PR #715, opened 2019-11-12
targets: explicitly mark the stack as NOLOAD

### LINKED ISSUE #594: ROM segments are non-contiguous with LLVM9
When building with LLVM 9, all of the smoke tests except wasm and avr fail with this error:
```
tinygo build -size short -o test.hex -target=pca10040            examples/blinky1
   5340      60     196 |    5400     256
error: ROM segments are non-contiguous: /tmp/tinygo703550982/main
```
There's not much more information, so I don't know if LLVM is stricter or tinygo is outputting something wrong now.

aykevl:
This prevents it from being of type PROGBITS in lld 9, it should always be NOBITS. This should fix the following error in lld 9:

    ROM segments are non-contiguous

Note: I haven't yet tested this.

fixes #594

### DIFF
--- a/targets/arm.ld
+++ b/targets/arm.ld
@@ -19,7 +19,7 @@ SECTIONS
     /* Put the stack at the bottom of RAM, so that the application will
      * crash on stack overflow instead of silently corrupting memory.
      * See: http://blog.japaric.io/stack-overflow-protection/ */
-    .stack :
+    .stack (NOLOAD) :
     {
         . = ALIGN(4);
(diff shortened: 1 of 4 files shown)


aykevl:
All the smoke tests have identical output, so I'm assuming that nothing changed when using `ld.lld-8`.

deadprogram:
All of my testing on LLVM8 worked as expected. Merging, hopefully this will allow us to get LLVM 9 support in for the next release!

Issue #595, opened 2019-09-26
Builtins' symbol index is no longer optional


QuLogic:
In `compileBuiltins`, there is a comment:
https://github.com/tinygo-org/tinygo/blob/cc313798f381532e4874a7166fa3c216234c005d/builtins.go#L248-L250
But with LLVM9, this appears no longer true. Running smoke tests fail to link:
```
tinygo build -size short -o test.hex -target=pca10040            examples/blinky1
ld.lld: error: .../compiler-rt.a: archive has no index; run ranlib to add one
error: failed to link /tmp/tinygo976334510/main: exit status 1
```

I could not find any symbol index writer in blakesmith/ar, so for now I just run `ranlib` before installing the `compiler-rt.a` files.

aykevl:
Oops, this definitely should be fixed. Although you can work around it by running `ranlib` and that should be sufficient for a distro package, I would really like to avoid requiring such an external tool (that's why TinyGo itself started generating these files in the first place). I'll see what I can do.

aykevl:
This is the relevant LLVM change: https://reviews.llvm.org/D63781

MaskRay:
This change is intentional. GNU linkers do not accept archives without a symbol table. lld used to be loose before https://reviews.llvm.org/D63781

```
% ld.bfd a.o a.a
ld.bfd: a.a: error adding symbols: archive has no index; run ranlib to add one
% gold a.o a.a
gold: error: a.a: no archive symbol table (run ranlib)
```

sbc100:
IIUC, before this change, lld would silently ignore libraries without symbol tables.  So I'm surprised this ever worked here (or perhaps there was a different change involved?).

aykevl:
In LLVM 8, lld was definitely able to read archives without a symbol table and we've relied on that behavior. For example, it was possible to use floating point operations on a Cortex-M0 (that has no FPU) after linking with compiler-rt, and this library definitely did not have a symbol table. This behavior breaks now we're trying to upgrade to LLVM 9.

I tried adding an empty symbol table for lld 9 (first file set to name `/` and containing 4 null bytes) but lld reports lots of error of undefined symbols. So it looks like either lld is now unable to read the individual members or (perhaps more likely) doesn't even attempt to read them when it sees a symbol table.

[...]

PR #717, opened 2019-11-12
builder: write a symbol table when writing out the compiler-rt lib

aykevl:
This should fix an issue with LLVM 9 (specifically, ld.lld-9).
See: https://github.com/tinygo-org/tinygo/issues/595

### DIFF
--- a/builder/ar.go
+++ b/builder/ar.go
@@ -1,0 +1,182 @@
+package builder
+
+import (
+	"bytes"
+	"debug/elf"
+	"encoding/binary"
+	"errors"
+	"io"
+	"os"
+	"path/filepath"
+	"time"
+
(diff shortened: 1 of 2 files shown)


aykevl:
All smoketests have identical outputs, so this PR should not have any side effects on LLVM 8.

deadprogram:
I suppose my only comment is that I would suggest implementing this in the form of a new function aka `createSymbolTable()` in part because the `CompileBuiltins()` is quite a few lines of code already.

aykevl:
That's a fair request. I did it slightly differently: I made a function that builds an archive file (with symbol table) from a list of object files - this makes it reusable. I also replaced a panic with an error return.

deadprogram:
Looks good, now merging.

PR #718, opened 2019-11-13
builder: make Clang header detection more robust

aykevl:
The header detection code failed way too easily, bailing out when there was more than one Clang version directory.

This fixes the following problem in LLVM 9 on Debian:

    testdata/cgo/main.h:1:10: fatal: 'stdbool.h' file not found
    testdata/cgo/main.go:5:10: note: in file included from testdata/cgo/main.go!cgo.c:3:

On my system, this was the cause for the error (more than one version). Only the `9.0.1` directory contains the header files:

```
$ ls -l /usr/lib/llvm-9/lib/clang/
total 0
drwxr-xr-x 1 root root  6 okt 27 14:40 9/
drwxr-xr-x 1 root root 36 nov  9 17:23 9.0.0/
drwxr-xr-x 1 root root 30 okt 27 14:40 9.0.1/
```

### DIFF
--- a/builder/env.go
+++ b/builder/env.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


deadprogram:
:mag_right: merging.

PR #719, opened 2019-11-13
all: use compiler-rt from the llvm-project directory

aykevl:
We don't need the separate submodule: compiler-rt is already included in the llvm-project repository.

This should hopefully make CI slightly faster too. Also, it will make sure compiler-rt is automatically upgraded on each LLVM version upgrade.


deadprogram:
:scissors: merging.

PR #720, opened 2019-11-13
fix incorrect starting value for optimized allocations in a loop

jadr2ddude:
Previously, we only zeroed stack allocations once in the entry block. However, when the original allocation is in a loop, it is necessary to zero the allocation on every iteration. This PR moves the allocation zeroing into the location where the `runtime.alloc` call used to be.

### DIFF
--- a/transform/transform_test.go
+++ b/transform/transform_test.go
@@ -70,8 +70,9 @@ func fuzzyEqualIR(s1, s2 string) bool {
 func filterIrrelevantIRLines(lines []string) []string {
 	var out []string
 	for _, line := range lines {
-		line = strings.TrimSpace(line) // drop '\r' on Windows
-		if line == "" || line[0] == ';' {
+		line = strings.Split(line, ";")[0]    // strip out comments/info
+		line = strings.TrimRight(line, "\r ") // drop '\r' on Windows and remove trailing spaces from comments
+		if line == "" {
 			continue
 		}
 		if strings.HasPrefix(line, "source_filename = ") {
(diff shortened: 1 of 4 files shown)


transform/transform_test.go
```diff
@@ -70,8 +70,10 @@ func fuzzyEqualIR(s1, s2 string) bool {
 func filterIrrelevantIRLines(lines []string) []string {
 	var out []string
 	for _, line := range lines {
-		line = strings.TrimSpace(line) // drop '\r' on Windows
-		if line == "" || line[0] == ';' {
+		line = strings.TrimSuffix(line, "\r") // drop '\r' on Windows
+		line = strings.Split(line, ";")[0]    // strip out comments/info
+		line = strings.TrimRight(line, " ")   // remove trailing spaces from comments
```
aykevl: Why is this necessary? It seems to be unnecessary for the PR here.

The new code appears to catch comments at the end of a line, but I think that can be done more cleanly this way:

```go
line = strings.Split(line, ";")[0] // strip out comments/info
line = strings.TrimSpace(line)     // drop '\r' on Windows and remove trailing spaces from comments
```

jadr2ddude: I took out the `strings.TrimSpace` because it also strips leading spaces, which is actually incorrect because it strips out the indentation.

jadr2ddude: I took out the `strings.TrimSpace` because it also strips leading spaces, which is actually incorrect because it strips out the indentation. This is more precise.
LLVM generates comments at the end of label lines.

aykevl: Indentation is not semantically relevant: you could use any kind of indentation in IR files.
The intention of this function was to compare whether two pieces of IR have any actual differences (excluding comments, whitespace, and other irrelevant details).

aykevl: >  LLVM generates comments at the end of label lines.

Ah, right. Good point.

PR #677, opened 2019-10-31
Refactor compile options

aykevl:
This is a large refactoring PR, split into multiple pieces to make it easier to review. The first three commits just move things around, the last commit actually changes anything.

It should not affect any build. If it does, that's a bug.

PR #721, opened 2019-11-13
Refactor interface lowering

aykevl:
This is something I've wanted to do for a long time, but couldn't do before doing other refactors (in particular: https://github.com/tinygo-org/tinygo/pull/677). It should not change any behavior, only move code around.

The first commit adds a llvmutil package. I couldn't think of a better way to do it: the code that's moved is hard to get right and should really not be copied. If you know of a better way, please let me know. Also, please let me know if you can think of a better package name/location.

Tests will be added soon, submitting now for initial review.


aykevl:
And now with tests.

compiler/llvm.go
```diff
@@ -24,79 +23,33 @@ func getUses(value llvm.Value) []llvm.Value {
...
 func (c *Compiler) createTemporaryAlloca(t llvm.Type, name string) (alloca, bitcast, size llvm.Value) {
-	alloca = c.createEntryBlockAlloca(t, name)
-	bitcast = c.builder.CreateBitCast(alloca, c.i8ptrType, name+".bitcast")
-	size = llvm.ConstInt(c.ctx.Int64Type(), c.targetData.TypeAllocSize(t), false)
-	c.builder.CreateCall(c.getLifetimeStartFunc(), []llvm.Value{size, bitcast}, "")
-	return
-}
-
-// createInstructionAlloca creates an alloca in the entry block, and places lifetime control intrinsics around the instruction
-func (c *Compiler) createInstructionAlloca(t llvm.Type, inst llvm.Value, name string) llvm.Value {
```
jadr2ddude: Is there a reason that this is handled differently than the rest?

aykevl: The only reason is that `createInstructionAlloca` was only called once so I just modified the caller.
That's not a very good reason, and I could certainly add a wrapper here. No real preference by me either way.

jadr2ddude: idk, doesnt make too much of a difference

PR #723, opened 2019-11-14
nrf: support for SoftDevice s140 PCA10056 board

thegrumpylion:
My source for the linker script file are the ble examples in nRF5 SDK v16.0.0. I am not sure if the values are correct but for the FLASH_TEXT the value is the same in all examples for pca10056. Now for the RAM origin seems to be application dependent. Is this correct?


aykevl:
> I am not sure if the values are correct but for the FLASH_TEXT the value is the same in all examples for pca10056. Now for the RAM origin seems to be application dependent. Is this correct?

That's correct. The flash size is fixed for a major revision number and the RAM size is application-dependent. However, it's currently fixed in TinyGo.

Sources:
* https://infocenter.nordicsemi.com/index.jsp?topic=%2Fsds_s140%2FSDS%2Fs1xx%2Fmem_usage%2Fmem_resource_map_usage.html&cp=4_5_3_0_13_0
* https://devzone.nordicsemi.com/nordic/short-range-guides/b/getting-started/posts/adjustment-of-ram-and-flash-memory

deadprogram:
Would this be the same basic setup to use with the reel board if loaded with the s140 softdevice?

thegrumpylion:
> However, it's currently fixed in TinyGo.
And that value is 0x20000000 + 0x000039c0 as in S132? Also, can you please explain why is fixed in TinyGo? Is it fixed by choice or some other factors?

> Actually, can you also add a smoketest to make sure this configuration keeps compiling in the future?
Sure.

> Would this be the same basic setup to use with the reel board if loaded with the s140 softdevice?
It should be (tm). I have also an nRF52840-mdk board that i would like to have S140 working on it but needs a bit of work as in creating a single .hex file since the bootloader is CMSIS-DAP and `nrfjprog` doesn't work

aykevl:
> And that value is 0x20000000 + 0x000039c0 as in S132? Also, can you please explain why is fixed in TinyGo? Is it fixed by choice or some other factors?

The RAM used/required by the SoftDevice can be configured, but must be set at compile time (because it is part of the linker script). Ideally you could configure this per application and eventually support for this should be added, but until then I think it's best to pick a reasonable number and use that. 0x39c0 (~14.4kB) should definitely be enough.

The amount of RAM required varies by application because different applications may use different features of the SoftDevice. This is configured when the SoftDevice is initialized, but at the moment go-bluetooth uses a fixed configuration with a fixed amount of memory required.
For more information, see [`sd_ble_cfg_set`](https://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.s132.api.v6.1.0%2Fgroup___b_l_e___c_o_m_m_o_n___f_u_n_c_t_i_o_n_s.html&cp=4_5_2_3_2_0_2_2_0&anchor=ga4edae2bac8c68b672c0fa101ed2c687f) and [`sd_ble_enable`](https://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.s132.api.v6.1.0%2Fgroup___b_l_e___c_o_m_m_o_n___f_u_n_c_t_i_o_n_s.html&cp=4_5_2_3_2_0_2_2_0&anchor=ga4edae2bac8c68b672c0fa101ed2c687f).

> I have also an nRF52840-mdk board that i would like to have S140 working on it but needs a bit of work as in creating a single .hex file since the bootloader is CMSIS-DAP and nrfjprog doesn't work

I believe you can also flash the SoftDevice using OpenOCD (it certainly worked on nrf51 devices). After that, you can just use `tinygo flash -taget=nrf52840-mdk`.

thegrumpylion:
@aykevl Thank you for the information and for the pointers. Everything is clear now.

aykevl:
Thank you for this contribution, merged now.

thegrumpylion:
Yay! My first contribution to TinyGo!! The first of many i hope :)

PR #729, opened 2019-11-16
compiler: move ReplacePanicsWithTrap pass to transforms

aykevl:
This moves the transformation pass to the right location, and adds tests to see that it actually works correctly.

### DIFF
--- a/transform/panic.go
+++ b/transform/panic.go
@@ -1,0 +1,33 @@
+package transform
+
+import (
+	"tinygo.org/x/go-llvm"
+)
+
+// ReplacePanicsWithTrap replaces each call to panic (or similar functions) with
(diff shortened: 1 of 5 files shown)


transform/panic.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"tinygo.org/x/go-llvm"
+)
+
+// ReplacePanicsWithTrap replaces each call to panic (or similar functions) with
+// calls to llvm.trap, to reduce code size. This is the -panic=trap command-line
+// option.
+func ReplacePanicsWithTrap(mod llvm.Module) {
+	builder := llvm.NewBuilder()
```
jadr2ddude: This should use a context. Maybe it should be `mod.Context().NewBuilder()`?

aykevl: Good catch! Fixed now.

PR #730, opened 2019-11-16
runtime/atsamd21: i2s initialization fixes

Infinoid:
I found these during code review, they look like bugs to me.  Unfortunately, I don't have atsamd21 hardware to test against, so please review carefully.

1. The I2S_SERCTRL_DATASIZE_* constants are 0-based, but DATASIZE is not at offset 0 within the SERCTRL register.  It should be shifted by 8, otherwise it will clobber the SERMODE and TXDEFAULT fields.
2. The Philips I2S spec is MSB-first.  The BITREV flag, when set, causes LSB-first operation.  I think the flag should be should be set in I2SStandardLSB mode, and clear in I2SStandardMSB and I2SStandardPhilips modes.

I think this PR will cause the `src/examples/i2s/i2s.go` example to print samples in the opposite order.  Were the samples printed correctly before?

### DIFF
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -877,29 +877,29 @@ func (i2s I2S) Configure(config I2SConfig) {
 	// now set serializer data size.
 	switch config.DataFormat {
 	case I2SDataFormat8bit:
-		i2s.Bus.SERCTRL1.SetBits(sam.I2S_SERCTRL_DATASIZE_8)
+		i2s.Bus.SERCTRL1.SetBits(sam.I2S_SERCTRL_DATASIZE_8 << sam.I2S_SERCTRL_DATASIZE_Pos)
 
 	case I2SDataFormat16bit:
-		i2s.Bus.SERCTRL1.SetBits(sam.I2S_SERCTRL_DATASIZE_16)
+		i2s.Bus.SERCTRL1.SetBits(sam.I2S_SERCTRL_DATASIZE_16 << sam.I2S_SERCTRL_DATASIZE_Pos)
 
 	case I2SDataFormat24bit:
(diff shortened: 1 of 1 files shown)


Infinoid:
The SERCTRL register definition can be found at http://ww1.microchip.com/downloads/en/DeviceDoc/SAMD21-Family-DataSheet-DS40001882D.pdf section 29.9.7.

The Philips I2S spec can be found at https://web.archive.org/web/20070102004400/http://www.nxp.com/acrobat_download/various/I2SBUS.pdf .  MSB-first operation is described in section 3.1, and shown in Figure 1.

src/machine/machine_atsamd21.go
```diff
@@ -877,29 +877,29 @@ func (i2s I2S) Configure(config I2SConfig) {
...
 		i2s.Bus.SERCTRL1.ClearBits(sam.I2S_SERCTRL_SLOTADJ)
+
+		// transfer LSB first
+		i2s.Bus.SERCTRL1.SetBits(sam.I2S_SERCTRL_BITREV)
 	} else {
 		// adjust left
 		i2s.Bus.SERCTRL1.SetBits(sam.I2S_SERCTRL_SLOTADJ)
-
-		// reverse bit order?
-		i2s.Bus.SERCTRL1.SetBits(sam.I2S_SERCTRL_BITREV)
```
deadprogram: I suggest instead of removing, change this line to `i2s.Bus.SERCTRL1.ClearBits(sam.I2S_SERCTRL_BITREV)` to make it more obvious what is happening.

deadprogram:
Seems like you are correct, very solid research, thank you for that.

The only code I have used the I2S peripheral with is https://github.com/tinygo-org/drivers/blob/master/microphone/microphone.go

I suspect that since all I did for that demo was to use a mic as a simple a VU meter, the fact that the samples were reversed was not noticed.

Thanks for the PR, now merging.

Issue #728, opened 2019-11-16
stack chain start not found!


ellemlabs:
After compiling a wasm file, if I attempt to invoke it using the typical means described in https://tinygo.org/webassembly/webassembly/, I get the error:

```
Uncaught (in promise) LinkError: WebAssembly.instantiate(): Import #12 module="env" function="memcpy" error: function import requires a callable
```

I'm not sure where to start with this. I believe it has something to do with the inner-workings of wasm_exec.js. I'm running the copy included with 0.9, but I've noticed it hasn't changed sense 0.8.

Regardless, is there anyway to debug this?

ellemlabs:
Using a `wasm2wat` and looking through the imports, it appears it's referencing to $memcpy, but that does not seem to be declared beforehand. I don't know too much about wasm or wat but this is certainly a start.

```
(module
  (type (;0;) (func (param i32)))
  (type (;1;) (func (result i32)))
  (type (;2;) (func (param i32 i32 i32) (result i32)))
  (type (;3;) (func (result f64)))
  (type (;4;) (func (param i32 i32 i32 i32 i32 i32 i32 i32 i32)))
  (type (;5;) (func (param i32 i32 i32 i32 i32 i32)))
  (type (;6;) (func (param i32 i32 i32 i32)))
  (type (;7;) (func (param i32 i32 i32 i32 i32)))
  (type (;8;) (func (param i32 i32 i32 i32 i32 i32 i32)))
  (type (;9;) (func))
  (type (;10;) (func (param i32 i32)))
  (type (;11;) (func (param i32) (result i32)))
  (type (;12;) (func (param i32 i32 i32 i32) (result i32)))
  (type (;13;) (func (param i32 i32) (result i32)))
  (type (;14;) (func (param i32 i32 i32)))
  (type (;15;) (func (param i64 i32 i32) (result i64)))
  (type (;16;) (func (param i64) (result i32)))
  (type (;17;) (func (param i32 i64)))
  (type (;18;) (func (param i32 i32 i32) (result f64)))
  (type (;19;) (func (param i32 i32 i32) (result i64)))
  (type (;20;) (func (param i32 i64 i64 i64 i64 i64) (result i32)))
  (type (;21;) (func (param i32 i64 i64 i32 i64) (result i32)))
  (type (;22;) (func (param i32 i32 i32 i32 i32 i64 i32 i32)))
[...]

PR #732, opened 2019-11-17
add code to handle programs which use heap allocations but never hit the GC

jadr2ddude:
Fixes the bug shown in https://github.com/tinygo-org/tinygo/issues/728#issuecomment-554684216, where a piece of code only calls `runtime.alloc` once and does a zero-size allocation.


aykevl:
To be honest, I don't like the addition of yet another independent test, that slows down CI. Perhaps it would be possible to move the GC pass to the transform package, and add a unit test instead?

Merging, because this fixes an actual bug.

PR #735, opened 2019-11-17
interp: fix inserting non-const values in a const aggregate

aykevl:
This bug was triggered by the following code:

    package main

    func foo() byte

    var array = [1]byte{foo()}

    func main() {
    }


deadprogram:
OK, merging.

Issue #514, opened 2019-08-19
Blocking interface calls


jadr2ddude:
Whenever we create a blocking implementation of an interface (at least with the mode using LLVM coroutines), this generates an inconsistency which prevents the code from compiling. We need to properly make interface wrappers blocking when necessary.

jadr2ddude:
This is a prerequisite for making mutexes block.

deadprogram:
Has this now been implemented?

aykevl:
I don't think so. I haven't done any work in that area.

PR #736, opened 2019-11-17
Blocking interface calls

jadr2ddude:
This PR fixes up function typing so that the interface `$invoke` and dispatch functions can be processed by the goroutine lowering pass. As this is currently structured, these functions will never need to be turned into coroutines because all blocking calls are tail calls.

### DIFF
--- a/compiler/interface.go
+++ b/compiler/interface.go
@@ -461,7 +461,9 @@ func (c *Compiler) getInterfaceInvokeWrapper(f *ir.Function) llvm.Value {
 	paramTypes := append([]llvm.Type{c.i8ptrType}, fnType.ParamTypes()[len(expandedReceiverType):]...)
 	wrapFnType := llvm.FunctionType(fnType.ReturnType(), paramTypes, false)
 	wrapper = llvm.AddFunction(c.mod, wrapperName, wrapFnType)
-	wrapper.LastParam().SetName("parentHandle")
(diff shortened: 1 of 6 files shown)


jadr2ddude:
fixes #514

compiler/interface.go
```diff
@@ -461,6 +461,7 @@ func (c *Compiler) getInterfaceInvokeWrapper(f *ir.Function) llvm.Value {
 	paramTypes := append([]llvm.Type{c.i8ptrType}, fnType.ParamTypes()[len(expandedReceiverType):]...)
 	wrapFnType := llvm.FunctionType(fnType.ReturnType(), paramTypes, false)
 	wrapper = llvm.AddFunction(c.mod, wrapperName, wrapFnType)
+	wrapper.LastParam().SetName("parentHandle")
```
aykevl: Can you add a check whether the to-be-wrapped function also has this name? That would protect against future changes that might break this assumption.

```suggestion
	if f.LLVMFn.LastParam().Name() == "parentHandle" {
		wrapper.LastParam().SetName("parentHandle")
	}
```

jadr2ddude:
Alright, pushed your change. Interesting that GitHub now has a button for this.

aykevl:
Thank you! Merged in 98eee7c22ab8f790a1a712619b105a1f9a45e271.

PR #737, opened 2019-11-17
compiler: move GC passes to the transform package

aykevl:
~~Not yet complete, I also need to add tests for `MakeGCStackSlots`.~~


aykevl:
More tests added.
Adding tests to GC stack slot creation code showed that there is a lot of low hanging fruit there.

aykevl:
To be clear: this PR is now complete and ready for review.

PR #738, opened 2019-11-19
Add SPI1 connected to NINA-W102 chip on Arduino Nano 33 IOT

bgould:
This PR adds a SPI interface called SPI1 on SERCOM2 for the connection to the integrated ESP chip.  Additionally it adds aliases NINA_TX, NINA_RX, and NINA_SPI to make use of this connection more portable across boards.


deadprogram:
Looks good, thanks for adding this @bgould now merging.

PR #739, opened 2019-11-19
Fixes for UART2 on Metro M4 Airlift Lite

bgould:
My initial implementation for UART2 on this Metro M4 Airlift Lite was simply incorrect.  This PR adds handlers to atsamd51.go so that this actually works.

### DIFF
--- a/src/machine/board_feather-m4.go
+++ b/src/machine/board_feather-m4.go
@@ -53,6 +53,12 @@ const (
 	UART_RX_PIN = D0
 )
 
+// UART2 pins
+const (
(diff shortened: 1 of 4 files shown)


deadprogram:
Hi @bgould the ItsyBitsy-M4 needs to have `UART2` defined, which is the reason that the build is failing.

```
src/machine/machine_atsamd51.go:680:2: undeclared name: UART2
```

Thanks!

deadprogram:
Hi @bgould thanks for the corrections to get the build passing. Now merging.

PR #741, opened 2019-11-19
runtime: use MSP/PSP registers for scheduling on Cortex-M

aykevl:
The Cortex-M architecture contains two stack pointers, designed to be used by RTOSes: MSP and PSP (where MSP is the default at reset). In fact, the ARM documentation recommends using the PSP for tasks in a RTOS.

This commit switches to using the PSP for goroutine stacks. Aside from being the recommended operation, this has the big advantage that the NVIC automatically switches to the MSP when handling interrupts. This avoids having to make every goroutine stack big enough that interrupts
can be handled on it.

Additionally, I've optimized the assembly code to save/restore registers (made possible by this change). For Cortex-M3 and up, saving all registers is just a single push instruction and restoring+branching is a single pop instruction. This results in a small code size improvement and probably a speed improvement as well (untested).

Sidenote: the fact that you can pop a number of registers and branch at the same time makes ARM not exactly a true RISC system. However, it's very useful in this case.

@deadprogram I think you may want to test this PR on some of your demos, to make sure it doesn't break anything important.

### DIFF
--- a/src/runtime/scheduler_cortexm.S
+++ b/src/runtime/scheduler_cortexm.S
@@ -19,76 +19,96 @@ tinygo_startTask:
     // After return, exit this goroutine. This is a tail call.
     bl    runtime.yield
 
+.section .text.tinygo_getSystemStackPointer
+.global  tinygo_getSystemStackPointer
+.type    tinygo_getSystemStackPointer, %function
+tinygo_getSystemStackPointer:
+    // The system stack pointer is always stored in the MSP register.
+    mrs r0, MSP
+    bx lr
+
+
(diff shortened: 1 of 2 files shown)


aykevl:
Tweaked the assembly a little bit to be a bit shorter (and possibly more efficient) on Cortex-M0.

deadprogram:
Will start testing this over the next couple of days.

deadprogram:
Ended up having some time here at the airport to do some testing, and had several M0 and M4 boards with me. I have run this branch against several of my most complex demos and worked flawlessly.

Merging, great work @aykevl

PR #745, opened 2019-11-22
revise defer to use heap allocations when running a variable number of times

jaddr2line:


compiler/defer.go
```diff
@@ -127,12 +128,22 @@ func (c *Compiler) emitDefer(frame *Frame, instr *ssa.Defer) {
...
+		alloca = llvmutil.CreateEntryBlockAlloca(c.builder, deferFrameType, "defer.alloca")
+		if c.NeedsStackObjects() {
+			c.trackPointer(alloca)
+		}
+	} else {
+		// This may be hit a variable number of times, so use a heap allocation.
+		size := c.targetData.TypeAllocSize(deferFrameType)
+		sizeValue := llvm.ConstInt(c.uintptrType, size, false)
+		allocCall := c.createRuntimeCall("alloc", []llvm.Value{sizeValue}, "defer.alloc.call")
+		alloca = c.builder.CreateBitCast(allocCall, llvm.PointerType(deferFrameType, 0), "defer.alloc")
```
aykevl: This `alloca` should be tracked as well. So I would suggest moving the above lines:

```go

		if c.NeedsStackObjects() {
			c.trackPointer(alloca)
		}
```
to below this if/else.

compiler/defer.go
```diff
@@ -127,12 +128,22 @@ func (c *Compiler) emitDefer(frame *Frame, instr *ssa.Defer) {
...
 	}
 
-	// Put this struct in an alloca.
-	alloca := c.builder.CreateAlloca(deferFrameType, "defer.alloca")
-	c.builder.CreateStore(deferFrame, alloca)
-	if c.NeedsStackObjects() {
-		c.trackPointer(alloca)
+	// Put this struct in an allocation.
+	var alloca llvm.Value
+	if instr.Block().Dominates(instr.Parent().Blocks[0]) {
```
aykevl: I've been thinking more about this, and I think it's not a complete solution. I think it's safe, but take the following code for example:

```go
func foo() {
    if a {
        return
    }
    defer bar()
}
```

As I understand dominance theory, block at `defer bar()` does not dominate the entry block and thus the defer would be heap allocated (while there is no good reason for that).

I'm pretty sure there must be an easy way to calculate whether some block can be executed more than once, but can't really think of one. This problem must be solved a long time ago as it is important for compiler optimizations.

deadprogram:
Any action happening on this PR?

jaddr2line:
It needs some major changes and I do not have time to implement them right now.

deadprogram:
Understood. Can just let it hang out until ready to land, then.

jaddr2line:
Alright, I updated the strategy to use a BFS to scan backwards. A defer will be stack-allocated unless there exists a path from the defer's block to itself. The only situation in which this would be slow that I can think of is something along the lines of:

```
// Thousands of basic blocks.

defer something()
defer something()
// thousands of defers
```

If a defer is in the entry block then the loop should only run once.

compiler/defer.go
```diff
@@ -34,6 +35,40 @@ func (c *Compiler) deferInitFunc(frame *Frame) {
 	c.builder.CreateStore(llvm.ConstPointerNull(deferType), frame.deferPtr)
 }
 
+// isCyclic checks if there is a path from a basic block to itself.
+func isCyclic(start *ssa.BasicBlock) bool {
```
aykevl: I think the name `isCyclic` is not immediately obvious in what it means. What about `isInLoop`?

aykevl:
Kicking the Windows build again...

aykevl:
Great! Thank you for working on this!

PR #746, opened 2019-11-22
Refactor compiler

aykevl:
Work-in-progress to refactor the compiler.

Not nearly finished, but sharing here for initial review to make sure this is the right direction. The main goal is to remove `Compiler.builder`, which is almost like a global variable (shared throughout the entire compiler instance). Instead, an IR builder should be made per function. This will make the structure of the compiler a lot simpler and should avoid lots of jumping around, for example to create wrapper functions. The last time I tried to remove the `SimpleDCE` mechanism (a blocker for building packages independently), wrapper functions were a major pain and resulted in needing to add a worklist for functions to be built after the current one is finished.

This change will be huge, so I decided to split it in many steps to make it reviewable. It could even be done in multiple PRs (if needed I could also split this PR up in multiple pieces). Most commits should be easy to review as the vast majority of the changes are mechanical (search+replace).

Some code is duplicated while the refactor is in progress. This duplication will be removed once it is finished.

### DIFF
--- a/compiler/calls.go
+++ b/compiler/calls.go
@@ -43,10 +43,10 @@ func (c *Compiler) createCall(fn llvm.Value, args []llvm.Value, name string) llv
 
 // Expand an argument type to a list that can be used in a function call
 // parameter list.
-func (c *Compiler) expandFormalParamType(t llvm.Type) []llvm.Type {
+func expandFormalParamType(t llvm.Type) []llvm.Type {
 	switch t.TypeKind() {
 	case llvm.StructTypeKind:
-		fields := c.flattenAggregateType(t)
+		fields := flattenAggregateType(t)
 		if len(fields) <= MaxFieldsPerParam {
 			return fields
 		} else {
@@ -82,7 +82,7 @@ func (c *Compiler) expandFormalParamOffsets(t llvm.Type) []uint64 {
 func (c *Compiler) expandFormalParam(v llvm.Value) []llvm.Value {
 	switch v.Type().TypeKind() {
 	case llvm.StructTypeKind:
-		fieldTypes := c.flattenAggregateType(v.Type())
+		fieldTypes := flattenAggregateType(v.Type())
 		if len(fieldTypes) <= MaxFieldsPerParam {
 			fields := c.flattenAggregate(v)
 			if len(fields) != len(fieldTypes) {
@@ -101,12 +101,12 @@ func (c *Compiler) expandFormalParam(v llvm.Value) []llvm.Value {
 
 // Try to flatten a struct type to a list of types. Returns a 1-element slice
 // with the passed in type if this is not possible.
-func (c *Compiler) flattenAggregateType(t llvm.Type) []llvm.Type {
+func flattenAggregateType(t llvm.Type) []llvm.Type {
 	switch t.TypeKind() {
 	case llvm.StructTypeKind:
 		fields := make([]llvm.Type, 0, t.StructElementTypesCount())
 		for _, subfield := range t.StructElementTypes() {
-			subfields := c.flattenAggregateType(subfield)
+			subfields := flattenAggregateType(subfield)
 			fields = append(fields, subfields...)
 		}
 		return fields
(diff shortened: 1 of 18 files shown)


aykevl:
This requires #773 to be merged first, to make this a cleaner PR.

aykevl:
Finally, this PR is ready for review. It doesn't fix the entire public interface of the compiler package (it still exposes `compiler.Phi` for example which is an internal detail) but it's a lot better now. But far more importantly, it removes the semi-global `llvm.Builder` used throughout the compiler and instead moves to a `compiler.builder`. This means that the state used during the compilation of a single function is now actually bound to a single function and not included in the semi-global `compilerContext` object.

It is a very large PR but it is split into many relatively small commits. The last time I tested, all but one commit can be compiled and tested individually, making it easy to spot regressions if there are any.

compiler/llvm.go
```diff
@@ -29,34 +29,34 @@ func getUses(value llvm.Value) []llvm.Value {
...
+func (b *builder) emitPointerPack(values []llvm.Value) llvm.Value {
+	return llvmutil.EmitPointerPack(b.Builder, b.mod, b.Config, values)
 }
 
 // emitPointerUnpack extracts a list of values packed using emitPointerPack.
-func (c *Compiler) emitPointerUnpack(ptr llvm.Value, valueTypes []llvm.Type) []llvm.Value {
-	return llvmutil.EmitPointerUnpack(c.builder, c.mod, ptr, valueTypes)
+func (b *builder) emitPointerUnpack(ptr llvm.Value, valueTypes []llvm.Type) []llvm.Value {
+	return llvmutil.EmitPointerUnpack(b.Builder, b.mod, ptr, valueTypes)
 }
```
jaddr2line: You seem to be changing the terminology from `emit` to `create`. Should these be changed to make it consistent?

compiler/compiler.go
```diff
@@ -364,26 +386,34 @@ func (c *Compiler) Compile(mainPath string) []error {
...
 // it as a LLVM type, creating it if necessary. It is a shorthand for
 // getLLVMType(getRuntimeType(name)).
-func (c *Compiler) getLLVMRuntimeType(name string) llvm.Type {
-	return c.getLLVMType(c.getRuntimeType(name))
+func (c *compilerContext) getLLVMRuntimeType(name string) llvm.Type {
+	fullName := "runtime." + name
+	typ := c.mod.GetTypeByName(fullName)
+	if typ.IsNil() {
+		println(c.mod.String())
+		panic("could not find runtime type: " + fullName)
```
jaddr2line: Again, try to use `fmt.Errorf`.

compiler/calls.go
```diff
@@ -13,40 +11,35 @@ import (
...
-	runtimePkg := c.ir.Program.ImportedPackage("runtime")
-	member := runtimePkg.Members[fnName]
-	if member == nil {
-		panic("trying to call runtime." + fnName)
+// createCall creates a new call to runtime.<fnName> with the given arguments.
+func (b *builder) createRuntimeCall(fnName string, args []llvm.Value, name string) llvm.Value {
+	fullName := "runtime." + fnName
+	fn := b.mod.NamedFunction(fullName)
+	if fn.IsNil() {
+		panic("trying to call non-existing function: " + fullName)
```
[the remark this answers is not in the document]
aykevl: Fixed the grammar.

I definitely agree that a real error message is better. However, that would also change lots of code so I want to do such refactoring at a later time to not complicate this one further.

compiler/llvm.go
aykevl: Yes, but I think that's best done in a separate PR. And in this particular case I want `llvmutil.EmitPointerUnpack` and `(b *builder) emitPointerUnpack` to match.

compiler/compiler.go
aykevl: Changing that here would imply changing lots of other places, which I think should be done separately. In fact, the panic here is already an improvement over the previous situation, which would probably panic with a nil pointer dereference or something similar.

deadprogram:
I spent a while reviewing the changes, which indeed were quite numerous. That said, it was a true refactoring and not a rewrite, so although tedious to review, it was consistent. Thanks for that @aykevl 

I then turned my attentions to some empirical testing using my latest collection of demos plus a couple of test programs/boards I have been using lately. That also took a while, but all of the code/hardware functioned exactly as expected.

Therefore, I am going to merge this PR. Hopefully it will help current and future contributors navigate the compiler code.

PR #748, opened 2019-11-24
interp: improve error reporting

aykevl:
This commit improves error reporting in several ways:

  * Location information is read from the intruction that causes the
    error, as far as that's available.
  * The package that is being interpreted is included in the error
    message. This may be the most useful part of the improvements.
  * The hashmap update intrinsics now doesn't panic, instead it logs a
    clear error (with location information, as in the first bullet
    point).

This is possible thanks to improvements in LLVM 9. This means that after
this change, TinyGo will depend on LLVM 9.

Example of the hashmap error improvement, before:

    panic: interface conversion: interp.Value is *interp.LocalValue, not *interp.MapValue

After:

```
# go/token
../../../../../usr/local/go/src/go/token/token.go:282:17: could not update map with string key
```

(Reading the relevant source code, this looks like a fixable error).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,5 +9,5 @@ require (
 	go.bug.st/serial.v1 v0.0.0-20180827123349-5f7892a7bb45
 	golang.org/x/sys v0.0.0-20191010194322-b09406accb47 // indirect
 	golang.org/x/tools v0.0.0-20190227180812-8dcc6e70cdef
-	tinygo.org/x/go-llvm v0.0.0-20191113125529-bad6d01809e8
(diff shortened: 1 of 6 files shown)


deadprogram:
I'm all in favor of this change. We're already a little behind many others to adopt LLVM 9. @jaddr2line any opinion?

jaddr2line:
I think this seems ok. We may eventually want to restructure some of this code, but that is a separate issue and should be done later.

aykevl:
Agreed, some code may need to be refactored a little bit to make it possible to replace the remaining panics with a normal error message. But I decided to do that at a later time, to keep this PR focused.

deadprogram:
:nine: merging.

PR #750, opened 2019-11-25
ci: build with Go 1.13 on Windows

aykevl:
The default Go version is 1.12. Because Go 1.13 introduced language changes (improved numeric constants), TinyGo compiled with Go 1.12 cannot handle Go 1.13 code such as the Go 1.13 standard library.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -146,6 +146,7 @@ tinygo-test:
 
 .PHONY: smoketest
 smoketest:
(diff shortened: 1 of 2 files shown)


aykevl:
This should now be ready.

deadprogram:
This will simplify things as well as helping Windows devs get the latest of everything. Now merging.

PR #754, opened 2019-11-26
interp: replace many panics with error messages

aykevl:
This PR replaces most panics in interp/frame.go and interp/scan.go with real error messages. The remaining ones are panics that should not happen when working with valid IR. It builds upon #748 to improve error handling in the interp package.

There are a number of panics left in interp/value.go, but replacing them will require some more restructuring so I will do that at a later time.


deadprogram:
:record_button: DON'T PANIC

merging...

PR #755, opened 2019-11-26
interp: make mapassign able to deal with non-constant values

aykevl:
TODO: tests, and extend this to `PutString`.


aykevl:
This may be useful eventually, but the original bug is now fixed in a different way (see #787). Closing for now.

PR #757, opened 2019-11-27
compiler: refactor func lowering to the transform package

aykevl:
This commit makes a number of changes:

  * It avoids a dependency on Compiler.emitStartGoroutine.
  * It moves the func-lowering pass to the transform package.
  * It adds testing to func lowering.

No functionality should have changed with this commit.


deadprogram:
:arrow_down: verified that this PR does not change any functionality of several of the more complex samples. Also very glad to see some tests on function lowering, which can be complex. Now merging.

PR #759, opened 2019-11-27
tools: avoid _paddingX in generated struct fields

aykevl:
This makes the generation script slightly simpler.

I checked whether any of the smoketest outputs changed with this, with the smoke tests of the main tinygo repository and the drivers repository. I could find only one difference: examples/apa102 in the drivers repository. However, when looking at the IR before and after, the difference was in some reflection data (which can indeed change when field names change). Therefore, I think that's an unrelated change (although it may be a good idea to test that example anyway).


deadprogram:
The APA102 example still works as before. Also, it really looks a lot better, as well as saving space.

Merging.

PR #760, opened 2019-11-28
tools: rewrite gen-device-avr in Go

aykevl:
This brings a big speedup. Not counting gofmt time, `make gen-device-avr` became about 3x faster. None of the generated files (Go, assembly, linker script) changed in the process.
In the future, it might be an idea to generate the AST in-memory and write it out already formatted.

Note: I haven't attempted any refactoring (apart from some obvious simplifications), that can be done at a later time.

This was somewhat of an experiment. Hopefully gen-device-svd.py can be rewritten next for an even bigger speedup (many more files to be created).


aykevl:
And now also in parallel. The total reduction of `make gen-device-avr` is now from about 12.5s to about 4s.

tools/gen-device-avr/gen-device-avr.go
```diff
@@ -0,0 +1,502 @@
...
+
+	maxInterruptNum := 0
+	for _, intr := range device.interrupts {
+		if intr.Index > maxInterruptNum {
+			maxInterruptNum = intr.Index
+		}
+	}
+
+	t := template.Must(template.New("go").Parse(`// Automatically generated file. DO NOT EDIT.
+// Generated by gen-device-avr.py from {{.metadata.file}}, see {{.metadata.descriptorSource}}
```
deadprogram: Should probably update this, to specify that the new Go-based generator,

tools/gen-device-avr/gen-device-avr.go
```diff
@@ -0,0 +1,502 @@
...
+func writeAsm(outdir string, device *Device) error {
+	// The interrupt vector, which is hard to write directly in Go.
+	out, err := os.Create(outdir + "/" + device.metadata["nameLower"].(string) + ".s")
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+	t := template.Must(template.New("asm").Parse(
+		`; Automatically generated file. DO NOT EDIT.
+; Generated by gen-device-avr.py from {{.file}}, see {{.descriptorSource}}
```
deadprogram: Same as above.

deadprogram:
Made a couple comments. Tested on DIgispark, worked as expected.

aykevl:
Fixed and rebased. Thanks for catching that!

deadprogram:
From personal experience, the ARM generator will be more difficult. It will also be that much sweeter to retire that code as well! Merging.

PR #762, opened 2019-11-29
riscv: use LLVM tools instead of GNU toolchain

aykevl:
Now that we use LLVM 9, RISC-V support in LLVM has far fewer bugs and we can avoid the GNU toolchain.

  * replace GNU linker with lld
  * replace GCC with clang

Additionally, RISC-V was promoted to stable so it can be enabled by default in CI.

WARNING: not confirmed to work on real hardware.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,7 +67,7 @@ commands:
       - run: go install .
       - run: go test -v ./cgo ./compileopts ./interp ./transform .
       - run: make gen-device -j4
-      - run: make smoketest RISCV=0
+      - run: make smoketest
       - save_cache:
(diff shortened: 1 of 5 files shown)


aykevl:
This PR doesn't work on real hardware, but will probably work once #764 is merged.

deadprogram:
I switched the branch to be merged into to be `dev` since that was probably an oversight. I then rebased it against `dev` since #764 has indeed been merged.

The HiFive1b blink worked exactly as expected.I am just waiting for the Windows CI build to finish before merging it.

deadprogram:
Well, also waiting for @aykevl to move from draft to ready to merge.

aykevl:
Rebased, now ready for review.

deadprogram:
This is really great! :scissors: now merging.

PR #763, opened 2019-11-29
Add support for debugging RISC-V using QEMU

aykevl:
With this PR, you can easily debug RISC-V code in QEMU. For example:

```
$ tinygo gdb -target=hifive1-qemu ./testdata/alias.go 
GNU gdb (GDB) 8.2.90.20190228-git
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "--host=x86_64-linux-gnu --target=riscv64-unknown-elf".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /tmp/tinygo689867347/main...
Remote debugging using :1234
0x00001000 in ?? ()
(gdb) c
Continuing.
x
apple
true
false
Remote connection closed
(gdb) 
```

You will currently need to have the following installed: `qemu-system-riscv` and `riscv64-unknown-elf-gdb`, both downloadable from SiFive: https://www.sifive.com/boards (scroll down to "Prebuilt RISC‑V GCC Toolchain and Emulator").

### DIFF
--- a/main.go
+++ b/main.go
@@ -268,11 +268,12 @@ func FlashGDB(pkgName, port string, ocdOutput bool, options *compileopts.Options
 		gdbInterface, openocdInterface := config.Programmer()
 		switch gdbInterface {
 		case "msd", "command", "":
-			if gdbInterface == "" {
-				gdbInterface = "command"
-			}
 			if openocdInterface != "" && config.Target.OpenOCDTarget != "" {
 				gdbInterface = "openocd"
+			}
+			if len(config.Target.Emulator) != 0 {
+				// Assume QEMU as an emulator.
+				gdbInterface = "qemu"
 			}
 		}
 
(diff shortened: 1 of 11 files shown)


deadprogram:
We should probably add the info on how to use this somewhere on the TinyGo web site. In any case, now merging.

deadprogram:
Something went wrong with the Azure CI build after merging this PR, can you please look at that @ayke?

aykevl:
I don't see how this PR could be related to the build failure. I have seen random errors in the cgo tests before, and only on Windows. Perhaps #775 will fix them (but we'll only know after a while with no errors).

PR #764, opened 2019-11-29
riscv: fix heap corruption

aykevl:
The .sdata and .sbss sections are created by the compiler, but were not
present in the linker script. That means that the compiler put them
after all other data/bss section, which happens to be where the heap
also resides.

This commit adds the .sdata and .sbss sections to the linker script,
which gets the blinky examples to work again on RISC-V.


deadprogram:
Tested on real hardware, worked as expected.

Now merging, thanks @aykevl for the fix!

PR #765, opened 2019-11-30
RISC-V tests in QEMU

aykevl:
Not everything is supported right now, but these tests should make sure that things that work now won't break in the future.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -55,6 +55,14 @@ func TestCompiler(t *testing.T) {
 		runPlatTests("cortex-m-qemu", matches, t)
 	})
 
+	if runtime.GOOS == "windows" || runtime.GOOS == "darwin" {
+		// Note: running only on Windows and macOS because Linux usually has an
+		// outdated QEMU version that doesn't support RISC-V yet.
+		t.Run("EmulatedHiFive1", func(t *testing.T) {
+			runPlatTests("hifive1-qemu", matches, t)
+		})
+	}
+
 	if runtime.GOOS == "linux" {
 		t.Run("ARMLinux", func(t *testing.T) {
 			runPlatTests("arm--linux-gnueabihf", matches, t)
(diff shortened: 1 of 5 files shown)


deadprogram:
@aykevl perhaps if you rebase against `dev` this PR might pass the Azure build?

aykevl:
Doubt it, it'll need the SiFive version of QEMU.

aykevl:
Okay, this is ready for review.

I've used a few dirty hacks to get this to work and it doesn't quite work yet on Windows for reasons I can't explain but at least this shows where I'm going.

I think the crude hacks could be avoided by using the `virt` machine for RISC-V instead of `sifive_e`. That target isn't as close to the HiFive1 as `sifive_e`, but it looks like it has a mechanism for exiting tests (the so-called SiFive Test Finisher). Any opinion on this? Rather have dirty hacks or a test that looks less like the real hardware?

Note: the SiFive test finisher is included in the SiFive build of QEMU but somehow didn't end up upstream. Unfortunately, using their version would slow down CI significantly.

main_test.go
```diff
@@ -55,6 +55,12 @@ func TestCompiler(t *testing.T) {
 		runPlatTests("cortex-m-qemu", matches, t)
 	})
 
+	if runtime.GOOS != "linux" {
```
deadprogram: Does this also need to exclude `freebsd`?

aykevl: Heh, good point. I'll simply check for windows or darwin instead, to be conservative.

deadprogram:
What is the state of this branch? It has a merge conflict that needs to be resolved, but otherwise what is needed for it to be merged?

aykevl:
Right now it has a hack to work around the lack of exiting from within QEMU. I think I should add a new `virt` RISC-V target to fix that.

aykevl:
Investigating this it seems like the first commit will be obsoleted by #871. It would be convenient if that PR was merged first.

aykevl:
Replaced in favor of #976.

PR #766, opened 2019-12-01
targets: simulate Circuit Playground Express in play.tinygo.org

aykevl:
This requires a number of changes to atsamd21 code. It's not a very clean separation, but I couldn't come up with a better one.


deadprogram:
Tested on actual CPE board, worked as expected with several test programs, including Gopherbot code. I´d like to merge, but perhaps should wait until whatever issue with the Azure CI from the last merge is fixed?

deadprogram:
Now merging.

PR #184, opened 2019-02-11
main: use internal libclang to build C and assembly files

aykevl:
This avoids a dependency on the clang-7 tool for release builds.

Note: this may need some more testing.

PR #769, opened 2019-12-04
builder: use builtin Clang when building statically

aykevl:
This will be a huge help for people installing TinyGo that don't have LLVM/Clang 9 already installed and in the `$PATH` variable.

Replaces #184.

I added an extra note to the LICENSE part of the README, would that be enough to cover the new code under the different license? I don't think adding it to the LICENSE file would be necessary as the files themselves already say that they are under a different license.

### DIFF
--- a/builder/compiler-builtin.go
+++ b/builder/compiler-builtin.go
@@ -1,0 +1,56 @@
+// +build byollvm
+
+package builder
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	"unsafe"
+
+	"github.com/tinygo-org/tinygo/goenv"
+)
+
+/*
+#cgo CXXFLAGS: -fno-rtti
+#include <stdbool.h>
+#include <stdlib.h>
+bool tinygo_clang_driver(int argc, char **argv);
+*/
+import "C"
+
+// runCCompiler invokes a C compiler with the given arguments.
+//
+// This version invokes the built-in Clang when trying to run the Clang compiler.
+func runCCompiler(command string, flags ...string) error {
+	switch command {
+	case "clang":
+		// Compile this with the internal Clang compiler.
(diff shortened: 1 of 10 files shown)


deadprogram:
I would suggest also adding the additional license info to https://github.com/tinygo-org/tinygo/blob/master/LICENSE

deadprogram:
OMG it compiles!!! 

One question is do we still need `lld` linker install, since this replaces clang-9?

builder/compiler-builtin.go
```diff
@@ -0,0 +1,57 @@
+// +build byollvm
+
+package builder
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	//"path/filepath"
```
deadprogram: Should remove this probably.

deadprogram:
@aykevl please note my question and comment on this PR. Thank you.

builder/compiler-builtin.go
aykevl: Oops, fixed.

aykevl:
> I would suggest also adding the additional license info to https://github.com/tinygo-org/tinygo/blob/master/LICENSE

What do you mean exactly? Copy the entire license text (something I'd rather not do) or just say something like this?

> TinyGo includes portions of LLVM, which is under the Apache License v2.0 with LLVM Exceptions.
> See https://llvm.org/LICENSE.txt for license information.

---

> One question is do we still need `lld` linker install, since this replaces clang-9?

We already bundle `lld` in static/release (`make`) builds for quite some time, so no. For `go install` builds nothing has changed.

deadprogram:
Just adding your propsed excerpt to our license is sufficient to comply with the Apache 2.0 license.

aykevl:
Updated.

deadprogram:
Thanks for the update. Now merging.

aykevl:
Great!

PR #775, opened 2019-12-07
cgo: don't run tests in parallel

aykevl:
Since LLVM 9, CGo sometimes randomly breaks with weird error messages on
Windows. I'm not sure why this is the case, but it might be related to
concurrency.

Disable concurrency for now, and hope that will make the errors go away.

---

This PR is for the issue described here: https://github.com/tinygo-org/tinygo/pull/763#issuecomment-562865172
I'm not sure whether it'll fix it, but I hope so.


deadprogram:
That worked, so now merging. Thanks @aykevl for the fix!

PR #778, opened 2019-12-08
machine: add Tx method to simulated SPI bus

aykevl:
This allows display drivers like st7789 to be used on the TinyGo Playground.


deadprogram:
:1234: merging.

PR #779, opened 2019-12-09
machine: support arduino-nano33 on play.tinygo.org

aykevl:
Similar to #766.


deadprogram:
Tested both UART and SPI interfaces with this PR on real hardware, worked as expected. Merging.

PR #780, opened 2019-12-14
tools: rewrite gen-device-svd in Go

aykevl:
This should make it more maintainable. Another big advantage that generation time (including gofmt) is now 3 times faster. No real attempt at refactoring has been made, that will need to be done at a later time.

While the outputs are sometimes slightly different (mostly in comments), I have verified that both the tinygo smoke tests and the drivers smoke tests have identical outputs (by comparing the outputs from `make`), so this should be a safe change.


deadprogram:
@aykevl see comment please...

aykevl:
I don't see a comment but I guess you mean `gen-device-svd.py` that should be renamed to `gen-device-svd.go`? I've fixed that now with a force push.

I've done another force push to remove the dependency on Python (no Python scripts are present anymore).

deadprogram:
Thanks for removing the python dependencies, I missed that one.

Very happily merging!

PR #781, opened 2019-12-15
Nano33 IoT: default SPI should be D13/D11/D12 instead of A2/A3/A6

conejoninja:
As per the official documentation : https://store.arduino.cc/arduino-nano-33-iot the SCK/MOSI/MISO pinas are D13/D11/D12. In order to avoid confusion, we should change the default SPI pins.


deadprogram:
Works as expected, so merging. Thanks @conejoninja for the PR!

PR #782, opened 2019-12-15
Interrupt API

aykevl:
Very much incomplete, but shows how it could work.

It works locally (after patching go-llvm) and has no code size impact for blinky2 on the pca10040.

### DIFF
--- a/src/machine/board_circuitplay_express_baremetal.go
+++ b/src/machine/board_circuitplay_express_baremetal.go
@@ -2,7 +2,10 @@
 
 package machine
 
-import "device/sam"
+import (
+	"device/sam"
+	"runtime/interrupt"
+)
 
 // UART1 on the Circuit Playground Express.
 var (
@@ -13,9 +16,8 @@ var (
 	}
 )
 
-//go:export SERCOM1_IRQHandler
-func handleUART1() {
-	defaultUART1Handler()
+func init() {
+	UART1.Interrupt = interrupt.New(sam.IRQ_SERCOM4, UART1.handleInterrupt)
 }
 
 // I2C on the Circuit Playground Express.
--- a/src/machine/board_trinket.go
+++ b/src/machine/board_trinket.go
@@ -2,7 +2,10 @@
 
 package machine
 
-import "device/sam"
+import (
+	"device/sam"
+	"runtime/interrupt"
+)
 
 // used to reset into bootloader
 const RESET_MAGIC_VALUE = 0xf01669ef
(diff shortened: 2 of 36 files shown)


aykevl:
I've taken care to produce useful error messages when an interrupt is defined twice. For example:

```
src/machine/machine_nrf.go:91:23: UART0_IRQHandler redeclared in this program
	previous declaration at src/machine/machine_nrf51.go:22
```

src/machine/board_trinket.go
```diff
@@ -51,9 +54,8 @@ var (
 	}
 )
 
-//go:export SERCOM1_IRQHandler
-func handleUART1() {
-	defaultUART1Handler()
+func init() {
+	UART1.interrupt = interrupt.New(sam.IRQ_SERCOM0, UART1.handleInterrupt)
```
aykevl: I found a bug here: UART1 is defined with SERCOM0 but the interrupt fired on SERCOM1. This PR fixes that.

src/machine/board_circuitplay_express_baremetal.go
```diff
@@ -13,9 +16,8 @@ var (
 	}
 )
 
-//go:export SERCOM1_IRQHandler
-func handleUART1() {
-	defaultUART1Handler()
+func init() {
+	UART1.interrupt = interrupt.New(sam.IRQ_SERCOM4, UART1.handleInterrupt)
```
aykevl: Same issue here: the UART uses SERCOM4 but the interrupt fired on SERCOM1.

deadprogram:
I just tried this branch, this was the result:

```
$ tinygo flash -target arduino-nano33 examples/serial
cannot find interrupt name for number 7
cannot find interrupt name for number 3
```

aykevl:
You need to update the generated Go files using `make gen-device` (or specifically `make gen-device-sam`).

deadprogram:
I should have realized that just from looking at the files changed in this PR. That did it.

I tried a few more examples that use interrupts, like UART and USB CDC and worked perfectly.

aykevl:
This PR is now ready from my point of view.

I've done some measurements regarding code/data size changes with this PR:

  * As expected, some programs got slightly larger while others got slightly smaller. This is probably just normal optimization noise that happens due to reordered functions.
  * Programs running on the samd21 chip became slightly smaller (about 80 bytes) and consumed less RAM (144 bytes). I think this is because the `UART1` struct (including buffer) can now be optimized away because most programs don't use it.

deadprogram:
Seems like `src/machine/machine_atsamd51.go` has the newer API for the USB and the older one for the UART? Is that intentional?

deadprogram:
Never mind I just realized that #831 and this PR can be rebased together, and then I suppose a future PR will complete the transition.

aykevl:
> Never mind I just realized that #831 and this PR can be rebased together, and then I suppose a future PR will complete the transition.

Exactly, that was my plan. Doing it in this PR (without #831) would complicate things.

deadprogram:
OK I have tested with SAMD21, SAMD51, NRF52840, and AVR. So far all have worked perfectly.

Now merging, this is giant leap forward for what we can do!

PR #784, opened 2019-12-16
machine: rename CPU_FREQUENCY -> CPUFrequency()

aykevl:
These all-caps constants aren't in the Go style, so rename it to `CPUFrequency` (which is more aligned with Go style). Additionally, make it a function so that it is possible to add support for changing the frequency in the future.

Tested by running `make smoketest`. None of the outputs did change.


deadprogram:
Seems like a good idea. Merging.

PR #785, opened 2019-12-16
flash: search for default serial port on macOS

deadprogram:
This PR searches for a reasonable default port value on macOS based on the typical serial port naming, so that if a single Arduino Nano33 is plugged into a Mac, the developer does not have to search for the correct port and enter it manually.

Before this PR:

```
ls /dev | grep usb
export NANO33_DEV_PATH=/dev/{YOUR TTY USBMODEM ID}
tinygo flash -target arduino-nano33 -port=$NANO33_DEV_PATH examples/blinky1
```

With this PR:

```
tinygo flash -target arduino-nano33 examples/blinky1
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -144,6 +144,14 @@ func Test(pkgName string, options *compileopts.Options) error {
 
 // Flash builds and flashes the built binary to the given serial port.
 func Flash(pkgName, port string, options *compileopts.Options) error {
+	if port == "" {
+		var err error
+		port, err = getDefaultPort()
+		if err != nil {
+			return err
+		}
+	}
+
 	config, err := builder.NewConfig(options)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -480,6 +496,23 @@ func parseSize(s string) (int64, error) {
 	return n, err
 }
 
+func getDefaultPort() (string, error) {
```
aykevl: Can you add a comment describing what this function does? For example:

> getDefaultPort returns the default serial port depending on the OS.

main.go
```diff
@@ -480,6 +496,23 @@ func parseSize(s string) (int64, error) {
...
+	port := "/dev/ttyACM0"
```
aykevl: Not important, but it might be useful to do the same on Linux, globbing for `/dev/ttyACM*`. After removing and quickly reinserting a USB cable, the number might change.
You could do that with a switch statement:

```go
switch runtime.GOOS {
case "linux":
// ...
case "darwin":
// ...
default:
// ...
}
```

Additionally, some serial ports present as /dev/ttyUSB0 etc. but AFAIK we don't support any of those yet.

main.go
deadprogram: Done.

main.go
deadprogram: Good idea. Done.

deadprogram:
Changes made based on suggestions and force pushed to branch.

deadprogram:
Tested on both macOS and Ubuntu.

aykevl:
This is one of those papercuts that will really improve the experience, especially for those just starting out with TinyGo.

PR #786, opened 2019-12-18
flash: use win32 wmi to try to find UF2 and COM ports when flashing

deadprogram:
This PR attempts to do for Windows 10 the same searching for default port and drive name when flashing boards as we are now doing for both Linux and macOS.

EDIT:
Both of these commands work:

```
tinygo flash -target circuitplay-express examples/blinky1
```

and

```
tinygo flash -target arduino-nano33 examples/blinky1
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"errors"
 	"flag"
 	"fmt"
@@ -209,7 +210,18 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 			flashCmd = strings.Replace(flashCmd, "{port}", port, -1)
 
 			// Execute the command.
-			cmd := exec.Command("/bin/sh", "-c", flashCmd)
+			var cmd *exec.Cmd
+			switch runtime.GOOS {
+			case "windows":
+				command := strings.Split(flashCmd, " ")
+				if len(command) < 2 {
+					return errors.New("invalid flash command")
+				}
+				cmd = exec.Command(command[0], command[1:]...)
+			default:
+				cmd = exec.Command("/bin/sh", "-c", flashCmd)
+			}
+
 			cmd.Stdout = os.Stdout
 			cmd.Stderr = os.Stderr
 			cmd.Dir = goenv.Get("TINYGOROOT")
@@ -438,9 +450,18 @@ func touchSerialPortAt1200bps(port string) error {
 
 func flashUF2UsingMSD(volume, tmppath string) error {
 	// find standard UF2 info path
-	infoPath := "/media/*/" + volume + "/INFO_UF2.TXT"
-	if runtime.GOOS == "darwin" {
+	var infoPath string
+	switch runtime.GOOS {
+	case "linux":
+		infoPath = "/media/*/" + volume + "/INFO_UF2.TXT"
+	case "darwin":
 		infoPath = "/Volumes/" + volume + "/INFO_UF2.TXT"
+	case "windows":
+		path, err := windowsFindUSBDrive()
+		if err != nil {
+			return err
+		}
+		infoPath = path + "/INFO_UF2.TXT"
 	}
 
 	d, err := filepath.Glob(infoPath)
@@ -456,9 +477,18 @@ func flashUF2UsingMSD(volume, tmppath string) error {
 
 func flashHexUsingMSD(volume, tmppath string) error {
 	// find expected volume path
-	destPath := "/media/*/" + volume
-	if runtime.GOOS == "darwin" {
+	var destPath string
+	switch runtime.GOOS {
+	case "linux":
+		destPath = "/media/*/" + volume
+	case "darwin":
 		destPath = "/Volumes/" + volume
+	case "windows":
+		path, err := windowsFindUSBDrive()
+		if err != nil {
+			return err
+		}
+		destPath = path + "/"
 	}
 
 	d, err := filepath.Glob(destPath)
@@ -470,6 +500,29 @@ func flashHexUsingMSD(volume, tmppath string) error {
 	}
 
 	return moveFile(tmppath, d[0]+"/flash.hex")
+}
+
+func windowsFindUSBDrive() (string, error) {
+	cmd := exec.Command("wmic", "PATH", "Win32_LogicalDisk",
+		"get", "DeviceID,", "VolumeName,",
+		"FileSystem,", "DriveType")
+
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	err := cmd.Run()
+	if err != nil {
+		return "", err
+	}
+
+	for _, line := range strings.Split(out.String(), "\n") {
+		words := strings.Fields(line)
+		if len(words) >= 3 {
+			if words[1] == "2" && words[2] == "FAT" {
+				return words[0], nil
+			}
+		}
+	}
+	return "", errors.New("unable to locate a USB device to be flashed")
 }
 
 // parseSize converts a human-readable size (with k/m/g suffix) into a plain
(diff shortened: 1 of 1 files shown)


deadprogram:
Tested on Windows 10 using Surface Pro. Was able to `tinygo flash` both UF2 boards as well as an Arduino Nano33 using `bossac`.

One important note, is that you do have to double tap the reset button to put any of these boards into bootloader mode. That could be an additional refinement for a future PR. This one is done and ready for review.

main.go
```diff
@@ -209,7 +210,18 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 			flashCmd = strings.Replace(flashCmd, "{port}", port, -1)
 
 			// Execute the command.
-			cmd := exec.Command("/bin/sh", "-c", flashCmd)
+			var cmd *exec.Cmd
+			switch runtime.GOOS {
+			case "windows":
+				command := strings.Split(flashCmd, " ")
```
aykevl: It's better to split commands using a dedicated command splitter, such as github.com/google/shlex, otherwise spaces may give invalid results.

main.go
```diff
@@ -472,6 +502,29 @@ func flashHexUsingMSD(volume, tmppath string) error {
 	return moveFile(tmppath, d[0]+"/flash.hex")
 }
 
+func windowsFindUSBDrive() (string, error) {
+	cmd := exec.Command("wmic", "PATH", "Win32_LogicalDisk",
+		"get", "DeviceID,", "VolumeName,",
+		"FileSystem,", "DriveType")
```
aykevl: A bit hacky but can't easily find a better way so OK.

main.go
```diff
@@ -505,6 +558,30 @@ func getDefaultPort() (port string, err error) {
 		portPath = "/dev/cu.usb*"
 	case "linux":
 		portPath = "/dev/ttyACM*"
+	case "windows":
+		cmd := exec.Command("wmic", "PATH", "Win32_SerialPort",
+			"get", "DeviceID")
```
aykevl: Did you see this?
http://a46554.blogspot.com/2018/05/go-lang-list-all-com-port-in-windows.html?m=1
Found it with a quick Google search. Not sure whether it works, but it might avoid an external command.

main.go
```diff
@@ -438,9 +450,18 @@ func touchSerialPortAt1200bps(port string) error {
...
+	case "linux":
+		infoPath = "/media/*/" + volume + "/INFO_UF2.TXT"
+	case "darwin":
 		infoPath = "/Volumes/" + volume + "/INFO_UF2.TXT"
+	case "windows":
+		path, err := windowsFindUSBDrive()
+		if err != nil {
+			return err
+		}
+		infoPath = path + "/INFO_UF2.TXT"
```
aykevl: Looks like this only checks the first FAT device available (that may or may not be the right one).
That can possibly be fixed later, but could be surprising to new users if no device can be found while it's clearly plugged in (along a USB thumb drive).

main.go
deadprogram: Was not aware of that package, just switched to using it.

main.go
deadprogram: We are already using the `wmic` command which is part of Windows itself. The command seems like the proper way to do this vs. search for some Windows registry entries that might change or even be entirely different based on system config.

main.go
deadprogram: Added some additional logic to ensure that we only use attached volume with the correct name that matches the board, like we are doing on Linux/macOS.

deadprogram:
Pushed an additional commit modifying this PR to take into account some of @aykevl suggestions.

main.go
deadprogram: Hmm shlex does not appear to work as I expected, and now my `wmic` calls are failing. Have to look deeper into how it works, but my previous solution was working well enough.

Basically just separating the command from the params is probably enough, since what was happening was the runtime on Windows does some extra insertions of `\` characters.

Anyhow, I think I will keep the other changes, but go back to my simpler parsing that works for the Windows rules.

deadprogram:
Updated the previous commit to go back to the simpler parsing for Windows, but with enhancements for ensuring the correct devices are used.

deadprogram:
OK it took a bit to get the correct magic syntax to pass thru the code only called on Windows:
https://github.com/golang/go/blob/master/src/syscall/exec_windows.go#L17

Now it works exactly as expected, tested both gain and they worked perfectly.

Ready for merge!

go.sum
```diff
@@ -29,4 +29,4 @@ tinygo.org/x/go-llvm v0.0.0-20191103200204-37e93e3f04e2/go.mod h1:fv1F0BSNpxMfCL
 tinygo.org/x/go-llvm v0.0.0-20191113125529-bad6d01809e8 h1:9Bfvso+tTVQg16UzOA614NaYA4x8vsRBNtd3eBrXwp0=
 tinygo.org/x/go-llvm v0.0.0-20191113125529-bad6d01809e8/go.mod h1:fv1F0BSNpxMfCL0zF3M4OPFbgYHnhtB6ST0HvUtu/LE=
 tinygo.org/x/go-llvm v0.0.0-20191124211856-b2db3df3f257 h1:o8VDylrMN7gWemBMu8rEyuogKPhcLTdx5KrUAp9macc=
-tinygo.org/x/go-llvm v0.0.0-20191124211856-b2db3df3f257/go.mod h1:fv1F0BSNpxMfCL0zF3M4OPFbgYHnhtB6ST0HvUtu/LE=
+tinygo.org/x/go-llvm v0.0.0-20191124211856-b2db3df3f257/go.mod h1:fv1F0BSNpxMfCL0zF3M4OPFbgYHnhtB6ST0HvUtu/LE=
```
aykevl: what's happening here? :thinking:

main.go
aykevl: Just a little bit worried that future changes may trip something up, but I guess it's fine for now.

main.go
```diff
@@ -505,6 +558,30 @@ func getDefaultPort() (port string, err error) {
 		portPath = "/dev/cu.usb*"
 	case "linux":
 		portPath = "/dev/ttyACM*"
+	case "windows":
+		cmd := exec.Command("wmic",
+			"PATH", "Win32_SerialPort", "WHERE", "Caption LIKE 'USB Serial%'", "GET", "DeviceID")
```
aykevl: Ugh that looks a lot like SQL, but whatever...

main.go
aykevl: I don't see the additional logic, but I do see that I missed the extra `volume` parameter to `windowsFindUSBDrive` which makes sure that only a volume with the correct name is returned. So that should be good.

main.go
```diff
@@ -472,6 +502,29 @@ func flashHexUsingMSD(volume, tmppath string) error {
 	return moveFile(tmppath, d[0]+"/flash.hex")
 }
 
+func windowsFindUSBDrive(volume string) (string, error) {
+	cmd := exec.Command("wmic",
+		"PATH", "Win32_LogicalDisk", "WHERE", "VolumeName = '"+volume+"'",
```
aykevl: There is no sanity-checking of the volume name. That's not a problem right now, but may be a problem if we want to accept untrusted volume names in the future.

main.go
deadprogram: That is correct: https://docs.microsoft.com/en-us/windows/win32/wmisdk/wmi-and-sql

main.go
deadprogram: We can add that in a future PR.

deadprogram:
Went back and rebased to remove the change to `go.sum` which was just adding a new line at the end? Not sure where that came from, but I removed and rebased it away.

This is now ready to go, within the parameters of what it is trying to do.

PR #787, opened 2019-12-19
interp: fix maps with string values

aykevl:


aykevl:
The first commit gets image/jpeg to work. The second commit breaks the syscall package...

aykevl:
Hopefully it'll work now...

deadprogram:
We for sure needed this. Merging, thanks @aykevl

PR #789, opened 2019-12-20
interp: remove accidental debug print

aykevl:
Accidentally left in the source in https://github.com/tinygo-org/tinygo/pull/787.


deadprogram:
Oops, missed that one. Merging.

PR #790, opened 2019-12-20
interp: add runtime fallback for mapassign operations

aykevl:
Some mapassign operations cannot (yet) be done by the interp package. Implement a fallback mechanism so that these operations can still be performed at runtime. In the future, this might be optimized further where possible.

---

This makes a few more packages importable: go/ast, go/format, go/parser, go/printer, go/scanner, and go/token (but probably all because the common dependency go/token became importable).

Eventually I'd like to add support for running gofmt in the playground (without network request!) but go/printer doesn't yet compile due to missing support for interfaces in map keys.


deadprogram:
:world_map: merging.

PR #641, opened 2019-10-19
Give readable names to subtests and run in parallel

jaddr2line:
This patch restructures the tests a bit so that the names of the subtests indicate the platform that they were run on. Additionally, the tests are now run in parallel. On my machine, this reduced the time needed to run the tests from 81 seconds to 28 seconds.

PR #792, opened 2019-12-20
Run tests partially in parallel

jaddr2line:
This change allows all parts of the testing process except for the actual build to be run in parallel. This drastically reduces the time required to run compiler tests. This is a less ambitious version of #641.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"sort"
+	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


deadprogram:
Good one @jaddr2line thanks for another useful contribution.

Before:
`ok  	github.com/tinygo-org/tinygo	95.085s`

After:
`ok  	github.com/tinygo-org/tinygo	31.702s`

Passes on all platforms, so now merging. Thank you!

PR #793, opened 2019-12-21
builder: protect against unsupported concurrency in lld

aykevl:
LLD does not support concurrency, and cannot be run at the same time as compiles. Therefore, use a `RWMutex` to read-lock compiles and write-lock links.

This commit moves the locking from the full program tests to the builder package itself, to provide a cleaner API boundary: users of `builder.Build` do not have to think about concurrency. This move also makes the above mentioned compile concurrency possible.

On my system, this results in about 36% faster tests.

---

Note: I've been busy with allowing packages to be compiled independently. When that is ready, `lldLock` will probably be locked per package.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -20,7 +20,9 @@ import (
 
 // Build performs a single package to executable Go build. It takes in a package
 // name, an output path, and set of compile options and from that it manages the
-// whole compilation process.
+// whole compilation process. It is safe to be called concurrently. However,
+// because of limitations in LLVM you may not call into LLVM functions while
+// Build is running.
 //
 // The error value may be of type *MultiError. Callers will likely want to check
 // for this case and print such errors individually.
--- a/builder/compiler-builtin.go
+++ b/builder/compiler-builtin.go
@@ -6,6 +6,8 @@ import (
 	"errors"
 	"os"
 	"os/exec"
+	"runtime"
+	"sync"
 	"unsafe"
 
 	"github.com/tinygo-org/tinygo/goenv"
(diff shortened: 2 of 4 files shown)


aykevl:
Hopefully this will fix the Windows issue.

deadprogram:
It appears to have fixed Windows. However it broke macOS.

aykevl:
It was a "too many open files" problem, which seems spurious to me. Fixed it by retrying the build.

However, I did find a file that wasn't closed here: https://github.com/tinygo-org/tinygo/pull/799 (that should reduce the chances of this happening).

builder/build.go
```diff
@@ -11,16 +11,22 @@ import (
...
 
+// Due to some problems with embedded Clang and LLD, we cannot run those in
+// parallel, or in parallel with Go compiles. Therefore, we put a lock around
+// builds and run everything else in parallel.
+var buildLock sync.RWMutex
+
 // Build performs a single package to executable Go build. It takes in a package
 // name, an output path, and set of compile options and from that it manages the
-// whole compilation process.
+// whole compilation process. It is safe to be called concurrently.
```
jaddr2line: We may need to clarify that the caller cannot use LLVM elsewhere while this is running.

aykevl:
I did another update. I don't see why LLD and Clang would interfere with each other or with regular LLVM building/linking so I switched to more fine grained locks instead (for just Clang and LLD, and Clang only on Windows).

builder/build.go
aykevl: Fixed.

deadprogram:
Seems to still be broken on macOS.

aykevl:
Ugh, builder/ar.go is really bad at properly closing files when there is an error. That may have triggered it.

aykevl:
We invoke lld in a separate process now so this is no longer an issue.

PR #794, opened 2019-12-21
 compiler: improve error locations in goroutine lowering 

aykevl:
Depends on #788. Should improve "async function foo used as function pointer" like errors.


deadprogram:
Merging.

PR #795, opened 2019-12-21
Added Adafruit's pybadge target

conejoninja:
It's my first board contribution, and I appreciate feedback. It's a variation of feather-m4, and compatible with it, most pins are the same, only doubt comes with the serial.

Speaker works, neopixels work, TFT works, I²C works, serial (println) works too, I tried a few random pins and they work too. Only thing not tested are the buttons, it uses a 8-channel shift register, a plugin should be used.


targets/pybadge.json
```diff
@@ -0,0 +1,8 @@
+{
+    "inherits": ["atsamd51j19a"],
+    "build-tags": ["sam", "atsamd51j19a", "pybadge"],
```
aykevl: `"sam", "atsamd51j19a"` are unnecessary, they are already part of the inherited target.

conejoninja: If I remove them, I get : 

```
error: no buildable Go source files in /workspace/go/src/github.com/tinygo-org/tinygo/src/device/sam
```

And I saw them in every other target, for example https://github.com/tinygo-org/tinygo/blob/master/targets/feather-m4.json#L3

aykevl: Ah, interesting. Seems to be a bug on this line:
https://github.com/tinygo-org/tinygo/blob/a44d74d4c05b771d1c013330949a5689855492aa/targets/atsamd51j19a.json#L4

It should have `atsamd51j19a` added to the build tags.

In that case, that can be fixed in a separate PR.

deadprogram:
From the looks of it, most of the buttons on the PyBadge are just GPIO pins. See https://github.com/adafruit/Adafruit_Arcada/blob/5b1ff3043c65ba4670ef33a8d110d908b5cfc8df/Adafruit_Arcada.cpp#L473-L493 for example.

conejoninja:
Buttons are behind a shift register : https://cdn-learn.adafruit.com/assets/assets/000/075/064/original/adafruit_products_PyBadge_Sch.png?1556577492  

BUTTON_LATCH, BUTTON_CLK and BUTTON_OUT are defined, but I think reading them should be done elsewhere.

On-board QSPI flash memory was not tested either.

targets/pybadge.json
conejoninja: Adding "atsamd54j19a" to atsamd51j19a.json works, but I think other targets have missing tags too, for example: https://github.com/tinygo-org/tinygo/blob/master/targets/atsamd21g18a.json

After that, I guess some tags could be removed from others targets too: https://github.com/tinygo-org/tinygo/blob/master/targets/feather-m4.json

deadprogram:
Thanks for the schematic. I will look in my parts boxes, but I do not think I have that particular shift register. However, as you said, that would be a future addition to the drivers repo anyhow.

conejoninja:
Added bit masks for the buttons, so it works with the _generic_ shift register driver https://github.com/tinygo-org/drivers/pull/107 (works great with pybadge, I need to check others) it will not pass CI until this PR is merged, which in turn needs* the build-tags fixed

deadprogram:
Don't forget to add to https://github.com/tinygo-org/tinygo/tree/dev#supported-boardstargets as well.

deadprogram:
@conejoninja if you add to the page with list of supported targets, then we can merge, no?

conejoninja:
Added :) thanks for your patience

deadprogram:
Thank you for adding this @conejoninja another fun member of the TinyGo family!

Squash/merging. Thanks again.

PR #796, opened 2019-12-21
tools: use byte padding to skip unused register ranges

aykevl:
This simplifies the code. The fields are blank anyway so there is no way to access them anyway (volatile or not).
Also do some other related simplifications of the code that result from this change.

I tested this by comparing all smoketests (from the compiler and from all drivers) to check whether the outputs are the same. That way, I found and fixed some bugs. After this, the only test case that has a different output is (again!) examples/apa102 from the drivers repo - and again in reflection data, not in any suspicious code.

### DIFF
--- a/tools/gen-device-svd/gen-device-svd.go
+++ b/tools/gen-device-svd/gen-device-svd.go
@@ -668,8 +668,7 @@ var (
 			}
 
 			var regType string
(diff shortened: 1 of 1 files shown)


deadprogram:
I really like this simplification. Merging, good one! :scissors:

PR #797, opened 2019-12-21
transform: make reflection sidetables constant globals

aykevl:
These globals are (and must be!) never modified by the reflect package. By marking them as constant, they will be put in read-only memory. This reduces RAM consumption on microcontrollers.

Discovered while working on #796.


deadprogram:
Merging, thanks!

PR #799, opened 2019-12-21
main: avoid leaving files open

aykevl:
Eventually, open files should be closed when the GC runs and the finalizer is called. However we shouldn't rely on that.

Using `ioutil.ReadFile` instead as it's simpler anyway.


deadprogram:
Also less code, which is good. Merging.

PR #800, opened 2019-12-22
Adafruit PyPortal Support

bgould:
https://www.adafruit.com/product/4116

LED: working
USBCDC: working
UART: not working (however only needed for reprogramming or getting debug output from ESP32)
SPI0: working
I2C: working
I2S: not added/tested yet

Temp sensor: working via tinygo.org/x/drivers/adt7410, pending https://github.com/tinygo-org/drivers/pull/109
Light sensor: working via `machine.ADC{machine.LIGHT} `
Resistive touch: working via github.com/bgould/tinygo-touchscreen
TFT: working via github.com/bgould/drivers/ili9341 (pending https://github.com/tinygo-org/drivers/pull/115)
TFT Backlight: PWM working
Neopixel: working via tinygo.org/x/drivers/ws2812
Wifi: working via tinygo.org/x/drivers/wifinina
QSPI flash: not tested; driver started but needs to get further along to be able to test

~~Also, auto-reset when flashing is not working properly.  Double-tap of the reset button is necessary to flash.  I think this is probably possible to get working however.~~ Fixed: this was due to different amount of RAM on ATSAMD51J20 than on G19/J19.  Added a constant for for hardware SRAM size and all is good.

### DIFF
--- a/src/machine/board_pyportal.go
+++ b/src/machine/board_pyportal.go
@@ -63,14 +63,14 @@ const (
 
 // Analog pins
 const (
-	A0 = PA02 // ADC/AIN[0]
-	A1 = D3   // ADC/AIN[2]
-	A2 = PA07 // ADC/AIN[3]
-	A3 = D4   // ADC/AIN[4]
-	A4 = PB00 // ADC/AIN[5]
-	A5 = PB01 // ADC/AIN[10]
-	A6 = PA06 // ADC/AIN[5]
-	A7 = PB08 // ADC/AIN[5]
+	A0 = PA02 // ADC0/AIN[0]
+	A1 = D3   // ADC0/AIN[4]
+	A2 = PA07 // ADC0/AIN[7]
+	A3 = D4   // ADC0/AIN[5]
+	A4 = PB00 // ADC0/AIN[12]
+	A5 = PB01 // ADC0/AIN[13]
+	A6 = PA06 // ADC0/AIN[6]
+	A7 = PB08 // ADC1/AIN[0]
 
 	AUDIO_OUT = A0
 	LIGHT     = A2
--- a/src/machine/board_pyportal.go
+++ b/src/machine/board_pyportal.go
@@ -21,12 +21,18 @@ const (
 	D7  = PB17 // NINA_RESETN
 	D8  = PB14 // NINA_CS
 	D9  = PB04 // TFT_RD
-	D10 = PB05 // TFT_RS
+	D10 = PB05 // TFT_DC
 	D11 = PB06 // TFT_CS
 	D12 = PB07 // TFT_TE
 	D13 = PB23 // built-in LED
 	D24 = PA00 // TFT_RESET
-	D25 = PB31 // TFT_WR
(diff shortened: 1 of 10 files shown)


src/machine/board_pyportal.go
```diff
@@ -0,0 +1,143 @@
...
+	UART_RX_PIN = D7
+)
+
+/*
+// UART2 pins
+const (
+	UART2_TX_PIN = A4
+	UART2_RX_PIN = A5
+)
+*/
```
aykevl: Why are these left commented out?

bgould: Right now the UART definitions in machine_atsamd51.go are pinned to specific SERCOM's: https://github.com/tinygo-org/tinygo/blob/master/src/machine/machine_atsamd51.go#L470

It looks like machine_atsamd21.go has a more flexible way of handling this, not sure if something similar could also apply to atsamd51... in any case I thought it might be best to comment this out temporarily while I try to get the other peripherals working.

aykevl: Yes I've been working on making atsamd51 pin configuration simpler (auto-detecting pads and pin modes from the given SERCOM and pin numbers). It'll take a while so maybe it's best to leave UART2 out until that is done.
@deadprogram any opinion on that? I generally prefer no code over commented out potentially useful code.

deadprogram: If the code that is commented was the result of a lot of research, and is going to be implemented next, then can leave in comments. Otherwise, agreed.

bgould: Sounds good to me I'll go ahead and remove any commented out code in my next commit and if necessary replace with more meaningful and specific comments as a reminder.

bgould:
The last major piece of the puzzle before this is no longer WIP is to put together a driver for the ILI9431 chip that drives the display so that the pin configuration can be verified.

Also there may be a bug in atsamd51's ADC implementation that I am investigating, where only the first pin used on either ADC0 or ADC1 is usable for subsequent reads via machine.ADC.Get(), no matter which other pins are configured.  For this reason I switched PB08 to use ADC1 which seems to temporarily solve the problem (at least for the touchscreen).  I will need to confirm the source of this issue and mitigate it in order for this board to be fully usable.

bgould:
Driver for ILI9341 is in a workable state now, so I will open a PR on the drivers repo for that, and also do some more testing of the ADC to see if I can figure out what is going on.  I'll then get everything cleaned up in this PR for review.

deadprogram:
I need to obtain one of these myself, after I saw your cool video on Adafruit @bgould

bgould:
Removing WIP from title; UART is not fully working yet, but is not really needed except for debug output from ESP chip, which most users do not need.  I plan to address UART in a separate pull request after the interrupt API is merged.  This PR should be ready for review.

deadprogram:
This is pretty incredbly great work @bgould I am excited to get one and start playing with it.

Now squashing/merging. Thank you so very much for this awesome contribution!

PR #801, opened 2019-12-22
added missing pinetime devkit to list of suppoerted boards/targets

conejoninja:
I noticed the PineTime DevKit is missing from the list of supported boards/targets. O know it's a devkit and not the final product yet, but I still think it should be included.


deadprogram:
Thanks for catching that omission @conejoninja now merging.

PR #597, opened 2019-09-26
machine/atsamd21: refactor SPI pin handling to only look at pin numbers

aykevl:
Pin mode and pad numbers are automatically calculated from the pin numbers, returning an error if no pinout could be found.

Not yet properly tested. However, while doing some initial testing, I found that there was an inconsistency in the Nano 33 IoT pinout for SPI0: the MISO pin is `A6` or `PA09` which has SERCOM0/PAD1 while the `DIpad` was set to `sercomRXPad0` - thus SERCOM0/PAD0 (pin PA08). I don't know which of the two it is supposed to be and whether I didn't miss anything but I'm surprised this worked. @deadprogram can you take a look?

Also, are there espat examples for the Nano 33 IoT (or other SPI tests) that should work with little or no modification?

This is just the first step, my goal is to convert UART and I2C in the same way. And also see whether this is possible on the SAMD51.

PR #802, opened 2019-12-23
machine/atsamd51: refactor SPI pin configuration to only look at pin numbers

aykevl:
This commit does the same thing as https://github.com/tinygo-org/tinygo/pull/597 but for samd51 series chips. Pin mode and pad numbers are automatically calculated from pin numbers, returning an error if no valid pinout is possible.

Code that uses SPI1 and code that does not configure all pins (for example, leaving out MISO) may need to be updated. Therefore, I think it's best to merge this after the coming release to have a bigger chance of finding bugs. Also, it should be updated after #800 has been merged.

### DIFF
--- a/src/machine/board_feather-m4.go
+++ b/src/machine/board_feather-m4.go
@@ -82,14 +82,8 @@ const (
 
 // SPI on the Feather M4.
 var (
-	SPI0 = SPI{Bus: sam.SERCOM1_SPIM,
-		SCK:         SPI0_SCK_PIN,
-		MOSI:        SPI0_MOSI_PIN,
-		MISO:        SPI0_MISO_PIN,
-		DOpad:       spiTXPad3SCK1,
-		DIpad:       sercomRXPad2,
-		MISOPinMode: PinSERCOM,
-		MOSIPinMode: PinSERCOM,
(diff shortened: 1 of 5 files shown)


deadprogram:
This should make it a lot easier to work with adding new SAMD51 boards and also give us more flexibility in configuration.

Now merging.

deadprogram:
@bgould if you have a chance, please check to make sure that SPI is still working as you expect in the `dev` branch on your various SAMD51 boards.

Thanks!

deadprogram:
I just created https://github.com/tinygo-org/tinygo/pull/841 because the PyPortal did need the manual pin mapping for SPI0 removed.

I merged #802 a little too quickly before running the smoke tests. My bad. Anyhow, the #841 should correct that.

aykevl:
Whoops, I wanted to update this PR but didn't get to it yet. Will send a fix soon (to fix the dev branch).

bgould:
> @bgould if you have a chance, please check to make sure that SPI is still working as you expect in the `dev` branch on your various SAMD51 boards.
> 
> Thanks!

FWIW I tested this out on my boards and everything seems good to me

PR #804, opened 2019-12-23
ci: remove symlink creation

aykevl:
These symlinks are now unnecessary because Clang (and previously lld) have been integrated into TinyGo.


deadprogram:
:scissors: merging.

PR #806, opened 2019-12-23
riscv: improve startup assembly

aykevl:
Now that we've switched to LLVM 9, we don't need a workaround anymore for the 'la' pseudo-instruction.
The assembly is slightly different but should be functionally identical.

Tested locally with the PR in #765.


deadprogram:
:+1: merging.

PR #807, opened 2019-12-23
ci: use QEMU from SiFive

aykevl:
WIP, made a PR for testing on Azure Pipelines.


aykevl:
Unfortunately this is a failed attempt: the SiFive version of QEMU doesn't include `qemu-system-arm` etc.

PR #810, opened 2019-12-24
Add initial FreeBSD support

dmgk:
This PR adds initial support for running TinyGo on FreeBSD using self-built LLVM. Both `test` and `smoketest` pass and I'm working on adding it to the ports tree.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,11 @@ $(LIBCLANG_PATH): $(LIBCLANG_FILES)
 else ifeq ($(shell uname -s),Darwin)
     MD5SUM = md5
     LIBCLANG_PATH = $(abspath $(LLVM_BUILDDIR))/lib/libclang.a
+else ifeq ($(shell uname -s),FreeBSD)
+    MD5SUM = md5
+    LIBCLANG_PATH = $(abspath $(LLVM_BUILDDIR))/lib/libclang.a
+    START_GROUP = -Wl,--start-group
+    END_GROUP = -Wl,--end-group
 else
     LIBCLANG_PATH = $(abspath $(LLVM_BUILDDIR))/lib/libclang.a
     START_GROUP = -Wl,--start-group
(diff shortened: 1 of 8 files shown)


aykevl:
Looks like this worked by accident. I expected `tinygo run ./testdata/stdlib.go` (part of `make test`) to have failed because of missing syscall intrinsics, but looking at the code it doesn't seem to check for GOOS when GOARCH=amd64 (but does this GOOS check for other architectures):

https://github.com/tinygo-org/tinygo/blob/923c2e7adafe419e9fa98662244446e1d05c4a75/compiler/syscall.go#L19

In any case, I don't have a FreeBSD system ready to test (and my quick attempt to run it in a VM failed) but the changes themselves look good so from my POV it can be merged.

aykevl:
Also: we don't have CI set up at the moment so I'm afraid it may accidentally break in the future. Unfortunately CircleCI doesn't support FreeBSD and I'd rather not add yet another CI system. @dmgk maybe you have any suggestions?

dmgk:
@aykevl `tinygo run ./testdata/stdlib.go` currently is expected to fail as I didn't look into supporting FreeBSD as a target yet. This PR is about making tinygo building on FreeBSD and being able to use it to target embedded platforms. Given that, I'm not sure why `make test` pass.

Although, adding freebsd/amd64 as a target should be fairly easy because AFAIK both Linux and FreeBSD use the same syscall conventions on amd64. Indeed, adding "freebsd" to a case expression in

[https://github.com/tinygo-org/tinygo/blob/923c2e7adafe419e9fa98662244446e1d05c4a75/compiler/syscall.go#L155](https://github.com/tinygo-org/tinygo/blob/923c2e7adafe419e9fa98662244446e1d05c4a75/compiler/syscall.go#L155)

makes testcase you posted above to pass (on FreeBSD 12.1-STABLE amd64). I'll update the PR shortly.

Regarding CI, the only CI that I know that supports FreeBSD is Cirrus CI which is probably not an option if you already set on CircleCI.

deadprogram:
I am also lacking any way to really test this PR, but looking over the changes they seem sane.

I changed the branch for this PR to `dev` as mentioned here: https://github.com/tinygo-org/tinygo/blob/master/CONTRIBUTING.md#how-to-use-our-github-repository

deadprogram:
It does not negatively impact any other Linux (at least as tested with Ubuntu), so merging as to facilitate more work be able to be done for FreeBSD.

Thank you @dmgk for your contribution.

PR #811, opened 2019-12-25
riscv: support sleeping in QEMU

aykevl:
QEMU doesn't support the RTC peripheral yet so work around it for now.

This makes the following command work:

    tinygo run -target=hifive1-qemu ./testdata/coroutines.go


deadprogram:
I suppose it is a hack, but appears to be the hack we need at the moment.

Merging.

aykevl:
It's a feature, not a hack! :wink:

PR #813, opened 2019-12-26
main: kill tests if they run too long

aykevl:
Sometimes, tests suddenly hang somewhere (in particular in emulators
where crashes often lead to hangs). Setting a limit has two advantages:

  1. Quickly killing test processes that are frozen (as opposed to waiting for the default 10min go test timeout).
  2. The output becomes visible, hopefully giving a clue what went wrong.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -14,6 +14,7 @@ import (
 	"sort"
 	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


deadprogram:
Good call, now merging.

PR #814, opened 2019-12-26
compiler: don't crash when encountering types.Invalid

aykevl:
This commit fixes a crash when trying to compile the following (invalid) code:

    package main

    import "unsafe"

    func main() {
    }

    type Foo struct {
       x DoesNotExist
    }

    const foo = unsafe.Sizeof(Foo{})

This commit fixes this situation. The result is a regular error message, indicating that `DoesNotExist` is not defined.


deadprogram:
A good check to have! Merging.

PR #815, opened 2019-12-26
main: remove getting a serial port in gdb subcommand

aykevl:
Remove this code for two reasons:

 1. It is not needed.
 2. It breaks `tinygo gdb` for debugging QEMU targets (such as cortex-m-qemu).


deadprogram:
:scissors: merging.

PR #819, opened 2019-12-30
targets: add target circuitplay-bluefruit

matloob:
Add a target for the Adafruit Circuit Playground Bluefruit, which is
based on the nRF52840. Adds the necessary code for the machine
package and the json and linker script files in the targets directory.
The machine package code is based on board_circuitplay_express.go,
with modifications made by consulting the wiring diagram on the
adafruit website here:
https://learn.adafruit.com/adafruit-circuit-playground-bluefruit/downloads

Also adds support to the uf2 conversion packacge to set the familyID
field. The Circuit Playground Bluefruit firmware rejects uf2 files
without the family id set to 0xADA52840 (and without the flag specifying
that the family id is present).

### DIFF
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -43,6 +43,7 @@ type TargetSpec struct {
 	FlashMethod      string   `json:"flash-method"`
 	FlashVolume      string   `json:"msd-volume-name"`
 	FlashFilename    string   `json:"msd-firmware-name"`
+	UF2FamilyID      string   `json:"uf2-family-id"`
 	OpenOCDInterface string   `json:"openocd-interface"`
 	OpenOCDTarget    string   `json:"openocd-target"`
 	OpenOCDTransport string   `json:"openocd-transport"`
@@ -108,6 +109,9 @@ func (spec *TargetSpec) copyProperties(spec2 *TargetSpec) {
 	}
 	if spec2.FlashFilename != "" {
 		spec.FlashFilename = spec2.FlashFilename
+	}
+	if spec2.UF2FamilyID != "" {
+		spec.UF2FamilyID = spec2.UF2FamilyID
 	}
 	if spec2.OpenOCDInterface != "" {
 		spec.OpenOCDInterface = spec2.OpenOCDInterface
--- a/src/machine/board_circuitplay_bluefruit.go
+++ b/src/machine/board_circuitplay_bluefruit.go
@@ -1,0 +1,73 @@
+// +build circuitplay_bluefruit
+
+package machine
+
+const HasLowFrequencyCrystal = true
+
+// GPIO Pins
+const (
+	D0  = P0_30
+	D1  = P0_14
+	D2  = P0_05
+	D3  = P0_04
+	D4  = P1_02
+	D5  = P1_15
+	D6  = P0_02
+	D7  = P1_06
+	D8  = P0_13
+	D9  = P0_29
+	D10 = P0_03
+	D11 = P1_04
+	D12 = P0_26
+	D13 = P1_14
+)
+
+// Analog Pins
+const (
+	A1 = P0_02
+	A2 = P0_29
+	A3 = P0_03
+	A4 = P0_04
+	A5 = P0_05
+	A6 = P0_30
+	A7 = P0_14
+	A8 = P0_28
+	A9 = P0_31
+)
+
+const (
+	LED       = D13
+	NEOPIXELS = D8
+
+	BUTTONA = D4
+	BUTTONB = D5
+	SLIDER  = D7 // built-in slide switch
+
+	BUTTON  = BUTTONA
+	BUTTON1 = BUTTONB
+
+	LIGHTSENSOR = A8
+	TEMPSENSOR  = A9
+)
+
+// UART0 pins (logical UART1)
+const (
+	UART_TX_PIN = P0_14 // PORTB
+	UART_RX_PIN = P0_30 // PORTB
+)
+
+// I2C pins
+const (
+	SDA_PIN = P0_05 // I2C0 external
+	SCL_PIN = P0_04 // I2C0 external
+
+	SDA1_PIN = P0_00 // I2C1 internal
+	SCL1_PIN = P0_01 // I2C1 internal
+)
+
+// SPI pins (internal flash)
+const (
+	SPI0_SCK_PIN  = P0_19 // SCK
+	SPI0_MOSI_PIN = P0_21 // MOSI
+	SPI0_MISO_PIN = P0_23 // MISO
+)
--- a/targets/circuitplay-bluefruit.json
+++ b/targets/circuitplay-bluefruit.json
@@ -1,0 +1,9 @@
+{
+    "inherits": ["nrf52840"],
+    "build-tags": ["circuitplay_bluefruit"],
+    "flash-method": "msd",
+    "msd-volume-name": "CPLAYBTBOOT",
+    "msd-firmware-name": "firmware.uf2",
+    "uf2-family-id": "0xADA52840",
+    "linkerscript": "targets/circuitplay-bluefruit.ld"
+}
--- a/targets/circuitplay-bluefruit.ld
+++ b/targets/circuitplay-bluefruit.ld
@@ -1,0 +1,10 @@
+
+MEMORY
+{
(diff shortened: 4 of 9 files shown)


deadprogram:
Hi @matloob thank you very much for the contribution. I do not have a circuitplay-bluefruit yet, but I tested with my Circuit Playground Express and it did not affect it, as expected. So if it works on your board, works for me.

Could you please also add the new board to the smoketests here: https://github.com/tinygo-org/tinygo/blob/master/Makefile#L149-L253

Lastly, please add the board to the list of supported boards here: https://github.com/tinygo-org/tinygo/blob/master/README.md#supported-boardstargets

Thanks!

src/machine/board_circuitplay_bluefruit.go
```diff
@@ -0,0 +1,125 @@
+// +build circuitplay_bluefruit
+
+package machine
+
+const HasLowFrequencyCrystal = true
+
+// Hardware pins
+const (
+	P0_00 Pin = 0
```
aykevl: These pin names are not board specific, so I think they should be moved into a generic nrf52 file (src/machine/board_nrf52832.go).

targets/circuitplay-bluefruit.ld
```diff
@@ -0,0 +1,10 @@
+
+MEMORY
+{
+    FLASH_TEXT (rw) : ORIGIN = 0x00000000+0x26000, LENGTH = 796K /* SoftDevice S140. See https://learn.adafruit.com/introducing-the-adafruit-nrf52840-feather/hathach-memory-map */
```
aykevl: It may be clearer to describe the length with a formula too: `LENGTH = 1M - 0x26000`. That way it is immediately obvious where that length comes from.

bgould:
I have this board, and I can test this out today.  Thank you for adding this @matloob !

bgould:
I get this error when trying to build/flash:
```
$ tinygo build -target=circuitplay-bluefruit -o /tmp/bluefruit.uf2  src/examples/blinky1/blinky1.go 
error: json: cannot unmarshal string into Go struct field TargetSpec.uf2-family-id of type uint32
```

deadprogram:
DId you rebuild tinygo itself using `go install`? the changes that @matloob made to the uf2 conversion require it.

compileopts/target.go
```diff
@@ -43,6 +43,7 @@ type TargetSpec struct {
 	FlashMethod      string   `json:"flash-method"`
 	FlashVolume      string   `json:"msd-volume-name"`
 	FlashFilename    string   `json:"msd-firmware-name"`
+	UF2FamilyID      uint32   `json:"uf2-family-id"`
```
bgould: I think this property needs to be added to the copyProperties() method below, something like:
```
  if spec2.UF2FamilyID != 0 {                                                   
    spec.UF2FamilyID = spec2.UF2FamilyID                                                                                                                                  
  }
```
Otherwise the value of this property is lost in resolveInherits()

targets/circuitplay-bluefruit.json
```diff
@@ -0,0 +1,9 @@
+{
+    "inherits": ["nrf52840"],
+    "build-tags": ["circuitplay_bluefruit"],
+    "flash-method": "msd",
+    "msd-volume-name": "CPLAYBTBOOT",
+    "msd-firmware-name": "firmware.uf2",
+    "uf2-family-id": "0xADA52840",
```
bgould: I was only able to make this work by specifying the uf2-family-id value as a base 10 integer, ie:
```
"uf2-family-id": 2913282112,
```
otherwise `encoding/json` complains about attempting to unmarshal a string into a uint32

bgould:
The blinky1 and ws2818 examples are working for me using the small tweaks that I've noted above.  However I'm not getting any serial device when running programs... should there be one or should I be using a serial converter or something?  I tried flash circuitpython onto the board and it creates /dev/ttyACM0 when connected

deadprogram:
The Circuit Playground Bluefruit is using USB CDC to communicate via serial, similar to how the SAMD21/SAMD51 boards do.

Our options seems to be:
- implement this ourselves, similar to what we have done with the SAMD21/SAMD51
- write a CGo wrapper around the aptly named https://github.com/hathach/tinyusb which is what all of the Adafruit boards are using now.

Any thoughts from anyone on this?

compileopts/target.go
matloob: Done

src/machine/board_circuitplay_bluefruit.go
matloob: Done

targets/circuitplay-bluefruit.ld
matloob: Done

compileopts/target.go
matloob: Done. Now I'm wondering how this ever worked for me...

targets/circuitplay-bluefruit.json
matloob: Done

aykevl:
I don't like the idea of depending on a C library for such a core feature but it may be the smarter option to reduce the maintenance burden. (It would also be a good use case to prove CGo support). However, I wonder is how well TinyUSB would integrate with TinyGo. For example, what about heap management?

@deadprogram any idea or guess how much work it would be to add support for the nrf52840 for USB? It seems like USB support is not coupled to the SAMD implementation so could theoretically be used from the nrf52840 as well.

In any case, I don't think we should block merging this PR over this issue. Output over USB can be added at a later time. This PR looks fine by me.

compileopts/target.go
deadprogram: I would suggest making this also a `string` and then using `hex.DecodeString()` or whatever to convert to the "real" type.

This just seems like it would make it a lot more readable in the actual JSON files themselves.

What do you think?

targets/circuitplay-bluefruit.json
```diff
@@ -0,0 +1,9 @@
+{
+    "inherits": ["nrf52840"],
+    "build-tags": ["circuitplay_bluefruit"],
+    "flash-method": "msd",
+    "msd-volume-name": "CPLAYBTBOOT",
+    "msd-firmware-name": "firmware.uf2",
+    "uf2-family-id": 2913282112,
```
deadprogram: See my previous comment about making this a string then calling `hex.DecodeString()` to make it more readable.

deadprogram:
Regarding the nrf52 USBD support, it is not trivial task of course. The current USB CDC is *mostly* decoupled I think, but this would be a real test. Also there are as always a number of interesting implementation details that are different.

Regarding using TinyUSB https://github.com/hathach/tinyusb/blob/6ed96cb93b92acf7cce3bd6b84dc5cd8882486c4/docs/getting_started.md#add-tinyusb-to-your-project seems encouraging.

From a bit of a deep dive, https://github.com/hathach/tinyusb/blob/bf76a1e49e6760ef5ac7dd255158ce9475ea8dd1/src/class/cdc/cdc_device.c#L55-L56 seems like TinyUSB allocates its own memory for buffers, but then we can pass our own CGo buffers for transfers: https://github.com/hathach/tinyusb/blob/master/examples/device/cdc_msc/src/main.c#L113-L116

I think it is worth trying a TinyUSB experiment for sure.

bgould:
I used a USB serial converter connected to the TX/RX pads and confirmed I can get output as expected from various examples.  Functionality looks good to me.

deadprogram:
As soon as @matloob can perhaps address my last feedback https://github.com/tinygo-org/tinygo/pull/819/files#r363045397 then I think we're good.

compileopts/target.go
matloob: Sounds good! and this allows us to distinguish between no family id (empty string) and a family id of zero!

targets/circuitplay-bluefruit.json
matloob: Done

matloob:
> As soon as @matloob can perhaps address my last feedback https://github.com/tinygo-org/tinygo/pull/819/files#r363045397 then I think we're good.

Done

deadprogram:
Thank you so very much @matloob for working on this PR and to @bgould for helping test it out.

Now merging.

PR #821, opened 2019-12-30
compiler: add support for CGO_ENABLED environment variable

### LINKED ISSUE #771: How to disable CGO?
Hello.

I need to set something like `CGO_ENABLED=0`. There's some code that has different implementations for cases if CGo is supported (`+build cgo`) and if CGo is not supported (`+build !cgo`). And I need to compile the second variant. It would be nice if you would help me to understand how can I do it without manually modifying all my dependencies (which could use CGo)?

aykevl:
This should fix #771. Not sure how to test it, but you can test locally by running tinygo with the ./testdata/cgo package with the following extra file (main2.go):

```go
// +build !cgo

package main

func main() {
    println("no cgo")
}
```

and run with `CGO_ENABLED` set to `1` or `0` to test the difference.

### DIFF
--- a/goenv/goenv.go
+++ b/goenv/goenv.go
@@ -18,6 +18,7 @@ var Keys = []string{
 	"GOROOT",
 	"GOPATH",
 	"GOCACHE",
+	"CGO_ENABLED",
 	"TINYGOROOT",
 }
 
@@ -57,6 +58,13 @@ func Get(name string) string {
 			panic("could not find cache dir: " + err.Error())
 		}
 		return filepath.Join(dir, "tinygo")
+	case "CGO_ENABLED":
+		val := os.Getenv("CGO_ENABLED")
+		if val == "1" || val == "0" {
+			return val
+		}
+		// Default to enabling CGo.
+		return "1"
 	case "TINYGOROOT":
 		return sourceDir()
 	default:
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -65,6 +65,12 @@ func (c *Config) BuildTags() []string {
 	return tags
 }
 
+// CgoEnabled returns true if (and only if) CGo is enabled. It is true by
+// default and false if CGO_ENABLED is set to "0".
+func (c *Config) CgoEnabled() bool {
+	return goenv.Get("CGO_ENABLED") == "1"
+}
+
 // GC returns the garbage collection strategy in use on this platform. Valid
 // values are "none", "leaking", and "conservative".
 func (c *Config) GC() string {
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -193,7 +193,7 @@ func (c *Compiler) Compile(mainPath string) []error {
 			GOOS:        c.GOOS(),
 			GOROOT:      goenv.Get("GOROOT"),
 			GOPATH:      goenv.Get("GOPATH"),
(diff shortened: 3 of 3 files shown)


goenv/goenv.go
```diff
@@ -57,6 +58,13 @@ func Get(name string) string {
 			panic("could not find cache dir: " + err.Error())
 		}
 		return filepath.Join(dir, "tinygo")
+	case "CGO_ENABLED":
+		val := os.Getenv("CGO_ENABLED")
+		if val == "1" || val == "0" {
+			return val
+		}
+		// Default to enabling CGo.
```
deadprogram: This addressed my only possible concern about this PR.

Not sure how useful it will be, but it does enforce the standard behavior people would expect.

aykevl: Not sure what you mean, is there something that needs changing?

This behavior here is slightly different from the main Go compiler. In TinyGo, CGo is always enabled unless it is explicitly disabled. In the main Go compiler, CGo is disabled by default when cross compiling. That doesn't make much sense for TinyGo as it supports CGo while cross compiling and cross compiling is in fact the main use case of TinyGo.

deadprogram: > This behavior here is slightly different from the main Go compiler.

Exactly, and we really do have to keep it that way. That was what I was saying, and you already handled it the way I would want to. So we're good. :)

xaionaro:
Sorry for the late answer:

Yes, it works, thank you.

```
experiment0@tinygo:~/go/src/github.com/u-root/u-root/bb$ tinygo build -o /tmp/bb main.go 
# os/user
../../../../../../.gimme/versions/go1.13.linux.amd64/src/os/user/cgo_lookup_unix.go:19:6: not implemented: build constraints in #cgo line
/usr/include/x86_64-linux-gnu/bits/confname.h:222:31: unexpected token ILLEGAL
/usr/include/x86_64-linux-gnu/bits/confname.h:224:31: unexpected token ILLEGAL
/usr/include/x86_64-linux-gnu/bits/confname.h:222:31: unexpected token ILLEGAL
/usr/include/x86_64-linux-gnu/bits/confname.h:224:31: unexpected token ILLEGAL
/usr/include/x86_64-linux-gnu/bits/confname.h:222:31: unexpected token ILLEGAL
/usr/include/x86_64-linux-gnu/bits/confname.h:224:31: unexpected token ILLEGAL
```
vs
```
experiment0@tinygo:~/go/src/github.com/u-root/u-root/bb$ CGO_ENABLED=0 tinygo build -o /tmp/bb main.go 
# os/user
../../../../../../.gimme/versions/go1.13.linux.amd64/src/os/user/lookup_stubs.go:91:14: Getgid not declared by package os
../../../../../../.gimme/versions/go1.13.linux.amd64/src/os/user/lookup_stubs.go:81:14: Getuid not declared by package os
../../../../../../.gimme/versions/go1.13.linux.amd64/src/os/user/lookup_stubs.go:30:19: UserHomeDir not declared by package os
../../../../../../.gimme/versions/go1.13.linux.amd64/src/os/user/lookup_stubs.go:34:16: Getenv not declared by package os
```
I will try to temporary add this stuff to the tinygo's environment. It's another issue :)

deadprogram:
Thanks for testing @xaionaro and to @aykevl for the fix.

Now merging.

PR #822, opened 2019-12-30
cgo: add support for identifiers

aykevl:
This doesn't yet support things like `#define foo foo` but that requires more complex compiler changes.


deadprogram:
Steps in the right direction, for sure! Now merging.

PR #823, opened 2019-12-31
loader: improve error messages for failed imports

aykevl:
Add location information (whenever possible) to failed imports. This helps in debugging where an incorrect import came from.

For example, show the following error message:

    /home/ayke/src/github.com/tinygo-org/tinygo/src/machine/machine.go:5:8: cannot find package "foobar" in any of:
        /usr/local/go/src/foobar (from $GOROOT)
        /home/ayke/src/foobar (from $GOPATH)

Instead of the following:

    error: cannot find package "foobar" in any of:
        /usr/local/go/src/foobar (from $GOROOT)
        /home/ayke/src/foobar (from $GOPATH)

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -251,13 +251,17 @@ func (c *Compiler) Compile(mainPath string) []error {
 			return []error{err}
 		}
 	} else {
-		_, err = lprogram.Import(mainPath, wd)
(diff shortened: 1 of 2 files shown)


deadprogram:
This should be very helpful when trying to find misc. packages in dependencies that are not loading correctly. 

Merging.

Issue #515, opened 2019-08-20
proposal: interrupt API


aykevl:
We need to define an API to interact with interrupts in TinyGo in an idiomatic way. At the moment, there is some support but they rely on magic comments and are hard to use correctly. For example:

```go
//go:export RTC1_IRQHandler
func handleRTC1() {
    // ...
}
```

Here are a few requirements I could come up with, for such an API:

* It should be portable across processor architectures. For example: Cortex-M, AVR, traditional ARM, RISC-V (both the older PLIC design and the upcoming standard, which will likely be CLIC-based), perhaps others? I've seen roughly these designs:
  * Full interrupt vector table with hardware stacking/unstacking, as implemented in Cortex-M.
  * Interrupt vector, but apart from saving/restoring the PC the interrupt controller expects the software to handle everything. This is the case for AVR and the new CLIC-based interrupt scheme in RISC-V.
  * Single entry point with software vectoring. This is the case for the old RISC-V design, can optionally be used with the new CLIC-based RISC-V interrupt design, and is used on the older ARM core in the GameBoy Advance.
* It should be efficient to implement. That is, it should not use extra RAM for the vector table on Cortex-M or always require jumping through a handler of some sort if that is not absolutely necessary.
* It should fit well in the Go language. That means it should not conflict too much with normal Go-style programs: lots of goroutines that perform blocking operations and a scheduler that switches between all these goroutines.
* It should be easy to use.

This is the design I came up with uses two different kinds of interrupt handlers, as I recognize there are certain interrupts that are easier to use in one style or in the other:

* Low-level interrupts, much like you would use in your classical C or assembly based embedded application. This should map as closely as possible to the hardware and be as efficient as possible. Such interrupts have limitations, such as being unable to use heap allocation or not being able to perform blocking operations (but non-blocking channel operations should still be allowed). The idea is that you register a function as an interrupt and possibly need to manually enable/configure it.
  These interrupts are usually run in response to certain events when some action needs to be performed, such as pressing a button.
* High level blocking operations. This is closer to how I/O is usually done in Go. The typical example would be waiting for an SPI transmission to be completed before continuing.

Here is my proposed API in pseudo-Go:

```go
// Low level interrupt style
type runtime.Interrupt struct { ... }
func runtime.NewInterrupt(interruptNumber uintptr, handler func()) runtime.Interrupt {}
func (i runtime.Interrupt) Enable() {}
func (i runtime.Interrupt) Disable() {}
[...]

PR #824, opened 2019-12-31
compiler: improve "function redeclared" error

aykevl:
This error can normally only happen with `//go:linkname` and such, but it's nice to get an informative error message in that case.


deadprogram:
I seem to recall we have run into this when upgrading versions of Go.

Also, I am all into informative error messages, so merging.

aykevl:
That's a good case yes, but perhaps more importantly this lays the foundation for better error handling in #515 in case an interrupt is defined twice at different locations.

PR #825, opened 2019-12-31
machine/arduino-nano33: fix UART1 and UART2

aykevl:
UART2 was configured with the wrong SERCOM for the used pins (PB22 and
PB23). However, after changing the SERCOM from 3 to 5 that led to a
conflict with UART1 (used for the on-board WiFi). But the used pins are
also usable from SERCOM 3, so in the end I switched SERCOM5 and SERCOM3
around.

With this change, I was able to get examples/echo working (after changing the UART to UART2 and changing the used pins to RX0 and TX1). I also verified that the on-board WiFi chip is still usable with the espconsole example in the drivers repo.

### DIFF
--- a/src/machine/board_arduino_nano33_baremetal.go
+++ b/src/machine/board_arduino_nano33_baremetal.go
@@ -8,12 +8,12 @@ import "device/sam"
 var (
 	UART1 = UART{
 		Buffer: NewRingBuffer(),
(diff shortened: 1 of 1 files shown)


deadprogram:
I tested with both the WiFiNINA mqttsub example, and also with my flightstick example and both performed as expected.

Now merging, thanks for the refinements!

PR #827, opened 2020-01-01
compiler: add globaldce pass to start of optimization pipeline

aykevl:
This reduces code size in a few cases when tested against the drivers smoketests (although there was one minor increase) without significantly increasing compile time. In fact, in my testing compile time appears to be going down a little bit (around 1%, within the noise).


aykevl:
Note: this PR is required for #782 to work well.

deadprogram:
I also did a little adhoc testing to make sure nothing critical was getting optimized away, and seems good.

Now merging, thanks!

PR #830, opened 2020-01-04
runtime: fix atsamd51 volatile usage

aykevl:
It was still using the (long removed) `//go:volatile` pragma for volatile variables, thus it was only accidentally working.

Tested with examples/blinky1 on an ItsyBitsy M4.


deadprogram:
Oops, that was probably how long it took to land that file from when it got started.

It is indeed amazing that it was working. cc/ @bgould 

Now merging. Thanks for finding this @aykevl

PR #603, opened 2019-09-30
machine/samd21: switch UART to use new pin configuration

aykevl:
This allows all possible UART pin configurations to be used and avoids some tricky configuration.

Will rebase on top of dev once #597 lands.

PR #831, opened 2020-01-05
Improve UART interrupts on atsamd51

aykevl:
This is a combination of 3 commits. The first was to get UART1 working at all. The second was something that looked like an obvious bug - but I didn't test whether I2C still works. The third is what this PR is all about. Here is the commit message:

```
A small footnote in the datasheet says that interrupt source numbers
correspond to the bit position in INTFLAG. We only need the RXC
interrupt for UART. In other words, ony the _2 interrupts (RXC is in the
2nd bit position) needs to be used for UART to work correctly.

In the future, more interrupts may be needed. They can then be added as
necessary.
```

(This PR is here to make #782 easier to complete).

### DIFF
--- a/src/machine/machine_atsamd51.go
+++ b/src/machine/machine_atsamd51.go
@@ -627,6 +627,7 @@ type UART struct {
 	Buffer *RingBuffer
 	Bus    *sam.SERCOM_USART_INT_Type
 	SERCOM uint8
+	IRQVal uint32 // RXC interrupt
 }
 
 var (
@@ -638,6 +639,7 @@ var (
 		Buffer: NewRingBuffer(),
 		Bus:    sam.SERCOM3_USART_INT,
 		SERCOM: 3,
+		IRQVal: sam.IRQ_SERCOM3_2, // RXC interrupt
 	}
 
 	// The second hardware serial port on the SAMD51. Uses the SERCOM0 interface.
@@ -645,6 +647,7 @@ var (
 		Buffer: NewRingBuffer(),
 		Bus:    sam.SERCOM0_USART_INT,
 		SERCOM: 0,
+		IRQVal: sam.IRQ_SERCOM0_2, // RXC interrupt
 	}
 )
 
@@ -696,7 +699,7 @@ func (uart UART) Configure(config UARTConfig) error {
 	config.TX.Configure(PinConfig{Mode: txPinMode})
 	config.RX.Configure(PinConfig{Mode: rxPinMode})
 
-	// reset SERCOM0
+	// reset SERCOM
 	uart.Bus.CTRLA.SetBits(sam.SERCOM_USART_INT_CTRLA_SWRST)
 	for uart.Bus.CTRLA.HasBits(sam.SERCOM_USART_INT_CTRLA_SWRST) ||
 		uart.Bus.SYNCBUSY.HasBits(sam.SERCOM_USART_INT_SYNCBUSY_SWRST) {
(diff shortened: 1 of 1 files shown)


deadprogram:
@bgould if you are able to test this PR with your current setup, it would be extremely helpful.

bgould:
Sorry I didn't notice this one until now.  I'll see if I can get this tested in the next day or two

aykevl:
I changed this PR to make the first commit like https://github.com/tinygo-org/tinygo/pull/603 instead of the change I did before (that I didn't fully trust anyway). I've tested both UART1 and UART2 on an ItsyBitsy M4 with the echo example. You can test it with a patch like this to the echo example:

```diff
diff --git a/src/examples/echo/echo.go b/src/examples/echo/echo.go
index 97a5dc4c..673de13b 100644
--- a/src/examples/echo/echo.go
+++ b/src/examples/echo/echo.go
@@ -9,13 +9,20 @@ import (
 
 // change these to test a different UART or pins if available
 var (
-       uart = machine.UART0
-       tx   = machine.UART_TX_PIN
-       rx   = machine.UART_RX_PIN
+       //uart = machine.UART1
+       //tx   = machine.UART_TX_PIN
+       //rx   = machine.UART_RX_PIN
+       uart = machine.UART2
+       tx   = machine.UART2_TX_PIN
+       rx   = machine.UART2_RX_PIN
 )
 
 func main() {
-       uart.Configure(machine.UARTConfig{TX: tx, RX: rx})
+       time.Sleep(time.Second)
+       time.Sleep(time.Second)
+       time.Sleep(time.Second)
+       err := uart.Configure(machine.UARTConfig{TX: tx, RX: rx})
+       println("error:", err)
        uart.Write([]byte("Echo console enabled. Type something then press enter:\r\n"))
 
        input := make([]byte, 64)
```

bgould:
I tested this on Metro M4 and it is working; also I added an appropriate UART on SERCOM4 to local machine_pyportal.go and confirmed it is working there as well.  A similar change is necessary to test Feather M4 and PyBadge I think; I can test Feather M4 as well if needed but I'm personally I'm convinced this approach is working.

I do however think that maybe these should be moved into the board-specific files since it doesn't make sense to always have UART1 be SERCOM3 for instance (though that was true of the first 2 samd51 boards that were implemented, which presumably is why this is here): https://github.com/tinygo-org/tinygo/blob/e1824c18d782158698f80e94c19e90dc52649b8c/src/machine/machine_atsamd51.go#L633-L652

I don't know if it makes sense to move those as part of this PR or a separate one, but I just wanted to mention that.  In either case, this PR certainly makes doing that easier and cleaner than it was before.

bgould:
Also, I think it is probably possible to eliminate the IRQVal field from the UART definition by doing something similar to what is implemented for samd21; that is to say that it is possible to derive IRQVal from the value of SERCOM.

If I am reading it correctly, this could be done for samd51 with something like:

    arm.EnableIRQ(sam.IRQ_SERCOM0_2 + (4 * uart.SERCOM))

deadprogram:
I am going to merge this PR, and then these UARTs can be refactored in another PR using the new interrupt API.

Thanks everyone!

aykevl:
> I do however think that maybe these should be moved into the board-specific files since it doesn't make sense to always have UART1 be SERCOM3 for instance

Yeah they really should, but I wanted to do that in a separate PR (or someone else can do that, of course).

> Also, I think it is probably possible to eliminate the IRQVal field from the UART definition by doing something similar to what is implemented for samd21; that is to say that it is possible to derive IRQVal from the value of SERCOM.

That's a good idea, but now that #782 has been merged this will be revised anyway.

PR #832, opened 2020-01-05
ci: produce zip as artifact in Azure Pipelines

aykevl:
Zip files are far more often used on Windows systems, so we provide zip files in releases. Unfortunately, previously the provided artifact in CI was really just a compressed .tar.gz file, which defeats the purpose.

This commit zips the release tarball itself so it can be downloaded as an artifact.


aykevl:
The build appears failed because I cancelled all the CircleCI builds, but those are not relevant to this PR.

deadprogram:
Yes, I can see that this did not change anything except the Azure build.

Now merging.

PR #833, opened 2020-01-05
atsamd51: Fixes/Improvements for ADC pins

bgould:
I found 2 problems with ADC handling for ATSAMD51 boards

1) in InitADC(), 10-bit resolution is being set instead of 12-bit resolution, and ADC.Get() assumes 12 bit resolution
2) The bits denoting which pin to use on an ADC bus are not cleared to be reset on subsequent reads. This causes the first pin that is read on an ADC bus to be the only pin that is ever read on that bus.  Also the "Data Ready" flag was not being cleared properly after the first garbage conversion, which causes inconsistent and incorrect values to be returned.

This PR fixes all described above, and is necessary to complete https://github.com/tinygo-org/tinygo/pull/800


deadprogram:
Thank you very much @bgould for researching and fixing this! Now merging.

PR #837, opened 2020-01-09
RISC-V interrupts

aykevl:
This too me a looong time to figure out between compiler bugs, few examples, and a small typo that cost me hours (`ret` vs `mret`).

You can now run the blinky2 example in QEMU and on the HiFive1b hardware:

```
tinygo run -target=hifive1-qemu examples/blinky2
tinygo flash -target=hifive1b examples/blinky2
```


deadprogram:
Seems like this should also free us up to implement the UART on this board.

Great work! Merging.

aykevl:
Note: the compiler bug mentioned above has now been fixed: https://reviews.llvm.org/D74596
This means that interrupts can be made more efficient, if needed.

PR #838, opened 2020-01-10
riscv: print exception PC and code

aykevl:
This can be useful for debugging critical bugs in code. I haven't added human-readable exceptions (such as "illegal instruction" or "stack overflow") yet, they can be added when they happen in practice (to avoid increasing code size unnecessarily).

A program like this:

```go
    // ...
    riscv.Asm(".long 0") // undef
    // ...
```

will cause an error like the following:

    fatal error: exception with mcause=2 pc=0x204002ce

where code 2 means "illegal instruction" according to the RISC-V docs (as expected).

Tested both on hardware (HiFive1 rev B) and in QEMU.

### DIFF
--- a/src/runtime/runtime_fe310.go
+++ b/src/runtime/runtime_fe310.go
@@ -69,8 +69,10 @@ func handleInterrupt() {
 			riscv.MIE.ClearBits(1 << 7) // MTIE bit
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


deadprogram:
Seems like a reasonable approach to me. Now merging. Thanks!

PR #839, opened 2020-01-10
hifive1b: add support for gdb subcommand

aykevl:
This makes debugging on the HiFive1 rev B much easier:

    tinygo gdb -target=hifive1b examples/echo


deadprogram:
Very useful, we really need this so we can the other interfaces like UART working.

Merging.

PR #840, opened 2020-01-11
machine/digispark: add clock speed and pin mappings

deadprogram:
This PR adds the clock speed and pin mappings to the existing Digispark board definition.


aykevl:
Looks good to me!

PR #841, opened 2020-01-11
machine/pyportal: remove manual SPI pin mapping as now handled by defaults

deadprogram:
This PR corrects the SPI pin mapping in the PyPortal board so that it builds correctly now that #802 has been merged.


aykevl:
Looks good to me. At the very least, it fixes CI.

PR #842, opened 2020-01-12
Fault handler

aykevl:
When this callback is set, it will be called when a critical error happens (such as a HardFault). Additionally, it will be called when a runtime panic happens if the `-panic=trap` option is used during compilation.

This PR does not implement handling of regular panics yet.

This PR is marked as WIP as it causes a ~200 byte increase in binary size, which is much higher than I expected. I'll see whether I can reduce that a bit.

@magiconair


deadprogram:

PR #846, opened 2020-01-13
gameboy-advance: include compiler-rt in build

aykevl:
This avoids errors like the following:

    ld.lld-9: error: undefined symbol: __umodsi3
    ld.lld-9: error: undefined symbol: __aeabi_uidivmod


deadprogram:
@conejoninja can you please test this with your newest set of GBA code?

conejoninja:
I used the interrupts, so this branch alone complains about not finding "runtime/interrupt" package, but interrupts-gba-2+gba-rtlib compiled without issues. I don't remember every error I previously got, but the ones I tried it worked fine with this change. :+1:

PR #848, opened 2020-01-14
main: fix race condition in tests

aykevl:
This caused most tests to run the zeroalloc.go test instead of what they should have been tested, and in turn explains most of the performance gains of parallel testing.

This commit fixes it by avoiding race conditions. Luckily, no tests started failing since then due to this.

@jaddr2line


jaddr2line:
oops. . .

aykevl:
Thanks for the quick review!
This must be the number one cause of race conditions in Go code, because it _looks_ correct.

PR #849, opened 2020-01-15
builder: refactor compiler-rt library

aykevl:
This refactor makes adding a new library (such as a libc) much easier in the future as it avoids a lot of duplicate code. Additionally, CI should become a little bit faster (~15s) as `build-builtins` now uses the build cache.

Example for picolibc:

```go
var picolibc = Library{
        name: "picolibc",
        cflags: func() []string {
                picolibcDir := filepath.Join(goenv.Get("TINYGOROOT"), "lib/picolibc/newlib/libc")
                return []string{"-Werror", "-Wall", "-std=c11", "-D_COMPILING_NEWLIB", "-D_POSIX_C_SOURCE=200809L", "-fshort-enums", "--sysroot=" + picolibcDir, "-I" + picolibcDir + "/tinystdio"}
        },
        sourceDir: "lib/picolibc/newlib/libc",
        sources: func(target string) []string {
                return picolibcSources
        },
}
```

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -129,21 +129,18 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(stri
 			return err
 		}
 
-		// Load builtins library from the cache, possibly compiling it on the
-		// fly.
-		var librt string
-		if config.Target.RTLib == "compiler-rt" {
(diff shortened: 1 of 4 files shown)


aykevl:
Replaced by #871.

PR #852, opened 2020-01-17
machine: avoid bytes package in USB logic

aykevl:
This greatly cuts down on compile time (by about 5x for small programs) and also makes the program a whole lot smaller. Overall it cuts down `make smoke-test` in the drivers repository by half (from 160s to 80s).

This will probably also fix the timeout issue in the Playground: https://github.com/tinygo-org/playground/issues/7

### DIFF
--- a/src/machine/usb.go
+++ b/src/machine/usb.go
@@ -3,7 +3,6 @@
 package machine
 
 import (
-	"bytes"
 	"errors"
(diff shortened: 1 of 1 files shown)


deadprogram:
Wow, the compile speed difference is really noticable! Great work, @aykevl you are going to save me so much time.

Now merging.

PR #853, opened 2020-01-18
machine/samd51: refactor i2c pin handling to auto-detect pin mode

aykevl:
With this change, it's no longer necessary to set a specific pin mode: it will get autodetected in the `Configure()` call.

Tested on an ItsyBitsy M4 with the mpu6050 example in the drivers repo.


deadprogram:
Very helpful to prevent incorrect settings.

Now merging.

PR #856, opened 2020-01-20
Support maps with arbitrary key types

aykevl:
This implementation simply casts types without special support to an interface, to make the implementation simpler and possibly reducing the code size too. It will likely be slower than the canonical Go implementation though (which builds special compare and hash functions at compile time).


aykevl:
Note: the JSON support I've been working on depends on this.

deadprogram:
Any other opinions before I merge this? @jaddr2line any thoughts?

jaddr2line:
Looks ok

deadprogram:
Now merging, thanks!

PR #858, opened 2020-01-21
flash: remove default port check for Digispark

deadprogram:
This PR remove default port check for Digispark as micronucleus communicates directly using HID. 

In the current release, running `tinygo flash` using  `-target digispark` results in error "unable to locate a USB device to be flashed".

### DIFF
--- a/main.go
+++ b/main.go
@@ -145,14 +145,6 @@ func Test(pkgName string, options *compileopts.Options) error {
 
 // Flash builds and flashes the built binary to the given serial port.
 func Flash(pkgName, port string, options *compileopts.Options) error {
-	if port == "" {
-		var err error
-		port, err = getDefaultPort()
-		if err != nil {
-			return err
-		}
-	}
-
 	config, err := builder.NewConfig(options)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


aykevl:
The "flash method" switch was really intended for entirely separate flash methods. In this case, `micronucleus` really means `command` without requesting a serial port.
A cleaner solution would be one of:
1. Make `micronucleus` a truly new flash method.
2. Only look for a port when it is really needed.

Actually, I think the requirement for a serial port should be lifted for many other targets as well. For example, you don't need a serial port when flashing using OpenOCD (but it usually works because usually a serial port is connected). So my proposal would be treat micronucleus like any other command, and only look for a serial port when needed. That means the following code would come both in the `if config.Target.PortReset == "true" {` block and when replacing the `{port}` string (if it exists in the command string).

```go
	if port == "" {
		var err error
		port, err = getDefaultPort()
		if err != nil {
			return err
		}
	}
```

deadprogram:
OK requested changes made and force pushed.

aykevl:
Great, thank you! Tested and works for the DigiSpark. Merging now.

PR #859, opened 2020-01-21
all: add Arduino Nano support

aykevl:
Completely untested apart from that the Uno still works.

@alankrantas can you test this PR? I don't have a Nano.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -253,6 +253,8 @@ smoketest:
 ifneq ($(AVR), 0)
 	$(TINYGO) build -size short -o test.hex -target=arduino             examples/blinky1
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=arduino-nano        examples/blinky1
+	@$(MD5SUM) test.hex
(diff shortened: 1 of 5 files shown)


alankrantas:
"error: open C:\tinygo\targets\atmega328p.json: The system cannot find the file specified."

Then I copied arduino.json as arduino-nano.json and simply added -b 57600 in the flash command, that works.

Also, Nano has analog in pins A0 to A7. Other pins are the same as Uno. Some Uno-clones are actually using Atmega328 just like Nanos, and they are always interchangeable with Atmega328p Unos.

aykevl:
To test this PR, you also need to copy over src/machine/board_arduino_nano.go.

aykevl:
Great! Thank you for testing this quickly!

A bit late, but now I remember that you can also download the build for this pull request here: https://dev.azure.com/tinygo/tinygo/_build/results?buildId=822&view=artifacts&type=publishedArtifacts

deadprogram:
Thanks everyone, now merging.

PR #860, opened 2020-01-21
RISC-V external interrupts

aykevl:
This PR implements interrupts for the RISC-V PLIC and uses this for UART interrupts. With this PR, the echo example works without modification.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,7 +116,7 @@ gen-device-sam: build/gen-device-svd
 	GO111MODULE=off $(GO) fmt ./src/device/sam
 
 gen-device-sifive: build/gen-device-svd
-	./build/gen-device-svd -source=https://github.com/posborne/cmsis-svd/tree/master/data/SiFive-Community lib/cmsis-svd/data/SiFive-Community/ src/device/sifive/
+	./build/gen-device-svd -source=https://github.com/posborne/cmsis-svd/tree/master/data/SiFive-Community -interrupts=software lib/cmsis-svd/data/SiFive-Community/ src/device/sifive/
 	GO111MODULE=off $(GO) fmt ./src/device/sifive
 
 gen-device-stm32: build/gen-device-svd
--- a/src/machine/machine_fe310.go
+++ b/src/machine/machine_fe310.go
@@ -4,6 +4,7 @@ package machine
 
 import (
 	"device/sifive"
+	"runtime/interrupt"
 )
 
 func CPUFrequency() uint32 {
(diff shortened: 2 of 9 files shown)


deadprogram:
I rebuilt the compiler, also ran `make gen-device-sifive` without problem.

However, when I run `tinygo flash -target hifive1b examples/echo` the compiler succeeds and flashes the board, but I have not been able to connect to it via picocom correctly.

This connects, but never responds to anything:

```
picocom -b 115200 /dev/ttyACM0
...
Terminal ready
```

This connects to what appears to be the debug output from the onboard ESP32:

```
picocom -b 115200 /dev/ttyACM1
...
 Terminal ready
ets Jun  8 2016 00:22:57

rst:0x1 (POWERON_RESET),boot:0x12 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:2
load:0x3fff0018,len:4
load:0x3fff001c,len:5552
ho 0 tail 12 room 4
load:0x40078000,len:0
load:0x40078000,len:13756
entry 0x40078fb4
I (30) boot: ESP-IDF v3.0.1 2nd stage bootloader
...
```

Am I doing something wrong here?

aykevl:
I experimented a bit and it seems like there is a floating input somewhere when I don't connect to the UART quicly enough:

    ��������������������panic: runtime error: index out of range

I'll see whether there is a way to pull the pin high or something.

aykevl:
I think it's a hardware problem. My findings:

* If I keep picocom connected while flasing, flasing fails with an erase error.
* If I connect a separate serial-to-USB to the serial pins (D0, D1), flashing also fails.

Therefore, it seems like the chip needs to be disconnected from serial to flash correctly.

* With a pulldown, I quickly hit a `panic: index out of range`. I guess this is how UART works.
* An external 1KΩ pullup doesn't seem to avoid this issue either, so an internal pullup probably won't help.

So I think the solution is just to disconnect UART while flashing, and to quickly connect it again while the chip is initializing (right after the green LED went on and off again). It also makes me wonder whether it's a good idea to enable UART interrupts unconditionally, as it will probably continuously interrupt when no UART is connected.

Thinking a bit more, I really wonder what's going on here. If even an external pullup won't help, what will? It feels like the programmer/debug chip on the board has faulty firmware and sends some garbage until someone connects using `picocom` or similar.

aykevl:
I tried updating the firmware on the HiFive1b. That fixed the bug that you couldn't flash while the serial was connected, but otherwise didn't fix the weird characters on the input.

```
~/src/github.com/tinygo-org/tinygo$ JLinkExe
SEGGER J-Link Commander V6.60f (Compiled Jan 20 2020 16:18:52)
DLL version V6.60f, compiled Jan 20 2020 16:18:34

Connecting to J-Link via USB...Updating firmware:  J-Link OB-K22-SiFive compiled Dec 12 2019 16:26:28
Replacing firmware: J-Link OB-K22-SiFive compiled May 27 2019 15:39:22
FAILED: Communication timed out - Can not execute firmware update
Cannot connect to J-Link
J-Link>

~/src/github.com/tinygo-org/tinygo$ JLinkExe
SEGGER J-Link Commander V6.60f (Compiled Jan 20 2020 16:18:52)
DLL version V6.60f, compiled Jan 20 2020 16:18:34

Connecting to J-Link via USB...Updating firmware:  J-Link OB-K22-SiFive compiled Dec 12 2019 16:26:28
Replacing firmware: J-Link OB-K22-SiFive compiled Sep  7 2018 BTL     
Waiting for new firmware to boot
New firmware booted successfully
O.K.
Firmware: J-Link OB-K22-SiFive compiled Dec 12 2019 16:26:28
Hardware version: V1.00
S/N: 979000738
VTref=3.300V


Type "connect" to establish a target connection, '?' for help
J-Link>
```

deadprogram:
When the ESP32 first boots, the UART that it uses for the debug port outputs data at 57600 baud, until it gets to the point in its own bootloader that it switches to the 115200 baudrate.

Perhaps just pausing on startup to let the ESP32 finish booting before connecting the TinyGo runtime to the UART?

deadprogram:
According to ESP32 docs here https://github.com/espressif/esptool/wiki/ESP32-Boot-Mode-Selection#other-pins if the ESP32 pin GPIO15 (MTDO) is pulled down, then boot messages will be suppressed.

According to the HiFive1B schematics here https://sifive.cdn.prismic.io/sifive%2Fa4546ced-0922-4d87-9334-e97c1a9fd9a5_hifive1.b01.schematics.pdf that pin is connected to the Freedom E310 via its GPIO_9 (SPI_CS2) which is pulled up by default due to resistor R10. Hence why the boot messages are displayed before we are ready for them.

deadprogram:
Since this mostly works, let's merge and mark this feature as experimental and continue working on it in additional PRs.

PR #861, opened 2020-01-21
GBA interrupts

aykevl:
This is based on #860.
By using the same vectoring method as RISC-V uses, this interrupt system does not need any RAM to store the interrupt vector table. Instead, it is implemented by the compiler as a big switch statement.
Other than that, this interrupt system is very bare bones. More features (such as priorities) could be added later if needed. However, I think it would be a good idea to consider interrupt priorities on RISC-V at the same time as the interrupt systems are very similar so that at least the design can be shared.

@kylelemons

### DIFF
--- a/src/machine/machine_gameboyadvance.go
+++ b/src/machine/machine_gameboyadvance.go
@@ -8,6 +8,25 @@ import (
 	"unsafe"
 )
 
+// Interrupt numbers as used on the GameBoy Advance. Register them with
+// runtime/interrupt.New.
+const (
+	IRQ_VBLANK  = 0
+	IRQ_HBLANK  = 1
+	IRQ_VCOUNT  = 2
+	IRQ_TIMER0  = 3
+	IRQ_TIMER1  = 4
+	IRQ_TIMER2  = 5
+	IRQ_TIMER3  = 6
+	IRQ_COM     = 7
+	IRQ_DMA0    = 8
+	IRQ_DMA1    = 9
+	IRQ_DMA2    = 10
+	IRQ_DMA3    = 11
+	IRQ_KEYPAD  = 12
+	IRQ_GAMEPAK = 13
+)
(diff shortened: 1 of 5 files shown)


conejoninja:
I can't say about the interrupt system, but my little example code compiles and works fine.

The smoke tests failed, so at least that needs to be fixed

aykevl:
I hope I fixed it with a force-push. It's some Go 1.12 and below compatibility issue (shift of signed numbers).

aykevl:
Finally fixed all the issues. Now the tests pass.

aykevl:
Rebased on top of the dev branch (after #837 was merged) and updated the commit message.

aykevl:
The only thing I'm not so sure about is where to store the interrupt numbers. They're in the machine package now because I didn't think the runtime/interrupt package was a good place, but the runtime/interrupt package is still aware of the number of interrupts (the `i < 14`) so maybe it doesn't help much after all.

deadprogram:
@conejoninja can you please try this branch one more time with your code before we merge?

conejoninja:
This compiles the code (could be found here ; https://github.com/conejoninja/gbatest I have a more complex demo, but had to simplify it to get some speed + avoid errors ) and it runs as expected.

This branch + gba-rtlib (already in dev) compiles and run fine with the added line (this alone gaves an undefined symbol: __umodsi3): 

```go
krgb = krgb%30
```

aykevl:
Great!

> This branch + gba-rtlib (already in dev) compiles and run fine with the added line (this alone gaves an undefined symbol: __umodsi3):
> 
>     krgb = krgb%30

Interesting! It looks like this has not been implemented in compiler-rt. I guess most cores have support for this already?
The modulo operator is pretty expensive, you could perhaps avoid it with `krgb = krgb%32`.

PR #862, opened 2020-01-21
main: differentiate between various serial/USB error messages

aykevl:
This way it is possible to determine the source of an error when it happens.


deadprogram:
Makes sense. Now merging.

PR #548, opened 2019-09-14
added teensy36 pin configuration

teoman002:

PR #863, opened 2020-01-22
Teensy 3.6

firelizzard18:
I want to use TinyGo for the Teensy 3.6, which is powered by the MK66FX1M0VMD18, a NXP Kinetis K66 processor. TinyGo does not currently support any NXP device.

* [x] Add SVD
   - posborne/cmsis-svd#93
* [x] `src/device/nxp/mk66f18.go`
   - I updated `gen-device-svd` to work with NXP SVDs
* [x] `src/runtime/runtime_nxpmk66f18.go` (and maybe `runtime_nxp.go`)
* [x] `src/machine/machine_nxpmk66f18.go` (and maybe `machine_nxp.go`)
* [x] `src/machine/board_teensy36.go`
   - I think I can just pull this from #548
* [x] `targets/nxpmk66f18.ld`, `targets/nxpmk66f18.json`
* [x] `targets/teensy36.json`


firelizzard18:
I would love any help I can get on this. I am out of my depth when it comes to the low level stuff, such as the runtime files and linker scripts. @PaulStoffregen, would you be willing to advise and/or help?

aykevl:
What I usually do is to try and find the most simple bare-metal C code that works, without any SDK. Just turning a LED on/off in an endless loop for example, without any delay even. You can often even ignore clock initialization. And then try and port it to Go, writing it directly in the `Reset_Handler`.

aykevl:
For the linker script, you could start with this (copied from the microbit):

```
MEMORY
{
    FLASH_TEXT (rw) : ORIGIN = 0x00000000, LENGTH = 256K /* .text */
    RAM (xrw)       : ORIGIN = 0x20000000, LENGTH = 16K
}

_stack_size = 2K;

INCLUDE "targets/arm.ld"
```

Flash/RAM is much bigger than that, but it should work.

This should also be a useful resource: https://www.seanet.com/~karllunt/bareteensy31.html

firelizzard18:
I made a stripped down blink program using the [Teensy 3 Arduino core](https://github.com/PaulStoffregen/cores/tree/master/teensy3). Besides `main.c`, the binary only includes `mk20dx128.c` and `pins_teensy.c` from the core. I have been removing bits from those files to see how small I can get.

firelizzard18:
I am trying to get blinky.go to compile, but I am running into odd errors. I'm building on macOS. Maybe XCode's clang is weird?

```golang
package main

import (
	"device/nxp"
	"runtime/volatile"
	"time"
	"unsafe"
)

func main() {
	mode := (*volatile.Register8)(unsafe.Pointer(uintptr(0x43FE1294)))
	config := nxp.PORTC.PCR5

	mode.Set(1)
	config.SetBits(nxp.PORT_PCR5_SRE | nxp.PORT_PCR5_DSE | 1<<nxp.PORT_PCR0_MUX_Pos)
	config.ClearBits(nxp.PORT_PCR5_ODE)

	for {
		nxp.GPIOC.PSOR.SetBits(1 << 5)
		time.Sleep(time.Millisecond * 1000)

		nxp.GPIOC.PCOR.SetBits(1 << 5)
		time.Sleep(time.Millisecond * 1000)
	}
}
```

```log
warning: optimization flag '-fsingle-precision-constant' is not supported
warning: argument unused during compilation: '-mthumb'
warning: argument unused during compilation: '-mfloat-abi=hard'
warning: argument unused during compilation: '-mfpu=fpv4-sp-d16'
TINYGO/src/device/arm/cortexm.s:1:1: error: unknown directive
.syntax unified
^
TINYGO/src/device/arm/cortexm.s:3:33: error: unexpected token in '.section' directive
.section .text.HardFault_Handler
                                ^
TINYGO/src/device/arm/cortexm.s:5:1: error: unknown directive
.type HardFault_Handler, %function
^
TINYGO/src/device/arm/cortexm.s:10:5: error: unknown use of instruction mnemonic without a size suffix
    mov r0, sp
    ^
TINYGO/src/device/arm/cortexm.s:16:14: error: unknown token in expression
    movs r3, #0
             ^
TINYGO/src/device/arm/cortexm.s:17:13: error: brackets expression not supported on this target
    ldr r3, [r3]
            ^
TINYGO/src/device/arm/cortexm.s:18:5: error: unknown use of instruction mnemonic without a size suffix
    mov sp, r3
    ^
TINYGO/src/device/arm/cortexm.s:21:5: error: invalid instruction mnemonic 'bl'
    bl handleHardFault
    ^~
TINYGO/src/device/arm/cortexm.s:25:31: error: unexpected token in '.section' directive
.section .text.SemihostingCall
                              ^
TINYGO/src/device/arm/cortexm.s:27:1: error: unknown directive
.type SemihostingCall, %function
^
TINYGO/src/device/arm/cortexm.s:29:5: error: invalid instruction mnemonic 'bkpt'
    bkpt 0xab
    ^~~~
TINYGO/src/device/arm/cortexm.s:30:5: error: invalid instruction mnemonic 'bx'
    bx lr
    ^~
error: failed to build src/device/arm/cortexm.s: failed to compile using built-in clang
```

I got the address for `mode` by expanding `*portModeRegister(13) = 1`: https://godbolt.org/z/tAR6U-.

aykevl:
Those Clang errors indicate that Clang does not get the correct `--target` flag. Take a look at this file for example: https://github.com/tinygo-org/tinygo/blob/master/targets/atsamd51g19a.json

My recommendation would be to just copy/modify that file, you can ignore float at least during the initial porting:

```json
{
	"inherits": ["cortex-m"],
	"llvm-target": "armv7em-none-eabi",
	"build-tags": ["teensy36", "teensy", "mk66f18", "nxp"],
	"cflags": [
		"--target=armv7em-none-eabi",
		"-Qunused-arguments"
	],
	""linkerscript": "targets/nxpmk66f18.ld",
	"extra-files": [
		"src/device/nxp/mk66f18.s"
	],
	"flash-command": "teensy_loader_cli -mmcu=mk66fx1m0 -v -w {hex}"
}
```

aykevl:
> I got the address for `mode` by expanding `*portModeRegister(13) = 1`: https://godbolt.org/z/tAR6U-.

That's an odd address. Maybe it is using bitbanding? Register addresses usually form a certain pattern.

firelizzard18:
> That's an odd address. Maybe it is using bitbanding? Register addresses usually form a certain pattern.

Ah, yes! There are bitband macros, but I had no idea what they were about. A partial expansion is `*((volatile uint8_t *)(GPIO_BITBAND_PTR(GPIOC_PDOR, CORE_PIN13_BIT) + 160)) = 1;` and the macros are:

```C
#define GPIO_BITBAND_ADDR(reg, bit) (((uint32_t)&(reg) - 0x40000000) * 32 + (bit) * 4 + 0x42000000)
#define GPIO_BITBAND_PTR(reg, bit) ((uint32_t *)GPIO_BITBAND_ADDR((reg), (bit)))
```

I updated `teensy36.json` so it builds now. I left in the float flags because its not complaining about those, but I did remove a number of other flags. Now I just have to get it to work!

firelizzard18:
I added bitband support (behind build tags) to `runtime/volatile`, so my test program is much cleaner: https://gist.github.com/firelizzard18/eaaa532f7ee18a7619b3b682490960ee/f82dac8ddbfdf3e993d6eed1a831df092c6004f0.

aykevl:
You should be able to simplify things by not using bitbanding. These regions are really just aliases for bits in registers. Instead, you can use SetBits and ClearBits which were designed for this purpose.

To be honest, I think it would be best to leave it out initially. As far as I know, bitbanding is mainly an optimization.

src/runtime/runtime_nxp.go
```diff
@@ -0,0 +1,13 @@
+// +build nxp
+
+package runtime
+
+type timeUnit int64
+
+//go:export Reset_Handler
+func main() {
+	// preinit()
+	// initAll()
```
aykevl: Be aware:
`preinit` initializes all global variables. Before that, they may contain any value.
`initAll` initializes the heap and runs all `init` functions, so may be rather important (`preinit` _must_ be run before that).

aykevl:
Have you gotten the LED to blink?

You could perhaps try without bitbanding. These lines should be equivalent:

```go
nxp.GPIOC.PDDR.Bit(5).Set(1)
nxp.GPIOC.PDDR.SetBits(1 << 5)
```

Also, you can try removing the `time.Sleep`: if the LED works, it should light up. Also: LEDs are often connected to VCC which means that setting the pin to low will enable it (somewhat counterintuitively).

aykevl:
Also, something that looks odd:

```go
nxp.GPIOC.PSOR.SetBits(1 << 5)
// ...
nxp.GPIOC.PCOR.SetBits(1 << 5)
```

I think this should be:

```go
nxp.GPIOC.PSOR.Set(1 << 5)
// ...
nxp.GPIOC.PCOR.Set(1 << 5)
```

If you haven't seen already, this is described in the 2000+ page Reference Manual for this chip, somewhere near the end in the GPIO peripheral.

src/runtime/runtime_nxp.go
firelizzard18: When I first added that stub, I thought `preinit` and `initAll` were something I had to define. I uncommented them once I realized they were defined in `runtime_cortexm.go`

firelizzard18:
I have not tried to load anything onto the board (excluding C/C++ programs). I am currently working on translating the initialization code from the Teensy core libraries. Once I have that translated, I'll start trying to get it to work.

I am using bit banding there because I blindly copied the contents of `pinMode` from the Teensy core libraries and then blindly translated that to go.

I'm not sure why I decided to use `SetBit`. After reading the spec, you are clearly right.

In the current version blinky, I am using `runtime.sleepTicks`, which (at the moment) is just a port of `delay` from the Teensy core libraries and does not sleep but spins until the appropriate duration has elapsed, according to `runtime.ticks`. `runtime.ticks` is a port of `micros` from the Teensy core libraries.

firelizzard18:
I've been putting my test program in a gist: https://gist.github.com/firelizzard18/eaaa532f7ee18a7619b3b682490960ee

firelizzard18:
I tried to create a PR to merge posborne/cmsis-svd into tinygo-org/cmsis-svd, but it seems that I can't do that

firelizzard18:
I have translated all the initialization code, but blink does not work. Time to start debugging...

firelizzard18:
The MK66FX1M0VMD18 requires a 16 byte flash config section at 0x400. To add this, I copied `arm.ld` into `nxpmk66f18.ld` and added it to `.text`. After doing this, I see that `.Lmain.go.main$gowrapper` is located at 0x400 instead of 0x1d0, but I do not see `.flashconfig`. What am I missing?

PaulStoffregen:
Hi, Paul here, the guy behind Teensy.  I can't get involved in Tiny Go, but I would like to quickly bring a couple issues about the 16 byte flash config to your attention.

The FSEC byte controls powerful features, like security which prevents external access to memory, and an option to disable mass erase.  If you activate both of these features, it will permanently disable reprogramming the flash.  Normal programming over USB with Teensy Loader tries to prevent this from ever happening.  It programs one of those bits which needs to be a 1 for that complete lock out to a 0.  If you experiment with arbitrary values for FSEC, you'll see the byte that ends up in flash isn't always the same as what you defined.  That is the lock-out safety feature at work, trying to prevent you from permanently bricking your Teensy!

Teensy Loader does allow you to turn on the security feature.  While this is recoverable, it does have some practical consequences.  Normally when you press the button on Teensy, it enters programming mode, but nothing in flash actually changes until Teensy Loader sends new code.  If you power cycle without programming, your code is still there and runs again.  But security changes all that.  Pressing the button causes complete flash erase.  So be aware that a touch of that button becomes a code "self destruct" for flash memory.  Not a big deal if you were going to upload new code, but something to keep in mind if the button is accidentally pressed.

If you don't use the flash controller to do your own erasing and programming of the flash, using only Teensy Loader to write hex files, you should be safe while experimenting.  But if you write your own code which erases those 16 bytes and writes them with new values, please remember you're running without a safety net.  You can permanently brick your Teensy if the wrong bit pattern ends up in those flash config bytes!

aykevl:
@PaulStoffregen that's very useful information, thank you for sharing!

@firelizzard18 I wanted to say "use GDB with a debug probe to see what's wrong" and while that may be a good idea, I then stumbled on [this post](https://medium.com/@mattmatic/preparing-teensy-3-6-for-swd-b014b0ce2999) which shows that it's unfortunately not easy to set up on the Teensy 3.6. Still, a real debugger (available for under $20, and around $2 if you want to play with AliExpress-quality products) would probably make it much easier to see where the problem lies.

I quickly looked at the linker script and it looks good to me.

firelizzard18:
I have a SWD debugger and a soldering iron, so I'll see if I can hook it up without destroying something :P

firelizzard18:
AFAICT, I fried my teensy, or at least the MKL02. I'll try debugging again once I have the equipment to completely remove the MKL02.

firelizzard18:
Debugging the teensy is a no-go. My soldering skills are not up to that level of work. I might buy a FRDM-K66, but for now I'm going to have fun comparing disassembly listings.

firelizzard18:
I have a bare minimum functioning (blinking) system, with the program below:

```golang
package main

import (
	"device/arm"
	"device/nxp"
	_ "machine"
)

func main() {
	nxp.GPIOC.PDDR.SetBits(1 << 5)
	nxp.PORTC.PCR5.SetBits(nxp.PORT_PCR5_SRE | nxp.PORT_PCR5_DSE | 1<<nxp.PORT_PCR0_MUX_Pos)
	nxp.PORTC.PCR5.ClearBits(nxp.PORT_PCR5_ODE)

	for {
		nxp.GPIOC.PSOR.Set(1 << 5)
		for i := 0; i < 1<<23; i++ {
			arm.Asm("nop")
		}

		nxp.GPIOC.PCOR.Set(1 << 5)
		for i := 0; i < 1<<25; i++ {
			arm.Asm("nop")
		}
	}
}
```

aykevl:
Awesome!
I think you should be able to go from there to a fully working target, but feel free to ask if you have any questions as to how to do that.

firelizzard18:
I can use `gc=none` or `gc=leaking`, but if TinyGo inserts scheduling into the process, it breaks. Since I don't want to debug that via blinking the LED, my plan is implement UART0, then USB serial, then scheduling.

aykevl:
> (which happens if I omit `-gc` and either call `Gosched()` anywhere or uncomment `initAll()`)

`initAll` should really be fixed, without it the program is in an inconsistent state. What is the issue you're getting?

firelizzard18:
> What is the issue you're getting?

The LED doesn't blink. Once I can at least printf-debug, I'll work on that more.

> `initAll` should really be fixed

Working on it. The system (hardware) is usable without it if I compile with `-gc=none`, in the sense that I can write directly to registers and the LED blinks and the UART prints, but as you might expect I can't use the HAL in that state. `initAll()` works when I compile with `-gc=leaking`, so that's where I'm at now. Still no scheduler, but I want buffered serial first, preferably USB serial.

firelizzard18:
I do have a specific question. How can I effectively call `runtime.Gosched()` from the `machine` package? Importing `runtime` would cause an import loop. IIRC Golang uses fancy linkage pragmas, but I'm not sure how to do that. I'm working on translating [`serial_putchar` from the Teensy core](https://github.com/PaulStoffregen/cores/blob/b0c38794ffca838ca458c5c79b994cfbdb69d39f/teensy3/serial1.c#L366):

```C

void serial_putchar(uint32_t c)
{
	uint32_t head, n;

	if (!(SIM_SCGC4 & SIM_SCGC4_UART0)) return;
	head = tx_buffer_head;
	if (++head >= SERIAL1_TX_BUFFER_SIZE) head = 0;
	while (tx_buffer_tail == head) {
		int priority = nvic_execution_priority();
		if (priority <= IRQ_PRIORITY) {
			if ((UART0_S1 & UART_S1_TDRE)) {
				uint32_t tail = tx_buffer_tail;
				if (++tail >= SERIAL1_TX_BUFFER_SIZE) tail = 0;
				n = tx_buffer[tail];
				UART0_D = n;
				tx_buffer_tail = tail;
			}
		} else if (priority >= 256) {
			yield();
		}
	}
	tx_buffer[head] = c;
	transmitting = 1;
	tx_buffer_head = head;
	UART0_C2 = C2_TX_ACTIVE;
}
```

aykevl:
> Still no scheduler, but I want buffered serial first, preferably USB serial.

I would highly recommend doing simple UART first and doing USB-CDC only when everything else is working. USB-CDC is vastly more complex but more importantly requires the heap to work (which in turn requires `initAll` to be run to initialize the heap). You can get serial output with a simple USB-serial adapter.

> I'm working on translating [`serial_putchar` from the Teensy core](https://github.com/PaulStoffregen/cores/blob/b0c38794ffca838ca458c5c79b994cfbdb69d39f/teensy3/serial1.c#L366):

Honestly that code looks way more complicated than necessary. It seems to involve an output buffer and some sort of scheduler, which you can mostly ignore for now (just busy loop instead). Most UART peripherals work by writing the byte to be sent to a special register (which may or may not do buffering in hardware) and then busy-waiting until the byte is sent. Sometimes you can do the reverse: wait until a "buffer empty" flag is set (or a "buffer full" bag is clear) and then writing the byte.

aykevl:
Looking at the datasheet, I suspect writing a byte to UART (after the UART has been configured) can work like this (in pseudocode):

```
func WriteBuyte(c byte) {
    for UARTx_TCFIFO.Get() != 0 {
    }
    UARTx_D.Set(c)
}
```

This is definitely not a fast implementation, but it might help to get started.

src/machine/machine_nxpmk66f18.go
```diff
@@ -83,3 +109,147 @@ func (p FastPin) Clear()       { p.PCOR.Set(true) }
...
+	r.C1.Set(nxp.UART_C1_ILT)
+	r.TWFIFO.Set(2) // tx watermark, causes S1_TDRE to set
+	r.RWFIFO.Set(4) // rx watermark, causes S1_RDRF to set
+	r.PFIFO.Set(nxp.UART_PFIFO_TXFE | nxp.UART_PFIFO_RXFE)
+	r.C2.Set(uartC2TXInactive)
+
+	// if s.Interrupt == interrupt.Interrupt{} {
+	s.Interrupt = u.newISR()
+	s.Interrupt.SetPriority(uartIRQPriority)
+	s.Interrupt.Enable()
```
aykevl: I don't think you'll need an interrupt for sending bytes.

firelizzard18:
> USB-CDC is vastly more complex but more importantly requires the heap to work

I was about to say I planned on translating [`usb_serial.c`](https://github.com/PaulStoffregen/cores/blob/master/teensy3/usb_serial.c) and that doesn't use a heap... but then I saw [`usb_mem.c`](https://github.com/PaulStoffregen/cores/blob/master/teensy3/usb_mem.c). I have things I can use as serial adapters, I just wanted to avoid the dongle madness that entails. But I think I'll wait.

src/machine/machine_nxpmk66f18.go
firelizzard18: I was translating Paul's code. In general, I'm trying to avoid reimplementing as much as I can. Understanding how to write low-level K66 code is not my goal.

firelizzard18:
Blinky seems to work now with the schedule and normal GC. Though when I call `time.Sleep` instead of my hook directly into `runtime.sleepTicks`, it does not work

firelizzard18:
I'm trying to debug a `nil pointer dereference`. Has TinyGo implemented stack tracing? I can't find it. My other thought was to print the caller's PC to the stack. Where can I look to figure out TinyGo's calling conventions? I assume that's implemented in the compiler somewhere?

firelizzard18:
`currentTask` in `yield()` is nil. What could be causing this? When I add a nil check to the if-statement, I get `fatal error: stack overflow with sp=***`.

https://github.com/tinygo-org/tinygo/blob/c0b8716d7f2cf33bcbb7e452e84260228a4e2b32/src/runtime/scheduler_tasks.go#L100-L107

jaddr2line:
@firelizzard18 you are probably suspending the scheduler, which is not a valid operation

firelizzard18:
## Current Status

* `putchar` works reliably since I switched from calling `machine.UART1.WriteByte` to directly implementing it via registers with a busy wait
* `time.Sleep` works reliably now that I removed `Gosched()` from `sleepTicks`
* `sleepTicks` now uses an actual timer
* I read the ARM documentation on hard faults and added [`fault.go`](https://github.com/tinygo-org/tinygo/blob/4593a297921a7014c9fe6c213b540b4c9a9d6889/src/device/arm/fault.go) and updated [`handleHardFault`](https://github.com/tinygo-org/tinygo/blob/4593a297921a7014c9fe6c213b540b4c9a9d6889/src/runtime/runtime_cortexm.go#L105) with more detailed messages.
* Next up is all the peripherals other than the UARTs

aykevl:
It seems to me like this PR implements enough of the Teensy 3.6 to be ready. However, it does a lot more than that, and quite a bit of that is unrelated to Teensy 3.6 support and should be done in separate PRs (such as the changes to hard fault handling and checking for an invalid yield - but please discuss these changes first in an issue before sending a PR!).

> I was translating Paul's code. In general, I'm trying to avoid reimplementing as much as I can. Understanding how to write low-level K66 code is not my goal.

I completely understand you tried to convert the existing code as-is to get something working. However, such code is often hard to maintain. If possible, please try to reduce the code as much as possible like you did for the UART.

Additionally, can you please rebase on the dev branch? It currently merges in all kinds of changes which makes the history hard to read and the diff larger than it needs to be.

firelizzard18:
> quite a bit of that is unrelated to Teensy 3.6 support and should be done in separate PRs

That was the plan. I was focused on getting everything to work, so I did things to help with debugging. I'll work on removing and/or separating those changes now.

> such as the changes to hard fault handling and checking for an invalid yield - but please discuss these changes first

I implemented this before asking because of the comment below. I ran into a situation where the hard fault handler was reporting a stack overflow when one was definitely not occuring. I saw this comment and I needed to figure out what was actually happening, so I did the TODO. I'll definitely make that a separate PR. I recognize that all of the new error messages will increase binary size, but I think checking the fault status to determine if the stack pointer is meaningful is a must. Otherwise, certain errors are incorrectly reported as a stack overflow. One of those errors is writing to a register of a disabled peripheral, which is definitely something an end-user might do.

https://github.com/tinygo-org/tinygo/blob/1d913a62bc0ceef841953058c2c6fb1dd2a40f6a/src/runtime/runtime_cortexm.go#L106-L109

firelizzard18:
I've rebased on top of the latest `dev` commit. I'll work on cleaning up the history now.

firelizzard18:
I have removed all changes outside of `src` and `targets` except my modifications to `tools/gen-device-svd`. The latter are necessary to process NXP's SVDs - without them it fails to generate anything. I removed changes in `src` that were out of scope or unnecessary.

Currently configuring a pin for input panic's with "todo", but implementing that won't take long. Once that's done I'll mark this ready for review.

tools/gen-device-svd/gen-device-svd.go
```diff
@@ -815,6 +921,28 @@ func writeGoRegisterBitfields(w *bufio.Writer, register *PeripheralField, name s
...
+		w.WriteString(": " + register.description)
+	}
+	w.WriteByte('\n')
+	for _, macro := range register.macros {
+		fmt.Fprintf(w, "func %[1]s(v %[2]s) %[2]s { return (v << %[1]s_Pos) & %[1]s_Msk }", macro.name, intType)
+		if macro.description != "" {
+			w.WriteString(" // " + macro.description)
+		}
+		w.WriteByte('\n')
+	}
```
aykevl: Why are you writing these macros? Is there something that was hard or impossible to do with the existing struct-based peripherals?

targets/teensy36.s
```diff
@@ -0,0 +1,18 @@
...
+    .byte 0xFF
+    .byte 0xFF
+    .byte 0xFF
+    .byte 0xFF
+    .byte 0xFF
+    .byte 0xFF
+    .byte 0xDE
+    .byte 0xF9
+    .byte 0xFF
+    .byte 0xFF
```
aykevl: Nit: this misses a newline at the end of the file.

targets/nxpmk66f18.ld
```diff
@@ -0,0 +1,93 @@
...
+    .text :
+    {
+        /* vector table MUST start at 0x0 */
+		. = 0;
+        _vector_table = .;
+        KEEP(*(.isr_vector))
+
+        /* flash configuration MUST be at 0x400 */
+        . = 0x400;
+        _flash_config = .;
```
aykevl: This defines a global symbol named `_flash_config`. I do not think that's necessary (at least not right now).

targets/nxpmk66f18.ld
```diff
@@ -0,0 +1,93 @@
...
+
+/* define output sections */
+SECTIONS
+{
+    /* Program code and read-only data goes to FLASH_TEXT. */
+    .text :
+    {
+        /* vector table MUST start at 0x0 */
+		. = 0;
+        _vector_table = .;
```
aykevl: This `_vector_table` symbol should not be necessary.

targets/nxpmk66f18.ld
```diff
@@ -0,0 +1,93 @@
...
+_stack_size = 2K;
+
+/* define output sections */
+SECTIONS
+{
+    /* Program code and read-only data goes to FLASH_TEXT. */
+    .text :
+    {
+        /* vector table MUST start at 0x0 */
+		. = 0;
```
aykevl: This file has a mix of spaces and tabs. While I personally prefer spaces (they're always the same width), the important thing is that it is consistent at least per file.

targets/nxpmk66f18.ld
```diff
@@ -0,0 +1,93 @@
...
+        /* flash configuration MUST be at 0x400 */
+        . = 0x400;
+        _flash_config = .;
+		KEEP(*(.flash_config))
+
+        /* everything else */
+        *(.text)
+        *(.text*)
+        *(.rodata)
+        *(.rodata*)
```
aykevl: Small nit: `.text*` also includes `.text`. I believe I introduced this error a while ago in one of the linker scripts and it stuck due to copying around, but it can be fixed here by adding an extra dot (`.text` and `.text.*` etc.).

src/runtime/runtime_nxpmk66f18.go
```diff
@@ -0,0 +1,400 @@
...
+// func abort() {
+// 	for {
+// 		// keep polling some communication while in fault
+// 		// mode, so we don't completely die.
+// 		if nxp.SIM.SCGC4.HasBits(nxp.SIM_SCGC4_USBOTG) usb_isr();
+// 		if nxp.SIM.SCGC4.HasBits(nxp.SIM_SCGC4_UART0) uart0_status_isr();
+// 		if nxp.SIM.SCGC4.HasBits(nxp.SIM_SCGC4_UART1) uart1_status_isr();
+// 		if nxp.SIM.SCGC4.HasBits(nxp.SIM_SCGC4_UART2) uart2_status_isr();
+// 	}
+// }
```
aykevl: This appears to be dead code?

targets/nxpmk66f18.ld
```diff
@@ -0,0 +1,93 @@
...
+    {
+        . = ALIGN(4);
+        . += _stack_size;
+        _stack_top = .;
+    } >RAM
+
+    /* Start address (in flash) of .data, used by startup code. */
+    _sidata = LOADADDR(.data);
+
+	/* todo add .usbdescriptortable .dmabuffers .usbbuffers */
```
aykevl: What's special about these symbols that they need an entry in the linker script?
If it's alignment, that has been handled by the `//go:align` pragma.

targets/nxpmk66f18.ld
```diff
@@ -0,0 +1,93 @@
...
+
+    /DISCARD/ :
+    {
+        *(.ARM.exidx)      /* causes 'no memory region specified' error in lld */
+        *(.ARM.exidx.*)    /* causes spurious 'undefined reference' errors */
+
+        /* all this makes it much harder to debug via disassembly */
+        *(.debug*)
+        *(.ARM.*)
+        *(.comment*)
```
aykevl: Wait - you're discarding all debug info???
The proper way to make disassembly readable is to only disassemble the things you need to disassemble. For example, with `objdump -d`, which dumps only code (as opposed to `objdump -D`).

src/runtime/volatile/register_nxpmk66f18.go
```diff
@@ -0,0 +1,79 @@
...
+
+//go:inline
+func bitbandAddress(reg uintptr, bit uint8) uintptr {
+	if uintptr(bit) > ptrBytes*8 {
+		panic("invalid bit position")
+	}
+	if reg < registerBase || reg >= registerEnd {
+		panic("register is out of range")
+	}
+	return (reg-registerBase)*ptrBytes*8 + uintptr(bit)*ptrBytes + bitbandBase
```
aykevl: Because this feature is specific to the Cortex-M series microcontrollers, you can safely assume that a pointer is 4 bytes in size.

src/runtime/volatile/register_nxpmk66f18.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// Special types that causes loads/stores to be volatile (necessary for
+// memory-mapped registers).
+type BitRegister struct {
+	Reg uint32
```
aykevl: Why `uint32`? As I understand it, these bits are 1 byte in size (thus `uint8`). You could probably even use `bool` directly for a potential efficiency gain but that's not supported right now by the `LoadUint*` and `StoreUint*` family of compiler intrinsics.

src/runtime/volatile/register_nxpmk66f18.go
```diff
@@ -0,0 +1,79 @@
+// +build nxp,mk66f18
+
+package volatile
+
+import "unsafe"
+
+const registerBase = 0x40000000
+const registerEnd = 0x40100000
+const bitbandBase = 0x42000000
```
aykevl: I would argue that only these constants should be in a file specific to this chip.
For example: bitband_nxpmk66f18.go for these constants and bitband.go for the methods below.

src/runtime/runtime_nxpmk66f18.go
```diff
@@ -0,0 +1,400 @@
...
+		micros += cycles / cyclesPerMicro
+	}
+
+	return timeUnit(micros)
+}
+
+//go:export LPTMR0_IRQHandler
+func wake() {
+	nxp.LPTMR0.CSR.Set(nxp.LPTMR0.CSR.Get()&^nxp.LPTMR0_CSR_TEN | nxp.LPTMR0_CSR_TCF) // clear flag and disable
+}
```
aykevl: This should be done using the interrupt API merged a while ago in TinyGo (see #782). That API should be a lot easier to use as well.
This also goes for all other interrupts declared in this PR.

src/runtime/runtime_nxpmk66f18.go
```diff
@@ -0,0 +1,400 @@
...
+	// TODO allow override
+	// > programs using the watchdog timer or needing to initialize hardware as
+	// > early as possible can implement startup_early_hook()
+
+	nxp.WDOG.STCTRLH.Set(nxp.WDOG_STCTRLH_ALLOWUPDATE)
+}
+
+func startupLateHook() {
+	// TODO allow override
+}
```
aykevl: This seems to be a C-ism left over from the code you ported over from the Teensy codebase.
I honestly haven't yet seen a good use case for such things that shouldn't be handled by the runtime. But if one emerges, it should be supported across all boards and not just the Teensy, so should be removed here.

src/runtime/runtime_nxpmk66f18.go
```diff
@@ -0,0 +1,400 @@
...
+
+// ???
+const asyncScheduler = false
+
+// convert from ticks (us) to time.Duration (ns)
+const tickMicros = 1000
+
+var cyclesPerMilli = machine.CPUFrequency() / 1000
+var cyclesPerMicro = machine.CPUFrequency() / 1000000
```
aykevl: Code is easier to optimize (and possibly easier to read as well) when these are declared where they are used.

src/runtime/runtime_nxpmk66f18.go
```diff
@@ -0,0 +1,400 @@
...
+	// 	usb_init();
+	// 	delay(TEENSY_INIT_USB_DELAY_AFTER);
+}
+
+func startupEarlyHook() {
+	// TODO allow override
+	// > programs using the watchdog timer or needing to initialize hardware as
+	// > early as possible can implement startup_early_hook()
+
+	nxp.WDOG.STCTRLH.Set(nxp.WDOG_STCTRLH_ALLOWUPDATE)
```
aykevl: What does this do?
If it enables the watchdog, I would prefer to keep it disabled unless it is actually used.

src/runtime/runtime_nxpmk66f18.go
```diff
@@ -0,0 +1,400 @@
...
+	DEFAULT_FTM_PRESCALE = 1
+
+	DefaultStdoutBaudRate = 115200
+)
+
+var (
+	SIM_SOPT2_IRC48SEL = nxp.SIM_SOPT2_PLLFLLSEL(3)
+	SMC_PMCTRL_HSRUN   = nxp.SMC_PMCTRL_RUNM(3)
+	SMC_PMSTAT_HSRUN   = nxp.SMC_PMSTAT_PMSTAT(0x80)
+)
```
aykevl: All these constants and globals should not be exported.

src/runtime/runtime_nxpmk66f18.go
```diff
@@ -0,0 +1,400 @@
...
+func startupLateHook() {
+	// TODO allow override
+}
+
+func putchar(c byte) {
+	u := &machine.UART1
+
+	// ensure the UART has been configured
+	if !u.SCGC.HasBits(u.SCGCMask) {
+		u.Configure(machine.UARTConfig{BaudRate: DefaultStdoutBaudRate})
```
aykevl: This default baudrate is not necessary as it is already the default in the machine package.

src/machine/uart_nxpmk66f18.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	for !u.TXBuffer.Put(c) {
+		gosched()
+	}
+
+	u.Transmitting.Set(1)
+	u.C2.Set(uartC2TXActive)
+	return nil
+}
```
aykevl, line 305: You could replace this code (and all related code) with the few simple lines of code you have in the runtime `putchar` function. More complex buffers can always be added at a later time (also, you said this code doesn't actually work?).

src/machine/board_teensy36.go
```diff
@@ -0,0 +1,87 @@
...
+
+// ClockFrequency returns the frequency of the external oscillator (16MHz)
+func ClockFrequency() uint32 { return 16000000 }
+
+// LED on the Teensy
+const LED Pin = 13
+
+var pins = []pin{
+	// {bit, control register, gpio register bank}
+	0:  {16, &nxp.PORTB.PCR16, nxp.GPIOB},
```
aykevl: These pins look like Arduino-style pins.
In TinyGo we've solved this differently. The `Pin` values contain the actual pin numbers in whatever form is convenient for the hardware. In this case, it looks like each port has 32 pins so pin 5 maps top port A pin 5, pin 33 to port B pin 1, etc. You can then define pin constants in this board file, such as `machine.D3` for digital pin 3 or however fits with that particular board. That gives readability without sacrificing performance or code density.

src/machine/board_teensy36.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"device/nxp"
+)
+
+// CPUFrequency returns the frequency of the ARM core clock (180MHz)
+func CPUFrequency() uint32 { return 180000000 }
+
+// ClockFrequency returns the frequency of the external oscillator (16MHz)
+func ClockFrequency() uint32 { return 16000000 }
```
aykevl, line 9: We don't use anything like that elsewhere in TinyGo, and don't think it should be added unless there is a real need for it (to avoid complicating the API).

targets/nxpmk66f18.ld
firelizzard18: When debugging, I wanted to see <_flash_config> in objdump instead of <some_function-some_offset>

targets/teensy36.s
firelizzard18: Fixed

targets/nxpmk66f18.ld
firelizzard18: Same reason as _flash_config, this was for debugging

targets/nxpmk66f18.ld
firelizzard18: Fixed (with spaces)

targets/nxpmk66f18.ld
firelizzard18: I'm not sure I follow. Are you suggesting `*(.text)` and `*(.text.*)` or `*(.text*)` and `*(.text.*)`? Or something else? I copied and pasted from one of the existing targets and/or the Teensy core's script.

src/runtime/runtime_nxpmk66f18.go
firelizzard18: I made `abort` overridable, so this was the Teensy's custom abort function. However, given the real reason for overriding the abort function is supporting seamless reprogramming, and that requires USB, I decided those changes could wait until a PR implementing USB. But I forgot to remove this function.

src/runtime/volatile/register_nxpmk66f18.go
firelizzard18: > Because this feature is specific to the Cortex-M series microcontrollers

Do you have suggestions on making this file usable for other Cortex-M's and not just the K66? I'm guessing they all do essentially the same math, but `registerBase`, `registerEnd`, and `bitbandBase` seem like they could be processor-specific.

I could change the build tags to `cortexm,bitband` so that targets could opt in, and then require them to define those constants. Except they probably should be defined in `runtime`, which means they would need to be exported.

src/runtime/volatile/register_nxpmk66f18.go
firelizzard18: Because the K66 docs describe these as 32-bit registers. Endianness is something I try to avoid thinking about. Would treating an address of a 32-bit register as an 8-bit value safely allow setting the low bit?

src/runtime/volatile/register_nxpmk66f18.go
firelizzard18: Your comments on #842 mentioned wanting to limit the number of places developers are required to make changes. Defining these in `runtime/volatile` would add a place. Of course, if `bitband.go` used the build tags `cortexm,bitband`, then targets could opt in. I don't think the constants could be defined in `machine`, `runtime`, or even `device/*` because it would cause a circular reference.

src/runtime/runtime_nxpmk66f18.go
firelizzard18: I'll remove it for now. The late hook could be implemented similarly to #842. That would not work for the early hook.

Based on what I've read in the PJRC forums, it seems like some workflows, peripheral initialization, and startup tasks can require the running code before all the initialization happens.

src/runtime/runtime_nxpmk66f18.go
firelizzard18: Fixed

src/runtime/runtime_nxpmk66f18.go
firelizzard18: Fixed

src/machine/uart_nxpmk66f18.go
firelizzard18: This code works now. I never really understood what Paul's code was doing exactly, so I took the general idea and reimplemented it.

src/machine/board_teensy36.go
firelizzard18: `sleepTicks` configures the low power timer to use the external reference clock for timing. This means that my math is wrong, because `cyclesPerMilli` and `cyclesPerMicro` are calculated from the CPU clock. `sleepTicks` should be using values calculated from `ClockFrequency()` instead of `CPUFrequency()`. The low power timer does have the option of using the 'internal reference clock', MCGIRCLK, but I'm not sure what that is.

src/runtime/runtime_nxpmk66f18.go
firelizzard18: "Enables updates to watchdog write-once registers, after the reset-triggered initial configuration window (WCT) closes, through unlock sequence." `WDOG_STCTRLH_WDOGEN` would be required to actually enable the watchdog.

src/runtime/runtime_nxpmk66f18.go
firelizzard18: I wanted to define them once since they are used in multiple places. Once I fix the sleepWait calculations, `cyclesPerMicro` will only be used in one place. However `cyclesPerMilli` is used as the systick reset value and thus is needed when calculating ticks. I thought using the same variable would clarify the relationship between those two.

src/runtime/runtime_nxpmk66f18.go
firelizzard18: I had some weird issues using that API so I used the old system to get everything working

targets/nxpmk66f18.ld
firelizzard18: I needed to see everything in the disassembly, for one because I was having issues getting the flash config to work. And given that I was experiencing weird issues that seemed totally inexplicably, I wanted to look through absolutely everything. At one point I was comparing assembly line by line.

targets/nxpmk66f18.ld
firelizzard18: It's probably alignment. I wanted to leave myself a reminder of what sections Teensy's LD script handled specially. I've reworded the comment to reflect that.

tools/gen-device-svd/gen-device-svd.go
firelizzard18: The macros encapsulate the `(x & FIELD_MASK) << FIELD_POS` pattern. They have no bearing on using or not using struct-based peripherals. Initially I created them to minimize the difference between Teensy's C code and my ported Go code. Now, I think they improve readability.

With macros:

```golang
nxp.SystemControl.SHPR3.Set(nxp.SystemControl_SHPR3_PRI_15(32) | nxp.SystemControl_SHPR3_PRI_14(32))
```

Without:

```golang
nxp.SystemControl.SHPR3.Set(((32 << SystemControl_SHPR3_PRI_15_Pos) & SystemControl_SHPR3_PRI_15_Msk) | ((32 << SystemControl_SHPR3_PRI_14_Pos) & SystemControl_SHPR3_PRI_14_Msk))
```

src/machine/board_teensy36.go
firelizzard18: With the way I've defined the pins, if the pin number is constant, everything optimizes to direct register writes. I've rearranged an instruction, removed a nop, and replaced movt/movw pairs with a single mov, but this is the result of a simple blink program:

```asm
    ; set bit 5 of PORTC PDDR
    mov r4, #0x400FF064
    ldr r0, [r4, #16]
    orr.w r0, r0, #0x20
    str r0, [r4, #16]

    ; configure PORTC PCR5
    mov r0, #0x4004B014
    mov.w r1, #0x144
    str r1, [r0, #0]

    ; set bit 5
    movs r5, #0x20
l:
    str r5, [r4, #0] ; of PORTC PSOR
    bl <time.Sleep>
    str r5, [r4, #4] ; of PORTC PCOR
    bl <time.Sleep>
    b.n l
```

Originally:

```golang

func main() {

	for {
		machine.LED.Set(true)
		time.Sleep(time.Second)

		machine.LED.Set(false)
		time.Sleep(time.Second)
	}
}
```

Based on exhaustively scanning through the K66 manual's port mux table, I can say that ports have the following pin ranges:

  * A: 0-29
  * B: 0-11, 16-23
  * C: 0-19
  * D: 0-15
  * E: 0-28

Architecturally, each port might be able to have 32 pins, but that's certainly not how it's actually arranged.

tools/gen-device-svd/gen-device-svd.go
aykevl: I would write that like this:

```
nxp.SystemControl.SHPR3.Set((32 << SystemControl_SHPR3_PRI_15_Pos) | (32 << SystemControl_SHPR3_PRI_14_Pos))
```

It is a few characters longer than the macro-based system (which actually aren't even macros but functions that hopefully get inlined) but it fits with how everything else is written so far. Of course, this can be changed in TinyGo but introducing another way to set constants will only increase the amount someone needs to learn to work with the system. In my view, a PR adding a new board is not the right place to introduce such a thing.

targets/nxpmk66f18.ld
aykevl: I think you can solve that by modifying the assembly:

```asm
.section .flash_config
.global  __flash_config
__flash_config:
```

targets/nxpmk66f18.ld
aykevl: Sorry wasn't very clear. This PR to fix it in arm.ld should make it clear: https://github.com/tinygo-org/tinygo/pull/937

> Are you suggesting `*(.text)` and `*(.text.*)` or `*(.text*)` and `*(.text.*)`?

The former.

targets/nxpmk66f18.ld
aykevl: :+1:

targets/nxpmk66f18.ld
aykevl: For the future: you can do that with the `-no-debug` flag to TinyGo.

src/runtime/volatile/register_nxpmk66f18.go
aykevl: Right. It's probably more trouble right now than it's worth so let's leave it as-is.

src/runtime/volatile/register_nxpmk66f18.go
aykevl: From what I've read, it seems like the width is always the same (they map a 1MB address space to a 32MB bitband region) but the exact addresses seem to differ.

src/runtime/volatile/register_nxpmk66f18.go
aykevl: Nevermind, it looks like I was wrong. Bitbanding seems to use 32-bit words instead of bytes.
For what it's worth, I've never seen a big endian Cortex-M chip. In general, big endian systems seem to almost disappear.

src/runtime/runtime_nxpmk66f18.go
aykevl: Can you try switching over now? You can do that one by one. If there is anything that doesn't quite fit or doesn't work, let me know.

src/machine/board_teensy36.go
aykevl: I'm not sure I follow. Are you saying there is a bug right now? Because `ClockFrequency` doesn't seem to be used in this PR.

src/machine/board_teensy36.go
aykevl: > With the way I've defined the pins, if the pin number is constant, everything optimizes to direct register writes.

That is likely only true as long as you're only setting one pin: the compiler will inline everything and/or const-propagate the pin. Otherwise the array will likely _not_ be optimized away.

It's fine if there are gaps. You could do this for example:

```go
const (
    PA00 Pin = 0
    PA01 Pin = 1
    ...
    PA29 Pin = 29
    PB00 Pin = 32
    ...
    PB11 Pin = 43
    PB16 Pin = 48
    ...
)
```

You can then use something like this:

```go
func (p Pin) getPortPin() (*nxp.GPIO_Type, uint32) {
    if p < 32 {
        return nxp.GPIOA, uint32(p)
    } else if p < 64 {
        return nxp.GPIOB, uint32(p) - 32
    } // etc
}

// Set the pin to high or low.    
// Warning: only use this on an output pin!    
func (p Pin) Set(high bool) {    
    port, bit := p.getPortPin()    
    if high {    
        port.PSOR.Set(1 << bit)    
    } else {    
        port.PCOR.Set(1 << bit)    
    }    
}    
```

This is also done for other chips, see for example:
https://github.com/tinygo-org/tinygo/blob/master/src/machine/machine_stm32f103xx.go#L35
https://github.com/tinygo-org/tinygo/blob/master/src/machine/machine_stm32f103xx.go#L94

Doing it this way makes sure that such low-level code only needs to be written once. A board using the same chip (or a different chip in the same family) can simply define a few constants without needing to worry about what `nxp.GPIOB` means. There are more advantages: with this separation I was able to simulate some boards on https://play.tinygo.org using the same pin mapping as the boards use on actual hardware.

There is one issue remaining, which is stuff like `&nxp.PORTB.PCR17`. Microchip has a similar issue with their SAM D chips. It's ugly, but I think you can work around it with some `unsafe.Pointer` hacking:

    port := ... // for example nxp.PORTB
    reg := (*volatile.Register32)(unsafe.Pointer(uintptr(unsafe.Pointer(port)) + pin * 4))

targets/nxpmk66f18.ld
firelizzard18: Fixed. I did that for anything matching `*(name*)`, including bss and data.

src/machine/board_teensy36.go
firelizzard18: Yes and no. The way I calculate the countdown for the timer is incorrect and results in sleeps 180MHz/16MHz = 11 times shorter than they should be. However, I use `ticks()` to determine when I'm done sleeping, so `sleepTicks` still blocks for the correct duration.

Given that the frequency of the external clock is board-specific, I put it in the board file, so other boards would be able to override it.

BTW I said `sleepWait` in my previous comment when I meant `sleepTicks`. I've edited it.

firelizzard18: To directly answer, yes there is something that I implemented incorrectly, and to fix it I need to know the frequency of the external oscillator. Since that is board-specific, I didn't want to hard code it. Do you have a suggestion for resolving this? I could use an entirely different timer, but the low power timer seemed like the best option for scheduler sleeps.

src/runtime/runtime_nxpmk66f18.go
firelizzard18: I've got it working now

src/machine/board_teensy36.go
firelizzard18: > 128 (untyped int constant) overflows int8

`Pin` is defined as `int8` in `machine.go`. While the K66 has less than 128 usable pins, if I use your suggested mechanism, I can't use port E without changing the definition of `Pin`.

firelizzard18: I implemented the change, but it required changing `Pin` to `int16`

deadprogram:
Seems like there are some merge conflicts needing to be resolved here.

deadprogram:
Any action on this PR? It needs merge conflicts resolved, please.

firelizzard18:
Are you referring to the open discussions? I am waiting for responses to my replies.

deezone:
@firelizzard18 The "merge conflicts" being referenced in the comments that need to be resolved are:
image

andrewrynhard:
Any plans in getting this info the next release of tinygo? Just bought a teensy 3.6 in preparation for this.

aykevl:
Sorry I've left this PR hang for so long, I was busy with other things. I hope to get back to it soon.

deadprogram:
I have a Teensy 3.6 coming in next week, so hopefully I can help get this PR landed.

src/machine/board_teensy36.go
aykevl: That may have an effect on many other targets. I think it's best to switch the `machine.Pin` type to `uint8` (and `machine.NoPin` to 0xff) which should also fix issues for a few other chips (like the stm32f7 family that often have a huge number of pins). This should be done in a separate PR, however.

If `uint8` is not enough, the type should instead be chip family specific instead of changing to `int16` for all chips.

firelizzard18: > This should be done in a separate PR, however.

So this PR should leave out port E for now? Or are you suggesting a separate PR as a prerequisite to this one?

> the type should instead be chip family specific

This seems like the cleanest solution to me. It looks like most low-level types are chip-specific, so it makes a lot of sense to me for the pin type to be chip-specific.

deadprogram: >So this PR should leave out port E for now? 

That seems like the simplest option to me.

aykevl: > > This should be done in a separate PR, however.
> 
> So this PR should leave out port E for now? Or are you suggesting a separate PR as a prerequisite to this one?

I intended it as a prerequisite for this PR, but you can also leave out port E for now. There are a few other targets that leave out some pins because of this.

firelizzard18: I opened #1190, removed port E, and added comment references to the issue.

deadprogram:
Hi @firelizzard18 can you please rebase against the `dev` branch again? I think there are some changes there for CircleCI that might fix your build error. Thanks!

deadprogram:
Hi @firelizzard18 looks like that last commit removed something that was needed, as the tests are failing:

```
tinygo build -size short -o test.hex -target=circuitplay-express examples/i2s
# machine
/home/circleci/lib/tinygo/src/machine/usb_setup.go:54:69: undeclared name: usb_BCD_DEVICE
/home/circleci/lib/tinygo/src/machine/usb_setup.go:6:2: "errors" imported but not used
```

firelizzard18:
@deadprogram I've been working on USB support, but I've hit a wall. Unless someone can help me with that, I'll probably just remove the last commit (which should fix CI).

1. The host (Windows) sends a reset
1. The host requests the device descriptor
1. The ISR transmits the device descriptor
1. The host sends a reset
1. Occasionally, the host sends SET_ADDRESS
1. And then nothing (Windows reports "device descriptor request failed")

My logging shows the same sequence of events happens with the C implementation, but it continues past that point (and works). Without a USB protocol analyzer, I'm not sure what to do.

aykevl:
I have heard Wireshark can intercept USB traffic, you could try that.

firelizzard18:
When running the C implementation, I can capture the control traffic I expect. When running the Go implementation, I don't see anything that I can ascribe to the Teensy. I don't completely know what I'm doing so maybe I'm filtering incorrectly. Or maybe Windows hides traffic from devices it can't properly enumerate.

deadprogram:
@firelizzard18 are you testing with the last commit in this branch? Asking in case I can grab some time to look into it myself, I have spent a fair amount of time by now looking at Wireshark USB dumps.

aykevl:
Alternatively I think the USB support can also be added at a later time.

deadprogram:
Actually, looking more closely, I think you should avoid any changes all all to the existing USB code at this time. There are changes made in this PR that would be much better made as part of some other PR specific to refactoring USB code, and also I am pretty sure that you could be using some of the shared USB code instead of code that you have added.

So that means just removing the last commit, it would appear. As @aykevl said we can take up getting the USB CDC working in a subsequent PR.

I will wire up my setup for testing my Teensy 3.6 now.

firelizzard18:
@deadprogram I pushed some very minor modifications. Mostly I added `arm.Asm("dsb 0xF")` after looking through Freescale's SDK. I pushed the USB changes to [`teensy-3.6-usb`](https://github.com/firelizzard18/tinygo/tree/teensy-3.6-usb)

> some other PR specific to refactoring USB code

I didn't refactor the USB code so much as rearrange it so I could selectively exclude `sendDescriptor` and `sendConfiguration`.

> you could be using some of the shared USB code instead of code that you have added

I could and planned to use the descriptor structs and constructors more. But for the initial implementation, I wanted to keep everything as close to Paul's implementation as I could. I planned to modify `getUSBDescriptor` to use those constructors instead of a byte array, once CDC worked.

`sendConfiguration` and therefore `sendDescriptor` does not allow any customization of the USB configuration. The USB support has to behave in a certain way in order to fully support the Teensy programmer. I don't know enough about it to know if full programmer support requires a custom USB config descriptor. Regardless, the Teensy can support many different USB functions, such as mouse, keyboard, joystick, MIDI, audio, etc. Supporting anything beyond CDC would require customization of the config descriptor.

> I have spent a fair amount of time by now looking at Wireshark USB dumps

Would you like me to email you some Windows/USBPcap dumps? Wireshark understands *.pcap files. Since I don't know how to isolate the Teensy, this would include all of my USB traffic (because basically everything in my system is on a single root hub), so I am unwilling to post them publicly.

firelizzard18:
FYI, all hell broke loose when I attempted to use UART0 directly from within the USB ISR. Hence why I write debug messages to a global variable.

deadprogram:
@firelizzard18 looks like you need to run `go fmt` on `src/machine/board_teensy36.go` please.

deadprogram:
I just tried this branch with the latest `dev` branch, and got the following:

```
$ tinygo flash -target teensy36 examples/blinky1
# runtime
/home/ron/.gvm/pkgsets/go1.14.2/global/src/github.com/tinygo-org/tinygo/src/runtime/runtime.go:64:9: undeclared name: ticksToNanoseconds
/home/ron/.gvm/pkgsets/go1.14.2/global/src/github.com/tinygo-org/tinygo/src/runtime/scheduler_any.go:10:31: undeclared name: nanosecondsToTicks
```

firelizzard18:
@deadprogram The build failure is fixed now. You might not be able to use `tinygo flash`. The Teensy CLI loader didn't work for me. I am using the GUI tool.

deadprogram:
Using this branch as of the last commit, I have blinky1 success!

```
$ tinygo flash -target teensy36 examples/blinky1
Teensy Loader, Command Line, Version 2.1
Read "/tmp/tinygo278174858/main.hex": 7188 bytes, 0.7% usage
Waiting for Teensy device...
 (hint: press the reset button)
Found HalfKay Bootloader
Read "/tmp/tinygo278174858/main.hex": 7188 bytes, 0.7% usage
Programming........
Booting
```

I did have to press the reset button when flashing, but then it worked.

deadprogram:
In further testing, I am able to run this branch against the TinyHCI and all of the current hardware passes.

This PR certainly seems functional enough to merge to keep working on additional features.

@aykevl or anyone else have any further comments here?

src/machine/machine_nxpmk66f18.go
```diff
@@ -0,0 +1,300 @@
...
+	var pcr *nxp.PORT_Type
+
+	if p < 32 {
+		gpio, pcr = nxp.GPIOA, nxp.PORTA
+	} else if p < 64 {
+		gpio, pcr = nxp.GPIOB, nxp.PORTB
+	} else if p < 96 {
+		gpio, pcr = nxp.GPIOC, nxp.PORTC
+	} else {
+		// } else if p < 128 {
```
deadprogram: Remove comments?

deadprogram:
@jaddr2line can you also please lookover this PR? Also anyone else who is interested. Thank you!

deadprogram:
Sorry @firelizzard18 but just one last merge conflict (hopefully!) to resolve, if you please.

src/machine/machine_nxpmk66f18.go
firelizzard18: Now that `machine.Pin` is a `uint8`, I dropped the commit that commented out that code

firelizzard18:
@deadprogram Merge conflicts resolved

deadprogram:
At long last, time to merge this pull request. I know there are still things to be done for this board, but it will be a lot easier to work on those in smaller, more specific PRs.

Thank you very much @firelizzard18 for all of your work on this contribution, and thanks everyone else who helped get it done.

Now rebasing/merging.

Issue #154, opened 2019-01-24
strings.IndexByte not yet implemented


deadprogram:
When trying to call `strings.Contains()` I get the following message:

```shell
/tmp/tinygo124212217/main.o: In function `strings.Index':
../../../../../../../gos/go1.11/src/strings/strings.go:995: undefined reference to `strings.IndexByte'
```

Does this mean that the `strings.IndexByte()` function is not yet implemented in TinyGo? If so, then where exactly should it be added?

aykevl:
The source says this:

```go
// IndexByte returns the index of the first instance of c in s, or -1 if c is not present in s.
func IndexByte(s string, c byte) int // in internal/bytealg
```

So it looks like it is implemented in the runtime, probably with hand-optimized assembly.
You can implement it in TinyGo in src/runtime/strings.go this way:

```go
//go:linkname indexByte strings.IndexByte
func indexByte(...) { ...}
```

For an initial implementation I would avoid optimizing it, just using the most obvious/naive implementation.

PR #867, opened 2020-01-23
Added indexBytePortable from standard library to link as implementation for internal/bytealg.IndexByte

bgould:
This PR copies the indexBytePortable function from the standard library to fix a linker error similar to the one in https://github.com/tinygo-org/tinygo/issues/154

### DIFF
--- a/src/runtime/bytes.go
+++ b/src/runtime/bytes.go
@@ -1,0 +1,11 @@
+package runtime
+
+//go:linkname indexBytePortable internal/bytealg.IndexByte
+func indexBytePortable(s []byte, c byte) int {
+	for i, b := range s {
+		if b == c {
+			return i
+		}
+	}
+	return -1
+}


deadprogram:
Which version of Go were you using when you hit that erro, @bgould ?

bgould:
I was using go version go1.13.3 linux/amd64

FWIW I encountered this when trying to compile https://golang.org/src/net/http/internal/chunked.go

deadprogram:
@aykevl any reason to not merge this?

aykevl:
To be sure, I checked when this function got added and it was [in Go 1.11](https://github.com/golang/go/commit/403ab0f2214f583db84a2dae275389be92072a35#diff-aee634d566bce3ce6544e16e42f8815a) which is our first supported version, so no build tags should be needed. This looks good to me.

deadprogram:
OK great. Thanks! Now merging.

PR #869, opened 2020-01-24
avr: support `tinygo run` with simavr

aykevl:
This commit adds support for running both the Arduino Uno and the
DigiSpark under a simulator.

Running `tinygo gdb` under a simulator is different from (for example)
what `tinygo gdb` does for other boards such as the PCA10040. However, I
think this is the better option as probably very few people actually
have a hardware debugger and supporting GDB with the simulator is then
much more useful.

### DIFF
--- a/targets/atmega328p.json
+++ b/targets/atmega328p.json
@@ -10,5 +10,6 @@
 	"extra-files": [
 		"targets/avr.S",
 		"src/device/avr/atmega328p.s"
-	]
+	],
+	"emulator": ["simavr", "-m", "atmega328p", "-f", "16000000"]
 }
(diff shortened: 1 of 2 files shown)


deadprogram:
@aykevl this branch now has merge conflicts.

aykevl:
Ah, will fix that.

However, since avr-gdb seems to be rather broken (at least when combined with TinyGo produced DWARF), maybe it's better to not include `tinygo gdb` support for now? What do you think?

deadprogram:
Hmm, yes probably better to hold off then. Can mark this PR as draft I suppose.

aykevl:
Rebased.
The second commit has now become unrelated, but it should be useful anyway.

aykevl:
Updated yet again to split out the error checking into a separate PR: #915.

PR #870, opened 2020-01-25
ci: fix Windows build failure for wasi-libc

aykevl:
The wasi-libc Makefile uses the `find` command line tool. Unfortunately, it was using the Windows find version instead of the MinGW version, leading to lots of errors at a later stage.

This commit prepends /usr/bin to `$PATH` to make sure the MinGW version is found first.


aykevl:
This is now ready, if the CI passes this time.

deadprogram:
OK great. That Windows is so very picky about paths.

Now merging.

PR #873, opened 2020-01-25
wasm: use wasi ABI for basic startup/stdout

aykevl:
This allows TinyGo-built binaries to run under wasmtime, for example:

    tinygo build -o test.wasm -no-debug -target=wasm examples/test
    wasmtime run test.wasm 0

Of course this doesn't bring full wasi compliance, but it's a start.

### DIFF
--- a/src/runtime/runtime_wasm.go
+++ b/src/runtime/runtime_wasm.go
@@ -2,35 +2,43 @@
 
 package runtime
 
+import "unsafe"
+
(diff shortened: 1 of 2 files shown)


deadprogram:
Can this be merged now? Who has tested this?

aykevl:
It is tested with automatic tests. Those should ensure that the changes to wasm_exec.js are correct - otherwise the test output wouldn't be what it should be.

deadprogram:
Now merging. Thank you!

PR #874, opened 2020-01-25
avr: use a garbage collector

aykevl:
This might sound crazy, but I think it's better to enable the GC by default to avoid surprises. It costs 1130 bytes of flash and 16 bytes of RAM (plus heap overhead) so it's not exactly free, but if needed it can easily be disabled with `-gc=leaking`. On the Uno (32kB flash, 2kB RAM) that's not massive, on the DigiSpark (8kB flash, 0.5kB RAM) that may be too much depending on the application.

Also added a smoke test for the `-gc=leaking` case: "if it isn't tested, it's broken".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -267,6 +267,8 @@ ifneq ($(AVR), 0)
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=digispark           examples/blinky1
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=digispark -gc=leaking examples/blinky1
+	@$(MD5SUM) test.hex
 endif
 	$(TINYGO) build -size short -o test.hex -target=hifive1b            examples/blinky1
 	@$(MD5SUM) test.hex
--- a/src/runtime/gc_conservative.go
+++ b/src/runtime/gc_conservative.go
@@ -318,7 +318,7 @@ func markRoots(start, end uintptr) {
 		}
 	}
 
-	for addr := start; addr != end; addr += unsafe.Sizeof(addr) {
+	for addr := start; addr != end; addr += unsafe.Alignof(addr) {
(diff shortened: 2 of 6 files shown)


deadprogram:
I have concentrated my testing on Digispark and my WS2812 branch.

Here are my stats:

```
$ tinygo flash -size short -target digispark -gc=conservative ./examples/ws2812/main.go 
    738       0     149 |     738     149
```

```
$ tinygo flash -size short -target digispark -gc=leaking ./examples/ws2812/main.go 
    654       0     149 |     654     149
```

Note that both of these programs execute as expected on the MCU. I do not see any difference other than the file size.

aykevl:
The real difference is of course for cases where you would otherwise see `runtime panic: out of memory`. I have tested this myself with:

    tinygo flash -target=arduino ./testdata/gc.go

(You can enable `gcDebug` in gc_conservative.go to see what the GC is doing).

deadprogram:
Also it seems like most of the code is optimized away when not needed.

Since the `blinky1` does not perform any real allocation, the overhead appears to be 200 bytes on arduino.:

```
$ tinygo flash -size short -gc=leaking -target arduino examples/blinky1
    490       0     658 |     490     658
```

```
$ tinygo flash -size short -gc=conservative -target arduino examples/blinky1
    690      54     658 |     744     712
```

One the other hand, the `echo` example 

```
$ tinygo flash -size short -gc=leaking -target arduino examples/echo
   1214     229     642 |    1443     871
```

```
$ tinygo flash -size short -gc=conservative -target arduino examples/echo
   2350     229     658 |    2579     887
```

In this case, we see the full 1136 bytes of overhead. So it is significant, but perhaps not really considering what you get. Also, since this can be turned off, it allows for specific needs from the developer. 

So pretty much I am satisfied that this can be merged, unless someone else would like to express a different opinion.

deadprogram:
OK, now merging. :wastebasket:

Issue #777, opened 2019-12-08
Expand board and device support for stm32f4discovery and peripherals


gwtnz:
The stm32f4discovery board has a number of devices on it (MEMS microphone, I2S sound device, accelerometer, etc) that currently don't have support in tinygo, nor is there support for:
* SPI
* I2C
* PWM
* I2S
on the chip.

The machine_stm32f407.go UART peripheral code style differs in from that in the machine_stm32f103xx.go file, and while perhaps being closer to a C-developer's natural low-level usage of the chip, programmatically seems like a more complex implementation making expansion for the other peripherals less straightfoward.

I propose combining the two styles and moving them up to the machine_stm32.go file and then handling chip- and board-specific nuances in the chip family files.

deadprogram:
It would be great to share code and make the implementations more consistent. I am in favor of this proposal.

gwtnz:
I'm happy to take a shot at a PR

deadprogram:
This has now been released, so closing. Thanks!

PR #875, opened 2020-01-26
Expand stm32f4 support

gwtnz:
Refer to Issue #777:

Attempts to leverage as much as possible from the stm32f1xx implementation for setting up GPIO, UARTs, SPI, I2C (not fully fleshed out yet) to expand support on the stm32f4xx boards and eventually other families in the stm32 line.
This moves the overall calls for creating a UART, SPI or I2C (and eventually other features like PWM, I2S, USB) into `machine_stm32_<peripheral>.go` with specific per-chip family calls in the lower level `machine_stm32fXXX.go` files for that chip.
Unfortunately there's not as much sharing as I'd hoped between the stm32f103 and the stm32f4xx (and other families), specifically around the registers such as MODER.

I lack stm32f103 boards to test this on, so have been relying on compilation only for those.


aykevl:
Can you perhaps rebase on top of the dev branch? That makes the diff much easier to read (and will be required before merging anyway).

Also, if possible, it would be much easier to review if you could split up the PR into different atomic commits (and possibly separate PRs). For example, one commit for SPI support, one for UART support, one for I2C support. What also helps is to separate pure refactors (moving code) and the actual change (such as implementing a feature) into separate commits. Small PRs are much easier and thus faster to review and merge, thus making it possible to build on top of it quickly.

> I lack stm32f103 boards to test this on, so have been relying on compilation only for those.

I have one so could test that.

gwtnz:
Ok I'll give the rebase and splitting up a shot.

gwtnz:
New PR here, initial step : https://github.com/tinygo-org/tinygo/pull/880

PR #876, opened 2020-01-26
compiler: add debug info for local variables

aykevl:
Local variables aren't scoped yet and not all are included, but this is a start. To use it, run `info locals` in GDB.

WIP: this depends on https://github.com/golang/tools/pull/201.


aykevl:
Now that https://github.com/golang/tools/pull/201 has been merged, this PR is ready for review.

deadprogram:
We certainly need the debug features. Now merging.

PR #605, opened 2019-09-30
arduino: fix avr hex not working when flashed

parasquid:
An optimization introduced in

https://github.com/tinygo-org/tinygo/commit/a04db67ea9d882eed9164321bcb503f76a65d2f1

seems to have broken arduino uno compiled hex. Setting optimzation
flags to 1, 2, or s builds proper hex binaries though.

These patches have been the result of troubleshooting over slack:

> @aykevl 
> that preinit also doesn't look right. Can you try this variant, with 8-bit stores instead of 32-bit stores?
> There might be some alignment issue: the _ebss might not be aligned resulting in ptr != unsafe.Pointer(&_ebss) never being true.

Co-authored-by: Ayke van Laethem <aykevanlaethem@gmail.com>
Co-authored-by: Jaden Weiss <jadr2ddude@gmail.com>

PR #878, opened 2020-01-26
runtime: fix external address declarations

aykevl:
This is the same problem as in https://github.com/tinygo-org/tinygo/pull/605, but other targets also suffer from it.

Discovered with the GBA target, but as pointed out in https://bugs.llvm.org/show_bug.cgi?id=42881#c1 this appears to be a bug in the way external globals are declared, not in LLVM. Therefore I decided that fixing it everywhere would be the best thing to do.


deadprogram:
Tested on various platforms and worked as expected.

Now merging.

PR #880, opened 2020-01-27
refactor GPIO for stm32f4, in prep for adding SPI, I2C, PWM etc

gwtnz:
See https://github.com/tinygo-org/tinygo/issues/777 for background.

Initial refactor of STM32Fxxx to move common functionality into the parent `machine_stm32.go` file.

The STM32F1xx series has a slightly different method of setting up GPIO from the other STM32Fxxx families, combing clock speed and pin function together. This PR moves GPIO handling to a common file for families that use the MODER, OSPEED and PUPDR registers, and puts high level setup calls in the overall `machine_stm32.go` file so that a common setup procedure can be used across families, with customizations/variations in their specific family or chip-level files.


src/machine/machine_stm32.go
```diff
@@ -2,6 +2,102 @@
...
+	}
+}
+
+// Get returns the current value of a GPIO pin.
+func (p Pin) Get() bool {
+	port := p.getPort()
+	pin := uint8(p) % 16
+	val := port.IDR.Get() & (1 << pin)
+	return (val > 0)
+}
```
gwtnz: Common GPIO calls

src/machine/machine_stm32_moder_gpio.go
```diff
@@ -0,0 +1,142 @@
+// +build stm32,!stm32f103xx
```
gwtnz: Much of this file came from the old `machine_stm32f407.go` file

src/machine/machine_stm32f407.go
```diff
@@ -171,20 +129,24 @@ func (uart UART) Configure(config UARTConfig) {
...
-		// use standard TX/RX pins PA2 and PA3
-		UART_TX_PIN.Configure(PinConfig{Mode: PinModeUartTX})
-		UART_RX_PIN.Configure(PinConfig{Mode: PinModeUartRX})
+	// Set the GPIO pins to defaults if they're not set
+	if config.TX == 0 {
+		config.TX = UART_TX_PIN
+	}
+	if config.RX == 0 {
+		config.RX = UART_RX_PIN
 	}
```
gwtnz: Setting defaults for TX and RX pins if not passed in on the `config` object. This is to handle the chip initialization in `runtime/runtime_stm32f407.go`

src/machine/machine_stm32_moder_gpio.go
```diff
@@ -0,0 +1,142 @@
...
+		p.SetAltFunc(AF5_SPI1_SPI2)
+	}
+}
+
+// SetAltFunc maps the given alternative function to the I/O pin
+func (p Pin) SetAltFunc(af AltFunc) {
+	port := p.getPort()
+	pin := uint8(p) % 16
+	pos := pin * 4
+	if pin >= 8 {
```
thinxer: pos -= 8

src/machine/machine_stm32f103xx.go
```diff
@@ -32,6 +33,20 @@ const (
...
+// Configure this pin with the given I/O settings.
+// stm32f1xx uses different technique for setting the GPIO pins than the stm32f407
+func (p Pin) configure(config PinConfig) {
+	// Configure the GPIO pin.
+	port := p.getPort()
+	pin := uint8(p) % 16
+	pos := uint8(p) % 8 * 4
+	if pin < 8 {
+		port.CRL.SetMasked(uint32(config.Mode), 0xf, pos)
+	} else {
```
thinxer: pos -= 8

src/machine/machine_stm32_moder_gpio.go
```diff
@@ -0,0 +1,142 @@
...
+	// Mode Flag
+	PinOutput        PinMode = 0
+	PinInput         PinMode = PinInputFloating
+	PinInputFloating PinMode = 1
+	PinInputPulldown PinMode = 2
+	PinInputPullup   PinMode = 3
+
+	// for UART
+	PinModeUartTX PinMode = 4
+	PinModeUartRX PinMode = 5
```
aykevl: The [usual Go convention](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) is to keep abbreviations in uppercase:

> Words in names that are initialisms or acronyms (e.g. "URL" or "NATO") have a consistent case. For example, "URL" should appear as "URL" or "url" (as in "urlPony", or "URLPony"), never as "Url". As an example: ServeHTTP not ServeHttp. For identifiers with multiple initialized "words", use for example "xmlHTTPRequest" or "XMLHTTPRequest".

However, in this case I'm not so sure as multiple abbreviations follow each other. [More discussion here](https://stackoverflow.com/a/39830270/559350). @deadprogram any opinion on this?

src/machine/machine_stm32.go
```diff
@@ -2,6 +2,102 @@
...
+//  5. If desired enable the nested vector interrupt control to generate interrupts.
+//  6. Program the AF/peripheral for the required configuration (eg baud rate for a USART) .
+
+// Given that the stm32 family has the AF and GPIO on different registers based on the chip,
+//  use the main function here for configuring, and use hooks in the more specific chip
+//  definition files
+// Also, the stm32f1xx series handles things differently from the stm32f0/2/3/4
+
+// EnableClock configures the STM32 to send clock signals to the given GPIO port
+func (p Pin) EnableClock() {
```
aykevl: Why is this exported? It's already done in `Configure`, so there's no need to repeat it anywhere.

src/machine/machine_stm32.go
```diff
@@ -2,6 +2,102 @@
...
+// configure sets the port I/O direction, speed, pull-up/down config etc
+func (p Pin) configure(config PinConfig) {
+	// perform GPIO configuration specific to this family
+}
+*/
+// ---------- General pin operations ----------
+
+// Configure this pin with the given I/O settings.
+func (p Pin) Configure(config PinConfig) {
+	p.EnableClock()
```
aykevl: `enableClock()` is also called in the moder version of `Pin.configure`, so it is essentially done twice. I think it should be removed here and done in the appropriate `Pin.configure` function when needed.

src/machine/machine_stm32f407.go
gwtnz: That's a reasonable assumption... could probably also see if the pins equalled each other:
```
if config.TX == config.RX {
...
```
I'll code it up as you recommend to keep the consistency across the project

src/machine/machine_stm32.go
```diff
@@ -2,6 +2,102 @@
...
+  default:
+    panic("machine: unknown port")
+  }
+}
+
+// configure sets the port I/O direction, speed, pull-up/down config etc
+func (p Pin) configure(config PinConfig) {
+	// perform GPIO configuration specific to this family
+}
+*/
```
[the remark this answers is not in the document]
gwtnz: This was intended to be guidelines for adding another family, e.g. `stm32f0` that had its own way of setting the registers, or a different mapping of bus to registers for example. I can remove it as folks can determine this from the actual code.

src/examples/stm32/f4-disco/blinky4/blinky4.go
```diff
@@ -0,0 +1,40 @@
+package main
+
+// Use goroutines to spin off led timing for each of the user LEDs on the stm32f4 discovery board
```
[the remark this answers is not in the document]
gwtnz: the stm32f4 discovery has 4 user LEDs; I didn't want folks getting the idea that they could e.g run this as is on an Arduino Uno. The LEDs aren't a strong argument for it to be on its own, but the example I've got for SPI to the discovery board's might be as that calls the accelerometer that's not even on most other discovery boards.

I'm happy to move the LEDs example though, and will add them to the smoke tests as described.

src/machine/machine_stm32_moder_gpio.go
gwtnz: can you elaborate a little? I'm not quite sure where the `pos -= 8` should be going...

src/machine/machine_stm32f103xx.go
gwtnz: see note above

src/machine/machine_stm32.go
```diff
@@ -2,6 +2,102 @@
...
+// Get returns the current value of a GPIO pin.
+func (p Pin) Get() bool {
+	port := p.getPort()
+	pin := uint8(p) % 16
+	val := port.IDR.Get() & (1 << pin)
+	return (val > 0)
+}
+
+// EnableAltFuncClock sends clock signals to the alternative function peripheral
+func EnableAltFuncClock(bus unsafe.Pointer) {
```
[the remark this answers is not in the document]
gwtnz: I was thinking it would be helpful for others developing for other peripherals that weren't in the base tinygo stm32 libraries yet, e.g. adding ADC. They could call the `EnableAltFuncClock` and it would set the clock up and enable the alternate fuction for them behind the scenes.
Granted, knowing you need to do this at all is almost the same as having to call the registers directly, so it may not be that useful as an export. I'll take it out.

src/machine/machine_stm32f407.go
```diff
@@ -171,20 +129,24 @@ func (uart UART) Configure(config UARTConfig) {
...
 
+	// use standard TX/RX pins PA2 and PA3
+	config.TX.Configure(PinConfig{Mode: PinModeUartTX})
+	config.RX.Configure(PinConfig{Mode: PinModeUartRX})
+
 	// Enable USART2 clock
-	stm32.RCC.APB1ENR.SetBits(stm32.RCC_APB1ENR_USART2EN)
+	EnableAltFuncClock(unsafe.Pointer(uart.Bus))
```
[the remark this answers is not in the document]
gwtnz: It was intended to make the peripherals implementations a little simpler - the next PR will be to refactor the UART setup, and then SPI, and so on, and the idea was to put all the common handling of the enabling of the peripheral in a single place; you'd just define your UART as using USART2 for example, and the common function would figure out what register needed to be set to which particular value.

src/runtime/volatile/register.go
```diff
@@ -160,3 +160,13 @@ func (r *Register32) ClearBits(value uint32) {
...
 	return (r.Get() & value) > 0
 }
+
+// SetMasked is a helper to simplify setting multiple bits high and/or low at
+// once. It is the volatile equivalent of:
+//
+//     r.Reg = (r.Reg & ^(mask << pos)) | value << pos
+//
+// go:inline
+func (r *Register32) SetMasked(value uint32, mask uint32, pos uint8) {
```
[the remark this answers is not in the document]
gwtnz: `ReplaceBits` works for me. I'll add in helpers for the other sizes too.

src/machine/machine_stm32f407.go
```diff
@@ -7,44 +7,40 @@ package machine
...
 }
 
+// Peripheral clock frequencies as set in runtime_stm32f407.go
+func APB1_Frequency() uint32 {
+	return CPUFrequency() / 4
+}
+
+func APB2_Frequency() uint32 {
+	return CPUFrequency() / 2
+}
```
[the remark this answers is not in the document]
gwtnz: They probably don't need to be exported. I'll update them.

src/machine/machine_stm32_moder_gpio.go
gwtnz: how about `PinModeUARTTX` and `PinModeUARTRX`? I can fix up the SPI and I2C constants too.

src/machine/machine_stm32.go
gwtnz: good catch, I'll fix that up.

gwtnz: Actually looking at it again, I'd put it here to make sure the clock got enabled so the individual `Pin.configure` calls didn't need to all ensure they called the clock enable. That would introduce a burden on each implementation of `configure`, though perhaps that's not a bad thing, since it needs to be included.

src/machine/machine_stm32.go
gwtnz: Along the same lines as the `EnableAltFunc()`, providing it as a building block for developers to add stuff external to the codebase without needing to write the lower level code directly when they wanted to use the GPIO.

src/machine/machine_stm32_moder_gpio.go
thinxer: Sorry, it should be `pin -= 8`.

```
	port := p.getPort()
	pin := uint8(p) % 16
	if pin >= 8 {
		pos := (pin - 8) * 4
		port.AFRH.SetMasked(...)
	} else {
		pos := pin * 4
		port.AFRL.SetMasked(...)
	}
```
When setting `AFRH`, the pin numbering starts with 8.

thinxer: The stm32f407 implementation uses only PA2 and PA3, so the bug has not been discovered. I was trying stm32f411 and found this.

Since there are many definitions (for example the USART interrupts) missing in the stm32f411 SVD files, I need some hacks to get it working. Probably I should wait until this patch is submitted.

In the meanwhile, I have discovered https://github.com/stm32-rs/stm32-rs, which fixed a lot of the bugs in the SVD files. Maybe we should generate device definitions from that repo.

src/machine/machine_stm32_moder_gpio.go
thinxer: Actually there pins are pretty standard. Periph.io define these pins in a standalone package and they are shared among all hardwares, for example: https://godoc.org/periph.io/x/periph/conn/uart.

src/machine/machine_stm32_moder_gpio.go
gwtnz: ah yes I see now. Thanks for the catch! I'll update the code. Good question on the SVD definitions; I was looking at getting a 401 and a 411 board so it's good to know there are gaps in the current implementation.

src/machine/machine_stm32f103xx.go
gwtnz: it looks like the f103 code was doing it right (line 42), but it seems the f407 code had the error I propagated. Thanks again for the catch!

src/examples/stm32/f4-disco/blinky4/blinky4.go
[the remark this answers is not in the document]
gwtnz: thinking about the SPI / accelerometer case, it's probably fine with the other examples too, as it just has general SPI <-> LIS302DL commands; this could work for any board / chip + the LIS302DL, so I think it can probably go with the rest of the examples.

src/machine/machine_stm32_moder_gpio.go
gwtnz: that's a pretty cool looking library! Good consideration for all the different types of pin usage likely to happen. I don't know which solution would be better:
- update the existing definitions in the one file, really just for the benefit of `stm32` users
- move the definitions to e.g. `uart.go`, `spi.go`, `i2c.go` etc in a similar way to the Periph.io library for other board/chip users to leverage in a common way
- import Periph.io and use its definitions directly

Thoughts? I'll go down the first path for now, as it's the minimal change, but this seems like a good thing to standardize for the tinygo platform.

src/machine/machine_stm32f407.go
```diff
@@ -7,44 +7,40 @@ package machine
...
 
+// Peripheral clock frequencies as set in runtime_stm32f407.go
+func apb1_Frequency() uint32 {
+	return CPUFrequency() / 4
+}
+
+func apb2_Frequency() uint32 {
+	return CPUFrequency() / 2
+}
+
```
gwtnz: I've left these in for now as they'll come into play with the UART and SPI refactor for calculating baud rates.

src/machine/machine_stm32_moder_gpio.go
thinxer: I agree with you on the first path to make this MR more feasible to be merged.

src/machine/machine_stm32f407.go
deadprogram: They should probably be named `apb1Frequency()` and `apb2Frequency()` to be more idiomatic function mames in Go.

But that is if they are left in at all. Seems like explicit inline  `(CPUFrequency() / 4)` is clearer than `apb1Frequency()` to me?

src/machine/machine_stm32f407.go
```diff
@@ -98,69 +94,31 @@ func (p Pin) enableClock() {
...
-func (p Pin) Set(high bool) {
-	port := p.getPort()
-	pin := p % 16
-	if high {
-		port.BSRR.Set(1 << uint8(pin))
-	} else {
-		port.BSRR.Set(1 << uint8(pin+16))
+// Enable peripheral clock
+func enableAltFuncClock(bus unsafe.Pointer) {
+	if bus == unsafe.Pointer(stm32.USART1) {
```
deadprogram: This should probably be a switch instead of `else if` for clarity.

```
switch bus {
    unsafe.Pointer(stm32.USART1):
        stm32.RCC.APB2ENR.SetBits(stm32.RCC_APB2ENR_USART1EN)
...
}
```

src/runtime/volatile/register.go
```diff
@@ -57,6 +57,16 @@ func (r *Register8) HasBits(value uint8) bool {
 	return (r.Get() & value) > 0
 }
 
+// ReplaceBits is a helper to simplify setting multiple bits high and/or low at
+// once. It is the volatile equivalent of:
+//
+//     r.Reg = (r.Reg & ^(mask << pos)) | value << pos
+//
+// go:inline
+func (r *Register8) ReplaceBits(value uint8, mask uint8, pos uint8) {
```
deadprogram: If we want to add `ReplaceBits()` to the `volatile` package, seems to me it would be best done in a separate PR.

src/machine/machine_stm32f407.go
gwtnz: I'm definitely ok with a rename, but hesitant to remove them, as they are referring to the peripheral bus speeds on the chip, as set in the `runtime` chip initialization code; ideally IMHO they'd go there. Otherwise for every peripheral the coder will need to refer back to the runtime chip init to determine what the current bus speeds are set to.
I agree that the explicit code makes clear what frequency is being used, but I feel it obscures why.

src/runtime/volatile/register.go
gwtnz: https://github.com/tinygo-org/tinygo/pull/897

src/machine/machine_stm32_moder_gpio.go
aykevl: @thinxer switching to stm32-rs for the SVD files may indeed be a good idea. It ensures we can re-use all the effort the Rust community put into fixing the SVD files :)

src/examples/stm32/f4-disco/blinky4/blinky4.go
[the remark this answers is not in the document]
aykevl: The main reason these examples are built in CI is to check that it compiles at all. Many trivial bugs are caught that way. So the specific example you're compiling doesn't matter all that much - as long as it basically works. As an example, the smoke test for the pca10056 board compiles examples/blinky1, even though it also has 4 LEDs on the board.
To be honest, I think a smoke test with just blinky1 or blinky2 would be fine, but I have no problem with adding another (small) example like blinky4. So I'm fine with the code as-is.

> I didn't want folks getting the idea that they could e.g run this as is on an Arduino Uno.

The same is also true for the blinky2 example. You'll immediately get a compiler error which should make the problem clear.

src/machine/machine_stm32_moder_gpio.go
aykevl: Pin modes fall roughly in two groups:

* Modes that are standard across chips and vendors, such as `PinInput`, `PinOutput`, `PinInputPullup`, etc. These modes are common across many chips and therefore should have a consistent name to make it easier to write portable code. Of course, when a given mode is not available on a chip, it should not be defined to get a clear compiler error.
* Modes that are chip (family/vendor) specific. Most chips have these, for example the SAM D family of chips has the "SERCOM" and "SERCOMAlt" modes used for serial protocols (SPI, I2C, ...). The UART Tx mode also falls in this category. They are inherently unportable and mostly used inside the machine package anyway so the exact name is less critical than the standard pin modes.

Long story short: the constants you have now look fine by me.

src/machine/machine_stm32f407.go
aykevl: Is this meant to be used in a future change? Because I can't find these functions being used anywhere in this PR.

Also, I agree with @deadprogram that these should be renamed to be more in Go style.

src/machine/machine_stm32_moder_gpio.go
```diff
@@ -0,0 +1,143 @@
...
+	PinModeSPIMISO PinMode = 10
+
+	// for analog/ADC
+	PinInputAnalog PinMode = 11
+
+	// for PWM
+
+	// Register values for the chip
+	// GPIOx_MODER
+	GPIOModeInput         = 0
```
aykevl: I guess you added these bitfields here because they don't exist in the stm32 package?

It might be better to move them over to the stm32 package (like src/device/stm32/stm32f103xx-bitfields.go) or maybe switch to the (hopefully more complete) stm32-rs package. But that can also be done at a later time.

src/machine/machine_stm32f407.go
gwtnz: yes, they show up in the UART and SPI refactors for calculating the baud rate those peripherals will end up with, and will probably appear in PWM when I get around to that.

Makefile
```diff
@@ -238,6 +238,8 @@ smoketest:
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=stm32f4disco        examples/blinky2
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=stm32f4disco        examples/blinky4
+	@$(MD5SUM) test.hex
```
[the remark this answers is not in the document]
gwtnz: i'm ok with taking it back out

src/machine/machine_stm32f407.go
gwtnz: I think given the confusion these are causing in this PR, I'll take them out for now and reintroduce them with the UART refactor which I'll submit once this one is in.

src/machine/machine_stm32f407.go
```diff
@@ -7,44 +7,31 @@ package machine
...
+	AF6_SPI3                          = 6
+	AF7_USART1_2_3                    = 7
+	AF8_USART4_5_6                    = 8
+	AF9_CAN1_CAN2_TIM12_13_14         = 9
+	AF10_OTG_FS_OTG_HS                = 10
+	AF11_ETH                          = 11
+	AF12_FSMC_SDIO_OTG_HS_1           = 12
+	AF13_DCMI                         = 13
+	AF14                              = 14
+	AF15_EVENTOUT                     = 15
```
aykevl: Same here as for `GPIOModeInput`, it seems better to me to move these constants over to src/device/stm32.

src/machine/machine_stm32.go
```diff
@@ -26,14 +21,6 @@ type AltFunc uint8
...
 // ---------- General pin operations ----------
 
-// Configure this pin with the given I/O settings.
-// Remember to call enableClock() in the chip-specific implementations!
-// Also call enableAltFuncClock() if using alternative output functions when
-//  configuring peripherals, e.g. UART, SPI etc.
-func (p Pin) Configure(config PinConfig) {
-	p.configure(config)
-}
-
```
gwtnz: I ended up removing this and setting the stm32f103 back to what it was before, as the other stm32 families needed an extra parameter for the alternate functions (UART, SPI etc).

src/machine/machine_stm32f407.go
```diff
@@ -101,15 +81,16 @@ func enableAltFuncClock(bus unsafe.Pointer) {
...
 }
 
 var (
-	// Both UART0 and UART1 refer to USART2.
 	UART0 = UART{
-		Buffer: NewRingBuffer(),
-		Bus:    stm32.USART2,
+		Buffer:          NewRingBuffer(),
+		Bus:             stm32.USART2,
+		AltFuncSelector: stm32.AF7_USART1_2_3,
```
gwtnz: These are somewhat dependent on the pins used, so ideally they'd be in the `UARTConfig` struct, but for now they're probably fine here; things just won't work if there's a mismatch between pins, bus and alt-func selected.

src/machine/machine_stm32f407.go
```diff
@@ -154,7 +135,7 @@ func (uart UART) Configure(config UARTConfig) {
 	// Enable USART2 port.
 	uart.Bus.CR1.Set(stm32.USART_CR1_TE | stm32.USART_CR1_RE | stm32.USART_CR1_RXNEIE | stm32.USART_CR1_UE)
 
-	// Enable RX IRQ.
+	// Enable RX IRQ. TODO: pick the right IRQ_xxx for the bus and the uart
 	intr := interrupt.New(stm32.IRQ_USART2, func(interrupt.Interrupt) {
 		UART1.Receive(byte((UART1.Bus.DR.Get() & 0xFF)))
 	})
```
gwtnz: This should probably live on the `UART1` definition, so it's more explicit about `USART2` interrupts being tied up there.

src/machine/machine_stm32.go
gwtnz: Would using interfaces for the peripherals make for a cleaner implementation?

gwtnz:
@aykevl @deadprogram any further thoughts on this?

src/machine/machine_stm32f407.go
aykevl: Hmm that extra field is unfortunate but okay for now.

src/machine/machine_stm32f407.go
```diff
@@ -197,24 +130,24 @@ func (uart UART) Configure(config UARTConfig) {
...
 	*/
-	stm32.USART2.BRR.Set(0x16c)
+	uart.Bus.BRR.Set(0x16c)
 
 	// Enable USART2 port.
-	stm32.USART2.CR1.Set(stm32.USART_CR1_TE | stm32.USART_CR1_RE | stm32.USART_CR1_RXNEIE | stm32.USART_CR1_UE)
+	uart.Bus.CR1.Set(stm32.USART_CR1_TE | stm32.USART_CR1_RE | stm32.USART_CR1_RXNEIE | stm32.USART_CR1_UE)
 
-	// Enable RX IRQ.
+	// Enable RX IRQ. TODO: pick the right IRQ_xxx for the bus and the uart
```
aykevl: You can take a look here for how this can be done:
https://github.com/tinygo-org/tinygo/blob/1d913a62bc0ceef841953058c2c6fb1dd2a40f6a/src/machine/board_circuitplay_express_baremetal.go#L10-L21
This pattern is very well optimized by the compiler.

src/machine/machine_stm32f407.go
gwtnz: yeah i couldn't come up with a better way to pass the alternate function parameter around; it's used on a number of the STM32 families, but even from one family to another the way it's set up varies :(

src/machine/machine_stm32f407.go
gwtnz: I was hoping to do something like:
```
var (
  UART1 = UART{
 		Buffer: NewRingBuffer(), 
 		Bus:     stm32.USART2, 
 		IRQ:     stm32.IRQ_USART2 , 
 	} 
) 

...

   intr := interrupt.New(uart.IRQ, func(interrupt.Interrupt) {
		uart.Receive(byte((uart.Bus.DR.Get() & 0xFF)))
	})
```

This way the `Configure()` call didn't need to know about specific bindings of IRQs/Buses to `uart` instances.

aykevl: Yeah I was hoping for that but both have a problem:

The first is not currently supported by the compiler. The `interrupt.New` function is a compiler intrinsic and will only accept constant parameters (both for the interrupt number and for the handler that you pass to it). In this case `uart.IRQ` is not constant. The problem with non-constant parameters is that it is incredibly hard (if not impossible) to optimize non-constant IRQ numbers.

The second is sadly not allowed by the Go specification because of a circular dependency. I really, really wished it was possible to write an interrupt handler like that because it looks so nice. The example I gave above works around that with an `init` function but is basically the same idea.

deadprogram:
I just discovered last night that after this commit, the blinky1 program compiles, but does not execute correctly on the bluepill. Nothing happens.

Just created https://github.com/tinygo-org/tinygo/issues/973 to track this.

PR #881, opened 2020-01-27
Implement `sync`

jaddr2line:
This is based on top of #828 and adds full support for the `sync` package, implementing blocking by using the `internal/task` package.

### DIFF
--- a/src/sync/cond.go
+++ b/src/sync/cond.go
@@ -54,6 +54,8 @@ func (c *Cond) Wait() {
 
 	// Temporarily unlock L.
 	c.L.Unlock()
+
+	// Re-acquire the lock before returning.
 	defer c.L.Lock()
 
 	// If we were signaled while unlocking, immediately complete.
@@ -62,16 +64,14 @@ func (c *Cond) Wait() {
 	}
 
 	// Remove the earlySignal frame.
-	if c.unlocking == &early {
-		c.unlocking = early.next
-	} else {
-		// Something else happened after the unlock - the earlySignal is somewhere in the middle.
-		// This would be faster but less space-efficient if it were a doubly linked list.
-		prev := c.unlocking
-		for prev.next != &early {
-			prev = prev.next
-		}
+	prev := c.unlocking
+	for prev != nil && prev.next != &early {
+		prev = prev.next
+	}
+	if prev != nil {
 		prev.next = early.next
+	} else {
+		c.unlocking = early.next
 	}
 
 	// Wait for a signal.
--- a/src/sync/cond.go
+++ b/src/sync/cond.go
@@ -1,0 +1,80 @@
+package sync
+
+import "internal/task"
+
+type Cond struct {
+	L Locker
+
+	unlocking *earlySignal
+	blocked   task.Stack
+}
+
+// earlySignal is a type used to implement a stack for signalling waiters while they are unlocking.
+type earlySignal struct {
+	next *earlySignal
+
+	signaled bool
+}
+
+func (c *Cond) trySignal() bool {
+	// Pop a blocked task off of the stack, and schedule it if applicable.
+	t := c.blocked.Pop()
+	if t != nil {
+		scheduleTask(t)
+		return true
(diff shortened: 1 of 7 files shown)


jaddr2line:
I have implemented most of `sync`, and I am not quite sure how to test this kind of stuff.

deadprogram:
Do you mean usage testing? If so then I can give this a try in the mqtt code we have that could really use a mutex.

aykevl:
Yeah not sure how to best test this either. The happy flow should already be mostly tested in testdata/channel.go, and that is the only thing that's easy to test. It's hard to prove the non-existence of a concurrency bug.

src/sync/cond.go
```diff
@@ -0,0 +1,80 @@
...
+	defer c.L.Lock()
+
+	// If we were signaled while unlocking, immediately complete.
+	if early.signaled {
+		return
+	}
+
+	// Remove the earlySignal frame.
+	if c.unlocking == &early {
+		c.unlocking = early.next
```
aykevl: This appears to be a special case of the loop below. IMHO removing this special case makes the code easier to read and should not result in any performance issues as the loop below will not even enter the first iteration.

src/sync/cond.go
```diff
@@ -0,0 +1,80 @@
...
+	} else {
+		// Something else happened after the unlock - the earlySignal is somewhere in the middle.
+		// This would be faster but less space-efficient if it were a doubly linked list.
```
aykevl: Something to consider: I don't think a doubly linked list would be a problem. At least with the current GC design it would not add any extra overhead as all objects are rounded up to 4 words anyway. Either choice seems fine by me, as a singly linked list is probably easier to reason about.

src/sync/cond.go
```diff
@@ -0,0 +1,80 @@
...
+		scheduleTask(t)
+		return true
+	}
+
+	// If there any tasks which are currently unlocking, signal one.
+	if c.unlocking != nil {
+		c.unlocking.signaled = true
+		c.unlocking = c.unlocking.next
+		return true
+	}
```
aykevl: So this code removes one entry from the `unlocking` queue. At the same time, `Wait` appears to pop off the earlySignal frame unconditionally. So wouldn't that result in a conflict (likely resulting in a nil pointer dereference in `Wait`)?

jaddr2line: There is no conflict, as marking it as signaled skips the pop.

src/sync/cond.go
jaddr2line: I am going to stick with a singly linked list, since the loop is only needed in rare cases where `Unlock` pauses or signals.

src/sync/cond.go
jaddr2line: I disagree, but I changed it.

jaddr2line:
CI had failed because of the issue described in #977, so I rebased on `dev`.

src/sync/cond.go
aykevl: (this was meant as a suggestion, not as a requirement)

My original idea was to replace this:

```go
	if c.unlocking == &early {
		c.unlocking = early.next
	} else {
		prev := c.unlocking
		for prev.next != &early {
			prev = prev.next
		}
		prev.next = early.next
	}
```

with this:

```go
	prev := c.unlocking
	for prev.next != &early {
		prev = prev.next
	}
	prev.next = early.next
```

but now I see that's not quite the same as it leaves `c.unlocking` unchanged. So I think the original code is easier to read that the modified version, it doesn't simplify the code as I hoped.

aykevl:
Looks good to me. However, some more testing would be good to have. @deadprogram you mentioned you might have some code with which you could test this?

jaddr2line:
@aykevl I added a test as requested.

aykevl:
Thanks a lot!

I have merged this without squasing because there were various commits that really were different features, although there were also some that might have been better squashed together with some other commit (such as the last one). I didn't want to lose that history.

PR #883, opened 2020-01-28
nrf52840: implement USB-CDC

syoder:
This is based on the implementation in `machine_atsamd51.go`, with changes made to use the USBD registers found on the nRF52840.

### DIFF
--- a/src/machine/board_pca10056.go
+++ b/src/machine/board_pca10056.go
@@ -28,12 +28,6 @@ const (
 	UART_RX_PIN Pin = 8
 )
 
-// UART0 is the NRF UART and UART1 is the nRF52840's USB device
-var (
-	UART0 = NRF_UART0
-	UART1 = USB
-)
-
 // ADC pins
 const (
 	ADC0 Pin = 3
--- a/src/machine/board_pca10056.go
+++ b/src/machine/board_pca10056.go
@@ -28,6 +28,12 @@ const (
 	UART_RX_PIN Pin = 8
 )
 
+// UART0 is the NRF UART and UART1 is the nRF52840's USB device
+var (
+	UART0 = NRF_UART0
+	UART1 = USB
+)
+
 // ADC pins
 const (
 	ADC0 Pin = 3
--- a/src/machine/usb_nrf52840.go
+++ b/src/machine/usb_nrf52840.go
@@ -3,12 +3,12 @@
 package machine
 
 import (
-	"device/nrf"
 	"device/arm"
+	"device/nrf"
 	"runtime/interrupt"
+	"runtime/volatile"
 	"unicode/utf16"
 	"unsafe"
-	"runtime/volatile"
 )
 
 // USBCDC is the USB CDC aka serial over USB interface on the nRF52840
@@ -68,7 +68,7 @@ var (
 	usbLineInfo      = cdcLineInfo{115200, 0x00, 0x00, 0x08, 0x00}
 	epinen           uint32
 	epouten          uint32
-	cdcInBusy				 volatile.Register8
+	cdcInBusy        volatile.Register8
 )
 
 // Configure the USB CDC interface. The config is here for compatibility with the UART interface.
@@ -95,7 +95,8 @@ func (usbcdc *USBCDC) Configure(config UARTConfig) {
 func (usbcdc *USBCDC) handleInterrupt(interrupt.Interrupt) {
 	// USBD ready event
 	if nrf.USBD.EVENTS_USBEVENT.Get() == 1 {
-		if (nrf.USBD.EVENTCAUSE.Get()&nrf.USBD_EVENTCAUSE_READY) > 0 {
+		nrf.USBD.EVENTS_USBEVENT.Set(0)
+		if (nrf.USBD.EVENTCAUSE.Get() & nrf.USBD_EVENTCAUSE_READY) > 0 {
 
 			// Configure control endpoint
 			initEndpoint(0, usb_ENDPOINT_TYPE_CONTROL)
@@ -106,7 +107,6 @@ func (usbcdc *USBCDC) handleInterrupt(interrupt.Interrupt) {
 
 			usbConfiguration = 0
 		}
-		nrf.USBD.EVENTS_USBEVENT.Set(0)
 		nrf.USBD.EVENTCAUSE.Set(0)
 	}
 
@@ -118,6 +118,7 @@ func (usbcdc *USBCDC) handleInterrupt(interrupt.Interrupt) {
 
 	if nrf.USBD.EVENTS_EP0DATADONE.Get() == 1 {
 		// done sending packet - either need to send another or enter status stage
+		nrf.USBD.EVENTS_EP0DATADONE.Set(0)
 		if sendOnEP0DATADONE.ptr != nil {
 			// previous data was too big for one packet, so send a second
 			sendViaEPIn(
@@ -132,7 +133,6 @@ func (usbcdc *USBCDC) handleInterrupt(interrupt.Interrupt) {
 			// no more data, so set status stage
 			nrf.USBD.TASKS_EP0STATUS.Set(1)
 		}
-		nrf.USBD.EVENTS_EP0DATADONE.Set(0)
 		return
 	}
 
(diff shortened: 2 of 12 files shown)


syoder:
This isn't done yet, but I wanted to post what I've done so far. I think it's getting close.

src/machine/usb_nrf52840.go
```diff
@@ -0,0 +1,541 @@
...
+	nrf.USBD.INTEN.Set(
+		nrf.USBD_INTEN_SOF |
+		nrf.USBD_INTEN_USBRESET |
+		nrf.USBD_INTEN_USBEVENT |
+		nrf.USBD_INTEN_EP0SETUP,
+	)
+
+	// enable IRQ
+	arm.SetPriority(nrf.IRQ_USBD, 0xc0)
+	arm.EnableIRQ(nrf.IRQ_USBD)
```
aykevl: This should use the new interrupt support as merged in #782.

src/machine/usb_nrf52840.go
```diff
@@ -0,0 +1,541 @@
...
+	nrf.USBD.EPINEN.Set(epinen)
+}
+
+func strToUTF16Descriptor(in string) []byte {
+	size := (len(in)<<1) + 2
+	out := make([]byte, size)
+	out[0] = byte(size)
+	out[1] = 0x03
+	// TODO: do a "real" conversion that supports more than just ascii chars
+	for i, char := range in {
```
aykevl: Note that you're already decoding Unicode code points here: `char` is actually a single Unicode code point (and `i` is the index into the string, which will increase by more than 1 when decoding non-ASCII characters).

syoder: I just copied what was done [here](https://github.com/tinygo-org/tinygo/blob/master/src/machine/machine_atsamd51.go#L1877-L1885) except I'm iterating runes instead of bytes. Neither this implementation or the other will work for anything except ascii characters. I was worried about pulling in the `unicode` package to convert, but it looks like I can do that fairly easily.

bgould:
I have this board and I can try testing soon.  Thanks for implementing this

syoder:
I seem to be stuck. The device seems to connect to my Mac and shows up as a serial port that I can open via `screen`. So I know all the setup / enumeration is happening properly. However, it locks up my main program - the only thing that works is stuff that runs in the interrupts.

I'm having this same problem with a nRF PWM driver that I'm writing too. It seems like whenever I enable an interrupt (even if the interrupt function is an empty function), everything stops working. 

In order to move this forward I think I need to connect a serial adapter to my board's UART pins so I can get some additional debugging output. I'll see if I can find something at my office tomorrow.

src/machine/usb_nrf52840.go
aykevl: Oh, it was more of a hint. And yeah, if possible it would be best to avoid the unicode package (it is really big and currently slow to compile).

syoder:
Questions that need to be answered:
 * Should everything in `src/machine/usb_nrf52840.go` be moved into `src/machine/machine_nrf52840.go`? I was looking for a way to break things out because those files are already so big, but 🤷‍♂ 
 * Should we set `machine.UART0` to be this new `USBCDC` object so that `putchar()` automatically writes to it and debug output goes to it? Or should we leave it at `machine.USB` as I have it now?
 * This currently sends PID/VID and string descriptors as set in `machine/usb.go` which are for an Arduino Zero. It looks like #884 refactors this a bit so that each board uses it's own - If that gets merged first, I should probably make a similar change here. If this gets merged first, that PR should include nrf52840 boards.
 * Same thing applies to fixing the way string descriptors are sent in the USB-CDC implementation for `sam` - if this gets merged first, the `strToUTF16LEDescriptor` function could be used there as well.

bgould:
> * Should we set `machine.UART0` to be this new `USBCDC` object so that `putchar()` automatically writes to it and debug output goes to it? Or should we leave it at `machine.USB` as I have it now?

For this one my vote would definitely be to make the USBCDC the default output.

aykevl:
> Should everything in src/machine/usb_nrf52840.go be moved into src/machine/machine_nrf52840.go? I was looking for a way to break things out because those files are already so big, but :man_shrugging: 

Considering how much code it is, it sounds good to keep it in a separate file (close to usb.go).

> Should we set `machine.UART0` to be this new USBCDC object so that `putchar()` automatically writes to it and debug output goes to it? Or should we leave it at `machine.USB` as I have it now?

That depends on the board. For the Bluefruit I think `putchar` should use USB-CDC while the pca10056 should use the UART (because that is what is normally connected).

syoder:
> That depends on the board. For the Bluefruit I think putchar should use USB-CDC while the pca10056 should use the UART (because that is what is normally connected).

What if we added something like this to  @`machine_nrf.go`:
```go
type ByteWriter interface {
  func WriteByte(c byte) error
}

var DEFAULT_OUTPUT ByteWriter
```

Then `putchar()` could use `machine.DEFAULT_OUTPUT` which could be set differently per board, but would also allow a developer to easily switch to whichever UART/USB device they wanted.

bgould:
I tested this on Circuit Playground Bluefruit and Adafruit CLUE boards, and it appears to work flawlessly so far.

With respect to the default output... I wonder if it might be enough to move the definition of the UART0 out of machine_nrf.go and into the individual board_*.go files?  I think that could be addressed in a separate if you didn't want this one to get larger.

syoder:
I wouldn't mind moving where `UART0` gets defined to the `board_*.go` files in this PR.

After a quick review, it looks like these are the currently supported nRF52840 based boards:
 * Circuit Playground Bluefruit
 * reel board
 * nRF52840-mdk
 * PCA10056

Of those, I'm guessing CPB and nRF52840-mdk would probably want USB-CDC to be UART0, while reel board and PCA10056 would probably want the actual UART? This is all just based on pictures of the boards on the internet, I'm not that familiar with any of them except for CPB.

syoder:
Looks like the smoke tests depend on `UART0` being defined in the `machine_*.go` files and fail if you try to define them in `board_*.go`. Unless anyone can tell me how to avoid this issue, I'll revert my last commit and we'll have to deal with that part in another PR.

src/machine/board_pca10056.go
```diff
@@ -28,6 +28,12 @@ const (
 	UART_RX_PIN Pin = 8
 )
 
+// UART0 is the NRF UART and UART1 is the nRF52840's USB device
+var (
+	UART0 = NRF_UART0
+	UART1 = USB
```
aykevl: The nrf52840 has an UART1 (actually UARTE1) so this would conflict.
I think it would be better to leave `UART1` undefined and expect programs to use `machine.USB` if they specifically need to write to that output.
(The same is true for other boards that primarily use a real UART for output).

aykevl:
> After a quick review, it looks like these are the currently supported nRF52840 based boards:
> 
>    * Circuit Playground Bluefruit
>    * reel board
>    * nRF52840-mdk
>    * PCA10056

I merged #893 so now there are five boards.

src/machine/board_pca10056.go
syoder: Should I also leave `UART1` undefined for boards that will use `USB` as the primary output? (programs would have to access `NRF_UART0` if they want to want to use the real UART)

aykevl: > Should I also leave `UART1` undefined for boards that will use `USB` as the primary output? (programs would have to access `NRF_UART0` if they want to want to use the real UART)

Not sure. @deadprogram any opinion on this?

deadprogram: So up until now, we have typically treated `UART0` as either a "real" UART or the USB-CDC interface, depending on what the board supports.

However, that now seems a little clumsy. I think we should make a chance and instead do something like what we do with `machine.LED` to have a well-known alias. We could name it something like `machine.SERIAL` or `machine.SERIALPORT`.

- Boards where there is a USB-CDC like basically all of the Adafruit boards, `machine.SERIAL = machine.USB`.
- Boards where there the "default" is actually a UART perhaps via a debug chip. then `machine.SERIAL = machine.UART0`.

If the user code needs to communicate with a specific UART, then that is easily handled. The default is the "sensible" option. And we can define all of the available ports per board without having to leave something out, or renumber.

aykevl: This sounds very reasonable. Completely agree.

That would also fix weird cases like this:

https://github.com/tinygo-org/tinygo/blob/1d913a62bc0ceef841953058c2c6fb1dd2a40f6a/src/machine/board_bluepill.go#L62-L68

deadprogram: `machine.Serial` seems good. Also sounds good to make those changes in a separate PR.

deadprogram:
Is there anything else outstanding on this PR before it can be merged? It would appear to be ready.

suda:
#904 will add three more nRF52 based boards, all of them default [UART0 to be the CDC and UART1 being the device's TX and RX pins](https://docs.particle.io/reference/device-os/firmware/boron/#serial) so definitely a +1 on UART0 being USB-CDC.

syoder:
As far as I know it's ready. Unless someone tells me something needs to be changed.

deadprogram:
OK, lets merge then. Massive thanks to @syoder for all your work on this, and also to everyone who helped test, provide feedback, and otherwise help make it happen!

PR #884, opened 2020-01-28
machine/usb: set the vid and pid to valid values supplied by Adafruit and Arduino

deadprogram:
The PR sets the vid and pid to valid values supplied by Adafruit and Arduino for boards that support USB CDC.

It might be able to help Windows 10 users get their devices to auto-mount as USB CDC without special drivers. cc/ @torntrousers

### DIFF
--- a/src/machine/board_feather-m0.go
+++ b/src/machine/board_feather-m0.go
@@ -102,3 +102,14 @@ const (
 	I2S_SD_PIN  = PA08
 	I2S_WS_PIN  = NoPin // TODO: figure out what this is on Feather M0.
 )
+
+// USB CDC identifiers
+const (
+	usb_STRING_PRODUCT      = "Feather M0 Express"
+	usb_STRING_MANUFACTURER = "Adafruit"
+)
+
+var (
+	usb_VID uint16 = 0x239A
+	usb_PID uint16 = 0x801B
+)
--- a/src/machine/board_feather-m0.go
+++ b/src/machine/board_feather-m0.go
@@ -105,7 +105,7 @@ const (
 
 // USB CDC identifiers
 const (
-	usb_STRING_PRODUCT      = "Feather M0 Express"
+	usb_STRING_PRODUCT      = "Adafruit Feather M0 Express"
 	usb_STRING_MANUFACTURER = "Adafruit"
 )
 
--- a/src/machine/board_particle_argon.go
+++ b/src/machine/board_particle_argon.go
@@ -94,3 +94,14 @@ const (
 	NFC1_PIN          Pin = 9
 	NFC2_PIN          Pin = 10
 )
+
+// USB CDC identifiers
+const (
+	usb_STRING_PRODUCT      = "Particle Argon"
+	usb_STRING_MANUFACTURER = "Particle"
+)
+
+var (
+	usb_VID uint16 = 0x239A
+	usb_PID uint16 = 0x8029
+)
--- a/src/machine/board_pca10056.go
+++ b/src/machine/board_pca10056.go
@@ -50,3 +50,14 @@ const (
 	SPI0_MOSI_PIN Pin = 45 // P1.13
 	SPI0_MISO_PIN Pin = 46 // P1.14
 )
+
+// USB CDC identifiers
+const (
+	usb_STRING_PRODUCT      = "Nordic nRF52840DK (PCA10056)"
+	usb_STRING_MANUFACTURER = "Nordic Semiconductor"
+)
+
+var (
+	usb_VID uint16 = 0x239A
+	usb_PID uint16 = 0x8029
+)
--- a/src/machine/board_arduino_nano33.go
+++ b/src/machine/board_arduino_nano33.go
@@ -90,3 +90,14 @@ const (
 	I2S_SD_PIN  Pin = PA08
 	I2S_WS_PIN      = NoPin // TODO: figure out what this is on Arduino Nano 33.
 )
+
+// USB CDC identifiers
+const (
+	usb_STRING_PRODUCT      = "Arduino NANO 33 IoT"
+	usb_STRING_MANUFACTURER = "Arduino"
+)
+
+var (
+	usb_VID uint16 = 0x2341
+	usb_PID uint16 = 0x8057
+)
--- a/src/machine/board_circuitplay_bluefruit.go
+++ b/src/machine/board_circuitplay_bluefruit.go
@@ -76,3 +76,14 @@ const (
 	SPI0_MOSI_PIN = P0_21 // MOSI
 	SPI0_MISO_PIN = P0_23 // MISO
 )
+
+// USB CDC identifiers
+const (
+	usb_STRING_PRODUCT      = "Adafruit Circuit Playground Bluefruit"
+	usb_STRING_MANUFACTURER = "Adafruit"
+)
+
+var (
+	usb_VID uint16 = 0x239A
+	usb_PID uint16 = 0x8045
+)
--- a/src/machine/board_circuitplay_express.go
+++ b/src/machine/board_circuitplay_express.go
@@ -88,3 +88,14 @@ const (
 	I2S_SD_PIN  = PA08
 	I2S_WS_PIN  = NoPin // no WS, instead uses SCK to sync
 )
+
+// USB CDC identifiers
+const (
+	usb_STRING_PRODUCT      = "Circuit Playground Express"
+	usb_STRING_MANUFACTURER = "Adafruit"
+)
+
+var (
+	usb_VID uint16 = 0x239A
+	usb_PID uint16 = 0x8018
+)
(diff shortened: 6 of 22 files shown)


syoder:
Another thing I discovered that might be tripping up windows is that I think we are sending the string descriptors incorrectly. See here: https://github.com/tinygo-org/tinygo/blob/master/src/machine/machine_atsamd51.go#L1876-L1887

There are several problems:
 * the header bytes (length and descriptor type) are being overwritten with the first character of the string
 * the ascii string is being converted to UTF-16BE, when it should be converted to UTF-16LE
 * at least on macOS, the USB host requests just the header (`wLength` == 2), and then afterwards requests the whole thing, otherwise it throws an error (but still allows it to connect)

deadprogram:
Hi @syoder I think you are totally correct about errors in that descriptor. I was looking in to that lasts night a little, but did not get quite as far as identifying specifics as you did.

I am going to try to correct a couple of those things today and see where that gets me.

src/machine/board_feather-m0.go
```diff
@@ -102,3 +102,14 @@ const (
 	I2S_SD_PIN  = PA08
 	I2S_WS_PIN  = NoPin // TODO: figure out what this is on Feather M0.
 )
+
+// USB CDC identifiers
+const (
+	usb_STRING_PRODUCT      = "Feather M0 Express"
```
torntrousers: Should this be "Feather M0 Express" or just "Feather M0"? I'm not sure what the difference is but there are separate board definitions in the Arduino IDE for each and the board I have is just the M0 Feather not the express board.

deadprogram: They have the same vendor ID (VID) and product ID (PID), so the description does not need to match. In any case, the initial descriptor 0 does not seem to be getting sent correctly.

aykevl:
What is the state of this PR? Does it fix the problem on Windows?

It looks good to me but I haven't tested it.

deadprogram:
We should merge in #883 first, and then I will rebase/cleanup this PR. Also I will try to correct the other errors in the sending of the descriptors.

sago35:
> What is the state of this PR? Does it fix the problem on Windows?

I fixed usb-cdc on windows 10.

I'm not sure if this fix is correct.
However, it has been confirmed that it works on Windows 10.

I have confirmed that when I connect pyportal / trinket-m0 on Windows 10, it is recognized as a USB serial device.
Also, it was confirmed that serial communication was possible with the test code described below.

However, I have only tested it on Windows 10, so I'm not sure it works on mac and linux.
Could you test it out?

https://github.com/sago35/tinygo/commit/d13764bd6ba126f4fa6f187943558f4c65b058ae

test code
https://gist.github.com/sago35/785bbfa745d4734e39e8185c1d58279d


I think that URB_FUNCTION_GET_MS_FEATURE_DESCRIPTOR in the figure below was bad, but I don't know.

image

sago35:
Reduced call conditions for sendZlp().
And rebased to the current dev branch.

* https://github.com/sago35/tinygo/tree/feature/usb-cdc-windows
  * https://github.com/sago35/tinygo/commit/f80cc339e5c2f18b559a2f94f59352c05329a816

deadprogram:
Made several updates to this branch to correct issues pointed out by @syoder and @sago35 although my fixes are slightly different.

I addressed the byte ordering and corrected skipping the header bytes when creating the descriptor strings. I also removed the case that an extra zero length packet was being sent in error along with the valid descriptor string

I did change the descriptor class being returned with the full descriptor, but I changed it to 0x02 as this is the correct base class for CDC https://www.usb.org/defined-class-codes#anchor_BaseClass02h

I have tested these changes on both Linux and Windows 10, using a Circuit Playground Express, a ItsyBitsy-M4, and a Circuit Playground Bluefruit. All performed as expected using the example/serial program.

The ATSAMD21 and ATSAMD51 boards also worked using the example/echo program. The Circuit Playground Bluefruit board did not work correctly with the example/echo program, in fact causing it to become unresponsive, however I think that issue is not related to this PR, and I will enter an issue to look into it separately.

That said, I think this PR is now ready for merge.

aykevl:
> The Circuit Playground Bluefruit board did not work correctly with the example/echo program, in fact causing it to become unresponsive, however I think that issue is not related to this PR, and I will enter an issue to look into it separately.

Is that a regression or was the problem already present?

deadprogram:
That problem is already present in the `dev` branch.

src/machine/board_feather-m0.go
ardnew: @torntrousers FYI the difference between "express" and non-"express" boards is the addition of extra SPI flash (for circuitpython)

src/machine/board_particle_argon.go
```diff
@@ -94,3 +94,14 @@ const (
...
+
+// USB CDC identifiers
+const (
+	usb_STRING_PRODUCT      = "Particle Argon"
+	usb_STRING_MANUFACTURER = "Particle"
+)
+
+var (
+	usb_VID uint16 = 0x239A
+	usb_PID uint16 = 0x8029
```
aykevl: This is the VID from Adafruit. Doesn't Particle have a different VID? (CC @suda).

src/machine/board_pca10056.go
```diff
@@ -50,3 +50,14 @@ const (
...
+// USB CDC identifiers
+const (
+	usb_STRING_PRODUCT      = "Nordic nRF52840DK (PCA10056)"
+	usb_STRING_MANUFACTURER = "Nordic Semiconductor"
+)
+
+var (
+	usb_VID uint16 = 0x239A
+	usb_PID uint16 = 0x8029
+)
```
aykevl: 1. Is there a reason you're re-using the VID from Adafruit?
2. It looks like the nrf52*-based boards do not use these `usb_STRING_PRODUCT` etc constants yet.

src/machine/board_particle_argon.go
deadprogram: The Adafruit drivers use the Adafruit VID for boards that do not have their own, such as Particle:
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/boards.txt#L444

src/machine/board_pca10056.go
deadprogram: 1. see https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/boards.txt#L419

2. That is true, I will add another commit that uses the constant.

deadprogram: Actually the nrf52840 boards all do use the constants. See https://github.com/tinygo-org/tinygo/blob/feature/usb-cdc-ids/src/machine/usb_nrf52840.go#L432 and https://github.com/tinygo-org/tinygo/blob/feature/usb-cdc-ids/src/machine/usb_nrf52840.go#L440

src/machine/board_particle_argon.go
suda: Yep, Particle has different PIDs for all three boards. I wanted to add them in a separate PR after this one is merged in 👍🏻

suda: The VID for 3rd gen is `0x2B04` and PIDs are respectively: `0xD00C` for Argon, `0xD00D` for Boron and `0xD00E` for Xenon ([source](https://github.com/particle-iot/device-os/blob/master/build/platform-id.mk#L287)).

deadprogram: I propose after we merge this PR then @suda can create a separate PR to connect the ids for the Particle boards, as he mentioned.

src/machine/board_pca10056.go
aykevl: Oh okay, then it's fine.

deadprogram:
Let me modify the product names that are inconsistent, then.

deadprogram:
OK, added another commit for name consistency.

sago35:
@deadprogram 

I tried feature/usb-cdc-ids in example/echo but it didn't work.
I'm just experimenting with swapping the source for tinygo-dev and the environment may be bad.
Please give uf2 made by `tinygo build -o app.uf2 -target pyportal .`
I want to check the operation with Windows10 Pro 64bit + pyportal.

image

```
# docker : tinygo/tinygo-dev
# example/echo

pushd /tinygo
git remote add gg https://github.com/tinygo-org/tinygo
git remote add gg https://github.com/tinygo-org/tinygo
git fetch gg
git checkout gg/feature/usb-cdc-ids
popd
tinygo build -o app.uf2 -target pyportal .
```

deadprogram:
Hello @sago35 you are correct that the echo example does not yet work on Windows. But it at least solves the original problem, which is to get the correct name, and the board driver+USB Serial port drivers to be installed.

The problem whatever it might be with Windows and reading data off of the USB CDC from the device will have to be solved in another PR. This is at least progress.

src/machine/board_particle_argon.go
deadprogram: This PR was merged, so @suda we are ready for your separate PR.

suda: @deadprogram https://github.com/tinygo-org/tinygo/pull/985 done!

PR #892, opened 2020-02-05
Add constants for QSPI pins for ATSAMD51

bgould:
The QSPI peripheral is only available on a fixed set of pins, so these can be added as constants for all atsamd51 boards.


bgould:
The constants in this PR are used in the QSPI example in https://github.com/tinygo-org/drivers/pull/124

aykevl:
Thank you! Seems reasonable to me so merging.

PR #893, opened 2020-02-07
Adding Adafruit CLUE Alpha

bgould:
https://www.adafruit.com/clue

They note that this is an alpha release of the board and that the pinout/sensors could change, so I've named this target "clue_alpha" in case such changes do occur.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -242,6 +242,8 @@ smoketest:
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=circuitplay-express examples/i2s
 	@$(MD5SUM) test.hex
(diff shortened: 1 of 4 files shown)


aykevl:
Thanks!
So of course the actual hardware produced never goes out of "alpha" - it's what it is today. So we can't just drop support for that. Therefore, I would like to propose the following for when this board comes out of the alpha phase:

  * If everything stays the same (or compatible), rename it to `clue` and provide a `clue_alpha.json` file for compatibility that simply inherits the `clue.json` file: `{"inherits": ["clue"]}`.
  * If the board changes after the alpha period, keep both versions around.

PR #895, opened 2020-02-07
Generalize -scheduler=none to support most platforms

jaddr2line:
This PR is based on top of #828, and generalizes the "none" scheduler to work on platforms other than AVR. It should now work on all platforms except WebAssembly. Init should also be run in the main goroutine when using a scheduler, but this is not yet possible as the runtime must be initialized before starting a goroutine.


jaddr2line:
I rebased this on the final merged version of #828, and added in the build tag fix. This is ready for review I think.

aykevl:
Oh well I'll merge it anyway. Not too big of a deal.

PR #896, opened 2020-02-09
nRF: fix PinInputPulldown

syoder:
Fixes a typo which made pins configured with `PinInputPulldown` on nRF SoCs not work correctly.


deadprogram:
Thanks for the correction @syoder now merging.

PR #897, opened 2020-02-09
include ReplaceBits method

gwtnz:
Allow multiple bit setting on a register, with masking.
Related to https://github.com/tinygo-org/tinygo/issues/777


aykevl:
Thank you! Looks reasonable to me.

PR #898, opened 2020-02-10
1200 baud auto-reset over USB for nrf52840

bgould:
This PR depends on https://github.com/tinygo-org/tinygo/pull/883, and should be rebased once that one is merged.  It implements the functionality for auto-resetting into the UF2 bootloader when a 1200 baud signal is received.

I protected this functionality with build tags since it depends on the UF2 bootloader, and also to enable disabling this functionality if desired.  If preferable however it could be included in the firmware by default without build tags.

### DIFF
--- a/src/machine/usb_nrf52840_reset_uf2.go
+++ b/src/machine/usb_nrf52840_reset_uf2.go
@@ -13,32 +13,34 @@ const (
 	DFU_MAGIC_OTA_RESET         = 0xA8
 )
 
+// checkShouldReset is called by the USB-CDC implementation to check whether to
+// reset into the bootloader/OTA and if so, resets the chip appropriately.
 func checkShouldReset() {
 	if usbLineInfo.dwDTERate == 1200 && usbLineInfo.lineState&usb_CDC_LINESTATE_DTR == 0 {
 		EnterUF2Bootloader()
 	}
 }
 
-// EnterSerialBootloader should perform a system reset in preparation
-// to switch to the bootloader to flash new firmware via serial/nrfutil
+// EnterSerialBootloader resets the chip into the serial bootloader. After
+// reset, it can be flashed using serial/nrfutil.
 func EnterSerialBootloader() {
 	arm.DisableInterrupts()
-	nrf.POWER.GPREGRET.Reg = uint32(DFU_MAGIC_SERIAL_ONLY_RESET)
+	nrf.POWER.GPREGRET.Set(DFU_MAGIC_SERIAL_ONLY_RESET)
 	arm.SystemReset()
 }
 
-// EnterUF2Bootloader should perform a system reset in preparation
-// to switch to the bootloader to flash new firmware via MSD/UF2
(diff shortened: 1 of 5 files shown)


deadprogram:
Hi @bgould can you please rebase now that #883 was merged? Thank you!

bgould:
Ok, rebased and tested on Adafruit Clue and Circuit Playground Bluefruit.  Should be good to go pending review

src/machine/usb_nrf52840_reset_uf2.go
```diff
@@ -0,0 +1,44 @@
...
+	if usbLineInfo.dwDTERate == 1200 && usbLineInfo.lineState&usb_CDC_LINESTATE_DTR == 0 {
+		EnterUF2Bootloader()
+	}
+}
+
+// EnterSerialBootloader should perform a system reset in preparation
+// to switch to the bootloader to flash new firmware via serial/nrfutil
+func EnterSerialBootloader() {
+	arm.DisableInterrupts()
+	nrf.POWER.GPREGRET.Reg = uint32(DFU_MAGIC_SERIAL_ONLY_RESET)
```
aykevl: This should use a volatile operation, not a direct register write.

```suggestion
	nrf.POWER.GPREGRET.Set(DFU_MAGIC_SERIAL_ONLY_RESET)
```

The same goes for all other `.Reg` stores.

src/machine/usb_nrf52840_reset_uf2.go
```diff
@@ -0,0 +1,44 @@
...
+	"device/nrf"
+)
+
+const (
+	DFU_MAGIC_SERIAL_ONLY_RESET = 0x4e
+	DFU_MAGIC_UF2_RESET         = 0x57
+	DFU_MAGIC_OTA_RESET         = 0xA8
+)
+
+func checkShouldReset() {
```
aykevl: Can you add some comment here? For example:

```go
// checkShouldReset checks whether to reset into the bootloader/OTA and if so, resets the chip appropriately.
```

src/machine/usb_nrf52840_reset_uf2.go
```diff
@@ -0,0 +1,44 @@
...
+	if usbLineInfo.dwDTERate == 1200 && usbLineInfo.lineState&usb_CDC_LINESTATE_DTR == 0 {
+		EnterUF2Bootloader()
+	}
+}
+
+// EnterSerialBootloader should perform a system reset in preparation
+// to switch to the bootloader to flash new firmware via serial/nrfutil
```
aykevl: Nit: there is no 'should', `EnterSerialBootloader` _will_ perform a system reset. Example:

```go
// EnterSerialBootloader resets the chip into the serial bootloader. After reset, it can be flashed using serial/nrfutil.
```

(not critical for merging)

bgould:
Thank you for the feedback @aykevl, I've incorporated your suggestions in the latest commit.

aykevl:
Looks good! I've merged these changes in the dev branch.

PR #899, opened 2020-02-11
Use go mod expected path for go.bug.st/serial.

QuLogic:
Since the [source install instructions](https://tinygo.org/getting-started/linux/#source-install) say to enable Go modules, then the path to go.bug.st/serial should use the go mod version. And that should not use `.v1`, [according to upstream](https://github.com/bugst/go-serial#gomod-transition).

Note, I was a bit lazy, and just edited the paths by hand without checking `go mod ...`, so hopefully CI will confirm that this is correct (mostly just unsure about the sums).


QuLogic:
Oh, well that's annoying. If I run `go mod tidy`, it still adds another `.v1` as an indirect dependency. I wonder if upstream is incorrect about this.

QuLogic:
OK, that was because it was set to an old commit. Now using v1.0.0 and everything works. [Here is the upstream diff](https://github.com/bugst/go-serial/compare/5f7892a7bb45...v1.0.0).

aykevl:
Thank you! I tested this locally and found a very similar diff after updating the dependency and running `go mod tidy` (the main difference was that locally a number of old LLVM versions from go.sum got deleted). I also tested this on the Arduino Nano 33, which it is still able to flash.

Will merge now.

QuLogic:
Yea, the old LLVM versions were deleted for me too. I didn't add them to the PR because they don't really hurt anything being there and it would make the PR cluttered, but you may want to remove them too.

Issue #437, opened 2019-07-05
error: interp: branch on a non-const-propagated constant expression


diamondburned:
This error happened when compiling https://github.com/diamondburned/smart-contract-go/blob/master/examples/transfer-back/main.go with `/tinygo build -o ./main.wasm`.

With `-printir`: http://0x0.st/zLo5.txt

`error: interp: branch on a non-constant` is returned with `-gc none`.

Since Google suggests something about `const`s not being supported, I wrote some smaller code to confirm that, but couldn't.

diamondburned:
Output of `llc -version`:

```sh
LLVM (http://llvm.org/):
  LLVM version 8.0.0
  Optimized build.
  Default target: x86_64-unknown-linux-gnu
  Host CPU: skylake

  Registered Targets:
    aarch64    - AArch64 (little endian)
    aarch64_be - AArch64 (big endian)
    amdgcn     - AMD GCN GPUs
    arm        - ARM
    arm64      - ARM64 (little endian)
    armeb      - ARM (big endian)
    bpf        - BPF (host endian)
    bpfeb      - BPF (big endian)
    bpfel      - BPF (little endian)
    hexagon    - Hexagon
    lanai      - Lanai
    mips       - MIPS (32-bit big endian)
    mips64     - MIPS (64-bit big endian)
    mips64el   - MIPS (64-bit little endian)
    mipsel     - MIPS (32-bit little endian)
    msp430     - MSP430 [experimental]
    nvptx      - NVIDIA PTX 32-bit
    nvptx64    - NVIDIA PTX 64-bit
    ppc32      - PowerPC 32
    ppc64      - PowerPC 64
    ppc64le    - PowerPC 64 LE
    r600       - AMD GPUs HD2XXX-HD6XXX
    sparc      - Sparc
    sparcel    - Sparc LE
    sparcv9    - Sparc V9
    systemz    - SystemZ
    thumb      - Thumb
    thumbeb    - Thumb (big endian)
    wasm32     - WebAssembly 32-bit
    wasm64     - WebAssembly 64-bit
    x86        - 32-bit X86: Pentium-Pro and above
    x86-64     - 64-bit X86: EM64T and AMD64
    xcore      - XCore
```

Output of `tinygo version`:

```
tinygo version 0.7.0-dev linux/amd64
```

If it helps:

```nix
{ pkgs ? import <nixpkgs> {} }:

let
  unstable = import <unstable> {};
[...]

PR #901, opened 2020-02-11
Support Go 1.14.

QuLogic:
I believe this works without issues, but I've also got to figure out LLVM 10 at the same time. I also have only checked 0.11.0, not 0.12.0 yet. So let's see how CI feels about this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -229,14 +229,14 @@ commands:
       - run:
           name: "Install dependencies"
           command: |
-            curl https://dl.google.com/go/go1.13.darwin-amd64.tar.gz -o go1.13.darwin-amd64.tar.gz
-            sudo tar -C /usr/local -xzf go1.13.darwin-amd64.tar.gz
+            curl https://dl.google.com/go/go1.14.darwin-amd64.tar.gz -o go1.14.darwin-amd64.tar.gz
+            sudo tar -C /usr/local -xzf go1.14.darwin-amd64.tar.gz
             ln -s /usr/local/go/bin/go /usr/local/bin/go
             HOMEBREW_NO_AUTO_UPDATE=1 brew install qemu
       - restore_cache:
           keys:
-            - go-cache-macos-v2-{{ checksum "go.mod" }}-{{ .Environment.CIRCLE_PREVIOUS_BUILD_NUM }}
-            - go-cache-macos-v2-{{ checksum "go.mod" }}
+            - go-cache-macos-v3-{{ checksum "go.mod" }}-{{ .Environment.CIRCLE_PREVIOUS_BUILD_NUM }}
+            - go-cache-macos-v3-{{ checksum "go.mod" }}
       - restore_cache:
           keys:
             - llvm-source-9-macos-v0
@@ -298,7 +298,7 @@ commands:
             sudo tar -C /usr/local --strip-components=1 -xf riscv64-unknown-elf-gcc-8.2.0-2019.05.3-x86_64-apple-darwin.tar.gz
       - run: make smoketest AVR=0
       - save_cache:
-          key: go-cache-macos-v2-{{ checksum "go.mod" }}-{{ .Environment.CIRCLE_BUILD_NUM }}
+          key: go-cache-macos-v3-{{ checksum "go.mod" }}-{{ .Environment.CIRCLE_BUILD_NUM }}
           paths:
             - ~/.cache/go-build
             - ~/.cache/tinygo
@@ -319,6 +319,11 @@ jobs:
   test-llvm9-go113:
     docker:
       - image: circleci/golang:1.13-buster
+    steps:
+      - test-linux
+  test-llvm9-go114:
+    docker:
+      - image: circleci/golang:1.14-buster
     steps:
       - test-linux
   assert-test-linux:
@@ -346,6 +351,7 @@ workflows:
       - test-llvm9-go111
       - test-llvm9-go112
       - test-llvm9-go113
+      - test-llvm9-go114
       - build-linux
       - build-macos
       - assert-test-linux
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,7 +12,7 @@ jobs:
   steps:
     - task: GoTool@0
       inputs:
-        version: '1.13.8'  
+        version: '1.14.1'
     - checkout: self
     - task: CacheBeta@0
       displayName: Cache LLVM source
--- a/builder/config.go
+++ b/builder/config.go
@@ -25,8 +25,8 @@ func NewConfig(options *compileopts.Options) (*compileopts.Config, error) {
 	if err != nil {
 		return nil, fmt.Errorf("could not read version from GOROOT (%v): %v", goroot, err)
 	}
-	if major != 1 || (minor != 11 && minor != 12 && minor != 13) {
-		return nil, fmt.Errorf("requires go version 1.11, 1.12, or 1.13, got go%d.%d", major, minor)
+	if major != 1 || minor < 11 || minor > 14 {
+		return nil, fmt.Errorf("requires go version 1.11, 1.12, 1.13, or 1.14, got go%d.%d", major, minor)
 	}
 	clangHeaderPath := getClangHeaderPath(goenv.Get("TINYGOROOT"))
 	return &compileopts.Config{
--- a/cgo/cgo_test.go
+++ b/cgo/cgo_test.go
@@ -11,6 +11,7 @@ import (
 	"go/types"
 	"io/ioutil"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 	"testing"
@@ -18,6 +19,22 @@ import (
 
 // Pass -update to go test to update the output of the test files.
 var flagUpdate = flag.Bool("update", false, "Update images based on test output.")
+
+// normalizeResult normalizes test results across Go versions.
+// Currently, it does so by using consistent newlines, and wrapping
+// function contents to be consistent with Go 1.14.
+func normalizeResult(result string) string {
+	actual := strings.Replace(result, "\r\n", "\n", -1)
+
+	// Make sure all functions are wrapped, even those that would otherwise be
+	// single-line functions. This is necessary because Go 1.14 changed the way
+	// such functions are wrapped and it's important to have consistent test
+	// results.
+	re := regexp.MustCompile(`func \((.+)\)( .*?) +{ (.+) }`)
+	actual = re.ReplaceAllString(actual, "func ($1)$2 {\n\t$3\n}")
+
+	return actual
+}
 
 func TestCGo(t *testing.T) {
 	var cflags = []string{"--target=armv6m-none-eabi"}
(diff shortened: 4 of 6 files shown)


QuLogic:
Ah, took a few tries, but finally some actual errors. I guess I didn't see them yet since I usually run smoke test first, which hit LLVM 10 issues.

QuLogic:
The failure in `TestCGo/types` appears to just be some line breaks:
```patch
diff --git a/cgo/testdata/types.out.go b/cgo/testdata/types.out.go
index 8a75008..1864e33 100644
--- a/cgo/testdata/types.out.go
+++ b/cgo/testdata/types.out.go
@@ -62,12 +62,16 @@ type C.union3_t = C.union_1
 type C.union_nested_t = C.union_3
 type C.unionarray_t = struct{ arr [10]C.uchar }
 
-func (s *C.struct_4) bitfield_a() C.uchar          { return s.__bitfield_1 & 0x1f }
-func (s *C.struct_4) set_bitfield_a(value C.uchar) { s.__bitfield_1 = s.__bitfield_1&^0x1f | value&0x1f<<0 }
+func (s *C.struct_4) bitfield_a() C.uchar { return s.__bitfield_1 & 0x1f }
+func (s *C.struct_4) set_bitfield_a(value C.uchar) {
+	s.__bitfield_1 = s.__bitfield_1&^0x1f | value&0x1f<<0
+}
 func (s *C.struct_4) bitfield_b() C.uchar {
 	return s.__bitfield_1 >> 5 & 0x1
 }
-func (s *C.struct_4) set_bitfield_b(value C.uchar) { s.__bitfield_1 = s.__bitfield_1&^0x20 | value&0x1<<5 }
+func (s *C.struct_4) set_bitfield_b(value C.uchar) {
+	s.__bitfield_1 = s.__bitfield_1&^0x20 | value&0x1<<5
+}
 func (s *C.struct_4) bitfield_c() C.uchar {
 	return s.__bitfield_1 >> 6
 }
@@ -100,19 +104,31 @@ func (union *C.union_1) unionfield_s() *C.short { return (*C.short)(unsafe.Point
 
 type C.union_1 struct{ $union uint64 }
 
-func (union *C.union_2) unionfield_area() *C.point2d_t  { return (*C.point2d_t)(unsafe.Pointer(&union.$union)) }
-func (union *C.union_2) unionfield_solid() *C.point3d_t { return (*C.point3d_t)(unsafe.Pointer(&union.$union)) }
+func (union *C.union_2) unionfield_area() *C.point2d_t {
+	return (*C.point2d_t)(unsafe.Pointer(&union.$union))
+}
+func (union *C.union_2) unionfield_solid() *C.point3d_t {
+	return (*C.point3d_t)(unsafe.Pointer(&union.$union))
+}
 
 type C.union_2 struct{ $union [3]uint32 }
 
-func (union *C.union_3) unionfield_point() *C.point3d_t    { return (*C.point3d_t)(unsafe.Pointer(&union.$union)) }
-func (union *C.union_3) unionfield_array() *C.unionarray_t { return (*C.unionarray_t)(unsafe.Pointer(&union.$union)) }
-func (union *C.union_3) unionfield_thing() *C.union3_t     { return (*C.union3_t)(unsafe.Pointer(&union.$union)) }
+func (union *C.union_3) unionfield_point() *C.point3d_t {
+	return (*C.point3d_t)(unsafe.Pointer(&union.$union))
+}
+func (union *C.union_3) unionfield_array() *C.unionarray_t {
+	return (*C.unionarray_t)(unsafe.Pointer(&union.$union))
+}
+func (union *C.union_3) unionfield_thing() *C.union3_t {
+	return (*C.union3_t)(unsafe.Pointer(&union.$union))
+}
 
 type C.union_3 struct{ $union [2]uint64 }
 
-func (union *C.union_union2d) unionfield_i() *C.int      { return (*C.int)(unsafe.Pointer(&union.$union)) }
-func (union *C.union_union2d) unionfield_d() *[2]float64 { return (*[2]float64)(unsafe.Pointer(&union.$union)) }
+func (union *C.union_union2d) unionfield_i() *C.int { return (*C.int)(unsafe.Pointer(&union.$union)) }
+func (union *C.union_union2d) unionfield_d() *[2]float64 {
+	return (*[2]float64)(unsafe.Pointer(&union.$union))
+}
 
 type C.union_union2d struct{ $union [2]uint64 }
 type C.enum_option C.int
```

The others look to be the same "branch on a non-constant":
* `TestCompiler/EmulatedCortexM3/stdlib.go`: `main_test.go`:139: failed to build: interp: branch on a non-constant
* `TestCompiler/Host/stdlib.go`: `main_test.go`:139: failed to build: interp: branch on a non-constant
* `TestCompiler/WebAssembly/stdlib.go`: `main_test.go`:139: failed to build: interp: branch on a non-constant
* `TestCompiler/ARM64Linux/stdlib.go`: `main_test.go`:139: failed to build: interp: branch on a non-constant

plus some odd things about node, which might be related to the above.

aykevl:
I really don't like the idea of having different tests for different versions of Go. Especially when the differences are basically just gofmt changes. I think this could be solved with a regex to make sure the output matches the Go 1.14 formatting (so that in a few years from now, we could simply drop the regex when Go 1.13 and below are no longer supported).

Regarding the other errors, I suspect they are variations on https://github.com/tinygo-org/tinygo/issues/437

QuLogic:
I suppose we could replace all newlines with a space, and then replace all space runs with a single space. Though that's broken code, I guess it would still check what's really important. Otherwise I'm not sure there's an easy regex to fix it as I don't know what triggers the extra wrapping.

QuLogic:
That's not to say I can't write a regex for it; it'll just seem a bit arbitrary.

aykevl:
My idea was more like this, to recognize the following pattern using a regex:

```
func <stuff>    { <stuff> }
```
and replace it with this:
```
func <stuff> {
    <stuff>
}
```

QuLogic:
OK, that's also doable, but it doesn't always affect every `func`, unless you just want to do it for all of them?

QuLogic:
I pushed a change that does the wrapping conservatively, only on the lines that changed. If you would prefer the wrapping to be always applied, that could be done instead.

deadprogram:
We are going to need this sooner rather than later, especially now that Homebrew installs Go 1.14 by default.

deadprogram:
Hopefully some work can be done on this :pray:

QuLogic:
Rebased so we can get nice things like 982b2d06ab640804c220068b8eeb845d198cc701 in the output.

QuLogic:
So it's pointing at `math/rand` (at least):
```
    TestCompiler/ARMLinux/stdlib.go: main.go:663: # math/rand
    TestCompiler/ARMLinux/stdlib.go: main.go:665: interp: branch on a non-constant
```
The diff between 1.13.8 and 1.14 for that directory is not very large: https://github.com/golang/go/compare/go1.13.8...go1.14#diff-6e6d9989336ec13fdae43d4534e5f162
Is it the global type assert on `rngSource` that's bugging it?

aykevl:
I think it's this:

```go
var _ *rngSource = globalRand.src.(*lockedSource).src
```

I'm working on a fix.

aykevl:
Here is a partial fix: https://github.com/tinygo-org/tinygo/pull/967

QuLogic:
Toggled the `Debug` flag in the tests, which gets us proper line numbers:
```
    TestCompiler/Host/stdlib.go: main.go:663: # math/rand
    TestCompiler/Host/stdlib.go: main.go:665: 296:31: interp: branch on a non-constant
```
and line 296 is indeed the `rngSource` type assert line.

deadprogram:
@QuLogic can you please rebase against dev, since #967 has been merged? Thank you.

QuLogic:
Unfortunately, I already did.

aykevl:
This should fix the problem with math/rand: https://github.com/tinygo-org/tinygo/pull/983

QuLogic:
Indeed, that fixes it and is a good catch; I was looking much deeper down the IR passes. Looks like the only thing left is some WASM stuff now.

QuLogic:
OK, so the wasm failure is because `wasm_exec.js` is out of sync with upstream. The one the tests are tripping on is [this addition of garbage collection](https://github.com/golang/go/commit/54e6ba6724dfde355070238f9abc16362cac2e3d). It seems somewhat straightforward to incorporate that, but there are many other changes to `wasm_exec.js` that maybe could be synced in  too if you want to.

deadprogram:
I have been wanting to update the `wasm_exec.js` for a while now, we've gotten pretty far out of sync.

aykevl:
> The main problem is going to be that [the type flag numbers changed](https://github.com/golang/go/commit/54e6ba6724dfde355070238f9abc16362cac2e3d#diff-956ed5abd94e3f82adfe493f134403b5R219).

Ouch, I'm afraid this may be a big problem. In the worst case we'll have to drop support for WebAssembly with Go 1.13 and lower.

aykevl:
@QuLogic since #966 it is possible to select the Go version in Azure Pipelines. It is still at Go 1.13 with this PR. I think it would be a good idea to update this PR to switch both the Windows and the macOS builds over to Go 1.14 to have a better idea of what still needs to be done.

aykevl:
> I pushed a change that does the wrapping conservatively, only on the lines that changed. If you would prefer the wrapping to be always applied, that could be done instead.

Oops I only read this properly now.
Yes I would prefer the wrapping to be applied in all cases, to make the normalizing code simpler. In fact, I've already written that (https://github.com/tinygo-org/tinygo/compare/dev...normalize-cgo-tests) and am now re-reading your comment again. So, feel free to take it (or not) for this PR.

deadprogram:
Regarding dropping support for lower versions of wasm, that is not a great option for all of the various projects now using tinygo, but I suspect most of them would accept this tradeoff in return for the updated functionality.

Any comments from WebAssembly devs using TinyGo?

One thing we should try to consider on this round of updates, is how we can keep the `wasm_exec.js` file in better sync, or at least more frequent sync.

QuLogic:
I've gone through most of the `wasm_exec.js` changes, but haven't posted them. I'll try and get that in a PR tomorrow.

QuLogic:
I made PRs for the simple bits of syncing wasm_exec. What remains is the following:
* Changes to fix argv/envp layout; TinyGo's wasm_exec has removed support for these, and I don't know why that decision was made other than 'not working'. This fixes _may_ make it work again.
* synchronous callbacks to js/wasm: https://github.com/golang/go/commit/6dd70fc5e391eb7a47be5eb6353107f38b73f161 + various followup commits. This has lots of corresponding changes to `src/runtime`, so I think not compatible with Go 1.13.
* reuse wasm memory DataView: https://github.com/golang/go/commit/07f04607370bf46c6f3a12ac3e182068da403b86. The new `resetMemoryDataView` needs to be called from the Go side.
* faketime renames: https://github.com/golang/go/commit/4af3c17f8cd37288a5bb9bc53e8b88c58f5624d1. It doesn't appear that TinyGo has any of the `go` section, so `nanotime` and `walltime` don't exist to be renamed.
* garbage collect references to JavaScript values: https://github.com/golang/go/commit/54e6ba6724dfde355070238f9abc16362cac2e3d. As mentioned earlier, the type flag values changed.

deadprogram:
@QuLogic @aykevl @jaddr2line is that list of items the last outstanding items? And is the main decision to drop support for WASM on versions of Go less than 1.14? Just trying to understand what we need to do here.

deadprogram:
Now that PR #1035 has been merged, I think this PR is no longer needed, and can be closed @QuLogic

QuLogic:
Sorry, I got a bit too busy to get back to this; thanks for finishing it up @aykevl.

PR #902, opened 2020-02-13
Autodetect llvm-ar and llvm-nm commands

### LINKED ISSUE #882: "llvm-ar-9" does not exist on Arch/Manjaro
In #845 we added a build of wasi-libc to the tinygo build. The build uses the commands `llvm-ar-9` and `llvm-nm-9`, however on Arch/Manjaro these are called `llvm-ar` and `llvm-nm` instead. This can be worked around using environment variables for now, but we should add a workaround for this before we release.

jaddr2line:
This PR fixes #882 by checking for both variants of command names - with and without the version number suffix. This also uses a more correct detection method than was previously used for ccache.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,8 +9,24 @@ LLD_SRC ?= llvm-project/lld
 
 # Default tool selection.
 CLANG ?= clang-9
-LLVM_AR ?= llvm-ar-9
-LLVM_NM ?= llvm-nm-9
+
+# Try to autodetect llvm-ar and llvm-nm
+ifneq (, $(shell command -v llvm-ar-9 2> /dev/null))
+    LLVM_AR ?= llvm-ar-9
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -23,7 +39,7 @@ MD5SUM = md5sum
 TINYGO ?= tinygo
 
 # Use CCACHE for LLVM if possible
-ifneq (, $(shell which ccache))
+ifneq (, $(shell command -v ccache 2> /dev/null))
```
aykevl: What's the difference between `which` and `command -v`?

jaddr2line: The `command` is a shell builtin, whereas `which` is a command. Also adding the redirect solved the issue where it would print error messages if `ccache` is not installed.

aykevl:
Okay, thanks for clarifying! Merged now.

PR #904, opened 2020-02-16
Add support for Particle Argon, Boron and Xenon boards

suda:
Tested on all three devices:
* ✅ GPIO
* ✅ UART
* ✅ I2C

This PR also adds the ability to change the UART ports when calling `Configure` on nRF platform (used to test UART on different pins).

Flashed using [Particle Debugger](https://store.particle.io/collections/accessories/products/particle-debugger) which works with `cmsis-dap` profile in OpenOCD.


src/machine/board_particle_argon.go
```diff
@@ -0,0 +1,67 @@
+// +build particle_argon
+
+package machine
+
+const HasLowFrequencyCrystal = true
+
+// More info: https://docs.particle.io/datasheets/wi-fi/argon-datasheet/
+// Board diagram: https://docs.particle.io/assets/images/xenon/xenon-block-diagram.png
```
conejoninja: Wrong diagram

src/machine/board_particle_boron.go
```diff
@@ -0,0 +1,73 @@
+// +build particle_boron
+
+package machine
+
+const HasLowFrequencyCrystal = true
+
+// More info: https://docs.particle.io/datasheets/cellular/boron-datasheet/
+// Board diagram: https://docs.particle.io/assets/images/xenon/xenon-block-diagram.png
```
conejoninja: wrong diagram url

src/machine/board_particle_argon.go
suda: Doh! Dang copypasta...

suda:
✅ Added the GPIO pins
✅ Fixed some incorrect pins in Boron
✅ Updated to support changes from #883 and set [`machine.Serial`](https://github.com/tinygo-org/tinygo/pull/883#discussion_r376875736) to the USB-CDC

Once #884 gets merged in, I can also set the correct Particle USB VID/PID

conejoninja:
We should add them to the tinygo website too https://github.com/tinygo-org/tinygo-site/tree/master/content/microcontrollers

src/machine/board_particle_argon.go
```diff
@@ -0,0 +1,96 @@
...
+
+// ESP32 coprocessor
+const (
+	ESP32_TXD_PIN       Pin = 36
+	ESP32_RXD_PIN       Pin = 37
+	ESP32_CTS_PIN       Pin = 39
+	ESP32_RTS_PIN       Pin = 38
+	ESP32_BOOT_MODE_PIN Pin = 16
+	ESP32_WIFI_EN_PIN   Pin = 24
+	ESP32_HOST_WK_PIN   Pin = 7
```
aykevl: Is this coprocessor compatible with the NINA on the Arduino Nano 33 IoT?
Looking at the constants it doesn't look like it, but checking just to be sure.

src/machine/board_particle_boron.go
```diff
@@ -0,0 +1,99 @@
...
+	UBLOX_RXD_PIN      Pin = 36
+	UBLOX_CTS_PIN      Pin = 38
+	UBLOX_RTS_PIN      Pin = 39
+	UBLOX_RESET_PIN    Pin = 12
+	UBLOX_POWER_ON_PIN Pin = 16
+	UBLOX_BUFF_EN_PIN  Pin = 25
+	UBLOX_VINT_PIN     Pin = 2
+)
+
+// Other periferals
```
aykevl: typo: peripherals

src/machine/board_particle_boron.go
```diff
@@ -0,0 +1,99 @@
...
+const (
+	SPI1_SCK_PIN  Pin = 19
+	SPI1_MOSI_PIN Pin = 20
+	SPI1_MISO_PIN Pin = 21
+	SPI1_CS_PIN   Pin = 17
+	SPI1_WP_PIN   Pin = 22
+	SPI1_HOLD_PIN Pin = 23
+)
+
+// u-blox coprocessor
```
aykevl: Optional, but it might be useful to indicate this is the LTE modem (to avoid confusion with the u-blox coprocessor on the Arduino Nano 33 IoT).

src/machine/board_particle_xenon.go
```diff
@@ -0,0 +1,85 @@
...
+const (
+	SPI1_SCK_PIN  Pin = 19
+	SPI1_MOSI_PIN Pin = 20
+	SPI1_MISO_PIN Pin = 21
+	SPI1_CS_PIN   Pin = 17
+	SPI1_WP_PIN   Pin = 22
+	SPI1_HOLD_PIN Pin = 23
+)
+
+// Other periferals
```
aykevl: typo

src/machine/board_particle_argon.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+// Other periferals
```
aykevl: typo

targets/particle-argon.json
```diff
@@ -0,0 +1,6 @@
+{
+	"inherits": ["particle-3rd-gen"],
+	"build-tags": ["particle_argon"],
+	"flash-method": "openocd",
+	"openocd-interface": "cmsis-dap"
```
aykevl: Strictly speaking, these two lines are optional because they are already inherited from the `particle-3rd-gen` target. But you could keep them here too if you think that's easier to read.

src/machine/board_particle_boron.go
suda: True. All Borons are using u-blox SARA modules. I'll change the prefix to `SARA`

src/machine/board_particle_argon.go
suda: I don't think so. It's a bare ESP32 connected via serial with HW flow control.

conejoninja:
Good for me too, should we merge now and add the linker script later?

aykevl:
I'd like to know what @suda thinks before merging.

suda:
Oooh the script would be great but I agree with Daniel, we can decouple it into a separate PR.

conejoninja:
Thanks for this PR! Great job :+1:  now merging

aykevl:
@conejoninja for the next PR: to keep a clean history can you squash such commits? They're mostly going back and forth fixing things while the core change is really just one thing. Ideally it would have been one commit per board but I think that would be hard to change at a later time. A clean history is convenient for things like making a changelog and bisecting bugs.

conejoninja:
no problem! I always have doubts between rebase and squash, but I just left a note to myself to squash from now on. :)

aykevl:
There is no hard and fast rule, but in general I try to keep one commit per actual change:

* Changes that make sense on their own should be kept in separate commits.
* Commits that fix issues in previous commits in the same PR (after reviews etc.) are not independent changes and should ideally be squashed into those original commits.

Not everyone knows the git-fu to do such things so it's not a big problem, but it helps to keep the history clean when possible.

PR #905, opened 2020-02-18
main: update cmsis-svd module

aykevl:
This reverts back to the upstream repository which has merged all necessary updates (including https://github.com/posborne/cmsis-svd/pull/85 and https://github.com/posborne/cmsis-svd/pull/93).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,7 +9,7 @@
 	url = https://github.com/avr-rust/avr-mcu.git
 [submodule "lib/cmsis-svd"]
 	path = lib/cmsis-svd
-	url = https://github.com/tinygo-org/cmsis-svd
+	url = https://github.com/posborne/cmsis-svd
 [submodule "lib/compiler-rt"]
 	path = lib/compiler-rt
 	url = https://github.com/llvm-mirror/compiler-rt.git


deadprogram:
Are we certain that we do not want to perhaps instead update our fork to include all the latest upstream commits? This at least gives us a semblance of version control ourselves.

What do you think, @aykevl ? Otherwise I will just merge this.

aykevl:
Submodules pin a particular version, so I don't see a problem with using the upstream version.

Using the upstream repo seems better to me, but I have no strong opinions about it. I just want to avoid forking the cmsis-svd repository more than necessary.

firelizzard18:
As @aykevl says, a submodule is always pinned to a particular commit sha. If you run `git ls-files -s lib`, the object ID for each folder is the commit sha that folder should be checked out to:

```
160000 9fe411cef1cef5de58e5957b89760759de44e393 0       lib/CMSIS
160000 6624554c02b237b23dc17d53e992bf54033fc228 0       lib/avr
160000 7d9b4167010feed9e59ac45e56519180295b3813 0       lib/cmsis-svd
160000 5bc79797e1f9184f39e1bd30fac75d995b72cea3 0       lib/compiler-rt
160000 3ab39a9d457bfe627473ed0e03a7f1161d9e4f27 0       lib/nrfx
```

The URLs in `.gitmodules` are only relevant when you explicitly ask git to update the submodule to a different commit. So functionally, the URL change has no effect until someone does something like `git pull` in `lib/cmsis-svd` (and IIRC `git pull` in particular only works if `.gitmodules` also has `branch = <name>`).

deadprogram:
OK, now merging.

PR #907, opened 2020-02-19
reflect: implement NumMethod and Implements

aykevl:
These two functions are needed by the encoding/json package.

### DIFF
--- a/src/reflect/type.go
+++ b/src/reflect/type.go
@@ -146,6 +146,20 @@ func (t Type) Kind() Kind {
 	}
 }
 
+// isBasic returns true if (and only if) this is a basic type.
+func (t Type) isBasic() bool {
+	return t%2 == 0
+}
+
+// isNamed returns true if (and only if) this is a named type.
+func (t Type) isNamed() bool {
+	if t.isBasic() {
+		return t>>6 != 0
+	} else {
+		return (t>>4)%2 != 0
+	}
+}
+
 // Elem returns the element type for channel, slice and array types, the
 // pointed-to value for pointer types, and the key type for map types.
 func (t Type) Elem() Type {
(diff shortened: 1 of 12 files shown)


src/reflect/type.go
```diff
@@ -401,10 +422,49 @@ func (t Type) AssignableTo(u Type) bool {
 }
 
 func (t Type) Implements(u Type) bool {
-	if t.Kind() != Interface {
+	if u.Kind() != Interface {
 		panic("reflect: non-interface type passed to Type.Implements")
 	}
-	return u.AssignableTo(t)
+	if u.NumMethod() == 0 {
```
niaow, line 432: This is subtly incorrect, as an interface can include unexported methods. I am not sure though whether there is an easy way to solve this right now. If you wish to leave this for further work, consider adding a comment about it.

deadprogram: Relevant: https://github.com/golang/go/issues/22075

aykevl: Hmm this might be a bold thing to do but what if we reverted to the old (incorrect) behavior? It would make the implementation of `Implements` a lot easier, at least as long as we don't need to fully support it. The limited implementation here should be enough for packages like encoding/json.

deadprogram: By which you mean, the same behavior as Go currently has? I would agree with that, lets implement the same bugs :clown_face:

aykevl: Yes, the same buggy behavior (to be bug-compatible :wink:). Reading the bug report the final decision on whether the docs or the implementation should be fixed does not seem to be made yet.

aykevl: Fixing this is slightly less trivial than I expected: I discovered that anonymous interface types (like `var x interface{ F() }`) do not have the correct `NumMethod`.

deadprogram: What is the status with this one, @aykevl ?

aykevl: I need to look into this further, haven't done that yet.

deadprogram: Reminder that we still need to resolve this. :smile_cat:

aykevl:
I have updated this branch to be bug-compatible with Go but now it isn't complete enough for encoding/json, so this will need a bit more work.

aykevl:
I have implemented Implements in a slightly different way (at compile time). This PR might still be useful one day, but for now I've solved it in a different way. Closing, for now.

Issue #151, opened 2019-01-23
Volatile structs compile but do not work as expected from interrupt handlers


deadprogram:
I am trying to create a struct that wraps around some volatile members. The code compiles, but the interrupt handlers do not seem to be able to access them as expected, aka nothing seems to happen.

I tried this initially:

```go
//go:volatile
type volatileByte byte

type RingBuffer struct {
   rxbuffer [bufferSize]volatileByte
   head volatileByte
   tail volatileByte
}
```

and also this

```go
//go:volatile
type volatileByte byte

//go:volatile
type RingBuffer struct {
   rxbuffer [bufferSize]volatileByte
   head volatileByte
   tail volatileByte
}
```

And also this

```go
//go:volatile
type RingBuffer struct {
   rxbuffer [bufferSize]byte
   head byte
   tail byte
}
```

They increased the size of my BIN file considerably, but did not work as expected.

aykevl:
You could test it by adding `-printir` and finding the offending function. The relevant `load` and `store` should have the `volatile` flag, but probably doesn't.
Unlike most programming languages, IR puts volatile on loads and stores instead of on types.

The first code sample should be the correct one.

deadprogram:
This is some of the Go code:

```go
//go:volatile
type volatileByte byte

type RingBuffer struct {
	rxbuffer [bufferSize]volatileByte
	head     volatileByte
	tail     volatileByte
}

// NewRingBuffer returns a new ring buffer.
func NewRingBuffer() *RingBuffer { //buf [bufferSize]volatileByte) *RingBuffer {
	//return &RingBuffer{rxbuffer: buf}
	return &RingBuffer{}
}

// Used returns how many bytes in buffer have been used.
func (rb RingBuffer) Used() uint8 {
	return uint8(rb.head - rb.tail)
}

// Put stores a byte in the buffer.
func (rb RingBuffer) Put(val byte) {
	if rb.Used() != bufferSize {
		rb.head++
		rb.rxbuffer[rb.head%bufferSize] = volatileByte(val)
	}
}

//go:export SERCOM0_IRQHandler
func handleUART0() {
	// should reset IRQ
	UART0.Buffer.Put(byte((UART0.Bus.DATA & 0xFF)))
	UART0.Bus.INTFLAG |= sam.SERCOM_USART_INTFLAG_RXC
}
```

and some of the IR generated:

```
[...]

PR #908, opened 2020-02-20
machine: remove old comment on volatile behavior

aykevl:
The volatile pragma has long since been replaced by builtins, so this commentis now outdated.

See: https://github.com/tinygo-org/tinygo/issues/151#issuecomment-588023760


deadprogram:
Thanks, now merging.

PR #910, opened 2020-02-22
move constant pointer packs into constant globals

jaddr2line:


jaddr2line:
For the most part, this is an improvement. However, this prevents strings from being optimized away from dead code. That requires a change to the GC pass which will be addressed in a separate PR.

compiler/llvmutil/wordpack.go
```diff
@@ -42,6 +42,24 @@ func EmitPointerPack(builder llvm.Builder, mod llvm.Module, config *compileopts.
...
+		for _, v := range values {
+			if !v.IsConstant() {
+				constant = false
+				break
+			}
+		}
+
+		if constant {
+			// The data is known at compile time, so store it in a constant global.
+			global := llvm.AddGlobal(mod, packedType, "")
```
aykevl: This should have a real name based on where it is declared. For example:

https://github.com/tinygo-org/tinygo/blob/1d913a62bc0ceef841953058c2c6fb1dd2a40f6a/compiler/compiler.go#L2214-L2215

compiler/llvmutil/wordpack.go
```diff
@@ -42,6 +42,24 @@ func EmitPointerPack(builder llvm.Builder, mod llvm.Module, config *compileopts.
...
+			global.SetInitializer(ctx.ConstStruct(values, false))
+			global.SetGlobalConstant(true)
```
aykevl: None of the uses of `EmitPointerPack` rely on this pointer being unique (I checked). Therefore you can mark this `unnamed_addr` using `SetUnnamedAddr`. This will let the compiler merge identical globals.

jaddr2line:
@aykevl I have addressed your comments.

Issue #716, opened 2019-11-12
Wasm callback functions working with Go 1.12, but not working with Go 1.13


justinclift:
Found an interesting regression (?).  Callback creation code that works with TinyGo 0.9.0 + Go 1.12, doesn't work when used with Go 1.13.

For example, the canvas demo in @johanbrandhorst's [wasm-experiments](https://github.com/johanbrandhorst/wasm-experiments) repo:

* https://github.com/johanbrandhorst/wasm-experiments/blob/99f4e0a7f0ddf0215dd0eaae8f7f210767c5bba6/canvas/main.go#L36-L41
* https://github.com/johanbrandhorst/wasm-experiments/blob/99f4e0a7f0ddf0215dd0eaae8f7f210767c5bba6/canvas/main.go#L69

The error given on the browser console is:

```
panic: syscall/js: Value.Call: property _makeFuncWrapper is not a function, got undefined
```

Not sure what to err... make of this, as `_makeFuncWrapper()`  is a function included in the TinyGo wasm_exec.js, and is a literal copy of the [mainline Go](https://github.com/golang/go/blob/54e6ba6724dfde355070238f9abc16362cac2e3d/misc/wasm/wasm_exec.js#L550-L557) one.

Anyone have ideas?

OneOfOne:
Minimal repro:

```
package main

import (
	"syscall/js"
)

func main() {
	js.Global().Get("document").Call("onLoad", js.FuncOf(fn))
	select {}
}

func fn(this js.Value, args []js.Value) interface{} {
	return ""
}
```

justinclift:
Good thinking @OneOfOne.:smile:

ellemlabs:
Confirming. I'm using TinyGo (tag v0.11) with Go (v1.13.5) I get the same error. It doesn't seem to be exclusive to callbacks, but the use of any sort of javascript interaction in general:

```
package main

import "syscall/js"

func test(this js.Value, inputs []js.Value) interface{} {
	println("test")
	return nil
}

func main() {
	js.Global().Set("testfunc", js.FuncOf(test)) 
}
```

aykevl:
`js.FuncOf` essentially creates a callback, so yes it is still caused by callbacks.

OneOfOne:
1.14 is almost out, the issue is still there.

jfreymuth:
Line 335 in tinygo's `wasm_exec.js` doesn't exist in the mainline version. Removing that line makes the callback example work for me.

[...]

PR #911, opened 2020-02-23
targets/wasm: update wasm_exec.js for go 1.13

jfreymuth:
See #716


aykevl:
Thank you for the fix!
It's unfortunate that this breaks older Go versions but the tests still pass and supporting current Go versions is definitely more important than older Go versions.

PR #913, opened 2020-02-24
Do not track const globals

jaddr2line:
We are currently tracking constant globals with the garbage collector, which is incorrect behavior. If a value is pointed to by a constant global, then it must either be another global or a constant user-defined address (which is not part of the heap). I discovered this because it was breaking optimizations in #910, and caused a size regression.


aykevl:
I tested the files in testdata/ and found that some were indeed a bit smaller (tests for channel, coroutines, and interface). Also, thanks for the nice tests.

Merging now.

aykevl:
This should have been obvious, but apparently I missed this optimization.

Issue #912, opened 2020-02-24
Calling yield from sleepTicks causes a stack overflow, but panics on nil pointer dereference


firelizzard18:
I am working on implementing a target for the Teensy 3.6 / NXP MK66 (#863). My initial implementation of `sleepTicks` called `Gosched`. When there is only one user routine, and that routine calls `time.Sleep` (`runtime.sleep`), this will cause a stack overflow (`sleep` => `yield` => `sleepTicks` => `Gosched` => `yield` => ...). **In this particular scenario, `currentTask` becomes `nil` and `yield` panics with `panic: runtime error: nil pointer dereference`.**

Obviously `sleepTicks` should not call `Gosched` or `yield`. But if that happens, `fatal error: stack overflow` is the proper response, not `panic: runtime error: nil pointer dereference`.

https://github.com/tinygo-org/tinygo/blob/c0b8716d7f2cf33bcbb7e452e84260228a4e2b32/src/runtime/scheduler_tasks.go#L100-L107

https://github.com/tinygo-org/tinygo/blob/27da0827f3653da6eb6be1363745cfac52990553/src/runtime/runtime_nxpmk66f18.go#L317-L338

jaddr2line:
You cannot call yield from inside sleepticks. It must be called from inside a goroutine.

firelizzard18:
@jaddr2line Yes, I understand that now, as I said:

> Obviously sleepTicks should not call Gosched or yield.

As I also said:

> But if that happens, `fatal error: stack overflow` is the proper response, not `panic: runtime error: nil pointer dereference`.

This issue is not about my incorrect code. This issue is about a secondary error that hid the real error in my code.

jaddr2line:
There is no stack overflow involved. There is no recursion. The `yield` call is supposed to switch back to the scheduler, which then continues doing whatever. This behavior is correct. You cannot call `yield` except from a goroutine, and in this case it is called from the scheduler.

jaddr2line:
See: https://github.com/tinygo-org/tinygo/blob/c0b8716d7f2cf33bcbb7e452e84260228a4e2b32/src/runtime/scheduler.go#L266

firelizzard18:
If there is no stack overflow, why does that error appear?

jaddr2line:
The nil pointer error appears because there is no current goroutine when on the scheduler

firelizzard18:
[...]

PR #914, opened 2020-02-24
Warn about unsafe calls to yield

firelizzard18:
As I discovered while working on #863 (#912 has additional details), calling `yield` from `sleepTicks` is unsafe due to possible infinite recursion and/or nil pointer dereference, and calling `yield` from `putchar` is unsafe because the latter is called in contexts where the former is not safe to call. In my specific circumstances, calling `yield` from `putchar` meant I did not see panic messages on the console, when panics occurred.

This change adds some call graph analysis that checks if the function(s) of concern, `yield`, is called directly or indirectly by the target functions, `sleepTicks` or `putchar`. This is done by recursively constructing a call graph of all the callees of the target functions and 'coloring' all nodes that can reach a/the function of concern. If `sleepTicks` or `putchar` are colored, there is a path from them to `yield`, and a warning is issued. This algorithm works for arbitrary functions of concern and target functions.

The warnings appear thus:

```
unsafe call to runtime.yield from runtime.sleepTicks: yield may call sleepTicks, which could lead to a stack overflow
       runtime.sleepTicks calls
       runtime.Gosched calls
       runtime.yield
unsafe call to runtime.yield from runtime.putchar: putchar must work in contexts where calling the scheduler is not safe
       runtime.putchar calls
       (*machine.UART).WriteByte calls
       runtime.Gosched calls
       runtime.yield
```

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -56,6 +56,8 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(stri
 		c.ApplyFunctionSections() // -ffunction-sections
 	}
 
+	c.CheckInvalidYield()
+
 	// Browsers cannot handle external functions that have type i64 because it
 	// cannot be represented exactly in JavaScript (JS only has doubles). To
 	// keep functions interoperable, pass int64 types as pointers to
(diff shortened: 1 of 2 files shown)


jaddr2line:
I am going to close this since `runtime.yield` has been replaced.

PR #915, opened 2020-02-24
main: check for error after starting the OCD daemon

aykevl:
This replaces a panic with an informative error message if the right tool (such as OpenOCD) is not installed.


conejoninja:
Looks good to me, thanks @aykevl

aykevl:
Thank you!

PR #916, opened 2020-02-24
main: use gdb-multiarch for debugging Cortex-M chips

aykevl:
Previously this was set to `arm-none-eabi-gdb`, but that has been replaced by `gdb-multiarch` in recent Debian/Ubuntu versions.


conejoninja:
Looks like the change was made in Ubuntu 18.04, I think it's safe to change it.

PR #917, opened 2020-02-24
Mark abort as noreturn

jaddr2line:
This marks the libc function `abort` as non-returning. This allows LLVM to optimize away code after panics. Also, this allows deadlocks to be properly propogated with the `coroutines` scheduler.


aykevl:
It would be useful to include the text of this PR in the commit message. That helps whoever looks at the git log figuring out what happened in the past. Also, can you format the first commit line to include the relevant package/section of the code (in this case `compiler:`)?
See [Good commit messages](https://tip.golang.org/doc/contribute.html#commit_messages) for an explanation.

It's no big deal if this is not always followed, but it helps to include it.

aykevl:
I have merged it with these changes to the commit message, I hope that's okay.

PR #919, opened 2020-02-25
Fault status decoding

firelizzard18:
These changes add functions to decode the fault status register(s) and update the hard fault handler with more detailed error messages. They were motivated by a bus error that was reported as a stack overflow, as well as by this TODO found in the hard fault handler:

https://github.com/tinygo-org/tinygo/blob/1d913a62bc0ceef841953058c2c6fb1dd2a40f6a/src/runtime/runtime_cortexm.go#L106-L109

Each new error string will add to binary sizes, but the fault decoding itself adds almost nothing. `GetFaultStatus()` compiles to a register load and each call such as `fault.Mem().InstructionAccessViolation()` compiles to a single instruction (usually a logical shift left `lsls`).

I discovered the incorrectly reported bus error when I tried to write to a register on a disabled peripheral. The Kinetis K66 uses clock gating to reduce power draw from unused peripherals. If a register on a peripheral is read or written before the peripheral's clock is enabled via the appropriate System Clock Gating Control register (SCGCx), the CPU raises a bus error and calls the hard fault handler. At least for the K66, this results in an "imprecise data bus error", which means that the BusFault Address Register and the stack pointer are invalid. This invalid stack pointer is interpreted by the existing handler as a stack overflow.

At a minimum, TinyGo should check the fault status to determine whether or not the stack pointer can be considered valid. This should prevent the hard fault handler from incorrectly reporting a bus error as a stack overflow. Writing to a register of a disabled peripheral is definitely something an end-user of TinyGo might do, if they are not a microcontroller expert and trying to use a peripheral directly.

### DIFF
--- a/src/device/arm/core_cortexm0.go
+++ b/src/device/arm/core_cortexm0.go
@@ -1,5 +1,0 @@
-// +build sam,atsamd21 nrf,nrf51
-
-package arm
-
-const Core = CortexM0
--- a/src/device/arm/core_cortexm0.go
+++ b/src/device/arm/core_cortexm0.go
@@ -1,0 +1,5 @@
+// +build sam,atsamd21 nrf,nrf51
+
+package arm
+
+const Core = CortexM0
--- a/src/runtime/runtime_cortexm.go
+++ b/src/runtime/runtime_cortexm.go
@@ -4,11 +4,12 @@ package runtime
 
 import (
 	"device/arm"
+	"machine"
 	"unsafe"
 )
 
-const showFaultAddresses = true
-const verboseFaultMessage = true
+const showFaultAddresses = false
+const verboseFaultMessage = false
 
 //go:extern _sbss
 var _sbss [0]byte
@@ -104,21 +105,7 @@ type interruptStack struct {
 func handleHardFault(sp *interruptStack) {
 	print("fatal error: ")
 
-	if arm.Core == arm.CortexM0 {
-		if uintptr(unsafe.Pointer(sp)) < 0x20000000 {
-			print("stack overflow")
-		} else {
-			print("HardFault")
-		}
-
-		print(" with sp=", sp)
-		if uintptr(unsafe.Pointer(&sp.PC)) >= 0x20000000 {
-			// Only print the PC if it points into memory.
-			// It may not point into memory during a stack overflow, so check that
-			// first before accessing the stack.
-			print(" pc=", sp.PC)
-		}
-	} else if arm.Core == arm.CortexM3 || arm.Core == arm.CortexM4 {
+	if machine.CPUArchitecture == "armv7m" || machine.CPUArchitecture == "armv7em" || machine.CPUArchitecture == "armv8m" {
 		fault := arm.GetFaultStatus()
 		validSP := !fault.Bus().ImpreciseDataBusError()
 
@@ -150,6 +137,21 @@ func handleHardFault(sp *interruptStack) {
 				// accessing the stack.
 				print(" pc=", sp.PC)
 			}
+		}
+
+	} else {
+		if uintptr(unsafe.Pointer(sp)) < 0x20000000 {
+			print("stack overflow")
+		} else {
+			print("HardFault")
+		}
+
+		print(" with sp=", sp)
+		if uintptr(unsafe.Pointer(&sp.PC)) >= 0x20000000 {
+			// Only print the PC if it points into memory.
+			// It may not point into memory during a stack overflow, so check that
+			// first before accessing the stack.
+			print(" pc=", sp.PC)
 		}
 	}
 
--- a/src/runtime/runtime_cortexm.go
+++ b/src/runtime/runtime_cortexm.go
@@ -7,7 +7,8 @@ import (
 	"unsafe"
 )
 
-const verboseFaultMessage = false
+const showFaultAddresses = true
+const verboseFaultMessage = true
 
 //go:extern _sbss
 var _sbss [0]byte
@@ -103,8 +104,7 @@ type interruptStack struct {
 func handleHardFault(sp *interruptStack) {
 	print("fatal error: ")
 
-	switch arm.Core {
-	case arm.CortexM0:
+	if arm.Core == arm.CortexM0 {
 		if uintptr(unsafe.Pointer(sp)) < 0x20000000 {
 			print("stack overflow")
 		} else {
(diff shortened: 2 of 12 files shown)


aykevl:
Quick testing shows that this increases code size by almost 1kB, which is huge. It also tries to access the fault handler on Cortex-M0, which is a problem. I think the code should stay more or less the same for Cortex-M0 as it doesn't offer many ways to discover what went wrong anyway.

What I would suggest is the following:

1. Provide a way (in the machine package?) to detect the CPU type that is then provided as a constant string (for example `machine.CPU`). The information can be obtained from SVD files (mostly, AFAIK stm32 doesn't have it) but it may be easier to put it directly in the relevant CPU family files (like src/machine/machine_atsamd51.go). Comparisons between two constant strings are trivially optimized away so can be used for CPU detection. (Sidenote: providing the CPU type in the machine package would also be useful for the ws2812 driver).
2. Use this information in the fault handler to have roughly two code paths: one for the Cortex-M0 which simply checks whether the stack pointer looks reasonable, and one for Cortex-M3 and up which has a proper check for a data bus error.
3. Instead of printing each fault as a string, print the raw fault bits. For example: `print(" CFSR=", uintptr(arm.SCB.CFSR.Get()))`. A quick google for "arm cfsr" shows the [official manual](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0552a/Cihcfefj.html) from where it is not hard to figure out the fault if you know how bits are stored in a hexadecimal number.

This is far less informative than printing the fault as a human-readable string, but it does provide the tools to debug this further without costing so much in terms of code size. Remember that this is included in every build and while 1kB might not sound like much, many of these features together add up quickly.

Sidenote: the tests fail because src/device/arm/fault.go is not formatted correctly (run `make fmt` to fix).

firelizzard18:
What about a normally false constant flag to print out full messages? In the usual case, the user will get a minimal error but binary size won’t increase much, but the flag will facilitate easier debugging.

My IDE (VSCode + vscode-golang + gopls) really doesn’t like TinyGo, so the automatic go fmt does not work consistently. I’ll probably add a local pre-commit hook to run make fmt.

firelizzard18:
I expected less size differences with `verboseFaultMessage = false`. I'll work on that tomorrow.

```
main-old0.hex 25992 bytes # baseline for feather-m0
main-m0.hex   26080 bytes # feather-m0 with new fault handler

main-old.hex  21768 bytes # baseline for feather-m4
main-new.hex  22480 bytes # feather-m4 with new fault handler
main-full.hex 24680 bytes # feather-m4 with new fault handler in verbose mode
```

firelizzard18:
As of the latest commit, there's no change for M0, and for M4:

  * Default: +184 bytes
  * With `showFaultAddresses`: +360 bytes
  * With `verboseFaultMessage`: +2472 bytes
  * With both flags: +2648 bytes

I discovered a couple of interesting things: if-statements comparing constant strings will optimize away, but switch statements will not; and given the three cases below, cases 1 and 2 reduce to effectively the same thing, but case 3 adds an extra 44 bytes and the generated code complexity increases noticeably. Switching which if-statement uses `Address()` makes no difference.

Case 1:
```golang
if fault.Mem().ValidAddress() {
	print(" mmar=", uintptr(arm.SCB.MMAR.Get()))
}
if fault.Bus().ValidAddress() {
	print(" bfar=", uintptr(arm.SCB.BFAR.Get()))
}
```

Case 2:
```golang
if addr, ok := fault.Mem().Address(); ok {
	print(" mmar=", addr)
}
if fault.Bus().ValidAddress() {
	print(" bfar=", uintptr(arm.SCB.BFAR.Get()))
}
```

Case 3:
```golang
if addr, ok := fault.Mem().Address(); ok {
	print(" mmar=", addr)
}
if addr, ok := fault.Bus().Address(); ok {
	print(" bfar=", addr)
}
```

aykevl:
> I discovered a couple of interesting things: if-statements comparing constant strings will optimize away, but switch statements will not

Huh that's interesting, I guess this is simply missing in the golang.org/x/tools/go/ssa package. However, it should not be very difficult to implement an optimization in TinyGo for this purpose. Note that once the code is in SSA form, there is no distinction between `if` and `switch`, everything is done with `goto` as that makes optimizations easier to implement.

> given the three cases below, cases 1 and 2 reduce to effectively the same thing, but case 3 adds an extra 44 bytes and the generated code complexity increases noticeably. Switching which if-statement uses `Address()` makes no difference.

This is likely related to some optimization triggering or not (I'd guess inlining).

src/device/arm/core_cortexm0.go
```diff
@@ -0,0 +1,5 @@
+// +build sam,atsamd21 nrf,nrf51
+
+package arm
+
+const Core = CortexM0
```
aykevl: Unfortunately, that's not entirely correct. The `atsamd21` is a Cortex-M0+ instead of a Cortex-M0. This is relevant because there are some small differences between the two cores (such as cycles per branch and privilege levels).

Perhaps more importantly, I do not think this is the right place for defining the `Core` constant. I think the right place is in the machine package in files for specific boards: that will reduce the number of places modifications need to be made to add support for a new board.

Thinking a bit longer about this, maybe simply providing a 'core' is not good enough:

* For fault status, what you're really interested in is the architecture (armv6m, armv7m, armv7em, armv8m.
* The runtime might be interested in which extensions are supported (such as floating point, or stack pointer limit checking in armv8m). It should preferably not need to know about the specific core.
* For timing sensitive operations, you need to know the exact core (to know the number of cycles an instruction takes).

firelizzard18: My motivation for putting the `Core` constant (and/or other constants) in the `arm` package is how it looks to the developer using TinyGo (vs devs working on TinyGo). I don't think a generic constant, e.g. `machine.CPU`, is the right choice for anything ARM-specific, so I would use `machine.ARMCore` etc. And IMO `arm.<Const>` (e.g. Core, Architecture, Extensions) flows better than `machine.ARM<Const>`, and better matches my expectations as a developer.

What about using `targets.json`? `llvm-target`, which is required, looks like it mostly matches architecture, and `cpu`, when it is used, seems to match the exact core. The only outlier is the GBA which defines `llvm-target` as `thumb4-none-eabi` and `cpu` as `arm7tdmi`. The compiler could use these, and possibly new `TargetSpec` fields, to (re)define constants.

firelizzard18:
Well, clearly I misunderstood the CPU field. That being said, I successfully implemented compiler-defined constants, though it's not pretty. I'd much rather manipulate the SSA or IR than inject a file, but all my attempts at the former failed miserably.

src/device/arm/core_cortexm0.go
firelizzard18: To expand on "better matches my expectations as a developer," I think it stems from the fact that the `machine` package is, in large part, a hardware abstraction layer, whereas the `device/*` packages provide low-level access to the microcontroller. So, "What architecture am I running on?" feels like far more of a low-level `device/*` kind of question than a `machine` hardware abstraction layer question.

One possibility is to create a `device/cpu` package and add the constants to that.

src/runtime/runtime_cortexm.go
```diff
@@ -100,21 +104,57 @@ type interruptStack struct {
 //go:export handleHardFault
 func handleHardFault(sp *interruptStack) {
 	print("fatal error: ")
-	if uintptr(unsafe.Pointer(sp)) < 0x20000000 {
-		print("stack overflow")
+
+	if machine.CPUArchitecture == "armv7m" || machine.CPUArchitecture == "armv7em" || machine.CPUArchitecture == "armv8m" {
```
firelizzard18: I first tried `if a := machine.CPUArchitecture; a == "armv7m" || a == "armv7em" || a == "armv8m" { ... }`, but the compiler did not optimize away the branch.

PR #925, opened 2020-02-26
main: increment version to 0.13.0-dev

deadprogram:
This PR updates the tinygo version to the current dev version. Better late then never!


aykevl:
...whoops

PR #926, opened 2020-02-26
compiler: fix deferred calls to exported functions

jaddr2line:
Previously using defer with an exported function generated an invalid function call due to differences between TinyGo's calling convention and the C calling convention.


bgould:
I tried this out, and it fixes my use case for https://github.com/bgould/go-littlefs/blob/master/go_lfs.go#L150

aykevl:
Great, looks good to me. Thanks for the quick fix!

PR #928, opened 2020-02-27
Makefile: fix LLVM tools detection for wasi-libc

aykevl:
This makes the `make wasi-libc` command much more reliable and makes the CI configuration simpler. Also, it avoids warnings when they are not relevant.

Now let's hope CI will accept this.


aykevl:
Thank you for the review!

PR #934, opened 2020-02-29
wasm: add runtime.NumCPU

justinclift:
Noticed this function call is missing, and - at least for wasm - it's trivial to implement. :wink:

### DIFF
--- a/src/runtime/runtime_wasm.go
+++ b/src/runtime/runtime_wasm.go
@@ -79,3 +79,9 @@ func ticks() timeUnit
 func abort() {
 	trap()
 }
+
+// WebAssembly supports only a single thread (at present)
+//go:export NumCPU()
+func NumCPU() int {
+	return 1
+}


aykevl:
Actually, it might make sense to use this implementation for all targets at the moment. While TinyGo definitely runs on hardware with more than one CPU, it does not support any form of parallelism right now. Returning 1 in `NumCPU` is lying a bit but should be fine for most use cases.

Is there something you need this function for?

The tests failed because the code is not formatted, try running `make fmt`.

justinclift:
> Is there something you need this function for?

Nothing in particular, I was just trying out some existing (mainline) Go code that uses WaitGroups and it also complained about this.  While I'm not going to mess with WaitGroups + TinyGo any time soon, this NumCPU piece seemed like an easy fix. :wink:

> The tests failed because the code is not formatted, try running make fmt.

Thanks, will do. :smile:

justinclift:
Just ran `goimports` on the source file.  Was hoping that'd let this pass, but it seems like CircleCI is having infrastructure issues?

deadprogram:
I think you need to rebase against the latest `dev`.

justinclift:
Rebased, lets see what happens now. :smile:

justinclift:
Closing this, as it's just not worth the time to investigate and fix whatever is going wrong.

PR #935, opened 2020-02-29
main: move some duplicate code to a single place

aykevl:
This makes fixes like https://github.com/tinygo-org/tinygo/pull/915 easier to apply (and in fact fixes a similar issue with the mgba debugger support). And in general, avoiding duplication is a good thing.

I tested this on the GBA, microbit, pca10040, cortex-m-qemu, hifive1b, and host targets.

### DIFF
--- a/main.go
+++ b/main.go
@@ -313,6 +313,7 @@ func FlashGDB(pkgName string, ocdOutput bool, options *compileopts.Options) erro
 
 		// Run the GDB server, if necessary.
 		var gdbCommands []string
+		var daemon *exec.Cmd
 		switch gdbInterface {
 		case "native":
 			// Run GDB directly.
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -363,66 +351,43 @@ func FlashGDB(pkgName string, ocdOutput bool, options *compileopts.Options) erro
...
 			gdbCommands = append(gdbCommands, "target remote :2345")
 
 			// Run in an emulator.
 			args := append(config.Target.Emulator[1:], tmppath, "-g")
-			daemon := exec.Command(config.Target.Emulator[0], args...)
+			daemon = exec.Command(config.Target.Emulator[0], args...)
 			daemon.Stdout = os.Stdout
 			daemon.Stderr = os.Stderr
+		case "msd":
+			return errors.New("gdb is not supported for drag-and-drop programmable devices")
```
deadprogram: How can you run GDB on devices such as ItsyBitsy-M0 with the debug pins attached to JLink device?

aykevl: By adding the `openocd-interface` and `openocd-target` keys to the JSON file (or I think with the `-programmer=jlink` flag). This PR doesn't change any of that, the change here is that some code got moved around.

deadprogram: Ah, OK. Thanks for the clarification.

deadprogram:
Now merging.

PR #936, opened 2020-03-01
[pyboard] replaced BUTTON_*_MASK by BUTTON_* to fit the new shifter

conejoninja:
driver


deadprogram:
It occurs to me that we could leave in the `_MASK` values and just add the `_BUTTON` values. Seems like it would clarify intent, and also not break any existing code.

deadprogram:
Just thinking that some comment about what these values are would be helpful.

conejoninja:
BUTTON_* are now in the shifter driver, are still needed?

deadprogram:
I think you are correct, this can just be closed now.

PR #937, opened 2020-03-01
arm: fix linker script

aykevl:
There were a few instances like `.text` and `.text*`. The first was redundant with the second, but the intention was to write `.text.*`.

This doesn't change anything (tested with `make smoketest`) but should avoid propagating this error in the future.

See: https://github.com/tinygo-org/tinygo/pull/863#discussion_r386060173


deadprogram:
Sounds good, now merging.

PR #938, opened 2020-03-01
added support for ADC1 on atsamd51 (PB04 / PB05 / PB06 / PB07)

conejoninja:
I tested the PB04 (A7 on the pybadge), according to the datasheet PB05-PB06-PB07 should work too but I can't say I fully understand it ¯\_(ツ)_/¯


deadprogram:
We need those pins! Now merging, thanks @conejoninja

PR #939, opened 2020-03-03
main: improve error reporting while running go test

aykevl:
This commit switches integration tests to use the same error reporting
mechanism as the tinygo compiler normally uses. It replaces errors like
this:

    main_test.go:139: failed to build: interp: branch on a non-constant

With this:

    main.go:693: # math/rand
    main.go:695: interp: branch on a non-constant

In this particular case the error isn't much better (it gives the
relevant package, though) but other errors should also include the
source location where they happen.

### DIFF
--- a/main.go
+++ b/main.go
@@ -675,38 +675,49 @@ func usage() {
 	flag.PrintDefaults()
 }
 
+// printCompilerError prints compiler errors using the provided logger function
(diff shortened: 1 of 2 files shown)


deadprogram:
This is for sure a good thing, it should help a lot with compatibility testing. Now merging.

PR #942, opened 2020-03-06
cgo: make -I and -L paths absolute

aykevl:
This is very useful for (conditionally) adding extra include paths relative to the package path.


deadprogram:
Makes good sense to me. Now merging.

PR #945, opened 2020-03-09
avr: use the correct RAM start address

aykevl:
Previously, the RAM was set to start at address 0. This is incorrect: on AVR, the first few addresses are taken up by memory-mapped I/O. The reason this didn't lead to problems (yet) was because the stack was usually big enough to avoid real problems and was put at the start of the RAM (including the I/O space).


aykevl:
Superseded by #946, which also includes this patch.

PR #946, opened 2020-03-10
Add support for the ATmega1284

aykevl:
This chip has 16kB of RAM, which is quite a lot for an AVR. This makes it useful for testing in simulation, for example using `simavr`.

This PR also includes two other small fixes I found while working on this PR.


deadprogram:
I do not have one of these to test with but looks ok to me. Now merging.

PR #947, opened 2020-03-10
runtime: fix channels on AVR

aykevl:
This commit makes channel code compilable on AVR, for when goroutines are supported on that platform. I have in fact managed to get initial support for channels working with this patch on a local branch that adds support for the stack-based scheduler.

### DIFF
--- a/src/runtime/panic.go
+++ b/src/runtime/panic.go
@@ -50,6 +50,11 @@ func slicePanic() {
 	runtimePanic("slice out of range")
 }
 
+// Panic when trying to create a new channel that is too big.
+func chanMakePanic() {
+	runtimePanic("new channel is too big")
+}
+
 func blockingPanic() {
 	runtimePanic("trying to do blocking operation in exported function")
 }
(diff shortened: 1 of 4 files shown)


aykevl:
I think this is slightly wrong (but close enough to correct that it should be fine). I will look into fixing this properly, in a similar way as slice bounds were fixed.

aykevl:
Fixed it more universally. Unfortunately, it appears that CircleCI builds are now broken.

aykevl:
Rebased to fix CI (after #950 got merged).

src/runtime/panic.go
```diff
@@ -50,6 +50,11 @@ func slicePanic() {
 	runtimePanic("slice out of range")
 }
 
+// Panic when trying to create a new channel that is too big.
+func chanMakePanic() {
+	runtimePanic("new channel too big")
```
jaddr2line:
```suggestion
	runtimePanic("new channel is too big")
```

aykevl:
Updated.

deadprogram:
Merging.

PR #949, opened 2020-03-11
runtime: return the correct type from the copy builtin

aykevl:
The copy builtin is defined as follows by the Go language spec:

    copy(dst, src []T) int
    copy(dst []byte, src string) int

In other words, it returns an `int`. The `runtime.sliceCopy` compiler intrinsic returned a `uintptr` instead, which led to a problem while compiling the strings package for AVR.

No other architecture should be affected by this change as the conversion from an `uintptr` to an `int` is a no-op on most architectures.


deadprogram:
OK, merging.

PR #950, opened 2020-03-11
ci: move Go version tests to buster containers

aykevl:
This fixes an error like the following:

    E: The method driver /usr/lib/apt/methods/https could not be found.
    N: Is the package apt-transport-https installed?

Apparently apt.llvm.org has been switched over to HTTPS. One solution could be to install apt-transport-https, but another (easier) solution is to switch to a newer container.

Note: I did not switch the other containers, to make sure TinyGo is still built with an older Debian release. That ensures the resulting binaries are relatively portable across distros, even relatively old distros.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ commands:
       - run:
           name: "Install apt dependencies"
           command: |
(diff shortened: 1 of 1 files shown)


deadprogram:
OK, understood that we need this to pass the CI. Merging.

Issue #724, opened 2019-11-15
Support Arduino MEGA 2560


kgolding:
Please add support for the Ardunio MEGA 2560 boards, as this board provides a larger number of GPIO's than most of the other boards (54 off).

https://www.arduino.cc/en/Main/arduinoBoard_Mega2560?setlang=en

ofauchon:
Hi. 

I found ATMega2560's declaration in tinygo sources : 

```
tinygo/src/device/avr/atmega2560.s
tinygo/src/device/avr/atmega2560.ld
tinygo/src/device/avr/atmega2560.go
```
So I tried to create a new local json target "arduino-mega2560.json" 

```
{
	"inherits": ["avr"],
	"llvm-target": "avr-atmel-none",
	"cpu": "atmega2560",
	"build-tags": ["arduino", "atmega2560", "atmega", "avr5"],
	"cflags": [
		"-mmcu=atmega2560"
	],
	"ldflags": [
		"-Wl,--defsym=_bootloader_size=512",
		"-Wl,--defsym=_stack_size=512"
	],
	"linkerscript": "src/device/avr/atmega2560.ld",
	"extra-files": [
		"targets/avr.S",
		"src/device/avr/atmega2560.s"
	],
	"flash-command": "avrdude -c arduino -p atmega256 -P {port} -U flash:w:{hex}"
```

... and run a simple build: 

```
$ tinygo build -target ../arduino-mega2560.json  -o main.bin main.go
/usr/bin/avr-ld: avr:6 architecture of input file `/tmp/tinygo576914937/main.o' is incompatible with avr output
/usr/bin/avr-ld: avr:6 architecture of input file `/tmp/tinygo576914937/extra-0-avr.S.o' is incompatible with avr output
/usr/bin/avr-ld: avr:6 architecture of input file `/tmp/tinygo576914937/extra-1-atmega2560.s.o' is incompatible with avr output
```
Some more debugs :

```
$ /usr/bin/avr-ld -plugin /usr/lib/gcc/avr/9.2.0/liblto_plugin.so -plugin-
opt=/usr/lib/gcc/avr/9.2.0/lto-wrapper -plugin-opt=-fresolution=/tmp/ccQJV7Rh.res 
-plugin-opt=-pass-through=-lgcc -plugin-opt=-pass-through=-lm -plugin-opt=-pass-through=-lc 
[...]

PR #951, opened 2020-03-12
Feature/mega2560

tmclane:
Initial pin / register / port mapping support for Arduino Mega2560.

With these changes `src/examples/blinky1` works perfectly.  I do not have a setup yet to test the rest of the pins since it seems my discrete components (leds, resistors, and the like) are packed away somewhere.  

I used the real pin numbers for all of the pins to be consistent since otherwise the A0-A15 would end up conflicting with D0-D15.  This helped with the math for offsets in some cases.  Unfortunately some of the pin offsets in the masks are semi-random.  I handled those individually where appropriate.

Note:  This is a draft since I do not believe it is complete as of yet.

Addresses #724

### DIFF
--- a/src/machine/board_arduino_mega2560.go
+++ b/src/machine/board_arduino_mega2560.go
@@ -8,7 +8,8 @@ func CPUFrequency() uint32 {
 }
 
 const (
-	LED Pin = PB7
+	AREF Pin = NoPin
+	LED  Pin = PB7
 
 	A0  Pin = PF0
 	A1  Pin = PF1
--- a/src/machine/board_arduino_mega2560.go
+++ b/src/machine/board_arduino_mega2560.go
@@ -1,0 +1,103 @@
+// +build arduino_mega2560
+
+package machine
+
+// Return the current CPU frequency in hertz.
+func CPUFrequency() uint32 {
+	return 16000000
+}
+
+const (
+	LED Pin = PB7
+
+	A0  Pin = PF0
+	A1  Pin = PF1
+	A2  Pin = PF2
+	A3  Pin = PF3
+	A4  Pin = PF4
+	A5  Pin = PF5
+	A6  Pin = PF6
+	A7  Pin = PF7
+	A8  Pin = PK0
+	A9  Pin = PK1
+	A10 Pin = PK2
+	A11 Pin = PK3
+	A12 Pin = PK4
+	A13 Pin = PK5
+	A14 Pin = PK6
+	A15 Pin = PK7
+
+	// Analog Input
+	ADC0  Pin = PF0
+	ADC1  Pin = PF1
+	ADC2  Pin = PF2
+	ADC3  Pin = PF3
+	ADC4  Pin = PF4
+	ADC5  Pin = PF5
+	ADC6  Pin = PF6
+	ADC7  Pin = PF7
+	ADC8  Pin = PK0
+	ADC9  Pin = PK1
+	ADC10 Pin = PK2
+	ADC11 Pin = PK3
+	ADC12 Pin = PK4
+	ADC13 Pin = PK5
+	ADC14 Pin = PK6
+	ADC15 Pin = PK7
+
+	// Digital pins
+	D0  Pin = PE0
+	D1  Pin = PE1
+	D2  Pin = PE6
+	D3  Pin = PE5
+	D4  Pin = PG5
+	D5  Pin = PE3
+	D6  Pin = PH3
+	D7  Pin = PH4
+	D8  Pin = PH5
+	D9  Pin = PH6
+	D10 Pin = PB4
+	D11 Pin = PB5
+	D12 Pin = PB6
+	D13 Pin = PB7
+	D14 Pin = PJ1
+	D15 Pin = PJ0
+	D16 Pin = PH1
+	D17 Pin = PH0
+	D18 Pin = PD3
+	D19 Pin = PD2
+	D20 Pin = PD1
+	D21 Pin = PD0
+	D22 Pin = PA0
+	D23 Pin = PA1
+	D24 Pin = PA2
+	D25 Pin = PA3
+	D26 Pin = PA4
+	D27 Pin = PA5
+	D28 Pin = PA6
+	D29 Pin = PA7
+	D30 Pin = PC7
+	D31 Pin = PC6
+	D32 Pin = PC5
+	D33 Pin = PC4
+	D34 Pin = PC3
+	D35 Pin = PC2
+	D36 Pin = PC1
+	D37 Pin = PC0
+	D38 Pin = PD7
+	D39 Pin = PG2
+	D40 Pin = PG1
+	D41 Pin = PG0
+	D42 Pin = PL7
+	D43 Pin = PL6
+	D44 Pin = PL5
+	D45 Pin = PL4
+	D46 Pin = PL3
+	D47 Pin = PL2
+	D48 Pin = PL1
+	D49 Pin = PL0
+	D50 Pin = PB3
+	D51 Pin = PB2
+	D52 Pin = PB1
+	D53 Pin = PB0
+)
--- a/src/machine/machine_atmega328p.go
+++ b/src/machine/machine_atmega328p.go
@@ -2,6 +2,54 @@
 
 package machine
 
-import "device/avr"
+import (
+	"device/avr"
+	"runtime/volatile"
+)
 
 const irq_USART0_RX = avr.IRQ_USART_RX
+
+// Configure sets the pin to input or output.
+func (p Pin) Configure(config PinConfig) {
+	if config.Mode == PinOutput { // set output bit
+		if p < 8 {
+			avr.DDRD.SetBits(1 << uint8(p))
+		} else if p < 14 {
+			avr.DDRB.SetBits(1 << uint8(p-8))
+		} else {
+			avr.DDRC.SetBits(1 << uint8(p-14))
+		}
+	} else { // configure input: clear output bit
+		if p < 8 {
+			avr.DDRD.ClearBits(1 << uint8(p))
+		} else if p < 14 {
+			avr.DDRB.ClearBits(1 << uint8(p-8))
+		} else {
+			avr.DDRC.ClearBits(1 << uint8(p-14))
+		}
+	}
+}
+
+// Get returns the current value of a GPIO pin.
+func (p Pin) Get() bool {
+	if p < 8 {
+		val := avr.PIND.Get() & (1 << uint8(p))
+		return (val > 0)
+	} else if p < 14 {
+		val := avr.PINB.Get() & (1 << uint8(p-8))
+		return (val > 0)
+	} else {
+		val := avr.PINC.Get() & (1 << uint8(p-14))
+		return (val > 0)
+	}
+}
+
+func (p Pin) getPortMask() (*volatile.Register8, uint8) {
+	if p < 8 {
+		return avr.PORTD, 1 << uint8(p)
+	} else if p < 14 {
+		return avr.PORTB, 1 << uint8(p-8)
+	} else {
+		return avr.PORTC, 1 << uint8(p-14)
+	}
+}
--- a/targets/arduino-mega2560.json
+++ b/targets/arduino-mega2560.json
@@ -1,0 +1,8 @@
+{
+    "inherits": ["atmega2560"],
+    "build-tags": ["arduino_mega2560"],
+    "ldflags": [
+        "-Wl,--defsym=_bootloader_size=8192"
+    ],
+    "flash-command":"avrdude -c wiring -b 115200 -p atmega2560 -P {port} -U flash:w:{hex} -v -D"
+}
--- a/targets/atmega2560.json
+++ b/targets/atmega2560.json
@@ -1,0 +1,18 @@
+{
+    "inherits": ["avr"],
+    "llvm-target": "avr-atmel-none",
+    "cpu": "atmega2560",
+    "build-tags": ["atmega2560", "atmega"],
+    "cflags": [
+        "-mmcu=atmega2560"
+    ],
+    "ldflags": [
+        "-mmcu=avr6",
+        "-Wl,--defsym=_stack_size=512"
+    ],
+    "linkerscript": "src/device/avr/atmega2560.ld",
+    "extra-files": [
+        "targets/avr.S",
+        "src/device/avr/atmega2560.s"
+    ]
+}
--- a/src/machine/machine_atmega.go
+++ b/src/machine/machine_atmega.go
@@ -5,53 +5,7 @@ package machine
 import (
 	"device/avr"
 	"runtime/interrupt"
-	"runtime/volatile"
 )
-
-// Configure sets the pin to input or output.
-func (p Pin) Configure(config PinConfig) {
-	if config.Mode == PinOutput { // set output bit
-		if p < 8 {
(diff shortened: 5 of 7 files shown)


tmclane:
Compiled and flashed with:


    ./build/tinygo flash -target targets/atmega2560.json examples/blinky1
    ./build/tinygo flash -target targets/atmega2560.json examples/serial

targets/atmega2560.json
```diff
@@ -0,0 +1,20 @@
...
+    "inherits": ["avr"],
+    "llvm-target": "avr-atmel-none",
+    "cpu": "atmega2560",
+    "build-tags": ["atmega2560"],
+    "cflags": [
+        "-mmcu=atmega2560"
+    ],
+    "ldflags": [
+        "-mmcu=avr6",
+        "-Wl,--defsym=_bootloader_size=0",
```
aykevl: I looked at the Arduino IDE and it appears that the bootloader is 8kB in size (253952 - 256kB = 8kB):

    mega.menu.cpu.atmega2560.upload.maximum_size=253952

targets/atmega2560.json
```diff
@@ -0,0 +1,20 @@
+{
+    "inherits": ["avr"],
+    "llvm-target": "avr-atmel-none",
+    "cpu": "atmega2560",
+    "build-tags": ["atmega2560"],
```
aykevl: Ideally the chip and the board would be separated, but they don't necessarily have to.

* If they are separated, there is one atmega2560.json that defines only the `atmega2560` and `atmega` build tag, and one arduino-mega2560.json that defines board-specific configuration (such as the `arduino_mega2560` build tag, the bootloader size, and the flash command).
* Otherise, this file should be named arduino-mega2560.json and also include the `arduino_mega2560` build tag.

Also, this chip is an ATmega so should define the `atmega` build tag.

src/runtime/runtime_atmega.go
```diff
@@ -1,4 +1,4 @@
-// +build avr,atmega
+// +build avr,atmega avr,atmega2560
```
aykevl: This change is not necessary with the addition of the `atmega` build tag.

src/machine/machine_atmega328p.go
```diff
@@ -0,0 +1,15 @@
+// +build avr,atmega328p
+
+package machine
+
+import "device/avr"
+
+const irq_USART0_RX = avr.IRQ_USART_RX
+
+// LED on the Arduino
+const LED Pin = 13
```
aykevl: The LED pin is board specific so should be defined in board_arduino-mega2560.go (with the appropriate build tag: `// +build arduino_mega2560`).

src/machine/board_arduino.go
```diff
@@ -1,15 +1,7 @@
-// +build arduino
+// +build arduino,!atmega2560
```
aykevl: This change should not be necessary.

src/machine/board_arduino.go
```diff
@@ -1,15 +1,7 @@
...
 package machine
 
-// Return the current CPU frequency in hertz.
-func CPUFrequency() uint32 {
-	return 16000000
-}
-
-// LED on the Arduino
-const LED Pin = 13
-
```
aykevl: This appears to be leftover from testing other changes?

jaddr2line:
It might make sense to change our naming scheme a bit. I propose that we change `arduino` -> `arduino-uno`, and call this target `arduino-mega`.

src/machine/board_arduino_mega2560.go
```diff
@@ -0,0 +1,88 @@
+// +build atmega2560
```
aykevl: This build tag should reflect the file name: `arduino_mega2560`.

targets/atmega2560.json
tmclane: Ok I was wondering how this should be broken down and what the `target` should be named.

src/machine/machine_atmega328p.go
tmclane: Ok thanks for the clarification.  It was unclear to me what pieces went in which files.  I have seen mixtures in other files that contradicted themselves.

src/machine/board_arduino.go
tmclane: believe it is actually.  I was trying to fix the CircleCI wasm build.

aykevl:
> It might make sense to change our naming scheme a bit. I propose that we change `arduino` -> `arduino-uno`,

Yes, I think changing the name of the current `arduino` target is a good idea (I've often thought about it but never done it). We could leave an alias for `arduino` but use `arduino-uno` in all examples.

>  and call this target `arduino-mega`.

Actually, there was an "Arduino Mega" based on the atmega1280. They now call it the "Arduino Mega 2560" to distinguish the two boards. Therefore I think it should be named `arduino-mega2560`.

targets/atmega2560.json
tmclane: I'll see if I can get this split up correctly!

targets/atmega2560.json
tmclane: > Thank you for this contribution!
> 
> This PR should be rebased on top of the dev branch after #946 has been merged.

Of course this will be done.

src/runtime/runtime_atmega.go
tmclane: > > It might make sense to change our naming scheme a bit. I propose that we change `arduino` -> `arduino-uno`,
> 
> Yes, I think changing the name of the current `arduino` target is a good idea (I've often thought about it but never done it). We could leave an alias for `arduino` but use `arduino-uno` in all examples.
> 
> > and call this target `arduino-mega`.
> 
> Actually, there was an "Arduino Mega" based on the atmega1280. They now call it the "Arduino Mega 2560" to distinguish the two boards. Therefore I think it should be named `arduino-mega2560`.



> > It might make sense to change our naming scheme a bit. I propose that we change `arduino` -> `arduino-uno`,
> 
> Yes, I think changing the name of the current `arduino` target is a good idea (I've often thought about it but never done it). We could leave an alias for `arduino` but use `arduino-uno` in all examples.
> 
> > and call this target `arduino-mega`.
> 
> Actually, there was an "Arduino Mega" based on the atmega1280. They now call it the "Arduino Mega 2560" to distinguish the two boards. Therefore I think it should be named `arduino-mega2560`.

Should this type of renaming be done in this same PR?
I'll make them separate commits at the very least.

targets/atmega2560.json
tmclane: The comment grouping on Github is being funny. 

I'll make the bootloader change.

src/runtime/runtime_atmega.go
aykevl: > > Actually, there was an "Arduino Mega" based on the atmega1280. They now call it the "Arduino Mega 2560" to distinguish the two boards. Therefore I think it should be named `arduino-mega2560`.
> 
> Should this type of renaming be done in this same PR?
> I'll make them separate commits at the very least.

The Arduino Mega, yes. But not `arduino` -> `arduino-uno`, that's a separate matter.

tmclane:
I have :

- rebased my branch back on top of the `atmega1284` branch
- refactored the targets
  - targets/atmega2560.json - atmega2560 generic settings
  - targets/arduino-mega2560.json - Arduino Mega2560 specific settings
- refactored the board / platform code
  - src/machine_atmega2560.go - machine specific settings (currently only the const definition)
  - src/board_arduino_mega2560.go - Arduino Mega 2560 pin definitions + pin function definitions

**Note:**  I had to add a build exclude tag to the `machine_atmega.go` file since otherwise the board specific functions like `func (p Pin) Configure(config PinConfig)` were defined twice.
I believe that this would be solved by moving the code out of `machine_atmega.go` and into a `Arduino Uno` specific file (which wasn't intended to be included in this PR).

aykevl:
> **Note:** I had to add a build exclude tag to the `machine_atmega.go` file since otherwise the board specific functions like `func (p Pin) Configure(config PinConfig)` were defined twice.

There is a difference between board files (like `board_arduino_mega2560.go`) and chip files (like `machine_atmega2560.go`).
This exclude really needs to be removed. Where that leads to a conflict because a function is defined multiple times:
* if it is the same for all the atmega chips (atmega328p, atmega2560, etc), it should be in the shared `machine_atmega.go` file.
* If it needs to be different, move it to more specific files.

For example, it seems like UART configuration can be shared between chips so `(UART) Configure` can remain in machine_atmega.go while a function like `getPortMask` will need to be moved to more specific files (`machine_atmega328p.go` and `machine_atmega2560.go`).

tmclane:
Ok I tested all digital pins 0 - 53 and found a few bugs and fixed them.
I am going to check the Analog pins next and figure out why PWM doesn't work as well.

tmclane:
Status update:

- A0 & A1 will not output anything in digital mode
- ADC works except that A0 and A1 float high unless grounded

src/machine/board_arduino_mega2560.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func (p Pin) getPortMask() (*volatile.Register8, uint8) {
+	_, port, mask := p.getRegisterPortMask()
+	return port, mask
+}
+
+// getRegisterPortMask returns the register, port, and mask for the pin
+func (p Pin) getRegisterPortMask() (*volatile.Register8, *volatile.Register8, uint8) {
+	if p <= A0 && p > A8 { // A0-A7
```
deadprogram: It would be a lot better to use a `switch` statement here. Also, it appears that the condition here is the opposite to what the comment says it is?

tmclane: I kept it as separate `if` statements because the original I copied as a guide used them (`machine_atmega.go` - before this branch)

Why do you think the comment is wrong?
it's not a subtraction operator in this case:  

```	A0  Pin = 97 // PF0
	A1  Pin = 96 // PF1
	A2  Pin = 95 // PF2
	A3  Pin = 94 // PF3
	A4  Pin = 93 // PF4
	A5  Pin = 92 // PF5
	A6  Pin = 91 // PF6
	A7  Pin = 90 // PF7
	A8  Pin = 89 // PK0
```

I guess I should have used set notation?

tmclane: The comments were intended to be pin ranges not mathematical operations.

tmclane: I'll update the code to use a `switch` statement.  I had thought about that originally but thought there was some reason not to since I hadn't seen one used anywhere (although I didn't look very hard)

deadprogram: My bad, I only just noticed that the A0 thru A9 is in descending order. Most of the boards this is ascending.

Either way, my comment about `switch` instead of a whole bunch of `if` statements still stands, if you please.

tmclane:
@deadprogram if => switch has been done. 

 For the `default` case what do you recommend to return? 
I am not happy with what is being returned at the moment.  I suppose the callers of this function could check for a bad mask and silently discard the request?

deadprogram:
One option for the default case is to panic, which should happen at compile time causing the compile to fail.

deadprogram:
BTW just tested blinky1 on both my arduino uno and also my mega2560 board, and it worked as expected on both.

src/machine/board_arduino_mega2560.go
```diff
@@ -0,0 +1,174 @@
...
+
+func (p Pin) getPortMask() (*volatile.Register8, uint8) {
+	_, port, mask := p.getRegisterPortMask()
+	return port, mask
+}
+
+// getRegisterPortMask returns the register, port, and mask for the pin
+func (p Pin) getRegisterPortMask() (*volatile.Register8, *volatile.Register8, uint8) {
+	switch p {
+	case A0, A1, A2, A3, A4, A5, A6, A7:
```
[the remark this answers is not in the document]
tmclane: Ok.  That makes sense.

src/machine/board_arduino_mega2560.go
```diff
@@ -0,0 +1,174 @@
...
+	ADC9  Pin = 9
+	ADC10 Pin = 10
+	ADC11 Pin = 11
+	ADC12 Pin = 12
+	ADC13 Pin = 13
+	ADC14 Pin = 14
+	ADC15 Pin = 15
+
+	// Digital pins
+	D0  Pin = 2  // PE0
```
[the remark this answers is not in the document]
tmclane: I will shift to this style.

src/machine/board_arduino_mega2560.go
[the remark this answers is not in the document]
tmclane: If I use `PF0` when I define it is it ok to use the ATMega2560 pin number for it's value?

i.e. 

```go
const (
    PF0 = 97  // A0 (on Mega 2560)

)
```

Or do you recommend just an `iota` sequence?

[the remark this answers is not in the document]
aykevl: Yes that's the intention. But you could perhaps make it more readable with something like this:

```go
const (
    PF0 Pin = 5*8 + 0
    PF1 Pin = 5*8 + 1
    PF2 Pin = 5*8 + 2
)
```
You could also define `portA`, `portB` constants like [here](https://github.com/tinygo-org/tinygo/blob/master/src/machine/board_stm32.go#L7) and write:

```go
const (
    PF0 = portF + 0
    PF1 = portF + 1
    PF2 = portF + 2
)
```


Where did you get 97 from? Or was it just an example?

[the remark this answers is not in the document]
tmclane: 97 is the Atmega2560 pin number according to my diagram of an Arduino Mega2560.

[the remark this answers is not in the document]
tmclane: Just checked the datasheet:  97 == A0

http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-2549-8-bit-AVR-Microcontroller-ATmega640-1280-1281-2560-2561_datasheet.pdf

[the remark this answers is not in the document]
tmclane: so it isn't board specific.  (was slightly worried it could be)

tmclane:
@aykevl Hopefully this latest push is more in line with what you were expecting.

tmclane:
FYI: I am thinking about changing the subtraction targets to be the `port<X>` constants instead of the lowest pin on each port.  Would the be nicer looking?

targets/arduino-mega2560.json
```diff
@@ -0,0 +1,8 @@
+{
+    "inherits": ["atmega2560"],
+    "build-tags": ["arduino_mega2560"],
+    "ldflags": [
+        "-Wl,--defsym=_bootloader_size=253952"
```
aykevl: It's the other way round: 253952 is the size available to the program (according to boards.txt in the Arduino IDE). The chip has 256kB of memory (262144) so the bootloader size must be 8192, or 8K.

```suggestion
        "-Wl,--defsym=_bootloader_size=8192"
```

tmclane: Ok I made this change.  I guess I misunderstood an earlier conversation when I put that in place.

tmclane: Where do I place the `253952` instead?  Is there another `defsym` value?

aykevl: Nowhere, it's just a different way of indicating which area of flash is available for the program.

aykevl:
Thank you! I've now merged this PR into the dev branch so it'll be part of the next release.

tmclane:
awesome to hear!

PR #955, opened 2020-03-14
Makefile: allow overriding the llvm-project directory

aykevl:
This allows setting a different directory for experimental changes. For example, I also have a llvm-project.master directory that tracks the master branch. It is very useful to be able to temporarily use that directory:

    make llvm-build.master LLVM_BUILDDIR=llvm-build.master LLVM_PROJECT=llvm-project.master


deadprogram:
Very reasonable to be able to test with LLVM next. Merging.

PR #435, opened 2019-07-04
Uses `go version' and zversion.go to detect version, fixes #433

diamondburned:
This first executes `go version` then parses it. If it fails, it reads `zversion.go`. If all else fails, it dies like usual.

PR #956, opened 2020-03-15
consider GOROOT/api as fallback to determine Go version

ardnew:
When using a development branch of Go (see #435), the file `$GOROOT/VERSION` doesn't exist, and `src/runtime/internal/sys/zversion.go` contains a version string like the following:
```go
const TheVersion = `devel +529988d62c Sun Mar 1 22:26:33 2020 +0000`
```
Instead of the expected format:
```go
const TheVersion = `go1.13.7`
```
Thus `tinygo` will fail to run without any existing means of determining version, for example:
```sh
$ tinygo flash -target pyportal .
error: could not read version from GOROOT (/home/andrew/Development/go/dev/goroot): could not parse Go version: version does not start with 'go' prefix
```

This change will look at the content of directory `$GOROOT/api` to find the semantically greatest version **only if** the above two checks first fail, and the value of `const TheVersion` begins with `"devel"`

Note that this still won't allow _current_ master branch to run, which will produce a version `go1.14`, resulting in:
```sh
$ tinygo flash -target pyportal .
error: requires go version 1.11, 1.12, or 1.13, got go1.14
```
But this is a separate unrelated issue (#901 I believe).


deadprogram:
I think this has been addressed already. @ardnew can we close this PR?

ardnew:
@deadprogram I'm not sure which commit would have addressed it, but I can test and verify it's resolved today and let you know.

I began always using a _release version_ of Go specifically for my TinyGo runtime, which I think most people do anyway.

ardnew:
Building `master@HEAD` from mainline Go, it will report a `devel` version number appended with a commit hash:

```
$ go version
go version devel go1.19-0c6d8bb109 2022/04/15 12:17:18 darwin/amd64
```

TinyGo doesn't immediately report any problems:
```
$ ./build/tinygo version
tinygo version 0.23.0-dev-2dc46a85 darwin/amd64 (using go version devel go1.19-0c6d8bb109 2022/04/15 12:17:18 and LLVM version 13.0.0)
```

But when actually trying to compile something, the following error is still generated:
```
$ ./build/tinygo build -target=arduino ./src/examples/blinky1/blinky1.go
error: could not read version from GOROOT (/usr/local/go/dev): could not parse Go version: version does not start with 'go' prefix
```

After rebasing this PR against latest `dev` branch, the logic in this change needed to be added to package `goenv`. It was also left in `builder`, but I'm not certain if this is used anymore?

Now building with the mainline Go development branch succeeds:

```
$ tinygo env
GOOS="darwin"
GOARCH="amd64"
GOROOT="/usr/local/go/dev"
GOPATH="/usr/local/go"
GOCACHE="/Users/andrew/Library/Caches/tinygo"
CGO_ENABLED="1"
TINYGOROOT="/Users/andrew/.tinygo"
$ tinygo version
tinygo version 0.23.0-dev-c3c09e9d darwin/amd64 (using go version devel go1.19-0c6d8bb109 2022/04/15 12:17:18 and LLVM version 13.0.0)
$ ./build/tinygo build -target=arduino -size=short ./src/examples/blinky1/blinky1.go
   2354      90     673 |    2444     763
```

deadprogram:
Closing due to age of PR and too many changes since then. Please reopen a new one if this is still needed. Thanks @ardnew

PR #957, opened 2020-03-15
WIP avr: add support for the Digispark Pro

aykevl:
This doesn't work yet and I'm not sure why. @deadprogram maybe you can take a look and see what's wrong? The original Digispark works fine with this PR.

### DIFF
--- a/targets/digispark-pro.json
+++ b/targets/digispark-pro.json
@@ -1,0 +1,21 @@
+{
+	"inherits": ["avr"],
+	"llvm-target": "avr-atmel-none",
+	"cpu": "attiny167",
+	"build-tags": ["digispark_pro", "attiny167", "attiny"],
+	"cflags": [
+		"-mmcu=attiny167"
+	],
+	"ldflags": [
+		"-mmcu=avr35",
+		"-Wl,--defsym=_bootloader_size=1540",
+		"-Wl,--defsym=_stack_size=128"
+	],
+	"linkerscript": "src/device/avr/attiny167.ld",
+	"extra-files": [
+		"targets/avr.S",
+		"src/device/avr/attiny167.s"
+	],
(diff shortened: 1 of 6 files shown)


targets/digispark-pro.json
```diff
@@ -0,0 +1,21 @@
+{
+	"inherits": ["avr"],
+	"llvm-target": "avr-atmel-none",
+	"cpu": "attiny167",
+	"build-tags": ["digispark_pro", "attiny167", "attiny"],
+	"cflags": [
+		"-mmcu=attiny167"
+	],
+	"ldflags": [
+		"-mmcu=avr35",
```
deadprogram: Should this also be `"-mmcu=attiny167"` like the `cflags` argument above?

deadprogram: Just saw https://github.com/tinygo-org/tinygo/pull/946/commits/2381cf9b661bcfc8187332216b2c88af53ba76b7 so probably I am incorrect? or maybe this should be `"-mmcu=avr25"`?

deadprogram:
Played around with it, but was not able to get anything to happen, nor to notice any specific problems. Very odd.

deadprogram:
CCing @ekettenburg just in case he might take an interest in what we are trying to do here.

targets/digispark-pro.json
aykevl: Exactly, apparently things go wrong when you specify the exact device. You can get the value of `-mmcu=avr*` from here: https://gcc.gnu.org/onlinedocs/gcc/AVR-Options.html (it lists the attiny167 under avr35).

aykevl:
I don't currently have plans to continue working on this PR, so closing. If I want to do it, I can always restore the code from this PR.

PR #958, opened 2020-03-16
Change samd51-usbcdc's timeout

sago35:
This PR fixes the usbcdc timeout value for samd51.

In the original source, sometimes bytes from fmt.Printf () disappear.
I think that the location of 3000 in samd21 should be 7500.

After this fix, the bytes no longer disappear in the range I'm trying.
However, I don't know if the timeout value is reasonable.

### DIFF
--- a/src/machine/machine_atsamd51.go
+++ b/src/machine/machine_atsamd51.go
@@ -1451,7 +1451,7 @@ func (usbcdc USBCDC) WriteByte(c byte) error {
 		setEPSTATUSSET(usb_CDC_ENDPOINT_IN, sam.USB_DEVICE_ENDPOINT_EPSTATUSSET_BK1RDY)
 
 		// wait for transfer to complete
-		timeout := 3000
+		timeout := 7500
 		for (getEPINTFLAG(usb_CDC_ENDPOINT_IN) & sam.USB_DEVICE_ENDPOINT_EPINTFLAG_TRCPT1) == 0 {
 			timeout--
 			if timeout == 0 {
@@ -1794,7 +1794,7 @@ func handleStandardSetup(setup usbSetup) bool {
 		setEPSTATUSSET(0, sam.USB_DEVICE_ENDPOINT_EPSTATUSSET_BK1RDY)
 
 		// wait for transfer to complete
-		timeout := 3000
+		timeout := 7500
 		for (getEPINTFLAG(0) & sam.USB_DEVICE_ENDPOINT_EPINTFLAG_TRCPT1) == 0 {
 			timeout--
 			if timeout == 0 {
@@ -1931,7 +1931,7 @@ func receiveUSBControlPacket() []byte {
 	setEPSTATUSCLR(0, sam.USB_DEVICE_ENDPOINT_EPSTATUSCLR_BK0RDY)
 
 	// Wait until OUT transfer is ready.
-	timeout := 300000
+	timeout := 750000
 	for (getEPSTATUS(0) & sam.USB_DEVICE_ENDPOINT_EPSTATUS_BK0RDY) == 0 {
 		timeout--
 		if timeout == 0 {
@@ -1940,7 +1940,7 @@ func receiveUSBControlPacket() []byte {
 	}
 
 	// Wait until OUT transfer is completed.
-	timeout = 300000
+	timeout = 750000
 	for (getEPINTFLAG(0) & sam.USB_DEVICE_ENDPOINT_EPINTFLAG_TRCPT1) == 0 {
 		timeout--
 		if timeout == 0 {


sago35:
Test script : https://raw.githubusercontent.com/tinygo-org/drivers/master/examples/adt7410/main.go

Previously: sometimes bytes from fmt.Printf () disappear.

```
temperature: 70.699997
temperature: 70.699997
temperature: 70.699997
tmperature: 70.699997
temperature: 70.699997
temperature: 70.699997
temperature: 70.587494
temperature: 70.587494
temperature: 70.925003
temperature: 70.925003
tmperature: 70.925003
temperature: 72.274994
temperature: 72.274994
temperature: 73.625000
```

aykevl:
This seems a little bit iffy to me. Is there any other way to detect a timeout without a magic constant? (@deadprogram)

deadprogram:
The timeout is to prevent a situation where it gets stuck waiting forever. By the way `timeout` is really just the number of times to retry, so it is a little misnamed.

The original value is completely arbitrary, so increasing it might be ok? 

If we do change this value, we should probably use the same values in the samd21 implementation.

sago35:
I'm sorry.
Increasing the timeout value was a bit too rough.

Of the four locations modified in this PR, I think only the location of WriteByte () is actually needed.

The current implementation returns `USEBCDC write byte timeout`, but I think it is because usb.go did not receive err.
Thus, `c byte` is lost when timeout == 0.

We can retry in Write() of usb.go, but we need to put a limit on retry.
For the correct implementation, Write() receives the err of usbcdc.WriteByte() and
If it was an err, Write() must also return err, returning n the number of bytes written.
And the process that receives err of Write() needs to rewrite the byte that could not be written.

I think the following is a good (?) way to proceed.

1. First, reduce the probability of occurrence by increasing the timeout
  * Should not affect anyone who has not timed out
2. Modify the part that calls usbcdc.Write () to write at the next timing

However, this may be a problem only for Windows 10 or a problem only for pyportal.
If the problem is not on Linux or Mac, it may be due to the current cdc starting to work.

In addition, my trial is trying below.
https://github.com/sago35/tinygo/commit/f80cc339e5c2f18b559a2f94f59352c05329a816

I will try it below.
https://github.com/tinygo-org/tinygo/commit/24e68b13fc358515da2ad0316a3c3ba792e8a996


```go
// src/machine/machine_atsamd51.go
func (usbcdc USBCDC) WriteByte(c byte) error {
    if usbLineInfo.lineState > 0 {
        // set the data
        // ...

        // wait for transfer to complete
        timeout := 3000
        for (getEPINTFLAG(usb_CDC_ENDPOINT_IN) & sam.USB_DEVICE_ENDPOINT_EPINTFLAG_TRCPT1) == 0 {
            timeout--
            if timeout == 0 {
                return errors.New("USBCDC write byte timeout")
            }
        }
    }

    return nil
}


// src/machine/usb.go
func (usbcdc USBCDC) Write(data []byte) (n int, err error) {
    for _, v := range data {
        usbcdc.WriteByte(v)
        // err != nil の時は、次の byte に進んでしまう
    }
    return len(data), nil
}
```

sago35:
In the following code, we confirmed that led is inverted when byte disappears.
I want to know a good fix for this problem.

```go
func (usbcdc USBCDC) WriteByte(c byte) error {
    // Supposedly to handle problem with Windows USB serial ports?
    if usbLineInfo.lineState > 0 {
        // set the data
        // ...

        // wait for transfer to complete
        timeout := 3000
        for (getEPINTFLAG(usb_CDC_ENDPOINT_IN) & sam.USB_DEVICE_ENDPOINT_EPINTFLAG_TRCPT1) == 0 {
            timeout--
            if timeout == 0 {
                led := LED
                led.Configure(PinConfig{Mode: PinOutput})
                led.Toggle()
                return errors.New("USBCDC write byte timeout")
            }
        }
    }

    return nil
}
```

deadprogram:
Hi @sago35 just wondering if the latest changes made to `dev` address this issue already?

sago35:
Hi @deadprogram 
The problem is still not solved at bugfix/usb-cdc-zlp-trcpt d34b260120e8287a9cc8a0d1dae5b04016038faf.

```
temperature: 76.212494
temperature: 76.212494
temperature: 76.212494
temperature: 76.212494
teperature: 76.212494
temperature: 76.212494
temperature: 76.099998
temperature: 75.987503
temperature: 76.212494
temperature: 76.212494
temperature: 76.212494
temperature: 76.324997
temperature: 76.212494
temperature: 75.987503
tmperature: 76.324997
tmperature: 76.212494
temperature: 76.212494
temperature: 75.875000
temperature: 75.987503
temperature: 75.987503
temperature: 76.099998
temperature: 75.875000
temperature: 75.987503
temperature: 75.762497
temperature: 75.425003
temperature: 75.312500
temperature: 75.649994
temperature: 75.649994
```


Maybe we need to write the following code and fix the USBCDC.Write() call part properly.
However, I don't know where the function call is coming from.


```go
// Write data to the USBCDC.
func (usbcdc USBCDC) Write(data []byte) (n int, err error) {
    for i, v := range data {
        err := usbcdc.WriteByte(v)
        if err != nil {
            return i-1, nil
        }
    }
    return len(data), nil
}
```

sago35:
I think it will be called from here, but if you call it in this way, it is better to keep sending it until the end with blocking.
However, long blocking process should be avoided if possible, so a good method is needed.

compiler/compiler.go:1270

aykevl:
@deadprogram Remeber the issues we had with flashing the esp32 coprocessor on the Arduino Nano 33 IoT? Perhaps it was caused by this issue.

sago35:
I think I'm going to close this PR.
Increasing the timeout value reduces the probability of data loss, but not to zero.
I'm still having this problem, so I'll launch another Issue #1175 .

Also, the current USB-CDC is very slow as it waits for each byte to be transmitted to complete.

Sending size can be increased to about 128 bytes, or DMA can be used to speed up the transmission process. It may be necessary.

PR #959, opened 2020-03-16
docker: use git submodule sync to fix broken docker dev build

deadprogram:
This PR adds the use of the `git submodule sync` command in the docker dev build to handle case where repo url changes, which fixes the broken dockerhub builds:

https://hub.docker.com/repository/registry-1.docker.io/tinygo/tinygo-dev/builds/2c37af2e-b3c7-4718-bb87-fe145312ad44


aykevl:
Not sure whether this is going to work but let's try it anyway.

aykevl:
And it works! Thank you!

PR #960, opened 2020-03-17
refactor stm32 UART code

gwtnz:
Step 2 of #777 : UART code cleanup

Moving the common parts of the STM32F1xx and STM32F4xx UART setup into a separate file, and simplifying the chip-specific pieces.

### DIFF
--- a/src/machine/machine_stm32f407.go
+++ b/src/machine/machine_stm32f407.go
@@ -79,75 +79,32 @@ func enableAltFuncClock(bus unsafe.Pointer) {
 	}
 }
 
-// UART
+//---------- UART related types and code
+
+// UART representation
 type UART struct {
 	Buffer          *RingBuffer
 	Bus             *stm32.USART_Type
+	Interrupt       interrupt.Interrupt
 	AltFuncSelector stm32.AltFunc
 }
 
-var (
-	UART0 = UART{
-		Buffer:          NewRingBuffer(),
-		Bus:             stm32.USART2,
-		AltFuncSelector: stm32.AF7_USART1_2_3,
-	}
-	UART1 = &UART0
-)
-
 // Configure the UART.
-func (uart UART) Configure(config UARTConfig) {
-	// Default baud rate to 115200.
-	if config.BaudRate == 0 {
-		config.BaudRate = 115200
-	}
-
-	// Set the GPIO pins to defaults if they're not set
-	if config.TX == 0 && config.RX == 0 {
-		config.TX = UART_TX_PIN
-		config.RX = UART_RX_PIN
-	}
-
-	// Enable USART clock
-	enableAltFuncClock(unsafe.Pointer(uart.Bus))
-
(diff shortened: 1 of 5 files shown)


src/device/stm32/stm32f103xx-bitfields.go
```diff
@@ -5,6 +5,10 @@
 
 package stm32
 
+// This is not used by the stm32f1xx series, but included here for the common
+// stm32 peripheral functions to operate
+type AltFunc uint8
```
gwtnz: I'm not thrilled about having this here since the stm32f1xx doesn't use it, but it's needed for the common UART file as that defines a UART struct referring to this type. The alternative that I can see is to put the type-definition of the UART struct in each multiple places, which doesn't seem great either.

deadprogram: The question is do the 4 lines of duplication for the type or the additional abstraction? I would probably go with the duplicate type, myself.

gwtnz: Yeah that seems reasonable, and the goal was to move the code to a common place, but the types can probably live closer to their uses (i.e. at the lower level chip-specific stuff).

deadprogram: Looking good now, thanks for making that change.

deadprogram:
Hello @gwtnz anything else needed in this PR now? Seems like it is now doing what you originally set out to do.

gwtnz:
This one is ready to go. I'm lining up an SPI and I2C one on the heels of this.

gwtnz:
@deadprogram yes, this one's ready for reviewing now :)

deadprogram:
I tested this branch on my Bluepill, with my fix from PR #974 applied, and it worked perfectly.

src/machine/machine_stm32f407.go
```diff
@@ -79,75 +79,32 @@ func enableAltFuncClock(bus unsafe.Pointer) {
...
-// WriteByte writes a byte of data to the UART.
-func (uart UART) WriteByte(c byte) error {
-	uart.Bus.DR.Set(uint32(c))
-
-	for !uart.Bus.SR.HasBits(stm32.USART_SR_TXE) {
+// UART baudrate calc based on the bus and clockspeed
+// NOTE: keep this in sync with the runtime/runtime_stm32f407.go clock init code
+func (uart UART) getBaudRateDivisor(br uint32) uint32 {
+	var clock uint32
+	switch unsafe.Pointer(uart.Bus) {
```
aykevl: I do not think you have to cast to `unsafe.Pointer` here, just switch on the raw pointers.

gwtnz:
good suggestion @aykevl . This cleans up that function a bunch too with all the casting removed.
The PR should be ready to go now.

deadprogram:
Thanks for all your efforts on this PR @gwtnz now merging.

PR #961, opened 2020-03-17
avr: add support for tasks scheduler

aykevl:
This adds support for the `-scheduler=tasks` flag for AVR. On most AVR chips you wouldn't want to run a real scheduler but it may be useful in some cases, especially on devices with more RAM. It is disabled by default.

### DIFF
--- a/src/internal/task/task_stack_avr.go
+++ b/src/internal/task/task_stack_avr.go
@@ -1,0 +1,88 @@
+// +build scheduler.tasks,avr
+
+package task
+
+import "unsafe"
+
+const stackSize = 256
+
+// calleeSavedRegs is the list of registers that must be saved and restored when
+// switching between tasks. Also see scheduler_avr.S that relies on the
+// exact layout of this struct.
+//
+// https://gcc.gnu.org/wiki/avr-gcc#Call-Saved_Registers
+type calleeSavedRegs struct {
+	r2r3   uintptr
+	r4r5   uintptr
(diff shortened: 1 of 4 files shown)


src/internal/task/task_stack_avr.go
```diff
@@ -0,0 +1,88 @@
...
+// https://gcc.gnu.org/wiki/avr-gcc#Call-Saved_Registers
+type calleeSavedRegs struct {
+	r2r3   uintptr
+	r4r5   uintptr
+	r6r7   uintptr
+	r8r9   uintptr
+	r10r11 uintptr
+	r12r13 uintptr
+	r14r15 uintptr
+	r16r17 uintptr
```
jaddr2line: Is there a reason that these are paired, rather than putting separate registers in separate byte-sized fields?

aykevl: Most of the toolchain treats these as register pairs instead of individual registers. It makes other things easier, such as:

```go

	// Pass the function to call in r2:r3.
	// This function is a compiler-generated wrapper which loads arguments out
	// of a struct pointer. See createGoroutineStartWrapper (defined in
	// compiler/goroutine.go) for more information.
	r.r2r3 = fn

	// Pass the pointer to the arguments struct in r4:45.
	r.r4r5 = uintptr(args)
```

aykevl:
Thank you for the review!

PR #963, opened 2020-03-17
riscv: switch to tasks-based scheduler

aykevl:
This means RISC-V will use a very similar scheduler as is used on Cortex-M.

WIP: there is some sort of crash when running testdata/math.go and testdata/reflect.go (apparently somewhere in `tinygo_pause` after exiting a goroutine).

### DIFF
--- a/src/internal/task/task_stack_tinygoriscv.go
+++ b/src/internal/task/task_stack_tinygoriscv.go
@@ -1,0 +1,88 @@
+// +build scheduler.tasks,tinygo.riscv
+
+package task
+
+import "unsafe"
+
+const stackSize = 1024
+
+// calleeSavedRegs is the list of registers that must be saved and restored when
+// switching between tasks. Also see scheduler_riscv.S that relies on the
+// exact layout of this struct.
+type calleeSavedRegs struct {
+	s0  uintptr // x8 (fp)
+	s1  uintptr // x9
+	s2  uintptr // x18
+	s3  uintptr // x19
+	s4  uintptr // x20
+	s5  uintptr // x21
+	s6  uintptr // x22
+	s7  uintptr // x23
+	s8  uintptr // x24
+	s9  uintptr // x25
+	s10 uintptr // x26
+	s11 uintptr // x27
+
+	pc uintptr
+}
+
+// registers gets a pointer to the registers stored at the top of the stack.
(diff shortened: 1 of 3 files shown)


deadprogram:
What is the status on this PR?

aykevl:
Still a WIP, I haven't worked on it since.

yannishuber:
@aykevl I am interested of doing a proof-of-concept kinda support for multi-core scheduling on the Kendryte K210 (I know this is not a goal for TinyGo, it would be as part of my bachelor's project). For this I think I have to first implement the task based scheduler. If you want I can continue your work of this PR unless you are working on it at the moment? Let me know if this would be OK for you.

aykevl:
Right now I'm not working on this, if you are interested in getting it up and running that would be great!

I was stuck at some unaligned load/store instruction (which is not supported on the hifive1b). I remember that I checked both QEMU and real hardware (the hifive1b) and only one gave a precise exception location, I think it was the hifive1b. I am suspecting two causes: either there is a bug elsewhere (such as a memory corruption) that creates this unaligned load/store, or the compiler assumes it can do an unaligned load/store even though the hardware doesn't support it.
(Sidenote: even though the hardware doesn't support it, it should be supported eventually to conform to the ISA specification. This can be done with a trap-and-emulate path. I was hesitant with it because I suspect the bug may be somewhere else).

aykevl:
> I am interested of doing a proof-of-concept kinda support for multi-core scheduling on the Kendryte K210 (I know this is not a goal for TinyGo, it would be as part of my bachelor's project).

Cool! It's not really a goal but support for it would be great to have, also for the ESP32 for example and for bigger targets such as the Raspberry Pi 3. I suspect multicore MCUs may come out more often in the future. The biggest limitation is that it shouldn't affect single core chips too much, ideally not at all: the only asynchronicity they support is interrupts and having code to deal with multiple cores would be wasteful. And while multicore MCUs are generally big and can easily handle a bit of extra code, small single core chips (the vast majority at the moment) may have more trouble with it.

yannishuber:
> Right now I'm not working on this, if you are interested in getting it up and running that would be great!

Great, I will work on this then, thank you for the indications.

aykevl:
Superseded by #1220.

PR #965, opened 2020-03-18
Move small passes

aykevl:
These two commits move two small passes to the transform package, and add tests. It is in preparation for #746, which is now almost finished.


aykevl:
Rebased to make CI succeed.

deadprogram:
This is a good example of refactoring. Does not change the functionality, just moves it, and adds tests. Thanks! Now merging.

PR #966, opened 2020-03-18
build: use go1.13.8 for MS Azure

deadprogram:
This PR specifies to use go1.13.8 instead of the Azure image default which is now go1.14 for MS Azure pipeline builds.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -10,6 +10,9 @@ jobs:
   pool:
     vmImage: 'VS2017-Win2016'
   steps:
+    - task: GoTool@0
+      inputs:
+        version: '1.13.8'  
     - checkout: self
     - task: CacheBeta@0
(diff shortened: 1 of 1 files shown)


aykevl:
It looks like they removed Go 1.13.8 from the image:
https://github.com/actions/virtual-environments/commit/c53984f34c9b7f4cee75ae3c176e915a2715d2ab#diff-55dcabaaca18e802a555dfb957187fbeR459

aykevl:
With this change, the addition of `/c/Go1.13/bin` to `$PATH` is not necessary anymore, can you remove that part from the `$PATH` to keep the azure-pipelines.yml file clean?

deadprogram:
Extra path entries have been removed. Now ready for merge.

aykevl:
Thank you!

PR #967, opened 2020-03-19
interp: add support for constant type asserts

aykevl:
This is needed for compiling math/rand on Go 1.14 (but it is not a complete solution).

Non-constant type asserts are not yet implemented, but should be relatively easy to add at a later time. They should result in a clear error message for now.


deadprogram:
Looks good, now merging.

PR #968, opened 2020-03-19
compiler: do not perform nil checking when indexing slices

aykevl:
The x/tools/go/ssa package splits slice loads/stores into two operations. So for code like this:

    x = p[3]

It outputs two instructions:

    x_ptr = &p[3]
    x = *x_ptr

This makes the IR simpler, but also means we're accidentally inserting more nil checks than necessary: the slice index operation has effectively already checked for nil by performing a bounds check. Therefore, omit nil pointer checks for pointers created by `*ssa.IndexAddr`.

This change is necessary to make sure a future removal of `runtime.isnil` will not cause the escape analysis pass to regress. Apart from that, it reduces code size slightly in many smoke tests (with no increases in code size).

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -1146,7 +1146,9 @@ func (c *Compiler) parseInstr(frame *Frame, instr ssa.Instruction) {
 	case *ssa.Store:
 		llvmAddr := c.getValue(frame, instr.Addr)
 		llvmVal := c.getValue(frame, instr.Val)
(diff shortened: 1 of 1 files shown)


deadprogram:
@aykevl this PR now needs a merge conflict resolved. Thanks.

aykevl:
Closing in favor of #990, which contains this commit but in a slightly different form (and rebased on top of dev after #746 landed). I can update this PR with that commit if needed if that makes review easier.

PR #970, opened 2020-03-20
nrf: add lib/nrfx/mdk to include dirs

aykevl:
This directory is needed for nrf.h and other headers that are used by
the SoftDevice. It is definitely needed for nrf52840 for example, but
I've also added it to the nrf51 as it will likely also require adding
this directory.


aykevl:
Replaced by #972.

PR #972, opened 2020-03-20
reelboard: add SoftDevice target reelboard-s140v7

aykevl:
Unfortunately, `tinygo flash` doesn't work. You have to merge the SoftDevice and the application hex with a mergehex tool and flash that to the board.

Replaces #970.


deadprogram:
> You have to merge the SoftDevice and the application hex with a mergehex tool and flash that to the board.

Where should we document that information?

aykevl:
It would be nice if this PR could be merged before the next release.

> Where should we document that information?

Not sure. Maybe on the TinyGo website? I can't think of a good place in this repository.

deadprogram:
Males sense. Now merging.

Issue #973, opened 2020-03-21
stm32: bluepill blinky1 does not work in dev branch


deadprogram:
I just discovered last night that after commit https://github.com/tinygo-org/tinygo/commit/c8a4994feb1d76f95bb3b49c0bfad9a04844cd0e, the blinky1 program compiles, but does not execute correctly on the bluepill. Nothing happens.

deadprogram:
#974 corrects this issue.

deadprogram:
This has now been released, so now closing. Thanks!

PR #974, opened 2020-03-21
machine/stm32f103xx: enable clock on configure of GPIO pin

deadprogram:
This PR corrects #973 by enabling the clock when a GPIO pin is configured.


aykevl:
@gwtnz what do you think?

gwtnz:
Yes I likely forgot to put that back in after switching back to leaving `Configure()` in the individual chip-family specific places

PR #976, opened 2020-03-22
riscv: implement VirtIO target

aykevl:
This allows running RISC-V tests in CI using QEMU, which should help catch bugs.

This PR is an improvement over #765 in that it doesn't hack around the lack of a test finisher but instead uses a new target (`-target=riscv-qemu`) that is based on VirtIO and actually includes a way to exit the QEMU process.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -54,6 +54,14 @@ func TestCompiler(t *testing.T) {
 		runPlatTests("cortex-m-qemu", matches, t)
 	})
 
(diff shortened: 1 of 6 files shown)


deadprogram:
Since there appear to be changes that also affect the hardware implementation, I will test on my actual board later today.

aykevl:
It only moves code around, so it's very unlikely to change in behavior (it may not even change the binary).

deadprogram:
Behavior verified, so now merging.

PR #977, opened 2020-03-22
main: extend test timeout from 1s to 10s

aykevl:
This should avoid the rather frequent "test ran too long, terminating..." error message that often occurs in CI and when running `go test` manually. Apparently I was too optimistic: some tests take longer than 1 second to run.


deadprogram:
Good idea, I was wondering about that. Merging.

PR #978, opened 2020-03-22
avr: clean up ATtiny definitions

aykevl:
Add definitions for constants like PB0.

This was originally based on #957 but as long as I can't get that chip to work, I'll post the cleanup here anyway.

Note: I've started pin numbers (like `PB0`) from 0 instead of 8 because there is only one port on the attiny85. It saves 14 bytes of code for blinky1.

### DIFF
--- a/src/machine/board_digispark.go
+++ b/src/machine/board_digispark.go
@@ -8,12 +8,12 @@ func CPUFrequency() uint32 {
 }
 
 const (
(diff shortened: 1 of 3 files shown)


deadprogram:
Tested with blinky1 and neopixels demos, and worked perfectly.

Merging, thanks!

PR #979, opened 2020-03-22
Stm32 refactor spi to add to stm32f4

gwtnz:
Extending #777, part 3 : Leverage the stm32f103 SPI implementation across more stm32 families.

Some of the constants might need renaming, not sure.

Also, @deadprogram or @aykevl if you're able to verify the SPI behavior on your f103-based boards, that would be awesome.

### DIFF
--- a/src/machine/machine_stm32_spi.go
+++ b/src/machine/machine_stm32_spi.go
@@ -1,0 +1,99 @@
+// +build stm32
+
+package machine
+
+// Peripheral abstraction layer for SPI on the stm32 family
+
+import (
+	"device/stm32"
+	"unsafe"
+)
+
+// SPIConfig is used to store config info for SPI.
+type SPIConfig struct {
+	Frequency uint32
+	SCK       Pin
+	MOSI      Pin
+	MISO      Pin
+	LSBFirst  bool
+	Mode      uint8
+}
+
+// Configure is intended to setup the STM32 SPI1 interface.
+// Features still TODO:
+// - support SPI2 and SPI3
+// - allow setting data size to 16 bits?
+// - allow setting direction in HW for additional optimization?
+// - hardware SS pin?
+func (spi SPI) Configure(config SPIConfig) {
+	// enable clock for SPI
+	enableAltFuncClock(unsafe.Pointer(spi.Bus))
+
+	// Get SPI baud rate based on the bus speed it's attached to
+	var conf uint32 = spi.getBaudRate(config)
+
+	// set bit transfer order
+	if config.LSBFirst {
+		conf |= stm32.SPI_CR1_LSBFIRST
+	}
+
+	// set polarity and phase on the SPI interface
+	switch config.Mode {
+	case Mode0:
+		conf &^= (1 << stm32.SPI_CR1_CPOL_Pos)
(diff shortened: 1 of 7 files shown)


src/machine/machine_stm32_spi.go
```diff
@@ -0,0 +1,107 @@
...
+// Peripheral abstraction layer for SPI on the stm32 family
+
+import (
+	"device/stm32"
+	"unsafe"
+)
+
+// SPI phase and polarity configs CPOL and CPHA
+const (
+	SPI_ClkLowValid1stEdge  = 0
```
deadprogram: Since these are well known modes do we even need to define them? https://en.wikipedia.org/wiki/Serial_Peripheral_Interface#Mode_numbers

If we do define them, it should probably be in the `spi.go` file, and should be `Mode0` `Mode1` etc.

src/machine/machine_stm32_spi.go
```diff
@@ -0,0 +1,107 @@
...
+	var conf uint32 = spi.getBaudRate(config)
+
+	// set bit transfer order
+	if config.LSBFirst {
+		conf |= stm32.SPI_CR1_LSBFIRST
+	}
+
+	// set polarity and phase on the SPI interface
+	switch config.Mode {
+	case SPI_ClkLowValid1stEdge:
```
deadprogram: See previous comment.

src/machine/machine_stm32_spi.go
gwtnz: Oh interesting, I wasn't aware there was a widespread definition of those; I just tried to capture what was in the datasheet in a human understandable way. I'll set up the `Mode0`, `..1`, etc in `spi.c`

gwtnz:
Hi @deadprogram , changes ready for review.

I've rebased back to before the `picolibc` merge from @aykevl as that's still failing for me with the Docker image; i think it might be to do with this: https://github.com/tinygo-org/tinygo/commit/f316ebc23b025e497931cfe420ec143f1a217b57#diff-439965b0f3510a2ea8e65750e8d106d9

gwtnz:
I added a line to the Dockerfile to re-copy the missing `picolibc.h` back in

deadprogram:
Hi @gwtnz can you please submit a separate PR with the Dockerfile changes, and remove that commit from this PR? It makes it a lot easier to review/handle changes.

Other than that, this PR looks great. I am going to setup my board for a test shortly.

gwtnz:
Awesome, thanks @deadprogram !

gwtnz:
@deadprogram This PR is now ready - I pulled out the Dockerfile change, then the other PR got merged in so this PR would have removed the change after rebasing; TL;DR it should be fine now :)

deadprogram:
Confirmed SPI is working on my Bluepill. Thanks very much for another contribution @gwtnz now merging.

deadprogram:
I ended up using git cherry-pick to only pull in commit 55ce55b86a219bd77215ab9f615c6d149c155a9d in order to keep things clean.

Thanks again! Now closing.

gwtnz:
yeah in retrospect I should probably have done an interactive rebase on this branch after the other PR was merged in. A lesson for next time! :)

PR #981, opened 2020-03-23
transform: replace panics with source locations

aykevl:
Panics are bad for usability: whenever something breaks, the user is shown a (not very informative) backtrace. Replace it with real error messages instead, that even try to display the Go source location.

(This PR is only for the interface lowering pass, other passes can be done later as needed).


deadprogram:
This is already proving to be very useful. Merging.

PR #982, opened 2020-03-23
riscv: add I2C support for the SiFive HiFive1 Rev B board

yannishuber:
This adds I2C read/write support on the HiFive1 Rev B board.


deadprogram:
Hello @yannishuber this is a very interesting looking contribution.

Just wondering if you really needed to update the CMSIS repo here? I do not see any changes relative to this board, so it would probably be better not in this PR.

Other than that, have you have a chance to test it out yet? It would be great to be able to land it before the next release.

Thank you.

yannishuber:
Hello @deadprogram, you are right the update of the CMSIS repo was unintentional, I am new working with submodules and still figuring it out :grin:.

About testing, I did some tests reading from an I2C thermometer which seems to work fine. I am gonna do some additional tests in the next few days with other sensors and if everything goes well this PR will be ready for review.

src/machine/machine_fe310.go
```diff
@@ -181,3 +182,147 @@ func (spi SPI) Transfer(w byte) (byte, error) {
...
+// bytes and stores them in r, and generates a stop condition on the bus.
+func (i2c I2C) Tx(addr uint16, w, r []byte) error {
+	var err error
+	if len(w) != 0 {
+		// send start/address for write
+		i2c.sendAddress(addr, true)
+
+		// ACK received (0: ACK, 1: NACK)
+		if i2c.Bus.CR_SR.HasBits(sifive.I2C_SR_RX_ACK) {
+			return errors.New("I2C write error: expected ACK not NACK")
```
aykevl: Can you put `errors.New` error values in unexported globals? That will avoid heap allocation at runtime.
(there are two more `errors.New` calls below)

src/machine/machine_fe310.go
```diff
@@ -181,3 +182,147 @@ func (spi SPI) Transfer(w byte) (byte, error) {
...
+		for _, b := range w {
+			err = i2c.writeByte(b)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	if len(r) != 0 {
+		// send start/address for read
+		i2c.sendAddress(addr, false)
```
aykevl: This might be problematic for some I2C devices that expect the write and read to happen in a single transaction.
This doesn't need fixing right away but something to be aware of if some hardware fails.

yannishuber: I am not sure to understand why that wouldn't be the case, the stop signal to end the transaction is only generated [after](https://github.com/tinygo-org/tinygo/pull/982/files#diff-8a77a1909742be6b60f3ea35660048ddR276) the write and read.

When calling `i2c.sendAddress()` the second time a repeated start is generated but this is still the same transaction.

aykevl: Ah I see. Well I hope devices can deal with repeated starts, then (you'd be surprised how many are slightly non conforming).

PR #983, opened 2020-03-23
interp: better support interface operations

aykevl:
This commit teaches the interp scanner that supported interface operations (type assertions, interface assertions) are supported.

This fixes a problem with math/rand in Go 1.14. The change is simple but it took me quite some time to figure out.


deadprogram:
Good work @aykevl now merging.

PR #987, opened 2020-03-24
interp: improve support for load/store instructions

aykevl:
This is a combination of two commits that allows a significant number of packages to compile (perhaps most importantly `math/big`):

  * crypto/dsa
  * crypto/ed25519
  * crypto/elliptic
  * crypto/rand
  * go/constant
  * math/big

However, it also breaks the image/color/palette package. I am hopeful that this will be fixed with the removal of the `runtime.isnil` mechanism which appears to cause this problem.

### DIFF
--- a/interp/frame.go
+++ b/interp/frame.go
@@ -92,7 +92,8 @@ func (fr *frame) evalBasicBlock(bb, incoming llvm.BasicBlock, indent string) (re
 		case !inst.IsALoadInst().IsNil():
 			operand := fr.getLocal(inst.Operand(0)).(*LocalValue)
 			var value llvm.Value
-			if !operand.IsConstant() || inst.IsVolatile() || (!operand.Underlying.IsAConstantExpr().IsNil() && operand.Underlying.Opcode() == llvm.BitCast) {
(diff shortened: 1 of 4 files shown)


aykevl:
This is now ready for review.

aykevl:
Well not quite, CI broke. I think it will be fixed with the removal of `runtime.isnil` which is close to finished (but depends on #746).

jaddr2line:
Looks good. Will put in a final review once this is rebased and CI passes.

deadprogram:
@aykevl I think you can now rebase this PR, as the dependencies are merged.

aykevl:
Rebased, and now it doesn't regress for image/color/palette.

deadprogram:
Hmm somethingn went wrong there:

```
# runtime
src/device/arm/arm.go:173:7: interp: branch on a non-constant
make: *** [Makefile:186: smoketest] Error 1
```

aykevl:
Well that's unfortunate. I'll need to look into that.

deadprogram:
Can this PR now be worked on again?

aykevl:
This PR is no longer relevant since #1430 has been merged.

PR #988, opened 2020-03-25
wasm_exec: Sync polyfills with Go 1.14.1.

QuLogic:
This ports over the following three commits:
https://github.com/golang/go/commit/9627180f0f1f016307f4987cec6594baf90d64ae
https://github.com/golang/go/commit/aff2f6ece896e0fe76a2c8853abf868f689006f0
https://github.com/golang/go/commit/42b79f08239216eeea3cc1b0febc992f91bd88de
https://github.com/golang/go/commit/ecba83520d4c34870e0f5f0997d59d4496957240

This is the easy stuff, with what should be no effect on existing results.

### DIFF
--- a/targets/wasm_exec.js
+++ b/targets/wasm_exec.js
@@ -5,38 +5,40 @@
 // This file has been modified for use by the TinyGo compiler.
 
 (() => {
-	// Map web browser API and Node.js API to a single common API (preferring web standards over Node.js API).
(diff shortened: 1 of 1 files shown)


aykevl:
Thank you! No tests break so I hope no real programs break. These commits should help to be compatible with other semi-browser environments like Electron.

PR #989, opened 2020-03-25
Copy picolibc header into Dockerfile

gwtnz:
The headerfile for picolibc gets copied into the Dockerfile early on during its build, but then removed immediately afterwards when pulling in the subprojects, causing a compilation error when trying to build `.hex` files using hte Docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,8 @@ RUN cd /tinygo/ && \
     git submodule sync && \
     git submodule update --init --recursive --force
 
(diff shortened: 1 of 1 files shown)


gwtnz:
@deadprogram , @aykevl proposed Dockerfile fix ready for review

aykevl:
Looks good to me. It feels like a hack but I can't think of a better fix so OK.

gwtnz:
Yeah it's not ideal; you need the directory structure from tinygo present but clean to get the submodules to land in the right places, but also need any additional files / directories to be available too...

PR #990, opened 2020-03-25
Remove runtime.isnil

aykevl:
Finally, I think this branch is ready.

The first three commits are commits to prepare for the change in the fourth commit, which is what this PR is about. They make sure escape analysis won't regress after `runtime.isnil` has been removed.

I found exactly one regression in escape analysis: in testdata/calls.go, but only on a host OS and on WebAssembly. Further investigation revealed that it has something to do with function pointers or closures.
Most tests I've done (including smoke tests) result in a reduction of code size however some get bigger (perhaps because the inliner made different choices somewhere). None of the baremetal tests resulted in an escape analysis regression. Overall I think this is an improvement.

(Note that the GitHub ordering of commits is wrong, they're sorted by date and not by actual commit order).

### DIFF
--- a/compiler/asserts.go
+++ b/compiler/asserts.go
@@ -158,12 +158,21 @@ func (b *builder) createNilCheck(inst ssa.Value, ptr llvm.Value, blockPrefix str
 		return
 	}
 
-	switch inst.(type) {
+	switch inst := inst.(type) {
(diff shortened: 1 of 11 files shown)


deadprogram:
Now merging, thanks @aykevl

PR #991, opened 2020-03-25
Add garbage collector that uses an external allocator

jaddr2line:
In cases where tinygo is running in an environment that it does not own, it may be desirable to allow tinygo to use a C heap. This is accomplished by prepending treap nodes to the beginning of every allocation.

Marking a node is implemented by removing it from the treap and then pushing it onto a scan queue (which is implemented as a linked list using the memory previously used for the treap node). Nodes are then popped off of this queue, scanned, and then pushed onto a new treap. Once the queue is empty, all currently active nodes will have been moved from the old treap to the new treap. All nodes from the old treap are then destroyed, and the new treap is swapped in.

All operations are done non-recursively. This is still somewhat WIP, and there is a lot of debugging code still left in right now.

### DIFF
--- a/src/runtime/runtime_unix.go
+++ b/src/runtime/runtime_unix.go
@@ -24,10 +24,6 @@ func exit(code int)
 //go:export clock_gettime
 func clock_gettime(clk_id int32, ts *timespec)
 
-const heapSize = 1 * 1024 * 1024 // 1MB to start
-
-var heapStart, heapEnd uintptr
-
 type timeUnit int64
 
 const tickMicros = 1
@@ -47,8 +43,7 @@ func postinit() {}
 // Entry point for Go. Initialize all packages and call main.main().
 //go:export main
 func main() int {
-	heapStart = uintptr(malloc(heapSize))
-	heapEnd = heapStart + heapSize
+	preinit()
 
 	run()
 
@@ -83,3 +78,10 @@ func ticks() timeUnit {
 func syscall_Exit(code int) {
 	exit(code)
 }
+
+func extalloc(size uintptr) unsafe.Pointer {
+	return malloc(size)
+}
+
+//go:export free
+func extfree(ptr unsafe.Pointer)
--- a/src/runtime/runtime_unix_heap.go
+++ b/src/runtime/runtime_unix_heap.go
@@ -1,0 +1,14 @@
+// +build darwin linux,!baremetal freebsd,!baremetal
+
+// +build gc.conservative gc.leaking
+
+package runtime
+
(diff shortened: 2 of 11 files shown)


jaddr2line:
This PR should be ready now.

src/runtime/runtime_unix.go
```diff
@@ -83,3 +78,10 @@ func ticks() timeUnit {
 func syscall_Exit(code int) {
 	exit(code)
 }
+
+func extalloc(size uintptr) unsafe.Pointer {
+	return malloc(size)
```
aykevl: To consider: you could use `calloc` instead and avoid zeroing the memory. In some cases this might be faster or use less memory. For example, for large allocations on Linux the libc will likely return a new `mmap` area that has been zeroed by the OS already (and might not even have RAM to back it).

src/runtime/runtime_unix_heap.go
```diff
@@ -0,0 +1,14 @@
+// +build darwin linux,!baremetal freebsd,!baremetal
```
aykevl: I only now spot this `freebsd,!baremetal` in the code, which doesn't make a lot of sense. The fact that there is `linux,!baremetal` is because baremetal targets pretend to be linux/arm to get the standard library to compile. Such a `!baremetal` exclusion is not necessary for FreeBSD.
But this is not something that needs fixing in this PR.

src/runtime/runtime_unix.go
jaddr2line: I am going to leave this, because I do not want to impose any constraints on any future external allocators.

jaddr2line:
> Ideally ./testdata/gc.go on the host, but those integration tests are currently a bit inflexible.

This has actually turned out to be surprisingly effective, and was able to catch tons of bugs.

> Until then I think it would be best to include a smoke test with this GC on the host OS to see whether it compiles at all.

I am not quite sure what you mean. Don't the existing tests already check that they compile?

aykevl:
> > Until then I think it would be best to include a smoke test with this GC on the host OS to see whether it compiles at all.
> 
> I am not quite sure what you mean. Don't the existing tests already check that they compile?

Ah, now I see. You also switched the default GC on Linux etc. Then it's fine.

One thing I only now notice: some documentation doesn't include the new GC type, such as the help text in the `tinygo` command (main.go) and the function documentation for `(*compileopts.Config).GC`.

jaddr2line:
Alright, I think I updated all of the docs in this repo.

aykevl:
Thanks for updating!

PR #992, opened 2020-03-26
runtime (gc): bundle small allocations together

jaddr2line:
This reduces the overhead of small allocations. Previously a pointer sized allocation would take up an entire block - leaving 3 pointer-widths of wasted space. Small allocations are now bundled together, and are treated as a single object for garbage-collection purposes. These could be tracked separately, but not without paying for space to store additional garbage collector metadata.

Most tests in `testdata` call `smallAlloc` at some point.

### DIFF
--- a/src/runtime/gc_conservative.go
+++ b/src/runtime/gc_conservative.go
@@ -213,10 +213,83 @@ func initHeap() {
 // collection cycle if needed. If no space is free, it panics.
 //go:noinline
 func alloc(size uintptr) unsafe.Pointer {
-	if size == 0 {
-		return unsafe.Pointer(&zeroSizedAlloc)
+	var ptr unsafe.Pointer
+	switch {
+	case size == 0:
+		// Use a fixed location for all zero-sized allocations.
+		ptr = unsafe.Pointer(&zeroSizedAlloc)
+	case size < bytesPerBlock:
+		// Bundle small allocations together.
+		ptr = smallAlloc(size)
+	default:
+		// Allocate multiple memory blocks.
+		ptr = bigAlloc(size)
 	}
 
+	// Zero the allocation.
+	memzero(ptr, size)
+
+	return ptr
+}
+
+// alignMask computes a bit mask which contains all bits which must be zero in the maximum alignment of a pointer of the given size.
+func alignMask(size uintptr) uintptr {
+	switch {
+	case size <= unsafe.Sizeof(uint8(0)):
+		return unsafe.Alignof(uint8(0)) - 1
+	case size <= unsafe.Sizeof(uint16(0)):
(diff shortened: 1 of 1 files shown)


src/runtime/gc_conservative.go
```diff
@@ -213,10 +213,92 @@ func initHeap() {
...
+	x = x | (x >> 2)
+	x = x | (x >> 4)
+	x = x | (x >> 8)
+	x = x | (x >> 16)
+	x = x | (x >> 32)
+
+	// The highest bit's value is the alignment.
+	// Shifting down gives us a mask that is one less than the alignment.
+	return x >> 1
+}
```
aykevl, line 251: I'm not sure what this does? Align the value to the size of the block?

jaddr2line: This computes a mask which can be used to align the allocation within the block. The maximum possible alignment of a type is the largest power of 2 less than or equal to the size. This returns a bitmask which is equal to `alignment-1`.

aykevl: The following code costs the same in code size (tested on a microbit) and to me is much more readable:

```go
        switch {
        case size <= 1:
                return 0
        case size <= 3:
                return 1
        case size <= 7:
                return 3
        default:
                return 7
        }
```

At the same time, the maximum alignment of anything on ARM, as far as I know, is 4. (The exception being the stack, but I think that's just the ABI and not the hardware - and we don't currently guarantee 8-byte alignment on the heap either). On AVR there is no alignment on RAM addresses. So it would be a waste to do 8-byte alignment on these platforms. You could perhaps use the `align` function or `unsafe.Alignof` to check for this.

Perhaps this?

```go
        switch {
        case size <= 1:
                return unsafe.Alignof(uint8(0)) - 1
        case size <= 3:
                return unsafe.Alignof(uint16(0)) - 1
        case size <= 7:
                return unsafe.Alignof(uint32(0)) - 1
        default:
                return unsafe.Alignof(uint64(0)) - 1
        }
```

aykevl:
I am somewhat cautious with this change. The benefit here is not nearly as obvious as with code size decreases or correctness fixes. In fact, a quick test with `tinygo build -o test.elf -size=short -target=microbit ./testdata/gc.go` shows a code size increase of 160 bytes and a static RAM increase of 16 bytes:

```
   2496       4    2276 |    2500    2280

   2656       4    2292 |    2660    2296
```

Therefore, I think the improvement should be measured in a real-world piece of code (such as the mqtt code we've been using for testing so far). How much is heap usage reduced? Is that reduction worth the cost, especially for small systems with little RAM (that would potentially benefit the most)?

niaow:
I think this can be done more efficiently with #1181 if we decide we want this

PR #994, opened 2020-03-26
wasm_exec: Port over safe features

QuLogic:
These are only new things on the wasm side, so they should do nothing until the Go side starts calling them, which should only be in Go 1.14.

### DIFF
--- a/targets/wasm_exec.js
+++ b/targets/wasm_exec.js
@@ -306,6 +306,13 @@
 						Reflect.set(v, p, x);
 					},
 
+					// func valueDelete(v ref, p string)
+					"syscall/js.valueDelete": (v_addr, p_ptr, p_len) => {
+						const v = loadValue(v_addr);
+						const p = loadString(p_ptr, p_len);
+						Reflect.deleteProperty(v, p);
+					},
+
 					// func valueIndex(v ref, i int) ref
 					"syscall/js.valueIndex": (ret_addr, v_addr, i) => {
 						storeValue(ret_addr, Reflect.get(loadValue(v_addr), i));
--- a/targets/wasm_exec.js
+++ b/targets/wasm_exec.js
@@ -381,6 +381,23 @@
 					//	mem().setUint8(sp + 24, loadValue(sp + 8) instanceof loadValue(sp + 16));
 					//},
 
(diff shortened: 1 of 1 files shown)


aykevl:
Have you tested whether the new functions (`syscall/js.copyBytesToGo` and `syscall/js.valueDelete`) work with TinyGo? The calling convention is very different and it's easy to break something.

deadprogram:
Any wasm people able to test this branch? @johanbrandhorst @justinclift

QuLogic:
Well, `copyBytesToGo` is just a copy of `copyBytesToJS` with the argument types reversed, and I tried to implement `valueDelete` the same way as the other `value*` functions, so they _should_ be correct. 

I have finished porting the other (breaking) changes to `wasm_exec.js` and it's currently crashing. I haven't figured out if I made a mistake here or in the later changes, so please hold off merging this one until I can figure that out.

justinclift:
As a visual look over, the order of the parameters to the function doesn't look correct yet:

```
"syscall/js.copyBytesToGo": (ret_addr, dest_addr, dest_len, dest_cap, source_addr) => {
```

That's from the perspective of looking at the existing (working) js.copyBytesToJS function signature:

```
"syscall/js.copyBytesToJS": (ret_addr, dest_addr, source_addr, source_len, source_cap) => {
```

The parameters there are in two groups:

* `ret_addr`, and `dest_addr`
* `source_addr`, `source_len`, and `source_cap`

Those two groups need to stick together, as the first is really a small structure for passing back the return values (`ret_addr`: "Boolean 'did it work?' value", `dest_addr`: "This is the address of the copied data"), and the second is a triplet that is effectively a Go slice structure.

Pretty sure yours will need to have a grouping something like that, though I'm not sure of the order and don't have time/interest to look at this just yet (am trying out Rust atm).

justinclift:
Just took some time looking over the function signature parameters, and they look ok.  Initially thought they weren't, but it does seem right.

That being said, I don't really have time/interest to test this properly.

deadprogram:
Is this PR ready to merge now?

deadprogram:
Since this only adds new functions as pointed out by @QuLogic and also since it is only required when Go 1.14 is used, I am going to now merge. Thanks for another contribution @QuLogic

PR #845, opened 2020-01-12
wasm: include wasi-libc

aykevl:
This allows CGo code to call some libc functions.

Fixes part of #844.

Issue #285, opened 2019-04-16
Compile packages independently


aykevl:
Packages should be compiled independently and then linked together. This provides several advantages in the long term:

* Packages can be compiled in parallel so compile time can be shorter.
* Unchanged packages can be cached, further cutting down on compile time.
* We can optionally avoid LTO to make incremental (debug) builds _much_ faster.

I would strive for the following architecture:

* Change the compiler package to compile just a single package (and not dependencies).
* Add a linker package that links several packages together and performs whole-program optimizations on them (LTO).
* Add a driver package (or repurpose the loader package?) that orchestrates all this work.
* Add some subcommands (`go tool compile`, `go tool link`) that call the compiler/linker directly. This can be useful for debugging.

I have taken a look at the [Go build cache](https://github.com/golang/go/blob/master/src/cmd/go/internal/cache/cache.go) and I think I'd implement a similar system. Every package has a hash of the input (build tags, file list, hashes of files, output hashes of packages it depends on) and produces an output hash, which is a hash over all the artifacts that it builds (mostly the serialized exported types).

My thinking is that a cached package would be an uncompressed zip file stored in the cache directory, with the following contents as files:

 * a list of packages it depends upon and their hashes (plain text)
 * a list of files included in the build and their mtimes / hashes (plain text)
 * a serialized form of the public API (some binary format, like [this one](https://golang.org/src/go/internal/gcimporter/gcimporter.go))
 * LLVM bytecode for this file (including C/C++ files, they are all linked together in one module / compile unit)

This is most likely too big to do at once, so I'd do it in multiple steps. But I'm creating this issue to give a high-level overview of what I would want to achieve at some point.

aykevl:
Some statistics when building `-target=circuitplay-express ./testdata/stdlib.go`, one of the slowest compiles at the moment:

```
  load:    37.230169ms
  parse:   90.345719ms
  types:   211.548187ms
  dce:     11.11218ms
  irgen:   700.464742ms
compile: 1.214556923s
interp:  1.264428106s
opt:     4.707457718s
codegen: 475.201528ms
link:    65.829285ms
total:   8.211343631s
```

The entire compiler runs serially at the moment, while most of this could be parallelized and/or cached.

 * the load step (which determines dependencies and to-be-compiled files) cannot be entirely parallelized, but dependencies can be loaded in parallel. However, there is not much to be gained here.
[...]

PR #995, opened 2020-03-26
LLVM memory builtins

aykevl:
Use LLVM intrinsics for memory operations (`memset`, `memcpy`, `memmove`). This gives more information to the optimizer and in the end lowers to libc functions instead of a custom TinyGo implementation. This is possible since #871 and #845.

Another motivation for this PR is to reduce the amount of code at the end of the `compiler.Compile` function, which should help with eventually breaking it down further into per-package compiles (#285).

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -330,7 +330,6 @@ func Compile(pkgName string, machine llvm.TargetMachine, config *compileopts.Con
 		return c.ctx.CreateEnumAttribute(attrKind, 0)
 	}
 	nocapture := getAttr("nocapture")
-	writeonly := getAttr("writeonly")
 	readonly := getAttr("readonly")
 
 	// Tell the optimizer that runtime.alloc is an allocator, meaning that it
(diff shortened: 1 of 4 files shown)


deadprogram:
This absolutely seems like a much better idea to me.

However, I am noticing that all of the programs are now a little larger, but a couple of them are now much larger than I would have expected.

Before this PR:

```
tinygo build -size short -o test.hex -target=itsybitsy-m0        examples/blinky1
   9172       8    4312 |    9180    4320
```

After this PR:

```
tinygo build -size short -o test.hex -target=itsybitsy-m0        examples/blinky1
   9548       8    4312 |    9556    4320
```

What do you think?

aykevl:
It definitely is `machine.sendUSBPacket`.

Before (at c4fd19be99c2695f25ac59057adef7a44ad61ad9):

```
   9380       8    4312 |    9388    4320
```

With this PR:

```
   9548       8    4312 |    9556    4320
```

With this PR and `//go:noinline` on `machine.sendUSBPacket`:

```
   9308       8    4312 |    9316    4320
```

So it definitely _can_ reduce code size, with a hint to the inliner. I think the inliner considers the LLVM builtin to be of a lower cost than a regular function call, although it is often translated to a function call in the end. Because of this lower cost, the inliner decided that `machine.sendUSBPacket` is just below the limit and decided to inline the function at all 11 places.
So, what do you think? I can add a commit to stop that function from being inlined.

---

Sidenote: Clang does this memory builtin translation too. When you call `memcpy` for example, it will emit a call to `llvm.memcpy.<something>` instead of a regular memcpy call.

deadprogram:
Very good investigative work to figure out the root cause. Sounds like a good solution to me.

So I assume that be a separate PR, in which case should I should merge this one when that one is ready?

aykevl:
Sounds good to me!

deadprogram:
Also now merging this. Good work!

PR #996, opened 2020-03-26
sync: implement sync.Map

aykevl:
This is a very simple implementation, just enough to get packages to compile.


deadprogram:
OK, merging. Thanks!

PR #997, opened 2020-03-27
Export implementations of malloc/free for use from C

bgould:
These are implementations of malloc and free that delegate to the GC implementation.

Note: I'm not totally sure what the correct build tag to use for this is, if any.  `baremetal` seems to work well but please feel free to suggest an alternative if there is something more appropriate.  Also if you think that there is so better design for how these should be implemented that is fine too.


aykevl:
I have done some size tests to make sure this doesn't blow up size unexpectedly, and it only seems to increase code size on very small blinking LED samples. Any real programs will have no or very moderate (16 bytes) increases in binary size, so I think it's fine.

Because this code is only relevant on baremetal targets, I think it makes sense to move it to src/runtime/baremetal.go, which is already protected by the same build constraint.

src/runtime/runtime_libc.go
```diff
@@ -0,0 +1,15 @@
+// +build baremetal
+
+package runtime
+
+import "unsafe"
+
+//go:export malloc
```
aykevl: The typical Go version is `//export`, not `//go:export`. TinyGo treats those identically, but I think it's best to move to `//export` everywhere (and not introduce new `//go:export` pragmas) to be consistent with the wider Go ecosystem.

The same goes for `libc_free` below.

aykevl:
Thank you! Squashed/merged your changed in the dev branch.

src/runtime/runtime_libc.go
bgould: Done

PR #998, opened 2020-03-27
machine: avoid binary size regression after LLVM memory intrinsics

aykevl:
Somehow moving to LLVM memory intrinsics for calls like memcpy made the
machine.sendUSBPacket get inlined. This is a problem because it is
called in many different functions and it is just big enough to cause a
significant file size increase.

Adding //go:noinline solves this problem and gets the examples/blinky1
program below the file size it was before this change (tested:
itsybitsy-m0, itsybitsy-m4, circuitplay-bluefruit).

This PR fixes the binary size regression seen in #995.

### DIFF
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -1736,6 +1736,7 @@ func cdcSetup(setup usbSetup) bool {
 	return false
 }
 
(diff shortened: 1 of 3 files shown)


deadprogram:
Thanks @aykevl now merging.

PR #1000, opened 2020-03-28
compiler: add support for anonymous type asserts

aykevl:
This is used for example by the errors package, which contains:

    if x, ok := err.(interface{ As(interface{}) bool }); ok && x.As(target) {
        return true
    }

The interface here is not a named type.


aykevl:
Also, this is issue/PR number 1000!

deadprogram:
Thanks @aykevl great milestone. Also a pretty sophisticated language feature, which seems fitting.

Now merging.

PR #1001, opened 2020-03-28
compiler: add debug info to goroutine start wrappers

aykevl:
In an effort to add debug info to _all_ functions, this PR adds debug information to goroutine start wrappers.


deadprogram:
Merging. :bug:

PR #1002, opened 2020-03-28
internal/task: add interrupt-safe queue

jaddr2line:
This commit adds a queue wrapper type which can be used to transfer tasks between interrupts and the scheduler.

### DIFF
--- a/src/internal/task/interruptqueue.go
+++ b/src/internal/task/interruptqueue.go
@@ -1,0 +1,66 @@
+package task
+
+import "runtime/volatile"
+
+// InterruptQueue is a specialized version of Queue, designed for working with interrupts.
+// It can be safely pushed to from an interrupt (assuming that a memory reference to the task remains elsewhere), and popped outside of an interrupt.
+// It cannot be pushed to outside of an interrupt or popped inside an interrupt.
+type InterruptQueue struct {
+	// This implementation uses a double-buffer of queues.
+	// bufSelect contains the index of the queue currently available for pop operations.
+	// The opposite queue is available for push operations.
+	bufSelect volatile.Register8
+	queues    [2]Queue
+}
+
+// Push a task onto the queue.
+// This can only be safely called from inside an interrupt.
+func (q *InterruptQueue) Push(t *Task) {
+	// Avoid nesting interrupts inside here.
+	var nest nonest
+	nest.Lock()
+	defer nest.Unlock()
+
+	// Push to inactive queue.
+	q.queues[1-q.bufSelect.Get()].Push(t)
+}
+
+// Check if the queue is empty.
+// This will return false if any tasks were pushed strictly before this call.
+// If any pushes occur during the call, the queue may or may not be marked as empty.
+// This cannot be safely called inside an interrupt.
+func (q *InterruptQueue) Empty() bool {
+	// Check currently active queue.
+	active := q.bufSelect.Get() & 1
+	if !q.queues[active].Empty() {
+		return false
+	}
+
+	// Swap to other queue.
+	active ^= 1
+	q.bufSelect.Set(active)
+
+	// Check other queue.
+	return q.queues[active].Empty()
+}
+
+// Pop removes a single task from the queue.
+// This will return nil if the queue is empty (with the same semantics as Empty).
+// This cannot be safely called inside an interrupt.
+func (q *InterruptQueue) Pop() *Task {
+	// Select non-empty queue if one exists.
+	if q.Empty() {
+		return nil
+	}
+
+	// Pop from active queue.
+	return q.queues[q.bufSelect.Get()&1].Pop()
+}
+
+// AppendTo pops all tasks from this queue and pushes them to another queue.
+// This operation has the same semantics as repeated calls to pop.
+func (q *InterruptQueue) AppendTo(other *Queue) {
+	for !q.Empty() {
+		q.queues[q.bufSelect.Get()&1].AppendTo(other)
+	}
+}
--- a/src/internal/task/nonest_arm.go
+++ b/src/internal/task/nonest_arm.go
@@ -1,0 +1,20 @@
+// +build arm,baremetal,!avr
+
+package task
+
+import "device/arm"
+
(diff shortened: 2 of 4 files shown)


src/internal/task/interruptqueue.go
```diff
@@ -0,0 +1,66 @@
...
+	queues    [2]Queue
+}
+
+// Push a task onto the queue.
+// This can only be safely called from inside an interrupt.
+func (q *InterruptQueue) Push(t *Task) {
+	// Avoid nesting interrupts inside here.
+	var nest nonest
+	nest.Lock()
+	defer nest.Unlock()
```
aykevl, line 22: I think it would be better to avoid `defer` here. `defer` might allocate heap memory (hopefully not in this case!) and the heap is currently not interrupt-safe and likely will never be.

src/internal/task/interruptqueue.go
```diff
@@ -0,0 +1,66 @@
+package task
+
+import "runtime/volatile"
+
+// InterruptQueue is a specialized version of Queue, designed for working with interrupts.
+// It can be safely pushed to from an interrupt (assuming that a memory reference to the task remains elsewhere), and popped outside of an interrupt.
+// It cannot be pushed to outside of an interrupt or popped inside an interrupt.
+type InterruptQueue struct {
```
aykevl, line 8: Do you have a reference for this scheme, or is this something you came up with yourself? I believe you once said it was based on interrupt/scheduler communication in another RTOS.
If it is based on an existing system, a reference would be very useful to understand the code.

src/internal/task/interruptqueue.go
```diff
@@ -0,0 +1,66 @@
...
+	// This implementation uses a double-buffer of queues.
+	// bufSelect contains the index of the queue currently available for pop operations.
+	// The opposite queue is available for push operations.
+	bufSelect volatile.Register8
```
aykevl, line 12: tl;dr: volatile is probably fine (no change needed) but atomics might be more appropriate.

I wonder if this should use atomics instead? (See #1113). Atomics have somewhat different guarantees and are intended for synchronization (where volatile is intended for memory-mapped I/O). In practice it will usually have the same effect, especially on microcontrollers.

Specifically:

1. Volatile guarantees there is exactly one load or store instruction. This is necessary for some memory-mapped I/O but prevents certain (probably irrelevant) compiler optimizations where values are speculatively loaded when they might not be needed. It does not guarantee the value is loaded or stored in a single operation (although, unofficially, you can rely on there being only one operation if the hardware has a load/store instruction for the given bit width).
2. Atomics guarantee that (from the perspective of other threads or interrupts), the value is loaded/stored atomically. This means, for example, that a 64-bit value is also loaded/stored atomically even if the processor does not support such an operation. In the case of Go (and usually C++ too), the operations will happen in the expected order respective to similar atomic operations (but not necessarily to regular load/store instructions). This is mostly relevant to out-of-ourder and/or multicore systems.

I think volatile will be fine in this case, but just wanted to warn for this particular (somewhat semantic) difference. No change is needed from my POV.

aykevl: Also, maybe something like `popSelect` or `receiverSelect` would be clearer? Because apparently it's the queue used by the receiver, and the sender (pusher) uses the other one. This is not immediately clear from the name `bufSelect`.

(Just a suggestion, not necessary to change).

src/internal/task/interruptqueue.go
```diff
@@ -0,0 +1,66 @@
...
+	nest.Lock()
+	defer nest.Unlock()
+
+	// Push to inactive queue.
+	q.queues[1-q.bufSelect.Get()].Push(t)
+}
+
+// Check if the queue is empty.
+// This will return false if any tasks were pushed strictly before this call.
+// If any pushes occur during the call, the queue may or may not be marked as empty.
```
aykevl, line 30: > the queue may or may not be marked as empty.

Could be me, but 'marked' sounds like it changes the queue, while it doesn't (at least, not in a relevant way). It only returns the current state of the queue.

src/internal/task/interruptqueue.go
aykevl: Also, I wonder if this could be written more clearly like so?

```go
var mask uintptr
if interrupt.SupportsNesting {
    // If the system allows nested interrupts, disable them during this operation to make it atomic.
    mask := interrupt.DisableInterrupts()
}
q.queues[1-q.bufSelect.Get()].Push(t)
if interrupt.SupportsNesting {
    interrupt.EnableInterrupts(mask)
}
```

That's a bit more verbose and of course needs some additions to the runtime/interrupts package, but it does look more obvious to me. What do you think @jaddr2line?
One thing in particular that makes me a little bit nervous is that the `nonest` mechanism essentially has a blacklist of architectures that allow nesting, but it is certainly possible more get added in the future (for example, it's possible to have nested interrupts in RISC-V although the current PLIC makes it a bit cumbersome).

src/internal/task/interruptqueue.go
niaow: An older implementation of InterruptQueue was based off of https://arxiv.org/pdf/0709.4558.pdf

However, it was complicated, and we can do better on platforms with atomics, so I just decided to swap it with a double buffer.

The scheme of seperating the interrupt queue from the runqueue was something I came up with myself, and it seemed to make more sense based on the constraints involved in TinyGo.

src/internal/task/interruptqueue.go
niaow: So this intentionally does not use atomics. On platforms with a hardware atomic CAS, this can be done a lot more efficiently. I wanted to separate this to have a different approach.

That said, I could probably get this to work with atomic loads and stores instead of volatile if we added an atomic 8-bit load/store, as this is what is needed to function efficiently on AVR.

aykevl: Ah I see. Note that with #1113, atomics are supported in most places. Not yet on AVR, but that should not be very difficult to implement (although it might be inefficient). 32-bit load/store atomic instructions on Cortex-M are basically free so they can be used.

niaow:
Closing in favor of #1142

PR #1006, opened 2020-03-28
runtime (scheduler): generalize scheduler beyond timers

jaddr2line:


src/runtime/scheduler_legacy.go
```diff
@@ -0,0 +1,100 @@
...
+		sleepQueue = t.Next
+		t.Next = nil
+		runqueue.Push(t)
+		awoke = true
+	}
+
+	return awoke
+}
+
+var curTime timeUnit
```
aykevl: Suggestion: I personally prefer to put global variables and consts at the top of the file, to make them more visible. I'm not aware of any official guidelines in this regard but it's something which I often see in code (including Go code).

jaddr2line: I tend to put variables closer to their use, as this makes it easier to track what they are used for.

src/runtime/scheduler.go
```diff
@@ -74,95 +70,51 @@ func runqueuePushBack(t *task.Task) {
...
-	if *q != nil {
-		// cut delay time between this sleep task and the next
-		(*q).Data -= t.Data
-	}
-	t.Next = *q
-	*q = t
-}
+const pollInterval = 256
 
 // Run the scheduler until all tasks have finished.
```
aykevl, line 77: I do not believe this sentence is true anymore with this PR: the scheduler runs until the main task has finished.

src/runtime/scheduler_legacy.go
```diff
@@ -0,0 +1,100 @@
...
+
+// poll checks for any events that are ready and pushes them onto the runqueue.
+// It returns true if any tasks were ready.
+func poll() bool {
```
aykevl, line 68: The name of this function is rather broad. Perhaps something like `pollForEvents` or `checkForNewTasks`? Or `fillRunqueue`?

src/runtime/scheduler_legacy.go
```diff
@@ -0,0 +1,100 @@
...
+	scheduleLog("  polling for events")
+	if sleepQueue == nil {
+		return false
+	}
+	curTime = ticks()
+	return pollSleepQueue(curTime)
+}
+
+// wait sleeps until any tasks are awoken by external events.
+func wait() {
```
aykevl, line 78: Same for this function. I think something like `waitForEvents` would be clearer.
(I'm assuming this will wait for the next interrupt soon, with `wfe`?)

niaow: Yes, #1040 adds an implementation for atsamd21 which directly uses `wfe`.

niaow:
Closing in favor of #1142

PR #1007, opened 2020-03-28
machine/atsamd21,atsamd51,nrf52840: correct USB CDC

deadprogram:
This PR corrects the USB CDC composite descriptors for the ATSAMD21, ATSAMD51, and NRF52840 processors.

Turns out that the clues provided by @sago35 were correct, the descriptor needed to be changed.

I tested the `example/echo` program with these boards: 

- Circuit Playground Express
- ItsyBitsy M4
- Circuit Playground Bluefruit

On these OS:

- Ubuntu
- macOS
- Windows 10.

### DIFF
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -1478,7 +1478,7 @@ func handleUSB(intr interrupt.Interrupt) {
 			setEPINTENCLR(0, sam.USB_DEVICE_EPINTENCLR_STALL1)
 		}
 	} else {
-		sendZlp(0)
+		sendZlp()
 	}
 
 	// Now the actual transfer handlers, ignore endpoint number 0 (setup)
@@ -1601,7 +1601,7 @@ func handleStandardSetup(setup usbSetup) bool {
 		} else if setup.wValueL == 0 { // ENDPOINTHALT
 			isEndpointHalt = false
 		}
-		sendZlp(0)
+		sendZlp()
 		return true
 
 	case usb_SET_FEATURE:
@@ -1610,7 +1610,7 @@ func handleStandardSetup(setup usbSetup) bool {
 		} else if setup.wValueL == 0 { // ENDPOINTHALT
 			isEndpointHalt = true
 		}
-		sendZlp(0)
+		sendZlp()
 		return true
 
 	case usb_SET_ADDRESS:
@@ -1665,7 +1665,7 @@ func handleStandardSetup(setup usbSetup) bool {
 			// Enable interrupt for CDC data messages from host
 			setEPINTENSET(usb_CDC_ENDPOINT_OUT, sam.USB_DEVICE_EPINTENSET_TRCPT0)
 
-			sendZlp(0)
+			sendZlp()
 			return true
 		} else {
 			return false
@@ -1679,7 +1679,7 @@ func handleStandardSetup(setup usbSetup) bool {
 	case usb_SET_INTERFACE:
 		usbSetInterface = setup.wValueL
 
-		sendZlp(0)
+		sendZlp()
 		return true
 
 	default:
@@ -1724,14 +1724,14 @@ func cdcSetup(setup usbSetup) bool {
 			} else {
 				// TODO: cancel any reset
 			}
-			sendZlp(0)
+			sendZlp()
 		}
 
 		if setup.bRequest == usb_CDC_SEND_BREAK {
 			// TODO: something with this value?
 			// breakValue = ((uint16_t)setup.wValueH << 8) | setup.wValueL;
 			// return false;
-			sendZlp(0)
+			sendZlp()
 		}
 		return true
 	}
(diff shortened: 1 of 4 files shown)


src/machine/machine_atsamd51.go
```diff
@@ -1967,53 +1969,29 @@ func sendDescriptor(setup usbSetup) {
...
-			b := make([]byte, len(prod)*2+2)
-			b[0] = byte(len(prod)*2 + 2)
-			b[1] = 0x03
-
-			for i, val := range prod {
-				b[i*2+2] = val
-				b[i*2+3] = 0
-			}
-
+			b := strToUTF16LEDescriptor(usb_STRING_PRODUCT)
```
aykevl: Note that this will likely result in an extra heap allocation. You can avoid this by allocating the buffer here and passing it to `strToUTF16LEDescriptor`, for example.

src/machine/machine_atsamd51.go
```diff
@@ -1967,53 +1969,29 @@ func sendDescriptor(setup usbSetup) {
...
-			b := make([]byte, len(prod)*2+2)
-			b[0] = byte(len(prod)*2 + 2)
-			b[1] = 0x03
-
-			for i, val := range prod {
-				b[i*2+2] = val
-				b[i*2+3] = 0
-			}
-
+			b := strToUTF16LEDescriptor(usb_STRING_MANUFACTURER)
```
aykevl: Same here.

src/machine/machine_atsamd51.go
deadprogram: I added another commit that addresses this.

src/machine/machine_atsamd51.go
deadprogram: Same as above.

deadprogram:
i added another commit to this PR that both addresses the heap allocation. More importantly, it removes a bunch of code duplication by moving the creation of the device descriptor into shared code for all USB CDC supported boards.

aykevl:
That looks much better!

jaddr2line:
This commit appears to have broken USB registration with my desktop.

deadprogram:
@jaddr2line what OS are you running on desktop? Also, what program are you running on your Arduino Nano33 board? You should try to test with `examples/serial` and `examples/echo`.

jaddr2line:
Linux (Manjaro) 4.14-rt, this is also the same machine that used to time out in the USB interrupt

deadprogram:
Now that we are sending the correct vendor and product id, you might need a udev rule for them to be able to be used by non-root user.

jaddr2line:
No, that is not it. The device is not enumerating. From `dmesg`:
```
[58032.758947] usb 1-7: new full-speed USB device number 73 using xhci_hcd
[58032.915657] usb 1-7: device descriptor read/64, error -71
[58033.152321] usb 1-7: device descriptor read/64, error -71
[58033.385607] usb 1-7: new full-speed USB device number 74 using xhci_hcd
[58033.542360] usb 1-7: device descriptor read/64, error -71
[58033.779007] usb 1-7: device descriptor read/64, error -71
[58033.885653] usb usb1-port7: attempt power cycle
[58034.528961] usb 1-7: new full-speed USB device number 75 using xhci_hcd
[58034.610115] usb 1-7: device descriptor read/8, error 0
[58034.740124] usb 1-7: device descriptor read/8, error 0
[58034.968948] usb 1-7: new full-speed USB device number 76 using xhci_hcd
[58035.046187] usb 1-7: device descriptor read/8, error 0
[58035.179221] usb 1-7: device descriptor read/all, error 0
[58035.179264] usb usb1-port7: unable to enumerate USB device
```

src/machine/machine_atsamd21.go
```diff
@@ -1477,6 +1477,8 @@ func handleUSB(intr interrupt.Interrupt) {
 			// clear stall request
 			setEPINTENCLR(0, sam.USB_DEVICE_EPINTENCLR_STALL1)
 		}
+	} else {
+		sendZlp()
```
jaddr2line: This line appears to have caused the regression. What is it supposed to be doing?

deadprogram: Interesting. I added to send a zlp when the other cases were not handled, which seemed to correct an issue with the USB enumeration on Windows 10, based on a comment from @sago35

Removing that line gets it to work on Linux and macOS, but does not work on Windows 10. I suppose that means we need to further investigate what Windows in sending as the initial request.

sago35: I made a proposal for sendZlp() below and in the comments that follow.
However, I'm not sure that this idea is correct.
https://github.com/tinygo-org/tinygo/pull/884#issuecomment-594490977

The samd51 I'm trying has five USB Device EndPoint Interrupt Flags.

https://www.mouser.com/datasheet/2/268/60001507A-1130176.pdf
P.1161.

I did an additional check on atsamd51 and USB-CDC worked if you called sendZlp() on TRCPT0.
It may be better to process TRCPT0 as follows.
My environment is Windows 10 + PyPortal (samd51).

https://github.com/sago35/tinygo/commit/f80cc339e5c2f18b559a2f94f59352c05329a816.

Since sendZlp() clears the bottom 12 bits of PCKSIZE, it assumes that BYTE_COUNT = 0.
The PCKSIZE is described on P.1197.

sago35: URB_FUNCTION_GET_MS_FEATURE_DESCRIPTOR, which probably exists only in Windows, needs to be handled well.
Maybe a careful reading of the Aruduino and circuitpython sources will solve the problem.

https://github.com/tinygo-org/tinygo/pull/884#issuecomment-594490977

sago35: https://github.com/adafruit/ArduinoCore-samd/blob/master/cores/arduino/USB/USBCore.cpp

sago35: This process is not in machine_atsamd51.go.
On the else side, sendZlp () may not be necessary or should not be called.

https://github.com/adafruit/ArduinoCore-samd/blob/master/cores/arduino/USB/USBCore.cpp#L951

deadprogram: Please see https://github.com/tinygo-org/tinygo/pull/1013 I think you have the right idea @sago35

PR #466, opened 2019-07-30
Avoid DCE at Go SSA level

aykevl:
Large change that:

1. Adds support for calling indirect functions, like the following (as used in the time package):
    ```go
    func foo(fn func()) {
        go fn()
    }
    ```
2. Compiles all code, instead of just the part that seems to be used.

This is in preparation of fixing #285. Not ready for merging, as it's not complete (some packages that previously compiled break with this change) and I want to review/improve it myself a bit, and add real commit messages.

This change will increase compile times, but should decrease massively once #285 is ready and package builds can be cached.

PR #1008, opened 2020-03-29
Remove SimpleDCE and add compiler testing

aykevl:
The first commit removes most of the ir package and either moves the code to compiler/symbol.go or removes it altogether. After this, not much remains of the ir package and it should be removed eventually.

The second commit relies on the added flexibility provided by the first commit to add testing to the compiler package. It only contains one very basic test but it provides a framework for future tests to be added in a similar way.

Warning: this PR may have consequences. I found that the image/png package stopped compiling due to a bug in the compiler (it produces invalid IR), probably because a function that previously was removed with SimpleDCE is now included in the build. It also slows down the compiler significantly, I've measured +12% (driver smoke tests) to +53% (tinygo smoke tests) increase in compile time. However, this PR is one of the many changes necessary to allow packages to be compiled independently which should eventually speed up the compiler by compiling packages in parallel and caching already compiled packages.

Replaces #466.


deadprogram:
This PR now has merge conflicts that need to be resolved.

aykevl:
Rebased. This PR should not be merged before the next release, it's too big and invasive and relies on followup changes (not yet made) to shrink compile time again.

aykevl:
> I found that the image/png package stopped compiling due to a bug in the compiler (it produces invalid IR), probably because a function that previously was removed with SimpleDCE is now included in the build.

This regression has now been fixed by #1040.

aykevl:
This PR has been superseded by other (merged) PRs.

PR #1010, opened 2020-03-29
runtime (scheduler): add revised arm scheduler with support for atsamd21

jaddr2line:
This PR is based off of #1006 and #1002. As of now, the queue is not yet exposed to anything. I will add that in a bit.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -252,6 +252,13 @@ func Compile(pkgName string, machine llvm.TargetMachine, config *compileopts.Con
 
 	c.loadASTComments(lprogram)
 
+	// Forcibly preload special types.
+	runtimePkg := c.ir.Program.ImportedPackage("runtime")
+	c.getLLVMType(runtimePkg.Type("_interface").Type())
+	c.getLLVMType(runtimePkg.Type("_string").Type())
+	c.getLLVMType(runtimePkg.Type("hashmap").Type())
+	c.getLLVMType(runtimePkg.Type("channel").Type())
+
 	// Declare runtime types.
 	// TODO: lazily create runtime types in getLLVMRuntimeType when they are
 	// needed. Eventually this will be required anyway, when packages are
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -1476,7 +1476,7 @@ func (b *builder) createExpr(expr ssa.Value) (llvm.Value, error) {
 	case *ssa.BinOp:
 		x := b.getValue(expr.X)
 		y := b.getValue(expr.Y)
-		return b.createBinOp(expr.Op, expr.X.Type(), x, y, expr.Pos())
+		return b.createBinOp(expr.Op, expr.X.Type(), expr.Y.Type(), x, y, expr.Pos())
 	case *ssa.Call:
 		return b.createFunctionCall(expr.Common())
 	case *ssa.ChangeInterface:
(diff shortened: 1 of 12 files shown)


compiler/compiler.go
```diff
@@ -252,6 +252,13 @@ func Compile(pkgName string, machine llvm.TargetMachine, config *compileopts.Con
 
 	c.loadASTComments(lprogram)
 
+	// Forcibly preload special types.
+	runtimePkg := c.ir.Program.ImportedPackage("runtime")
+	c.getLLVMType(runtimePkg.Type("_interface").Type())
+	c.getLLVMType(runtimePkg.Type("_string").Type())
+	c.getLLVMType(runtimePkg.Type("hashmap").Type())
+	c.getLLVMType(runtimePkg.Type("channel").Type())
```
aykevl: Is this really necessary? I would expect these types to be included with the loop below.

jaddr2line: Unfortunately, some of these need to be pre-defined before they can be used by other runtime types.

src/runtime/timer.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	if baseUpdated {
+		// The base time has changed, so the timer interrupt needs to be reconfigured.
+		c.set()
+	}
+}
+
+// poll checks for any expired timers and schedules them.
+// This must be called periodically by the scheduler.
```
aykevl, line 92: This feels like a possible problem. Why is there a need for polling?

aykevl:
I looked at the code, but I'm not fully grasping what it is doing. I looked at #1002 (which is pretty clear), this PR, and #1010 but I miss the big picture. Reading through the code I see a lot of newly added code and some changed code but somehow I'm missing the purpose. Questions I get are:

* Is the legacy scheduler intended to be removed eventually? Because I suspect it will be fine for WebAssembly for example. If not, I don't think it should be called 'legacy'.
* A function like `ticks` has been moved, but I'm not sure why? It certainly distracts from the actual changes.
* I see a lot of references to polling. Is there really any polling going on? Why? Because generally you'd want to avoid polling and I can't think of a reason why you would need it.
* There is a whole new timer controller that I'm wondering about.

So can you maybe give a high level overview of what this particular PR is doing, why things are done the way they are, and what's up with all the polling? Also, how this will eventually result in things like interrupt-safe conditional variables?

src/runtime/timer.go
niaow: We have 2 kinds of platforms in TinyGo:
1. microcontrollers -  where all events come from interrupts
2. "hosted" platforms (wasm, embedded linux, etc.) - where we need to periodically poll for events

We *could* use 2 different scheduler implementations, one interrupt based and one polling, but that would make some stuff more complicated and result in 2 different sets of behavior.

aykevl:
I'm a bit late here. But still. Is there a reason why you went for this almost entirely lock-free approach? I do not think it's a problem to disable interrupts for a short time when working with the runqueue.

niaow:
> Is the legacy scheduler intended to be removed eventually? Because I suspect it will be fine for WebAssembly for example. If not, I don't think it should be called 'legacy'.

It is legacy. WebAssembly has somewhat different semantics - we never need to re-check the timer, and all of the time queueing can be done at the end.

> A function like ticks has been moved, but I'm not sure why? It certainly distracts from the actual changes.

Ticks was not moved. The code before it was deleted, and new code was put there, and git decided that was how it wanted the diff to appear.

> I see a lot of references to polling. Is there really any polling going on? Why? Because generally you'd want to avoid polling and I can't think of a reason why you would need it.

On hosted systems, there is no way around polling. On microcontrollers, we need to copy awoken tasks to the runqueue (although we may want to look into more frequent polling there, since it is cheap).

> There is a whole new timer controller that I'm wondering about.

I think the timer system still currently makes sense to leave separate, since we have to constantly reorder and reschedule things. The timer controller mechanism is designed to be a simplified version of working with interrupt timers, and is supposed to become the standard mechanism for implementing runtime timers.

niaow:
> Also, how this will eventually result in things like interrupt-safe conditional variables?

The interrupt-safe condition variables push awoken tasks onto the interrupt queue. This change feeds those tasks into the runqueue and lets them be resumed.

> I'm a bit late here. But still. Is there a reason why you went for this almost entirely lock-free approach? I do not think it's a problem to disable interrupts for a short time when working with the runqueue.

I guess we could do this. I was just worried that we could potentially be doing a lot of manipulations to the runqueue, and this would mean that a large total time would end up spent with interrupts off. If you want, I can redesign this a bit to instead disable interrupts.

src/runtime/timer.go
aykevl: What kinds of events need to be polled for on hosted platforms? Epoll?
I may need to look more into it to understand the problem better.

niaow:
Closing in favor of #1142

PR #1011, opened 2020-03-29
runtime/sync: add interrupt-safe condition variable

jaddr2line:
This PR uses #1010 and implements an interrupt-safe condition variable. A demo is provided that uses systicks.


niaow:
Closing in favor of #1142

PR #1013, opened 2020-03-31
machine/atsamd21,atsamd51: send zero length packet when USB transfer …

deadprogram:
This PR should correct the last issues for atsamd21 and atsamd51 USB CDC by sending a zero length packet when USB transfer is complete on endpoint 0.

This fix is 100% due to awesome research by @sago35 thank you very much!

### DIFF
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -1477,8 +1477,6 @@ func handleUSB(intr interrupt.Interrupt) {
 			// clear stall request
 			setEPINTENCLR(0, sam.USB_DEVICE_EPINTENCLR_STALL1)
 		}
-	} else {
-		sendZlp()
 	}
 
 	// Now the actual transfer handlers, ignore endpoint number 0 (setup)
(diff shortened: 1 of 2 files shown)


aykevl:
I can confirm that this PR works for me (Linux kernel version 5.4) on both the Arduino Nano 33 IoT and the Adafruit Circuit Playground Bluefruit.

Unrelated: I noticed that for the CPB it gives an error:

    error: failed to flash /tmp/tinygo446225229/main.uf2: rename /media/ayke/CPLAYBTBOOT/flash.uf2.tmp /media/ayke/CPLAYBTBOOT/flash.uf2: no such file or directory

It still works, but this (somewhat cosmetic) error should probably be fixed eventually.

sago35:
USB-CDC now works without adding sendZlp() in handleUSBIRQ().
But it's not yet known which changes are affecting them.
I've only fixed the code in SAMD51.
Tested with pyportal (samd51) + windows10 only.

The changes are outlined below.

* made to resemble the arduino source
 * Added usbPackMessage (probably unnecessary)
 * Fix handleStandardSetup().
 * Fix sendDescriptor()
* ClearBits() and then SetBits() from SetBits().
* Return Ack for Start of frame.
* Removed sendZlp() in handleUSBIRQ().

I'm going to organize it tomorrow so that it's only a necessary change.
If possible, I would like you to check the operation on Linux, etc.
@deadprogram

https://github.com/sago35/tinygo/commit/22b4a22c402aa224391faf2be3006f166725618a

sago35:
Works for me.

deadprogram:
This branch now appears good to merge.

PR #1014, opened 2020-04-01
main: switch to LLVM 10

aykevl:
This commit also adds a bit of version independence, in particular for external commands. It also adds the LLVM version to the `tinygo version` command, which might help while debugging.

~TODO: make TinyGo work with LLVM version 9 too, for Fedora (and perhaps other distributions). I'm making a PR now so that I can see whether the Windows build will work.~ done

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,14 +14,14 @@ commands:
       - run:
           name: "Install apt dependencies"
           command: |
-            echo 'deb https://apt.llvm.org/buster/ llvm-toolchain-buster<<parameters.llvm>> main' | sudo tee /etc/apt/sources.list.d/llvm.list
+            echo 'deb https://apt.llvm.org/buster/ llvm-toolchain-buster-<<parameters.llvm>> main' | sudo tee /etc/apt/sources.list.d/llvm.list
             wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
             sudo apt-get update
             sudo apt-get install \
-                llvm<<parameters.llvm>>-dev \
-                clang<<parameters.llvm>> \
-                libclang<<parameters.llvm>>-dev \
-                lld<<parameters.llvm>> \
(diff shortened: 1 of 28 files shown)


aykevl:
Looks like I'm hitting this problem: https://reviews.llvm.org/D66324#1680231

aykevl:
I'm afraid fixing this will require some redesigning of how certain tools (`cc1` mainly) are invoked.

aykevl:
This PR is now ready for review. I've intentionally kept the last commit separate so that it can be reverted if/when the lld bug is fixed upstream.

deadprogram:
I was able to install LLVM 10 using the apt repo, and compile several code examples successfully using this branch.

When I first tried under this branch, I did get

```
error: failed to link /tmp/tinygo339089044/main: none of these commands were found in your $PATH: ld.lld-10 ld.lld
```

I probably did not have the proper symlink setup for LLVM 9? Now that I have LLVM 10 installed I cannot backtest LLVM 9 very easily.

That said, from my POV I think we should merge this now.

aykevl:
> ```
> error: failed to link /tmp/tinygo339089044/main: none of these commands were found in your $PATH: ld.lld-10 ld.lld
> ```

Did you install `ld.lld-10`?

    sudo apt-get install llvm-10-dev libclang-10-dev lld-10 clang-10

deadprogram:
The error occurred before I had installed llvm10, but already had llvm9 installed.

aykevl:
Ah I understand now.

QuLogic:
Ah cool, did you run into any issues with ARM sections? IIRC, that's what I got before stopping to focus on Go 1.14.

I will try this out myself, but probably won't be able to get to it for a few days.

aykevl:
I haven't seen anything about ARM sections, no. I did find a different linker bug, though.

deadprogram:
Been testing this branch, and seems ready to merge. Last call for feedback!

deadprogram:
OK, let's merge!

Issue #1009, opened 2020-03-29
runtime (gc): memory corruption due to untracked allocations


deadprogram:
The latest `dev` branch has some issues when running the https://github.com/tinygo-org/drivers/tree/master/examples/wifinina/mqttsub example.

After some period of time, somewhere between 100 and 1500 messages sent/received, the program on the Arduino Nano33 IoT crashes, giving one of the following outputs:

```
panic: runtime error: out of memory
```

```
fatal error: HardFault with sp=0x200007b4 pc=0xf7f8fbbd
````

```
panic: runtime error: goroutine stack overflow
```

```
panic: runtime error: nil pointer dereference
```

Note that it is one of the above when it crashes, not all of them.

**Comment by aykevl** [comment]
This looks like a memory corruption. Does it also happen with `-scheduler=coroutines`?

**Comment by aykevl** [comment]
Also, it would help a lot if you could bisect this issue to the offending commit (assuming that the master branch works).

**Comment by deadprogram** [comment]
It runs without errors when I use

```
tinygo flash -target arduino-nano33 -scheduler=coroutines ./examples/wifinina/mqttsub/
```

I will bisect to try to find the commit that breaks things.

**Comment by deadprogram** [comment]
It appears this is the answer:

```
# first bad commit: [6a50f25a48385fc78397b6f0f1c83650130d8d16] refactor coroutine lowering and tasks
```

**Comment by aykevl** [comment]
Okay, great. That's at least something. @jaddr2line can you perhaps take a look?

**Comment by jaddr2line** [comment]
I suspect that this is due to a small change in stack usage causing an overflow. Can you try raising the stack size a bit? Those numbers sound like potential garbage collection times.

**Comment by jaddr2line** [comment]
Also the only relevant change I can think of from that commit is that the task data is no longer on top of the stack. It may have overflowed before, but is now corrupting something other than the task data.

**Comment by aykevl** [comment]
Here is the stack size you can try raising (for example to 2048):

https://github.com/tinygo-org/tinygo/blob/57320c09222560dd3656d3c4bb05a16c1849e9a5/src/internal/task/task_stack_cortexm.go#L7

**Comment by deadprogram** [comment]
Raising the stack size to 2048 had no effect.

**Comment by jaddr2line** [comment]
There appears to be an issue with GC tracking right now. This patch forces the `conservative` GC to reclaim memory earlier for debugging purposes:
```
[...]

PR #1016, opened 2020-04-01
track phi nodes in stack slots

jaddr2line:
As demonstrated in #1009, allocations are sometimes released too early because the stack slot is hit multiple times. This PR attempts to fix this by generating separate stack slots for PHI nodes.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -1025,6 +1025,18 @@ func (b *builder) createFunctionDefinition() {
 			phi.llvm.AddIncoming([]llvm.Value{llvmVal}, []llvm.BasicBlock{llvmBlock})
 		}
 	}
+
+	if b.NeedsStackObjects() {
+		// Track phi nodes.
+		for _, phi := range b.phis {
+			insertPoint := llvm.NextInstruction(phi.llvm)
+			for !insertPoint.IsAPHINode().IsNil() {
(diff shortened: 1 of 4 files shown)


jaddr2line:
Note: this does not yet actually emit the trackpointer calls.

jaddr2line:
This PR seems to have solved memory corruption in the case described here: https://github.com/tinygo-org/tinygo/issues/1009#issuecomment-606684785

However, it appears that reflect is still broken on wasm.

deadprogram:
I have tested this branch with the mqttsub example, and I am still getting

```
panic: runtime error: goroutine stack overflow
```

I also tried with both this branch alone, with #1017 alone, and also with this branch and #1017 rebased together. Same error.

jaddr2line:
@deadprogram This PR does not affect bare-metal targets. That is a different bug, which I will hopefully fix today.

aykevl:
Looks good to me. I checked the size impact and it was not as bad as I was afraid it would be.

jaddr2line:
I will eventually try to optimize this later, but right now I have to prioritize bugfixes.

aykevl:
Yes I understand. I just wanted to make sure the binary size doesn't blow up unexpectedly (and if it does, only when it is absolutely needed).

PR #1017, opened 2020-04-01
compiler: track the result of string concatenation

jaddr2line:
Before this commit, the garbage collector was able to collect string values while they were still in use.


aykevl:
Thank you! I never realized a binary operator could also result in a new local heap pointer (binops are normally for arithmetic).

PR #1018, opened 2020-04-02
runtime (gc): scan callee-saved registers while marking stack on baremetal

jaddr2line:
Before this PR, some allocations on baremetal would not be tracked by the garbage collector while in use. This is because the only location that they were stored in at the time that the GC was invoked was in a callee-saved register. This PR works around the problem by pushing all callee-saved registers onto the active stack before scanning it.

### DIFF
--- a/src/runtime/gc_stack_raw.go
+++ b/src/runtime/gc_stack_raw.go
@@ -3,12 +3,38 @@
 
 package runtime
 
+import "internal/task"
+
 // markStack marks all root pointers found on the stack.
 //
 // This implementation is conservative and relies on the stack top (provided by
 // the linker) and getting the current stack pointer from a register. Also, it
 // assumes a descending stack. Thus, it is not very portable.
 func markStack() {
-	// Mark system stack.
-	markRoots(getSystemStackPointer(), stackTop)
+	// Scan the current stack, and all current registers.
+	scanCurrentStack()
+
+	if !task.OnSystemStack() {
+		// Mark system stack.
+		markRoots(getSystemStackPointer(), stackTop)
+	}
+}
+
+//go:export tinygo_scanCurrentStack
+func scanCurrentStack()
+
+//go:export tinygo_scanstack
+func scanstack(sp uintptr) {
+	// Mark current stack.
+	// This function is called by scanCurrentStack, after pushing all registers onto the stack.
+	// Callee-saved registers have been pushed onto stack by tinygo_localscan, so this will scan them too.
+	if task.OnSystemStack() {
+		// This is the system stack.
+		// Scan all words on the stack.
+		markRoots(sp, stackTop)
+	} else {
+		// This is a goroutine stack.
+		// It is an allocation, so scan it as if it were a value in a global.
+		markRoot(0, sp)
+	}
 }
--- a/src/runtime/scheduler_avr.S
+++ b/src/runtime/scheduler_avr.S
@@ -187,3 +187,55 @@ tinygo_switchToScheduler:
 
     // Return into the scheduler, as if tinygo_switchToTask was a regular call.
     ret
+
+.global tinygo_scanCurrentStack
+.type tinygo_scanCurrentStack, %function
+tinygo_scanCurrentStack:
+    // Save callee-saved registers.
+    push r29 // Y
+    push r28 // Y
+    push r17
+    push r16
+    push r15
+    push r14
+    push r13
+    push r12
+    push r11
+    push r10
+    push r9
+    push r8
+    push r7
+    push r6
+    push r5
+    push r4
+    push r3
+    push r2
+
+    // Scan the stack.
+    in  r24, 0x3d; SPL
+    in  r25, 0x3e; SPH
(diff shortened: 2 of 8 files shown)


src/runtime/scheduler_avr.S
```diff
@@ -187,3 +187,55 @@ tinygo_switchToScheduler:
 
     // Return into the scheduler, as if tinygo_switchToTask was a regular call.
     ret
+
+.global tinygo_localscan
+.type tinygo_localscan, %function
+tinygo_localscan:
+    // Save callee-saved registers.
+    push r29 // Y
+	  push r28 // Y
```
aykevl, line 196: This file contains some inconsistent whitespace.

(While I personally prefer spaces over tabs, it is far more important to be consistent with the surrounding code).

src/runtime/gc_stack_raw.go
```diff
@@ -3,12 +3,38 @@
...
+	localscan()
+
+	if !task.OnSystemStack() {
+		// Mark system stack.
+		markRoots(getSystemStackPointer(), stackTop)
+	}
+}
+
+//go:export tinygo_localscan
+func localscan()
```
aykevl, line 24: I think `localscan` is not very descriptive. What about `scanstackwrapper`, `markStackWrapper`, `markRegistersAndStack`?

jaddr2line:
The comments have been addressed.

deadprogram:
When running

```
tinygo flash -target arduino-nano33 ./examples/wifinina/mqttsub/
```

I am still getting this error

```
panic: runtime error: goroutine stack overflow
```

jaddr2line:
Yes. This raises stack usage as a side-effect. That isnt memory corruption, it is a legitimate stack overflow. I think that stack usage should go back down when we make the gc non-recursive.

src/runtime/scheduler_avr.S
```diff
@@ -187,3 +187,55 @@ tinygo_switchToScheduler:
...
+    pop r10
+    pop r11
+    pop r12
+    pop r13
+    pop r14
+    pop r15
+    pop r16
+    pop r17
+    pop r28 // Y
+    pop r29 // Y
```
aykevl, line 241: This code can perhaps be replaced by the following:

```asm
in  r26, 0x3f  ; SREG
cli
in  r24, 0x3d  ; SPL
in  r25, 0x3e  ; SPH
adiw r24, 18
out  0x3d, r24 ; SPL
out  0x3e, r25 ; SPH
out  0x3f, r26 ; SREG
```

I'm not sure whether that's worth the trouble as it is fairly complex.

src/runtime/scheduler_cortexm.S
```diff
@@ -111,3 +111,35 @@ tinygo_swapTask:
...
+    pop {r4-r11, pc}
+    #else
+    pop {r4-r7}
+    pop {r0-r3}
+    mov r8, r0
+    mov r9, r1
+    mov r10, r2
+    mov r11, r3
+    pop {pc}
+    #endif
```
aykevl, line 145: You should be able to replace this with the following:

```
        add     sp, #32
        pop     {pc}
```

That's one instruction more on thumb2 but a whole lot faster in all cases.

src/runtime/scheduler_tinygoriscv.S
```diff
@@ -0,0 +1,29 @@
...
+   sw s6,  36(sp)
+   sw s5,  32(sp)
+   sw s4,  28(sp)
+   sw s3,  24(sp)
+   sw s2,  20(sp)
+   sw s1,  16(sp)
+   sw s0,  12(sp)
+
+   // Scan the stack.
+   addi a0, sp, 0
```
aykevl, line 22: You could perhaps use `mv a0, sp` here for better readability (it's an alias).

src/runtime/scheduler_avr.S
jaddr2line: I think I am just going to stick with the current approach here.

src/runtime/scheduler_cortexm.S
jaddr2line: Done.

src/runtime/scheduler_tinygoriscv.S
jaddr2line: I made the change, and confirmed that it builds. Do you have a board which you can test this on?

deadprogram: I will test this shortly.

deadprogram: Confirmed that the changes to this PR work as expected on my risc-v board.

deadprogram:
Tested on all platforms, and it is working as expected.

Now merging, great work on this @jaddr2line

src/runtime/scheduler_tinygoriscv.S
aykevl: For the future, you can run the test in an emulator, for example `tinygo run -target=riscv-qemu ./testdata/gc.go`. There is another (`-target=hifive1-qemu`) but I accidentally broke it (fixed in https://github.com/tinygo-org/tinygo/pull/1026).

PR #1019, opened 2020-04-02
builder: run tools (clang, ...) as separate processes

aykevl:
This is necessary because LLVM defines many options in global variables that are modified when invoking Clang. In particular, LLVM 10 seems to have a bug in which it always sets the `-pgo-warn-misexpect` flag. Setting it multiple times (over various cc1 invocations) results in an error:

    clang (LLVM option parsing): for the --pgo-warn-misexpect option: may only occur zero or one times!

This is fixed by running the Clang invocation in a new `tinygo` invocation.

Because we've had issues with lld in the past, also run lld in a separate process so similar issues won't happen with lld in the future.


deadprogram:
OK, merging.

PR #1020, opened 2020-04-03
main (test): remove global build lock in tests

jaddr2line:
Previously, due to bugs in LLD, we were forced to put a global lock around builds. Now that LLD and clang are being run in seperate processes, this is no longer necessary. This change allows tests to be run concurrently.


jaddr2line:
Huh. Nevermind.

aykevl:
Perhaps you'll need a lock around libclang?

PR #1021, opened 2020-04-04
machine: move errors.New calls to globals

aykevl:
Calling `errors.New` in an error path causes a heap allocation at an already unfortunate moment. It is more efficient to create these error values in globals and return these constant globals. If these errors are not used (because the related code was optimized out), the globals will also be optimized out.

Feel free to suggest more (or less) precise error messages, I haven't put much thought in which errors should be merged and which shouldn't.

---

This commit also helps avoid a possible recursive code path: `runtime.alloc` may print an error message when the heap is full but the USB-CDC code to print it may also cause an allocation to happen.

### DIFF
--- a/src/device/arm/arm.go
+++ b/src/device/arm/arm.go
@@ -35,6 +35,8 @@ import (
 	"unsafe"
 )
 
+var errCycleCountTooLarge = errors.New("requested cycle count is too large, overflows 24 bit counter")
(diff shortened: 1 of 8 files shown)


deadprogram:
In addition to the memory benefits, it also helps clarify the code itself.

Now merging.

PR #1022, opened 2020-04-04
runtime: avoid recursion in printuint64 function

aykevl:
This function is called from `runtime.printitf`, which is called from `runtime._panic`, and is therefore the leaf function of many call paths. This makes analyzing stack usage very difficult.

Also forwarding `printuint32` to `printuint64` as it reduces code size in the few examples I've tested. Printing numbers is not often done so it doesn't matter if it's a bit slow (the serial connection is probably slower anyway).

### DIFF
--- a/src/runtime/print.go
+++ b/src/runtime/print.go
@@ -47,23 +47,8 @@ func printint16(n int16) {
 	printint32(int32(n))
 }
 
-//go:nobounds
 func printuint32(n uint32) {
-	digits := [10]byte{} // enough to hold (2^32)-1
-	// Fill in all 10 digits.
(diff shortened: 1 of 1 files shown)


jaddr2line:
Is there a reason we are making this one non-recursive, and not the others (e.g. `printuint8`)?

aykevl:
No particular reason, except that they didn't show up in my stack depth analysis (because they are only recursive on 8-bit systems). I should probably make those non-recursive too.

deadprogram:
Is this PR ready to merge? Looks good to me.

aykevl:
Yes it is ready. The code can be improved further (as @jaddr2line said), but that code is only recursive on AVR.

PR #1024, opened 2020-04-04
machine/atsamd21,atsamd51: switch to modern interrupt registration method for UART

aykevl:
This saves about 112 bytes in flash and 288 bytes in RAM when the UART is not used.


deadprogram:
Glad you did this @aykevl now merging.

PR #1027, opened 2020-04-06
WIP: for feedback from tinygo folks

iansmith:
@aykevl @jaddr2line 

This PR is to confirm that I have the files setup "right" to be a good citizen in tinygo.  

The test program (note this does UART synchronously, not yet to the point of interrupts):
```
package main

import "machine"

var testMyBSS int

func main() {
	// set up serial console
	machine.UARTInit()

	// say hello
	machine.UARTPuts("Hello World!\n")

	// echo everything back
	for {
		machine.UARTPutc(machine.UARTGetc())
	}
}
```

If you want to test locally:
```
$ tinygo build -target rpi3 -o uart main.go
$ llvm-objcopy -O binary uart kernel8.img
```
Assuming you have Qemu 4.2 you can run with:
```
$ qemu-system-aarch64 -M raspi3 -kernel kernel8.img -serial null -serial stdio
```


src/machine/board_rpi3.go
```diff
@@ -0,0 +1,182 @@
...
+var MemoryMappedIO = uintptr(0x3F000000)
+
+
+// XXX initialized in preinit so we avoid interp.Unsupported
+var Aux *AuxPeripheralsRegisterMap //= (*AuxPeripheralsRegisterMap)(unsafe.Pointer(MemoryMappedIO + 0x00215000))
+var GPIO *GPIORegisterMap //= (*GPIORegisterMap)(unsafe.Pointer(MemoryMappedIO + 0x00200000))
+var SysTimer *SysTimerRegisterMap //= (*SysTimerRegisterMap)(unsafe.Pointer(MemoryMappedIO + 0x3000))
+
+type AuxPeripheralsRegisterMap struct {
+	InterruptStatus           uint32 //0x00
```
niaow: You can use volatile registers in order to ensure that semantics are preserved correctly. You can look at the device files in `device/sam` or similar for inspiration (after you have run `make gen-device`).

src/machine/board_rpi3.go
```diff
@@ -0,0 +1,182 @@
+// +build rpi3
+
+package machine
+
+import (
+	"device/arm"
+)
+
+//specific to raspberry pi 3
+var MemoryMappedIO = uintptr(0x3F000000)
```
niaow: This should be changed to be a constant to simplify things.

src/machine/board_rpi3.go
```diff
@@ -0,0 +1,182 @@
...
+
+
+// XXX initialized in preinit so we avoid interp.Unsupported
+var Aux *AuxPeripheralsRegisterMap //= (*AuxPeripheralsRegisterMap)(unsafe.Pointer(MemoryMappedIO + 0x00215000))
+var GPIO *GPIORegisterMap //= (*GPIORegisterMap)(unsafe.Pointer(MemoryMappedIO + 0x00200000))
+var SysTimer *SysTimerRegisterMap //= (*SysTimerRegisterMap)(unsafe.Pointer(MemoryMappedIO + 0x3000))
```
niaow: I think that using volatile registers and making `MemoryMappedIO` a constant should fix the issue and let you avoid the preinit.

src/runtime/runtime_rpi3.go
```diff
@@ -0,0 +1,140 @@
...
+	//}
+
+	machine.UARTInit()
+}
+
+// main on bare metal (bootloader uses this path)
+//go:export main
+func main() {
+	preinit()
+	primaryMain()
```
niaow: If you rebase on `dev`, there is a convenient wrapper called `run()` which you can use to replace this. Then your code could be changed to:
```
preinit()
run()
machine.Abort()
```

If the scheduler does not work yet, you can set the scheduler to `none` in the device file, like we currently do for avr: https://github.com/tinygo-org/tinygo/blob/cbaa58a2d9f4c29a2f0fbd79130fee62b8a1f72a/targets/avr.json#L8

src/runtime/runtime_unix.go
```diff
@@ -1,4 +1,4 @@
-// +build darwin linux,!baremetal freebsd,!baremetal
+// +build (darwin linux,!baremetal freebsd,!baremetal),!rpi3
```
niaow: The board is marked as `baremetal`, which will make this unnecessary.

src/runtime/runtime_rpi3.go
```diff
@@ -0,0 +1,140 @@
...
+//go:extern
+func putchar(c byte) {
+	//MiniUARTSend(c) if you prefer the mini uart
+	machine.UARTPutc(c)
+}
+
+//
+// Wait MuSec waits for at least n musecs based on the system timer.
+//
+func WaitMuSec(n uint32) {
```
niaow: Does this need to be exported?

src/runtime/runtime_rpi3.go
```diff
@@ -0,0 +1,140 @@
...
+	for r < t {
+		arm.AsmFull(`mrs x27, cntpct_el0
+			str x27,{r}`, map[string]interface{}{"r": &r})
+ 	}
+}
+
+//
+// SysTimer gets the 64 bit timer's value.
+//
+func SysTimer() uint64 {
```
niaow: Does this need to be exported?

If this needs to be exported, maybe put it in `machine`.

src/runtime/runtime_rpi3.go
```diff
@@ -0,0 +1,140 @@
...
+	"device/arm"
+)
+
+const tickMicros = int64(1)
+
+type timeUnit int64
+
+//go:export sleepticks sleepticks
+func sleepTicks(n timeUnit) {
+	WaitMuSec(uint32(n))
```
niaow: You can handle longer durations by checking for overflow.
```suggestion
	for n > timeUnit(^uint32(0)) {
		WaitMuSec(^uint32(0))
		n -= timeUnit(^uint32(0))
	}
	WaitMuSec(uint32(n))
```

src/machine/board_rpi3.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func UARTPuts(s string) {
+	for i := 0; i < len(s); i++ {
+		UARTPutc(s[i])
+	}
+}
+
+func Abort() {
+	arm.Asm("nop")
```
niaow: How does this function as an abort? Wouldn't you need a loop around this?

Makefile
```diff
@@ -137,7 +137,7 @@ llvm-source: llvm-project/README.md
 # Configure LLVM.
 TINYGO_SOURCE_DIR=$(shell pwd)
 $(LLVM_BUILDDIR)/build.ninja: llvm-source
-	mkdir -p $(LLVM_BUILDDIR); cd $(LLVM_BUILDDIR); cmake -G Ninja $(TINYGO_SOURCE_DIR)/llvm-project/llvm "-DLLVM_TARGETS_TO_BUILD=X86;ARM;AArch64;RISCV;WebAssembly" "-DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=AVR" -DCMAKE_BUILD_TYPE=Release -DLIBCLANG_BUILD_STATIC=ON -DLLVM_ENABLE_TERMINFO=OFF -DLLVM_ENABLE_ZLIB=OFF -DLLVM_ENABLE_PROJECTS="clang;lld" -DLLVM_TOOL_CLANG_TOOLS_EXTRA_BUILD=OFF $(LLVM_OPTION)
+	mkdir -p $(LLVM_BUILDDIR); cd $(LLVM_BUILDDIR); cmake -G Ninja $(TINYGO_SOURCE_DIR)/llvm-project/llvm "-DLLVM_TARGETS_TO_BUILD=X86;ARM;AArch64;RISCV;WebAssembly" "-DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=AVR" -DCMAKE_BUILD_TYPE=Release -DLIBCLANG_BUILD_STATIC=ON -DLLVM_ENABLE_TERMINFO=OFF -DLLVM_ENABLE_ZLIB=OFF -DLLVM_ENABLE_PROJECTS="clang;lld;compiler-rt" -DLLVM_TOOL_CLANG_TOOLS_EXTRA_BUILD=OFF $(LLVM_OPTION)
```
aykevl: We include compiler-rt in a different way so this change is unnecessary.

src/machine/board_rpi3.go
```diff
@@ -0,0 +1,182 @@
...
+	ControlStatus uint32 //0x00
+	CounterLower32 uint32 //0x04
+	CounterHigher32 uint32 //0x08
+	reservedGPU0 uint32 //0x0C
+	Compare1 uint32 //0x10
+	reservedGPU2 uint32 //0x14
+	Compare3 uint32 //0x18
+}
+
+func UARTInit() {
```
aykevl: This UART should have an interface consistent with other targets. Something like:

```go
type UART struct {
    // You can leave this empty.
    // If there are multiple UARTs with a similar memory map (at different addresses)
    // you can include a pointer to the block of registers here.
}

func (uart UART) Configure(config UARTConfig) error {
    // this has the contents of UARTInit
}

func (uart UART) WriteByte(c byte) error {
    // this has the contents of UARTPutc
}
```

src/runtime/runtime_rpi3.go
```diff
@@ -0,0 +1,140 @@
...
+	machine.SysTimer = (*machine.SysTimerRegisterMap)(unsafe.Pointer(machine.MemoryMappedIO + 0x3000))
+
+
+
+	// Initialize .bss: zero-initialized global variables.
+	//ptr := unsafe.Pointer(&_sbss)
+	//for ptr != unsafe.Pointer(&_ebss) {
+	//	*(*uint32)(ptr) = 0
+	//	ptr = unsafe.Pointer(uintptr(ptr) + 4)
+	//}
```
aykevl: This looks necessary to me, why is it commented out?

src/runtime/runtime_rpi3.go
```diff
@@ -0,0 +1,140 @@
...
+	primaryMain()
+}
+
+func primaryMain() {
+	initAll()
+	callMain()
+	machine.Abort()
+}
+
+//go:extern
```
aykevl: Unless you want to link with an external library, you don't need to mark this as `//go:extern`.

targets/arm64.json
```diff
@@ -0,0 +1,24 @@
...
+	"goos":"linux",
+	"llvm-target":"aarch64-elf",
+	"tags":"arm64",
+	"compiler":"clang",
+	"linker":"ld.lld",
+	"rtlib": "compiler-rt",
+	"cflags": [
+		"--target=aarch64-elf",
+		"-fshort-enums",
+		"-nostdlibinc",
```
aykevl: Instead of `-nostdlibinc`, you can add `"libc": "picolibc"` (on the dev branch).

aykevl:
Note: I have changed the PR from master to dev, as that's what we're working on (and dev has many changes now not included in master).

src/machine/board_rpi3.go
iansmith: Done. Much cleaner.

src/machine/board_rpi3.go
iansmith: Done.

src/machine/board_rpi3.go
iansmith: Yep, it did.

src/runtime/runtime_rpi3.go
iansmith: Yep, done.

src/runtime/runtime_rpi3.go
iansmith: Done.

src/runtime/runtime_rpi3.go
iansmith: Done.

src/machine/board_rpi3.go
iansmith: Changed to wfe (wait for event) which means "go to sleep and wait for something" ... that will never happen, in this case.

src/runtime/runtime_unix.go
iansmith: Yep, fixed.

src/runtime/runtime_rpi3.go
iansmith: Allowed 64bit wait values.

Makefile
iansmith: Yep, fixed.

src/runtime/runtime_rpi3.go
iansmith: Because I could not figure out how to get it to work. I was getting some undefined symbols so I need to investigate.  After this PR gets merged, I will fix.

src/machine/board_rpi3.go
aykevl: I would still recommend a loop around the `wfe` here. Even though it _should_ not continue after the abort, it may happen due to unexpected circumstances. A loop will only cost one extra instruction.

Also, all other targets put `abort` in the runtime and keep it unexported. There is no reason a program should call it directly: if programs need a way to exit then `os.Exit` should be implemented instead (which still doesn't make a lot of sense on an embedded system).

src/runtime/runtime_rpi3.go
aykevl: This really needs to be fixed, otherwise many programs won't work correctly (globals will not be zeroed if necessary). What is the error you're getting?

iansmith: I put this back in the code, although at the moment I don't have any code that sets ebss!=sbss...

targets/arm64.json
iansmith: Fixed.  Required a change to the picolib library builder to get it to work.   The fix is not included in this PR.

```
$ git diff builder/picolibc.go
diff --git a/builder/picolibc.go b/builder/picolibc.go
index 31e2f05..ede39c3 100644
--- a/builder/picolibc.go
+++ b/builder/picolibc.go
@@ -12,7 +12,7 @@ var Picolibc = Library{
        name: "picolibc",
        cflags: func() []string {
                picolibcDir := filepath.Join(goenv.Get("TINYGOROOT"), "lib/picolibc/newlib/libc")
-               return []string{"-Werror", "-Wall", "-std=gnu11", "-D_COMPILING_NEWLIB", "-fshort-enums", "--sysroot=" + picolibcDir, "-I" + picolibcDir + "/tinystdio", "-I" + goenv.Get("TINYGOROOT") + "/lib/picolibc-include"}
+               return []string{"-Werror", "-Wall", "-std=gnu11", "-D_COMPILING_NEWLIB","-I"+picolibcDir+"/include", "-fshort-enums", "--sysroot=" + picolibcDir, "-I" + picolibcDir + "/tinystdio", "-I" + goenv.Get("TINYGOROOT") + "/lib/picolibc-include"}
        },
        sourceDir: "lib/picolibc/newlib/libc",
        sources: func(target string) []string {
```

src/machine/board_rpi3.go
iansmith: Updated to reflect your UART defn.  However, right now it's pretty dumb because I'm using the (simplest) UART of the many on an RPI.  When I get to doing UART1 and so forth then it will be time to do it properly.

src/runtime/runtime_rpi3.go
iansmith: Removed.

src/runtime/runtime_rpi3.go
aykevl: What do you mean? `_ebss` and `_sbss` are set by the linker, so what comes between is what the compiler/linker put there (for example, some globals used by the heap that are zero-initialized).

Makefile
```diff
@@ -84,6 +84,7 @@ $(LIBCLANG_PATH): $(LIBCLANG_FILES)
 else ifeq ($(shell uname -s),Darwin)
     MD5SUM = md5
     LIBCLANG_PATH = $(abspath $(LLVM_BUILDDIR))/lib/libclang.a
+    CGO_CXXFLAGS=-I/Library/Developer/CommandLineTools/usr/include/c++/v1
```
aykevl: This is a platform-specific change that really doesn't belong in this PR. And honestly I feel like this patches over some existing issue on your system: it seems to work fine for most other people.

Makefile
```diff
@@ -103,7 +104,7 @@ LLD_LIBS = $(START_GROUP) -llldCOFF -llldCommon -llldCore -llldDriver -llldELF -
 # For static linking.
 ifneq ("$(wildcard $(LLVM_BUILDDIR)/bin/llvm-config*)","")
     CGO_CPPFLAGS=$(shell $(LLVM_BUILDDIR)/bin/llvm-config --cppflags) -I$(abspath $(LLVM_BUILDDIR))/tools/clang/include -I$(abspath $(CLANG_SRC))/include -I$(abspath $(LLD_SRC))/include
-    CGO_CXXFLAGS=-std=c++11
+    CGO_CXXFLAGS+=-std=c++11
```
aykevl: (related to the above comment)

Makefile
```diff
@@ -154,7 +155,7 @@ llvm-source: $(LLVM_PROJECTDIR)/README.md
 # Configure LLVM.
 TINYGO_SOURCE_DIR=$(shell pwd)
 $(LLVM_BUILDDIR)/build.ninja: llvm-source
-	mkdir -p $(LLVM_BUILDDIR); cd $(LLVM_BUILDDIR); cmake -G Ninja $(TINYGO_SOURCE_DIR)/$(LLVM_PROJECTDIR)/llvm "-DLLVM_TARGETS_TO_BUILD=X86;ARM;AArch64;RISCV;WebAssembly" "-DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=AVR" -DCMAKE_BUILD_TYPE=Release -DLIBCLANG_BUILD_STATIC=ON -DLLVM_ENABLE_TERMINFO=OFF -DLLVM_ENABLE_ZLIB=OFF -DLLVM_ENABLE_PROJECTS="clang;lld" -DLLVM_TOOL_CLANG_TOOLS_EXTRA_BUILD=OFF $(LLVM_OPTION)
+	mkdir -p $(LLVM_BUILDDIR); cd $(LLVM_BUILDDIR); cmake -G Ninja $(TINYGO_SOURCE_DIR)/$(LLVM_PROJECTDIR)/llvm "-DLLVM_TARGETS_TO_BUILD=X86;ARM;AArch64;RISCV;WebAssembly" "-DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=AVR" -DCMAKE_BUILD_TYPE=Release -DLIBCLANG_BUILD_STATIC=ON -DLLVM_ENABLE_TERMINFO=OFF -DLLVM_ENABLE_ZLIB=OFF -DLLVM_ENABLE_PROJECTS="clang;clang-tools-extra" -DLLVM_TOOL_CLANG_TOOLS_EXTRA_BUILD=ON $(LLVM_OPTION)
```
aykevl: There is still a diff here. I do not believe we need `clang-tools-extra` for anything in TinyGo.

builder/library.go
```diff
@@ -67,7 +67,7 @@ func (l *Library) Load(target string) (path string, err error) {
 	defer os.RemoveAll(dir)
 
 	// Precalculate the flags to the compiler invocation.
-	args := append(l.cflags(), "-c", "-Oz", "-g", "-ffunction-sections", "-fdata-sections", "-Wno-macro-redefined", "--target="+target, "-fdebug-prefix-map="+dir+"="+remapDir)
+	args := append(l.cflags(), "-v","-c", "-Oz", "-g", "-ffunction-sections", "-fdata-sections", "-Wno-macro-redefined", "--target="+target, "-fdebug-prefix-map="+dir+"="+remapDir)
```
aykevl: This appears to be a leftover from some debugging, and should be removed.

targets/arm64.json
```diff
@@ -0,0 +1,24 @@
+{
+	"build-tags": ["arm64"],
+	"goarch":"arm64",
+	"goos":"linux",
+	"llvm-target":"aarch64-elf",
+	"tags":"arm64",
+	"compiler":"clang",
+	"linker":"ld.lld",
+	"rtlib": "compiler-rt",
+	"libc": "picolibc",
```
aykevl, line 10: I think it's best to avoid picolibc for now. I would like to first see why Clang doesn't use the usual `<sysroot>/include` but instead uses `<sysroot>/usr/include`, it doesn't make sense that it only does this for aarch64 (and not for arm, riscv, and avr) and I'm hopeful there is some sort of flag to tell it not to do this. In any case, those changes can be done in a separate PR (a libc should not be necessary for initial bringup of this target).

src/runtime/runtime_rpi3.go
```diff
@@ -0,0 +1,89 @@
...
+
+// abort is called by panic().
+func abort() {
+	machine.Abort()
+}
+
+
+// called once the memory allocator is ready and heap initialized
+func postinit() {
+	// Initialize .bss: zero-initialized global variables.
```
aykevl: This doesn't make sense. You can't initialize the heap without first clearing the `.bss` section. You need to clear the BSS section (and initialize the `.data` section) before running any normal code.

src/device/arm/arm-cortex-a53.S
```diff
@@ -0,0 +1,31 @@
...
+    ldr     x1, =_start
+    mov     sp, x1
+
+    // clear bss
+    //ldr     x1, =__bss_start
+    //ldr     w2, =__bss_size
+3:  //cbz     w2, 4f
+    //str     xzr, [x1], #8
+    //sub     w2, w2, #1
+    //cbnz    w2, 3b
```
aykevl: Why did you comment this out? I think it is necessary. If it's because of problems with the linker, you need to rename `__bss_start` to `_sbss` and `__bss_end` to `_ebss`, because those are the names that are used in the linker script.
You can also do this in Go code and in fact there is a function `preinit` which does that, but it doesn't seem to be called from anywhere. Either assembly or Go is fine, although I generally tried to keep it in Go to make the code easier to read.

targets/rpi3.ld
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+SECTIONS
+{
+    . = 0x80000;
+    .text : { KEEP(*(.text.boot)) *(.text .text.* .gnu.linkonce.t*) }
+    .rodata : { *(.rodata .rodata.* .gnu.linkonce.r*) }
+    PROVIDE(_data = .);
+
+		/* Globals with initial value */
```
aykevl: This file uses inconsistent whitespace. I don't care much which variant you pick (tabs or spaces), but please keep it consistent.

src/runtime/runtime_rpi3.go
```diff
@@ -0,0 +1,89 @@
...
+// There are constants defined for most some interrupts in
+// src/device/rpi3.go.
+//func EnableInterruptController(bits uint32) {
+//	volatile.StoreUint32((*uint32)(dev.ENABLE_IRQS_1), bits)
+//}
+
+// EnableIRQ turns on the I in daif reg
+//func EnableIRQ() {
+//	dev.Asm("msr daifclr, #2")
+//}
```
aykevl: Such dead code shouldn't be kept around, to keep the diff clean. Feel free to put it in a separate branch for when you get interrupts to work, but don't leave around unused code.

aykevl:
There are a number of conflicts, which you can fix by rebasing on top of the dev branch (although it appears like you already did some of that in #1048?)

Makefile
iansmith: Yeah, you are right. And these problems seem to have been fixed with LLVM10 (it now seems to know about the weird placement of /usr/include on a mac)

Makefile
iansmith: Yeah, you are right.

builder/library.go
iansmith: Done.

targets/arm64.json
iansmith: Removed.

src/device/arm/arm-cortex-a53.S
iansmith: Fixed. Eliminated dead code in asm, then moved to using the go code that is already in use by other runtimes (risc v, avr, etc).

targets/rpi3.ld
iansmith: Yeah, my editor (goland) didn't know about that type of file so things were not consistent.  I used the goland "convert spaces to tabs" to fix it.

src/runtime/runtime_rpi3.go
iansmith: Yep, this code is in the upcoming PR that has interrupts working for the UART.  I deleted for this PR, however.

src/runtime/runtime_rpi3.go
iansmith: Fixed, see other comment.

Makefile
iansmith: Fixed. And, no, it was not needed.

iansmith:
@aykevl It appears that the problem with picolib is affecting circle-ci as well. Check the errors in this build.  This is the same trouble I had on my mac.

https://app.circleci.com/pipelines/github/iansmith/tinygo/10/workflows/2e5badf7-1a2e-4297-a333-62dc856cd707/jobs/61

src/runtime/runtime_rpi3.go
aykevl: I still don't see where .bss is initialized? You have a `postinit` function which appears to do that, but it isn't called anywhere.
My proposal would be:

* Rename `postinit` to `preinit`.
* Call `preinit()` just before `run()` call.
* Add a new empty `postinit` function. It is only really used to enable interrupts on AVR but because how things are designed it has to be defined in every target.

Also I don't see where `.data` is initialized. It might not be necessary on the Raspberry Pi 3, though.

iansmith:
@aykevl @jaddr2line 

I implemented all the suggestions from Ayke on overriding and machine generation of the hardware MMIO registers.  

Close the other PRs that were confusing the issue.

deadprogram:
Closing due to age of this PR. Please open a new issue if anything here is still needed. Thanks!

PR #1028, opened 2020-04-06
runtime (gc): remove recursion from "conservative" GC

### LINKED ISSUE #436: runtime: Garbage collector should not recurse
At least within the context of WebAssembly, the conservative GC recursing is causing stack overflows on larger heaps. Ideally it should be possible to make it not recurse.

jaddr2line:
This PR attempts to fix #436 and hopefully reduce stack usage.

The GC modifications take a similar approach to the MicroPython GC, in that a fixed-size stack is used to track allocations that need to be scanned. If the stack overflows, this forces a full rescan of all marked allocations. This is potentially a lot slower in some extreme cases, but does not blow up the stack like the current version does.

This seems to pass some basic sanity checks, but we should try this out with some larger tinygo programs before merging it.

### DIFF
--- a/src/runtime/gc_conservative.go
+++ b/src/runtime/gc_conservative.go
@@ -45,6 +45,7 @@ const (
 	bytesPerBlock      = wordsPerBlock * unsafe.Sizeof(heapStart)
 	stateBits          = 2 // how many bits a block state takes (see blockState type)
 	blocksPerStateByte = 8 / stateBits
+	markStackSize      = 4 * unsafe.Sizeof((*int)(nil)) // number of to-be-marked blocks to queue before forcing a rescan
 )
 
 var (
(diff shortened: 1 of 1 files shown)


deadprogram:
Any feedback on this PR @aykevl ?

It seems needed in order to address #1009

jaddr2line:
Alright, I split the marking into 2 phases as requested.

deadprogram:
Let me know when it is ready to retest, please.

jaddr2line:
I have made the requested changes. @aykevl is there anything else before he retests?

src/runtime/gc_conservative.go
```diff
@@ -45,6 +45,7 @@ const (
 	bytesPerBlock      = wordsPerBlock * unsafe.Sizeof(heapStart)
 	stateBits          = 2 // how many bits a block state takes (see blockState type)
 	blocksPerStateByte = 8 / stateBits
+	markStackSize      = 4 * unsafe.Sizeof((*int)(nil)) // number of to-be-marked blocks to queue before forcing a rescan
```
aykevl, line 48: This number seems arbitrary but I guess it's reasonable.

jaddr2line: I basically had decided that it would be reasonable to spend 64 bytes on ARM, and then decided to reduce that to 16 on AVR. I basically just went with `4*(wordWidth^2)` bytes spent on the stack, since 256 bytes also seemed fine for AMD64 where there is plenty of memory.

deadprogram:
Retested and working as expected.

Now merging, great work @jaddr2line on this important set of fixes!

PR #1035, opened 2020-04-07
Go 1.14 support

aykevl:
This is a continuation of #901. I have modified the PR of @QuLogic to include the bits needed to get all tests to run on Go 1.14 and extended it a bit to switch to Go 1.14 entirely.

Note that this PR drops support for WebAssembly on Go 1.13 (at least when you import the syscall/js package) but I think working around that would be difficult. It should definitely be possible to create a workaround (for example, by exporting a function that when called returns the Go minor version) but I'd like to avoid such hacks if possible. Therefore I would propose keeping it as-is for now and only adding a workaround if there is enough demand (which I don't expect).

This will probably conflict with #1014 but the conflicts should be easy to resolve.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -232,8 +232,8 @@ commands:
       - run:
           name: "Install dependencies"
           command: |
-            curl https://dl.google.com/go/go1.13.darwin-amd64.tar.gz -o go1.13.darwin-amd64.tar.gz
-            sudo tar -C /usr/local -xzf go1.13.darwin-amd64.tar.gz
+            curl https://dl.google.com/go/go1.14.darwin-amd64.tar.gz -o go1.14.darwin-amd64.tar.gz
+            sudo tar -C /usr/local -xzf go1.14.darwin-amd64.tar.gz
             ln -s /usr/local/go/bin/go /usr/local/bin/go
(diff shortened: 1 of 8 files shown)


deadprogram:
So this PR would replace #901 basically, by incorporating part of it plus some other code?

Then also what about #994

aykevl:
> So this PR would replace #901 basically, by incorporating part of it plus some other code?

Yes, exactly.

> Then also what about #994

It is not necessary for Go 1.14 support. It should be merged evenutally, however.

deadprogram:
@aykevl this branch is now ready for conflict resolution, please.

aykevl:
Rebased and fixed conflicts.

deadprogram:
Tested with this branch and Go 1.14.2 with some of my demos on AVR, SAMD21, SAMD51, and Hi5 boards, and all worked perfectly. I think this is ready for merge.

deadprogram:
Now merging, thanks to @QuLogic and @aykevl for getting this done!

anatol:
Great news folks! Now when `dev` branch supports Golang 1.14 and LLVM10 is there a chance to get a new release? It will help users at systems like Arch Linux that have the new versions of dependencies already.

aykevl:
Yes that will happen very soon now.

PR #1036, opened 2020-04-07
internal/bytealg: reimplement bytealg in pure Go

jaddr2line:
Previously, we implemented individual bytealg functions via linknaming, and had to update them every once in a while when we hit linker errors. Instead, this change reimplements the bytealg package in pure Go. If something is missing, it will cause a compiler error rather than a linker error. This is easier to test and maintain.

NOTE: The next Go release (1.15) is set to include Rabin-Karp string searches, which will require adding some code to this.


aykevl:
It looks like you copied code from the main Go implementation. Is that true? If so, this should be indicated at the top of the file.

jaddr2line:
I do not believe that I did. Should we try copying the main Go implementation and modifying it?

aykevl:
No, I was just asking to be sure (for copyright reasons).

aykevl:
Thank you, this is a much-needed cleanup of the ad hoc string/bytes algorithms.
I hope a pure Go implementation is fast enough. If it is not, we should first consider optimizing the compiler instead of writing assembly implementations (as that will benefit all code).

PR #1039, opened 2020-04-08
runtime (gc): implement a best-fit allocator

jaddr2line:
This is an experimental change to the memory allocator which attempts to reduce fragmentation. It uses a doubly-linked list sorted in ascending order to track free regions of the heap. Allocations are fulfilled with the smallest free region exceeding the required size. This causes small holes in the heap to be filled first, and allows large empty regions to form.


niaow:
Replaced by #1181

PR #1041, opened 2020-04-09
compiler: unexport some exported symbols

aykevl:
Some symbols (constants/types/methods) were exported while they are an implementation detail. To keep the public API clean, unexport them.

The public API of the compiler package now consists of just the `Compile` and the `NewTargetMachine` functions.


deadprogram:
Always good to remove superfluous API surface. Merging.

PR #1042, opened 2020-04-09
compiler: add parameter names to IR

aykevl:
This makes viewing the IR easier because parameters have readable names.

This also makes it easier to write compiler tests (still a work in progress), that work in LLVM 9 and LLVM 10, as LLVM 10 started printing value names for unnamed parameters. Naming them would avoid that issue.

Example function:

```go
func printstring(s string)
```

Before:

```llvm
define internal void @runtime.printstring(i8* %0, i64 %1, i8* %context, i8* %parentHandle) unnamed_addr {
```

After:

```llvm
define internal void @runtime.printstring(i8* %s.data, i64 %s.len, i8* %context, i8* %parentHandle) unnamed_addr {
```

### DIFF
--- a/compiler/calls.go
+++ b/compiler/calls.go
@@ -2,6 +2,7 @@ package compiler
 
 import (
 	"go/types"
+	"strconv"
 
 	"tinygo.org/x/go-llvm"
 )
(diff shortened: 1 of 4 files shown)


deadprogram:
This PR now has a merge conflict that needs to be resolved, please.

aykevl:
Rebased.

deadprogram:
@jaddr2line any feedback on this PR?

deadprogram:
OK, merging. Thanks!

Issue #1032, opened 2020-04-06
transform (coroutines): undef task handle in call to closure


asdine:
```go
func foo() {
	ch := make(chan int)

	fn := func(nb int) {
		ch <- nb
	}

	go func() {
		fn(10)
	}()

	fmt.Println(<-ch)
}

func main() {
  foo()
}
```

This causes the following compilation error:

```
async function exp.foo$1 used as function pointer
```

The same happens when I replace the fn closure by a method.

I compiled to wasm using the `0.12.0` docker image with the following params:

```bash
docker run ... \
    tinygo/tinygo:0.12.0 tinygo build -o /dist/exp.wasm -target wasm --no-debug exp
```

jaddr2line:
On the current development branch, that part seems to have been fixed an replaced by a different problem: it now segfaults.

jaddr2line:
```
tail call void @"/tmp/indirectasync.go.foo$1"(i64 10, i8* nonnull %ch, i8* undef)
```

It appears that we are passing in an undefined task handle, which is causing a segfault when we attempt to update the state.

jaddr2line:
v0.13 is getting close to release, so the fix is probably going to have to be pushed back into the next release window.

jaddr2line:
Alright, I found the cause of the problem: https://github.com/tinygo-org/tinygo/blob/5308e8903e8c1a79151c8262fc2992eb405ce049/transform/func-lowering.go#L147

This causes the pass to leave behind an indirect function call, which causes the coroutine lowering pass to fail to detect the call. As a result, the call is not lowered, and the parent handle is not provided.

PR #1043, opened 2020-04-10
transform (func-lowering): remove specializations from function value lowering and fix lowering of a function value of an unimplemented type

jaddr2line:
Previously, the function value lowering pass had special cases for when there were 0 or 1 function implementations. However, the results of the pass were incorrect in both of these cases. This change removes the specializations and fixes the transformation.

In the case that there was a single function implementation, the compiler emitted a select instruction to obtain the function pointer. This selected between null and the implementing function pointer.
While this was technically correct, it failed to eliminate indirect function calls. This prevented discovery of these calls by the coroutine lowering pass, and caused async function calls to be passed through unlowered. As a result, the generated code had undefined behavior (usually resulting in a segfault).

In the case of no function implementations, the lowering code was correct. However, the lowering code was not run. The discovery of function signatures was accomplished by scanning implementations, and when there were no implementations nothing was discovered or lowered.

For maintainability reasons, I have removed both specializations rather than fixing them. This substantially simplifies the code, and reduces the amount of variation that we need to worry about for testing purposes. The IR now generated in the cases of 0 or 1 function implementations can be efficiently simplified by LLVM's optimization passes. Therefore, there should not be a substantial regression in terms of performance or machine code size.

I have also inspected some final output IR, and it seems that LLVM optimizes it to something similar to what we had wanted for a single function implementation:
```
  %switch = icmp eq i64 %.unpack2, 0
  br i1 %switch, label %func.nil, label %func.call1
```

### DIFF
--- a/transform/func-lowering.go
+++ b/transform/func-lowering.go
@@ -6,6 +6,7 @@ package transform
 import (
 	"sort"
 	"strconv"
+	"strings"
 
 	"github.com/tinygo-org/tinygo/compiler/llvmutil"
 	"tinygo.org/x/go-llvm"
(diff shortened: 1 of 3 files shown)


jaddr2line:
This fixes #1032

aykevl:
This caused a conflict, can you rebase on top of the dev branch to fix that?

jaddr2line:
I finally managed to rebase.

aykevl:
Tested the size and while it does affect code size on wasm (mainly), it is usually slightly reduced.
Will review the code soon.

aykevl:
Thank you for fixing this bug!

transform/func-lowering.go
```diff
@@ -55,17 +56,30 @@ func LowerFuncValues(mod llvm.Module) {
 	funcValueWithSignaturePtr := llvm.PointerType(mod.GetTypeByName("runtime.funcValueWithSignature"), 0)
```
aykevl, line 56: Sidenote: this will eventually need to be changed. I discovered that types may be renamed when LLVM modules are merged (when there are duplicates). This is a big problem in the interface lowering pass and I'm not yet sure how to fix that.

PR #1044, opened 2020-04-10
transform (coroutines): remove map iteration from coroutine lowering pass

jaddr2line:
The coroutine lowering pass had issues where it iterated over maps, sometimes resulting in non-deterministic output. This change removes many of the maps and ensures that the transformations are deterministic.

### DIFF
--- a/transform/coroutines.go
+++ b/transform/coroutines.go
@@ -115,13 +115,22 @@ type asyncFunc struct {
 	// callers is a set of all functions which call this async function.
 	callers map[llvm.Value]struct{}
 
-	// returns is a map of terminal basic blocks to their return kinds.
-	returns map[llvm.BasicBlock]returnKind
+	// returns is a list of returns in the function, along with metadata.
+	returns []asyncReturn
 
-	// calls is the set of all calls in the asyncFunc.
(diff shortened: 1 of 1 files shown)


transform/coroutines.go
```diff
@@ -451,46 +500,48 @@ func (c *coroutineLoweringPass) analyzeFuncReturns(fn *asyncFunc) {
...
 	}
 }
 
 // categorizeCalls categorizes all asynchronous calls into regular vs. async and matches them to their callers.
 func (c *coroutineLoweringPass) categorizeCalls() {
 	// Sort calls into their respective callers.
-	for _, async := range c.asyncFuncs {
-		async.calls = map[llvm.Value]struct{}{}
+	for _, async := range c.asyncFuncsOrdered {
+		async.calls = []llvm.Value{}
```
aykevl, line 512: You usually don't need to explicitly create a slice this way. Most operations are the same between nil slices and zero-length slices, except for comparing against nil. I believe that is the only difference.
(Maps are different, adding an element to a map requires the map to be created first).

jaddr2line: Fixed.

PR #1046, opened 2020-04-11
fix for invoking the tools when target is arm64

iansmith:
@aykevl I added the change for the problem we observed with picolibc.

This fix is necessary or you cannot compile picolibc due to a difference in the -I CFLAGS.  Weirdly, that seems to depend on the _target_.

This fix also passes the target, if supplied, to the extra assembly files when building.

It passes smoketests for me locally, with AVR=0.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -170,7 +170,11 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(stri
 		for i, path := range config.ExtraFiles() {
 			abspath := filepath.Join(root, path)
 			outpath := filepath.Join(dir, "extra-"+strconv.Itoa(i)+"-"+filepath.Base(path)+".o")
-			err := runCCompiler(config.Target.Compiler, append(config.CFlags(), "-c", "-o", outpath, abspath)...)
+			targ:=config.Target.Triple
+			if targ!="" {
+				targ="--target="+config.Target.Triple
+			}
+			err := runCCompiler(config.Target.Compiler, append(config.CFlags(), "-c", targ,"-o", outpath, abspath)...)
 			if err != nil {
 				return &commandError{"failed to build", path, err}
 			}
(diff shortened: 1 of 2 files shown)


builder/build.go
```diff
@@ -170,7 +170,11 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(stri
 		for i, path := range config.ExtraFiles() {
 			abspath := filepath.Join(root, path)
 			outpath := filepath.Join(dir, "extra-"+strconv.Itoa(i)+"-"+filepath.Base(path)+".o")
-			err := runCCompiler(config.Target.Compiler, append(config.CFlags(), "-c", "-o", outpath, abspath)...)
+			targ:=config.Target.Triple
+			if targ!="" {
+				targ="--target="+config.Target.Triple
+			}
+			err := runCCompiler(config.Target.Compiler, append(config.CFlags(), "-c", targ,"-o", outpath, abspath)...)
```
aykevl, line 177: Instead of adding the `--target` flag here, the `Target` struct should be set up correctly. Assuming this is for the Raspberry Pi 3, adding it to the `cflags` key of the JSON file should be enough (if not, you can print `config.CFlags()` here to check whether `--target` is already included).

aykevl: Also note that this code does not affect the compilation of picolibc.

iansmith: The other change in this commit is the picolibc one.

This change with the --target is to make the smoketests pass. My config for rpi3 sets the target properly. Do you want me to investigate why it is not set during tests?

iansmith: Oh, I see what you mean. Took me a sec. I'll fix this to not use this mechanism and remove this from the PR.

aykevl:
I think I found the underlying issue and have written a patch here: https://github.com/tinygo-org/tinygo/pull/1054. Basically it skips the `--sysroot` parameter entirely and uses a Clang-internal flag to set a system include path (which is what we want here).
Can you test it with your rpi3 branch?

aykevl:
I believe this has been fixed now that #1054 is merged. If not, please let me know.

PR #1048, opened 2020-04-12
rpi3 support with interrupts

iansmith:
* Allows you to set interrupts, although in a very simple way 
* Removed hack in builder that pushed triple to the assembler (now using rpi3.json)
* Included the change to builder/picolibc.go to make that deal with arm64 targets

@aykevl Please take a look.


aykevl:
This seems to be based on #1027? I think it's best to first finish and merge that PR.

iansmith:
Closed at Ayke's suggestion.  Doing it all in the rpi3_setup_files branch as requested.

PR #1051, opened 2020-04-13
compiler/llvm.go: fix typo

watsonso:
compiler/llvm.go: fix typo


aykevl:
Thank you for spotting this!
The CI tests failed but because this is just a comment change there is no way it could have caused those failures. Merging now.

PR #1052, opened 2020-04-13
flash: retry 3 times when attempting to reset the serial port

deadprogram:
This PR adds up to 3 retries when attempting to reset the target board by switching the serial port to 1200 baud speed change. Should hopefully help with boards that need a longer time to prepare their bootloader.


aykevl:
Looks reasonable and doesn't seem to affect the case when there is no error.

PR #1054, opened 2020-04-13
builder: fix picolibc include path

aykevl:
Previously we used `--sysroot` to set the sysroot explicitly. Unfortunately, this flag is not used directly by Clang to set the include path (`<sysroot>/include`) but is instead interpreted by the toolchain code. This means that even when the toolchain is explicitly set (using the `--sysroot` parameter), it may still decide to use a different include path such as `<sysroot>/usr/include` (such as on baremetal aarch64).

This commit uses the Clang-internal `-internal-isystem` flag which sets the include directory directly (as a system include path). This should be more robust.

The reason the `--sysroot` parameter has so far worked is that all existing targets happened to add `<sysroot>/include` as an include path.

The relevant Clang code is here:
https://github.com/llvm/llvm-project/blob/release/9.x/clang/lib/Driver/Driver.cpp#L4693-L4739
So far, RISC-V is handled by `RISCVToolchain`, Cortex-M targets by `BareMetal` (which seems to be specific to ARM unlike what the name says) and aarch64 fell back to `Generic_ELF`.

### DIFF
--- a/builder/picolibc.go
+++ b/builder/picolibc.go
@@ -12,7 +12,7 @@ var Picolibc = Library{
 	name: "picolibc",
 	cflags: func() []string {
 		picolibcDir := filepath.Join(goenv.Get("TINYGOROOT"), "lib/picolibc/newlib/libc")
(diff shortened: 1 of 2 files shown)


deadprogram:
Seems like this is working as expected. Can we merge it now?

aykevl:
Yes, this is ready to merge.

deadprogram:
OK, merging.

PR #1055, opened 2020-04-13
main: update version for beginning of v0.14 development cycle

jaddr2line:

Issue #918, opened 2020-02-25
-target arduino "PinInputPullup not declared by package machine".


yevsia:
Traget board : Arduino UNO.
Goal : Up-Down -> "Pull-up Resistor".
Result build : PinInputPullup not declared by package machine.

Code :
```
package main

import (
	"machine"
	"time"
)

const (
	led    = machine.Pin(13)
	button = machine.Pin(12)
)

func main() {
	led.Configure(machine.PinConfig{Mode: machine.PinOutput})
	button.Configure(machine.PinConfig{Mode: machine.PinInputPullup})

	for {
		if !button.Get() {
			led.High()
		} else {
			led.Low()
		}

		time.Sleep(time.Millisecond * 10)
	}
}
```
Comment : Arduino UNO don't support Pull-up Resistors?

deadprogram:
Many AVR processors support a built-in pullup resistor. This would just require some additional code, which we would be most pleased to accept should someone happen to submit a PR with this feature.

deadprogram:
Support for this feature was just merged into the `dev` branch as #1109 and will be part of the next release.

bobross419:
This (PinInputPullup) worked for me on arduino-mega2560 R3.  I was finally able to read digital button inputs.  Thanks for putting this together!

deadprogram:
This was released with v0.14.0 so now closing. Please reopen if needed. Thanks!

yevsia:
Thanks, works fine)

PR #1056, opened 2020-04-14
main: add support for LLVM 11

aykevl:
This can be useful to test improvements in LLVM master and to make it
possible to support LLVM 11 for the most part already before the next
release. That also allows catching LLVM bugs early to fix them upstream.

Note that tests do not yet pass for this LLVM version, but the TinyGo
compiler can be built with the binaries from apt.llvm.org (at the time
of making this commit).

---

One concrete example of what I want to do with this change is to test/improve AVR support: upstream master has a number of improvements that aren't included yet in LLVM 10.

Note that #1055 should be merged first.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -413,12 +413,12 @@ jobs:
     steps:
       - test-linux:
           llvm: "10"
-  test-llvm10-go115:
+  test-llvm11-go115:
     docker:
       - image: circleci/golang:1.15-buster
     steps:
       - test-linux:
-          llvm: "10"        
+          llvm: "11"
   assert-test-linux:
     docker:
       - image: circleci/golang:1.14-stretch
@@ -450,7 +450,7 @@ workflows:
       - test-llvm10-go112
       - test-llvm10-go113
       - test-llvm10-go114
-      - test-llvm10-go115
+      - test-llvm11-go115
       - build-linux
       - build-macos
       - assert-test-linux
--- a/builder/build.go
+++ b/builder/build.go
@@ -129,6 +129,17 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 		}
 	}
 
+	// LLVM 11 by default tries to emit tail calls (even with the target feature
+	// disabled) unless it is explicitly disabled with a function attribute.
+	// This is a problem, as it tries to emit them and prints an error when it
+	// can't with this feature disabled.
+	// Because as of september 2020 tail calls are not yet widely supported,
+	// they need to be disabled until they are widely supported (at which point
+	// the +tail-call target feautre can be set).
+	if strings.HasPrefix(config.Triple(), "wasm") {
+		transform.DisableTailCalls(mod)
+	}
+
 	// Make sure stack sizes are loaded from a separate section so they can be
 	// modified after linking.
 	var stackSizeLoads []string
--- a/builder/objcopy.go
+++ b/builder/objcopy.go
@@ -57,7 +57,7 @@ func extractROM(path string) (uint64, []byte, error) {
 
 	progs := make(progSlice, 0, 2)
 	for _, prog := range f.Progs {
-		if prog.Type != elf.PT_LOAD || prog.Filesz == 0 {
+		if prog.Type != elf.PT_LOAD || prog.Filesz == 0 || prog.Off == 0 {
 			continue
 		}
 		progs = append(progs, prog)
@@ -69,6 +69,12 @@ func extractROM(path string) (uint64, []byte, error) {
 
 	var rom []byte
 	for _, prog := range progs {
+		romEnd := progs[0].Paddr + uint64(len(rom))
+		if prog.Paddr > romEnd && prog.Paddr < romEnd+16 {
+			// Sometimes, the linker seems to insert a bit of padding between
+			// segments. Simply zero-fill these parts.
+			rom = append(rom, make([]byte, prog.Paddr-romEnd)...)
+		}
 		if prog.Paddr != progs[0].Paddr+uint64(len(rom)) {
 			return 0, nil, objcopyError{"ROM segments are non-contiguous: " + path, nil}
 		}
--- a/cgo/libclang_config.go
+++ b/cgo/libclang_config.go
@@ -1,5 +1,5 @@
 // +build !byollvm
-// +build !llvm9
+// +build !llvm9,!llvm11
 
 package cgo
 
--- a/cgo/libclang_config_llvm11.go
+++ b/cgo/libclang_config_llvm11.go
@@ -1,0 +1,14 @@
+// +build !byollvm
+// +build llvm11
+
+package cgo
+
+/*
+#cgo linux   CFLAGS:  -I/usr/lib/llvm-11/include
+#cgo darwin  CFLAGS:  -I/usr/local/opt/llvm@11/include
+#cgo freebsd CFLAGS:  -I/usr/local/llvm11/include
+#cgo linux   LDFLAGS: -L/usr/lib/llvm-11/lib -lclang
+#cgo darwin  LDFLAGS: -L/usr/local/opt/llvm@11/lib -lclang -lffi
+#cgo freebsd LDFLAGS: -L/usr/local/llvm11/lib -lclang
+*/
+import "C"
--- a/go.mod
+++ b/go.mod
@@ -10,5 +10,5 @@ require (
 	github.com/marcinbor85/gohex v0.0.0-20200531091804-343a4b548892
 	go.bug.st/serial v1.0.0
 	golang.org/x/tools v0.0.0-20200216192241-b320d3a0f5a2
-	tinygo.org/x/go-llvm v0.0.0-20200503225853-345b2947b59d
+	tinygo.org/x/go-llvm v0.0.0-20200503224449-70c558526021
 )
--- a/go.sum
+++ b/go.sum
@@ -46,5 +46,5 @@ golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898 h1:/atklqdjdhuosWIl6AIbO
 golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
 gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
 gopkg.in/yaml.v2 v2.2.2/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
-tinygo.org/x/go-llvm v0.0.0-20200503225853-345b2947b59d h1:hcX7vpB067GWM/EH4sGGOti0PMgIx+0bbZwUXctOIvE=
-tinygo.org/x/go-llvm v0.0.0-20200503225853-345b2947b59d/go.mod h1:fv1F0BSNpxMfCL0zF3M4OPFbgYHnhtB6ST0HvUtu/LE=
+tinygo.org/x/go-llvm v0.0.0-20200503224449-70c558526021 h1:d8T98WXGjrTgDmMXgxa6nb9EAYXGXwnzXygnJl6d+ac=
+tinygo.org/x/go-llvm v0.0.0-20200503224449-70c558526021/go.mod h1:fv1F0BSNpxMfCL0zF3M4OPFbgYHnhtB6ST0HvUtu/LE=
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -3,6 +3,7 @@ package interp
 import (
 	"io/ioutil"
 	"os"
+	"regexp"
 	"strings"
 	"testing"
 
@@ -66,6 +67,8 @@ func runTest(t *testing.T, pathPrefix string) {
 	}
 }
 
+var alignRegexp = regexp.MustCompile(", align [0-9]+$")
+
 // fuzzyEqualIR returns true if the two LLVM IR strings passed in are roughly
 // equal. That means, only relevant lines are compared (excluding comments
 // etc.).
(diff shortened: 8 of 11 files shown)


yevsia:
Try to fix this problem please [https://github.com/tinygo-org/tinygo/issues/918](url), if it's possible in new LLVM...

aykevl:
@yevsia the LLVM version and #918 are unrelated. You're very welcome to add a patch to implement pullups on AVR.

deadprogram:
Looks like a rebase is needed...

QuLogic:
In Fedora, we are already starting to use LLVM 11 for Fedora 33. I started testing out TinyGo 0.14.1 which obviously doesn't yet support it. I pulled the `llvm11` branch from go-llvm, which is fortunately sufficient to build and run its tests. I have not done any deep investigation in these, but for TinyGo itself, there remain a few problems:
* The old commit of wasi-libc is a bit broken with clang 11:
  * Differences in `predefined_macros.txt`; this is fixed upstream by https://github.com/WebAssembly/wasi-libc/pull/187.
  * Differences in `undefined-symbols.txt`; I see some upstream comments that that file won't be needed (https://github.com/WebAssembly/wasi-libc/issues/209#issuecomment-658319203), but haven't found any changes to that effect yet.
  * Various warnings from incorrect functions calls; these are probably fixed by upstream's update to [musl 1.1.24](https://github.com/WebAssembly/wasi-libc/commit/79a9b40837bfbf27dfbc991182ab8c4f5e98ba09) and/or [musl 1.2.0](https://github.com/WebAssembly/wasi-libc/commit/575e1579a4ebaa6dccb884ca657188a92982af23), but since these are non-fatal, I did not confirm or test yet.
* Some smoke tests fail:
  * for `circuitplay-express`, `pca10040`, `microbit`, `microbit-s110v8`, `itsybitsy-m0`, `itsybitsy-m4`, `feather-m0`, `trinket-m0`, `clue_alpha`, `pybadge`, `metro-m4-airlift`, `pyportal`, `pca10056-s140v7`, `reelboard-s140v7`, `wioterminal`, `pygamer`, `xiao`, `feather-nrf52840` with errors similar to:
     ```
     tinygo build -size short -o test.hex -target=circuitplay-express examples/i2s
        8888       8    4264 |    8896    4272
     error: ROM segments are non-contiguous: /tmp/tinygo064621349/main
     ```
  * for `gameboy-advance` with link errors:
     ```
     tinygo build -size short -o test.gba -target=gameboy-advance     examples/gba-display || true
     ld.lld: warning: lld uses blx instruction, no object with architecture supporting feature detected
     ld.lld: error: no memory region specified for section '.text.examples/gba-display.main'
     ld.lld: error: no memory region specified for section '.text.handleInterrupt'
     ld.lld: error: no memory region specified for section '.text.main'
     ld.lld: error: no memory region specified for section '.text.memset'
     error: failed to link /tmp/tinygo264710633/main: exit status 1
     ```
  * for `nintendo-switch` with a GC error:
     ```
     tinygo build             -o test.elf -target=nintendoswitch      examples/serial || true
     musttail call must precede a ret with an optional bitcast
       musttail call fastcc void @"(*internal/task.Queue).Push"(%"internal/task.Task"* nonnull %task.current1), !dbg !1213
     error: GC pass caused a verification failure
     ```
  * `wasm-ld` doesn't understand `--no-threads`; it looks like this was a simple rename: https://reviews.llvm.org/D76885 Naturally, they said "--no-threads is used rarely."... This is a straightforward switch to `--threads=1`, though wasi-sdk removed it entirely, so it may not be necessary.

niaow:
```
ld.lld: warning: lld uses blx instruction, no object with architecture supporting feature detected
```
Huh, I thought ditching thumb solved this, but maybe not?

```
musttail call must precede a ret with an optional bitcast
```
I don't see how tinygo could create this by itself? Unfortunately, this looks like it could just be an LLVM bug?

QuLogic:
Surprisingly, updating to latest commit of wasi-libc (https://github.com/WebAssembly/wasi-libc/commit/215adc8ac9f91eb055311acc72683fd2eb1ae15a) 'just worked', and was able to compile. I'm not sure exactly which commit fixed it and if you want update directly to it (there are [many large updates](https://github.com/WebAssembly/wasi-libc/compare/a280fead2ae71b9a230d3b48c1f95867431888e4...215adc8ac9f91eb055311acc72683fd2eb1ae15a) like the above-mentioned musl updates). This plus the `wasm-ld` ldflags change are enough to fix wasm builds. Unfortunately, `go test` fails with:
```
error: <unknown>:0:0: in function runtime.chanSend.resume void (%runtime.chanSend.Frame*): WebAssembly 'tail-call' feature not enabled
```

QuLogic:
> ```
> ld.lld: warning: lld uses blx instruction, no object with architecture supporting feature detected
> ```
> 
> Huh, I thought ditching thumb solved this, but maybe not?

I reverted Clang/LLD/LLVM to 10 and this warning still happens, so I guess not? It's not specific to 11 at least.

QuLogic:
> * Differences in `undefined-symbols.txt`; I see some upstream comments that that file won't be needed ([WebAssembly/wasi-libc#209 (comment)](https://github.com/WebAssembly/wasi-libc/issues/209#issuecomment-658319203)), but haven't found any changes to that effect yet.

I missed it earlier, but this one is fixed by https://github.com/WebAssembly/wasi-libc/pull/152.

So the minimum commit to avoid the two diff errors is https://github.com/WebAssembly/wasi-libc/commit/38b930a26b60558522bc34adc7621fff17925b4a, which is about halfway between what TinyGo uses and the latest commit.

QuLogic:
The `musttail` assert seems to have come from https://reviews.llvm.org/D79258, but I dumped ssa (`-dumpssa`) and IR `(-printir`) and see no mention of it in tinygo's stuff. You may be right that it's an LLVM bug.

aykevl:
Looking into this now.

aykevl:
Right now I'm hitting this:

```
error: <unknown>:0:0: in function runtime.chanSend.resume void (%runtime.chanSend.Frame*): WebAssembly 'tail-call' feature not enabled
```

... which I don't quite understand.

aykevl:
I have made a separate PR to update the wasi-libc dependency: https://github.com/tinygo-org/tinygo/pull/1309

QuLogic:
The tail-call thing seems to have to do with the `channel.go` test and compiling IR to an object file. That is, if you create bitcode or IR, then it is fine with it:
```bash
$ go run . build -target wasm -wasm-abi=js -print-stacks -o channel.o testdata/channel.go                                
error: scheduler.go:76:15: in function runtime.chanSend.resume void (%runtime.chanSend.Frame*): WebAssembly 'tail-call' feature not enabled

exit status 1

$ go run . build -target wasm -wasm-abi=js -print-stacks -o channel.bc testdata/channel.go

$ go run . build -target wasm -wasm-abi=js -print-stacks -o channel.ll testdata/channel.go
```

There seem to be [many wasm features](https://github.com/llvm/llvm-project/blob/65eb11306e921bb0299100dfc61e79858f903c1b/llvm/lib/Target/WebAssembly/WebAssembly.td) that are not listed in `targets/wasm.json`, but even still I was not able to figure out how to get `wasm-ld` to enable them. I could pass complete garbage to the `--features` option and it wouldn't complain. So this didn't work:
```bash
$ wasm-ld --features=+tail-call --allow-undefined --stack-first --export-all --no-demangle $TINYGOROOT/lib/wasi-libc/sysroot/lib/wasm32-wasi/libc.a channel.bc -o channel.wasm
wasm-ld: error: scheduler.go:76:15: in function runtime.chanSend.resume void (%runtime.chanSend.Frame*): WebAssembly 'tail-call' feature not enabled
wasm-ld: error: scheduler.go:76:15: in function runtime.chanSelect.resume void (%runtime.chanSelect.Frame*): WebAssembly 'tail-call' feature not enabled
wasm-ld: error: scheduler.go:76:15: in function runtime.chanRecv.resume void (%runtime.chanRecv.Frame*): WebAssembly 'tail-call' feature not enabled
```

I was able to get a further using `llc`, as it has a `--disable-tail-calls` option:
```
$ llc --filetype=obj --march=wasm32 --disable-tail-calls -o channel.o channel.ll  # also accept channel.bc

$ wasm-ld --allow-undefined --stack-first --export-all --no-demangle $TINYGOROOT/lib/wasi-libc/sysroot/lib/wasm32-wasi/libc.a -o channel.wasm channel.o

$ node targets/wasm_exec.js channel.wasm 
len, cap of channel: 1 2 false
len, cap of channel: 0 0 false
recv from open channel: 1 true
received num: 2
received num: 3
slept
received num: 4
received num: 5
received num: 6
received num: 7
received num: 8
recv from closed channel: 0 false
complex128: (+7.000000e+000+1.050000e+001i)
sum of n: 149
sum: 25
sum: 29
sum: 33
sum(100): 4950
deadlocking
select no-op
after no-op
sum: 5
did send one
select one n: 0
select n from chan: 55
select n from closed chan: 0
select send
sum: 235
non-concurrent channel recieve: 1
non-concurrent channel recieve: 2
closed buffered channel recieve: 3
closed buffered channel recieve: 4
closed buffered channel recieve: 0
hybrid buffered channel recieve: 2
blocking select sum: 3
```
So if you can figure out how to transform that `--disable-tail-calls` from `llc` into something that affects `LLVMTargetMachineEmitToMemoryBuffer`, then I think everything should work again.

QuLogic:
Another path with `llc` that I have not totally figured out yet, its `-mcpu` option has a `bleeding-edge` CPU that seems able to enable all features:
```bash
$ llc --filetype=obj --march=wasm32 -mcpu=bleeding-edge -o channel.o channel.ll
```
but it maybe enables too much:
```bash
$ wasm-ld --allow-undefined --stack-first --export-all --no-demangle $TINYGOROOT/lib/wasi-libc/sysroot/lib/wasm32-wasi/libc.a -o channel.wasm channel.o
wasm-ld: error: 'atomics' feature is used by channel.o, so --shared-memory must be used
```
and that then requires a larger memory size:
```bash
$ wasm-ld --shared-memory --allow-undefined --stack-first --export-all --no-demangle $TINYGOROOT/lib/wasi-libc/sysroot/lib/wasm32-wasi/libc.a -o channel.wasm channel.o
wasm-ld: error: maximum memory too small, 67344 bytes needed
```
which compiles, but doesn't run:
```bash
$ wasm-ld --shared-memory --max-memory=131072 --allow-undefined --stack-first --export-all --no-demangle $TINYGOROOT/lib/wasi-libc/sysroot/lib/wasm32-wasi/libc.a -o channel.wasm channel.o

$ node targets/wasm_exec.js channel.wasm 
[CompileError: WebAssembly.instantiate(): invalid memory limits flags @+718]
```
I don't know what to fix there, but it does suggest that if you could figure out how to enable features like tail-call, but not the other stuff like atomics, you might be able compile and run.

aykevl:
@QuLogic thanks a lot for all the research you did! Yes, it appears to be a target feature and `"features": ["+tail-call"]` does indeed work, but I don't think that's a solution: Chrome hasn't even added support for tail calls. Instead, we need to convince LLVM to not try to emit tail calls.

aykevl:
Also, it's not a linker issue but a compiler issue. This is how you can reproduce it without invoking the linker:

    tinygo build -o test.o -target=wasm ./testdata/channel.go

aykevl:
I think I have a solution, let's see what CI thinks of it.

aykevl:
Tests passed, so this should be ready for merging. The only thing I noticed was that the commit message is wrong (it says tests don't pass), if approved I can fix that while merging (or the one who merges it can fix it during "squash and merge").

In essence, to fix that last error I added a `"disable-tail-calls"="true"` attribute to every function when compiling to WebAssembly, which does exactly what it says.

aykevl:
I compared code size and while it obviously changed, it went both ways. Some programs increased in size, others decreased.

QuLogic:
Do you mean with LLVM11, or with/without tail calls?

aykevl:
With LLVM 11. With/without tail calls isn't really a useful statistic as it isn't enabled by default in any major browser and it wasn't enabled in LLVM 10 anyway (otherwise wasm output would not have worked in a browser).

QuLogic:
> * for `nintendo-switch` with a GC error:
>   ```
>   tinygo build             -o test.elf -target=nintendoswitch      examples/serial || true
>   musttail call must precede a ret with an optional bitcast
>     musttail call fastcc void @"(*internal/task.Queue).Push"(%"internal/task.Task"* nonnull %task.current1), !dbg !1213
>   error: GC pass caused a verification failure
>   ```

This one is fixed by #1368.

Then this PR fixes everything else with LLVM 11.

QuLogic:
I suppose you're waiting for the final LLVM 11 release, but to let you know, I've already been building this against LLVM 11 [on Rawhide](https://koji.fedoraproject.org/koji/buildinfo?buildID=1613687) and it's worked fine for all tests.

deadprogram:
LLVM 11 has just been released: https://lists.llvm.org/pipermail/llvm-announce/2020-October/000089.html

aykevl:
I'm not waiting for anything - I'd like to see this PR merged. Is there anything blocking it?

deadprogram:
No blockers from my side, I misunderstood the status.

Now merging, thanks @aykevl and @QuLogic

deadprogram:
Getting CI failures such as 

```
wasm-ld-11: error: unknown argument: --no-threads
``` 

which appears in https://app.circleci.com/pipelines/github/tinygo-org/tinygo/3773/workflows/2ff8b6ab-e33b-4946-81e3-b68f66dae25d/jobs/19529

QuLogic:
That should have been removed by #1308?

Also, don't forget to merge the llvm11 branch in https://github.com/tinygo-org/go-llvm

deadprogram:
Just added https://github.com/tinygo-org/tinygo/pull/1437 which should correct that build error, I think? Just waiting for CI to finish.

deadprogram:
Yes, confirm that #1437 corrects that build error.

aykevl:
> That should have been removed by #1308?

That was for targets/wasm.json. Presumably targets/wasi.json was a copy of it, so also copied `--no-threads`.
Merged the PR so the dev branch should be green again.

PR #1058, opened 2020-04-15
runtime: do not put scheduler and GC code in the same section

aykevl:
This allows dead code elimination and avoids linker errors with `-scheduler=leaking`.

### DIFF
--- a/src/runtime/scheduler_avr.S
+++ b/src/runtime/scheduler_avr.S
@@ -188,6 +188,7 @@ tinygo_switchToScheduler:
     // Return into the scheduler, as if tinygo_switchToTask was a regular call.
     ret
 
+.section .text.tinygo_scanCurrentStack
 .global tinygo_scanCurrentStack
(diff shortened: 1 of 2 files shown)


aykevl:
Note that the GBA is still broken:

```
$ tinygo build -o test.elf -target=gameboy-advance -size=short ./testdata/print.go
ld.lld-10: warning: lld uses blx instruction, no object with architecture supporting feature detected
ld.lld-10: error: undefined symbol: tinygo_scanCurrentStack
>>> referenced by gc_stack_raw.go:15 (src/runtime/gc_stack_raw.go:15)
>>>               /tmp/tinygo577525292/main.o:(runtime.alloc)
error: failed to link /tmp/tinygo577525292/main: exit status 1
```

jaddr2line:
Aside from the `blx`, is there anything in `scheduler_cortexm` that would not work for GBA?

deadprogram:
Merging.

Issue #894, opened 2020-02-07
GBA: lld inserts unsupported blx instruction


trevor403:
It seems that division is still not possible on the GBA target.

Before #846 if you attempted to compile a function that used division, you would get an error due to missing arithmetic library functions from LLVM like `__umodsi3` (see https://github.com/tinygo-org/tinygo/issues/843#issuecomment-573540947)

It also seems as though GC and Optimization settings play into how well behaved code will run on GBA. I've noticed that the CompilerRT functions seem to hang the GBA program. It looks like this is due to stack corruption.

This makes it impossible to do something like `strconv.FormatUint(uint64(104), 10)`

This may also be due to the fact that 64bit data types don't fit into the registers on the GBA but I have also tried with 32bit math and get similar failures. 

I will comment back shortly with the following
- branch used
- test program
- build command

aykevl:
> This may also be due to the fact that 64bit data types don't fit into the registers on the GBA but I have also tried with 32bit math and get similar failures.

What the compiler supports and what the architecture supports is mostly separate. Example: you can use int64_t and float on AVR (with a C compiler) while the architecture supports none of that. The compiler will make sure to either split up the operation into multiple instructions (example: `add`/`adc`) or make calls to a runtime library such as compiler-rt.

> It also seems as though GC and Optimization settings play into how well behaved code will run on GBA. I've noticed that the CompilerRT functions seem to hang the GBA program. It looks like this is due to stack corruption.

I would be very surprised if it was stack corruption. Why do you think it's this?

A test program would be very much appreciated, that way I can debug this locally.

trevor403:
Sure here is an example

Here is some sample code https://github.com/trevor403/division-fail
It builds using the `tinygo/tinygo:0.12.0` docker image

trevor403:
Also note that if you switch the code at the bottom of the program to:
```
	// only use 100 to pass fastSmalls
	num := uint64(100)
	numString := strconv.FormatUint(uint64(num), 10)
	printString("num "+numString, 9)
```
This does not error unlike the loop. I think this is due to compile time optimizations for arithmetic with constants.

aykevl:
I ran the program with `tinygo gdb -target=gameboy-advance -size=short -opt=1` and found the bug:

```
(gdb) c
Continuing.

Program received signal SIGILL, Illegal instruction.
[...]

PR #1060, opened 2020-04-19
runtime: copy stack scan assembly for GBA 

jaddr2line:
This is a continuation of #1058 that attempts to fix the issue on GBA. It now appears to link, but the display example does not appear to run right now:

image

### DIFF
--- a/src/runtime/scheduler_gba.S
+++ b/src/runtime/scheduler_gba.S
@@ -1,0 +1,19 @@
+.section .text.tinygo_scanCurrentStack
+.global  tinygo_scanCurrentStack
+.type    tinygo_scanCurrentStack, %function
+tinygo_scanCurrentStack:
+    // Save callee-saved registers onto the stack.
+    mov r0, r8
+    mov r1, r9
+    mov r2, r10
+    mov r3, r11
+    push {r0-r3, lr}
+    push {r4-r7}
+
+    // Scan the stack.
+    mov r0, sp
+    bl tinygo_scanstack
+
+    // Restore stack state and return.
+    add sp, #32
+    pop {pc}
--- a/targets/gameboy-advance.json
+++ b/targets/gameboy-advance.json
@@ -25,7 +25,8 @@
 	],
 	"linkerscript": "targets/gameboy-advance.ld",
 	"extra-files": [
-		"targets/gameboy-advance.s"
+		"targets/gameboy-advance.s",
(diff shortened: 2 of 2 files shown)


jaddr2line:
I looked at the IR, and the only reason the allocator is included is in order to export `malloc`, which is not used. The collector is never being invoked, so I do not think that this is the cause, and it should probably be okay to merge this.

jaddr2line:
I found the cause of the breakage in the GBA example, and will file it separately. Add an infinite loop to the end of the display example to work around this for now if you want to test.

jaddr2line:
The GBA example also works without modification with `-scheduler=none`.

jaddr2line:
I rebased this now that #1058 was merged, and also renamed it to follow conventions. This is ready to be reviewed.

jaddr2line:
I did not mean to close this.

deadprogram:
The example compiled and linked. Adding the infinite loop then lets it run as expected. Hence, I think ready to merge now.

@jaddr2line you will submit another PR to address the issue with the example?

jaddr2line:
The issue with the example is an instance of #894. I do not really know how to fix it, and it is a lot more complicated of an issue.

src/runtime/scheduler_gba.S
```diff
@@ -0,0 +1,23 @@
+.section .text.tinygo_scanCurrentStack
+.global  tinygo_scanCurrentStack
+.type    tinygo_scanCurrentStack, %function
+tinygo_scanCurrentStack:
+    // Save callee-saved registers onto the stack.
+    #if defined(__thumb2__)
+    push {r4-r11, lr}
```
aykevl, line 7: The GBA doesn't support Thumb 2.
(In ARM mode, however, you could do the push as above).

jaddr2line:
> This change looks mostly good. However, I'm not sure about the fact that this code lives in a file named scheduler (it doesn't have anything to do with scheduling). Perhaps it would be better to move it to a common src/runtime/gc_arm.S?

I agree with this, but for now I just wanted to be consistent with what we have already. We may want to refactor this later, but I did not want to make this change more complicated than it needed to be.

src/runtime/scheduler_gba.S
jaddr2line: Fixed.

deadprogram:
Thanks for the refinements @jaddr2line and @aykevl now merging.

PR #1071, opened 2020-04-22
riscv: fix infinite loop in garbage collector

yannishuber:
This fixes an infinite loop in the riscv scheduler caused by the `ra` register which gets overwritten.

The bug occures when `runtime.GC()` is called.

### DIFF
--- a/src/runtime/scheduler_tinygoriscv.S
+++ b/src/runtime/scheduler_tinygoriscv.S
@@ -22,6 +22,9 @@ tinygo_scanCurrentStack:
    mv a0, sp
    call tinygo_scanstack
 
+   // Restore return address.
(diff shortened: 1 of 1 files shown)


yannishuber:
I am also thinking, shouldn't all registers (`s0-s11`) be restored before returning ?

aykevl:
Great, that's definitely a bug. I wonder why we haven't found it previously: we do RISC-V testing in CI (which includes GC tests).

> I am also thinking, shouldn't all registers (`s0-s11`) be restored before returning ?

No. These registers are all callee saved so `tinygo_scanstack` will restore them if they are changed. The reason they're saved here is to ensure they get scanned by the GC, which does not look at values saved in registers.

PR #1074, opened 2020-04-26
support for external board definition

iansmith:
@aykevl @deadprogram @jaddr2line 

Don't know what you guys think of this, but I've been getting some mileage out of being able to define a machine/board and tons of register maps without touching the tinygo source.

It's up to you, but I think the virtue is that somebody who installs with "brew" or similar package manager and may not have the source (or know how to access it) can write new machine/board definitions.

### DIFF
--- a/src/runtime/runtime_external.go
+++ b/src/runtime/runtime_external.go
@@ -1,0 +1,68 @@
+// +build machine_external
+
+package runtime
+
+type timeUnit int64
+var tickMicros = int64(1)
+var asyncScheduler = false
+
+/*
+ * Use the interface external_* below to expose to the tinygo runtime your
+ * external implementation of the runtime.  External here means outside the
+ * source tree.
+ *
+ * Example (and the //export is mandatory)
+ *
+ * //export runtime.external_putchar
+ * func putchar(c uint8) {
+ *  //your implementation goes here
+ * }
+*/
+
+//
+// external interface
+//
+
+//go:extern external_ticks()
+func external_ticks() timeUnit
+
+//go:extern external_sleep_ticks
+func external_sleep_ticks(timeUnit)
+
+//go:extern external_abort()
+func external_abort()
+
+//go:extern putchar
+func external_putchar(uint8)
+
+//go:extern external_postinit
+func external_postinit()
+
+//
+// connect the "internal" calls to the external
+//
+
+func ticks() timeUnit {
+	return external_ticks()
+}
+
+func sleepTicks(d timeUnit) {
+	external_sleep_ticks(timeUnit(d))
+}
+
+func postinit() {
+	external_postinit()
+}
+
+func putchar(b uint8) {
+	external_putchar(b)
+}
+
+func abort() {
+	external_abort()
+}
+
+//so main() outside the tree can call internal run
+func Run() {
+	run()
+}
--- a/src/machine/board_external.go
+++ b/src/machine/board_external.go
@@ -1,0 +1,4 @@
+// +build machine_external
+
+package machine
+


src/runtime/runtime_external.go
```diff
@@ -0,0 +1,68 @@
...
+ * func putchar(c uint8) {
+ *  //your implementation goes here
+ * }
+*/
+
+//
+// external interface
+//
+
+//go:extern external_ticks()
```
aykevl, line 26: `//go:extern` in TinyGo only works for global variables, not for functions. You should use `//export` instead.

src/runtime/runtime_external.go
```diff
@@ -0,0 +1,68 @@
...
+ * external implementation of the runtime.  External here means outside the
+ * source tree.
+ *
+ * Example (and the //export is mandatory)
+ *
+ * //export runtime.external_putchar
+ * func putchar(c uint8) {
+ *  //your implementation goes here
+ * }
+*/
```
aykevl, line 20: The Go convention is to use `//` comments instead of `/* ... */` comments, even for larger blocks.

src/runtime/runtime_external.go
```diff
@@ -0,0 +1,68 @@
...
+func external_putchar(uint8)
+
+//go:extern external_postinit
+func external_postinit()
+
+//
+// connect the "internal" calls to the external
+//
+
+func ticks() timeUnit {
```
aykevl, line 45: Why not just export these functions directly? For example:

```go
//export tinygo_ticks
func ticks() timeUnit
```

iansmith:
@aykevl I agree it has non-zero maintenance costs and there are some minor things above that need fixing.

The reason that I would for the inclusion of this is so that a target can be put in tinygo, like rpi3, that is suitable and straightforward without preventing others from doing more sophisticated things.

For context, I'm trying to do various things with the rpi3 that involve the memory mapped hardware.  The peripherals document (https://www.raspberrypi.org/documentation/hardware/raspberrypi/bcm2835/BCM2835-ARM-Peripherals.pdf) is 205 pages and contains a large number of memory mapped registers and associated constants.  In addition, there is the ARM architectural reference manual for the A53 chip which (it is 1000+ pages!) contains a great many more memory mapped registers and constants for controlling how the chip behaves.  There are also undocumented features of the peripherals (e.g. random number generator) that are memory mapped, and a complex GPU (Broadcom VideoCore IV) that is partially memory mapped.  Naturally, all of these memory mapped devices interact with the MMU if one is use memory protection or virtual address translation.

Compounding this is that the linker scripts that are needed for a device as complex as this tend to be different for different situations. For example, I use a linker script that allows me to include a bitmap font inside my binary so I can draw text on the screen via the GPU.  That's not needed just to use the UART for output.  Similarly, the link-time behavior of something booting "raw" on bare metal needs a different behavior than something starting from a bootloader.  The needed link behavior is controllable with parameters that can be passed to the boot process, so they are not identical for all users.  Even the defaults differ between different board revisions!

I argue that a "simple, straightforward" bare metal target for RPI3 is a good thing for tinygo.  Doing something "complete" would likely be 10s of thousands of lines of code, and would require near-constant changes to the tinygo sources to update memory mapped io, linker scripts, etc.  By just allowing external definitions, the basics can be in tinygo but folks doing more complex things can do so without needing to constantly hack the "rpi3 definition" in the source.

I want to continue using Tinygo, I'm getting a lot of use out of it! I'd like to contribute a "basic" RPI3 definition and I'd like folks that use my software to not need to change the source.

thanks, ian

aykevl:
So if I understand it correctly, there are a few reasons you want to supply external board definitions:

1. For custom linker scripts. I think the example you gave (GPU bitmaps) is something that would be useful for TinyGo in general: for example TinyFont would benefit from it a lot (CC @conejoninja). It might be solved with https://github.com/golang/go/issues/35950. Alternatively, you can have a default linker script that ships with TinyGo and override the linker script with a custom JSON target file (you can inherit the main rpi3 target and override the `linkerscript` property).
2. Controlling how the binary is loaded. I think that would be solved either with two different TinyGo targets ("raw" and from a bootloader) or again by providing a good-enough default with the possibility of overriding the linker script. Again, this is already possible today by overriding the `linkerscript` property.
3. Register definitions. This one is a bit more complicated.

I can see a few possible solutions for the register definition question.

1. Develop from within TinyGo. Additions to register descriptors should be quick and easy to merge, as they will generally have no side effects.
2. Write the register definitions in a machine readable format such as SVD (the Embedded Rust folks have some tooling to make this less painful). This is from time to time imported into TinyGo (possibly through the posborne/cmsis-svd repo) but projects that want to use the latest and greatest can generate the most up to date register definitions and use that instead. Another advantage of this approach is that other projects can benefit from it too.
3. Use Go modules. For example, use tinygo.org/x/devices/bcm2835 which is stuck at a particular version but can be freely updated to newer revisions by a package that needs it (or even be replaced with a different package). It seems to me like this would give the best results - unfortunately Go modules are not yet supported in TinyGo.

PR #1080, opened 2020-05-04
main: update go-llvm to fix macOS build problem

aykevl:
The go-llvm package was still trying to use LLVM version 9 on MacOS.


deadprogram:
Merging.

Issue #1077, opened 2020-05-01
flash: touchSerialPortAt1200bps() always returns an error on windows


sago35:
In my environment, touchSerialPortAt1200bps() always returns an error.
However, touchSerialPortAt1200bps() is used to transition to boot.
After the transition to boot, touchSerialPortAt1200bps() does not return an error.

Test:
windows10 pro
feather-m4

sago35:
When I add a debug print to touchSerialPortAt1200bps(),

```go
func touchSerialPortAt1200bps(port string) (err error) {
	retryCount := 3
	for i := 0; i < retryCount; i++ {
		// Open port
		p, e := serial.Open(port, &serial.Mode{BaudRate: 1200})
		if e != nil {
			fmt.Println("DEBUG:", i, e.Error()) // <<-- debug print
			time.Sleep(1 * time.Second)
			err = e
			continue
		}
		defer p.Close()

		p.SetDTR(false)
		return nil
	}
	return fmt.Errorf("opening port: %s", err)
}
```

I get something like this.
When serial.Open() returns "Invalid serial port", we are transitioning to boot.

```
DEBUG: 0 Invalid serial port
DEBUG: 1 Serial port not found
DEBUG: 2 Serial port not found
opening port: Serial port not found
```

sago35:
Only the feather-m4 is connected to the USB.
The COM port is set to the feather-m4 port.

sago35:
In TinyGo, go.bug.st/serial.Open returns an error because it is resetting immediately at the following location.
It returns an error, but it transitions to boot.

https://github.com/tinygo-org/tinygo/blob/dev/main.go#L480

When using arduino software, serial.Open is not an error.
So tinygo flash succeeds.
The reason is that it is reset after 250 ticks.

https://github.com/adafruit/ArduinoCore-samd/blob/master/cores/arduino/USB/CDC.cpp#L125

```c++
bool CDC_Setup(USBSetup& setup)
{
    // ...
    if (requestType == REQUEST_HOSTTODEVICE_CLASS_INTERFACE)
    {
        // ...
        if (r == CDC_SET_LINE_CODING || r == CDC_SET_CONTROL_LINE_STATE)
        {
            // auto-reset into the bootloader is triggered when the port, already
            // open at 1200 bps, is closed. We check DTR state to determine if host 
            // port is open (bit 0 of lineState).
[...]

PR #1085, opened 2020-05-07
flash: fix touchSerialPortAt1200bps on windows

sago35:
#1077 

This PR solves a problem that causes tinygo flash to fail in a windows environment.
The fix isn't the right one, but it's a step forward because until now it wasn't possible to `tinygo flash` the `flash-method == msd target` on windows.

I use runtime.GOOS to affect only windows.
Also, even if you get the judgement wrong, it should fail on subsequent msd-volume-name checks.

BEFORE: We couldn't tinygo flash in windows + msd
AFTER: We can do tinygo flash with windows + msd too!

I'd like to see it merged, as it would be more convenient for windows users (who don't have jlink, etc.).

### DIFF
--- a/main.go
+++ b/main.go
@@ -479,6 +479,13 @@ func touchSerialPortAt1200bps(port string) (err error) {
 		// Open port
 		p, e := serial.Open(port, &serial.Mode{BaudRate: 1200})
 		if e != nil {
+			if runtime.GOOS == `windows` {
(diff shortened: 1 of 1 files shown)


deadprogram:
Finally was able to test this, and it worked as promised. Thank you very much @sago35 now merging.

sago35:
Thank you.

PR #1089, opened 2020-05-09
machine/arduino-nano33: use USB flag and remove debug flag for bossac flasher

deadprogram:
This PR uses the `-U` flag to ensure that device can be found by `bossac` flashing tool when the Arduino Nano33 IoT is not connected to the default port.


deadprogram:
I also added another commit to remove the `-d` debug flag to reduce console noise when performing flashing.

PR #1090, opened 2020-05-09
ci: add archlinux release job

johanbrandhorst:
Adds the arch-release job, which automatically updates
the tinygo-bin AUR package on every new git tag with
a semver version.

This will require setting the `$ARCH_RELEASE_SSH_PRIVATE_KEY` secret CI variable to a value that I have transferred to @deadprogram.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -319,7 +319,53 @@ commands:
             - ~/.cache/go-build
             - ~/.cache/tinygo
             - /go/pkg/mod
-
+  arch-release:
+    steps:
+      - run:
+          name: Install dependencies
+          command: pacman -Sy --noconfirm openssh git pacman-contrib binutils
+      - run:
+          name: Create TinyGo user
+          command: useradd -m tinygo
+      - run:
+          name: Become TinyGo user
+          command: su tinygo
+      - run:
+          name: Start SSH Agent
+          command: eval $(ssh-agent -s)
+      - run:
+          name: Add ARCH_RELEASE_SSH_PRIVATE_KEY identity
+          command: echo "${ARCH_RELEASE_SSH_PRIVATE_KEY}" | tr -d '\r' | ssh-add -
+      - run:
+          name: Create SSH directory
+          command: mkdir -p ~/.ssh && chmod 700 ~/.ssh
+      - run:
+          name: Add aur.archlinux.org to known hosts
+          command: ssh-keyscan aur.archlinux.org >> ~/.ssh/known_hosts
+      - run:
+          name: Clone tinygo-bin repo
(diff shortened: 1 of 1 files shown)


niaow:
Can you switch this to the `dev` branch?

johanbrandhorst:
> Can you switch this to the `dev` branch?

Done :)

.circleci/config.yml
```diff
@@ -319,7 +319,53 @@ commands:
...
+          name: Create TinyGo user
+          command: useradd -m tinygo
+      - run:
+          name: Become TinyGo user
+          command: su tinygo
+      - run:
+          name: Start SSH Agent
+          command: eval $(ssh-agent -s)
+      - run:
+          name: Add SSH_PRIVATE_KEY identity
```
deadprogram: Can we rename this `ARCH_RELEASE_SSH_PRIVATE_KEY`?

johanbrandhorst: Done!

.circleci/config.yml
```diff
@@ -375,3 +426,12 @@ workflows:
       - build-linux
       - build-macos
       - assert-test-linux
+  release:
+    jobs:
+      - arch-release:
+          filters:
+            branches:
+              ignore: /.*/
```
deadprogram, line 434: Just wondering, should we only do the release once the semver is in the `master` branch?

Semi-related: I would like to rename that to `release` branch soon.

deadprogram:
The needed ENV var has been added:

image

.circleci/config.yml
johanbrandhorst: This doesn't care about branches, only tags. If you make the tag against `master`, it will run on that tag. If we make releases on another branch and don't want to make arch-releases for those, we may need to tweak the tag regex.

deadprogram: I think @aykevl has been creating the tag on `dev` for each release.

aykevl: Release tags are created on the master branch by creating a GitHub release (which will automatically create a git tag). The process is that the new release commit (updating the CHANGELOG etc) is merged into dev, then fast-forward merged into master, and then a GitHub release is created.

> Semi-related: I would like to rename that to `release` branch soon.

Somewhat off-topic indeed but I agree, `release` describes the branch much better (alternatively, it could be `stable` which I've also seen on a few repositories).

deadprogram: That being the case, we can merge this, and then see what happens when we cut our next release.

deadprogram:
Merging now, so it is in place for the next release. Thank you @johanbrandhorst for your work on this!

PR #1093, opened 2020-05-10
avr: use standard pin numbering

aykevl:
This commit changes pin numbering for atmega328 based boards (Uno, Nano) to use the standard format, where pin number is determined by the pin/port. Previously, pin numbers were based on what the Uno uses, which does not seem to have a clear pattern.

One difference is that counting starts at port B, as there is no port A. So PB0 is 0, PB1 is 1… PC0 is 8.

This commit also moves PWM code to the atmega328 file, as it may not be generic to all ATmega chips.

---

This PR prepares for more changes (such as pin pullup, pin interrupts) which I wanted to do on a clean base.

Tested: blinking an LED on a few pins from all ports, the echo example, the button example, and the PWM example.


deadprogram:
Ran the integration tests, only thing worth noting is that all AVR code should now be changed to use the correct `machine.DX` constants, as the previous pure numeric values may not point to the pin you expect. Example is `D11` which is not the same as `11`. In any case, the tests pass with the correct naming, so now merging. Thanks for the cleanup @aykevl

aykevl:
Yeah this is definitely something that might break backwards compatibility, but I think it's for the better (not all atmega boards might use the pin numbering that's used by Adafruit).

PR #1095, opened 2020-05-11
sam: return an error when an incorrect PWM pin is used

### LINKED ISSUE #1038: PWM.Configure produces a nil pointer dereference
Device: `arduino-nano33`

The code in `src/examples/pwm/pwm.go` produces the following error:

```
```

it looks like the `PWM.getTimer` method is returning nil from `src/machine/machine_atsamd21.go:1196` which doesn't seem to have a matching set of Pin numbers to the `src/machine/board_arduino_nano33.go:13` file.

Are these numbers supposed to match the pins described as having PWM?

aykevl:
Previously it would trigger a nil pointer panic.

fixes #1038

### DIFF
--- a/src/examples/pwm/pwm.go
+++ b/src/examples/pwm/pwm.go
@@ -28,13 +28,16 @@ func main() {
 	machine.InitPWM()
 
 	red := machine.PWM{redPin}
-	red.Configure()
+	err := red.Configure()
+	checkError(err, "failed to configure red pin")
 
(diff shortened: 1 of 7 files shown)


deadprogram:
@aykevl this needs a merge conflict resolved, please.

aykevl:
Rebased.

deadprogram:
Thanks @aykevl now merging.

PR #1097, opened 2020-05-11
Chan and map support improvements

cornelk:


compiler/compiler.go
```diff
@@ -1201,9 +1201,7 @@ func (b *builder) createBuiltin(args []ssa.Value, callName string, pos token.Pos
 		var llvmCap llvm.Value
 		switch args[0].Type().(type) {
 		case *types.Chan:
-			// Channel. Buffered channels haven't been implemented yet so always
-			// return 0.
```
aykevl, line 1205: This used to be the case, but not anymore as you found out.

Issue #541, opened 2019-09-12
How to parse cli arguments?


dongweigogo:
Since tinygo does not support flag package currently, is there a way to get around this problem?

aykevl:
Exactly, we need to add support for the `os.Args` array, which should be initialized from here:

https://github.com/tinygo-org/tinygo/blob/7d481c179faa35c9b2d53774a2d13f923cd50591/src/runtime/runtime_unix.go#L47-L48

talentlessguy:
hi, any progress on this?

deadprogram:
I know that the Args stubs were merged. Has anyone tried to implement the needed OS support for linux or darwin (or windows)?

aykevl:
See #1410.

deadprogram:
This was released with v0.18 so now closing. Thank you!

PR #1098, opened 2020-05-11
os: add Args stub

cornelk:
Tnis allows code that relies on it to compile, the `flags` package compiles now. No real parameters are passed but Args[0] contains a static mock string.

Setting Args as recommended https://github.com/tinygo-org/tinygo/issues/541#issuecomment-531483733 did not work tho:

`tinygo/src/runtime/runtime_unix.go`
```
func main() int {
	args = []string{"/proc/self/exe2"}
	preinit()
	run()
```

has no effect, os.Args will still hold the old value

### DIFF
--- a/src/os/proc.go
+++ b/src/os/proc.go
@@ -9,6 +9,15 @@ import (
 	"syscall"
 )
 
+// Args hold the command-line arguments, starting with the program name.
+var Args []string
+
+func init() {
+	Args = runtime_args()
+}
+
+func runtime_args() []string // in package runtime
+
 // Exit causes the current program to exit with the given status code.
 // Conventionally, code zero indicates success, non-zero an error.
 // The program terminates immediately; deferred functions are not run.
--- a/src/runtime/runtime.go
+++ b/src/runtime/runtime.go
@@ -24,9 +24,12 @@ func GOROOT() string {
 	return "/usr/local/go"
 }
 
+// TODO: fill with real args.
+var args = []string{"/proc/self/exe"}
+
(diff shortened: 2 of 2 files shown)


aykevl:
Thank you, this is a welcome improvement to get more packages to compile (even if they aren't fully supported yet).

aykevl:
I've looked a bit into why you can't initialize the arguments in `runtime.main` and there are two reasons:

1. You can't do heap allocations before the `initHeap` call. This can be worked around by calling `malloc` directly.
2. The interp package gets in the way. It assumes that all global variables are initialized with a package initializer (globals, `init` functions). Therefore, setting argumens in the main function _at runtime_ will be ignored by package initializers, which often run _at compile time_ (mainly for code size).

sevki:
@aykevl i saw this got merged, but it seems to break the dev branch with `-gc=none` is there another issue tracking the progress of this or work that needs to be done... I'm waiting to start a new job in a month so I have plenty of free time on my hands :wink:

aykevl:
> i saw this got merged, but it seems to break the dev branch with `-gc=none`

Are you sure this is what broke `-gc=none`? `-gc=none` is already broken in the master branch (and I believe it has been broken for a while now):

```
tinygo flash -target=pca10040 -gc=none examples/blinky1
Global is external, but doesn't have external or weak linkage!
i8* ()* @runtime.alloc
error: verification failure after LLVM optimization passes
```

aykevl:
Here is a PR to fix the `-gc=none` bug: https://github.com/tinygo-org/tinygo/pull/1236

PR #1099, opened 2020-05-11
extend stdlib to allow import of more packages

### LINKED ISSUE #752: Add Runtime Error Interface
I would like to add Runtime Error interface defined here: https://golang.org/pkg/runtime/#Error
to solve the problem of compiling text/template/parse as defined here: https://tinygo.org/lang-support/stdlib/#text-template-parse

I am testing some wasm libraries and they include text-template-parse. Adding this error interface may solve the problem.

cornelk:
Closes #752

these changes allow a successful import of the following go 1.14 stdlib packages:

```
	_ "database/sql"
	_ "debug/elf"
	_ "debug/macho"
	_ "debug/pe"
	_ "debug/plan9obj"
	_ "html"
	_ "mime"
	_ "testing"
```


aykevl:
This PR conflicts with #1012, which I think should be merged first as it also cleans up the os package a bit.

cornelk:
@aykevl rebased now on current dev branch after #1012 got merged

src/os/errors.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"errors"
+)
+
+var (
+	ErrInvalid    = errors.New("invalid argument")
+	ErrPermission = errors.New("permission denied")
+	ErrClosed     = errors.New("file already closed")
+)
```
aykevl: These errors are really the same type of errors as the ones in file.go. I agree that it is a good idea to move them into a separate file, so can you move those errors also together with these errors?

src/os/errors.go
```diff
@@ -0,0 +1,32 @@
...
+
+func IsPermission(err error) bool {
+	return err == ErrPermission
```
aykevl, line 21: Not complete (it doesn't unwrap the error) but I guess it's fine for now.

src/reflect/type.go
```diff
@@ -455,11 +471,74 @@ type StructField struct {
...
+		tag = tag[i+1:]
+
+		if key == name {
+			// TODO strconv causes import cycle
+			//value, err := strconv.Unquote(qvalue)
+			//if err != nil {
+			//	break
+			//}
+			//return value, true
+			return qvalue, true
```
aykevl: You will probably need to reimplement this in the reflect package.

src/reflect/type.go
```diff
@@ -455,11 +471,74 @@ type StructField struct {
...
+type StructTag string
+
+// Get returns the value associated with key in the tag string.
+func (tag StructTag) Get(key string) string {
+	v, _ := tag.Lookup(key)
+	return v
+}
+
+// Lookup returns the value associated with key in the tag string.
+func (tag StructTag) Lookup(key string) (value string, ok bool) {
```
aykevl, line 496: Get and Lookup will need tests. Please add them to testdata/reflect.go. You can probably just try to lookup a few tags and if they are present, print them (just like many other properties of types/values are printed).

aykevl: I think it may be better to use something other than a string for `StructTag` eventually. The last time I looked through the standard library, no package was using `StructTag` as a string but instead just called the `Get` and `Lookup` functions. Therefore, I think it would be feasible to do the key/value splitting at compile time, avoiding the code size cost of doing it at runtime (and possibly speeding things up as well, although that's not the goal).

aykevl: Are you sure you actually need `Get` and `Lookup`? You could perhaps simplify things by stubbing them out. At least when I tested the encoding/json package, I think I managed to do without for a proof-of-concept.

cornelk: the `StructTag` type with it's functions is actively used in common libraries `github.com/BurntSushi/toml` and `gopkg.in/yaml.v2`

cornelk: I added your proposal to do it at compile time as a TODO comment. For this MR i focused on getting more libraries to compile.

cornelk:
@aykevl i updated the PR

src/reflect/type.go
aykevl: I checked both libraries.
* `github.com/BurntSushi/toml` passes the `.Tag` field as a `StructTag` to `getOptions` and only uses the `.Get` function. That would allow doing the key/value parsing at compile time.
* `gopkg.in/yaml.v2` unfortunately does use the raw string. It looks like they fall back to using the whole string if the struct tag is not in a key/value format. I think they shouldn't do this as it breaks the convention, but it's probably hard to persuade them to change this (there is probably code that depends on this behavior).

Not sure what to do here, maybe using raw strings is better after all (even though it will likely increase code size). Maybe there are optimizations possible here, but I'm not sure whether that's possible in a sane way.

deadprogram:
@aykevl any further feedback? Seems like all of the comments have been addressed.

deadprogram:
Reminder to @aykevl about this PR. LGTM.

src/reflect/strconv.go
```diff
@@ -0,0 +1,248 @@
...
+// license that can be found in the LICENSE file.
+
+package reflect
+
+import (
+	"unicode/utf8"
+)
+
+// ErrSyntax indicates that a value does not have the right syntax for the target type.
+var ErrSyntax = badSyntax{}
```
aykevl: This file adds a number of exported symbols. Because this is the reflect package, these (copied) symbols should not be exported.

src/reflect/type.go
aykevl: Did you write this code yourself or did you copy it? If you copied it, it should have a copyright note (just like the strconv.go file).

src/runtime/mprof.go
```diff
@@ -0,0 +1,6 @@
+package runtime
+
+// Stack is a stub, not implemented
+func Stack(buf []byte, all bool) int {
```
aykevl: You could add this to stack.go, which has similar functions (to avoid yet another small file).

None of these are likely to be implemented considering what TinyGo is designed for.

cornelk:
@aykevl comments are addressed, I will split it up next time.

deadprogram:
Thank you very much for working on this contribution, @cornelk now that all of the comments have been addressed, I am going to squash/merge. Thanks again!

deadprogram:
The tests previous passed, but now the CI fails for all Linux platforms. Perhaps rebasing again against `dev` would have been a good idea before I merged this. My bad.

We now need to fix this right away.

deadprogram:
Thanks to @jaddr2line root problem was discovered (of course unrelated to this PR) and I just pushed a fix to the CircleCI build.

PR #1100, opened 2020-05-11
cgo: add support for negative numeric constants in tokenizer

### LINKED ISSUE #770: CGo does not support macros with symbol values
```
xaionaro@ubuntu:~/go/src/github.com/u-root/u-root/bb$ CGO_ENABLED=0 tinygo build -o bb main.go 
# os/user
../../../../../../.gimme/versions/go1.13.linux.amd64/src/os/user/cgo_lookup_unix.go:19:6: not implemented: build constraints in #cgo line
/usr/include/x86_64-linux-gnu/bits/confname.h:222:31: unexpected token ILLEGAL
/usr/include/x86_64-linux-gnu/bits/confname.h:224:31: unexpected token ILLEGAL
/usr/include/x86_64-linux-gnu/bits/confname.h:222:31: unexpected token ILLEGAL
/usr/include/x86_64-linux-gnu/bits/confname.h:224:31: unexpected token ILLEGAL
/usr/include/x86_64-linux-gnu/bits/confname.h:222:31: unexpected token ILLEGAL
/usr/include/x86_64-linux-gnu/bits/confname.h:224:31: unexpected token ILLEGAL
```

```
xaionaro@ubuntu:~/go/src/github.com/u-root/u-root/bb$ head -224 /usr/include/x86_64-linux-gnu/bits/confname.h | tail -4
    _SC_GETGR_R_SIZE_MAX,
#define	_SC_GETGR_R_SIZE_MAX		_SC_GETGR_R_SIZE_MAX
    _SC_GETPW_R_SIZE_MAX,
#define	_SC_GETPW_R_SIZE_MAX		_SC_GETPW_R_SIZE_MAX
```

cornelk:
Closes #770

before:

```
# net
/usr/include/netdb.h:617:23: unexpected token ILLEGAL
/usr/include/netdb.h:618:22: unexpected token ILLEGAL
/usr/include/netdb.h:624:23: unexpected token ILLEGAL
```

failing lines were like:
`# define EAI_NONAME   -2    /* NAME or SERVICE is unknown.  */`

### DIFF
--- a/cgo/const.go
+++ b/cgo/const.go
@@ -131,7 +131,7 @@ func (t *tokenizer) Next() {
 			t.value = t.buf[:1]
 			t.buf = t.buf[1:]
 			return
-		case c >= '0' && c <= '9':
+		case (c >= '0' && c <= '9') || c == '-':
 			// Numeric constant (int, float, etc.).
 			// Find the last non-numeric character.
 			tokenLen := len(t.buf)
@@ -140,7 +140,7 @@ func (t *tokenizer) Next() {
 				if c == '.' {
 					hasDot = true
 				}
-				if c >= '0' && c <= '9' || c == '.' || c == '_' || c >= 'a' && c <= 'z' || c >= 'A' && c <= 'Z' {
+				if c >= '0' && c <= '9' || c == '-' || c == '.' || c == '_' || c >= 'a' && c <= 'z' || c >= 'A' && c <= 'Z' {
 					tokenLen = i + 1
 				} else {
 					break
--- a/cgo/const_test.go
+++ b/cgo/const_test.go
@@ -15,6 +15,7 @@ func TestParseConst(t *testing.T) {
 		Go string
 	}{
 		{`5`, `5`},
+		{`-5`, `-5`},
 		{`(5)`, `(5)`},
 		{`(((5)))`, `(5)`},
 		{`)`, `error: 1:1: unexpected token )`},


aykevl:
I think this will cause trouble in the future when expressions are implemented. For example: how should `5 -2` be parsed? Obviously as three tokens (`5` `-` `2`) to get an expression but with this change it would be parsed as `5` `-2`, and two consecutive numbers aren't valid C (like in `5 4`).

At the moment, the cgo package doesn't support parsing C expressions but it should eventually learn to do this. For example, the Nordic SoftDevice contains code like this for error messages:

```c
#define NRF_ERROR_SDM_BASE_NUM  (0x1000)
#define NRF_ERROR_SDM_INCORRECT_CLENR0                  (NRF_ERROR_SDM_BASE_NUM + 2)
```

aykevl:
I should add that your work is certainly appreciated, just that this particular PR may not be the best solution for negative numeric constants.

typeless:
I did an experiment using @cznic's https://gitlab.com/cznic/cc to parse the #defines. (Where the author uses it to translate the entire SQLite codebase.)
It almost works except that the `cc` package doesn't have a `stddef.h` shipped like a normal C compiler.
That can be solved by importing a `stddef.h` from somewhere else or just creating a dummy one for it.
However, I am not sure if it is the way worth pursuing as it still cannot solve my problem immediately (for the lack of say, symbols defined in the `os` package).

PS. I wanted to use TinyGo to compile CLI utilities runnin on embeded Linux devices.

aykevl:
If we're going to use a C parser, it'll be the same we use everywhere: libclang. It seems like a bad idea to me to use two different C parsers for different purposes.

That said, I really think the right way forward is not using a C parser at all but a hand-rolled parser that only parses the most common types of C expressions and convert those expressions to Go expressions, just like we do already: we parse the C string in the `#define` to a Go AST node and let the Go compiler compile that expression further. Maybe there are cases where that's not enough but I think it will work in the vast majority of `#define`s.

> However, I am not sure if it is the way worth pursuing as it still cannot solve my problem immediately (for the lack of say, symbols defined in the `os` package).
> 
> PS. I wanted to use TinyGo to compile CLI utilities runnin on embeded Linux devices.

Can you raise a new issue for this?

typeless:
@aykevl 
````
/home/mural/src/golang.org/go1.16.3/src/net/fd_unix.go:202:12: NewFile not declared by package os
/home/mural/src/golang.org/go1.16.3/src/net/pipe.go:156:16: ErrDeadlineExceeded not declared by package os
/home/mural/src/golang.org/go1.16.3/src/net/pipe.go:169:16: ErrDeadlineExceeded not declared by package os
/home/mural/src/golang.org/go1.16.3/src/net/pipe.go:188:16: ErrDeadlineExceeded not declared by package os
/home/mural/src/golang.org/go1.16.3/src/net/pipe.go:204:17: ErrDeadlineExceeded not declared by package os
/home/mural/src/golang.org/go1.16.3/src/net/cgo_resnew.go:21:17: cannot initialize 2 variables with 1 values
/home/mural/src/golang.org/go1.16.3/src/net/cgo_unix.go:36:52: undeclared name: C.GoString
/home/mural/src/golang.org/go1.16.3/src/net/cgo_unix.go:107:17: cannot initialize 2 variables with 1 values
/home/mural/src/golang.org/go1.16.3/src/net/cgo_unix.go:161:17: cannot initialize 2 variables with 1 values
/home/mural/src/golang.org/go1.16.3/src/net/cgo_unix.go:190:11: undeclared name: C.GoString
````
The above is what I got when compiling my app using `TinyGo` + `cc` + `stddef.h` from GCC 9.
https://github.com/tinygo-org/tinygo/issues/1888 looks like the same issue about the missing `ErrDeadlineExceeded`.
I think most of these, if not all, can be fixed by replacing the `net` package (https://github.com/tinygo-org/tinygo/issues/1602).

The remaining error messages like `undeclared name: C.GoString` and `cannot initialize 2 variables with 1 values` are more likely on my part though.

Issue #1091, opened 2020-05-09
Channels block forever in wasm


bradleypeabody:
Channels do not appear to be working properly (both send and receive block forever) in wasm.  Here's the specific info on the issue and then a more involved question below with which I'm trying to understand and decide what my next steps are for TinyGo+Vugu support.

Anything I've tried where main() is doing:
```
ch := make(chan bool, 1)
// do some JS setup
<-ch
```

And then some JS event happens and calls back into the Go code and a channel send i.e. `ch <- true` happens - the send blocks, and the the receive never resumes.

This reproduces it: https://github.com/bradleypeabody/tgtestb  On `master` the channels block as described above.  On `dev` it does not appear to arrive back into Go from JS at all, which is probably an unrelated problem.

The issue is very similar to this one: https://github.com/tinygo-org/tinygo/issues/791 - and I tried using the exact code from that issue and does not work (blocks forever, same as the example above).

---

With that said, I have two main questions:

1. **Is there already a good solution on the table for the "stack unwinding" problem in TinyGo+wasm?**  As I understand the problem (and I am by no means an expert on this so please correct me if I have this wrong), the Wasm instruction set does not have actual jump instructions like `je`, etc. but instead forces logic into condition blocks, and you can't manually save and restore the stack to implement these sorts of "and control transfers over to this context in this case" situations.  (Although it certainly appears like there is active discussion about mechanisms that could affect this or improve the situation e.g. https://github.com/WebAssembly/design/issues/1340 https://github.com/WebAssembly/exception-handling/issues/105)

I believe in a previous discussion, [Asyncify](https://kripken.github.io/blog/wasm/2019/07/16/asyncify.html) was mentioned as a possible solution.

If something like that isn't feasible, it might not even make sense to try to fully solve this until wasm provides some underlying suspend/resume mechanism that can be used for this (because that would make things much easier and saner and probably offload much of the work to LLVM).  On the other hand, realistically the necessary Wasm features could take years to make their way all the way into desktop browsers.

The reason I'm bringing all this up is I'm trying to get an idea of how well supported these cases will be, at least for the time being.

[...]

PR #1102, opened 2020-05-12
call scheduler from resume

bradleypeabody:
Adding scheduler call per discussion with @jaddr2line here https://github.com/tinygo-org/tinygo/issues/1091 resolves the issue where JS event handlers are not executing anything.  I tested this on two examples which were exhibiting the bug and both are working now.

### DIFF
--- a/src/runtime/runtime_wasm.go
+++ b/src/runtime/runtime_wasm.go
@@ -58,6 +58,7 @@ func resume() {
 	go func() {
 		handleEvent()
 	}()
+	scheduler()
 }
 
 //export go_scheduler


aykevl:
This looks reasonable to me, @jaddr2line can you perhaps double check that this is correct?

(I'm not sure how we could test for this, as we don't have any Go/JS tests right now)

bradleypeabody:
Fwiw on Vugu I used this https://hub.docker.com/r/chromedp/headless-shell/ as a headless browser, made a little test server program and then serve up sample wasm programs into the headless browser click around on them with https://github.com/chromedp/chromedp - that works pretty well.  I could take a crack at trying to set this up for tinygo with a few small test cases to start.  I believe the above pieces would be the only requirements and would only be needed for testing.  Is this worth looking into?

deadprogram:
That sounds like a very cool set of integration tests to me!

aykevl:
@jaddr2line thanks, just wanted to make sure I didn't miss anything important (as it seems like the kind of thing that could break in subtle ways).

PR #1105, opened 2020-05-12
arm: make FPU configuration consistent

aykevl:
Eventually we might want to start using the FPU, but the easy option right now is to simply disable it everywhere. Previously, it depended on whether Clang was built as part of TinyGo or it was an external binary (but Go code compiled with TinyGo never used the FPU). By setting the floating point mode explicitly, such inconsistencies are avoided.

This commit creates a new cortex-m4 target which can be the central place for setting FPU-related settings across all Cortex-M4 chips.

Tested on the pca10040 and the ItsyBitsy-m4.

### DIFF
--- a/builder/library.go
+++ b/builder/library.go
@@ -69,7 +69,7 @@ func (l *Library) Load(target string) (path string, err error) {
 	// Precalculate the flags to the compiler invocation.
 	args := append(l.cflags(), "-c", "-Oz", "-g", "-ffunction-sections", "-fdata-sections", "-Wno-macro-redefined", "--target="+target, "-fdebug-prefix-map="+dir+"="+remapDir)
 	if strings.HasPrefix(target, "arm") || strings.HasPrefix(target, "thumb") {
-		args = append(args, "-fshort-enums", "-fomit-frame-pointer")
+		args = append(args, "-fshort-enums", "-fomit-frame-pointer", "-mfloat-abi=soft")
 	}
 	if strings.HasPrefix(target, "riscv32-") {
 		args = append(args, "-march=rv32imac", "-mabi=ilp32", "-fforce-enable-int128")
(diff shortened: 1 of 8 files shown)


aykevl:
What we need before we can enable the FPU:

* Proper save/restore of FPU registers during task switching. This will have an overhead (64 byte of stack per goroutine and about 17 cycles).
* The right features enabled so that TinyGo will also emit floating point instructions instead of libcalls.

sago35:
PyPortal (atsamd51j20a) worked correctly with the binary in the CI results section of Azure Pipelines.

Later, I'll try feather-m4 (atsamd51j19a) and wioterminal (atsamd51p19a).

sago35:
The feather-m4 (atsamd51j19a) also works.

It works without changing the atsamd51j19a.json, is that as intended?
I'm thinking that it works because I changed the options in LLVM, is that correct?

sago35:
@aykevl 
If possible, I'd like to you to merge into dev.

sago35:
wioterminal (atsamd51p19a) also works.

sago35:
Please update this file too. @aykevl 

targets/atsamd51j19a.json

```json
{
	"inherits": ["cortex-m4"],
	"build-tags": ["atsamd51j19", "atsamd51", "sam"],
	"cflags": [
		"-Qunused-arguments"
	],
	"linkerscript": "targets/atsamd51.ld",
	"extra-files": [
		"src/device/sam/atsamd51j19a.s"
	]
}
```

deadprogram:
@aykevl do you want to make the change proposed by @sago35 before I merge this PR?

aykevl:
I've updated the atsamd51j19a, the nrf52840, and the stm32f4 with the same changes. Somehow I missed those (this time I grepped for `v7em`). It should be ready now, although I haven't tested all of those (I only tested floats on the nrf52840).

deadprogram:
Looks good and all tests passing. Now merging, thanks @aykevl and @sago35

PR #1106, opened 2020-05-13
ci: do not install the SiFive toolchain

aykevl:
We don't need it anymore since we use lld to link RISC-V binaries.


deadprogram:
:scissors: merging.

Issue #1103, opened 2020-05-12
build on ubuntu 20 fails


gshields:
with :
 clang --version
clang version 10.0.0-4ubuntu1 
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
 uname -a
Linux 6540 5.4.0-29-generic #33-Ubuntu SMP Wed Apr 29 14:32:27 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
go version 
go version go1.14.2 linux/amd64

verified that clang is working to compile other programs 

Downloaded tinygo source

~/go/tinygo$ make
Fetch and build LLVM first by running:
  make llvm-source
  make llvm-build

Make fails to compile tinygo version as of 5/11/20

aykevl:
What is the error you are getting? (I'm assuming in `make llvm-build`).

aykevl:
If you mean that this is the error:

```
~/go/tinygo$ make
Fetch and build LLVM first by running:
make llvm-source
make llvm-build
```

Then you need to do what the error message says: run `make llvm-build`.

However, I think what you really want is not `make` but `go install`.

gshields:
tried this:
 sudo dpkg -i tinygo_0.13.1_amd64.deb
[sudo] password for ghs:
Selecting previously unselected package tinygo.
(Reading database ... 222884 files and directories currently installed.)
Preparing to unpack tinygo_0.13.1_amd64.deb ...
Unpacking tinygo (0.13.1) ...
Setting up tinygo (0.13.1) ...
ghs@6540:~$ tinygo
tinygo: command not found
I looked in /usr/local and there was no tinygo
tried to uninstall package and found it was not install
dpkg -P tinygo_0.13.1_amd64
dpkg: warning: ignoring request to remove tinygo_0.13.1_amd64 which isn't
installed



Then I tried to install using git source since this was on the instruction
git clone --recursive https://github.com/tinygo-org/tinygo.git
cd tinygo

I follow your suggestion to use go and I used go get and it worked!
Thanks

aykevl:
> ```
> ghs@6540:~$ tinygo
> tinygo: command not found
> ```
> I looked in /usr/local and there was no tinygo

I tried installing myself and a `/usr/local` did appear.
You should also add `/usr/local/tinygo/bin` to your `$PATH` variable:

    export PATH=$PATH:/usr/local/tinygo/bin

> tried to uninstall package and found it was not install
> ```
[...]

PR #1107, opened 2020-05-13
ci: build .deb files along with .tar.gz files for Debian

### LINKED ISSUE #725: Deb package does not set path to tinygo command
After a fresh install on Kubuntu 19.04, the tinygo command is not in the PATH and thus can't be easily run.

The proposal is to add a symlink like so `ln -s /usr/local/tinygo/bin/tinygo .`

aykevl:
This should make installing TinyGo easier on Debian based systems by automatically creating a symlink in `/usr/local/bin`. For example, see #1103 and #725. It also makes the release process a little bit more robust and reproducible as the Debian package is built in CI instead of on my laptop.

fixes #725

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -205,12 +205,20 @@ commands:
           name: "Test TinyGo"
           command: make test
       - run:
+          name: "Install fpm"
+          command: |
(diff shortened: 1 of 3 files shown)


deadprogram:
An excellent idea! Now merging.

PR #941, opened 2020-03-05
Go module support

aykevl:
This is more of a mockup than anything serious. I want to see whether creating a new GOROOT per build is feasible. Right now creating such a GOROOT costs about 4ms on my system (after a warmed-up cache) so is not a significant overhead per build.

I'm not sure whether the current approach (with symlinks) is going to be viable on Windows.

The first commit adds the `go list` command, which might be enough to support TinyGo in IDEs such as VS Code.

PR #1108, opened 2020-05-13
Nintendo Switch Basic Support

racerxdl:
I added basic Nintendo Switch support to TinyGo.

It still relies on devKitPro and libnx to interact with Horizon OS (The Nintendo Switch Operating System), I hope that changes soon.

Still, that does work. At first I added the libnx calls here, but I found out its easier to make an external library to that. So I created this:

https://github.com/racerxdl/gonx

Without `gonx` the application can run inside nintendo switch, but can't use the display or anything until we implement full golang calls for the Switch IPCs.

I also created this: https://github.com/racerxdl/go-switch-examples

I will port each example from the switchbrew examples repo ( https://github.com/switchbrew/switch-examples/ ) and implement the nescessary calls on gonx as needed.

For compiling to nintendo switch its simple by using:

```bash
tinygo build -target nintendo-switch -o hello-world.elf
```
For running on real devices a `NRO` file needs to be generated, that can be done with elf2nro (which is inside devKitPro) or with [linkle](https://github.com/MegatonHammer/linkle):

```bash
linkle nro hello-world.elf hello-world.nro
```

Here is a video of working: https://twitter.com/lucasteske/status/1259650256897298432

And the vt52-example inside examples folder prints this:
image


trevor403:
Wow!

.gitignore
```diff
@@ -14,3 +14,4 @@ src/device/stm32/*.s
 vendor
 llvm-build
 llvm-project
+.idea
```
aykevl: This is unrelated to Switch support.
You can add it to your local .git/info/exclude or ~/.config/git/ignore.

builder/tools.go
```diff
@@ -2,10 +2,9 @@ package builder
 
 import (
 	"errors"
+	"github.com/tinygo-org/tinygo/goenv"
 	"os"
 	"os/exec"
-
-	"github.com/tinygo-org/tinygo/goenv"
```
aykevl: This is an unintended change?

src/machine/machine_generic.go
```diff
@@ -1,4 +1,5 @@
 // +build !baremetal
+// +build !nintendoswitch
```
aykevl: Why does this file not work? It looks like machine_nintendoswitch_stub.go is very similar.

src/runtime/print_nintendoswitch.go
```diff
@@ -0,0 +1,14 @@
...
+package runtime
+
+import (
+	"reflect"
+	"unsafe"
+)
+
+//go:nobounds
+func printstring(s string) {
+	sh := (*reflect.StringHeader)(unsafe.Pointer(&s))
```
aykevl: This custom `printstring` is unnecessary, all standard output should be handled by `putchar`.

src/os/file_nintendoswitch.go
```diff
@@ -0,0 +1,33 @@
+// +build nintendoswitch
+
+package os
+
+import "unsafe"
+
+// Read is unsupported on this system.
+func (f *File) Read(b []byte) (n int, err error) {
+	n = libc_read(int(f.fd), unsafe.Pointer(&b[0]), uint64(len(b)))
```
aykevl: Is there something about src/os/file_unix.go that doesn't work?
In general, it's better to use the syscall package for these calls and let the syscall package deal with platform specific issues (as some packages may call syscall directly). In this case, I think syscall_libc.go should work.

src/runtime/print_nintendoswitch.go
racerxdl: It can't be handled by putchar since the SVC requires a full string. Otherwise, this happens:
image

src/machine/machine_generic.go
racerxdl: It is very similar but `machine_generic` expects some assembly functions:

```go
//export __tinygo_gpio_configure
func gpioConfigure(pin Pin, config PinConfig)

//export __tinygo_gpio_set
func gpioSet(pin Pin, value bool)

//export __tinygo_gpio_get
func gpioGet(pin Pin) bool


//export __tinygo_spi_configure
func spiConfigure(bus uint8, sck Pin, mosi Pin, miso Pin)

//export __tinygo_spi_transfer
func spiTransfer(bus uint8, w uint8) uint8

//export __tinygo_adc_read
func adcRead(pin Pin) uint16


//export __tinygo_pwm_set
func pwmSet(pin Pin, value uint16)


//export __tinygo_i2c_configure
func i2cConfigure(bus uint8, scl Pin, sda Pin)

//export __tinygo_i2c_transfer
func i2cTransfer(bus uint8, w *byte, wlen int, r *byte, rlen int) int

//export __tinygo_uart_configure
func uartConfigure(bus uint8, tx Pin, rx Pin)

//export __tinygo_uart_read
func uartRead(bus uint8, buf *byte, bufLen int) int

//export __tinygo_uart_write
func uartWrite(bus uint8, buf *byte, bufLen int) int
```

On nintendoswitch_stub I just replace with empty functions or returning uninmplemented when needed.

src/os/file_nintendoswitch.go
racerxdl: hmm I will move it to syscall then

racerxdl: That's true. But Nintendo switch has read, seek, close, open calls. Should I implement then on `syscall_libc` mapping to libc native? That would break `picolibc` right? Maybe its better to create a `syscall_nintendoswitch`.

I will do that tonight.

.gitignore
racerxdl: Fixed

builder/tools.go
racerxdl: Fixed

racerxdl:
> I did a first pass over the PR.
> 
> There is a bit of code in builder/build.go that I'm not so happy about, I wonder how that can best be fixed? Ideally no environment variable is needed (apart from making sure the compiler is included in `$PATH`), but otherwise the best place to configure this would be compileopts, not builder.

The issue is that there still need to have the devkitpro env to get the libnx include path. It's always inside devkitpro folder, but you dont know where is installed. All devKitPro projects expect a DEVKITPRO environment to build.

src/machine/machine_generic.go
aykevl: There are no assembly functions. Their main purpose is to support https://play.tinygo.org (WebAssembly). As long as you don't actually call these, you can just leave the functions undefined.

src/runtime/print_nintendoswitch.go
aykevl: Ah I see. But even if you special-case `printstring`, other print functions will still print variables vertically.
What we did in WebAssembly was to buffer a string until a line terminator and then print the entire string using `console.log`. That is done in JavaScript, but you can do something similar in Go. Something like:

```go
var stdoutBuffer = make([]byte, 0, 80)

func putchar(c byte) {
    if c == '\n' {
        // print entire line...
        stdoutBuffer = stdoutBuffer[:0]
        return
    }
    stdoutBuffer = append(stdoutBuffer, c)
}
```
Also, what does the libnx `write` call do? I would assume it has the same problem?

src/os/file_nintendoswitch.go
aykevl: Yes, I think mapping the "syscalls" to the libc `open`, `close` etc calls is the best strategy.

> That would break `picolibc` right?

I'm not sure what you mean. You can only use one libc. So if you're using libnx (which seems to act as a libc), you shouldn't use picolibc. Picolibc is only used if you say so in the target JSON file.

src/machine/machine_generic.go
racerxdl: Hmm, I need to debug a bit more. But using that one gives me this:
```
Function return type does not match operand type of return inst!
  ret i64 %15, !dbg !24567
 i32error: verification error after IR construction
```

I will check what's the issue

src/runtime/print_nintendoswitch.go
racerxdl: the write call on libnx is actually the write from newlib. It just writes the data to the specified file descriptor. But the stdout/stderr/stdin needs to be pre-initialized to point somewhere, otherwise they're closed.

Since this print calls are internal to the engine (for example for panics), I use the System Call  OutputDebugString to output to Emulator Terminal.

I will try the solution of buffering the output.

src/os/file_nintendoswitch.go
racerxdl: I mean break the picolibc for non nintendo switch cases. I will check if picolibc implements open / read / close calls as well.

aykevl:
> The issue is that there still need to have the devkitpro env to get the libnx include path. It's always inside devkitpro folder, but you dont know where is installed. All devKitPro projects expect a DEVKITPRO environment to build.

I was playing around with it and you're right, even the bundled compilers haven't been configured to include libnx by default. Which makes things a whole lot more complicated.

---

I tried running examples/nintendoswitch in Yuzu but it doesn't work for me. That may also because of my Yuzu build (which I only barely managed to get to work). Does the example work in Yuzu?

src/os/file_nintendoswitch.go
aykevl: > I mean break the picolibc for non nintendo switch cases.

I'm not sure what you mean here. TinyGo only uses picolibc on baremetal systems, where the libc version of the syscall package is not used. In any case, this PR shouldn't modify non-Switch targets.

racerxdl: Ahh so its okay. I will implement the other methods and keep in the default libc syscall.

racerxdl:
The example does work on Yuzu (the screenshot is from yuzu). But I don't remember if you need the basic Nintendo Switch dump to make it work.

To avoid legal issues with Nintendo proprietary assets, to run commercial games on yuzu you need to dump keys from your console and some content from the NAND flash. That's a requirement for commercial games because they're signed and encrypted. The NRO files are neither, so I don't know what's the minimum requirements to run. I will check here (since I have everything) what's the minimum setup to get it working.

aykevl:
Can you maybe share a working .nro file that I can test? To determine which is the cause (the build or Yuzu).

racerxdl:
Sure, this one works here:

[vt52-demo.nro.zip](https://github.com/tinygo-org/tinygo/files/4624899/vt52-demo.nro.zip)
image

aykevl:
Well that version sadly doesn't run, so I'm afraid my Yuzu build still doesn't work.
One time it even crashed:

```
Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
dynarmic: POSIX SigHandler: Exception was not in registered code blocks (rip 0x007f596a4da8bb)
fish: “yuzu tmp/vt52-demo.nro” terminated by signal SIGSEGV (Address boundary error)
```

(fish is my shell)

racerxdl:
Hmm thats odd. As I remember it gives a message box saying what you need to run, not crashing. Maybe its some issue with wayland? Here I'm using Xorg instead, I never tested on wayland.

aykevl:
This is what it says on startup, is that a problem?

Screenshot from 2020-05-14 00-19-49

racerxdl:
That's the thing, for commercial games I'm 100% sure you need everything. Homebrew need only a subset. In about an hour I can confirm what's needed.

aykevl:
I have also managed to produce a test.nro that works in yuzu, so I have all the tools I need for local testing.

aykevl:
Played around a bit with avoiding libnx, but all I got was a panic in linkle:

```
thread 'main' panicked at 'Invalid module offset 3531603969', <::std::macros::panic macros>:5:6
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

src/runtime/print_nintendoswitch.go
racerxdl: image
Buffering works fine.

src/os/file_nintendoswitch.go
racerxdl: Done, everything on syscall_libc now.

src/machine/machine_generic.go
racerxdl: For some reason, after all the changes I did, this error dissapeared. LOL

src/syscall/syscall_libc.go
```diff
@@ -1,38 +1,58 @@
...
 
 package syscall
 
 import (
+	"errors"
 	"unsafe"
 )
 
+var (
+	notImplemented = errors.New("os: not implemented")
```
cornelk: should be "syscall: not implemented"

deadprogram:
@racerxdl can you please rebase this branch against `dev` to resolve the merge conflicts? Thank you.

racerxdl:
Done @deadprogram ! I also squashed all commits.

aykevl:
We recently changed compiler/compiler.go in #941 so I'm afraid this will need another conflict resolution.

builder/build.go
```diff
@@ -27,8 +28,25 @@ import (
 // The error value may be of type *MultiError. Callers will likely want to check
 // for this case and print such errors individually.
 func Build(pkgName, outpath string, config *compileopts.Config, action func(string) error) error {
+	var err error
+	var machine llvm.TargetMachine
+
+	isNintendoSwitch := false
```
aykevl: Just to be clear: I do not think code like this is mergeable. I understand you did it as a proof of concept but eventually this will need to be implemented in a proper way before merging.
My suggestion would be to add an extra field to the target JSON file that indicates the relocation model (static or PIC), while defaulting to static.

src/syscall/syscall.go
```diff
@@ -1,3 +1,5 @@
 package syscall
 
+type Signal int
+
```
aykevl: This seems to be an unrelated change?

racerxdl:
@aykevl just changed the way the PIC stuff reaches the compiler. Now it goes on TargetSpec and comes from json spec. I modified the original `NewTargetMachine` to read the RelocationModel from TargetSpec and select the appropriate one (or static if invalid defined)

deadprogram:
Hello, everyone. I would love to help get this PR merged before the next release. Seems like there are again (or still?) some merge conflicts to resolve.

After that, what else will be needed to get it into a mergeable state? Thank you.

racerxdl:
Done @deadprogram. Not sure what's next.

builder/build.go
```diff
@@ -27,11 +28,16 @@ import (
 // The error value may be of type *MultiError. Callers will likely want to check
 // for this case and print such errors individually.
 func Build(pkgName, outpath string, config *compileopts.Config, action func(string) error) error {
+	var err error
+	var machine llvm.TargetMachine
```
aykevl, line 32: It looks like you didn't undo these changes, which are now just noise.

builder/build.go
```diff
@@ -156,6 +163,29 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(stri
 			ldflags = append(ldflags, librt)
 		}
 
+		root := goenv.Get("TINYGOROOT")
+		dkp := os.Getenv("DEVKITPRO")
+		// If Nintendo Switch, add libnx
+		if config.Target.Linker == "devKitPro" {
```
aykevl, line 169: I still don't really agree with adding code here that is specific for a particular board. I was under the impression that the special linker and libnx wasn't necessary for initial Switch support?

compileopts/target.go
```diff
@@ -266,16 +271,17 @@ func defaultTarget(goos, goarch, triple string) (*TargetSpec, error) {
...
+		GOOS:            goos,
+		GOARCH:          goarch,
+		BuildTags:       []string{goos, goarch},
+		Compiler:        "clang",
+		Linker:          "cc",
+		CFlags:          []string{"--target=" + triple},
+		GDB:             "gdb",
+		PortReset:       "false",
+		FlashMethod:     "native",
+		RelocationModel: "static",
```
aykevl, line 284: This is unnecessary: it defaults to `static` already.

compiler/compiler.go
```diff
@@ -109,7 +110,18 @@ func NewTargetMachine(config *compileopts.Config) (llvm.TargetMachine, error) {
 		codeModel = llvm.CodeModelLarge
 	}
 
-	machine := target.CreateTargetMachine(config.Triple(), config.CPU(), features, llvm.CodeGenLevelDefault, llvm.RelocStatic, codeModel)
+	switch config.Target.RelocationModel {
```
aykevl, line 113:
```suggestion
	switch config.RelocationModel() {
```

compiler/compiler.go
```diff
@@ -109,7 +110,18 @@ func NewTargetMachine(config *compileopts.Config) (llvm.TargetMachine, error) {
...
+	case "static":
+		relocationModel = llvm.RelocStatic
+	case "pic":
+		relocationModel = llvm.RelocPIC
+	case "dynamicnopic":
+		relocationModel = llvm.RelocDynamicNoPic
+	default:
+		relocationModel = llvm.RelocStatic
```
aykevl, line 121: This should return an error instead of silently defaulting to static with an unknown relocation model.

src/examples/nintendoswitch/vt52-example.go
```diff
@@ -0,0 +1,71 @@
+// VT52 Demo example
+// Same as https://github.com/switchbrew/switch-examples/tree/master/graphics/printing/vt52-demo
+package main
+
+import (
+	"fmt"
+	"github.com/racerxdl/gonx/nx"
```
aykevl, line 7: I'm not sure we want to depend on an external package for a test. Is there something interesting you can do with what is in TinyGo already?

src/runtime/runtime_nintendoswitch.go
```diff
@@ -0,0 +1,75 @@
...
+func main() int {
+	preinit()
+	run()
+
+	return exit(0) // Call libc_exit to cleanup libnx
+}
+
+// sleepTicks argument are actually in microseconds
+func sleepTicks(d timeUnit) {
+	usleep(uint(d))
```
aykevl, line 22: If I understand correctly:

* `armGetSystemTick` returns the time in 19200000Hz units.
* `usleep` (as the name indicates) sleeps in microseconds.

The `timeUnit` type assumes there is just one unit, not two. In fact, in the implementation here values of `timeUnit` can be either in nanoseconds or in the Switch-specific time unit (with intervals of around 52ns). This is not consistent with other targets, where it means just one thing.

I would suggest switching to nanoseconds everywhere for consistency. The conversion to/from microseconds is obvious. The conversion to/from the system tick timer is less obvious, and assuming 52ns tick interval is slightly incorrect. You can use this instead, which is exact:

```go
func ticks() timeUnit {
	// the following formula is equivalent to this but with less risk of overflowing:
	//     getArmSystemTimeNs() * 1e9 / 19200000
	return getArmSystemTimeNs() * 625 / 12
}
```

src/runtime/runtime_nintendoswitch_heap.go
```diff
@@ -0,0 +1,28 @@
...
+// +build gc.conservative gc.leaking
+
+package runtime
+
+import "unsafe"
+
+const heapSize = 16 * 1024 * 1024
+
+//go:extern _stack_top
+var stackTopSymbol [0]byte
```
aykevl, line 12: Where is this defined?

src/runtime/runtime_nintendoswitch_svc.go
```diff
@@ -0,0 +1,17 @@
+// +build nintendoswitch
```
aykevl, line 1: It seems to me that this could be in the runtime_nintendoswitch.go file, or is there a reason to keep it separate?

src/syscall/syscall_baremetal.go
```diff
@@ -12,6 +12,7 @@ package syscall
 
 // A Signal is a number describing a process signal.
 // It implements the os.Signal interface.
+
```
aykevl, line 15: This looks unintentional?

src/syscall/syscall_libc.go
```diff
@@ -1,46 +1,66 @@
...
 	"unsafe"
 )
 
+var (
+	notImplemented = errors.New("syscall: not implemented")
+)
+
 func Close(fd int) (err error) {
-	return ENOSYS // TODO
+	errN := libc_close(int32(fd))
```
aykevl, line 15: What are all these syscall changes for? Are they necessary for initial support?
It seems better to me to do it in a separate PR and do it properly (without all the TODOs here).

targets/nintendo-switch.json
```diff
@@ -0,0 +1,35 @@
+{
+  "llvm-target": "aarch64",
+  "build-tags": ["nintendoswitch", "arm64"],
+  "goos": "linux",
+  "goarch": "arm64",
+  "compiler": "aarch64-none-elf-gcc",
+  "linker": "devKitPro",
```
aykevl, line 7: Wasn't it possible to link using `lld`?

aykevl: Also, if you can use `aarch64-none-elf-gcc`, would `clang` (with the right options) work? The less dependencies, the better.

racerxdl: It wasn't. LLD can't link stuff pre-compiled with GCC with custom linker script :( - It gave several errors. Also it does need DevKitPro because of the libc required for all nintendo switch deps (the DevKitPro is a modified picolibc).

We might eventually want to port devKitPro picolibc to tinygo and replace the current one, because the picolibc from devKitPro is more generic than the one we have on tinygo today. I checked if it was easy to port it, but there are some things that require changes on other tinygo targets, so I think its better a separated PR for that eventually.

src/runtime/runtime_nintendoswitch_heap.go
racerxdl: That's from devKitPro libc. I will put a comment there.

src/syscall/syscall_libc.go
racerxdl: Yes, for nintendo switch that is actually needed (it does have file I/O). Actually I'm not sure why they weren't implemented like that since the picolibc included on tinygo has that libc calls as stubs.

builder/build.go
racerxdl: It is needed. The support is complete useless without being able to link libnx. I tried libtransistor (which is made using LLVM) but its basically abandoned and has some issues.

targets/nintendo-switch.json
aykevl: Okay, now I understand the situation better.

The `linker` field here really refers to an actual command. I think instead of repurposing the `linker` field, you should do something like this:

```json
    "libc": "libnx",
```

That is actually the field intended for changing the libc, which is the main reason all these custom build steps are needed. You can then check for libnx (similar to picolibc) and if it is used, add the correct paths by querying the `DEVKITPRO` environment variable. Not how I would hope it would work, but I think that is the cleanest way to do it.

aykevl: On second thought, you need to run a command in the `$DEVKITPRO` directory. Which is not known statically. This requires some logic to determine the correct path.

This is somewhat of a mismatch between how I intended TinyGo to work and how DevKitPro works. I'm not sure how to fix that exactly. Maybe as it is right now.

deadprogram:
> If possible, please add a smoke test for this PR. It will make it obvious when there is a compile-time breaking change.

Yes, please, @racerxdl is this possible? Ideally something with no external dependencies. Does not have to do much, just shows that the core compiler itself is working.

builder/build.go
deadprogram: Is there something needed still for this comment?

compileopts/target.go
deadprogram: Is action required on this comment?

compiler/compiler.go
deadprogram: Looks like something still required here?

src/examples/nintendoswitch/vt52-example.go
deadprogram: Same question from me, even the most minimal example is better. Then from the docs page on the tinygo.org site we can point to other resources such as the gonx package.

src/runtime/runtime_nintendoswitch.go
deadprogram: I think perhaps #1127 is the model to look at now, of how to do this.

targets/nintendo-switch.json
deadprogram: > Maybe as it is right now.

@aykevl do you mean as this PR has already implemented?

racerxdl:
Hi @deadprogram / @aykevl - sorry I didn't had much time to look these days. This weekend I will try to get a look in all comments and do the nescessary fixes / comments.

About the smoke test, yes I can put a very small program just to test that everything is compiling. The thing is that the only change for Nintendo Switch is basically the devKitpro call for libnx linking, otherwise is just a normal arm64. So If I remove the devKitPro dependency for a smoke test, it will just test the ARM64 compiler part. Is that what's intended?

The last time I was working on that, I tried to make the libnx linkage externally to the tinygo (by doing this https://github.com/racerxdl/gonx/blob/dev/base.go#L6-L7 ) but that does seen to trigger a bug in CGO ( #1148 ) which @aykevl wasn't able to simulate. (I think because he is importing from a local folder and not a external library).

If I manage to do a external linkage, I can put a very small CRT for nintendo switch inside tinygo so it works as a application inside the nintendo switch. Also I think in that case, the lld can link it (just not link when using libnx).

deadprogram:
Thanks @racerxdl that sounds great!

racerxdl:
Hi @deadprogram / @aykevl while I work on the external integration of libnx for this branch, I did another setup that does not require devKitPro or anything, but can only print to Nintendo Switch emulator debug console. 

https://github.com/tinygo-org/tinygo/pull/1226

But this will serve as a base for external linkage of libnx. I will change this PR (#1108) to draft and then when its ready again I switch to a PR again.

targets/nintendo-switch.json
aykevl: Yes. I don't like it but I don't see a better way, so maybe just do it as the PR is now. It's unfortunate that devkitPro assumes the use of global variables, which isn't very compatible with TinyGo right now.

PR #1109, opened 2020-05-15
AVR PinInputPullup

aykevl:
This PR contains two somewhat related (but distinct) commits:

1. A refactor of AVR GPIO pins. It reduces duplication and at with that, also reduces the amount of new code needed when adding support for a new AVR chip. Additionally, I discovered a bug in the ATmega2560 `Pin.Get()` function which is fixed in the unified `Pin.Get()` function.
2. Added support for `PinInputPullup`. The motivation for the refactor was to make this simpler to implement, and as a result this only needed a code change in one place.

`PinInputPullup` will make AVR pin interrupt support much easier to use. I have code for that that I just need to clean up, but I thought it would be a good idea to make this PR separately first.

I have tested this on an Arduino Uno and a DigiSpark. Note that for the DigiSpark, you can't use all pins. For example, P3 already has a pullup to 3.3V for USB and will cause weird behavior if you try to use it as a regular pullup. I was very confused at first why pullups didn't seem to work on the DigiSpark while the assembly looked fine.

### DIFF
--- a/src/examples/button/button.go
+++ b/src/examples/button/button.go
@@ -12,7 +12,7 @@ const (
 
 func main() {
 	led.Configure(machine.PinConfig{Mode: machine.PinOutput})
-	button.Configure(machine.PinConfig{Mode: machine.PinInput})
+	button.Configure(machine.PinConfig{Mode: machine.PinInputPullup})
 
 	for {
 		if button.Get() {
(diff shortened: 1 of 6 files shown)


deadprogram:
Good work @aykevl thanks for this. Now merging.

PR #1112, opened 2020-05-15
main: add pause flag to be used with tinygo flash

deadprogram:
This PR adds a `-pause` flag to be used with the `tinygo flash` command. It allows for adjusting the pause time after a 1200 baud reset, to allow for slower machines to give enough time for the connected MCU to finish coming up in the bootloader and be available for flashing.

### DIFF
--- a/main.go
+++ b/main.go
@@ -156,7 +156,7 @@ func Test(pkgName string, options *compileopts.Options) error {
 }
 
 // Flash builds and flashes the built binary to the given serial port.
-func Flash(pkgName, port string, options *compileopts.Options) error {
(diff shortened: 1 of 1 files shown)


aykevl:
As an alternative: would it be possible to recheck every second (or so) until the chip becomes available? That would avoid the compiler flag and automatically pick the right "pause" duration. There would probably need to be an upper limit (10s?) after which the device is declared "not found".

deadprogram:
Agreed. Closing in favor of new PR.

deadprogram:
I replaced this PR with https://github.com/tinygo-org/tinygo/pull/1129

PR #298, opened 2019-04-23
WIP compiler: implement atomic operations in package sync/atomic

aykevl:
Implement functions in package sync/atomic using LLVM intrinsics.
Does not work yet:

 * the compiler emits unresolved symbols like `__sync_fetch_and_add_8` on Cortex-M, we need to implement something for that
 * atomic operations do not seem to be implemented in the WebAssembly backend: it crashes with a nasty `LIBCLANG FATAL ERROR: Cannot select` error
 * some functions need to be added to go-llvm.

I am making the PR anyway for future reference.

PR #1113, opened 2020-05-16
compiler: add support for atomic operations

aykevl:
This also implements DisableInterrupts/EnableInterrupts for RISC-V, as those operations were needed to implement a few libcalls.

This PR replaces the out-of-date and incomplete PR #298. Changes since then:

* Added implementations of libcalls for Cortex-M and RISC-V.
* Added support for the `sync/atomic.Value` type.
* Removed src/runtime/atomic.go, which had a few (unsafe!) implementations of atomic operations.


deadprogram:
@jaddr2line any feedback before we merge?

PR #1121, opened 2020-05-19
machine: refactor PWM support

aykevl:
This commit lays the groundwork and refactors the SAM D21 implementation.

This is very much WIP, as the API hasn't been agreed upon yet. It is really just a proof of concept to make sure the API works in the real world. The API implemented here gives a lot of flexibility and control over PWM while keeping device specific details to a minimum. It also tries to be as simple to use as possible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -341,8 +341,6 @@ smoketest:
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=feather-m4          examples/pwm
 	@$(MD5SUM) test.hex
-	$(TINYGO) build -size short -o test.hex -target=pyportal            examples/pwm
-	@$(MD5SUM) test.hex
 ifneq ($(STM32), 0)
 	$(TINYGO) build -size short -o test.hex -target=bluepill            examples/blinky1
 	@$(MD5SUM) test.hex
--- a/src/examples/pwm/arduino.go
+++ b/src/examples/pwm/arduino.go
@@ -1,0 +1,12 @@
+// +build arduino
+
+package main
+
+import "machine"
+
+var (
+	// Configuration on an Arduino Uno.
+	pwm  = machine.Timer2
+	pinA = machine.PB3 // pin 11 on the Uno
+	pinB = machine.PD3 // pin 3 on the Uno
+)
--- a/src/examples/pwm/feather-m4.go
+++ b/src/examples/pwm/feather-m4.go
@@ -1,0 +1,11 @@
+// +build feather_m4
+
+package main
+
+import "machine"
+
+var (
+	pwm  = machine.TCC0
+	pinA = machine.D12
+	pinB = machine.D13
+)
--- a/src/examples/pwm/itsybitsy-m0.go
+++ b/src/examples/pwm/itsybitsy-m0.go
@@ -1,0 +1,11 @@
+// +build itsybitsy_m0
+
+package main
+
+import "machine"
+
+var (
+	pwm  = machine.TCC0
+	pinA = machine.D3
+	pinB = machine.D4
+)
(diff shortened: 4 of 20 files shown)


deadprogram:
What is the status with this PR now?

aykevl:
I now have a PWM abstraction for four chips: atsamd21, atsamd51, nrf52xxx, and atmega328p. The abstraction works with a WIP servo and tone driver, proving that it can be used for purposes that require tight control over frequency and duty cycle.

Next up: update the documentation so that it is usable and possibly also add a method or something so that it remains possible to easily drive LEDs using PWM without worrying about PWM instances.

aykevl:
I updated the PR again, I think it is now more or less ready.

I have also updated the documentation. I'm not entirely ready yet, but here is the near-final result: https://github.com/tinygo-org/tinygo-site/blob/pwm/content/microcontrollers/machine/pwm.md

For individual boards, I've started adding documentation on pin to PWM mapping: https://github.com/tinygo-org/tinygo-site/blob/pwm/content/microcontrollers/circuit-playground-express.md#pins
Again, feedback welcome. I'm now just listing the names of the PWM peripherals as they are defined in the chip. I've considered renaming them to `PWM0` etc. but some chips have different kinds of PWMs for different purposes (for example, TC0 vs TCC0 on a samd51 - both are timer peripherals capable of PWM).

To be implemented at a later date:

* Phase correct PWM. I'm sure these have a use case but I haven't come across one yet.
* One shot PWM, to produce a single precisely timed pulse. Useful for a hub75 driver.
* Interrupts at match and overflow. Again useful for a hub75 driver.
* Features to be able to use these timers as actual times, with the ability to get and set the time.

deadprogram:
Been studying this PR for a while and I think the latest version is pretty great. What would be the next steps to help it land?

deadprogram:
@conejoninja @ardnew @sago35 or anyone else have any opinions?

aykevl:
I didn't really look at it recently but it should be ready now. What is missing is some work on the documentation, but that can perhaps happen after this PR lands.

Of course, this is a breaking change so might need a few more eyes for review.

conejoninja:
I don't fully understand the details, but looks good to me. documentation is clear enough and it would be great to finally have a servo driver, so this is a first step

ardnew:
> * Features to be able to use these timers as actual times, with the ability to get and set the time.

I'm not real certain what you mean by this. Do you mean a common API using various chip's timer peripherals (e.g., TC/TCC on SAMD21) as general-purpose timers (with common functionality such as interrupt on overflow, etc.) and/or system timers such as SysTick and Sleep? If so, that seems like an entirely separate effort from this PWM support

aykevl:
> Do you mean a common API using various chip's timer peripherals (e.g., TC/TCC on SAMD21) as general-purpose timers (with common functionality such as interrupt on overflow, etc.)

Yes, this. PWM and timers may seem like different things but in practice they aren't. Most of the time a single peripheral can be used as both a PWM and a timer. In fact, the core of a PWM _is_ a timer. The nrf chips do have a dedicated PWM (that would be harder to use as a timer) and have a dedicated RTC (that can't easily be used as a PWM) but most other chips I've looked at (avr, sam, stm32) don't make a clear distinction.

For most chips, I do not intend to use these PWM peripherals as system times however maybe it would be a good idea on AVR: adding an API to use one of the timer/PWM peripherals as a system timer (because AVRs usually don't have a system timer).

sago35:
It looks very good.

Nerzal:
So after merge conflicts are solved, this PR can finally be merged? :)

deadprogram:
@aykevl reminder that this PR needs a merge conflict resolved. Thanks.

aykevl:
Rebased.

deadprogram:
Looks like just one last merge conflict to be resolved.

aykevl:
Rebased again. It was conflicting with #1718.

deadprogram:
By popular demand, now merging this PR.

:infinity:

Nerzal:
thank u <3

PR #1125, opened 2020-05-20
cgo: Add LDFlags support and float types

racerxdl:
I added LDFlags support to CGO. It just parses the `#cgo LDFLAGS` statements and pass to the linker.

WIth this, I'm able to compile that:

```go
package main

/*
#cgo LDFLAGS: -lm

#include <math.h>
#include <stdio.h>
*/
import "C"

import "fmt"

func main() {
    f := float64(0.4)
    f2 := float64(0.5)

    res := C.fmax(f, f2)
    fmt.Println("FLOAT: ", float64(res))
}
```

Which uses libm function `fmax`

This should also help in #1108 for [gonx](https://github.com/racerxdl/gonx) link the libnx externally and not internally on tinygo.

### DIFF
--- a/cgo/cgo.go
+++ b/cgo/cgo.go
@@ -41,6 +41,7 @@ type cgoPackage struct {
 	elaboratedTypes map[string]*elaboratedTypeInfo
 	enums           map[string]enumInfo
 	anonStructNum   int
+	ldflags         []string
 }
 
 // constantInfo stores some information about a CGo constant found by libclang
@@ -156,7 +157,7 @@ typedef unsigned long long  _Cgo_ulonglong;
 // newly created *ast.File that should be added to the list of to-be-parsed
 // files. If there is one or more error, it returns these in the []error slice
 // but still modifies the AST.
-func Process(files []*ast.File, dir string, fset *token.FileSet, cflags []string) (*ast.File, []error) {
+func Process(files []*ast.File, dir string, fset *token.FileSet, cflags []string) (*ast.File, []string, []error) {
 	p := &cgoPackage{
 		dir:             dir,
 		fset:            fset,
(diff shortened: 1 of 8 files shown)


cgo/cgo.go
```diff
@@ -115,6 +116,8 @@ var cgoAliases = map[string]string{
 	"C.uint32_t":  "uint32",
 	"C.uint64_t":  "uint64",
 	"C.uintptr_t": "uintptr",
+	"C.float":     "float32",
+	"C.double":    "float64",
```
aykevl: I wasn't aware that these were necessary. Is there code that uses `C.float` and `C.double` instead of simply `float32` and `float64`? So far, using floats in testdata/cgo seems to have worked well so far, so I'm not sure what this fixes exactly. Note that a C `float` is already an alias for `float32` and C `double` is already an alias for `float64`.

In any case, these should really be done in a separate PR or a cleanly separated commit, as this is unrelated to `LDFLAGS` support.

aykevl (commented):
The `LDFLAGS` parts look fine by me. But see the inline comment below.

cgo/cgo.go
racerxdl: Well it does work fine for tinygo because the types can be mapped directly. On go you need to convert the types so usually the cgo codes use that C.float or C.double. Its not needed for tinygo but it increases the compatibility with oficial golang toolchain.

I will separate it in another pull request.

aykevl: Thanks! Makes sense.

You can run `go test ./cgo` to test the changes locally. It may also be a good idea to add a test to testdata/cgo, to check that `C.float` etc works.

aykevl (approved):

aykevl:
I changed the base branch. Note that we're working from the dev branch during the development, master points to the latest release.

racerxdl:
Ahh great to know. I will use `dev` branch now on. Thanks!

PR #1127, opened 2020-05-21
runtime: refactor time handling

aykevl:
This commit refactors both determining the current time and sleeping for
a given time. It also improves precision for many chips.

  * The nrf chips had a long-standing TODO comment about a slightly
    inaccurate clock. This should now be fixed.
  * The SAM D2x/D5x chips may have a slightly more accurate clock,
    although probably within the error margin of the RTC. Also, by
    working with RTC ticks and converting in the least number of places,
    code size is often slightly reduced (usually just a few bytes, up to
    around 1kB in some cases).
  * I believe the HiFive1 rev B timer was slightly wrong (32768Hz vs
    30517.6Hz). Because the datasheet says the clock runs at 32768Hz,
    I've used the same conversion code here as in the nrf and sam cases.
  * I couldn't test both stm32 timers, so I kept them as they currently
    are. It may be possible to make them more efficient by using the
    native tick frequency instead of using microseconds everywhere.

This commit also replaces the incorrectly-named `tickMicros`.

Note that this may increase code size on nrf chips, probably because a simple shift (the power-of-two divide) got replaced with a 64-bit divide. But I think it's needed, as it fixes a real issue.

As indicated, I had trouble flashing my Blue Pill board, so if anyone can test whether it didn't break that would be great. The change is actually very small so I don't expect any breakage.

@mickey-happygolucky can you test the STM32F4 Discovery maybe? You can test examples/blinky1.

### DIFF
--- a/src/runtime/runtime.go
+++ b/src/runtime/runtime.go
@@ -61,7 +61,7 @@ func memequal(x, y unsafe.Pointer, n uintptr) bool {
 }
 
 func nanotime() int64 {
-	return int64(ticks()) * tickMicros
+	return ticksToNanoseconds(ticks())
 }
 
 // timeOffset is how long the monotonic clock started after the Unix epoch. It
--- a/src/runtime/runtime_arm7tdmi.go
+++ b/src/runtime/runtime_arm7tdmi.go
@@ -9,8 +9,6 @@ import (
 
 type timeUnit int64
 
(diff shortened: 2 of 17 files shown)


deadprogram:
Tested on HCI and also ran test manually on my Bluepill. A much needed refactoring! Now merging.

PR #1129, opened 2020-05-23
main: use auto-retry (up to 10 times) to locate MSD for UF2 and HEX

deadprogram:
This PR causes the `tinygo flash` command to auto-retry (up to 10 seconds) to locate MSD for UF2 and HEX files. It replaces #1112

### DIFF
--- a/main.go
+++ b/main.go
@@ -491,6 +491,8 @@ func touchSerialPortAt1200bps(port string) (err error) {
 	return fmt.Errorf("opening port: %s", err)
 }
 
+const maxMSDRetries = 10
+
 func flashUF2UsingMSD(volume, tmppath string) error {
 	// find standard UF2 info path
 	var infoPath string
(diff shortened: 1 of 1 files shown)


aykevl (approved):
I think this PR is fine, and can be merged as-is. However, there are a few ways in which it could be improved:

* You could perhaps reduce the delay from 1 second to something lower (100ms? 500ms?). Reading that directory is not a very expensive operation and a delay of 1s will introduce an average waiting time of 0.5s, which quickly adds up over many edit-build-test cycles.
* Similarly, now that it will retry if the board is not yet available, the hardcoded 3 second delay in the `Flash` function may be reduced. Again, this can reduce flash time a bit as the delay will usually be less than that.
* There is some duplicate code, which is unfortunate.

deadprogram:
Made some changes and refactored to reduce duplication, @aykevl  please take another look.

deadprogram:
Ready for another review, please.

aykevl:
Thanks, merged.
I should note that the last commit changes the timeout from 10s to 5s, not sure whether that was intentional.

deadprogram:
It was intentional, as I was also trying to speed up the process based on your previous comment.

PR #1136, opened 2020-05-26
Gameboy Advance multiboot target

trevor403:
There is a feature on the GBA called multiboot. The bios uses the EXT1 port on the GBA to initiate a multiboot protocol over SIO (a custom protocol that is similar to SPI) to load a game directly into EWRAM and boot it by executing a branch instruction.

This PR adds the `gameboy-advance-mb` target which changes the linker script to support EWRAM loading.

I also updated the crt0 assembly to include some missing header info that is required on GBA for multiboot. It is important to include the multiboot header for all GBA roms as the bios will clobber some bytes while loading the rom. 

NOTE: this code is not functional likely due to an issue in the garbage collector which treats _sbss as part of heap. 
I am able to step through the main loop but as soon as the gc get's called, the program starts executing nop memory.

### DIFF
--- a/targets/gameboy-advance-mb.ld
+++ b/targets/gameboy-advance-mb.ld
@@ -1,0 +1,84 @@
+OUTPUT_ARCH(arm)
+ENTRY(_start)
+
+/* Note: iwram is reduced by 96 bytes because the last part of that RAM
+ * (starting at 0x03007FA0) is used for interrupt handling.
+ */
+MEMORY {
+    ewram   : ORIGIN = 0x02000000, LENGTH = 256K   /* on-board work RAM (2 wait states) */
+    iwram   : ORIGIN = 0x03000000, LENGTH = 32K-96 /* in-chip work RAM (faster) */
+}
+
+__stack_size_irq = 1K;
+__stack_size_usr = 2K;
+
+SECTIONS
+{
+    .text :
+    {
+        KEEP (*(.init))
+        *(.text)
+        . = ALIGN(4);
+    } >ewram
+
+    .rodata :
+    {
+        . = ALIGN(4);
+        *(.rodata)
+        *(.rodata*)
+        . = ALIGN(4);
+    } >ewram
+
+    /* Put the stack at the bottom of RAM, so that the application will
+     * crash on stack overflow instead of silently corrupting memory.
+     * See: http://blog.japaric.io/stack-overflow-protection/ */
+    .stack (NOLOAD) :
+    {
+        . = ALIGN(4);
+        _stack_top_irq = .;
+        . += __stack_size_irq;
+        _stack_top = .;
+        . += __stack_size_usr;
+    } >iwram
+
+    /* Start address (in flash) of .data, used by startup code. */
+    _sidata = LOADADDR(.data);
+
+    /* Globals with initial value */
+    .data :
+    {
+        . = ALIGN(4);
+        _sdata = .;        /* used by startup code */
+        *(.data)
+        *(.data*)
+        *(.iwram .iwram.*)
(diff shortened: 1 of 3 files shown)


trevor403:
I think it might be a good idea to rename the `gameboy-advance` target to `gameboy-advance-rom` or `gameboy-advance-cart` to we can use `gameboy-advance` as an inherited target.

niaow:
I am not quite sure that your heap start and end are set correctly. Could that be why you are having issues?

trevor403:
@jaddr2line that is almost likely why I am having issues... It looks like the gc is trying to zero `0x20000c0 <_rom_header_end>` which is odd.

I can mess around with breakpoints for a couple hours tonight and try to figure out why this is happening. Unfortunately there are no locals in the elf file so it may be harder said than done.

trevor403:
I spent all night in GDB but I was not able to track down the corruption... my `heapStart` is set just fine, the address is far past the end of the program.

targets/gameboy-advance-mb.ld
```diff
@@ -0,0 +1,84 @@
...
+    /DISCARD/ :
+    {
+        *(.ARM.exidx)      /* causes 'no memory region specified' error in lld */
+        *(.ARM.exidx.*)    /* causes spurious 'undefined reference' errors */
+    }
+}
+
+/* For the memory allocator. */
+_heap_start = 0x2020000;
+_heap_end = ORIGIN(ewram) + LENGTH(ewram);
```
aykevl, line 82: Based on this, you're using the area from `0x2020000` to `0x2040000` as heap. Is that correct?

targets/gameboy-advance-mb.ld
```diff
@@ -0,0 +1,84 @@
...
+    .data :
+    {
+        . = ALIGN(4);
+        _sdata = .;        /* used by startup code */
+        *(.data)
+        *(.data*)
+        *(.iwram .iwram.*)
+        . = ALIGN(4);
+        _edata = .;        /* used by startup code */
+    } >iwram AT>ewram
```
aykevl, line 57: This confuses me. Normally you would copy the `.data` section from flash to RAM, and the `>iwram AT>ewram` part tells the linker to put the area at ewram but relocate it as if it was at iwram. In this case, everything resides in RAM so there is no need to copy the `.data` section. Therefore, I would suggest to remove `_sidata` as it is no longer necessary (if it is, it is still copying the `.data` section even though it shouldn't) and replacing `>iwram AT>ewram` with just `>iwram`.

targets/gameboy-advance-mb.ld
```diff
@@ -0,0 +1,84 @@
...
+        . = ALIGN(4);
+        _sbss = .;         /* used by startup code */
+        *(.bss)
+        *(.bss*)
+        *(COMMON)
+        . = ALIGN(4);
+        _ebss = .;         /* used by startup code */
+    } >iwram
+
+    /* TODO: find ewram ABSOLUTE for _heap_start */
```
aykevl, line 71: You could let the heap start after iwram, in which case you could set `_heap_start` to `_ebss`.

aykevl (commented):
Some comments inline, that may help debug the issue.
How exactly is the code and data loaded? Is everything directly loaded into RAM at the addresses specified in the ELF file or is there some sort of translation going on?

trevor403:
Thanks for the code review @aykevl I'll take a look after work and play with the linker script

deadprogram:
Due to being pretty far out of date, I am now closing this PR.

PLease see https://github.com/tinygo-org/tinygba and friends for where the TinyGo GBA dev is happening now. Thanks @trevor403

PR #1137, opened 2020-05-26
compiler: fix a few crashes due to named types

aykevl:
There were a few cases left where a named type would cause a crash in the compiler. While going through enough code would have found them eventually, I specifically looked for the `Type().(` pattern: a `Type()` call that is then used in a type assert. Most of those were indeed bugs, although for some I couldn't come up with a reproducer so I left them as-is.

### DIFF
--- a/compiler/channel.go
+++ b/compiler/channel.go
@@ -12,7 +12,7 @@ import (
 )
 
 func (b *builder) createMakeChan(expr *ssa.MakeChan) llvm.Value {
-	elementSize := b.targetData.TypeAllocSize(b.getLLVMType(expr.Type().(*types.Chan).Elem()))
+	elementSize := b.targetData.TypeAllocSize(b.getLLVMType(expr.Type().Underlying().(*types.Chan).Elem()))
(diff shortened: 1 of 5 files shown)


deadprogram:
I think perhaps TinyHCI might have found an actual issue on AVR with this change.

> analogReadHalfVoltage = fail
> expected: 'val <= 65535/2+256 && val >= 65535/2-256'
> actual: �������������������v

that is not the uint16 value that we have been previously getting, or am I missing something else here?

aykevl:
I have seen that error before in other PRs, it might be a flaky test. I do not think it's related to this PR as (in theory) it should not affect any code that worked before, only allow more code to be compiled.

aykevl:
Here is a similar error: https://github.com/tinygo-org/tinygo/runs/712762185 (with code that previously passed: https://github.com/tinygo-org/tinygo/runs/702316486).

deadprogram:
I think it was a short in the wiring to V, that caused the value to overflow from too many max values being added together. I changed the board wiring slightly to give a little more space to the resistors, and now it is working as expected.

deadprogram:
Hence, merging. Thanks!

PR #1138, opened 2020-05-27
Interface nil check

aykevl:
I ran into an issue where I did a method call on a nil interface and it
resulted in a HardFault. Luckily I quickly realized what was going on so
I could fix it, but I think undefined behavior is definitely the wrong
behavior in this case. This commit therefore changes such calls to cause
a nil panic instead of introducing undefined behavior.

This does have a code size impact. It's relatively minor, much lower
than I expected. When comparing the before and after of the drivers
smoke tests (probably the most representative sample available), I found
that most did not change at all and those that did change, normally not
more than 100 bytes (16 or 32 byte changes are typical).

Right now the pattern is the following:

    switch typecode {
    case 1:
        call method 1
    case 2:
        call method 2
    default:
        nil panic
    }

I also tried the following (in the hope that it would be easier to
optimize), but it didn't really result in a code size reduction:

    switch typecode {
    case 1:
        call method 1
    case 2:
        call method 2
    case 0:
        nil panic
    default:
        unreachable
    }

Some code got smaller, while other code (the majority) got bigger. Maybe
this can be improved once [range](https://llvm.org/docs/LangRef.html#range-metadata) is finally [allowed](https://github.com/rust-lang/rust/issues/50156) on function
parameters, but it'll probably take a while before that is implemented.

---

The first commit refactors some code that makes the second commit easier to implement. It also has a (very small) code size impact.

### DIFF
--- a/transform/interface-lowering.go
+++ b/transform/interface-lowering.go
@@ -292,58 +292,40 @@ func (p *lowerInterfacesPass) run() error {
 
 		methodSet := use.Operand(1).Operand(0) // global variable
 		itf := p.interfaces[methodSet.Name()]
-		if len(itf.types) == 0 {
-			// This method call is impossible: no type implements this
-			// interface. In fact, the previous type assert that got this
-			// interface value should already have returned false.
-			// Replace the function pointer with undef (which will then be
-			// called), indicating to the optimizer this code is unreachable.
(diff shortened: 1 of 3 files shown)


deadprogram:
Very good check to have, certainly worth the very few bytes it adds. Now merging, thanks @aykevl

PR #1111, opened 2020-05-15
[WIP] Interrupts pins samd51

sago35:
#1094

I wrote the code for SAMD51J19A.
I checked with feather-m4 and it seems to be working fine.
I haven't confirmed this with the SAMD51J20A yet.

https://github.com/sago35/tinygo/tree/interrupts-pins-samd51

I'm struggling with how to write the following parts.
Is there a better way to do it?

1. `sam.EIC.CTRLA.Set(0)` or `sam.EIC.CTRLA.ClearBits(0x02)`
    * https://github.com/sago35/tinygo/blob/interrupts-pins-samd51/src/machine/machine_atsamd51.go#L352
2. Is it better to define it outside of SetInterrupt()?
    * https://github.com/sago35/tinygo/blob/interrupts-pins-samd51/src/machine/machine_atsamd51.go#L384
3. I'd like to write a little nicer, but interrupt.New() only accepts const values, so this is how it's implemented.
    * https://github.com/sago35/tinygo/blob/interrupts-pins-samd51/src/machine/machine_atsamd51.go#L393
4. I want to make an example, but the feather-m4 I have doesn't have a button.
Creating an example for pybadge is one thing, but since there is no machine.Button, the other examples need to be modified.

PR #1139, opened 2020-05-28
sam: add support for pin change interrupts (samd5x)

sago35:
#1094
#1111 

I wrote the code for SAMD51J19A and SAMD51J20A.
I checked with feather-m4 and pyportal, and it seems to be working fine.

https://github.com/sago35/tinygo/tree/interrupts-pins-samd51

I'm struggling with how to write the following parts.
Is there a better way to do it?

1. `sam.EIC.CTRLA.Set(0)` or `sam.EIC.CTRLA.ClearBits(0x02)`
    * https://github.com/sago35/tinygo/blob/interrupts-pins-samd51/src/machine/machine_atsamd51.go#L352
2. Is it better to define it outside of SetInterrupt()?
    * https://github.com/sago35/tinygo/blob/interrupts-pins-samd51/src/machine/machine_atsamd51.go#L384
3. I'd like to write a little nicer, but interrupt.New() only accepts const values, so this is how it's implemented.
    * https://github.com/sago35/tinygo/blob/interrupts-pins-samd51/src/machine/machine_atsamd51.go#L393
4. ~~I want to make an example, but the feather-m4 I have doesn't have a button.
Creating an example for pybadge is one thing, but since there is no machine.Button, the other examples need to be modified.~~
    * Since wioterminal has a button, I'm going to make examples with #1124.

### DIFF
--- a/src/machine/machine_atsamd51.go
+++ b/src/machine/machine_atsamd51.go
@@ -44,6 +44,26 @@ const (
 	PinInputPulldown PinMode = 18
 )
 
+type PinChange uint8
+
+// Pin change interrupt constants for SetInterrupt.
+const (
+	PinRising  PinChange = sam.EIC_CONFIG_SENSE0_RISE
+	PinFalling PinChange = sam.EIC_CONFIG_SENSE0_FALL
+	PinToggle  PinChange = sam.EIC_CONFIG_SENSE0_BOTH
+)
+
+// Callbacks to be called for pins configured with SetInterrupt. Unfortunately,
+// we also need to keep track of which interrupt channel is used by which pin,
+// as the only alternative would be iterating through all pins.
+//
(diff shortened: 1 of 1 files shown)


sago35:
The SAMD51P19 has 128 pins, so there's a bit more code.
Also, since the 128 pin setting is not yet in machine_atsamd51.go, it would be better to change it in #1124.

sago35:
@deadprogram 
I cherry-picked it.

deadprogram:
Now merging, thanks for working on this @sago35

deadprogram:
You might want to submit another PR to change `sam.EIC.CTRLA.ClearBits(0x02)` to `sam.EIC.CTRLA.ClearBits(sam.EIC_CTRLA_ENABLE)`

PR #1141, opened 2020-05-29
sam: fix register access for interrupts pins in samd51

sago35:
#1139

@deadprogram 
This PR modifies the register access for interrupts pins on samd51.

Since only the EIC.CTRLA.ENABLE bit is required, only the EIC.CTRLA.ENABLE bit should be accessed and changed.

I tested it with feather-m4 and confirmed that it works.

### DIFF
--- a/src/machine/machine_atsamd51.go
+++ b/src/machine/machine_atsamd51.go
@@ -335,7 +335,7 @@ func (p Pin) SetInterrupt(change PinChange, callback func(Pin)) error {
 	pinCallbacks[extint] = callback
 	interruptPins[extint] = p
 
-	if (sam.EIC.CTRLA.Get() & sam.EIC_CTRLA_ENABLE) == 0 {
+	if !sam.EIC.CTRLA.HasBits(sam.EIC_CTRLA_ENABLE) {
 		// EIC peripheral has not yet been initialized. Initialize it now.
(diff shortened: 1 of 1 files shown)


src/machine/machine_atsamd51.go
```diff
@@ -335,7 +335,7 @@ func (p Pin) SetInterrupt(change PinChange, callback func(Pin)) error {
 	pinCallbacks[extint] = callback
 	interruptPins[extint] = p
 
-	if (sam.EIC.CTRLA.Get() & 0x02) == 0 {
+	if (sam.EIC.CTRLA.Get() & sam.EIC_CTRLA_ENABLE) == 0 {
```
deadprogram: You could use the `volatile` package `HasBits()` function here:

```go
if !sam.EIC.CTRLA.HasBits(sam.EIC_CTRLA_ENABLE) {
```

sago35: The source code has been modified.
It's better to use HasBits().

sago35:
I tested it with feather-m4 and confirmed that it works.
c9b37632aaf6fba0559d7f5e48948d1b5c68a274

deadprogram:
Squashed and merged, thank you @sago35

PR #1142, opened 2020-05-30
runtime: make channels work in interrupts

niaow:
This works right now, but there are some pieces here and there which are sub-optimal or have limited support. This is mostly a proof of concept that I hacked together tonight.

### DIFF
--- a/src/internal/task/queue.go
+++ b/src/internal/task/queue.go
@@ -1,5 +1,7 @@
 package task
 
+import "runtime/interrupt"
+
 const asserts = false
 
 // Queue is a FIFO container of tasks.
@@ -10,7 +12,9 @@ type Queue struct {
 
 // Push a task onto the queue.
 func (q *Queue) Push(t *Task) {
+	i := interrupt.Disable()
 	if asserts && t.Next != nil {
+		interrupt.Restore(i)
 		panic("runtime: pushing a task to a queue with a non-nil Next pointer")
 	}
 	if q.tail != nil {
@@ -21,12 +25,15 @@ func (q *Queue) Push(t *Task) {
 	if q.head == nil {
 		q.head = t
 	}
+	interrupt.Restore(i)
 }
 
 // Pop a task off of the queue.
 func (q *Queue) Pop() *Task {
+	i := interrupt.Disable()
 	t := q.head
 	if t == nil {
+		interrupt.Restore(i)
 		return nil
 	}
 	q.head = t.Next
@@ -34,11 +41,13 @@ func (q *Queue) Pop() *Task {
 		q.tail = nil
 	}
 	t.Next = nil
+	interrupt.Restore(i)
 	return t
 }
 
 // Append pops the contents of another queue and pushes them onto the end of this queue.
 func (q *Queue) Append(other *Queue) {
+	i := interrupt.Disable()
 	if q.head == nil {
 		q.head = other.head
 	} else {
@@ -46,6 +55,15 @@ func (q *Queue) Append(other *Queue) {
 	}
 	q.tail = other.tail
 	other.head, other.tail = nil, nil
+	interrupt.Restore(i)
+}
+
+// Empty checks if the queue is empty.
+func (q *Queue) Empty() bool {
+	i := interrupt.Disable()
+	empty := q.head == nil
+	interrupt.Restore(i)
+	return empty
 }
 
 // Stack is a LIFO container of tasks.
@@ -57,19 +75,24 @@ type Stack struct {
 
 // Push a task onto the stack.
 func (s *Stack) Push(t *Task) {
+	i := interrupt.Disable()
 	if asserts && t.Next != nil {
+		interrupt.Restore(i)
 		panic("runtime: pushing a task to a stack with a non-nil Next pointer")
 	}
 	s.top, t.Next = t, s.top
+	interrupt.Restore(i)
 }
 
 // Pop a task off of the stack.
 func (s *Stack) Pop() *Task {
+	i := interrupt.Disable()
 	t := s.top
 	if t != nil {
 		s.top = t.Next
 		t.Next = nil
 	}
+	interrupt.Restore(i)
 	return t
 }
 
@@ -89,10 +112,13 @@ func (t *Task) tail() *Task {
 // Queue moves the contents of the stack into a queue.
 // Elements can be popped from the queue in the same order that they would be popped from the stack.
 func (s *Stack) Queue() Queue {
+	i := interrupt.Disable()
 	head := s.top
 	s.top = nil
-	return Queue{
+	q := Queue{
 		head: head,
 		tail: head.tail(),
 	}
+	interrupt.Restore(i)
+	return q
 }
--- a/src/runtime/arch_cortexm.go
+++ b/src/runtime/arch_cortexm.go
@@ -102,3 +102,7 @@ func procPin() {
 func procUnpin() {
 	arm.EnableInterrupts(procPinnedMask)
 }
+
+func waitForEvents() {
+	arm.Asm("wfe")
+}
--- a/src/runtime/arch_tinygoriscv.go
+++ b/src/runtime/arch_tinygoriscv.go
@@ -90,3 +90,11 @@ func procPin() {
 func procUnpin() {
 	riscv.EnableInterrupts(procPinnedMask)
 }
+
+func waitForEvents() {
+	mask := riscv.DisableInterrupts()
+	if !runqueue.Empty() {
+		riscv.Asm("wfi")
+	}
+	riscv.EnableInterrupts(mask)
+}
--- a/src/runtime/chan.go
+++ b/src/runtime/chan.go
@@ -25,6 +25,7 @@ package runtime
 
 import (
 	"internal/task"
+	"runtime/interrupt"
 	"unsafe"
 )
 
@@ -308,13 +309,17 @@ func (ch *channel) trySend(value unsafe.Pointer) bool {
 		return false
 	}
 
+	i := interrupt.Disable()
+
 	switch ch.state {
 	case chanStateEmpty, chanStateBuf:
 		// try to dump the value directly into the buffer
 		if ch.push(value) {
 			ch.state = chanStateBuf
+			interrupt.Restore(i)
 			return true
 		}
+		interrupt.Restore(i)
 		return false
 	case chanStateRecv:
 		// unblock reciever
@@ -328,16 +333,21 @@ func (ch *channel) trySend(value unsafe.Pointer) bool {
 			ch.state = chanStateEmpty
 		}
 
+		interrupt.Restore(i)
 		return true
 	case chanStateSend:
 		// something else is already waiting to send
+		interrupt.Restore(i)
 		return false
 	case chanStateClosed:
+		interrupt.Restore(i)
 		runtimePanic("send on closed channel")
 	default:
+		interrupt.Restore(i)
 		runtimePanic("invalid channel state")
 	}
 
+	interrupt.Restore(i)
 	return false
 }
 
@@ -350,6 +360,8 @@ func (ch *channel) tryRecv(value unsafe.Pointer) (bool, bool) {
 		// this is non-blocking, so just say no
 		return false, false
 	}
+
+	i := interrupt.Disable()
 
 	switch ch.state {
 	case chanStateBuf, chanStateSend:
@@ -373,6 +385,7 @@ func (ch *channel) tryRecv(value unsafe.Pointer) (bool, bool) {
 				ch.state = chanStateEmpty
 			}
 
+			interrupt.Restore(i)
 			return true, true
 		} else if ch.blocked != nil {
 			// unblock next sender if applicable
@@ -386,19 +399,24 @@ func (ch *channel) tryRecv(value unsafe.Pointer) (bool, bool) {
 				ch.state = chanStateEmpty
 			}
 
+			interrupt.Restore(i)
 			return true, true
 		}
+		interrupt.Restore(i)
 		return false, false
 	case chanStateRecv, chanStateEmpty:
 		// something else is already waiting to recieve
+		interrupt.Restore(i)
 		return false, false
 	case chanStateClosed:
 		if ch.pop(value) {
+			interrupt.Restore(i)
 			return true, true
 		}
 
 		// channel closed - nothing to recieve
 		memzero(value, ch.elementSize)
+		interrupt.Restore(i)
 		return true, false
 	default:
 		runtimePanic("invalid channel state")
@@ -447,14 +465,18 @@ type chanSelectState struct {
 // This operation will block unless a value is immediately available.
 // May panic if the channel is closed.
 func chanSend(ch *channel, value unsafe.Pointer, blockedlist *channelBlockedList) {
+	i := interrupt.Disable()
+
 	if ch.trySend(value) {
 		// value immediately sent
 		chanDebug(ch)
+		interrupt.Restore(i)
 		return
 	}
 
 	if ch == nil {
 		// A nil channel blocks forever. Do not schedule this goroutine again.
+		interrupt.Restore(i)
 		deadlock()
 	}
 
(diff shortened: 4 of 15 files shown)


src/runtime/chan.go
```diff
@@ -476,14 +487,18 @@ func chanSend(ch *channel, value unsafe.Pointer) {
...
 		chanDebug(ch)
+		exitCritical()
 		return ok
 	}
 
 	if ch == nil {
 		// A nil channel blocks forever. Do not schedule this goroutine again.
+		exitCritical()
 		deadlock()
 	}
```
niaow, line 517: TODO: dont allocate in a critical section

Or better yet, don't allocate at all

src/runtime/runtime_atsamd21.go
```diff
@@ -302,8 +305,18 @@ func timerSleep(ticks uint32) {
...
 
-	for timerWakeup.Get() == 0 {
-		arm.Asm("wfi")
+wait:
+	arm.Asm("wfi")
+	if timerWakeup.Get() != 0 {
+		return true
+	}
+	if hasScheduler {
+		// The interurpt may have awoken a goroutine, so bail out early.
```
niaow, line 316: This might have a race condition somewhere?

niaow:
@aykevl Is this design more of what you were looking for, rather than the other scheduler design?

go.sum
```diff
@@ -41,6 +41,7 @@ golang.org/x/sys v0.0.0-20190215142949-d0b11bdaac8a/go.mod h1:STP8DvDyc/dI5b8T5h
 golang.org/x/sys v0.0.0-20190412213103-97732733099d/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/sys v0.0.0-20191128015809-6d18c012aee9 h1:ZBzSG/7F4eNKz2L3GE9o300RX0Az1Bw5HF7PDraD+qU=
 golang.org/x/sys v0.0.0-20191128015809-6d18c012aee9/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
+golang.org/x/sys v0.0.0-20200116001909-b77594299b42 h1:vEOn+mP2zCOVzKckCZy6YsCtDblrpj/w7B9nxGNELpg=
```
niaow: I am not sure why this keeps getting added.

src/runtime/chan.go
niaow: fix put into #1144

go.sum
aykevl: I have the same issue, I think it just needs to be added?

src/runtime/runtime_atsamd21.go
```diff
@@ -302,8 +305,18 @@ func timerSleep(ticks uint32) {
 	// enable IRQ for CMP0 compare
 	sam.RTC_MODE0.INTENSET.SetBits(sam.RTC_MODE0_INTENSET_CMP0)
 
-	for timerWakeup.Get() == 0 {
-		arm.Asm("wfi")
+wait:
+	arm.Asm("wfi")
```
aykevl: Huh, interesting. There is a subtle difference between `wfe` and `wfi`. I think it's best to ~always use `wfe`, as it avoids a race condition if the interrupt that you're waiting for happens right before the `wfi` instruction (the `wfi` would wait until the next interrupt, while `wfe` would be a `nop` in that case).

But it's somewhat unrelated as this code was already present.

src/runtime/arch_cortexm.go
```diff
@@ -102,3 +102,7 @@ func procPin() {
 func procUnpin() {
 	arm.EnableInterrupts(procPinnedMask)
 }
+
+func waitForEvents() {
+	arm.Asm("wfi")
```
aykevl: Same as in the previous comment (`wfi` vs `wfe`).

src/runtime/arch_tinygoriscv.go
```diff
@@ -90,3 +90,7 @@ func procPin() {
 func procUnpin() {
 	riscv.EnableInterrupts(procPinnedMask)
 }
+
+func waitForEvents() {
+	riscv.Asm("wfi")
```
aykevl: Reading the RISC-V specification, I think the same race condition is present. I'm not sure how to work around that, but there appears to be a trick with disabling/enabling interrupts.

src/runtime/chan.go
```diff
@@ -308,6 +308,9 @@ func (ch *channel) trySend(value unsafe.Pointer) bool {
 		return false
 	}
 
+	critical()
+	defer exitCritical()
```
aykevl: I'm somewhat uncomfortable with `defer` in an interrupt, as there isn't a guarantee `defer` won't allocate. Maybe it's better to be explicit and move the `exitCritical()` to where it should be run.

src/runtime/critical_cortexm.go
```diff
@@ -0,0 +1,22 @@
...
+package runtime
+
+import "device/arm"
+
+var criticalLevel uint32
+var criticalState uintptr
+
+func critical() {
+	if criticalLevel == 0 {
+		criticalState = arm.DisableInterrupts()
```
aykevl: There is a much better way to do this, and it is with saving and restoring the previous interrupt-enabled state:

    mask := arm.DisableInterrupts()
    // critical section
    arm.EnableInterrupts(mask)

It is safe to use this pattern in a nested fashion. Note that `EnableInterrupts` is a bit of a misnomer, maybe it should be called `RestoreInterrupts` as it restores the interrupt-enabled state from before the `DisableInterrupts` call.

The added benefit is that the code is (probably, untested) much smaller as there are no globals involved.

aykevl (commented):
I still need to look at this properly (and I intend to do so very soon), but I quickly skimmed through the code.

src/runtime/critical_cortexm.go
niaow: I was trying to make something that would work also on platforms where interrupt disabling worked differently, although now i guess for them we can use a global and return a boolean.

niaow:
I don't want to scope-creep this PR any further, so I would suggest adding support for other boards in separate PRs after this is merged, because otherwise we would basically have to test every board to merge this.

deadprogram:
@jaddr2line please resolve merge conflict.

niaow:
Rebased.

src/runtime/runtime_atsamd21.go
niaow: Fixed.

niaow:
Rebased again after the go mod tidy.

src/runtime/runtime_atsamd21.go
```diff
@@ -301,9 +305,21 @@ func timerSleep(ticks uint32) {
...
-	for timerWakeup.Get() == 0 {
-		arm.Asm("wfi")
+wait:
+	arm.Asm("wfe")
+	if timerWakeup.Get() != 0 {
+		return true
+	}
+	if hasScheduler {
+		// The interurpt may have awoken a goroutine, so bail out early.
+		rtcInterrupt.Disable()
```
aykevl: Instead of enabling/disabling the entire RTC_MODE0 interrupt, you could instead disable just the CMP0 bit (as is done in the interrupt handler). Or, if that causes problems, perhaps enable the CMP0 bit in initRTC and control the interrupt by enabling/disabling the entire peripheral (`rtcInterrupt`). Using both together might cause problems.

src/runtime/runtime_atsamd21.go
```diff
@@ -280,7 +284,7 @@ func ticks() timeUnit {
 }
 
 // ticks are in microseconds
-func timerSleep(ticks uint32) {
+func timerSleep(ticks uint32) bool {
```
aykevl, line 288: It may be worth documenting what the bool means. It looks like it means that the sleep has completed (and not been interrupted by some other interrupt)?

src/runtime/critical_cortexm.go
```diff
@@ -0,0 +1,13 @@
...
+
+import "device/arm"
+
+func critical() uintptr {
+	return arm.DisableInterrupts()
+}
+
+func exitCritical(state uintptr) {
+	arm.EnableInterrupts(state)
+}
```
aykevl: While I think we need this functionality, I think it's best put in the runtime/interrupt package. What do you think? The package can provide a platform independent way to disable/restore interrupts (or provide critical sections, as they're often called). This can perhaps be done in a separate PR.

(I wanted to do that but never got around to actually doing it).

aykevl (commented):
I have taken a bit more in-depth look, I need to look at the write barrier next.

Sorry, lots of things going on right now in my life so don't have as much time as I would hope.

Sidenote (semi-related): the functionality I'm most waiting for is some sort of simple condition variable that one goroutine can wait on and an interrupt can signal. That's what is needed by the vast majority of hardware, for example to queue a DMA SPI transfer and wait for it to finish. Or, in my case, wait for a BLE operation to complete (in the nrf SoftDevice) while another goroutine is allowed to run. Right now I just wait for it to complete, which is a waste:

```go
// Wait for the next advertisement packet to arrive.
// TODO: use some sort of condition variable once the scheduler supports
// them.
arm.Asm("wfe")
if gotScanReport.Get() == 0 {
	// Spurious event. Continue waiting.
	continue
}
gotScanReport.Set(0)
```

Of course, that should happen after this PR is merged (which lays the foundation for more interrupt <-> goroutine communication).

src/runtime/critical_cortexm.go
niaow: Alright, I will move it there.

aykevl:
I think the write barrier system is solid. I am however not sure whether disabling interrupts in the GC is ever a good idea. Even the possibility of it happening may be a problem for real time applications, while endless GC loops seem rather unlikely to me (except for really badly designed interrupt handlers). I'm inclined to say it's better to risk unbounded GC than to disable interrupts in the GC.

I'll think a bit about the best strategy.

aykevl:
Actually, it seems to me that every heap pointer move would be a task pointer move from some arbitrary location (for example the blocked list in a channel) to the runqueue. Therefore it might be enough to scan the runqueue with interrupts disabled, or rather (after the heap has been scanned) do something like this:

```python
while True:
  task = nil
  disable interrupts
    for t in runqueue:
      if t not marked:
        task = t
        break
  enable interrupts
  if task == nil:
    break # finished scanning
  scan task
```

Maybe we have already discussed this, but I don't see an obvious flaw here. The advantage here is that the number of goroutines is limited, especially on bare metal systems it may be just a few or only one. Certainly iterating through the runqueue will be much more predictable than rescanning the entire heap.

niaow:
@aykevl I have removed the write barrier with a pattern like what you suggested.

niaow:
Alright, this should finally be done!

deadprogram:
I am doing some testing with this PR now...

deadprogram:
So this PR currently causes the ItsyBitsy-M4 SAMD51 board to become unresponsive. I could recover by using the manual bootloader and then flashing code from the previous version...

niaow:
Unresposive when running what?

deadprogram:
Once flashed, the MCU "locks up" and even the USB CDC code no longer works. The IRQ that is uses to detect the USB CDC is no longer called, so the MCU no longer goes into bootloader when switching to 1200 baud.

deadprogram:
When I was finally able to get GDB to connect (had a bad solder joint on my connector) here is the stack trace I was able to obtain:

```
(gdb) continue                                                                                                                                                               
Continuing.                                                                                                                                                                  
^C                                                                                                                                                                           
Program received signal SIGTRAP, Trace/breakpoint trap.                                                                                                                      
runtime.timerSleep (ticks=<optimized out>) at /home/ron/.gvm/pkgsets/go1.14.2/global/src/github.com/tinygo-org/tinygo/src/runtime/runtime_atsamd51.go:290                    
290             for timerWakeup.Get() == 0 {                                                                                                                                 
(gdb) bt                                                                                                                                                                     
#0  runtime.timerSleep (ticks=<optimized out>) at /home/ron/.gvm/pkgsets/go1.14.2/global/src/github.com/tinygo-org/tinygo/src/runtime/runtime_atsamd51.go:290                
#1  0x0000655c in runtime.sleepTicks (d=16384) at /home/ron/.gvm/pkgsets/go1.14.2/global/src/github.com/tinygo-org/tinygo/src/runtime/runtime_atsamd51.go:252                
#2  0x00006530 in runtime.scheduler () at /home/ron/.gvm/pkgsets/go1.14.2/global/src/github.com/tinygo-org/tinygo/src/runtime/scheduler.go:153                               
#3  0x00006456 in runtime.run () at /home/ron/.gvm/pkgsets/go1.14.2/global/src/github.com/tinygo-org/tinygo/src/runtime/scheduler_any.go:24                                  
#4  0x00006254 in Reset_Handler () at /home/ron/.gvm/pkgsets/go1.14.2/global/src/github.com/tinygo-org/tinygo/src/runtime/runtime_atsamd51.go:20
```

I discovered that the same problem of MCU locks up even on `blinky1` exists on the nrf52840 when I accidentally flashed my Circuit Playground Bluefruit with code from this branch. Oops!

deadprogram:
The issue I reported regarding M4 processors was corrected by #1182 which has now been merged into `dev`. @jaddr2line if you rebase `dev` into this branch then all the tests should pass now.

niaow:
Rebased.

aykevl:
I did a size comparison before and after this PR:

* Somehow amd64 is affected by this: code size is usually increased a bit. I did not expect that to happen. Not a big problem but it might be a symptom of something.
* Similarly, WebAssembly outputs tend to be a bit larger.
* As expected, code size is increased a little bit for most baremetal targets (usually around 150-200 bytes). I think this increase is worth the extra functionality it provides.

niaow:
> Somehow amd64 is affected by this: code size is usually increased a bit.

This probably has to do with the GC changes.

aykevl:
Right, that makes sense. For amd64 I don't really care but for wasm it's a bit unfortunate that it adds extra overhead for something that doesn't happen on WebAssembly.

niaow:
I have now put parts of the conservative collector behind the baremetal build tag, and they will only build if the scheduler is enabled

The test failure is completely unrelated, and is an instance of #1185

deadprogram:
@jaddr2line if you rebase against dev the tests should pass now due to https://github.com/tinygo-org/tinygo/commit/1ad69538588c076c72066fd78f2d5a6089edb05b

niaow:
Rebased.

src/internal/task/queue.go
```diff
@@ -10,7 +12,9 @@ type Queue struct {
 
 // Push a task onto the queue.
 func (q *Queue) Push(t *Task) {
+	i := interrupt.Disable()
```
aykevl, line 15: It might be more efficient to put all these locks where the queue/stack are used.

src/runtime/gc_conservative.go
```diff
@@ -292,9 +294,39 @@ func GC() {
...
 	// Mark phase: mark all reachable objects, recursively.
-	markGlobals()
 	markStack()
-	finishMark()
+	markGlobals()
+
+	if baremetal && hasScheduler {
+		// Channel operations in interrupts may move task pointers around while we are marking.
+		// Therefore we need to scan the runqueue seperately.
+		var markedTaskQueue task.Queue
```
aykevl, line 303: I was hoping you could just go through the runqueue instead of (essentially) rewriting it, but it seems fine. And perhaps safer.

aykevl (commented):
I did another size comparison. The last commit (to reduce size) did not have an effect on amd64 but did reduce code size on wasm. Still there is a small (much more acceptable) size increase on wasm. Interestingly, it did sometimes reduce .bss size by 4 bytes on baremetal targets.

Size results (for diffing):
https://gist.github.com/aykevl/ea54c616224e5f8af55545cef7387d77

This is ready to go from my POV. @deadprogram any last comments before merging?

(some thoughts below inline, probably not worth changing)

deadprogram:
From my POV this PR is ready to go. The small comments made by @aykevl are either not worth changing, or can be looked at for future refinements.

Now merging, incredible work again from @jaddr2line thank you!! :cake: :fireworks:

PR #1144, opened 2020-05-30
compiler/runtime: move the channel blocked list onto the stack

### LINKED ISSUE #1143: runtime: chansend and chanrecv operations should not allocate
While working on #1142, I ran into the issue that channels may allocate. Since they are now becoming critical sections, this becomes more of an issue since this can cause the garbage collector to run with interrupts disabled.

We should modify the code generator to stack-allocate the `channelBlockedList`.

niaow:
Previously, chansend and chanrecv allocated a heap object before blocking on a channel. This object was used to implement a linked list of goroutines blocked on the channel. The chansend and chanrecv now instead accept a buffer to store this object in as an argument. The compiler now creates a stack allocation for this object and passes it in.

Fixes #1143

### DIFF
--- a/compiler/channel.go
+++ b/compiler/channel.go
@@ -35,12 +35,17 @@ func (b *builder) createChanSend(instr *ssa.Send) {
 	valueAlloca, valueAllocaCast, valueAllocaSize := b.createTemporaryAlloca(valueType, "chan.value")
 	b.CreateStore(chanValue, valueAlloca)
 
-	// Do the send.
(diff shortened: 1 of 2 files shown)


aykevl:
I checked the binary size impact. The only difference was (as expected) for testdata/channel.go, increasing by around 0.6% (wasm, microbit)  to 2.5% (amd64). Not too much (and the lack of GC is good), but good to be aware of. Normally such a size impact would be a problem but in normal code channels aren't used nearly as much as in testdata/channel.go. And it brings the advantage that blocking on a channel send/receive won't trigger the GC, which is expensive.

PR #1145, opened 2020-05-30
Dockerfile: avoid duplicate LLVM apt line

aykevl:
This line causes problems when installing software: `apt-get` complains that there are duplicate lines.

I'm not sure how to test this, apart from building the container. But it _looks_ correct.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,10 +27,7 @@ COPY --from=tinygo-base /go/bin/tinygo /go/bin/tinygo
 COPY --from=tinygo-base /tinygo/src /tinygo/src
 COPY --from=tinygo-base /tinygo/targets /tinygo/targets
 
-RUN wget -O- https://apt.llvm.org/llvm-snapshot.gpg.key| apt-key add - && \
-    echo "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-10 main" >> /etc/apt/sources.list && \
-    apt-get update && \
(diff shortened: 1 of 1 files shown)


aykevl:
I was hitting this problem while trying to install a Windows cross compiler in the tinygo-dev image, to test the Bluetooth package I'm writing.

Dockerfile
```diff
@@ -27,10 +27,7 @@ COPY --from=tinygo-base /go/bin/tinygo /go/bin/tinygo
 COPY --from=tinygo-base /tinygo/src /tinygo/src
 COPY --from=tinygo-base /tinygo/targets /tinygo/targets
 
-RUN wget -O- https://apt.llvm.org/llvm-snapshot.gpg.key| apt-key add - && \
-    echo "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-10 main" >> /etc/apt/sources.list && \
-    apt-get update && \
-    apt-get install -y libllvm10 lld-10
+RUN apt-get install -y libllvm10 lld-10
```
johanbrandhorst, line 30: Is this line even necessary if it's inheriting from tinygo-base? Looks like it's already installed in the base.

aykevl: I thought so too at first, but it installs different packages.

johanbrandhorst (approved):

deadprogram:
Looks good, now merging. Thanks!

PR #1146, opened 2020-05-31
builder: use newer version of gohex

aykevl:
This version adds error handling for the DumpIntelHex method, and thus fixes a TODO comment.


deadprogram:
:1234: now merging.

PR #1149, opened 2020-06-01
riscv: add new kendryte k210 chip

yannishuber:
This PR adds support for the Kendryte [K210](https://s3.cn-north-1.amazonaws.com.cn/dl.kendryte.com/documents/kendryte_datasheet_20181011163248_en.pdf) chip and the Sipeed [MAix Bit](https://www.seeedstudio.com/Sipeed-MAix-BiT-for-RISC-V-AI-IoT-p-2872.html) board.

TODO:

- [x]  Runtime code
- [x]  UART
- [x]  GPIO
- [x]  GPIO interrupts
- [x]  SPI
- [x]  I²C
- [x]  Unit tests


yannishuber:
The CMSIS-SVD repository has temporarly been changed and will be reset to the original upstream once https://github.com/posborne/cmsis-svd/pull/104 is merged.

targets/riscv.json
```diff
@@ -1,26 +1,12 @@
 {
```
niaow: We may need to split our target specs into 32 and 64 bit RISC-V? I think that we also probably need to use a build tag to distinguish between the two.

yannishuber: I agree, now there are two files and I also added a new `tinygo.riscv64` build tag. However I have kept the `tinygo.riscv` build tag for 32-bit architectures to assure backward compatibility. I don't know if it would be better to replace it by `tinygo.riscv32` to be more explicit?

compileopts/target.go
```diff
@@ -272,6 +276,7 @@ func defaultTarget(goos, goarch, triple string) (*TargetSpec, error) {
 		GDB:         "gdb",
 		PortReset:   "false",
 		FlashMethod: "native",
+		CodeModel:   "default",
```
aykevl: You could leave this out because an empty code model is already equivalent to "default" (to reduce clutter).

targets/fe310.json
```diff
@@ -1,5 +1,5 @@
 {
-	"inherits": ["riscv"],
+	"inherits": ["riscv32"],
 	"features": ["+a", "+c", "+m"],
-	"build-tags": ["fe310", "sifive"]
+	"build-tags": ["fe310", "sifive"]
```
aykevl: It looks like you added some trailing whitespace here.

targets/maixbit.ld
```diff
@@ -0,0 +1,11 @@
+
+MEMORY
+{
+    RAM (xrw) : ORIGIN = 0x80000000, LENGTH = (6 * 1024 * 1024)
```
aykevl: I didn't know such an expression was allowed, but you can make the code more readable with `6M` (meaning 6 megabytes).

targets/maixbit.ld
```diff
@@ -0,0 +1,11 @@
...
+}
+
+REGION_ALIAS("FLASH_TEXT", RAM);
```
aykevl: This seems to indicate that flash and RAM are in the same region. How does that work exactly? Is there a bootloader that loads the entire binary into RAM at reset?

targets/riscv64.json
```diff
@@ -0,0 +1,30 @@
+{
```
aykevl: Perhaps you could make a parent riscv.json that is inherited by riscv32.json and riscv64.json?
I agree it's a good idea to split. Using a common parent might avoid some bugs where some things are unintentionally different between RV32 and RV64.

aykevl (commented):
Overall this looks good. I have left a number of comments inline, however.
Regarding cmsis-svd, while I would prefer to use upstream, I think practically it would be better to fork it into tinygo-org (as we did previously) and keeping it there, even as cmsis-svd upstream merges PRs. Switching between submodule remotes has resulted in trouble before so let's do it once for good.

Looking at this PR, there are really three somewhat distinct changes (two necessary as precondition to K210 support):

* Adding support for non-default code models.
* Adding support for 64-bit RISC-V.
* Adding support for the K210 chip.

You could cleanly separate these in different commits (with a rebase and force-push) or actually make separate PRs, what you prefer. Separate PRs may be faster to merge individually.

targets/maixbit.ld
yannishuber: There is no integrated flash in the K210 chip. Exactly, the executable is flashed onto an external SPI flash and is copied at reset by a bootloader stored on the ROM to the RAM. I added the alias to keep the `riscv.ld` file unchanged.

targets/riscv64.json
yannishuber: done

targets/maixbit.ld
yannishuber: done

yannishuber:
@aykevl I made two PRs in the tinygo-org/cmsis-svd repo to update it and to incorporate the unmerged fix (https://github.com/tinygo-org/cmsis-svd/pull/3 and https://github.com/tinygo-org/cmsis-svd/pull/4). I will update the submodule repository as soon as these are merged.

I will also separate this pull request into 3 distinct ones as suggested.

deadprogram:
All of the PRs to the tinygo fork of cmsis-svd are done. Please continue with the rest of the steps so we can get this merged! Thank you.

targets/maixbit.ld
aykevl: Right. That's not very common for baremetal systems, hence my confusion. Thanks for explaining.

deadprogram:
I should be receiving a Sipeed MAix BiT board next week, so I can also help test.

targets/maixbit.json
```diff
@@ -0,0 +1,6 @@
+{
+	"inherits": ["k210"],
+	"build-tags": ["maixbit"],
+	"linkerscript": "targets/maixbit.ld",
+	"flash-command": "kflash -p /dev/ttyUSB0 {hex}"
```
deadprogram: You should use `{port}` here instead of assuming `/dev/ttyUSB0`.

deadprogram:
So this PR is the precursor to adding the board and machine files for the MAiX BiT, correct? @yannishuber are you planning on doing this? I just got my boards today.  :smile_cat:

yannishuber:
@deadprogram yes exactly, i also have this board and I am currently working on it. It should be done soon.

deadprogram:
So should this PR still be considered draft, or should we look at merging it?

deadprogram:
@yannishuber is this still a draft PR, or should it be able to be merged into `dev`?

yannishuber:
@deadprogram Sorry I didn't see your first reply. No I was thinking of adding the code related to the maix-bit in this PR but I can also make a separate PR if you want. I will work on this today.

deadprogram:
It can be in the same PR, since that lets us test the full implementation that way. Thank you!

src/runtime/arch_tinygoriscv64.go
```diff
@@ -0,0 +1,92 @@
+// +build tinygo.riscv64
+
+package runtime
+
+import "device/riscv"
+
+const GOARCH = "arm64" // riscv pretends to be arm
```
yannishuber, line 5: Could we use `GOARCH = "riscv64"` here? It seems to have "experimental" support since [Go 1.14](https://golang.org/doc/go1.14#riscv).

deadprogram: @jaddr2line or @aykevl ?

aykevl: Unfortunately not, if we want to retain support for older Go versions. And honestly the `GOARCH` value doesn't matter a whole lot in practice.

src/device/riscv/handleinterrupt64.S
```diff
@@ -0,0 +1,44 @@
+.section .text.handleInterruptASM
+.global handleInterruptASM
+.type handleInterruptASM,@function
+handleInterruptASM:
+    // Save and restore all registers, because the hardware only saves/restores
+    // the pc.
+    // Note: we have to do this in assembly because the "interrupt"="machine"
+    // attribute is broken in LLVM: https://bugs.llvm.org/show_bug.cgi?id=42984
```
aykevl: Sidenote (unrelated to this PR): this should be fixed with LLVM 11 so that we won't need this workaround anymore.

src/machine/board_k210.go
```diff
@@ -0,0 +1,317 @@
+// +build maixbit
+
```
aykevl, line 2: Maybe you can add a link to the datasheet here?
For example https://github.com/kendryte/kendryte-doc-datasheet/blob/master/en/SUMMARY.md. I couldn't easily find a PDF that looked authoritative (although a PDF might be preferred).

src/machine/machine_k210.go
```diff
@@ -0,0 +1,94 @@
...
+func (p Pin) Set(high bool) {
+}
+
+// Get returns the current value of a GPIO pin.
+func (p Pin) Get() bool {
+	return true
+}
+
+type FPIOA struct {
+	Bus *kendryte.FPIOA_Type
```
aykevl: What is this exactly? It appears to be a kind of I/O multiplexer, similar to what most other MCUs have (like the atsamd chips). But this particular interface is probably very chip specific, which hurts portability. Would it be possible to implement just the GPIO bits here (in the empty `Pin` methods above)?

src/machine/machine_k210.go
```diff
@@ -0,0 +1,94 @@
...
+var (
+	FPIOA0 = FPIOA{Bus: kendryte.FPIOA}
+)
+
+func (fpioa FPIOA) Init() {
+	// Enable APB0 clock.
+	kendryte.SYSCTL.CLK_EN_CENT.Set(kendryte.SYSCTL_CLK_EN_CENT_APB0_CLK_EN)
+
+	// Enable FPIOA peripheral.
+	kendryte.SYSCTL.CLK_EN_PERI.Set(kendryte.SYSCTL_CLK_EN_PERI_FPIOA_CLK_EN)
```
aykevl: This looks like a job for the runtime to do at reset.

src/machine/machine_k210.go
```diff
@@ -0,0 +1,94 @@
...
+	intr.Enable()
+}
+
+func (uart *UART) handleInterrupt(interrupt.Interrupt) {
+	rxdata := uart.Bus.RXDATA.Get()
+	c := byte(rxdata)
+	if uint32(c) != rxdata {
+		// The rxdata has other bits set than just the low 8 bits. This probably
+		// means that the 'empty' flag is set, which indicates there is no data
+		// to be read and the byte is garbage. Ignore this byte.
```
aykevl, line 379: Huh, this looks copied from the fe310 chip. Just to confirm: is it indeed a similar peripheral?

src/runtime/arch_tinygoriscv64.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+//export __atomic_fetch_add_8
+func __atomic_fetch_add_8(ptr *uint64, value uint64, ordering int32) uint64 {
+	mask := riscv.DisableInterrupts()
+	oldValue := *ptr
+	*ptr = oldValue + value
+	riscv.EnableInterrupts(mask)
+	return oldValue
+}
```
aykevl: This should not be necessary for the K210 chip as it implements the atomic extension. And even if it is necessary, it's better to move it to src/runtime/baremetal.go (using the runtime/interrupt package to disable/restore interrupts) - but that can be done later when necessary.

src/runtime/interrupt/interrupt_k210.go
```diff
@@ -0,0 +1,19 @@
+// +build k210
```
aykevl, line 1: This file may need to be extended since https://github.com/tinygo-org/tinygo/pull/1171 got merged.

src/runtime/runtime_k210.go
```diff
@@ -0,0 +1,159 @@
...
+func postinit() {}
+
+//export main
+func main() {
+
+	// Both harts should disable all interrupts on startup.
+	initPLIC()
+
+	// Only use one hart for the moment.
+	if riscv.MHARTID.Get() == 0 {
```
aykevl: Perhaps you can reformat this code to be more Go style, to reserve indentation for exceptions.

```go
	if riscv.MHARTID.Get() != 0 {
		abort()
	}
	// normal code
```

targets/riscv.ld
```diff
@@ -18,7 +18,7 @@ SECTIONS
      * See: http://blog.japaric.io/stack-overflow-protection/ */
     .stack (NOLOAD) :
     {
-        . = ALIGN(4);
+        . = ALIGN(8);
```
aykevl: This is a change for all chips but probably fine.

aykevl (commented):
This chip seems to be pretty similar to the SiFive fe310, so much that I wonder whether some code can be reused between them. That would require putting the SiFive and Kendryte vendor code in the same package (perhaps device/riscv?).

src/machine/machine_k210.go
yannishuber: Yes it is. From the kendryte standalone SDK:
```c
typedef struct _uarths_rxdata
{
    /* Bits [7:0] is data */
    uint32_t data : 8;
    /* Bits [30:8] is 0 */
    uint32_t zero : 23;
    /* Bit 31 is empty status */
    uint32_t empty : 1;
} __attribute__((packed, aligned(4))) uarths_rxdata_t
```

targets/riscv.ld
yannishuber: Yes I know, according to the RISC-V ABI the stack should even be 16 bytes aligned but since none of the chips currently supported by TinyGo have the Q extension I think we can settle with an alignment of 8 bytes. (more details [here](https://github.com/riscv/riscv-elf-psabi-doc/issues/21))

src/runtime/interrupt/interrupt_k210.go
yannishuber: Looking at #1171 and the `interrupt_sifive.go` file I think this is correct but I may be missing something.

src/machine/machine_k210.go
yannishuber: Yes it is very chip specific. I think it is possible to only implement GPIO, I will look into that.

src/machine/board_k210.go
yannishuber: done

src/runtime/runtime_k210.go
yannishuber: done

src/runtime/arch_tinygoriscv64.go
yannishuber: Ok I will remove this file for now.

yannishuber: Huh not the whole file actually but the atomic operations

src/machine/machine_k210.go
yannishuber: Yes this function is called by the runtime in `initPeripherals()`

src/machine/machine_k210.go
```diff
@@ -0,0 +1,166 @@
...
+		p.fpioaSetIOPull(fpioaPullDown)
+		input = true
+
+	case PinOutput:
+		p.fpioaSetIOPull(fpioaPullDown)
+		input = false
+
+	}
+
+	if p >= D0 && p <= D7 {
```
aykevl:
```suggestion
	if p >= P08 && p <= P15 {
```

These `D0` etc pins are board specific.

src/machine/machine_k210.go
```diff
@@ -0,0 +1,166 @@
...
+		} else {
+			kendryte.GPIOHS.OUTPUT_EN.SetBits(1 << gpioPin)
+			kendryte.GPIOHS.INPUT_EN.ClearBits(1 << gpioPin)
+		}
+	}
+}
+
+// Set the pin to high or low.
+func (p Pin) Set(high bool) {
+	if p >= D0 && p <= D7 {
```
aykevl: Same here.

src/machine/machine_k210.go
```diff
@@ -0,0 +1,166 @@
...
+		} else {
+			kendryte.GPIOHS.OUTPUT_VAL.ClearBits(1 << gpioPin)
+		}
+	}
+}
+
+// Get returns the current value of a GPIO pin.
+func (p Pin) Get() bool {
+	var val uint32
+	if p >= D0 && p <= D7 {
```
aykevl: And here.

src/machine/machine_k210.go
```diff
@@ -0,0 +1,166 @@
...
+	case PinInputPullUp:
+		p.fpioaSetIOPull(fpioaPullUp)
+		input = true
+
+	case PinInputPullDown:
+		p.fpioaSetIOPull(fpioaPullDown)
+		input = true
+
+	case PinOutput:
+		p.fpioaSetIOPull(fpioaPullDown)
```
aykevl: Is this correct? Should an output pin be pulled low?

src/machine/machine_k210.go
yannishuber: I changed this to use the chip pins. thank you

src/machine/machine_k210.go
yannishuber: This is how Kendryte does it in its SDK. However, I did some tests and the pull setting seems to be ignored if it is an output pin. To avoid confusion I changed it to pull none.

src/runtime/interrupt/interrupt_k210.go
deadprogram: Is this part now working as expected?

deadprogram:
Just tested blinky1 on my Maixbit and worked perfectly first time. Great job @yannishuber wow! I will continue testing.

src/runtime/interrupt/interrupt_k210.go
yannishuber: Yes, interruptions seem to work.

deadprogram:
Tested with the integrations tests from https://github.com/tinygo-org/tinyhci/pull/4 and works as expected.

Looking at the commits, it is not clear which if any could be squashed together.

@yannishuber @aykevl @jaddr2line opinions on that, or should we just rebase into `dev`?

yannishuber:
@deadprogram personally I would squash 779f796f808ef054cf08a18525ccc6f7f65a3db7...2e3871780c64c2232a1d92d87566e0a9c197e7a4 together, but the other commits are meaningful I think. Unless you want me to squash everything to keep just one commit.

deadprogram:
I for sure do not want to squash them all into a single commit. I agree the commits are meaningful, and also I think we want to keep some more granularity to our commit log.

Your plan seems reasonable to me. Still like for @aykevl and @jaddr2line to comment before we merge.

yannishuber:
I noticed that the floating point registers where not saved upon interruptions and fixed it (even if it will soon not be needed anymore as @aykevl pointed out). I also refactored the RISC-V assembly files to be more flexible for the different extensions. Instead of creating a separate file for RV32 and RV64 this is based on the cflags and preprocessor macros.

yannishuber:
I am stuck with the stack overflow prevention. The solution in place now in `target/riscv.ld` to put the stack at the bottom of RAM is not possible because the chip has no embedded flash and thus the `.text` section must be placed at the lowest address in RAM (the bootloader jumps to this address). I thought about using the Physical Memory Protection feature of RISC-V to protect the `.text` section but unfortunately it is not implemented on the K210. I don't know if there is another simple way to prevent this or if we should just stick with the usual stack placement and acknowledge the potential risk of stack overflow? (knowing that the 6M of RAM offer plenty of space before a stack overflow would actually corrupt the heap.)

aykevl:
> I am stuck with the stack overflow prevention.

Don't worry about it. Stack overflow protection used to be effective on Cortex-M chips but isn't as effective now that each goroutine has its own stack. In any case, there is still some sort of protection in that the scheduler will check for overflows when switching the current stack (and so far that has caught many stack overflows in my experience). In any case, this can easily be added at a later date.

Makefile
```diff
@@ -320,6 +324,8 @@ endif
 	@$(MD5SUM) test.hex
 	$(TINYGO) build             -o wasm.wasm -target=wasm               examples/wasm/export
 	$(TINYGO) build             -o wasm.wasm -target=wasm               examples/wasm/main
+	$(TINYGO) build -size short -o test.hex -target=maixbit             examples/blinky1
+	@$(MD5SUM) test.hex
```
aykevl, line 326: Can you keep these smoke tests somewhat bundled together? For example, you could move it next to the hifive1b test above.

src/machine/machine_k210.go
```diff
@@ -0,0 +1,642 @@
...
+
+// Configure this pin with the given configuration.
+// The pin must already be set as GPIO or GPIOHS pin.
+func (p Pin) Configure(config PinConfig) {
+	var input bool
+
+	// Check if the current pin's FPIOA function is either GPIO or GPIOHS.
+	f := p.FPIOAFunction()
+	if f < FUNC_GPIOHS0 || f > FUNC_GPIO7 {
+		return // The pin is not configured as GPIO or GPIOHS.
```
aykevl, line 79: I would expect `Configure` to set up the pin as a GPIO pin, even if it is already configured as something else (so calling `Configure` is a way to undo a different configuration). Would it be possible to reconfigure this pin as a GPIO pin, or is there a reason not to do that?

If the pin is again used by a different peripheral (such as SPI), that peripheral should set the correct pin mode for use by that peripheral.

src/machine/machine_k210.go
```diff
@@ -0,0 +1,642 @@
...
+	}
+}
+
+// Set the pin to high or low.
+func (p Pin) Set(high bool) {
+
+	// Check if the current pin's FPIOA function is either GPIO or GPIOHS.
+	f := p.FPIOAFunction()
+	if f < FUNC_GPIOHS0 || f > FUNC_GPIO7 {
+		return // The pin is not configured as GPIO or GPIOHS.
```
aykevl, line 129: This check is reasonable however (although it's also the caller's responsibility to not call `Set` on a pin that is not configured as GPIO pin).

src/machine/machine_k210.go
```diff
@@ -0,0 +1,642 @@
...
+	fpioaPullUp
+)
+
+// GPIOHS pin interrupt events.
+const (
+	PinRising PinChange = iota + 1
+	PinFalling
+	PinToggle
+	PinHigh
+	PinLow = 8
```
aykevl: This wasn't very clear to me, especially with the regular enum that was used as a bitfield in `SetInterrupt`. Maybe this is more readable:

```suggestion
	PinRising PinChange = 1 << iota
	PinFalling
	PinHigh
	PinLow
	PinToggle = PinRising | PinFalling
```

src/machine/machine_k210.go
```diff
@@ -0,0 +1,642 @@
...
+	}
+	if change&PinFalling != 0 {
+		kendryte.GPIOHS.FALL_IE.SetBits(1 << gpioPin)
+	}
+	if change&PinHigh != 0 {
+		kendryte.GPIOHS.HIGH_IE.SetBits(1 << gpioPin)
+	}
+	if change&PinLow != 0 {
+		kendryte.GPIOHS.LOW_IE.SetBits(1 << gpioPin)
+	}
```
aykevl: What are the semantics for `PinHigh` and `PinLow`? The interrupt keeps being called continuously as long as the line is high (or low)?
I intentionally did not implement this on other chips as it seemed less useful to me and I wasn't sure how it should be defined exactly in a portable way.

src/machine/machine_k210.go
```diff
@@ -0,0 +1,642 @@
...
+
+	handleInterrupt := func(inter interrupt.Interrupt) {
+
+		pin := uint8(inter.GetNumber() - kendryte.IRQ_GPIOHS0)
+
+		if kendryte.GPIOHS.RISE_IE.HasBits(1 << pin) {
+			kendryte.GPIOHS.RISE_IE.ClearBits(1 << pin)
+			kendryte.GPIOHS.RISE_IP.SetBits(1 << pin)
```
aykevl: I suspect there is a race condition here: `kendryte.GPIOHS.RISE_IP.SetBits(1 << pin)` reads the value, modifies it, and writes it back. If a different pin triggers an interrupt the interrupt pending flag may be zeroed again. The same is true for the other three interrupt sources below.

I suspect this chip has atomic register accesses, which perhaps should be used for this purpose. They are not used currently.

yannishuber:
> ...now that each goroutine has its own stack.

@aykevl is that also the case with the coroutines scheduler?

> In any case, this can easily be added at a later date.

Well in that case I think this PR is finally ready.

src/runtime/scheduler_tinygoriscv.S
```diff
@@ -1,32 +1,42 @@
+#ifdef RV64
```
aykevl: Aren't there already preprocessor macros for this? For example, it appears there is `__riscv_mul` (for the M extension) and `__riscv_xlen` which defines XLEN.

src/device/riscv/handleinterrupt.S
```diff
@@ -0,0 +1,130 @@
+#ifdef F_EXTENSION
```
aykevl: Same here, I think you can use `__riscv_flen`. See: https://github.com/riscv/riscv-toolchain-conventions

targets/riscv.ld
aykevl: Is this still necessary now that the K210 has a separate linker script?

aykevl:
You may want to rebase on the dev branch to make sure this PR still works after #1142 has been merged.

src/machine/machine_k210.go
yannishuber: @aykevl Yes it would be possible to configure a non-GPIO pin as GPIO automatically. However I think it is better to assign a none default functions to a pin using the `SetFPIOAFunction()`. Imagine that situation:
```go
machine.D0.Configure(machine.PinConfig{Mode: PinOutput}) // D0 has a JTAG function as default, imagine we configure it automatically as GPIOHS0
machine.D1.Configure(machine.PinConfig{Mode: PinInputPullNone}) // D1 has also a JTAG function as default, but we also configure it as GPIOHS0
machine.D0.High() // This would have no effect now because D1 configured GPIOHS0 as input pin and D0 also has the GPIOHS0 function, so D0 is now an input pin.
```
So this would lead to confusing situations. I know the `SetFPIOAFunction()` adds a step before you can use a certain pin (unless you use a pin with a default function which is already GPIO, see `board_maixbit.go`) but at least there is no risk that two pins have the same function unless you do it on purpose. What do you think?

src/machine/machine_k210.go
yannishuber: > The interrupt keeps being called continuously as long as the line is high (or low)?

Yes, I agree it is not very useful, I will remove it.

src/machine/machine_k210.go
aykevl: Does that mean that with the PR as-is, you also need to call `SetFPIOAFunction` to do anything with these pins? Which implies that examples/blinky1 for example won't work?

aykevl: Also, am I correct that you can actually use the debug pins for something else? That seems dangerous, I'm not sure that should be allowed.

yannishuber: No, if the default FPIOA function for the pin is already GPIO or GPIOHS you do not have to call `SetFPIOAFunction` (the led pins `D12`,`D13`,`D14` are GPIO as default). This would only be needed for the pins `D0` - `D7` or the K210 pins `P36`-`P47` (although they aren't even mapped to external pins on the MAix Bit).

yannishuber: > Also, am I correct that you can actually use the debug pins for something else? That seems dangerous, I'm not sure that should be allowed.

Yes you can, but you would have to do a `SetFPIOAFunction`.

targets/riscv.ld
yannishuber: Not on 32-Bit systems no. Should I change it back to 4 Bytes or should we still respect the default stack alignment of 16 Bytes even if it works with 4 Bytes?

Makefile
yannishuber: done

src/machine/machine_k210.go
yannishuber: done

src/machine/machine_k210.go
yannishuber: Absolutely, I changed it to use the instructions from the atomic extension.

src/runtime/scheduler_tinygoriscv.S
yannishuber: I didn't know this, thank you.

yannishuber:
> You may want to rebase on the dev branch to make sure this PR still works after #1142 has been merged.

I already did that. The changes from #1142 are incorporated.

targets/riscv.ld
aykevl: Ok, for this line (the stack) it makes sense to do the alignment. But for the other alignments below I think they should be reverted back to 4.

yannishuber: done

src/machine/machine_k210.go
aykevl: Ok, I think we can leave it as-is for now. Cross-platform code can simply consider these pins to have a fixed function (like AVR chips for example). I'm not so sure about exposing `SetFPIOAFunction` and `FPIOAFunction`, they add chip specific changes that technically are not necessary (`Configure` calls can do the same, and in fact do just that on most other chips such as SAM D chips).

aykevl (commented):
I think this is ready to merge. I'm a bit unsure about the nonstandard API for pin modes, but I guess it's fine for now as it's still usable this way (but with fixed pins).

deadprogram:
OK, the time has come. Now rebasing into `dev` 

Incredble work @yannishuber and fantastic effort all around. Thank you!

PR #1152, opened 2020-06-03
Add support for SAMD51 with 128 pins

sago35:
This PR adds support for PCxx and PDxx.

SAMD51 series has a chip with 100/128 pins.
More information can be found on P.19 of the Datasheet.

ATSAMD5x DataSheet: https://files.seeedstudio.com/wiki/Wio-Terminal/res/ATSAMD51.pdf

Separated from #1124


deadprogram:
Thank you very much for splitting this out @sago35 makes it a lot easier to understand. Now merging, thanks!

PR #1155, opened 2020-06-03
Standardize SAMD51 UART settings

sago35:
Because SERCOM used by UART is different in each board, I divided the file by this PR.

I've written about which SERCOM to use in the comments below.
https://github.com/tinygo-org/tinygo/pull/1124#issuecomment-633205440

### DIFF
--- a/src/machine/board_feather-m4.go
+++ b/src/machine/board_feather-m4.go
@@ -1,11 +1,6 @@
-// +build sam,atsamd51,feather_m4
+// +build feather_m4
 
 package machine
-
-import (
-	"device/sam"
-	"runtime/interrupt"
-)
 
 // used to reset into bootloader
 const RESET_MAGIC_VALUE = 0xf01669ef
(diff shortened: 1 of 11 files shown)


deadprogram:
In order to make sure that the simulators as used on the TinyGo Playground, it would be better to add "baremetal" files for each board with these additions. 

Here is an example of how this has been done for the Arduino Nano33:

https://github.com/tinygo-org/tinygo/blob/master/src/machine/board_arduino_nano33_baremetal.go

sago35:
I have created a baremetal file.

However, at the moment I can't confirm that the change is correct because playground doesn't work in my local environment (win10).

Are these changes correct?


I thought the rules were as follows.

* board_xxx.go
    * buildtag is only for xxx
   * Only const and usb_VID and usb_PID
* board_xxx_baremetal.go
    * buildtag is the same as the one before splitting
    * Other than the above

deadprogram:
I think you have it correct, and I tested with the PyBoard and seemed to work as expected.

@aykevl any feedback before I squash/merge this?

aykevl:
It's a good idea to move the SERCOM configuration to the board files, as this configuration is board specific.

I'm not sure whether the split to "baremetal" files is necessary. The main purpose is to allow boards to exist on play.tinygo.org, and right now none of the boards are used there. So this split is essentially unused. I'd be fine with merging this as-is, but it might add some unnecessary clutter.

deadprogram:
I think that we very likely will want to add a few of these boards to the Playground.

Now merging to help make that happen. Thanks @sago35 for your work on this contribution!

PR #1158, opened 2020-06-05
compiler: add support for custom code model

yannishuber:
This PR adds a new compiler option to change the code model used by LLVM. This is needed to support the 64-bit RISC-V architecture.

https://github.com/tinygo-org/tinygo/pull/1149


aykevl (approved):
Looks simple enough. Thanks!

PR #1159, opened 2020-06-05
risc-v: add support for 64-bit CPUs

yannishuber:
This PR adds target definitions for both 32 and 64-bit RISC-V chips.

https://github.com/tinygo-org/tinygo/pull/1149

### DIFF
--- a/targets/riscv64.json
+++ b/targets/riscv64.json
@@ -1,0 +1,14 @@
+{
+	"inherits": ["riscv"],
+	"llvm-target": "riscv64--none",
+	"build-tags": ["tinygo.riscv64"],
+	"code-model": "medium",
+	"cflags": [
+		"--target=riscv64--none",
+		"-march=rv64gc",
+		"-mabi=lp64"
+	],
+	"ldflags": [
+		"-melf64lriscv"
+	]
+}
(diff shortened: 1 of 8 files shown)


targets/riscv64.json
```diff
@@ -0,0 +1,14 @@
+{
+	"inherits": ["riscv"],
+	"llvm-target": "riscv64--none",
+	"build-tags": ["tinygo.riscv64"],
+	"code-model": "medium",
```
niaow: Should this be left as default, or do all 64-bit RISC-V systems need this?

yannishuber: No you are right, this belongs in the chip's definition since it may not always be needed.

targets/riscv64.json
```diff
@@ -0,0 +1,13 @@
+{
+	"inherits": ["riscv"],
+	"llvm-target": "riscv64--none",
+	"build-tags": ["tinygo.riscv64"],
+	"cflags": [
+		"--target=riscv64--none",
+		"-march=rv64gc",
```
aykevl, line 7: Assuming RV64GC is a big assumption but I'm fine with it for now (it can always be changed). The riscv32.json makes a similar assumption.

aykevl (approved):
Looks good to me!

PR #1162, opened 2020-06-07
hifive1b: zero PLIC threshold value at startup

### LINKED ISSUE #1128: hifive1b: UART is not receiving data
The UART on the HiFive1b is not receiving data. The program `examples/echo` compiles and flashes, but it does not appear to receive any data. 

The HiFIve1b is able to send data, as the `examples/serial` works.

Another note is that it takes a long time before the serial port comes available.

yannishuber:
The PLIC threshold value must be zeroed at startup to permit all interrupt priorities. Fixes #1128.


deadprogram:
Thank you very much for this fix @yannishuber it is exactly what I needed.

Now merging.

Issue #1163, opened 2020-06-07
gen-device-svd: unexported clusters


yannishuber:
When generating the Go definition for the K210 chip the `gen-device-svd.go` script generates cluster with lowercase names and are therefore not exported.

The result of the generation for the PLIC type for example is the following:
```go
// Platform-Level Interrupt Controller
type PLIC_Type struct {
	PRIORITY       [1024]volatile.Register32
	PENDING        [32]volatile.Register32
	_              [3968]byte
	target_enables [4]struct {
		ENABLE [32]volatile.Register32
	}
	_       [2088448]byte
	targets [4]struct {
		THRESHOLD volatile.Register32
		CLAIM     volatile.Register32
		_         [4084]byte
		_RESERVED volatile.Register32
	}
}
```
`target_enables` and `target` should be uppercase.

aykevl:
The PR has been merged, so closing.

PR #1164, opened 2020-06-07
gen-device-svd: fix lowercase cluster name

yannishuber:
Not sure if this is the right way to do this but this produces the expected output in issue  #1163:
```go
// Platform-Level Interrupt Controller
type PLIC_Type struct {
	PRIORITY       [1024]volatile.Register32
	PENDING        [32]volatile.Register32
	_              [3968]byte
	TARGET_ENABLES [4]struct {
		ENABLE [32]volatile.Register32
	}
	_       [2088448]byte
	TARGETS [4]struct {
		THRESHOLD volatile.Register32
		CLAIM     volatile.Register32
		_         [4084]byte
		_RESERVED volatile.Register32
	}
}
```

### DIFF
--- a/tools/gen-device-svd/gen-device-svd.go
+++ b/tools/gen-device-svd/gen-device-svd.go
@@ -342,6 +342,11 @@ func readSVD(path, sourceURL string) (*Device, error) {
 				firstAddress := clusterRegisters[0].address
 				dimIncrement = int(lastAddress - firstAddress)
 			}
+
+			if !unicode.IsUpper(rune(clusterName[0])) && !unicode.IsDigit(rune(clusterName[0])) {
+				clusterName = strings.ToUpper(clusterName)
+			}
+
 			p.registers = append(p.registers, &PeripheralField{
 				name:        clusterName,
 				address:     baseAddress + clusterOffset,


deadprogram:
Seems legit to me. @aykevl any comment before we merge this?

tools/gen-device-svd/gen-device-svd.go
```diff
@@ -342,6 +342,11 @@ func readSVD(path, sourceURL string) (*Device, error) {
 				firstAddress := clusterRegisters[0].address
 				dimIncrement = int(lastAddress - firstAddress)
 			}
+
+			if !unicode.IsUpper(rune(clusterName[0])) && !unicode.IsDigit(rune(clusterName[0])) {
```
aykevl, line 346: Strictly speaking the cluster name cannot start with a digit, or it would be an invalid identifier in most major programming languages.

aykevl (approved):
Looks good to me!

aykevl:
Two remarks:

* Can you format the commit message like you formatted the PR title? That's the format usually used in TinyGo (and Go).
* There has been an issue like this before. Maybe it's best to write a 'filter' function that does this translation on all names, to fix this more systematically.

yannishuber:
@aykevl Yes sure, I will keep that in mind. About the filter, I will try to come up with something and do a PR in a few days.

PR #1165, opened 2020-06-08
Various bugfixes

aykevl:
This is a collection of three more-or-less independent commits, fixing various issues. I didn't bother with separate PRs because they're so small and trivial.


deadprogram:
OK, merging.

PR #1168, opened 2020-06-10
main: go mod tidy

aykevl:
Clean up the go.mod and go.sum which have gotten a bit messy, and add an extra line to go.sum that keeps reappearing locally for some reason (so it seems important).


deadprogram:
Thanks @aykevl :broom:

PR #1169, opened 2020-06-11
machine/hifive1b: add definitions for UART0 pins

deadprogram:
This PR add definitions for UART0 pins. It is a lot more reliable to use them vs. trying to use the UART connection via the SEGGER interface.

### DIFF
--- a/src/machine/board_hifive1b.go
+++ b/src/machine/board_hifive1b.go
@@ -37,8 +37,8 @@ const (
 
 const (
 	// TODO: figure out the pin numbers for these.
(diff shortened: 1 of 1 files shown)


aykevl:
Does this do anything at all? I don't see any code in src/machine/machine_fe310.go that would use these pin numbers.

deadprogram:
Those are the default pins for the UART which I have not found a way to change. The main reason to add them is really to serve as documentation. I have these pins in use for the TinyHCI tests.

aykevl:
Ah I see. That should be fine.

PR #1171, opened 2020-06-14
 runtime/interrupt: add cross-chip disable/restore interrupt support

aykevl:
This should cover all supported baremetal targets.


niaow (approved):
LGTM

PR #1172, opened 2020-06-15
runtime: add cheap atomic condition variable

niaow:
Earlier on slack @aykevl suggested that something like this would be needed.

Usage example with systick:
```Go
package main

import (
	"device/arm"
	"machine"
	"runtime"
)

var cond runtime.Cond

func main() {

	// timer fires 10 times per second
	arm.SetupSystemTimer(machine.CPUFrequency() / 10)

	for {
		machine.LED.Low()
		cond.Wait()
		machine.LED.High()
		cond.Wait()
	}
}

//export SysTick_Handler
func timer_isr() {
	cond.Notify()
}
```

Current issues making this WIP/not ready:
1. #1142 needs to be merged
2. Where should this go? It was originally in `runtime/interrupt`, but that caused an import cycle.
3. The no-scheduler implementation is still suboptimal (it should wait instead of using a polling loop).

### DIFF
--- a/src/runtime/cond.go
+++ b/src/runtime/cond.go
@@ -1,0 +1,90 @@
+// +build !scheduler.none
+
+package runtime
+
+import (
+	"internal/task"
+	"sync/atomic"
+	"unsafe"
+)
+
+// notifiedPlaceholder is a placeholder task which is used to indicate that the condition variable has been notified.
+var notifiedPlaceholder task.Task
+
+// Cond is a simplified condition variable, useful for notifying goroutines of interrupts.
+type Cond struct {
+	t *task.Task
+}
+
+// Notify sends a notification.
+// If the condition variable already has a pending notification, this returns false.
+func (c *Cond) Notify() bool {
+	for {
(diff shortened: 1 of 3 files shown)


niaow:
> The no-scheduler implementation is still suboptimal (it should wait instead of using a polling loop).

This part has now been resolved.

niaow:
> Where should this go?

@aykevl any opinion?

deadprogram:
Reminder to @aykevl and @niaow about this PR. :smiley_cat:

niaow:
This _should_ work decently now. However, my computer isn't recognizing my ARM boards right now so I have not been able to test on hardware.

deadprogram:
Tested with the provided example on ATSAMD21 board (Circuit Playground Express) and then also on nrf52840 (Circuit Playground Bluefruit) and both worked exactly as expected.

@aykevl any feedback before this gets merged?

deadprogram:
We really need this for something specific. Code looks good, and works as promised, so now merging. Thanks @niaow for this feature!

PR #1173, opened 2020-06-16
Added Adafruit PyGamer Target

APDevice:
I have added support for the Adafruit PyGamer. The board is nearly identical to the PyBadge, just with a few changes and additions. Chiefly the directional buttons have been replaced with an analog stick and there is an microSD card slot for additional storage.

This is my first attempt to add a device, though I did test it on the hardware to make sure that I had assigned the analog stick correctly

Schematic: https://learn.adafruit.com/assets/76005


deadprogram:
Hi @APDevice first of all, thanks for the contribution.

I have changed the branch to be against the `dev` branch as mentioned here: https://github.com/tinygo-org/tinygo/blob/master/CONTRIBUTING.md#how-to-use-our-github-repository

I do not think you meant to delete the `.gitignore` file, did you? We need that file, you probably want to remove that commit and then force push back to your feature branch.

Thank you!

APDevice:
> Hi @APDevice first of all, thanks for the contribution.
> 
> I have changed the branch to be against the `dev` branch as mentioned here: https://github.com/tinygo-org/tinygo/blob/master/CONTRIBUTING.md#how-to-use-our-github-repository
> 
> I do not think you meant to delete the `.gitignore` file, did you? We need that file, you probably want to remove that commit and then force push back to your feature branch.
> 
> Thank you!

AH yes, sorry. I got a little confused with that. I believe I have fixed it now!

src/machine/board_pygamer.go
```diff
@@ -0,0 +1,147 @@
...
+
+// UART2 var is on SERCOM0, defined in atsamd51.go
+
+// I2C pins
+const (
+	SDA_PIN = PA12 // SDA: SERCOM2/PAD[0]
+	SCL_PIN = PA13 // SCL: SERCOM2/PAD[1]
+)
+
+// I2C on the ItsyBitsy M4.
```
deadprogram: That should say "PyGamer" instead of  "ItsyBitsy M4" right?

deadprogram:
Please also add a "smoke test" to the Makefile like this: https://github.com/tinygo-org/tinygo/blob/dev/Makefile#L296-L297

Thank you!

src/machine/board_pygamer.go
APDevice: Oh, I hadn’t even touched that; Since this is a variant of the pyBadge, I just copied and modified that. Carrying over all the quirks in the process, it seems. I can go ahead and fix it here though.

APDevice:
So I added the smoke test, but now it's saying it's conflicted with the dev makefile, and the error just suggests that the conflict is with a blank line. I’m not sure how to resolve the issue.

deadprogram:
If you had originally created your feature branch against the `dev` branch instead of the `master` branch, you probably would not have this problem.

You can resolve by changing all of this:

```
<<<<<<< master
	$(TINYGO) build -size short -o test.hex -target=pygamer             examples/blinky1
	@$(MD5SUM) test.hex
=======
>>>>>>> dev
```

to this:

```
	$(TINYGO) build -size short -o test.hex -target=pygamer             examples/blinky1
	@$(MD5SUM) test.hex
```

You could use the web editor or however you normally edit code.

Hope that helps!

APDevice:
Well the confusing thing is that that’s not in the file. I even double checked. 

Ok, so I merged my dev with my master. It’s sloppy I realize, but it seems to resolve the issue.

deadprogram:
Thank you very  much for working on this @APDevice I will now squash/merge into `dev`. Thanks again!

PR #1174, opened 2020-06-17
machine/samd21: add adc settings

sago35:
Added some pin settings.

ATSAMD2x datasheet: https://files.seeedstudio.com/wiki/Seeeduino-XIAO/res/ATSAMD21G18A-MU-Datasheet.pdf


deadprogram:

PR #1176, opened 2020-06-18
machine: add DMA support to spi tx

sago35:
I made the call to SetupDMA() to use DMA to send.
If no call is made, send it without DMA.

At the moment, it is wioterminal-only and most of the possible values cannot be set.
The source code is messy and not clean.
However, I think the future of effective use of the DMA can be examined on the basis of this PR.

* before (w/o DMA) : 40 - 42 fps
    * 5c85bbc6036ba84896f1332b6b52120c1107fb40
* after (with DMA) : 62 - 64 fps
    * 169aa2ac3675be0ff5b55d4b994e67cfc41d2fdd


sago35:
ref: https://github.com/tinygo-org/drivers/pull/153#issuecomment-645390455

aykevl:
I think we'll always want to use DMA. That is an easier to use API and likely faster overall.

sago35:
I'm going to try and make it.

sago35:
I plan to switch to the method used in the following PR.

#3985

PR #1182, opened 2020-06-21
device/arm: do not mask fault handlers in critical sections

niaow:
It appears that we were incorrectly leaving faults masked permanently. On Cortex-M4 systems, this led to the microcontroller becoming non-responsive and passing garbage to the debugger. As far as I can tell, we aren't actually faulting anywhere, so this is probably a hardware bug. We should figure out what is going wrong, but until we do we should probably stop bricking microcontrollers.

### DIFF
--- a/src/device/arm/arm.go
+++ b/src/device/arm/arm.go
@@ -196,7 +196,7 @@ func SetPriority(irq uint32, priority uint32) {
 func DisableInterrupts() uintptr {
 	return AsmFull(`
 		mrs {}, PRIMASK
-		cpsid if
+		cpsid i
 	`, nil)
 }
 


deadprogram:
We are not using FIQ requests anyhow, so as you said this seems like it was incorrect. Since it also corrects the issue and passes all the tests, I am merging. Great job finding this @jaddr2line

aykevl:
I checked this and it is indeed a bug. I am not sure why disabling faults would lock the MCU but fault handling (HardFault etc) should certainly not be disabled.

I probably copied this code from somewhere, my bad for not checking what those instructions do exactly.

aykevl:
> We are not using FIQ requests anyhow

Do you mean fast interrupts, like on ARM cores? Those do not exist on the Cortex-M4. We do handle faults, by printing a `hard fault` message.

niaow:
oh wait that is fiq not fault i think

niaow:
Huh. There are actually 2 instructions with the same name, and this lines up with some stuff i saw. . .

PR #1183, opened 2020-06-22
machine/samd21: basic implementation for DAC

deadprogram:
This PR adds a very basic DAC implementation for the SAMD21. It does not address any timer/DMA functionality needed to create a complex waveform, that will be for future PRs.

### DIFF
--- a/src/examples/dac/dac.go
+++ b/src/examples/dac/dac.go
@@ -1,0 +1,38 @@
+// Example using the DAC
+//
+package main
+
+import (
+	"machine"
+	"time"
+)
+
+func main() {
+	enable := machine.PA30
+	enable.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	enable.Set(true)
+
+	speaker := machine.A0
+	speaker.Configure(machine.PinConfig{Mode: machine.PinOutput})
+
+	machine.DAC0.Configure(machine.DACConfig{})
+
+	data := []uint16{4096, 2058, 1024, 512, 256, 128}
+
+	for {
+		for _, val := range data {
+			play(val)
+			time.Sleep(500 * time.Millisecond)
+		}
+	}
+}
+
+func play(val uint16) {
+	for i := 0; i < 100; i++ {
+		machine.DAC0.WriteData(val)
+		time.Sleep(2 * time.Millisecond)
+
+		machine.DAC0.WriteData(0)
+		time.Sleep(2 * time.Millisecond)
+	}
+}
--- a/src/examples/dac/dac.go
+++ b/src/examples/dac/dac.go
@@ -1,5 +1,7 @@
-// Example using the DAC
+// Simplistic example using the DAC on the Circuit Playground Express.
 //
+// To actually use the DAC for producing complex waveforms or samples requires a DMA
+// timer-based playback mechanism which is beyond the scope of this example.
 package main
 
 import (
--- a/src/examples/dac/dac.go
+++ b/src/examples/dac/dac.go
@@ -19,7 +19,7 @@ func main() {
 
 	machine.DAC0.Configure(machine.DACConfig{})
 
-	data := []uint16{4096, 2058, 1024, 512, 256, 128}
+	data := []uint16{32768, 8192, 2048, 512, 0}
 
 	for {
 		for _, val := range data {
@@ -31,10 +31,10 @@ func main() {
 
 func play(val uint16) {
 	for i := 0; i < 100; i++ {
-		machine.DAC0.WriteData(val)
+		machine.DAC0.Set(val)
 		time.Sleep(2 * time.Millisecond)
 
-		machine.DAC0.WriteData(0)
+		machine.DAC0.Set(0)
 		time.Sleep(2 * time.Millisecond)
 	}
 }
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -2251,8 +2251,6 @@ func (dac DAC) Configure(config DACConfig) {
 func (dac DAC) WriteData(value uint16) error {
 	sam.DAC.DATA.Set(value & 0x3FF)
 	syncDAC()
-	sam.DAC.CTRLA.SetBits(sam.DAC_CTRLA_ENABLE)
-	syncDAC()
 	return nil
 }
 
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -2249,25 +2249,10 @@ func (dac DAC) Configure(config DACConfig) {
 
 // WriteData writes a single value to the DAC.
 func (dac DAC) WriteData(value uint16) error {
-	syncDAC()
 	sam.DAC.DATA.Set(value & 0x3FF)
 	syncDAC()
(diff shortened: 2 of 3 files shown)


src/examples/dac/dac.go
```diff
@@ -0,0 +1,40 @@
...
+
+	for {
+		for _, val := range data {
+			play(val)
+			time.Sleep(500 * time.Millisecond)
+		}
+	}
+}
+
+func play(val uint16) {
```
aykevl, line 32: Why 'play'? Just wondering why this particular name, it makes me think of something related to audio, which this isn't.

src/machine/machine_atsamd21.go
```diff
@@ -2209,3 +2209,69 @@ func ResetProcessor() {
...
+	sam.DAC.DATA.Set(value & 0x3FF)
+	syncDAC()
+	sam.DAC.CTRLA.SetBits(sam.DAC_CTRLA_ENABLE)
+	syncDAC()
+	return nil
+}
+
+// Write is intended to satisfy the Writer interface.
+// This current implementation is probably not very useful.
+func (dac DAC) Write(p []uint16) (n int, err error) {
```
aykevl: This does not satisfy the `Writer` interface, in fact it breaks this support because the name is the same while the signature is different. Maybe it's better to leave it out until a good use case can be found.

src/machine/machine_atsamd21.go
```diff
@@ -2209,3 +2209,69 @@ func ResetProcessor() {
...
+	for _, val := range p {
+		dac.WriteData(val)
+	}
+	return len(p), nil
+}
+
+// Close the DAC.
+func (dac DAC) Close() error {
```
aykevl: Probably unnecessary when `Write` doesn't exist (and even then, not sure how it's useful).

src/machine/machine_atsamd21.go
```diff
@@ -2209,3 +2209,69 @@ func ResetProcessor() {
...
+	}
+
+	// enable
+	sam.DAC.CTRLB.Set(sam.DAC_CTRLB_EOEN | sam.DAC_CTRLB_REFSEL_AVCC)
+	sam.DAC.CTRLA.Set(sam.DAC_CTRLA_ENABLE)
+}
+
+// WriteData writes a single value to the DAC.
+func (dac DAC) WriteData(value uint16) error {
+	syncDAC()
```
aykevl, line 2254: This first `syncDAC` should not be necessary, if the DAC is synchronized after every register access. Maybe a `syncDAC` is missing at the end of the `Configure` method?

aykevl (commented):
So, this DAC basically creates a particular voltage (not PWM) on a pin? That's interesting, I haven't seen that often. I wonder if there are other chips with similar functionality, to match the API?

src/examples/dac/dac.go
deadprogram: Actually, this example does play an audio tone at 6 different amplitudes using the Circuit Playground Express built-in speaker.

src/machine/machine_atsamd21.go
deadprogram: Removed.

src/machine/machine_atsamd21.go
deadprogram: Removed.

src/machine/machine_atsamd21.go
deadprogram: You were correct. Removed.

src/examples/dac/dac.go
aykevl: Huh that's strange. What kind of speaker is that? Wouldn't a speaker normally need a sine wave (or something) to play a tone instead of a particular voltage?

deadprogram: The loop turns the power on and off acting like a very simple wave ala PWM. Of course just setting a single voltage would not produce any sound! :)

src/examples/dac/dac.go
```diff
@@ -0,0 +1,40 @@
...
+	enable := machine.PA30
+	enable.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	enable.Set(true)
+
+	speaker := machine.A0
+	speaker.Configure(machine.PinConfig{Mode: machine.PinOutput})
+
+	machine.DAC0.Configure(machine.DACConfig{})
+
+	data := []uint16{4096, 2058, 1024, 512, 256, 128}
```
sago35: Since samd21 is a 10-bit DAC, numbers above 1024 are meaningless.
But I want to use the same test with samd51, so I think it is better to leave it as it is.

Samd51 is a 12-bit DAC and its maximum value is 4095.
In addition, since 2058 also looks wrong, isn't 2048 the correct value?

In other words, I think the following is correct.

```go
data := []uint16{4095, 2048, 1024, 512, 256, 128}
```

src/machine/machine_atsamd21.go
```diff
@@ -2209,3 +2209,54 @@ func ResetProcessor() {
...
+	// enable
+	sam.DAC.CTRLB.Set(sam.DAC_CTRLB_EOEN | sam.DAC_CTRLB_REFSEL_AVCC)
+	sam.DAC.CTRLA.Set(sam.DAC_CTRLA_ENABLE)
+}
+
+// WriteData writes a single value to the DAC.
+func (dac DAC) WriteData(value uint16) error {
+	sam.DAC.DATA.Set(value & 0x3FF)
+	syncDAC()
+	sam.DAC.CTRLA.SetBits(sam.DAC_CTRLA_ENABLE)
```
sago35: From reading the datasheet, I don't think the following is necessary
It actually works if you comment out.

```go
	sam.DAC.CTRLA.SetBits(sam.DAC_CTRLA_ENABLE)
	syncDAC()
```

deadprogram: You are quite correct @sago35 now removed. Thank you.

deadprogram:
I think all feedback has now been addressed in this PR.

aykevl:
The implementation looks good to me. However, I have some concerns about the API.

```go
func (dac DAC) WriteData(value uint16)
```

Similar to the ADC, I would suggest this value should use the full 16-bit scale to be more portable across chips. As stated above, the number of bits varies by chip, but it doesn't seem likely it will be more than 16 bits on common chips. For comparison, the DAC in the stm32f103 is also 12 bits, like the samd51.
I'm assuming the DAC by default uses the VCC (usually 3.3V) as reference voltage for the maximum value?

I'm also not sure about the name, `WriteData` seems a bit generic. What about `Set` for example, analogous to the ADC `Get` function?

deadprogram:
I have now renamed the function to `Set(value uint16)` and also changed to expect 16-bit values and scale the input value. This does seems like the correct way to handle, @aykevl thanks for the suggestions.

@sago35 you will have to change your PR #1198 a little more than expected once this one has been merged, assuming that the changes are acceptable to everyone.

aykevl (approved):
LGTM, thanks!

deadprogram:
OK commits squashed to keep it clean, and merged. Thanks @aykevl and @sago35 for your reviews. Also @sago35 you are now unblocked to complete #1198

PR #1188, opened 2020-06-24
machine/stm32f4disco: add target file for updated board

deadprogram:
This PR adds an additional target file for updated versions of the STM32F4 Discovery boards that have the updated ST-Link-2-1 onboard.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -259,6 +259,8 @@ smoketest:
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=stm32f4disco        examples/blinky2
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=stm32f4disco-1       examples/blinky1
+	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=circuitplay-bluefruit examples/blinky1
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=circuitplay-express examples/i2s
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -259,6 +259,10 @@ smoketest:
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=stm32f4disco        examples/blinky2
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=stm32f4disco-1       examples/blinky1
+	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=stm32f4disco-1       examples/blinky2
+	@$(MD5SUM) test.hex
```
aykevl: Only one smoke test should be fine. The only thing that an extra smoke test tests is that the `LEDx` constants exist, which is not something that is likely to break unintentionally (and it slows down tests).

aykevl (commented):
That's unfortunate. There is no interface that will work for both (`stlink-v2-1` or `stlink-v2`)?

deadprogram:
I was not able to get openocd to recognize the onboard jlink adaptor unless I had the full name with the `-1` suffix.

Makefile
deadprogram: Removed extra smoketest.

aykevl:
That is unfortunate. And the reverse doesn't work either, driving the older boards with `stlink-v2-1`?

deadprogram:
I do not have the older one to test with, so not sure on that.

PR #1196, opened 2020-06-26
Implement defers

waj334:


compiler/defer.go
```diff
@@ -151,9 +155,87 @@ func (b *builder) createDefer(instr *ssa.Defer) {
 		values = append(values, context)
 		valueTypes = append(valueTypes, context.Type())
 
+	} else if  builtin, ok := instr.Call.Value.(*ssa.Builtin); ok {
+		var funcName string
+		switch builtin.Name() {
+		case "close":
+			funcName = "chanClose"
+		default:
+			b.addError(instr.Pos(), fmt.Sprint("TODO: Implement defer for ", builtin.Name()))
```
aykevl: No need for `fmt.Sprint`, you can simply add the strings together.
```suggestion
			b.addError(instr.Pos(), "todo: Implement defer for " + builtin.Name())
```

compiler/defer.go
```diff
@@ -151,9 +155,87 @@ func (b *builder) createDefer(instr *ssa.Defer) {
...
+			funcValue = b.getValue(expr)
+			switch ty := expr.X.Type().(type) {
+			case *types.Pointer:
+				sig = ty.Elem().Underlying().(*types.Signature)
+			default:
+				sig = ty.Underlying().(*types.Signature).Results().At(0).Type().Underlying().(*types.Signature)
+			}
+		}
+
+		if funcValue.IsNil() == false && sig != nil {
```
aykevl: Don't do `== false`, just use the boolean.
```suggestion
		if !funcValue.IsNil() && sig != nil {
```

compiler/defer.go
```diff
@@ -151,9 +155,87 @@ func (b *builder) createDefer(instr *ssa.Defer) {
...
+			sig = expr.Call.Value.Type().Underlying().(*types.Signature).Results().At(0).Type().Underlying().(*types.Signature)
+		case *ssa.UnOp:
+			funcValue = b.getValue(expr)
+			switch ty := expr.X.Type().(type) {
+			case *types.Pointer:
+				sig = ty.Elem().Underlying().(*types.Signature)
+			default:
+				sig = ty.Underlying().(*types.Signature).Results().At(0).Type().Underlying().(*types.Signature)
+			}
+		}
```
aykevl: You're reimplementing part of the compiler here. Expressions should not be evaluated here, use `b.getValue` for that and store the result (which should be a func value). The code as-is will likely miss some edge cases.

aykevl (commented):
You will need to update testdata/calls.txt with the output of running testdata/calls.go for the tests to pass.

compiler/defer.go
waj334: From my understanding, I can't call a func value directly. So I needed to obtain the call's signature in order to get the function pointer needed for b.createCall later in createRunDefers. Is there another way to accomplish this?

EDIT:
Immediately answered my own question

compiler/defer.go
```diff
@@ -151,9 +154,60 @@ func (b *builder) createDefer(instr *ssa.Defer) {
...
+			valueTypes = append(valueTypes, llvmParam.Type())
+		}
+
 	} else {
-		b.addError(instr.Pos(), "todo: defer on uncommon function call type")
-		return
+		funcValue := b.getValue(instr.Call.Value)
+		funcValueType := b.getLLVMRuntimeType("funcValue")
+
+		if !funcValue.IsNil() && funcValue.Type() == funcValueType {
```
aykevl: This check is not necessary, and in fact should be removed. There are two ways a func value is represented in the IR, as a `runtime.funcValue` or as a tuple with a function pointer and context pointer. Most code is oblivious to the difference, and this code should be too. In fact, you shouldn't try to inspect this func value, instead you should just pass along this value to the place where it is called and call `decodeFuncValue` to get the raw function pointer and context parameter (which you can then call as usual).

waj334:
> You will need to update testdata/calls.txt with the output of running testdata/calls.go for the tests to pass.

I checked the output against the calls.txt file and they appear to be the same. I'm sure I'm not crazy, but I think encoding might be the problem perhaps???

compiler/compiler.go
```diff
@@ -74,7 +74,14 @@ type builder struct {
 	deferFuncs        map[*ir.Function]int
 	deferInvokeFuncs  map[string]int
 	deferClosureFuncs map[*ir.Function]int
+	deferExprFuncs    map[interface{}]int
```
aykevl:
```suggestion
	deferExprFuncs    map[ssa.Value]int
```

compiler/compiler.go
```diff
@@ -74,7 +74,14 @@ type builder struct {
...
 	selectRecvBuf     map[*ssa.Select]llvm.Value
+	deferBuiltinFuncs map[interface{}]deferBuiltin
```
aykevl:
```suggestion
	deferBuiltinFuncs map[ssa.Value]deferBuiltin
```

aykevl (commented):
Looks good to me, except these things. The added type safety may be useful.

aykevl:
Thank you! The `defer` keyword is now almost entirely implemented.

PR #1198, opened 2020-06-28
machine/samd51: add DAC support

sago35:
This PR adds a very basic DAC implementation for the SAMD51.
It is the SAMD51 version of the implementation created in #1183.

There are two DACs in the SAMD51, but so far I've only implemented one.

### DIFF
--- a/src/machine/machine_atsamd51.go
+++ b/src/machine/machine_atsamd51.go
@@ -2297,3 +2297,60 @@ func ResetProcessor() {
 
 	arm.SystemReset()
 }
+
+// DAC on the SAMD51.
+type DAC struct {
+}
+
+var (
+	DAC0 = DAC{}
+)
+
+// DACConfig placeholder for future expansion.
+type DACConfig struct {
+}
+
+// Configure the DAC.
+// output pin must already be configured.
+func (dac DAC) Configure(config DACConfig) {
+	// Turn on clock for DAC
+	sam.MCLK.APBDMASK.SetBits(sam.MCLK_APBDMASK_DAC_)
+
+	// Use Generic Clock Generator 4 as source for DAC.
+	sam.GCLK.PCHCTRL[42].Set((sam.GCLK_PCHCTRL_GEN_GCLK4 << sam.GCLK_PCHCTRL_GEN_Pos) | sam.GCLK_PCHCTRL_CHEN)
+	for sam.GCLK.SYNCBUSY.HasBits(sam.GCLK_SYNCBUSY_GENCTRL_GCLK4 << sam.GCLK_SYNCBUSY_GENCTRL_Pos) {
+	}
+
+	// reset DAC
+	sam.DAC.CTRLA.Set(sam.DAC_CTRLA_SWRST)
+
+	// wait for reset complete
+	for sam.DAC.CTRLA.HasBits(sam.DAC_CTRLA_SWRST) {
+	}
+	for sam.DAC.SYNCBUSY.HasBits(sam.DAC_SYNCBUSY_SWRST) {
+	}
+
+	// enable
+	sam.DAC.CTRLB.Set(sam.DAC_CTRLB_REFSEL_VREFPU)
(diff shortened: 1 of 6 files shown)


sago35:
I think it would be better to merge this PR in the following order: 1.

1. merge #1183
2. rebase
Update example

sago35:
To work with pyportal, you need to make the following changes

```go
	enable := machine.SPK_SD
```

src/machine/machine_atsamd51.go
```diff
@@ -2297,3 +2297,60 @@ func ResetProcessor() {
...
+	sam.DAC.CTRLA.Set(sam.DAC_CTRLA_SWRST)
+
+	// wait for reset complete
+	for sam.DAC.CTRLA.HasBits(sam.DAC_CTRLA_SWRST) {
+	}
+	for sam.DAC.SYNCBUSY.HasBits(sam.DAC_SYNCBUSY_SWRST) {
+	}
+
+	// enable
+	sam.DAC.CTRLB.Set(sam.DAC_CTRLB_REFSEL_VREFPU)
```
sago35: When 0 is written to DATA0 while DAC_CTRLB_REFSEL_VDDANA is set, the voltage becomes VREF/2.
I think this is the same problem as the following, but I don't know what is happening.
https://www.avrfreaks.net/comment/2620686#comment-2620686

deadprogram: But it works as expected with `sam.DAC_CTRLB_REFSEL_VREFPU`?

sago35: @deadprogram 
Yes.
The voltage at 4095 is a little low, but it's almost exactly what you want.
On the other hand, VDDANA seems to work like differential mode.

image

deadprogram: Well, the proof is in the signal analysis above. Good image and work on that @sago35 I am satisfied it is correct!

deadprogram:
Hello @sago35 this PR is now ready for revision to match the updated DAC API, please.

sago35:
@deadprogram 
In pyportal, values below 0x2000 are almost inaudible.
I would like to fix the sample as follows, is this OK in circuitplay-express?

```
diff --git a/src/examples/dac/dac.go b/src/examples/dac/dac.go
index cc84082..570e349 100644
--- a/src/examples/dac/dac.go
+++ b/src/examples/dac/dac.go
@@ -15,7 +15,7 @@ func main() {
 
        machine.DAC0.Configure(machine.DACConfig{})
 
-       data := []uint16{32768, 8192, 2048, 512, 0}
+       data := []uint16{0xFFFF, 0x8000, 0x4000, 0x2000, 0x1000}
 
        for {
                for _, val := range data {
```

src/machine/machine_atsamd51.go
```diff
@@ -2331,7 +2331,7 @@ func (dac DAC) Configure(config DACConfig) {
 	}
 
 	// enable
-	sam.DAC.CTRLB.Set(sam.DAC_CTRLB_REFSEL_VREFPU)
+	sam.DAC.CTRLB.Set(sam.DAC_CTRLB_REFSEL_VREFPU << sam.DAC_CTRLB_REFSEL_Pos)
```
sago35, line 2334: Register was in the wrong position, so I fixed it.
It does not work on VDDANA at this time.
It appears to be working with VREFPU or VREFPB.

sago35: VREFPU
image

VDDANA
I don't know why, but the VDDANA waveform looks like this
image

deadprogram: According to this post https://www.avrfreaks.net/comment/2620671#comment-2620671

> 1.8.2 DAC
> The selection of VDDANA as the DAC reference in DAC.CTRLB.REFSEL is non-functional.
> Workaround
> The VDDANA must be connected externally to a VREF pin and DAC.CTRLB.VREFAU must be selected.

Not sure why those spikes occur, but in any case seems like not expected to work anyhow.

deadprogram:
OK, hopefully last change for this PR. Can you please add a smoke test for the dac example with `-target pyportal`?

sago35:
I had forgotten about smoke test.
I just added it.

deadprogram:
Looks great, thank you so very much for another wonderful contribution, @sago35 

Now squashing/merging to keep commit log as clean as possible. Thanks again!

PR #1203, opened 2020-06-30
avr: fix target triple

aykevl:
It was `avr-atmel-none`, which is incorrect. It must be `avr-unknown-unknown`.

Additionally, there is no reason to specify the target triple per chip, it can be done for all AVR chips at once as it doesn't vary like Cortex-M chips.

---

Comment that triggered this PR: https://github.com/rust-lang/rust/issues/44052#issuecomment-649807495

### DIFF
--- a/targets/atmega1284p.json
+++ b/targets/atmega1284p.json
@@ -1,6 +1,5 @@
 {
     "inherits": ["avr"],
-    "llvm-target": "avr-atmel-none",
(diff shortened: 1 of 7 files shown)


deadprogram:
Tested using Arduino Uno and DIgispark, worked as expected. Now merging, thanks for this great research @aykevl and kudos to the Rust community for all their related LLVM work.

PR #1211, opened 2020-07-05
machine: fix atmega2560 mapping for pins D2 and D5

brandentimm:
This PR corrects the port mappings for pins D2 and D5 on atmega2560.  Tested working on my board (elegoo mega2560 R3).


deadprogram:
Thank you for the fix, @brandentimm now merging.

PR #1213, opened 2020-07-06
docs: add submodule update step prior to 'make release'

brandentimm:
The instructions for 'make release' were missing a step to update submodules, which would cause the task to fail.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -82,8 +82,11 @@ The result should not contain libclang or libLLVM.
 
 Now that we have a working static build, it's time to make a release tarball:
 
-    git submodule update --init
     make release
+
+If you did not clone the repository with the `--recursive` option, you will get errors until you initialize the project submodules:
+
+    git submodule update --init
 
 The release tarball is stored in build/release.tar.gz, and can be extracted with
 the following command (for example in ~/lib):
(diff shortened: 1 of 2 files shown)


deadprogram:
Hi @brandentimm thanks for the PR.

I think maybe a better place to do this might be as part of this step: https://github.com/tinygo-org/tinygo/blob/bf3f195f6667a826f83d0b113c9cfb6164c1f723/BUILDING.md#download-the-source

What do you think?

brandentimm:
Makes sense, I'll update it and re-submit.

brandentimm:
@deadprogram well now I'm seeing what I initially missed in the section you referenced, the note about using --recursive when cloning.  So I think adding an additional note in there seems redundant.  

Perhaps instead we add a note in the 'make release' section calling out that if you haven't cloned recursively, you'll need to run  'git submodule update --init'?

deadprogram:
I think by the time people see that notice about `--recursive` they will have already cloned. So maybe we should say if you have not used that flag to run the submodule command.

aykevl:
I think this is a useful addition, so can be merged from my point of view.

brandentimm:
I've clarified the need for submodule update in the release step.

deadprogram:
Thank you for helping reduce friction for people getting stuck on that @brandentimm now going to squash/merge.

Thanks again!

PR #1215, opened 2020-07-07
risc-v: disable linker relaxations during gp initialization

yannishuber:
Linker relaxations should be disabled when initializing the global pointer to avoid ending up with 
```asm
mv gp, gp
```
LLVM does not seem to use linker relaxations at the moment but this could change in the future leading to unexpected results.


aykevl:
Looks reasonable, thank you! The last time I looked lld indeed did not support relaxation yet (which is not a big deal on the currently supported RISC-V chips).

PR #1216, opened 2020-07-07
all: changeover to eliminate all direct use of master/slave terminology

deadprogram:
This PR is to changeover to eliminate all direct use of master/slave terminology both for the Git repo, and also the TinyGo peripherals. See https://www.oshwa.org/a-resolution-to-redefine-spi-signal-names

We will also need to change the `master` branch name to `release`, and change the default branch in Github.


deadprogram:
@johanbrandhorst any way to change the branch used by the repo used for the Arch release? https://github.com/tinygo-org/tinygo/blob/dev/.circleci/config.yml#L368

johanbrandhorst:
> @johanbrandhorst any way to change the branch used by the repo used for the Arch release? https://github.com/tinygo-org/tinygo/blob/dev/.circleci/config.yml#L368

Couldn't find any information about this so I've posted a question on the AUR forum: https://bbs.archlinux.org/viewtopic.php?pid=1914409#p1914409.

deadprogram:
Thanks!

aykevl:
I could be wrong, but the way I understood https://www.oshwa.org/a-resolution-to-redefine-spi-signal-names was that the new signal names are (usually) SDO and SDI (similar to TX and RX). Additionally, I find `SDO` and `SDI` easier to understand.

johanbrandhorst:
> > @johanbrandhorst any way to change the branch used by the repo used for the Arch release? https://github.com/tinygo-org/tinygo/blob/dev/.circleci/config.yml#L368
> 
> Couldn't find any information about this so I've posted a question on the AUR forum: https://bbs.archlinux.org/viewtopic.php?pid=1914409#p1914409.

It doesn't seem to be possible right now, but I've asked if it can be changed in the future https://bbs.archlinux.org/viewtopic.php?pid=1914523#p1914523.

deadprogram:
@aykevl I like the SDO/SDI better as well. Changes made and force pushed.

@johanbrandhorst thanks for checking into it.

deadprogram:
I think that I have now handled all of the boards currently in the `dev` branch. This PR should now be ready.

sevki:
@deadprogram could you update the default branch?
Just spent an ungodly amount of time compiling an obsolete branch :joy:

deadprogram:
@sevki see https://github.com/tinygo-org/tinygo/blob/master/CONTRIBUTING.md#how-to-use-our-github-repository

src/machine/i2s.go
```diff
@@ -41,14 +41,14 @@ const (
...
-	Stereo            bool
+	SCK                Pin
+	WS                 Pin
+	SD                 Pin
+	Mode               I2SMode
+	Standard           I2SStandard
+	ClockSource        I2SClockSource
+	DataFormat         I2SDataFormat
+	AudioFrequency     uint32
+	PrimaryClockOutput bool
```
aykevl: You use `I2SModeController` and `I2SModePeripheral` above (which seem fine, although "source"/"destination" or "sender"/"receiver" might be better in the case of I2S which is in a single direction). This is called Primary, which does not seem consistent.

src/runtime/interrupt/interrupt_gameboyadvance.go
```diff
@@ -8,9 +8,9 @@ import (
 )
 
 var (
-	regInterruptEnable       = (*volatile.Register16)(unsafe.Pointer(uintptr(0x4000200)))
-	regInterruptRequestFlags = (*volatile.Register16)(unsafe.Pointer(uintptr(0x4000202)))
-	regInterruptMasterEnable = (*volatile.Register16)(unsafe.Pointer(uintptr(0x4000208)))
+	regInterruptEnable        = (*volatile.Register16)(unsafe.Pointer(uintptr(0x4000200)))
+	regInterruptRequestFlags  = (*volatile.Register16)(unsafe.Pointer(uintptr(0x4000202)))
+	regInterruptPrimaryEnable = (*volatile.Register16)(unsafe.Pointer(uintptr(0x4000208)))
```
aykevl: The "master enable" bit controls all interrupts at once, like cli/sei on AVR for example. I don't think "primary" is the correct term to describe this.

I'm not sure what to call this. Maybe it's better to keep calling it "master enable", because that is what upstream documentation uses: https://www.akkit.org/info/gbatek.htm#gbainterruptcontrol. Leaving it shouldn't matter much as it's not exposed in an API.

src/machine/machine_stm32_spi.go
```diff
@@ -56,20 +56,20 @@ func (spi SPI) Configure(config SPIConfig) {
...
 	spi.Bus.CR1.Set(conf)
 
 	// init pins
-	if config.SCK == 0 && config.MOSI == 0 && config.MISO == 0 {
+	if config.SCK == 0 && config.SDO == 0 && config.SDI == 0 {
 		config.SCK = SPI0_SCK_PIN
-		config.MOSI = SPI0_MOSI_PIN
-		config.MISO = SPI0_MISO_PIN
+		config.SDO = SPI0_SDO_PIN
+		config.SDI = SPI0_SDI_PIN
```
aykevl, line 72: Is this still correct in peripheral mode? In peripheral mode, SDO/SDI are reversed while MOSI/MISO aren't (or the other way around, depending on how you look at it).

aykevl (commented):
While I agree this is a good change in principle, we should be careful to use the correct terminology. I'm worried some changes may make things less clear. We can't really do this kind of thing again without good reason because it will break existing programs.

src/runtime/interrupt/interrupt_gameboyadvance.go
niaow: Maybe `regGlobalInterruptEnable` if we still wanted to change it?

aykevl: That's a good idea, thanks!

src/machine/machine_stm32_spi.go
deadprogram: We do not current have any support for peripheral mode on any board anyhow, so once we do, we can address that.

src/machine/i2s.go
deadprogram: Changed to "source" and "receiver".

src/runtime/interrupt/interrupt_gameboyadvance.go
deadprogram: Done.

deadprogram:
Updated again based on feedback, and also for new boards added since I started this PR.

deadprogram:
Anything else needed for this PR?

src/machine/machine_atmega.go
```diff
@@ -27,7 +27,7 @@ func (i2c I2C) Configure(config I2CConfig) {
 
 	// twi bit rate formula from atmega128 manual pg. 204:
 	// SCL Frequency = CPU Clock Frequency / (16 + (2 * TWBR))
-	// NOTE: TWBR should be 10 or higher for master mode.
+	// NOTE: TWBR should be 10 or higher for primary mode.
```
aykevl: You might consider calling this controller mode instead of primary mode for consistency.

src/machine/machine_atsamd21.go
```diff
@@ -1157,7 +1157,7 @@ func (spi SPI) Configure(config SPIConfig) error {
 		dataOrder = 1
 	}
 
-	// Set SPI master
+	// Set SPI mode to chip
```
aykevl: Again, I believe the new name is "controller".

src/machine/machine_atsamd51.go
```diff
@@ -1111,7 +1111,7 @@ func (i2c I2C) Configure(config I2CConfig) error {
 		i2c.Bus.SYNCBUSY.HasBits(sam.SERCOM_I2CM_SYNCBUSY_SWRST) {
 	}
 
-	// Set i2c master mode
+	// Set i2c chip mode
```
aykevl: And here

aykevl (approved):
A few nits for consistency, but really it looks good to me now. Feel free to fix this (or not) and merge, LGTM.

src/machine/machine_atsamd21.go
deadprogram: Changed.

src/machine/machine_atmega.go
deadprogram: Changed.

src/machine/machine_atsamd51.go
deadprogram: Done.

aykevl:
Looks good to me, so merging. There is a CI failure but that one already existed before this PR (and should be fixed separately).

PR #1217, opened 2020-07-07
openocd flash, and regex error matching

audstanley:
This fixes the CPUTAPID problem by attempting to use the default CPUTAPID first, and if openocd fails, tries again by setting the CPUTAPID to 0.

This also detects when the software reset has failed (which happens once in a while [ for me it's like after 50 flashes - so it's rare]), prompts the user that they need to hold the reset pin, it runs the flash command, and then runs it again for a successful flash.  This has worked perfectly every time so far with this fix.

This seems to me that this is the same problem (with the software reset) that Lady Ada was experiencing a long while ago:
https://www.youtube.com/watch?v=-EIaDxSIBYw&t=06m48s
...without the need to unplug, wait a minute, and plug back in.  So that's good.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"bytes"
 	"errors"
 	"flag"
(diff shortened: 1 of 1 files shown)


audstanley:
I accidentally left some comments at the very bottom of the changes that can be removed.

PR #1218, opened 2020-07-07
openOCD fix with CPUTAPID as well as fix for "Unable to reset target"

audstanley:
This should explain the problems that I have been having : 

[openocd fix](https://www.youtube.com/watch?v=4-srV5S6vTA)

### DIFF
--- a/main.go
+++ b/main.go
@@ -273,11 +273,9 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 			fmt.Println("---------------- TINYGO - OPENOCD WRAPPER ----------------")
 			attemptNum := 1
 			args, err := config.OpenOCDConfiguration()
-
 			if err != nil {
 				return err
 			}
-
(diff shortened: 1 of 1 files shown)


aykevl:
I do not think this PR is ready for review. It adds a lot of noise (while OpenOCD already produces a lot of noise) and I do not think reading from stdin is acceptable for a tool such as TinyGo (instead, it might just exit with an error indicating the board needs a reset).

Instead, I would suggest the following. Add a new field to the JSON configuration (see compileopts/target.go) similar to `openocd-transport` that indicates the CPUTAPID. If it is set, a new command can be inserted when constructing the OpenOCD command in `OpenOCDConfiguration` (`set CPUTAPID <id>`), again similar to `openocd-transport`.

audstanley:
Fair enough,  
thank you for the feedback, and I'll take a look at the code and the json config.  I'll close this pull request, and the other as well. 🙂

PR #1220, opened 2020-07-08
riscv: switch to tasks-based scheduler

yannishuber:
This pull request is based on the work of @aykevl in #963.

### DIFF
--- a/src/runtime/arch_tinygoriscv32.go
+++ b/src/runtime/arch_tinygoriscv32.go
@@ -2,7 +2,7 @@
 
 package runtime
 
-const GOARCH = "arm" // riscv pretends to be arm
+const GOARCH = "riscv"
 
 // The bitness of the CPU (e.g. 8, 32, 64).
 const TargetBits = 32
--- a/src/runtime/arch_tinygoriscv64.go
+++ b/src/runtime/arch_tinygoriscv64.go
@@ -2,7 +2,7 @@
 
 package runtime
 
-const GOARCH = "arm64" // riscv pretends to be arm
+const GOARCH = "riscv64"
 
 // The bitness of the CPU (e.g. 8, 32, 64).
 const TargetBits = 64
--- a/src/device/riscv/handleinterrupt.S
+++ b/src/device/riscv/handleinterrupt.S
@@ -1,7 +1,5 @@
 #ifdef __riscv_flen
 #define NREG 48
-#define LFREG flw
-#define SFREG fsw
 #else
 #define NREG 16
 #endif
@@ -10,10 +8,18 @@
 #define REGSIZE 8
 #define SREG sd
 #define LREG ld
+#ifdef __riscv_flen
+#define LFREG fld
+#define SFREG fsd
+#endif
 #else
 #define REGSIZE 4
 #define SREG sw
 #define LREG lw
+#ifdef __riscv_flen
+#define LFREG flw
+#define SFREG fsw
+#endif
 #endif
 
 .section .text.handleInterruptASM
--- a/src/internal/task/task_stack_tinygoriscv.go
+++ b/src/internal/task/task_stack_tinygoriscv.go
@@ -6,26 +6,6 @@ import "unsafe"
 
 const stackSize = 1024
 
-// calleeSavedRegs is the list of registers that must be saved and restored when
-// switching between tasks. Also see scheduler_riscv.S that relies on the
-// exact layout of this struct.
-type calleeSavedRegs struct {
-	s0  uintptr // x8 (fp)
-	s1  uintptr // x9
-	s2  uintptr // x18
-	s3  uintptr // x19
-	s4  uintptr // x20
-	s5  uintptr // x21
-	s6  uintptr // x22
-	s7  uintptr // x23
-	s8  uintptr // x24
-	s9  uintptr // x25
-	s10 uintptr // x26
-	s11 uintptr // x27
-
-	pc uintptr
-}
-
 // registers gets a pointer to the registers stored at the top of the stack.
 func (s *state) registers() *calleeSavedRegs {
 	return (*calleeSavedRegs)(unsafe.Pointer(s.sp))
@@ -49,7 +29,6 @@ func (s *state) archInit(stack []uintptr, fn uintptr, args unsafe.Pointer) {
 	// Store the initial sp for the startTask function (implemented in assembly).
 	s.sp = uintptr(unsafe.Pointer(&stack[uintptr(len(stack))-(unsafe.Sizeof(calleeSavedRegs{})/unsafe.Sizeof(uintptr(0)))]))
 
-	// Initialize the registers.
 	// These will be popped off of the stack on the first resume of the goroutine.
 	r := s.registers()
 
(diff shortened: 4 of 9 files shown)


yannishuber:
@aykevl So I found that the bug you mentioned in https://github.com/tinygo-org/tinygo/pull/963#issuecomment-653757977 was related to the stack size. With this patch, both `testdata/math.go` and `testdata/reflect.go` run without errors. Should the stack size be bigger as default ?

```diff
diff --git src/internal/task/task_stack_tinygoriscv.go src/internal/task/task_stack_tinygoriscv.go
index ae4a94a..079de43 100644
--- src/internal/task/task_stack_tinygoriscv.go
+++ src/internal/task/task_stack_tinygoriscv.go
@@ -4,7 +4,7 @@ package task
 
 import "unsafe"
 
-const stackSize = 1024
+const stackSize = 2048
 
 // calleeSavedRegs is the list of registers that must be saved and restored when
 // switching between tasks. Also see scheduler_riscv.S that relies on the
```

yannishuber:
Also I get the following LLVM error when building `testdata/math.go` for a RV64GC chip:
```
LLVM ERROR: Cannot select: 0x7f2cd0876170: f64 = fmaximum 0x7f2cd089bbd0, 0x7f2cd089bb00, math.go:173:22 @[ math.go:33:35 ]
  0x7f2cd089bbd0: f64,ch = CopyFromReg 0x7f2cd071a608, Register:f64 %3, bits.go:51:55 @[ asinh.go:48:22 @[ math.go:17:55 @[ math.go:9:37 ] ] ]
    0x7f2cd08cbce0: f64 = Register %3
  0x7f2cd089bb00: f64 = fadd 0x7f2cd089bbd0, 0x7f2cd08cc2f8, math.go:33:40
    0x7f2cd089bbd0: f64,ch = CopyFromReg 0x7f2cd071a608, Register:f64 %3, bits.go:51:55 @[ asinh.go:48:22 @[ math.go:17:55 @[ math.go:9:37 ] ] ]
      0x7f2cd08cbce0: f64 = Register %3
    0x7f2cd08cc2f8: f64,ch = load<(load 8 from constant-pool)> 0x7f2cd071a608, 0x7f2cd09dd898, undef:i64
      0x7f2cd09dd898: i64 = PseudoLLA TargetConstantPool:i64<double 1.000000e+00> 0
        0x7f2cd0875748: i64 = TargetConstantPool<double 1.000000e+00> 0
      0x7f2cd0876448: i64 = undef
In function: command-line-arguments.main
```
I think it comes from how LLVM handles the floating point registers but I am far from an expert. Any idea? When building the `blinky2` example, it works fine on RV32 and RV64.

EDIT: This was because of the unsupported LLVM intrinsics in `runtime/math.go`

yannishuber:
> So I found that the bug you mentioned in [#963 (comment)](https://github.com/tinygo-org/tinygo/pull/963#issuecomment-653757977) was related to the stack size.

Nevermind, there is another issue with `testdata/math.go` even on RV32 without any floating point extension.

yannishuber:
Without using the LLVM intrinsics in `runtime/math.go` (by using GOARCH=riscv64) the task scheduler seems to work on RV64GC even with the stack size of 1024. For RV32 without F extension there seems to be an issue with the soft float conversion because the values when running `testdata/math.go` are completely wrong (and stack size has to be bigger than 1024).

EDIT: with the `-opt=1` flag the `testdata/math.go` program produces the correct results on the hifive1b also with stack size of 1024. So it seems that one of the optimization passes causes the problem.

yannishuber:
Now this works on both the maixbit and the hifive1b. On the maixbit running `testdata/channel.go` traps on an misaligned load but the other test programs work. I think the misaligned load/store issue can be addressed in another PR. What do you think @aykevl ?

src/runtime/arch_tinygoriscv32.go
```diff
@@ -2,7 +2,7 @@
 
 package runtime
 
-const GOARCH = "arm" // riscv pretends to be arm
+const GOARCH = "riscv"
```
aykevl, line 5: Was this change intentional? I think the `GOARCH` value here should be kept in sync with the `GOARCH` in the target JSON file (targets/riscv.json) and with the build tag used there. Because those other two cannot be changed yet because of backwards compatibility, I think this should be kept at `"arm"`, even if that's incorrect. We can change that once we drop Go support before 1.14.

src/runtime/arch_tinygoriscv64.go
```diff
@@ -2,7 +2,7 @@
 
 package runtime
 
-const GOARCH = "arm64" // riscv pretends to be arm
+const GOARCH = "riscv64"
```
aykevl, line 5: Same here.

aykevl (commented):
> Now this works on both the maixbit and the hifive1b. On the maixbit running `testdata/channel.go` traps on an misaligned load but the other test programs work. I think the misaligned load/store issue can be addressed in another PR. What do you think @aykevl ?

So this PR would break maixbit support? That would be unfortunate.

I think this PR looks fine (except for the two things below). However, I would prefer if the misaligned load would be fixed first to not break anything with the this PR. Especially with the upcoming release.

src/runtime/arch_tinygoriscv32.go
yannishuber: Yes, for RISC-V architectures without floating-point extension (like the hifive1b) it is fine to use the arm `GOARCH`, however with the floating-point extension (the maixbit), LLVM produces errors (see https://github.com/tinygo-org/tinygo/pull/1220#issuecomment-655605906) when using the LLVM intrinsic functions in `runtime/math.go`. These functions are used only on targets with the arm or wasm `GOARCH` but do not seem supported on RISC-V with floating point extension. So changing the `GOARCH` value is only necessary to avoid using the LLVM intrinsics in `runtime/math.go`. We could check with another variable to avoid using those intrinsic functions but I thought it would be cleaner using the GOARCH and since the CI seemed to pass even on the older Go versions I thought it didn't impact backwards compatibility. How do you suggest to avoid using the LLVM intrinsics without changing the GOARCH?
https://github.com/tinygo-org/tinygo/blob/a9ba6ebad9ad85cacd9674a1af9229489a811f68/src/runtime/math.go#L58-L64

yannishuber:
> So this PR would break maixbit support? That would be unfortunate.

Yes you are right, I will come up with a support for misaligned load/stores to fix this, however it will probably not be ready for this release.

src/runtime/arch_tinygoriscv32.go
aykevl: Okay, that's a good reason to change the `GOARCH` value.

Ideally we should be checking which floating point intrinsics are supported on RISC-V so that they can be used directly. Using those intrinsics may be significantly faster than calling math functions (and reduces code size as well).

That said, I do believe this is a RISC-V backend bug. It should create calls to libc functions (such as `ceil`) instead of crashing.

aykevl:
@yannishuber are you still working on this? I would really like to see this land, but it shouldn't break anything (in this case, the maixbit).

aykevl:
I found that there were a few bugs in the scheduler that resulted in random errors such as "unaligned load". The only unaligned load there was, was as a result of memory corruption.

I have rebased my original PR and fixed this bug. It implements the tasks scheduler for RV32 only at the moment.

aykevl:
Support for this was merged in https://github.com/tinygo-org/tinygo/pull/2145.

PR #1221, opened 2020-07-08
targets/maixbit: cleanup output from kflash command

deadprogram:
This PR cleans up the output from the kflash command by removing ansi colors that were cluttering the markdown use by TinyHCI.

I also turned on the "verbose" output, which is still rather compact.


deadprogram:
What do you think @yannishuber seems ok?

yannishuber:
@deadprogram Looks good to me

deadprogram:
Thanks @yannishuber

PR #1222, opened 2020-07-08
Fix portL mappings for atmega2560

brandentimm:
Due to a typo in the machine package, port L pins were not mapped correctly on atmega2560.  This fixes the package so they are correctly mapped, tested working on my atmega2560.


deadprogram:
Thanks for the fix @brandentimm now squashing/merging.

PR #1223, opened 2020-07-09
Teensy 3.6 USB CDC

firelizzard18:
USB CDC support for Teensy 3.6 (NXP MK66F18)

### DIFF
--- a/src/machine/board_teensy36.go
+++ b/src/machine/board_teensy36.go
@@ -8,6 +8,24 @@ func CPUFrequency() uint32 { return 180000000 }
 // ClockFrequency returns the frequency of the external oscillator (16MHz)
 func ClockFrequency() uint32 { return 16000000 }
 
+const (
+	usb_STRING_MANUFACTURER = "Teensy"
+	usb_STRING_PRODUCT      = "Teensy 3.6 USB Serial"
+
+	usb_DEVICE_CLASS    = 0x02
+	usb_DEVICE_SUBCLASS = 0
+	usb_DEVICE_PROTOCOL = 0
+
+	usb_EP0_SIZE    = 64
+	usb_BUFFER_SIZE = usb_EP0_SIZE
+
+	usb_VID        = 0x16C0
+	usb_PID        = 0x0483 // teensy usb serial
+	usb_BCD_DEVICE = 0x0277 // teensy 3.6
+)
+
+var usb_STRING_SERIAL string
+
 // LED on the Teensy
 const LED = PC05
 
@@ -99,3 +117,17 @@ const (
 	defaultUART4RX = D34
 	defaultUART4TX = D33
 )
+
+//go:linkname sleepTicks runtime.sleepTicks
+func sleepTicks(int64)
+
+func InitPlatform() {
+	// for background about this startup delay, please see these conversations
+	// https://forum.pjrc.com/threads/36606-startup-time-(400ms)?p=113980&viewfull=1#post113980
+	// https://forum.pjrc.com/threads/31290-Teensey-3-2-Teensey-Loader-1-24-Issues?p=87273&viewfull=1#post87273
+
+	sleepTicks(25 * 1000) // 50 for TeensyDuino < 142
+	// usb_STRING_SERIAL = readSerialNumber()
+	USB0.Configure()
+	sleepTicks(275 * 1000) // 350 for TeensyDuino < 142
+}
--- a/src/machine/usb_nxpmk66f18.go
+++ b/src/machine/usb_nxpmk66f18.go
@@ -1,0 +1,626 @@
+// +build nxp,mk66f18
+
+package machine
+
+import (
+	"device/nxp"
+	"fmt"
+	"reflect"
+	"runtime/interrupt"
+	"runtime/volatile"
+	"unsafe"
+)
+
+const debugUSB = true
+
+var USB0 = USB{USB0_Type: nxp.USB0, SCGC: &nxp.SIM.SCGC4, SCGCMask: nxp.SIM_SCGC4_USBOTG, ENDPT: (*[16]register8Aligned32)(unsafe.Pointer(&nxp.USB0.ENDPT0))}
+
+//go:align 512
+var usbBufferDescriptorRecords [usb_MaxBDTSize]usbBufferDescriptor
+var usbBufferDescriptorTable = usbBufferDescriptorTable_t{records: &usbBufferDescriptorRecords}
+
+func init() {
+	USB0.Interrupt = interrupt.New(nxp.IRQ_USB0, USB0.serviceInterrupt)
+}
+
+//go:linkname millisSinceBoot runtime.millisSinceBoot
+func millisSinceBoot() uint64
+
+type USB struct {
+	*nxp.USB0_Type
+	SCGC      *volatile.Register32
+	SCGCMask  uint32
+	Interrupt interrupt.Interrupt
+	ENDPT     *[16]register8Aligned32
+
+	CDCRXBuffer RingBuffer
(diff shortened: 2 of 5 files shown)


firelizzard18:
At one point, it looked like this version was working. Or at least I was seeing far more log messages from the USB ISR. But now it doesn't get past the device descriptor:

```
usb istat 0x01
usb istat 0x08
usb stat 0x00
usb bd tok 0xD
usb setup raw 8006000100004000
usb setup request=6 type=80 value=100 index=0
usb setup get descriptor
usb ep0 transmit 18 byte(s): 12010002ef020140c0168304000101020301
usb istat 0x08
usb stat 0x08
usb bd tok 0x9
usb istat 0x01
```

Except now, I'm not even getting that... All I get is "usb done" (indicating that `machine.USB0.Configure()` completed).

@deadprogram @aykevl Any ideas? As far as I can tell, things are failing early enough that Wireshark doesn't see anything.

src/machine/board_teensy36.go
```diff
@@ -99,3 +117,17 @@ const (
...
+
+func InitPlatform() {
+	// for background about this startup delay, please see these conversations
+	// https://forum.pjrc.com/threads/36606-startup-time-(400ms)?p=113980&viewfull=1#post113980
+	// https://forum.pjrc.com/threads/31290-Teensey-3-2-Teensey-Loader-1-24-Issues?p=87273&viewfull=1#post87273
+
+	sleepTicks(25 * 1000) // 50 for TeensyDuino < 142
+	// usb_STRING_SERIAL = readSerialNumber()
+	USB0.Configure()
+	sleepTicks(275 * 1000) // 350 for TeensyDuino < 142
```
aykevl, line 132: Is there a reason not to include this directly in `postinit`? It's not like you're supposed to call `InitPlatform` (except for the runtime).

aykevl (commented):
I never worked directly with USB-CDC so I don't think I can help you much debugging it, unfortunately.

(one quick comment below, I haven't really looked through the code)

src/machine/board_teensy36.go
firelizzard18: I thought I would leave room for non-Teensy boards that are based on the NXP MK66F18. I believe these delays are more related to how the Teensy programmer detects boards than to the chip itself.

aykevl: Ah, I see. So the function is only intended to be called from the runtime but put in the machine package because it is board specific? In that case, it would be useful to add a comment explaining that, so that it's clear this function isn't supposed to be called elsewhere.

src/machine/usb_nxpmk66f18.go
```diff
@@ -0,0 +1,626 @@
...
+import (
+	"device/nxp"
+	"fmt"
+	"reflect"
+	"runtime/interrupt"
+	"runtime/volatile"
+	"unsafe"
+)
+
+const debugUSB = true
```
aykevl, line 14: Is this intentional? It looks like it should be `false`.

src/machine/usb_nxpmk66f18.go
```diff
@@ -0,0 +1,626 @@
+// +build nxp,mk66f18
+
+package machine
+
+import (
+	"device/nxp"
+	"fmt"
```
aykevl, line 7: Don't import fmt in the machine package. This is a _huge_ dependency with a large dependency graph and importing it in the machine package may lead to circular dependencies. It also bloats the binary and increases compile time.

Instead, I would suggest to use `println` for debugging (and disabled by default).

src/machine/usb_nxpmk66f18.go
```diff
@@ -0,0 +1,626 @@
...
+
+	if len(data) >= 1024 {
+		panic("USB Buffer Descriptor overflow")
+	}
+
+	slice := (*reflect.SliceHeader)(unsafe.Pointer(&data))
+
+	*bde.retain = data
+	bde.record.address.Set(uint32(slice.Data))
+	bde.Describe(uint32(slice.Len), data1)
```
aykevl, line 603: The reflect package is a relatively large dependency and shouldn't be used in the machine package. It is also possible to do without in this case:

```suggestion
	*bde.retain = data
	bde.record.address.Set(uint32(uintptr(unsafe.Pointer(&data[0]))))
	bde.Describe(len(data), data1)
```

src/machine/usb_nxpmk66f18.go
```diff
@@ -0,0 +1,626 @@
...
+}
+
+func (bde usbBufferDescriptorEntry) Data() []byte {
+	length := bde.record.descriptor.Get() >> 16
+
+	var data []byte
+	slice := (*reflect.SliceHeader)(unsafe.Pointer(&data))
+	slice.Len = uintptr(length)
+	slice.Cap = uintptr(length)
+	slice.Data = uintptr(bde.record.address.Get())
```
aykevl, line 613: Where does `bde.record.address` come from? I'm guessing it is some sort of DMA buffer, did you put it in a global somewhere (for example) to make sure it is not garbage collected?
I think it would be much better to somehow keep the original slice around and use that, to avoid some unsafe code.

src/machine/usb_nxpmk66f18.go
```diff
@@ -0,0 +1,626 @@
...
+
+	state := interrupt.Disable()
+	log, u.Log = u.Log, nil
+	interrupt.Restore(state)
+
+	if log == nil {
+		return []string{}
+	}
+	return log
+}
```
aykevl, line 63: This looks like debug code. While functions that aren't called won't take up extra space (they get optimized away), the `Log` buffer cannot be optimized away. Maybe you can make it a separate global (separate from the `USB` type) if you want to keep it?

src/machine/usb_nxpmk66f18.go
```diff
@@ -0,0 +1,626 @@
...
+		usbBufferDescriptorTable.Get(0, true, true).Reset()                      // EP0 TX odd
+
+		// activate endpoint 0
+		u.configureEndpoint(0, USBEndpointConfiguration{
+			Transmit:  true,
+			Receive:   true,
+			Handshake: true,
+		})
+
+		// clear all ending interrupts
```
aykevl, line 171: typo?
```suggestion
		// clear all pending interrupts
```

src/machine/usb_nxpmk66f18.go
```diff
@@ -0,0 +1,626 @@
...
+func (u *USB) CDC() USBCDC {
+	return USBCDC{&u.CDCRXBuffer}
+}
+
+type USBCDC struct {
+	Buffer *RingBuffer
+}
+
+func (u USBCDC) WriteByte(byte) {
+	panic("not implemented")
```
aykevl, line 493: This seems strange, you implement a `USBCDC` type that doesn't actually work?

src/machine/usb_nxpmk66f18.go
```diff
@@ -0,0 +1,626 @@
...
+
+	usb_BDT_TOK_PID_Mask = 0b00111100
+	usb_BDT_TOK_PID_Pos  = 2
+
+	usb_TOK_PID_OUT   = 0x1
+	usb_TOK_PID_IN    = 0x9
+	usb_TOK_PID_SETUP = 0xD
+)
+
+type register8Aligned32 struct {
```
aykevl, line 515: Just to be clear, this struct is not aligned on a 32-bit (4 byte) boundary. It is still 8 bit aligned but simply takes up four bytes. Was this intentional?

aykevl (commented):
I'm not sure I understand the purpose of this PR. It doesn't seem to implement USB-CDC support, so what does it add in functionality?

deadprogram:

PR #1225, opened 2020-07-09
compileopts: automatically add -g flag when including debug symbols

aykevl:
Debug information is often useful and there is no reason to include it
for Go code but not for C code. Also, disabling debug information should
disable it entirely, not just for Go code.

---

This will also be useful for automatically determining stack sizes, as such support (in my attempts) relies on DWARF stack frame information.


deadprogram:
Makes sense to me, now merging.

PR #1226, opened 2020-07-10
Nintendo Switch Minimal Support

racerxdl:
That pullrequest is a minimum support for nintendo switch (can just print to emulator debug output) which doesn't require any external library.

That does not invalidate #1108 but will be the base for changes in #1108.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -342,6 +342,8 @@ endif
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=pca10040 -opt=1     examples/blinky1
 	@$(MD5SUM) test.hex
+	$(TINYGO) build             -o test.elf -target=nintendoswitch      examples/serial
+	@$(MD5SUM) test.elf
 
 wasmtest:
 	$(GO) test ./tests/wasm
--- a/src/runtime/runtime_nintendoswitch.go
+++ b/src/runtime/runtime_nintendoswitch.go
@@ -1,0 +1,73 @@
+// +build nintendoswitch
+
+package runtime
+
+type timeUnit int64
+
+const asyncScheduler = false
+
+func postinit() {}
+
+// Entry point for Go. Initialize all packages and call main.main().
+//export main
+func main() int {
+	preinit()
+	run()
+
+	// Call exit to correctly finish the program
+	// Without this, the application crashes at start, not sure why
+	return exit(0)
+}
+
+// sleepTicks sleeps for the specified system ticks
+func sleepTicks(d timeUnit) {
+	sleepThread(uint64(ticksToNanoseconds(d)))
+}
+
+// armTicksToNs converts cpu ticks to nanoseconds
+// Nintendo Switch CPU ticks has a fixed rate at 19200000
+// It is basically 52 ns per tick
+// The formula 625 / 12 is equivalent to 1e9 / 19200000
+func ticksToNanoseconds(tick timeUnit) int64 {
+	return int64(tick * 625 / 12)
+}
+
+func nanosecondsToTicks(ns int64) timeUnit {
+	return timeUnit(12 * ns / 625)
+}
+
+func ticks() timeUnit {
+	return timeUnit(ticksToNanoseconds(timeUnit(getArmSystemTick())))
+}
+
+var stdoutBuffer = make([]byte, 0, 120)
+
+func putchar(c byte) {
+	if c == '\n' || len(stdoutBuffer)+1 >= 120 {
+		NxOutputString(string(stdoutBuffer))
+		stdoutBuffer = stdoutBuffer[:0]
+		return
+	}
+
+	stdoutBuffer = append(stdoutBuffer, c)
+}
+
+func usleep(usec uint) int {
+	sleepThread(uint64(usec) * 1000)
+	return 0
+}
+
+func abort() {
+	for {
+		exit(1)
+	}
+}
+
+//export sleepThread
+func sleepThread(nanos uint64)
+
+//export exit
+func exit(code int) int
+
+//export armGetSystemTick
+func getArmSystemTick() int64
--- a/src/runtime/runtime_nintendoswitch.s
+++ b/src/runtime/runtime_nintendoswitch.s
@@ -1,0 +1,45 @@
+.section .text.armGetSystemTick, "ax", %progbits
+.global  armGetSystemTick
+.type    armGetSystemTick, %function
+.align 2
+armGetSystemTick:
+    mrs x0, cntpct_el0
+    ret
+
+.section .text.nxOutputString, "ax", %progbits
+.global nxOutputString
+.type nxOutputString, %function
+.align 2
+.cfi_startproc
+nxOutputString:
+	svc 0x27
+	ret
+.cfi_endproc
+
+.section .text.exit, "ax", %progbits
+.global exit
+.type exit, %function
+.align 2
+exit:
+	svc 0x7
+	ret
+
+.section .text.setHeapSize, "ax", %progbits
+.global setHeapSize
+.type setHeapSize, %function
+.align 2
+setHeapSize:
+	str x0, [sp, #-16]!
+	svc 0x1
+	ldr x2, [sp], #16
+	str x1, [x2]
+	ret
+
+
+.section .text.sleepThread, "ax", %progbits
+.global sleepThread
+.type sleepThread, %function
+.align 2
+sleepThread:
+	svc 0xB
+	ret
--- a/src/runtime/runtime_nintendoswitch_heap.go
+++ b/src/runtime/runtime_nintendoswitch_heap.go
@@ -1,0 +1,31 @@
+// +build nintendoswitch
+
+// +build gc.conservative gc.leaking
+
+package runtime
+
+import "unsafe"
+
+const heapSize = 0x2000000 * 16 // Default by libnx
+
+//go:extern _stack_top
+var stackTopSymbol [0]byte
+
+var (
+	heapStart = uintptr(0)
+	heapEnd   = uintptr(0)
+	stackTop  = uintptr(unsafe.Pointer(&stackTopSymbol))
+)
+
+//export setHeapSize
+func setHeapSize(addr *uintptr, length uint64) uint64
+
+func preinit() {
+	setHeapSize(&heapStart, heapSize)
+
+	if heapStart == 0 {
+		panic("failed to allocate heap")
+	}
+
+	heapEnd = heapStart + heapSize
+}
--- a/src/runtime/runtime_nintendoswitch_svc.go
+++ b/src/runtime/runtime_nintendoswitch_svc.go
@@ -1,0 +1,22 @@
+// +build nintendoswitch
+
+package runtime
+
+import (
+	"unsafe"
+)
+
+// Result nxOutputString(const char *str, u64 size)
+//export nxOutputString
+func nxOutputString(str *uint8, size uint64) uint64
+
+func NxOutputString(str string) uint64 {
+	strData := (*_string)(unsafe.Pointer(&str))
+	return nxOutputString((*uint8)(unsafe.Pointer(strData.ptr)), uint64(strData.length))
+}
+
+//export malloc
+func extalloc(size uintptr) unsafe.Pointer
+
+//export free
+func extfree(ptr unsafe.Pointer)
--- a/src/syscall/syscall_libc.go
+++ b/src/syscall/syscall_libc.go
@@ -1,4 +1,4 @@
-// +build darwin
+// +build darwin nintendoswitch
 
 package syscall
 
--- a/src/syscall/syscall_nintendoswitch.go
+++ b/src/syscall/syscall_nintendoswitch.go
@@ -1,0 +1,48 @@
+// +build nintendoswitch
+
+package syscall
+
+import "errors"
+
+// A Signal is a number describing a process signal.
+// It implements the os.Signal interface.
+type Signal int
+
+const (
+	_ Signal = iota
+	SIGCHLD
+	SIGINT
+	SIGKILL
+	SIGTRAP
+	SIGQUIT
+	SIGTERM
+)
+
+// File system
+
+const (
+	Stdin  = 0
+	Stdout = 1
+	Stderr = 2
+)
+
+const (
+	O_RDONLY = 0
+	O_WRONLY = 1
+	O_RDWR   = 2
+
+	O_CREAT  = 0100
+	O_CREATE = O_CREAT
+	O_TRUNC  = 01000
+	O_APPEND = 02000
+	O_EXCL   = 0200
+	O_SYNC   = 010000
+
+	O_CLOEXEC = 0
+)
+
+var dummyError = errors.New("unknown syscall error")
+
+func getErrno() error {
+	return dummyError
+}
--- a/targets/nintendoswitch.ld
+++ b/targets/nintendoswitch.ld
@@ -1,0 +1,61 @@
+OUTPUT_FORMAT(elf64-littleaarch64)
+OUTPUT_ARCH(aarch64)
+ENTRY(_start)
+
+PHDRS
+{
+  text    PT_LOAD     FLAGS(5);
+  rodata  PT_LOAD     FLAGS(4);
+  data    PT_LOAD     FLAGS(6);
+  bss     PT_LOAD     FLAGS(6);
+  dynamic PT_DYNAMIC;
+}
+
+SECTIONS
+{
+  . = 0;
+
+  .text : ALIGN(0x1000) {
+    HIDDEN(__text_start = .);
+    KEEP(*(.text.jmp))
+
+    . = 0x80;
+
+    *(.text .text.*)
+  }
+
+  /* Read-only sections */
+
+  . = ALIGN(0x1000);
+
+  .rodata : { *(.rodata .rodata.*) } :rodata
+  .mod0 : {
+    KEEP(crt0.nso.o(.data.mod0))
+    KEEP(crt0.nro.o(.data.mod0))
+    KEEP(crt0.lib.nro.o(.data.mod0))
+  }
+
+  /* Read-write sections */
+   . = ALIGN(0x1000);
+
+  .data : {
+    *(.data .data.*)
+  } :data
+
+  .dynamic : {
+    HIDDEN(__dynamic_start = .);
+    *(.dynamic)
+  }
+
+  /* BSS section */
+
+  . = ALIGN(0x1000);
+
+  .bss : {
+    HIDDEN(__bss_start = .);
+    *(.bss .bss.*)
+    *(COMMON)
+    . = ALIGN(8);
+    HIDDEN(__bss_end = .);
+  } :bss
+}
--- a/targets/nintendoswitch.s
+++ b/targets/nintendoswitch.s
@@ -1,0 +1,51 @@
+.section .text.jmp, "x"
+.global _start
+_start:
+    b start
+    .word _mod_header - _start
+
+.section .data.mod0
+    .word 0, 8
+
+.global _mod_header
+_mod_header:
+    .ascii "MOD0"
+    .word __dynamic_start - _mod_header
+    .word __bss_start - _mod_header
+    .word __bss_end - _mod_header
+    .word 0, 0 // eh_frame_hdr start/end
+    .word 0 // runtime-generated module object offset
+
+.section .text, "x"
+.global start
+start:
+
+    // save lr
+    mov  x7, x30
+
+    // get aslr base
+    bl   +4
+    sub  x6, x30, #0x88
+
+    // context ptr and main thread handle
+    mov  x5, x0
+    mov  x4, x1
+
+    // clear .bss
+    adrp x5, __bss_start
+    add x5, x5, #:lo12:__bss_start
+    adrp x6, __bss_end
+    add x6, x6, #:lo12:__bss_end
+
+bssloop:
+    cmp x5, x6
+    b.eq run
+    str xzr, [x5]
+    add x5, x5, 8
+    b bssloop
+
+run:
+    // call entrypoint
+    adrp x30, exit
+    add  x30, x30, #:lo12:exit
+    b    main
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -284,6 +284,16 @@ func (c *Config) CodeModel() string {
 	return "default"
 }
 
+// RelocationModel returns the relocation model in use on this platform. Valid
+// values are "static", "pic", "dynamicnopic".
+func (c *Config) RelocationModel() string {
+	if c.Target.RelocationModel != "" {
+		return c.Target.RelocationModel
+	}
+
+	return "static"
+}
+
 type TestConfig struct {
 	CompileTestBinary bool
 	// TODO: Filter the test functions to run, include verbose flag, etc
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -50,6 +50,7 @@ type TargetSpec struct {
 	OpenOCDTransport string   `json:"openocd-transport"`
 	JLinkDevice      string   `json:"jlink-device"`
 	CodeModel        string   `json:"code-model"`
+	RelocationModel  string   `json:"relocation-model"`
 }
 
 // copyProperties copies all properties that are set in spec2 into itself.
@@ -133,6 +134,10 @@ func (spec *TargetSpec) copyProperties(spec2 *TargetSpec) {
 	}
 	if spec2.CodeModel != "" {
 		spec.CodeModel = spec2.CodeModel
+	}
+
+	if spec2.RelocationModel != "" {
+		spec.RelocationModel = spec2.RelocationModel
 	}
 }
 
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -93,6 +93,7 @@ func NewTargetMachine(config *compileopts.Config) (llvm.TargetMachine, error) {
 	features := strings.Join(config.Features(), ",")
 
 	var codeModel llvm.CodeModel
+	var relocationModel llvm.RelocMode
 
 	switch config.CodeModel() {
 	case "default":
@@ -109,7 +110,16 @@ func NewTargetMachine(config *compileopts.Config) (llvm.TargetMachine, error) {
 		codeModel = llvm.CodeModelLarge
 	}
 
-	machine := target.CreateTargetMachine(config.Triple(), config.CPU(), features, llvm.CodeGenLevelDefault, llvm.RelocStatic, codeModel)
+	switch config.RelocationModel() {
+	case "static":
+		relocationModel = llvm.RelocStatic
+	case "pic":
+		relocationModel = llvm.RelocPIC
+	case "dynamicnopic":
+		relocationModel = llvm.RelocDynamicNoPic
+	}
+
+	machine := target.CreateTargetMachine(config.Triple(), config.CPU(), features, llvm.CodeGenLevelDefault, relocationModel, codeModel)
 	return machine, nil
 }
 
--- a/loader/goroot.go
+++ b/loader/goroot.go
@@ -165,7 +165,7 @@ func mergeDirectory(goroot, tinygoroot, tmpgoroot, importPath string, overrides 
 // with the TinyGo version. This is the case on some targets.
 func needsSyscallPackage(buildTags []string) bool {
 	for _, tag := range buildTags {
-		if tag == "baremetal" || tag == "darwin" {
+		if tag == "baremetal" || tag == "darwin" || tag == "nintendoswitch" {
 			return true
 		}
 	}
--- a/src/runtime/runtime_nintendoswitch_noheap.go
+++ b/src/runtime/runtime_nintendoswitch_noheap.go
@@ -1,0 +1,7 @@
+// +build nintendoswitch
+
+// +build gc.none gc.extalloc
+
+package runtime
+
+func preinit() {}
--- a/src/runtime/runtime_unix.go
+++ b/src/runtime/runtime_unix.go
@@ -1,4 +1,5 @@
 // +build darwin linux,!baremetal freebsd,!baremetal
+// +build !nintendoswitch
 
 package runtime
 
--- a/src/runtime/runtime_unix_heap.go
+++ b/src/runtime/runtime_unix_heap.go
@@ -1,4 +1,5 @@
 // +build darwin linux,!baremetal freebsd,!baremetal
+// +build !nintendoswitch
 
 // +build gc.conservative gc.leaking
 
--- a/src/runtime/runtime_unix_noheap.go
+++ b/src/runtime/runtime_unix_noheap.go
@@ -1,5 +1,7 @@
 // +build darwin linux,!baremetal freebsd,!baremetal
 
+// +build !nintendoswitch
+
 // +build gc.none gc.extalloc
 
 package runtime
--- a/src/syscall/tables_baremetal.go
+++ b/src/syscall/tables_baremetal.go
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// +build baremetal
+// +build baremetal nintendoswitch
 
 package syscall
 
--- a/targets/nintendoswitch.json
+++ b/targets/nintendoswitch.json
@@ -1,0 +1,28 @@
+{
+  "llvm-target": "aarch64",
+  "build-tags": ["nintendoswitch", "arm64"],
+  "goos": "linux",
+  "goarch": "arm64",
+  "compiler": "clang",
+  "linker": "ld.lld",
+  "rtlib": "compiler-rt",
+  "libc": "picolibc",
+  "gc": "conservative",
+  "relocation-model": "pic",
+  "cflags": [
+    "-target", "aarch64-none-linux-gnu",
+    "-mtune=cortex-a57",
+    "-fPIE",
+    "-Werror",
+    "-Qunused-arguments",
+    "-fshort-enums",
+    "-fomit-frame-pointer",
+    "-fno-exceptions", "-fno-unwind-tables",
+    "-ffunction-sections", "-fdata-sections"
+  ],
+  "linkerscript": "targets/nintendoswitch.ld",
+  "extra-files": [
+    "targets/nintendoswitch.s",
+    "src/runtime/runtime_nintendoswitch.s"
+  ]
+}


racerxdl:
@aykevl / @deadprogram I can't find out why the test with go1.12 is failing but others are not. Still that does seen to be fine.

deadprogram:
The CI fail is something to do with the TinyGo cache cc @jaddr2line 

I pushed the branch myself and it built correctly, so the code in this branch should also be considered as passing CI.

targets/nintendo-switch.json
```diff
@@ -0,0 +1,31 @@
...
+  "compiler": "clang",
+  "linker": "ld.lld",
+  "rtlib": "compiler-rt",
+  "libc": "picolibc",
+  "gc": "conservative",
+  "relocation-model": "pic",
+  "cflags": [
+    "-target", "aarch64-none-linux-gnu",
+    "-mtune=cortex-a53",
+    "-g",
```
deadprogram: @racerxdl see https://github.com/tinygo-org/tinygo/pull/1225 do you need to remove the `-g` flag from here?

deadprogram:
> That pullrequest is a minimum support for nintendo switch (can just print to emulator debug output) which doesn't require any external library.

I agree this is a much better first step at Nintendo Switch support.

targets/nintendo-switch.json
racerxdl: Oh, I think it was a copy-paste issue. Fixing it.

racerxdl:
omg, now is failing in more tests, but all related to `.cache` stuff.

deadprogram:
You did rebase against the latest `dev` right?

racerxdl:
Ok I did forgot that @deadprogram. I think it was that!

danielhe4rt (approved):

src/examples/nintendoswitch/debug-hello.go
```diff
@@ -0,0 +1,23 @@
+// This program is a minimal Nintendo Switch program that just outputs messages to
+// emulator console
+package main
+
+import (
+	"runtime"
+	"time"
+)
+
+func print(data string) {
```
niaow: Try renaming this to avoid overloading the language builtin of the same name. Maybe `debugPrint`?

src/runtime/runtime_nintendoswitch.s
```diff
@@ -0,0 +1,54 @@
...
+.cfi_endproc
+
+
+.section .text.exit, "ax", %progbits
+.global exit
+.type exit, %function
+.align 2
+exit:
+	svc 0x7
+	ret
```
jaddr2line, line 34: Is this `ret` instruction actually reachable?

src/runtime/runtime_nintendoswitch_svc.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"unsafe"
+)
+
+// Result nxOutputString(const char *str, u64 size)
+//go:export nxOutputString
+func NxOutputString(str *byte, size uint64) uint64
+
+//go:export malloc
+func extalloc(size uintptr) unsafe.Pointer
```
niaow: Change to `//export` for consistency

targets/nintendo-switch.json
```diff
@@ -0,0 +1,28 @@
...
+  "goarch": "arm64",
+  "compiler": "clang",
+  "linker": "ld.lld",
+  "rtlib": "compiler-rt",
+  "libc": "picolibc",
+  "gc": "conservative",
+  "relocation-model": "pic",
+  "cflags": [
+    "-target", "aarch64-none-linux-gnu",
+    "-mtune=cortex-a53",
```
niaow: Is there a reason we are tuning for the LITTLE core instead of the big core? This isn't necessarily a problem, it just is different from the other switch software builds I have looked at.

src/runtime/runtime_nintendoswitch_heap.go
```diff
@@ -0,0 +1,34 @@
...
+	heapStart = uintptr(0)
+	heapEnd   = uintptr(0)
+	stackTop  = uintptr(unsafe.Pointer(&stackTopSymbol))
+)
+
+//export setHeapSize
+func setHeapSize(addr *uintptr, length uint64) uint64
+
+//export sleepThread
+func sleepThread(nanos uint64)
```
niaow: Shouldn't this be in `src/runtime/runtime_nintendoswitch.go`? This does not seem to be associated with the heap.

src/runtime/runtime_nintendoswitch.go
```diff
@@ -0,0 +1,72 @@
...
+func main() int {
+	preinit()
+	run()
+
+	return exit(0) // Call libc_exit to cleanup libnx
+}
+
+// sleepTicks
+func sleepTicks(d timeUnit) {
+	usleep(uint(ticksToNanoseconds(d) / 1000))
```
niaow: We seem to be dividing by 1000 and then re-multiplying, effectively rounding this down. Is this intentional? Can we flatten this instead to a call to `sleepThread`?

src/runtime/runtime_nintendoswitch.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func usleep(usec uint) int {
+	sleepThread(uint64(usec) * 1000)
+	return 0
+}
+
+//export abort
+func abort() {
+	exit(1)
```
niaow: I would suggest adding a `for{}` around this exit. This will tell the compiler that `abort` never returns, allowing it to omit generating machine code following a panic. This should cause a non-negligible reduction in code size and improve performance.

src/runtime/runtime_nintendoswitch_heap.go
```diff
@@ -0,0 +1,34 @@
...
+
+func preinit() {
+	setHeapSize(&heapStart, heapSize)
+
+	if heapStart == 0 {
+		panic("Cannot allocate heap")
```
niaow:
```suggestion
		runtimePanic("failed to allocate heap")
```

src/runtime/runtime_nintendoswitch.go
```diff
@@ -0,0 +1,72 @@
...
+
+	stdoutBuffer = append(stdoutBuffer, c)
+}
+
+func usleep(usec uint) int {
+	sleepThread(uint64(usec) * 1000)
+	return 0
+}
+
+//export abort
```
niaow: Why is this exported?

src/examples/nintendoswitch/debug-hello.go
```diff
@@ -0,0 +1,23 @@
+// This program is a minimal Nintendo Switch program that just outputs messages to
+// emulator console
+package main
```
niaow: I think you may want to revise the path. Right now it builds as `tinygo build -target nintendo-switch examples/nintendoswitch`. You may want to change this to something like `examples/nintendoswitch/debug-hello`.

src/runtime/runtime_nintendoswitch.go
```diff
@@ -0,0 +1,72 @@
...
+// Entry point for Go. Initialize all packages and call main.main().
+//export main
+func main() int {
+	preinit()
+	run()
+
+	return exit(0) // Call libc_exit to cleanup libnx
+}
+
+// sleepTicks
```
niaow: ?

niaow (dismissed):
A few minor issues and then I think we should be about ready to go.

targets/nintendo-switch.json
racerxdl: Oh that's a mistake. Nintendo Switch doesn't uses the A53. It think I misplaced this. let me fix it.

niaow: Well, it has 4 A53s and 4 A57s, but I would assume that we would want to tune for the A57s since tuning will only make a substantial difference for math-intensive workloads.

racerxdl: The Horizon OS from nintendo switch keeps the A53 cores disabled all the time. There is no scheduling to it, at least anyone never has been able to reach them.

niaow: Ah. Yeah it seems that Nintendo left them always off. . .

src/examples/nintendoswitch/debug-hello.go
racerxdl: That's true. I normalized everything to `nintendoswitch` now.

src/runtime/runtime_nintendoswitch.go
racerxdl: I think I copied from somewhere. Fixed!

src/runtime/runtime_nintendoswitch_heap.go
racerxdl: That's true. My bad.

src/runtime/runtime_nintendoswitch.go
racerxdl: Hmm, true. Let me do it.

src/runtime/runtime_nintendoswitch.s
```diff
@@ -0,0 +1,54 @@
...
+    mrs x0, cntpct_el0
+    ret
+
+.section .text.armGetSystemTickFreq, "ax", %progbits
+.global  armGetSystemTickFreq
+.type    armGetSystemTickFreq, %function
+.align 2
+armGetSystemTickFreq:
+    mrs x0, cntfrq_el0
+    ret
```
aykevl: This function does not seem to be called anywhere. Can it be removed?
Same goes for the `armGetSystemTickFreq` declaration above.

src/runtime/runtime_nintendoswitch.s
```diff
@@ -0,0 +1,54 @@
...
+
+.section .text.nxOutputString, "ax", %progbits
+.global nxOutputString
+.type nxOutputString, %function
+.align 2
+.cfi_startproc
+nxOutputString:
+	svc 0x27
```
aykevl, line 15: You could look at `arm.SVCall2` for these `svc` wrappers. It has the benefit of avoiding assembly.

src/runtime/runtime_nintendoswitch.go
racerxdl: Just explained better :D

src/runtime/runtime_nintendoswitch_svc.go
```diff
@@ -0,0 +1,17 @@
...
+
+package runtime
+
+import (
+	"unsafe"
+)
+
+// Result nxOutputString(const char *str, u64 size)
+//export nxOutputString
+func NxOutputString(str *byte, size uint64) uint64
```
aykevl: Strings in TinyGo are represented as exactly this, so it might work to simply declare this function as ` func NxOutputString(s string) uint64`. While it is an implementation detail, it is unlikely to change.

src/syscall/syscall_libc.go
```diff
@@ -1,46 +1,66 @@
...
 	"unsafe"
 )
 
+var (
+	notImplemented = errors.New("syscall: not implemented")
+)
+
 func Close(fd int) (err error) {
-	return ENOSYS // TODO
+	errN := libc_close(int32(fd))
```
aykevl: I still wonder why you need this? This PR does not include any other mention of the `Close` function. And while this is a useful addition, it may be better done separately. For example: it should ideally return a correct error.

The same goes for all other libc changes below, I don't see how they are necessary for minimal Nintendo Switch support?

src/examples/nintendoswitch/debug-hello.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func debugPrint(data string) {
+	t := []byte(data)
+	l := uint64(len(t))
+	runtime.NxOutputString(&t[0], l)
+}
+
+func main() {
+	print("Hello world!")
```
niaow: You missed one.

targets/nintendoswitch.ld
```diff
@@ -0,0 +1,81 @@
...
+OUTPUT_ARCH(aarch64)
+ENTRY(_start)
+
+PHDRS
+{
+  text PT_LOAD FLAGS(5);
+  rodata PT_LOAD FLAGS(4);
+  data PT_LOAD FLAGS(6);
+  bss PT_LOAD FLAGS(6);
+  dynamic PT_DYNAMIC;
```
aykevl: nit: can you align these?

src/examples/nintendoswitch/debug-hello.go
racerxdl: omg lol

targets/nintendoswitch.ld
```diff
@@ -0,0 +1,81 @@
...
+  . = 0;
+
+  .text : ALIGN(0x1000) {
+    HIDDEN(__text_start = .);
+    KEEP(*(.text.jmp))
+
+    . = 0x80;
+
+    *(.text .text.*)
+    *(.plt .plt.*)
```
aykevl: Where did you get this linker script from?
It contains many things that do not seem to be relevant, for example the PLT or exception handling sections (`.eh_frame`).

src/runtime/runtime_nintendoswitch.go
```diff
@@ -0,0 +1,77 @@
...
+	stdoutBuffer = append(stdoutBuffer, c)
+}
+
+func usleep(usec uint) int {
+	sleepThread(uint64(usec) * 1000)
+	return 0
+}
+
+func abort() {
+	for {
```
niaow, line 61: Shouldn't you be calling exit somewhere here?

targets/nintendoswitch.s
```diff
@@ -0,0 +1,50 @@
+.section .text.jmp, "x"
+.global _start
+_start:
+        b start
+	.word _mod_header - _start
```
aykevl: This file uses a mix of tabs and spaces. Can you format it to use one or the other?
(My preference is spaces but that doesn't matter much, as long as it's consistent).

deadprogram:
Looks like a merge conflict needs to be resolved, and just a few more feedback items for this PR to get this to land, correct?

src/runtime/runtime_nintendoswitch.s
racerxdl: I tried to use `arm.SVCall2` but it uses `r0` which is 32 bit, and the aarch64 requires a 64 bit register `x0`.

If I go to `compiler/inlineasm.go` and change the function `emitSVCall(args []ssa.Value)`:

```go
constraints := "={r0}"
```
to 
```go
constraints := "={x0}"
```

and 

```
constraints += ",{r" + strconv.Itoa(i) + "}"
```

to 

```
constraints += ",{x" + strconv.Itoa(i) + "}"
```

it does work.

So to avoid changing that inline asm code and not having a .s file I'm doing this:

```go
func NxOutputString(str *byte, size uint64) uint64 {
    res := arm.AsmFull(
        `mov x0, {str}
         mov x1, {size}
         svc 0x27`,
        map[string]interface{}{
            "str":  uintptr(unsafe.Pointer(str)),
            "size": size,
        })
	return uint64(res)
}
```

Is that better?

racerxdl:
Fixed @aykevl / @deadprogram !

src/runtime/runtime_nintendoswitch.s
deadprogram: That makes more sense even to me. @aykevl seems legit?

deadprogram:
@jaddr2line I think that all of your feedback has been addressed.

src/runtime/runtime_nintendoswitch.s
niaow: I think we should fix `emitSVCall` eventually, I guess this is fine for now

deadprogram:
@jaddr2line by which I mean https://github.com/tinygo-org/tinygo/pull/1226#pullrequestreview-446836840

src/runtime/runtime_nintendoswitch.s
aykevl: Adding support for aarch64 should be relatively easy. @racerxdl do you have any information on the clobbered/retained registers after a `svc` instruction? On arm it's the usual calling convention hence the clobbered registers.
The inline assembly you posted above may not be correct because of clobbered registers, so if you don't know it may be best to assume the usual calling convention (which will be used when you call an assembly function).

Makefile
```diff
@@ -342,6 +342,7 @@ endif
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=pca10040 -opt=1     examples/blinky1
 	@$(MD5SUM) test.hex
+	$(TINYGO) build             -o test.elf  -target=nintendoswitch     examples/nintendoswitch
```
aykevl: Nit: please also add the `@$(MD5SUM) test.hex`.

src/runtime/runtime_nintendoswitch.s
racerxdl: Hi @aykevl , yes they're all mapped here: https://switchbrew.org/wiki/SVC

Which in the case of this SVC is this: https://switchbrew.org/wiki/SVC#OutputDebugString

image

src/examples/nintendoswitch/debug-hello.go
```diff
@@ -0,0 +1,17 @@
...
+// emulator console
+package main
+
+import (
+	"runtime"
+	"time"
+)
+
+func main() {
+	runtime.NxOutputString("Hello world!")
```
aykevl: In principle, `examples/serial` should work as well. Would that work as a smoke test? It would reduce the amount of code specific to the Switch and proves that `println` works correctly.

src/syscall/syscall_libc.go
```diff
@@ -1,26 +1,27 @@
...
-	return ENOSYS // TODO
+	return ENOSYS
 }
 
 func Write(fd int, p []byte) (n int, err error) {
 	buf, count := splitSlice(p)
 	n = libc_write(int32(fd), buf, uint(count))
 	if n < 0 {
-		err = getErrno()
+		err = errors.New("error writing to fd") // TODO: Return errno
```
aykevl: I would prefer if you defined `getErrno` so that errno remains correct for other platforms (for example macOS). If there is no way to get the errno currently, you can implement a dummy `getErrno` (in syscall_nintendoswitch.go).

```go
var dummyError = errors.New("unknown syscall error")
func getErrno() error {
    return dummyError
}
```

src/syscall/syscall_libc.go
```diff
@@ -60,6 +61,5 @@ func splitSlice(p []byte) (buf *byte, len uintptr) {
 	return slice.buf, slice.len
 }
 
-// ssize_t write(int fd, const void *buf, size_t count)
 //export write
-func libc_write(fd int32, buf *byte, count uint) int
+func libc_write(fd int32, buf *byte, size uint) int
```
aykevl: There are still a few changes in this file that look unintentional. They're not necessarily bad, but probably shouldn't be in this PR.

src/syscall/syscall_nintendoswitch.go
```diff
@@ -0,0 +1,41 @@
+// +build nintendoswitch
+
+package syscall
+
+// A Signal is a number describing a process signal.
+// It implements the os.Signal interface.
+
```
aykevl: Nit: this empty line removes the doc string from the `Signal` type.

targets/nintendoswitch.s
```diff
@@ -0,0 +1,50 @@
+.section .text.jmp, "x"
+.global _start
+_start:
+        b start
+  .word _mod_header - _start
```
aykevl: Can you format this file? Again, I don't care too much about how it's exactly formatted but right now the indentation is a bit messy.

src/runtime/runtime_nintendoswitch.s
aykevl: That's not what I mean. Often, such syscall instructions (`svc` is very similar to the amd64 `syscall` instruction) clobber a number of registers, possibly including their input registers. This means the registers many have any value. For a function call to an unknown function (such as assembly) the compiler knows that some registers will be clobbered and won't assume the values in the registers remain the same. For regular inline assembly,  the compiler is not aware of this and it may (or may not) lead to corruption at runtime.

aykevl: The `arm.SVCall` functions are implemented in a way to be aware of this: they clobber `r1`, `r2` and `r3` following the ARM calling convention (`r0` is an output register and thus is implicitly clobbered). Inline assembly does not automatically mark these registers as clobbered.

racerxdl: Ahhh, so this function just inlines whatever I put in that assembly statement inside the function its being called?

 Well I'm not sure to be honest. The SVC calls in both libnx and libtransistor are done in pure assembly but called from C (they're exported as normal functions).

You think its safer to keep then in a `.s` file just to avoid register corruption?

deadprogram: @aykevl what do you think?

Makefile
deadprogram: @racerxdl please see this comment.

src/examples/nintendoswitch/debug-hello.go
deadprogram: @racerxdl also this comment.

src/syscall/syscall_libc.go
deadprogram: LIkewise @racerxdl this comment too.

src/syscall/syscall_libc.go
deadprogram: @racerxdl this one.

src/syscall/syscall_nintendoswitch.go
deadprogram: @racerxdl this as well.

targets/nintendoswitch.s
deadprogram: @racerxdl this is the last one.

src/examples/nintendoswitch/debug-hello.go
racerxdl: Just tested, it does indeed work because of `putchar` implementation. I removed this nintendo switch example and changed the smoke test to build examples/serial.

Makefile
racerxdl: Done

src/syscall/syscall_libc.go
racerxdl: Done!

src/syscall/syscall_nintendoswitch.go
racerxdl: Done!

src/syscall/syscall_libc.go
racerxdl: Done!

targets/nintendoswitch.s
racerxdl: omg, that file is full broken. Fixed now.

src/runtime/runtime_nintendoswitch.s
racerxdl: @deadprogram / @aykevl I think that's the only missing point now.

deadprogram: @aykevl anything on this?

Makefile
```diff
@@ -342,6 +342,8 @@ endif
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=pca10040 -opt=1     examples/blinky1
 	@$(MD5SUM) test.hex
+	$(TINYGO) build             -o test.elf  -target=nintendoswitch     examples/serial
+	@$(MD5SUM) test.elf
```
aykevl: Sorry to be nitpicking here but you need `test.hex` as output to be able to get a proper md5sum. Otherwise most trivial changes (even a new comment somewhere) may cause a different output.

src/runtime/runtime_nintendoswitch_svc.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+// Result nxOutputString(const char *str, u64 size)
+func NxOutputString(str string) uint64 {
+	strData := (*_string)(unsafe.Pointer(&str))
+
+	res := arm.AsmFull(
+		`mov x0, {str}
+         mov x1, {size}
+         svc 0x27`,
```
aykevl: Looking at this again, this cannot be safe. The compiler has no idea that registers `x0` and `x1` are being overwritten. I would suggest switching back to calling a special function in assembly (at least until `SVCall2` is supported on arm64, which does tell the compiler that a number of registers may be overwritten).

src/syscall/syscall_libc.go
```diff
@@ -60,6 +60,5 @@ func splitSlice(p []byte) (buf *byte, len uintptr) {
 	return slice.buf, slice.len
 }
 
-// ssize_t write(int fd, const void *buf, size_t count)
```
aykevl: Nit: you probably didn't intend to remove this line?

aykevl (commented):
This is really close, just two nits and a comment on `NxOutputString`. I'm afraid the current code just happens to work but may stop working when the compiler decides to arrange things slightly differently.

Makefile
racerxdl: Weird, if I try with `hex` it gives me this error:

```
$ tinygo build             -o test.hex  -target=nintendoswitch     examples/serial
error: ROM segments are non-contiguous: /tmp/tinygo527744533/main
```

src/runtime/runtime_nintendoswitch.s
racerxdl: Ok reverted to use assembly code instead.

src/runtime/runtime_nintendoswitch_svc.go
racerxdl: Reverted to assembly code :D

src/syscall/syscall_libc.go
racerxdl: Oh, yeah. I thought I added it. Let me rollback that.

aykevl:
@racerxdl can you rebase this PR on top of the dev branch? That should fix the assert-test-linux test failure.

The tinygo-CI (Windows) test also fails but that's a separate issue not caused by this PR.

racerxdl:
> @racerxdl can you rebase this PR on top of the dev branch? That should fix the assert-test-linux test failure.
> 
> The tinygo-CI (Windows) test also fails but that's a separate issue not caused by this PR.

Ok just did the rebase! Lets see if it fixes ^^

deadprogram:
The Windows build failure was fixed by https://github.com/tinygo-org/tinygo/pull/1261

If you rebase from dev it should now pass.

racerxdl:
@deadprogram it did it \o/

aykevl:
Looking good!

aykevl:
@racerxdl I was looking into adding support for `tinygo run` to Nintendo Switch support, however I couldn't get it to work.
I tried the following:

```
tinygo build -o test.elf -target=nintendoswitch examples/test
elf2nro test.elf test2.nro
yuzu test2.nro
```

However, with that the program is stuck at the loading screen.
Where can I find the debug output, if there are any?

I confirmed that `yuzu` itself is working, with a .nro file that I created earlier.

racerxdl:
Hi @aykevl , just change the log filter to `*:Debug` in Yuzu Configuration -> General -> Debug:

image


The print statements should appear in the console window:

image


Since the code does not initialize the GPU, it will show the Loading screen forever. But should run the code just fine.

Next week my vacation starts, I will start implementing the IPC stuff :100:

aykevl:
Okay, thanks a lot!
The debug console itself doesn't appear on Linux (it's Windows-only), but enabling debug messages lets me see them on stderr where I can filter them as usual. So that works.
However, I can't seem to get the expected output. I get a lot of these messages:

```
[   2.637636] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: [   2.637816] HW.Memory <Error> core/memory.cpp:Write:669: Unmapped Write64 0x00000000 @ 0x0000000000004038
```

However, the program itself appears to be running because the number of `OutputDebugString` error messages almost matches the number of expected lines of output (it's one less with examples/test, but I get the expected number of lines with testdata/alias.go and testdata/print.go).

racerxdl:
Hey @aykevl on linux it does work fine if you open yuzu from a terminal (I do that here, the print screen is from linux).

There is the bug I mentioned in slack which in some cases causes a invalid memory read / write which we couldnt find out why (it depends a lot how the code is done, I'm still trying to debug it. Its easier to test with a panic).

But the example you compiled works, besides some Unmapped R/W:

```
[  19.002905] Core <Error> core/core.cpp:Load:238: Failed to find title id for ROM (Error 0)
[  19.004145] Frontend <Info> yuzu/main.cpp:BootGame:1043: Booting game: 0000000000000000 | test.nro
[  19.081462] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read64 @ 0x00004040
[  19.081467] HW.Memory <Error> core/memory.cpp:Write:668: Unmapped Write64 0x08303FC8 @ 0x0000000000004040
[  19.081491] HW.Memory <Debug> core/memory.cpp:MapPages:561: Mapping 00000000C0000000 onto 0000002088200000-0000002108200000
[  19.725916] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read64 @ 0x00004040
[  19.725922] HW.Memory <Error> core/memory.cpp:Write:668: Unmapped Write64 0x08303E00 @ 0x0000000000004040
[  19.726231] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: Hello world from Go!
[  19.726509] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: The answer is: 42
[  19.726547] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: 5 ** 2 = 25
[  19.726574] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: 3 + 12 = 15
[  19.726771] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: fib(11) = 89
[  19.726805] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: sumrange(100) = 5050
[  19.726834] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: strlen foo: 3
[  19.727746] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read64 @ 0x00004040
[  19.727750] HW.Memory <Error> core/memory.cpp:Write:668: Unmapped Write64 0x08303C38 @ 0x0000000000004040
[  19.728670] HW.Memory <Error> core/memory.cpp:Write:668: Unmapped Write64 0x00000000 @ 0x0000000000004040
[  19.728712] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read64 @ 0x00004040
[  19.728715] HW.Memory <Error> core/memory.cpp:Write:668: Unmapped Write64 0x08303C38 @ 0x0000000000004040
[  19.728719] HW.Memory <Error> core/memory.cpp:Write:668: Unmapped Write64 0x00000000 @ 0x0000000000004040
[  19.728869] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: map length: 2
[  19.729356] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: map read: answer = 42
[  19.729703] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624:   answer = 42
[  19.729742] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624:   foo = 3
[  19.729900] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: map length: 1
[  19.729904] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x00003020
[  19.729923] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x00003021
[  19.729926] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x00003022
[  19.729926] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x00003023
[  19.729927] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x00003024
[  19.729927] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x00003025
[  19.729928] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x00003026
[  19.729929] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x00003027
[  19.729947] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read64 @ 0x00003028
[  19.729994] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: map read: data = 0
[  19.729997] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x00003020
[  19.729998] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x00003021
[  19.729999] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x00003022
[  19.729999] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x00003023
[  19.730000] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x00003024
[  19.730001] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x00003025
[  19.730001] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x00003026
[  19.730002] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x00003027
[  19.730003] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read64 @ 0x00003028
[  19.730134] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: len/cap foo: 4 4
[  19.730171] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: len/cap bar: 3 5
[  19.730210] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: len/cap foo[1:2]: 1 3
[  19.730237] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: foo[3]: 5
[  19.730379] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: sum foo: 12
[  19.730450] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: copy foo -> bar: 3
[  19.730497] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: sum bar: 7
[  19.730690] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: thing: foo
[  19.730824] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: is int: 5
[  19.730871] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: is byte: 120
[  19.730964] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x000020B4
[  19.730967] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x000020B5
[  19.730967] HW.Memory <Error> core/memory.cpp:Read:627: Unmapped Read8 @ 0x000020B6
[  19.730969] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: is string: [  19.731128] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: is Thing: foo
[  19.731232] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: is *Thing: foo
[  19.731270] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: is *Thing: foo
[  19.731341] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: is Doubler: 6
[  19.731477] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: Stringer.String(): foo
[  19.731582] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: Stringer.(*Thing).String(): foo
[  19.731698] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: hello from function pointer: 5
[  19.731793] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: deferring...
[  19.731863] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: ...run as defer 2
[  19.731874] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: ...run as defer 1
[  19.732013] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: bound method: foo
[  19.732099] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: thing inside closure: foo
[  19.732208] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: inside fp closure: foo 3
[  19.732342] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: lower to upper char: h -> H
[  19.732433] HW.Memory <Error> core/memory.cpp:Write:668: Unmapped Write64 0x00000000 @ 0x0000000000004040
[  19.732452] Kernel.SVC <Info> core/hle/kernel/svc.cpp:ExitProcess:1376: Process 81 exiting
```

With only the OutputDebugString lines:

```
[  19.726231] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: Hello world from Go!
[  19.726509] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: The answer is: 42
[  19.726547] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: 5 ** 2 = 25
[  19.726574] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: 3 + 12 = 15
[  19.726771] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: fib(11) = 89
[  19.726805] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: sumrange(100) = 5050
[  19.726834] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: strlen foo: 3
[  19.728869] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: map length: 2
[  19.729356] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: map read: answer = 42
[  19.729703] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624:   answer = 42
[  19.729742] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624:   foo = 3
[  19.729900] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: map length: 1
[  19.729994] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: map read: data = 0
[  19.730134] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: len/cap foo: 4 4
[  19.730171] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: len/cap bar: 3 5
[  19.730210] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: len/cap foo[1:2]: 1 3
[  19.730237] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: foo[3]: 5
[  19.730379] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: sum foo: 12
[  19.730450] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: copy foo -> bar: 3
[  19.730497] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: sum bar: 7
[  19.730690] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: thing: foo
[  19.730824] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: is int: 5
[  19.730871] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: is byte: 120
[  19.730969] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: is string:
[  19.731128] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: is Thing: foo
[  19.731232] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: is *Thing: foo
[  19.731270] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: is *Thing: foo
[  19.731341] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: is Doubler: 6
[  19.731477] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: Stringer.String(): foo
[  19.731582] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: Stringer.(*Thing).String(): foo
[  19.731698] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: hello from function pointer: 5
[  19.731793] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: deferring...
[  19.731863] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: ...run as defer 2
[  19.731874] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: ...run as defer 1
[  19.732013] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: bound method: foo
[  19.732099] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: thing inside closure: foo
[  19.732208] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: inside fp closure: foo 3
[  19.732342] Debug.Emulated <Debug> core/hle/kernel/svc.cpp:OutputDebugString:624: lower to upper char: h -> H
```

I still want to get the VT52 example working with pure golang stuff. I have two week vacation, so lets see if I can make it since its a better example ^^

aykevl:
Well for me that doesn't work. However, I'm suspecting I see a linker error (or a bad linker/elf2nro interaction) that I'm now investigating. It could explain why none of the strings were printed properly.

When I remove some apparently unnecessary sections, yuzu crashes with a SIGSEGV so that's interesting at least.

racerxdl:
That's odd. Which version of yuzu are you running?
Just as a reference:

Thats my elf2nro
```
sha1sum /opt/devkitpro/tools/bin/elf2nro
467b694c206ea4ebd5d774f5b9f1e8d86f02e78e  /opt/devkitpro/tools/bin/elf2nro
```

racerxdl:
Ah, I just noticed there was some changes in my local compiler, maybe they're related. I will check that here and return in few moments.

racerxdl:
@aykevl just confirmed that my changes fixed, but I haven't published them. I will do that tonight and send a PR fixing that. The upstream `dev` branch gives the Unmapped Write / Read errors for every OutputDebugString.

aykevl:
Okay, that would be great! Right now I'm looking into whether it's possible to skip the elf2nro / linkle step and generate .nro files directly with the linker script (using `objcopy -O binary`). So far it seems to be possible, but I'm hitting some issues.

racerxdl:
@aykevl I can check that as well this weekend if you want. There are other projects that does that (I currently use linkle which is made in rust). Maybe we can do a small snippet in tinygo to generate the NRO file directly.

aykevl:
I have done it, with some assembly and linker script magic I made TinyGo output .nro files directly:

https://github.com/tinygo-org/tinygo/compare/dev...nintendoswitch-nro

I'll wait with a PR until you've submitted the fix, as I can't fully test it right now.

racerxdl:
@aykevl this fixes here the outputs: https://github.com/tinygo-org/tinygo/pull/1368

PR #1229, opened 2020-07-10
Detect stack overflow via memory protections

firelizzard18:
On the Teensy 3.6 (NXP MK66F18), a hard fault is *not* triggered when the stack overflows past the end of RAM. Instead it appears that the microcontroller resets. To handle stack overflow properly, I've used the memory protection unit to mark the last 1kB of RAM as inaccessible, which forces a fault.

### DIFF
--- a/src/runtime/runtime_nxpmk66f18.go
+++ b/src/runtime/runtime_nxpmk66f18.go
@@ -35,6 +35,8 @@ import (
 	"device/arm"
 	"device/nxp"
 	"machine"
+	"runtime/volatile"
+	"unsafe"
 )
 
 const (
@@ -56,6 +58,7 @@ func main() {
 	initSystem()
 	arm.Asm("CPSIE i")
 	initInternal()
+	initMemoryProtection()
 
 	run()
 	abort()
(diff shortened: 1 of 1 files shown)


firelizzard18:
Calling `f()` now causes a fault (before, it would repeat forever):

```
stackTop: 0x1fff0800
globalsStart: 0x1fff0800
globalsEnd: 0x1fff10fc
heapStart: 0x1fff10fc
heapEnd: 0x20030000

0x1fff3100
0x1fff31d0
0x1fff32a0
...
0x2002f980
0x2002fa50
0x2002fb20
fatal error: data bus error (imprecise)
```

```golang
func f(n int) int {
	np := uintptr(unsafe.Pointer(&n))
	println(np)
	return f(n+1) + 1
}
```

However, the heap, globals, and stack bounds concern me. Is that what they should be? I assume that `n int` is passed on the stack, but the addresses look like they belong to the heap. Is the Teensy's linker script doing something wrong? RAM extends from `0x1FFF0000` to `0x2002FFFF`.

niaow:
> I assume that n int is passed on the stack, but the addresses look like they belong to the heap.

No, it actually ends up being allocated on the heap, since we have not yet implemented a way to tell the optimizer that printing the pointer does not escape it (comparisons are considered escaping operations unless you add additional constraints because. . . C). That also does not really matter because the compiler has no need to implement `f` recursively. Your function is being tail-call optimized. In other words, what the compiler is actually doing:
```Go
func f(n int) {
    for m := n; ; m++ {
        np := runtime.alloc(unsafe.Sizeof(int(0)))
        *(*int)(np) = m
        println(uintptr(np))
    }
}
```

If the compiler realized it did not escape, it would do this instead:
```Go
func f(n int) {
    for {
        println(uintptr(unsafe.Pointer(&n)))
    }
}
```

Additionally, `f` is being run in a goroutine rather than on the system stack. This means that it is on a seperate heap-allocated stack.

I don't know where your fault is coming from, but it is not coming from `f`.

niaow:
```Go
protectStart := heapEnd - 0x400
protectEnd := heapEnd
```

Shouldn't this make a fault trigger when using the last kilobyte of the heap? This section of the heap will be used regularly, and is the most likely explanation I can find as to your fault. Shouldn't this be:
```Go
protectStart := heapEnd
protectEnd := heapEnd + 0x400
```

Or am I missing something?

niaow:
Also, the stack grows downward rather than upward, so I think you would really need to protect `heapEnd` through `0xFFFFFFFF` because a system stack overflow would mean that the stack pointer was underflowing.

firelizzard18:
> Shouldn't this be: [heapEnd, heapEnd+1K]

The protected region must be within the address range for the RAM. On the K66's memory map, there is a reserved section after the SRAM. Accessing this region does not cause a fault, but seems to reset the processor. Protecting this region does not work, so the protected region must be at the end of RAM. A better solution than mine would be moving the heap end back by 1K, so the heap ends before the protected region.

> I don't know where your fault is coming from, but it is not coming from `f`.

From the log messages and fault registers, it's pretty clear that `f` is causing the error one way or another. Calls to `f` are obviously filling up the heap with allocations (based on the logged addresses), and the NXP's MPU error registers indicate an attempted write to an address within the protected region. If I add `if np >= 0x2002f800 { return 0 }` to `f`, no fault occurs and everything continues happily.

niaow:
Ok, so if I understand correctly, you need to actually reserve the region rather than leaving it in the heap.

firelizzard18:
I think I've been seriously misunderstanding what's happening. I think what I've been interpreting as an overflow causing a processor reset is actually... the garbage collector doing its job.

The root of my confusion is that the default HardFault handler reports a stack overflow in many situations where that is not the case. In fact, I don't think I've ever seen an actual stack overflow. So I've been trying to manually produce one, to see what the Teensy's behavior is for a real overflow. But I haven't actually made anything except an infinite loop that pressures the GC...

niaow:
It is also worth noting that none of these allocations are held for very long. If everything is configured right, then this should never terminate.

firelizzard18:
That agrees with what I was seeing, but not with my expectations, so I assumed something bad was happening.

firelizzard18:
A real stack overflow produces a hard fault, so I'm closing this

PR #1231, opened 2020-07-10
builder: try to determine stack size information at compile time

aykevl:
For now, this is just an extra flag that can be used to print stack
frame information, but this is intended to provide a way to determine
stack sizes for goroutines at compile time in many cases.

Stack sizes are often somewhere around 350 bytes so are in fact not all
that big usually. Once this can be determined at compile time in many
cases, it is possible to use this information when available and as a
result increase the fallback stack size if the size cannot be determined
at compile time. This should reduce stack overflows while at the same
time reducing RAM consumption in many cases.

Interesting output for testdata/channel.go:

    function                                 stack usage (in bytes)
    Reset_Handler                            332
    .Lcommand-line-arguments.fastreceiver    220
    .Lcommand-line-arguments.fastsender      192
    .Lcommand-line-arguments.iterator        192
    .Lcommand-line-arguments.main$1          184
    .Lcommand-line-arguments.main$2          200
    .Lcommand-line-arguments.main$3          200
    .Lcommand-line-arguments.main$4          328
    .Lcommand-line-arguments.receive         176
    .Lcommand-line-arguments.selectDeadlock  72
    .Lcommand-line-arguments.selectNoOp      72
    .Lcommand-line-arguments.send            184
    .Lcommand-line-arguments.sendComplex     192
    .Lcommand-line-arguments.sender          192
    .Lruntime.run$1                          548

This shows that the stack size (if these numbers are correct) can in
fact be determined automatically in many cases, especially for small
goroutines. One of the great things about Go is lightweight goroutines,
and reducing stack sizes is very important to make goroutines
lightweight on microcontrollers.

---

This PR does not directly depend on #1225 but is more useful with that PR.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -4,11 +4,13 @@
 package builder
 
 import (
+	"debug/elf"
 	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -16,6 +18,7 @@ import (
 	"github.com/tinygo-org/tinygo/compiler"
 	"github.com/tinygo-org/tinygo/goenv"
 	"github.com/tinygo-org/tinygo/interp"
+	"github.com/tinygo-org/tinygo/stacksize"
 	"github.com/tinygo-org/tinygo/transform"
 	"tinygo.org/x/go-llvm"
 )
@@ -216,6 +219,11 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(stri
 			}
 		}
 
+		// Print goroutine stack sizes, as far as possible.
+		if config.Options.PrintStacks {
+			printStacks(mod, executable)
+		}
+
 		// Get an Intel .hex file or .bin file from the .elf file.
 		if outext == ".hex" || outext == ".bin" || outext == ".gba" {
 			tmppath = filepath.Join(dir, "main"+outext)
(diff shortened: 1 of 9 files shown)


aykevl:
Some ideas what this can be useful for:

* Determine stack size for new goroutines (obviously).
* Allocate main goroutine statically, thus avoiding a heap allocation at startup. This may be useful to avoid the GC and might allow optimizing it away for simple programs. This may make the tasks scheduler more viable on AVR, for example.
* Determine the Cortex-M main stack (for the `Reset_Handler` and interrupts) at compile time. Interrupts should be short, so it's likely their stack sizes can be determined automatically. If the number of interrupt levels is known (often 4 or 8 on Cortex-M chips), then the biggest 4 or 8 interrupt stacks can be summed up to know the maximum stack depth.

Right now it's limited to ARM, but I see no reason why it couldn't be extended to other architectures (as long as they don't do crazy things with stack frames).

Also, because this is binary analysis as much as possible (as opposed to IR based analysis), the traditional toolchain components (such as the linker) can be used as usual in many cases. This will simplify the implementation.

sago35:
I'll try to output the sdcard + filesystem stack size later.

sago35:
> I'll try to output the sdcard + filesystem stack size later.

There seems to be some kind of recursive call.


```
function                         stack usage (in bytes)
Reset_Handler                    308
.Lruntime.run$1                  recursive, (*fmt.pp).printArg may call itself
```

sago35:
I did a little research on the cause.


The following is from src/fmt/print.go (*fmt.pp) in Go1.14. 

```
.Lruntime.run$1                  recursive, (*fmt.pp).printArg may call itself
.Lruntime.run$1                  recursive, (*fmt.pp).badVerb may call itself
```

The following comes from struct-related processing in TinyGo's src/reflect.

```
.Lruntime.run$1                  recursive, (reflect.Type).Field may call itself
```

Even if we are able to resolve these, we are still left with code that appears to be CGO this time.

```
.Lruntime.run$1                  unknown, __aeabi_memcpy does not have stack frame information
```

deadprogram:
@sago35 are you saying that this PR has that problem, or are you saying that the changes in this PR helped you find that problem?

sago35:
@deadprogram 
The changes in this PR helped me find that problem.

Thanks to this PR, we can now see where it is a recursive call.
I think it's very good.

deadprogram:
Thanks for the clarification, @sago35 

Now merging, this is going to be very useful, @aykevl great work!

aykevl:
> The following is from src/fmt/print.go (*fmt.pp) in Go1.14.
> 
> ```
> .Lruntime.run$1                  recursive, (*fmt.pp).printArg may call itself
> .Lruntime.run$1                  recursive, (*fmt.pp).badVerb may call itself
> ```

Yes, I don't think we can do much about that except to avoid the fmt package.

> The following comes from struct-related processing in TinyGo's src/reflect.
> 
> ```
> .Lruntime.run$1                  recursive, (reflect.Type).Field may call itself
> ```

I suspect this is because `(reflect.Type).Field` calls `(reflect.Type).Align`, which again calls `(reflect.Type).Field`. Not sure whether this is fixable other than avoiding reflection (structs may have other struct fields).

> Even if we are able to resolve these, we are still left with code that appears to be CGO this time.
> 
> ```
> .Lruntime.run$1                  unknown, __aeabi_memcpy does not have stack frame information
> ```

This is in fact fixable. The function is implemented in assembly so the compiler has no idea how much stack space it uses (but it can determine the calls it makes through relocations). I have added a hardcoded frame size for `__aeabi_uidivmod` as it appeared in many places, but more should be added over time to improve the algorithm.

PR #1235, opened 2020-07-11
builder: remove optimization level 0

niaow:
Currently, turning optimizations off causes compile failures.
We rely on the optimizer removing some dead symbols.
Avoid providing an option that does not work right now.
In the future once everything has been fixed we can re-enable this.


aykevl:
Seems reasonable, so merged.

I looked into fixing this before, but never figured out how to fix it. Some TinyGo passes introduce dead globals that should be removed but aren't.

PR #1236, opened 2020-07-12
all: fix -gc=none

aykevl:
This option was broken for a long time, in part because we didn't test for it. This commit fixes that and adds a test to make sure it won't break again unnoticed.


niaow (approved):
LGTM

aykevl:
@deadprogram TinyHCI appears to be stuck?

deadprogram:
All passed, now merging.

PR #1237, opened 2020-07-12
internal/task,transform: fix coroutines scheduling order

niaow:
This change applies a loop around resuming a task.
This means that a task will not suspend implicitly when returning from a function.

### DIFF
--- a/src/internal/task/task_coroutine.go
+++ b/src/internal/task/task_coroutine.go
@@ -13,33 +13,46 @@ type rawState uint8
 //export llvm.coro.resume
 func (s *rawState) resume()
 
-type state struct{ *rawState }
+type state struct {
+	*rawState
+
+	// retcon is set to true when a call returns and awakens it caller.
+	retcon bool
+}
 
 //export llvm.coro.noop
 func noopState() *rawState
 
 // Resume the task until it pauses or completes.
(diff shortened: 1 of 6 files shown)


src/internal/task/task_coroutine.go
```diff
@@ -13,33 +13,46 @@ type rawState uint8
 //export llvm.coro.resume
 func (s *rawState) resume()
 
-type state struct{ *rawState }
+type state struct {
+	*rawState
+
+	// retcon is set to true when a call returns and awakens it caller.
+	retcon bool
```
aykevl, line 20: What does `retcon` mean? "Return continue"?

aykevl (commented):
While the change itself seems fine, it does increase code size a little bit on WebAssembly. Therefore, I wonder why this is necessary? The Go specification does not define a particular order between goroutines unless explicit synchronization is used.

If this is for interrupts, I do not think this is a complete fix. I think interrupts are only really safe when used with the tasks scheduler as it is much easier to reason about ordering and heap allocation (no heap allocation on function calls, for example). The tasks scheduler should be implementable on any reasonable baremetal target.

niaow:
I guess I will work on the coroutines replacement instead

PR #1238, opened 2020-07-12
Update code to use ReplaceBits

firelizzard18:
Update existing code to:

```go
// Replace this
x.Set((x.Get() &^ (mask << pos)) | val<<pos)

// With this
x.ReplaceBits(val, mask, pos)
```

ReplaceBits was introduced by #897


deadprogram:
Looks good, thanks for this cleanup @firelizzard18 now merging.

PR #1239, opened 2020-07-13
runtime/volatile: add GetBits

firelizzard18:
Adds another convenience method for registers to read multiple bits at an offset. This change will allow cleaner code:

```go
// instead of
(nxp.MCG.S.Get() & nxp.MCG_S_CLKST_Msk) != (2 << nxp.MCG_S_CLKST_Pos)

// use
nxp.MCG.S.GetBits(nxp.MCG_S_CLKST_11, nxp.MCG_S_CLKST_Pos) != 2
```

### DIFF
--- a/src/runtime/volatile/register.go
+++ b/src/runtime/volatile/register.go
@@ -67,6 +67,16 @@ func (r *Register8) ReplaceBits(value uint8, mask uint8, pos uint8) {
 	StoreUint8(&r.Reg, LoadUint8(&r.Reg)&^(mask<<pos)|value<<pos)
 }
 
+// GetBits is a helper to simplify reading multiple bits at a specific offset.
+// It is the volatile equivalent of:
+//
+//     (r.Reg >> pos) & mask
+//
+// go:inline
+func (r *Register8) GetBits(mask uint8, pos uint8) uint8 {
+	return (LoadUint8(&r.Reg) >> pos) & mask
+}
+
 type Register16 struct {
 	Reg uint16
 }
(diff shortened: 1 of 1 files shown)


deadprogram:
Isn't this mostly the same as https://github.com/tinygo-org/tinygo/blob/dev/src/runtime/volatile/register.go#L56

```
nxp.MCG.S.HasBits(nxp.MCG_S_CLKST_11 << nxp.MCG_S_CLKST_Pos)
```

Also, all of the related functions in volatile return a `bool`.

firelizzard18:
`HasBits` returns whether any bit is set at the given positions, which is why it returns a boolean, whereas `GetBits` extracts the value of the specified bits, regardless of whether they are all set or not, which is why it returns a `uintN`.

| `MCG.S` | `HasBits` | `GetBits` |
| - | - | - |
| `xxxx00xx` | `false` | `00` |
| `xxxx01xx` | `true` | `01` |
| `xxxx10xx` | `true` | `10` |
| `xxxx11xx` | `true` | `11` |

- `nxp.MCG.S.HasBits(nxp.MCG_S_CLKST_11 << nxp.MCG_S_CLKST_Pos)` returns true if `MCG.S` is not `xxxx00xx`
- `nxp.MCG.S.GetBits(nxp.MCG_S_CLKST_11, nxp.MCG_S_CLKST_Pos)` returns `AB` given `MCG.S` is `xxxxABxx`
- `nxp.MCG.S.GetBits(nxp.MCG_S_CLKST_11, nxp.MCG_S_CLKST_Pos) != 2` returns true only if `MCG.S` is `xxxx10xx`

aykevl:
I think there is a better way to do this: by treating these struct fields as bitfields (which they are) and auto generating getters/setters for these bitfields in gen-device-svd. At the moment, `HasBits` is already much too easy to misuse.

Compare this:

```go
// current code
(nxp.MCG.S.Get() & nxp.MCG_S_CLKST_Msk) != (2 << nxp.MCG_S_CLKST_Pos)

// this PR
nxp.MCG.S.GetBits(nxp.MCG_S_CLKST_11, nxp.MCG_S_CLKST_Pos) != 2

// my proposal
nxp.MCG.S.GetCLKST() != 2
```

This would mean that the struct would look more like this (all autogenerated from the SVD file):

```go
type MCG_Type struct {
    // ... other registers
    S MCG_S_Type
    // ... other registers
}

type MCG_S_Type struct {
    volatile.Register8
}

//go:inline
func (field *MCG_S_Type) GetCLKST() uint8 {
    return (field.Register8.Get() & nxp.MCG_S_CLKST_Msk) >> nxp.MCG_S_CLKST_Pos
}
```

Other options: adding setters, treating single-bit fields as booleans (returning a `bool` instead of an integer), ...
Eventually we could hopefully replace and remove `HasBits`/`SetBits`/`ReplaceBits` entirely.

deadprogram:
Now closing this, since we have added generated function wrappers for many of the bit values in question. Thank you everyone!

PR #1241, opened 2020-07-13
Cleanup Teensy 3.6 linker script

firelizzard18:
I replaced most of `nxpmk66f18.ld` with `INCLUDE "targets/arm.ld"`. I left the `.text` section, since the K66 needs that arranged specially, but I verified (with `objdump`) that the linker script change does not change the binaries produced.


deadprogram:
Looks great @firelizzard18 :scissors: 

Now merging, thanks!

PR #1242, opened 2020-07-13
Added board definition for Feather nRF52840 Express

bgould:
New board definition for Adafruit Feather nRF52840 Express

https://www.adafruit.com/product/4062


deadprogram:
I do not have one to test with, but seems correct to me.

Now merging, thanks @bgould I am going to use this PR as basis to add the ItsyBitsy nrf52840.

Thank you!

Issue #1245, opened 2020-07-15
transform (gc): GEP not pinned


lei-april:
The following code produces different results between the official Go compiler and Tinygo.

```go
package main

const DataLen = 32

type Foo struct {
	data [DataLen]byte
}

func MakeFoo(from []byte) *Foo {
	foo := new(Foo)
	copy(foo.data[:], from)
	return foo
}

func main() {
	foo := MakeFoo([]byte("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"))
	println(string(foo.data[:]))

	bar := MakeFoo([]byte("bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"))
	println(string(bar.data[:]))
}
```

With official Go:
```
$ go run t.go
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
```

With TinyGo:
```
$ tinygo build -o t t.go && ./t
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

```

If I change the field type of `Foo` from array to slice, then TinyGo produces the same result as official Go.

```diff
@@ -3,11 +3,12 @@ package main
 const DataLen = 32

 type Foo struct {
-   data [DataLen]byte
+   data []byte
 }

 func MakeFoo(from []byte) *Foo {
    foo := new(Foo)
+   foo.data = make([]byte, DataLen)
    copy(foo.data[:], from)
    return foo
 }
```

I'm not sure if there's something wrong with the code I posted, or it's a bug in TinyGo. Any idea?

niaow:
This appears to somehow be an issue with the `extalloc` GC that we use on *nix systems. It seems to work with our other GC, although I am not entirely sure what is going on yet.

niaow:
Okay, so it looks like the return value of `MakeFoo` (in `main`) is being compiled incorrectly, and is not being added as a pointer pinned by the GC. Then when you convert it to a string `string([]byte(foo.data[:]))` it does another allocation. On the second time, you appear to be hitting the first GC high water mark, triggering the GC to run and freeing `foo` and `bar`.

niaow:
[...]

PR #1247, opened 2020-07-15
transform: track 0-index GEPs

niaow:
It appears that LLVM is turning bitcasts into 0-index GEPs. This caused stuff to not be tracked, resulting in use-after-free issues. This solution is sub-optimal, but is the most reasonable solution I could come up with without redesigning the stack slots pass.


niaow:
Fixes #1245

aykevl:
Looks reasonable. I feel like this may not fix the root cause, but it fixes a bug without being too intrusive so it's fine.

PR #1252, opened 2020-07-16
arm: use CFI directives for stack usage

aykevl:
Call Frame Information is stored in the .debug_frame section and is used
by debuggers for unwinding. For assembly, this information is not known.
Debuggers will normally use heuristics to figure out the parent function
in the absence of call frame information.

This usually works fine, but is not enough for determining stack sizes.
Instead, I hardcoded the stack size information in
stacksize/stacksize.go, which is somewhat fragile. This change uses CFI
assembly directives to store this information instead of hardcoding it.

This change also fixes the following error message that would appear in
GDB:

More information on CFI:
  * https://sourceware.org/binutils/docs/as/CFI-directives.html
  * https://www.imperialviolet.org/2017/01/18/cfi.html

### DIFF
--- a/src/runtime/scheduler_cortexm.S
+++ b/src/runtime/scheduler_cortexm.S
@@ -1,12 +1,21 @@
+// Only generate .debug_frame, don't generate .eh_frame.
+.cfi_sections .debug_frame
+
 .section .text.tinygo_startTask
 .global  tinygo_startTask
 .type    tinygo_startTask, %function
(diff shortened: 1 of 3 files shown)


deadprogram:
Any feedback on this @jaddr2line ?

niaow (approved):
LGTM

deadprogram:
Cool, now merging. Thanks!

PR #1253, opened 2020-07-18
Automatically determine stack size

aykevl:
This is a big change that will determine the stack size for many
goroutines automatically. Functions that aren't recursive and don't call
function pointers can in many cases have an automatically determined
worst case stack size. This is useful, as the stack size is usually much
lower than the previous hardcoded default of 1024 bytes: somewhere
around 200-500 bytes is common.

A side effect of this change is that the default stack sizes (including
the stack size for other architectures such as AVR) can now be changed
in the config JSON file, making it tunable per application.

---

I checked the binary size impact:

* Non-ARM chips are not affected.
* For ARM chips, there is usually a small increase in binary size (around 30 bytes) and a small increase in static RAM consumption (usually 4 bytes).

I think this is worth the cost as RAM usage may be significantly reduced for smaller programs and RAM is more valuable than flash.

Because this is a big change, it should not be included in the next release but sit a while in the dev branch for testing.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -5,6 +5,7 @@ package builder
 
 import (
 	"debug/elf"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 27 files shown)


aykevl:
...well, it appears I still have things to do before this is ready. There are a number of test failures.

aykevl:
Worked around a bug in the LLVM 9 assembler, which should fix the test-llvm9-go111 failure.

niaow (approved):
LGTM

deadprogram:
This is a pretty intense change, thanks for looking it over @niaow and thanks for all the work that went into it @aykevl 

Going to merge now, let's see how it performs in the wild!

PR #1259, opened 2020-07-23
ci: do not cache TinyGo cache dir

aykevl:
It doesn't seem to speed up the build and it causes issues with a stale cache.

---

Note: this might not fix the problem right away, it may need another CI run to fully take effect (as the list of directories to cache is not checked when restoring the cache).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,7 +93,6 @@ commands:
           key: wasi-libc-sysroot-systemclang-v1
           paths:
             - lib/wasi-libc/sysroot
-      - run: tinygo clean
       - run: go test -v -tags=llvm<<parameters.llvm>> ./cgo ./compileopts ./interp ./transform .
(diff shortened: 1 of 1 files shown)


aykevl:
```
main_test.go:156: # syscall
main_test.go:156: ../.cache/tinygo/goroot-go1.11.13-3b52482cdcddfd5e3da9a7e43c690743bd4e95dab23234f22a2a77b16dbbb0a4-syscall/src/time/sys_unix.go:11:2: open /home/circleci/.cache/tinygo/goroot-go1.11.13-3b52482cdcddfd5e3da9a7e43c690743bd4e95dab23234f22a2a77b16dbbb0a4-syscall/src/syscall/syscall_nintendoswitch.go: no such file or directory
```
See that's what I mean. I restarted the build, rerunning it should fix it.

deadprogram:
Now merging.

PR #1260, opened 2020-07-23
main: add -target flag to tests

aykevl:
This makes it easy to test one particular architecture, for example:

    go test -v -target=hifive1-qemu

This speeds up testing and allows testing targets that are not included in the test by default (such as RISC-V tests on Linux).


deadprogram:
That is pretty useful! Now merging.

PR #1261, opened 2020-07-23
ci: fix Windows QEMU version

aykevl:
It appears that version 2020.07.22 or 2020.07.23 introduced a breaking
change in RISC-V. We will have to fix this eventually, but for now it's
easiest to just pin the QEMU version. Once this new QEMU version
(version 5?) is more widely available, it becomes easier to debug and
fix the underlying cause.


aykevl:
Yup, tests pass.

deadprogram:
Merging, thanks @aykevl !

Issue #1262, opened 2020-07-23
Dev Branch / 0.14 - compiler Panics    "panic: interp: load from a bitcast"


markfarnan:
Dev Branch -  Commit:  50a677e9b791bbf34eaa9effca231a036a284ce8
Target:  WASM

Compiles without error in : 0.13

```
panic: interp: load from a bitcast

goroutine 1 [running]:
github.com/tinygo-org/tinygo/interp.(*LocalValue).Load(0xc00c7c2250, 0xc00c7c2250)
        /home/dev/external/tinygo/interp/values.go:61 +0x220
github.com/tinygo-org/tinygo/interp.(*LocalValue).Load(0xc00c7c2240, 0x7f3a0000001d)
        /home/dev/external/tinygo/interp/values.go:58 +0x12f
github.com/tinygo-org/tinygo/interp.(*frame).evalBasicBlock(0xc0065c4990, 0x7f3aa087a9d0, 0x0, 0xc00c7c12a0, 0xc, 0xc000080000, 0x7f3b0084a108, 0x0, 0x7f3a00000000, 0x0, ...)
        /home/dev/external/tinygo/interp/frame.go:98 +0xd684
github.com/tinygo-org/tinygo/interp.(*evalPackage).function(0xc0065c74e0, 0x7f3ac807d2f8, 0xc00920d740, 0x3, 0x4, 0xc00c7c12a0, 0xc, 0x4, 0x0, 0x0)
        /home/dev/external/tinygo/interp/interp.go:113 +0x238
github.com/tinygo-org/tinygo/interp.(*frame).evalBasicBlock(0xc0065c5750, 0x7f3acab22b60, 0x0, 0xc00c7bbe58, 0x8, 0xc000080000, 0x7f3b0084a108, 0x0, 0x7f3a00000000, 0x0, ...)
        /home/dev/external/tinygo/interp/frame.go:570 +0x6463
github.com/tinygo-org/tinygo/interp.(*evalPackage).function(0xc0065c74e0, 0x7f3ac8043c08, 0xc009f46100, 0xa, 0x10, 0xc00c7bbe58, 0x8, 0x10, 0x0, 0x0)
        /home/dev/external/tinygo/interp/interp.go:113 +0x238
[...]

PR #1263, opened 2020-07-23
interp: replace some panics with error messages

aykevl:
A number of functions now return errors instead of panicking, which should help greatly when investigating interp errors. It at least shows the package responsible for it.

---

Not everything is covered yet, but it does increase the number of panics greatly. In particular, it should show a proper error message for `Value.Load`, which leads to a problem in #1262.


deadprogram:
This would have been really useful for tracking down some driver errors I was looking into last week!

Now merging, thank you @aykevl

PR #1265, opened 2020-07-24
interp: fix sync/atomic.Value load/store methods

aykevl:
These methods do some unsafe pointer casting but can be assumed to not have significant side effects. Simply call these functions at runtime instead of compile time.

This is a partial fix for importing image/png, see #1262.


aykevl:
I have added a second commit that should fix #1262 entirely.

@jaddr2line I believe this is correct, but can you please double check that I didn't miss anything?

niaow:
@deadprogram hardware ci seems to be having issues

niaow (approved):
This seems correct to me. In the future we should really try to add tests for this kind of stuff.

deadprogram:
TinyHCI restarted, all tests passing.

deadprogram:
Now merging.

PR #1275, opened 2020-08-01
Reduce current consumption when using the SoftDevice

aykevl:
This PR reduces current consumption in sleep mode when using the SoftDevice on an nrf.


deadprogram:
Makes sense to me, so now merging. Thanks @aykevl

PR #1281, opened 2020-08-05
simplify copyProperties using reflection

supcik:
Here is my proposal for the implementation of `copyProperties` using `reflection`. According to @aykevl:

> if it is a non-zero-length string replace it, if it is a slice append it. There is one exception: `Emulator`.

To avoid a hard-coded exception, I implemented it using a new _tag_ (`copy:"replace"`) in the struct.

The `panic` are here for _defensive programming_ reasons only and should never be called. This implementation is less _safe_ than the original one because the type checks occur at run time and not at compile time, but the code is shorter and more conform with the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).


compileopts/target.go
```diff
@@ -53,91 +54,35 @@ type TargetSpec struct {
...
+	for i := 0; i < specType.NumField(); i++ {
+		field := specType.Field(i)
+		switch kind := field.Type.Kind(); kind {
+		case reflect.String: // for strings, just copy the field of spec2 to spec (if not empty)
+			if v := spec2Value.Field(i); v.Len() > 0 {
+				specValue.Field(i).Set(v)
+			}
+		case reflect.Slice: // for slices...
+			tag, ok := field.Tag.Lookup("copy")
+			if ok && tag == "replace" {
```
niaow: I think this might be clearer as a conditional and a switch statement?

supcik: Yes, you are right... but I would need to duplicate the line

```go
specValue.Field(i).Set(reflect.AppendSlice(specValue.Field(i), spec2Value.Field(i)))
```

for the "append" case and if no tag is given.

On the other hand, I never had to use the explicit "append" tag because this is currently the default

supcik: Maybe I'm being too strict with my DRY principle ;-)

supcik: Another option could be to use the tag `copy:""`(without value) and if this is present, then the field will be replaced instead of appened.

supcik: The code would then be : 

```go
case reflect.Slice: // for slices...
	if _, ok := field.Tag.Lookup("copy"); ok {
		// ... copy the field of spec2 to spec if the field has a "copy" tag and is not []
		if v := spec2Value.Field(i); v.Len() > 0 {
			specValue.Field(i).Set(v)
		}
	} else {
		// ... else append the field of spec2 to spec
		specValue.Field(i).Set(reflect.AppendSlice(specValue.Field(i), spec2Value.Field(i)))
	}
```

supcik (commented):
Here is a new proposal using a `switch`instead of a `if`. I also renamed the function from `copyProperties` to `overrideProperties` because this function is used to build the spec using _inheritance_ so I think that `override` is more appropriate.

deadprogram:
Hello @supcik could you please look into rebasing against the `dev` branch and resolve the merge conflict? Thank you very much.

supcik:
Thank you @deadprogram, for your reply.

I rebased against the `dev` branch and completed the generic inheritance assignments for **pointers** (override if not `nil`) and for **uint64** (override if not zero)

I also added some unit tests for the override function.

deadprogram:
Seems like the `make wasi-libc` task of the `test-llvm9-go111` build is the only one failing. That seems very odd?

supcik:
Indeed. And the error is strange too:
```
/bin/sh: 1: clang-10: not found
```

I don't see how this is related to the change that I made.

supcik:
... And I am also confused : why does the `llvm9` test try to use `clang-10` ?

supcik:
It could be a cache problem. During this run, the "Restoring Cache" step was a miss:

```
No cache is found for key: wasi-libc-sysroot-systemclang-v2
```

Perhaps a new run would "solve" this issue.

supcik:
Perhaps the auto-detection of `llvm` does not work (https://github.com/tinygo-org/tinygo/blob/release/Makefile#L11)

```
# Try to autodetect LLVM build tools.
ifneq (, $(shell command -v llvm-build/bin/clang 2> /dev/null))
    CLANG ?= $(abspath llvm-build/bin/clang)
else
    CLANG ?= clang-10
endif
```

aykevl:
I have seen this before, it's probably fixed by doing the build again (which I've just triggered).

compileopts/target.go
```diff
@@ -56,100 +57,62 @@ type TargetSpec struct {
...
+				case "append":
+					// or append the field of child to spec
+					appendField(specValue.Field(i), childValue.Field(i))
+				default:
+					panic("override mode must be 'copy' or 'append' (default). I don't know how to '" + tag + "'.")
+				}
+			} else {
+				// if no tag, then append the field of child to spec
+				appendField(specValue.Field(i), childValue.Field(i))
+			}
```
aykevl: This can be simplified (but also see the comment above):

```suggestion
			switch tag := field.Tag.Get("override"); tag {
			case "copy":
				// copy the field of child to spec if not empty
				copyFieldIfNotEmpty(specValue.Field(i), childValue.Field(i))
			case "append", "":
				// or append the field of child to spec
				appendField(specValue.Field(i), childValue.Field(i))
			default:
				panic("override mode must be 'copy' or 'append' (default). I don't know how to '" + tag + "'.")
			}
```

aykevl (commented):
I think this is a good idea, but perhaps the DRY principle is taken a bit far with all these locally defined functions (`copyFieldIfNotEmpty` etc). I suspect the code will be about as long and will be easier to understand if these functions are inlined. For example: all logic in the `reflect.Slice` case can be ignored if the slice has a length of zero.

While less lines of code are often beneficial, IMHO the real benefit is making code easier to understand and modify. Therefore, it shouldn't introduce too much indirection.

supcik:
> I think this is a good idea, but perhaps the DRY principle is taken a bit far with all these locally defined functions (`copyFieldIfNotEmpty` etc). I suspect the code will be about as long and will be easier to understand if these functions are inlined. For example: all logic in the `reflect.Slice` case can be ignored if the slice has a length of zero.
> 
> While less lines of code are often beneficial, IMHO the real benefit is making code easier to understand and modify. Therefore, it shouldn't introduce too much indirection.

You are right... I just committed a _simplified_ version.

supcik:
Thank you @deadprogram and @aykevl for your constructive comments.

supcik:
@deadprogram, @aykevl, is there anything I shall do to move this PR forward ?

aykevl (approved):
Looks good to me, thank you!

PR #1289, opened 2020-08-11
esp: add support for the Espressif ESP32 chip

aykevl:
This PR adds very minimal ESP32 support. You can use it as follows:

```sh
tinygo build -o test.elf -target=esp32 -size=short examples/serial
esptool.py --chip=esp32 elf2image test.elf
esptool.py --chip=esp32 --port /dev/ttyUSB0 write_flash 0x1000 test.bin -ff 80m -fm dout

picocom --baud 115200 /dev/ttyUSB0  # optional
```

Support for other important things (GPIO control, `tinygo flash`, etc) can be added in later PRs. I've left them out to keep this as small and reviewable as possible, especially considering the LLVM branch switch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,16 +44,28 @@ commands:
           command: |
             wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
             sudo apt install ./google-chrome-stable_current_amd64.deb
+  install-xtensa-toolchain:
+    parameters:
+      variant:
+        type: string
+    steps:
+      - run:
+          name: "Install Xtensa toolchain"
+          command: |
+            curl -L https://github.com/espressif/crosstool-NG/releases/download/esp-2020r2/xtensa-esp32-elf-gcc8_2_0-esp-2020r2-<<parameters.variant>>.tar.gz -o xtensa-esp32-elf-gcc8_2_0-esp-2020r2-<<parameters.variant>>.tar.gz
+            sudo tar -C /usr/local -xf xtensa-esp32-elf-gcc8_2_0-esp-2020r2-<<parameters.variant>>.tar.gz
+            sudo ln -s /usr/local/xtensa-esp32-elf/bin/xtensa-esp32-elf-ld /usr/local/bin/xtensa-esp32-elf-ld
+            rm xtensa-esp32-elf-gcc8_2_0-esp-2020r2-<<parameters.variant>>.tar.gz
   llvm-source-linux:
     steps:
       - restore_cache:
           keys:
-            - llvm-source-10-v0
+            - llvm-source-10-v1
       - run:
           name: "Fetch LLVM source"
           command: make llvm-source
       - save_cache:
-          key: llvm-source-10-v0
+          key: llvm-source-10-v1
           paths:
             - llvm-project
   build-wasi-libc:
@@ -95,7 +107,7 @@ commands:
             - lib/wasi-libc/sysroot
       - run: go test -v -tags=llvm<<parameters.llvm>> ./cgo ./compileopts ./interp ./transform .
       - run: make gen-device -j4
-      - run: make smoketest
+      - run: make smoketest XTENSA=0
       - run: make wasmtest
       - save_cache:
           key: go-cache-v2-{{ checksum "go.mod" }}-{{ .Environment.CIRCLE_BUILD_NUM }}
@@ -121,6 +133,8 @@ commands:
                 gcc-avr \
                 avr-libc
       - install-node
+      - install-xtensa-toolchain:
+          variant: "linux-amd64"
       - restore_cache:
           keys:
             - go-cache-v2-{{ checksum "go.mod" }}-{{ .Environment.CIRCLE_PREVIOUS_BUILD_NUM }}
@@ -128,7 +142,7 @@ commands:
       - llvm-source-linux
       - restore_cache:
           keys:
-            - llvm-build-10-linux-v0-assert
+            - llvm-build-10-linux-v1-assert
       - run:
           name: "Build LLVM"
           command: |
@@ -146,7 +160,7 @@ commands:
               make ASSERT=1 llvm-build
             fi
       - save_cache:
-          key: llvm-build-10-linux-v0-assert
+          key: llvm-build-10-linux-v1-assert
           paths:
             llvm-build
       - run: make ASSERT=1
@@ -179,6 +193,8 @@ commands:
                 gcc-avr \
                 avr-libc
       - install-node
+      - install-xtensa-toolchain:
+          variant: "linux-amd64"
       - restore_cache:
           keys:
             - go-cache-v2-{{ checksum "go.mod" }}-{{ .Environment.CIRCLE_PREVIOUS_BUILD_NUM }}
@@ -186,7 +202,7 @@ commands:
       - llvm-source-linux
       - restore_cache:
           keys:
-            - llvm-build-10-linux-v0
+            - llvm-build-10-linux-v1
       - run:
           name: "Build LLVM"
           command: |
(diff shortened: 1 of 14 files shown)


aykevl:
This is now ready for review.

.circleci/config.yml
```diff
@@ -44,16 +44,38 @@ commands:
           command: |
             wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
             sudo apt install ./google-chrome-stable_current_amd64.deb
+  install-xtensa-toolchain:
+    parameters:
+      variant:
+        type: string
+    steps:
+      # Cache the file because the Espressif download website is not particularly fast.
```
igrr: Another option is to download the same file from GitHub Releases, which should be pretty fast: https://github.com/espressif/crosstool-NG/releases/tag/esp-2020r2

targets/esp32.ld
```diff
@@ -0,0 +1,95 @@
...
+{
+    /* Data RAM. Allows byte access.
+     * There are various data RAM regions:
+     *   SRAM2: 0x3FFA_E000..0x3FFD_FFFF (72 + 128 = 200K)
+     *   SRAM1: 0x3FFE_0000..0x3FFF_FFFF (128K)
+     * This gives us 328K of contiguous RAM, which is the largest span possible.
+     * SRAM1 has other addresses as well but the datasheet seems to indicate
+     * these are aliases.
+     */
+    DRAM  (rw) : ORIGIN = 0x3FFAE000, LENGTH = 200K + 128K /* Internal SRAM 1 + 2 */
```
igrr, line 13: One thing to watch out for is that the startup stack is somewhere in that range. Might be useful to add an assert that the .data/.rodata sections don't overlap the stack used by the bootloader.

aykevl: That's what gave me trouble initially. I solved it by setting the stack pointer at startup, see src/device/esp/esp32.S. That also makes the system more flexible, as it is no longer bound by the stack that was set by the boot ROM.

.circleci/config.yml
aykevl: Good idea, thanks!

aykevl (commented):
@igrr thanks a lot for checking this PR!

.circleci/config.yml
aykevl: I've updated the PR with this. That makes the CI file smaller and simpler (no caches needed).

targets/esp32.ld
igrr: My point is that the ROM bootloader is using certain region as stack, and if the app's .data section happens to overlap the ROM bootloader stack, the bootloader will overwrite its own stack, while trying to load app's .data.

src/device/esp/esp32.S
```diff
@@ -0,0 +1,14 @@
+// Only calling it call_start_cpu0 for consistency with ESP-IDF.
+.section .text.call_start_cpu0
+1:
+    .long _stack_top
+.global call_start_cpu0
+call_start_cpu0:
+    l32r a1, 1b
```
igrr: In Xtensa window ABI changing a1 like it is done here is not sufficient, i think. If any of the previous register windows contain registers not spilled onto the stack, these registers will get saved to their original frame locations, once the window overflow happens. This might not be a practical issue, unless the boot ROM stack overlaps with app's .bss and the window overflow happens *after* .bss is zeroed out. If that happens, some part of .bss will contain spilled registers.

To avoid this issue, disable WOE in PS register, set WINDOWBASE to 1 << WINDOWSTART, and re-enable WOE in PS.  Here is the code snippet which does this: https://github.com/espressif/esp-idf/blob/c77c4ccf6c43ab09fd89e7c907bf5cf2a3499e3b/components/xtensa/include/xt_instr_macros.h#L26

targets/esp32.ld
aykevl: Right. I see. Thank you for explaining. In that case, I think there are two options:
* Use the stack that is used by the ROM bootloader.
* Let the application load `.data` instead of letting the ROM bootloader do that.

Is the bootloader stack documented somewhere?

igrr: PRO CPU stack is at 0x3ffe1320 — 0x3ffe3f20, and APP CPU stack is at 0x3ffe5230 — 0x3ffe7e30
(I think these values are defined somewhere in IDF, but I can't find the link right now. They can be extracted from the ROM code ELF file, provided at https://dl.espressif.com/dl/esp32_rom.elf.)

In addition to that there are .bss and .data sections used by various ROM functions, at 
0x3ffe0000 — 0x3ffe0440, 0x3ffe3f20 — 0x3ffe4350, 0x3ffae000 — 0x3ffae6e0. If you do not plan to call ROM functions (other than stateless functions like memcpy and memset), and do not plan to call any libraries which in turn call ROM functions, you can ignore these other reserved regions.

On the ESP32-S2, the situation is a bit more sane, as the reserved regions are near the end of DRAM, and don't make "holes" in the continuous region you would otherwise use for the application.

aykevl: Thanks, that is some great insight. Also thanks for the ROM file which will be useful. I will take a look and improve this PR accordingly.
It may be possible to use the given area for the heap anyway and block out the parts that are used by the ROM code. That shouldn't need any changes to the TinyGo heap. Although I'll probably ignore that for this PR and do that only when we actually want to call these special ROM functions.

> On the ESP32-S2, the situation is a bit more sane, as the reserved regions are near the end of DRAM, and don't make "holes" in the continuous region you would otherwise use for the application.

Glad to hear! Unfortunately the ESP32 will be around for a while and we'll need to adjust to it somehow.

aykevl:
@igrr pointed out some flaws in the PR so I will mark this a draft again until I've fixed the issues.

targets/esp32.ld
igrr: > It may be possible to use the given area for the heap anyway and block out the parts that are used by the ROM code.

Right, that's what we do in IDF — these ROM stack regions are getting added to the heap allocator once the startup is complete. And the regions used for ROM .data/.bss are marked as reserved.

> Unfortunately the ESP32 will be around for a while and we'll need to adjust to it somehow.

Of course, it was more of me saying that we realize that reserved regions in the middle of DRAM are tricky to work with, and in the subsequent chips the situation is somewhat improved.

aykevl:
I think I've fixed all the issues pointed out by @igrr now. I'm still using a custom stack, but hopefully did the stack switch correctly this time. I've also added an assert to make sure the `.data` section won't overlap with the startup stack (although that should not normally happen in normal usage, with around 200kB for globals).

deadprogram:
I think the only hesitation I still have is depending on the Xtensa fork of LLVM for everything, instead of using the official LLVM repo. Perhaps we should fork the Xtensa fork ourselves, and then use that dependency, just in case something else changes again. Not sure on that, perhaps I am just being over-careful?

I will find some of my ESP32 boards for some device level testing.

aykevl:
Yes, good point. I have forked llvm-project into tinygo-org and pushed the branch and updated this PR accordingly, so that we have our own copy of LLVM (just to be on the safe side).

deadprogram:
Took a long time to build the LLVM. Also worth mentioning somewhere is you need to install the `xtensa-esp32-elf-ld` linker.

The program built the first time, and flashed/worked exactly as expected on my ESP32 board!

Seems like at this point we can merge this. We still have a lot of work to do to make it useful, but this is a massive step forward, and a long time in the making.

deadprogram (approved):

aykevl:
> Also worth mentioning somewhere is you need to install the `xtensa-esp32-elf-ld` linker.

Ah yes, forgot to mention that. Maybe I'll make a PR to Espressif one day to implement it in `ld.lld`, until then we need this dependency. However, note that you can in fact use the ESP8266 toolchain too (`xtensa-lx106-elf-ld`), but it seemed less confusing to me to just use the ESP32 version.

deadprogram:
OK, merging. Incredible work @aykevl this is very exciting moment.

aykevl:
Awesome! :tada:

andreisfr:
Great job @aykevl !

PR #1299, opened 2020-08-18
main: embed git-hash in tinygo-dev executable

sago35:
When reading issues and slack (tinygo channel), I'm finding it increasingly difficult to get the exact version of tinygo executable.
I think we should embed git-hash in it as well as go.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1001,11 +1001,11 @@ func main() {
 		if s, err := goenv.GorootVersionString(goenv.Get("GOROOT")); err == nil {
 			goversion = s
 		}
-		if !strings.HasSuffix(goenv.Version, "-dev") {
-			fmt.Printf("tinygo version %s %s/%s (using go version %s and LLVM version %s)\n", goenv.Version, runtime.GOOS, runtime.GOARCH, goversion, llvm.Version)
-		} else {
-			fmt.Printf("tinygo version %s-%s %s/%s (using go version %s and LLVM version %s)\n", goenv.Version, gitSha1, runtime.GOOS, runtime.GOARCH, goversion, llvm.Version)
+		version := goenv.Version
+		if strings.HasSuffix(goenv.Version, "-dev") && gitSha1 != "" {
+			version += "-" + gitSha1
 		}
+		fmt.Printf("tinygo version %s %s/%s (using go version %s and LLVM version %s)\n", version, runtime.GOOS, runtime.GOARCH, goversion, llvm.Version)
 	case "env":
 		if flag.NArg() == 0 {
 			// Show all environment variables.
--- a/main.go
+++ b/main.go
@@ -28,6 +28,10 @@ import (
 	"go.bug.st/serial"
 )
 
+var (
+	gitSha1 string
+)
+
 // commandError is an error type to wrap os/exec.Command errors. This provides
 // some more information regarding what went wrong while running a command.
 type commandError struct {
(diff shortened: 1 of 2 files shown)


sago35:
before:
tinygo version 0.15.0-dev linux/amd64 (using go version go1.14.4 and LLVM version 10.0.1)

after:
tinygo version 0.15.0-dev-ddd5d24 linux/amd64 (using go version go1.14.4 and LLVM version 10.0.1)

deadprogram:
@sago35 maybe the easiest thing is to start a new PR for this by using a new branch of `dev` and then `git cherry-pick ddd5d24`

main.go
```diff
@@ -28,6 +28,10 @@ import (
 	"go.bug.st/serial"
 )
 
+var (
+	gitSha1 string
```
aykevl, line 34: Can you add a comment here explaining that this variable is set to the git SHA1 hash with a build flag?

main.go
```diff
@@ -995,7 +999,11 @@ func main() {
 		if s, err := goenv.GorootVersionString(goenv.Get("GOROOT")); err == nil {
 			goversion = s
 		}
-		fmt.Printf("tinygo version %s %s/%s (using go version %s and LLVM version %s)\n", goenv.Version, runtime.GOOS, runtime.GOARCH, goversion, llvm.Version)
+		if !strings.HasSuffix(goenv.Version, "-dev") {
+			fmt.Printf("tinygo version %s %s/%s (using go version %s and LLVM version %s)\n", goenv.Version, runtime.GOOS, runtime.GOARCH, goversion, llvm.Version)
+		} else {
+			fmt.Printf("tinygo version %s-%s %s/%s (using go version %s and LLVM version %s)\n", goenv.Version, gitSha1, runtime.GOOS, runtime.GOARCH, goversion, llvm.Version)
+		}
```
aykevl: You could keep a single `Printf` line here, reducing duplication:

```suggestion
		version := goenv.Version
		if !strings.HasSuffix(goenv.Version, "-dev") && gitSha1 != "" {
			version += "-" + gitSha1
		}
		fmt.Printf("tinygo version %s %s/%s (using go version %s and LLVM version %s)\n", version, runtime.GOOS, runtime.GOARCH, goversion, llvm.Version)
```

Also, you can build TinyGo without going through the Makefile so there needs to be a check that `gitSha1` is set.

aykevl (commented):
This is a good idea, thanks for looking into it!

aykevl:
From the smoke tests:
```
tinygo version 0.15.0-dev-bd9fdd0 windows/amd64 (using go version go1.15 and LLVM version 10.0.0)
tinygo version 0.15.0-dev-b088340 linux/amd64 (using go version go1.14.7 and LLVM version 10.0.1)
tinygo version 0.15.0-dev-b088340 darwin/amd64 (using go version go1.14 and LLVM version 10.0.1)
```

The Windows git hash seems to be different as it works from a custom PR commit: https://github.com/tinygo-org/tinygo/commit/bd9fdd0a1e6cf92577ecc402e55b77f81444bc58, I think it's fine.

main.go
sago35: Added a comment.
I've added a link to stackoverflow, but it may not be necessary.

main.go
sago35: The less duplication, the better.

sago35:
```
tinygo version 0.15.0-dev-1bc61d7 windows/amd64 (using go version go1.15 and LLVM version 10.0.0)
tinygo version 0.15.0-dev-b8e87c0 linux/amd64 (using go version go1.14.7 and LLVM version 10.0.1)
tinygo version 0.15.0-dev-b8e87c0 darwin/amd64 (using go version go1.14 and LLVM version 10.0.1)
```

Hmmm, I'd like the windows version to be the same gitSha1 as the linux and macos versions.

https://dev.azure.com/tinygo/tinygo/_build/results?buildId=1995&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=9c33736e-cef6-544f-3943-acf16442661a&l=508

aykevl:
I'm pretty sure this is only for pull request builds. For normal release builds (and local builds), the real git has should be used.

sago35:
link: #1300

PR #377, opened 2019-05-25
wasm: fix wasm-ld hang

aykevl:
See the following bugs for more information:
https://bugs.llvm.org/show_bug.cgi?id=41508
https://bugs.llvm.org/show_bug.cgi?id=37064

This should be removed again when we switch to LLVM 9, it may have a very small (basically negligible) compile time cost.

PR #1308, opened 2020-08-23
Remove --no-threads from wasm-ld calls.

QuLogic:
The bugs linked from the original addition of this flag are fixed:
https://bugs.llvm.org/show_bug.cgi?id=41508
or 'possibly fixed':
https://bugs.llvm.org/show_bug.cgi?id=37064#c8
since LLVM 8.0.1. TinyGo only support LLVM 9+, and as noted in the
original PR (#377), this can be removed when switching to 9.

Additionally, this flag was dropped from LLVM 11.

### DIFF
--- a/targets/wasm.json
+++ b/targets/wasm.json
@@ -12,7 +12,6 @@
 	],
 	"ldflags": [
 		"--allow-undefined",
-		"--no-threads",
 		"--stack-first",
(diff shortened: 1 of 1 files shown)


aykevl:
Thank you for investigating and fixing! This flag is indeed not necessary any more.

PR #1309, opened 2020-08-24
wasm: update wasi-libc dependency

aykevl:
Several updates are necessary for LLVM 11 support, so simply update to the latest commit.

Thanks to @QuLogic for investigating.


deadprogram:
Updates! Now merging.

Issue #1278, opened 2020-08-03
Flashing on an adafruit nRF52 board causes an error message to appear


sago35:
When I try to do tinygo flash with feather-nrf52840, I get an error message.
tinygo flash seems to succeed.

The reset appears to be running immediately after the following.
https://github.com/tinygo-org/tinygo/blob/dev/main.go#L82

It's the same on both ubuntu and windows.Is this the same for everyone?
We might want to include a note of caution in your documentation.

```
# ubuntu 20.04
$ tinygo flash --target feather-nrf52840 --size short ./src/examples/blinky1
   code    data     bss |   flash     ram
   6256      12    4064 |    6268    4076
error: failed to flash /tmp/tinygo749323199/main.uf2: rename /media/sago35/FTHR840BOOT/flash.uf2.tmp /media/sago35/FTHR840BOOT/flash.uf2: no such file or directory
```

```
# windows10 pro
$ .\bin\tinygo.exe flash --target feather-nrf52840 --size short .\src\examples\blinky1
   code    data     bss |   flash     ram
   6856      12    4064 |    6868    4076
error: failed to flash C:\tinygo-pr\tinygo0.14.0dev_1afe528\temp\tinygo458391831\main.uf2: rename D:\/flash.uf2.tmp D:\/flash.uf2: A device which does not exist was specified.
```

sago35:
> Benjamin Gould  19:57
I'm able to reproduce the same problem using `circuitplay-bluefruit` and `clue_alpha` targets as well

sago35:
https://github.com/adafruit/Adafruit_nRF52_Bootloader

sago35:
I don't know if this is relevant or not.
https://github.com/hathach/tinyusb/issues/375
https://github.com/adafruit/Adafruit_nRF52_Bootloader/issues/120

bgould:
FWIW this is working ok on my Debian laptop, but reproducible using the same board on my Windows computer as well as an Ubuntu VM running in VirtualBox on that same Windows computer.  These are the bootloader details for that board:
```
UF2 Bootloader 0.2.6 lib/nrfx (v1.1.0-1-g096e770) lib/tinyusb (legacy-525-ga1c59649) s140 6.1.1                 
Model: Adafruit Feather nRF52840 Express                                        
Board-ID: NRF52-Bluefruit-v0                                                    
Bootloader: s140 6.1.1                                                          
Date: Dec 21 2018
```

sago35:
[...]

PR #1313, opened 2020-08-24
main: use simpler file copy instead of file renaming when flashing UF2 MSD

deadprogram:
This PR use simpler file copy instead of file renaming to avoid issues on nrf52840 UF2 bootloaders when flashing MSD drives. It corrects #1278

### DIFF
--- a/main.go
+++ b/main.go
@@ -58,33 +58,23 @@ func moveFile(src, dst string) error {
 	return os.Remove(src)
 }
 
-// copyFile copies the given file from src to dst. It copies first to a .tmp
-// file which is then moved over a possibly already existing file at the
-// destination.
+// copyFile copies the given file from src to dst. It can copy over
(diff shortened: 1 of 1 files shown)


deadprogram:
Flashes without error on both ItsyBitsy nRF52840, and also Circuit Playground Bluefruit.

sago35:
Flashes without error on feather-nrf52840

aykevl:
Seems good to me. The reason for the temporary file was to get somewhat atomic behavior for copying files, which may be important when the output is stored on a regular host system. However, in this case the `copyFile` function is only called during a release build (where atomicity doesn't really matter) and duing MSD flashing, so it's fine to change the behavior.

sago35:
I checked to be sure, and the following was fine.

* xiao
* wioterminal
* feather-m4

PR #1316, opened 2020-08-26
loader: use ioutil.TempDir to create a temporary directory

aykevl:
... instead of generating one with math/rand. The problem was that
math/rand is deterministic across runs, resulting in a possible race
when trying to create the same directory between two processes.
Additionally, because I used `os.MkdirAll`, no error was reported when
the directory already existed. The solution to this is to use the stdlib
function designed for this: ioutil.TempDir.

### DIFF
--- a/loader/goroot.go
+++ b/loader/goroot.go
@@ -10,13 +10,11 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"math/rand"
 	"os"
 	"os/exec"
 	"path"
 	"path/filepath"
 	"runtime"
-	"strconv"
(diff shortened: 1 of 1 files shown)


deadprogram:
The CircleCI tests all are failing for same reason:

```
main.go:763: error: stat /home/circleci/.cache/tinygo: no such file or directory
```

aykevl:
I see. That's an actual bug, will fix...

aykevl:
Fixed, this should work now.

deadprogram:
Thanks @aykevl now merging.

PR #1322, opened 2020-08-26
feather-stm32f405: add support for Adafruit STM32F405 Feather Express

ardnew:
## Notes

The exact chip on this board is `STM32F405RG`. 

A similar chip (`STM32F407`) already has support, required for target board `stm32f4disco`. Instead of duplicating largely-identical implementations, the following files were renamed with changes made to support _**both**_ targets:

|Deleted (`STM32F407`-only)|Created (`STM32F405`+`STM32F407`)|
|:--------|:--------|
|`src/machine/machine_stm32f407.go`|`src/machine/machine_stm32f40x.go`|
|`src/runtime/runtime_stm32f407.go`|`src/runtime/runtime_stm32f40x.go`|
|`src/device/stm32/stm32f407xx-altfunc-bitfields.go`|`src/device/stm32/stm32f40x-altfunc-bitfields.go`|

Note that because the `STM32F407` implementation already includes basic support for UART and SPI, the same support was automatically inherited by the `STM32F405`. No effort was made to *remove* UART/SPI support for `feather-stm32f405` from this common `STM32F405/STM32F407` implementation. However, the multiple interfaces available for both of these peripherals — (6)×UART, (3)×SPI — on this board have **not** been implemented (their declarations have been included in `src/machine/board_feather-stm32f405.go` for future reference, but are commented out). Only the (1)×UART and (1)×SPI primary interfaces have been implemented and enabled.

I2C is *not* currently implemented for `stm32F4disco`, and was excluded via build tag in `src/machine/i2c.go`. Similarly, a build tag exclusion was added for `feather_stm32f405` to the same file.

#### Programming

Flashing this board using its DFU bootloader is a bit awkward/cumbersome (see Adafruit docs linked below), but it is possible — without requiring an external programmer — by pulling the `B0` pin high (you can use the board's 3.3V output pin) at bootup when connected to host PC via USB. This puts the device in bootloader mode.

Once in bootloader mode, I've verified the device can be programmed using either `STM32CubeProgrammer` (requires ***Oracle*** JDK 8) or the open-source tool `dfu-util` (available in most package manager repos, verified using Ubuntu 20.04).

> Alternatively, I've been using a J-Link Mini EDU SWD programmer connected to the header on the reverse side of the board without issue.

|Programmer|URL|
|:-------------:|:----|
|`STM32CubeProgrammer`|https://www.st.com/en/development-tools/stm32cubeprog.html|
|`dfu-util`|http://dfu-util.sourceforge.net/|

#### ST
|Title|URL|
|:----:|:----|
|Overview|https://www.st.com/content/st_com/en/products/microcontrollers-microprocessors/stm32-32-bit-arm-cortex-mcus/stm32-high-performance-mcus/stm32f4-series/stm32f405-415/stm32f405rg.html|
|Datasheet|https://www.st.com/resource/en/datasheet/stm32f405rg.pdf|
|UserManual|https://www.st.com/resource/en/reference_manual/dm00031020-stm32f405415-stm32f407417-stm32f427437-and-stm32f429439-advanced-armbased-32bit-mcus-stmicroelectronics.pdf|

#### Adafruit

|Title|URL|
|:----:|:----|
|Overview|https://learn.adafruit.com/adafruit-stm32f405-feather-express/overview|
|Bootloader|https://learn.adafruit.com/adafruit-stm32f405-feather-express/dfu-bootloader-details|
|Schematic|https://learn.adafruit.com/adafruit-stm32f405-feather-express/downloads|

#### STM32duino 

A lot of info was derived from these relevant chip/board-specific implementations in [Arduino_Core_STM32](https://github.com/stm32duino/Arduino_Core_STM32)

|Title|URL|
|:----:|:----|
|Variant|https://github.com/stm32duino/Arduino_Core_STM32/tree/master/variants/FEATHER_F405|
|HAL|https://github.com/stm32duino/Arduino_Core_STM32/tree/master/system/STM32F4xx|


sago35:
> > Also, there are some duplicate definitions:
> > These should be in the machine_stm32f40x.go file (or alternatively in a new board_stm32f40x.go file).
> 
> Thanks. I moved those defs from board_feather-stm32f405.go to machine_stm32f40x.go, and apparently forgot to remove them from board_stm32f4disco.go.
> Wonder why make test didn't catch this? Or, how did you catch it?

The stm32f405 build works fine, but the stm32f4disco build now fails.
Normally you can detect this with smoke-test, so it's a good idea to check before git-push.

```
tinygo build -size short -o test.hex -target=stm32f4disco        examples/blinky1
# machine
/home/circleci/lib/tinygo/src/machine/machine_stm32f40x.go:15:2: PA0 redeclared in this block
/home/circleci/lib/tinygo/src/machine/board_stm32f4disco.go:11:2: 	other declaration of PA0
/home/circleci/lib/tinygo/src/machine/machine_stm32f40x.go:16:2: PA1 redeclared in this block
/home/circleci/lib/tinygo/src/machine/board_stm32f4disco.go:12:2: 	other declaration of PA1
/home/circleci/lib/tinygo/src/machine/machine_stm32f40x.go:17:2: PA2 redeclared in this block
/home/circleci/lib/tinygo/src/machine/board_stm32f4disco.go:13:2: 	other declaration of PA2
/home/circleci/lib/tinygo/src/machine/machine_stm32f40x.go:18:2: PA3 redeclared in this block
/home/circleci/lib/tinygo/src/machine/board_stm32f4disco.go:14:2: 	other declaration of PA3
```

https://app.circleci.com/pipelines/github/tinygo-org/tinygo/3477/workflows/c977b6c1-5c12-4334-affc-ef5777b96a61/jobs/17141

src/device/stm32/stm32f40x-altfunc-bitfields.go
```diff
@@ -1,7 +1,7 @@
-// These are the supported alternate function numberings on the stm32f407
-// +build stm32,stm32f407
+// +build stm32
+// +build stm32f405 stm32f407
 
-// Alternate function settings on the stm32f4xx series
+// Alternate function numberings on the stm32f40x series
```
sago35, line 4: This change should be separated into a separate PR.
The same goes for i2c.

If this is specific to stm32f4 you can add a buildtag like stm32f4.

sago35: Before this PR, you might want to change your stm32f4disco configuration.
I'm not sure of the correct name, but the sources common to stm32f4disco and stm32f405 need to be cut out in a buildtag, for example stm32f4.

Then adding stm32f405 would work, I think.

ardnew: ah, `make smoketest` is what I needed to be running, not `make test`. thanks. 

I simply removed the pin definitions from `board_stm32f4disco.go`, since they can be shared by all boards with this chip, and now `make smoketest` passes as expected. 

are you suggesting i still make a new build tag anyway? there are several files that use that same 2-line build tag change I made like you are seeing there in `stm32f40x-altfunc-bitfields.go`. I should instead replace all instances of those 2 lines with the new build tag and update their target configurations?

sago35: For this PR, it's better to change only the files that are needed for stm32f405.
It's easier for us to review.

> are you suggesting i still make a new build tag anyway?

yes.
An example of the change is as follows
https://github.com/sago35/tinygo/commit/1f74ac8089b629d23473fcc8986918a8da8c4112

Again, for this PR it's best to only make changes that pertain to stm32f405, so changes to the following files should be in a separate PR so they can be merged in first.

* src/machine/board_stm32.go
* src/machine/i2c.go
* targets/stm32f4disco.json
* src/device/stm32/stm32f407xx-altfunc-bitfields.go -> stm32f4-altfunc-bitfields.go
* src/machine/board_stm32f4disco.go
* src/machine/machine_stm32f407.go
* src/runtime/runtime_stm32f407.go

ardnew: thank you very much for the detailed feedback. hopefully I'm following your intent and will split this into the following new PRs:

- new build tag `stm32f4` - updates build constraints in existing board `stm32f4disco` and machine/runtime `stm32f407`
- new build tag `feather-stm32f405` - creates new target, utilizing new tag `stm32f4`, and implements board `feather-stm32f405` and machine/runtime `stm32f405`
- integrate UART support
- integrate SPI support
- integrate I2C support

please let me know if any of these are still too large.

(good thing I made this a draft!)

aykevl: I think it's fine to bundle everything in one PR for basic support (that way you can test the PR actually works!), but it would indeed be helpful to split out new peripheral support in separate PRs.

src/machine/machine_stm32f40x.go
```diff
@@ -0,0 +1,324 @@
...
+	PE13 = portE + 13
+	PE14 = portE + 14
+	PE15 = portE + 15
+
+	PH0 = portH + 0
+	PH1 = portH + 1
+)
+
+func CPUFrequency() uint32 {
+	return 168_000_000
```
aykevl, line 105: This is not supported in older Go versions, which breaks the tests.
```suggestion
	return 168000000
```
or
```suggestion
	return 168e6
```

PR #1315, opened 2020-08-25
remove thread-safe wrapper

sago35:
This PR is for confirmation purposes only.

Although the following is written in the source code, it may be possible to run it parrallel in practice.

> // Due to some problems with LLD, we cannot run links in parallel, or in parallel with compiles.
> // Therefore, we put a lock around builds and run everything else in parallel.
https://github.com/tinygo-org/tinygo/blob/release/main_test.go#L118

In my hand, the test has finished without any failure.

before : 47 sec
after : 8 sec

PR #1325, opened 2020-08-28
loader: be more robust when creating the cached GOROOT

aykevl:
This commit fixes two issues:

  * Do not try to create the cached GOROOT multiple times in parallel.
    This may happen in tests and is a waste of resources (and thus
    speed).
  * Check for an "access denied" error when trying to rename a directory
    over an existing directory. On *nix systems, this results in the
    expected "file exists" error. Unfortunately, Windows gives an access
    denied. This commit fixes the Windows behavior.

Both fixes should fix the Windows issue in #1315.

### DIFF
--- a/loader/goroot.go
+++ b/loader/goroot.go
@@ -15,10 +15,13 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"sync"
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Very subtle, good catch @aykevl now merging.

PR #1329, opened 2020-08-29
nrf: add SoftDevice support for the Circuit Playground Bluefruit

aykevl:
Untested, therefore this is a draft.

### DIFF
--- a/src/machine/board_circuitplay_bluefruit.go
+++ b/src/machine/board_circuitplay_bluefruit.go
@@ -2,7 +2,7 @@
 
 package machine
 
-const HasLowFrequencyCrystal = true
+const HasLowFrequencyCrystal = false
 
 // GPIO Pins
 const (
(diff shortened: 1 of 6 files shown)


bgould:
I have a couple of these boards so I'm happy to test it out

aykevl:
Tested and works (after fixing some issues). This is now ready for review.

deadprogram:
Worked as expected, we probably should make the same change for all the Adafruit nRF52840 based boards, since they also should support this.

aykevl:
I added support for the other three boards, of which I was only able to test the ItsyBits nRF52840. It appears that the Bluefruit is the exception here with a bare chip and all other boards use an existing BLE module (and appear to have a low-frequency crystal).

deadprogram:
Thanks, @aykevl very awesome! Now merging.

PR #1331, opened 2020-08-30
cortexm: fix stack size calculation with interrupts

aykevl:
Interrupts store 32 bytes on the current stack, which may be a goroutine
stack. After that the interrupt switches to the main stack pointer so
nothing more is pushed to the current stack. However, these 32 bytes
were not included in the stack size calculation.

This commit adds those 32 bytes. The code is rather verbose, but that is
intentional to make sure it is readable. This is tricky code that's hard
to get right, so I'd rather keep it well documented.


deadprogram:
Thanks for making it very clear. Now merging.

PR #1334, opened 2020-08-30
Fix arch release job

johanbrandhorst:
Instead of using su, which is blocking, set the
user explicitly for each command.


deadprogram:
You would know better than I would if this will work @johanbrandhorst so now merging so we can find out next release. :smile_cat:

PR #1335, opened 2020-08-30
flash: call PortReset only on other than openocd

sago35:
Before this fix, PortReset == true is judged and portReset is executed first.
If your microcontroller is crashing, portReset will fail.

 We don't need the portReset when we write from openocd.


deadprogram:
That makes sense to me, thanks @sago35 merging.

PR #1338, opened 2020-08-31
esp32: add support for basic GPIO

aykevl:
GPIO is much more advanced on the ESP32, but this is a starting point. It gets examples/blinky1 to work.

I chose to support the ESP32-WROOM-32 instead of an actual board as this is the module that is used on the vast majority of ESP32 boards. I also don't have a "standard" ESP32 board, just some random dev board from AliExpress that is really a breakout board with USB-to-serial chip for programming.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -347,7 +347,7 @@ ifneq ($(AVR), 0)
 	@$(MD5SUM) test.hex
 endif
 ifneq ($(XTENSA), 0)
-	$(TINYGO) build -size short -o test.bin -target=esp32               examples/serial
+	$(TINYGO) build -size short -o test.bin -target=esp32-wroom-32      examples/blinky1
 endif
 	$(TINYGO) build -size short -o test.hex -target=hifive1b            examples/blinky1
 	@$(MD5SUM) test.hex
--- a/src/machine/board_esp32-wroom-32.go
+++ b/src/machine/board_esp32-wroom-32.go
@@ -1,0 +1,6 @@
+// +build esp32_wroom_32
+
+package machine
+
+// Blue LED on the ESP32-WROOM-32 module.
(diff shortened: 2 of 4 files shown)


sago35:
It worked well.
My test environment is as follows.

https://www.switch-science.com/catalog/3210/

Makefile
```diff
@@ -347,7 +347,7 @@ ifneq ($(AVR), 0)
 	@$(MD5SUM) test.hex
 endif
 ifneq ($(XTENSA), 0)
-	$(TINYGO) build -size short -o test.bin -target=esp32               examples/serial
+	$(TINYGO) build -size short -o test.bin -target=esp32-wroom-32      examples/blinky1
```
sago35, line 350: esp-wroom-32 might be a better choice.

aykevl: It appears that esp-wroom-32 is the old name.
I can't find esp-wroom-32 on their product page: https://www.espressif.com/en/products/modules/esp32
Wikipedia claims (without citation) that it is the old name:

> First publicly available ESP32 module board created by Espressif.[11] FCC Part 15.247 tested (FCC ID: 2AC7Z-ESPWROOM32).[12] Based on ESP32-D0WDQ6 chip. Originally named "ESP-WROOM-32".

sago35: You're right.

deadprogram:
I have an "original" board from about 4 years ago labeled `ESP-WROOM-32` and then several brand new boards labeled "ESP32-WROOM-32`. Both are able to be flashed with this code.

Merging, fantastic work @aykevl what's next? What should we help with? :smile_cat:

aykevl:
> What should we help with? smile_cat

The usual stuff: I2C, SPI, ...

PR #857, opened 2020-01-20
arduino: make avrdude command line compatible with Windows

aykevl:
On Windows, it is common that there is a colon in the path. `avrdude` will treat that as a separator and everything behind it as the file format specifier instead of defaulting to Intel hex format.

By explicitly specifying the Intel hex format (with `:i`), this issue should be fixed.

fixes #851 (hopefully)

PR #1341, opened 2020-09-02
arduino-mega2560: fix flashing on Windows

aykevl:
Without the extra `:i` at the end, avrdude will misinterpret the colon in Windows paths. **Untested**.

Also see https://github.com/tinygo-org/tinygo/pull/857.


deadprogram:

PR #1345, opened 2020-09-03
Makefile: check whether submodules have been downloaded in some common cases

aykevl:
This should help new people making this very common mistake.


deadprogram:
A good check to perform. Merging, thanks @aykevl

PR #1346, opened 2020-09-03
all: run test binaries in the correct directory

aykevl:
Test binaries must be run in the source directory of the package to be tested. This wasn't done, leading to a few "file not found" errors.

This commit implements this. Unfortunately, it does not allow more packages to be tested as both affected packages (debug/macho and debug/plan9obj) will still fail with this patch even though the "file not found" errors are gone.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -24,22 +24,35 @@ import (
 	"tinygo.org/x/go-llvm"
 )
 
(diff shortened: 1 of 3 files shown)


deadprogram:
That makes sense to me. Merging.

PR #1347, opened 2020-09-03
ci: run `tinygo test` for known-working packages

aykevl:
These packages are known to pass tests with `tinygo test`. It's still a very short list, but hopefully this list can be expanded to eventually cover most or all of the standard library.


aykevl:
Removed `testing/iotest` from the list of tested packages, which appears to not have *_test.go files in older Go versions.

deadprogram:
I :heart: tests! Merging.

PR #1348, opened 2020-09-03
test: support non-host tests 

aykevl:
For example, for running tests with `-target=wasm` or `-target=cortex-m-qemu`. It looks at the output to determine whether tests were successful in the absence of a status code.

~~Depends on #1346 because they would otherwise conflict.~~

### DIFF
--- a/main.go
+++ b/main.go
@@ -127,22 +127,50 @@ func Test(pkgName string, options *compileopts.Options) error {
 	}
 
 	return builder.Build(pkgName, ".elf", config, func(result builder.BuildResult) error {
-		cmd := exec.Command(result.Binary)
-		cmd.Stdout = os.Stdout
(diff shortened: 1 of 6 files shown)


aykevl:
This is now ready for review.

deadprogram:
How do I use this @aykevl?  It applies to the `tinygo test` command, but is there a specific example I can try out?

aykevl:
You can run the three tests supported at this state, for example:

    tinygo test -target=cortex-m-qemu container/list
    tinygo test -target=cortex-m-qemu container/ring
    tinygo test -target=cortex-m-qemu text/scanner

(math and unicode/utf8 were added after this PR branched so cannot be tested here)

deadprogram:
Thanks @aykevl that works as expected. Now merging.

PR #1353, opened 2020-09-04
esp32: add libgcc ROM functions to linker script

aykevl:
They are copied from ESP-IDF. They are always available (burned in the mask ROM) so best to use them.

I do not think a list of functions and addresses can be copyrighted (they don't seem to be creative in any way, unlike the implementations themselves which can be done in a creative/novel way). And if they are, they're licensed under an Apache 2.0 license so if there is a problem it's easy to add that.
@deadprogram any opinions on this matter?

These functions are also available in compiler-rt and picolibc, but sadly the Xtensa backend has bugs (https://github.com/espressif/llvm-project/issues/33 and https://github.com/espressif/llvm-project/issues/34 respectively) so they can't be compiled yet.

### DIFF
--- a/targets/esp32.ld
+++ b/targets/esp32.ld
@@ -97,8 +97,106 @@ _heap_end = ORIGIN(DRAM) + LENGTH(DRAM);
 
 _stack_size = 4K;
 
-/* From ESP-IDF, included here as long as picolibc doesn't compile.
+/* From ESP-IDF:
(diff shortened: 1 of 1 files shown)


deadprogram:
Seems like a necessary duplication at least for now. If they get those bug fixed, it would be better to use picolibc probably. In any case, now merging.

PR #1354, opened 2020-09-04
esp32: export machine.PortMask* for bitbanging implementations

aykevl:
This is useful for some drivers, in particular for the WS2812 driver.


deadprogram:
:traffic_light: now merging, thanks @aykevl

PR #1356, opened 2020-09-05
ci: set git-fetch-depth to 1

sago35:
I think this PR will make windows CI a little faster.

> steps:
> - checkout: self | none | repository name # self represents the repo where the initial Pipelines YAML file was found
>   fetchDepth: number  # the depth of commits to ask Git to fetch; defaults to no limit

https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema%2Cparameter-schema#checkout

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -14,6 +14,7 @@ jobs:
       inputs:
         version: '1.15'
     - checkout: self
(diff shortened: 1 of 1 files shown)


deadprogram:
Good idea @sago35 now merging.

PR #1361, opened 2020-09-06
List usable targets with a targets subcommand

aykevl:
This is something that should be useful for autocomplete and similar tools.

### DIFF
--- a/main.go
+++ b/main.go
@@ -8,6 +8,7 @@ import (
 	"go/scanner"
 	"go/types"
 	"io"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"os/signal"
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -921,6 +922,35 @@ func main() {
...
+			if err != nil {
+				fmt.Fprintln(os.Stderr, "could not list target:", err)
+				os.Exit(1)
+				return
+			}
+			if spec.FlashMethod == "" && spec.FlashCommand == "" && spec.Emulator == nil {
+				// This doesn't look like a regular target file, but rather like
+				// a parent target (such as targets/cortex-m.json).
+				continue
+			}
```
sago35, line 949: The nintendoswitch.json is not listed, so either change the conditions a bit or put a trick in the json.

My only concern is that it may be difficult to keep this rule in place in the future. 
If I have trouble in the future, I will consider it.

deadprogram: I have same issue, which is that the `nintendoswitch` target does not appear in the list.

aykevl:
Yes, I'm aware of the Nintendo Switch limitation:

> This patch adds the `tinygo targets` command, which lists usable targets (targets that can be used in the `-target` flag).
> 
> The assumption here is that usable targets can either be flashed or emulated by TinyGo. There is one exception where it doesn't work yet: the nintendoswitch target. Right now it requires some manual steps to build a .nro file which can then be run by yuzu, hence why it doesn't show up in the list.

I looked into adding emulator support, but it's hard to test with the issue I hit here: https://github.com/tinygo-org/tinygo/pull/1226#issuecomment-688322890

deadprogram:
I propose we merge this PR now, and then once the 2 nintendoswitch related PRs are merged, another PR can be created that adds it to the list.

aykevl:
Yes, that was what I was trying to suggest.

deadprogram:
OK, now merging. Let the awesome tooling commence!

PR #1362, opened 2020-09-06
main: add cached GOROOT to info subcommand

aykevl:
This is necessary for an upcoming VS Code extension to support TinyGo, and may be useful for other people wanting to use proper autocomplete etc. in their IDE.


deadprogram:
Thanks @aykevl now merging.

sago35:
Gopls worked correctly on windows 10.

PR #1368, opened 2020-09-08
nintendoswitch: Fix invalid memory read / write in print calls

racerxdl:
These fixes are for what it looks like a compiler bug. Me and @deadprogram tried for a while to debug what caused that problems, and it looks like it is some unknown compiler bug which it doesn't seens to happen for other platforms.

There are two bugs:

1. Depending on how the compiler optimizes the `putchar`, the slicing of a pre-allocated slice triggers a new memory allocation, which in some parts of the code isn't possible.
2. The panic call in `runtime_nintendoswitch_heap.go` if only there, works just fine. But if the target code has another panic, neither of the panic's output is printed correctly (the string passed to the print call has a null pointer)

I can't be sure what's causing these errors, but this fixes it for now. I haven't got any issues in target code if that runtime code is like this.

@aykevl with this, the issues you were having in https://github.com/tinygo-org/tinygo/pull/1226 should be gone.

### DIFF
--- a/src/runtime/runtime_nintendoswitch.go
+++ b/src/runtime/runtime_nintendoswitch.go
@@ -40,11 +40,11 @@ func ticks() timeUnit {
 	return timeUnit(ticksToNanoseconds(timeUnit(getArmSystemTick())))
 }
 
-var stdoutBuffer = make([]byte, 120, 120)
+var stdoutBuffer = make([]byte, 120)
 var position = 0
 
 func putchar(c byte) {
-	if c == '\n' || position >= 119 {
+	if c == '\n' || position > len(stdoutBuffer) {
 		nxOutputString(&stdoutBuffer[0], uint64(position))
 		position = 0
 		return
--- a/src/runtime/runtime_nintendoswitch.go
+++ b/src/runtime/runtime_nintendoswitch.go
@@ -40,21 +40,18 @@ func ticks() timeUnit {
 	return timeUnit(ticksToNanoseconds(timeUnit(getArmSystemTick())))
 }
 
-var stdoutBuffer = make([]byte, 0, 120)
+var stdoutBuffer = make([]byte, 120, 120)
+var position = 0
 
 func putchar(c byte) {
-	if c == '\n' || len(stdoutBuffer)+1 >= 120 {
-		NxOutputString(string(stdoutBuffer))
-		stdoutBuffer = stdoutBuffer[:0]
+	if c == '\n' || position >= 119 {
+		nxOutputString(&stdoutBuffer[0], uint64(position))
+		position = 0
 		return
 	}
 
-	stdoutBuffer = append(stdoutBuffer, c)
-}
-
-func usleep(usec uint) int {
-	sleepThread(uint64(usec) * 1000)
-	return 0
+	stdoutBuffer[position] = c
+	position++
 }
 
 func abort() {
--- a/src/runtime/runtime_nintendoswitch_heap.go
+++ b/src/runtime/runtime_nintendoswitch_heap.go
@@ -4,17 +4,12 @@
 
 package runtime
 
-import "unsafe"
-
 const heapSize = 0x2000000 * 16 // Default by libnx
-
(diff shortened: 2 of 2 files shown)


aykevl:
Well that's interesting, and something I'd like to investigate further.

racerxdl:
The panic thing is very easy to simulate, just make an code that panics anything. With the nxOutputDebug string in `src/runtime/runtime_nintendoswitch_heap.go`, the panic prints out correctly in Yuzu debug output. If you use `panic` in `src/runtime/runtime_nintendoswitch_heap.go` it gives an Invalid Memory Read for each character in the string.

src/runtime/runtime_nintendoswitch_heap.go
```diff
@@ -24,7 +19,12 @@ func preinit() {
 	setHeapSize(&heapStart, heapSize)
 
 	if heapStart == 0 {
-		panic("failed to allocate heap")
+		// Panic should work, but if we call panic here, it triggers a compiler bug that breaks
+		// all print calls in the whole application. So far we couldn't figure out why, that's
+		// why we're using a byte array and nxOutputString directly
+		nxOutputString(&failedToAllocateHeapString[0], uint64(len(failedToAllocateHeapString)))
+		exit(1)
+		//panic("failed to allocate heap")
```
aykevl: This appears unrelated, when I revert this change it still works. However, it would be a good idea to use a `runtimePanic` instead which doesn't allocate (assuming that `putchar` doesn't allocate):

```suggestion
		runtimePanic("failed to allocate heap")
```

src/runtime/runtime_nintendoswitch.go
```diff
@@ -40,21 +40,18 @@ func ticks() timeUnit {
...
 
-var stdoutBuffer = make([]byte, 0, 120)
+var stdoutBuffer = make([]byte, 120, 120)
+var position = 0
 
 func putchar(c byte) {
-	if c == '\n' || len(stdoutBuffer)+1 >= 120 {
-		NxOutputString(string(stdoutBuffer))
-		stdoutBuffer = stdoutBuffer[:0]
+	if c == '\n' || position >= 119 {
```
aykevl: Nit:

```suggestion
	if c == '\n' || position > len(stdoutBuffer) {
```

src/runtime/runtime_nintendoswitch.go
```diff
@@ -40,21 +40,18 @@ func ticks() timeUnit {
 	return timeUnit(ticksToNanoseconds(timeUnit(getArmSystemTick())))
 }
 
-var stdoutBuffer = make([]byte, 0, 120)
+var stdoutBuffer = make([]byte, 120, 120)
```
aykevl: Very minor nitpick:

```suggestion
var stdoutBuffer = make([]byte, 120)
```

aykevl (commented):
I have confirmed that the changes in `putchar` do indeed fix the issue. Looking at the IR, this change makes sure `putchar` is not considered a potentially blocking function anymore, which is good. I haven't looked into why this change fixes that, but it might be because it proves to the compiler the function doesn't allocate (the compiler probably isn't smart enough to determine that the slice append could not allocate).

The `panic` below appears unrelated so I think it would be best to revert it (or change to `runtimePanic`) for code readability.

aykevl:
I managed to reproduce the incorrect behavior in the old code (with a plain panic in preinit). This issue is fixed by using `runtimePanic` instead.

src/runtime/runtime_nintendoswitch_heap.go
racerxdl: Great! Any major differences between runtimePanic vs panic?

aykevl: Yes, `runtimePanic` is somewhat lighter weight and should generally not do heap allocations (which is important if you're trying to panic when memory is full, for example). The `panic` function might call an interface method that allocates, so the compiler can't prove it won't allocate.

aykevl:
Thank you!

PR #1371, opened 2020-09-09
nrf52840: use higher priority for USB-CDC code

aykevl:
This ensures that stdout (`println` etc) keeps working in interrupts.

Generally you shouldn't print anything in an interrupt. However, printing things for debugging is very useful and printing panic messages can be critical when the code doesn't work for some reason.

This is a partial fix for https://github.com/tinygo-org/bluetooth/issues/15. It doesn't fix the underlying issue but for me I can see the panic message with this fix (while the device would just hang without this fix).


deadprogram:

PR #1372, opened 2020-09-09
esp32: configure the I/O matrix for GPIO pins

aykevl:
Only some pins (notably including GPIO2 aka machine.LED) have GPIO for the default function 1. Other pins (such as GPIO 15) had a different function by default. Function 3 means GPIO for all the pins, so always use that when configuring a pin to use as a GPIO pin.

In the future, the mux configuration will need to be updated for other functions such as SPI, I2C, etc.

### DIFF
--- a/src/machine/machine_esp32.go
+++ b/src/machine/machine_esp32.go
@@ -20,18 +20,43 @@ type PinMode uint8
 const (
 	PinOutput PinMode = iota
 	PinInput
+	PinInputPullup
+	PinInputPulldown
 )
 
 // Configure this pin with the given configuration.
 func (p Pin) Configure(config PinConfig) {
-	if config.Mode == PinOutput {
+	var muxConfig uint32 // The mux configuration.
+
+	// Configure this pin as a GPIO pin.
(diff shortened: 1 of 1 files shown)


src/machine/machine_esp32.go
```diff
@@ -20,18 +20,43 @@ type PinMode uint8
...
 	PinOutput PinMode = iota
 	PinInput
+	PinInputPullup
+	PinInputPulldown
 )
 
 // Configure this pin with the given configuration.
 func (p Pin) Configure(config PinConfig) {
-	if config.Mode == PinOutput {
+	var muxConfig uint32 // The mux configuration.
```
deadprogram, line 29: Perhaps all this should be broken out into a separate function, since as you mention a variation of setting each of these will be required for I2C.

aykevl: I'll try implementing SPI to see whether this abstraction works.

aykevl: I implemented SPI and it requires only a small modification to this function.

deadprogram:
OK, in that case merging. Thanks @aykevl

PR #1373, opened 2020-09-12
support WASI target

mathetake:
This PR adds `wasi` target

Screenshot from 2020-09-23 10-28-29

## Changes in this PR
- [x] add `targets/wasi.json`
- [x] implement `runtime.ticks`
- [x] implement `runtime.sleepTicks`
- [x] add the emulator/test

## Future TODOs 
- [ ] add WASI systemcall tests in `tests/wasi/*`
- [ ] documentation
- [ ] support file system related syscalls

I will create an issue for tracking these TODOs after merging this PR
___

## Motivation

we in the [Envoyproxy](https://github.com/envoyproxy/envoy) community are working on [proxy-wasm](https://github.com/proxy-wasm/spec) project. The objective here is to provide web developers with the way to write Envoy extensions through WASM <-> Proxy ABI with their own favorite languages.

The host environment is not Go ABI (as in wasm_exec.js) at all and is kind of the extension of WASI, and therefore we would like to produce WASI compatible wasm binaries with TinyGo. We can compile proxy-wasm ABI compatible WASM binaries with the current TinyGo but we cannot import almost all of existing libraries due to the syscall/js centric syscall implementation. That is too bad for developer experience with proxy-wasm extensions.

## Consideration

I know there are ongoing WASI issues https://github.com/golang/go/issues/31105 and https://github.com/golang/go/issues/38248 in the official compiler, but I think it's going to take multiple months (even years) for them to be completed. Once they're completed, I will do the integration with TinyGo for consistency and remove TinyGo specific wasi syscall implementation.

___ 

~This is not completed yet but I would like to have opinions and direction from TinyGo maintainers before I work on the detail.~ Now ready for review :smile: 

P.S. Thanks for this awesome project! Without TinyGo, there would be no potential for Go developers to write proxy-wasm extensions.


mathetake:
@deadprogram Could you please take a look at? If there's no concern, I am going to work on the detail.

src/runtime/runtime_wasi.go
```diff
@@ -0,0 +1,86 @@
+// +build wasm,wasi
+
+package runtime
+
+import "unsafe"
+
+type timeUnit float64
```
aykevl: This file is almost a duplicate of src/runtime/runtime_wasm.go. Can you separate out the bits that are different in js and in wasi in separate files? Something like:

runtime_wasm.go: contains all the code shared for WebAssembly targets (the vast majority)
runtime_wasi.go: contains just the WASI-specific bits
runtime_js.go: contains just the JS-specific bits

src/syscall/syscall_wasi.go
```diff
@@ -0,0 +1,123 @@
+// +build wasi
```
aykevl: This file is identical to syscall_baremetal.go (except for this build tag), so it should be possible to merge the two.

aykevl (commented):
Originally I wanted to wait until upstream Go supports WASI. Unfortunately, that doesn't seem to be happening (at least not in the near future) so it does make sense to add support in TinyGo independent of upstream support.

This PR contains a lot of code duplication. While some duplication might be necessary, it would be best to avoid as much as possible. Also, I'm missing a targets/wasi.json to be able to use this.
(Yes I know that this PR is a draft and needs more work, just wanted to point out a few things that will be needed for review).

mathetake:
time.Now has become available

Screenshot from 2020-09-17 21-47-18

mathetake:
runtime.sleepTicks is now working with poll_oneoff API:)

Screenshot from 2020-09-19 11-09-28

mathetake:
@deadprogram @aykevl I've made this PR ready for review so PTAL when you have time! :smile:

transform/gc.go
```diff
@@ -19,6 +19,9 @@ func MakeGCStackSlots(mod llvm.Module) bool {
 		}
 		stackChainStart := mod.NamedGlobal("runtime.stackChainStart")
 		if !stackChainStart.IsNil() {
+			if shouldHiveGlobals {
+				stackChainStart.SetVisibility(llvm.HiddenVisibility)
```
aykevl: Instead of making this symbol hidden, would it work to set the linkage to internal? That might allow some more optimizations.

transform/gc.go
```diff
@@ -8,7 +8,7 @@ import (
 
 // MakeGCStackSlots converts all calls to runtime.trackPointer to explicit
 // stores to stack slots that are scannable by the GC.
-func MakeGCStackSlots(mod llvm.Module) bool {
+func MakeGCStackSlots(mod llvm.Module, shouldHiveGlobals bool) bool {
```
aykevl: There is a typo here.

```suggestion
func MakeGCStackSlots(mod llvm.Module, shouldHideGlobals bool) bool {
```

However, would it work to always disable this global? If it works for WASI, it will likely work for JS too. The global was never meant to be exposed. Making it universally hidden or internal avoids a lot of complexity and simplifies this PR.
```

File: targets/wasi.json [hunk 5]
```diff
@@ -0,0 +1,22 @@
+{
+	"llvm-target":   "wasm32--wasi",
+	"build-tags":    ["js", "wasm", "wasi"],
```
aykevl: I would _really_ like to avoid the complexity around the JS/WebAssembly compatibility. So instead of adding the `js` and `wasm` build tags, can you only include the `wasi` build tag instead? It would mean the target is treated more like a baremetal target instead of a WebAssembly target at the source code level.

The main reason is the added complexity around the loader. There is already the issue with overriding the syscall or not overriding it, and I would like to keep those exceptions to an absolute minimum.

For this to work, you may need to add the `wasi` build tag to a few source files in addition to the `baremetal` build tag.

For one example of the complexities of trying to be compatible with `js`/`wasm`:

```go
// TODO: add crypto/rand
```

aykevl (commented):
While this is a good start, the main problem is that WASI support here is trying to also include JS/WebAssembly build tags. This won't work, as this is not supported by the standard library. Therefore, there are two options:

  * Match `js/wasm` support exactly, including support for the syscall/js package. I do not think this is a good path to take, as the things that work for JS don't work so well for WASI.
  * Pretend to be something else. This is the path that baremetal support in TinyGo takes: it pretends to be GOOS=linux GOARCH=arm even on other architectures such as AVR and RISC-V. While this may seem odd, it means that we can use the existing standard library with very few patches (mainly the syscall package).

mathetake:
>  it pretends to be GOOS=linux GOARCH=arm

OK, that makes sense. I'll follow this path and refactor the code :+1:

mathetake:
@aykevl Thanks you for kind review. I followed your comments and made the following changes:

- use `GOARCH=arch` and `GOOS=linux` instead of wasm/js
- set `InternalLinkage` instead of `HiddenVisibility`, and enabled this by default

Indeed, these changes reduced the complexity of this PR! Please take another look :smile: Thanks!

aykevl (approved):
Looks good to me, thank you!
While the GOOS=linux GOARCH=arm is unfortunate, I really don't see a better way to integrate this in a sane way. So let's hope upstream Go eventually adds wasi support natively that we can rely on.

mathetake:
give me a sec to resolve conflict:)

loader/loader.go
```diff
@@ -169,6 +169,7 @@ func Load(config *compileopts.Config, inputPkgs []string, clangHeaders string, t
 			}
 			return nil, err
 		}
+
```
aykevl: Can you remove this empty line? It is causing a conflict.

I have changed this PR to use the dev branch, see: https://github.com/tinygo-org/tinygo/blob/release/CONTRIBUTING.md#how-to-use-our-github-repository

mathetake: sorry I branched off from release branch and that is why this conflict happens. Thank you for the pointer, and I will keep that in mind

PR #1376, opened 2020-09-13
feather-stm32f405: add UART support

ardnew:
This adds support for the three UART peripherals on the target `feather-stm32f405`.

### DIFF
--- a/src/machine/board_feather-stm32f405.go
+++ b/src/machine/board_feather-stm32f405.go
@@ -2,6 +2,11 @@
 
 package machine
 
+import (
+	"device/stm32"
+	"runtime/interrupt"
+)
+
 const (
 	NUM_DIGITAL_IO_PINS = 39
 	NUM_ANALOG_IO_PINS  = 7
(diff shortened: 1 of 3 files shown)


src/machine/board_feather-stm32f405.go
```diff
@@ -113,7 +118,30 @@ const (
...
+		Bus:             stm32.USART1,
+		AltFuncSelector: stm32.AF7_USART1_2_3,
+	}
+	UART0 = UART1
+)
+
+func initUART() {
+	UART1.Interrupt = interrupt.New(stm32.IRQ_USART3, UART1.handleInterrupt)
+	UART2.Interrupt = interrupt.New(stm32.IRQ_USART6, UART2.handleInterrupt)
+	UART3.Interrupt = interrupt.New(stm32.IRQ_USART1, UART3.handleInterrupt)
```
ardnew, line 143: Should we be automatically installing these interrupts for the non-default UARTs (`UART2`/`UART3`)? The default `UART1` is the only one required by other packages (e.g. `runtime.putchar()`)

The other UARTs will function with this left as-is, but it's more of a question of convention and how other targets provide extra peripherals to the user

deadprogram: > Should we be automatically installing these interrupts for the non-default UARTs 

Possibly not? If a developer wants to use them for a specific application, they can do that themselves.

aykevl: Probably yes. This is a bit of magic code that gets optimized away if the UART is not in use (so no interrupt should be registered if you don't use UART2 for example).

aykevl:
The code itself looks fine.

deadprogram:
Thanks for working on this contribution @ardnew now merging.

PR #1377, opened 2020-09-13
feather-stm32f405: add SPI support

ardnew:
This adds support for the three SPI peripherals on the target `feather-stm32f405`.

~~This included extending the STM32 family's SPI implementation to now support the following options:~~

~~- Direction (full/half duplex)~~
~~- Hardware or software chip/slave-select~~
~~- 8-bit or 16-bit data frame format~~

### DIFF
--- a/src/machine/machine_stm32_spi.go
+++ b/src/machine/machine_stm32_spi.go
@@ -44,37 +44,6 @@ type SPIConfig struct {
 	CSMode    uint8
 }
 
-//	const (
-//		modeSlave  = 0
-//		modeMaster = stm32.SPI_CR1_MSTR | stm32.SPI_CR1_SSI
-//
-//		direction2Lines   = 0
-//		direction2LinesRx = stm32.SPI_CR1_RXONLY
-//		direction1Line    = stm32.SPI_CR1_BIDIMODE
-//
-//		dataSize8Bit  = 0
-//		dataSize16Bit = stm32.SPI_CR1_DFF
-//
-//		clkPolarityLow  = 0
-//		clkPolarityHigh = stm32.SPI_CR1_CPOL
-//
-//		clkPhase1Edge = 0
-//		clkPhase2Edge = stm32.SPI_CR1_CPHA
-//
-//		slaveSelectSoft       = stm32.SPI_CR1_SSM
-//		slaveSelectHardInput  = 0
-//		slaveSelectHardOutput = stm32.SPI_CR2_SSOE << 16
-//
-//		firstBitMSB = 0
-//		firstBitLSB = stm32.SPI_CR1_LSBFIRST
-//
-//		tiModeDisable = 0
-//		tiModeEnable  = stm32.SPI_CR2_FRF
-//
-//		crcDisable = 0
-//		crcEnable  = stm32.SPI_CR1_CRCEN
-//	)
-
 // Configure is intended to setup the STM32 SPI1 interface.
 func (spi SPI) Configure(config SPIConfig) {
 
@@ -234,11 +203,10 @@ func (spi SPI) Transfer(w byte) (byte, error) {
 	for spi.Bus.SR.HasBits(stm32.SPI_SR_BSY) {
 	}
 
-	// clear the overrun flag.
-	// note: this assumes 2-lines directionality (the default and only supported
-	//       direction at the moment).
-	//       otherwise, it is not appropriate to clear this flag.
-	_ = spi.Bus.SR.Get()
+	// clear the overrun flag (only in full-duplex mode)
+	if !spi.Bus.CR1.HasBits(stm32.SPI_CR1_RXONLY | stm32.SPI_CR1_BIDIMODE | stm32.SPI_CR1_BIDIOE) {
+		spi.Bus.SR.Get()
+	}
 
 	// disable the SPI interface
 	spi.Bus.CR1.ClearBits(stm32.SPI_CR1_SPE)
--- a/src/machine/machine_stm32_spi.go
+++ b/src/machine/machine_stm32_spi.go
@@ -9,28 +9,6 @@ import (
 	"unsafe"
 )
 
-// SPI directionality
-const (
-	DirFullDuplex   = 0
-	DirSimplexRx    = 1
-	DirHalfDuplexRx = 2
-	DirHalfDuplexTx = 3
-)
-
-// SPI data frame format/size
-const (
-	DataWidth8Bit  = 0
-	DataWidth16Bit = 1
-)
-
-// SPI chip/slave select mode. Use any GPIO pin in software mode, otherwise use
-// one of the pins with NSS alternate function in hardware mode.
-const (
-	CSSoft    = 0
-	CSHardIn  = 1
-	CSHardOut = 2
-)
-
 // SPIConfig is used to store config info for SPI.
 type SPIConfig struct {
 	Frequency uint32
@@ -39,9 +17,6 @@ type SPIConfig struct {
 	SDI       Pin
 	LSBFirst  bool
 	Mode      uint8
-	Direction uint8
-	DataSize  uint8
-	CSMode    uint8
 }
 
 // Configure is intended to setup the STM32 SPI1 interface.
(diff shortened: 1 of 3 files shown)


src/machine/machine_stm32f405.go
```diff
@@ -32,8 +33,45 @@ type SPI struct {
...
+	// subtract 1 from the log2 value, keeping a lower bound of 0
+	if div < 0 {
+		div = 0
+	} else if div > 0 {
+		div--
+	}
+
+	// finally, shift the enumerated value into position for SPI CR1
+	return uint32(div) << stm32.SPI_CR1_BR_Pos
+}
```
ardnew: @aykevl this is the function you previously expressed concern about, since other machine packages sometimes use a switch statement and hardcoded frequencies.

an argument supporting your concern is that this code is used by exactly one chip (STM32F405), so the fact that it is a _more generalized solution_ isn't actually benefiting anything (but in principle could be re-used elsewhere)

wanted to point this out so it isn't overlooked in review, in case the consensus would rather use the other approach.

sago35: ~~The value is about 6.25Mhz with the following settings.~~
~~Is this as intended?~~

```go
	machine.SPI0.Configure(machine.SPIConfig{
		Frequency: 21000000,
	})
```

The value of CR1 register seems to be 0x344 at this time.

ss

src/machine/machine_stm32f405.go
```diff
@@ -32,8 +33,45 @@ type SPI struct {
...
+	if freq > clock {
+		freq = clock
+	}
+
+	// calculate the exact clock divisor (freq=clock/div -> div=clock/freq).
+	// truncation is fine, since it produces a less-than-or-equal divisor, and
+	// thus a greater-than-or-equal frequency.
+	// divisors only come in consecutive powers of 2, so we can use log2 (or,
+	// equivalently, bits.Len - 1) to convert to respective enum value.
+	div := bits.Len32(clock/freq) - 1
```
sago35, line 78: It seems to raise an exception when freq == 0, but I don't think it's an exception when I run it.
I think it would be better to set some kind of initial value.

src/machine/machine_stm32f405.go
sago35: My instrument was set up incorrectly.
At 21,000,000, it appears to be set to 41MHz.
Compared to samd51, it appears to be about twice as fast.

ss

deadprogram:
Also, @ardnew please note merge conflict that will need resolution via rebase against `dev` branch. Thanks.

src/machine/machine_stm32f405.go
ardnew: Hmmm, the SPI0 interface is connected to hardware bus `stm32.SPI2` on `APB1`, which has a **maximum** clock freq of 42 MHz, and the smallest divisor is DIV2, so i don't think its even possible for that SPI interface to have a freq faster than 21 MHz. Can you double check this?

src/machine/machine_stm32f405.go
ardnew: Yes, a minimum freq check should be added corresponding to the maximum check immediately before this statement

src/machine/machine_stm32f405.go
aykevl: My main concern here was the use of floating points and advanced math operations. While I am still not super happy about math/bits, it's not too big of a dependency so is okay.
The machine and runtime packages are a bit special when it comes to dependencies: they can't import anything that might (recursively) import the runtime. This has led to problems in the past, so I'd rather keep the list of imported packages to simple leaf packages.
(Also, right now TinyGo is rather slow and doesn't cache compiled packages. So keeping the dependency tree small also helps to reduce compile time. But this should eventually get fixed).

Regarding the 21/42 MHz: I think the 42MHz refers to the time the clock is high and the 21MHz refers to the entire pulse. So yes, the SPI peripheral is running at 21MHz (as expected).

src/machine/machine_stm32_spi.go
```diff
@@ -17,83 +39,178 @@ type SPIConfig struct {
...
+	spi.configurePins(config)
+
 	// Get SPI baud rate based on the bus speed it's attached to
 	var conf uint32 = spi.getBaudRate(config)
 
 	// set bit transfer order
 	if config.LSBFirst {
 		conf |= stm32.SPI_CR1_LSBFIRST
+	} else {
+		conf &^= stm32.SPI_CR1_LSBFIRST
```
aykevl: Why do you clear the bits here and below? `getBaudRate` should have all those bits zeroed.

A few days ago I used a very similar pattern here (and in other places in the PR):
https://github.com/tinygo-org/tinygo/pull/1379/files#diff-cb056e43ce0e9bc6e837cd32b2b9ac66R352-R357

src/machine/machine_stm32_spi.go
```diff
@@ -17,83 +39,178 @@ type SPIConfig struct {
 	SDI       Pin
 	LSBFirst  bool
 	Mode      uint8
+	Direction uint8
+	DataSize  uint8
+	CSMode    uint8
```
aykevl: Can you please explain the semantics of all these?

* What is the impact of the `Direction` flag on the Tx and Transfer functions?
* What impact has `DataSize`, considering that `Transfer` and `Tx` only support byte slices?
* How is the CS pin configured?

While there are other chips with such features, they aren't implemented for TinyGo. And I definitely think such things could be useful, but they must make sense in the provided API. For example, instead of setting a "soft" or "hard" CS pin, you could pass in the pin itself (or `NoPin`).

Right now I think all these features are already more or less supported in TinyGo with software. Half duplex communication is implemented by leaving either the `w` or the `r` parameter of `Tx` empty, the data size is not relevant on the wire (it's all just bits), and CS is easily emulated in software. In other words, I suspect these are performance optimizations instead of supporting something you could not do otherwise.

All in all I think it would be better to leave them unimplemented for now until one (or multiple!) use cases appear and to model an API based on that. The reason is that it is very easy to add an API but very hard to remove or change it later.

src/machine/machine_stm32_spi.go
ardnew: Right, the clearing isn't necessary. At one point I had separate functions for these individual bits in `CR1` (so the user could enable/disable features individually at runtime). Tried to simplify the interface and merged them altogether back into this sole `Configure` function. Will remove the clearing statements. (Are they not optimized away?)

src/machine/machine_stm32f405.go
sago35: The following sources were used to acquire the waveforms of each.
I think stm32 is twice as fast as stm32.
Maybe the correct one is feather-m4.

feather-stm32f405 + Freq 21MHz
image


feather-m4 + Freq 24MHz
image


```go
package main

import (
	"machine"
	"time"
)

func main() {
	spi0 := machine.SPI0
	spi0.Configure(machine.SPIConfig{
		Frequency: 21000000, // for feather-stm32f405
		//Frequency: 24000000, // for feather-m4
	})

	for {
		spi0.Tx([]byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 9}, nil)
		time.Sleep(1000 * time.Millisecond)
	}
}
```

ardnew: Can you test with an equivalent program using Arduino? That would probably be the best reference measure.

I'm thinking the Feather M4 is being driven half the frequency requested, simply because the STM32F4 datasheet indicates that SPI bus is *incapable of operating at a freq greater than 21 MHz*

src/machine/machine_stm32_spi.go
ardnew: Well if you'd rather I just remove them for now until other machines come up with a common API, I can do that -without scouring back through the reference manual to try (and probably fail) to define their behaviors.

The one difference that might affect bus stability is clearing the overrun flag for full-duplex mode (lines 206-209). If we default to this mode, and the user implements a half or simplex mode in software, then we could be inadvertently clearing important error flags.

The motivation for clearing this flag in this mode comes from ST's own SPI driver source code.

src/machine/machine_stm32f405.go
sago35: I understand.
I'm wrong and you're right.

I didn't suspect samd51, but it seems that tinygo's samd51 is wrong.

* feather-stm32f405 : ok : 1 / 0.000000048 = 20.83MHz
* feather-m4 : ng : 1 / 0.000000084 = 11.90MHz

Arduino feather-m4 + Freq 24000000
image

```c

#include <SPI.h>
SPISettings MAX6675Setting (24000000, MSBFIRST, SPI_MODE0);

// the setup function runs once when you press reset or power the board
void setup() {
  // initialize digital pin LED_BUILTIN as an output.
  pinMode(LED_BUILTIN, OUTPUT);
  SPI.begin();
}

// the loop function runs over and over again forever
void loop() {
  SPI.beginTransaction(MAX6675Setting);
  SPI.transfer(0xFF);
  SPI.endTransaction();
  digitalWrite(LED_BUILTIN, HIGH);   // turn the LED on (HIGH is the voltage level)
  delay(500);                       // wait for a second
  digitalWrite(LED_BUILTIN, LOW);    // turn the LED off by making the voltage LOW
  delay(500);                       // wait for a second
}

```

sago35: > Regarding the 21/42 MHz: I think the 42MHz refers to the time the clock is high and the 21MHz refers to the entire pulse. So yes, the SPI peripheral is running at 21MHz (as expected).

I understand.

sago35: The SPI Frequency of atsamd will be corrected by the following PR.

#1387

src/machine/machine_stm32_spi.go
deadprogram: Hi @ardnew I think removing is the correct answer, at lest for now. Thanks!

aykevl: Yes, indeed. The reason is that APIs (especially one as fundamental as the machine package) should have small, well  defined interfaces.

deadprogram:
Seems like the requested changes have all been made to the PR, is that correct?

aykevl (commented):
The code looks clean, however I'm wondering about this:

* The SPI peripheral is enabled at the end of the `Configure` call
* The SPI peripheral is enabled at the start of `Transfer`, if it isn't enabled already
* The SPI peripheral is disabled at the end of `Transfer`.

Is it intentional to disable the SPI peripheral between transferring each byte? Why do you enable it then in `Configure`?

ardnew:
> Is it intentional to disable the SPI peripheral between transferring each byte? 

It was intentional on my part, yes, as this is what the STM32 HAL/LL drivers do before and after each transfer - I'm not sure *why* exactly they do it though. Clears pending flags? Paranoid attempt to minimize bus traffic? Helps prevent the user from re-configuring the interface while its enabled?

> Why do you enable it then in `Configure`?

Good catch, indeed one of these is unnecessary. I think enabling/disabling for each transfer is the safer option, if not the most efficient. But I also think its a little pedantic and would be fine leaving it enabled between transfers unless someone objects.

deadprogram:
I do not think that you want to enable/disable on event call to `Tx()`, only on `Configure()`. It might "work" but will slow down operation considerably.

deadprogram:
Hi @ardnew any update on removing the register enable from the `Tx()` method? Thank you.

ardnew:
@deadprogram Sorry -- got a little distracted with other efforts! 

Requested change implemented.

deadprogram:
Thanks @ardnew now squash/merging.

PR #1379, opened 2020-09-13
esp32: add SPI support

aykevl:
I have tested a version of this with a st7735 display and it actually worked when all the cables were just right (which was really hard). After that I did a few modifications but the signals still look good on my logic analyzer.

Screenshot from 2020-09-13 14-58-27

Although I should say I'm not quite sure whether the mode parameter is set correctly.


deadprogram:
Excellent work @aykevl now merging.

PR #1382, opened 2020-09-14
nintendoswitch: Add dynamic loader for runtime loading PIE sections

racerxdl:
Since Nintendo Switch has mandatory [ASLR](https://en.wikipedia.org/wiki/Address_space_layout_randomization) it is required to build executables as PIE. That means some sections might be re-alocated by the application loader.

This patch adds a dynamic loader function for AArch64 elf that patches the entries with a provided image base. In case of Nintendo Switch that image base is loaded early before getting into preinit to be sure that nothing that requires a dynamic portion is called.

That also adds a maximum of 4095 bytes pad into romExtract function, because since Nintendo Switch OS requires a page-aligned   section, it might have up to 4095 bytes of difference between sections.

I also added a dummy `write` call to allow `fmt.printf` statements to be used. This is a easy way to test the dynamic loading since  the embedded `println` is inlined in the code not requiring any dynamic section.

That makes work both with internal .nro file generation and external generation by a elf file.

With that, a bit more complex program runs without any error in the emulator:

```go
package main

import (
    "fmt"
    "time"
)

func main() {
    println("Hello world!")

    if true {
        fmt.Printf("Number %d - String %s\n", 1234, "Test")
    }

    println("HUEBR\n")
    i := 0
    for i < 10 {
       fmt.Printf("Cycle %d\n", i)
       time.Sleep(time.Second)
       i++
    }
    panic("FINISH")
}
```

image

### DIFF
--- a/targets/nintendoswitch.s
+++ b/targets/nintendoswitch.s
@@ -51,9 +51,8 @@ start:
     mov  x5, x0
     mov  x4, x1
 
-    // Save lr, context pointer, main thread handler
-    adrp x0, _aslr_base
-    str  x6, [x0, #:lo12:_aslr_base]
+    // Save ASLR Base to use later
+    mov x0, x6
 
     // clear .bss
     adrp x5, __bss_start
@@ -70,26 +69,10 @@ bssloop:
 
 run:
     // process .dynamic section
-    adrp x0, _aslr_base
-    ldr  x0, [x0, #:lo12:_aslr_base]
+    // ASLR base on x0
     adrp x1, _DYNAMIC
     add  x1, x1, #:lo12:_DYNAMIC
     bl   __dynamic_loader
 
-    // set LR to svcExitProcess if it's null
-    adrp x3, exit
-    add x3, x3, #:lo12:exit
-    cmp x30, xzr
-    csel x30, x3, x30, eq
-
     // call entrypoint
-    mov x3, sp
-    sub sp, sp, 0x10
-    stp x29, x30, [sp]
     b    main
-
-.section .data.horizon
-.align 8
-.global _aslr_base // Placeholder for ASLR Base Address
-_aslr_base:
-    .dword 0
--- a/targets/nintendoswitch.s
+++ b/targets/nintendoswitch.s
@@ -6,8 +6,7 @@
 _start:
     b start
     .word _mod_header - _start
-    .word 0
-    .word 0
+    .ascii "HOMEBREW"
 
     .ascii "NRO0"              // magic
     .word 0                    // version (always 0)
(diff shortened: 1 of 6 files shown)


targets/nintendoswitch.s
```diff
@@ -6,20 +6,19 @@
...
 
     // segment headers
-    .word __text_start
+    .word 0 // __text_start
     .word __text_size
-    .word __rodata_start
+    .word 0 //__rodata_start
     .word __rodata_size
-    .word __data_start
+    .word 0 //__data_start
```
aykevl, line 21: Why is this necessary? It seems to me that this would cause issues when the .nro file is loaded when all segments start at 0 and overlap the text segment.

targets/nintendoswitch.s
```diff
@@ -45,18 +44,17 @@ _mod_header:
...
+    adrp x6, _start
 
     // context ptr and main thread handle
     mov  x5, x0
     mov  x4, x1
 
+    // Save lr, context pointer, main thread handler
+    adrp x0, _aslr_base
+    str  x6, [x0, #:lo12:_aslr_base]
+
```
aykevl: I don't understand this. Here you are saving the `x6` variable, and you load it again after clearing the .bss section?

I see that bss zeroing overwrites registers `x5` and `x6`, but doesn't touch `x0` (which you're loading the ASLR base into). So why not move it to `x0` instead?

```suggestion
    mov x0, x6
```

targets/nintendoswitch.s
```diff
@@ -71,7 +69,27 @@ bssloop:
...
+    add x3, x3, #:lo12:exit
+    cmp x30, xzr
+    csel x30, x3, x30, eq
+
     // call entrypoint
-    adrp x30, exit
-    add  x30, x30, #:lo12:exit
+    mov x3, sp
+    sub sp, sp, 0x10
+    stp x29, x30, [sp]
```
aykevl: Why do you need to store these two registers on the stack? Are they used by `svcExitProcess`?

It would make more sense to me to make sure `main` never returns and exit it from within the Go `main` function, Go is much easier to understand than assembly.

targets/nintendoswitch.s
racerxdl: Thats necessary for NRO environment (which isn't the case of the emulator that is a NSO environment). For now, I'm not sure how to fill that, since linker complains that can't put a relocatable symbol there :(

targets/nintendoswitch.s
racerxdl: Oh, these are junk from older code. I will remove it.

targets/nintendoswitch.s
racerxdl: Done!

targets/nintendoswitch.s
deadprogram: So with this PR can a TinyGo program still run correctly in the emulator?

racerxdl: Yes, it does work fine. The screenshot is with this PR running on Emulator ^^

deadprogram:
This is really great progress @racerxdl thanks for all your work on this!

Now merging.

PR #1385, opened 2020-09-15
microbit: reelboard: flash using OpenOCD when needed

aykevl:
When using a SoftDevice, the MSD flash method is not appropriate as it
will erase the entire flash area before writing the new firmware. This
also wipes the SoftDevice. Instead, use OpenOCD to only rewrite the
parts of flash that need to be rewritten and leave the SoftDevice alone.


deadprogram:
This will no doubt help a lot of people avoid overwriting their boards. We just need to inform them they need to install openocd in the docs.

Anyhow, thanks @aykevl now merging.

PR #1386, opened 2020-09-15
stacksize: deal with DW_CFA_advance_loc1

aykevl:
In some cases this operation is emitted. It appears to be emitted when a switch is lowered to a jump table in the ARM backend.


ofauchon:
This commit,solves my compilation error (error: could not parse executable for stack size analysis: could not decode .debug_frame bytecode op 0x2 (for address 0x80010d0)).
Thanks

deadprogram:
That is one obscure bug, good find! Now merging.

PR #1387, opened 2020-09-16
atsamd5x, atsamd2x: fix BAUD value

sago35:
In the case of atsamd5x, the BAUD calculation was wrong.
Therefore, with Freq = 24MHz, it runs at 12MHz.

In the case of atsamd2x, BAUD calculations were underflowing when Freq > 24MHz.
If you underflow the BAUD calculation, the frequency will be 94 kHz, but maybe it should work at maximum speed.

samd21 underflow:
image

### DIFF
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -1188,7 +1188,10 @@ func (spi SPI) Configure(config SPIConfig) error {
 	}
 
 	// Set synch speed for SPI
-	baudRate := (CPUFrequency() / (2 * config.Frequency)) - 1
+	baudRate := CPUFrequency() / (2 * config.Frequency)
+	if baudRate > 0 {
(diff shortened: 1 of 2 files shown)


sago35:
testcode:

```go
package main

import (
	"machine"
	"time"
)

func main() {
	spi0 := machine.SPI0
	spi0.Configure(machine.SPIConfig{
		Frequency: 48000000,
		//Frequency: 24000000,
		//Frequency: 12000000,
	})

	for {
		spi0.Tx([]byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 9}, nil)
		time.Sleep(1000 * time.Millisecond)
	}
}
```

deadprogram:
Looking at the datasheets, looks like this was an error from the beginning. Thanks for doing so much work on it, @sago35 now merging.

PR #1388, opened 2020-09-16
bluepill: Enable stm32's USART2 on bluepill boards

ofauchon:
This PR enables the second stm32's UARTs (USART2) on bluepill boards and maps it to UART1 tinygo serial device.


deadprogram:
Thanks for the contribution @ofauchon now merging.

PR #1390, opened 2020-09-17
Let the math package pass all tests

aykevl:
This PR contains two commits, that together get the math package to pass all tests with `tinygo test math`.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -2044,17 +2044,17 @@ func (b *builder) createBinOp(op token.Token, typ, ytyp types.Type, x, y llvm.Va
 			case token.QUO: // /
 				return b.CreateFDiv(x, y, ""), nil
 			case token.EQL: // ==
-				return b.CreateFCmp(llvm.FloatUEQ, x, y, ""), nil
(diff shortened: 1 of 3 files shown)


compiler/compiler.go
```diff
@@ -2578,7 +2578,7 @@ func (b *builder) createUnOp(unop *ssa.UnOp) (llvm.Value, error) {
 			if typ.Info()&types.IsInteger != 0 {
 				return b.CreateSub(llvm.ConstInt(x.Type(), 0, false), x, ""), nil
 			} else if typ.Info()&types.IsFloat != 0 {
-				return b.CreateFSub(llvm.ConstFloat(x.Type(), 0.0), x, ""), nil
+				return b.CreateFNeg(x, ""), nil
```
aykevl, line 2581: Apparently I needed to use a special instruction to get the negative of a floating point number instead of simply doing `0-x`.

aykevl:
Changed and simplified the logic for changing the import path so that (in theory, untested) it also works when the math package is imported as part of a test.

deadprogram:
Good work @aykevl now merging.

PR #1392, opened 2020-09-17
nrf: add I2C error checking

### LINKED ISSUE #1343: nrf: I2C Tx() fails on NACK
The i2c.readByte has no checks for error conditions. This makes the function to loop forever if receiving a NACK instead of an ACK.

dmlambea:
This adds basic support for error checking during I2C Tx, to avoid infinite loops.

Fixes #1343

### DIFF
--- a/src/machine/machine_nrf.go
+++ b/src/machine/machine_nrf.go
@@ -6,7 +6,12 @@ import (
 	"device/nrf"
 	"errors"
 	"runtime/interrupt"
+
+	_ "unsafe" // for go:linkname
 )
+
+//go:linkname gosched runtime.Gosched
+func gosched()
 
 var (
 	ErrTxInvalidSliceSize = errors.New("SPI write and read slices must be same size")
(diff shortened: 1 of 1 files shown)


src/machine/machine_nrf.go
```diff
@@ -295,19 +307,30 @@ func (i2c I2C) signalStop() {
...
 // writeByte writes a single byte to the I2C bus.
-func (i2c I2C) writeByte(data byte) {
+func (i2c I2C) writeByte(data byte) error {
 	i2c.Bus.TXD.Set(uint32(data))
 	for i2c.Bus.EVENTS_TXDSENT.Get() == 0 {
+		if e := i2c.Bus.EVENTS_ERROR.Get(); e != 0 {
+			i2c.Bus.EVENTS_ERROR.Set(0)
+			return errI2CBusError
+		}
+		gosched()
```
aykevl: Please avoid using `gosched` right now. This will eventually have better support using DMA and interrupts, but not right now.

src/machine/machine_nrf.go
```diff
@@ -258,12 +263,19 @@ func (i2c I2C) Configure(config I2CConfig) {
 // Tx does a single I2C transaction at the specified address.
 // It clocks out the given address, writes the bytes in w, reads back len(r)
 // bytes and stores them in r, and generates a stop condition on the bus.
-func (i2c I2C) Tx(addr uint16, w, r []byte) error {
+func (i2c I2C) Tx(addr uint16, w, r []byte) (err error) {
 	i2c.Bus.ADDRESS.Set(uint32(addr))
+	defer func() {
```
aykevl, line 263: `defer` is best avoided especially in such low level code as it can be hard to optimize. For example, I have been looking into `recover` support and `recover` support will make `defer` much more expensive.

I would suggest using a `goto` here for error handling, for details: https://eli.thegreenplace.net/2009/04/27/using-goto-for-error-handling-in-c

aykevl (commented):
Thank you! This is definitely an improvement, we should be checking for errors here.

src/machine/machine_nrf.go
dmlambea: Yes, I'm familiar with this style, like in the Linux source code. I think it's not idiomatic Go, but I agree with you that microcontrollers require special considerations.

I'm modifying it all.  Thanks!  :-)

deadprogram:
Thanks for the contribution @dmlambea and the making the reqtested changes. Now squashing/merging.

deadprogram:
I think I merged too soon. Unfortunately this PR fails the TinyHCI tests for the bbc micro:bit, seems like the i2c "locks up" and never returns data to the test program, which is located here: https://github.com/tinygo-org/tinyhci/blob/master/microbit/main.go 

How should we proceed?

aykevl:
I might have a fix, essentially extra code was added for the cleanup but the `defer` wasn't removed: https://github.com/tinygo-org/tinygo/pull/1417. Hopefully that fixes the I2C issue.

Good thing we have TinyHCI to detect this! :sweat_smile:

Issue #1320, opened 2020-08-26
There is a problem after writing atsamd / nrf5 with openocd


sago35:
#1310 makes it easier to write from openocd with feather-m4 and others.
However, two problems remain.

### 1. touchSerialPortAt1200bps() is executed when programmer is specified

As for 1, the following json does not have the flash-1200-bps-reset setting and will not cause a problem.
However, it is not reasonable to separate json, so you need to set flash-1200-bps-reset = false when -programmer is specified.

```
{
    "inherits": ["atsamd51j19a"],
    "build-tags": ["sam", "atsamd51j19a", "feather_m4"]
}
```

If you solve this problem, we will be able to write to the board in panic state.
(Normally flash-1200-bps-reset cannot be performed on a board in panic state.)

Also, we will be able to do rewrites very quickly.

### 2. Sometimes bootloader mode is triggered after writing

I think the reason for this problem is that the reset is performed multiple times when writing.
I don't know why, but changing `program *.hex reset exit` to `program *.hex verify reset exit` solved the problem.
I don't know if this is the right solution, but I'd like to fix it in some way.

sago35:
Both issues 1 and 2 also occurred with feather-nrc52840.

sago35:
> 1. touchSerialPortAt1200bps() is executed when programmer is specified

fixed. (flash: call PortReset only on other than openocd #1335)

deadprogram:
This issue was corrected in v0.15.0 which was just released. Now closing.

sago35:
> 2. Sometimes bootloader mode is triggered after writing
This one is still unresolved.

If I check with feather-m4 and wioterminal now, it's almost certainly a bootloader state.
The writes are successful, so when I reset it, it starts working normally.

Trying it with feather-nrf52840 and xiao did not reproduce the problem.

* environment
    * ubuntu 20.04
    * jlink edu mini
    * cmsis-dap-v2 debug adptor (XIAO)

deadprogram:
@sago35 is this still an issue?

sago35:
> Sometimes bootloader mode is triggered after writing
This one is still unresolved.

@deadprogram 

For the adafruit board, updating the bootloader fixed the problem.
However, I still have the following problem with wioterminal.

I would like #1397 to be merged.

However, at this point, I still don't know why adding verify eliminates the problem.
This is a problem with openocd, and the same problem occurs when I run openocd directly without tinygo.

niaow:
[...]

PR #1397, opened 2020-09-18
flash: add verify on openocd-flash

sago35:
before : Sometimes bootloader mode is triggered after writing.
after : It's always working right.

https://github.com/tinygo-org/tinygo/issues/1320#issuecomment-694752172

I've tested it in the following environments.
To sum up, it looks like verifying is required in the case of samd5x.

| board | debugger | before | after |
|--------|-----------|-----|----|
| feather-m4 | cmsis-dap-v2 | ng | ok |
| feather-m4 | jlink edu mini | ng | ok |
| wioterminal | cmsis-dap-v2 | ng | ok |
| wioterminal | jlink edu mini | ng | ok |
| feather-nrf52840 | jlink edu mini | ok | ok |
| xiao | cmsis-dap-v2 | ok | ok |
| xiao | jlink edu mini | ok | ok |

### DIFF
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -59,6 +59,7 @@ type TargetSpec struct {
 	OpenOCDTarget    string   `json:"openocd-target"`
 	OpenOCDTransport string   `json:"openocd-transport"`
 	OpenOCDCommands  []string `json:"openocd-commands"`
+	OpenOCDVerify    *bool    `json:"openocd-verify"` // enable verify when flashing with openocd
 	JLinkDevice      string   `json:"jlink-device"`
 	CodeModel        string   `json:"code-model"`
 	RelocationModel  string   `json:"relocation-model"`
--- a/main.go
+++ b/main.go
@@ -453,7 +453,11 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 			if err != nil {
 				return err
 			}
-			args = append(args, "-c", "program "+filepath.ToSlash(result.Binary)+" reset exit")
+			exit := " reset exit"
+			if config.Target.OpenOCDVerify != nil && *config.Target.OpenOCDVerify {
+				exit = " verify" + exit
+			}
+			args = append(args, "-c", "program "+filepath.ToSlash(result.Binary)+exit)
 			cmd := executeCommand(config.Options, "openocd", args...)
 			cmd.Stdout = os.Stdout
 			cmd.Stderr = os.Stderr
--- a/targets/wioterminal.json
+++ b/targets/wioterminal.json
@@ -6,5 +6,6 @@
     "flash-1200-bps-reset": "true",
     "flash-method": "msd",
     "msd-volume-name": "Arduino",
-    "msd-firmware-name": "firmware.uf2"
+    "msd-firmware-name": "firmware.uf2",
+    "openocd-verify": true
 }


deadprogram:
Seems like a change that should not have any negative consequences. @aykevl any opinion before I merge?

aykevl:
It should be fine, although `tinygo flash` might become a bit slower. However, I am wondering why this change is necessary. What verifying does is add a bit of delay, perhaps that helps avoid some race condition.

sago35:
I tried it out using a hex created in the Arduino IDE.
The result is the same, and without vefiry it seems to stay in the bootloader.

I only tried it with feather-m4.

```
# Convert bin file from Arduino output to a hex file
objcopy --input-target=binary --output-target=ihex b1.feather_m4.bin b1.feather_m4.hex --change-addresses 0x4000

# program           reset exit : ng : stay in th bootloader
/usr/local/bin/openocd -f interface/cmsis-dap.cfg -c 'transport select swd' -f target/atsame5x.cfg -c 'program b1.feather_m4.hex reset exit'

# program vefiry reset exit : ok
/usr/local/bin/openocd -f interface/cmsis-dap.cfg -c 'transport select swd' -f target/atsame5x.cfg -c 'program b1.feather_m4.hex verify reset exit'
```

sago35:
For feather-m4, the problem was solved by updating the bootloader.
However, there is no new bootloader in wioterminal.
So I am still hoping that this PR will be incorporated.


NG: 

```
UF2 Bootloader v2.0.0-adafruit.7 SFHWRO
Model: Feather M4 Express
Board-ID: SAMD51J19A-FeatherM4-v0
```

OK:

```
UF2 Bootloader v3.13.0 SFHWRO
Model: Feather M4 Express
Board-ID: SAMD51J19A-FeatherM4-v0
```

aykevl:
Ok, seems good to me. Still wondering why this is necessary, it _should_ not be necessary.
I'm a little bit worried about merging this right before a release although it probably won't be a problem. @deadprogram any opinion on this?

sago35:
I'm not entirely sure why, but there are definitely some more checks in the new version, as shown below.

see : src/main.c
https://github.com/adafruit/uf2-samdx1/compare/v2.0.0-adafruit.7..v3.13.0#diff-e0cf5b28d9b6b600f0af2bc78e8fd30ec675fd731a5da86f0c4283ffc0e40176

left: v2.0.0-adafruit.7
right: v3.13.0
image

Seeed's bootloader does not have that code.
https://github.com/Seeed-Studio/uf2-samdx1/blob/master/src/main.c#L156-L169

sago35:
I looked at the source of the adafruit bootloader, but could not find the cause.
The oldest bootloader I could build did not cause the problem.

I still have problems with wioterminal, so I want to add verify.
It is safer to merge this change now, just after the release of 0.17.

sago35:
I still feel inconvenienced without this change when developing with `Wio Terminal`.

sago35:
If for some reason verification is needed after program, it is now possible with the -ocd-verify option.
This will solve the problem of openocd-flash in WioTerminal without affecting other targets.

sago35:
Only wioterminal requires verify after openocd flash.
This item is not expected to change often.
So I made the change in target/*.json instead of tinygo's command line option.

Now ready for review again.

deadprogram:
And this cannot be done using `openocd-commands` in the `wioterminal.json`?

sago35:
It is not configurable by `eopenocd-commands`.
Verify must be added to `-c 'program main.hex reset exit'`.

```
# reset exit
openocd -f interface/cmsis-dap.cfg -c 'transport select swd' -f target/atsame5x.cfg -c 'program main.hex reset exit'

# verify reset exit
openocd -f interface/cmsis-dap.cfg -c 'transport select swd' -f target/atsame5x.cfg -c 'program main.hex verify reset exit'

# openocd-commands
openocd -f interface/cmsis-dap.cfg -c 'OPENOCD-COMMANDS' -c 'transport select swd' -f target/atsame5x.cfg -c 'program main.hex verify reset exit'
```

https://github.com/tinygo-org/tinygo/blob/e1b139149722bad78e3de1e8784c03702a553b1f/compileopts/config.go#L457-L460

deadprogram:
OK understood now @sago35 thanks for your patience in explaining. Now merging.

Issue #1351, opened 2020-09-04
Arduino UNO need SPI support for testing driver Lora sx127x


yevsia:
There are 2 Arduino UNO boards with Lora SX1278 available, but there is no SPI support at the moment, implementation is necessary if possible, thanks.

Nerzal:
I'd say this issue can now be closed, as SPI support is coming in the next release thanks to #1614

deadprogram:
Tagged for next release, at which point it can be closed.

deadprogram:
This has been released with v0.17.0 so now closing. Thanks!

PR #1398, opened 2020-09-19
machine: add atmega328p spi support

Nerzal:
Hi, here is my approach to implement SPI for the atmega328p. 

It still needs to be tested, i did already create the pr to get some feedback, if i am on the right track and to get to know if i put the things in the right places. 

I'll do some testing in the next few days :) 

Fixes: #1351


.gitignore
```diff
@@ -18,4 +18,4 @@ src/device/kendryte/*.go
 src/device/kendryte/*.s
 vendor
 llvm-build
-llvm-project
+llvm-project
```
aykevl: This is an unrelated change that should not be included in this PR (although it is a good change).

interp/README.md
```diff
@@ -51,13 +51,13 @@ But why is it necessary at all? The answer is that globals with initializers are
...
-  * Dead globals are trivial to optimize away.
-  * Constant globals are easier to detect. Remember that Go does not have global
+* Dead globals are trivial to optimize away.
+* Constant globals are easier to detect. Remember that Go does not have global
     constants in the same sense as that C has them. Constants are useful because
     they can be propagated and provide some opportunities for other
     optimizations (like dead code elimination when branching on the contents of
     a global).
-  * Constants are much more efficent on microcontrollers, as they can be
+* Constants are much more efficent on microcontrollers, as they can be
```
aykevl: These changes are unrelated to this PR, can you please undo them? They might have been added by the IDE (if you use one).

aykevl (commented):
Can you please adjust the API to match the API for other supported chips? Take a look [here](https://tinygo.org/microcontrollers/machine/microbit/#type-spi) and [here](https://tinygo.org/microcontrollers/machine/microbit/#type-spi).

Specifically:

  * The clock speed should be a `uint32` that contains the exact frequency. If you can't match the frequency, use it as an upper bound (so for example, if 6MHz is requested in `SPIConfig` but you only support 4MHz and 8MHz, use 4MHz instead). This can be done easily with a switch (`case config.Frequency <= 4e6:` etc).
  * Peripheral mode is not yet supported on any supported chips, so it may be best to leave it out. If it gets supported, it might be best to do that as a separate peripheral (`type SPIPeripheral`) - I'm not sure on that one.
  * `LSB` should be `LSBFirst` to match other chips.
  * `CPOL` is a subset of `Mode`, so please remove it.
  * Do not configure the SPI peripheral with functions other than `(SPI) Configure`. While it is easy to support on the AVR, some chips only support configuring the peripheral when it is disabled. This also keeps the public API small which is a good thing.
  * Please replace `Receive` and `Send` with a single `Transfer` function that does both a receive and a send (which is how SPI works, of course).

Nerzal:
Thanks for the review and helpful feedback! 

I removed the Periphal support for now. 
I try to change the Clock-Config later today.

Nerzal:
What just came to my mind: Is there a reason, why we don't have interfaces for SPI and things, that have the same API on different chips?

aykevl:
> What just came to my mind: Is there a reason, why we don't have interfaces for SPI and things, that have the same API on different chips?

The interface types themselves do not need to exist in the machine package, only where they are used.
Right now we don't use interfaces in the drivers repository, instead we directly use `machine.SPI`. We recently changed this for `I2C`, it uses an interface now in the drivers repository (`drivers.I2C`) so that non-machine-package I2C implementations can also be used (for example, mocks in tests). We will likely make the same change for SPI in the future.

Nerzal:
okay, i guess i have changed everything as wished :) 
now it only needs testing. 

i'll do some basic testing later on

deadprogram:
I finally was able to test with an Arduino Uno and a small strip of APA102 LEDs and it did not work for me.

This was the code I was attempting to use: https://github.com/tinygo-org/drivers/blob/release/examples/apa102/main.go

Nerzal:
Yeah sorry, i also had no time to test the changes yet. 
I'll try to find some time soon

Nerzal:
Just saw, that i had this change uncommitted. But that most likely will not fix the issues.

Still need to find some time for testing and fixing, i have not forgotten this PR, but it may take another week or two, sorry!

src/machine/machine_atmega328p.go
```diff
@@ -88,3 +89,132 @@ func (pwm PWM) Set(value uint16) {
...
+
+	// set default frequency
+	if config.Frequency == 0 {
+		config.Frequency = 6000000
+	}
+
+	var clockDivider uint8
+	switch {
+	case config.Frequency >= 93750 && config.Frequency < 187500:
+		clockDivider = SPI_CLOCK_FOSC128
```
aykevl, line 155: I have two requests for changes to these switch cases.

1. It's good enough to just do a single check for each frequency if you reverse the order of these checks. For an example, take a look here:

https://github.com/tinygo-org/tinygo/blob/1dec9dcbc466583fb0aa16c337ddf6f47ee73cb4/src/machine/machine_nrf.go#L364-L379

2. This chip can be run at various frequencies (unlike the nRF chips for the code above), for example it may need to be run at a lower frequency when the voltage is lower. Therefore, I think it would be better to base the calculation on the result of `CPUFrequency()`. Take a look here as an example:

https://github.com/tinygo-org/tinygo/blob/1dec9dcbc466583fb0aa16c337ddf6f47ee73cb4/src/machine/machine_atmega.go#L28-L32

src/machine/machine_atmega328p.go
```diff
@@ -88,3 +89,132 @@ func (pwm PWM) Set(value uint16) {
...
+	SPI_CLOCK_MASK          uint8 = 0x03
+	SPI_2XCLOCK_MASK        uint8 = 4
+)
+
+// SPIConfig
+type SPIConfig struct {
+	LSBfirst bool
+	// Frequency defaults to 6000000Hertz if not set
+	Frequency uint32
+	// Mode can be 0, 1, 2, 3, 4, 5
```
aykevl: There are only four modes: 0, 1, 2, and 3.

src/machine/machine_atmega328p.go
```diff
@@ -88,3 +89,132 @@ func (pwm PWM) Set(value uint16) {
...
+	// Use default pins if not set.
+	if config.SCK == 0 && config.SDO == 0 && config.SDI == 0 {
+		config.SCK = PB5
+		config.SDO = PB3
+		config.SDI = PB4
+	}
+
+	// set default frequency
+	if config.Frequency == 0 {
+		config.Frequency = 6000000
```
aykevl: That's a pretty high frequency! I recently switched all boards over to use 4MHz by default if unset, I wouldn't go over that to be honest. If you can only choose between 6MHz and 3MHz, I'd pick 3MHz (although I suspect it can do 4MHz as that's a nice division of the 16MHz clock speed that's used on the Uno).

Nerzal: lowered it to 4MHz for now

Nerzal:
I'll do some testing now, and try to find out why it is not working properly

src/machine/machine_atmega328p.go
Nerzal: 2. Does this really make any sense here? We do not set a specific frequency in the SPI Controller Register, but we set a clock divider. So i guess, even if a custom clock rate is being used it will still use the clock divider on that custom clock rate. 

But i am really not an expert for the atmega328p. That is just how i understand the documentation.

Nerzal: Okay, i got it now, yes i'll try to consider that. Taking a nap sometimes helps understanding things :)

Nerzal:
I'm not quite sure if it is correct to configure the Digital Pin 10 as output pin in here, because that seems to be specific for the arduino UNO boards.

Nerzal:
I still haven't managed to get the ST7735 to draw something, but the issue that the transfer did block after the first byte is resolved. 
So the problem i do now face could be due to my changes to the driver. 
I currently have nothing else to test with

Nerzal:
closed in favor of #1614

PR #1399, opened 2020-09-20
Also check lib64 for clang include path.

QuLogic:
On 64-bit Fedora, `lib64` is where the clang headers are, not `lib`. For multiarch systems, both will exist, but it's likely you want 64-bit, so check that first.


aykevl (approved):
Looks good to me, thanks!

Issue #1037, opened 2020-04-08
runtime timers are unimplemented


ryskiwt:
Using time.NewTicker causes a runtime error:

```go
package main

import (
	"time"
)

func main() {
	ticker := time.NewTicker(time.Second)
	defer ticker.Stop()

	<-ticker.C

	neverDone := make(chan struct{})
	<-neverDone
}
```

Error:

```
Uncaught (in promise) LinkError: WebAssembly.instantiate(): Import #12 module="env" function="time.startTimer" error: function import requires a callable
```

Compiled using the 0.12.0 docker image:

```
docker run ... \
    -e GOOS=js -e GOARCH=wasm \
    tinygo/tinygo:0.12.0 tinygo build -target wasm -o hoge.wasm .
```

jaddr2line:
TinyGo does not implement the `runtimeTimer` system which is used to implement `time.NewTicker` and `time.AfterFunc`. We either need to somehow implement it (not currently possible due to how `//go:linkname` is implemented), or replace parts of the `time` package.

ryskiwt:
Thanks for reply. I got it.
Is there any timeline or roadmap for implementing `go:linkname` ?

jaddr2line:
So `//go:linkname` has been implemented, but in TinyGo it preserves type information. Since `time.runtimeTimer` is an unexported type, another package cannot linkname `time.startTimer`, because the types would be different and cause a compiler error.

jaddr2line:
For now, if possible, use `time.Sleep` or loops. I do not know when we are going to be able to efficiently implement this.

jaddr2line:
As a workaround, I wonder whether we could do something like `//go:linkname` but with types.

aykevl:
Maybe we need to support `//go:linkname` with slightly-mismatched types after all. Yes, we could allow custom type names for named types, but doing something like that would still be quite invasive. I still don't like the `//go:extern` trick for globals, it requires parsing the AST. I would prefer to avoid one more dependency on having the AST available while building the LLVM IR.

[...]

PR #1402, opened 2020-09-22
Implement time.NewTimer and time.NewTicker

aykevl:
This has been requested before in #1037 and it is necessary for using the stdlib testing package. I have compared the code size before and after and except for very nontrivial cases the code size is the same (even with the scheduler additions). All examples in the drivers repo remain the same.

### DIFF
--- a/src/runtime/scheduler.go
+++ b/src/runtime/scheduler.go
@@ -16,6 +16,7 @@ package runtime
 
 import (
 	"internal/task"
+	"runtime/interrupt"
 )
 
 const schedulerDebug = false
@@ -27,6 +28,7 @@ var (
 	runqueue           task.Queue
 	sleepQueue         *task.Task
 	sleepQueueBaseTime timeUnit
+	timerQueue         *timerNode
 )
 
 // Simple logging, for debugging.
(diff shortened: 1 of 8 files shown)


src/runtime/scheduler.go
```diff
@@ -134,20 +176,43 @@ func scheduler() {
...
 
+		// Check for expired timers to trigger.
+		if timerQueue != nil && now >= timerQueue.whenTicks() {
+			scheduleLog("--- timer awoke")
+			// Pop timer from queue.
+			tn := timerQueue
+			timerQueue = tn.next
+			tn.next = nil
+			// Run the callback stored in this timer node.
+			tn.callback(tn)
```
niaow, line 187: I don't think this is really safe with coroutine lowering right now?

Any user code can define a function value of the type `func(interface{}, uintptr)` which I think is sufficient to break this?

aykevl: I'm afraid you're right. I'll have to think a bit about how to fix this. It's not trivial, as the signature is set by the time package.

Thank you for taking a look. I hadn't thought of this.

niaow: So on coroutines we should do `go tn.callback(tn)`. I dunno, there is probably a way to improve this.

niaow: Oh, nevermind, you fixed it?

aykevl: No I didn't, I haven't really thought about how to do this in a way that works with the coroutines scheduler.

dkegel-fastly:
When asyncify lands might be a good time to revisit this...?

deadprogram:
> When asyncify lands might be a good time to revisit this...?

@dkegel-fastly that is a very good question. @aykevl what do you think?

sago35:
@aykevl 
Any progress on this?

Implementing `time.startTimer` may result in some of the following tests being ok
https://tinygo.org/docs/reference/lang-support/stdlib/

* bufio
* compress/gzip
* encoding/base64
* go/printer
* image/jpeg
* io
* mime/quotedprintable

It would also be nice to be able to use time.After, time.Tick, etc.

kenbell:
@sago35 my port of @aykevl code seems to be working well to me.  It's pretty much Ayke's code, so not sure if he wants to take it forward.  If not, I think my PR is ready to review: https://github.com/tinygo-org/tinygo/pull/2954

aykevl:
Closing in favor of #2954

PR #1404, opened 2020-09-23
Use direct stack access for GC except on WebAssembly

aykevl:
Scan the stack directly instead of using our LLVM-based workaround to store frames in stack slots.

This has some benefits:

  * Much smaller code size, for larger programs around 33%. Probably faster too, I didn't measure this.
  * It might avoid some bugs in the stack frame code. The raw stack access code gets the most test coverage because of the baremetal targets.
  * It makes it possible to use conventional (stack-based) goroutines instead of goroutines based on C++ coroutine as is needed for WebAssembly. I already have a proof-of-concept of this for amd64.

The downside is of course that adding support for a new architecture becomes slightly more difficult (mainly due to `tinyg_scanCurrentStack`).

Affected targets: Linux, MacOS, GameBoy Advance and the Nintendo Switch.

---

~I know that the build fails, it's probably easy to fix but packages.debian.org is down making it somewhat difficult.~ should be fixed now, it was not as trivial as expected.


aykevl:
I found a bug, will push a fix soon.

aykevl:
Fixed the bug. There was a stack alignment issue on amd64 that I discovered while working on better testing support.

aykevl:
Slightly updated to refactor a bit more (and also split the code refactor in a separate commit). With this, all `tinygo_scanCurrentStack` functions are in src/runtime/gc_*.S files to keep things organized.

deadprogram:
@niaow can you please look at this PR when you can? Thanks.

niaow (approved):

deadprogram:
I do not have any programs that demonstrate the size difference. @aykevl can you please point me to an example of this?

aykevl:
Before:

```
$ tinygo run -size=short ./testdata/stdlib.go
  37601    9995      64 |   47596   10059
stdin:  /dev/stdin
stdout: /dev/stdout
stderr: /dev/stderr
pseudorandom number: 1298498081
strings.IndexByte: 2
strings.Replace: An-example-string
```

After:

```
$ tinygo run -size=short ./testdata/stdlib.go
  25297    6627      64 |   31924    6691
stdin:  /dev/stdin
stdout: /dev/stdout
stderr: /dev/stderr
pseudorandom number: 1298498081
strings.IndexByte: 2
strings.Replace: An-example-string
```

That's a 33% reduction in code size.

deadprogram:
Thanks for the clarification. That makes me want to have this on bare metal too! :mag: 

Now merging.

aykevl:
> Thanks for the clarification. That makes me want to have this on bare metal too!

Well, we already do this on baremetal. This PR basically makes the mentioned platforms (Linux etc) more like baremetal and less like WebAssembly as far as the GC is concerned. So no, this will not lead to a size reduction on baremetal.

PR #1405, opened 2020-09-24
implement reflect.Swapper

mathetake:
This PR implements reflect.Swapper which is used by "sort" package


src/reflect/swapper.go
```diff
@@ -1,5 +1,70 @@
...
+			is := *(*[]int16)(v.value)
+			return func(i, j int) { is[i], is[j] = is[j], is[i] }
+		case 1:
+			is := *(*[]int8)(v.value)
+			return func(i, j int) { is[i], is[j] = is[j], is[i] }
+		}
+	}
+
+	header := (*SliceHeader)(v.value)
+	tmp := unsafe.Pointer(&make([]byte, size)[0])
```
mathetake, line 28: - Note that this swapping space is created by `unsafe_New` function in the official implementation. 
- I'm worried that this is not GC safe, and it seems like we should hold the variable pointing to this array and take the unsafe.Pointer inside of the closure.

Any suggestion?

src/reflect/swapper.go
```diff
@@ -1,5 +1,70 @@
...
+	switch v.Len() {
+	case 0:
+		return func(i, j int) { panic("reflect: slice index out of range") }
+	case 1:
+		return func(i, j int) {
+			if i != 0 || j != 0 {
+				panic("reflect: slice index out of range")
+			}
+		}
+	}
```
aykevl: You could leave this out, as it increases code size. It might also increase speed but sorting slices of length zero or one is probably a no-op anyway.

src/reflect/swapper.go
```diff
@@ -1,5 +1,70 @@
...
+			is := *(*[]int32)(v.value)
+			return func(i, j int) { is[i], is[j] = is[j], is[i] }
+		case 2:
+			is := *(*[]int16)(v.value)
+			return func(i, j int) { is[i], is[j] = is[j], is[i] }
+		case 1:
+			is := *(*[]int8)(v.value)
+			return func(i, j int) { is[i], is[j] = is[j], is[i] }
+		}
+	}
```
aykevl: While this code looks correct to me, it is really a performance optimization. I would prefer to keep code size small and just fall back to the generic implementation below.

src/reflect/swapper.go
aykevl: TinyGo currently uses a conservative garbage collector, so this is _currently_ safe. We might need to rethink this once we start to use a precise garbage collector, but even then if we only do a GC during memory allocation it should be fine.

src/reflect/swapper.go
```diff
@@ -1,5 +1,70 @@
...
+
+	return func(i, j int) {
+		if uint(i) >= uint(header.Len) || uint(j) >= uint(header.Len) {
+			panic("reflect: slice index out of range")
+		}
+		val1 := unsafe.Pointer(header.Data + uintptr(i)*size)
+		val2 := unsafe.Pointer(header.Data + uintptr(j)*size)
+		memmove(tmp, val1, size)
+		memmove(val1, val2, size)
+		memmove(val2, tmp, size)
```
aykevl: You can use `memcpy` here instead of `memmove` as none of the buffers will overlap. At least in theory `memcpy` could also be faster.
It also avoids adding the new `memmove` function in the reflect package.

aykevl (commented):
Overall, this PR looks really good: it even includes extensive tests for the Swapper functionality!
Eventually I'd like to use `tinygo test sort` in this case which can replace these custom tests, but `tinygo test` doesn't support the sort package yet.

A few comments below.

src/reflect/swapper.go
mathetake: Thank you for the commet:) I have now much better understanding

mathetake:
@aykevl Thanks for the kind review :slightly_smiling_face: I didn't care about the resulting code size when I was writing this patch. Thanks! PTAL

deadprogram:
LGTM

@aykevl can we merge this?

aykevl (approved):

PR #1410, opened 2020-09-25
unix: support command line arguments

### LINKED ISSUE #456: os Args are not implemented
I tried to compile Go program with `cli` package but it requires `os.Args` in `flag.go`

`os.Args` is used [here](https://github.com/golang/go/blob/f518a96e009157c51f8ecf619032bbb7873228f9/src/flag/flag.go#L579)

**tinygo version**: tinygo version 0.7.0 linux/amd64

aykevl:
This has been a long requested feature.

fixes #456 and #541 

example usage:

```go
package main

import "os"

func main() {
	println("num args:", len(os.Args))
	for _, arg := range os.Args {
		println("arg:", arg)
	}
}
```

```
$ tinygo build -o args ./args.go
$ ./args hello world
num args: 3
arg: ./args
arg: hello
arg: world
```

### DIFF
--- a/interp/scan.go
+++ b/interp/scan.go
@@ -60,6 +60,10 @@ func (e *evalPackage) hasSideEffects(fn llvm.Value) (*sideEffectResult, *Error) 
 		return &sideEffectResult{severity: sideEffectNone}, nil
 	case name == "runtime.trackPointer":
 		return &sideEffectResult{severity: sideEffectNone}, nil
+	case name == "os.runtime_args":
+		// Special-casing this one because the (only) global it accesses changes
+		// before runtime.initAll is called.
+		return &sideEffectResult{severity: sideEffectLimited}, nil
 	case name == "llvm.dbg.value":
 		return &sideEffectResult{severity: sideEffectNone}, nil
 	case name == "(*sync/atomic.Value).Load" || name == "(*sync/atomic.Value).Store":
--- a/main_test.go
+++ b/main_test.go
@@ -13,7 +13,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
-	"sort"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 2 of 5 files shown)


bgould:
This is an exciting new feature, and looks good to me.  Do you think that we should have any tests for this functionality?

deadprogram:
Agreed with @bgould this is very exciting! Also agree that some test coverage would be a very good idea.

deadprogram:
@aykevl can you please both handle the merge conflict, and also consider adding a unit test? Thank you.

deadprogram:
Reminder to @aykevl to please resolve and merge conflict, and consider adding unit test. Thanks.

aykevl:
I have gotten it working, but somehow this change causes a segmentation fault when I run the stdlib test. The issue seems to be in the interp package where somehow an unaligned memory location is loaded from.

I plan to finish this PR after #1430 is finished and merged, as that PR should avoid this issue (hopefully).

That said, this PR is now rebased to fix the conflicts and has a test (which is indeed a good idea). I made the test in such a way that it should be easy to add environment variables in the future, which I assume will come up soon enough.

arp242:
I was looking at this yesterday – as I'm kind of waiting on it – and it segfaults when you try to import the `os` package; this doesn't just happen from the tests, it also happens when you run a program compiled with this change. I couldn't really figure out the cause though 😅

aykevl:
Yeah I haven't found the root cause yet. It seems to be related to the interp package and/or the way globals are packed into one giant global to make them scannable by the GC, and it might just be a LLVM compiler bug even. Either I can look for that, or maybe just give up on that and work on the new interp package instead (which should sidestep this problem entirely).

jfcg:
would be great to have os.Args, really fundamental stuff. i want to try on some tools with tinygo.
thanks..

hanabi1224:
Looking forward to having this!

aykevl:
Closed in favor of https://github.com/tinygo-org/tinygo/pull/1812.

aykevl:
Now _actually_ closing.

PR #1411, opened 2020-09-25
testing: implement dummy Helper method

aykevl:
This lets a few more packages pass tests: container/heap and encoding/ascii85.

---

After putting a lot of work in getting the standard library test package to work, I'm not sure whether that's the way forward. It consumes a lot RAM, which is often limited on microcontrollers (for example, it appears to run each test in a separate goroutine). And ideally we'd run tests directly on microcontrollers too. That's why I now plan to improve the testing package so that it is more compatible with the upstream testing package and get more packages to pass tests that way.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,8 +195,10 @@ test: wasi-libc
 # implied -v flag).
 .PHONY: tinygo-test
 tinygo-test:
+	$(TINYGO) test container/heap
 	$(TINYGO) test container/list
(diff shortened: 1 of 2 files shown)


aykevl:
The one failure appears to be unrelated to the code changes.

deadprogram:
Passed TinyHCI after restart.

> And ideally we'd run tests directly on microcontrollers too.

Yes. That is exactly what we want.

PR #1412, opened 2020-09-25
testing: implement some benchmark stubs

aykevl:
This allows the following packages to pass tests:

  * crypto/des
  * ~crypto/rc4~
  * encoding/hex

... which are two outdated ciphers and one potentially useful package. Still, it's a good thing to get as many packages to pass tests, if nothing else it helps to catch regressions in the compiler.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -198,7 +198,9 @@ tinygo-test:
 	$(TINYGO) test container/heap
 	$(TINYGO) test container/list
 	$(TINYGO) test container/ring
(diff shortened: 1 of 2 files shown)


aykevl:
Removed crypto/rc4 as it doesn't work on Go 1.11.

aykevl:
Rebased, after a conflict with #1411

deadprogram:
Looking good, now merging. Thanks @aykevl

PR #1413, opened 2020-09-25
nintendoswitch: Fix crash when printing very long lines.

racerxdl:
A silly mistake on print buffer length check was making unmapped memory access. This fixes it.


aykevl (approved):
I remember looking at this and thinking about correctness, but somehow I still didn't catch this off-by-one.

PR #1414, opened 2020-09-26
arm64: Add support for system calls (SVC)

racerxdl:
Since ARM64 (AArch64) uses `X` instead `R` on registers, this change is required to `SVCall` to work.

I also added other registers to clobbered registers by the AArch64 calling convention.


aykevl:
Thank you! Looks reasonable to me.

PR #1417, opened 2020-09-27
nrf: fix double stop signal in I2C

aykevl:
It looks like https://github.com/tinygo-org/tinygo/pull/1392 introduced a problem, which I suspect may be caused by sending a stop signal twice.


deadprogram:
Thanks @aykevl now merging.

Issue #1418, opened 2020-09-28
wasm tests timeout on i686


QuLogic:
While updated to 0.15.0, I also started running the wasm tests. This works on 64-bit systems, but fails on 32-bit: https://koji.fedoraproject.org/koji/taskinfo?taskID=52392695
```
+ make wasmtest
go test ./tests/wasm
2020/09/28 07:13:24 Starting server at ":8826" for dir: /tmp/wasm_test630700884
2020/09/28 07:13:24 Command: tinygo build -o /tmp/wasm_test630700884/chan.wasm -target wasm testdata/chan.go; err=<nil>; full output:
2020/09/28 07:13:25 Command: tinygo build -o /tmp/wasm_test630700884/event.wasm -target wasm testdata/event.go; err=<nil>; full output:
2020/09/28 07:13:28 Command: tinygo build -o /tmp/wasm_test630700884/fmt.wasm -target wasm testdata/fmt.go; err=<nil>; full output:
2020/09/28 07:13:29 Command: tinygo build -o /tmp/wasm_test630700884/fmtprint.wasm -target wasm testdata/fmtprint.go; err=<nil>; full output:
--- FAIL: TestChan (5.70s)
    chan_test.go:31: unable to execute *log.EnableParams: context deadline exceeded
2020/09/28 07:13:30 Command: tinygo build -o /tmp/wasm_test630700884/log.wasm -target wasm testdata/log.go; err=<nil>; full output:
--- FAIL: TestEvent (6.26s)
    event_test.go:41: log1: 
    event_test.go:42: log2: 
    event_test.go:44: context deadline exceeded
--- FAIL: TestFmt (8.96s)
    fmt_test.go:38: log1: 
    fmt_test.go:40: unable to execute *log.EnableParams: context deadline exceeded
--- FAIL: TestFmtprint (10.36s)
    fmtprint_test.go:34: log1: 
    fmtprint_test.go:36: unable to execute *log.EnableParams: context deadline exceeded
--- FAIL: TestLog (11.01s)
    log_test.go:39: log1: 
    log_test.go:41: unable to execute *log.EnableParams: context deadline exceeded
FAIL
FAIL	_/builddir/build/BUILD/tinygo-0.15.0/tests/wasm	11.010s
FAIL
```
I tried increasing the timeouts from 5 to 15 seconds, but that didn't seem to have an effect. I'm not sure if this is a bug in tinygo, or the chrome driver stuff.

aykevl:
[...]

PR #1420, opened 2020-09-29
ci: fix `make wasmtest`

aykevl:
This fixes issue https://github.com/tinygo-org/tinygo/issues/1418. In short, it appears there was a race condition that was only visible on `GOOS=386` but not on `GOOS=amd64`. Updating to a more recent chromedp version fixes the issue.


aykevl:
@deadprogram it appears that some devices are failing right now.

deadprogram:
Had a power failure, now restarted and all passing.

deadprogram:
Now merging, thanks @aykevl

PR #1421, opened 2020-09-30
add wasm-abi field in TargetSpec && set generic for WASI by default

mathetake:
- add `wasm-abi` field in TargetSpec (defaults to the empty string)
- change default value of `-wasm-abi` flag from `js` to the empty string
- override `wasm-abi` value by `-wasm-abi` when the flag is set to non-empty string
- set `wasm-abi` field values for `wasm.json` and `wasi.json`
    - `js` for wasm.json
    - `generic` for wasi.json

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -79,10 +79,7 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 	// keep functions interoperable, pass int64 types as pointers to
 	// stack-allocated values.
 	// Use -wasm-abi=generic to disable this behaviour.
-	if config.Options.WasmAbi != "" {
-		config.Target.WasmAbi = config.Options.WasmAbi
-	}
-	if config.Target.WasmAbi == "js" {
+	if config.WasmAbi() == "js" {
 		err := transform.ExternalInt64AsPtr(mod)
 		if err != nil {
 			return err
--- a/builder/build.go
+++ b/builder/build.go
@@ -79,7 +79,10 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 	// keep functions interoperable, pass int64 types as pointers to
 	// stack-allocated values.
 	// Use -wasm-abi=generic to disable this behaviour.
-	if config.Options.WasmAbi == "js" && strings.HasPrefix(config.Triple(), "wasm") {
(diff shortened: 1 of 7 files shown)


targets/wasi.json
```diff
@@ -18,5 +18,6 @@
 		"--no-demangle",
 		"{root}/lib/wasi-libc/sysroot/lib/wasm32-wasi/libc.a"
 	],
-	"emulator":      ["wasmtime"]
+	"emulator":      ["wasmtime"],
+	"wasm-abi":      "generic"
```
mathetake, line 22: I am not sure if this is necessary. It seems ok to just not set and use default empty string

builder/build.go
```diff
@@ -79,7 +79,10 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 	// keep functions interoperable, pass int64 types as pointers to
 	// stack-allocated values.
 	// Use -wasm-abi=generic to disable this behaviour.
-	if config.Options.WasmAbi == "js" && strings.HasPrefix(config.Triple(), "wasm") {
+	if config.Options.WasmAbi != "" {
+		config.Target.WasmAbi = config.Options.WasmAbi
```
aykevl: Please do not modify the `config.Target` field, it should contain exactly what is extracted from the JSON files.

Instead, you can either:

 * Add a getter to the `config` variable (`*compileopts.Config`), so you can simply call `config.WasmAbi()` to get the value.
 * Use a local variable instead.

The getter would be slightly cleaner, as it matches the pattern of other configurations (`CGO_ENABLED`, `GC`, `NeedsStackObjects`, etc).

targets/wasi.json
aykevl: It is not necessary, but it doesn't harm either.

builder/build.go
mathetake: added the getter in https://github.com/tinygo-org/tinygo/pull/1421/commits/d61d49c69b4ce0032b527089db852d8984bcf331

deadprogram: @mathetake thanks for the addition. 

@aykevl anything else needed for us to merge this now?

aykevl:
No that's all. Thank you!

PR #1424, opened 2020-10-04
Make lib64 clang include path check more robust.

QuLogic:
On Fedora 33+, there is a buggy package that installs to `/usr/lib64/clang/{version}/lib`, even on 32-bit systems. The original
code from #1399 sees the `/usr/lib64/clang/{version}` directory, checks for an `include` subdirectory, and then gives up because it doesn't exist.

To be more robust, check both `/usr/lib64/clang/{version}/include` and `/usr/lib/clang/{version}/include`, sorted by version subdirectory.

### DIFF
--- a/builder/env.go
+++ b/builder/env.go
@@ -6,6 +6,9 @@ import (
 	"os/exec"
 	"path/filepath"
 	"sort"
+	"strings"
+
+	"tinygo.org/x/go-llvm"
 )
 
 // getClangHeaderPath returns the path to the built-in Clang headers. It tries
@@ -26,6 +29,7 @@ func getClangHeaderPath(TINYGOROOT string) string {
 
 	// It looks like we are built with a system-installed LLVM. Do a last
 	// attempt: try to use Clang headers relative to the clang binary.
+	llvmMajor := strings.Split(llvm.Version, ".")[0]
 	for _, cmdName := range commands["clang"] {
 		binpath, err := exec.LookPath(cmdName)
 		if err == nil {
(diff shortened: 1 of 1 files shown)


QuLogic:
Hmm, still not as robust as I'd like, as 9.0.0 sorts after 10.0.0... Though I guess that's an existing bug.

QuLogic:
I added a normalization that [should work](https://play.golang.org/p/QiiGqjZj7wp).

QuLogic:
Oops, I totally messed up #1399, as `dirnames` does _not_ contain full paths and thus doesn't really contain both `lib` and `lib64` paths. This should be fixed now.

aykevl:
TinyGo should only use the LLVM major version that matches the version it was built with. However, it appears that it will look for any LLVM version and use that. Is that true?

On Debian you can have multiple LLVM versions installed simultaneously. It looks like Fedora also has this ability, and uses the LLVM major version in the binary path. If so, I would suggest only using LLVM major versions that match the LLVM major version that TinyGo was built with. This also avoids the problem of sorting LLVM 9.0.0 before LLVM 10.0.0.

QuLogic:
That's true, but it's how it's always worked. I wouldn't mind using only the version it compiled against, but I don't know how to get that number?

QuLogic:
OK, I just noticed how the major version is pulled to populate `commands["clang"]`, so how does that look?

deadprogram:
@QuLogic did you see the test failure here?

https://app.circleci.com/pipelines/github/tinygo-org/tinygo/3894/workflows/4da7f4b4-ce0a-43cb-8156-f45592ee64a0/jobs/20464

QuLogic:
I saw it, but I don't know how it's related, as it's failing in the Makefile which I didn't change.

deadprogram:
And now the tests all appear to be passing. Weird.

Anyhow, @aykevl any other feedback on this PR?

aykevl:
> And now the tests all appear to be passing. Weird.

Yes, I think I restarted them. After one test caches the result of the wasi-libc build, it will be used by the other test in the next run hence why simply restarting it fixes the issue.

PR #1425, opened 2020-10-05
board/teensy40: Add support for Teensy 4.0

ardnew:
This PR adds very basic support for Teensy 4.0 (capable of running `src/examples/blinky1`), using a new target `machine` and `runtime` package for NXP's i.MXRT1062 MCU. These packages should be reusable for future Teensy 4.1 support as well.

### Initialization

  - The flash configuration binary blob defined in `targets/teensy40.s` is sourced (with attribution) from the Teensyduino core ([hardware/avr/cores/teensy4/bootdata.c](https://github.com/PaulStoffregen/cores/blob/master/teensy4/bootdata.c)).

### Memory

  - The MCU on this board has no on-chip flash memory. The Teensy adds a 2 MiB NOR flash chip connected to the FlexSPI serial interface. The memory is partitioned for TinyGo as follows (see the **Limitations** section further below):

|RAM|Name|Address|Size|Contents|
|:----:|:------:|:-------:|:----:|:---------|
|RAM1|`ITCM`|`0x00000000`|512 KiB|*(Unused!)*|
|RAM1|`DTCM`|`0x20000000`|512 KiB|Stack and globals|
|RAM2|`RAM`|`0x20200000`|512 KiB|Heap|
|Flash|`Flash`|`0x60000000`|2 MiB|Code|

>  For a more complete description of these memory regions, and the motivation for this configuration in TinyGo, see the **Memory Layout** section on the product page: https://www.pjrc.com/store/teensy40.html

### Clocks

  - The ARM PLL is configured to generate a core clock frequency of **600 MHz**. 

  - SysTick is configured to use the onboard 24 MHz OSC clock source instead of the insane ARM core clock (in case the user wants to scale down or sleep the ARM core, it will not affect SysTick). 

  - The `machine` package's delay/sleep timer uses a dedicated 32-bit timer peripheral (`PIT`) also sourced from the 24 MHz OSC.

### GPIO

There are technically 9 GPIO peripherals in the MCU (`GPIO1`-`GPIO9`). However, half of these are just the high-speed counterparts connected to the exact same pins. From the i.MXRT1062 Processor Reference Manual (**Chapter 12 - GPIO**):

>  High-speed GPIOs exist in this device:
>    - GPIO1-5 are standard-speed GPIOs that run off the IPG_CLK_ROOT, while
>      GPIO6-9 are high-speed GPIOs that run at the AHB_CLK_ROOT frequency.
>      See the table "System Clocks, Gating, and Override" in CCM chapter.
>    - Regular GPIO and high speed GPIO are paired (GPIO1 and GPIO6 share the
>      same pins, GPIO2 and GPIO7 share, etc). The IOMUXC_GPR_GPR26-29
>      registers are used to determine if the regular or high-speed GPIO
>      module is used for the GPIO pins on a given port.

So to get rid of confusion, only four GPIO ports are defined in the TinyGo machine package:

  - `portA`: `GPIO1`/`GPIO6`
  - `portB`: `GPIO2`/`GPIO7`
  - `portC`: `GPIO3`/`GPIO8`
  - `portD`: `GPIO4`/`GPIO9`

By default, the device is configured to always use only the high-speed GPIO for every pin.

## Limitations

1. Teensyduino takes advantage of the ITCM memory by copying the user's executable code to this region at startup (leaving the initialization code in flash). I have been unable to duplicate this functionality in TinyGo, so all executable code is currently placed in flash. The ITCM region is preserved in the linker script so that someone with more insight can fix this in the future.

2. The `machine` package's sleep timer uses an interrupt to indicate when rollover occurs. A volatile flag variable is set in the ISR to coordinate with the main goroutine when this occurs. In the main goroutine, waiting for this volatile flag is typically done with ARM instruction `wfi`. However, this instruction is causing a hard fault here. So the main goroutine instead spins in an infinite loop while checking this flag. This should be replaced with `wfi` once the reason for the hard fault is determined.

3. _**It produces really large code**_ (see below). The vast majority of this support package is based off of NXP's MCUXpresso i.MXRT1062 SDK and _not_ the Teensyduino BSP. This means the initialization code is much more generalized and capable of supporting non-Teensy devices, but on the other hand a lot of superfluous definitions and logic probably (i.e., definitely) not needed for Teensy 4.0 support. I'm not sure if this is the source of the bulk, or maybe the executable layout described by the linker script is affecting this, or...? Luckily the Teensy has enormous amounts of memory, so I haven't felt too obligated to strip this down. It wouldn't hurt to take a second pass and look for wasted space.
```
tinygo build -size short -o test.hex -target=teensy40 examples/blinky1
18336      48  131272 |   18384  131320
```


src/runtime/runtime_cortexm_hardfault_config.go
```diff
@@ -0,0 +1,27 @@
+// +build nxp,!mimxrt1062
```
deadprogram: Seems to me the contents of this file should be added to a new file in the  `device` package similar to what you did for https://github.com/tinygo-org/tinygo/tree/release/src/device/stm32

src/runtime/runtime_cortexm_hardfault_config_mimxrt1062.go
```diff
@@ -0,0 +1,27 @@
+// +build nxp,mimxrt1062
```
deadprogram: Same as previous comment about adding to `device` package.

src/runtime/runtime_mimxrt1062_clock.go
```diff
@@ -0,0 +1,1242 @@
...
+	clkIpLpuart8.enable(true) //  ~
+
+	// ----------------------------------------------------------------- FLEXIO --
+	clkIpFlexio1.enable(true) // enable FLEXIO clock gates
+	clkIpFlexio2.enable(true) //  ~
+}
+
+type clock uint32
+
+// Named oscillators
```
deadprogram: Also consider adding to a `device` package file.

src/runtime/runtime_mimxrt1062_clock.go
```diff
@@ -0,0 +1,1242 @@
...
+	modeClkWait clockMode = 1 // Transfer to wait mode
+	modeClkStop clockMode = 2 // Transfer to stop mode
+)
+
+func (m clockMode) set() {
+	nxp.CCM.CLPCR.Set((nxp.CCM.CLPCR.Get() & ^uint32(nxp.CCM_CLPCR_LPM_Msk)) |
+		((uint32(m) << nxp.CCM_CLPCR_LPM_Pos) & nxp.CCM_CLPCR_LPM_Msk))
+}
+
+const (
```
deadprogram: `device` probably.

src/runtime/runtime_mimxrt1062_mpu.go
```diff
@@ -0,0 +1,349 @@
...
+
+	// [7] QSPI flash:
+	//          2 MiB, +ACCESS, #NORMAL, +EXEC, -share, +CACHE, +BUFFER, -subregion
+	mpu.setRBAR(7, 0x60000000)
+	mpu.setRASR(rs2MB, apFull, exNormal, true, false, true, true, false)
+
+	mpu.enable(true)
+}
+
+// MPU Type Register Definitions
```
deadprogram: Seems like `device` bitfield stuff too.

deadprogram:
First of all, wow, this is a lot of effort, thank you so much @ardnew for working on it.

I made some comments about consts that seem to me would be better off being put into `device/nxp` files. See what you think. I probably missed some, but you get the idea.

src/runtime/runtime_cortexm_hardfault_config.go
ardnew: Moved to new file `src/device/nxp/hardfault.go`

src/runtime/runtime_cortexm_hardfault_config_mimxrt1062.go
ardnew: Moved to new file `src/device/nxp/hardfault_mimxrt1062.go`

src/runtime/runtime_mimxrt1062_clock.go
ardnew: Moved to new file `src/device/nxp/clock_mimxrt1062.go`

src/runtime/runtime_mimxrt1062_clock.go
ardnew: Same as previous (moved to new file `src/device/nxp/clock_mimxrt1062.go`)

src/runtime/runtime_mimxrt1062_mpu.go
ardnew: Moved to new file `src/device/nxp/mpu_mimxrt1062.go`

ardnew:
@deadprogram I moved quite a bit to the `device/nxp` package - all of the bitfield definitions, types, and methods for interacting with the chip-specific clocks and MPU. If moving those methods wasn't intended, it would take substantial refactoring to replace all of them with functions (after moving them back to the runtime package) accepting the receivers as argument; it didn't seem appropriate to separate them this way anyhow.

The code that actually **uses** these types and methods for configuring the device were left in the runtime package.

src/machine/board_teensy40.go
```diff
@@ -0,0 +1,177 @@
+// +build teensy40
+
+package machine
+
+const (
+	NUM_DIGITAL_IO_PINS = 40
+	NUM_ANALOG_IO_PINS  = 14
+)
```
aykevl: Are these used anywhere? If not, I would suggest removing them (I suspect they are necessary in the Arduino environment).

src/machine/board_teensy40.go
```diff
@@ -0,0 +1,177 @@
...
+	initSPI()
+	initI2C()
+}
+
+// -- LEDs ---------------------------------------------------------------------
+
+const (
+	NUM_BOARD_LED = 1
+
+	LED_BUILTIN = D13
```
aykevl: That's also an Arduino thing. What is called `LED_BUILTIN` in Arduino, is just `machine.LED` in TinyGo. No need for long names if you have proper namespacing.

src/machine/board_teensy40.go
```diff
@@ -0,0 +1,177 @@
...
+	initLED()
+	initUART()
+	initSPI()
+	initI2C()
+}
+
+// -- LEDs ---------------------------------------------------------------------
+
+const (
+	NUM_BOARD_LED = 1
```
aykevl: What's the purpose of these constants, `NUM_BOARD_LED`, `NUM_UART_INTERFACES`, etc?

src/machine/board_teensy40.go
```diff
@@ -0,0 +1,177 @@
...
+	A12 = PA30 // D26 | [AD_B1_14] {128+3}
+	A13 = PA31 // D27 | [AD_B1_15] {128+4}
+)
+
+func init() {
+	initLED()
+	initUART()
+	initSPI()
+	initI2C()
+}
```
aykevl: Initialization should happen in the runtime, not in the machine package. Or otherwise, it should happen in the `Configure` call of a particular peripheral.

For example, clock initialization is usually done in the runtime directly. UART initialization is done in the `UART.Configure` method.
In this case, all these functions are empty so there really is no purpose of including them other than making the code longer (and thus harder to read).

src/machine/machine_mimxrt1062.go
```diff
@@ -0,0 +1,591 @@
...
+type muxSelect struct {
+	mux uint8                // AF mux selection (NOT a Pin type)
+	sel *volatile.Register32 // AF selection register
+}
+
+// connect configures the IOMUXC controller to route a given pin with alternate
+// function to a desired peripheral (see godoc comments on type muxSelect).
+func (s muxSelect) connect() {
+	s.sel.Set(uint32(s.mux))
+}
```
aykevl: This code appears to be dead.

src/machine/uart.go
```diff
@@ -1,4 +1,4 @@
-// +build avr esp nrf sam sifive stm32 k210 nxp
+// +build avr esp nrf sam sifive stm32 k210 nxp,!mimxrt1062
```
aykevl: Maybe it would be better to explicitly include devices instead of explicitly excluding devices?

src/device/nxp/hardfault.go
```diff
@@ -0,0 +1,29 @@
+// Hand created file. DO NOT DELETE.
+// Hardfault aliases for definitions that have inconsistent naming (which are
+// auto-generated by gen-device-svd.go) among devices in package nxp.
+
+// +build nxp,!mimxrt1062
+
+package nxp
+
+const (
+	HardFault_CFSR_IACCVIOL    = SystemControl_CFSR_IACCVIOL
```
aykevl, line 10: It appears that these values are not specific to NXP chips but instead for the whole Cortex-M7 family. Take a look here:
https://developer.arm.com/documentation/dui0646/b/cortex-m7-peripherals/system-control-block/configurable-fault-status-register?lang=en

I would suggest moving this into src/device/arm/arm.go (which implies hand-defining them, but that only needs to be done once). That also avoids NXP-specific code in src/runtime/runtime_cortexm_hardfault_debug.go.

targets/teensy40.json
```diff
@@ -0,0 +1,25 @@
+{
+  "inherits": ["cortex-m"],
+  "llvm-target": "armv7em-none-eabi",
+  "cpu": "cortex-m7",
+  "build-tags": ["teensy40", "teensy", "mimxrt1062", "nxp"],
+  "automatic-stack-size": false,
+  "default-stack-size": 131072,
```
aykevl: Isn't 128K a bit on the large side? I would suggest 2K or 4K instead (in fact, most chips use 1K and that's usually enough).

targets/teensy40.json
```diff
@@ -0,0 +1,25 @@
...
+  "cflags": [
+    "--target=armv7em-none-eabi",
+    "-Qunused-arguments",
+    "-mfloat-abi=hard",
+    "-mfpu=fpv5-d16"
+  ],
+  "ldflags": [
+    "--emit-relocs",
+    "--gc-sections"
+  ],
```
aykevl: These flags are already defined in cortex-m.json so don't need to be repeated here.

targets/teensy40.json
```diff
@@ -0,0 +1,25 @@
...
+  "llvm-target": "armv7em-none-eabi",
+  "cpu": "cortex-m7",
+  "build-tags": ["teensy40", "teensy", "mimxrt1062", "nxp"],
+  "automatic-stack-size": false,
+  "default-stack-size": 131072,
+  "cflags": [
+    "--target=armv7em-none-eabi",
+    "-Qunused-arguments",
+    "-mfloat-abi=hard",
+    "-mfpu=fpv5-d16"
```
aykevl: Unfortunately, floating point is not yet supported in TinyGo stack switching so I would suggest using softfloat until it is. Otherwise, if multiple goroutines are using the FPU there is a real risk of one goroutine corrupting the float registers of the other goroutine.

That said, these flags only apply to C code so are of little importance to most Go code (and in fact would introduce a binary incompatibility between Go and C).

aykevl (commented):
I haven't looked at all code yet, it's a lot of code to review.
But here is something to get started.

Overall, it looks pretty good although it's very unfortunate such a large amount of code is needed for just blinking an LED.

src/machine/board_teensy40.go
ardnew: Nope, not used, will remove

src/machine/board_teensy40.go
ardnew: Same, not used in TinyGo packages, will remove

src/machine/board_teensy40.go
ardnew: Mimicking the Arduino headers again. Will remove

src/machine/board_teensy40.go
ardnew: These were not intended to initialize the peripherals (which *is* done in the `runtime` package, including clock initialization).

Quite a few other boards register interrupts (e.g. UART RX) in the `machine` package's `func init()`. These were I guess placeholders for any similar initialization. But like you see, they are empty in this initial PR, so I can remove them until they are needed.

src/machine/machine_mimxrt1062.go
ardnew: Dead code for this PR, since GPIO is the default mux selection for all pins, but this is very much needed in all subsequent peripheral PRs.

src/machine/uart.go
ardnew: Going forward that seems reasonable. It seems to me kinda risky (and a lot of work) trying to go back and making sure you've accounted for all appropriate devices in existing code like this

src/device/nxp/hardfault.go
ardnew: The values weren't the problem here, but rather the identifiers that were generated from the SVD. The values actually apply to several Cortex-M families (not just the M7).

For some reason, the NXP MIMXRT1062 SVD used `SystemControl_` prefix, while other Cortex-M devices used `SCB_`. The existing code trying to reference `SCB_` fields thus failed to compile for this chip (see: `runtime/runtime_cortexm_hardfault_debug.go`), even though the values did exist (as you noted), they just needed to use correct identifiers with a different prefix. 

So I don't think it would be appropriate to move these definitions to a more global scope, since most Cortex-M devices do not use `SystemControl_`-prefixed identifiers for these field values.

Hope that makes sense and doesn't miss the point you're making!

targets/teensy40.json
ardnew: Definitely on the large side! Wasn't sure if that would be problematic or not. 

The reason for this, is that an entire 128K physical RAM bank (whose accesses are all performed via the 600 MHz AHB bus) is completely unused in TinyGo! This seemed like an unfortunate waste of resources, so I just handed it all over for the stack. The heap is allocated on the other physical 128K RAM bank.

(I need a guru such as yourself to help allocate code to this unused bank during startup so that we have super fast code accesses. Currently it is all stored in flash, and we rely on the MPU (which is configured in this PR!) to cache these for fast code accesses from flash)

targets/teensy40.json
ardnew: Thanks, wasn't sure how the inheritance worked among these JSON files - if it intelligently merges JSON arrays so that unique elements are preserved among both parent and child, or if it would override the parent if the child provided its own definition..

Anyhow, I think your comment implies that it *Does The Right Thing*. Will remove.

targets/teensy40.json
ardnew: Interesting! Will remove, thanks.

targets/teensy40.json
ardnew: Changed to 4K (4096)

aykevl: This is the stack size for goroutines, which are allocated on the heap. The scheduler has a different stack which is set in the linker script (`_stack_size`).

src/device/nxp/hardfault.go
aykevl: What I suggest is to ignore the identifiers in the SVD file entirely and define them manually, as they are NXP specific and even vary by chip (apparently) while their values are the same across all chips (M7 or maybe lower). In fact, most SVD files do not contain definitions for these peripherals that are defined by ARM.

I know that these identifiers are used in `runtime/runtime_cortexm_hardfault_debug.go`, and that's the problem. A generic Cortex-M runtime file should not import a NXP device package, that's a bug that I apparently missed previously. So basically, `runtime/runtime_cortexm_hardfault_debug.go` should be modified to import device/arm instead of device/nxp.

src/machine/uart.go
aykevl: No, it's not risky at all. The smoke tests will fail if you made a mistake. And in this case, there is only one file to include: the chip in the Teensy 3.6.

```suggestion
// +build avr esp nrf sam sifive stm32 k210 nxpmk66f18
```

ardnew: Oh you are referring to this one board, not all boards included by each of the other build tags too... that makes more sense.

ardnew: Also, I should have realized this is only a temporary build constraint until the UART PR #1454 gets merged, which should probably keep the `nxp` tag, as it applies to both Teensy 3.6 and 4.0

src/device/nxp/hardfault.go
ardnew: The routines `GetFaultStatus()`, `(MemFaultStatus) Address()`, and `(BusFaultStatus) Address()` all refer to a `nxp.SystemControl` block of registers. 

The `SCB_Type` in package `device/arm` doesn't have the registers they are referencing.

Besides manually re-defining the original fields in, say, `device/arm/faultstatus.go`, it seems the `CFSR`, `MMFAR`, and `BFAR` registers will all also need to be manually defined here as well.

Were you expecting these changes with your original suggestion? Or should this be moved to its own PR?

aykevl: > The routines `GetFaultStatus()`, `(MemFaultStatus) Address()`, and `(BusFaultStatus) Address()` all refer to a `nxp.SystemControl` block of registers.

Yes, and that's a problem. This should be changed to not import device/nxp and import device/arm instead.

> The `SCB_Type` in package `device/arm` doesn't have the registers they are referencing.

Yes, so they should be added to `SCB_Type`.

> Besides manually re-defining the original fields in, say, `device/arm/faultstatus.go`, it seems the `CFSR`, `MMFAR`, and `BFAR` registers will all also need to be manually defined here as well.

What do you mean re-defining? It is true that these registers will need to be written by hand in the device/arm package, but only once.

ardnew: I mean "re-defining" in the sense that several (many?) devices, such as these `device/nxp` implementations, already have these fields defined in their SVD auto-generated files. So, for these cases, we are duplicating the field and register definitions by also adding them to package `device/arm`. 

I _didn't_ mean "re-define" as in renaming/moving or changing the values in those existing generated packages.

ardnew: I wasn't sure if changing the type definition of `SCB_Type` in `device/arm` was warranted for this PR. 

But I will go ahead and make these changes and push in a few hours

ardnew: Marking this conversation resolved, see 4091bca for all requested changes.

deadprogram:
I think one last merge conflict, and this can be merged!

ardnew:
rebased against `dev`, no more conflicts on my end. 

verified via `make smoketest` and LED blinking with flash command on a _real_ device:

`tinygo flash -target=teensy40 src/examples/blinky1`

deadprogram:
Now merging, thank you for all the work on this PR!

PR #1426, opened 2020-10-07
wasm: use the fixed length buffer for putchar

mathetake:
this PR follows the Nintendo Switch's implementation of puchar for WASM/WASI target

### DIFF
--- a/src/runtime/runtime_wasm.go
+++ b/src/runtime/runtime_wasm.go
@@ -34,16 +34,10 @@ var (
 )
 
 func putchar(c byte) {
-	if putcharPosition >= putcharBufferSize {
-		putcharIOVec.bufLen = putcharPosition
-		fd_write(stdout, &putcharIOVec, 1, &putcharNWritten)
-		putcharPosition = 0
-	}
-
 	putcharBuffer[putcharPosition] = c
 	putcharPosition++
 
-	if c == '\n' {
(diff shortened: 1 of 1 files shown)


mathetake:
@aykevl Can you please take a look? :bow:

aykevl:
The Nintendo Switch has the same issue. It solves this by buffering up data and writing it out on the first `\n` received, or when the buffer is full. I think that's a much simpler approach that works just as well in practice.

https://github.com/tinygo-org/tinygo/blob/47dc1087e81463d8614b4b7437bbd3c0a489bbbc/src/runtime/runtime_nintendoswitch.go#L58-L70

> Theoretically we could wait for `\n` to come into the host just like in `wasm_exec.js`, but this potentially leads to memory leakage since 1.) we cannot force all the developers to emit `\n` for sure, and 2.) we switch contexts and message emitters will change constantly, and therefore we have to manage log buffers for each request.

Simply creating a new line after a certain number of characters (for example, 120) avoids memory leakage. And I don't see the need for buffers for each separate request, they go to the same stdout anyway.

mathetake:
@aykevl Thanks for the comment, and I'm now convinced that buffering characters in fixed length slice is enough. PTAL

aykevl:
Thank you! I've squashed/merged the changes.

PR #1429, opened 2020-10-07
compiler: fix undefined behavior in wordpack

aykevl:
Previously, EmitPointerPack would generate an out-of-bounds read from an
alloca. This commit fixes that by creating an alloca of the appropriate
size instead of using the size of the to-be-packed data (which might be
smaller than a pointer).

I discovered this error while working on a rewrite of the interp
package, which checks for out-of-bounds reads and writes. There I
discovered this issue when the image package was compiled.


aykevl:
Apparently this doesn't work well without the interp rewrite.

Issue #1409, opened 2020-09-25
Wasm: Panic trying to make exported function async


asdine:
The compiler panics trying to compile what it appears to be functions that expect other functions.

Target: wasm
Tinygo version: Docker version of 0.15.0

Stacktrace:

```
panic: trying to make exported function async: (github.com/genjidb/genji/document.Stream).Iterate$invoke

goroutine 1 [running]:
github.com/tinygo-org/tinygo/transform.(*coroutineLoweringPass).findAsyncFuncs(0xc008591040)
        /home/circleci/project/transform/coroutines.go:186 +0xa64
github.com/tinygo-org/tinygo/transform.(*coroutineLoweringPass).load(0xc008591040, 0x20, 0x9a16790)
        /home/circleci/project/transform/coroutines.go:305 +0x36c
github.com/tinygo-org/tinygo/transform.LowerCoroutines(0x8be16c0, 0x8be1601, 0x0, 0x0)
        /home/circleci/project/transform/coroutines.go:70 +0x15f
github.com/tinygo-org/tinygo/transform.Optimize(0x8be16c0, 0xc000336060, 0x2, 0x2, 0x5, 0x0, 0x0, 0x0)
        /home/circleci/project/transform/optimizer.go:120 +0xef0
github.com/tinygo-org/tinygo/builder.Build(0x7ffd9a4f9efb, 0x21, 0x7ffd9a4f9ed2, 0x10, 0xc000336060, 0xc008591ba8, 0x0, 0x0)
        /home/circleci/project/builder/build.go:109 +0x337a
main.Build(0x7ffd9a4f9efb, 0x21, 0x7ffd9a4f9ed2, 0x10, 0xc000300000, 0x0, 0x12)
        /home/circleci/project/main.go:94 +0xc5
main.main()
        /home/circleci/project/main.go:856 +0x1f96
```

Besides the error and the stacktrace above, it is impossible for me to write a simple code sample for you to reproduce it because the behavior of the compiler is really difficult to grasp: Two statements cause the compiler to fail, but if I use statement A or statement B separately, the compilation succeeds. If you have any tips on how I could debug this more efficiently I will try to give more details.

aykevl:
This looks like a duplicate of https://github.com/tinygo-org/tinygo/issues/733 so closing this issue in favor of it.

Issue #733, opened 2019-11-17
panic: trying to make exported function async


ellemlabs:
I'm back with more weekend-ruining excitement.

I put a channel receive inside a method that cannot be directly called from the javascript. When compiling I get the following error. I notice that if I start putting other channel receives inside of arbitrary functions than I can't reproduce the error.

I'm not sure if this is a bug or the compiler is trying to tell me to stop what I'm doing.

```
panic: trying to make exported function async: (StructName).FunctionName$invoke

goroutine 1 [running]:
github.com/tinygo-org/tinygo/compiler.(*Compiler).markAsyncFunctions(0xc0001421e0, 0xc000000180, 0xc000036000, 0xc0064ff0d0)
	/home/circleci/project/compiler/goroutine-lowering.go:852 +0x5ae6
github.com/tinygo-org/tinygo/compiler.(*Compiler).lowerCoroutines(0xc0001421e0, 0xa9bf6b, 0x7fc8e8090a90)
	/home/circleci/project/compiler/goroutine-lowering.go:180 +0x40
github.com/tinygo-org/tinygo/compiler.(*Compiler).LowerGoroutines(0xc0001421e0, 0x7fc8e8091580, 0x7fc8e808f360)
	/home/circleci/project/compiler/goroutine-lowering.go:131 +0x6d
github.com/tinygo-org/tinygo/compiler.(*Compiler).Optimize(0xc0001421e0, 0x2, 0x2, 0x5, 0x0, 0x0)
	/home/circleci/project/compiler/optimizer.go:83 +0x5b7
main.Compile(0x7ffdf414c21f, 0x10, 0x7ffdf414c205, 0x19, 0xc000142000, 0xc0064ffe98, 0xc0064ffc68, 0x0, 0x0)
	/home/circleci/project/main.go:188 +0x370b
main.Build(0x7ffdf414c21f, 0x10, 0x7ffdf414c205, 0x19, 0x7ffdf414c1fd, 0x4, 0xc0000e7e98, 0xc0000ee250, 0x8cd78e)
	/home/circleci/project/main.go:343 +0xe2
main.main()
	/home/circleci/project/main.go:836 +0x128d
```

jadr2ddude:
This appears to just be #514 .

ellemlabs:
Do you suppose https://github.com/tinygo-org/tinygo/pull/736/commits/0e036d4fa9c88ab63f2ecb1a7746ede23af4f065 will fix it?

[...]

PR #1430, opened 2020-10-07
interp: rewrite entire package 

aykevl:
For a full explanation, see interp/README.md. In short, this rewrite is a redesign of the partial evaluator which improves it over the previous partial evaluator. The main functional difference is that when interpreting a function, the interpretation can be rolled back when an unsupported instruction is encountered (for example, an actual unknown instruction or a branch on a value that's only known at runtime). This also means that it is no longer necessary to scan functions to see whether they can be interpreted: instead, this package now just tries to interpret it and reverts when it can't go further.

This new design has several benefits:

  * Most errors coming from the interp package are avoided, as it can simply skip the code it can't handle. This has long been an issue.
  * The memory model has been improved, which means some packages now pass all tests that previously didn't pass them.
  * Because of a better design, it is in fact a bit faster than the previous version.

This means the following packages now pass tests with `tinygo test`:

  * hash/adler32: previously it would hang in an infinite loop
  * math/cmplx: previously it resulted in errors

---

This PR depends on #1412, which should be merged first. Also, it requires a change to the go-llvm package.


deadprogram:
#1412 has been merged. Not sure what change is needed to go-llvm or if that has already been made.

aykevl:
No, a different change is needed. I'll get back to this, it really helps to make TinyGo a more correct Go compiler.

aykevl:
Looks like this _almost_ works, but needs some work to support some Go versions.

aykevl:
Note that this should probably be merged after the next release, to give it some time to find possible bugs.

deadprogram:
Indeed :smiley_cat:

aykevl:
Updated the PR, I think it is ready now. This newer version fixes an issue with the `append` builtin and (in a separate commit) fixes a bug with an assertion in the coroutine lowering pass (https://github.com/tinygo-org/tinygo/issues/1409 and https://github.com/tinygo-org/tinygo/issues/733).

ofauchon:
With tinygo (dev branch), I could not compile my application: 

```
traceback:
/usr/lib/go/src/bytes/bytes.go:
  br i1 %14, label %for.body, label %for.done, !dbg !2460
/home/olivier/go/src/github.com/jacobsa/crypto/cmac/subkey.go:29:27:
  %9 = call { i8*, i32, i32 } @bytes.Repeat(i8* %6, i32 %7, i32 %8, i32 16, i8* undef, i8* undef), !dbg !2457
github.com/jacobsa/crypto/cmac/<init>:
  call void @"github.com/jacobsa/crypto/cmac.init#1"(i8* undef, i8* undef), !dbg !2452
```

Then I applied the 3 commits 

```
git cherry-pick b3794a9ea8af6b7c71112a15ad54470db060c66c
git cherry-pick 54ce74a784e05ac6c7d009aa840c802cf66fa1b6
git cherry-pick 0b3b692d533630e306bc815ac3387073bd954f8c
go install
```

This fixed the previous error ! 

Thanks 

Olivier

interp/interp.go
```diff
@@ -1,59 +1,73 @@
...
-	name := "runtime.initAll"
-	e := &Eval{
-		Mod:          mod,
-		TargetData:   llvm.NewTargetData(mod.DataLayout()),
-		Debug:        debug,
-		dirtyGlobals: map[llvm.Value]struct{}{},
+	r := runner{
+		mod:           mod,
+		targetData:    llvm.NewTargetData(mod.DataLayout()),
+		debug:         debug,
```
niaow, line 39: Eventually we should standardize this type of functionality throughout the compiler, but this makes sense for now.

interp/memory.go
```diff
@@ -0,0 +1,1392 @@
...
+		obj := mv.get(objectIndex)
+		if obj.marked < mark {
+			obj = obj.clone()
+			obj.marked = mark
+			if mv.objects == nil {
+				mv.objects = make(map[uint32]object)
+			}
+			mv.objects[objectIndex] = obj
+		}
+		// TODO: also mark objects referenced by this object.
```
niaow: Is this going to break anything now?

interp/memory.go
```diff
@@ -0,0 +1,1392 @@
...
+func (v literalValue) asPointer(r *runner) (pointerValue, error) {
+	return pointerValue{}, errLiteralToPointer
+}
+
+func (v literalValue) asRawValue(r *runner) rawValue {
+	var buf []byte
+	switch value := v.value.(type) {
+	case uint64:
+		buf = make([]byte, 8)
+		binary.LittleEndian.PutUint64(buf, value)
```
niaow, line 359: I don't think we have to deal with any big-endian machines right now, but we should probably document this limitation somewhere.

niaow (changes requested):
Looks good, one minor comment

interp/memory.go
aykevl: Not sure whether this is necessary. If we ever add support for a big endian system (if they are still around), grepping for `LittleEndian` seems pretty obvious to me. Right now there is at least one other place that assumes little endian systems: builder/build.go (although it's perhaps less important as each architecture needs special code to be supported there).

interp/memory.go
aykevl: In my testing, no. However it is technically unsound and should probably be fixed.

deadprogram: So, does this require additional work now, or should we merge and deal with in future effort?

interp/memory.go
deadprogram: Makes sense to me!

aykevl: I will document this at the top of the file just to be sure.

interp/memory.go
aykevl: Yes, I think I shouldn't be lazy and fix this (even if it doesn't seem to affect the behavior right now).

aykevl:
Okay, I've written the missing bits (see the force push for a diff). Ready for review again.

interp/memory.go
aykevl: Fixed now.

deadprogram:
What is the status of this PR? It would appear that all feedback has now been addressed.

aykevl:
Ready from my POV but maybe @niaow wants to comment.

niaow (approved):

deadprogram:
Here we go!! :rocket: :raccoon:

aykevl:
Unfortunately this appears to have broken at least support for the Circuit Playground Express. Even examples/blinky1 doesn't work anymore. ~It appears to be related to `machine.RingBuffer` (the `Put` and `Used` methods).~ I haven't investigated further yet.

deadprogram:
OK that seems odd, because the TinyHCI tests passed on CPE board after the merge. 

However I am able right now to verify the problem on CPE myself. Also confirm it is commit 30df912565c2ab2d69efb40a2b2859a1736a88a3 that causes the problem.

aykevl:
I'll take a look, but probably not before Saturday.

deadprogram:
I wonder if it is specifically related to USB-CDC code somehow? 

I notice that now that this merged into `dev` in subsequent PRs that most of the ARM boards are failing, except for Microbit which along with RISC-V and AVR boards are still passing TinyHCI:
https://github.com/tinygo-org/tinygo/pull/1543/checks?check_run_id=1601942130

aykevl:
Yes, probably. I will investigate soon.

aykevl:
What I've discovered so far is that `-opt=1` seems to mask the issue. Not sure why or how yet, don't rely on it.

aykevl:
I have figured out the issue, the fix is in #1547.

PR #1431, opened 2020-10-08
wasm: allow users to provide custom sections

mathetake:
This PR adds `wasm-custom-sections` flag for specifying WASM custom sections. Ref: https://reviews.llvm.org/D43097

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -86,6 +86,10 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 		}
 	}
 
+	if len(config.Options.WasmCustomSections) > 0 {
+		transform.AddWASMCustomSections(mod, config.Options.WasmCustomSections)
+	}
+
 	// Optimization levels here are roughly the same as Clang, but probably not
 	// exactly.
 	errs = nil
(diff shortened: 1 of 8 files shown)


aykevl:
Can you give some context, and explain why this needs to be in TinyGo?
Alternatively, can you add those custom sections later to the generated WebAssembly file? I would assume there are tools that can do this.

mathetake:
> Alternatively, can you add those custom sections later to the generated WebAssembly file? I would assume there are tools that can do this.

Yeah, that's true. closing.

Mossaka:
> Can you give some context, and explain why this needs to be in TinyGo?

There is a need to add custom sections for "processed-by" field that describes the producer of the wasm module. See a discussion [here](https://github.com/bytecodealliance/wit-bindgen/issues/617).

aykevl:
Is it possible to add these custom sections in C? If so, would it be possible to use CGo instead?
(Note that while CGo support in TinyGo is not as completely supported as upstream Go, it does not depend on external tools so it's much easier to use).

Mossaka:
> Is it possible to add these custom sections in C?

Not sure, I will have to try it out

PR #1432, opened 2020-10-09
Add instanceof support for WebAssembly

elct9620:
The `instanceOf` is commented out in the `wasm_exec.js`, but it can simply be implemented in this case.

I think using `sp` and `sp + 8` and return bool can work correctly, and also confirmed in my local machine.

### DIFF
--- a/targets/wasm_exec.js
+++ b/targets/wasm_exec.js
@@ -407,9 +407,9 @@
 					},
 
 					// func valueInstanceOf(v ref, t ref) bool
-					//"syscall/js.valueInstanceOf": (sp) => {
-					//	mem().setUint8(sp + 24, loadValue(sp + 8) instanceof loadValue(sp + 16));
-					//},
+					"syscall/js.valueInstanceOf": (v_addr, t_addr) => {
+ 						return loadValue(v_attr) instanceof loadValue(t_addr);
+					},
 
 					// func copyBytesToGo(dst []byte, src ref) (int, bool)
 					"syscall/js.copyBytesToGo": (ret_addr, dest_addr, dest_len, dest_cap, source_addr) => {


deadprogram:
@aykevl can you comment on this please?

targets/wasm_exec.js
```diff
@@ -407,9 +407,9 @@
 					},
 
 					// func valueInstanceOf(v ref, t ref) bool
-					//"syscall/js.valueInstanceOf": (sp) => {
-					//	mem().setUint8(sp + 24, loadValue(sp + 8) instanceof loadValue(sp + 16));
-					//},
+					"syscall/js.valueInstanceOf": (sp) => {
+ 						return loadValue(sp) instanceof loadValue(sp + 8);
+					},
```
aykevl: I do not think this is correct. Have you tested the code?

The calling convention is different in TinyGo, so it should probably be something like this:

```suggestion
					"syscall/js.valueInstanceOf": (vPtr, tPtr) => {
 						return loadValue(vPtr) instanceof loadValue(tPtr);
					},
```

Although maybe this was already the layout in memory by chance.

aykevl (changes requested):

targets/wasm_exec.js
elct9620: Yes, I have tested the code. 
But the second argument may have value and I didn't check for it. The offset in TinyGo is `8` unit, but we should use the second argument instead of offsetting it by manual.

elct9620: I will update it later.

elct9620:
Update implement to use `loadValue(v_attr)` and `loadValue(t_attr)` to retrieve JavaScript references to process `instanceOf`

elct9620:
The CircleCI have some error, but I cannot see it 🤔

aykevl:
The CircleCI error is a bug in Go 1.15.3, which we're working around using #1443. Can you rebase this branch on top of the `dev` branch? That should fix it.

deadprogram:
Thank you very much for this contribution @elct9620 now merging.

PR #1434, opened 2020-10-13
nintendoswitch: refactor and hblauncher environment parsing

racerxdl:
*	Added homebrew launcher parser (for overriden heap)
*	Removed unused stuff (moved to gonx)
*	Kept require code at minimum to work in a real device

This makes possible to gonx run in a real device: https://twitter.com/lucasteske/status/1312667426283192320

There might be still some adds in the future for:

* Override stdout/stderr/stdin externally
* Thread Management (for goroutines work)
* Hardware Syncronization Mechanisms.

But so far that makes possible to develop everything in gonx. I'm still polishing the gonx code but I'm working on this branch: https://github.com/racerxdl/gonx/tree/dev-ipc


src/runtime/runtime_nintendoswitch.go
```diff
@@ -8,24 +8,31 @@ type timeUnit int64
...
 //export main
-func main() int {
+func main() {
 	preinit()
-
-	// Obtain the initial stack pointer right before calling the run() function.
-	// The run function has been moved to a separate (non-inlined) function so
-	// that the correct stack pointer is read.
-	stackTop = getCurrentStackPointer()
 	runMain()
```
aykevl: If you don't use `getCurrentStackPointer`, you could avoid the `runMain` indirection and directly call `run`.

src/runtime/runtime_nintendoswitch.s
```diff
@@ -1,45 +1,51 @@
...
-.section .text.sleepThread, "ax", %progbits
-.global sleepThread
-.type sleepThread, %function
-.align 2
-sleepThread:
-	svc 0xB
-	ret
+FUNC runtime.getCurrentStackPointer
+    mov x0, sp
+    ret
```
aykevl: This function is not necessary anymore since https://github.com/tinygo-org/tinygo/pull/1404.

src/runtime/runtime_nintendoswitch_init.go
```diff
@@ -0,0 +1,138 @@
...
+
+// GetContextPtr returns the hblauncher context
+func GetContextPtr() uintptr {
+	return context
+}
+
+// GetMainThreadHandle returns the main thread handler if any
+func GetMainThreadHandle() uintptr {
+	return mainThread
+}
```
aykevl: I suppose these are necessary for gonx?
I do not think this should be exposed in the public API of the runtime. If they are really necessary, maybe it would be a better idea to make them unexported and use them from gonx with `//go:linkname`. This makes it clear this is not a regular API but an implementation detail of the TinyGo runtime.

deadprogram:
Hi @racerxdl any chance of getting these feedback items addressed for the pending release? :smile_cat:

racerxdl:
Sure @deadprogram ! I will take a look on these tonight! ^^

src/runtime/runtime_nintendoswitch_init.go
racerxdl: Great! I didn't knew I could do that with go:linkname

deadprogram:
@aykevl please check out the changes made here. Thanks.

targets/nintendoswitch.json
```diff
@@ -9,9 +9,11 @@
   "libc": "picolibc",
   "gc": "conservative",
   "relocation-model": "pic",
+  "features": ["+a57","+strict-align","+crc","+crypto"],
   "cflags": [
-    "-target", "aarch64-none-linux-gnu",
+    "-target", "aarch64-unknown-none",
     "-mtune=cortex-a57",
+    "-mcpu=cortex-a57",
```
aykevl: The `-mcpu` flag should match the `cpu` field in the JSON file, see targets/esp32.json for example. I think that can replace the `features` field entirely.

Also, I think `-mtune` is unnecessary in this case, as it should tune for the processor specified in `-mcpu`.

aykevl (commented):
Looks reasonable to me.

Going forward, I would actually prefer if all runtime files for one target (in this case, nintendoswitch) would be kept in a single file to reduce the number of files in the runtime package (which is already quite large).

racerxdl:
Done @aykevl ! Moved everything to a single file (tried to organize a bit, but sadly the init routine is big). And tuned the .json file :D

deadprogram:
Looks ready to go. Any last feedback before merge @aykevl ?

aykevl (approved):
Looks good to me!

PR #1437, opened 2020-10-13
targets/wasi: remove --no-threads flag as no longer in LLVM 11 linker

deadprogram:
This PR removes the `--no-threads` flag from the wasi.json target file, as that flag as no longer in the LLVM 11 linker.


deadprogram:
Needed to restore the `dev` branch CI to a passing state.

Issue #1396, opened 2020-09-18
(ESP8266) Pin.Get() undefined


alankrantas:
For ESP8266 type machine.Pin has no field or method Get.

Also, should ESP8266 and ESP32 define machine.LED (Pin(2) on most boards) as well?

aykevl:
> For ESP8266 type machine.Pin has no field or method Get.

Yes, this is not yet implemented. Feel free to submit a patch :)

> Also, should ESP8266 and ESP32 define machine.LED (Pin(2) on most boards) as well?

`machine.LED` is defined for the esp32-wroom-32 as it has an on-board LED. It is not defined for the esp8266 target as it is not guaranteed to have this LED (even though almost all development boards have it). So if we add support for a board (and not just the chip itself), we can define `machine.LED`. In other words, this is working as intended.

jbpratt:
I think this can be closed!

alankrantas:
Just tested it, works in both PinInput and PinOutput mode. Thanks everyone!

PR #1438, opened 2020-10-13
WIP: Esp8266 Get Function

vertoforce:
I really am not positive about what I'm doing, but the more I get my feet wet the more I can help :)

For anyone looking to make similar changes, it looks like the registers are defined [here](https://github.com/cnlohr/nosdk8266/blob/f189866d1ecbcdb7a7fe71be3c200458cd257b22/useful/esp8266.h).

You can see the esp8266 technical specs [here](https://www.espressif.com/sites/default/files/documentation/esp8266-technical_reference_en.pdf)

Fixes https://github.com/tinygo-org/tinygo/issues/1396

### DIFF
--- a/src/machine/machine_esp8266.go
+++ b/src/machine/machine_esp8266.go
@@ -110,11 +110,10 @@ func (p Pin) Configure(config PinConfig) {
 // Get returns the current value of a GPIO pin when the pin is configured as an
 // input.
 func (p Pin) Get() bool {
-	if p < 32 {
-		return esp.GPIO.IN.Get()&(1<<p) != 0
-	} else {
-		return esp.GPIO.IN1.Get()&(1<<(p-32)) != 0
-	}
+	// See this document for details
+	// https://www.espressif.com/sites/default/files/documentation/esp8266-technical_reference_en.pdf
+
+	return esp.GPIO.IN.Get()&(1<<p) != 0
 }
 
 // Set sets the output value of this pin to high (true) or low (false).
--- a/src/machine/machine_esp8266.go
+++ b/src/machine/machine_esp8266.go
@@ -107,6 +107,16 @@ func (p Pin) Configure(config PinConfig) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


vertoforce:
Reading more in to the datasheets...

vertoforce:
I'm not sure why you are bit shifting the input pin in the setting of values (`1<<p`), but I removed it in the getting.  I might be wrong there.

deadprogram:
Hello @vertoforce thanks for contributing to TinyGo.

Regarding the GPIO input, this is info from https://www.espressif.com/sites/default/files/documentation/esp8266-technical_reference_en.pdf:

> 2.2.3. GPIO Input Register
> bit[15:0] the input status bit (readable and writable):
> If the related bit is set to be 1, the IO pin status is high level. If the related bit is set to be 0,
> the IO pin status is low level. Bit[15:0] contains 16 GPIO input status bits. 

This is a very common pattern on microcontrollers, which is the state of the bit in a particular position corresponds to that specific pin.

So, pin 0 is bit 0, pin 1 is bit 1, pin 2 is bit 2, etc.  This is why you frequently see (1 << pin) which refers to that bit position.

So you would probably want your code to be something like:

```go
return esp.GPIO.GPIO_IN.Get()&(1 << p) != 0
```

Hope that helps!

vertoforce:
Ah right, thanks @deadprogram! I had a brain-dead moment there forgetting what bitshifts really do.

Updated the PR :)

deadprogram:
Can you please run `go fmt` on your changes, @vertoforce that should allow it to pass the CI tests. Thanks.

vertoforce:
Whoops my bad, done!

deadprogram:
@vertoforce I think you are almost there. If you rebase your branch against the latest `dev` branch then the CircleCI tests should be able to pass.

vertoforce:
Sweet, did it.  Thanks for the help :)

deadprogram:
Looking good now @vertoforce I am going to squash/merge.

Thanks!

PR #1439, opened 2020-10-14
machine/esp32, targets/esp32: correct board definitions

deadprogram:
This PR correct board definitions for actual ESP32 boards not processor variants, also defines all labeled pins.


aykevl:
Looks reasonable, thank you!

The CI failure is likely a flaky test and certainly unrelated to this PR, so ignoring that.

PR #1440, opened 2020-10-14
all: add -output= flag for setting print/panic output

aykevl:
This flag allows setting the standard output to a few different values:

*   `-output=none`:   do not produce output. This is useful for some systems that lack a native output (e.g. Digispark) or for WebAssembly where the additional wasi output dependency may be a problem.
*   `-output=native`: use native output for the target, for example stdout on an OS or the native output of QEMU under emulation.
* `-output=uart`:   use UART output, the default for most microcontrollers.
*  `-output=custom`: allow setting a custom output with runtime.SetOutput. This provides flexibility for those who need it.

It is quite an invasive change, and it might break things. I have not tested this extensively.

In particular, this includes a change to the Teensy 3.6 to be more like other targets. It now configures the UART at startup and thus no longer needs to configure it on demand when needed. It also doesn't call `runtime.Gosched()` from the machine package anymore, greatly simplifying the code (this is better solved with #1172 anyway).

Even though there are some more calls, I compared code size before and after and it doesn't change, except for atmega where it is two bytes (one instruction) reduced for some reason. But examples/serial still works on my Uno so I don't think this is something to worry about.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -337,6 +337,10 @@ smoketest:
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=itsybitsy-nrf52840  examples/blinky1
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=teensy36            examples/blinky1
+	@$(MD5SUM) test.hex
+	$(TINYGO) build             -o test.nro -target=nintendoswitch      examples/serial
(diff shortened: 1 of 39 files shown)


aykevl:
This depends on #1441 which should fix CI for this PR.

aykevl:
https://github.com/tinygo-org/tinygo/pull/1922 has been merged, which is the same functionality, so closing.

PR #1443, opened 2020-10-17
ci: work around bug in Go 1.15.3

aykevl:
This works around a bug in Go 1.15.3 by pinning an older version.
See: https://github.com/golang/go/issues/42032


deadprogram:
:safety_pin: merging.

aykevl:
This should be fixed in Go 1.15.4: https://github.com/golang/go/issues/42151

PR #1445, opened 2020-10-17
Use stack based scheduler on Linux, MacOS and Nintendo Switch

aykevl:
These two PRs add support for the stack based scheduler on Linux, MacOS and the Nintendo Switch. This requires adding support for four new architectures, which becomes a lot easier after some refactoring to reduce the architecture-specific code for stack switching.

This brings these platforms closer to baremetal TinyGo and in fact to the standard Go implementation.

I have intentionally kept ARM support separate from Cortex-M support even if it's very similar. The reason is that there is a special "process stack pointer" which was specifically made for this purpose and is useful to avoid using the goroutine stack as interrupt stack (implemented in hardware). Non-Cortex-M chips don't have this or don't allow access to it, for example on Linux.

### DIFF
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -239,14 +239,16 @@ func defaultTarget(goos, goarch, triple string) (*TargetSpec, error) {
 	// No target spec available. Use the default one, useful on most systems
 	// with a regular OS.
 	spec := TargetSpec{
-		Triple:    triple,
-		GOOS:      goos,
-		GOARCH:    goarch,
-		BuildTags: []string{goos, goarch},
-		Linker:    "cc",
-		CFlags:    []string{"--target=" + triple},
-		GDB:       []string{"gdb"},
-		PortReset: "false",
+		Triple:           triple,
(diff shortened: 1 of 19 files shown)


deadprogram:
Seems like something weird went wrong with some of the tests?

aykevl:
Yes, I need to debug this. It looks like a memory corruption.

aykevl:
I have found why the tests fail: it's the extalloc GC.

This works: `tinygo test -gc=conservative text/scanner`

This does not:

```
$ tinygo test text/scanner
=== RUN   TestNext
--- PASS: TestNext
=== RUN   TestScan
```

(and many other errors that indicate a memory corruption)

aykevl:
This PR depends on #1860.

aykevl:
This is now ready for review.

niaow (approved):

deadprogram:
Thanks for working on this @aykevl and to @niaow for reviewing.

Now merging.

PR #1446, opened 2020-10-18
nrf: give more flexibility in picking SPI speeds

aykevl:
Instead of only allowing a limited number of speeds, use the provided
speed as an upper bound on the allowed speed. The reasoning is that
picking a higher speed than requrested will likely result in malfunction
while picking a lower speed will usually only result in slower
operation.
This behavior matches the ESP32 at least.

---

Tested on a SK9822 LED strip with an nrf52832.

It was inspired by code like this: https://github.com/tinygo-org/drivers/pull/208/files#diff-3ea9b9d98ece4f5cf1c374d3876eeda73d4b9fc617d1bc3431cd988af2bfeaebL41
Also, this behavior should allow more flexibility in configuring SPI peripherals: a driver or program should pick the maximum speed allowed by the driver instead of configuring a specific speed per chip (for portable code).

### DIFF
--- a/src/machine/machine_nrf.go
+++ b/src/machine/machine_nrf.go
@@ -357,23 +357,25 @@ func (spi SPI) Configure(config SPIConfig) {
 	// set frequency
 	var freq uint32
 
-	switch config.Frequency {
-	case 125000:
-		freq = nrf.SPI_FREQUENCY_FREQUENCY_K125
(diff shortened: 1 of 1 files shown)


deadprogram:
Very good idea @aykevl thanks. Now merging.

PR #1447, opened 2020-10-18
spi: switch default frequency to 4MHz

aykevl:
Let's use the same default frequency everywhere, for consistency.
It could be any frequency, but 4MHz is already used for other chips and
it seems like a reasonable frequency to me (not too fast for most chips
but still reasonably fast). Oh, and 4MHz is slow enough that it can be
inspected by a Saleae Logic 4 (that sadly has been discontinued).

Related to #1446 which also changes the default from 500kHz to 4MHz.


deadprogram:
Totally in favor of this. Now merging.

PR #1451, opened 2020-10-20
nrf: fix nrf52832 flash size

aykevl:
I've accidentally specified just half of the available flash in the
linker script. This change fixes that.

There is in fact a 256kB version of the nrf52832, but it also has 32kB
of RAM so if you had used that it wouldn't actually work right now.
Also, extending the available flash should not affect existing programs
(as I haven't seen any run into size limitations yet).

---

Discovered while preparing for GoLab.


deadprogram:
:arrow_double_up: nice. Now merging.

Issue #1175, opened 2020-06-18
USB-CDC times out


sago35:
Sometimes bytes from fmt.Printf () disappear.

```
temperature: 70.699997
temperature: 70.699997
temperature: 70.699997
tmperature: 70.699997
temperature: 70.699997
temperature: 70.699997
temperature: 70.587494
temperature: 70.587494
temperature: 70.925003
temperature: 70.925003
tmperature: 70.925003
temperature: 72.274994
temperature: 72.274994
temperature: 73.625000
```

The reason is that there is no retry after a timeout.
However, increasing the number of busy loops will have an impact on performance.

```go
// src/machine/machine_atsamd51.go
func (usbcdc USBCDC) WriteByte(c byte) error {
    if usbLineInfo.lineState > 0 {
        // set the data
        // ...

        // wait for transfer to complete
        timeout := 3000
        for (getEPINTFLAG(usb_CDC_ENDPOINT_IN) & sam.USB_DEVICE_ENDPOINT_EPINTFLAG_TRCPT1) == 0 {
            timeout--
            if timeout == 0 {
                return errors.New("USBCDC write byte timeout")
            }
        }
    }

    return nil
}


// src/machine/usb.go
func (usbcdc USBCDC) Write(data []byte) (n int, err error) {
    for _, v := range data {
        usbcdc.WriteByte(v)
        // !!! If err! = nil, go to the next byte !!!
    }
    return len(data), nil
}
```

sago35:
Sending some batches instead of using putchar will be faster.
If you want to send them in batches, you may want to make the timeout value longer.

By sending a certain number of bytes at once on the TinyGo side, I think the performance will improve considerably.

### Before Change

https://github.com/sago35/tinygo/tree/b0eb3ff5362f7c5e80b23407ef6f7531cb8ab8ff

It takes 3.1ms to output 51 characters.
49 characters in between A and 2 characters in between B.
It is slow because it waits for the completion of transmission for each character.
And sometimes it times out.

![image](https://user-images.githubusercontent.com/9251039/85401577-38727080-b595-11ea-9625-64fd48976f7c.png)

![image](https://user-images.githubusercontent.com/9251039/85402112-16c5b900-b596-11ea-88f9-9aafab6a50bd.png)


### After the change

https://github.com/sago35/tinygo/tree/e5e8ac99d126af0fa5050601c10f1bf113b10d46

It takes 170us to output 51 characters.
It is about 20 times faster.

[...]

PR #1452, opened 2020-10-20
atsamd5x: improve USBCDC

sago35:
Increased speed by increasing the size of the data to be sent at a time.
I also made it so that it waits more than 1ms before timeout, so it probably won't time out.

Right now we're using the Get() process from a RingBuffer, but we can optimize it further and send the data from the RingBuffer directly.
However, the code is complicated by the fact that the ADDR register needs to be a 4 byte align.

When the PR is merged, #1175 in atsamd5x is solved.

This PR is designed to flush() every 1ms.

### DIFF
--- a/src/machine/machine_atsamd51.go
+++ b/src/machine/machine_atsamd51.go
@@ -902,7 +902,10 @@ type UART struct {
 
 var (
 	// UART0 is actually a USB CDC interface.
-	UART0 = USBCDC{Buffer: NewRingBuffer()}
+	UART0 = USBCDC{
+		Buffer:   NewRingBuffer(),
+		TxBuffer: NewRingBuffer(),
+	}
 )
 
 const (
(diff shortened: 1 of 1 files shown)


sago35:
Since up to 128 bytes can be sent in a single pass, writing up to 128 bytes is very fast.
It is not very efficient when split into multiple Flush()'s, as there is a gap between Flush() and Flush().
However, it is much faster than the original implementation.

We can make it even faster by using the send completion interrupt.


* 10242 byte
  * old : 139.580 ms :  73377 B/s
  * new :  80.004 ms : 128018 B/s
* 256 byte
  * old : 3.509 ms :  72955 B/s
  * new : 1.235 ms : 207287 B/s
* 128 byte
  * old : 1.711 ms :  74810 B/s
  * new : 0.154 ms : 831168 B/s
* 14 byte
  * old : 184.75  us :  75778 B/s
  * new :  17.41  us : 804135 B/s


testcode:

```
testdata := [10242]byte{}

machine.LED.High()
machine.UART0.Write(testdata[:])
machine.LED.Low()
```

sago35:
You can increase the size of the RingBuffer by setting the following head and tail to volatile.Register16
If the size is increased, the amount of data that can be sent in one flush() will be larger and the transfer efficiency will improve.



```go
// src/machine/buffer.go
type RingBuffer struct {
	rxbuffer [bufferSize]volatile.Register8
	head     volatile.Register8
	tail     volatile.Register8
}
```

If you want to make it larger than 128, you must also change the following
Because the size of udd_ep_in_cache_buffer is 128.

```go
usbEndpointDescriptors[usb_CDC_ENDPOINT_IN].DeviceDescBank[1].ADDR.Set(uint32(uintptr(unsafe.Pointer(&udd_ep_in_cache_buffer[usb_CDC_ENDPOINT_IN]))))
```

deadprogram:
@sago35 it appears you have implemented the interrupt solution:

```go
if i == usb_CDC_ENDPOINT_IN {
	UART0.waitTxc = false
}
```
However it appears that your proposed `udd_ep_in_cache_buffer` size increase is not in this PR.

Is that correct?

deadprogram:
Also, it would seem that a similar change could be made to the SAMD21 support, is that true @sago35 ?

sago35:
@deadprogram 

> However it appears that your proposed udd_ep_in_cache_buffer size increase is not in this PR.

I'll do a `git push` later.

> Also, it would seem that a similar change could be made to the SAMD21 support, is that true

Yes.
SAMD21 can be modified in the same way.

but,,,
There's a race condition that needs to be resolved for now.
I'd like to use `interrupt.Disable()`, but it's not currently available in the machine package.

aykevl:
I'm not sure I fully understand this code, but I don't see anything weird with it.

Sidenote: this gives me even more reason that USB should be separated out in a new package. I think the code would be easier to follow if the code that sends a buffer over USB and the code that buffers for the next transmission is cleanly separated with an API.

deadprogram:
@sago35 what is the specific race condition you are seeing?

sago35:
> what is the specific race condition you are seeing?

Calling usbcdc.TxBuffer.Put(c) with multiple priorities will cause problems.

interrupt.Disable() can be built and used with interp-rewrite branch.
I believe #1430 will be merged with tinygo 0.17.0, so this PR will also be targeted at 0.17.0.

> Note that this should probably be merged after the next release, to give it some time to find possible bugs.
> https://github.com/tinygo-org/tinygo/pull/1430#issuecomment-718715299

src/machine/machine_atsamd51.go
```diff
@@ -1708,33 +1711,55 @@ func (pwm PWM) getMux() PinMode {
...
+				return nil
+			}
+			usbcdc.waitTxc = true
 
-		// clean multi packet size of bytes already sent
-		usbEndpointDescriptors[usb_CDC_ENDPOINT_IN].DeviceDescBank[1].PCKSIZE.ClearBits(usb_DEVICE_PCKSIZE_MULTI_PACKET_SIZE_Mask << usb_DEVICE_PCKSIZE_MULTI_PACKET_SIZE_Pos)
+			// set the data
+			for i := uint8(0); i < sz; i++ {
+				udd_ep_in_cache_buffer[usb_CDC_ENDPOINT_IN][i], _ = usbcdc.TxBuffer.Get()
+			}
```
sago35, line 1734: This process is a byte copy, but it's a waste of time.
I will create a new PR that improves on this part of the process.

src/machine/machine_atsamd51.go
```diff
@@ -1708,33 +1711,55 @@ func (pwm PWM) getMux() PinMode {
...
+
+// WriteByte writes a byte of data to the USB CDC interface.
+func (usbcdc USBCDC) WriteByte(c byte) error {
+	// Supposedly to handle problem with Windows USB serial ports?
+	if usbLineInfo.lineState > 0 {
+		// `timeout == 10000` waits for a bit over 1ms (120MHz)
+		timeout := 10000
+		ok := false
+		for !ok {
+			ok = usbcdc.TxBuffer.Put(c)
```
sago35, line 1762: RingBuffer.Put() does not consider calling from multiple priorities.
interrupt.Disable() is required.

deadprogram:
Now that #1481 has been merged, I will close this PR. Thanks again for the great work @sago35

PR #1456, opened 2020-10-22
all: replace underscores with dashes in target names

aykevl:
This is the convention, so use it everywhere.

Motivated by: https://github.com/tinygo-org/tinygo-site/pull/122


ardnew:
I think this is an invalid change that will not work with board-specific source files:

1. When implementing STM32F405 Feather, it was a real head-scratcher as to why my Go source files were not being found by the compiler when I used `feather-stm32f405` build tag (in target JSON file as well as source file build constraint).

2. Turned out the build constraints were never being satisfied. After changing `feather-stm32f405` to `feather_stm32f405` (JSON + source), the candidate source files were then being compiled.

Have you tested this change with a source that actually uses one of these tags (e.g. `// +build clue-alpha`) and **not** just one of its parent/catch-all tags (e.g., `// +build nrf`) ?

ardnew:
From `go help buildconstraint` output:

```
A build constraint is evaluated as the OR of space-separated options.
Each option evaluates as the AND of its comma-separated terms.
Each term consists of letters, digits, underscores, and dots.      <--- HERE
A term may be negated with a preceding !.
```

It doesn't seem hyphens are permitted in a build constraint term

aykevl:
Oops, I need to read more carefully. I meant to replace `-target=clue_alpha` with `-target=clue-alpha`, but accidentally also replaced `-tags=clue_alpha` (which is already correct). I updated the PR to fix this.

I think dashes look better than underscores, but sadly they can't be used in build tags.

aykevl:
Apparently I had made another mistake with `-tags=circuitplay-express`. Fixed now.

And I discovered that I probably made a mistake in 8333c171f4a2ec8570a93587e61e1adf65ec04bc which I've also fixed.

deadprogram:
Seems like there is still some problem with `itsybitsy-m4` and `arduino-nano33` as after flashing, neither one responds to USB CDC.

aykevl:
It seems very unlikely to me that these failures are caused by this change.

deadprogram:
Seems like some other PR is getting built and causing the problem I mentioned, but you are quite correct that it is not this one.

Now merging, thanks for simplifying this @aykevl

PR #1457, opened 2020-10-22
Add os.LookupEnv() stub

arp242:
os.Getenv() was already stubbed out, but os.LookupEnv() wasn't. This
will allow me to compile my program unmodified without using separate
files and build tags.


aykevl:
Thank you, looks reasonable!

There was a CI failure but it must have been unrelated to this PR.

PR #1458, opened 2020-10-23
board/teensy40: Add ADC support

ardnew:
### This PR is baselined on top of the Teensy 4.0 external interrupt support PR (#1484)

This PR adds analog input support (ADC1/ADC2) for Teensy 4.0 on 14 GPIO pins (`A0`-`A13`). 

I mirrored the API used in machine `feather-m4`, assuming it is probably a good example for TinyGo convention.

This has been tested using a cheap potentiometer with `src/examples/adc/adc.go`. Note you must also change the pin from `ADC0` to one of this machine's analog pins, e.g. `A0`. All other code can be left as-is (although the call to `machine.InitADC()` is unnecessary - it is implemented as an empty stub for backwards compatibility with existing code such as this example).

The following describes each analog pin, its equivalent digital pin, its IOMUX pad, and finally which ADC channel(s) it is connected to. It is defined in `src/machine/board_teensy40.go`:

```go
// Analog pins
const (
	//  = Pin  // Dig | [Pad]      {ADC1/ADC2}
	A0  = PA18 // D14 | [AD_B1_02] {  7 / 7  }
	A1  = PA19 // D15 | [AD_B1_03] {  8 / 8  }
	A2  = PA23 // D16 | [AD_B1_07] { 12 / 12 }
	A3  = PA22 // D17 | [AD_B1_06] { 11 / 11 }
	A4  = PA17 // D18 | [AD_B1_01] {  6 / 6  }
	A5  = PA16 // D19 | [AD_B1_00] {  5 / 5  }
	A6  = PA26 // D20 | [AD_B1_10] { 15 / 15 }
	A7  = PA27 // D21 | [AD_B1_11] {  0 / 0  }
	A8  = PA24 // D22 | [AD_B1_08] { 13 / 13 }
	A9  = PA25 // D23 | [AD_B1_09] { 14 / 14 }
	A10 = PA12 // D24 | [AD_B0_12] {  1 / -  }
	A11 = PA13 // D25 | [AD_B0_13] {  2 / -  }
	A12 = PA30 // D26 | [AD_B1_14] {  - / 3  }
	A13 = PA31 // D27 | [AD_B1_15] {  - / 4  }
)
```

### DIFF
--- a/src/machine/machine_mimxrt1062.go
+++ b/src/machine/machine_mimxrt1062.go
@@ -7,6 +7,7 @@ import (
 	"math/bits"
 	"runtime/interrupt"
 	"runtime/volatile"
+	"sync"
 )
 
 // Peripheral abstraction layer for the MIMXRT1062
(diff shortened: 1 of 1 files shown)


deadprogram:

ardnew:
rebased against `dev`

src/machine/machine_mimxrt1062.go
```diff
@@ -879,3 +880,190 @@ func (p Pin) getMuxMode(config PinConfig) uint32 {
...
+
+	// begin calibration
+	nxp.ADC1.GC.Set(average | nxp.ADC_GC_CAL)
+	nxp.ADC2.GC.Set(average | nxp.ADC_GC_CAL)
+
+	adcStatusGlobal.startCalibrating()
+	adcStatusGlobal.waitForCalibration()
+}
+
+func (a ADC) Get() uint16 {
```
aykevl, line 926: This is missing documentation.
Also, what is the output value? In other versions it is scaled to the full 16 bit width, so for example if the output has a resolution of 12 bits it's shifted left by 4 before returning.

src/machine/machine_mimxrt1062.go
```diff
@@ -879,3 +880,190 @@ func (p Pin) getMuxMode(config PinConfig) uint32 {
...
+	Resolution uint32
+	Samples    uint32
+}
+
+// InitADC is not used by this machine. Use `(ADC).Configure()`.
+func InitADC() {}
+
+// Configure initializes the receiver ADC's Pin and the ADC1/ADC2 instances for
+// immediate usage with default settings.
+func (a ADC) Configure() { a.ConfigureMode(ADCConfig{}) }
```
aykevl: I think it makes sense to change the API of `ADC.Configure` (for all chips) to the API you're using here. The reason is that an ADC actually has a number of important configurations for different tradeoffs:

 - Reference voltage (often VCC, but sometimes 1.1V or some other value).
 - Accuracy vs time. I know that the nrf chips have this tradeoff (where more accurate sampling takes a lot longer but is necessary if there is a big resistor on the input) and I would guess other ADCs have something similar.

Of course, it's not necessary to implement them all right away but with this I think it makes sense to add an `ADCConfig` to the peripheral.

src/machine/machine_mimxrt1062.go
```diff
@@ -879,3 +880,190 @@ func (p Pin) getMuxMode(config PinConfig) uint32 {
...
+}
+
+func (a ADC) Get() uint16 {
+	if ch1, ch2, ok := a.Pin.getChannel(); ok {
+		adcStatusGlobal.waitForCalibration()
+		if noChannel != ch1 {
+			nxp.ADC1.HC0.Set(uint32(ch1))
+			for !nxp.ADC1.HS.HasBits(nxp.ADC_HS_COCO0) {
+			}
+			return uint16(nxp.ADC1.R0.Get() & 0xFFFF)
```
aykevl: Strictly speaking the `& 0xFFFF` is unnecessary here but it doesn't do any harm either (and perhaps makes the code more obvious).

src/machine/machine_mimxrt1062.go
```diff
@@ -879,3 +880,190 @@ func (p Pin) getMuxMode(config PinConfig) uint32 {
...
+		average = nxp.ADC_GC_AVGE
+	}
+
+	return mode, average
+}
+
+const noChannel = uint8(0xFF)
+
+// getChannel returns the input channel for ADC1 & ADC2 of the receiver Pin p.
+func (p Pin) getChannel() (adc1, adc2 uint8, ok bool) {
```
aykevl: Because this is specifically for ADC, I think it would make sense to put that in the name (something like `getADCChannel`).

src/machine/machine_mimxrt1062.go
```diff
@@ -879,3 +880,190 @@ func (p Pin) getMuxMode(config PinConfig) uint32 {
...
+	switch p.getPort() {
+	case portA:
+		switch p.getPos() {
+		case 12: // [AD_B0_12]:       ADC1_IN1        ~
+			return 1, noChannel, true
```
aykevl: I think you could do this simpler, by directly matching on pin IDs:

```go
switch p {
case PA12:
    // ...
case PA13:
    // ...
// ...
}
```

The compiler will figure out the best way to lower such switches using jump tables and other mechanisms so that should not be a concern.

src/machine/machine_mimxrt1062.go
```diff
@@ -879,3 +880,190 @@ func (p Pin) getMuxMode(config PinConfig) uint32 {
...
+			for nxp.ADC1.GC.HasBits(nxp.ADC_GC_CAL) {
+			} // wait for calibration
+			w.Done()
+		}(&work)
+		go func(w *sync.WaitGroup) {
+			for nxp.ADC2.GC.HasBits(nxp.ADC_GC_CAL) {
+			} // wait for calibration
+			w.Done()
+		}(&work)
+		work.Wait()
```
aykevl: Why not keep it simple? Goroutines won't help here as they're busy-waiting anyway (and I'd like to avoid expensive operations such as goroutines in the machine package).

```suggestion
		for nxp.ADC1.GC.HasBits(nxp.ADC_GC_CAL) {
		} // wait for calibration
		for nxp.ADC2.GC.HasBits(nxp.ADC_GC_CAL) {
		} // wait for calibration
```

src/machine/machine_mimxrt1062.go
```diff
@@ -879,3 +880,190 @@ func (p Pin) getMuxMode(config PinConfig) uint32 {
...
+			return noChannel, 4, true
+		default:
+			return noChannel, noChannel, false
+		}
+	default:
+		return noChannel, noChannel, false
+	}
+}
+
+type adcStatus struct{ volatile.Register32 }
```
aykevl: Actually, why do you need to use volatile here?

src/machine/machine_mimxrt1062.go
ardnew: Hmm, it seems I have unintentionally lost a commit (during a previous rebase, presumably). 

However, I was actually _scaling_ the returned value uniformly to 16 bits [0, 65535], instead of just left-shifting.

For the 12-bit case, I don't think it matters much, as you'll just have a slight jump from 0 to 16 in the outputs, and a uniform distribution thereafter to 65535.

But for the 8-bit case, left-shifting seems like a bad idea, with a gap from 0 to 256. That seems more noticeable.

I don't believe performance between the two alternatives matters here, with scaling requiring just 1 integer multiply + 1 integer divide (if I remember correctly). The real bottleneck in ADC reads will be the conversions themselves. For example, with default settings (10-bit (requires 20 ADC clock cycles), 4 samples), each ADC read will require 80 ADC clock cycles just to perform the conversion.

I will add the docs to this function along with one of these 2 scaling techniques. Is it okay if I do the arithmetic scaling (not the bit-shifting)?

src/machine/machine_mimxrt1062.go
ardnew: Agreed. However, unless you are wanting to maintain backwards compatibility, I would think the ADC peripheral should be like all other peripherals and receive its `ADCConfig` as argument to `Configure`, and completely ditch the `ConfigureMode` that I'm using here.

src/machine/machine_mimxrt1062.go
ardnew: Will keep for clarity

src/machine/machine_mimxrt1062.go
ardnew: Renamed to `getADCChannel`

src/machine/machine_mimxrt1062.go
ardnew: Indeed, much simpler. Changed to switch on receiver `Pin p`

src/machine/machine_mimxrt1062.go
ardnew: Replaced goroutines with simpler spin locks as suggested

src/machine/machine_mimxrt1062.go
ardnew: Added godocs and 16-bit scaling to method `Get`

src/machine/machine_mimxrt1062.go
ardnew: `type adcStatus` has been removed entirely in favor of a simpler implementation. please see changes in latest commit.

src/machine/machine_mimxrt1062.go
ardnew: @aykevl would you rather I keep this code as is (i.e., keep it API-**compatible** with other `machine` package implementations)? 

or should I introduce a new API-**incompatible** `func (*ADC) Configure(ADCConfig)`, and remove my current `func (*ADC) ConfigureMode(ADCConfig)`, under the presumption that you are planning to change those others to match this new API-**incompatible** function

ardnew:
rebased against dev

src/machine/machine_mimxrt1062.go
aykevl: I think it would actually be better to (temporarily) break compatibility and meanwhile also update the signature of the other ADC implementations.

ardnew: updated to use the more conventional API: `func (*ADC) Configure(ADCConfig)`

ardnew:
rebased against dev (3eb9dca695254b9d59f30597ea6c409e1b65e7ee)

deadprogram:
Looks like this PR is ready to go @ardnew 

Do you want to squash commits together so I can merge, or would you prefer that I do so, or something else entirely?

deadprogram:
I am going to squash/merge in the interest of getting this unstuck (which is entirely my fault for not doing sooner).

Thanks @ardnew

PR #1463, opened 2020-10-27
main: update go.sum

aykevl:
This diff keeps reappearing so I ran `go mod tidy` and here is the result.


deadprogram:
Merging, thanks @aykevl

PR #1464, opened 2020-10-27
compiler: implement negate for complex numbers

aykevl:
This is necessary to complete #1430 (to get math/cmplx to work). Somehow negating complex numbers hasn't been necessary before, which says something about how often complex numbers are used in Go.


deadprogram:
:heavy_minus_sign: merging. Thanks @aykevl

PR #1465, opened 2020-10-28
testing: add Run method

aykevl:
This patch adds subtests via the Run function. This gets two more
packages to pass tests: encoding/base32 and hash/fnv.


deadprogram:
:three: :two: merging.

PR #1468, opened 2020-10-29
runtime: allow ranging over a nil map

aykevl:
This appears to be allowed by the specification, at least it is allowed by the main Go implementation: https://play.golang.org/p/S8jxAMytKDB

Allow it in TinyGo too, for consistency.

Found because it is triggered with `tinygo test flags`. This doesn't make the flags package pass all tests, but is a step closer.


deadprogram:
Seems legit, now merging. Thanks @aykevl

PR #1469, opened 2020-10-29
reflect: add support for DeepEqual

aykevl:
The implementation has been mostly copied from the Go reference implementation with some small changes to fit TinyGo.

Source: https://github.com/golang/go/blob/77a11c05d6a6f766c75f804ea9b8796f9a9f85a3/src/reflect/deepequal.go

This allows the unicode/utf16 package to pass tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -210,11 +210,13 @@ TEST_PACKAGES = \
 	index/suffixarray \
 	math \
 	math/cmplx \
+	net/mail \
+	reflect \
 	testing \
 	testing/iotest \
 	text/scanner \
-	text/scanner \
 	unicode \
+	unicode/utf16 \
 	unicode/utf8 \
 
 # Test known-working standard library packages.
--- a/src/reflect/all_test.go
+++ b/src/reflect/all_test.go
@@ -1,0 +1,188 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package reflect_test
+
+import (
+	"math"
+	. "reflect"
+	"testing"
+)
+
+type Basic struct {
+	x int
+	y float32
+}
+
+type NotBasic Basic
+
+type DeepEqualTest struct {
+	a, b interface{}
+	eq   bool
+}
+
+// Simple functions for DeepEqual tests.
+var (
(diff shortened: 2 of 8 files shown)


aykevl:
Unfortunately this causes a stack overflow on Cortex-M chips. Maybe the default stack size should be increased.

aykevl:
Rebased. Hopefully it works this time.

I've also fixed a bug in the reflect `v.Interface()` method in a separate commit.

aykevl:
Rebased and fixed some bugs. Now it appears to fail on `-target=hifive1-qemu` only.

Makefile
```diff
@@ -213,8 +213,8 @@ TEST_PACKAGES = \
 	testing \
```
dgryski, line 215: You should be able to add `net/mail` here too.

dgryski:
Can we import any of `DeepEqual` tests from Go's `reflect/all_test.go` file ?

Makefile
aykevl: Thanks! Will do.

aykevl:
> Can we import any of `DeepEqual` tests from Go's `reflect/all_test.go` file ?

Working on it, and in the process I found a bug that I'm now investigating.

aykevl:
So apparently the failure for hifive1-qemu is a stack overflow.

aykevl:
I simply upped the stack size of `-target=hifive1-qemu` which should fix the stack overflow. I think this is fine because the target isn't meant to be running on actual hardware anyway (and already has a different modification that makes it impossible to use on actual hardware).

dgryski (approved):
LGTM

dgryski:
Just noticed (with my testenv stubs branch: https://github.com/dgryski/tinygo/tree/dgryski/testenv-stubs ) which allows the `compress/flate` test to run, there is a segfault in `reflect.DeepEqual()` in the `TestWriterReset` test when run on my Mac.  The wasi version passes.

However, since a slightly-buggy DeepEqual() that exists is better than one that doesn't, this still gets a :+1: from my end.  We can figure out what's happening with the `compress/flate` test when the test stub code is merged and it's easier to debug.

Edit: n/m, looks like my working branch had an older version of this PR on it.  The updated version passes fine!

deadprogram:
Awesome! Thanks everyone, now merging.

PR #1454, opened 2020-10-21
board/teensy40: Add UART support

ardnew:
### This PR is baselined on top of the Teensy 4.0 external interrupt support PR (#1484)

This PR adds support for 7 (_seven_!) UART interfaces on the Teensy 4.0.

Testing was performed by connecting a serial USB adapter to digital pins 0 and 1, and flashing the `src/examples/serial` program to repeatedly print `hello world!` to my serial console. It has also been indirectly tested very frequently for debugging purposes while developing all subsequent feature PRs.

The public UART API is the same as most other TinyGo devices. The default UART (`UART1`) is configured by the runtime package during initialization, so `println()` and family will output (115200 baud) on UART TX digital pin 1 by default. 

Each UART is also allocated a unique interrupt handler `func (*UART).handleInterrupt(Interrupt))` during runtime initialization, which handles both transmit and receive events.

The following table (from `src/machine/board_teensy40.go`) succinctly describes all of the important properties of each of the 7 supported UART interfaces available in TinyGo:

```go
// #=====================================================#
// |                        UART                         |
// #===========#===========#=============#===============#
// | Interface | Hardware  | Clock(Freq) |  RX/TX  : Alt |
// #===========#===========#=============#=========-=====#
// |   UART1   |  LPUART6  | OSC(24 MHz) |  D0/D1  : 2/2 |
// |   UART2   |  LPUART4  | OSC(24 MHz) |  D7/D8  : 2/2 |
// |   UART3   |  LPUART2  | OSC(24 MHz) | D15/D14 : 2/2 |
// |   UART4   |  LPUART3  | OSC(24 MHz) | D16/D17 : 2/2 |
// |   UART5   |  LPUART8  | OSC(24 MHz) | D21/D20 : 2/2 |
// |   UART6   |  LPUART1  | OSC(24 MHz) | D25/D24 : 2/2 |
// |   UART7   |  LPUART7  | OSC(24 MHz) | D28/D29 : 2/2 |
// #===========#===========#=============#=========-=====#
```

PR #1471, opened 2020-10-31
board/teensy40: Add I2C support

ardnew:
### This PR is baselined on top of the Teensy 4.0 UART support PR (#1454)

This PR adds support for 3 (_three_!) I2C interfaces on the Teensy 4.0. 

The public I2C API is the same as most other TinyGo devices with no substantial deviations I'm aware of.

Testing this implementation has been done entirely through 2 new drivers I'm actively developing and are not yet available.

The following table (from `src/machine/board_teensy40.go`) succinctly describes all of the important properties of each of the 3 supported I2C interfaces available in TinyGo:

```go
// #====================================================#
// |                         I2C                        |
// #===========#==========#=============#===============#
// | Interface | Hardware | Clock(Freq) | SDA/SCL : Alt |
// #===========#==========#=============#=========-=====#
// |   I2C1    |  LPI2C1  | OSC(24 MHz) | D18/D19 : 3/3 |
// |   I2C2    |  LPI2C3  | OSC(24 MHz) | D17/D16 : 1/1 |
// |   I2C3    |  LPI2C4  | OSC(24 MHz) | D25/D24 : 0/0 |
// #===========#==========#=============#=========-=====#
```

### DIFF
--- a/src/machine/machine_mimxrt1062.go
+++ b/src/machine/machine_mimxrt1062.go
@@ -342,10 +342,10 @@ func (jt *pinJumpTable) dispatchInterrupt(interrupt.Interrupt) {
 		if status := gpio.ISR.Get() & gpio.IMR.Get(); status != 0 {
 			gpio.ISR.Set(status) // clear interrupt
 			for status != 0 {
-				p := Pin(bits.TrailingZeros32(status))
-				i := Pin(port + p)
-				jt.lut[i](i)
-				status &^= 1 << p
+				off := Pin(bits.TrailingZeros32(status)) // ctz
+				pin := Pin(port + off)
+				jt.lut[pin](pin)
+				status &^= 1 << off
 			}
 		}
 	}
--- a/src/machine/machine_mimxrt1062_i2c.go
+++ b/src/machine/machine_mimxrt1062_i2c.go
@@ -1,0 +1,725 @@
+// +build mimxrt1062
+
+package machine
+
+// I2C peripheral abstraction layer for the MIMXRT1062
+
+import (
+	"device/nxp"
+	"errors"
+)
+
+const (
+	TWI_FREQ_BUS     = 24000000        // LPI2C root clock is on 24 MHz OSC
+	TWI_FREQ_100KHZ  = 100000          // StandardMode (100 kHz)
+	TWI_FREQ_400KHZ  = 400000          // FastMode (400 kHz)
+	TWI_FREQ_1MHZ    = 1000000         // FastModePlus (1 MHz)
+	TWI_FREQ_5MHZ    = 5000000         // UltraFastMode (5 MHz)
+	TWI_FREQ_DEFAULT = TWI_FREQ_100KHZ // default to StandardMode (100 kHz)
+)
+
+var (
+	errI2CWriteTimeout       = errors.New("I2C timeout during write")
+	errI2CReadTimeout        = errors.New("I2C timeout during read")
+	errI2CBusReadyTimeout    = errors.New("I2C timeout on bus ready")
+	errI2CSignalStartTimeout = errors.New("I2C timeout on signal start")
+	errI2CSignalReadTimeout  = errors.New("I2C timeout on signal read")
+	errI2CSignalStopTimeout  = errors.New("I2C timeout on signal stop")
+	errI2CAckExpected        = errors.New("I2C error: expected ACK not NACK")
+	errI2CBusError           = errors.New("I2C bus error")
+	errI2CNotConfigured      = errors.New("I2C interface is not yet configured")
+)
+
+// I2CConfig is used to store config info for I2C.
+type I2CConfig struct {
+	Frequency uint32
+	SDA       Pin
+	SCL       Pin
+}
+
+func (c I2CConfig) getPins() (sda, scl Pin) {
+	if 0 == c.SDA && 0 == c.SCL {
+		// default pins if none specified
+		return I2C_SDA_PIN, I2C_SCL_PIN
+	}
+	return c.SDA, c.SCL
+}
+
+type I2C struct {
+	Bus *nxp.LPI2C_Type
+
+	// these hold the input selector ("daisy chain") values that select which pins
+	// are connected to the LPI2C device, and should be defined where the I2C
+	// instance is declared (e.g., in the board definition). see the godoc
+	// comments on type muxSelect for more details.
+	muxSDA, muxSCL muxSelect
+
+	// these are copied from I2CConfig, during (*I2C).Configure(I2CConfig), and
+	// should be considered read-only for internal reference (i.e., modifying them
+	// will have no desirable effect).
+	sda, scl  Pin
+	frequency uint32
+
+	// auxiliary state data used internally
+	configured bool
+}
+
+type I2CDirection bool
+
+const (
+	DirectionWrite I2CDirection = false
+	DirectionRead  I2CDirection = true
+)
+
+// I2C enumerated types
+type (
+	resultFlag   uint32
+	statusFlag   uint32
+	transferFlag uint32
+	commandFlag  uint32
+	stateFlag    uint32
+)
+
+const (
+	// general purpose results
+	resultSuccess         resultFlag = 0x0 // success
+	resultFail            resultFlag = 0x1 // fail
+	resultReadOnly        resultFlag = 0x2 // read only failure
+	resultOutOfRange      resultFlag = 0x3 // out of range access
+	resultInvalidArgument resultFlag = 0x4 // invalid argument check
+	// I2C-specific results
+	resultBusy                 resultFlag = 0x0384 + 0x0 // the controller is already performing a transfer
+	resultIdle                 resultFlag = 0x0384 + 0x1 // the peripheral driver is idle
+	resultNak                  resultFlag = 0x0384 + 0x2 // the peripheral device sent a NAK in response to a byte
+	resultFifoError            resultFlag = 0x0384 + 0x3 // FIFO under run or overrun
+	resultBitError             resultFlag = 0x0384 + 0x4 // transferred bit was not seen on the bus
+	resultArbitrationLost      resultFlag = 0x0384 + 0x5 // arbitration lost error
+	resultPinLowTimeout        resultFlag = 0x0384 + 0x6 // SCL or SDA were held low longer than the timeout
+	resultNoTransferInProgress resultFlag = 0x0384 + 0x7 // attempt to abort a transfer when one is not in progress
+	resultDmaRequestFail       resultFlag = 0x0384 + 0x8 // DMA request failed
+	resultTimeout              resultFlag = 0x0384 + 0x9 // timeout polling status flags
+)
+
+const (
+	statusTxReady         statusFlag = nxp.LPI2C_MSR_TDF  // transmit data flag
+	statusRxReady         statusFlag = nxp.LPI2C_MSR_RDF  // receive data flag
+	statusEndOfPacket     statusFlag = nxp.LPI2C_MSR_EPF  // end Packet flag
+	statusStopDetect      statusFlag = nxp.LPI2C_MSR_SDF  // stop detect flag
+	statusNackDetect      statusFlag = nxp.LPI2C_MSR_NDF  // NACK detect flag
+	statusArbitrationLost statusFlag = nxp.LPI2C_MSR_ALF  // arbitration lost flag
+	statusFifoErr         statusFlag = nxp.LPI2C_MSR_FEF  // FIFO error flag
+	statusPinLowTimeout   statusFlag = nxp.LPI2C_MSR_PLTF // pin low timeout flag
+	statusI2CDataMatch    statusFlag = nxp.LPI2C_MSR_DMF  // data match flag
+	statusBusy            statusFlag = nxp.LPI2C_MSR_MBF  // busy flag
+	statusBusBusy         statusFlag = nxp.LPI2C_MSR_BBF  // bus busy flag
+
+	// all flags which are cleared by the driver upon starting a transfer
+	statusClear statusFlag = statusEndOfPacket | statusStopDetect | statusNackDetect |
+		statusArbitrationLost | statusFifoErr | statusPinLowTimeout | statusI2CDataMatch
+
+	// IRQ sources enabled by the non-blocking transactional API
+	statusIrq statusFlag = statusArbitrationLost | statusTxReady | statusRxReady |
+		statusStopDetect | statusNackDetect | statusPinLowTimeout | statusFifoErr
+
+	// errors to check for
+	statusError statusFlag = statusNackDetect | statusArbitrationLost | statusFifoErr |
+		statusPinLowTimeout
+)
+
+// LPI2C transfer modes
+const (
+	transferDefault       transferFlag = 0x0 // transfer starts with a start signal, stops with a stop signal
+	transferNoStart       transferFlag = 0x1 // don't send a start condition, address, and sub address
+	transferRepeatedStart transferFlag = 0x2 // send a repeated start condition
+	transferNoStop        transferFlag = 0x4 // don't send a stop condition
+)
+
+// LPI2C FIFO commands
+const (
+	commandTxData commandFlag = (0x0 << nxp.LPI2C_MTDR_CMD_Pos) & nxp.LPI2C_MTDR_CMD_Msk // transmit
+	commandRxData commandFlag = (0x1 << nxp.LPI2C_MTDR_CMD_Pos) & nxp.LPI2C_MTDR_CMD_Msk // receive
+	commandStop   commandFlag = (0x2 << nxp.LPI2C_MTDR_CMD_Pos) & nxp.LPI2C_MTDR_CMD_Msk // generate STOP condition
+	commandStart  commandFlag = (0x4 << nxp.LPI2C_MTDR_CMD_Pos) & nxp.LPI2C_MTDR_CMD_Msk // generate (REPEATED)START and transmit
+)
+
+// LPI2C transactional states
+const (
+	stateIdle              stateFlag = 0x0
+	stateSendCommand       stateFlag = 0x1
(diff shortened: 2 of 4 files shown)


deadprogram:

ardnew:
rebased against `dev`

src/machine/machine_mimxrt1062.go
```diff
@@ -341,10 +341,10 @@ func (jt *pinJumpTable) dispatchInterrupt(interrupt.Interrupt) {
...
 			gpio.ISR.Set(status) // clear interrupt
 			for status != 0 {
-				p := Pin(bits.TrailingZeros32(status))
-				i := Pin(port + p)
-				jt.lut[i](i)
-				status &^= 1 << p
+				off := Pin(bits.TrailingZeros32(status)) // ctz
+				pin := Pin(port + off)
+				jt.lut[pin](pin)
+				status &^= 1 << off
```
aykevl, line 348: Looks unrelated but certainly an improvement so OK.

src/machine/machine_mimxrt1062_i2c.go
```diff
@@ -0,0 +1,725 @@
...
+	// are connected to the LPI2C device, and should be defined where the I2C
+	// instance is declared (e.g., in the board definition). see the godoc
+	// comments on type muxSelect for more details.
+	muxSDA, muxSCL muxSelect
+
+	// these are copied from I2CConfig, during (*I2C).Configure(I2CConfig), and
+	// should be considered read-only for internal reference (i.e., modifying them
+	// will have no desirable effect).
+	sda, scl  Pin
+	frequency uint32
```
aykevl: Why do you need to store this configuration? It doesn't seem necessary.

src/machine/machine_mimxrt1062_i2c.go
```diff
@@ -0,0 +1,725 @@
...
+
+	// auxiliary state data used internally
+	configured bool
```
aykevl: This flag is unused so can be removed.

src/machine/machine_mimxrt1062_i2c.go
```diff
@@ -0,0 +1,725 @@
...
+}
+
+type I2CDirection bool
+
+const (
+	DirectionWrite I2CDirection = false
+	DirectionRead  I2CDirection = true
+)
```
aykevl: These symbols are exported but only seem to be used internally, so should not be exported.

src/machine/machine_mimxrt1062_i2c.go
```diff
@@ -0,0 +1,725 @@
...
+	shift(I2CDirection) uint32
+}
+
+// address7Bit and address10Bit stores the unshifted original I2C peripheral
+// address in an unsigned integral data type and implements the addressable
+// interface to reformat addresses as required for read/write operations.
+//   TODO:
+//    add 10-bit address support
+type (
+	address7Bit uint8
```
aykevl: It looks like you did this for future compatibility?

I am generally hesitant to include code that does not currently serve a purpose, especially when it complicates things (here with an interface value, even). Right now, only 7 bit addresses are supported.

Take a look at this code from the AVR:

https://github.com/tinygo-org/tinygo/blob/3cdc110462369500aedfc1511f7c409c49200e35/src/machine/machine_atmega.go#L70-L75

To me, this is much simpler and much easier to understand. It might need revisiting once 10 bit addresses get supported (although I haven't seen any such devices yet) but those 3 LOC for address calculation are all that's really needed.

src/machine/machine_mimxrt1062_i2c.go
```diff
@@ -0,0 +1,725 @@
...
+	// return an error if the bus is already in use by another controller
+	if i2c.isBusBusy() {
+		return resultBusy
+	}
+	// clear all flags
+	i2c.Bus.MSR.Set(uint32(statusClear))
+	// turn off auto-stop
+	i2c.Bus.MCFGR1.ClearBits(nxp.LPI2C_MCFGR1_AUTOSTOP)
+
+	cmd := newCommandBuffer(7)
```
aykevl: To me it looks like `commandBuffer` can easily be replaced with regular slice operations: `make([]byte, 0, 7)` and `append`. That would be shorter in code, (in my opinion) easier to understand and probably even more efficient.

src/machine/machine_mimxrt1062_i2c.go
ardnew: Removed from struct and from `Configure`

src/machine/machine_mimxrt1062_i2c.go
ardnew: Removed

src/machine/machine_mimxrt1062_i2c.go
ardnew: unexported type and constants:

```go
type i2cDirection bool

const (
	directionWrite i2cDirection = false
	directionRead  i2cDirection = true
)
```

src/machine/machine_mimxrt1062_i2c.go
ardnew: Also have never encountered a device with 10-bit address. Will refactor to remove this address generalization

ardnew: Not sure if it was intentional, but there is an inconsistency with the type of peripheral addresses in the `machine` package.

In `machine/i2c.go`, we have the general purpose `ReadRegister` and `WriteRegister` that would both be unable to support future 10-bit addresses (which their godoc comments rightfully warn the user) with a `uint8` parameter:

```go
func (i2c I2C) WriteRegister(address uint8, register uint8, data []byte) error
func (i2c I2C) ReadRegister(address uint8, register uint8, data []byte) error
```

Yet all of the machine implementation-specific `Tx` methods (which the above depend on) provision future 10-bit support with a `uint16` parameter:

```
$ grep -RF 'I2C) Tx('
 src/machine/machine_k210.go:func (i2c I2C) Tx(addr uint16, w, r []byte) error {
 src/machine/machine_mimxrt1062_i2c.go:func (i2c I2C) Tx(addr uint16, w, r []byte) error {
 src/machine/machine_attiny.go:func (i2c I2C) Tx(addr uint16, w, r []byte) error {
 src/machine/machine_atsamd21.go:func (i2c I2C) Tx(addr uint16, w, r []byte) error {
 src/machine/machine_atmega.go:func (i2c I2C) Tx(addr uint16, w, r []byte) error {
 src/machine/machine_generic.go:func (i2c I2C) Tx(addr uint16, w, r []byte) error {
 src/machine/machine_stm32_i2c.go:func (i2c I2C) Tx(addr uint16, w, r []byte) error {
 src/machine/machine_atsamd51.go:func (i2c I2C) Tx(addr uint16, w, r []byte) error {
 src/machine/machine_stm32f103xx.go:func (i2c I2C) Tx(addr uint16, w, r []byte) error {
 src/machine/machine_fe310.go:func (i2c I2C) Tx(addr uint16, w, r []byte) error {
 src/machine/machine_nrf.go:func (i2c I2C) Tx(addr uint16, w, r []byte) (err error) {
```

This seems backwards to me. Specifically, the _machine-independent_ API `ReadRegister`/`WriteRegister` should be insulated from future changes that add 10-bit addressing support by simply widening their `address` type to `uint16`.

The 7-bit-only policy should instead be considered a defect (or at least an incomplete implementation), enforced by convention and documentation, so that adding 10-bit support in the future is a transparent upgrade to current users and would not become an API-breaking change.

src/machine/machine_mimxrt1062_i2c.go
ardnew: Ah, indeed. Got caught up porting C code and didn't think through the overall intent. Will remove this type.

ardnew:
rebased against dev

deadprogram:
Sorry I have not looked at this PR sooner. I will get out my Teensy 4.0 and try it out in a bit.

deadprogram:
Was trying this PR but https://github.com/tinygo-org/tinygo/pull/1454#issuecomment-778838171

General comment: should define a I2C0 to point to I2C1 in this PR?

src/machine/machine_mimxrt1062_i2c.go
aykevl: Well, yeah you are right in that it is inconsistent. However, I do not agree that one or the other is necessarily a defect, they are just inconsistent with each other.

It might be worth changing the API here, or even removing them entirely as `WriteRegister` forces a heap allocation which you generally wouldn't want (even though these methods are very convenient). In any case, this discussion is somewhat separate from this PR.

This might be an interesting read: https://www.kernel.org/doc/Documentation/i2c/ten-bit-addresses

This as well: https://github.com/rust-embedded/embedded-hal/issues/147

ardnew:
rebased against dev (3eb9dca695254b9d59f30597ea6c409e1b65e7ee)

src/machine/machine_mimxrt1062_i2c.go
```diff
@@ -0,0 +1,646 @@
...
+	SCL       Pin
+}
+
+func (c I2CConfig) getPins() (sda, scl Pin) {
+	if 0 == c.SDA && 0 == c.SCL {
+		// default pins if none specified
+		return I2C_SDA_PIN, I2C_SCL_PIN
+	}
+	return c.SDA, c.SCL
+}
```
sago35: I think Pin needs to be specified according to the value of muxSelect{}.
If I'm right about this, why not add the Pin information to muxSelect{}?

Or, depending on the settings of I2CConfig-SDA, etc., maxSelect{} should be generated automatically.

src/machine/machine_mimxrt1062_i2c.go
```diff
@@ -0,0 +1,646 @@
...
+	// set FIFO watermarks (RX=1, TX=1)
+	mfcr := (uint32(0x1) << nxp.LPI2C_MFCR_RXWATER_Pos) & nxp.LPI2C_MFCR_RXWATER_Msk
+	mfcr |= (uint32(0x1) << nxp.LPI2C_MFCR_TXWATER_Pos) & nxp.LPI2C_MFCR_TXWATER_Msk
+	i2c.Bus.MFCR.Set(mfcr)
+
+	// configure clock using receiver frequency
+	i2c.setFrequency(freq)
+
+	// clear reset, and enable the interface
+	i2c.Bus.MCR.Set(nxp.LPI2C_MCR_MEN)
```
sago35, line 290: When I tried it in drivers/examples/ssd1306/i2c_128x64, the transmission immediately after initialization disappeared.
Perhaps a wait loop like the following is needed.
Timeout may also be needed.

```suggestion
	i2c.Bus.MCR.Set(nxp.LPI2C_MCR_MEN)

	// wait for the I2C bus to idle
	for {
		if i2c.Bus.MSR.Get()&nxp.LPI2C_MSR_BBF == 0 {
			break
		}
	}
```

src/runtime/runtime_mimxrt1062_clock.go
```diff
@@ -166,7 +166,7 @@ func initClocks() {
 	nxp.ClockIpLpi2c2.Enable(false) //
 	nxp.ClockIpLpi2c3.Enable(false) //
```
sago35, line 168: ClockIpLpi2c4.Enable() is required.
The same change is required for enablePeripheralClocks().

sago35 (commented):
@ardnew 
Both i2c0 and i2c2 worked, but i2c4 did not.
I think it needs a few fixes, but for the most part it is very good.

deadprogram:
@ardnew if you get a chance to respond to @sago35 feedback on this board it would be great. Thank you!

src/machine/machine_mimxrt1062_i2c.go
ardnew: Not sure if I'm understanding this one correctly, but adding pin information to `muxSelect{}` would: 
1. Require another (large) lookup table that would take considerable effort to create, and 
2. The user would still need to define which `muxSelect{}` to associate with their `I2C{}`. 

The intent of this method was to automatically determine the pins in case the user has intentionally left their `I2C{}` _unconfigured_ – to let TinyGo determine the default settings.

There are two separate devices that must be configured to connect a given pin to a given peripheral. For example, digital pin `18` (port `A` pin `PA17`, also known as `AD_B1_01`) can provide I2C SDA functionality to `I2C1` (peripheral `LPI2C1`):
|Alt Fn|GPIO|IOMUXC|
|:-----:|:----:|:-------:|
|I2C SDA|Configure `AD_B1_01` as `LPI2C1 SDA`|Connect `LPI2C1 SDA` to `AD_B1_01`|

The tricky part is while `AD_B1_01` can be configured with different alternate functions (e.g., I2C SDA, SPI MOSI, etc.), **multiple different pins can provide I2C SDA functionality to `LPI2C1`**.

In the code, the GPIO configuration is done in `(Pin).Configure(PinConfig)` via `(Pin).getMuxMode(PinConfig)`, and IOMUXC connection is done in `(*I2C).Configure(I2CConfig)` via this `(I2CConfig).getPins()` method. 

Similar to how `(Pin).getMuxMode(PinConfig)` automatically determines the alternate function mode based on both `Pin` and peripheral (via `PinConfig`), this `(I2CConfig).getPins()` could be expanded to automatically determine the pins based on peripheral alone (we already know the alternate function mode is I2C SDA/SCL). Something _similar_ to this:

```go
func (c I2CConfig) getPins(i2c *I2C) (sda, scl Pin) {
	if 0 == c.SDA && 0 == c.SCL {
		// default pins if none specified
		switch i2c {
		case I2C1:
			return I2C1_SDA_PIN, I2C1_SCL_PIN
		case I2C2:
			return I2C2_SDA_PIN, I2C2_SCL_PIN
		case I2C3:
			return I2C3_SDA_PIN, I2C3_SCL_PIN
		}
		// the given *I2C is unsupported
		return I2C_SDA_PIN, I2C_SCL_PIN
	}
	return c.SDA, c.SCL
}
```

I'll go ahead and update this method if this looks right to you.

src/runtime/runtime_mimxrt1062_clock.go
ardnew: Good catch! Wonder why I missed this? 😕

src/machine/machine_mimxrt1062_i2c.go
ardnew: OK, looks great to me. Reworded slightly:

```go
for i2c.Bus.MSR.Get()&nxp.LPI2C_MSR_BBF != 0 {
}
```

src/machine/machine_mimxrt1062_i2c.go
ardnew: I've added default SDA, SCL pins to type `I2C`:

```diff
  type I2C struct {
  	Bus *nxp.LPI2C_Type

+  	// these pins are initialized by each global I2C variable declared in the
+  	// board_teensy4x.go file according to the board manufacturer's default pin
+  	// mapping. they can be overridden with the I2CConfig argument given to
+  	// (*I2C).Configure(I2CConfig).
+  	sda, scl Pin
  
  	// these hold the input selector ("daisy chain") values that select which pins
  	// are connected to the LPI2C device, and should be defined where the I2C
  	// instance is declared (e.g., in the board definition). see the godoc
  	// comments on type muxSelect for more details.
  	muxSDA, muxSCL muxSelect
  }
```

And then for each peripheral, added something like:

```diff
  	_I2C1 = I2C{
  		Bus: nxp.LPI2C1,
+  		sda: I2C1_SDA_PIN, // D18 (PA17 [AD_B1_01])
+  		scl: I2C1_SCL_PIN, // D19 (PA16 [AD_B1_00])
-  		muxSDA: muxSelect{ // D18 (PA17 [AD_B1_01])
+  		muxSDA: muxSelect{
  			mux: nxp.IOMUXC_LPI2C1_SDA_SELECT_INPUT_DAISY_GPIO_AD_B1_01_ALT3,
  			sel: &nxp.IOMUXC.LPI2C1_SDA_SELECT_INPUT,
  		},
-  		muxSCL: muxSelect{ // D19 (PA16 [AD_B1_00])
+  		muxSCL: muxSelect{
  			mux: nxp.IOMUXC_LPI2C1_SCL_SELECT_INPUT_DAISY_GPIO_AD_B1_00_ALT3,
  			sel: &nxp.IOMUXC.LPI2C1_SCL_SELECT_INPUT,
  		},
  	}
```

The method `(I2CConfig).getPins() (Pin, Pin)` was replaced with a method to override these default pins when the user calls `(*I2C).Configure(I2CConfig)`:

```diff
+  func (i2c *I2C) setPins(c I2CConfig) (sda, scl Pin) {
+  	// if both given pins are defined, or either receiver pin is undefined.
+  	if 0 != c.SDA && 0 != c.SCL || 0 == i2c.sda || 0 == i2c.scl {
+  		// override the receiver's pins.
+  		i2c.sda, i2c.scl = c.SDA, c.SCL
+  	}
+  	// return the selected pins.
+  	return i2c.sda, i2c.scl
+  }
  
  // Configure is intended to setup an I2C interface for transmit/receive.
  func (i2c *I2C) Configure(config I2CConfig) {
  	// init pins
-  	sda, scl := config.getPins()
+  	sda, scl := i2c.setPins(config)
  
  	// configure the mux and pad control registers
  	sda.Configure(PinConfig{Mode: PinModeI2CSDA})
  	scl.Configure(PinConfig{Mode: PinModeI2CSCL})
  
  	// configure the mux input selector
  	i2c.muxSDA.connect()
  	i2c.muxSCL.connect()
```

Now, whether the user calls `p.Configure(I2CConfig{})` with `p` being `I2C0`, `I2C1`, `I2C2`, `I2C3`, or `I2C4`, it will use the correct pins for the given peripheral by default.

ardnew:
Addressed remaining comments and rebased against dev (b5c5d95b689979780ea01ed4c8d4fdb40602535b)

deadprogram (approved):

deadprogram:
Since https://github.com/tinygo-org/tinygo/pull/2924 passed CI I will squash/merge this PR now. Thank you very much @ardnew

PR #1473, opened 2020-10-31
Add missing return pointer restore for regular coroutine tail calls

### LINKED ISSUE #1467: Miscompilation of hash/crc64: return pointer is not restored on tail call
I have noticed a miscompilation in hash/crc64:

    tinygo test -opt=1 hash/crc64

I have tracked it to `hash/crc64.tableSum`, which looks like this:

```go
func tableSum(t *Table) uint64 {
	var a [2048]byte
	b := a[:0]
	if t != nil {
		for _, x := range t {
			b = appendUint64(b, x)
		}
	}
	return Checksum(b, MakeTable(ISO))
}
```

If I change it to this, the test passes:

```go
func tableSum(t *Table) uint64 {
	var a [2048]byte
	b := a[:0]
	if t != nil {
		for _, x := range t {
			b = appendUint64(b, x)
		}
	}
	checksum := Checksum(b, MakeTable(ISO))
	return checksum
}
```

The only difference are the last two lines, before it would return the result of `Checksum` directly and afterwards it returns them in a new variable. This _should_ not make a difference, except that the debug information is slightly different. In fact, when I compare the IR the only difference is two calls to `llvm.dbg.value` which are added in the working version.

I suspect the bug lies in the coroutine code as the output after the coroutine transformation is quite different, as a lot more extra calls are inserted in the working version (for example, a call to `(*internal/task.Task).getReturnPtr`). This tells me two things:

1. The coroutine optimization currently doesn't ignore calls to `llvm.dbg.value` while it should (debug information should not affect the program itself in any way).
2. It appears there is a bug in the coroutine tail call optimization.

To see the IR output, you can dump the function with a small change to the TinyGo source:

```diff
$ git diff
diff --git a/transform/optimizer.go b/transform/optimizer.go
index 917083ce..30b40c22 100644
--- a/transform/optimizer.go
+++ b/transform/optimizer.go
@@ -117,10 +117,12 @@ func Optimize(mod llvm.Module, config *compileopts.Config, optLevel, sizeLevel i
        switch config.Scheduler() {
        case "coroutines":
                // Lower async as coroutines.
+               mod.NamedFunction("hash/crc64.tableSum").Dump()
                err := LowerCoroutines(mod, config.NeedsStackObjects())
                if err != nil {
                        return []error{err}
                }
+               mod.NamedFunction("hash/crc64.tableSum").Dump()
        case "tasks":
                // No transformations necessary.
        case "none":
```

and run `tinygo test -opt=1 hash/crc64`.

@niaow can you maybe take a look?

niaow:
This fixes #1467 where a normal suspending call followed by a plain tail call would result in the tail return value being written to the return pointer of the normal suspending call. This is fixed by saving the return pointer at the start of the function and restoring it before initiating a plain tail call.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,6 +188,7 @@ tinygo-test:
 	$(TINYGO) test encoding/base32
 	$(TINYGO) test encoding/hex
 	$(TINYGO) test hash/fnv
+	$(TINYGO) test hash/crc64
 	$(TINYGO) test math
 	$(TINYGO) test text/scanner
 	$(TINYGO) test unicode/utf8
--- a/transform/coroutines.go
+++ b/transform/coroutines.go
@@ -735,10 +735,10 @@ func (c *coroutineLoweringPass) lowerFuncCoro(fn *asyncFunc) {
 	task := c.builder.CreateCall(c.current, []llvm.Value{llvm.Undef(c.i8ptr), fn.rawTask}, "task")
 	parentState := c.builder.CreateCall(c.setState, []llvm.Value{task, coroState, llvm.Undef(c.i8ptr), llvm.Undef(c.i8ptr)}, "task.state.parent")
 	// Get return pointer if needed.
-	var retPtr llvm.Value
-	if fn.hasValueStoreReturn() {
-		retPtr = c.builder.CreateCall(c.getRetPtr, []llvm.Value{task, llvm.Undef(c.i8ptr), llvm.Undef(c.i8ptr)}, "task.retPtr")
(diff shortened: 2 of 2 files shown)


niaow:
It looks like the setup for `test-llvm11-go115` is broken?

niaow:
Ah, i see what happened and i am fixing the makefile

niaow:
I think #1474 should fix the CI failure

deadprogram:
Can you please rebase now so we can make sure that tests are passing? Thanks.

niaow:
Done.

aykevl:
Looks good to me, thank you!

Can you also add the hash/crc64 test to the Makefile? (see `tinygo-test:`).

niaow:
@aykevl done

deadprogram:
Thanks very much @niaow now merging.

PR #1475, opened 2020-10-31
avr: properly support the .rodata section

aykevl:
Unfortunately, the .rodata section can't be stored in flash. Instead, an
explicit .progmem section should be used, which is supported in LLVM as
address space 1 but not exposed to normal programs.

Eventually a pass should be written that converts trivial const globals
of which all loads are visible to be in addrspace 1, to get the benefits
of storing those globals directly in ROM.


deadprogram:
:scissors: merging. Thanks @aykevl

PR #1479, opened 2020-11-04
builder: simplify detection of Clang header paths

aykevl:
This detection mechanism should work on both Debian and Fedora, and hopefully other Linux distributions too.

By simplifying this it should be possible to more easily add exceptions for specific systems.

---

Note, this should be merged after the coming release to avoid potential disruption.

### DIFF
--- a/builder/env.go
+++ b/builder/env.go
@@ -1,10 +1,9 @@
 package builder
 
 import (
-	"io/ioutil"
 	"os"
-	"os/exec"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Hello @QuLogic does this look like it will address some of the issues you have pointed out in Fedora?

deadprogram:
@QuLogic did you happen to see this PR? What do you think?

QuLogic:
I did not fully test this, but it seems okay, except that I don't think `Glob` is guaranteed to be in any order: golang/go#17153

deadprogram:
Oops, somehow overlooked this PR. Is it still relevant/needed?

aykevl:
> except that I don't think `Glob` is guaranteed to be in any order: [golang/go#17153](https://github.com/golang/go/issues/17153)

Fair enough. I've added a `sort.Strings` call to be sure.

> Is it still relevant/needed?

It's still a simplification and simplifying things is usually a good thing. So I'd say yes.

deadprogram:
Just wondering if this still works as we expect?

deadprogram:
@aykevl is this PR still going to be worked on, or should we close it?

aykevl:
Similar changes have been done here: https://github.com/tinygo-org/tinygo/pull/3948

PR #1481, opened 2020-11-04
atsamd5x: improve USBCDC

sago35:
The following approaches were used to speed up the process.

* It used to send every 1 byte, but it now holds a maximum of 63 bytes before sending
*  if the buffer is full, send buffer
* If there is data at the time of the handleUSBIRQ every 1ms, send buffer

Unlike #1452, it does not require additional memory.
The code is a little more complicated, but it looks like it works very well.

* 13 byte : fmt.Printf("hello world\r\n")
  * old : 202.814 us
  * new :  51.330 us
* 11 byte : println("hello world")
  * old : 171.274 us
  * new :  11.920 us
* 1024 byte + "\r\n"
  * old : 13.999962 ms
  * new :  1.150692 ms
* 1024 * 10 byte + "\r\n"
  * old : 139.649294 ms
  * new :  11.612682 ms

`interrupt.Disable()` cannot be built unless interp-rewrite is merged.

Samd2x can be made faster in the same way.
Once we have reviewed this PR, we will create a PR.

### DIFF
--- a/src/machine/machine_atsamd51.go
+++ b/src/machine/machine_atsamd51.go
@@ -11,6 +11,7 @@ import (
 	"device/arm"
 	"device/sam"
 	"runtime/interrupt"
+	"runtime/volatile"
 	"unsafe"
 )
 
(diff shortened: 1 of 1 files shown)


sago35:
The reason why it is set to draft is that we need to merge the interp-rewrite branches to be able to build.
I would like a review.

I tested it with the following code with WioTerminal.

```go
package main

import (
	"fmt"
	"machine"
	"time"
)

var buf = [1024 * 10]byte{}

var (
	debugPort1 = machine.BCM3
)

func main() {
	debugPort1.Configure(machine.PinConfig{Mode: machine.PinOutput})

	for i := range buf {
		if ((i + 1) & 0x0F) == 0 {
			buf[i] = byte(fmt.Sprintf("%X", byte(i>>4)&0xF)[0])
		} else {
			buf[i] = '.'
		}
	}
	buf[0] = '+'

	for {
		debugPort1.High()

		machine.UART0.Write(buf[:1024*10])
		//machine.UART0.Write(buf[:1024])
		machine.UART0.Write([]byte("\r\n"))
		//fmt.Printf("hello world\r\n")
		//println("hello world")

		debugPort1.Low()
		time.Sleep(time.Second)
	}
}

```

sago35:
* Use udd_ep_in_cache_buffer[usb_CDC_ENDPOINT_IN][128] as two 64 byte buffers.
  * The term udd_ep_in_cache_buffer is abbreviated to buffer below
* WriteByte()
  * Set `c byte` to buffer[TxIdx].
  * Can write up to 63 bytes.
  * The reason it's 63 bytes instead of 64 bytes is because `TxIdx & 0x3F` treats it as a size
  * Call Flush() if you've written to the maximum
* Flush()
  * Flush() sends the data in the buffer
  * Calculate the size to be sent using `TxIdx & 0x3F`
  * When the size is greater than 0, it is actually sent to USBCDC
  * After sending, set TxIdx to the next buffer
    * For example, when TxIdx is 0 to 63, TxIdx is set to 64. And Set TxIdx to 0 when the value is 64 to 127.
  * Set UART0.sent to true if it is sent once.
    * If not using UART0.sent, USBCDC status is not known at WriteByte() time.
* handleUSBIRQ()
  * Interrupts every 1ms
  * Calling Flush()

sago35:
It is buildable on the current dev branch so it is ready for review.

I wish I could make the implementation more straightforward, but I can't think of a good way to do it right now.

The code for testing can be found at
https://github.com/tinygo-org/tinygo/pull/1481#issuecomment-721702039

I checked it on Windows 10 and found the following

* 1024 * 10 byte + "\r\n"
  * old :  514.359 ms
  * new : 11.633ms

src/machine/machine_atsamd51.go
```diff
@@ -1733,36 +1734,81 @@ func (pwm PWM) getMux() PinMode {
...
-// WriteByte writes a byte of data to the USB CDC interface.
-func (usbcdc USBCDC) WriteByte(c byte) error {
-	// Supposedly to handle problem with Windows USB serial ports?
+// Flush flushes buffered data.
+func (usbcdc *USBCDC) Flush() error {
 	if usbLineInfo.lineState > 0 {
-		// set the data
-		udd_ep_in_cache_buffer[usb_CDC_ENDPOINT_IN][0] = c
+		idx := usbcdc.TxIdx.Get()
+		sz := idx & 0x3F
```
deadprogram: I suggest replace `0x3F` with some const like `maxBufferIdx` to make it more clear instead of the magic value.

src/machine/machine_atsamd51.go
```diff
@@ -1733,36 +1734,81 @@ func (pwm PWM) getMux() PinMode {
...
+				// waiting for the next flush(), because the transmission is not complete
+				return nil
+			}
+			usbcdc.waitTxc = true
 
-		// clean multi packet size of bytes already sent
-		usbEndpointDescriptors[usb_CDC_ENDPOINT_IN].DeviceDescBank[1].PCKSIZE.ClearBits(usb_DEVICE_PCKSIZE_MULTI_PACKET_SIZE_Mask << usb_DEVICE_PCKSIZE_MULTI_PACKET_SIZE_Pos)
+			// set the data
+			usbEndpointDescriptors[usb_CDC_ENDPOINT_IN].DeviceDescBank[1].ADDR.Set(uint32(uintptr(unsafe.Pointer(&udd_ep_in_cache_buffer[usb_CDC_ENDPOINT_IN][idx&0xC0]))))
+			usbcdc.TxIdx.Set(0x40 - idx&0xC0)
```
deadprogram: What is the significance of `0x40` and `0xC0` here? Same as above, I suggest using some appropriately named const values here.

src/machine/machine_atsamd51.go
```diff
@@ -1733,36 +1734,81 @@ func (pwm PWM) getMux() PinMode {
...
+// WriteByte writes a byte of data to the USB CDC interface.
+func (usbcdc *USBCDC) WriteByte(c byte) error {
+	// Supposedly to handle problem with Windows USB serial ports?
+	if usbLineInfo.lineState > 0 {
+		ok := false
+		for {
+			mask := interrupt.Disable()
+
+			idx := UART0.TxIdx.Get()
+			if (idx & 0x3F) != 0x3F {
```
deadprogram: See previous comments about const.

deadprogram:
The code does appear to work. I made a few comments about using `const` values for clarity.

@aykevl do you have any feedback?

src/machine/machine_atsamd51.go
sago35: I named it usbcdcTxSizeMask.

src/machine/machine_atsamd51.go
sago35: I rewrote it as follows

```go
			if bk == usbcdcTxBank1st {
				usbcdc.TxIdx.Set(usbcdcTxBank2nd)
			} else {
				usbcdc.TxIdx.Set(usbcdcTxBank1st)
			}
```

src/machine/machine_atsamd51.go
sago35: I named it usbcdcTxSizeMask.

note: The 0x3F on the left is usbcdcTxSizeMask, but the 0x3F on the right is not really usbcdcTxSizeMask, but something like usbcdcTxSizeMax.

sago35:
Updated. @deadprogram

src/machine/machine_atsamd51.go
```diff
@@ -1733,36 +1734,94 @@ func (pwm PWM) getMux() PinMode {
...
+	usbcdcTxBank2nd  uint8 = usbcdcTxSizeMask + 1
+)
+
+// Flush flushes buffered data.
+func (usbcdc *USBCDC) Flush() error {
 	if usbLineInfo.lineState > 0 {
-		// set the data
-		udd_ep_in_cache_buffer[usb_CDC_ENDPOINT_IN][0] = c
+		idx := usbcdc.TxIdx.Get()
+		sz := idx & usbcdcTxSizeMask
```
deadprogram, line 1754: These lines:

```go
sz := idx & usbcdcTxSizeMask
bk := idx & usbcdcTxBankMask
if 0 < sz {
```

could be shortened to:

```go
if 0 < idx & usbcdcTxSizeMask {
```

src/machine/machine_atsamd51.go
```diff
@@ -1733,36 +1734,94 @@ func (pwm PWM) getMux() PinMode {
...
+			if usbcdc.waitTxc {
+				// waiting for the next flush(), because the transmission is not complete
+				return nil
+			}
+			usbcdc.waitTxc = true
 
-		usbEndpointDescriptors[usb_CDC_ENDPOINT_IN].DeviceDescBank[1].ADDR.Set(uint32(uintptr(unsafe.Pointer(&udd_ep_in_cache_buffer[usb_CDC_ENDPOINT_IN]))))
+			// set the data
+			usbEndpointDescriptors[usb_CDC_ENDPOINT_IN].DeviceDescBank[1].ADDR.Set(uint32(uintptr(unsafe.Pointer(&udd_ep_in_cache_buffer[usb_CDC_ENDPOINT_IN][bk]))))
+			if bk == usbcdcTxBank1st {
```
deadprogram, line 1766: If you change the lines above as I was suggesting, this line can be:

```go
if idx & usbcdcTxBankMask == usbcdcTxBank1st {
```

src/machine/machine_atsamd51.go
```diff
@@ -1733,36 +1734,94 @@ func (pwm PWM) getMux() PinMode {
...
+// WriteByte writes a byte of data to the USB CDC interface.
+func (usbcdc *USBCDC) WriteByte(c byte) error {
+	// Supposedly to handle problem with Windows USB serial ports?
+	if usbLineInfo.lineState > 0 {
+		ok := false
+		for {
+			mask := interrupt.Disable()
+
+			idx := UART0.TxIdx.Get()
+			sz := idx & usbcdcTxSizeMask
```
deadprogram: Couldn't these lines

```go
sz := idx & usbcdcTxSizeMask
if sz < usbcdcTxSizeMask {
```

just be

```go
if idx & usbcdcTxSizeMask < usbcdcTxSizeMask {
```

deadprogram:
Made a few small comments @sago35 that look like we can shorten by a couple of lines of code.

src/machine/machine_atsamd51.go
sago35: L1777 also uses sz, so it would be easier to understand if you don't change it.
Would it be better to change it?

src/machine/machine_atsamd51.go
sago35: L1765 also uses bk, so it would be easier to understand if you don't change it.
Would it be better to change it?

src/machine/machine_atsamd51.go
sago35: I've changed it.

sago35:
Updated. @deadprogram 

I thought USBCDC.Flush() should not be changed.
Should I change it?

src/machine/machine_atsamd51.go
deadprogram: OK, not worth worrying about anyway, I'm sure.

deadprogram: Sure, can just leave it how it is.

deadprogram:
I think this is now fine how it is. @sago35 can you please squash into a single commit? Thank you!

sago35:
squashed. @deadprogram

deadprogram:
This is really great, thank you for working on it @sago35 just waiting for CI to finish to merge.

Now we just need the same thing for SAMD21! :smile_cat:

sago35:
I will prepare the samd21 version as well.

PR #1482, opened 2020-11-04
main: update go-llvm to fix LLVM build tags for Linux

aykevl:
This makes sure TinyGo should be buildable again on Darwin.


deadprogram:
Merging.

PR #1484, opened 2020-11-04
board/teensy40: Add external interrupt support

ardnew:
### This PR is baselined on top of the original Teensy 4.0 BSP PR (#1425)

This PR adds support for external interrupts on _**all**_ GPIO pins (both routed and not routed to board headers).

This allows the user to attach interrupt handlers via `func (p Pin) SetInterrupt(change PinChange, callback func(Pin)) error`.

There are 5 detection modes (of type `PinChange`) supported, using the conventional identifiers as other TinyGo boards. Their purpose should be self-evident:

```go
const (
	PinLow     PinChange = 0
	PinHigh    PinChange = 1
	PinRising  PinChange = 2
	PinFalling PinChange = 3
	PinToggle  PinChange = 4
)
```

There are no other publicly visible changes to the existing Teensy 4 API other than this one method added to the `machine` package.

### DIFF
--- a/src/machine/machine_mimxrt1062.go
+++ b/src/machine/machine_mimxrt1062.go
@@ -19,39 +19,39 @@ type PinMode uint8
 
 const (
 	// GPIO
-	PinInput           PinMode = 0
-	PinInputPullUp     PinMode = 1
-	PinInputPullDown   PinMode = 2
-	PinOutput          PinMode = 3
-	PinOutputOpenDrain PinMode = 4
-	PinDisable         PinMode = 5
+	PinInput PinMode = iota
+	PinInputPullUp
+	PinInputPullDown
+	PinOutput
+	PinOutputOpenDrain
+	PinDisable
 
 	// ADC
-	PinInputAnalog PinMode = 6
+	PinInputAnalog
 
 	// UART
-	PinModeUARTTX PinMode = 7
-	PinModeUARTRX PinMode = 8
+	PinModeUARTTX
+	PinModeUARTRX
 
 	// SPI
-	PinModeSPISDI PinMode = 9
-	PinModeSPISDO PinMode = 10
-	PinModeSPICLK PinMode = 11
-	PinModeSPICS  PinMode = 12
+	PinModeSPISDI
+	PinModeSPISDO
+	PinModeSPICLK
+	PinModeSPICS
 
 	// I2C
-	PinModeI2CSDA PinMode = 13
-	PinModeI2CSCL PinMode = 14
+	PinModeI2CSDA
+	PinModeI2CSCL
 )
 
 type PinChange uint8
 
 const (
-	PinLow     PinChange = 0
-	PinHigh    PinChange = 1
-	PinRising  PinChange = 2
-	PinFalling PinChange = 3
-	PinToggle  PinChange = 4
+	PinLow PinChange = iota
+	PinHigh
+	PinRising
+	PinFalling
+	PinToggle
 )
 
 // pinJumpTable represents a function lookup table for all 128 GPIO pins.
--- a/src/machine/machine_mimxrt1062.go
+++ b/src/machine/machine_mimxrt1062.go
@@ -4,6 +4,8 @@ package machine
 
 import (
 	"device/nxp"
+	"math/bits"
+	"runtime/interrupt"
 	"runtime/volatile"
 )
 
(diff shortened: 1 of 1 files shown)


deadprogram:

ardnew:
rebased against `dev`

src/machine/machine_mimxrt1062.go
```diff
@@ -17,29 +19,56 @@ type PinMode uint8
...
 const (
 	// GPIO
-	PinInput PinMode = iota
-	PinInputPullUp
-	PinInputPullDown
-	PinOutput
-	PinOutputOpenDrain
-	PinDisable
+	PinInput           PinMode = 0
+	PinInputPullUp     PinMode = 1
```
aykevl: Are these changes intentional? The code looks cleaner to me as it was before (with `iota`).

src/machine/machine_mimxrt1062.go
```diff
@@ -301,6 +330,101 @@ func (p Pin) Toggle() {
...
 }
 
+// dispatchInterrupt invokes the user-provided callback functions for external
+// interrupts generated on the high-speed GPIO pins.
+//
+// Unfortunately, all four high-speed GPIO ports (A-D) are connected to just a
+// single interrupt control line. Therefore, the interrupt status register (ISR)
+// must be checked in all four GPIO ports on every interrupt.
+func (jt *pinJumpTable) dispatchInterrupt(interrupt.Interrupt) {
+	handle := func(gpio *nxp.GPIO_Type, port Pin) {
```
aykevl, line 340: This is probably okay but note that closures might cause a heap allocation, which is not allowed in interrupts (see #1460).

src/machine/machine_mimxrt1062.go
```diff
@@ -301,6 +330,101 @@ func (p Pin) Toggle() {
...
+		gpio.ISR.Set(mask)       // clear any pending interrupt (W1C)
+		gpio.IMR.ClearBits(mask) // disable external interrupt
+	}
+	// enable or disable the interrupt based on number of defined callbacks
+	if pinISR.numDefined > 0 {
+		if nil == pinInterrupt {
+			// create the Interrupt if it is not yet defined
+			irq := interrupt.New(nxp.IRQ_GPIO6_7_8_9, pinISR.dispatchInterrupt)
+			pinInterrupt = &irq
+			pinInterrupt.Enable()
```
aykevl, line 417: I don't really understand this, why do you use `&irq`?
```suggestion
			irq.Enable()
```

aykevl (commented):
I didn't look too deeply into the implementation but superficially this looks good to me.

src/machine/machine_mimxrt1062.go
ardnew: I keep a copy of the interrupt (as _global_ `pinInterrupt`) so that:

1. We don't have to create/enable it when it already exists and is already enabled (`nil != pinInterrupt`)
2. We can disable it when all user callbacks are removed (`pinISR.numDefined == 0`)

I have `pinInterrupt` defined as a pointer (`*interrupt.Interrupt`), and hence use `&irq`, so that I can check if it has been created or not via comparison to `nil`. Wasn't sure of a simple, equivalent way to test a _**value**_ of type `interrupt.Interrupt`.

#### _Related_:

For some reason, `interrupt.New` doesn't follow Go convention and returns a value type instead of a pointer. In this respect, I think it should have been named `interrupt.Make`. In particular, should have been either of these two:

> `func New(id int, handler func(Interrupt)) *Interrupt`
> `func Make(id int, handler func(Interrupt)) Interrupt`

ardnew: Wasn't sure of a way to assign directly to the _pointer_ `pinInterrupt` from the _value_ returned by `interrupt.New`, without using an intermediate variable `irq` and taking its address `&irq`.

Using `pinInterrupt = &interrupt.New(...)`  or `pinInterrupt = &(interrupt.New(...))` isn't valid

src/machine/machine_mimxrt1062.go
ardnew: Hmm, why is this okay then? I can verify it works, and it is an interrupt service routine. It _looks like_ a closure over the receiver variable `jt`... otherwise, maybe it is just an anonymous function? 

Or you're saying closures are okay if and only if they don't make an allocation. 

Of course I can refactor and pull `handle` out as a proper function if you recommend it

I also wasn't sure if the call to `bits.TrailingZeros32` was not allowed or not recommended in an ISR.

src/machine/machine_mimxrt1062.go
ardnew: They were intentional, but only because I was trying to match the convention from SAMD51. Looks like SAMD21 and STM32 do the same (see: `machine_atsamd21.go`, `machine_atsamd51.go`, and `machine_stm32_moder_gpio.go`)

I also found it useful to be able to immediately look up the integer value of these modes when debugging.

Now that these aren't being debugged anymore, there isn't really a benefit. 

I will revert these back to using `iota`.

src/machine/machine_mimxrt1062.go
deadprogram: I think we should probably revisit this specific point, but in a future PR, since there are enough interrelated PRs already! :smiley_cat:

src/machine/machine_mimxrt1062.go
deadprogram: This is going to need to be addressed further, but can be handled in a future PR.

deadprogram:
Thanks for working on this @ardnew now merging.

src/machine/machine_mimxrt1062.go
aykevl: Ah I see. Maybe `interrupt.New` is badly named then. It's actually a compiler intrinsic instead of a regular function call, so it will be evaluated at compile time instead of runtime and the return value is compiled down to a constant integer. This magic makes sure that the interrupt handler can be optimized away if the interrupt is never enabled.

Maybe something like `interrupt.Request` is better? Suggestions are definitely welcome.

src/machine/machine_mimxrt1062.go
aykevl: The reason why it's _probably_ okay in this case is that if only a single pointer is captured in a closure, it avoids a heap allocation. This is an implementation detail and there is no guarantee that this will remain the case (although I see no reason to change this).

If you want to be sure, you can try to apply #1460.

Issue #1408, opened 2020-09-25
findPinPadMapping panics (index out of range) when noPin is passed


conejoninja:
On the *pybadge* it's not possible to use the onboard display, the error might be caused by this change : https://github.com/tinygo-org/tinygo/commit/1451eeaf414176f2b96298516f3bf2c3296f8037

SPI1_SDI_PIN is defined as noPin (255) https://github.com/tinygo-org/tinygo/blob/release/src/machine/board_pybadge.go#L103
pinPadMapping is 64 length https://github.com/tinygo-org/tinygo/blob/release/src/machine/machine_atsamd51.go#L267

Code to reproduce the issue

```go
	machine.SPI1.Configure(machine.SPIConfig{
		SCK:       machine.SPI1_SCK_PIN,
		SDO:       machine.SPI1_SDO_PIN,
		SDI:       machine.SPI1_SDI_PIN,
		Frequency: 8000000,
	})
```

This fix the issue, but not sure about the implications : https://github.com/conejoninja/tinygo/commit/6c39859ec32b0d28ce3aadd0f0966079fc5cc380

sago35:
@conejoninja 
You can maybe set up a PB12 instead of NoPin as with the arduino.
You can't read the value because it's not connected, but I think the SPI setup will work!

https://github.com/adafruit/ArduinoCore-samd/blob/master/variants/pybadge_m4/variant.cpp#L104


> This fix the issue, but not sure about the implications : [conejoninja@6c39859](https://github.com/conejoninja/tinygo/commit/6c39859ec32b0d28ce3aadd0f0966079fc5cc380)

On a different note, I think it would be good to change it so that it doesn't become panic.
So I think this change is necessary as well.

sago35:
https://github.com/adafruit/ArduinoCore-samd/blob/master/variants/pygamer_m4/variant.cpp#L100-L101

PB12 is doubly defined here.
This seems to be a mistake on the Arduino side.

aykevl:
@conejoninja your proposed fix looks correct. Can you make a PR with it?

conejoninja:
@aykevl similar code was already merged here : https://github.com/tinygo-org/tinygo/commit/db27541b1a44a903feeeef91840314a56fcdc725

we still need to consider adding an extra check in findPinPadMapping so pin/2 isn't out of range

aykevl:
[...]

PR #1485, opened 2020-11-04
machine/atsamd51: handle configuration of output-only SPI where input pin not connected

### LINKED ISSUE #1483: SPI1.Configure crashes for pygamer
I want to use TFT on pygamer, but calling machine.SPI1.Configure crash

```go
	machine.SPI1.Configure(machine.SPIConfig{
		SCK:       machine.SPI1_SCK_PIN,
		SDO:       machine.SPI1_SDO_PIN,
		SDI:       machine.SPI1_SDI_PIN,
	})
```

jypelle:


src/machine/machine_atsamd51.go
```diff
@@ -1331,12 +1331,13 @@ func (spi SPI) Configure(config SPIConfig) error {
 
 	// Determine the input pinout (for SDI).
 	var dataInPinout uint32
-	SDIPinMode, SDIPad, ok := findPinPadMapping(spi.SERCOM, config.SDI)
+	var SDIPinMode PinMode
```
sago35, line 1334: I think it's a good change.

If we don't use SDI, is it OK for SDIPinMode and dataInPinout to be 0?
Do you know anything about it?

jypelle: For SDIPinMode, it's only used when config.SDI != NoPin so it's ok.

For dataInPinout, I concede I'm not 100% sure. What i can say is that my adafruit pygamer tft screen works seemlessly with this change.

sago35: The proposed amendment to #1408 ([conejoninja@6c39859](https://github.com/conejoninja/tinygo/commit/6c39859ec32b0d28ce3aadd0f0966079fc5cc380)) is almost the same code.
I think it's fine.

sago35 (dismissed):
Looks good to me

sago35:
findPinPadMapping is also called by other functions, such as the UART.
Therefore, it is better to include a crash prevention mechanism in findPinPadMapping().

conejoninja:
I'd have set the variables dataInPinout & SDIPinMode, but as stated in my commit, I didn't know the implications of such change (setting them or not). I strongly recommend to do one way or another, until this is added, displays on the pygamer/pybadge/pyedge/... remain broken.

Linking the opened issue here : https://github.com/tinygo-org/tinygo/issues/1408

deadprogram:
Thanks everyone who has worked on this. Now merging.

src/machine/machine_atsamd51.go
aykevl: Yes I think this is fine. With `dataInPinout` left zero, it will still read data from some arbitrary pin but as long as it is not read that should be fine (see `CTRLA.DIPO` field of the SPI peripheral, [here](http://ww1.microchip.com/downloads/en/DeviceDoc/60001507E.pdf#page=986)).

PR #1487, opened 2020-11-08
ci: update CircleCI, Azure, and Docker builds to Go 1.15.4

deadprogram:
This PR updates the CircleCI build to use Go 1.15.4 now that https://github.com/golang/go/issues/42032 has been resolved.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -426,7 +426,7 @@ jobs:
           llvm: "10"
   test-llvm11-go115:
     docker:
-      - image: circleci/golang:1.15.4-buster
+      - image: circleci/golang:1.15-buster
     steps:
       - test-linux:
           llvm: "11"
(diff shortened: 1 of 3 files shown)


deadprogram:
Once the first commit passed CI, I also added a commit to update Azure and then another to change the `dev` Docker builds to Go 1.15.4

aykevl (commented):
I would actually prefer if this didn't pin to a particular Go minor version. I pinned it to Go 1.15.2 to work around a Go bug but instances like that should be rare. In fact, that bug was a hotfix for a different issue for Go on Android and would normally have gone in a major version, probably catching the issue before the final release.

aykevl:
Azure CI fails, I'm looking into it. It appears that the caching mechanism fails to work.

aykevl:
Or maybe it was a temporary failure, it seems to work now.

PR #1488, opened 2020-11-08
machine/qtpy: add board definition for Adafruit QTPy

deadprogram:
This PR adds the board definition for the Adafruit QTPy board (https://www.adafruit.com/product/4600)

### DIFF
--- a/targets/qtpy.json
+++ b/targets/qtpy.json
@@ -1,0 +1,8 @@
+{
+    "inherits": ["atsamd21e18a"],
+    "build-tags": ["sam", "atsamd21e18a", "qtpy"],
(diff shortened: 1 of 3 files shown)


aykevl:
That's a cute little board!

targets/qtpy.json
```diff
@@ -0,0 +1,8 @@
+{
+    "inherits": ["atsamd21e18a"],
+    "build-tags": ["sam", "atsamd21e18a", "qtpy"],
```
aykevl, line 3: The build tags could be simpler, see #1490 (fixing this for all other SAM boards too).

PR #1489, opened 2020-11-08
ESP8266 tasks scheduler

aykevl:
I have chosed to call this implementation `esp8266` instead of `xtensa` as it has been written specifically for the ESP8266 and there are no other easily available Xtensa chips with the CALL0 ABI (no windowing) that I know of. The only other related chip is the ESP32, which does implement register windowing and thus needs a very different implementation.

The first commit in this PR refactors the stack switching implementation to make it a lot simpler and easier to extend with new architectures. It is the same refactoring commit as in #1445.


aykevl:
@igrr, @andreisfr: can one of you maybe give me a hint as to how I can implement stack switching on the ESP32?

As you can see, I have managed to get stack switching to work on the ESP8266. It wasn't  very difficult as it's just like any other RISC ISA. However, the ESP32 is different. I've read a lot about the register window mechanism but I still don't fully understand it. This includes the (apparently unofficial) Xtensa ISA document that's somewhere on the internet.

I think it should work like this:

1. Make sure all registers are spilled to the stack
2. Save the remaining registers (including WindowBase and WindowStart?).
3. Set the stack pointer of the new stack.
4. Load the registers saved in 2, but from the new stack (of the new task).

You can probably get an idea what it should do by looking at `tinygo_swapTask`. It pushes callee-saved registers on the stack, swaps the stack pointer (while saving the old) and pops callee-saved registers from the stack again. This is the way the scheduler switches over to a task or a task switches over to the scheduler.
Of course, this won't work with the windowing ABI as the unsaved registers are kept across task (goroutine) switches. So if you have any source I can look at that would be great!

igrr:
@aykevl For switching between the stacks, I think it should be sufficient to:

1. save the current window (a0-15) on to the stack, plus any additional registers you need (e.g. floating point registers, threadptr). If you are not switching between task and interrupt context, you don't need to save exception-related registers.
2. flush all the remaining register windows
3. disable window exceptions, reset `WindowStart` to `1 << WindowBase`
4. set new stack pointer
5. restore registers from stack
6. re-enable window exceptions

Steps 3,4,6 can be seen here: https://github.com/espressif/esp-idf/blob/c77c4ccf6c43ab09fd89e7c907bf5cf2a3499e3b/components/xtensa/include/xt_instr_macros.h#L26.

Another higher-level option is to use `setjmp` and `longjmp` for stack switching. These functions perform similar steps internally ([source](https://github.com/espressif/newlib-esp32/blob/esp_based_on_3_0_0/newlib/libc/machine/xtensa/setjmp.S)). Here is an example of switching between the two stacks using setjmp/longjmp: https://github.com/espressif/esp-idf/commit/de662d5e112eae00ca8806c3fbabe8e54584d949#diff-d05b943a7721428370c66db316aa1980573db1bb1b2fcae4f8793e994214d4c9R29-R80

aykevl:
Great, thank you! I was looking for the source of the `setjmp` in the ESP32 ROM, looks like it's from esp32-newlib. The comments there should really help. Unfortunately `setjmp` and `longjmp` do not exactly match what I want (which is more like `swapcontext` or a RTOS task switch), but it's a useful starting point.

One thing that confuses me is the `SYS_nop` system call. I can see how interrupts need to flush register windows but triggering an interrupt just to save registers seems expensive to me.

igrr:
> One thing that confuses me is the SYS_nop system call. I can see how interrupts need to flush register windows but triggering an interrupt just to save registers seems expensive to me.

I think the reason for this implementation is that PS is a privileged register, and user mode code can not directly write to it to perform the above mentioned steps (disable WOE, etc). So triggering a syscall is a relatively portable way that can be used from the standard library, without making assumptions on whether the code will run in user or kernel mode.

aykevl:
Finally, I did it! The solution looks very simple (it certainly is very little code) but so far it seems to work well. See #1494.

@igrr thanks for the help! While I did things a bit differently it really helped to get me on the right track. If you could take a look at that PR (just the last commit) that would be great!

igrr:
Happy that you got it to work @aykevl! I left a comment on the linked PR.

deadprogram:
Thanks @aykevl and @igrr now merging.

PR #1491, opened 2020-11-09
atsamd5x atsamd2x: add a length check to findPinPadMapping

sago35:
findPinPadMapping does not have a length check, which causes a crash.

Related to the following issues and PR
#1483 
#1408

### DIFF
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -136,7 +136,8 @@ var pinPadMapping = [32]byte{
 // found" (indicated by returning ok=false). The pad number is returned to
 // calculate the DOPO/DIPO bitfields of the various serial peripherals.
 func findPinPadMapping(sercom uint8, pin Pin) (pinMode PinMode, pad uint32, ok bool) {
-	if len(pinPadMapping) <= int(pin)/2 {
+	if int(pin)/2 >= len(pinPadMapping) {
+		// This is probably NoPin, for which no mapping is available.
 		return
 	}
 
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -136,7 +136,7 @@ var pinPadMapping = [32]byte{
 // found" (indicated by returning ok=false). The pad number is returned to
 // calculate the DOPO/DIPO bitfields of the various serial peripherals.
 func findPinPadMapping(sercom uint8, pin Pin) (pinMode PinMode, pad uint32, ok bool) {
-	if len(pinPadMapping) <= int(pin) {
+	if len(pinPadMapping) <= int(pin)/2 {
 		return
 	}
 
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -136,6 +136,10 @@ var pinPadMapping = [32]byte{
 // found" (indicated by returning ok=false). The pad number is returned to
 // calculate the DOPO/DIPO bitfields of the various serial peripherals.
 func findPinPadMapping(sercom uint8, pin Pin) (pinMode PinMode, pad uint32, ok bool) {
+	if len(pinPadMapping) <= int(pin) {
+		return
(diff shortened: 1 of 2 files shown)


aykevl:
This should be fixed by #1485. If not, I would consider that a bug elsewhere (not in `findPinPadMapping`).

deadprogram:
Hello @sago35 it appears that this commit has the unintended side effect of disabling the I2C on the Arduino Nano33:

https://github.com/tinygo-org/tinygo/pull/1491/checks?check_run_id=1373182847

sago35:
I don't know why it's failing so far.
I'll try to find out why.

```
    digitalReadVoltage = pass
    digitalReadGround = pass
    digitalWriteOn = pass
    digitalWriteOff = pass
    analogReadVoltage = pass
    analogReadGround = pass
    analogReadHalfVoltage = pass
    i2cConnectionNoPower = pass
    i2cConnectionPower = fail
```

test code:
https://github.com/tinygo-org/tinyhci/blob/master/arduino-nano33/main.go#L269-L280

sago35:
I checked with Xiao and ADT7410, but it did not reproduce.


```go
func i2cConnection() {
	powerpin.Configure(machine.PinConfig{Mode: machine.PinOutput})

	a := adt7410.New(machine.I2C0)
	//a.Configure()
	accel = a

	printtest("i2cConnectionNoPower")

	// should not be connected when not powered
	powerpin.Low()
	if accel.Connected() {
		printtestresult("fail")
	} else {
		printtestresult("pass")
	}

	if true {
		temp := accel.ReadTempC()
	}

	printtest("i2cConnectionPower")
	// turn on power and should be connected now
	powerpin.High()
	time.Sleep(200 * time.Millisecond)

	accel.Configure()
	time.Sleep(500 * time.Millisecond)

	if !accel.Connected() {
		printtestresult("fail")
		return
	}

	printtestresult("pass")

	if true {
		temp := accel.ReadTempC()
	}

}
```

sago35:
The length check was incorrect.
It has been corrected and should be fine.

src/machine/machine_atsamd21.go
```diff
@@ -136,6 +136,10 @@ var pinPadMapping = [32]byte{
 // found" (indicated by returning ok=false). The pad number is returned to
 // calculate the DOPO/DIPO bitfields of the various serial peripherals.
 func findPinPadMapping(sercom uint8, pin Pin) (pinMode PinMode, pad uint32, ok bool) {
+	if len(pinPadMapping) <= int(pin)/2 {
+		return
```
aykevl: I would suggest:

1. Adding a comment explaining why this early return is necessary.
2. (style nitpick) swapping the condition, this is the order that I more commonly see.

```suggestion
	if int(pin)/2 >= len(pinPadMapping) {
		// This is probably NoPin, for which no mapping is available.
		return
```

src/machine/machine_atsamd51.go
```diff
@@ -322,6 +322,10 @@ var pinPadMapping = [64]uint16{
 // (indicated by returning ok=false). The pad number is returned to calculate
 // the DOPO/DIPO bitfields of the various serial peripherals.
 func findPinPadMapping(sercom uint8, pin Pin) (pinMode PinMode, pad uint32, ok bool) {
+	if len(pinPadMapping) <= int(pin)/2 {
+		return
```
aykevl: Same here.

aykevl (commented):
I changed my mind. I think this is a good change, as people might pass `machine.NoPin` or any arbitrary number as a pin number and this is a very simple way to catch those issues and return an error indicating an invalid configuration.
I do have a few style issues though, but they are minor and can perhaps be ignored.

src/machine/machine_atsamd51.go
sago35: updated.

aykevl:
Thank you! Merged.

PR #1494, opened 2020-11-11
Goroutine support on the ESP32

aykevl:
This has been a *lot* of work, trying to understand the Xtensa windowed registers ABI. But in the end I managed to come up with a very simple implementation that so far seems to work very well.

I tested this with both blinky examples (with blinky2 slightly edited) and ./testdata/coroutines.go to verify that it actually works.
Most development happened on the ESP32 QEMU fork from Espressif (https://github.com/espressif/qemu/wiki) but I also verified that it works on a real ESP32.


src/device/esp/esp32.S
```diff
@@ -43,7 +43,7 @@ call_start_cpu0:
     rsync
 
     // Jump to the runtime start function written in Go.
-    j main
+    call4 main
```
aykevl, line 46: This was a bug in the existing ESP32 support: it didn't adhere to the ABI before. Using `call4` makes it adhere to the ABI.

This wasn't a problem for very small programs, but it became a problem once registers needed to be saved to the stack.

igrr:
@aykevl The ESP32 part looks correct to me! Indeed it looks nice and simple.

I see only one possible issue, related to the interrupt handling. When spilling windows, after the `rotw` instruction, the a1 register will no longer contain the stack pointer. This might be an issue for the interrupt handlers implementation, as usually low and medium level interrupt handlers attempt to save the stack frame onto the task's stack. In this case, a1 may contain an arbitrary value from another register window, so saving the stack frame to a1 - FRAME_SIZE will fail.
It is possible to modify the interrupt handlers to be compatible with this, of course. I have looked into this approach once, but haven't found a reliable way to link all the stack frames to each other — so that if the program is stopped in the interrupt handler by the debugger, correct backtrace would be shown.

By the way, since you are already using QEMU for development, you may consider something like https://github.com/wasm3/wasm3/blob/0f599517ae18325a9eb774da8f063de4a9366b6f/.github/workflows/tests.yml#L363-L385 as part of your CI jobs.

aykevl:
Thank you! Yes, interrupts is something I thought about but forgot in the end. I've now fixed it by simply disabling interrupts during the `rotw` instructions.

> By the way, since you are already using QEMU for development, you may consider something like https://github.com/wasm3/wasm3/blob/0f599517ae18325a9eb774da8f063de4a9366b6f/.github/workflows/tests.yml#L363-L385 as part of your CI jobs.

Looks good! I will keep that in mind. Running tests in CI is something I would really like to have eventually once enough of the ESP32 is supported to make that feasible (we already have QEMU based tests for Cortex-M and RISC-V and I hope to have tests for AVR eventually too).

src/internal/task/task_stack_esp32.S
```diff
@@ -0,0 +1,86 @@
...
+    // a3 = oldStack *uintptr
+
+    // Reserve 32 bytes on the stack. It really needs to be 32 bytes, with 16
+    // extra at the bottom to adhere to the ABI.
+    entry sp, 32
+
+    // Disable interrupts while flushing registers. This is necessary because
+    // interrupts might want to use the stack pointer (at a2) which will be some
+    // arbitrary register while registers are flushed.
+    rsil a4, 15
```
igrr: I think this should be XCHAL_EXCM_LEVEL — high level interrupts are not expected to rely on `a1` being the correct stack pointer. Unless you plan to invoke this function from higher-level interrupts — in that case, XCHAL_NUM_INTLEVELS can be used instead.

~~By the way, any specific reason to choose `a4` here?  It may be part of the window which isn't flushed to the stack yet.~~ Edit: sorry, I forgot that you are implementing a solicited context switch (not a exception/preemptive one) so a4 needs only to be flushed, not saved.

aykevl: Okay, thank you! Fixed it.

Right now TinyGo doesn't support Xtensa interrupts so it's a bit irrelevant but it's best to make it safe for when they get supported.

deadprogram:
Fantastic work on this @aykevl and thank you very much @igrr for your review/corrections.

Now merging.

PR #1500, opened 2020-11-16
atsamd21: fix SPI when SDI is set to NoPin

aykevl:
**Untested**

This should fix an issue when SDI is set to `machine.NoPin`. This is already fixed for the samd51, in a basically identical way.

### DIFF
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -1096,11 +1096,15 @@ func (spi SPI) Configure(config SPIConfig) error {
 	}
 
 	// Determine the input pinout (for SDI).
-	SDIPinMode, SDIPad, ok := findPinPadMapping(spi.SERCOM, config.SDI)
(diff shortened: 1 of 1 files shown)


deadprogram:
TinyHCI tests pass, so seems good.

sago35:
The following solves the problem.
So we can close this PR without merging it.

2540172cc5c684772480f895a682591c63dee7bb

sago35:
Correction.

Since the code for samd5x has been changed by the following PR, it is better to change the code for samd2x in the same way.
In other words, it is better to merge this PR.

https://github.com/tinygo-org/tinygo/pull/1485

sago35:
@aykevl 
What do you think?

sago35:
The differences are as follows
image

deadprogram:
Closing due to age. Please open a new PR if this is something important. Thanks everyone!

PR #1501, opened 2020-11-18
WIP: Add support for Maxim MAX32620(fthr)

elliots:
Very wip. 

I doubt I'll be able to make this work, but I'm adding here in case anyone else is interested and has more of a clue.

Whats been done - 
  - copied https://github.com/MaximIntegratedMicros/arduino-max326xx repo into lib (if its needed, should be a submodule instead)
  - generated from the svd in http://www.mxim.net/microcontroller/pack/Maxim.MAX32620.1.0.2.pack (but didnt add to the cmsis-svd repo)
  - added stub `max32620` machine (nothing is set up at all. no gpio, no uart, nothing)
  - added `max32620fthr` board (just the pins)
  - added `max32620fthr` target

There's a test script in `TEST BUILD SCRIPT`. Which I think was a good name.

What happens... 

 - "something" builds, and is copied to the board
 - the board boots, and puts the following in `FAIL.TXT`

```
In application programming write failed.
```

The contents of DETAILS.TXT on the msd, if thats any use.
```
# DAPLink Firmware - see https://mbed.com/daplink
Unique ID: 000000008b808b3200000000000000000000000097969904
HIC ID: 97969904
Auto Reset: 0
Automation allowed: 1
Daplink Mode: Bootloader
Bootloader Version: 0242
Git SHA: 33e3ba1b4a53f190d35bd40421a6c67f9c1842a2
Local Mods: 1
USB Interfaces: MSD
Bootloader CRC: 0x300aa5fd
Interface CRC: 0x671b24c3
```


aykevl:
My first guess would be that if flashing fails, something is wrong with the memory map.

For reference: https://pdfserv.maximintegrated.com/en/an/AN6242.pdf

elliots:
Aha... when i use the memory map from https://github.com/MaximIntegratedMicros/arduino-max326xx/blob/master/arm/system/CMSIS/Device/Maxim/MAX32620/Source/GCC/max32620_bl.ld (which leaves space for the bootloader) instead of https://github.com/MaximIntegratedMicros/arduino-max326xx/blob/master/arm/system/CMSIS/Device/Maxim/MAX32620/Source/GCC/max32620.ld (which does not have the bootloader) it no longer creates the FAIL.TXT.

And it no longer starts the bootloader (with its usb disk device) by default anymore!

I mean, it does nothing at all... but I couldn't expect it to I haven't actually implemented anything to even flash a light.

So it's either progress, or not progress, but at least its something new.

PR #1514, opened 2020-11-27
build: remove release build job for arch release

deadprogram:
This PR removes the CircleCI release build job for arch release until it can be debugged properly.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -344,60 +344,6 @@ commands:
           paths:
             - ~/.cache/go-build
             - /go/pkg/mod
-  arch-release:
-    steps:
-      - run:
-          name: Install dependencies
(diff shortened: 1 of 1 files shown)


aykevl:
If we're going to remove this, wouldn't it be cleaner to remove it entirely? See 149c9533e2d2df5b4c67e4af5034afac56d358a0. We can always re-add it if necessary by reverting this commit, nothing is lost, it's all in the git history.

CC @johanbrandhorst

johanbrandhorst:
Sure, go ahead and remove it. I don't think we ever got to the bottom of the problem, unfortunately, but feel free to rip the whole thing out.

deadprogram:
OK removed the entire set of build steps, and pushed. Now ready for merge.

aykevl:
Thank you!

PR #1517, opened 2020-12-01
nrf: use SPIM peripheral instead of the legacy SPI peripheral

aykevl:
This newer peripheral supports DMA (through EasyDMA) and should generally be faster and more reliable (especially at high speeds). Importantly for some operations: interrupts (within 255 byte buffers) will not interfere with the SPI transfer.

The motivation for this was to be able to control WS2812 LED strips using SPI. Unfortunately it's not possible to fully disable interrupts while the SoftDevice is in use, so WS2812 LEDs would flicker under heavy BLE traffic. The new WS2812-over-SPI code depends on https://github.com/tinygo-org/drivers/pull/208 so will have to wait a bit, but it's able to control 8 LEDs (up to 16, untested) without flickering and without disabling interrupts, which can be useful in some cases.

The nrf51 code is unchanged, only moved around. The nrf52 configuration code has been slightly changed, the main change is the `Transfer` and `Tx` methods.


aykevl:
@deadprogram it appears that two devices are stuck in TinyHCI.

deadprogram:
TinyHCI fixed.

aykevl:
Thank you!

deadprogram:
Tested on itsybitsy-nrf52840 and works as expected. Thanks @aykevl now merging.

PR #1520, opened 2020-12-03
all: switch to LLVM 11 for static builds

aykevl:
This commit switches to LLVM 11 for builds with LLVM linked statically (e.g. `make`). It does not yet switch the default for builds dynamically linked to LLVM (like when you run `go install`), that should be done in a later change.

This commit also changes to use the default host toolchain (normally GCC) instead of Clang as the default compiler in CI. There were some issues with Clang 3.8 in CI and hopefully this will fix it.

This PR is now possible since the LLVM fork from Espressif was updated, see: https://github.com/espressif/llvm-project/issues/40

Perhaps the most important improvement will be a number of changes to the AVR backend that make it much more usable. Another change is a bugfix in the Xtensa backend necessary for proper floating point functioning.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,12 +68,12 @@ commands:
     steps:
       - restore_cache:
           keys:
-            - llvm-source-10-v1
+            - llvm-source-11-v1
       - run:
           name: "Fetch LLVM source"
           command: make llvm-source
       - save_cache:
-          key: llvm-source-10-v1
+          key: llvm-source-11-v1
           paths:
             - llvm-project
   build-wasi-libc:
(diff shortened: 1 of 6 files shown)


aykevl:
Note: still working on this. Windows appears to need some extra work.

aykevl:
> Windows appears to need some extra work.

Let's just say that was an understatement. I still haven't figured it out.

aykevl:
_Finally_, this is ready for review/merge. The secret code was `-DLLVM_ENABLE_PIC=OFF`, using this in fact makes the Makefile a bit simpler (no hack to manually create libclang.a).

deadprogram:
That was some pretty amazingly hard effort to figure that out, good one @aykevl now merging.

PR #1521, opened 2020-12-03
ci: switch to Go 1.15 for MacOS builds

aykevl:
Unfortunately, CircleCI doesn't seem to provide Debian stretch builds
with Go 1.15. We should be using Debian stretch (an older distro) to
make sure the tinygo binary runs on as many Linux systems as possible
(including older ones), and I think using Go 1.14 for these builds is
unfortunate but the better tradeoff.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -274,8 +274,8 @@ commands:
       - run:
           name: "Install dependencies"
           command: |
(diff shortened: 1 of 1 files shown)


deadprogram:
Makes sense. Now merging.

PR #1522, opened 2020-12-03
esp32: get float32/float64 to work correctly

aykevl:
This PR gets float (both float32 and float64) to work, with this pr testdata/float.go can be run on the ESP32.

It depends on #1520 for correct functioning but it should be possible to merge before #1520 is merged (it shouldn't result in a regression). You can test whether it works with the following command:

    tinygo flash -target=esp32-mini32 -port=/dev/ttyUSB0 ./testdata/float.go

EDIT: or maybe it needs to be based on #1520 anyway.

### DIFF
--- a/src/device/esp/esp32.S
+++ b/src/device/esp/esp32.S
@@ -42,6 +42,11 @@ call_start_cpu0:
     wsr.ps a2
     rsync
 
+    // Enable the FPU (coprocessor 0 so the lowest bit).
+    movi a2, 1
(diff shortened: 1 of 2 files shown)


deadprogram:
@aykevl probably just need to rebase this again against dev to pickup your recent fix.

deadprogram:
Confirmed that rebase fixed the tests, so now merging. More good work @aykevl the ESP32 support is getting better and better!

PR #1526, opened 2020-12-08
Nucleo f722ze

kenbell:
I've validated UART and LED functionality.


deadprogram:
Hi @kenbell thank you very much for this contribution. I have a slightly different STM32F7 Nucleo board, hopefully I can help with some testing.

One question I have is why did you create the file machine_stm32_uart-2.go? If the code in there is specific to the stm32f7x then it should probably go into one of the other files. I did not notice what what different at first glance, but I did not go deeply into it.

I am working on stm32f407 support for i2c right now, and I realized there are too many files. Also some duplication I am trying to reduce, but that will be in a future iteration.

kenbell:
@deadprogram ST have changed up the USART - from the SVD, looks like they've re-arranged the registers in memory, and they've also separated read and write registers:

- `RDR` and `TDR` are used to read and write (whereas other design uses `DR` for both)
- `ISR` is the new name for `SR`

This code is just using the new register names - for basic async UART functionality the same logic works.

I think there's a couple options:
1. abstract the register names somehow, so same logic applies, but using chip-specific registers
2. duplicate the code for this new UART design

I don't know all the chips the new design applies to - for now perhaps it makes sense to assume it's all `f7` chips and a `machine_stm32_f7_uart.go`?

deadprogram:
> I don't know all the chips the new design applies to - for now perhaps it makes sense to assume it's all f7 chips and a machine_stm32_f7_uart.go?

Yes, that sounds like a good idea to me. Thanks!

deadprogram:
Thank you for making the changes @kenbell I am going to squash/merge now. Thanks again!

PR #1528, opened 2020-12-10
test: run tests for AVR

aykevl:
This runs the currently working compiler tests for AVR using simavr. I
picked the atmega1284p target as it seems like the most suitable one
(most importantly, with a lot of RAM) so that tests are less likely to
hit resource limitations.

One extra change this required is using a different implementation of
the runtime.abort() function. The new one is smaller and will prevent
interrupts from running on exit, which should at least in theory be
fine. This pattern is required as it signals to simavr the program is
finished.

---

Now that #1520 is ready, I want to work towards better AVR support and dropping the gcc-avr and avr-libc dependencies if possible. One way to track progress is by getting tests to pass, both to show improvements in new PRs and to catch regressions.


niaow:
I am surprised that I missed this when making #2458, but I think this is no longer needed.

aykevl:
Yeah, this PR was way outdated anyway and much simpler now that simavr is widely available.

PR #1529, opened 2020-12-10
matrixportal-m4: Add support for board Adafruit Matrix Portal M4

ardnew:
This PR adds support for a new SAMD51 (with ESP32 coprocessor) board from Adafruit:

- Product: https://www.adafruit.com/product/4745
- Documentation: https://learn.adafruit.com/adafruit-matrixportal-m4

The special feature of this board is its HUB75 connector that lets users plug it directly onto common RGB LED matrix panels.


deadprogram:
Thanks for another coll contribution @ardnew I am going to squash/merge.

Issue #1533, opened 2020-12-15
ADC: Accept configuration struct as argument to method Configure


ardnew:
I think it makes sense to change the API of `ADC.Configure` (for all chips) to the API you're using here. The reason is that an ADC actually has a number of important configurations for different tradeoffs:

 - Reference voltage (often VCC, but sometimes 1.1V or some other value).
 - Accuracy vs time. I know that the nrf chips have this tradeoff (where more accurate sampling takes a lot longer but is necessary if there is a big resistor on the input) and I would guess other ADCs have something similar.

Of course, it's not necessary to implement them all right away but with this I think it makes sense to add an `ADCConfig` to the peripheral.

_Originally posted by @aykevl in https://github.com/tinygo-org/tinygo/pull/1458#discussion_r522904082_

deadprogram:
This has been released with v0.17.0 so now closing. Thanks!

PR #1534, opened 2020-12-15
ADC: Accept configuration struct as argument to method Configure

ardnew:
This PR provides a basic implementation for #1533 

Currently, none of the targets affected make use of the field `Reference` (AREF), although both `atsamd21` and `atsamd51` are capable of using it and will should be updated to handle it in the future.

### DIFF
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -343,7 +343,7 @@ func (a ADC) Configure(config ADCConfig) {
 
 	// Averaging (see datasheet table in AVGCTRL register description)
 	var samples uint32
-	switch config.Resolution {
+	switch config.Samples {
 	case 1:
 		samples = sam.ADC_AVGCTRL_SAMPLENUM_1
 	case 2:
--- a/src/machine/machine_atsamd51.go
+++ b/src/machine/machine_atsamd51.go
@@ -757,7 +757,7 @@ func (a ADC) Configure(config ADCConfig) {
 
 		// Averaging (see datasheet table in AVGCTRL register description)
 		var samples uint32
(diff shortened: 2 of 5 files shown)


deadprogram:
@ardnew can you please resolve the merge conflicts? Thank you.

ardnew:
@deadprogram rebased against `dev`

Looks like there were changes to nRF52 file structure, combining the similar machine methods into `src/machine/machine_nrf528xx`.

@aykevl can you verify I've placed the changes in the correct source file?

src/machine/adc.go
```diff
@@ -0,0 +1,12 @@
+package machine
+
+// Hardware abstraction layer for the analog-to-digital conversion (ADC)
+// peripheral.
+
+// ADCConfig holds ADC configuration parameters. If left unspecified, the zero
+// value of each parameter will use the peripheral's default settings.
```
aykevl: It may be worth adding that not all fields may be supported on all hardware.

aykevl (commented):
The change looks good to me. However, I think it would be a good idea to run the device tests on this, to make sure it doesn't break anything. @deadprogram can you maybe run the TinyHCI tests manually?

deadprogram:
Whoops, I forgot this is a breaking change to the API, so I need to modify the TinyHCI tests in a branch to be able to pass. Working on that.

deadprogram:
OK confirmed that passes TinyHCI with branch for the breaking API change.

Deployed that change to master. I rebased to squash and then merged this PR.

Thanks for the contribution @ardnew

deadprogram:
Perhaps because of the rebase, seems like Github is not recognizing that this PR was merged, so now closing manually.

PR #1538, opened 2020-12-21
Use clang to do linking.

### LINKED ISSUE #1536: 386Linux test fails on non-x86 systems
When building on [aarch64](https://koji.fedoraproject.org/koji/taskinfo?taskID=57500892), this test set fails:
```
/usr/bin/ld: i386 architecture of input file `/tmp/tinygo269308519/main.o' is incompatible with aarch64 output
/usr/bin/ld: i386 architecture of input file `/tmp/tinygo269308519/extra-0-gc_386.S.o' is incompatible with aarch64 output
/usr/bin/ld: /tmp/tinygo269308519/main.o: in function `runtime.printuint64':
/builddir/build/BUILDROOT/tinygo-0.16.0-1.fc34.aarch64/usr/lib64/tinygo/src/runtime/print.go:85: undefined reference to `__udivdi3'
collect2: error: ld returned 1 exit status
```
and many repetitions of the same.

Now, I don't have a cross-compiler or cross-glibc installed, but I also couldn't find anything in `targets/` that specified what to call (if there were anything at all; I'm not sure if linux targets use compilers or linked-to llvm). I'm not sure if I'm missing some compiler, need to patch a binary name to use the right one, or this sort of thing just isn't supported and should be skipped on other arches?

QuLogic:
This avoids having to guess what the linker name-with-build-triplet is, as we can just pass the target to it in the same way as the compilation stage.

Close #1536.

### DIFF
--- a/builder/tools.go
+++ b/builder/tools.go
@@ -38,6 +38,19 @@ func runCCompiler(command string, flags ...string) error {
 
 // link invokes a linker with the given name and flags.
 func link(linker string, flags ...string) error {
+	if hasBuiltinTools && linker == "clang" {
+		// Compile this with the internal Clang compiler.
+		headerPath := getClangHeaderPath(goenv.Get("TINYGOROOT"))
+		if headerPath == "" {
+			return errors.New("could not locate Clang headers")
+		}
+		flags = append(flags, "-I"+headerPath)
+		cmd := exec.Command(os.Args[0], append([]string{"clang"}, flags...)...)
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+		return cmd.Run()
+	}
+
 	if hasBuiltinTools && (linker == "ld.lld" || linker == "wasm-ld") {
 		// Run command with internal linker.
 		cmd := exec.Command(os.Args[0], append([]string{linker}, flags...)...)
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -238,8 +238,9 @@ func defaultTarget(goos, goarch, triple string) (*TargetSpec, error) {
 		GOARCH:    goarch,
 		BuildTags: []string{goos, goarch},
 		Compiler:  "clang",
-		Linker:    "cc",
 		CFlags:    []string{"--target=" + triple},
+		Linker:    "clang",
+		LDFlags:   []string{"--target=" + triple, "-fuse-ld=lld"},
 		GDB:       "gdb",
 		PortReset: "false",
 	}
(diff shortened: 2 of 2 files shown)


aykevl:
I agree this is is an important cleanup. Unfortunately it fails to link, it appears to use the system linker even when cross compiling (which won't work with ld.bfd):

    /usr/bin/ld: unrecognised emulation mode: armelf_linux_eabi
    Supported emulations: elf_x86_64 elf32_x86_64 elf_i386 elf_iamcu elf_l1om elf_k1om i386pep i386pe

I think this can be fixed with an extra linker parameter `-fuse-ld=lld`

I also see this error:

    main.go:803: error: failed to link /tmp/tinygo728791357/main: none of these commands were found in your $PATH: clang-11

And this one, on MacOS:

    main.go:803: error: failed to link /var/folders/yy/d215g62s727f4w2137t9jl2m0000gn/T/tinygo016104012/main: fork/exec /usr/local/opt/llvm@11/bin/clang-11: no such file or directory

I have a suspicion what is causing it, but maybe it's best to first try out the first fix to see how well it works. In short, the name `clang` (which should be treated specially) is not recognized as a linker in builder/tools.go. For that, the functions `runCCompiler` and `link` should probably be merged to a single function.

QuLogic:
I don't recall closing this...

QuLogic:
Thanks for the pointers. As a first step, I just copied the `clang` special case from `runCCompiler` to `link`, and those can be merged together later if it works. Right now, I'm not really sure where these arguments on which CI is failing come from. There doesn't appear to be `-EL`, `--hash-style=both`, or `-demangle` in the tinygo source code.

QuLogic:
Oh, I missed that it was complaining about `-fuse-ld=lld` first, and those arguments are probably coming from `clang` attempting to pass `lld` arguments to something that isn't that. I guess later those CI configurations will need to be modified to install lld instead of the cross-gcc files.

The configurations that actually run appear to still be using `ld.bfd`, perhaps. Maybe this update will work better.

aykevl:
For MacOS, it looks like you're also switching the compiler (from `cc` to `clang`), where you will probably want to use the system linker. As far as I know, `ld64.lld` is not yet very usable (in fact, they're rewriting the linker right now).

And this:

    error: invalid linker name in argument '-fuse-ld=lld'

Is only present when building LLVM statically, so I think something else needs to be done. Apparently Clang has a check to see whether lld is installed, but in this case it should ignore that and try to use it anyway (as lld is built into TinyGo when linking LLVM statically).

deadprogram:
Hello @QuLogic and @aykevl is this PR still active or should it be closed? Thanks.

aykevl:
We've switched a long time ago to invoke LLD directly so that TinyGo won't rely too much on the system configuration (in many cases, not at all). So this is not really an issue today.

PR #1540, opened 2020-12-22
Update to current chromedp.

QuLogic:
They've added a new argument to `WaitFunc`, which I've just decided to ignore, as it seems to work without using it for anything.


aykevl:
Can you please run `go mod tidy` to clean up go.sum?
If that doesn't work, you can try adding an (empty) go.mod file to some directories (src, llvm-project, llvm-build, build, lib). That is a way to make the Go toolchain ignore those directories (they will be considered a new module).

Other than that it looks reasonable. Pinging @bradleypeabody in case he wants to comment on it.

bradleypeabody:
It looks like the latest tag on the chromedp project is [v0.6.4](https://github.com/chromedp/chromedp/tree/v0.6.4).

This PR is [using v0.5.4](https://github.com/tinygo-org/tinygo/pull/1540/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R8). 

@QuLogic Is there any reason not to upgrade to the latest version?  If not, I think that would best.

QuLogic:
> Can you please run `go mod tidy` to clean up go.sum?

I did; it doesn't really ever delete anything.

> @QuLogic Is there any reason not to upgrade to the latest version? If not, I think that would best.

It was simply not the latest version at the time.

bradleypeabody:
Cool, LGTM.

I suspect the extra go.sum entries are due to this: https://github.com/golang/go/issues/33008. I suggest not worrying about it for now and let it get cleaned up later with Go 1.16.

deadprogram:
Thanks for this contribution @QuLogic and thanks @bradleypeabody for reviewing.

Now merging.

PR #1542, opened 2020-12-23
Use httptest to serve wasm test files.

QuLogic:
This picks a port automatically, so avoids any conflicts that might arise from running the tests in parallel.

### DIFF
--- a/tests/wasm/chan_test.go
+++ b/tests/wasm/chan_test.go
@@ -11,6 +11,9 @@ func TestChan(t *testing.T) {
 
 	t.Parallel()
 
+	wasmTmpDir, server, cleanup := startServer(t)
+	defer cleanup()
+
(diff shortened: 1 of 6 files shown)


QuLogic:
Ah, looks like `testing.TempDir` was added in 1.15 only.

deadprogram:
@bradleypeabody or anyone else have any feedback on this PR?

bradleypeabody:
Seems fine to me.  Earlier I made an attempt to get some of these tests running on the older Go versions but for various reasons it didn't work.  I don't see a build tag at the top of setup_test.go, so I'm a bit surprised the tests for 1.12 and 1.13 don't fail on that file - the wasm test stuff must be disabled in some other way on those versions.

That said, since targeting wasm tests at <1.14 seems unrealistic at this point (and more unnecessary as time goes on), I'm good with this change.

deadprogram:
Thanks @QuLogic  for working on this and to @bradleypeabody for reviewing.

Now merging.

PR #1543, opened 2020-12-23
compiler: fix non-int integer constants

aykevl:
Before this change, the compiler could panic with the following message:

    panic: 20 not an Int

That of course doesn't make much sense. But it apparently is expected behavior, see https://github.com/golang/go/issues/43165 for details.

This commit fixes this issue by converting the constant to an integer if needed.


deadprogram:
For sure an odd feature. Merging for the sake of language compatibility.

PR #1547, opened 2020-12-26
interp: fix alignment of untyped globals

aykevl:
During a run of interp, some memory (for example, memory allocated
through runtime.alloc) may not have a known LLVM type. This memory is
alllocated by creating an i8 array.
This does not necessarily work, as i8 has no alignment requirements
while the allocated object may have allocation requirements. Therefore,
the resulting global may have an alignment that is too loose.
This works on some microcontrollers but notably does not work on a
Cortex-M0 or Cortex-M0+, as all load/store operations must be aligned.

This commit fixes this by setting the alignment of untyped memory to the
maximum alignment. The determination of "maximum alignment" is not
great but should get the job done on most architectures.

---

This PR fixes an issue with #1430.

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -21,6 +21,8 @@ type runner struct {
 	targetData    llvm.TargetData
 	builder       llvm.Builder
 	pointerSize   uint32                   // cached pointer size from the TargetData
+	i8ptrType     llvm.Type                // often used type so created in advance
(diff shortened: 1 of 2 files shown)


aykevl:
@deadprogram you might want to restart some of the TinyHCI devices, many appear to be locked up.

deadprogram:
The itsybitsy-nrf52840 hci failure is not related to this fix. The other boards are all passing again.

Now merging, thanks for the quick fix @aykevl !

PR #1548, opened 2020-12-28
machine/nrf52840: ensure that USB CDC interface is only initialized once

deadprogram:
This PR is to ensure that the USB CDC interface is only initialized once. This avoids a lockup when performing a soft reset, such as when using the UF2 bootloader.


aykevl:
Looks good to me!

I think most/all these objects (such as `USB`, `UART0`, `SPI0`, etc) should eventually be pointers to avoid issues with accidentally copying them.

PR #1549, opened 2020-12-28
machine/clue: correct volume name and add alias for release version of Adafruit Clue

deadprogram:
This PR corrects the volume name and adds an alias for release version of Adafruit Clue board so we can use `-target=clue`.


deadprogram:
Since I am the only one around who currently has a Clue, I am going to merge.

PR #1550, opened 2020-12-29
Use rust SVD library for STM32

kenbell:
Have validated using `make smoketest`.

Had to adjust `gen-device-svd` to put a 'read', 'write', 'read-write' suffix when needed, because of this patch:
https://github.com/stm32-rs/stm32-rs/blob/master/peripherals/wwdg/wwdg.yaml

This was resulting in two identical constant declarations for 'Finished' in the Go source.  Solved by taking into account this 'usage' field which is set to 'read', 'write' or 'read-write' per the SVD XML schema.

The suffix is only added when there are multiple 'enumeratedValues' sections, indicating there is potential for a naming collision.  Otherwise suffix is not added so as to avoid gratuitous changes to symbolic names.

This change requires some environment changes to generate the device/stm32 source:

- python3 to be installed
- svdtools to be installed (`pip3 install --upgrade --user svdtools`)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -20,3 +20,6 @@
 [submodule "lib/picolibc"]
 	path = lib/picolibc
 	url = https://github.com/keith-packard/picolibc.git
+[submodule "lib/stm32-rs"]
+	path = lib/stm32-rs
+	url = https://github.com/stm32-rs/stm32-rs.git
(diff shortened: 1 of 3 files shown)


kenbell:
Closing as better approach available in https://github.com/tinygo-org/tinygo/pull/1557

PR #1525, opened 2020-12-07
fix GBA ROM header

mogenson:
Populate the GBA ROM header so that emulators and physical Game Boy Advance consoles recognize the ROM as a valid game.

This PR removes the need for post-processing of a GBA ROM by a tool like `gbafix`. It's based on [this commit](https://github.com/rust-console/gba/commit/016d9a577db6aeff0354a6d55da3fb9a57891b24) to the [rust-console/gba](https://github.com/rust-console/gba) project: 

Testing:
```
$ tinygo build -target=gameboy-advance -o gba-display.gba examples/gba-display
$ file gba-display.gba
gba-display.gba: Game Boy Advance ROM image (, Rev.00)
$ mgba gba-display.gba # also copy rom to flashcart
```

Note: The reserve space at the end of the header had to be hand-tuned to this specific value of 98 bytes. Can anyone here tell me why? Something to do with word boundaries for ARM vs Thumb mode? Padding with 158 bytes also worked, but anything less than 98 bytes failed to boot.

PR #1558, opened 2021-01-04
Add Nintendo Logo bytes to gameboy-advance target

trevor403:
This will create ROMs that can function on real hardware and emulators with ROM Header validation.

A side effect of this issue can be observed here https://github.com/tinygo-org/tinygo/issues/894#issuecomment-737578880


aykevl:
I have tested this but at least for me it doesn't fix the crash.

```
$ git checkout trevor403/patch-1
Previous HEAD position was 2f4200a0 Add support for additional openocd commands (#1492)
HEAD is now at 4a82e889 Add Nintendo Logo bytes to gameboy-advance target

$ tinygo build -o test.gba -target=gameboy-advance examples/gba-display
ld.lld-10: warning: lld uses blx instruction, no object with architecture supporting feature detected

$ mgba test.gba
The game crashed!

$ mgba --version
mgba 0.7.0 (0ddade3)
```

Also see the discussion at https://github.com/tinygo-org/tinygo/pull/1525.

racerxdl:
btw @aykevl that patch does fix running on real hardware though:
image

QuLogic:
This seems to have been replaced by #1525, I think?

PR #1561, opened 2021-01-05
Support for STM32L0 MCUs and Dragino LGT92 device

ofauchon:
This PR provides support for : 

- STM32L0 microcontroller series. 
- Dragino LGT92 device (which is based on stm32l072czt6  )

Working:  

 - LPUART and UART
 - GPIOs
 - SPI
 - Timers

Not working yet: 

- i2c
- power saving (LP modes)


src/machine/board_lgt92.go
```diff
@@ -0,0 +1,107 @@
...
+
+package machine
+
+import (
+	"device/stm32"
+	"runtime/interrupt"
+)
+
+const (
+	LED1 = PA12
```
aykevl, line 11: Can you also add this?

```go
	LED = LED1
```

Many examples use `machine.LED` for the first (or only) LED available.

src/machine/board_lgt92.go
```diff
@@ -0,0 +1,107 @@
...
+	stm32.RCC.AHBENR.SetBits(stm32.RCC_AHBENR_MIFEN)
+
+	// Enable Leds GPIO output
+	LED_RED.Configure(PinConfig{Mode: PinOutput})
+	LED_GREEN.Configure(PinConfig{Mode: PinOutput})
+	LED_BLUE.Configure(PinConfig{Mode: PinOutput})
+
+	LED_RED.Low()
+	LED_GREEN.Low()
+	LED_BLUE.Low()
```
aykevl: Why are you doing this in the board file?

- Configuring outputs needs to be done by the program, unless it's really necessary they should be configured by a program that wants to use these LEDs.
- Is there any support for EEPROM? If not, I would suggest leaving it disabled (enabling unused peripherals may cost power).
- If you need to enable the GPIO peripheral (`stm32.RCC.IOPENR.SetBits(stm32.RCC_IOPENR_IOPAEN)`), this should be done in `Pin.Configure`. It looks like this is already done, so these lines are unnecessary.

src/machine/machine_stm32l0.go
```diff
@@ -0,0 +1,193 @@
...
+	case 2:
+		return stm32.GPIOC
+	case 3:
+		return stm32.GPIOD
+	case 4:
+		return stm32.GPIOE
+		//	case 5:
+		//		return stm32.GPIOF
+		//	case 6:
+		//		return stm32.GPIOG
```
aykevl: If `GPIOF` etc. don't exist, you don't need to include them here. If you want to explain why you left them out, you could add a comment saying "GPIOF etc. do not exist on this chip".

src/machine/machine_stm32l0.go
```diff
@@ -0,0 +1,193 @@
...
+	case unsafe.Pointer(stm32.SPI1): // SPI1 clock enable
+		stm32.RCC.APB2ENR.SetBits(stm32.RCC_APB2ENR_SPI1EN)
+	case unsafe.Pointer(stm32.ADC): // ADC clock enable
+		stm32.RCC.APB2ENR.SetBits(stm32.RCC_APB2ENR_ADCEN)
+	case unsafe.Pointer(stm32.USART1): // USART1 clock enable
+		stm32.RCC.APB2ENR.SetBits(stm32.RCC_APB2ENR_USART1EN)
+		//	case unsafe.Pointer(stm32.TIM8): // TIM8 clock enable
+		//		stm32.RCC.APB2ENR.SetBits(stm32.RCC_APB2ENR_TIM8EN)
+		//	case unsafe.Pointer(stm32.TIM1): // TIM1 clock enable
+		//		stm32.RCC.APB2ENR.SetBits(stm32.RCC_APB)
```
aykevl: This is again code that should be removed: TIM1 and TIM8 do not seem to exist on this chip.

src/machine/machine_stm32l0x2.go
```diff
@@ -0,0 +1,108 @@
...
+		localFrequency = localFrequency * 2
+	}
+
+	// set frequency dependent on PCLK prescaler. Since these are rather weird
+	// speeds due to the CPU freqency, pick a range up to that frquency for
+	// clients to use more human-understandable numbers, e.g. nearest 100KHz
+
+	// These are based on APB2 clock frquency (84MHz on the discovery board)
+	// TODO: also include the MCU/APB clock setting in the equation
+	switch true {
```
aykevl: You can leave out the `true` here, a switch can also work as a replacement for a long `if` / `else if` / `else` chain.

```suggestion
	switch {
```

src/machine/machine_stm32l0x2.go
```diff
@@ -0,0 +1,108 @@
...
+		conf = stm32.SPI_PCLK_8
+		// NOTE: many SPI components won't operate reliably (or at all) above 10MHz
+		// Check the datasheet of the part
+	case localFrequency < 21000000:
+		conf = stm32.SPI_PCLK_4
+	case localFrequency < 42000000:
+		conf = stm32.SPI_PCLK_2
+	default:
+		// None of the specific baudrates were selected; choose the lowest speed
+		conf = stm32.SPI_PCLK_256
```
aykevl: Ideally we'd default to 4MHz, but if that is not possible you could use a value just below that.
You can add such a check at the top by checking for 0:

```go
if config.Frequency == 0 {
    config.Frequency = 4e6
}
```

src/machine/machine_stm32l0x2.go
```diff
@@ -0,0 +1,108 @@
...
+
+	// set frequency dependent on PCLK prescaler. Since these are rather weird
+	// speeds due to the CPU freqency, pick a range up to that frquency for
+	// clients to use more human-understandable numbers, e.g. nearest 100KHz
+
+	// These are based on APB2 clock frquency (84MHz on the discovery board)
+	// TODO: also include the MCU/APB clock setting in the equation
+	switch true {
+	case localFrequency < 328125:
+		conf = stm32.SPI_PCLK_256
```
aykevl: It would be useful to make this calculation explicit. For example:

```go
const apb2Frequency = 84e6

// ...

switch {
case localFrequency >= apb2Frequency / 2:
    conf = stm32.SPI_PCLK_2
case localFrequency >= apb2Frequency / 4:
    conf = stm32.SPI_PCLK_4
// ...
case localFrequency >= apb2Frequency / 128:
    conf = stm32.SPI_PCLK_128
default:
    conf = stm32.SPI_PCLK_256
}
```

This way the calculation is very explicit. This kind of calculation is always done at compile time, so you don't need to worry about any overhead it might introduce.
You can take a look here as an example: https://github.com/tinygo-org/tinygo/pull/1446

targets/stm32l0x2.json
```diff
@@ -0,0 +1,22 @@
...
+    "cflags": [
+        "--target=armv6m-none-eabi",
+        "-Qunused-arguments"
+    ],
+    "linkerscript": "targets/stm32l072czt6.ld",
+    "extra-files": [
+        "src/device/stm32/stm32l0x2.s"
+    ],
+    "flash-method": "openocd",
+    "openocd-interface": "stlink-v2",
```
aykevl:
```suggestion
```

These two lines are specific to the board, so shouldn't be in the chip specification.

targets/stm32l0x2.json
```diff
@@ -0,0 +1,22 @@
...
+    "build-tags": [
+        "stm32l0",
+        "stm32l0x2",
+        "stm32"
+    ],
+    "cflags": [
+        "--target=armv6m-none-eabi",
+        "-Qunused-arguments"
+    ],
+    "linkerscript": "targets/stm32l072czt6.ld",
```
aykevl:
```suggestion
```

Because this linker script is for a more specific chip (with a given amount of memory), it shouldn't be included in the more generic chip family specification.

aykevl (commented):
Thank you, looks good to me! I have a number of code cleanliness issues but nothing major.

aykevl:
And before I forget: can you also add a smoke test to the Makefile? You can see a list of commands there already to run the smoke tests.

src/machine/board_lgt92.go
ofauchon: Done

ofauchon:
Hi @aykevl,

Thanks for your detailed review, and all the suggestions. 
I think I made all the proposed changes. 

Please tell me if you have other improvements.

Olivier

aykevl (approved):
Thank you for the fixes! Looks good to me now.

deadprogram:
I will squash/merge this PR now. Thank you very much for the contribution @ofauchon

PR #1562, opened 2021-01-05
README: fix bluepill link

wuhanstudio:
Hi, it seems like the STM32Duino wiki url is invalid, because the wiki has moved to a new site.

[https://stm32duinoforum.com/forum/wiki_subdomain/index_title_Blue_Pill.html](https://stm32duinoforum.com/forum/wiki_subdomain/index_title_Blue_Pill.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ The following 32 microcontroller boards are currently supported:
 * [PineTime DevKit](https://www.pine64.org/pinetime/)
 * [SiFIve HiFive1](https://www.sifive.com/boards/hifive1)
 * [ST Micro "Nucleo F103RB"](https://www.st.com/en/evaluation-tools/nucleo-f103rb.html)
-* [ST Micro STM32F103XX "Bluepill"](http://wiki.stm32duino.com/index.php?title=Blue_Pill)
+* [ST Micro STM32F103XX "Bluepill"](https://stm32-base.org/boards/STM32F103C8T6-Blue-Pill)
 * [ST Micro STM32F407 "Discovery"](https://www.st.com/en/evaluation-tools/stm32f4discovery.html)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -74,7 +74,7 @@ The following 32 microcontroller boards are currently supported:
 * [PineTime DevKit](https://www.pine64.org/pinetime/)
 * [SiFIve HiFive1](https://www.sifive.com/boards/hifive1)
 * [ST Micro "Nucleo F103RB"](https://www.st.com/en/evaluation-tools/nucleo-f103rb.html)
-* [ST Micro STM32F103XX "Bluepill"](http://wiki.stm32duino.com/index.php?title=Blue_Pill)
+* [ST Micro STM32F103XX "Bluepill"](https://stm32duinoforum.com/forum/wiki_subdomain/index_title_Blue_Pill.html)
```
aykevl: The wiki has not moved. The original wiki has gone down and someone made a read-only copy.

In this case, I think it is better to link to the archived copy of this webpage: https://web.archive.org/web/20190527040051/http://wiki.stm32duino.com/index.php?title=Blue_Pill
The archive.org copy is much more likely to stay up and makes it clear it's a read-only copy. If this wiki becomes alive again, we can adjust the URL again to the live version.

wuhanstudio: Hi, indeed it's not proper to use a read-only copy.

What do you think of this one, it's from stm32duino offical website for bluepill:

[https://stm32-base.org/boards/STM32F103C8T6-Blue-Pill](https://stm32-base.org/boards/STM32F103C8T6-Blue-Pill)

[https://github.com/stm32duino/Arduino_Core_STM32/blob/master/README.md#generic-stm32f1-boards](https://github.com/stm32duino/Arduino_Core_STM32/blob/master/README.md#generic-stm32f1-boards)

aykevl: Yes, that URL also looks good to me.

PR #1564, opened 2021-01-06
stacksize: add support for DW_CFA_offset_extended

aykevl:
It should be possible to ignore this directive, but we still have to consume the two operands.

@firelizzard18 can you test this change with the Teensy 3.6?


deadprogram:
Works on my Teensy3.6 with blinky1 as expected, so now merging. Thanks @aykevl

PR #1563, opened 2021-01-05
Add support for the P1AM-100 (similar to Arduino MKR)

quentinmit:
`tinygo flash -target=p1am-100 examples/blinky1` works unmodified. I haven't fully exercised all the peripherals yet. I'm not sure if you have a test suite I can try running to exercise all the peripherals, but if you do please LMK. I plan to test the SPI interface next, since this board has SPI peripherals built-in.

PR #1565, opened 2021-01-06
machine/p1am-100: rename MISO/MOSI

quentinmit:
This makes p1am-100 work again after commit d1c4ed6.

See #1563


deadprogram:
Thanks @quentinmit for the quick update. Now merging.

Issue #1535, opened 2020-12-16
wasm memory bug/question


yuval-k:
Hello everyone! 
Than you for making tinygo wasm compatible. I've been testing it out, trying to use with envoy proxy, and I have some questions I would love your help with. I tried answering these myself by looking at the code, but I want to make sure I'm not missing anything:

I noticed that using different arguments to `--heap-size` still produces the exact same binary, with initial wasm memory set to 3 wasm-pages (i.e.: `(memory $0 3)`) I do see this argument passed in the ld flags, so I suspect that this is a bug? 

IIUC the code correctly, the initial memory is all that is available - i.e. the `memory.grow` instruction is not used to grow memory dynamically. This means that all the memory usage expected should be included in the heap-size. is that accurate?

Thank you for your reply!

asayah:
@deadprogram any update on this one?

djannot:
+1

aykevl:
Hi! Thank you for your interest in using WebAssembly. I have made a PR that implements automatically growing the heap when needed: #1567. This removes the `-heap-size` parameter entirely as it is not necessary anymore, which I believe is the proper fix (`-heap-size` was always just a workaround as the TinyGo heap previously was unable to grow in size).

djannot:
@aykevl do you know when this fix will be released ?

deadprogram:
This has been released with v0.17.0 so now closing. Thanks!

Issue #1327, opened 2020-08-29
[runtime error: out of memory] due to repeated call to demo splitter method


Kusers:
Error screenshot
https://github.com/Kusers/tinygoDemo/blob/master/runerr.png
javascript code
https://github.com/Kusers/tinygoDemo/blob/master/run_err_js.txt
How can we solve this problem? I look forward to your reply. Thank you very much

aykevl:
Passing data between JavaScript and WebAssembly (Go) currently leaks memory. You can slow it down a bit with the `-heap-size` parameter.

Kusers:
If I use C language, will it still happen?

deadprogram:
@Kusers this is a really a general problem with memory management and any WASM code that wants to pass data to/from JavaScript, regardless of what was used to compile it.

FrankReh:
The memory leak caused by having no finalizers called? I've run into that today.

Does getting a gc for wasm to understand objects that are being freed, and keeping track of their finalizers, require knowledge of the LLVM side of things or should it be enough to know the Go and JavaScript side of things?

I'm not even sure the conservative gc is being used for the wasm target. I saw a question about it early last year but the answer was that channels and reflection were first on the TODO list.

I don't know what the setup time is for a wasm instantiation but is a viable alternative to let the entire instance be collected by the JavaScript VM and start a new wasm instance when memory gets low?

deadprogram:
This has been released with v0.17.0 so now closing. Thanks!

Issue #460, opened 2019-07-26
Misleading option `-heap-size` for wasm


lei-april:
This options was introduced by commit d627208c48501e649a0c2a477f77556af1e35a99. From what I see, it actually configures size of whole wasm linear memory, not just the "heap".

As regard to wasm bytecodes generated by the LLVM backend, the linear memory consists of three sections: an explicit stack, the data section and the heap. 

From source of wasm-ld:
```cpp
// The default memory layout is as follows, from low to high.
//
//  - initialized data (starting at Config->GlobalBase)
//  - BSS data (not currently implemented in llvm)
//  - explicit stack (Config->ZStackSize)
//  - heap start / unallocated
```

However, the argument of option `-heap-size` seems to apply to the whole linery memory:
```go
if spec.GOARCH == "wasm" {
    // Round heap size to next multiple of 65536 (the WebAssembly page
    // size).
    heapSize := (config.heapSize + (65536 - 1)) &^ (65536 - 1)
    ldflags = append(ldflags, "--initial-memory="+strconv.FormatInt(heapSize, 10))
}
```

So I think it's a little confusing to use the phrase "heap size" here.

deadprogram:
Perhaps this should be renamed to something like `memory-size` to more accurately describe what it does. @aykevl what do you think?

aykevl:
The original intent was that `-heap-size` would also configure the heap size on *nix systems (using `malloc` to allocate the heap on startup). However, this turned out to be a bit more tricky to implement than I originally expected so I left it out.

Changing to `-memory-size` sounds good.

aykevl:
#1567 sidesteps this issue by removing this flag entirely.

deadprogram:
This has been released with v0.17.0 so now closing. Thanks!

PR #1567, opened 2021-01-09
Automatically grow the heap on WebAssembly

aykevl:
This commit makes the heap growable on WebAssembly. The heap design is not fundamentally changed although the heap layout is changed. This has been something I wanted to implement long ago but I never got around doing it. In the end, it took me a lot less time than I expected.

This should have very little impact on baremetal targets. Firmware size increased slightly but while working on this I also found an optimization in the heap to make things easier for the optimizer so that firmware size is in fact usually reduced a little bit with this PR on baremetal targets. The WebAssembly output has grown a little bit of course, with this new ability to grow the heap when it is full.

This fixes https://github.com/tinygo-org/tinygo/issues/1535 and https://github.com/tinygo-org/tinygo/issues/460 and improves https://github.com/tinygo-org/tinygo/issues/1327

### DIFF
--- a/src/runtime/arch_wasm.go
+++ b/src/runtime/arch_wasm.go
@@ -17,6 +17,9 @@ var heapStartSymbol [0]byte
 //export llvm.wasm.memory.size.i32
 func wasm_memory_size(index int32) int32
 
+//export llvm.wasm.memory.grow.i32
+func wasm_memory_grow(index int32, delta int32) int32
+
 var (
 	heapStart = uintptr(unsafe.Pointer(&heapStartSymbol))
 	heapEnd   = uintptr(wasm_memory_size(0) * wasmPageSize)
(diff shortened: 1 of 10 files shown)


src/runtime/arch_wasm.go
```diff
@@ -30,3 +33,32 @@ func align(ptr uintptr) uintptr {
...
+
+	// Increase the heap. After setting the new heapEnd, calculateHeapAddresses
+	// will update metadataStart and the memcpy will copy the metadata to the
+	// new location.
+	// The new metadata will be bigger than the old metadata, but a simple
+	// memcpy is fine as it only copies the old metadata and the new memory will
+	// have been zero initialized.
+	heapEnd = uintptr(wasm_memory_size(0) * wasmPageSize)
+	calculateHeapAddresses()
+	memcpy(unsafe.Pointer(metadataStart), unsafe.Pointer(oldMetadataStart), oldMetadataSize)
```
niaow: This section should probably live inside of `gc_conservative.go`, as this references variables from a separate build tag. We could maybe have something like `resizeHeap` (I am not doing great with names right now) which would sit inside of the GC implementations. If the GC did not support this, it could avoid calling `growHeap` and implement `resizeHeap` as a no-op. Ideally, `leaking`/similar should still work.

deadprogram: That seems like a very reasonable suggestion to me. :+1:

aykevl: Yes, that makes sense. I have updated the PR with this change (with `setHeapEnd`) that is also supported with `-gc=none` and `-gc=leaking` (it doesn't make sense with `-gc=extalloc`).

niaow (changes requested):
It looks like this just needs a gofmt on `src/runtime/gc_conservative.go` to fix CI, and then it should be good.

aykevl:
Oops, thanks for letting me know! I've fixed it.
My VSCode was having some trouble and didn't automatically format source files for some reason.

niaow (approved):

deadprogram:

PR #1569, opened 2021-01-12
Examples for Dragino lgt92 device

ofauchon:
- Fix bad openocd-target definition in lgt92.json
- Example applications for Dragino LGT92 device


ofauchon:
These examples are  very specific to lgt92 boards.
Not sure they should be added to tinygo project. 

https://github.com/tinygo-org/tinygo-zoo  may be a better place for them. 

I'm closing this PR and I'll try to re-submit them in the zoo repository

PR #1571, opened 2021-01-13
Add tests to the compiler package

aykevl:
This is my next attempt to make the compiler more modular (also see #1008). This time I managed to add tests to the compiler without too many complications.

Testing is still quite limited as it isn't yet possible to test anything that refers to runtime types or functions, which are the most interesting tests. ~I did this in #1008~ but left it out in this PR to keep things simple and focused. I intend to introduce runtime types in a later PR and with that change, add more compiler tests.

The first commit of this PR is a small change that renames the main package in the IR. This could have real world impact but I think it is an improvement as output binaries now better match the symbol name conventions of the Go toolchain.

Updates #285.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -114,7 +114,7 @@ commands:
           key: wasi-libc-sysroot-systemclang-v2
           paths:
             - lib/wasi-libc/sysroot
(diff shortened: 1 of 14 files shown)


niaow (approved):

deadprogram:
Thanks @aykevl for the changes and to @niaow for reviewing. Now merging.

PR #1572, opened 2021-01-13
New tinygo -x option to print commands 

ofauchon:
New -x tinygo flag to print external commands (ex: flasher, emulator, debugger commands)

### DIFF
--- a/main.go
+++ b/main.go
@@ -33,7 +33,6 @@ var (
 	// This variable is set at build time using -ldflags parameters.
 	// See: https://stackoverflow.com/a/11355611
 	gitSha1 string
-	options *compileopts.Options
 )
 
 // commandError is an error type to wrap os/exec.Command errors. This provides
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -33,6 +33,7 @@ var (
 	// This variable is set at build time using -ldflags parameters.
 	// See: https://stackoverflow.com/a/11355611
 	gitSha1 string
+	options *compileopts.Options
```
aykevl: Globals are a bad idea. Can you pass this options object manually to all `executeCommand` calls? You can usually use `config.Options` for that.

ofauchon: You are right, globals should be avoided....
But I thought it was tolerated when data is  initialized once and reused multiple times (which is the case of config object ?)

However, I'll follow your advice and remove this global .

deadprogram:
Thanks for the contribution and the changes @ofauchon now squash/merging.

PR #1574, opened 2021-01-15
compiler: saturate float-to-int conversions

### LINKED ISSUE #276: Float to integer conversion may result in undefined behavior
The Go spec says this about conversions:

> In all non-constant conversions involving floating-point or complex values, if the result type cannot represent the value the conversion succeeds but the result value is implementation-dependent.

In other words, all conversions (including float-to-integer) must be implementation defined and not undefined. At the moment, we use LLVM [`fptui`](https://llvm.org/docs/LangRef.html#fptoui-to-instruction) and [`fptosi`](https://llvm.org/docs/LangRef.html#fptosi-to-instruction) instructions, which have this to say on the matter:

> The ‘`fptoui`’ instruction converts its floating-point operand into the nearest (rounding towards zero) unsigned integer value. If the value cannot fit in ty2, the result is a poison value.

> The ‘`fptosi`’ instruction converts its floating-point operand into the nearest (rounding towards zero) signed integer value. If the value cannot fit in ty2, the result is a poison value.

This basically means that floating point values that cannot be represented as integers (Infinity, NaN, negative numbers in unsigned integers) will produce an arbitrary value that if used may invoke undefined behavior.

This should be fixed. It is probably easiest to wait until https://reviews.llvm.org/D54749 is merged to avoid bloating the code too much.

Similar issue in Rust: https://github.com/rust-lang/rust/issues/10184

niaow:
This works around some UB in LLVM, where an out-of-bounds conversion would produce a poison value.
The selected behavior is saturating, except that NaN is mapped to the minimum value.

Fixes #276


aykevl:
The code itself looks reasonable, although I haven't studied it in depth.

Can you maybe add a few more test cases? I'm a little bit nervous around edge cases with floats, so for example I think it would be a good idea to test the values around the edges. You can look here for some examples: https://github.com/rust-lang/rust/blob/19e63bd50eb756e35416d7a8f01e2d63f419b373/src/test/run-pass/saturating-float-casts.rs

aykevl:
Also, I'm not sure whether this was intentional but the behavior in this PR differs slightly from the LLVM builtin, which we'll probably want to use eventually: https://reviews.llvm.org/D54749. The builtin converts NaN to 0 instead of the minimal integer value. Using the same behavior would help with testing when we can use that builtin instead.

niaow:
> The builtin converts NaN to 0 instead of the minimal integer value.

Yeah, this was something I noticed. I had omitted the extra step to do this for signed integers as it seemed to be a bit of extra complexity for now, but I can add it in.

aykevl:
I'm fine with it as it is. It just might need changing in the future.
The rationale for `NaN` to 0 conversion according to Rust is the following ([source](https://internals.rust-lang.org/t/help-us-benchmark-saturating-float-casts/6231)):

> * Matches the behaviour of ARM’s `vcvt` instruction, making this have no overhead there.
> * Should optimize reasonably well on other platforms.

I don't know whether softfloat implementations have the same behavior and whether LLVM optimizes this pattern, but at least in theory this pattern is cheap/free on ARM.

deadprogram:
Thank you @niaow for the contribution and @aykevl for the review. Now merging.

niaow:
This was. . . not ready to be merged

I guess I will open a new PR to actually finish this.

deadprogram:
Oops, my bad. I misinterpreted the comments.

Issue #1575, opened 2021-01-15
compile: slicing an array does not produce a nil check


niaow:
This behavior is easily demonstrated here:
```
package main

func main() {
        var x *[4]uint32
        println(x[:][0])
}
```

Expected behavior (`go run`):
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x45dc9f]

goroutine 1 [running]:
main.main()
        /home/niaow/bad.go:5 +0x1f
exit status 2
```

Actual behavior (`tinygo run`):
```
0
```

deadprogram:
This has been released with v0.17.0 so now closing. Thanks!

PR #1576, opened 2021-01-15
compiler: use LLVM memory move intrinsics for slice copy

niaow:
I am experimenting to see whether copy can be done more efficiently, blocked by #1575


aykevl:
`runtime.sliceCopy` already uses the `llvm.memmove` builtin.
Is there any other advantage I have overlooked?

niaow:
I don't really know. I don't think the compiler can really do much without inlining `sliceCopy`, although I guess this is effectively inlining. I kinda just wanted to try this out and see what happens. It seems to be finding bugs I guess.

aykevl:
I see.

I think if inlining makes it smaller/faster, a `//go:inline` pragma may be appropriate. In any case, if possible I would very much prefer such things to be done through runtime calls rather than doing it directly in the compiler as Go code is much easier to maintain.

PR #1577, opened 2021-01-15
compiler: emit a nil check when slicing an array pointer

### LINKED ISSUE #1575: compile: slicing an array does not produce a nil check

niaow:
Fixes #1575 

Once the compiler testing gets a bit further we can probably add some tests to verify that assertions are inserted correctly.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -1852,6 +1852,7 @@ func (b *builder) createExpr(expr ssa.Value) (llvm.Value, error) {
 				low,
 			}
 
+			b.createNilCheck(expr.X, value, "slice")
 			b.createSliceBoundsCheck(llvmLen, low, high, max, lowType, highType, maxType)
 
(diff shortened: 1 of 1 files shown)


aykevl:
Looks reasonable, thank you!

I believe the same code is used when slicing a regular array, hence why I never added a nil check. But in the case of a regular array the check _should_ be optimized away.
This might be worth testing however.

PR #1579, opened 2021-01-15
builder: parallelize most of the build

aykevl:
This commit parallelizes almost everything that can currently be parallelized. With that, it also introduces a framework for easily parallelizing other parts of the compiler.

Code for baremetal targets already compiles slightly faster with this PR because it can parallelize the compilation of supporting assembly files. However, the speedup is especially noticeable when libraries (compiler-rt, picolibc) also need to be compiled: they will be compiled in parallel next to the Go files using all available cores. On my dual core laptop (4 cores if you count hyperthreading) this cuts compilation time roughly in half when compiling something for a Cortex-M board after running `tinygo clean` (thus with a cold cache). It will likely provide a bigger speedup on a system with more cores.

The end goal is caching the build result of individual packages. That's something really big so this commit lays the groundwork for it while already providing real world benefits.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -65,10 +65,286 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 		return err
 	}
 
+	// The slice of jobs that orchestrates most of the build.
+	// This is somewhat like an in-memory Makefile with each job being a
+	// Makefile target.
+	var jobs []*compileJob
+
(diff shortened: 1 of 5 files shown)


niaow (approved):

deadprogram:
Thanks @aykevl for this awesome feature and @niaow for the review.

PR #1581, opened 2021-01-16
main: use LLVM 11 by default when linking LLVM dynamically

aykevl:
This doesn't affect the release builds but it is helpful for TinyGo developers.


deadprogram:
Thanks @aykevl this will avoid accidental confusion when building from sources.

PR #1582, opened 2021-01-16
compiler: test float to int conversions and fix upper-bound calculation

niaow:
This should have gone into #1574


niaow:
@aykevl I think this should address some of your concerns from #1574, and indeed the additional testing revealed a single broken edge case (the expression for the upper bound was missing parentheses, resulting in UB).

aykevl (approved):
Thank you for the added testing and the fix, this looks good.

PR #1583, opened 2021-01-18
Add support for Go 1.16.

QuLogic:
This is not yet available in CircleCI images, but it is in testing in Fedora Rawhide. As such, these are only the changes needed to get it to work right now, without any attempt at backwards-compatibility.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -268,8 +268,8 @@ commands:
       - run:
           name: "Install dependencies"
           command: |
-            curl https://dl.google.com/go/go1.15.5.darwin-amd64.tar.gz -o go1.15.5.darwin-amd64.tar.gz
-            sudo tar -C /usr/local -xzf go1.15.5.darwin-amd64.tar.gz
+            curl https://dl.google.com/go/go1.16.darwin-amd64.tar.gz -o go1.16.darwin-amd64.tar.gz
+            sudo tar -C /usr/local -xzf go1.16.darwin-amd64.tar.gz
             ln -s /usr/local/go/bin/go /usr/local/bin/go
             HOMEBREW_NO_AUTO_UPDATE=1 brew install qemu
       - install-xtensa-toolchain:
@@ -370,6 +370,12 @@ jobs:
     steps:
       - test-linux:
           llvm: "11"
+  test-llvm11-go116:
+    docker:
+      - image: circleci/golang:1.16-buster
+    steps:
+      - test-linux:
+          llvm: "11"
   assert-test-linux:
     docker:
       - image: circleci/golang:1.14-stretch
@@ -396,6 +402,7 @@ workflows:
       - test-llvm10-go113
       - test-llvm10-go114
       - test-llvm11-go115
+      - test-llvm11-go116
       - build-linux
       - build-macos
       - assert-test-linux
--- a/cgo/cgo_test.go
+++ b/cgo/cgo_test.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"go/ast"
+	"go/build"
 	"go/format"
 	"go/parser"
 	"go/token"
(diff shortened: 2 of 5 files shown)


deadprogram:
Is this still the case with `go1.16rc1` that was just released?

Also, what else is required for backward compatibility?

aykevl:
Thank you for starting this work!

For backwards incompatible changes, I would suggest we update the testcase to the latest Go version (1.16) and skip the test on older versions. That way all tests can pass and everything is still tested on the latest (soon to be most used) version of Go.

QuLogic:
Yes, looks like CircleCI added 1.16 now, and this change is still needed there, as you can see from the build.

> For backwards incompatible changes, I would suggest we update the testcase to the latest Go version (1.16) and skip the test on older versions. That way all tests can pass and everything is still tested on the latest (soon to be most used) version of Go.

I guess there are no current examples of skipping for these types of tests, which use external data?

aykevl:
> I guess there are no current examples of skipping for these types of tests, which use external data?

You can check [`runtime.Version()`](https://pkg.go.dev/runtime#Version) and call `t.Skip()` if it's below Go 1.16.

QuLogic:
TBH, that might have been a bit overkill, but now those tests can be skipped in any which way build constraints can be specified.

cgo/cgo_test.go
```diff
@@ -49,6 +111,8 @@ func TestCGo(t *testing.T) {
 				t.Fatal("could not parse Go source file:", err)
 			}
 
+			checkBuildConstraint(t, f)
```
aykevl: Well, honestly, I think it would be easier to understand if you just checked for that file specifically. For example (untested):

```go
minorVersion, err := strconv.Atoi(strings.Split(runtime.Version, ".")[1])
if err != nil {
    t.Fatal("could not parse Go version", runtime.Version)
}
// ...
if name == "errors" && minorVersion < 16 {
    t.Skip()
}
```

That's a very simple test and tests exactly what is needed. In fact, checking for GOOS/GOARCH should not be necessary in CGo tests because we use a fixed GOOS/GOARCH. The only things that we _don't_ control are the libclang version (which shouldn't affect the result much) and the Go version (which can certainly affect the results, for example due to formatting changes, see `normalizeResult`).

deadprogram:
Hello @QuLogic just wondering if you saw the last feedback? Now that Go 1.16 is actually out we need it! :smiley_cat:

cgo/cgo_test.go
QuLogic: I don't like parsing `runtime.Version()` as the value is not super consistent if you have a dev version, or even alpha/beta/rc. It's safer to check the build context, and I threw the other ones in because they were there and easy to do, but if you think that's too much, I can roll that back.

deadprogram: > the value is not super consistent if you have a dev version, or even alpha/beta/rc

Good point.

Regarding the other values, I think @aykevl is just trying to keep it as simple as possible for the moment, and avoid adding anything we're not using right now. Probably removing at present is a good idea, if you do not mind.

aykevl: > the value is not super consistent if you have a dev version, or even alpha/beta/rc

Agreed, however this is only for tests (which we normally only run on release versions). So I don't think it is that much of an issue.

That said, `build.Default.ReleaseTags` is interesting. Perhaps you could simply check whether `go1.16` is in the list? That should be a very simple test and avoids the issue with dev versions.

aykevl: > I think @aykevl is just trying to keep it as simple as possible for the moment, and avoid adding anything we're not using right now.

Yes, exactly. Usually the simpler something is, the less likely it is to break and the easier it is to maintain.

.circleci/config.yml
```diff
@@ -370,6 +370,12 @@ jobs:
     steps:
       - test-linux:
           llvm: "11"
+  test-llvm11-go116:
+    docker:
+      - image: circleci/golang:1.16-rc-buster
```
aykevl: Can you update this to the final version? (Probably circleci/golang:1.16-buster)

aykevl (commented):
This looks good to me, apart from one small thing (see inline comment)

There are a few other places where the Go version could be updated, it would be great to have that as well:

* in MacOS, see the `build-macos` CircleCI job
* on Windows, see the GoTool@0 task in azure-pipelines.yml
* on Linux (still on Go 1.14 somehow), see the build-linux and assert-test-linux jobs

but it's more important to get Go 1.16 supported so this can come later if necessary.

.circleci/config.yml
QuLogic: It wasn't available earlier... Maybe now.

QuLogic:
> on Linux (still on Go 1.14 somehow), see the build-linux and assert-test-linux jobs

These would also mean updating stretch to buster; is that fine?

aykevl:
> > on Linux (still on Go 1.14 somehow), see the build-linux and assert-test-linux jobs
> 
> These would also mean updating stretch to buster; is that fine?

Oh. In that case, please keep it at stretch. That's necessary to keep the TinyGo binary as portable as possible (because of dynamically linked libraries).

aykevl (approved):
LGTM, assuming all tests pass.

deadprogram:
Thanks very much for working on this @QuLogic now merging!

PR #1585, opened 2021-01-19
docker: update dev docker image to use llvm11

deadprogram:
This PR updates the dev docker image to use llvm11, fixing the broken build on Docker hub.


aykevl (approved):
Oops, I should have included this in #1581.

Issue #1518, opened 2020-12-02
Vendored packages used by the standard library cannot be resolved


sykesm:
While experimenting with current HEAD + #1430, I've noticed that packages that import vendored code fail to build.

Using the ECDSA [example](https://golang.org/pkg/crypto/ecdsa/#example_) from the standard library, the build fails with the following error:
```
# vendor/golang.org/x/crypto/cryptobyte
/usr/local/go/src/vendor/golang.org/x/crypto/cryptobyte/asn1.go:14:2: could not import golang.org/x/crypto/cryptobyte/asn1 (package not imported: golang.org/x/crypto/cryptobyte/asn1)
```
It looks like this may be because the loader implementation does not make use of the `ImportMap` information from `go list` that provides the relationship between the source import and the import path.

PR #1588, opened 2021-01-21
loader: support imports from vendor directories

aykevl:
This fixes https://github.com/tinygo-org/tinygo/issues/1518.

It feels a bit like a hack, but that's the whole situation with ` [*.test]` anyway.


deadprogram:
Confirm that this corrects the issue. Thanks @aykevl now merging.

PR #1589, opened 2021-01-21
compiler: support all kinds of deferred builtins

aykevl:
This change extends defer support to all supported builitin functions.
Not all of them make sense (such as len, cap, real, imag, etc) but this change for example adds support for `defer(delete(m, key))` which is used in the Go 1.15 encoding/json package.


deadprogram:
From testing, I can confirm that it helps get further with compiling `encoding/json`.

niaow (approved):

deadprogram:
:hourglass_flowing_sand: now merging. Thanks!

PR #1590, opened 2021-01-22
reflect: implement PtrTo

aykevl:
This is necessary for encoding/json support.


deadprogram:
From testing combined with #1589 I can confirm that it helps get even further with compiling `encoding/json`.

deadprogram:
Thanks @aykevl looks good. Now merging.

PR #1592, opened 2021-01-24
compiler: remove unnecessary main.main call workaround

aykevl:
Since https://github.com/tinygo-org/tinygo/pull/1571 (in particular, the first
commit that sets the main package path), the main package is always named
"main". This makes the callMain() workaround in the runtime unnecessary and
allows directly calling the main.main function with a //go:linkname pragma.

This is just a small refactor to clean up the code a bit.


deadprogram:
Nice :scissors: merging.

PR #1593, opened 2021-01-24
compiler: move the setting of attributes to getFunction

aykevl:
This is a small refactor to move code away from `compiler.CompilePackage`, with the goal that `compiler.CompilePackage` will eventually be removed entirely in favor of `compiler.CompilePackage`.


deadprogram:
Rebase is now required against `dev` please.

aykevl:
Rebased. I was hoping git would be smart enough for this, but apparently not.

deadprogram:
I also prefer the style of the definition that comes with this refactor. Now merging, thanks @aykevl

PR #1594, opened 2021-01-24
compiler: create runtime types lazily when needed

aykevl:
This fixes a longstanding TODO comment and similar to https://github.com/tinygo-org/tinygo/pull/1593 it removes some code out of the `compiler.CompileProgram` function that doesn't need to be there.


deadprogram:
:scissors: lazy and even fewer lines of code. Nice.

PR #1600, opened 2021-01-28
compiler: move settings to a separate Config struct

aykevl:
Moving settings to a separate config struct has two benefits:
  - It decouples the compiler a bit from other packages, most
    importantly the compileopts package. Decoupling is generally a good
    thing.
  - Perhaps more importantly, it precisely specifies which settings are
    used while compiling and affect the resulting LLVM module. This will
    be necessary for caching the LLVM module.
    While it would have been possible to cache without this refactor, it
    would have been very easy to miss a setting and thus let the
    compiler work with invalid/stale data.

---

This commit is extracted from a local branch that caches built packages and that way slightly speeds up incremental builds (where you modify only one package).


deadprogram:
Good one @aykevl now merging.

PR #1606, opened 2021-01-30
reflect: check for access in the Interface method call

aykevl:
This fixes a type system loophole. The following program would incorrectly run in TinyGo, while it would trigger a panic in Go:

    package main

    import "reflect"

    func main() {
        v := reflect.ValueOf(struct {
            x int
        }{})
        x := v.Field(0).Interface()
        println("x:", x.(int))
    }

Playground link: https://play.golang.org/p/nvvA18XFqFC

The panic in Go is the following:

    panic: reflect.Value.Interface: cannot return value obtained from unexported field or method

I've shortened the message in TinyGo to save a little bit of space.

This change also required a change to the runtime package, which was relying on this (incorrect) behavior for map key hashing.

It is currently hard to test this change because there isn't really a practical way of testing panics (`recover` is unimplemented), but at least the existing tests should check for regressions.

---

Apparently this fails because `reflect.valueInterfaceUnsafe` is optimized out. I could fix it, but I know it works with the local branch that caches package builds. Maybe I can finish this PR after that change lands.

### DIFF
--- a/src/reflect/value.go
+++ b/src/reflect/value.go
@@ -28,6 +28,13 @@ func (v Value) isIndirect() bool {
 	return v.flags&valueFlagIndirect != 0
 }
 
+// isExported returns whether the value represented by this Value could be
+// accessed without violating type system constraints. For example, it is not
+// set for unexported struct fields.
+func (v Value) isExported() bool {
+	return v.flags&valueFlagExported != 0
+}
+
 func Indirect(v Value) Value {
 	if v.Kind() != Ptr {
 		return v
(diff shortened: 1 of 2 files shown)


aykevl:
It looks like the CI failures would be fixed by rebasing on top of #1612. There are other workarounds, but #1612 would be the simplest solution.

aykevl:
Rebased, in the hope that CI tests now pass.

deadprogram:
There is now a merge conflict to be resolved here, please.

aykevl:
Rebased. Thank you for the reminder!

deadprogram:
Definitely seems like a good idea to restrict this. Now merging, thanks @aykevl

PR #1607, opened 2021-01-30
machine/circuitplay-bluefruit: correct internal I2C pin mapping

deadprogram:
This PR corrects the internal I2C pin mapping for the Adafruit Circuit Playground Bluefruit.


conejoninja:
It matches the diagram, thanks for your contribution

PR #1466, opened 2020-10-29
Allow interrupts in stm32f103xx

tom-horn:
Old PR for this same change: #1442.

PR #1608, opened 2021-01-31
bluepill: fix sleep logic

kenbell:
This fixes issue where sleep was not reliably returning if an interrupt occurred during a sleep.

### DIFF
--- a/src/runtime/runtime_stm32f103.go
+++ b/src/runtime/runtime_stm32f103.go
@@ -117,14 +117,7 @@ func nanosecondsToTicks(ns int64) timeUnit {
 
 // sleepTicks should sleep for specific number of microseconds.
 func sleepTicks(d timeUnit) {
-	for d != 0 {
-		ticks()            // update timestamp
-		ticks := uint32(d) // current scaling only supports 100 usec to 6553 msec
-		if !timerSleep(ticks) {
-			return
-		}
-		d -= timeUnit(ticks)
-	}
+	timerSleep(uint32(d))
 }
 
 // number of ticks (microseconds) since start.
@@ -143,8 +136,7 @@ func ticks() timeUnit {
 }
 
 // ticks are in microseconds
-// returns false if an interrupt occured
-func timerSleep(ticks uint32) bool {
+func timerSleep(ticks uint32) {
 	timerWakeup.Set(0)
 
 	// STM32 timer update event period is calculated as follows:
@@ -191,19 +183,10 @@ func timerSleep(ticks uint32) bool {
 	// Enable the timer.
 	stm32.TIM3.CR1.SetBits(stm32.TIM_CR1_CEN)
 
-wait:
-	arm.Asm("wfi")
-	if timerWakeup.Get() != 0 {
-		return true
+	// Wait for interrupt to fire
+	for timerWakeup.Get() == 0 {
+		arm.Asm("wfi")
 	}
-
-	if hasScheduler {
-		return false
-	} else {
-		// keep looping until the routine exits or is interrupted
-		goto wait
-	}
-
 }
 
 func handleTIM3(interrupt.Interrupt) {


aykevl:
It looks like this will remove support for channels in interrupts, introduced in https://github.com/tinygo-org/tinygo/pull/1142 (for the stm32f103 in https://github.com/tinygo-org/tinygo/pull/1466).

kenbell:
Ah - that explains why the code was so complex :(

I'll take a look.

kenbell:
@aykevl - I've taken a look.  I've been using the 'systick' example which appears to use a channel with an interrupt handler.

The example works with both 'dev branch HEAD' and with this fix.  I've tried with several scheduler settings:

| Scheduler | dev HEAD | fix |
|--|--|--|
|none|UART output: `panic: runtime error: scheduler is disabled`|UART output: `panic: runtime error: scheduler is disabled`|
|coroutines|rapid flashing LED|rapid flashing LED|
|tasks|rapid flashing LED|rapid flashing LED|

I don't fully understand the scheduler code, but i think this means:

1. Channels appear to be working with the change
2. Scheduler `none` is already non-functional
3. The code i removed was special-casing a case that doesn't work anyway

It's quite likely i've missed the mark here and not looking at the right case that may now be broken with this PR?  If so, if you can point me in the right direction that would be awesome.

deadprogram:
I can verify that the PR fixes my issues withe the UART on Bluepill. Echo example is now working perfectly.

deadprogram:
@aykevl does this address your previous concerns?

src/runtime/runtime_stm32f103.go
```diff
@@ -117,14 +117,7 @@ func nanosecondsToTicks(ns int64) timeUnit {
...
 // sleepTicks should sleep for specific number of microseconds.
 func sleepTicks(d timeUnit) {
-	for d != 0 {
-		ticks()            // update timestamp
-		ticks := uint32(d) // current scaling only supports 100 usec to 6553 msec
-		if !timerSleep(ticks) {
-			return
-		}
-		d -= timeUnit(ticks)
-	}
```
aykevl, line 127: This code is necessary. It makes sure that you can sleep for longer durations than the timer supports.

src/runtime/runtime_stm32f103.go
```diff
@@ -191,19 +183,10 @@ func timerSleep(ticks uint32) bool {
...
-
-	if hasScheduler {
-		return false
-	} else {
-		// keep looping until the routine exits or is interrupted
-		goto wait
+	// Wait for interrupt to fire
+	for timerWakeup.Get() == 0 {
+		arm.Asm("wfi")
```
aykevl, line 188: This change means that waking up a goroutine from an interrupt won't work anymore. That's still better than if the chip is non-functional entirely so might be intentional.

aykevl (commented):
I think this PR is basically a revert of this PR: https://github.com/tinygo-org/tinygo/pull/1466

However, it is still not entirely correct. Perhaps it's better to just revert that commit? I think that will fix the issue with sleep.

kenbell:
I copied the pattern from one of the other STM32 chips.  Taking a quick survey, there are two patterns:

### Algorithm 1

1. Set timer interrupt for at most 'd' (timer may not have resolution to wait for whole of 'd')
2. Wait for the timer interrupt (ignoring other exits from wfi)
3. Return to scheduler

```
runtime_stm32f7x2.go
runtime_stm32f405.go
runtime_stm32f407.go
runtime_stm32l5x2.go
runtime_stm32f103.go (my change)
```

### Algorithm 2

1. Set timer interrupt for at most 'd' (timer may not have resolution to wait for whole of 'd')
2. Wait for the timer interrupt (ignoring other exits from wfi)
3. If timer didn't have resolution for full sleep duration, go back to step 1
4. Return to scheduler

```
runtime_stm32l0.go
runtime_stm32f103.go (original, before #1466)
```

### Proposal

1. Revert change #1466 to unbreak sleep in the short-term (i.e. what @aykevl proposes)
2. Specify the expected semantics of sleepTicks precisely
3. Harmonize the STM32 sleep functionality to prevent divergence (with the exception of stm32f103 they all use a timer, only stm32f103 uses RTC - they call all use timers?)

My understanding of the spec for sleepTicks is:
```
If hasScheduler
  Sleep for at least the specified duration (allowing for overhead of setting up the sleep) unless any interrupt fires (including the timer interrupt), in which case return to the scheduler

Else (ie hasScheduler = false)
  Sleep for at least the specified duration (allowing for overhead of setting up the sleep), do not return early regardless of any interrupts that fire
```

### Bugs

If I've understood correctly, there's an additional bug in #1466 that if two interrupts had fired (the timer interrupt + one other), the logic assumed only the timer interrupt had fired and would not return to the scheduler early.

Less significant, the logic in `runtime32l0.go` also accumulates errors if it goes back to timerSleep multiple times since it doesn't recalculate from a fixed reference time - it just subtracts the 'asked for sleep duration' from the 'total sleep duration' not allowing for overheads.

### Harmonizing STM32 sleep

Given the subtlety around the sleep logic, it really seems like we should have a single implementation for STM32 chips - the basic timers are essentially identical on the chips i did a quick look at, so it should be possible to have a common implementation.  Part of this would be to move the chips that use the RTC for 'ticks' to use a timer.

It looks like it should be possible to eliminate 80% of the code from the `runtime_stm32xxxx.go` files.

deadprogram:
Just read the last comment @kenbell that is excellent research and summary, thank you for that!

Regarding the RTC being used for timer on STM32F103 I think that one is mine: https://github.com/tinygo-org/tinygo/commit/f71e1bcf03b0e9d74e34e63c057d14218754892f#diff-42a4d81db88297c7aae5ed71a34b5643f0f1ed35819c021a734714e38be5a568

I had not really thought thru the timer/clock thing and how `Sleep()` should work, and also it was one of the early implementations.

In any case, I think that your plan sounds like a good one. Reducing the code by having a single implementation, that also happens to work as desired would be a double win. :balloon: 

So I will revert that commit to start the process. Could you please close this PR, and also copy your great research info into a new GH issue that we can use to track this?

Thank you!

kenbell:
Closing as we're going the route of reverting #1466 and have raised #1666 to harmonize STM32 tick / sleep logic.

PR #1609, opened 2021-01-31
STM32 I2C: replace addressable interface with simpler type

### LINKED ISSUE #1568: feather-stm32f405: I2C not working
I finally acquired an Adafruit feather-stm32f405 board. Flashed it with the latest TInyGo `dev` branch. The MPU6050 example from here does not run correctly: https://github.com/tinygo-org/drivers/blob/release/examples/mpu6050/main.go

It only returns zero values...

```
Terminal ready
0 0 0
0 0 0
0 0 0
0 0 0
0 0 0
```

@ardnew any thoughts?

ardnew:
The `addressable` interface type is unnecessary and can be removed, which simplifies some of the code. This doesn't resolve #1568, but it's a start with diagnosing the issue.

### DIFF
--- a/src/machine/machine_stm32_i2c.go
+++ b/src/machine/machine_stm32_i2c.go
@@ -91,20 +91,6 @@ const (
 	frameNoOption     = 0xFFFF0000
 )
 
-type i2cDirection bool
-
-const (
-	directionWrite i2cDirection = false
-	directionRead  i2cDirection = true
-)
-
-func (dir i2cDirection) shift7Bit(addr uint16) uint32 {
-	if addr <<= 1; dir == directionRead {
-		addr |= 1
-	}
-	return uint32(addr) & 0xFF
-}
-
 // I2C fast mode (Fm) duty cycle
 const (
 	DutyCycle2    = 0
@@ -169,7 +155,7 @@ func (i2c I2C) Configure(config I2CConfig) {
 }
 
 func (i2c I2C) Tx(addr uint16, w, r []byte) error {
-	// a := address7Bit(addr)
+
 	if err := i2c.controllerTransmit(addr, w); nil != err {
 		return err
 	}
@@ -181,13 +167,13 @@ func (i2c I2C) Tx(addr uint16, w, r []byte) error {
 
 func (i2c I2C) controllerTransmit(addr uint16, w []byte) error {
 
+	if !i2c.waitForFlag(flagBUSY, false) {
+		return errI2CBusReadyTimeout
+	}
+
 	// ensure peripheral is enabled
 	if !i2c.Bus.CR1.HasBits(stm32.I2C_CR1_PE) {
 		i2c.Bus.CR1.SetBits(stm32.I2C_CR1_PE)
-	}
-
-	if !i2c.waitForFlag(flagBUSY, false) {
-		return errI2CBusReadyTimeout
 	}
 
 	// disable POS
@@ -252,7 +238,7 @@ func (i2c I2C) controllerRequestWrite(addr uint16, option transferOption) error 
 	}
 
 	// send peripheral address
-	i2c.Bus.DR.Set(directionWrite.shift7Bit(addr))
+	i2c.Bus.DR.Set(uint32(addr) << 1)
 
 	// wait for address ACK from peripheral
 	if !i2c.waitForFlagOrError(flagADDR, true) {
@@ -264,13 +250,13 @@ func (i2c I2C) controllerRequestWrite(addr uint16, option transferOption) error 
 
 func (i2c I2C) controllerReceive(addr uint16, r []byte) error {
 
+	if !i2c.waitForFlag(flagBUSY, false) {
+		return errI2CBusReadyTimeout
+	}
+
 	// ensure peripheral is enabled
 	if !i2c.Bus.CR1.HasBits(stm32.I2C_CR1_PE) {
 		i2c.Bus.CR1.SetBits(stm32.I2C_CR1_PE)
-	}
-
-	if !i2c.waitForFlag(flagBUSY, false) {
-		return errI2CBusReadyTimeout
 	}
 
 	// disable POS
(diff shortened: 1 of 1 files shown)


src/machine/machine_stm32_i2c.go
```diff
@@ -191,27 +169,27 @@ func (i2c I2C) Configure(config I2CConfig) {
 }
 
 func (i2c I2C) Tx(addr uint16, w, r []byte) error {
-	a := address7Bit(addr)
-	if err := i2c.controllerTransmit(a, w); nil != err {
+	// a := address7Bit(addr)
```
aykevl: This is dead code and can be removed:

```suggestion
```

src/machine/machine_stm32_i2c.go
```diff
@@ -274,13 +252,7 @@ func (i2c I2C) controllerRequestWrite(addr addressable, option transferOption) e
...
 
 	// send peripheral address
-	switch addr.bitSize() {
-	case 7: // 7-bit peripheral address
-		i2c.Bus.DR.Set(addr.toWrite())
-
-	case 10: // 10-bit peripheral address
-		// TODO
-	}
+	i2c.Bus.DR.Set(directionWrite.shift7Bit(addr))
```
aykevl: Personally I'd simplify this further by avoiding the indirection for what is really just a oneliner:

```suggestion
	i2c.Bus.DR.Set(uint32(addr) << 1)
```

The I2C protocol is fixed, this format is not going to change. Therefore I don't see the potential flexibility of moving this in a helper function.

src/machine/machine_stm32_i2c.go
```diff
@@ -464,13 +436,7 @@ func (i2c I2C) controllerRequestRead(addr addressable, option transferOption) er
...
 
 	// send peripheral address
-	switch addr.bitSize() {
-	case 7: // 7-bit peripheral address
-		i2c.Bus.DR.Set(addr.toRead())
-
-	case 10: // 10-bit peripheral address
-		// TODO
-	}
+	i2c.Bus.DR.Set(directionRead.shift7Bit(addr))
```
aykevl: Same here:
```suggestion
	i2c.Bus.DR.Set(uint32(addr)<<1 | 1)
```

src/machine/machine_stm32_i2c.go
```diff
@@ -191,27 +169,27 @@ func (i2c I2C) Configure(config I2CConfig) {
...
 
 	// ensure peripheral is enabled
 	if !i2c.Bus.CR1.HasBits(stm32.I2C_CR1_PE) {
 		i2c.Bus.CR1.SetBits(stm32.I2C_CR1_PE)
 	}
 
+	if !i2c.waitForFlag(flagBUSY, false) {
+		return errI2CBusReadyTimeout
+	}
+
```
aykevl: This  is not a pure refactor: these lines might change behavior. Was this intentional? Does this fix an issue? If not, I'd revert these changes unless this change improves something (for example, fixing #1568).

src/machine/machine_stm32_i2c.go
```diff
@@ -290,17 +262,17 @@ func (i2c I2C) controllerRequestWrite(addr addressable, option transferOption) e
...
+func (i2c I2C) controllerReceive(addr uint16, r []byte) error {
 
 	// ensure peripheral is enabled
 	if !i2c.Bus.CR1.HasBits(stm32.I2C_CR1_PE) {
 		i2c.Bus.CR1.SetBits(stm32.I2C_CR1_PE)
 	}
 
+	if !i2c.waitForFlag(flagBUSY, false) {
+		return errI2CBusReadyTimeout
+	}
```
aykevl: Same here: this code has moved (perhaps unintentionally).

aykevl (commented):
I agree this is a good simplification. I never really liked the use of interfaces here, especially around something as low-level as I2C (I don't see the benefit here while introducing complexity and thus reducing readability). However, it looks like this is not just a refactor but also changes the behavior of the I2C peripheral.

deadprogram:
Tried testing a few different I2C devices on the feather stm32f405, both with and without 10k pullup resistors, and was not able to get any to respond. Of course, that was true before this PR, sadly.

deadprogram:
@ardnew not sure if you have seem the comments on this PR? Thanks!

src/machine/machine_stm32_i2c.go
ardnew: Removed

src/machine/machine_stm32_i2c.go
ardnew: Replaced with inline shift, removed type and constants for `i2cDirection`

src/machine/machine_stm32_i2c.go
ardnew: Replaced with inline shift, removed type and constants for `i2cDirection`

src/machine/machine_stm32_i2c.go
ardnew: These changes were intentional on my part to test if it fixed the problem being reported #1568 (which I can't reproduce), but it sounds like it had no effect. (Doesn't the original implementation order look backwards? It potentially checks for a peripheral flag before the peripheral is enabled....)

Reverted back to original ordering.

src/machine/machine_stm32_i2c.go
ardnew: These changes were intentional on my part to test if it fixed the problem being reported #1568 (which I can't reproduce), but it sounds like it had no effect. (Doesn't the original implementation order look backwards? It potentially checks for a peripheral flag before the peripheral is enabled....)

Reverted back to original ordering.

ardnew:
@deadprogram sorry, overlooked these comments on my phone. committed updates to address each of them.

deadprogram:
Thanks @ardnew seems helpful to do things in the smallest increments possible so we can get things working well. This also makes it a lot easier to review things more quickly.

@aykevl looks like your feedback has been addressed. Anything else before merging this?

aykevl:
Yes this looks good to me.

deadprogram:
Thanks for the contribution @ardnew now merging.

PR #1610, opened 2021-01-31
machine/clue: correct for lack of low frequency crystal

deadprogram:
This PR changes the Adafruit Clue which it seems does not possess a low frequency crystal.


aykevl (approved):
Not sure why this is needed, but as discussed this is necessary for BLE support.

deadprogram:
OK, merging. Thanks for actually figured out the work-around here @aykevl

PR #1611, opened 2021-02-01
all: go mod tidy

aykevl:
Clean up lots of unnecessary dependencies using `go mod tidy`. These seem to be introduced in https://github.com/tinygo-org/tinygo/pull/1540.


deadprogram:
:broom: thanks @aykevl now merging.

PR #1612, opened 2021-02-01
Build packages independently

aykevl:
This commit switches from the previous behavior of compiling the whole program at once, to compiling every package in parallel and linking the LLVM bitcode files together for further whole-program optimization. This is a small performance win on incremental builds, but it has several advantages in the future:

  - There are many more things that can be done per package in parallel, avoiding the bottleneck at the end of the compiler phase. This should speed up the compiler further in the future.
  - This change is a necessary step towards a non-LTO build mode for fast incremental builds that only rebuild the changed package, when compiler speed is more important than binary size.
  - This change refactors the compiler in such a way that it will be easier to inspect the IR for one package only. Inspecting this IR will be very helpful for compiler developers.


aykevl:
I'm starting to suspect I'm hitting a LLVM bug here.

    fatal error: undefined instruction with sp=0x200007b4 pc=nil

For some reason LLVM thinks it can emit a `ldrexd` instruction, which appears to be unsupported on most Cortex-M chips. However, it looks like QEMU allows this instruction as long as the target address is 8 byte aligned, which it happened to be before this PR (by chance) and no longer is after this PR. Still looking into it.

aykevl:
I think I have a fix. Will test the patch and submit to upstream LLVM for review. Meanwhile I think we'll need a workaround in TinyGo for this bug (which should not be too difficult).

aykevl:
Patch submitted: https://reviews.llvm.org/D95891
But this is a separate issue (unrelated to this PR, only triggered by it due to random changes) so I hope to have a workaround in TinyGo until this patch is accepted upstream and trickles down to all the LLVM versions we support.

aykevl:
Fix for the atomics issue: https://github.com/tinygo-org/tinygo/pull/1613

aykevl:
This is now ready for review.

aykevl:
... although it looks like I need to fix the Windows build :unamused:

aykevl:
I fixed the Windows build and rebased against the dev branch. This is now ready for review.

aykevl:
All tests are passing now.

niaow (approved):

deadprogram:
Good work on this PR @aykevl and thanks for reviewing @niaow 

Now merging.

PR #1613, opened 2021-02-02
compiler: work around an ARM backend bug in LLVM

aykevl:
Because of a bug in the ARM backend of LLVM, the cmpxchg instruction is
lowered using ldrexd/strexd instructions which don't exist on Cortex-M
cores. This leads to an "undefined instruction" exception at runtime.
Therefore, this patch works around this by lowering directly to a call
to the __sync_val_compare_and_swap_8 function, which is what the backend
should be doing.

For details, see: https://reviews.llvm.org/D95891

To test this patch, you can run the code on a Cortex-M3 or higher
microcontroller, for example:

    tinygo flash -target=pca10040 ./testdata/atomic.go

Before this patch, this would trigger an error. With this patch, the
behavior is correct. The error (without this patch) could look like
this:

    fatal error: undefined instruction with sp=0x200007cc pc=nil

---

My intention is to remove the code in a year or so when hopefully all relevant LLVM versions are patched. That's why I included the LLVM version ("LLVM 11") so it's easier to grep for (I did something similar with other LLVM workarounds).

### DIFF
--- a/compiler/atomic.go
+++ b/compiler/atomic.go
@@ -1,6 +1,8 @@
 package compiler
 
 import (
+	"strings"
+
 	"golang.org/x/tools/go/ssa"
 	"tinygo.org/x/go-llvm"
 )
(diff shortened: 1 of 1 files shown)


deadprogram:
@aykevl one of the test suites is still failing just in case you did not notice.

aykevl:
I noticed but thank you for letting me know. I've pushed a fix.

aykevl:
All tests pass!

deadprogram:
Great job @aykevl amazing work! Now merging.

PR #1614, opened 2021-02-04
Add SPI support for Atmega based chips.

schmidtw:
This is based on @Nerzal's #1398 PR, but is a bit of a rework and
expansion to all the Atmega based chips.

### DIFF
--- a/src/machine/machine_atmega.go
+++ b/src/machine/machine_atmega.go
@@ -5,6 +5,8 @@ package machine
 import (
 	"device/avr"
 	"runtime/interrupt"
+	"runtime/volatile"
+	"unsafe"
 )
 
 // I2CConfig is used to store config info for I2C.
@@ -154,4 +156,109 @@ func (uart UART) WriteByte(c byte) error {
 	}
 	avr.UDR0.Set(c) // send char
 	return nil
+}
+
+// SPIConfig is used to store config info for SPI.
+type SPIConfig struct {
+	Frequency uint32
+	LSBFirst  bool
+	Mode      uint8
+}
+
+// SPI is for the Serial Peripheral Interface
+// Data is taken from http://ww1.microchip.com/downloads/en/DeviceDoc/ATmega48A-PA-88A-PA-168A-PA-328-P-DS-DS40002061A.pdf page 169 and following
+type SPI struct {
+	// The registers for the SPIx port set by the chip
+	spcr *volatile.Register8
+	spdr *volatile.Register8
+	spsr *volatile.Register8
+
+	// The io pins for the SPIx port set by the chip
+	sck Pin
+	sdi Pin
+	sdo Pin
+	cs  Pin
+}
+
+// Configure is intended to setup the SPI interface.
+func (s SPI) Configure(config SPIConfig) error {
+
+	// This is only here to help catch a bug with the configuration
+	// where a machine missed a value.
+	if s.spcr == (*volatile.Register8)(unsafe.Pointer(uintptr(0))) ||
+		s.spsr == (*volatile.Register8)(unsafe.Pointer(uintptr(0))) ||
+		s.spdr == (*volatile.Register8)(unsafe.Pointer(uintptr(0))) ||
+		s.sck == 0 || s.sdi == 0 || s.sdo == 0 || s.cs == 0 {
+		return errSPIInvalidMachineConfig
+	}
+
+	// Make the defaults meaningful
+	if config.Frequency == 0 {
+		config.Frequency = 4000000
+	}
+
+	// Default all port configuration bits to 0 for simplicity
+	s.spcr.Set(0)
+	s.spsr.Set(0)
+
+	// Setup pins output configuration
+	s.sck.Configure(PinConfig{Mode: PinOutput})
+	s.sdi.Configure(PinConfig{Mode: PinInput})
+	s.sdo.Configure(PinConfig{Mode: PinOutput})
+
+	// Prevent CS glitches if the pin is enabled Low (0, default)
+	s.cs.High()
+	// If the CS pin is not configured as output the SPI port operates in
+	// slave mode.
+	s.cs.Configure(PinConfig{Mode: PinOutput})
+
+	frequencyDivider := CPUFrequency() / config.Frequency
+
+	switch {
+	case frequencyDivider >= 128:
+		s.spcr.SetBits(avr.SPCR_SPR0 | avr.SPCR_SPR1)
+	case frequencyDivider >= 64:
+		s.spcr.SetBits(avr.SPCR_SPR1)
+	case frequencyDivider >= 32:
+		s.spcr.SetBits(avr.SPCR_SPR1)
+		s.spsr.SetBits(avr.SPSR_SPI2X)
+	case frequencyDivider >= 16:
+		s.spcr.SetBits(avr.SPCR_SPR0)
+	case frequencyDivider >= 8:
+		s.spcr.SetBits(avr.SPCR_SPR0)
+		s.spsr.SetBits(avr.SPSR_SPI2X)
+	case frequencyDivider >= 4:
+		// The clock is already set to all 0's.
+	default: // defaults to fastest which is /2
+		s.spsr.SetBits(avr.SPSR_SPI2X)
+	}
+
+	switch config.Mode {
+	case Mode1:
+		s.spcr.SetBits(avr.SPCR_CPHA)
+	case Mode2:
+		s.spcr.SetBits(avr.SPCR_CPOL)
+	case Mode3:
+		s.spcr.SetBits(avr.SPCR_CPHA | avr.SPCR_CPOL)
+	default: // default is mode 0
+	}
+
+	if config.LSBFirst {
+		s.spcr.SetBits(avr.SPCR_DORD)
+	}
+
+	// enable SPI, set controller, set clock rate
+	s.spcr.SetBits(avr.SPCR_SPE | avr.SPCR_MSTR)
+
+	return nil
+}
+
+// Transfer writes the byte into the register and returns the read content
+func (s SPI) Transfer(b byte) (byte, error) {
+	s.spdr.Set(uint8(b))
+
+	for !s.spsr.HasBits(avr.SPSR_SPIF) {
+	}
+
+	return byte(s.spdr.Get()), nil
 }
--- a/src/machine/machine_atmega1284p.go
+++ b/src/machine/machine_atmega1284p.go
@@ -69,3 +69,13 @@ func (p Pin) getPortMask() (*volatile.Register8, uint8) {
 		return avr.PORTD, 1 << uint8(p-portD)
 	}
 }
+
+// SPI configuration
+var SPI0 = SPI{
+	spcr: avr.SPCR,
+	spsr: avr.SPSR,
+	spdr: avr.SPDR,
+	sck:  PB7,
+	sdo:  PB5,
+	sdi:  PB6,
+	cs:   PB4}
--- a/src/machine/machine_atmega2560.go
+++ b/src/machine/machine_atmega2560.go
@@ -126,3 +126,13 @@ func (p Pin) getPortMask() (*volatile.Register8, uint8) {
 		return avr.PORTA, 255
 	}
 }
+
+// SPI configuration
+var SPI0 = SPI{
+	spcr: avr.SPCR,
+	spdr: avr.SPDR,
+	spsr: avr.SPSR,
+	sck:  PB1,
+	sdo:  PB2,
+	sdi:  PB3,
+	cs:   PB0}
--- a/src/machine/machine_atmega328pb.go
+++ b/src/machine/machine_atmega328pb.go
@@ -1,4 +1,4 @@
-// +build avr,atmega328p
+// +build avr,atmega328pb
 
 package machine
 
@@ -7,7 +7,7 @@ import (
 	"runtime/volatile"
 )
 
-const irq_USART0_RX = avr.IRQ_USART_RX
+const irq_USART0_RX = avr.IRQ_USART0_RX
 
 // getPortMask returns the PORTx register and mask for the pin.
 func (p Pin) getPortMask() (*volatile.Register8, uint8) {
@@ -91,10 +91,19 @@ func (pwm PWM) Set(value uint16) {
 
 // SPI configuration
 var SPI0 = SPI{
-	spcr: avr.SPCR,
-	spdr: avr.SPDR,
-	spsr: avr.SPSR,
+	spcr: avr.SPCR0,
+	spdr: avr.SPDR0,
+	spsr: avr.SPSR0,
 	sck:  PB5,
 	sdo:  PB3,
 	sdi:  PB4,
 	cs:   PB2}
+
+var SPI1 = SPI{
+	spcr: avr.SPCR1,
+	spdr: avr.SPDR1,
+	spsr: avr.SPSR1,
+	sck:  PC1,
+	sdo:  PE3,
+	sdi:  PC0,
+	cs:   PE2}
--- a/src/machine/spi.go
+++ b/src/machine/spi.go
@@ -1,4 +1,4 @@
-// +build !baremetal sam stm32,!stm32f7x2 fe310 k210
+// +build !baremetal sam stm32,!stm32f7x2 fe310 k210 atmega
 
 package machine
 
@@ -13,7 +13,8 @@ const (
 )
 
 var (
-	ErrTxInvalidSliceSize = errors.New("SPI write and read slices must be same size")
+	ErrTxInvalidSliceSize      = errors.New("SPI write and read slices must be same size")
+	errSPIInvalidMachineConfig = errors.New("SPI port was not configured properly by the machine")
 )
 
 // Tx handles read/write operation for SPI interface. Since SPI is a syncronous write/read


schmidtw:
This is my first bit of coding with this code base & I'm not entirely sure I have the code in the right places.  I am fairly certain that the SPI port works (I have it connected to my [Saleae](https://www.saleae.com) and am able to confirm that the program below produces the right output using an Arduino Uno board.

```go
package main

import (
	"fmt"
	"machine"
	"time"
)

func main() {
	time.Sleep(time.Millisecond * 1000)
	fmt.Printf("hello world!\n")

	time.Sleep(time.Millisecond * 1000)

	cs := machine.SPI0_CS_PIN
	cs.High()
	cs.Configure(machine.PinConfig{Mode: machine.PinOutput})

	machine.SPI0.Configure(machine.SPIConfig{
		LSBFirst:  true,
		Frequency: 5,
		Mode:      0, //machine.Mode0,
	})
	out := []byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 0xf0, 0x0f, 0xff}

	for {
		time.Sleep(time.Millisecond * 1000)

		fmt.Printf("SPI Out!\n")
		cs.Low()
		machine.SPI0.Tx(out, nil)
		cs.High()
	}
}
```

schmidtw:
I'm not sure where the right place to declare `SPI0` at.  If I do not declare it [here](https://github.com/tinygo-org/tinygo/pull/1614/files#diff-d1606be86f97c2b5674e97efa76e43e1b2a6cb95923b64b279e40050865dce86R56) when I try to compile the example program using `tinygo flash -target arduino spi.go` the SPI0 port is not defined.

Suggestions?

Nerzal:
Just out of curiosity - can you tell me what was wrong with my code? I did not manage to get it to run correctly.

And just to be sure - have you made sure that your "tinygo flash ..." commands refers to the correct tinygo version, that you have built? 
What i do most of the time to make sure it uses the correct tinygo binary, i build tinygo with the -o flag  like

```bash
 go build -o my_tinygo
```

and then directly use that binary like in the following example: 

```bash
./my_tinygo flash --target=arduino path/to/main.go
```

I think that could be your issue, because the placement of the SPI0 looks correct to me :)

src/machine/machine_atmega.go
```diff
@@ -155,3 +155,91 @@ func (uart UART) WriteByte(c byte) error {
...
+type SPI struct{}
+
+// Configure is intended to setup the SPI interface.
+func (spi SPI) Configure(config SPIConfig) error {
+
+	// Make the defaults meaningful
+	if config.Frequency == 0 {
+		config.Frequency = 4000000
+	}
+	if 0 == config.SCK {
```
Nerzal: You might want to resolve the Yoda Conditions, by switching the order of the check. 

Example: config.SCK == 0  instead of 0 == config.SCK

src/machine/machine_atmega.go
```diff
@@ -155,3 +155,91 @@ func (uart UART) WriteByte(c byte) error {
...
+		avr.SPCR.SetBits(avr.SPCR_SPR1)
+	case frequencyDivider >= 32:
+		avr.SPCR.SetBits(avr.SPCR_SPR1)
+		avr.SPSR.SetBits(avr.SPSR_SPI2X)
+	case frequencyDivider >= 16:
+		avr.SPCR.SetBits(avr.SPCR_SPR0)
+	case frequencyDivider >= 8:
+		avr.SPCR.SetBits(avr.SPCR_SPR0)
+		avr.SPSR.SetBits(avr.SPSR_SPI2X)
+	case frequencyDivider >= 4:
```
Nerzal, line 230: golang has no default fall through switch cases. 
So i fear, that you either need to add the fallthrough statement in that case, or just omit this case, to let it go through the default one.

Nerzal (changes requested):
Mostly looks good to me, only found 2 little things :+1:

schmidtw:
> Just out of curiosity - can you tell me what was wrong with my code? I did not manage to get it to run correctly.

Good question.  I had to go back to try to figure this out.  The one thing I know is an issue is the way you set the clock.  The bits for the speed were using the bitmask of the register, but the SPI_2XCLOCK_MASK is 0x01 of the SPSR register and SPI_CLOCK_MASK is 0x03 of the SPCR register.  This means that for SPI_CLOCK_FOSC128 (3), your code was setting SPR0, SPR1 & 2XCLOCK when only SPR0 & SPR1 should have been set.

```
avr.SPSR.SetBits(clockDivider & SPI_2XCLOCK_MASK)
...
avr.SPCR.SetBits(avr.SPCR_SPE | avr.SPCR_MSTR | clockDivider&SPI_CLOCK_MASK)
```

Beyond that I'm not sure if there was much that was actually wrong.  I couldn't get the code to work with my initial test program, saw the clock issue and then pursued the approach to explicitly zero out all the registers used to make sure they were initialized to a known value and work from there.  After longer than I would like to admit I discovered that in my test program I was using the LED pin on the Arduino to signal when I should be getting a small burst of SPI... and the LED is the same pin as the SCK.  Once I removed the LED code everything worked.  By this time the code was very close to what I submitted.  I tried to normalize the code to match the patterns of what the code around it and your patch were after.


> And just to be sure - have you made sure that your "tinygo flash ..." commands refers to the correct tinygo version, that you have built?

Yes, but good question.

> What i do most of the time to make sure it uses the correct tinygo binary, i build tinygo with the -o flag like
> 
> ```shell
>  go build -o my_tinygo
> ```
> 
> and then directly use that binary like in the following example:
> 
> ```shell
> ./my_tinygo flash --target=arduino path/to/main.go
> ```
> 
> I think that could be your issue, because the placement of the SPI0 looks correct to me :)

Yeah, the command line you suggest works for building the test program when I have SPI0 where it is in the PR.  However, that definition seems to conflict with what is [here](https://github.com/tinygo-org/tinygo/blob/cca0eab3da7065d568ca073df61f03beecbd2086/src/machine/machine_generic.go#L8), which breaks the CI build.

src/machine/machine_atmega.go
schmidtw: Good call out - I'm going to add a comment in the code to make this a bit more clear.  The `>= 4` case results in the bits being cleared to 0 which means no setting is needed for this case.  The default value is anything faster than 4 & is configured differently.

Nerzal:
@schmidtw thanks for pointing out, what was broken in my PR. 
I'll just close my PR in favor to this one :)

src/machine/board_arduino.go
```diff
@@ -43,3 +43,11 @@ const (
 	UART_TX_PIN Pin = PD1
 	UART_RX_PIN Pin = PD0
 )
+
+// SPI pins
+const (
+	SPI0_SCK_PIN Pin = PB5
+	SPI0_SDO_PIN Pin = PB3
+	SPI0_SDI_PIN Pin = PB4
+	SPI0_CS_PIN  Pin = PB2
```
aykevl: I see that some other boards also have a defined CS pin, but to be honest I don't think it should be used like this. The whole point of the chip select pin is to select which (of multiple) external devices are currently in use, and therefore you should use one per external device, not one per SPI peripheral.

```suggestion
```

src/machine/machine_atmega.go
```diff
@@ -155,3 +155,94 @@ func (uart UART) WriteByte(c byte) error {
...
+	}
+	if config.SCK == 0 {
+		config.SCK = SPI0_SCK_PIN
+	}
+	if config.SDO == 0 {
+		config.SDO = SPI0_SDO_PIN
+	}
+	if config.SDI == 0 {
+		config.SDI = SPI0_SDI_PIN
+	}
```
aykevl: This doesn't make sense to me. AVR chips have assigned fixed pins to certain SPI peripherals, so there is no point in configuring different pins.

src/machine/machine_atmega.go
```diff
@@ -155,3 +155,94 @@ func (uart UART) WriteByte(c byte) error {
...
+
+	// Default all port configuration bits to 0 for simplicity
+	avr.SPCR.Set(0)
+	avr.SPSR.Set(0)
+
+	// Setup pins output configuration
+	config.SCK.Configure(PinConfig{Mode: PinOutput})
+	config.SDI.Configure(PinConfig{Mode: PinInput})
+	config.SDO.Configure(PinConfig{Mode: PinOutput})
```
aykevl: Same here.

src/machine/machine_avr.go
```diff
@@ -154,3 +154,6 @@ var (
 	// UART0 is the hardware serial port on the AVR.
 	UART0 = UART{Buffer: NewRingBuffer()}
 )
+
+// SPI on AVR.
+var SPI0 = SPI{}
```
aykevl: This should be in the same file as where the SPI is defined, as SPI on AVR depends on fixed pin numbers.

src/examples/spi/spi.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"machine"
+	"time"
+)
+
+func main() {
+	cs := machine.SPI0_CS_PIN
+	cs.High()
+	cs.Configure(machine.PinConfig{Mode: machine.PinOutput})
```
aykevl: @deadprogram do you have an opinion on whether to include an example here?

In any case, if you include an example you should also add a line to the smoke tests in the Makefile to make sure it continues to compile.

aykevl (commented):
I think the SPI implementation should live in machine_atmega328p.go, and definitions like `SPI0_SCK_PIN` too. The reason is that the ports used for SPI cannot be configured in software, unlike many other (more powerful) microcontrollers. That should fix errors like the following:

```
/go/bin/tinygo build -size short -o test.hex -target=atmega1284p         examples/serial
# machine
/home/circleci/project/src/machine/machine_atmega.go:181:16: undeclared name: SPI0_SCK_PIN
/home/circleci/project/src/machine/machine_atmega.go:184:16: undeclared name: SPI0_SDO_PIN
/home/circleci/project/src/machine/machine_atmega.go:187:16: undeclared name: SPI0_SDI_PIN
```

aykevl:
> And just to be sure - have you made sure that your "tinygo flash ..." commands refers to the correct tinygo version, that you have built?
> What i do most of the time to make sure it uses the correct tinygo binary, i build tinygo with the -o flag like
> 
> ```shell
>  go build -o my_tinygo
> ```

You don't need to build TinyGo each time you make a change to the machine package, only when you make a change to the compiler.  The standard library is not included in the `tinygo` binary.

src/machine/board_arduino.go
schmidtw: I'm fine removing the CS pin.

I initially went with the various documents from Atmel (all have a CS or SS pin defined even though they're just GPIO).  In the past AVR32 or PIC chips I've seen have had a special group of CS pins so that the SPI hardware could do the enable/disable automatically based on which chip you wanted to send/receive from ... all via interrupts.  From what I can tell, that pattern doesn't exist in any of the machines tinygo supports.

schmidtw: I moved the implementation up to atmega since from what I could tell all the atmega chips have the same SPI hardware register implementation.

I agree with the pins being defined per chip or board or by the user.  It's just really nice to be able to wire & when there's only one possible choice it "just works".  This seems consistent with I2C and USART as well.

aykevl: > In the past AVR32 or PIC chips I've seen have had a special group of CS pins so that the SPI hardware could do the enable/disable automatically based on which chip you wanted to send/receive from ... all via interrupts. From what I can tell, that pattern doesn't exist in any of the machines tinygo supports.

Yeah, I'm aware this exists. We might want to add support for that eventually, but would need to find a way to express it in a portable way. And in any case, it only makes sense once DMA is supported (which it currently isn't, for the most part).

> I moved the implementation up to atmega since from what I could tell all the atmega chips have the same SPI hardware register implementation.

No. The atmega1284p for example has 3 SPI peripherals, so it can't use the exact same registers: there need to be 3 sets of them with different names. Maybe it has the same names for SPI0 but it can't have the same names for SPI1 and SPI2.

src/machine/machine_atmega.go
schmidtw: I based this on the prior PR as well as trying to be consistent with the other SPI implementations for consistency.  I'm fine deleting code that isn't used.

src/machine/machine_atmega.go
schmidtw: Something needs to configure the pins for output mode.  If there is only one & the user asks to use the SPI port, it needs to be setup right or the SPI hardware doesn't work correctly.  I think this should stay with the configuration code.

src/examples/spi/spi.go
schmidtw: Happy to add to the smoke tests.  I apologize for the broken build & missing this.  I tried running the tests, but since I didn't rebuild llvm from scratch something in the make file doesn't seem happy.  I'm working on running them one by one.

src/machine/board_arduino.go
schmidtw: I must be missing something, but I don't see the other 2 SPI peripherals in the datasheet.  I see the USART as SPI (many of the chips have that), but I didn't try to address those other 2 modes.

src/machine/machine_atmega.go
aykevl: Yes, but it should use the fixed pins for this chip:

```suggestion
	SPI0_SCK_PIN.Configure(PinConfig{Mode: PinOutput})
	SPI0_SDI_PIN.Configure(PinConfig{Mode: PinInput})
	SPI0_SDO_PIN.Configure(PinConfig{Mode: PinOutput})
```

src/examples/spi/spi.go
aykevl: You don't need to build with a custom LLVM build. At least if you do `make smoketest TINYGO=tinygo` you can pick the binary that you'd like to use.

src/machine/board_arduino.go
schmidtw: I found a different supported chip that has the pattern you mentioned. `atmega1284p`

My larger question is would you rather have a bunch of duplicated code for each processor or have a more generic solution that works for all the Atmega chips?

schmidtw: I found a different chip that matches your example: `atmega384pb`

My question is would you rather have a more generic implementation that all the Atmega chips can use, or do you want the code duplicated for each chip?

aykevl: I'm referring to this product page: https://www.microchip.com/wwwproducts/en/ATMEGA1284P

But maybe you're right, with a quick scan I can't find other SPI buses. Which kind of makes sense, as you can't really do much with multiple SPI peripherals if you don't also use DMA (to work with them in parallel).

aykevl: > My question is would you rather have a more generic implementation that all the Atmega chips can use, or do you want the code duplicated for each chip?

Ideally a generic solution if it is possible, but maybe it is not possible. For example, when I was working on a refactor of PWM support I couldn't really come up with a working generic solution.

src/machine/machine_atmega.go
```diff
@@ -155,3 +157,109 @@ func (uart UART) WriteByte(c byte) error {
...
+	// The io pins for the SPIx port set by the chip
+	sck Pin
+	sdi Pin
+	sdo Pin
+}
+
+// Configure is intended to setup the SPI interface.
+func (s SPI) Configure(config SPIConfig) error {
+
+	// This is only here to help catch a bug with the configuration
```
schmidtw, line 186: I ran into missing configuration the first time I added a chip.  It was pretty annoying, so I decided I'd try to detect and return an error.  If you don't like this I can delete it with no hurt feelings.

deadprogram: Having `Configure()` return an error is a standardization we are doing for all I2C interfaces, so seems like a good thing here as well.

src/machine/machine_atmega.go
```diff
@@ -155,3 +157,109 @@ func (uart UART) WriteByte(c byte) error {
...
+	s.spsr.Set(0)
+
+	// Setup pins output configuration
+	s.sck.Configure(PinConfig{Mode: PinOutput})
+	s.sdi.Configure(PinConfig{Mode: PinInput})
+	s.sdo.Configure(PinConfig{Mode: PinOutput})
+
+	// Ensure the CS pins are all high before enabling them to prevent any
+	// momentary low glitches that can confuse the other chips
+	for _, cs := range config.CSPins {
```
deadprogram: I do not really see what benefit having the `CSPins` slice provides here. Especially since it is not used anywhere else.

src/machine/machine_atmega.go
```diff
@@ -155,3 +157,109 @@ func (uart UART) WriteByte(c byte) error {
 	avr.UDR0.Set(c) // send char
 	return nil
 }
+
+// SPIConfig is used to store config info for SPI.
+type SPIConfig struct {
+	Frequency uint32
+	LSBFirst  bool
+	Mode      uint8
+	CSPins    []Pin
```
deadprogram: This should probably be removed. See my comment below.

src/examples/spi/spi.go
deadprogram: Seems like there will be no need for a specific example, by standardizing a bit more with the other SPI implementations. Specifically, removing  `[]CSPins`.

src/machine/machine_atmega.go
schmidtw: It's not included elsewhere but some place needs to be the first :).  (There isn't a governing interface for SPI that I could see).

I'll argue that the CS pins are missing from the other impls and not the other way around.  Here are a few examples of code that all have a bug on Atmega chips that may or may not have issues on different chips:

- https://github.com/tinygo-org/drivers/blob/e376785596dc8269f3e8aa42a9bf75fb1457febc/wifinina/wifinina.go#L271
- https://github.com/tinygo-org/drivers/blob/e376785596dc8269f3e8aa42a9bf75fb1457febc/flash/transport_spi.go#L44
- https://github.com/tinygo-org/drivers/blob/e376785596dc8269f3e8aa42a9bf75fb1457febc/bmi160/bmi160.go#L31

The output pin goes low for a few ns & then high.  For some chips this can signal the start of a transaction & inadvertently cause the first few attempts to use a chip not work.

I definitely agree with standardizing the SPI port usage.  I'm hopeful that driver or application code can delegate pin configuration to the SPI code for the chip (like it does for the other SPI pins).

But if this is the only thing keeping this code from being merged I'll remove it.

schmidtw: @deadprogram If the only thing blocking this merge is the CS handling - I'll remove it to get the rest of the code in.  I'm fine dealing with it in a later PR as well.  I'm only pinging you here because there's so many comments & I don't want you to have to waste your time looking through them if this is the only blocker.

deadprogram: I think we should probably not try to make changes to the interface for all SPI as part of this PR.

That would better served as part of a broader conversation of possible changes to the SPI interface.

deadprogram:
Just tried to test this branch using my Arduino Uno and a small strip of APA102 LED's with the example at https://github.com/tinygo-org/drivers/blob/release/examples/apa102/main.go

I changed the number of LEDs to 6, as that is all I have in my tiny strip. At first it did not work at all. When I removed the params from the `SPIConfig{}` to use the defaults, if I do a hard power reset it will light up the strip once sometimes. The times when it lights up, the animated effect does not occur.

schmidtw:
I'll look into this.

schmidtw:
I looked at the output of the SPI pins with my logic analyser & based on what I see the code wanting to send out the SPI output matches.  I'm not an APA102 export nor do I have any of those.

This is with my Arduino Uno.  I modified the program to only represent 2 LEDs:
```
	leds := make([]color.RGBA, 2)
```

Here is the output waveform.  The clock is 500kHz.  I added a CS so I could easily trigger the analyser.

Screenshot from 2021-02-09 11-49-44

As I wrote this up I decided to re-run and capture without the CS pin being configured.  The SPI output is not there.  So what I have found is that the CS pin configuration is critical for the SPI peripheral hardware to actually work.  The Atmel datasheets indicate that if the specified (default) CS pin is not in the right state, the SPI peripheral will start in slave mode, not master mode.

I added this to the example right before the `apa102.New()` is called & the waveform works as expected:
```
	cs := machine.PB2
	cs.High()
	cs.Configure(machine.PinConfig{Mode: machine.PinOutput})
```

Here's what it looks like with the CS pin configured, but not used (the SPI decoder didn't run since I have it configured to require the CS signal but the waveform is the same):
Screenshot from 2021-02-09 12-00-14

deadprogram:
That is pretty interesting. Confirmed on my end by adding the cs pin as you suggested,

Hmm, AVR SPI is pretty specific about its requirements it seems.

schmidtw:
Since the application actually controls the CS operation for AVRs the important thing is that during configuration/enabling of the peripheral the CS is in the right state.  After that, it does not matter what is used for CS if one is used at all.  In theory you could use the pin for something else, but that will be quite complex to manage & shouldn't live in this code.

I tried adding the least invasive change possible in the latest change I pushed.

deadprogram:
Something else: the APA102 freezes after about 10-15 seconds.

Keep in mind that this APA102 is one of my main devices I use for SPI testing on every other platform, using that same code in the drivers repo. So hopefully the same will be able to be true for AVR.

deadprogram:
That last commit you pushed seems to have resolved that freeze. That CS pin really means it! :smile_cat:

deadprogram:
Was also able to run the MCP3008 example on Arduino Uno: https://github.com/tinygo-org/tinygo/blob/release/src/examples/mcp3008/mcp3008.go :tada:

deadprogram:
Also confirmed MCP3008 example on Arduino Mega 2560.

@schmidtw if you do not mind, could you please squash the commits in this PR so I can merge it now? 

Thank you so much for working on this, SPI on AVR has been a long needed feature!

deadprogram:
Let me also not forget @Nerzal who did so much work on this to get it started! A true group effort.

Nerzal:
Should we add the info, that the CS pin needs to be configured prior to configure the SPI interface to the tinygo-site ? 
Or is this "common sense" ?

deadprogram:
Now merging, thanks again everyone!

aykevl:
Sorry to be so late in the review. I think the code can be further simplified so that it takes a lot less space (at the cost of being less portable). Maybe I can do that in a new PR? The current solution with a global with pointers will be rather space inefficient, something that is a problem on microcontrollers.

deadprogram:
That sounds very good to me!

PR #1616, opened 2021-02-04
machine: make I2C.Configure signature consistent

aykevl:
It's better to always return an error value (even if it is nil) for consistency. I noticed this while working on some documentation.


deadprogram:
Good to make this consistent. Now merging.

PR #1617, opened 2021-02-06
machine: remove TWI_FREQ_* constants

aykevl:
These constants:

  * Have the wrong name (TWI instead of I2C)
  * Are not formatted according to Go guidelines
  * Are not necessary: if you want `TWI_FREQ_400KHZ` you could write that directly with 400e3.

Therefore I propose that we removed them.

There is also a small change to nrf chips to also support 250kHz I2C and to machine_stm32_i2c.go to accept other values than just 100kHz/400kHz (but 100kHz/400kHz will be picked anyway). These changes allow the use of other frequencies than just 100kHz and 400kHz in I2CConfig while still doing something reasonable.

---

I have tested this on the pca10040 board. Also see https://github.com/tinygo-org/drivers/pull/229.

### DIFF
--- a/src/machine/i2c.go
+++ b/src/machine/i2c.go
@@ -6,12 +6,6 @@ import (
 	"errors"
 )
 
-// TWI_FREQ is the I2C bus speed. Normally either 100 kHz, or 400 kHz for high-speed bus.
-const (
-	TWI_FREQ_100KHZ = 100000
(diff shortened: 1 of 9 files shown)


deadprogram:
These were already marked as deprecated, and are no longer being used, so this PR is a bit out of date. Hence, closing. Please open a new one if you want to actually remove those 2 lines.

PR #1618, opened 2021-02-06
all: update go-llvm to use LLVM 11 on macOS

aykevl:
@kenbell can you confirm that this works on macOS with LLVM 11?


kenbell:
Looks good
```
% go install
go: downloading tinygo.org/x/go-llvm v0.0.0-20210206225315-7fe719483a0f
% tinygo version
tinygo version 0.17.0-dev darwin/amd64 (using go version go1.15.8 and LLVM version 11.0.0)
```

deadprogram:
Thanks @aykevl for the fix and @kenbell for verifying it. Merging.

PR #1621, opened 2021-02-07
cgo: add support for variadic functions

aykevl:
This doesn't yet add support for actually making use of variadic functions, but at least allows (unintended) variadic functions like the following to work:

    void foo();

### DIFF
--- a/cgo/cgo.go
+++ b/cgo/cgo.go
@@ -54,9 +54,10 @@ type constantInfo struct {
 // functionInfo stores some information about a CGo function found by libclang
 // and declared in the AST.
 type functionInfo struct {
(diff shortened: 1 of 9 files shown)


aykevl:
WebAssembly doesn't work yet. I think I know how to fix it.

aykevl:
This is a bit more involved, but WebAssembly should work now.

deadprogram:
Is this PR ready for review now?

aykevl:
Yes this is ready for review.

deadprogram:
Seems like a very good work-around for handling what is probably not intended by most developers.

Merging, thanks @aykevl

PR #1625, opened 2021-02-10
ci: remove bad symlink workaround on Windows

aykevl:
This workaround is no longer needed as the symlink in the LLVM source tree has been removed.

I have tested this in https://github.com/tinygo-org/tinygo/pull/1623 and verified that the `bad_symlink` symlink has been removed from the LLVM tree in LLVM 11. Therefore, there is no reason to (attempt to) remove this symlink anymore.

Upstream bug that I reported: https://github.com/microsoft/azure-pipelines-tasks/issues/13418

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -24,11 +24,7 @@ jobs:
       displayName: Download LLVM source
       inputs:
         targetType: inline
-        script: |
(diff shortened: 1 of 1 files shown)


deadprogram:
:scissors: thanks @aykevl merging.

PR #1626, opened 2021-02-10
esp8266: add compiler-rt library 

aykevl:
With this change, it is possible to compile ./testdata/float.go for the ESP8266 and run it successfully. Previously it would result in many linker error like this:

    /tmp/tinygo494494333/main.o:(.literal.runtime.printfloat64+0x0): undefined reference to `__unorddf2'
    /tmp/tinygo494494333/main.o:(.literal.runtime.printfloat64+0x4): undefined reference to `__gtdf2'
    /tmp/tinygo494494333/main.o:(.literal.runtime.printfloat64+0xc): undefined reference to `__nedf2'
    /tmp/tinygo494494333/main.o:(.literal.runtime.printfloat64+0x10): undefined reference to `__ltdf2'
    /tmp/tinygo494494333/main.o:(.literal.runtime.printfloat64+0x1c): undefined reference to `__gedf2'

I have verified that the output on the serial console matches ./testdata/float.txt when run on the ESP8266.

In the future the second commit will also be necessary for adding compiler-rt to AVR targets.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -171,7 +171,7 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 	// Add compiler-rt dependency if needed. Usually this is a simple load from
 	// a cache.
 	if config.Target.RTLib == "compiler-rt" {
-		path, job, err := CompilerRT.load(config.Triple(), dir)
+		path, job, err := CompilerRT.load(config.Triple(), config.CPU(), dir)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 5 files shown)


aykevl:
Added another commit that enables support for CGo on the ESP32/ESP8266 (see commit message).

deadprogram:
This is going to be incredibly helpful at getting more advanced features running on these chips based on existing libraries.

Thanks @aykevl now merging.

PR #1627, opened 2021-02-11
avr: use Clang for compiling C and assembly files

aykevl:
This is one step towards an avr-gcc free TinyGo toolchain, something that will make it easier to use AVR targets in TinyGo.

I have tested some code samples with an Arduino Nano and a Digispark.


deadprogram:
Does this completely remove the `avr-gcc` dependency?

aykevl:
No. It removes one instance of it. Other cases are harder to remove, I'm looking into it. But it seemed like a good idea to do it step by step (not all at once).

deadprogram:
Makes sense to me! Now merging.

aykevl:
I've looked more deeply in removing all uses of avr-gcc and it looks like we need two upstream patches first:

* https://reviews.llvm.org/D86546
* https://reviews.llvm.org/D86547

They're still in review (I'm not sure I can LGTM them). Hopefully they will be included in LLVM 13.

aykevl:
I did look at this before and thought I had covered everything, but sadly I hadn't. The bugs I'm hitting are in some compiler-rt builtins that don't have tests (which is surprising for LLVM and rather sloppy) - therefore I missed them when I ran all builtin tests on AVR. I hope I've covered everything now.

(While doing that I also discovered a rather critical AVR backend bug in LLVM 12, luckily I found it so early in the release cycle just after the LLVM 12 branch).

PR #1628, opened 2021-02-11
builder: wait for running jobs to finish

aykevl:
This avoids external commands from finishing after the TinyGo command exits. For example, when testing out compiler-rt on AVR, I got the following error:

    $ go install; and tinygo run -target=atmega1284p ./testdata/calls.go
    [... Clang error removed ...]
    error: failed to build /home/ayke/src/github.com/tinygo-org/tinygo/lib/compiler-rt/lib/builtins/extendsfdf2.c: exit status 1
    error: unable to open output file '/tmp/tinygo361380649/build-lib-compiler-rt/ffsdi2.c.o': 'No such file or directory'
    1 error generated.

That last error ("unable to open output file") is a spurious error because the temporary directory has already been removed. This commit waits until all running jobs have finished before returning, so that these errors won't happen.

### DIFF
--- a/builder/jobs.go
+++ b/builder/jobs.go
@@ -110,6 +110,12 @@ func runJobs(jobs []*compileJob) error {
 			fmt.Println("## finished:", job.description, "(time "+job.duration.String()+")")
 		}
 		if job.err != nil {
+			// Wait for running jobs to finish.
+			for numRunningJobs != 0 {
+				<-doneChan
+				numRunningJobs--
+			}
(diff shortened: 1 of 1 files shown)


deadprogram:
I wonder if this might fix another problem flashing that I sometimes have? In any case is a very good idea overall, so now merging.

aykevl:
> I wonder if this might fix another problem flashing that I sometimes have?

I don't know until you tell me what problem you have ;)

Issue #1504, opened 2020-11-20
WebAssembly binary built targeting the WASI interface cannot read environment variables


flavio:
I looks like `os.Getenv` is not working for WASM modules produced for the WASI target.

# How to reproduce the issue

Given the following program called `env-test.go`:

```go
package main

import (
	"fmt"
	"os"
)

func main() {
	fmt.Printf("HOME -> |%s|\n", os.Getenv("HOME"))
}
```

Build the program using latest (0.16.0) release of TinyGo:

```bash
$ tinygo build -o env-test.wasm -wasm-abi=generic -target=wasi env-test.go
```

And then run it with `wasmtime`:

```bash
$ wasmtime run --env HOME=flavio env-test.wasm
HOME -> ||
```

As you can see the `os.Getenv` returns an empty string.

# Environment details

* tinygo version 0.16.0 linux/amd64 (using go version go1.15.2 and LLVM version 10.0.1)
* wasmtime 0.21.0

flavio:
I would like to help fixing this issue, is there anybody that can give me some guidance? I tried to understand how the WASI target is implemented internally, but I didn't succeed (yet) :heart:

aykevl:
Yes, certainly I would like to help. Do you have specific things you would like to know? (See https://github.com/tinygo-org/tinygo/issues/1505#issuecomment-736872118).

mathetake:
Here's the PR to solve this: https://github.com/tinygo-org/tinygo/pull/1632

deadprogram:
This has been released with v0.17.0 so now closing. Thanks!

PR #1632, opened 2021-02-14
WASI & darwin: support env variables based on libc

mathetake:
Signed-off-by: Takeshi Yoneda <takeshi@tetrate.io>

resolves https://github.com/tinygo-org/tinygo/issues/1504


src/os/envi_wasi.go
```diff
@@ -0,0 +1,78 @@
+// +build wasi
+
+package os
+
+import (
+	"reflect"
+	"unsafe"
+)
+
+// TODO: this is not thread safe. When we support threading proposal, we need to reconsider the implementation.
```
mathetake: if we invoke `ensureEnv` in the init function, then we do not need to care about this even if we support threading proposal. But I'm not sure we should always call `environ_get` and `environ_sizes_get`.

mathetake:
I just realized that we can depend on libc's Getenv system call, which is simpler and generic.

deadprogram:
Any feedback on this WASM people? @bradleypeabody @aykevl anyone else?

src/syscall/syscall_libc_other.go
```diff
@@ -1,4 +1,4 @@
-// +build nintendoswitch
+// +build nintendoswitch wasi
```
aykevl: I do not think this is correct: this file is for targets that do not have a native libc (and therefore have no known constants for `SIGCHLD` etc.). It is likely that the constants in this file (for `SIGCHLD`, `O_APPEND`, etc) do not match with the values from wasi. Instead, you should add a new file with the correct constants as used in WASI.

src/os/env.go
```diff
@@ -1,3 +1,5 @@
+// +build !wasi
```
aykevl: Instead of adding an exception for WASI, I would suggest implementing `Getenv` for all platforms (like you did in env_wasi.go) and making sure that `syscall.Getenv` is implemented in all syscall implementations - perhaps implementing a dummy method that always returns `"", false` for unsupported platforms. In other words, please rename env_wasi.go over env.go and implement `syscall.Getenv` as needed.

main_test.go
```diff
@@ -108,15 +108,22 @@ func TestCompiler(t *testing.T) {
...
+var additionalArgs = map[string][]string{
+	"testdata/wasi_env.go": {"--env", "ENV1=VALUE1", "--env", "ENV2=VALUE2"},
+}
+
 func runPlatTests(target string, matches []string, t *testing.T) {
 	t.Parallel()
 
 	for _, path := range matches {
-		path := path // redefine to avoid race condition
+		if strings.Contains(path, "wasi") && target != "wasi" {
```
aykevl: I think you can simplify this by calling `runTest("testdata/wasi_env.go", "wasi", t, {"--env", "ENV1=VALUE1", "--env", "ENV2=VALUE2"})` directly in the loop above:

```go
		t.Run("WASI", func(t *testing.T) {
			runPlatTests("wasi", matches, t)
			runTest("testdata/wasi_env.go", "wasi", t, {"--env", "ENV1=VALUE1", "--env", "ENV2=VALUE2"})
		})
```

src/syscall/syscall_libc_other.go
mathetake: yeah actually I already have syscall_libc_wasi.go for [file system support branch](https://github.com/mathetake/tinygo/compare/wasi-env-vars...mathetake:libc-wasi?expand=1) which I will submit after this PR. Thanks for clarification!

testdata/wasi/env.go
```diff
@@ -0,0 +1,15 @@
...
+	"os"
+)
+
+func main() {
+	fmt.Println(os.Getenv("ENV1"))
+	v, ok := os.LookupEnv("ENV2")
+	if !ok {
+		panic("ENV2 not found")
+	}
+	fmt.Println(v)
```
aykevl: Can you use `println` instead of `fmt.Println`? The `fmt` package is very big and (currently) will slow down tests if used.

src/syscall/syscall_nonlibc.go
```diff
@@ -0,0 +1,7 @@
+// +build !darwin,!nintendoswitch,!wasi
```
aykevl: When is this needed? Isn't this just the baremetal case? In that case, I think this file should be removed and `Getenv` should be moved back to syscall_baremetal.go.

aykevl (commented):
Overall looking quite good.

Would it be possible to also test `Getenv` on macOS? In theory it should work there as well.

mathetake:
Thanks, I'm on it

src/syscall/syscall_libc.go
```diff
@@ -48,18 +54,36 @@ func Getpid() (pid int) {
...
+	ptr := uintptr(unsafe.Pointer(raw))
+	for size := uintptr(0); ; size++ {
+		v := *(*byte)(unsafe.Pointer(ptr))
+		if v == 0 {
+			src := *(*[]byte)(unsafe.Pointer(&sliceHeader{buf: raw, len: size, cap: size}))
+
+			// copy the original char*
+			dst := make([]byte, len(src))
+			copy(dst, src)
+			return *(*string)(unsafe.Pointer(&sliceHeader{buf: &dst[0], len: size, cap: size})), true
```
aykevl: I think you can simplify this. The returned value should be constant, so you can use it directly. Something like this:

```go
	return *(*string)(unsafe.Pointer(&stringHeader{
		ptr: unsafe.Pointer(raw),
		len: size,
	}))
```

aykevl: Also, as you're using a libc anyway it might be easier/simpler to just call `strlen` to get the string length. The current code is hard to read with all the `unsafe` and pointer casts.

mathetake: I am not sure how safe it is actually. Even though we haven't supported `Setenv`, but once we support that, I guess the byte array this string based on would be changed after calling `Setenv` on the same key. What do you think?

aykevl: I see, you are right. From the [Open Group](https://pubs.opengroup.org/onlinepubs/009696799/functions/getenv.html):

> The return value from getenv() may point to static data which may be overwritten by subsequent calls to getenv(), setenv(), or unsetenv().

aykevl: But in that case, you can convert it to a `[]byte` (as you're already doing) and then simply convert it to a `string`. This conversion makes a copy because if the `[]byte` is modified, the string must not be modified. Therefore, you can simplify this to:

```go
src := *(*[]byte)(unsafe.Pointer(&sliceHeader{buf: raw, len: size, cap: size}))
return string(str), true
```

mathetake: OK, that looks much simpler. Thanks!

mathetake:
I think all the comments are addressed including the test for macos. @aykevl please take another look! Thanks :slightly_smiling_face:

bradleypeabody:
LGTM, I don't have anything to add. Utilizing libc where it's already available seems like a sensible solution.

deadprogram:
Looks good @mathetake could you please squash the commits so we can merge?

Thank you for working on this!

mathetake:
squashed. Thanks!

deadprogram:
Thanks for working on this @mathetake 

@aykevl are we good to merge now?

main_test.go
```diff
@@ -58,6 +58,14 @@ func TestCompiler(t *testing.T) {
 	if runtime.GOOS != "windows" {
 		t.Run("Host", func(t *testing.T) {
 			runPlatTests("", matches, t)
+			if runtime.GOOS == "darwin" {
+				for _, env := range [][2]string{{"ENV1", "VALUE1"}, {"ENV2", "VALUE2"}} {
+					if err := os.Setenv(env[0], env[1]); err != nil {
+						t.Fatal(err)
+					}
+				}
+				runTest("testdata/libc/env.go", "", t)
```
aykevl: This modifies global state, which I think is a bad idea. Perhaps you can change the second parameter of `runtTest` to accept a slice of environment variables?

```suggestion
				runTest("testdata/libc/env.go", "", t, []string{"ENV1=VALUE1", "ENV2=VALUE2"})
```

main_test.go
```diff
@@ -104,6 +112,8 @@ func TestCompiler(t *testing.T) {
 
 		t.Run("WASI", func(t *testing.T) {
 			runPlatTests("wasi", matches, t)
+			runTest("testdata/libc/env.go", "wasi", t,
+				[]string{"--env", "ENV1=VALUE1", "--env", "ENV2=VALUE2"}...)
```
aykevl: Same here:

```suggestion
			runTest("testdata/libc/env.go", "wasi", t, []string{"ENV1=VALUE1", "ENV2=VALUE2"})
```

main_test.go
```diff
@@ -191,7 +200,7 @@ func runTest(path, target string, t *testing.T) {
 			cmd = exec.Command(binary)
 		} else {
 			args := append(spec.Emulator[1:], binary)
-			cmd = exec.Command(spec.Emulator[0], args...)
+			cmd = exec.Command(spec.Emulator[0], append(args, additionalArgs...)...)
 		}
 	}
```
aykevl, line 208: Here you could do something like:

```go
if len(spec.Emulator) != 0 && spec.Emulator[0] == "wasmtime" {
    for _, v := range environmentVars {
        cmd.Args = append(cmd.Args, "--env", v)
    }
} else {
    cmd.Env = append(cmd.Env, environmentVars...)
}
```

This only sets the environment variables for the newly created process.

aykevl (commented):
Looking good, I'm only concerned by the `os.Setenv` call which I think can be easily avoided.

mathetake:
resolved & squashed :+1:

mathetake:
kindly ping @aykevl

deadprogram:
Looking good now @mathetake thank you very much for this contribution! Now merging.

PR #1639, opened 2021-02-17
machine/nrf52840: rename usb support files to be consistent

deadprogram:
This PR renames the usb support files for the nrf52840 to be consistent with other naming by architecture/chip.


aykevl:
You mean this is more consistent with files such as machine_stm32_i2c.go and machine_mimxrt1062_uart.go?
In that case, sure, looks fine by me. Not something I really thought about but I agree it's a bit more consistent.

deadprogram:
Thanks @aykevl now merging.

Issue #1637, opened 2021-02-17
samd : If you disconnect USBCDC after communication is established, it will cause an infinite loop.


sago35:
I thought `usbLineInfo.lineState` would return to 0 after disconnecting communication, but it did not.
Even when it returns to 0, USBCDC.WriteByte() will go into an infinite loop depending on the timing of the disconnect.

* #1636
* #1481

We will discuss how to fix it.
A simple solution would be to stop sending if it takes more than Xmsec to send a single message.

sago35:
If the connection is never established, or never disconnected, there is no problem.

deadprogram:
This has been released with v0.17.0 so now closing. Thanks!

PR #1640, opened 2021-02-18
atsamd5x: avoid infinite loop when USBCDC is disconnected

sago35:
Fix issue #1637 

Samd2x should be able to be fixed in the same way.
Once we have reviewed this PR, we will create a PR.

### DIFF
--- a/src/machine/machine_atsamd51.go
+++ b/src/machine/machine_atsamd51.go
@@ -1734,17 +1734,19 @@ func (pwm PWM) getMux() PinMode {
 
 // USBCDC is the USB CDC aka serial over USB interface on the SAMD21.
 type USBCDC struct {
-	Buffer  *RingBuffer
-	TxIdx   volatile.Register8
-	waitTxc bool
-	sent    bool
+	Buffer            *RingBuffer
+	TxIdx             volatile.Register8
+	waitTxc           bool
+	waitTxcRetryCount uint8
+	sent              bool
 }
 
 const (
-	usbcdcTxSizeMask uint8 = 0x3F
-	usbcdcTxBankMask uint8 = ^usbcdcTxSizeMask
-	usbcdcTxBank1st  uint8 = 0x00
-	usbcdcTxBank2nd  uint8 = usbcdcTxSizeMask + 1
+	usbcdcTxSizeMask          uint8 = 0x3F
+	usbcdcTxBankMask          uint8 = ^usbcdcTxSizeMask
+	usbcdcTxBank1st           uint8 = 0x00
+	usbcdcTxBank2nd           uint8 = usbcdcTxSizeMask + 1
+	usbcdcTxMaxRetriesAllowed uint8 = 5
 )
 
 // Flush flushes buffered data.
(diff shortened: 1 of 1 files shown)


src/machine/machine_atsamd51.go
```diff
@@ -1734,17 +1734,19 @@ func (pwm PWM) getMux() PinMode {
...
+	waitTxcTimer uint8
+	sent         bool
 }
 
 const (
 	usbcdcTxSizeMask uint8 = 0x3F
 	usbcdcTxBankMask uint8 = ^usbcdcTxSizeMask
 	usbcdcTxBank1st  uint8 = 0x00
 	usbcdcTxBank2nd  uint8 = usbcdcTxSizeMask + 1
+	usbcdcTxTimeout  uint8 = 5
```
sago35: The timeout was set at about 5 ms.
It usually takes less than 1 ms to transmit.

src/machine/machine_atsamd51.go
```diff
@@ -1806,6 +1809,14 @@ func (usbcdc *USBCDC) WriteByte(c byte) error {
 
 			if ok {
 				break
+			} else if usbcdcTxTimeout < UART0.waitTxcTimer {
+				mask := interrupt.Disable()
+				UART0.waitTxc = false
+				UART0.waitTxcTimer = 0
+				usbcdc.TxIdx.Set(0)
+				usbLineInfo.lineState = 0
```
sago35, line 1817: I don't know if I can initialize `usbLineInfo.lineState` .
However, in my environment, there was no problem because cdcSetup() was called when I connected again.

env: windows10 + teraterm 4.105

src/machine/machine_atsamd51.go
```diff
@@ -1734,17 +1734,19 @@ func (pwm PWM) getMux() PinMode {
...
 // USBCDC is the USB CDC aka serial over USB interface on the SAMD21.
 type USBCDC struct {
-	Buffer  *RingBuffer
-	TxIdx   volatile.Register8
-	waitTxc bool
-	sent    bool
+	Buffer       *RingBuffer
+	TxIdx        volatile.Register8
+	waitTxc      bool
+	waitTxcTimer uint8
```
deadprogram: Perhaps this might be better named `waitTxcRetryCount` ?

src/machine/machine_atsamd51.go
deadprogram: Perhaps this might be better named `usbcdcTxMaxRetriesAllowed` ?

src/machine/machine_atsamd51.go
deadprogram: > there was no problem because cdcSetup() was called when I connected again

That is what I would expect to happen when it is working correctly. I will try this branch on Linux/macOS.

src/machine/machine_atsamd51.go
sago35: Thank you.
You're right, it's retry, not timeout.

src/machine/machine_atsamd51.go
sago35: Thank you.
You're right, it's retry, not timeout.

src/machine/machine_atsamd51.go
deadprogram: Seems to work as expected on both.

deadprogram:
Looks and works as expected. Can you please squash commits @sago35 thanks!

sago35:
Squashed. @deadprogram

deadprogram:
Thanks @sago35 now merging.

PR #1641, opened 2021-02-19
atsamd2x: avoid infinite loop when USBCDC is disconnected

sago35:
Fix issue #1637


deadprogram:
Thanks @sago35 now merging.

PR #1644, opened 2021-02-20
Fix typo in wasm_exec.js

ybkimm:
There seems to be a typo in the `syscall/js.valueLoadString` implementation.

https://github.com/tinygo-org/tinygo/blob/b689f14bb2c49c39fdd958be318ef3fb2e6d5ffb/targets/wasm_exec.js#L409-L412

I think it should be `loadValue(v_addr)`, not `v_attr`.


deadprogram:
Looks like a pretty obvious typo to me, thank you for finding and fixing.

Now merging.

PR #1655, opened 2021-02-22
interp: support GEP on fixed (MMIO) addresses

aykevl:
GetElementPtr would not work on values that weren't pointers. Because fixed addresses (often used in memory-mapped I/O) are integers rather than pointers in interp, it would return an error.

This resulted in the teensy40 target not compiling correctly since the interp package rewrite. This commit should fix that.

This should fix the issue in https://github.com/tinygo-org/tinygo/pull/1650.

### DIFF
--- a/interp/errors.go
+++ b/interp/errors.go
@@ -11,19 +11,17 @@ import (
 	"tinygo.org/x/go-llvm"
 )
 
-var errLiteralToPointer = errors.New("interp: trying to convert literal value to pointer")
-
 // These errors are expected during normal execution and can be recovered from
 // by running the affected function at runtime instead of compile time.
 var (
-	errExpectedPointer        = errors.New("interp: trying to use an integer as a pointer (memory-mapped I/O?)")
+	errIntegerAsPointer       = errors.New("interp: trying to use an integer as a pointer (memory-mapped I/O?)")
 	errUnsupportedInst        = errors.New("interp: unsupported instruction")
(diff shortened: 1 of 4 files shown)


deadprogram:
Has a few problems on Xtensa it seems: https://app.circleci.com/pipelines/github/tinygo-org/tinygo/4393/workflows/5352a6dd-7b6c-41d9-8203-614d32dc9111/jobs/24204/parallel-runs/0/steps/0-117?invite=true

aykevl:
That's not on Xtensa, that's on the pca10040.
...perhaps I should have run a few more tests before submitting this PR.

aykevl:
@ardnew if you are available, can you maybe re-test this PR with the Teensy40? I don't think my recent change would affect the Teensy 4.0, but it would be good to check just in case.

deadprogram:
I just tried the Teensy 4.0 with this branch and it did not correct the issue.

aykevl:
Did you try the other UART (not UART0)?

ardnew:
There are some changes I've made to the Teensy 4.0 that could help address this issue. I've found that `UART.txBuffer` is getting cleaned up too early, causing a segfault in the interrupt handler. Moving its allocation to the `UART` declaration (same place `UART.Buffer` is allocated) fixes the issue. 

Have there been changes to the garbage collector? I'm not sure which commit to look for that would have introduced a change like this.

deadprogram:
Tested on Teensy 4.0 with the revised #1650 so this PR here can now be merged.

PR #1656, opened 2021-02-23
build: add apt update that seems needed for when dependencies update …

deadprogram:
This PR adds an apt update to steps that seem to need them for when dependencies update or packages are removed.


deadprogram:
We need this in order to correct the current problems on CircleCI.

deadprogram:
Going to merge to fix the failing builds...

PR #1636, opened 2021-02-16
atsamd2x: improve USBCDC

sago35:
This PR improves the USBCDC speed of atsamd2x.
The atsamd5x version is #1481 .

I checked it on Windows 10 and found the following

* 13 byte : fmt.Printf("hello world\r\n")
  * old : 648.298 us
  * new :  88.284 us
* 11 byte : println("hello world")
  * old :  592.844 us
  * new :  22.224 us
* 1024 byte + "\r\n"
  * old :  51.735 ms
  * new :  1.906 ms
* 1024 * 10 byte + "\r\n"
    * old : 513.601 ms
    * new : 18.888ms

PR #1660, opened 2021-02-25
nrf52840: improve USBCDC

sago35:
This PR improves the USBCDC speed of nrf52840.
The atsamd5x version is #1481 .
The atsamd2x version is #1636  .

* 13 byte : fmt.Printf("hello world\r\n")
  * old : 631 us
  * new : 64 us
* 11 byte : println("hello world")
  * old : 320 us
  * new : 15 us
* 1024 byte + "\r\n"
  * old : 28.746 ms
  * new : 1.249 ms
* 1024 * 10 byte + "\r\n"
  * old : 515.017 ms
  * new : 12.746 ms

windows 10 + feather-nrf52840

### DIFF
--- a/src/machine/usb_nrf52840.go
+++ b/src/machine/usb_nrf52840.go
@@ -12,22 +12,100 @@ import (
 
 // USBCDC is the USB CDC aka serial over USB interface on the nRF52840
 type USBCDC struct {
-	Buffer       *RingBuffer
-	interrupt    interrupt.Interrupt
-	initcomplete bool
+	Buffer            *RingBuffer
+	interrupt         interrupt.Interrupt
+	initcomplete      bool
+	TxIdx             volatile.Register8
+	waitTxc           bool
+	waitTxcRetryCount uint8
+	sent              bool
+}
+
+const (
+	usbcdcTxSizeMask          uint8 = 0x3F
+	usbcdcTxBankMask          uint8 = ^usbcdcTxSizeMask
+	usbcdcTxBank1st           uint8 = 0x00
+	usbcdcTxBank2nd           uint8 = usbcdcTxSizeMask + 1
+	usbcdcTxMaxRetriesAllowed uint8 = 5
+)
+
+// Flush flushes buffered data.
+func (usbcdc *USBCDC) Flush() error {
+	if usbLineInfo.lineState > 0 {
+		idx := usbcdc.TxIdx.Get()
+		sz := idx & usbcdcTxSizeMask
+		bk := idx & usbcdcTxBankMask
+		if 0 < sz {
+
+			if usbcdc.waitTxc {
+				// waiting for the next flush(), because the transmission is not complete
+				usbcdc.waitTxcRetryCount++
+				return nil
+			}
+			usbcdc.waitTxc = true
+			usbcdc.waitTxcRetryCount = 0
+
+			// set the data
+			enterCriticalSection()
+			sendViaEPIn(
+				usb_CDC_ENDPOINT_IN,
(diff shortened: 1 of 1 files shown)


sago35:
Even in the unmodified version, it often crashes immediately after tinygo-flash.
When it crashes, I sometimes get the following message

> panic: runtime error: goroutine stack overflow

After resetting several times, the problem will no longer occur.

src/machine/usb_nrf52840.go
```diff
@@ -117,6 +196,11 @@ func (usbcdc *USBCDC) Configure(config UARTConfig) {
 }
 
 func (usbcdc *USBCDC) handleInterrupt(interrupt.Interrupt) {
+	if nrf.USBD.EVENTS_SOF.Get() == 1 {
+		nrf.USBD.EVENTS_SOF.Set(0)
+		UART0.Flush()
+	}
+
```
sago35: Added SOF interrupt because I wanted periodic processing every 1ms

src/machine/usb_nrf52840.go
```diff
@@ -12,22 +12,100 @@ import (
...
-		sendViaEPIn(
-			usb_CDC_ENDPOINT_IN,
-			&udd_ep_in_cache_buffer[usb_CDC_ENDPOINT_IN][0],
-			1,
-		)
+		ok := false
+		for {
+			mask := interrupt.Disable()
+
+			idx := UART0.TxIdx.Get()
```
sago35: In this function, access is made in the form UART0.XXX instead of usbcdc.XXX.
I don't know why the usbcdc.XXX access method is slow, but it is.

https://twitter.com/sago35tk/status/1363850766486675458

src/machine/usb_nrf52840.go
```diff
@@ -12,22 +12,100 @@ import (
...
+func (usbcdc *USBCDC) Flush() error {
+	if usbLineInfo.lineState > 0 {
+		idx := usbcdc.TxIdx.Get()
+		sz := idx & usbcdcTxSizeMask
+		bk := idx & usbcdcTxBankMask
+		if 0 < sz {
+
+			if usbcdc.waitTxc {
+				// waiting for the next flush(), because the transmission is not complete
+				usbcdc.waitTxcRetryCount++
```
sago35, line 42: Unlike samd2x/samd5x, I decided to increment the `waitTxcRetryCount` here.
I think it would be better to unify the samd code as well.

src/machine/usb_nrf52840.go
sago35: The reelboard test was failing, so I rewrote it to USB instead of UART0.

deadprogram:
Once thing I have noticed is that you really need to update nrf52840 bootloader to the latest version from Adafruit (0.4.0) to avoid issues with flashing. See https://github.com/tinygo-org/tinygo-site/pull/143 for info on how to do this.

sago35:
Updated to 0.4.1.
When jlink is not used, there seems to be no problem.
After writing from jlink, it looks like I need to perform a reset several times myself.

deadprogram:
I have not tried to use jlink on this board, but that does not entirely surprise me. Might need an issue reported upstream to the bootloader repo.

src/machine/usb_nrf52840.go
deadprogram: Good idea.

deadprogram:
Absolutely works better/faster @sago35 thanks for working on this! Now merging.

deadprogram:
I meant to squash this first, but was too quick on the click. Sorry about that.

PR #1663, opened 2021-02-27
Revert "Allow interrupts in stm32f103xx (#1466)" as discussed in #1608

deadprogram:
This reverts commit 3bb994da9f42118fe9cfcc8945e8ec9d3b01da27 from #1466 as discussed in #1608


aykevl:
Thank you!

PR #1667, opened 2021-02-27
stm32: add i2c for stm32f407

kenbell:
Verified on a STM32F407G-DISC1 board using a BME280 sensor + SSD1306 led display over i2c.

The new code in machine_stm32f407.go is a direct copy of the i2c code in machine_stm32f405.go - at some point this logic should probably be harmonized.

The fixes to stm32_i2c are enough to make this work for me.  I have found that a startup delay before initializing i2c is necessary - i suspect there may be power/voltage transients or similar causing issues if i2c is enabled too quickly after boot.  Why i think this: (at least on my disco board)
- startup delay not required with debugging hardware attached (i2c logic analyzer, scope, UART)
- UART has a tendency to back-feed the board
- The delay is not needed for a soft reset
- I've occasionally seen a 'double boot' on power-up where it looks like the MCU does a reset (possible 'brown-out'?)

### DIFF
--- a/src/machine/board_stm32f4disco.go
+++ b/src/machine/board_stm32f4disco.go
@@ -66,3 +66,15 @@ var (
 	}
 	SPI1 = &SPI0
 )
+
+const (
+	I2C0_SCL_PIN = PB6
+	I2C0_SDA_PIN = PB9
+)
+
+var (
+	I2C0 = I2C{
+		Bus:             stm32.I2C1,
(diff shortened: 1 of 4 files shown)


deadprogram:
Great work @kenbell the I2C0 port is now working perfectly and the UART is still working, so I would call that success!

Now merging, thank you very much for working on this.

PR #1668, opened 2021-02-27
Libc open read close

oflebbe:
Provide a simple implementation for open, read and close for libc backed systems.

Not to sure about the data types to use.
Reading is really slow, but at least seems to work on MacOS.


oflebbe:
pull to wrong branch

PR #1659, opened 2021-02-24
WASI & darwin: support basic file io based on libc

mathetake:
This is the first PR towards file system support in libc target.

resolves #1505

PR #1669, opened 2021-02-27
libc: provide implementation for open, read, close

oflebbe:
Implement simple wrappers for open, read and close for libc backed system.

Not too sure about data types to use.
At least it seems to work on MacOS/Darwin, though very slow.

### DIFF
--- a/src/syscall/syscall_libc.go
+++ b/src/syscall/syscall_libc.go
@@ -13,7 +13,12 @@ type sliceHeader struct {
 }
 
 func Close(fd int) (err error) {
-	return ENOSYS // TODO
+	n := libc_close(int32(fd))
+
+	if n < 0 {
+		return getErrno()
+	}
+	return nil
 }
 
 func Write(fd int, p []byte) (n int, err error) {
@@ -26,7 +31,12 @@ func Write(fd int, p []byte) (n int, err error) {
 }
 
 func Read(fd int, p []byte) (n int, err error) {
-	return 0, ENOSYS // TODO
+	buf, count := splitSlice(p)
+	n = libc_read(int32(fd), buf, uint(count))
+	if n < 0 {
+		err = getErrno()
+	}
+	return
 }
 
 func Seek(fd int, offset int64, whence int) (off int64, err error) {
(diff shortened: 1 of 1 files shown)


src/syscall/syscall_libc.go
```diff
@@ -26,15 +31,29 @@ func Write(fd int, p []byte) (n int, err error) {
...
 
 func Seek(fd int, offset int64, whence int) (off int64, err error) {
 	return 0, ENOSYS // TODO
 }
 
 func Open(path string, mode int, perm uint32) (fd int, err error) {
-	return 0, ENOSYS // TODO
+	buf, count := splitSlice([]byte(path))
+	cstr := libc_strndup(buf, count)
+	defer libc_free(cstr)
```
aykevl, line 49: You can avoid a heap allocation here by avoiding the string to []byte conversion. You can create a new `splitString` function and pass the resulting pointer directly to `libc_strndup`.

src/syscall/syscall_libc.go
```diff
@@ -76,10 +95,28 @@ func splitSlice(p []byte) (buf *byte, len uintptr) {
 	return slice.buf, slice.len
 }
 
+// int open(const char *path, int oflag, int perm)
+//export open
+func libc_open(path *byte, oflag int, perm uint32) int
```
aykevl, line 100: `int` in C is practically always 32-bit (int32) on both 32-bit and 64-bit systems, while `int` in Go is 64-bit on 64-bit systems. Therefore, this signature is subtly wrong in an almost-compatible way. This is better:

```suggestion
func libc_open(path *byte, oflag int32, perm uint32) int32
```

aykevl (commented):
The types look almost correct.

>  At least it seems to work on MacOS/Darwin, though very slow.

How did you test this?
It would be great to have some tests for this, possibly including Linux.

src/syscall/syscall_libc.go
oflebbe: Where is a suitable place in the file tree for a test ? 
Can we use go test framework ?

deadprogram:
Is there any duplication with this PR and https://github.com/tinygo-org/tinygo/pull/1659 ?

oflebbe:
Oh yes silly me. Didn't checked the open PR's first . Seems like  #1659 seem to contain my proposal  .

oflebbe:
Withdrawing PR in favor of #1659

PR #1670, opened 2021-02-28
gdb: enable to specify multiple candidates for gdb

### LINKED ISSUE #1662: How to do tinygo-gdb on windows
The following changes make me use gdb-multiarch, but I still think I need to use arm-none-eabi-gdb.exe on windows.

* 1a2f051b

If rename arm-none-eabi-gdb.exe to gdb-multiarch.exe, it can be debugged with tinygo-gdb.
However, I think it is better to specify it with an argument like `tinygo gdb -debugger=arm-none-eabi-gdb ...` rather than renaming it.
Since I already created something that works, I will create a Pull-Request if this approach is okay.

or,,,
Is there any other better way?

windows 10 + tinygo 0.16 or later + cortex-m4

```
error: failed to run gdb with C:\Users\sago3\AppData\Local\Temp\tinygo160016967\main: exec: "gdb-multiarch": executable file not found in
 %PATH%
```

arm-neno-eabi-gdb for windows can be found at:
https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-rm/downloads

sago35:
fixes #1662 (How to do tinygo-gdb on windows)

### DIFF
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"io"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"reflect"
 	"runtime"
@@ -42,7 +43,7 @@ type TargetSpec struct {
 	ExtraFiles       []string `json:"extra-files"`
 	Emulator         []string `json:"emulator" override:"copy"` // inherited Emulator must not be append
 	FlashCommand     string   `json:"flash-command"`
-	GDB              string   `json:"gdb"`
+	GDB              []string `json:"gdb"`
 	PortReset        string   `json:"flash-1200-bps-reset"`
 	FlashMethod      string   `json:"flash-method"`
 	FlashVolume      string   `json:"msd-volume-name"`
@@ -240,7 +241,7 @@ func defaultTarget(goos, goarch, triple string) (*TargetSpec, error) {
 		Compiler:  "clang",
 		Linker:    "cc",
 		CFlags:    []string{"--target=" + triple},
-		GDB:       "gdb",
+		GDB:       []string{"gdb"},
 		PortReset: "false",
 	}
 	if goos == "darwin" {
@@ -253,7 +254,7 @@ func defaultTarget(goos, goarch, triple string) (*TargetSpec, error) {
 	}
 	if goarch != runtime.GOARCH {
 		// Some educated guesses as to how to invoke helper programs.
-		spec.GDB = "gdb-multiarch"
+		spec.GDB = []string{"gdb-multiarch"}
 		if goarch == "arm" && goos == "linux" {
 			spec.CFlags = append(spec.CFlags, "--sysroot=/usr/arm-linux-gnueabihf")
 			spec.Linker = "arm-linux-gnueabihf-gcc"
@@ -270,4 +271,18 @@ func defaultTarget(goos, goarch, triple string) (*TargetSpec, error) {
 		}
 	}
 	return &spec, nil
+}
+
+// LookupGDB looks up a gdb executable.
+func (spec *TargetSpec) LookupGDB() (string, error) {
+	if len(spec.GDB) == 0 {
+		return "", errors.New("gdb not configured in the target specification")
+	}
+	for _, d := range spec.GDB {
+		_, err := exec.LookPath(d)
+		if err == nil {
+			return d, nil
+		}
+	}
+	return "", errors.New("no gdb found configured in the target specification (" + strings.Join(spec.GDB, ", ") + ")")
 }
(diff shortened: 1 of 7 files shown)


compileopts/target.go
```diff
@@ -271,3 +272,16 @@ func defaultTarget(goos, goarch, triple string) (*TargetSpec, error) {
 	}
 	return &spec, nil
 }
+
+func (spec *TargetSpec) Debugger() (string, error) {
```
aykevl: Can you rename this to `GDB`?

```suggestion
func (spec *TargetSpec) GDB() (string, error) {
```

We might want to add support for other debuggers in the future, such as LLDB (which can possibly be included in the TinyGo release, unlike GDB). LLDB will need special support, it is not a drop-in replacement.

main.go
```diff
@@ -484,11 +485,11 @@ func FlashGDB(pkgName string, ocdOutput bool, options *compileopts.Options) erro
 		for _, cmd := range gdbCommands {
 			params = append(params, "-ex", cmd)
 		}
-		cmd := executeCommand(config.Options, config.Target.GDB, params...)
+		cmd := executeCommand(config.Options, gdb, params...)
 		cmd.Stdin = os.Stdin
 		cmd.Stdout = os.Stdout
 		cmd.Stderr = os.Stderr
-		err := cmd.Run()
+		err = cmd.Run()
```
aykevl: Was this intentional?

compileopts/target.go
sago35: If I rename it to GDB(), I get the following error.
Is there any other name that would be better?

```
compileopts\target.go:276:6: type TargetSpec has both field and method named GDB
```

main.go
sago35: It was an unnecessary change.
Fixed.

compileopts/target.go
aykevl: Ah, I see. I can't think of a good name, maybe `GDBCommand` or `LookupGDB`?

sago35: I renamed Debugger() to LookupGDB().

compileopts/target.go
```diff
@@ -253,7 +254,7 @@ func defaultTarget(goos, goarch, triple string) (*TargetSpec, error) {
 	}
 	if goarch != runtime.GOARCH {
 		// Some educated guesses as to how to invoke helper programs.
-		spec.GDB = "gdb-multiarch"
+		spec.GDB = []string{"gdb-multiarch", "arm-none-eabi-gdb"}
```
aykevl: This change is not correct. This code is for regular desktop usage, where `gdb-multiarch` is usually the correct name and `arm-none-eabi-gdb` never is.

compileopts/target.go
```diff
@@ -271,3 +272,17 @@ func defaultTarget(goos, goarch, triple string) (*TargetSpec, error) {
...
+	if len(spec.GDB) == 0 {
+		return "", errors.New("gdb not configured in the target specification")
+	}
+	for _, d := range spec.GDB {
+		_, err := exec.LookPath(d)
+		if err == nil {
+			return d, nil
+		}
+	}
+	return "", errors.New("no gdb found configured in the target specification")
```
aykevl: Can you maybe include the command names that were checked in the error message? That makes it easier to know how to fix the issue based on the error message.

targets/cortex-m7.json
```diff
@@ -5,5 +5,5 @@
 		"--target=armv7em-none-eabi",
 		"-Qunused-arguments"
 	],
-	"gdb": "arm-none-eabi-gdb"
+	"gdb": ["arm-none-eabi-gdb"]
```
aykevl: Strictly speaking this line is not necessary, but it was already present before so it's fine (no change needed). But you can remove it if you want.

compileopts/target.go
sago35: Added information to error message.

```
error: no gdb found configured in the target specification (gdb-multiarch, arm-none-eabi-gdb)
```

compileopts/target.go
sago35: arm-none-eabi-gdb has been removed.

targets/cortex-m7.json
sago35: Removed because it is defined in the inherited cortex-m.json

sago35:
Please review this PR.
 @aykevl

oflebbe:
This patch is needed for MacOS as well. +1

deadprogram:
I think this PR seems OK to merge if squashed. @aykevl what do you think?

deadprogram:
@sago35 can you please squash this PR so we can merge it?

@aykevl any further feedback?

sago35:
I squashed and rebased it. @deadprogram

deadprogram:
Thanks @sago35 now merging.

sago35:
Thank you!

Issue #1678, opened 2021-03-04
Document make wasi-libc


fgsch:
This is not mentioned anywhere I can see. It would be a good idea to document it.

aykevl:
Please take a look at https://github.com/tinygo-org/tinygo/pull/1679 which should make the error message a lot clearer and more helpful. For example, I got this error message after removing the libc:

```
$ go install; and tinygo run -target=wasi ./testdata/cgo
error: could not find wasi-libc, perhaps you need to run `make wasi-libc`?
```

That's a lot more helpful than a linker error.

fgsch:
@aykevl I will take a look. Thank you.

Can we still document this step in the building instructions?

aykevl:
Done: https://github.com/tinygo-org/tinygo-site/pull/150

fgsch:
Re #1679, works as advertised. Thank you!

deadprogram:
This has been released with v0.17.0 so now closing. Thanks!

PR #1679, opened 2021-03-04
wasi: specify wasi-libc in a different way

aykevl:
This way is more consistent with how picolibc is specified and allows
generating a helpful error message. This error message should never be
generated for TinyGo binary releases, only when doing local development.

This should improve https://github.com/tinygo-org/tinygo/issues/1678. It's a common question so it seems like a idea to me to make the error message more helpful.


deadprogram:
Very good idea @aykevl now merging.

PR #1680, opened 2021-03-04
all: remove support for LLVM 9

aykevl:
This LLVM version breaks CI and is now relatively rather old anyway, so remove support for it.

This also reverts a workaround for LLVM 9, see a9568932b ("maixbit: workaround to avoid medium code model").

I have not yet removed Go 1.11 and Go 1.12 support, because it might be useful for people running Debian stable (which has Go 1.11 in the repository). We will do that after the release.


deadprogram:
A good step forward @aykevl thanks! :scissors: 

Merging.

PR #1685, opened 2021-03-05
gdb: support -ocd-output on windows

sago35:
To make the escape sequence ansi coloring work on windows, I used go-colorable.
With go-colorable, there is no need to write branches for windows and other operating systems.
https://github.com/mattn/go-colorable

before:
image

after:
image


aykevl:
Looks good to me, thank you!

Issue #1634, opened 2021-02-15
Package runtime/debug is not in GOROOT 


leobragaz:
Hi everyone, I was looking at [this](https://github.com/tinygo-org/tinygo/issues/1472) and it was closed apparently because it was related to problems in supporting wasm with older go version but I'm having the same issue with go `1.15`

```
package runtime/debug is not in GOROOT (/Users/leonardo/Library/Caches/tinygo/goroot-go1.15.5-093539b4bdf036f3692361094b6ad0277884029bf081496eed0130db628030c1/src/runtime/debug)
package command-line-arguments
        imports github.com/cosmos/cosmos-sdk/crypto/hd
        imports github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1
        imports github.com/cosmos/cosmos-sdk/codec
        imports github.com/cosmos/cosmos-sdk/codec/types
```
Is it still unsupported?

yaron2:
I'm getting the same thing with Go 1.15.

aykevl:
Can you give the code that triggers it so I can reproduce the issue?

leobragaz:
https://github.com/desmos-labs/go-key-manager

aykevl:
I have created a dummy package here: https://github.com/tinygo-org/tinygo/pull/1687

If you try it, you'll see that the next build failure is the lack of encoding/json support.

dgryski:
Closing.  We should file issues for the pieces from the `debug` package we actually need.

PR #1687, opened 2021-03-05
runtime: add dummy debug package

aykevl:
This package does not implement any methods, which is of course not useful. However, by creating this package in advance it's possible to see the next issue that's preventing something from building in TinyGo.

Motivated by: https://github.com/tinygo-org/tinygo/issues/1634

### DIFF
--- a/src/runtime/debug/debug.go
+++ b/src/runtime/debug/debug.go
@@ -1,0 +1,2 @@
+// Package debug is a dummy package that is not yet implemented.
+package debug


ardnew:
This may actually resolve a long-standing bug with TinyGo and the GoSublime (`margo.sh`) plugin for SublimeText. A missing `runtime/debug` package prevents their tooling from recognizing valid GOROOTs. 

Dunno if a dummy/stub like this is sufficient, or if it actually expects some package content. Will verify once this gets merged.

deadprogram:
Merging, let us see what it exposes as a next step.

PR #1689, opened 2021-03-05
builder: add support for -opt=0

aykevl:
This optimization level wasn't working before because some passes expect some globals to be cleaned up afterwards. Cleaning these globals is easy, just add the pass necessary for it. This shouldn't reduce the usefulness of the -opt=0 build flag as most optimizations are still skipped.


deadprogram:
Makes sense @aykevl now merging.

PR #1690, opened 2021-03-05
wasm: switch to wasi_snapshot_preview1

aykevl:
This is a hopefully more stable interface and it appears more people are using wasi_snapshot_preview1 than wasi_unstable, so let's switch to it.


aykevl:
https://github.com/tinygo-org/tinygo/pull/1691 looks like a better version, which includes adjustments to ABI changes in this snapshot.

PR #1692, opened 2021-03-05
attiny: remove dummy UART

aykevl:
I think it's better not to provide a UART0 global at all than one that does nothing.


deadprogram:
Makes sense. Merging. :scissors:

PR #1698, opened 2021-03-08
transform: optimize string comparisons against ""

aykevl:
This optimizes a common pattern like:

    if s != "" {
        ...
    }

to:

    if len(s) != 0 {
        ...
    }

This avoids a runtime call and thus produces slightly better code. With testdata/stdlib.go I saw two instances of this and the optimization reduced binary size by 80 bytes (with `-target=cortex-m-qemu`). That's admittedly rather low, but the main goal of this optimization is so that the most readable code can be written in all cases and the compiler optimizes it as needed.


deadprogram:
:zero: merging

PR #1699, opened 2021-03-08
Test runtime calls in the compiler

aykevl:
This PR adds support for testing runtime calls in the compiler package. This allows us to add many interesting tests to the compiler. I've started out with some slice and string tests.


aykevl:
Updated. Hopefully tests now pass. I've changed the tests of the compiler package to only test on LLVM 11, which makes it much easier to write complex but readable tests.

aykevl:
Rebased to fix merge conflicts after #1612 was merged.

deadprogram:
This is a very good idea! Now merging thanks @aykevl

PR #1701, opened 2021-03-09
docs: add QT Py and Teensy 4.0

sago35:


deadprogram:
Oops, we missed a couple! :smile_cat: 

Thanks @sago35

Issue #1658, opened 2021-02-24
Tinygo on Linux uses go version 0.1


allthingssecurity:
When I try to run a tinygo program it says it is using version 0.1 of go instead of something between 1.11 and 1.15.
I tried deb package of both tinygo 0.15 and 0.16 but I get same error

deadprogram:
How exactly did you produce that version? I ask because normally you can tell by running the command `tinygo version` like this:

```shell
$ tinygo version
tinygo version 0.17.0-dev-418ab8b linux/amd64 (using go version go1.15 and LLVM version 11.0.0)
```

I am not aware of any other way to do so.

raqbit:
When you set the tinygo cached GOROOT (as found by `tinygo info <board>`) as your GOROOT in GoLand, it also shows `v0.1.0` as the go version, not sure why, but probably related.

aykevl:
In that case I think GoLand picks up the wrong version number.

TinyGo will use the Go installation that it finds by resolving the `go` command. I can assure you that it won't magically use Go 0.1, which is now very old.

raqbit:
Found it, Goland picks up the version number from `src/runtime/internal/sys/zversion.go` in `GOROOT` via a regex.

Found it [here](https://github.com/go-lang-plugin-org/go-lang-idea-plugin/blob/e73dab44213ffa76b3d6a853aecb109929c3e2b5/src/com/goide/sdk/GoSdkUtil.java#L372) in the legacy open source go-lang-idea-plugin, the code of which probably still lives in GoLand.

So it seems like https://github.com/tinygo-org/tinygo/blob/release/src/runtime/internal/sys/zversion.go should be linked to the go version that TinyGo finds.

aykevl:
Apparently I did not really understand the issue.

> When you set the tinygo cached GOROOT (as found by `tinygo info <board>`) as your GOROOT in GoLand

It looks like GoLand then also passes this cached GOROOT on to TinyGo, which expects the root of the Go installation instead of this cached GOROOT. See my comment on #1705 for more details.
This is indeed a bug. I could say that GoLand shouldn't pass on this GOROOT but that's actually sensible behavior. Instead, I think TinyGo should recognize this situation and deal with it (again, as detailed in #1705).

deadprogram:
This appears to be have been resolved. Please reopen if needed.

PR #1705, opened 2021-03-10
Use physical path instead of cached GOROOT in function getGoroot (fixes #433, #1658)

### LINKED ISSUE #433: Could not read version from GOROOT: /usr/local/go
Hello. Was grabbing `tinygo` manually when I had this:

```sh
$ tinygo build -o ../sc.wasm -target wasm32-unknown-unknown .
error: could not read version from GOROOT: /nix/store/ypfhc6f07z6ij7yn5hrz87s74jh7l8w2-go-1.12.1/share/go
```

Tried Docker as well:

```sh
$ docker run --rm -v "$PWD" tinygo/tinygo tinygo build -o ../sc.wasm -target wasm32-unknown-unknown .
Unable to find image 'tinygo/tinygo:latest' locally
latest: Pulling from tinygo/tinygo
11f34d1ed4d8: Pull complete 
b74da5bb8068: Pull complete 
babdd2d5058f: Pull complete 
f913f1fc17b5: Pull complete 
698bf23304a5: Pull complete 
4e4c4039cb7f: Pull complete 
fe88c201175d: Pull complete 
a27575ff07ae: Pull complete 
Digest: sha256:07dcaa92012e5f0b8c4e138db799e54a9f1b1729f8aa16c85d4232a9017dcea4
Status: Downloaded newer image for tinygo/tinygo:latest
error: could not read version from GOROOT: /usr/local/go
```

In case it matters, the repository is [diamondburned/smart-contract-go](https://github.com/diamondburned/smart-contract-go).

ardnew:
I keep running into the same error as others (#433 and #1658), where TinyGo resolves an incorrect Go runtime version, and generates the message:

```
error: requires go version 1.11 through 1.15, got go0.1
error: command failed
```

On Linux, this appears to be caused by the symlink created in the cached `GOROOT`:

```sh
andrew@trip ~ readlink -f $( tinygo info teensy40 | \grep -oP "GOROOT:\s*\K.+" )/src/runtime/internal/sys/zversion.go
/usr/local/src/tinygo/src/runtime/internal/sys/zversion.go
andrew@trip ~ cat /usr/local/src/tinygo/src/runtime/internal/sys/zversion.go
package sys

const TheVersion = `go0.1.0`
```

The problem is manifested here: https://github.com/tinygo-org/tinygo/blob/c60c36f0a8bbfd62fc61a583d4d2c7a8bf64bd58/goenv/version.go#L51-L52

This PR adds logic that tries to resolve a better path to `zversion.go` by checking if `GOROOT/bin` is a symlink (which it always will be inside TinyGo's cached `GOROOT`), and using the real, physical path to `GOROOT/bin` as the root directory for finding `src/runtime/internal/sys/zversion.go`. This way we are using the actual Go runtime directory's `zversion.go`, and not the one stored in TinyGo's sources.

Note that if this method fails, it falls back to using the original method currently implemented.

### DIFF
--- a/goenv/goenv.go
+++ b/goenv/goenv.go
@@ -4,6 +4,7 @@ package goenv
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"os/user"
@@ -221,19 +222,22 @@ func isCachedGoroot(goroot string) (bool, string) {
 		"pkg": os.ModeSymlink,
 	}
 
-	entry, _ := os.ReadDir(goroot)
-	for _, e := range entry {
-		if mode, ok := cacheEntry[e.Name()]; ok {
-			if mode != e.Type() {
+	info, err := ioutil.ReadDir(goroot)
+	if err != nil {
+		return false, ""
+	}
+	for _, f := range info {
+		if mode, ok := cacheEntry[f.Name()]; ok {
+			if mode != f.Mode().Type() {
 				return false, ""
 			}
 			// Remove the verified entry
-			delete(cacheEntry, e.Name())
+			delete(cacheEntry, f.Name())
 			if mode&os.ModeSymlink == 0 {
 				continue
 			}
 			// Entry is a symlink, get the physical GOROOT to which it points
-			root, err := os.Readlink(filepath.Join(goroot, e.Name()))
+			root, err := os.Readlink(filepath.Join(goroot, f.Name()))
 			if err != nil {
 				return false, ""
 			}
--- a/goenv/goenv.go
+++ b/goenv/goenv.go
@@ -228,7 +228,7 @@ func isCachedGoroot(goroot string) (bool, string) {
 	}
 	for _, f := range info {
 		if mode, ok := cacheEntry[f.Name()]; ok {
-			if mode != f.Mode().Type() {
+			if mode != f.Mode()&os.ModeType {
 				return false, ""
 			}
 			// Remove the verified entry
--- a/goenv/goenv.go
+++ b/goenv/goenv.go
@@ -4,7 +4,6 @@ package goenv
 
 import (
 	"fmt"
-	"io/fs"
 	"os"
 	"os/exec"
 	"os/user"
@@ -215,11 +214,11 @@ func isCachedGoroot(goroot string) (bool, string) {
 	// As these are encountered, they are deleted from the map. If, after
 	// reading all entries in the given path, this map has zero remaining
 	// elements, then we have a cached GOROOT.
-	cacheEntry := map[string]fs.FileMode{
-		"src": fs.ModeDir,
-		"bin": fs.ModeSymlink,
-		"lib": fs.ModeSymlink,
-		"pkg": fs.ModeSymlink,
+	cacheEntry := map[string]os.FileMode{
+		"src": os.ModeDir,
+		"bin": os.ModeSymlink,
+		"lib": os.ModeSymlink,
+		"pkg": os.ModeSymlink,
 	}
 
 	entry, _ := os.ReadDir(goroot)
@@ -230,7 +229,7 @@ func isCachedGoroot(goroot string) (bool, string) {
 			}
 			// Remove the verified entry
 			delete(cacheEntry, e.Name())
-			if mode&fs.ModeSymlink == 0 {
+			if mode&os.ModeSymlink == 0 {
 				continue
 			}
 			// Entry is a symlink, get the physical GOROOT to which it points
--- a/goenv/goenv.go
+++ b/goenv/goenv.go
@@ -4,6 +4,7 @@ package goenv
 
 import (
 	"fmt"
+	"io/fs"
 	"os"
 	"os/exec"
 	"os/user"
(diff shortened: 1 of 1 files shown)


aykevl:
I would argue the issue here is not in `GorootVersionString` but in how GOROOT is determined in the first place. A GOROOT set to the cached GOROOT (in ~/.cache/tinygo) will surely cause trouble. Therefore, I think the more straightforward solution here is to ignore the GOROOT environment variable if it points into the cache directory.

When writing this logic I had not considered that GOROOT could be set to the cached GOROOT of TinyGo. But now I see this, it does make sense to do so in an IDE (which might pass on this environment variable to tools it invokes).

ardnew:
> Therefore, I think the more straightforward solution here is to ignore the GOROOT environment variable if it points into the cache directory.

Without a properly-defined environment variable, how would you suggest a TinyGo GOROOT be resolved? Perhaps this is an unrecoverable error?

> A GOROOT set to the cached GOROOT (in ~/.cache/tinygo) will surely cause trouble.

I was under the impression that a GOROOT pointing to one of the cached directories (`~/.cache/tinygo/goroot-goVER-HASH-syscall`) was the _only_ permissible GOROOT definition for TinyGo - hence, this is why we use `tinygo info target` to define GOROOT when launching IDEs and tools. After all, these cached TinyGo GOROOTs are very different (that is, they have different content) from any installed standard Go's GOROOT.

aykevl:
There are three different GOROOT-like paths in TinyGo:

* The path which points to the root of the Go toolchain (GOROOT).
* The path which points to the root of the TinyGo toolchain (TINYGOROOT).
* The path which merges the two and is stored in the TinyGo cache (cached GOROOT).

TinyGo will try various things to obtain a valid GOROOT, first trying the GOROOT environment variable, then trying the location of the `go` command, and finally trying a few fallback paths. For details, see:

https://github.com/tinygo-org/tinygo/blob/138befd8a25d2a5c65e2e740038a8c3a44e95222/goenv/goenv.go#L142-L144

TinyGo will then merge the GOROOT and TINYGOROOT to a cached GOROOT, and will pass this path on to the `go list` command to resolve package dependencies. This cached GOROOT is also used in IDEs so that IDEs will resolve the correct dependencies (the machine package, for example).

The bug here is that this cached GOROOT is then set as the GOROOT environment variable when calling TinyGo. TinyGo expects the root of the Go toolchain but instead gets the cached GOROOT, which it then tries to merge _again_ with TINYGOROOT. Which of course gives a rather odd result.

I think the best way to fix this is by improving the logic of `getGoroot` to either detect this and ignore the passed in GOROOT (simple fix, perhaps not entirely correct) or maybe better to return the root of the Go installation by reading it from this cached GOROOT.

I hope that clears things up!

ardnew:
> I think the best way to fix this is by improving the logic of getGoroot to either detect this and ignore the passed in GOROOT (simple fix, perhaps not entirely correct) or maybe better to return the root of the Go installation by reading it from this cached GOROOT.

@aykevl I've updated this PR to implement your suggestion here (reverted my original commit). So instead of modifying `func GorootVersionString`, I've modified `getGoroot` as follows:

#### Previous logic
```
function getGoroot:
  if environment variable ${GOROOT} is defined:
    return ${GOROOT}
  ...
```

#### Changed logic
```
function getGoroot:
  if environment variable ${GOROOT} is defined:
    if the path at ${GOROOT} appears to be a TinyGo cache:
      return the standard Go GOROOT referenced by the TinyGo cache ${GOROOT}
    return ${GOROOT}
  ...
```

To identify a GOROOT as a TinyGo cache, it needs to meet the following conditions:
1. It is a directory containing at least the following files with exactly these file modes:
   - `src`: `ModeDir`
   - `bin`: `ModeSymlink`
   - `lib`: `ModeSymlink`
   - `pkg`: `ModeSymlink`
2. All of the above symlinks point to subdirectories of the _**exact same parent directory**_ (the GOROOT)

This resolved the issue for me locally, but I definitely need some help verifying these conditions are correct in general (not just Linux).

ardnew:
Perhaps we can place a file in the root of the TinyGo cache upon cache creation, or even in its parent directory (e.g., `~/.cache/tinygo/`), whose existence means that path is a cached GOROOT?

Or maybe we can reliably identify the path to where these caches are being created, and assume any subdirectory of that path is a cached GOROOT?

aykevl:
Yes, that code looks roughly correct. But I think it can be simplified a lot. Because TinyGo itself creates this cached GOROOT, we don't need to guess what's in there. So I think the following would be good enough:

```diff
 	goroot := os.Getenv("GOROOT")
 	if goroot != "" {
+		if symlink, err := os.Readlink(filepath.Join(goroot, "bin")); err == nil {
+			return filepath.Base(symlink)
+		}
 		return goroot
 	}
```

I think it's unlikely upstream Go will ever make the bin directory a symlink.

ardnew:
OK, so no need to really inspect the `GOROOT` that rigorously, I'll update the PR

(Also I believe you meant `filepath.Dir` instead of `filepath.Base`)

ardnew:
@aykevl Updated once more with your latest comments. Please review.

ardnew:
Been testing this change with all sorts of valid and invalid configurations:

- `GOROOT` environment variable
   - Using `tinygo` command manually 
- VSCode with TinyGo extension
   - Toggling between different targets
   - Setting `{"go.toolsEnvVars":{"GOROOT"}}` in `.vscode/settings.json` manually
- SublimeText3 with GoSublime, gopls, and goguru extensions
   - Setting `GOPATH` in margo configuration
   - Setting `GS_GOPATH` and other `GS_*` environment variables

Not once have I encountered the error message `got go0.1` again! I think this is a stable change.

aykevl:
Thank you! Looks good to me.

There was one failing test but that looks like a flaky test, unrelated to this PR.

PR #1707, opened 2021-03-12
builder: cache C and assembly file outputs 

aykevl:
This probably won't speed up the build on multicore systems (the build
is still dominated by the whole-program optimization step) but should be
useful at a later date for other optimizations. For example, I intend to
eventually optimize each package individually including C files, which
should enable cross-language optimizations (inlining C functions into Go
functions, for example). For that to work, accurate dependency tracking
is important.

Depends on #1612 so marked as WIP.


aykevl:
Rebased, this is now ready for review.

deadprogram:
I had to study this PR for a bit. Thanks for all the great comments @aykevl it was extremely helpful.

Now merging.

Issue #1624, opened 2021-02-10
Standalone C files don't get header search path from #cgo ?


oflebbe:
Hi,

thanks for hinting for using proper prototypes.

However, only slightly more involved commands still won't compile

See https://github.com/oflebbe/testesp32 

Maybe the CFLAGS flag isn't honored for stand-alone C files ?

aykevl:
> Maybe the CFLAGS flag isn't honored for stand-alone C files ?

Yes, that's true. Fixing this shouldn't be too hard, but it conflicts with https://github.com/tinygo-org/tinygo/pull/1612 so I'd like to get that PR ready first. In the meantime, you could work around it like this:

```diff
--- a/demo.c
+++ b/demo.c
@@ -1,4 +1,4 @@
-#include <foo.h>
+#include "include/foo.h"
 
 int fortytwo()
 {
```

Basically, do an include relative to the current directory.

aykevl:
Based on the repository name, you might be interested in this PR: https://github.com/tinygo-org/tinygo/pull/1626

aykevl:
#1626 does not fix this issue, so I've removed the next-release label.

oflebbe:
Indeed !  #1626 looks very interesting.
Wanted to poke into compiling and linking with ESP-IDF Headers and binaries .

aykevl:
Here is a fix: https://github.com/tinygo-org/tinygo/pull/1708

deadprogram:
Looks like this was addressed some time ago, so now closing. Please reopen if needed. Thank you!

PR #1708, opened 2021-03-12
cgo: add support for CFLAGS in .c files

aykevl:
This patch adds support for passing CFLAGS added in #cgo lines of the CGo preprocessing phase to the compiler when compiling C files inside packages. This is expected and convenient but didn't work before.

This fixes https://github.com/tinygo-org/tinygo/issues/1624. It is a draft because it needs #1612 to be merged first.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -430,7 +430,7 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 		job := &compileJob{
 			description: "compile extra file " + path,
 			run: func(job *compileJob) error {
-				result, err := compileAndCacheCFile(abspath, dir, config)
+				result, err := compileAndCacheCFile(abspath, dir, config.CFlags(), config)
 				job.result = result
 				return err
 			},
@@ -443,12 +443,13 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 	// TODO: do this as part of building the package to be able to link the
 	// bitcode files together.
 	for _, pkg := range lprogram.Sorted() {
+		pkg := pkg
 		for _, filename := range pkg.CFiles {
 			abspath := filepath.Join(pkg.Dir, filename)
 			job := &compileJob{
 				description: "compile CGo file " + abspath,
 				run: func(job *compileJob) error {
-					result, err := compileAndCacheCFile(abspath, dir, config)
+					result, err := compileAndCacheCFile(abspath, dir, pkg.CFlags, config)
 					job.result = result
 					return err
 				},
--- a/builder/cc.go
+++ b/builder/cc.go
@@ -57,7 +57,7 @@ import (
 //   depfile but without invalidating its name. For this reason, the depfile is
 //   written on each new compilation (even when it seems unnecessary). However, it
 //   could in rare cases lead to a stale file fetched from the cache.
-func compileAndCacheCFile(abspath, tmpdir string, config *compileopts.Config) (string, error) {
+func compileAndCacheCFile(abspath, tmpdir string, cflags []string, config *compileopts.Config) (string, error) {
 	// Hash input file.
 	fileHash, err := hashFile(abspath)
 	if err != nil {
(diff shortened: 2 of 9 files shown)


oflebbe:
Eerrr, no this doesn't work for me. 

The code in cgo.go sets a variable cflags (lower case) and the build/builder.go code uses loader.Package.CFlags a variable in a different package as far as I can tell.

There is more what I don't get: in builder/build.go the code is compiled with "clang" without any target specifc flags. Even if we had CFLAGS appended to the command line, this will not link.

I had no luck to see any flags from #cgo: supplied to the compiler.

aykevl:
@oflebbe did you test the code? There is a testcase in testdata/cgo which tests whether the feature works (and it does).

Maybe you are confused by the first two commits which are unrelated? Only the third (and last) commit is relevant to this feature.

aykevl:
> There is more what I don't get: in builder/build.go the code is compiled with "clang" without any target specifc flags.

What do you mean? Target specific flags are always used, otherwise an object file for the wrong architecture would be created.

oflebbe:
I used the code mentioned in the original ticket https://github.com/oflebbe/testesp32.git with the flags from to README.
Used the branch cgo-cflags-files to compile tinygo from sources
Additionally I added a debug statement 
  fmt.Printf("Holla: %s, %v\n", config.Target.Compiler, cflags)  in Front of the runCCompiler statement to see the flags and invocation.

What I see is
 /usr/local/tinygo/bin/tinygo build -target pca10040 -o dummy.hex example/main.go
...
Holla: clang, [-c -o /var/folders/3p/my6cgbr953q73wcb34_9d5gw0000gp/T/tinygo272042384/pkg3.0-demo.c.o /Users/olaf/Devel/testesp32/demo.c]
...

Not sure if I missed something along the way

aykevl:
What if you try this?

    /usr/local/tinygo/bin/tinygo build -target pca10040 -o dummy.hex ./example

(My laptop is doing a Windows upgrade so not available right now...)

oflebbe:
Thanks for your suggestion, never thought about using it that way. 
However
```
Holla: clang, [-c -o /var/folders/3p/my6cgbr953q73wcb34_9d5gw0000gp/T/tinygo653743909/pkg3.0-demo.c.o /Users/olaf/Devel/testesp32/demo.c]
/Users/olaf/Devel/testesp32/demo.c:1:10: fatal error: 'foo.h' file not found
#include <foo.h>
         ^~~~~~~
1 error generated.
failed to run tool: clang
error: failed to build /Users/olaf/Devel/testesp32/demo.c: exit status 1

```
Using a Mac, btw, in case it makes it matters.

oflebbe:
BTW: I never got the build option ` -x	Print commands`  to work . It would have been very helpful here.

aykevl:
I will look into this once I can use my laptop again (Windows upgrade).

oflebbe:
I appreciate your efforts, tinygo is great. Take your time.

aykevl:
Thank you!

> BTW: I never got the build option ` -x Print commands` to work . It would have been very helpful here.

`-x` will be fixed in #1707.

aykevl:
I found the issue and it was... a race condition. I'll submit a fix soon.

aykevl:
It should now be fixed. Thank you for testing!

The fix was very simple:

```diff
  	for i, pkg := range lprogram.Sorted() {
+		pkg := pkg
 		for j, filename := range pkg.CFiles {
```

Result:

```
$ tinygo build -target pca10040 -o dummy.hex -size=short example/main.go
   4824       8    2208 |    4832    2216
```

oflebbe:
Thanks, works for me now! I cannot reproduce the failing smoketest test-llvm11-go115 on my mac. Maybe a retry should help?

aykevl:
I think I found why the smoke test failed and fixed it in #1724.

aykevl:
Rebased, this is now ready for review.

deadprogram:
@aykevl a merge conflict resolution is needed again, please.

aykevl:
I've rebased this PR.

deadprogram:
Seems like something not working as expected here?

https://app.circleci.com/pipelines/github/tinygo-org/tinygo/4670/workflows/6315ed1a-ffa4-4a71-984d-bda3926051e2/jobs/26422

aykevl:
Yeah, it's signal number 9 so I suspect an OOM in the container. I've restarted the failing build.

deadprogram:
Thanks @aykevl now merging.

PR #1709, opened 2021-03-12
Fix fe310 SPI read

tctromp:
Reading the RXDATA register removes the data from the FIFO. Attempting to read the data register immediately after checking for data would return the next item in the FIFO.

 Saving the data while checking resolves this problem.

The `sifive.QSPI_RXDATA_DATA_Msk` mask is also unnecessary as the data will be cast to a byte.


deadprogram:
Thank you very much for the fix @tctromp it is appreciated! Now merging.

PR #1711, opened 2021-03-13
stm32l0: use unified UART logic

kenbell:
Brings stm32L0 into line with other STM32 boards


deadprogram:
Thanks @kenbell looks good! :scissors:

PR #1712, opened 2021-03-14
Fix data shift/mask in func newUSBSetup

ardnew:
My Go linter pointed me to a suspicious piece of code in `src/machine/usb.go` where a shift is performed that is wider than its operand. The exact error message was:

> data[5] (8 bits) too small for shift of 8

I'm not sure how to exercise this bug, but, left as-is, the logic doesn't make sense. This PR widens the operand to 16 bits before applying the left-shift, which seems to be what was originally intended.


sago35:
I think this is correct.

sago35 (approved):

deadprogram:
Makes sense to me. Now merging, thanks @ardnew and @sago35

PR #1714, opened 2021-03-14
Add board support for Adafruit Grand Central M4 (SAMD51)

ardnew:
## Status

This PR adds basic support for the [Adafruit Grand Central M4 (SAMD51)](https://learn.adafruit.com/adafruit-grand-central/overview), with the following peripherals implemented using the existing support in machine `atsamd51`:

- `GPIO` **x70** (!)
- `ADC`, `DAC`
- `PWM`
- `UART0` (USB CDC virtual serial port)
- `UART1`, `UART2`, `UART3`, `UART4`
- `I2C0`, `I2C1`
- `SPI0`, `SPI1`

## Limitations

Note that there appears to be an issue with _some_ ADC channels, which are not mapped to the same pins defined in `machine_atsamd51.go`. Also, only 1 of the board's 2 DACs are available, and PWM is only supported on pins configured with a `TCC` timer/counter (not on pins with `TC`).

These will all require changes that may affect existing `atsamd51` targets to correct.


src/machine/machine_atsamd51p20.go
```diff
@@ -0,0 +1,70 @@
...
+	sam.GCLK.PCHCTRL[sam.PCHCTRL_GCLK_TCC0].Set((sam.GCLK_PCHCTRL_GEN_GCLK0 << sam.GCLK_PCHCTRL_GEN_Pos) |
+		sam.GCLK_PCHCTRL_CHEN)
+	sam.GCLK.PCHCTRL[sam.PCHCTRL_GCLK_TCC2].Set((sam.GCLK_PCHCTRL_GEN_GCLK0 << sam.GCLK_PCHCTRL_GEN_Pos) |
+		sam.GCLK_PCHCTRL_CHEN)
+	sam.GCLK.PCHCTRL[sam.PCHCTRL_GCLK_TCC4].Set((sam.GCLK_PCHCTRL_GEN_GCLK0 << sam.GCLK_PCHCTRL_GEN_Pos) |
+		sam.GCLK_PCHCTRL_CHEN)
+}
+
+// getTimer returns the timer to be used for PWM on this pin
+func (pwm PWM) getTimer() *sam.TCC_Type {
```
sago35, line 31: The same code should be used for ATSAMD51P20 and ATSAMD51P19 (wioterminal), except that the ROM/RAM is different.
So the only difference in machine_atsamd51pXX.go will be HSRAM_SIZE.
I think `PWM.getTime()` should include both.
I think it would be better to keep the definitions that exist in `machine_atsamd51p19.go` .

https://github.com/tinygo-org/tinygo/blob/release/src/machine/machine_atsamd51p19.go#L30

targets/grandcentral-m4.json
```diff
@@ -0,0 +1,11 @@
+{
+    "inherits": ["atsamd51p20a"],
+    "build-tags": ["grandcentral_m4"],
+    "flash-1200-bps-reset": "true",
+    "flash-method": "msd",
+    "msd-volume-name": "GCM4BOOT",
+    "msd-firmware-name": "firmware.uf2",
+    "openocd-transport": "swd",
+    "openocd-interface": "jlink",
+    "openocd-target": "atsame5x"
```
sago35: openocd-target is not needed since it is defined in atsamd51p20a.json

sago35:
* Board
    * Name
        * Adafruit Grand Central M4 Express
    * Board developer / designer / manufacturer
        * Adafruit
    * schematic
        * https://learn.adafruit.com/assets/69175
    * pinout and stuff
        * 
    * URL
        * https://www.adafruit.com/product/4064
    * User manual for the board (if it exists)
        * https://learn.adafruit.com/adafruit-grand-central/overview
* Chip
    * chip name
        * ATSAMD51P20
    * Clock / ROM / RAM size
        * 120 MHz / 1024 KB / 256 KB
    * datasheet URL of the chip
        * https://www.microchip.com/wwwproducts/en/ATSAMD51P20A
* URL of the source, such as arduino (if it exists)
    * https://github.com/adafruit/ArduinoCore-samd/tree/master/variants/grand_central_m4
    * https://github.com/adafruit/ArduinoCore-samd/blob/master/boards.txt#L633-L696

src/machine/board_grandcentral-m4.go
```diff
@@ -0,0 +1,244 @@
...
+	SDCARD_SDO_PIN = SD0_SDO_PIN //
+	SDCARD_SDI_PIN = SD0_SDI_PIN //
+	SDCARD_CS_PIN  = SD0_CS_PIN  //
+	SDCARD_DET_PIN = SD0_DET_PIN //
+)
+
+// Other peripheral constants
+const (
+	RESET_MAGIC_VALUE      = 0xF01669EF // Used to reset into bootloader
+	EXTERNAL_FLASH_DEVICES = "GD25Q16C" // QSPI
```
sago35, line 225: `GD25Q64` seems to be correct.
https://cdn-learn.adafruit.com/assets/assets/000/069/175/original/adafruit_products_Grand_Central_Sch1.png?1547234817

sago35: `QD25Q64` seems to be correct
However, if you never refer to this definition, you may want to remove it.

https://cdn-learn.adafruit.com/assets/assets/000/069/175/original/adafruit_products_Grand_Central_Sch1.png?1547234817

targets/grandcentral-m4.json
```diff
@@ -0,0 +1,11 @@
+{
+    "inherits": ["atsamd51p20a"],
+    "build-tags": ["grandcentral_m4"],
+    "flash-1200-bps-reset": "true",
+    "flash-method": "msd",
+    "msd-volume-name": "GCM4BOOT",
+    "msd-firmware-name": "firmware.uf2",
+    "openocd-transport": "swd",
```
sago35: `openocd-transport` is not needed since it is defined in atsamd51p20a.json

src/machine/machine_atsamd51p20.go
ardnew: `getTimer` appears different between `atsamd15p19.go` (Wio) and `atsamd51p20.go` (GC). For example, `PA16` is defined on the P19, but it is _not_ defined on the P20; the P20 would return a nil timer for `PA16`.

Note that I've based these mappings for the P20 on the Arduino core ([see here](https://github.com/adafruit/ArduinoCore-samd/blob/aebf3cb686daa2bb8ee8e85febf6ed656241d933/variants/grand_central_m4/variant.cpp#L29)), so it is possible their pin definitions are not the only permissable configuration; i.e., the P19 implementation of `getTimer` might be valid for the P20 as well (I have no idea). I just wanted to point out they are currently not equivalent.

If this is the case, though, I think the J20 is a closer match to the P20, since it also has an equal value for `HSRAM_SIZE`, but otherwise has an identical implementation of `getTimer` to the P19.

### Different PR 

However, I'm not sure this effort should be restricted to these two variants, and thus should not be part of this PR. For example, all of the following source files are identical (except for `HSRAM_SIZE`):

- `machine_atsamd51j19.go`
- `machine_atsamd51j20.go`
- `machine_atsamd51p19.go`
- `machine_atsamd51p20.go` (if its pins can use a different configuration than implemented in this PR)

I propose that all of these files be combined into a single source file.

Furthermore, since `HSRAM_SIZE` can be derived from the linker script for each of this chips, I would recommend we not redefine here in the source file, and let the linker script define it. Then this common source file can just use a `//go:extern HSRAM_SIZE` declaration to get the appropriate value.

targets/grandcentral-m4.json
ardnew: Thanks, will remove

ardnew:
> * Chip
>   
>   * chip name
>     
>     * ~~ATSAMD51J19~~

Should be the **ATSAMD51P20**

ardnew:
> * Chip
>   
>   * chip name
>     
>     * ~~ATSAMD51J19~~

@sago35 Should be the **ATSAMD51P20**

src/machine/board_grandcentral-m4.go
ardnew: Indeed, forgot to research if this is needed somewhere, will remove

targets/grandcentral-m4.json
ardnew: Thanks, will remove

src/machine/machine_atsamd51p20.go
sago35: Yes, let's try to handle it with different PR.

It will probably need to be set for each Pin Count in GJNP. However, if we define a dummy pin (PCxx / PDxx) that does not exist in the G series, we may be able to use the same source. However, in that case, it will be difficult to check for errors.

sago35: So, I think this part is fine as it is now.
Thank you for your comments.

sago35 (approved):

deadprogram:
Thank you very much for another board contributed @ardnew and to @sago35 for such a detailed review. You two are fantastic!

Now squashing/merging.

PR #1715, opened 2021-03-14
all: prototype of RAM persistence across reset

kenbell:
See issue #1716 which proposes adding persistence regions to TinyGo


src/runtime/persistence/persistence.go
```diff
@@ -0,0 +1,70 @@
...
+	ReadAt(dest []byte, offset int64) (n int, err error)
+
+	// WriterAt reads from the persistence area, it is signature
+	// compatible with Go's `io.WriterAt` interface
+	WriteAt(src []byte, offset int64) (n int, err error)
+
+	// SubAllocation creates a new persistence object that
+	// represents a sub-allocation of the main persistence
+	// area
+	SubAllocation(offset int, len int) (p Region, err error)
```
ardnew: What is the imagined use-case of this? Seems like it could be left out (for now) to simplify the interface until implementations have been fleshed out and tested. 

It seems like more of a convenience that isn't truly essential for this initial eval.

compiler/ramallocation.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"go/types"
+
+	"golang.org/x/tools/go/ssa"
+	"tinygo.org/x/go-llvm"
+)
+
+func (b *builder) createRAMAllocation(instr *ssa.CallCommon) error {
+	// Get the size, which must be a compile-time constant.
+	size, ok := instr.Args[0].(*ssa.Const)
```
ardnew: It could be really useful to be able to pass in a type and determine the size to allocate accordingly. 

This would let users define persistent variables; when their type definition changes, the reserved region grows/shrinks accordingly, without having to redo any arithmetic and update the `NewRAM` call separately (or risk forgetting to do so).

Just an idea, which may be a bit more elaborate than necessary for now.

Example use case:

```go
type Config struct {id int}

func main() {

	p := persistence.NewRAM(Config)

	// would be cool to either re-use the `make` builtin somehow, or create an equivalent:
	// p := persist(Config)
	// q := persist([]rune, 1024)
}
```

targets/arm.ld
```diff
@@ -69,3 +80,7 @@ _heap_start = _ebss;
 _heap_end = ORIGIN(RAM) + LENGTH(RAM);
 _globals_start = _sdata;
 _globals_end = _ebss;
+
+/* For the RAM persistence module. */
+_persist_start = _spersist;
+_persist_end = _epersist;
```
ardnew: Can you not just define these above in place of `_spersist` and `_epersist`? There doesn't appear to be anything else using these symbols, so I suggest getting rid of them.

targets/arm.ld
```diff
@@ -31,6 +31,17 @@ SECTIONS
         _stack_top = .;
     } >RAM
 
+    /* Persistent RAM that survives reset. */
+    .persist (NOLOAD) :
+    {
+        . = ALIGN(4);
+        _spersist = .;
+        *(.persist)
+        *(.persist.*)
```
ardnew, line 41: I've never understood the convention with these includes, are they both necessary? Wouldn't `*(.persist*)`, or, in this case, simply `*(.persist)` be sufficient?

ardnew (approved):
This is really neat! Only a few questions/comments added

targets/arm.ld
aykevl: It's simply a glob. So `.persist` matches exactly the string `.persist` while `.persist.*` matches anything that starts with `.persist.`. This emulates somewhat the similar conventions around `-ffunction-sections` and `-fdata-sections` but may not be necessary in this case.

compiler/ramallocation.go
aykevl: That may sound like a good idea, but has some issues in practice. For example: you can't put pointers in this area (and thus no slices, strings, etc.) because they live on the heap which is reinitialized on reset.

This issue details why this is a bad idea for a very similar use case on server hardware: https://github.com/golang/go/issues/43810

compiler/ramallocation.go
```diff
@@ -0,0 +1,44 @@
...
+	// We create a buffer 'constant' in LLVM terms as an array of desired size.  Since
+	// the Go runtime code can't know this type (array dimensions), we also have a pointer
+	// that points at the allocated buffer.  In order to prevent the symbols from being
+	// optimized out, the pointer `_persist_ptr` must be referenced from the runtime
+	// code.
+
+	// _persist$buf is the allocated space.
+	bufType := llvm.ArrayType(llvm.Int8Type(), int(size.Int64()))
+	buf := llvm.AddGlobal(b.mod, bufType, "_persist$buf")
+	buf.SetLinkage(llvm.ExternalLinkage)
```
aykevl: Why external linkage? It should be internal linkage to help the optimizer.

src/examples/persistence/main.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"time"
+	"unsafe"
+
+	"runtime/persistence"
+)
+
+func main() {
+	p := persistence.NewRAM(uint32(unsafe.Sizeof(byte(0)) * 100))
```
aykevl: You can assume `unsafe.Sizeof(byte)` is always 1. Doing something different would break the Go language specification: https://golang.org/ref/spec#Size_and_alignment_guarantees

I have absolutely no plans of supporting architectures that do not support byte addressing. There is hardware that doesn't support byte addressing, but it's all very special purpose (DSPs etc.). And even if TinyGo would support this hardware, it would still follow the Go specification (whatever that might look like in practice).

aykevl (commented):
While I think this is an interesting idea, I have a few concerns with this implementation.

1. I do not think we should be mixing RAM and flash here. While reading RAM and flash is usually similar (just read as any normal pointer), writing most definitely is not. For example: in flash, usually a whole page (1024 bytes, 4069 bytes, or even bigger) needs to be erased at a time.
2. I don't really see the purpose of the special Region type? Why not return a byte slice from `NewRAM`?
3. Why shouldn't libraries allocate persistent RAM?
4. How can an application detect whether the memory area has been initialized or not? Should it care?
5. What about compatibility with the Go toolchain?

Especially the last point is important to me. I think it is fairly simple to make this compatible with the Go toolchain by doing something similar to the new file embedding support in Go 1.16. I think the API could be simplified to the following:

```go
//go:persistent
var somePersistentMemory [100]byte
```

The semantics could for example be: the memory area is reset to zero with a power-on reset but otherwise left as-is (for example, with a wake up from deep sleep). In the Go toolchain the variable is always zero initialized on startup.
That requires detecting a power-on reset on every supported chip. However, I think this feature is almost useless without a way to know whether some data has been written to memory or not: RAM is not necessarily all zeroes on startup, it's more likely to be random data.

Overall, I'm curious whether you have a use case in mind? You've mentioned a few examples where it could be useful, but few practical examples that currently apply to TinyGo. It really helps to know how it will be used to shape what the API looks like.

src/runtime/persistence/persistence.go
kenbell: I was thinking along the lines of slices in Go - being able to divide the space up, so parts can be handed off to different libraries without risk of bounds overflow.

compiler/ramallocation.go
kenbell: Yeah - i think serialization needs to be explicit.

compiler/ramallocation.go
kenbell: I'll admit - i don't really know what the values do, I was bashing on values until one worked :)

I'll take another look.

src/examples/persistence/main.go
kenbell: Yeah - you're right, that example is excessively complex - should just be 'i need a region of X bytes'

kenbell:
> 1. I do not think we should be mixing RAM and flash here.

Reading your thoughts, I think you're right..   There's enough different between them that trying to pretend they're the same will add a lot of complexity without much (if any) clear benefit.


> 2. I don't really see the purpose of the special Region type? Why not return a byte slice from NewRAM?

I wasn't sure how that would interact with the GC.  It would be a lot cleaner if that was the interface.  I'll investigate.


> 3. Why shouldn't libraries allocate persistent RAM?

There's an ordering constraint I'm not sure how to fix.  I don't think the compiler / linker chain guarantees the order symbols are packed into sections - so on a rebuild the layout could vary.  For RAM, this probably isn't a concern (good chance the RAM is wiped anyway on a re-flash) - but for Flash, would be good to make the ordering explicit, which is why I left it to the app to allocate the data to libraries.  I'm hoping we can have it so that on a reasonable sub-set of devices, flashing a new app will leave the config intact.

This may be another good reason to separate RAM and Flash, since letting libraries allocate persistent RAM on the understanding it will not be guaranteed (or even likely) to be preserved across re-builds / re-flashing seems a good approach.


> 4. How can an application detect whether the memory area has been initialized or not? Should it care?

I'm thinking magic values guarding the memory region, or some kind of CRC - would be use-case specific.  If the guard values are wrong, or the CRC is invalid then assume uninitialized and do the initialization.   STM32 does the same for RTC - it has a few bytes that are explicitly there for detecting if the RTC has been initialized - they document to set to a magic value of the developer's choosing after initializing the RTC.


> 5. What about compatibility with the Go toolchain?

I really like the idea of using `[]byte` and `//go:persistent` - feels a lot more elegant.  I'll take a look at trying to do that.

targets/arm.ld
kenbell: I was just aiming to follow the pattern in the file - i'm more than happy to just use _spersist, _epersist.  It would work fine.

kenbell:
> Overall, I'm curious whether you have a use case in mind? You've mentioned a few examples where it could be useful, but few practical examples that currently apply to TinyGo. It really helps to know how it will be used to shape what the API looks like.

I've put a sketch of an prototypical app in the issue / proposal #1716, here (https://github.com/tinygo-org/tinygo/issues/1716#issuecomment-798997491)

src/runtime/persistence/ram.go
```diff
@@ -0,0 +1,83 @@
...
+	p := uintptr(r.start) + uintptr(offset)
+	for i := 0; i < len(dest); i++ {
+		dest[i] = *(*byte)(unsafe.Pointer(p))
+		p++
+	}
+
+	return len(dest), nil
+}
+
+// WriterAt reads from the persistence area, it is signature
```
ardnew: Should be `WriteAt` here

compiler/ramallocation.go
ardnew: How about the `type` argument in `NewRAM(type)` is required to implement a byte serialization interface? Something like the following:

```go
type RAMSerializer interface {
	Offset() int // zero unless its a subregion
	Size() int
	Serialize() ([]byte, error)
	Deserialize([]byte) error
}
```

This could facilitate higher-level `Get`/`Set` operations in the persistence API, making it friendlier to interact with; instead of _always_ having to juggle byte slices, the `ReadAt`/`WriteAt` methods could be invoked automatically internally via `RAMSerializer` interface.

Perhaps this functionality could just as well be implemented in user code... But if the pattern is useful and repeated often enough (which we cannot possibly know yet), why not eventually include it with the core API?

deadprogram:
Just wondering if this is actually now better addressed via the recent flash interface changes?

aykevl:
No, I don't think the flash API is appropriate for this. Flash is very different from RAM.

I do quite like the design here: https://github.com/tinygo-org/tinygo/issues/1716#issuecomment-799125817
I don't think the magic seed in that example is really necessary though: practically all MCUs provide a way to read the reset reason and therefore can know whether the reset was due to a power-on reset (RAM is gone) or a software reset (RAM is still available). We just need to make this available in the machine package.

@kenbell any plans to continue working on this? Perhaps start by rebasing on the dev branch.

kenbell:
Yeah - I'd kinda forgotten about this.  I'll take a look in the next few days.

PR #1717, opened 2021-03-14
Support C++ files for cgo as well

oflebbe:
Hi, was looking into repurposing the Arduino Wifi.h for esp32 tinygo. But the rabbit hole is too deep, at least for now.

However, for this experiment I needed to support C++ cgo files (since Arduino is C++), so I implemented C++ support in monkey see - monkey do mode.  I surely had some invalid shortcuts.

Consider it for inclusion, if you see it useful. The initial usecase fot it is now void, unfortunately. But it might be helpful to add C++ libs in future?


aykevl:
Interesting! Thank you for working on this.

At the moment it's not really in a mergeable state as the third commit is partially #1708 and partially your own changes. But C++ support would certainly be useful. To make this mergeable, it will also need a test and will need to pass CI tests. You can add a test to ./testdata/cgo.

oflebbe:
Recreated patch, added missing files and fixed tests .

oflebbe:
Built on top of  cgo-cflags-files branch.

testdata/cgo/main_cpp.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+
+extern "C" int fortythree() {
+	FortyThree ft;
+
+	return ft.get();
+}
+
+extern "C" int other() {
+   return OTHER_CONSTANT;
```
aykevl, line 22: It looks like you're not checking the value here in main.go.

main.go
```diff
@@ -888,6 +889,10 @@ func main() {
 		options.CFlags = strings.Split(*cFlags, " ")
 	}
 
+	if *cxxFlags != "" {
+		options.CXXFlags = strings.Split(*cxxFlags, " ")
```
aykevl, line 893: I plan on removing `options.CFlags` (because it's not necessary anymore with better CGo support) so I think it's better to not add it here for this PR. It's a bit inconsistent but not a problem in my opinion.

aykevl (commented):
Thank you, this looks pretty good! There is a bit of duplication unfortunately, but maybe it's not worth fixing that (removing the duplication may introduce additional complexity).

One thing I wonder about: wouldn't the CXXFLAGS always be CFLAGS plus some extra CFLAGS? I don't know how the Go compiler deals with this, but there is of course a fair bit of overlap between them.

testdata/cgo/main_cpp.cpp
oflebbe: Right! Improving this.

main.go
oflebbe: Maybe problematic: 

In the good old days with larger C/C++ codebases, CPPFLAGS was used for flags common to C and C++ , using CFLAGS for only C specific Flags and CXXFLAGS. However I wanted to have a shortcut implementing C++ support, as I wrote previously. So I hesitated from implementing CPPFLAGS and changing all target json files. 

Not sure how to proceed. My gut feeling is to have code duplicated now and rework later if this really is limiting tinygo, though not exactly clean code.

PR #1718, opened 2021-03-15
Group together STM32 smoke tests.

QuLogic:
And then allow them to be disabled with one option.

Unfortunately, I can't ship the STM32 svd files because of their license, so I need an easy way to skip these smoke tests. This option is similar to the `AVR` and `XTENSA` ones.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,10 @@ fmt-check:
 	@unformatted=$$(gofmt -l $(FMT_PATHS)); [ -z "$$unformatted" ] && exit 0; echo "Unformatted:"; for fn in $$unformatted; do echo "  $$fn"; done; exit 1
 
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Can you please resolve the current merge conflict @QuLogic thanks.

deadprogram:
Thanks @QuLogic now merging.

aykevl:
Hmm, that's unfortunate. But if you say so...

I hope ST will eventually add a license to these files, it's weird they don't. (At the same time I think it's very unlikely they will sue anybody for using the SVD files for their own chips).

QuLogic:
Alas, there is a license, but it's not open source friendly: https://github.com/posborne/cmsis-svd/blob/master/data/STMicro/License.html

I did ask on their forums, and they did say they'd look into it, but who knows. Maybe with more ARM vendors using ASL2.0/BSD, they might start releasing it that way too, but it may only be for new files.

Issue #1653, opened 2021-02-22
LLVM ERROR: Cannot select: intrinsic %llvm.coro.size


ALTree:
```
$ tinygo version
tinygo version 0.16.0 linux/amd64 (using go version go1.15.8 and LLVM version 10.0.1)
```
```
package main

func main() {
	var ch chan string
	var m map[byte]string
	_ = m[0]
	_ = <-ch
}
```
```
$ tinygo build -o out crash.go 
LLVM ERROR: Cannot select: intrinsic %llvm.coro.size
```

deadprogram:
Is that actually valid Go code?

```
fatal error: all goroutines are asleep - deadlock!

goroutine 1 [chan receive (nil chan)]:
main.main()
	/tmp/sandbox401551102/prog.go:7 +0x65

Program exited: status 2.
```

aykevl:
It is valid in the sense that it is well defined: a program like this is defined to exit with a panic. This is certainly a compiler bug, either in TinyGo or in LLVM.

ALTree:
@deadprogram Yes, it's valid, in the sense that it conforms to the [Go Specification](https://golang.org/ref/spec). 

As @aykevl pointed out, in general a compiler shouldn't crash when compiling valid -or invalid- Go code. Valid code should generate a binary, which then may or may not panic at runtime, like mine does; invalid Go code should print a compilation error. In this case LLVM is crashing while processing a valid Go program, so this can be classified as a compiler bug.

niaow:
This should be addressed now that coroutines are no longer used (and will be removed by #2174 )

ALTree:
Confirmed fixed on 0.25

PR #1720, opened 2021-03-15
Test with LLVM 12.

QuLogic:
Just a check to see if CI will work; I already tested locally and found it crashed on wasm stuff, so there's still some work to do.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -377,6 +377,12 @@ jobs:
     steps:
       - test-linux:
           llvm: "11"
+  test-llvm12-go116:
+    docker:
+      - image: circleci/golang:1.16-buster
+    steps:
+      - test-linux:
+          llvm: "12"
   assert-test-linux:
     docker:
       - image: circleci/golang:1.17-stretch
@@ -400,6 +406,7 @@ workflows:
     jobs:
       - test-llvm10-go115
       - test-llvm11-go116
+      - test-llvm12-go116
       - build-linux
       - build-macos
       - assert-test-linux
--- a/cgo/libclang_config.go
+++ b/cgo/libclang_config.go
@@ -1,5 +1,5 @@
 // +build !byollvm
-// +build !llvm10
+// +build !llvm10,!llvm12
 
 package cgo
 
--- a/cgo/libclang_config_llvm12.go
+++ b/cgo/libclang_config_llvm12.go
@@ -1,0 +1,14 @@
+// +build !byollvm
+// +build llvm12
+
+package cgo
+
+/*
+#cgo linux   CFLAGS:  -I/usr/lib/llvm-12/include
+#cgo darwin  CFLAGS:  -I/usr/local/opt/llvm@12/include
+#cgo freebsd CFLAGS:  -I/usr/local/llvm12/include
+#cgo linux   LDFLAGS: -L/usr/lib/llvm-12/lib -lclang
+#cgo darwin  LDFLAGS: -L/usr/local/opt/llvm@12/lib -lclang -lffi
+#cgo freebsd LDFLAGS: -L/usr/local/llvm12/lib -lclang
+*/
+import "C"
--- a/go.mod
+++ b/go.mod
@@ -12,5 +12,5 @@ require (
 	go.bug.st/serial v1.1.2
 	golang.org/x/sys v0.0.0-20210510120138-977fb7262007
 	golang.org/x/tools v0.1.6-0.20210813165731-45389f592fe9
-	tinygo.org/x/go-llvm v0.0.0-20210325115028-e7b85195e81c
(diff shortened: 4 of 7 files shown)


aykevl:
```
The following additional packages will be installed:
  adwaita-icon-theme at-spi2-core binfmt-support binutils-aarch64-linux-gnu
```

`adwaita-icon-theme`?? Something is odd here. I suspect some packages are pulled in that shouldn't be pulled in (but that may be a problem with the container or Debian, not this PR).

QuLogic:
There are two failures in the transform tests:
* In `maps.ll`, the main difference seems to be that `testUnused` contains two extra `hashmap` lines:
  ```
  %hashmap.value = alloca i32, align 4
  %hashmap.value.bitcast = bitcast i32* %hashmap.value to i8*
   ```
* In `coroutines.ll`, it appears that one of the `arguments` lines has added `nofree nosync`; differences on these lines appear to be ignored for LLVM<10 by `filterIrrelevantIRLines`, but it's been changed again for 12.

There's also the weird issue of not finding `-lclang`, and final abort due to:
```
LLVM ERROR: Cannot select: intrinsic %llvm.coro.size
```
Not sure if that's related to #1653 or some other bug.

QuLogic:
Running `TestHost` localls, I can find that `testdata/channel.go` crashes after the `deadlocking` printout, which seems consistent with the WASI results, which seem to have trapped:
```
    1: wasm trap: unreachable
       wasm backtrace:
           0: 0x4258 - <unknown>!main.selectDeadlock
           1: 0x34f7 - <unknown>!main.main.resume
           2: 0x2517 - <unknown>!_start
       note: run with `WASMTIME_BACKTRACE_DETAILS=1` environment variable to display more information
```

I believe the `%llvm.coro.size` abort comes from the `testdata/json.go` file.

aykevl:
I've made an issue over at the Xtensa fork so they will hopefully update their fork (which we use for ESP32 support): https://github.com/espressif/llvm-project/issues/46

aykevl:
It appears there were some changes to coroutines in LLVM 12. The following code leads to a crash:

```go
package main

func main() {
	go selectDeadlock()
}

func selectDeadlock() {
	println("deadlocking")
	select {}
	println("unreachable")
}
```

QuLogic:
Any ideas on finishing this? We're almost at LLVM 13 time now.

aykevl:
I've made the `llvm12` branch where I continue working on support for LLVM 12.

aykevl:
Progress:
After some digging, I found that the `llvm.coro.size` assert seems to be caused by added `noreturn` attributes. They are correct before the coroutine transformation pass, but not anymore afterwards. Removing the attribute gets testdata/channel.go to work again but doesn't fix testdata/json.go so it's not a complete fix. Also, simply removing the attribute seems incorrect.

```diff
diff --git a/transform/coroutines.go b/transform/coroutines.go
index 2aeff43c..1fc39635 100644
--- a/transform/coroutines.go
+++ b/transform/coroutines.go
@@ -1029,6 +1029,7 @@ func (c *coroutineLoweringPass) fixAnnotations() {
                                f.RemoveEnumAttributeAtIndex(i, kindID)
                        }
                }
+               f.RemoveEnumFunctionAttribute(llvm.AttributeKindID("noreturn"))
        }
 }
 
```

aykevl:
I've been further experimenting in the llvm12 branch: https://github.com/tinygo-org/tinygo/compare/dev...llvm12
Still looking into why goroutines are crashing.

aykevl:
Here is a small reproducer:

```go
package main

import "time"

func main() {
	go selectDeadlock()
	time.Sleep(1)
}

func selectDeadlock() {
	println("deadlocking")
	select {}
	println("unreachable")
}
```

aykevl:
So the issue is that some functions don't return, the `selectDeadlock` above for example doesn't return. In coroutine lowering, goroutines are started by simply calling them. When those functions don't return (even if they simply exit without panicking), this will lead to unreachable instructions in the parent goroutine.

Here is another reproducer:

```go
package main

import "time"

func main() {
	go doPanic()
	time.Sleep(1)
}

func doPanic() {
	println("hello")
	panic("abort! abort!")
}
```

aykevl:
An even more innocent example, no `panic` in sight!

```go
package main

import "time"

func main() {
	go goroutine()
	time.Sleep(1)
	println("main hello")
}

func goroutine() {
	for {
		println("go hello")
		time.Sleep(time.Second)
	}
}
```

It triggers an error because `goroutine` is called from main but doesn't actually return.
I'm now convinced this is a bug in how we do goroutines. It has always been buggy, but LLVM didn't do the optimizations yet that would catch the bug. But in LLVM 12, the optimizer has become smart enough to catch the (erroneous) IR that we were emitting.

dgryski:
This can probably be closed now.

QuLogic:
Yes indeed.

PR #1721, opened 2021-03-15
build: make msd flash faster

sago35:
This is mainly for cases where msd flash is used instead of openocd.
The basic idea is to run the `touchSerialPortAt1200bps()` in parallel with the build.

It makes the environment a little more comfortable for tinygopher, which doesn't have a debugger.
If the build takes more than 3 seconds, it will be at least 3 seconds faster.
If the build takes less than 3 seconds, there is not much improvement.

* Core i5-5200 + ubuntu 20.04
    * before : 8.6 sec
    * after : 3.5 sec
* Ryzen 7 3700X + Windows10
    * before : 5.0 sec
    * after : 4.5 sec

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -44,7 +44,7 @@ type BuildResult struct {
 //
 // The error value may be of type *MultiError. Callers will likely want to check
 // for this case and print such errors individually.
-func Build(pkgName, outpath string, config *compileopts.Config, action func(BuildResult) error) error {
+func Build(pkgName, outpath string, config *compileopts.Config, preAction func() error, action func(BuildResult) error) error {
 	compilerConfig := &compiler.Config{
 		Triple:          config.Triple(),
 		CPU:             config.CPU(),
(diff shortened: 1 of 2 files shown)


builder/build.go
```diff
@@ -87,6 +87,14 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 	// Makefile target.
 	var jobs []*compileJob
 
+	if preAction != nil {
+		// Add job to preAction.
```
sago35, line 91: I couldn't think of any good comments here.

main.go
```diff
@@ -250,12 +250,14 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 
 			err := touchSerialPortAt1200bps(port)
 			if err != nil {
-				return &commandError{"failed to reset port", result.Binary, err}
+				return &commandError{"failed to reset port", "", err}
```
sago35, line 253: At the time of processing this part, result.Binary is undefined, so it is assumed to be empty.

deadprogram:
@sago35 is this something that you are still working on/wanting? Or should we close it due to age?

sago35:
I’ll close it for now due to its age. @deadprogram

PR #1724, opened 2021-03-17
compiler: do not use llvm.GlobalContext()

aykevl:
This is a leftover from a long time ago, when everything was still in
the global context. The fact that this uses the global context is most
certainly a bug.

I have seen occasional crashes in the build-packages-indepedently branch
(and PRs based on it) which I suspect are caused by this bug. I think
this is a long-dormant bug that only surfaced when doing the compilation
steps in parallel.


sago35:
I'm not sure what the change is, but I think it's the correct change.
This change may resolve the following PRs.

#1315

sago35 (approved):

aykevl:
>  This change may resolve the following PRs.

That is definitely possible.

deadprogram:
Merging, thanks @aykevl and @sago35

Issue #1415, opened 2020-09-26
Undefined compiler runtime symbols on AVR


rammanokar:
Getting below error when running bmp180 example code. 

**Program Code:** 
``` 
https://raw.githubusercontent.com/tinygo-org/drivers/release/examples/bmp180/main.go
```

**Command Executed:**
```bash
 tinygo build -o test.hex -target arduino bmp180.go
 ```
**Tinygo Version:** 
``` tinygo version 0.15.0 linux/amd64 (using go version go1.15.2 and LLVM version 10.0.1) ```

**tinygo env**
```
GOOS="linux"
GOARCH="amd64"
GOROOT="/home/***/sdk/go1.15.2"
GOPATH="/home/***/go:/home/ram/tinygo:"
GOCACHE="/home/***/.cache/tinygo"
CGO_ENABLED="1"
TINYGOROOT="/usr/local/lib/tinygo"
```

**Error:**
```
tinygo build -o test.hex -target arduino bmp180.go 
/tmp/tinygo299647481/main.o: In function `runtime.printfloat32':
bmp180.go:(.text.runtime.printfloat32+0x3a): undefined reference to `__extendsfdf2'
/usr/lib/gcc/avr/5.4.0/../../../avr/bin/ld: Dwarf Error: found dwarf version '0', this reader only handles version 2, 3 and 4 information.
bmp180.go:(.text.runtime.printfloat32+0x74): undefined reference to `__adddf3'
/usr/lib/gcc/avr/5.4.0/../../../avr/bin/ld: Dwarf Error: found address size '9', this reader can not handle sizes greater than '8'.
bmp180.go:(.text.runtime.printfloat32+0xc2): undefined reference to `__eqdf2'
/usr/lib/gcc/avr/5.4.0/../../../avr/bin/ld: Dwarf Error: found dwarf version '12', this reader only handles version 2, 3 and 4 information.
bmp180.go:(.text.runtime.printfloat32+0xec): undefined reference to `__unorddf2'
/usr/lib/gcc/avr/5.4.0/../../../avr/bin/ld: Dwarf Error: found dwarf version '0', this reader only handles version 2, 3 and 4 information.
bmp180.go:(.text.runtime.printfloat32+0x140): undefined reference to `__eqdf2'
/usr/lib/gcc/avr/5.4.0/../../../avr/bin/ld: Dwarf Error: found dwarf version '0', this reader only handles version 2, 3 and 4 information.
[...]

Issue #1282, opened 2020-08-07
TinyGo fails to compile a Vecty program


marwan-at-work:
Hi there, 
I have a simple [Vecty](https://github.com/gopherjs/vecty) program that compiles with Go but fails to compile with TinyGo:

```
package main

import (
	"github.com/gopherjs/vecty"
	"github.com/gopherjs/vecty/elem"
)

type comp struct {
	vecty.Core
}

func (c *comp) Render() vecty.ComponentOrHTML {
	return elem.Div(vecty.Text("hello"))
}

func main() {
	vecty.RenderBody(&comp{})
}
```

I was using Go modules and TinyGo v0.14.0 when running the following and got the error:

```
root@05844ebaa1e7:/app# tinygo build -o wasm.wasm -target=wasm .
# github.com/gopherjs/vecty
/go/pkg/mod/github.com/gopherjs/vecty@v0.0.0-20200328200803-52636d1f7aba/dom_wasmjs_gopherjs.go: interp: unknown GEP

traceback:
github.com/gopherjs/vecty/<init>:41:24:
  %15 = call %runtime._interface @"github.com/gopherjs/vecty.wrapObject"([0 x %runtime.funcValue] %12, i64 %13, i64* %14, i8* undef, i8* undef), !dbg !216
```

This is issue is a follow up on https://github.com/tinygo-org/tinygo/issues/93 - see [this comment](https://github.com/tinygo-org/tinygo/issues/93#issuecomment-670445323) for more context. 

It would be great to either solve this on the TinyGo side or the Vecty side. 

CC: @slimsag @deadprogram 

Thanks!

bgould:
FWIW the "unknown GEP" error from the interp package usually means that there is some code in a global initializer that cannot be  executed by the interpreter in the TinyGo compiler that is responsible for computing global variables (and marking the resulting values as constants when appropriate).  This is a documented difference from the standard Go implementation, which always defers the computation of globals and the execution of initializers to when the program executes (see last bullet on this page - https://tinygo.org/compiler-internals/differences-from-go/). The compile-time interpretation of initializers is an important aspect of how the TinyGo compiler is able to emit smaller binaries that can run in resource-constrained environments.

In my experience, a lot of the time the "unknown GEP" error happens when functions called from global initializers return an interface, and similar scenarios usually involving interfaces (if I'm not mistaken, global sentinel error values created with `errors.New(...)` are a common cause of this). 
[...]

Issue #1087, opened 2020-05-08
Trying to implement `reflect.New()`


lei-april:
Hi,

I'm trying to implement the missing function `reflect.New()` in TinyGo, and here's my minimal patch:
```diff
diff --git a/src/reflect/value.go b/src/reflect/value.go
index a1c6937..6e953da 100644
--- a/src/reflect/value.go
+++ b/src/reflect/value.go
@@ -650,8 +650,16 @@ func Zero(typ Type) Value {
 	panic("unimplemented: reflect.Zero()")
 }
 
+//go:linkname alloc runtime.alloc
+func alloc(size uintptr) unsafe.Pointer
+
+func PtrTo(t Type) Type {
+	return (t << 5) + Type((Ptr-19)<<1) + 1
+}
+
 func New(typ Type) Value {
-	panic("unimplemented: reflect.New()")
+	data := alloc(typ.Size())
+	return Value{PtrTo(typ), data, 0}
 }
 
 type funcHeader struct {
```

When testing it with the following sample code:
```go
package main

import "reflect"

type FooInterface interface {
	Hello()
}

type Foo struct {
}

func (f *Foo) Hello() {
	println("Hello")
}

func main() {
	foo := reflect.New(reflect.TypeOf(Foo{})).Interface().(FooInterface)
	foo.Hello()
}
```

I got the error: `panic: runtime error: type assert failed`

However, if I change the sample code a little bit, it then works fine:
```go
func touch(i interface{}) {
}

func main() {
	touch(&Foo{}) // ???
	foo := reflect.New(reflect.TypeOf(Foo{})).Interface().(FooInterface)
	foo.Hello()
}
```

It seems that somehow I need to make the compiler see a conversion from `*Foo` to `interface{}`.

Do you have any insights on this? Thanks.

jaddr2line:
Quick explanation of what is happening:

Type asserts are implemented by generating a switch statement. TinyGo tries to compute a list of all implementing types, and then generates a type switch across them.

The way that TinyGo gets this list of all implementing types is by looking at every interface creation operation. In this case, you are synthesizing a type at runtime, and therefore the compiler does not discover the implementation.
So it looks like the compiler will also need to add cases for:
1. a pointer to a type which has been converted to an interface
2. all fields of types which have been converted to an interface

Other synthesized types cannot implement non-empty interfaces, so we do not need any gigantic changes to how we handle this yet.

deadprogram:
[...]

PR #1727, opened 2021-03-18
Improve reflection

aykevl:
This is a big PR (based on #1612) that improves reflect support significantly:

1. `compiler: merge runtime.typecodeID and runtime.typeInInterface`: this commit makes reflect support more correct, see testdata/reflect.go for a case where it behaved incorrectly before. It increases code size in many cases.
2. `compiler: do not check for impossible type asserts`: reduces the code size in many cases introduced in the previous commit.
3. `interp: handle (reflect.Type).Elem()`: prepare for the last commit. Shaves off a few bytes in many cases.
4. `interp: add support for runtime.interfaceMethod`: also prepares for the last commit.
5. `reflect: let reflect.Type be of interface type`: this is the big one. The encoding/json package relies on being able to compare `reflect.Type` variables with nil, which this commit makes possible by changing `reflect.Type` to an interface type (like standard Go). All the previous commits are needed to reduce the associated code size increase. There is still some code size increase, but it's only for some bigger programs (for example, that use the fmt package).

One side effect of this PR (in commit 2) is that it improves (but does not fix) issue https://github.com/tinygo-org/tinygo/issues/1415 when you're printing `error` values (e.g. `err := ...; println(err)`).

This is a draft until #1612 is merged. I think it would be possible to separate this PR from #1612 but that might make the first commit more complicated (it was causing linker errors until I rebased on top of #1612).

### DIFF
--- a/interp/interpreter.go
+++ b/interp/interpreter.go
@@ -155,11 +155,8 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 				// which case this call won't even get to this point but will
 				// already be emitted in initAll.
 				continue
-			case callFn.name == "(reflect.Type).Elem" || strings.HasPrefix(callFn.name, "runtime.print") || callFn.name == "runtime._panic" || callFn.name == "runtime.hashmapGet":
+			case strings.HasPrefix(callFn.name, "runtime.print") || callFn.name == "runtime._panic" || callFn.name == "runtime.hashmapGet":
 				// These functions should be run at runtime. Specifically:
-				//   * (reflect.Type).Elem is a special function. It should
-				//     eventually be interpreted, but fall back to a runtime call
-				//     for now.
 				//   * Print and panic functions are best emitted directly without
 				//     interpreting them, otherwise we get a ton of putchar (etc.)
 				//     calls.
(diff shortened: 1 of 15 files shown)


johanbrandhorst:
This might be interesting for running Vecty on TinyGo webassembly (#1282, hexops/vecty/issues/269) CC @slimsag @marwan-at-work.

aykevl:
@johanbrandhorst it'll probably help with Vecty support, but I think support won't be complete yet. For example, `reflect.New` is not yet supported.

Things I'd like to work on after this PR is merged:

  * `(reflect.Type).Implements`. I'm not yet entirely sure how to implement this. It is required for encoding/json support. I might just rely on the fact that in almost all cases the interface is known and it can be converted to an interface type assert.
  * `reflect.New`, this is a little bit complicated because it requires some internal changes, see https://github.com/tinygo-org/tinygo/issues/1087.
  * Improved dead code elimination using [LLVM type metadata](https://llvm.org/docs/TypeMetadata.html) (if possible).

marwan-at-work:
@aykevl thank you and cannot wait for TinyGo to support Vecty :)

aykevl:
Rebased after #1612 was merged.

interp/interpreter.go
```diff
@@ -155,7 +155,7 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 				// which case this call won't even get to this point but will
 				// already be emitted in initAll.
 				continue
-			case callFn.name == "(reflect.Type).Elem" || strings.HasPrefix(callFn.name, "runtime.print") || callFn.name == "runtime._panic" || callFn.name == "runtime.hashmapGet":
+			case strings.HasPrefix(callFn.name, "runtime.print") || callFn.name == "runtime._panic" || callFn.name == "runtime.hashmapGet":
 				// These functions should be run at runtime. Specifically:
 				//   * (reflect.Type).Elem is a special function. It should
 				//     eventually be interpreted, but fall back to a runtime call
```
niaow: This comment looks out of date.

niaow (changes requested):

interp/interpreter.go
aykevl: Thank you for spotting that! I've removed it (it's basically a TODO that's now done).

aykevl:
Updated with suggestion from @niaow and rebased on the dev branch to fix a merge conflict.

deadprogram:
This is very exciting PR right here. :fire: 

Now merging.

marwan-at-work:
@deadprogram @aykevl thank you both ❤️ 

Is it worth just waiting until reflect.New is implemented before thoroughly testing Vecty? 

Thanks!

aykevl:
@marwan-at-work not sure, I haven't really done anything with Vecty yet. You can of course try it.

PR #1728, opened 2021-03-19
housekeeping: ignore files generated by smoketests

kenbell:


sago35 (approved):

aykevl:
Thank you!

I have some local gitignore changes (in .git/info/exclude) so I don't see them, but it's a good idea to ignore these test outputs for everybody.

PR #1729, opened 2021-03-19
goenv: Support Go snap installations by finding the goroot

### LINKED ISSUE #1706: TinyGo does not work with Go installed from Snap
Some people are reporting TinyGo does not work with Go installed from Snap.

See: https://github.com/tinygo-org/tinygo/issues/433#issuecomment-671953170

Nerzal:
I guess this does not need many words :) 

Fixes #1706 

Tested it and i can now build & flash code using TinyGo with a snap installation. 
Also tinygo version output looks good now :) 

```bash
$ which go
/snap/bin/go
```

```bash
./my_tinygo_18 version
tinygo version 0.18.0-dev linux/amd64 (using go version go1.16.2 and LLVM version 11.1.0)
```

current seems to always link on the currenty installed version
```bash
$ ls -lra /snap/go/
insgesamt 8
lrwxrwxrwx  1 root root    4 Mär 19 16:17 current -> 7221
drwxr-xr-x 12 root root  390 Mär 16 23:36 7221
drwxr-xr-x 20 root root 4096 Mär 19 16:17 ..
drwxr-xr-x  3 root root 4096 Mär 19 16:17 .
```

### DIFF
--- a/goenv/goenv.go
+++ b/goenv/goenv.go
@@ -171,8 +171,9 @@ func getGoroot() string {
 	switch runtime.GOOS {
 	case "linux":
 		candidates = []string{
-			"/usr/local/go", // manually installed
-			"/usr/lib/go",   // from the distribution
+			"/usr/local/go",     // manually installed
+			"/usr/lib/go",       // from the distribution
+			"/snap/go/current/", // installed using snap
 		}
 	case "darwin":
 		candidates = []string{


oflebbe:
Ah I see now, why it fixes the problem.  tinygo from the DEB install defaults to path at compile time "/usr/local/go" of tinygo. 
Manual workaround is to use,
`GOROOT=/snap/go/current tinygo build ....`
or apply the proposed patch.

deadprogram:
Thanks for the fix @Nerzal and for testing it out @oflebbe now merging.

PR #1730, opened 2021-03-20
stm32: move f103 (bluepill) to common i2c code

kenbell:
The I2C implementation in the F103 chips is register-compatible with the implementation in F405 / F407, so use that implementation.

Further reduction of duplicate code is possible since these per-chip functions are all essentially identical across all three processors, except for the clock frequency:

- `getFreqRange`
- `getRiseTime`
- `getSpeed`

I'll look at removing that duplication as a separate change.

### DIFF
--- a/src/machine/board_bluepill.go
+++ b/src/machine/board_bluepill.go
@@ -85,6 +85,6 @@ const (
 
 // I2C pins
 const (
-	SDA_PIN = PB7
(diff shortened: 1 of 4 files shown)


deadprogram:
I love PRs that remove LoC, I have to admit. :smiley_cat: 

Merging, thanks @kenbell

PR #426, opened 2019-06-27
machine: add generic board support on non-baremetal hardware

aykevl:
Instead of trying to modify periperhals directly, external functions are
called. For example, __tinygo_gpio_set sets a GPIO pin to a specified
value (high or low). It is expected that binaries made this way will be
linked with some extra libraries that implement support for these
functions.

One particularly interesting case is this experimental board simulator:
https://github.com/aykevl/tinygo-play
Compiling code to WebAssembly with the correct build tag for a board
will enable this board to be simulated in the browser.

Atmel/Microchip based SAMD boards are not currently supported, because
their I2C/SPI support is somewhat uncommon and harder to support in the
machine API. They may require a modification to the machine API for
proper support.

PR #1731, opened 2021-03-20
stm32: housekeeping - remove empty file

kenbell:
This file appears unused and the build tag seems a little odd, might as well remove it


aykevl:
Yes, this looks reasonable.

This file was previously used to define some constants for multiple boards, but I don't think it was ever really needed? They were introduced in https://github.com/tinygo-org/tinygo/pull/426 (by me), possibly by mistake.

PR #1733, opened 2021-03-20
fix pca10040 flashing

oflebbe:
The webpage https://tinygo.org/microcontrollers/pca10040/ states to install the nordic nrfjprog and the segger jlink tools .

This was the only way I was able to get tinygo flash --target pca10040 working on windows. Current configuration choses openocd (not mentioned on the webpage) which does work on macos and linux, but not on windows. On windows openocd (from scoop) does complain about some libusb support missing.

I double checked that nrfjprog works correctly on Windows and MacOS with tinygo. 

Personally I prefer to set it to nrfjprog, even though one has to install additional tools.

### DIFF
--- a/targets/pca10040.json
+++ b/targets/pca10040.json
@@ -1,7 +1,7 @@
 {
 	"inherits": ["nrf52"],
 	"build-tags": ["pca10040"],
-	"flash-method": "openocd",
+	"flash-method": "command",
 	"flash-command": "nrfjprog -f nrf52 --sectorerase --program {hex} --reset",
 	"openocd-interface": "jlink",
 	"openocd-transport": "swd"


oflebbe:
I just found the -programmer flag .  If you do not like the patch I can work on changing the documentation instead.

aykevl:
I think it's good to use OpenOCD as it is easy to install on many platforms and is open source (and I think the scoop version should really be fixed if it doesn't work).

I'm personally fine with either default. I have nrfjprog installed because some operations are easier with it (e.g. erasing the entire chip). Maybe other people have an opinion?



> I just found the -programmer flag . If you do not like the patch I can work on changing the documentation instead.

Adding this to the documentation sounds like a good idea.

oflebbe:
Ok, I tried to improve the site https://github.com/tinygo-org/tinygo-site/pull/157 . Actually digging much deeper into Windows  I managed to work with openocd on pca10040 as well. (WinUSB to the win) . So I agree to use OpenOCD as default. Closing PR

sago35:
Hi @oflebbe 

openocd 0.11.0 was released on 2021/03/07.
There is an official binary for windows now.
I made sure that I could write with feather-nrf52840 + jlink using the official binary on windows 10.

http://openocd.org/getting-openocd/
https://github.com/ntfreak/openocd/releases

On windows, you can use the following method to openocd.
I don't know if it's better to write it in the documentation.

1. Change jlink driver to WinUSB driver in zadig
2. Download openocd 0.11 or newer, and pass the PATH
3. tinygo flash --target XXXX --programmer jlink examples/b1


The build log is as follows.

```
$ tinygo flash --target feather-nrf52840 --size short --programmer jlink examples/blinky1
   7640      32    4080 |    7672    4112
Open On-Chip Debugger 0.11.0 (2021-03-07-12:52)
Licensed under GNU GPL v2
For bug reports, read
        http://openocd.org/doc/doxygen/bugs.html
swd
Info : J-Link EDU Mini V1 compiled Jun  9 2020 13:36:46
Info : Hardware version: 1.00
Info : VTarget = 3.299 V
Info : clock speed 1000 kHz
Info : SWD DPIDR 0x2ba01477
Info : nrf51.cpu: hardware has 6 breakpoints, 4 watchpoints
Info : starting gdb server for nrf51.cpu on 3333
Info : Listening on port 3333 for gdb connections
target halted due to debug-request, current mode: Thread
xPSR: 0x01000000 pc: 0x00000a80 msp: 0x20000400
** Programming Started **
Info : nRF52840-xxAA(build code: D0) 1024kB Flash, 256kB RAM
Warn : Adding extra erase range, 0x00027df8 .. 0x00027fff
** Programming Finished **
** Verify Started **
** Verified OK **
** Resetting Target **
shutdown command invoked
```

PR #1734, opened 2021-03-20
wasm: only export explicitly exported functions

aykevl:
Previously we used the `--export-all` linker flag to export most functions. However, this is not needed and increases WebAssembly binary size by around 200 bytes. Instead, we should only export the functions marked with the `//export` pragma.

I'm investigating how feasible it is to do some post-link transformations (for stack switching and garbage collection). While looking into it, I found that there were a lot more symbols exported from WebAssembly than I would expect. Hence this PR to fix this.


deadprogram:
:scissors: good one, now merging.

PR #1738, opened 2021-03-22
nrf: Added support for writing with nrfjprog on all nrf boards

sago35:
`tinygo flash -programmer command` to write using nrfjprog

This PR makes it possible for all nrf devices to support the same writing style.
Basically, it should be possible to write with openocd, but if for some reason someone wants to use nrfjprog, the `-programmer command` option can be used.

It is related to the following
https://github.com/tinygo-org/tinygo-site/pull/157

### DIFF
--- a/targets/pca10056.json
+++ b/targets/pca10056.json
@@ -2,7 +2,6 @@
 	"inherits": ["nrf52840"],
 	"build-tags": ["pca10056"],
 	"flash-method": "command",
-	"flash-command": "nrfjprog -f nrf52 --sectorerase --program {hex} --reset",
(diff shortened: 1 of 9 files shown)


targets/pca10056.json
```diff
@@ -2,7 +2,6 @@
 	"inherits": ["nrf52840"],
 	"build-tags": ["pca10056"],
 	"flash-method": "command",
```
sago35, line 4: Only pca10056 has command as the default.
It might be better to adapt it to openocd, but I don't know.
I will fix it if someone wants to.

PR #1739, opened 2021-03-23
reflect: implement a number of stub functions

aykevl:
These stub functions are necessary for the encoding/json package. They don't seem to be called in trivial cases, so leave them as simple stubs for now.

Together with #1727 this gets the encoding/json package to compile. It doesn't work yet (`.Implements` needs to be implemented) but it's a big step forward.


aykevl:
I have a local change that gets `.Implements` to work, it can successfully run `json.Marshal([]string{"foo", "bar"})`.

deadprogram:
You are on a roll @aykevl! Now merging.

PR #1741, opened 2021-03-24
Initial JSON support

aykevl:
This PR adds initial support for the encoding/json package. It is far from complete (structs don't work yet, for example) but it seems to actually work for encoding some trivial types.


deadprogram:
Still much to be done, but a most impressive step forward, and with the unit tests to prove it. Good work @aykevl 

Now merging.

atdiar:
Does tinygo support the use of encoding/json for unmarshalling json formatted data into the basic go types (besides custom structs) now?
I need to unmarshal into map[string]interface{}.

PR #1742, opened 2021-03-24
Support for pca10059 board

oflebbe:
This is a board support without a softdevice.

The pca10059 is a NRF52840 Chip on a USB Dongle, similar to the nrf52850-mdk-usb-dongle. Amazingly most of the definitions already present for nrf52850-mdk-usb-dongle seem to be suitable for this board as well. 
It uses a MBR with "Open Firmware" to be flashed either by the nRF Connect software or by the python tool "nrfutil". 
I did my best to document the LED and ADC ports, though I only used blinky examples for now and the UART over USB serial.

Tested with tinygo build, tinygo flash on macos. 

When I am more into tinygo,  I will surely look into supporting an softdevice, in an followup patch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -267,6 +267,10 @@ smoketest:
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=pca10056            examples/blinky2
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=pca10059            examples/blinky1
+	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=pca10059            examples/blinky2
+	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=itsybitsy-m0        examples/blinky1
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=feather-m0          examples/blinky1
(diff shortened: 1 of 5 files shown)


oflebbe:
hi @deadprogram , possible to include this into tinygo?

deadprogram:
I thought that the PCA10059 was this dongle from Nordic Semi? https://www.nordicsemi.com/Software-and-tools/Development-Kits/nRF52840-Dongle

oflebbe:
It is!

deadprogram:
Then shouldn't https://github.com/tinygo-org/tinygo/pull/1742/files#diff-1edbec16299d4167b07a2c5fb2a3c443bc7a0b05839fd611cbb0aeee9a3966adR57-R58 refer to that specifically?

oflebbe:
Ironically, both the usb_VID, usb_PID are the same as the nrf52850-mdk-usb-dongle and  nrf52850-mdk. 
My mac actually is using the same Product Name as tinygo (see pic)

Bildschirmfoto 2021-03-28 um 12 58 43

oflebbe:
Looks like I was extrapolating from MacOS , Not sure what manufactorer name should be: Should it be the cleartext of the Vendor ID ? Than  nrf52850-mdk-usb-dongle and nrf52850-mdk are wrong as well.

Looking at other sources for the USB Product 0xcafe Id for Nordics Vendor ID is not generally known. Not sure what we are using the names for ...

deadprogram:
We send the name as part of returning the USB descriptor. So we can set to anything we want. Generally we have tried to stick to whatever the vendor uses themselves. This case is a little different, so I think what you have is cool. Do you mind squashing into a single commit please?

oflebbe:
Thanks! Squashed it.

deadprogram:
Great, thanks @oflebbe now merging.

PR #1743, opened 2021-03-24
reflect: implement Value.CanAddr

aykevl:
It is used in the crypto/sha512 test, for example. And it is very simple to implement.


deadprogram:
Looks good, now merging. Thanks @aykevl

PR #1752, opened 2021-03-28
compiler: fix "fragment covers entire variable" bug

aykevl:
This bug could sometimes be triggered by syscall/js code it seems. But it's a generic bug, not specific to WebAssembly.

I have fixed it by skipping zero-length fields of structs when splitting them for variable passing.


deadprogram:
Seems like a reasonable guard to have in place. Now merging, thanks @aykevl

PR #1753, opened 2021-03-28
reflect: implement Sizeof and Alignof for func values

aykevl:
This is a small change that appears to be necessary for encoding/json support. It's simple enough to implement.


deadprogram:
:triangular_ruler: merging.

PR #1453, opened 2020-10-21
atsamd5x: improve SPI

sago35:
I made it faster by rewriting SPI.Tx().
Originally, there was a gap of about 400ns per byte, but this is a significant improvement.

After this PR is merged, drivers/ili9341 should be able to simplify it.
https://github.com/tinygo-org/drivers/pull/208

old: 6ab0106af37abf65bdc9e41912d7e8a46db9bd8d
image

new txrx():
image

new tx():
image

new rx():
image

PR #1754, opened 2021-03-29
atsamd5x: add SPI.TxN()

sago35:
I created `SPI.TxN()` to repeat `SPI.Tx()` n times.

```go
// pseudo code
SPI.TxN(tx, rx []byte, n int) {
    for i := 0; i < n; i++ {
        SPI.Tx(tx, rx)
    }
}
```

`SPI.Tx()` is not enough to run ili9341 on standard SPI for speed reasons.
In particular, `SPI.TxN()` is faster than SPI.Tx() in the following applications.

* Fill with a specific color in ili9341
    * https://github.com/tinygo-org/drivers/pull/242

`SPI.Tx()` is slower because it waits for the SPI to finish communication for each function call.

testcode:

```go
package main

import (
	"fmt"
	"machine"
	"time"
)

func main() {
	//led := machine.LED
	//led.Configure(machine.PinConfig{Mode: machine.PinOutput})

	spi := machine.SPI0
	spi.Configure(machine.SPIConfig{
		Frequency: 48000000,
	})

	txbuf := make([]byte, 4)
	rxbuf := make([]byte, 4)

	for i := range txbuf {
		txbuf[i] = byte(i * 2)
		rxbuf[i] = byte(i * 2)
	}

	testNo := 0
	for {
		//led.Toggle()
		time.Sleep(time.Millisecond * 500)

		switch testNo % 4 {
		case 0:
			// txrx
			spi.TxN(txbuf, rxbuf, 16)

			for i := range txbuf {
				if txbuf[i] != rxbuf[i] {
					fmt.Printf("%d : got %02X want %02X\r\n", i, rxbuf[i], txbuf[i])
				} else {
					fmt.Printf(".")
				}
				rxbuf[i] = 0
			}
		case 1:
			// tx
			spi.TxN(txbuf, nil, 16)
		case 2:
			// rx
			spi.TxN(nil, rxbuf, 16)
		case 4:
			// skip
		}

		testNo++
	}
}
```

The speed is about the same compared to #1453.

txrxn:
image

txn:
image

rxn:
image

### DIFF
--- a/src/machine/machine_atsamd51.go
+++ b/src/machine/machine_atsamd51.go
@@ -1560,6 +1560,76 @@ func (spi SPI) txrx(tx, rx []byte) {
 	rx[len(rx)-1] = byte(spi.Bus.DATA.Get())
 }
 
+// TxN handles read/write operation for SPI interface. The difference with Tx() is that
+// it repeats the process n times.
+//
+func (spi SPI) TxN(w, r []byte, n int) error {
+	switch {
+	case w == nil:
+		// read only, so write zero and read a result.
+		spi.rxn(r, n)
+	case r == nil:
+		// write only
+		spi.txn(w, n)
+
+	default:
+		// write/read
+		if len(w) != len(r) {
+			return ErrTxInvalidSliceSize
+		}
+
+		spi.txrxn(w, r, n)
+	}
+	return nil
+}
+
+func (spi SPI) txn(tx []byte, n int) {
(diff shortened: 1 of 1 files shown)


deadprogram:
@sago35 is this PR something you are still working on, or should we close it?

sago35:
@deadprogram 
At this point, it would be better to adopt #3985. I will close this PR.

deadprogram:
Thanks for the reply @sago35 now closing.

PR #1755, opened 2021-03-29
atsamd21, atsamd51, nrf52840: unify usbcdc code

sago35:
The code for the three targets has the same form of USBCDC speedup.
In this PR, I unified the points where each was slightly different.
This change will make it easier to maintain in the future.

The changes are as follows

* Move `usbcdc.waitTxcRetryCount++` to `USBCDC.Flush()`
* Make `USBCDC.WriteByte()` not a pointer receiver.
* `USBCDC.FLush()` is now called from SoF only.

### DIFF
--- a/src/machine/machine_atsamd51.go
+++ b/src/machine/machine_atsamd51.go
@@ -1850,6 +1850,7 @@ func (usbcdc *USBCDC) Flush() error {
 
 			if usbcdc.waitTxc {
 				// waiting for the next flush(), because the transmission is not complete
+				usbcdc.waitTxcRetryCount++
 				return nil
 			}
 			usbcdc.waitTxc = true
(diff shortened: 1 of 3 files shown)


src/machine/machine_atsamd51.go
```diff
@@ -1882,7 +1883,7 @@ func (usbcdc *USBCDC) Flush() error {
 }
 
 // WriteByte writes a byte of data to the USB CDC interface.
-func (usbcdc *USBCDC) WriteByte(c byte) error {
+func (usbcdc USBCDC) WriteByte(c byte) error {
```
deadprogram: Is there a specific reason you changed the signature here? I think we probably do want to keep it as a pointer receiver.

sago35: No, there is no particular reason.
I just made it look the same as `machine_nrf52840_usb.go` .

I'll fix it.

sago35:
Updated. @deadprogram

deadprogram:
Can you please squash these commits together @sago35? Thanks!

sago35:
Squashed and rebased.

deadprogram:
Thanks for this unification @sago35 

@ardnew is working on an even bigger unification, so this will certainly help with that as well.

Now merging.

PR #1758, opened 2021-03-29
Improve caching / speed up build by running interp per package

aykevl:
Run package initializers per package, in addition to once for the entire program. It is still run for the entire program (to benefit from whole-program visibility), but it doesn't have to do much in that case so it can be very fast.

This results in a significant speedup in some cases. For example, this runs over twice as fast with a warm cache:

    tinygo build -o test.elf ./testdata/stdlib.go

This should help a lot with edit-compile-test cycles, that typically only modify a single package.

This required some changes to the interp package to deal with globals created in a previous run of the interp package and to deal with external globals (that can't be loaded from).

The main package that is sped up is the unicode package, which may take over a second to run `interp.RunFunc` for. Moving this into the (cached) package build saves this time in recompilations.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -55,6 +55,7 @@ type BuildResult struct {
 type packageAction struct {
 	ImportPath      string
 	CompilerVersion int // compiler.Version
+	InterpVersion   int // interp.Version
 	LLVMVersion     string
 	Config          *compiler.Config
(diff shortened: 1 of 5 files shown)


deadprogram:
:racehorse: merging. Thanks @aykevl

PR #1759, opened 2021-03-30
//go:section pragma

aykevl:
This PR adds a new pragma for functions and globals to set the
section name. This can be useful to place a function or global in a
special device specific section, for example:

  * Functions may be placed in RAM to make them run faster, or in flash
    (if RAM is the default) to not let them take up RAM.
  * DMA memory may only be placed in a special memory area.
  * Some RAM may be faster than other RAM, and some globals may be
    performance critical thus placing them in this special RAM area can
    help.
  * Some (large) global variables may need to be placed in external RAM,
    which can be done by placing them in a special section.

To use it, you have to place a function or global in a special section,
for example:

```go
//go:section .externalram
var externalRAMBuffer [1024]byte
```

This can then be placed in a special section of the linker script, for
example something like this:

    .bss.extram (NOLOAD) : {
        *(.externalram)
    } > ERAM

This PR also adds some testing for pragmas, for regression testing.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -715,11 +715,11 @@ func (c *compilerContext) createPackage(irbuilder llvm.Builder, pkg *ssa.Package
 		member := pkg.Members[name]
 		switch member := member.(type) {
 		case *ssa.Function:
+			// Create the function definition.
+			b := newBuilder(c, irbuilder, member)
 			if member.Blocks == nil {
 				continue // external function
 			}
-			// Create the function definition.
-			b := newBuilder(c, irbuilder, member)
 			b.createFunction()
 		case *ssa.Type:
 			if types.IsInterface(member.Type()) {
@@ -758,8 +758,8 @@ func (c *compilerContext) createPackage(irbuilder llvm.Builder, pkg *ssa.Package
 		case *ssa.Global:
 			// Global variable.
 			info := c.getGlobalInfo(member)
+			global := c.getGlobal(member)
 			if !info.extern {
-				global := c.getGlobal(member)
 				global.SetInitializer(llvm.ConstNull(global.Type().ElementType()))
 				global.SetVisibility(llvm.HiddenVisibility)
 			}
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -762,6 +762,9 @@ func (c *compilerContext) createPackage(irbuilder llvm.Builder, pkg *ssa.Package
 			if !info.extern {
 				global.SetInitializer(llvm.ConstNull(global.Type().ElementType()))
 				global.SetVisibility(llvm.HiddenVisibility)
+				if info.section != "" {
+					global.SetSection(info.section)
+				}
 			}
 		}
 	}
@@ -786,6 +789,9 @@ func (b *builder) createFunction() {
 	if !b.info.exported {
 		b.llvmFn.SetVisibility(llvm.HiddenVisibility)
 		b.llvmFn.SetUnnamedAddr(true)
+	}
+	if b.info.section != "" {
+		b.llvmFn.SetSection(b.info.section)
 	}
 	if b.info.exported && strings.HasPrefix(b.Triple, "wasm") {
 		// Set the exported name. This is necessary for WebAssembly because
--- a/compiler/compiler_test.go
+++ b/compiler/compiler_test.go
@@ -45,6 +45,7 @@ func TestCompiler(t *testing.T) {
 		{"float.go", ""},
 		{"interface.go", ""},
 		{"func.go", ""},
+		{"pragma.go", ""},
 		{"goroutine.go", "wasm"},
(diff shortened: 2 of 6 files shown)


aykevl:
@ardnew @kenbell @sago35 this may be of interest.

kenbell:
nice!

sago35:
I'll try it later.

kenbell:
I've updated the 'persistent ram' prototype to use this mechanism: #1715 

Seems to work fine.

deadprogram:
@aykevl one more merge conflict needs to be resolved, please.

sago35:
@aykevl 

The following source code worked fine.
When writing to `targets/arm.ld`, I would like to name it something like the `beginning of .data`.
But I don't know what name would be appropriate.

```
/* targets/arm.ld */

    /* Globals with initial value */
    .data :
    {
        . = ALIGN(4);
        _sdata = .;        /* used by startup code */
        *(.canram)
        *(.canram.*)
        *(.data)
        *(.data.*)
        . = ALIGN(4);
        _edata = .;        /* used by startup code */
    } >RAM AT>FLASH_TEXT
```

```go
//go:section .canram
//go:align 4
var CANRxFifo [2][1024]byte

//go:section .canram
//go:align 4
var CANTxFifo [2][1024]byte

//go:section .canram
//go:align 4
var CANEvFifo [2][1024]byte
```

deadprogram:
@aykevl reminder this has merge conflict that needs resolution, please.

sago35:
With the following code, I was able to place the variable dummy in .text.
There is a PR to put tinyfont data in the const area, but it seems to be correct to use this pragma.

* https://github.com/tinygo-org/tinyfont/pull/23


```go
package main

import (
	"fmt"
	"machine"
	"time"
)

//go:section .constdata
var dummy = []byte("123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890" +
	"123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890" +
	"123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890" +
	"123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890" +
	"123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890" +
	"123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890" +
	"123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890" +
	"123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890" +
	"123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890")

func main() {
	led := machine.LED

	for {
		led.Toggle()
		println("hello world!")
		fmt.Printf("%p %q\r\n", &dummy, string(dummy))
		time.Sleep(time.Second)
	}
}
```

```
    .text :
    {
        KEEP(*(.isr_vector))
        *(.text)
        *(.text.*)
        *(.rodata)
        *(.rodata.*)
        *(.constdata)
        *(.constdata.*)
        . = ALIGN(4);
    } >FLASH_TEXT
```

sago35:
@aykevl 
I want a section pragma.
Please resolve conflicts and rebase.

aykevl:
I have rebased the PR to resolve the merge conflict.

> With the following code, I was able to place the variable dummy in .text.
> There is a PR to put tinyfont data in the const area, but it seems to be correct to use this pragma.

Hmm, this can be a problem. Maybe `//go:section` should be an unsafe feature for this reason, just like `//go:linkname`. The problem is that modifying such an array is not possible and there is no way the type system can prevent this.
A better solution might be to store font data as strings: strings can be indexed in the same way as byte arrays but will be stored in flash because they cannot be modified.
An alternative is file embed support, which isn't yet implemented in Go.

deadprogram:
> Maybe //go:section should be an unsafe feature for this reason, just like //go:linkname

Absolutely yes, this should also require `unsafe`.

aykevl:
Okay, I've updated it to require an import of `unsafe` in packages that use `//go:section` because it allows to do unsafe things.
Just import package unsafe to use it, even if that's `import _ "unsafe"`.

deadprogram:
Thanks @aykevl for the revisions. Now merging.

PR #1763, opened 2021-04-01
compiler: decouple func lowering from interface type codes

aykevl:
There is no good reason for func values to refer to interface type codes. The only thing they need is a stable identifier for function signatures, which is easily created as a new kind of globals. Decoupling makes it easier to change interface related code.

This change should not have any effect on the generated code, it's just a refactor.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -23,7 +23,7 @@ import (
 // Version of the compiler pacakge. Must be incremented each time the compiler
 // package changes in a way that affects the generated LLVM module.
 // This version is independent of the TinyGo version number.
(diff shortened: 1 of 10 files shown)


deadprogram:
@aykevl merge conflict resolution needed, please.

aykevl:
I'll need to make some bigger changes to adjust this PR after #1770.
As it happens, this pass is the same as the one that's triggering the bug in #1771. I might do them both at the same time.

aykevl:
I've rebased it against the dev branch and updated it a bit to be more robust. The assert-test-linux CI job is still failing because this doesn't include #1783 yet.

deadprogram:
Looking good now merging.

Thanks @aykevl

PR #1767, opened 2021-04-03
support nrfutil zip format

### LINKED ISSUE #1757: Cannot flash pca10059 in windows environment

oflebbe:
Support for flashing the pca10059 from windows. Fixes #1757 . Works for me (even automatic port detection, what a surprise)
@sago35 would you like to check it  as well?


src/machine/board_pca10059.go
```diff
@@ -23,9 +23,9 @@ const (
 // ADC pins
 const (
 	ADC1 Pin = 2
-	ADC2 Pin = 4
-	ADC3 Pin = 29
-	ADC4 Pin = 31
+	ADC2 Pin = 29
+	ADC3 Pin = 31
+	ADC4 Pin = 4 // only available as a pad on downside
```
sago35: Does this change have anything to do with nrfutil zip?
If not, please give us another PR so that we can understand the commit history better.

Also, do you have any reason for this change?
I did not understand the meaning of the following.

> move Pin 0.4 to the end of list since almost unreachable

builder/nrfutil.go
```diff
@@ -0,0 +1,26 @@
...
+		fmt.Printf("%s\n ", strings.Join(cmdLine, " "))
+	}
+
+	cmd := exec.Command(cmdLine[0], cmdLine[1:]...)
+	err := cmd.Run()
+	if err != nil {
+		return fmt.Errorf("could not run nrfutil pkg generate: %w", err)
+	}
+	return nil
+}
```
sago35: The following error message was output.

```
C:\tinygo\tinygo>tinygo flash --target pca10059 --size short -x -port COM33 ./src/examples/blinky1
   7368      32    4080 |    7400    4112
nrfutil pkg generate --hw-version 52 --sd-req 0x0 --debug-mode --application C:\Users\sago3\AppData\Local\Temp\tinygo022542263\main.hex C
:\Users\sago3\AppData\Local\Temp\tinygo022542263\main.zip
 error: could not run nrfutil pkg generate: exit status 4294967295
```

It seems that nrfutil on windows needs to handle standard output.
I downloaded nrfutil from the following URL

https://github.com/NordicSemiconductor/pc-nrfutil/releases/tag/v6.1

```
$ nrfutil.exe version
nrfutil version 6.1.0
```

So, I don't know the details, but I would like to add the following to `tinygo flash` from windows.

```suggestion
import (
	"fmt"
	"io/ioutil"
	"os/exec"
	"strings"

	"github.com/tinygo-org/tinygo/compileopts"
)

// https://infocenter.nordicsemi.com/index.jsp?topic=%2Fug_nrfutil%2FUG%2Fnrfutil%2Fnrfutil_intro.html

func makeDFUFirmareImage(options *compileopts.Options, infile, outfile string) error {
	cmdLine := []string{"nrfutil", "pkg", "generate", "--hw-version", "52", "--sd-req", "0x0", "--debug-mode", "--application", infile, outfile}

	if options.PrintCommands {
		fmt.Printf("%s\n ", strings.Join(cmdLine, " "))
	}

	cmd := exec.Command(cmdLine[0], cmdLine[1:]...)
	// It seems that nrfutil on windows needs to handle standard output
	cmd.Stdout = ioutil.Discard
	err := cmd.Run()
	if err != nil {
		return fmt.Errorf("could not run nrfutil pkg generate: %w", err)
	}
	return nil
}
```

compileopts/config.go
```diff
@@ -238,6 +238,9 @@ func (c *Config) BinaryFormat(ext string) string {
 		// More information:
 		// https://github.com/Microsoft/uf2
 		return "uf2"
+	case ".zip":
```
sago35, line 257: The name zip is probably too common and will cause problems later.
I think the name `nrf-dfu`, as written by @aykevl , is fine.

main.go
```diff
@@ -221,6 +221,8 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 			fileExt = ".bin"
 		case strings.Contains(config.Target.FlashCommand, "{uf2}"):
 			fileExt = ".uf2"
+		case strings.Contains(config.Target.FlashCommand, "{zip}"):
```
sago35: The name zip is probably too common and will cause problems later.
I think the name `nrf-dfu`, as written by @aykevl , is fine.

targets/pca10059.json
```diff
@@ -2,5 +2,5 @@
 	"inherits": ["nrf52840"],
 	"build-tags": ["pca10059"],
 	"linkerscript": "targets/pca10059.ld",
-	"flash-command": "nrfutil pkg generate --hw-version 52 --sd-req 0x0 --application {hex} --application-version 1 /tmp/tinygo_$$.zip && nrfutil dfu usb-serial -pkg /tmp/tinygo_$$.zip -p {port} -b 115200 && rm -f /tmp/tinygo_$$.zip"
+	"flash-command": "nrfutil dfu usb-serial -pkg {zip} -p {port} -b 115200"
```
sago35: The name zip is probably too common and will cause problems later.
I think the name `nrf-dfu`, as written by @aykevl , is fine.

sago35:
@oflebbe 
Thank you for your contribution.
Most of it is good, but I did make a minor point.
Please check it.

compileopts/config.go
oflebbe: Yeah, I am aware of it. But there are some assumption right into tinygo:
1) We can chose suffixes (we can't since `nrfutil` does only accept `zip` for dfu packages, sadly.
2) filename suffices correspond 1:1 to methods (as you have written `zip` is widely used for many things, so a `tinygo build -o xx.zip ` is not  unambigous.
3) filename suffices and methods have to have the same name  (coded in main.go)

I cannot work around 1 and 2 but will provide a change for (3).

src/machine/board_pca10059.go
oflebbe: THis slipped in accidently. Motivation was support for the `examples/adc` code. The `PIN 0.4` aka `4` of the pca10059 board is located on the bottom of the USB stick in a quite unaccessible manner.
See 
Bildschirmfoto 2021-04-05 um 18 59 46

builder/nrfutil.go
oflebbe: Great thanks! Not sure why my windows installation wasn't picky about that

oflebbe:
Thanks for your review. I reworked the part for zip / nrf-dfu detection.

sago35:
You can use the BinaryFormat() function as in esp32.
For example, you can use the following json

```
{
	"inherits": ["nrf52840"],
	"build-tags": ["pca10059"],
	"linkerscript": "targets/pca10059.ld",
	"binary-format": "nrf-dfu",
	"flash-command": "nrfutil dfu usb-serial -pkg {zip} -p {port} -b 115200"
}
```

see:
https://github.com/tinygo-org/tinygo/blob/release/compileopts/config.go#L228-L230
https://github.com/tinygo-org/tinygo/blob/release/targets/esp32.json#L18

compileopts/config.go
```diff
@@ -238,6 +238,9 @@ func (c *Config) BinaryFormat(ext string) string {
...
+		// Nordic's proprietary DFU update package build by nrfutil
+		return "nrf-dfu"
```
sago35, line 259: It is better to use nrf-dfu only when binary-format is specified for json.
The zip file format may be used for other purposes in the future.

You will need to change the JSON as well.


```suggestion
		if c.Target.BinaryFormat != "" {
			return c.Target.BinaryFormat
		}
		return "zip"
```

targets/pca10059.json
```diff
@@ -2,5 +2,5 @@
 	"inherits": ["nrf52840"],
 	"build-tags": ["pca10059"],
 	"linkerscript": "targets/pca10059.ld",
-	"flash-command": "nrfutil pkg generate --hw-version 52 --sd-req 0x0 --application {hex} --application-version 1 /tmp/tinygo_$$.zip && nrfutil dfu usb-serial -pkg /tmp/tinygo_$$.zip -p {port} -b 115200 && rm -f /tmp/tinygo_$$.zip"
+	"flash-command": "nrfutil dfu usb-serial -pkg {nrf-dfu} -p {port} -b 115200"
```
sago35, line 5:
```suggestion
	"binary-format": "nrf-dfu",
	"flash-command": "nrfutil dfu usb-serial -pkg {zip} -p {port} -b 115200"
```

main.go
```diff
@@ -210,27 +210,38 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 	var fileExt string
 
 	flashMethod, _ := config.Programmer()
+	var fileToken string
```
sago35, line 267: By using binary-format, I think this change will be unnecessary.

src/machine/board_pca10059.go
sago35: I see.
Could you please separate the PRs for easier review?
Thank you.

compileopts/config.go
aykevl: I agree with @sago35 here.

compileopts/config.go
aykevl: The filename suffix can remain `.zip`. That's fine. You can choose what format this zip file is using `BinaryFormat`, as suggested by @sago35.

aykevl (commented):
This is very close, but not entirely. I agree with the suggestions from @sago35, they would make the .zip extension be the DFU firmware format when selected using `BinaryFormat`.

Is there anything that is blocking this?

oflebbe:
hi i wasn‘t aware that there is a pr
still open. let me look into it in the next days

oflebbe:
rebased on dev

deadprogram:
https://github.com/tinygo-org/tinygo/pull/1767/commits/7cad1b2c890e8e01cb75d0b5074fcb167ec1bead is a merge commit, which is why all of the 130+ extra commits appear in this branch. 

I would personally probably create a new branch off of dev, and then use `git cherry-pick` to pull in your own couple of commits for your specific changes into this new branch. Or something like that?

oflebbe:
Will be continued in #1931

PR #1768, opened 2021-04-04
transform: fix bug in interface lowering when signatures are renamed

### LINKED ISSUE #1765: compiler: regression failure for build on Adafruit Clue demo with ST7789 display
A regression bug was introduced with e2f532709f97b37575d4216c9605323221b63b93 as determined using `git bisect`:

```
$ tinygo flash -size short -target=clue .
{ i8, i8, i8, i8 } %5
 %"image/color.CMYK" = type { i8, i8, i8, i8 }  %9 = call fastcc %runtime._interface @"(*tinygo.org/x/drivers/st7789.Device).FillRectangle"(%"tinygo.org/x/drivers/st7789.Device"* %8, i16 %1, i16 %2, i16 %3, i16 %4, { i8, i8, i8, i8 } %5, i8* %6, i8* %7), !dbg !9963
{ i8, i8, i8, i8 } %3
 %"image/color.CMYK" = type { i8, i8, i8, i8 }  call fastcc void @"(*tinygo.org/x/drivers/st7789.Device).SetPixel"(%"tinygo.org/x/drivers/st7789.Device"* %6, i16 %1, i16 %2, { i8, i8, i8, i8 } %3, i8* %4, i8* %5), !dbg !9963
error: optimizations caused a verification failure
```

The code was that I used in my FOSDEM 2021 talk on the Adafruit Clue:
https://github.com/deadprogram/fosdem-2021/blob/main/demo/scanner-ble/main.go

aykevl:
In rare cases the signature might change as a result of LLVM renaming
some named struct types when multiple LLVM modules are merged. The
easiest workaround is to detect such mismatched signatures and adding a
bitcast: this should be safe as the underlying data is effectively of
the same type.

This should fix #1765.

Note that this doesn't have tests as I don't really know the exact behavior of LLVM module merging and I have no idea what a good test would look like. Additionally, I intend to refactor this code in the future in a way that might avoid this issue entirely.

### DIFF
--- a/transform/interface-lowering.go
+++ b/transform/interface-lowering.go
@@ -703,6 +703,20 @@ func (p *lowerInterfacesPass) createInterfaceMethodFunc(itf *interfaceInfo, sign
 			// function.
 			receiver = p.builder.CreateBitCast(receiver, function.FirstParam().Type(), "")
 		}
+
+		// Check whether the called function has the same signature as would be
+		// expected from the parameters. This can happen in rare cases when
+		// named struct types are renamed after merging multiple LLVM modules.
+		paramTypes := []llvm.Type{receiver.Type()}
+		for _, param := range params {
+			paramTypes = append(paramTypes, param.Type())
+		}
+		calledFunctionType := function.Type()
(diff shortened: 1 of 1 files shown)


deadprogram:
Confirmed that I can now build that program with this PR. Thanks for the quick fix @aykevl now merging.

PR #1771, opened 2021-04-05
transform: use IPSCCP pass instead of the constant propagation pass

aykevl:
The constant propagation pass is removed in LLVM 12, so this pass needs
to be replaced anyway. The direct replacement would be the SCCP (sparse
conditional constant propagation) pass, but perhaps a better replacement
is the IPSCCP pass, which is an interprocedural version of the SCCP
pass and propagates constants across function calls if possible.

This is not always a code size reduction, but it appears to reduce code
size in a majority of cases. It certainly reduces code size in almost
all WebAssembly tests I did.

### DIFF
--- a/transform/optimizer.go
+++ b/transform/optimizer.go
@@ -66,7 +66,7 @@ func Optimize(mod llvm.Module, config *compileopts.Config, optLevel, sizeLevel i
 		defer goPasses.Dispose()
 		goPasses.AddGlobalDCEPass()
 		goPasses.AddGlobalOptimizerPass()
-		goPasses.AddConstantPropagationPass()
+		goPasses.AddIPSCCPPass()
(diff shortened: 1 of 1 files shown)


deadprogram:
Code size reduction, what's not to love!

Now merging, thanks @aykevl

deadprogram:
Looks like the assert failure was real in this case:

https://app.circleci.com/pipelines/github/tinygo-org/tinygo/4698/workflows/737f4183-f460-4f69-8b69-1f3977c1c815/jobs/26630

deadprogram:
@aykevl did you see the above failure?

aykevl:
Ouch, yes that looks like a bug.

aykevl:
I can reproduce it locally. I might not be able to fix it today, hopefully tomorrow (but it's first on my priority list!).

PR #1774, opened 2021-04-06
ci: limit number of jobs for assert-test-linux

aykevl:
This job is causing OOM errors on CircleCI so limit it to just two jobs
(which should be fine on a 2CPU executor). Hopefully this fixes the
errors in CI that have occured recently.


deadprogram:
Looks good, now merging.

Thanks @aykevl!

PR #1775, opened 2021-04-06
Inject build time variables

### LINKED ISSUE #1045: Inject build-time variables
With go you can inject variables using ldflags like this:

`go build -o ./main.wasm -ldflags "-X main.Version=$(VERSION)" ./main.go`

Attempting this with tinygo:

`tinygo build -o ./main.wasm -target wasm -ldflags "-X main.Version=$(VERSION)"  ./main.go`

Gives:
```
tinygo:wasm-ld: error: unknown argument: -X
tinygo:wasm-ld: error: cannot open main.Version=1.0.0: No such file or directory
error: failed to link /var/folders/jc/3lr9ltt96k90h_mg_xq6pvbw0000gn/T/tinygo833831446/main: failed to link using built-in wasm-ld
```

Is there an equivalent way to achieve this in tinygo?

aykevl:
This PR implements replacing some global variables with a different value, if the global variable has no initializer. For example, if you have:

```go
package main

var version string
```

you can replace the value with `-ldflags="-X main.version=0.2"`.

Right now it only works for uninitialized globals. The Go tooling also supports initialized globals (`var version = "<undefined>"`) but that is a bit hard to combine with how initialized globals are currently implemented.

The current implementation still allows caching package IR files while making sure the values don't end up in the build cache. This means compiling a program multiple times with different values will use the cached package each time, inserting the string value only late in the build process.

This PR also has a few commits to remove the older `-cflags` and `-ldflags` (which are not necessary anymore as they are supported directly in CGo) and to clean up and improve testing.

Fixes #1045.


deadprogram:
It would appear the Windows build is failing on this.

aykevl:
I hope I fixed it.

aykevl:
I fixed it... but assert-test-linux is failing again. I hope #1777 will now really fix that.

aykevl:

deadprogram:
Tested and worked perfectly. Thanks @aykevl

PR #1777, opened 2021-04-06
ci: limit test runs of assert-test-linux to two jobs

aykevl:
Instead of the regular build, it's the `make test` line that fails due
to OOM. This is because testing means that a lot of test binaries need
to be built while the regular build only needs to link one binary.

This improves https://github.com/tinygo-org/tinygo/pull/1774 and should
hopefully actually fix the OOM errors.


deadprogram:
Seems to have worked, so now merging.

PR #1779, opened 2021-04-07
modules: add latest go-llvm because seems like older SHA is missing?

deadprogram:
This PR adds the latest go-llvm because seems like older SHA is missing from all the errors that seem to just have appeared.


deadprogram:
Merging just to stop the build errors!

PR #1782, opened 2021-04-08
compiler: check for errors

aykevl:
Some errors were generated but never returned or never checked in the test function. That's a problem. Therefore this commit fixes this oversight (by me).


deadprogram:
:heavy_check_mark: now merging

PR #1783, opened 2021-04-09
transform: fix func lowering assertion failure

aykevl:
The func-lowering pass has started to fail in the dev branch, probably
as a result of replacing the ConstPropagation pass with the IPSCCP pass.
This commit makes the code a bit more robust and should be able to
handle all possible cases (not just ptrtoint).

### DIFF
--- a/transform/allocs.go
+++ b/transform/allocs.go
@@ -43,6 +43,18 @@ func OptimizeAllocs(mod llvm.Module) {
 			continue
 		}
 
+		if size == 0 {
+			// If the size is 0, the pointer is allowed to alias other
+			// zero-sized pointers. Use the pointer to the global that would
+			// also be returned by runtime.alloc.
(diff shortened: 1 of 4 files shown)


aykevl:
Sigh. Something else is broken now, and it might be a bug in LLVM itself.

aykevl:
Yup, it appears to be a bug in LLVM: https://bugs.llvm.org/show_bug.cgi?id=49916

Now I'll need to see whether there is a way to work around it...

aykevl:
I think I have a workaround. Let's see whether CI passes this time.

Both commits fix different issues and both should be included. I believe both are surfaced by switching to the IPSCCP pass.

aykevl:
Finally, assert-test-linux passes!

deadprogram:
:bug: in LLVM is always the hardest category. Good work on this @aykevl now merging.

PR #1788, opened 2021-04-11
Fix RGBA color interpretation for GameBoyAdvance

Agurato:
It appears that the implementation for color parameter in SetPixel was wrong.
This PR aims to better retranscribe original color (24 bit) to GBA palette (15-bit)

### DIFF
--- a/src/machine/machine_gameboyadvance.go
+++ b/src/machine/machine_gameboyadvance.go
@@ -54,7 +54,7 @@ func (d FramebufDisplay) Size() (x, y int16) {
 }
 
 func (d FramebufDisplay) SetPixel(x, y int16, c color.RGBA) {
-	d.port[y][x].Set(uint16(c.R)&0x1f | uint16(c.G)&0x1f<<5 | uint16(c.B)&0x1f<<10)
+	d.port[y][x].Set((uint16(c.R) >> 3) | ((uint16(c.G) >> 3) << 5) | ((uint16(c.B) >> 3) << 10))
 }
 
 func (d FramebufDisplay) Display() error {


deadprogram:
Anybody who is into GBA want to comment before I merge?

src/machine/machine_gameboyadvance.go
```diff
@@ -54,7 +54,7 @@ func (d FramebufDisplay) Size() (x, y int16) {
 }
 
 func (d FramebufDisplay) SetPixel(x, y int16, c color.RGBA) {
-	d.port[y][x].Set(uint16(c.R)&0x1f | uint16(c.G)&0x1f<<5 | uint16(c.B)&0x1f<<10)
+	d.port[y][x].Set(((uint16(c.R) >> 3) & 31) | (((uint16(c.G) >> 3) & 31) << 5) | (((uint16(c.B) >> 3) & 31) << 10))
```
aykevl: The colors here are `uint8`, so after you shift right by 3 the `& 31` is unnecessary. It will probably be optimized away anyway, but the code would perhaps be cleaner by leaving the mask out.

Not a big issue, only worth fixing if you think it's worth it.

aykevl (commented):
Yeah, this looks like a bug to me. I wrote the code originally so you can blame me ;)

I'm fine with merging, have you tested this in an emulator or on an actual GBA?

src/machine/machine_gameboyadvance.go
Agurato: Indeed the bit mask is useless here, I just modified it in following commit.

I tested the code with both emulators [mGBA](https://mgba.io/) and [VisualBoyAdvance-M](https://vba-m.com/). I don't have any physical GBA to test it further.

deadprogram:
Thanks @Agurato for the quick change.

Would you mind squashing into a single commit, or would you prefer that I just do so when merging?

Agurato:
@deadprogram I squashed both commits into a single one. Thanks for the quick review !

deadprogram:
Thanks @Agurato now merging.

Issue #1785, opened 2021-04-10
Support SPI on STM32 L4 series


kenbell:
Some SVD patches are needed to get the same symbolic constants available as other platforms.  Have raised https://github.com/stm32-rs/stm32-rs/pull/535

Once SVD patches are merged, I'll update the stm32-svd repo

kenbell:
PR to merge to stm32-svd repo: https://github.com/tinygo-org/stm32-svd/pull/3

deadprogram:
This was released with v0.18 so now closing. Thank you!

PR #1789, opened 2021-04-11
stm32: support SPI on L4 series

kenbell:
Implements: #1785 
Depends on: https://github.com/tinygo-org/stm32-svd/pull/3

Includes fix from @ofauchon / @aykevl to do 8-bit writes to the DR register, even if declared as 16-bit register.

The `getBaudRate` function is not great - cut and paste from other boards with clock-specific magic values.  Will raise a separate PR to have a single implementation of `getBaudRate` for STM32 which accepts the peripheral bus freq as input.

### DIFF
--- a/src/machine/board_nucleol432kc.go
+++ b/src/machine/board_nucleol432kc.go
@@ -13,20 +13,53 @@ const (
 	LED_GREEN   = PB3
 )
 
-// UART pins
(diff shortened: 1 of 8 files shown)


deadprogram:
Merging, thanks for adding this support @kenbell

PR #1797, opened 2021-04-14
Clean up target files

aykevl:
This is a PR with two mostly unrelated commits that I've put together because one led to the other and they will conflict. See the individual commit messages for details. In summary, this cleans up the target JSON files with redundant lines.

### DIFF
--- a/targets/avr.json
+++ b/targets/avr.json
@@ -9,7 +9,8 @@
 	"scheduler": "none",
 	"default-stack-size": 256,
 	"cflags": [
-		"--target=avr-unknown-unknown"
+		"--target=avr-unknown-unknown",
+		"-Werror"
 	],
 	"ldflags": [
 		"-T", "targets/avr.ld",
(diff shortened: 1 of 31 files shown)


targets/avr.json
```diff
@@ -9,7 +9,8 @@
 	"scheduler": "none",
 	"default-stack-size": 256,
 	"cflags": [
-		"--target=avr-unknown-unknown"
+		"--target=avr-unknown-unknown",
+		"-Werror"
```
aykevl, line 13: I've added this line because I've seen warnings recently regarding unused parameters in AVR builds that weren't caught before. This commit fixes these warnings (by adding `-Qunused-arguments`) and with this flag also ensures that the compilation will fail if this happens again.

deadprogram: :+1:

deadprogram:
:scissors: merging. Thanks @aykevl

PR #1799, opened 2021-04-14
microbit: remove LED constant

aykevl:
There doesn't appear to be a user-controllable LED outside of the LED
matrix. In fact, the pin assigned for this was P13, which was connected
to the SPI SCK pin.


deadprogram:
:scissors: merging.

PR #1800, opened 2021-04-14
microbit-v2: add support for S113 SoftDevice

aykevl:
This currently doesn't work with `tinygo flash` yet (even with `-programmer=openocd`), you can use pyocd instead. For example, from the Bluetooth package:

    tinygo build -o test.hex -target=microbit-v2-s113v7 ./examples/advertisement/
    pyocd flash --target=nrf52 test.hex

I intend to add support for pyocd to work around this issue, so that a simple `tinygo flash` suffices.

If the micro:bit doesn't have the SoftDevice flashed on it, you can restore it by downloading the following file onto it: https://support.microbit.org/support/solutions/articles/19000021613-reset-the-micro-bit-to-factory-defaults

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -226,8 +226,6 @@ smoketest:
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=microbit            examples/microbit-blink
 	@$(MD5SUM) test.hex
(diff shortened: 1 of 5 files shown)


deadprogram:
:satellite: now merging, thanks @aykevl

PR #1802, opened 2021-04-14
nrf52833: add PWM support

aykevl:
This chip wasn't included in the PR for PWM support. Adding this support is very easy, luckily.


deadprogram:
It is indeed short. Now merging, thanks @aykevl

PR #1812, opened 2021-04-16
Implement environment variables and command line arguments

aykevl:
This PR:

 * Implements environment variables on Linux (MacOS and WASI were already supported)
 * Implements command line arguments on Linux, MacOS and WASI.

While these two changes are more or less separate, implementing the first made it a lot cleaner to write tests for the second.


aykevl:
Hmmm, maybe I should split out the wasi-libc change.

aykevl:
I solved it a little bit differently. I like this way, it's a bit more flexible.

deadprogram:
This is rather exciting feature, I think this now makes it possible to create useful CLI tools, writ small.

Merging, thanks for working on this one @aykevl

PR #1813, opened 2021-04-16
WIP: Implement Benchmark()

dkegel-fastly:
"make test" fails locally because it can't find string.h, presumably I need to teach it how to find include files on mac 10.15.

WIP because "cd tests/tinygotest; tinygo test" fails here with 
```
error: package github.com/tinygo-org/tinygo/tests/tinygotest.test imports main but couldn't find dependency
```

Creating a draft pull request to see what CI says about the first problem.


dkegel-fastly:
oh, bother.  didn't tell it dev.

PR #1815, opened 2021-04-16
Fix formatting of machine_atmega1280.go

dkegel-fastly:
Happy fmt-check, happy ci.


dkegel-fastly:
The download failure appears to be spurious...?

aykevl:
Thank you! This looks reasonable.

deadprogram:
Oops, I should have checked more carefully thanks for the fix!

Issue #1816, opened 2021-04-17
Cannot indicate which ST-LINK to use when multiple plugged in


kenbell:
Currently, if multiple ST-LINK are plugged in to development box, it's not possible to indicate which one to use for flashing a target.  This makes it cumbersome to develop for & test multiple targets.

It is possible to select the desired ST-LINK by using the `hla_serial` openocd command, eg:
```
hla_serial 066CFF485648756687031446
```

This can be specified in the target configuration today, but that binds the serial number of a ST-LINK programmer into the target, which is undesirable.

To avoid hardcoding a specific way to identify programming devices, propose to add a new tinygo command-line option `programmer-cmd` (there is an existing command-line option `programmer` which takes values like `openocd`).  This would work like this:
```shell
tinygo gdb -target stm32f4disco-1 -programmer-cmd "hla_serial 066CFF485648756687031446" ./examples/ina228
```

The new `programmer-cmd` parameter would take one (or more) commands to be executed by the programmer software (eg openocd).  Commands can be separated by commas.  There would be no particular restrictions on the commands that can be supplied.

deadprogram:
This was released with v0.18 so now closing. Thank you!

PR #1492, opened 2020-11-09
Add support for additional openocd commands

supcik:
This PR addresses the issue #1279. Your comments are warmly welcome.

PR #1817, opened 2021-04-17
openocd commands in tinygo command line

kenbell:
Implements #1816 to enable identification of specific ST-LINK devices on tinygo command line.

Because the precise command is specific to the openocd transport in use, this implements as a generic 'openocd command' facility.

### DIFF
--- a/builder/config.go
+++ b/builder/config.go
@@ -17,10 +17,18 @@ func NewConfig(options *compileopts.Options) (*compileopts.Config, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	if options.OpenOCDCommands != nil {
+		// Override the OpenOCDCommands from the target spec if specified on
+		// the command-line
+		spec.OpenOCDCommands = options.OpenOCDCommands
+	}
+
 	goroot := goenv.Get("GOROOT")
 	if goroot == "" {
 		return nil, errors.New("cannot locate $GOROOT, please set it manually")
 	}
+
 	major, minor, err := goenv.GetGorootVersion(goroot)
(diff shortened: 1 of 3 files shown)


kenbell:
@aykevl @deadprogram - is this something that can be merged?  It's super helpful for me with multiple STM32 boards I want to verify changes on, so that openocd will target the correct ST-link to flash the board.

sago35:
It does almost the same thing as #1492, but the difference is that it can be run from the command line.

If possible, it might be better to merge with OpenOCDCommands so that it can be configured from JSON, but can be overridden from the command line.

note:
Currently, OpenOCDCommands are only used by bluepill-clone.
The only thing we are implementing is the CPUTAPID setting.

deadprogram:
@sago35 that seems like a very good suggestion.

@kenbell what do you think?

kenbell:
Yeah - seems like a good idea, so:

1. Change the name to something like `-openocdcommands`
2. Change the logic so that if `-openocdcommands` is provided on command-line, the `OpenOCDCommands` json (if any) will be overwritten by the command-line.

I'll update the PR

kenbell:
@sago35 - I've pushed a new change.  Is this the kind of thing you were thinking?   An example of using is this:
```bash
tinygo flash -target bluepill -ocd-commands "hla_serial 49FF73067077485240271967" -size short ./blinky/
```

sago35 (approved):
Looks good to me.

deadprogram:
Thanks @kenbell and @sago35 for working on this.

Now merging.

Issue #1796, opened 2021-04-14
Speed up llvm-source step and llvm-build step in ci


sago35:
note: If I make a Pull-Request, a cache will probably be created in circleci, so I'll write it in the issue first.

I have improved the following two points.
I think it will speed up the CI in most cases.

However, since we are using a cache-reducing method, the cache may need to be updated in future dev versions.
Typically, this happens when you try to use a file that is not currently in use.
I would especially like to know if you have any opinions on this disadvantage.

* llvm-source-*
    * Change the cache to remove all files except *.h and *.def under the llvm-project directory.
    * This change will reduce the restore-cache on linux from 21 sec to 1 sec
* llvm-build-*
    * Change the cache to remove *.o
    * This change will reduce the restore-cache on linux from 11 sec to 7 sec

The changes can be found below.
For testing purposes, the CI targets have been narrowed down to build-linux, build-macos and assert-test-linux, but will eventually be restored.
I will also rename the cache to an appropriate name.

* changes
    * https://github.com/tinygo-org/tinygo/compare/dev...sago35:ci-llvm-build
* CI result
    * old
        * https://app.circleci.com/pipelines/github/tinygo-org/tinygo/4727/workflows/20480dd8-d147-4818-bd0c-0a7b598e44a2
    * new
        * https://app.circleci.com/pipelines/github/sago35/tinygo/38/workflows/a4108973-ce2e-4a34-930c-90031977eede

![image](https://user-images.githubusercontent.com/9251039/114635347-819bd880-9cff-11eb-8ad1-71cd3c4f1e18.png)

aykevl:
> * llvm-source-*
> 
>     * Change the cache to remove all files except *.h and *.def under the llvm-project directory.
> 
>     * This change will reduce the restore-cache on linux from 21 sec to 1 sec

Yes, this is a good idea. I wanted to make this change but haven't done it yet. However, I think it is better to cache specific directories instead of specific file extensions. For example, the directory that is currently needed is `llvm-project/clang/lib/Headers`. Like this:

```yaml
       - save_cache:
          key: llvm-source-11-v1
          paths:
            - llvm-project/clang/lib/Headers
```




> * llvm-build-*
> 
>     * Change the cache to remove *.o
> 
>     * This change will reduce the restore-cache on linux from 11 sec to 7 sec

[...]

PR #1819, opened 2021-04-18
ci: improve llvm-source cache

sago35:
This PR makes restoring-cache-llvm-source faster.
As described in #1796, the processing time is reduced from 21 sec to 1 sec.

Instead, if the make llvm-build is needed, it needs to be downloaded again with make llvm-source.
However, the only case where this will increase processing time is when doing llvm-build, so only the first time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,14 +68,17 @@ commands:
     steps:
       - restore_cache:
           keys:
-            - llvm-source-11-v1
+            - llvm-source-11-v2
       - run:
           name: "Fetch LLVM source"
           command: make llvm-source
       - save_cache:
-          key: llvm-source-11-v1
+          key: llvm-source-11-v2
           paths:
-            - llvm-project
+            - llvm-project/clang/lib/Headers
+            - llvm-project/clang/include
+            - llvm-project/lld/include
+            - llvm-project/llvm/include
   build-wasi-libc:
     steps:
       - restore_cache:
@@ -160,6 +163,9 @@ commands:
           command: |
             if [ ! -f llvm-build/lib/liblldELF.a ]
             then
+              # fetch LLVM source
+              rm -rf llvm-project
+              make llvm-source
               # install dependencies
               sudo apt-get install cmake ninja-build
               # hack ninja to use less jobs
@@ -224,6 +230,9 @@ commands:
           command: |
             if [ ! -f llvm-build/lib/liblldELF.a ]
             then
+              # fetch LLVM source
+              rm -rf llvm-project
+              make llvm-source
               # install dependencies
               sudo apt-get install cmake ninja-build
               # hack ninja to use less jobs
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -287,14 +300,21 @@ commands:
...
-            - llvm-source-11-macos-v1
+            - llvm-source-11-macos-v2
       - run:
           name: "Fetch LLVM source"
-          command: make llvm-source
+          command: |
+            if [ ! -e llvm-project ]
+            then
+              make llvm-source
+            fi
```
aykevl: This change is unnecessary: `make llvm-source` will exit immediately if the llvm-project directory exists.

.circleci/config.yml
```diff
@@ -68,14 +68,21 @@ commands:
...
-            - llvm-source-11-v1
+            - llvm-source-11-v2
       - run:
           name: "Fetch LLVM source"
-          command: make llvm-source
+          command: |
+            if [ ! -e llvm-project ]
+            then
+              make llvm-source
+            fi
```
aykevl: Same here.

aykevl (commented):
Looks good to me apart from the two issues below.

.circleci/config.yml
sago35: You are correct.
I fixed it and force-pushed it.

.circleci/config.yml
sago35: You are correct.
I fixed it and force-pushed it.

sago35:
```
#!/bin/bash -eo pipefail
make llvm-source

git clone -b xtensa_release_11.0.0 --depth=1 https://github.com/tinygo-org/llvm-project llvm-project
fatal: destination path 'llvm-project' already exists and is not an empty directory.
Makefile:156: recipe for target 'llvm-project/README.md' failed
make: *** [llvm-project/README.md] Error 128

Exited with code exit status 2

CircleCI received exit code 2
```
I'm getting an error when I try to run git-clone again because of the dependency on llvm-source/README.md.
It might be easier to modify the Makefile, but do you have a better idea?
A simple fix might be as follows

Makefile:
```diff
 # Get LLVM sources.
 $(LLVM_PROJECTDIR)/README.md:
 	git clone -b xtensa_release_11.0.0 --depth=1 https://github.com/tinygo-org/llvm-project $(LLVM_PROJECTDIR)
- llvm-source: $(LLVM_PROJECTDIR)/README.md
+ llvm-source: $(LLVM_PROJECTDIR)
```

aykevl:
You could do that, or use one of the cached directories like this:

```diff
 # Get LLVM sources.
-$(LLVM_PROJECTDIR)/README.md:
+$(LLVM_PROJECTDIR)/llvm:
 	git clone -b xtensa_release_11.0.0 --depth=1 https://github.com/tinygo-org/llvm-project $(LLVM_PROJECTDIR)
- llvm-source: $(LLVM_PROJECTDIR)/README.md
+ llvm-source: $(LLVM_PROJECTDIR)/llvm
```

aykevl (approved):
Looks good to me, thank you for speeding up the CI!

PR #1821, opened 2021-04-18
main: add -print-allocs flag that lets you print all allocations

aykevl:
This flag, if set, is a regexp for function names. If there are heap
allocations in the matching function names, these heap allocations will
be printed with an explanation why the heap allocation exists (and why
the object can't be stack allocated).

TODO: write tests.

### DIFF
--- a/compileopts/options.go
+++ b/compileopts/options.go
@@ -2,6 +2,7 @@ package compileopts
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 )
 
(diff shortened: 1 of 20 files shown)


FrankReh:
Flag in main.go is still -print-allocs, do you want -print-heap-allocs?

aykevl:
I've added some tests written in Go, that not only test the stack alloc transform but also show what escapes and what doesn't.

I'm quite happy with how the test looks, it's much more readable than many other tests.

deadprogram:
OK this is one incredibly useful feature for those who aspire to a very small memory footprint.

Now merging, great work @aykevl

PR #1823, opened 2021-04-19
machine: avoid heap allocations in USB code

aykevl:
This commit replaces most heap allocations in USB related code with stack allocations. This is important for several reasons:

  - It avoids running the GC unnecessarily.
  - It reduces code size by 400-464 bytes.
  - USB code might be called from interrupt handlers. The heap may be in an inconsistent state when that happens if main thread code also performs heap allocations.

The last one is by far the most important one: not doing heap allocations in interrupts is critical for correctness. But the code size reduction alone should be worth it.

There are two heap allocations in USB related code left: in the function `receiveUSBControlPacket` (SAMD21 and SAMD51). This heap allocation must also be removed because it runs in an interrupt, but I've left that for a future change.

---

This PR is made easy by the change in #1821, which provides a convenient way of showing all heap allocations in the code or in a single package (in this case, the machine package).
Eventually this USB code will likely be replaced by the new USB stack that's being written by @ardnew but it's a relatively simple improvement until that work is complete.

@deadprogram can you maybe take a look at `receiveUSBControlPacket`? I see that the callers only use the first 7 bytes, so it should be relatively simple to replace with a `[7]byte` but I'm not sure what to do with `bytesread`.

### DIFF
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -2298,7 +2298,7 @@ func handleStandardSetup(setup usbSetup) bool {
 func cdcSetup(setup usbSetup) bool {
 	if setup.bmRequestType == usb_REQUEST_DEVICETOHOST_CLASS_INTERFACE {
 		if setup.bRequest == usb_CDC_GET_LINE_CODING {
-			b := make([]byte, 7)
+			var b [cdcLineInfoSize]byte
 			b[0] = byte(usbLineInfo.dwDTERate)
 			b[1] = byte(usbLineInfo.dwDTERate >> 8)
 			b[2] = byte(usbLineInfo.dwDTERate >> 16)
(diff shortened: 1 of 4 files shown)


deadprogram:
Yes, the `SET_LINE_CODING` packet is the same as that for `GET_LINE_CODING` see https://github.com/tinygo-org/tinygo/blob/dev/src/machine/machine_atsamd21.go#L2300-L2308

Also see https://cscott.net/usb_dev/data/devclass/usbcdc11.pdf section 6.2.13

deadprogram:
@aykevl do you want me to make a PR with that change, or do you want to add a commit to this one, or something else entirely?

deadprogram:
@aykevl please see https://github.com/tinygo-org/tinygo/commit/53b0a9f5f2afe4b38e605dc02785668573b496e0 which I believe should handle the remaining changes you requested. If you like it can git cherry-pick into this branch then we can merge.

deadprogram:
Reminder to @aykevl please let me know what you think about my proposed changes in https://github.com/tinygo-org/tinygo/commit/53b0a9f5f2afe4b38e605dc02785668573b496e0

aykevl:
I've added a few inline comments to the commit.

deadprogram:
OK 52699a81267b8471b1a17d868da94f41f09ddaac created to correct typo. I left the other changes as they were because I think it is perhaps a little clearer what is happening. What do you think?

deadprogram (approved):

aykevl:
@deadprogram I've cherry picked your commit to the branch, so now this PR should be complete.

deadprogram:
Thanks @aykevl now merging.

Issue #1805, opened 2021-04-15
regression at 312f5d3 "builder: run interp per package": "panic: interface conversion"?


dkegel-fastly:
An app here built ok with tinygo 0.17 (either from release branch using go install or using "brew install tinygo").

The same app now fails with tinygo built from dev branch ( v0.17.0-99-g0d66475 ) using go install:
```
$ tinygo build -target=wasi -scheduler=none -wasm-abi=generic -o bin/main.wasm
panic: interface conversion: interp.value is interp.rawValue, not *interp.mapValue

goroutine 52 [running]:
github.com/tinygo-org/tinygo/interp.(*runner).run(0xc00447e1b0, 0xc003d98be0, 0xc000929190, 0x6, 0x7, 0xc00073d8c0, 0xc0053554c8, 0x8, 0x0, 0x4257d65, ...)
	/Users/dkegel/go/src/github.com/tinygo-org/tinygo/interp/interpreter.go:467 +0xc695
github.com/tinygo-org/tinygo/interp.(*runner).run(0xc00447e1b0, 0xc003d98b90, 0x0, 0x0, 0x0, 0x0, 0x43f4c7b, 0x4, 0x0, 0x0, ...)
	/Users/dkegel/go/src/github.com/tinygo-org/tinygo/interp/interpreter.go:522 +0x5ed3
github.com/tinygo-org/tinygo/interp.Run(0x3a7c4c00, 0xc005a7bd00, 0x0, 0x0)
	/Users/dkegel/go/src/github.com/tinygo-org/tinygo/interp/interp.go:112 +0x69f
github.com/tinygo-org/tinygo/builder.optimizeProgram(0x3a7c4c00, 0xc0002721e0, 0xc0051c4780, 0x2)
	/Users/dkegel/go/src/github.com/tinygo-org/tinygo/builder/build.go:649 +0x3f
github.com/tinygo-org/tinygo/builder.Build.func2(0xc00414bf20, 0x0, 0x0)
	/Users/dkegel/go/src/github.com/tinygo-org/tinygo/builder/build.go:384 +0x7c6
github.com/tinygo-org/tinygo/builder.jobWorker(0xc0012b5260, 0xc0012b5200)
	/Users/dkegel/go/src/github.com/tinygo-org/tinygo/builder/jobs.go:165 +0xe2
created by github.com/tinygo-org/tinygo/builder.runJobs
[...]

PR #1827, opened 2021-04-20
Fix regression in maps after running interp per package

aykevl:
The bug https://github.com/tinygo-org/tinygo/issues/1805 was introduced in #1758. This PR fixes this issue by removing the affected code entirely. This is why the bugfix doesn't come with a test: I'm not sure how I would test code that isn't there anymore.

To make that work, I had to fix a bug in the interp package that I found while working on this change.

I expected a small reduction in code size, but that didn't happen. The code in testdata/map.go is affected, but it's only reduced in size. I haven't investigated why. All other code (even rather extensive tests such as testdata/json.go) are not affected, which is great. This shows that the special map support in the interp package was never really needed.


dkegel-fastly:
Seems to solve the problem here.

aykevl:
Apparently I did not test enough, the testdata/map.go test is failing. I will need to investigate this further.

aykevl:
Looks like I fixed it (see commit message for details).

deadprogram:
Great work @aykevl! Now merging.

:world_map:

PR #1833, opened 2021-04-23
Copy os.DirFS to the os package

arp242:
This wasn't implemented; this is just an exact copy of the code from the
Go standard library.

### DIFF
--- a/src/os/file_go_116.go
+++ b/src/os/file_go_116.go
@@ -5,6 +5,7 @@ package os
 import (
 	"io"
 	"io/fs"
+	"runtime"
 )
 
 type (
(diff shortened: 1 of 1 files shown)


deadprogram:
Anyone have any feedback on this?

DrGo:
yes, needed to use the new FS interface.

deadprogram:
@DrGo sorry I cannot understand what you mean. It is not using an interface you think it should be?

arp242:
I think what DrGo intended is that you need this patch to use the new `fs` interface introduces in Go 1.16. At least, that's how I read it.

I actually ran in to some more problems with missing functions after this. I've been planning to look in to this to do this better; TinyGo can already "merge" Go stdlib packages with its own versions, but it's a bit limited. Instead of manually copying stuff like this, it's probably a better idea to expand that so it's automatically "merged" from stdlib.

This is mostly useful for the wasm use case, because a lot of these things aren't used by microcontrollers.

DrGo:
yes... since Go 1.16, I have been using fs.FS wherever access to a filesystem is needed and then inject this implementation when access to the underlying OS is needed. It removes the dependency on a specific filesystem and makes testing a breeze (eg using a mock in-memory filesystem). I cannot say much about how high of a priority this change is or about the best way to implement it (copy-paste or automatic merge).

sago35:
Please rebase to the latest dev branch.

It is copied from `src/os/file.go` in go1.16.
If it is not used, it has no effect, so it seems to be safe to merge it.

To make it easier to see the differences, I would like the order of implementation to be moved to before ReadFile().

dkegel-fastly:
I will have a look at this...

dkegel-fastly:
Maybe after readdir lands, then I can pull in tests, too.

dkegel-fastly:
os.DirFS has landed in dev.   I can't quite land the tests yet, but they were useful.

PR #1835, opened 2021-04-23
atsamd51 atsame5x: change to allow specifying the ADC Bus number

sago35:
For PB08 and PB09, and PA08 and PA09, it was not possible to select which ADC instance to use until now.
Since the two ADC instances can be sampled at the same time, we want to be able to select which ADC instance to use.

> https://ww1.microchip.com/downloads/en/DeviceDoc/SAM_D5x_E5x_Family_Data_Sheet_DS60001507G.pdf
> The SAM D5x/E5x has two ADC instances, ADC0 and ADC1.
> The two inputs can be sampled simultaneously, as each ADC includes sample and hold circuits.

|    | ADC0 | ADC1 |
| ---- | - | - |
| PB03 |  15 |  |
| PC00 |     |  10 |
| PC01 |     |  11 |
| PC02 |     |   4 |
| PC03 |     |   5 |
| PA02 |   0 |  |
| PA03 |   1 |  |
| PB04 |     |   6 |
| PB05 |     |   7 |
| PD00 |     |  14 |
| PD01 |     |  15 |
| PB06 |     |   8 |
| PB07 |     |   9 |
| PB08 |   2 |   0 |
| PB09 |   3 |   1 |
| PA04 |   4 |  |
| PA05 |   5 |  |
| PA06 |   6 |  |
| PA07 |   7 |  |
| PA08 |   8 |   2 |
| PA09 |   9 |   3 |
| PA10 |  10 |  |
| PA11 |  11 |  |
| PC30 |     |  12 |
| PC31 |     |  13 |
| PB00 |  12 |  |
| PB01 |  13 |  |
| PB02 |  14 |  |

I tried to change the source code, but there may be a better way.
Also, as I was making changes in src/examples/adc, I ended up with incompatible changes in environments other than atsamx5x.
Would it be better to make sure that all environments except atsamx5x are not incompatible?

### DIFF
--- a/src/machine/machine.go
+++ b/src/machine/machine.go
@@ -44,4 +44,5 @@ func (p Pin) Low() {
 
 type ADC struct {
 	Pin Pin
+	Bus uint8
 }
--- a/src/examples/adc/adc.go
+++ b/src/examples/adc/adc.go
@@ -14,7 +14,7 @@ func main() {
 	led := machine.LED
 	led.Configure(machine.PinConfig{Mode: machine.PinOutput})
 
-	sensor := machine.ADC{machine.ADC2}
+	sensor := machine.ADC{Pin: machine.ADC2}
 	sensor.Configure(machine.ADCConfig{})
 
 	for {
--- a/src/machine/adc.go
+++ b/src/machine/adc.go
@@ -9,4 +9,11 @@ type ADCConfig struct {
 	Reference  uint32 // analog reference voltage (AREF) in millivolts
 	Resolution uint32 // number of bits for a single conversion (e.g., 8, 10, 12)
 	Samples    uint32 // number of samples for a single conversion (e.g., 4, 8, 16, 32)
+	Bus        uint8  // bus Number of ADC
 }
+
+const (
+	AdcBusAuto = 0
+	AdcBus0    = 1
+	AdcBus1    = 2
+)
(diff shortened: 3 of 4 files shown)


src/machine/machine.go
```diff
@@ -39,4 +39,5 @@ func (p Pin) Low() {
 
 type ADC struct {
 	Pin Pin
+	Bus uint8
```
sago35, line 47: Due to this change, src/examples/adc needs to be modified

sago35:
This is not directly related to this PR, but I noticed that the following sources can be unified.
I'll discuss it in another PR / Issue.

* machine_atsamd51.go
* machine_atsame5x.go

sago35:
> 
> 
> This is not directly related to this PR, but I noticed that the following sources can be unified.
> I'll discuss it in another PR / Issue.
> 
>     * machine_atsamd51.go
> 
>     * machine_atsame5x.go

Fixed with #1845

deadprogram:
I think there is probably a different way you can solve this @sago35 that does not require breaking the current API.

For example, your function `getADCBus()` could lookup which bus matches that pin. In the case that you really need to set which bus to use, then perhaps adding the bus number to https://github.com/tinygo-org/tinygo/blob/dev/src/machine/adc.go#L8

What do you think?

sago35:
The original version of getADCBus() was implemented in such a way that if there were two candidates, it would return one of them.
When a new getADCBus() decides which bus to use, there are two rules to follow.

If there is only one bus that the Pin can use, it will return that bus.
If there are two busses available for the Pin, the bus is determined from the information stored in the ADC-struct.

Hmm, I can't think of a good way to do this.

sago35:
The dev branch is in progress, so I rebased it.

PR #1838, opened 2021-04-24
compiler: improve position information

aykevl:
In many cases, position information is not stored in Go SSA instructions
because they don't exit directly in the source code. This includes
implicit type conversions, implicit returns at the end of a function,
the creation of a (hidden) slice when calling a variadic function, and
many other cases. I'm not sure where this information is supposed to
come from, but this patch takes the value (usually) from the value the
instruction refers to. This seems to work well for these implicit
conversions.

I've also added a few extra tests to the heap-to-stack transform pass,
of which one requires this improved position information.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -964,11 +964,59 @@ func (b *builder) createFunction() {
 	}
 }
 
+// posser is an interface that's implemented by both ssa.Value and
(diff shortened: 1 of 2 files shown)


deadprogram:
:world_map: thanks @aykevl now merging.

PR #1840, opened 2021-04-26
transform: split interface and reflect lowering

aykevl:
These two passes are related, but can definitely work independently.
Which is what this change does: it splits the two passes. This should
make it easier to change these two new passes in the future.

This change now also enables slightly better testing by testing these
two passes independently. In particular, the reflect lowering pass got
some actual tests: it was barely unit-tested before.

I have verified that this doesn't really change code size, at least not
on the microbit target. Two tests do change, but in a very minor way
(and in opposite direction).

### DIFF
--- a/transform/allocs_test.go
+++ b/transform/allocs_test.go
@@ -2,7 +2,6 @@ package transform_test
 
 import (
 	"go/token"
-	"go/types"
 	"io/ioutil"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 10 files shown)


aykevl:
Huh, I did not expect _that_ error. I'll need to investigate why these method sets weren't properly cleaned.

aykevl:
Fixed the error, this is now ready for review.

deadprogram:
:arrow_lower_right: merging. Thanks @aykevl

PR #1842, opened 2021-04-27
Target for Arduino Nano w/ new bootloader

raqbit:
Since January 2018, Arduino Nanos (and clones) are sold with a new bootloader (see [note](https://www.arduino.cc/en/Guide/ArduinoNano#:~:text=note)), which requires programming at 115200 baud instead of the 57600 baud required by the old one.

This PR adds a target for these new Nanos which simply inherits the "old" target but with a changed baud rate in the flash command.


deadprogram:
Thanks @Raqbit for the updated target file. Now merging.

PR #1844, opened 2021-04-28
Add -llvm-features parameter

fgsch:
With this is possible to enable e.g., SIMD in WASM using `-llvm-features +simd128`.  
Multiple features can be specified separated by comma, e.g.,` -llvm-features +simd128,+tail-call`.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -59,6 +59,7 @@ type Config struct {
 	DefaultStackSize   uint64
 	NeedsStackObjects  bool
 	Debug              bool // Whether to emit debug information in the LLVM module.
+	LLVMFeatures       string
 }
 
 // compilerContext contains function-independent data that should still be
@@ -185,7 +186,12 @@ func NewTargetMachine(config *Config) (llvm.TargetMachine, error) {
 	if err != nil {
 		return llvm.TargetMachine{}, err
 	}
-	features := strings.Join(config.Features, ",")
+
+	feat := config.Features
+	if len(config.LLVMFeatures) > 0 {
+		feat = append(feat, config.LLVMFeatures)
+	}
+	features := strings.Join(feat, ",")
 
 	var codeModel llvm.CodeModel
 	var relocationModel llvm.RelocMode
--- a/builder/build.go
+++ b/builder/build.go
@@ -94,6 +94,7 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 		DefaultStackSize:   config.Target.DefaultStackSize,
 		NeedsStackObjects:  config.NeedsStackObjects(),
 		Debug:              config.Debug(),
+		LLVMFeatures:       config.LLVMFeatures(),
 	}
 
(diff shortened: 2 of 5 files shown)


compiler/compiler.go
```diff
@@ -185,11 +186,17 @@ func NewTargetMachine(config *Config) (llvm.TargetMachine, error) {
...
-	features := strings.Join(config.Features, ",")
 
+	var features string
 	var codeModel llvm.CodeModel
 	var relocationModel llvm.RelocMode
 
+	if strings.HasPrefix(config.Triple, "wasm") {
+		features = config.WasmFeatures
+	} else {
+		features = strings.Join(config.Features, ",")
```
fgsch: This asymmetry bothers me a bit. I was thinking about whether we can move the join call when we retrieve the features but decided to keep it in case people didn't like it.

deadprogram: Did you really mean to disable the "normal" features if `wasm-features` is being used? Seems like there is probably a case where you need both. Perhaps this is enough here (untested):

```go
features := strings.Join(config.Features, ",")
if strings.HasPrefix(config.Triple, "wasm") && len(config.WasmFeatures) > 0 {
        features = strings.Join([]string{features, config.WasmFeatures}, ",")
}
```

deadprogram: Actually the above has flaw. This code here is what I *really* meant :smile_cat: 

```go
feat := config.Features
if strings.HasPrefix(config.Triple, "wasm") && len(config.WasmFeatures) > 0 {
        feat = append(feat, config.WasmFeatures)
}
features := strings.Join(feat, ",")
```

fgsch: @deadprogram Yes, that was on purpose.  I can change the code so both are taking into account but I don't think is needed at the moment and would be easy to change it later.

fgsch: As discussed offline, I've added your suggestion. Thank you.

aykevl:
These features aren't really specific to WebAssembly (except for one arbitrary test). These are for the LLVM backend.

Maybe a better name for the flag would be `-llvm-features` to make it clear what the flag is for? Then you could also rename the various fields from `WasmFeatures` to just `Features` and drop the wasm test.

fgsch:
I originally proposed `-target-features` that was target agnostic, but based on our convo in the Tinygo slack channel, I thought we didn't want to expose the features to all platform - only to wasm. 
Are we now saying that we want to make these general target features?

deadprogram:
> Are we now saying that we want to make these general target features?

"`llvm-features`" apparently yes. :smile_cat: 

@fgsch could you please squash into a single commit?

@aykevl any further feedback?

deadprogram:
Also, @fgsch can you please rebase against `dev` and resolve any merge conflicts? Thank you.

fgsch:
Sure, I will do in a bit.

fgsch:
Squashed and rebased.

deadprogram:
Thanks for working on this @fgsch now merging.

PR #1846, opened 2021-04-29
Support chained interrupt handlers

kenbell:
Multiple calls to interrupt.New are permitted with handlers called sequentially in undefined order.

The implementation works by changing the interrupt 'wrapper' from always being a single call to the actual handler, to being multiple calls to all the handlers in the chain in sequence.

Most of the change is re-arranging the interrupt lowering code such that it's possible to loop over multiple handlers with the same interrupt number when generating the wrapper function.

By way of example, this:
```golang
package main

import (
	"runtime/interrupt"
)

func handler1(interrupt.Interrupt) {
	println("h1")
}

func handler2(interrupt.Interrupt) {
	println("h2")
}

func main() {
	intr1 := interrupt.New(22, handler1)
	intr1.Enable()

	intr2 := interrupt.New(22, handler2)
	intr2.Enable()
}
```

The interrupt handler compiles to this at default optimization level (CAN1_SCE_IRQHandler is interrupt 22 on this device):
```asm
.text:08000d60 <CAN1_SCE_IRQHandler>:
    .text:08000d60 80 b5                            push	{r7, lr}
    .text:08000d62 41 f2 4a 00                      movw	r0, #4170	; 0x104a
    .text:08000d66 02 21                            movs	r1, #2
    .text:08000d68 c0 f6 00 00                      movt	r0, #2048	; 0x800
    .text:08000d6c ff f7 92 fa                      bl	0x08000294 <runtime.printstring>
    .text:08000d70 ff f7 9c fa                      bl	0x080002ac <runtime.printnl>
    .text:08000d74 41 f2 4c 00                      movw	r0, #4172	; 0x104c
    .text:08000d78 02 21                            movs	r1, #2
    .text:08000d7a c0 f6 00 00                      movt	r0, #2048	; 0x800
    .text:08000d7e ff f7 89 fa                      bl	0x08000294 <runtime.printstring>
    .text:08000d82 bd e8 80 40                      ldmia.w	sp!, {r7, lr}
    .text:08000d86 ff f7 91 ba                      b.w	0x080002ac <runtime.printnl>
```
Note the two handlers have both been successfully inlined into the wrapper.  This implementation has minimal or no additional overhead in the output binary (depending on efficient optimization).

### DIFF
--- a/transform/interrupt.go
+++ b/transform/interrupt.go
@@ -73,10 +73,10 @@ func LowerInterrupts(mod llvm.Module, sizeLevel int) []error {
 	// Create a function type with the signature of an interrupt handler.
 	fnType := llvm.FunctionType(ctx.VoidType(), nil, false)
 
-	// Collect a slice of interrupt handle objects. The fact that they still
+	// Collect a map of interrupt handle objects. The fact that they still
 	// exist in the IR indicates that they could not be optimized away,
 	// therefore we need to make real interrupt handlers for them.
-	var handlers []llvm.Value
+	handleMap := map[int][]llvm.Value{}
 	handleType := mod.GetTypeByName("runtime/interrupt.handle")
 	if !handleType.IsNil() {
 		handlePtrType := llvm.PointerType(handleType, 0)
@@ -84,14 +84,48 @@ func LowerInterrupts(mod llvm.Module, sizeLevel int) []error {
 			if global.Type() != handlePtrType {
 				continue
 			}
-			handlers = append(handlers, global)
+
+			// Get the interrupt number from the initializer
+			initializer := global.Initializer()
+			num := int(llvm.ConstExtractValue(initializer, []uint32{1, 0}).SExtValue())
+			pkg := packageFromInterruptHandle(global)
+
+			handles, exists := handleMap[num]
+
+			// If there is an existing interrupt handler, ensure it is in the same package
+			// as the new one.  This is to prevent any assumptions in code that a single
+			// compiler pass can see all packages to chain interrupt handlers. When packages are
+			// compiled to separate object files, the linker should spot the duplicate symbols
+			// for the wrapper function, failing the build.
+			if exists && packageFromInterruptHandle(handles[0]) != pkg {
+				errs = append(errs, errorAt(global,
+					fmt.Sprintf("handlers for interrupt %d (%s) in multiple packages: %s and %s",
+						num, handlerNames[int64(num)], pkg, packageFromInterruptHandle(handles[0]))))
+				continue
+			}
+
+			handleMap[num] = append(handles, global)
 		}
 	}
 
-	// Iterate over all handler objects, replacing their ptrtoint uses with a
+	// Output interrupts in numerical order for reproducible builds (Go map
+	// intentionally randomizes iteration order of maps).
+	interrupts := make([]int, 0, len(handleMap))
+	for k := range handleMap {
+		interrupts = append(interrupts, k)
+	}
+	sort.Ints(interrupts)
+
+	// Iterate over all handle objects, replacing their ptrtoint uses with a
 	// real interrupt ID and creating an interrupt handler for them.
-	for _, global := range handlers {
-		initializer := global.Initializer()
+	for _, interrupt := range interrupts {
+		handles := handleMap[interrupt]
+
+		// There is always at least one handler for each interrupt number.  We
+		// arbitrarily take the first handler to attach any errors to.
+		first := handles[0]
+
+		initializer := first.Initializer()
 		num := llvm.ConstExtractValue(initializer, []uint32{1, 0})
 		name := handlerNames[num.SExtValue()]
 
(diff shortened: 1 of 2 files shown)


kenbell:
I found there was a problem with the implementation if chained handlers were declared in different modules (linker symbol collision), so I've pushed an addition change to update the name of the 'handles' from `runtime/interrupt.$interruptXX` to `<callingfn>.$interruptXX`, e.g.  `runtime.initTickTimer$interrupt55`

aykevl:
While this sounds good in theory, I have a few concerns:

* > Multiple calls to interrupt.New are permitted with handlers called sequentially _in undefined order_.
  
  This doesn't seem like a good idea. I think they should have a defined order, for example the same order that packages are initialized.
* Some interrupt handlers need to reset an interrupt flag, otherwise the interrupt keeps triggering. While now the responsibility for clearing this flag is well defined (there is only one handler, so that handler is responsible), this PR makes that a bit ambiguous as multiple handlers may be responsible.
* This change assumes the interrupt lowering pass can see the entire program at once.

The last issue is my main concern. While the pass currently already assumes full view of the program, I believe this is something fixable so that this pass can run per package instead of for the whole program. However, this PR bakes in this assumption much harder: I don't see a way to define multiple interrupt handlers per interrupt when package are compiled independently (to object files) without expensive things like global arrays. Eventually I hope to add a non-LTO mode for fast incremental compilations: where packages are compiled to object files and linked like most compilers do. It might even become the default if it works well enough. And for that reason, I would like to keep the option open for compiling interrupt handlers per package without loss of optimizations.

I think doing this manually costs perhaps a bit more code, but would be easier to read and leaves room for future optimizations.

kenbell:
> * > Multiple calls to interrupt.New are permitted with handlers called sequentially _in undefined order_.
>   
>   
>   This doesn't seem like a good idea. I think they should have a defined order, for example the same order that packages are initialized.
> * Some interrupt handlers need to reset an interrupt flag, otherwise the interrupt keeps triggering. While now the responsibility for clearing this flag is well defined (there is only one handler, so that handler is responsible), this PR makes that a bit ambiguous as multiple handlers may be responsible.

The use-case here is that independent devices share an interrupt.  E.g. for STM32F4, there's one interrupt number for both TIM1 roll-over and TIM10: IRQ_TIM1_UP_TIM10.

Personally, I don't think undefined order is a problem.  When using chained interrupts, the code needs to be guarded.  E.g. the handler for TIM10 will be invoked when TIM1 rolls-over, but it should guard its functionality by checking if the interrupt flag in TIM10 peripheral is set.

(TIM10 is uses one interrupt for both counter roll-over and PWM channel roll-over, TIM1 has separate interrupts for counter and channel roll-over)

> I think doing this manually costs perhaps a bit more code, but would be easier to read and leaves room for future optimizations.

I've prototyped it, it's quite ugly and binds together unrelated device implementations - e.g. TIM6 and the DAC implementations need to be a special case for the interrupt (e.g. IRQ_TIM6_DAC).  To have generic code, it also adds overhead (like a lookup table) onto all instances of a device instead of targeting it at the specific devices where a chained interrupt is actually required.

> * This change assumes the interrupt lowering pass can see the entire program at once.
> 
> The last issue is my main concern. While the pass currently already assumes full view of the program, I believe this is something fixable so that this pass can run per package instead of for the whole program. However, this PR bakes in this assumption much harder: I don't see a way to define multiple interrupt handlers per interrupt when package are compiled independently (to object files) without expensive things like global arrays. Eventually I hope to add a non-LTO mode for fast incremental compilations: where packages are compiled to object files and linked like most compilers do. It might even become the default if it works well enough. And for that reason, I would like to keep the option open for compiling interrupt handlers per package without loss of optimizations.

In practice, this could be scoped to just a single package to get almost all the benefit - i.e. the `machine` package, making it compatible with creating object files.  The 'multiple packages' case is more of completeness than being needed in practice.

If we define it to only work for the case where all handlers for a given interrupt are in the same package, would that address this concern?

kenbell:
I've added a change that explicitly limits handlers to being present in the same package to be chained.  It encodes the package into the handle name.

To clarify the use-case here (the original example may be misleading), the code that would rely on chaining is like this:
```golang
func (t *TIM) handleUPInterrupt(interrupt.Interrupt) {
	if t.Device.SR.HasBits(stm32.TIM_SR_UIF) {
		// clear the update flag
		t.Device.SR.ClearBits(stm32.TIM_SR_UIF)

		if t.callback != nil {
			t.callback()
		}
	}
}
```

In this case, the `handleUPInterrupt` for both TIM1 and TIM10 would be called each time `IRQ_TIM1_UP_TIM10` fires.  The peripheral's `TIM_SR_UIF` flag gates the logic.  There's no dependency on ordering of the TIM1 and TIM10 logic.

aykevl:
> I've added a change that explicitly limits handlers to being present in the same package to be chained. It encodes the package into the handle name.

Thank you! Yes, that would resolve my biggest concern.

> In this case, the `handleUPInterrupt` for both TIM1 and TIM10 would be called each time `IRQ_TIM1_UP_TIM10` fires. The peripheral's `TIM_SR_UIF` flag gates the logic.

So the interrupt flags are per interrupt source? That would make the logic indeed much more straightforward.

kenbell:
Yes - interrupt flags are per peripheral, as far as I can tell it's literally just the interrupt numbers that are shared.

deadprogram:
@kenbell can you please rebase against `dev`? That should fix your CI failure.

kenbell:
@deadprogram thanks - have rebased.

By way of a working example of using this, this is the in-progress PWM support for stm32.  Note TIM1 / TIM10 and other timer peripherals share interrupt numbers.

https://github.com/tinygo-org/tinygo/pull/1867/files#diff-9b429ebca851cfb19ab60d97eed65c83146bda283fbfe7297d57392419f3c4c5R480

transform/interrupt.go
```diff
@@ -73,25 +73,50 @@ func LowerInterrupts(mod llvm.Module, sizeLevel int) []error {
...
+			handleMap[num] = append(handles, global)
 		}
 	}
 
-	// Iterate over all handler objects, replacing their ptrtoint uses with a
+	// Iterate over all handle objects, replacing their ptrtoint uses with a
 	// real interrupt ID and creating an interrupt handler for them.
-	for _, global := range handlers {
-		initializer := global.Initializer()
+	for _, handles := range handleMap {
```
aykevl: This introduces non-deterministic behavior in the compiler because the iteration over the map is done in unspecified order.

Example:

```
$ tinygo build -o test.hex -target=microbit examples/serial && md5sum test.hex
59c5d997db197b0e4707f7541935be4f  test.hex

$ tinygo build -o test.hex -target=microbit examples/serial && md5sum test.hex
ab431d70b7686fe593c8fbd7073ad885  test.hex
```

aykevl (changes requested):
This generally looks good to me except there is one thing that needs fixing: the `handleMap` iteration order. I don't care too much about which order you pick as long as it is stable (for example, the order in which they're added to the map or simply by numerically sorting the interrupt numbers).

transform/interrupt.go
kenbell: Thanks - yes, that definitely needs fixing.  I'll work on a fix.

kenbell: I've pushed a fix to sort the handlers in interrupt number order.  Builds are now repeatable:
```shell
$ tinygo build -o test.hex -target=microbit examples/serial && md5sum test.hex
d988eef1e3bc84fc8def75acb677470c  test.hex
$ tinygo build -o test.hex -target=microbit examples/serial && md5sum test.hex
d988eef1e3bc84fc8def75acb677470c  test.hex
```

kenbell:
@aykevl is this PR OK to be merged?  I think I took care of the build reproducibility - are there any other concerns?   I'd like to get the STM32 timer changes merged that depend on this PR.   Thanks!

aykevl:
Yes, the updates look good to me! It can be merged from my point of view.

aykevl (approved):

kenbell:
@deadprogram - I think this is ready to be merged.  Then I can rebase #1867.  Thanks!

deadprogram:
Amazing work on this contribution @kenbell thank you very much!

Now merging.

PR #1849, opened 2021-05-02
ci: bump Xcode version to use macOS 10.14

aykevl:
The CircleCI macOS builds are failing, probably due to the old macOS version that's used. This version (10.13 High Sierra) isn't supported anymore on Homebrew so it seems best to me to simply bump the version.

I picked Xcode 11.1.0 because 10.3.0 is somehow triggering an error while trying to install QEMU (the Python install fails).

Because of this newer Xcode version, I had to add an extra flag (-isysroot) to the default command line for MacOS. The reason is that this newer Xcode version no longer stores header files in /usr/local, an SDK must be specified manually. With this change, the default SDK is used.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -294,16 +294,16 @@ commands:
           variant: "macos"
       - restore_cache:
           keys:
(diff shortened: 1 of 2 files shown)


aykevl:

deadprogram:
:trophy: thanks @aykevl now merging.

PR #1852, opened 2021-05-03
Fix I2c initialization for stm32f1/f4 

ofauchon:
Hi, 

This PR fixes "panic: runtime error: nil pointer dereference" when using i2c on stm32f1/f4. 
Panic rise in i2c Configure function, as i2c.Bus value is nil:

https://github.com/tinygo-org/tinygo/blob/1f73941c43dac9bc91ca651479655fea5c01c538/src/machine/machine_stm32_i2c_reva.go#L120-L124

The fix  was tested on bluepill +  hd44780-I2C Driver.


kenbell (approved):
Change looks good to me

deadprogram:
Thanks for the fix @ofauchon and for reviewing @kenbell now merging.

PR #1853, opened 2021-05-03
wasm: scan globals conservatively

aykevl:
Make the GC globals scan phase conservative instead of precise on
WebAssembly. This reduces code size at the risk of introducing some
false positives.

This is a stopgap measure to mitigate an issue with the precise scanning
of globals that doesn't track all pointers. It works for regular globals
but globals created in the interp package don't always have a type and
therefore may be missed by the AddGlobalsBitmap pass.
The same issue is present on Linux and macOS, but is not as noticeable
there.

The proper fix (for Linux and MacOS) would be either making globals conservative or making globals actually precise, instead of almost-but-not-correctly precise as it is now (there are false negatives, which is a problem).

### DIFF
--- a/src/runtime/arch_wasm.go
+++ b/src/runtime/arch_wasm.go
@@ -14,6 +14,9 @@ const TargetBits = 32
 //go:extern __heap_base
 var heapStartSymbol [0]byte
 
+//go:extern __global_base
+var globalsStartSymbol [0]byte
+
 //export llvm.wasm.memory.size.i32
 func wasm_memory_size(index int32) int32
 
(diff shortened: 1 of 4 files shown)


aykevl:
There is an issue with this patch, it needs a small change to work correctly.

aykevl:

deadprogram:
This seems like one of those seemingly "simple" changes that *could* have some wide-ranging consequences.

@dkegel-fastly @fgsch you have been using the WASM capabilities quite heavily very recently. Any comments from either of you?

Anyone else doing any serious WebAssembly in TinyGo with comments before we merge?

dkegel-fastly:
I can test this Monday afternoon, prolly...

fgsch:
I can also take a deeper look at some point today or tomorrow - I have built a few wasm modules with this PR but nothing beyond that.

fgsch:
Running code with this change and so far so good 👍

deadprogram:
Awesome, thanks for the update @aykevl and for testing it out @fgsch 

Now merging.

PR #1355, opened 2020-09-05
Smoketest with go script

sago35:
#1330

I created a PR to make it easier to check the CI results.
There is no plan to merge it for now.

PR #1858, opened 2021-05-05
atsame5x: add support for CAN

sago35:
In this PR, we will add CAN functionality to atsame5x.
This is the first CAN feature for TinyGo.

The design of the API has not yet been finalized.
It depends on atsame5x for many parts.
However, CAN does indeed work now.

examples/can
```
CAN0 2
-> 00000123 8 []byte{0x1, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xef}
-> 00000789 A []byte{0x2, 0x24, 0x46, 0x67, 0x89, 0xab, 0xcd, 0xef, 0x1, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xef}
CAN0 2
-> 00000123 8 []byte{0x1, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xef}
-> 00000789 A []byte{0x2, 0x24, 0x46, 0x67, 0x89, 0xab, 0xcd, 0xef, 0x1, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xef}
```

examples/caninterrupt
```
1 123 8 01 23 45 67 89 AB CD EF
2 456 3 AA BB CC
1 123 8 01 23 45 67 89 AB CD EF
2 456 3 AA BB CC
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -343,7 +343,11 @@ smoketest:
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=atsame54-xpro       examples/blinky1
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=atsame54-xpro       examples/can
+	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=feather-m4-can      examples/blinky1
+	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=feather-m4-can      examples/caninterrupt
 	@$(MD5SUM) test.hex
 	# test pwm
 	$(TINYGO) build -size short -o test.hex -target=itsybitsy-m0        examples/pwm
--- a/src/examples/can/feather-m4-can.go
+++ b/src/examples/can/feather-m4-can.go
@@ -1,0 +1,15 @@
+// +build feather_m4_can
+
+package main
+
+import (
+	"machine"
+)
+
+func init() {
+	// power on the CAN Transceiver
+	// https://learn.adafruit.com/adafruit-feather-m4-can-express/pinouts#can-bus-3078990-8
+	boost_en := machine.BOOST_EN
+	boost_en.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	boost_en.High()
+}
--- a/src/examples/can/main.go
+++ b/src/examples/can/main.go
@@ -1,0 +1,53 @@
+package main
+
+import (
+	"fmt"
+	"machine"
+	"time"
+)
+
+func main() {
+	can1 := machine.CAN1
+	can1.Configure(machine.CANConfig{
+		TransferRate:   machine.CANTransferRate500kbps,
+		TransferRateFD: machine.CANTransferRate1000kbps,
+		Rx:             machine.CAN1_RX,
+		Tx:             machine.CAN1_TX,
+		Standby:        machine.CAN1_STANDBY,
+	})
+
+	can0 := machine.CAN0
+	can0.Configure(machine.CANConfig{
+		TransferRate:   machine.CANTransferRate500kbps,
+		TransferRateFD: machine.CANTransferRate1000kbps,
+		Rx:             machine.CAN0_RX,
+		Tx:             machine.CAN0_TX,
+		Standby:        machine.NoPin,
+	})
+
+	rxMsg := machine.CANRxBufferElement{}
+
+	for {
+		can1.Tx(0x123, []byte{0x01, 0x23, 0x45, 0x67, 0x89, 0xAB, 0xCD, 0xEF}, false, false)
+		can1.Tx(0x789, []byte{0x02, 0x24, 0x46, 0x67, 0x89, 0xAB, 0xCD, 0xEF, 0x01, 0x23, 0x45, 0x67, 0x89, 0xAB, 0xCD, 0xEF}, true, false)
+		time.Sleep(time.Millisecond * 1000)
+
+		sz0 := can0.RxFifoSize()
+		if sz0 > 0 {
+			fmt.Printf("CAN0 %d\r\n", sz0)
+			for i := 0; i < sz0; i++ {
+				can0.RxRaw(&rxMsg)
+				fmt.Printf("-> %08X %X %#v\r\n", rxMsg.ID, rxMsg.DLC, rxMsg.Data())
+			}
+		}
+
+		sz1 := can1.RxFifoSize()
+		if sz1 > 0 {
+			fmt.Printf("CAN1 %d\r\n", sz1)
+			for i := 0; i < sz1; i++ {
+				can1.RxRaw(&rxMsg)
+				fmt.Printf("-> %08X %X %#v\r\n", rxMsg.ID, rxMsg.DLC, rxMsg.Data())
+			}
+		}
+	}
+}
--- a/src/examples/caninterrupt/main.go
+++ b/src/examples/caninterrupt/main.go
@@ -1,0 +1,75 @@
+package main
+
+import (
+	"device/sam"
+	"fmt"
+	"machine"
+	"time"
+)
+
+type canMsg struct {
+	ch   byte
+	id   uint32
+	dlc  byte
+	data []byte
+}
+
+func main() {
+	ch := make(chan canMsg, 10)
+	go func() {
+		for {
+			select {
+			case m := <-ch:
+				fmt.Printf("%d %03X %X ", m.ch, m.id, m.dlc)
+				for _, d := range m.data {
+					fmt.Printf("%02X ", d)
+				}
+				fmt.Printf("\r\n")
+			}
+
+		}
+	}()
+
+	can1 := machine.CAN1
+	can1.Configure(machine.CANConfig{
+		TransferRate:   machine.CANTransferRate500kbps,
+		TransferRateFD: machine.CANTransferRate1000kbps,
+		Rx:             machine.CAN1_RX,
+		Tx:             machine.CAN1_TX,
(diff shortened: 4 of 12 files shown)


src/examples/can/atsame54-xpro.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"machine"
+)
+
+func init() {
+	// Activate CAN Transceiver
+	stb := machine.CAN1_STANDBY
+	stb.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	stb.Low()
```
aykevl: What does this do?

src/examples/can/feather-m4-can.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"machine"
+)
+
+func init() {
+	// power on the CAN Transceiver
+	boost_en := machine.BOOST_EN
+	boost_en.Configure(machine.PinConfig{Mode: machine.PinOutput})
+	boost_en.High()
```
aykevl: And what does this do?

src/examples/can/main.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"device/sam"
+	"fmt"
+	"machine"
+	"time"
+)
+
+func main() {
+	can1 := machine.CAN{Bus: sam.CAN1}
```
aykevl: I think it would be better if the machine package defined `CAN1` this way, so you can use `machine.CAN1` directly. For example, like this:

```go
var CAN1 = &CAN{Bus: sam.CAN1}
```

src/examples/can/main.go
```diff
@@ -0,0 +1,52 @@
...
+		TransferRateFD: machine.CANTransferRate1000kbps,
+		Rx:             machine.CAN0_RX,
+		Tx:             machine.CAN0_TX,
+	})
+
+	rxMsg := machine.CANRxBufferElement{}
+
+	for {
+		can1.Tx(0x123, 8, []byte{0x01, 0x23, 0x45, 0x67, 0x89, 0xAB, 0xCD, 0xEF}, false, false)
+		can1.Tx(0x789, 16, []byte{0x02, 0x24, 0x46, 0x67, 0x89, 0xAB, 0xCD, 0xEF, 0x01, 0x23, 0x45, 0x67, 0x89, 0xAB, 0xCD, 0xEF}, true, false)
```
aykevl: The length is specified twice: once as a parameter and once in the byte slice. It should be enough to specify it once:

```suggestion
		can1.Tx(0x123, []byte{0x01, 0x23, 0x45, 0x67, 0x89, 0xAB, 0xCD, 0xEF}, false, false)
		can1.Tx(0x789, []byte{0x02, 0x24, 0x46, 0x67, 0x89, 0xAB, 0xCD, 0xEF, 0x01, 0x23, 0x45, 0x67, 0x89, 0xAB, 0xCD, 0xEF}, true, false)
```

src/examples/caninterrupt/main.go
```diff
@@ -0,0 +1,71 @@
...
+	}()
+
+	can1 := machine.CAN{Bus: sam.CAN1}
+	can1.Configure(machine.CANConfig{
+		TransferRate:   machine.CANTransferRate500kbps,
+		TransferRateFD: machine.CANTransferRate1000kbps,
+		Rx:             machine.CAN1_RX,
+		Tx:             machine.CAN1_TX,
+	})
+	can1.SetInterrupt(sam.CAN_IE_RF0NE, func(machine.CAN) {
```
aykevl: What does `sam.CAN_IE_RF0NE` mean?

Also, I think you'll want to pass a pointer, not a value:

```suggestion
	can1.SetInterrupt(sam.CAN_IE_RF0NE, func(*machine.CAN) {
```

src/machine/machine_atsame5x_can.go
```diff
@@ -0,0 +1,456 @@
...
+
+//go:align 4
+var CANTxFifo [2][(8 + 64) * CANTxFifoSize]byte
+
+//go:align 4
+var CANEvFifo [2][(8) * CANEvFifoSize]byte
+
+type CAN struct {
+	Bus     *sam.CAN_Type
+	Channel byte
```
aykevl: It looks like `Channel` is unnecessary: it is always 0 when `Bus` equals `*sam.CAN0` and 1 when `Bus` equals `sam.CAN1`.
If you want to simplify the code, you could add this helper for example:

```go
func (can *CAN) instance() byte {
    if can.Bus == sam.CAN0 {
        return 0
    } else {
        return 1
    }
}
```

aykevl (commented):
I don't know much about CAN so I find it hard to judge this API. But I've added some review comments anyway.

src/examples/can/main.go
sago35: Indeed.
Changed.

src/examples/can/main.go
sago35: Indeed, the length was extra.
I fixed it.

src/examples/caninterrupt/main.go
sago35: Changed.

src/machine/machine_atsame5x_can.go
sago35: Changed.

src/examples/caninterrupt/main.go
sago35: `sam.CAN_IE_RF0NE` is  `Rx FIFO 0 New Message Interrupt Enable` register.
This interrupt is generated every time a CAN frame is received.

sago35: The RF0NE is described in the following page 1141.

https://cdn-learn.adafruit.com/assets/assets/000/097/530/original/SAM_D5xE5x_Family_Data_Sheet_DS60001507F.pdf?1607462315

src/examples/can/feather-m4-can.go
sago35: This operation is used to turn on the power of the CAN transceiver.
Since the on-board CAN transceiver requires 5V, the power supply can be controlled in consideration of LiPo drive.

https://learn.adafruit.com/adafruit-feather-m4-can-express/pinouts#can-bus-3078990-8

src/examples/can/atsame54-xpro.go
sago35: Many CAN transceivers have Standby-mode.
In order to be able to communicate with CAN, it needs to be set to Normal-mode, so the port is manipulated.

* atsame54-xpro : ATA6561
    * https://www.microchip.com/wwwproducts/en/ata6561
* feather-m4-can : TCAN1051-Q1
    * https://www.ti.com/product/TCAN1051-Q1

sago35:
@aykevl 
Thank you for reviewing this.
I have responded to the comments received.

Makefile
```diff
@@ -338,9 +338,9 @@ smoketest:
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=p1am-100            examples/blinky1
 	@$(MD5SUM) test.hex
-	$(TINYGO) build -size short -o test.hex -target=atsame54-xpro       examples/blinky1
+	$(TINYGO) build -size short -o test.hex -target=atsame54-xpro       examples/can
```
deadprogram, line 346: Should probably build both `blinky1` and also `can` in the smoke test, since they are testing different things, right?

src/examples/can/atsame54-xpro.go
deadprogram: Perhaps adding an interface function such as `func (can *CAN) Standby(on bool) error` would be a good feature to clarify intent and also require less datasheet specific knowledge.

If the specific CAN transceiver does not support this mode, the function could return an error that was "Standby not supported" or something like that.

src/examples/can/feather-m4-can.go
deadprogram: Same as above comment, but adding `func (can *CAN) Boost(enable bool) error`

src/examples/can/atsame54-xpro.go
sago35: Standby Pin was added to CAN.Configure(). This simplifies the transition from Standby-mode to Normal-mode. Thanks @deadprogram .

On the other hand, I don't think Standby(bool) is necessary for the CAN function at the moment, because I haven't created a side to stop CAN. I will PR it again in the future when I want to stop CAN. A typical use of Standby is to reduce the power consumption when the microcontroller is put to sleep.

src/examples/can/feather-m4-can.go
sago35: This process is specific to feather-m4-can, so I don't think it is necessary for the CAN function. I added the above URL to the comments.

Makefile
sago35: I restored the blinky1 test.

I had to remove the blinky1 test, which bothered me. My concern is that smoketest is getting more and more bloated, and CI time is increasing. Also, the local tests are taking longer.

This concern will be addressed significantly in #1355.

src/examples/can/atsame54-xpro.go
deadprogram: OK make sense. Thanks!

Makefile
deadprogram: Thanks for adding back.

deadprogram:
@sago35 I think this looks good to be merged. Can you please squash the commits? Thank you!

sago35:
@deadprogram 
Thank you for reviewing it.
I squashed and rebased it.

deadprogram:
Really great work on this @sago35 it opens up a whole new world of possibilities! :car: :truck: :ship: 

Now merging, thanks again.

PR #1859, opened 2021-05-05
qtpy: fix i2c setting

sago35:
Fix QT Py's i2c to SERCOM1.

https://github.com/adafruit/ArduinoCore-samd/blob/master/variants/qtpy_m0/variant.h#L159-L170


deadprogram:
Thanks @sago35 now merging.

PR #1860, opened 2021-05-05
unix: use conservative GC by default

aykevl:
This commit does two things:

 1. It makes it possible to grow the heap on Linux and MacOS by
    allocating 1GB of virtual memory on startup and then slowly using it
    as necessary, when running out of available heap space.
 2. It switches the default GC to be the conservative GC (previously
    extalloc). This is good for consistency with other platforms that
    all use this same GC.

This makes the extalloc GC unused by default.

@niaow the direct motivation is to fix a bug in the extalloc GC, see: https://github.com/tinygo-org/tinygo/pull/1445#issuecomment-832300285
But maybe you have a different solution?

### DIFF
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -80,12 +80,7 @@ func (c *Config) GC() string {
 	if c.Target.GC != "" {
 		return c.Target.GC
 	}
-	for _, tag := range c.Target.BuildTags {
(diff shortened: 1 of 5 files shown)


niaow (approved):
This looks fine, although idk what is up with tinyci rn

deadprogram:
:+1: merging, thanks @aykevl and @niaow

PR #1719, opened 2021-03-15
atsamd21: improve SPI

sago35:
This is the samd21 version of #1453 .

Since samd21 has a slower CPU than samd51, we created a special function to speed it up.

I think it will be necessary for tinygo-org/drivers/ili9341 to run faster.
It's also better to be fast for the sdcard SPI driver.

I will eventually use DMA to send/receive, but it will be very late to work on DMA support for samd21.

----

### old

old: 7d1ce24be5c60525e14b25a29bc0833b454ab7d4
image

----

### new
txrx24mhz: 3bc7451c04c63486638d37d0d38279265823157c
image

tx24mhz: 3bc7451c04c63486638d37d0d38279265823157c
image

rx: 3bc7451c04c63486638d37d0d38279265823157c
image

txrx with SPI.Freq == 12MHz: 3bc7451c04c63486638d37d0d38279265823157c
When the SPI Freq is less than 12Mhz, it seems that no special processing is required.
image

PR #1861, opened 2021-05-05
atsamd21: remove special handling for SPI-24mhz

sago35:
Speedup at SPI Freq 24Mhz was added in #1719.
However, at this point, I have confirmed that data may be lost.
Therefore, we decided to stop the special handling, although the speed will be a little slower.


deadprogram:
:hammer_and_wrench: merging. Thanks @sago35

Issue #1862, opened 2021-05-05
regression: tinygo apps crash in one wasi runtime if argc == 0?


dkegel-fastly:
Since https://github.com/tinygo-org/tinygo/commit/7b761fac7805083dd0263ed376b10e106726ef8f,
tinygo apps crash on one wasm runtime.  It seems the runtime used by my employer does not provide argv[0].
I could not reproduce the problem with wasmtime or wasmer, but the problem is quite reproducible in our environment.

I also tested on mac and linux, using execve to run a tinygo app without argv[0], but that worked without a crash;
linux actually does allow running with no argv[0], believe it or not, but that didn't cause a problem as it did with wasi.

The change to tinygo to avoid the crash in the wasi case is, well, tiny.  Pull request coming.

deadprogram:
Since this issue was introduced in-between releases I am going to mark as closed. Thanks @dkegel-fastly for the report and the fix!

PR #1863, opened 2021-05-05
Don't crash on empty argv in wasi

dkegel-fastly:
Fixes regression #1862

### DIFF
--- a/src/runtime/runtime_wasm_wasi.go
+++ b/src/runtime/runtime_wasm_wasi.go
@@ -30,6 +30,9 @@ func init() {
 	// these args (argv).
 	var argc, argv_buf_size uint32
 	args_sizes_get(&argc, &argv_buf_size)
+	if argc == 0 {
+		return
+	}
 
 	// Obtain the command line arguments
 	argsSlice := make([]unsafe.Pointer, argc)


aykevl:
I would expect that at least one parameter would be provided, but the WASI spec is silent on this. So, I guess we'll need to deal with that.

LGTM.

tschneidereit:
For better or worse, WASI inherits this from POSIX, which has this to say:
> Early proposals required that the value of argc passed to main() be "one or greater". This was driven by the same requirement in drafts of the ISO C standard. In fact, historical implementations have passed a value of zero when no arguments are supplied to the caller of the exec functions. This requirement was removed from the ISO C standard and subsequently removed from this volume of POSIX.1-2017 as well. The wording, in particular the use of the word should, requires a Strictly Conforming POSIX Application to pass at least one argument to the exec function, thus guaranteeing that argc be one or greater when invoked by such an application. In fact, this is good practice, since many existing applications reference argv[0] without first checking the value of argc.

([Source](https://pubs.opengroup.org/onlinepubs/9699919799/functions/exec.html))

deadprogram:
Thanks for the fix @dkegel-fastly and for the additional details @tschneidereit 

Now merging.

PR #1864, opened 2021-05-06
machine: move PinMode to central location

aykevl:
It is always implemented exactly the same way (as an uint8) so there is no reason to implement it in each target separately.

This also makes it easier to add some documentation to it.


deadprogram:
:star: merging.

PR #1865, opened 2021-05-06
main: match `go test` output

### LINKED ISSUE #1831: Panic when building/testing a package without any code
```
$ cat *.go
// mylib is documented here.
package mylib

$ tinygo test .
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x6c2b65]

goroutine 21 [running]:
github.com/tinygo-org/tinygo/compiler.CompilePackage(0xc000262640, 0xb, 0xc00021f2c0, 0xc0000f5c20, 0x12398a0, 0xc00023c0b0, 0x0, 0x0, 0x0, 0x0, ...)
        /data/src/tinygo/compiler/compiler.go:253 +0x105
github.com/tinygo-org/tinygo/builder.Build.func1(0xc0000f5d40, 0x0, 0x0)
        /data/src/tinygo/builder/build.go:199 +0x128
github.com/tinygo-org/tinygo/builder.jobWorker(0xc000282120, 0xc0002820c0)
        /data/src/tinygo/builder/jobs.go:165 +0xe2
        /data/src/tinygo/builder/jobs.go:84 +0xc5
```

Same results for `tinygo build`.

Trying to build/test this is a bit silly of course; the case where I ran in to this is that I have a library where all the actual code is in subpackages, with just a `doc.go` and `go.mod` at the top-level. Still, it shouldn't panic.

Looking at `CompilePackage()`, it's `ssaPkg.Prog.ImportedPackage("runtime")` that returns `nil`; I suppose because it doesn't have any code, it doesn't import `runtime`. I'm not sure if adding a `!= nil` check here is the best way to fix it though.

    $ tinygo version
    tinygo version 0.18.0-dev linux/amd64 (using go version go1.16.3 and LLVM version 11.0.0)

(this is commit 80caf2d)

aykevl:
This commit makes the output of `tinygo test` similar to that of `go test`. It changes the following things in the process:

  * Running multiple tests in a single command is now possible. They aren't paralellized yet.
  * Packages with no test files won't crash TinyGo, instead it logs it in the same way the Go toolchain does.

This fixes #1831.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -39,6 +39,11 @@ type BuildResult struct {
 	// The directory of the main package. This is useful for testing as the test
 	// binary must be run in the directory of the tested package.
 	MainDir string
+
+	// ImportPath is the import path of the main package. This is useful for
+	// correctly printing test results: the import path isn't always the same as
+	// the path listed on the command line.
+	ImportPath string
 }
 
 // packageAction is the struct that is serialized to JSON and hashed, to work as
@@ -637,8 +642,9 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 		return fmt.Errorf("unknown output binary format: %s", outputBinaryFormat)
 	}
 	return action(BuildResult{
-		Binary:  tmppath,
-		MainDir: lprogram.MainPkg().Dir,
+		Binary:     tmppath,
+		MainDir:    lprogram.MainPkg().Dir,
+		ImportPath: lprogram.MainPkg().ImportPath,
 	})
 }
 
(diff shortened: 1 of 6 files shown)


builder/build.go
```diff
@@ -39,6 +39,11 @@ type BuildResult struct {
 	// The directory of the main package. This is useful for testing as the test
 	// binary must be run in the directory of the tested package.
 	MainDir string
+
+	// ImportPath is the import path of the main package. This is useful for
+	// correctly printing test results: the import path isn't always the same as
+	// the path listed on the command line.
+	ImportPath string
```
deadprogram, line 46: I needed exactly this. :star:

deadprogram:
:heavy_check_mark: thanks @aykevl now merging.

PR #1866, opened 2021-05-06
Draft: Stm32's Tick Timer optimization

ofauchon:
This PR is an attempt to optimize STM32 's timers.

Current Tick Timer is configured to fire one interrupt every millisecond (with TICK_RATE = 1000) 
In some situation, 1khz is not enough (eg: some drivers needs acurate, sub-millisecond time.Sleeps )

Of course, it's possible to increase TICK_RATE to improve granularity and precision. 
But it increases interrupt rate at the same level, which is a waste of CPU's resources (and probably energy).

Thanks to @aykevl suggestion, I tried to implement a new tick timer's behaviour: 

You can now adjust two parameters: 

- TICK_RATE : defines  the desired tick rate. (eg: we want 32000 ticks/sec) 
- TICK_INTFREQ : defines the desired timer's interrupt frequency ( 1000 int / sec for example)

With this patch, tick timer will be configured to overflow at TICK_INTFREQ frequency 
During the interrupt, the global tickCount value will be updated (every 1ms when TICK_INTFREQ=1000 , which is not very accurate)

But the tick() function now uses the current coarse tickCount , and adds the "missing part of the time" since the interrupt 
(by derivating the timer's counter register) 

Any comments welcome !

@kenbell: Could you please review the code and make some quick tests on that PR ? 

Olivier

Thanks to :
@kenbell  for the awesome work on STM32's timers standardization
@aykevl for the suggestions, and support on #Tinygo.

### DIFF
--- a/src/runtime/runtime_stm32_timers.go
+++ b/src/runtime/runtime_stm32_timers.go
@@ -11,11 +11,11 @@ package runtime
 // Interface
 // ---------
 // For each MCU, the following constants should be defined:
-//   TICK_FREQ           The desired frequency of ticks, e.g. 10000 for 10KHz ticks
-//   TICK_TIMER_INTFREQ  The desired frequency of timer interrupts ex: 1000 for 1Khz interrupts
-//   TICK_TIMER_IRQ      Which timer to use for counting ticks (e.g. stm32.IRQ_TIM7)
+//   TICK_RATE             The desired rate of ticks, e.g. 10000 for 10KHz ticks
+//   TICK_TIMER_INTFREQ    The desired frequency of timer interrupts ex: 1000 for 1Khz interrupts
+//   TICK_TIMER_IRQ        Which timer to use for counting ticks (e.g. stm32.IRQ_TIM7)
 //   TICK_TIMER_CLOCKFREQ  The frequency the clock feeding the sleep timer is set to (e.g. 84MHz)
-//   SLEEP_TIMER_IRQ  Which timer to use for sleeping (e.g. stm32.IRQ_TIM3)
+//   SLEEP_TIMER_IRQ       Which timer to use for sleeping (e.g. stm32.IRQ_TIM3)
 //   SLEEP_TIMER_CLOCKFREQ The frequency the clock feeding the sleep timer is set to (e.g. 84MHz)
 //
 // The type alias `arrtype` should be defined to either uint32 or uint16 depending on the
@@ -23,7 +23,6 @@ package runtime
 
 import (
 	"device/stm32"
-	"machine"
 	"runtime/interrupt"
 	"runtime/volatile"
 )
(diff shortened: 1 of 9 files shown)


src/runtime/runtime_stm32_timers.go
```diff
@@ -58,7 +64,13 @@ func nanosecondsToTicks(ns int64) timeUnit {
 // number of ticks (microseconds) since start.
 //go:linkname ticks runtime.ticks
 func ticks() timeUnit {
-	return timeUnit(tickCount.Get())
+	// Global tickCount variable is updated at every interrupt, but this is a coarse estimation
+	// of elapsed time. (tickCount never includes elapsed time between two interrupts)
+	// In this function, we need a precise value of elapsed time.
+	// This is the reason we complement tickCount with extra time computed from timer's current value.
+	fractionnal := (TICKS_PER_INTS * tickTimer.Device.CNT.Get()) / timerARR
```
kenbell: Typo?  Should be 'fractional'?

src/runtime/runtime_stm32_timers.go
```diff
@@ -67,26 +79,27 @@ func ticks() timeUnit {
...
+	timerPSC = uint32(TICK_TIMER_CLOCKFREQ / TICK_TIMER_INTFREQ)
+	timerARR = uint32(1)
 
-	// Get the pre-scale into range, with interrupt firing
-	// once per tick.
-	for psc > 0x10000 || period == 1 {
-		psc >>= 1
-		period <<= 1
+	// Find correct prescaler/Counter limit
+	for timerARR < 512 {
```
kenbell: Why 512?  If it's to get precision in ARR, might want to do something more like this? https://github.com/tinygo-org/tinygo/pull/1867/files#diff-45e7f001091fdbec2fb598052e4dd8232744d1a2974bd2e658a29c874e8e5855R124

(borrowed from another PWM implementation - it chooses PSC to get max precision in ARR)

src/runtime/runtime_stm32_timers.go
```diff
@@ -45,6 +47,10 @@ var (
 
 	// The timer used for sleeping
 	sleepTimer *timerInfo
+
+	// timer Prescaler and Auto-reload
+	timerPSC uint32
```
kenbell: It looks like timerPSC is only used inside one function, perhaps don't have as a global?

kenbell (commented):
Looks good - left a few comments.

I'm working on converting the timer functionality into 'TIM' objects in the machine package that will do PWM, but can also simplify the tick/sleep implementation in runtime.  I'll make sure to include this logic when I make that change.

kenbell:
I've checked a couple of examples on STM32F407G-DISC1 and NUCLEO-L552ZE, worked fine.

ofauchon:
I'm pleased to know you are working on a major rewrite of timer for PWM. 
Should I continue my work on this PR, or you prefer I suspend It and takeover these changes in your PWM branch? 

I can make some experiments with systick for tick timer too, if that can be useful. 

Olivier

kenbell:
If you can take a look at Systick, that would be awesome.  I really liked your idea about sleep implementation vary depending on how long the sleep is - if it's longer than the 'interrupt' interval, wait for interrupt - if it's less than that, then do a busy wait.

Doing that I think we can use systick for sleep and for ticks and not use any timers?

ofauchon:
This PR has been superseded by @kenbell work on timers refactor:
 
https://github.com/tinygo-org/tinygo/pull/1867

PR #1867, opened 2021-05-07
stm32: implement PWM and tick/sleep by exposing machine.TIMx peripherals

kenbell:
This is a rework of the STM32 timer logic:
1. Code to configure timers has been made generic (meets PWM interface) and move to `machine` package
2. Code using timers has been re-built to use new logic in `machine` package
3. For supported boards, documented timers have been created as machine.TIMx devices
4. Reduced from 1 timer for tick and 1 for sleep to a single timer that uses output compare for granular sleep

Verified on:
- bluepill
- stm32f4disco-1
- nucleo-l432kc
- nucleo-l031k6
- nucleo-l552ze
- nucleo-f722ze

Depends on:
#1846 (pending)
#1885 (fixed)


kenbell:
@ardnew @ofauchon - would be great if you could take a look at this

kenbell:
@deadprogram - I can't see why the circleci jobs failed.  If you can get logs I can take a look.

deadprogram:
Looks like you just need to run `go fmt` on the code, :smile_cat:

ofauchon:
@kenbell  
I'll start testing your code on STM32F103 (BluePill) andSTM32L0 (LGT92) immediately.

ofauchon:
Hi ... seems there is something broken when I try to test this branch: 

```
[olivier@psx kenbell-tinygo]$ git pull
[olivier@psx kenbell-tinygo]$ git branch
  master
  release
* stm32-pwm
[olivier@psx blinky1]$ #go build && go install && tinygo clean 
[olivier@psx kenbell-tinygo]$ cd src/examples/blinky1
[olivier@psx blinky1]$ tinygo build --target=bluepill -o blinky1.elf blinky1.go 
/home/olivier/dev/contrib/kenbell-tinygo/src/machine/board_bluepill.go:6:2: package device/stm32 is not in GOROOT (/home/olivier/.cache/tinygo/goroot-go1.15.6-1cb89e78f9212eeca2ff2e69cb49170c39199d0c47ae97fcbab3a3f355599082-syscall/src/device/stm32)
```

Did I make something wrong when installing your tinygo code ?

deadprogram:
@ofauchon don't forget to run `make gen-device`.

ofauchon:
> @ofauchon don't forget to run `make gen-device`.

Yes, that's it : 

```
git submodule update --init && make gen-device-stm32
```

ofauchon:
@kenbell 

I made some performance tests on timer/sleep  (bluepill target) : 

I use this kind of code  
```
for i := 0; i < 3; i++ {
   machine.PA0.High() 
   time.Sleep(t)
   machine.PA0.Low()
}
```
.... With 1s, 100ms, 10ms, 1ms, 100uS, 10uS, 1uS  periods. 

And I measured the results with a 8MSamples/sec logic analyzer: 


|period| result|
|-|-|
|1s |  OK => 1.001 sec |
|100ms |  OK => 0.1001 sec |
|10ms |  OK/STRANGE => 10ms then 2x20ms image |
|1ms |  OK => 1,052 ms |
|100us |  FAIR => 0.130 us |
|10us |  NOK => 34uS  |
|1us |  NOK => 32uS  |


Based on these first results, my opinion is that this patch outperforms the current timers's implementation for the board. 
(Previously, I could never Sleep properly under the millisecond). 

Maybe we could have a closer look to the 10ms sleep test (the one with the the screenshot), which  produce unexpected results (10ms, then 20ms, then 20ms) . 

Please note  this patch also solves the problems I had with HD44780 i2c driver's too ! 

I'll try to make some extra tests on LGT92 board soon to confirm these results. 

Anyway, you did a awesome work @kenbell !  thx 

Olivier

kenbell:
@ofauchon thanks!  10ms is suspiciously the same as the interrupt frequency (100Hz) - probably an edge-case I need to take a look at.

kenbell:
Hmm - i can't repro on my bluepill - i've done a bunch of resets, this is what I get...
image


This is the full test program I'm using:
```golang
package main

import (
	"machine"
	"time"
)

func main() {
	pin := machine.PA1
	pin.Configure(machine.PinConfig{Mode: machine.PinOutput})
	t := 10*time.Millisecond

	for i := 0; i < 3; i++ {
		pin.High() 
		time.Sleep(t)
		pin.Low()
	}
}
```

Might be a weird hardware timing variation?  I've got another couple of bluepills hopefully headed my way i'll try out.

ofauchon:
Hi @kenbell 

Here is my complete timer test code. 
Could you please test it with your bluepill   

[test_timers.zip](https://github.com/tinygo-org/tinygo/files/6500313/test_timers.zip)

thx

Olivier

kenbell:
I tried your code on my bluepill and on the bluepill-clone that arrived today.  Both showed 10ms working fine:
image

(I had to flip the LED and PA1 pins in your code as pin0 was doing the delay tests)

I think there's something 'special' about your bluepill :)   I don't know if it makes a difference, but i'm powering mine using an STLINK-V2 clone through the 'debug pins'.

I think i see some possibilities for some race conditions in the code where the timer interrupt at just the wrong moment might cause the interrupt to be missed for one cycle - which could explain why you see 20ms delay.  I'll do some speculative changes and ask you to try them.

interp/memory.go
```diff
@@ -196,6 +196,8 @@ func (mv *memoryView) markExternal(llvmValue llvm.Value, mark uint8) {
 				element := llvm.ConstExtractValue(llvmValue, []uint32{uint32(i)})
 				mv.markExternal(element, mark)
 			}
+		case llvm.PointerTypeKind:
+			// Do nothing
```
aykevl: This is incorrect, although it may appear to work. I'll look into it and try to find a good solution.

ofauchon:
@kenbell :

I tested this PR on STM32L0 (LGT92 device) 
=> No problem with timers during my tests.

interp/memory.go
ofauchon: @kenbell : 

For my education, why did you add this extra code ?

aykevl: It's a workaround, here is a fix: https://github.com/tinygo-org/tinygo/pull/1891

kenbell: Thanks @aykevl - I'll take a look at #1891

ofauchon:
@kenbell : 
I made some timing tests with another bluepill.
I could not reproduce  the timing bug with the new one. 
=> Maybe faulty hardware ?

kenbell:
Looks like bluepill is OK, so I think the outstanding work here is to improve #1846 before this is ready to be merged.

deadprogram:
#1846 has been merged now, so all clear to proceed.

kenbell:
@deadprogram - thanks.  I think this change is good to merge.

deadprogram:
Pretty incredible work on this @kenbell and also @ofauchon for all the rounds of testing. Thanks to you both!

Now merging.

deadprogram:
OK, so it turns out that https://github.com/tinygo-org/tinygo/pull/1867/commits/d8e04e4df2ff7496d3e06e77e3c404f63c2fe1b1 breaks the WASM tests. I am planning to revert just that one commit, and then it can be addressed in a new PR.

PR #1870, opened 2021-05-08
runtime: improve timers on nrf, and samd chips

aykevl:
This commit improves the timers on various microcontrollers to better
deal with counter wraparound. The result is a reduction in RAM size of
around 12 bytes and a small effect (sometimes positive, sometimes
negative) on flash consumption. But perhaps more importantly: getting
the current time is now interrupt-safe (it previously could result in a
race condition) and the timer will now be correct when the timer isn't
retrieved for a long duration. Before this commit, a call to `time.Now`
more than 8 minutes after the previous call could result in an incorrect
time.

For more details, see:
https://www.eevblog.com/forum/microcontrollers/correct-timing-by-timer-overflow-count/msg749617/#msg749617

I have made this improvement while working on better documentation for adding new chip support to TinyGo, in particular how to add support for keeping time and for sleeping.

CC @kenbell and @ofauchon: this might be of interest.

### DIFF
--- a/src/runtime/runtime_atsamd21.go
+++ b/src/runtime/runtime_atsamd21.go
@@ -217,11 +217,19 @@ func initRTC() {
 	waitForSync()
 
 	rtcInterrupt := interrupt.New(sam.IRQ_RTC, func(intr interrupt.Interrupt) {
-		// disable IRQ for CMP0 compare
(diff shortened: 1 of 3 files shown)


deadprogram:
This is a very clever technique indeed.

Now merging, good :watch: @aykevl

PR #1871, opened 2021-05-08
runtime: remove the asyncScheduler constant

aykevl:
There is no reason to specialize this per chip as it is only ever used for JavaScript. Not only that, it is causing confusion and is yet another quirk to learn when porting the runtime to a new microcontroller.


deadprogram:
:scissors: merging.

PR #1872, opened 2021-05-09
runtime (gc): simplify extalloc and correct GC scan bounds

niaow:
This change replaces the overcomplicated treap system used previously by the extalloc GC.
The treap was difficult to understand, and added a lot of complexity.
It has been replaced by a simple sorted list, which has roughly the same space overhead and time complexity.

Additionally, a bug was found in the termination condition for the scanning code for both collectors.
On the extalloc collector, this would cause a trailing pointer to be missed.
On the conservative collector, this would cause the scanner to read out of bounds on architectures where pointer size was greater than alignment (AVR mostly).

### DIFF
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -93,7 +93,12 @@ func (c *Config) GC() string {
 	if c.Target.GC != "" {
 		return c.Target.GC
 	}
-	return "conservative"
+	for _, t := range c.Target.BuildTags {
+		if t == "baremetal" || t == "tinygo.wasm" {
+			return "conservative"
+		}
+	}
+	return "extalloc"
 }
 
 // NeedsStackObjects returns true if the compiler should insert stack objects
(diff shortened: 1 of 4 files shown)


niaow:
After this, all tests seem to pass on extalloc again. This seems to have fixed the memory corruption problem as far as I can tell (which seems to have been the incorrect trailing bound).

niaow:
Hmm, stuff still seems to be broken in one place.

niaow:
I have rebased this for now.

dgryski:
Do we want tests for the searching and sorting code?

niaow:
I turned stuff back on for now, but `opt=0` is currently broken do to an unimplemented constant expression in interp.

niaow:
This has been rebased and I have added a tweak to interp to make it work. The interp tweak will be moved into a separate PR probably.

PR #1873, opened 2021-05-09
Glob by /dev/cu.usb* for default port on darwin

ysoldak:
On macOS, there are usually multiple serial ports (Bluetooth, tethered iPhone) active in addition to plugged in board.
Shall ignore all but /dev/cu.usb* to get default port of our board connected via usb.

### DIFF
--- a/main.go
+++ b/main.go
@@ -706,7 +706,9 @@ func getDefaultPort(portCandidates []string) (port string, err error) {
 	switch runtime.GOOS {
 	case "freebsd":
 		ports, err = filepath.Glob("/dev/cuaU*")
-	case "darwin", "linux", "windows":
+	case "darwin":
+		ports, err = filepath.Glob("/dev/cu.usb*")
+	case "linux", "windows":
 		var portsList []*enumerator.PortDetails
 		portsList, err = enumerator.GetDetailedPortsList()
 		if err != nil {
@@ -720,8 +722,6 @@ func getDefaultPort(portCandidates []string) (port string, err error) {
 		if ports == nil || len(ports) == 0 {
 			// fallback
 			switch runtime.GOOS {
-			case "darwin":
-				ports, err = filepath.Glob("/dev/cu.usb*")
 			case "linux":
 				ports, err = filepath.Glob("/dev/ttyACM*")
 			case "windows":


sago35:
I didn't have macOS, so I couldn't test it.

Could you please run the following code?
If you can branch using the information provided below, it would probably be better to do so.

Maybe IsUSB is true for all boards assumed by TinyGo. @deadprogram 

```go
package main

import (
	"fmt"

	"go.bug.st/serial/enumerator"
)

func main() {
	portsList, err := enumerator.GetDetailedPortsList()
	if err != nil {
		panic(err)
	}
	for _, p := range portsList {
		fmt.Printf("%#v\n", p)
	}
}
```

Here are the results of my run on Windows 10.

```
&enumerator.PortDetails{Name:"COM50", IsUSB:true, VID:"239A", PID:"80CD", SerialNumber:"", Product:"USB シリアル デバイス (COM50)"}
&enumerator.PortDetails{Name:"COM1", IsUSB:false, VID:"", PID:"", SerialNumber:"", Product:"通信ポート (COM1)"}
&enumerator.PortDetails{Name:"COM48", IsUSB:true, VID:"2886", PID:"802F", SerialNumber:"", Product:"USB シリアル デバイス (COM48)"}
```

on Ubuntu 20.04.

```
&enumerator.PortDetails{Name:"/dev/ttyACM0", IsUSB:true, VID:"2886", PID:"802d", SerialNumber:"", Product:""}
```

deadprogram:
As documented here https://pkg.go.dev/go.bug.st/serial

> If a port is a virtual USB-CDC serial port (for example an USB-to-RS232 cable or a microcontroller development board) is possible to retrieve the USB metadata, like VID/PID or USB Serial Number, with the GetDetailedPortsList function in the enumerator package

> This library tries to avoid the use of the "C" package (and consequently the need of cgo) to simplify cross compiling. Unfortunately the USB enumeration package for darwin (MacOSX) requires cgo to access the IOKit framework. This means that if you need USB enumeration on darwin you're forced to use cgo

From that description, it is unclear if we can retrieve that info on macOS with our current build config.

ysoldak:
```
&enumerator.PortDetails{Name:"/dev/cu.Bluetooth-Incoming-Port", IsUSB:false, VID:"", PID:"", SerialNumber:"", Product:""}
&enumerator.PortDetails{Name:"/dev/cu.usbmodem14101", IsUSB:true, VID:"2341", PID:"8057", SerialNumber:"", Product:""}
```
Do you guys want I reimplement this using "IsUSB"? I'm fine with that, if you think that's better/cleaner.

deadprogram:
> Do you guys want I reimplement this using "IsUSB"? I'm fine with that, if you think that's better/cleaner.

I am excited that it works, and yes, I think that is a better implementation.

ysoldak:
Shall we use "IsUSB" just for darwin or we can use it across OSes?
Not sure what @sago35 output from Windows 10 tells, is it useable there?

sago35:
@ysoldak 

It' s good to filter by `IsUSB` in windows too.

ysoldak:
Made an alternative PR, uses IsUSB #1874

deadprogram:
Closing because #1874 did this very nicely. Thanks again @ysoldak

PR #1874, opened 2021-05-10
Discover USB ports only

ysoldak:
This will ignore f.ex. bluetooth

Alternative implementation of #1873


deadprogram:
Works perfectly for me @ysoldak thanks for making this change!

Now merging.

ysoldak:
Thanks to @sago35 for proposing use of IsUSB 🥇

PR #1876, opened 2021-05-10
compiler: only check for default stack size with tasks scheduler

aykevl:
With `-scheduler=none`, there are no goroutines so there is no stack size anywhere. Therefore, don't check whether the default stack size is set.


deadprogram:
Looks good, now merging.

PR #1880, opened 2021-05-13
Use machine.Serial as the default output

aykevl:
Previously, the machine.UART0 object had two meanings:

  - it was the first UART on the chip
  - it was the default output for println

These two meanings conflict, and resulted in workarounds like:

  - Defining UART0 to refer to the USB-CDC interface (atsamd21,
    atsamd51, nrf52840), even though that clearly isn't an UART.
  - Defining NRF_UART0 to avoid a conflict with UART0 (which was
    redefined as a USB-CDC interface).
  - Defining aliases like UART0 = UART1, which refer to the same
    hardware peripheral (stm32).

This PR changes this to use a new machine.Serial object for the
default serial port. It might refer to the first or second UART
depending on the board, or even to the USB-CDC interface. Also, UART0
now really refers to the first UART on the chip, no longer to a USB-CDC
interface.

To make that work, I also made both the `machine.USBCDC` and the `machine.UART` pointer receivers so that these objects can more easily be assigned to different variables without copying.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -349,6 +349,8 @@ smoketest:
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=feather-m4-can      examples/caninterrupt
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=arduino-nano33      examples/blinky1
+	@$(MD5SUM) test.hex
 	# test pwm
 	$(TINYGO) build -size short -o test.hex -target=itsybitsy-m0        examples/pwm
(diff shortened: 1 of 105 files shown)


aykevl:
To my big surprise there wasn't a smoketest for `arduino-nano33`, so I've added it in this PR.

aykevl:
The four failing TinyHCI tests should be fixed with this PR: https://github.com/tinygo-org/tinyhci/pull/5

deadprogram:
TinyHCI fixed, and passes the four failing tests.

Now merging, thanks @aykevl

PR #1881, opened 2021-05-13
Add support for net.IP

fgsch:
As discussed on slack, this adds support for `net.IP`. 

The code is copied from Go 1.16, and the names were kept as found there.
Files were trimmed to remove unneeded code at this stage.

### DIFF
--- a/loader/goroot.go
+++ b/loader/goroot.go
@@ -208,6 +208,7 @@ func pathsToOverride(needsSyscallPackage bool) map[string]bool {
 		"internal/reflectlite/": false,
 		"internal/task/":        false,
 		"machine/":              false,
+		"net/":                  true,
 		"os/":                   true,
 		"reflect/":              false,
 		"runtime/":              false,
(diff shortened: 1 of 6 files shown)


aykevl:
This looks like the right way forward to me, so LGTM. It's a straight copy-paste from the Go standard library with parts stripped out. However, I'd like to hear what @soypat thinks of these changes.

(Sorry @soypat for ignoring your PR so long, there are so many things going on in TinyGo that it's hard for me to keep up).

soypat:
LGTM. One minor comment on the auxiliary functions `xtoi`, `ubtoa`, `dtoa` and similar: Would it be reasonable to include a tinygo package for these in the future? They seem to be of use especially in embedded programming and could reduce program sizes.

About my PR: Don't worry about it Ayke, my PR has been quite monolithic. I really did not expect it to be approved for a while. What's more is I've been working on a slimmer improved version. I'll comment soon on the status.

deadprogram:
All makes sense to me. Thanks @fgsch for working on this and to @soypat and @aykevl for looking it over.

Now merging.

PR #1890, opened 2021-05-19
ci: disable building some optional Clang components

aykevl:
This commit disables the Clang static analyzer and ARCMigrate components of Clang. These aren't used at the moment in TinyGo so don't need to be enabled. This reduces the build by 200 files (2909 -> 2709) and thus should improve build time and possibly also reduce the TinyGo binary size.

The idea comes from here (via LLVM weekly):
https://www.cambus.net/speedbuilding-llvm-clang-in-5-minutes/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -157,7 +157,7 @@ commands:
       - llvm-source-linux
       - restore_cache:
           keys:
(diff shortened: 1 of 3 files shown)


deadprogram:
:scissors: :scissors: :scissors: 

Merging, thanks @aykevl

Issue #1885, opened 2021-05-15
interp panic during compilation


kenbell:
branch: fork off dev
build command: tinygo build -target stm32f4disco-1 -o test.elf --opt=0 ./src/examples/pwm/

panic:
```
panic: interp: unknown type kind in markExternalValue

goroutine 51 [running]:
github.com/tinygo-org/tinygo/interp.(*memoryView).markExternal(0xc000d5b860, 0x7f3fb4434f00, 0x7f3fb4434f02)
        /home/kenneth/work/tinygo/interp/memory.go:200 +0x7e5
github.com/tinygo-org/tinygo/interp.(*memoryView).markExternal(0xc000d5b860, 0x7f3fb424d628, 0x7f3fb424d602)
        /home/kenneth/work/tinygo/interp/memory.go:168 +0x372
github.com/tinygo-org/tinygo/interp.(*memoryView).markExternal(0xc000d5b860, 0x7f3fb41a1248, 0x7f3fb41a1202)
        /home/kenneth/work/tinygo/interp/memory.go:168 +0x372
github.com/tinygo-org/tinygo/interp.(*memoryView).markExternal(0xc000d5b860, 0x7f3fb43a7568, 0x7f3fb43a7502)
        /home/kenneth/work/tinygo/interp/memory.go:168 +0x372
github.com/tinygo-org/tinygo/interp.(*memoryView).markExternal(0xc000d5b860, 0x7f3fb46db168, 0x7f3fb46db102)
        /home/kenneth/work/tinygo/interp/memory.go:168 +0x372
github.com/tinygo-org/tinygo/interp.(*memoryView).markExternal(0xc000d5b860, 0x7f3fb43f63b8, 0x7f3fb43f6302)
        /home/kenneth/work/tinygo/interp/memory.go:168 +0x372
github.com/tinygo-org/tinygo/interp.(*memoryView).markExternal(0xc000d5b860, 0x7f3fb46db318, 0x7f3fb444b602)
        /home/kenneth/work/tinygo/interp/memory.go:168 +0x372
github.com/tinygo-org/tinygo/interp.(*memoryView).markExternalStore(...)
        /home/kenneth/work/tinygo/interp/memory.go:111
[...]

PR #1891, opened 2021-05-19
interp: ignore inline assembly in markExternal

aykevl:
The markExternal function is used when a global (function or global
variable) is somehow run at runtime. All the other globals it refers to
are from then on no longer known at compile time, so can't be used by
the interp package anymore.

This can also include inline assembly. While it is possible to modify
globals that way, it is only possible to modify exported globals:
similar to calling an undefined function (in C for example).

Fixes https://github.com/tinygo-org/tinygo/issues/1885.

### DIFF
--- a/interp/memory.go
+++ b/interp/memory.go
@@ -178,6 +178,9 @@ func (mv *memoryView) markExternal(llvmValue llvm.Value, mark uint8) {
 		default:
 			panic("interp: unknown constant expression")
 		}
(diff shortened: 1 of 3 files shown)


kenbell:
@aykevl - thanks.

I've tried this locally for my timer change and it seems to work great!

deadprogram:
@aykevl thanks for the fix and thanks @kenbell for testing it out.

Now merging.

Issue #1886, opened 2021-05-16
Getuid not declared by package os


iarkaroy:
I am working on an wasm project. When building with normal go command like:
```
GOARCH=wasm GOOS=js go build -o app.wasm
```
it is working perfectly, but when building with tinygo:
```
tinygo build -o app.wasm -target=wasm
```
it is showing errors:
```
../go/pkg/mod/github.com/google/uuid@v1.2.0/dce.go:47:42: Getuid not declared by package os
../go/pkg/mod/github.com/google/uuid@v1.2.0/dce.go:55:41: Getgid not declared by package os
```
Go version: `1.16.4`

Any suggestion?

aykevl:
I have made a patch which should fix this: https://github.com/tinygo-org/tinygo/pull/1892

deadprogram:
@iarkaroy can you please check if this corrects your issue? Thank you.

gedw99:
@iarkaroy did you have any success ?
 I am also Interested is go-app project being able to compile with tinygo

iarkaroy:
When building, I am getting this error:
```
cgo/libclang.go:19:10: fatal error: clang-c/Index.h: No such file or directory
   19 | #include <clang-c/Index.h> // if this fails, install libclang-10-dev
```
I have already installed `libclang-10-dev`. Is there anything else I need to do?

skhwhu:
I am also Interested is go-app project being able to compile with tinygo, can you build a fixed tinygo.exe for windows

aykevl:
@iarkaroy that error message is probably outdated, you may need to install libclang-11-dev instead. For details, see https://tinygo.org/docs/guides/build/#with-system-installed-llvm.

aykevl:
@skhwhu you can download a build from here: https://dev.azure.com/tinygo/tinygo/_build/results?buildId=3391&view=artifacts&pathAsName=false&type=publishedArtifacts

skhwhu:
@aykevl thank you very much for you propmt reply. I download tinygo.exe from the above link, but I find "Getuid" doen;t fixed!

C:\Users\SKH\go\pkg\mod\github.com\google\uuid@v1.2.0\dce.go:47:42: Getuid not declared by package os
C:\Users\SKH\go\pkg\mod\github.com\google\uuid@v1.2.0\dce.go:55:41: Getgid not declared by package os

aykevl:
[...]

PR #1892, opened 2021-05-19
syscall: add support for getting uid/gid/pid and related numbers

aykevl:
This is for compatibility with the existing syscall package and in fact
these functions are implemented in the baremetal syscall version. The
numbers will most likely be fake for WebAssembly but that's the job of
wasi-libc to sort out.

I'm not sure how to add tests for these, as they are normally unknown by
definition (unlike command line arguments and environment variables
which can be set by a test program).

See: https://github.com/tinygo-org/tinygo/issues/1886

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -23,7 +23,7 @@ import (
 // Version of the compiler pacakge. Must be incremented each time the compiler
 // package changes in a way that affects the generated LLVM module.
 // This version is independent of the TinyGo version number.
-const Version = 11 // last change: change method name globals
+const Version = 12 // last change: implement syscall.rawSyscallNoError
 
 func init() {
 	llvm.InitializeAllTargets()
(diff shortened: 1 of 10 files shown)


fgsch:
fwiw, this is missing the functions in the os package to fix #1886.

aykevl:

> fwiw, this is missing the functions in the os package to fix #1886.

What do you mean? It does define `Getuid`.

fgsch:
> What do you mean? It does define Getuid.

This is defining the functions under syscall, but the error in #1886 is in the os package.
In other words, they need to be defined under os as well, same as in the stdlib.

aykevl:
Eh, oops. Thank you for pointing that out. I'll submit a fix.

fgsch:
This can go in as is and the changes to os can happen in a separate PR IMO.

aykevl:
I have revamped this PR to also add support for these functions in the os package and added (limited) testing that they can in fact be compiled. This surfaced an issue on Linux which I've also fixed. So now, `os.Getuid` and related functions should at least compile on most platforms.

Unfortunately, it's hard to test these functions as they are sometimes either unavailable entirely (wasm, baremetal) or otherwise their values depend on the environment.

deadprogram:
Seems more solid now that at least some smoke testing is happening. Let's merge and keep going from here.

Issue #1888, opened 2021-05-17
wasm will report an error when using the crypto/x509 package


vicnoah:
`package vrsa

import (
	"crypto/rand"
	"crypto/rsa"
	"crypto/x509"
	"encoding/pem"
	"io"
)

// GenerateRSAKey 生成rsa公私钥
func GenerateRSAKey(priv io.Writer, pub io.Writer, bits int) (err error) {
	// GenerateKey函数使用随机数据生成器random生成一对具有指定字位数的RSA密钥
	// Reader是一个全局、共享的密码用强随机数生成器
	privateKey, err := rsa.GenerateKey(rand.Reader, bits)
	if err != nil {
		return
	}
	// 保存私钥
	// 通过x509标准将得到的ras私钥序列化为ASN.1 的 DER编码字符串
	X509PrivateKey := x509.MarshalPKCS1PrivateKey(privateKey)
	// 使用pem格式对x509输出的内容进行编码
	// 构建一个pem.Block结构体对象
	privateBlock := pem.Block{Type: "RSA Private Key", Bytes: X509PrivateKey}
	// 将数据保存到文件
	err = pem.Encode(priv, &privateBlock)
	if err != nil {
		return
	}
	// 保存公钥
	// 获取公钥的数据
	publicKey := privateKey.PublicKey
	// X509对公钥编码
	X509PublicKey, err := x509.MarshalPKIXPublicKey(&publicKey)
	if err != nil {
		return
	}
	// pem格式编码
	// 创建一个pem.Block结构体对象
	publicBlock := pem.Block{Type: "RSA Public Key", Bytes: X509PublicKey}
	err = pem.Encode(pub, &publicBlock)
	return
}
`
Error message:
/snap/go/current/src/net/pipe.go:156:16: ErrDeadlineExceeded not declared by package os
/snap/go/current/src/net/pipe.go:169:16: ErrDeadlineExceeded not declared by package os
/snap/go/current/src/net/pipe.go:188:16: ErrDeadlineExceeded not declared by package os
/snap/go/current/src/net/pipe.go:204:17: ErrDeadlineExceeded not declared by package os

When I commented out the x509 package this error no longer occurred but unfortunately my program compiled and reported other errors

fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x470b493]

runtime stack:
runtime.throw(0x4ea9aba, 0x2a)
[...]

PR #1893, opened 2021-05-19
Add a stub for os.ReadDir()

fgsch:
This is needed by crypto/x509 in some configuration.
Related to #1888.


deadprogram:
Looks good to me. Thanks @fgsch now merging.

PR #1894, opened 2021-05-19
gdb: Use "extended-remote" instead of "remote"

ysoldak:
My gdb complains "remote" command is deprecated
On top of that "extended-remote" allows "disconnect" command that enables attaching from another debug client, like an IDE (I use VSCode)
See https://sourceware.org/gdb/current/onlinedocs/gdb/Connecting.html

### DIFF
--- a/main.go
+++ b/main.go
@@ -434,7 +434,7 @@ func FlashGDB(pkgName string, ocdOutput bool, options *compileopts.Options) erro
 		case "native":
 			// Run GDB directly.
 		case "openocd":
-			gdbCommands = append(gdbCommands, "target remote :3333", "monitor halt", "load", "monitor reset halt")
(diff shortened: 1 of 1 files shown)


aykevl:
This looks reasonable. However, can you also update all the other GDB commands in a similar way? I've tested most of them with `extended-remote` and it looks like they all work just fine.

ysoldak:
> This looks reasonable. However, can you also update all the other GDB commands in a similar way? I've tested most of them with `extended-remote` and it looks like they all work just fine.

Thanks, will do!

ysoldak:
Fixed for all GDB commands, @aykevl please check

aykevl (approved):
Yes, looks good to me. Thank you for the update!

PR #1895, opened 2021-05-20
Add support for http.Cookie and http.Header

fgsch:
Along with a few more things such as HTTP statuses.


aykevl:
Why do you want to reimplement the HTTP package? That seems like a very complex package that you wouldn't want to reimplement. Unlike the net package, the http package doesn't have a lot of hardware specific dependencies.

fgsch:
Superseded by #1911 and friends. Closing.

PR #1896, opened 2021-05-20
cgo: make const parser capable of prefix and infix expressions

aykevl:
I converted the CGo const parser to a Pratt parser following the book "Writing An Interpreter In Go" by Thorsten Ball. It is now capable of parsing some more numeric constants, see cgo/const_test.go for some expressions that can now be parsed.

We might need to switch to something different eventually. The trouble is that C is much more flexible with constants than Go. For example, the `!x` prefix automatically converts the value `x` to a boolean while Go doesn't do such a thing.

See https://github.com/tinygo-org/bluetooth/pull/65 for an example where this PR would be useful.

### DIFF
--- a/cgo/const.go
+++ b/cgo/const.go
@@ -11,14 +11,30 @@ import (
 	"strings"
 )
 
+var prefixParseFns map[token.Token]func(*tokenizer) (ast.Expr, *scanner.Error)
+
+func init() {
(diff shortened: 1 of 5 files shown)


deadprogram:
Thanks for working on this @aykevl now merging.

Issue #1887, opened 2021-05-17
Compiling ANTLR with tinygo returning Function return type does not match operand type of return inst!


jjeb:
Hi, I am trying to compile [ANTLR4](https://github.com/antlr/antlr4/blob/master/doc/index.md) using tinyGo, I am fully aware that there is some challenges so I am not really looking for a solution but to understand a little bit more about the error so I can try to figure it out: 

```
▶ tinygo version
tinygo version 0.18.0 darwin/amd64 (using go version go1.16 and LLVM version 11.0.0)

github.com/tinygo-antlr  tinygo18_single_import ✗   
▶ tinygo build -o fe-test/public/main.wasm -target wasm main.go
Function return type does not match operand type of return inst!
  ret %"github.com/antlr/antlr4/runtime/Go/antlr.ATN.187"* %3, !dbg !27382
 %"github.com/antlr/antlr4/runtime/Go/antlr.ATN"*Function return type does not match operand type of return inst!
  ret %"github.com/antlr/antlr4/runtime/Go/antlr.ATN.187"* %5, !dbg !27382
 %"github.com/antlr/antlr4/runtime/Go/antlr.ATN"*Function return type does not match operand type of return inst!
  ret %"github.com/antlr/antlr4/runtime/Go/antlr.ATN.187"* %9, !dbg !27382
 %"github.com/antlr/antlr4/runtime/Go/antlr.ATN"*Function return type does not match operand type of return inst!
  ret %"github.com/antlr/antlr4/runtime/Go/antlr.ParserATNSimulator.206"* %3, !dbg !27311
 %"github.com/antlr/antlr4/runtime/Go/antlr.ParserATNSimulator"*Function return type does not match operand type of return inst!
  ret %"github.com/antlr/antlr4/runtime/Go/antlr.ParserATNSimulator.206"* %5, !dbg !27311
 %"github.com/antlr/antlr4/runtime/Go/antlr.ParserATNSimulator"*Function return type does not match operand type of return inst!
  ret %"github.com/antlr/antlr4/runtime/Go/antlr.ParserATNSimulator.206"* %9, !dbg !27311
 %"github.com/antlr/antlr4/runtime/Go/antlr.ParserATNSimulator"*Function return type does not match operand type of return inst!
  ret %"github.com/antlr/antlr4/runtime/Go/antlr.ParserATNSimulator.206"* %10, !dbg !27311
[...]

PR #1898, opened 2021-05-20
transform: work around renamed return type after merging LLVM modules

aykevl:
This fix is very similar to https://github.com/tinygo-org/tinygo/pull/1768, but now for the return type. It fixes the issue in https://github.com/tinygo-org/tinygo/issues/1887.

Like #1768, I'm not sure how to test this as it is very specific to certain renames that LLVM does and that don't seem very reproducable.

### DIFF
--- a/transform/interface-lowering.go
+++ b/transform/interface-lowering.go
@@ -578,7 +578,7 @@ func (p *lowerInterfacesPass) getInterfaceMethodFunc(itf *interfaceInfo, signatu
 			paramTypes = append(paramTypes, param.Type())
 		}
 		calledFunctionType := function.Type()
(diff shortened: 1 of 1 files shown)


deadprogram:
I do not understand the implications of this change sufficiently, even though it is just one line. Are we sure this is safe?

aykevl:
Yes, I'm pretty sure it is :)
Also, this continues to be a problem (see https://github.com/tinygo-org/tinygo/issues/1887#issuecomment-894928366 and https://github.com/tinygo-org/tinygo/issues/2197).

niaow (changes requested):
I think this resolves the problem with this specific bug, although it looks like we may have another similar bug in another pass. Can we add a test to the transform first (put something into the `testdata/interface.ll`)?

niaow (approved):
Nevermind, I actually see what you mean now. Sorry. LGTM.

deadprogram:
Thanks @aykevl for the work-around and @niaow for reviewing. Now merging.

Issue #1879, opened 2021-05-13
[Question]tinygo build wasm error when use protobuf


zhenjunMa:
### Question

when i use struct compiled by protobuf, and use tinygo build to a wasm file,  i got an error.

### How to reproduce

model.proto
```
syntax = "proto3";

package model;

message Request {
	string name = 1;
}
```

build command
`protoc -I. --go_out=. model.proto`

hello.go
```
package main

import (
	"TinyGoProtoBuf/model"
)

func main() {
	_ = &model.Request{
		Name: "a",
	}
}

```
build command:
`tinygo build -o wasm.wasm -target wasm ./hello.go`

error:
![image](https://user-images.githubusercontent.com/28104581/118070017-c0dd4800-b3d7-11eb-8b07-4f014d66af62.png)

i have uploaded the total project: https://github.com/zhenjunMa/TinyGoProtoBuf

any idea about how to solve this?   or tinygo can not support protobuf now.

zhenjunMa:
hello.go

```
package main

import (
	"encoding/json"
	"fmt"
)

type request struct {
	name string `json:"name"`
}

func main() {
	req := &request{
		name: "hello",
	}

	b, _ := json.Marshal(req)

	fmt.Println(len(b))

}

```

i tried to use json like above, but still got an error:

![image](https://user-images.githubusercontent.com/28104581/118084832-f09a4900-b3f3-11eb-98e5-4610749849c5.png)

so, if i can use any other way to serialize a struct ?

dengliangjun:
I met those compiled errors as the same with yours
You should update tinygo version to 0.18.
However, Maybe you will meet new errors as the same with mine.
i.e. 
xxxx/protobuf/proto/clone.go : AddpendSlice not declared by package reflect
xxxx/protobuf/proto/discard.go : t.FieldByName undefined ( type reflect.Type has no field or method FieldByName)
xxxx......
xxxx.....

If you solve this problem, Please tell me. we can solve this problem jointly.
email: dengliangjun@skycto.com

aykevl:
Some of these issues are fixed in the latest release. Can you try again with the latest release?

dengliangjun:
> Some of these issues are fixed in the latest release. Can you try again with the latest release?

Hi, I need your help, My Env :
tinygo=0.18.0
go=1.13
target=wasm
my source code like ：
struct Order {
    Name string ‘json:name’
    Price int 'json:price'
}
[...]

PR #1902, opened 2021-05-22
Implement reflect.AppendSlice

aykevl:
This implementation basically just calls through to the runtime implementation of the `append` built-in (`runtime.sliceAppend`) with a few extra checks to ensure type safety.

See: https://github.com/tinygo-org/tinygo/issues/1879#issuecomment-844650456


deadprogram:
Very cool! :heavy_plus_sign: 

Now merging.

PR #1903, opened 2021-05-23
os: add FileMode constants from Go 1.16

sago35:
If you try to build the following code, you will get an error.
In this PR, we will add the constants needed for the above.
It is completely copied from the Go 1.16 source code.

https://github.com/tinygo-org/drivers/pull/265

before:
```
$ tinygo flash --target wioterminal --size short --programmer cmsis-dap ./examples/sdcard/tinyfs/
# github.com/tinygo-org/tinyfs/fatfs
..\..\..\..\pkg\mod\github.com\tinygo-org\tinyfs@v0.0.0-20210514090915-924e60a7bcf8\fatfs\go_fatfs.go:163:11: ModeDir not declared by package os
```

### DIFF
--- a/src/os/file_go_116.go
+++ b/src/os/file_go_116.go
@@ -83,3 +83,31 @@ func WriteFile(name string, data []byte, perm FileMode) error {
 	}
 	return err
 }
+
+// The defined file mode bits are the most significant bits of the FileMode.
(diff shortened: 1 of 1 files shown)


deadprogram:
Thanks for the update @sago35 now merging.

PR #1905, opened 2021-05-25
Various fixes to starting a goroutine

aykevl:
This PR is a collection of related but more or less independent changes to starting a new goroutine. It includes:

  * Some refactoring
  * IR tests for starting a new goroutine
  * Starting a builtin function as a goroutine (useless but now such code won't cause a panic)
  * Two small optimizations, one for closures in general and one specifically for starting a goroutine

I was investigating how to add support for code like this:

```go
func foo(itf fmt.Stringer) {
    go itf.String()
}
```

Which currently results in an error. Unfortunately it is a bit harder to implement than I expected and might need some more refactoring to implement cleanly. But I decided to create a PR for most small fixes leading up to that so these fixes won't need to wait any longer.


compiler/goroutine.go
```diff
@@ -5,25 +5,110 @@ package compiler
...
+		default:
+			panic("StaticCallee returned an unexpected value")
+		}
+		if !context.IsNil() {
+			params = append(params, context) // context parameter
+			hasContext = true
+		}
+		funcPtr = b.getFunction(callee)
+	} else if builtin, ok := instr.Call.Value.(*ssa.Builtin); ok {
+		// We cheat. None of the builtins do any long or blocking operation, so
```
deadprogram, line 52: :1st_place_medal:

deadprogram:
A very useful collection of fixes! Thanks @aykevl now merging.

PR #1908, opened 2021-05-26
Do not disable interrupts on abort

### LINKED ISSUE #1889: Panic messages are not always printed out
Sometimes my program panics and crashes.

Tinygo runtime tries its best to notify over serial console where exactly problem was, but this message usually fails to be delivered. Probably system dies too soon, could we make sure it prints out the message fully before halting?

An example of a message would be really nice to see in the console: "goroutine stack overflow"
See https://github.com/tinygo-org/tinygo/blob/release/src/internal/task/task_stack.go#L43

I use Arduino Nano 33 IoT board.

ysoldak:
Allows panic messages to be printed fully into serial console.

I've been testing this tiny fix for several days now.
So far panic messages been delivered reliably to serial console w/o any adverse side effects.

Thanks Ayke for the hint where to look.

Fixes #1889

### DIFF
--- a/src/runtime/runtime_cortexm_abort.go
+++ b/src/runtime/runtime_cortexm_abort.go
@@ -7,9 +7,6 @@ import (
 )
 
 func abort() {
-	// disable all interrupts
-	arm.DisableInterrupts()
-
 	// lock up forever
 	for {
 		arm.Asm("wfi")


deadprogram:
This makes total sense to me. Of course the serial would stop without any interrupts!

Now merging, thanks for the fix @ysoldak

PR #1911, opened 2021-05-28
Add sync.NewCond

fgsch:
Required by net/http.

This part of a larger effort to get net/http working under TinyGo.

PR #1912, opened 2021-05-28
Add runtime stubs required for net/http

fgsch:
Continued from #1911.


deadprogram:
I do not know enough about this part of `runtime` to really comment.

sago35 (approved):
Looks good to me.

aykevl:
Looks reasonable to me. I'm sure I've written this code before but apparently never made a PR for it.

PR #1915, opened 2021-05-29
RP2040 ignore generated files

kenbell:
This aligns 2040 with other families


deadprogram:
Thanks @kenbell now merging.

PR #1916, opened 2021-05-29
scheduler: task.Data made 64bit to avoid overflow

kenbell:
This is to re-add the change that broke the build to increase size of task.Data (to avoid overflows):
* original add was here: https://github.com/tinygo-org/tinygo/commit/fd8938c6346010979992986b2a1b512449e5c866
* revert was here: https://github.com/tinygo-org/tinygo/commit/ed2db8a26d8b5f3fdb9bf6365002bc6eeb19c758

I can't reproduce locally (make test doesn't break for me), but I believe I've found the cause of the build break, that the compiler test data needs to be updated since the task struct is included in the IR matched against the compiler output.


kenbell:
I think this is ready for review.  I did see some panics in WebAssembly doing `make test` - but after cleaning all go and tinygo caches, those have gone away.

deadprogram:
This now seems good to me. Any comments from anyone else before this is merged?

ysoldak:
I believe, I saw overflows while debugging my Nano33 IoT. Do not know how to reproduce though.

aykevl:
This looks reasonable. I've wanted to try and avoid this but it seems better to do it anyway. Overflows are a really bad kind of bug.

PR #1919, opened 2021-05-30
nrf: avoid heap allocations

aykevl:
This is a PR with two unrelated commits that avoid unnecessary heap allocations.


deadprogram:
Perhaps we want a larger buffer at some point? 

In any case this PR does not address that question. Looks good, now merging.

aykevl:
The buffer is currently only used for the `Transfer` method, which only transfers a single byte at a time. The `Tx` method doesn't need this buffer as it gets the buffer from the caller. So no, I don't think the buffer needs to be any bigger.

PR #1920, opened 2021-05-31
increase stack size for access to sdcard

sago35:
I had to change default-stack-size to use drivers/sdcard.
As discussed in the following PR, change the configuration of the board with the built-in sdcard reader.

https://github.com/tinygo-org/drivers/pull/265#issuecomment-851433104


deadprogram:
:+1: thanks @sago35 now merging.

Issue #1339, opened 2020-08-31
Blue Pill Interrupt on GPIO input state change


olablt:
I have Blue Pill STM32f103C8T6

How to create Interrupts on 3 GPIO inputs, to fire after each state change?
Is it possible?

Here is an example Arduino code using interrupts. I am not sure how to do this with TinyGo
https://github.com/chrisalbertson/quadratureBluePill/blob/master/quadratureBluePill.h

ofauchon:
Hi. 

I can't find implementation of (Pin)SetInterrupt for STM32 in the Tinygo sources. 

So I guess GPIO interrupts are not yet available for STM32. 

However, you can find some example of low-level code to configure GPIO interrupts  : 

- on STM32L0: https://github.com/ofauchon/tinygo/blob/lgt92_examples/src/examples/lgt92/01_button_interrupt.go
- on STM32F103: https://github.com/ofauchon/experiments/blob/master/tinygo/bluepill-misc/test_gpio_int.go

Hope that helps

Olivier

kenbell:
Closing since GPIO interrupt support is merged

PR #1921, opened 2021-06-01
stm32: GPIO input interrupt support

kenbell:
Adds support for GPIO input interrupts using the same interface as other families of chips.

Also extended stm32f103 to support configurable pull-up / pull-down on input pins.

ST has moved the location of the EXTI config registers between 3 different places over time (AFIO, SYSCFG, EXTI).  I have called these 'reva', 'revb', 'revc' (similar to i2c) in absence of any better naming scheme.

Similarly, more recent models appear to have more EXTI channels, so two variants of the code to work with both older and more recent designs.

### DIFF
--- a/src/machine/machine_stm32_gpio_reva.go
+++ b/src/machine/machine_stm32_gpio_reva.go
@@ -1,0 +1,92 @@
+// +build stm32,!stm32l4,!stm32l5
+
+package machine
+
+import (
+	"device/stm32"
+)
+
+// This variant of the GPIO input interrupt logic is for
+// chips with a smaller number of interrupt channels
+// (that fits in a single register).
+
+//
+// STM32 allows one interrupt source per pin number, with
+// the same pin number in different ports sharing a single
+// interrupt source (so PA0, PB0, PC0 all share).  Only a
+// single physical pin can be connected to each interrupt
+// line.
+//
+// To call interrupt callbacks, we record here for each
+// pin number the callback and the actual associated pin.
+//
+
+// Callbacks for pin interrupt events
+var pinCallbacks [16]func(Pin)
+
+// The pin currently associated with interrupt callback
+// for a given slot.
+var interruptPins [16]Pin
+
+// SetInterrupt sets an interrupt to be executed when a particular pin changes
+// state. The pin should already be configured as an input, including a pull up
+// or down if no external pull is provided.
+//
+// This call will replace a previously set callback on this pin. You can pass a
+// nil func to unset the pin change interrupt. If you do so, the change
+// parameter is ignored and can be set to any value (such as 0).
+func (p Pin) SetInterrupt(change PinChange, callback func(Pin)) error {
+	port := uint32(uint8(p) / 16)
+	pin := uint8(p) % 16
+
+	enableEXTIConfigRegisters()
+
+	if callback == nil {
+		stm32.EXTI.IMR.ClearBits(1 << pin)
+		pinCallbacks[pin] = nil
+		return nil
+	}
+
+	if pinCallbacks[pin] != nil {
+		// The pin was already configured.
+		// To properly re-configure a pin, unset it first and set a new
+		// configuration.
+		return ErrNoPinChangeChannel
+	}
+
+	// Set the callback now (before the interrupt is enabled) to avoid
+	// possible race condition
+	pinCallbacks[pin] = callback
+	interruptPins[pin] = p
+
+	crReg := getEXTIConfigRegister(pin)
+	shift := (pin & 0x3) * 4
+	crReg.ReplaceBits(port, 0xf, shift)
+
+	if (change & PinRising) != 0 {
+		stm32.EXTI.RTSR.SetBits(1 << pin)
+	}
+	if (change & PinFalling) != 0 {
+		stm32.EXTI.FTSR.SetBits(1 << pin)
+	}
+	stm32.EXTI.IMR.SetBits(1 << pin)
+
+	intr := p.registerInterrupt()
+	intr.SetPriority(0)
+	intr.Enable()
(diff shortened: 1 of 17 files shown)


kenbell:
This should address #1339 - thanks @ofauchon for the code linked in that issue that was the basis for this change.

src/machine/machine_stm32_gpio_reva.go
```diff
@@ -0,0 +1,85 @@
...
+	intr.Enable()
+
+	return nil
+}
+
+func handlePinInterrupt(pin uint8) {
+	if stm32.EXTI.PR.HasBits(1 << pin) {
+		// Writing 1 to the pending register clears the
+		// pending flag
+		stm32.EXTI.PR.SetBits(1 << pin)
```
aykevl: SetBits seems incorrect here. It is a read-modify-write, so all other interrupts pending on EXTI will be cleared.

src/machine/machine_stm32_gpio_reva.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+//
+// STM32 allows one interrupt source per pin number, with
+// the same pin number in different ports sharing a single
+// interrupt source (so PA0, PB0, PC0 all share).  Only a
+// single physical pin can be connected to each pin number.
+//
+// To call interrupt callbacks, we record here for each
+// pin number the callback and the actual associated pin.
```
aykevl, line 21: Do I read this correctly that if you set an interrupt on `PA0`, and then one on `PB0`, that the interrupt on `PA0` is overwritten?

src/machine/machine_stm32_gpio_reva.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func handlePinInterrupt(pin uint8) {
+	if stm32.EXTI.PR.HasBits(1 << pin) {
+		// Writing 1 to the pending register clears the
+		// pending flag
+		stm32.EXTI.PR.SetBits(1 << pin)
+
+		callback := pinCallbacks[pin]
+		if callback != nil {
```
aykevl, line 88: This is racy, at least in the current implementation. A `func()` value is a doubleword, which is written in two operations. This means that one could be nil while the other isn't. Maybe the `SetInterrupt` method can be written in such a way that the callback is only ever modified when the pin interrupt is disabled? I've done that in other pin interrupt implementations. In that case, you also don't need to check whether the callback is nil (but it's better to do anyway so that the compiler won't insert a nil check).

aykevl:
> ST has moved the location of the EXTI config registers between 3 different places over time (AFIO, SYSCFG, EXTI). I have called these 'reva', 'revb', 'revc' (similar to i2c) in absence of any better naming scheme.

You could perhaps take a look at which naming scheme stm32-rs is using.

src/machine/machine_stm32_gpio_reva.go
kenbell: I'll correct the comment it's misleading - writing 1 for a bit clears that bit only, from the docs:

```
PRx: Pending bit
0: No trigger request occurred
1: selected trigger request occurred
This bit is set when the selected edge event arrives on the external interrupt line. This bit is
cleared by writing a ‘1’ into the bit. 
```

src/machine/machine_stm32_gpio_reva.go
kenbell: Yes - there's physically a mux that selects one (and only one) of PA0, PB0, PC0, PD0, etc
image

src/machine/machine_stm32_gpio_reva.go
kenbell: Thanks - will fix

src/machine/machine_stm32_gpio_reva.go
kenbell: Oh - hang-on, I get what you mean about read-modify-write now.  I need to use Set instead.

kenbell:
I've pushed a new change that:

- Fixes the SetBits bug
- Renames the 'reva', 'revb', 'revc' variants of exti to be _afio.go, _syscfg.go, _exti.go  (stm32rs doesnt really have any helpful names, so these at least seem more mnemonic)
- Fixes the race in SetInterrupt / handlePinInterrupt to set the handler function whilst the interrupt is not enabled

I've verified the changes on a bunch of boards.

I think this should be good to go.

src/machine/machine_stm32_gpio_reva.go
deadprogram: Has this been addressed?

deadprogram:
It appears all feedback has been addressed. I plan to merge later today.

deadprogram:
Now merging, thanks for another great contribution @kenbell

src/machine/machine_stm32_gpio_reva.go
aykevl: The API in all other implementations is that (from the API user perspective) every pin has a separate interrupt handler. With this implementation, setting an interrupt on pin PB0 would replace an interrupt on PA0, which is probably not what you'd expect given the current API.

aykevl: Also see this PR: https://github.com/tinygo-org/tinygo-site/pull/191

kenbell: With the latest changes, if you do something like this:
```go
machine.PA13.SetInterrupt(machine.PinChange,xxxx)
machine.PC13.SetInterrupt(machine.PinChange,xxxx)
```
You get an error: `could not configure pin interrupt: machine: no channel available for pin interrupt` on the second call to use slot 13.

Given the hardware constraints, it's probably the closest implementation to the interface?

PR #1922, opened 2021-06-01
all: add a flag to the command line to select the serial implementation

aykevl:
This can be very useful for some purposes:

  * It makes it possible to disable the UART in cases where it is not needed or needs to be disabled to conserve power.
  * It makes it possible to disable the serial output to reduce code size, which may be important for some chips. Sometimes, a few kB can be saved this way.
  * It makes it possible to override the default, for example you might want to use an actual UART to debug the USB-CDC implementation.

It also lowers the dependency on having machine.Serial defined, which is often not defined when targeting a chip. Eventually, we might want to make it possible to write `-target=nrf52` or `-target=atmega328p` for example to target the chip itself with no board specific assumptions.

The defaults don't change. I checked this by running `make smoketest XTENSA=0` before and after and comparing the results.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -430,6 +430,8 @@ endif
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=pca10040 -opt=1     examples/blinky1
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=pca10040 -serial=none examples/echo
+	@$(MD5SUM) test.hex
 	$(TINYGO) build             -o test.nro -target=nintendoswitch      examples/serial
 	@$(MD5SUM) test.nro
 	$(TINYGO) build -size short -o test.hex -target=pca10040 -opt=0     ./testdata/stdlib.go
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -55,7 +55,7 @@ func (c *Config) GOARCH() string {
 
 // BuildTags returns the complete list of build tags used during this build.
 func (c *Config) BuildTags() []string {
-	tags := append(c.Target.BuildTags, []string{"tinygo", "gc." + c.GC(), "scheduler." + c.Scheduler()}...)
+	tags := append(c.Target.BuildTags, []string{"tinygo", "gc." + c.GC(), "scheduler." + c.Scheduler(), "serial." + c.Serial()}...)
 	for i := 1; i <= c.GoMinorVersion; i++ {
 		tags = append(tags, fmt.Sprintf("go1.%d", i))
 	}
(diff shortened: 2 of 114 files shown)


kenbell (commented):
I see failure against STM32 (should this fail, or should it 'do nothing'):

```
$ tinygo flash -target bluepill -serial none ./src/examples/echo/
# github.com/tinygo-org/tinygo/src/examples/echo
/home/kenneth/work/tinygo/src/examples/echo/echo.go:19:7: uart.Write undefined (type machine.NullSerial has no field or method Write)
/home/kenneth/work/tinygo/src/examples/echo/echo.go:24:11: uart.Buffered undefined (type machine.NullSerial has no field or method Buffered)
/home/kenneth/work/tinygo/src/examples/echo/echo.go:25:20: uart.ReadByte undefined (type machine.NullSerial has no field or method ReadByte)
/home/kenneth/work/tinygo/src/examples/echo/echo.go:30:10: uart.Write undefined (type machine.NullSerial has no field or method Write)
/home/kenneth/work/tinygo/src/examples/echo/echo.go:31:10: uart.Write undefined (type machine.NullSerial has no field or method Write)
/home/kenneth/work/tinygo/src/examples/echo/echo.go:32:10: uart.Write undefined (type machine.NullSerial has no field or method Write)
/home/kenneth/work/tinygo/src/examples/echo/echo.go:33:10: uart.Write undefined (type machine.NullSerial has no field or method Write)
```

sago35:
I would like to define DefaultUART in atsamd21/atsamd51 as well.
In the current state, this will result in the following error, but in most cases it is sufficient to specify UART0 for the first time.

```
$ tinygo flash --target xiao --serial uart --programmer cmsis-dap --size short ./src/examples/echo
# machine
..\..\Users\sago3\AppData\Local\tinygo\goroot-go1.16-641d833c9cde99e59ff7f49213b86411a51f7b285c3bdc52fcd17d2b43471339-syscall\src\machine\serial-uart.go:6:14: undeclared name: DefaultUART
```

deadprogram:
Merge conflicting needs resolution @aykevl if you please.

aykevl:
> I see failure against STM32 (should this fail, or should it 'do nothing'):

Not sure. We could add the missing (dummy) methods, sure. On the other hand, maybe it's better to fail because this is not a real serial port.

> I would like to define DefaultUART in atsamd21/atsamd51 as well.
> In the current state, this will result in the following error, but in most cases it is sufficient to specify UART0 for the first time.

If the board has an UART on the silkscreen, then yes this can be added. But I don't agree it should be added for all samd21/samd51, only for the boards that have TX/RX on the silkscreen.

aykevl:
Updated to resolve the merge conflicts. I've verified that the produced binaries are still identical with the default serial implementation.

Regarding the samd21/samd51: I think this is best done as a separate PR. I've found that they don't define the UART instance yet, so it's not as simple as defining `DefaultUART`.

deadprogram:
> On the other hand, maybe it's better to fail because this is not a real serial port.

IMO the least surprising thing to happen when `serial=none` would be to disable all output, if there is a serial port or not.

aykevl:
I've updated the PR with some more dummy methods so that examples/echo can be compiled (but of course it won't show any output).

deadprogram:
This is for sure needed for field-deployed devices, along with other purposes you've mentioned.

We might want to put a warning in the docs for people who use `-serial=none` on a device, and then cannot figure out why the auto-reset for `tinygo flash` isn't working anymore.

In any case, worked for the tests I made, and is passing TinyHCI, so merging.

Thanks everyone who helped with this one!

PR #1913, opened 2021-05-28
Add os stubs required for net/http

fgsch:
Continued from #1912.

PR #1923, opened 2021-06-01
Add reflect stubs required for net/http

fgsch:
Continued from #1913.

### DIFF
--- a/src/reflect/makefunc.go
+++ b/src/reflect/makefunc.go
@@ -1,5 +1,5 @@
 package reflect
 
 func MakeFunc(typ Type, fn func(args []Value) (results []Value)) Value {
-	return Value{}
+	panic("unimplemented: reflect.MakeFunc()")
 }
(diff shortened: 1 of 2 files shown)


src/reflect/makefunc.go
```diff
@@ -0,0 +1,5 @@
+package reflect
+
+func MakeFunc(typ Type, fn func(args []Value) (results []Value)) Value {
+	return Value{}
```
aykevl: I do not think returning `Value{}` is correct here. For example, `reflect.MakeFunc(...).Kind()` would return `reflect.Invalid` instead of `reflect.Func`. Therefore, I think this should panic instead.

fgsch: Yeah, I wasn't sure about this. Thanks for pointing it out. I will update it.

fgsch: Pushed.

PR #1925, opened 2021-06-02
main: don't consider compile-only tests as failing

aykevl:
Previously a command like the following would incorrectly print FAIL:

    tinygo test -c math

This commit fixes this issue by defaulting to a passing test (the test is marked as passed if it isn't run).
Thanks to @ysoldak for finding this bug!


deadprogram:
:heavy_check_mark: now merging.

PR #1926, opened 2021-06-02
Fix testing main packages

aykevl:
Reported by @ysoldak.

This PR fixes testing the main package. Before this PR, it would show an internal error like this:

    $ tinygo test -target arduino-nano33 -c ./tinytest/
    error: package github.com/foo/bar/tinytest.test imports main but couldn't find dependency


ysoldak:
Verified, works for me!

deadprogram:
Thanks @aykevl for the fix and @ysoldak for the review.

Now merging.

PR #1927, opened 2021-06-02
main: make flash-command portable and safer to use

aykevl:
Previously, flash-command would assume it could execute a command straight via /bin/sh, at least on non-Windows systems. Otherwise it would just split the command using `strings.Split`. This is all a bit hacky, so I've replaced it with a proper solution: splitting the command _before_ substituting various paths using a real shell splitter (shlex.Split, from Google). This solves a few things:

  * It guards against special characters in path names. This can be an issue on Windows where the temporary path may contain spaces (this is uncommon on POSIX systems).
  * It is more portable, by disallowing the use of a shell. That way, it doesn't differentiate between Windows and non-Windows anymore.

This depends on https://github.com/tinygo-org/tinygo/pull/1767 otherwise the PCA10059 board can no longer be flashed.

Tested on an Arduino Nano (avrdude), Arduino Nano 33 (bossac), and DigiSpark (micronucleus).

### DIFF
--- a/main.go
+++ b/main.go
@@ -314,8 +314,10 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 		case "", "command":
 			// Create the command.
 			flashCmd := config.Target.FlashCommand
-			fileToken := "{" + fileExt[1:] + "}"
-			flashCmd = strings.ReplaceAll(flashCmd, fileToken, result.Binary)
+			flashCmdList, err := shlex.Split(flashCmd)
+			if err != nil {
+				return fmt.Errorf("could not parse flash command %#v: %w", flashCmd, err)
+			}
 
 			if strings.Contains(flashCmd, "{port}") {
 				var err error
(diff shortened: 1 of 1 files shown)


sago35 (approved):
I think it's very good.

deadprogram:
#1767 was replaced by #1931 which was merged into `dev`. So this PR can now be continued.

aykevl:
Rebased and un-drafted so this is now ready for review.

main.go
```diff
@@ -314,8 +314,10 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 		case "", "command":
 			// Create the command.
 			flashCmd := config.Target.FlashCommand
-			fileToken := "{" + fileExt[1:] + "}"
-			flashCmd = strings.ReplaceAll(flashCmd, fileToken, result.Binary)
+			flashCmdList, err := shlex.Split(flashCmd)
```
deadprogram, line 317: :+1: for `shlex`

deadprogram:
Looks good, so now merging. Thanks for this @aykevl

Issue #1910, opened 2021-05-28
windows issue: unable to flash arduino nano33 


loreaman:
``` 
tinygo flash -target arduino-nano33 .\step0

bossac: extra arguments found
Try 'bossac -h' or 'bossac --help' for more information
error: failed to flash C:\Users\Lorenzo Amante\AppData\Local\Temp\tinygo006274015\main.bin: exit status 1
```

sago35:
If you add the `-x` option as follows, how does it appear?

```
$ tinygo flash -target arduino-nano33 -x .\step0
```

loreaman:
Hi, thank you for the reply. Unfortunately I am getting again the error. I will paste the output here:
```
C:\Users\Lorenzo Amante\Documents\PROJECTS\TINYGO\gophercon-2019\sensor\arduino>tinygo flash -target arduino-nano33 -x .\step0
ld.lld --emit-relocs --gc-sections -L C:\Users\Lorenzo Amante\scoop\apps\tinygo\current -T targets/atsamd21.ld -o C:\Users\Lorenzo Amante\AppData\Local\Temp\tinygo270044323\main C:\Users\Lorenzo Amante\AppData\Local\Temp\tinygo270044323\main.o C:\Users\Lorenzo Amante\scoop\apps\tinygo\current\pkg\armv6m-none-eabi\compiler-rt.a C:\Users\Lorenzo Amante\scoop\apps\tinygo\current\pkg\armv6m-none-eabi\picolibc.a C:\Users\Lorenzo Amante\AppData\Local\tinygo\obj-9ae6912157365e41ef4601bc6ddfde4fb334865ec0b3445a142a6c0f.o C:\Users\Lorenzo Amante\AppData\Local\tinygo\obj-7d1f71c6e0b8e3e91c6e0bc9d1b3b8b6257cb2b7ccddcb662876d57a.o C:\Users\Lorenzo Amante\AppData\Local\tinygo\obj-789b0940e03ceb173eba2c87acf8891d566caff16c9621f0ebdf6957.o C:\Users\Lorenzo Amante\AppData\Local\tinygo\obj-029fbf7e06de2115289d368f10f043505bb949b6f059b5f5b18312c9.o
bossac -i -e -w -v -R -U --port=COM12 --offset=0x2000 C:\Users\Lorenzo Amante\AppData\Local\Temp\tinygo270044323\main.bin
bossac: extra arguments found
Try 'bossac -h' or 'bossac --help' for more information
[...]

PR #1928, opened 2021-06-02
main: escape commands while printing them with the -x flag

aykevl:
This should make issues such as the one in https://github.com/tinygo-org/tinygo/issues/1910 more obvious.


sago35 (approved):
Looks good to me.
The commands shown are not directly executable, especially on windows, but they are easier to understand.

deadprogram:
:100: merging, thank you!

PR #1933, opened 2021-06-05
runtime: expose heap stats

ysoldak:
Only stats about heap for now.

Zero effect on runtime and compiled size when `runtime.ReadMemStats` not used in application code (expected to be optimised out).

Example included.

Sample output (from Arduino Nano 33 IoT):
```
Doing  56  iterations
Heap before GC. Used:  6528  Free:  16656  Meta:  368
Heap after  GC. Used:  1056  Free:  22128  Meta:  368
Doing  0  iterations
Heap before GC. Used:  1056  Free:  22128  Meta:  368
Heap after  GC. Used:  1056  Free:  22128  Meta:  368
Doing  94  iterations
Heap before GC. Used:  15568  Free:  7616  Meta:  368
Heap after  GC. Used:  1056  Free:  22128  Meta:  368

```

### DIFF
--- a/src/examples/memstats/memstats.go
+++ b/src/examples/memstats/memstats.go
@@ -1,0 +1,32 @@
+package main
+
+import (
+	"math/rand"
+	"runtime"
+	"time"
+)
+
+func main() {
+
+	ms := runtime.MemStats{}
+
+	for {
+		escapesToHeap()
+		runtime.ReadMemStats(&ms)
+		println("Heap before GC. Used: ", ms.HeapInuse, " Free: ", ms.HeapIdle, " Meta: ", ms.GCSys)
+		runtime.GC()
+		runtime.ReadMemStats(&ms)
+		println("Heap after  GC. Used: ", ms.HeapInuse, " Free: ", ms.HeapIdle, " Meta: ", ms.GCSys)
+		time.Sleep(5 * time.Second)
+	}
+
+}
+
+func escapesToHeap() {
+	n := rand.Intn(100)
+	println("Doing ", n, " iterations")
+	for i := 0; i < n; i++ {
+		s := make([]byte, i)
+		_ = append(s, 42)
+	}
+}
--- a/src/runtime/mstats.go
+++ b/src/runtime/mstats.go
@@ -1,0 +1,69 @@
+// +build gc.conservative
+
+package runtime
+
+// Memory statistics
+
+// Subset of memory statistics from upstream Go.
+// Works with conservative gc only.
+
+// A MemStats records statistics about the memory allocator.
+type MemStats struct {
+	// General statistics.
+
+	// Sys is the total bytes of memory obtained from the OS.
+	//
+	// Sys is the sum of the XSys fields below. Sys measures the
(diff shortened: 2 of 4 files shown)


ysoldak:
I've kept `uint64` to be API-compatible with upstream GO, in our case uint shall be enough, of course.

src/runtime/mstats.go
```diff
@@ -0,0 +1,44 @@
...
+	// Heap memory statistics.
+
+	// HeapSys is bytes of heap memory, total.
+	HeapSys uint64
+
+	// HeapIdle is bytes in idle (unused) blocks.
+	HeapIdle uint64
+
+	// HeapInuse is bytes in in-use blocks.
+	HeapInuse uint64
```
aykevl, line 35: Right, I think I see your reasoning. It counts all blocks, which may be bigger than the objects themself (which is somewhat like heap spans in upstream Go).
Seems reasonable.

src/runtime/mstats.go
```diff
@@ -0,0 +1,44 @@
...
+package runtime
+
+// Memory statistics
+
+// Subset of memory statistics from upstream Go.
+// Works with conservative gc only.
+
+// A MemStats records statistics about the memory allocator.
+type MemStats struct {
+	// Heap memory statistics.
```
aykevl, line 21: You could possibly add:

- `Sys`:  the total allocated memory: `heapEnd - heapStart`. Technically this should maybe also include the scheduler stack, but I think it's fine leaving it out for now (it is not easy to obtain on POSIX, for example).
- `HeapReleased`: always 0, we don't currently release memory back to the OS.

src/examples/memstats/memstats.go
```diff
@@ -0,0 +1,32 @@
+package main
```
aykevl, line 1: This example should be included in the Makefile (near the top of the `smoketest` target, with the other example smoke tests).

aykevl (commented):
I think it would perhaps make more sense to put the code that reads the memory statistics inside gc_conservative.go, as it is specific to that GC. Perhaps even the entire mstats.go file. But that's just a suggestion, overall I think this PR is really nice and clean.

> I've kept `uint64` to be API-compatible with upstream GO, in our case uint shall be enough, of course.

Yes, this is a good idea.

src/examples/memstats/memstats.go
ysoldak: ✅

src/runtime/mstats.go
ysoldak: ✅

ysoldak:
Anything I can do to improve this PR to be accepted?

deadprogram:
Since all feedback items have been addressed, I am going to merge now. 

Thanks @ysoldak

deadprogram:
Actually I will squash PR first @ysoldak unless you want to do so yourself?

ysoldak:
Squashed and rebased 👍

deadprogram:
Now merging, thanks for working on this @ysoldak :bar_chart:

PR #1936, opened 2021-06-06
Prototype: rp2040: two-pass link to build 2nd stage boot

kenbell:
This is a prototype of using two linker passes to generate and insert the RP2040 stage 2 boot loader into the binary.


kenbell:
Consensus is that #1937 is the better approach, so closing this PR.

PR #1937, opened 2021-06-07
rp2040: assemble / checksum 2nd stage bootloader

kenbell:
This prototype assembles the RP2040 second stage bootloader from source and uses binary patching to inject the CRC after linking

### DIFF
--- a/builder/elfpatch.go
+++ b/builder/elfpatch.go
@@ -1,0 +1,57 @@
+package builder
+
+import (
+	"debug/elf"
+	"fmt"
+	"os"
+)
+
+func getElfSectionData(executable string, sectionName string) ([]byte, elf.FileHeader, error) {
+	elfFile, err := elf.Open(executable)
+	if err != nil {
+		return nil, elf.FileHeader{}, err
+	}
+	defer elfFile.Close()
+
+	section := elfFile.Section(sectionName)
+	if section == nil {
+		return nil, elf.FileHeader{}, fmt.Errorf("could not find %s section", sectionName)
+	}
+
+	data, err := section.Data()
+
+	return data, elfFile.FileHeader, err
+}
+
+func replaceElfSection(executable string, sectionName string, data []byte) error {
+	fp, err := os.OpenFile(executable, os.O_RDWR, 0)
+	if err != nil {
+		return err
+	}
+	defer fp.Close()
+
+	elfFile, err := elf.Open(executable)
+	if err != nil {
+		return err
+	}
+	defer elfFile.Close()
+
+	section := elfFile.Section(sectionName)
+	if section == nil {
+		return fmt.Errorf("could not find %s section", sectionName)
+	}
+
+	// Implicitly check for compressed sections
+	if section.Size != section.FileSize {
+		return fmt.Errorf("expected section %s to have identical size and file size, got %d and %d", sectionName, section.Size, section.FileSize)
+	}
+
+	// Only permit complete replacement of section
+	if section.Size != uint64(len(data)) {
+		return fmt.Errorf("expected section %s to have size %d, was actually %d", sectionName, len(data), section.Size)
+	}
+
+	// Write the replacement section data
+	_, err = fp.WriteAt(data, int64(section.Offset))
+	return err
+}
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -40,6 +40,7 @@ type TargetSpec struct {
 	LDFlags          []string `json:"ldflags"`
 	LinkerScript     string   `json:"linkerscript"`
 	ExtraFiles       []string `json:"extra-files"`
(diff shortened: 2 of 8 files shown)


targets/pico_boot_stage2.S
```diff
@@ -1,23 +1,420 @@
...
+//
+// This implementation has been made 'stand-alone' by including necessary code /
+// symbols from the included files in the reference implementation directly into
+// the source.  Care has been taken to preserve ordering and it has been verified
+// the generated binary is byte-for-byte identical to the reference code binary.
+//
+// Note: the stage 2 boot loader must be 256 bytes in length and have a checksum
+// present.  In TinyGo, the linker script is responsible for allocating 256 bytes
+// for the .boot2 section and the target JSON doc indicates a two-pass link is
+// required to generate and inject the checksum.
```
aykevl: This isn't really true for this PR.

targets/pico_boot_stage2.S
```diff
@@ -1,23 +1,420 @@
...
+
+    pop {r1, pc}
+
+#endif
+
+.global literals
+literals:
+.ltorg
+
+.end
```
aykevl: Looks like this is missing a newline at the end of the file. Many editors insert it automatically, but not all.

compileopts/target.go
```diff
@@ -40,6 +40,7 @@ type TargetSpec struct {
 	LDFlags          []string `json:"ldflags"`
 	LinkerScript     string   `json:"linkerscript"`
 	ExtraFiles       []string `json:"extra-files"`
+	ElfPatches       []string `json:"elf-patches"`              // Patches that should be applied to the ELF after linking
```
aykevl: I'm not entirely convinced this level of abstraction is the right level. While we might add more patches in the future, I think a simple boolean (like `automatic-stack-size`, for example `rp2040-boot-patch`) would be simpler to understand and modify. It's always possible to add a slice of patches in the future.

Alternatively, you could leave out a change to the target JSON out entirely and instead use a specially named section (like `.rp2040_boot2`) which will be modified if present.

builder/elfpatch.go
```diff
@@ -0,0 +1,57 @@
+package builder
+
+import (
+	"debug/elf"
+	"fmt"
+	"os"
+)
+
+func getElfSectionData(executable string, sectionName string) ([]byte, elf.FileHeader, error) {
+	elfFile, err := elf.Open(executable)
```
aykevl, line 10: You might consider moving the opening of ELF files to some place earlier, so that it is only opened once instead of for each call to `getElfSectionData` and `replaceElfSection`. Optional, the current code looks good and is certainly easy to understand/modify.

aykevl (commented):
I quite like this PR, more than the two-pass link prototype. The reason is that this is much more modular and doesn't depend on a particular build mode. For example, we might want to use ThinLTO in the future which might interfere with the two-pass link (by making the link slower, for example).

The build failure is probably just a flaky build, trying again should fix it.

targets/pico_boot_stage2.S
kenbell: thanks - i've updated the comment:
```
// Note: the stage 2 boot loader must be 256 bytes in length and have a checksum
// present.  In TinyGo, the linker script is responsible for allocating 256 bytes
// for the .boot2 section and the build logic patches the checksum into the
// binary after linking, controlled by the '<target>.json' flag 'rp2040-boot-patch'.
```

targets/pico_boot_stage2.S
kenbell: fixed

compileopts/target.go
kenbell: I'm less certain about special section names - means the build logic has to open the elf for all targets to see if the section exists?  For now I've gone with the suggestion of `rp2040-boot-patch` in target spec.

builder/elfpatch.go
kenbell: Yeah - i was undecided about 'going too far' on something that wraps up `debug/elf` with editing capability.  I'll leave it like it is for now and can make the wrapper more complete later if needed.

kenbell:
> I quite like this PR, more than the two-pass link prototype. The reason is that this is much more modular and doesn't depend on a particular build mode. For example, we might want to use ThinLTO in the future which might interfere with the two-pass link (by making the link slower, for example).

Agree - i think this is the better approach, I'll drop the other PR.

kenbell:
I think this is ready for review - it doesn't add support for feather-rp2040, but is a good stand-alone step to avoid including the stage2 boot loader as a binary blob.

The 2nd stage should be byte-for-byte identical to the previous binary blob.

deadprogram:
Tested on my pico and worked perfectly.

aykevl:
Thank you for the update! Looks good to me.

Issue #1918, opened 2021-05-30
Can NEOPIXEL(S) name be unified?


sago35:
`NEOPIXEL(S)` is now defined with the following name
[drivers/examples/ws2812](https://github.com/tinygo-org/drivers/blob/release/examples/ws2812/others.go#L9) expects the name `NEOPIXELS`.
So we would like to unify these names.
Even if there is only one `NEOPIXEL`, we want to use the name `NEOPIXELS`.

If you want to ignore compatibility, you can change them all to `NEOPIXELS`.
If you want to keep compatibility, you can set an additional name `NEOPIXELS`.

If anyone has a better idea, please let me know.

```
$ git grep -i NEOPIXEL
src/machine/board_circuitplay_bluefruit.go:40:  NEOPIXELS = D8
src/machine/board_circuitplay_express.go:43:    NEOPIXELS = D8
src/machine/board_clue_alpha.go:74:     NEOPIXEL = D18
src/machine/board_feather-m4-can.go:20: D7  = PB03 // neopixel power
src/machine/board_feather-m4-can.go:21: D8  = PB02 // built-in neopixel
src/machine/board_feather-m4-can.go:46: NEOPIXELS = D8
src/machine/board_feather-m4-can.go:127:        // turn on neopixel
src/machine/board_feather-m4.go:15:     D8  = PB03 // built-in neopixel
src/machine/board_feather-nrf52840.go:17:       D8  = P0_16 // NeoPixel
src/machine/board_feather-nrf52840.go:61:       NEOPIXEL = D8
src/machine/board_feather-stm32f405.go:26:      D8  = PC0  // GPIO, Neopixel
src/machine/board_feather-stm32f405.go:81:      NUM_BOARD_NEOPIXEL = 1
src/machine/board_feather-stm32f405.go:84:      LED_NEOPIXEL = D8
src/machine/board_grandcentral-m4.go:97:        D88 = PC24 // On-board NeoPixel
src/machine/board_grandcentral-m4.go:135:       NEOPIXEL_PIN    = D88 // (PC24)
src/machine/board_grandcentral-m4.go:141:       NEOPIXEL = NEOPIXEL_PIN
src/machine/board_matrixportal-m4.go:16:        D4  = PA23  // NeoPixel                       EXTI7
src/machine/board_matrixportal-m4.go:77:        NEOPIXEL = D4
src/machine/board_metro-m4-airlift.go:26:       D40 = PB22 // built-in neopixel
[...]

PR #1940, opened 2021-06-07
machine: add definition for ws2812

sago35:
This PR adds WS2812 to the pin definition if there is a ws2812 on the board.
This makes the code for `drivers/examples/ws2812` identical.

#1918


deadprogram:
:traffic_light: now merging.

PR #1932, opened 2021-06-05
machine/rp2040: add basic support for ADC

deadprogram:
This PR adds support for the RP2040's ADC to be able to read analog data. Tested on my real RP2040 Pico with the code in `examples/adc`.

PR #1942, opened 2021-06-07
board/nano-rp2040: pins and blinking led

ysoldak:
All info I could find about pins on **Arduino Nano RP2040 Connect**.

Flashed successfully and verified with *blinky1*.

Boot procedure seem to be identical to **Pico**.

Putting it into mass storage mode for flashing is kind of pain: must short `REC` and `GND` pins before plugging in.

There is code in ArduinoCore-mbed that implements switch to flashing mode by double tap. I can't make use of it, hopefully someone smarter than me could fix this for TinyGo, would be great help.
https://github.com/arduino/ArduinoCore-mbed/blob/master/variants/NANO_RP2040_CONNECT/double_tap_usb_boot.cpp

P.S. Please advise on better name for the target if you feel `arduino-nano-rp2040-connect` is just too long.
On the other hand `pico` is too short. I'd vote for renaming it to `rpi-pico` or `rpi-rp2040-pico`, while it's still not too late?

### DIFF
--- a/src/machine/board_nano-rp2040.go
+++ b/src/machine/board_nano-rp2040.go
@@ -1,0 +1,99 @@
+// +build nano_rp2040
+
+// This contains the pin mappings for the Arduino Nano RP2040 Connect board.
+//
+// Sometimes the board is not detected even when the board is connected to your computer.
+// To solve this, place a jumper wire between the REC and GND pins, then connect the board to your computer.
+//
+// For more information, see: https://store.arduino.cc/nano-rp2040-connect
+// Also
+// - Datasheets: https://docs.arduino.cc/hardware/nano-rp2040-connect
+// - Nano RP2040 Connect technical reference: https://docs.arduino.cc/tutorials/nano-rp2040-connect/rp2040-01-technical-reference
+//
+package machine
+
+// Digital Pins
+const (
+	D2  Pin = GPIO25
+	D3  Pin = GPIO15
+	D4  Pin = GPIO16
+	D5  Pin = GPIO17
+	D6  Pin = GPIO18
+	D7  Pin = GPIO19
+	D8  Pin = GPIO20
+	D9  Pin = GPIO21
+	D10 Pin = GPIO5
+	D11 Pin = GPIO7
+	D12 Pin = GPIO4
+	D13 Pin = GPIO6
+	D14 Pin = GPIO26
+	D15 Pin = GPIO27
+	D16 Pin = GPIO28
+	D17 Pin = GPIO29
+	D18 Pin = GPIO12
+	D19 Pin = GPIO13
+)
+
+// Analog pins
+const (
+	A0 Pin = ADC0
+	A1 Pin = ADC1
+	A2 Pin = ADC2
+	A3 Pin = ADC3
+)
+
+// Onboard LED
+const (
+	LED = GPIO6
+)
+
+// UART1 pins
+const (
+	UART_TX_PIN Pin = GPIO0
+	UART_RX_PIN Pin = GPIO1
+)
+
+// I2C pins
+const (
+	SDA_PIN Pin = GPIO12
+	SCL_PIN Pin = GPIO13
+)
+
+// SPI pins
+const (
+	SPI0_SCK_PIN Pin = GPIO6
+	SPI0_SDO_PIN Pin = GPIO7
+	SPI0_SDI_PIN Pin = GPIO4
+)
+
+// NINA-W102 Pins
+const (
+	NINA_SCK Pin = GPIO14
+	NINA_SDO Pin = GPIO11
+	NINA_SDI Pin = GPIO8
+
+	NINA_CS     Pin = GPIO9
+	NINA_ACK    Pin = GPIO10
+	NINA_GPIO0  Pin = GPIO0
+	NINA_RESETN Pin = GPIO3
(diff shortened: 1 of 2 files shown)


src/machine/board_rp2040.go
```diff
@@ -0,0 +1,41 @@
+// +build rp,rp2040 arduino_nano_rp2040_connect pico
```
kenbell: This applies to all rp2040?  If so should build tag just be:
`// +build rp2040`

ysoldak: I have no opinion, was following this example
https://github.com/tinygo-org/tinygo/blob/release/src/machine/board_atsamd21.go#L1
No prob to change, what shall it be then?

kenbell: I'd err on the side of simplicity, unless some other folks want to jump in with other thoughts.

ysoldak: OK, let's wait a day, if no-one objects, I go and simplify to just "rp2040"

kenbell:
On naming, most of the other boards don't include the manufacturer (for better or worse) - adafruit boards are just `feather-xx` for example.

How about either `nano-rp2040-connect` or `rp2040-connect` for this board?

ysoldak:
> On naming, most of the other boards don't include the manufacturer (for better or worse) - adafruit boards are just `feather-xx` for example.
> 
> How about either `nano-rp2040-connect` or `rp2040-connect` for this board?

I like `rp2040-connect`.

Out of scope of this PR entirely and probably better discuss on Slack, but just since we are on it...
"arduino-" targets seem to be outliers among other boards. Do we want to break this pattern or perhaps rename, say, `arduino-nano33` to more appropriate `nano33-iot` some day?
I plan to add **Arduino Nano33 BLE [Sense]** board also, targets could be `nano33-ble` and `nano33-ble-sense`.

deadprogram:
Hi @ysoldak please resolve the conflict that appeared after I merged the ADC PR #1932

src/machine/board_rp2040.go
deadprogram: I agree that a single tag is best when possible.

ysoldak:
> Hi @ysoldak please resolve the conflict that appeared after I merged the ADC PR #1932

No prob, I've decided to drop that common `board_rp2040.go` anyway, since `GPXX` are indeed silkscreen on Pico and not something common for all rp2040 boards.

deadprogram:
Based on your last few comments, I think I have made some incorrect assumptions about pin naming in the recently merged ADC implementation. Seems like it would not compile for rp2040-connect although it would probably work.

https://github.com/tinygo-org/tinygo/commit/9912dd6db1496eefd806d9cceb9527b6478b7485#diff-ebe247a53831fec6d096aba454c0d6e7e1ea527abe01be43ecc5284411b60b11R25

and 

https://github.com/tinygo-org/tinygo/commit/9912dd6db1496eefd806d9cceb9527b6478b7485#diff-ebe247a53831fec6d096aba454c0d6e7e1ea527abe01be43ecc5284411b60b11R50-R56

are the lines I am referring to.

ysoldak:
Pin numbers shall be same though... 26..29

Can I help with this somehow?

deadprogram:
I think we need a processor-based pin naming based on the datasheet:
image

Then each board can have in its board file whatever is used for the silkscreen. 

In other words, kinda back to what you originally had, but with a different naming scheme.

What do you think?

ysoldak:
I was just copying atsamd.
But good idea, yeah. Shall we handle it separately, different PR, I mean?

deadprogram:
Yes, separate PR just to keep it clean, please. Also that lets me use that for my upcoming PR adding UART.

ysoldak:
Well, then we just need to merge this PR then and make new with `board_rp2040.go` in it.

deadprogram:
I think `machine_rp2040.go` would be the place to put the "real" pins, not a board file.

ysoldak:
See #1944

deadprogram:
@ysoldak #1944 was merged, so can you please update this PR now to reflect those changes?

src/machine/board_nano-rp2040.go
```diff
@@ -0,0 +1,99 @@
...
+	D15 Pin = GPIO27
+	D16 Pin = GPIO28
+	D17 Pin = GPIO29
+	D18 Pin = GPIO12
+	D19 Pin = GPIO13
+)
+
+// Analog pins
+const (
+	A0 Pin = GPIO26
```
deadprogram: Do you perhaps want to use

```go
A0 = ADC0 
```

Since they are also already defined in `machine_rp2040.go`?

ysoldak: I was just looking at pin diagram for Arduino Nano RP2040 Connect board itself.
Does it make any difference?

Nano-RP2040_analog-pins

ysoldak: I mean will ADC code fail in this case? I see it checks in `switch` for `ADCX`, will it fail checking against `GPIO26` even if they have same int values?

ysoldak: nvm, changed that. I was just curios would it make any difference.

deadprogram:
Thanks for this @ysoldak now merging.

PR #1944, opened 2021-06-09
machine/rp2040: GPIO and ADC pins

ysoldak:
Something like this, @deadprogram ?


deadprogram:
Looks great to me @ysoldak thank you!

Now merging.

PR #1945, opened 2021-06-09
machine/rp2040: add UART support

deadprogram:
This PR is to add support for UART on the RP2040. TX works as expected, but RX is not working yet.

Note this is not USB-CDC support, so you will need to connect an FTDI adaptor to pins 0 and 1 to test it.

EDIT: RX is now working.

### DIFF
--- a/src/machine/machine_rp2040_uart.go
+++ b/src/machine/machine_rp2040_uart.go
@@ -1,0 +1,133 @@
+// +build rp2040
+
+package machine
+
+import (
+	"device/rp"
+	"runtime/interrupt"
+)
+
+// UART on the RP2040.
+type UART struct {
+	Buffer    *RingBuffer
+	Bus       *rp.UART0_Type
+	Interrupt interrupt.Interrupt
+}
+
+// Configure the UART.
+func (uart *UART) Configure(config UARTConfig) error {
+	initUART(uart)
+
+	// Default baud rate to 115200.
+	if config.BaudRate == 0 {
+		config.BaudRate = 115200
+	}
+
+	// Use default pins if pins are not set.
+	if config.TX == 0 && config.RX == 0 {
+		// use default pins
+		config.TX = UART_TX_PIN
+		config.RX = UART_RX_PIN
+	}
+
+	uart.SetBaudRate(config.BaudRate)
+
+	// default to 8-1-N
+	uart.SetFormat(8, 1, ParityNone)
+
+	// Enable the UART, both TX and RX
+	uart.Bus.UARTCR.SetBits(rp.UART0_UARTCR_UARTEN |
+		rp.UART0_UARTCR_RXE |
+		rp.UART0_UARTCR_TXE)
+
+	// set GPIO mux to UART for the pins
+	config.TX.Configure(PinConfig{Mode: PinUART})
+	config.RX.Configure(PinConfig{Mode: PinUART})
+
+	// Enable RX IRQ.
+	uart.Interrupt.SetPriority(0x80)
+	uart.Interrupt.Enable()
+
+	// setup interrupt on receive
+	uart.Bus.UARTIMSC.Set(rp.UART0_UARTIMSC_RXIM)
+
+	return nil
+}
+
+// SetBaudRate sets the baudrate to be used for the UART.
+func (uart *UART) SetBaudRate(br uint32) {
+	div := 8 * 125 * MHz / br
+
+	ibrd := div >> 7
+	var fbrd uint32
+
+	switch {
+	case ibrd == 0:
+		ibrd = 1
+		fbrd = 0
+	case ibrd >= 65535:
+		ibrd = 65535
+		fbrd = 0
+	default:
+		fbrd = ((div & 0x7f) + 1) / 2
+	}
+
+	// set PL011 baud divisor registers
+	uart.Bus.UARTIBRD.Set(ibrd)
+	uart.Bus.UARTFBRD.Set(fbrd)
+
+	// PL011 needs a (dummy) line control register write.
+	// See https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_uart/uart.c#L93-L95
+	uart.Bus.UARTLCR_H.SetBits(0)
+}
+
+// WriteByte writes a byte of data to the UART.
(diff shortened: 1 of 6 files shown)


ysoldak:
Verified on Nano-RP2040, works!
Had to copy over UART init block from `board_pico.go`. Would be nice to share it somehow?

deadprogram:
Now both TX and RX are working.

Also, can be used on other rp2040 boards without adding the UART code to every `board_*-rp2040.go` file. Please give it a try @ysoldak (and anyone else, of course).

ysoldak:
Verified, works on Arduino Nano RP2040 Connect, both TX and RX.
Been using `src/examples/echo/echo.go`

kenbell:
Working good for me on a pico and echo example

src/machine/machine_rp2040_uart.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+// UART on the RP2040.
+type UART struct {
+	Buffer    *RingBuffer
+	Bus       *rp.UART0_Type
+	Interrupt interrupt.Interrupt
+}
+
+func (uart *UART) Configure(config UARTConfig) error {
```
aykevl, line 18: Can you add some documentation to the public functions `Configure` and `SetBaudRate`?

src/machine/machine_rp2040_uart.go
```diff
@@ -0,0 +1,132 @@
...
+	// set GPIO mux to UART for the pins
+	config.TX.Configure(PinConfig{Mode: PinUART})
+	config.RX.Configure(PinConfig{Mode: PinUART})
+
+	// Enable RX IRQ.
+	uart.Interrupt.SetPriority(0x80)
+	uart.Interrupt.Enable()
+
+	// setup interrupt on receive
+	uart.Bus.UARTIMSC.Set(rp.UART0_UARTIMSC_RXIM)
```
aykevl, line 52: Looks like you're only supporting interrupts on UART0?

src/machine/machine_rp2040_uart.go
```diff
@@ -0,0 +1,132 @@
...
+	for uart.Bus.UARTFR.HasBits(rp.UART0_UARTFR_TXFF) {
+	}
+
+	// write data
+	uart.Bus.UARTDR.Set(uint32(c))
+	return nil
+}
+
+// SetFormat for number of data bits, stop bits, and parity for the UART.
+func (uart *UART) SetFormat(databits, stopbits int, parity UARTParity) error {
```
aykevl: This is a new API. It looks clean and reasonable, although perhaps it makes sense to make `databits` and `stopbits` 8-bit numbers.

src/machine/uart.go
```diff
@@ -1,11 +1,19 @@
-// +build atmega esp nrf sam sifive stm32 k210 nxp
+// +build atmega esp nrf sam sifive stm32 k210 nxp rp2040
 
 package machine
 
 import "errors"
 
 var errUARTBufferEmpty = errors.New("UART buffer empty")
 
+type UARTParity int
```
aykevl, line 10: I think this also needs some documentation.

src/machine/uart.go
```diff
@@ -1,11 +1,19 @@
...
+
+const (
+	ParityNone UARTParity = 0
```
aykevl, line 15: Perhaps you can add some documentation indicating that `ParityNone` is by far the most commonly used?

aykevl (commented):
Looks good, with a few nits.

src/machine/machine_rp2040_uart.go
deadprogram: The `UART0_UARTIMSC_RXIM` bitfield is generated and applies to either UART. 

`Bus` in this case being either `UART0` or `UART1`, and each UARTs has its own `UARTIMSC` register.

src/machine/machine_rp2040_uart.go
deadprogram: Done.

src/machine/machine_rp2040_uart.go
deadprogram: Done.

src/machine/uart.go
deadprogram: Done.

src/machine/uart.go
deadprogram: Also done.

deadprogram:
All feedback has been addressed and all tests passing. Should be ready to merge.

deadprogram:
OK, now merging. Thanks everyone for feedback and testing!

mattn:
`src/examples/echo/echo.go` works on Raspberry Pi Pico? On my Windows 10 does not recognize serial port (like COM7) with tinygo. I could read/write with pico-sdk and stdio_init_all.

```c
#include <stdio.h>
#include "pico/stdlib.h"

int main() {
    stdio_init_all();
    int i = 1;
    while (true) {
        if (i % 15 == 0)
          printf("FizzBuzz\n");
        else if (i % 3 == 0)
          printf("Fizz\n");
        else if (i % 5 == 0)
          printf("Buzz\n");
        else
          printf("%d\n", i);
        sleep_ms(1000);
        i++;
    }
    return 0;
}
```

sago35:
Hi @mattn 

I tried it with tinygo, the latest dev branch.
It seems to be working as follows.

$ tinygo version 0.20.0-dev-f57e9622 windows/amd64 (using go version go1.16.5 and LLVM version 11.0.0)

image

It is connected and checked as follows.
Don't forget to connect `GND`.
Don't forget to adjust the serial speed.
The pico also defaults to 115200bps.

https://github.com/tinygo-org/tinygo/blob/04f520040e6d60ace0728d0016cbe1e75bc44f71/src/machine/machine_rp2040_uart.go#L22-L24

* PC  <--> XIAO <--> pico
* PC vs XIAO
    * USBCDC
* XIAO vs pico
    * XIAO Rx <--> pico Tx
    * XIAO Tx <--> pico Rx

image

deadprogram:
Yes, the UART is not the same as USB-CDC support, which is not yet complete for rp2040. That means you must use the UART pins as @sago35 has done.

mattn:
@deadprogram @sago35 Thanks.

PR #1946, opened 2021-06-10
board/nano-33-ble: pins, blinking leds and serial

ysoldak:
Support Arduino Nano 33 BLE [Sense] boards.

Flashed and verified with `blinky1`, `blinky2` and `echo` examples.

Requires special version of bossac, that can be easily obtained either via Arduino IDE or by simple download.
See comments in the header of the board definition. (thanks @deadprogram for finding the second option!)

~Examples changed a bit, hope that's OK.~

### DIFF
--- a/src/machine/board_nano-33-ble.go
+++ b/src/machine/board_nano-33-ble.go
@@ -1,0 +1,100 @@
+// +build nano_33_ble
+
+// This contains the pin mappings for the Arduino Nano 33 BLE [Sense] boards.
+//
+// Flashing the board requires special version of bossac.
+//
+// This executable can be obtained two ways:
+// 1) In Arduino IDE, install support for the board ("Arduino Mbed OS Nano Boards")
+//    Search for "tools/bossac/1.9.1-arduino2/bossac" in Arduino IDEs directory
+// 2) Download https://downloads.arduino.cc/packages/package_index.json
+//    Search for "bossac-1.9.1-arduino2" in that file
+//    Download tarball for your OS and unpack it
+//
+// Once you have the executable, make it accessible in your PATH as "bossac_arduino2".
+//
+// It is possible to replace original bossac with this new one (this only adds support for nrf chip).
+// In that case make "bossac_arduino2" symlink on it, for the board target to be able to find it.
+//
+// For more information, see:
+// - https://store.arduino.cc/arduino-nano-33-ble
+// - https://store.arduino.cc/arduino-nano-33-ble-sense
+//
+package machine
+
+const HasLowFrequencyCrystal = true
+
+// Digital Pins
+const (
+	D2  Pin = P1_11
+	D3  Pin = P1_12
+	D4  Pin = P1_15
+	D5  Pin = P1_13
+	D6  Pin = P1_14
+	D7  Pin = P0_23
+	D8  Pin = P0_21
+	D9  Pin = P0_27
+	D10 Pin = P1_02
+	D11 Pin = P1_01
+	D12 Pin = P1_08
+	D13 Pin = P0_13
+)
+
+// Analog pins
+const (
+	A0 Pin = P0_04
+	A1 Pin = P0_05
+	A2 Pin = P0_30
+	A3 Pin = P0_29
+	A4 Pin = P0_31
+	A5 Pin = P0_02
+	A6 Pin = P0_28
+	A7 Pin = P0_03
+)
+
+// Onboard LEDs
+const (
+	LED         = LED_BUILTIN
+	LED1        = LED_RED
+	LED2        = LED_GREEN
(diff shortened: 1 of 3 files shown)


src/examples/echo/echo.go
```diff
@@ -25,7 +25,7 @@ func main() {
 			data, _ := uart.ReadByte()
 
 			switch data {
-			case 13:
+			case 10, 13:
```
ysoldak: Makes it work on Mac

src/examples/blinky2/blinky2.go
```diff
@@ -11,34 +11,18 @@ import (
...
 func main() {
-	go led1()
-	led2()
+	go led(machine.LED1, time.Millisecond*1000, "")
+	led(machine.LED2, time.Millisecond*420, "  ")
 }
 
-func led1() {
-	led := machine.LED1
+func led(led machine.Pin, interval time.Duration, padding string) {
```
ysoldak: Just merged functions in one. Makes it easier ad-hoc test more than 2 LEDs.

src/machine/board_nano-33-ble.go
```diff
@@ -0,0 +1,80 @@
+// +build nano_33_ble
+
+package machine
+
+const HasLowFrequencyCrystal = false
```
ysoldak: I don't know if this board actually has it.
It works with this value, probably will work even better if enabled, I just don't know how to test.

src/machine/board_nano-33-ble.go
```diff
@@ -0,0 +1,80 @@
...
+	A4 Pin = P0_31
+	A5 Pin = P0_02
+	A6 Pin = P0_28
+	A7 Pin = P0_03
+)
+
+// Onboard LEDs
+const (
+	LED         = LED_BUILTIN
+	LED1        = LED_RED
```
ysoldak, line 58: Numbered LEDs just to make it easier with the examples.

targets/nano-33-ble.ld
```diff
@@ -0,0 +1,11 @@
+
+MEMORY
+{
+    FLASH_TEXT (rw) : ORIGIN = 0x10000, LENGTH = 1M
```
ysoldak: ORIGIN is important, hence separate .ld file

targets/nano-33-ble.json
```diff
@@ -0,0 +1,7 @@
+{
+	"inherits": ["nrf52840"],
+	"build-tags": ["nano_33_ble"],
+	"flash-command": "bossac_arduino2 -d -i -e -w -v -R --port={port} {bin}",
```
ysoldak, line 4: special version of bossac and no `--offset`

src/machine/board_nano-33-ble.go
aykevl: If it works with `HasLowFrequencyCrystal = true`, there is probably a low frequency oscillator. Most boards (except for the smallest/cheapest) have one.

targets/nano-33-ble.ld
```diff
@@ -0,0 +1,11 @@
...
+    RAM (rwx)       : ORIGIN = 0x20000000, LENGTH = 256K
+}
+OUTPUT_FORMAT ("elf32-littlearm", "elf32-bigarm", "elf32-littlearm")
```
aykevl: I don't think this line is needed.

targets/nano-33-ble.ld
aykevl: I see. However, this is not quite correct: the chip has only 1M of flash and `LENGTH = 1M` will go over this. It should be this:

```suggestion
    FLASH_TEXT (rw) : ORIGIN = 0x10000, LENGTH = 0xf0000
```

Unfortunately, this also means it's probably not possible to use the SoftDevice on this chip without destroying the bootloader.

aykevl (commented):
Can you please separate the changes to the examples into a new PR? That makes it easier to discuss these changes.

ysoldak:
- Added docs to the board file.
- Removed examples from the PR.
- Adjusted `.ld` file according to suggestions, verified -- still works.
- Low frequency crystal boolean set to "true" and it still works, leaving it in this state.
- Squashed and rebased on current dev.

aykevl:
Thank you!
It's a bit unfortunate that this needs a special version of bossac but not much we can do about that :shrug:

deadprogram:
> Unfortunately, this also means it's probably not possible to use the SoftDevice on this chip without destroying the bootloader.

I expected that. 

There is UF2 bootloader available for this board https://github.com/adafruit/Adafruit_nRF52_Bootloader/tree/master/src/boards/arduino_nano_33_ble

However flashing the bootloader will require connecting to the debug pins on bottom of the board. Not too easy for most users.

PR #1948, opened 2021-06-13
compiler: refactor method names

aykevl:
This commit includes two changes:

  * It makes unexported interface methods package-private, so that it's
    not possible to type-assert on an unexported method in a different
    package.
  * It makes the globals used to identify interface methods defined
    globals, so that they can (eventually) be left in the program for an
    eventual non-LTO build mode.

---

I have extracted this commit from my local branch to refactor the reflect package.


deadprogram:
The change is clearly defined, and proven by tests, so looks good to me. Now merging, thanks @aykevl

PR #1950, opened 2021-06-13
BlackMagic/BMP Initial support for flashing and debugging

ofauchon:
This PR brings  basic support for flashing and debugging with BlackMagic Interface. 
(https://github.com/blacksphere/blackmagic)

You just have to use **-programmer=bmp** to switch to BlackMagic debugger
(Default BMP Debug port is /dev/ttyACM0, but you can change it with **-port /dev/whatever**

How to use it : 
```
# For flashing
tinygo flash -x  -target=bluepill -programmer=bmp blinky1.go

# For debugging 
tinygo gdb -x  -target=bluepill -programmer=bmp blinky1.go
```
At the moment, it was only tested on BluePill targets (stm32f103).

PS: If you don't already know BlackMagic Probe, you should read hackaday.com review: 

https://hackaday.com/2016/12/02/black-magic-probe-the-best-arm-jtag-debugger/

_"...
First off, it’s got a JTAG and a UART serial port in one device. You can flash the target, run your code, use the serial port for printf debugging like you know you want to, and then fall back on full-fledged JTAG-plus-GDB when you need to, all in one dongle. It’s just very convenient.
...
BMP’s killer feature is that it runs a GDB server on the probe. It opens up a virtual serial port that you can connect to directly through GDB on your host computer. No need to hassle around with OpenOCD configurations, or to open up a second window to run [texane]’s marvelous st-util. Just run GDB, "target extended-remote /dev/ttyACM0" and you’re debugging.
"_

### DIFF
--- a/main.go
+++ b/main.go
@@ -288,6 +288,8 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 		fileExt = filepath.Ext(config.Target.FlashFilename)
 	case "openocd":
 		fileExt = ".hex"
+	case "bmp":
+		fileExt = ".elf"
 	case "native":
 		return errors.New("unknown flash method \"native\" - did you miss a -target flag?")
 	default:
@@ -380,6 +382,25 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 				return &commandError{"failed to flash", result.Binary, err}
 			}
 			return nil
+		case "bmp":
+			gdb, err := config.Target.LookupGDB()
+			if err != nil {
+				return err
+			}
+			var bmpGDBPort string
+			bmpGDBPort, _, err = getBMPPorts()
+			if err != nil {
+				return err
+			}
+			args := []string{"-ex", "target extended-remote " + bmpGDBPort, "-ex", "monitor swdp_scan", "-ex", "attach 1", "-ex", "load", filepath.ToSlash(result.Binary)}
+			cmd := executeCommand(config.Options, gdb, args...)
+			cmd.Stdout = os.Stdout
+			cmd.Stderr = os.Stderr
+			err = cmd.Run()
+			if err != nil {
+				return &commandError{"failed to flash", result.Binary, err}
+			}
+			return nil
 		default:
 			return fmt.Errorf("unknown flash method: %s", flashMethod)
 		}
@@ -434,6 +455,13 @@ func FlashGDB(pkgName string, ocdOutput bool, options *compileopts.Options) erro
 		switch gdbInterface {
 		case "native":
 			// Run GDB directly.
+		case "bmp":
+			var bmpGDBPort string
+			bmpGDBPort, _, err = getBMPPorts()
+			if err != nil {
+				return err
+			}
+			gdbCommands = append(gdbCommands, "target extended-remote "+bmpGDBPort, "monitor swdp_scan", "compare-sections", "attach 1", "load")
 		case "openocd":
 			gdbCommands = append(gdbCommands, "target extended-remote :3333", "monitor halt", "load", "monitor reset halt")
 
@@ -839,6 +867,35 @@ func getDefaultPort(portFlag string, usbInterfaces []string) (port string, err e
 	}
 
 	return "", errors.New("port you specified '" + strings.Join(portCandidates, ",") + "' does not exist, available ports are " + strings.Join(ports, ", "))
+}
+
+// getBMPPorts returns BlackMagicProbe's serial ports if any
+func getBMPPorts() (gdbPort, uartPort string, err error) {
+	var portsList []*enumerator.PortDetails
+	portsList, err = enumerator.GetDetailedPortsList()
+	if err != nil {
+		return "", "", err
+	}
+	var ports []string
+	for _, p := range portsList {
+		if !p.IsUSB {
+			continue
+		}
+		if p.VID != "" && p.PID != "" {
+			vid, vidErr := strconv.ParseUint(p.VID, 16, 16)
+			pid, pidErr := strconv.ParseUint(p.PID, 16, 16)
+			if vidErr == nil && pidErr == nil && vid == 0x1d50 && pid == 0x6018 {
+				ports = append(ports, p.Name)
+			}
+		}
+	}
+	if len(ports) == 2 {
+		return ports[0], ports[1], nil
+	} else if len(ports) == 0 {
+		return "", "", errors.New("no BMP detected")
+	} else {
+		return "", "", fmt.Errorf("expected 2 BMP serial ports, found %d - did you perhaps connect more than one BMP?", len(ports))
+	}
 }
 
 func usage() {
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -285,6 +285,9 @@ func (c *Config) Programmer() (method, openocdInterface string) {
 	case "openocd", "msd", "command":
 		// The -programmer flag only specifies the flash method.
 		return c.Options.Programmer, c.Target.OpenOCDInterface
+	case "bmp":
+		// The -programmer flag only specifies the flash method.
+		return c.Options.Programmer, ""
 	default:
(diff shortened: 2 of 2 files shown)


main.go
```diff
@@ -377,6 +379,20 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 				return &commandError{"failed to flash", result.Binary, err}
 			}
 			return nil
+		case "bmp":
+			bmpDebugPort := "/dev/ttyACM0"
```
aykevl: This is Linux specific. Can you make it more cross platform? You can take a look at `getDefaultPort` for example (possibly just calling that function).

main.go
```diff
@@ -431,6 +447,12 @@ func FlashGDB(pkgName string, ocdOutput bool, options *compileopts.Options) erro
 		switch gdbInterface {
 		case "native":
 			// Run GDB directly.
+		case "bmp":
+			bmpDebugPort := "/dev/ttyACM0"
```
aykevl: Same here.

main.go
```diff
@@ -377,6 +379,20 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 				return &commandError{"failed to flash", result.Binary, err}
 			}
 			return nil
+		case "bmp":
+			bmpDebugPort := "/dev/ttyACM0"
+			if len(port) > 0 {
+				bmpDebugPort = port
+			}
+			args := []string{"-ex", "target extended-remote " + bmpDebugPort, "-ex", "monitor swdp_scan", "-ex", "attach 1", "-ex", "load", filepath.ToSlash(result.Binary)}
+			cmd := executeCommand(config.Options, "arm-none-eabi-gdb", args...)
```
aykevl: This is also Linux specific. Can you please use `LookupGDB` instead?

aykevl (commented):
The Black Magic Probe is certainly an interesting debugger. This PR is more or less good, but has some platform specific things that need to be fixed to make it work everywhere (not just Linux).

main.go
ofauchon: Hi Aykevl ! 
I'll try to improve the PR to make it cross platform. 
Thanks for your feedback.

main.go
ofauchon: OK thanks

main.go
ofauchon: I could use LookupGDB without problem, 'arm-none-eabi-gdb' was properly detected. 
(Only tested on Linux, with a couple of STM32 boards)

main.go
ofauchon: I could change with getDefaultPort quite easilly. 

Now I get correct messages like "error: multiple serial ports available - use -port flag, available ports are /dev/ttyACM0, /dev/ttyACM1, /dev/ttyACM2" when not passing -port parameter. 

And I also made some tests with passing multiple ports eg: -port=COM1,/dev/ttyACM2, 
It worked properly too (/dev/ttyACM2 was used)

Unfortunatly, I could not test it under windows and MacOS yet.

main.go
ofauchon: Fixed the same way as above comment.

ofauchon:
Please note I added a extra "port string" parameter in FlashGDB(), like in Flash() function 
This is because BMP needs "gdb target remote <serial_port>" for both flashing and debugging.

Hope this doesn't hurt...

aykevl:
I have made a PR to filter the ports by VID/PID: https://github.com/tinygo-org/tinygo/pull/1956
I think you can use that to get the correct serial port of the BMP, without needing to use the `-port` command line parameter.

ofauchon:
Hi @aykevl . 

serial "vid/pid" filtering is a nice feature!  
However, it seems to fail detecting BlackMagic probe port . 

The problem is that BMP interface exposes two acm ports (One for the debugger, and another for extra USART) eg:/dev/ttyACM1 and /dev/ttyACM2 on my setup

```
$ lsusb | grep Black
Bus 001 Device 069: ID 1d50:6018 OpenMoko, Inc. Black Magic Debug Probe (Application)

$ usb-devices
T:  Bus=01 Lev=01 Prnt=01 Port=00 Cnt=01 Dev#= 69 Spd=12  MxCh= 0
D:  Ver= 2.00 Cls=ef(misc ) Sub=02 Prot=01 MxPS=32 #Cfgs=  1
P:  Vendor=1d50 ProdID=6018 Rev=01.00
S:  Manufacturer=Black Sphere Technologies
S:  Product=Black Magic Probe (SWLINK) v1.7.1-194-g8089e05
S:  SerialNumber=5396DBBB
C:  #Ifs= 6 Cfg#= 1 Atr=80 MxPwr=100mA
I:  If#=0x0 Alt= 0 #EPs= 1 Cls=02(commc) Sub=02 Prot=00 Driver=cdc_acm  <<<<
I:  If#=0x1 Alt= 0 #EPs= 2 Cls=0a(data ) Sub=00 Prot=00 Driver=cdc_acm  <<<<
I:  If#=0x2 Alt= 0 #EPs= 1 Cls=02(commc) Sub=02 Prot=00 Driver=cdc_acm <<<<
I:  If#=0x3 Alt= 0 #EPs= 2 Cls=0a(data ) Sub=00 Prot=00 Driver=cdc_acm <<<<
I:  If#=0x4 Alt= 0 #EPs= 0 Cls=fe(app. ) Sub=01 Prot=01 Driver=(none)
I:  If#=0x5 Alt= 0 #EPs= 1 Cls=ff(vend.) Sub=ff Prot=ff Driver=(none)
``` 

When I try to add something like this in the Flash() and FlashGDB() functions:   
```
case "bmp":
...
port, err = getDefaultPort(port, []string{"acm:1d50:6018"})"  
```
 I get : 

```
$ tinygo flash -x -target=bluepill -programmer=bmp   src/examples/blinky1/blinky1.go 
....
error: multiple serial ports available - use -port flag, available ports are /dev/ttyACM1, /dev/ttyACM2

```

I see two solutions : 

- Add an extra optional 'index' parameter in vid/pid declaration : ex: "acm:1d50:6018:1"
- Accept my proposal of adding "port" parameter in FlashGDB() function : Same as in Flash() function, 
for specifying the correct port.  

Thanks

Olivier

aykevl:
I think the best solution is to extract some logic from `getDefaultPort`, something like this:

```go
func getBMPPorts() (serialPort, monitorPort string, err error) {
	var portsList []*enumerator.PortDetails
	portsList, err = enumerator.GetDetailedPortsList()
	if err != nil {
		return "", "", err
	}

	var ports []string
	for _, p := range portsList {
		if !p.IsUSB {
			continue
		}
		if p.VID != "" && p.PID != "" {
			vid, vidErr := strconv.ParseUint(p.VID, 16, 16)
			pid, pidErr := strconv.ParseUint(p.PID, 16, 16)
			if vidErr == nil && pidErr == nil && vid == 0x1d50 && pid == 0x6018 {
				ports = append(ports, p.Name)
			}
		}
	}
	if len(ports) == 2 {
		return ports[0], ports[1], nil
	} else if len(ports) == 0 {
		return "", "", errors.New("no BMP detected")
	} else {
		return "", "", fmt.Errorf("expected 2 BMP serial ports, found %d - did you perhaps connect more than one BMP?", len(ports))
	}
}
```

ofauchon:
Hi @aykevl. 

The dedicated "getBMPPorts()" function seems the best way to detect properly BMP serial interfaces. 

I made the changes, and it's working well on Linux (Still no Windows for testing)

Could you please review the PR  ?

Thanks 

Olivier

ofauchon:
Hi @aykevl,
Do you think this PR could be merged, or needs improvements ? 

Thanks

sago35 (approved):
Looks good to me.

deadprogram:
Now merging, thanks for all this great work @ofauchon and also to @sago35 and @aykevl for reviewing.

aykevl:
Sorry for the delay! This PR looks good to me, thanks @deadprogram for merging.

PR #1951, opened 2021-06-13
smoke&readme: add missing boards

ysoldak:
Smoke and ~Mirrors~ Readme.

Boards added:
- Arduino Nano 33 BLE
- Arduino Nano 33 BLE Sense
- Arduino Nano RP2040 Connect
- Raspberry Pi Pico

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -356,7 +356,11 @@ smoketest:
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=arduino-nano33      examples/blinky1
 	@$(MD5SUM) test.hex
-	$(TINYGO) build -size short -o test.hex -target=pico       			examples/blinky1
+	$(TINYGO) build -size short -o test.hex -target=pico                examples/blinky1
+	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=nano-33-ble         examples/blinky1
+	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=nano-rp2040         examples/blinky1
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=feather-rp2040 		examples/blinky1
 	@$(MD5SUM) test.hex
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ See the [getting started instructions](https://tinygo.org/getting-started/) for 
 
 You can compile TinyGo programs for microcontrollers, WebAssembly and Linux.
 
-The following 63 microcontroller boards are currently supported:
+The following 67 microcontroller boards are currently supported:
 
 * [Adafruit Circuit Playground Bluefruit](https://www.adafruit.com/product/4333)
 * [Adafruit Circuit Playground Express](https://www.adafruit.com/product/3333)
@@ -69,7 +69,10 @@ The following 63 microcontroller boards are currently supported:
 * [Arduino Mega 2560](https://store.arduino.cc/arduino-mega-2560-rev3)
 * [Arduino MKR1000](https://store.arduino.cc/arduino-mkr1000-wifi)
 * [Arduino Nano](https://store.arduino.cc/arduino-nano)
-* [Arduino Nano33 IoT](https://store.arduino.cc/nano-33-iot)
+* [Arduino Nano 33 BLE](https://store.arduino.cc/nano-33-ble)
+* [Arduino Nano 33 BLE Sense](https://store.arduino.cc/nano-33-ble-sense)
+* [Arduino Nano 33 IoT](https://store.arduino.cc/nano-33-iot)
+* [Arduino Nano RP2040 Connect](https://store.arduino.cc/nano-rp2040-connect)
 * [Arduino Uno](https://store.arduino.cc/arduino-uno-rev3)
 * [Arduino Zero](https://store.arduino.cc/usa/arduino-zero)
 * [BBC micro:bit](https://microbit.org/)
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -356,7 +356,11 @@ smoketest:
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=arduino-nano33      examples/blinky1
 	@$(MD5SUM) test.hex
-	$(TINYGO) build -size short -o test.hex -target=pico       			examples/blinky1
+	$(TINYGO) build -size short -o test.hex -target=pico                examples/blinky1
```
ysoldak, line 359: whitespace fix: this line had tabs in place of spaces

README.md
```diff
@@ -68,7 +68,10 @@ The following 62 microcontroller boards are currently supported:
 * [Arduino Mega 2560](https://store.arduino.cc/arduino-mega-2560-rev3)
 * [Arduino MKR1000](https://store.arduino.cc/arduino-mkr1000-wifi)
 * [Arduino Nano](https://store.arduino.cc/arduino-nano)
-* [Arduino Nano33 IoT](https://store.arduino.cc/nano-33-iot)
+* [Arduino Nano 33 BLE](https://store.arduino.cc/nano-33-ble)
+* [Arduino Nano 33 BLE Sense](https://store.arduino.cc/nano-33-ble-sense)
```
sago35, line 73: `targets/nano-33-ble-sense.json` does not exist.
Is this configuration correct?
If there is even a slight difference in pin placement, I think I need to add `board_nano-33-ble-sense.go` or something similar.

ysoldak: Yes, there is bit of mess there.
Silkscreen pins are exactly the same.
It's just Sense variant has more modules on board that can be accessed via I2C or SPI.

Compare:
https://content.arduino.cc/assets/Pinout-NANOble_latest.pdf
and
https://content.arduino.cc/assets/Pinout-NANOsense_latest.pdf

sago35: It looks like nano-33-ble-sense is just nano-33-ble with some additional part implementations.
If this is the case, `targets/nano-33-ble-sense` may not be necessary.

The difference between the board name in README.md and the name specified by -target is getting more and more difficult.

README.md
```diff
@@ -68,7 +68,10 @@ The following 62 microcontroller boards are currently supported:
...
+* [Arduino Nano 33 IoT](https://store.arduino.cc/nano-33-iot)
```
sago35, line 74: Do I use `-target arduino-nano33` to flash this board?

README.md
sago35: I think it would be easier for tinygo users to understand if there was a targets/nano-33-ble-sense.json.
See targets/clue.json and targets/clue-alpha.json.

https://github.com/tinygo-org/tinygo/blob/release/targets/clue.json
https://github.com/tinygo-org/tinygo/blob/release/targets/clue-alpha.json

Do you have any thoughts on this? @deadprogram

README.md
ysoldak: Yes, this is inherited inconsistency. We may want to have aliases.

README.md
ysoldak: Can we perhaps have `targets/nano-33-ble-sense.json` with only content as following?
```
{"inherits":["nano-33-ble"]}
```
This basically implements an alias.

deadprogram: > Can we perhaps have targets/nano-33-ble-sense.json with only content as following?

That seems to me to be the most consistent approach.

ysoldak:
OK, I'm planning to add two aliases with the approach described in comments above.
- `nano-33-iot` -> `arduino-nano33`
- `nano-33-ble-sense` -> `nano-33-ble`

Shall I add smoke tests for them to `Makefile`?

ysoldak:
Added target alias and smoke for it: `nano-33-ble-sense` -> `nano-33-ble`

Tell me if you feel like we can benefit from having `nano-33-iot` alias to `arduino-nano33` as well.
Otherwise, feel free to merge.

Makefile
```diff
@@ -356,7 +356,13 @@ smoketest:
...
+	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=nano-33-ble         examples/blinky1
+	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=nano-33-ble-sense   examples/blinky1
+	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=nano-rp2040         examples/blinky1
```
aykevl: Honestly I don't think we need smoke tests for aliases. They only make the CI longer and the only thing they test is that the alias works (which it should, as long as the JSON file isn't modified).

aykevl (commented):
Overall looks okay to me. I'm not too big of a fan of aliases, the one we have (clue/clue-alpha) exists for backwards compatibility reasons. In this case, we might eventually make the BLE Sense board have some more definitions, so it's fine by me. I do not think aliases will help users, they only add to the list of targets possibly confusing people.

Makefile
ysoldak: Sure, depends what we want to test with smokes (I can imagine json become corrupted after, say, some refactoring campaign).
I remove this then, unless @sago35 or @deadprogram have anything to say?

deadprogram: :scissors:

ysoldak:
Removed smoke and rebased. @aykevl @deadprogram

ysoldak:
Wanna merge before release?

ysoldak:
Nano-RP2040 smoke was failing after rebase.
Fix is trivial, a file rename, hope it's fine to have it in the scope of this PR.

deadprogram:
Thanks for the updates @ysoldak now merging.

PR #1956, opened 2021-06-17
main: detect specific serial port IDs based on USB vid/pid

aykevl:
This makes it possible to flash a board even when there are multiple
different kinds of boards attached, e.g. an Arduino Uno and a Circuit
Playground Express. You can find the VID/PID pair in several ways:

 1. By running `lsusb` before and after attaching the board and looking
    at the new USB device.
 2. By grepping for `usb_PID` and `usb_VID` in the TinyGo source code.
 3. By checking the Arduino IDE boards.txt from the vendor.

Note that one board may have multiple VID/PID pairs:

  * The bootloader and main program may have a different PID, so far
    I've seen that the main program generally has the bootloader PID
    with 0x8000 added.
  * The software running on the board may have an erroneous PID, for
    example from a different board. I've seen this happen a few times.
  * A single board may have had some revisions which changed the PID.
    This is particularly true for the Arduino Uno.

As a fallback, if the given VID/PID pair isn't found, the whole set of
serial ports will be used.

There are many boards which I haven't included yet simply because I
couldn't test them.

### DIFF
--- a/targets/pyportal.json
+++ b/targets/pyportal.json
@@ -3,6 +3,7 @@
     "build-tags": ["pyportal"],
     "flash-1200-bps-reset": "true",
     "flash-method": "msd",
+    "serial-port": ["acm:239a:8035", "acm:239a:35", "acm:239a:8036"],
     "msd-volume-name": "PORTALBOOT",
     "msd-firmware-name": "firmware.uf2",
     "default-stack-size": 2048
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -45,6 +45,7 @@ type TargetSpec struct {
 	FlashCommand     string   `json:"flash-command"`
 	GDB              []string `json:"gdb"`
 	PortReset        string   `json:"flash-1200-bps-reset"`
+	SerialPort       []string `json:"serial-port"` // serial port IDs in the form "acm:vid:pid" or "usb:vid:pid"
 	FlashMethod      string   `json:"flash-method"`
 	FlashVolume      string   `json:"msd-volume-name"`
 	FlashFilename    string   `json:"msd-firmware-name"`
--- a/main.go
+++ b/main.go
@@ -15,6 +15,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync/atomic"
 	"time"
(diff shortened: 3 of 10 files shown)


targets/pyportal.json
```diff
@@ -3,6 +3,7 @@
     "build-tags": ["pyportal"],
     "flash-1200-bps-reset": "true",
     "flash-method": "msd",
+    "serial-port": ["acm:239a:8035", "acm:239a:35", "acm:239a:8036"],
```
aykevl, line 6: Note: `239a:8036` is used by the default firmware on the board. I believe this is a bug in the firmware.

ysoldak:
Verified, works with recently added Arduino Nano 33 BLE.

Hint: on Mac, VID and PID values are easily obtained from System Information utility, see screenshot.
Menu flow: ` > About This Mac > System Report...`

nano-33-ble

aykevl:
Thank you for checking!
I think this misses a case: the PID is 0x805a in TinyGo (the bootloader and main program often have a different PID). You can test this by having a different serial port plugged in at the same time (e.g. some other board with a UF2 bootloader, or an Arduino Uno). If it complains about multiple serial ports while flashing, then the configuration is not complete.

ysoldak:
I had Nano 33 IoT plugged in to different port, in app mode, doing some debugging via serial console. Is that enough?

aykevl:
(after some private discussion)

I have added the VID/PID for the Arduino Nano 33 BLE too.

ysoldak:
Be informed, in the moment of testing, Arduino Nano 33 BLE [Sense] board requires manual switch to bootloader mode (double tap). Flashing in "application" mode does not seem to work. But that's not related to this PR, just general observation, hopefully helps someone.

aykevl:
https://github.com/tinygo-org/tinygo/pull/1959 should hopefully fix the double tap issue. It is otherwise unrelated to this PR.

deadprogram:
@sago35 any opinions on this before I merge it?

sago35 (approved):

sago35:
Looks good to me.

deadprogram:
Thanks @aykevl for this improvement, and thanks @ysoldak and @sago35 for reviewing.

Now merging.

Issue #1957, opened 2021-06-17
size of int on darwin inconsistent


oflebbe:
Hi, I was wondering about a test failure on my mac (testdata/filesystem.go) failed with "panic: should be non exist error"

Digging deeper I found that comparison of the fd filedescriptor in Open https://github.com/tinygo-org/tinygo/blob/dev/src/syscall/syscall_libc.go#L47 does not fail because the "int" datatype in go (and tinygo) is a 64 bit datatype , rather in the "C" binding is a 32bit data type. 

For example in go on GOARCH=amd64 GOOS=darwin
```package main

import "unsafe"
import "C"

func main() {
	var a int = 1
	println(unsafe.Sizeof(a))
	var b C.int = 2
	println(unsafe.Sizeof(b))

}
```

outputs 
`8 4`

Looks like the tinygo promise to have an easy interop between C and tinygo might not work on this platform:
https://tinygo.org/docs/reference/lang-support/

I am not too sure how to proceed from here:
* Make tinygo int 32bit on darwin ?
* Supply a different implementation for darwin  libc bindings, explicit done as int32?
* introduce a C.int datatype like in go for interop ?

Did I miss something along the way ?

oflebbe:
seems like @aykevl  recommended to use int32 in #1669 . I didn't recognized this accepting the code now in tinygo.

aykevl:
Yes, `int32` would be the correct choice (because practically all modern OSes nowadays uses 32-bit int in C). Despite the name, C int and Go int are two different types and you shouldn't assume they are the same size.

aykevl:
Here is a patch: https://github.com/tinygo-org/tinygo/pull/1958
Can you test whether it fixes the bug on your system?

oflebbe:
Alright! the tinygo tests are now running flawless on macos 11.4 with xcode 12.5 .
Thanks @aykevl !

deadprogram:
This has been released with v0.19.0 so now closing. Thanks everyone!

PR #1958, opened 2021-06-17
syscall: fix int type in libc version

aykevl:
Int in Go and C are two different types (hence why CGo has C.int). The code in syscall assumed they were of the same type, which led to a bug: https://github.com/tinygo-org/tinygo/issues/1957

While the C standard makes no guarantees on the size of int, in most modern operating systems it is 32-bits so Go int32 would be the correct choice.

### DIFF
--- a/src/syscall/syscall_libc.go
+++ b/src/syscall/syscall_libc.go
@@ -13,7 +13,7 @@ type sliceHeader struct {
 }
 
 func Close(fd int) (err error) {
-	if libc_close(fd) < 0 {
+	if libc_close(int32(fd)) < 0 {
 		err = getErrno()
 	}
 	return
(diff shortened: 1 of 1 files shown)


deadprogram:

yishangru:
Should this fix be treated as a deviation from the expected Go behavior? For Go, int is intended to be int64 for the 64-bit system, while Tinygo will use 32 bit. This will bring unexpected int overflow.

aykevl:
> For Go, int is intended to be int64 for the 64-bit system, while Tinygo will use 32 bit. This will bring unexpected int overflow.

No. TinyGo uses 32-bit int on 32-bit systems and 64-bit int on 64-bit systems. So it is compatible with Go behavior. The only exception is WebAssembly, where TinyGo uses 32-bit ints and upstream Go uses 64-bit ints for reasons unclear to me (it's an architecture with 32-bit pointers).

PR #1959, opened 2021-06-18
nrf52840: add support for flashing with the BOSSA tool

aykevl:
This only works with a custom bossac build from Arduino, not with the upstream version. It avoids needing the manual "double tap" to enter bootloader mode before flashing firmware.

Untested. Can somebody with an Arduino Nano 33 BLE please test this PR?


ysoldak:
Verified, works with my Arduino Nano 33 BLE Sense.
In both app and bootloader modes.
GJ!

aykevl:
Thank you for testing!

deadprogram:
On a roll! Merging, thanks @aykevl and @ysoldak

PR #1960, opened 2021-06-18
cortexm: bump default stack size to 2048 bytes

aykevl:
Previously it was 1024 bytes, which occasionally ran into a stack overflow. I hope that 2048 bytes will be enough for most purposes.

I've also removed some 2048-byte stack size settings in JSON files, which are unnecessary now that the parent (cortex-m.json) sets them.

Also see https://github.com/tinygo-org/tinygo/pull/1920, which does this for a limited number of boards.


deadprogram:
:two: :zero: :four: :eight: :heavy_check_mark:

PR #1962, opened 2021-06-19
Add crypto/rand package

aykevl:
This package provides access to an operating system resource
(cryptographic numbers) and so needs to be replaced with a TinyGo
version that does this in a different way.

I've made the following choices while adding this feature:

  - I'm (so far) only using syscalls (getrandom/getentropy) instead of
    /dev/random, because those syscalls are easier to use and more
    reliable.
  - I have implemented the WASI version using the getentropy call from
    wasi-libc that has the same interface as the getentropy call on
    MacOS, so that only a single implementation needs to be written and
    maintained.
  - I've left bare-metal implementations to a future patch. This because
    it's hard to reliably get cryptographically secure random numbers on
    embedded devices: most devices do not have a hardware PRNG for this
    purpose.

I've also fixed a bug in the interp package that was triggered by the crypto/rand package.

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -129,7 +129,7 @@ func Run(mod llvm.Module, debug bool) error {
 
 	// Update all global variables in the LLVM module.
 	mem := memoryView{r: r}
-	for _, obj := range r.objects {
+	for i, obj := range r.objects {
 		if obj.llvmGlobal.IsNil() {
 			continue
 		}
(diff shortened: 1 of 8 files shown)


kevlar700:
Out of interest. Is there a way to manually push bytes from Trngs found on many newer embedded chips?

aykevl:
Updated to hopefully fix the issue on older Linux versions (that don't have `getrandom` yet).

> Out of interest. Is there a way to manually push bytes from Trngs found on many newer embedded chips?

You can replace the `rand.Reader` variable, which will have this effect. You can replace it with a version that reads from the chip's RNG.

deadprogram:
Very cool work @aykevl this is pretty exciting! :black_joker: 

Now merging.

PR #1964, opened 2021-06-21
wasi: remove wasm build tag

aykevl:
The wasm build tag together with GOARCH=arm was causing problems in the internal/cpu package. In general, I think having two architecture build tag will only cause problems (in this case, wasm and arm) so I've removed the wasm build tag and replaced it with tinygo.wasm.

This is similar to the tinygo.riscv build tag, which is used for older Go versions that don't yet have RISC-V support in the standard library (and therefore pretend to be GOARCH=arm instead).

The situation after this PR:

* `-target=wasm` will keep using GOOS=js GOARCH=wasm
* `-target=wasi` already had GOOS=linux GOARCH=arm, but will have the wasm build tag removed for compatibility with the internal/cpu package.

Once WASI gets fully supported by the Go standard library, we can remove this workaround and use GOOS=wasi GOARCH=wasm, which would be the correct way to do this.

### DIFF
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -89,7 +89,7 @@ func (c *Config) NeedsStackObjects() bool {
 	switch c.GC() {
 	case "conservative", "extalloc":
 		for _, tag := range c.BuildTags() {
-			if tag == "wasm" {
+			if tag == "tinygo.wasm" {
 				return true
 			}
 		}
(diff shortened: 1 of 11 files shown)


deadprogram:
@aykevl thanks for the change and @fgsch thank you for testing.

Now merging.

PR #1965, opened 2021-06-21
Add more stubs for 1.15

fgsch:
Fix importing net/http.


src/os/file_go_other.go
```diff
@@ -44,3 +44,8 @@ const (
 func (m FileMode) IsDir() bool {
 	return false
 }
+
+// IsRegular is a stub, always returning false
+func (m FileMode) IsRegular() bool {
+	return false
```
aykevl, line 50: This is `m & ModeType == 0`.

PR #1976, opened 2021-06-30
machine/feather-nrf52: fix pin definition of uart

sago35:
The pin settings for TX and RX of the UART were reversed, which has been corrected.
Also, the DefaultUART setting has been added.

Notes:
The adafruit arduino code does show D0 of feather-nrf52840(-sense) as TX.
D0 of feather-m4 is RX.

Also, considering the position of the D2 pin, the pin definition of feather-nrf52840(-sense) seems to be wrong.
However, perhaps the pin names should be adapted to the Arduino.

https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/variants/feather_nrf52840_express/variant.cpp#L29-L30
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/variants/feather_nrf52840_sense/variant.cpp#L29-L30

https://github.com/adafruit/ArduinoCore-samd/blob/master/variants/feather_m4/variant.cpp#L37-L38

### DIFF
--- a/src/machine/board_feather-nrf52840-sense.go
+++ b/src/machine/board_feather-nrf52840-sense.go
@@ -72,8 +72,8 @@ const (
 
 // UART0 pins (logical UART1)
 const (
-	UART_RX_PIN = D0
-	UART_TX_PIN = D1
+	UART_RX_PIN = D1
+	UART_TX_PIN = D0
 )
 
 // I2C pins
@@ -98,4 +98,8 @@ const (
 var (
 	usb_VID uint16 = 0x239A
 	usb_PID uint16 = 0x8088
+)
+
+var (
+	DefaultUART = UART0
 )
--- a/src/machine/board_feather-nrf52840.go
+++ b/src/machine/board_feather-nrf52840.go
@@ -72,8 +72,8 @@ const (
 
 // UART0 pins (logical UART1)
 const (
-	UART_RX_PIN = D0
-	UART_TX_PIN = D1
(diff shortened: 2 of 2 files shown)


sago35:
I used the following code and checked it with feather-nrf52840 and feather-nrf52840-sense.
This PR makes the TX and RX follow the silk of the board.

I believe this PR is correct, but I'm hoping someone can confirm that I'm not mistaken.

image

```go
// This is a echo console running on the device UART.
// Connect using default baudrate for this hardware, 8-N-1 with your terminal program.
package main

import (
	"machine"
	"time"
)

var (
	uart = machine.UART0
	tx   = machine.UART_TX_PIN
	rx   = machine.UART_RX_PIN
)

func main() {
	uart.Configure(machine.UARTConfig{TX: tx, RX: rx})

	input := make([]byte, 64)
	i := 0
	for {
		if uart.Buffered() > 0 {
			data, _ := uart.ReadByte()

			switch data {
			case 13:
				// return key
				uart.Write([]byte("\r\n"))
				uart.Write([]byte("You typed: "))
				uart.Write(input[:i])
				uart.Write([]byte("\r\n"))
				i = 0
			default:
				// just echo the character
				uart.WriteByte(data)
				input[i] = data
				i++
			}
		}
		time.Sleep(10 * time.Millisecond)
	}
}
```

deadprogram:
Looks good to me, thanks @sago35 now merging.

PR #1984, opened 2021-07-04
cortexm: fix stack overflow because of unaligned stacks

aykevl:
On ARM, the stack has to be aligned to 8 bytes on function calls, but
not necessarily within a function. Leaf functions can take advantage of
this by not keeping the stack aligned so they can avoid pushing one
register. However, because regular functions might expect an aligned
stack, the interrupt controller will forcibly re-align the stack when an
interrupt happens in such a leaf function (controlled by the STKALIGN
flag, defaults to on). This means that stack size calculation (as used
in TinyGo) needs to make sure this extra space for stack re-alignment is
available.

This commit fixes this by aligning the stack size that will be used for
new goroutines.

Additionally, it increases the stack canary size from 4 to 8 bytes, to
keep the stack aligned. This is not strictly necessary but is required
by the AAPCS so let's do it anyway just to be sure.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -953,15 +953,19 @@ func modifyStackSizes(executable string, stackSizeLoads []string, stackSizes map
 		if fn.stackSizeType == stacksize.Bounded {
 			stackSize := uint32(fn.stackSize)
 
-			// Adding 4 for the stack canary. Even though the size may be
-			// automatically determined, stack overflow checking is still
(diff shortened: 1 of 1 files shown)


oflebbe:
Works for me as well  +1

deadprogram:
Thanks for the fix @aykevl and thanks @oflebbe for the review. Now merging.

Issue #1884, opened 2021-05-15
array indexes ignored when initializing pointers to array elements at package level


knieriem:
When I'm initializing a pointer to an array element at package level the following way,

```go
var a = [4]int{1, 2, 3, 4}

var p = &a[2]
```
`p` will point to a[0] instead of a[2] starting with tinygo v0.17.0. The array index, in this example: 2, appears to be ignored, and replaced by zero.

I am observing this for initializations at package level and within init, but not for pointers initialized within functions. I'm using the following tinygo versions:

* 0.18.0 linux/amd64 (using go version go1.16 and LLVM version 11.1.0)
* 0.17.0 linux/amd64 (using go version go1.16 and LLVM version 11.1.0)
* 0.16.0 linux/amd64 (using go version go1.15.6 and LLVM version 10.0.1)

Could there be a problem with my LLVM packages? I'm using Debian packages from apt.llvm.org/buster/.
I can see the behaviour when running this program. It initializes pointer variables at
various levels to the same array element, and checks whether each variable contains the correct pointer:

```go
package main

var (
	a = [4]int{1, 2, 3, 4}

	ptop = &a[2]

	pinit *int

	pfunc *int
)

func init() {
	pinit = &a[2]
}

func f() {
	pfunc = &a[2]
}

func main() {
	plocal := &a[2]
	f()
	test(ptop, "package")
	test(pinit, "init")
	test(pfunc, "function")
	test(plocal, "local")
	test(&a[2], "value")
}

func test(p *int, level string) {
	status := "OK"
	if p == &a[0] {
		status = "ERR"
	}
	println(p, *p, level, status)
```

With tinygo v0.16 it prints the value of the third position, "3", and "OK" in all cases, with tinygo v0.17 and v0.18 the output is:

    0x0000000000404068 1 package ERR
    0x0000000000404068 1 init ERR
    0x0000000000404078 3 function OK
    0x0000000000404078 3 local OK
    0x0000000000404078 3 value OK

knieriem:
btw, the first revision, where the new behaviour starts, appears to be 30df912565c2.

knieriem:
A partial fix appears to be to change `llvm.Type{}` to `llvmType` in interp/memory.go:912. Possibly a typo.

```go
	case llvm.PointerTypeKind:
		if v.buf[0] > 255 {
			// This is a regular pointer.
[...]

PR #1986, opened 2021-07-08
interp: don't ignore array indices for untyped objects

aykevl:
This fixes https://github.com/tinygo-org/tinygo/issues/1884.
My original plan to fix this was much more complicated, but then I
realized that the output type doesn't matter anyway and I can simply
cast the type to an *i8 and perform a GEP on that pointer.


deadprogram:

Issue #1987, opened 2021-07-09
machine/rp2040: SPI implementation.


soypat:
I was wondering if anyone was working on implementing SPI on the pico's microcontroller. If not I can give it a try.

Solved by https://github.com/tinygo-org/tinygo/pull/1991

deadprogram:
Released with 0.20.0 so now closing. Thank you!

PR #1991, opened 2021-07-09
rp2040: SPI implementation

soypat:
Not tested yet. I've never contributed before and looking forward to learning what the workflow looks like. Solves #1987

### DIFF
--- a/src/machine/board_pico.go
+++ b/src/machine/board_pico.go
@@ -37,3 +37,20 @@ const (
 	// Onboard crystal oscillator frequency, in MHz.
 	xoscFreq = 12 // MHz
 )
+
+// SPI default pins
+const (
+	// Default Serial Clock Bus 0 for SPI communications
+	SPI0_SCK_PIN = GPIO18
+	// Default Serial Out Bus 0 for SPI communications
+	SPI0_SDO_PIN = GPIO19 // Tx
+	// Default Serial In Bus 0 for SPI communications
+	SPI0_SDI_PIN = GPIO16 // Rx
+
+	// Default Serial Clock Bus 1 for SPI communications
+	SPI1_SCK_PIN = GPIO10
+	// Default Serial Out Bus 1 for SPI communications
+	SPI1_SDO_PIN = GPIO11 // Tx
+	// Default Serial In Bus 1 for SPI communications
+	SPI1_SDI_PIN = GPIO12 // Rx
+)
--- a/src/machine/machine_rp2040_spi.go
+++ b/src/machine/machine_rp2040_spi.go
@@ -1,0 +1,361 @@
+// +build rp2040
+
+package machine
+
+import (
+	"device/rp"
+	"errors"
+)
+
+// SPI on the RP2040
+var (
+	SPI0  = &_SPI0
+	_SPI0 = SPI{
+		Bus: rp.SPI0,
+	}
+	SPI1  = &_SPI1
+	_SPI1 = SPI{
+		Bus: rp.SPI1,
+	}
+)
+
+// SPIConfig is used to store config info for SPI.
+type SPIConfig struct {
+	Frequency uint32
+	// LSB not supported on rp2040.
+	LSBFirst bool
+	// Mode's two most LSB are CPOL and CPHA. i.e. Mode==2 (0b10) is CPOL=1, CPHA=0
+	Mode uint8
+	// Number of data bits per transfer. Valid values 4..16. Default and recommended is 8.
+	DataBits uint8
+	// Serial clock pin
+	SCK Pin
+	// TX or Serial Data Out (MOSI if rp2040 is master)
+	SDO Pin
+	// RX or Serial Data In (MISO if rp2040 is master)
+	SDI Pin
+}
+
+var (
+	ErrLSBNotSupported    = errors.New("SPI LSB unsupported on PL022")
+	ErrTxInvalidSliceSize = errors.New("SPI write and read slices must be same size")
+	ErrSPITimeout         = errors.New("SPI timeout")
+	ErrSPIBaud            = errors.New("SPI baud too low or above 66.5Mhz")
+)
+
+type SPI struct {
+	Bus *rp.SPI0_Type
+}
+
+// time to wait on a transaction before dropping. Unit in Microseconds for compatibility with ticks().
+const _SPITimeout = 10 * 1000 // 10 ms
+
+// Tx handles read/write operation for SPI interface. Since SPI is a syncronous write/read
+// interface, there must always be the same number of bytes written as bytes read.
+// The Tx method knows about this, and offers a few different ways of calling it.
+//
+// This form sends the bytes in tx buffer, putting the resulting bytes read into the rx buffer.
+// Note that the tx and rx buffers must be the same size:
+//
+// 		spi.Tx(tx, rx)
+//
+// This form sends the tx buffer, ignoring the result. Useful for sending "commands" that return zeros
+// until all the bytes in the command packet have been received:
+//
+// 		spi.Tx(tx, nil)
+//
+// This form sends zeros, putting the result into the rx buffer. Good for reading a "result packet":
+//
+// 		spi.Tx(nil, rx)
+//
+// Remark: This implementation (RP2040) allows reading into buffer with a custom repeated
+// value on tx.
+//
+// 		spi.Tx([]byte{0xff}, rx) // may cause unwanted heap allocations.
+//
+// This form sends 0xff and puts the result into rx buffer. Useful for reading from SD cards
+// which require 0xff input on SI.
+func (spi SPI) Tx(w, r []byte) (err error) {
+	switch {
+	case w == nil:
+		// read only, so write zero and read a result.
+		err = spi.rx(r, 0)
+	case r == nil:
+		// write only
+		err = spi.tx(w)
+	case len(w) == 1 && len(r) > 1:
+		// Read with custom repeated value.
+		err = spi.rx(r, w[0])
+	default:
+		// write/read
+		err = spi.txrx(w, r)
+	}
+	return err
+}
+
+// Write a single byte and read a single byte from TX/RX FIFO.
+func (spi SPI) Transfer(w byte) (byte, error) {
+	var deadline = ticks() + _SPITimeout
+	for !spi.isWritable() {
+		if ticks() > deadline {
+			return 0, ErrSPITimeout
+		}
+	}
+
+	spi.Bus.SSPDR.Set(uint32(w))
+
+	for !spi.isReadable() {
+		if ticks() > deadline {
+			return 0, ErrSPITimeout
+		}
+	}
+	return uint8(spi.Bus.SSPDR.Get()), nil
+}
+
+func (spi SPI) SetBaudRate(br uint32) error {
+	const freqin uint32 = 125 * MHz
+	const maxBaud uint32 = 66.5 * MHz // max output frequency is 66.5MHz on rp2040. see Note page 527.
+	// Find smallest prescale value which puts output frequency in range of
+	// post-divide. Prescale is an even number from 2 to 254 inclusive.
+	var prescale, postdiv uint32
+	for prescale = 2; prescale < 255; prescale += 2 {
+		if freqin < (prescale+2)*256*br {
+			break
+		}
+	}
+	if prescale > 254 || br > maxBaud {
+		return ErrSPIBaud
+	}
+	// Find largest post-divide which makes output <= baudrate. Post-divide is
+	// an integer in the range 1 to 256 inclusive.
+	for postdiv = 256; postdiv > 1; postdiv-- {
+		if freqin/(prescale*(postdiv-1)) > br {
+			break
+		}
+	}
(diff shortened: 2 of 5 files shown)


soypat:
Let me know of any comments so I can apply fixes and suggestions to i2c and PWM implementation branches!

efectn:
Have you tested yet? If not, i can test when i'm free.

soypat:
> Have you tested yet? If not, i can test when i'm free.

@efectn  I have not. Having trouble building TinyGo :(

soypat:
I've managed to build tinygo and flash programs to the pico. Currently dont have SPI devices on hand so I'll probably be PRing the i2c branch as soon as I get it working and then fix this branch (already noticed a couple bugs in how I use ReplaceBits)

soypat:
So I have not gotten the driver working. I'm going to take a break from this driver and try the PWM out. I've managed to get the Loopback function working but struggling reading data from my SPI device (_see details below_).

<details><summary>Example program with loopback test</summary>

```go
package main

import (
	"device/rp"
	"machine"
	"time"
)

var (
	LED  = machine.GPIO25
	b    = machine.SPI{Bus: rp.SPI0}
	uart = machine.UART0
	MISO = machine.GPIO4
	CS   = machine.GPIO5
	SCK  = machine.GPIO6
	MOSI = machine.GPIO7
)

func main() {
	// configure UART to print stuff
	{
		err := uart.Configure(machine.UARTConfig{BaudRate: 9600})
		if err != nil {
			panic(err)
		}
	}
	println("start program")
	time.Sleep(time.Millisecond)

	err := b.Configure(machine.SPIConfig{Frequency: 8.1e6,
		SDO: MOSI,
		SDI: MISO,
		SCK: SCK})

	if err != nil {
		println(err.Error())
	} else {
		println("spi baud:", b.GetBaudRate())
	}
	// Loopback test.
	b.Bus.SSPCR1.SetBits(rp.SPI0_SSPCR1_LBM)
	for i := 0; i < 10; i++ {
		ret, err := b.Transfer(uint8(i))
		if err != nil {
			println(err.Error())
		}
		println(i, ret)
	}
	// Output:
	// 0 0
	// 1 1
	// 2 2
	// 3 3
	// 4 4
	// 5 5
	// 6 6
	// 7 7
	// 8 8
	// 9 9
	// End loopback test.
	b.Bus.SSPCR1.ClearBits(rp.SPI0_SSPCR1_LBM)

	LED.Configure(machine.PinConfig{Mode: machine.PinOutput})
	CS.Configure(machine.PinConfig{Mode: machine.PinInputPullup})
	CS.Configure(machine.PinConfig{Mode: machine.PinOutput})
	CS.High() // Motorola SPI has Active Low.
	// Do SPI stuff
	for { // blinky when done.
		time.Sleep(time.Second)
		LED.Set(!LED.Get())
	}
}
```
</details>

src/machine/machine_rp2040_spi.go
```diff
@@ -0,0 +1,326 @@
...
+	// Serial clock pin
+	SCK Pin
+	// TX or Serial Data Out (MOSI if rp2040 is master)
+	SDO Pin
+	// RX or Serial Data In (MISO if rp2040 is master)
+	SDI Pin
+}
+
+var (
+	errLSBNotSupported    = errors.New("SPI LSB unsupported on PL022")
```
deadprogram: This should be capitalized in order for it to be exported for callers to use.

src/machine/machine_rp2040_spi.go
```diff
@@ -0,0 +1,326 @@
...
+	ErrTxInvalidSliceSize = errors.New("SPI write and read slices must be same size")
+	ErrSPITimeout         = errors.New("SPI timeout")
+	errSPIBaud            = errors.New("SPI baud too low or above 66.5Mhz")
```
deadprogram: Same as above.

src/machine/machine_rp2040_spi.go
```diff
@@ -0,0 +1,326 @@
...
+		config.DataBits = 8
+	}
+	if config.Frequency == 0 {
+		config.Frequency = defaultBaud
+	}
+	// SPI pin configuration
+	config.SCK.setFunc(fnSPI)
+	config.SDO.setFunc(fnSPI)
+	config.SDI.setFunc(fnSPI)
+	// TODO verify this code below.
```
deadprogram: Should this be removed, or is it needed?

src/machine/machine_rp2040_spi.go
```diff
@@ -0,0 +1,326 @@
...
+	default:
+		// write/read
+		err = spi.txrx(w, r)
+	}
+	return err
+}
+
+// Write a single byte and read a single byte from TX/RX FIFO.
+func (spi SPI) Transfer(w byte) (byte, error) {
+	for !spi.isWritable() {
```
deadprogram, line 99: This would be a good place for a timeout if the port is not writable after a predefined period of time.

src/machine/machine_rp2040_spi.go
```diff
@@ -0,0 +1,326 @@
...
+	}
+
+	spi.Bus.SSPDR.Set(uint32(w))
+
+	for !spi.isReadable() {
```
deadprogram, line 107: Same as above, we do not want to get looked in if it never is readable.

src/machine/machine_rp2040_spi.go
```diff
@@ -0,0 +1,326 @@
...
+		for !spi.isWritable() {
+		}
+		spi.Bus.SSPDR.Set(uint32(tx[i]))
+	}
+	// Drain RX FIFO, then wait for shifting to finish (which may be *after*
+	// TX FIFO drains), then drain RX FIFO again
+	for spi.isReadable() {
+		spi.Bus.SSPDR.Get()
+	}
+	for spi.isBusy() {
```
deadprogram, line 293: Timeout would be helpful here.

src/machine/machine_rp2040_spi.go
```diff
@@ -0,0 +1,326 @@
...
+func (spi SPI) txrx(tx, rx []byte) error {
+	plen := len(tx)
+	if plen != len(rx) {
+		return ErrTxInvalidSliceSize
+	}
+	// Never have more transfers in flight than will fit into the RX FIFO,
+	// else FIFO will overflow if this code is heavily interrupted.
+	const fifoDepth = 8
+	var rxleft, txleft = plen, plen
+	for txleft != 0 || rxleft != 0 {
```
deadprogram: A timeout would also be good here.

src/machine/machine_rp2040_spi.go
```diff
@@ -0,0 +1,326 @@
...
+	const freqin uint32 = 125 * MHz
+	prescale := spi.Bus.SSPCPSR.Get()
+	postdiv := ((spi.Bus.SSPCR0.Get() & rp.SPI0_SSPCR0_SCR_Msk) >> rp.SPI0_SSPCR0_SCR_Pos) + 1
+	return freqin / (prescale * postdiv)
+}
+
+// Configure is intended to setup/initialize the SPI interface.
+// Default baudrate of 115200 is used if Frequency == 0. Default
+// word length (data bits) is 8. SPI0 default bus is
+//  SCK = GPIO18
```
deadprogram: This info should be in the board file for the rp2040 so it can be more easily found.

src/machine/machine_rp2040_spi.go
```diff
@@ -0,0 +1,326 @@
...
+//  SDO = GPIO19
+//  SDI = GPIO16
+// No pin configuration is needed of SCK, SDO and SDI needed after calling Configure.
+func (spi SPI) Configure(config SPIConfig) error {
+	const defaultBaud uint32 = 115200
+	if config.SCK == 0 {
+		switch spi.Bus {
+		case rp.SPI0:
+			// These pins appear darkened "default" on pico pinout pictures on the web.
+			config.SCK = GPIO18
```
deadprogram: This will overwrite any values passed in by the programmer. This means a board such as the RP2040 Nano Connect will not be able to use the onboard WiFiNina chip.

deadprogram:
In order to test this on my rp2040 pico board, I had to add the following code:

```
diff --git a/src/machine/board_pico.go b/src/machine/board_pico.go
index 3a4d67b..e60fbec 100644
--- a/src/machine/board_pico.go
+++ b/src/machine/board_pico.go
@@ -2,6 +2,10 @@
 
 package machine
 
+import (
+       "device/rp"
+)
+
 // GPIO pins
 const (
        GP0  Pin = GPIO0
@@ -37,3 +41,18 @@ const (
        // Onboard crystal oscillator frequency, in MHz.
        xoscFreq = 12 // MHz
 )
+
+// SPI pins
+const (
+       SPI0_SCK_PIN Pin = GPIO18
+       SPI0_SDO_PIN Pin = GPIO19
+       SPI0_SDI_PIN Pin = GPIO16
+)
+
+// SPI on the RP2040
+var (
+       SPI0  = &_SPI0
+       _SPI0 = SPI{
+               Bus: rp.SPI0,
+       }
+)
```

`SPI0` needs to be hooked up to a specific bus for it to be able to be used. The pin definitions are just added for clarity. See my comments above.

deadprogram:
By the way, the SPI interface worked on my pico using an APA102 LED strip. Good work @soypat!

src/machine/machine_rp2040_spi.go
soypat: I'm not sure. Pico-sdk does not configure SCK, but then again they do skip some things the datasaheet strongly recommends.  If we could get eyes from someone familiar with SPI...

From the datasheet: 
_When CPOL clock polarity control bit is LOW, it produces a steady state LOW value on the SCK pin. If the CPOL clock polarity bit is high, a steady statew high value is placed on the SCK pin when data is not being transferred._

src/machine/machine_rp2040_spi.go
soypat: The pico-sdk does not use timeouts in SPI. From what I've tested this function does **not** hang if there is nothing connected, then again it would not hurt to add a timeout.

src/machine/machine_rp2040_spi.go
soypat: Should I create a `board_rp2040.go` file? Or do you mean `board_pico.go`?

src/machine/machine_rp2040_spi.go
soypat: These are predetermined "default" values in case a zero-valued `SPIConfig` struct is passed to `Configure`. I'm not against removing it but it's nice to have a shorthand initialization, I know I've spent too much time mixing up MISO and MOSI.

soypat: Alright, I think I understand the importance of changing this now. The issue with defining pins like so:
```go
       SPI0_SCK_PIN Pin = GPIO18
       SPI0_SDO_PIN Pin = GPIO19
       SPI0_SDI_PIN Pin = GPIO16
```
is that these are not the only SPI0 pins available. Maybe we should call them `DEFAULT_SPI0_SCK` and such?

src/machine/machine_rp2040_spi.go
soypat: ~From what I've tested this function does not hang~ scratch that, today it hanged while testing the ENC28J60 under suboptimal conditions. I'll fix this.

src/machine/machine_rp2040_spi.go
deadprogram: Ideally those constants should be defined in the board file, and then used in the machine file. That way each board can specify which pins are assigned to which function. A bunch of the Adafruit SAMD21/51 boards are setup this way.

src/machine/machine_rp2040_spi.go
deadprogram: This should be controlled entirely by the PIO function for SPI in this case. The datasheet is just describing what it is going to be doing in order to support whichever of the 4 modes you are selecting. That means you can remove, please.

src/machine/board_pico.go
```diff
@@ -37,3 +37,20 @@ const (
 	// Onboard crystal oscillator frequency, in MHz.
 	xoscFreq = 12 // MHz
 )
+
+// SPI default pins
+const (
+	// Serial Clock Bus 0 for SPI communications
+	Default_SPI0_SCK = GPIO18
```
deadprogram: It would be better to use the same naming convention for these e.g. `SPI0_SCK_PIN` just for consistency's sake. If we want to do something like change the naming, best to do that to all boards at the same time in a PR specific to that idea so we can get agreement from the community.

soypat: @deadprogram Consider it done!

deadprogram:
I think you still need to add the following to `machine_rp2040.go`:

```go
// SPI on the RP2040
var (
	SPI0  = &_SPI0
	_SPI0 = SPI{
		Bus: rp.SPI0,
	}
	SPI1  = &_SPI1
	_SPI1 = SPI{
		Bus: rp.SPI1,
	}
)
```

Add that, and squash everything into a single commit, and I think we are about ready to merge this! :smile_cat:

deadprogram:
Thanks very much for all the work on this PR! Now merging.

Issue #1990, opened 2021-07-09
Another memory corruption problem?


fgsch:
The code below prints:

`len(b)=55265: "foo" -> ""`

Tinygo version:

`tinygo version 0.19.0 darwin/amd64 (using go version go1.16.5 and LLVM version 11.0.0)`

Code:

```
package main

import (
	"fmt"
	"os"
)

func main() {
	pre := os.Getenv("PATH")
	b := make([]byte, 55265)
	post := os.Getenv("PATH")
	fmt.Printf("len(b)=%d: %q -> %q\n", len(b), pre, post)
}
```

Compilation and testing:

- `tinygo build -target=wasi -o bug.wasm bug.go`
- `wasmtime run  bug.wasm --env PATH=foo`

Changing the len of b will affect the output.

~This might be a duplicate of #1966.~

dkegel-fastly:
Thanks for finding a minimal reproducer!

aykevl:
Indeed, this minimal reproducer made it much easier to spot the underlying issue!

I think I know the bug, this should be a fix: https://github.com/tinygo-org/tinygo/pull/1992
I'm not sure whether this is related to #1966 but if you can test it, that would be great.

fgsch:
Thanks. #1992 fixes the issue in this PR but #1966 still fails.

fgsch:
Credits should go to @peterbourgon for repro.

fgsch:
This can be marked as next-release now too 🎉

deadprogram:
Released with 0.20.0 so now closing. Thank you!

PR #1992, opened 2021-07-10
wasm: override dlmalloc heap implementation from wasi-libc

aykevl:
These two heaps conflict with each other, so that if any function uses
the dlmalloc heap implementation it will eventually result in memory
corruption.

This commit fixes this by implementing all heap-related functions. This
overrides the functions that are implemented in wasi-libc. That's why
all of them are implemented (even if they just panic): to make sure no
program accidentally uses the wrong one.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -470,33 +470,10 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 		linkerDependencies = append(linkerDependencies, job)
 	}
 
-	// Add libc dependency if needed.
-	root := goenv.Get("TINYGOROOT")
-	switch config.Target.Libc {
-	case "picolibc":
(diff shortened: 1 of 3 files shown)


aykevl:
Huh, this appears to fail when using Go 1.15 or below? That doesn't make a lot of sense. I'll need to start digging...

aykevl:
Looks like I fixed it! The problem was that wasi-libc (a static library) was added before the main object file. Instead, static libraries should follow object files on the linker command line.

aykevl:
Not sure why the arduino-nano33 test is failing but it's probably just a flaky test.

deadprogram:
Had a power failure, so had to reset the board. Now passing.

fgsch:
I can confirm this fixes #1990.

deadprogram:
Thanks for the fix @aykevl and thanks @fgsch for verifying.

Now merging.

Issue #1981, opened 2021-07-01
panic: runtime error: nil pointer dereference


fgsch:
Full error:

```
panic: runtime error: nil pointer dereference
error: failed to run compiled binary /some/path/main: signal: abort trap
```

TinyGo version:

`tinygo version 0.19.0 darwin/amd64 (using go version go1.16.5 and LLVM version 11.0.0)`

Command line:

`tinygo run main.go`

Repro:

```go
package main

import "flag"

var _ = flag.Bool("foo", false, "")

func main() {
}
```

Also fails with TinyGo 0.17.0 and 0.18.0.

Originally found in https://github.com/tinygo-org/tinygo/pull/1814. Reporting it here so it can be tracked independently.

fgsch:
Backtrace:

```
#1  0x00007ffff7df3859 in __GI_abort () at abort.c:79
#2  0x00000000004014f6 in runtime.runtimePanic (msg=...) at /home/fgsch/tinygo/src/runtime/panic.go:20
#3  0x00000000004017f4 in runtime.nilPanic () at /home/fgsch/tinygo/src/runtime/panic.go:32
#4  0x0000000000401c10 in (*flag.FlagSet).Var (f=0x0, value=..., name=..., usage=...) at /opt/go-1.16.4/src/flag/flag.go:862
#5  0x0000000000401a73 in runtime.initAll () at /home/fgsch/tinygo/src/runtime/runtime.go:33
#6  0x0000000000401a42 in runtime.run$1 () at /home/fgsch/tinygo/src/runtime/scheduler_any.go:19
#7  0x0000000000401a2c in <goroutine wrapper> () at /home/fgsch/tinygo/src/runtime/scheduler_any.go:18
#8  0x0000000000401c31 in tinygo_startTask () at /home/fgsch/tinygo/src/internal/task/task_stack_amd64.S:29
```

fgsch:
Similar crash when running under wasi:

```
panic: runtime error: nil pointer dereference
Error: failed to run main module `bug2.wasm`

Caused by:
    0: failed to invoke command default
    1: wasm trap: unreachable
       wasm backtrace:
           0:  0x6bf - runtime.abort
                           at /Users/fgsch/foss/tinygo/src/runtime/runtime_tinygowasm.go:55:6
                     - runtime.runtimePanic
                           at /Users/fgsch/foss/tinygo/src/runtime/panic.go:20:7
           1:  0x826 - runtime.nilPanic
[...]

Issue #1966, opened 2021-06-21
Memory corruption? 


fgsch:
While testing #1964 with the code further down, I got:

```
021/06/21 23:32:49 net/http: invalid Cookie.Domain "<lots of garbage>"; dropping domain attribute
"foo="
```

The same code compiled with ~`-gc leaking` or~ `-scheduler none` works as expected (just prints `"foo="`).

Code:

```go
package main

import (
        "fmt"
        "net/http"
)

func main() {
        c := http.Cookie{Name:"foo"}
        fmt.Printf("%q\n", c.String())
}
```

Command line:

`tinygo build -target wasi -o test.wasm test.go`

fgsch:
I can reproduced this issue with 0.17.0, 0.18.0 and 0.19.0.

fgsch:
Reduced repro (updated):

```go
package main

import "log"

type foo struct {
	bar string
	qux string
}

func main() {
	func(f *foo) {
		log.Printf("")
		if len(f.bar) != 0 {
			panic("len(f.bar) not 0")
		}
	}(&foo{})
}
```

fgsch:
Fails with 0.17.0 and 0.18.0 as well.

aykevl:
I have further narrowed down this bug to this:

```go
package main

import (
	"os"
	"time"
)

var std = &Logger{out: os.Stderr}

type foo struct {
	a int
	b int
	c int
}

type Writer interface {
	Write([]byte) (int, error)
}

type Logger struct {
	out Writer
}

func main() {
	f(&foo{})
}

func f(f *foo) {
	println("f:", f.a, f.b, f.c)
	time.Time{}.Location().String()
	std.out.Write(nil)
	println("f:", f.a, f.b, f.c)
}
```

It prints the following:

```
$ tinygo run -target=wasi .
f: 0 0 0
f: 0 2 0
```

But clearly it should print the following:

```
$ go run .
f: 0 0 0
f: 0 0 0
```

If you want to help, it would be great if you could break this issue down even further. My strategy so far is inlining functions from packages, seeing whether the copy still reproduces the issue, and then incrementally removing pieces that don't matter while keeping the memory corruption.

fgsch:
I've fiddled with it a bit more and it looks like locking the mutex, and thus interacting with the scheduler, is related. 
Updated code below.

```go
package main

import (
	"os"
	"sync"
)

var std = &Logger{out: os.Stderr}

type foo struct {
	a int
	b int
[...]

PR #1994, opened 2021-07-11
wasm: strip debug info and compress relocations

aykevl:
This patch reduces code size significantly by stripping debug
information at the link stage and additionally compressing relocations
(which is only possible when debug information has been stripped).

For example, when compiling testdata/alias.go with -target=wasi, I get
the following results:

```
Previously, without -no-debug: 25326 bytes
Previously, with -no-debug:     5591 bytes
With this patch:                4762 bytes
With this patch and -no-debug:  4747 bytes
```

In other words, by stripping debug information and compressing
relocations by default, the resulting binary is smaller than it would
otherwise be even with -no-debug. Adding -no-debug may give a small
extra reduction in code size, but after this patch it's not very
significant.

This patch currently does not provide a way to opt out of this behavior.
If there is a need, an option can be added at a later time to change
this behavior.

@dkegel-fastly @fgsch because you've been rather involved in WebAssembly lately: what do you think of this change? Is it a good change of defaults?

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -100,7 +100,7 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 		AutomaticStackSize: config.AutomaticStackSize(),
 		DefaultStackSize:   config.Target.DefaultStackSize,
 		NeedsStackObjects:  config.NeedsStackObjects(),
-		Debug:              config.Debug(),
+		Debug:              true,
 		LLVMFeatures:       config.LLVMFeatures(),
 	}
 
@@ -537,6 +537,39 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 		// no library specified, so nothing to do
 	default:
 		return fmt.Errorf("unknown libc: %s", config.Target.Libc)
+	}
+
+	// Strip debug information with -no-debug.
+	if !config.Debug() {
+		for _, tag := range config.BuildTags() {
+			if tag == "baremetal" {
+				// Don't use -no-debug on baremetal targets. It makes no sense:
+				// the debug information isn't flashed to the device anyway.
+				return fmt.Errorf("stripping debug information is unnecessary for baremetal targets")
+			}
+		}
+		if config.Target.Linker == "wasm-ld" {
+			// Don't just strip debug information, also compress relocations
+			// while we're at it. Relocations can only be compressed when debug
+			// information is stripped.
+			ldflags = append(ldflags, "--strip-debug", "--compress-relocations")
+		} else {
+			switch config.GOOS() {
+			case "linux":
+				// Either real linux or an embedded system (like AVR) that
+				// pretends to be Linux. It's a ELF linker wrapped by GCC in any
+				// case.
+				ldflags = append(ldflags, "-Wl,--strip-debug")
+			case "darwin":
+				// MacOS (darwin) doesn't have a linker flag to strip debug
+				// information. Apple expects you to use the strip command
+				// instead.
+				return errors.New("cannot remove debug information: MacOS doesn't suppor this linker flag")
+			default:
+				// Other OSes may have different flags.
+				return errors.New("cannot remove debug information: unknown OS: " + config.GOOS())
+			}
+		}
 	}
 
 	// Create a linker job, which links all object files together and does some
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -209,9 +209,8 @@ func (c *Config) CFlags() []string {
 		cflags = append(cflags, "-nostdlibinc", "-Xclang", "-internal-isystem", "-Xclang", filepath.Join(root, "lib", "picolibc", "newlib", "libc", "include"))
 		cflags = append(cflags, "-I"+filepath.Join(root, "lib/picolibc-include"))
 	}
-	if c.Debug() {
-		cflags = append(cflags, "-g")
-	}
+	// Always emit debug information. It is optionally stripped at link time.
+	cflags = append(cflags, "-g")
 	return cflags
 }
 
@@ -250,8 +249,9 @@ func (c *Config) VerifyIR() bool {
 	return c.Options.VerifyIR
 }
 
-// Debug returns whether to add debug symbols to the IR, for debugging with GDB
-// and similar.
+// Debug returns whether debug (DWARF) information should be retained by the
+// linker. By default, debug information is retained but it can be removed with
+// the -no-debug flag.
 func (c *Config) Debug() bool {
 	return c.Options.Debug
 }
--- a/main.go
+++ b/main.go
@@ -1019,7 +1019,7 @@ func main() {
 	printStacks := flag.Bool("print-stacks", false, "print stack sizes of goroutines")
 	printAllocsString := flag.String("print-allocs", "", "regular expression of functions for which heap allocations should be printed")
 	printCommands := flag.Bool("x", false, "Print commands")
-	nodebug := flag.Bool("no-debug", false, "disable DWARF debug symbol generation")
+	nodebug := flag.Bool("no-debug", false, "strip debug information")
 	ocdCommandsString := flag.String("ocd-commands", "", "OpenOCD commands, overriding target spec (can specify multiple separated by commas)")
 	ocdOutput := flag.Bool("ocd-output", false, "print OCD daemon output during debug")
 	port := flag.String("port", "", "flash port (can specify multiple candidates separated by commas)")


dkegel-fastly:
Ideally tinygo might follow go's lead here, and aim for a good debugging experience for optimized programs ( https://blog.golang.org/debug-opt ).  Failing that, make sure it's easy to debug by adding an option.

Removing debug info, with no option to bring it back, feels... strange.  So, maybe make it its own option that can be turned on and off?   And/or potentially make it -opt 3 ?  I dunno much about tinygo optimization.  Bigger compilers have both lots of little optimizations which can be controlled individually, and are controlled grossly in groups with -O1, -O2, -O3, or -Os; it would be natural for tinygo to move a bit in that direction.

aykevl:
I agree that debug information should be available for most targets. However, WebAssembly is a bit different. Here are my considerations:

* Compressing relocations is only possible when debug information is stripped because of limitations in DWARF emission (there is no technical reason why this isn't possible, [RISC-V manages to do it](https://zhuanlan.zhihu.com/p/357112171)). This compression does save around 8% of code size (or around 3% after compression) so it's a worthwhile optimization in my opinion.
* WebAssembly is still hard to debug with DWARF, I believe only Chrome implements it.
* Most people will want small binaries by default.

That said, you've got me convinced that it should be an option (that is probably off by default on WebAssembly for the above reasons). This makes small binaries the default while leaving the option to debug those binaries if needed.

> Ideally tinygo might follow go's lead here, and aim for a good debugging experience for optimized programs ( https://blog.golang.org/debug-opt ).

This is actually somewhat unrelated. The blog post describes how to _improve_ debug information that's already present. In short, there is no reason why only unoptimized builds can be debugged (even though that's what many people think). While optimizations will make debugging harder, DWARF has been catching up and adding more and more features that allow it to better debug optimized programs - basically to catch up with modern optimizations.

Luckily TinyGo doesn't have to deal with this, mostly. LLVM is responsible for most optimizations and thus also responsible for preserving debug information. The choice that TinyGo can make is whether to emit it at all.

dkegel-fastly:
The blog post said Google wanted to make production binaries work well with debuggers.  That's on topic.

That said, there's lots of history for debugging defaulting to off back in the old days, so you wouldn't be going out on a huge limb.

FWIW I suspect server side wasm would be ok with larger files, and might usually prefer to keep the debug info around to get those sweet, sweet backtraces.

I found https://thenewstack.io/the-pain-of-debugging-webassembly/ interesting.  I do hope the situation gets better.

aykevl:
> FWIW I suspect server side wasm would be ok with larger files, and might usually prefer to keep the debug info around to get those sweet, sweet backtraces.

Are there any WebAssembly runtimes that support DWARF other than Chrome? I'm not aware of any but I could be wrong.

Note that there is also a "names" section that provides some symbol names and can thus give a rough backtrace (but without accounting for inlining unfortunately). If I remember correctly, Chrome has support for this section.

From [the blog post you mention](https://thenewstack.io/the-pain-of-debugging-webassembly/ ):

> Cool. Except a lot of WASM tools don’t account for things like byte offsets. For good reason, because WASM actually has two formats: a binary format and WAST, which is a text representation. Within DWARF, these two formats are no longer interchangeable in the presence of debug information, as otherwise the offsets would require rewriting. Bummer.

I'm actually considering doing something like this, for improved GC and scheduler support. While they kind of work at the moment, you can see from bugs like #1966 that they're not working very well. The reason is that WebAssembly is very limited compared to other instruction sets: it doesn't allow direct access to the stack (for security). While there are proposals that would allow TinyGo to work well here (with [stack switching](https://github.com/WebAssembly/stack-switching) and [using the host GC](https://github.com/WebAssembly/gc)), both are still proposals that see very little activity. By modifying WebAssembly after linking, both of these might be possible to implement in a more robust way.
That would unfortunately mean that debugging wouldn't be possible if these are used (requiring workarounds like `-gc=leaking` and `-scheduler=none`) until these features are implemented in WebAssembly runtimes.
I wish all of that wasn't necessary, but the current state of WebAssembly isn't great for languages that aren't C like (C/C++/Rust).

fgsch:
I agree with @dkegel-fastly; if anything this should be optional.

To illustrate this, trying the code in #1981, before this change:

```
Error: failed to run main module `bug2.wasm`

Caused by:
    1: wasm trap: unreachable
       wasm backtrace:
           0:  0x6bf - runtime.abort
                           at /Users/fgsch/foss/tinygo/src/runtime/runtime_tinygowasm.go:55:6
                     - runtime.runtimePanic
                           at /Users/fgsch/foss/tinygo/src/runtime/panic.go:20:7
           1:  0x826 - runtime.nilPanic
                           at /Users/fgsch/foss/tinygo/src/runtime/panic.go:32:14
           2:  0xcd0 - (*flag.FlagSet).Var
                           at /usr/local/Cellar/go/1.16.5/libexec/src/flag/flag.go:862:23
           3:  0xc83 - <unknown>!runtime.initAll
           4:  0xbfd - runtime.run$1
                           at /Users/fgsch/foss/tinygo/src/runtime/scheduler_any.go:19:10
           5:  0xbf3 - runtime.run
                           at /Users/fgsch/foss/tinygo/src/runtime/scheduler_any.go:18:2
           6:  0xbe3 - _start
                           at /Users/fgsch/foss/tinygo/src/runtime/runtime_wasm_wasi.go:21:5                      
```

After this change:

```
Error: failed to run main module `bug2.wasm`

Caused by:
    1: wasm trap: unreachable
       wasm backtrace:
           0:  0x5f1 - <unknown>!<wasm function 10>
           1:  0x723 - <unknown>!<wasm function 18>
           2:  0xaf2 - <unknown>!<wasm function 34>
           3:  0xab5 - <unknown>!<wasm function 33>
           4:  0xa4e - <unknown>!<wasm function 32>
           5:  0xa48 - <unknown>!<wasm function 31>
           6:  0xa40 - <unknown>!<wasm function 30>
```

aykevl:
Sweet! Which runtime is this? Is it public? It shows the source locations (and not just names) so it must be reading the DWARF sections.

> FWIW I suspect server side wasm would be ok with larger files, and might usually prefer to keep the debug info around to get those sweet, sweet backtraces.

Out of interest, what's the reason for looking into TinyGo here? Is it the reduced code size, WASI support, or something else?

fgsch:
@aykevl this is wasmtime with `WASMTIME_BACKTRACE_DETAILS=1`.

aykevl:
Cool! I just found out right before I saw your comment. I had to update wasmtime from v0.20.0 to v0.28.0 for it to work, though.

aykevl:
Interestingly, the `--strip-debug` flag also strips the name section which stores the names of functions. That kind of makes sense, but previously the names would be retained with `-no-debug` (and tools like wasmtime could at least show a limited backtrace).

aykevl:
I have an improved branch at https://github.com/tinygo-org/tinygo/compare/dev...debug-flag. It strips debug information at link time instead of at compile time and also replaces the `-no-debug` flag with `-debug=true/false`. The debug flag defaults to false for wasm, true for everything else.
Now I'm thinking: is this the right solution? With wasmtime supporting DWARF, maybe it makes sense to keep emitting debug information even on WebAssembly. What do you think?

dkegel-fastly:
Special cases may be confusing, seems like users who want the extra smallness will turn it on. 

Keeping the debug info around by default will be handy once more runtimes support backtraces.

So I'm in the "debuggable by default" camp, I guess.

dkegel-fastly:
Also, I'd just leave the -no-debug flag the way it is; it's shorter to type than -debug=false :-)

mathetake:
Sorry to weigh in here. As a user, I would give +1 to @dkegel-fastly and I would like to keep debug info by default. At least, we in Envoy/Proxy-Wasm project already have used name sections for emitting backtraces, so I definitely would love to keep name section at least. Anyway thanks for your efforts here to reduce the binary size, that would be awesome for everyone.

FWIW, we haven't support DWARF in Envoy proxy yet https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/149

aykevl:
Ok, you got me convinced. I'll keep the defaults as-is, with debug information. I've force-pushed an update.

> Also, I'd just leave the -no-debug flag the way it is; it's shorter to type than -debug=false :-)

It was necessary when the default depends on the target (for example, you can't re-enable debug information with `-no-debug` if it's disabled for wasm). But it's not necessary anymore if everything always emits debug information.

> Sorry to weigh in here.

I'm always interested in feedback especially from people affected by the change :)

dkegel-fastly:
That's more like it :-)

deadprogram:
@aykevl some merge conflicts are needing to be resolved, please.

deadprogram:
Reminder to @aykevl to please resolve merge conflict here. Thanks!

aykevl:
Rebased.

deadprogram:
Thanks @aykevl for working on this and thanks to everyone who helped provide feedback and guidance.

Now merging.

PR #1999, opened 2021-07-14
Fix bugs in interp package

aykevl:
This PR fixes two more or less separate bugs that I discovered while looking into https://github.com/tinygo-org/tinygo/issues/1981. Further testing shows that it also fixes https://github.com/tinygo-org/tinygo/issues/1966.


deadprogram:
:bug: :zap: good work @aykevl :detective: 

Merging.

Issue #1830, opened 2021-04-23
Using the regexp package gives either a stack overlow or "panic: expected exactly one index"


arp242:
With the following:

```
package main

import "regexp"

var reTest = regexp.MustCompile(`^\w[\w -]+:$`)

func main() {}
```

I get:

```
panic: expected exactly one index

goroutine 6 [running]:
github.com/tinygo-org/tinygo/interp.(*runner).runAtRuntime(0xc0000c1320, 0xc001271900, 0x36, 0x53, 0xc0021352c0, 0x3, 0x3, 0x1f0c0a0, 0xc000344108, 0x11, ...)
	/data/src/tinygo/interp/interpreter.go:940 +0xd25
github.com/tinygo-org/tinygo/interp.(*runner).run(0xc0000c1320, 0xc001271900, 0xc002d9c610, 0x8, 0x1f, 0xc0012591d0, 0xc0003441f8, 0x18, 0x0, 0xc002a972c0, ...)
	/data/src/tinygo/interp/interpreter.go:100 +0xa85a
github.com/tinygo-org/tinygo/interp.(*runner).run(0xc0000c1320, 0xc0012716d0, 0xc002a40e10, 0x5, 0x7, 0xc000bbfec0, 0xc000e23668, 0x14, 0x0, 0xc002f9d660, ...)
	/data/src/tinygo/interp/interpreter.go:481 +0x6473
github.com/tinygo-org/tinygo/interp.(*runner).run(0xc0000c1320, 0xc001271680, 0xc0029fc010, 0x6, 0x7, 0xc0008de420, 0xc00281bf90, 0x10, 0x0, 0xc002872a10, ...)
	/data/src/tinygo/interp/interpreter.go:481 +0x6473
github.com/tinygo-org/tinygo/interp.(*runner).run(0xc0000c1320, 0xc001271630, 0xc002801e90, 0x4, 0x7, 0xc0008bf6e0, 0xc0027d4f60, 0xc, 0x0, 0x3ab7f900, ...)
	/data/src/tinygo/interp/interpreter.go:481 +0x6473
github.com/tinygo-org/tinygo/interp.(*runner).run(0xc0000c1320, 0xc001270c30, 0xc002801910, 0x4, 0x7, 0xc0008b7200, 0xc0027d4780, 0x8, 0x0, 0x200000003, ...)
	/data/src/tinygo/interp/interpreter.go:481 +0x6473
[...]

Issue #1306, opened 2020-08-22
regexp compile failures on darwin/wasm


ribasushi:
**TLDR**
`regexp` is listed as fully implemented, yet code using it fails compilation.

**Version**
`tinygo version 0.14.0 darwin/amd64 (using go version go1.14.7 and LLVM version 10.0.0)`

**Failure**
```
~$ tinygo run tiny.go 
# command-line-arguments
/usr/local/go/src/regexp/syntax/parse.go:1535:5: interp: branch on a non-constant

traceback:
/usr/local/go/src/regexp/syntax/parse.go:774:28:
  %108 = call { %runtime._string, %runtime._interface } @"(*regexp/syntax.parser).parseClass"(%"regexp/syntax.parser"* %19, i8* %106, i64 %107, i8* undef, i8* undef), !dbg !1831
/usr/local/go/src/regexp/regexp.go:170:25:
  %4 = call { %"regexp/syntax.Regexp"*, %runtime._interface } @"regexp/syntax.Parse"(i8* %2, i64 %3, i16 %mode, i8* undef, i8* undef), !dbg !1739
/usr/local/go/src/regexp/regexp.go:133:16:
  %4 = call { %regexp.Regexp*, %runtime._interface } @regexp.compile(i8* %2, i64 %3, i16 212, i1 false, i8* undef, i8* undef), !dbg !1696
/usr/local/go/src/regexp/regexp.go:309:24:
  %4 = call { %regexp.Regexp*, %runtime._interface } @regexp.Compile(i8* %2, i64 %3, i8* undef, i8* undef), !dbg !1698
command-line-arguments/<init>:8:34:
  %0 = call %regexp.Regexp* @regexp.MustCompile(i8* getelementptr inbounds ([12 x i8], [12 x i8]* @"command-line-arguments.init$string", i32 0, i32 0), i64 12, i8* undef, i8* undef), !dbg !1692
```

**Expectation**
```
~$ go run tiny.go 
(?m)^([^\n])
(?m)^\s{0,12}([^\s\n\-])
(?m)^(\s*)\-\-
\bMaxPayload\b
```

**Testcase**
```
package main

import (
	"fmt"
	"regexp"
)

var indenter = regexp.MustCompile(`(?m)^([^\n])`)
var nonOptIndenter = regexp.MustCompile(`(?m)^\s{0,12}([^\s\n\-])`)
var dashStripper = regexp.MustCompile(`(?m)^(\s*)\-\-`)
[...]

PR #2001, opened 2021-07-15
Fix interp stack overflow with self-referencing data types

aykevl:
This PR partially fixes https://github.com/tinygo-org/tinygo/issues/1848. It's relatively big: it adds object layout information to the IR which is normally optimized away but is used by the interp package to avoid problematic cases like linked list, tree structures, etc.

It also improves https://github.com/tinygo-org/tinygo/issues/1306 and https://github.com/tinygo-org/tinygo/issues/1830 but doesn't yet fix them. Somehow they're still failing.

### DIFF
--- a/compiler/llvm.go
+++ b/compiler/llvm.go
@@ -1,6 +1,11 @@
 package compiler
 
 import (
+	"fmt"
+	"go/token"
+	"go/types"
+	"math/big"
+
 	"github.com/tinygo-org/tinygo/compiler/llvmutil"
 	"tinygo.org/x/go-llvm"
 )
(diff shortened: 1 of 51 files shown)


aykevl:
There appears to be one real bug in the arduino test:

```
error: interp: ptrtoint integer size does not equal pointer size
```

I'll need to look into that. Other than that, this PR passes all tests.

dkegel-fastly:
FWIW, after this patch, my app builds fine except that on slower systems (our CI), I get a build timeout, and have to raise the one minute timeout thusly:

```
sed -i 's,time.Minute,3 * time.Minute,' interp/interpreter.go
```

Time shows the compile took 2 minutes 25 seconds, so a 3 minute timeout should just do on the ci for this app today.

Making it adjustable from the commandline might be nice, otherwise I'll probably have to keep building from source after the release.

(And yes, I should figure out what takes so long to compile, and get rid of it... it is probably a regexp-compile-like-thing in an init function in a dependency.)

aykevl:
I've investigated the error in the Arduino Uno test, and I think I know what's going on. But not sure yet how to fix this.

Note to self: the problem is that `runtime.funcValue` has an `id` field that is of type `uintptr`, but a pointer to a global is stored in it for optimization reasons. I don't know why this triggers on the Arduino Uno test and not on WebAssembly where the same problem exists.

compiler/llvm.go
```diff
@@ -52,3 +57,199 @@ func (c *compilerContext) makeGlobalArray(buf []byte, name string, elementType l
...
+		return big.NewInt(1)
+	case llvm.StructTypeKind:
+		ptrs := big.NewInt(0)
+		if typ.StructName() == "runtime.funcValue" {
+			// Hack: the type runtime.funcValue contains an 'id' field which is
+			// of type uintptr, but before the LowerFuncValues pass it actually
+			// contains a pointer (ptrtoint) to a global. This trips up the
+			// interp package. Therefore, make the id field a pointer for now.
+			typ = c.ctx.StructType([]llvm.Type{c.i8ptrType, c.i8ptrType}, false)
+		}
```
aykevl, line 215: I've worked around the build failure for the Arduino Uno (which could also affect WebAssembly) using this hack. I don't like it, but I don't really see a cleaner way without some major refactoring. And this seems like a relatively small, unlikely to be problematic hack.

dkegel-fastly:
That's great, thanks.  (Unfortunately rebasing the branch on current dev appears to have yielded a compiler we can't use because of https://github.com/tinygo-org/tinygo/issues/2101, sigh.  Evidently the sands shifted on us.)

aykevl:
All tests are passing! Well, except for the circuitplay-express test which appears to be failing due to a hardware issue.
CircleCI looks like it's still in progress but that's only because I was messing with some CircleCI settings. All those tests have in fact long been passed.

@dkegel-fastly yeah but that must be a separate issue.

dgryski:
So, seems like this can be merged?

dgryski:
While testing golang.org/x/crypto, I think I found a bug traced to this branch:

```
package main

var testCases = []struct {
	msg []byte
	out [8]byte
}{
	{
		msg: []byte("Hi There"),
		out: [8]byte{
			0x87, 0xaa, 0x7c, 0xde, 0xa5, 0xef, 0x61, 0x9d,
		},
	},
}

func main() {
	println(testCases[0].out[0])
}
```

produces

```
# main
main/<init>:10:10: interp: pointer index with something other than a struct or array?
  store i8 -86, i8* %8, align 1, !dbg !35

traceback:
main/<init>:10:10:
  store i8 -86, i8* %8, align 1, !dbg !35
```

deadprogram:
Merge conflicts need resolution, please.

aykevl:
I've rebased this PR.

@dgryski Thank you! I've updated the PR to fix this issue. It was actually not directly in the PR but rather caused by this PR as a side effect. Basically, I fixed it by simplifying the code and thus eliminating the error entirely.

deadprogram:
@aykevl merge conflict needs resolution, please.

aykevl:
I have rebased the branch again.

deadprogram:
Seems like all tests failing in this branch.

aykevl:
I've updated the PR, it should be fixed now.

deadprogram:
All tests are now passing, and code looks good. Thanks @aykevl for working on this and thanks @dkegel-fastly and @dgryski for all the great work testing and reviewing.

Now merging!

PR #2003, opened 2021-07-16
Add a mechanism to disable interp passes

fgsch:
This would be handy when debugging code generation issues and should
not be used in normal circumstances.
Discussed with @aykevl

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -304,19 +304,21 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 					newGlobal.SetName(name)
 				}
 
-				// Try to interpret package initializers at compile time.
-				// It may only be possible to do this partially, in which case
-				// it is completed after all IR files are linked.
-				pkgInit := mod.NamedFunction(pkg.Pkg.Path() + ".init")
-				if pkgInit.IsNil() {
-					panic("init not found for " + pkg.Pkg.Path())
-				}
-				err := interp.RunFunc(pkgInit, config.DumpSSA())
-				if err != nil {
-					return err
-				}
-				if err := llvm.VerifyModule(mod, llvm.PrintMessageAction); err != nil {
-					return errors.New("verification error after interpreting " + pkgInit.Name())
+				if config.InterpPackage() {
+					// Try to interpret package initializers at compile time.
+					// It may only be possible to do this partially, in which case
+					// it is completed after all IR files are linked.
+					pkgInit := mod.NamedFunction(pkg.Pkg.Path() + ".init")
+					if pkgInit.IsNil() {
+						panic("init not found for " + pkg.Pkg.Path())
+					}
+					err := interp.RunFunc(pkgInit, config.DumpSSA())
+					if err != nil {
+						return err
+					}
+					if err := llvm.VerifyModule(mod, llvm.PrintMessageAction); err != nil {
+						return errors.New("verification error after interpreting " + pkgInit.Name())
+					}
 				}
 
 				// Run function passes for each function in the module.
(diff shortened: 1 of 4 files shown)


builder/build.go
```diff
@@ -231,19 +231,21 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
...
-				if pkgInit.IsNil() {
-					panic("init not found for " + pkg.Pkg.Path())
-				}
-				err := interp.RunFunc(pkgInit, config.DumpSSA())
-				if err != nil {
-					return err
-				}
-				if err := llvm.VerifyModule(mod, llvm.PrintMessageAction); err != nil {
-					return errors.New("verification error after interpreting " + pkgInit.Name())
+				if config.InterpPackage() {
```
aykevl, line 309: This configuration option has an effect on the cached bitcode for the package, so must be included in the `packageAction` struct. Otherwise you're going to end up with incorrect/stale caches. (If that's the case for you locally, you can run `tinygo clean` to clean it up).

compileopts/config.go
```diff
@@ -367,6 +367,22 @@ func (c *Config) LLVMFeatures() string {
 	return c.Options.LLVMFeatures
 }
 
+func (c *Config) InterpPackage() bool {
```
aykevl, line 456: Can you add some short documentation to these methods? Something like "InterpPackage returns whether the interp package should be run per package, according to the -interp-pass flag".

aykevl (commented):
In general this looks good, except for the two things below.

That said, I'm wondering how these debug options can best be separated from other options. Maybe by prefixing all these options with `-internal` for example, like `-internal-interp-pass` and `-internal-printir`. Regular users shouldn't use these options. Or even make them a bit harder to use using something like `-build-debug="interp-all,dumpssa"`. But that's for a separate issue/PR, not for this one.

aykevl:
CI is failing, probably because the `InterpPass` option isn't set when running the tests in main_test.go. One simple way to fix this is to also allow the empty string to indicate "all" in compileopts/config.go, or by reversing the logic:

```go

func (c *Config) InterpProgram() bool {
	switch c.Options.InterpPass {
	case "package", "none":
		return false
	}
	return true
}
```

compileopts/config.go
fgsch: Done.

builder/build.go
fgsch: Done.

fgsch:
Thank you for the feedback @aykevl. I've update the code.

deadprogram:
Any feedback on this now @aykevl?

dgryski:
On the subject of configuring `interp`, being able to customize the timeout value (currently one minute) would also be useful.

aykevl:
So I still don't think this change is correct in general. Some packages absolutely require the interp pass to run for correct functioning. Although, maybe this is only true for the runtime package and not for other packages.

One thing that would be correct is to disable the whole program interp pass (`interp.Run`), that one can I think be safely disabled. Maybe that already solves the biggest issues?

> being able to customize the timeout value (currently one minute) would also be useful.

I guess? But I think that would only be for debugging. If it runs more than a few seconds I'd argue that there must be a bug somewhere (even if it's just a performance bug). So, not sure about this but not against it either.

fgsch:
The idea behind this PR was to disable the interp passes for debugging, and not to generate fully functional code.
In other words, this is purely for debugging and should not  be used for anything else.

Would you be more confortable if we prefix the parameter with `-internal` as you suggested in https://github.com/tinygo-org/tinygo/pull/2003#pullrequestreview-709457492 or should I go ahead and remove the package option, only leaving the program option?

fgsch:
Thinking a bit more about this, perhaps one piece of information missing here is the fact that this was meant to be used together with `-printir` to capture the generated IR without the interp (package) pass.

PR #1193, opened 2020-06-25
runtime (gc): add lightweight `list` GC

niaow:
This saves about a kilobyte of flash, and attempts to address the AVR size increase issue in #1181. Similar to #1181 and the avr-libc allocator, this is a best-fit allocator.

PR #2004, opened 2021-07-18
runtime (gc): add a new "list" GC

niaow:
This memory manager uses a simple linked list to track active allocations, and measures the gaps between them to find available memory. Performance is not great for large heaps, but is sufficient for microcontrollers with single-digit kilobytes of memory.

This is effectively a simplified version of #1193 containing only the GC itself so that it is easier to review.

The `./testdata/gc.go` test was run on an Arduino Uno and passes. It is slow.

The false-positive rate problem still applies to microcontrollers where the address space is mostly used.

### DIFF
--- a/src/runtime/gc_list.go
+++ b/src/runtime/gc_list.go
@@ -1,0 +1,467 @@
+// +build gc.list
+
+// This memory manager is partially inspired by the memory allocator included in avr-libc.
+// Unlike avr-libc malloc, this memory manager stores an allocation list instead of a free list.
+// There is an overhead of 2 pointer-widths/allocation (up from avr-libc's 1 pointer-width/allocation).
+// The allocation list is stored in ascending address order.
+// The set of free spans is implicitly derived from the gaps between allocations.
+// This allocator has an effective allocation complexity of O(n) and worst-case GC complexity of O(n^2).
+// Due to architectural quirks, as well as tiny heaps, this should almost always be faster than the standard conservative collector on AVR.
+
+package runtime
+
+import (
+	"internal/task"
+	"runtime/interrupt"
+	"unsafe"
+)
+
+// Set gcDebug to true to print debug information.
+const (
+	gcDebug   = false   // print debug info
+	gcAsserts = gcDebug // perform sanity checks
+)
+
+// setHeapEnd is called to expand the heap. The heap can only grow, not shrink.
+func setHeapEnd(newHeapEnd uintptr) {
+	if gcAsserts && newHeapEnd <= heapEnd {
+		panic("gc: setHeapEnd didn't grow the heap")
+	}
+
+	heapEnd = newHeapEnd
+
+	// There is no other metadata to update since the free space is just computed from the heap bounds and the allocations list.
+}
+
+// Initialize the memory allocator.
+func initHeap() {
+	// No initialization is required: the only state is the allocation list, which starts empty.
+	// This function must exist anyway since other GCs need initialization.
+	// It should be inlined and optimized away by the compiler.
+}
+
+var gcRunning bool
+
+// alloc tries to find some free space on the heap, possibly doing a garbage
+// collection cycle if needed. If no space is free, it panics.
+//go:noinline
+func alloc(size uintptr) unsafe.Pointer {
+	if gcAsserts && gcRunning {
+		runtimePanic("alloc while GC is running")
+	}
+
+	if size == 0 {
+		return unsafe.Pointer(&zeroSizedAlloc)
+	}
+
+	var ranGC bool
+tryAlloc:
+	// Search for available memory.
+	alloc := findMem(size)
+	if alloc == nil {
+		// There is no free span large enough for the allocation.
+
+		if ranGC {
+			// Even after running the GC, there is not enough memory.
+			if growHeap() {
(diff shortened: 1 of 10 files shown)


niaow:
Size comparison:
```
[niaow@finch tinygo]$ build/tinygo build -target=arduino -size short -gc=conservative -o /tmp/gc.hex ./testdata/gc.go 
   3768      89     689 |    3857     778
[niaow@finch tinygo]$ build/tinygo build -target=arduino -size short -gc=list -o /tmp/gc.hex ./testdata/gc.go 
   2712      94     686 |    2806     780
```

src/runtime/gc_list.go
```diff
@@ -0,0 +1,460 @@
...
+		remainingAllocs.dump()
+	}
+
+	return remainingAllocs
+}
+
+// sort the allocation list in ascending address order.
+func (list *allocList) sort() {
+	// Sort the list by repeatedly moving the lowest-address element to the front.
+	// This is selection sort (https://en.wikipedia.org/wiki/Selection_sort).
```
aykevl, line 275: I'm slightly worried that a quadratic algorithm in a GC might be a problem. Maybe not on the AVR though with its very limited memory.

I think there is a way to avoid this sorting, at least if it's fine to limit the maximum object size to half the address space (which seems like a reasonable limitation to me). You can use one bit of `allocHeader.len` (e.g. the top bit) to store the mark state. Then, during sweeping, you can iterate through all objects and remove the ones that haven't been marked.

src/runtime/gc_list.go
```diff
@@ -0,0 +1,460 @@
...
+// mark a GC root at the address addr.
+func markRoot(addr, root uintptr) {
+	if !looksLikePointer(root) {
+		// This is not a pointer.
+		// Skip searching for it.
+		return
+	}
+
+	// Search the list of unmarked allocations for this root address.
+	alloc := activeAllocs.search(root)
```
aykevl, line 359: This also looks like a quadratic algorithm here: for every possible root, all allocations are scanned.

aykevl (commented):
I looked at the code (not super in-depth though) and it looks reasonable to me. Thank you for this effort! This is indeed much easier to review when it's separated from a precise GC (which will be quite a bit of work and may also be useful for `-gc=conservative`).

Can you add a test for this GC? You can add a test in main_test.go, see the `opt=1`, `opt=0`, and `ldflags` tests.

src/runtime/gc_list.go
niaow: Unfortunately yes. Not sure if this can/should be changed. This worked much better with the precise GC because there were far fewer roots to deal with. This is also fairly reasonable on AVR where I don't really see more than a dozen allocations very often.

src/runtime/gc_list.go
niaow: That seems reasonable.

aykevl: It's just an idea for how it can be improved. The code as-is already looks good to me.

src/runtime/gc_list.go
```diff
@@ -0,0 +1,460 @@
...
+	// Zero the allocation.
+	ptr := unsafe.Pointer(uintptr(unsafe.Pointer(alloc)) + unsafe.Sizeof(allocHeader{}))
+	memzero(ptr, size)
+
+	return ptr
+}
+
+// findMem searches the heap for a free span large enough to contain an allocation of the specified size.
+// If there are no sufficiently large free spans available, this returns nil.
+func findMem(size uintptr) *allocHeader {
```
aykevl, line 95: Two things I realize now:

- The memory should be aligned. You can use the `align` function for this. This is a no-op on AVR and should be optimized away entirely. On other platforms, it is necessary for correct functioning.
  ```go
  size = align(size)
  ```
- You might want to use a `zeroSizedAlloc` global (like in gc_conservative.go) to return for when the size is zero. This does sometimes happen in practice, although it is rare. The Go spec explicitly allows this:
  > Pointers to distinct zero-size variables may or may not be equal.
  
  On the other hand, it's rare and costs RAM/flash, so it would be reasonable to leave this out.

niaow: Adding the 0-sized alloc in because it only costs 1 byte of RAM (and saves 28 bytes of flash for some reason).

niaow: I am not entirely sure what you mean with the alignment. Assuming that `heapStart` is aligned (which I picked up from the conservative collector), `findMem` should always return appropriately aligned data. The header preserves pointer/word alignment, and every subsequent `start` is corrected with an align call.

src/runtime/gc_list.go
niaow: Actually, I think this may be better as-is. That tagging strategy would actually also make the scan quadratic (the list needs to be fully traversed each time we need to find what to scan). I think the current strategy with the scan stack is at least a bit more readable.

niaow:
@aykevl I am currently planning to do some GC cleanup to try to ensure that all of the conservative GCs behave similarly. Should we merge this first or should I push this to later?

deadprogram:
Either way, please note merge conflicts @niaow

niaow:
I am going to close this. This GC has too many issues and we have other priorities right now.

PR #2007, opened 2021-07-20
runtime: fix time base for time.Now()

### LINKED ISSUE #68: The "time" package
With docker image:

```
cat demo.go && tinygo run demo.go && date +%s
package main

import "time"

func main() {
        println(time.Now().Unix())
}
22598
1541263654
```
Off by orders of magnitude, nanoseconds output also looks suspicious.
```
cat demo2.go && tinygo run demo2.go 
package main

import "time"

func main() {
        println(time.Now().String())
}
  %101 = extractvalue { i8, i1 } %100, 0
 i32  call void @runtime.sliceBoundsCheck(i32 %109, i8 %101, i32 %108)
Both operands to a binary operator are not of the same type!
  %111 = sub i32 %108, i8 %101
Both operands to a binary operator are not of the same type!
  %112 = sub i32 %109, i8 %101
LLVM ERROR: Broken module found, compilation aborted!
```

aykevl:
This function previously returned the atomic time, that isn't affected
by system time changes but also has a time base at some arbitrary time
in the past. This makes sense for baremetal platforms (which typically
don't know the wall time) but it gives surprising results on Linux and
macOS: time.Now() usually returns a time somewhere near the start of
1970.

This commit fixes this by obtaining both time values: the monotonic time
and the wall clock time. This is also how the Go runtime implements the
time.now function.

---

This fixes #68.

Note that with this change, `runtime.AdjustTimeOffset` is only available on baremetal systems. That doesn't seem like a problem to me.

### DIFF
--- a/src/runtime/baremetal.go
+++ b/src/runtime/baremetal.go
@@ -52,3 +52,24 @@ func syscall_Exit(code int) {
 }
 
 const baremetal = true
+
+// timeOffset is how long the monotonic clock started after the Unix epoch. It
+// should be a positive integer under normal operation or zero when it has not
+// been set.
+var timeOffset int64
+
+//go:linkname now time.now
+func now() (sec int64, nsec int32, mono int64) {
+	mono = nanotime()
+	sec = (mono + timeOffset) / (1000 * 1000 * 1000)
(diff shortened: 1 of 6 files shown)


deadprogram:
:watch: thanks @aykevl now merging.

PR #2008, opened 2021-07-20
machine: add support to set size of RingBuffer

sago35:
Initially, the size of the RingBuffer used by the UART is 128.

https://github.com/tinygo-org/tinygo/blob/64d048c47c95bc0a70b4cff109afc1f9039705a4/src/machine/buffer.go#L7

This is not a problem in most cases, but it can sometimes be a problem when handling the wifi of the rtl8720dn.
For example, the size of a small HTTP header is about 200 bytes, and I want to process it all in a single transfer.

https://github.com/tinygo-org/drivers/pull/273

In this PR, the size can be set without changing the existing code.
In this case, I used variadic, but if there is a better way, I would like to adopt it.

### DIFF
--- a/src/machine/buffer.go
+++ b/src/machine/buffer.go
@@ -4,32 +4,38 @@ import (
 	"runtime/volatile"
 )
 
-const bufferSize = 128
+const defaultRingBufferSize = 128
 
 // RingBuffer is ring buffer implementation inspired by post at
 // https://www.embeddedrelated.com/showthread/comp.arch.embedded/77084-1.php
 type RingBuffer struct {
-	rxbuffer [bufferSize]volatile.Register8
-	head     volatile.Register8
-	tail     volatile.Register8
+	rxbuffer []volatile.Register8
+	head     volatile.Register16
+	tail     volatile.Register16
 }
 
 // NewRingBuffer returns a new ring buffer.
-func NewRingBuffer() *RingBuffer {
-	return &RingBuffer{}
+func NewRingBuffer(size ...int) *RingBuffer {
+	sz := defaultRingBufferSize
+	if len(size) > 0 {
+		sz = size[0]
+	}
+	return &RingBuffer{
+		rxbuffer: make([]volatile.Register8, sz),
+	}
 }
 
 // Used returns how many bytes in buffer have been used.
-func (rb *RingBuffer) Used() uint8 {
-	return uint8(rb.head.Get() - rb.tail.Get())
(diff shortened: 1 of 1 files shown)


src/machine/buffer.go
```diff
@@ -4,32 +4,38 @@ import (
...
+	head     volatile.Register16
+	tail     volatile.Register16
 }
 
 // NewRingBuffer returns a new ring buffer.
-func NewRingBuffer() *RingBuffer {
-	return &RingBuffer{}
+func NewRingBuffer(size ...int) *RingBuffer {
+	sz := defaultRingBufferSize
+	if len(size) > 0 {
```
ardnew, line 20: I believe this `RingBuffer` implementation depends on its size being a power of 2 to support filling _all_ slots in the buffer (although you might not notice an issue until that many slots are filled). 

Might either verify `sz` satisfies that condition, or round up to ensure it does.

aykevl:
I do not think this is a good idea. I have intentionally picked an array instead of a slice to avoid extra indirection and thus overhead. If this is an issue, it may be better to add an API to call a callback on an incoming byte instead (somewhat similar to pin interrupts).

>  For example, the size of a small HTTP header is about 200 bytes, and I want to process it all in a single transfer.

Is this just a performance problem or does the code behave incorrectly for larger buffers?

src/machine/buffer.go
sago35: I agree.
Surely some sort of check is needed.

sago35:
> it may be better to add an API to call a callback on an incoming byte instead (somewhat similar to pin interrupts).

I'll try it.

> Is this just a performance problem or does the code behave incorrectly for larger buffers?

I use a larger buffer for communication to match the Arduino implementation.
Using a larger buffer makes the implementation simpler.

At first it was important to have a simple implementation, but now it looks like I can refactor it.
I will try to refactor it a bit so that I don't have to resize the RingBuffer.

sago35:
With the following commit, this PR is no longer needed.
Close this PR.

https://github.com/tinygo-org/drivers/pull/287#issuecomment-888284428

ardnew:
> I will try to refactor it a bit so that I don't have to resize the RingBuffer.

You may have already resolved this, but I wanted to suggest simply duplicating the logic of this `RingBuffer` using a larger array (or slice) private to one of the types you are working with.

It is a very small implementation after all, so it might be easier to copy its functionality than refactoring your code to work with this specific type.

Issue #1101, opened 2020-05-12
Asyncify for goroutines in WebAssembly?


kripken:
Context: @aykevl [wrote](https://github.com/WebAssembly/design/issues/1340#issuecomment-626258448):

> Stackful coroutines (called goroutines). We've had to go through great pains to get these working in TinyGo and we're still hitting corner cases where it doesn't work. If it works, it's inefficient.

The discussion there on new wasm capabilities is the right long-term solution, of course, but I wonder if maybe until then, [Asyncify](https://kripken.github.io/blog/wasm/2019/07/16/asyncify.html) might be useful here?

With Asyncify you run [a utility](https://github.com/WebAssembly/binaryen#tools) on a wasm file, and get a modified wasm file that can pause and resume execution at various points. It's easy to use - you don't need to think about how it works or handle corner cases, just tell it which calls can pause, and set up a little runtime code to call things. This does add overhead, about 50% in size and speed on average in the worst cases (where it ends up instrumenting almost everything), but in the best cases it's very efficient - for example, it can avoid instrumenting inner loops where possible, letting such code run at full speed.

The idea here would be to leave many of the low-level goroutine pause/resume details to Asyncify. All TinyGo would do is have some special function call (which you don't need to implement; it's a marker for Asyncify) that indicates a pause/resume point. Asyncify is run on the wasm. Then when the wasm runs, a goroutine that reaches such a point can return to the runtime code, which can then decide which other goroutine to proceed to run.

If there's interest I'd be happy to collaborate on investigating this! (I wrote Asyncify, and have helped integrate it in various projects, but I don't know that much about TinyGo.)

jaddr2line:
This is unfortunately used on platforms other than WASM. I am currently working on a rebuild of the system which should work better than the current version (and probably better than asyncify).

kripken:
I see, thanks @jaddr2line ! Sounds irrelevant then, closing...

niaow:
It probably makes more sense to switch to asyncify now. Asyncify is more mature, most other platforms have switched to stack-based goroutines. Additionally the previously mentioned rebuild of the coroutines system didnt happen.

aykevl:
I have investigated this, and in fact looked into the feasibility of reimplementing an Asyncify-like pass in Go (to avoid a large JavaScript dependency) but there is one very big downside: it destroys debug information. This is important, see #1994 for why. Unless Asyncify learns to retain debug information, I'm afraid it is a blocker.

[...]

Issue #2009, opened 2021-07-20
[WASM] __heap_base not always aligned


fgsch:
While investigating a crash using #2003 and with @aykevl help, we found the heap start (__heap_base) is not always aligned.
For example building with:

```
tinygo build -interp-pass=none -target wasi -o main.wasm code.go
```

And setting https://github.com/tinygo-org/tinygo/blob/dev/src/runtime/gc_conservative.go#L240 to true, I get:

```
heapStart:         0x00013551
heapEnd:           0x00040000
total size:        0x0002caaf
metadata size:     0x00000b2a
metadataStart:     0x0003f4d6
# of blocks:       0x00002bf8
# of block states: 0x00002ca8
```

aykevl:
This may be a LLVM bug, this might be a way to fix it:

```diff
diff --git a/lld/wasm/Writer.cpp b/lld/wasm/Writer.cpp
index 63e3dac5d298..5ceb4d76e981 100644
--- a/lld/wasm/Writer.cpp
+++ b/lld/wasm/Writer.cpp
@@ -312,6 +312,7 @@ void Writer::layoutMemory() {
     // Set `__heap_base` to directly follow the end of the stack or global data.
     // The fact that this comes last means that a malloc/brk implementation
     // can grow the heap at runtime.
+    memoryPtr = alignTo(memoryPtr, 4);
     log("mem: heap base   = " + Twine(memoryPtr));
     WasmSym::heapBase->setVA(memoryPtr);
   }
```
I haven't tested this patch yet (right now compiling LLVM).

fgsch:
I can confirm the patch fixes this issue.

niaow:
According to #2010, it appears that we may actually need 8 byte alignment.

aykevl:
@fgsch thanks for verifying!

I've made a patch here: https://reviews.llvm.org/D106499, which aligns to 8 bytes.

aykevl:
There are two workarounds from the TinyGo side:

* remove the `--stack-first` flag
* manually align `heapStart`

The second workaround looks like this:

```diff
diff --git a/src/runtime/gc_conservative.go b/src/runtime/gc_conservative.go
index e9682592..b6d18455 100644
--- a/src/runtime/gc_conservative.go
+++ b/src/runtime/gc_conservative.go
@@ -228,6 +228,7 @@ func setHeapEnd(newHeapEnd uintptr) {
 // This function can be called again when the heap size increases. The caller is
[...]

Issue #1972, opened 2021-06-28
Extreme high memory usage when building wasm


zluo01:
When try to build my code into a wasm with tinygo which command `tinygo build -o main.wasm -target wasm ./main.go` . It yields memory usage over 100GB and long build time (eventually kill due to insufficient memory). [source](https://github.com/zluo01/color-thief/blob/master/wasm/main.go) code here.

<img width="973" alt="Screen Shot 2021-06-27 at 20 37 39" src="https://user-images.githubusercontent.com/12102011/123573691-05f4e800-d794-11eb-80a4-ddc03685818f.png">

zluo01:
try with the latest 0.19 release, yet the problem still exists.  I find out this [file](https://github.com/zluo01/color-thief/blob/master/wsm/wsm.go) cause the problem. Yet the only external thing I use is `math.Sqrt`. I wonder what can be the reason that cause tinygo to fail the build ? Thank you.

inkeliz:
I have the same issue compiling the Gio app (https://todo.sr.ht/~eliasnaur/gio/151). It eats all the memory and then leave an "LLVM ERROR: out of memory". :\

gedw99:
@Inkeliz Does this help ? https://github.com/tinygo-org/tinygo/issues/2108#issuecomment-918700845

It reservees the memory and i think will unblock you.. Let me know

inkeliz:
I didn't try because I can't compile, and it doesn't seem to be related.

The #2108 is about to limit the RAM on the runtime. It says, "I build this file by the command `tinygo build -o a.wasm -target=wasi a.go`", so there's no issue with the compilation. The question is how to limit the memory during the runtime. 

However, the issue that @zluo01 and I are facing is on the compilation, not the runtime. We can't compile at all. The compiler (tinygo) uses all the memory available and then crashes with out-of-memory, even with 256GB of RAM. 

When we run `tinygo build`, it starts to use all the memory, and after hours it crashes.

gedw99:
Does it work with the gio kitchen example ?

inkeliz:
I suspect that this bug is from  `wasm-opt` rather than TinyGo. 

If you still have the code, can you compile with GoCompiler (`GOARCH=wasm GOOS=js go build ./main.go` and then use `wasm-opt -O3 main.wasm -o main.wasm` or `wasm-opt -Oz main.wasm -o main.wasm`?

[...]

PR #2011, opened 2021-07-21
internal/task: use asyncify on WASM

niaow:
This currently requires manually running the compiled wasm through `wasm-opt`.
Very little works so far.
Issue: #1101


dgryski:
What work is remaining to get this PR finished?  I'd love to help out any way I can.

niaow:
This is mainly blocked by #2035 I think

dgryski:
While that bug is getting fixed, is there work that could be done on this issue even with having to run wasm-opt manually?

aykevl:
I'm almost inclined to say that we should perhaps just get this finished (even without #2035). That means compiling for WebAssembly requires an extra binary - but I think that's only required when using the new Asincify based scheduler?

Thoughts?
I don't like adding an external dependency (it makes distribution a whole lot harder) but right now concurrency on WebAssembly is already somewhat broken and this PR might fix it.

niaow:
Alright, I can modify this to just shell out to wasm-opt for now when using asyncify.

niaow:
I made that change and attempted to debug some issues. It appears that after the initial launch suspend, resumed goroutines will not suspend again. I don't quite understand it yet.

dgryski:
I'll take a look at this on Monday and see what I can come up with.

niaow:
I have fixed what was probably the biggest bug. However, there are still some things left to be fixed apparently, such as weird WASI crashes and some sort of unicode error in the reflect test (. . . and i still havent added binaryen to CI).

dgryski:
\o/. This works great on my little test `ping pong` program (included below)

```
package main

import (
	"time"
)

func pinger(pinger <-chan int, ponger chan<- int, who string) {
	for {
		<-pinger
		println(who)
		time.Sleep(time.Second)
		ponger <- 1
	}
}

func main() {
	ping := make(chan int)
	pong := make(chan int)

	go pinger(ping, pong, "ping")
	go pinger(pong, ping, "pong")

	ping <- 1

	select {}
}
```

However, I'm getting an unaligned pointer error with `gc=leaking` on `wasmtime` when trying to read the clock.

I managed to fix it with this patch.

```
 func ticks() timeUnit {
-       var nano uint64
-       clock_time_get(0, timePrecisionNanoseconds, &nano)
+       var b [16]byte
+       ptr := uintptr(unsafe.Pointer(&b[0]))
+       ptr += 8 + (^ptr & 0x7) + 1
+
+       clock_time_get(0, timePrecisionNanoseconds, (*uint64)(unsafe.Pointer(ptr)))
+       nano := *(*uint64)(unsafe.Pointer(ptr))
        return timeUnit(nano)
 }

```

It seems like a proper fix lies elsewhere, although I'm not sure where..

Trying to run the above ping-pong code with `gc=conservative` instead produces:

```
Caused by:
    1: wasm trap: uninitialized element
       wasm backtrace:
           0: 0x6c376 - <unknown>!<wasm function 266>
           1: 0x1489c - <unknown>!<wasm function 68>
           2: 0x2e336 - <unknown>!<wasm function 158>
           3: 0x2d77d - <unknown>!<wasm function 156>
           4: 0x2d636 - <unknown>!<wasm function 155>
```

I'm still looking into this..

niaow:
> However, I'm getting an unaligned pointer error with `gc=leaking` on `wasmtime` when trying to read the clock.

I suspect this may be fixed by #2009. This PR is kinda old so I can rebase tommorow.

dgryski:
Adding `-g` to the options provided to `wasm-opt` with the following patch:

```
diff --git a/builder/build.go b/builder/build.go
index df329bd7..7568e8ae 100644
--- a/builder/build.go
+++ b/builder/build.go
@@ -641,7 +641,7 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
                                default:
                                        return fmt.Errorf("unknown opt level: %q", config.Options.Opt)
                                }
-                               cmd := exec.Command("wasm-opt", "--asyncify",
+                               cmd := exec.Command("wasm-opt", "-g", "--asyncify",
                                        "--optimize-level", strconv.Itoa(optLevel),
                                        "--shrink-level", strconv.Itoa(shrinkLevel),
                                        executable, "--output", executable)
```

gives me a much better error:
```
Caused by:
    1: In func wasi_snapshot_preview1::args_sizes_get at write size: Pointer not aligned to 4: Region { start: 1066183, len: 4 }
       wasm backtrace:
           0: 0x320f - <unknown>!runtime.initAll
           1: 0x307f - runtime.run$1
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/scheduler_any.go:19:10
           2: 0x301d - <goroutine wrapper>
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/scheduler_any.go:18:2
           3: 0x3cb6 - <unknown>!tinygo_launch
           4:  0x538 - (*internal/task.Task).Resume
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/internal/task/task_asyncify.go:94:17
           5: 0x2edb - runtime.scheduler
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/scheduler.go:171:11
                     - runtime.run
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/scheduler_any.go:24:11
                     - _start
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/runtime_wasm_wasi.go:21:5

```

So indeed, something about asyncify is futzing with the alignment of the pointers and making things unhappy.

dgryski:
I rebased my local copy against `dev` with no success.  Same pointer alignment issue.  I'll keep looking at this..

dgryski:
Patching gc_leaking.go with
```
 func initHeap() {
-       // preinit() may have moved heapStart; reset heapptr
-       heapptr = heapStart
+       // preinit() may have moved heapStart; reset heapptr and make sure it's aligned
+       heapptr = align(heapStart)
 }
```

allows both gc=conservative and gc=leaking to run the pingpong example.

niaow:
For some reason I thought chocolatey had binaryen. I guess I will need to switch over the install then.

dgryski:
Something about the asyncify code is causing https://github.com/tinygo-org/tinygo/issues/2172 to fail

azure-pipelines.yml
```diff
@@ -15,6 +15,19 @@ jobs:
...
       fetchDepth: 1
+    - task: Bash@3
+      displayName: Install Dependencies
+      inputs:
+        targetType: inline
+        script: |
+          # install ninja
+          choco install ninja
+          # hack ninja to use fewer jobs
+          echo -e 'C:\\ProgramData\\Chocolatey\\bin\\ninja -j4 %*' > /usr/bin/ninja.bat
```
aykevl: There is no reason to do this when not building LLVM.

compiler/compiler_test.go
```diff
@@ -93,6 +101,7 @@ func TestCompiler(t *testing.T) {
 				Scheduler:          config.Scheduler(),
 				FuncImplementation: config.FuncImplementation(),
 				AutomaticStackSize: config.AutomaticStackSize(),
+				DefaultStackSize:   4096,
```
aykevl: I think this should match the target JSON for consistency.

```suggestion
				DefaultStackSize:   config.DefaultStackSize(),
```

src/internal/task/task_asyncify.go
```diff
@@ -0,0 +1,129 @@
...
+func align(p uintptr) uintptr
+
+// initialize the state and prepare to call the specified function with the specified argument bundle.
+func (s *state) initialize(fn uintptr, args unsafe.Pointer, stackSize uintptr) {
+	// Save the entry call.
+	s.entry = fn
+	s.args = args
+
+	// Create a stack.
+	stackSize = align(stackSize)
```
aykevl: Normally, I would expect an aligned stack to be passed in. For example, a stack size of 2048 is very much aligned.
I never really thought about aligning it, but yeah, technically it might need to be aligned.

I don't think this needs to be done at runtime (although there is a good chance it's optimized away in this case): it's only ever specified in code (in compileopts/target.go for example) or in target JSON files. The only way to easily change it is by modifying the JSON file directly, for example. If we're going to align it, I think it would be better to do it at compile time instead of at runtime, for example here:

https://github.com/tinygo-org/tinygo/blob/157382600535c95bdbf5d6e8c4b9e5ca71749018/compiler/goroutine.go#L125

This can be done by aligning to 16 bytes, I believe that's nowadays a pretty standard alignment (even on riscv32 and i386, for example).

src/internal/task/task_asyncify.go
```diff
@@ -0,0 +1,129 @@
...
+
+// Pause suspends the current task and returns to the scheduler.
+// This function may only be called when running on a goroutine stack, not when running on the system stack.
+func Pause() {
+	// This is mildly unsafe but this is also the only place we can do this.
+	if *(*uintptr)(unsafe.Pointer(currentTask.state.asyncifysp)) != stackCanary {
+		runtimePanic("stack overflow")
+	}
+
+	currentTask.state.unwind()
```
aykevl, line 94: So I guess once this call returns, we're back after rewinding?

targets/wasi.json
```diff
@@ -5,6 +5,9 @@
 	"goarch":        "arm",
 	"linker":        "wasm-ld",
 	"libc":          "wasi-libc",
+	"scheduler":     "asyncify",
+	"gc":            "conservative",
```
aykevl: This GC was already the default, so this is not technically needed. But many other targets also do it so :shrug:

testdata/channel.go
```diff
@@ -244,7 +244,7 @@ func receive(ch <-chan int) {
 func sender(ch chan int) {
 	for i := 1; i <= 8; i++ {
 		if i == 4 {
-			time.Sleep(time.Microsecond)
+			time.Sleep(time.Second)
```
aykevl: Why did you increase this to a second? That's some valuable time spent in a test. If you want to make it longer than a microsecond, please consider making it multiple microseconds or even a whole millisecond.

aykevl (commented):
This looks pretty clean. I would like to take a more closer look at some parts, but so far I only have some relatively minor comments.

testdata/channel.go
niaow: This was a temporary change I made when debugging a timer queue crash. I will revert this when cleaning up.

src/internal/task/task_asyncify.go
niaow: Yes.

aykevl:
> For some reason I thought chocolatey had binaryen. I guess I will need to switch over the install then.

`scoop` has it, not `choco`.

niaow:
I have created a fix for the WASM scheduler recursion failure: #2178

dgryski:
I think this branch still has an odd failure.  Trying to build the following program with
`tinygo build -scheduler=asyncify -gc=leaking -target=wasi -o m.wasm main.go`


```
package main

import (
	"context"
)

func main() {
	ctx := context.Background()

	select {
	case <-ctx.Done():
	}
}
```

produces the following build failure:
```
%runtime.channel.56* null
 %runtime.channel.80*  %1 = call i1 bitcast (i1 (%runtime.channel*, i8*, %runtime.channelBlockedList*, i8*, i8*)* @runtime.chanRecv to i1 (%runtime.channel.80*, i8*, %runtime.channelBlockedList.79*, i8*, i8*)*)(%runtime.channel.56* null, i8* null, %runtime.channelBlockedList.79* %chan.blockedList, i8* undef, i8* null), !dbg !1627
```

niaow:
I have been poking around, and it currently appears that the interface lowering pass is the first pass introducing type errors when building that code.

niaow:
I identified the issue and seperated it into #2197

niaow:
Hmm, it looks like LLVM 10 may have issues with the not-yet-resolved symbols for asyncify. I am not really sure how we want to handle this. I guess we could disable this on LLVM 10? I am not sure of a better option.

dgryski:
Given that llvm 13 is out now, how many versions behind is tinygo committed to supporting?

aykevl:
> Hmm, it looks like LLVM 10 may have issues with the not-yet-resolved symbols for asyncify. I am not really sure how we want to handle this. I guess we could disable this on LLVM 10? I am not sure of a better option.

Don't let that be a blocker. If LLVM 10 is an issue, we'll drop LLVM 10 support. That's a really old version by now. (Usually we support the last two LLVM versions or so, to make things easier for distributions). Even more so because this PR is a blocker to add support for newer LLVM versions.

aykevl:
I've created #2216 to drop support for LLVM 10 so that this PR can continue without being burdened by that LLVM version.

niaow:
Alright, sounds good. I will take a look at it tomorrow.
The main other issues I still need to deal with right now are:
1. Fix Windows CI
2. Figure out why `test-llvm11-go116` is timing out

azure-pipelines.yml
```diff
@@ -15,6 +15,16 @@ jobs:
...
       fetchDepth: 1
+    - task: Bash@3
+      displayName: Install Dependencies
+      inputs:
+        targetType: inline
+        script: |
+          # install qemu
+          choco install qemu --version=2020.06.12
+          # install binaryen
+          wget https://github.com/WebAssembly/binaryen/releases/download/version_101/binaryen-version_101-x86_64-windows.tar.gz
```
aykevl: Apparently wget isn't included on this Windows version: https://github.com/actions/virtual-environments/blob/main/images/win/Windows2016-Readme.md
You may have better luck with curl, or by installing emscripten using Scoop instead: ScoopInstaller/Main#2740 (installing Scoop should be pretty fast).

deadprogram:
Now that #2216 has been merged into `dev` branch, this PR should be able to continue. :smile_cat:

niaow:
So it appears that we may need to require binaryen version 102 instead of 101. This makes it a bit more annoying to install, since 102 does not have pre-built binaries.

aykevl:
> So it appears that we may need to require binaryen version 102 instead of 101.

Are you referring to this failure?

```
build/tinygo build             -o test.wasm -tags=hifive1b             examples/blinky1
error: wasm-opt failed: signal: segmentation fault (core dumped)
```

niaow:
> Are you referring to this failure?

Yeah, I think that is fixed in binaryen 102. I am not 100% sure, although it seems to match up with a bug report in 101, and I have not been able to replicate any similar fault on 102.

deadprogram:
Any action on this PR? Quite a few people are eagerly awaiting it (see what I did there?).

niaow:
> Any action on this PR? Quite a few people are eagerly awaiting it (see what I did there?).

Sorry I haven't gotten a chance to work on it this week.
Tomorrow I should hopefully be able to get back to it.
I am going to try to get all of the targets to build binaryen 102 from source and bundle it in the release tarball.

niaow:
Getting this to work on stretch is seemingly more difficult than I anticipated, due to the requirement for a newer cmake version.

aykevl:
Did you verify that the bug is indeed there in version 101 and fixed in 102? It could be something unrelated to the version.

aykevl:
With version 101:

```
$ $WASMOPT --version
wasm-opt version 101 (version_101)

$ tinygo build -o test.wasm -tags=hifive1b -size=short examples/blinky1
  29870     771  130301 |   30641  131072
```

I'm compiling version 102 right now.

niaow:
> Did you verify that the bug is indeed there in version 101 and fixed in 102? It could be something unrelated to the version.

It is a non-deterministic bug, and does not replicate on my system. It doesn't seem to replicate on 102 so far, so it may be right it may be wrong but cant tell for sure.

aykevl:
Hmm, perhaps with was a flaky test then?
If it can't be reproduced, I'd say don't worry about bundling wasm-opt in all TinyGo releases.

niaow:
Not a flaky test. This is a legitimate bug, and reproduces randomly in CI (generally at least one of the builds fails, but rarely more than that) with 101.

niaow:
I think(?) I have fixed all of the build errors at this point. I just need to wait for all of the CI run (and Windows may take a while).

niaow:
I had removed the `rm -rf llvm-source` without realizing why it was done. I updated the comments in the CI config to try to explain it better, because it was surprising behavior.

deadprogram:
All tests are passing now!

niaow:
@deadprogram windows smoke tests are still running: https://github.com/niaow/tinygo/runs/4132134414?check_suite_focus=true

niaow:
Windows smoke tests failed.

niaow:
It looks like I messed up the autodetection on Windows (I need to add a `.exe` suffix).

niaow:
The commit I just pushed should fix it, although it will take 2 hours to find out. I think this is ready for review to start now.

.circleci/config.yml
```diff
@@ -25,7 +25,9 @@ commands:
                 qemu-system-arm \
                 qemu-user \
                 gcc-avr \
-                avr-libc
+                avr-libc \
+                cmake \
+                ninja-build
```
aykevl, line 30: Can you do this installation only when these tools are needed in the build-binaryen step? That speeds up the CI in the common case.

.circleci/config.yml
```diff
@@ -141,9 +175,14 @@ commands:
                 qemu-system-arm \
                 qemu-user \
                 gcc-avr \
-                avr-libc
+                avr-libc \
+                ninja-build \
+                python3
```
aykevl, line 190: Same here.

.circleci/config.yml
```diff
@@ -206,9 +240,14 @@ commands:
...
```
aykevl, line 255: And here.

.circleci/config.yml
```diff
@@ -283,7 +317,7 @@ commands:
             curl https://dl.google.com/go/go1.17.darwin-amd64.tar.gz -o go1.17.darwin-amd64.tar.gz
             sudo tar -C /usr/local -xzf go1.17.darwin-amd64.tar.gz
             ln -s /usr/local/go/bin/go /usr/local/bin/go
-            HOMEBREW_NO_AUTO_UPDATE=1 brew install qemu
+            HOMEBREW_NO_AUTO_UPDATE=1 brew install qemu cmake ninja
```
aykevl, line 330: And here.

.circleci/config.yml
```diff
@@ -76,6 +85,29 @@ commands:
...
             - llvm-project/lld/include
             - llvm-project/llvm/include
+  hack-ninja-jobs:
+    steps:
+      - run:
+          name: "Hack Ninja to use less jobs"
+          command: |
+            echo -e '#!/bin/sh\n/usr/bin/ninja -j3 "$@"' > /go/bin/ninja
+            chmod +x /go/bin/ninja
+  build-binaryen:
```
aykevl: Suggestion: this might be clearer when the name is `build-binaryen-linux`. Because the cache is only for Linux.
Similarly, I think it's better to rename the cache key to something like `binaryen-linux-v1`.

.gitignore
```diff
@@ -1,4 +1,3 @@
-build
```
aykevl, line 1: Oh, I see you've moved the .gitignore line into a separate file in the build directory.
I guess that's fine, just wondering what the reason is?

Makefile
```diff
@@ -169,6 +169,12 @@ $(LLVM_BUILDDIR)/build.ninja: llvm-source
 $(LLVM_BUILDDIR): $(LLVM_BUILDDIR)/build.ninja
 	cd $(LLVM_BUILDDIR); ninja $(NINJA_BUILD_TARGETS)
 
+# Build Binaryen
+.PHONY: binaryen
+binaryen: build/wasm-opt
+build/wasm-opt:
+	cd lib/binaryen; cmake -G Ninja . -DBUILD_STATIC_LIB=ON; ninja
```
aykevl:
```suggestion
	cd lib/binaryen && cmake -G Ninja . -DBUILD_STATIC_LIB=ON && ninja
```

Not sure whether this works as-is in a Makefile but generally not continuing on errors is a good thing.

aykevl (commented):
I've left a few comments about how the CI can be sped up a bit, but in general this looks very good.

I have one question though: what happens if you export a function and do a blocking call in there? JavaScript can't block, so it must return _something_ instead. This could be surprising to users, especially if this behavior is not part of the function signature.

aykevl:
> The commit I just pushed should fix it, although it will take 2 hours to find out. I think this is ready for review to start now.

Yeah, caching on GitHub Actions is a bit weird. It should work, but for some reason there are a lot of cache misses.

.gitignore
niaow: I needed to keep the `build` directory, and for some reason was struggling to get it to track as-is (negated patterns seemed to be behaving weirdly). This is just how I normally do that.

.circleci/config.yml
niaow: I did this like this for 2 reasons:
1. We need to install these differently on buster and stretch.
2. We need these for both LLVM and Binaryen, and avoiding a double install seemed annoying.

I guess I could potentially restore both caches before installing, although it seems to make this build config a lot more complicated.

niaow: This appears to only make CI 1 second slower. I don't think the extra complexity is worthwhile.

niaow:
> I have one question though: what happens if you export a function and do a blocking call in there? JavaScript can't block, so it must return something instead. This could be surprising to users, especially if this behavior is not part of the function signature.

Assuming that you use `js.FuncOf`. . .  the current behavior (unchanged by this PR) is that stuff will break. Go defines this to deadlock, but our implementation is still non-conformant (I hope to fix this in a future PR).

As for raw exported functions, they will panic in `task.Pause` if there are no running goroutines. If there are running goroutines (Go calls into JS calls into Go which then blocks), this will break and bad things will happen. This particular issue is a lot more difficult to solve, and will require us to provide our own implementation of `syscall/js` with some changes.

.circleci/config.yml
aykevl: > This appears to only make CI 1 second slower. I don't think the extra complexity is worthwhile.

How did you measure this? It seems rather unlikely that installing software is that fast.
(If it is really just 1 second, then installing it unconditionally would be fine).

niaow: I looked at 2 a build on the `dev` branch and a build from this PR. The "install apt dependencies steps" took 22 seconds on this PR and 23 seconds on the dev branch. Looking through more CI runs on `dev` now I see a lot more variation, but the difference in download size is 180 MB to 186 MB which gives me a seemingly appropriate estimate of 3%. These packages are relatively small compared to the other packages which we install.

aykevl:
> As for raw exported functions, they will panic in `task.Pause` if there are no running goroutines. If there are running goroutines (Go calls into JS calls into Go which then blocks), this will break and bad things will happen.

Right. That was what I was wondering about. I suspect many people today call exported functions from TinyGo, as described [here](https://tinygo.org/docs/guides/webassembly/). Maybe `-scheduler=none` is the solution in that case, but maybe that's a bit too drastic. In any case, we have to provide some answer to the question of how to call into Go functions from JavaScript.
(This is why I prefer to keep `-scheduler=coroutines` around for one release if possible).

niaow:
Something seems really wrong with the Windows build, so that might take a bit to debug. https://github.com/niaow/tinygo/runs/4132565997?check_suite_focus=true#step:17:72

(This means that binaryen crashed when just run with `--version`)

.circleci/config.yml
aykevl: Ok. Well maybe it is fine then. We can revisit this later, especially when we move away from Debian stretch.

aykevl:
@niaow you might want to rebase on top of the dev branch, #2245 should avoid rebuilding LLVM on each push.

niaow:
It looks like the version of GCC used by Debian Stretch is too old to build `binaryen` correctly, and that the cache was hiding this fact.

niaow:
I managed to work around that by using the copy of clang that we build.
Now the only remaining issue is Windows.

niaow:
If my debugging attempts are correct, it looks like binaryen is still trying to dynamically link against `libstdc++` and `libgcc` from mingw, and failing (error code `0xc0000139`).
```
[niaow@finch bin]$ objdump -p wasm-opt.exe | grep "DLL Name:"
	DLL Name: libgcc_s_seh-1.dll
	DLL Name: KERNEL32.dll
	DLL Name: msvcrt.dll
	DLL Name: libwinpthread-1.dll
	DLL Name: USER32.dll
	DLL Name: libstdc++-6.dll
```

aykevl:
For the TinyGo build I use `-static -static-libgcc -static-libstdc++`, I think you might need those for binaryen too.

niaow:
Yeah I just realized that. Adding now.

aykevl:
I can try rebooting to Windows to test some build options, would that help?

niaow:
I suspect that I just fixed it. If not, I finally have a Windows VM setup so I can test there.

niaow:
All tests are now passing.

niaow:
I did a quick squash/cleanup, so I think this is ready for a final review.

deadprogram:
Do we need to modify the dev Dockerfile if you want to use this scheduler? https://github.com/tinygo-org/tinygo/blob/dev/Dockerfile#L23-L28

Dockerfile
```diff
@@ -30,7 +30,7 @@ COPY --from=tinygo-base /tinygo/targets /tinygo/targets
 RUN cd /tinygo/ && \
     apt-get update && \
     apt-get install -y make clang-11 libllvm11 lld-11 && \
-    make wasi-libc
+    make wasi-libc binaryen
```
deadprogram, line 34: Oops, just noticed this. Never mind my other question.

Makefile
```diff
@@ -54,6 +54,8 @@ ifeq ($(OS),Windows_NT)
     CGO_LDFLAGS += -static -static-libgcc -static-libstdc++
     CGO_LDFLAGS_EXTRA += -lversion
 
+	BINARYEN_OPTION += -DCMAKE_EXE_LINKER_FLAGS='-static-libgcc -static-libstdc++'
```
deadprogram: I'm almost embarrassed to point out something so trivial as an extra tab here :joy_cat:

niaow: Huh. I guess this is what happens when using 4-wide tabs in editor :facepalm:

Dockerfile
niaow: I just double checked and this doesn't actually work anymore, so I do actually need to fix it.

deadprogram (approved):
All tests are now passing, and seems good to me.

deadprogram:
@aykevl any last comments before merge?

gedw99:
Is this the fix to get WASM compilation working again ?

deadprogram:
> Is this the fix to get WASM compilation working again ?

WASM compilation is working afaik. This PR is to add asyncify as an option for task scheduling for WASM programs.

gedw99:
> > Is this the fix to get WASM compilation working again ?
> 
> WASM compilation is working afaik. This PR is to add asyncify as an option for task scheduling for WASM programs.

ok thank for the response. 

https://github.com/tinygo-org/tinygo/issues/1972 is blocking i believe. I think is for gio when building to Web ( wasm ).

aykevl (commented):
This looks good to me.
I can't really comment on the Dockerfile changes, is the issue there fixed now? I don't know a whole lot about Docker.

goenv/goenv.go
```diff
@@ -67,11 +70,98 @@ func Get(name string) string {
...
+		}
+
+		return findWasmOpt()
 	default:
 		return ""
 	}
 }
 
+// Find wasm-opt, or exit with an error.
+func findWasmOpt() string {
```
aykevl, line 91: [this comment was intended to be part of the review but I forgot to click the "start review" button]

I think this function can be simplified. In short: the bundled wasm-opt binaries don't need to be checked for a version.

I would suggest the following logic:

```
if $TINYGOROOT/bin/wasm-opt[.exe] exists:
    return this path
if $TINYGOROOT/build/wasm-opt[.exe] exists:
    return this path
look for wasm-opt using LookPath
if not found:
    print or return error
check for version
if version too old:
    return error
return this path
```

This is not a blocker to me.

niaow: I think we should still be doing a version check regardless for now, until we can actually link it in properly.

niaow:
The Dockerfile issues have been fixed.

goenv/goenv.go
niaow: This has already caught one issue, so it is still helpful.

deadprogram:
OK merge time! Great work on this @niaow thank you. Also thanks to everyone who helped review/test.

deadprogram (approved):

Issue #2010, opened 2021-07-21
runtime (wasi): 64-bit integers require 8 byte alignment


niaow:
While working on asyncify support, I ran into a crash where `wasmtime` failed to store the timestamp. Interestingly, my debug code also caused it to reproduce on the `dev` branch.

This bug can be reproduced by attempting to print the address of the timestamp before passing it to `clock_time_get`:
```
func ticks() timeUnit {
	var nano uint64
	println(&nano)
	clock_time_get(0, timePrecisionNanoseconds, &nano)
	return timeUnit(nano)
}
```

The address (`0x0001064c`) appears to be 4-byte aligned. I don't think that is incorrect on 32-bit wasm?
However it breaks wasmtime:
```
[niaow@finch tinygo]$ build/tinygo build -target wasi -gc=leaking -opt=1 -o goroutines.wasm ./testdata/goroutines.go
[niaow@finch tinygo]$ WASMTIME_BACKTRACE_DETAILS=1 wasmtime goroutines.wasm 
main 1
sub 1
0x0001064c
Error: failed to run main module `goroutines.wasm`

Caused by:
    0: failed to invoke command default
    1: In func clock_time_get:write timestamp:
       wasm backtrace:
           0: 0x1cc2 - runtime.ticks
                           at /home/niaow/go/src/github.com/tinygo-org/tinygo/src/runtime/runtime_wasm_wasi.go:89:16
           1: 0x1f50 - runtime.addSleepTask
                           at /home/niaow/go/src/github.com/tinygo-org/tinygo/src/runtime/scheduler.go:92:14
           2: 0x2025 - time.Sleep
                           at /home/niaow/go/src/github.com/tinygo-org/tinygo/src/runtime/scheduler_any.go:10:14
           3: 0x24e6 - main.sub
                           at /home/niaow/go/src/github.com/tinygo-org/tinygo/testdata/goroutines.go:91:12
           4: 0x1ed9 - main.main
                           at /home/niaow/go/src/github.com/tinygo-org/tinygo/testdata/goroutines.go:11:2
           5: 0x1d47 - runtime.run$1
                           at /home/niaow/go/src/github.com/tinygo-org/tinygo/src/runtime/scheduler_any.go:21:11
           6: 0x1cf8 - runtime.run
                           at /home/niaow/go/src/github.com/tinygo-org/tinygo/src/runtime/scheduler_any.go:18:2
           7: 0x1ce5 - _start
[...]

PR #2014, opened 2021-07-25
wasm: align heap to 16 bytes

aykevl:
This PR fixes two things:

  * It changes the alignment to 16 bytes (from 4), to match `max_align_t` in C.
  * It manually aligns heapStart on WebAssembly, to work around a bug in `wasm-ld` with `--stack-first` (see https://reviews.llvm.org/D106499).

### DIFF
--- a/src/runtime/arch_tinygowasm.go
+++ b/src/runtime/arch_tinygowasm.go
@@ -32,9 +32,11 @@ var (
 
 const wasmPageSize = 64 * 1024
 
-// Align on word boundary.
 func align(ptr uintptr) uintptr {
-	return (ptr + 3) &^ 3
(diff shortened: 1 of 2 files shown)


fgsch:
Perhaps this can also add a test from https://github.com/tinygo-org/tinygo/issues/2010 ?

aykevl:
Hmm, perhaps yes. I'm not sure though as checking for alignment does not guarantee anything: the pointer can also be aligned by chance.

fgsch:
Either way, this looks good to me 👍

niaow (approved):

deadprogram:
Thank you, everyone. Now merging.

Issue #1993, opened 2021-07-10
machine/rp2040: PWM implementation


soypat:
I'm on it! See https://github.com/soypat/tinygo/blob/patch-pico-pwm/src/machine/machine_rp2040_pwm.go

deadprogram:
Released with 0.20.0 so now closing. Thank you!

PR #2013, opened 2021-07-24
machine/rp2040: implement i2c

soypat:
Please take a look. Tested it with mpu6050 driver. Solves #1989.

PR #2015, opened 2021-07-25
machine/rp2040: add PWM implementation

soypat:
Please take a look. Tested and works with on-board LED and GPIO5. Solves #1993 .


soypat:
Tests have failed probably because two functions are missing: `boolToBit` and `umax32` which are due to be added with #2013. Below is an example of usage (PWM blinky)

<details><summary> PWM Blinky Example </summary>

```go
package main

import (
	"machine"
	"time"
)

var (
	LED  = machine.GPIO25
	pin2 = machine.GPIO26
	uart = machine.UART0
)

func main() {
	// configure UART to print stuff
	{
		err := uart.Configure(machine.UARTConfig{BaudRate: 9600})
		if err != nil {
			panic(err)
		}
	}

	pin := LED

	pin.Configure(machine.PinConfig{Mode: machine.PinPWM})
	pwm, err := machine.NewPWM(pin)
	if err != nil {
		println(err.Error())
	}
	var per uint64 = 1e9 / 500
	err = pwm.Configure(machine.PWMConfig{Period: per})
	if err != nil {
		println(err.Error())
	}
	ch, err := pwm.Channel(pin)
	if err != nil {
		println(err.Error())
	}
	println("slice", pwm.Peripheral(pin), " ch:", ch)
	if !pwm.IsEnabled() {
		println("NOT ENABLED")
	}
	println("period:", pwm.Period(), "target_period:", per)
	time.Sleep(time.Millisecond)

	for {

		for i := 1; i < 255; i++ {
			pwm.Set(ch, pwm.Top()/uint32(i))
			time.Sleep(time.Millisecond * 5)
		}
	}

}

```

</details>

deadprogram:
Ready for resolve of merge conflicts, please. :smile_cat:

soypat:
@deadprogram Done dealio

deadprogram:
Hi @soypat please see CI failure here: https://app.circleci.com/pipelines/github/tinygo-org/tinygo/5251/workflows/3e9e5c36-d044-4e6c-9fbd-340d32a20509/jobs/30455?invite=true#step-124-287

soypat:
Hello @deadprogram, seems like I used the wrong function name for `u32max`. Should be fixed by now.

src/machine/machine_rp2040_pwm.go
```diff
@@ -0,0 +1,494 @@
...
+
+// periodToFrequency converts a period (nanoseconds) to Hz.
+// Accurate for values under 1 second (1e9ns).
+//go:inline
+func frequencyToPeriod(f uint32) (T uint32) {
+	if f == 0 {
+		return 0xffff_ffff
+	}
+	return 1e9 / f
+}
```
aykevl: These two functions aren't used anywhere, so can be removed.

src/machine/machine_rp2040_pwm.go
```diff
@@ -0,0 +1,494 @@
...
+
+// Determine the PWM channel that is attached to the specified GPIO.
+// Each slice 0 to 7 has two channels, A and B.
+//go:inline
+func pwmGPIOToChannel(gpio Pin) (channel uint8) {
+	return uint8(gpio) & 1
+}
+
+// Returns the period of a clock cycle for the raspberry pi pico in nanoseconds.
+//go:inline
```
aykevl: Please don't use `//go:inline` unless you have measured this gives a performance or size improvement - and even then only do it when needed. In this case, I doubt it makes a difference (the compiler will likely already inline the function).

src/machine/machine_rp2040_pwm.go
```diff
@@ -0,0 +1,494 @@
...
+func (pwm *pwms) initSlice() {
+	pwm.slice = make(pwmGroup, 8)
+	s := pwm.slice
+	hw := pwm.hw
+	// PWM0
+	s[0].cc = &hw.CH0_CC
+	s[0].csr = &hw.CH0_CSR
+	s[0].ctr = &hw.CH0_CTR
+	s[0].div = &hw.CH0_DIV
+	s[0].top = &hw.CH0_TOP
```
aykevl: Hmm, it's a bit ugly that the RP2040 SVD file doesn't contain an array of slices.
I think this would be an appropriate use of `unsafe`:

```go
type pwmGroup struct {
    CSR volatile.Register32
    DIV volatile.Register32
    CTR volatile.Register32
    CC  volatile.Register32
    TOP volatile.Register32
}

func getPWMGroup(index uintptr) *pwmGroup {
    return (*pwmGroup)(unsafe.Pointer(rp.PWM))
}
```

You can then use it like `getPWMGroup(0).CSR.Set(...)`.

src/machine/machine_rp2040_pwm.go
```diff
@@ -0,0 +1,494 @@
...
+
+// setPeriod sets the pwm peripheral period (frequency). Calculates DIV_INT and sets it from following equation:
+//  cycles = (TOP+1) * (CSRPHCorrect + 1) * (DIV_INT + DIV_FRAC/16)
+// where cycles is amount of clock cycles per PWM period.
+//go:inline
+func (pwm *pwms) setPeriod(sliceNum uint8, period uint64) {
+	targetPeriod := uint32(period)
+	periodPerCycle := getPeriod()
+	top := pwm.getWrap(sliceNum)
+	phc := pwm.getPhaseCorrect(sliceNum)
```
aykevl: At the moment, phase correct PWM isn't supported so I would suggest not taking this into account.
(I believe phase correct PWM is only relevant for specific motor control applications)

aykevl (commented):
It appears that the PWM of the RP2040 is structured a little bit differently than other chips: it groups all PWM timers into one peripheral. Most chips have multiple PWMs where each PWM instance has just one timer. However, if you look closely you can see that it is actually not that different: it's as if the RP2040 has 8 PWM peripherals that each have two channel outputs. Therefore, I would suggest exposing it like that:

```go
var (
    PWM0 = ...
    PWM1 = ...
    PWM2 = ...
    // etc
}
```

In this API, you can use the PWM like this:
```go
pwm := machine.PWM0
pwm.Configure(...)        // enables the given PWM
ch, _ := pwm.Channel(pin) // configures the pin
ch.Set(ch, value)         // set the channel value
```

I understand the need for a `NewPWM`, and I agree we should have something like that (also for other chips). However, there is one big catch: it's not clear which PWMs alias each other. For example, if you set the frequency for pin 2, it will also change the frequency for pin 3, 18, and 19. This is in fact one of the main reasons of the somewhat complex API. That's why I would suggest leaving it out for now.

(Also, I think the name `NewPWM` is misleading as the function doesn't create anything new: PWM resources are fixed hardware resources and can't be made on the fly).

soypat:
@deadprogram I think it's all good now

soypat:
@aykevl I've applied sweeping changes inspired by your comments. There may be more use of unsafe but I think this is much cleaner. Let me know what you think!

deadprogram:
CI tests are failing, and there is a merge conflict. Can you please look at that @soypat thank you!

src/machine/machine_rp2040_pwm.go
soypat: Do you mean I should leave it commented out as a TODO for when phase correct is added, if ever?

soypat:
I changed a method to a function. One can obtain the peripheral number with `machine.PWMPeripheral`. I think it makes more sense than having it as a method.

deadprogram:
Was not able to get working with the PWM example in the main TinyGo examples. Some mappings are missing, and I might not have them correct for the Pico board.

soypat:
@deadprogram I had to add a `pin.Configure` call inside `pwm.Channel` for the API to be the same as other implementations. I have tested it and it works with the on-board LED on the pico (which I added to the `examples/pwm` files.

deadprogram:
Tested and is now working as expected. Squashing/merging now, thanks for working on this @soypat

PR #2021, opened 2021-07-31
wasm: add support for the crypto/rand package

aykevl:
This is done via wasi-libc and the WASI interface, for ease of maintenance (only one implementation for both WASI and JS/browsers).


deadprogram:
Any comments from WASM/WASI oriented people?

fgsch:
Reads good to me 👍

deadprogram:
Seems random enough, merging. :game_die: :game_die:

PR #2022, opened 2021-08-01
transform: improve GC stack slot pass to work around a bug

### LINKED ISSUE #1790: wasm: musttail call must precede a ret with an optional bitcast
```
% tinygo build -o try.wasm -target wasm
musttail call must precede a ret with an optional bitcast
  musttail call fastcc void @runtime.runqueuePushBack(%"internal/task.Task"* nonnull %parentHandle.reload1), !dbg !1172
error: GC pass caused a verification failure
```
Version reports
```
tinygo version 0.17.0 darwin/amd64 (using go version go1.16.3 and LLVM version 11.1.0)
```
Problem does not manifest when -opt=1 is used. Problem also reproduced from the latest dev branch.

Ran into this when building for wasm where a js.FuncOf was being called. Reduced the problem to find it is some combination of calls from the syscall/js package which I reduced further to a stand alone main.go file below. But the problem goes away if the sync.Lock call is removed so that sync package I felt I had to keep in the reduction. Many attempts at reducing the main.go further also made the problem go away so I could not reduce it down to just one line that makes a difference. Sorry there are so many functions in this reproduction.

Removing the sync.Lock makes the problem go away. Also just removing the one call to panic makes the problem go away. The great job LLVM does at removing dead code though makes it hard to guess why a line does or does not affect tail call logic. Even just removing the *int field from the struct effects the outcome, even though the code doesn't use the field - guess it is the size of the struct.

```
// +build js,wasm

package main

import "sync"

const (
	T0 int = iota
	T1
	T2
)

var (
	mu sync.Mutex
	v1 = make1(1) // change 1 to 0, get another kind of crash
	v2 = v1.Method1()
)

func main() {
	_ = LockAndReturn()
}
func LockAndReturn() error {
	mu.Lock()
	mu.Unlock()
	return nil
}

type S1 struct {
	f1 int
	f2 *int
}

func make1(b byte) S1 {
	return S1{f1: (int(b))}
}

func (v S1) Method1() S1 {
	if t := v.Type(); !isObject(t) {
		panic(&ValueError{})
	}
	return makeValue(valueGet(0, ""))
}

func makeValue(r int) S1 {
	return S1{}
}

func valueGet(v int, p string) int

type ValueError struct {
	Method string
}

func (e *ValueError) Error() string {
	return "something" + e.Method
}

func isObject(t int) bool {
	return t == T1 || t == T2
}

func (v S1) Type() int {
	if v.f1 == 0 {
		return T0
	}
	return T2
}
```

aykevl:
Bug 1790 ("musttail call must precede a ret with an optional bitcast")
is caused by the GC stack slot pass inserting a store instruction
between a musttail call and a return instruction. This is not allowed in
LLVM IR.

One solution would be to remove the musttail. That would probably work,
but 1) the go-llvm API doesn't support this and 2) this might have
unforeseen consequences. What I've done in this commit is to move the
store instruction to a position earlier in the basic block, just after
the last access to the GC stack slot alloca.

Thanks to @fgsch for a very small repro, which I've used as a regression
test.

fixes #1790

### DIFF
--- a/testdata/goroutines.go
+++ b/testdata/goroutines.go
@@ -76,6 +76,8 @@ func main() {
 	testGoOnBuiltins()
 
 	testCond()
+
+	testIssue1790()
 }
 
 func acquire(m *sync.Mutex) {
@@ -197,4 +199,14 @@ func testCond() {
 	if !ok {
 		panic("missing queued notification")
 	}
+}
+
+var once sync.Once
+
+// This tests a fix for issue 1790:
+// https://github.com/tinygo-org/tinygo/issues/1790
+func testIssue1790() *int {
+	once.Do(func() {})
+	i := 0
+	return &i
 }
--- a/transform/gc.go
+++ b/transform/gc.go
@@ -246,6 +246,7 @@ func MakeGCStackSlots(mod llvm.Module) bool {
 		}
 
 		// Do a store to the stack object after each new pointer that is created.
+		pointerStores := make(map[llvm.Value]struct{})
 		for i, ptr := range pointers {
 			// Insert the store after the pointer value is created.
 			insertionPoint := llvm.NextInstruction(ptr)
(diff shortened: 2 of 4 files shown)


fgsch:
This fixes the issue I ran into and reads well to my limited knowledge eyes.

deadprogram:
Thanks @aykevl for the fix and @fgsch for the repro and review.

Now merging.

Issue #2006, opened 2021-07-19
Go 1.17 compabability


deadprogram:
Has anyone looked into the upcoming Go 1.17 release, and if TinyGo will require any changes to accommodate it?

aykevl:
I looked at the release notes and there are in fact a few changes to the language which likely will need to be supported in TinyGo too. The `unsafe` changes probably don't have an effect (except that TinyGo may need to be built with Go 1.17 to support Go 1.17), but the new slice to array cast will likely require a change in TinyGo.

https://tip.golang.org/doc/go1.17#language

aykevl:
I intend to wait a bit until the dust is settled on the x/tools/go/ssa package: https://github.com/golang/go/issues/46987

aykevl:
Also, the math package got some problematic changes that I'm not sure how to solve.
In other words, supporting Go 1.17 will unfortunately be a bit more work.

oflebbe:
@aykevl Just for my curiosity : Regarding math I see only the MaxUInt etc constants mentioned, which IMO doesn't sound too alarming to me. Can you go into more detail what might be problematic to solve ?

aykevl:
The problem is this commit: https://github.com/golang/go/commit/1d20a362d0ca4898d77865e314ef6f73582daef0. Previously, it was simple to forward the math functions to the naive/non-optimized versions. But with that commit, it becomes a bit more complicated. (This is why I started work on supporting the optimized assembly versions).
Also see this PR: https://github.com/tinygo-org/tinygo/pull/164

oflebbe:
Any chance to have [FPU](https://github.com/tinygo-org/tinygo/commit/f79e66ac2e66a5338ec3ce3c4c992567b07c4959) reenabled again , or did I miss it ? Otherwise assembly will be really hard.

aykevl:
> Any chance to have [FPU](https://github.com/tinygo-org/tinygo/commit/f79e66ac2e66a5338ec3ce3c4c992567b07c4959) reenabled again

This is only for the Cortex-M. The FPU is always enabled on desktop systems (386, amd64, arm64). The assembly support I'm working on will only be available for targets supported upstream (running on Linux and MacOS), not for any other architectures. It's not possible (or very hard) to automatically convert aarch64 (arm64) machine code to T32 (32-bit ARM, Cortex-M) machine code.

oflebbe:
Ah interesting. 25years ago I did an 387 FPU binding for the linux libm :)
[...]

PR #2030, opened 2021-08-06
Improve/refactor math package

aykevl:
This PR does many things which all depend on each other:

- It bumps the minimum required x86 processor to the Pentium 4, which seems plenty old to me but gets the math package to pass all tests (by avoiding the problematic x87 coprocessor).
- It fixes math package tests on arm64 and wasm.
- It prepares for changes in the math package in Go 1.17 (see #2006).
- As an added bonus, it also adds support for some hashing packages: crypto/md5, crypto/sha1, crypto/sha256, and crypto/sha512.


deadprogram:
Took a while to look this over, as it does do a lot. As far as I am concerned this is a pragmatic solution to the problem of Go 1.17+. Also dropping Pentium 4 is legit.

Anyone else want to chime in before I merge this collection of math goodies?

aykevl:
> Also dropping Pentium 4 is legit.

Actually the Pentium 4 will then be the oldest supported processor. We'll be dropping support for Pentium 3 and older.

deadprogram:
OK I am satisfied this is a big improvement. Now merging, thanks @aykevl 

:heavy_plus_sign: :heavy_minus_sign:

PR #2032, opened 2021-08-06
crypto/rand: switch to arc4random_buf on WASI and BSD-like systems

aykevl:
This doesn't have the potential blocking issue of the getentropy call (which calls WASI random_get when using wasi-libc) and should therefore be a lot faster.

For context, this is what the random_get documentation says:

> Write high-quality random data into a buffer. This function blocks
> when the implementation is unable to immediately provide sufficient
> high-quality random data. This function may execute slowly, so when
> large mounts of random data are required, it's advisable to use this
> function to seed a pseudo-random number generator, rather than to
> provide the random data directly.


fgsch:
👍

deadprogram:
:infinity: :game_die: merging.

PR #2031, opened 2021-08-06
ci: run stdlib tests for other architectures

aykevl:
Currently this runs the tests for amd64, 386, arm, arm64, and wasi. I
tried adding wasm and a baremetal target (such as riscv-qemu) but they
don't pass the tests yet and should be fixed in the future. At least
this change should make sure that existing tests continue to pass on a
variety of architectures.

Depends on #2030 for 386 and wasm fixes, so marked as draft until then.

PR #2034, opened 2021-08-06
more llvm builtins

oflebbe:
Use more llvm builtins.

llvm.minnum and maxnum are not IEEE754 compliant as the don't handle NaN, Inf and zeros correctly. 

There are a few more llvm builtins ... So if you agree to the general approach I can add them as well

### DIFF
--- a/src/runtime/math.go
+++ b/src/runtime/math.go
@@ -57,7 +57,7 @@ func math_cbrt(x float64) float64
 
 //go:linkname math_Ceil math.Ceil
 func math_Ceil(x float64) float64 {
-	if GOARCH == "arm64" || GOARCH == "wasm" {
+	if GOARCH == "arm64" || GOARCH == "wasm" || GOARCH == "i386" || GOARCH == "amd64" {
 		return llvm_ceil(x)
 	}
 	return math_ceil(x)
@@ -70,10 +70,18 @@ func llvm_ceil(x float64) float64
 func math_ceil(x float64) float64
 
 //go:linkname math_Cos math.Cos
-func math_Cos(x float64) float64 { return math_cos(x) }
+func math_Cos(x float64) float64 {
+	if GOARCH == "arm64" || GOARCH == "wasm" || GOARCH == "i386" || GOARCH == "amd64" {
+		return llvm_cos(x)
+	}
+	return math_cos(x)
+}
 
 //go:linkname math_cos math.cos
 func math_cos(x float64) float64
+
+//export llvm.cos.f64
+func llvm_cos(x float64) float64
 
 //go:linkname math_Cosh math.Cosh
 func math_Cosh(x float64) float64 { return math_cosh(x) }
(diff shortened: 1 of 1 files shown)


oflebbe:
Arg on linux it needs to link against libm . But how can I place -lm at the _end_ of the link line ???

aykevl:
As long as `tinygo test math` passes on all relevant architectures, sure. It would be nice if LLVM can wrangle a little bit more performance out of the code by knowing what exactly it is doing. However, I think it would be better to do this based on #2030 (specifically the commit "compiler: move LLVM math builtin support into the compiler").
You can test whether it works across architectures this way (see https://github.com/tinygo-org/tinygo/pull/2031):

```
GOARCH=amd64 tinygo test math
GOARCH=386 tinygo test math
GOARCH=arm64 tinygo test math
GOARCH=arm tinygo test math
tinygo test -target=wasi math
```

Honestly, I'm a little bit afraid this would expose TinyGo a bit too much to variations in libc/libm implementations. Maybe it works, but various libc/libm implementations might have subtly different implementations of the math functions.

> Arg on linux it needs to link against libm . But how can I place -lm at the _end_ of the link line ???

Right. I'm using ld.lld on my computer because it is faster so that's when I tried to do this, it worked with `-lm` at the start of the linker flags. GNU ld (normally used on Linux) has the problem that it has much stricter requirements on parameter order - a problem ld.lld mostly avoids. I think the best solution would be to add it as a new libc version here:

https://github.com/tinygo-org/tinygo/blob/64d048c47c95bc0a70b4cff109afc1f9039705a4/builder/build.go#L473-L495

Eventually I'd like to switch to musl so that TinyGo can produce fully static executables on Linux (and can more easily cross-compile) but we're not there yet.

src/runtime/math.go
```diff
@@ -169,28 +178,59 @@ func math_log2(x float64) float64
...
+		if y == 0 && x == 0 {
+			// if both are negative zero
+			if math.Signbit(x) {
+				return x // returns -0.0
+			}
+			if math.Signbit(y) {
+				return y // returns -0.0
+			}
+			return 0
+		}
```
aykevl: With all of this complexity (and with the requirement to import the math package from the runtime!) I would actually much prefer to just use the standard Go  versions. This is basically a re-implementation of the version in the math package and just as long.

oflebbe: Fully agree.

oflebbe:
> Eventually I'd like to switch to musl so that TinyGo can produce fully static executables on Linux (and can more easily cross-compile) but we're not there yet.

I had a quick look over the fence at ziglang.org yesterday. Zig is providing crosscompilation (incl. musl) for C Code as an first-class option . All one need is using "zig cc" . I was asking myself if tinygo could borrow the libc within zig or just use the zig cc as (optional?) C compiler

oflebbe:
Ok, looking even more deeply into it...  @aykevl  you are right, too much internals of libc/libm need to be exposed . withdrawing proposal and looking into #2030 instead

PR #2035, opened 2021-08-07
builder: run binaryen after the linker when generating wasm

niaow:
Right now it builds and links.
It needs a lot of cleanup still, and debug symbol support had to be disabled for it to build.
This can use the `wasm-opt` command when dynamically built and `libbinaryen` when built normally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,6 +96,7 @@ ifneq ("$(wildcard $(LLVM_BUILDDIR)/bin/llvm-config*)","")
     CGO_LDFLAGS+=$(abspath $(LLVM_BUILDDIR))/lib/lib$(LIBCLANG_NAME).a -L$(abspath $(LLVM_BUILDDIR)/lib) $(CLANG_LIBS) $(LLD_LIBS) $(shell $(LLVM_BUILDDIR)/bin/llvm-config --ldflags --libs --system-libs $(LLVM_COMPONENTS)) -lstdc++ $(CGO_LDFLAGS_EXTRA)
 endif
 ifneq ("$(wildcard lib/binaryen/lib/libbinaryen.a)","")
+	CGO_CPPFLAGS+=-I$(abspath lib/binaryen/src/)
 	CGO_LDFLAGS+=$(abspath lib/binaryen/lib/libbinaryen.a)
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,9 @@ ifneq ("$(wildcard $(LLVM_BUILDDIR)/bin/llvm-config*)","")
     CGO_CXXFLAGS=-std=c++14
     CGO_LDFLAGS+=$(abspath $(LLVM_BUILDDIR))/lib/lib$(LIBCLANG_NAME).a -L$(abspath $(LLVM_BUILDDIR)/lib) $(CLANG_LIBS) $(LLD_LIBS) $(shell $(LLVM_BUILDDIR)/bin/llvm-config --ldflags --libs --system-libs $(LLVM_COMPONENTS)) -lstdc++ $(CGO_LDFLAGS_EXTRA)
 endif
-
+ifneq ("$(wildcard lib/binaryen/lib/libbinaryen.a)","")
+	CGO_LDFLAGS+=$(abspath lib/binaryen/lib/libbinaryen.a)
+endif
 
 clean:
 	@rm -rf build
@@ -168,6 +170,13 @@ $(LLVM_BUILDDIR)/build.ninja: llvm-source
 $(LLVM_BUILDDIR): $(LLVM_BUILDDIR)/build.ninja
 	cd $(LLVM_BUILDDIR); ninja $(NINJA_BUILD_TARGETS)
 
+# Configure binaryen.
+lib/binaryen/build.ninja:
+	cd lib/binaryen; cmake -G Ninja -DBUILD_STATIC_LIB=ON -DBUILD_LLVM_DWARF=OFF .
+
+# Build binaryen.
+lib/binaryen/lib/libbinaryen.a: lib/binaryen/build.ninja
+	ninja -C lib/binaryen lib/libbinaryen.a
 
 # Build wasi-libc sysroot
 .PHONY: wasi-libc
--- a/builder/build.go
+++ b/builder/build.go
@@ -621,6 +621,30 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 				}
 			}
 
+			// Run wasm-opt if necessary.
+			if config.GOARCH() == "wasm" {
+				var cfg BinaryenConfig
+				switch config.Options.Opt {
+				case "none", "0":
+				case "1":
+					cfg.OptLevel = 1
+				case "2":
+					cfg.OptLevel = 2
+				case "s":
+					cfg.OptLevel = 2
+					cfg.ShrinkLevel = 1
+				case "z":
+					cfg.OptLevel = 2
+					cfg.ShrinkLevel = 2
+				default:
+					return fmt.Errorf("unknown opt level: %q", config.Options.Opt)
+				}
+				err := WasmOpt(executable, executable, cfg)
+				if err != nil {
+					return fmt.Errorf("wasm-opt failed: %w", err)
+				}
+			}
+
 			if config.Options.PrintSizes == "short" || config.Options.PrintSizes == "full" {
 				sizes, err := loadProgramSize(executable)
 				if err != nil {
--- a/builder/tools-builtin.go
+++ b/builder/tools-builtin.go
@@ -4,13 +4,18 @@ package builder
 
 import (
 	"errors"
+	"fmt"
+	"os"
+	"sync"
 	"unsafe"
 )
 
 /*
 #cgo CXXFLAGS: -fno-rtti
+#cgo LDFLAGS: -lbinaryen
 #include <stdbool.h>
 #include <stdlib.h>
+#include <binaryen-c.h>
 bool tinygo_clang_driver(int argc, char **argv);
 bool tinygo_link_elf(int argc, char **argv);
 bool tinygo_link_wasm(int argc, char **argv);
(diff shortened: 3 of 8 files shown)


niaow:
To build binaryen, you can use:
```
make lib/binaryen/lib/libbinaryen.a
```

niaow:
So I now have CI working on all non-windows platforms. For some reason it looks like the build process is stalling in a smoketest with WASM on Windows.

niaow:
This has now been rebased on the latest dev.

niaow:
It looks like the windows problem is that it is doing. . . nothing:
image

niaow:
The windows bug appears to be a memory access violation in the `ExternalInt64AsPtr` transform pass

niaow:
There also appears to be an issue with CGo support of some sort, which I have not been able to untangle.

builder/tools-builtin.go
```diff
@@ -52,3 +56,51 @@ func RunTool(tool string, args ...string) error {
 	}
 	return nil
 }
+
+// wasmOptLock exists because the binaryen C API uses global state.
+// In the future, we can hopefully avoid this by re-execing.
+var wasmOptLock sync.Mutex
```
aykevl, line 62: We already do this for Clang and LLD, so it wouldn't be too much trouble to add another for wasm-opt. Either way, that's easy to do at a later time (no need to do it from the beginning).

aykevl (commented):
Interesting work! Yeah, getting things to build reliable in a cross-platform way can be very painful. I have had plenty of experience with this (especially when originally porting TinyGo to Windows, that took me a few days of debugging). It would be pretty awesome to have wasm-opt built into TinyGo.

Makefile
```diff
@@ -95,7 +95,10 @@ ifneq ("$(wildcard $(LLVM_BUILDDIR)/bin/llvm-config*)","")
     CGO_CXXFLAGS=-std=c++14
     CGO_LDFLAGS+=$(abspath $(LLVM_BUILDDIR))/lib/lib$(LIBCLANG_NAME).a -L$(abspath $(LLVM_BUILDDIR)/lib) $(CLANG_LIBS) $(LLD_LIBS) $(shell $(LLVM_BUILDDIR)/bin/llvm-config --ldflags --libs --system-libs $(LLVM_COMPONENTS)) -lstdc++ $(CGO_LDFLAGS_EXTRA)
 endif
-
+ifneq ("$(wildcard lib/binaryen/lib/libbinaryen.a)","")
+	CGO_CPPFLAGS+=-I$(abspath lib/binaryen/src/)
+	CGO_LDFLAGS+=$(abspath lib/binaryen/lib/libbinaryen.a)
+endif
```
aykevl, line 101: I think these flags should always be added when linking statically against LLVM.

Makefile
```diff
@@ -168,6 +171,13 @@ $(LLVM_BUILDDIR)/build.ninja: llvm-source
 $(LLVM_BUILDDIR): $(LLVM_BUILDDIR)/build.ninja
 	cd $(LLVM_BUILDDIR); ninja $(NINJA_BUILD_TARGETS)
 
+# Configure binaryen.
+lib/binaryen/build.ninja:
+	cd lib/binaryen; cmake -G Ninja -DBUILD_STATIC_LIB=ON -DBUILD_LLVM_DWARF=OFF .
+
+# Build binaryen.
+lib/binaryen/lib/libbinaryen.a: lib/binaryen/build.ninja
+	ninja -C lib/binaryen lib/libbinaryen.a
```
aykevl, line 180: I think this should be a dependency of the `tinygo:` make target.

aykevl:
I tried compiling wasm-opt using LLVM, using the following line in the Makefile:

    cd lib/binaryen; cmake -G Ninja -DBUILD_STATIC_LIB=ON -DBUILD_LLVM_DWARF=ON -DLLVM_SOURCE="$(abspath $(LLVM_PROJECTDIR)/llvm)" -DLLVM_BUILD="$(abspath $(LLVM_BUILDDIR))" .

And the following patch to binaryen:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 45b4a0282..35ebdb51e 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,11 +120,17 @@ if(NOT EMSCRIPTEN)
   endif()
 endif()
 
+set(LLVM_SOURCE "${CMAKE_CURRENT_SOURCE_DIR}/third_party/llvm-project" CACHE STRING "LLVM source directory (llvm-project/llvm)")
+set(LLVM_BUILD "" CACHE STRING "LLVM build directory")
+
 # Compiler setup.
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/src)
 if(BUILD_LLVM_DWARF)
-  include_directories(${CMAKE_CURRENT_SOURCE_DIR}/third_party/llvm-project/include)
+  include_directories(${LLVM_SOURCE}/include)
+  if(LLVM_BUILD)
+    include_directories(${LLVM_BUILD}/include)
+  endif()
 endif()
 
 # Add output directory to include path so config.h can be found
@@ -293,7 +299,7 @@ add_subdirectory(src/emscripten-optimizer)
 add_subdirectory(src/passes)
 add_subdirectory(src/support)
 add_subdirectory(src/wasm)
-add_subdirectory(third_party)
+#add_subdirectory(third_party)
 
 # Configure lit tests
 add_subdirectory(test/lit)
@@ -308,9 +314,9 @@ set(binaryen_objs
     $<TARGET_OBJECTS:cfg>
     $<TARGET_OBJECTS:support>)
 
-if(BUILD_LLVM_DWARF)
-  SET(binaryen_objs ${binaryen_objs} $<TARGET_OBJECTS:llvm_dwarf>)
-endif()
+#if(BUILD_LLVM_DWARF)
+#  SET(binaryen_objs ${binaryen_objs} $<TARGET_OBJECTS:llvm_dwarf>)
+#endif()
 
 # Sources.
 
diff --git a/src/wasm/wasm-debug.cpp b/src/wasm/wasm-debug.cpp
index 60e97a27e..d38af1c36 100644
--- a/src/wasm/wasm-debug.cpp
+++ b/src/wasm/wasm-debug.cpp
@@ -20,7 +20,7 @@
 #ifdef BUILD_LLVM_DWARF
 #include "llvm/ObjectYAML/DWARFEmitter.h"
 #include "llvm/ObjectYAML/DWARFYAML.h"
-#include "llvm/include/llvm/DebugInfo/DWARFContext.h"
+#include "llvm/DebugInfo/DWARF/DWARFContext.h"
 
 std::error_code dwarf2yaml(llvm::DWARFContext& DCtx, llvm::DWARFYAML::Data& Y);
 #endif
```

but it's failing because Binaryen is built for an older version of LLVM. So it doesn't look like static linking while preserving DWARF debuginfo would be possible at all (and preserving DWARF is IMHO a necessity, certainly if you ask @dkegel-fastly).

builder/build.go
```diff
@@ -621,6 +621,30 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 				}
 			}
 
+			// Run wasm-opt if necessary.
+			if config.GOARCH() == "wasm" {
```
aykevl, line 625: This should be `strings.HasPrefix(config.Triple(), "wasm")` because WASI uses GOOS=linux GOARCH=arm because of reasons (see #1964 for details).

aykevl:
@niaow is this something you intend to continue working on? Otherwise I can take over this PR if needed.

Right now it seems to me that it is not possible to link against libbinaryen while libbinaryen is built with DWARF support. Because we do want DWARF support, I do not think linking against libbinaryen is a reasonable option. Therefore, we always need to call the external wasm-opt binary.
We can either bundle wasm-opt (~12MB original, ~8MB stripped, or ~3MB compressed) or expect the user to install this manually. A manual install is less of a problem than it may seem: binaryen is already packaged for MacOS (Homebrew) and Windows (Chocolatey). And it is only necessary for WebAssembly. But bundling will definitely be less of a hassle for new users.

ghost:
@aykevl Recommended way to install tinygo on Windows - scoop manager, but unfortunately binaryen doesn't supported there, but any way, it's not a very promise feature to be able to abfuscate Wasm-code to run it after that locally in Browser, more important to be able to optimize or shrink code for microcontrollers (here we already have LLVM).

dkegel-fastly:
I think the motivation for allowing binaryen as a postprocessing pass is to try fixing https://github.com/tinygo-org/tinygo/issues/2101 using asyncify, see https://github.com/tinygo-org/tinygo/pull/2011 
so this is about correctness more than optimization...

aykevl:
> Recommended way to install tinygo on Windows - scoop manager, but unfortunately binaryen doesn't supported there,

Binaryen can easily be added to Scoop. No problem (but see https://github.com/WebAssembly/binaryen/issues/4148).

> it's not a very promise feature to be able to abfuscate Wasm-code to run it after that locally in Browser, more important to be able to optimize or shrink code for microcontrollers (here we already have LLVM).

Maybe not for you, but some people really want to use TinyGo for WebAssembly.

Also, @dkegel-fastly is correct. The goal is not to shrink binary size. The goal is to be able to use Asyncify which is part of binaryen.

ghost:
@aykevl @dkegel-fastly thanks, now I see it, [Binaryen's Asyncify](https://kripken.github.io/blog/wasm/2019/07/16/asyncify.html) really looks interesting.

ghost:
@aykevl I am a big fan of WebAssembly, some time ago [I even thought](https://lists.w3.org/Archives/Public/public-webassembly/2018Nov/0000.html) about using it directly on Microcontrollers, like now [wasm3](https://github.com/wasm3/wasm3) do, but after all I think that your approarch with tinygo looks more progmatic.

aykevl:
I have created a PR to add Binaryen to Scoop: https://github.com/ScoopInstaller/Main/pull/2740

> @aykevl I am a big fan of WebAssembly, some time ago [I even thought](https://lists.w3.org/Archives/Public/public-webassembly/2018Nov/0000.html) about using it directly on Microcontrollers, like now [wasm3](https://github.com/wasm3/wasm3) do

I actually made something very similar! Please take a look here: https://play.tinygo.org/. It implements the same machine package but runs in the browser using WebAssembly.

niaow:
> @niaow is this something you intend to continue working on? Otherwise I can take over this PR if needed.
> 
> Right now it seems to me that it is not possible to link against libbinaryen while libbinaryen is built with DWARF support. Because we do want DWARF support, I do not think linking against libbinaryen is a reasonable option. Therefore, we always need to call the external wasm-opt binary. We can either bundle wasm-opt (~12MB original, ~8MB stripped, or ~3MB compressed) or expect the user to install this manually. A manual install is less of a problem than it may seem: binaryen is already packaged for MacOS (Homebrew) and Windows (Chocolatey). And it is only necessary for WebAssembly. But bundling will definitely be less of a hassle for new users.

I think i can get back to the asyncify work a bit now, but I updated #2011 before to shell out to binaryen for now, which I think is the most reasonable option until binaryen can be built with tinygo's copy of LLVM. I think it is probably more appropriate just to close this for now.

PR #2039, opened 2021-08-10
testing: add support for the -test.v flag

aykevl:
This flag is passed automatically with the (new) -v flag for TinyGo. For example, this prints all the test outputs:

    $ tinygo test -v crypto/md5
    === RUN   TestGolden
    --- PASS: TestGolden
    === RUN   TestGoldenMarshal
    --- PASS: TestGoldenMarshal
    === RUN   TestLarge
    --- PASS: TestLarge
    === RUN   TestBlockGeneric
    --- PASS: TestBlockGeneric
    === RUN   TestLargeHashes
    --- PASS: TestLargeHashes
    PASS
    ok  	crypto/md5	0.002s

This prints just a summary:

    $ tinygo test crypto/md5
    PASS
    ok  	crypto/md5	0.002s

(The superfluous 'PASS' message may be removed in the future).

This is especially useful when testing a large number of packages:

    $ tinygo test crypto/md5 crypto/sha1 crypto/sha256 crypto/sha512
    PASS
    ok  	crypto/md5	0.002s
    PASS
    ok  	crypto/sha1	0.043s
    PASS
    ok  	crypto/sha256	0.002s
    PASS
    ok  	crypto/sha512	0.003s

At the moment, the -test.v flag is not supplied to binaries running in emulation. I intend to fix this after
https://github.com/tinygo-org/tinygo/pull/2038 lands by refactoring `runPackageTest`, `Run`, and `runTestWithConfig` in the main package which al do something similar.

PR #2043, opened 2021-08-14
Improve/refactor testing of binaries running under qemu-system-*

aykevl:
The first commit fixes the exit code of cortex-m-qemu, hifive1-qemu, and riscv-qemu. The second commit uses this fix to simplify `tinygo test` a bit and improve the output a little bit (removing of a redundant "PASS" line). These two commits could perhaps be squashed but because these changes are somewhat independent I decided to keep them in separate commits.

The `-v` flag feature for emulators is still pending (see #2039), this is just a refactor I'd like to do before getting to the `-v` flag on baremetal and JS.


aykevl:

deadprogram:
I am not the biggest exert on QEMU by any means, but looking thru the changes they seem reasonable, and also passing all tests, so now merging. Thanks for working on this @aykevl

PR #2044, opened 2021-08-14
ci: bump Go version in Linux builds to Go 1.16

aykevl:
Not sure why it was at 1.14. It seems like a better idea to build releases using the latest stable Go version.

This should also fix the CI build failure in #2031.


aykevl:
Update: I've now also switched the assert-test-linux and build-linux builds from stretch to buster because the stretch builds aren't provided anymore from apt.llvm.org.

PR #2048, opened 2021-08-14
Go 1.17

aykevl:
This is still a WIP, not all features are supported yet.

See #2006


aykevl:
I've now also implemented the [three new language features](https://tip.golang.org/doc/go1.17#language) of Go 1.17.

aykevl:
This should be ready once CircleCI publishes the 1.17 containers: https://hub.docker.com/r/circleci/golang/tags?page=1&ordering=last_updated&name=1.17-stretch

dkegel-fastly:
Have you tried building with circleci's rc2 image, e.g. circleci/golang:1.17rc2-buster ?

niaow (approved):
This looks fairly complete, and should be good to go once circleci is ready

.circleci/config.yml
```diff
@@ -379,12 +379,12 @@ jobs:
           llvm: "11"
   assert-test-linux:
     docker:
-      - image: circleci/golang:1.16-stretch
+      - image: circleci/golang:1.17-stretch
```
deadprogram, line 382: I am not sure that CircleCI actually plans to release a 1.17 version of this image. The reason is that they consider this image legacy. See https://circleci.com/developer/images/image/cimg/go

aykevl: The problem is that we rely on a specific Debian version. The cimg images seem to be based on Ubuntu (probably a recent version?). The version/variant of the underlying OS is important because it is the lowest that the TinyGo binaries will run on. For example, if we switched to Debian bullseye (the latest version), the prebuilt binaries of TinyGo wouldn't run on Debian buster anymore (the previous Debian version). Therefore, we need to use a base image that is as old as possible but still somewhat supported.

I'm thinking that we should probably create our own images with all software preinstalled, that should hopefully cut down on CI build time (no need to install anything) but more importantly provides stability (no packages that suddenly disappear from upstream). But perhaps not before this PR is merged.

In this case, the -rc2 image versions have been published so I'm hoping the final version will also be published soon.

deadprogram: I am not too sure on that. It has been a lot longer than previous releases, and the repo with the legacy images is now archived and read-only: https://github.com/CircleCI-Public/circleci-dockerfiles

In addition the LLVM 11 images seemed to have issues, which is why I merged https://github.com/tinygo-org/tinygo/commit/931f87f96a36565cf01291e9a306a808884e5b97 so that the GH action would stop failing on the *current* version.

deadprogram: Looks like it was finally pushed: https://hub.docker.com/layers/circleci/golang/1.17-stretch

deadprogram:
CI now passing. Merging, thanks for working on this @aykevl

PR #2049, opened 2021-08-15
Fix maximum slice size

aykevl:
Extracted from #2048. This brings the maximum slice size (in `make([]T, len)`) the same as in upstream Go. I've split this in two commits so that the second commit clearly shows the (small) change in the test output.

This is unlikely to have much (if any) real world impact, but I made the change for consistency.


deadprogram:
Makes sense to me. Should pass all tests now that 1.13 and 1.14 support have been removed. Merging.

PR #2050, opened 2021-08-15
reflect: add StructField.IsExported method

aykevl:
This field was introduced in Go 1.17 and is used by the encoding/json package (starting with Go 1.17).

This commit was extracted from #2048 because the tests don't depend on Go 1.17.


deadprogram:
Now that #2048 was merged this code is already in `dev` so now closing.

PR #2051, opened 2021-08-15
ci: drop support for Go 1.13 and 1.14

aykevl:
They aren't supported anymore in CI, and because untested code is broken code, let's remove support for these Go versions altogether.

This should address the recent CI failures.


QuLogic (approved):

deadprogram:
:scissors: thanks @aykevl now merging.

Issue #2054, opened 2021-08-17
tinygo problems blocking support for newly released go 1.17?


dkegel-fastly:
Lucky me, I happened to install the latest go right when go 1.17 was released, and discovered A Problem Or Two :-) running tinygo.
````
$ go version
go version go1.17 darwin/amd64
$ tinygo version
tinygo version 0.20.0-dev darwin/amd64 (using go version go1.17 and LLVM version 11.1.0)
````

First problem:

```
$ tinygo build -o foo
error: could not read version from GOROOT (/usr/local/go): Invalid go version output:
```

dtruss shows
```
 open("/usr/local/go/src/runtime/internal/sys/zversion.go\0", 0x1000000, 0x0)		 = 3 0
 read(0x3, "// Code generated by go tool dist; DO NOT EDIT.\n\npackage sys\n\nconst StackGuardMultiplierDefault = 1\n\0", 0x200)		 = 100 0
 write(0x2, "error: could not read version from GOROOT (/usr/local/go): Invalid go version ...
```

Looks like https://github.com/golang/go/commit/5159c836410535 removed the line ```const TheVersion = `%s``` that tinygo's GorootVersionString() was expecting.

That's an easy fix - make the first hunk fall through if zversion.go doesn't match the regexp.

Once past that, the next problem is:

```
$ tinygo build -o foo
error: requires go version 1.15 through 1.16, got go1.17
```

tinygo's builder/config.go says
```
	if major != 1 || minor < 15 || minor > 16 {
		return nil, fmt.Errorf("requires go version 1.15 through 1.16, got go%d.%d", major, minor)
```

Once past that, tinygo's own test suite fails as follows (lightly summarized):
```
$ go test
Undefined symbols for architecture x86_64:
  "_math.archFloor", referenced from:
      _runtime.run$1$gowrapper in main.o
  "_math.archHypot", referenced from:
      _runtime.run$1$gowrapper in main.o
  "_math.archMax", referenced from:
      _runtime.run$1$gowrapper in main.o
  "_math.archCeil", referenced from:
      _runtime.run$1$gowrapper in main.o
  "_math.archMin", referenced from:
      _runtime.run$1$gowrapper in main.o
  "_math.archTrunc", referenced from:
      _runtime.run$1$gowrapper in main.o
  "_math.archExp", referenced from:
      _math.Exp in main.o
  "_math.archLog", referenced from:
      _math.Log in main.o
...
[...]

PR #2055, opened 2021-08-17
Try building with go 1.17 and see what breaks

dkegel-fastly:
See https://github.com/tinygo-org/tinygo/issues/2054

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -377,6 +377,12 @@ jobs:
     steps:
       - test-linux:
           llvm: "11"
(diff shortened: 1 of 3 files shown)


aykevl:
Did you see #2048? I've made a PR for Go 1.17 already that (as far as I can see) passes all tests. It's only waiting on updated containers from CircleCI.

dkegel-fastly:
Ha, no, thanks!  I stumbled into this accidentally, should have checked for an existing pull request!

PR #2062, opened 2021-08-20
rp2040: add DormantUntil and begin work on IRQ API

soypat:
Begin adding interrupt API. Not tested.

Add the [`_goto_dormant`](https://github.com/raspberrypi/pico-extras/blob/ccd88d320f6d759d102cf65a1345d06f1d730f32/src/rp2_common/pico_sleep/sleep.c) interrupt.


soypat:
Pulled against release, sorry

PR #2063, opened 2021-08-20
machine/rp2040: Interrupts and add `machine.Sleep` API

soypat:
Begin adding interrupt API. ~Not tested.~ Tested and works! Example added under `examples/dormant` directory. Below is the program output of the example:
```
$ monitor
start program
going into dormant state until rising edge on pin 10
awoken!
```

Add the [`_goto_dormant`](https://github.com/raspberrypi/pico-extras/blob/ccd88d320f6d759d102cf65a1345d06f1d730f32/src/rp2_common/pico_sleep/sleep.c) interrupt.


ysoldak:
Hey, I've tested your branch with Nano RP2040, it does work here too! Had to add config for my board in example, obviously, see below.

nano-rp2040.go
```go
// +build nano_rp2040
package main

import "machine"

var (
	interruptPin = machine.D14
)

```

I've tried also improve the example, put it in cycle to see interrupt does trigger more than once. And here I have some problems with. Seem like board goes to sleep not in the exact moment you expect it do that, but few lines later. Can be optimisations, of course. I'll try and prepare some clean piece of code to debug and post here.

deadprogram:
@soypat can you please rebase `dev` into this branch? Thank you.

ysoldak:
@deadprogram I've rebased locally on upstream/dev to see how it works and if that fixes the problem I'm experiencing.

So, here is my streamlined and minimal variant of dormant.go example:
```go
package main

import (
	"machine"
	"time"
)

func main() {
	led := machine.LED
	interruptPin.Configure(machine.PinConfig{Mode: machine.PinInputPulldown})
	for {
		println("START")
		led.Low()
		println("  LOW")
		interruptPin.DormantUntil(machine.PinRising)
		println("  TOUCH")
		led.High()
		println("  HIGH")
		time.Sleep(5 * time.Second)
		println("END")
	}
}
```

I expect it to print "START", followed by "LOW" and hang there until I short D14 pin with 3v3. Then I shall print "TOUCH", "HIGH", wait for 5 sec keeping LED shining, print "END" and loop back to "START", "LOW" and go dormant again with LED off.
Instead, it goes dormant seem like after "HIGH" and keep LED on.
I understand, there can be buffering, so "LOW" actually not get printed when it should, but what about LED then?
I've tried insert some busy loops here and there too to fight possible buffering effect. Still the same. See output below.

```
12:32:29.662 -> START
12:32:29.662 ->   LOW
12:32:29.662 ->   TOUCH
12:32:29.662 ->   HIGH
12:32:41.217 -> END
12:32:41.217 -> START
12:32:41.217 ->   LOW
12:32:41.217 ->   TOUCH
12:32:41.217 ->   HIGH
12:32:58.141 -> END
12:32:58.141 -> START
12:32:58.141 ->   LOW
12:32:58.141 ->   TOUCH
12:32:58.141 ->   HIGH
```

soypat:
@deadprogram I'm not sure what you mean by rebasing this branch, it's already 1 commit ahead (I'm used to squash rebasing) what commands should I be executing? I've got quite the reputation over at gonum for fudging things up in the tree- don't want to bring that luggage here.

@ysoldak That's odd for sure. I'd suspect it does not happen without the for loop? The example I provided seemed to work fine.

ysoldak:
@soypat I believe he means
```shell
git rebase 4b1f92600f3f201923f67312aad3658fe5c06067 
```

ysoldak:
> That's odd for sure. I'd suspect it does not happen without the for loop? The example I provided seemed to work fine.

The very first time it halts as expected, after printing "LOW". I'll try to check w/o "for" loop, will just copy-paste couple of times and see.

ysoldak:
So, I've tried w/o loop, just doing it two times.

```go
package main

import (
	"machine"
	"time"
)

func main() {
	led := machine.LED
	interruptPin.Configure(machine.PinConfig{Mode: machine.PinInputPulldown})

	println("START1")
	led.Low()
	println("  LOW")
	interruptPin.DormantUntil(machine.PinRising)
	println("  TOUCH")
	led.High()
	println("  HIGH")
	time.Sleep(5 * time.Second)
	println("END1")

	println("START2")
	led.Low()
	println("  LOW")
	interruptPin.DormantUntil(machine.PinRising)
	println("  TOUCH")
	led.High()
	println("  HIGH")
	time.Sleep(5 * time.Second)
	println("END2")

	for {
		time.Sleep(time.Hour)
	}
}
```


Full output, from boot to last println:
```
14:18:57.916 -> START1
14:18:57.916 ->   LOW
14:20:40.962 ->   TOUCH
14:20:40.962 ->   HIGH
14:20:45.979 -> END1
14:20:45.979 -> START2
14:20:45.979 ->   LOW
14:20:45.979 ->   TOUCH
14:20:45.979 ->   HIGH
14:21:12.595 -> END2
```

aykevl:
To rebase, do the following.

Fetch the new version:

```
git checkout dev
git pull
```

Go back to your branch:

```
git checkout rp2040-dormant
```

Rebase on the dev branch:

```
git rebase dev
```

Push your changes:

```
git push -f
```

I hope that helps!

src/examples/dormant/pico.go
```diff
@@ -0,0 +1,16 @@
...
+package main
+
+import "machine"
+
+var (
+	interruptPin = machine.GP10
+)
+
+func init() {
+	err := machine.UART0.Configure(machine.UARTConfig{BaudRate: 9600})
```
aykevl: Why do you configure the UART like this? It's already configured by default.

src/machine/machine_rp2040.go
```diff
@@ -114,3 +114,8 @@ func init() {
 	UART0.Interrupt = interrupt.New(rp.IRQ_UART0_IRQ, _UART0.handleInterrupt)
 	UART1.Interrupt = interrupt.New(rp.IRQ_UART1_IRQ, _UART1.handleInterrupt)
 }
+
+// GetCoreNumber returns the core number the call was made from.
+func GetCoreNumber() uint32 {
```
aykevl: This is a new API, and should therefore have some thought how it should be designed. For example, the ESP32 also has two cores and may need an API like this.

My suggestion would be this:

```go
func CurrentCore() int
func NumCores() int
```

* No `Get` prefix, as is usual in Go (similar to `CPUFrequency()`)
* Return an `int`, which is a more common data type (and is the same size as `uint32` on most microcontrollers)
* Name the function differently. Initially I was a bit confused and thought that `GetCoreCount` meant the total count, not the current core.
* As a bonus: add a function to get the total number of cores.

aykevl (commented):
This PR mixes a few things together that feels a bit uncomfortable to me:

- It adds a new API for getting the current core number. Definitely useful, but not directly related to dormant state (except that it is also used internally).
- It has a partial implementation of pin change interrupts.
- It adds a new `DormantUntil` API, which appears very RP2040 specific.

My suggestion would be to complete the pin interrupt API (which would be useful even without dormant state!) and then if that's ready, add dormant state to it. For example, I can imagine that the API could be simplified to `machine.Sleep(...)` or something similar.

Regarding the DormantUntil API: I wonder if this can be generalized a bit? Many chips have a deep sleep state from which they can be awoken using a pin interrupt. I believe AVR chips have this, maybe others too.

ysoldak:
It's trickier than that when one's code in a fork.
Suppose you have two remotes: upstream (this repo) and origin (your repo).

What you need to do is:

Fetch this repo
```
git fetch upstream
```
Move your changes on top of this repo's dev branch
```
git rebase upstream/dev
```
Push you changes to your repo into rp2040-dormant branch
```
git push -f
```

aykevl:
It really depends on how the local git checkout is organized. If the dev branch has the TinyGo repo as its upstream, it would work. But not if the dev branch has the fork as its upstream. Your solution is perhaps more reliable though.

ysoldak:
>It really depends on how the local git checkout is organized. If the dev branch has the TinyGo repo as its upstream, it would work. But not if the dev branch has the fork as its upstream. Your solution is perhaps more reliable though.

Yes, this is what I mean. It can _look_ like everything is OK, but actually the branch is rebased on fork's dev branch and not upstream's.

soypat:
I have applied most changes. I have not, however, added the rest of the API as I am confused about how the C SDK maps over to tinygo. The way I understand it the C sdk allows setting only two interrupts, one for each core. The documentation mentions the following

> Each core is equipped with a standard ARM Nested Vectored Interrupt Controller (NVIC) which has 32 interrupt inputs. Each NVIC has the same interrupts routed to it, with the exception of the GPIO interrupts: there is one GPIO interrupt per bank, per core. These are completely independent, so e.g. core 0 can be interrupted by GPIO 0 in bank 0, and core 1 by GPIO 1 in the same bank.

I'm not really sure what this means (banks/cpu). I've read up on the compiler internalsbut still feel out of the loop on what exactly `interrupt.New` is doing under the hood.

 I've looked at other implementations in tinygo and found that `atsamd51.go` stores callbacks in a `[16]func(Pin)` array, which is stored in RAM as far as I know.

soypat:
PTAL. I have refactored the interrupt API and added functions to fulfill API with respect to other boards. I will be testing it shortly.

ysoldak:
@soypat check build logs
```
build/tinygo build -size short -o test.hex -target=nano-rp2040         examples/blinky1
# machine
/home/circleci/project/src/machine/machine_rp2040_gpio.go:305:3: undeclared name: GP2
```

Oh, I know. This smoke builds for Nano RP2040 Connect, and it has no `GP2`. You shall use `GPIOXX` to be compatible.

src/machine/machine_rp2040_gpio.go
```diff
@@ -208,3 +211,196 @@ func (p Pin) Set(value bool) {
...
+	}
+	// first acquire which interrupt has been triggered.
+	for i = 0; i < 4; i++ {
+		flags = irq.intS[i].Get()
+		if flags != 0 {
+			break
+		}
+	}
+	if flags == 0 {
+		GP2.acknowledgeInterrupt(1 << uint32(PinRising))
```
ysoldak: Is this GPIO2? But on Nano RP2040 Connect it's connected to Nina wifi module...

soypat: Whoops, my bad. Was doing some debugging on my pico and pushed my progress after an exhausting debug session. Will fix

soypat:
I'm closing this PR in favor of splitting it into 2 PRs, one containing the basic `pin.SetInterrupt()` API (#2215) and the other containing the pico's dormant `machine.Sleep()` API.

PR #2070, opened 2021-08-31
Minor changes to support go 1.17

fgsch:
I came across this while looking at something else. I believe this is needed but I have not tested it.

### DIFF
--- a/src/os/file_go_new.go
+++ b/src/os/file_go_new.go
@@ -1,4 +1,4 @@
-// +build go1.16
+// +build go1.16 go1.17
 
 package os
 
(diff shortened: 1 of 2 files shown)


fgsch:
cc @mathetake @sago35

mathetake (approved):

sago35 (approved):

deadprogram:
Thank you! Now merging.

aykevl:
I do not think this is necessary. A given Go version adds its own version and all previous versions as build tags. So for example, Go 1.16 adds build tags for `go1.1` all the way to `go1.16` (but not `go1.17` of course). So adding a build tag `go1.16` means "Go 1.16 and later".



> I believe this is needed but I have not tested it.

That's why you need to test ;)

fgsch:
Bugger! Thanks for the cluebat @aykevl and feel free to revert this.

PR #2072, opened 2021-08-31
targets: add openocd configuration for rp2040

sago35:
This PR adds flash and gdb via openocd to the target on rp2040.
For example, in the case of the Raspberry Pi Pico, you can write the following

Now there is no need to press BOOTSEL.

```
$ tinygo flash --target pico --size short --programmer picoprobe examples/blinky1

$ tinygo gdb --target pico --size short --programmer picoprobe examples/blinky1
```

### DIFF
--- a/targets/rp2040.json
+++ b/targets/rp2040.json
@@ -9,5 +9,7 @@
     "rp2040-boot-patch": true,
     "extra-files": [
         "src/device/rp/rp2040.s"
-    ]
+    ],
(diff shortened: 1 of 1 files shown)


sago35:
Here is a movie of an actual example that works.

https://twitter.com/sago35tk/status/1432669557546635270?s=20

deadprogram:
Thanks @sago35 now merging.

PR #2073, opened 2021-08-31
machine/feather-rp2040: add pin name definition for feather

sago35:
It is better to have a name as feather for better compatibility.
In this PR, the pin name as a feather will be added.

* https://learn.adafruit.com/adafruit-feather-rp2040-pico/pinouts
    * Analog Pins
    * Digital Pins


deadprogram:
Thanks @sago35 now merging.

PR #2074, opened 2021-08-31
nrf52840: fix ram size

sago35:
Since the nRF52840 has 256KB of RAM, it is necessary to modify the configuration.

https://learn.adafruit.com/introducing-the-adafruit-nrf52840-feather/hathach-memory-map

> [CFG] SoftDevice config requires more SRAM than provided by the linker.
> App Ram Start must be at least 0x20004180 (provided 0x20003200).
> Please update linker file or re-config SoftDevice.

https://www.nordicsemi.com/-/media/Software-and-other-downloads/Product-Briefs/nRF52840-SoC-product-brief.pdf?la=en&hash=EDF4C48A053E7943AD3C9DD3963B626D768B5885

> nRF52840 Product Brief Version 2.3
> KEY FEATURES
> 64 MHz Arm® Cortex-M4 with FPU
> 1 MB Flash + 256 KB RAM


aykevl:
This looks good to me, thank you! The chip indeed has a lot more RAM than was available with the old configuration.

PR #2077, opened 2021-09-02
Implement os.Getwd

fgsch:


src/syscall/syscall_libc_darwin.go
```diff
@@ -43,6 +43,7 @@ const (
 	EEXIST      Errno = 0x11
 	EINTR       Errno = 0x4
 	ENOTDIR     Errno = 0x14
+	EINVAL      Errno = 22
```
aykevl: Can you make this hexadecimal to match the other constants?

src/syscall/str.go
```diff
@@ -22,3 +22,13 @@ func uitoa(val uint) string {
 	buf[i] = byte(val + '0')
 	return string(buf[i:])
 }
+
+// clen returns the index of the first NULL byte in n or len(n) if n contains no NULL byte.
+func clen(n []byte) int {
```
aykevl, line 27: You could perhaps call this `strlen` (although I don't particularly care about the name as it is an implementation detail).

aykevl (commented):
Generally looks good to me.
You could perhaps add a test (in testdata/env.go for example) that the function returns no error and a non-empty string.

src/syscall/str.go
fgsch: I copied this from upstream so I'd prefer to keep it as is.

src/syscall/syscall_libc_darwin.go
fgsch: What if I covert them all to decimal? Having them in hexadecimal is a bit odd to me.

deadprogram: That sounds like it would be better done in a separate PR. For this PR, perhaps you could please make it a hex value like the others?

fgsch: Sure, updated.

fgsch:
I will add a test and update this space.

src/syscall/file_hosted.go
```diff
@@ -0,0 +1,25 @@
+// +build !baremetal,!wasi,!wasm
+
+// This file assumes there is a libc available that runs on a real operating
+// system.
+
+package syscall
+
+const pathMax = 1024
```
aykevl, line 8: Note that `PATH_MAX` doesn't really apply nowadays, see: https://insanecoding.blogspot.com/2007/11/pathmax-simply-isnt.html
But I guess it's good enough for now...

src/syscall/syscall_libc_darwin.go
aykevl: I don't mind which form it is, as long as it's consistent. Feel free to make a PR to convert these to decimal if you want to :)

src/syscall/file_hosted.go
fgsch: fwiw, this was copied from the standard library, and different platforms use different values there.

codefromthecrypt:
What do you think about using env instead of empty string for wasm?

WASI has env, someone could hint by adding the "PWD" variable and use it similar to how go does: https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/os/getwd.go#L33-L39

codefromthecrypt:
ignore above comment, I mistook `syscall.Getwd` for `os.Getwd`!

codefromthecrypt:
eek.. I forgot tinygo has its own `os.Getwd` so this is an issue. https://github.com/tinygo-org/tinygo/blob/cf0b7edc78e42038a0bb522b3f1a5b76928e730e/src/os/file.go#L248-L250

PR #2082, opened 2021-09-03
src/runtime: reset heapptr to heapStart after preinit()

dgryski:
heapptr is assinged to heapStart (which is 0) when it's declared, but preinit()
may have moved the heap somewhere else.  Set heapptr to the proper value
of heapStart when we initialize the heap properly.

This allows the leaking allocator to work on unix.


deadprogram:
Since support for Go 1.13/1.14 has been dropped for the upcoming release, I will ignore those CI failures.

Thank you very much for all of the research that has gone into this fix @dgryski 

Now merging.

PR #2084, opened 2021-09-04
tinygo: add a flag for creating cpu profiles

dgryski:


deadprogram:
Was there some change you wanted for this @aykevl or is it good to merge now?

dgryski:
I believe there was the idea of coming up with a naming convention or hidden set of flags that were useful for developers of tinygo instead of users of tinygo.  `-printir` would also fall into this category.

aykevl:
> I believe there was the idea of coming up with a naming convention or hidden set of flags that were useful for developers of tinygo instead of users of tinygo. `-printir` would also fall into this category.

Yes, I think that would be a good idea. But this seems useful on its own so I've just merged it.

PR #2056, opened 2021-08-17
machine/arduino_mkrwifi1010: add board definition for Arduino MKR Wifi 1010

sago35:
Added arduino-mkrwifi1010 (samd21g18) board support.
This board is very similar to arduino-nano33-iot with samd21 + NINA-W102 configuration.
It is also very similar to arduino-mkr1000.
This time, I checked the source code of both while working on it.

* Board
    * Name
        * Arduino MKR WiFi 1010
    * Board developer / designer / manufacturer
        * Arduino
    * schematic
        * https://store.arduino.cc/usa/mkr-wifi-1010
            * https://content.arduino.cc/assets/MKRWiFi1010V2.0_sch.pdf
    * pinout and stuff
        * https://store.arduino.cc/usa/mkr-wifi-1010
            * see `Documentation`
    * URL
        * https://store.arduino.cc/usa/mkr-wifi-1010
    * User manual for the board (if it exists)
        * https://www.arduino.cc/en/Guide/MKRWiFi1010
* Chip
    * chip name
        * samd21g18
    * Clock / ROM / RAM size
        * 48MHz / 256KB / 32KB
    * datasheet URL of the chip
        * https://content.arduino.cc/assets/mkr-microchip_samd21_family_full_datasheet-ds40001882d.pdf
* URL of the source, such as arduino (if it exists)
    * https://github.com/arduino/ArduinoCore-samd/blob/master/variants/mkrwifi1010/

At least the following examples have been confirmed.
I also checked the SPI with the code I created.

* GPIO: examples/blinky1
* UART: examples/echo
* I2C: drivers/examples/ssd1306/i2c_128x64
* WIFININA: drivers/examples/wifinina/webclient

PR #2086, opened 2021-09-05
machine/arduino_mkrwifi1010: fix pin definition of NINA_RESETN

sago35:
I made a mistake and corrected it.

https://github.com/arduino/ArduinoCore-samd/blob/master/variants/mkrwifi1010/variant.cpp#L155

refs: #2056


deadprogram:
Thank you @sago35 for the correction. Now merging.

PR #2087, opened 2021-09-05
esp32c3: add support for this chip

aykevl:
This change adds support for the ESP32-C3, a new chip from Espressif. It is a RISC-V core so porting was comparatively easy.

Most peripherals are shared with the (original) ESP32 chip, but with subtle differences. Also, the SVD file I've used gives some peripherals/registers a different name which makes sharing code harder. Eventually, when an official SVD file for the ESP32 is released, I expect that a lot of code can be shared between the two chips.

More information: https://www.espressif.com/en/products/socs/esp32-c3

TODO:
  - stack-based scheduler
  - interrupts (see https://github.com/espressif/svd/issues/6)
  - most peripherals (SPI, I2C, PWM, etc) - many can probably be shared with the original ESP32 chip

Also, the SVD file itself needs to be added to cmsis-svd. I'm waiting a bit until the last small issues with the SVD file are hopefully resolved.

Ping @igrr and @jessebraham, this might be of interest.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,6 +124,7 @@ build/gen-device-svd: ./tools/gen-device-svd/*.go
 
 gen-device-esp: build/gen-device-svd
 	./build/gen-device-svd -source=https://github.com/posborne/cmsis-svd/tree/master/data/Espressif-Community -interrupts=software lib/cmsis-svd/data/Espressif-Community/ src/device/esp/
+	./build/gen-device-svd -source=https://github.com/posborne/cmsis-svd/tree/master/data/Espressif -interrupts=software lib/cmsis-svd/data/Espressif/ src/device/esp/
 	GO111MODULE=off $(GO) fmt ./src/device/esp
 
 gen-device-nrf: build/gen-device-svd
@@ -432,6 +433,8 @@ ifneq ($(XTENSA), 0)
 	$(TINYGO) build -size short -o test.bin -target=nodemcu             examples/blinky1
 	@$(MD5SUM) test.bin
 endif
+	$(TINYGO) build -size short -o test.bin -target=esp32c3           	examples/serial
+	@$(MD5SUM) test.bin
 	$(TINYGO) build -size short -o test.hex -target=hifive1b            examples/blinky1
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=hifive1-qemu        examples/serial
--- a/src/device/esp/esp32c3.S
+++ b/src/device/esp/esp32c3.S
@@ -1,0 +1,49 @@
+// This is a very minimal bootloader for the ESP32-C3. It only initializes the
+// flash and then continues with the generic RISC-V initialization code, which
+// in turn will call runtime.main.
+// It is written in assembly (and not in a higher level language) to make sure
+// it is entirely loaded into IRAM and doesn't accidentally call functions
+// stored in IROM.
+//
+// For reference, here is a nice introduction into RISC-V assembly:
+// https://www.imperialviolet.org/2016/12/31/riscv.html
+
+.section .init
+.global call_start_cpu0
+.type call_start_cpu0,@function
+call_start_cpu0:
+    // At this point:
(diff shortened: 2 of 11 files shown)


src/device/esp/esp32c3.S
```diff
@@ -0,0 +1,49 @@
+// This is a very minimal bootloader for the ESP32-C3. It only initializes the
+// flash and then continues with the generic RISC-V initialization code, which
```
igrr, line 2: I see you have figured out cache initialization part already, but just in case, you might find this useful:
https://github.com/espressif/esp32c3-direct-boot-example

aykevl: Oh wow, I wasn't aware of that! That would have been quite useful!

For previous chips (ESP8266, ESP32), we basically load everything in RAM (loaded by the ROM bootloader) and just execute from there. I got that working on the ESP32-C3, but of course that's not ideal: there is only so much RAM available and it's kind of hard to make that work well with linker scripts. So I wanted to see whether I could get execution from flash working, ideally while keeping firmware size low. Luckily I managed to do that. It's a bit hacky, but it sort of works. Linker scripts are very much not made for the instruction/data bus split (with the same underlying address space) in the ESP32 family. I might make the same change for other chips in the family.

Oh, also, just as a side note, I think there is a bug in the ESP32-C3 reference manual (page 80):

> The CPU accesses the external memory via the cache. According to the MMU (Memory Management Unit)
settings, the cache maps the CPU’s address to the external memory’s physical address. Due to this address
mapping, **the ESP32-C3 can address up to 16 MB external flash**.
> 
> Using the cache, ESP32-C3 is able to support the following address space mappings. **Note that the instruction
bus address space (8MB) and the data bus address space (8 MB) is always shared.**

Emphasis mine. Those two statements appear to be contradictory. In my experiments, it appears that the address space is indeed shared and if something is mapped in the instruction bus address space, it's also mapped in the data bus address space at the same relative location. So I don't see how it's possible to access 16MB of flash that way. This 16MB is also repeated on page 76 ("Supports up to 16 MB external flash"), which also seems wrong.

aykevl:
Updated to the latest SVD file and updated the cmsis-svd repository to the esp32-svd branch (see https://github.com/tinygo-org/cmsis-svd/pull/9). I think this is now ready for review.

Makefile
```diff
@@ -124,6 +124,7 @@ build/gen-device-svd: ./tools/gen-device-svd/*.go
 
 gen-device-esp: build/gen-device-svd
 	./build/gen-device-svd -source=https://github.com/posborne/cmsis-svd/tree/master/data/Espressif-Community -interrupts=software lib/cmsis-svd/data/Espressif-Community/ src/device/esp/
+	./build/gen-device-svd -source=https://github.com/posborne/cmsis-svd/tree/master/data/Espressif -interrupts=software lib/cmsis-svd/data/Espressif/ src/device/esp/
```
zdima, line 127: the repository https://github.com/posborne/cmsis-svd/tree/master/data/Espressif not exists and  `make gen-device-esp` failing:
```$ make gen-device-esp 
./build/gen-device-svd -source=https://github.com/posborne/cmsis-svd/tree/master/data/Espressif-Community -interrupts=software lib/cmsis-svd/data/Espressif-Community/ src/device/esp/
lib/cmsis-svd/data/Espressif-Community/esp32.svd
lib/cmsis-svd/data/Espressif-Community/esp8266.svd
./build/gen-device-svd -source=https://github.com/posborne/cmsis-svd/tree/master/data/Espressif -interrupts=software lib/cmsis-svd/data/Espressif/ src/device/esp/
cannot find input directory: lib/cmsis-svd/data/Espressif/
make: *** [gen-device-esp] Error 1
```

aykevl: If you're testing this locally, you need to run `git submodule update --init`.

deadprogram:
Thank you for working on this @aykevl and thanks @igrr for review and @zdima for testing.

Now merging!

PR #2088, opened 2021-09-06
internal/task, runtime: add subsections_via_symbols to assembly files…

### LINKED ISSUE #2081: link errors on darwin when using -gc=leaking or -scheduler=coroutines
The default configuration includes assembly files which have link errors (due to undefined references) if you build with `-gc=leaking` or `-scheduler=coroutines`.

If using `leaking`, the included `gc_amd64.S` file expects an implementation of `scanstack`, which is only provided for `conservative` and `extalloc`.

Similarly, the `task_stack_amd64.S` file expects an implementation of `_task_pause`, which is a compiler intrinsic for `-scheduler=coroutines`.

https://github.com/tinygo-org/tinygo/blob/release/compileopts/target.go#L262

dgryski:
… on darwin

This allows the assembly routines in these files to be stripped as dead
code if they're not referenced.  This solves the link issues on MacOS
when the `leaking` garbage collector or the `coroutines` scheduler
are selected.

Fixes #2081

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -451,6 +451,9 @@ endif
 	@$(MD5SUM) test.nro
 	$(TINYGO) build -size short -o test.hex -target=pca10040 -opt=0     ./testdata/stdlib.go
 	@$(MD5SUM) test.hex
+ifneq ($(OS),Windows_NT)
+	$(TINYGO) build -o test.elf -gc=leaking -scheduler=none examples/serial
+endif
 
 
 wasmtest:
--- a/src/internal/task/task_stack_amd64.S
+++ b/src/internal/task/task_stack_amd64.S
@@ -72,3 +72,8 @@ tinygo_swapTask:
 
     // Return into the new task, as if tinygo_swapTask was a regular call.
     ret
+
+#ifdef __MACH__ // Darwin
+// allow these symbols to stripped as dead code
+.subsections_via_symbols
+#endif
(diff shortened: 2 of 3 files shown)


aykevl:
Looks good to me!

Can you also add a smoke test? At the bottom of the smoke tests in the Makefile, there is a section "test various compiler flags" where you could add:

```
$(TINYGO) build -size short -o test.elf -gc=leaking -scheduler=none examples/serial
```

This makes sure it continues to work in future versions and across all supported operating systems.

dgryski:
Huh, CI failed on MacOS ?

```
/usr/local/bin/tinygo build -size short -o test.elf -gc=leaking -scheduler=none examples/serial
error: bad magic number '[207 250 237 254]' in record at byte 0x0
```

aykevl:
Oh, right, the `-size short` flag should be removed. It's not (yet) supported on MacOS.

deadprogram:
@dgryski now a different build error, this time on Windows:

```
build/tinygo build -o test.elf -gc=leaking -scheduler=none examples/serial
error: Building Windows binaries is currently not supported. Try specifying a different target
mingw32-make: *** [Makefile:426: smoketest] Error 1
```

https://dev.azure.com/tinygo/tinygo/_build/results?buildId=3770&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=632617cc-49d1-5f82-37d2-600ea5832b67&l=411

deadprogram:
Thanks for the contribution @dgryski and thanks @aykevl for reviewing. Now merging.

PR #2091, opened 2021-09-07
Shorten some make targets

QuLogic:


QuLogic:
Annoyingly, it seems stretch is too old to have a reasonably new cmake.

Makefile
```diff
@@ -457,17 +460,26 @@ endif
 wasmtest:
 	$(GO) test ./tests/wasm
 
-build/release: tinygo gen-device wasi-libc
+build/compiler-rt.%.a: build/tinygo$(EXE)
+	./build/tinygo$(EXE) build-library -target=$* -o $@ compiler-rt
```
aykevl, line 467: Normally we use `$(TINYGO)` for this. Would that also work?

.circleci/config.yml
```diff
@@ -135,6 +135,8 @@ commands:
       - run:
           name: "Install apt dependencies"
           command: |
+            echo "deb http://deb.debian.org/debian stretch-backports main" | \
+                sudo tee /etc/apt/sources.list.d/backports.list
```
aykevl: Why do we need a new CMake version?

aykevl (commented):
Generally these cleanups look good, but I'm wondering about the new CMake version?

aykevl:
It looks like there is a problem with the new CMake version on Debian stretch.

There is also a failure on Windows:

```
mingw32-make: *** No rule to make target 'build/compiler-rt.armv6m-none-eabi.a', needed by 'library'.  Stop.
```

Makefile
QuLogic: Ah, I wrote this long ago and rebased; probably you've changed how this is done a bit. That'll likely work too.

.circleci/config.yml
QuLogic: The version on stretch didn't support `-B`/`-S`, though in older versions there was an undocumented `-H`, but I haven't tested that out on old or new versions. But it seems like I might need to install a bit more than a CMake backport, so I might revert that part.

Makefile
QuLogic: Actually, the [replaced lines](https://github.com/tinygo-org/tinygo/pull/2091/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L491-L496) are using the built version, so it seems like this should too?

QuLogic:
Fixing that failure on Windows depends on how you feel about the inline comment.

Makefile
aykevl: Right, I see. That makes sense. It looks good to me as-is.

aykevl:
This generally looks good, thanks for the cleanups! However, I think we should remain on stretch for now. Eventually we'll have to switch, at which point we'll automatically have the newer CMake version. (Also, note that no build failure is triggered now because we're caching the LLVM build - but the PR as is doesn't seem correct).
I would suggest undoing the change that requires the newer CMake version or waiting until we switch to buster or another distro.

Makefile
aykevl: Note that the `$(EXE)` is not necessary here: Windows will execute `./build/tinygo.exe` when calling `./build/tinygo`. It's not harmful either, though.

deadprogram:
Closing due to age of the PR and too many changes since then. Please open a new PR if these things are still needed. Thanks @QuLogic

PR #2094, opened 2021-09-08
compiler: fix equally named structs in different scopes

aykevl:
For example, in this code:

    type kv struct {
           v float32
    }

    func foo(a *kv) {
           type kv struct {
                   v byte
           }
    }

Both `kv` types would be given the same LLVM type, even though they are different types! This is fixed by only creating a LLVM type once per Go type (types.Type).

As an added bonus, this change gives a performance improvement of about 0.4%. Not that much, but certainly not nothing for such a small change.


deadprogram:
Thanks for this fix @aykevl now merging.

PR #2095, opened 2021-09-08
runtime: fix a suspicious bitwise operation

dgryski:
The `0 << nxp.SIM_CLKDIV1_OUTDIV1_Pos` term was duplicated.
No effect other than triggering a static analysis check.

### DIFF
--- a/src/runtime/runtime_nxpmk66f18.go
+++ b/src/runtime/runtime_nxpmk66f18.go
@@ -140,7 +140,7 @@ func initSystem() {
 
 	// now program the clock dividers
 	// config divisors: 180 MHz core, 60 MHz bus, 25.7 MHz flash, USB = IRC48M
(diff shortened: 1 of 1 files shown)


deadprogram:
Looking here seems like it might not even be needed? https://github.com/PaulStoffregen/cores/blob/master/teensy3/mk20dx128.c#L965

I suppose it is probably fine in the corrected version, unless you prefer to remove @dgryski

dgryski:
I have no opinion; I was just looking through results from static analysis tooling.

aykevl:
Interesting. It's shifting 0 so there should be no behavior difference.

Tagging @firelizzard18 who wrote this.

firelizzard18:
Definitely unintentional. @deadprogram linked the code that I (incorrectly) translated.

Since this is doing a write, the inclusion of `0 << pos` must be a noop - any bits that aren't explicitly set must end up zero, so all you actually need is `2 << pos2 | 6 << pos4`. But I like the symmetry of including all of the positions. I find it somewhat more readable when all positions are explicitly used.

deadprogram:
Thanks for the clarification @firelizzard18 

That said, now merging. Thanks @dgryski

PR #2096, opened 2021-09-08
builder: add missing error check for ioutil.TempFile()

dgryski:


deadprogram:
:heavy_check_mark: thanks @dgryski now merging.

Issue #2041, opened 2021-08-12
Crash with -target=wasi and -gc=leaking


ckaznocha:
I'm experiencing some runtime issues with tinygo produced wasi with the `leaking` gc. I think I've been able to distill the issue down to its core.

```
> ./tinygo version
tinygo version 0.20.0-dev-5e5ce98d darwin/amd64 (using go version go1.16.6 and LLVM version 11.0.0)
```

```
> wasmtime --version
wasmtime 0.29.0
```

Source code:
```Go
package main

import (
	"fmt"
	"sync"
)

var (
	once  sync.Once
	ready = make(chan struct{})
)

func main() {
	once.Do(func() {
		close(ready)
	})

	<-ready

	fmt.Println()
}
```

Expected output:
```
```

Actual output (modified to remove the full file paths):
```
> ./tinygo build -target=wasi -gc=leaking -o test.wasm && WASMTIME_BACKTRACE_DETAILS=1 wasmtime test.wasm
Error: failed to run main module `test.wasm`

Caused by:
    0: failed to invoke command default
    1: wasm trap: unreachable
       wasm backtrace:
           0: 0x4a87 - (*internal/task.Task).setReturnPtr
                           at {src_path}/github.com/tinygo-org/tinygo/src/internal/task/task_coroutine.go:51:4
                     - main.main
                           at {src_path}/github.com/fastly/ckaznocha/test/main.go:14:9
           1: 0x452f - (*internal/task.Task).Resume
                           at {src_path}/github.com/tinygo-org/tinygo/src/internal/task/task_coroutine.go:23:12
                     - runtime.scheduler
                           at {src_path}/github.com/tinygo-org/tinygo/src/runtime/scheduler.go:171:11
                     - runtime.run
                           at {src_path}/github.com/tinygo-org/tinygo/src/runtime/scheduler_any.go:24:11
                     - _start
                           at {src_path}/github.com/tinygo-org/tinygo/src/runtime/runtime_wasm_wasi.go:21:5

```

A few observations:

- The program runs correctly with `-opt=1`
- The program runs correctly with the `conservative` gc but that causes other issues for me.
- The program runs correctly if I change `fmt.Println()` to `log.Println()`

dgryski:
I've been looking into this with an execution trace from `wasm3-strace` and a decompilation of the wasm output from `wasm-decompile`.

The `unreachable` code is from this line:
```
function main_main_resume(a:int) {
[...]

PR #2100, opened 2021-09-08
compiler: avoid zero-sized alloca in channel operations

aykevl:
This works around a bug in LLVM (https://bugs.llvm.org/show_bug.cgi?id=49916) but seems like a good change in general.

Fixes https://github.com/tinygo-org/tinygo/issues/2041.

### DIFF
--- a/compiler/channel.go
+++ b/compiler/channel.go
@@ -32,8 +32,15 @@ func (b *builder) createChanSend(instr *ssa.Send) {
 
 	// store value-to-send
 	valueType := b.getLLVMType(instr.X.Type())
-	valueAlloca, valueAllocaCast, valueAllocaSize := b.createTemporaryAlloca(valueType, "chan.value")
-	b.CreateStore(chanValue, valueAlloca)
+	isZeroSize := b.targetData.TypeAllocSize(valueType) == 0
+	var valueAlloca, valueAllocaCast, valueAllocaSize llvm.Value
+	if isZeroSize {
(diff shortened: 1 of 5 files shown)


dgryski:
Are there any other places that zero-byte structs will be affected with a similar problem?

deadprogram:
@aykevl not sure if you noticed that all of the CI builds are failing: https://app.circleci.com/pipelines/github/tinygo-org/tinygo/5375/workflows/4ce07cea-c635-4dee-8075-6546395e080c/jobs/31173?invite=true#step-117-660

aykevl:
> @aykevl not sure if you noticed that all of the CI builds are failing

Oops, yeah I just read the message from CircleCI. It should e fixed now.

aykevl:
> Are there any other places that zero-byte structs will be affected with a similar problem?

Yes. Some places have a zero sized alloca. This might be the last one remaining:

https://github.com/tinygo-org/tinygo/blob/485a9284e74ea9ff8f0f37f2402ae80448b33f10/compiler/compiler.go#L1533-L1537

If you want to catch them, you can build LLVM using `ASSERTS=1` on the `make` command line (something like `make llvm-build ASSERTS=1`) and then build TinyGo using it.
I'm not sure I'm going to bother fixing this, unless it results in actual bugs: it will be fixed when we upgrade to LLVM 12 anyway. So I'd rather focus now on getting LLVM 12 to work properly (and after that, LLVM 13).

deadprogram:
:no_entry_sign:  :zero: 

Thanks @aykevl now merging.

PR #2102, opened 2021-09-08
cgo: don't normalize CGo tests anymore

aykevl:
Normalization was required because previously we supported Go 1.13 and Go 1.14 at the same time. Now we've dropped support for both so this normalization is not necessary anymore.

CGo support remains the same. It's just the test outputs that aren't normalized anymore.

(This is just to remove some cruft to keep the code base clean)


deadprogram:
Thanks for the cleanup here @aykevl now merging.

PR #2103, opened 2021-09-08
targets: add DefaultUART to adafruit boards

sago35:
Added the required configuration to #1880 for the adafruit board.

The following boards, which have not been changed in this PR, are already configured.

* feather-nrf52840-sense
* feather-nrf52840
* feather-rp2040
* feather-stm32f405

The recently added mdbt50qrx-uf2 does not have a port and does not need to be configured.

### DIFF
--- a/src/machine/board_circuitplay_bluefruit.go
+++ b/src/machine/board_circuitplay_bluefruit.go
@@ -83,3 +83,7 @@ var (
 	usb_VID uint16 = 0x239A
 	usb_PID uint16 = 0x8045
 )
+
+var (
+	DefaultUART = UART0
+)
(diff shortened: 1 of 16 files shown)


sago35:
The case with tags is failing.
DefaultUART setting should be done in *_baremetal.go if it exists.

```
$ tinygo build             -o /tmp/test.wasm -tags=circuitplay_express  examples/blinky1
# machine
**\src\machine\board_circuitplay_express.go:105:16: undeclared name: UART1
```

sago35:
Since pygamer did not have a UART setting, it was removed.
The setting of metro-m4-airlift was wrong and has been corrected.
rebase.

I think the CI will be successful now.

deadprogram:
Thanks for catching this @sago35 now merging.

PR #2104, opened 2021-09-08
target: fix pid for mdbt50qrx-uf2

sago35:
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/5e72fd0a1dbd2ecce0b2a00b139e4b102a240b74/boards.txt#L505-L511


deadprogram:
Thank you for the correction @sago35 now merging.

PR #2107, opened 2021-09-11
teensy40: Enable receive watermark interrupt (Fixes #2092)

### LINKED ISSUE #2092: Teensy 4.0's machine.UART.handleInterrupt() does not work
@ardnew 
The program in examples/echo does not work.
It seems that `UART.handleInterrupt()` is not working.

The UART transmission by fmt.Printf() seems to be working.

ardnew:
This PR fixes a few problems that were apparent while debugging #2092, and it also simplifies the `(*UART).Configure` code a bit.

The most important change was to enable the receive and idle line interrupts (`LPUART_CTRL_RIE`, `LPUART_CTRL_ILIE`) during the config sequence (not sure if/how this was even working before?)

Some smaller periphery updates were also made as they generally improve the Teensy 4 package and do affect the stability/performance of UART:

- Move local registers of `enableDcache(bool)` to package scope (see: `dcacheCcsidr`, `dcacheSets`, and `dcacheWays` in `src/device/nxp/mimxrt1062_mpu.go`)
	- This eliminates a potentially-very-frequent heap allocation, as `enableDcache(bool)` is intended to be called before each DMA beat (I think that's the right term?)
- Reduce TinyGo stack size defined in linker script (`targets/mimxrt1062-teensy40.ld`) from an astronomically unreasonable `128K` to a conventional `4K`.
	- This frees up `124K` of DTCM RAM for application code. Enough said.

### DIFF
--- a/src/device/nxp/mimxrt1062_mpu.go
+++ b/src/device/nxp/mimxrt1062_mpu.go
@@ -2,6 +2,7 @@
 // Type definitions, fields, and constants associated with the MPU peripheral
 // of the NXP MIMXRT1062.
 
+//go:build nxp && mimxrt1062
 // +build nxp,mimxrt1062
 
 package nxp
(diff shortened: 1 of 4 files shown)


sago35 (approved):
The problem in #2092 has been resolved and examples/echo is now working successfully.
I think the code is also good.

deadprogram:
Thanks @ardnew for the fixes/improvments, as well as to @sago35 for review/testing.

Now merging.

PR #2111, opened 2021-09-15
unix: check for mmap error and act accordingly

aykevl:
At startup, a large chunk of virtual memory is used up by the heap. This
works fine in emulation (qemu-arm), but doesn't work so well on an
actual Raspberry Pi. Therefore, this commit reduces the requested amount
until a heap size is found that works on the system.

This can certainly be improved, but for now it's an important fix
because it allows TinyGo built binaries to actually run on a Raspberry
Pi with just 1GB RAM. Previously, they would segfault.


deadprogram:
This will be quite helpful @aykevl thanks for the improvement. :pie:

PR #2112, opened 2021-09-15
arm: switch to Thumb instruction set on ARM

aykevl:
This reduces binary size substantially, for two reasons:

  - It switches to a much more modern architecture: ARMv4 vs ARMv7.
  - It switches to Thumb2, which is a lot denser than regular ARM.

Practically all modern and not-so-modern ARM chips support Thumb2, so this seems like a safe change to me.

The size in numbers:

  - Code size for testdata/stdlib.go is reduced by about 35%.
  - Binary size for testdata/stdlib.go (when compiling with -no-debug to strip debug information) is reduced by about 16%.


deadprogram:
Good one @aykevl :+1: merging.

PR #2114, opened 2021-09-15
os: add SEEK_SET, SEEK_CUR, and SEEK_END

aykevl:
These are deprecated but still used by some code.


deadprogram:
Thanks @aykevl now merging.

Issue #2101, opened 2021-09-08
Memory corruption with scheduler=coroutines


dgryski:
```
package main

import (
        "context"
        "os"
)

func main() {
        flog := os.Stdout
        println("outside flog is", flog)

        ServeFunc(func(context context.Context) {
                println("inside flog is: ", flog)
        })
}

func Serve(h Handler) {
        _, cancel := context.WithCancel(context.Background())
        defer cancel()
        h.ServeHTTP(nil)
}

func ServeFunc(f HandlerFunc) {
        Serve(f)
}

type Handler interface {
        ServeHTTP(ctx context.Context)
}

type HandlerFunc func(ctx context.Context)

func (f HandlerFunc) ServeHTTP(ctx context.Context) {
        f(ctx)
}
```

Produces the following output:

```
outside flog is 0x00010704
panic: runtime error: nil pointer dereference
Error: failed to run main module `m.wasm`

Caused by:
    0: failed to invoke command default
    1: wasm trap: unreachable
       wasm backtrace:
           0: 0x2e1e - runtime.abort
                           at /Users/dgryski/go/src/tinygo.org/tinygo/src/runtime/runtime_tinygowasm.go:63:6
                     - runtime.runtimePanic
                           at /Users/dgryski/go/src/tinygo.org/tinygo/src/runtime/panic.go:20:7
           1:  0xc65 - runtime.nilPanic
                           at /Users/dgryski/go/src/tinygo.org/tinygo/src/runtime/panic.go:32:14
           2: 0xb9ae - main.Serve
                           at /Users/dgryski/go/src/github.com/dgryski/wasmplay/cmd/closenil/main.go
           3: 0x51f5 - (*internal/task.Task).Resume
                           at /Users/dgryski/go/src/tinygo.org/tinygo/src/internal/task/task_coroutine.go:23:12
                     - runtime.scheduler
                           at /Users/dgryski/go/src/tinygo.org/tinygo/src/runtime/scheduler.go:171:11
                     - runtime.run
                           at /Users/dgryski/go/src/tinygo.org/tinygo/src/runtime/scheduler_any.go:24:11
                     - _start
                           at /Users/dgryski/go/src/tinygo.org/tinygo/src/runtime/runtime_wasm_wasi.go:21:5

```

This was run with tinygo build from `dev` and the patch from #2100 .

I've tried to shrink this down as much as I can.  The `println` inside the closure generates a nil-pointer crash when trying to restore `flog` from the coroutine context.

[...]

PR #2117, opened 2021-09-15
transform (coroutines): fix memory corruption for tail calls that reference stack allocations

niaow:
This change fixes a bug in which `alloca` memory lifetimes would not extend past the suspend of an asynchronous tail call. This would typically manifest as memory corruption, and could happen with or without normal suspending calls within the function.

Fixes a bug from #2101

### DIFF
--- a/transform/coroutines.go
+++ b/transform/coroutines.go
@@ -600,11 +600,11 @@ func (c *coroutineLoweringPass) lowerFuncsPass() {
 			continue
 		}
 
-		if len(fn.normalCalls) == 0 {
-			// No suspend points. Lower without turning it into a coroutine.
+		if len(fn.normalCalls) == 0 && fn.fn.FirstBasicBlock().FirstInstruction().IsAAllocaInst().IsNil() {
+			// No suspend points or stack allocations. Lower without turning it into a coroutine.
 			c.lowerFuncFast(fn)
 		} else {
-			// There are suspend points, so it is necessary to turn this into a coroutine.
+			// There are suspend points or stack allocations, so it is necessary to turn this into a coroutine.
 			c.lowerFuncCoro(fn)
 		}
 	}
@@ -827,6 +827,7 @@ func (c *coroutineLoweringPass) lowerFuncCoro(fn *asyncFunc) {
 	}
 
 	// Lower returns.
+	var postTail llvm.BasicBlock
 	for _, ret := range fn.returns {
 		// Get terminator instruction.
 		terminator := ret.block.LastInstruction()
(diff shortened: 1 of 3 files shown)


transform/coroutines.go
```diff
@@ -600,11 +600,11 @@ func (c *coroutineLoweringPass) lowerFuncsPass() {
 			continue
 		}
 
-		if len(fn.normalCalls) == 0 {
-			// No suspend points. Lower without turning it into a coroutine.
+		if len(fn.normalCalls) == 0 && fn.fn.FirstBasicBlock().FirstInstruction().IsAAllocaInst().IsNil() {
```
aykevl, line 603: I don't think alloca instructions are always at the start of the entry block. I think it would be safer to check the entire entry block for alloca instructions, just in case some are not the first instruction.

(Technically they can be anywhere in the function but we check in other places that this isn't possible).

transform/coroutines.go
```diff
@@ -886,10 +887,37 @@ func (c *coroutineLoweringPass) lowerFuncCoro(fn *asyncFunc) {
 			call.EraseFromParentAsInstruction()
 		}
 
-		// Replace terminator with branch to cleanup.
+		// Replace terminator with a branch to the exit.
+		var exit llvm.BasicBlock
+		if ret.kind == returnNormal || ret.kind == returnVoid || fn.fn.FirstBasicBlock().FirstInstruction().IsAAllocaInst().IsNil() {
```
aykevl, line 913: Same here.

aykevl (commented):
Looks good to me, apart from those two comments.

transform/coroutines.go
niaow: If there are any alloca instructions that are not at the start of the entry block, the coroutine lowering pass crashes iirc.

aykevl: IIRC it crashes if they are not in the entry block. but I think they are allowed if they are not directly at the start of the entry block.

deadprogram:
@niaow any chance of addressing this feedback from @aykevl? I really really want to merge this! :smiley_cat:

niaow:
I added a pre-processing step which ensures that all allocas in the entry block exist at the start. I had previously made sure that this never happened because a while back we hit an issue where a basic block split moved some allocas out of the entry block.

deadprogram:
That seems like a good solution to me, @niaow 

@aykevl are we good to merge?

transform/coroutines.go
```diff
@@ -763,6 +763,27 @@ func (c *coroutineLoweringPass) lowerCallReturn(caller *asyncFunc, call llvm.Val
...
 // See https://llvm.org/docs/Coroutines.html for more information on these intrinsics.
 func (c *coroutineLoweringPass) lowerFuncCoro(fn *asyncFunc) {
+	// Ensure that any alloca instructions in the entry block are at the start.
+	// Otherwise, block splitting would result in unintended behavior.
+	{
+		// Skip alloca instructions at the start of the block.
+		inst := fn.fn.FirstBasicBlock().FirstInstruction()
+		for !inst.IsAAllocaInst().IsNil() {
+			inst = llvm.NextInstruction(inst)
+		}
```
aykevl, line 773: I believe this won't work correctly if there are no alloca instructions in the entry block: it will eventually reach the last instruction in the block and crash on a nil pointer dereference.

transform/coroutines.go
```diff
@@ -763,6 +763,27 @@ func (c *coroutineLoweringPass) lowerCallReturn(caller *asyncFunc, call llvm.Val
...
+		// Find any other alloca instructions and move them after the other allocas.
+		c.builder.SetInsertPointBefore(inst)
+		for !inst.IsNil() {
+			next := llvm.NextInstruction(inst)
+			if !inst.IsAAllocaInst().IsNil() {
+				inst.RemoveFromParentAsInstruction()
+				c.builder.Insert(inst)
+			}
+			inst = next
+		}
```
aykevl, line 784: This moves alloca instructions together, which I guess is fine, but I don't see how this addresses the issue I raised before?

> I don't think alloca instructions are always at the start of the entry block. I think it would be safer to check the entire entry block for alloca instructions, just in case some are not the first instruction.

transform/coroutines.go
niaow: The loop condition is the other way around. If there are no alloca instructions this will run 0 times.

transform/coroutines.go
niaow: This moves all alloca instructions to the start of the entry block, so later the check does correctly apply to all allocas in the entry block. This also fixes the potential issues where an alloca could be moved by a later transformation into another block by `SplitBasicBlock`.

transform/coroutines.go
aykevl: Oh, I see. I got confused by the double negation.

aykevl (approved):
Ok, seems good to me!

Issue #2036, opened 2021-08-07
[esp32] SPI not working as expected


Nebdir:
my setup:
esp32 [link](https://www.az-delivery.de/en/products/esp32-developmentboard)
waveshare 1.5 inch oled display [link](https://www.waveshare.com/1.5inch-oled-module.htm)
goland ide with tinygo plugin
tinygo 0.19.0

tl;dr;
display is working with esp, arduino spi and arduino ide no additional config, but not with default tinygo esp32 spi. soft spi with tinygo is working. docs missread? implementation wrong? board not expected for implementation?

my story in all the details of the last 24h:

ive tried to get a simple program working to control a 1.5 inch oled display from waveshare.

for that ive tried this first using arduino and its spi library.

```
#include <SPI.h>
void setup() {
  pinMode(17, OUTPUT); // CS
  pinMode(16, OUTPUT); // RST
  pinMode(4, OUTPUT); // DC
  SPI.begin();
  digitalWrite(16,HIGH);
  delay(100);
  digitalWrite(16,LOW);
  delay(100);
  digitalWrite(16,HIGH);
  delay(100);
  SPI.transfer(0xaf);
  delay(300);
  SPI.transfer(0xa5);
}
void loop() {}
```

that works as intended and resets enables and fills the display with white pixels.

but trying that with tinygo doesnt work as intended, 

```
package main

import (
	"machine"
	"time"
)

func main() {
	dc := machine.Pin(4)
	dc.Configure(machine.PinConfig{Mode: machine.PinOutput})
	rst := machine.Pin(16)
	rst.Configure(machine.PinConfig{Mode: machine.PinOutput})
	cs := machine.Pin(17)
	cs.Configure(machine.PinConfig{Mode: machine.PinOutput})
	dc.Low()
	cs.Low()
	rst.High()
	time.Sleep(100 * time.Millisecond)
	rst.Low()
	time.Sleep(100 * time.Millisecond)
	rst.High()
	time.Sleep(100 * time.Millisecond)
	machine.SPI2.Transfer(0xaf)
	time.Sleep(300 * time.Millisecond)
	machine.SPI2.Transfer(0xa5)
}
```
ive tried to find the difference between those two implementations in the spi protcol. ive done all testing with the same wiring.
ive experimented with all mentioned frequencies and modes, no good results. so ive decided to build a very slow software spi emulation at the bare metal of the protocol:

```
package main

import (
	"machine"
	"time"
)

func main() {
[...]

PR #2119, opened 2021-09-16
esp32: fix spi extra configuration

sago35:
This PR corrects the timing of SCK and SDO.
This PR fixes the problem in #2036.

I am not sure if this change is correct, but I believe it is at this point for the following reasons.
If there is a better way to change it, please let me know.

### 1 ###
USER_USR_DUMMY is not used in espressif/arduino-esp32.
If USER_USR_DUMMY is set, 0x00 will be inserted at the beginning of SPI communication.
This appears to be unnecessary.

### 2 ###
MOSI_DELAY_MODE is also used only in the slave side below.

https://github.com/espressif/arduino-esp32/blob/master/tools/sdk/esp32/include/hal/esp32/include/hal/spi_ll.h#L370-L425

### 3 ###
`table 29 on page 128` is not found.
The following PDF does not appear to require the MOSI_DELAY_MODE setting.

https://www.espressif.com/sites/default/files/documentation/esp32_technical_reference_manual_en.pdf

image

### 4 ###
The actual waveform before and after the modification is shown below.

old w/40e6 Hz - ili9341 did not work
Depending on the USER_USR_DUMMY setting, an extra 0x00 will be output
image

old w/20e6 Hz - ili9341 worked
image

old w/4e6 Hz - ili9341 did not work
The timing of SCK and SDO is almost simultaneous due to the MOSI_DELAY setting.
image

new w/40e6 Hz - ili9341 did not work (note-1)
image

new w/20e6 Hz - ili9341 worked
image

new w/4e6 Hz - ili9341 worked
image

note-1:
As a separate issue, I think the CLOCK register setting needs to be fixed. After changing the code a bit, I got the correct waveform output even at 40Mhz. I am going to create another PR.

### DIFF
--- a/src/machine/machine_esp32.go
+++ b/src/machine/machine_esp32.go
@@ -384,16 +384,17 @@ func (spi SPI) Configure(config SPIConfig) error {
 	if config.Frequency >= 40e6 {
 		// Delay mode must be set to 0 and SPI_USR_DUMMY_CYCLELEN should be set
 		// to 0 (the default).
+		spi.Bus.USER1.Set(0)
 		userReg |= esp.SPI_USER_USR_DUMMY
 	} else if config.Frequency >= 20e6 {
 		// Nothing to do here, delay mode should be set to 0 according to the
 		// datasheet.
 	} else {
-		// Follow the delay mode as given in table 29 on page 128 of the
+		// Follow the delay mode as given in table 7-3 on page 125 of the
 		// reference manual.
 		// Note that this is only specified for SPI frequency of 10MHz and
 		// below (≤Fapb/8), so 13.3MHz appears to be left unspecified.
-		ctrl2Reg |= delayMode << esp.SPI_CTRL2_MOSI_DELAY_MODE_Pos
+		ctrl2Reg |= delayMode << esp.SPI_CTRL2_MISO_DELAY_MODE_Pos
 	}
 	// Enable full-duplex communication.
 	userReg |= esp.SPI_USER_DOUTDIN


sago35:
The ESP32 SPI was created in #1379.
I would like to know if you have any comments on this change. @aykevl

sago35:
For reference, the waveform of atsamd51 looks like this 
Since it is in Mode0, SCK is sampled on the rising edge.

image

deadprogram:
@aykevl any comments before this gets merged?

src/machine/machine_esp32.go
```diff
@@ -363,38 +363,16 @@ func (spi SPI) Configure(config SPIConfig) error {
...
-		userReg |= esp.SPI_USER_USR_DUMMY
-	} else if config.Frequency >= 20e6 {
-		// Nothing to do here, delay mode should be set to 0 according to the
-		// datasheet.
-	} else {
-		// Follow the delay mode as given in table 29 on page 128 of the
-		// reference manual.
-		// Note that this is only specified for SPI frequency of 10MHz and
-		// below (≤Fapb/8), so 13.3MHz appears to be left unspecified.
-		ctrl2Reg |= delayMode << esp.SPI_CTRL2_MOSI_DELAY_MODE_Pos
```
aykevl, line 396: I think the mistake is that I did this to the MOSI pin, while it should have been used for MOSI.

src/machine/machine_esp32.go
```diff
@@ -363,38 +363,16 @@ func (spi SPI) Configure(config SPIConfig) error {
...
-		delayMode = 2
 		pinReg |= esp.SPI_PIN_CK_IDLE_EDGE
 	}
-	// Extra configuration necessary for correct data input at high frequencies.
-	// This is only necessary when MISO goes through the GPIO matrix (which it
-	// currently does).
-	if config.Frequency >= 40e6 {
-		// Delay mode must be set to 0 and SPI_USR_DUMMY_CYCLELEN should be set
-		// to 0 (the default).
-		userReg |= esp.SPI_USER_USR_DUMMY
```
aykevl: From the datasheet:

> If GP-SPI output clock frequency is clkapb/2, register SPI_MISO_DELAY_MODE should be set to 0 and the dummy phase should be enabled (SPI_USR_DUMMY = 1) for one clkspi clock cycle (SPI_USR_DUMMY_CYCLELEN = 0) when configuring the clock polarity;

So the datasheet says the `SPI_USER_USR_DUMMY` bit must be set.

I wonder if the datasheet is wrong or something else is going on?

src/machine/machine_esp32.go
```diff
@@ -363,38 +363,16 @@ func (spi SPI) Configure(config SPIConfig) error {
 	// (mode), among others.
 	var ctrl2Reg, userReg, pinReg uint32
 	// For mode configuration, see table 29 in the reference manual (page 128).
```
aykevl: I think I was referring to an older version of the TRM. Please see table 7-3 on page 125 of https://www.espressif.com/sites/default/files/documentation/esp32_technical_reference_manual_en.pdf.

aykevl (commented):
Can you try whether SPI works correctly when you apply this small patch?

```diff
-		ctrl2Reg |= delayMode << esp.SPI_CTRL2_MOSI_DELAY_MODE_Pos
+		ctrl2Reg |= delayMode << esp.SPI_CTRL2_MISO_DELAY_MODE_Pos
```

I think that is what I originally intended and I think it is needed so that MISO (reading data, SDI) works correctly.

src/machine/machine_esp32.go
sago35: If I change it to the following, it seems to work fine.

```go
ctrl2Reg |= delayMode << esp.SPI_CTRL2_MISO_DELAY_MODE_Pos
```

image

sago35: It was at 4MHz that I thought it was working.
At 4Mhz, the read succeeds regardless of the MOSI_DELAY_MODE setting.
However, when MOSI_DELAY_MODE is set at 10 MHz and 13.3 MHz, as shown below, it does not read properly.

image

The expected value is 0xE8, but if it doesn't read properly, it will read 0xF4, a value that has been shifted by one bit.

```
MISO_DELAY_MODE
    40   MHz: F4
    20   MHz: F4
    13.3 MHz: F4 !!
    10   MHz: F4 !!
     8   MHz: E8
     6.5 MHz: E8

NO DELAY
    40   MHz: F4
    20   MHz: F4
    13.3 MHz: E8
    10   MHz: E8
     8   MHz: E8
     6.5 MHz: E8
```

It still looks like it is better not to set MOSI_DELAY_MODE.
In the following source, MOSI_DELAY_MODE is not used.

The DELAY_MODE is not used except in the selected slave configuration.

https://github.com/espressif/arduino-esp32/blob/master/tools/sdk/esp32/include/hal/esp32/include/hal/spi_ll.h#L370-L425

src/machine/machine_esp32.go
sago35:
```go
		// Delay mode must be set to 0 and SPI_USR_DUMMY_CYCLELEN should be set
		// to 0 (the default).
```

image


The source code says that the initial value of SPI_USR_DUMMY_CYCLELEN is 0.
However, in TRM, ResetValue=7.
By adding the following initialization, the dummy cycle is no longer output to the bus.

```go
spi.Bus.USER1.Set(0)
```

src/machine/machine_esp32.go
sago35: Updated comments.

src/machine/machine_esp32.go
sago35: Added initialization of SPI_USR_DUMMY_CYCLELEN.

https://github.com/tinygo-org/tinygo/blob/86a4c71ebbe07d4399520589d039951edf0eada2/src/machine/machine_esp32.go#L387

aykevl:
> The source code says that the initial value of SPI_USR_DUMMY_CYCLELEN is 0.
> However, in TRM, ResetValue=7.

You are correct! Yes, that's a mistake. That means that you need `spi.Bus.USER1.Set(0)` if `SPI_USER_USR_DUMMY` is set.

Is it now correct with `spi.Bus.USER1.Set(0)` to correct this mistake? Or do you still need to disable `SPI_USER_USR_DUMMY` for it to work?

sago35:
As a result of checking, if USR_DUMMY is present, the communication fails because it seems to be sending one bit too many.
At this point, I think it is better to remove the `USER1.Set(0) + USR_DUMMY` .

```go
		spi.Bus.USER1.Set(0)
		userReg |= esp.SPI_USER_USR_DUMMY
```

40MHz + w/o USER1.Set(0) + w/o USR_DUMMY : OK
image

40MHz + w/o USER1.Set(0) + with USR_DUMMY (== dev branch) : NG
image

40MHz + with USER1.Set(0) + with USR_DUMMY : NG
Communication failure due to underlined 1 bit too many
image

80MHz + w/o USER1.Set(0) + w/o USR_DUMMY : OK
image

80MHz + with USER1.Set(0) + with USR_DUMMY : NG
Communication failure due to underlined 1 bit too many
image

sago35:
In order to run at 40Mhz, the Frequency needs to be set to 40e6, and the condition needs to be changed as follows.
It is difficult to read from the TRM, but it seems that freq > 40e6 is probably correct.

image

deadprogram:
Any more feedback on this @aykevl or is it now good to merge?

sago35:
I think the first version is correct.
In the first version, the following parts were all removed.

https://github.com/tinygo-org/tinygo/blob/86a4c71ebbe07d4399520589d039951edf0eada2/src/machine/machine_esp32.go#L381-L398

----

Also, I think it would be better to incorporate the following changes.
If you add the following changes, the upper limit of the SPI clock will be increased from 20Mhz to 40MHz.

https://github.com/tinygo-org/tinygo/pull/2119#issuecomment-926253226

sago35:
Newly created #2148 .
I believe #2148 is correct, not this PR.

aykevl:
Closing because #2148 has been merged.

PR #2120, opened 2021-09-16
stm32: add support for PortMask* functions for WS2812 support

aykevl:
This also requires support in the tinygo.org/x/drivers/ws2812 package, which I've already partially written: https://github.com/tinygo-org/drivers/pull/314


deadprogram:
Very cool, thanks @aykevl :diving_mask:

PR #2122, opened 2021-09-16
builder: simplify running of jobs

aykevl:
Instead of keeping a slice of jobs to run, let the runJobs function
determine which jobs should be run by investigating all dependencies.
This has two benefits:

  - The code is somewhat cleaner, as no 'jobs' slice needs to be
    maintained while constructing the dependency graph.
  - Eventually, some jobs might not be required by any dependency.
    While it's possible to avoid adding them to the slice, the simpler
    solution is to build a new slice from the dependencies which will
    only include required dependencies by design.

---

Note: this is a change that I've made at least twice now, but never in a finished branch yet. So I decided to extract it so that future changes to the builder package will be easier and result in fewer conflicts.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -124,11 +124,6 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 		return err
 	}
 
-	// The slice of jobs that orchestrates most of the build.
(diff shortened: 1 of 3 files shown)


deadprogram:
Good "job" @aykevl now merging.

PR #2123, opened 2021-09-17
add support for esp32c3 on-chip random generator to crypto/rand package.

zdima:
The ESP32-C3 contains a true random number generator. This will allow utilize this for cryptographical operations, among other things.

### DIFF
--- a/src/crypto/rand/rand_esp32c3.go
+++ b/src/crypto/rand/rand_esp32c3.go
@@ -16,7 +16,7 @@ import (
 const MHZ = 1000000
 
 func init() {
-	Reader = &reader{}
+	Reader = &esp32c3RndReader{}
 
 	// When using the random number generator, make sure at least either the SAR ADC,
 	// high-speed ADC1, or RTC20M_CLK2 is enabled. Otherwise, pseudo-random numbers will be returned.
@@ -61,8 +61,9 @@ func init() {
 
 }
 
-type reader struct {
-	lastCpuTick uint32
+type esp32c3RndReader struct {
+	lastCpuTick     uint32
+	minimalCPUTicks uint32
 }
 
 func getApbFreqHZ() uint32 {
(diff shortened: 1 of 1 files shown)


src/crypto/rand/rand_esp32c3.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+type reader struct {
+}
+
+func (r *reader) Read(b []byte) (n int, err error) {
+	if len(b) != 0 {
+		for i := 0; i < len(b); {
+			r := esp.APB_CTRL.RND_DATA.Get()
+			a := (*[4]byte)(unsafe.Pointer(&r))[:]
```
aykevl: This duplicates the `r` variable, which is a bit confusing. Can you rename the `r` variable (and possibly other single letter variables) to something more readable?

src/crypto/rand/rand_esp32c3.go
```diff
@@ -0,0 +1,59 @@
...
+	// Note:
+	//  1. Note that, when the Wi-Fi module is enabled, the value read from the high-speed
+	//     ADC can be saturated in some extreme cases, which lowers the entropy. Thus, it
+	//     is advisable to also enable the SAR ADC as the noise source for the random
+	//     number generator for such cases.
+	//  2. Enabling RTC20M_CLK increases the RNG entropy. However, to ensure maximum entropy,
+	//     it’s recommended to always enable an ADC source as well.
+
+	// Enable SAR ADC
+	esp.SYSTEM.PERIP_CLK_EN0.SetBits(esp.SYSTEM_PERIP_CLK_EN0_APB_SARADC_CLK_EN)
```
aykevl, line 38: It appears that enabling random number generation is a bit more complicated than this:

https://github.com/espressif/esp-idf/blob/v4.3/components/bootloader_support/src/bootloader_random_esp32c3.c

I'm worried that it is not truly random without this complete initialization.

aykevl (changes requested):
I think the current implementation is not safe against predictable numbers. Specifically:

> When there is noise coming from the SAR ADC, the random number generator is fed with a 2-bit entropy in one clock cycle of RTC20M_CLK (20 MHz), which is generated from an internal RC oscillator (see Chapter 6 Reset and Clock for details). Thus, it is advisable to read the RNG_DATA_REG register at a maximum rate of 1 MHz to obtain the maximum entropy.

For details on how this can be implemented, see the `esp_random` function:

https://github.com/espressif/esp-idf/blob/v4.3/components/esp32c3/hw_random.c

That said, adding a RNG like this seems like a good idea (if implemented correctly).

src/crypto/rand/rand_esp32c3.go
zdima: I will move it to draft until have move details on initialization and the  #2325 with #2333 are move so I would have better access to register's fields.

deadprogram:
This support was already added in different PR so now closing. Thanks for working on this @zdima

Issue #1428, opened 2020-10-07
ESP8266 i2c support


tombrk:
The docs (https://tinygo.org/microcontrollers/esp8266/) list that the ESP8266 i2c feature is not yet implemented.

To interface sensors like the BME280, I2C is required. What needs to be done so i2c can be used?

aykevl:
The ESP8266 unfortunately does not have hardware support for I2C. See https://github.com/tinygo-org/tinygo-site/pull/125.

aykevl:
To add support for I2C, a generic software I2C library needs to be written. I think the right place for this would be the machine package. Perhaps tinygo.org/x/drivers/bitbang with type `I2C`?

So you could use it like this:

```go

import "tinygo.org/x/drivers/bitbang"
import "tinygo.org/x/drivers/bme280"

func main() {
    i2c := bitbang.NewI2C(scl, sda)
    sensor := bme280.New(i2c)
    sensor.Configure(...)
    // ...
}
```

DanValenza:
I am interested in working on the I2C bitbang library, I've found [these specs](https://www.ti.com/lit/an/slva704/slva704.pdf) for the I2C protocol, are there stable implementations in other languages out there we could refer to?

DanValenza:
I did a bit of digging and [this](https://github.com/esp8266/Arduino/blob/master/cores/esp8266/core_esp8266_si2c.cpp) is from the Arduino core library for ESP8266

cweagans:
https://github.com/periph/devices/tree/main/bitbang may be of interest here. It might be more expedient to separate that implementation from the rest of the periph.io suite, but then again, adopting periph.io to enhance tinygo device support across the board might be a net gain?

deadprogram:
This was implemented in https://github.com/tinygo-org/drivers/pull/319 and was released. Now closing, thank you!

PR #2125, opened 2021-09-21
machine: add support for i2c-bitbang

sago35:
Added I2C software implementation for esp8266 (#1428) , esp32, etc.
It can also be used for atsamd51 and others, but be aware that it is not a very accurate process.

Since the esp32 has hardware I2C, it should really implement hardware I2C instead of SI2C.
However, I thought I would lose motivation if I implemented the hardware I2C driver first, so I implemented the software I2C first.

The Frequency setting section could be written better.
Right now it runs at a Frequency that differs from the value you set to some extent, but not exactly.
If this is a problem, it needs to be implemented differently.
The time package is not available from the machine package, so we need to consider what to do.

esp32 + SI2C 100kHz (Actually, it runs at about 80 kHz)
image

esp32 + SI2C 400kHz (Actually, it runs at about 275 kHz)
image

atsamd51 + SI2C 100kHz (Actually, it runs at about 40 kHz)
image

atsamd51 + SI2C 400kHz (Actually, it runs at about 140 kHz)
image

It can be used in the following way

```go
package main

import (
	"fmt"
	"machine"
	"time"

	"tinygo.org/x/drivers/adt7410"
)

var (
	i2c    = &machine.SI2C{}
	sensor = adt7410.New(i2c)
)

func main() {
	i2c.Configure(machine.SI2CConfig{
		Frequency: machine.TWI_FREQ_400KHZ,
		SDA:       machine.IO23,
		SCL:       machine.IO18,
	})

	sensor.Configure()

	for {
		temp := sensor.ReadTempF()
		time.Sleep(time.Second)
	}

}
```

### DIFF
--- a/src/machine/i2c_generic.go
+++ b/src/machine/i2c_generic.go
@@ -1,0 +1,305 @@
+package machine
+
+import (
+	"device/arm"
+	"errors"
+)
+
+// TWI_FREQ is the I2C bus speed. Normally either 100 kHz, or 400 kHz for high-speed bus.
+const (
+	TWI_FREQ_100KHZ = 100000
+	TWI_FREQ_400KHZ = 400000
+)
+
+var (
+	errI2CWriteTimeout       = errors.New("I2C timeout during write")
+	errI2CReadTimeout        = errors.New("I2C timeout during read")
+	errI2CBusReadyTimeout    = errors.New("I2C timeout on bus ready")
+	errI2CSignalStartTimeout = errors.New("I2C timeout on signal start")
+	errI2CSignalReadTimeout  = errors.New("I2C timeout on signal read")
+	errI2CSignalStopTimeout  = errors.New("I2C timeout on signal stop")
+	errI2CAckExpected        = errors.New("I2C error: expected ACK not NACK")
+	errI2CBusError           = errors.New("I2C bus error")
+	errI2CConfigInvalid      = errors.New("I2C config error")
+)
+
+// SI2C is an I2C implementation by Software. Since it is implemented by
+// software, it can be used with microcontrollers that do not have I2C
+// function. This is not efficient but works around broken or missing drivers.
+type SI2C struct {
+	scl     Pin
+	sda     Pin
+	nack    bool
+	waitNum uint32
+}
+
+// SI2CConfig is used to store config info for SI2C.
+type SI2CConfig struct {
+	Frequency uint32
+	SCL       Pin
+	SDA       Pin
+}
+
+// Configure is intended to setup the SI2C interface.
+func (i2c *SI2C) Configure(config SI2CConfig) error {
+	// Default SI2C bus speed is 100 kHz.
+	if config.Frequency == 0 {
+		config.Frequency = TWI_FREQ_100KHZ
+	}
+
+	if config.SDA == 0 || config.SCL == 0 {
+		return errI2CConfigInvalid
+	}
+	i2c.scl = config.SCL
+	i2c.sda = config.SDA
+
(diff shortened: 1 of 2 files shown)


src/machine/i2c_generic.go
```diff
@@ -0,0 +1,305 @@
...
+// Many I2C-compatible devices are organized in terms of registers. This method
+// is a shortcut to easily read such registers. Also, it only works for devices
+// with 7-bit addresses, which is the vast majority.
+func (i2c *SI2C) ReadRegister(address uint8, register uint8, data []byte) error {
+	return i2c.Tx(uint16(address), []byte{register}, data)
+}
+
+func (i2c *SI2C) waitHalfCycle() {
+	for i := uint32(0); i < i2c.waitNum; i++ {
+		arm.Asm("nop")
```
sago35, line 303: Now we have a dependency on device/arm.
To support RISCV, etc., you may want to change to a different method.

sago35: The dependency on this part makes it arm-specific code.
Instead of i2c_generic.go, i2c_arm.go would be a better name.

src/machine/i2c_generic.go
```diff
@@ -0,0 +1,305 @@
...
+		i2c.waitHalfCycle()
+		i2c.scl.High()
+		i2c.waitHalfCycle()
+		i2c.waitHalfCycle()
+	}
+
+	i2c.scl.Low()
+	i2c.waitHalfCycle()
+	i2c.waitHalfCycle()
+	i2c.sda.Configure(PinConfig{Mode: PinInputPullup})
```
sago35, line 164: PinInputPullup seems not to exist in teensy40.

src/machine/i2c_generic.go
```diff
@@ -0,0 +1,305 @@
...
+	i2c.scl.High()
+	i2c.scl.Configure(PinConfig{Mode: PinOutput})
+
+	return nil
+}
+
+// SetBaudRate sets the communication speed for the SI2C.
+func (i2c *SI2C) SetBaudRate(br uint32) {
+	// TODO: Should be more accurate
+	i2c.waitNum = CPUFrequency() / br / 32
```
sago35, line 70: There seems to be no CPUFrequency() definition for pico.

```
/c/tinygo/tinygo/build/tinygo build -size short -o test.hex -target=pico                examples/blinky1
# machine
..\..\Users\sago3\AppData\Local\tinygo\goroot-go1.16.8-54e32ce58bba44d45bb499bf66eb2ae8700e2aa8a551addd3687341f3e1f1e61-syscall\src\machine\i2c_arm.go:72:16: undeclared name: CPUFrequency
```

sago35:
After talking with @aykevl , I decided to implement it in tinygo.org/x/drivers/i2c-bitbang.
After creating a PR in drivers, this PR will be closed.

src/machine/i2c_generic.go
aykevl: Then it should be added. This is just a simple function that returns the (fixed) CPU frequency of the RP2040.

src/machine/i2c_generic.go
aykevl: You can add architecture specific versions of `waitHalfCycle` in separate files and keep the rest of the code generic.

sago35:
Created a PR at tinygo.org/x/drivers.
This PR will be closed.

https://github.com/tinygo-org/drivers/pull/319

src/machine/i2c_generic.go
sago35: -> #2134

Issue #2118, opened 2021-09-16
Cannot compile with C FFI.


ogre2007:
Code
```
package main

/*
#include <stdlib.h>

int  sum(int a, int b) {
	return a+b;
}

*/
import "C"

import (
	"fmt"
	"unsafe"
	)
func main() {
        x := C.sum(1,2)
        _ = x
}
```
result:
```
/mnt/d/little-gopher/main.go:106: undefined reference to `sum'
collect2: error: ld returned 1 exit status
error: failed to link /tmp/tinygo929362267/main: exit status 1
```

aykevl:
At the moment, defining functions in CGo headers is not yet supported. See https://github.com/tinygo-org/tinygo/pull/2126 for a draft PR. However, you can put the function in a separate C file (main.c for example) which will be linked as usual.

PR #2126, opened 2021-09-21
cgo: support functions defined in CGo headers

aykevl:
This fixes https://github.com/tinygo-org/tinygo/issues/2118 and perhaps others.

Not complete. Merging Clang and TinyGo IR introduces some weird side effects which I would like to solve.


aykevl:
This is now ready. It doesn't support static functions defined in headers yet, I'd like to fix that in a future PR. But regular (non-static) functions work fine now.

deadprogram:
This is going be very useful! Now merging, thanks @aykevl

PR #2128, opened 2021-09-21
os: don't try to read executable path on baremetal

aykevl:
Baremetal systems usually pretend to be GOOS=linux, so an extra build tag is needed here.


deadprogram:
Given the way we use build tags and GOOS, seems necessary.

Thanks @aykevl now merging.

PR #2130, opened 2021-09-22
Clean up -target flag

aykevl:
- fix a bug introduced in #2112
- make sure that this issue is tested for, to prevent re-introducing something like it in the future
- clean up the `-target` flag, use `GOOS`/`GOARCH` environment variables if you want to cross compile (e.g. `GOOS=linux GOARCH=arm tinygo build ...`).

### DIFF
--- a/builder/config.go
+++ b/builder/config.go
@@ -13,7 +13,7 @@ import (
 // uses the currently active GOPATH (from the goenv package) to determine the Go
 // version to use.
 func NewConfig(options *compileopts.Options) (*compileopts.Config, error) {
-	spec, err := compileopts.LoadTarget(options.Target)
+	spec, err := compileopts.LoadTarget(options)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


deadprogram:
Something wrong with the Windows build it seems, @aykevl

aykevl:
I've amended the last commit to (hopefully) fix the issue on Windows.

deadprogram:
Now it is failing on different CI build. Also merge conflicts.

aykevl:

> Now it is failing on different CI build.

Yeah, it appears that tests are running so long that they hit the 10 minute timeout. I'm not sure why they do now, I don't think much has changed recently (or perhaps CircleCI is getting slower?). Maybe we can increase the timeout?

deadprogram:
OK, seems good now. Merging, thanks @aykevl

PR #2133, opened 2021-09-22
all: fix staticcheck warnings

aykevl:
This is a loose collection of small fixes flagged by staticcheck:

  - dead code
  - regexp expressions not using backticks (\`foobar` / "foobar")
  - redundant types of slice and map initializers
  - misc other fixes

Not all of these seem very useful to me, but in particular dead code is
nice to fix. I've fixed them all just so that if there are problems,
they aren't hidden in the noise of less useful issues.


deadprogram:
I am all in favor of following staticcheck's recommendations. Now merging, thanks @aykevl

PR #1897, opened 2021-05-20
add rp2040, pico

rythmkraze:
- Adds support for the new Raspberry Pi `Pico` board along with the onboard `rp2040` mcu.
- Contains minimal implementation, only enough to get `examples/blinky1` working
- All of the mcu peripheral setup code is based on the well structured and well documented Raspberry Pi [pico-sdk](https://github.com/raspberrypi/pico-sdk), with little changes to make it idiomatic Go.
- Instead of having all peripheral code in a single package `machine`, I have attempted to split each peripheral into its own package. We can move them all into package `machine` if it doesn't fit well with the current style of organization.

I am not well versed in assembly and linker scripts and not a regular coder either :) Please guide me through the review/correction process. 

Additional Resources:
- [RP2040 Getting Started Page](https://www.raspberrypi.org/documentation/rp2040/getting-started/)
- [Pico Board Datasheet](https://datasheets.raspberrypi.org/pico/pico-datasheet.pdf)
- [RP2040 MCU Datasheet](https://datasheets.raspberrypi.org/rp2040/rp2040-datasheet.pdf)

PR #2134, opened 2021-09-23
rp2040: add CPUFrequency()

sago35:
I believe it's 125 MHz from the source code.

### DIFF
--- a/src/machine/machine_rp2040_clocks.go
+++ b/src/machine/machine_rp2040_clocks.go
@@ -14,6 +14,10 @@ const (
 	MHz = 1000000
 )
 
+func CPUFrequency() uint32 {
+	return 125 * MHz
+}
+
 // clockIndex identifies a hardware clock
 type clockIndex uint8
 


src/machine/machine_rp2040_clocks.go
```diff
@@ -14,6 +14,10 @@ const (
 	MHz = 1000000
 )
 
+func CPUFrequency() uint32 {
+	return 125 * MHz
```
aykevl, line 18: According to the datasheet (https://datasheets.raspberrypi.org/rp2040/rp2040-datasheet.pdf), it's (up to) 133MHz.
I didn't write the code, so I don't know what frequency it's actually running at.

See: https://github.com/tinygo-org/tinygo/pull/1897, CC @rythmkraze.

alankrantas: In MicroPython it runs at 125 MHz by default, but you can set it as high as 270 MHz (and this applies to both cores). Probably not recommended for long term use though.

deadprogram:
Good enough for now to at least have a sensible default.

Merging, thanks @sago35

PR #2135, opened 2021-09-24
cgo: add support for stdio in picolibc

aykevl:
This is simple: just add the required header file. But this PR also adds a simple test to make sure it remains supported.


aykevl:
This doesn't work yet on WebAssembly. That's fixable, but requires a bit more CGo work.

aykevl:
Updated so that a function like `puts` works when called from Go (CGo), even on baremetal systems! It will forward the written bytes to the `putchar` function in the runtime. It also works on WebAssembly now.

deadprogram:
This will be very useful for hacky debug sessions, among other more serious uses. :smile_cat: 

Thanks @aykevl now merging.

PR #2136, opened 2021-09-24
cgo: implement rudimentary C array decaying

aykevl:
This is just a first step. It's not complete, but it gets some real world C code to parse.

This signature, from the ESP-IDF:

    esp_err_t esp_wifi_get_mac(wifi_interface_t ifx, uint8_t mac[6]);

Was previously converted to something like this (pseudocode):

    C.esp_err_t esp_wifi_get_mac(ifx C.wifi_interface_t, mac [6]uint8)

But this is not correct. C array parameters will decay. The array is passed by reference instead of by value. Instead, this would be the correct signature:

    C.esp_err_t esp_wifi_get_mac(ifx C.wifi_interface_t, mac *uint8)

So that it can be called like this (using CGo):

    var mac [6]byte
    errCode := C.esp_wifi_get_mac(C.ESP_IF_WIFI_AP, &mac[0])

This stores the result in the 6-element array mac.

### DIFF
--- a/cgo/libclang.go
+++ b/cgo/libclang.go
@@ -196,7 +196,7 @@ func tinygo_clang_globals_visitor(c, parent C.GoCXCursor, client_data C.CXClient
 			}
 			fn.args = append(fn.args, paramInfo{
 				name:     argName,
-				typeExpr: p.makeASTType(argType, pos),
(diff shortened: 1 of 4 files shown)


deadprogram:
A good step. Merging, thanks @aykevl

PR #2137, opened 2021-09-24
feather-stm32f405, feather-rp2040: add I2C pin names

sago35:
Added a Pin name to access I2C like other feather boards such as feather-m4.

feather-stm32f405
https://learn.adafruit.com/assets/83680

feather-rp2040
https://learn.adafruit.com/assets/100337


deadprogram:
Thanks @sago35 now merging.

PR #2138, opened 2021-09-24
WIP: support for WiFi on the ESP32-C3

aykevl:
This is an experimental branch. It's not complete. See https://github.com/tinygo-org/drivers/pull/320.

I plan on cleaning up some commits from here to submit separately, to hopefully make reviewing easier. See #2135 and #2136 for example.


Dorbmon:
Is this pull request still active?

aykevl:
Everything except for FreeRTOS compatibility has been merged already, so let's close this. I don't plan on implementing wifi/BLE support by implementing these FreeRTOS shims so that part can be dropped.

Also see: https://github.com/tinygo-org/drivers/pull/650

PR #2140, opened 2021-09-26
src/internal/bytealg: fix indexing error in Compare()

### LINKED ISSUE #2139: bytes.Compare return wrong value
```go
package main
import "bytes"
func main() {
	a := []byte("02056f8d91fc8996bd3f2fa4dc2dd60d6cdd05811db0065dd4f191b8bbd8eda5d4")
	b := []byte("02cbb8488ce044116dd4411c7c76ca44bb42b76400a0eb8353bf62a327423da26c")
	r := bytes.Compare(a, b)
	println("++++bytes.Compare:", r)
}
```

expected output:
```
++++bytes.Compare: -1
```

got:
```
++++bytes.Compare: 0
```

dgryski:
Fixes #2139

### DIFF
--- a/src/internal/bytealg/bytealg.go
+++ b/src/internal/bytealg/bytealg.go
@@ -15,9 +15,9 @@ func Compare(a, b []byte) int {
 	// Compare for differing bytes.
 	for i := 0; i < len(a) && i < len(b); i++ {
 		switch {
-		case a[0] < b[0]:
+		case a[i] < b[i]:
 			return -1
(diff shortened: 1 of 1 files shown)


dgryski:
For manual testing, I imported https://github.com/golang/go/blob/master/src/bytes/compare_test.go and ensured that the tests all pass.  We should probably do more testing to ensure the rest of the bytes package is also green.

deadprogram:
Thanks for the fix @dgryski now merging.

PR #2144, opened 2021-09-27
Add testing.short and t.Parallel()

dgryski:
This makes it easier to run standard Go tests unmodified (although many uses of testing.Short() use t.Skip() which is still unimplemented.)

t.Parallel() is just a stub, similar to t.Helper().


aykevl (approved):
Looks good to me!

deadprogram:
Thanks @dgryski now merging.

PR #2145, opened 2021-09-27
Tasks scheduler for RISC-V

aykevl:
This is now even more important to support the ESP32-C3.
This PR doesn't add support for the scheduler on the ESP32-C3, that will require some more work. But it is a prerequisite.

Also see #1220 and #963. This is a continuation of #963.


niaow (approved):
I don't see anything wrong here, although I do not have a physical machine to test this on.

deadprogram:
Thanks @aykevl for the implementation and to @niaow for looking this over.

Now merging.

PR #2146, opened 2021-09-27
ESP32-C3: implement atomics

aykevl:
32-bit atomics need to be implemented for the ESP32-C3 because it lacks the A (atomic) extension. With this commit, flashing ./testdata/atomic.go to the ESP32-C3 works correctly and produces the expected output on the serial console.


deadprogram:
:atom: looking good to me @aykevl thanks! Now merging.

PR #2147, opened 2021-09-28
main: implement `tinygo lldb` subcommand

aykevl:
LLDB mostly works on most platforms, but it is still lacking in some features. For example, it doesn't seem to support RISC-V yet (coming in LLVM 12), it only partially supports AVR (no stacktraces), and it doesn't seem to support the Ctrl-C keyboard command when running a binary for another platform (e.g. with `GOOS=arm64`). However, it does mostly work, even on baremetal systems.

Example when running `tinygo lldb -target=microbit examples/serial`:

Screenshot from 2021-09-28 20-11-05

This, to me, is a lot more readable than the following exact same command sequence using GDB (`tinygo gdb -target=microbit examples/serial`):

Screenshot from 2021-09-28 20-14-51

I'm not sure why the two debuggers behave differently (the breakpoint location is the same), but having some diversity is probably a good thing. If GDB is too old or crashes during debugging, then there is always another option available.

### DIFF
--- a/builder/commands.go
+++ b/builder/commands.go
@@ -21,6 +21,7 @@ func init() {
 	commands["clang"] = []string{"clang-" + llvmMajor}
 	commands["ld.lld"] = []string{"ld.lld-" + llvmMajor, "ld.lld"}
 	commands["wasm-ld"] = []string{"wasm-ld-" + llvmMajor, "wasm-ld"}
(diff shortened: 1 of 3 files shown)


deadprogram:
Very interesting new feature! Thanks @aykevl now merging.

PR #2148, opened 2021-09-29
esp32: remove extra configuration

sago35:
I tried to fix the spi extra configration in #2119, but it didn't work.
The Arduino-Core source does not use MISO_DELAY on the SPI-Master side, so I think removing it as shown in this PR is the right thing to do at this point.

> MOSI_DELAY_MODE is also used only in the slave side below.
> https://github.com/espressif/arduino-esp32/blob/master/tools/sdk/esp32/include/hal/esp32/include/hal/spi_ll.h#L370-L425


aykevl:
I'm not sure why this needs to be removed, but you've done all the testing and if it fixed the problem, then it seems like an improvement. I wonder why there is such a difference with the datasheet.

In any case, let's merge this.

PR #2154, opened 2021-09-30
cgo: run CGo parser for all CGo fragments in a file

aykevl:
Previously, libclang was run on each fragment (import "C") separately. However, in regular Go it's possible for later fragments to refer to types in earlier fragments so they must have been parsed as one.

This commit changes the behavior to run only one C parser invocation for each Go file.

In practice, it means that code like this now works:

```go
package main

// typedef int myint;
import "C"

// typedef myint myint2;
import "C"

func main() {
    var _ C.myint2
}
```

Previously, it would exit with the following error:

```
 tinygo run foobar.go
# command-line-arguments
foobar.go!cgo.c:2:12: error: unknown type name 'myint'
```

This follows the behavior of upstream Go, which also accepts this code.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -21,6 +21,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/tinygo-org/tinygo/cgo"
 	"github.com/tinygo-org/tinygo/compileopts"
 	"github.com/tinygo-org/tinygo/compiler"
 	"github.com/tinygo-org/tinygo/goenv"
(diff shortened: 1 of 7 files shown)


deadprogram:
@aykevl if you rebase `dev` into this branch it will probably pass the CI now.

aykevl:
Rebased, and changed the PR a little bit. Added a `cgo.Version` constant for better cache invalidation.

deadprogram:
> This follows the behavior of upstream Go, which also accepts this code.

Certainly a real need for us as well. Thanks for the enhancement! Now merging.

PR #2155, opened 2021-09-30
compiler: add nounwind attribute

aykevl:
This attribute is also set by Clang when it compiles C source files (unless `-fexceptions` is set). The advantage is that no unwind tables are emitted on Linux (and perhaps other systems) and therefore binaries are slightly smaller. It also avoids `__aeabi_unwind_cpp_pr0` on ARM when using the musl libc.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -23,7 +23,7 @@ import (
 // Version of the compiler pacakge. Must be incremented each time the compiler
 // package changes in a way that affects the generated LLVM module.
 // This version is independent of the TinyGo version number.
-const Version = 20 // last change: fix export math functions issue
+const Version = 21 // last change: add nounwind attribute
 
 func init() {
 	llvm.InitializeAllTargets()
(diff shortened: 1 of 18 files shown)


aykevl:
Unfortunately there seems to be a problem with wasmtime.dev.

aykevl:
Or more likely: a problem with CircleCI. The command `curl https://wasmtime.dev/install.sh -sSf` runs fine on my laptop.

aykevl:
Looks like the problem is that the Docker containers don't contain the ISRG root certificate. The older certificate chain (from IdenTrust) is now expired.
For details, see: https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/

deadprogram:
Same here @aykevl if you rebase `dev` into this branch it will probably pass the CI now.

aykevl:
Rebased.

aykevl:
Rebased again to fix a merge conflict.

aykevl:
This PR should fix the build failures in the drivers repository. @deadprogram can you take a look?

ysoldak:
Gogogo, we have many PRs in drivers waiting for this.

deadprogram:
Thanks everyone for working on this! Merging.

PR #2159, opened 2021-10-04
build: double timeout period without output for make test

deadprogram:
This PR fixes issues with the CircleCI build by doubling the timeout period without output for make test.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -253,6 +253,7 @@ commands:
       - run:
           name: "Test TinyGo"
           command: make test
+          no_output_timeout: 20m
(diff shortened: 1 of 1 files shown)


aykevl (approved):
Looks good to me. However, I believe Go itself also has a timeout of 10 minutes for tests, so we might hit that one next. For now, 20 minutes should be good enough.

(Off-topic: we might want to investigate ways to lower test duration)

deadprogram:
> (Off-topic: we might want to investigate ways to lower test duration)

please oh please! :+1: 

Merging.

PR #2160, opened 2021-10-04
ci: switch to GitHub Actions for Windows builds

aykevl:
GitHub Actions is faster and much better integrated into GitHub than Azure Pipelines, and is in general easier to use. Therefore, it seems like a good idea to me to switch to GitHub Actions instead.

It does have one major downside though: single-file release artifacts aren't really supported. This is something that should come to GitHub Actions, but isn't there yet. A workaround is to download the artifact, extract it, and upload the extracted file. Still, the improved performance might be worth it: it builds LLVM in 1½ hour instead of ~3 hours.


aykevl:
Not sure why CircleCI builds are still indicated as running, they are in fact long finished (successfully) if you click on the Details link.

azure-pipelines.yml
```diff
@@ -1,88 +0,0 @@
...
-      displayName: Build LLVM
-      inputs:
-        targetType: inline
-        script: |
-          if [ ! -f llvm-build/lib/liblldELF.a ]
-          then
-            # install dependencies
-            choco install ninja
-            # hack ninja to use fewer jobs
-            echo -e 'C:\\ProgramData\\Chocolatey\\bin\\ninja -j4 %*' > /usr/bin/ninja.bat
```
sago35, line 43: This part does not exist in the GitHubActions yaml.
Is it OK to delete it?
Was it a patch for azure-pipelines?

aykevl: I hope so, yes. I haven't really tested it: the cache that's now used was executed with this code still present.
I can add it again to be sure.

sago35:
It looked like circleci was succeeding at everything but failing to notify pr. I ran circleci again, and this time it seemed to succeed.

sago35:
When I was building with azule-pipelines, the version hash was a little different.
In GitHubActions, it's the same as the commit hash (== 6a14398a535562378d9b4b188d9b526e37bca065).
I think that's very good.



> C:\Users\sago3\Downloads\release-double-zipped\release\tinygo>.\bin\tinygo.exe  version
> tinygo version 0.21.0-dev-6a14398 windows/amd64 (using go version go1.16.8 and LLVM version 11.0.0)

aykevl:
> When I was building with azule-pipelines, the version hash was a little different.
> In GitHubActions, it's the same as the commit hash (== [6a14398](https://github.com/tinygo-org/tinygo/commit/6a14398a535562378d9b4b188d9b526e37bca065)).
> I think that's very good.

I think that's because Azure Pipelines uses the merge commit to run the tests against. That's not necessarily a bad thing.

deadprogram:
Any last comments before this gets merged?

deadprogram:
Thanks @aykevl for working on this, and thanks @sago35 for reviewing.

Now merging.

PR #2161, opened 2021-10-05
rp2040: i2c baud rate handling improvements

ysoldak:
Noticed my Nano RP2040 sometimes acts strange, I2C dies, after reset.
(I use ResetOnAbort patch (#2162), so my board restarts on panic w/o power cycling.)

Apparently, it was not re-initialising correctly. Copied baud rate default settings from SPI and it seem to work.
Verified with lsm6dsox IMU over I2C on Nano RP2040.

Also, use recently introduced `CPUFrequency()` instead of hardcoded value.

P.S. On Nano RP2040 one needs to execute following sequence in addition to this fix before configuring I2C, since IMU may turn its I2C off when INT1 pin for some reason reads "high" for a moment.
```
machine.GPIO24.Configure(machine.PinConfig{Mode: machine.PinOutput})
machine.GPIO24.High()
machine.GPIO24.Low()
```

### DIFF
--- a/src/machine/machine_rp2040_i2c.go
+++ b/src/machine/machine_rp2040_i2c.go
@@ -116,10 +116,10 @@ func (i2c *I2C) Configure(config I2CConfig) error {
 	return i2c.init(config)
 }
 
-// SetBaudrate sets the I2C frequency. It has the side effect of also
+// SetBaudRate sets the I2C frequency. It has the side effect of also
 // enabling the I2C hardware if disabled beforehand.
 //go:inline
-func (i2c *I2C) SetBaudrate(br uint32) error {
+func (i2c *I2C) SetBaudRate(br uint32) error {
 
 	if br == 0 {
 		return ErrInvalidI2CBaudrate
@@ -210,7 +210,7 @@ func (i2c *I2C) init(config I2CConfig) error {
 
 	// Always enable the DREQ signalling -- harmless if DMA isn't listening
 	i2c.Bus.IC_DMA_CR.Set(rp.I2C0_IC_DMA_CR_TDMAE | rp.I2C0_IC_DMA_CR_RDMAE)
-	return i2c.SetBaudrate(config.Frequency)
+	return i2c.SetBaudRate(config.Frequency)
 }
 
 // reset sets I2C register RESET bits in the reset peripheral and then clears them.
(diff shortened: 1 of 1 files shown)


soypat:
LGTM

sago35 (approved):
Looks good to me.

src/machine/machine_rp2040_i2c.go
```diff
@@ -96,6 +96,7 @@ func (i2c *I2C) Tx(addr uint16, w, r []byte) error {
 //  SDA: 2, 6, 10, 14, 18, 26
 //  SCL: 3, 7, 11, 15, 19, 27
 func (i2c *I2C) Configure(config I2CConfig) error {
+	const defaultBaud uint32 = 115200
```
aykevl: Why 115200? That looks more like an UART baud rate. I think you intended something like 100kHz (100_000).

src/machine/machine_rp2040_i2c.go
```diff
@@ -116,8 +120,13 @@ func (i2c *I2C) Configure(config I2CConfig) error {
 // enabling the I2C hardware if disabled beforehand.
 //go:inline
 func (i2c *I2C) SetBaudrate(br uint32) error {
```
aykevl: Not an issue with this PR, but I see that this function is not spelled correctly. It should be `SetBaudRate` as in the SAM D21/D51 chips.

src/machine/machine_rp2040_i2c.go
ysoldak: I don't know, I've just copied from SPI code and verified it works.
If 100kHz is more correct and still works, I'm all to change it. Let me verify.

src/machine/machine_rp2040_i2c.go
ysoldak: Let's have a follow-up? Changing everywhere misspelled.

src/machine/machine_rp2040_i2c.go
ysoldak: Fixed

src/machine/machine_rp2040_i2c.go
ysoldak: OK, I thought this misspelling is more wide spread, apparently only this file is affected. Fixed.

ysoldak:
@aykevl Please take a look, your comments addressed

deadprogram:
Thanks @ysoldak now merging.

PR #2162, opened 2021-10-05
runtime: reset on abort for cortexm

ysoldak:
Panics happen. And right now (and still by default after this PR) system just hangs.
Feels like it can be beneficial to let users choose whether to hang or restart the system on panic.

It is not always possible to have easy physical access to embedded devices. Some may have been deployed to run 24/7 at a remote site.

Please suggest better API, this is just some quick attempt to make it useable.

Verified on my Nano RP2040, 24/7 setup, panicking and rebooting on wifinina and lsm6dsox weird situations that hard to catch and debug.

### DIFF
--- a/src/runtime/runtime_cortexm_abort.go
+++ b/src/runtime/runtime_cortexm_abort.go
@@ -6,16 +6,11 @@ import (
 	"device/arm"
 )
 
-var ResetOnAbort = false
-
 func exit(code int) {
 	abort()
 }
 
 func abort() {
-	if ResetOnAbort {
-		arm.SystemReset()
-	}
 	// lock up forever
 	for {
 		arm.Asm("wfi")
--- a/src/runtime/runtime_cortexm_abort.go
+++ b/src/runtime/runtime_cortexm_abort.go
@@ -6,11 +6,16 @@ import (
 	"device/arm"
 )
 
+var ResetOnAbort = false
+
 func exit(code int) {
 	abort()
 }
 
 func abort() {
+	if ResetOnAbort {
+		arm.SystemReset()
+	}
 	// lock up forever
 	for {
 		arm.Asm("wfi")
--- a/src/examples/onpanic/onpanic.go
+++ b/src/examples/onpanic/onpanic.go
@@ -1,0 +1,26 @@
+// +build cortexm
+
+package main
+
+import (
+	"device/arm"
+	"runtime"
+	"time"
+)
+
+// This example shows how to reset system on panic.
+// In your application you may: reset, blink an LED to indicate failure, or do something else.
+
+func main() {
+
+	runtime.OnPanic = func() {
+		println("PANIC")
+		time.Sleep(time.Second)
+		arm.SystemReset() // available on cortexm, your system may have to do this differently
+	}
+
+	println("START")
+	time.Sleep(time.Second)
+	panic("AAA!!!111")
+
+}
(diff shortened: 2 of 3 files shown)


src/runtime/runtime_cortexm_abort.go
```diff
@@ -6,11 +6,16 @@ import (
 	"device/arm"
 )
 
+var ResetOnAbort = false
```
aykevl: I think this can be generalized with a function like the following:

```go
var OnAbort func()
```

which is then called from the `abort` function if it is not nil. The big advantage is that applications can then do what they like to do: reset, blink an LED to indicate failure, or something else. But note that when this function is called, the program could be in a bad state and heap allocations might be impossible for example.

Or perhaps even this:

```go
var OnAbort func(code int)
```

Also, this API needs some documentation.

Still, I'm not 100% convinced this is a good idea: panics do indeed happen, but when they happen that would be a bug somewhere which needs to be handled instead of ignored.

ysoldak: Good suggestion, Ayke, thank you. I'll try and refactor + document.

I see your point, panics indeed indicate the errors that must be fixed. And once we give people an escape hatch to ignore them they may misuse it.
Still, I believe there is room for valid use cases. For example, when it's crucial for a remote device to be up 24/7, nice to have such "safety net", if you will. A prudent developer then make devices indicate panic did happen and must be investigated later, but shall not prevent main functionality from running.
I like to think about it as k8s would kill-restart unhealthy pods for devs to have un-interrupted service, while still providing info there was some error to look at.

ysoldak:
Made it more flexible with function, as suggested.
OnPanic in `runtime/panic.go` to enable it for all architectures, not just cortexm.
Example of resetting system on panic (for cortexm).

Once and if we settle on a final implementation, I squash commits in this PR.

ysoldak:
After a short discussion in slack, I feel like I shall not push this PR anymore.
It's problematic with extending public API in general and runtime package in particular (that we want to keep close API-wise to upstream Go).

Feel free to close PR. @aykevl 

I've solved resets on panic for me in my fork and just wanted to enable others do the same. But the costs outweight benefits and if anyone has the same problem they shall be able to find the solution now.
I'll try and solve the problem (resets on panic for 24/7 devices) with classic defer/recover and, if successful, can probably write about the pattern in tinygo docs instead.

soypat:
```go
func main() {
	defer func() {
		a := recover()
		if a != nil {
			println("panic encountered. recovering:",a)
			time.Sleep(time.Second)
			go main()
			runtime.Goexit() // Courtesy of the #darkarts channel on Slack.
		}
	}()
	panic("oh no!")
}
```
I believe this has the desired effect on upstream Go. Not sure if it would work on TinyGo though.

ysoldak:
> I believe this has the desired effect on upstream Go. Not sure if it would work on TinyGo though.

Unfortunately no.
This does not work in TinyGo, since `abort()` function just lock on `arm.Asm("wfi")` and never executes `defer`

Check an example in my fork's branch
https://github.com/ysoldak/tinygo/blob/defer-on-panic_does-not-work/src/examples/resetonpanic/resetonpanic.go

stevegt:
Recommend close this PR as @ysoldak suggests and focus on #2742.

deadprogram:
Closing as requested.

PR #2163, opened 2021-10-06
esp32c3: use tasks scheduler by default

aykevl:
Now that the tasks scheduler has been implemented for 32-bit RISC-V, it can be used for the ESP32-C3.

As an example, the following now works:

    tinygo flash -target=esp32c3 ./testdata/goroutines.go


deadprogram:
Thanks @aykevl now merging.

Issue #2153, opened 2021-09-30
Some functions of the net package are not supported. 


dosgo:
![image](https://user-images.githubusercontent.com/9545458/135422518-2d108673-0478-4b4d-ad7e-bbd44b598838.png)

Are these unsupported functions because they are not available for embedded applications or are they?

PR #2165, opened 2021-10-09
Add `net.Interface` implementation

soypat:
Resolves part of https://github.com/tinygo-org/tinygo/issues/2153.

It does not provide a fully working `net.Interface` implementation but may allow compilation of some programs which make little to no use of it while still calling it in non-critical code.

### DIFF
--- a/src/net/interface.go
+++ b/src/net/interface.go
@@ -1,0 +1,252 @@
+// The following is copied from Go 1.16 official implementation.
+
+// Copyright 2011 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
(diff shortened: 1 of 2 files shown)


deadprogram:
Seems legit to me. @aykevl @dgryski anyone else have any comments?

dgryski (approved):
LGTM

deadprogram:
Now merging thanks @soypat

deadprogram:
OK yet another example of why I need to make sure that PRs are recently rebased against `dev` before merging. Sorry @soypat but can you please help look into why the CI in `dev` is now failing?

PR #2167, opened 2021-10-10
ESP32C3 interrupts, GPIO and UART support

zdima:
Added and integrated interrupts, GPIO and UART.
The test examples are:
src/examples/esp/esp32c3/pininterrupt.go
src/examples/esp/esp32c3/echo.go
The signal vector remain in .text.
Moving it to the .init or even to the bottom of the .text segment breaks the signals functionality.


src/device/esp/esp32c3_intr.S
```diff
@@ -0,0 +1,223 @@
...
+	.option norvc
+
+	j _panic_handler            /* exception handler, entry 0 */
+	.rept 24
+	j _interrupt_handler        /* 24 identical entries, all pointing to the interrupt handler */
+	.endr
+	j _panic_handler            /* Call panic handler for ETS_T1_WDT_INUM interrupt (soc-level panic)*/
+	j _panic_handler            /* Call panic handler for ETS_CACHEERR_INUM interrupt (soc-level panic)*/
+	.rept 5
+	j _interrupt_handler        /* 6 identical entries, all pointing to the interrupt handler */
```
aykevl, line 108: Are there 5 or 6 identical entries? The comment and `.rept` statement disagree.

src/device/esp/esp32c3_intr.S
```diff
@@ -0,0 +1,223 @@
...
+	sw   t0,  31*REGSIZE(sp)
+	csrr t0, mtval
+	sw   t0,  32*REGSIZE(sp)
+	csrr t0, mhartid
+	sw   t0,  33*REGSIZE(sp)
+
+	addi t0, sp, CONTEXT_SIZE /* restore sp with the value when trap happened */
+	sw   t0,  34*REGSIZE(sp)
+
+	/* Call handleException(sp) or handleException(sp)
```
aykevl, line 135: handleException or handleException? This doesn't look right.

src/runtime/runtime_esp32c3.go
```diff
@@ -63,3 +67,39 @@ func abort() {
...
+		callInterruptHandler(id)
+	}
+}
+
+//export handleException
+func handleException(code uint32) {
+	println("*** Exception: code:", code)
+	for {
+		riscv.Asm("wfi")
+	}
```
aykevl, line 100: This should call `abort`.

src/device/esp/esp32c3_intr.S
```diff
@@ -0,0 +1,223 @@
...
+	/* Increase interrupt threshold level */
+	li   t2, 0x7fffffff
+	and  t1, s1, t2     /* t1 = mcause & mask */
+	slli t1, t1, 2      /* t1 = mcause * 4 */
+	li   t2, CPU_INT_PRI_0_REG
+	add  t1, t2, t1     /* t1 = INTC_INT_PRIO_REG + 4 * mcause */
+	lw   t2, 0(t1)      /* t2 = INTC_INT_PRIO_REG[mcause] */
+	addi t2, t2, 1      /* t2 = t2 +1 */
+	sw   t2, 0(t0)      /* CPU_INT_THRESH = t2 */
+	fence
```
aykevl, line 191: You are implementing nested interrupts here. Is this intentional? Did you test nested interrupts?

src/examples/esp/esp32c3/echo.go
```diff
@@ -0,0 +1,42 @@
...
+	"time"
+)
+
+func main() {
+
+	uartConfig := &machine.UARTConfig{
+		BaudRate: 115200,
+		TX:       machine.Pin(2),
+		RX:       machine.Pin(5),
+	}
```
aykevl, line 14: Why does this need a new example?

src/machine/machine_esp32c3.go
```diff
@@ -14,131 +13,92 @@ func CPUFrequency() uint32 {
 	return 160e6 // 160MHz
 }
 
+type Module int
```
aykevl, line 16: What's a module? It seems like this is something internal to the ESP32-C3, and therefore should not be part of the public API.

aykevl (commented):
This is a very large PR. I haven't gone through all of it yet.
Can you perhaps make it smaller, so that it is easier to review and maintain?

* This PR implements both pin interrupts and UART. I would prefer it it only implemented one of them (you can always make a PR with the other one later).
* This PR implements nested interrupts, but do we really need those? I would suggest simple non-nested interrupts (like in the fe310 chip) for now, to keep things easy to understand.
* More generally, it looks like you've copied a lot of code from ESP-IDF. Please make sure you understand what every piece of it is doing, and if you don't understand it, remove it. For example, you are passing two parameters to `handleInterrupt` from esp32c3_intr.S but don't actually use those parameters.
* Why do you move code from machine_esp32c3.go to new files? It's done for some other chips, but I honestly don't think it improves readability (it will make the list of files in the machine package very long). And in this case, it makes the diff bigger and thus the code becomes harder to review.

zdima:
@aykevl 
I have submitted another PR https://github.com/tinygo-org/tinygo/pull/2170 for interrupt only

sago35:
@zdima 
It would be nice to be able to use UART interrupts, etc.

Issue #1401, opened 2020-09-21
[Support] Arduino Pro Mini, ATmega328P, 5V/16MHz


ProximaB:
I have manage to compile tinygo and upload it to Arduino Mini Pro with avrisp mkii and prove that it works.
I used avrdude tool to upload it with serial programmer.
```bash
sudo tinygo build -target arduino -o output.hex main.go
avrdude -avrdude.conf -v -patmega328p -carduino -P/dev/ttyUSB0 -b57600 -D -Uflash:w:./output.hex 
 ```
Config used (basic arduino avrdude config): https://pastebin.com/xVtzAfm4

```
avrdude: verifying flash memory against ./output.hex:
avrdude: load data flash data from input file ./output.hex:
avrdude: input file ./output.hex auto detected as Intel Hex
avrdude: input file ./output.hex contains 558 bytes
```

**Comment by aykevl** [comment]
Cool!
Is this a request to add support for this board?

**Comment by ProximaB** [comment]
Yeah, that's it. We could add some instructions on the website: https://tinygo.org/microcontrollers/. That would be great.

**Comment by aykevl** [comment]
Actually, it would be best to add native support for this in TinyGo. Seeing this, it should not be very difficult. It looks like the board is very similar to the Arduino Uno.

**Comment by ProximaB** [comment]
Alright, I would like to contribute and add native support for Arduino Pro mini to TinyGo.
That's really tiny cheap board If it could communicate by serial bus ESP-01S it would be ready for small IoT projects. Both of that components you can get for less than one dollar. 😀

**Comment by duarte2025** [comment]
For arduino pro mini would not be the pin mapping? Well, the microcontroller is the same as the arduino one, if I'm not mistaken.

**Comment by ProximaB** [comment]
@duarte2025 That's right. I would like to add some instructions how to boot devices as well, because I had some troubles with avrdude.

**Comment by ProximaB** [comment]
I have noticed that the code lack of consistence in naming analog pins or one way of structuring code. That will cause code portability issues. 
[board_arduino.go#L12](https://github.com/ProximaB/tinygo/blob/6b8940421e169a751dd039d8c7d1b1c4a1465fb2/src/machine/board_arduino.go#L12
)
```go
	D0  = PD0 // RX
	D1  = PD1 // TX
...
	ADC0 Pin = PC0
	ADC1 Pin = PC1
....
	UART_TX_PIN Pin = PD1
	UART_RX_PIN Pin = PD0
```
[...]

PR #2168, opened 2021-10-11
Add support Arduino Pro Mini #1401

0xDEAD:
Pro Mini is basically a Uno with the same pinout on a
smaller board. Especially clones of this board are very
popular because of their low price.

This solves #1401


aykevl:
This looks like an exact copy of targets/arduino.json.
I guess that's fine, but in that case the file can be simplified to the following:
```json
{
    "inherits": "arduino"
}
```
This makes `arduino-pro-mini` an alias for `arduino`.

deadprogram:
Removing due to age/inactivity. Please open a new PR if needed, thanks!

PR #2170, opened 2021-10-12
esp32c3 interrupt implementation 

zdima:
@aykevl this is only interrupt implementation for esp32c3.
No exported APIs except enum values of `HandlerID` type to initialize chip's modules.
Tested with pin src/examples/pininterrupt/pininterrupt.go (not checked in)

### DIFF
--- a/src/runtime/interrupt/interrupt_esp32c3.go
+++ b/src/runtime/interrupt/interrupt_esp32c3.go
@@ -10,91 +10,40 @@ import (
 	"unsafe"
 )
 
-const (
-	// Each module of the ESP32C3 uses different interrupt registers.
-	// Below is the list of modules used from machine package to initialize interrupt handler
-	GPIO HandlerID = iota
-	UART0
-	UART1
-	USB
-	I2C0
-	I2S1
-	TIMG0
-	TIMG1
-	UHCI0
-	RMT
-	SPI1
-	SPI2
-	TWAI
-	RNG
-	WIFI
-	BT
-	WIFI_BT_COMMON
-	BT_BASEBAND
-	BT_LC
-	RSA
-	AES
-	SHA
-	HMAC
-	DS
-	GDMA
-	SYSTIMER
-	SARADC
-)
-
-var (
-	ErrInterruptNotImplemented = errors.New("interrupt number is not implemented")
-)
-
-type HandlerID int
-
+// Enable register CPU interrupt with interrupt.Interrupt.
+// The ESP32-C3 has 31 CPU independent interrupts.
+// The Interrupt.New(x, f) (x = [1..31]) attaches CPU interrupt to function f.
+// Caller must map the selected interrupt using following sequence (for example using id 5):
+//
+//    // map interrupt 5 to my XXXX module
+//    esp.INTERRUPT_CORE0.XXXX_INTERRUPT_PRO_MAP.Set( 5 )
+//    _ = Interrupt.New(5, func(interrupt.Interrupt) {
+//        ...
+//    }).Enable()
 func (i Interrupt) Enable() error {
+	if i.num < 1 && i.num > 31 {
+		return errors.New("interrupt for ESP32-C3 must be in range of 1 through 31")
+	}
 	mask := riscv.DisableInterrupts()
 	defer riscv.EnableInterrupts(mask)
 
-	// map mapRegister to CPU interrupt
-	intr, err := HandlerID(i.num).set()
-	if err != nil {
-		return err
-	}
+	// enable CPU interrupt number i.num
+	esp.INTERRUPT_CORE0.CPU_INT_ENABLE.SetBits(1 << i.num)
 
-	esp.INTERRUPT_CORE0.CPU_INT_ENABLE.SetBits(1 << intr)
-	esp.INTERRUPT_CORE0.CPU_INT_TYPE.SetBits(1 << intr)
-	// Set threshold to 5
-	reg := (*volatile.Register32)(unsafe.Pointer((uintptr(unsafe.Pointer(&esp.INTERRUPT_CORE0.CPU_INT_PRI_0)) + uintptr(intr)*4)))
+	// Set pulse interrupt type (rising edge detection)
+	esp.INTERRUPT_CORE0.CPU_INT_TYPE.SetBits(1 << i.num)
+
+	// Set default threshold to 5
+	reg := (*volatile.Register32)(unsafe.Pointer((uintptr(unsafe.Pointer(&esp.INTERRUPT_CORE0.CPU_INT_PRI_0)) + uintptr(i.num)*4)))
 	reg.Set(5)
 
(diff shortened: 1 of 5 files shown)


src/runtime/interrupt/interrupt_esp32c3.go
```diff
@@ -0,0 +1,198 @@
...
+
+	} else {
+		// Topmost bit is clear, so it is an exception of some sort.
+		// We could implement support for unsupported instructions here (such as
+		// misaligned loads). However, for now we'll just print a fatal error.
+		handleException(interruptNumber)
+	}
+}
+
+//export handleException
```
aykevl: Why is this exported?

src/runtime/runtime_esp32c3.go
```diff
@@ -63,3 +68,28 @@ func abort() {
 		riscv.Asm("wfi")
 	}
 }
+
+// Init initialize the interrupt controller and called from runtime once.
```
aykevl:
```suggestion
// interruptInit initialize the interrupt controller and called from runtime once.
```

src/runtime/interrupt/interrupt_esp32c3.go
```diff
@@ -0,0 +1,198 @@
...
+	cause := riscv.MCAUSE.Get()
+	interruptNumber := uint32(cause & 0x1f)
+
+	if cause&(1<<31) != 0 && interruptNumber > 0 {
+		// disable this interrupt
+		interruptBit := uint32(1 << interruptNumber)
+		esp.INTERRUPT_CORE0.CPU_INT_ENABLE.ClearBits(interruptBit)
+
+		// reset this interrupt
+		if esp.INTERRUPT_CORE0.CPU_INT_TYPE.Get()&(interruptBit) != 0 {
```
aykevl: The extra parentheses are unnecessary.

```suggestion
		if esp.INTERRUPT_CORE0.CPU_INT_TYPE.Get() & interruptBit != 0 {
```

src/runtime/interrupt/interrupt_esp32c3.go
```diff
@@ -0,0 +1,198 @@
...
+
+	riscv.Asm("fence")
+	return nil
+}
+
+// getHandlerIDsForInterrupt return registered codes for interrupt
+func getHandlerIDsForInterrupt(interrupt int) []HandlerID {
+	switch interrupt {
+	case 1:
+		return []HandlerID{GPIO}
```
zdima: @aykevl, is this bad in terms of allocation?

niaow: You could probably fix this by moving the arrays into globals.

zdima: Thanks, move to global.

src/runtime/interrupt/interrupt_esp32c3.go
```diff
@@ -0,0 +1,173 @@
...
+		// We could implement support for unsupported instructions here (such as
+		// misaligned loads). However, for now we'll just print a fatal error.
+		handleException(mcause)
+	}
+}
+
+//export handleException
+func handleException(mcause uintptr) {
+	// TODO need to get location of actual MEPC from the stack stash created in src/device/riscv/handleinterrupt.S
+	println("*** Exception:     pc:", riscv.MEPC.Get())
```
aykevl: I don't understand the TODO, isn't this the actual MEPC?

aykevl (commented):
I've been thinking about this PR and how to implement interrupts correctly on the ESP32-C3.
I think it would actually make more sense to make the `Interrupt` type be the 31 interrupt vectors. It would then work as follows:

* Code that wants to use an interrupt can do so using `interrupt.New`, for example `interrupt.New(5, func(interrupt.Interrupt) { ... })`.
* It will have to manually map the interrupt source to the interrupt vector, for example `esp.INTERRUPT_CORE0.GPIO_INTERRUPT_PRO_MAP.Set(5)`.
* The `handleInterrupt` logic now becomes a lot simpler. It can now directly call `callInterruptHandler` with the interrupt number from `mcause`.

It is still possible to map multiple interrupt sources to a single interrupt vector: `interrupt.New` can be used for multiple handlers on a single interrupt number. For example:

```go
// configuring GPIO:
intr := interrupt.New(5, func(interrupt.Interrupt) {
    status := esp.GPIO.STATUS.Get()
    if status != 0 {
        // handle GPIO interrupt
    }
})
intr.Enable()
esp.INTERRUPT_CORE0.GPIO_INTERRUPT_PRO_MAP.Set(5)

// configuring UART:
intr := interrupt.New(5, func(i interrupt.Interrupt) {
    flag := esp.UART0.INT_ST.Get()
    if flag != 0 {
        // handle UART interrupt
    }
})
intr.Enable()
esp.INTERRUPT_CORE0.UART0_INTERRUPT_PRO_MAP.Set(5)
```

(These two don't need to be on the same interrupt vector, I'm only showing that it is possible).

What do you think?

src/runtime/interrupt/interrupt_esp32c3.go
zdima: will remove it was part of the interrupt table

src/runtime/interrupt/interrupt_esp32c3.go
zdima: I don't know how to access the saved registers on the stack in handleinterrupt.S.
I want to print all the registers here and I can add it If you can suggest how I can access location of the stack created by handleinterrupt.S

zdima:
I didn't know you can call `interrupt.New` with the same id. I never get response from you on how does it work in the slack PM :(. If I knew, perhaps the implementation will be much simpler.
So how does it work if make a call `interrupt.New(5, func(interrupt.Interrupt) {})` twice?

zdima:
@aykevl not sure if you get notification on my comments above

src/runtime/interrupt/interrupt_esp32c3.go
zdima: removed

aykevl (approved):
Looks good to me, thank you for all the work!

aykevl:
CI is failing, but I think that's an unrelated issue.

aykevl:
Oh now I see. You have made the PR against the espnet branch. Can you rebase it against the dev branch?

zdima:
I could, but don’t we need other changes you made like stdio, pico lib and FreeRTOS?

aykevl:
> I could, but don’t we need other changes you made like stdio, pico lib and FreeRTOS?

Yes, but by merging this in the dev branch it will become available to everybody, not just those working from the espnet branch. That's why I also made a few other PRs like #2135, #2136, and #2143. In general I prefer working as much as possible directly from the dev branch for independent features.
After the PR is merged, it can easily be added to the espnet branch too.

zdima:
this is merged with PR #2200

PR #2171, opened 2021-10-13
rutime/gc_leaking: ensure heapptr is aligned

dgryski:


aykevl:
Can you make this so that alignment is only done for wasm? That would make it more obvious what it's for and avoids increasing code size on other platforms. And perhaps also with a comment, see: https://github.com/tinygo-org/tinygo/pull/2014

```go
ptr := heapStart
if GOARCH == "wasm" {
    ptr = align(ptr)
}
heapptr = ptr
```

(This is optimized to `heapptr = heapStart` on non-wasm architectures).

dgryski:
Fixed and force-pushed.

aykevl:
Thank you!

Issue #2172, opened 2021-10-13
infinite recursion trying to create recursive function type


dgryski:
Tinygo does not like `fnType` in the below program:

```
package main

type fnType func(int) (fnType, error)

func printNode(int) (fnType, error) {
        return nil, nil
}

func main() {
}
```

tinygo errors with a stack overflow; the callstack is filled with
```
	/Users/dgryski/go/src/github.com/tinygo-org/tinygo/compiler/func.go:99 +0x8c fp=0xc020d40560 sp=0xc020d40520 pc=0x429c7ec
github.com/tinygo-org/tinygo/compiler.(*compilerContext).makeLLVMType(0xc0005b0000, {0x4413240, 0xc000ca67e0})
	/Users/dgryski/go/src/github.com/tinygo-org/tinygo/compiler/compiler.go:388 +0x2b4 fp=0xc020d40628 sp=0xc020d40560 pc=0x4284cd4
github.com/tinygo-org/tinygo/compiler.(*compilerContext).getLLVMType(0xc0005b0000, {0x4413240, 0xc000ca67e0})
	/Users/dgryski/go/src/github.com/tinygo-org/tinygo/compiler/compiler.go:321 +0x6e fp=0xc020d40668 sp=0xc020d40628 pc=0x428498e
github.com/tinygo-org/tinygo/compiler.(*compilerContext).makeLLVMType(0xc0005b0000, {0x44131f0, 0xc000128e80})
	/Users/dgryski/go/src/github.com/tinygo-org/tinygo/compiler/compiler.go:383 +0x771 fp=0xc020d40730 sp=0xc020d40668 pc=0x4285191
github.com/tinygo-org/tinygo/compiler.(*compilerContext).getLLVMType(0xc0005b0000, {0x44131f0, 0xc000128e80})
	/Users/dgryski/go/src/github.com/tinygo-org/tinygo/compiler/compiler.go:321 +0x6e fp=0xc020d40770 sp=0xc020d40730 pc=0x428498e
github.com/tinygo-org/tinygo/compiler.(*compilerContext).getRawFuncType(0xc0005b0000, 0xc000ca67e0)
	/Users/dgryski/go/src/github.com/tinygo-org/tinygo/compiler/func.go:125 +0x4cf fp=0xc020d40828 sp=0xc020d40770 pc=0x429cd4f
github.com/tinygo-org/tinygo/compiler.(*compilerContext).getFuncType(0xc0005b0000, 0x0)
```

[...]

PR #2175, opened 2021-10-13
compiler: add support for recursive function types

aykevl:
This adds support for a construct like this:

    type foo func(fn foo)

Unfortunately, LLVM cannot create function pointers that look like this.
LLVM only supports named types for structs (not for pointers) and thus
can't add a pointer to a function type of the same type to a parameter
of that function type.

The fix is simple: cast all function pointers to a void function, in
LLVM IR:

    void ()*

Raw function pointers are cast to this type before storing, and cast
back to the regular function type before calling. This means that
function parameters will never refer to its own type because raw
function types are fixed at that one type.

Somehow, this does have an effect on binary size in some cases. The
effect is small and goes both ways. On top of that, there is work
underway in LLVM which would make all pointer types opaque (without a
pointee type). This would make this whole commit useless and therefore
should fix any size increases that might happen.
https://llvm.org/docs/OpaquePointers.html

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -23,7 +23,7 @@ import (
 // Version of the compiler pacakge. Must be incremented each time the compiler
 // package changes in a way that affects the generated LLVM module.
 // This version is independent of the TinyGo version number.
-const Version = 22 // last change: check for divide by zero
+const Version = 23 // last change: fix recursive function types
 
 func init() {
 	llvm.InitializeAllTargets()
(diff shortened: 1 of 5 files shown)


aykevl:
Rebased to fix the merge conflict.
This PR solves https://github.com/tinygo-org/tinygo/issues/2172.

niaow:
I just gave this a rebase so that #2209 is applied and CI should run (I hope this was okay?)

niaow (approved):
This seems to be working as expected. LGTM.

deadprogram:
Thanks @aykevl for working on this and thanks @niaow for review.

Now merging.

PR #2177, opened 2021-10-14
transform: remove some dead code

aykevl:
This showed up in the linter and it makes sense to just remove it.


deadprogram:
:scissors: :heart: now merging. Thanks!

PR #2178, opened 2021-10-14
runtime (js/wasm): avoid scheduler recursion

niaow:
The wasm "log" test attaches a callback to a button and then triggers a click event.
The click event then calls back into the WASM module, spawning another copy of the scheduler loop.

This change detects this kind of nesting and will now spawn the event handler but not the scheduler.
A debug assertion has also been added to the scheduler to help track down similar issues in the future.


niaow:
Nevermind, I seem to have messed up something between testing and committing. . .

PR #2180, opened 2021-10-15
src/testing: stub B.ReportAllocs()

dgryski:
This allows test packages that use this feature in their benchmarks
to build and run (if not the benchmarks themselves).


aykevl:
Does this allow any new packages to pass tests? If so, can you add those to `TEST_PACKAGES`?

aykevl:
Thank you!

PR #2181, opened 2021-10-15
src/runtime: add another set of invalid unicode runes to encodeUTF8()

dgryski:
Behaviour is now identical between Go and tinygo for the following test program:

```
package main

import (
	"fmt"
	"unicode/utf8"
)

func main() {
	for i := 0; i < utf8.MaxRune; i++ {
		r := rune(i)
		got := string(r)
		want := string(0xFFFD) // "invalid rune" rune

		if got == want {
			fmt.Printf("0x%x\n", i)
		}
	}
}
```

### DIFF
--- a/src/runtime/string.go
+++ b/src/runtime/string.go
@@ -160,6 +160,9 @@ func encodeUTF8(x rune) ([4]byte, uintptr) {
 		b1 := 0xc0 | byte(x>>6)
 		b2 := 0x80 | byte(x&0x3f)
 		return [4]byte{b1, b2, 0, 0}, 2
+	case 0xd800 <= x && x <= 0xdfff:
+		// utf-16 surrogates are replaced with "invalid code point"
+		return [4]byte{0xef, 0xbf, 0xbd, 0}, 3
 	case x <= 0xffff:
 		b1 := 0xe0 | byte(x>>12)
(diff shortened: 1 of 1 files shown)


src/runtime/string.go
```diff
@@ -160,6 +160,10 @@ func encodeUTF8(x rune) ([4]byte, uintptr) {
 		b1 := 0xc0 | byte(x>>6)
 		b2 := 0x80 | byte(x&0x3f)
 		return [4]byte{b1, b2, 0, 0}, 2
+
+	case 0xd800 <= x && x <= 0xdfff:
+		return [4]byte{0xef, 0xbf, 0xbd, 0}, 3
+
```
aykevl: 1. Can you please follow the existing convention and leave out the newlines? (Whether or not there should be newlines is debatable, but consistency is important).
2. While the information is in the link, it's not immediately obvious what happens. Can you add a short comment saying these are UTF-16 surrogates? That makes the code a bit easier to understand.

aykevl (commented):
Re-running the tests now, hopefully that fixes the CI timeout.

src/runtime/string.go
dgryski: Fixed and force-pushed.

PR #2182, opened 2021-10-15
compileopts: add `purego` buildtag to dodge unsupported asm code

dgryski:


aykevl (approved):

aykevl:
Actually, you said that now some packages will pass all tests? Can you please add those to the Makefile (`TEST_PACKAGES`) so that they are also tested in CI?

dgryski:
The tests enabled by `purego` were in the /x/crypto repo, not stdlib.

There are other stdlib packages that will pass with only minor modifications (such as `crypto/sha1` which passes after removing `issue15617_test.go` which requires mmap and the sycall package).

aykevl:
> The tests enabled by `purego` were in the /x/crypto repo, not stdlib.

Hmmm. In that case, I'm not entirely sure we should add this build tag to TinyGo itself.
Not sure what the proper fix would be. Adding support for Go assembly would probably help in many cases, but perhaps not all.
Perhaps it's best to let users use the `purego` tag manually until Go assembly support has been added. At that point, we can check whether any further action is needed.

dgryski:
SGTM. Closing for now.

PR #2183, opened 2021-10-15
Simplify some SERCOM declarations

aykevl:
This simplifies some SERCOM declarations:

  * It avoids lots of repetitive and error-prone code (I found two bugs).
  * It reduces code size, thus simplifying maintenance.
  * It avoids the need for special *_baremetal.go files, and thus allows all these boards to be used in the [TinyGo playground](https://play.tinygo.org).

Overall, it removes over 400 lines of (difficult to test) code, which I think is an improvement. In my tests, there was sometimes a change in the produced binaries but there was no change in code size.


sago35 (approved):
Looks good to me.
I think it is very good, especially that _baremetal.go is gone.

deadprogram:
Thanks @aykevl for working on this, and thanks @sago35 for review.

Now merging.

PR #2184, opened 2021-10-16
Makefile: remove crypto/sha1 from TEST_PACKAGES

dgryski:
Due to issue15617_test's use of /x/sys/unix.Mmap, this package
no longer builds and runs out of the box on Mac

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,7 +195,6 @@ TEST_PACKAGES = \
 	crypto/dsa \
 	crypto/md5 \
 	crypto/rc4 \
-	crypto/sha1 \
 	crypto/sha256 \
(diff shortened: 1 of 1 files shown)


dgryski:
I don't understand why this crypto/sha1 passes in CI but fails locally.

aykevl:
Well, it does pass for me. I'm using linux/amd64, maybe it just doesn't work on a different OS? Or a different Go version?

dgryski:
Fixed by [`5a07ffe` (#2286)](https://github.com/tinygo-org/tinygo/pull/2286/commits/5a07ffea0352458c4048db6be4a849b1815a2763)

PR #2188, opened 2021-10-20
hifive1b: increase CPU frequency from 16MHz to 320MHz

aykevl:
This chip can run so much faster! Let's update the default frequency.

Also, change the UART implementation to be more fexible regarding the clock frequency.


deadprogram:
:racehorse: thanks @aykevl looks good. Now merging.

PR #2189, opened 2021-10-20
fe310: add support for bit banging drivers

aykevl:
This is necessary to add support for WS2812.


deadprogram:
:zero: :one: thanks @aykevl now merging.

PR #2192, opened 2021-10-20
machine: support Pin.Get() function when the pin is configured as output

aykevl:
To my surprise, this is supported on all the devices I could test so therefore it makes sense to change the API to allow this.

Tested on all the devices that have a changed comment. Unfortunately I could not test on the NXP chips (because I don't have them) and on the RP2040 (because my Pico somehow doesn't show up when I try to flash it).

I only needed to change the nrf series to enable the pin input when it is configured as an output.


deadprogram:
:round_pushpin: thank you @aykevl now merging.

PR #2193, opened 2021-10-20
Proposal to fix #2187 WASI gc=conservative frees data which is still malloc'ed by cgo code 

### LINKED ISSUE #2187: WASI gc=conservative frees data which is still malloc'ed by cgo code
The code to malloc, calloc and free in [src/runtime/arch_tinygowasm.go](https://github.com/tinygo-org/tinygo/blob/65c1978965bc2f00d07765978f977617ba40b28f/src/runtime/arch_tinygowasm.go#L69) does not ensure that malloc'ed memory is not freed during GC.
I experienced the gc=conservative freeing memory which was still used during the execution of a cgo function. The only option to avoid this was to use gc=leaking.

Not sure if this workaround has any undesired side-effects, but I could resolve the issue by keeping a reference to the alloced cgo pointers in Go.

Here is what I did:
1. Add a map to the var (...) section of arch_tinygowasm.go: `cgoAllocations = make(map[uintptr]uintptr)`
2. Fill the map in libc_malloc and libc_calloc: `cgoAllocations[uintptr(allocResult)] = size`
3. Delete from the map in libc_free: `delete(cgoAllocations, uintptr(ptr))`

fritzbauer:
I reported [WASI gc=conservative frees data which is still malloc'ed by cgo code #2187](https://github.com/tinygo-org/tinygo/issues/2187) which describes the issue I was facing. This PR is my proposal to resolve the issue.

Tests are passing except for the arm & aarch64 ones:
exec: "aarch64-linux-gnu-gcc": executable file not found in $PATH
exec: "arm-linux-gnueabihf-gcc": executable file not found in $PATH

I do not have the arm & aarch64 toolchains installed on my system and
could not find any build targets to create them for the tinygo tests,
therefore I skipped them.

Thank you!

### DIFF
--- a/src/runtime/arch_tinygowasm.go
+++ b/src/runtime/arch_tinygowasm.go
@@ -24,10 +24,10 @@ func wasm_memory_size(index int32) int32
 func wasm_memory_grow(index int32, delta int32) int32
 
 var (
-	heapStart    = uintptr(unsafe.Pointer(&heapStartSymbol))
-	heapEnd      = uintptr(wasm_memory_size(0) * wasmPageSize)
-	globalsStart = uintptr(unsafe.Pointer(&globalsStartSymbol))
-	globalsEnd   = uintptr(unsafe.Pointer(&heapStartSymbol))
+	heapStart       = uintptr(unsafe.Pointer(&heapStartSymbol))
+	heapEnd         = uintptr(wasm_memory_size(0) * wasmPageSize)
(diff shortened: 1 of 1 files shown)


aykevl:
See https://github.com/tinygo-org/tinygo/issues/2187#issuecomment-948565530.

fritzbauer:
This was addressed with https://github.com/tinygo-org/tinygo/pull/3148

PR #2194, opened 2021-10-20
Implement realloc for WASI cgo

fritzbauer:
Using wasi-libc from wasi-sdk 12.0 I experienced
a panic because realloc is not implemented yet.
Not sure if the current wasi-libc shipped with tinygo
was patched to avoid realloc, otherwise this
implementation might help to be prepared for newer
versions of wasi-libc.

### DIFF
--- a/src/runtime/arch_tinygowasm.go
+++ b/src/runtime/arch_tinygowasm.go
@@ -84,8 +84,19 @@ func libc_calloc(nmemb, size uintptr) unsafe.Pointer {
 
 //export realloc
 func libc_realloc(ptr unsafe.Pointer, size uintptr) unsafe.Pointer {
-	runtimePanic("unimplemented: realloc")
-	return nil
+	newMalloc := libc_malloc(size)
+	newMallocPtr := uintptr(newMalloc)
+	oldPtr := uintptr(ptr)
+
+	var i uintptr
+	for i = 0; i < size; i++ {
+		//unsafe.Add() only available for go>=1.17; can be adjusted later
+		//*(*byte)(unsafe.Add(newMalloc,i)) = *(*byte)(unsafe.Add(ptr,i))
+		*(*byte)(unsafe.Pointer(newMallocPtr + i)) = *(*byte)(unsafe.Pointer(oldPtr + i))
+	}
+
+	libc_free(ptr)
+	return newMalloc
 }
 
 //export posix_memalign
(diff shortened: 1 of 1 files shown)


src/runtime/arch_tinygowasm.go
```diff
@@ -84,8 +84,23 @@ func libc_calloc(nmemb, size uintptr) unsafe.Pointer {
 
 //export realloc
 func libc_realloc(ptr unsafe.Pointer, size uintptr) unsafe.Pointer {
-	runtimePanic("unimplemented: realloc")
-	return nil
+	newMalloc := libc_malloc(size)
+	if newMalloc == nil || ptr == nil {
+		return newMalloc
+	}
```
aykevl, line 90: Please call `alloc` directly. It won't return a nil pointer.

```suggestion
	newMalloc := alloc(size)
```

src/runtime/arch_tinygowasm.go
```diff
@@ -84,8 +84,23 @@ func libc_calloc(nmemb, size uintptr) unsafe.Pointer {
...
+
+	newMallocPtr := uintptr(newMalloc)
+	oldPtr := uintptr(ptr)
+
+	var i uintptr
+	for i = 0; i < size; i++ {
+		//unsafe.Add() only available for go>=1.17; can be adjusted later
+		//*(*byte)(unsafe.Add(newMalloc,i)) = *(*byte)(unsafe.Add(ptr,i))
+		*(*byte)(unsafe.Pointer(newMallocPtr + i)) = *(*byte)(unsafe.Pointer(oldPtr + i))
+	}
```
aykevl, line 100: There is a function for this available in the runtime: `memcpy`.

aykevl (commented):
There is a big problem with this PR: it doesn't know the size of the old allocation and therefore doesn't know how many bytes to copy. This is known by some heap allocations (-gc=conservative in particular) but not by all. I think a realloc has to know the size of the old allocation to properly function. So perhaps it's best to add this feature to the supported GCs and panic if it isn't available?

src/runtime/arch_tinygowasm.go
fritzbauer: I saw that -gc=conservative never returns nil, but was not sure if it can be assumed for other gcs. 

Regarding ptr I was trying to stick to POSIX and the musl implementation:
> If ptr is a null pointer, realloc() shall be equivalent to malloc() for the specified size.
https://pubs.opengroup.org/onlinepubs/9699919799/

fritzbauer:
Not knowing the size is a fair point. Storing the allocations including their size as suggested in #2187 would help, but I agree this adds more memory consumption. One might argue that webassembly is not as limited in memory as µCs are.

I do not mind to close this PR and be patient until it gets properly implemented in the gc.
Thank you for reviewing!

Issue #2190, opened 2021-10-20
Dividing by zero should panic


learnforpractice:
Giving the following code
```go
package main

import "os"

func main() {
	a := 0
	if len(os.Args) == 1000 {
		a = 1 //never goes here， just to make sure the code can be compiled
	}
	println(1/a)
}
```

Compiling the code with go compiler and  run it, it will throw the following error

```
panic: runtime error: integer divide by zero
```

But after compiling it with tinygo, it just outputs "1".

aykevl:
You are correct, the spec says that a divide by zero must cause a runtime panic. Right now we use whatever LLVM does, which matches C (meaning, it is undefined behavior to divide by zero). This will need to be fixed.

A related issue is signed integer divide overflow, when you divide int32(-2147483648) by -1 this is undefined behavior in C and LLVM but defined for Go: it must result in the number -2147483648.

learnforpractice:
Yes, the same behavior in clang.

aykevl:
I've made a fix here: https://github.com/tinygo-org/tinygo/pull/2195
The code size impact of it is surprisingly small.

Thank you for reporting this issue!

learnforpractice:
Great! It was amazing. You solved this issue so quickly.

deadprogram:
Released with 0.21 so now closing. Thank you!

PR #2195, opened 2021-10-20
compiler: properly implement div and rem operations

aykevl:
The division and remainder operations were lowered directly to LLVM IR.
This is wrong however because the Go specification defines exactly what
happens on a divide by zero or signed integer overflow and LLVM IR
itself treats those cases as undefined behavior. Therefore, this commit
implements divide by zero and signed integer overflow according to the
Go specification.

This does have an impact on the generated code, but it is surprisingly
small. I've used the drivers repo to test the code before and after, and
to my surprise most driver smoke tests are not changed at all. Those
that are, have only a small increase in code size. At the same time,
this change makes TinyGo more compliant to the Go specification.

This fixes https://github.com/tinygo-org/tinygo/issues/2190.


deadprogram:
Please resolve merge conflict. Thanks!

aykevl:
Rebased.

deadprogram:
Thanks @aykevl for the fix. Merging.

PR #2198, opened 2021-10-22
Added realloc implementation to GCs

fritzbauer:
When using the latest wasi-libc I experienced a panic on an attempt to call realloc. My first attempt to add it to arch_tinygowasm.go was not good (PR #2194) for obvious reasons. 
So here is another suggestion.    

Somehow I got motivated to help. Please stop me, if it is more annoying than helping.

### DIFF
--- a/src/runtime/gc_conservative.go
+++ b/src/runtime/gc_conservative.go
@@ -341,6 +341,28 @@ func alloc(size uintptr, layout unsafe.Pointer) unsafe.Pointer {
 	}
 }
 
+func realloc(ptr unsafe.Pointer, size uintptr) unsafe.Pointer {
+	if ptr == nil {
+		return alloc(size, nil)
+	}
+
+	ptrAddress := uintptr(ptr)
+	endOfTailAddress := blockFromAddr(ptrAddress).findNext().address()
+
+	// this might be a few bytes longer than the original size of
+	// ptr, because we align to full blocks of size bytesPerBlock
+	oldSize := endOfTailAddress - ptrAddress
+	if size <= oldSize {
+		return ptr
+	}
+
+	newAlloc := alloc(size, nil)
+	memcpy(newAlloc, ptr, oldSize)
+	free(ptr)
+
+	return newAlloc
+}
+
 func free(ptr unsafe.Pointer) {
 	// TODO: free blocks on request, when the compiler knows they're unused.
 }
--- a/src/runtime/gc_leaking.go
+++ b/src/runtime/gc_leaking.go
@@ -35,6 +35,18 @@ func alloc(size uintptr, layout unsafe.Pointer) unsafe.Pointer {
 	return unsafe.Pointer(addr)
 }
 
+func realloc(ptr unsafe.Pointer, size uintptr) unsafe.Pointer {
+	newAlloc := alloc(size, nil)
+	if ptr == nil {
+		return newAlloc
+	}
+	// according to POSIX everything beyond the previous pointer's
+	// size will have indeterminate values so we can just copy garbage
+	memcpy(newAlloc, ptr, size)
+
+	return newAlloc
+}
+
 func free(ptr unsafe.Pointer) {
 	// Memory is never freed.
 }
--- a/src/runtime/arch_tinygowasm.go
+++ b/src/runtime/arch_tinygowasm.go
@@ -84,8 +84,7 @@ func libc_calloc(nmemb, size uintptr) unsafe.Pointer {
 
 //export realloc
 func libc_realloc(ptr unsafe.Pointer, size uintptr) unsafe.Pointer {
-	runtimePanic("unimplemented: realloc")
-	return nil
+	return realloc(ptr, size)
 }
 
 //export posix_memalign
--- a/src/runtime/gc_extalloc.go
+++ b/src/runtime/gc_extalloc.go
@@ -630,6 +630,10 @@ func alloc(size uintptr, layout unsafe.Pointer) unsafe.Pointer {
 	}
 }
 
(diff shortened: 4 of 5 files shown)


src/runtime/gc_conservative.go
```diff
@@ -340,6 +340,30 @@ func alloc(size uintptr) unsafe.Pointer {
 	}
 }
 
+//go:noinline
```
aykevl: I don't think `//go:noinline` is needed?

src/runtime/gc_conservative.go
```diff
@@ -340,6 +340,30 @@ func alloc(size uintptr) unsafe.Pointer {
...
+	if ptr == nil {
+		return newAlloc
+	}
+
+	ptrAddress := uintptr(ptr)
+	endOfTailAddress := blockFromAddr(ptrAddress).findNext().address()
+
+	//this might be a few bytes longer than the original size of
+	//ptr, because we align to full blocks of size bytesPerBlock
+	ptrSize := endOfTailAddress - ptrAddress - 1
```
aykevl: This looks like an off-by-one.

```suggestion
	ptrSize := endOfTailAddress - ptrAddress
```

Also, perhaps the name `oldSize` or `objectSize` is a bit clearer than `ptrSize`? It's not the size of the pointer, it's the size of the object that the pointer points to.

src/runtime/gc_leaking.go
```diff
@@ -35,6 +35,18 @@ func alloc(size uintptr) unsafe.Pointer {
...
 }
 
+func realloc(ptr unsafe.Pointer, size uintptr) unsafe.Pointer {
+	newAlloc := alloc(size)
+	if ptr == nil {
+		return newAlloc
+	}
+	// according to POSIX everything beyond the previous pointer's
+	// size will have indeterminate values so we can just copy garbage
+	memcpy(newAlloc, ptr, size)
```
aykevl, line 45: Yeah... I'm afraid that's the best we can do for the leaking "GC".

src/runtime/gc_conservative.go
```diff
@@ -340,6 +340,30 @@ func alloc(size uintptr) unsafe.Pointer {
 	}
 }
 
+//go:noinline
+func realloc(ptr unsafe.Pointer, size uintptr) unsafe.Pointer {
+	newAlloc := alloc(size)
```
aykevl: For the conservative GC, we know the size of the underlying buffer so we can often reuse that buffer. Therefore, I would suggest first calculating the underlying buffer size (like you do below) and do nothing if the buffer is already large enough for the requested size.

aykevl (commented):
Thank you! This looks pretty reasonable, with a few comments.

src/runtime/gc_conservative.go
fritzbauer: Agree to rename it. Not sure about the off-by-one. Afaik block.findNext() returns the next head/free block, which is already outside of the previous allocation. Therefore - 1 was supposed to calculate till the last byte of the allocation.

aykevl: > Not sure about the off-by-one. Afaik block.findNext() returns the next head/free block, which is already outside of the previous allocation.

That's correct.

> Therefore - 1 was supposed to calculate till the last byte of the allocation.

No. That's an off-by-one.
Imagine you have an allocation of two blocks, which will be 32 bytes (assuming a 32-bit system). Let's say it is located at address 1024. Two blocks further will be 1024 + 32 = 1056. If you have only those two addresses and want to calculate the size of the first one, you can subtract the two: 1056 - 1024 = 32.
You may be confused by the fact that addresses start at 0. So the block at address 1024 can be indexed from byte 0 (address 1024) up to and including byte 31 (address 1024+31=1055). The full size is 32 bytes, but offsets 0..31 are inside the allocation.

src/runtime/gc_conservative.go
fritzbauer: Thank you! Will change it :)

src/runtime/gc_conservative.go
fritzbauer: Thank you for the explanation! This helped a lot!

deadprogram:
Hello @fritzbauer thank you for this contribution.

Do you want to squash these commits and rebase against the latest `dev` branch please?

aykevl (approved):

fritzbauer:
Thanks! Done. Unfortunately the tests failed, but seems not related to the changes.

deadprogram:
Perhaps something else changed recently? https://github.com/tinygo-org/tinygo/runs/4457364312?check_suite_focus=true#step:16:143 seems like a "legit" error of some kind.

fritzbauer:
Thanks for the link, this was one of the runners, I did not check. Found https://github.com/tinygo-org/tinygo/commit/f24a93c51d8b565aedfa6edebd3c638c9125f468 which adds the layout parameter to alloc. Wondering if I should add it to this realloc function definition as well @aykevl ?

aykevl:
`runtime.realloc` is only called from the libc function `realloc`, which doesn't provide type information. So it's not necessary to add it there. In contrast, `runtime.alloc` is a compiler intrinsic.

fritzbauer:
Perfect. Squashed again now :)

deadprogram:
Sorry to bother you, but could you please rebase against `dev` just one more time? You need to pickup this commit to pass the CI build: https://github.com/tinygo-org/tinygo/commit/9734f349a37630golang82e8a1fa9ba712023db2c9249e thank you!

fritzbauer:
did not find this commit, just rebased against the latest dev revision and it seems all checks completed successfully now

deadprogram:
Thank you very much @fritzbauer merging!

PR #2200, opened 2021-10-23
add support for CPU interrupts for ESP32-C3

zdima:
Support for CPU interrupts for ESP32-C3


aykevl (approved):

aykevl:
I've also added this commit to the espnet branch, for easy development.

PR #2201, opened 2021-10-24
Fix gen-device-svd to handle 64-bit

zdima:
Fix gen-device-svd to handle 64-bit registers masks.


aykevl:
Thank you!
With some quick testing (nrf, sifive, kendryte, esp) I only found a difference with the Kendryte chips, presumably because that's the only 64-bit MCU that we support. But still, it should fix a bug.

Issue #1961, opened 2021-06-18
todo: go on interface call


fgsch:
This was discussed on slack a while back, but for the record, the code below fails with `/usr/local/Cellar/go/1.16.5/libexec/src/net/http/transport.go:1587:5: todo: go on interface call`:

```go
package main

import (
	"fmt"
	"net/http"
)

func main() {
	fmt.Println(http.Cookie{})
}
```

TinyGo version:

`tinygo version 0.19.0-dev-1913cb76 darwin/amd64 (using go version go1.16.5 and LLVM version 11.0.0)`

Invocation:

`tinygo build -target wasm -o main.wasm main.go`

aykevl:
I looked into this before and I believe my conclusion was that it's easier to finish refactoring the reflect package first.

esimov:
This error is still happening for me. Is there any resolution regarding this issue?

aykevl:
I have implemented this, after refactoring how interfaces are type asserted and interface methods are called: https://github.com/tinygo-org/tinygo/pull/2202

esimov:
Thank you. I've tried to install it again via Homebrew, but I'm afraid is not updated yet, because I'm still receiving the above error.

aykevl:
It's only a PR yet, not yet merged into the development branch even. So no, it's not part of Homebrew yet.

esimov:
All right, thanks.

dgryski:
Support for "go on interfaces" as been merged and tinygo 0.21 was released with this fix.  Closing this bug.  Please re-open if you're still encountering problems.

PR #2202, opened 2021-10-24
compiler: add support for the go keyword on interface methods

aykevl:
Most of this PR is a refactor of interface lowering, to make future changes possible. The result of it is that it is now trivial to add support for the go keyword on interface methods, as requested in https://github.com/tinygo-org/tinygo/issues/1961.

### DIFF
--- a/transform/rtcalls.go
+++ b/transform/rtcalls.go
@@ -109,77 +109,64 @@ func OptimizeReflectImplements(mod llvm.Module) {
 	if implementsSignature.IsNil() {
 		return
 	}
-	interfaceMethod := mod.NamedFunction("runtime.interfaceMethod")
-	if interfaceMethod.IsNil() {
-		return
-	}
-	interfaceImplements := mod.NamedFunction("runtime.interfaceImplements")
-	if interfaceImplements.IsNil() {
-		return
-	}
 
 	builder := mod.Context().NewBuilder()
 	defer builder.Dispose()
 
 	// Get a few useful object for use later.
-	zero := llvm.ConstInt(mod.Context().Int32Type(), 0, false)
 	uintptrType := mod.Context().IntType(llvm.NewTargetData(mod.DataLayout()).PointerSize() * 8)
 
-	defer llvm.VerifyModule(mod, llvm.PrintMessageAction)
-
-	for _, use := range getUses(implementsSignature) {
-		if use.IsACallInst().IsNil() {
+	// Look up the (reflect.Value).Implements() method.
+	var implementsFunc llvm.Value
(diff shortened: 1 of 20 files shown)


aykevl:
Updated to fix merge conflicts and (hopefully) fix the CI failures.

transform/rtcalls.go
```diff
@@ -109,77 +109,64 @@ func OptimizeReflectImplements(mod llvm.Module) {
...
 
 	builder := mod.Context().NewBuilder()
 	defer builder.Dispose()
 
 	// Get a few useful object for use later.
-	zero := llvm.ConstInt(mod.Context().Int32Type(), 0, false)
 	uintptrType := mod.Context().IntType(llvm.NewTargetData(mod.DataLayout()).PointerSize() * 8)
 
-	defer llvm.VerifyModule(mod, llvm.PrintMessageAction)
+	// Look up the (reflect.Value).Implements() method.
```
niaow, line 119: I think I am missing something: why can't we look this function up by name?

niaow:
I think most of this looks fine, just a bit confused (see the question i posted above)

transform/rtcalls.go
aykevl: Because it is named something like `interface:{Align:func:{}{basic:int},AssignableTo:func:{named:reflect.Type}{basic:bool},Bits:func:{}{basic:int},Comparable:func:{}{basic:bool},ConvertibleTo:func:{named:reflect.Type}{basic:bool},Elem:func:{}{named:reflect.Type},Field:func:{basic:int}{named:reflect.StructField},FieldAlign:func:{}{basic:int},Implements:func:{named:reflect.Type}{basic:bool},Key:func:{}{named:reflect.Type},Kind:func:{}{named:reflect.Kind},Len:func:{}{basic:int},Name:func:{}{basic:string},NumField:func:{}{basic:int},NumMethod:func:{}{basic:int},Size:func:{}{basic:uintptr},String:func:{}{basic:string}}.Implements$invoke` and therefore depends on the precise set of methods in src/reflect/type.go.
So yeah, we could check that name directly, but every change to the `reflect.Type` interface would also need a change here. Doable, but kind of a burden.

niaow: Oh, I think I misunderstood what this was doing. Alright.

niaow (approved):

deadprogram:
Thanks @aykevl for working on this and thank you @niaow for review.

Now merging.

deadprogram:
Might have been better to rebase `dev` into this branch first, because after merging the CI is failing in `dev` branch.

@aykevl can you please take a look? Thanks!

PR #2204, opened 2021-10-25
Nano-33-BLE: SoftDevice s140v7 support

ysoldak:
Verified on my Arduino Nano 33 BLE Sense.

Installed `nrfjprog`, soldered [SWD debug wires](https://create.arduino.cc/projecthub/arduinocc/arduino-nano-33-iot-debugging-633ad8) and connected to J-Link EDU debug probe.

In [my branch](https://github.com/ysoldak/bluetooth/tree/nrf-s140v7) of `tinygo-org/bluetooth`
```
nrfjprog -f nrf52 --eraseall
nrfjprog -f nrf52 --program s140_nrf52_7.3.0/s140_nrf52_7.3.0_softdevice.hex
tinygo flash -target nano-33-ble-s140v7 ./examples/advertisement/
```

Reverting back to Arduino's bootloader is as easy as:
```
nrfjprog -f nrf52 --eraseall
nrfjprog -f nrf52 --program ~/Library/Arduino15/packages/arduino/hardware/mbed_nano/2.5.2/bootloaders/nano33ble/bootloader.hex --reset
```
Note: you may need to unplug your device and plug it again, to really power-cycle it.

### DIFF
--- a/targets/nano-33-ble-s140v7.json
+++ b/targets/nano-33-ble-s140v7.json
@@ -1,0 +1,6 @@
+{
+	"inherits": ["nano-33-ble", "nrf52840-s140v7"],
+	"flash-command": "nrfjprog -f nrf52 --sectorerase --program {hex} --reset",
+	"flash-1200-bps-reset": "false",
+	"openocd-interface": "jlink"
+}
--- a/targets/nrf52840-s140v7.ld
+++ b/targets/nrf52840-s140v7.ld
@@ -5,9 +5,10 @@ MEMORY
     RAM (xrw)       : ORIGIN = 0x20000000 + 0x000039c0, LENGTH = 256K - 0x000039c0
 }
 
-_stack_size = 4K;
+_stack_size = 4K + __softdevice_stack;
 
 /* This value is needed by the Nordic SoftDevice. */
 __app_ram_base = ORIGIN(RAM);
+__softdevice_stack = DEFINED(__softdevice_stack) ? __softdevice_stack : 0;
 
 INCLUDE "targets/arm.ld"
--- a/targets/nano-33-ble-sense.json
+++ b/targets/nano-33-ble-sense.json
@@ -1,3 +1,0 @@
-{
-	"inherits": ["nano-33-ble"]
-}
--- a/targets/nrf52840-s140v7.json
+++ b/targets/nrf52840-s140v7.json
@@ -1,4 +1,7 @@
 {
 	"build-tags": ["softdevice", "s140v7"],
-	"linkerscript": "targets/nrf52840-s140v7.ld"
+	"linkerscript": "targets/nrf52840-s140v7.ld",
+	"ldflags": [
+		"--defsym=__softdevice_stack=0x700"
+	]
 }


targets/nano-33-ble-s140v7.json
```diff
@@ -0,0 +1,11 @@
+{
+	"inherits": ["nrf52840", "nrf52840-s140v7"],
+	"build-tags": ["nano_33_ble-s140v7", "nano_33_ble"],
+	"serial": "usb",
+	"flash-method": "command",
+	"flash-command": "nrfjprog -f nrf52 -c 1000 --sectorerase --program {hex} --reset",
+	"msd-volume-name": "JLINK",
```
aykevl: >     "msd-volume-name": "JLINK",

Does this actually work? In my experience, flashing an application using the MSD (mass storage device) method will erase whatever there is on the chip already, including the bootloader and SoftDevice.

targets/nano-33-ble-sense-s140v7.json
```diff
@@ -0,0 +1,3 @@
+{
+	"inherits": ["nano-33-ble-s140v7"]
```
aykevl: Why did you add this alias? I would recommend against aliases to avoid confusion - just use a single ID for this target.

targets/nrf52840-s140v7.ld
```diff
@@ -5,9 +5,10 @@ MEMORY
     RAM (xrw)       : ORIGIN = 0x20000000 + 0x000039c0, LENGTH = 256K - 0x000039c0
 }
 
-_stack_size = 4K;
+_stack_size = 4K + __softdevice_stack;
```
aykevl, line 8: :+1: (though at this point, 4K is perhaps a bit large and could be reduced a bit)

targets/nano-33-ble-sense-s140v7.json
ysoldak: Shall we remove `nano-33-ble-sense.json` then?

targets/nano-33-ble-s140v7.json
ysoldak: I don't fully understand effect of all these options here.
This was copied over from `pca10056.json`
Better remove all "msd"?

targets/nrf52840-s140v7.ld
ysoldak: What would be the recommended value? 3K?

ysoldak: How one shall think when deciding on the stack size?
The board has 256KB SRAM.
https://docs.arduino.cc/hardware/nano-33-ble

targets/nano-33-ble-s140v7.json
ysoldak: Removed

targets/nrf52840-s140v7.ld
ysoldak: Oh, wait. This file is not just about Nano 33 BLE, but generally for all nrf52840...
Anyway, this code was borrowed from `nrf52833-s113v7` and 4K is no problem there.
I'd propose we leave this as is or adjust in `nrf52833-s113v7` too then.

Side note: Shall I add same thing to `nrf52-s132v6` ?

targets/nano-33-ble-sense-s140v7.json
ysoldak: Removed

aykevl: Oh - now I see. Because there was also a `nano-33-ble-sense`.

targets/nano-33-ble-s140v7.json
```diff
@@ -0,0 +1,8 @@
+{
+	"inherits": ["nrf52840", "nrf52840-s140v7"],
+	"build-tags": ["nano_33_ble-s140v7", "nano_33_ble"],
```
aykevl: What I did for other boards, was that I inherited two files: the board target (in this case `nano-33-ble`) and the chip+SoftDevice target (in this case `nrf52840-s140v7`). That means that most board-specific configurations can be kept per board.

Something like this (untested):

```json
{
  "inherits": ["nano-33-ble", "nrf52840-s104v7"],
  "flash-command": "nrfjprog -f nrf52 --sectorerase --program {hex} --reset",
  "openocd-interface": "jlink"
}
```

In addition, note that build tags may not contain dashes.

targets/nrf52.json
```diff
@@ -12,5 +12,5 @@
 		"src/device/nrf/nrf52.s"
 	],
 	"openocd-transport": "swd",
-	"openocd-target": "nrf51"
+	"openocd-target": "nrf52"
```
aykevl: This change is unrelated to the Nano 33 BLE: the Nano 33 BLE is a nrf52840, not a nrf52 (aka nrf52832). Yes, this is a bit confusing, but is also the convention used by Nordic Semiconductors.

Note that in my experience, the nrf51 OpenOCD target will work just fine for nrf52xxx series chips and I have intentionally used it here so that the board files are compatible with older versions of OpenOCD (this is one of those cases where comments in JSON would have been very useful: to explain that there is a good reason for this).

targets/nrf52840-s140v7.ld
aykevl: Yeah this was only a suggestion, the code is fine as it is now. 2K is probably more than enough to run the scheduler and application interrupts (which is what this stack size is for with `-scheduler=tasks`, normal goroutines run on separate stacks) but 4K won't really do harm on a 256K RAM chip.

targets/nano-33-ble-sense-s140v7.json
ysoldak: Once we have comments in target files, we could add info in “nano-33-ble” that it shall be used for “sense” variant too… so we skip an extra almost empty file. Or you prefer I add them back, since I’ve removed them now.

targets/nrf52840.json
```diff
@@ -12,5 +12,5 @@
 		"src/device/nrf/nrf52840.s"
 	],
 	"openocd-transport": "swd",
-	"openocd-target": "nrf51"
+	"openocd-target": "nrf52"
```
aykevl: See my other comment before. This probably doesn't do what you think it does.

targets/nano-33-ble-sense-s140v7.json
aykevl: I'm fine either way. But if you add the `nano-33-ble-sense.json` file back, then there should also be such an alias for the -s140v7 variant.

targets/nano-33-ble-s140v7.json
ysoldak: Thanks, will fix

targets/nrf52.json
ysoldak: Yes, OK, so I roll back this and the sister change and we can comment about this later, when #2214 or similar PR enables us have comments in target files.

ysoldak:
Rebased on dev to trigger a new build

deadprogram:
Thanks @ysoldak for working on this and @aykevl for reviewing.

Now merging.

PR #2209, opened 2021-10-28
ci: increase timeout to 20 minutes

aykevl:
PR #2159 wasn't enough, there are many more places that default to a 10 minute timeout.


deadprogram:
:hourglass_flowing_sand: thanks @aykevl merging.

PR #2210, opened 2021-10-29
Support for compiling c++ source file.

ghost:
This pull request adds support for compiling C++ source files.

See example at "src/examples/callcpp"

Developers can specify cxxflags in target json file like this:

```
{
	"cflags": [
...
	],
	"cxxflags": [
		"-std=c++17"
	],
...
}
```

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -538,6 +538,21 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 			}
 			linkerDependencies = append(linkerDependencies, job)
 		}
+		for _, filename := range pkg.CXXFiles {
+			abspath := filepath.Join(pkg.Dir, filename)
+			flags := make([]string, len(pkg.CFlags)+len(pkg.CXXFlags))
+			copy(flags, pkg.CFlags)
+			copy(flags[len(pkg.CFlags):], pkg.CXXFlags)
+			job := &compileJob{
+				description: "compile CGo C++ file " + abspath,
+				run: func(job *compileJob) error {
+					result, err := compileAndCacheCFile(abspath, dir, flags, config.Options.PrintCommands)
+					job.result = result
+					return err
+				},
+			}
+			linkerDependencies = append(linkerDependencies, job)
+		}
 	}
 
 	// Linker flags from CGo lines:
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -38,6 +38,7 @@ type TargetSpec struct {
 	AutoStackSize    *bool    `json:"automatic-stack-size"` // Determine stack size automatically at compile time.
 	DefaultStackSize uint64   `json:"default-stack-size"`   // Default stack size if the size couldn't be determined at compile time.
 	CFlags           []string `json:"cflags"`
+	CXXFlags         []string `json:"cxxflags"`
 	LDFlags          []string `json:"ldflags"`
 	LinkerScript     string   `json:"linkerscript"`
 	ExtraFiles       []string `json:"extra-files"`
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -51,6 +51,7 @@ type PackageJSON struct {
 	GoFiles  []string
 	CgoFiles []string
 	CFiles   []string
+	CXXFiles []string
 
 	// Dependency information
 	Imports   []string
@@ -72,6 +73,7 @@ type Package struct {
 	Files      []*ast.File
 	FileHashes map[string][]byte
 	CFlags     []string // CFlags used during CGo preprocessing (only set if CGo is used)
+	CXXFlags   []string
 	CGoHeaders []string // text above 'import "C"' lines
 	Pkg        *types.Package
 	info       types.Info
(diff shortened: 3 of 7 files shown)


loader/loader.go
```diff
@@ -51,6 +51,7 @@ type PackageJSON struct {
 	GoFiles  []string
 	CgoFiles []string
 	CFiles   []string
+	CXXFiles   []string
```
aykevl: Please run `go fmt loader/loader.go` to correctly format this file.

compileopts/target.go
```diff
@@ -38,6 +38,7 @@ type TargetSpec struct {
 	AutoStackSize    *bool    `json:"automatic-stack-size"` // Determine stack size automatically at compile time.
 	DefaultStackSize uint64   `json:"default-stack-size"`   // Default stack size if the size couldn't be determined at compile time.
 	CFlags           []string `json:"cflags"`
+	CXXFlags         []string `json:"cxxflags"`
```
aykevl, line 41: I don't see why a `cxxflags` key is necessary?
C flags are important in the target file because they define things like the float ABI. But these flags are also used for C++. I can't think of a reason why you would want to configure C++ flags in a target file.

builder/build.go
```diff
@@ -538,6 +538,21 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 			}
 			linkerDependencies = append(linkerDependencies, job)
 		}
+		for _, filename := range pkg.CXXFiles {
+			abspath := filepath.Join(pkg.Dir, filename)
+			flags := make([]string, len(pkg.CFlags)+len(pkg.CXXFlags))
+			copy(flags, pkg.CFlags)
+			copy(flags[len(pkg.CFlags):], pkg.CXXFlags)
+			job := &compileJob{
+				description: "compile CGo CPP file " + abspath,
```
aykevl: CPP is usually used for the C preprocessor.

```suggestion
				description: "compile CGo C++ file " + abspath,
```

aykevl (commented):
Can you add the example to the Makefile so it gets tested? You can add it here for example:

```diff
 ifneq ($(OS),Windows_NT)
 	$(TINYGO) build -o test.elf -gc=leaking -scheduler=none examples/serial
+	$(TINYGO) build -o test.elf                             examples/callcpp
 endif
```

compileopts/target.go
learnforpractice: `cxxflags` is needed for compiling C++ source code. If you move cxx flags such as `-std=c++17` to `cflags`, you will get the following error when compiling the C source code.
```
error: invalid argument '-std=c++17' not allowed with 'C'
```

learnforpractice:
Thank you! The changes have been made.


About why adding CXXFlags to TargetSpec

cxxflags is needed for compiling C++ source code. If you move cxx flags such as -std=c++17 to cflags, you will get the following error when compiling the C source code.

error: invalid argument '-std=c++17' not allowed with ‘C'

learnforpractice:
To make it more clear, I have created an example to show why CXXFlags is not useless in TargetSpec.

https://github.com/learnforpractice/cxxflags-example <https://github.com/learnforpractice/cxxflags-example>

Sorry about causing confusion. I tried to make the pull request as clean as possible.

learnforpractice:
To make it more clear, I have created an example to show why CXXFlags is not useless in TargetSpec.

https://github.com/learnforpractice/cxxflags-example <https://github.com/learnforpractice/cxxflags-example>

Sorry about causing confusion. I tried to make the pull request as clean as possible.

aykevl:
Actually, I think there is a much better way to do that: using `#cgo CXXFLAGS`. See: https://pkg.go.dev/cmd/cgo

In fact, I think some of the logic in this PR is not correct:

> All the cgo CPPFLAGS and CFLAGS directives in a package are concatenated and used to compile C files in that package. All the CPPFLAGS and CXXFLAGS directives in a package are concatenated and used to compile C++ files in that package. All the CPPFLAGS and FFLAGS directives in a package are concatenated and used to compile Fortran files in that package. All the LDFLAGS directives in any package in the program are concatenated and used at link time. All the pkg-config directives are concatenated and sent to pkg-config simultaneously to add to each appropriate set of command-line flags. 

In other words, CGo normally doesn't mix CFLAGS and CXXFLAGS like is done in this PR.

I would suggest the following:

* Use `#cgo CFLAGS`, `#cgo CPPFLAGS`, and the `cflags` key from the target spec to compile C files.
* Use `#cgo CXXFLAGS`, `#cgo CPPFLAGS`, and (again) the `cflags` key from the target spec to compile C++ files.

This means that you can use a particular version of C++ like this:

```go
// #cgo CXXFLAGS: -std=c++17
import "C"
```

To avoid duplication I think it makes sense to use the `cflags` field in the TargetSpec for both C and C++. The `cflags` field is mostly used for low level details such as the architecture and ABI, which apply to both C and C++.

learnforpractice:
Yeah, that’s more complex than I thought, thanks for pointing that out.

Since CXXFLAGS is not supported by tinygo cgo currently, I need to fix it or waiting for it to be fixed if I want to use it.

deadprogram:
Hello @learnforpractice is this PR something that you are still working on? Thanks!

PR #2214, opened 2021-10-30
targets: use YAML instead of JSON

ysoldak:
This PR tests an idea of using YAML instead of JSON for target files.
Minimal changes required. Content of the target files can stay same, since JSON is subset of YAML.

An attempt was made to automatically convert contents to YAML with `json2yaml` python script and `yq` cli util.
Results were not satisfying, since both tools would unroll arrays to multiple lines, and that looks especially ugly for "emulator" values and unnecessary verbose for "inherit" and "build-tags" values.

So decision was made to keep contents unchanged save several Arduino targets where comments were added to show usefulness of the change.

Please take a look at `arduino-nano.yaml` and `arduino-nano-new.yaml`.
Without comments (taken from respective commit introducing `arduino-nano-new` target) it is unclear what target one shall use, when and why.

Another PR #2222 tests the same idea but allows both JSON and YAML to be used for target files. With intention to move to YAML fully eventually.


main.go
```diff
@@ -1318,7 +1318,7 @@ func main() {
 			return
 		}
 		for _, entry := range entries {
-			if !entry.Mode().IsRegular() || !strings.HasSuffix(entry.Name(), ".json") {
+			if !entry.Mode().IsRegular() || !strings.HasSuffix(entry.Name(), ".yaml") {
 				// Only inspect JSON files.
```
Nerzal: that commands also wants an update :)

ysoldak: No, that’s for `go list` if I understood correctly

ysoldak: Oh, wait, probably you are right, thanks!

ysoldak: Fixed

ysoldak:
Some Arduino target file contents converted to proper YAML to give better taste of how it can look like.

compileopts/target.go
```diff
@@ -118,15 +118,15 @@ func (spec *TargetSpec) load(r io.Reader) error {
...
 // - targets/ directory inside the compiler sources
-// - a relative or absolute path to custom (project specific) target specification .json file;
+// - a relative or absolute path to custom (project specific) target specification .yaml file;
 //   the Inherits[] could contain the files from target folder (ex. stm32f4disco)
-//   as well as path to custom files (ex. myAwesomeProject.json)
+//   as well as path to custom files (ex. myAwesomeProject.yaml)
 func (spec *TargetSpec) loadFromGivenStr(str string) error {
 	path := ""
-	if strings.HasSuffix(str, ".json") {
+	if strings.HasSuffix(str, ".yaml") {
```
bgould, line 126: This comment applies to a lot of this change; I think we should do something like accept both `.json` and `.yaml` as valid extensions, rather than just completely switch from one to the other.

ysoldak: I don't see a point to have mix of .json and .yaml files. That would introduce new code with risk of bugs that is doomed to be removed shortly after. And removal can also bring bugs.

Classic JSON is a subset of YAML and can be parsed by YAML parsers, so having .yaml extension for [classic] .json is not that bad actually.

Reformatting all target files -- that we can do. But since it will require curated, though automatic, conversion, I'd like to save this for the last step when everyone agrees we shall move to YAML and not something else, or not move at all.

I can make the conversion, but let's first decide we go this way at all?

bgould:
My opinion is that I think it might be good to support YAML for targets, but if so it should be implemented in stages and I think it is important to ensure backward compatibility.  Some considerations, as a starting point for discussion:

- If we go and rename all of the target files like this and merge into dev, how will that affect open PRs (i.e. will it cause them to need to be rebased etc)?
- This adds a new upstream dependency to the tinygo compiler (not just a test dependency like some of the others in go.mod). Maybe this is not a big deal, but it should at least be considered, since `encoding/json` exists in the standard library and is maintained by the Go project itself

If we are ok with the upstream dependency and are going to implement this change, might we start with just switching to the YAML parser internally and support both `.yaml` and `.json` extensions?  That won't break the workflow of people with existing custom target files or existing branches/forks.  Additionally I think that might be a lot easier to review.  After switching the parser and confirming that it works in a backwards-compatible way, then we handle converting the target files separately.

ysoldak:
Makes sense, @bgould
I did not really think about existing PRs and how this change would affect them. (Do we have many about target files, anyway?)
My only concern in supporting both .json and .yaml is it can easily become a python2/3 story.
I'm sure we will finally arrive to the bright yaml-only future, but on the way it can be mess and confusion.

As for new upstream dependency, I step down and let current maintainers express their opinion.

OK, I going to make a sister PR with both .json and .yaml supported in parallel. And we shall see if that looks better.

ysoldak:
Abandoning this PR in favour of #2222

PR #2215, opened 2021-10-30
machine/rp2040: add interrupt API

soypat:
I've added a first stub for implementing the interrupt API for the rp2040. 

:warning:  It has a bug I have not been able to fix: I am unable to acquire gpio pin status on the interrupt callback so the callback for now has pin 255 as the placeholder interrupt pin. 

:question: Also noticed my VSCode started editing the build tags from `// +build rp2040` to `//go:build rp2040` not sure why, still builds fine. I can edit this if it is a problem.

### DIFF
--- a/src/machine/machine_rp2040_sync.go
+++ b/src/machine/machine_rp2040_sync.go
@@ -1,0 +1,73 @@
+//go:build rp2040
+// +build rp2040
+
+package machine
+
+import (
+	"device/rp"
+)
+
+// machine_rp2040_sync.go contains interrupt and
+// lock primitives similar to those found in Pico SDK's
+// irq.c
+
+const (
+	// Number of spin locks available
+	_NUMSPINLOCKS = 32
+	// Number of interrupt handlers available
+	_NUMIRQ               = 32
+	_PICO_SPINLOCK_ID_IRQ = 9
+	_NUMBANK0_GPIOS       = 30
+)
+
+// Clears interrupt flag on a pin
+func (p Pin) acknowledgeInterrupt(change PinChange) {
+	ioBank0.intR[p>>3].Set(p.ioIntBit(change))
+}
+
+// Basic interrupt setting via ioBANK0 for GPIO interrupts.
+func (p Pin) setInterrupt(change PinChange, enabled bool) {
+	// Separate mask/force/status per-core, so check which core called, and
+	// set the relevant IRQ controls.
+	switch CurrentCore() {
+	case 0:
+		p.ctrlSetInterrupt(change, enabled, &ioBank0.proc0IRQctrl)
+	case 1:
+		p.ctrlSetInterrupt(change, enabled, &ioBank0.proc1IRQctrl)
+	}
+}
+
+// ctrlSetInterrupt acknowledges any pending interrupt and enables or disables
+// the interrupt for a given IRQ control bank (IOBANK, DormantIRQ, QSPI).
+//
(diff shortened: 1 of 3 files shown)


ysoldak:
AFAIK, "go:build rp2040" is new format, introduced in go 1.17

soypat:
Thanks for the clarification yurii. I just realized I PRed against `release`, fixed.

deadprogram:
Looks correct to me. Anyone have any feedback before I merge?

ysoldak:
I may try and test, but dunno how.
An example would be helpful probably?

soypat:
Here's an example for the Raspberry Pi Pico. Change pins accordingly. `cause` intended to be shorted to `effct`

Should print 5 interrupts messages and then stop running the interrupt after it is set to `nil`

```go
package main
​
import (
	"machine"
	"time"
)
​
const (
	// This pin shall be physically shorted to effct to test out interrupt
	cause = machine.GP2
	// this pin shall be the pin with the enabled interrupt. One may short this pin to 3.3V to cause interrupt.
	effct = machine.GP3
	// This pin will display interrupt trigger
	led = machine.LED
	// This will be the type of event that triggers the interrupt.
	event = machine.PinRising
	// event = machine.PinRising | machine.PinFalling //  One can also OR PinChange's so interrupt triggers on either or.
)
​
func main() {
	var err error
	// Configure pin that triggers interrupt
	cause.Configure(machine.PinConfig{Mode: machine.PinOutput})
	if cause.Get() {
		println("cause was high, may trigger immediate interrupt")
	}
	cause.Low()
​
	effct.Configure(machine.PinConfig{Mode: machine.PinInput})
	// err = machine.Sleep(effct, event)
	// pe(err)
	err = effct.SetInterrupt(event, handler)
	pe(err)
	println("wait for interrupt...")
	sleep(2)
	var count int
	for {
		println("prep signal in t minus 1s")
		sleep(1)
		println("send signal")
		cause.High()
		sleep(2)
		println("signal off")
		cause.Low()
		sleep(2)
		count++
		if count > 5 {
			pe(effct.SetInterrupt(event, nil)) // removes interrupt
		}
	}
}
​
// The star of the show. This happens when interrupt triggers.
func handler(p machine.Pin) {
	led.Set(!led.Get())
	println(p)
}
​
// shorthand second sleep.
func sleep(d int) {
	time.Sleep(time.Duration(d) * time.Second)
}
​
// Print error if present.
func pe(err error) {
	if err != nil {
		println(err.Error())
	}
}
```

src/machine/machine_rp2040_sync.go
```diff
@@ -0,0 +1,73 @@
...
+// +build rp2040
+
+package machine
+
+import (
+	"device/rp"
+)
+
+// machine_rp2040_sync.go contains interrupt and
+// lock primitives similat to thos found in Pico SDK's
```
ysoldak:
```suggestion
// lock primitives similar to those found in Pico SDK's
```

ysoldak:
Works on Nano RP2040 Connect.

Verified with slightly modified example `src/examples/pininterrupt`
- set custom pin for button since board has no `machine.BUTTON`
- added naive debouncing logic

I was shorting GND with respective pin and the board was reliably changing state, indicated by built-in LED.

deadprogram:
Thanks for working on this @soypat and thanks @ysoldak for testing/reviewing.

Now merging.

PR #2216, opened 2021-10-30
all: drop support for LLVM 10

aykevl:
It's a rather old version and it is interfering with #2011, so remove it.


deadprogram:
:scissors: thanks @aykevl about time we did this.

Merging.

PR #2217, opened 2021-10-31
transform (interface): fix merge error from #2202

niaow:
This fixes the failure caused by merging #2202


deadprogram:
Thanks for the fix @niaow now merging.

PR #2218, opened 2021-11-01
Match Clang optsize/minsize attributes

aykevl:
This PR adjusts code size function attributes (`optsize` and `minsize`) to match Clang. I've found that while the code size changes vary a lot, on average there is a slight reduction in code size.

I've kept the default at `-opt=z` (which now more aggressively optimizes for code size) because for where TinyGo is used, I think `-opt=z` is more appropriate. If performance is a concern, users should use `-opt=2` or `-opt=s` instead.


deadprogram:
:mag: thanks @aykevl now merging.

PR #2219, opened 2021-11-01
stacksize: hardcode some more frame sizes for __aeabi_* functions

aykevl:
These functions are defined in compiler-rt in assembly and therefore don't have stack size information. However, they're often called so these missing functions often inhibit stack size calculation.

Example, before:

    $ tinygo build -o test.elf -target=cortex-m-qemu -print-stacks ./testdata/float.go
    function                         stack usage (in bytes)
    Reset_Handler                    unknown, __aeabi_memclr does not have stack frame information
    runtime.run$1                    unknown, __aeabi_dcmpgt does not have stack frame information

After:

    $ tinygo build -o test.elf -target=cortex-m-qemu -print-stacks ./testdata/float.go
    function                         stack usage (in bytes)
    Reset_Handler                    260
    runtime.run$1                    224

### DIFF
--- a/stacksize/stacksize.go
+++ b/stacksize/stacksize.go
@@ -201,7 +201,25 @@ func CallGraph(f *elf.File, callsIndirectFunction []string) (map[string][]*CallN
 	case elf.EM_ARM:
 		knownFrameSizes := map[string]uint64{
 			// implemented with assembly in compiler-rt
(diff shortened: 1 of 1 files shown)


deadprogram:
Looking good :straight_ruler: now merging.

PR #2221, opened 2021-11-02
ci: switch to Github Actions for macOS build

deadprogram:
In similar fashion to https://github.com/tinygo-org/tinygo/pull/2160 this PR switches to use Github Actions for the macOS build. It has the same limitation regarding the "double-zip" of the build artifact.


aykevl (approved):
Looks good to me!
Any reason this is still in draft?

aykevl:
Actually, maybe CircleCI is better for now. Caching in CircleCI is at least a lot more straightforward than in GitHub Actions. For example, I'm afraid that GitHub Actions will rebuild LLVM on every release, just like Azure Pipelines did. Which slows down releases.

aykevl:
It looks like we need this. The 60 minute job timeout also exists for MacOS.

.github/workflows/build-macos.yml
```diff
@@ -0,0 +1,90 @@
+name: CI for tinygo-dev on macOS
+
+on: push
+
+jobs:
+  build-macos:
+    name: Build for macOS
+    runs-on: macos-11
```
aykevl: I'm a little bit worried that this will produce binaries that won't run on older MacOS systems.
Does anyone know how backwards compatible such binaries are? @dgryski maybe?

aykevl:
> It looks like we need this. The 60 minute job timeout also exists for MacOS.

Actually, maybe not. The MacOS executor is just fast enough to build LLVM within 60 minutes.

.github/workflows/build-macos.yml
```diff
@@ -0,0 +1,104 @@
...
+        id: cache-binaryen
+        with:
+          key: binaryen-macos-v1
+          path: build/wasm-opt
+      - name: Build Binaryen
+        if: steps.cache-binaryen.outputs.cache-hit != 'true'
+        run: make binaryen
+      - name: Test TinyGo
+        shell: bash
+        run: make test
```
niaow: Is there a reason you removed the `-short` flag?

.circleci/config.yml
```diff
@@ -99,104 +99,6 @@ commands:
...
-    steps:
-      - checkout
-      - submodules
-      - run:
-          name: "Install dependencies"
-          command: |
-            curl https://dl.google.com/go/go1.17.darwin-amd64.tar.gz -o go1.17.darwin-amd64.tar.gz
-            sudo tar -C /usr/local -xzf go1.17.darwin-amd64.tar.gz
-            ln -s /usr/local/go/bin/go /usr/local/bin/go
-            HOMEBREW_NO_AUTO_UPDATE=1 brew install cmake ninja
```
niaow, line 112: These things should still all be installed unconditionally, as the logic is a bit complicated to which we need

.github/workflows/build-macos.yml
deadprogram: I switched to `macos-10.15` for backward compat.

.github/workflows/build-macos.yml
deadprogram: Added it. Not sure that was ever in the macOS build?

niaow: It was, and it makes this several minutes faster

deadprogram:
All tests green, with LLVM cache complete in 17 min. When binaryen cache is used should be just over 10 minutes total.

deadprogram:
Now merging, thanks @aykevl and @niaow for the reviews and suggestions.

PR #2222, opened 2021-11-03
targets: support yaml

ysoldak:
This PR tests an idea of using YAML in addition to JSON for target files.
Minimal changes required. Content of the target files can stay same and parsed by YAML decoder, since JSON is a subset of YAML.

Please take a look at arduino-nano.yaml and arduino-nano-new.yaml.
Without comments (taken from respective commit introducing arduino-nano-new target) it is unclear what target one shall use, when and why.

This PR is less radical than #2214 allowing both JSON and YAML to be used for target files. With intention to move to YAML fully eventually.


ysoldak:
@bgould please, take a look

ysoldak:
Keeping in sync with dev in hope we have a chance to review this proposal and, if accepted, merge early in next release cycle to have time and catch regressions, if any.

ysoldak:
Some thoughts on why YAML.

I want to:
1) have comments in target files;
2) don't shake the codebase too much.

YAML provides both: drop-in replacement of decoder (that can handle JSON too) and comments are natively supported.

[TOML](https://toml.io/en/) looks good too, but will require more changes since format is totally different.

aykevl:
This looks reasonable to me.
One comment: I think you can remove the `json:"` keys from the target spec because this PR also parses the JSON files with a YAML parser. That would avoid some duplication.

Are there other people who have thoughts about this? Not sure who to tag, @deadprogram what do you think?

ysoldak:
Thanks, makes sense. Fixed.

compileopts/target.go
```diff
@@ -121,15 +121,18 @@ func (spec *TargetSpec) load(r io.Reader) error {
...
 	path := ""
-	if strings.HasSuffix(str, ".json") {
+	if strings.HasSuffix(str, ".json") || strings.HasSuffix(str, ".yaml") {
 		path, _ = filepath.Abs(str)
 	} else {
-		path = filepath.Join(goenv.Get("TINYGOROOT"), "targets", strings.ToLower(str)+".json")
+		path = filepath.Join(goenv.Get("TINYGOROOT"), "targets", strings.ToLower(str)+".yaml")
+		if _, err := os.Stat(path); errors.Is(err, os.ErrNotExist) {
+			path = filepath.Join(goenv.Get("TINYGOROOT"), "targets", strings.ToLower(str)+".json")
+		}
```
aykevl, line 135: What about renaming all files to `.yaml`? YAML is backwards compatible so you could keep the existing JSON files but with a different file extension. Then, in a separate PR, we could mass-format the target JSON/YAML files in a consistent way.

There are tools to mass-rename files, I think you can use the `rename` command for this on Linux.

aykevl (commented):
Also see: https://github.com/tinygo-org/tinygo/pull/3533#issuecomment-1528895352

Sorry I never got back to you. I think this is a good idea, but I think the PR could be made simpler (see the comment below):

soypat:
I'm conflicted; on one side:
- I like comments in JSON, there is a lot of value in that

On the other side, I've made the mistake of using YAML to configure hardware at work... **once**. It is appealing at first sight, no question about it. Reads easy. But the spec is just so loose that you can easily construct YAML that looks OK but that enters a weird parsing edge case that prevents part of your configuration YAML to be parsed *without errors*. 

After several costly issues due to the duplicitous nature of YAML we ended up switching to Go source to model our data as config files. I understand this is not a possibility in this case, so I'd like to propose a middle ground: Keep files as JSON but allow comments, maybe preprocess the JSON file before decoding it, removing comments. Should be easy enough?

Related: https://noyaml.com/

ysoldak:
Several months (and slack threads) later, we decided to settle on a less radical change, namely JSONC (JSON with comments). Next best would be [JSON5](https://json5.org/) and YAML only after them two.

@dankegel
I much prefer jsonc to json5, on the theory that smaller attack surfaces are better.
There are several jsonc's for go, I haven't looked at them, but:
- https://github.com/muhammadmuzzammil1998/jsonc
- https://github.com/tidwall/jsonc
- https://github.com/komkom/jsonc
- https://github.com/akshaybharambe14/go-jsonc

@aykevl
^ exactly, the smaller the better (and comments are the main feature we want, anyway). And if we decide in the future that jsonc is too limiting, we can easily upgrade to json5 or even yaml if we want.

ysoldak:
Closing in favour of #3811

PR #2212, opened 2021-10-30
builder: improve accuracy of the -size=full flag 

aykevl:
This PR improves accuracy of the `-size=full` flag in a big way. Instead of relying on symbol names to figure out by which package symbols belong, it will instead mostly use DWARF debug information (specifically, debug line tables and debug information for global variables) relying on symbols only for some specific things. This is much more accurate: it also accounts for inlined functions.

For example, here is how it looked previously when compiling a personal project:

     code  rodata    data     bss |   flash     ram | package
     1902     333       0       0 |    2235       0 | (bootstrap)
       46     256       0       0 |     302       0 | github
        0     454       0       0 |     454       0 | handleHardFault$string
      154      24       4       4 |     182       8 | internal/task
     2498      83       5    2054 |    2586    2059 | machine
        0      16      24     130 |      40     154 | machine$alloc
     1664      32      12       8 |    1708      20 | main
        0       0       0     200 |       0     200 | main$alloc
     2476      79       0      36 |    2555      36 | runtime
      576       0       0       0 |     576       0 | tinygo
     9316    1277      45    2432 |   10638    2477 | (sum)
    11208       -      48    6548 |   11256    6596 | (all)

And here is how it looks now:

```
   code  rodata    data     bss |   flash     ram | package
------------------------------- | --------------- | -------
   1509       0      12      23 |    1521      35 | (unknown)
    660       0       0       0 |     660       0 | C compiler-rt
     58       0       0       0 |      58       0 | C picolibc
      0       0       0    4096 |       0    4096 | C stack
    174       0       0       0 |     174       0 | device/arm
      6       0       0       0 |       6       0 | device/sam
    598     256       0       0 |     854       0 | github.com/aykevl/ledsgo
    320      24       0       4 |     344       4 | internal/task
   1414      99      24    2181 |    1537    2205 | machine
    726     352      12     208 |    1090     220 | main
   3002     542       0      36 |    3544      36 | runtime
    848       0       0       0 |     848       0 | runtime/volatile
     70       0       0       0 |      70       0 | time
    550       0       0       0 |     550       0 | tinygo.org/x/drivers/ws2812
------------------------------- | --------------- | -------
   9935    1273      48    6548 |   11256    6596 | whole program
```

There are some notable differences:

  * Odd packages like `main$alloc` and `handleHardFault$string` are gone, instead their code is put in the correct package.
  * C libraries and the stack are now included in the list, they were previously part of the `(bootstrap)` pseudo-package or not calculated at all.
[...]

PR #2223, opened 2021-11-03
transform: refactor interrupt lowering

aykevl:
Instead of doing everything in the interrupt lowering pass, generate
some more code in gen-device to declare interrupt handler functions and
do some work in the compiler so that interrupt lowering becomes a lot
simpler.

This has several benefits:

  - Overall code is smaller, in particular the interrupt lowering pass.
  - The code should be a bit less "magical" and instead a bit easier to
    read. In particular, instead of having a magic
    runtime.callInterruptHandler (that is fully written by the interrupt
    lowering pass), the runtime calls a generated function like
    device/sifive.InterruptHandler where this switch already exists in
    code.
  - Debug information is improved. This can be helpful during actual
    debugging but is also useful for other uses of DWARF debug
    information.

For an example on debug information improvement, this is what a
backtrace might look like before this commit:

    Breakpoint 1, 0x00000b46 in UART0_IRQHandler ()
    (gdb) bt
    #0  0x00000b46 in UART0_IRQHandler ()
    #1  <signal handler called>
    [..etc]

Notice that the debugger doesn't see the source code location where it
has stopped.

After this commit, breaking at the same line might look like this:

    Breakpoint 1, (*machine.UART).handleInterrupt (arg1=..., uart=<optimized out>) at /home/ayke/src/github.com/tinygo-org/tinygo/src/machine/machine_nrf.go:200
    200			uart.Receive(byte(nrf.UART0.RXD.Get()))
    (gdb) bt
    #0  (*machine.UART).handleInterrupt (arg1=..., uart=<optimized out>) at /home/ayke/src/github.com/tinygo-org/tinygo/src/machine/machine_nrf.go:200
    #1  UART0_IRQHandler () at /home/ayke/src/github.com/tinygo-org/tinygo/src/device/nrf/nrf51.go:176
    #2  <signal handler called>
    [..etc]

By now, the debugger sees an actual source location for UART0_IRQHandler
(in the generated file) and an inlined function.


aykevl:
Oops, I need to update the tests.

aykevl:
The motivation for this PR is actually #2212. With this PR, the `-size=full` flag increases the number of bytes in the binary that can be attributed to a particular package because interrupt handlers are now correctly counted.

deadprogram:
Any comments from @ardnew @ysoldak @soypat @sago35 or anyone else who is into interrupts?

ysoldak:
Sorry, I'm useless in this code review, but I can test regression on nrf52840 (Nano BLE), AVR (Uno), RP2040 (Nano RP2040) and atsamd21g18a (Nano 33 IoT).

Is this example would be enough to flash on aforementioned boards and try if it still works?
https://github.com/tinygo-org/tinygo/blob/release/src/examples/pininterrupt/pininterrupt.go

soypat:
I'm in pretty much the same spot as Yurii but I can test out on Arduino Uno, Mega and the pico sometime next week

sago35:
I don't have the ability to review the compiler/ and transform/ sources yet either.
All other changes are checked.

sago35:
I think it can be deleted.

https://github.com/tinygo-org/tinygo/blob/afb038e63b8b18d9694e72291e9678d9067fae5c/src/runtime/runtime_k210.go#L157-L159

ysoldak:
Regression tested.

Verified with slightly modified "pininterrupt" example.

Works on:
- Nano 33 BLE (target: nano-33-ble, chip: nrf52840 -> cortex-m4)
- Nano 33 IoT (target: arduino-nano33, chip: atsamd21g18a -> cortex-m0plus)

Does not work on:
- Uno (target: arduino, chip: atmega328p -> avr)
  - no support?
- Nano RP2040 (target: nano-rp2040, chip: rp2040 -> cortex-m0plus)
  - support not merged, see #2215

deadprogram:
Thanks for the reviews, everyone. Now merging.

PR #2225, opened 2021-11-03
targets: match LLVM triple to the one Clang uses

aykevl:
The target triples have to match mostly to be able to link LLVM modules (for CGo). Linking LLVM modules is already possible (the triples are already close enough), but testing becomes much easier when they match exactly.

For macOS, I picked "macosx10.12.0". That's an old and unsupported version, but I had to pick _something_. Clang by default uses "macos10.4.0", which is much older.


deadprogram:
:three: merging. Thanks @aykevl

PR #2226, opened 2021-11-04
transform: add debug info in interface lowering pass

aykevl:
This is fake debug info. It doesn't point to a source location because there is no source location. However, it helps to correctly attribute code size usage to particular packages (with `-size=full`).

I've also updated builder/sizes.go with some debugging helpers.


deadprogram:
:no_entry_sign: :bug: 

Merging, thanks @aykevl

PR #2227, opened 2021-11-04
os: add File.WriteString and File.WriteAt

### LINKED ISSUE #2157: WriteString and WriteAt for *os.File are not implemented
When trying to use [github.com/dlclark/regexp2](https://github.com/dlclark/regexp2), I encounter the following error:
```
os.Stdout.WriteString undefined (type *os.File has no field or method WriteString)
```
I looked at the stdlib of tinygo and noticed that `WriteString` and `WriteAt` are not implemented. `WriteString` seems like a pretty simple feature to implement, but I do not know enough about tinygo to be able to tell for sure.

dgryski:
WriteString just does the simple and and converts the passed string
to a byte-slice.  This can be made zero-copy later with unsafe, if needed.

WriteAt returns ErrNotImplemented, to match Seek() and ReadAt().

Fixes #2157

### DIFF
--- a/src/os/file.go
+++ b/src/os/file.go
@@ -111,6 +111,16 @@ func (f *File) Write(b []byte) (n int, err error) {
 	return
 }
 
+// WriteString is like Write, but writes the contents of string s rather than a
+// slice of bytes.
+func (f *File) WriteString(s string) (n int, err error) {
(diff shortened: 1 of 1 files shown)


deadprogram:
Looks legit, thanks for the fix @dgryski now merging.

PR #2228, opened 2021-11-04
musl: reduce binary file size

aykevl:
With some small tweaks, I managed to reduce binary size by around 4.4kB which is most of the code size increase caused by #2110. I also reduced code size on WASI in the case that the os package is not imported.


deadprogram:
:+1: looks good, thanks @aykevl now merging.

PR #2229, opened 2021-11-05
Fix SPI on stm32f103

JAicewizard:
There was a mistake somewhere in the evolution of the SPI driver for the stm32f103. In the original driver it does exist, but somewhere along the line the shift for the frequency multiplier was left out. I didnt have time to track this down more yesterday, I will today. This fixes the SPI code so that it woks again. This was tested using the WIP sx127x LoRa driver by sending messages over LoRa.

### DIFF
--- a/src/machine/machine_stm32f103.go
+++ b/src/machine/machine_stm32f103.go
@@ -358,7 +358,7 @@ func (spi SPI) getBaudRate(config SPIConfig) uint32 {
 	default:
 		conf |= stm32.SPI_CR1_BR_Div256
 	}
-	return conf
+	return conf << stm32.SPI_CR1_BR_Pos
 }
 
 // Configure SPI pins for input output and clock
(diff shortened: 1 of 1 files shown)


JAicewizard:
Preferably, I would actually turn this into something like the the other stm32 devices which calculate the divisor so tat other frequencies that dont exactly match are also supported by rounding to the nearest matching frequency.

kenbell:
Looks good to me - this is an example on another stm32 where we do the shift: https://github.com/tinygo-org/tinygo/blob/release/src/machine/machine_stm32f407.go#L124

JAicewizard:
This seems like these divider constants ware diferent at first: https://github.com/tinygo-org/tinygo/blob/dccfae485c9c5830e1ec012dfb9eb5af5fd10699/src/device/stm32/stm32f103xx-bitfields.go
And then they ware changed in https://github.com/tinygo-org/tinygo/commit/154c7c691bc5c25ba31b84d2583e833ede881de4 to use the generated constants which are different than the ones they replaced.

kenbell (approved):

deadprogram:
Thanks for the fixes @JAicewizard and to @kenbell for review. Now merging.

PR #2230, opened 2021-11-05
machine/rp2040: refactor PWM code. fix Period calculation

soypat:
Fixes Period calculation. Now has pinpoint accuracy. 

```pwm.Period()``` before could overflow for periods above 8ms. Fixed in new version.

Move ```cpuPeriod()``` to be next to relevant functions.

PTAL!

### DIFF
--- a/src/machine/machine_rp2040_clocks.go
+++ b/src/machine/machine_rp2040_clocks.go
@@ -18,6 +18,12 @@ func CPUFrequency() uint32 {
 	return 125 * MHz
 }
 
+// Returns the period of a clock cycle for the raspberry pi pico in nanoseconds.
+// Used in PWM API.
+func cpuPeriod() uint32 {
+	return 1e9 / CPUFrequency()
+}
+
 // clockIndex identifies a hardware clock
 type clockIndex uint8
 
--- a/src/machine/machine_rp2040_pwm.go
+++ b/src/machine/machine_rp2040_pwm.go
@@ -1,3 +1,4 @@
+//go:build rp2040
 // +build rp2040
 
 package machine
@@ -5,12 +6,13 @@ package machine
 import (
 	"device/rp"
 	"errors"
+	"math"
 	"runtime/volatile"
 	"unsafe"
 )
 
 var (
-	ErrPeriodTooBig = errors.New("period outside valid range 1..4e9ns")
+	ErrBadPeriod = errors.New("period outside valid range 8ns..268ms")
 )
 
 const (
@@ -50,8 +52,21 @@ func getPWMGroup(index uintptr) *pwmGroup {
 	return (*pwmGroup)(unsafe.Pointer(uintptr(unsafe.Pointer(rp.PWM)) + 0x14*index))
 }
 
+// Hardware Pulse Width Modulation (PWM) API
 // PWM peripherals available on RP2040. Each peripheral has 2 pins available for
 // a total of 16 available PWM outputs. Some pins may not be available on some boards.
+//
+// The RP2040 PWM block has 8 identical slices. Each slice can drive two PWM output signals, or
+// measure the frequency or duty cycle of an input signal. This gives a total of up to 16 controllable
+// PWM outputs. All 30 GPIOs can be driven by the PWM block
+//
+// The PWM hardware functions by continuously comparing the input value to a free-running counter. This produces a
+// toggling output where the amount of time spent at the high output level is proportional to the input value. The fraction of
+// time spent at the high signal level is known as the duty cycle of the signal.
+//
+// The default behaviour of a PWM slice is to count upward until the wrap value (\ref pwm_config_set_wrap) is reached, and then
+// immediately wrap to 0. PWM slices also offer a phase-correct mode, where the counter starts to count downward after
+// reaching TOP, until it reaches 0 again.
 var (
 	PWM0 = getPWMGroup(0)
 	PWM1 = getPWMGroup(1)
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/machine/machine_rp2040_pwm.go [hunk 1]
```diff
@@ -256,21 +248,31 @@ func (pwm *pwmGroup) setDivMode(mode uint32) {
...
-	targetPeriod := uint32(period)
-	periodPerCycle := getPeriod()
+func (pwm *pwmGroup) setPeriod(period uint32) {
+	periodPerCycle := cpuPeriod()
 	top := pwm.getWrap()
 	phc := pwm.getPhaseCorrect()
-	_, frac := pwm.getClockDiv()
+	// We first modify TOP if desired period is outside divider threshold bounds.
+	limitPeriod := (1 + phc) * periodPerCycle * (1 + top)
+	switch {
```
soypat: New logic to fix bug mentioned in Slack by Fatih. Now modifies Top() value if period is unreachable by modifying just the clock div. Will test tomorrow.

soypat: Fatih Beşer has confirmed these changes solved the bug!

src/machine/machine_rp2040_pwm.go
```diff
@@ -256,21 +255,34 @@ func (pwm *pwmGroup) setDivMode(mode uint32) {
...
+	//  DIV_INT + DIV_FRAC/16 = cycles / ( (TOP+1) * (CSRPHCorrect+1) )  // DIV_FRAC/16 is always 0 in this equation
 	// where cycles must be converted to time:
 	//  target_period = cycles * period_per_cycle ==> cycles = target_period/period_per_cycle
-	Int := targetPeriod/((1+phc)*periodPerCycle*(1+top)) - frac/16
-	if Int > 0xff {
-		Int = 0xff
-	}
-	pwm.setClockDiv(uint8(Int), 0)
+	rhs := 16 * uint64(period) / uint64((1+phc)*periodPerCycle*(1+top)) // right-hand-side of equation, scaled so frac is not divided
+	Int := (rhs / 16) & 0xff
```
deadprogram: Kinda of a nit-pick, but I would have named this `whole` or something instead of `Int` which is probably only exported due to name collision with `int` which seems like a call for this to be named something else. :wink:

soypat: Your wish is my command :wink: , also good call

src/machine/machine_rp2040_clocks.go
```diff
@@ -18,6 +18,13 @@ func CPUFrequency() uint32 {
 	return 125 * MHz
 }
 
+// Returns the period of a clock cycle for the raspberry pi pico in nanoseconds.
+// Used in PWM API. On RP2040 this number should be
+//  1e9/125MHz == 8 (8 nanoseconds)
+func cpuPeriod() uint32 {
+	return 1e9 / CPUFrequency()
```
aykevl, line 24: This works at exactly 125MHz, because 1e9 / 125e6 happens to be exactly 8. Other frequencies likely won't produce an exact number. I don't know whether there will be a way to change the CPU frequency, but this is something to consider.

src/machine/machine_rp2040_pwm.go
```diff
@@ -94,29 +107,29 @@ func (pwm *pwmGroup) peripheral() uint8 {
...
-	if period > 0xffff_ffff {
-		return ErrPeriodTooBig
-	}
 	if period == 0 {
 		period = 1e5
 	}
-	p.setPeriod(period)
+	// Maximum Period is 268369920ns on rp2040, given by (16*255+15)*8*(1+0xffff)*(1+1)/16
+	// With no phase shift max period is half of this value.
+	if period > 268e9 || period < 8 {
```
aykevl: I think it would be easier to move this check into `setPeriod`. That code already contains the code to calculate the maximum possible period and doing it here can introduce bugs. For example, it doesn't look like this code accounts for phase correct PWM and thus is not complete.

src/machine/machine_rp2040_pwm.go
```diff
@@ -180,6 +194,12 @@ func (p *pwmGroup) SetTop(top uint32) {
 	p.setWrap(uint16(top))
 }
 
+// SetCounter sets counter control register. Max value is 16bit (0xffff).
+// Useful for synchronising two different PWM peripherals.
+func (p *pwmGroup) SetCounter(ctr uint32) {
+	p.CTR.Set(ctr)
+}
+
```
aykevl, line 196: I guess this is analogous to the `Counter()` method?
I guess this is fine. I mostly added `Counter()` for debugging purposes though.

src/machine/machine_rp2040_pwm.go
```diff
@@ -256,21 +253,34 @@ func (pwm *pwmGroup) setDivMode(mode uint32) {
...
 // where cycles is amount of clock cycles per PWM period.
-func (pwm *pwmGroup) setPeriod(period uint64) {
-	targetPeriod := uint32(period)
-	periodPerCycle := getPeriod()
+func (pwm *pwmGroup) setPeriod(period uint32) {
+	periodPerCycle := cpuPeriod()
 	top := pwm.getWrap()
 	phc := pwm.getPhaseCorrect()
-	_, frac := pwm.getClockDiv()
+	// We first modify TOP if desired period is outside divider threshold bounds.
```
aykevl: I did it the other way around for other implementations. I intentionally do not change the divider but instead only change the TOP value, because that's usually far more precise. Why do you try to modify the prescaler instead?

aykevl (commented):
This looks fine, although I have a few comments how it could be improved.

src/machine/machine_rp2040_pwm.go
soypat: The counter method has been used by someone to sync PWM signals. I'd leave this.

aykevl:
My suggestion for this PR is to implement `SetPeriod` slightly differently, namely as follows:

1. calculate the maximum possible period that is supported with the current configuration (without changing any of it)
2. if the requested period is higher than that, return an error
3. set the period

I think this is more precise than changing the prescaler in most cases. The prescaler is initially set in `Configure` and `SetPeriod` can set the period to a fraction of the configured prescaler.

soypat:
@aykevl maybe I did not express myself too well, the prescaler is indispensable to reaching the full range of PWM frequencies. Without it you can leave it at 256 but will not be able to access frequencies below 256*8ns = 2 microseconds, and even for frequencies above and near 2us you will have poor dynamic range of the dutycycle. I propose finding a "best of both worlds" approach:
1. Calculate prescaler for best period match for Top value at 95% (or similar) of MaxTop
2. Adjust Top value to make up for the poor accuracy of prescale
3. Set calculated Prescale from step 1 and calculated top from step 2.

This method lets us have near perfect dynamic range of dutycycle and precise period calculation!

soypat:
Setting Period 200Hz, 50% DC on RP2040:

Before: 
image
After!
image

deadprogram:
Merging so we can get this fix into the next release. Thanks @soypat and @aykevl for working on this!

soypat:
Thank you guys!

PR #2231, opened 2021-11-05
Build programs for Windows

aykevl:
I got _slightly_ distracted. But it actually seems to work! Not yet tested on Windows though...


aykevl:
It works! This PR is now ready for review.

src/runtime/runtime_windows.go
```diff
@@ -0,0 +1,210 @@
...
+
+//go:linkname now time.now
+func now() (sec int64, nsec int32, mono int64) {
+	var time uint64
+	_GetSystemTimeAsFileTime(&time)
+	sec = int64(time / 10_000_000)
+	nsec = int32((time - (uint64(sec) * 10_000_000)) * 100)
+	mono = ticksToNanoseconds(ticks())
+	return
+}
```
sago35: I think it needs to be converted to Unix timestamp.

* time.Now()
    * tinygo run on windows : 2390/11/15 21:41:01
    * go run on windows : 2021/11/15 21:41:01

https://stackoverflow.com/questions/20370920/convert-current-time-from-windows-to-unix-timestamp-in-c-or-c

```suggestion
//go:linkname now time.now
func now() (sec int64, nsec int32, mono int64) {
	var time uint64
	_GetSystemTimeAsFileTime(&time)
	
	// convert from windows to unix time stamp
	time -= 0x019DB1DED53E8000
	
	sec = int64(time / 10_000_000)
	nsec = int32((time - (uint64(sec) * 10_000_000)) * 100)
	mono = ticksToNanoseconds(ticks())
	return
}
```

sago35 (commented):
I'm still trying to figure out this code.
`time.Now()` should be fixed at least.

I've read other codes, but I can't decide if they are good or not.
I have built some programs and confirmed that they work on windows.
I think it is a very good first step.

src/runtime/runtime_windows.go
aykevl: Thanks! I will take a look.

aykevl:
I have updated the branch to fix `time.Now`.

deadprogram:
This is definitely far enough along to let people start trying it out. Great work @aykevl and thanks for helping review @sago35 now merging.

QuLogic:
So I managed to get this [linking against system mingw](https://github.com/QuLogic/tinygo/commit/e981e11100f9cb9c79bfd62a7e57df60929cda7b), but I wonder if this is some optional thing that should be done upstream here?

aykevl:
@QuLogic Hmm, I'm not sure what to think of that change. I understand the desire for using existing packages. But in this case, it might result in different build outputs than what TinyGo would normally produce, which users might not expect. In particular, I believe MinGW usually links directly to msvcrt.dll whereas TinyGo links to the newer (and hopefully more stable) ucrt.dll. But that depends on how MinGW has been built, I believe.
Do all the Windows tests pass on Linux, with the tests running in Wine?

Here is some background: https://mingwpy.github.io/ucrt.html

QuLogic:
> @QuLogic Hmm, I'm not sure what to think of that change. I understand the desire for using existing packages. But in this case, it might result in different build outputs than what TinyGo would normally produce, which users might not expect. In particular, I believe MinGW usually links directly to msvcrt.dll whereas TinyGo links to the newer (and hopefully more stable) ucrt.dll. But that depends on how MinGW has been built, I believe.

Well, I'm not calling MinGW, just passing the `kernel32.lib` and `ucrt.lib` files that were created from the `.a` files that tinygo uses here (and that your link says to use in general.)

> Do all the Windows tests pass on Linux, with the tests running in Wine?

Which tests run wine exactly? I guess they are optional, as I didn't seem to need wine when running `go test`/`make tinygo-test`/`make smoketest`. Is `main_test.go` missing a `t.Run` for Windows?

The single `test.exe` that is created in `make smoketest` _does_ run in Wine. It appears to be linked against the expected CRT files:

```
$ llvm-readobj --coff-imports test.exe

File: test.exe
Format: COFF-x86-64
Arch: x86_64
AddressSize: 64bit
Import {
  Name: KERNEL32.dll
  ImportLookupTableRVA: 0x5AB4
  ImportAddressTableRVA: 0x5B4C
  Symbol: GetLastError (630)
  Symbol: GetProcAddress (710)
  Symbol: LoadLibraryExW (990)
  Symbol: SetLastError (1334)
  Symbol: VirtualAlloc (1486)
}
Import {
  Name: api-ms-win-crt-string-l1-1-0.dll
  ImportLookupTableRVA: 0x5AE4
  ImportAddressTableRVA: 0x5B7C
  Symbol: memset (117)
  Symbol: strcpy (122)
  Symbol: strlen (125)
}
Import {
  Name: api-ms-win-crt-private-l1-1-0.dll
  ImportLookupTableRVA: 0x5B04
  ImportAddressTableRVA: 0x5B9C
  Symbol: memcpy (1144)
}
Import {
  Name: api-ms-win-crt-runtime-l1-1-0.dll
  ImportLookupTableRVA: 0x5B14
  ImportAddressTableRVA: 0x5BAC
  Symbol: abort (86)
}
Import {
  Name: api-ms-win-crt-heap-l1-1-0.dll
  ImportLookupTableRVA: 0x5B24
  ImportAddressTableRVA: 0x5BBC
  Symbol: malloc (26)
}
Import {
  Name: api-ms-win-crt-stdio-l1-1-0.dll
  ImportLookupTableRVA: 0x5B34
  ImportAddressTableRVA: 0x5BCC
  Symbol: putchar (127)
  Symbol: puts (128)
}
```

Issue #2234, opened 2021-11-06
nrf52840: floats do not work anymore


ysoldak:
Simple program as below would not work on nrf52840 chip anymore. Affected board is for ex. Nano 33 BLE.
```
package main
import "time"
func main() {
	x := 100.1
	time.Sleep(5 * time.Second)
	for {
		println(x)
		time.Sleep(time.Second)
	}
}
```
Commit that broke things is this:
https://github.com/tinygo-org/tinygo/commit/29206cf0a4a8efcdabe50306090fe6125f6b2d0c

If I remove `"cpu": "cortex-m4",` line from `cortex-m4.json` target file, it works again.

aykevl:
Thanks for the report! I think I know what's wrong and how to fix it. I'll take a look soon.

deadprogram:
Fixed by https://github.com/tinygo-org/tinygo/pull/2235 in cycle, so now closing.

PR #2235, opened 2021-11-06
Set/update target JSON `features` key

aykevl:
This makes sure that the LLVM target features match the one generated by Clang:

  - This fixes a bug (#2234) introduced when setting the target CPU for all targets: Cortex-M4 would now start using floating point operations while they were disabled in C.
  - This will make it possible in the future to inline C functions in Go and vice versa. This will need some more work though.

There is a code size impact. Cortex-M4 targets are increased slightly in binary size while Cortex-M0 targets tend to be reduced a little bit. Other than that, there is little impact.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -136,7 +136,6 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 		DefaultStackSize:   config.Target.DefaultStackSize,
 		NeedsStackObjects:  config.NeedsStackObjects(),
 		Debug:              true,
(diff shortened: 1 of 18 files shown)


ysoldak:
floats work again for nano-33-ble if flashed with tinygo built from this branch.

deadprogram:
Some tests in the CI failing, however.

deadprogram:
Thanks for the quick fix @aykevl and @ysoldak for the quick bug report/verification.

Now merging.

PR #2241, opened 2021-11-07
compiler: add `"target-cpu"` and `"target-features"` attributes

aykevl:
This matches Clang, and with that, it adds support for inlining between Go and C because LLVM only allows inlining if the `"target-cpu"` and `"target-features"` string attributes match.

For example, take a look at the following code:

```go
// int add(int a, int b) {
//   return a + b;
// }
import "C"

func main() {
    println(C.add(3, 5))
}
```

The 'add' function is not inlined into the main function before this commit, but after it, it will be inlined and trivially be optimized to `println(8)`.

The PRs #2220 and #2235 were preparations to make this PR possible.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -23,7 +23,7 @@ import (
 // Version of the compiler pacakge. Must be incremented each time the compiler
 // package changes in a way that affects the generated LLVM module.
 // This version is independent of the TinyGo version number.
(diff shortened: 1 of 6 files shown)


deadprogram:
Seems like this is the commit that completes the trilogy then. Now merging, thanks @aykevl

PR #2245, opened 2021-11-10
ci: improve caching for GitHub Actions

aykevl:
Previously the cache would be stale for every new branch.
With this change, PRs use the cache from the base branch and therefore don't need to rebuild LLVM from scratch.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,6 +1,11 @@
 name: Windows
 
-on: push
+on:
+  pull_request:
+  push:
+    branches:
+      - dev
+      - release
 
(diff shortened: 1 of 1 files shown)


aykevl:
This seems to work: https://github.com/tinygo-org/tinygo/runs/4168639924 did use the cache even though the branch doesn't have any caches yet (I cancelled them all before they could complete).

deadprogram:
That seems promising! I will try that same syntax out in the https://github.com/tinygo-org/tinygo/pull/2221 branch.

deadprogram:
Also, now merging.

Issue #2236, opened 2021-11-06
Failed to compile github.com/hajimehoshi/ebiten/v2/examples/2048


hajimehoshi:
```sh
git clone https://github.com/hajimehoshi/ebiten
cd ebiten
docker pull tinygo/tinygo:latest
docker run --rm -v $(pwd):/src -w /src tinygo/tinygo-dev:latest tinygo build -o wasm.wasm -target=wasm -tags=example ./examples/2048/
```

```
go: downloading golang.org/x/image v0.0.0-20210628002857-a66eb6448b8d
go: downloading golang.org/x/text v0.3.6
panic: interp: load from object with external store

goroutine 9 [running]:
github.com/tinygo-org/tinygo/interp.(*memoryView).load(0xc00cd00270, {0xc000000cef}, 0x4)
        /tinygo/interp/memory.go:273 +0x2a5
github.com/tinygo-org/tinygo/interp.(*runner).run(0xc017c80120, 0xc00a4a19f0, {0xc00d24db10, 0x4, 0x2}, 0xc01ae4e810, {0xc00d069428, 0x8})
        /tinygo/interp/interpreter.go:395 +0x4eae
github.com/tinygo-org/tinygo/interp.(*runner).run(0xc017c80120, 0xc00a4a19a0, {0x0, 0x0, 0x0}, 0x0, {0x7e025d, 0x4})
        /tinygo/interp/interpreter.go:490 +0x8555
github.com/tinygo-org/tinygo/interp.Run({0xc018681dc8}, 0x5f)
        /tinygo/interp/interp.go:112 +0x630
github.com/tinygo-org/tinygo/builder.optimizeProgram({0x7f62ce14d550}, 0xc000070ea0)
        /tinygo/builder/build.go:761 +0x33
github.com/tinygo-org/tinygo/builder.Build.func2(0x0)
        /tinygo/builder/build.go:456 +0x66d
github.com/tinygo-org/tinygo/builder.jobWorker(0x0, 0x0)
        /tinygo/builder/jobs.go:191 +0xad
created by github.com/tinygo-org/tinygo/builder.runJobs
        /tinygo/builder/jobs.go:110 +0x229
```

Zyko0:
Same issue here

aykevl:
I have a fix for this issue (and another issue that came up after that), will make PRs soon.

aykevl:
[...]

PR #2246, opened 2021-11-11
interp: take care of constant globals

aykevl:
Constant globals can't have been modified, even if a pointer is passed
externally. Therefore, don't treat it as such in hasExternalStore.

In addition, it doesn't make sense to update values of constant globals
after the interp pass is finished. So don't do this.

TODO: track whether objects are actually modified and only update the
globals if this is the case.

---

This fixes the `panic: interp: load from object with external store` issue in #2236.


deadprogram:
Thanks @aykevl now merging.

PR #2247, opened 2021-11-11
interp: fix reverting of extractvalue/insertvalue with multiple indices

aykevl:
This is the second bug in https://github.com/tinygo-org/tinygo/issues/2236, uncovered after the first one is fixed using #2246.


deadprogram:

PR #2252, opened 2021-11-11
compiler: fix indices into strings and arrays

aykevl:
This PR fixes two bugs at once:

 1. Indices were incorrectly extended to a bigger type. Specifically,
    unsigned integers were sign extended and signed integers were zero
    extended. This commit swaps them around.
 2. The getelementptr instruction was given the raw index, even if it
    was a uint8 for example. However, getelementptr assumes the indices
    are signed, and therefore an index of uint8(200) was interpreted as
    an index of int8(-56).

---

This bug caused several packages to fail the tests, including math/bits and golang.org/x/image/ccitt.

### DIFF
--- a/compiler/asserts.go
+++ b/compiler/asserts.go
@@ -15,22 +15,16 @@ import (
 // createLookupBoundsCheck emits a bounds check before doing a lookup into a
 // slice. This is required by the Go language spec: an index out of bounds must
 // cause a panic.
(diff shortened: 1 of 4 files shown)


dgryski:
This plus the reflect.DeepEqual() fix means that all the stdlib compress/* package tests can now run and pass with minimal modification.

deadprogram:
Great fix @aykevl thanks! Now merging.

PR #2254, opened 2021-11-12
pass testing arguments to wasmtime

dgryski:
Being paranoid here and only putting the testVerbose etc checks inside the wasmtime if block.  I don't know how other emulators take their arguments.

### DIFF
--- a/main.go
+++ b/main.go
@@ -244,8 +244,18 @@ func runPackageTest(config *compileopts.Config, result builder.BuildResult, test
 		cmd.Dir = result.MainDir
 	} else {
 		// Run in an emulator.
-		// TODO: pass the -test.v flag if needed.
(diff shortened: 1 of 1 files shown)


deadprogram:
You are a much more serious user of wasmtime than I am, so I trust this works as you expect.

Now merging, thanks @dgryski

aykevl:
While I think it's a good change, I think a better (long term) solution is to merge the logic for `tinygo test`, `tinygo run` and the logic in make_test.go for running binaries with parameters (and environment variables).

@dgryski is this something you want to do or something I should take a look at?

See also: https://github.com/tinygo-org/tinygo/pull/2043

dgryski:
I probably won't have time to look at a larger fix :(

PR #2256, opened 2021-11-12
nano-33-ble: internal i2c config, power led and sensors switch

ysoldak:
Adds missing pin definitions to work with onboard sensors

### DIFF
--- a/src/machine/board_nano-33-ble.go
+++ b/src/machine/board_nano-33-ble.go
@@ -1,9 +1,14 @@
+//go:build nano_33_ble
 // +build nano_33_ble
 
 // This contains the pin mappings for the Arduino Nano 33 BLE [Sense] boards.
+// - https://store.arduino.cc/arduino-nano-33-ble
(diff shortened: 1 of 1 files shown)


src/machine/board_nano-33-ble.go
```diff
@@ -72,8 +83,19 @@ const (
...
+
+	// I2C0 (external) pins
+	SDA0_PIN = P0_31
+	SCL0_PIN = P0_02
+
+	// I2C1 (internal) pins
+	SDA1_PIN = P0_14
+	SCL1_PIN = P0_15
+
+	I2C_PULLUP = P1_00 // Set high for I2C to work
```
ysoldak, line 98: This is odd, but required for internal I2C to work.
Extracted from https://github.com/arduino/ArduinoCore-mbed/blob/master/variants/ARDUINO_NANO33BLE/variant.cpp#L159

deadprogram:
Thanks for adding the mapping here @ysoldak now merging.

PR #2260, opened 2021-11-14
all: add support for GOARM

aykevl:
This environment variable can be set to 5, 6, or 7 and controls which ARM version (ARMv5, ARMv6, ARMv7) is used when compiling for GOARCH=arm.

The default value is usually ARMv5. This is an old architecture variant and perhaps it makes sense to use a more recent one such as ARMv6.

We could even go further and support ARMv4 if anybody is interested. It should be pretty simple to add this if needed.

### DIFF
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -5,6 +5,7 @@ package compileopts
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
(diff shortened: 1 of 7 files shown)


compileopts/target.go
```diff
@@ -240,7 +254,14 @@ func defaultTarget(goos, goarch, triple string) (*TargetSpec, error) {
...
 		spec.CPU = "generic"
-		spec.Features = "+armv7-a,+dsp,+fp64,+vfp2,+vfp2sp,+vfp3d16,+vfp3d16sp,-thumb-mode"
+		switch strings.Split(triple, "-")[0] {
+		case "armv5":
+			spec.Features = "+armv5t,+strict-align,-thumb-mode"
+		case "armv6":
+			spec.Features = "+armv6,+dsp,+fp64,+strict-align,+vfp2,+vfp2sp,-thumb-mode"
+		case "armv7":
+			spec.Features = "+armv7-a,+dsp,+fp64,+vfp2,+vfp2sp,+vfp3d16,+vfp3d16sp,-thumb-mode"
+		}
```
niaow, line 269: Should we check for unrecognized options here?

aykevl: Do you mean an unrecognized architecture string (`armv5` etc)?
The only possible values are set in `LoadTarget`: `armv5`, `armv6`, and `armv7`. And the three possible GOARM values are again checked in builder_test.go to have the expected `spec.Features`.

niaow: Oh, I guess I just missed the other check. Nevermind.

niaow (approved):

aykevl:
I changed the PR a bit to default to `GOARM=6` instead of `GOARM=5`. Still not sure which is the better one, but ARMv6 is supported on most single board computers and (I think) all Raspberry Pis. The difference between ARMv5 and ARMv6 is pretty big.

deadprogram:

PR #2262, opened 2021-11-14
wip

chmllr:


aykevl:
As you probably found out, it's not practically possible to use a `GOOS` (or `GOARCH`) that is not supported by the Go toolchain.

I think this use case is best served with `-target=wasi` and a regular package that provides wrappers for the ic0 module.

chmllr:
@aykevl Thanks a lot! Right, I've used this PR to ask a question in the Slack and got the same response. In the end I was able to use `linux/arm` pair by adding a lot of build tags. But the resulting WASM module still had 3 issues:
- imports for `memset`, `memcpy`: no idea where they come from, but I need a clean module, without any imports except those I define in the runtime
- inability to configure the import module name: `//go:wasm-module` does not seem to do anything inside the Go file I compile - all functions were exported with the `env`
- the functions in my newly added runtime did not appear as imports at all, only function inside the Go file I was compiling.

I stopped my experiments for now at that point as I'm not getting anywhere :(

chmllr:
PS: I think I wasn't clear in my reponse: I cannot use `wasi` as it imports WASI syscalls which cannot be implemented on the platform I was trying to add. It's a pity that WASM is not integrated into tinygo as a platform agnostic IR (which it is).

aykevl:
> It's a pity that WASM is not integrated into tinygo as a platform agnostic IR (which it is).

It's not for the same reason there are no platform agnostic x86 binaries that run on both Windows and Linux for example. If you want to do anything useful with them, you have to interact with the environment in some way.

However, I think it is possible to use `-target=wasi` anyway by avoiding any use of WASI. For example, with a very simple Go file:

```go
package main

func main() {
}
```

and by disabling panic messages (which print using WASI):

```
$ tinygo build -o test.wasm -target=wasi -panic=trap test.go
$ wasm-objdump -x -j import test.wasm

test.wasm:	file format wasm 0x1

Section Details:

Section not found: import
```

chmllr:
> If you want to do anything useful with them, you have to interact with the environment in some way.

Correct, there is little value in binaries without calls to the runtime environment. What I've tried to say is that in my naive imagination, I'd first try to provide a platform agnostic WASM compilation (should be actually easier than to integrate WASI) as a basis for the support of arbitrary runtimes defined by the set of their syscalls: WASI, JS, you name it. 

> However, I think it is possible to use -target=wasi anyway by avoiding any use of WASI

Thanks again for your help as it looks very promising! However, I ran the command on the suggested simple Go file, but I get a different output from `wasm-objdump`:

```
$ cat test.go
package main

func main() {
}
$ tinygo build -o test.wasm -target=wasi -panic=trap test.go
$ wasm-objdump -x -j import test.wasm

test.wasm:	file format wasm 0x1

Section Details:

Import[2]:
 - func[0] sig=0 <runtime.args_sizes_get> <- wasi_snapshot_preview1.args_sizes_get
 - func[1] sig=0 <runtime.args_get> <- wasi_snapshot_preview1.args_get
```

I'm using now the version from Homebrew:

```
$ tinygo version
tinygo version 0.20.0 darwin/amd64 (using go version go1.17.2 and LLVM version 11.0.0)
```

Is it possible that you have a different optimization level enabled?

chmllr:
So I played with the `-opt` settings, but as expected it didn't help as `z` is the default anyway. It looks like these 2 imports are used inside `_start`. I'm curious why it's not the case for you.

aykevl:
This was changed in the dev branch: https://github.com/tinygo-org/tinygo/pull/2228/commits/f42d4a7647ee2356b6870fa9d153c2c5020b81e5

chmllr:
Awesome, thanks so much! I'll try to continue my experiments using just the `wasi` target.

I have a general question: in case I _would_ need some modifications to release a Go SDK for the [Internet Computer Protocol](https://internetcomputer.org/) (ICP), how open would you be to add a new target/runtime? What kind of platform establishment metrics do you observe to decide if `tinygo` should support it officially?

The ICP is currently in the top 25 blockchain projects and has a market cap of roughly $9B USD. The eco system is growing rapidly, but there are only two languages right now which can be used on the platform, because only these languages have usable WASM compilers: Rust and Motoko (DFINITY's own language). I work for DFINITY and am a long time fan of Golang. I think it would be a big deal enabling Go support on IC.

Things I have in mind are for example: the IC smart contracts support 2 types of calls: an update call and a query call. The query calls are immutable wrt. the heap, that means the WASM memory changes are just discarded after running a query. To save costs it make sense to disable GC on queries. All of that — I would guess — is not possible with the WASI runtime right now.

chmllr:
FYI it looks like I cannot checkout the dev branch you refer to. When I navigate to the commit I see its contents, but Github says: "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository." Looks like it's a private branch. Do you have an ETA when it will land on the `release` branch?

PR #1798, opened 2021-04-14
PWM Support for atmega1280

MrNinso:
This is the initial pwm support for atmega1280

I need to test this code a little bit more on my arduino mega 1280 before became ready for review, but suggestions and fixes are welcome.

this code is theoretically working on atmega2560, but I don't have an arduino mega 2560, but if anyone can test this I will add it to this pull request.

this code is based in #1121

PR #2264, opened 2021-11-14
Add support for PWM for adruino mega 2560

alexandrevilain:
Hello :) 

I tried to use PWM on Arduino mega 2560 but I found that it was not implemented.

I found PR https://github.com/tinygo-org/tinygo/pull/1798 by @MrNinso saying "this code is theoretically working on atmega2560, but I don't have an Arduino mega 2560" so I copy-pasted the code on `src/machine/machine_atmega2560.go`, compiled it locally, and it works well on my Arduino 2560 R3. 

I don't like copy-pasting such a big part of code, maybe there is a way using build tags that would make support for PWM on both mega 1280 and mega2560 more DRY ? 

I'm pretty new to the tinygo project and in embedded systems ecosystem, so any tips or ideas would be appreciated 😃

### DIFF
--- a/src/machine/machine_atmega2560.go
+++ b/src/machine/machine_atmega2560.go
@@ -4,6 +4,7 @@ package machine
 
 import (
 	"device/avr"
+	"runtime/interrupt"
 	"runtime/volatile"
 )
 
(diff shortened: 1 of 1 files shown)


soypat:
Seems to be working A-OK with the examples/pwm example. 
image

I'd suggest you add `src/examples/pwm/arduino_mega2560.go` file with these contents so it can be run as example!
```go
package main

import "machine"

var (
	pwm  = machine.Timer1
	pinA = machine.PB5
	pinB = machine.PB6
)
```

src/machine/machine_atmega2560.go
```diff
@@ -127,6 +128,799 @@ func (p Pin) getPortMask() (*volatile.Register8, uint8) {
...
+			} else {
+				avr.PORTH.ClearBits(1 << 6) // PH6 low
+				avr.TCCR2A.ClearBits(avr.TCCR2A_COM2B0)
+			}
+		}
+	case 3:
+		// Note: the COM3A0/COM3B0 bit is not set with the configuration below.
+		// It will be set the following call to Set(), however.
+		switch channel {
+		case 0: // channel A, PE3
```
soypat, line 622: Maybe this switch can be collapsed a bit
```go
case 3:
    // Note: the COM3A0/COM3B0 bit is not set with the configuration below.
    if inverting {
        avr.PORTE.SetBits(1<<(channel+3))
    } else {
         avr.PORTE.ClearBits(1<<(channel+3))
    }
```
If I'm not mistaken, in which case the ones below can probably be also collapsed a tad bit.

soypat (approved):
I'm unfamiliar with AVR register convention, but it works and overall looks good to me.

src/machine/machine_atmega2560.go
alexandrevilain: Nice catch! Thanks :) 
As this code is a copy/past from the atmega1280, is there a way to share the same code for both mega1280 and mega2560 ? 
I don't see any build tag that would match both of them.

soypat: They are two separate processors so they have different build tags for now. Maybe we could move the pwm code to a file called `machine_atmega_1280_2560.go` and add both builds tags like so 
```// +build avr,atmega2560 avr,atmega1280```

alexandrevilain:
Hi! It seems that it does not work as expected. 
I'm closing for now, and I'll try to make it work later. 

Thanks for the review @soypat .

PR #2265, opened 2021-11-14
os: Implement and test Mkdir() and Remove()

dkegel-fastly:
While running tinygo test on random packages, discovered that Mkdir() was not yet implemented.

### DIFF
--- a/src/os/os_unix_test.go
+++ b/src/os/os_unix_test.go
@@ -1,0 +1,81 @@
+// +build darwin linux,!baremetal freebsd,!baremetal
+
+package os_test
+
+import (
+	. "os"
+	"strconv"
+	"testing"
+	"time"
+)
+
+func randomName() string {
+	// fastrand() does not seem available here, so fake it
+	ns := time.Now().Nanosecond()
+	pid := Getpid()
+	return strconv.FormatUint(uint64(ns^pid), 10)
+}
+
+func TestMkdir(t *testing.T) {
+	dir := "TestMkdir" + randomName()
+	Remove(dir)
+	err := Mkdir(dir, 0755)
+	defer Remove(dir)
+	if err != nil {
+		t.Errorf("Mkdir(%s, 0755) returned %v", dir, err)
+	}
+	// tests the "directory" branch of Remove
+	err = Remove(dir)
+	if err != nil {
+		t.Errorf("Remove(%s) returned %v", dir, err)
+	}
+}
+
+func writeFile(t *testing.T, fname string, flag int, text string) string {
+	f, err := OpenFile(fname, flag, 0666)
+	if err != nil {
+		t.Fatalf("Open: %v", err)
+	}
+	n, err := f.WriteString(text)
+	if err != nil {
+		t.Fatalf("WriteString: %d, %v", n, err)
+	}
+	f.Close()
+	data, err := ReadFile(f.Name())
+	if err != nil {
+		t.Fatalf("ReadFile: %v", err)
+	}
+	return string(data)
+}
+
+func TestRemove(t *testing.T) {
+	f := "TestRemove" + randomName()
+
+	err := Remove(f)
+	if err == nil {
+		t.Errorf("TestRemove: remove of nonexistent file did not fail")
+	} else {
+		// FIXME: once we drop go 1.15, switch this to fs.PathError
+		if pe, ok := err.(*PathError); !ok {
+			t.Errorf("TestRemove: expected PathError, got err %q", err.Error())
+		} else {
+			if pe.Path != f {
+				t.Errorf("TestRemove: PathError returned path %q, expected %q", pe.Path, f)
+			}
+		}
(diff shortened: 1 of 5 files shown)


src/os/os_unix_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"os"
+	. "os"
+	"testing"
+)
+
+func TestMkdir(t *testing.T) {
+	// FIXME: put test files in unique directory
+	foo := "/tmp/_Test_Mkdir_.tmp"
```
aykevl: I'm worried that this will cause trouble.
I think we should at the very least try to make a unique directory based on the current time, possibly including the PID. Alternatively, crypto/rand could perhaps be used for a unique name.

src/os/os_unix_test.go
```diff
@@ -0,0 +1,57 @@
+// +build darwin linux,!baremetal freebsd,!baremetal
+
+package os_test
+
+import (
+	"os"
+	. "os"
```
aykevl, line 6: Any reason to import os twice?

src/syscall/syscall_libc.go
```diff
@@ -51,11 +51,30 @@ func Open(path string, flag int, mode uint32) (fd int, err error) {
 }
 
 func Mkdir(path string, mode uint32) (err error) {
-	return ENOSYS // TODO
+	data := append([]byte(path), 0)
```
aykevl: I see that existing code also does this.
If you want to improve on that, you could create a new helper function that creates a C string from a Go string with only a single allocation (make a []byte slice of one byte larger and copy the string into it).

src/os/os_unix_test.go
dkegel-fastly: Well.... the 2nd one is forced, but the first one is so I can use os. as a prefix on os functions from the test code.

src/os/os_unix_test.go
dkegel-fastly: I'll implement MkdirTemp and use that, I think.  
First step: https://github.com/tinygo-org/tinygo/pull/2266

dkegel-fastly: No crypto/rand for me, alas:
```
	imports os
	imports crypto/rand
	imports math/big
	imports fmt
	imports os: import cycle not allowed
```
On to door #2.

src/os/os_unix_test.go
aykevl: What do you mean, "forced"?
Either you can `import "os"` and use `os.Mkdir` etc or you can `import . "os"` and call `Mkdir`. Mixing the two seems confusing to me.

src/os/os_unix_test.go
aykevl: Huh, weird. This shouldn't happen: the test package is named os_test.

src/os/os_unix_test.go
dkegel-fastly: Saw it in https://golang.org/src/os/os_test.go and thought it was normal.  Nevermind.

aykevl: Huh, weird. I don't know why they're doing that...

dgryski:
Go itself calls into a simple xorshift rng provided by / used by the runtime: https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/runtime/stubs.go;l=124

src/os/os_unix_test.go
dkegel-fastly: Just using time and pid for now.

src/syscall/syscall_libc.go
dkegel-fastly: done.

src/os/os_unix_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"testing"
+	"time"
+)
+
+func randomName() string {
+	// fastrand() does not seem available here, so fake it
+	ns := time.Now().Nanosecond()
+	pid := Getpid()
+	return fmt.Sprintf("%d", uint(ns^pid))
```
aykevl: Would something slightly less heavy duty like `strconv.Itoa` also work?

aykevl (approved):
LGTM.
EDIT: Looks like you need to fix some merge conflicts.

src/os/os_unix_test.go
dgryski: Or `internal/itoa` ?

As for a small rng, we might want to use `Xorshift32` or `XorshiftMult64` from https://github.com/dgryski/trifles/blob/master/inthash/inthash.go

dkegel-fastly: internal is not allowed in tests, but strconv makes sense.

I'm not sure it's worth bothering with a small rng in this pull request.
A better random filename generator should land along with TempFile et al, though, and could be used instead of this one.

deadprogram:
Thanks @dkegel-fastly for the contribution and @aykevl @dgryski for reviewing.

Now merging.

PR #2266, opened 2021-11-15
Hoist itoa out from src/syscall into src/internal/itoa

dkegel-fastly:
because it would be useful elsewhere in the stdlib tree, too. 
Also switch to the slightly different version from the standard go tree.

Follows in the footsteps of https://github.com/golang/go/commit/061a6903a232cb868780b for tinygo-specific code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -216,6 +216,7 @@ TEST_PACKAGES = \
 	hash/crc64 \
 	html \
 	index/suffixarray \
+	internal/itoa \
 	math \
 	math/cmplx \
 	net/mail \
--- a/loader/goroot.go
+++ b/loader/goroot.go
@@ -228,6 +228,7 @@ func pathsToOverride(needsSyscallPackage bool) map[string]bool {
 		"internal/bytealg/":     false,
 		"internal/reflectlite/": false,
 		"internal/task/":        false,
+		"internal/itoa/":        false, // TODO: Remove when we drop support for go 1.16
 		"machine/":              false,
 		"net/":                  true,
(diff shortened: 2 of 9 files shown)


aykevl:
Why does the internal/itoa package need to be replaced? It seems to be working fine for me without replacing it:

```diff
diff --git a/src/syscall/errno.go b/src/syscall/errno.go
index b67a685c..ec6cc1fb 100644
--- a/src/syscall/errno.go
+++ b/src/syscall/errno.go
@@ -1,5 +1,7 @@
 package syscall
 
+import "internal/itoa"
+
 // Most code here has been copied from the Go sources:
 //   https://github.com/golang/go/blob/go1.12/src/syscall/syscall_js.go
 // It has the following copyright note:
@@ -18,7 +20,7 @@ package syscall
 type Errno uintptr
 
 func (e Errno) Error() string {
-       return "errno " + itoa(int(e))
+       return "errno " + itoa.Itoa(int(e))
 }
 
 func (e Errno) Temporary() bool {
```

In fact the test passes:

```
$ tinygo test internal/itoa
ok  	internal/itoa	0.000s
```

dkegel-fastly:
Oh, good point.  I will try using it as-is.  It is only just dawning on me how this stuff fits together.

dkegel-fastly:
It seems internal/itoa is new to go 1.17, and is not present in go 1.16.
So perhaps we do need to duplicate it for now, if we want to use it (and I would like to use it in os).

aykevl:
Thank you!

PR #2271, opened 2021-11-16
builder: support -size= flag on the esp32

aykevl:
This fixes a small mistake when calculating binary size for an Xtensa file. Previously it would exit with the following error:

    $ tinygo build -o test.elf -size=short -target=esp32-mini32 examples/serial
    panic: runtime error: index out of range [65521] with length 18

Now it runs as expected:

    $ tinygo build -o test.elf -size=short -target=esp32-mini32 examples/serial
       2897       0    4136 |    2897    4136

I'm not sure why it works in CI but not locally. It might be a different linker version (we don't use the LLVM linker for Xtensa).


sago35 (approved):
Looks good to me.
Thanks.

deadprogram:
Thanks for the quick fix @aykevl and the quick testing @sago35 now merging.

PR #2272, opened 2021-11-16
ci: simplify test-linux jobs

aykevl:
Instead of doing lots of repetitive tests in test-llvm11-go115 and test-llvm11-go116, do those tests only once in assert-test-linux and only run smoke tests for older Go versions.

Benefits:
  - This should reduce total CI time, because these jobs don't do tests that are done elsewere anyway. They only do the minimal work necessary to prove that the given Go/LLVM version works.
  - Doing all tests in assert-test-linux hopefully catches bugs that might not be found in regular LLVM builds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,28 +6,6 @@ commands:
       - run:
           name: "Pull submodules"
           command: git submodule update --init
-  apt-dependencies:
-    parameters:
-      llvm:
-        type: string
-    steps:
-      - run:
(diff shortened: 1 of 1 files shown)


aykevl:
Before this PR, the test-llvm11-go115 and test-llvm-go116 jobs were taking about 16-20 minutes to complete.
With this PR, it is down to 10-11 minutes.

deadprogram:
Looking good @aykevl :scissors: thanks! Now merging.

PR #2273, opened 2021-11-16
ci: shorten build-linux job

aykevl:
Split building the release and smoke-testing the release in two, and don't redo some tests that are already done by assert-test-linux.

Some benefits:
  - Lower overall CI time because tests aren't done multiple times.
  - TinyHCI can run earlier because the build-linux job is finished as soon as the build artifact is ready.

It does however have the downside of an extra job, which costs a few seconds to spin up and a few seconds to push and pull the workspace. But even with this, overall CI time is down by a few minutes per workflow run.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -247,18 +247,10 @@ commands:
             sudo apt-get update
             sudo apt-get install --no-install-recommends \
                 libgnutls30 libssl1.0.2 \
-                qemu-system-arm \
-                qemu-user \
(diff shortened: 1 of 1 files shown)


deadprogram:
:hourglass_flowing_sand: thanks @aykevl now merging.

deadprogram:
`build-linux` job of the `dev` branch before this PR build time 12m 39s

`build-linux` job of the `dev` branch with this PR build time 3m 47s

PR #2274, opened 2021-11-16
ci: move all cross compilation tests to Linux

aykevl:
The idea here is as follows:
  - Run all Linux and cross compilation tests in the asser-test-linux job.
  - Only run native tests on MacOS and Windows.

This reduces testing time on MacOS and Windows, which are generally more expensive in CI. Also, by not duplicating tests in Windows and MacOS we can reduce overall CI usage a bit.

The downside is of course that assert-test-linux will take a longer time to run. I've mitigated that slightly by moving the important test (`make test`) above some other less important tests like `make smoketest`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -163,20 +163,17 @@ commands:
           command: |
             sudo apt-get update
             sudo apt-get install --no-install-recommends \
-                libgnutls30 libssl1.0.2 \
                 qemu-system-arm \
+                qemu-system-riscv32 \
                 qemu-user \
                 gcc-avr \
                 avr-libc \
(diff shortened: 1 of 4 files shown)


deadprogram:
`assert-test-linux` already takes a long time to run. This at least frees up some other resources somewhere else.

Now merging, thanks @aykevl

aykevl:
> `assert-test-linux` already takes a long time to run. This at least frees up some other resources somewhere else.

Yes. I plan on making it faster by parallelizing the work, but it's not easy. It basically requires a rewrite of `runJobs` in builder/jobs.go.

PR #2276, opened 2021-11-17
machine/rp2040: add `machine.Sleep` api

soypat:
@aykevl  @deadprogram @ysoldak   PTAL! This is a new API addition to `machine` package. Provides a way to set processor in a low power state until an interrupt is triggered via a PinEvent.

```go
// Sets raspberry pico in dormant state until interrupt event is triggered
// usage:
//  err := machine.Sleep(interruptPin, machine.PinRising)
// Above code will cause program to sleep until a high level signal
// is found on interruptPin. Be sure to configure pin as input beforehand.
// Taken from https://github.com/raspberrypi/pico-extras/ -> pico/sleep.h
func Sleep(pin Pin, change PinChange) error {
```

### DIFF
--- a/src/machine/machine_rp2040_sync.go
+++ b/src/machine/machine_rp2040_sync.go
@@ -20,6 +20,27 @@ const (
 	_NUMBANK0_GPIOS       = 30
 )
 
+// Sets raspberry pico in dormant state until interrupt event is triggered
+// usage:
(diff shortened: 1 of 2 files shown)


soypat:
Superceded by https://github.com/hybridgroup/tinyglobo/blob/main/powman/powman.go. Will be included into tinygo in a separate PR

PR #2277, opened 2021-11-17
wasm: support -scheduler=none

aykevl:
Previously, `-scheduler=none` wasn't possible for WASM targets:

    $ tinygo run -target=wasm -scheduler=none ./testdata/stdlib.go
    src/runtime/runtime_wasm_js.go:34:2: attempted to start a goroutine without a scheduler

With this commit, it works just fine:

    $ tinygo run -target=wasm -scheduler=none ./testdata/stdlib.go
    stdin:  /dev/stdin
    stdout: /dev/stdout
    stderr: /dev/stderr
    pseudorandom number: 1298498081
    strings.IndexByte: 2
    strings.Replace: An-example-string

~It also reduces binary file size a bit.~

Supporting `-scheduler=none` has some benefits:

  * it reduces file size a lot compared to having a scheduler
  * it allows JavaScript to call exported functions

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -164,7 +164,7 @@ func runPlatTests(options compileopts.Options, tests []string, t *testing.T) {
 		t.Parallel()
 		runTest("env.go", options, t, []string{"first", "second"}, []string{"ENV1=VALUE1", "ENV2=VALUE2"})
 	})
-	if options.Target == "wasi" {
+	if options.Target == "wasi" || options.Target == "wasm" {
 		t.Run("alias.go-scheduler-none", func(t *testing.T) {
 			t.Parallel()
 			options := compileopts.Options(options)
--- a/src/runtime/runtime_wasm_js.go
+++ b/src/runtime/runtime_wasm_js.go
@@ -29,34 +29,6 @@ func setEventHandler(fn func()) {
 	handleEvent = fn
 }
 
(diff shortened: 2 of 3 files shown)


src/runtime/runtime_wasm_js_asyncify.go
```diff
@@ -0,0 +1,16 @@
+//go:build wasm && !wasi && scheduler.asyncify
+// +build wasm,!wasi,scheduler.asyncify
+
+package runtime
+
+//export go_scheduler
```
niaow: Both functions are required for both schedulers.

src/runtime/runtime_wasm_js_coroutines.go
```diff
@@ -0,0 +1,20 @@
+//go:build wasm && !wasi && scheduler.coroutines
+// +build wasm,!wasi,scheduler.coroutines
+
+package runtime
+
+//export resume
+func resume() {
```
niaow: Same here: both functions are required for both schedulers.

niaow (changes requested):

aykevl:
Oh, okay, thanks! I've updated the PR with these changes.

niaow (approved):

deadprogram:

PR #2279, opened 2021-11-17
compiler: fix string compare functions

dgryski:
Before:
	x < x false
	x <= x true
	x == x true
	x >= x false
	x > x true

After:
	x < x false
	x <= x true
	x == x true
	x >= x true
	x > x false


dgryski:
`sort` package tests still fail.  Investigating.

dgryski:
Fixed. Remaining `sort` test failures are due to lack of `SkipNow()` support.

aykevl:
Looks like you need to run `go test ./compiler -update` to update the IR.

Also, can you add some tests to testdata/binop.go that show that this bug is fixed?

aykevl (approved):
LGTM.

deadprogram:
Good work on these fixes @dgryski thank you!

Now merging.

PR #2283, opened 2021-11-18
builder: fix off-by-one in size calculation

aykevl:
> There are two hard things in computer science: cache invalidation, naming things, and off-by-one errors.

Because of this bug, sometimes the last object in a section might not be attributed correctly to a source location.

Before:

```
$ tinygo build -size=full -o test ./testdata/alias.go
     33    2046      24      86 |    2103     110 | (unknown)
   2843       8       0     560 |    2851     560 | C musl
    333      24       0       8 |     357       8 | internal/task
     58       6       0       0 |      64       0 | main
   1778     112       8      50 |    1898      58 | runtime
   5045    2196      32     704 |    7273     736 | total
```

After:

```
$ tinygo build -size=full -o test ./testdata/alias.go
     28    2038      20      78 |    2086      98 | (unknown)
   2848      16       4     568 |    2868     572 | C musl
    333      24       0       8 |     357       8 | internal/task
     58       6       0       0 |      64       0 | main
   1778     112       8      50 |    1898      58 | runtime
   5045    2196      32     704 |    7273     736 | total
```

As you can see, all four data kinds (code, rodata, data, bss) are reduced a bit in the `(unknown)` pseudo-package and have the size correctly attributed (in this case, to musl). My goal is of course to make `(unknown)` eventually disappear with everything correctly attributed :)

### DIFF
--- a/builder/sizes.go
+++ b/builder/sizes.go
@@ -96,6 +96,17 @@ const (
 	memoryStack
 )
 
+func (t memoryType) String() string {
+	return [...]string{
+		0:            "-",
+		memoryCode:   "code",
+		memoryData:   "data",
+		memoryROData: "rodata",
+		memoryBSS:    "bss",
+		memoryStack:  "stack",
+	}[t]
+}
+
 // Regular expressions to match particular symbol names. These are not stored as
 // DWARF variables because they have no mapping to source code global variables.
 var (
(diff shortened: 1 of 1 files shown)


niaow (approved):
This looks good to me, but I think it might make sense to try to add some form of testing soon.

niaow:
Actually wait this is on the wrong branch.

aykevl:
I'm not sure how this can be tested. Many different things can change the output. The only thing I can think of would be checking in a binary. Or do you have a better idea?

niaow:
I can't specifically think of a better way to do this than with checked-in binaries. I know that the `debug/pe` tests bundle some binaries for this purpose: https://github.com/golang/go/tree/master/src/debug/pe/testdata, although I suspect we would want something a bit more minimalist. Doesn't block this PR, but at some point it may be worth looking into.

deadprogram:
:1234: merging.

PR #2284, opened 2021-11-18
windows: add support for the -size= flag

aykevl:
Like WebAssembly, it misses information on strings and other anonymous symbols. However, most code (.text) is covered.


niaow (approved):
The size calculation code seems to have gotten a bit big, so I think we should eventually reorganize it. However, I do not think that it needs to be done in this PR. This seems to work.

PR #2285, opened 2021-11-18
all: remove FreeBSD support

aykevl:
FreeBSD support has been broken for a long time, probably since https://github.com/tinygo-org/tinygo/pull/1860 (merged in May). Nobody has complained yet, so I am going to assume nobody uses it.

This doesn't remove support for FreeBSD entirely: the code necessary to build TinyGo on FreeBSD is still there. It just removes the code necessary to build binaries targetting FreeBSD. But again, it could very well be broken as we don't test it.

If anybody wants to re-enable support for FreeBSD, they would be welcome to do that. But I think it would at the very least need a smoke test of some sort.

CC @dmgk


sago35 (approved):

deadprogram:
Agreed, if a new contributor steps up, we can help add actually working support back in.

Now merging, thanks @aykevl and @sago35 for review.

PR #2286, opened 2021-11-18
ci: run stdlib tests on macos

aykevl:


dgryski:
Probably blocked on https://github.com/tinygo-org/tinygo/pull/2184

aykevl:
Hopefully this fixes the issue.

dgryski:
This builds on my mac, but fails with
```
=== RUN   TestOutOfBoundsRead
```

Again, I don't know why Linux doesn't have the same issue.

aykevl:
It could very well be that I made a mistake in my `syscall.Mmap` implementation for MacOS. I didn't test it.

dgryski:
This is passing locally for me now and in CI.

src/syscall/syscall_libc.go
```diff
@@ -80,6 +80,22 @@ func Getenv(key string) (value string, found bool) {
 	}
 }
 
+func Mmap(fd int, offset int64, length int, prot int, flags int) (data []byte, err error) {
+	addr := libc_mmap(nil, uintptr(length), int32(prot), int32(flags), int32(fd), uintptr(offset))
+	if addr == unsafe.Pointer(^uintptr(0)) {
+		return nil, getErrno()
+	}
+	return (*[30]byte)(addr)[:length], nil
```
aykevl: Well there is my bug. I meant to write:

```suggestion
	return (*[1 << 30]byte)(addr)[:length], nil
```

dgryski (approved):
LGTM and CI passes.

aykevl:
Thanks! Looks to me too :)

PR #2289, opened 2021-11-19
ci: move assert-test-linux to GitHub Actions

aykevl:
This is needed since CircleCI has imposed a limit of 60 minutes on the length of a job - even if that happens only infrequently.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,0 +1,110 @@
+name: Linux
+
+on:
+  pull_request:
+  push:
+    branches:
+      - dev
+      - release
+
+jobs:
+  assert-test-linux:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+      - name: Install apt dependencies
+        run: |
+          sudo apt-get install --no-install-recommends \
+              qemu-system-arm \
+              qemu-system-riscv32 \
+              qemu-user \
+              gcc-avr \
+              avr-libc \
+              ninja-build
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.17'
+      - name: Install Node.js
+        uses: actions/setup-node@v2
+        with:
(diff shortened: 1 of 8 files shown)


niaow:
This looks good so far

aykevl:
I'm having some trouble with the wasm tests. They're timing out sometimes.

aykevl:
So I actually forgot to add the `ASSERT=1`, which is the whole point of the assert-test-linux job...

This PR should now be more or less ready. I have a few CI jobs commented out during development because I don't need them, I'll update them once this PR is ready.

My idea is to move at least the Linux jobs over GitHub Actions, but not all at once to make it easier to review and to minimize disruption in the process. (Although `build-linux` and `test-linux-build` will need to be moved at once).

.github/workflows/linux.yml
```diff
@@ -0,0 +1,110 @@
...
+  assert-test-linux:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+      - name: Install apt dependencies
+        run: |
+          sudo apt-get install --no-install-recommends \
```
niaow, line 20: Is apt-get update no longer necessary here?

.github/workflows/windows.yml
```diff
@@ -1,7 +1,7 @@
 name: Windows
 
 on:
-  pull_request:
```
niaow: Was this just temporary for debugging?

aykevl: Yes. I'll remove it before marking this PR as ready.

.github/workflows/linux.yml
aykevl: Apparently not.
And if it is, we can always add it back.

niaow (approved):
This mostly looks good to me. I can't seem to find anything wrong with it now. It does seem a bit slower in a few places, but not by that much.

deadprogram:
Merging. Thanks @aykevl and @niaow

PR #2290, opened 2021-11-19
internal/task: use a lock-free queue

niaow:
This changes task.Stack and task.Queue to use atomic compare-and-swap operations.
With this change, Push can also be executed safely from multiple threads/cores.
Additionally, the queue is actually LIFO now (not sure what I was thinking before).

Not entirely sure about this yet.

### DIFF
--- a/src/internal/task/queue.go
+++ b/src/internal/task/queue.go
@@ -1,69 +1,63 @@
 package task
 
-import "runtime/interrupt"
+import (
+	"sync/atomic"
+	"unsafe"
+)
 
 const asserts = false
 
 // Queue is a FIFO container of tasks.
 // The zero value is an empty queue.
 type Queue struct {
-	head, tail *Task
+	// in is a stack used to buffer incoming tasks.
+	in Stack
+
+	// out is a singly linked list of tasks in oldest-first order.
+	// Once empty, it is refilled by dumping and flipping the input stack.
+	out *Task
 }
 
 // Push a task onto the queue.
+// This is atomic.
 func (q *Queue) Push(t *Task) {
-	i := interrupt.Disable()
-	if asserts && t.Next != nil {
-		interrupt.Restore(i)
-		panic("runtime: pushing a task to a queue with a non-nil Next pointer")
-	}
-	if q.tail != nil {
-		q.tail.Next = t
-	}
-	q.tail = t
-	t.Next = nil
-	if q.head == nil {
-		q.head = t
-	}
-	interrupt.Restore(i)
+	q.in.Push(t)
 }
 
 // Pop a task off of the queue.
+// This cannot be called concurrently.
 func (q *Queue) Pop() *Task {
-	i := interrupt.Disable()
-	t := q.head
-	if t == nil {
-		interrupt.Restore(i)
-		return nil
-	}
-	q.head = t.Next
-	if q.tail == t {
-		q.tail = nil
-	}
-	t.Next = nil
-	interrupt.Restore(i)
-	return t
-}
+	next := q.out
+	if next == nil {
+		// Dump the input stack.
+		s := q.in.dump()
 
-// Append pops the contents of another queue and pushes them onto the end of this queue.
-func (q *Queue) Append(other *Queue) {
-	i := interrupt.Disable()
-	if q.head == nil {
-		q.head = other.head
-	} else {
-		q.tail.Next = other.head
+		// Flip it.
+		var prev *Task
+		for t := s.top; t != nil; {
+			next := t.Next
+			t.Next = prev
+			prev = t
+			t = next
+		}
+		if prev == nil {
+			// The queue is empty.
+			return nil
+		}
+
+		// Save it in the output list.
+		next = prev
 	}
-	q.tail = other.tail
-	other.head, other.tail = nil, nil
-	interrupt.Restore(i)
+
+	q.out = next.Next
+	next.Next = nil
+	return next
 }
 
 // Empty checks if the queue is empty.
+// This cannot be called concurrently with Pop.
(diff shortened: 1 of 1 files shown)


niaow:
Oops, i meant FIFO not LIFO. Also xtensa seems to still need work?

niaow:
Size impact statistics:
- Arduino Uno Blinky (tasks): adds 172 bytes
- Arduino Nano 33 Blinky: adds 84 bytes
- Metro M4 Airlift Blinky: adds 40 bytes
- ESP32-Mini32 Blinky: adds 40 bytes
- Linux Channel Test: adds 76 bytes

src/runtime/runtime_esp8266.go
```diff
@@ -115,3 +117,23 @@ func abort() {
 		device.Asm("waiti 0")
 	}
 }
+
+//export __sync_val_compare_and_swap_4
+func __sync_val_compare_and_swap_4(ptr *uint32, expected, desired uint32) uint32 {
```
aykevl: Huh, interesting. These functions haven't been implemented before:

```
$ build/tinygo build -o test.elf -target=esp8266 ./testdata/atomic.go
/tmp/tinygo787979806/main.o:(.literal.runtime.run$1$gowrapper+0x0): undefined reference to `__sync_fetch_and_add_4'
/tmp/tinygo787979806/main.o:(.literal.runtime.run$1$gowrapper+0x4): undefined reference to `__sync_fetch_and_add_8'
/tmp/tinygo787979806/main.o:(.literal.runtime.run$1$gowrapper+0x8): undefined reference to `__sync_lock_test_and_set_4'
/tmp/tinygo787979806/main.o:(.literal.runtime.run$1$gowrapper+0xc): undefined reference to `__sync_lock_test_and_set_8'
/tmp/tinygo787979806/main.o:(.literal.runtime.run$1$gowrapper+0x10): undefined reference to `__sync_val_compare_and_swap_4'
/tmp/tinygo787979806/main.o:(.literal.runtime.run$1$gowrapper+0x14): undefined reference to `__sync_val_compare_and_swap_8'
/tmp/tinygo787979806/main.o:(.literal.main.testValue+0x0): undefined reference to `sync/atomic.runtime_procPin'
/tmp/tinygo787979806/main.o:(.literal.main.testValue+0x4): undefined reference to `sync/atomic.runtime_procUnpin'
error: failed to link /tmp/tinygo787979806/main: exit status 1
```

If that isn't too difficult, it would be great to have these implemented in a separate PR. Perhaps as a generalization of the functions in src/runtime/arch_cortexm.go - after all, this can be implemented the same way on all unicore processors (which are in effect all MCUs TinyGo supports as TinyGo doesn't use more than one core). I think src/runtime/baremetal.go (or even a new src/runtime/atomic.go) would be a good place for these functions.

src/runtime/arch_avr.go
```diff
@@ -14,3 +17,14 @@ func align(ptr uintptr) uintptr {
 }
 
 func getCurrentStackPointer() uintptr
+
+//export __sync_val_compare_and_swap_2
+func __sync_val_compare_and_swap_2(ptr *uint32, expected, desired uint32) uint32 {
```
aykevl: The same goes for AVR.

Note that atomics are somewhat broken on AVR. If you see weird behavior on AVR, it could be the backend that is emitting invalid instructions.
For more details, see: https://reviews.llvm.org/D97127

src/runtime/runtime_avr.go
```diff
@@ -93,3 +95,12 @@ func abort() {
 		sleepWDT(WDT_PERIOD_2S)
 	}
 }
+
+//export __sync_lock_test_and_set_2
+func __sync_lock_test_and_set_2(ptr *uint16, value uint16) uint16 {
```
aykevl: Whatever the exact location, these two functions should _definitely_ be together in the same file.

aykevl (commented):
Out of curiosity: do you have plans to use the changes in this PR? Any new features this will add?

I haven't looked too deeply into the queue code but I don't see anything surprising with a quick glance.

src/runtime/runtime_avr.go
niaow: Yeah i think this was a mistake

src/runtime/runtime_esp8266.go
niaow: Yeah, I was planning on doing something like that soon.

niaow:
> Out of curiosity: do you have plans to use the changes in this PR?

I am not intending to do anything specific right now, but this might be useful once we want to implement multicore support for something.

> Any new features this will add?

Right now:
1. The queue. . .  is actually a queue instead of a weird mostly-stack
2. On devices with support for atomics, this does not need any critical sections

The Push operation can also be safely called concurrently. . .  but only if the tasks are pinned safely (which is currently an issue, but this is more of a future thing).

src/runtime/runtime_esp8266.go
niaow: Moved into #2307

niaow:
This has been rebased now that #2307 is merged

niaow:
I still need to test this fully on AVR though

niaow:
Now that #2458 and #2459 are in, I consider this ready.

deadprogram:
Seems like this is perhaps not working correctly on the HiFive1b? I tried several times, and also ran some other test code successfully.

jclab-joseph:
@aykevl If this is correct, would you be willing to merge it?
I'm going to try implementing threads again.

aykevl:
I think we can close this now?
I don't see what features it adds or what bugs it fixes, while it does increase binary size.

PR #2291, opened 2021-11-19
targets: change LLVM features to match vanilla Clang

aykevl:
I mistakenly believed the difference was in LLVM version 11.0.0 vs LLVM 11.1.0. However, the difference is in whether we use the Debian version of Clang.

The Debian version has had lots of patches. I'm not sure which is to blame, but it could be this one:
https://salsa.debian.org/pkg-llvm-team/llvm-toolchain/-/blob/snapshot/debian/patches/clang-arm-default-vfp3-on-armv7a.patch

This PR is a preparation for LLVM 12 support.

### DIFF
--- a/builder/builder_test.go
+++ b/builder/builder_test.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
 	"testing"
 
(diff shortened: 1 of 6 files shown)


niaow (approved):
LGTM

aykevl:
Thanks!

PR #2292, opened 2021-11-19
ci: fix Binaryen cache on Windows

aykevl:
The wrong path was used to cache binaryen, so it wasn't actually getting
cached. Therefore, wasm-opt was rebuilt on every new PR (slowing down
the "Build TinyGo release tarball" a lot).

EDIT: in draft to check that it now actually works.


niaow:
Oops. . . :facepalm:

aykevl:
Seems to be working! Now ready for review.

niaow (approved):

PR #2293, opened 2021-11-19
housekeeping: wasm optional in smoketest and fix make clean

kenbell:
make clean was broken because it removed the 'build' directory, which meant the checked-in .gitignore in the build directory was deleted leading to a 'dirty' repo.

This PR:
1. makes WASM optional in `make smoketest` (do WASM=0 to disable)
2. removes .gitignore from build and replaces with rule in parent .gitignore, so `make clean` is 'clean'
3. changes `make binaryen` to create build directory (if not present)
4. adds a few more smoketest output files to .gitignore as a convenience


niaow:
This looks good. Can you rebase?

aykevl:
Also looks good to me.

niaow (approved):

PR #2295, opened 2021-11-19
tinygo: support -run for tests

### LINKED ISSUE #2294: test: add -run flag
Hi,
`-run regex` flag to test command allows to run tests whose names match the `regex`. It can be quite convenient if you have lots of tests and you want to quickly run just a subset of them.

dgryski:
Fixes #2294

### DIFF
--- a/src/testing/testing.go
+++ b/src/testing/testing.go
@@ -18,8 +18,9 @@ import (
 
 // Testing flags.
 var (
-	flagVerbose bool
-	flagShort   bool
+	flagVerbose   bool
+	flagShort     bool
+	flagRunRegexp string
 )
 
 var initRan bool
@@ -33,6 +34,7 @@ func Init() {
 
 	flag.BoolVar(&flagVerbose, "test.v", false, "verbose: print additional output")
 	flag.BoolVar(&flagShort, "test.short", false, "short: run smaller test suite to save time")
+	flag.StringVar(&flagRunRegexp, "test.run", "", "run: regexp of tests to run")
 }
 
 // common holds the elements common between T and B and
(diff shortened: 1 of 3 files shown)


src/testing/testing.go
```diff
@@ -255,7 +258,28 @@ func (m *M) Run() int {
 	}
 
 	failures := 0
+
+	match := func(s string) bool { return true }
+
+	if flagRunRegexp != "" {
+		re, err := regexp.Compile(flagRunRegexp)
+		if err != nil {
+			fmt.Printf("testing: invalid regexp for -test.run: %v", err)
```
aykevl: Shouldn't this print to stderr?
Also, a possibly cheaper alternative is this:

```suggestion
			fmt.Println("testing: invalid regexp for -test.run:", err.Error())
```

Also, looking at that it appears that the line is incorrect anyway: it misses a newline at the end.

aykevl (commented):
Regular expressions already work? I wasn't aware of that.

src/testing/testing.go
dgryski: Yes, regexps work (in general).  There's a Bus Error during `TestOnePassCutoff` I haven't investigated (I think it might be in `Compile` or `Simplify`), and the other test I need to disable is `TestRE2Exhaustive` because it just takes *so* long (and I'm assuming there isn't a bug that's causing an infinite loop -- again, haven't verified this.

The rest of the tests pass.  Although, annoying this PR introduces an import cycle when trying to test the regexp package!  Oh noes!  I wonder if we need a different fix..

aykevl: >  Although, annoying this PR introduces an import cycle when trying to test the regexp package! Oh noes! I wonder if we need a different fix..

Upstream this is solved via the internal/testdeps package. I think we should be using that instead of importing the testing package directly.
(I haven't investigated what the state of internal/testdeps is in TinyGo).

dgryski: As for printing to stderr, I checked this and the standard Go toolchain prints this message to stdout.

deadprogram:
@dgryski please note build failure in Windows: https://github.com/tinygo-org/tinygo/runs/4270476056?check_suite_focus=true#step:13:440

dgryski:
@deadprogram I think the windows failure is a flaky test and unrelated to my change.  Should go away when we rerun CI.

testdata/testing.go
```diff
@@ -34,9 +34,13 @@ var benchmarks = []testing.InternalBenchmark{}
...
+	m := testing.MainStart(testdeps{}, tests, benchmarks, examples)
 	exitcode := m.Run()
 	if exitcode != 0 {
 		println("exitcode:", exitcode)
 	}
 }
+
+type testdeps struct{}
+
+func (testdeps) MatchString(pat, str string) (bool, error) { return true, nil }
```
aykevl, line 46: That... works, I guess, but could lead to unexpected results.

aykevl (approved):
Looks good. However, I would prefer if this actually did pass CI (meaning, it should probably be rebased on the dev branch once #2302 is merged).

testdata/testing.go
dgryski: Well, we have a bit of a problem in that I can't actually import `testing/internal/testdeps` here, so I need to fake it.  I figured this was ok since it was just a test package (and I didn't want to pull in all of regexp just for this.)
 The reason I need to do this at all is because otherwise in the `testing` package I was getting a panic for the unchecked type assertion.  I suppose I could fix that to allow `nil` being passed in here, and warn saying that `-run` support is not available in that case.

dgryski: I'll rebase this on dev once #2303 is merged.

dgryski:
Rebased.  Hopefully CI passes this time.

deadprogram:
Passing CI, so now merging. Thanks @dgryski

PR #2298, opened 2021-11-20
os: implement CreateTemp

dkegel-fastly:
Also:
- make TempDir() usable on Windows
- added the missing Unwrap method on PathError, else errors.Is() fails when called on PathError

Includes a minimally functional stub of fastrand(), which should probably move into runtime and be improved later.


src/os/tempfile.go
```diff
@@ -1,5 +1,94 @@
 package os
 
+import (
+	"errors"
+	"internal/itoa"
+	"time"
+)
+
+// export for testing
+var ErrPatternHasSeparator = errPatternHasSeparator
```
dgryski: Things that are exported for testing only (and not otherwise as part of the public API) are normally put in `export_test.go`.  That file will get included when testing, and ignored otherwise.

src/os/tempfile_test.go
```diff
@@ -0,0 +1,79 @@
...
+			tmpfile, err := CreateTemp(tmpDir, tt.pattern)
+			if tmpfile != nil {
+				defer tmpfile.Close()
+			}
+			if tt.wantErr {
+				if err == nil {
+					t.Errorf("CreateTemp(..., %#q) succeeded, expected error", tt.pattern)
+				}
+				if !errors.Is(err, ErrPatternHasSeparator) {
+					t.Logf("TODO: CreateTemp(..., %#q): %v, expected ErrPatternHasSeparator", tt.pattern, err)
```
dgryski: I have a feeling this is related to reflect and/or the comparable logic.

src/os/tempfile.go
dkegel-fastly: TIL!

dkegel-fastly: Alas, tinygo doesn't seem to support export_test.go.

dgryski: Hmm.. pretty sure I tested this before commenting. Let me double check.

dgryski: WorksForMe(tm)

```
diff --git a/src/os/tempfile.go b/src/os/tempfile.go
index b5452248..c88ec04a 100644
--- a/src/os/tempfile.go
+++ b/src/os/tempfile.go
@@ -6,9 +6,6 @@ import (
        "time"
 )

-// export for testing
-var ErrPatternHasSeparator = errPatternHasSeparator
-
 // fake_fastrand() is a placeholder for fastrand()
 // TODO: provide fastrand in runtime, as go does.  It is hard to implement properly elsewhere.
 // See https://github.com/tinygo-org/tinygo/issues/1037
~/go/src/github.com/tinygo-org/tinygo/src/os $ cat export_test.go
package os

// export for testing
var ErrPatternHasSeparator = errPatternHasSeparator
~/go/src/github.com/tinygo-org/tinygo/src/os $ ../../tinygo clean && ../../tinygo test os
ok  	os	0.373s

```

dkegel-fastly: OMG I have been doing it wrong for months. 

I  have been informed that the way to test code in the src/os directory is not, in fact, to run "tinygo test" in the src/os directory or to pass the name of the test file to tinygo test as one might expect, but instead to run "tinygo test os" anywhere in the source tree.

Running "tinygo test" in the src/os directory somewhat amazingly runs the tests on github.com/tinygo-org/tinygo/src/os rather than on the local files, hence my confusion.

dkegel-fastly:
This only slightly rare failure on Windows,
```
CreateTemp(..., "tempfile_test*") error: open C:\Users\RUNNER~1\AppData\Local\Temp/tempfile_test1772762933: The file exists.
```
indicates that fake_fastrand is too fake.

I made it less fake by adding a minimal actual RNG.

src/os/tempfile_test.go
dkegel-fastly: If this PR is accepted, I'll try to make a minimal test case and file an issue...

dkegel-fastly:
Rebased and squashed a bit.

src/os/tempfile_test.go
dkegel-fastly: Turns out PathError was missing an Unwrap method.  Fixed!

deadprogram:
Yet another merge conflict @dkegel-fastly :smiley_cat:

dkegel-fastly:
Closing in favor of individual pull requests for the three problems this solves.

PR #2300, opened 2021-11-21
Add LLVM 12 support

aykevl:
Originally based on a PR by @QuLogic, but extended a lot to get all tests to pass.

Work in progress. The remaining parts is build-linux, which can't complete its run on CircleCI because of the 1 hour job run limit.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,12 +22,12 @@ commands:
     steps:
       - restore_cache:
           keys:
-            - llvm-source-11-v2
+            - llvm-source-12-v1
       - run:
           name: "Fetch LLVM source"
           command: make llvm-source
       - save_cache:
-          key: llvm-source-11-v2
+          key: llvm-source-12-v1
           paths:
             - llvm-project/clang/lib/Headers
             - llvm-project/clang/include
@@ -118,12 +118,12 @@ commands:
             - go-cache-macos-v3-{{ checksum "go.mod" }}
       - restore_cache:
           keys:
-            - llvm-source-11-macos-v3
+            - llvm-source-12-macos-v1
       - run:
           name: "Fetch LLVM source"
           command: make llvm-source
       - save_cache:
-          key: llvm-source-11-macos-v3
+          key: llvm-source-12-macos-v1
           paths:
             - llvm-project/clang/lib/Headers
             - llvm-project/clang/include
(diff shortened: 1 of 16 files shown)


dgryski:
Aside from coroutine issues, does this fix any issues that were hitting tinygo?

aykevl:
Not directly, but it's important to stay up to date with LLVM. Some reasons:

* Newer versions have improvements and bug fixes. For example, the heap alignment bug for WebAssembly that's fixed in LLVM 13. Another example: the MachO linker becomes usable in LLVM 12 which means we can cross compile executables for MacOS like we already do for Windows and Linux.
* Linux distributions prefer to use only a single (recent) LLVM version

aykevl:
This is now, _finally_, ready for review. It took a lot of work (most importantly, #2011) to get LLVM 12 supported but so far all tests seem to be passing.

deadprogram:
Seems like all TinyHCI tests are passing except MAIt BiT which is our only RISC-V 64-bit platform. Have to dig in a little more here. Program compiles and flashes just fine, but no response from serial.

aykevl:
Hmm, that's kind of difficult for me to debug as I don't have the hardware. I don't see anything obviously wrong with the binary either.
Can you try debugging this locally? For example with a very small program (I usually use `./testdata/alias.go` for this purpose).

One thing you could try is to set the linker to the LLVM 11 linker and see whether that works. You can do this by modifying the maixbit.json file and add

    "linker": "ld.lld-11",

(Or similar, depending on how you've installed LLVM).

deadprogram:
Using a container with this build, I can determine that even blinky1 does not run on the Maixbit with this branch. The container I used for testing did not have LLVM installed, as it uses the build from the CI. I will have to try a different container for this test.

aykevl:
You can also build locally, using `go install`. You will then need to provide the LLVM version in a tag: `go install -tags=llvm11` or `go install -tags=llvm12`. This PR still defaults to LLVM 11 for `go install`.

deadprogram:
I will setup jlink and gdb to try to look into this a bit more. In the meantime, since it is not commonly used apparently, I have disabled that board from TinyHCI in the interest of getting this PR merged sooner.

Also have installed locally, so I will try out with some of my various demos and toys.

deadprogram:
Confirmed that a couple of my Bluetooth examples are working as expected with this branch. As far as I am concerned, we should merge it, and then later look into any specific issues on the Maxibit.

niaow (approved):
LGTM. I tried it out locally and cannot find any issues. Not sure about the maixbit issue.

deadprogram:
Thanks @QuLogic for starting this off and @aykevl for finishing it off. Also thanks @niaow and @dgryski for review.

Now merging!

PR #2302, opened 2021-11-21
ci: move Linux release builds to GitHub Actions

aykevl:
~Work in progress. Making a PR because it won't run otherwise.~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,13 +6,6 @@ commands:
       - run:
           name: "Pull submodules"
           command: git submodule update --init
-  install-cmake:
-    steps:
-      - run:
-          name: "Install CMake"
-          command: |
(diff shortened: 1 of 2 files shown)


aykevl:
This is now ready.

Note that TinyHCI will likely need to be updated to pull the release artifact from GitHub Actions instead of from CircleCI.

The motivation for this PR is to avoid the 1 hour build limit on CircleCI, which we can't stay within when we're building LLVM. Right now we usually stay within that limit, but the cache will eventually expire at which point we have to do something. Also: LLVM 12 support (#2300) needs to rebuild LLVM.

deadprogram:
I have made a number of changes to TinyHCI to accommodate the new workflow. It was a bit more complicated to obtain the build artifacts, and then also process them, which is why this took longer than expected.

Since the TinyHCI tests are now passing, I am going to merge this PR now. Thanks for working on it @aykevl

PR #2303, opened 2021-11-22
ci: cache Go cache for faster build times

aykevl:
This appears to reduce CI time for Windows by 1-2 minutes.


deadprogram:
Less is more. Merging, thanks @aykevl

PR #2306, opened 2021-11-22
os, syscall: implement ReadAt aka Pread

dkegel-fastly:
WIP as I learn how this works on other machines...

Only the last commit is new, the first two are just https://github.com/tinygo-org/tinygo/pull/2298

So don't merge until 2298 lands.


dkegel-fastly:
Premature.  Will reopen later.

PR #2308, opened 2021-11-23
runtime: allow comparing interfaces in reflectValueEqual()

dgryski:
A few tests (such as `crypto/hmac` in the stdlib) panic with `comparing un-comparable type`.  It seems like the runtime is not able to compare interfaces.

This is *a* PR to address this, but I don't know if it's the right one.

### DIFF
--- a/src/runtime/interface.go
+++ b/src/runtime/interface.go
@@ -74,6 +74,8 @@ func reflectValueEqual(x, y reflect.Value) bool {
 			}
 		}
 		return true
+	case reflect.Interface:
+		return reflectValueEqual(x.Elem(), y.Elem())
 	default:
 		runtimePanic("comparing un-comparable type")
 		return false // unreachable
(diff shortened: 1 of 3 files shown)


aykevl:
Yes, I think this is correct. Took me a while to figure it out, because I didn't realize `valueInterfaceUnsafe` actually returns an interface type. Therefore, it's equivalent to the following:

```go
return interfaceEqual(valueInterfaceUnsafe(x), valueInterfaceUnsafe(y))
```

The following might be a little bit simpler - but that probably depends on who's reading the code:

```go
return reflectValueEqual(x.Elem(), y.Elem())
```

Can you add a few tests to testdata/binop.go? It looks like the comparison is something like this:

```go
type foo struct {
    n   int
    itf interface{}
}

// ...
var a = foo{3, float32(5}}
var b = foo{3, float32{3}}
return a == b
```

This code probably doesn't work before this patch but probably does work with this patch (not tested).
I know that the `reflectValueEqual` isn't tested a lot, but let that not be a reason to not add tests for new code!

dgryski:
Fixed compare logic to recurse with `.Elem()`.  Added failing test which is fixed with this patch.  PTAL.

aykevl:
Thank you!

PR #2309, opened 2021-11-23
Multiple improvements for stm32wle targets

ofauchon:
This PR provide multiple improvements for STM32WLE5 targets:

- board/stm32: Add STM32 Nucleo WL55JC board
- board/lorae5,nucleo-wl55jc: Add SubGhz SPI3 interface
- stm32/stm32wl: Set 48Mhz HSE32/PLL as default Clock, SubGhz Radio device init

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ See the [getting started instructions](https://tinygo.org/getting-started/) for 
 
 You can compile TinyGo programs for microcontrollers, WebAssembly and Linux.
 
-The following 72 microcontroller boards are currently supported:
+The following 74 microcontroller boards are currently supported:
 
 * [Adafruit Circuit Playground Bluefruit](https://www.adafruit.com/product/4333)
 * [Adafruit Circuit Playground Express](https://www.adafruit.com/product/3333)
@@ -115,6 +115,7 @@ The following 72 microcontroller boards are currently supported:
 * [ST Micro "Nucleo" L031K6](https://www.st.com/ja/evaluation-tools/nucleo-l031k6.html)
 * [ST Micro "Nucleo" L432KC](https://www.st.com/ja/evaluation-tools/nucleo-l432kc.html)
 * [ST Micro "Nucleo" L552ZE](https://www.st.com/en/evaluation-tools/nucleo-l552ze-q.html)
+* [ST Micro "Nucleo" WL55JC](https://www.st.com/en/evaluation-tools/nucleo-wl55jc.html)
 * [ST Micro STM32F103XX "Bluepill"](https://stm32-base.org/boards/STM32F103C8T6-Blue-Pill)
 * [ST Micro STM32F407 "Discovery"](https://www.st.com/en/evaluation-tools/stm32f4discovery.html)
 * [X9 Pro smartwatch](https://github.com/curtpw/nRF5x-device-reverse-engineering/tree/master/X9-nrf52832-activity-tracker/)
--- a/src/machine/board_lorae5.go
+++ b/src/machine/board_lorae5.go
@@ -9,8 +9,8 @@ import (
 )
 
 const (
-	LED1 = PB5
-	LED  = LED1 // Default LED
+	// We assume a LED is connected on PB5
+	LED = PB5 // Default LED
 )
 
 // SubGhz (SPI3)
@@ -46,6 +46,11 @@ var (
 		RxAltFuncSelector: AF7_USART1_2,
 	}
 	DefaultUART = UART0
+
+	// SPI
+	SPI3 = SPI{
+		Bus: stm32.SPI3,
+	}
 )
 
 func init() {
--- a/src/machine/machine_stm32wle5.go
+++ b/src/machine/machine_stm32wle5.go
@@ -7,6 +7,7 @@ package machine
 
 import (
 	"device/stm32"
+	"math/bits"
 	"runtime/interrupt"
 	"runtime/volatile"
 	"unsafe"
@@ -28,12 +29,15 @@ const (
 	AF15_EVENTOUT          = 15
 )
 
-func CPUFrequency() uint32 {
-	return 4e6
-}
+const (
+	SYSCLK        = 48e6
+	APB1_TIM_FREQ = SYSCLK
+	APB2_TIM_FREQ = SYSCLK
+)
 
-const APB1_TIM_FREQ = 4e6
-const APB2_TIM_FREQ = 4e6
+func CPUFrequency() uint32 {
+	return SYSCLK
+}
 
 const (
 	PA0  = portA + 0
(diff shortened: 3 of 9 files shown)


README.md
```diff
@@ -115,6 +115,7 @@ The following 72 microcontroller boards are currently supported:
 * [ST Micro "Nucleo" L031K6](https://www.st.com/ja/evaluation-tools/nucleo-l031k6.html)
 * [ST Micro "Nucleo" L432KC](https://www.st.com/ja/evaluation-tools/nucleo-l432kc.html)
 * [ST Micro "Nucleo" L552ZE](https://www.st.com/en/evaluation-tools/nucleo-l552ze-q.html)
+* [ST Micro "Nucleo" WL55JC](https://www.st.com/en/evaluation-tools/nucleo-wl55jc.html)
```
kenbell, line 118: Should increment the count of supported boards in the text above

src/runtime/runtime_stm32wle5.go
```diff
@@ -4,21 +4,125 @@
...
 	initTickTimer(&machine.TIM1)
+
+	// Radio init
+	subGhzInit()
+}
+
+func initCLK() {
+
+	// Set Power Voltage Regulator Range 2
+	//stm32.PWR.CR1.ReplaceBits(0b10, stm32.PWR_CR1_VOS_Msk, stm32.PWR_CR1_VOS_Pos)
```
kenbell: Should either un-comment, or completely remove

src/runtime/runtime_stm32wle5.go
```diff
@@ -4,21 +4,125 @@
...
+	for !stm32.RCC.CR.HasBits(stm32.RCC_CR_PLLRDY) {
+	}
+
+	// Enable PLL System Clock output.
+	stm32.RCC.PLLCFGR.SetBits(stm32.RCC_PLLCFGR_PLLREN)
+	for !stm32.RCC.CR.HasBits(stm32.RCC_CR_PLLRDY) {
+	}
+
+	// Set Flash Latency of 2 and wait until it's set properly
+	stm32.FLASH.ACR.ReplaceBits(0b010, 0b111, stm32.Flash_ACR_LATENCY_Pos)
```
kenbell: Can use symbolic constants for these values?

src/runtime/runtime_stm32wle5.go
```diff
@@ -4,21 +4,125 @@
...
+	for !stm32.RCC.CR.HasBits(stm32.RCC_CR_HSERDY) {
+	}
+
+	// Disable PLL
+	stm32.RCC.CR.ClearBits(stm32.RCC_CR_PLLON)
+	for stm32.RCC.CR.HasBits(stm32.RCC_CR_PLLRDY) {
+	}
+
+	// Configure PLL
+	stm32.RCC.PLLCFGR.Set(0x23020613)
```
kenbell: Can use symbolic constants here?  E.g. other STM32 worked on recently tend to have something like this:
```
	// Configure the PLL
	stm32.RCC.PLLCFGR.ReplaceBits(
		(1)| // 1 = RCC_PLLSOURCE_MSI
			(PLL_M-1)<<stm32.RCC_PLLCFGR_PLLM_Pos|
			(PLL_N<<stm32.RCC_PLLCFGR_PLLN_Pos)|
			(((PLL_Q>>1)-1)<<stm32.RCC_PLLCFGR_PLLQ_Pos)|
			(((PLL_R>>1)-1)<<stm32.RCC_PLLCFGR_PLLR_Pos)|
			(PLL_P<<stm32.RCC_PLLCFGR_PLLPDIV_Pos),
		stm32.RCC_PLLCFGR_PLLSRC_Msk|stm32.RCC_PLLCFGR_PLLM_Msk|
			stm32.RCC_PLLCFGR_PLLN_Msk|stm32.RCC_PLLCFGR_PLLP_Msk|
			stm32.RCC_PLLCFGR_PLLR_Msk|stm32.RCC_PLLCFGR_PLLPDIV_Msk,
		0)
```

src/runtime/runtime_stm32wle5.go
```diff
@@ -4,21 +4,125 @@
...
+	}
+
+	// Set clock source to PLL (0x3)
+	stm32.RCC.CFGR.ReplaceBits(0b11, 0b11, stm32.RCC_CFGR_SW_Pos)
+	for (stm32.RCC.CFGR.Get() & stm32.RCC_CFGR_SWS_Msk) != 0xc {
+	}
+
+}
+
+// SubGhzInit() enables SX126x radio module
```
kenbell: I don't think this should be in the runtime package - runtime is really about setting up clocks and time.

Should probably think about this as an optional peripheral that apps may (or may not) use - with it either:
1 - being in machine package
2 - being supported in the tinygo drivers repo

I'm not sure which is best - it's not a SPI/I2C peripheral - and might be a bit 'specialized' for the machine package?  Perhaps @aykevl  or @deadprogram  have some thoughts?

kenbell (commented):
Thanks for improving the support - suggest just a small amount of replacing magic values with symbolics (where possible) - and needs some thought about how (and when) to initialize the radio peripheral.

Also - looks like a change may be missing based on the compile errors?

src/runtime/runtime_stm32wle5.go
ofauchon: SX126x Radio (which is part of STM32WL) will have its own tinygo driver soon, so the code can be reused when using SX126x standalone module.
 
But this function is specific to STM32WL series (as SX126x radio unit and MCU are on the same die)
 
IMHO it's entirely part of the whole chipset initialisation process.

I'm not sure moving this function to tinygo-driver makes sense (I don't see other situations you might need this)

But having it in machine package instead, would not hurt. 

Other comments welcome .

deadprogram: `machine` package is a good place for this. Perhaps it should be exported so users who want to turn on this functionality can do so, but it is not always turned on by default?

deadprogram:
Once you have updated from any changes from feedback, @ofauchon can you please rebase against the latest `dev` branch? Hopefully that will get CI to build successfully.

deadprogram:
Something else must be wrong here @ofauchon see https://app.circleci.com/pipelines/github/tinygo-org/tinygo/6087/workflows/7a6938bf-d87d-4690-8782-209e919ae383/jobs/34339?invite=true#step-117-360

src/runtime/runtime_stm32wle5.go
```diff
@@ -4,21 +4,125 @@
...
+	initCLK()
 
 	// UART init
 	machine.Serial.Configure(machine.UARTConfig{})
 
 	// Timers init
 	initTickTimer(&machine.TIM1)
+
+	// Radio init
+	subGhzInit()
```
deadprogram: CI is now passing @ofauchon will you move `subGhzInit()` to `machine` package now, change to `SubGhzInit()`, and remove this call here so initialization is optional to the developer? :smile_cat:

src/runtime/runtime_stm32wle5.go
ofauchon: Removed

src/runtime/runtime_stm32wle5.go
ofauchon: All binaries replaced with symbolic constants.

src/runtime/runtime_stm32wle5.go
ofauchon: I could rewrite the PLL init with symbolic constants . Done

src/runtime/runtime_stm32wle5.go
ofauchon: SubGhzInit function moved in "src/machine/machine_stm32wle5.go" and exported so users can use it or not

ofauchon:
@deadprogram @aykevl @kenbell . 
I made all the suggested changes. 

The final PR Code  was tested on both LoraE5 and nucleo-wl55jc: 

- Simple timers tests OK , 
- UART OK, 
- Radio module SX126x simple tests (Continuous TX checked with a SDR)

Olivier

src/machine/board_lorae5.go
```diff
@@ -9,8 +9,8 @@ import (
 )
 
 const (
-	LED1 = PB5
-	LED  = LED1 // Default LED
+	// We assume a LED is connected on PB5
+	LED = PB5 // Default LED
```
aykevl, line 13: It looks like this change is unrelated to the PR. Was it intentional?

src/machine/machine_stm32wle5.go
```diff
@@ -221,6 +225,61 @@ func (p Pin) registerInterrupt() interrupt.Interrupt {
...
+}
+
+func (spi SPI) getBaudRate(config SPIConfig) uint32 {
+	var clock uint32
+	switch spi.Bus {
+	case stm32.SPI1:
+		clock = CPUFrequency()
+	case stm32.SPI2, stm32.SPI3:
+		clock = CPUFrequency()
+	}
```
aykevl, line 257: I don't understand this part. In all cases the clock equals the CPU frequency? Why not write `clock := CPUFrequency()` instead?

aykevl (commented):
Generally looks good to me.
I'm a little bit unsure about the `SubGhzInit` function. IIRC it's on the same chip, but connected through an (on-chip) SPI bus?
I'm almost inclined to say it should be implemented entirely in a driver package, not in the machine package. Not sure. On the other hand, it does need some special initialization so it's not just any other SPI peripheral.

src/machine/board_lorae5.go
ofauchon: This board don't actually have a LED. 
So I thought I would be a good idea to remove it. 

But, that breaks blinky1 , so I rolled-back the change.

src/machine/machine_stm32wle5.go
ofauchon: IIRC, this code was (quickly) borrowed from STM32F405

I guess the lines are the same because the both SPI1 and SPI2/SPI3 's bus use a 1:1 Prescaler. 
(in the current STM32WL setup) 

In the future, we may improve / tune the clocks, and the lines may turn into :

```
	switch spi.Bus {
	case stm32.SPI1:
		clock = CPUFrequency() /2 
	case stm32.SPI2, stm32.SPI3:
		clock = CPUFrequency() /4
	}
```
So we have two solutions: 

  - Add comments to explain the PRESCALERs are 1:1 in the current configuration
  - Review and complete the work on SYSCLK/BUS/SPI clocks , prescalers... but that will take time
 
As I plan to work on Power Saving for this chip in a near future, 
i'd prefer not spending too much time on the clocks right now, and juste add
some comment. 

Tell me what you feel the most confortable with . 

Olivier

deadprogram: Add comment about keeping separate to support future improvements, and IMO its all good.

ofauchon: I added the comment. 
Thank you all for the time spend on this PR !

ofauchon:
> Generally looks good to me. I'm a little bit unsure about the `SubGhzInit` function. IIRC it's on the same chip, but connected through an (on-chip) SPI bus? I'm almost inclined to say it should be implemented entirely in a driver package, not in the machine package. Not sure. On the other hand, it does need some special initialization so it's not just any other SPI peripheral.

Yes, SPI3/SubGhz Is an internal SPI bus dedicated to SX126X Radio Submodule communication.

I agree this SubGhzInit is not ideal it looks like an ugly quirk in machine code.

I'll remove this function for the moment, and try to implement this in the tinygo-driver.

src/machine/board_lorae5.go
kenbell: FWIW in the LoRa-e5 schematic, PB5 is connected to a RED LED through a jumper (see bottom right of the schematic): https://files.seeedstudio.com/products/113990934/LoRa-E5%20Dev%20Board%20v1.0.pdf

It certainly blinked the LED for me :)

ofauchon: image

I understand why i've never seen this led blinking and why I connected an external LED (hopefully on PB5)

kenbell: :-) the official dev board does look a little less artisanal.

deadprogram:
Thanks for responding to all the feedback @ofauchon 

Now merging.

PR #2312, opened 2021-11-24
stm32: pull-up on I2C lines

### LINKED ISSUE #2310: nucleo-l432kc i2c appears to be nonfunctional
I have been trying to use this nucleo board for I2c, to no avail.  

The nucleo-l432kc pins are wired as SCL PB6 (D5) and SDA PB7 (D4).

I have several devices on the i2c bus.

After repeatedly failing to be able to speak with the devices using machine.Tx(), I then loaded https://github.com/toyo/I2Cscan and tried scanning the bus for any devices.  None could be seen.

Without any rewiring, I wrote an Arduino sketch to do the same, and it worked perfectly.

I then removed the wires and hooked them instead to a feather-nrf52840, and used that target for tinygo.  It worked perfectly.

As such, my conclusion is that the nucleo-l432kc I2C implementation is not functioning properly.

Thank you for your consideration of the issue.

kenbell:
Fixes #2310

This changes the default for all STM32 (except stm32f103)

### DIFF
--- a/src/machine/machine_stm32_moder_gpio.go
+++ b/src/machine/machine_stm32_moder_gpio.go
@@ -1,3 +1,4 @@
+//go:build stm32 && !stm32f103
 // +build stm32,!stm32f103
 
 package machine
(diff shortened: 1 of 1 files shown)


deadprogram:
Thanks for the quick update @kenbell now merging.

PR #2316, opened 2021-11-25
os, syscall: implement ReadAt for unix, add stub for other OSs

dkegel-fastly:
Windows will take more work, so test is skipped there.

Lets the following pass on mac or linux:
```
git clone https://github.com/golang/image
cd image/tiff
tinygo test
```

### DIFF
--- a/src/os/filesystem.go
+++ b/src/os/filesystem.go
@@ -46,6 +46,9 @@ type FileHandle interface {
 	// Read reads up to len(b) bytes from the file.
 	Read(b []byte) (n int, err error)
 
+	// ReadAt reads up to len(b) bytes from the file starting at the given absolute offset
+	ReadAt(b []byte, offset int64) (n int, err error)
+
 	// Write writes up to len(b) bytes to the file.
 	Write(b []byte) (n int, err error)
 
--- a/src/os/os_test.go
+++ b/src/os/os_test.go
@@ -1,0 +1,48 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package os_test
+
+import (
+	"io"
+	. "os"
+	"runtime"
+	"testing"
+)
+
+// localTmp returns a local temporary directory not on NFS.
+func localTmp() string {
+	return TempDir()
+}
+
+func newFile(testName string, t *testing.T) (f *File) {
+	// TODO: use CreateTemp when it lands
+	f, err := OpenFile(TempDir()+"/_Go_"+testName, O_RDWR|O_CREATE, 0644)
+	if err != nil {
+		t.Fatalf("TempFile %s: %s", testName, err)
+	}
+	return
+}
+
+func TestReadAt(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Log("TODO: implement Pread for Windows")
+		return
+	}
+	f := newFile("TestReadAt", t)
+	defer Remove(f.Name())
+	defer f.Close()
+
+	const data = "hello, world\n"
+	io.WriteString(f, data)
+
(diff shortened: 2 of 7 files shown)


src/syscall/syscall_libc.go
```diff
@@ -140,6 +150,10 @@ func libc_getenv(name *byte) *byte
 //export read
 func libc_read(fd int32, buf *byte, count uint) int
 
+// ssize_t pread(int fd, void *buf, size_t count, off_t offset);
+//export pread
+func libc_pread(fd int32, buf *byte, count uint, offset int) int
```
aykevl:
```suggestion
// ssize_t pread(int fd, void *buf, size_t count, off_t offset);
//export pread
func libc_pread(fd int32, buf *byte, count uint, offset int64) int
```

This works because:

* MacOS is always 64-bit
* Linux uses musl, which always uses a 64-bit integer for `off_t` to avoid the need for a separate pread64 function.

And this should fix the TODO above.

src/os/os_test.go
```diff
@@ -0,0 +1,49 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package os_test
+
+import (
+	"io"
+	"os"
+	. "os"
```
aykevl: Here is the same issue as before. Do you want to import `"os"` and use `os.OpenFile`, or import `. "os"` and use `OpenFile` (no prefix)?

src/os/os_test.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		t.Fatalf("TempFile %s: %s", testName, err)
+	}
+	return
+}
+
+func TestReadAt(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Log("TODO: implement Pread for Windows")
+		return
```
aykevl, line 31: Hmm, this _should_ be `t.Skip` but it's not implemented so :shrug:

src/os/filesystem.go
```diff
@@ -46,6 +46,9 @@ type FileHandle interface {
 	// Read reads up to len(b) bytes from the file.
 	Read(b []byte) (n int, err error)
 
+	// Pread reads up to len(b) bytes from the file starting at the given absolute offset
+	Pread(b []byte, offset int64) (n int, err error)
```
aykevl: Why call this `Pread`? I think calling it `ReadAt` is nicer to match [`io.ReaderAt`](https://pkg.go.dev/io#ReaderAt) (in fact, you might want to embed `io.ReaderAt` here instead).

dkegel-fastly: I was just following upstream's os.File ReadAt's example, but I guess that's not a very good reason.  ReaderAt isn't available in go 1.15, so probably not time to embed that just yet.

src/os/os_test.go
dkegel-fastly: Again, I was just following upstream.  I'll fix it.

src/syscall/syscall_libc.go
dkegel-fastly: Done.

src/os/filesystem.go
dkegel-fastly: It has been ReadAtified.

src/os/os_test.go
dkegel-fastly: Done

deadprogram:
Even more conflicts here for your tender loving care @dkegel-fastly if you please.

dkegel-fastly:
That test failure looks spurious...

deadprogram:
Passing after a re-request.

dgryski (approved):
LGTM

deadprogram:

Issue #2317, opened 2021-11-25
stm32 i2c is not blocking


kenbell:
This may be related to #2310 (possibly similar root cause)

In at least some cases, i2c is dropping messages.  This sample code (with no i2c devices attached) results in lots of i2c messages that are dropped on nucleo-l432kc:
```
package main

import (
	"machine"
)

func main() {
	machine.I2C0.Configure(machine.I2CConfig{})
	w := []byte{0x42, 0x24}
	r := make([]byte, 2)

	for i := 0; i < 255; i++ {
		machine.I2C0.Tx(uint16(i), w, r)
	}
}
```
Logic Analyzer output:
![Screenshot from 2021-11-25 17-38-06](https://user-images.githubusercontent.com/509378/143483778-cb3e4260-30ed-4317-aabe-b2c7a63c8154.png)

PR #2320, opened 2021-11-25
stm32: fix timeout for i2c comms

kenbell:
ticks changed to 16ns causing timeouts to be very, very short.  This change updates timeouts for 16ns ticks.

This was causing a problem where the peripheral was busy, but instead of waiting for a reasonable delay to wait for peripheral to stop being busy, instead the code timed out.  This was particularly problematic for i2c scan code that expected a transmission error to indicate an address was unused, rather than the i2c peripheral in the MCU was busy.

This issue is symptomatic of the wider issue of the circular dependency...   runtime package uses machine package and machine package needs some notion of time to implement peripheral support.  We could really do with a 'clean' way for machine package to measure time.  For now, this is a quick fix to unbreak i2c.

### DIFF
--- a/src/machine/machine_stm32_i2c_revb.go
+++ b/src/machine/machine_stm32_i2c_revb.go
@@ -1,3 +1,4 @@
+//go:build stm32l5 || stm32f7 || stm32l4 || stm32l0
 // +build stm32l5 stm32f7 stm32l4 stm32l0
 
 package machine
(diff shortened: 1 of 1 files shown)


kenbell:
This is a fix for #2317 and a possible fix for #2310.  After this fix the logic analyzer shows this (as expected, each address is scanned in turn):
Screenshot from 2021-11-25 22-13-34

deadprogram:
Great work, thank you for getting to the bottom of this so quickly. Makes total sense, and agree some future cleanup will be in order.

Now merging. Thanks again @kenbell

PR #1977, opened 2021-06-30
main: improve help

sago35:
The following frequently used commands have been added to `tinygo help` .
If the English I used for help is not correct, I would like you to tell me how to write it.

* `tinygo targets`
* `tinygo info`
* `tinygo version`

Also, the help for `No command-line arguments supplied.` no longer displays an extra flag: character, as shown below.
In addition, I added gitSha1 to the version displayed by `usage()`.

before:
```
$ tinygo
TinyGo is a Go compiler for small places.
version: 0.19.0-dev
usage: C:\tinygo\tinygo\build\tinygo.exe command [-printir] [-target=<target>] -o <output> <input>

commands:
  build: compile packages and dependencies
  run:   compile and run immediately
  test:  test packages
  flash: compile and flash to the device
  gdb:   run/flash and immediately enter GDB
  env:   list environment variables used during build
  list:  run go list using the TinyGo root
  clean: empty cache directory (C:\Users\sago3\AppData\Local\tinygo)
  help:  print this help text

flags:

```

after:

```
$ tinygo
No command-line arguments supplied.
TinyGo is a Go compiler for small places.
version: 0.19.0-dev-3a1dee69
usage: C:\tinygo\tinygo\build\tinygo.exe command [-printir] [-target=<target>] -o <output> <input>

commands:
  build:   compile packages and dependencies
  run:     compile and run immediately
  test:    test packages
  flash:   compile and flash to the device
  gdb:     run/flash and immediately enter GDB
  env:     list environment variables used during build
  list:    run go list using the TinyGo root
  clean:   empty cache directory (C:\Users\sago3\AppData\Local\tinygo)
  targets: list targets
  info:    show info for specified target
  version: show version
  help:    print this help text

```

PR #2321, opened 2021-11-26
housekeeping: add missing tinygo commands to help

kenbell:
I noticed some of the commands are missing, like `tinygo targets`.  This adds them.  The order is somewhat arbitrary but i tried to keep them somewhat grouped with the existing documented commands.


dgryski (approved):
LGTM

deadprogram:
Did you happen to see https://github.com/tinygo-org/tinygo/pull/1977? It was another attempt to do the same thing as this PR but was never quite finished.

kenbell:
I didn't see that one.  I'll cancel this one - let's finish off #1977

PR #2322, opened 2021-11-26
os: TempDir(): obey TMPDIR on unix, TMP on win, etc

dkegel-fastly:
Uses upstream go code, lightly adjusted.

Adds smoke test.

### DIFF
--- a/src/os/file_anyos_test.go
+++ b/src/os/file_anyos_test.go
@@ -1,0 +1,25 @@
+package os_test
+
+import (
+	. "os"
+	"testing"
+)
+
+// TestTempDir assumes that the test environment has a filesystem with a working temp dir.
+func TestTempDir(t *testing.T) {
+	name := TempDir() + "/_os_test_TestTempDir"
+	Remove(name)
+	f, err := OpenFile(name, O_RDWR|O_CREATE, 0644)
+	if err != nil {
+		t.Errorf("OpenFile %s: %s", name, err)
+		return
+	}
+	err = f.Close()
+	if err != nil {
+		t.Errorf("Close %s: %s", name, err)
+	}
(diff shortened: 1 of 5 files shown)


niaow:
If this contains upstream Go code, appropriate license headers should be added.

dgryski (approved):
LGTM

deadprogram:
@dkegel-fastly this PR now needs to be rebased against `dev` to resolve merge conflicts, if you please.

src/os/file_anyos_test.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+// TestTempDir assumes that the test environment has a filesystem with a working temp dir.
+func TestTempDir(t *testing.T) {
+	name := TempDir() + "/_os_test_TestTempDir"
+	Remove(name)
+	f, err := OpenFile(name, O_RDWR|O_CREATE, 0644)
+	if err != nil {
+		t.Errorf("OpenFile %s: %s", name, err)
+		return
```
aykevl, line 15: Suggestion: I believe this can be replaced with `t.Fatal`.

aykevl (approved):
Looks reasonable to me!

src/os/file_anyos_test.go
dkegel-fastly: It would still need the return at the moment, I went for lower cognitive dissonance :-)

deadprogram:
Thanks everyone! Now merging.

src/os/file_anyos_test.go
aykevl: Fair enough. Indeed, `t.Fatal` isn't obviously returning (even though the effect is similar).

PR #2324, opened 2021-11-27
os.Remove: avoid double-wrapping err; fixes TODO in test

dkegel-fastly:
Removing a nonexistent directory should generate true for IsNotExist(err).

It didn't because fs.Remove had already wrapped the underlying error, and double-wrapping confuses things.

This is like part two of https://github.com/tinygo-org/tinygo/pull/2265, which introduced the problem.

### DIFF
--- a/src/os/file.go
+++ b/src/os/file.go
@@ -47,7 +47,7 @@ func Remove(path string) error {
 	}
 	err := fs.Remove(suffix)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


dgryski (approved):
LGTM

deadprogram:
Thanks @dkegel-fastly for the fix and @dgryski for reviewing. Now merging.

Issue #2166, opened 2021-10-10
register.ReplaceBits not compatible with SVD generated mask


zdima:
The SVD define all bit mask (*Msk) as a physical location of the bits, i.e. 0x00300000.
The mask argument of a `register.ReplaceBits` is not really a bit mask, but the size of a field in bits.
To get correct result, one should add extra shift when calling it and use position value 0:

https://github.com/tinygo-org/tinygo/blob/64d048c47c95bc0a70b4cff109afc1f9039705a4/src/machine/machine_stm32f103.go#L699

The usage on `register.ReplaceBits` is mixed with assumptions that mask argument is a positioned bit mask or it is a size of a field mask.

This may produce incorrect result when position is anything but 0, and lead to many hours of debugging "why it is not working?!".
https://github.com/tinygo-org/tinygo/blob/157382600535c95bdbf5d6e8c4b9e5ca71749018/src/machine/machine_k210.go#L540
And when correction is made, it introduce unnecessary shift calls.
```
kendryte.SYSCTL.CLK_TH5.ReplaceBits((0x03<<kendryte.SYSCTL_CLK_TH5_I2C2_CLK_Pos),
     kendryte.SYSCTL_CLK_TH5_I2C2_CLK_Msk, 0)
```

The common sense and instinct call to use predefined (generated) bit mask as is:
```
kendryte.SYSCTL.CLK_TH5.ReplaceBits(0x03, kendryte.SYSCTL_CLK_TH5_I2C2_CLK_Msk,
     kendryte.SYSCTL_CLK_TH5_I2C2_CLK_Pos)
```

The `ReplaceBits` need a fix to use a real bit mask and code that uses the mask as 'size of a field' fixed.
```
StoreUint#(&r.Reg, LoadUint#(&r.Reg)&^mask|value<<pos)
```

zdima:
The existing ReplaceBits call changes will cause regression in local repo and outside projects.
The set of setter and getters provided as part of the SVD with PR #2325.

PR #2325, opened 2021-11-27
Provide Set/Get for each register field described in SVD files (2).

zdima:
The new generated code structure is 100% compatible with existing code and almost identical with exception of clustered registers.
The new code define named structure for clustered registers instead of anonymous to allow preserve direct access and also provide Set/Get for bit fields of those clustered registers.
The Set/Get for the rest of the registers are added directly to existing top level struct, which allow persevere structure and add functionality.
**The size of the binaries produced by smoketest is ~98% the same**. Few binary had increase of 20 bytes. Very few with binary size >40KB increased by 100 bytes.
@aykevl
During testing found some SVD files with incorrect code, which produces wrong structures by existing SVD-GEN (and new code):
peripheral:FMC cluster:FMC_SDBANK size: 4 struct size: 12 in:
> lib/stm32-svd/svd/stm32h743.svd 
> lib/stm32-svd/svd/stm32h743v.svd
> lib/stm32-svd/svd/stm32h747cm4.svd
> lib/stm32-svd/svd/stm32h747cm7.svd
> lib/stm32-svd/svd/stm32h753.svd
> lib/stm32-svd/svd/stm32h753v.svd
> lib/stm32-svd/svd/stm32h7b3.svd 

peripheral: PORT cluster: PORT_GROUP size: 112 struct size: 116 in:
> lib/cmsis-svd/data/Atmel/ATSAML11D14A.svd
> lib/cmsis-svd/data/Atmel/ATSAML11D15A.svd
> lib/cmsis-svd/data/Atmel/ATSAML11D16A.svd
> lib/cmsis-svd/data/Atmel/ATSAML11E14A.svd
> lib/cmsis-svd/data/Atmel/ATSAML11E15A.svd
> lib/cmsis-svd/data/Atmel/ATSAML11E16A.svd

many errors in all nrf* except nrf51.svd

### DIFF
--- a/tools/gen-device-svd/gen-device-svd.go
+++ b/tools/gen-device-svd/gen-device-svd.go
@@ -128,19 +128,28 @@ type Peripheral struct {
 // A PeripheralField is a single field in a peripheral type. It may be a full
 // peripheral or a cluster within a peripheral.
 type PeripheralField struct {
-	Name        string
-	Address     uint64
-	Description string
-	Registers   []*PeripheralField // contains fields if this is a cluster
-	Array       int
-	ElementSize int
-	Bitfields   []Bitfield
+	Name         string
+	Address      uint64
+	Description  string
+	Registers    []*PeripheralField // contains fields if this is a cluster
+	Array        int
+	ElementSize  int
+	Constants    []Constant
+	ShortName    string     // name stripped of "spaced array" suffix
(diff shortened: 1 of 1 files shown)


zdima:
related to #2166 and fixes by providing set/get methods for individual fields.

aykevl (approved):
Thank you! Looks good to me and I couldn't find a difference in the binaries with or without this patch.

PR #2328, opened 2021-11-28
Implement os.Stat

dkegel-fastly:
Mostly just copied over from go's standard library and lightly massaged to fit tinygo.

### DIFF
--- a/src/os/export_windows_test.go
+++ b/src/os/export_windows_test.go
@@ -1,0 +1,12 @@
+// Copyright 2016 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package os
+
+// Export for testing.
+
+var (
+	FixLongPath     = fixLongPath
+	CanUseLongPaths = canUseLongPaths
+)
--- a/src/os/file.go
+++ b/src/os/file.go
@@ -171,11 +171,6 @@ func (f *File) Stat() (FileInfo, error) {
 	return nil, &PathError{"stat", f.name, ErrNotImplemented}
 }
 
-// Sync is a stub, not yet implemented
-func (f *File) Sync() error {
-	return ErrNotImplemented
-}
-
 func (f *File) SyscallConn() (syscall.RawConn, error) {
 	return nil, ErrNotImplemented
 }
@@ -188,16 +183,6 @@ func (f *File) Fd() uintptr {
 // Truncate is a stub, not yet implemented
 func (f *File) Truncate(size int64) error {
 	return &PathError{"truncate", f.name, ErrNotImplemented}
-}
-
-const (
-	PathSeparator     = '/' // OS-specific path separator
-	PathListSeparator = ':' // OS-specific path list separator
-)
-
-// IsPathSeparator reports whether c is a directory separator character.
-func IsPathSeparator(c uint8) bool {
-	return PathSeparator == c
 }
 
 // PathError records an error and the operation and file path that caused it.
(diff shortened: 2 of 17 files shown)


src/syscall/syscall_libc.go
```diff
@@ -193,6 +274,17 @@ func libc_mkdir(pathname *byte, mode uint32) int32
 //export rmdir
 func libc_rmdir(pathname *byte) int32
 
+// The odd $INODE64 suffix is an Apple compatibility feature, see https://assert.cc/posts/darwin_use_64_bit_inode_vs_ctypes/
+// Without it, you get the old, smaller struct stat from mac os 10.2 or so.
```
aykevl: This is specific to macOS so this needs to be in syscall_libc_darwin.go.

src/syscall/syscall_libc.go
```diff
@@ -68,6 +69,86 @@ func Rmdir(path string) (err error) {
 	return
 }
 
+type Timespec struct {
+	Sec  int64
+	Nsec int64
+}
+
+// Go chose Linux's field names for Stat_t, see https://github.com/golang/go/issues/31735
+type Stat_t struct {
```
aykevl: Same here. This should probably live in syscall_libc_darwin.go.

src/syscall/syscall_libc.go
```diff
@@ -68,6 +69,86 @@ func Rmdir(path string) (err error) {
...
+	Blocks    int64
+	Blksize   int32
+	Flags     uint32
+	Gen       uint32
+	Lspare    int32
+	Qspare    [2]int64
+}
+
+// Source: https://github.com/apple/darwin-xnu/blob/main/bsd/sys/_types/_s_ifmt.h
+// Consistent with values we use from lib/wasi-libc/sysroot/include/sys/stat.h
```
aykevl: Even with that, I'd prefer if this lived in syscall_libc_darwin.go in the (unlikely) case that a different OS uses different constants. I would prefer a little copying over deduplication in this case, unless these constants are some sort of standard (which they don't appear to be).

(Also, sidenote: this syscall package is not used on Linux. We use the upstream Go syscall package on Linux).

aykevl (commented):
Haven't looked through all the code yet.

src/syscall/syscall_libc.go
dkegel-fastly: Thanks.  I had it in syscall_libc_darwin originally, but moved it to get nintendo switch to pass, which was probably dumb :-)

src/syscall/syscall_libc.go
dkegel-fastly: Sorted.

src/syscall/syscall_libc.go
dkegel-fastly: Sorted

src/syscall/syscall_libc.go
dkegel-fastly: Sorted.

dkegel-fastly:
Tested on both mac and wasi using the program
```
package main

import (
	"fmt"
	"os"
)

func main() {
	finfo, err := os.Stat("foo.dat")
	if err != nil {
		fmt.Printf("Error %s\n", err.Error())
		return
	}
        fmt.Printf("Name %v\n", finfo.Name())
        fmt.Printf("Size %v\n", finfo.Size())
        fmt.Printf("Mode %v\n", finfo.Mode())
        fmt.Printf("ModTime %v\n", finfo.ModTime())
        fmt.Printf("IsDir %v\n", finfo.IsDir())
}
```
with the commands
```
echo "hello" > foo.dat
go run stat.go
tinygo run stat.go
tinygo build -target=wasi -scheduler=coroutines -gc=leaking -wasm-abi=generic -o stat.wasm stat.go && wasmtime --dir=. stat.wasm
```
and all three give the same answers (except for mode on wasi, which I think is a known issue).

dgryski (approved):
Given that the majority of this is taken from upstream Go, and the few libc bits needed for tinygo integration works, I'd say LGTM.

dgryski:
Do any more stdlib packages pass with this change?

dkegel-fastly:
I don't expect so, but path/filepath only needs MkdirAll, Chdir and Chmod, and I have MkdirAll ready to submit as soon as this pull request is merged.  Maybe I can set my sights on path/filepath for next weekend.

deadprogram:
Thanks everyone! Now merging.

PR #2330, opened 2021-11-29
runtime: use LLVM intrinsic to read the stack pointer

aykevl:
This results in slightly smaller code and simplifies reading the stack pointer on AVR.


deadprogram:
What an excellent idea! Thanks @aykevl now merging.

Issue #2326, opened 2021-11-28
Embrace run-time determination of target characteristics in a standard way


rayozzie:
We - Blues Wireless - make a cellular peripheral that connects with host MCU's ranging from 8-bit MCUs and the 16-bit MSP430, all the way up to Raspberry Pi's.  Our developer audience is split roughly in half between RPi-class and microcontroller-class, and the microcontrollers are hugely fragmented.

Developer experience means a lot to us:  dev.blues.io

As such, we have implemented the moral equivalent of a "user agent" within the libraries that the developer uses to talk with our Notecard, and we use that info to try to make our developer tools and Examples better based on trends of which Microcontrollers are used more than others.

In full golang, we do this - courtesy of the great shirou/gopsutil package:
https://github.com/blues/note-go/blob/main/notecard/ua.go

In tinygo, I was limited to doing this:
https://github.com/blues/note-tinygo/blob/master/ua.go

I have several comments.

1) I was pretty surprised that runtime.Version() isn't implemented.  You might consider implementing this.

2) I would like to propose a very simple feature that will take about 30m of editing, wherein a standard feature is implemented across all target .go's such as this:  https://github.com/tinygo-org/tinygo/blob/release/src/machine/board_arduino_nano33.go
In these GO files, you would implement these two strings - both required in all machine implementations:

// Target identifiers
const (
	target_DISPLAY = "Arduino NANO 33 IoT"
	target_ID = "arduino-nano33"
) 

These would ultimately prove to be very helpful not just for analytics, but also in case the developer needs to do conditional code based on specific oddball characteristics of a specific MCU.

3) It would be great if there were a standard set of methods implemented by all platforms, analogous to the gopsutil functions.  I say this because as more and more people do implement reusable library functions, they will want to likely do conditional code based upon machine capabilities.  You might even consider having standard "UART interface enumerators" and "I2C interface enumerators", and so on.
It would be truly great if the golang dev (and thus ua.go) could get "target name" in a standard way across targets, by defining standard const's for things like "nucleo-l432kc" and "STMicroelectronics Nucleo L432KC", because I think people might want to do run-time checks as well as analytics.

kenbell:
TinyGo supports build-time strings the same syntax as regular Go: https://github.com/tinygo-org/tinygo/blob/release/main.go#L1094

Could perhaps extend this to inject the 'target name' into a known string variable during build?

aykevl:
Thank you for your interest! These are all good ideas.



[...]

PR #2332, opened 2021-11-29
machine: add Device constant

aykevl:
This field contains the microcontroller name that we're compiling for, or "generic" if we're not running on a microcontroller.

See: https://github.com/tinygo-org/tinygo/issues/2326


deadprogram:
Very cool, thanks @aykevl

PR #2336, opened 2021-11-30
Makefile: add more stdlib packages for testing

dgryski:


dgryski:
Not quite yet.  `compress/lzw` still needs too much memory, and `debug/plan9obj` needs ReadAt which is still stubbed on Windows.

PR #2337, opened 2021-12-01
docker: update Dockerfile for LLVM 12 changes

deadprogram:
This PR updates the dev Dockerfile for LLVM 12 changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,22 @@
-# TinyGo base stage installs the most recent Go 1.17.x, LLVM 11 and the TinyGo compiler itself.
-FROM golang:1.17 AS tinygo-base
+# tinygo-llvm stage obtains the llvm source for TinyGo
+FROM golang:1.17 AS tinygo-llvm
 
-RUN wget -O- https://apt.llvm.org/llvm-snapshot.gpg.key| apt-key add - && \
-    echo "deb http://apt.llvm.org/bullseye/ llvm-toolchain-bullseye-11 main" >> /etc/apt/sources.list && \
-    apt-get update && \
-    apt-get install -y llvm-11-dev libclang-11-dev lld-11 git
+RUN apt-get update && \
+    apt-get install -y apt-utils make cmake clang-11 binutils-avr gcc-avr avr-libc ninja-build
+
+COPY ./Makefile /tinygo/Makefile
+
+RUN cd /tinygo/ && \
+    make llvm-source
+
+# tinygo-llvm-build stage build the custom llvm with xtensa support
+FROM tinygo-llvm AS tinygo-llvm-build
+
+RUN cd /tinygo/ && \
+    make llvm-build
+
+# tinygo-compiler stage builds the compiler itself
+FROM tinygo-llvm-build AS tinygo-compiler
 
 COPY . /tinygo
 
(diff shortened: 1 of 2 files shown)


deadprogram:
Note that the `dev` docker build is now based on the Xtensa fork of LLVM, so it can be used for ESP32 boards.

Dockerfile
```diff
@@ -1,80 +1,41 @@
...
-COPY --from=tinygo-base /tinygo/Makefile /tinygo/
-COPY --from=tinygo-base /tinygo/tools /tinygo/tools
-COPY --from=tinygo-base /tinygo/lib /tinygo/lib
+COPY --from=tinygo-compiler /tinygo/build/tinygo /go/bin/tinygo
 
 RUN cd /tinygo/ && \
-    apt-get update && \
-    apt-get install -y apt-utils make clang-11 binutils-avr gcc-avr avr-libc && \
-    make gen-device
+    make wasi-libc binaryen gen-device
```
sago35: It may be better to use something like "make gen-device -j".
The only difference is the speed.

sago35 (commented):
I think it's good.
It may be better to add -j.

Dockerfile
deadprogram: Is `-j` actually used by `gen-device` task? I did not see that flag has any effect on that, and `wasi-libc` already has a `-j4` in the makefile itself. Or am I missing something here?

aykevl:
Why do we build LLVM manually? Why not use the LLVM build from apt.llvm.org?
You can select LLVM 12 by adding `-tags=llvm12` to the build command. Or we should change the default from LLVM 11 to LLVM 12 in go-llvm.

Dockerfile
aykevl: Yes, `gen-device` is sped up with the `-j` flag. It runs `gen-device-avr`, `gen-device-sam`, etc in parallel.

Dockerfile
```diff
@@ -1,80 +1,41 @@
...
 
-# tinygo-all stage installs the needed dependencies to compile TinyGo programs for all platforms.
-FROM tinygo-wasm AS tinygo-all
+# tinygo-tools stage installs the needed dependencies to compile TinyGo programs for all platforms.
+FROM tinygo-compiler AS tinygo-tools
 
-COPY --from=tinygo-base /tinygo/Makefile /tinygo/
-COPY --from=tinygo-base /tinygo/tools /tinygo/tools
-COPY --from=tinygo-base /tinygo/lib /tinygo/lib
+COPY --from=tinygo-compiler /tinygo/build/tinygo /go/bin/tinygo
```
thaJeztah: This looks redundant, because this stage already uses `FROM tinygo-compiler`

Just glancing over (sorry for the drive-by comment), but it looks like with this changes, the Dockerfile effectively become a single-stage build (with some intermediate "breakpoints"); was that the intent?

```dockerfile
FROM golang:1.17 AS tinygo-llvm
...
FROM tinygo-llvm AS tinygo-llvm-build
..
FROM tinygo-llvm-build AS tinygo-compiler
...
FROM tinygo-compiler AS tinygo-tools
```

From the above; `tinygo-tools` "extends" `tinygo-compiler`, "extends" `tinygo-llvm-build`, "extends" `tinygo-llvm`, "extends" `tinygo-llvm`, which "extends" `golang:1.17`

(admitted, before this, it looks like "most" things were already copied (`COPY --from=tinygo-base` etc))

This means that the final image will contain all sources that were copied in (and intermediate layers, some of which look to be deleting files)

deadprogram: > This means that the final image will contain all sources that were copied in (and intermediate layers, some of which look to be deleting files)

Yes, the idea with the `dev` build was always that you could use it to add features to TinyGo and rebuild from within the image itself.

deadprogram:
> Why do we build LLVM manually? Why not use the LLVM build from apt.llvm.org?

I thought we needed the manual build in order to pickup the Xtensa bits. I want to be able to use the `dev` build on ESP32 targets.

Dockerfile
deadprogram: OK, I can update that.

deadprogram: Updated.

Dockerfile
deadprogram: Removed the redundant copy, thanks for the feedback @thaJeztah

deadprogram:
@aykevl any comment on my response to your comment? :ping_pong:

aykevl:
Well, okay.
But why aren't we just pulling the build artifact from GHA and use that in the image? That would be a lot simpler.

(Also: it might be a good idea to make a custom GitHub Action for pulling the GHA build artifact to use in tests - similar to how there is `actions/setup-go@v2` we could have `tinygo-org/setup-tinygo@v1`).

deadprogram:
> But why aren't we just pulling the build artifact from GHA and use that in the image? That would be a lot simpler.

Because this image is intended to also be able to be used for local development on TinyGo itself. Although it might need a few tweaks now to be used in that fashion.

> Also: it might be a good idea to make a custom GitHub Action for pulling the GHA build artifact to use in tests

That is for sure an excellent idea, we should work on that.

deadprogram:
Unless there are other changes requested, I think this PR should be ready to merge now.

aykevl:
I'm just thinking it's rather wasteful to build LLVM for every merge into the dev branch. However:

> I thought we needed the manual build in order to pickup the Xtensa bits. I want to be able to use the `dev` build on ESP32 targets.

Yeah, for that you need a custom build.

deadprogram:
I am pretty sure that the multi-stage build will cache the parts that do not change, such as the LLVM build steps. They certainly do when I use this image locally.

deadprogram:
Working on getting some caching working here.

deadprogram:
I have added caching that now correctly handles the llvm-source build. This PR is finally ready!

deadprogram:
Now merging, thanks @thaJeztah @aykevl @sago35 for the feedback and reviews.

PR #2340, opened 2021-12-01
src/testing: stub b.RunParallel() and PB.Next()

dgryski:
This allows `archive/zip` and a few other packages tests to build, but sadly not pass tests.


deadprogram:
Thanks @dgryski now merging.

PR #2342, opened 2021-12-01
src/os: fix build tags for non-windows PathSeparator

dgryski:


deadprogram:
Thanks @dgryski merging.

PR #2343, opened 2021-12-02
fix: parseint float ops

fdymylja:
adds a merge instruction in tinygo which makes the error comparison through a string comparison rather than interface comparison


fdymylja:
ops. wrong repo!

Issue #2297, opened 2021-11-20
STM32 ADC implementation needed


deadprogram:
Has anyone started on any implementation for ADC on any STM32 platforms? @kenbell @ofauchon or anyone else who is into STM32?

PR #2345, opened 2021-12-04
machine/stm32f103, stm32f4: Add ADC support

deadprogram:
This is WIP on adding support for ADC on the STM32F4 family of processors.


deadprogram:
Intended to help address #2297

deadprogram:
Adding STM32F103 aka "Bluepill" to this PR as well.

deadprogram:
Tested on my Adafruit Feather STM32F405 board and worked as expected.

kenbell (approved):
LGTM

I'm a complete novice on ADC input - but if I understand correctly, this code uses a single ADC in scan mode to scan all analog pins configured for analog input.

It seems like the only true commonality between f4 and f103 is possibly the 'get' function.  I don't think that's worth unifying though.  It looks like it should be able to re-use the f4 implementation for other classes of STM32 in future.

deadprogram:
Thanks for reviewing @kenbell now merging.

PR #2346, opened 2021-12-04
machine: Interfaces for compile-time checking SPI/UART/I2C. Resolves #2334

### LINKED ISSUE #2334: Proposal: add unexported interfaces for compile-time interface implementation check
I don't mean to shoot down https://github.com/kenbell/tinygo-target-validator, but it would be relatively simple to check for interface implementation on various APIs. It consists of adding some `machine` package wide interfaces and asserting interface fulfillment as so:

`machine/interfaces_uart.go`
```go
//go:build atsamd21 || rp2040

// This can be wrapped in an init() function to prevent uarter from appearing as a package wide type while developing.

type uarter interface {
       	io.Reader
	io.Writer

	Buffered() int
}

// If this causes error then UART0 does not implement the expected UART interface. Please check
// method implementation or remove your target from build tags.
var _ uarter = UART0
```

It would be relatively simple to implement and maintain. I can help out adding them.

soypat:
Resolves #2334.

Added 3 files to consolidate SPI/UART/I2C interfaces to standarize way of programming tinygo targets. As mentioned in #2334, this adds no size augmentation for `microbit` target. Other targets were not tested.


deadprogram:
@soypat I think we have already been doing some of this work elsewhere. Should this PR be closed now?

soypat:
Hey @deadprogram! Where are we doing that work?

deadprogram:
For example https://github.com/tinygo-org/tinygo/blob/release/src/machine/i2c.go#L9-L18

soypat:
You are right Ron! Looks like good work was done in the meantime. Closing!

PR #2347, opened 2021-12-04
os: implement and smoketest os.Chdir and os.Chmod

dkegel-fastly:
And add the same build tags on file_anyos_test.go as on file_anyos.go, else it fails on some bare metalish targets in "make smoketest".

With this, "tinygo test io/ioutil" compiles, but does not yet pass.


dgryski (approved):
LGTM.  Nice!

deadprogram:
Thank you @dkegel-fastly now merging.

Issue #2348, opened 2021-12-05
builder: library headers rename can fail when run concurrently


niaow:
We do not currently have any form of inter-process cache locking. As a result, the directory rename when building picolibc can (rarely) fail when run concurrently. Example:
```
error: rename /home/niaow/.cache/tinygo/picolibc-thumbv6m-unknown-unknown-eabi-cortex-m0plus/include.tmp2615988351 /home/niaow/.cache/tinygo/picolibc-thumbv6m-unknown-unknown-eabi-cortex-m0plus/include: file exists
````

sago35:
At least as of #1355 , there were cases of crashes during build.
The reason for this is cache inconsistency between multiple tinygo processes.

https://github.com/tinygo-org/tinygo/pull/1355#issuecomment-826015110

aykevl:
I _think_ it would be safe to use the existing file if `os.Rename` doesn't work.

Slightly better would be to use a lock, and check again whether the target has already been built by another process. This avoids rebuilding libraries when there is a cached version already (and thus wasting CPU time).

niaow:
I split the locking piece into #2392. We still need both, but the locking piece is more complex as it requires restructuring a few parts of the build to avoid deadlocks.

PR #1349, opened 2020-09-04
Add -j option to smoketest

sago35:
#1330 

I created a PR to make it easier to check the CI results.
There is no plan to merge it for now.

PR #2349, opened 2021-12-05
Makefile: run smoke tests concurrently

niaow:
This change breaks each smoke test into a separate build rule.
The outputs are saved, and then all printed together at the end.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,7 +92,7 @@ commands:
           paths:
             - lib/wasi-libc/sysroot
       - run: make gen-device -j4
-      - run: make smoketest XTENSA=0
+      - run: make smoketest -j2 XTENSA=0
       - save_cache:
           key: go-cache-v2-{{ checksum "go.mod" }}-{{ .Environment.CIRCLE_BUILD_NUM }}
           paths:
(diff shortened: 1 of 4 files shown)


niaow:
Before:
```
real    2m19.197s
user    11m51.756s
sys     4m32.056s
```

After:
```
real    0m34.854s
user    21m25.876s
sys     9m13.691s
```

I am pretty sure that the increase in user time on my machine is a result of multiple builds of the same thing running at the same time, which is a problem which still needs to be solved.

niaow:
This is also blocked by #2348

niaow:
I also seem to have forgotten yet again that `build` does not exist consistently in CI.

sago35:
This PR is similar to the following PR.

#1355
#1349

niaow:
Huh. I guess I missed that.

niaow:
Hmm. I might try to rework the cache.

aykevl:
Hmm. I'm not entirely convinced to be honest. The speed improvements are certainly tempting, but it would be unfortunate to lose the ability to specify the build commands as a list of commands. I really like how these commands are very easy to read and modify at the moment.

Perhaps there is a way to get the best of both worlds?

PR #2350, opened 2021-12-05
os: implement and smoketest os.Setenv, os.Unsetenv, and os.Clearenv

dkegel-fastly:
Makes joho/dotenv a bit happier...

### DIFF
--- a/src/os/env_test.go
+++ b/src/os/env_test.go
@@ -46,6 +46,34 @@ func TestUnsetenv(t *testing.T) {
 	}
 }
 
+func TestClearenv(t *testing.T) {
+	const testKey = "GO_TEST_CLEARENV"
+	const testValue = "1"
+
+	// reset env
+	defer func(origEnv []string) {
+		for _, pair := range origEnv {
+			// Environment variables on Windows can begin with =
+			// https://blogs.msdn.com/b/oldnewthing/archive/2010/05/06/10008132.aspx
+			i := strings.Index(pair[1:], "=") + 1
+			if err := Setenv(pair[:i], pair[i+1:]); err != nil {
+				t.Errorf("Setenv(%q, %q) failed during reset: %v", pair[:i], pair[i+1:], err)
+			}
+		}
+	}(Environ())
+
+	if err := Setenv(testKey, testValue); err != nil {
+		t.Fatalf("Setenv(%q, %q) failed: %v", testKey, testValue, err)
+	}
+	if _, ok := LookupEnv(testKey); !ok {
+		t.Errorf("Setenv(%q, %q) didn't set $%s", testKey, testValue, testKey)
+	}
+	Clearenv()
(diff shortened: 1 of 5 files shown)


dankegel:
Ha, maybe I needed to build tinygo on linux with CGO_ENABLED=1  to see this problem.

dgryski:
Still gota CI failure:
```
tinygo:ld.lld: error: undefined symbol: syscall.setenv_c
>>> referenced by env_unix.go:128 (/opt/hostedtoolcache/go/1.17.3/x64/src/syscall/env_unix.go:128)
>>>               /tmp/tinygo1840787456/main.o:(os_test.TestLookupEnv)
failed to run tool: ld.lld
error: failed to link /tmp/tinygo1840787456/main: exit status 1
```

dkegel-fastly:
Indeed.  I do not know how to wire up cgo inside runtime for linux.  
cf. discussion at https://gophers.slack.com/archives/CDJD3SUP6/p1638750577308900

dkegel-fastly:
Punted on Linux for the moment, other platforms should still be ok.

aykevl:
> Punted on Linux for the moment, other platforms should still be ok.

You can put this in the runtime package:

```go
//go:linkname syscall_setenv_c syscall.setenv_c
func syscall_setenv_c(k string, v string) {
    // call C setenv like you do on MacOS for example
}
```

Or is there a reason that doesn't work?

dkegel-fastly:
Thanks for the tip.  I had no idea how to do that...

deadprogram:
Is this PR ready for merge now? @dkegel-fastly @dgryski @aykevl

dkegel-fastly:
I think so.  I've asked @dgryski for review...

src/os/env_test.go
```diff
@@ -0,0 +1,92 @@
...
+	value, ok := LookupEnv(smallpox) // Should not exist.
+	if ok || value != "" {
+		t.Fatalf("%s=%q", smallpox, value)
+	}
+	defer Unsetenv(smallpox)
+	err := Setenv(smallpox, "virus")
+	if err != nil {
+		t.Fatalf("failed to release smallpox virus")
+	}
+	_, ok = LookupEnv(smallpox)
```
dgryski, line 88: Do you want to check the returned key here?

dgryski (changes requested):
Maybe fix up the test, then LGTM.

src/os/env_test.go
dkegel-fastly: The test is the upstream test verbatim.

IMHO the upstream test is good enough, even without checking the other return value of LookupEnv... and I'd like to avoid diverging.

dgryski: SGTM

dgryski (approved):
LGTM

deadprogram:

PR #2352, opened 2021-12-06
Also search for wasm-opt under libexec/

fgsch:
With this change, we could drop the dependency on binaryen in the macos package (homebrew) and copy the binary as part of the release process to libexec/.

In the future, we could extend this change to all the platforms and let tinygo pick it up from there, but we can burn that bridge when we get there.

### DIFF
--- a/goenv/goenv.go
+++ b/goenv/goenv.go
@@ -113,6 +113,7 @@ func findWasmOpt() string {
 	searchPaths := []string{
 		tinygoroot + "/bin/wasm-opt",
 		tinygoroot + "/build/wasm-opt",
+		tinygoroot + "/libexec/wasm-opt", // for Homebrew
 	}
 
 	var paths []string


fgsch:
It seems that most (all?) of the errors are due to timeouts.

deadprogram:
Not all https://app.circleci.com/pipelines/github/tinygo-org/tinygo/6179/workflows/10c9f4c8-ccf4-46f2-9c5f-27303471d623/jobs/34623?invite=true#step-114-5

fgsch:
@deadprogram that doesn't look related to this PR though

fgsch:
@deadprogram can you re-run this please?

fgsch:
force pushed it after updating the date and the errors are gone 🤷

aykevl:
I don't think I fully understand why this is necessary but the change itself looks reasonable.
Can you maybe add a short comment why it looks in libexec? Just something like ` // for Homebrew` might already help the reader.

fgsch:
This is needed because we cannot add the wasm-opt binary to bin in Homebrew without creating a symlink to /usr/local/bin, which might clash with an existing or future binaryen installation.

I've added the comment though ideally we would use it it for all the supported platforms. 
But we can burn that bridge when we get there.

deadprogram:
Based on what @niaow said about Windows, since the homebrew version is new enough, we should probably not build binaryen for macOS, Which make this PR no longer needed, nor https://github.com/tinygo-org/homebrew-tools/pull/18

We probably also want to modify the CI build to use the homebrew version.

deadprogram:
Now that https://github.com/tinygo-org/tinygo/pull/2527 has merged, this PR can be closed. Thanks @fgsch

Issue #2339, opened 2021-12-01
issue with maps indexed by float64? golang/geo/s2 TestCellFaces fails


dgryski:
```
--- FAIL: TestCellFaces
    edge (0.000000000000000000000000, 0.000000000000000000000000, 0.000000000000000000000000) counts wrong, got 1, want 2
    edge (0.000000000000000000000000, 0.000000000000000000000000, 0.000000000000000000000000) counts wrong, got 1, want 2
    edge (0.000000000000000000000000, 0.000000000000000000000000, 0.000000000000000000000000) counts wrong, got 1, want 2
    edge (0.000000000000000000000000, 0.000000000000000000000000, 0.000000000000000000000000) counts wrong, got 1, want 2
```

These are entries from a map indexed by a struct.   Here is another program that demonstrates the map-of-structs problem, although I'm not 100% certain it's the same issue as the failing test.

```
package main

func main() {
        m := make(map[float64]int)
        m[0]++
        for k, v := range m {
                println(k, v)
        }
}
```

On tinygo, this shows that the map key does not have the expected value.
```
~/go/src/github.com/fastly/dgryski/wasmplay/cmd/structmap $ go run main.go
+0.000000e+000 1
~/go/src/github.com/fastly/dgryski/wasmplay/cmd/structmap $ ~/go/src/github.com/tinygo-org/tinygo/build/tinygo run main.go
+1.383384e-322 1
```

dgryski:
Update: seems like the value that's coming out is the `interface{}` type code (`0x1c`).  The float64 `reflect.Kind` is 14.  Type codes for basic types are that type shifted over one bit, which for float64 makes it 28 == 0x1c.  Using https://evanw.github.io/float-toy/ (or `math.Frombits`) we can see that for a float64, the bit pattern `0x1c` is indeed `+1.383384e-322`.

aykevl:
See https://github.com/tinygo-org/tinygo/pull/2353 which should fix this bug.

PR #2353, opened 2021-12-07
Fix ranging over a map with non-trivially-comparable key types

aykevl:
Some map keys are hard to compare, such as floats. They are stored as if
the map keys are of interface type instead of the key type itself. This
makes working with them in the runtime package easier: they are compared
as regular interfaces.

Iterating over maps didn't care about this special case though. It just
returns the key, value pair as it is stored in the map. This is buggy,
and this commit fixes this bug.

This should fix https://github.com/tinygo-org/tinygo/issues/2339.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -1792,20 +1792,7 @@ func (b *builder) createExpr(expr ssa.Value) (llvm.Value, error) {
 		if expr.IsString {
 			return b.createRuntimeCall("stringNext", []llvm.Value{llvmRangeVal, it}, "range.next"), nil
 		} else { // map
-			llvmKeyType := b.getLLVMType(rangeVal.Type().Underlying().(*types.Map).Key())
-			llvmValueType := b.getLLVMType(rangeVal.Type().Underlying().(*types.Map).Elem())
-
-			mapKeyAlloca, mapKeyPtr, mapKeySize := b.createTemporaryAlloca(llvmKeyType, "range.key")
-			mapValueAlloca, mapValuePtr, mapValueSize := b.createTemporaryAlloca(llvmValueType, "range.value")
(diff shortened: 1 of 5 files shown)


dgryski:
So, `golang/geo/s2` which initially led me to this bug appears to still be failing.  It looks like perhaps it's not handling `0 == -0` in the struct comparison?  But otherwise, this definitely fixes the float-in-map case that I originally filed.  I'll file the `0 == -0` bug as a separate issue.

```
~/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/golang/geo/s2 $ ~/go/src/github.com/tinygo-org/tinygo/build/tinygo test
--- FAIL: TestCellFaces
    edge (-0.707106781186547461715008, -0.707106781186547461715008, 0.000000000000000000000000) counts wrong, got 1, want 2
    edge (-0.707106781186547461715008, -0.707106781186547461715008, -0.000000000000000000000000) counts wrong, got 1, want 2
    edge (-0.000000000000000000000000, -0.707106781186547461715008, -0.707106781186547461715008) counts wrong, got 1, want 2
    edge (0.000000000000000000000000, -0.707106781186547461715008, -0.707106781186547461715008) counts wrong, got 1, want 2
FAIL	github.com/golang/geo/s2	0.574s
FAIL
```

Edit: With this PR merged, this is a minimal reproducer for this negative-zero case: https://github.com/tinygo-org/tinygo/issues/2356

dgryski:
This PR also fixes the failing test for /x/text/message/catalog, which I've now added to my test corpus: https://github.com/dgryski/tinygo-test-corpus/commit/a788627497949af71fab0e9806dffa5bce7cfb08

dgryski (approved):
LGTM

deadprogram:

PR #2355, opened 2021-12-08
src/runtime: strengthen hash function for structs and arrays

dgryski:
Using `|` to combine hash values with slowly turn every bit on.
Hashes combined with `^` with keep more entropy.


aykevl (approved):

deadprogram:

PR #2357, opened 2021-12-08
src/runtime: improve float/complex hashing

dgryski:
This is not perfect, but the best I could do without
revamping all the hash funtions to take a seed.

### DIFF
--- a/src/runtime/hashmap.go
+++ b/src/runtime/hashmap.go
@@ -356,6 +356,24 @@ func hashmapStringDelete(m *hashmap, key string) {
 //go:linkname valueInterfaceUnsafe reflect.valueInterfaceUnsafe
 func valueInterfaceUnsafe(v reflect.Value) interface{}
 
+func hashmapFloat32Hash(ptr unsafe.Pointer) uint32 {
+	f := *(*uint32)(ptr)
+	if f == 0x80000000 {
+		// convert -0 to 0 for hashing
+		f = 0
+	}
+	return hashmapHash(unsafe.Pointer(&f), 4)
+}
+
+func hashmapFloat64Hash(ptr unsafe.Pointer) uint32 {
+	f := *(*uint64)(ptr)
+	if f == 0x8000000000000000 {
+		// convert -0 to 0 for hashing
+		f = 0
+	}
+	return hashmapHash(unsafe.Pointer(&f), 8)
+}
+
 func hashmapInterfaceHash(itf interface{}) uint32 {
 	x := reflect.ValueOf(itf)
 	if x.RawType() == 0 {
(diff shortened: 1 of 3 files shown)


dgryski:
Draft because I still need to write some tests for this.

testdata/map.go
```diff
@@ -109,6 +112,48 @@ func main() {
...
 	testBigMap(squares, 40)
 	println("tested growing of a map")
+
+	floatcmplx()
+}
+
+func floatcmplx() {
+
+	var zero float64
+	var negz float64 = math.Copysign(zero, -1)
```
aykevl: This also works:
```suggestion
	var negz float64 = -zero
```

aykevl (commented):
Looks like you like coffee, but not the decaf kind?

I can see a few ways to implement this:

1. As you've done, by returning a fixed value for a zero floating point value.
2. By checking the bit pattern. For example, you could convert all +0.0 values to -0.0 values (or vice versa).
3. By checking the value, and converting to 0 if necessary

Example for 2:

```go
case reflect.Float32:
    f := (*uint32)(ptr)
    if *f == 0 {
        *f = 0x80000000 // -0.0
    }
    return hashmapHash(ptr, 4)
```

Example for 3:

```go
case reflect.Float32, reflect.Float64:
    f := x.Float()
    if f == 0 {
        f = 0 // not a no-op: this converts -0.0 to +0.0
    }
    return hashmapHash(unsafe.Pointer(f), 8)
```

The advantage of 2) is that it doesn't require any float64 operations, which is important for embedded devices without 64-bit FPU (that is, almost all of them). And it's probably the fastest of all options. The advantage of 3) is that it doesn't change the current behavior much and is maybe easier to read (but that's subjective) - but of course it requires softfloat implementations of float32-to-float64 and float64 comparisons. Even for programs that don't use float64 otherwise.
What you have done also works of course, but it requires a float32 to float64 conversion and comparison between two float64 values. Not sure which is the best one. I'm personally not a big fan of magic numbers but it should definitely work.

testdata/map.go
dgryski: Yeah, I wasn't a huge fan of the hard-coded hash values either.  I'll rewrite this with option number (2).

dgryski: Hmm... just realized we don't want to actually modify the key value that's being passed in, which resetting the value via the pointer I believe is doing.

aykevl: Oops, you're right. Wasn't thinking it through enough. This works around that:

```go
case reflect.Float32:
    f := *(*uint32)(ptr)
    if f == 0 {
        f = 0x80000000 // -0.0
    }
    return hashmapHash(unsafe.Pointer(&f), 4)
```

aykevl: Sidenote: I changed +0.0 to -0.0 in the example because comparing against 0 is slightly more efficient than comparing against other values in some instruction sets. It's a micro-optimization though, and probably doesn't make much difference.

dgryski:
PTAL

testdata/map.go
dgryski: I can only imagine that the difference in comparing against zero vs not is totally outweighed by the actual hash function calculation.

dgryski: Hmm... my approach of having two hash function calls probably has additional code side overhead than just resetting the value.  I'll update my PR.

src/runtime/hashmap.go
```diff
@@ -356,6 +356,24 @@ func hashmapStringDelete(m *hashmap, key string) {
...
 func valueInterfaceUnsafe(v reflect.Value) interface{}
 
+func hashmapFloat32Hash(ptr unsafe.Pointer) uint32 {
+	f := *(*uint32)(ptr)
+	if f == 0x80000000 {
+		// convert -0 to 0 for hashing
+		f = 0
+	}
+	return hashmapHash(unsafe.Pointer(&f), 4)
+}
```
aykevl, line 366: The IR for this looks pretty neat! Translated back to pseudo-Go, it's this:

```go
var stackalloc [1]uint64 // uninitialized stack allocation
stackalloc[0] = 0
val := *ptr
comparison = val == -2147483648
val = comparison ? 0 : val // this is a single LLVM IR instruction
*(*uint32)(&stackalloc) = val
return hashmapHash(&stackalloc, 4)
```

(the `[1]uint64` is because of reasons, it should be optimized)

aykevl: https://github.com/tinygo-org/tinygo/pull/2362

aykevl (approved):

PR #2358, opened 2021-12-08
src/runtime: fix nil map dereference

### LINKED ISSUE #2341: archive/zip: nil pointer deference in TestReader
With `b.RunParallel` stubbed, the `archive/zip` tests now build, but annoyingly do not pass.

The nil pointer panic seems to come from looking up an entry in a nil map (reader.go: 149).
```
func (z *Reader) decompressor(method uint16) Decompressor {
        dcomp := z.decompressors[method] <-- here
        if dcomp == nil {
                dcomp = decompressor(method)
        }
        return dcomp
}
```

dgryski:
Operations on nil maps are accepted and shouldn't
panic. The base hashmapGet/hashmapDelete handled
nil-maps correctly, but the hashmapBinary versions
could segfault accessing the nil map while trying
to hash the key.

Fixes #2341

### DIFF
--- a/src/runtime/hashmap.go
+++ b/src/runtime/hashmap.go
@@ -307,16 +307,24 @@ func hashmapNext(m *hashmap, it *hashmapIterator, key, value unsafe.Pointer) boo
 // Hashmap with plain binary data keys (not containing strings etc.).
 
 func hashmapBinarySet(m *hashmap, key, value unsafe.Pointer) {
+	// TODO: detect nil map here and throw a better panic message?
 	hash := hashmapHash(key, uintptr(m.keySize))
 	hashmapSet(m, key, value, hash, memequal)
 }
 
 func hashmapBinaryGet(m *hashmap, key, value unsafe.Pointer, valueSize uintptr) bool {
+	if m == nil {
+		memzero(value, uintptr(valueSize))
+		return false
+	}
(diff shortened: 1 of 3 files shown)


aykevl (commented):
Looks good to me, but can you add tests for hashmapBinaryGet and hashmapBinaryDelete? It can be as simple as getting and deleting a value from a nil `map[int]int` map in testdata/map.go to check that it doesn't panic.

dgryski:
PTAL.  (I actually had these already and thought they were included in this commit.  Must've missed them..)

aykevl (approved):

dgryski:
The wasmtest keep failing in CI :( https://github.com/tinygo-org/tinygo/runs/4458708630?check_suite_focus=true#step:22:15

aykevl:
Yeah, I'm not sure what's going on there. I needed to fix them when moving to GitHub Actions, but they are still buggy. They worked fine on CircleCI.

I'm suspecting race conditions.

dgryski:
This also handles a failure in dgryski/go-postings: https://github.com/dgryski/tinygo-test-corpus/commit/d3d2fad502f2f986b4d8fc35c5567331be2473e7

deadprogram:
Thank you very much @dgryski for this fix! Now merging.

PR #2359, opened 2021-12-08
net/interface: use internal implementation for itoa.Uitoa

deadprogram:
This PR switches to use the internal implementation for itoa.Uitoa for the recent additions for `net.Interface`. It should fix the CI build issues introduced by #2165

### DIFF
--- a/src/net/interface.go
+++ b/src/net/interface.go
@@ -8,6 +8,7 @@ package net
 
 import (
 	"errors"
+	"internal/itoa"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


src/net/interface.go
```diff
@@ -226,7 +227,7 @@ func (zc *ipv6ZoneCache) name(index int) string {
 		zoneCache.RUnlock()
 	}
 	if !ok { // last resort
-		name = uitoa(uint(index))
+		name = itoa.Uitoa(uint(index))
```
soypat, line 230: nice catch. What about `dtoi` on line 250?

deadprogram: That one IS defined here https://github.com/tinygo-org/tinygo/blob/dev/src/net/parse.go#L14

deadprogram:
Merging so we can fix the CI. Thanks for reviewing @soypat

Issue #2354, opened 2021-12-07
regression: http-get example in tinygo-drivers fails to compile with dev branch


ysoldak:
wifinina http-get example in tinygo-drivers repository fails to compile with tinygo from dev branch head.
It compiles fine with latest tinygo release (0.21.0)

Offending change identified to be ec9fd3fb38c5aa97345cae517d6a15204581b851

```
tinygo build -target=arduino-nano33 -size=full -o test.bin ./examples/wifinina/http-get
# path/filepath
/usr/local/Cellar/go/1.17.2/libexec/src/path/filepath/path.go:382:16: Lstat not declared by package os
/usr/local/Cellar/go/1.17.2/libexec/src/path/filepath/match.go:249:18: Lstat not declared by package os
/usr/local/Cellar/go/1.17.2/libexec/src/path/filepath/match.go:324:16: Stat not declared by package os
/usr/local/Cellar/go/1.17.2/libexec/src/path/filepath/path.go:465:18: Lstat not declared by package os
/usr/local/Cellar/go/1.17.2/libexec/src/path/filepath/path.go:501:18: Lstat not declared by package os
/usr/local/Cellar/go/1.17.2/libexec/src/path/filepath/symlink.go:84:17: Lstat not declared by package os
```

ysoldak:
PR that fails CircleCi due to this:
https://github.com/tinygo-org/drivers/pull/346

dgryski:
Figuring the correct build tags for the `os` package for each platform is a nightmare :(

Why was this not caught when that (or another) PR was merged?

dankegel:
Looking...

dankegel:
By expanding imports, I found a series of simpler reproducers of the form
```
package main

import foo

func main() {
}
```

where foo is one of:
"tinygo.org/x/drivers/net/http"
"crypto/tls"
"path/filepath"

but I'm not sure how helpful that is.

ysoldak:
I mean, no wonder filesystem stuff is not available on "baremetal", shall drivers code be fixed instead that been working by coincidence before? Or shall we have no-op Stat and Lstat stubs for baremetal?

dgryski:
I think `ErrNotImplemented` stubs for everything on baremetal is the way to go, until we create some sort of rudimentary in-memory file system or something...

deadprogram:
Deciding on filesystem for bare metal also depends on the existence of some device to provide this. See https://github.com/tinygo-org/tinyfs

[...]

PR #2362, opened 2021-12-08
transform: allocate the correct amount of bytes in an alloca

aykevl:
When I wrote the code originally, I didn't know about SetAlignment so I hacked a way around it by allocating [...]uintptr types. However, this allocates a few too many bytes in some cases. This commit changes this to only allocate the space that we actually need.

The code size effect is mixed, but generally positive. The combined average of the drivers repo is reduced by 0.27% with more programs being reduced in size than are increasing in size (although I couldn't run all tests due to #2354).


deadprogram:
Very good, this seems much more correct! Now merging, thanks @aykevl

PR #2363, opened 2021-12-09
Makefile: add regexp/syntax to TEST_PACKAGES

dgryski:


dgryski:
Still taking too long during CI

PR #2364, opened 2021-12-09
src/os/stat.go: fix regression on baremetal targets

dkegel-fastly:
Fix https://github.com/tinygo-org/tinygo/issues/2354

Doesn't seem to handle wasi yet, but that is not a regression: 
```
$ build/tinygo test -target wasi os
Error: failed to run main module `/var/folders/v0/0k9hwftd29b18z9z7s2g2sfc0000gn/T/tinygo186587511/main`

Caused by:
    0: failed to instantiate "/var/folders/v0/0k9hwftd29b18z9z7s2g2sfc0000gn/T/tinygo186587511/main"
    1: unknown import: `env::chmod` has not been defined
FAIL	os	0.032s
```

### DIFF
--- a/src/os/stat.go
+++ b/src/os/stat.go
@@ -1,5 +1,3 @@
-// +build !baremetal,!js
-
 // Copyright 2017 The Go Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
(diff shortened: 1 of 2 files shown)


ysoldak:
Verified works for me (http-get example in drivers repo, nano-rp2040 board)

dgryski (approved):
LGTM

deadprogram:
Thank you for the fix @dkegel-fastly and thanks @dgryski and @ysoldak for verifying. Now merging.

PR #2365, opened 2021-12-09
Add support for LLVM 13

aykevl:
This adds support for building with `-tags=llvm13` and switches to LLVM 13 for tinygo binaries that are statically linked against LLVM.

Some notes on this PR:

  * Added `-mfloat-abi=soft` to all Cortex-M targets because otherwise nrfx would complain that floating point was enabled on Cortex-M0. That's not the case, but with `-mfloat-abi=soft` the `__SOFTFP__` macro is defined which silences this warning.
    See: https://reviews.llvm.org/D100372
  * Changed from `--sysroot=<root>` to `-nostdlib -isystem <root>/include` for musl because with Clang 13, even with `--sysroot` some system libraries are used which we don't want.
  * Changed all `-Xclang -internal-isystem -Xclang` to simply `-isystem`, for consistency with the above change. It appears to have the same effect.
  * Moved WebAssembly function declarations to the top of the file in task_asyncify_wasm.S because (apparently) the assembler has become more strict.

@QuLogic pinging you in case you hadn't seen this yet

### DIFF
--- a/builder/cc1as.cpp
+++ b/builder/cc1as.cpp
@@ -115,29 +115,26 @@ bool AssemblerInvocation::CreateFromArgs(AssemblerInvocation &Opts,
   // Any DebugInfoKind implies GenDwarfForAssembly.
   Opts.GenDwarfForAssembly = Args.hasArg(OPT_debug_info_kind_EQ);
 
-  if (const Arg *A = Args.getLastArg(OPT_compress_debug_sections,
-                                     OPT_compress_debug_sections_EQ)) {
-    if (A->getOption().getID() == OPT_compress_debug_sections) {
-      // TODO: be more clever about the compression type auto-detection
-      Opts.CompressDebugSections = llvm::DebugCompressionType::GNU;
-    } else {
-      Opts.CompressDebugSections =
-          llvm::StringSwitch<llvm::DebugCompressionType>(A->getValue())
-              .Case("none", llvm::DebugCompressionType::None)
-              .Case("zlib", llvm::DebugCompressionType::Z)
-              .Case("zlib-gnu", llvm::DebugCompressionType::GNU)
-              .Default(llvm::DebugCompressionType::None);
-    }
+  if (const Arg *A = Args.getLastArg(OPT_compress_debug_sections_EQ)) {
+    Opts.CompressDebugSections =
+        llvm::StringSwitch<llvm::DebugCompressionType>(A->getValue())
+            .Case("none", llvm::DebugCompressionType::None)
+            .Case("zlib", llvm::DebugCompressionType::Z)
+            .Case("zlib-gnu", llvm::DebugCompressionType::GNU)
+            .Default(llvm::DebugCompressionType::None);
   }
 
   Opts.RelaxELFRelocations = Args.hasArg(OPT_mrelax_relocations);
+  if (auto *DwarfFormatArg = Args.getLastArg(OPT_gdwarf64, OPT_gdwarf32))
+    Opts.Dwarf64 = DwarfFormatArg->getOption().matches(OPT_gdwarf64);
   Opts.DwarfVersion = getLastArgIntValue(Args, OPT_dwarf_version_EQ, 2, Diags);
   Opts.DwarfDebugFlags =
       std::string(Args.getLastArgValue(OPT_dwarf_debug_flags));
   Opts.DwarfDebugProducer =
       std::string(Args.getLastArgValue(OPT_dwarf_debug_producer));
-  Opts.DebugCompilationDir =
-      std::string(Args.getLastArgValue(OPT_fdebug_compilation_dir));
+  if (const Arg *A = Args.getLastArg(options::OPT_ffile_compilation_dir_EQ,
+                                     options::OPT_fdebug_compilation_dir_EQ))
+    Opts.DebugCompilationDir = A->getValue();
   Opts.MainFileName = std::string(Args.getLastArgValue(OPT_main_file_name));
 
   for (const auto &Arg : Args.getAllArgValues(OPT_fdebug_prefix_map_EQ)) {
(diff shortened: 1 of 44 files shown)


aykevl:
Huh, for some reason the goroutines test starts to fail on AVR with LLVM 13.

QuLogic:
Backporting the first commit to `v0.21.0` fixes most of the tests except for `TestCompiler/WebAssembly/channel.go`, which errors with the output:
```
    main_test.go:381: failed to run: signal: interrupt
    main_test.go:395: stdout: len, cap of channel: 1 2 false
    main_test.go:395: stdout: len, cap of channel: 0 0 false
    main_test.go:395: stdout: recv from open channel: 1 true
    main_test.go:395: stdout: received num: 2
    main_test.go:395: stdout: received num: 3
    main_test.go:395: stdout: slept
    main_test.go:395: stdout: received num: 4
    main_test.go:395: stdout: received num: 5
    main_test.go:395: stdout: received num: 6
    main_test.go:395: stdout: received num: 7
    main_test.go:395: stdout: received num: 8
    main_test.go:395: stdout: recv from closed channel: 0 false
    main_test.go:395: stdout: complex128: (+7.000000e+000+1.050000e+001i)
    main_test.go:395: stdout: sum of n: 149
    main_test.go:395: stdout: --- test ran too long, terminating...
```
I did not bother with the second commit, as AVR tests are not enabled in 0.21.0. I suppose I should try out `dev` as well some time, but that requires forward-porting the patches.

QuLogic:
Running this against `dev` does appear to get `TestBuild/WebAssembly/channel.go` passing, but other new tests are failing in `TestTest`. I haven't determined if that's related to this PR or `dev` directly.

aykevl:
Maybe it works now? Let's see what CI thinks.

@QuLogic I hope to get a release out soon, with this PR in it. We need to catch up on LLVM versions!

deadprogram:
This PR has passed all CI including TinyHCI!

aykevl:
Ok, ready for review now!
@niaow pinging you because the first commit modifies AVR tests a bit.

builder/cc1as.cpp
```diff
@@ -419,12 +424,12 @@ bool ExecuteAssembler(AssemblerInvocation &Opts, DiagnosticsEngine &Diags) {
     Failed = Parser->Run(Opts.NoInitialTextSection);
   }
 
-  // Close Streamer first.
```
niaow, line 422: Is this no longer an issue? I don't quite understand what is going on here.

cgo/libclang_config.go
```diff
@@ -1,5 +1,5 @@
 // +build !byollvm
-// +build !llvm12
+// +build !llvm12,!llvm13
```
niaow, line 2: Do these build tags still work correctly with LLVM 12 as the current default?

niaow:
I just tried `go test -v -tags llvm13` right now on my system (Manjaro/Arch uses LLVM 13.0.0 by default now) and it seems that:
1. All Linux tests currently segfault (host and emulated)
2. wasi-libc cannot be imported due to missing (u)int*_t types

All other targets seem to work fine, not quite sure what is happening yet.

niaow:
It currently appears to be crashing when starting the libc:
```
(gdb) bt
#0  0x0000000000206b0a in __init_libc (envp=0x7fffffffde28, pn=0x7fffffffe19f "/tmp/tinygo3925493316/main") at lib/musl/src/env/__libc_start_main.c:24
#1  0x0000000000206cdc in __libc_start_main (main=0x2041d0 <main>, argc=<optimized out>, argv=0x7fffffffde18) at lib/musl/src/env/__libc_start_main.c:79
#2  0x0000000000206ad6 in _start ()
```

The debug location unfortunately does not seem to be very accurate - it points at the opening bracket.

niaow:
The crashing instruction (on host) is:
```
mov %fs:0x28,%rax
```

Which is doing something with TLS before TLS is initialized if I understand correctly. I don't quite see why it is there yet.

At that point in the program, GDB reports that the `fs` segment register is 0 (because it has not been set yet).

niaow:
This appears to be a stack check

niaow:
Adding `-fno-stack-protector` to musl's CFLAGS seems to fix all of the segfaults, so the only remaining issue is wasi-libc

niaow:
The `wasi-libc` problem seems to be an issue with `getClangHeaderPath` on Arch. The clang binary is stored at `/usr/bin/clang-13` (with a symlink at `/usr/bin/clang`), but the headers are stored at `/usr/lib/clang/13.0.0/include`. It looks like we were previously pulling in headers from both the main clang header path and the header path from my locally compiled assert copy of clang 12, but after deleting the assert copy it just cannot find some headers.

niaow:
With #2498, this works for me

QuLogic:
I haven't tested the latest version of this, but I did test `dev` and I see the same failure there with LLVM 12 as I do with LLVM13 + this PR. I'm bisecting that now to see what introduced the problem.

QuLogic:
Ah, the problem with `dev` is that some recent PRs pulled in `internal/itoa` and the backport is not visible when running tests (but seems fine for build), so I probably have some messed up `GOPATH` setting or something. I guess previously tests didn't import anything that needed backports in the Go version I was using, so this didn't fail, but was probably still wrong.

deadprogram:
Thanks for looking at this @QuLogic sounds like if there are any further issues we can resolve them individually.

Thanks also @niaow for your testing. Please see my comment on PR #2498 I am doing to merge this PR so you should both change the target branch for your PR and also rebase.

Great work on this @aykevl it is very important for us to be able to use the latest LLVM, not to mention all the upstream fixes you have been contributing.

Now merging!

QuLogic:
OK, so backporting this + 1d2c17753a9a7353a5ef753de7b002d5e625a6d3 to v0.21.0 works on LLVM 12. For LLVM 13, I also needed to backport e4de7b49575c28fcd3a446e7478687e9611d110a.

For `dev` directly, I figured out the problem is that I'm running the equivalent of `go test ./...`, and that picks up tests in `src`, but those should not be run by `go`, but by `tinygo test`. I'm not totally sure why this didn't happen in 0.21 or older, but I've excluded that directory, and everything is good on both LLVM 12 & 13.

builder/cc1as.cpp
aykevl: This file is basically a copy of https://github.com/llvm/llvm-project/blob/main/clang/tools/driver/cc1as_main.cpp with some small modifications. So I'm just syncing up the changes to LLVM 13.

PR #2366, opened 2021-12-09
builder: use build ID as cache key

aykevl:
Instead of storing an increasing version number in relevant packages (compiler.Version, interp.Version, cgo.Version, ...), read the build ID from the currently running executable. This has several benefits:

  * All changes relevant to the compiled packages are caught.
  * No need to bump the version for each change to these packages. This avoids merge conflicts.
  * During development, `go install` is usually enough. No need to run `tinygo clean` most of the time.

Of course, the drawback is that it might be updated a bit more often
than necessary but I think the overall benefit is big.

Regular release users shouldn't see any difference. Because the tinygo binary stays the same, the cache works well.

---

I made this PR mostly to make development for myself easier. I've been running `rm ~/.cache/tinygo/pkg-*` far too often to clean the cache. Previously I just ran `tinygo clean` but since #2110 that has become a lot more expensive.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -24,7 +24,6 @@ import (
 	"strings"
 
 	"github.com/gofrs/flock"
-	"github.com/tinygo-org/tinygo/cgo"
 	"github.com/tinygo-org/tinygo/compileopts"
 	"github.com/tinygo-org/tinygo/compiler"
 	"github.com/tinygo-org/tinygo/goenv"
@@ -64,9 +63,8 @@ type BuildResult struct {
 // implementation of an imported package changes.
 type packageAction struct {
 	ImportPath       string
-	CGoVersion       int // cgo.Version
-	CompilerVersion  int // compiler.Version
-	InterpVersion    int // interp.Version
+	CompilerBuildID  string
+	TinyGoVersion    string
 	LLVMVersion      string
 	Config           *compiler.Config
 	CFlags           []string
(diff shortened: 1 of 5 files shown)


dgryski:
Too bad we can't use Go's `buildid` package directly :(

aykevl:
Yeah, I've so often wanted to use their internal packages. I wish they were importable somehow. Even if they don't provide API stability guarantees, having access to them would be great (for API stability, there is of course version pinning using Go modules).

At the same time, by using GNU build IDs, I think we're also including all the non-Go code (LLVM mostly).

niaow:
Another option could be to take the hash of the executable. This is probably more expensive, but is portable.

Hmm. . . nevermind it might be a bit too expensive.

deadprogram:
Not quite building yet on Windows: https://github.com/tinygo-org/tinygo/runs/4474797055?check_suite_focus=true#step:14:355

aykevl:
I think I've fixed the Windows failure. Let's see whether it passes CI.

aykevl:
> Another option could be to take the hash of the executable. This is probably more expensive, but is portable.
> 
> Hmm. . . nevermind it might be a bit too expensive.

Yeah exactly, that was what I first wanted to do but didn't because it seemed unreasonably expensive. Luckily it's not necessary with build IDs :)

Note that the latest push (to fix Windows) only uses the Go build ID, not the full binary build ID. I think that's not too big of a problem though.

niaow (approved):
I don't know what is going on with tinyhci right now, but this looks good to me

PR #2371, opened 2021-12-10
Additions to os and syscall

dgryski:
Motivated by getting archive/zip tests to pass.

### DIFF
--- a/src/os/file.go
+++ b/src/os/file.go
@@ -182,11 +182,6 @@ func (f *File) Seek(offset int64, whence int) (ret int64, err error) {
 	return 0, &PathError{"seek", f.name, ErrNotImplemented}
 }
 
-// Stat is a stub, not yet implemented
-func (f *File) Stat() (FileInfo, error) {
-	return nil, &PathError{"stat", f.name, ErrNotImplemented}
-}
-
 func (f *File) SyscallConn() (syscall.RawConn, error) {
 	return nil, ErrNotImplemented
 }
--- a/src/os/file.go
+++ b/src/os/file.go
@@ -10,6 +10,7 @@
 package os
 
 import (
+	"errors"
 	"io"
 	"syscall"
 )
@@ -111,14 +112,29 @@ func (f *File) Read(b []byte) (n int, err error) {
 	return
 }
 
+var errNegativeOffset = errors.New("negative offset")
+
 // ReadAt reads up to len(b) bytes from the File at the given absolute offset.
 // It returns the number of bytes read and any error encountered, possible io.EOF.
 // At end of file, Read returns 0, io.EOF.
 func (f *File) ReadAt(b []byte, offset int64) (n int, err error) {
-	n, err = f.handle.ReadAt(b, offset)
-	if err != nil && err != io.EOF {
-		err = &PathError{"readat", f.name, err}
+	if offset < 0 {
+		return 0, &PathError{Op: "readat", Path: f.name, Err: errNegativeOffset}
 	}
+
+	for len(b) > 0 {
+		m, e := f.handle.ReadAt(b, offset)
+		if e != nil {
+			if err != io.EOF {
+				err = &PathError{"readat", f.name, err}
+			}
+			break
+		}
+		n += m
+		b = b[m:]
+		offset += int64(m)
+	}
+
 	return
 }
 
--- a/src/syscall/syscall_libc_darwin.go
+++ b/src/syscall/syscall_libc_darwin.go
@@ -1,3 +1,4 @@
+//go:build darwin
 // +build darwin
 
 package syscall
@@ -172,6 +173,15 @@ func Stat(path string, p *Stat_t) (err error) {
 	return
 }
 
+func Fstat(fd int, p *Stat_t) (err error) {
+	n := libc_fstat(int32(fd), unsafe.Pointer(p))
+
+	if n < 0 {
+		err = getErrno()
+	}
+	return
+}
+
 func Lstat(path string, p *Stat_t) (err error) {
 	data := cstring(path)
 	n := libc_lstat(&data[0], unsafe.Pointer(p))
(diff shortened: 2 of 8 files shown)


src/os/file.go
```diff
@@ -115,10 +116,23 @@ func (f *File) Read(b []byte) (n int, err error) {
 // It returns the number of bytes read and any error encountered, possible io.EOF.
 // At end of file, Read returns 0, io.EOF.
 func (f *File) ReadAt(b []byte, offset int64) (n int, err error) {
-	n, err = f.handle.ReadAt(b, offset)
-	if err != nil && err != io.EOF {
-		err = &PathError{"readat", f.name, err}
+	if offset < 0 {
+		return 0, &PathError{Op: "readat", Path: f.name, Err: errors.New("negative offset")}
```
aykevl: Note that `errors.New` causes a heap allocation. You can avoid this by declaring it as an (unexported) global variable.
(Just a suggestion, not critical since it's not for a resource constrained system).

src/os/file.go
```diff
@@ -115,10 +116,23 @@ func (f *File) Read(b []byte) (n int, err error) {
...
+	}
+
+	for len(b) > 0 {
+		m, e := f.handle.ReadAt(b, offset)
```
aykevl, line 126: You could consider moving this to the implementation of `f.handle.ReadAt`. Not sure about that. It would make `f.handle.ReadAt` also follow the `io.ReaderAt` documentation:

> ReadAt reads len(p) bytes into p starting at offset off in the underlying input source. It returns the number of bytes read (0 <= n <= len(p)) and any error encountered.

src/syscall/syscall_libc_darwin.go
```diff
@@ -188,6 +198,10 @@ func Lstat(path string, p *Stat_t) (err error) {
 //export stat$INODE64
 func libc_stat(pathname *byte, ptr unsafe.Pointer) int32
 
+// int fstat(int fd, struct stat * buf);
+//export fstat$INODE64
+func libc_fstat(fd int, ptr unsafe.Pointer) int32
```
aykevl: Note that a C int us different from a Go int. On a 64 bit architecture, C usually uses a 32 bit int while Go usually uses a 64-bit int. Use int32 to make sure it's always a 32-bit int.

src/syscall/syscall_libc_wasi.go
```diff
@@ -225,6 +235,10 @@ func Lstat(path string, p *Stat_t) (err error) {
 //export stat
 func libc_stat(pathname *byte, ptr unsafe.Pointer) int32
 
+// int fstat(fd int, struct stat * buf);
+//export fstat
+func libc_fstat(fd int, ptr unsafe.Pointer) int32
```
aykevl: Same here.

aykevl (commented):
Just a few notes. I'm aware this is still a WIP.

src/os/file.go
dgryski: The loop upstream is in the `File` methods, not in the `handle` version: https://github.com/golang/go/blob/master/src/os/file.go#L127

src/syscall/syscall_libc_darwin.go
dgryski: Hmm... so what should the type of `fd` be here then?

aykevl: `int32`, to match C. Otherwise you might run into weird ABI issues.

dgryski:
Interesting wasi failure:
```
~/go/src/go.googlesource.com/go/src/archive/zip $ ~/go/src/github.com/tinygo-org/tinygo/build/tinygo test -target=wasi archive/zip
panic: sync: unlock of unlocked Mutex
Error: failed to run main module `/var/folders/5b/_hr1d1fd3qn4t9vtfx5p61wm0000gp/T/tinygo23514247/main`

Caused by:
    1: wasm trap: unreachable
       wasm backtrace:
           0: 0x5863 - <unknown>!runtime._panic
           1: 0x9e19 - <unknown>!(*sync.Mutex).Unlock
           2: 0xa333 - <unknown>!(*sync.Once).Do
           3: 0xaf601 - <unknown>!archive/zip.newFlateReader
           4: 0x9fee3 - <unknown>!(*archive/zip.File).Open
           5: 0xbc18a - <unknown>!archive/zip.readTestFile
           6: 0xba3c8 - <unknown>!archive/zip.TestReader$1
```

Edit: totally different failure when building with `-opt=1`, so I think it's unrelated to this change.  But clearly something to investigate.

Edit: Adding some printlns to the map code makes me think of memory corruption, possibly due to a stack overflow. I saw a similar issue when hunting down a wasi test failure in `golang/snappy`.

dgryski:
Windows runs the stdlib tests, which means I either need to remove `archive/zip` from `TEST_PACKAGES` or figure out how to write `fstat` for Windows.

aykevl:
`fstat` is available on Windows: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fstat-fstat32-fstat64-fstati64-fstat32i64-fstat64i32?view=msvc-170
You can use it like any other libc, like on MacOS for example.
You can use Wine to test. I've written most Windows support on Linux, by testing through Wine.

dkegel-fastly:
OK, looking at the windows side of this, using Wine as Ayke suggests.

But first I suppose I should get dev to pass tests on wine.  
See https://github.com/tinygo-org/tinygo/pull/2390 for first step down that rabbithole.

deadprogram:
#2390 has been merged.

Also there is a merge conflict in this branch. :smiley_cat:

dkegel-fastly:
Per discussion with @dgryski, closing in favor of the updated versions split into https://github.com/tinygo-org/tinygo/pull/2443 and https://github.com/tinygo-org/tinygo/pull/2457

Issue #2374, opened 2021-12-10
Importing "image/color" breaks scheduler=tasks


cstrahan:
Importing `"image/color"` breaks `scheduler=tasks`:

```
$ tinygo version
tinygo version 0.21.0 linux/amd64 (using go version go1.17.1 and LLVM version 11.0.0)
```

```go
// main.go
package main

import _ "image/color"

func main() {
}
```

```
$ tinygo flash -target arduino-mega2560 -scheduler=tasks
Invalid bitcast
i8* bitcast (%runtime._interface (i16, i8*, i8*, i8*) addrspace(1)* @"image/color.rgbaModel" to i8*)
Invalid bitcast
i8* bitcast (%runtime._interface (i16, i8*, i8*, i8*) addrspace(1)* @"image/color.rgba64Model" to i8*)
Invalid bitcast
i8* bitcast (%runtime._interface (i16, i8*, i8*, i8*) addrspace(1)* @"image/color.nrgbaModel" to i8*)
Invalid bitcast
i8* bitcast (%runtime._interface (i16, i8*, i8*, i8*) addrspace(1)* @"image/color.nrgba64Model" to i8*)
Invalid bitcast
i8* bitcast (%runtime._interface (i16, i8*, i8*, i8*) addrspace(1)* @"image/color.alphaModel" to i8*)
Invalid bitcast
i8* bitcast (%runtime._interface (i16, i8*, i8*, i8*) addrspace(1)* @"image/color.alpha16Model" to i8*)
Invalid bitcast
i8* bitcast (%runtime._interface (i16, i8*, i8*, i8*) addrspace(1)* @"image/color.grayModel" to i8*)
Invalid bitcast
i8* bitcast (%runtime._interface (i16, i8*, i8*, i8*) addrspace(1)* @"image/color.gray16Model" to i8*)
Invalid bitcast
i8* bitcast (%runtime._interface (i16, i8*, i8*, i8*) addrspace(1)* @"image/color.yCbCrModel" to i8*)
Invalid bitcast
i8* bitcast (%runtime._interface (i16, i8*, i8*, i8*) addrspace(1)* @"image/color.nYCbCrAModel" to i8*)
Invalid bitcast
i8* bitcast (%runtime._interface (i16, i8*, i8*, i8*) addrspace(1)* @"image/color.cmykModel" to i8*)
```

cstrahan:
Minimal reproducer:

```go
package main

type wrappedFunc struct {
	f func()
}

[...]

PR #2376, opened 2021-12-10
interp: work around AVR function pointers in globals

aykevl:
Not sure how to test this, since we don't really have any AVR tests configured.

See https://github.com/tinygo-org/tinygo/issues/2374


deadprogram:
This is one obscure bug, great job finding it @aykevl thank you!

Now merging.

PR #2380, opened 2021-12-13
os: implement and test os.MkdirTemp

dkegel-fastly:
Also:
- remove one TODO now that CreateTemp exists
- add a testcase for os.stat to expose a problem found by upstream's test for MkdirTemp
- add a forgotten handleSyscallError in os.stat in unix to fix that problem there
- skip tests that expose same problem on Windows for now; it's kind of subtle


src/os/tempfile.go
```diff
@@ -95,6 +95,48 @@ func prefixAndSuffix(pattern string) (prefix, suffix string, err error) {
...
+		}
+		if IsExist(err) {
+			if try++; try < 10000 {
+				continue
+			}
+			return "", &PathError{Op: "mkdirtemp", Path: dir + string(PathSeparator) + prefix + "*" + suffix, Err: ErrExist}
+		}
+		if IsNotExist(err) {
+			_, serr := Stat(dir)
+			ine := IsNotExist(err)
```
dgryski: Should this be checking `serr` instead of `err` ?

dgryski (changes requested):

src/os/tempfile.go
dkegel-fastly: thanks.  That is partially reverted (and evidently broken) debugging code.  Fixing...

dgryski (approved):
LGTM

src/os/tempfile.go
dkegel-fastly: Squashed this fix in:

```
 		if IsNotExist(err) {
-			_, serr := Stat(dir)
-			ine := IsNotExist(err)
-			if ine {
-				return "", serr
+			if _, err := Stat(dir); IsNotExist(err) {
+				return "", err
 			}
 		}
```

to make it mirror the upstream test as intended.

deadprogram:
Thank you @dkegel-fastly you are on a serious roll! Thanks for keeping up the review pace too @dgryski 

Now merging.

PR #2381, opened 2021-12-13
runtime: remove extalloc

niaow:
The extalloc collector has been broken for a while, and it doesn't seem reasonable to fix right now.
In addition, after a recent change it no longer compiles.
In the future similar functionality can hopefully be reintroduced, but for now this seems to be the most reasonable option.


dgryski:
Also update https://github.com/tinygo-org/tinygo-site/blob/release/content/docs/reference/usage/important-options.md ?

aykevl (approved):

deadprogram:
Please resolve merge conflicts @niaow so this can be merged. Thanks!

deadprogram:
Thank you @niaow now merging.

PR #2384, opened 2021-12-14
interp: move initializaton to runtime on interp timeout

### LINKED ISSUE #2383: interp timeout should move computations to runtime, not fail the build
If an `init()` functions runs for "too long" tinygo produces output similar to
```
/Users/dgryski/go/src/go.googlesource.com/go/src/math/rand/rand.go:337:31: interp: running for more than a minute, timing out (executed calls: 312055)
  %0 = load %"math/rand.Rand"*, %"math/rand.Rand"** @"math/rand.globalRand", align 8, !dbg !2802
```
and fails the build.  Instead, that computation should be moved to runtime.

dgryski:
This also resets the timeout timer on each new interpretred top-level function
call.

Fixes #2383

### DIFF
--- a/interp/interpreter.go
+++ b/interp/interpreter.go
@@ -17,9 +17,14 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 	locals := make([]value, len(fn.locals))
 	r.callsExecuted++
 
+	var timeout bool
+
 	if time.Since(r.start) > time.Minute {
 		// Running for more than a minute. This should never happen.
-		return nil, mem, r.errorAt(fn.blocks[0].instructions[0], fmt.Errorf("interp: running for more than a minute, timing out (executed calls: %d)", r.callsExecuted))
+		if r.debug {
+			fmt.Fprintln(os.Stderr, "moving long running function initialization to runtime:", fn.name)
+		}
+		timeout = true
 	}
 
 	// Parameters are considered a kind of local values.
@@ -97,7 +102,7 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 				operands = append(operands, v)
 			}
 		}
-		if isRuntimeInst {
+		if isRuntimeInst || timeout {
 			err := r.runAtRuntime(fn, inst, locals, &mem, indent)
 			if err != nil {
 				return nil, mem, err
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -111,6 +111,7 @@ func Run(mod llvm.Module, debug bool) error {
 		if r.debug {
 			fmt.Fprintln(os.Stderr, "call:", fn.Name())
 		}
+		r.start = time.Now()
 		_, mem, callErr := r.run(r.getFunction(fn), nil, nil, "    ")
 		call.EraseFromParentAsInstruction()
 		if callErr != nil {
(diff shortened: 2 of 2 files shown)


interp/interpreter.go
```diff
@@ -17,9 +17,14 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 	locals := make([]value, len(fn.locals))
 	r.callsExecuted++
 
+	var timeout bool
+
 	if time.Since(r.start) > time.Minute {
 		// Running for more than a minute. This should never happen.
-		return nil, mem, r.errorAt(fn.blocks[0].instructions[0], fmt.Errorf("interp: running for more than a minute, timing out (executed calls: %d)", r.callsExecuted))
+		if r.debug {
```
kenbell, line 24: For MCUs, always logging is useful, without worrying about debug flags?  If there’s a fixed flash budget to meet, would want to be able to easily track down code that interp gave up on and needs attention.

niaow:
I am a bit concerned about our use of the timer here. This change could potentially make builds non-deterministic. If this is going to impact the generated binary, it might make sense to switch to a different timeout mechanism - like a counter.

dgryski:
Yeah, I think we only have the `callsExecuted` counter which I'm not sure is exactly what we want.  I'd love to use a counter instead but that makes it tricky for users to have something that "just works" reasonably.  Maybe we have a timeout for most people, but that's also configurable with a counter limit for people who want more control in reproducible builds?

niaow:
I don't think having a flag or doing non-reproducible builds is really appropriate here.

dgryski:
Well, I can remove the minute timeout and try to add an instruction counter somewhere to trigger the switch instead.  But I do think that a configurable "let it run ~forever" flag somehow might be useful.

aykevl:
I agree with @niaow, I think timing out is a really bad idea. It makes builds not reproducible. And I very strongly prefer reproducible builds for many reasons.

Timing out after a number of calls is better, but I don't really like that either. While it is technically reproducible, it does introduce odd behavior when you're on the fence of hitting the limit:
Imagine you're debugging an interp bug that only happens when recovering from uninterpretable code. Now you remove some unrelated code and suddenly the bug is masked, giving you a false sense of where to look.

aykevl:
@dgryski can you give an example of code that fails in such a way? There is probably an interp bug that can be fixed. The interp package should never run for more than a minute. Essentially, it shouldn't be much slower than what Go would run at init time anyway. It is expected to be slower (because it's an interpreter, not compiled code) but one minute is in almost all cases excessive.
(Also, did you verify that letting it run for longer than a minute actually compiles the code?)

dgryski:
The `bytes` and `strings` package timeout while trying to create random input buffers.  These packages pass if left to run, or if the random initialization is moved to runtime instead (and the panic tests are removed).  The `cloudflare/ahocorasick` package tests also change from erroring with a timeout to passing.

Edit: example of bytes init code that times out: https://github.com/golang/go/blob/master/src/bytes/bytes_test.go#L1693

niaow:
Hmm yeah actually I think that might be a bug or unintended effect from interp. That loop includes a synchronization barrier (or at least it should), so if interp is trying to run that then it may just be effectively unrolling that loop.

Update: no

aykevl:
I also found `tinygo test regexp/syntax`, which (IIRC) fails with the same error.
Basically, I'm afraid we need another redesign of the interp package to make it much faster again.

dgryski:
This is not the correct approach.

PR #2387, opened 2021-12-14
tinygo: set cmd.Dir even when running emulators

### LINKED ISSUE #2367: tinygo test -target wasi compress/flate fails because it needs access to multiple directories
With today's dev:

```
$ tinygo test compress/bzip2
ok  	compress/bzip2	0.533s
$ tinygo test -target wasi compress/bzip2
panic: open testdata/e.txt.bz2: file not found
```
Here's the source line that fails:

https://github.com/golang/go/blob/8ff254e30ba9d9c13747fe42213c3f54b47c58e7/src/compress/bzip2/bzip2_test.go#L216

dgryski:
This allows compress/bzip2 to pass with -target=wasi

Fixes #2367

### DIFF
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -232,6 +232,13 @@ func LoadTarget(options *Options) (*TargetSpec, error) {
 		spec.ExtraFiles = append(spec.ExtraFiles, "src/internal/task/task_asyncify_wasm.S")
 	}
 
+	// TODO(dgryski): handle CFLAGS and LDFLAGS here too?
+	var emu []string
+	for _, s := range spec.Emulator {
+		emu = append(emu, strings.ReplaceAll(s, "{root}", goenv.Get("TINYGOROOT")))
+	}
+	spec.Emulator = emu
+
 	return spec, nil
 }
 
--- a/main.go
+++ b/main.go
@@ -244,7 +244,6 @@ func runPackageTest(config *compileopts.Config, result builder.BuildResult, test
 			flags = append(flags, "-test.run="+testRunRegexp)
 		}
 		cmd = executeCommand(config.Options, result.Binary, flags...)
-		cmd.Dir = result.MainDir
 	} else {
 		// Run in an emulator.
 		args := append(config.Target.Emulator[1:], result.Binary)
@@ -264,6 +263,7 @@ func runPackageTest(config *compileopts.Config, result builder.BuildResult, test
 		}
 		cmd = executeCommand(config.Options, config.Target.Emulator[0], args...)
 	}
+	cmd.Dir = result.MainDir
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	err := cmd.Run()
--- a/targets/wasm.json
+++ b/targets/wasm.json
@@ -13,6 +13,6 @@
 		"--stack-first",
 		"--no-demangle"
 	],
-	"emulator":      ["node", "targets/wasm_exec.js"],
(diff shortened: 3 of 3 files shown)


dkegel-fastly:
LGTM.  Rescues GOOS=windows tinygo test debug/macho , too!

niaow:
Huh. . .  I actually seem to still get an access error:
```
panic: open testdata/e.txt.bz2: errno 2
```

niaow:
This also seems to prevent wasm (js wasm) from even starting at all, since it seems to use a relative path:
```
Error: Cannot find module '/usr/lib/go/src/compress/bzip2/targets/wasm_exec.js'
```

dgryski:
Hmm.. can we use `os.ExpandEnv` to handle constructing the path here?

aykevl (approved):

aykevl (dismissed):
> This also seems to prevent wasm (js wasm) from even starting at all, since it seems to use a relative path:
> 
> ```
> Error: Cannot find module '/usr/lib/go/src/compress/bzip2/targets/wasm_exec.js'
> ```

You are correct, this is a regression.
The easiest solution may be to use the same `{root}` pattern as is used in CFlags. See for example:
https://github.com/tinygo-org/tinygo/blob/06df31944c6af706d6cb8b30a2e92d19060436d2/targets/nrf51.json#L4-L8

dgryski:
Looks like `{root}` isn't expanded everywhere:

```
Error: Cannot find module '/Users/dgryski/go/src/go.googlesource.com/go/src/compress/bzip2/{root}/targets/wasm_exec.js'
```

Where would be the best place to handle that substitution?

niaow:
It looks like we handle it for `CFLAGS` in `compileopts/config.go` when transforming the config values. It might make sense to put it there as well?

dgryski:
I was starting to look at `compileopts.LoadTarget` where we already do a bit of munging (for the asyncify extra files, for example).  So `{root}` would get replaced at load time.  We might even be able to have that replacement run *only* at Load time and not have to do it individually for the CFlags and LDFlags.

dgryski:
`tinygo test -target=wasm compress/bzip2` seems to be failing because the file stuff seems to be unimplemented.  I've added a patch to replace `{root}` in the emulator line so the test at least runs now (even if it doesn't pass.)

dgryski:
This is ready for review now.

niaow (approved):

aykevl:
> I was starting to look at `compileopts.LoadTarget` where we already do a bit of munging (for the asyncify extra files, for example). So `{root}` would get replaced at load time. We might even be able to have that replacement run _only_ at Load time and not have to do it individually for the CFlags and LDFlags.

Actually, I was thinking we should do it the other way round. I don't think I wrote this down anywhere, but the way I see it is that `compileopts.Target` is a read-only version of the combined JSON files. We do modify it in `defaultTarget` but that's essentially generating a similar target struct, just without an underlying JSON file (and is not modified after creation).
So:
- `compileopts.Options`: command line options and some related things
- `compileopts.Target`: JSON file, or inferred from the current system (read-only)
- `compileopts.Config`: combines the two and provides the "true" values in methods, see for example `Config.Scheduler()` that prefers the command line option, then the one specified in the target JSON file, and falls back to the coroutines scheduler (sidenote: that one should probably be updated).

Anyway, that is the design as I have it in my mind.

dgryski:
So, should the logic in this PR be moved to `compileopts.Config.Emulator()` ?

aykevl:
That would be most consistent with the rest of the code.

PR #2388, opened 2021-12-14
internal/task: swap stack chain when switching goroutines

niaow:
This change swaps the stack chain when switching goroutines, ensuring that the chain is maintained consistently.
This is only really currently necessary with asyncify on wasm.


aykevl (approved):
Looks reasonable. A bit unfortunate about the changes for non-asyncify schedulers but I don't see an easy way to avoid that.

PR #2390, opened 2021-12-15
os/path_windows_test.go: don't fail on Wine

dkegel-fastly:
"Just test with Wine", they said.  OK, fine, here we go.

Wine does not fully implement all the quirks of win32.
In particular, it appears to allow a dot to indicate the current directory as a component in extended length paths; native doesn't.
This caused one perhaps overly-strict test case in upstream's path_windows_test.go to fail.
So skip that bit on Wine.

Detecting Wine using os.Getenv is good enough, even though the proper way is to access the registry (yecch).

### DIFF
--- a/src/os/path_windows_test.go
+++ b/src/os/path_windows_test.go
@@ -51,6 +51,12 @@ func TestFixLongPath(t *testing.T) {
 
 // TODO: bring back upstream version's TestMkdirAllLongPath once os.RemoveAll and t.TempDir implemented
 
+// isWine returns true if executing on wine (Wine Is Not an Emulator), which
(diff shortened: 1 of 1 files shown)


aykevl (approved):

deadprogram:

aykevl:
> Detecting Wine using os.Getenv is good enough, even though the proper way is to access the registry (yecch).

Yeah, for tests this seems good enough. If we need to detect it for non-test code (I hope not!), then we might need something more reliable.

For fun: https://github.com/golang/go/issues/18537
Instead of reading a magic pointer, I decided to use a standard DLL function. You might understand why ;)

PR #2393, opened 2021-12-15
builder: handle concurrent library header rename

niaow:
When a library is built concurrently by multiple TinyGo processes, they may sometimes both build the headers. In that case a directory rename may fail due to conflict. This change detects and handles the conflict similar to how GOROOT construction does.

This change addresses #2348

### DIFF
--- a/builder/library.go
+++ b/builder/library.go
@@ -4,6 +4,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/tinygo-org/tinygo/compileopts"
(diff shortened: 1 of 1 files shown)


builder/library.go
```diff
@@ -92,7 +93,23 @@ func (l *Library) load(config *compileopts.Config, tmpdir string) (job *compileJ
...
-				return nil, err
+				switch {
+				case os.IsExist(err):
+					// Another invocation of TinyGo also seems to have already created the headers.
+
+				case runtime.GOOS == "windows" && os.IsPermission(err):
+					// On Windows, a rename with a destination directory that already
+					// exists does not result in an IsExist error, but rather in an
+					// access denied error. To be sure, check for this case by checking
+					// whether the target directory exists.
```
aykevl, line 104: Windows and its quirks...

aykevl (approved):

Issue #1268, opened 2020-07-27
time.Sleep with a negative argument hangs forever


rogpeppe:
commit: 50a677e9b791bbf34eaa9effca231a036a284ce8
platform: itsybitsy-m4

If the duration passed to `time.Sleep` is negative, it seems to hang up forever rather than returning immediately.
This means that an idiom like:

     time.Sleep(time.Until(deadline))

can deadlock if the deadline has passed.

niaow:
I will take a look at this soon to confirm whether this is still a problem and address it.

niaow:
I have confirmed that this issue is still maybe possible in some form, and made a fix: #2394

PR #2394, opened 2021-12-15
runtime: handle negative sleep times

niaow:
This change fixes the edge case where a negative sleep time is provided.
When this happens, the call now returns immediately (as specified by the docs for time.Sleep).

This change addresses #1268


niaow:
An example of a target on which a negative sleep will fail is `metro-m4-airlift` with `-scheduler=none`

deadprogram:
Thank you for this fix @niaow now merging.

PR #2395, opened 2021-12-15
src/reflect: make sure indirect pointers are handled correctly

dgryski:


src/reflect/value.go
```diff
@@ -355,13 +352,16 @@ func (v Value) Len() int {
...
 	case Chan:
-		return chanlen(v.value)
+		return chanlen(v.pointer())
 	case Map:
-		return maplen(v.value)
+		return maplen(v.pointer())
 	case Slice:
-		return int((*sliceHeader)(v.value).len)
+		if v.pointer() == nil {
+			return 0
```
aykevl: A slice is always stored as a pointer in reflect.Value (because it's bigger than a pointer). So I think `v.value == nil` would be more appropriate.
The same is true below: I think `int((*sliceHeader)(v.value).len)` is safe.

src/reflect/value.go
```diff
@@ -355,13 +352,16 @@ func (v Value) Len() int {
...
+		}
+		return int((*sliceHeader)(v.pointer()).len)
 	case String:
-		return int((*stringHeader)(v.value).len)
+		return int((*stringHeader)(v.pointer()).len)
```
aykevl: Same here.
Also, if slices can have a nil value, then strings probably as well. They are very similar as far as reflect is concerned. (I need to think this through a bit more when I'm not tired).

aykevl (commented):
I need to review this when I'm not tired, but here is some early feedback.

src/reflect/type.go
```diff
@@ -364,7 +364,7 @@ func (t rawType) elem() rawType {
 	}
 }
 
-// stripPrefix removes the "prefix" (the first 5 bytes of the type code) from
+// stripPrefix removes the "prefix" (the low 5 bits of the type code) from
```
aykevl, line 367: 👍 my bad

dgryski:
This is now passing all the tests and my test corpus too.

dgryski:
I should probably add a test for the specific failing minimized test cases this patch fixes.

dgryski:
Added a test for the map issue.  Also removed some of the extra checks that seemed to be superfluous.  Running my test corpus now.

dgryski:
Removing the incorrect fix for the indirect array indexing so this can get merged.

dgryski:
Fixed the broken indirect array indexing patch.  Ready for final review.

src/reflect/all_test.go
```diff
@@ -127,7 +127,7 @@ var deepEqualTests = []DeepEqualTest{
 	{&loopy1, &loopy1, true},
 	{&loopy1, &loopy2, true},
 	//{&cycleMap1, &cycleMap2, true},
-	{&cycleMap1, &cycleMap3, false},
+	//{&cycleMap1, &cycleMap3, false},
```
aykevl, line 130: I guess this was passing by accident before?
I just checked which passed and commented those that did not (and verified they were all related to unimplemented features such as maps).

src/reflect/value.go
```diff
@@ -531,7 +525,7 @@ func (v Value) Index(i int) Value {
 			}
 		}
 
-		if size > unsafe.Sizeof(uintptr(0)) {
+		if size > unsafe.Sizeof(uintptr(0)) || v.isIndirect() {
 			// The element fits in a pointer, but the array does not.
```
aykevl: It might be worth updating this comment, because after this PR it's somewhat inaccurate.
Something like:

> The element fits in a pointer, but the array is not stored in the pointer directly.

aykevl (commented):
Looks reasonable to me.
I think I introduced `isIndirect` at a later time, which led to missing some cases where it needed to be handled.

src/reflect/all_test.go
dgryski: Yes, this bug was causing the maps equality code to thing the maps were different before it actually got to comparing the map keys, when it would the fail with unimplemented `reflect.Value.MapKeys()` call.

aykevl (approved):

deadprogram:
Thank you for the fixes @dgryski and thanks for reviewing @aykevl 

Now merging.

PR #2396, opened 2021-12-15
runtime (gc): correct scan bounds

niaow:
This fixes 2 bugs in the GC scan bounds:
1. On AVR, the GC could sometimes read one byte past the end of a block due to the difference between pointer size and alignment.
2. On WASM, the linker does not properly align the marker for the end of the globals section. A manual alignment operation has been added to markGlobals to work around this.

### DIFF
--- a/src/runtime/gc_conservative.go
+++ b/src/runtime/gc_conservative.go
@@ -430,7 +430,17 @@ func markRoots(start, end uintptr) {
 		if start >= end {
 			runtimePanic("gc: unexpected range to mark")
 		}
+		if start%unsafe.Alignof(start) != 0 {
+			runtimePanic("gc: unaligned start pointer")
+		}
+		if end%unsafe.Alignof(end) != 0 {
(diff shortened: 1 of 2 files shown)


src/runtime/gc_conservative.go
```diff
@@ -430,8 +430,18 @@ func markRoots(start, end uintptr) {
...
+			runtimePanic("gc: unaligned start pointer")
+		}
+		if end%unsafe.Alignof(end) != 0 {
+			runtimePanic("gc: unaligned end pointer")
+		}
 	}
 
+	// Reduce the end bound to avoid reading too far on platforms where pointer alignment is smaller than pointer size.
+	// If the size of the range is 0, then end will be slightly below start after this.
+	end -= unsafe.Sizeof(end) - unsafe.Alignof(end)
```
aykevl, line 443: :+1: nice, this should trivially compile down to nothing on non-AVR instruction sets.

aykevl (approved):
LGTM, thanks for debugging this!

Issue #2369, opened 2021-12-10
tinygo test -target wasi os fails with unknown import env::chmod


dkegel-fastly:
With today's dev on mac:

```
$ tinygo test os
ok  	os	0.235s
$ tinygo test -target wasi os
Error: failed to run main module `/var/folders/v0/0k9hwftd29b18z9z7s2g2sfc0000gn/T/tinygo491621975/main`

Caused by:
    0: failed to instantiate "/var/folders/v0/0k9hwftd29b18z9z7s2g2sfc0000gn/T/tinygo491621975/main"
    1: unknown import: `env::chmod` has not been defined
FAIL	os	0.032s
```

aykevl:
Apparently wasi-libc doesn't define `chmod`.

dkegel-fastly:
I gather wasi-filesystem is the part of wasi that defines things like chmod, and hadn't gotten around to that function yet?

https://github.com/WebAssembly/WASI/blob/main/phases/ephemeral/docs.md mentioned but did not define a function path_permissions_set.

https://github.com/WebAssembly/wasi-filesystem/pull/36 appears to move a bunch of cheese, but interestingly defines a function change_file_permissions_at, so maybe they're getting around to it slowly.

dkegel-fastly:
I think this can be closed.  tinygo test -target wasi os gets past this.

PR #2397, opened 2021-12-16
os: Chmod: don't test on wasi yet, wasi-libc does not yet support it

dkegel-fastly:
Lets "tinygo test -target wasi os" compile, if not pass.

For https://github.com/tinygo-org/tinygo/issues/2369

### DIFF
--- a/src/os/os_chmod_test.go
+++ b/src/os/os_chmod_test.go
@@ -1,0 +1,27 @@
+// +build !baremetal,!js,!wasi
+
+// TODO: Move this into os_test.go (as upstream has it) when wasi supports chmod
+
+package os_test
+
+import (
+	. "os"
+	"runtime"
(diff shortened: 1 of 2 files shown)


dgryski (approved):
LGTM

src/os/os_chmod_test.go
```diff
@@ -0,0 +1,27 @@
+// +build !baremetal,!js,!wasi
+
+// TODO: Move this into os_test.go (as upstream has it) when wasi supports chmod
```
niaow: Should this keep the license header from `os_test.go`?

dkegel-fastly: I keep forgetting that...

niaow (approved):
LGTM. Will merge once CI finishes.

PR #2407, opened 2021-12-18
sync: fix concurrent read-lock on write-locked RWMutex

niaow:
This bug can be triggered by the following series of events:
- A acquires a write lock
- B starts waiting for a read lock
- C starts waiting for a read lock
- A releases the write lock

After this, both B and C are supposed to be resumed as a read-lock is available.
However, with the previous implementation, only C would be resumed immediately.
Other goroutines could immediately acquire the read lock, but B would not be resumed until C released the read lock.

### DIFF
--- a/src/sync/mutex.go
+++ b/src/sync/mutex.go
@@ -5,8 +5,6 @@ import (
 	_ "unsafe"
 )
 
-// These mutexes assume there is only one thread of operation and cannot be accessed safely from interrupts.
-
 type Mutex struct {
 	locked  bool
 	blocked task.Stack
--- a/src/sync/mutex.go
+++ b/src/sync/mutex.go
@@ -5,8 +5,7 @@ import (
 	_ "unsafe"
 )
 
-// These mutexes assume there is only one thread of operation: no goroutines,
-// interrupts or anything else.
+// These mutexes assume there is only one thread of operation and cannot be accessed safely from interrupts.
 
 type Mutex struct {
 	locked  bool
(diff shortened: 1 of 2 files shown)


src/sync/mutex.go
```diff
@@ -41,35 +40,132 @@ func (m *Mutex) Unlock() {
...
+		// The mutex is already unlocked.
+		panic("sync: unlock of unlocked RWMutex")
+
+	default:
+		// The mutex is read-locked instead of write-locked.
+		panic("sync: write-unlock of read-locked RWMutex")
+	}
+
+	switch {
+	case rw.maybeUnblockReaders():
```
dgryski, line 88: Ended up reading the stdlib RWMutex implementation to see that they have the same priority of writer->readers over writer->writer lock.

src/sync/mutex.go
```diff
@@ -41,35 +40,132 @@ func (m *Mutex) Unlock() {
...
+	case rw.maybeUnblockWriter():
+		// Transferred to another writer.
+
+	default:
+		// Nothing is waiting for the lock.
+		rw.state = rwMutexStateUnlocked
+	}
 }
 
 func (rw *RWMutex) RLock() {
```
dgryski, line 100: The stdlib docs suggest that a reader could expect to block if there's a pending writer.  This implementation seems like it could starve a writer.

dgryski (changes requested):
Needs either fixes to fairness or documented that it's not.

niaow:
I have added a bit of documentation as requested.

src/sync/mutex.go
niaow: Yeah. Also due to scheduler issues, it is possible for this to happen even if the mutex was implemented fairly.

dgryski (approved):
LGTM

PR #2410, opened 2021-12-19
Stm32 rng l0 l5

kenbell:
All stm32 that support hardware RNG are now implemented, so changing stm32_rng.go logic to only exclude supported MCUs that lack RNG hardware.


kenbell:
This PR depends on an stm32-svd PR to support stm32l552: https://github.com/tinygo-org/stm32-svd/pull/5

deadprogram:
https://github.com/tinygo-org/stm32-svd/pull/5 was merged @kenbell so you should now be clear to proceed.

kenbell:
@deadprogram - thanks.  I think this should be good to go.

deadprogram:
Thanks for the random commits @kenbell now merging. :game_die:

Issue #2392, opened 2021-12-15
builder: use advisory locking


niaow:
When multiple TinyGo processes miss the same cache entry at the same time, they will all build the same code simultaneously. This can waste large amounts of CPU time.

In many cases it should be possible to work around this by using `flock`. However, this is not available everywhere, so the system must still work when it is not present.

This has been split out from #2348.

niaow:
The https://github.com/gofrs/flock package seems to handle all platforms supported by TinyGo, so that should work for us I think.

PR #2411, opened 2021-12-19
builder: use flock to avoid double-compiles

niaow:
This change uses flock (when available) to acquire locks for build operations.
This allows multiple tinygo processes to run concurrently without building the same thing twice.


niaow:
This implements #2392

deadprogram:
Looking good to me @niaow 

@aykevl any feedback before I merge?

sago35 (approved):
Looks good to me.
This change also makes #1355 work correctly.

deadprogram:
Thanks for the improvement @niaow and for the review @sago35 now merging.

PR #2412, opened 2021-12-19
main: run compiler tests and package tests concurrently

niaow:
This PR now changes the following:
- Multiple concurrent builds can be run during the main tests
- `tinygo test` can now build and run multiple tests at once
- `tinygo test` output is now buffered until the previous test finishes, to provide consistent output ordering
- A semaphore has been added to limit the number of concurrent running jobs across all builds
- `runJobs` can now detect and describe dependency cycles
- The time complexity of `runJobs` has been reduced from `O(n^2)` to `O(n*log(n))` while retaining the original job ordering

### DIFF
--- a/builder/jobs.go
+++ b/builder/jobs.go
@@ -4,8 +4,12 @@ package builder
 // parallel while taking care of dependencies.
 
 import (
+	"container/heap"
+	"errors"
 	"fmt"
 	"runtime"
+	"sort"
+	"strings"
 	"time"
 )
 
@@ -29,7 +33,6 @@ type compileJob struct {
 	dependencies []*compileJob
 	result       string // result (path)
 	run          func(*compileJob) (err error)
-	state        jobState
 	err          error         // error if finished
 	duration     time.Duration // how long it took to run this job (only set after finishing)
 }
@@ -44,41 +47,20 @@ func dummyCompileJob(result string) *compileJob {
 	}
 }
 
-// readyToRun returns whether this job is ready to run: it is itself not yet
-// started and all dependencies are finished.
-func (job *compileJob) readyToRun() bool {
-	if job.state != jobStateQueued {
-		// Already running or finished, so shouldn't be run again.
-		return false
-	}
-
-	// Check dependencies.
-	for _, dep := range job.dependencies {
-		if dep.state != jobStateFinished {
-			// A dependency is not finished, so this job has to wait until it
-			// is.
-			return false
-		}
-	}
-
-	// All conditions are satisfied.
-	return true
-}
-
 // runJobs runs the indicated job and all its dependencies. For every job, all
 // the dependencies are run first. It returns the error of the first job that
 // fails.
 // It runs all jobs in the order of the dependencies slice, depth-first.
 // Therefore, if some jobs are preferred to run before others, they should be
 // ordered as such in the job dependencies.
-func runJobs(job *compileJob, parallelism int) error {
-	if parallelism == 0 {
-		// Have a default, if the parallelism isn't set. This is useful for
+func runJobs(job *compileJob, sema chan struct{}) error {
+	if sema == nil {
+		// Have a default, if the semaphore isn't set. This is useful for
 		// tests.
-		parallelism = runtime.NumCPU()
+		sema = make(chan struct{}, runtime.NumCPU())
 	}
-	if parallelism < 1 {
-		return fmt.Errorf("-p flag must be at least 1, provided -p=%d", parallelism)
+	if cap(sema) == 0 {
+		return errors.New("cannot 0 jobs at a time")
 	}
 
 	// Create a slice of jobs to run, where all dependencies are run in order.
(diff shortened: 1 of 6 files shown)


niaow:
I ran this alongside #2411 on my machine (using an assert build of LLVM):
```
with build lock:
real    4m37.297s
user    12m36.769s
sys     2m54.530s

without build lock:
real    0m34.659s
user    14m35.942s
sys     4m15.562s
```

The user and system times do go up, but not by an unexpectedly large amount.

niaow:
Huh, even without #2411 this seems to have improved CI times a bit.

deadprogram:
In any case #2411 has been merged now.

niaow:
It looks like we may have somehow hit an OOM in assert-test-linux. I am a bit unsure whether this means we just need to lower the concurrency. This test does not increase the maximum concurrency.

niaow:
Or actually the OOM may be from the multiple concurrent loader runs. . .

niaow:
The OOM looks like it might be because each tested package stores its own copy of the entire dependency tree, all at once.
Limiting the number of concurrent calls to `Test` in addition seems to fix this.

niaow:
This also gives me a few ideas on how to make the compiler more efficient once this gets merged.

niaow:
The last change made CI 4-6 minutes faster.

deadprogram:
@aykevl any feedback on this PR? We need all the speed improvements on testing we can get.

builder/jobs.go
```diff
@@ -97,64 +79,91 @@ func runJobs(job *compileJob, parallelism int) error {
...
+			case sema <- struct{}{}:
+				// Start a job.
+				job := jobs[heap.Pop(&ready).(int)]
 				if jobRunnerDebug {
-					fmt.Println("## start:   ", jobToRun.description)
+					fmt.Println("## start:   ", job.description)
 				}
-				jobToRun.state = jobStateRunning
-				workerChan <- jobToRun
+				go runJob(job, doneChan)
```
aykevl, line 123: This could theoretically be a performance issue, but now I think about it, it probably isn't. It certainly simplifies the code to not worry about the goroutine worker pool.

main.go
```diff
@@ -1505,3 +1551,121 @@ func main() {
...
+	// Append the output.
+	var prev []byte
+	if len(buf.output) > 0 && !buf.output[len(buf.output)-1].stderr {
+		prev = buf.output[len(buf.output)-1].data
+		buf.output = buf.output[:len(buf.output)-1]
+	}
+	buf.output = append(buf.output, outputEntry{
+		stderr: false,
+		data:   append(prev, data...),
+	})
```
aykevl: I think this can be simplified to the following:

```suggestion
if len(buf.output) == 0 || !buf.output[len(buf.output)-1].stderr {
    buf.output = append(buf.output, outputEntry{
        stderr: false,
    })
}
last := buf.output[len(buf.output)-1]
last.data = append(last.data, data...)
```

At least I think it's simpler :shrug: essentially it first makes sure the last output entry is of the current output type (stdout, stderr) and then just appends to that output entry.

main.go
```diff
@@ -1505,3 +1551,121 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+// testOutputBuf is used to buffer the output of concurrent tests.
+type testOutputBuf struct {
+	mu             sync.Mutex
+	output         []outputEntry
+	stdout, stderr io.Writer
+	outerr, errerr error
```
aykevl, line 1572: To be honest, I don't think it's necessary to handle these errors. Stdout and stderr could theoretically fail, but if that happens, I'm pretty sure we have bigger things to worry about. And we couldn't even report it (as already pointed out above, "There was an error writing to stdout or stderr, so we probbably cannot print this.").

aykevl (commented):
Generally looks good to me. However, a few notes:

* The test output buffering seems rather complicated. I'm not sure how to simplify it with these features, though - although I suspect upstream Go doesn't have these. It is neat that it writes test output right away when it is the last test to run.
* I intend to move more work to jobs eventually, so that the last commit might not be necessary anymore in the future. Essentially, my goal is to only run `go list` per build and move all parsing/typechecking/etc to parallel jobs.

I actually tried to make a PR for this same feature but it never really got very far.

So in summary: LGTM with some (optional) suggestions for improvements

main.go
niaow: It is possible to report that something went wrong by producing a bad exit code - which this code does.

main.go
niaow: There are a few typos here, but I did manage to implement a version of this tweak

niaow:
@aykevl Is this good to merge now?

main.go
aykevl: Hmm, fair enough. Although I still don't really see the point TBH considering that almost no code checks the result of `fmt.Println` for example.

aykevl:
Yeah, LGTM. It's certainly a big improvement.

PR #2419, opened 2021-12-21
docker: remove only subdirectories (which are updated as submodules) …

deadprogram:
This PR corrects the Dockerfile by removing only subdirectories (which are updated as submodules) from lib to keep picolibc file in its correct place. This is required to correctly build any code running bare metal, and will fix the build on the `drivers` repo.


deadprogram:
Merging so the `drivers` repo build can get fixed.

Issue #1185, opened 2020-06-23
loader: goroot caching issue hit on dev


niaow:
The issue with goroot caching hit dev now. I think we need to make a smarter mechanism for keying our cache in order to avoid this continuing to happen during development.

deadprogram:
I think this was already addressed?

niaow:
It was not, but it should be hopefully fixed by #2421

PR #2421, opened 2021-12-22
loader: elminate goroot cache inconsistency

niaow:
This change breaks the merged goroot creation process into 2 steps:
1. List all overrides
2. Construct a goroot with the specified overrides

Now step 2 is cached using a hash of the results from step 1.
This eliminates cache inconsistency, but means that step 1 needs to be run on every build.
This is relatively acceptable, as step 1 only takes about 3 ms (assuming the directory tree is in the OS filesystem cache).

### DIFF
--- a/loader/goroot.go
+++ b/loader/goroot.go
@@ -14,8 +14,8 @@ package loader
 import (
 	"crypto/sha512"
 	"encoding/hex"
+	"encoding/json"
 	"errors"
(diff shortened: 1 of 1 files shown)


niaow:
This addresses #1185, and would mean that `tinygo clean` is no longer needed every time a package is modified.

aykevl (approved):

Issue #2382, opened 2021-12-13
google/open-location-code/go tests fail on wasi due to asynchronous file loading


dgryski:
This `init()` function here https://github.com/google/open-location-code/blob/main/go/olc_test.go#L63 loads some test files asynchronously.  On wasi, this causes the test to fail opening the file.  Changing the init function to load them in sequence allows the test to pass.

dgryski:
cc @niaow for scheduler bugs

dgryski:
I wonder if we're `wasi-libc` is insufficiently thread-safe for how we're using it?

niaow:
What is the failure though?

dgryski:
```
$ wasmtime run  ./go.test --dir=. --dir=.. --dir=../test_data  --dir=../.. -- -test.v
panic: open ../test_data/validityTests.csv: errno 76
Error: failed to run main module `./go.test`

Caused by:
    0: failed to invoke command default
    1: wasm trap: unreachable
       wasm backtrace:
           0: 0x14cd - <unknown>!runtime._panic
           1: 0x2cbd6 - <unknown>!github.com/google/open-location-code/go.mustReadLines
           2: 0x29b94 - <unknown>!github.com/google/open-location-code/go.init#1$1
           3: 0x2cfa8 - <unknown>!github.com/google/open-location-code/go.init#1$1$gowrapper
           4: 0x3067f - <unknown>!tinygo_launch
           5: 0x7c9b - <unknown>!(*internal/task.Task).Resume
           6: 0x107e7 - <unknown>!_start
       note: using the `WASMTIME_BACKTRACE_DETAILS=1` environment variable to may show more debugging information

```

`76` is the wasi "not allowed" error you get when you haven't passed the correct `--dir` parameter to `wasmtime` to allow it access to that particular directory.

niaow:
This bug appears to be caused or triggered by interp.

niaow:
The actual issue here is that interp is placing these on the runqueue, which causes them to run before the inits are run, causing things to break horribly.

dgryski:
This test now passes.

```
dgryski@beekin go % ~/go/src/github.com/tinygo-org/tinygo/build/tinygo test -c -target=wasi
dgryski@beekin go % wasmtime run ./go.test --dir=. --dir=.. -- -test.v
=== RUN   TestCheck
--- PASS: TestCheck
=== RUN   TestEncode
--- PASS: TestEncode
=== RUN   TestDecode
[...]

PR #2422, opened 2021-12-22
interp: run goroutine starts and checks at runtime

niaow:
This change prevents interp from trying to execute goroutine starts or checks.
This fixes a bug where a goroutine started by an init function would run before the init function.

This addresses #2382

### DIFF
--- a/interp/interpreter.go
+++ b/interp/interpreter.go
@@ -197,7 +197,8 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 				// which case this call won't even get to this point but will
 				// already be emitted in initAll.
 				continue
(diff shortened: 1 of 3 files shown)


niaow:
I am not entirely happy with this solution, but I don't really have a better one right now.

src/internal/task/task_asyncify.go
```diff
@@ -45,6 +45,7 @@ type stackState struct {
 
 // start creates and starts a new goroutine with the given function and arguments.
 // The new goroutine is immediately started.
+//go:noinline
```
aykevl: Many places in interp assume no inlining has yet happened, so I think it's safe to leave out these `//go:noinline` pragmas.

aykevl (commented):
The design looks good to me.

src/internal/task/task_asyncify.go
niaow: Alright yeah I just removed these.

deadprogram:
Thanks for the fix @niaow and thanks @aykevl for the review. Now merging.

Issue #2423, opened 2021-12-23
time.Sleep doesn't work as expected on Arduino Nano


cweagans:
I'm doing some testing with an Arduino Nano and it looks like `time.Sleep` doesn't do the right thing. Here's the code I'm using to test:

```go
package main

import (
	"machine"
	"time"
)

const delay = 16

func main() {
	pin := machine.D12
	pin.Configure(machine.PinConfig{Mode: machine.PinOutput})
	for {
		pin.High()
		time.Sleep(time.Millisecond * delay)
		pin.Low()
		time.Sleep(time.Millisecond * delay)
	}
}
```

When I change `delay` to different powers of 2, here is the timing of each iteration of the for loop (as measured by my logic analyzer):

* `2` : `5 us`
* `4` : `5 us`
* `8` : `5 us`
* `16`: `5 us`
* `32`: `33ms`
* `64`: `101ms`
* `128`: `236ms`
* `256`: `506ms`
* `512`: `1.01s` 

I would definitely expect a little variance given that Arduinos aren't exactly high precision devices, but this seems pretty unpredictable.

Compare to the following Arduino sketch:

```c
void setup() {
  pinMode(12, OUTPUT);
}

void loop() {
  const int wait = 2;
  digitalWrite(12, HIGH);
  delay(wait);
  digitalWrite(12, LOW);
  delay(wait);
}
```

Modifying `wait` in the same way as before:

* `2` : `2 ms`
* `4` : `8 ms`
* `8` : `16 ms`
* `16`: `32 ms`
* `32`: `63.9 ms`
* `64`: `127.9ms`
* `128`: `255.9ms`
* `256`: `511.7ms`
* `512`: `1s` 

It seems like mostly accurate timing is doable, but I don't understand enough about tinygo under the hood to know why the behavior is different. I saw a couple of issues open for the Arduino Uno (which also has the atmega328p), so maybe it's something to do with the support for this specific mcu?

While I'd like to get this resolved, I'm more interested in getting up to speed with TinyGo. Can you help me get pointed in the right direction so that I can debug and fix this myself? I'm still coming up to speed with everything, but my goal is to work on TinyGo more regularly. Understanding (hopefully) little problems like this help a ton!

ysoldak:
Seem to be very MCU dependent.
I've verified your sketch on couple of my boards:

Nano 33 BLE (nrf52840)
- `4`:  `8ms`
- `128`: `256ms`
Perfect!

[...]

PR #2428, opened 2021-12-24
Interrupt based time for AVR. 

zdima:
Timer-0 is used as interrupt source.
Keep all time in nanoseconds to simplify the math and increase performance.
Adjust tick cost when timer-0 is reconfigured by PWM (the time precision affected when timer-0 reconfigured).

### DIFF
--- a/src/machine/machine_avr.go
+++ b/src/machine/machine_avr.go
@@ -4,6 +4,7 @@ package machine
 
 import (
 	"device/avr"
+	"runtime/interrupt"
 	"runtime/volatile"
 	"unsafe"
 )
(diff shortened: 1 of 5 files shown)


src/machine/machine_avr.go
```diff
@@ -142,3 +143,77 @@ func (a ADC) Get() uint16 {
...
+	adjustMonotonicTimer()
+
+	// - Unmask interrupt
+	avr.TIMSK0.SetBits(avr.TIMSK0_TOIE0)
+}
+
+func adjustMonotonicTimer() {
+	// adjust the tickNanos
+	tickNanos = currentTickNanos()
+	println("adjustMonotonicTimer tickNanos=", tickNanos, "ticks=", Ticks)
```
eliasnaur: Remove println if it's for debugging during development of this PR.

cweagans:
This PR helps a lot. See test results in related issue:

https://github.com/tinygo-org/tinygo/issues/2423#issuecomment-1000945390

src/machine/machine_avr.go
zdima: removed

src/runtime/runtime_avr.go
```diff
@@ -56,32 +61,49 @@ func postinit() {
...
-		d -= 1
+	waitTill := ticks() + d
+	// recalibrate if we have some time (>100ms) and it was a while when we did it last time.
+	if d > 100000 {
+		now := waitTill - d
+		if nextTimerRecalibrate < now {
+			nextTimerRecalibrate = now + timerRecalibrateInterval
+			machine.AdjustMonotonicTimer()
+		}
+	}
```
zdima, line 86: @aykevl, you had added TODO comment some time back, and I have added code to recalibrate timer with interval 1 minute when sleep is called for 100ms or more. Do you think one minute interval is too short?

deadprogram: I doubt we need to recalibrate more often. Pretty sure one minute is sufficient.

src/machine/machine_avr.go
```diff
@@ -142,3 +143,76 @@ func (a ADC) Get() uint16 {
...
+
+func currentTickNanos() int64 {
+	// this time depends on clk_IO, prescale, mode and OCR0A
+	// assuming the clock source is CPU clock
+	prescaler := int64(avr.TCCR0B.Get() & 0x7)
+	clock := (int64(1e12) / prescaler) / int64(CPUFrequency())
+	mode := avr.TCCR0A.Get() & 0x7
+
+	/*
+		Mode	WGM02	WGM01	WGM00	Timer/Counter		TOP		Update of OCRx at	TOV Flag Set on
```
deadprogram: Looks like small formatting problem here. :smiley_cat:

deadprogram:
This looks like a really great contribution once we can get it merged. Solves a longstanding issue! 

I left some minor feedback. Also perhaps you want to squash a couple of the commits?

src/machine/machine_avr.go
zdima: fixed

zdima:
fixed comments and commits are squash into one.

deadprogram:
@aykevl or anyone else have any further feedback before we merge?

deadprogram:
Thank you for all your work on this contribution @zdima now merging.

src/runtime/runtime_avr.go
aykevl: But `AdjustMonotonicTimer` doesn't calibrate anything? Therefore calling it every minute is totally unnecessary.

In my comment I was referring to calibrating the WDT based on the timer. The timer is already accurate (or as accurate as it can get).

aykevl (commented):
Normally all time related code lives in the runtime. It looks like the only reason it is in the machine package is so that the machine package can call `AdjustMonotonicTimer`. Perhaps it can be moved to the runtime package and called via a special `//go:linkname`? That would avoid exposing lots of internal functions in the machine package.

Also, this PR is probably going to increase power consumption and possibly reduce performance because of the constant interrupts. That may be a reasonable tradeoff, but I'm mentioning it here just in case.

zdima:
> But AdjustMonotonicTimer doesn't calibrate anything? Therefore calling it every minute is totally unnecessary.

The "calibrate" in this case means adjust the increment value for each interrupt, which depends on timer0 settings.
There is code to address time0 changes when made by PWM. But user can change the time0 settings behind out back and this will address this scenario.

> this PR is probably going to increase power consumption and possibly reduce performance because of the constant interrupts.

Agree. User can address this by changing timer0 settings to reduce interrupt intervals. This will reduce power consumption (also  may affect accuracy of timer, which is another tradeoff).

zdima:
> called via a special //go:linkname

Yes, that would be better. I didn't want to expose those internal functions and that would save me some time :(
Should I create new PR with change to use //go:linkname?

aykevl:
Looks like this PR broke `-opt=1` on AVR: https://github.com/tinygo-org/tinygo/issues/2478

> There is code to address time0 changes when made by PWM. But user can change the time0 settings behind out back and this will address this scenario.

Yes, but this is true for many other things. If users use hardware that is used by the runtime (such as Timer0), they should not expect the runtime to magically deal with that. Therefore, this is in my opinion unnecessary.

> Agree. User can address this by changing timer0 settings to reduce interrupt intervals. This will reduce power consumption (also may affect accuracy of timer, which is another tradeoff).

A timer will always increase power consumption, even at low frequency. This varies by chip of course. Other chips (such as nrf) have dedicated low-power clocks for this purpose running from a separate crystal.

Anyway, just wanted to mentioned this to note as a side effect.

PR #2429, opened 2021-12-25
goenv: add GO variable to use alternate go versions

niaow:
This adds a GO environment variable which may be used to select and test other Go versions.

For example:
```
go install golang.org/dl/go1.15.4
go1.15.4 download
GO=go1.15.4 make smoketest
```

### DIFF
--- a/goenv/goenv.go
+++ b/goenv/goenv.go
@@ -12,10 +12,12 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 )
 
 // Keys is a slice of all available environment variable keys.
 var Keys = []string{
+	"GO",
 	"GOOS",
 	"GOARCH",
 	"GOROOT",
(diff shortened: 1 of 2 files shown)


dkegel-fastly:
This is very helpful. 

Should CI use it to verify "make tinygo-test" works with the older gos we support?

niaow:
Maybe, or at least verify that everything builds. That would probably be good for a job running concurrently with `test-linux-build`. I would probably wait for #2412 for that though since `tinygo-test` is quite slow.

niaow:
@aykevl can you take a look at this when you get a chance since it changes GOROOT and GOPATH detection?

fgsch:
Is there any difference to `GOROOT=$(/path/to/alternative/go env GOROOT) tinygo` or this is just for simplicity?

niaow:
> Is there any difference to GOROOT=$(/path/to/alternative/go env GOROOT) tinygo or this is just for simplicity?

Yes, it changes what version of Go is used to run `go list`.

goenv/goenv.go
```diff
@@ -39,15 +41,51 @@ var TINYGOROOT string
...
-			return dir
+		if goos := os.Getenv("GOOS"); goos != "" {
+			return goos
 		}
 		return runtime.GOOS
 	case "GOARCH":
-		if dir := os.Getenv("GOARCH"); dir != "" {
-			return dir
+		if goarch := os.Getenv("GOARCH"); goarch != "" {
+			return goarch
```
aykevl, line 88: :+1:

goenv/goenv.go
```diff
@@ -249,79 +279,38 @@ func getHomeDir() string {
...
 
-	// Check for some standard locations, as a last resort.
-	var candidates []string
-	switch runtime.GOOS {
-	case "linux":
-		candidates = []string{
-			"/usr/local/go",     // manually installed
-			"/usr/lib/go",       // from the distribution
-			"/snap/go/current/", // installed using snap
-		}
```
aykevl, line 287: This removes functionality. Is this intended?
Normally this is used to detect a Go installation when the `go` binary is not in `$PATH`.

aykevl (commented):
Hmm. I'm not so sure about calling `go env` to get these variables. I'm also not sure about adding yet another environment variable, and especially one that is not used by the Go toolchain (although that's pretty obvious in this particular case).

Here is another proposal. What if instead of adding a `GO` environment variable we actually fix all the places in the source code where we simply call `go` and instead always resolve the `go` binary relative to `GOROOT`? For example here:

https://github.com/tinygo-org/tinygo/blob/06df31944c6af706d6cb8b30a2e92d19060436d2/loader/list.go#L27

I think it should be changed to something like this:

```go
    cmd := exec.Command(filepath.Join(goenv.Get("GOROOT"), "bin", "go"), args...)
```

Then you could run `GOROOT=$HOME/sdk/go1.15.4 make smoketest`.

That isn't as easy to use, but I think it's something we have to do anyway. The current behavior is probably not what users would expect.
Once those bugs are fixed, we can maybe add a `GO` environment variable for `GOROOT` detection.

goenv/goenv.go
niaow: The `go list` command already does not use this autodetection, so if that was intended, it does not work.

niaow:
> That isn't as easy to use, but I think it's something we have to do anyway. The current behavior is probably not what users would expect.

I think this behavior is more appropriate, as it uses the behavior defined by the upstream Go implementation. It should also handle things like `GOENV` correctly.

goenv/goenv.go
niaow: The `/usr/local/go` kinda makes sense I guess, but the installation instructions describe adding it to path, so I don't think that makes too much sense here. As for the others, they are automatically in `$PATH`.

niaow:
Or hmm.. .  . that really doesn't work well either. I am not really sure of any approach which would provide behavior that makes any sense.

PR #2430, opened 2021-12-25
os: add Perm stub on Go 1.15

niaow:
This adds a stub for Perm on Go 1.15, which allows the OS package to compile again.

### DIFF
--- a/src/os/file_go_other.go
+++ b/src/os/file_go_other.go
@@ -1,3 +1,4 @@
+//go:build !go1.16
 // +build !go1.16
 
 package os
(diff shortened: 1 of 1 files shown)


niaow:
This can be tested locally with #2429, although I think I need to come up with a better long-term solution.

dkegel-fastly:
Tested with #2429 and something like
```
$ go install golang.org/dl/go1.15.15@latest
$ go1.15.15 download
$ go install
$ GO=go1.15.15 TINYGO=$HOME/go/bin/tinygo make tinygo-test
$ TINYGO=$HOME/go/bin/tinygo make tinygo-test
```

dkegel-fastly (approved):
LGTM

Issue #2400, opened 2021-12-17
wasmtest: TestChan failing randomly in CI


niaow:
It seems that TestChan is failing occasionally in CI under `assert-test-linux` due to a timeout.
This could just be a timeout (10 seconds does seem like it should be plenty though) or it could be a bug.

niaow:
I am somewhat unsure of what is going on. No approach I try seems to be able to get this test (the section inside the timeout) to take over 250 ms locally (running it pinned to a single core to try to CPU-starve it, running it repeatedly, etc.).

zdima:
I have PR #2431 (draft) that I believe will address this issue. 
Is there a way to trigger CI without making additional commits or better run “reduced” test on CI environment with “make wasmtest” only?

deadprogram:
> Is there a way to trigger CI without making additional commits

I can help you with that. I will retrigger this right now as another test. Also please see my comments.

dkegel-fastly:
This is happening a lot still.

deadprogram:
I think this was addressed some time ago. Now closing, please reopen if needed. Thanks everyone!

PR #2431, opened 2021-12-27
fix wasmtest randomly failing in CI

zdima:
Fix https://github.com/tinygo-org/tinygo/issues/2400
The result is not what expected when the go routine start faster and println("2") finish printing before println("4") does.

### DIFF
--- a/tests/wasm/chan_test.go
+++ b/tests/wasm/chan_test.go
@@ -22,7 +22,6 @@ func TestChan(t *testing.T) {
 		chromedp.Navigate(server.URL+"/run?file=chan.wasm"),
 		waitLog(`1
 4
-2
 3
 true`),
 	)
--- a/tests/wasm/chan_test.go
+++ b/tests/wasm/chan_test.go
@@ -18,14 +18,17 @@ func TestChan(t *testing.T) {
 	ctx, cancel := chromectx()
 	defer cancel()
 
-	err = chromedp.Run(ctx,
-		chromedp.Navigate(server.URL+"/run?file=chan.wasm"),
-		waitLog(`1
+	for i := 0; i < 40; i++ {
+		err = chromedp.Run(ctx,
+			chromedp.Navigate(server.URL+"/run?file=chan.wasm"),
+			waitLog(`1
 4
+2
 3
 true`),
-	)
-	if err != nil {
-		t.Fatal(err)
+		)
+		if err != nil {
+			t.Fatal(err)
+		}
 	}
 }
--- a/tests/wasm/setup_test.go
+++ b/tests/wasm/setup_test.go
@@ -162,7 +162,7 @@ func waitInnerTextMatch(sel string, re *regexp.Regexp) chromedp.QueryAction {
 				return nodes, err
 			}
 			if !re.MatchString(ret) {
-				// log.Printf("found text: %s", ret)
+				log.Printf("found text: [%s]\n", ret)
 				return nodes, errors.New("unexpected value: " + ret)
 			}
 
--- a/tests/wasm/setup_test.go
+++ b/tests/wasm/setup_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 	"net/http"
 	"net/http/httptest"
 	"os/exec"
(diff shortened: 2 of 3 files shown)


niaow:
This really should not have any impact, since scheduling order should be deterministic with the current implementation (unless we have a scheduler bug here. . . which is probably the case if this is in fact a fix). Now that I look at it, the channel is also buffered which means that only the recv from the outer goroutine blocks. It might make sense to update this test instead so that the output is guaranteed by the Go spec if we are going to change this.

deadprogram:
Maybe the test should be something like this?

```go
	println("1")
	go func() {
                println("2")
		ch <- true
	}()
	v := <-ch
	println(v)
```

zdima:
I think the test intent is to verify the '4' seen before '3' - main routine block until . 
And more I look at it, more I think my fix also going to fail sometime.
How about to have a small sleep before println("2")? 
That for sure make "4" appear before "2", and result will be as expected:
1, 4, 2, 3, true.

zdima:
> Maybe the test should be something like this?
> 
> ```go
> 	println("1")
> 	go func() {
>                 println("2")
> 		ch <- true
> 	}()
> 	v := <-ch
> 	println(v)
> ```

This will also produce the same result "all the time" assuming channel works.

niaow:
The issue I have here is that the TinyGo scheduler does not work like that. It runs a goroutine until it blocks, then it moves onto the next one. The order should be fully deterministic as-is, unless something is horribly wrong (which I am worried about).

niaow:
The `1`, `4`, `2`, `3`, `true` output should always be the output with the current implementation. This test is not guaranteed with future implementations, but with the current one this should always be correct unless we have a serious bug.

zdima:
> 

In this case, the debug output must be added to see why the waitFor is not finishing.
Let me try this in my PR and will see.

zdima:
Why it is not failing when we need it ?! ;)
@deadprogram , can you start "Linux / assert-test-linux" again? 
Can I start it on my own without extra commit and your help?

deadprogram:
See https://github.com/tinygo-org/tinygo/runs/4644837804?check_suite_focus=true#step:22:7

zdima:
yes, I saw this. As I understand the test, the WASM program uses stdout to print a messages, and those messages (I think) added to special HTML element with is "log". Since the content of that element is empty, it can indicate the map of stdout to element didn't worked. 
Could it be a timing issue, where log element is not ready to map with stdout, but we didn't wait for it assuming the user didn't care about log and didn't create element? 
I am have zero experience with WASM and I think we need someone who understand this internals to debug farther.

niaow:
It looks like we may also not currently handle the error from `WebAssembly.instantiateStreaming`, so if that failed it would do so silently (I think?). I can check that out later.

aykevl:
> Could it be a timing issue

Most certainly yes. My observations:

* This was never an issue on CircleCI, it only started to be an issue on GitHub Actions
* It seems to happen less frequently on slower computers? I can easily reproduce it on my laptop, which is not very fast.
* Running multiple tests in parallel seems to increase the chance for it to happen, which is why I disabled that in https://github.com/tinygo-org/tinygo/pull/2289/commits/ea7f6185506e52302f58b4cf6097383b1d736e89.

So I'm pretty sure this is a timing issue or race condition. And my guess is that it is somewhere in the test or in chromedp, not in the TinyGo scheduler.

aykevl:
I'm also a bit worried about this PR because it hides the bug instead of fixing it.

zdima:
This PR is for debugging the issue. Adding debug messages since we don’t know what is going on inside. definitely looks like timing issue:

> Could it be a timing issue, where log element is not ready to map with stdout, but we didn't wait for it assuming the user didn't care about log and didn't create element?

zdima:
> So I'm pretty sure this is a timing issue or race condition. And my guess is that it is somewhere in the test or in chromedp, not in the TinyGo scheduler.

@aykevl, @deadprogram (@niaow)
This particular test is failing too often. I am sure you see this failure with your PRs as well.
Do you think this test case should be temporary disabled until someone from WASM group can fix it?

deadprogram:
I think this can be closed now. Thanks everyone who worked on it!

PR #2434, opened 2021-12-28
Add support for the stm32f469-discovery board

eliasnaur:


deadprogram:
Hello @eliasnaur looks like you need to rebase this branch against `dev` please.

eliasnaur:
I did that already and GitHubs says there are no conflicts. Did I miss anything?

EDIT: rebased and force-pushed.

kenbell:
@eliasnaur - i want to go through these changes in a bit of detail which may take a day or two.  It's really great you've unified the code - and there's some patterns here that we can re-use in other STM32 families.

I notice some constants have changed, e.g. now:
```go
	if config.Frequency > 10000 {
		min = 4 // fast mode (Fm)
	}
```
the F407 implementation was 100k, not 10k - 100K intuitively seemed like it was the correct value, but i need to double-check the documentation.
```go
	if config.Frequency > 100000 {
		min = 4 // fast mode (Fm)
	}
```

I've spun up your branch on the stm32f407 disco, and basic blink and UART looks great.  Have you had the change to test I2C or SPI?

In the meantime, would be great to update `README.md` to add the new target and also to add the target to the `smoketests` command in the `Makefile`.

eliasnaur:
> @eliasnaur - i want to go through these changes in a bit of detail which may take a day or two. It's really great you've unified the code - and there's some patterns here that we can re-use in other STM32 families.
> 

Thanks! I suspect runtime_stm32f405.go can be replaced as well, but I'm not proficient enough to make that call, and don't have a 405 board to test on.

> I notice some constants have changed, e.g. now:
> 
> ```go
> 	if config.Frequency > 10000 {
> 		min = 4 // fast mode (Fm)
> 	}
> ```
> 
> the F407 implementation was 100k, not 10k - 100K intuitively seemed like it was the correct value, but i need to double-check the documentation.
> 
> ```go
> 	if config.Frequency > 100000 {
> 		min = 4 // fast mode (Fm)
> 	}
> ```
> 

Indeed, I mentioned the frequency limit in one of the commit, but should have made that change clearer. In particular because I don't know what the correct limit is (or are, if they vary per board).

> I've spun up your branch on the stm32f407 disco, and basic blink and UART looks great. Have you had the change to test I2C or SPI?
>

I don't have a 407, so I assume you mean testing 469. I don't have any hardware other than the 469 board; can I test I2C and SPI without?
 
> In the meantime, would be great to update `README.md` to add the new target and also to add the target to the `smoketests` command in the `Makefile`.

Done.

src/machine/machine_stm32f4.go
```diff
@@ -592,3 +615,180 @@ func initRNG() {
...
+}
+
+func (i2c *I2C) getFreqRange(config I2CConfig) uint32 {
+	// all I2C interfaces are on APB1
+	clock := CPUFrequency() / 4
+	// convert to MHz
+	clock /= 1000000
+	// must be between 2 MHz (or 4 MHz for fast mode (Fm)) and 50 MHz, inclusive
+	var min, max uint32 = 2, 50
+	if config.Frequency > 10000 {
```
kenbell: Per RM0386, §29.3 and RM0090, §27.3 standard mode is up-to 100KHz, fast mode up to 400KHz.  In fast-mode, minimum is 4MHz peripheral input clock.  It looks like the original F405 code was incorrect.   Can you change this to 100K.

README.md
```diff
@@ -118,6 +118,7 @@ The following 75 microcontroller boards are currently supported:
 * [ST Micro "Nucleo" WL55JC](https://www.st.com/en/evaluation-tools/nucleo-wl55jc.html)
 * [ST Micro STM32F103XX "Bluepill"](https://stm32-base.org/boards/STM32F103C8T6-Blue-Pill)
 * [ST Micro STM32F407 "Discovery"](https://www.st.com/en/evaluation-tools/stm32f4discovery.html)
+* [ST Micro STM32F469 "Discovery"](https://www.st.com/content/st_com/en/products/evaluation-tools/product-evaluation-tools/mcu-mpu-eval-tools/stm32-mcu-mpu-eval-tools/stm32-discovery-kits/32f469idiscovery.html)
```
kenbell, line 121: Can you also increment count by 1 at the top of the list

Makefile
```diff
@@ -453,6 +453,10 @@ ifneq ($(STM32), 0)
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=stm32f4disco-1      examples/pwm
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=stm32f469disco      examples/blinky1
+	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=stm32f469disco      examples/blinky2
```
kenbell: It's probably fine just to have a single blinky built as part of the smoketest, can remove blinky2 IMO

kenbell (changes requested):
Requesting a few minor changes (see inline comments).  The I2C comparison should be 100K, and just a couple over minor housekeeping bits and pieces.

I've tested I2C and GPIO on stm32f407-disco and works great.  My testing of SPI wasn't successful - but it was equally broken before this change, so needs separate investigation / PR.

Overall - this looks great.  With a couple small changes, I think this is good to go.

kenbell:
I've confirmed SPI is also working great on stm32f407 - earlier problems were with my setup.

src/machine/machine_stm32f4.go
eliasnaur: Thanks for checking up on this!

eliasnaur:
Thank you for your thorough review. All requested changes should be done now.

kenbell (approved):

kenbell:
@eliasnaur - thanks for the great contribution. Merging now.

PR #2435, opened 2021-12-28
Makefile: update tool search for LLVM 12

niaow:
This updates the tool search to detect build tools suffixed by the version 12, and prefers them over version 11.


kenbell (approved):
lgtm

i had a change locally that just switched clang-11 --> clang-12, but this looks better since it detects.

kenbell:
actually - @niaow  should this change also update go.mod to take the latest go-llvm that defaults to llvm-12?

niaow:
No, it should not.

aykevl (approved):

aykevl:
> should this change also update go.mod to take the latest go-llvm that defaults to llvm-12?

That would be a good idea though - in a separate PR.

PR #2443, opened 2021-12-29
Read, ReadAt: handle partial reads, add test for length zero reads.

dkegel-fastly:
This is part 1 of https://github.com/tinygo-org/tinygo/pull/2371 plus a corresponding test.


dkegel-fastly:
wait, upstream does have a test for this, let me pull that in.

dkegel-fastly:
Upstream tests were illuminating.   There are a few more we should import, but this is enough for today.

TestRead0 exposed a similar problem for Read, so I pulled in the upstream fix.

deadprogram:
Thanks @dkegel-fastly looking good. Now merging.

PR #2444, opened 2021-12-29
runtime (chan): support blocking without a scheduler

niaow:
This change modifies the implementation of channels such that they can still block without a scheduler.
An example that works with this change is `examples/systick` which sends to a channel from an interrupt.

### DIFF
--- a/src/runtime/chan.go
+++ b/src/runtime/chan.go
@@ -39,6 +39,10 @@ func chanDebug(ch *channel) {
 	}
 }
 
+// pretendTask is a placeholder used when no scheduler is available.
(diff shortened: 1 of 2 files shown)


niaow:
It currently uses a fake task for compatibility, which is kinda messy. I am not entirely sure of a good way to do this which doesn't involve either making the existing implementation much more complex or maintaining an entire additional implementation of channels.

niaow:
Hmm. Actually, I might have a better idea once coroutines are removed.

niaow:
This is very out of date.

Issue #2368, opened 2021-12-10
tinygo test -target wasi compress/zlib fails early?


dkegel-fastly:
With today's dev on mac, compress/zlib seems to crash very early on target wasi:
```
$ tinygo test  compress/zlib
ok  	compress/zlib	0.633s
$ tinygo test -target wasi compress/zlib
Error: failed to run main module `/var/folders/v0/0k9hwftd29b18z9z7s2g2sfc0000gn/T/tinygo646659090/main`

Caused by:
    0: failed to invoke command default
    1: wasm trap: unreachable
       wasm backtrace:
           0: 0x78bbf - <unknown>!asyncify_stop_unwind
           1: 0x767a9 - <unknown>!tinygo_launch
           2: 0x6281 - <unknown>!(*internal/task.Task).Resume
           3: 0x1d248 - <unknown>!_start
       note: using the `WASMTIME_BACKTRACE_DETAILS=1` environment variable to may show more debugging information
       
FAIL	compress/zlib	0.043s
```

dgryski:
With optimizations off, this crashes in the conservative garbage collector.  With `-gc=leaking`, the tests pass.  Note that `wasmtime` also needs `--dir=..` to allow the test package to access `../testdata`.

```
~/go/src/go.googlesource.com/go/src/compress/zlib $ ~/go/src/github.com/tinygo-org/tinygo/build/tinygo test -target=wasi -c -opt=1 compress/zlib
~/go/src/go.googlesource.com/go/src/compress/zlib $ WASMTIME_BACKTRACE_DETAILS=1 wasmtime run ./zlib.test --dir=. --dir=.. -- -test.v
=== RUN   TestDecompressor
--- PASS: TestDecompressor
=== RUN   TestWriter
Error: failed to run main module `./zlib.test`

Caused by:
    0: failed to invoke command default
    1: wasm trap: out of bounds memory access
       wasm backtrace:
           0: 0x212e9 - runtime.markRoots
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/gc_conservative.go:414:11
           1: 0x20cd3 - runtime.markStack
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/gc_stack_portable.go:31:12
           2: 0x20b7e - runtime.GC
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/gc_conservative.go:355:11
[...]

PR #2447, opened 2021-12-30
targets (wasi/wasm): raise default stack size to 16 KiB

niaow:
In some cases, 8 KiB is not enough for both the C stack and the asyncify stack.
This gets the compress/zlib tests to fail without crashing.


niaow:
See #2368 for context

aykevl (approved):
Seems reasonable. (16kB seems a lot to me but I'm used to MCUs so :shrug:)

dkegel-fastly:
I was doing part of this locally already, so thanks!

PR #2433, opened 2021-12-28
ci: try to build LLVM with thread support

aykevl:
This should fix a number of concurrency/threading issues.

PR #2448, opened 2021-12-30
internal/task: remove coroutines

niaow:


niaow:
This is a bit bigger than I expected, so I don't think it is necessarily a good idea to include this for this release. Next release cycle I would like to refactor a lot of the things that used to depend on coroutines.

compileopts/config.go
```diff
@@ -159,31 +159,6 @@ func (c *Config) OptLevels() (optLevel, sizeLevel int, inlinerThreshold uint) {
...
-		// regular function pointer.
-		return "doubleword"
-	case "none", "coroutines":
-		// As "doubleword", but with the function pointer replaced by a unique
-		// ID per function signature. Function values are called by using a
-		// switch statement and choosing which function to call.
-		// Pick the switch implementation with the coroutines scheduler, as it
-		// allows the use of blocking inside a function that is used as a func
-		// value.
-		return "switch"
```
aykevl, line 181: FYI: this does have an effect on AVR which used `-scheduler=none` by default. This used to be necessary because function pointers used to be buggy in AVR but I believe it should work now (except a very small increase in binary size sometimes which I think is acceptable).

aykevl (commented):
Phew! That's indeed a big PR. I did expect it to be pretty big though...

I think I would have tried to split this into multiple commits (remove coroutines scheduler, remove switch implementation of functions, and then remove the `parentHandle` parameter). But I'm not going to complain.

We might want to test AVR some more just to be sure, see below.

Anyway, giving a preliminary LGTM now.

niaow:
> I think I would have tried to split this into multiple commits

Yeah, I can try to do that

niaow:
This PR has now been split into 3 less-huge commits.

aykevl (approved):
Looks good to me!
Feel free to fix (probably trivial) merge conflicts and merge the PR.

niaow:
I need to look into why there is an issue with the closure on Windows.

niaow:
It seems to be a random error. . . which is concerning. I also distinctly remember running into that exact error before (so I don't think it is caused by this PR). I strongly suspect that the actual fix is #2433, which I now realize I forgot to finish debugging.

niaow:
I created #2551 for now

niaow:
IDK, does anyone object to merging this now?

dkegel-fastly:
I'm all for merging it if it will prevent bad user experiences.  We can always resurrect coroutines later...

PR #2453, opened 2021-12-31
WIP: math/bits: use LLVM intrinsics

niaow:
This has some issues with WASM right now, so it still needs work.
Mostly putting this here so that I don't forget about it again.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -716,6 +716,10 @@ func (c *compilerContext) createPackage(irbuilder llvm.Builder, pkg *ssa.Package
 			// Create the function definition.
 			b := newBuilder(c, irbuilder, member)
 			if member.Blocks == nil {
+				if cd, ok := compDefs[b.llvmFn.Name()]; ok {
+					// This is a compiler-defined function.
+					b.fillCompDef(cd)
+				}
 				continue // external function
 			}
 			b.createFunction()
--- a/compiler/intrinsics.go
+++ b/compiler/intrinsics.go
@@ -102,3 +102,91 @@ func (b *builder) createMathOp(call *ssa.CallCommon) (llvm.Value, bool) {
 	}
 	return b.CreateCall(llvmFn, args, ""), true
 }
+
+// fillCompDef fills the body of a compiler-defined function.
+func (b *builder) fillCompDef(cd compdef) {
+	// Create a basic block.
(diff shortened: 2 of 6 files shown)


niaow:
This is currently blocked by building `compiler-rt` for wasm.

compiler/compiler.go
```diff
@@ -716,6 +716,10 @@ func (c *compilerContext) createPackage(irbuilder llvm.Builder, pkg *ssa.Package
 			// Create the function definition.
 			b := newBuilder(c, irbuilder, member)
 			if member.Blocks == nil {
+				if cd, ok := compDefs[b.llvmFn.Name()]; ok {
+					// This is a compiler-defined function.
+					b.fillCompDef(cd)
+				}
```
aykevl, line 722: We have lots of compiler-overridden functions already. We currently replace `*ssa.Call` instructions to them with their value inline, but perhaps it makes sense to replace the contents of the functions instead (at least, in some cases). See the `syscall.Syscall*` functions. In any case, I think we should be consistent and do one or the other, I would be fine with changing most of those call replacements to replacing the definition of functions as a refactor.

aykevl (commented):
Is there a reason why you copied math/bits? If not absolutely needed, I'd prefer to keep using the upstream version.

niaow:
> Is there a reason why you copied math/bits?

I had done it this way since it required replacing almost every function in `math/bits`.
This can be done in the compiler, but it would effectively become a full rewrite of `math/bits`. . .  but within the compiler.

compiler/compiler.go
niaow: I think I can turn this into a replacement in this case.

niaow:
Additionally there are some other smaller details like using the runtime's divide-by-zero panic function (the upstream version tries to linkname it, but it works differently in the upstream version).

niaow:
One thing which I just realized is that the rotate functions do not currently have the specified constant-time property on AVR. . .

aykevl:
> > Is there a reason why you copied math/bits?
> 
> I had done it this way since it required replacing almost every function in `math/bits`.

Right, I see. That seems reasonable.

> Additionally there are some other smaller details like using the runtime's divide-by-zero panic function (the upstream version tries to linkname it, but it works differently in the upstream version).

We might want to switch to how upstream does it some day, so that runtime panics can be caught. But that's for a separate PR.

dgryski:
Dup https://github.com/tinygo-org/tinygo/pull/3620

Issue #1906, opened 2021-05-26
<undefined reference> problem while compiling a binary for Arm from Linux


cheer-Yuan:
Hello,

I am trying to compile a program of ARM with tinygo from Linux, when I enter the following command : 
```
GOARCH=arm tinygo build -o <OutputName> ./main.go
```
I got :
```
/usr/lib/gcc-cross/arm-linux-gnueabihf/9/../../../../arm-linux-gnueabihf/bin/ld: /tmp/tinygo859602615/main.o: in function `syscall.Seek':
/usr/local/go/src/syscall/syscall_linux_arm.go:60: undefined reference to `syscall.seek'
```
The relevant codes indicated here above are :
```
// Underlying system call writes to newoffset via pointer.
// Implemented in assembly to avoid allocation.
func seek(fd int, offset int64, whence int) (newoffset int64, err Errno)

func Seek(fd int, offset int64, whence int) (newoffset int64, err error) {
	newoffset, errno := seek(fd, offset, whence)
	if errno != 0 {
		return 0, errno
	}
	return newoffset, nil
}
```
I have just use
```
sudo apt-get install arm-linux-gnueabihf
```
to install this package.

However,
```
GOARCH=arm go build -o <OutputName> ./main.go
```
works without any problem.
Using CGO_ENABLED="0" flag doesn't change anything.
I have searched on the Internet and found no similar issues. Thank you for any possible information.

Environment :
```
Linux 5.4.0-73-generic #82-Ubuntu SMP Wed Apr 14 17:39:42 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
go version go1.16.3 linux/amd64
CGO_ENABLED="1"
GOROOT="/usr/local/go"
TINYGOROOT="/usr/local/lib/tinygo"
```

dkegel-fastly:
Also seems to affect i386.  See log
https://github.com/tinygo-org/tinygo/runs/4687146838?check_suite_focus=true
from https://github.com/tinygo-org/tinygo/pull/2457

aykevl:
I have a local branch to implement Go assembly. It's not complete (only works on amd64 and arm64 now) but once finished that branch should fix this issue.

dkegel-fastly:
May be fixed already by https://github.com/tinygo-org/tinygo/pull/2571 in 0.22?

Please try 0.22 or dev and see if that works better.

aykevl:
This is fixed now, see #2571.

PR #2457, opened 2021-12-31
src/{syscall, os}: add Fstat, with tests

dkegel-fastly:
This is the Fstat part of https://github.com/tinygo-org/tinygo/pull/2371, plus tests and a windows implementation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -262,6 +262,9 @@ TEST_PACKAGES_BASE = \
 
 # Standard library packages that pass tests natively
 TEST_PACKAGES = \
+ifneq ($(OS),Windows_NT)   # archive/zip requires ReadAt, which is not yet supported on windows
+	archive/zip \
+endif
 	$(TEST_PACKAGES_BASE)
 
 # Standard library packages that pass tests on wasi
--- a/Makefile
+++ b/Makefile
@@ -261,11 +261,15 @@ TEST_PACKAGES_BASE = \
 	unicode/utf8 \
 
 # Standard library packages that pass tests natively
-TEST_PACKAGES = \
-ifneq ($(OS),Windows_NT)   # archive/zip requires ReadAt, which is not yet supported on windows
-	archive/zip \
-endif
+TEST_PACKAGES := \
 	$(TEST_PACKAGES_BASE)
+
+# archive/zip requires ReadAt, which is not yet supported on windows
+ifneq ($(OS),Windows_NT)
+TEST_PACKAGES := \
+	$(TEST_PACKAGES) \
+	archive/zip
+endif
 
 # Standard library packages that pass tests on wasi
 TEST_PACKAGES_WASI = \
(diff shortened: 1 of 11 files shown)


dkegel-fastly:
Beyond the current build failure, I think this doesn't work on wasi yet.

dkegel-fastly:
On wasi, I think this requires https://github.com/tinygo-org/tinygo/pull/2469 before it'll pass.
On i386 and arm, it seems to need syscall.seek to be implemented, not sure why.
( cf. https://github.com/tinygo-org/tinygo/issues/1906 ).

dkegel-fastly:
This now works around #1906 by leaving Fstat a stub on i386-linux and arm-linux.

The workaround is a single commit that should be reverted once tinygo supports go assembly.

dkegel-fastly:
May need a tweak to let wasi tests pass still.  I will push a new version tonight.

dgryski (approved):
LGTM.  Can we add `archive/zip` to the passing tests yet?

dkegel-fastly:
Oh, yeah, when I rebase tonight I will try adding that.

dkegel-fastly:
archive/zip is not passing on wasi yet, so I added it to the passing-on-native-tests list.

I don't quite understad *why* it isn't passing on wasi... probably means this pull request isn't fully baked yet.

dkegel-fastly:
Also not passing on windows.  Back to the kitchen!

dgryski:
Started looking into why it's not passing on wasi and fell down the rabbit hole of `io/fs` interfaces and implementations.

dkegel-fastly:
I broke my gaze away from that, too.

dkegel-fastly:
aha, archive/zip failure on windows is expected, it needs ReadAt.  Excluding that test on windows.

dgryski:
The `TestFSModTime` test (which is the failing test on wasi) passes when run independently.  This points to a garbage collection issue or similar.  However, running with `gcAsserts = true`  fails:

This is worth opening its own bug for.

```
~/go/src/go.googlesource.com/go/src/archive/zip $ tinygo test -target=wasi archive/zip
panic: runtime error: gc: metadata array is too small
Error: failed to run main module `/var/folders/5b/_hr1d1fd3qn4t9vtfx5p61wm0000gp/T/tinygo272422465/main`

Caused by:
    1: wasm trap: wasm `unreachable` instruction executed
       wasm backtrace:
           0: 0x9bc6 - <unknown>!runtime.runtimePanic
           1: 0x23d43 - <unknown>!runtime.calculateHeapAddresses
           2: 0x2d431 - <unknown>!_start

FAIL	archive/zip	2.088s
```

dkegel-fastly:
Let's land this now, link to the new bug, and deal with that later.

dgryski:
Filed https://github.com/tinygo-org/tinygo/issues/2554

dkegel-fastly:
Ship it!

deadprogram:
OK that was a bit complicated to follow at first, but I understand the reasoning behind how this PR is setup. Merging, counting on you both @dkegel-fastly and @dgryski to help get it sorted out once the "real" solution is available.

PR #2458, opened 2022-01-01
main (test): run tests on AVR

niaow:
This is still somewhat WIP. This doesn't intend to fix any of these bugs, just to ensure that we do not pick up more.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -6,6 +6,7 @@ package main
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"flag"
 	"fmt"
 	"io/ioutil"
@@ -72,6 +73,7 @@ func TestCompiler(t *testing.T) {
 	}
 
 	t.Run("Host", func(t *testing.T) {
+		t.Parallel()
 		runPlatTests(optionsFromTarget("", sema), tests, t)
 	})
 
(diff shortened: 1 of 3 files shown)


main_test.go
```diff
@@ -127,34 +129,107 @@ func TestCompiler(t *testing.T) {
...
 
 	t.Run("EmulatedRISCV", func(t *testing.T) {
+		t.Parallel()
 		runPlatTests(optionsFromTarget("riscv-qemu", sema), tests, t)
 	})
 
+	t.Run("AVR", func(t *testing.T) {
+		// LLVM backend crash:
+		// LIBCLANG FATAL ERROR: Cannot select: t3: i16 = JumpTable<0>
+		// This bug is non-deterministic, and only happens when run concurrently with non-AVR tests.
```
aykevl, line 144: Ugh. Yeah that's bad. I hope we can fix this (after this PR is merged, in LLVM probably).

aykevl (commented):
Thanks a lot! This looks reasonable. In fact, I've wanted to write this patch for some time now but never got around to actually do it.

niaow:
Huh, actually this seems ready

aykevl (approved):

PR #2459, opened 2022-01-01
compiler: work around AVR atomics bugs

niaow:
The AVR backend has several critical atomics bugs.
This change invokes libcalls for some atomic operations on AVR.
Now `testdata/atomic.go` compiles and runs correctly.

This came up when working on #2458, as it triggered a full crash rather than a compiler error or test failure.

### DIFF
--- a/compiler/atomic.go
+++ b/compiler/atomic.go
@@ -1,6 +1,7 @@
 package compiler
 
 import (
+	"fmt"
 	"strings"
 
 	"golang.org/x/tools/go/ssa"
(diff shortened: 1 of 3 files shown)


niaow:
This PR has been rebased and now enables the atomics test on AVR.

niaow:
Huh, I am not sure yet why the test fails in CI but passes locally.

niaow:
Ah, it looks like recent versions of GCC emulate double-precision floating point, and older ones do not: https://github.com/gcc-mirror/gcc/commit/f30dd607669212de135dec1f1d8a93b8954c327c

I am a bit unsure of why that is being used though (nevermind that is just from the `atomic.Value` tests).

niaow:
I have disabled the test for now, and added a descriptive comment instead.

niaow:
This has now been rebased and simplified a bit

compiler/atomic.go
```diff
@@ -16,6 +17,20 @@ func (b *builder) createAtomicOp(call *ssa.CallCommon) (llvm.Value, bool) {
 	case "AddInt32", "AddInt64", "AddUint32", "AddUint64", "AddUintptr":
 		ptr := b.getValue(call.Args[0])
 		val := b.getValue(call.Args[1])
+		if strings.HasPrefix(b.Triple, "avr") {
+			// AtomicRMW does not work on AVR as intended:
+			// - There are some register allocation issues (fixed by https://reviews.llvm.org/D97127 which is not yet in a usable LLVM release)
```
aykevl, line 22: This is part of LLVM 13, so that particular issue shouldn't be a concern anymore.

compiler/atomic.go
```diff
@@ -73,6 +88,23 @@ func (b *builder) createAtomicOp(call *ssa.CallCommon) (llvm.Value, bool) {
 	case "StoreInt32", "StoreInt64", "StoreUint32", "StoreUint64", "StoreUintptr", "StorePointer":
 		ptr := b.getValue(call.Args[0])
 		val := b.getValue(call.Args[1])
+		if strings.HasPrefix(b.Triple, "avr") {
+			// SelectionDAGBuilder is currently missing the "are unaligned atomics allowed" check for stores.
+			vType := val.Type()
+			isPointer := vType.TypeKind() == llvm.PointerTypeKind
+			if isPointer {
+				// libcalls only supports integers, so cast to an integer.
```
aykevl, line 96: Technically they just don't care, they only see a bunch of bytes that they need to store. But that doesn't really matter.

main_test.go
```diff
@@ -140,7 +140,8 @@ func TestBuild(t *testing.T) {
 		for _, t := range tests {
 			switch t {
 			case "atomic.go":
-				// Not supported due to unaligned atomic accesses.
+				// Requires GCC 11.2.0 or above for interface comparison.
+				// https://github.com/gcc-mirror/gcc/commit/f30dd607669212de135dec1f1d8a93b8954c327c
```
aykevl, line 144: Sidenote: some day we'll drop the avr-gcc and avr-libc dependencies so that this case becomes supported.

aykevl (approved):
Looks good to me.

deadprogram:
Thanks @niaow for working on this, and thanks for reviewing @aykevl now merging.

aykevl:
I have created a patch to fix this in the AVR backend of LLVM: https://reviews.llvm.org/D117725

PR #2460, opened 2022-01-01
testing: implement Skip

niaow:
The Skip methods were left unimplemented due to a lack of runtime.Goexit.
While runtime.Goexit does not currently run defers, it is close enough to the intended functionality that all current tests we have pass with it.

### DIFF
--- a/src/testing/benchmark.go
+++ b/src/testing/benchmark.go
@@ -181,14 +181,23 @@ func (b *B) launch() {
 func (b *B) Run(name string, f func(b *B)) bool {
 	b.hasSub = true
 	sub := &B{
-		common:    common{name: name},
+		common: common{
+			name:     name,
+			finished: make(chan struct{}),
+		},
 		benchFunc: f,
 		benchTime: b.benchTime,
 	}
-	if sub.run1() {
-		sub.run()
-	}
-	b.add(sub.result)
+	go func() {
+		defer close(sub.finished)
+
+		if sub.run1() {
+			sub.run()
+		}
+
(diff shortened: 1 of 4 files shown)


niaow:
Time to run the test suite normally with a purged cache:
```
real    1m53.792s
user    21m35.909s
sys     3m10.601s
```

Time to run the test suite with `-short` with a purged cache:
```
real    0m53.079s
user    19m8.192s
sys     2m46.130s
```

We are still really bottlenecked by building, but it does save 2 minutes of CPU time from a few very slow tests (although I guess I ran this with an assert build. . . ).

dgryski:
Nice.  I think we'll be able to add a few more stdlib packages when this lands.  Maybe regexp?

src/testing/benchmark.go
```diff
@@ -181,14 +181,23 @@ func (b *B) launch() {
...
+			finished: make(chan struct{}),
+		},
 		benchFunc: f,
 		benchTime: b.benchTime,
 	}
-	if sub.run1() {
-		sub.run()
-	}
-	b.add(sub.result)
+	go func() {
```
aykevl, line 191: I very intentionally avoided running tests in a goroutine, to reduce memory consumption. In fact, this was the main reason why I decided to reimplement the testing package instead of reusing the standard library version.
I think we should continue to avoid goroutines in the testing package.

aykevl (commented):
Instead of goroutines, we can implement this functionality using `recover` (#2331). That will be much more light-weight.

PR #2462, opened 2022-01-01
compiler: support RawSyscall.  os: implement Pipe for darwin.

dankegel:
At bare minimum, one should be able to write a short message to the write end of a Pipe, then read it from the read end.

I didn't see an obvious test like that upstream, so I threw one together.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -1440,7 +1440,7 @@ func (b *builder) createFunctionCall(instr *ssa.CallCommon) (llvm.Value, error) 
 			return b.emitSV64Call(instr.Args)
 		case strings.HasPrefix(name, "(device/riscv.CSR)."):
 			return b.emitCSROperation(instr)
-		case strings.HasPrefix(name, "syscall.Syscall"):
+		case strings.HasPrefix(name, "syscall.Syscall") || strings.HasPrefix(name, "syscall.RawSyscall"):
 			return b.createSyscall(instr)
 		case strings.HasPrefix(name, "syscall.rawSyscallNoError"):
 			return b.createRawSyscallNoError(instr)
--- a/src/os/pipe_test.go
+++ b/src/os/pipe_test.go
@@ -1,0 +1,56 @@
+//go:build windows || darwin || (linux && !baremetal && !wasi)
+// +build windows darwin linux,!baremetal,!wasi
+
+// Copyright 2021 The Go Authors. All rights reserved.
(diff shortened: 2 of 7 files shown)


dkegel-fastly:
assert-test-linux looks like a false pass; if you look at the "test stdlib packages" step, there is an error.

It feels like this test found about three problems...?

aykevl:
Huh, weird. Without further investigation, I'd look into #2412 which changes the `tinygo test` logic.

dkegel-fastly:
And... maybe underneath pipe is all still a stub.

dkegel-fastly:
I have a draft of unstubbed pipe on my local tree for mac passing this test, will finish when I get a chance.

dkegel-fastly:
All is well except on linux, which is suffering from #2466 and #2474.
Once #2466 is fixed, this will go red; once #2474 is fixed, this will go green.

dkegel-fastly:
Includes Ayke's suggested fix for RawSyscall now.

deadprogram:
Please correct merge conflict @dankegel thank you.

compiler/compiler.go
```diff
@@ -1435,7 +1435,8 @@ func (b *builder) createFunctionCall(instr *ssa.CallCommon) (llvm.Value, error)
 			return b.emitSV64Call(instr.Args)
 		case strings.HasPrefix(name, "(device/riscv.CSR)."):
 			return b.emitCSROperation(instr)
-		case strings.HasPrefix(name, "syscall.Syscall"):
+		case strings.HasPrefix(name, "syscall.Syscall") || strings.HasPrefix(name, "syscall.RawSyscall"):
+			// TODO: implement b.createRawSyscall for real to reduce overhead slightly
```
aykevl: Which overhead? TinyGo just calls syscalls directly without any stack switching overhead like in upstream Go (in fact, the call is inlined so there's no real way to optimize it further). So this comment seems misleading.

Also, note that `b.createRawSyscall` is a different 'raw' than `syscall.RawSyscall`.

aykevl (commented):
Looks good to me, except for one nit.

deadprogram:
Also, please resolve merge conflict @dankegel thank you!

compiler/compiler.go
dankegel: Removed comment.

deadprogram:
Thanks @dankegel now merging.

PR #2465, opened 2022-01-01
compiler: add correct debug location to init instructions

aykevl:
This adds proper debug locations to interp errors. For example, when trying to use the unicode package on AVR (which currently doesn't work), the following error is shown with this PR:

    /usr/local/go1.17/src/unicode/casetables.go:13:31: interp: ptrtoint integer size does not equal pointer size

Before this PR, that error was not as detailed:

    unicode/<init>:13:31: interp: ptrtoint integer size does not equal pointer size

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -78,6 +78,7 @@ type compilerContext struct {
 	diagnostics      []error
 	astComments      map[string]*ast.CommentGroup
 	pkg              *types.Package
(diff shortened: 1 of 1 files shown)


niaow (approved):
LGTM

aykevl:
I also found the AVR bug, will submit a fix separately. It's a one-line patch.

deadprogram:
Thanks everyone! Now merging.

PR #2468, opened 2022-01-01
compiler: fix emission of large object layouts

aykevl:
Large object layouts don't fit in a pointer-sized integer and therefore
need to be stored in a global instead. However, the way the data was
stored in these globals was not correct for buffers that don't have
pointers near the end. This commit fixes this issue by using math/big
FillBytes() instead of Bytes().

This gets the unicode package to compile on AVR.


dgryski:
Does this affect wasm at all?

aykevl:
Yes, it affects all architectures. But it is much more likely to trigger on AVR. On WebAssembly, it would only trigger for objects that are over 104 bytes in size.

niaow (approved):
LGTM

deadprogram:
Thank you, all. Now merging.

PR #2470, opened 2022-01-01
avr: fix time.Sleep() in init code

### LINKED ISSUE #2464: runtime (avr): timer code fails with -opt=1 or during init()
When debugging `testdata/goroutine.go`, I came up with this minimal test case (the loop isn't necessary, but it makes things easier to test on the actual hardware):
```
package main

import "time"

func init() {
        time.Sleep(time.Millisecond)
}

func main() {
        for {
                println("hello")
        }
}
```

If you check the scheduler output, the chip goes to sleep at which point simavr recognizes that it cannot wake up. This also reproduced on real hardware (my Arduino Uno). It reproduces with or without a scheduler.

Another way to replicate this is to run `examples/blinky1` with `-opt=1`.

aykevl:
In the early days of TinyGo, the idea of `postinit` was to enable interrupts only after initializers have run. Which kind of makes sense... except that `time.Sleep` is allowed in init code and `time.Sleep` requires interrupts to be enabled. Therefore, interrupts must be enabled while initializers are being run.

This commit simply moves the enabling of interrupts to a point right before running package initializers. It also removes `runtime.postinit`, which is not necessary anymore (and was only used on AVR).

This fixes #2464.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -171,9 +171,6 @@ func TestCompiler(t *testing.T) {
 			case "float.go", "math.go", "print.go":
 				// Stuck in runtime.printfloat64.
 
-			case "goroutines.go":
-				// The main() never runs.
-
 			case "interface.go":
 				// Several comparison tests fail.
 
--- a/src/runtime/runtime_arm7tdmi.go
+++ b/src/runtime/runtime_arm7tdmi.go
@@ -28,8 +28,6 @@ var _sidata [0]byte
 //go:extern _edata
 var _edata [0]byte
 
-func postinit() {}
-
 // Entry point for Go. Initialize all packages and call main.main().
 //export main
 func main() {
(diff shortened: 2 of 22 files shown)


niaow:
It looks like `testdata/goroutines.go` now passes with this change. Can you enable it so that this gets tested?

aykevl:
Yes, indeed! Thanks for testing! I've updated the PR.

niaow:
The Windows CI seems to have hit a very weird bug, where the closure-captured value changes from 20 to 0. I am not really sure what happened here though? I can't really think of a way that this could cause that.

aykevl:
Probably a fluke. I re-ran the test and now it passes. Something to remember though if we see it again, because it could indicate a serious memory issue.

niaow (approved):

PR #2471, opened 2022-01-01
all: use compiler-rt from LLVM

aykevl:
This is important to stay up to date with developments in LLVM. The compiler-rt builtins don't change very often, but sometimes they do. And the latest LLVM versions have some fixes in them for AVR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,15 +22,16 @@ commands:
     steps:
       - restore_cache:
           keys:
-            - llvm-source-13-v1
+            - llvm-source-13-v3
       - run:
           name: "Fetch LLVM source"
           command: make llvm-source
       - save_cache:
-          key: llvm-source-13-v1
+          key: llvm-source-13-v3
           paths:
             - llvm-project/clang/lib/Headers
             - llvm-project/clang/include
+            - llvm-project/compiler-rt
             - llvm-project/lld/include
             - llvm-project/llvm/include
   hack-ninja-jobs:
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -35,10 +35,11 @@ jobs:
         uses: actions/cache@v2
         id: cache-llvm-source
         with:
-          key: llvm-source-13-macos-v1
+          key: llvm-source-13-macos-v2
           path: |
             llvm-project/clang/lib/Headers
             llvm-project/clang/include
+            llvm-project/compiler-rt
             llvm-project/lld/include
             llvm-project/llvm/include
       - name: Download LLVM source
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -38,10 +38,11 @@ jobs:
         uses: actions/cache@v2
         id: cache-llvm-source
         with:
-          key: llvm-source-13-linux-v1
+          key: llvm-source-13-linux-v2
           path: |
             llvm-project/clang/lib/Headers
             llvm-project/clang/include
+            llvm-project/compiler-rt
             llvm-project/lld/include
             llvm-project/llvm/include
       - name: Download LLVM source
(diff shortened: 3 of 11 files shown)


QuLogic:
So the aim here would be to install the version of `compiler-rt` that matches LLVM? So I'd have to download the corresponding source tarball when building a package instead of using the fixed version from before?

aykevl:
> So I'd have to download the corresponding source tarball when building a package instead of using the fixed version from before?

The compiler-rt library should be part of a release build. In the case of binary releases from us, it would be packaged just like picolibc for example. In the case of release builds for Linux distributions, it should somehow be shipped together with the package (either bundled with the tinygo package or as a dependency). Users shouldn't need to download compiler-rt manually.

Not sure what you were asking but I hope that answers is.

aykevl:
It looks like some tests are failing due to https://github.com/espressif/llvm-project/issues/52.

aykevl:
With this PR and some local changes (because AVR doesn't use compiler-rt yet), I was able to compile testdata/print.go on AVR:

```
$ tinygo run -target=atmega1284p ./testdata/print.go
Loaded 31530 .text at address 0x0
Loaded 143 .data
hello world!..
42..
100000000..
abc..
a b c..
123..
123..
-123..
12345..
12345..
-12345..
12345678..
12345678..
-12345678..
123456789012..
123456789012..
-123456789012..
-Inf..
+3.Ojt���e+000..
(+0.000000e+000+0.000000e+000i)..
(0:nil)..
map[2]..
true false..
```

... not quite correct yet, but getting close. And I think the last issues can be solved with https://reviews.llvm.org/D86547.

QuLogic:
> > So I'd have to download the corresponding source tarball when building a package instead of using the fixed version from before?
> 
> The compiler-rt library should be part of a release build. In the case of binary releases from us, it would be packaged just like picolibc for example. In the case of release builds for Linux distributions, it should somehow be shipped together with the package (either bundled with the tinygo package or as a dependency). Users shouldn't need to download compiler-rt manually.
> 
> Not sure what you were asking but I hope that answers is.

Yes, I'm asking as a packager, not a user. I assume we can't use a dependency on a pre-compiled package, because tinygo would need the compiled-for-random architecture versions and not just the native ones.

aykevl:
> Yes, I'm asking as a packager, not a user. I assume we can't use a dependency on a pre-compiled package, because tinygo would need the compiled-for-random architecture versions and not just the native ones.

TinyGo has many different architectures. You _could_ precompile all of them but that might not be reasonable, considering that we might end up with like 30 different architecture variants (see builder/builder_test.go: every target and CPU combination could need a different build).
Would it be possible to package compiler-rt builtins in source form? Normally TinyGo expects that releases bundle these builtin sources, but with a patch it can also read the sources from somewhere else. Note that we only need the builtins, not all the sanitizer runtimes (but they wouldn't harm either).

QuLogic:
> Would it be possible to package compiler-rt builtins in source form? Normally TinyGo expects that releases bundle these builtin sources, but with a patch it can also read the sources from somewhere else. Note that we only need the builtins, not all the sanitizer runtimes (but they wouldn't harm either).

Yes, this is how I've been packaging it so far, but that was when the version was different from the system one (which was always tied to the LLVM version). Now that tinygo's compiler-rt is tied to the LLVM version (with the PR), the question naturally arises whether the packaged version could be used. But the answer appears to be no, since the packaged version is just the native build and not every arch that tinygo supports, so things shall remain as they were.

aykevl:
All tests are passing now.

> Yes, this is how I've been packaging it so far, but that was when the version was different from the system one (which was always tied to the LLVM version). Now that tinygo's compiler-rt is tied to the LLVM version (with the PR), the question naturally arises whether the packaged version could be used. But the answer appears to be no, since the packaged version is just the native build and not every arch that tinygo supports, so things shall remain as they were.

:+1:

The version doesn't have to match LLVM exactly. Ideally it's the same version as the one TinyGo is built against, but one earlier or later is usually fine. So if Fedora already packages compiler-rt in source form, that could also be used instead.

aykevl:
Rebased.

I'd like to see this included because it is required for an upcoming change that will greatly improve AVR support.

deadprogram:
Let's do this, seems like a very good idea overall. Now merging, thanks @aykevl for working on it and @QuLogic for the feedback.

aykevl:
Thank you!
I just realized this also requires LLVM 14 (a few important compiler-rt bugs affecting AVR have been fixed in LLVM 14), but we have to do that anyway soon.

Issue #2466, opened 2022-01-01
assert-test-linux "test stdlib packages" reports success despite test failing to link?


dkegel-fastly:
In the CI results for https://github.com/tinygo-org/tinygo/pull/2462, 
in the assert-test-linux CI run, 
in the "test stdlib packages" step, there is an error linking one test, but the build does not terminate.

Log fragment:
```
tinygo:ld.lld: error: undefined symbol: syscall.RawSyscall
>>> referenced by zsyscall_linux_amd64.go:1661 (/opt/hostedtoolcache/go/1.17.5/x64/src/syscall/zsyscall_linux_amd64.go:1661)
>>>               /tmp/tinygo2693500925/main.o:(os_test.TestSmokePipe)
failed to run tool: ld.lld
```

This bug isn't about the link failure; it's about the test suite passing despite the link failure.

niaow:
I think I see what the problem is here.

dkegel-fastly:
Thanks!  Filed https://github.com/tinygo-org/tinygo/issues/2474 for the missing syscall magic.

PR #2472, opened 2022-01-02
main (tinygo test): fail tests on build error

niaow:
This fixes a bug where build errors (e.g. linker errors) would be ignored by tinygo test.

This addresses #2466 

This is a bug in #1925 which became exposed by #2412


niaow:
I am also working on some tests for our tests for after this (#2473)

dkegel-fastly (approved):
LGTM

PR #2473, opened 2022-01-02
main (test): add tests to `tinygo test`

niaow:
This adds some tests of `tinygo test`, to ensure that we handle failures and emulators consistently.


deadprogram:
This looks very good @niaow we needed this!

deadprogram (approved):

Issue #2480, opened 2022-01-03
os.IsNotExist does not work properly on Windows?


dkegel-fastly:
dev's TestStatBadDir in os_anyos_test.go currenly skips on Windows because os.IsNotExist doesn't seem to work there.

It appears that syscall_windows.go's Errno.Is() function appears to be broken for us; the following kludge lets the test pass without skipping:

```
diff a/src/os/errors.go b/src/os/errors.go
@@ -60,12 +61,34 @@ func underlyingErrorIs(err, target error) bool {
        }
        // To preserve prior behavior, only examine syscall errors.
        e, ok := err.(syscall.Errno)
-       return ok && e.Is(target)
+       if !ok {
+               return false
+       }
+       if runtime.GOOS == "windows" {
+               // syscall_windows.go's Errno.Is() function appears to be broken for us, so recapitulate it here, badly and partially.
+               switch target {
+                       //case ErrPermission:
+                       //      return e == ERROR_ACCESS_DENIED
+                       //case ErrExist:
+                       //      return e == ERROR_ALREADY_EXISTS ||
+                       //              e == ERROR_DIR_NOT_EMPTY ||
+                       //              e == ERROR_FILE_EXISTS
+                       case ErrNotExist:
+                               return e.Error() == "File not found." ||
+                                       e.Error() == "Path not found."
+                       //      return e == ERROR_FILE_NOT_FOUND ||
+                       //              e == _ERROR_BAD_NETPATH ||
+                       //              e == ERROR_PATH_NOT_FOUND
+               }
+       }
+       return e.Is(target)
 }
```

deadprogram:
@dkegel-fastly I think this was addressed in the last release?

dkegel-fastly:
Nope.  Still present, verified on dev that the test fails if I remove the exception for windows.

deadprogram:
OK thanks for checking.

Issue #2109, opened 2021-09-14
SEEK_CUR not declared by package os


typeless:
I am trying to use tinygo to build a binary for an embedded Linux device.
I got the following errors:
```
$ tinygo build -o out.tiny .
../../cap/v4/cap.go:767:27: SEEK_CUR not declared by package os
../../cap/v4/cap.go:851:46: SEEK_SET not declared by package os
```
TinyGo version:
```
$ tinygo version
tinygo version 0.20.0-dev-ca4f251 linux/amd64 (using go version go1.17 and LLVM version 11.0.0)
```
My build OS:
```
$ cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04.3 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.3 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

ardnew:
It appears a package you are using depends on `os` from the Go standard library, which obviously isn't available for baremetal TinyGo.

Can you share the output of command `tinygo list -json` from the same directory you are trying to build from?

ardnew:
Also, if you are trying to build native ELF executables for Ubuntu, see: https://tinygo.org/docs/guides/linux/ 

~~Otherwise, you need to specify a build target for tinygo, such as `tinygo build -o=out.tiny -target=arduino .`~~

typeless:
I tried with `GOARCH=arm` as mentioned in the guides, but it gave the same build error too.

This is a proprietary project, which contains some information that I am afraid unsuitable to share.

```
$ tinygo list -json
{
	"Dir": "/home/mural/src/castlestech.com/lbox/cmd/uld",
	"ImportPath": "castlestech.com/lbox/cmd/uld",
	"Name": "main",
	"Target": "/home/mural/bin/uld",
	"Root": "/home/mural/src/castlestech.com/lbox",
	"Module": {
		"Path": "castlestech.com/lbox",
		"Main": true,
		"Dir": "/home/mural/src/castlestech.com/lbox",
[...]

PR #2481, opened 2022-01-03
os: port os.Rename + tests from upstream

dkegel-fastly:
Skips a couple fancy checks for now.
Skips part of the test on Windows because of https://github.com/tinygo-org/tinygo/issues/2480

TODO: fix 2480 and unskip test
TODO: pull in rest of upstream tests, fix problems they find

Requested in https://github.com/tinygo-org/tinygo/issues/2109

### DIFF
--- a/src/syscall/syscall_libc.go
+++ b/src/syscall/syscall_libc.go
@@ -95,6 +95,16 @@ func Rmdir(path string) (err error) {
 	return
 }
 
+func Rename(from, to string) (err error) {
+	fromdata := cstring(from)
+	todata := cstring(to)
+	fail := int(libc_rename(&fromdata[0], &todata[0]))
+	if fail < 0 {
+		err = getErrno()
+	}
+	return
+}
+
 func Unlink(path string) (err error) {
 	data := cstring(path)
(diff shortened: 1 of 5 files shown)


src/syscall/syscall_libc.go
```diff
@@ -270,6 +280,10 @@ func libc_mkdir(pathname *byte, mode uint32) int32
 //export rmdir
 func libc_rmdir(pathname *byte) int32
 
+// int rename(const char *from, *to);
+//export rename
+func libc_rename(from, too *byte) int32
```
dgryski, line 285: typo for `too` or intentional?

dgryski (approved):
LGTM. `os` is really fleshing out.

src/syscall/syscall_libc.go
dkegel-fastly: Typo, but let's leave it out of superstition :-) and because I'm getting tired of clicking "retry all builds" to work around the spurious CI problems...

deadprogram:
Thank you @dkegel-fastly and @dgryski now merging.

PR #2482, opened 2022-01-04
arm: Explicitly disable unwind tables

### LINKED ISSUE #2477: undefined symbol: __aeabi_unwind_cpp_pr0 for ARM target
With LLVM 12 support landed, I'm trying to update the Fedora packages to be usable. I'm running in to some trouble running the arm smoke test:
```
$ make smoketest STM32=0 XTENSA=0
...
GOOS=linux GOARCH=arm /builddir/build/BUILDROOT/tinygo-0.21.0-2.fc34.aarch64/usr/bin/tinygo build -size short -o test.elf       ./testdata/cgo
ld.lld: error: undefined symbol: __aeabi_unwind_cpp_pr0
>>> referenced by divsi3.c
>>>               divsi3.c.o:(.ARM.exidx.text.__divsi3+0x0) in archive /builddir/build/BUILD/tinygo-0.21.0/tinygo.ISH612/tinygo/compiler-rt-armv6-unknown-linux-gnueabihf-generic/lib.a
>>> referenced by udivsi3.c
>>>               udivsi3.c.o:(.ARM.exidx.text.__udivsi3+0x0) in archive /builddir/build/BUILD/tinygo-0.21.0/tinygo.ISH612/tinygo/compiler-rt-armv6-unknown-linux-gnueabihf-generic/lib.a
>>> referenced by udivmodsi4.c
>>>               udivmodsi4.c.o:(.ARM.exidx.text.__udivmodsi4+0x0) in archive /builddir/build/BUILD/tinygo-0.21.0/tinygo.ISH612/tinygo/compiler-rt-armv6-unknown-linux-gnueabihf-generic/lib.a
>>> referenced 64 more times
error: failed to link /tmp/tinygo570417244/main: exit status 1
```

Formerly, I only ran native Linux tests (i.e., `go test` ones) on the matching platform (e.g., amd64 on x86_64), because Fedora did not have cross-compiled libc available even though there was the cross-compiler. Now that tinygo uses its own copy of musl, running tests for all platforms might be feasible again. However, the smoke test shows that arm is still not quite working. I see that #2155 is supposed to fix something similar, but it's in 0.21.0 already.

QuLogic:
Some clang builds (e.g., Fedora's) enable unwind tables by default. As tinygo does not need nor support them, that leads to undefined symbols when linking.

Fixes #2477

### DIFF
--- a/builder/library.go
+++ b/builder/library.go
@@ -149,7 +149,7 @@ func (l *Library) load(config *compileopts.Config, tmpdir string) (job *compileJ
 		}
 	}
 	if strings.HasPrefix(target, "arm") || strings.HasPrefix(target, "thumb") {
-		args = append(args, "-fshort-enums", "-fomit-frame-pointer", "-mfloat-abi=soft")
+		args = append(args, "-fshort-enums", "-fomit-frame-pointer", "-mfloat-abi=soft", "-fno-unwind-tables")
 	}
 	if strings.HasPrefix(target, "riscv32-") {
 		args = append(args, "-march=rv32imac", "-mabi=ilp32", "-fforce-enable-int128")
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -265,6 +265,7 @@ func defaultTarget(goos, goarch, triple string) (*TargetSpec, error) {
 		spec.Features = "+cx8,+fxsr,+mmx,+sse,+sse2,+x87"
 	case "arm":
 		spec.CPU = "generic"
+		spec.CFlags = append(spec.CFlags, "-fno-unwind-tables")
 		switch strings.Split(triple, "-")[0] {
 		case "armv5":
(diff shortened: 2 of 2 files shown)


deadprogram:
@QuLogic just wanted to point out https://github.com/tinygo-org/tinygo/pull/2331 which I think will need unwind tables if using C++ exceptions if I am not mistaken.

QuLogic:
That's probably okay. This PR currently only minimally affects arm builds where that option would have been disabled by default on Debian, and that PR would have had to explicitly enable it to use (but it doesn't, so I'll assume that arm is not yet a supported platform or it doesn't need unwind tables.)

QuLogic:
I don't know why LLVM needs to rebuild, and times out, for this PR but not others.

deadprogram:
@QuLogic you should rebase against `dev` as the macOS build is no longer being performed by CircleCI.

QuLogic:
Oops, thought I did, but it must have been before pulling.

deadprogram:
Thanks @QuLogic now merging.

aykevl (commented):
It's already merged, but just wanted to say this looks good to me.

PR #2476, opened 2022-01-02
os: implement File.Seek, add smoke test

dkegel-fastly:
This was simpler than I thought...

PR #2485, opened 2022-01-04
os: implement File.Seek, add smoke test

dkegel-fastly:
After staring hard at upstream, I decided to use the same pattern they use, rather than adding a FileHandle method for Seek.


dkegel-fastly:
Ayke should compare this with #2476, come to think of it, before this is merged.  Some thought is required.

dkegel-fastly:
As long as FileHandle is there to allow mounting, operations on open files need to go through it... so closing in favor of #2476.  Sorry for the waffling.

PR #2487, opened 2022-01-05
Partially support BenchmarkXXX functions with the -bench and -test.bench flags

dkegel-fastly:
Had this sitting around half-finished for a long time, just today I needed benchmarks, and suddenly the last little bit got much easier to write :-)

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -140,6 +140,8 @@ jobs:
           submodules: true
       - name: Install apt dependencies
         run: |
+          echo "Show cpuinfo; sometimes useful when troubleshooting"
+          cat /proc/cpuinfo
           sudo apt-get update
           sudo apt-get install --no-install-recommends \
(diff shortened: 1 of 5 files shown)


dkegel-fastly:
Will go green after https://github.com/tinygo-org/tinygo/pull/2488 lands or the images are updated.

( Although it might take several retries... there seem to be multiple flaky CI failure modes, see e.g. the three failed runs before https://github.com/tinygo-org/tinygo/runs/4722737811 )

dkegel-fastly:
I think this is ready.   I hope to add support for sub-benchmarks soon, but even without them it seems useful.
The three commits are best reviewed individually.

Let's land it with all three commits intact, I am looking forward to seeing what /proc/cpuinfo says on a bad ci run.

deadprogram:
OK this should be very interesting. Thanks for working on this, been wanting some benchmarks myself for a while.

Now merging.

PR #2498, opened 2022-01-08
builder: update configuration to support Arch Linux LLVM 13 installation

niaow:
This PR makes two changes which make TinyGo work with the default installed LLVM/clang on Arch:
1. Adds `-fno-stack-protector` to musl - it is on by default on Arch, which causes TLS to be used before initialization
2. Updates the clang headers search path to handle the differences in installation paths

### DIFF
--- a/builder/env.go
+++ b/builder/env.go
@@ -84,6 +84,20 @@ func getClangHeaderPath(TINYGOROOT string) string {
 		}
 	}
 
+	// On Arch Linux, the clang executable is stored in /usr/bin rather than being symlinked from there.
+	// Search directly in /usr/lib for clang.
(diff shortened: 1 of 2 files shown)


deadprogram:
The CI fail should be fixed by a rebase against `llvm13` branch, which now has been rebased against `dev`.

deadprogram:
I already changed the base branch for this PR to `dev` so now it should just need a rebase to pass CI.

niaow:
Ugh, macos just seems to be timing out repeatedly.

deadprogram:
I think I will merge anyway, and see if the `dev` branch will have better luck on the macOS build.

Thanks @niaow

PR #2499, opened 2022-01-08
os: implement readdir for darwin and linux

dkegel-fastly:
Other operating systems are left as an exercise for the reader :-)

### DIFF
--- a/src/os/dir_darwin.go
+++ b/src/os/dir_darwin.go
@@ -1,0 +1,156 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package os
+
+import (
+	"io"
+	"runtime"
+	"syscall"
+	"unsafe"
+)
+
+// Auxiliary information if the File describes a directory
+type dirInfo struct {
+	dir uintptr // Pointer to DIR structure from dirent.h
+}
+
+func (d *dirInfo) close() {
+	if d.dir == 0 {
+		return
+	}
+	closedir(d.dir)
+	d.dir = 0
+}
+
+func (f *File) readdir(n int, mode readdirMode) (names []string, dirents []DirEntry, infos []FileInfo, err error) {
+	if f.dirinfo == nil {
+		dir, call, errno := darwinOpenDir(syscallFd(f.handle.(unixFileHandle)))
+		if errno != nil {
+			return nil, nil, nil, &PathError{Op: call, Path: f.name, Err: errno}
+		}
+		f.dirinfo = &dirInfo{
+			dir: dir,
+		}
+	}
+	d := f.dirinfo
+
+	size := n
+	if size <= 0 {
+		size = 100
+		n = -1
+	}
+
+	var dirent syscall.Dirent
+	var entptr *syscall.Dirent
+	for len(names)+len(dirents)+len(infos) < size || n == -1 {
+		if errno := readdir_r(d.dir, &dirent, &entptr); errno != 0 {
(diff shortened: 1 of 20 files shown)


dkegel-fastly:
The i386 problem this ran into in CI may be related to https://github.com/tinygo-org/tinygo/issues/1906

It can be reproduced with the commands

```
make
cd tests/testing/pass
GOARCH=386 ../../../build/tinygo test
```
in this branch, but not in dev.

dgryski (approved):
LGTM.  Large patch, but almost all of it seems to be from upstream Go. And it has tests, which pass.  👍

deadprogram:
Something not working quite right here: https://github.com/tinygo-org/tinygo/runs/4845686957?check_suite_focus=true#step:17:420

```
tinygo:ld.lld: error: undefined symbol: syscall.seek
```

dkegel-fastly:
Yes, see https://github.com/tinygo-org/tinygo/pull/2499#issuecomment-1013768071

The real fix may involves Ayke's fix for https://github.com/tinygo-org/tinygo/issues/1906

Until then, maybe I could figure out some clever way to exclude the architectures with the problem...?

dkegel-fastly:
Needs a tiny bit more cleanup.  Also can't land until the go assembly feature lands.  So marking draft again for now.

dkegel-fastly:
Rebased atop https://github.com/tinygo-org/tinygo/pull/2571 for the seek fix

dkegel-fastly:
https://github.com/tinygo-org/tinygo/pull/2571 should land first, but otherwise this is ready for review.

src/os/dir_darwin.go
```diff
@@ -140,6 +140,12 @@ func darwinOpenDir(fd syscallFd) (uintptr, string, error) {
 			break
 		}
 	}
+	// Without this test, darwinOpenDir would happily open non-directories, causing trouble in at least Glob().
```
dkegel-fastly: @aykevl syscall.Fdopendir needs reviewing, I think.

dkegel-fastly: sorted.

dankegel:
I think this is ready, now that I've tested it with io/fs and fixed the problem this found on Darwin.

dkegel-fastly:
Oh, ok, maybe it's too late for 0.22.

dgryski (approved):
LGTM

deadprogram:

src/os/file_other.go
```diff
@@ -22,6 +22,19 @@ const isOS = false
...
+// file is the real representation of *File.
+// The extra level of indirection ensures that no clients of os
+// can overwrite this data, which could cause the finalizer
+// to close the wrong file descriptor.
+type file struct {
+	handle FileHandle
+	name   string
+}
+
+func NewFile(fd FileHandle, name string) *File {
```
QuLogic, line 34: This differs from Go in taking an interface for `fd` instead of a `uintptr`, which seems to break if passing, e.g., file descriptor 3 as done in [Go 1.18's fuzzing code](https://github.com/golang/go/blob/d588f487703e773ba4a2f0a04f2d4141610bff6b/src/internal/fuzz/sys_posix.go#L61) (cf. #2515)

dkegel-fastly: IIRC that was taken straight from upstream?  Maybe upstream has changed, or maybe I'm misremembering.

Either way, please file a bug and we can sort it.

QuLogic: Well, it's been that way for 10 years, it appears. I'll just open a PR as I've already made the changes locally.

PR #2501, opened 2022-01-09
Add JSON output to info command

cweagans:
I'm working on a neovim plugin and it would be very helpful to have the output of `tinygo info` in a machine readable format. My intent is to make it so that I can open a list of all targets, select one, and auto-configure the neovim environment variables to use the appropriate `GOROOT` and all that.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1514,21 +1514,23 @@ func main() {
 		}
 		if *flagJSON {
 			json, _ := json.MarshalIndent(struct {
-				GOROOT    string   `json:"goroot"`
-				GOOS      string   `json:"goos"`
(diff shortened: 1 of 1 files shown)


dgryski (approved):
LGTM.  Why no LLVM Triple in the JSON?

cweagans:
Ah, that would be because I overlooked it :) Pushed an update to include it.

dgryski (approved):
LGTM

deadprogram:
@cweagans thank you for this contribution. I am going to squash into a single commit and then merge.

@dgryski thanks for providing feedback and review.

aykevl (commented):
Looks reasonable :+1:

PR #2502, opened 2022-01-09
stm32: add blues wireless swan

kenbell:
I've verified:
* Timer setup (blink frequency)
* I2C using BME280
* Serial IO (UART)

### DIFF
--- a/src/machine/machine_stm32l4.go
+++ b/src/machine/machine_stm32l4.go
@@ -117,6 +117,25 @@ const (
 	PE15 = portE + 15
 )
 
+// IRQs are defined here as they vary in the SVDs, but do have consistent mapping
+// to Timer Interrupts.
+const (
+	irq_TIM1_BRK_TIM15     = 24
+	irq_TIM1_UP_TIM16      = 25
+	irq_TIM1_TRG_COM_TIM17 = 26
+	irq_TIM1_CC            = 27
+	irq_TIM2               = 28
+	irq_TIM3               = 29
+	irq_TIM4               = 30
+	irq_TIM5               = 50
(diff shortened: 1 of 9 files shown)


deadprogram:
Check this out @rayozzie :smile_cat:

kenbell:
Updated to use stm32 dfu mode by default...
```tinygo flash -target swan <path>```  << use DFU
```tinygo flash -target swan -programmer openocd <path>``` << use stlink

File: src/machine/machine_stm32l4.go [hunk 1]
```diff
@@ -117,6 +117,25 @@ const (
 	PE15 = portE + 15
 )
 
+// IRQs are defined here as they vary in the SVDs, but do have consistent mapping
+// to Timer Interrupts.
+const (
+	irq_TIM1_BRK_TIM15     = 24
```
deadprogram, line 123: Perhaps you want to put these into `device/stm32` similar to how https://github.com/tinygo-org/tinygo/blob/release/src/device/sam/atsamd51x-bitfields.go is?

kenbell: How strongly do you feel about it?

I kinda feel icky about putting hand-crafted code in there.  Long-term better to work with rust team on getting the IRQ constants aligned.  I've made these non-exported so they don't become an implicit API for apps.

deadprogram: I like your long-term solution much better!

deadprogram:
Thanks for another great contribution @kenbell now merging.

rayozzie:
Indeed, my deep appreciation!

PR #2505, opened 2022-01-09
testing: implement Cleanup(), with tests

dkegel-fastly:
t.Cleanup() will be needed when we implement t.TempDir().

Land it now with tests to reduce the size of the future pull request for t.TempDir().

Also update testdata/testing.go to run properly with go 1.16 (it crashed without this).
This was useful while I was debugging the tRunner() change.

### DIFF
--- a/src/testing/sub_test.go
+++ b/src/testing/sub_test.go
@@ -1,0 +1,80 @@
+// Copyright 2016 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package testing
+
+import (
+	"reflect"
+)
+
+func TestCleanup(t *T) {
+	var cleanups []int
+	t.Run("test", func(t *T) {
+		t.Cleanup(func() { cleanups = append(cleanups, 1) })
+		t.Cleanup(func() { cleanups = append(cleanups, 2) })
+	})
+	if got, want := cleanups, []int{2, 1}; !reflect.DeepEqual(got, want) {
+		t.Errorf("unexpected cleanup record; got %v want %v", got, want)
+	}
+}
+
+func TestRunCleanup(t *T) {
+	outerCleanup := 0
+	innerCleanup := 0
+	t.Run("test", func(t *T) {
+		t.Cleanup(func() { outerCleanup++ })
+		t.Run("x", func(t *T) {
+			t.Cleanup(func() { innerCleanup++ })
+		})
+	})
+	if innerCleanup != 1 {
+		t.Errorf("unexpected inner cleanup count; got %d want 1", innerCleanup)
+	}
+	if outerCleanup != 1 {
+		t.Errorf("unexpected outer cleanup count; got %d want 0", outerCleanup)
+	}
+}
+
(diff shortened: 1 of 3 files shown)


src/testing/sub_test.go
```diff
@@ -0,0 +1,101 @@
...
+	t.Run("test", func(t *T) {
+		t.Cleanup(func() { cleanups = append(cleanups, 1) })
+		t.Cleanup(func() { cleanups = append(cleanups, 2) })
+	})
+	if got, want := cleanups, []int{2, 1}; !reflect.DeepEqual(got, want) {
+		t.Errorf("unexpected cleanup record; got %v want %v", got, want)
+	}
+}
+
+func TestConcurrentCleanup(t *T) {
```
dgryski: This is almost certainly racy, since its accesses the `t.cleanups` slice with no mutex.

src/testing/testing.go
```diff
@@ -204,35 +208,70 @@ func (c *common) Parallel() {
...
 		if flagVerbose {
-			fmt.Fprintf(&t.output, sub.indent+"--- PASS: %s\n", sub.name)
-			t.output.Write(sub.output.Bytes())
+			fmt.Fprintf(t.w, t.indent+"--- PASS: %s\n", t.name)
+			t.w.Write(t.output.Bytes())
 		}
 	}
+}
+
+// Run runs a subtest of f t called name. It waits until the subtest is finished
```
dgryski: Seems like this comment is missing a word or two?

dgryski (changes requested):
Almost there.  Looking good!

src/testing/sub_test.go
dkegel-fastly: Removed the premature test.

src/testing/testing.go
dkegel-fastly: Repaired word salad.

dgryski (approved):
LGTM

deadprogram:
Please resolve merge conflict. Thanks.

deadprogram:
Thank you @dkegel-fastly now merging.

testdata/testing.go
```diff
@@ -33,14 +35,26 @@ var benchmarks = []testing.InternalBenchmark{}
 
 var examples = []testing.InternalExample{}
 
+var errMain = errors.New("testing: unexpected use of func Main")
+
+// matchStringOnly is part of upstream, and is used below to provide a dummy deps to pass to MainStart
+// so it can be run with go (tested with go 1.16) to provide a baseline for the regression test.
+// See c56cc9b3b57276.  Unfortunately, testdeps is internal, so we can't just use &testdeps.TestDeps{}.
```
aykevl, line 42: > Unfortunately, testdeps is internal, so we can't just use &testdeps.TestDeps{}.
	
~That's not the reason: the testing package can most certainly import internal packages because they're both part of GOROOT. The reason that testdeps exists is so that the testing package can import useful packages while those packages can still import the testing package (for their own tests) without creating a circular dependency.~

aykevl: Oops, nevermind. The comment is correct. I didn't realize the code was in testdata/testing.go (instead of src/testing.go).

Issue #2409, opened 2021-12-19
interp: wait loops run forever


niaow:
When investigating some failures in interp, I came up with this test case:
```
package main

import(
        "sync/atomic"
        "time"
)

func init() {
        var x uint32
        go func() {
                atomic.StoreUint32(&x, 1)
        }()
        for atomic.LoadUint32(&x) != 1 {
                time.Sleep(time.Millisecond)
        }
}

func main() {
}
```

I think the correct behavior here would be to pass the loop through as is.

niaow:
I think I am mixing things up. I am very confused right now.

aykevl:
This code contains a branch based on a value that's only known at runtime, so it _should_ bail out and run the whole function at runtime. These lines are somewhat relevant:

https://github.com/tinygo-org/tinygo/blob/9fa667ce6341fe653f2cac571dd0a262fb390cdf/interp/interpreter.go#L100-L101

Basically, if any of the instruction operands (in this case, a `llvm.CondBr`) is not known, it attempts to run that instruction at runtime.
However, branch instructions can't easily be run at runtime, so abort the whole function:

https://github.com/tinygo-org/tinygo/blob/9fa667ce6341fe653f2cac571dd0a262fb390cdf/interp/interpreter.go#L987-L988

I originally attempted to make branch instructions work, but seeing this I think it's a good idea to always bail and run the function at runtime. We absolutely don't want loops like this to be unrolled.

niaow:
> I originally attempted to make branch instructions work, but seeing this I think it's a good idea to always bail and run the function at runtime. We absolutely don't want loops like this to be unrolled.

Yeah. 

I need to look back at this later, but I suspect that the atomic operation in this specific example is being run at interp time, assuming (incorrectly) that `x` could not have been modified by anything else.

PR #2508, opened 2022-01-10
interp: always run atomic and volatile loads/stores at runtime

niaow:
This addresses part of the issue in #2409 (but does not completely solve it).

### DIFF
--- a/interp/interpreter.go
+++ b/interp/interpreter.go
@@ -524,7 +524,7 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 				return nil, mem, r.errorAt(inst, err)
 			}
 			size := operands[1].(literalValue).value.(uint64)
-			if mem.hasExternalStore(ptr) {
+			if inst.llvmInst.IsVolatile() || inst.llvmInst.Ordering() != llvm.AtomicOrderingNotAtomic || mem.hasExternalStore(ptr) {
 				// If there could be an external store (for example, because a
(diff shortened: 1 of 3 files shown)


niaow:
I am still a bit unfamiliar with parts of interp, so I don't know if there is a better way to do this?

deadprogram:
@niaow can you please rebase this against `dev`? Thanks.

interp/interpreter.go
```diff
@@ -928,12 +928,18 @@ func (r *runner) runAtRuntime(fn *function, inst instruction, locals []value, me
 		if inst.llvmInst.IsVolatile() {
 			result.SetVolatile(true)
 		}
+		if ordering := inst.llvmInst.Ordering(); ordering != llvm.AtomicOrderingNotAtomic {
+			result.SetOrdering(ordering)
```
aykevl, line 932: I think the volatile flag should also be set here (and similarly below for the store instruction).

aykevl (commented):
This actually looks reasonable to me. The idea here is that any synchronisation barrier would result in an interp bail out, which seems correct to me (and volatile/atomic operations are some sort of synchronisation barriers).

interp/interpreter.go
niaow: The volatile flag is already set for volatile loads/stores in the previous if statement. You can see this working in the changed part of `interp/testdata/revert.out.ll`.

deadprogram:
Thanks @niaow now merging.

PR #2510, opened 2022-01-11
interp: show breadcrumbs to help developer find slow init functions

dkegel-fastly:
Also raise timeout to 90 seconds.

The output seems good enough to find the piggy init function that had been bothering me.

### DIFF
--- a/interp/interpreter.go
+++ b/interp/interpreter.go
@@ -17,10 +17,7 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 	locals := make([]value, len(fn.locals))
 	r.callsExecuted++
 
-	if time.Since(r.start) > time.Minute {
-		// Running for more than a minute. This should never happen.
-		return nil, mem, r.errorAt(fn.blocks[0].instructions[0], fmt.Errorf("interp: running for more than a minute, timing out (executed calls: %d)", r.callsExecuted))
-	}
+	t0 := time.Since(r.start)
 
 	// Parameters are considered a kind of local values.
 	for i, param := range params {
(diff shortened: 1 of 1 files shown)


dgryski (approved):
Seems useful until we can fix the larger interp issues.

dkegel-fastly:
With this, "make tinygo-test" outputs 
```
interp: slow: startms 484 endms 4325 func html.populateMaps
interp: slow: startms 368 endms 4328 func (*sync.Once).Do
interp: slow: startms 367 endms 4331 func html.UnescapeString
interp: slow: startms 363 endms 4334 func html.init
interp: slow: startms 114 endms 24186 func compress/flate.largeDataChunk
interp: slow: startms 76 endms 24294 func compress/flate.init
```

which is a bit embarassing.
Maybe I should raise the threshold for that print to five seconds.

Is there anything we can do to speed up https://github.com/golang/go/blob/1abe9c1c73739786bb927342c4072e229affea8f/src/compress/flate/deflate_test.go#L78 ?

dgryski:
Another similar init function is https://github.com/golang/go/blob/master/src/bytes/bytes_test.go#L1693

deadprogram:

interp/interpreter.go
```diff
@@ -106,6 +103,17 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 		}
 		switch inst.opcode {
 		case llvm.Ret:
+			t1 := time.Since(r.start)
+			if t1-t0 > time.Second {
+				// Provide some breadcrumbs for user trying to find their slow init functions.
+				fmt.Fprintln(os.Stderr, "interp: slow: startms", int(t0.Milliseconds()), "endms", int(t1.Milliseconds()), "func", fn.name)
```
aykevl, line 109: Hmm, IMHO this is a bit ugly. Perhaps it would have been better to store this information in the error object so it can be printed in `printCompilerError` in the main package?

dkegel-fastly: But it's not an error.  It's a diagnostic aid that gives the user early warning that they have a slow init, and where to look for it.

We could hide it behind a verbose flag, though.

aykevl: Yeah, I feel strongly that we shouldn't do any logging of this kind without an opt-in (via a flag for example). Normally we don't print anything for successful commands unless explicitly asked to (such as with `-size=short`), following Unix tradition. And I think that's a good tradition.

PR #2511, opened 2022-01-11
Run go mod tidy

QuLogic:
Mainly because the two things marked 'indirect' are _not_.


deadprogram:
Thank you @QuLogic now merging.

Issue #2406, opened 2021-12-18
testing flags (testVerbose/testShort/testRunRegex) should move to config.TestConfig


dgryski:
I didn't see this struct was there when I added them.  The struct even has a TODO to add these flags there already.

aykevl:
To be honest, I'd say that we don't need a separate struct for this. Just put everything directly in `compileopts.Config`. But that's just my personal opinion (I'm not going to bikeshed about this).

galamdring:
Should this be closed with the related PR merged?

dgryski:
The related PR is just me reminding myself in a comment that I still need to do this.

galamdring:
Understood. Just looking for where I can start contributing.

aykevl:
Looks like this is fixed in the dev branch.

PR #2514, opened 2022-01-11
testing: benchmark: add -benchtime, spiff up -bench

dkegel-fastly:
I needed to run one particular subbenchmark, and I couldn't stand the thought of rewriting the benchmark, so I grabbed that feature from upstream and plopped it down into tinygo.  (See "Running specific tests or benchmarks" in https://go.dev/blog/subtests .)

I didn't need it for tests just yet, so I will leave adding that behavior to -run for a later pull request.

-benchtime came along for the ride, as it was in the same area, and is quite useful.

### DIFF
--- a/main.go
+++ b/main.go
@@ -179,7 +179,7 @@ func Build(pkgName, outpath string, options *compileopts.Options) error {
 
 // Test runs the tests in the given package. Returns whether the test passed and
 // possibly an error if the test failed to run.
-func Test(pkgName string, stdout, stderr io.Writer, options *compileopts.Options, testCompileOnly, testVerbose, testShort bool, testRunRegexp string, testBenchRegexp string, outpath string) (bool, error) {
+func Test(pkgName string, stdout, stderr io.Writer, options *compileopts.Options, testCompileOnly, testVerbose, testShort bool, testRunRegexp string, testBenchRegexp string, testBenchTime string, outpath string) (bool, error) {
 	options.TestConfig.CompileTestBinary = true
 	config, err := builder.NewConfig(options)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


dankegel:
Rebased and split into two commits (one for -bench, one for -benchtime) for easier review.

main.go
```diff
@@ -179,7 +179,7 @@ func Build(pkgName, outpath string, options *compileopts.Options) error {
 
 // Test runs the tests in the given package. Returns whether the test passed and
 // possibly an error if the test failed to run.
-func Test(pkgName string, stdout, stderr io.Writer, options *compileopts.Options, testCompileOnly, testVerbose, testShort bool, testRunRegexp string, testBenchRegexp string, outpath string) (bool, error) {
+func Test(pkgName string, stdout, stderr io.Writer, options *compileopts.Options, testCompileOnly, testVerbose, testShort bool, testRunRegexp string, testBenchRegexp string, testBenchTime string, outpath string) (bool, error) {
```
dgryski, line 182: I really need to do https://github.com/tinygo-org/tinygo/issues/2406

dgryski (approved):
LGTM.  Love this more usable test and benchmarking infrastructure.

src/testing/match.go
```diff
@@ -5,9 +5,131 @@
 package testing
```
deadprogram, line 5: I think this is missing copyright here?

dkegel-fastly: It's the file verbatim from upstream, and the first four lines are
```
// Copyright 2015 The Go Authors. All rights reserved.
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.

```
so I think it's set on copyright.

deadprogram: I blame github web UI. :)

PR #2518, opened 2022-01-13
main (test): integrate test corpus runner

niaow:
This allows us to test a large corpus of external packages against the compiler.

This is based off of https://github.com/dgryski/tinygo-test-corpus

There are still a few things I need to split out before this is ready

PR #2520, opened 2022-01-14
compiler: predeclare runtime.trackPointer

niaow:
When a package only uses runtime.trackPointer to create interface packs, the compiler fails to find runtime.trackPointer. This change predeclares it alongside runtime.alloc and updates the tests to use runtime.trackPointer when the test's target uses it.

I ran into this while working on #2518 (which currently includes a hacky version of this fix).


dgryski (approved):
Not an expert on LLVM, but this LGTM.

aykevl:
Looks good to me.

PR #2522, opened 2022-01-14
testing: show duration

dkegel-fastly:
Also rearrange a couple functions to make diffing with upstream a little easier; the two commits are best reviewed one at a time.

Doesn't include Cleanup() in the duration yet, but this is good enough for the moment.

### DIFF
--- a/testdata/testing.txt
+++ b/testdata/testing.txt
@@ -1,10 +1,10 @@
---- FAIL: TestBar
+--- FAIL: TestBar (0.00s)
     log Bar
     log g
         h
         i
         
-    --- FAIL: TestBar/Bar2
+    --- FAIL: TestBar/Bar2 (0.00s)
         log Bar2
             a
             b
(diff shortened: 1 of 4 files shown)


testdata/testing.txt
```diff
@@ -1,10 +1,10 @@
---- FAIL: TestBar
+--- FAIL: TestBar (0.00s)
```
niaow, line 1: This seems a bit concerning here. Do we somehow disable the timer, or does this rely on the tests running quickly?

dkegel-fastly: Good catch. 

Fixed.  main_test.go now rewrites the received bytes.

dkegel-fastly:
Best merged without squashing (I think).

Accordingly, I force-pushed the fix to Nia's issue to keep things tidy... and also included this fix (failed tests shouldn't be printed as skipped):
```
 func (t *T) report() {
        dstr := fmtDuration(t.duration)
        format := t.indent + "--- %s: %s (%s)\n"
-       if t.Failed() && !t.Skipped() {
+       if t.Failed() {
```

dgryski (approved):
LGTM

Issue #2492, opened 2022-01-06
"LLVM ERROR: Cannot select: t3: i16 = JumpTable" in CI on AVR target


dkegel-fastly:
Got an ugly llvm crash in CI just now: https://github.com/tinygo-org/tinygo/runs/4722070252

```
=== CONT  TestBuild/AVR/goroutines.go
LLVM ERROR: Cannot select: t3: i16 = JumpTable<0>
In function: machine.AdjustMonotonicTimer
SIGABRT: abort
...
FAIL	github.com/tinygo-org/tinygo	62.689s
make: *** [Makefile:196: test] Error 1
```

Happened again in https://github.com/tinygo-org/tinygo/runs/4722481707
```
=== CONT  TestBuild/AVR/slice.go
LLVM ERROR: Cannot select: t3: i16 = JumpTable<0>
In function: runtime.alloc
SIGABRT: abort
```

Seems to happen at about 2m10s into Test TinyGo; successful runs seem to take longer to get to the same point.

Here’s a good run (a retry of the previous one): https://github.com/tinygo-org/tinygo/runs/4722602927

No idea why it doesn't happen every time.  Different software or hardware on different github ci nodes?

dkegel-fastly:
AVR may have some dirty laundry: https://github.com/llvm/llvm-project/issues/31494#issuecomment-980975376

dkegel-fastly:
Happened again in https://github.com/tinygo-org/tinygo/runs/4729218984, again at 2 minutes 16 seconds, line 413.
```
=== CONT  TestBuild/AVR/slice.go
LLVM ERROR: Cannot select: t3: i16 = JumpTable<0>
In function: runtime.alloc
SIGABRT: abort
```

dkegel-fastly:
FWIW, I snuck a "cat /proc/cpuinfo" into the top of the apt-get install step in https://github.com/tinygo-org/tinygo/pull/2487, the pull request that fails half the time with this problem.

On the first run, it said the cpu was a 2 core Xeon 8370C (so a small slice of a 32 core chip), and it did not hit this bug.

aykevl:
> AVR may have some dirty laundry: [llvm/llvm-project#31494 (comment)](https://github.com/llvm/llvm-project/issues/31494#issuecomment-980975376)

Not anymore :) see https://reviews.llvm.org/D107853#3224989

[...]

PR #2523, opened 2022-01-14
main (test): skip AVR tests

niaow:
The AVR tests fail inconsistently now due to non-deterministic backend bugs.
This disables them until this can be understood and fixed.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -133,8 +133,8 @@ func TestBuild(t *testing.T) {
 	t.Run("AVR", func(t *testing.T) {
 		// LLVM backend crash:
 		// LIBCLANG FATAL ERROR: Cannot select: t3: i16 = JumpTable<0>
-		// This bug is non-deterministic, and only happens when run concurrently with non-AVR tests.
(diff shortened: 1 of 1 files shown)


niaow:
I don't really want to do this, but right now the tests fail so often that they have become a significant problem.

deadprogram:
Which tests are failing here? Seems like CI has had no errors in this area lately?

niaow:
>  Seems like CI has had no errors in this area lately?

We have had quite a few, we have just been restarting the jobs. For example:
https://github.com/tinygo-org/tinygo/runs/4820340443?check_suite_focus=true

> Which tests are failing here?

All of them can fail, it is just random which one happens to fail.

niaow:
See #2492

dkegel-fastly:
for the love of $deity please merge this.  My mouse hand is getting sore from clicking "rerun all jobs".

deadprogram:
:crying_cat_face: merging. Hopefully we can figure out the root cause soon.

PR #2524, opened 2022-01-14
os (wasi): fix opening files in read-only mode

niaow:
On wasi, O_RDWR is a bitwise or of read and write mode. As a result, the bit test result was incorrect, and rewrote it to read-write mode. This corrects the bit test to work on all targets.

### DIFF
--- a/src/os/file_anyos.go
+++ b/src/os/file_anyos.go
@@ -1,3 +1,4 @@
+//go:build !baremetal && !js
 // +build !baremetal,!js
 
 // Portions copyright 2009 The Go Authors. All rights reserved.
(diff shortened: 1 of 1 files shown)


src/os/file_anyos.go
```diff
@@ -95,7 +96,7 @@ func (fs unixFilesystem) OpenFile(path string, flag int, perm FileMode) (FileHan
 	if flag&O_WRONLY != 0 {
 		syscallFlag |= syscall.O_WRONLY
 	}
-	if flag&O_RDWR != 0 {
+	if flag&O_RDWR == O_RDWR {
```
dgryski: I would be happier if we fixed the rest of the tests to be against the flag value rather than `!= 0`.

niaow:
I have removed the flag conversion, as it is not necessary.

dgryski (approved):
LGTM

dkegel-fastly (commented):
LGTM

PR #2526, opened 2022-01-15
Makefile: Move slow tests to end of tinygo-test. .github: test-linux-build: run tinygo-test-wasi-fast

dkegel-fastly:
With the Makefile change, impatient developers now see test results quicker, and can skip the slowest tests if they like.

With the ci change, all but the slowest stdlib packages get tested on wasi.


dgryski (approved):
LGTM

deadprogram:
Good idea @dkegel-fastly thanks for this. Now merging.

PR #2527, opened 2022-01-15
ci (Windows/Mac): use binaryen from scoop/brew

niaow:
The scoop and brew package managers now bundle up-to-date copies of binaryen.
As a result, there is no longer a strong need for us to build and package our own copy.

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -16,10 +16,10 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: '1.17'
-      - name: Install QEMU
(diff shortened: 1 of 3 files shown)


deadprogram:
Looks good. We need to make sure https://github.com/ScoopInstaller/Main/blob/master/bucket/tinygo.json is updated with this dependency and also https://github.com/tinygo-org/tinygo-site/blob/dev/content/getting-started/install/windows.md

deadprogram:
cc @thushan

deadprogram:
Thanks @niaow now merging.

thushan:
Will get onto that @deadprogram.

PR #2529, opened 2022-01-16
os: implement os.Symlink and os.Readlink

dkegel-fastly:
Needed by path/filepath, among other tests.


dgryski (approved):
LGTM

deadprogram:
:chains: thanks @dkegel-fastly and @dgryski now merging.

Issue #2533, opened 2022-01-16
nRF52840 + SoftDevice: I2C unreliable


ysoldak:
I2C hangs on Tx (writeByte) when trying to use it on a board (Nano 33 BLE) flashed with SoftDevice (s140v7).

Target I'm using is "nano-33-ble-s140v7"
It hangs not immediately, but after some time, 1-5mins.

**No I2C issues observed w/o SoftDevice. (target: "nano-33-ble")**

So what seem to happening is our writeByte waits for confirmation of byte sent and that confirmation (an event) never actually comes.

---

Observed

Both `EVENTS_STOPPED` and `EVENTS_SUSPENDED` registers are set (not zero) in the loop waiting for `EVENTS_TXDSENT` in `writeByte()` when running with SoftDevice; while both are always zeros while running w/o SoftDevice.
Surprisingly, it still works in SoftDevice case with both registers non-zero. And after I2C hangs, it's only `EVENTS_SUSPENDED` that's still non-zero, and `EVENTS_STOPPED` is zero then.

https://github.com/tinygo-org/tinygo/blob/dev/src/machine/machine_nrf.go#L313-L323

---
Reproducer

https://github.com/ysoldak/HeadTracker/tree/i2c-hangs

Have Arduino Nano 33 BLE [Sense], flash SoftDevice on it and the application code from repository above.
There is Makefile to help you with.
```
make jlink-softdevice
make jlink-flash
OR
make jlink-debug
```

ysoldak:
From docs (can be something?):

>Shared resources
The TWI shares registers and other resources with other peripherals that have the same ID as the TWI.
>Therefore, you must disable all peripherals that have the same ID as the TWI before the TWI can be configured and used. Disabling a peripheral that has the same ID as the TWI will not reset any of the registers that are shared with the TWI. It is therefore important to configure all relevant TWI registers explicitly to secure that it operates correctly.
>The Instantiation table in Instantiation shows which peripherals have the same ID as the TWI.

https://infocenter.nordicsemi.com/index.jsp?topic=%2Fps_nrf52840%2Fmemory.html&anchor=topic

ysoldak:
Issue narrowed down to advertisement part of Bluetooth stack.
So SoftDevice per se is not a problem, but advertisement part of it conflicts with TWI (I2C).

ysoldak:
After some more debugging I'm able to catch the moment of I2C bus error and ERRORSRC register points at OVERRUN error.
[...]

PR #2537, opened 2022-01-17
nrf: fix race in i2c

ysoldak:
Fixes: #2533 

There is a race in nRF implementation of I2C, this PR fixes it.
We want to trigger RESUME task only after we fully read the first byte.

SoftDevice has nothing to do with this per se, that's Bluetooth use that makes whole system bit slower and probability of the race condition rises.

### DIFF
--- a/src/machine/machine_nrf.go
+++ b/src/machine/machine_nrf.go
@@ -1,3 +1,4 @@
+//go:build nrf
 // +build nrf
 
 package machine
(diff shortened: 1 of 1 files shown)


ysoldak:
This is TWI read flow, we were triggering RESUME right away, before even reading up the first byte.
TWI_Reading

deadprogram:
Thank you for working on the fix and proving the clear explanation @ysoldak now merging.

Issue #2445, opened 2021-12-30
github.com/pkg/errors build fails with "type *runtime.Func has no field or method FileLine"


lz1998:
```bash
../../../../../../pkg/mod/github.com/pkg/errors@v0.9.1/stack.go:39:16: fn.FileLine undefined (type *runtime.Func has no field or method FileLine)
```

dgryski:
We can easily add these fields so this package compiles, but they will be stubbed out and not contain the true values.

dkegel-fastly:
What package were you trying to build?

lz1998:
> What package were you trying to build?

errors.New

errors.Wrapf

dkegel-fastly:
I can add stubs for those this week if nobody beats me to it...

dankegel:
Try the dev branch now...

deadprogram:
Closing since it appears to have been addressed. Please reopen if needed.

PR #2538, opened 2022-01-17
runtime: add stubs for Func.FileLine and Frame.PC

dkegel-fastly:
With this, 'tinygo test' in github.com/pkg/errors at least compiles and passes a few tests:
```
    $ git clone github.com/pkg/errors
    $ cd errors
    $ tinygo test -c
    $ for a in $(go test -list Test | grep Test); do ./errors.test -test.run $a -test.v  > $a.log 2>&1; done
    $ grep -l PASS *.log | wc -l
      19
    $ grep -l FAIL *.log | wc -l
      11
```
For https://github.com/tinygo-org/tinygo/issues/2445

### DIFF
--- a/src/runtime/stack.go
+++ b/src/runtime/stack.go
@@ -11,6 +11,10 @@ func (f *Func) Name() string {
 	return ""
 }
 
+func (f *Func) FileLine(pc uintptr) (file string, line int) {
+	return "", 0
+}
+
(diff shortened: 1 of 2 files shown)


dgryski (approved):
LGTM.

deadprogram:
Looking good @dkegel-fastly thanks now merging.

PR #2541, opened 2022-01-18
wasmtest: fix resource cleanup and add logging

niaow:
The chromedp context was not cancelled, so resources may have been leaking.
Additionally this waits for the browser to start before the timer starts, and extends the timeout to 20 seconds.
Logging from chromedp has also been enabled, which may help identify possible issues?


deadprogram:
Very logical. Good work @niaow now merging.

PR #2542, opened 2022-01-18
Makefile: reorganize tool detection

niaow:
There are now a large number of paths that need to be searched, and this started to get a little bit unwieldy. Additionally, brew paths were searched unconditionally, resulting in warnings every time the Makefile was run.
This reorganizes the detection paths into a parameterized list of search paths by version, which is appended to on Mac. This list is then expanded across all versions. The loop and filtering has been moved into the detect function.

Additionally, a helpful error message is displayed upon use of a missing tool:
```
Makefile:204: *** failed to locate llvm-blah at any of: llvm-build/bin/llvm-blah llvm-blah-13 llvm-blah-12 llvm-blah-11 llvm-blah.  Stop.
```


deadprogram:
A most excellent idea, thanks for doing this @niaow 

Now merging.

PR #2543, opened 2022-01-18
Makefile: fix tool autodetect conversion to absolute path

niaow:
After the last change, $(abspath x) no longer returned the actual absolute path.
This uses which instead.


deadprogram:
Makes sense thanks @niaow now merging.

PR #2545, opened 2022-01-19
nrf: i2c fix race in stop condition

ysoldak:
A follow-up for #2537 that appeared to be just a partial fix for #2533

This time race in stop condition is addressed. See comments in the code.

I've also took an opportunity to fix `Configure()` method by disabling the bus before configuring it (same as in SPI case, see `machine_nrf528xx.go:141`).

---

This is my understanding of how race is happening.

Happy scenario:
Shortcut triggers stop condition implicitly and we call `signalStop()` before slave responds.
Event comes back, we nullify it and roll happily further.

Unhappy scenario:
After shortcut has fired and before we call `signalStop()`, an interrupt occurs that pauses us and I2C peripheral has time to respond and set event flag that we see in `signalStop()` method. But task is still active and it fires on subsequent `Tx()` call after STARTTX, stops slave and we never receive `TXDSENT` event we expect in `writeByte()`.
This last bit I'm not entirely sure about. I did observe it hangs in `writeByte()`, but I'm not sure a task can be dormant and fire only after STARTTX or STARTRX.

### DIFF
--- a/src/machine/machine_nrf.go
+++ b/src/machine/machine_nrf.go
@@ -225,6 +225,9 @@ type I2CConfig struct {
 
 // Configure is intended to setup the I2C interface.
 func (i2c *I2C) Configure(config I2CConfig) error {
+
+	i2c.Bus.ENABLE.Set(nrf.TWI_ENABLE_ENABLE_Disabled)
+
 	// Default I2C bus speed is 100 kHz.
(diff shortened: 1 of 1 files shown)


deadprogram (approved):

deadprogram:
Thank you for working on this fix, @ysoldak now merging.

PR #2547, opened 2022-01-19
loader: respect $GOROOT when running `go list`

aykevl:
This makes it possible to select the Go version using GOROOT, and works even if the `go` binary is not in $PATH.

This is an alternative of sorts to #2429 and something that we should have been doing anyway.


niaow (approved):
LGTM. Seems to work decently for me.

PR #2548, opened 2022-01-19
Makefile: use absolute path for llvm-build/bin/* binaries

aykevl:
This un-breaks `make wasi-libc` on my system. Not sure which commit introduced this issue, possibly #2543 or #2542.


niaow:
I am unsure what is happening here? The abspath seems to do nothing on my system, and the which seems to do nothing on your system, so I guess we need both?

niaow:
What is the output of `which llvm-build/bin/clang` on your system?

niaow:
Nevermind. Not sure what is different here.

niaow (approved):

aykevl:
I use a different shell (fish), so:

```
~/src/github.com/tinygo-org/tinygo$ bash

~/src/github.com/tinygo-org/tinygo$ which llvm-build/bin/clang
llvm-build/bin/clang
```

PR #2549, opened 2022-01-19
wasm: remove heap allocator from wasi-libc

aykevl:
This would conflict with our own heap. We previously defined all those functions to make sure it's not used, but with a more recent wasi-libc version (https://github.com/WebAssembly/wasi-libc/pull/250) we can simply not compile the wasi-libc heap, which is the proper fix.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,10 +85,10 @@ commands:
       - run: go install -tags=llvm<<parameters.llvm>> .
       - restore_cache:
           keys:
-            - wasi-libc-sysroot-systemclang-v3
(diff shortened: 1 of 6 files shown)


niaow:
Can you rebase this on `dev` since the makefile has been fixed?

aykevl:
Rebased.

aykevl:
This seems to be working now, except that TinyHCI is acting up.

deadprogram:
TinyHCI restarted, all passing now.

niaow (approved):

deadprogram:
Thanks! Merging.

Issue #2090, opened 2021-09-06
interp timeout compiling large maps with runtime inits


dgryski:
```
package main

import "strconv"

var x map[string]int

func init() {
	x = map[string]int{}
	for i := 0; i < 65536; i++ {
		x[strconv.Itoa(i)] = i
	}
}

func main() {
	println(len(x))
}
```

Compilation fails with
```
/Users/dgryski/go/src/tinygo.org/tinygo/src/runtime/func.go:26:6: interp: running for more than a minute, timing out (executed calls: 182712)
  %val = alloca %runtime._string, align 8, !dbg !447

traceback:
/Users/dgryski/go/src/tinygo.org/tinygo/src/runtime/func.go:26:6:
  %val = alloca %runtime._string, align 8, !dbg !447
/Users/dgryski/go/src/tinygo.org/tinygo/src/runtime/hashmap.go:134:16:
  %69 = call i32 @runtime.getFuncPtr(i8* %keyEqual.context, i32 %keyEqual.funcptr, i8* @"reflect/types.funcid:func:{basic:unsafeptr,basic:unsafeptr,basic:uintptr}{basic:bool}", i8* undef, i8* null), !dbg !604
/Users/dgryski/go/src/tinygo.org/tinygo/src/runtime/hashmap.go:337:12:
  call void @runtime.hashmapSet(%runtime.hashmap* %m, i8* %key, i8* %value, i32 %3, i8* undef, i32 ptrtoint (%runtime._interface* @"runtime.hashmapStringEqual$withSignature" to i32), i8* undef, i8* undef), !dbg !458
main/<init>:
  call void @runtime.hashmapStringSet(%runtime.hashmap* %3, i8* %14336, i32 %14337, i8* bitcast (i32* @"main$alloca" to i8*), i8* undef, i8* null), !dbg !439
```

dgryski:
Would adding `//go:tinygo-init-at-runtime` or something be a reasonable work-around to move some globals to initialization at runtime instead of at compile-time with the `interp` pass ?

dkegel-fastly:
It doesn't feel like a very exciting prospect, somehow.

dkegel-fastly:
This patch attempts to help the developer figure out what init function is being slow, so he can rework it:
```
--- a/interp/interpreter.go
+++ b/interp/interpreter.go
@@ -19,7 +19,7 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 
[...]

PR #2555, opened 2022-01-19
interpreter.go: double timeout to let a real world project build. 

dkegel-fastly:
Waiting is a bit less frustrating now that we print breadcrumbs.

See https://github.com/tinygo-org/tinygo/issues/2090

The new value was chosen to be about twice the value needed to let a real world app build; it currently fails after printing
```
interp: slow: startms 338 endms 91958 
```


dgryski:
Another limp-along solution would be to make this configurable with a flag.

dgryski (approved):
LGTM.

deadprogram:
Not the worst hack in our toolbox, so merging.

Issue #2563, opened 2022-01-20
os: problem with Readlink on baremetal


deadprogram:
For some reason there is a new failure in the smoke tests for the TinyGo driver's repo, I think is probably in some way due to the implementation in PR https://github.com/tinygo-org/tinygo/pull/2529

The error is:

```
tinygo build -size short -o ./build/test.hex -target=wioterminal ./examples/rtl8720dn/webclient/
go: downloading golang.org/x/net v0.0.0-20210614182718-04defd469f4e
go: downloading golang.org/x/text v0.3.6
# path/filepath
/usr/local/go/src/path/filepath/symlink.go:103:19: Readlink not declared by package os
make: *** [Makefile:208: smoke-test] Error 1
```

https://app.circleci.com/pipelines/github/tinygo-org/drivers/1104/workflows/c928d9b8-975a-4915-bae9-67ce62e940df/jobs/1139?invite=true#step-105-400

dkegel-fastly:
Reproducible without drivers repo using by building path/filepath test for bare metal target, I think, e.g.
```
tinygo test -c -target=pybadge path/filepath
# path/filepath
/usr/local/go/src/path/filepath/symlink.go:103:19: Readlink not declared by package os
```

deadprogram:
Confirmed on my machine:

```
$ tinygo test -c -target=pybadge path/filepath
# path/filepath
/home/ron/.gvm/gos/go1.17.5/src/path/filepath/symlink.go:103:19: Readlink not declared by package os
FAIL
```

dkegel-fastly:
Fix with proper test coming tonight.

dkegel-fastly:
https://github.com/tinygo-org/tinygo/pull/2565 should sort this.

deadprogram:
Drivers build went green thanks to you @dkegel-fastly so now closing.

Issue #2530, opened 2022-01-16
:bug: build error on target "nintendoswitch" when import "fmt"


ca1e:
```golang
package main
import "fmt"
func main() {
	fmt.Println("Hello,world")
}
```
$tinygo build -target nintendoswitch -o dummy
```
ld.lld: error: undefined symbol: os.runtime_args
>>> referenced by scheduler_any.go:19
ld.lld: error: undefined symbol: errno
>>> referenced by syscall_libc.go:0
```

ca1e:
go version: 1.17.6
tinygo version:  0.21.0

ca1e:
tinygo: version 0.22.0-dev-751ac85, here is another errors:
```
# os
~\tinygo\src\os\types_unix.go:20:18: Stat_t not declared by package syscall
~\tinygo\src\os\stat_linux.go:46:32: Timespec not declared by package syscall
~\tinygo\src\os\file_anyos.go:185:16: S_ISUID not declared by package syscall
~\tinygo\src\os\file_anyos.go:188:16: S_ISGID not declared by package syscall
~\tinygo\src\os\file_anyos.go:191:16: S_ISVTX not declared by package syscall
~\tinygo\src\os\stat_linux.go:19:31: S_IFMT not declared by package syscall
~\tinygo\src\os\stat_linux.go:20:15: S_IFBLK not declared by package syscall
~\tinygo\src\os\stat_linux.go:22:15: S_IFCHR not declared by package syscall
~\tinygo\src\os\stat_linux.go:24:15: S_IFDIR not declared by package syscall
~\tinygo\src\os\stat_linux.go:26:15: S_IFIFO not declared by package syscall
~\tinygo\src\os\stat_linux.go:28:15: S_IFLNK not declared by package syscall
~\tinygo\src\os\stat_linux.go:30:15: S_IFREG not declared by package syscall
~\tinygo\src\os\stat_linux.go:32:15: S_IFSOCK not declared by package syscall
~\tinygo\src\os\stat_linux.go:35:25: S_ISGID not declared by package syscall
~\tinygo\src\os\stat_linux.go:38:25: S_ISUID not declared by package syscall
~\tinygo\src\os\stat_linux.go:41:25: S_ISVTX not declared by package syscall
~\tinygo\src\os\stat_linux.go:52:43: Stat_t not declared by package syscall
~\tinygo\src\os\stat_unix.go:22:37: Stat not declared by package syscall
[...]

Issue #2389, opened 2021-12-14
interp: ptrtoint integer size does not equal pointer size


niaow:
I think this error was supposed to be used internally for rollback, but it seems that it replicates on AVR:
```
package main

import "fmt"

func printOne() {
        fmt.Println("one")
}

func main() {
        f := printOne
        f()
}
```

This example is pulled from #349 which I was checking on.

```
[niaow@finch tinygo]$ build/tinygo run -target arduino /tmp/bleh.go
error: interp: ptrtoint integer size does not equal pointer size
```

dgryski:
Is this the issue that was mentioned in the patch for the conservative collector here: https://github.com/tinygo-org/tinygo/pull/1872/commits/f870ed5cbab8abeacd61466c1a739a3f16bc098e ?

niaow:
No, it is not. Although idk if the patch might impact it.

dgryski:
Is this patch then still something I should create a PR for?

```
diff --git a/src/runtime/gc_conservative.go b/src/runtime/gc_conservative.go
index cabd41427..faa4091a5 100644
--- a/src/runtime/gc_conservative.go
+++ b/src/runtime/gc_conservative.go
@@ -410,7 +410,7 @@ func markRoots(start, end uintptr) {
 		}
 	}
 
-	for addr := start; addr < end; addr += unsafe.Alignof(addr) {
+	for addr := start; addr+unsafe.Sizeof(unsafe.Pointer(nil)) <= end; addr += unsafe.Alignof(addr) {
 		root := *(*uintptr)(unsafe.Pointer(addr))
 		markRoot(addr, root)
 	}
```

niaow:
I do not know. I will need to look into that again. That should not be related to this issue.

niaow:
So the answer there is that yes that code still needs to be fixed but also my fix had a bug in it because `markRoots` already had another load-bearing bug.

niaow:
For future reference, that patch has been turned into #2396

dgryski:
This issue is also happening when trying to build https://github.com/google/open-location-code/, and I'm pretty sure that's a recent failure.

aykevl:
This issue is most likely unrelated to any runtime GC that might happen. The interp package intercepts calls to `runtime.alloc` and replaces them with creating a new global.

> I think this error was supposed to be used internally for rollback

[...]

PR #2565, opened 2022-01-20
os: add stubs for Readlink and File.Seek for baremetal targets

dkegel-fastly:
With smoke test, currently enabled on pybadge (chosen at random).

Fixes https://github.com/tinygo-org/tinygo/issues/2563

TODO:
- enable test on arduino; currently fails with "interp: ptrtoint integer size..." (https://github.com/tinygo-org/tinygo/issues/2389 )
- enable test on nintendoswitch; currently fails with many missing definitions (https://github.com/tinygo-org/tinygo/issues/2530 )

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -92,4 +92,4 @@ jobs:
           path: build/tinygo.darwin-amd64.tar.gz
       - name: Smoke tests
         shell: bash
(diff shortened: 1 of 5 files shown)


deadprogram:
Thank you for the quick action on the fix @dkegel-fastly :heart: 

Now merging.

PR #2567, opened 2022-01-21
Switch default to llvm13

fgsch:


aykevl (approved):
Looks reasonable, thanks!

PR #2571, opened 2022-01-22
os: Seek: provide stub for syscall.seek on i386 and arm.

dkegel-fastly:
Lets 'GOARCH=386 tinygo test os' compile on linux.

Replaces the patch "Kludge: work around lack of syscall.seek on 386 and arm, #1906", which tried to do the same thing but at a higher level, thus failing to catch the direct reference to syscall.Seek in the time package.

Also skip tests that depend on seek so "GOARCH=386 tinygo test os" passes again on linux.

TODO: add CI runners that test os on 386 and arm?


dkegel-fastly:
This is part of https://github.com/tinygo-org/tinygo/pull/2499 (the pull request that exposed the need for this fix), so you could just merge that and close this one if you like.

deadprogram:
Thanks for the fix @dkegel-fastly let's get it in for 0.22

Now merging.

PR #2575, opened 2022-01-24
all: release v0.22.0

aykevl:


deadprogram:
Looks ready to me! :shipit:

aykevl:
The only thing that I'd like to include is https://github.com/tinygo-org/tinygo/pull/2577.
For the interp issues, I think https://github.com/tinygo-org/tinygo/pull/2561 could be a very good solution. But it could also have unintended side effects, so maybe including it in v0.22.0 is a bit risky?

deadprogram:
We also made a lot of changes to the CI process, including migrating most things to Github Actions. Probably worth mentioning someplace.

deadprogram:
Merged #2577 build almost done.

Agree, let's hold off on #2561 until next release.

dankegel:
https://github.com/tinygo-org/tinygo/pull/2571 might be good to include, as it fixes linking problems in the time package for arm and 386.

dkegel-fastly:
Here's the link error 2571 fixes; it's pretty annoying.  People will want to be able to print Now().
https://github.com/tinygo-org/tinygo/issues/2578

dkegel-fastly:
Next showstopper: https://github.com/tinygo-org/tinygo/issues/2580 
I have a safe workaround locally, and will submit it as a pull request after testing.

aykevl:
> We also made a lot of changes to the CI process, including migrating most things to Github Actions. Probably worth mentioning someplace.

Not sure about that.
How I make the changelog is that I try to include changes that are relevant to end users. This means that not all commits are listed and some changes are merged together. For example:

* I don't include CI changes because they are mostly relevant to TinyGo developers, not to its users.
* Similarly, I usually don't include changes to stdlib tests.
* Some commits are merged. For example, if a board is added in one commit and another adds some more pin definitions, I don't include that other commit (it's still part of "adding the board").

But maybe the fact that release binaries are now mostly built on GitHub Actions could be relevant to end users?

aykevl:
I think this is now ready. Or did I miss anything important?

deadprogram:
:rocket:

aykevl:
This might also be useful for the release: https://github.com/tinygo-org/tinygo/pull/2585

deadprogram:
OK merged #2585 since it is in fact related to the release process.

PR #2576, opened 2022-01-24
esp32c3: add support for input pin

sago35:
This PR adds the basic PinInput functionality of ESP32-C3.

https://www.espressif.com/sites/default/files/documentation/esp32-c3_technical_reference_manual_en.pdf

If Pin.Get() does not exist, then GPIO does not exist as documented in tinygo-site.
https://github.com/tinygo-org/tinygo-site/blob/release/doc-gen/main.go#L365

### DIFF
--- a/src/machine/machine_esp32c3.go
+++ b/src/machine/machine_esp32c3.go
@@ -95,6 +95,13 @@ func (p Pin) Set(value bool) {
 	}
 }
 
+// Get returns the current value of a GPIO pin when configured as an input or as
+// an output.
+func (p Pin) Get() bool {
+	reg := &esp.GPIO.IN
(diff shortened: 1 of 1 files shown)


sago35:
I used m5stamp-c3 and tested it with the following code.


```go
package main

import (
	"fmt"
	"machine"
	"time"
)

const (
	button = machine.IO7
)

func main() {
	button.Configure(machine.PinConfig{Mode: machine.PinInputPullup})

	for {
		if button.Get() {
			fmt.Printf("on\r\n")
		} else {
			fmt.Printf("off\r\n")
		}

		time.Sleep(time.Millisecond * 10)
	}
}
```

deadprogram:

PR #2577, opened 2022-01-24
interp: don't log from the interp package

aykevl:
We shouldn't be logging anything when not explicitly requested via a flag or some other way.


deadprogram:
Agreed. @dkegel-fastly perhaps you can add this back in under a flag for the next release?

dankegel:
Sure

Issue #2580, opened 2022-01-24
"tinygo build" aborts on Alpine with "error: could not find build ID"


dkegel-fastly:
https://github.com/tinygo-org/tinygo/pull/2366 seems to have introduced a regression.

Discovered by building and using current dev branch inside an Alpine 3.14 container, which worked in 0.21.

tinygo/builder/buildid.go was expecting the gnu build id to be in a section named .note.gnu.build-id,
but on Alpine, it appears to be in a section named .note.go.buildid. 

The obvious change (accepting either) allows tinygo to successfully build programs on alpine, pull request to follow.

dkegel-fastly:
https://github.com/golang/go/blob/84eefdc933410907495e42aac872036403851ffa/src/cmd/internal/buildid/buildid.go#L143 suggests that go.buildid is the proper name, and gnu.build-id was working by accident?

There is also gnu.buildid floating around the internet, just to make things confusinger.

deadprogram:
Closing since this was correct in-stream for the current release. Thanks!

PR #2581, opened 2022-01-24
builder/buildid.go: accept alternate name for buildid section on Linux

dkegel-fastly:
On Ubuntu, using standard go, both go and gnu buildid sections are present.
On Alpine, the gnu buildid section is absent, which caused tinygo to abort early.

It is possible that we could hit a situation where only the gnu
buildid section is present, so accept either one just in case.

Fixes https://github.com/tinygo-org/tinygo/issues/2580

### DIFF
--- a/builder/buildid.go
+++ b/builder/buildid.go
@@ -31,7 +31,8 @@ func ReadBuildID() ([]byte, error) {
 			return nil, err
 		}
 		for _, section := range file.Sections {
-			if section.Type != elf.SHT_NOTE || section.Name != ".note.gnu.build-id" {
+			if section.Type != elf.SHT_NOTE ||
+				(section.Name != ".note.gnu.build-id" && section.Name != ".note.go.buildid") {
 				continue
 			}
 			buf := make([]byte, section.Size)


dgryski (approved):
LGTM

aykevl:
Which one comes first?
Usually I think we should prefer the GNU build ID, because it should (I think) also cover the non-Go parts.

dkegel-fastly:
gnu takes priority currently.

deadprogram:
Thank you for the quick fix @dkegel-fastly merging.

aykevl:
> gnu takes priority currently.

Not for me:

```
$ objdump -h ~/bin/tinygo | grep build
  3 .note.go.buildid   00000064 000000000020031c 
  4 .note.gnu.build-id 00000018 0000000000200380 
 31 .go.buildinfo      00000020 000000000086f070 DATA
```

...but I'm also using ld.lld instead of ld.bfd as the linker, which could affect the result.

aykevl:
See #2584 (but it's not necessary for the coming release).

dkegel-fastly:
my bad!  Answered without thinking.

PR #2584, opened 2022-01-25
builder: prefer GNU build ID over Go build ID

aykevl:
The GNU build ID covers the Go build ID, and probably some more.

### DIFF
--- a/builder/buildid.go
+++ b/builder/buildid.go
@@ -30,6 +30,7 @@ func ReadBuildID() ([]byte, error) {
 		if err != nil {
 			return nil, err
 		}
+		var gnuID, goID []byte
 		for _, section := range file.Sections {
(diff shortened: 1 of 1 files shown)


builder/buildid.go
```diff
@@ -40,7 +41,16 @@ func ReadBuildID() ([]byte, error) {
 			if uint64(n) != section.Size || err != nil {
 				return nil, fmt.Errorf("could not read build id: %w", err)
 			}
-			return buf, nil
+			if section.Name == ".note.gnu.build-id" {
+				gnuID = buf
```
dkegel-fastly, line 45: To be squeaky minimal, I'd just return immediately rather than keeping gnuID around...

dkegel-fastly (approved):
LGTM

deadprogram:
Merging, thanks @aykevl

PR #2585, opened 2022-01-25
build: fix build-library subcommand

aykevl:
This subcommand has been broken for a while, since libraries also use the CPU flag. This commit fixes this.

Previously, libraries were usable for most Cortex-M cores. But with the addition of the CPU field, I've limited it to three popular cores: the Cortex-M0 (microbit), Cortex-M0+ (atsamd21), and Cortex-M4 (atsamd21, nrf52, and many others).

In the future we might consider also building libraries for the current OS/arch so that libraries like musl are already precompiled.

---

@QuLogic this PR might be relevant to you.


deadprogram:
OK, seems like a worthwhile fix to pull in before a release. Now merging.

QuLogic:
Thanks for the heads up.

PR #2588, opened 2022-01-26
src/runtime: fix off-by-one in metadata size calculation due to rounding

dgryski:
This fixes the `archive/zip` stdlib test corpus and `dgryski/go-factor` from the corpus on `target=wasi`

### DIFF
--- a/src/runtime/gc_conservative.go
+++ b/src/runtime/gc_conservative.go
@@ -143,7 +143,7 @@ func (b gcBlock) findNext() gcBlock {
 // State returns the current block state.
 func (b gcBlock) state() blockState {
 	stateBytePtr := (*uint8)(unsafe.Pointer(uintptr(metadataStart) + uintptr(b/blocksPerStateByte)))
-	return blockState(*stateBytePtr>>((b%blocksPerStateByte)*2)) % 4
+	return blockState(*stateBytePtr>>((b%blocksPerStateByte)*stateBits)) & blockStateMask
 }
 
 // setState sets the current block to the given state, which must contain more
@@ -151,7 +151,7 @@ func (b gcBlock) state() blockState {
 // from head to mark.
 func (b gcBlock) setState(newState blockState) {
 	stateBytePtr := (*uint8)(unsafe.Pointer(uintptr(metadataStart) + uintptr(b/blocksPerStateByte)))
-	*stateBytePtr |= uint8(newState << ((b % blocksPerStateByte) * 2))
+	*stateBytePtr |= uint8(newState << ((b % blocksPerStateByte) * stateBits))
 	if gcAsserts && b.state() != newState {
 		runtimePanic("gc: setState() was not successful")
 	}
@@ -160,7 +160,7 @@ func (b gcBlock) setState(newState blockState) {
 // markFree sets the block state to free, no matter what state it was in before.
 func (b gcBlock) markFree() {
 	stateBytePtr := (*uint8)(unsafe.Pointer(uintptr(metadataStart) + uintptr(b/blocksPerStateByte)))
-	*stateBytePtr &^= uint8(blockStateMask << ((b % blocksPerStateByte) * 2))
+	*stateBytePtr &^= uint8(blockStateMask << ((b % blocksPerStateByte) * stateBits))
 	if gcAsserts && b.state() != blockStateFree {
 		runtimePanic("gc: markFree() was not successful")
 	}
@@ -174,7 +174,7 @@ func (b gcBlock) unmark() {
 	}
 	clearMask := blockStateMask ^ blockStateHead // the bits to clear from the state
 	stateBytePtr := (*uint8)(unsafe.Pointer(uintptr(metadataStart) + uintptr(b/blocksPerStateByte)))
-	*stateBytePtr &^= uint8(clearMask << ((b % blocksPerStateByte) * 2))
+	*stateBytePtr &^= uint8(clearMask << ((b % blocksPerStateByte) * stateBits))
 	if gcAsserts && b.state() != blockStateHead {
 		runtimePanic("gc: unmark() was not successful")
 	}
--- a/src/runtime/gc_conservative.go
+++ b/src/runtime/gc_conservative.go
@@ -240,7 +240,7 @@ func calculateHeapAddresses() {
 	totalSize := heapEnd - heapStart
 
 	// Allocate some memory to keep 2 bits of information about every block.
-	metadataSize := totalSize / (blocksPerStateByte * bytesPerBlock)
+	metadataSize := (totalSize + blocksPerStateByte*bytesPerBlock) / (1 + blocksPerStateByte*bytesPerBlock)
 	metadataStart = unsafe.Pointer(heapEnd - metadataSize)
 
 	// Use the rest of the available memory as heap.
--- a/src/runtime/gc_conservative.go
+++ b/src/runtime/gc_conservative.go
@@ -110,7 +110,11 @@ func (b gcBlock) pointer() unsafe.Pointer {
 
 // Return the address of the start of the allocated object.
 func (b gcBlock) address() uintptr {
-	return heapStart + uintptr(b)*bytesPerBlock
+	addr := heapStart + uintptr(b)*bytesPerBlock
+	if gcAsserts && addr > uintptr(metadataStart) {
+		runtimePanic("gc: block pointing inside metadata")
+	}
+	return addr
 }
 
 // findHead returns the head (first block) of an object, assuming the block
--- a/src/runtime/gc_conservative.go
+++ b/src/runtime/gc_conservative.go
@@ -14,7 +14,7 @@ package runtime
 // any free space, it will perform a garbage collection cycle and try again. If
 // it still cannot find any free space, it gives up.
 //
-// Every block has some metadata, which is stored at the beginning of the heap.
+// Every block has some metadata, which is stored at the end of the heap.
 // The four states are "free", "head", "tail", and "mark". During normal
 // operation, there are no marked blocks. Every allocated object starts with a
 // "head" and is followed by "tail" blocks. The reason for this distinction is
--- a/src/runtime/gc_conservative.go
+++ b/src/runtime/gc_conservative.go
@@ -134,7 +134,7 @@ func (b gcBlock) findNext() gcBlock {
 	if b.state() == blockStateHead || b.state() == blockStateMark {
 		b++
 	}
-	for b.state() == blockStateTail {
+	for b.address() < uintptr(metadataStart) && b.state() == blockStateTail {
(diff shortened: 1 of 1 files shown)


src/runtime/gc_conservative.go
```diff
@@ -240,7 +240,7 @@ func calculateHeapAddresses() {
 	totalSize := heapEnd - heapStart
 
 	// Allocate some memory to keep 2 bits of information about every block.
-	metadataSize := totalSize / (blocksPerStateByte * bytesPerBlock)
+	metadataSize := align((totalSize / (blocksPerStateByte * bytesPerBlock)) + 1)
```
niaow: I do not think that this is correct. First, `align` works differently on different targets depending on their required alignment. . .  which is not what we need here. Second, the inner expression isn't rounding down, it is rounding up and then adding 1.

If we want to round up, I think this should be:
```suggestion
	metadataSize := (totalSize + (blocksPerStateByte * bytesPerBlock) - 1) / (blocksPerStateByte * bytesPerBlock)
```

niaow (changes requested):
As currently written, this fix is not correct.

niaow:
Actually, now that I think about it more, the optimal number of metadata bytes is:
```
(totalSize + blocksPerStateByte * bytesPerBlock + bytesPerBlock - 1) / (blocksPerStateByte * bytesPerBlock + bytesPerBlock)
```

niaow:
*fixed

dgryski:
Do we need to do anything to ensure the metadataStart is appropriately aligned or does it not matter?

niaow:
The backing type of the metadata is an array of bytes. There is no alignment requirement.

dgryski:
I think I was more concerned about coming at it from the other side.  If we're allocating blocks starting from heapStart, our rounding will make sure we stop before we get to a (potentially) unaligned metadataStart.

dgryski:
Failing CI:
```
=== CONT  TestBuild/WASI/sort.go
Error: failed to run main module `/tmp/TestBuildWASItesting.go2615567584/001/test`

Caused by:
    1: wasm trap: out of bounds memory access
       wasm backtrace:
           0: 0x9981 - <unknown>!(runtime.gcBlock).state
           1: 0x10fe - <unknown>!runtime.alloc
           2: 0x299e3 - <unknown>!strconv.genericFtoa
           3: 0x3456d - <unknown>!(*fmt.fmt).fmtFloat
           4: 0x342aa - <unknown>!(*fmt.pp).fmtFloat
           5: 0x2e846 - <unknown>!(*fmt.pp).printArg
           6: 0x317a8 - <unknown>!(*fmt.pp).doPrintf
           7: 0x37a85 - <unknown>!fmt.Sprintf
           8: 0x1a478 - <unknown>!testing.tRunner
           9: 0x3d1dd - <unknown>!(*testing.T).Run
          10: 0x3e073 - <unknown>!main.TestAllLowercase
          11: 0x1a2d2 - <unknown>!testing.tRunner
          12: 0x3d1dd - <unknown>!(*testing.T).Run
          13: 0x19eb7 - <unknown>!testing.runTests$1
          14: 0x1a2d2 - <unknown>!testing.tRunner
          15: 0x15e7d - <unknown>!runtime.run$1
          16: 0x13d81 - <unknown>!runtime.run$1$gowrapper
          17: 0x3e68c - <unknown>!tinygo_launch
          18: 0x13c4a - <unknown>!_start
```

niaow:
Hmm. I suspect that we have a different issue also involved here, but not sure what it is yet.

dgryski:
Building with `gcAsserts = true` gives `panic: runtime error: gc: metadata array is too small`.  I don't *think* my other patches in this PR are anything but cosmetic, so I guess it doesn't like the new calculation?

dgryski:
My math seems to suggest this calculation should be
```
        metadataSize := (totalSize + blocksPerStateByte*bytesPerBlock) / (1 + blocksPerStateByte*bytesPerBlock)
```

Given that we don't want to track pointers in the metadata block itself, so we only need to track the metadata for the non-metadata blocks.

Derivation follows:

```
total = available + metadata
and
available = metadata * blocksPerStateByte * bytesPerBlock

So,

total = metadata * blocksPerStateByte * bytesPerBlock + metadata
total = metadata * (blocksPerStateByte * bytesPerBlock + 1)
total / (blocksPerStateByte * bytesPerBlock + 1) = metadata

Then to round metadata up, we have
metadata = (total + blocksPerStateByte * bytesPerBlock + 1 - 1) / (blocksPerStateByte * bytesPerBlock + 1)

giving
metadata := (totalSize + blocksPerStateByte * bytesPerBlock) / (blocksPerStateByte * bytesPerBlock + 1)
```

Does that sound right?  Did I miss something?

dgryski:
Hmm.. Not quite. Fails on the corpus for `github.com/google/der-ascii/cmd/ascii2der`

soypat:
Using good ol' `gopls` as a calculator in VSCode I tried the following:

```go
const (
		// ptrs
		heapStart, heapEnd = 0, 1024
		totalSize          = heapEnd - heapStart
		// =8
		meta               = (totalSize + blocksPerStateByte*bytesPerBlock) / (1 + blocksPerStateByte*bytesPerBlock)
                // = 1024
		avail = meta * blocksPerStateByte * bytesPerBlock
		// = 1032
		totalUsed = avail + meta
)
```
Shouldn't `totalUsed` always be equal or less than `totalSize`?

dgryski:
Good point. Clearly rounding doing weird things here. I'll look at this more when my kids are asleep tonight.

dgryski:
Ah, I think the problem in my example is the calculation of `avail`, since the actual code figures out the number of available blocks based on the free space remaining, *not* the number of bits that are available to be tracked in the metadata.

dgryski:
The `ascii2der` test passes if I bump the wasi stack size to 32k from 16k.

dgryski:
Hm, but the same failure here:
```
2022/01/27 21:44:20 Error: failed to run main module `/var/folders/5b/_hr1d1fd3qn4t9vtfx5p61wm0000gp/T/tinygo1673466671/main`

Caused by:
    1: wasm trap: out of bounds memory access
       wasm backtrace:
           0: 0x1c55d - (runtime.gcBlock).state
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/gc_conservative.go:154:20
           1: 0x22aa6 - (runtime.gcBlock).findNext
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/gc_conservative.go:145:13
           2: 0x1c66c - runtime.startMark
```
with `gonum/integrate/quad`

So, I'm glad we have the test corpus but I'm annoyed that we didn't see this problem until I started trying to fix the `archive/zip` issue, and now all the fixes we're trying seem to produce more errors ..

src/runtime/gc_conservative.go
```diff
@@ -240,7 +240,7 @@ func calculateHeapAddresses() {
 	totalSize := heapEnd - heapStart
 
 	// Allocate some memory to keep 2 bits of information about every block.
-	metadataSize := totalSize / (blocksPerStateByte * bytesPerBlock)
+	metadataSize := (totalSize + blocksPerStateByte*bytesPerBlock + bytesPerBlock - 1) / (blocksPerStateByte*bytesPerBlock + bytesPerBlock)
```
zdima: I think this should be 
```
metadataSize := 1+(totalSize / ( 1+blocksPerStateByte*bytesPerBlock )
```
Yet a better version is to use extra check:
```
	metadataSize := (totalSize / (1 + blocksPerStateByte*bytesPerBlock))
	for (totalSize-metadataSize)/bytesPerBlock > (metadataSize * blocksPerStateByte) {
		metadataSize++
	}

```
check this test: https://go.dev/play/p/3LRcr_8wZ38

Also, not sure how important this could be, it would be good if metadata start at aligned position.

niaow: There are no alignment requirements on the metadata.

dgryski:
I'm pretty sure at this point there are other assumptions in the garbage collector that "fixing" the off-by-one then breaks.  Probably time to add a few more asserts and see what happes.

dgryski:
```
 // Some globals + constants for the entire GC.
@@ -117,10 +117,18 @@ func (b gcBlock) address() uintptr {
 // points to an allocated object. It returns the same block if this block
 // already points to the head.
 func (b gcBlock) findHead() gcBlock {
+       var origB gcBlock
+       if gcAsserts {
+               origB = b
+       }
+
        for b.state() == blockStateTail {
                b--
        }
        if gcAsserts {
+               if b > origB {
+                       runtimePanic("gc: b wrapped while looking for head")
+               }
                if b.state() != blockStateHead && b.state() != blockStateMark {
                        runtimePanic("gc: found tail without head")
                }
@@ -131,6 +139,10 @@ func (b gcBlock) findHead() gcBlock {
 // findNext returns the first block just past the end of the tail. This may or
 // may not be the head of an object.
 func (b gcBlock) findNext() gcBlock {
+       if gcAsserts && b.address() > uintptr(metadataStart) {
+               runtimePanic("gc: block pointing inside metadata")
+
+       }
        if b.state() == blockStateHead || b.state() == blockStateMark {
                b++
        }
```

@niaow Are these assertions valid?  Should I add another commit with them?

I suppose the findNext() assert could also check after every `b++`.

dgryski:
Seems like the current failure I'm hitting is if we're at the last gcBlock, and it's allocated, and we call `findNext()`, we fall off the end of the heap.

dgryski:
Ignore the green CI: The corpus still fails due to the above-mentioned `findNext()` issue.

dgryski:
Note that even with these fixes archive/zip still effectively hangs on wasi.

niaow:
I do not think that underflow check can ever trigger on a non-AVR target, as underflow should trigger a crash. If you want to check for underflow correctly, you need to check `b == 0` inside the loop before the `b--`.

While that metadata address check is correct, I would prefer to put it inside of the address calculation rather than in `findNext`.

dgryski:
PTAL

niaow (approved):
This looks like it is at least an improvement to me. If this doesn't result in any new breakage, I think this should be okay to merge.

deadprogram:
Thanks everyone! Now merging.

deadprogram:
Is this CI fail in `dev` related to this now being merged? https://github.com/tinygo-org/tinygo/runs/5027206715?check_suite_focus=true#step:18:50

dgryski:
Can't say for sure.  I can all my test corpus through this change with `gcAsserts = true` on both native and wasi and it was fine there.  I can test if it fails under wine tomorrow, otherwise we can revert and see if the failure goes away...

deadprogram:
Ran it yet again and passed. So not related to this change, but rather ongoing issue with CI or ?

PR #2593, opened 2022-01-29
machine: add no-op io.Reader implementation to NullSerial

bgould:
When compiling with `-serial=usb` or `-serial=uart`, `machine.Serial` implements `drivers.UART`.

Currently the same is not true for `-serial=none`, because `machine.NullSerial` does not implement `io.Reader` as necessary: https://github.com/tinygo-org/drivers/blob/a15f2167cc7e46864d82e08904a3ddaa3e610d5e/uart.go#L8

Adding a no-op Read() method so that NullSerial implements `drivers.UART`

### DIFF
--- a/src/machine/machine_rp2040_gpio.go
+++ b/src/machine/machine_rp2040_gpio.go
@@ -10,7 +10,7 @@ import (
 	"unsafe"
 )
 
-type io struct {
+type rpio struct {
 	status volatile.Register32
 	ctrl   volatile.Register32
 }
@@ -22,7 +22,7 @@ type irqCtrl struct {
 }
 
 type ioBank0Type struct {
-	io                 [30]io
+	io                 [30]rpio
 	intR               [4]volatile.Register32
 	proc0IRQctrl       irqCtrl
 	proc1IRQctrl       irqCtrl
--- a/src/machine/serial.go
+++ b/src/machine/serial.go
@@ -1,6 +1,9 @@
 package machine
 
-import "errors"
+import (
+	"errors"
+	"io"
+)
 
 var errNoByte = errors.New("machine: no byte read")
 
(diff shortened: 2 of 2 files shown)


bgould:
renamed the unexported `io` struct in machine_rp2040_gpio.go to avoid naming collision with `io` package

src/machine/serial.go
```diff
@@ -44,3 +47,14 @@ func (ns NullSerial) Buffered() int {
 func (ns NullSerial) Write(p []byte) (n int, err error) {
 	return len(p), nil
 }
+
+// Read is a no-op; always returns that zero bytes were read with no error
+func (ns NullSerial) Read(p []byte) (n int, err error) {
+	return 0, nil
```
pymq, line 53: Just wondering, shouldn't it return io.EOF to show that there is nothing to read?

As per doc
> // Implementations of Read are discouraged from returning a
// zero byte count with a nil error, except when len(p) == 0.
// Callers should treat a return of 0 and nil as indicating that
// nothing happened; in particular it does not indicate EOF.

The same as `ReadByte()` always returns an error

https://github.com/tinygo-org/tinygo/blob/3458726234bdd758cca3d55cc72af83954f6379d/src/machine/serial.go#L32-L34

aykevl: Agreed, returning an error seems more correct to me.

src/machine/serial.go
```diff
@@ -44,3 +47,14 @@ func (ns NullSerial) Buffered() int {
...
+}
+
+// type check to ensure NullSerial implements interfaces for drivers.UART
+var (
+	_ io.Reader = NullSerial{}
+	_ io.Writer = NullSerial{}
```
aykevl, line 59: :+1:

src/machine/machine_rp2040_gpio.go
```diff
@@ -22,7 +22,7 @@ type irqCtrl struct {
 }
 
 type ioBank0Type struct {
-	io                 [30]io
+	io                 [30]rpio
```
aykevl, line 25: Was this change intentional?

bgould: Importing the `io` package as part of the type check causes a compile error in the smoke test for RP2040 because it defines an internal type called `io` as well.  The alternative would have been to import `io` with an alias in `serial.go`, but I thought it would be better and more appropriate to rename the type for the rp2040 to prevent this problem from happening again in the future.

src/machine/serial.go
bgould: Do you think should return `io.EOF` or should it return `errNoByte` (the same as `ReadByte`)?

deadprogram:
Closing due to bitrot. Please open a new PR against the latest if this is still needed. Thank you @bgould

PR #2594, opened 2022-01-29
machine/rp2040: fix PWM bug. Resolve #2592 [closed due to bad git HEAD]

### LINKED ISSUE #2592: PWM only produces max frequency of 1.8khz
No matter what value I give "period" in the code below, I can't get a higher PWM on the pin of 1.8Khz.

Save the following file as main.go....
[main.txt](https://github.com/tinygo-org/tinygo/files/7962672/main.txt)

soypat:
PTAL. Have not had a chance to test it but should regardless be more correct. Resolves https://github.com/tinygo-org/tinygo/issues/2592


soypat:
I'm going to go ahead and match dev head with a new PR. sorry.

PR #2595, opened 2022-01-29
machine/rp2040: fix PWM bug. Resolve #2592

### LINKED ISSUE #2592: PWM only produces max frequency of 1.8khz

soypat:
PTAL. Have not had a chance to test it but should regardless be more correct. Resolves https://github.com/tinygo-org/tinygo/issues/2592

### DIFF
--- a/src/machine/machine_rp2040_pwm.go
+++ b/src/machine/machine_rp2040_pwm.go
@@ -278,8 +278,14 @@ func (pwm *pwmGroup) setPeriod(period uint64) error {
 	rhs := 16 * period / ((1 + phc) * periodPerCycle * (1 + topStart)) // right-hand-side of equation, scaled so frac is not divided
 	whole := rhs / 16
 	frac := rhs % 16
-	if whole > 0xff {
+	switch {
+	case whole > 0xff:
 		whole = 0xff
+	case whole == 0:
+		// whole calculation underflowed so setting to minimum
+		// permissible value in DIV_INT register.
+		whole = 1
+		frac = 0
 	}
 
 	// Step 2 is acquiring a better top value. Clearing the equation:


src/machine/machine_rp2040_pwm.go
```diff
@@ -280,6 +280,11 @@ func (pwm *pwmGroup) setPeriod(period uint64) error {
 	frac := rhs % 16
 	if whole > 0xff {
```
deadprogram: Perhaps change all of this to:

```go
switch {
   case whole == 0:
      // do thing
   case whole > 0xff:
      // do other thing
}
```

soypat: Great idea! I've force pushed changes.

src/machine/machine_rp2040_pwm.go
```diff
@@ -278,8 +278,14 @@ func (pwm *pwmGroup) setPeriod(period uint64) error {
 	rhs := 16 * period / ((1 + phc) * periodPerCycle * (1 + topStart)) // right-hand-side of equation, scaled so frac is not divided
 	whole := rhs / 16
 	frac := rhs % 16
-	if whole > 0xff {
+	switch whole {
+	case 0xff:
```
kenbell: This isn't equivalent?  This won't clamp 0x100 since it's not exactly 0xff?

soypat: darn. I was about to break it even more. Sorry, I'm juggling university work and tinygo in a dangerous mix.

deadprogram:
Thanks for the fix @soypat and for the review @kenbell now merging.

PR #2596, opened 2022-01-30
rp2040: watchdog

ysoldak:
Implements hardware watchdog timer for RP2040.

Implementation inspired by [Pico SDK](https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_watchdog/watchdog.c)
It ended up quite different though.

I could not use `time.Duration` to have nice "period" configuration parameter instead of "ticks" due to an import loop.

Examples demonstrate basic usage, watchdog triggering on runtime panic and while spinning.
Basic example added to the list of smoke tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -477,6 +477,9 @@ endif
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=feather-m4          examples/pwm
 	@$(MD5SUM) test.hex
+	# test watchdog
+	$(TINYGO) build -size short -o test.hex -target=nano-rp2040         examples/watchdog/basic
+	@$(MD5SUM) test.hex
 ifneq ($(STM32), 0)
 	$(TINYGO) build -size short -o test.hex -target=bluepill            examples/blinky1
 	@$(MD5SUM) test.hex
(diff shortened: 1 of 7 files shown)


ysoldak:
This is printout from the basic example:
```
Hardware reset
10
9
8
7
6
5
4
3
2
1
0

Timer reset
10
9
8
7
6

Force reset
10
9
8
7
6
5
4
3
2
1
0

Timer reset
10
9
8
```

ysoldak:
@ofauchon What do you think?

ysoldak:
I've observed a strange behaviour when building with this change.
Time seem to flow much slower with these changes. I've probably messed up with some timer.

This behaviour manifests itself quite remarkably: while the board is plugged in and flashed (`tinygo flash`) it works just fine as usual. But once you unplug the board and plug it in again (hard power cycle) you can directly tell your code runs much slower, very easy to see with `blinky1` example.

deadprogram:
Now that https://github.com/tinygo-org/tinygo/pull/3855 was merged, this PR seems obsolete, so should probably close it.

ysoldak:
> Now that #3855 was merged, this PR seems obsolete, so should probably close it.

That PR does not cover all functionality though. I'll try and rebase and sync with dev branch.

deadprogram:
@ysoldak just noticed this PR again. What do you think about it?

ysoldak:
@deadprogram I'll take a look. If nothing from me in a week, feel free to close it.

PR #2603, opened 2022-02-03
os: disable TestClearenv for now, it is very flaky. Also reduce verbosity.

dkegel-fastly:
The problem with TestClearenv is that it assumes os.Environ() returns a copy of the environment, but in tinygo, it returns pointers to the live environment.  Thus in tinygo, os.Environ() can go invalid the moment any environment variable is deleted.

Probably need to decide whether this behavior is ok, and if so, rewrite that test to call os.Environ() after each deletion, or something.

See https://github.com/tinygo-org/tinygo/issues/2646

### DIFF
--- a/src/os/env_test.go
+++ b/src/os/env_test.go
@@ -122,33 +122,7 @@ func TestUnsetenv(t *testing.T) {
 	}
 }
 
-func TestClearenv(t *testing.T) {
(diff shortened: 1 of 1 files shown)


deadprogram:
Thanks @dkegel-fastly merging as the most pragmatic for now.

PR #2604, opened 2022-02-03
Add board support for ESP32-C3-12f Kit

zdima:
Provide board LED pins and other pins assignment.


aykevl (changes requested):
Can you also add this board to the README and add a smoke test to the Makefile?
Other than that, this PR looks good to me.

zdima:
@aykevl I updated readme and added smoke test for new board

deadprogram:
Thanks @zdima sorry about the delay. Now merging.

PR #2605, opened 2022-02-03
os: isWine: be compatible with older versions of wine, too

dkegel-fastly:
Turns out wine 5.x doesn't export WINEUSERNAME.


aykevl (approved):
Seems reasonable to me.

PR #2606, opened 2022-02-03
Implement os.RemoveAll()

dkegel-fastly:
Tests could use a little more love sometime, but this is good enough for now.

TODO: removeall_test.go: on windows: enable once Readdir is implemented
TODO: read_test.go: on wasi: enable after figuring out stat-on-a-symlink issue

### DIFF
--- a/src/os/read_test.go
+++ b/src/os/read_test.go
@@ -1,0 +1,102 @@
+//go:build !baremetal && !js && !wasi
+// +build !baremetal,!js,!wasi
+
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package os_test
+
+import (
+	"bytes"
+	. "os"
+	"path/filepath"
+	"testing"
+)
+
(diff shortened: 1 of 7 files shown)


src/os/read_test.go
```diff
@@ -0,0 +1,101 @@
...
+	. "os"
+	"path/filepath"
+	"testing"
+)
+
+func checkNamedSize(t *testing.T, path string, size int64) {
+	// TODO: this statement fails on wasi, possibly it objects to reading Stat on a symlink
+	dir, err := Stat(path)
+	if err != nil {
+		// TODO: on wasi, t.Fatalf crashes with null ptr when printing err
```
dgryski: Remove the TODO since we know why this fails.

dgryski (approved):
This is pretty much all upstream Go wth the appropriate copyright tags.  This LGTM.

src/os/read_test.go
dkegel-fastly: Removed already in the code you reviewed, I think.

deadprogram:

PR #2608, opened 2022-02-03
Fix LLVM build constraints

QuLogic:
The include and linker search paths probably do nothing since they won't exist, but `-lffi` will break building if you don't have `libffi.so` installed.

### DIFF
--- a/cgo/libclang_config_llvm11.go
+++ b/cgo/libclang_config_llvm11.go
@@ -5,12 +5,12 @@ package cgo
 
 /*
 #cgo linux        CFLAGS:  -I/usr/lib/llvm-11/include
-#cgo darwin amd64 CFLAGS:  -I/usr/local/opt/llvm@11/include
-#cgo darwin arm64 CFLAGS:  -I/opt/homebrew/opt/llvm@11/include
+#cgo darwin,amd64 CFLAGS:  -I/usr/local/opt/llvm@11/include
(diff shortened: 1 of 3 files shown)


aykevl:
Ah, I see. The issue is that `darwin amd64` means `darwin || amd64`, not `darwin && amd64` as I expected (the build constrains used here were my suggestion). However, I don't think the fix here is correct. It passes CI only because the files aren't used on MacOS in CI (but are used by TinyGo developers).

aykevl:
Not sure what the correct syntax is. Maybe `darwin,amd64`? I don't have any hardware I can test this on, unfortunately.

QuLogic:
Unfortunately, when the new `go:build` constraints were added, all the docs for the old form were removed. I guess they forgot that `cgo` constraints still use the old syntax.

Looking at the [`go:build` draft proposal](https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md#background-syntax), you can see that the old syntax includes an example of `GOOS/GOARCH`:

> For example, this line indicated that the file should build on Linux for any architecture, or on Windows only for 386:
>
> ```
> // +build linux windows/386
> ```

But I didn't read far enough, as it looks like comma replaced slash:

> In January 2012, [CL 5554079](https://golang.org/cl/5554079) added support for custom build tags (such as appengine), **changed slash to comma,** and introduced ! for negation.

So I'll switch to that here.

deadprogram:
Makes sense to me. Thanks @QuLogic now merging.

PR #2613, opened 2022-02-04
rp2040: I2C minor fixes

ysoldak:
- Removes parasitic printouts
- Use `internal/itoa` instead of `strconv`
- Return abort reason instead of byte index (discussed with @soypat on slack and agreed that would be more appropriate)


soypat:
LGTM!

deadprogram:
Thank you @ysoldak and @soypat now merging.

PR #2614, opened 2022-02-05
rp2040: fix incorrect inline assembly

aykevl:
The register r0 was used unconditionally. This is a bug: the compiler doesn't know it is clobbered and might consider it alive across the inline assembly expression.

The fix is simple. It could probably be optimized, but this should at least fix the bug. I discovered it while working on LLVM 14 support.

---

Note: this is UNTESTED. I don't have a rp2040 with me at the moment so can't test it myself.


ysoldak:
How to test? I can try if understand what to look for.
Is it enough just to flash "blink" example and see if it still works?

deadprogram:
Tested and worked as expected for me, e.g. no change.

Merging, thanks @aykevl

PR #2586, opened 2022-01-25
builder: add support for cross compiling to Darwin

aykevl:
This means that it will be possible to generate a Darwin binary on any platform (Windows, Linux, and MacOS of course), including CGo. Of course, the resulting binaries can only run on MacOS itself.

The binary links against libSystem.dylib, which is a shared library. The macos-minimal-sdk repository (created by me) contains open source header files and generated symbol stubs so we can generate a stub libSystem.dylib without copying any closed source code.

PR #2615, opened 2022-02-05
Implement some missing Signal bits

QuLogic:
Some bits of the `os.Signal` interface were not implemented, so I copied them out of the `syscall_js.go` from upstream Go. Also added the "always-available" (but not really, because the build constraints) signal aliases to `os`.

These are needed for parts of the fuzzing implementation in Go 1.18.

### DIFF
--- a/src/os/exec.go
+++ b/src/os/exec.go
@@ -55,3 +55,7 @@ func (p *Process) Wait() (*ProcessState, error) {
 func (p *Process) Kill() error {
 	return ErrNotImplemented
 }
+
+func (p *Process) Signal(sig Signal) error {
+	return ErrNotImplemented
+}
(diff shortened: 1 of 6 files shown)


QuLogic:
Also added `os.Process.Signal` which is needed in `src/internal/testenv/testenv.go` for running `make tinygo-test`.

dkegel-fastly:
If you don't have a mac handy, maybe https://github.com/tinygo-org/tinygo/pull/2586 could be of some use!

QuLogic:
Fortunately, it appears the errors are the same as the other systems (i.e., missing `Signal.Signal` implementation), so it should be sufficient to copy it in the same way.

deadprogram:
Seems reasonable to me. @QuLogic can you please rebase against `dev` just to make sure this still builds correctly on all platforms? Thanks.

QuLogic:
Seems good still.

deadprogram:
Thanks @QuLogic now merging.

PR #2617, opened 2022-02-06
testing: implement testing.TempDir

dkegel-fastly:
Also enable testing/fstest and io/fs in "make tinygo-test" as first consumers.

Not supported yet on windows and wasi, where os.Readdir and os.RemoveAll are unimplemented.


dgryski (approved):
LGTM

niaow (approved):

PR #2618, opened 2022-02-06
board: add definition for Teensy 4.1

bgould:


deadprogram:
Looking very good @bgould 

Can you please add to smoketests alongside https://github.com/tinygo-org/tinygo/blob/dev/Makefile#L445

bgould:
thanks for reminding me @deadprogram I always seem to forget that.  It is added now, as well as updated the readme.

deadprogram:
Thanks @bgould now merging.

PR #2619, opened 2022-02-06
os: Use a uintptr for NewFile

QuLogic:
This appears to have been how it is upstream for about 10 years. Using an interface breaks if a file descriptor number is passed directly to `NewFile`, e.g., `NewFile(3, "fuzz_in")` as used in [Go 1.18 fuzzing code](https://github.com/golang/go/blob/d588f487703e773ba4a2f0a04f2d4141610bff6b/src/internal/fuzz/sys_posix.go#L61).

### DIFF
--- a/src/os/file_anyos.go
+++ b/src/os/file_anyos.go
@@ -21,9 +21,9 @@ func init() {
 // Stdin, Stdout, and Stderr are open Files pointing to the standard input,
 // standard output, and standard error file descriptors.
 var (
-	Stdin  = NewFile(unixFileHandle(syscall.Stdin), "/dev/stdin")
-	Stdout = NewFile(unixFileHandle(syscall.Stdout), "/dev/stdout")
-	Stderr = NewFile(unixFileHandle(syscall.Stderr), "/dev/stderr")
+	Stdin  = NewFile(uintptr(syscall.Stdin), "/dev/stdin")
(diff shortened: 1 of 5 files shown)


dkegel-fastly:
Here's when tinygo changed to not use a uintptr directly:
https://github.com/tinygo-org/tinygo/pull/1012

So the essence of your change is that since file_unix.go and file_other.go use disjoint build constraints, they could go back to sharing uintptr rather than defining separate file handle types...?

dgryski:
This seems like a reasonable change, especially if that's what upstream does.  Thoughts @aykevl since you were the one who moved it away from a `uintptr` ?

QuLogic:
> So the essence of your change is that since file_unix.go and file_other.go use disjoint build constraints, they could go back to sharing uintptr rather than defining separate file handle types...?

Well, I think they do still have to define their own separate file handle types; it's just that the `FileHandle` type is no longer the public API for it, but `uintptr` instead.

aykevl:
Thank you! Yes, this seems like a reasonable change to improve compatibility.

aykevl:
Somewhat unrelated: I've been thinking we should perhaps remove the whole virtual filesystem thing and instead let application code rely on `io/fs` (which didn't exist yet when I implemented VFS support). This works for read-only cases but doesn't work if writing to files is also needed.
But this should perhaps be discussed in a different place.

PR #2620, opened 2022-02-07
Fix cross-Linux setup on non-amd64 arches

QuLogic:
In that case, an emulator is needed for amd64, and tests should be run for amd64 as a _cross_ test.

I'm not 100% on the 386 and arm sides as it turns out qemu is broken there for some arches, but it works to put amd64 in qemu.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -31,6 +31,13 @@ const TESTDATA = "testdata"
 
 var testTarget = flag.String("target", "", "override test target")
 
+var supportedLinuxArches = map[string]string{
+	"AMD64Linux": "linux/amd64",
+	"X86Linux":   "linux/386",
+	"ARMLinux":   "linux/arm/6",
+	"ARM64Linux": "linux/arm64",
+}
+
 var sema = make(chan struct{}, runtime.NumCPU())
 
 func TestBuild(t *testing.T) {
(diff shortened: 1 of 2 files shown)


main_test.go
```diff
@@ -31,6 +31,13 @@ const TESTDATA = "testdata"
 
 var testTarget = flag.String("target", "", "override test target")
 
+var supportedArches = map[string]string{
```
aykevl: We support other OS/arch combinations too, such as windows/amd64. Maybe we can call this `supportedLinuxArches` or something like that?

aykevl (commented):
Looks good to me, with one nit.
Tests were certainly rather amd64-centered because that's what most people use and that's what in CI. It's good to make them more portable.

main_test.go
QuLogic: Yes, good point. I suppose once #2586 is in, this could maybe be expanded, but I'll change this now.

aykevl: > I suppose once https://github.com/tinygo-org/tinygo/pull/2586 is in, this could maybe be expanded, but I'll change this now.

I don't think so: #2586 is only for cross compiling. You can't actually run those binaries on the local host.

PR #2625, opened 2022-02-08
os, runtime: enable os.Stdin for baremetal target

sago35:
This PR enables os.Stdin for the baremetal target.

There are two items that need to be discussed and agreement.

1. ~~Can an os package import a machine package?~~
    * solved: Could not import machine package from os package. I used go:link instead.
2. ~~There are changes that will *break compatibility*~~
    * ~~I've changed `src/machine/uart.go` and `src/machine/usb.go`, and the behavior changes significantly.~~
    * ~~`bufio.Scanner.Scan()` will not work without this change (error message is `multiple Read calls return no data or error`)~~
    * This PR only adds os.Stdin, so there are no compatibility-breaking changes.

### DIFF
--- a/src/os/file_other.go
+++ b/src/os/file_other.go
@@ -38,9 +38,26 @@ func NewFile(fd uintptr, name string) *File {
 	return &File{&file{stdioFileHandle(fd), name}}
 }
 
-// Read is unsupported on this system.
+// Read reads up to len(b) bytes from machine.Serial.
+// It returns the number of bytes read and any error encountered.
 func (f stdioFileHandle) Read(b []byte) (n int, err error) {
-	return 0, ErrUnsupported
+	if len(b) == 0 {
+		return 0, nil
+	}
+
+	size := buffered()
+	for size == 0 {
+		gosched()
+		size = buffered()
+	}
+
+	if size > len(b) {
+		size = len(b)
+	}
+	for i := 0; i < size; i++ {
+		b[i] = getchar()
+	}
+	return size, nil
 }
 
 func (f stdioFileHandle) ReadAt(b []byte, off int64) (n int, err error) {
@@ -77,6 +94,15 @@ func (f stdioFileHandle) Fd() uintptr {
 
 //go:linkname putchar runtime.putchar
 func putchar(c byte)
+
+//go:linkname getchar runtime.getchar
+func getchar() byte
+
+//go:linkname buffered runtime.buffered
+func buffered() int
+
+//go:linkname gosched runtime.Gosched
+func gosched() int
 
 func Pipe() (r *File, w *File, err error) {
 	return nil, nil, ErrNotImplemented
--- a/Makefile
+++ b/Makefile
@@ -358,6 +358,8 @@ smoketest:
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=pca10040            examples/echo
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=pca10040            examples/echo2
+	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=circuitplay-express examples/i2s
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=pca10040            examples/mcp3008
--- a/src/examples/echo2/echo2.go
+++ b/src/examples/echo2/echo2.go
@@ -1,0 +1,31 @@
+// This is a echo console running on the os.Stdin and os.Stdout.
+// Stdin and os.Stdout are connected to machine.Serial in the baremetal target.
+//
+// Serial can be switched with the -serial option as follows
+// 1. tinygo flash -target wioterminal -serial usb examples/echo2
+// 2. tinygo flash -target wioterminal -serial uart examples/echo2
+//
+// This example will also work with standard Go.
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+)
+
+func main() {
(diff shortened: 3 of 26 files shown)


src/os/file_other.go
```diff
@@ -38,7 +39,7 @@ func NewFile(fd FileHandle, name string) *File {
 
 // Read is unsupported on this system.
 func (f stdioFileHandle) Read(b []byte) (n int, err error) {
-	return 0, ErrUnsupported
+	return machine.Serial.Read(b)
```
sago35: Maybe it would be better to change here without changing uart.go and usb.go.
I think it would be better to call runtime.Gosched(), but it can't be called from the machine package.

```go
func (f stdioFileHandle) Read(b []byte) (n int, err error) {
	size := 0
	for size == 0 {
		size = machine.Serial.Buffered()
		runtime.Gosched()
	}
	return machine.Serial.Read(b)
}
```

src/os/file_other.go
```diff
@@ -38,7 +39,7 @@ func NewFile(fd FileHandle, name string) *File {
 
 // Read is unsupported on this system.
```
sago35:
```suggestion
// Read reads up to len(b) bytes from machine.Serial.
// It returns the number of bytes read and any error encountered.
```

deadprogram:
Looks like CI fail here https://github.com/tinygo-org/tinygo/runs/5110984825?check_suite_focus=true#step:17:492

kenbell:
@sago35 if I read this correctly, it makes UART blocking on read whereas it was non-blocking before.  I can see this breaking existing code.

it would be better to put the blocking behaviour outside of UART implementation?

src/os/file_other.go
kenbell: I think this proposal is much better

sago35:
@kenbell 
I also want to maintain compatibility as much as possible.
However, if you do not block it, bufio.Scanner.Scan will fail.
In other words, I believe that the standard Go blocks it.

sago35:
If we do not change uart or usbcdc, but only os.Stdin, the following problem occurs.
This will probably be very confusing.

```go
// It works as expected.
scanner := bufio.NewScanner(os.Stdin)
```

```go
// Not working as expected.
scanner := bufio.NewScanner(machine.Serial)
```

src/os/file_other.go
sago35: https://github.com/tinygo-org/tinygo/pull/2625#issuecomment-1033627274

sago35:
@kenbell @deadprogram 
I would be happy to talk more about os.Stdin in the following thread.
https://gophers.slack.com/archives/CN4R2DQV7/p1646872768911309

src/os/file_other.go
sago35: Updated to reflect @kenbell's point.
The problem with using machine.UART has been separated in #2739.

sago35:
> if I read this correctly, it makes UART blocking on read whereas it was non-blocking before. I can see this breaking existing code.
> it would be better to put the blocking behaviour outside of UART implementation?

This PR only adds os.Stdin, so there are no compatibility-breaking changes.

sago35:
@kenbell @deadprogram 
Compatibility issues have been resolved and are ready for review.
Changes impacting compatibility have been moved to #2739.

With this PR, the following code will work in both standard Go and tinygo-baremetal.
I would like to see this change merged, as it would considerably improve the efficiency of development in baremetal.

https://github.com/tinygo-org/tinygo/blob/cc3009b5c15e0772f9f62c3c0ff6789e2bb961cf/src/examples/echo2/echo2.go

src/os/file_other.go
```diff
@@ -78,6 +95,15 @@ func (f stdioFileHandle) Fd() uintptr {
 //go:linkname putchar runtime.putchar
 func putchar(c byte)
 
+//go:linkname getchar runtime.getchar
+func getchar() byte
+
+//go:linkname buffered runtime.buffered
+func buffered() int
+
+//go:linkname gosched runtime.Gosched
```
deadprogram, line 104: Does this need to be defined here? I thought it already was elsewhere.

sago35: Defined below, but I believe the following is the definition of `machine.gosched()`.
Since it is not possible to import runtime packages, etc. from os packages, `go:linkname` is used here.
This defines `os.gosched()`.

https://github.com/tinygo-org/tinygo/blob/cc3009b5c15e0772f9f62c3c0ff6789e2bb961cf/src/machine/machine_nxpmk66f18_uart.go#L63-L64

deadprogram:
@ardnew @bgould any implications in this PR for the USB 2 work that is happening?

ardnew:
> @ardnew @bgould any implications in this PR for the USB 2 work that is happening?

After a cursory review, I don't see any potential blockers with respect to the USB 2.0 package. But I'm still a little confused as to what we gain with this capability. Is this to create a type-safe, consistent IO interface? As opposed to the awkward juggling we are currently doing with `Serial`/`UART`/`USB` constants?

sago35:
What we get with this PR is support for `os.Stdin.Read()` on baremetal targets.
With this PR, what `os.Stdin` points to is `machine.Serial`.

~~But now I see that `os.Stdin.ReadAt()` is not implemented and should be added.~~
added: The implementation of `os.Stdin.ReadAt()` is covered in another PR.

old :
https://github.com/tinygo-org/tinygo/blob/486b99961d1ef918341ce2f7d29b20388da253dc/src/os/file_other.go#L41-L48

deadprogram:
Any other feedback or comments before we merge this?

deadprogram:
Merging in order to keep this process with improvements for serial/USB communication moving along. Thanks @sago35

PR #2626, opened 2022-02-08
machine: add DefaultUART to wioterminal

sago35:
Activate the -serial option for tinygo.


deadprogram:
Thanks @sago35 now merging.

PR #2628, opened 2022-02-11
compiler: move allocations > 256 bytes to the heap

dgryski:


aykevl:
What's the rationale for this change?
(I'm not against it, but didn't expect this to be an issue anywhere).

dgryski:
In the snappy encoder on wasi, it tries to allocate a 32kb object on a 16k stack and the subsequent memset() call wipes out some globals, including `metadataStart`, corrupting the garbage collector: https://github.com/golang/snappy/blob/master/encode_other.go#L130

aykevl:
I see. I just read the #tinygo-dev discussion :)

It does seem reasonable to me to do this: a 32kB stack object is just way too large. 256 bytes is maybe a little bit on the low side? I think it depends on the target and the amount of stack/heap allocated for it (256 bytes is a lot on AVR, but 1kB would be perfectly fine on x86).

aykevl (approved):
Let's just merge this as it does fix a bug.

compiler/compiler.go
```diff
@@ -1571,8 +1571,9 @@ func (b *builder) createExpr(expr ssa.Value) (llvm.Value, error) {
 	switch expr := expr.(type) {
 	case *ssa.Alloc:
 		typ := b.getLLVMType(expr.Type().Underlying().(*types.Pointer).Elem())
-		if expr.Heap {
-			size := b.targetData.TypeAllocSize(typ)
+		size := b.targetData.TypeAllocSize(typ)
+		// move all allocations larger than 256 bytes to the heap
```
aykevl: Actually, can you maybe add a comment that this matches `maxStackAllocs` in transform/allocs.go? And maybe also a comment to transform/allocs.go to point back here? Just so that anybody looking at the code in the future knows that both values need to be updated at the same time.

dgryski: Comments added.

deadprogram:

PR #2629, opened 2022-02-11
i2c for esp32-c3

zdima:
Added support for I2C with 7-bit addressing for ESP32C3.
Tested with bh1750 (reads) and SSD1306 (writes).

### DIFF
--- a/src/machine/i2c.go
+++ b/src/machine/i2c.go
@@ -1,5 +1,5 @@
-//go:build atmega || nrf || sam || stm32 || fe310 || k210 || rp2040
-// +build atmega nrf sam stm32 fe310 k210 rp2040
+//go:build atmega || nrf || sam || stm32 || fe310 || k210 || rp2040 || esp32c3
+// +build atmega nrf sam stm32 fe310 k210 rp2040 esp32c3
 
 package machine
 
--- a/src/machine/machine_esp32c3.go
+++ b/src/machine/machine_esp32c3.go
@@ -405,6 +405,7 @@ func (i2c *I2C) transmit(addr uint16, cmd []i2cCommand, timeoutMS int) error {
 			}
 			count := 32
 			bytes := len(c.data) - c.head
+			// Only last byte in sequence must be sent with ACK set to 1 to indicate end of data.
 			split := bytes <= count
 			if split {
 				bytes--
@@ -412,16 +413,17 @@ func (i2c *I2C) transmit(addr uint16, cmd []i2cCommand, timeoutMS int) error {
 			if bytes > 32 {
 				bytes = 32
 			}
-			readTo = c.data[c.head : c.head+bytes]
 			reg.Set(i2cCMD_READ | uint32(bytes) | c.ack)
 			reg = (*volatile.Register32)(unsafe.Pointer((uintptr(unsafe.Pointer(reg)) + 4)))
 
 			if split {
 				reg.Set(i2cCMD_READ | 1 | c.ack)
 				reg = (*volatile.Register32)(unsafe.Pointer((uintptr(unsafe.Pointer(reg)) + 4)))
+				readTo = c.data[c.head : c.head+bytes+1] // read bytes + 1 last byte
 				cmdIdx++
 			} else {
 				reg.Set(i2cCMD_END)
+				readTo = c.data[c.head : c.head+bytes]
 				reg = nil
 			}
 
--- a/src/machine/machine_esp32c3.go
+++ b/src/machine/machine_esp32c3.go
@@ -316,23 +316,17 @@ type i2cCommandType = uint32
 type i2cAck = uint32
 
 const (
-	i2cCMD_RSTART i2cCommandType = 6 << 11
-	i2cCMD_WRITE  i2cCommandType = 1<<11 | 1<<8 // WRITE + ack_check_en
-	i2cCMD_READ   i2cCommandType = 3 << 11
-	i2cCMD_STOP   i2cCommandType = 2 << 11
-	i2cCMD_END    i2cCommandType = 4 << 11
-)
-
-const (
-	i2cAckNone i2cAck = 0
-	i2cAckHigh i2cAck = 0x00000700
-	i2cAckLow  i2cAck = 0x00000100
+	i2cCMD_RSTART   i2cCommandType = 6 << 11
+	i2cCMD_WRITE    i2cCommandType = 1<<11 | 1<<8 // WRITE + ack_check_en
+	i2cCMD_READ     i2cCommandType = 3<<11 | 1<<8 // READ + ack_check_en
+	i2cCMD_READLAST i2cCommandType = 3<<11 | 5<<8 // READ + ack_check_en + NACK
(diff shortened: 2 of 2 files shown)


src/machine/i2c.go
```diff
@@ -1,5 +1,4 @@
-//go:build atmega || nrf || sam || stm32 || fe310 || k210 || rp2040
-// +build atmega nrf sam stm32 fe310 k210 rp2040
+// +build atmega nrf sam stm32 fe310 k210 rp2040 esp32c3
```
sago35, line 2: Please do not delete definitions from go 1.17 or later.

```suggestion
//go:build atmega || nrf || sam || stm32 || fe310 || k210 || rp2040 || esp32c3
// +build atmega nrf sam stm32 fe310 k210 rp2040 esp32c3
```

sago35:
@zdima 
I tried with the ADT7410 example.
However, m5stamp-c3 did not produce the expected waveform.
feather-m4-can (atsamd51) is working correctly.
Can you think of any possible causes?

m5stamp-c3
image

feather-m4-can
image

```go
package main

import (
	"fmt"
	"machine"
	"time"

	"tinygo.org/x/drivers/adt7410"
)

var (
	i2c    = machine.I2C0
	sensor = adt7410.New(i2c)
)

func main() {
	i2c.Configure(machine.I2CConfig{Frequency: machine.TWI_FREQ_400KHZ, SCL: machine.IO6, SDA: machine.IO7})
	//i2c.Configure(machine.I2CConfig{Frequency: machine.TWI_FREQ_400KHZ})
	sensor.Configure()

	for {
		temp := sensor.ReadTempC()
		time.Sleep(time.Second)
	}

}
```

sago35:
I rewrote it as follows, and it works a little better.
But it still doesn't work well.

m5stamp-c3
image

```go
func main() {

	i2c := machine.I2C0
	i2c.Configure(machine.I2CConfig{Frequency: machine.TWI_FREQ_400KHZ, SCL: machine.IO6, SDA: machine.IO7})
	//i2c.Configure(machine.I2CConfig{Frequency: machine.TWI_FREQ_400KHZ})
	sensor := adt7410.New(i2c)
	sensor.Configure()

	for {
		temp := sensor.ReadTempC()
		time.Sleep(time.Second)
	}

}
```

sago35:
I have confirmed that the SSD1306 does work.

src/machine/i2c.go
zdima: my bad. I have fix it.

zdima:
@sago35 

> @zdima I tried with the ADT7410 example. However, m5stamp-c3 did not produce the expected waveform. feather-m4-can (atsamd51) is working correctly. Can you think of any possible causes?

From the wave form it look like the signal capture is partial. The signal with address is not captured and only data portion shown.

> I rewrote it as follows, and it works a little better.

I think this is because sensor is initialized with i2c interface before i2c get initialized. This is only thing that may provide unexpected result. In case you need to keep sensor as a global variable, you could add function that return configured i2c interface and use it to initialize sensor while program is loading.

sago35:
@zdima 
I did some more research.

---

The following source code cannot be loaded by m5stamp-c3.
It looks like there is no address transmission on read.


```go
		i2c.Tx(0x48, []byte{0x00}, buf[:2])
```

m5stamp-c3
image

feather-m4-can
image


---

In the case of the following source code, m5stamp-c3 can also read the data of ADT7410.


```go
		i2c.Tx(0x48, []byte{0x00}, []byte{})
		i2c.Tx(0x48, []byte{}, buf[:2])
```

m5stamp-c3
image

feather-m4-can
image

sago35:
@zdima can you please rebase this branch against dev?

zdima:
@sago35 I don't have I2C device with registers. Can you try new code and let me know if this fix the issue?

sago35:
@zdima 

The waveform is now good enough to communicate.

image

However, there are some waveform changes that are too fast, perhaps due to timing issues with register operations.

image

image


In addition, this process reads 2 bytes, but the second byte is always 0xFF.
I don't understand what the following source code is doing, but it seems to me that only one byte is being processed.

https://github.com/zdima/tinygo/blob/2bfdd894e76138efca8b3738cc22a9bd1a85d089/src/machine/machine_esp32c3.go#L409-L411

zdima:
> However, there are some waveform changes that are too fast,

The registry operations are all complete before it start sending the entire sequence. I don't think I can do anything with this.

> I don't understand what the following source code is doing

The reason for this code is to set for last byte NACK. 

> this process reads 2 bytes, but the second byte is always 0xFF.

Yes, looks like because of this twist with the last byte, the last byte was not getting read from the controller.
Please try the last changes.

sago35:
ADT7410 is now working.
Thank you for the quick response.

src/machine/machine_esp32c3.go
```diff
@@ -152,3 +156,324 @@ func (uart *UART) WriteByte(b byte) error {
...
+	}
+	return nil
+}
+
+// Tx does a single I2C transaction at the specified address.
+// It clocks out the given address, writes the bytes in w, reads back len(r)
+// bytes and stores them in r, and generates a stop condition on the bus.
+func (i2c *I2C) Tx(addr uint16, w, r []byte) (err error) {
+	// timeout in microseconds.
+	const timeout = 100 * 1000 // 40ms is a reasonable time for a real-time system.
```
sago35, line 466: Probably 40 is correct here.

```suggestion
	const timeout = 40 // 40ms is a reasonable time for a real-time system.
```

sago35:
@zdima Now, it looks like it is failing to read a single byte.
Here, we want to read only one byte from address==0xA1, but it keeps reading until timeout.

https://github.com/tinygo-org/drivers/blob/a15f2167cc7e46864d82e08904a3ddaa3e610d5e/bme280/bme280.go#L65-L69

image

osechet:
Hi @zdima!
is there any news about this PR? I just tested it with a [Lolin C3 Mini v1](https://www.wemos.cc/en/latest/c3/c3_mini_1_0_0.html) and an SSD1306 OLED display and it works perfectly. The test program is visible here if it can help: https://gist.github.com/osechet/d6d3e2d3bc8ad33435705c3a94dc85af.

The program is flashed using the _esp32c3_ target: `tinygo flash -port COM3 -target esp32c3 .\ssd1306`
tinygo version: 0.27.0 windows/amd64
go version: 1.19.4

deadprogram:
I am also very interested in finding out the status of this PR and helping out however needed. :smile_cat:

deadprogram:
Just submitted PR https://github.com/tinygo-org/tinygo/pull/4011 

What do we think is needed if anything to get this PR completed?

deadprogram:
I did some testing last night.

The good news is that with a couple small changes this branch worked to read/write data.
The not as good news is that as @sago35 reported the PR does not work correctly when writing a single byte. I will look into that this evening to see if I can resolve.

aykevl:
Looks like the code here has been merged already in #4014 .

PR #2631, opened 2022-02-12
samd21,samd51: fix usbcdc initialization when -serial=uart

sago35:
After writing with -serial=uart, tinygo flash will fail.
The reason for the failure is that the USBCDC initialization is not performed.
This PR will fix this problem.

I have confirmed that it works with the following targets:

* ~~nrf : feather-nrf52840-sense~~
* samd21 : qtpy
* samd51 : wioterminal

### DIFF
--- a/src/runtime/runtime_atsamd21.go
+++ b/src/runtime/runtime_atsamd21.go
@@ -7,7 +7,6 @@ import (
 	"device/arm"
 	"device/sam"
 	"machine"
-	"reflect"
 	"runtime/interrupt"
 	"runtime/volatile"
 	"unsafe"
@@ -31,7 +30,7 @@ func init() {
 
 	// connect to USB CDC interface
 	machine.Serial.Configure(machine.UARTConfig{})
-	if reflect.TypeOf(machine.Serial) != reflect.TypeOf(machine.USB) {
+	if !machine.USB.Configured() {
 		machine.USB.Configure(machine.UARTConfig{})
 	}
 }
--- a/src/runtime/runtime_atsamd21.go
+++ b/src/runtime/runtime_atsamd21.go
@@ -7,6 +7,7 @@ import (
 	"device/arm"
 	"device/sam"
 	"machine"
+	"reflect"
 	"runtime/interrupt"
 	"runtime/volatile"
 	"unsafe"
@@ -30,6 +31,9 @@ func init() {
 
 	// connect to USB CDC interface
 	machine.Serial.Configure(machine.UARTConfig{})
+	if reflect.TypeOf(machine.Serial) != reflect.TypeOf(machine.USB) {
+		machine.USB.Configure(machine.UARTConfig{})
+	}
 }
 
 func putchar(c byte) {
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -1742,6 +1742,7 @@ type USBCDC struct {
 	waitTxc           bool
 	waitTxcRetryCount uint8
 	sent              bool
+	configured        bool
 }
 
 var (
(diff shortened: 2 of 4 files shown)


sago35:
I see that machine.USB is not defined in some cases.
I wish there was a better way to categorize them.

```
/go/bin/tinygo build -size short -o test.hex -target=pca10040            examples/blinky1
# runtime
/home/circleci/project/src/runtime/runtime_nrf.go:33:62: USB not declared by package machine
/home/circleci/project/src/runtime/runtime_nrf.go:34:11: USB not declared by package machine
```

sago35:
Once nrf has been excluded from the target.
I think this will make the CI successful.

I think it is better to think about nrf separately.
So far I haven't come up with a good way to do that.

src/runtime/runtime_atsamd21.go
```diff
@@ -7,6 +7,7 @@ import (
 	"device/arm"
 	"device/sam"
 	"machine"
+	"reflect"
```
deadprogram: I do not think we want to require the "reflect" package here, it has to increase the minimum memory footprint, no?

bgould:
I believe implementing this feature in this way would sort of negate the purpose of -serial=uart ... I think the reason someone might use that feature would be to skip initializing the USB.  If the user is interested in having both the USB initialized for the reset functionality and have the default serial output to a UART, I they can already achieve that by using -serial=usb and replacing machine.Serial with their preferred UART similar to this:

https://github.com/bgould/keyboard-firmware/blob/76d63fd76e6ddbe16fc5cb02538d73e30dc89991/devices/kint41/main_usbhid.go#L21

sago35:
> I think the reason someone might use that feature would be to skip initializing the USB.

I hadn't thought about the reason being to skip the USB initialization.

..

The advantage of `-serial=uart` in my opinion is that it can immediately output println, etc. right after startup.
When `-serial=usb`, the serial output will disappear until the USB setup is finished.

In this kind of usage, there may be cases where you want to write with `-serial=uart` and then write with `-serial=uart` again.

I write from openocd, so I don't have trouble in many cases.
However, it felt like a very troubling behavior as I wrote the documentation.

sago35:
I think some people want to msd-flash even when -serial=uart.
For that person, USB-CDC is not really necessary, but he or she would like to have 1200bps-port-open.

sago35:
https://github.com/tinygo-org/tinygo/pull/1922

>     It makes it possible to disable the UART in cases where it is not needed or needs to be disabled to conserve power.
>     It makes it possible to disable the serial output to reduce code size, which may be important for some chips. Sometimes, a few kB can be saved this way.
>     It makes it possible to override the default, for example you might want to use an actual UART to debug the USB-CDC implementation.

In terms of code size, the current way of doing #2631 is not good.

src/runtime/runtime_atsamd21.go
sago35: Using reflect makes a big difference in ROM size, doesn't it?
I hadn't noticed that.

wioterminal:
| - | flash (byte) | ram (byte) |
| -- | --: | --: |
| dev (4b2edc9a26a84b4870d61ddf09f70369ab9f6f1e) | 5260 | 4276 |
| #2631 (reflect) | 13000 | 6500 |
| #2631 (w/o reflect) | 8160 | 6504 |

sago35: I fixed it with a new commit.
However, I'm wondering what I should do about @bgould's point.

deadprogram:
Seems to me that if we do not have this, users with UF2 bootloader can end up with device they cannot flash. Size will be bigger but it is at least avoiding an unexpected tricky problem for users.

deadprogram:
@sago35 do you want to squash these commits so I can merge, please?

sago35:
@deadprogram 
Please `Squash and merge`.

deadprogram:
Thanks @sago35 done.

PR #2632, opened 2022-02-12
debug: pyOCD support

ysoldak:
This adds [pyOCD](https://pyocd.io) support, rudimentary, uses existing openocd target definitions.

Tested to work with [picoprobe](https://www.raspberrypi.com/documentation/microcontrollers/raspberry-pi-pico.html#debugging-using-another-raspberry-pi-pico) debugging RP2040 board.
Download [picoprobe.uf2](https://datasheets.raspberrypi.com/soft/picoprobe.uf2) file and flash it to any RP2040 board to convert it to a debugger. 

Motivation:
This is the only straight way to debug RP2040 boards I was able to find.
`openocd` (0.11.0 or HEAD) would not work for me via `cmsis-dap` or `jlink` interfaces (and different debuggers, including Particle DAPLink, XIAO DAPLink and JLink EDU I've tried), while `picoprobe` interface is not available in official `openocd`, and `pyOCD` does have it.

To use `pyocd`, please specify `--programmer=pyocd` in `tinygo gdb` command.

`--persist` is there to keep pyOCD running when automatically launched gdb is detached by user to connect to pyOCD from another app, like an IDE.

### DIFF
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -370,7 +370,7 @@ func (c *Config) Programmer() (method, openocdInterface string) {
 	case "":
 		// No configuration supplied.
 		return c.Target.FlashMethod, c.Target.OpenOCDInterface
-	case "openocd", "msd", "command":
+	case "openocd", "pyocd", "msd", "command":
 		// The -programmer flag only specifies the flash method.
 		return c.Options.Programmer, c.Target.OpenOCDInterface
 	case "bmp":
--- a/main.go
+++ b/main.go
@@ -557,6 +557,20 @@ func Debug(debugger, pkgName string, ocdOutput bool, options *compileopts.Option
 				daemon.Stdout = w
 				daemon.Stderr = w
 			}
+		case "pyocd":
+			port = ":3333"
+			gdbCommands = append(gdbCommands, "monitor halt", "load", "monitor reset halt")
+			daemon = executeCommand(config.Options, "pyocd", "gdbserver", "--persist", "-t", config.Target.OpenOCDTarget)
+			if ocdOutput {
+				// Make it clear which output is from the daemon.
+				w := &ColorWriter{
+					Out:    colorable.NewColorableStderr(),
+					Prefix: "pyocd: ",
+					Color:  TermColorYellow,
+				}
+				daemon.Stdout = w
+				daemon.Stderr = w
+			}
 		case "jlink":
 			port = ":2331"
 			gdbCommands = append(gdbCommands, "load", "monitor reset halt")


ysoldak:
Picoprobe_wiring

``` 
Pico A GND -> Pico B GND
Pico A GP2 -> Pico B SWCLK
Pico A GP3 -> Pico B SWDIO
Pico A GP4/UART1 TX -> Pico B GP1/UART0 RX
Pico A GP5/UART1 RX -> Pico B GP0/UART0 TX
```

Extracted from [Getting Started with Pico](https://datasheets.raspberrypi.com/pico/getting-started-with-pico.pdf)

sago35:
I haven't looked at it in detail, but I was able to gdb tinygo with feather-nrf52840-sense via pyocd.
It would be nice if we could do tinygo flash as well.

sago35:
I think it is difficult to combine with openocd-target to make it work.
If the openocd configuration happens to match the pyocd name, it may work.

For nrf52840, the openocd-target setting of nrf51 is correct.
However, for pyocd, the correct setting is nrf52840.

I think it is a bit of a maintenance challenge to get these things to live together.
Do you have any good ideas?
Maybe you're not thinking about anything other than the rp2040.

ysoldak:
Thanks for looking, @sago35 
Yes, I am mostly concerned about rp2040, other targets I can debug already now.
As said, this is rudimentary support, i.e. minimal and PoC. I shall probably convert this PR to draft.

compileopts/config.go
```diff
@@ -370,7 +370,7 @@ func (c *Config) Programmer() (method, openocdInterface string) {
 	case "":
 		// No configuration supplied.
 		return c.Target.FlashMethod, c.Target.OpenOCDInterface
-	case "openocd", "msd", "command":
+	case "openocd", "pyocd", "msd", "command":
 		// The -programmer flag only specifies the flash method.
 		return c.Options.Programmer, c.Target.OpenOCDInterface
```
aykevl, line 375: Can you add a new setting here instead of reusing `OpenOCDInterface`?

main.go
```diff
@@ -557,6 +557,20 @@ func Debug(debugger, pkgName string, ocdOutput bool, options *compileopts.Option
 				daemon.Stdout = w
 				daemon.Stderr = w
 			}
+		case "pyocd":
+			port = ":3333"
+			gdbCommands = append(gdbCommands, "monitor halt", "load", "monitor reset halt")
+			daemon = executeCommand(config.Options, "pyocd", "gdbserver", "--persist", "-t", config.Target.OpenOCDTarget)
```
aykevl, line 563: Same here. Can you add a new `PyOCDTarget` here in addition to `OpenOCDTarget`? As @sago35 said, the openocd and pyocd interface names are sometimes different.

Also, I'm not sure about the `--persist` flag. It is not something that is officially supported at the moment and I can imagine it has surprising effects.

PR #2633, opened 2022-02-13
Cross-build of TinyGo to 32-bit Linux ARM

aykevl:
Instead of me doing it on my Raspberry Pi at home, let's do it in CI! It needs some configuration so that cross compilation works well.

Unfortunately, `llvm-config` is a bit complicated. I'm now running it using `qemu-arm`, which is not pretty, but it works.

Eventually, these changes might allow for us to cross compile from X86 Macs to M1 Macs - until the CI situation for macOS arm64 has been sorted out.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -84,7 +84,6 @@ jobs:
         run: make wasi-libc
       - name: Install fpm
         run: |
-          sudo apt-get install ruby ruby-dev
           sudo gem install --no-document fpm
       - name: Build TinyGo release
         run: |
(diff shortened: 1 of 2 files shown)


aykevl:
This is now ready for review.
I did some light testing of the resulting release files (.tar.gz and .deb) and they seem to run fine on my Raspberry Pi.

deadprogram:
Looks like you might need to `apt update` here: https://github.com/tinygo-org/tinygo/runs/5477614919?check_suite_focus=true#step:3:65

aykevl:
Rebased and trying again...

aykevl:
If we can get this in the coming release that would be great, then I don't have to make these ARM builds every time :)

deadprogram:
The build succeeded, and we certainly do need this!

Merging, great work on this @aykevl

PR #2634, opened 2022-02-13
all: move stm32 files to separate repository

aykevl:
I would like to move all autogenerated files in src/device/* to this repository. This is a first step.

There are multiple goals here:

  - This change reduces CI time by not regenerating all these files from scratch for every build. This is rather expensive and takes up a major part of CI time.
  - Go language tools often struggle with analyzing TinyGo source code because of all the special packages in GOROOT. This change is one step in the direction of making this easier for these tools. (They mainly struggle with the machine package, but I think moving the device/stm32 package is much lower risk to test such a change).
  - It should allow programs to override the tinygo.org/x/device module. I haven't tested this yet, it probably needs a modification to $GOROOT/src/vendor/modules.txt. This is not very use for this particular package, but could be very useful for the machine package.
  - _Maybe_ we'll ship these files separately some day, to be downloaded when needed. Not sure whether we'll ever need such a thing. But this new separation makes it a bit easier.

In the future, we might consider moving the machine package too. But that will likely be a breaking change and so not something we should do lightly.

@QuLogic FYI

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,13 +11,10 @@ src/device/sam/*.go
 src/device/sam/*.s
 src/device/sifive/*.go
 src/device/sifive/*.s
-src/device/stm32/*.go
-src/device/stm32/*.s
 src/device/kendryte/*.go
 src/device/kendryte/*.s
 src/device/rp/*.go
 src/device/rp/*.s
-vendor
 llvm-build
 llvm-project
 build/*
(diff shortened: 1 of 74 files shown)


aykevl:
@deadprogram TinyHCI seems to be having problems, any idea what's going on? (Maybe it was triggered by #2633 but somehow also affected this PR?)

sago35 (approved):
Looks good to me.
Putting the device under vendor is a good idea.

deadprogram:
TinyHCI ran out of disk space. Cleared up and restarted, everything passing now.

Totally agree with @sago35 `vendor` just sounds correct for this.

Now merging, thanks @aykevl

PR #2637, opened 2022-02-15
darwin: add support for arm64 GOARCH (aka Apple Silicon)

aykevl:
This patch adds support for generating GOOS=darwin GOARCH=arm64 binaries. This means that it will become possible to run `go test` on recent Macs, for example.

Untested, except that it compiles and produces a MachO ARM64 binary. If someone with a M1 Mac can test it, that would be great.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -635,6 +635,7 @@ endif
 	GOOS=linux GOARCH=arm $(TINYGO) build -size short -o test.elf       ./testdata/cgo
 	GOOS=windows GOARCH=amd64 $(TINYGO) build -size short -o test.exe   ./testdata/cgo
 	GOOS=darwin GOARCH=amd64 $(TINYGO) build              -o test       ./testdata/cgo
+	GOOS=darwin GOARCH=arm64 $(TINYGO) build              -o test       ./testdata/cgo
 ifneq ($(OS),Windows_NT)
 	# TODO: this does not yet work on Windows. Somehow, unused functions are
 	# not garbage collected.
--- a/src/internal/task/task_stack_arm64.S
+++ b/src/internal/task/task_stack_arm64.S
@@ -1,7 +1,12 @@
+#ifdef __MACH__
+.global  _tinygo_startTask
+_tinygo_startTask:
+#else
 .section .text.tinygo_startTask
 .global  tinygo_startTask
 .type    tinygo_startTask, %function
 tinygo_startTask:
+#endif
     .cfi_startproc
     // Small assembly stub for starting a goroutine. This is already run on the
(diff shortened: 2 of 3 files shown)


aykevl:
It appears that CI breaks because arm64 isn't supported yet in the new macho linker in LLVM 11 and LLVM 12. I think it would be reasonable to simply remove support for these LLVM versions.

deadprogram:
The default for us is now LLVM 13 and LLVM 14 is coming out soonish. I would agree with removal myself. 

It can be installed without package manager via BYOLLVM on older LTS distros if that is a problem for someone who wants to compile everything from sources including TinyGo itself.

arbourd:
> This patch adds support for generating GOOS=darwin GOARCH=arm64 binaries. This means that it will become possible to run `go test` on recent Macs, for example.
> 
> Untested, except that it compiles and produces a MachO ARM64 binary. If someone with a M1 Mac can test it, that would be great.

I can give it a shot re: tests. Other than building it, what's the best for it? Just running the CLI I guess?

aykevl:
Updated, hopefully the tests now pass.

> I can give it a shot re: tests. Other than building it, what's the best for it? Just running the CLI I guess?

Basically just do `tinygo run <somepackagepath>` and see if it works. You can also build a binary using `tinygo build -o testbinary <somepacakgename>`, run it to test, and check that it's arm64 using `file testbinary`.

dkegel-fastly:
So close!  Everything works but the $INODE64 suffixed syscalls:
```
% build/tinygo test os
tinygo:ld.lld: error: undefined symbol: _readdir_r$INODE64
>>> referenced by /var/folders/hw/6t6vc0192yz_29lh30b3b7hh0000gp/T/tinygo1353498533/main.o
```

I tried the obvious fix, 
https://github.com/dkegel-fastly/tinygo/commit/b77a9710dfcfbb2d2702833fc1cd151e6471bcbc
but the resulting binaries seem to abort very early in main, as they are branching to runtime.runMain()...?
(Copying libc_syscall_darwin_amd64.go to libc_syscall_darwin_arm64.go restores the previous
behavior, so the split into different files isn't the problem.)

aykevl:
Other tests build and run correctly on arm64?

> So close! Everything works but the $INODE64 suffixed syscalls:

Ah, yes, it seems like `$INODE64` isn't used on arm64. Sadly that means we have to use a patch like yours to support both systems.

dkegel-fastly:
It's not sad, really.  Upstream does the same kind of thing.

deadprogram:
This looks ready to me. Now merging!

PR #2638, opened 2022-02-16
all: add support for ThinLTO

aykevl:
ThinLTO optimizes across LLVM modules at link time. This means that
optimizations (such as inlining and const-propagation) are possible
between C and Go. This makes this change especially useful for CGo, but
not just for CGo. By doing some optimizations at link time, the linker
can discard some unused functions and this leads to a size reduction on
average. It does increase code size in some cases, but that's true for
most optimizations.

I've excluded a number of targets for now (wasm, avr, xtensa, windows,
macos). They can probably be supported with some more work, but that
should be done in separate PRs.

Overall, this change results in an average 3.24% size reduction over all
the tinygo.org/x/drivers smoke tests.

TODO: this commit runs part of the pass pipeline twice. We should set
the PrepareForThinLTO flag in the PassManagerBuilder for even further
reduced code size (0.7%) and improved compilation speed.

---

This PR took me a _long_ time to prepare. The change itself is pretty small, but it required a lot of investigation to avoid increases in code size.
Some day, I'd like to go further and compile every Go package in a separate LLVM module and rely on ThinLTO for optimizations across modules. That may result in a code size increase, but will likely be a huge improvement in compilation speed. As a preparation, this PR runs ThinLTO between C and Go files (and later possibly also including the C standard library).

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -105,6 +105,15 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 		defer os.RemoveAll(dir)
 	}
 
+	// Look up the build cache directory, which is used to speed up incremental
+	// builds.
+	cacheDir := goenv.Get("GOCACHE")
+	if cacheDir == "off" {
+		// Use temporary build directory instead, effectively disabling the
+		// build cache.
+		cacheDir = dir
+	}
+
 	// Check for a libc dependency.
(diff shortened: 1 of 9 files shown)


deadprogram:
I looked at this more carefully and it seems like a very good change, very well thought out. I then rebased against `dev` and made sure it passed all the tests, so now I will merge. Thanks for working on this @aykevl

PR #2639, opened 2022-02-16
all: add -work flag

aykevl:
This flag has the same behavior as in upstream Go. It can be useful while debugging certain issues.


dgryski (approved):
LGTM

deadprogram:
OK, now merging. Thanks @aykevl and @dgryski for review.

PR #2644, opened 2022-02-18
Refresh macos-minimal-sdk to fix regression; github: also run 'make tinygo-test' on mac

dkegel-fastly:
On my mac, "make tinygo-test" started complaining
```
ld.lld: error: undefined symbol: _readdir_r$INODE64
```
today.  Seems CI isn't running that test yet on Mac, so let's try adding it.

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -76,6 +76,8 @@ jobs:
         run: make test GOTESTFLAGS="-v -short"
       - name: Build TinyGo release tarball
         run: make release -j3
+      - name: Test stdlib packages
+        run: make tinygo-test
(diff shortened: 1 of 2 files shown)


aykevl:
@dkegel-fastly can you update the lib/macos-minimal-sdk submodule? It should pass now with https://github.com/aykevl/macos-minimal-sdk/issues/1 fixed.

deadprogram (approved):

deadprogram:
Thanks for fixing @dkegel-fastly now merging.

dkegel-fastly:
You could also mark the interface as experimental.
We're happy to adapt to a different way of getting the info later.

PR #2645, opened 2022-02-19
test: fix assertion for multiple packages

aykevl:
I see no reason why it isn't possible to run `tinygo test -c` with multiple packages. It'll just create multiple test outputs. I think the intended flag was the `-o` flag, which indeed doesn't make much sense with multiple packages.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1411,8 +1411,8 @@ func main() {
 		if len(pkgNames) == 0 {
 			pkgNames = []string{"."}
 		}
-		if *testCompileOnlyFlag && len(pkgNames) > 1 {
-			fmt.Println("cannot use -c flag with multiple packages")
+		if outpath != "" && len(pkgNames) > 1 {
+			fmt.Println("cannot use -o flag with multiple packages")
 			os.Exit(1)
 		}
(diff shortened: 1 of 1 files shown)


niaow:
IMO, this behavior is correct as-is. Upstream does not support this right now. If we do decide to support this, I think we should match whatever behavior upstream Go chooses (likely https://github.com/golang/go/issues/15513#issuecomment-959775908).

aykevl:
But it seems like we did support this in the past (at least, it's in my shell history), and I think it's reasonable behavior.
Right now:

* just `-c` produces exactly what you'd expect: it writes a bunch of test binaries to the current directory. Except that it's disallowed. I don't see why except that it's not something that upstream supports.
* `-o` does _not_ do something reasonable at the moment: it overwrites the given file name multiple times which is basically never what you'd want (and probably interferes badly with concurrency)

Hence why I made this PR. We can always add directory support to `-o` at a later time.

niaow (approved):

deadprogram:
Looks good thanks @aykevl and @niaow merging

PR #2647, opened 2022-02-22
src/os: make Environ() return a copy of the environment

### LINKED ISSUE #2646: os.Environ() returns live environment rather than copy
Tinygo's os tests include TestClearenv from upstream, but it fails frequently because it assumes os.Environ() returns a copy of the environment.
This is the case upstream, but in tinygo, it returns pointers to the live environment. Thus in tinygo, os.Environ() can go invalid the moment any environment variable is deleted.

We should probably disable TestClearenv until that's fixed, otherwise TestClearenv will continue to fail, causing spurious failures on unrelated changes in CI.

dgryski:
Fixes #2646

### DIFF
--- a/src/os/env.go
+++ b/src/os/env.go
@@ -8,6 +8,7 @@ package os
 
 import (
 	"internal/testlog"
+	"strings"
 	"syscall"
 )
 
@@ -137,5 +138,13 @@ func Clearenv() {
 // Environ returns a copy of strings representing the environment,
 // in the form "key=value".
 func Environ() []string {
-	return syscall.Environ()
+	orig := syscall.Environ()
+	single := strings.Join(orig, "")
+	env := make([]string, len(orig))
+	for i, v := range orig {
+		s := single[:len(v)]
+		env[i] = s
+		single = single[len(v):]
+	}
+	return env
 }
--- a/src/os/env.go
+++ b/src/os/env.go
@@ -8,7 +8,6 @@ package os
 
 import (
 	"internal/testlog"
-	"strings"
 	"syscall"
 )
 
@@ -138,13 +137,5 @@ func Clearenv() {
 // Environ returns a copy of strings representing the environment,
 // in the form "key=value".
 func Environ() []string {
-	orig := syscall.Environ()
-	single := strings.Join(orig, "")
-	env := make([]string, len(orig))
-	for i, v := range orig {
-		s := single[:len(v)]
-		env[i] = s
-		single = single[len(v):]
-	}
-	return env
+	return syscall.Environ()
 }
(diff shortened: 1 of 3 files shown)


dankegel:
this pull request should also bring back the test deleted by https://github.com/tinygo-org/tinygo/pull/2603

src/os/env.go
```diff
@@ -137,5 +138,13 @@ func Clearenv() {
 // Environ returns a copy of strings representing the environment,
 // in the form "key=value".
 func Environ() []string {
-	return syscall.Environ()
+	orig := syscall.Environ()
+	single := strings.Join(orig, "")
```
dkegel-fastly: Upsteam does the copy in syscall.Environ(), we might want to do the same in case someone calls that directly?

dgryski:
That probably makes sense.  I'll move this logic (or equivalent) there.

dkegel-fastly:
Would it make sense to do a single allocation to copy libc_environ and then point to bits of it?  That might be faster...?

dgryski:
Yeah, I was torn about doing that because we need to loop over the environment a few times instead of just once.  Let me try that approach and see what it looks like

dgryski:
Now with a single allocation.

dkegel-fastly (commented):
LGTM

src/syscall/syscall_libc.go
```diff
@@ -235,20 +235,41 @@ func Mprotect(b []byte, prot int) (err error) {
...
 		rawEnvVar.length = length
-		envs = append(envs, envVar)
-		// This is the Go equivalent of "environ++" in C.
+		// pull off the number of bytes we need for this environment variable
+		var bs []byte
+		bs, b = b[:length], b[length:]
+		// copy over the bytes to the Go heap
+		copy(bs, envVar)
+		// convert trimmed slice to string
+		s := *(*string)(unsafe.Pointer(&bs))
```
aykevl, line 269: This is a dirty trick... but I have to admit, this whole function isn't exactly clean and I wrote it originally.
(Dirty because it's not guaranteed that slice/string layout will stay the way it is now - although I don't see why it would need to change).

aykevl (approved):

aykevl:
> Now with a single allocation.

Note that this can have the side effect of keeping more alive than needed. Imagine you only need a single value out of the environment. So you call `Environ()`, pick the one value, and keep that string around. Now all the environment strings can't be garbage collected.
Maybe that's fine, just noting in case that wasn't the intention.

dgryski:
Good point about a single environment variable pinning the entire allocation. I think I'll add a comment to this function about that

dkegel-fastly:
Yeah, that went through my head too, I should have mentioned it.

PR #2654, opened 2022-02-25
Revert "all: move stm32 files to separate repository"

sago35:
This PR is to revert #2634.

Since the problem is reproduced in the local environment, I think it is better to revert and then PR again after the countermeasure is completed.
More information about the problem can be found below.

https://github.com/tinygo-org/drivers/pull/379

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,10 +11,13 @@ src/device/sam/*.go
 src/device/sam/*.s
 src/device/sifive/*.go
 src/device/sifive/*.s
+src/device/stm32/*.go
+src/device/stm32/*.s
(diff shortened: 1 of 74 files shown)


sago35:
Whenever someone does CI using tinygo/tinygo-dev, it always results in an error.
This is a very bad problem.

deadprogram:
Agreed. Reverting is the best option for now.

aykevl (commented):
Just wanted to say LGTM. I (or somebody else!) need to investigate how to implement this properly. I probably should have marked #2634 as draft to avoid this...

PR #2655, opened 2022-02-27
WIP: redesign interp

niaow:
This is an attempt to redesign interp. It is still relatively early on, but the main goals are to track side-effects more accurately and keep time complexity down.


niaow:
This is still a bit more complex than I want (especially with `errRuntime` vs reverts)

niaow:
There is one known problem right now: the memory management assumes that a store of a non-byte-multiple integer (mainly `i1`) is zero extended when storing and truncated when loading. This is not really correct, but should not actually be a problem. I don't really have a better idea for this.

aykevl:
> There is one known problem right now: the memory management assumes that a store of a non-byte-multiple integer (mainly `i1`) is zero extended when storing and truncated when loading. This is not really correct, but should not actually be a problem. I don't really have a better idea for this.

Why would that not be correct? All modern general-purpose ISAs are byte addressable, not bit addressable. LLVM backends lower `i1` to `i8`. I think this assumption is correct.

niaow:
From the LLVM [docs for store](https://llvm.org/docs/LangRef.html#store-instruction):
> When writing a value of a type like i20 with a size that is not an integral number of bytes, it is unspecified what happens to the extra bits that do not belong to the type, but they will typically be overwritten.

If I understand what this is saying, this is unspecified behavior. However, yes I expect that this is at least generally correct.

niaow:
Now most of the base tests pass. The main exceptions are:
- The `reflect.go` and `json.go` tests fail (this failure can be replicated by completely turning interp off, so it does not seem to be a bug in interp)
- The WASM target doesn't really work right now because of bugs in ExternalInt64AsPtr (wasi is fine)
- `env.go`, `stdlib.go`, and `testing.go` fail on emulated Cortex-M3 and RISC-V

niaow:
The `env.go` issue is caused by `internal/testlog.logger` is getting corrupted. So far it is difficult to track down why.

niaow:
Well I found the problem: it was a stack overflow in `unicode.init`, since it was not interpreted due to the allocas.

aykevl:
> The `reflect.go` and `json.go` tests fail (this failure can be replicated by completely turning interp off, so it does not seem to be a bug in interp)

This is probably because those rely on `OptimizeReflectImplements`, which in turn relies on interp.
...yes, that's a bit fragile.

niaow:
The remaining issue with `env.go` now is that maps and other special data structures do not have attached type information (which causes the allocation to be left on the heap). I think that we do need to pass type information to these through the constructors, which implies getting the frontend to emit the compound type info.

niaow:
With the latest commit, the `reflect` and `json` tests pass.

FrankReh (commented):
Given my confusion regarding pointerAlign vs pointerSize, the first parameter name to createObjectLayout1 is then a misnomer. Maybe both should be passed to createObjectLayout1 or maybe createObjectLayout1 should should compute those two values again on its own.

compiler/llvm.go
```diff
@@ -102,6 +101,70 @@ func (c *compilerContext) createObjectLayout(t llvm.Type, pos token.Pos) llvm.Va
 		layout := (uint64(1) << 1) | 1
 		return llvm.ConstIntToPtr(llvm.ConstInt(c.uintptrType, layout, false), c.i8ptrType)
 	}
+	return c.createObjectLayout1(pointerSize, objectSizeBytes, bitmap, pos)
```
FrankReh: My version passed a pointerSize. But with the change to line 163, this should also be the alignment. Maybe better to pass both or pass neither and just recompute them in createObjectLayout1.

compiler/llvm.go
```diff
@@ -102,6 +101,70 @@ func (c *compilerContext) createObjectLayout(t llvm.Type, pos token.Pos) llvm.Va
...
+		eSizeBytes, ebitmap,
+	)
+
+	// We know the bucket will be large enough to contain a pointer
+	// so no check for that.
+
+	// We know there are pointers, we added one ourself for the bucket overhead
+	// so no check for that.
+
+	return c.createObjectLayout1(pointerAlign, mSizeBytes, mbitmap, pos)
```
FrankReh: This shouldn't pass the alignment unless the earlier call does too.

compiler/llvm.go
```diff
@@ -102,6 +101,70 @@ func (c *compilerContext) createObjectLayout(t llvm.Type, pos token.Pos) llvm.Va
...
+}
+
+func (c *compilerContext) createObjectLayout1(pointerSize, objectSizeBytes uint64, bitmap *big.Int, pos token.Pos) llvm.Value {
+	pointerAlignment := c.targetData.PrefTypeAlignment(c.i8ptrType)
```
FrankReh: This is one of the causes of confusion. I got the alignment here, without passing it from the caller.

niaow:
I am not really sure how I feel about this. I have gotten a lot working so far, but this is just way too big and complicated right now.

niaow:
Some things that went right with this attempt:
1. The fixed-shape tree used for memory works quite well and has predictable overhead
2. The bit cat and slice mechanisms simplify memory handling a lot
3. Side-effect calculation is fairly accurate
4. The structure of each instruction using offsets into a list to get inputs and then appending its outputs to that list works well
5. The structure of the constructed representation works well, and the depth-first traversal of the dominator tree simplifies the construction of the internal interp representation
6. It is relatively easy to debug behavior
7. The "resolve" process on arguments is effective and flexible

Some things that went wrong with this attempt:
1. The whole distinction between `errRuntime` and `errRevert` is very confusing
2. There are still a few cases where repeated reverts can result in `O(bad)` overhead with repeated restoration of memory state
3. Memory lifetimes get stretched when they are used at runtime
4. Structs in interfaces are quite heavy, so functions use more memory than they really need to
5. The lack of access to the bodies of functions from imported packages (e.g. `errors.New`) forces extra work to be done after linking
8. The process of implementing an instruction is excessively complicated
9. The bit-concatenation type is not properly comparable
10. Its just too big, although this might be necessary
11. The "resolve" process on arguments is unnecessarily slow

PR #2664, opened 2022-03-02
Implement runtime.fastrand and runtime.memhash

aykevl:
This is needed for the hash/maphash package.

Unfortunately, I couldn't get the hash/maphash package to pass tests because it's too slow. I think it hits the quadratic complexity of the current map implementation. So I'm not quite sure how to test this. Maybe we should wait with this PR until the quadratic time complexity of the map implementation has been fixed?

### DIFF
--- a/src/runtime/algorithm.go
+++ b/src/runtime/algorithm.go
@@ -3,6 +3,8 @@ package runtime
 // This file implements various core algorithms used in the runtime package and
 // standard library.
 
+import "unsafe"
+
 // This function is used by hash/maphash.
 func fastrand() uint32 {
 	xorshift32State = xorshift32(xorshift32State)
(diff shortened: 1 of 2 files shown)


src/runtime/algorithm.go
```diff
@@ -0,0 +1,55 @@
...
+	xorshift32State = xorshift32(xorshift32State)
+	return xorshift32State
+}
+
+var xorshift32State uint32 = 1
+
+func xorshift32(x uint32) uint32 {
+	// Algorithm "xor" from p. 4 of Marsaglia, "Xorshift RNGs".
+	// This could probably be improved if needed.
+	x ^= x << 13
```
dgryski: The constants `7,1, 9` have *slightly* better statistical properties, but not by much: https://github.com/dgryski/trifles/blob/master/inthash/inthash.go#L153

src/runtime/algorithm.go
```diff
@@ -0,0 +1,55 @@
...
+func memhash(p unsafe.Pointer, seed, s uintptr) uintptr {
+	if unsafe.Sizeof(uintptr(0)) > 4 {
+		return seed ^ uintptr(hash64(p, s))
+	}
+	return seed ^ uintptr(hash32(p, s))
+}
+
+// Get FNV-1a hash of the given memory buffer.
+//
+// https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function#FNV-1a_hash
```
dgryski, line 36: I should lookup the state-of-the-art for embedded hash functions.  (I dislike fnv, but that's just my personal bias...). Also, I know from experiments that the issue with the maps isn't the hashing, it's the bucket chains.

dgryski (approved):
LGTM.  Nice!  Do we also need a `fastrandn` implementation?

src/runtime/algorithm.go
aykevl: This is not new code, I'm only moving the code to a shared file.

dgryski: Yeah, this was more areas of research for my own interests if it becomes a bottleneck.

src/runtime/algorithm.go
aykevl: Thanks! I have updated the code. The (7, 1, 9) constants are also better on AVR: they're much easier to shift than (13, 17, 5) because AVR doesn't have a barrel shifter.

Note sure how to best cite the algorithm, suggestions welcome :)

dgryski (approved):
LGTM

deadprogram:
:slot_machine: now merging. Thanks @aykevl and @dgryski

Issue #1941, opened 2021-06-07
wasm: js.CopyBytesToGo does not copy bytes to Go


yurivish:
I encountered a bug when trying to pass data from JS to WASM using `js.CopyBytesToGo` to copy over the contents of a `Uint8ClampedArray`.

`js.CopyBytesToGo` errors with 

>panic: syscall/js: CopyBytesToGo: expected src to be an Uint8Array or Uint8ClampedArray

 despite the fact that `src` is a `Uint8ClampedArray`. 

Compiling the same code with Go's built-in WASM support succeeded, leading me to think this is a bug in TinyGo.

Here's a minimal example that reproduces the issue. It instantiates a `Uint8ClampedArray` and attemps to copy its contents into a Go slice, resulting in the same error:

main.go:
```go
package main

import (
	"fmt"
	"syscall/js"
)

func main() {
	size := 100
	src := js.Global().Get("Uint8ClampedArray").New(size)
	dst := make([]byte, size)
	fmt.Println("copied: ", js.CopyBytesToGo(dst, src))
}
```

index.html:
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8" />
    <title>Hello World - Go</title>
  </head>
  <body>
    <script src="./wasm_exec.js"></script>
    <script type="module" src="./index.js"></script>
  </body>
</html>
```

Build `tinygo_bug.wasm` with `tinygo build -o tinygo_bug.wasm -target wasm ./tinygo_bug.go`

Place the above into a folder with [wasm_exec.js](https://github.com/tinygo-org/tinygo/blob/164632616496e21f5f11098316c650ab5ad9356c/targets/wasm_exec.js) and the index.js file from the Gist below, then run a local web server to test; the error should appear in the console. I've confirmed that this code runs successfully when compiled with Go and prints `copied: 100`.

Here's a gist with the above files for convenience: https://gist.github.com/yurivish/d2490b3040c5dbc7426840e0e476513e

This report is on macOS Catalina (10.15.7 ) with TinyGo version 0.18.0 and Go version go1.16 darwin/amd64.

ghost:
I just tested it here, and see the same panic.  Works with "Uint8Array".

tinygo version 0.15.0 darwin/amd64 (using go version go1.15.3 and LLVM version 10.0.1) macos 10.12.4

PR #2671, opened 2022-03-03
syscall/js: allow copyBytesTo(Go|JS) to use Uint8ClampedArray

samuel:
This change allow the `js.CopyBytesToX` functions to also work with `Uint8ClampedArray`. The same change was made to Go: https://github.com/golang/go/commit/f0e8b81aa34120e21642c569912bde00ccd33393

I've tested that this change alone is enough to make it work. These types of arrays are common when working with the HTML Canvas.

Fixes https://github.com/tinygo-org/tinygo/issues/1941

### DIFF
--- a/targets/wasm_exec.js
+++ b/targets/wasm_exec.js
@@ -424,7 +424,7 @@
 
 						const dst = loadSlice(dest_addr, dest_len);
 						const src = loadValue(source_addr);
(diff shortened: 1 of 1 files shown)


aykevl (approved):
Looks good to me! Running CI now.

deadprogram:
Thank you @samuel for the contribution. Now merging.

PR #2673, opened 2022-03-03
src/runtime: use memzero for leaking collector

dgryski:


aykevl (approved):
This looks good to me.
Did you check whether this has any concrete benefits, like smaller binary size? (Not necessary to test IMHO but would be nice to see).

dgryski:
I didn't do any checks other than making sure a few test programs still compile and run.  Do you have a nicer one-liner to figure out binary size changes?

aykevl:
I guess just `tinygo build -size=short ...`. Or if you want to get fancy: https://gist.github.com/aykevl/bc593ce06b22d64d8dcbb335a40a1e40.
I often do this myself for things that are supposed to improve code size to see whether they really do.

dgryski:
Hmm.. Should add those scripts to the main tinygo repo?  They seem useful.

PR #2675, opened 2022-03-03
gc_leaking.go: don't inline alloc.  Fixes #2674.

### LINKED ISSUE #2674: -gc=leaking causes significant code bloat?
A large app, built with default gc, yields a 5.0MB .wasm file.
With -gc=leaking, the .wasm file is 7.3MB (and seems 2x faster).

At @dgryski and @cfallin's suggestion, I added //go:noinline on func alloc() in gc_leaking.go.
This brought the size down to 4.7MB, reducing its size by 30%.
First quick benchmark seem to show It also slowed down slightly, but was still much faster than without -gc=leaking.

dkegel-fastly:
Prevent a surprising amount of code bloat from inlining alloc() when using -gc=leaking.

### DIFF
--- a/src/runtime/gc_leaking.go
+++ b/src/runtime/gc_leaking.go
@@ -14,6 +14,9 @@ import (
 // Ever-incrementing pointer: no memory is freed.
 var heapptr = heapStart
 
+// Inlining alloc() speeds things up slightly but bloats the executable by 50%,
+// see https://github.com/tinygo-org/tinygo/issues/2674.  So don't.
+//go:noinline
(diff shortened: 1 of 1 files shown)


dgryski (approved):
LGTM.  @aykevl any concerns about merging this?

deadprogram:
Seems good to everyone and only impacts one GC, so merging.

deadprogram:
Seems good to everyone and only impacts one GC, so merging.

aykevl:
Just a bit curious why it behaves the way it does, but if it speeds things up then :shrug: 
I don't think `runtime.alloc` should be inlined in general.

PR #2676, opened 2022-03-04
src/runtime: implement runtime.Version() function

ZauberNerd:
This adds the `Version()` function of the `runtime` package which embeds
the go version that was used to build tinygo.

For programs that are compiled with tinygo the version can be overriden
via the:
`tinygo build -ldflags="-X 'runtime.buildVersion=abc'"` flag.
Otherwise it will continue to use the go version with which tinygo was
compiled.

### DIFF
--- a/src/runtime/extern.go
+++ b/src/runtime/extern.go
@@ -3,3 +3,15 @@ package runtime
 func Callers(skip int, pc []uintptr) int {
 	return 0
 }
+
+// buildVersion is the Tinygo tree's version string at build time.
+//
+// This is set by the linker.
+var buildVersion string
+
+// Version returns the Tinygo tree's version string.
+// It is the same as goenv.Version, or in case of a development build,
+// it will be the concatenation of goenv.Version and the git commit hash.
+func Version() string {
+	return buildVersion
+}
--- a/Makefile
+++ b/Makefile
@@ -208,7 +208,7 @@ lib/wasi-libc/sysroot/lib/wasm32-wasi/libc.a:
 # Build the Go compiler.
 tinygo:
 	@if [ ! -f "$(LLVM_BUILDDIR)/bin/llvm-config" ]; then echo "Fetch and build LLVM first by running:"; echo "  make llvm-source"; echo "  make $(LLVM_BUILDDIR)"; exit 1; fi
-	CGO_CPPFLAGS="$(CGO_CPPFLAGS)" CGO_CXXFLAGS="$(CGO_CXXFLAGS)" CGO_LDFLAGS="$(CGO_LDFLAGS)" $(GO) build -buildmode exe -o build/tinygo$(EXE) -tags byollvm -ldflags="-X main.gitSha1=`git rev-parse --short HEAD`" .
+	CGO_CPPFLAGS="$(CGO_CPPFLAGS)" CGO_CXXFLAGS="$(CGO_CXXFLAGS)" CGO_LDFLAGS="$(CGO_LDFLAGS)" $(GO) build -buildmode exe -o build/tinygo$(EXE) -tags byollvm -ldflags="-X github.com/tinygo-org/tinygo/goenv.GitSha1=`git rev-parse --short HEAD`" .
 
 test: wasi-libc
 	CGO_CPPFLAGS="$(CGO_CPPFLAGS)" CGO_CXXFLAGS="$(CGO_CXXFLAGS)" CGO_LDFLAGS="$(CGO_LDFLAGS)" $(GO) test $(GOTESTFLAGS) -timeout=20m -buildmode exe -tags byollvm ./builder ./cgo ./compileopts ./compiler ./interp ./transform .
--- a/builder/build.go
+++ b/builder/build.go
@@ -204,6 +204,21 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 	var packageJobs []*compileJob
 	packageBitcodePaths := make(map[string]string)
 	packageActionIDs := make(map[string]string)
+
+	if config.Options.GlobalValues["runtime"]["buildVersion"] == "" {
+		version := goenv.Version
+		if strings.HasSuffix(goenv.Version, "-dev") && goenv.GitSha1 != "" {
+			version += "-" + goenv.GitSha1
+		}
+		if config.Options.GlobalValues == nil {
+			config.Options.GlobalValues = make(map[string]map[string]string)
+		}
+		if config.Options.GlobalValues["runtime"] == nil {
+			config.Options.GlobalValues["runtime"] = make(map[string]string)
+		}
+		config.Options.GlobalValues["runtime"]["buildVersion"] = version
+	}
+
 	for _, pkg := range lprogram.Sorted() {
 		pkg := pkg // necessary to avoid a race condition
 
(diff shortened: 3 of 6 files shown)


src/runtime/extern.go
```diff
@@ -3,3 +3,10 @@ package runtime
 func Callers(skip int, pc []uintptr) int {
 	return 0
 }
+
+// Version returns the Go tree's version string.
+// It is either the commit hash and date at the time of the build or,
+// when possible, a release tag like "go1.3".
+func Version() string {
+	return "TODO: not implemented"
```
aykevl: This is actually something that's not too difficult to support. The compiler already supports the `-X` flag and we could add the version relatively easily. See these lines for a starting point:

https://github.com/tinygo-org/tinygo/blob/3883550c446fc8b14115366284316290f3d3c4b1/builder/build.go#L846-L850

We could for example add `runtime.buildVersion` to `GlobalValues` earlier in the build. The version itself can be obtained using `goenv.GorootVersionString`.

ZauberNerd: I've updated the PR to your suggestion.
Now the tinygo runtime.Version() is always the version of golang that was used to build tinygo, but not the version of tinygo itself, which I think would be more useful?

Also, for some reason I wasn't able to set the version via `go build -tags=llvm11 -ldflags="-X 'runtime.buildVersion=abc'" -o build/tinygo` - shouldn't that work too, so that tinygo could set its own version during build?

deadprogram: > Now the tinygo runtime.Version() is always the version of golang that was used to build tinygo, but not the version of tinygo itself, which I think would be more useful?

I agree that we should return the version of TinyGo used to build the binary here, which is what the main Go seems to do with respect to the version of Go used to build binary.

ZauberNerd: I've moved and exported `main.gitSha1` to `goenv.GitSha1` to be able to access it from both the `main` and `builder` package.
Now the `runtime.Version()` function returns a version string similar to what `tinygo version` returns (`0.23.0` - or, if it ends in `-dev` it will be the concatenation of the version and git sha `0.23.0-dev-f5170c0b`).

https://github.com/tinygo-org/tinygo/pull/2676/files#diff-d871f892c3b7d415467578ac1742cccab676ec4cc21022b0b9b34f5741237eb7R208-R217

ZauberNerd:
~~I get a few errors here, most are related to WASI, but one is:~~
```
tinygo:ld.lld: error: undefined symbol: notARealFunction
>>> referenced by builderr.go:9 (/home/ubuntu/tinygo/tests/testing/builderr/builderr.go:9)
>>>               /home/ubuntu/.cache/tinygo/thinlto/llvmcache-0A6BB726F2D94B251D948A1883CD488484B0EB09:(github.com/tinygo-org/tinygo/tests/testing/builderr_test.TestThing)
failed to run tool: ld.lld
```
~~I found: https://github.com/tinygo-org/tinygo/pull/2708 and the linked issues, but I have rebased and the error still persists.~~

**Edit:** I found `--- PASS: TestTest/Host/Fail (14.24s)` in the logs, so it seems the above error is just printed, but the test seems to pass.

The other ones seem to come from wasmtime:
* https://github.com/bytecodealliance/wasmtime/issues/2587
* https://docs.rs/wasmtime/latest/src/wasmtime/linker.rs.html#810

And they look like:
```
--- FAIL: TestTest (0.00s)
    --- FAIL: TestTest/WASI (0.00s)
        --- FAIL: TestTest/WASI/Pass (32.81s)
            main_test.go:611: Error: failed to run main module `/tmp/tinygo2920845472/main`
            main_test.go:611: 
            main_test.go:611: Caused by:
            main_test.go:611:     0: failed to instantiate "/tmp/tinygo2920845472/main"
            main_test.go:611:     1: command export 'runtime.buildVersion' is not a function
            main_test.go:611: FAIL      github.com/tinygo-org/tinygo/tests/testing/pass 4.192s
            main_test.go:523: test failed
FAIL
```

Apparently there's a switch to allow unknown exports: https://github.com/bytecodealliance/wasmtime/pull/2879, but I want to first understand why and how this error came to be before enabling any further options.
Or maybe one of you has seen this error already / has an idea what is happening?

ZauberNerd:
Adding the `--allow-unknown-exports` to the `args` here:
https://github.com/tinygo-org/tinygo/blob/3883550c446fc8b14115366284316290f3d3c4b1/main.go#L276-L305
does fix the issue, but I don't understand why it is necessary.

And when I just build the test file `tinygo test -target wasi -c os`, it will generate `os.test` which, when converted to WAT, shows that `runtime.buildVersion` is mentioned in the exports:
```
  (table $T0 58 58 funcref)
  (memory $memory 3)
  (global $g0 (mut i32) (i32.const 65536))
  (global $runtime.buildVersion i32 (i32.const 92376))
  (global $g2 (mut i32) (i32.const 0))
  (global $g3 (mut i32) (i32.const 0))
  (export "memory" (memory 0))
  (export "malloc" (func $malloc))
  (export "free" (func $free))
  (export "calloc" (func $calloc))
  (export "realloc" (func $realloc))
  (export "_start" (func $_start))
  (export "runtime.buildVersion" (global 1))
  (export "asyncify_start_unwind" (func $asyncify_start_unwind))
  (export "asyncify_stop_unwind" (func $asyncify_stop_unwind))
  (export "asyncify_start_rewind" (func $asyncify_start_rewind))
  (export "asyncify_stop_rewind" (func $asyncify_stop_rewind))
  (export "asyncify_get_state" (func $asyncify_get_state))
```
Which I assume has somehow to do with the variable being set by the linker?
Ideally, I wouldn't want the export showing up here instead of setting the `--allow-unknown-exports` flag.

ZauberNerd:
Ok, so in the linked issue the proposed fix is to remove `--export-dynamic` linker flag from Rust when targeting `wasm32-wasi`: https://github.com/rust-lang/rust/pull/81255
We can do the same by removing `--export-dynamic` in the `targets/wasi.json` file: https://github.com/tinygo-org/tinygo/blob/3883550c446fc8b14115366284316290f3d3c4b1/targets/wasi.json#L14

ZauberNerd:
Compiling a minimal program:
```golang
package main

import "runtime"

func main() {
    println(runtime.Version())
}
```
to WASI `tinygo build -o main.wasm -target wasi ./main.go` and running it with wasmtime: `wasmtime ./main.wasm`  works fine when removing the `--export-dynamic` linker flag.
With the `--export-dynamic` flag set, the program crashes when running it with wasmtime `wasmtime ./main.wasm` but works when running it with: `wasmtime --allow-unknown-exports ./main.wasm`.

And looking at the resulting WAT files in both cases it seems the only change is whether the `runtime.buildVersion` exists in the exports or not.

ZauberNerd:
~~Now there's only the `env.go` tests failing:~~
~~It seems like these tests embed the `os.Environ()` and `os.Args` in the binary via the linker~~

**Edit:** Nevermind, seems like I overwrote `config.Options.GlobalValues["runtime"]` before.

deadprogram (approved):

aykevl:
Thank you! This looks good to me.

I hope that `--export-dynamic` doesn't change anything important.

aykevl:
Note that programs using `runtime.Version()` _might_ expect the Go version string (from the currently used GOROOT), instead of the TinyGo version. That's hard to know, so let's keep it as-is.

PR #2677, opened 2022-03-04
src/runtime: add stub for debug.ReadBuildInfo()

ZauberNerd:
This adds the necessary structs and the `ReadBuildInfo()` function to
the runtime/debug module to allow to compile code that tries to access
it.
The stub itself returns ok=false, so that calling code should not try
to read from the nil pointer that is returned instead of the actual
BuildInfo struct.

### DIFF
--- a/src/runtime/debug/debug.go
+++ b/src/runtime/debug/debug.go
@@ -15,3 +15,28 @@ func SetMaxStack(n int) int {
 func Stack() []byte {
 	return nil
 }
+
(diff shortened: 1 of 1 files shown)


dgryski (approved):
LGTM

aykevl (approved):
Waiting for CI to finish...

deadprogram:
Thank you @ZauberNerd now merging.

PR #2678, opened 2022-03-04
src/runtime: add stub for runtime.NumCPU()

ZauberNerd:
This adds a stub for the `NumCPU()` function from the `runtime`
package.
This change allows code to compile that tries to access this function.

I guess for most hardware boards and WASM setting this value to `1` is
fine. And as far as I can see it shouldn't break or change existing
applications, because the function previously did not exist at all.

### DIFF
--- a/src/runtime/debug.go
+++ b/src/runtime/debug.go
@@ -1,0 +1,10 @@
+package runtime
+
+// NumCPU returns the number of logical CPUs usable by the current process.
+//
+// The set of available CPUs is checked by querying the operating system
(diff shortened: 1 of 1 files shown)


dgryski (approved):
LGTM

aykevl (approved):
Looks good to me. The value 1 should be fine for most purposes, especially since we don't support multithreading at the moment. Once we do, we should probably return a real value here.

deadprogram:
Now merging, thank you @ZauberNerd

PR #2679, opened 2022-03-04
src/os: add stubs for exec.ExitError and ProcessState.ExitCode

ZauberNerd:
This commit adds stubs for the `ExitError` struct of the `exec` package
and `ProcessState.ExitCode()` of the `os` package.

Since the `os/exec` is listed as unsupported, stubbing these methods
and structs should be enough to get programs that use these to compile.

### DIFF
--- a/src/os/exec.go
+++ b/src/os/exec.go
@@ -34,6 +34,12 @@ func (p *ProcessState) Success() bool {
 	return false // TODO
 }
 
+// ExitCode returns the exit code of the exited process, or -1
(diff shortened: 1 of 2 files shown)


dkegel-fastly (approved):
ExitCode isn't provided on Windows upstream, but that's probably ok.

deadprogram:
Thanks @ZauberNerd for the contribution and to @dkegel-fastly for reviewing.

Now merging.

PR #2680, opened 2022-03-04
src/os: implement os.File.Fd() method

ZauberNerd:
This commits adds a OS-specific `Fd()` function for `file_anyos.go` and
`file_other.go`, which returns a uintptr to the underlying file
descriptor.

Creating this as a draft PR, because I'm unsure, if this is the correct way to go about implementing it.

### DIFF
--- a/src/os/file.go
+++ b/src/os/file.go
@@ -179,9 +179,19 @@ func (f *File) SyscallConn() (syscall.RawConn, error) {
 	return nil, ErrNotImplemented
 }
 
+// fd is an internal interface that is used to try a type assertion in order to
+// call the Fd() method of the underlying file handle if it is implemented.
+type fd interface {
+	Fd() uintptr
+}
+
 // Fd returns the file handle referencing the open file.
 func (f *File) Fd() uintptr {
-	panic("unimplemented: os.file.Fd()")
+	handle, ok := f.handle.(fd)
+	if ok {
+		return handle.Fd()
+	}
+	return 0
 }
 
 // Truncate is a stub, not yet implemented
--- a/src/os/file_anyos.go
+++ b/src/os/file_anyos.go
@@ -118,6 +118,10 @@ func (f unixFileHandle) Close() error {
 	return handleSyscallError(syscall.Close(syscallFd(f)))
 }
 
+func (f unixFileHandle) Fd() uintptr {
+	return uintptr(f)
+}
+
 // Chmod changes the mode of the named file to mode.
 // If the file is a symbolic link, it changes the mode of the link's target.
 // If there is an error, it will be of type *PathError.
--- a/src/os/file_anyos_test.go
+++ b/src/os/file_anyos_test.go
@@ -68,3 +68,21 @@ func TestChdir(t *testing.T) {
 		t.Errorf("Remove %s: %s", dir, err)
 	}
 }
+
+func TestStandardFd(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Log("TODO: TestFd fails on Windows, skipping")
+		return
+	}
+	if fd := Stdin.Fd(); fd != 0 {
+		t.Errorf("Stdin.Fd() = %d, want 0", fd)
+	}
+
+	if fd := Stdout.Fd(); fd != 1 {
+		t.Errorf("Stdout.Fd() = %d, want 1", fd)
+	}
+
+	if fd := Stderr.Fd(); fd != 2 {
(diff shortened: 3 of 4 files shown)


dgryski:
Thanks.  Can you add a test or two?

dkegel-fastly:
It looks ok.  At least a trivial test would be nice.  
(Upstream barely tests it at all, using it to force blocking mode (!) and as a probe for an assertion in https://go.dev/src/os/readfrom_linux_test.go.  If we pass that test I'd be impressed.)

Every time we fatten the FileHandle interface, it diverges from https://pkg.go.dev/tinygo.org/x/tinyfs#FileHandle, but I don't know whether that breaks users of tinyfs somehow.

dgryski:
Should we be having PRs that update tinyfs at all?

/cc @aykevl

ZauberNerd:
> It looks ok. At least a trivial test would be nice. (Upstream barely tests it at all, using it to force blocking mode (!) and as a probe for an assertion in https://go.dev/src/os/readfrom_linux_test.go. If we pass that test I'd be impressed.)

I'm thinking maybe it's possible to `os.Open("/proc/self/<something>")` to read the file descriptor of the just opened file?
I know about `/proc/self/fd/*`, but for that I would need to know the descriptor already.

ZauberNerd:
Otherwise I can at least test that `Stdin.Fd() == 0`, `Stdout.Fd() == 1` etc.

dkegel-fastly:
I... guess you could pass the fd to NewFile and make sure you can use that file?

(cf. upstream's TestNewFileBlock(), which passes it an fd from a syscall and makes sure it can do file.Read().)

dkegel-fastly:
Offhand I would avoid /proc for this test.

ZauberNerd:
@dkegel-fastly  I'm not sure, if I understood you correctly, but I tried to write some meaningful tests - let me know if there's anything I should change or if I can mark this PR as ready for review.

dkegel-fastly:
Looks like it's unhappy on windows; if you have wine, you can reproduce it without windows, e.g.
```
$ GOOS=windows tinygo test -c os
$ wine os.test -test.v -test.run TestFd
=== RUN   TestFd
--- FAIL: TestFd (0.00s)
    Failed to read 5 bytes from file descriptor: readat as-fd: operation not implemented
    File descriptor contents not equal to file contents.
FAIL
```
This is not surprising, as fds are very different on windows, and upstream's "pollfd" layer that hides the difference (I think) has not been ported yet.

So just skip that test on windows, as TestMkdirTempBadDir() does.

deadprogram:
Please rebase against the latest dev, so that the Windows tests can run correctly. Thanks!

dkegel-fastly (approved):
LGTM

src/os/filesystem.go
```diff
@@ -57,6 +57,8 @@ type FileHandle interface {
 
 	// Close closes the file, making it unusable for further writes.
 	Close() (err error)
+
+	Fd() uintptr
```
aykevl: Instead of adding a new method to this interface (which breaks packages like tinyfs!), I would suggest type-asserting in the `Fd()` function.

src/os/file.go
```diff
@@ -181,7 +181,7 @@ func (f *File) SyscallConn() (syscall.RawConn, error) {
 
 // Fd returns the file handle referencing the open file.
 func (f *File) Fd() uintptr {
-	panic("unimplemented: os.file.Fd()")
+	return f.handle.Fd()
```
aykevl: Here is an example:

```suggestion
	handle, ok := f.handle.(interface {
		Fd() uintptr
	})
	if ok {
		return handle.Fd()
	}
	return 0
```

(This inline interface is perhaps a bit ugly though)

src/os/filesystem.go
dkegel-fastly: tinyfs was probably broken when we added the Seek() and ReadAt() methods to FileHandle, but yeah, avoiding further divergence would probably be good.

dkegel-fastly: See also https://github.com/tinygo-org/tinygo/issues/2709

dkegel-fastly: Tinyfs itself doesn't care about the discrepancy, its tests still pass because it has its own copy of File and FileHandle, and doesn't use tinygo's...?

ZauberNerd: Updated to the suggestion from @aykevl.

deadprogram:
Now merging. Thanks @ZauberNerd

PR #2681, opened 2022-03-04
src/os: export correct values for os.DevNull for each OS

ZauberNerd:
This commit introduces `os.DevNull` exports for all supported OS
targets.
I had to introduce a new file `file_wasi.go` in order to export a
different value for `file_windows.go` (`NUL`) and keep `/dev/null` for
all other targets.

I'll create this PR as draft, because I have a few open questions:
~~I'm not sure, if this works correctly with the build tags or if the `file_anyos.go` takes precedence over the `file_windows.go`.~~
Other options that I see:
1. export from `file_other.go`, `file_unix.go` and `file_windows.go`, ignoring WASI
  1.1 create a separate `file_wasi.go` to be able to export `DevNull` there too (seems overkill?)
2. export only from `file_anyos.go` and `file_other.go`, ignoring Windows

### DIFF
--- a/src/os/file_unix.go
+++ b/src/os/file_unix.go
@@ -1,6 +1,9 @@
 //go:build darwin || (linux && !baremetal)
 // +build darwin linux,!baremetal
 
+// target wasi sets GOOS=linux and thus the +linux build tag,
+// even though it doesn't show up in "tinygo info target -wasi"
+
 // Portions copyright 2009 The Go Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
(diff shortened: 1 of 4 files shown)


src/os/file_unix.go
```diff
@@ -1,6 +1,9 @@
 //go:build darwin || (linux && !baremetal)
 // +build darwin linux,!baremetal
 
+// target wasi sets GOOS=linux and thus the +linux build tag,
+// even though it doesn't show up in "tinygo info target -wasi
```
dkegel-fastly: looks like the comment was truncated

ZauberNerd: Fixed.

QuLogic:
If this will be supported everywhere, that will be helpful for #2515.

dkegel-fastly (approved):
LGTM

deadprogram:
Thank you @ZauberNerd :fire: merging.

PR #2682, opened 2022-03-05
src/os: Add UserHomeDir() function to os package

ZauberNerd:
This commit adds the `UserHomeDir()` function to the `os` package.

### DIFF
--- a/src/os/file.go
+++ b/src/os/file.go
@@ -12,6 +12,7 @@ package os
 import (
 	"errors"
 	"io"
+	"runtime"
 	"syscall"
 )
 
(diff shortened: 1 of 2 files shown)


dkegel-fastly:
Please also port the corresponding test if possible, e.g. TestUserHomeDir()

dkegel-fastly:
Looks like it failed "make tinygo-test-wasi-fast".

ZauberNerd:
Ok, so it fails because `$HOME` is not defined: https://github.com/tinygo-org/tinygo/pull/2682/files#diff-9104369aad12ebcc262dc47f26321742fe4b10d31e30bc53832e869f4a7b3bceR277
And the [`t.SkipF()`](https://github.com/tinygo-org/tinygo/pull/2682/files#diff-ad834cc3078ff26b65749c4d2bf9bb98dc5f7069502c99aa21aa34275e770e1bR14) function didn't seem to do what I expected.

But this begs the question: Is it correct, that it returns an empty string / error on WASI (and possibly on baremetal too?)

The test failures on Windows are unrelated - the `MinoruSekine/setup-scoop@v1` fails.

dkegel-fastly:
I suspect / would be a fine thing to return on WASI if HOME is not set (cf.https://github.com/WebAssembly/WASI/issues/771 ).

t.SkipF() is unfortunately broken for now on Tinygo, with no fix planned any time soon; you can use t.Log(("Skipping...") followed by return as a workaround.

deadprogram:
Please rebase against the latest dev, so that the Windows tests can run correctly. Thanks!

dkegel-fastly (changes requested):
upstream has the test in os_test.go rather than file_test.go, suggest moving it to end of os_anyos_test.go.  Otherwise this is identical to upstream and ready to go.

dkegel-fastly (approved):
LGTM

deadprogram:
Thank you @ZauberNerd and @dkegel-fastly now merging.

Issue #2670, opened 2022-03-03
add support for .syso files


eliasnaur:
The [`gioui.org/shader/piet`](https://git.sr.ht/~eliasnaur/gio-shader/tree/main/item/piet) package uses .syso files to embed pre-built object files into Go programs. It seems TinyGo ignores them:

```
tinygo flash -target stm32f469disco ./stm32f4
ld.lld: error: undefined hidden symbol: elements_coroutine_begin
>>> referenced by elements_abi.c
>>>               /Users/e/Library/Caches/tinygo/obj-920bd271d98c454017dee52698a38d15be7e3a20326ee84e434c6d2a.o:(elements_program_info)

ld.lld: error: undefined hidden symbol: elements_coroutine_await
>>> referenced by elements_abi.c
>>>               /Users/e/Library/Caches/tinygo/obj-920bd271d98c454017dee52698a38d15be7e3a20326ee84e434c6d2a.o:(elements_program_info)

ld.lld: error: undefined hidden symbol: elements_coroutine_destroy
>>> referenced by elements_abi.c
>>>               /Users/e/Library/Caches/tinygo/obj-920bd271d98c454017dee52698a38d15be7e3a20326ee84e434c6d2a.o:(elements_program_info)
error: failed to link /var/folders/3p/_lbqc_4n5t764s1c44sql13w0000gn/T/tinygo2157102769/main: exit status 1
```

aykevl:
After a quick Google search, I didn't find much information about .syso files. They appear to be basically just object files. Is that correct? Do you have any further information on them?
(If they are just object files, it should be relatively easy to add support for them in TinyGo by simply adding them to the linker command line).

aykevl:
Oh but you're trying to use them on a Cortex-M4? The Cortex-M4 uses the Thumb2 instruction set, which is different from the ARM instruction set. You won't be able to use elements_linux_arm.syso on the stm32f469, for example.

If these are files compiled from C, it's probably a better idea to just compile those via CGo.

dgryski:
@aykevl I believe they are just object files, and conveniently available via `go list`:
```
        SysoFiles       []string   // .syso object files to add to archive
```

eliasnaur:
Yes, they're just object files with a distinct filename extension. Best resource I found was https://zchee.github.io/golang-wiki/GcToolchainTricks/.

[...]

PR #2686, opened 2022-03-07
builder,loader: link .syso files

eliasnaur:


aykevl (commented):
This looks fine and I think it would work, but I'm a bit unsure we should do it like this. Because we have many targets (baremetal and wasi) that pretend to be something they are not, usually linux/arm. And I don't think we should link .syso files in these cases.
We could for example add `(compileopts.Config).IsRegularOS()` (name TBD) that checks for particular build tags, I think `wasi` and `baremetal` would be enough.
For normal OSes (linux, MacOS, Windows), linking them should be fine.

Another consideration: TinyGo (mostly) uses the C calling convention while Go uses a special calling convention. I'm not entirely sure how .syso files are normally used but if they are intended to be used directly from Go files, this calling convention mismatch would be a problem.

eliasnaur:
My STM32F469 board is baremetal, right? If so, not linking .syso files for baremetal targets won't fix #2670 for me. I agree that linking linux/arm syso files are probably a bad idea: at the very least I'd like to discern `_arm.syso` files from, say, `_arm_thumb2.syso` files built for the target.

dgryski:
I believe syso files use the standard C calling convention.

aykevl:
> My STM32F469 board is baremetal, right? If so, not linking .syso files for baremetal targets won't fix #2670 for me.

Correct.
I'm not sure what the best solution is here. Linking _linux_arm.syso files would indeed be bad for existing code (like [gioui.org/shader/piet](https://git.sr.ht/~eliasnaur/gio-shader/tree/main/item/piet)). But then there is no way to link .syso files for baremetal systems: _linux_thumb2.go won't work as a build exclusion rule with the Go toolchain because `thumb2` isn't a recognized OS/arch name.

What about this: we just say that if users want to link object files in TinyGo on baremetal systems, they have to link .o (or .a) files directly using `#cgo LDFLAGS`. It seems to me that .syso files are normally used as a way to avoid CGo, which is a limitation that doesn't exist in TinyGo. It's perhaps a bit ugly, but I can't really think of a situation where it can't work (unless for some weird reason CGo stops working in TinyGo).

eliasnaur:
Completely agree: automatic .syso linking is for avoiding Cgo and manually adding `#cgo LDFLAGS` is a small price to pay. Thank you for your thoughts.

Issue #1553, opened 2020-12-29
maps need growing logic when they fill up


kokizzu:
Tried to compile and run [this code](https://github.com/kokizzu/kokizzu-benchmark/blob/master/assoc-small/map.go), but it doesn't complete within 1000s (while other languages/compiler completed within 1-10s)

```
// map.go
package main

import (
        "bytes"
        "fmt"
        "strconv"
)

type MSD map[string]float64

const MAX_DATA = 1230000

var i2ch []byte

func init() {
        i2ch = []byte{'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'B', 'c', 'D', 'e', 'F'}
}
func get_first_digit(d float64) int {
        for d > 10 {
                d /= 10
        }
        return int(d)
}
func to_rhex(v int) string {
        hex := bytes.Buffer{}
        for v > 0 {
                hex.WriteByte(i2ch[v%16])
                v /= 16
        }
        return hex.String()
}
func set_or_inc(m map[string]float64, key string, set, inc int, ctr *int) {
        if _, ok := m[key]; !ok {
                m[key] = float64(set)
        } else {
                m[key] += float64(inc)
                *ctr += 1
        }
}
func main() {
        m := MSD{}
        dup1, dup2, dup3 := 0, 0, 0
        for z := MAX_DATA; z > 0; z-- {
                val2 := MAX_DATA - z
                val3 := MAX_DATA*2 - z
                key1 := strconv.Itoa(z)
                key2 := strconv.Itoa(val2)
                key3 := to_rhex(val3)
                set_or_inc(m, key1, z, val2, &dup1)
                set_or_inc(m, key2, val2, val3, &dup2)
                set_or_inc(m, key3, val3, z, &dup3)
        }
        fmt.Println(dup1, dup2, dup3)
        total, verify, count := 0, 0, 0
        for k, v := range m {
                total += get_first_digit(v)
                verify += len(k)
                count += 1
        }
        fmt.Println(total, verify, count)
}
```
![image](https://user-images.githubusercontent.com/1061610/103266354-fc0b4b80-49e1-11eb-8081-d2210c57e696.png)

kokizzu:
I'm using ubuntu 20.04 64-bit, default target

kokizzu:
sometimes segfault:

```
$ time tinygo run map.go             
error: failed to run compiled binary /tmp/tinygo923402887/main: signal: segmentation fault (core dumped)
[...]

PR #2687, opened 2022-03-07
src/runtime: make maps grow automatically

dgryski:
Should be working.  Need to add a map-growing-and-iterator test case.


dgryski:
Things I might like to do (in no particular order):
   - zero out fields in the iterators (when done) and map (when deleted) to avoid pinning allocations
   - switch hash32 to leveldb hash (faster hash)
   - include per-map hash seed
   - (maybe) find improved 64-bit hash function

dgryski:
This was the test program I was using:

```
package main

func main() {

	m := make(map[int]int)

	const N = 1000

	for i := 0; i < 10; i++ {
		m[i] = i
	}

	var deleted bool
	for k, v := range m {
		println(k)
		if k == 5 {
			// grow map
			for i := 10; i < 1000; i++ {
				m[i] = i
			}

			// delete some elements
			for i := 0; i < 10; i++ {
				delete(m, i)
			}
			deleted = true
			continue
		}

		// make sure we never see a deleted element later in our iteration
		if deleted && v < 10 {
			println("saw deleted element", v)
		}
	}

	if len(m) != 990 {
		println("bad length post grow/delete")
	}

	var mcount int
	for k, v := range m {
		if k != v {
			println("element mismatch", k, v)
		}
		if v < 10 {
			println("saw deleted element post-grow")
		}
		mcount++
	}

	if mcount != 990 {
		println("bad number of elements post-grow:", mcount)
	}

	println("done")

}
```

src/runtime/hashmap.go
```diff
@@ -83,7 +116,7 @@ func hashmapLenUnsafePointer(p unsafe.Pointer) int {
 
 // Set a specified key to a given value. Grow the map if necessary.
 //go:nobounds
-func hashmapSet(m *hashmap, key unsafe.Pointer, value unsafe.Pointer, hash uint32, keyEqual func(x, y unsafe.Pointer, n uintptr) bool) {
+func hashmapSet(m *hashmap, key unsafe.Pointer, value unsafe.Pointer, hash uint32, keyEqual func(x, y unsafe.Pointer, n uintptr) bool, keyHash func(y unsafe.Pointer, size uintptr) uint32) {
```
dgryski: I had the thought of using `hashmapAlgorithms[m.alg]` entries as needed, instead of passing functions pointers around to the various functions that need them.

compiler/map.go
```diff
@@ -17,22 +17,27 @@ func (b *builder) createMakeMap(expr *ssa.MakeMap) (llvm.Value, error) {
...
+		alg = 2
 	} else if hashmapIsBinaryKey(keyType) {
 		// Trivially comparable keys.
 		llvmKeyType = b.getLLVMType(keyType)
+		alg = 1
 	} else {
 		// All other keys. Implemented as map[interface{}]valueType for ease of
 		// implementation.
 		llvmKeyType = b.getLLVMRuntimeType("_interface")
+		alg = 2
```
aykevl: This looks like a bug:

```suggestion
		alg = 3
```

src/runtime/hashmap.go
```diff
@@ -12,12 +12,30 @@ import (
...
 	keySize    uint8 // maybe this can store the key type as well? E.g. keysize == 5 means string?
 	valueSize  uint8
 	bucketBits uint8
+	alg        uint8
+}
+
+type hashmapAlgorithm uint8
+
+const (
+	hashmapAlgorithmBinary hashmapAlgorithm = iota + 1
```
aykevl: Why not start at 0? It would avoid the zero entry in `hashmapAlgorithms` (and possibly make the rest of the code very slightly smaller).

src/runtime/hashmap.go
aykevl: Yes, that would probably be better.
The optimizer has a lot of trouble looking through function pointers, so it's generally a good idea to get rid of them. Using `hashmapAlgorithms[alg]` still is a function pointer though: it's unlikely the optimizer manages to look through it. The most efficient way (although perhaps not the most readable way) is to use a switch:

```go
func hashmapEqual(x, y unsafe.Pointer, n uintptr, alg int) bool {
	switch alg {
	case hashmapAlgorithmString:
		return hashmapStringEqual(x, y, n)
	case hashmapAlgorithmBinary:
		// ..etc
}
```

Regular function calls still allow inlining and various other interprocedural optimizations such as const propagations.

src/runtime/hashmap.go
```diff
@@ -158,6 +195,32 @@ func hashmapInsertIntoNewBucket(m *hashmap, key, value unsafe.Pointer, tophash u
...
+	n.bucketBits = m.bucketBits + 1
+	numBuckets := uintptr(1) << n.bucketBits
+	bucketBufSize := unsafe.Sizeof(hashmapBucket{}) + uintptr(m.keySize)*8 + uintptr(m.valueSize)*8
+	n.buckets = alloc(bucketBufSize*numBuckets, nil)
+
+	// use a hashmap iterator to go through the old map
+	var it hashmapIterator
+
+	var key = alloc(uintptr(m.keySize), nil)
+	var value = alloc(uintptr(m.valueSize), nil)
```
aykevl, line 236: You could perhaps merge these two allocations:

```suggestion
	var key = alloc(uintptr(m.keySize)+uintptr(m.valueSize), nil)
	var value = unsafe.Pointer(uintptr(key) + uintptr(m.keySize))
```
(...although this might need some more alignment to be correct)

src/runtime/hashmap.go
```diff
@@ -249,26 +312,29 @@ func hashmapDelete(m *hashmap, key unsafe.Pointer, hash uint32, keyEqual func(x,
...
 	if m == nil {
-		// Iterating over a nil slice appears to be allowed by the Go spec:
-		// https://groups.google.com/g/golang-nuts/c/gVgVLQU1FFE?pli=1
-		// https://play.golang.org/p/S8jxAMytKDB
+		// From the spec: If the map is nil, the number of iterations is 0.
 		return false
 	}
 
-	numBuckets := uintptr(1) << m.bucketBits
+	if it.buckets == unsafe.Pointer(uintptr(0)) {
```
aykevl:
```suggestion
	if it.buckets == nil {
```

aykevl (commented):
Overall this looks pretty good. Nice and simple. There are a few small issues though.

I think this PR would need at least a test case for iterator behavior (elements that are removed during iteration are not visited in the iteration, and generally that modifying a map while iterating doesn't result in weird behavior).

I would prefer if other changes (such as using a different hash key) would be done in separate PRs to make things easier to review.

Also, did you benchmark this change? It's supposed to be a performance improvement, so I would expect it to at least improve performance for large maps.

src/runtime/hashmap.go
dgryski: Ok, I'll fix up the code to rely on the passed `alg` value and switches.

compiler/map.go
dgryski: Oops, nice catch.

src/runtime/hashmap.go
dgryski: Because I initially wanted to catch places where `alg` wasn't initialized and having the nil function pointers would make it obvious.

I'll remove it since I'm more confident I'm catching it now.  (I used to initialize the fields in the new map by hand, but now I'm just copying the entire struct and replacing the fields I care about.)

src/runtime/hashmap.go
dgryski: 👍

dgryski:
PTAL

dgryski:
The benchmark I ran was the test case from https://github.com/tinygo-org/tinygo/issues/1553 which goes from taking ~forever to taking 5-10s.

aykevl:
The code itself looks good.

I did however test this PR with the drivers package and there are a number of packages where binary size grows by ~800 bytes. I would expect this for complex drivers, but most of these were simple sensor reading drivers (over I2C for example), seemingly without using maps anywhere. That's rather surprising.
You can see the results here: https://gist.github.com/aykevl/fbb713b830d9e588e64230bf3b07ed3e

aykevl:
Oh I see. The size increase is in the interface code. Which makes sense: the compiler can now no longer prove that something like `hashmapInterfaceEqual` isn't used.
This might be fixable.

aykevl:
This is the patch I tried: https://gist.github.com/aykevl/13c87508f5a7e1cd18b7d285bcac4f85
It reduces binary size slightly in a few rare cases (-0.02% overall) but doesn't fix the main issue here.

... I think it might be better to use function pointers after all.

dgryski:
Ok, I'll switch back to function pointers.

dgryski:
Seems like switching back to function pointers loaded set via the `alg` value seems to allow the dead code pass to work.  (I couldn't figure out how to run the full drivers code size test).

Before:

```
~/go/src/github.com/tinygo-org/tinygo/testdata $ tinygo build -target=wasi -size=full mapgrowth.go
  13105       0    1131  129770 |   14236  130901 | (unknown)
   1122       0       0       4 |    1122       4 | internal/task
   1609       0       0       0 |    1609       0 | main
  10432       0       0       0 |   10432       0 | reflect
  14395       0      12     155 |   14407     167 | runtime
  40663       0    1143  129929 |   41806  131072 | total
```

After:
```
~/go/src/github.com/tinygo-org/tinygo/testdata $ tinygo build -target=wasi -size=full mapgrowth.go
   6030       0     279  130622 |    6309  130901 | (unknown)
   1122       0       0       4 |    1122       4 | internal/task
   1649       0       0       0 |    1649       0 | main
   8293       0      12     155 |    8305     167 | runtime
  17094       0     291  130781 |   17385  131072 | total
```

dgryski:
Cleaned up the commit sequence a bit.

deadprogram:
@aykevl any last feedback before we merge?

deadprogram (approved):

deadprogram:
Now merging, thanks for all the work @dgryski

deadprogram:
@dgryski I perhaps merged too soon? Please check this CI failure: https://github.com/tinygo-org/tinygo/runs/5815785664?check_suite_focus=true#step:9:11

dgryski:
Bizarre test failure..  `hash/adler32 has no maps`.

dgryski:
Oh yay, the crash goes away if I enable gc asserts...

FrankReh:
Just some thoughts. No real help probably.

I couldn't find maps used in hash/adler32 either but the testing package uses several maps that start with default size allocation so might be growing when a lot of tests are run. One could run just hash/adler32 and see if it passes when it is the only test that matches the test -run <regex>. You see the maps in src/testing/testing.go. One could also tweak the make calls for the maps to give them a much larger starting size to see if that changes things. Also there are a few packages that adler32_test.go pulls in and many packages that src/testing/testing.go pulls in and some of them might have maps that grow in this test situation.

It's common for results to change when trying to track down memory corruption because anything we do changes the function frame size and changes the sizes allocated and often the location of heap memory being returned. Even the amount that gets inlined changes when we add prints to a function, sometimes causing more frames on the stack which can cause a GC to see some memory is used when otherwise it might think it is free (I know anyone reading this probably knows that already). And of course when and how often GC itself is called can be changed whenever something about the image under test changes. I sometimes resort to calling runtime.GC manually and adding a runtime.ZeroFreeMemory() and calling that, and even changing the GC so it resets the next address, all to try to get some better apples to apples tests when tracking down why something heap related has started to fail. I've also found forcing GC to grow memory for wasm builds early, before actual test code runs, can make it easier to see what's going on. Larger wasm memory will mean less churn and less variability to results as functions change from debugging iteration to iteration. [All that to say this isn't so uncommon.]

Looking at the new code is probably the best way to track down the cause of what is probably new memory corruption but I didn't see anything yet. Maybe someone else can.

Short of that, running a subset of tests, either manually or by tweaking the Makefile, might shed light on the problem. Like does the corruption cause a different test to fail or is it always the adler32 test. And then a manual binary search on when the new functionality is enabled could lead to the map type or the growth type or the iteration type that is at issue - the function hashmapShouldGrow could use a global counter during debugging to also decide when to open the gate for the new functionality - presumably there are no failures if this function always returns false, even with the new code in the build, so a binary walk of counter values to see at what value do the problems start would allow adding a panic once that value was hit in the function and then figuring out what map was about to grow and what the circumstances around the map or an iteration around that map were.

If someone can give instructions for how to see a test failure running darwin builds, or wasm builds, I would try to investigate myself. If it's just CI or Linux environments where it fails, I'll have to stay on the sidelines.

PR #2688, opened 2022-03-07
Add support for Go assembly

aykevl:
Work in progress. The math package works, but many other packages (like crypto/aes) don't work yet.
This draft is also for amd64/arm64 only.


aykevl:
Closing in favor of #3103.

PR #2694, opened 2022-03-09
Per-map hash seeds

dgryski:
Created on top of https://github.com/tinygo-org/tinygo/pull/2687

### DIFF
--- a/src/runtime/hashmap.go
+++ b/src/runtime/hashmap.go
@@ -13,12 +13,13 @@ import (
 // The underlying hashmap structure for Go.
 type hashmap struct {
 	buckets    unsafe.Pointer // pointer to array of buckets
+	seed       uintptr
 	count      uintptr
 	keySize    uint8 // maybe this can store the key type as well? E.g. keysize == 5 means string?
 	valueSize  uint8
 	bucketBits uint8
 	keyEqual   func(x, y unsafe.Pointer, n uintptr) bool
-	keyHash    func(key unsafe.Pointer, size uintptr) uint32
+	keyHash    func(key unsafe.Pointer, size, seed uintptr) uint32
 }
 
 type hashmapAlgorithm uint8
@@ -74,6 +75,7 @@ func hashmapMake(keySize, valueSize uint8, sizeHint uintptr, alg uint8) *hashmap
 
 	return &hashmap{
 		buckets:    buckets,
+		seed:       uintptr(fastrand()),
 		keySize:    keySize,
 		valueSize:  valueSize,
 		bucketBits: bucketBits,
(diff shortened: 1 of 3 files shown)


dgryski:
Draft because it's still failing tests.

dgryski:
Bah, these commits ended up with too many changes.  I'm going to split out a few of them into a smaller commit.

dgryski:
I don't understand this CI failure.

dgryski:
Pulled out the hash function changes to see if that's what CI doesn't like.

dgryski:
Huh, so indeed something about the new hash functions causes the CI issue.  I'll keep that out for now.

deadprogram:
@dgryski can you now rebase this PR and resolve merge conflicts? Thanks!

dgryski:
Rebased against `dev` now that map growth PR has landed.

dgryski:
ping @aykevl

src/runtime/hashmap.go
```diff
@@ -413,15 +418,15 @@ func hashmapBinaryGet(m *hashmap, key, value unsafe.Pointer, valueSize uintptr)
...
 	}
-	hash := hash32(key, uintptr(m.keySize))
+	hash := hash32(key, uintptr(m.keySize), m.seed)
 	return hashmapGet(m, key, value, valueSize, hash)
 }
 
 func hashmapBinaryDelete(m *hashmap, key unsafe.Pointer) {
 	if m == nil {
 		return
 	}
```
aykevl, line 428: Now that all the `hashmap*Set`, `hashmap*Get`, and `hashmap*Delete` functions have nil checks themselves, I think the nil checks can be removed from the functions `hashmapSet`, `hashmapGet`, and `hashmapDelete`.

aykevl (commented):
The code itself looks reasonable, with one nit below. I've also checked the binary size impact and it's actually quite minor. Of all the driver examples, only a few changed size (apparently maps aren't used much in driver tests). And for the regular tests, the biggest increases were in `testdata/testing.go`, `testdata/map.go`, `testdata/reflect.go`, which is either somewhat unusual Go code (map, reflect) or simply very large (testing).

Can you explain a bit more about why this is needed? Is it for the [hashmap DoS attack](https://lwn.net/Articles/474912/)? I'm generally hesitant with merging things that increase code size, unless there's a clear benefit.

dgryski:
Yes, this is to protect against hash flooding attacks. While less important for microcontroller targets, the people running wasm web services will find this a welcome security fix. Incoming HTTP Headers, for example, are generally stored in maps and being user controlled makes them an easy avenue to exploit.

aykevl (approved):
Thank you for the explanation!

deadprogram:
:tada: thank you @dgryski now merging!

PR #2697, opened 2022-03-10
atsamd51: allow faster frequency setting when using SPI

sago35:
For applications requiring SPI speed, the current 48 MHz setting is too slow.
If SERCOM_FREQ_REF is 48 MHz, the actual maximum SPI speed is 24 MHz.

ILI9341 such as Wio Terminal can operate up to 80MHz SPI clock.
First, I suggest setting it to GCLK0 (120MHz), which is easy to set.


deadprogram:
Hello @sago35 seems like this may have introduced an error on i2c for samd51.

The TinyHCI test is consistently failing on this commit. See https://github.com/tinygo-org/tinygo/pull/2697/checks?check_run_id=5494732040

sago35:
Yes, I should have checked all of SERCOM's features.
I will check and correct this later.

sago35:
I changed the design to branch the GCLK selection only on SPI.
GCLK1 is always used for UART and I2C, so the behavior is the same as before.

I have checked below and it works as expected.

* UART : 115200
* I2C : 100kHz, 400kHz
* SPI : 24MHz, 30MHz, 60MHz

deadprogram (approved):

deadprogram:
Thanks @sago35 now merging.

Issue #2699, opened 2022-03-10
optimize map[string] lookup from []byte key


FrankReh:
The current system works but enlists an allocation and memcopy when it calls runtime_stringFromBytes.

The go compiler team did this with their commit:
    https://github.com/golang/go/commit/f5f5a8b6209f84961687d993b93ea0d397f5d5bf#diff-3437cd20ec7506421b8d8b653efa9bfe
    "cmd/gc, runtime: optimize map[string] lookup from []byte key"

Maybe tinygo can do something similar.

If tinygo could enlist escape analysis, it might even go further to use the string([]byte) construct generally for variables or argument passing.

As it stands, there is a mechanism enlisting unsafe used in the wild and the strings package for converting a byte slice to a string without the allocation and copy. And the support by tinygo for unsafe already makes it work. See the example further down.

The basic map lookup under consideration:
```go
func bytes2StringLookup(m map[string]bool, b []byte) bool {

	return m[string(b)]

}
```

This is compiled by tinygo to this wasm function (note the call to runtime_stringFromBytes):

```
function bytes2StringLookup(a:int, b:int):int {
  var c:int = stack_pointer - 32;
  stack_pointer = c;
  c[5]:long@4 = 1L;
  var d:int = 99708[0]:int;
  99708[0]:int = c + 16;
  c[4]:int = d;

  runtime_stringFromBytes(c, b, 32, 32);

  99708[0]:int = d;
  (c + 24)[0]:int = (d = c[0]:int);
  runtime_hashmapStringGet(a, d, c[1]:int, c + 15, 1);
  d = c[15]:ubyte;
  stack_pointer = c + 32;
  return d;
}
```

Here is the unsafe workaround that can be applied manually when a developer wants to take the responsibility for misuse into their own hands. This conversion is already used by the strings.Builder and works in tinygo as tinygo supports strings.Builder without modification.

```go
func unsafeGetString(b []byte) string {
	return *(*string)(unsafe.Pointer(&b)) // As done in strings.Builder
}

func bytes2UnsafeStringLookup(m map[string]bool, b []byte) bool {
	return m[unsafeGetString(b)]
}
```

Some benchmarks showing bytes, strings, and manual/unsafe conversions between the two. Starred are the relevant results for this issue.

```
tinygo/wasm running in nodejs on older mac:
BenchmarkRangeBytes
BenchmarkRangeBytes              	83960235	        14.45 ns/op
[...]

PR #2701, opened 2022-03-10
compiler: optimize map[string([]byte))] lookups

dgryski:


dgryski:
This doesn't seem to work (according to the benchmarks in https://github.com/tinygo-org/tinygo/issues/2699 ) and I have no idea why.

dgryski:
So, my PR adds is the definitions for `%5` and `%6`, but `%3`, `%4`, and the call to `runtme.trackPointer` still seem to be hanging around.  Not sure how to remove those.  Probably at a later stage in a transform optimization pass.

```
  %3 = call %runtime._string @runtime.stringFromBytes(i8* %b.data, i32 %b.len, i32 %b.cap, i8* undef), !dbg !62164
  %4 = extractvalue %runtime._string %3, 0, !dbg !62164
  call void @runtime.trackPointer(i8* %4, i8* undef), !dbg !62164
  %5 = insertvalue { i8*, i32 } undef, i8* %b.data, 0, !dbg !62165
  %6 = insertvalue { i8*, i32 } %5, i32 %b.len, 1, !dbg !62165
```

FrankReh:
okay. glad you are making progress in understanding this. I was wondering why I didn't see it working either. I really don't understand that level of IR or the passes the tinygo compiler is making so not sure how I can help at the moment.

aykevl:
There is a very similar optimization already: `OptimizeStringToBytes` (see transform/rtcalls.go). But you want the opposite, essentially.

I'm thinking of ways to make this `[]byte`-to-`string` optimization more general. But it's kind of difficult. For the `string`-to-`[]byte` conversion optimization, it's easy: just check that none of the uses of the conversion modify the slice. But the other way round is hard because even though that particular function might not modify the slice, there can be aliases that might be modified.

compiler/compiler.go
```diff
@@ -1775,6 +1776,40 @@ func (b *builder) createExpr(expr ssa.Value) (llvm.Value, error) {
...
+								key := b.getValue(convert.X)
+
+								keyBuf := b.CreateExtractValue(key, 0, "map.keyBuf")
+								keyLen := b.CreateExtractValue(key, 1, "map.keyLen")
+
+								// Create the string
+								str := b.ctx.ConstStruct([]llvm.Value{
+									llvm.Undef(b.i8ptrType),
+									llvm.Undef(b.uintptrType),
+								}, false)
```
aykevl, line 1798: This looks unsafe. A string is not just a struct, it's a `runtime._string` named struct. ~And I don't think these values you're inserting are actually constants.~ nevermind misread the code.

dgryski:
This is not the approach.

PR #2705, opened 2022-03-12
Add stub for runtime numcgocall and numgoroutine

attriaayush:
This adds a stub for `NumCgoCall`, `NumGoroutine` and `Version` functions from the runtime package.

Since the multi-threading is currently not supported, hard coding `NumGoroutine` value to 1 (main thread), `NumCgoCall` to 0 ~and lastly for the `Version` It is either the commit hash, date of the build or release tag but I have just put in `v0.1.0` for now~. From what I can see this change should not break anything since these functions don't already exist.

### DIFF
--- a/src/runtime/debug.go
+++ b/src/runtime/debug.go
@@ -8,3 +8,17 @@ package runtime
 func NumCPU() int {
 	return 1
 }
+
+// NumCgoCall returns the number of cgo calls made by the current process.
+func NumCgoCall() int {
+	return 0
+}
+
+// NumGoroutine returns the number of goroutines that currently exist.
+func NumGoroutine() int {
+	return 1
+}
+
+func Version() string {
+	return "v0.1.0"
+}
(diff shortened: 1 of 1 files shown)


dgryski:
Note that `Version` is also being discussed here: https://github.com/tinygo-org/tinygo/pull/2676

attriaayush:
> Note that `Version` is also being discussed here: #2676

@dgryski That makes sense, happy to remove the changes related to `Version` in that case from this PR.

deadprogram (approved):

aykevl:
> Since the multi-threading is currently not supported, hard coding `NumGoroutine` value to 1 (main thread)

TinyGo does support goroutines. So returning 1 is technically incorrect. But TinyGo doesn't currently track which goroutines are still alive (they just get garbage collected when finished) so there is no way to know.
But for compatibility, this might be fine.

> `NumCgoCall` to 0

For compatibility, I think returning 0 would be fine.

Can you update the comments to point out these are stubs and don't return the real value?

PR #2706, opened 2022-03-12
main: calculate default output path if -o is not specified

aykevl:
This matches the Go command and is generally convenient to have.


deadprogram (approved):

sago35 (approved):

deadprogram:
Thank you for adding this @aykevl and thanks for review @sago35 now merging.

PR #2707, opened 2022-03-12
cgo: slightly improve error messages

aykevl:
Updating them to libclang-13-dev was a good change, but we can go even further:

  * The suggestion didn't apply to MacOS.
  * The suggestion would need to be updated with every LLVM release, which is a maintenance burden.
  * The suggestion is wrong when compiling with `-tags=llvm12` for example to choose a different LLVM version.

Therefore, link to the build documentation instead.

### DIFF
--- a/cgo/libclang.go
+++ b/cgo/libclang.go
@@ -16,7 +16,7 @@ import (
 )
 
 /*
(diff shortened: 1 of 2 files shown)


deadprogram:
Thanks for the clarification here @aykevl now merging.

PR #2691, opened 2022-03-08
Makefile: chmod before fpm. 

dankegel:
This works around a likely permissions problem in "tinygo build-library"

(Originally removed a -j3, but testing on a hex core machine showed -j3 really does speed things up a bit.)

Fixes #2685

Issue #2685, opened 2022-03-06
Build Error on Ubuntu


charles-d-burton:
Installed using the .deb package for ubuntu.  I'm getting the following permissions error, it's easy enough to fix, just posting this here because I think the .deb needs the permissions updated

```bash
❯ tinygo build -target pico -o test.u2f
tinygo:ld.lld: error: cannot open /usr/local/lib/tinygo/pkg/thumbv6m-unknown-unknown-eabi-cortex-m0plus/compiler-rt/lib.a: Permission denied
tinygo:ld.lld: error: cannot open /usr/local/lib/tinygo/pkg/thumbv6m-unknown-unknown-eabi-cortex-m0plus/picolibc/lib.a: Permission denied
failed to run tool: ld.lld
error: failed to link /tmp/tinygo157798778/main: exit status 1
❯ tinygo version
tinygo version 0.22.0 linux/amd64 (using go version go1.17.7 and LLVM version 13.0.0)
```

The permissions on those files are set to read by root onlY

```bash
❯ ls -l /usr/local/lib/tinygo/pkg/thumbv6m-unknown-unknown-eabi-cortex-m0plus/picolibc/lib.a
-rw------- 1 root root 705066 Jan 25 19:00 /usr/local/lib/tinygo/pkg/thumbv6m-unknown-unknown-eabi-cortex-m0plus/picolibc/lib.a
```

charles-d-burton:
I fixed it locally with the following command
```bash
❯ sudo find /usr/local/lib/tinygo/pkg/ -type f | sudo xargs chmod 644
❯ tinygo build -target pico -o test.u2f
```

dankegel:
Confirming present in 0.21 and 0.22 (but not in 0.20, which didn't contain those files):
```
$ wget https://github.com/tinygo-org/tinygo/releases/download/v0.22.0/tinygo_0.22.0_amd64.deb
$ ar x tinygo_0.22.0_amd64.deb 
$ tar tvf data.tar.gz | grep ".-----"
-rw------- 0/0          401872 2021-11-18 05:31 ./usr/local/lib/tinygo/pkg/armv6m-unknown-unknown-eabi/compiler-rt/lib.a
drwx------ 0/0               0 2021-11-18 05:31 ./usr/local/lib/tinygo/pkg/armv6m-unknown-unknown-eabi/picolibc/include/
[...]

PR #2708, opened 2022-03-12
builder: use correct permission bits when creating a library

aykevl:
Previously, the wrong permission bits were emitted by `tinygo build-library`. This commit fixes that, by `chmod`'ing to reasonable default permission bits.

See: https://github.com/tinygo-org/tinygo/issues/2685 and https://github.com/tinygo-org/tinygo/pull/2691.


deadprogram:
Thanks for a more complete fix to this issue. Merging.

PR #2710, opened 2022-03-13
build: trigger builds of dev branches of repos in TinyGo ecosystem

deadprogram:
This PR triggers builds of `dev` branches of all the repos in the TinyGo ecosystem e.g. Bluetooth, TinyFS, TinyFont, TinyDraw.

The idea is to make it immediate that we know about breaking changes the same way that we currently do for the Drivers repo.


deadprogram:
Working exactly as expected!

PR #2711, opened 2022-03-13
On baremetal platforms, use simpler test matcher.  Fixes #2666.

### LINKED ISSUE #2666: regression: "panic: runtime error: out of memory" error on some tests on target cortex-m-qemu
Using the command

    tinygo test -v -target=cortex-m-qemu $foo

the following tests pass on both 0.21 and 0.23-dev:

container/heap
container/list
container/ring
encoding
hash/fnv
internal/itoa
reflect
testing/iotest
unicode/utf16

but the following tests pass on 0.21 and fail on 0.23-dev:

crypto/des
encoding/ascii85
encoding/hex
math/cmplx
net/mail
text/scanner
unicode/utf8

In all cases, the log shows "panic: runtime error: out of memory".

dkegel-fastly:
On baremetal targets, there is no argv anyway, so even this minimal matcher is probably overkill... 
but it does work, in case anyone does jam arguments in somehow.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -239,3 +239,4 @@ jobs:
           rm xtensa-esp32-elf-gcc8_2_0-esp-2020r2-linux-amd64.tar.gz
       - run: make smoketest
       - run: make wasmtest
+      - run: make tinygo-baremetal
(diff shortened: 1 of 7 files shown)


dgryski (approved):
LGTM

deadprogram:

PR #2714, opened 2022-03-14
interp: prevent an off-by-one during interp debug

dgryski:


interp/interpreter.go
```diff
@@ -493,8 +493,8 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 				// possibly trying to recover from it if it failed to execute.
 				if r.debug {
 					argStrings := make([]string, len(operands))
```
QuLogic: This should be `len(operands)-1`, I think, or you get a trailing comma, unless that's desired for some reason.

dgryski: Oh, you're right. I fixed the out-of-bounds crash but didn't then check the output was correct.

dgryski: Fixed off-by-one in the off-by-one fix.

deadprogram:
Looking good, thanks @dgryski for the fix and @QuLogic for the review. Now merging.

PR #2716, opened 2022-03-15
os: Add some exec.ProcessState stubs

QuLogic:
There was a bit more, but that got in in another PR, so I figured I should split this out from #2515 before someone else does it again.


dkegel-fastly (approved):
LGTM

deadprogram:
Thanks @QuLogic and @dkegel-fastly for review. Now merging.

PR #2719, opened 2022-03-17
Implement getpagesize and munmap.  For go 1.18.

dkegel-fastly:
"make test" with go 1.18 installed fails with
```
    main_test.go:541: test error: could not compile: /usr/local/go/src/internal/fuzz/sys_posix.go:37:11: Munmap not declared b
```
See https://github.com/tinygo-org/tinygo/pull/2515#issuecomment-1031295016


deadprogram:
Seems legit to me! Thanks @dkegel-fastly now merging.

Issue #2718, opened 2022-03-16
tinygo does not yet build and pass tests with go 1.18 installed


slzatz:
I am sure you're aware but hoping you can clarify the situation.  With go 1.18 released, tinygo won't build with the check for go version that is in builder/config.go.  Should we reinstall 1.17 or is 1.18 compatibility coming soon?

dkegel-fastly:
No ETA yet.  Personally, I'm sticking with 1.17 for the moment.

It's being worked on; see https://github.com/tinygo-org/tinygo/pull/2515 and related pull requests.

dkegel-fastly:
See https://github.com/tinygo-org/tinygo/pull/2725

dkegel-fastly:
Closing duplicate

dkegel-fastly:
See https://github.com/tinygo-org/tinygo/issues/2762 for the remaining test.

PR #2721, opened 2022-03-18
compiler.go: createBuiltin: accept alias for slice.

dkegel-fastly:
Seems to get 1.18 tests closer to passing.

I don't know llvm, so this was a bit of a guess.

For https://github.com/tinygo-org/tinygo/pull/2515 / https://github.com/tinygo-org/tinygo/issues/2718

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -1233,7 +1233,7 @@ func (b *builder) createBuiltin(argTypes []types.Type, argValues []llvm.Value, c
 	case "cap":
 		value := argValues[0]
 		var llvmCap llvm.Value
-		switch argTypes[0].(type) {
+		switch argTypes[0].Underlying().(type) {
 		case *types.Chan:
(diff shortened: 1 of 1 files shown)


aykevl (changes requested):
There is a better way to fix this:

```diff
- switch argTypes[0].(type) {
+ switch argTypes[0].Underlying().(type) {
```

This strips off the named type and makes the type switch work directly with the underlying type (which should be a `*types.Chan` or `*types.Slice`).

aykevl (approved):
Looks good to me. A test would be good but I couldn't get one to work, the following already works for me:

```go
package main

type Buffer []byte

func main() {
	l := make(Buffer, 3, 5)
	println(len(l), cap(l))
}

func foo(l Buffer) {
	println(len(l), cap(l))
}
```

But it's small and matches the len builtin, so merging.

PR #2723, opened 2022-03-19
On go 1.18, skip TestTest on targets lacking PROT_READ upstream

dkegel-fastly:
Path of least resistance FTW :-)

Works around
```
$ make test GOTESTFLAGS="-run TestTest/EmulatedCortexM3/Pass"
...
main_test.go:520: test error: could not compile: /usr/local/go/src/internal/fuzz/sys_posix.go:19:18: PROT_READ not declared by package syscall
```

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -486,11 +486,19 @@ func TestTest(t *testing.T) {
 			}
 		}
 
+		_, minor, err := goenv.GetGorootVersion(goenv.Get("GOROOT"))
+		if err != nil {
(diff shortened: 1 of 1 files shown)


aykevl:
To be honest, I don't think this is the right direction. I think we should define those values in the syscall package (even if that's a bit dirty). We can copy the values from Linux, for example.

dkegel-fastly:
Closing in favor of https://github.com/tinygo-org/tinygo/pull/2724

Issue #2616, opened 2022-02-06
'sub' in constant expression is unhandled


QuLogic:
Running tests (using several patches from #2515) with Go 1.18 fails with:
```
i64 sub (i64 ptrtoint ([0 x i8]* @"internal/fuzz._ecounters" to i64), i64 ptrtoint ([0 x i8]* @"internal/fuzz._counters" to i64))
panic: unknown constant expr
```

I _think_ this comes from [this bit of code in the fuzzer](https://github.com/golang/go/blob/d588f487703e773ba4a2f0a04f2d4141610bff6b/src/internal/fuzz/counters_supported.go#L20):
```
	size := uintptr(unsafe.Pointer(&_ecounters)) - uintptr(addr)
```

deadprogram:
@aykevl can you please take a look?

aykevl:
I'm starting to think it may be better to stub out the entire internal/fuzz package because it makes too many assumptions about the compiler and linker.
(If we ever want to implement fuzz testing in TinyGo, we could revisit this decision).

deadprogram:
I would agree with that decision for now.

lmaltsis:
I have run into this issue independently. here's a minimal trigger:

```go
package main

import (
    "unsafe"
)

var glob uintptr

func init() {
    b := make([]byte, 1)
    glob = uintptr(unsafe.Pointer(&b[len(b)-1])) - 1
}

func main() {}
```

PR #2725, opened 2022-03-19
Support Go 1.18, updated

dkegel-fastly:
This is https://github.com/tinygo-org/tinygo/pull/2515 rebased against current dev, with a few new commits fixing missing signal values, adding a missing stub or two, and crucially, fixing #2726 , now with Damian's magic hack when goroot is 1.18 or above.

The testdata change in this branch differs from qulogic's slightly for no good reason.

(Includes  @QuLogic's stub fuzzer to avoid https://github.com/tinygo-org/tinygo/issues/2616;
one of the commits in this branch addresses an other little problem that come up along the way to that failure.)


dgryski:
This PR passes the test corpus, which is a nice vote of confidence.

dgryski:
Just realized my `interface{} -> any` rewrite patch should check that we're running 1.18 or above before doing the replacement.

dkegel-fastly:
OK, the interface{} -> any rewrite patch  is now versioned as you suggested.

aykevl:
IMHO it's not necessary to keep testing Go 1.17 on Windows or MacOS, unless you think there is a risk of regression that isn't caught by existing Linux/Go1.17 tests.

dkegel-fastly:
Yeah, I only started doing the 1.17 builds again when I saw 1.18 failed on windows, just wanted to see if 1.17 still worked there (it does).  I was going to remove it once 1.18 works on windows.

deadprogram:
@dkegel-fastly can you please resolve the merge conflict and rebase against `dev` so we can see what is still left to do.

deadprogram:
So looks like just Windows 1.18 not passing, if I am reading the output correctly.

dkegel-fastly:
Yup.

Tests fail on windows with 1.18.
It seems the compiler crashes when building compress/flate and index/suffixarray?
Error message may vary, latest is
```
runtime: VirtualAlloc of 802816 bytes failed with errno=1455
fatal error: out of memory
```
https://github.com/tinygo-org/tinygo/runs/5871904692?check_suite_focus=true

dkegel-fastly:
Because windows builds are failing with go 1.18, but people want to start using go 1.18, I removed
the CI changes; when this lands, motivated early adopters will be able to build dev with go 1.18
and help find problems.  I can open another PR for the CI changes...

deadprogram:
Seems like a practical approach to me. @aykevl any comment before merge?

deadprogram (approved):

deadprogram:
Let's goooooooooooooooooooo! Merging, thank you @dkegel-fastly

aykevl:
A bit late, but I think it's fine to merge. However, I do think we should at least try to fix the Windows issue before release or a lot of people will be very confused.

deadprogram:
Agreed for sure we need to fix it. I think @dkegel-fastly will now create a PR just for the CI changes so we can make sure all platforms are working. This PR was just to unblock people on other platforms from making a build from `dev` and having it work for their testing purposes.

aykevl:
Hmm, `GOOS=windows tinygo test compress/flate` works on my Linux system (with wine). That's unfortunate. Looks like it needs an actual Windows system.

dkegel-fastly:
Yup.  Doesn't reproduce on wine here, either.

dkegel-fastly:
The abort is in the compiler, and would presumably show up if you could run the windows version of go in wine, but wine can't quite handle that yet.

dkegel-fastly:
I just tried running the compiler under wine, roughly:
```
$ wine64 msiexec /i go1.18.windows-amd64.msi
$ cd ~/.wine/drive_c
$ mkdir -p users/dkegel/go
$ export GOPATH='c:\users\dkegel\go'
$ cd "Program Files"/go/bin
$ wine64 go.exe test testing
...
0134:fixme:process:SetProcessPriorityBoost (FFFFFFFFFFFFFFFF,1): stub
ok      testing 7.645s

$ wine64 go.exe test compress/flate
...
write C:\users\dkegel\Temp\go-build1476500739\b023\vet.cfg: Invalid handle.
...
--- PASS: Example_synchronization (0.00s)
PASS
ok      compress/flate  6.089s
FAIL
04f4:fixme:file:NtLockFile I/O completion on lock not implemented yet
$ wine64 --version
wine-6.0.2
```
so while there is some problem on exit in wine, it's not the problem we're looking for, and you probably do need to run on real windows.

When running go.exe to build or test something, Mac OS 11.6.5 puts up a dialog saying delete executable or cancel about 20 times; click cancel each time to allow running a downloaded executable.  (Presumably there is some way to bless go.exe and its tools, but I haven't checked.)

dkegel-fastly:
See https://github.com/tinygo-org/tinygo/issues/2762 for the bug.

PR #2727, opened 2022-03-21
Add reflect.Value.FieldByIndexErr stub

QuLogic:


src/reflect/value.go
```diff
@@ -861,6 +861,11 @@ func (v Value) FieldByIndex(index []int) Value {
 	panic("unimplemented: (reflect.Value).FieldByIndex()")
 }
 
+// FieldByIndexErr returns the nested field corresponding to index.
+func (v Value) FieldByIndexErr(index []int) (Value, error) {
+	return Value{}, &ValueError{Method: "FieldByIndexErr"}
```
QuLogic, line 866: This is probably the wrong error to return, but importing `errors` causing an import cycle. So it's good enough for the stub to compile, but maybe you'd want a more fleshed out error.

dankegel:
What does the stub fix?

QuLogic:
This part of the smoke test in Go 1.18:
```
$ make smoketest
.../tinygo/build/tinygo version
tinygo version 0.23.0-dev-093f0fd4 linux/amd64 (using go version go1.18 and LLVM version 13.0.0)
# compile-only platform-independent examples
cd tests/text/template/smoke && /home/elliott/code/tinygo/tb/tinygo/build/tinygo test -c && rm -f smoke.test
# text/template
/usr/lib/golang/src/text/template/exec.go:638:27: receiver.FieldByIndexErr undefined (type reflect.Value has no field or method FieldByIndexErr)
FAIL
make: *** [Makefile:343: smoketest] Error 1
```

deadprogram (approved):

src/reflect/value.go
deadprogram: Agreed that it is enough for now.

Issue #2361, opened 2021-12-08
RP2040 Temperature sensor support


spearson78:
The RP2040 chip includes a built in temperature sensor. This can be read using the lower level device/rp APIs but it would be helpful for newcomers to have access to the temperature sensor either directly in the machine package or the drivers package.

Here is the code I have been using in my project.

```
import (
	"device/rp"
)
func InitTempSensor() {

	rp.ADC.CS.SetBits(rp.ADC_CS_TS_EN)
}

func ReadTemp() float32 {
	rp.ADC.CS.SetBits(uint32(4) << rp.ADC_CS_AINSEL_Pos)
	rp.ADC.CS.SetBits(rp.ADC_CS_START_ONCE)

	for !rp.ADC.CS.HasBits(rp.ADC_CS_READY) {
	}

	// rp2040 uses 12-bit sampling, so scale to 16-bit
	raw := uint16(rp.ADC.RESULT.Get())

	conversion_factor := float32(3.3) / float32(1<<12)
	result := float32(raw) * conversion_factor
	temp := float32(27) - (result-float32(0.706))/float32(0.001721)

	return temp
}
```

deadprogram:
I would think that you should be able to use the exist ADC support in the RP2040 combined with the thermistor driver? I have not looked too closely, but see https://github.com/tinygo-org/drivers/blob/release/thermistor/thermistor.go

spearson78:
Thanks, I'll take a look.

spearson78:
The machine and thermistor packages assume that ADC is always performed through a machine.Pin. This is not the the case for the rp2040 temperature sensor it is an additional ADC channel. I could create a fake Pin for the RP2040 but any attempt to use this pin as a standard GPIO pin would either need to be protected against which would make all GPIO slower just for this special case or if we omit the checks would lead to undefined behavior.

I'm not sure that the assumption that ADC channels are always bound to a machine.Pin will hold true for all microcontrollers especially for things like internal temperature sensors.

I think even the Arduino has an internal temp sensor not connected to a pin, https://microchipdeveloper.com/8avr:avrtemp however it seems here that the Arduino concept of analog pins doesn't work and the ADC channel has to be read using lower level code. https://forum.arduino.cc/t/using-the-internal-temperature-sensor/8201

The options I see are to either

[...]

PR #2737, opened 2022-03-27
Pico adc input ch support

lincolngill:
Hi, This request adds ADC input channel support to the rp2040 ADC. This allows access to the ADC peripheral via the input mux channels as well as retaining the current access via machine.ADC pins. The enhancement includes:
1) Can now access the internal temperature sensor via channel 4. Which is not associated with a pin.
2) Fixes a bug. Could not use multiple ADC pins/channels before because the CS.AINSEL register bits where only getting set, not replaced. 
3) Added a Mutex to make it safe to read multiple ADC channels concurrently.
4) Added adc_rp2040 example to show off new features.

### DIFF
--- a/src/machine/machine_rp2040_adc.go
+++ b/src/machine/machine_rp2040_adc.go
@@ -39,11 +39,12 @@ func InitADC() {
 }
 
 // Configure sets the ADC pin to analog input mode.
-// Does nothing if the GPIO is not an ADC muxed pin.
-func (a ADC) Configure(config ADCConfig) {
-	if c, err := a.GetADCChannel(); err == nil {
-		c.Configure(config)
+func (a ADC) Configure(config ADCConfig) error {
+	c, err := a.GetADCChannel()
+	if err != nil {
+		return err
 	}
+	return c.Configure(config)
 }
 
 // Get returns a one-shot ADC sample reading.
@@ -75,7 +76,7 @@ func (a ADC) GetADCChannel() (c ADCChannel, err error) {
 
 // Configure sets the channel's associated pin to analog input mode or powers on the temperature sensor for ADC_TEMP_SENSOR.
 // The powered on temperature sensor increases ADC_AVDD current by approximately 40 μA.
-func (c ADCChannel) Configure(config ADCConfig) {
+func (c ADCChannel) Configure(config ADCConfig) error {
 	if config.Reference != 0 {
 		adcAref = config.Reference
 	}
@@ -86,6 +87,7 @@ func (c ADCChannel) Configure(config ADCConfig) {
 		// Enable temperature sensor bias source
 		rp.ADC.CS.SetBits(rp.ADC_CS_TS_EN)
 	}
+	return nil
 }
 
 // getOnce returns a one-shot ADC sample reading from an ADC channel.
--- a/src/machine/machine_rp2040_adc.go
+++ b/src/machine/machine_rp2040_adc.go
@@ -5,10 +5,29 @@ package machine
 
 import (
 	"device/rp"
+	"sync"
 )
 
+// ADC peripheral mux channel. 0-4.
+// Use channel constants to access valid channels.
+type ADCChannel uint8
+
+const (
+	ADC0_CH ADCChannel = iota
+	ADC1_CH
+	ADC2_CH
+	ADC3_CH     // Note: GPIO29 not broken out on pico board
+	ADC_TEMP_CH // Internal temperature sensor
+)
+
+// Will be needed for round robin config when implemented
+type ADCChConfig struct{}
+
+// Used to serialise ADC sampling
+var adcLock sync.Mutex
+
+// Reset the ADC peripheral.
 func InitADC() {
-	// reset ADC
 	rp.RESETS.RESET.SetBits(rp.RESETS_RESET_ADC)
 	rp.RESETS.RESET.ClearBits(rp.RESETS_RESET_ADC)
 	for !rp.RESETS.RESET_DONE.HasBits(rp.RESETS_RESET_ADC) {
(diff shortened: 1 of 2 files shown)


lincolngill:
Should address #2361

src/machine/machine_rp2040_adc.go
```diff
@@ -20,43 +39,112 @@ func InitADC() {
...
+		p.Configure(PinConfig{Mode: PinAnalog})
+	}
+	if c == ADC_TEMP_CH {
+		// Enable temperature sensor bias source
+		rp.ADC.CS.SetBits(rp.ADC_CS_TS_EN)
+	}
+}
+
+// The GetOnce method returns a one-shot ADC sample reading
+func (c ADCChannel) GetOnce() uint16 {
```
deadprogram: Not sure about exposing this as public function, due to not existing on other boards. Perhaps for now it should be internal package function only?

src/machine/machine_rp2040_adc.go
```diff
@@ -20,43 +39,112 @@ func InitADC() {
...
+	adcLock.Unlock()
+
+	// rp2040 is a 12-bit ADC, scale raw reading to 16-bits.
+	return uint16(rp.ADC.RESULT.Get()) << 4
+}
 
-	// rp2040 uses 12-bit sampling, so scale to 16-bit
-	return uint16(rp.ADC.RESULT.Get() << 4)
+// The GetVoltage method does a one-shot sample and returns the reading as a voltage between 0.0 and 3.3V
+func (c ADCChannel) GetVoltage() (volts float32) {
```
deadprogram: This has been left out on other platforms for various reasons, such as having different analog reference voltages on some, and also avoiding use of floats. Perhaps it should be removed for now?

src/machine/machine_rp2040_adc.go
```diff
@@ -20,43 +39,112 @@ func InitADC() {
...
+	return float32(c.GetOnce()>>4) * float32(3.3) / float32(1<<12)
 }
 
+// The GetTemp method does a one-shot sample of the internal temperature sensor.
+// Only works on the ADC_TEMP_CH channel. aka AINSEL=4. Other channels will return 0.0.
+func (c ADCChannel) GetTemp() (celsius float32) {
```
deadprogram: It would be good if this function matched the name/signature of other temperature functions such as those in the drivers repo. See https://github.com/tinygo-org/drivers/blob/5741ceb9d179dab8b88e290fda2ad5486345303c/mag3110/mag3110.go#L52

deadprogram:
Hello @lincolngill thank you for working on this code contribution. I made a few comments, please see what you think.

src/machine/machine_rp2040_adc.go
```diff
@@ -20,43 +39,112 @@ func InitADC() {
 	waitForReady()
 }
 
-// Configure configures a ADC pin to be able to be used to read data.
+// Below methods are for standard tinygo, ADC{} pin based access to the ADC peripheral.
```
soypat: The comment on this function should follow standard [Go documentation convention](https://gist.github.com/adamveld12/c0d9f0d5f0e1fba1e551#doc-comments)

src/machine/machine_rp2040_adc.go
```diff
@@ -20,43 +39,112 @@ func InitADC() {
...
+	if p, ok := c.Pin(); ok {
+		p.Configure(PinConfig{Mode: PinAnalog})
+	}
+	if c == ADC_TEMP_CH {
+		// Enable temperature sensor bias source
+		rp.ADC.CS.SetBits(rp.ADC_CS_TS_EN)
+	}
+}
+
+// The GetOnce method returns a one-shot ADC sample reading
```
soypat: Same as above https://gist.github.com/adamveld12/c0d9f0d5f0e1fba1e551#doc-comments. This comment could be rewritten as:
`// GetOnce returns a one-shot ADC sample reading`
Be good to also fix the other function/methods comments

src/machine/machine_rp2040_adc.go
soypat: @lincolngill  I can try to add these methods with `int32` return values as millicelsius and microvolts after this PR comes through if you'd prefer!

src/machine/machine_rp2040_adc.go
```diff
@@ -20,43 +39,112 @@ func InitADC() {
...
-		return 0
-	case ADC1:
-		return 1
-	case ADC2:
-		return 2
-	case ADC3:
-		return 3
+// The Pin method returns the GPIO Pin associated with the ADC mux channel, if it has one.
+// Returns !ok if the channel does not have an associated GPIO, or if the channel number is invalid.
+func (c ADCChannel) Pin() (p Pin, ok bool) {
```
soypat: Same as above, probably better to return an error? i.e:
`var errBadPinValue = errors.New("no associated pin with value")`

src/machine/machine_rp2040_adc.go
```diff
@@ -20,43 +39,112 @@ func InitADC() {
...
+	if c, ok := a.GetADCChannel(); ok {
+		return c.GetOnce()
+	}
+	// Not an ADC pin!
+	return 0
+}
+
+// The GetADCChannel method returns the channel associated with the ADC pin.
+// Returns !ok if the pin is not an ADC pin
+func (a ADC) GetADCChannel() (c ADCChannel, ok bool) {
```
soypat: Probably cleaner to return a package level error with precise information of what went wrong instead of `ok bool`.

`var errBadPinValue = errors.New("no associated pin with value")`

src/machine/machine_rp2040_adc.go
```diff
@@ -20,43 +39,112 @@ func InitADC() {
 	waitForReady()
 }
 
-// Configure configures a ADC pin to be able to be used to read data.
+// Below methods are for standard tinygo, ADC{} pin based access to the ADC peripheral.
+
+// The Configure method sets the ADC pin to analog input mode.
+// Note the ADCConfig parameter is ignored.
+// Does nothing if the GPIO is not an ADC muxed pin.
 func (a ADC) Configure(config ADCConfig) {
```
soypat: Why do we have two Configure methods for the same pin (or so it seems from what I can tell)? It seems each pin is configured individually and that each ADC contains a single corresponding Pin? Can't we simplify this to a single `ADCConfig` type which specifies all of the channel configuration- or am I missing something?

soypat (changes requested):

soypat:
@lincolngill Thanks for the contribution :heart:  It really is great to see contributions coming into the rp2040 implementation!
I made a few comments about some little things, let me know of any doubts or any help you need with any of my suggestions!

src/machine/machine_rp2040_adc.go
lincolngill: Will change to private. This was a way to get the raw temperature sensor ADC reading if using an external ref voltage less than 3.3V. I'll update code to use ADCConfig.Reference (if set) when calculating the voltage and temperature.

src/machine/machine_rp2040_adc.go
lincolngill: Will change to integer arithmetic and embed in ReadTemperature()

src/machine/machine_rp2040_adc.go
lincolngill: Based on the comments. It might be best if I rename ADC_TEMP_CH to TEMP_SENSOR and make it the only public channel. This ensures the pin based channels are only accessible via the standard ADC type methods. And the temperature sensor would be accessed like:
```
ts := machine.TEMP_SENSOR
ts.Configure(machine.ADCConfig{})
milliDegrees int32, err := ts.ReadTemperature()
```
Does that sound alright?
The alternative is to write a separate device driver for the temperature sensor but IMO it's better to just treat it as another ADC channel. The ADC peripheral doesn't distinguish it from the other input channels and it will be simpler to implement the more advanced ADC features if it's not treated as a separate device.

src/machine/machine_rp2040_adc.go
lincolngill: Will tidy up.

src/machine/machine_rp2040_adc.go
lincolngill: Yep using ADCConfig makes sense. I was originally thinking the channel config might be quite specific to rp2040 when advanced ADC features get implemented in the future. It's a mute point, Currently no config attributes are used. I'll get rid of the ADCChConfig type.

lincolngill: Configuring each individual ADC pin with ADCConfig{} is existing behaviour. Seems a bit odd to me too. The ADC H/W abstraction seems to focus on the input pins rather than a configurable peripheral, with configurable inputs.

src/machine/machine_rp2040_adc.go
deadprogram: agree with @soypat return better to return `error`

src/machine/machine_rp2040_adc.go
deadprogram: That makes sense. No need for separate driver, my point was to have the same interface.

src/machine/machine_rp2040_adc.go
soypat: Maybe having separate ADCs that makes sense? If each input pin can be configured independently it would make sense to have a separate ADC type for each pin. 

I just gave a _very_ cursory glance at the datasheet- it really seems like there is no independent configuring of pins, but I could be wrong.

lincolngill: @soypat Maybe I'm missing something. The definition of ADC type is in machine.go as
 `
type ADC struct {
	Pin Pin
}`
So the stock adc.go example would suggest each ADC pin receives an ADCConfig. This is also how the other boards are implemented. Not sure how to do a single config without diverging from the standard ADC API. Any suggestions?

The rp2040 doesn't currently use any ADCConfig fields but I plan to use the ADCConfig.Reference voltage in the temperature calc.

I'm going to make only the ADC_TEMP_SENSOR channel public and change it's Configure method to use ADCConfig. The pin based inputs will just have their public ADC methods. Like now.

soypat: Oh snap, nevermind what I was nagging about then. If that's how other boards do it then so it shall be.

deadprogram: OK, so nothing further required here on this point.

deadprogram:
@lincolngill likewise really happy to see this in progress. Seems like just a few small things to get it merged. Please let me know if  I can do anything to help out.

deadprogram:
We are getting ready for a new release soonish. It would be great to be able to land this PR in it, anything I can do to help?

lincolngill:
> We are getting ready for a new release soonish. It would be great to be able to land this PR in it, anything I can do to help?

Latest commit should resolve all previous PR comments and suggestions.

deadprogram:
@soypat any other feedback here before we merge?

soypat:
LGTM :+1: 

Edit: just noticed the `Configure` methods don't return an error. Would it not be wise to stick to the convention of usually returning an error on Configure, even if it is nil? Thinking of interfaces and such

deadprogram:
> Would it not be wise to stick to the convention of usually returning an error on Configure

Yes. Good catch. Can you please update this one last time @lincolngill ?

lincolngill:
> > Would it not be wise to stick to the convention of usually returning an error on Configure
> 
> Yes. Good catch. Can you please update this one last time @lincolngill ?

Configure() for ADC and ADCChannel now return an error. But a search shows that machine_rp2040_adc.go is now the only board that does this
Screenshot from 2022-04-26 19-36-43
!

lincolngill:
Also spotted and fixed a a bug in InitADC. Wasn't resetting the ADC ref voltage value.

deadprogram:
Thank you for all of the great work on this PR @lincolngill and thanks to @soypat for review.

Now merging!

Issue #2259, opened 2021-11-13
machine/rp2040: I2C not ACKing upon succesful read transaction


soypat:
RP2040 Transaction Write to *Address 0x40*, @ 0x00, and Read  @ 0x21
![image](https://user-images.githubusercontent.com/26156425/141661304-47b6683a-8cfb-4c6c-99cd-180047f3a21d.png)

Same transaction on Arduino Mega 2560:
![image](https://user-images.githubusercontent.com/26156425/141661341-b483e6e1-2b25-4980-8ce1-35a3cdf7b292.png)

Notice last ACK bit is set on the Mega, not so on RP2040

deadprogram:
Is anyone looking into this?

soypat:
@ysoldak This may interest you.
I think I may have found the root cause. See [`func (i2c) Tx()`](https://github.com/tinygo-org/tinygo/blob/a5d905f19ba5008a88d4594bfa816ed6de1df7d0/src/machine/machine_rp2040_i2c.go#L77), this line should be changed to 
```go
if err := i2c.tx(uint8(addr), w, len(r) > 0, timeout); nil != err {
```
So that `nostop=true`, thus stop condition is not issued if there is a transaction pending on the read buffer.

Making the change breaks [`disable()`](https://github.com/tinygo-org/tinygo/blob/a5d905f19ba5008a88d4594bfa816ed6de1df7d0/src/machine/machine_rp2040_i2c.go#L183), presumably mid-transaction during `rx` call. The best approach would seem to be to combine `tx` and `rx` functions into a single function with two loops which disables the peripheral a single time during the transaction. I'll see if I fix it these coming days.

ysoldak:
@soypat thanks for looking at it, happy to help testing

soypat:
Alright, here's the first revision. seems to be working and I can confirm the spurious STOP condition when a read is pending is gone.
<details>

```go
func (i2c *I2C) Tx(addr uint16, w, r []byte) error {
	return i2c.txrx(uint8(addr), w, r)
}

func (i2c *I2C) txrx(addr uint8, tx, rx []byte) (err error) {
        // timeout in microseconds.
	const timeout = 40 * 1000 // 40ms is a reasonable time for a real-time system.
	deadline := ticks() + timeout
	if addr >= 0x80 || isReservedI2CAddr(addr) {
		return ErrInvalidTgtAddr
	}
	txlen := len(tx)
	rxlen := len(rx)
[...]

PR #2743, opened 2022-04-01
rp2040: fix spurious i2c STOP during write+read transaction

soypat:
@ysoldak @deadprogram  PTAL. This fixes a subtle bug which causes a STOP condition to be issued between write/read operation. It is does not solve https://github.com/tinygo-org/tinygo/issues/2259 but is somewhat related: it does fix a NACK I was seeing in single byte write transactions too (read transaction NACK is still present on last byte read). The spurious STOP condition is visible in the image in the link on the pico, not so with the arduino. 

It would be great to get some testing feedback! It has worked great with my l3gd20 gyro so far!


deadprogram (approved):

deadprogram:
Thanks @soypat now merging.

PR #2744, opened 2022-04-02
fix: flash-command for maixbit board

saisasidhar:
Latest version of kflash tool requires a --Board/-B argument to be able to run the tool

### DIFF
--- a/targets/maixbit.json
+++ b/targets/maixbit.json
@@ -3,5 +3,5 @@
 	"build-tags": ["maixbit"],
 	"serial": "uart",
 	"linkerscript": "targets/maixbit.ld",
-	"flash-command": "kflash -p {port} --noansi --verbose {bin}"
+	"flash-command": "kflash --Board bit -p {port} --noansi --verbose {bin}"
 }


deadprogram (approved):

deadprogram:
Thanks @saisasidhar now merging.

deadprogram:
Hello @saisasidhar which version of kflash is it that you say requires that argument? I notice that with this change I can no longer flash my Maixbit with `goE` firmware option. The autodetect IS working for me with kflash 0.8.3 which is the version installed by running

```
sudo pip3 install kflash
```

I would for sure prefer autodetect if possible. What do you think?

saisasidhar:
Hello @deadprogram, I picked up the [release .exe file from Sipeed/kflash.py](https://github.com/sipeed/kflash.py/releases/tag/v1.0.0). 

Just saw that this 1.0.0 exe release is 91 commits behind master on sipeed/kflash.py and one of those commits includes a change from `required=True` to `required=False` on the `--Board` argument. 

Sorry for creating the confusion, I was using a release from early 2019 ! and releases on pypi are much more recent. Please feel free to revert my commit.

deadprogram:
Thanks @saisasidhar for looking into it. Reverted earlier.

PR #2746, opened 2022-04-05
m5stamp-c3: add pin setting of UART

sago35:
It is better to have UART_TX_PIN defined in order to use `src/examples/echo`.

Unlike other ESP32C3 boards, the m5stamp has its UART pins connected to the Type-C connector via a serial conversion chip.
For this reason, it is very useful to define UART_TX_PIN.

https://docs.m5stack.com/en/core/stamp_c3


deadprogram (approved):

deadprogram:

PR #2748, opened 2022-04-05
feat: add basic WASI `net.Listener` via `sock_accept`

rvolosatovs:
# Description
Add support for `sock_accept` recently added to WASI https://github.com/WebAssembly/WASI/pull/458

The implementation still lacks polling support, but that is something best left for another PR, since that would be a magnitudes bigger change.

Relevant documentation on flags can be found here https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#fdflags

I followed the upstream Go implementation as much as possible with minimal changes

# Testing

## Server
Save the following program to `main.go` in your working directory:
```go
package main

import (
	"fmt"
	"io"
	"log"
	"net"
	"os"
	"strconv"
	"time"
)

func handle(conn net.Conn) error {
	defer conn.Close()

	var b [128]byte
	n, err := conn.Read(b[:])
	if err != nil {
		if err == io.EOF {
			return nil
		}
		return fmt.Errorf("failed to read from connection: %s", err)
	}
	log.Printf("Read '%s'\n", b[:n])

	res := fmt.Sprintf("Hello, %s", b[:n])
	n, err = conn.Write([]byte(res))
	if err != nil {
		return fmt.Errorf("failed to write to connection: %s", err)
	}
	log.Printf("Wrote '%s'\n", res[:n])
	return nil
}

func run() error {
	fds, err := strconv.Atoi(os.Getenv("FD_COUNT"))
	if err != nil {
		return fmt.Errorf("failed to parse FD_COUNT: %w", err)
	}
	if fds != 4 {
		return fmt.Errorf("FD_COUNT must be 4, got %d", fds)
	}

	ln, err := net.FileListener(os.NewFile(uintptr(3), "socket"))
	if err != nil {
		return fmt.Errorf("failed to listen on fd 3: %w", err)
	}

	for {
		log.Println("Waiting for connection...")
		conn, err := ln.Accept()
		if err != nil {
			if err, ok := err.(*os.SyscallError); ok && err.Timeout() {
				log.Println("Accept timed out, retrying in a second...")
				time.Sleep(time.Second)
				continue
			}
			return fmt.Errorf("failed to accept connection: %w", err)
		}
		log.Println("Accepted connection")

		if err := handle(conn); err != nil {
			return fmt.Errorf("failed to handle connection: %w", err)
		}
		log.Println("---")
	}
}

func init() {
	log.SetOutput(os.Stderr)
	log.SetFlags(0)
}

func main() {
	if err := run(); err != nil {
		log.Fatalf("Failed to run: %s", err)
	}
}
```

Compile with:
```
$ tinygo build -target wasi main.go
```

### Wasmtime
Works on `wasmtime 0.38.0`:
```
$ wasmtime run --env FD_COUNT=4 --tcplisten 127.0.0.1:9000 main.wasm
Waiting for connection...
Accept timed out, retrying in a millisecond...
Waiting for connection...
Accept timed out, retrying in a millisecond...
Waiting for connection...
Accept timed out, retrying in a millisecond...
Waiting for connection...
Accepted connection
Read 'test
'
Wrote 'Hello, test
'
---
```

### Enarx
Works on https://github.com/enarx/enarx with the following config:
```toml
[[files]]
kind = "stdin"

[[files]]
kind = "stdout"

[[files]]
kind = "stderr"

[[files]]
kind = "listen"
prot = "tcp"
port = 9000
```

For example, on Linux:
```
$ curl -sL -o enarx https://github.com/enarx/enarx/releases/download/v0.6.0/enarx-x86_64-unknown-linux-musl
$ echo 26823747c19c21fc7c5a7a6d44337d731735c780044b9da90365e63aae15f68d enarx | sha256sum -c
$ chmod +x enarx
$ ./enarx run --wasmcfgfile Enarx.toml main.wasm
Waiting for connection...
Accepted connection
Read 'test
'
Wrote 'Hello, test
'
---
```

## Client
Start in a terminal window:
```
while :; do sleep 1 && echo test | nc 127.0.0.1 9000; done
Hello, test
```

### DIFF
--- a/src/net/fd_wasi.go
+++ b/src/net/fd_wasi.go
@@ -1,0 +1,57 @@
+//go:build wasi
+// +build wasi
+
+package net
+
+import (
+	"io"
+	"os"
+	"syscall"
+	"time"
+)
+
+// Network file descriptor.
+type netFD struct {
+	*os.File
+
+	net   string
+	laddr Addr
+	raddr Addr
+}
+
+// Read implements the Conn Read method.
+func (c *netFD) Read(b []byte) (int, error) {
+    // TODO: Handle EAGAIN and perform poll
+	return c.File.Read(b)
+}
+
+// Write implements the Conn Write method.
+func (c *netFD) Write(b []byte) (int, error) {
+    // TODO: Handle EAGAIN and perform poll
+	return c.File.Write(b)
+}
+
+// LocalAddr implements the Conn LocalAddr method.
+func (*netFD) LocalAddr() Addr {
+	return nil
+}
+
+// RemoteAddr implements the Conn RemoteAddr method.
+func (*netFD) RemoteAddr() Addr {
+	return nil
+}
+
+// SetDeadline implements the Conn SetDeadline method.
+func (*netFD) SetDeadline(t time.Time) error {
+	return ErrNotImplemented
+}
+
+// SetReadDeadline implements the Conn SetReadDeadline method.
+func (*netFD) SetReadDeadline(t time.Time) error {
+	return ErrNotImplemented
+}
+
+// SetWriteDeadline implements the Conn SetWriteDeadline method.
+func (*netFD) SetWriteDeadline(t time.Time) error {
+	return ErrNotImplemented
+}
--- a/src/net/fd_wasi.go
+++ b/src/net/fd_wasi.go
@@ -4,7 +4,6 @@
 package net
 
 import (
-	"io"
 	"os"
 	"syscall"
 	"time"
@@ -21,13 +20,13 @@ type netFD struct {
 
 // Read implements the Conn Read method.
 func (c *netFD) Read(b []byte) (int, error) {
-    // TODO: Handle EAGAIN and perform poll
+	// TODO: Handle EAGAIN and perform poll
 	return c.File.Read(b)
 }
 
 // Write implements the Conn Write method.
 func (c *netFD) Write(b []byte) (int, error) {
-    // TODO: Handle EAGAIN and perform poll
+	// TODO: Handle EAGAIN and perform poll
 	return c.File.Write(b)
 }
 
@@ -54,4 +53,26 @@ func (*netFD) SetReadDeadline(t time.Time) error {
 // SetWriteDeadline implements the Conn SetWriteDeadline method.
 func (*netFD) SetWriteDeadline(t time.Time) error {
 	return ErrNotImplemented
+}
+
+type listener struct {
+	*os.File
+}
+
+// Accept implements the Listener Accept method.
+func (l *listener) Accept() (Conn, error) {
+	fd, err := syscall.SockAccept(int(l.File.Fd()), syscall.O_NONBLOCK)
+	if err != nil {
+		return nil, wrapSyscallError("sock_accept", err)
+	}
+	return &netFD{File: os.NewFile(uintptr(fd), "conn"), net: "file+net", laddr: nil, raddr: nil}, nil
+}
+
+// Addr implements the Listener Addr method.
+func (*listener) Addr() Addr {
+	return nil
+}
+
+func fileListener(f *os.File) (Listener, error) {
+	return &listener{File: f}, nil
 }
--- a/src/syscall/syscall_libc_wasi.go
+++ b/src/syscall/syscall_libc_wasi.go
@@ -309,6 +309,14 @@ func Getpagesize() int {
 	return 65536
 }
 
+func SockAccept(fd int, flags int) (nfd int, err error) {
+	var retptr0 __wasi_fd_t
+	if n := sock_accept(__wasi_fd_t(fd), __wasi_fdflags_t(flags), &retptr0); n != 0 {
+		return -1, Errno(n)
+	}
+	return int(retptr0), nil
+}
+
 // int stat(const char *path, struct stat * buf);
 //
 //export stat
@@ -323,3 +331,14 @@ func libc_fstat(fd int32, ptr unsafe.Pointer) int32
 //
 //export lstat
 func libc_lstat(pathname *byte, ptr unsafe.Pointer) int32
+
+type (
+	__wasi_errno_t = uint16
+
+	__wasi_fd_t      = int32
+	__wasi_fdflags_t = uint16
+)
+
+//go:wasm-module wasi_snapshot_preview1
+//export sock_accept
+func sock_accept(fd __wasi_fd_t, flags __wasi_fdflags_t, retptr0 *__wasi_fd_t) __wasi_errno_t
--- a/src/net/error_posix.go
+++ b/src/net/error_posix.go
@@ -1,0 +1,24 @@
+// The following is copied and adapted from Go 1.18 official implementation.
+
+// Copyright 2017 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build aix || darwin || dragonfly || freebsd || (js && wasm) || linux || netbsd || openbsd || solaris || windows || wasi
+// +build aix darwin dragonfly freebsd js,wasm linux netbsd openbsd solaris windows wasi
+
+package net
+
+import (
+	"os"
+	"syscall"
+)
+
+// wrapSyscallError takes an error and a syscall name. If the error is
+// a syscall.Errno, it wraps it in a os.SyscallError using the syscall name.
+func wrapSyscallError(name string, err error) error {
+	if _, ok := err.(syscall.Errno); ok {
+		err = os.NewSyscallError(name, err)
+	}
+	return err
+}
--- a/src/net/file.go
+++ b/src/net/file.go
@@ -1,0 +1,31 @@
+//go:build wasi
+// +build wasi
+
+// The following is copied from Go 1.18 official implementation.
+
+// Copyright 2015 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package net
+
+import (
+	"os"
+)
+
+type fileAddr string
+
+func (fileAddr) Network() string  { return "file+net" }
+func (f fileAddr) String() string { return string(f) }
+
+// FileListener returns a copy of the network listener corresponding
+// to the open file f.
+// It is the caller's responsibility to close ln when finished.
+// Closing ln does not affect f, and closing f does not affect ln.
+func FileListener(f *os.File) (ln Listener, err error) {
+	ln, err = fileListener(f)
+	if err != nil {
+		err = &OpError{Op: "file", Net: "file+net", Source: nil, Addr: fileAddr(f.Name()), Err: err}
+	}
+	return
+}
--- a/src/net/net.go
+++ b/src/net/net.go
@@ -1,4 +1,4 @@
-// The following is copied from Go 1.18 official implementation.
+// The following is copied and adapted from Go 1.18 official implementation.
 
 // Copyright 2009 The Go Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
@@ -8,6 +8,7 @@ package net
 
 import (
 	"io"
+	"syscall"
 	"time"
 )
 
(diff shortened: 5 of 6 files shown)


deadprogram:
Hello @rvolosatovs could you please rebase against the latest `dev` branch so this can be more easily reviewed?

cc/ @dkegel-fastly @dgryski @aykevl

npmccallum:
@rvolosatovs Can you rebase this?

rvolosatovs:
@deadprogram apologies for the delay, I was on vacation right after we spoke and then had a few very busy weeks.
I've rebased, updated the implementation and added an example for testing the PR is ready for review, PTAL!
cc @dkegel-fastly @dgryski @aykevl

There is a CI build failure on MacOS, but it does not seem to be caused by this PR - I may be wrong however

npmccallum:
@deadprogram Are you able to review this?

rvolosatovs:
I've extracted some of the required changes into smaller scoped PRs in hopes of a simpler review.
@dkegel-fastly @dgryski @aykevl @deadprogram 

This PR is now blocked by:
https://github.com/tinygo-org/tinygo/pull/3051
https://github.com/tinygo-org/tinygo/pull/3050
https://github.com/tinygo-org/tinygo/pull/3052
https://github.com/tinygo-org/tinygo/pull/3053
https://github.com/tinygo-org/tinygo/pull/3054

rvolosatovs:
@deadprogram would further splitting this PR into smaller scoped chunks assist review?

rvolosatovs:
Filed #3061 , marking as draft again

codefromthecrypt:
@rvolosatovs interesting (esp that the snapshot01 version was updated after the fact!).

I was curious the motivation of this? Is this a proof of concept or are folks making raw socket listeners? Reason I ask is often there are higher level libs that need some changes to support even if sock_accept works.

Finally, I noticed in the past due to lack of description in the spec sometimes you need to look at another impl to see how they handle things. Did you notice how this is implemented in rust or clang?? Ex the resulting wasm (can inspect backwards via wasm2wat) looks similar between here and rust or C of the same?

Cheers!

src/net/fd_wasi.go
```diff
@@ -0,0 +1,78 @@
...
+
+// Write implements the Conn Write method.
+func (c *netFD) Write(b []byte) (int, error) {
+	// TODO: Handle EAGAIN and perform poll
+	return c.File.Write(b)
+}
+
+// LocalAddr implements the Conn LocalAddr method.
+func (*netFD) LocalAddr() Addr {
+	return nil
```
codefromthecrypt, line 35:
```suggestion
	return laddr
```

codefromthecrypt, line 35:
```suggestion
	return laddr
```

src/net/fd_wasi.go
```diff
@@ -0,0 +1,78 @@
...
+	return ErrNotImplemented
+}
+
+type listener struct {
+	*os.File
+}
+
+// Accept implements the Listener Accept method.
+func (l *listener) Accept() (Conn, error) {
+	fd, err := syscall.SockAccept(int(l.File.Fd()), syscall.O_NONBLOCK)
```
codefromthecrypt, line 64:
```suggestion
	// Note: WASI doesn't define SOCK_CLOEXEC
	fd, err := syscall.SockAccept(int(l.File.Fd()), syscall.O_NONBLOCK)
```

src/net/fd_wasi.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return nil, wrapSyscallError("sock_accept", err)
+	}
+	return &netFD{File: os.NewFile(uintptr(fd), "conn"), net: "file+net", laddr: nil, raddr: nil}, nil
```
codefromthecrypt, line 68: maybe a fake addr instead of nil? or comment why? I think usually these are derived from fd, right?

src/net/fd_wasi.go
```diff
@@ -0,0 +1,78 @@
...
+func (*netFD) SetReadDeadline(t time.Time) error {
+	return ErrNotImplemented
+}
+
+// SetWriteDeadline implements the Conn SetWriteDeadline method.
+func (*netFD) SetWriteDeadline(t time.Time) error {
+	return ErrNotImplemented
+}
+
+type listener struct {
```
codefromthecrypt, line 58:
```suggestion
// wasiListener implements Listener with syscall.SockAccept.
type wasiListener struct {
```

src/net/fd_wasi.go
```diff
@@ -0,0 +1,78 @@
...
+	"os"
+	"syscall"
+	"time"
+)
+
+// Network file descriptor.
+type netFD struct {
+	*os.File
+
+	net   string
```
codefromthecrypt, line 16: unused, so set or remove?

src/syscall/syscall_libc_wasi.go
```diff
@@ -309,6 +309,14 @@ func Getpagesize() int {
 	return 65536
 }
 
+func SockAccept(fd int, flags int) (nfd int, err error) {
+	var retptr0 __wasi_fd_t
```
codefromthecrypt, line 313: verify flags or return EINVAL?

src/net/fd_wasi.go
```diff
@@ -0,0 +1,78 @@
...
+
+// Read implements the Conn Read method.
+func (c *netFD) Read(b []byte) (int, error) {
+	// TODO: Handle EAGAIN and perform poll
+	return c.File.Read(b)
+}
+
+// Write implements the Conn Write method.
+func (c *netFD) Write(b []byte) (int, error) {
+	// TODO: Handle EAGAIN and perform poll
```
codefromthecrypt, line 29:
```suggestion
	// TODO: Handle EAGAIN and perform poll similar to poll/fd_unix.go
```

src/net/fd_wasi.go
```diff
@@ -0,0 +1,78 @@
...
+	*os.File
+
+	net   string
+	laddr Addr
+	raddr Addr
+}
+
+// Read implements the Conn Read method.
+func (c *netFD) Read(b []byte) (int, error) {
+	// TODO: Handle EAGAIN and perform poll
```
codefromthecrypt, line 23:
```suggestion
	// TODO: Handle EAGAIN and perform poll similar to poll/fd_unix.go
```

src/net/fd_wasi.go
```diff
@@ -0,0 +1,78 @@
...
+package net
+
+import (
+	"os"
+	"syscall"
+	"time"
+)
+
+// Network file descriptor.
+type netFD struct {
```
codefromthecrypt, line 13:
```suggestion
type wasiFD struct {
```

src/net/fd_wasi.go
```diff
@@ -0,0 +1,78 @@
...
+
+package net
+
+import (
+	"os"
+	"syscall"
+	"time"
+)
+
+// Network file descriptor.
```
codefromthecrypt, line 12:
```suggestion
// wasiCon implements Con with a file returned by syscall.SockAccept.
```

src/net/fd_wasi.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	return &netFD{File: os.NewFile(uintptr(fd), "conn"), net: "file+net", laddr: nil, raddr: nil}, nil
+}
+
+// Addr implements the Listener Addr method.
+func (*listener) Addr() Addr {
+	return nil
+}
+
+func fileListener(f *os.File) (Listener, error) {
```
codefromthecrypt, line 76:
```suggestion
// fileListener is used by FileListener in file.go
func fileListener(f *os.File) (Listener, error) {
```

codefromthecrypt (commented):
Added some review to help this along.

I suspect when things stabilize, unit tests would make sense, though I notice Go doesn't really test these files directly either.

rvolosatovs:
> @rvolosatovs interesting (esp that the snapshot01 version was updated after the fact!).
> 
> I was curious the motivation of this? Is this a proof of concept or are folks making raw socket listeners? Reason I ask is often there are higher level libs that need some changes to support even if sock_accept works.

The motivation is to be able to accept a connection from within a WebAssembly module via WASI. In [Wasmtime](https://github.com/bytecodealliance/wasmtime) and [Enarx](https://github.com/enarx/enarx) this works by calling [`sock_accept`](https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#-sock_acceptfd-fd-flags-fdflags---resultfd-errno) on pre-opened file descriptors 

> Finally, I noticed in the past due to lack of description in the spec sometimes you need to look at another impl to see how they handle things. Did you notice how this is implemented in rust or clang?? Ex the resulting wasm (can inspect backwards via wasm2wat) looks similar between here and rust or C of the same?

Both Rust (https://github.com/rust-lang/rust/pull/93158) and C (https://github.com/WebAssembly/wasi-libc/pull/287) support was added by @haraldh, who we work together with at @profianinc. We're making sure to have all implementations consistent.

Note, that big difference of `sock_accept` from `accept` family, is that the peer address cannot be determined (at least for now) as part of the "accept".

src/net/fd_wasi.go
rvolosatovs: This is directly copy-pasted and adapted from https://github.com/golang/go/blob/a0441c7ae3dea57a0553c9ea77e184c34b7da40f/src/net/fd_posix.go#L16-L17, that's the approach I saw throughout this project so I followed that.

Changing this comment would make it inconsistent with upstream, which means it would not be as trivial to sync this with upstream by copy-pasting.

If we're changing this, then we should change the whole implementation as well, since the way this is designed in the standard library is not necessarily the best approach here given the set of use cases this project tries to cover.

src/net/fd_wasi.go
rvolosatovs: https://github.com/tinygo-org/tinygo/pull/2748/files#r979834379

src/net/fd_wasi.go
rvolosatovs: https://github.com/tinygo-org/tinygo/pull/2748/files#r979834379

src/net/fd_wasi.go
rvolosatovs: Well, not necessarily same as https://github.com/golang/go/blob/master/src/internal/poll/fd_unix.go
It's circling back to https://github.com/tinygo-org/tinygo/pull/2748/files#r979834379 in my head - is it a goal of this project to be as similar to upstream Go and minimize diff or not? A simpler and more minimal poll implementation could probably suit this project better, but I am not sure what's the preferred approach. I was hoping this PR could start a discussion

src/net/fd_wasi.go
rvolosatovs: See above

src/net/fd_wasi.go
rvolosatovs, line 35: This is not available on WASI

src/net/fd_wasi.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// LocalAddr implements the Conn LocalAddr method.
+func (*netFD) LocalAddr() Addr {
+	return nil
+}
+
+// RemoteAddr implements the Conn RemoteAddr method.
+func (*netFD) RemoteAddr() Addr {
+	return nil
```
rvolosatovs, line 40: This is not available on WASI

src/net/fd_wasi.go
rvolosatovs: None of this is available on WASI

src/net/fd_wasi.go
codefromthecrypt: ok understand what you mean. I think in another world I would probably comment things like this in the source, but ack that's not the practice. I'll dismiss my feedback as most of them are me being curious why things are defining unused symbols or otherwise. People who know the conventions already somehow won't need to see comments like this.

src/net/fd_wasi.go
rvolosatovs, line 76: Upstream Go does not do this https://github.com/golang/go/blob/a0441c7ae3dea57a0553c9ea77e184c34b7da40f/src/net/file_unix.go#L89-L102

If a structure of the code is confusing, I don't think that explaining everything in comments is the right solution - restructuring the code is, however.
As I said, I tried to keep as close as possible to upstream Go implementation - if that's not a requirement, then I'd rather rewrite this whole thing in a way that makes much more sense especially for this use case

src/net/fd_wasi.go
codefromthecrypt: I'll leave this one to the maintainers to decide. thanks for clarifying!

src/net/fd_wasi.go
codefromthecrypt: I'll delete all my comments which summarize as "please comment this" :D for now, I guess people will know that if something is nil, it is intentional.

src/net/fd_wasi.go
codefromthecrypt: Personally, I would love to see your impl, but for tinygo whatever maintainers suggest is best as they have the long term stake.

src/syscall/syscall_libc_wasi.go
rvolosatovs: Standard library does not do that for `Accept4` https://github.com/golang/go/blob/a0441c7ae3dea57a0553c9ea77e184c34b7da40f/src/syscall/syscall_linux.go#L661-L677 :thinking: 
Is it really responsibility of TinyGo standard library or rather the WASI implementation?

codefromthecrypt: WASI has no guidance on things like this. the spec is very sparse. There's no discussion of host vs guest responsibilities. knowing this, yeah most hosts will also validate and return EINVAL conventionally https://github.com/WebAssembly/WASI/blob/snapshot-01/phases/snapshot/docs.md

lacking guidance, what I do is look at what wasi-libc does and prefer that, though agree you can also look at how go does it. this isn't a huge deal which way, imho

src/net/fd_wasi.go
dkegel-fastly: To the style question, my impression is that the project tries in this order:
1. directly use upstream files (no copying needed)
2. borrow a subset of upstream files, minimizing diffs but only borrowing the lines that are needed
3. replace upstream files with something more suited to tinygo's use case (but still kind of minimizing diffs if possible)

dkegel-fastly: Upstream's networking is... tricky because of its interaction with goroutines, as you've noticed.
How useful is networking without getting that bit right?  
Will your code (e.g. netfd.Read()) pause all goroutines while blocking on networking? 
I haven't checked yet.

hlubek:
I played around with this PR and tried to use it to start an HTTP server with `net/http` over the `FileListener`. Basically my WASM/WASI code looks like this (following the example from this PR):

```go
func run() error {
	ln, err := net.FileListener(os.NewFile(uintptr(3), "socket"))
	if err != nil {
		return fmt.Errorf("failed to listen on fd 3: %w", err)
	}

	mux := http.NewServeMux()
	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		fmt.Fprintf(w, "Hello, %s", r.URL.Path[1:])
	})

	srv := &http.Server{
		Handler: mux,
	}

	err = srv.Serve(ln)
	if err != nil {
		return fmt.Errorf("failed to serve: %w", err)
	}

	return nil
}
```

It seems to work after I did the following changes:

* Added a dummy `LocalAddr` and `RemoteAddr`, otherwise the Go HTTP implementation had a nil dereference issue
* Implement a very naive polling via checking for `syscall.EAGAIN` and sleeping for a few milliseconds (in `Read`, `Write` and `Accept`):
  ```go
  func (c *netFD) Read(b []byte) (int, error) {
	var (
		n   int
		err error
	)
	for {
		n, err = c.File.Read(b)
		if errors.Is(err, syscall.EAGAIN) {
			time.Sleep(30 * time.Millisecond)
			continue
		}
		break
	}
	return n, err
  }
  ```

Doing this I could run the server via `wasmtime run --tcplisten 127.0.0.1:9000 server.wasm` and a Curl succeeded. 😁 

What would be a better way to perform the polling? Is `poll_oneoff` supposed to work for `sock_accept` and reading / writing from the file descriptor?

I could imagine that the `Listener` returned by `FileListener` should not poll by default but just return `EAGAIN` / a syscall error indicating a timeout and the user code could wrap the listener to control the polling behaviour.

It would be really great to have some kind of support of this in TinyGo, so web services with Go could be run in WASM/WASI.

codefromthecrypt:
@hlubek if were me I would probably scour GitHub for anyone using wasi-libc in C and see how they are doing it. https://github.com/WebAssembly/wasi-libc/blob/63e4489d01ad0262d995c6d9a5f1a1bab719c917/libc-bottom-half/sources/accept.c then compare polling approaches. I think these network functions are more used in rust for various docker demos.

chriso:
We're proposing this be added to Go (with `GOOS=wasip1`). See https://go-review.googlesource.com/c/go/+/493358. It'll use the native netpoll facility so you won't have to manually handle EAGAIN.

chriso:
This landed in Go via https://github.com/golang/go/commit/a17de43e (and prerequisites https://github.com/golang/go/commit/41893389 and https://github.com/golang/go/commit/c5c21845)

rvolosatovs:
> This landed in Go via https://github.com/golang/go/commit/a17de43e (and prerequisites https://github.com/golang/go/commit/41893389 and https://github.com/golang/go/commit/c5c21845)

Yes, it finally happened! Closing

codefromthecrypt:
@deadprogram I think this issue is not closed in TinyGo, even if it is solved for Go. There will be users who can't use Go, yet due to lack of exports. Should this be re-opened as an issue until closed? or this codebase repurposed to complete regardless of what Go does?

Issue #2749, opened 2022-04-05
"panic: runtime error: integer divide by zero" when make-ing an array of zero-sized elems


ALTree:
```
$ tinygo version
tinygo version 0.22.0 linux/amd64 (using go version go1.17.8 and LLVM version 13.0.0)
```

Building this program:

```
package main

import "fmt"

func main() {
	i := 4
	m := make([]struct{}, i)
	fmt.Println(m)
}
```

triggers a panic in tinygo:

```
$ tinygo build -o crash.o crash.go 

panic: runtime error: integer divide by zero

goroutine 54 [running]:
github.com/tinygo-org/tinygo/compiler.(*compilerContext).maxSliceSize(0xc000654680, {0x512df38})
	/home/runner/work/tinygo/tinygo/compiler/compiler.go:1552 +0xb5
github.com/tinygo-org/tinygo/compiler.(*builder).createExpr(0xc00113e360, {0x5134910, 0xc00136d440})
	/home/runner/work/tinygo/tinygo/compiler/compiler.go:1793 +0x472
github.com/tinygo-org/tinygo/compiler.(*builder).createInstruction(0xc00113e360, {0x51348c8, 0xc00136d440})
	/home/runner/work/tinygo/tinygo/compiler/compiler.go:1129 +0x8ea
github.com/tinygo-org/tinygo/compiler.(*builder).createFunction(0xc00113e360)
	/home/runner/work/tinygo/tinygo/compiler/compiler.go:1034 +0x1865
github.com/tinygo-org/tinygo/compiler.(*compilerContext).createPackage(0xc000654680, {0xc00161f2c0}, 0xc001305500)
	/home/runner/work/tinygo/tinygo/compiler/compiler.go:773 +0x62f
github.com/tinygo-org/tinygo/compiler.CompilePackage({0xc0003f9da0, 0xc00157ae10}, 0xc000462540, 0xc001305500, {0x5}, 0x0, 0x0)
	/home/runner/work/tinygo/tinygo/compiler/compiler.go:280 +0x377
github.com/tinygo-org/tinygo/builder.Build.func1(0x0)
	/home/runner/work/tinygo/tinygo/builder/build.go:258 +0x196
github.com/tinygo-org/tinygo/builder.runJob(0xc00133c540, 0x0)
	/home/runner/work/tinygo/tinygo/builder/jobs.go:222 +0x4f
created by github.com/tinygo-org/tinygo/builder.runJobs
[...]

PR #2750, opened 2022-04-05
compiler: allow slices of empty structs.

dkegel-fastly:
Fixes https://github.com/tinygo-org/tinygo/issues/2749

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -1554,7 +1554,11 @@ func (c *compilerContext) maxSliceSize(elementType llvm.Type) uint64 {
 	// Determine the maximum allowed size for a slice. The biggest possible
 	// pointer (starting from 0) would be maxPointerValue*sizeof(elementType) so
 	// divide by the element type to get the real maximum size.
-	maxSize := maxPointerValue / c.targetData.TypeAllocSize(elementType)
+	elementSize := c.targetData.TypeAllocSize(elementType)
+	if elementSize == 0 {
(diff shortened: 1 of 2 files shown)


aykevl (approved):

compiler/compiler.go
```diff
@@ -1554,7 +1554,11 @@ func (c *compilerContext) maxSliceSize(elementType llvm.Type) uint64 {
 	// Determine the maximum allowed size for a slice. The biggest possible
 	// pointer (starting from 0) would be maxPointerValue*sizeof(elementType) so
 	// divide by the element type to get the real maximum size.
-	maxSize := maxPointerValue / c.targetData.TypeAllocSize(elementType)
+	elementSize := c.targetData.TypeAllocSize(elementType)
+	if elementSize == 0 {
+		elementSize = 1
```
soypat, line 1559: Can't you set maxSize=0 for this case to avoid allocations altogether?

deadprogram: I think that is not the same thing. Empty struct still requires allocation, since not `nil`.

deadprogram:

compiler/compiler.go
aykevl: Correct, they require something in the data pointer so that `slice[:0:0]` remains non-nil. It doesn't have to be a true allocation however, it might as wel be a fixed pointer constant (like the one returned by `runtime.alloc(0)`).

PR #2751, opened 2022-04-06
Fall back to `go version` when GOROOT version detection fails

eliasnaur:
This change makes the Go version detection logic succeed on my NixOS 21.11.

### DIFF
--- a/goenv/version.go
+++ b/goenv/version.go
@@ -58,9 +58,9 @@ func GorootVersionString(goroot string) (string, error) {
 		return string(data), nil
 
 	} else if data, err := ioutil.ReadFile(filepath.Join(
-		goroot, "src", "runtime", "internal", "sys", "zversion.go")); err == nil {
+		goroot, "src", "internal", "buildcfg", "zbootstrap.go")); err == nil {
 
-		r := regexp.MustCompile("const TheVersion = `(.*)`")
+		r := regexp.MustCompile("const version = `(.*)`")
 		matches := r.FindSubmatch(data)
 		if len(matches) != 2 {
 			return "", errors.New("Invalid go version output:\n" + string(data))


aykevl:
Do you know why the existing code fails? It tries both `$GOROOT/VERSION` and `$GOROOT/src/runtime/internal/sys/zversion`. Are both of those files missing in Nix?

(I'd like to understand why it fails before accepting this solution, I suspect there might be something else going on).

eliasnaur:
`VERSION` is not present, and `zversion.go` no longer contains `TheVersion` in recent Go versions. Here's the content of `zversion.go` from a recent Go checkout, built the regular way with `make.bash`:

```
// Code generated by go tool dist; DO NOT EDIT.

package sys

const StackGuardMultiplierDefault = 1
```

builder/config.go
```diff
@@ -47,3 +45,23 @@ func NewConfig(options *compileopts.Options) (*compileopts.Config, error) {
...
+	goroot := goenv.Get("GOROOT")
+	if goroot == "" {
+		return 0, 0, errors.New("cannot locate $GOROOT, please set it manually")
+	}
+
+	return goenv.GetGorootVersion(goroot)
+}
+
+func getGoVersion() (major, minor int, err error) {
+	ver, err := exec.Command("go", "version").CombinedOutput()
```
aykevl: Please use the `go` binary from the detected GOROOT. Something like:

```suggestion
	ver, err := exec.Command(filepath.Join(goroot, "bin", "go"), "version").CombinedOutput()
```

This avoids issues if the GOROOT is set manually via an environment variable.

aykevl (commented):
> `VERSION` is not present,

Hmm, that's unfortunate. It's part of the official release tarballs, I wonder why Nix doesn't include the version file?

> and `zversion.go` no longer contains `TheVersion` in recent Go versions

Ah, yes. I think that's the reason why we added the `$GOROOT/VERSION` check.
We could perhaps replace that check by reading the version from `$GOROOT/src/internal/buildcfg/zbootstrap.go`? It also contains the version number (and is hopefully present on NixOS).

builder/config.go
eliasnaur: Good point, but no longer relevant.

eliasnaur:
> > `VERSION` is not present,
> 
> Hmm, that's unfortunate. It's part of the official release tarballs, I wonder why Nix doesn't include the version file?
> 

Not sure, I'm very new to NixOS.

> > and `zversion.go` no longer contains `TheVersion` in recent Go versions
> 
> Ah, yes. I think that's the reason why we added the `$GOROOT/VERSION` check. We could perhaps replace that check by reading the version from `$GOROOT/src/internal/buildcfg/zbootstrap.go`? It also contains the version number (and is hopefully present on NixOS).

`zbootstrap.go` exists and contains the version. Thanks, I've updated the PR.

deadprogram:
It would appear this is ready to merge. @aykevl is anything else missing or needed here?

deadprogram:
It would appear this is ready to merge. @aykevl is anything else missing or needed here?

deadprogram (approved):

deadprogram:
Merging, thanks @eliasnaur

PR #2753, opened 2022-04-06
compiler: fix difference in aliases in interface methods

aykevl:
There used to be a difference between `byte` and `uint8` in interface
methods. These are aliases, so they should be treated the same.
This patch introduces a custom serialization format for types,
circumventing the `Type.String()` method that is slightly wrong for our
purposes.

This also fixes an issue with the `any` keyword in Go 1.18, which
suffers from the same problem (but this time actually leads to a crash).


deadprogram (approved):

deadprogram:

PR #2755, opened 2022-04-07
src/runtime: second draft of map growth code

### LINKED ISSUE #1553: maps need growing logic when they fill up

dgryski:
Fixes #1553


dgryski:
~For some reason, tinygo/llvm didn't like the `n := *m` and `*m = n` lines from my original patch.  That points to a larger issue that needs to be fixed, but this solves the failing `hash/adler32` test.~

Edit: nope, this was totally incorrect.

dgryski:
Still have some `uintptr` potential overflow issues to solve.

deadprogram:
It is all passing and looks good to my little eye. @aykevl any feedback here?

src/runtime/hashmap.go
```diff
@@ -57,12 +68,70 @@ func hashmapMake(keySize, valueSize uint8, sizeHint uintptr) *hashmap {
...
+	}
+}
+
+func hashmapShouldGrow(m *hashmap) bool {
+
+	if heapStart == heapEnd {
+		// don't try to grow if we have no memory
+		// this also triggers during interp
+		return false
 	}
```
aykevl: I don't agree with this. I'm pretty sure this indicates a bug somewhere else in the code.
(I need to review the rest of the code still, it's probably fine)

dgryski: Yeah, this was the workaround I found. I don't like it either.

deadprogram: OK, so what do we do now? If this indicates a bug elsewhere, is there a way we can trigger it and find it?

If we are not going to do this right now, we could create an issue to track doing so, and then merge this PR (pending any other feedback) in the interest of unblocking this feature.

aykevl: I have a strong suspicion this is an interp bug. One possible solution (that's still a hack but a lot cleaner) is to add `runtime.hashmapGrow` to the functions that are excluded from interp:

https://github.com/tinygo-org/tinygo/blob/9eb4a6268aa8d61de9d6c03d6d02c0f0c2be157f/interp/interpreter.go#L204-L217

aykevl:
```
heapStart: 0x0009fff0
heapEnd: 0x000a0000
```
Actually, it might be a GC issue after all. These two addresses are awkwardly close together.

aykevl: I think I have a fix. I'll test it a bit more to confirm.

aykevl: Here is the fix: https://github.com/tinygo-org/tinygo/pull/2812

deadprogram: @dgryski #2812 is in `dev` now can you please rebase and make suitable adjustments to this PR?

deadprogram:
OK, seems like the main issue with this PR has now been addressed. Are we good to merge?

main_test.go
```diff
@@ -57,6 +57,7 @@ func TestBuild(t *testing.T) {
 		"json.go",
 		"map.go",
 		"math.go",
+		"mapgrowth.go",
```
aykevl: Please keep this list sorted.
Also, why not merge mapgrowth.go with map.go? Unfortunately, compilations are still relatively slow and adding new tests slows down everything a lot.

src/runtime/hashmap.go
```diff
@@ -57,14 +68,65 @@ func hashmapMake(keySize, valueSize uint8, sizeHint uintptr) *hashmap {
...
+func hashmapKeyEqualAlg(alg hashmapAlgorithm) func(x, y unsafe.Pointer, n uintptr) bool {
+	switch alg {
+	case hashmapAlgorithmBinary:
+		return memequal
+	case hashmapAlgorithmString:
+		return hashmapStringEqual
+	case hashmapAlgorithmInterface:
+		return hashmapInterfaceEqual
+	default:
+		panic("unknown hashmap equal algorithm")
```
aykevl: If we get here, that's a compiler bug. It should never ever be reached. I think returning nil is the better option to save binary size (or even something like `__builtin_unreachable()` in C, which isn't yet exposed in TinyGo).

src/runtime/hashmap.go
```diff
@@ -57,14 +68,65 @@ func hashmapMake(keySize, valueSize uint8, sizeHint uintptr) *hashmap {
...
+func hashmapKeyHashAlg(alg hashmapAlgorithm) func(key unsafe.Pointer, n uintptr) uint32 {
+	switch alg {
+	case hashmapAlgorithmBinary:
+		return hash32
+	case hashmapAlgorithmString:
+		return hashmapStringPtrHash
+	case hashmapAlgorithmInterface:
+		return hashmapInterfacePtrHash
+	default:
+		panic("unknown hashmap hash algorithm")
```
aykevl: Same here.

aykevl (commented):
This looks pretty reasonable. Just a few nits below.
There is an impact on code size, but in many cases it's not too big (especially considering this adds a new feature). And it may be avoidable by using the function pointers in a slightly different way.

main_test.go
dgryski: It's a different file because I just added the test program I was using for development and dropped it into the testdata directory.  I'll add it to the `map.go` file properly.

dgryski:
PTAL (while CI churns...)

src/runtime/hashmap.go
```diff
@@ -57,14 +68,67 @@ func hashmapMake(keySize, valueSize uint8, sizeHint uintptr) *hashmap {
...
+	switch alg {
+	case hashmapAlgorithmBinary:
+		return memequal
+	case hashmapAlgorithmString:
+		return hashmapStringEqual
+	case hashmapAlgorithmInterface:
+		return hashmapInterfaceEqual
+	default:
+		// compiler bug :(
+		return nil
```
aykevl, line 95: Technically correct. I think I would have called it "unreachable" but :shrug:

aykevl (approved):

src/runtime/hashmap.go
dgryski: I'll fix that in a follow up PR.

deadprogram:
At long last merging! Fantastic work @dgryski

PR #2761, opened 2022-04-10
ci: update supported go versions from 1.15-1.17 to 1.15-1.18

dkegel-fastly:
Currently shows one or two tinygo-test failures on windows with 1.18

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,15 +102,15 @@ commands:
       - run: make fmt-check
 
 jobs:
-  test-llvm11-go115:
+  test-llvm11-go116:
     docker:
-      - image: circleci/golang:1.15-buster
+      - image: circleci/golang:1.16-buster
     steps:
       - test-linux:
           llvm: "11"
-  test-llvm12-go116:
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -101,21 +101,21 @@ commands:
       - run: make fmt-check
 
 jobs:
-  test-llvm11-go115:
+  test-llvm11-go116:
```
aykevl: IMHO we should either drop support for Go 1.15 entirely or continue to support in including in CI. We can for example test Go 1.15 and 1.17: it's unlikely that with those tests, Go 1.16 will somehow break.
(I don't have a strong preference, but unless it holds back some new features I'd be in favor of keeping Go 1.15 support for now).

dkegel-fastly: I changed the 1.16 to 1.15 in circleci.

deadprogram (approved):

deadprogram:
Merging now so we can get closer to a release. Your patient work on this is essential @dkegel-fastly thank you!

PR #2774, opened 2022-04-13
cgo: refactor

aykevl:
This is a large refactor of the cgo package. It should fix a number of smaller problems and be a bit more strict (like upstream CGo): it for example requires every Go file in a package to include the header files it needs instead of piggybacking on imports in earlier files.

The main benefit is that it should be a bit more maintainable and easier to add new features in the future (like static functions).

This breaks the tinygo.org/x/bluetooth package (https://github.com/tinygo-org/bluetooth/pull/97), which should be updated before this change lands.

### DIFF
--- a/cgo/cgo.go
+++ b/cgo/cgo.go
@@ -18,7 +18,6 @@ import (
 	"go/scanner"
 	"go/token"
 	"path/filepath"
-	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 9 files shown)


aykevl:
This PR changes the internal structure of the cgo package:
- Instead of scanning the Go files for C references and then scanning the C files for declarations, the order is reversed: all C declarations are scanned and then the Go AST is scanned for `C.foo` symbols to replace. This should fix some issues with the const parser in a future change.
- CGo is processed independently per Go file and duplicate definitions are deduplicated as needed. This matches upstream CGo.

deadprogram:
> CGo is processed independently per Go file and duplicate definitions are deduplicated as needed. This matches upstream CGo.

:100:

deadprogram:
> This is a large refactor of the cgo package.

You do not exaggerate. I am still digesting it, but so far I like what I see.

deadprogram:
I suspect a change may be needed for https://github.com/tinygo-org/tinyfs

aykevl:
> I suspect a change may be needed for https://github.com/tinygo-org/tinyfs

Didn't test it, but yes there's a good chance that package needs an update.

deadprogram:
Seems like tinyfs is still working as expected even with this refactor. I think we should merge unless for some reason we want to hold until after next release.

deadprogram:
Let's merge this now. Thanks @aykevl for the good work!

PR #2776, opened 2022-04-13
cgo: implement support for static functions

aykevl:
This depends on #2774.

Static functions are used in tinygo.org/x/bluetooth but so far we've mostly just worked around them. This implements proper support for static functions.

### DIFF
--- a/cgo/cgo.go
+++ b/cgo/cgo.go
@@ -32,6 +32,7 @@ type cgoPackage struct {
 	errors          []error
 	currentDir      string // current working directory
 	packageDir      string // full path to the package to process
+	importPath      string
 	fset            *token.FileSet
 	tokenFiles      map[string]*token.File
 	definedGlobally map[string]ast.Node
@@ -39,12 +40,15 @@ type cgoPackage struct {
 	cflags          []string // CFlags from #cgo lines
 	ldflags         []string // LDFlags from #cgo lines
 	visitedFiles    map[string][]byte
+	cgoHeaders      []string
 }
 
 // cgoFile holds information only for a single Go file (with one or more
 // `import "C"` statements).
 type cgoFile struct {
 	*cgoPackage
+	file    *ast.File
+	index   int
 	defined map[string]ast.Node
 	names   map[string]clangCursor
 }
@@ -158,9 +162,10 @@ func GoBytes(ptr unsafe.Pointer, length C.int) []byte {
 // functions), the CFLAGS and LDFLAGS found in #cgo lines, and a map of file
 // hashes of the accessed C header files. If there is one or more error, it
 // returns these in the []error slice but still modifies the AST.
-func Process(files []*ast.File, dir string, fset *token.FileSet, cflags []string, clangHeaders string) (*ast.File, []string, []string, []string, map[string][]byte, []error) {
+func Process(files []*ast.File, dir, importPath string, fset *token.FileSet, cflags []string, clangHeaders string) (*ast.File, []string, []string, []string, map[string][]byte, []error) {
 	p := &cgoPackage{
 		currentDir:      dir,
+		importPath:      importPath,
 		fset:            fset,
 		tokenFiles:      map[string]*token.File{},
 		definedGlobally: map[string]ast.Node{},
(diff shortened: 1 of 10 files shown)


aykevl:
```
Error: /Users/runner/work/tinygo/tinygo/testdata/cgo/extra.go:9:78: error: aliases are not supported on darwin
void _Cgo_static_8fa7a76b0410251fc60e_headerfunc_static(void) __attribute__((alias("headerfunc_static")));
```

_grumble_

aykevl:
I expected this to be necessary for https://github.com/tinygo-org/drivers/pull/401 but apparently it's not.

deadprogram:
Now that #2774 has been merged, this PR should now be able to be moved forward.

aykevl:
Unfortunately, not. It doesn't work on MacOS yet. I need to work around that somehow.

ysoldak:
> Unfortunately, not. It doesn't work on MacOS yet. I need to work around that somehow.

I've found this workaround for MacOS, can we use it?
https://github.com/besser82/libxcrypt/blob/b8c6d0475f5e1a87a6dfbc9f62b7402b69052049/lib/crypt-port.h#L191-L198

```
/* Darwin (Mach-O) doesn't support alias attributes or symbol versioning.
   It does, however, support symbol aliasing at the object file level.  */
#ifdef __APPLE__

# define _strong_alias(name, aliasname)         \
  __asm__(".globl _" #aliasname);               \
  __asm__(".set _" #aliasname ", _" #name);     \
  extern __typeof(name) aliasname __THROW
```

aykevl:
@ysoldak thanks for this suggestion! I was aware of the linker feature but didn't know how to use it from C.

Unfortunately, the linker we use doesn't support aliases:

```
ld.lld-14: error: TODO: support symbols of type 10
ld.lld-14: error: TODO: support symbols of type 10
ld.lld-14: error: undefined symbol: __Cgo_static_425e352dd21e64a2b260_headerfunc_static
>>> referenced by /tmp/tinygo2273409226/main.o

ld.lld-14: error: undefined symbol: __Cgo_static_f2b9c44701219e7a3dfe_headerfunc_static
>>> referenced by /tmp/tinygo2273409226/main.o
error: failed to link /tmp/tinygo2273409226/main: exit status 1
```

Perhaps I can try to add support...

ysoldak:
"the linker we use doesn't support aliases", you mean this?
> ld.lld-14: error: TODO: support symbols of type 10

aykevl:
Yes. "symbol of type 10" refers to aliases.

aykevl:
I've added a workaround for MacOS, let's see whether it works.

ysoldak:
It works on my Mac at least. macOS 12.5.1
I've tried
```
GOOS=darwin GOARCH=amd64 tinygo  build  -size short -o test       ./testdata/cgo
  12734     112     145 |   12846     257
```
and
```
GOOS=darwin GOARCH=arm64 tinygo  build  -size short -o test       ./testdata/cgo
  12704     120     136 |   12824     256
```

aykevl:
Reminder: this PR is ready for review.

cgo/cgo.go
```diff
@@ -210,13 +215,13 @@ func Process(files []*ast.File, dir string, fset *token.FileSet, cflags []string
 		}
 	}
 	// Patch some types, for example *C.char in C.CString.
-	cf := p.newCGoFile()
+	cf := p.newCGoFile(nil, -1)
```
deadprogram: Is there some special significance to the `-1` and if so perhaps a named `const` instead?

cgo/cgo.go
```diff
@@ -1142,27 +1152,55 @@ func (f *cgoFile) getASTDeclNode(name string, found clangCursor, iscall bool) as
...
+			var params []string
+			for _, param := range node.Type.Params.List {
+				params = append(params, param.Names[0].Name)
+			}
+			callInst := fmt.Sprintf("%s(%s);", name, strings.Join(params, ", "))
+			if node.Type.Results != nil {
+				callInst = "return " + callInst
+			}
+			aliasDeclaration := fmt.Sprintf(`
+#ifdef __APPLE__
```
deadprogram, line 1183: :crying_cat_face: oh Apple

deadprogram (approved):
I am really looking forward to being able to compile some libs that have not been easily importable until now.

cgo/cgo.go
aykevl: No, not really. I'm just creating a dummy `*cgoFile` object. I have added a comment that should hopefully make this clearer.

cgo/cgo.go
aykevl: The worst part is that the linker actually _does_ support aliases, it's just the compiler that doesn't.
...and I don't feel like creating a Clang patch and pushing for its inclusion.

deadprogram:
Merging!!

Issue #2777, opened 2022-04-14
SIGSEGV in the compiler in transform.OptimizeAllocs call tree


ALTree:
```
$ tinygo version
tinygo version 0.22.0 linux/amd64 (using go version go1.17.8 and LLVM version 13.0.0)
```

The following program:

```
package main

var P *int

func f(p []int, m map[int]int) int {
	m[p[0]] = *P
	return 0
}

func main() {
	var m map[int]int
	f(make([]int, 1/f(nil, m)), m)
}
```

Occasionally triggers a panic in the compiler with stack trace:

```
$ tinygo build -o crash.o crash.go

fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x615 pc=0x4d3b3da]

runtime stack:
runtime.throw({0x50c8043, 0x7f6c4446a5c0})
	/opt/hostedtoolcache/go/1.17.6/x64/src/runtime/panic.go:1198 +0x71
runtime.sigpanic()
	/opt/hostedtoolcache/go/1.17.6/x64/src/runtime/signal_unix.go:719 +0x396

goroutine 53 [syscall]:
runtime.cgocall(0xe2ac70, 0xc0006e77f8)
	/opt/hostedtoolcache/go/1.17.6/x64/src/runtime/cgocall.go:156 +0x5c fp=0xc0006e77d0 sp=0xc0006e7798 pc=0xadc57c
tinygo.org/x/go-llvm._Cfunc_LLVMConstIntGetZExtValue(0x7f6c44011840)
	_cgo_gotypes.go:3365 +0x4d fp=0xc0006e77f8 sp=0xc0006e77d0 pc=0xccfccd
tinygo.org/x/go-llvm.Value.ZExtValue.func1({0x0})
	/home/runner/go/pkg/mod/tinygo.org/x/go-llvm@v0.0.0-20220121152956-4fa2ab2718f3/ir.go:854 +0x3f fp=0xc0006e7830 sp=0xc0006e77f8 pc=0xce761f
tinygo.org/x/go-llvm.Value.ZExtValue({0xc000464f50})
	/home/runner/go/pkg/mod/tinygo.org/x/go-llvm@v0.0.0-20220121152956-4fa2ab2718f3/ir.go:854 +0x19 fp=0xc0006e7848 sp=0xc0006e7830 pc=0xce75b9
[...]

PR #2778, opened 2022-04-14
transform: fix poison value in heap-to-stack transform

aykevl:
In https://github.com/tinygo-org/tinygo/issues/2777, a poison value
ended up in `runtime.alloc`. This shouldn't happen, especially not for
well written code. So I'm not sure why it happens. But here is a fix
anyway.


deadprogram:
Seems logical. Now merging, thanks @aykevl

PR #2780, opened 2022-04-15
Support command line parameters in `tinygo run`

aykevl:
This PR adds support for command line parameters to `tinygo run`. The biggest change is a refactor to make that possible.

Running normally:

```
$ tinygo run ./testdata/env.go foobar -asdf
[...]
arg: foobar
arg: -asdf
```

Running on an emulated Cortex-M chip:
```
$ tinygo run -target=cortex-m-qemu ./testdata/env.go foobar -asdf
[...]
arg: foobar
arg: -asdf
```

Running in wasmtime:

```
$ tinygo run -target=wasi ./testdata/env.go foobar -asdf
[...]
arg: foobar
arg: -asdf
```

(This is mostly just refactoring to make `tinygo run -target=esp32` easier to implement)

PR #2781, opened 2022-04-15
esp: support CGo

aykevl:
Without this patch, the include directory isn't found and picolibc.h (used indirectly by stdint.h for example) can't be found.

I would like to add tests for this but we currently don't run Xtensa tests. This should be possible however using https://github.com/espressif/qemu/wiki (see also: https://github.com/tinygo-org/tinygo/pull/2780).

Simple way to verify this patch: `tinygo build -o test.bin -target=esp32 ./testdata/cgo/`


deadprogram:
Awesome, thanks @aykevl

PR #2785, opened 2022-04-19
builder: free LLVM IR modules after use

aykevl:
This reduces the TinyGo memory consumption when running `make tinygo-test` from 5.8GB to 5GB on my laptop. This means there are still memory leaks, but some leaks have been fixed.


deadprogram:
In theory this is a good idea, but why are all the tests failing?

aykevl:
Second try. With much more things freed and hopefully also the bugs in the first try fixed.

deadprogram:
Looking good now @aykevl thank you!

PR #2786, opened 2022-04-19
main: use shared code for both run and test subcommands

aykevl:
This means that we don't need duplicate code to pass parameters to wasmtime and that the following actually produces verbose output (it didn't before this PR):

    tinygo test -v -target=cortex-m-qemu math


deadprogram (approved):
Makes sense, thanks @aykevl now merging.

PR #2790, opened 2022-04-20
compiler: remove support for memory references in AsmFull

aykevl:
Memory references (`*m` in LLVM IR inline assembly) need a pointer type starting in LLVM 14. This is a bit inconvenient and requires a new API in the go-llvm package.

Instead of doing that, I'd like to remove support for memory references from AsmFull (and possibly AsmFull entirely if possible: it's hard to use correctly).

This breaks tinygo.org/x/drivers/ws2812 for AVR, ARM, and RISC-V which need to be updated. Probably using CGo.

### DIFF
--- a/compiler/inlineasm.go
+++ b/compiler/inlineasm.go
@@ -98,7 +98,10 @@ func (b *builder) createInlineAsmFull(instr *ssa.CallCommon) (llvm.Value, error)
 			case llvm.IntegerTypeKind:
 				constraints = append(constraints, "r")
 			case llvm.PointerTypeKind:
-				constraints = append(constraints, "*m")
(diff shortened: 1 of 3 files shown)


deadprogram:
Tested with https://github.com/tinygo-org/drivers/pull/401 which was merged, and worked as expected. Is there something else I am missing here or can we merge now?

deadprogram (approved):

deadprogram:
Merging since it is working for me.

aykevl:
> Is there something else I am missing here or can we merge now?

I doubt inline assembly is used in many places. But the docs need to be updated, for which I've created a PR: https://github.com/tinygo-org/tinygo-site/pull/258

PR #2791, opened 2022-04-20
avr: simplify timer-based time

aykevl:
Simplify the interrupt-based timer code in a few ways:

  - Do not recalibrate the timer every 100ms. Instead, rely on the fact
    that the machine package will calbrate the timer if necessary if it
    makes changes to Timer0.
  - Do not configure Timer0 and then set nanosecondsInTick based on that
    value. Instead, use a fixed value.

These two changes together mean that in code that doesn't use PWM,
nanosecondsInTick will be constant which makes the TIMER0_OVF interrupt
handler a lot smaller.

Together this reduces the code size of AVR binaries by about 1200 bytes,
making it pretty close to the pre-timer code size (only about 250 bytes
larger).

It also somehow fixes a problem with
tinygo.org/x/drivers/examples/ws2812 on the Arduino Uno. I'm not quite
sure what was going wrong, but bisecting pointed towards the timer code
(https://github.com/tinygo-org/tinygo/pull/2428) and with this
simplification the bug appears to be gone.

### DIFF
--- a/src/runtime/runtime_avr.go
+++ b/src/runtime/runtime_avr.go
@@ -77,14 +77,6 @@ func nanosecondsToTicks(ns int64) timeUnit {
 // Sleep this number of ticks of nanoseconds.
 func sleepTicks(d timeUnit) {
 	waitTill := ticks() + d
-	// recalibrate if we have some time (>100ms) and it was a while when we did it last time.
-	if d > 100000 {
-		now := waitTill - d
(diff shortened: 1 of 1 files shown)


deadprogram:
Please see comment https://github.com/tinygo-org/drivers/pull/401#issuecomment-1104037443

deadprogram (approved):
This change seems good, I will enter separate issue for the Digispark itself.

PR #2792, opened 2022-04-20
LLVM 14 support

aykevl:
This PR adds support for LLVM 14 without dropping support for older versions. It switches the LLVM version for static builds, but doesn't switch the default for system LLVM builds yet because Homebrew doesn't have LLVM 14 yet.

This PR is similar to #2766 but is more complete. I ran a bunch of tests locally and it seems to work fine.

PR #2793, opened 2022-04-22
Llvm14 windows test

deadprogram:
Justa test.


aykevl:
If we merge #2792 and rebase this PR on top of it, those long CI times aren't necessary (because LLVM will have been cached already).

PR #2794, opened 2022-04-22
Run ESP32 inside QEMU

aykevl:
This should make it much easier to debug this chip:

Screenshot_20220422_191206

The text "hello world!" is printed by the emulated firmware, and the startup messages ("ets Jul 29 2019 12:21:46" etc) are part of the built-in ROM, which is also emulated.

For this to work, you need:

 - `qemu-system-xtensa` from https://github.com/espressif/qemu/wiki (I built it from source)
 - `xtensa-esp32-elf-gdb`, optional (needed only if you want to debug). At least version esp-2021r2, otherwise you get [this error](https://stackoverflow.com/questions/70644426/build-gdb-for-esp32-and-qemu).

### DIFF
--- a/targets/arduino-nano.json
+++ b/targets/arduino-nano.json
@@ -6,5 +6,5 @@
 		"-Wl,--defsym=_stack_size=512"
 	],
 	"flash-command": "avrdude -c arduino -p atmega328p -b 57600 -P {port} -U flash:w:{hex}:i",
-	"emulator": ["simavr", "-m", "atmega328p", "-f", "16000000"]
+	"emulator": "simavr -m atmega328p -f 16000000 {}"
 }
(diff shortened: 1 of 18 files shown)


targets/arduino-nano.json
```diff
@@ -6,5 +6,5 @@
 		"-Wl,--defsym=_stack_size=512"
 	],
 	"flash-command": "avrdude -c arduino -p atmega328p -b 57600 -P {port} -U flash:w:{hex}:i",
-	"emulator": ["simavr", "-m", "atmega328p", "-f", "16000000"]
+	"emulator": "simavr -m atmega328p -f 16000000 {}"
```
deadprogram, line 9: Too bad about the breaking change, but I prefer the new format.

deadprogram:
Merging so we can keep making progress on ESP32. Thanks @aykevl

PR #2796, opened 2022-04-23
docker: update Dockerfile to use Go 1.18

deadprogram:
This PR updates the Dockerfile used for the `dev` build to use Go 1.18.1

It also updates the linker for Xtensa to the LLVM 14 version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # tinygo-llvm stage obtains the llvm source for TinyGo
-FROM golang:1.17 AS tinygo-llvm
+FROM golang:1.18 AS tinygo-llvm
 
 RUN apt-get update && \
     apt-get install -y apt-utils make cmake clang-11 binutils-avr gcc-avr avr-libc ninja-build


Dockerfile
```diff
@@ -18,11 +18,11 @@ RUN cd /tinygo/ && \
...
 
-ARG xtensa_version="1.22.0-80-g6c4433a-5.2.0"
+ARG xtensa_version="esp-14.0.0-20220415"
 RUN cd /tmp/ && \
-    wget -q https://dl.espressif.com/dl/xtensa-esp32-elf-linux64-${xtensa_version}.tar.gz && \
-    tar xzf xtensa-esp32-elf-linux64-${xtensa_version}.tar.gz && \
-    cp ./xtensa-esp32-elf/bin/xtensa-esp32-elf-ld /usr/local/bin/ && \
+    wget -q https://github.com/espressif/llvm-project/releases/download/${xtensa_version}/xtensa-esp32-elf-llvm14_0_0-${xtensa_version}-linux-amd64.tar.xz && \
+    tar xvf xtensa-esp32-elf-llvm14_0_0-${xtensa_version}-linux-amd64.tar.xz && \
+    cp ./xtensa-esp32-elf-clang/bin/xtensa-esp32-elf-ld /usr/local/bin/ && \
```
aykevl: Why did you change to the Clang version? As far as I know, it's still experimental (and the linker - which is the only thing we need - is the same as in the GCC version anyway).

deadprogram: I assumed it was the same version, since that  github repo was linked to via the espressif site.

Also, I could not find an updated version for LLVM 14 on the dl.espressif.com pages.

aykevl: I don't think this needs an update. We're only using the linker here, which has remained the same for years.

> Also, I could not find an updated version for LLVM 14 on the dl.espressif.com pages.

That's not relevant, as this is not about LLVM. We're only copying the GNU linker here.

deadprogram: OK got it. Changed back to the original linker.

aykevl (approved):

PR #2797, opened 2022-04-23
thingsplus_rp2040: support sparkfun thingsplus rp2040 board

oflebbe:
Support for this board https://www.sparkfun.com/products/17745

### DIFF
--- a/targets/thingsplus-rp2040.json
+++ b/targets/thingsplus-rp2040.json
@@ -1,0 +1,11 @@
+{
+    "inherits": [
+        "rp2040"
+    ],
+    "serial": "uart",
+    "build-tags": ["thingsplus_rp2040"],
+    "linkerscript": "targets/thingsplus-rp2040.ld",
+    "extra-files": [
+        "targets/pico-boot-stage2.S"
+    ]
+}
--- a/targets/thingsplus-rp2040.ld
+++ b/targets/thingsplus-rp2040.ld
@@ -1,0 +1,10 @@
+
+MEMORY
+{
+    /* Reserve exactly 256 bytes at start of flash for second stage bootloader */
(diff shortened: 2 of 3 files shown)


targets/thingsplus-rp2040.json
```diff
@@ -0,0 +1,11 @@
+{
+    "inherits": [
+        "rp2040"
+    ],
+    "serial": "uart",
+    "build-tags": ["thingsplus_rp2040"],
+    "linkerscript": "targets/thingsplus-rp2040.ld",
```
deadprogram, line 7: I think you do not need a separate linker file `targets/thingsplus-rp2040.ld` since it is the same as the pico board you can  specify `targets/pico.ld` here and not add the separate file in this PR.

targets/thingsplus-rp2040.ld
```diff
@@ -0,0 +1,10 @@
+
```
deadprogram, line 1: See comment above about not needing this file.

targets/thingsplus-rp2040.json
oflebbe: the ld file is different. The thingsplus has 16mb of flash, while pico has 2mb. but that‘s the only diff, i have to admit

targets/thingsplus-rp2040.ld
oflebbe: see above : flash section is much larger

targets/thingsplus-rp2040.json
deadprogram: Oops I missed that difference. In that case, never mind!

deadprogram:
Thanks @oflebbe now merging.

Issue #2800, opened 2022-04-23
"tinygo test -target wasi compress/lzw" fails on windows?


dankegel:
Using today's dev branch plus a fix for #2799, on a (relatively) freshly set up windows 11 system, compress/lzw seems to fail on wasi:
```
$ ~/src/tinygo/build/tinygo test -target wasi -v compress/lzw
FAIL    compress/lzw    16.797s
```

deadprogram:
Is this still an issue?

ghost:
By me:

```shell
C:\>tinygo test -target wasi -v compress/lzw
error: wasm-opt failed: exit status 0xc0000409
FAIL

C:\>tinygo version
tinygo version 0.24.0 windows/amd64 (using go version go1.18.4 and LLVM version 14.0.0)
```

dgryski:
@dulanov That seems like a different error than the one @dankegel reported.

dgryski:
`compress/lzw` isn't playing nicely with wasi at all.

https://github.com/tinygo-org/tinygo/issues/2525

Closing as a dupe for now.

PR #2801, opened 2022-04-23
github: Run tinygo-test-wasi-fast on Windows

### LINKED ISSUE #2799: On windows 11, "tinygo test -target wasi  os" fails because wasi does not support cd .. yet
With a fresh windows 11 machine and today's dev (including clang 14):
```
$ ~/src/tinygo/build/tinygo test -target wasi os
Returning nil error open C:\Users\dank\AppData\Local\Temp\tinygotmp1446174477/TestFd.txt147587007: file not found
Caused by:
           1: 0x63ad - runtime.nilPanic
                           at C:\Users\dank\src\tinygo\src\runtime\panic.go:32:14
           2: 0x332f5 - (*os.File).Name
           3: 0x755c5 - os_test.TestFd
                           at C:\Users\dank\src\tinygo\src\os\file_anyos_test.go:98:21
```

Running wasmtime os.test under procmon, I noticed something funny: temporary files being opened were at the wrong location, as if wasmtime was confused by drive letters.  Building just the test by itself, then creating a writable directory C:\TMP,
then running in a CMD window, I saw that the command
```
wasmtime --dir=. --dir=.. --dir=..\.. --dir=/tmp --env=TMPDIR=/tmp os.test
```
succeeds, but the commands
```
wasmtime --dir=. --dir=.. --dir=..\.. --dir=c:/tmp --env=TMPDIR=c:/tmp os.test
wasmtime --dir=. --dir=.. --dir=..\.. --dir=\tmp --env=TMPDIR=\tmp os.test
wasmtime --dir=. --dir=.. --dir=..\.. --dir=c:\tmp --env=TMPDIR=c:\tmp os.test
```
all fail.  It seems that TMPDIR must be a unix style directory... which makes sense, since wasi is unix-ish.
And since wasmtime probably does an exact compare on the dirs to grant access, it needs the unix style directory, too.

The workaround is easy as long as TMPDIR is on the default drive...

This seems like a regression; wasn't this working before?  Maybe not; we haven't been running tinygo-test-wasm in ci on windows...

dkegel-fastly:
Also:
- Makefile: add list of passing stdlib tests on windows
- Makefile: demote flate/lzw a bit as it fails on windows wasi (#2800)
- main.go: on windows, wasmtime wants tmpdir expressed as a unix path; fixes #2799

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -96,3 +96,5 @@ jobs:
         run: make smoketest TINYGO=$(PWD)/build/tinygo AVR=0 XTENSA=0
       - name: Test stdlib packages
         run: make tinygo-test
+      - name: Test stdlib packages on wasi
+        run: make tinygo-test-wasi-fast
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -71,6 +71,9 @@ jobs:
       - name: Build wasi-libc
         if: steps.cache-wasi-libc.outputs.cache-hit != 'true'
         run: make wasi-libc
+      - name: Install wasmtime
+        run: |
+          scoop install wasmtime
       - name: Test TinyGo
         shell: bash
         run: make test GOTESTFLAGS="-v -short"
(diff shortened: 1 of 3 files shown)


dkegel-fastly:
FWIW, looks like _getdrive() just looks at the first chars of GetCurrentDirectory(), so the TODO might be easier to fix than I feared... (it was command.com that kept separate current directories per drive, not the kernel; I keep forgetting that.)

deadprogram:
Seems like drive mapping solution not quite working yet, @dkegel-fastly 

https://github.com/tinygo-org/tinygo/runs/6143551542?check_suite_focus=true#step:20:10

dkegel-fastly:
Indeed.  I'm regrouping.

Issue #2623, opened 2022-02-07
Seeed Studio XIAO BLE (and BLE Sense)


therebelrobot:
Hi there! I'm wanting to add support for the newest XIAO microcontroller from Seeed Studio, the [XIAO BLE](https://wiki.seeedstudio.com/XIAO-BLE-Sense-Getting-Started/). I've worked with the SAMD XIAO prior to this, but just got my hands on a BLE and a BLE Sense, both based on the Nordic nRF52840 MCU, which is covered by TinyGo in other boards. The XIAO BLE Sense is build on the same tech and even has the same documentation in Seeed's website, so I plan on building that support for both.

I'm trying to follow the [contributing doc](https://tinygo.org/docs/guides/contributing/) as closely as I can, including making this issue, and will for sure follow [the wiki on adding a new board](https://github.com/tinygo-org/tinygo/wiki/Adding-a-new-board). 

Is there documentation handy that shows how to link the local github build to use for development testing? (If it's in the contributing/wiki ignore me, I'm still making my way through them both)

deadprogram:
Hello @therebelrobot 

Please see https://tinygo.org/docs/guides/build/

therebelrobot:
Hi @deadprogram! I was able to get building working locally, thank you for that link.

I had an issue with tinygo not being able to find the msd volume when defined in the new device configuration, do you know of any way of scanning a device for available msd volumes? I'm at a loss for figuring out the correct volume name, and [the device's forum suggested one](https://forum.seeedstudio.com/t/xiao-ble-flash-method/262869/3) that didn't work for finding it.

therebelrobot:
For context, here's the info.txt file that shows up when the device is put into bootloader mode, maybe that'll help?

```
UF2 Bootloader 0.6.2-12-g459adc9-dirty lib/nrfx (v2.0.0) lib/tinyusb (0.10.1-293-gaf8e5a90) lib/uf2 (remotes/origin/configupdate-9-gadbb8c7)
Model: Seeed XIAO nRF52840
Board-ID: Seeed_XIAO_nRF52840_Sense
Date: Nov 30 2021
SoftDevice: S140 7.3.0
```

and the device configuration as i currently have it:

```json
{
  "inherits": ["nrf52840"],
  "build-tags": ["xiaoble"],
  "serial": "usb",
  "serial-port": ["acm:2886:0045", "acm:2886:8045", "acm:2886:0145"],
  "flash-1200-bps-reset": "true",
[...]

PR #2802, opened 2022-04-24
Seeed XIAO BLE nRF52840 (Sense) board support

ysoldak:
This PR adds support for a new nrf52840 based boards from Seeed Studio.

Seeed XIAO BLE is an ultra-small size, ultra-low power Bluetooth development board based on the Nordic nRF52840.
It features an onboard Bluetooth antenna, onboard battery charging chip, and 21*17.5mm thumb size, which makes it ideal for IoT projects. "Sense" variant is designed for IoT and AI applications, it adds 6 Dof IMU and a microphone.

This nice board has NRF SoftDevice (s140v7) pre-flashed together with UF2 bootloader. This makes it an ideal first board to experiment with TinyGo and Bluetooth (no debug probe or soldering required).
Currently, I'm using it instead of Arduino Nano 33 BLE in my hobby TinyGo-based [Head Tracker](https://github.com/ysoldak/HeadTracker) project.

Wiring of on-board LSM6DS3TR IMU sensor is a bit odd. LSM_PWR pin must be set to an unusual mode (not just "Output" and is 32bit, instead of 8bit that's supported by TinyGo now). At first, I've added this code here in "init()" method, but moved it to "drivers" repository instead, in "xiao_ble" specific code for lsm6ds3 driver (PR soon).

Here is the pin mode code, that makes IMU usable on this board. Just for history and reference:
```
	// Following lines are XIAO BLE Sense specific, they have nothing to do with sensor per se
	// Implementation adapted from https://github.com/Seeed-Studio/Seeed_Arduino_LSM6DS3/blob/master/LSM6DS3.cpp#L68-L77

	// Special mode for IMU power pin on this board.
	// Can not use pin.Configure() directly due to special mode and 32 bit size
	pinConfig := uint32(nrf.GPIO_PIN_CNF_DIR_Output<<nrf.GPIO_PIN_CNF_DIR_Pos) |
		uint32(nrf.GPIO_PIN_CNF_INPUT_Disconnect<<nrf.GPIO_PIN_CNF_INPUT_Pos) |
		uint32(nrf.GPIO_PIN_CNF_PULL_Disabled<<nrf.GPIO_PIN_CNF_PULL_Pos) |
		uint32(nrf.GPIO_PIN_CNF_DRIVE_H0H1<<nrf.GPIO_PIN_CNF_DRIVE_Pos) |
		uint32(nrf.GPIO_PIN_CNF_SENSE_Disabled<<nrf.GPIO_PIN_CNF_SENSE_Pos)
	nrf.P1.PIN_CNF[8].Set(pinConfig) // LSM_PWR == P1_08

	// Enable IMU
	machine.LSM_PWR.High()
```

### DIFF
--- a/src/machine/board_xiao-ble.go
+++ b/src/machine/board_xiao-ble.go
@@ -1,0 +1,113 @@
+//go:build xiao_ble
+// +build xiao_ble
+
+// This file contains the pin mappings for the Seeed XIAO BLE nRF52840 [Sense] boards.
+//
+// Seeed XIAO BLE is an ultra-small size, ultra-low power Bluetooth development board based on the Nordic nRF52840.
+// It features an onboard Bluetooth antenna, onboard battery charging chip, and 21*17.5mm thumb size, which makes it ideal for IoT projects.
+//
+// Seeed XIAO BLE nRF52840 Sense is a tiny Bluetooth LE development board designed for IoT and AI applications.
+// It features an onboard antenna, 6 Dof IMU, microphone, all of which make it an ideal board to run AI using TinyML and TensorFlow Lite.
+//
+// SoftDevice (s140v7) is pre-flashed on this board already.
+// See https://github.com/tinygo-org/bluetooth
+//
+// - https://www.seeedstudio.com/Seeed-XIAO-BLE-nRF52840-p-5201.html
+// - https://www.seeedstudio.com/Seeed-XIAO-BLE-Sense-nRF52840-p-5253.html
+//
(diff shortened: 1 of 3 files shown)


deadprogram:
Thanks for another great contribution @ysoldak now merging.

sago35:
#2623

ysoldak:
This is PR for on-board IMU support: https://github.com/tinygo-org/drivers/pull/410

sago35:
@ysoldak 
I am very interested.

PR #2803, opened 2022-04-24
build: use Go 1.18.1 for all builds

deadprogram:
This PR is to ensure that we use Go 1.18.1 for all builds on GHA.

It also updates Windows build to use the `windows-2022` runner. 

See https://github.blog/changelog/2021-11-16-github-actions-windows-server-2022-with-visual-studio-2022-is-now-generally-available-on-github-hosted-runners/


dkegel-fastly (approved):
Eventually we might want to build on one version of windows and smoke test on the oldest one we support.

PR #2804, opened 2022-04-24
machine/circuitplay-bluefruit: move pin mappings so board can be compiled for WASM

deadprogram:
This PR moves the pin mappings used by the `nrf52840` so that the Circuit Playground Bluefruit can have programs compiled into WASM for use in Playground.

### DIFF
--- a/src/machine/board_nrf52840.go
+++ b/src/machine/board_nrf52840.go
@@ -1,0 +1,56 @@
+//go:build nrf52840 || circuitplay_bluefruit || reelboard || clue || itsybitsy_nrf52840
+// +build nrf52840 circuitplay_bluefruit reelboard clue itsybitsy_nrf52840
+
+package machine
+
+// Hardware pins
+const (
+	P0_00 Pin = 0
+	P0_01 Pin = 1
+	P0_02 Pin = 2
+	P0_03 Pin = 3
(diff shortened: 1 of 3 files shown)


src/machine/board_nrf52840.go
```diff
@@ -0,0 +1,56 @@
+//go:build nrf52840 || circuitplay_bluefruit
+// +build nrf52840 circuitplay_bluefruit
```
aykevl: You could perhaps add a few other boards as well that are based on the nrf52840 (reelboard, itsybitsy-nrf52840, clue, to name a few). That way they can more easily be added to the playground later.

aykevl (commented):
I think we need to add a smoke test for simulation of at least one nrf52840 based board to the Makefile. Something like `$(TINYGO) build -size short -o test.wasm -tags=circuitplay_bluefruit             examples/blinky1`.

Other than that, LGTM.

src/machine/board_nrf52840.go
deadprogram: Done.

deadprogram:
Smoke test added.

deadprogram:
Merging, thanks for the feedback @aykevl

Issue #2652, opened 2022-02-24
unknown import: `env::sync/atomic.StoreUint32` has not been defined


kyleconroy:
Using the latest dev, I'm unable to run code generated by protoc when it's compiled to WASM / WASI.

```
Error: failed to run main module `example.wasm`

Caused by:
    0: failed to instantiate "example.wasm"
    1: unknown import: `env::sync/atomic.StoreUint32` has not been defined
make: *** [run] Error 1
```

A full reproduction can be found here: https://github.com/kyleconroy/tinygo-protoc. You'll need to have `wasmtime` installed to run the `example.wasm` file.

dgryski:
For me this also fails on native.

dgryski:
Minimal reproducer:

From https://github.com/protocolbuffers/protobuf-go/blob/master/internal/impl/extension.go#L131

```
import (
	"sync/atomic"
)

var global uint32

func foo() {
	defer atomic.StoreUint32(&global, 1)

	atomic.StoreUint32(&global, 2)
}

func main() {
	foo()
	println(atomic.LoadUint32(&global))
}
```

dgryski:
Potential solution: include a real copy of this function (instead of a compiler intrinsic) but marked as "always-inline"

kyleconroy:
@dgryski I'd love to implement this, but I understood about one-fifth of your comment. If you have any pointers on where to get started, let me know.

skabbes:
@kyleconroy I'm certainly not an expert - just feeling my way in the dark - but what I believe he means is that `atomic.StoreUint32` is already implemented - you can see it here:
https://github.com/tinygo-org/tinygo/blob/3883550c446fc8b14115366284316290f3d3c4b1/compiler/atomic.go#L88

and I believe it calls this at runtime:

https://github.com/tinygo-org/tinygo/blob/3883550c446fc8b14115366284316290f3d3c4b1/src/runtime/atomics_critical.go#L123

What I _think_ is happening is that the call to atomic.StoreX is being directly rewritten to a llmv atomic store.
https://github.com/tinygo-org/tinygo/blob/3883550c446fc8b14115366284316290f3d3c4b1/compiler/compiler.go#L1446

And the actual function for `atomic.StoreUint32` is not being generated in the target runtime.

Maybe here is a good place to look:

https://github.com/tinygo-org/tinygo/blob/3883550c446fc8b14115366284316290f3d3c4b1/compiler/defer.go#L348

[...]

PR #2805, opened 2022-04-24
fix: Define and inline atomic operations

kyleconroy:
Implement the atomic package in Go, but take advantage of the existing atomic methods defined using LLVM intrinsics. I picked the `llvm_` prefix quickly, so happy to change that to something else.

Fixes https://github.com/tinygo-org/tinygo/issues/2652

```
package main

import (
	"sync/atomic"
)

var global uint32

func foo() {
	defer atomic.StoreUint32(&global, 1)
	atomic.StoreUint32(&global, 2)
}

func main() {
	foo()
	println(atomic.LoadUint32(&global))
}
```

### DIFF
--- a/compiler/atomic.go
+++ b/compiler/atomic.go
@@ -14,7 +14,7 @@ import (
 func (b *builder) createAtomicOp(call *ssa.CallCommon) (llvm.Value, bool) {
 	name := call.Value.(*ssa.Function).Name()
 	switch name {
-	case "AddInt32", "AddInt64", "AddUint32", "AddUint64", "AddUintptr":
+	case "llvm_AddInt32", "llvm_AddInt64", "llvm_AddUint32", "llvm_AddUint64", "llvm_AddUintptr":
 		ptr := b.getValue(call.Args[0])
 		val := b.getValue(call.Args[1])
 		if strings.HasPrefix(b.Triple, "avr") {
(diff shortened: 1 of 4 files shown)


compiler/atomic.go
```diff
@@ -14,7 +14,7 @@ import (
 func (b *builder) createAtomicOp(call *ssa.CallCommon) (llvm.Value, bool) {
 	name := call.Value.(*ssa.Function).Name()
 	switch name {
-	case "AddInt32", "AddInt64", "AddUint32", "AddUint64", "AddUintptr":
```
skabbes, line 17: @kyleconroy I think you want to keep the old ones here too, not remove them.  If you do that, I don't _think_ you'll need to pull in `atomic/value.go` as well.

kyleconroy: I think you have to copy it all over because of the changes to `loader/goroot.go`

skabbes: Ah, when I was reading the code I thought it was maybe file by file (not package by package).

skabbes: But I think the maintainers are going to want to keep the original (non llvm_) AST rewrites anyway (at least for now) to 100% preserve legacy behavior, but let's see what they say.

kyleconroy:
@dgryski Are you the best person to review this change?

dgryski:
@aykevl or @niaow have more experience with the llvm and atomics bits.

aykevl (commented):
This approach probably works fine, but it has one downside: it copies the sync/atomic package into the TinyGo tree (which I'd rather avoid).

There is another way to solve this: by defining the functions in LLVM IR instead of transforming calls to it. So for example instead of recognizing calls to `sync/atomic.LoadInt32` and directly creating an atomic load inline, it is also possible to define `sync/atomic.LoadInt32` with a custom body that does the given operation. This should preserve the ability to take pointers to these functions.

kyleconroy:
> it is also possible to define sync/atomic.LoadInt32 with a custom body that does the given operation. This should preserve the ability to take pointers to these functions.

Is this done anywhere else inside the TinyGo codebase?

skabbes:
@aykevl took a stab here:
https://github.com/tinygo-org/tinygo/pull/2821

aykevl:
> > it is also possible to define sync/atomic.LoadInt32 with a custom body that does the given operation. This should preserve the ability to take pointers to these functions.
> 
> Is this done anywhere else inside the TinyGo codebase?

Sorry for not replying!
It is not, but I've been working on a feature that would do this:
https://github.com/tinygo-org/tinygo/commit/4e9f222d07eef9993592f84666c18eae8e0743c6#diff-d3cde2dec4b7c2263747b17d43f704418e6657b49583254628bcad1d1052471bR773

dgryski:
Closing as this was implemented in https://github.com/tinygo-org/tinygo/pull/2920

PR #2807, opened 2022-04-25
ci: run tinygo-test-wasi-fast on windows

### LINKED ISSUE #2799: On windows 11, "tinygo test -target wasi  os" fails because wasi does not support cd .. yet

dkegel-fastly:
Also:
- Makefile: tinygo-test*: disable compress-lzw on wasi, enable on windows; #2800
- TestChdir: don't use Chdir(".."), as wasi does not implement that yet; fixes #2799

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -71,6 +71,9 @@ jobs:
       - name: Build wasi-libc
         if: steps.cache-wasi-libc.outputs.cache-hit != 'true'
         run: make wasi-libc
+      - name: Install wasmtime
+        run: |
+          scoop install wasmtime
       - name: Test TinyGo
         shell: bash
         run: make test GOTESTFLAGS="-v -short"
@@ -96,3 +99,5 @@ jobs:
         run: make smoketest TINYGO=$(PWD)/build/tinygo AVR=0 XTENSA=0
       - name: Test stdlib packages
         run: make tinygo-test
+      - name: Test stdlib packages on wasi
+        run: make tinygo-test-wasi-fast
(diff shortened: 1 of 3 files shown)


deadprogram (approved):

deadprogram:
Now merging, thanks @dkegel-fastly

PR #2811, opened 2022-04-26
machine/thingplus: rename to the singular thing as that is correct

deadprogram:
This PR renames the "thingsplus" to the singular "thingplus" as that is the correct name.

https://www.sparkfun.com/products/17745


deadprogram:
@oflebbe

oflebbe:
OMG. Yes you are right.

oflebbe (approved):
+1

PR #2812, opened 2022-04-26
runtime: be able to deal with a very small heap

aykevl:
See the comment in the source for details.

Also see the discussion in https://github.com/tinygo-org/tinygo/pull/2755, which originally triggered this bug.

Somewhat surprising, this results in a slight code size decrease for ARM targets of a few bytes.


deadprogram:
Looking good! Thanks @aykevl now merging.

PR #2820, opened 2022-05-01
os: PathError now alias for fs.PathError, fixes #2817

### LINKED ISSUE #2817: "tinygo test io/ioutil" fails because os.PathError has not moved to io/fs as in upstream
Currently, "tinygo test io/ioutil" fails like this post go 1.15:
```
--- FAIL: TestTempDir_BadDir (0.00s)
    TempDir error = &T{Op:"stat", Path:"/tmp/TestTempDir_BadDir216851801/not-exist", Err:(T)(0x247638)}; want PathError for path "/tmp/TestTempDir_BadDir216851801/not-exist" satisifying os.IsNotExist
```
while the nearly identical test in os,  TestMkdirTempBadDir, passes.

See this comment in os/file.go:
```
// TODO: PathError moved to io/fs in go 1.16 and left an alias in os/errors.go.
// Do the same once we drop support for go 1.15.
type PathError struct {
```

Patch coming.

dkegel-fastly:
This lets us add io/ioutil to "make test-tinygo" on linux and mac.

This requires dropping go 1.15 support.

Fixes https://github.com/tinygo-org/tinygo/issues/2817

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,9 +102,9 @@ commands:
       - run: make fmt-check
 
 jobs:
-  test-llvm11-go115:
+  test-llvm11-go116:
     docker:
-      - image: circleci/golang:1.15-buster
+      - image: circleci/golang:1.16-buster
     steps:
       - test-linux:
           llvm: "11"
@@ -118,5 +118,5 @@ jobs:
 workflows:
   test-all:
     jobs:
-      - test-llvm11-go115
+      - test-llvm11-go116
       - test-llvm12-go117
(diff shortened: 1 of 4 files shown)


aykevl:
Dropping Go 1.15 seems good to me. However, please also do it completely (also removing any cruft left over for Go 1.15). While grepping for `1\.15` I found some files that still contain code for Go 1.15:

- BUILDING.md
- go.mod
- src/os/file.go
- src/os/os_anyos_test.go
- src/os/tempfile_test.go

Maybe it would be better to do this in a separate PR? Or at least in one commit that comes before the "os: PathError now alias for fs.PathError" commit.

dkegel-fastly:
Closing in favor of https://github.com/tinygo-org/tinygo/pull/2825

Issue #2667, opened 2022-03-03
Serialize and deserialize protocol buffers


kyleconroy:
As outlined in https://github.com/tinygo-org/tinygo/issues/447, TinyGo should support common serialization formats. While most people may think of JSON, I personally want TinyGo to support [protocol buffers](https://developers.google.com/protocol-buffers). This issue is meant to track my progress on adding Protocol Buffer support to TinyGo.

## Design

The default Protocol Buffer package (https://pkg.go.dev/google.golang.org/protobuf) makes heavy use of reflection. TinyGo only supports a small amount of the `reflect` package. Heavy use of reflection leads to slow code, so the folks at PlanetScale wrote [vtprotobuf](https://github.com/planetscale/vtprotobuf), which generates explicit serialization / deserialization code. This code works today with TinyGo. However, it requires the existing, slow code to compile.

So the plan is simple: Support compiling the code generated by protoc-gen-go and running and of the code contained in `init()` functions. Use the code generated by vtprotobuf to serialize and deserialize the data.

## Roadmap
- [x] Compile code generated by https://github.com/planetscale/vtprotobuf
- [x] Compile code generated by [protoc-gen-go](https://developers.google.com/protocol-buffers/docs/reference/go-generated)
   - [x] https://github.com/tinygo-org/tinygo/pull/2624
   - [x] https://github.com/tinygo-org/tinygo/issues/2806
- [ ] Import code generated by protoc-gen-go
  - [x] https://github.com/tinygo-org/tinygo/issues/2652
  - [x] https://github.com/tinygo-org/tinygo/issues/2668

paralin:
@kyleconroy This is a great idea, I'm using vtprotobuf a lot currently and it works great. Having some kind of auto-transpiler that automatically converts the reflection-based protobuf-go to use vtprotobuf would be great.

Quick question: as of March 2023, it seems like most of the reflection features are supported - is there anything missing for the existing reflection version to work currently?

dgryski:
And even more reflect is working, including the last bug on you roadmap.

gitslav:
@kyleconroy @dgryski looking at the roadmap is it time for the confetti emoji?

rockwotj:
So it seems that the default protobuf compiler is not supported still even with the improvements to reflection. I have yet to test `vtprotobuf`

```
% ./wasmtime-v14.0.4-x86_64-linux/wasmtime run protobuf-example.wasm
[...]

PR #2821, opened 2022-05-01
compiler: implement deferred calls to sync/atomic.*

skabbes:
Fixes: https://github.com/tinygo-org/tinygo/issues/2652 , https://github.com/tinygo-org/tinygo/pull/2805
And is part of a larger effort to get Protobufs working with tinygo. https://github.com/tinygo-org/tinygo/issues/2667

Previously discussed [here](https://github.com/tinygo-org/tinygo/issues/2652), and [here](https://github.com/tinygo-org/tinygo/pull/2805).

**Problem**
Deferrred calls to the sync/atomic package would fail with a linker error, since those _functions_ don't really exist in the LLVM representation.

```go
package main

import (
	"sync/atomic"
)

var global uint32

func foo() {
	defer atomic.StoreUint32(&global, 2) // the offending line
	atomic.StoreUint32(&global, 1)
	println(atomic.LoadUint32(&global))
}

func main() {
	foo()
	println(atomic.LoadUint32(&global))
}

```

image

**Solution**
After some discussion in the referenced issues, the solution I pursued is to share the previous rewrite code with `createRunDefers`.  (In effect the deferred calls to atomic.* are inlined in exactly the same way).

image

**Next Steps**
The amount of code changed is small, and easy to understand, and given this has never worked before - I am fairly confident it won't break anything. The tests 'prove' it works, but given the "run defers" code seems particularly hairy (I would love a second set of eyes on it).

@aykevl this is a slight modification of your original idea [here](https://github.com/tinygo-org/tinygo/pull/2805#pullrequestreview-957580759), do you mind taking a look?

_Please feel free to hack away at my code if you'd find that faster than coordinating the work with me over comments._

### DIFF
--- a/compiler/atomic.go
+++ b/compiler/atomic.go
@@ -11,12 +11,12 @@ import (
 // createAtomicOp lowers an atomic library call by lowering it as an LLVM atomic
 // operation. It returns the result of the operation and true if the call could
 // be lowered inline, and false otherwise.
-func (b *builder) createAtomicOp(call *ssa.CallCommon) (llvm.Value, bool) {
-	name := call.Value.(*ssa.Function).Name()
+func (b *builder) createAtomicOp(call *ssa.Function, args []llvm.Value) (llvm.Value, bool) {
+	name := call.Name()
 	switch name {
 	case "AddInt32", "AddInt64", "AddUint32", "AddUint64", "AddUintptr":
-		ptr := b.getValue(call.Args[0])
-		val := b.getValue(call.Args[1])
+		ptr := args[0]
+		val := args[1]
 		if strings.HasPrefix(b.Triple, "avr") {
 			// AtomicRMW does not work on AVR as intended:
 			// - There are some register allocation issues (fixed by https://reviews.llvm.org/D97127 which is not yet in a usable LLVM release)
@@ -35,8 +35,8 @@ func (b *builder) createAtomicOp(call *ssa.CallCommon) (llvm.Value, bool) {
 		// Return the new value, not the original value returned by atomicrmw.
 		return b.CreateAdd(oldVal, val, ""), true
 	case "SwapInt32", "SwapInt64", "SwapUint32", "SwapUint64", "SwapUintptr", "SwapPointer":
-		ptr := b.getValue(call.Args[0])
-		val := b.getValue(call.Args[1])
+		ptr := args[0]
+		val := args[1]
 		isPointer := val.Type().TypeKind() == llvm.PointerTypeKind
 		if isPointer {
 			// atomicrmw only supports integers, so cast to an integer.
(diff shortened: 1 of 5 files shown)


skabbes:
@aykevl I added tests for each individual call to the sync/atomic.* package.  Would you be willing to review this? I think it is pretty straightforward, as it is leveraging the existing code for rewriting atomics (no changes there).

The build failures seem completely unrelated, maybe due to go 1.18, or some CI configuration changes to windows.

skabbes:
I think windows is failing due to an upstream scoop issue:
https://github.com/ScoopInstaller/Scoop/issues/4917

skabbes:
I think https://github.com/tinygo-org/tinygo/pull/2853 will resolve build issues 🤞

deadprogram:
@skabbes please rebase against `dev` now to see if it fixes the build on this branch. Thanks!

skabbes:
@deadprogram all green!

skabbes:
FYI - as deadprogram has requested of me on other PRs - I can clean up / squash / whatever the commits once we have completed a code review.

aykevl:
While this probably works fine, I think there is a much more elegant way: by defining the functions (like `sync/atomic.AddInt32`) directly instead of changing all the calls to it. This will automatically make deferred calls work correctly without any special cases (and avoid the hairiness of it all).

Here is how to do that. First you can detect the special sync/atomic functions here:

```diff
diff --git a/compiler/compiler.go b/compiler/compiler.go
index dd7a8ea3..6981fc06 100644
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -773,6 +773,9 @@ func (c *compilerContext) createPackage(irbuilder llvm.Builder, pkg *ssa.Package
                        // Create the function definition.
                        b := newBuilder(c, irbuilder, member)
                        if member.Blocks == nil {
+                               if member.Pkg.Pkg.Path() == "sync/atomic" && token.IsExported(member.Name()) {
+                                       b.implementAtomicOp()
+                               }
                                continue // external function
                        }
                        b.createFunction()
```

And then you can fill them with atomic instructions here:

```go
func (b *builder) implementAtomicOp() {
	entry := b.ctx.AddBasicBlock(b.llvmFn, "entry")
	b.Builder.SetInsertPointAtEnd(entry)
	switch b.fn.Name() {
	case "AddInt32":
		ptr := b.llvmFn.Param(0)
		val := b.llvmFn.Param(1)
		oldVal := b.CreateAtomicRMW(llvm.AtomicRMWBinOpAdd, ptr, val, llvm.AtomicOrderingSequentiallyConsistent, true)
		// Return the new value, not the original value returned by atomicrmw.
		newVal := b.CreateAdd(oldVal, val, "")
		b.CreateRet(newVal)
	default:
		b.addError(b.fn.Pos(), "unknown atomic operation: "+b.fn.Name())
		b.CreateUnreachable()
	}
}
```

(This is not complete of course, it only implements `sync/atomic.AddInt32` and doesn't implement the AVR special case).

skabbes:
OK, thanks - I don't think I'll be able to get around to the changes. But maybe someone else can pick up the slack.

aykevl:
@skabbes ok, I've implemented this as I described, see #2920.

dgryski:
Closing as this was implemented in https://github.com/tinygo-org/tinygo/pull/2920

PR #2825, opened 2022-05-02
Drop support for go 1.15

### LINKED ISSUE #2817: "tinygo test io/ioutil" fails because os.PathError has not moved to io/fs as in upstream

dkegel-fastly:
1.15 specific files deleted.
1.16 specific files folded carefully into generic files, with goal of reducing diff with upstream.
Follows upstream 1.16 in making PathError etc. be aliases for the same errors in io/fs.

This fixes #2817 and lets us add io/ioutil to "make test-tinygo" on linux and mac.


aykevl (approved):
LGTM, thanks for the cleanup!

deadprogram:

PR #2830, opened 2022-05-05
avr: add full support for float64 and fix tests

aykevl:
With this change, you can run most tests under AVR:

```
$ go test -target=simavr -run Build -v
[...]
--- PASS: TestBuild (0.00s)
    --- PASS: TestBuild/alias.go (1.71s)
    --- PASS: TestBuild/string.go (2.07s)
    --- PASS: TestBuild/init_multi.go (1.26s)
    --- PASS: TestBuild/goroutines.go (4.22s)
    --- PASS: TestBuild/init.go (1.40s)
    --- PASS: TestBuild/map.go (4.84s)
    --- PASS: TestBuild/interface.go (3.44s)
    --- PASS: TestBuild/slice.go (1.85s)
    --- PASS: TestBuild/print.go (1.93s)
    --- PASS: TestBuild/sort.go (2.56s)
    --- PASS: TestBuild/float.go (2.58s)
    --- PASS: TestBuild/calls.go (2.27s)
    --- PASS: TestBuild/zeroalloc.go (1.24s)
    --- PASS: TestBuild/binop.go (2.26s)
    --- PASS: TestBuild/atomic.go (2.44s)
    --- PASS: TestBuild/go1.17.go (0.99s)
    --- PASS: TestBuild/structs.go (0.96s)
PASS
ok      github.com/tinygo-org/tinygo    9.966s
```

This has taken a lot of work behind the scenes, this PR only shows a small part of it! A number of patches were needed to Clang and compiler-rt to make compiler-rt work for AVR.

PR #2831, opened 2022-05-05
all: remove support for LLVM 11 and LLVM 12

aykevl:
This removes a lot of backwards compatibility cruft and makes it possible to start using features that need LLVM 13 or newer.
For example:

  * https://github.com/tinygo-org/tinygo/pull/2637
  * https://github.com/tinygo-org/tinygo/pull/2830


compiler/compiler_test.go
```diff
@@ -28,18 +28,11 @@ type testCase struct {
 func TestCompiler(t *testing.T) {
 	t.Parallel()
 
-	// Check LLVM version.
+	// Determine LLVM version.
 	llvmMajor, err := strconv.Atoi(strings.SplitN(llvm.Version, ".", 2)[0])
```
deadprogram, line 32: Should we fail test if `llvmMajor < 13`?

deadprogram (approved):

src/runtime/gc_conservative.go
```diff
@@ -233,14 +233,6 @@ func setHeapEnd(newHeapEnd uintptr) {
 // This function can be called again when the heap size increases. The caller is
 // responsible for copying the metadata to the new location.
 func calculateHeapAddresses() {
-	if GOARCH == "wasm" {
-		// This is a workaround for a bug in wasm-ld: wasm-ld doesn't always
-		// align __heap_base and when this memory is shared through an API, it
-		// might result in unaligned memory. For details, see:
-		// https://reviews.llvm.org/D106499
-		// It should be removed once we switch to LLVM 13, where this is fixed.
```
dgryski, line 241: Does this need to be left in place for llvm12 ?

src/runtime/gc_globals_conservative.go
```diff
@@ -4,21 +4,10 @@
...
 // This implementation marks all globals conservatively and assumes it can use
 // linker-defined symbols for the start and end of the .data section.
 func markGlobals() {
-	end := globalsEnd
-	if GOARCH == "wasm" {
-		// This is a workaround for a bug in wasm-ld: wasm-ld doesn't always
-		// align __heap_base and when this memory is shared through an API, it
-		// might result in unaligned memory. For details, see:
-		// https://reviews.llvm.org/D106499
-		// It should be removed once we switch to LLVM 13, where this is fixed.
```
dgryski, line 20: Does this need to be left in place for llvm12 ?

Edit: ignore, thought this was only dropping support for llvm11, not 11 and 12.

dgryski (approved):
LGTM

compiler/compiler_test.go
aykevl: Why? We're dropping support for LLVM < 13.

deadprogram:
Thanks @aykevl

Issue #2795, opened 2022-04-23
tinygo/lib/compiler-rt/lib/builtins/absvdi2.c: could not locate Clang headers


arbourd:
Strange new behaviour when building an arm64 container from source (which worked fine prior to `v0.22.0`). I don't have any experience with `clang` or `llvm` so I'd love some help understanding why this is happening.

Ran:
- `make llvm-source` ✅
- `make llvm-build` ✅
- `make wasi-libc` ✅
- `make` ✅
- `make release` ❌

Fails on the last bit of `build/release` https://github.com/tinygo-org/tinygo/blob/v0.22.0/Makefile#L644:

```
#0 148.0 ./build/tinygo build-library -target=cortex-m0     -o build/release/tinygo/pkg/thumbv6m-unknown-unknown-eabi-cortex-m0/compiler-rt     compiler-rt
#0 148.1 error: failed to build /root/tinygo/lib/compiler-rt/lib/builtins/absvdi2.c: could not locate Clang headers
```

aykevl:
Can you tell a bit more about the environment? OS, etc

arbourd:
Absolutely! Thank for taking the time @aykevl.

I am attempting to build both `amd64` and `arm64` on Linux using a different Dockerfile (for our purposes). It is here if that helps. It is located here: https://github.com/suborbital/subo/blob/6e49f277e7c064c9f8a2de08b878adb0c66c9c9b/builder/docker/tinygo/Dockerfile.base#L28-L29

We're purposely skipping some other steps. With `v0.21.0` we've had no issues building. I'm guessing it might have been [this change](https://github.com/tinygo-org/tinygo/commit/3883550c446fc8b14115366284316290f3d3c4b1) that is causing issues, I just don't understand why the clang headers would have moved 😅

aykevl:
I don't think it's https://github.com/tinygo-org/tinygo/commit/3883550c446fc8b14115366284316290f3d3c4b1, but you can confirm by reverting this commit.

TinyGo assumes that the Clang headers are in a particular location. The logic is here: https://github.com/tinygo-org/tinygo/blob/release/builder/env.go#L17
I see you're storing the LLVM source in a non-default location, that may be causing these issues. If you really want to know which commit is causing this, you could try bisecting this issue.

aykevl:
[...]

PR #2834, opened 2022-05-06
makefile: Use release built tinygo in Makefile

arbourd:
Closes https://github.com/tinygo-org/tinygo/issues/2795

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -710,12 +710,12 @@ endif
 	@cp -rp llvm-project/compiler-rt/LICENSE.TXT  build/release/tinygo/lib/compiler-rt-builtins
 	@cp -rp src                          build/release/tinygo/src
 	@cp -rp targets                      build/release/tinygo/targets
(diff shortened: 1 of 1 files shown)


dgryski:
Given that we copy the tinygo binary from the build directory to the release directory earlier in the makefile, I'm not sure this patch accomplishes anything.

arbourd:
It's referenced in #2795. If you're like me and are using a LLVM directory that isn't the default, it doesn't know where the headers are. The built "release" binary does, because it's properly copied the headers over in an earlier step.

This solution was suggested by @aykevl: https://github.com/tinygo-org/tinygo/issues/2795#issuecomment-1111179833

dgryski:
Ah, thanks for the background. If this works, then LGTM.

aykevl:
@dgryski the `tinygo` binary locates some directories relative to itself, so yes this changes some things.

aykevl (approved):

deadprogram:
Thank you for working on this @arbourd and thanks everyone else for reviewing. Now merging.

Issue #2760, opened 2022-04-09
Add support for Mac M1 (foremost using Docker)


supcik:
I just managed to compile and install tinygo on the new Mac M1 inside a Docker container. This allows me to use tinygo on the M1 while waiting for a native version.

I think that it is not so hard to extend the _official_ Docker image "tinygo/tinygo" for arm64 as well (multi-architecture image).

A small issue is the [extraction of `xtensa-esp32-elf-ld`](https://github.com/tinygo-org/tinygo/blob/3883550c446fc8b14115366284316290f3d3c4b1/Dockerfile#L25).

If `xtensa-esp32-elf-ld` is really needed (can someone confirm ?), then it is possible to compile it using the instructions from [the espressif web site](https://espressif-docs.readthedocs-hosted.com/projects/esp-idf/en/latest/get-started/linux-setup-scratch.html)

_Note 1: I just had to use the default branch instead of the `esp-2019r2` tag described on the web page_.
_Note 2: The tool can't be compiled as root, and we have to handle this in the Dockerfile_

So, if `xtensa-esp32-elf-ld` is required, it is possible to compile it instead of extracting from a tar file. It will take more time, but it allows getting a native version of the linker.

I would be happy to work on this issue if you think that it makes sense.

deadprogram:
> So, if xtensa-esp32-elf-ld is required, it is possible to compile it instead of extracting from a tar file. It will take more time, but it allows getting a native version of the linker.

This is a requirement for compiling for esp8266 and esp32 targets.

supcik:
Thank you for your reply. I am working on a PR.

Wouldn't it be useful to mention the installation of xtensa-esp32-elf-ld in the "Build from source" documentation page (https://tinygo.org/docs/guides/build/) ?

aykevl:
I've been working on replacing `xtensa-esp32-elf-ld` with native ld.lld, but I need to finish it some day: https://github.com/espressif/llvm-project/pull/53 (it almost works but still has some issues with some corner cases).
Might be worth including that patch in our fork of LLVM though to avoid the `xtensa-esp32-elf-ld` dependency. It should be good enough for TinyGo.

aykevl:
I think that TinyGo should also work with Rosetta 2 emulation. Or are there any problems with it?

supcik:
[...]

PR #2835, opened 2022-05-06
ci: Add linux/arm64 build

arbourd:
Based on the discussion in #2760, this PR adds an `arm64` cross target and test.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -346,3 +346,103 @@ jobs:
           path: |
             /tmp/tinygo.linux-arm.tar.gz
             /tmp/tinygo_armhf.deb
+  build-linux-arm64:
+    # Build ARM64 Linux binaries, ready for release.
+    # It is set to "needs: build-linux" because it modifies the release created
+    # in that process to avoid doing lots of duplicate work and to avoid
+    # complications around precompiled libraries such as compiler-rt shipped as
+    # part of the release tarball.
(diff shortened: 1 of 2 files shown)


.github/workflows/linux.yml
```diff
@@ -346,3 +346,105 @@ jobs:
...
+    # in that process to avoid doing lots of duplicate work and to avoid
+    # complications around precompiled libraries such as compiler-rt shipped as
+    # part of the release tarball.
+    runs-on: ubuntu-18.04
+    needs: build-linux
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: true
```
aykevl: This can be removed. It slows down the build and the submodules that are actually needed (binaryen) are pulled specifically when needed.

aykevl (commented):
Looks good to me, except for one detail.
This probably means we're also going to be publishing linux/arm64 binaries as part of the release. I think that's fine, just wanted to check whether there are any comments on that.

arbourd:
@aykevl I'm happy to release separate artifacts, one per architecture. I was just copying the armhf build, but I think it does make sense for these things to be different.

deadprogram:
So that means we can merge this, correct?

aykevl (approved):
Yes, LGTM.

deadprogram:
Thank you very much @arbourd now merging.

PR #2837, opened 2022-05-08
Skip slice-copy test for LLVM < 14

### LINKED ISSUE #2836: slice-copy interp test fails with Go 1.16 / LLVM 13
I'm not sure which of these is the issue, as we don't have a system with the intermediate combinations. 

When building on Fedora 36, everything passes with Go 1.18 and LLVM 14. When building on Fedora 35 with Go 1.16 and LLVM 13, the slice-copy test fails, and the diff is:
```diff
diff --git a/interp/testdata/slice-copy.out.ll b/interp/testdata/slice-copy.out.ll
index 28175643..9f88df09 100644
--- a/interp/testdata/slice-copy.out.ll
+++ b/interp/testdata/slice-copy.out.ll
@@ -1,6 +1,8 @@
 target datalayout = "e-m:e-i64:64-f80:128-n8:16:32:64-S128"
 target triple = "x86_64--linux"
 
+@"main$alloc.1" = internal unnamed_addr constant [6 x i8] c"\05\00{\00\00\04", align 8
+
 declare void @runtime.printuint8(i8) local_unnamed_addr
 
 declare void @runtime.printint16(i16) local_unnamed_addr
@@ -15,6 +17,7 @@ entry:
   call void @runtime.printuint8(i8 3)
   call void @runtime.printuint8(i8 3)
   call void @runtime.printint16(i16 5)
-  call void @runtime.printint16(i16 5)
+  %int16SliceDst.val = load i16, i16* bitcast ([6 x i8]* @"main$alloc.1" to i16*), align 2
+  call void @runtime.printint16(i16 %int16SliceDst.val)
   ret void
 }
```
No other interp tests appear to fail, and all other tests pass otherwise.

QuLogic:
Assuming we don't want to do anything more complex.

Fixes #2836

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -3,6 +3,7 @@ package interp
 import (
 	"io/ioutil"
 	"os"
+	"strconv"
 	"strings"
 	"testing"
 
@@ -10,6 +11,12 @@ import (
 )
 
 func TestInterp(t *testing.T) {
+	llvmVersion, err := strconv.Atoi(strings.Split(llvm.Version, ".")[0])
+	if err != nil {
+		// Note: this should never happen and if it does, it will always happen
+		// for a particular build because llvm.Version is a constant.
+		panic(err)
+	}
 	for _, name := range []string{
 		"basic",
 		"phi",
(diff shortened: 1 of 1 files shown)


aykevl (approved):

aykevl:
IR output often varies by LLVM version. We're only testing IR output in CI for the latest/current LLVM version that we support to avoid churn. For older LLVM versions, we do fewer tests: mainly smoke tests (see .circleci/config.yml). I guess we could also do at least some tests that the produced binaries behaves as expected, but we currently don't.

QuLogic:
I run as much as possible on Fedora, but of course that's only for releases.

PR #2838, opened 2022-05-08
Makefile: fix forced rebuild of LLVM

aykevl:
This is a small change that makes sure not to force a rebuild of LLVM every time.

For example, I might run:

    make llvm-source
    make llvm-build ASSERT=1

And then I might make some temporary changes to LLVM to test out a patch for example. So I run:

    make llvm-build

...and my whole build cache gets destroyed.

This commit addresses this issue by not forcing a re-run of CMake with every `make llvm-build` invocation.


deadprogram:
Very good idea @aykevl now merging.

Issue #2824, opened 2022-05-01
armhf linux build fails "tinygo test os" in TestDirFS?


dankegel:
On raspberry pi 4 64 bits, I installed https://github.com/tinygo-org/tinygo/releases/download/v0.23.0/tinygo_0.23.0_armhf.deb (and the right libc++, see #2815 ) and go 1.17, and ran "tinygo test os".

The test fails with 
```
--- FAIL: TestDirFS (0.00s)
    TestFS found errors:
        b: failed TestReader:
        	Seek(0, 1) from EOF = 0, operation not supported, want 0, nil
        dir/x: failed TestReader:
        	Seek(0, 1) from EOF = 0, operation not supported, want 0, nil
        a: failed TestReader:
        	Seek(0, 1) from EOF = 0, operation not supported, want 0, nil
    FailNow is incomplete, requires runtime.Goexit()
```

Same result with go 1.18 (with the fixed tinygo binary from #2823).

QuLogic:
This also [fails on i686](https://koji.fedoraproject.org/koji/taskinfo?taskID=86789232), so seems like maybe a 32-bit problem in general?

QuLogic:
In `os_test.go`, I see
```
func TestSeek(t *testing.T) {
        if runtime.GOARCH == "386" || runtime.GOARCH == "arm" {
                t.Log("TODO: implement seek for 386 and arm")
                return
        }
```
so this probably needs to be translated into `TestFS` as well, but I don't think there's a copy of it in tinygo that can be patched.

QuLogic:
I also see
```
--- FAIL: TestSymlink (0.00s)
    TestFS found errors:
        hello: failed TestReader:
        	Seek(0, 1) from EOF = 0, operation not supported, want 13, nil
        hello.link: failed TestReader:
        	Seek(0, 1) from EOF = 0, operation not supported, want 13, nil
    FailNow is incomplete, requires runtime.Goexit()
FAIL
```
but this one I cannot figure how to get from `TestSymlink` to a `Seek`.

QuLogic:
Ah, the latter is from `testing/fstest`; I worked around that by skipping it on i686.

deadprogram:
Now that we have 64-bit ARM Linux build, it this still a problem?

dankegel:
Well... Yeah, it's still a problem for 32 bit systems, and will be until tinygo supports go's assembly implementation of seek,
[...]

PR #2840, opened 2022-05-09
Fix 32bit testing

QuLogic:
Or really, it's more of a workaround, by simply disabling the tests that can't run, as mentioned in #2824.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -328,7 +328,6 @@ TEST_PACKAGES_FAST = \
 # Additional standard library packages that pass tests on individual platforms
 TEST_PACKAGES_LINUX := \
 	archive/zip \
-	compress/flate \
 	compress/lzw \
 	crypto/hmac \
 	debug/dwarf \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -295,13 +295,16 @@ TEST_PACKAGES_FAST = \
...
 TEST_PACKAGES_LINUX := \
 	archive/zip \
-	compress/flate \
 	compress/lzw \
 	debug/dwarf \
 	debug/plan9obj \
 	io/fs \
-	io/ioutil \
+	io/ioutil
+ifneq ($(shell uname -p),i686)
```
dankegel: On my 64 bit Ubuntu system:
```
$ uname -p
x86_64
```

And for grins, on my 64 bit mac:
```
$ uname -p
i386
```

Maybe use getconf LONG_BIT ?  That should work on all unix systems.

QuLogic: Sure, I wasn't sure that'd end up very portable, but that suggestion sounds reasonable.

aykevl:
The `syscall.seek` issue will be fixed by #3103.

I can reproduce the "out of memory" issue on my laptop though:

```
$ GOARCH=arm tinygo test -v compress/flate
qemu: uncaught target signal 6 (Aborted) - core dumped
FAIL    compress/flate  1337.215s
FAIL
```

Not sure what's causing it, as both amd64 and 386 should have a heap size of 1GB. Also, I don't see why compress/flate would use such a large amount of RAM during tests. Perhaps the GC has too many false positives so it keeps memory around?

QuLogic:
I no longer do any 32-bit builds, so not sure what to do with this PR; maybe just close it?

deadprogram:
Given the age of the PR and the content, I will close. We can open something new should we pursue this.

Thank you everyone!

Issue #2842, opened 2022-05-09
"make test-corpus" fails on linux in google/go-patchutils?


dkegel-fastly:
On my fresh ubuntu 22.04 VM, with a freshly built dev, "make test-corpus" crashes if
google/go-patchutils is in corpus.yaml.
(Evidently that is a recent addition, but not that recent, https://github.com/tinygo-org/tinygo/commit/cbfa85be6ad018b447f2c47401f4ebf8193b0887 )

Here's a log.
[bad.txt](https://github.com/tinygo-org/tinygo/files/8652720/bad.txt)

dkegel-fastly:
A corpus.yaml containing just google/go-patchutils fails in the linux vm but doesn't fail on my mac (outside the vm, I mean).

dankegel:
Confirming it fails on a plain old linux box, no VM involved.

dgryski:
I'll take a look at this.

dgryski:
Passes on mac. Will need to test this in a VirtualBox Ubuntu.

dgryski:
Got this failing on my ubuntu vm.  It's an error in *building* the code, not in *running* the code.  Interesting.

dgryski:
```
2022/05/10 21:49:52 compiling: github.com/google/go-patchutils$string.219
```
Possibly the function that's failing to compile.

dgryski:
With llvm-13 on Linux:
```
2022/05/11 04:19:26 compiling: github.com/google/go-patchutils$string.219
panic: runtime error: index out of range [-1]

goroutine 86 [running]:
github.com/tinygo-org/tinygo/interp.(*runner).run(0xc00460e000, 0xc002f9bc20, {0xc004ea3e90, 0x4, 0x0?}, 0xc004245b30, {0xc004d054a0, 0x14})
	/home/dgryski/go/src/github.com/tinygo-org/tinygo/interp/interpreter.go:902 +0x61f9
github.com/tinygo-org/tinygo/interp.(*runner).run(0xc00460e000, 0xc002f9ba40, {0xc00414d8d0, 0x3, 0x0?}, 0xc004245680, {0xc004ef7ff0, 0x10})
	/home/dgryski/go/src/github.com/tinygo-org/tinygo/interp/interpreter.go:501 +0x80b4
```

where line 902 is:
```
        return nil, mem, r.errorAt(bb.instructions[len(bb.instructions)-1], errors.New("interp: reached end of basic block without terminator"))
```

[...]

PR #2843, opened 2022-05-09
Comment out google/go-patchutils in testdata/corpus.yaml

dkegel-fastly:
Works around 
https://github.com/tinygo-org/tinygo/issues/2842


deadprogram:
I think this PR should no longer be needed thanks to #2863 if I am not mistaken.

dkegel-fastly:
No longer needed, whew.

PR #2847, opened 2022-05-11
RP2040 sleep improvements

kenbell:
This change moves sleep implementation from a busy loop to using a timer alarm, so the CPU can go idle using 'wfe'.

Power usage improvements are 'modest', but clear.

Test environment:

- board: feather-rp2040
- code: blinky1
- power: 4.2V through battery connector
- tool: Nordic Power Profiler Kit II

Before:
no-sleep

After:
with-sleep

Approx 3mA improvement, or 13.5% battery usage.

### DIFF
--- a/src/runtime/runtime_rp2040.go
+++ b/src/runtime/runtime_rp2040.go
@@ -12,6 +12,9 @@ import (
 // machineTicks is provided by package machine.
 func machineTicks() uint64
 
+// machineLightSleep is provided by package machine.
+func machineLightSleep(uint64)
+
 type timeUnit uint64
 
 // ticks returns the number of ticks (microseconds) elapsed since power up.
@@ -32,6 +35,16 @@ func sleepTicks(d timeUnit) {
 	if d == 0 {
 		return
 	}
+
+	if hasScheduler {
+		// With scheduler, sleepTicks may return early if an interrupt or
+		// event fires - so scheduler can schedule any go routines now
+		// eligible to run
+		machineLightSleep(uint64(d))
+		return
+	}
+
+	// Busy loop
 	sleepUntil := ticks() + d
 	for ticks() < sleepUntil {
 	}
--- a/src/machine/machine_rp2040.go
+++ b/src/machine/machine_rp2040.go
@@ -86,6 +86,11 @@ func ticks() uint64 {
 	return timer.timeElapsed()
 }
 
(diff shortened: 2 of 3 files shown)


deadprogram:
Looks good to me, anyone else have feedback on this change?

deadprogram:
cc @ysoldak @lincolngill @soypat

src/runtime/runtime_rp2040.go
```diff
@@ -32,6 +35,16 @@ func sleepTicks(d timeUnit) {
 	if d == 0 {
 		return
 	}
+
+	if hasScheduler {
```
soypat, line 39: Given your comment above 
> (up to approx 72mins per RP2040 datasheet, 4.6.3. Alarms).

Should there not be a check here similar to `if hasScheduler && d < 72*minute` ?

ysoldak:
Tested for regression.
Slightly modified "blinky2", printing out `time.Now().UnixMilli()`.

Result: `time.Sleep()` still works as expected on `nano-rp2040` target.

src/runtime/runtime_rp2040.go
ysoldak: Agree.
And after that we probably want a follow-up ticket to not busy loop for periods higher than 72 mins. Like light sleeping in chunks of 70 mins.

kenbell: @soypat @ysoldak  - thanks for review.  I don't think we need to change anything here.  The sleep duration is slightly unsual when there's a scheduler - what it really means is 'there is currently no go routine that will be runnable for at least d period'.  It's safe to return early, and even desirable, if a interrupt handler has done something to make another go routine runnable (e.g. sent a message on a channel).

The max sleep from the datasheet is due to truncation because the RP2040 hardware has a smaller register for alarms.  All that means is the 'lightSleep' may return early if the scheduler asks to sleep for > 72mins.  This is safe and normal, since the scheduler will look for any go routines to run, and if none - sleep once again.

We could do `d = min(d, 72mins)` - but it's of no real benefit, makes the code a little more complex and just (potentially) increases the binary size by a few bytes.

soypat (approved):

deadprogram:
Thanks everyone! Now merging.

kenbell:
Linking to issue https://github.com/tinygo-org/tinygo/issues/2877 which this fixes by allowing scheduler to run after interrupts.

PR #2853, opened 2022-05-17
build: install scoop without update to avoid bug in ScoopInstaller

deadprogram:
This PR in intended to provide a temporary work-around for https://github.com/ScoopInstaller/Scoop/issues/4917


deadprogram:
Got past that error in the Windows build, so changed to "ready for review".

deadprogram:
All green! :smile_cat:

dkegel-fastly (approved):
LGTM FWIW :-)

Issue #2806, opened 2022-04-25
(constant -4 of type uintptr) overflows uintptr


kyleconroy:
TinyGo can't compile the the [pointer_unsafe.go](https://github.com/protocolbuffers/protobuf-go/blob/master/internal/impl/pointer_unsafe.go#L142) file from https://github.com/protocolbuffers/protobuf-go. I get the following error message:

```
unsafe.Sizeof(unsafe.Pointer(nil)) - unsafe.Sizeof((MessageState literal)) (constant -4 of type uintptr) overflows uintptr
```

aykevl:
This comment above it is relevant:

```
// Static check that MessageState does not exceed the size of a pointer.
```

So it appears that `MessageState` is larger than a pointer.
Without further information, this would be hard to debug. How can I reproduce this?

---

That said, this is most certainly unsafe code so that throws away all promises of portability. But quickly looking at it, I don't see why it fails.

skabbes:
My guess is that possibly TinyGo (or maybe the specific architecutre you are targeting) is considering a `uintptr` to be unsigned (the `u` in uintptr), and therefore -4 is an invalid value (or it is wrapping around to a very very large number).

skabbes:
I had a comment previously that I deleted, I did not read aykevl's comment closely enough.

The [MessageState](https://github.com/protocolbuffers/protobuf-go/blob/8a7ba0762cb3b39fc0536379eac2f7fa5796f187/internal/impl/message_reflect.go#L360) struct embeds a [few structs](https://github.com/protocolbuffers/protobuf-go/blob/8a7ba0762cb3b39fc0536379eac2f7fa5796f187/internal/pragma/pragma.go#L11) with 0 fields.

My _new_ guess is that these structs take space in TinyGo, but do not take space in "regular Go".

kyleconroy:
> Without further information, this would be hard to debug. How can I reproduce this?

Sorry about that. Here's the reproduction.

```sh
go mod init github.com/tinygo-issues
go get google.golang.org/protobuf
```

```go
package main

import (
	"fmt"

	_ "google.golang.org/protobuf/runtime/protoimpl"
)

func main() {
	fmt.Println("protoc")
}
```

```sh
tinygo build -o example.wasm -gc=leaking -scheduler=none -wasm-abi=generic -target=wasi
# google.golang.org/protobuf/internal/impl
[...]

PR #2854, opened 2022-05-18
compiler: make sure zero-sized array types are as small as possible

skabbes:
As part of an https://github.com/tinygo-org/tinygo/issues/2667 - an effort to use protobufs with tinygo (particularly with WASM):

Fixes https://github.com/tinygo-org/tinygo/issues/2806

**Context**
The protobuf lib at `google.golang.org/protobuf/runtime/protoimpl` includes some code includes some code that acts as a "non-comparable" static check.  I have simplified the code to its minimal repro:

```go
package main

import (
	"unsafe"
)

// DoNotCompare can be embedded in a struct to prevent comparability.
type DoNotCompare [0]func()

type MessageState struct {
	DoNotCompare
	data *int32
}

// Static check that MessageState does not exceed the size of a pointer.
const _ = unsafe.Sizeof(unsafe.Pointer(nil)) - unsafe.Sizeof((MessageState{}))

func main() {
	println("protoc minimal example")
}
```

**Problem**

`alignof([0]func())` was 2 words large, should have only been 1.

### DIFF
--- a/compiler/sizes.go
+++ b/compiler/sizes.go
@@ -21,6 +21,12 @@ func (s *stdSizes) Alignof(T types.Type) int64 {
 	// of alignment of the elements and fields, respectively.
 	switch t := T.Underlying().(type) {
 	case *types.Array:
+		if t.Len() == 0 {
+			// 0-sized arrays, always have 0 size.
+			// And from the spec, should have an alignment of _at least_ 1
+			return 1
+		}
+
 		// spec: "For a variable x of array type: unsafe.Alignof(x)
 		// is the same as unsafe.Alignof(x[0]), but at least 1."
 		return s.Alignof(t.Elem())
--- a/testdata/reflect.go
+++ b/testdata/reflect.go
@@ -52,6 +52,12 @@ func main() {
 	println("\nvalues of interfaces")
 	var zeroSlice []byte
 	var zeroFunc func()
+	// by embedding a 0-array func type in your struct, it is not comparable
+	type doNotCompare [0]func()
+	type notComparable struct {
+		doNotCompare
+		data *int32
+	}
 	var zeroMap map[string]int
 	var zeroChan chan int
 	n := 42
@@ -169,6 +175,10 @@ func main() {
 	assertSize(reflect.TypeOf("").Size() == unsafe.Sizeof(""), "string")
 	assertSize(reflect.TypeOf(new(int)).Size() == unsafe.Sizeof(new(int)), "*int")
 	assertSize(reflect.TypeOf(zeroFunc).Size() == unsafe.Sizeof(zeroFunc), "func()")
+
+	// make sure embedding a zero-sized "not comparable" struct does not add size to a struct
+	assertSize(reflect.TypeOf(doNotCompare{}).Size() == unsafe.Sizeof(doNotCompare{}), "[0]func()")
+	assertSize(unsafe.Sizeof(notComparable{}) == unsafe.Sizeof((*int32)(nil)), "struct{[0]func(); *int32}")
 
 	// Test that offset is correctly calculated.
 	// This doesn't just test reflect but also (indirectly) that unsafe.Alignof


skabbes:
Needs https://github.com/tinygo-org/tinygo/pull/2853 to be merged to pass windows build.

@deadprogram would you be the right person to review this?  I think it is pretty straightforward.

deadprogram:
@skabbes can you please rebase against `dev` now to see if CI passing? Thanks!

skabbes:
@deadprogram all green!

deadprogram:
@skabbes it would have been better to rebase without that last merge commit. I usually use the command line for this e.g.

```
git checkout dev
git pull --rebase origin dev
git checkout my-feature-branch
git rebase dev
git push myfork my-feature-branch -f
```

skabbes:
Ya, some projects prefer merge - some rebase. I'll squash them and force-push if that's how you prefer in tinygo.

deadprogram:
I see we need to clarify that on https://tinygo.org/docs/guides/contributing/

skabbes:
Happy to do it however the team finds best for them - I probably could have figured that out by looking at a few other PRs. But appreciate you letting me know.  I'll rebase off dev from now on - and already did it for this branch.

If you want me to squash these too, just say the word!

skabbes:
> I see we need to clarify that on https://tinygo.org/docs/guides/contributing/

Added a PR to clarify:
https://github.com/tinygo-org/tinygo-site/pull/267

compiler/sizes.go
```diff
@@ -21,6 +21,11 @@ func (s *stdSizes) Alignof(T types.Type) int64 {
 	// of alignment of the elements and fields, respectively.
 	switch t := T.Underlying().(type) {
 	case *types.Array:
+		if t.Len() == 0 {
+			// 0-sized arrays, always have 0 size - and therefore can be minimally aligned
+			return s.PtrSize
```
aykevl: I think this should have an alignment of 1, not `PtrSize`. You can test this with a `Sizeof(struct{ a [0]int; b byte}{})`.

skabbes: Ya, I went back and forth on this - thank you for the clarification.

skabbes: I can update to 1 (I thought originally 1 was the correct value too).  But I was unable to corroborate that with the upstream go compiler (perhaps that is not / should not be the source of truth in this case however).

```go
package main

import (
	"unsafe"
)

type ZeroInt [0]int

type ZeroIntRaw struct {
	f [0]int
	b byte
}

type ZeroIntEmbedded struct {
	ZeroInt
	b byte
}

func main() {
	println("sizeof ZeroInt", unsafe.Sizeof(ZeroInt{}))
	println("alignof ZeroInt", unsafe.Alignof(ZeroInt{}))

	println("sizeof ZeroIntRaw", unsafe.Sizeof(ZeroIntRaw{}))
	println("alignof ZeroIntRaw", unsafe.Alignof(ZeroIntRaw{}))

	println("sizeof ZeroIntEmbedded", unsafe.Sizeof(ZeroIntEmbedded{}))
	println("alignof ZeroIntEmbedded", unsafe.Alignof(ZeroIntEmbedded{}))
}

```

The go compiler reports these all as `8`
```
sizeof ZeroInt 0
alignof ZeroInt 8
sizeof ZeroIntRaw 8
alignof ZeroIntRaw 8
sizeof ZeroIntEmbedded 8
alignof ZeroIntEmbedded 8
```

skabbes: (M1 Mac is the architecture I tried this on).

deadprogram:
I think you should perhaps remove commit https://github.com/tinygo-org/tinygo/pull/2854/commits/d0e10e8b8bbc9dbcbf11cef3a64185dbb00825ea now since it was basically replaced by https://github.com/tinygo-org/tinygo/pull/2854/commits/8a4eecf4aedcd0023571986244a6514287dbed3f

skabbes:
> I think you should perhaps remove commit [d0e10e8](https://github.com/tinygo-org/tinygo/commit/d0e10e8b8bbc9dbcbf11cef3a64185dbb00825ea) now since it was basically replaced by [8a4eecf](https://github.com/tinygo-org/tinygo/commit/8a4eecf4aedcd0023571986244a6514287dbed3f)

OK, so I assume that you do no rewrite the commits and squash them when PRs are merged (as many other project do).  You would like me to have a perfectly clean commit history, merged as-is into dev.

I'll just squash them both together in that case, since there is no point in having so many commits in dev.  But there _is_ value in having them here in this PR, so everyone review can see my thought process as I went through the changes.

deadprogram:
@skabbes you totally get what we're trying to do, and thank you for the contribution and also for being a cool collaborator!

skabbes:
@deadprogram thanks! do you mind reviewing the code?

deadprogram:
Thanks for making the requested changes @skabbes now merging.

dgryski:
Seems reasonable.  I know in the past there have been issues with trailing 0-byte elements in structs.  Is there any danger of that happening again here?

skabbes:
I unfortunately do not have the context to answer that, but it as least seems that the previous intent of the code was that 0-sized structs have alignof = 1.  I don't see a reason why 0-sized _array_ types should be any different.

https://github.com/tinygo-org/tinygo/blob/4c7449efe5f220eaa4ec0b76af83a0bc8ec0eac2/compiler/sizes.go#L55-L59

aykevl:
After some testing, I found that this PR isn't entirely correct either. Proof:

```go
package main

import "unsafe"

func main() {
        v := struct {
                a byte
                b [0]func()
                c byte
        }{}
        println("unsafe.Alignof: ", int(unsafe.Alignof([0]func(){})))
        println("unsafe.Offsetof:", int(unsafe.Offsetof(v.c)))
        println("actual offset:  ", int(uintptr(unsafe.Pointer(&v.c))-uintptr(unsafe.Pointer(&v))))
}
```

The `unsafe.Offsetof` and `actual offset` values must match, but they do not on 32-bit architectures (on the latest dev branch with this PR included). For example, for wasm:

```
unsafe.Alignof:  1
unsafe.Offsetof: 1
actual offset:   4
```

I think this is the proper fix:

```diff
diff --git a/compiler/sizes.go b/compiler/sizes.go
index 57108323..36b8d1ab 100644
--- a/compiler/sizes.go
+++ b/compiler/sizes.go
@@ -51,6 +51,8 @@ func (s *stdSizes) Alignof(T types.Type) int64 {
                if t.Info()&types.IsString != 0 {
                        return s.PtrSize
                }
+       case *types.Signature:
+               return s.PtrSize
        }
        a := s.Sizeof(T) // may be 0
        // spec: "For a variable x of any type: unsafe.Alignof(x) is at least 1."
```

This is a bit different from upstream Go because func values in TinyGo consist of two pointers instead of one.

aykevl:
Also, to add: TinyGo does not necessarily match Go when it comes to size and alignment. Always check these things by checking against Clang or by comparing pointer values/indices.
The `unsafe` package can't be used either because it is computed using the values from compiler/sizes.go, and must adhere to what LLVM will make of it (not the other way around).

deadprogram:
Seems I merged this a little too quickly, then. @skabbes are you able to submit another PR with this change?

skabbes:
@aykevl thank you - your comment really clarified a mental model for me.  The unsafe package simply reports what they layout was - it does not determine it.  compiler/sizes.go kind sort of "configures" LLVM, but at the end of the day whatever memory layout clang generates is the source of truth.

@deadprogram I'll give this another shot - add some tests that check the invariants Ayke laid out above.

aykevl:
> @aykevl thank you - your comment really clarified a mental model for me. The unsafe package simply reports what they layout was - it does not determine it. compiler/sizes.go kind sort of "configures" LLVM, but at the end of the day whatever memory layout clang generates is the source of truth.

Yes, basically. Except this bit:

> compiler/sizes.go kind sort of "configures" LLVM,

It doesn't. LLVM does its own thing. The only thing compiler/sizes.go is used for, is `unsafe.Sizeof`/`unsafe.Alignof`/`unsafe.Offsetof`. Of these, `unsafe.Sizeof` and `unsafe.Alignof` are used by the reflect package for things like `Type.Size()`, `Type.Align()`, and `Value.Field(i)`. So essentially it has to match whatever LLVM will do to avoid tricky bugs.

skabbes:
Perfect, that plus a little extra reading of compiler.go - I think I got it.

I put your suggested changes up for PR https://github.com/tinygo-org/tinygo/pull/2862, along with to some tests for offsetof alignof and the _actual_ offset.

PR #2857, opened 2022-05-20
machine/badger2040: support for Badger 2040

deadprogram:
This PR provides support for Badger 2040.

### DIFF
--- a/src/machine/board_badger2040.go
+++ b/src/machine/board_badger2040.go
@@ -1,0 +1,81 @@
+//go:build badger2040
+// +build badger2040
+
+// This contains the pin mappings for the Badger 2040 Connect board.
+//
+// For more information, see: https://shop.pimoroni.com/products/badger-2040
+// Also
+// - Badger 2040 schematic: https://cdn.shopify.com/s/files/1/0174/1800/files/badger_2040_schematic.pdf?v=1645702148
+//
+package machine
+
+const (
+	LED Pin = GPIO25
+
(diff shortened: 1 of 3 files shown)


deadprogram:
@conejoninja it seems this is working now for you as expected? I changed to "ready for review".

src/machine/board_badger2040.go
```diff
@@ -0,0 +1,81 @@
...
+	I2C0_SDA_PIN Pin = GPIO4
+	I2C0_SCL_PIN Pin = GPIO5
+
+	I2C1_SDA_PIN Pin = NoPin
+	I2C1_SCL_PIN Pin = NoPin
+)
+
+// SPI pins.
+const (
+	SPI0_SCK_PIN Pin = NoPin
```
conejoninja: should we define SPI0_SCK_PIN as EPD_SCK_PIN, same for EPD_SDO_PIN ?

conejoninja:
It's working fine. We still miss a few things:

- Exposed pins on the back (left) : 3v, sda, scl, int, tx, rx, swc, swd
- QSPI pins for flash access
- USB-CDC :(


*Neither buttons nor I²C were tested yet.

src/machine/board_badger2040.go
deadprogram: Done.

deadprogram:
Thanks for the review @conejoninja now merging.

Issue #2859, opened 2022-05-20
"tinygo test os" fails to build on M1 mac due to extra $INODE64 suffix on some syscalls


dkegel-fastly:
With today's dev:
```
% build/tinygo test os     
tinygo:ld.lld: error: undefined symbol: _readdir_r$INODE64
>>> referenced by /var/folders/hw/6t6vc0192yz_29lh30b3b7hh0000gp/T/tinygo207431203/main.o
```

This is because Apple is sunsetting the old compatibility suffixes as planned by not requiring them on new architectures.

deadprogram:
Closing since resolved "in-release".

PR #2860, opened 2022-05-20
darwin: remove now redundant syscall suffix for arm64

dkegel-fastly:
Fixes https://github.com/tinygo-org/tinygo/issues/2859


deadprogram:
Thank you for the quick fix @dkegel-fastly now merging.

PR #2863, opened 2022-05-21
interp: fix some buggy localValue handling

aykevl:
Bug:

 1. `fn.locals[v.value]` returns 0 (the default value) if `v.value` is not part of the fn.locals map.
 2. `locals[fn.locals[v.value]]` then returns the first local value, which is usually non-nil
 3. This incorrect value is then used as the operand value.

The manifestation of this convoluted bug was https://github.com/tinygo-org/tinygo/issues/2842. It didn't occur more often probably because it only seems to happen in practice with inline assembly.

Fixes https://github.com/tinygo-org/tinygo/issues/2842

### DIFF
--- a/interp/interpreter.go
+++ b/interp/interpreter.go
@@ -219,6 +219,9 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 				if err != nil {
 					return nil, mem, err
 				}
+			case callFn.name == "internal/task.Pause":
+				// Task scheduling isn't possible at compile time.
+				return nil, mem, r.errorAt(inst, errUnsupportedRuntimeInst)
 			case callFn.name == "runtime.nanotime" && r.pkgName == "time":
 				// The time package contains a call to runtime.nanotime.
 				// This appears to be to work around a limitation in Windows
--- a/interp/interpreter.go
+++ b/interp/interpreter.go
@@ -81,12 +81,28 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 		if inst.opcode != llvm.PHI {
 			for _, v := range inst.operands {
 				if v, ok := v.(localValue); ok {
-					if localVal := locals[fn.locals[v.value]]; localVal == nil {
+					index, ok := fn.locals[v.value]
+					if !ok {
+						// This is a localValue that is not local to the
+						// function. An example would be an inline assembly call
+						// operand.
+						isRuntimeInst = true
+						break
+					}
+					localVal := locals[index]
+					if localVal == nil {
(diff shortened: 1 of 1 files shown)


dankegel:
Locally, make test GOTESTFLAGS="-run TestBuild/EmulatedCortexM3 -v" on my 12GB linux box triggers the OOM killer after a few minutes with this patch, but passes quickly without it.

aykevl:
Hmm, maybe the fix triggers some sort of bug? I'll need to investigate what's going on.

aykevl:
This... is interesting. Roughly what I expected, actually (I expected some sort of assembly or synchronization primitive to go wrong).

```
/home/ayke/src/github.com/tinygo-org/tinygo/src/runtime/runtime_cortexm_qemu.go:70:10: too many runtime instructions
  call void asm sideeffect "wfi", ""(), !dbg !596

traceback:
/home/ayke/src/github.com/tinygo-org/tinygo/src/runtime/runtime_cortexm_qemu.go:70:10:
  call void asm sideeffect "wfi", ""(), !dbg !596
/home/ayke/src/github.com/tinygo-org/tinygo/src/runtime/runtime_cortexm_qemu.go:57:6:
  call void @runtime.exit(i32 1, i8* undef), !dbg !588
/home/ayke/src/github.com/tinygo-org/tinygo/src/runtime/panic.go:20:7:
  call void @runtime.abort(i8* undef), !dbg !596
/home/ayke/src/github.com/tinygo-org/tinygo/src/runtime/panic.go:32:14:
  call void @runtime.runtimePanic(i8* getelementptr inbounds ([23 x i8], [23 x i8]* @"runtime$string.69", i32 0, i32 0), i32 23, i8* undef), !dbg !589
/home/ayke/src/github.com/tinygo-org/tinygo/src/internal/task/task_stack.go:43:18:
  call void @runtime.nilPanic(i8* undef), !dbg !591
/home/ayke/src/github.com/tinygo-org/tinygo/src/runtime/scheduler_any.go:16:12:
  call void @"internal/task.Pause"(i8* undef), !dbg !600
/home/ayke/src/github.com/tinygo-org/tinygo/testdata:
  call void @time.Sleep(i64 1000000, i8* undef), !dbg !589
```

~I guess I'll just mark inline assembly like the `br` instruction with unknown operands.~

aykevl:
I've force-pushed another commit. It should be fixed now.

I also did a size check to be sure but not much changed. The only change was that testdata/goroutine.go got a little bit smaller in code size.

deadprogram (approved):
Looks good to me! Thanks @aykevl

Issue #2856, opened 2022-05-19
fatal: undefined symbol: syscall.Getpagesize


Theodoree:
env:

```
os:        debian 9
go:        1.18.1
tinygo:    0.23
llvm:      14.0.0

GOOS="linux"
GOARCH="amd64"
GOROOT="/usr/local/go"
GOPATH="/go"
GOCACHE="/root/.cache/tinygo"
CGO_ENABLED="1"
TINYGOROOT="/usr/local/lib/tinygo"
```

code:
```
package main
import (
     "fmt"
     "syscall"
)
func main() {
     fmt.Println(syscall.Getpagesize())
}
```

execute a command: `GOOS=linux GOARCH=amd64 tinygo build`
```
tinygo:ld.lld: error: undefined symbol: syscall.Getpagesize
>>> referenced by m.go:9 (/opt/repos/pkg/m.go:9)
>>>               /root/.cache/tinygo/thinlto/llvmcache-1F80CCA51D93267E9887A910A33CFCDC9CFE6A98:(runtime.run$1$gowrapper)
failed to run tool: ld.lld
error: failed to link /tmp/tinygo811094212/main: exit status 1
```
---
but darwin can be compiled normally, l don't  know way.

dankegel:
https://github.com/golang/go/blob/128279e5034ca29bad4320eef81a8abd5b40ea7e/src/syscall/syscall.go#L99 suggests it needs to be added to tinygo's runtime package on Linux.

dkegel-fastly:
Does https://github.com/tinygo-org/tinygo/pull/2899 work for you?

dkegel-fastly:
A fix has landed in the dev branch.

What app were you trying to build?

PR #2868, opened 2022-05-23
runtime: linux: implement syscall.Getpagesize like upstream

dkegel-fastly:
Should fix https://github.com/tinygo-org/tinygo/issues/2856, once the code is actually called.

Filed the PR to ask about the linking problem.

### DIFF
--- a/src/runtime/os_linux.go
+++ b/src/runtime/os_linux.go
@@ -1,13 +1,7 @@
-// Copyright 2009 The Go Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style
-// license that can be found in the LICENSE file.
-
 //go:build linux
 // +build linux
 
 package runtime
-
-import "unsafe"
 
 const GOOS = "linux"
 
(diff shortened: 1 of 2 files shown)


aykevl:
I think directly calling `getpagesize` would be a cleaner solution. You can add it to the musl build with this patch:

```diff
diff --git a/builder/musl.go b/builder/musl.go
index fa29496a..5dae0428 100644
--- a/builder/musl.go
+++ b/builder/musl.go
@@ -115,6 +115,7 @@ var Musl = Library{
                        "internal/libc.c",
                        "internal/syscall_ret.c",
                        "internal/vdso.c",
+                       "legacy/*.c",
                        "malloc/*.c",
                        "mman/*.c",
                        "signal/*.c",
```

dkegel-fastly:
Ok, I'll give that a shot!

aykevl:
And regarding the issue with os_linux.go: I would suggest splitting the file in two. One is `os_linux.go` but only for actual Linux (no baremetal or nintendoswitch), and one is `os_fakelinux.go` (name to be decided) which is only for the not-actually-Linux systems (baremetal, nintendoswitch, wasi) that just need to match an existing operating system for Go standard library compatibility.

dkegel-fastly:
Good. I was using os_real_linux.go, but I prefer your idea.

dkegel-fastly:
Closing in favor of https://github.com/tinygo-org/tinygo/pull/2899

PR #2872, opened 2022-05-26
Align API of  PortMaskSet and PortMaskClear to the other MCU implementations 

oflebbe:
Because of https://github.com/tinygo-org/drivers/pull/407 . Align the API of PortMaskSet and PortMaskClear. Not too sure if this a good idea since data types are coming straight from the SVD files.

### DIFF
--- a/src/machine/machine_rp2040_gpio.go
+++ b/src/machine/machine_rp2040_gpio.go
@@ -88,8 +88,8 @@ const (
 	PinSPI
 )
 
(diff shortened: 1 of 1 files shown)


aykevl (commented):
Can you run `go fmt` on the file? According to CI, it's not correctly formatted.
Other than that, it looks good to me. It makes the API consistent with other chips.

deadprogram:
Thank you very much @oflebbe now merging.

PR #2873, opened 2022-05-28
Add support for the embed package

aykevl:
Like the title says.
I needed to do some refactoring to cleanly support this feature, which I've done in separate commits for easier review.


builder/build.go
```diff
@@ -909,6 +961,110 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 	})
 }
 
+// createEmbedObjectFile creates a new object file with the given contents, for
+// the embed package.
+func createEmbedObjectFile(data, hexSum, sourceFile, sourceDir, tmpdir string, compilerConfig *compiler.Config) (string, error) {
+	// TODO: this works for small files, but can be a problem for larger files.
```
deadprogram, line 967: What exactly do you mean "small" here? Please provide specific number.

deadprogram:
Not sure I realized before this PR how much there is to the Go `embed` implementation. Still reading...

builder/build.go
aykevl: I don't have a specific number and I didn't measure it. But I imagine it might start to slow down compilation when reaching in the megabytes and cause real problems in the gigabytes.

deadprogram:
Anyone have any feedback to chime in here?

aykevl:
Pushed a small change so that `tinygo run -target=simavr ./testdata/embed` runs without a stack overflow.

> Not sure I realized before this PR how much there is to the Go `embed` implementation.

I expected it to be more difficult. That's why I avoided it. (The "this works for small files, but can be a problem for larger files" comment is the thing that made it simpler - for now at least).

sago35:
Variables using embed seem to consume only ROM if not modified.
This is very good.

aykevl:
> Variables using embed seem to consume only ROM if not modified. This is very good.

Yes, except on AVR.
To be more precise, on non-AVR:
 - `embed.FS` and `string` is always stored in ROM
 - `[]byte` is only stored in ROM if the compiler can prove it isn't modified, otherwise it's loaded in RAM

So if you want to make sure a file is always stored in ROM, avoid `[]byte`.
The exception is AVR, due to architecture limitations. This might be fixed some day, but should happen in a separate change. (Strings on AVR are also stored in RAM because of the same limitations).

deadprogram:
Thank you for all the work on this @aykevl and thanks @sago35 for testing/review. Now merging.

Issue #2870, opened 2022-05-24
Compile packages independently, link using LTO


aykevl:
After #285, I'd like to move one step further: by compiling packages entirely separately and doing optimizations across packages using ThinLTO (or, optionally, full LTO if desired). The main benefit is that compilation should be a _lot_ faster. Both with a cold cache (by parallelizing codegen) and with small changes to the source code (by reusing most packages). We should be able to get close to the speed of the `go` toolchain: TinyGo is currently a lot slower.

How we currently compile packages is as follows:

 1. LLVM IR for packages is generated in parallel (and cached in ~/.cache/tinygo).
 2. This IR is then merged together to create one huge LLVM module with the IR of all packages.
 3. Some generic LLVM optimizations and TinyGo specific transformation passes are applied to all this combined IR.
 4. The IR is then written to a temporary location, either as bitcode (for ThinLTO) or as an object file (for non-ThinLTO builds).
 5. The linker (usually lld) is invoked to link everything together to generate an executable. In the ThinLTO case, lld creates an object file internally and caches it.

What I'd like to see:

 1. LLVM IR for packages is generated in parallel (and cached in ~/.cache/tinygo), as before. TinyGo specific optimizations need to be done in this phase.
 2. The linker (lld) is then used to link all bitcode files together, using ThinLTO.

This means there is no phase in which all IR is combined into one big module, which avoids the serial step that currently takes up most of the compile time.

This is no small task. We currently rely heavily on merging all packages together to perform some (required) optimization passes. These will need to be changed in some way to work well with LTO, by modifying them or replacing them with something else:

 - [x] We don't support ThinLTO yet for some targets (see #2867, #2865 for example).
 - [x] Some targets need the `AddGlobalsBitmap` pass to be able to scan global variables in the GC mark phase. It should be possible to convert this to simply scanning the `.data`/`.bss` sections everywhere (see #2867, #2869 for example).
 - [ ] WebAssembly uses the `MakeGCStackSlots` pass. We need to make this pass run per package. In the future, the [WebAssembly GC](https://github.com/WebAssembly/gc) would be an alternative.
 - [x] Reflect information is currently processed for the whole program in `LowerReflect`. I've been working on a replacement in #2640 but it's going to cost something. In return, the compiler itself becomes easier to understand and new reflect features are easier to add.
[...]

PR #2867, opened 2022-05-22
Use ThinLTO on Windows

aykevl:
Switch Windows builds to use ThinLTO. This gets us closer to using ThinLTO everywhere.

Also change the way globals are scanned during GC, to avoid a bug with precise globals and ThinLTO.

This is WIP until #2865 is merged.

PR #2875, opened 2022-05-28
nintendoswitch: scan globals conservatively

aykevl:
This is a step towards #2870, similar to #2867 and #2869. After this, I think only Linux remains.

I did not test this patch, I don't know how to do it. I know there is `yuzu`, but I don't know how to get it to show `println` messages.
@racerxdl can you check that this change didn't break anything?


racerxdl:
Let me check :D

racerxdl:
@aykevl  LGTM!

image

Btw I couldnt get Yuzu to print it as well, but Ryujinx prints the svcCalls in console (the printscreen above is from ryujinx).
https://github.com/Ryujinx/Ryujinx

I think Ryujinx has a headless mode which we can use for automated tests since for homebrews we dont need any nintendo switch keys or related stuff. I will check that later ^^

deadprogram:
Thanks @aykevl and @racerxdl now merging.

PR #2878, opened 2022-05-30
corpus: make non-working packages easy to uncomment

aykevl:
This makes it easier to test for changes, just remove the hash sign in front of some packages to test them.
The contents of the corpus hasn't changed.


deadprogram:
Good idea! Merging.

PR #2879, opened 2022-05-30
gc: drop support for 'precise' globals

aykevl:
Precise globals require a whole program optimization pass that is hard to support when building packages separately. This patch removes support for these globals by converting the last use (Linux) to use linker-defined symbols instead.

For details, see: https://github.com/tinygo-org/tinygo/issues/2870


niaow (approved):
LGTM

PR #2880, opened 2022-05-30
chore: improves error messages

jcchavezs:
While trying to debug https://github.com/corazawaf/coraza/runs/6651956715?check_suite_focus=true#step:6:7 I stumbled uppon different errors which were lacking context and did not help at all.

I am up to introducing better errors everywhere but I would start introducing these error messages to start the discussion.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -290,7 +290,7 @@ func (p *Program) Parse() error {
 	for _, pkg := range p.sorted {
 		err := pkg.Parse()
 		if err != nil {
-			return err
+			return fmt.Errorf("when parsing the package %q: %w", pkg.ImportPath, err)
 		}
 	}
 
@@ -298,7 +298,7 @@ func (p *Program) Parse() error {
 	for _, pkg := range p.sorted {
 		err := pkg.Check()
 		if err != nil {
-			return err
+			return fmt.Errorf("when checking the package %q: %w", pkg.ImportPath, err)
 		}
 	}
 


aykevl:
This actually breaks proper error reporting. For example, if I introduce a syntax error in testdata/alias.go, I get the following:

```
$ tinygo run ./testdata/alias.go
# command-line-arguments
testdata/alias.go:27:1: expected declaration, found ')'
testdata/alias.go:35:2: expected declaration, found println
```

With this patch, I get this error message that is not as useful:

```
$ tinygo run ./testdata/alias.go
error: when parsing the package "command-line-arguments": could not compile: /home/ayke/src/github.com/tinygo-org/tinygo/testdata/alias.go:27:1: expected declaration, found ')' (and 1 more errors)
```

aykevl (changes requested):
I do believe that we can improve error messages in various parts of the compiler, but I don't think this particular patch is an improvement.
As for the issue in https://github.com/corazawaf/coraza/runs/6651956715, I can't really tell from the error message what's going wrong. But my first suspicion would be to look at this function:

https://github.com/tinygo-org/tinygo/blob/cf0b7edc78e42038a0bb522b3f1a5b76928e730e/loader/loader.go#L443-L457

jcchavezs:
@aykevl thanks for the answer. I have been trying to debug the problem and indeed `imported.Pkg` is `nil`, the error comes from https://github.com/tinygo-org/tinygo/blob/release/loader/loader.go#L299 specifically in https://github.com/tinygo-org/tinygo/blob/release/loader/loader.go#L377.

I tried to debug this but I did not find a way to build a unit test nor to fix it in local. 

The steps to reproduce it is to clone https://github.com/corazawaf/coraza/pull/254 and run `tinygo test ./...`. As you can see in the code, the packages are there.

aykevl:
Without looking, could it simply be the `...` syntax? I've not used it before in TinyGo. Listing each package separately on the command line should work however.

PR #2881, opened 2022-05-30
Start moving optimizations to the package build

aykevl:
This is for https://github.com/tinygo-org/tinygo/issues/2870.

I'd like to move optimization passes to the package build whenever possible, so that #2870 becomes easier. It should also improve compile time a little bit when the package is already cached although this is probably negligible with the current PR.

This is just a first step to get things going. More changes can be done in separate PRs.


deadprogram (approved):
Looking good, thanks @aykevl

PR #1829, opened 2021-04-21
Implement common USB 2.0 package

ardnew:
This PR serves as working area and development notes/discussion for a common USB 2.0 API and package implementation reusable by various TinyGo targets.

A functioning model — including a Device Controller driver with CDC-ACM device class support — is provided for Teensy 4 (NXP iMXRT1062). This can be used to iterate and evaluate changes to the architecture.

1. The first step would be to document at a high-level the design of this functional prototype. From this, I think more developers could then be involved in discussing (and performing) what changes should be made, and it should help navigate the complexities of the implementation. I'll take this task.
2. The next step should probably add support for another target. This process would help reveal which components of the existing architecture can be shared, and which components must be implemented on a per-target basis (e.g., USB PHY interaction, descriptor storage layout, etc.). I know TinyGo currently has USB (CDC-ACM) device support for SAMD21, SAMD51, and nRF52840, so one of these would be a good candidate to start with.

This is more than enough work for a single PR. But I feel it's all necessary to develop a robust baseline API for future USB targets.

PR #2889, opened 2022-06-02
USBHID (keyboard / mouse) support

sago35:
This PR adds USBHID (keyboard / mouse) support.
The following microcontrollers are targeted

* atsamd51
* atsamd21
* nrf52840 ~~(WIP: nrf52840's USBCDC / USBHID is broken)~~ (It works fine now)

Unlike #1829, it extends the existing USB implementation.
Also, the keyboard implementation is made by copying the @ardnew implementation.
The USBHID Support being added in this PR is to buy time until #1829 is completed.

The remaining tasks are as follows

* [x] Add smoketest
* [x] Make it work with atsamd51
* [x] Make it work with atsamd21
* [x] Make it work with nrf52840


sago35:
The commit message is probably correct as follows

```
samd21,samd51,nrf52840: add support for USBHID (keyboard / mouse)
```

sago35:
Now ready for review.

A summary of the changes in each file is as follows

* M       Makefile
    * Add smoketest
* A       src/examples/hid-keyboard/main.go
    * Add example
* A       src/examples/hid-mouse/main.go
    * Add example
* M       src/machine/machine_atsamd21.go
    * Add callback invocation when using HID
    * Add branch to support HID
    * Add SendUSBHIDPacket() for interrupt transfer
* M       src/machine/machine_atsamd51.go
    * Same as samd21
* M       src/machine/machine_nrf52840_usb.go
    * Same as samd21
* M       src/machine/usb.go
    * The processing for the descriptor was removed and it is now just a string of bytes
        * VID and PID are now initialized with values such as machine.VID
    * USB device settings can be switched by rewriting usbDescriptor
        * If EnableHID() is called, the descriptor is changed
    * Support for some enumeration instructions required for USBHID support
* A       src/machine/usb_descriptor.go
    * descriptor is defined
        * Initialized with the same contents as arduino
* A       src/machine/usb/hid/buffer.go
    * ringbuffer for interrupt transfer
    * It is based on machine/buffer.go
    * Size could be optimized a bit more
* A       src/machine/usb/hid/hid.go
* A       src/machine/usb/hid/keyboard.go
    * Extracted source code from ardnew
    * Source code for Keyboard operations.
        * Currently, mediakey etc. are not supported.
* A       src/machine/usb/hid/keycode.go
    * Extracted source code from ardnew
* A       src/machine/usb/hid/mouse.go
    * Source code for mouse operations.

deadprogram:
@sago35 this PR is really quite useful to me, because it helps me to actually study the HID API a little bit. I will provide feedback later on about how I think we should modify it, but consider how the API would need to handle the following 3 use cases:

- HID gamepad/joystick
- MIDI
- MSD

I think the current implementation makes some assumptions about keyboard/mouse that might not be sufficient.

What do you think?

sago35:
@deadprogram Thanks for the comment.
gamepad/joystick, MIDI and MSD should be considered separately from this PR.
My thoughts as of now are as follows.

### src/machine/usb.go

I have simplified USB Enumeration in this PR.
With this change, Descriptors for devices such as CDC + MIDI can be easily created.

To support MIDI and MSD, it should be necessary to add callbacks to BulkIn and BulkOut.

### src/machine/usb/hid/

There may be something missing in the keyboard or mouse functionality, but I don't understand what it is.

Gamepad and Joystick support should be implemented here.
Of course, `src/machine/usb_descriptor.go` must also be changed.

Based on my optimistic view, I believe that all implementations can be done with only the following Interface.
Ah, but it does not handle the data received from InterruptOut, so that will need to be added in the future.

https://github.com/tinygo-org/tinygo/blob/a1e62773a18a7713d17ad86f9658455ff62f221c/src/machine/usb/hid/hid.go#L34-L46

### src/machine/usb/midi

I have not been able to confirm much yet, but after the enumeration is done, BulkIn and BulkOut need to be implemented.
I don't know how much functionality midi has, but if it is just for serial communication, it is not too difficult.

### src/machine/usb/msd

I am not sure what I need for msd.
Perhaps BulkIn and BulkOut should be implemented.
And we will need to handle File and Dir according to the MSD way.
I think we should then implement the io/fs interface.

deadprogram:
@sago35 I was imagining something more like this: https://gist.github.com/deadprogram/4e55c3d0e663295839dd66f6517d8aa4

sago35:
@deadprogram 

I would like to complete USB enumeration as soon as possible, so I think it is better not to have the hid.Enable(kb) part.
(This is because if enumeration is not completed quickly, USBCDC will be unavailable for a longer period of time)
It is better to have the enumeration fixed when the import is done.

`keyboard.NewKeyboard()` might be good.
But perhaps `keyboard.New()` or `hid.NewKeyboard()` would be better.

```go
package main

import (
	"machine"
	"machine/usb/hid"
	"time"
)

func main() {
	button := machine.BUTTON
	button.Configure(machine.PinConfig{Mode: machine.PinInputPullup})

	kb := hid.NewKeyboard()
	mouse := hid.NewMouse()

	for {
		if !button.Get() {
			kb.Write([]byte("tinygo"))
			mouse.Move(0, 0)
			time.Sleep(200 * time.Millisecond)
		}
	}
}
```

deadprogram:
I also thought we should put the different devices into their own package folders, e.g.

```
"machine/usb/hid/keyboard"
"machine/usb/hid/mouse"
```

deadprogram:
Lastly, would be good to use an array of handlers. For example (not syntax checked or anything):

```go
type hidDevicer interface {
    callback()
}

var devices [5]hidDevicer
var size int

func SetCallbackHandler(d hidDevicer) {
    callbacks[size] = d
    size++
}

func PerformCallbacks() {
    for(i := 0; i < size; i++ {
        if done := devices[i].callback(); done {
		return
	}
    }
}
```

Then

sago35:
@deadprogram 
Package structure has been changed.
Certainly a good suggestion for the future.

src/machine/usb/hid/mouse/mouse.go
```diff
@@ -0,0 +1,90 @@
...
+
+func init() {
+	if Mouse == nil {
+		Mouse = newMouse()
+		hid.SetCallbackHandler(Mouse)
+	}
+}
+
+// NewMouse returns hid-mouse.
+func NewMouse() *mouse {
```
deadprogram: This can be just `New()` now.

src/machine/usb/hid/keyboard/keyboard.go
```diff
@@ -0,0 +1,482 @@
...
+
+func init() {
+	if Keyboard == nil {
+		Keyboard = newKeyboard()
+		hid.SetCallbackHandler(Keyboard)
+	}
+}
+
+// NewKeyboard returns hid-keybord.
+func NewKeyboard() *keyboard {
```
deadprogram: This can be just `New()` now.

src/machine/usb/hid/mouse/mouse.go
sago35: Yes, it is.
That is very good.

src/machine/usb/hid/keyboard/keyboard.go
sago35: Yes, it is.
That is very good.

deadprogram:
Tested on my Circuit Playground Express, and I had to change the example to this for it to work:

```go
package main

import (
	"machine"
	"machine/usb/hid/keyboard"
	"time"
)

func main() {
	button := machine.BUTTON
	button.Configure(machine.PinConfig{Mode: machine.PinInputPulldown})

	kb := keyboard.New()

	for {
		if button.Get() {
			kb.Write([]byte("tinygo"))
			time.Sleep(200 * time.Millisecond)
		}
	}
}
```

Otherwise it was amazing!

src/machine/usb/hid/keyboard/keyboard.go
```diff
@@ -0,0 +1,482 @@
...
+func (kb *keyboard) Callback() bool {
+	if b, ok := kb.buf.Get(); ok {
+		hid.SendUSBPacket(b)
+		return true
+	}
+	return false
+}
+
+func (kb *keyboard) ready() bool {
+	//return kb.dc != nil && kb.hc != nil
```
deadprogram: remove commented out line?

src/machine/usb/hid/keyboard/keyboard.go
```diff
@@ -0,0 +1,482 @@
...
+	b[3] = c.key()
+	b[4] = 0
+	b[5] = 0
+	b[6] = 0
+	b[7] = 0
+	b[8] = 0
+	if !kb.sendKey(false, b[:]) {
+		return hid.ErrHIDReportTransfer
+	}
+	//time.Sleep(10 * time.Millisecond)
```
deadprogram: remove commented line?

src/machine/usb/hid/keyboard/keyboard.go
```diff
@@ -0,0 +1,482 @@
...
+
+	b[1] = 0
+	b[3] = 0
+	if !kb.sendKey(false, b[:]) {
+		return hid.ErrHIDReportTransfer
+	}
+	//time.Sleep(10 * time.Millisecond)
```
deadprogram: Remove commented line?

src/machine/usb/hid/keyboard/keyboard.go
```diff
@@ -0,0 +1,482 @@
...
+//   0xC200 - 0xDFFF  UTF-8 packed        (U+0080 to U+07FF)  [USES LAYOUT]
+//   0xE000 - 0xE0FF  Modifier key        (bitmap, 8 keys, Shift/Ctrl/Alt/GUI)
+//   0xE200 - 0xE2FF  System key          (HID usage code, page 1)
+//   0xE400 - 0xE7FF  Media/Consumer key  (HID usage code, page 12)
+//   0xF000 - 0xFFFF  Normal key          (HID usage code, page 7)
+func (kb *keyboard) Press(c Keycode) error {
+	if err := kb.Down(c); nil != err {
+		return err
+	}
+	//time.Sleep(10 * time.Millisecond)
```
deadprogram: remove comment?

deadprogram:
I think if we correct these small items, and squash commits, we can merge this @sago35

sago35:
I made a squash commit.
Thank you for your review.

deadprogram:
Thank you for all the work on this PR @sago35 very exciting!

Now merging.

PR #2891, opened 2022-06-02
Attach USB DP to the correct pin on Matrix Portal M4

kjagiello:
USB DP wasn't connected to the right pin on Matrix Portal M4. This PR fixes it. I have verified the fix by running `examples/serial`.


deadprogram:
Thanks for the correction @kjagiello now merging.

PR #2893, opened 2022-06-03
transform (MakeGCStackSlots): do not move the stack chain pop earlier

### LINKED ISSUE #2876: Memory corruption in WebAssembly
While trying to enable ThinLTO on WebAssembly, I've found an issue with memory corruption. I managed to reproduce it with a small patch that _should_ be harmless but triggers this bug.

```diff
diff --git a/src/runtime/gc_conservative.go b/src/runtime/gc_conservative.go
index 00cadc30..f2a0b268 100644
--- a/src/runtime/gc_conservative.go
+++ b/src/runtime/gc_conservative.go
@@ -168,6 +168,11 @@ func (b gcBlock) markFree() {
        if gcAsserts && b.state() != blockStateFree {
                runtimePanic("gc: markFree() was not successful")
        }
+       addr := b.address()
+       *(*uintptr)(unsafe.Pointer(addr + 0)) = 0
+       *(*uintptr)(unsafe.Pointer(addr + 4)) = 0
+       *(*uintptr)(unsafe.Pointer(addr + 8)) = 0
+       *(*uintptr)(unsafe.Pointer(addr + 12)) = 0
 }
 
 // unmark changes the state of the block from mark to head. It must be marked
@@ -264,6 +269,7 @@ func alloc(size uintptr, layout unsafe.Pointer) unsafe.Pointer {
        if size == 0 {
                return unsafe.Pointer(&zeroSizedAlloc)
        }
+       GC()
 
        neededBlocks := (size + (bytesPerBlock - 1)) / bytesPerBlock
```

What it does, is:

1. Zero memory freed by the GC (it's free memory, so should be safe).
2. Run the GC before each allocation.

And then I run any program like this:

```
$ tinygo run -target=wasi -opt=1 ./testdata/alias.go
Error: failed to run main module `/tmp/tinygo2191174038/main`

Caused by:
    1: wasm trap: uninitialized element
       wasm backtrace:
           0: 0x2701 - <unknown>!tinygo_launch
           1:  0x537 - (*internal/task.Task).Resume
                           at /home/ayke/src/github.com/tinygo-org/tinygo/src/internal/task/task_asyncify.go:110:17
           2: 0x23a4 - runtime.scheduler
                           at /home/ayke/src/github.com/tinygo-org/tinygo/src/runtime/scheduler.go:167:11
           3: 0x22fc - runtime.run
                           at /home/ayke/src/github.com/tinygo-org/tinygo/src/runtime/scheduler_any.go:28:11
           4: 0x2283 - _start
                           at /home/ayke/src/github.com/tinygo-org/tinygo/src/runtime/runtime_wasm_wasi.go:21:5
```

(The `-opt=1` is important, otherwise it won't reproduce).

Apparently `t.state.entry` is zero, which it shouldn't be.

niaow:
Previously, the MakeGCStackSlots pass would attempt to pop the stack chain before a tail call. This resulted in use-after-free bugs when the tail call allocated memory and used a value allocated by its caller. Instead of trying to move the stack chain pop, remove the tail flag from the call.

Fixes #2876

### DIFF
--- a/transform/gc.go
+++ b/transform/gc.go
@@ -271,37 +271,17 @@ func MakeGCStackSlots(mod llvm.Module) bool {
 		// Make sure this stack object is popped from the linked list of stack
 		// objects at return.
 		for _, ret := range returns {
-			inst := ret
-			// Try to do the popping of the stack object earlier, by inserting
-			// it not right before the return instruction but moving the insert
-			// position up.
-			// This is necessary so that the GC stack slot pass doesn't
-			// interfere with tail calls (in particular, musttail calls).
-			for {
-				prevInst := llvm.PrevInstruction(inst)
-				if prevInst == parent {
-					break
-				}
-				if _, ok := pointerStores[prevInst]; ok {
-					// Pop the stack object after the last store instruction.
-					// This can probably be made more efficient: storing to the
-					// stack chain object and then immediately popping isn't
-					// useful.
-					break
-				}
-				if prevInst.IsNil() {
-					// Start of basic block. Pop the stack object here.
-					break
-				}
-				if !prevInst.IsAPHINode().IsNil() {
-					// Do not insert before a PHI node. PHI nodes must be
-					// grouped at the beginning of a basic block before any
-					// other instruction.
-					break
-				}
-				inst = prevInst
+			// Check for any tail calls at this return.
+			prev := llvm.PrevInstruction(ret)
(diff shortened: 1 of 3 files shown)


aykevl (approved):
Thank you! I still see some issues when I enable ThinLTO but there are fewer of them, so this is most certainly an improvement.

PR #2896, opened 2022-06-07
wioterminal: fix I2C definition

sago35:
This PR corrects the name for I2C in WioTerminal.

I2C0 in the schematic was named I2C1 in TinyGo.
Also, I2C1 in the schematic was named I2C0 in TinyGo.

image

I think this PR is doing the right thing, but compatibility is broken.
However, I believe it is better to merge in order to reduce future problems.


sago35:
image

deadprogram:
Agreed @sago35 now merging, thank you for the fix!

PR #2899, opened 2022-06-11
Getpagesize(): add test, implement for Linux & Windows

dkegel-fastly:
It was already implemented for Darwin.

Also add stubs for wasi, nintendo, etc.

Should fix https://github.com/tinygo-org/tinygo/issues/2856

### DIFF
--- a/src/syscall/syscall_libc_wasi.go
+++ b/src/syscall/syscall_libc_wasi.go
@@ -296,6 +296,11 @@ func Pipe2(p []int, flags int) (err error) {
 	return ENOSYS // TODO
 }
 
+func Getpagesize() int {
+	// per upstream
+	return 65536
+}
+
 // int stat(const char *path, struct stat * buf);
 //export stat
 func libc_stat(pathname *byte, ptr unsafe.Pointer) int32
--- a/src/syscall/syscall_nonhosted.go
+++ b/src/syscall/syscall_nonhosted.go
@@ -205,3 +205,9 @@ type Timeval struct {
 	Sec  int64
 	Usec int64
 }
+
+func Getpagesize() int {
+	// There is no right value to return here, but 4096 is a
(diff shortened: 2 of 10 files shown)


deadprogram (approved):

src/syscall/syscall_libc_wasi.go
```diff
@@ -296,6 +296,10 @@ func Pipe2(p []int, flags int) (err error) {
 	return ENOSYS // TODO
 }
 
+func Getpagesize() int {
+	return 4096 // TODO
```
aykevl: This is 64kB in Go, which also makes a lot more sense to me (it matches the wasm page size): https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/runtime/os_js.go;l=105

src/syscall/syscall_nonhosted.go
```diff
@@ -205,3 +205,7 @@ type Timeval struct {
 	Sec  int64
 	Usec int64
 }
+
+func Getpagesize() int {
+	return 4096 // TODO
```
aykevl: Hmm. I'd say this is not TODO, it's simply unavailable on most baremetal systems because there is no virtual memory and thus no pages. Returning something like 4 or 8 might be more appropriate but I think 4096 is fine for compatibility.
I think putting something like that in a comment would be better than simply TODO (because there is nothing to do here).

aykevl (commented):
Seems good to me, with some nits.

src/syscall/syscall_libc_wasi.go
dkegel-fastly: Done.

src/syscall/syscall_nonhosted.go
dkegel-fastly: Done.

PR #2902, opened 2022-06-11
ci: disable autocrlf on windows

sago35:
The source code included in TinyGo, which is created with GitHub Actions, is in CRLF.
I suspect this is due to git autocrlf, so I changed the setting.

https://github.com/tinygo-org/tinygo/releases/download/v0.23.0/tinygo0.23.0.windows-amd64.zip

Maybe this change will make it LF.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -22,6 +22,11 @@ jobs:
         shell: bash
         run: |
           scoop install ninja binaryen
+      - name: git config
+        shell: bash
+        run: |
+          git config --global core.autocrlf false
+          git config --global core.eol lf
       - name: Checkout
         uses: actions/checkout@v2
         with:


aykevl:
Why is this needed? CRLF is the standard newline format on Windows, which seems good to me.

sago35:
For example, executing `go fmt` will result in `LF`.

I believe some tools will not work outside of `LF`.
I don't know now, but at least in the past it didn't work well with `vim-go`.

Some tools attempt to output source code that is partially `LF`.
https://github.com/davidrjenni/reftools

The source code of Go is all `LF`, even the Windows version.

aykevl:
> I believe some tools will not work outside of `LF`.
> I don't know now, but at least in the past it didn't work well with `vim-go`.

That sounds like a bug in `vim-go`.

> The source code of Go is all `LF`, even the Windows version.

That's a good point.
However, I've also found that using CRLF on Windows uncovered several bugs. Bugs that would otherwise not be found in CI and only when testing locally with autocrlf at the default (enabled). Therefore, I would prefer to keep autocrlf enabled unless there is a good practical reason not to.

sago35:
Okay, this PR is now closed.

PR #2903, opened 2022-06-11
compiler: add support for type parameters (aka generics)

aykevl:
...that was surprisingly easy.

Doing it now because Go 1.19 requires generics support in the standard library.


deadprogram:
Of course it does not work with Go 1.16 https://app.circleci.com/pipelines/github/tinygo-org/tinygo/7520/workflows/f618ef84-d1b1-42a4-a2e8-a75e9feed751/jobs/37590

testdata/generics.go
```diff
@@ -0,0 +1,14 @@
...
+func main() {
+	println("add:", Add(3, 5))
+	println("add:", Add(int8(3), 5))
+}
+
+type Integer interface {
+	int | int8 | int16 | int32 | int64
+}
+
+func Add[T Integer](a, b T) T {
```
deadprogram, line 12: OK this is very very exciting.

aykevl:
Same for Go 1.17.
Generics can be super useful and are required for Go 1.19. So I'd propose removing support for Go 1.16 and Go 1.17.
It is perhaps possible to keep Go 1.6 and Go 1.17 supported with a bit of effort, but without generics of course. I'm not sure that's worth it TBH.

Any opinions on this? (Ping @QuLogic)

dkegel-fastly:
Dropping 1.17 could make things hard for developers working on two different projects, one of which requires go 1.17 be installed.

If it really isn't hard to support 1.17, I would like to see us do that.  We can drop the support after one more major release of go.

aykevl:
Ok, kept support for Go 1.17 (and probably Go 1.16 as well).

deadprogram:
> and probably Go 1.16 as well

Tests pass using 1.16 so yes, that appears to be the case.

deadprogram:
A very fine and reasonable patch to cover all the current cases, IMO. Great work @aykevl

soypat:
Jeepers, christmas came early

@aykevl Edit to add meaningful value to this comment: I'd really like type parameter support. I have quite a few uses for it for personal projects and might have a real use for them at work with a generic FSM I built recently.

dgryski:
Test failure running the corpus with new generics code:
```
~/go/src/github.com/tinygo-org/tinygo $ go test -test.v -run=Corpus/github.com/google/btree  -corpus=testdata/corpus.yaml
=== RUN   TestCorpus
=== PAUSE TestCorpus
=== CONT  TestCorpus
=== RUN   TestCorpus/github.com/google/btree
=== PAUSE TestCorpus/github.com/google/btree
=== CONT  TestCorpus/github.com/google/btree
    main_test.go:518: go: creating new go.mod: module github.com/tinygo/tinygo-corpus-test
    main_test.go:518: go: downloading github.com/google/btree v1.1.2
    main_test.go:518: go: added github.com/google/btree v1.1.2
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc03e5804f8 stack=[0xc03e580000, 0xc05e580000]
fatal error: stack overflow
```

And the btree code with generics: https://github.com/google/btree/blob/master/btree_generic.go

dkegel-fastly:
Filed https://github.com/tinygo-org/tinygo/issues/2957 for the overflow with google/btree.

PR #2904, opened 2022-06-11
all: drop support for Go 1.16 and Go 1.17

aykevl:
Draft to await result of the discussion in https://github.com/tinygo-org/tinygo/pull/2903#issuecomment-1152944907.


dgryski:
While we're here we should probably update the calls to `ioutil/*` to the appropriate `io` or `os` methods.

QuLogic:
Since this seems a more relevant place to follow up than https://github.com/tinygo-org/tinygo/pull/2903#issuecomment-1152944907, I'll post here. We have Go 1.16 in Fedora 35, Go 1.18 in Fedora 36, and Go 1.19 in Fedora Rawhide (to be 37). Fedora 35 goes EOL one month after Fedora 37 is out (so approximately November).

However, since #3039 is now in, dropping support for LLVM 13 _already_ means that `dev` would no longer work on Fedora 35. So there's no reason not to apply this one as well, from a Fedora point of view.

QuLogic:
Oh, actually, since TinyGo 0.24.0 requires golang.org/x/tools 0.1.11, which itself requires Go 1.17, I think Fedora 35 is already unsupported, so its constraints are no longer an issue.

aykevl:
@QuLogic thanks!

I have now rebased and updated it after multiple changes in the dev branch (mostly generics and timers). Right now I have a local branch that adds support for Go assembly files on Linux (32-bit and 64-bit) but it uses generics - because I didn't want to copy the ELF writing code between ELF32 and ELF64. Therefore, this branch requires Go 1.18.

aykevl:
@deadprogram looks like the TinyHCI runner might need a Go update.

deadprogram:
@aykevl fixed TinyHCI and now all passing.

deadprogram:
@aykevl thanks for all the work on this, and thanks @QuLogic for the detailed review.

Now merging! :scissors: :scissors:

deadprogram:
Something not working as expected after merge into `dev`

CI failing on macOS
https://github.com/tinygo-org/tinygo/runs/8090351014?check_suite_focus=true#step:12:19781

aykevl:
Yes, the timers test is flaky on MacOS.

PR #2906, opened 2022-06-11
all: Go 1.19 support

aykevl:
This adds early Go 1.19 support. There are a number of things that don't work yet, but the smoke tests pass so it's at least working for a significant subset of programs.

This change also switches from CircleCI convenience images to upstream Go images. This makes it a bit easier to use the latest Go versions. Also, the convenience images are not updated anymore.


deadprogram:
:one: :black_circle: :one: :nine:

PR #2908, opened 2022-06-14
fixes tinygo test run

jcchavezs:
This PR aims to fix the run of `tinygo test ./...` for packages like `github.com/corazawaf/coraza` where root package import subpackages.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -19,6 +19,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"sort"
 	"strconv"
 	"strings"
 	"unicode"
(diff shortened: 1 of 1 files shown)


dkegel-fastly:
Could you update the smoke test to minimally hit that case, if it's not too hard?

deadprogram:
@jcchavezs @dkegel-fastly not sure, but wasn't this something that was added in a different way subsequently? In other words, should it be closed now?

dkegel-fastly:
I suspect so.  A run of 'go test ./...' in coraza with tinygo 0.33.0 mostly succeeds, and I suspect will do even better on 0.34.0 (since the only error I see are three panic: unimplemented: internal/abi.FuncPCABI0's, which 0.34 may have improved).

Closing for now, reopen if needed.

PR #2909, opened 2022-06-15
runtime: include .data section in globals scan

aykevl:
All the globals are between _etext and _end. We were scanning only between _edata and _end, which mainly consists of the .bss section. Scanning from _etext ensures that the .data section is also included.

This bug didn't result in issues in CI, but did result in a bug in the recover branch: https://github.com/tinygo-org/tinygo/pull/2331. This patch fixes this bug.

### DIFF
--- a/src/runtime/os_linux.go
+++ b/src/runtime/os_linux.go
@@ -24,21 +24,94 @@ const (
 	clock_MONOTONIC_RAW = 4
 )
 
-//go:extern _edata
-var globalsStartSymbol [0]byte
+// For the definition of the various header structs, see:
+// https://refspecs.linuxfoundation.org/elf/elf.pdf
+// Also useful:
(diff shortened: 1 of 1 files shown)


deadprogram:
https://github.com/tinygo-org/tinygo/runs/6905013421?check_suite_focus=true#step:17:1334 is still failing

aykevl:
I hope I've really fixed it now. The new patch makes `scanGlobals` more like Windows and MacOS.

aykevl:
Looks like this patch gets #2331 to work! Finally!

Instead of merging this PR, it may be easier to merge #2331, which also contains the commit in this PR (otherwise #2331 would need to be rebased after this PR got merged).

deadprogram:
Closing this PR as advised by @aykevl since this commit was merged into `dev` with PR https://github.com/tinygo-org/tinygo/pull/2331

PR #2911, opened 2022-06-16
wasm: Use newer WebAssembly features

aykevl:
This PR will start to use a few more WebAssembly features, such as bulk memory operations. This results in a significant code size saving. How much it saves varies a lot but it's typically around 1300 bytes per binary.

This change is possible by bumping our minimum Node.js version to 14. The previous LTS version (12) has been marked end of life, so we can start to depend on features in the current oldest LTS version, which is version 14. Browsers have been supporting these features for a long time now, it's just Node.js that prevented us from doing this before.

CC @dgryski @matthewcp @natemoo-re @justinclift for opinions

EDIT: I also looked at the multivalue extension but using that appears to be a bit more difficult. It might be useful to look at next.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,10 +86,10 @@ commands:
       - run: go install -tags=llvm<<parameters.llvm>> .
       - restore_cache:
           keys:
-            - wasi-libc-sysroot-systemclang-v5
+            - wasi-libc-sysroot-systemclang-v6
       - run: make wasi-libc
       - save_cache:
-          key: wasi-libc-sysroot-systemclang-v5
+          key: wasi-libc-sysroot-systemclang-v6
           paths:
             - lib/wasi-libc/sysroot
       - run: make gen-device -j4
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -67,7 +67,7 @@ jobs:
         uses: actions/cache@v2
         id: cache-wasi-libc
         with:
-          key: wasi-libc-sysroot-v3
+          key: wasi-libc-sysroot-v4
           path: lib/wasi-libc/sysroot
       - name: Build wasi-libc
         if: steps.cache-wasi-libc.outputs.cache-hit != 'true'
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -77,7 +77,7 @@ jobs:
         uses: actions/cache@v2
         id: cache-wasi-libc
         with:
-          key: wasi-libc-sysroot-linux-asserts-v4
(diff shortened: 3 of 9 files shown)


justinclift:
The concept sounds like a good idea, but I don't have time in the short to medium future to do any review. :innocent:

deadprogram:
I looked over the specific features this enables, and count me in favor.

deadprogram:
I cannot find a downside at present to enable this, so now merging.

deadprogram:
OK that did not merge as smoothly as expected.

@aykevl please see https://github.com/tinygo-org/tinygo/runs/6979373526?check_suite_focus=true#step:14:253

Seems like that is occuring on all OS builds that failed CI.

aykevl:
See: https://github.com/tinygo-org/tinygo/pull/2923

codefromthecrypt:
This is good progress. One thing to note is probably don't use anything by default not in the 2.0 draft (this doesn't) https://webassembly.github.io/spec/core/

Another is that multi-value (especially multiple return) is very popular, so hopefully that can end up working one day.

Thanks for the work on this!

codefromthecrypt:
another feature that would be incredibly helpful, and possibly less work than multi-value is reference types (externref)

https://webassembly.github.io/spec/core/syntax/types.html#syntax-reftype 

This allows users to pass host pointers opaquely, so that no one can do arithmetic on them, or serialize them in wasm. It might not change the wasm side terribly, but it would dramatically simplify the host side.

aykevl:
Yes, I'd like to add multivalue support. Unfortunately it triggers an LLVM bug: https://github.com/llvm/llvm-project/issues/55136#issuecomment-1185493030.

I'm not sure how reference types can be integrated into TinyGo. I suppose they could be used in the syscall/js package for `js.Value` but I haven't looked into it yet.

codefromthecrypt:
@aykevl thanks for replying and giving context.

wrt reference types, one simple use case is something like a security token or correlation context. The host passes it into wasm compiled by tinygo, and that wasm can make host callbacks with the same parameter. In any case, the wasm itself doesn't need to serialize it or write it to memory. Once something like that works, the rest should follow!

aykevl:
@codefromthecrypt yes, I see. My question is more how this should be represented in Go. Go doesn't have a concept of reference types, except maybe for `js.Value`. It also makes the assumption that all values can be stored in linear memory, which is not true for reference types. Therefore, I think it's best to wait until we support the WebAssembly GC and stack switching (hopefully, if possible!) which does support reference types.

codefromthecrypt:
gotcha. yeah on the host side we use uintptr for lack of a better fit. There's no coupling between js in the core spec, though true js binding is the only thing I've seen actually use externref.

PR #2912, opened 2022-06-16
atmega2560: multiple uarts

ysoldak:


src/machine/machine_atmega.go
```diff
@@ -143,38 +165,37 @@ func (uart *UART) Configure(config UARTConfig) {
...
-	avr.UCSR0C.Set(avr.UCSR0C_UCSZ01 | avr.UCSR0C_UCSZ00)
+	uart.statusRegC.Set(avr.UCSR0C_UCSZ01 | avr.UCSR0C_UCSZ00)
+}
+
+func (uart *UART) handleInterrupt(intr interrupt.Interrupt) {
+	// Read register to clear it.
+	data := uart.dataReg.Get()
+
+	// Ensure no error.
+	if !uart.statusRegA.HasBits(avr.UCSR0A_FE0 | avr.UCSR0A_DOR0 | avr.UCSR0A_UPE0) {
```
ysoldak, line 187: Can look confusing we use constants *associated* with UART0 here, but they are actually the same for all UARTs. Keeps the code size smaller and cleaner.
Same comment for several lines above.

deadprogram: Understood, that is not uncommon with the generated wrappers from the SVD files.

deadprogram:
@ysoldak thank you very much for another cool contribution!

Just a small comment. This PR probably could have been separated into 2 commits, one that added the feature, the other adding the new example.

In any case, not worth the extra effort. Now merging.

PR #2913, opened 2022-06-16
build: cancel in-progress builds

ysoldak:
This will cancel old, no-longer-actual in-progress builds when a new commit triggers new workflow build.
This shall help with resource utilisation.

See https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -7,6 +7,10 @@ on:
       - dev
       - release
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build-macos:
     name: build-macos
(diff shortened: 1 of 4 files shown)


aykevl (approved):
Thank you! The last time I looked into this I couldn't find an official way to do this.

deadprogram:
I think this change has the unintended side effect of not letting all of the Linux build to run. Some of these are missing from this PR:
image

ysoldak:
they are still building...

ysoldak:
Seems like `actions/cache@v2` does not play nicely with this configuration and have to re-build LLVM

deadprogram:
Thanks very much @ysoldak now merging.

PR #2915, opened 2022-06-17
avr: add support for `recover()`

aykevl:
You can see that it works with the following command:

    tinygo run -target=simavr ./testdata/recover.go

This also gets the following tests to pass again (which started failing since #2331):

    go test -run=Build -target=simavr -v

Adding support for AVR was a bit more compliated because it's also necessary to save and restore the Y register. That's why the first two commits move things around a bit.

### DIFF
--- a/compiler/defer.go
+++ b/compiler/defer.go
@@ -71,7 +71,7 @@ func (b *builder) deferInitFunc() {
 		// possibly broken by the C alloca function).
 		// The frame pointer is _not_ saved, because it is marked as clobbered
 		// in the setjmp-like inline assembly.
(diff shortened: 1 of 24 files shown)


aykevl:
Pushed a fix for the test failure.

deadprogram:
You are not kidding that it was a bit more complicated. Digesting the changes now. 

Hopefully after this the WebAssembly `recover()` is lot easier.

aykevl:
> Hopefully after this the WebAssembly `recover()` is lot easier.

Unfortunately not, I'm afraid. WebAssembly is a really weird architecture.

deadprogram:
Let's merge it! Thanks @aykevl

PR #2916, opened 2022-06-17
avr: fix race condition in stack write

aykevl:
If an interrupt happens between the writes to SPL and SPH, the stack
pointer is inconsistent and terrible things will happen. Therefore,
disable interrupts while updating the stack pointer.

Interrupts are restored _before_ the write to SPH. This is safe, because
interrupts are re-enabled with a one cycle delay. The avr-gcc and Clang
compilers do the same thing when they need to update the stack pointer.

It's almost impossible to test for this bug, but it should make firmware
just a little bit more reliable.


deadprogram:
:checkered_flag: merging.

PR #2918, opened 2022-06-18
all: run gofmt on all source code

aykevl:
Some source code wasn't part of `FMT_PATHS` so wasn't checked for correct formatting. This change includes all this source code and excludes cgo/testdata because it contains files that can't be parsed.

### DIFF
--- a/src/net/interface_tinygo.go
+++ b/src/net/interface_tinygo.go
@@ -1,3 +1,4 @@
+//go:build tinygo
 // +build tinygo
 
 package net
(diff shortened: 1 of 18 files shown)


src/net/interface_tinygo.go
```diff
@@ -1,3 +1,4 @@
+//go:build tinygo
 // +build tinygo
```
aykevl, line 2: Huh, this doesn't make much sense. This is part of the TinyGo source so this build tag is always set. :shrug:

deadprogram: Who knows? Seems benign anyhow.

PR #2919, opened 2022-06-18
runtime: use the main (startup) stack for the main goroutine

aykevl:
Instead of always starting a new goroutine for the main goroutine, run
the main goroutine on the system stack.
The system stack is not occupied with scheduling, instead each goroutine
that wants to pause itself calls into the scheduler which will switch to
the next task (goroutine) to run, or sleeps.

There are various advantages of this over the previous system:

  * When the program doesn't start a goroutine, the code size and RAM
    consumption is close to what you'd get with `-scheduler=none`.
  * When the program does start a goroutine, there is still a reduction
    in RAM consumption because only one extra stack is needed.
  * Because tasks directly switch to the next task to run, only a single
    task switch is needed instead of two (goroutine -> scheduler ->
    goroutine). This should improve task switching performance.

I kept the current behavior for WebAssembly/Asyncify. I looked into how
the same benefits can be realized for WebAssembly but couldn't easily
find how to do that. Maybe this can be done separately, or maybe we'll
just wait for the stack switching proposal to finish.

The code for Cortex-M is currently more complicated than I'd like, and
therefore can sometimes result in a slight increase in code size. I'd
like to fix this eventually but am still looking into good ways to do
this. I still think this change is generally beneficial because many
programs see big reductions in code size when compiling for Cortex-M.

---

This is a draft because I'd like to merge this after the coming release: this is a rather big and possibly risky change.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -1228,10 +1228,15 @@ func determineStackSizes(mod llvm.Module, executable string) ([]string, map[stri
 	// Goroutines need to be started and finished and take up some stack space
 	// that way. This can be measured by measuing the stack size of
 	// tinygo_startTask.
-	if numFuncs := len(functions["tinygo_startTask"]); numFuncs != 1 {
-		return nil, nil, fmt.Errorf("expected exactly one definition of tinygo_startTask, got %d", numFuncs)
+	var baseStackSize uint64
+	var baseStackSizeType stacksize.SizeType
+	var baseStackSizeFailedAt *stacksize.CallNode
+	if len(gowrappers) != 0 {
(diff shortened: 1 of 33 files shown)


niaow:
Hmm. For something like this to work on WASM, I think we could work with a more minimal loop on the outside (just passing the switch target through a global). We would however need some logic similar to Cortex-M to ensure that external calls use the main stack. I also think it may make sense to move the scheduler into `internal/task` instead of linknaming it, since these are getting tangled together now.

deadprogram:
Closing since it is so out of date. Thanks!

PR #2921, opened 2022-06-20
compiler: drop support for macos syscalls via inline assembly

aykevl:
This has always been unsupported on MacOS and has in fact been removed from upstream Go a few releases ago. So do the same for TinyGo.
Linux seems to be the only supported OS with a stable syscall interface.


dgryski (approved):
LGTM.  Is there anything currently using this code?

aykevl:
Not that I'm aware. Hence why I'm making this PR.

deadprogram:
:scissors: :wastebasket: merging.

PR #2923, opened 2022-06-21
compiler: update tests after adding new wasm features

aykevl:
I don't understand why this wasn't caught in CI. It should have. In any case, because the llvm-features string was updated, these IR outputs were updated.


dgryski (approved):
LGTM

deadprogram:

PR #2925, opened 2022-06-22
Add support for printing slices via print/println

### LINKED ISSUE #2922: print() complains "unknown arg type: []byte"
Trying to use a byte array throws a compilation error : 

unknown arg type: []byte

Doesnt matter if self written or in any predesigned struct like bytes.buffer{}

tinygo version 0.23.0 windows/amd64 (using go version go1.18.3 and LLVM version 14.0.0)
MCU used: RPI PICO

fgsch:
Fixes #2922.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -1541,6 +1541,12 @@ func (b *builder) createBuiltin(argTypes []types.Type, argValues []llvm.Value, c
 			case *types.Pointer:
 				ptrValue := b.CreatePtrToInt(value, b.uintptrType, "")
 				b.createRuntimeCall("printptr", []llvm.Value{ptrValue}, "")
+			case *types.Slice:
+				bufptr := b.CreateExtractValue(value, 0, "")
+				buflen := b.CreateExtractValue(value, 1, "")
+				bufcap := b.CreateExtractValue(value, 2, "")
+				ptrValue := b.CreatePtrToInt(bufptr, b.uintptrType, "")
+				b.createRuntimeCall("printslice", []llvm.Value{ptrValue, buflen, bufcap}, "")
 			default:
 				return llvm.Value{}, b.makeError(pos, "unknown arg type: "+typ.String())
 			}
--- a/src/runtime/print.go
+++ b/src/runtime/print.go
@@ -39,6 +39,17 @@ func printint8(n int8) {
 	}
 }
 
+func printuint(n uintptr) {
+	switch unsafe.Sizeof(n) {
+	case 2:
+		printuint16(uint16(n))
+	case 4:
+		printuint32(uint32(n))
+	case 8:
+		printuint64(uint64(n))
+	}
+}
+
 func printuint16(n uint16) {
 	printuint32(uint32(n))
 }
@@ -323,14 +334,7 @@ func printitf(msg interface{}) {
 		// cast to underlying type
 		itf := *(*_interface)(unsafe.Pointer(&msg))
 		putchar('(')
-		switch unsafe.Sizeof(itf.typecode) {
-		case 2:
-			printuint16(uint16(itf.typecode))
-		case 4:
-			printuint32(uint32(itf.typecode))
-		case 8:
-			printuint64(uint64(itf.typecode))
-		}
+		printuint(uintptr(itf.typecode))
 		putchar(':')
 		print(itf.value)
 		putchar(')')
@@ -371,4 +375,13 @@ func printbool(b bool) {
 	} else {
 		printstring("false")
 	}
+}
+
+func printslice(ptr, len_, cap_ uintptr) {
+	putchar('[')
+	printuint(len_)
+	putchar('/')
+	printuint(cap_)
+	putchar(']')
+	printptr(ptr)
 }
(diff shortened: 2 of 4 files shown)


compiler/compiler.go
```diff
@@ -1519,6 +1519,8 @@ func (b *builder) createBuiltin(argTypes []types.Type, argValues []llvm.Value, c
 			case *types.Pointer:
 				ptrValue := b.CreatePtrToInt(value, b.uintptrType, "")
 				b.createRuntimeCall("printptr", []llvm.Value{ptrValue}, "")
+			case *types.Slice:
+				b.createRuntimeCall("printslice", []llvm.Value{value}, "")
```
aykevl: You need to split up the slice in pointer/len/cap and cast the pointer to a consistent type (like unsafe.Pointer).

fgsch: I had that originally but this works the same. Is there any reason to use one over the other?

aykevl: Yes, try printing a slice that's not a byte slice. []int for example.

testdata/print.go
```diff
@@ -45,4 +45,7 @@ func main() {
 
 	// print bool
 	println(true, false)
+
+	// print slice
+	// println([]byte{'H', 'e', 'l', 'l', 'o', ' ', 'W', 'o', 'r', 'l', 'd'})
```
aykevl: I recommend printing `[]byte(nil)` and `[]int(nil)`. Not a full test but using a non-nil slice is not possible.

fgsch: Good idea. I will update it.

compiler/compiler.go
fgsch: After I updated the tests with your suggestion I noticed this 😅

fgsch: Done.

testdata/print.go
fgsch: Done.

src/runtime/print.go
```diff
@@ -372,3 +372,12 @@ func printbool(b bool) {
 		printstring("false")
 	}
 }
+
+func printslice(ptr uintptr, len_, cap_ int64) {
```
aykevl: `len` and `cap` are also of `uintptr` type. They just happen to match `int64` on a 64-bit system.

src/runtime/print.go
```diff
@@ -349,7 +349,7 @@ func printmap(m *hashmap) {
 
 func printptr(ptr uintptr) {
 	if ptr == 0 {
-		print("nil")
+		print("0x0")
```
aykevl: Any reason for this change? I know it didn't match upstream exactly, but `nil` seems fine to me.

fgsch: To match upstream. I will revert it.

src/runtime/print.go
fgsch: I think I can use `int` in this case too to avoid converting it back? That should work in both 32 and 64 bit systems.

aykevl: There are also 16-bit systems: AVR. While `len(slice)` returns an int on AVR to comply with the spec, it is internally stored as a pointer-sized integer (which is 16 bits on AVR) and thus matches `uintptr`.
You can add a new `printuintptr` like this:

```go
func printuintptr(n uintptr) {
    switch unsafe.Sizeof(n) {
    case 2:
        printuint16(uint16(n))
    // etc
    }
}
```

You can test whether it works there by running `tinygo run -target=simavr ./testdata/print.go`.

fgsch: Thanks again for the cluebat. I will update it.

fgsch: I used `print(uint(len_))` and `print(uint(cap_))` instead. Hopefully this is good.

dkegel-fastly:
Try this:
```
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -261,6 +261,7 @@ jobs:
         uses: actions/checkout@v2
       - name: Install apt dependencies
         run: |
+          sudo apt-get update
           sudo apt-get install --no-install-recommends \
               qemu-user \
               g++-arm-linux-gnueabihf \
```

fgsch:
@aykevl updated and tested in avr. Is there anything else I should address before this can be merged?

src/runtime/print.go
aykevl: This works, but it results in larger code than necessary on AVR. (Note that the switch on `unsafe.Sizeof` is resolved at compile time, not at runtime). So I'd really prefer a `printuintptr` that prints numbers based on the actual integer size.

Here is some code that also does this, and can be factored out:

https://github.com/tinygo-org/tinygo/blob/b65447c7d567eea495805656f45472cc3c483e03/src/runtime/print.go#L326-L333

fgsch: Ok, I will update it accordingly.

fgsch: Done

deadprogram:
Seems to me this is ready, once the commits are squashed, if you please. Or would you prefer I do that for you?

fgsch:
I will do it shortly. Thank you!

fgsch:
@deadprogram done.

deadprogram:
Thanks for implementing this @fgsch now merging.

PR #2928, opened 2022-06-23
darwin: print full size information for -size=full

aykevl:
The MachO file format is a bit weird and doesn't store the DWARF debug
information directly in the file. Instead, it has to be looked up in the
original object file. This makes reading the DWARF debug information for
code size usage a bit more difficult. However, it works with this
change.

```
$ GOOS=darwin tinygo build -o test -size=full ./testdata/alias.go
    111    4306      56       7 |    4473      63 | (unknown)
    347       0       0      16 |     347      16 | internal/task
     65       0       0       0 |      65       0 | main
   1822       0       8      82 |    1830      90 | runtime
   2345    4306      64     105 |    6715     169 | total
```

I noticed this wasn't finished while writing the changelog and got distracted. Note that it doesn't always work well with LLVM 13, use LLVM 14 instead. Or build the same thing again, the 2nd time it should pick up debug information.

### DIFF
--- a/builder/sizes.go
+++ b/builder/sizes.go
@@ -127,7 +127,7 @@ var (
 // readProgramSizeFromDWARF reads the source location for each line of code and
 // each variable in the program, as far as this is stored in the DWARF debug
 // information.
-func readProgramSizeFromDWARF(data *dwarf.Data, codeOffset uint64) ([]addressLine, error) {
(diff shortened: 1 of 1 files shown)


dgryski (approved):
LGTM.  Tested, seems to work.

deadprogram:
Thanks everyone! Now merging.

PR #2927, opened 2022-06-23
all: update to version 0.24.0

aykevl:

PR #2929, opened 2022-06-23
ci: fix CI failure due to missing apt-get update

aykevl:
There are failures in #2928 and #2927. Hopefully this fixes them.


deadprogram:
Not done building all the jobs, but at least one that failed previously is passing with this PR.

PR #2930, opened 2022-06-25
main: use tags parser from buildutil

aykevl:
This should add support for things like quotes around tags, if they are ever needed.

Only making this change now because I happened to stumble across buildutil.TagsFlag.

I would prefer to merge this after the coming release. Not because I think it's risky but because it doesn't fix a particular bug and I'd like to avoid introducing a new bug.


dgryski (approved):
Seems reasonable.  Since this passes CI (and our CI tests uses tags) this is probably fine to merge.

PR #2932, opened 2022-06-26
all: use LLVM 14 by default

aykevl:
This also adds support for LLVM 14 from Homebrew on MacOS.


dgryski:
Should this land for 0.24 ?

aykevl:
This is mainly useful for people who build TinyGo manually using apt.llvm.org or Homebrew, otherwise it doesn't really matter. So I don't really have a preference.

dgryski (approved):
LGTM

deadprogram:
Since LLVM 14 is now the default for Homebrew, let's merge.

PR #2933, opened 2022-06-26
Update usb package and add support for USBMIDI / rp2040

sago35:
This PR includes the following changes.
Perhaps it would be better to break it up into multiple PRs.

1. Refactor machine/usb
    * Move usbcdc to machine/usb/cdc
2. Add machine/usb/midi
3. Add support for rp2040

I am testing below for now.

* qtpy (atsamd21)
* feather-m4-can (atsame51)
* feather-nrf52840-sense (nrf52840)
* pico (rp2040)
* feather-rp2040 (rp2040)

USBMIDI tests can be performed, for example.
https://mikatahara.github.io/SendReceive/

The rp2040 may act a little weird when flashing from `tinygo flash -programmer=picoprobe`.
In that case, press the reset button once and `tinygo flash -programmer=picoprobe` again.
If `openocd` is not used, there is no problem.

I would like to merge this PR into TinyGo 0.25.

The following tasks still remain to be completed

* [x] Add support for other boards that use rp2040
* [x] `-serial=uart` to work as well
* [x] Delete unnecessary comments
* [ ] Making CI a Success


sago35:
If you use an rp2040 board other than pico and feather-rp2040, please update the following files.
It would be great if you could make a commit.

* src/machine/board_xxxx.go
    * see https://github.com/tinygo-org/tinygo/blob/695cbc2a50a39f7530eb69819ffb071753fc5a57/src/machine/board_pico.go
* targets/xxxx.json
    * see https://github.com/tinygo-org/tinygo/blob/695cbc2a50a39f7530eb69819ffb071753fc5a57/targets/pico.json

sago35:
Once the CI is successful, I plan to break this PR into multiple PRs.
Perhaps divide it as follows.

1. targets/*.json + machine/board_xxx.go
2. refactor usb
3. add support for rp2040
4. add support for usb-midi

But even at this point I would still like a lot of different opinions.

sago35:
It was divided into the following PRs

* #2937
* #2973
* MIDI PR will be created later.

PR #2934, opened 2022-06-27
makefile: add badger2040, nano-rp2040, thingplus-rp2040 to smoketest

sago35:
We have probably forgotten to add.
I added this because I think it needs to be tested against the usb package update.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -557,6 +557,10 @@ endif
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=macropad-rp2040 	examples/blinky1
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=badger2040          examples/blinky1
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -557,6 +557,12 @@ endif
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=macropad-rp2040 	examples/blinky1
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=badger2040          examples/blinky1
+	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=nano-rp2040         examples/blinky1
```
deadprogram: This board is already in the Makefile. :smiley_cat:

sago35: oops..

deadprogram:
Thanks @sago35

PR #2940, opened 2022-06-29
Pin name consistency

aykevl:
This is necessary for https://github.com/tinygo-org/tinygo-site/pull/272.

I've split this up as much as possible, for easier review. Most commits should be non-functional or purely additive.

Perhaps this should be merged _after_ the release, to avoid unintentional breakage. There should be no breakage, but you never know when so many lines change.


aykevl:
Rebased to fix a merge conflict. It is ready now.

deadprogram:
That was a lot of manual effort, thanks for doing it. We need to keep things tidy from now on.

PR #2943, opened 2022-06-29
build: back to windows-2019 image, also Go 1.18.3

deadprogram:
Restore functioning build for Windows by dropping back to older `windows-2019` runner.

Also use latest Go release 1.18.3


aykevl:
Looks like LLVM wasn't rebuilt here, so it might not have worked.

deadprogram:
It did not break cache, and ALSO failed tests, so not going to work apparently.

PR #2944, opened 2022-06-29
ci: disable Ccache on Windows

aykevl:
Let's see whether this works.
Commented out the part after the LLVM build so we can get a cache. After LLVM finishes, that part should be uncommented to run all tests.

PR #2948, opened 2022-06-30
ci: use Bash for Windows release artifacts

deadprogram:
ci: use Bash for Windows release artifacts


deadprogram:
Also includes the fix from https://github.com/tinygo-org/tinygo/pull/2944

Now merging.

PR #2949, opened 2022-07-01
targets: make leveldb runtime hash default for wasi

dgryski:
Arguably we should adjust other targets, but lets start small...


deadprogram (approved):
LGTM

PR #2953, opened 2022-07-02
Initial support for board XIAO RP2040

conejoninja:
Looks like lib/wasi-libc was modified/updated, not sure why/how. Don't know how to drop it from the commit either.

Initial support, onboard RED led works, need to check all other pins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ See the [getting started instructions](https://tinygo.org/getting-started/) for 
 
 You can compile TinyGo programs for microcontrollers, WebAssembly and Linux.
 
-The following 85 microcontroller boards are currently supported:
+The following 86 microcontroller boards are currently supported:
 
 * [Adafruit Circuit Playground Bluefruit](https://www.adafruit.com/product/4333)
 * [Adafruit Circuit Playground Express](https://www.adafruit.com/product/3333)
@@ -114,6 +114,7 @@ The following 85 microcontroller boards are currently supported:
 * [Raytac MDBT50Q-RX Dongle (with TinyUF2 bootloader)](https://www.adafruit.com/product/5199)
 * [Seeed Seeeduino XIAO](https://www.seeedstudio.com/Seeeduino-XIAO-Arduino-Microcontroller-SAMD21-Cortex-M0+-p-4426.html)
 * [Seeed XIAO BLE](https://www.seeedstudio.com/Seeed-XIAO-BLE-nRF52840-p-5201.html)
+* [Seeed XIAO RP2040](https://www.seeedstudio.com/XIAO-RP2040-v1-0-p-5026.html)
 * [Seeed LoRa-E5 Development Kit](https://www.seeedstudio.com/LoRa-E5-Dev-Kit-p-4868.html)
 * [Seeed Sipeed MAix BiT](https://www.seeedstudio.com/Sipeed-MAix-BiT-for-RISC-V-AI-IoT-p-2872.html)
 * [Seeed Wio Terminal](https://www.seeedstudio.com/Wio-Terminal-p-4509.html)
--- a/src/machine/board_xiao-rp2040.go
+++ b/src/machine/board_xiao-rp2040.go
@@ -1,0 +1,79 @@
+//go:build xiao_rp2040
+// +build xiao_rp2040
+
+// This file contains the pin mappings for the Seeed XIAO RP2040 boards.
+//
+// XIAO RP2040 is a microcontroller using the Raspberry Pi RP2040 chip.
+//
+// - https://wiki.seeedstudio.com/XIAO-RP2040/
+//
+package machine
+
+// Digital Pins
+const (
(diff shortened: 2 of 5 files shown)


aykevl:
Looks like you reverted the wasi-libc change I did recently.
One way to fix this is as follows:
```
git reset HEAD~ # undo last commit
git add # add files as usual (but NOT lib/wasi-libc)
git commit # redo the commit
git push -f # update the PR
```
Hope that helps!

conejoninja:
Thanks!

I already did a
```
git reset --hard origin/release
```
before branching out, but looks like I missed the 
```
git submodule update
```

kenbell:
@conejoninja - looks like the PR is against release branch, you'll want to target the dev branch.

conejoninja:
Thanks, changed

README.md
```diff
@@ -114,6 +114,7 @@ The following 85 microcontroller boards are currently supported:
 * [Raytac MDBT50Q-RX Dongle (with TinyUF2 bootloader)](https://www.adafruit.com/product/5199)
 * [Seeed Seeeduino XIAO](https://www.seeedstudio.com/Seeeduino-XIAO-Arduino-Microcontroller-SAMD21-Cortex-M0+-p-4426.html)
 * [Seeed XIAO BLE](https://www.seeedstudio.com/Seeed-XIAO-BLE-nRF52840-p-5201.html)
+* [Seeed XIAO RP2040](https://www.seeedstudio.com/XIAO-RP2040-v1-0-p-5026.html)
```
kenbell, line 117: There's a count of boards at the top of the list, increment by 1.

kenbell:
You probably want to also add to the `smoketest` target in the Makefile to make sure it doesn't get broken.

src/machine/board_xiao-rp2040.go
```diff
@@ -0,0 +1,78 @@
...
+
+// USB CDC identifiers
+const (
+	usb_STRING_PRODUCT      = "XIAO RP2040"
+	usb_STRING_MANUFACTURER = "Seeed"
+)
+
+var (
+	usb_VID uint16 = 0x2e8a
+	usb_PID uint16 = 0x0003
```
sago35: Looking at the following, the PID of the XIAO RP2040 appears to be 0x000A.
Perhaps it would be better to adapt it to the Arduino environment.
However, I am not sure if 0x000A is really the correct value.
If 0x0003 is the value used by the Arduino, then the PID in #2936 may also need to be changed slightly.

https://github.com/earlephilhower/arduino-pico/blob/master/boards.txt

conejoninja: In bootloader mode lsusb output is : 

```
Bus 001 Device 024: ID 2e8a:0003 
```

This is the as badger2040 : https://github.com/tinygo-org/tinygo/blob/release/src/machine/board_badger2040.go#L80

conejoninja:
Thanks all for your patience and comments, it's been a while since my last contribution and I forget a few things.

src/machine/board_xiao-rp2040.go
sago35: I confirmed this with xiao-samd21 and xiao-rp2040.

* xiao samd21
  * When built on Arduino, it is 2886/802F.
  * The bootloader is 2886/002F.
  * `board_xiao.go` is set to 2886/802F.
* xiao rp2040
  * When built on Arduino, it is 2E8A/000A.
  * The bootloader is 2886/0003.

I think 2E8A/000A is correct this time, since at least tinygo-flashed state is not bootloader.

https://github.com/tinygo-org/tinygo/blob/b65447c7d567eea495805656f45472cc3c483e03/targets/xiao.json#L5
https://github.com/tinygo-org/tinygo/blob/b65447c7d567eea495805656f45472cc3c483e03/src/machine/board_xiao.go#L96-L99

deadprogram:
@conejoninja do you want to squash these commits into a single one, or would you prefer I do at merge time?

conejoninja:
It wasn't clear to me if  usbPID should be 0x0003 or 0x000a

```go
usb_PID uint16 = 0x0003
```

It works with both values, but lsusb only shows 0x0003 while on bootloader mode, while running the code it doesn't show on the pc. I wasn't able to get 0x000a value.


Please advise.

Anyway, squashed if 0x0003 is ok.

sago35:
@conejoninja 
At least a product ID of 0x0003 indicates a working bootloader.
This value can be set to any value from the program.
If set to 0x000A, the rp2040 application will appear to be running from the Arduino IDE, not the bootloader.
So 0x000a would be better.
2e8a/000a is the same value as the pico application, but that is not a problem.


When 2e8a/0004 is used, it is recognized as picoprobe.

conejoninja:
changed to 0x000a

sago35 (approved):

deadprogram:
Thanks @conejoninja @sago35 and @kenbell now merging.

PR #2954, opened 2022-07-02
Implement time.NewTimer and time.NewTicker 

kenbell:
This is a quick port of https://github.com/tinygo-org/tinygo/pull/1402 to current dev - seems to be functional...

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -65,6 +65,7 @@ func TestBuild(t *testing.T) {
 		"stdlib.go",
 		"string.go",
 		"structs.go",
+		"timers.go",
 		"zeroalloc.go",
 	}
 	_, minor, err := goenv.GetGorootVersion(goenv.Get("GOROOT"))
@@ -180,6 +181,14 @@ func runPlatTests(options compileopts.Options, tests []string, t *testing.T) {
 	}
 
 	for _, name := range tests {
+		if options.GOOS == "linux" && (options.GOARCH == "arm" || options.GOARCH == "386") {
+			switch name {
+			case "timers.go":
+				// Timer tests do not work because syscall.seek is implemented
+				// as Assembly in mainline Go and causes linker failure
+				continue
+			}
+		}
 		if options.Target == "simavr" {
 			// Not all tests are currently supported on AVR.
 			// Skip the ones that aren't.
@@ -206,6 +215,11 @@ func runPlatTests(options compileopts.Options, tests []string, t *testing.T) {
 
 			case "cgo/":
 				// CGo does not work on AVR.
+				continue
+
+			case "timers.go":
+				// Doesn't compile:
+				//   panic: compiler: could not store type code number inside interface type code
 				continue
 
 			default:
--- a/src/runtime/timer_go116.go
+++ b/src/runtime/timer_go116.go
@@ -1,0 +1,36 @@
+//go:build !go1.18
+// +build !go1.18
+
+// Portions copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package runtime
+
+type puintptr uintptr
+
+// Package time knows the layout of this structure.
+// If this struct changes, adjust ../time/sleep.go:/runtimeTimer.
+type timer struct {
+	// If this timer is on a heap, which P's heap it is on.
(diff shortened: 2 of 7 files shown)


kenbell:
@aykevl this is pretty much your code.  Let me know if you'd sooner take it over.

sago35:
Very nice. This code works on windows11.

```go
package main

import (
	"fmt"
	"runtime"
	"time"
)

func main() {
	run()
	fmt.Printf("done\n")
}

func run() {
	ch := time.After(1 * time.Second)

	for {
		select {
		case <-ch:
			fmt.Printf("expired\n")
			return
		default:
			runtime.Gosched()
		}
	}

}
```

And with this change, the following tests will be OK.
This is very nice.

* encoding/base64
* image/jpeg
* mime/quotedprintable

https://tinygo.org/docs/reference/lang-support/stdlib/

deadprogram:
@aykevl what do you think? This would be a great feature to get in to this release.

deadprogram:
@aykevl what do you think about this PR? It would be great to merge it! :smile_cat:

main_test.go
```diff
@@ -181,6 +182,14 @@ func runPlatTests(options compileopts.Options, tests []string, t *testing.T) {
 	}
 
 	for _, name := range tests {
+		if options.GOOS == "linux" && (options.GOARCH == "arm" || options.GOARCH == "386") {
+			switch name {
+			case "timers.go":
+				// Timer tests do not work because syscall.seek is implemented
+				// as Assembly in mainline Go and causes linker failure
```
aykevl, line 188: Ugh I really don't like this, but don't see an easy fix so :shrug:.

src/runtime/timer_go116.go
```diff
@@ -0,0 +1,32 @@
+//go:build !go1.18
+// +build !go1.18
+
+package runtime
+
+type puintptr uintptr
+
+// Package time knows the layout of this structure.
+// If this struct changes, adjust ../time/sleep.go:/runtimeTimer.
+type timer struct {
```
aykevl, line 14: These two files (timer_go116.go and timer_go118.go) seem to be copied from the Go source tree and should therefore have an appropriate notice at the top. Something like this:

https://github.com/tinygo-org/tinygo/blob/f0391eac258ca07ed724526d9f838c4a48333572/src/os/file.go#L1-L3

aykevl (commented):
Overall looks good to me. After all, it's mostly my own code :)

I have a few comments though and a request to add this patch, so that `go test -target=simavr` continues to work:

```diff
--- a/main_test.go
+++ b/main_test.go
@@ -218,6 +218,11 @@ func runPlatTests(options compileopts.Options, tests []string, t *testing.T) {
                                // CGo does not work on AVR.
                                continue
 
+                       case "timers.go":
+                               // Doesn't compile:
+                               //   panic: compiler: could not store type code number inside interface type code
+                               continue
+
                        default:
                        }
                }
```

(I also should add that you seem to have removed some of the uglier bits of #1402, which is good).

aykevl:
@kenbell perhaps you can also include some of the original commit message in this commit?

deadprogram:
Looks like need to run `go fmt`:

```
Unformatted:
  src/runtime/timer_go118.go
make: *** [Makefile:180: fmt-check] Error 1
```

kenbell:
@deadprogram - thanks, yeah.  Hopefully fixed that.  The MacOS builds seem to be unreliable with the test-case as well.  I think we need to either tweak the test case, or exclude on MacOS.  I'll try tweaking it next, see if it can be more reliable.  I'm guessing the CI worker is being paged out or something so go routines aren't running at the resolution of the test case?

deadprogram:
I am very excited about this feature.

Thank you @kenbell for actually getting it in, and thanks @aykevl for all the original work on it.

Now merging!

deadprogram:
Hmm, seems like macOS has troubles passing CI https://github.com/tinygo-org/tinygo/runs/7972019922?check_suite_focus=true#step:12:19780

Trying to run it a third time now...

kenbell:
@deadprogram - might want to back this out from dev until it's more robust in CI.  I've been gradually backing off the timing on the test case, but looks like not far enough to be reliable yet. :(

I'll take another pass at increasing the timeouts.

Issue #2938, opened 2022-06-28
RP2040 sometimes does not start up


sago35:
As shown in the following image, pressing the RESET button while Blinky1 is running may not start it.
It always seems to work immediately after `tinygo flash`.
It may not work after pressing RESET or turning the power back on.
There may be a problem with the initialization code.

Reproduced with TinyGo 0.23 and dev branch.

![image](https://user-images.githubusercontent.com/9251039/176149592-4f7df52e-43cd-46f0-ba44-13607901c4f1.png)

kenbell:
@sago35 I've noticed some slight weirdness on reset sometimes - putting a sleep at the start of main fixed it for me.  Wonder if it's just a timing thing?

sago35:
After making the changes in #2960, the problem has been completely resolved.

From @kenbell's advice, I was able to check pico and notice it.
Thank you very much.

PR #2960, opened 2022-07-05
feather-rp2040,macropad-rp2040: fix qspi-flash settings

sago35:
This PR resolves the issue of #2938

feather-rp2040 and macropad-rp2040 seem to be configured differently because they use slightly slower qspi-flash.
It would be good to keep it together with the pico-sdk source code.

PICO_FLASH_SPI_CLKDIV == 4
https://github.com/raspberrypi/pico-sdk/blob/2e6142b15b8a75c1227dd3edbe839193b2bf9041/src/boards/include/boards/adafruit_feather_rp2040.h#L80-L82
https://github.com/raspberrypi/pico-sdk/blob/2e6142b15b8a75c1227dd3edbe839193b2bf9041/src/boards/include/boards/adafruit_macropad_rp2040.h#L170-L172

PICO_FLASH_SPI_CLKDIV == 2
https://github.com/raspberrypi/pico-sdk/blob/2e6142b15b8a75c1227dd3edbe839193b2bf9041/src/boards/include/boards/pico.h#L69-L71
(All other boards are CLKDIV==2)

When adding RP2040 boards in the future, it is better to check the following
https://github.com/raspberrypi/pico-sdk/tree/master/src/boards/include/boards

### DIFF
--- a/targets/macropad-rp2040-boot-stage2.S
+++ b/targets/macropad-rp2040-boot-stage2.S
@@ -1,17 +1,17 @@
-// Adafruit Feather RP2040 Stage 2 Bootloader
+// Adafruit MacroPad RP2040 Stage 2 Bootloader
 
 //
 // This file defines the parameters specific to the flash-chip found
-// on the Adafruit Feather RP2040.  The generic implementation is in
+// on the Adafruit MacroPad RP2040.  The generic implementation is in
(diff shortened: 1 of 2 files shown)


deadprogram:
Amazing work finding and fixing this @sago35 and @kenbell thank you!

PR #2937, opened 2022-06-27
usb: refactor

sago35:
This PR is a refactoring of the USB that is part of #2933 
To be more specific, the following changes have been made

* Move usbcdc code to `machine/usb/cdc`
* Move usb code from `machine_samdXX.go` to `machine_samdXX_usb.go` (samd21 and samd51)
* Change Bootloader transition function in nrf to ResetProcessor
* Change callback from usb to three types (In / Out / ControlIn)
* Add `machine.InitSerial()` in runtime.init()
    * Configure may fail due to the initialization order of init()
* Split nrf code into nrf52840 and others (runtime/runetime_nrf*.go)

PR #2962, opened 2022-07-05
serial: use common initialization for serial

sago35:
This PR is intended to further break #2937 into smaller pieces for easier review.

This change is seemingly unnecessary, but is necessary to move usbcdc to machine/usb/cdc.
We can change it later if we find a better way.

After this PR, the final code will look like this

https://github.com/tinygo-org/tinygo/blob/0a54fad649c64e2a5e66dc9ffe561acb49c8e257/src/machine/serial-usb.go#L4-L10

https://github.com/tinygo-org/tinygo/blob/0a54fad649c64e2a5e66dc9ffe561acb49c8e257/src/machine/usb/cdc/usbcdc.go#L212-L215

### DIFF
--- a/src/machine/serial-none.go
+++ b/src/machine/serial-none.go
@@ -5,3 +5,7 @@ package machine
 
 // Serial is a null device: writes to it are ignored.
 var Serial = NullSerial{}
+
+func InitSerial() {
+	Serial.Configure(UARTConfig{})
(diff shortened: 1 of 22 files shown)


deadprogram (approved):
Make sense, encapsulates changes into a small surface area.

PR #2963, opened 2022-07-05
samd21,samd51,nrf52840: unify bootloader entry process

sago35:
This PR is intended to further break #2937 into smaller pieces for easier review.
This PR must be merged after #2962 

Unify the functions that go into Bootloarder to make it easier to call from `machine/usb/cdc` .

PR #2965, opened 2022-07-05
samd21,samd51,nrf52840: refactor usb initialization

sago35:
This PR is intended to further break #2937 into smaller pieces for easier review.
This PR must be merged after #2963 

This PR separates USB Device initialization from USBCDC.
Also move some common items to usb.go.
Renames several functions with the same meaning.
Separate runtime for nrf52840

I have confirmed that examples/echo works on the following boards
* feather-m0
* feather-m4-can
* feather-nrf52840-sense

### DIFF
--- a/src/runtime/runtime_nrf52840.go
+++ b/src/runtime/runtime_nrf52840.go
@@ -1,5 +1,5 @@
-//go:build nrf && !nrf52840
-// +build nrf,!nrf52840
+//go:build nrf && nrf52840
+// +build nrf,nrf52840
 
 package runtime
 
@@ -28,6 +28,7 @@ func main() {
 }
 
 func init() {
+	machine.USBDev.Configure(machine.UARTConfig{})
 	machine.InitSerial()
 	initLFCLK()
 	initRTC()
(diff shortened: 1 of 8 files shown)


deadprogram:
Ready for rebase against `dev` please.

sago35:
Rebased.

src/runtime/runtime_nrf52840.go
```diff
@@ -0,0 +1,149 @@
+//go:build nrf && nrf52840
```
deadprogram, line 1: Seems like a lot of changes are needed for the nrf52840, how was this working before without all of this initialization?

sago35: Until now, `machine.Serial.Configure` has only been called.
Except for nrf52840, `machine.Serial` was set to use `serial-uart.go`.
So we were able to use the same source code for all nrf's including nrf52840.

However, that does not allow us to merge the changes in #2631
In this case, nrf52840 wants to perform USB initialization with or without serial-uart.go.

https://github.com/tinygo-org/tinygo/blob/b65447c7d567eea495805656f45472cc3c483e03/src/runtime/runtime_nrf.go#L30-L34

deadprogram: OK, understood. Thank you for the clarification.

PR #2967, opened 2022-07-06
Add reflect side table for map type info

dgryski:


dgryski:
PR is broken and doesn't work but contains my first attempt at adding a side table so we have the appropriate type information for maps for improved reflect support.

dgryski:
@aykevl Should I try to finish this or is https://github.com/tinygo-org/tinygo/pull/2640 sufficiently close to being done and we should focus our efforts there?

dgryski:
New stuff is better.

PR #2966, opened 2022-07-06
samd21,samd51,nrf52840: change usbSetup and sendZlp to public

sago35:
This PR is intended to further break https://github.com/tinygo-org/tinygo/pull/2937 into smaller pieces for easier review.
This PR must be merged after https://github.com/tinygo-org/tinygo/pull/2965

PR #2968, opened 2022-07-06
samd21,samd51,nrf52840: refactor handleStandardSetup and initEndpoint

sago35:
This PR is intended to further break https://github.com/tinygo-org/tinygo/pull/2937 into smaller pieces for easier review.
This PR must be merged after https://github.com/tinygo-org/tinygo/pull/2966

handleStandardSetup() is now unified in this PR.
~~The next PR moves handleStandardSetup() to usb.go.
That next PR will complete everything.~~

In the second (or 3rd??) commit of this PR, handleStandardSetup was moved to usb.go.


sago35:
It looks like the nrf is a little broken.
Will check.

deadprogram:
Also ready for rebase against `dev`.

sago35:
Rebased.
~~If there is no problem, you may squash merge.~~


> It looks like the nrf is a little broken.

~~1 commit added.~~

* ~~Add nrf52840 workaround for errata~~
* ~~Fix multiple calls to configure~~

Fixing configure is a separate PR.

sago35:
Rebased.
This PR change is fine.
If there is no problem, you may squash merge.

---

However, the problem is that I erased this part of nrf52840, which makes it easy for initialization to fail.
And I think this part is needed for samd21 and samd51 as well.
I will create a separate PR for this issue.

https://github.com/tinygo-org/tinygo/blob/1b2e764835fb5278de144caf059cb094bad4db23/src/machine/machine_nrf52840_usb.go#L172-L174

For example, the following cases can cause exactly the problem

https://github.com/tinygo-org/tinygo/blob/b65447c7d567eea495805656f45472cc3c483e03/src/examples/echo/echo.go#L18

PR #2971, opened 2022-07-07
atsamd21,atsamd51,nrf52840: improve usb-device initialization

sago35:
This PR is intended to further break https://github.com/tinygo-org/tinygo/pull/2937 into smaller pieces for easier review.
This PR must be merged after https://github.com/tinygo-org/tinygo/pull/2968

This PR fixes the following issues

* Fixed problem when configure is called multiple times (-> dev.initcomplete)
* Add workaround to errata 187 in nrf52840

> the problem is that I erased this part of nrf52840, which makes it easy for initialization to fail.
> And I think this part is needed for samd21 and samd51 as well.
> 
> https://github.com/tinygo-org/tinygo/blob/1b2e764835fb5278de144caf059cb094bad4db23/src/machine/machine_nrf52840_usb.go#L172-L174
> 
> For example, the following cases can cause exactly the problem
> 
> https://github.com/tinygo-org/tinygo/blob/b65447c7d567eea495805656f45472cc3c483e03/src/examples/echo/echo.go#L18

### DIFF
--- a/src/machine/machine_nrf52840_usb.go
+++ b/src/machine/machine_nrf52840_usb.go
@@ -153,6 +153,15 @@ func exitCriticalSection() {
 
 // Configure the USB peripheral. The config is here for compatibility with the UART interface.
 func (dev *USBDevice) Configure(config UARTConfig) {
+	if dev.initcomplete {
+		return
+	}
+
+	state := interrupt.Disable()
+	defer interrupt.Restore(state)
+
+	nrf.USBD.USBPULLUP.Set(0)
+
 	// Enable IRQ. Make sure this is higher than the SWI2 interrupt handler so
 	// that it is possible to print to the console from a BLE interrupt. You
 	// shouldn't generally do that but it is useful for debugging and panic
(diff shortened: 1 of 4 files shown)


deadprogram:
Ready for rebase against `dev` please.

src/machine/machine_nrf52840_usb.go
```diff
@@ -161,19 +170,33 @@ func (dev *USBDevice) Configure(config UARTConfig) {
 	intr.SetPriority(0x40) // interrupt priority 2 (lower number means more important)
 	intr.Enable()
 
+	// enable interrupt for end of reset and start of frame
+	nrf.USBD.INTEN.Set(nrf.USBD_INTENSET_USBEVENT | nrf.USBD_INTENSET_SOF)
+
+	// errata 187
+	// https://infocenter.nordicsemi.com/topic/errata_nRF52840_EngB/ERR/nRF52840/EngineeringB/latest/anomaly_840_187.html
```
deadprogram, line 177: Impressive to find this! Great research @sago35

PR #2972, opened 2022-07-07
samd21,samd51,nrf52840: move usbcdc to machine/usb/cdc

sago35:
This PR is intended to further break https://github.com/tinygo-org/tinygo/pull/2937 into smaller pieces for easier review.
This PR must be merged after https://github.com/tinygo-org/tinygo/pull/2968 and  #2971 

Finally, the functionality is the same as the source code in #2937.
After this PR is merged, the following can be added

* add support for rp2040's USB
* add support usb-midi

PR #2973, opened 2022-07-07
rp2040: add support for usb

sago35:
This PR can be merged after #2972 is merged.

### DIFF
--- a/src/machine/machine_rp2040_usb.go
+++ b/src/machine/machine_rp2040_usb.go
@@ -128,30 +128,31 @@ func handleUSBIRQ(intr interrupt.Interrupt) {
 }
 
 func initEndpoint(ep, config uint32) {
-	val := uint32(0x80000000) | uint32(0x20000000)
-	offset := ep*2*64 + 0x100
+	val := uint32(usbEpControlEnable) | uint32(usbEpControlInterruptPerBuff)
+	offset := ep*2*USBBufferLen + 0x100
 	val |= offset
 
 	switch config {
 	case usb_ENDPOINT_TYPE_INTERRUPT | usbEndpointIn:
-		val |= 0x0C000000
+		val |= usbEpControlEndpointTypeInterrupt
 		usbDPSRAM.EPxControl[ep].In = val
 
 	case usb_ENDPOINT_TYPE_BULK | usbEndpointOut:
-		val |= 0x08000000
+		val |= usbEpControlEndpointTypeBulk
 		usbDPSRAM.EPxControl[ep].Out = val
-		usbDPSRAM.EPxBufferControl[ep].Out = 0x00000040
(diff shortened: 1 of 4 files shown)


sago35:
At the moment, it is not possible to enter the bootloader from USBCDC.

We can probably enter the bootloader by implementing the following
https://github.com/raspberrypi/pico-sdk/blob/2e6142b15b8a75c1227dd3edbe839193b2bf9041/src/rp2_common/pico_bootrom/include/pico/bootrom.h#L144-L165

sago35:
The following options (`-serial usb`) are required to use USB CDC.

```
$ tinygo flash --size short --target feather-rp2040 --serial usb ./src/examples/echo
```

src/machine/machine_rp2040_usb.go
```diff
@@ -0,0 +1,309 @@
...
+			// Standard Requests
+			ok = handleStandardSetup(setup)
+		} else {
+			// Class Interface Requests
+			if setup.WIndex < uint16(len(usbSetupHandler)) && usbSetupHandler[setup.WIndex] != nil {
+				ok = usbSetupHandler[setup.WIndex](setup)
+			}
+		}
+
+		if ok {
```
deadprogram: Is there still some working pending here to be done, or should this be removed?

src/machine/machine_rp2040_usb.go
```diff
@@ -0,0 +1,309 @@
...
+
+	// Bus is reset
+	if (status & rp.USBCTRL_REGS_INTS_BUS_RESET) > 0 {
+		rp.USBCTRL_REGS.SIE_STATUS.Set(rp.USBCTRL_REGS_SIE_STATUS_BUS_RESET)
+		initEndpoint(0, usb_ENDPOINT_TYPE_CONTROL)
+	}
+}
+
+func initEndpoint(ep, config uint32) {
+	val := uint32(0x80000000) | uint32(0x20000000)
```
deadprogram: It would be great to have some const definitions for `0x80000000`, `0x20000000` and the other "magic numbers" here.

deadprogram:
@sago35 seems like this PR is so close! What do you think about my feedback?

src/machine/machine_rp2040_usb.go
sago35: USB-stall implemented.
I have confirmed that I can send stalls, but as far as I can see I have not encountered an opportunity to send stalls.

src/machine/machine_rp2040_usb.go
sago35: Added const definition.

sago35:
@deadprogram Updated.

deadprogram:
Fantastic work here @sago35 now squash/merging.

diceptia (approved):

PR #2981, opened 2022-07-12
usb: rename callback to handler to keep consistent

deadprogram:
This PR renames all of the USB "callbacks" to "handlers" to keep consistent.

### DIFF
--- a/src/machine/usb/midi/midi.go
+++ b/src/machine/usb/midi/midi.go
@@ -12,9 +12,9 @@ const (
 var Midi *midi
 
 type midi struct {
-	buf            *RingBuffer
-	callbackFuncRx func([]byte)
-	waitTxc        bool
+	buf       *RingBuffer
+	rxHandler func([]byte)
+	waitTxc   bool
 }
 
 func init() {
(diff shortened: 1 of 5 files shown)


src/machine/usb/midi/midi.go
```diff
@@ -32,12 +32,12 @@ func newMidi() *midi {
 	m := &midi{
 		buf: NewRingBuffer(),
 	}
-	machine.EnableMIDI(m.Callback, m.CallbackRx, nil)
+	machine.EnableMIDI(m.Handler, m.rxHandler, nil)
```
sago35, line 35: RxHandler is correct here.

```suggestion
	machine.EnableMIDI(m.Handler, m.RxHandler, nil)
```

sago35 (changes requested):
@deadprogram 
It is very easy to understand.
Thank you very much.

Please correct only the part I commented on.
Otherwise, very good.

src/machine/usb/midi/midi.go
deadprogram: Since there is already a `RxHandler()` method, the internal field cannot be exported. Otherwise you end up with error:

```
/home/ron/Development/tinygo/tinygo/src/machine/usb/midi/midi.go:75:16: field and method with the same name RxHandler
```

deadprogram: In other words, I think it is correct the way I have it.

sago35: @deadprogram 
Since `m.rxHandler` at the time `machine.EnableMIDI` is called is nil, the callback is no longer called.
`examples/usb-midi` toggles LEDs whenever MIDI is received.

https://mikatahara.github.io/SendReceive/

image

deadprogram: You are certainly correct @sago35 I got confused by the naming :smiley_cat: 

Corrected in PR https://github.com/tinygo-org/tinygo/pull/2988

PR #2986, opened 2022-07-14
rp2040: change default for serial to usb

sago35:
This change breaks compatibility.
However, in many cases it is better to default to USB rather than UART for serial.

An additional serial-port setting has been added.
This is a necessary parameter after the user is able to enter the bootloader via USBCDC.


deadprogram (approved):
Agree that it is better to have RP2040 act like all other boards with UF2 bootloaders/USB support.

PR #2990, opened 2022-07-14
rp2040: add support for EnterBootloader()

sago35:
It should be merged after #2989

This PR adds flash-1200-bps-reset support to the rp2040.
In other words, the BOOTSEL button no longer needs to be pressed.
(Of course, if something goes wrong, you will need BOOTSEL)

PR #2992, opened 2022-07-15
rp2040: change to ignore PortReset failures

sago35:
It should be merged after https://github.com/tinygo-org/tinygo/pull/2990

This PR fixes the tinygo-flash failure issue for rp2040 already in the bootloader.
In my environment, this does not occur on Windows, but on Linux (ubuntu20.04).

This becomes a problem when the following software is written.

```go
package main

import (
	"runtime/interrupt"
)

func main() {
	interrupt.Disable()
	panic("panic")
}
```

If you want to recover without having to make this PR change, you must make the following changes before tinygo flash

1. Set flash-1200-bps-reset in `targets/rp2040.json` to false
2. Press bootsel to start up.
3. tinygo flash

### DIFF
--- a/main.go
+++ b/main.go
@@ -350,16 +350,14 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 		// do we need port reset to put MCU into bootloader mode?
 		if config.Target.PortReset == "true" && flashMethod != "openocd" {
 			port, err := getDefaultPort(port, config.Target.SerialPort)
-			if err != nil {
-				return err
+			if err == nil {
+				err = touchSerialPortAt1200bps(port)
+				if err != nil {
+					return &commandError{"failed to reset port", result.Binary, err}
+				}
+				// give the target MCU a chance to restart into bootloader
+				time.Sleep(3 * time.Second)
 			}
-
-			err = touchSerialPortAt1200bps(port)
-			if err != nil {
-				return &commandError{"failed to reset port", result.Binary, err}
-			}
-			// give the target MCU a chance to restart into bootloader
-			time.Sleep(3 * time.Second)
 		}
 
 		// this flashing method copies the binary data to a Mass Storage Device (msd)


sago35:
@deadprogram 
If there is a better way, I would like to adopt it.

sago35:
I have found that this change is necessary for macos as well.
https://gophers.slack.com/archives/CDJD3SUP6/p1657877945224479?thread_ts=1657666757.320569&cid=CDJD3SUP6

sago35:
@deadprogram  We need to decide what to do about this PR issue before tinygo-0.25 is released.
What is the best way to do this?
Maybe the VID/PID of the bootloader of rp2040 is as follows

* VID: 0x2E8A
* PID: 0x0003

main.go
```diff
@@ -349,17 +349,23 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 	return builder.Build(pkgName, fileExt, config, func(result builder.BuildResult) error {
 		// do we need port reset to put MCU into bootloader mode?
 		if config.Target.PortReset == "true" && flashMethod != "openocd" {
-			port, err := getDefaultPort(port, config.Target.SerialPort)
-			if err != nil {
-				return err
-			}
+			err := func() error {
```
kenbell: Might be cleaner to separate out this logic into a function rather than do an inline function?

compileopts/target.go
```diff
@@ -23,45 +23,46 @@ import (
...
+	CFlags             []string `json:"cflags"`
+	LDFlags            []string `json:"ldflags"`
+	LinkerScript       string   `json:"linkerscript"`
+	ExtraFiles         []string `json:"extra-files"`
+	RP2040BootPatch    *bool    `json:"rp2040-boot-patch"` // Patch RP2040 2nd stage bootloader checksum
+	Emulator           string   `json:"emulator"`
+	FlashCommand       string   `json:"flash-command"`
+	GDB                []string `json:"gdb"`
+	PortReset          string   `json:"flash-1200-bps-reset"`
+	PortResetIgnoreErr string   `json:"flash-1200-bps-reset-ignore-error"`
```
kenbell: Do we need this setting?  Might be best to always ignore errors?  (I can't really think of a reason where a developer would turn it off)

kenbell (approved):
This fixes issue #3013 - just a couple of minor suggestions for implementation.

compileopts/target.go
sago35: I tried to reduce the impact on other targets as much as possible.
However, we certainly may not have to worry that much about it.

@deadprogram 
What do you think?

main.go
sago35: That would probably be in the form of including it in touchSerialPortAt1200bps().

compileopts/target.go
sago35: Hmmm, it is hard to understand the error message when there are multiple options at the same time.


Before correction
```
# When two Wio Terminals are present
$ tinygo flash --target wioterminal --size short ./b1
   7824     144    6260 |    7968    6404
error: multiple serial ports available - use -port flag, available ports are COM37, COM105
```

```
# When Wio Terminal does not exist
$ tinygo flash --target wioterminal --size short ./b1
   7824     144    6260 |    7968    6404
error: multiple serial ports available - use -port flag, available ports are COM48, COM104
```


After correction (always ignore error)
```
# When two Wio Terminals are present
$ tinygo flash --target wioterminal --size short ./b1
   7844     144    6260 |    7988    6404
error: failed to flash C:\Users\sago3\AppData\Local\Temp\tinygo1961070370\main.uf2: unable to locate a USB device to be flashed
```

```
# When Wio Terminal does not exist
$ tinygo flash --target wioterminal --size short ./b1
   7844     144    6260 |    7988    6404
error: failed to flash C:\Users\sago3\AppData\Local\Temp\tinygo3745000764\main.uf2: unable to locate a USB device to be flashed
```

deadprogram: Seems to me it would be better to either always ignore the PortReset failures, or else to find the root cause on the RP2040. Having this be "Yet Another Setting" is perhaps not the cleanest solution.

sago35: The root cause is clear.

The rp2040 bootloader fails to open the serial port because it does not define a serial port.

A bootloader such as samd51 defines a serial port and ignores 1200bps-open.
Therefore, flash-1200-bps-reset works well with and without bootloader.

sago35: PortReset failures are now ignored.
Error messages degrade in some cases, such as when multiple connections are made, but should be improved separately.

sago35:
Now ready for review.

> PortReset failures are now ignored.
> Error messages degrade in some cases, such as when multiple connections are made, but should be improved separately.

deadprogram:
Thanks for the detailed explanations @sago35 

Good idea also to defer working on the multiple connection error modes for now in favor of getting this family of boards to have a nice developer experience.

Now merging, thank you!

PR #2993, opened 2022-07-15
runtime: rename printuint to printuintptr

aykevl:
This is arguably the correct name, and is consistent with other print functions.


fgsch (approved):
👍

PR #2996, opened 2022-07-15
darwin: don't clobber X18 and FP registers

aykevl:
These are reserved registers on MacOS so don't need to be clobbered, and result in a compiler waring:

    ld.lld-14: warning: inline asm clobber list contains reserved registers: X18, FP
    Reserved registers on the clobber list may not be preserved across the asm statement, and clobbering them may lead to undefined behaviour.

I didn't test the resulting binaries, but at least they compile without the warning anymore.
(I don't think the clobber would harm because the compiler wouldn't use the given registers anyway, but the warning is annoying).

### DIFF
--- a/compiler/defer.go
+++ b/compiler/defer.go
@@ -161,7 +161,17 @@ str x2, [x1, #8]
 mov x0, #0
 1:
 `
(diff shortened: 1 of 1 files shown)


deadprogram:
Can someone please test this with a macOS system?

dgryski:
@deadprogram This needs testing specifically on an M1 machine.

dkegel-fastly:
I will test it... I have covid, though, so I may be a bit slow.

dkegel-fastly (approved):
WFM

MacOS M1 no longer warns about clobbers during "make tinygo-test".  Test results otherwise unchanged.

deadprogram:
Thanks for the fix @aykevl and thank you @dkegel-fastly for testing it out. Now merging.

Issue #2951, opened 2022-07-01
Call parameter type does not match function signature! when method input parameter is a generic type


yoursunny:
**main.go**

```go
package main

type C[V any] struct{}

func (c *C[V]) F(v N[V]) {}

type N[V any] struct{}

func main() {
	var c C[int]
	var n N[int]
	c.F(n)
}
```

Commands and error message:

```bash
yoursunny@localhost:~/snippet$ tinygo version
tinygo version 0.24.0 linux/amd64 (using go version go1.18.3 and LLVM version 14.0.0)

yoursunny@localhost:~/snippet$ tinygo build -target=wasm main.go
Call parameter type does not match function signature!
  %0 = bitcast i8* %c to %main.C*, !dbg !15
 %main.C.0*  call void @"(*main.C[int]).F[int]"(%main.C* %0, i8* undef), !dbg !17
error: verification error after compiling package command-line-arguments
```

nik0sc:
I faced a similar problem when compiling some code that uses generics. What I found is that this error occurs when there is at least one pointer argument to the method (including the receiver). So in the example code, if the method F has a value receiver `func (c C[V]) F(v N[V]) {}` the code compiles without errors.

aykevl:
See: https://github.com/tinygo-org/tinygo/pull/2998

deadprogram:
Closing since this is part of release 0.29.0 thank you!

PR #2998, opened 2022-07-15
compiler: fix issue with methods on generic structs

aykevl:
This is a fix for https://github.com/tinygo-org/tinygo/issues/2951, which may or may not be an upstream Go bug.


deadprogram:
Note CI failures. :boom:

aykevl:
Oops, it's actually a bug in this PR. I'll try to fix it.

aykevl:
Reported upstream, as I'm not sure how to fix it inside TinyGo: https://github.com/golang/go/issues/53914

dgryski:
Is moving to `typeutil.Map` as suggested in the bug an option for us?

pkedy:
Hi, @aykevl. I have a use case that needs this fix so I tested using `typeutil.Map` locally. I also wanted to run the fix through the full test suite so I opened #3003. All tests are passing. Feel free copy the diffs into this PR.

aykevl:
Fixed in #3003.

PR #3003, opened 2022-07-19
compiler: fix issue with methods on generic structs

pkedy:
Testing the fix for #2951 as discussed in #2998 (using `typeutil.Map`).

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -18,6 +18,7 @@ import (
 	"github.com/tinygo-org/tinygo/compiler/llvmutil"
 	"github.com/tinygo-org/tinygo/loader"
 	"golang.org/x/tools/go/ssa"
+	"golang.org/x/tools/go/types/typeutil"
 	"tinygo.org/x/go-llvm"
 )
 
@@ -70,7 +71,7 @@ type compilerContext struct {
 	cu               llvm.Metadata
 	difiles          map[string]llvm.Metadata
 	ditypes          map[types.Type]llvm.Metadata
-	llvmTypes        map[types.Type]llvm.Type
+	llvmTypes        typeutil.Map
 	machine          llvm.TargetMachine
 	targetData       llvm.TargetData
 	intType          llvm.Type
@@ -95,7 +96,6 @@ func newCompilerContext(moduleName string, machine llvm.TargetMachine, config *C
 		DumpSSA:     dumpSSA,
 		difiles:     make(map[string]llvm.Metadata),
 		ditypes:     make(map[types.Type]llvm.Metadata),
-		llvmTypes:   make(map[types.Type]llvm.Type),
 		machine:     machine,
 		targetData:  machine.CreateTargetData(),
 		astComments: map[string]*ast.CommentGroup{},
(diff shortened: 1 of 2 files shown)


compiler/compiler.go
```diff
@@ -389,9 +390,17 @@ func (c *compilerContext) makeLLVMType(goType types.Type) llvm.Type {
 			// in LLVM IR, named structs are implemented as named structs in
 			// LLVM. This is because it is otherwise impossible to create
 			// self-referencing types such as linked lists.
+			//
+			// Note: *types.Named isn't unique when working with generics.
+			// See https://github.com/golang/go/issues/53914
+			existing := c.llvmTypes.At(goType)
```
aykevl: This shouldn't be necessary with `typeutil.Map`.

aykevl (commented):
The fix I had in mind was to do this only for `*types.Named`, but doing it like this also works. (And might even be a tiny bit faster, because a `typeutil.Map` lookup is probably fastest than a CGo call).

pkedy:
@aykevl comment addressed. Please take another look.

pkedy:
Hold off on merging this. I'm getting `symbol multiply defined!` linker errors and trying to work through it.

pkedy:
@aykevl I was able to repro the issue and committed a test case to this PR. This is the error.

```
error: Linking globals named '(*github.com/tinygo-org/tinygo/testdata/generics/value.mapper[int, int]).Get[int int]$1': symbol multiply defined!
```

It seems to occur when multiple packages reference a package with generics in it AND the package with generics uses callback functions. There needs to be something to prevent duplicate global symbols between packages. Unfortunately, I don't have any idea where to look for the fix. Any ideas?

aykevl:
I think these generic functions should have linkage `linkonce_odr` (there is a method `SetLinkage` that you can use). That means they are treated the same as C++ template instantiations, and duplicates are eliminated.

I don't have the mental bandwidth right now to look into it. Ping me again in a week if this doesn't work or you can't figure it out.

pkedy:
@aykevl That seems to have worked! In the latest commit, I'm not sure that I'm detecting generic functions the best way (checking for the existence of `/\[[^\]]+\]/` in the name), but even my more complex scenario works. Perhaps you when you have time, you can check if there is a more sane approach. :smile:

Update: All the tests pass except for on Windows... story of my life.

aykevl:
Happy it works now!
Yes, there is a better way. See here (bottom of the Function documentation): https://pkg.go.dev/golang.org/x/tools/go/ssa#Function

> A function is generic iff it has a non-empty TypeParams list and an empty TypeArgs list. TypeParams lists the type parameters of the function's Signature or the receiver's type parameters for a method.

pkedy:
@aykevl Per the [Function documentation](https://pkg.go.dev/golang.org/x/tools/go/ssa#Function), I put this code in `createFunctionStart` to see which methods would list just based on non-empty TypeParams.

```golang
if b.fn.Signature.TypeParams().Len() > 0 {
	fmt.Println("Generic:", b.llvmFn.Name())
}
```

Unfortunately, nothing printed. Also, I'm not seeing an obvious method on Function/Signature for TypeArgs. Strangely, there are these two fields that have `_` prefixes so they are not accessible outside the `ssa` package.

```golang
	_TypeParams []*typeparams.TypeParam // the type paramaters of this function. len(TypeParams) == len(_TypeArgs) => runtime function
	_TypeArgs   []types.Type            // type arguments for for an instantiation. len(_TypeArgs) != 0 => instantiation
```

Admittedly, I have not had time to dig deeper.

To understand the Windows test failure, I need to fire up my gaming machine.

aykevl:
What about _only_ merging the first two commits (squashed together)? It has a seemingly obvious fix that is certainly an improvement, while the `symbol multiply defined!` appears to be a different issue that we only see now because of the previous fix. And the `symbol multiply defined!` issue needs a bit more investigation to fix (I'm sure there is a cleaner way to fix it somehow - there must be a way to detect generic instantiations somehow without looking at the symbol name).

pkedy:
> What about only merging the first two commits (squashed together)?
> the `symbol multiply defined!` appears to be a different issue ...  and needs a bit more investigation

💯 agree! I've reverted the last commit and squashed the rest.

pkedy:
Once this is merged, I'll create a separate issue and a draft PR that updates the generics test with the scenario that causes `symbol multiply defined!` and we can go from there.

aykevl (approved):

Issue #3004, opened 2022-07-19
USB Enumeration fails when trying to use crypto/rand with samd51


sago35:
I would like to make the reproduction code smaller, but have not been able to do so yet.

The problem will not occur if GetRNG() in machine_atsamd51.go is modified as follows.
![image](https://user-images.githubusercontent.com/9251039/179770994-c67f51ad-eefd-4b5b-b502-d2d2bc8c8e7e.png)

The code we are reproducing so far is as follows.
The examples/rand code is implemented at the beginning of the main function.

```go
package main

import (
	"crypto/rand"
	"encoding/hex"
	"fmt"
	"time"

	"tinygo.org/x/drivers/examples/rtl8720dn"
	"tinygo.org/x/drivers/net"
	"tinygo.org/x/drivers/net/mqtt"
)

var (
	ssid     string // 設定が必要
	password string // 設定が必要
	server   string = "tcp://test.mosquitto.org:1883"
	topic           = "tinygo-book"
)

func main() {
	var result [32]byte
	for {
		rand.Read(result[:])
		encodedString := hex.EncodeToString(result[:])
		println(encodedString)
		time.Sleep(time.Second)
	}

	//rtl8720dn.Debug(true)
	rtl, err := rtl8720dn.Setup()
	if err != nil {
		failMessage(err.Error())
	}
	err = rtl.ConnectToAccessPoint(ssid, password, 10*time.Second)
	if err != nil {
		failMessage(err.Error())
	}

	// tinygo.org/x/drivers/netを使うために必要
	net.UseDriver(rtl)

	opts := mqtt.NewClientOptions()
	var rndbuf [4]byte
	//rand.Read(rndbuf[:])
	opts.AddBroker(server).SetClientID(fmt.Sprintf("tinygo-client-% X", rndbuf))
	fmt.Printf("client id : %s\r\n", fmt.Sprintf("tinygo-client-% X", rndbuf))

	fmt.Printf("Connectng to MQTT : %s\r\n", server)
	cl := mqtt.NewClient(opts)
	if token := cl.Connect(); token.Wait() && token.Error() != nil {
		failMessage(token.Error().Error())
	}
	defer cl.Disconnect(100)

	for i := 0; ; i++ {
		fmt.Printf("Publishing MQTT message\r\n")
		data := []byte(fmt.Sprintf(`{"e":[{"n":"hello %d","v":101}]}`, i))
		token := cl.Publish(topic, 0, false, data)
		token.Wait()
[...]

PR #3005, opened 2022-07-20
usb: adjust buffer alignment (samd21, samd51, nrf52840)

sago35:
This PR solves the USB Buffer align problem for samd21 and samd51 and nrf52840.
(#3004)

* samd51
    * Bits 31:0 – ADDR[31:0] Data Pointer Address Value These bits define the data pointer address as an absolute word address in RAM. The two least significant bits must be zero to ensure the start address is 32-bit aligned.
* samd21
    * Bits 31:0 – ADDR[31:0]: Data Pointer Address Value These bits define the data pointer address as an absolute word address in RAM.The two least significant bits must be zero to ensure the start address is 32-bit aligned.
* nrf52840
    * Note: The control endpoint buffer size in Data RAM can be of any size in bytes, and there is no constraint to keep it 32-bit aligned.
    * The maximum size of a bulk/interrupt transaction in USB full-speed is 64 bytes, and it has to be a multiple of 4 bytes and 32-bit aligned in Data RAM. However, the amount of data bytes transmitted on the USB data endpoint can be of any size (up to 64 bytes).
* rp2040
    * Already aligned.
	usbDPSRAM = (*USBDPSRAM)(unsafe.Pointer(uintptr(0x50100000)))

### DIFF
--- a/src/machine/usb.go
+++ b/src/machine/usb.go
@@ -56,16 +56,21 @@ var (
 var (
 	usbEndpointDescriptors [usb.NumberOfEndpoints]usb.DeviceDescriptor
 
-	udd_ep_control_cache_buffer [256]uint8
-	udd_ep_in_cache_buffer      [7][64]uint8
-	udd_ep_out_cache_buffer     [7][64]uint8
-
 	isEndpointHalt        = false
 	isRemoteWakeUpEnabled = false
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Great finding @sago35 thank you! Now merging.

PR #3006, opened 2022-07-20
xiao-ble: add support for flash-1200-bps-reset

sago35:
I had been flashing from openocd and was unaware of this.
I think it would be better to be able to write from usb like the regular nrf52840 board or xiao-samd21.

And, although not directly related to this change, the initial value of serial was changed to usb.
This is another change to match other boards.
However, it is important to note that **compatibility can be broken**.

@deadprogram @ysoldak 
What do you think?

### DIFF
--- a/targets/xiao-ble.json
+++ b/targets/xiao-ble.json
@@ -1,8 +1,9 @@
 {
 	"inherits": ["nrf52840"],
-	"build-tags": ["xiao_ble", "softdevice", "s140v7"],
-	"serial": "uart",
+	"build-tags": ["xiao_ble", "nrf52840_reset_uf2", "softdevice", "s140v7"],
+	"serial": "usb",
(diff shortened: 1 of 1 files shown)


deadprogram:
I think it is a good idea to make to the same as the other nrf52840 based boards, even if it breaks compatibility. The current implementation is much less convenient than the new one in this PR, so it seems quite worth it to me.

ysoldak:
I have vague memories magic number for xiao ble is different and current implementation of nrf52840_reset_uf2 would not work. But I may be wrong, if it works, sure, change it.

ysoldak:
I’m AFK until August. Then, I can try and verify. Unless @sago35 tried already, of course.

sago35:
I have checked with both xiao-ble and xiao-ble-sense.
It is working fine.

deadprogram:
Thank you @sago35 for the fix and @ysoldak for the feedback. Now merging.

PR #3007, opened 2022-07-20
machine/usb/midi: add NoteOn, NoteOff, and SendCC methods

deadprogram:
This PR updates the `machine/usb/midi` package to add `NoteOn()`, `NoteOff()`, and `SendCC()` methods to have a more complete API.


sago35 (approved):
I think it is a very good change.
I think the operation has become very easy to understand.

deadprogram:

PR #3009, opened 2022-07-21
samd51: improve TRNG

sago35:
Two improvements were implemented.

I stopped using the rngInitDone variable.
This determination can be made by whether MCLK is running or not.

I made sure to wait for DATARDY.
Before the change, the first 4 bytes of the following code would be 0x00000000.
Also, consecutive calls to machine.GetRNG() with --opt=s may take the same value.
I believe crypto/rand needs to be modified because it needs to take a different value each time.

```
package main

import (
	"crypto/rand"
	"encoding/hex"
	"time"
)

func main() {
	time.Sleep(3 * time.Second)
	var result [32]byte
	for {
		rand.Read(result[:])
		encodedString := hex.EncodeToString(result[:])
		println(encodedString[:])
		time.Sleep(time.Second)
	}
}
```

### DIFF
--- a/src/machine/machine_atsamd51.go
+++ b/src/machine/machine_atsamd51.go
@@ -2067,18 +2067,16 @@ func (dac DAC) syncDAC() {
 	}
 }
 
-var rngInitDone = false
-
 // GetRNG returns 32 bits of cryptographically secure random data
 func GetRNG() (uint32, error) {
-	if !rngInitDone {
+	if !sam.MCLK.APBCMASK.HasBits(sam.MCLK_APBCMASK_TRNG_) {
 		// Turn on clock for TRNG
 		sam.MCLK.APBCMASK.SetBits(sam.MCLK_APBCMASK_TRNG_)
 
 		// enable
 		sam.TRNG.CTRLA.Set(sam.TRNG_CTRLA_ENABLE)
-
-		rngInitDone = true
+	}
+	for !sam.TRNG.INTFLAG.HasBits(sam.TRNG_INTFLAG_DATARDY) {
 	}
 	ret := sam.TRNG.DATA.Get()
 	return ret, nil


sago35:
I have modified the commit message.
Now ready for review.

sago35:
### INTFLAG

Bit 0 – DATARDY Data Ready
This flag is set when a new random value is generated, and an interrupt will be generated if INTENCLR/SET.DATARDY=1.
This flag is cleared by writing a '1' to the flag or by reading the DATA register.
Writing a '0' to this bit has no effect.

### MCLK
Bit 10 – TRNG TRNG APBC Mask Clock Enable

0 The APBC clock for the TRNG is stopped.
1 The APBC clock for the TRNG is enabled.

deadprogram:
> I believe crypto/rand needs to be modified because it needs to take a different value each time.

Can you please explain this comment?

sago35:
Without the change, the same value could be read even if rand.Read() was executed multiple times.
crypto/rand.Read should read a different value each time it is called.

deadprogram:
In other words, something like this?

```go
func (r *reader) Read(b []byte) (n int, err error) {
	if len(b) == 0 {
		return
	}

	var randomByte uint32
	for i := range b {
		randomByte, err = machine.GetRNG()
		if err != nil {
			return n, err
		}
		b[i] = byte(randomByte)
	}

	return len(b), nil
}
```

sago35:
> In other words, something like this?

No.
DataSheet says that 32-bit random can be got.

> Passed NIST Special Publication 800-22 Tests Suite 
> Passed Diehard Random Tests Suite 
> May be used as Entropy Source for seeding an NIST approved DRNG (Deterministic RNG) as required by FIPS PUB 140-2 and 140-3 
> Provides a 32-bit random number every 84 clock cycles

So, we can keep the original logic.

https://github.com/tinygo-org/tinygo/blob/3edbf7442b03a1154a62573137c0108c339feba8/src/crypto/rand/rand_baremetal.go#L17-L36

----

I sorry for not being able to explain well in English.
Running this source with Wio Terminal produced the following results.


```
package main

import (
	"fmt"
	"machine"
	"time"
)

func main() {
	time.Sleep(3 * time.Second)
	var result [4]uint32

	for {
		result[0], _ = machine.GetRNG()
		result[1], _ = machine.GetRNG()
		result[2], _ = machine.GetRNG()
		result[3], _ = machine.GetRNG()

		for _, x := range result {
			fmt.Printf("%08X\r\n", x)
		}
		time.Sleep(10 * time.Second)
	}
}
```


```
// dev branch (latest)
00000000        <--- Output before random value generation.
8C7C2280
8C7C2280        <--- The same value is output as before because it is output before the random value is generated.
8F9E4995
```

```
// samd51-trng branch
1B919561
F50BEE9C
5F479D66
D57CD0DA
```

deadprogram:
Sorry I think it was my misunderstanding as much as anything. Got it, that is some great work on this @sago35 thank you! Now merging.

aykevl (commented):
Good catch! This was a very serious bug.

Issue #3012, opened 2022-07-22
USB CDC not flashing RP2040 reliably


kenbell:
It seems like `tinygo flash` only works immediately after a hardware reset and not after a soft reset.

OS:
```sh
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04 LTS
Release:	22.04
Codename:	jammy
```

TinyGo:
```tinygo version 0.25.0-dev-3047d8f3 linux/amd64 (using go version go1.18.3 and LLVM version 14.0.0)```

Connections:
* RP2040 board (challenger RP2040) attached to USB (CDC maps to /dev/ttyACM0)

Scenario:
* Plug-in board to USB (observe `/dev/ttyACM0` present)
* Run `tinygo flash -target challenger-rp2040 ./src/examples/serial/` (observe flash succeeds)
* Run `tinygo flash -target challenger-rp2040 ./src/examples/serial/` (observe error `error: unable to locate a serial port`)

**Comment by kenbell** [comment]
(I have reproduced this issue on a vanilla Pico)

**Comment by sago35** [comment]
@kenbell 
I'll try it later.
Could you also try the following PR code?

#2991

**Comment by kenbell** [comment]
Thanks @sago35 - I've just tried that PR.  It's not helping with my scenario unfortunately.

**Comment by kenbell** [comment]
It might be related to that errata.  I've been having the rp2040 boards plugged into a USB hub with some other devices.  Using a USB port directly on my PC it's a lot more reliable.

**Comment by sago35** [comment]
#2991 code may not be well implemented yet.
Because RP2040-E5 is not reproduced in my environment.

**Comment by deadprogram** [comment]
Perhaps related to same issue as https://github.com/micropython/micropython/issues/8904

**Comment by sago35** [comment]
tinygo flash failed when USB-CDC exchange was executed from another microcontroller with the same USB-HUB.
To be clear, USB Enumeration has failed.

#2991 source will allow successful USB Enumeration.
Perhaps a workaround for errata is done.

However, USB enumeration of rp2040-bootloader almost always fails after portReset.
This is the same thing that happens when resetting while holding down Bootsel.
I think the bootloader on my B1 chip is not able to handle the RP2040-E5.

my feather-rp2040:
```
UF2 Bootloader v2.0
Model: Raspberry Pi RP2
Board-ID: RPI-RP2
```

**Comment by kenbell** [comment]
This is a dmesg from my system...
```
[...]

PR #3015, opened 2022-07-25
rp2040: add resetBlock for USBCTRL

sago35:
This PR adds a resetBlock for USBCTRL.
This is the same as initializing an Arduino.
I have confirmed that it works correctly with pico.

* https://github.com/tinygo-org/tinygo/issues/3012#issuecomment-1192958468
    * https://github.com/raspberrypi/pico-examples/blob/a7ad17156bf60842ee55c8f86cd39e9cd7427c1d/usb/device/dev_lowlevel/dev_lowlevel.c#L184-L186


deadprogram:
That makes total sense @sago35 good research! Now merging, thank you.

PR #3016, opened 2022-07-25
RP2040 fix USB device reset

### LINKED ISSUE #3012: USB CDC not flashing RP2040 reliably

kenbell:
This change makes the RP2040 USB much more reliable in the case where a bus fault occurs and the host tries to re-initialize the device.  In that case, the RP2040 should be expecting to respond to packets addressed to device '0'.

This fixes #3012 where a USB hub is failing to allow the initial USB setup packet handshake to complete in some cases (reason unknown).  When the host then tries to 'start-over', the RP2040 was failing to ACK USB packets (probably because it was expecting to respond to the original address).  This logic was borrowed from the Pico bootrom code, which always resets the device address on USB bus reset.

Before fix, bus can get in this state (notice, no ACK from RP2040 to packets on address 0):
image

After fix, although there are more USB resets that should be required, the bus does eventually get itself in a good state:
image

### DIFF
--- a/src/machine/machine_rp2040_usb.go
+++ b/src/machine/machine_rp2040_usb.go
@@ -128,6 +128,8 @@ func handleUSBIRQ(intr interrupt.Interrupt) {
 	// Bus is reset
 	if (status & rp.USBCTRL_REGS_INTS_BUS_RESET) > 0 {
 		rp.USBCTRL_REGS.SIE_STATUS.Set(rp.USBCTRL_REGS_SIE_STATUS_BUS_RESET)
+		rp.USBCTRL_REGS.ADDR_ENDP.Set(0)
+
 		initEndpoint(0, usb.ENDPOINT_TYPE_CONTROL)
 	}
 }


sago35 (approved):
Thanks for the great improvement.

deadprogram:
Thank you @kenbell for the fix and @sago35 for reviewing. Now merging.

PR #3017, opened 2022-07-25
target: add gopherbot and gopherbot2 aliases to simplify for newer users

deadprogram:
This PR adds `gopherbot` and `gopherbot2` aliases to simplify flashing process for newer users.

### DIFF
--- a/targets/gopherbot.json
+++ b/targets/gopherbot.json
@@ -1,0 +1,3 @@
+{
+	"inherits": ["circuitplay-express"]
+}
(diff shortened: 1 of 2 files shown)


kenbell:
Should update README as well?   (is Makefile smoketest also a good idea?)

deadprogram:
No need for smoketest with alias, it is already being tested.

As far as README, I am starting to wonder about the ever growing list of boards... not sure.

sago35 (approved):
🤖🤖🤖

PR #3018, opened 2022-07-25
esp32: fix WDT reset on the MCH2022 badge

aykevl:


sago35:
https://www.espressif.com/en/support/documents/technical-documents
https://www.espressif.com/sites/default/files/documentation/esp32_technical_reference_manual_en.pdf

```
// P.506
Both the MWDTs, as well as the RWDT, can be protected from accidental writing. To accomplish this, they have a
write-key register (TIMERS_WDT_WKEY for the MWDT, RTC_CNTL_WDT_WKEY for the RWDT.) On reset, these
registers are initialized to the value 0x50D83AA1. When the value in this register is changed from 0x50D83AA1,
write protection is enabled. Writes to any WDT register, including the feeding register (but excluding the write-key
register itself), are ignored. The recommended procedure for accessing a WDT is:
1. Disable the write protection
2. Make the required modification or feed the watchdog
3. Re-enable the write protection
```

sago35 (approved):
It might be better to run `Re-enable the write protection`.
However, it is probably not necessary to change it.

deadprogram:

aykevl:
@sago35 that would be helpful if the WDT is enabled. But in this case, we disable the WDT so I don't see a reason to enable protection.

PR #3021, opened 2022-07-26
esp32: optimize SPI transmit

aykevl:
For write-only operations (in SPI displays for example), the transmit speed is doubled with this relatively small change.


deadprogram:
See https://github.com/tinygo-org/tinygo/runs/7529106699?check_suite_focus=true#step:10:497 for build error

aykevl:
Oops, my Makefile was incorrect and wasn't recompiling the binary after I made some changes. The code should work now (and be even slightly faster!)

deadprogram:
Can you please rebase against `dev` now that the CI fix is in?

deadprogram (approved):

aykevl:
Rebased.

deadprogram:
Thanks @aykevl now merging.

PR #3023, opened 2022-07-27
usb/midi: add definition of MIDI note number

sago35:


src/machine/usb/midi/notes.go
```diff
@@ -0,0 +1,108 @@
+package midi
+
+// Note represents a MIDI note number. For example, Note(69) is A4 or 440Hz.
+type Note uint8
+
+// Define all the notes in a format similar to the Tone library in the Arduino
+// IDE.
+const (
+	A0 Note = iota + 21 // 27.5Hz
+	AS0
```
deadprogram, line 10: What do you think about using Unicode like `A♯0` for the sharps?

deadprogram: Also, in a separate section to not disturb the `iota` how about also putting in consts for the flats e.g. `B♭0 = A♯0`

sago35: Interesting idea.
However, for compatibility with Arduino, I would like to keep AS0 etc. and define A♯0 etc.

deadprogram: That makes sense. I was also thinking it might not be as easy for people to type, even though it would look nice to read.

sago35: It seems to be impossible to build.
I get the same error with Go.

https://go.dev/play/p/CDpt003Xhbg

```
./prog.go:4:3: invalid character U+266F '♯' in identifier
./prog.go:5:3: invalid character U+266D '♭' in identifier

Go build failed.
```

deadprogram: I think I misunderstood what was possible here. Never mind. :black_cat:

deadprogram (approved):

deadprogram:
Thank you @sago35 :musical_keyboard:

PR #3025, opened 2022-07-27
build/linux/arm: pin dotenv Rubygem version to 2.7.6

deadprogram:
This PR updates the Linux ARM builds to use the 2.7.6 version of the dotenv Rubygem. This is needed due to https://github.com/jordansissel/fpm/issues/1918

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -98,6 +98,7 @@ jobs:
         run: make wasi-libc
       - name: Install fpm
         run: |
+          gem install --version 2.7.6 dotenv
           gem install --no-document fpm
       - name: Build TinyGo release
         run: |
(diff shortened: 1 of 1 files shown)


deadprogram:
That last change is the correct one. Once everything else is completed, I will replace all these commits with a single one that does the needed pinning to specific version.

aykevl (approved):
Seems like a fine workaround.

Hopefully I'll be able to also make an Alpine/musl build for arm/arm64 so we can use a recent Perl version that hopefully doesn't need this workaround anymore.

deadprogram:
> use a recent Perl version

:rofl:

aykevl:
Eeeh yeah I clearly wasn't very focused...

PR #3027, opened 2022-07-27
machine/rp2040: turn off pullup/down when not input type not specified

deadprogram:
This PR turns off the pullup/pulldown resistor when the input type has not specified it.

This is needed to leave the input pin in a floating mode such as that needed by the Makeybutton driver.


kenbell (approved):
LGTM

Issue #3031, opened 2022-07-28
`symbol multiply defined!` linker error when multiple packages import a package with anonymous functions inside generic functions


pkedy:
As identified in #3003, in some special cases, a `symbol multiply defined!` linker error occurs when multiple packages reference a package that contains generic functions with anonymous functions inside. This would affect functional programming patterns use cases and perhaps others.

In #3003, an attempt was made to set the linkage for generic functions to `linkonce_odr` so that they are treated the same as C++ template instantiations, and duplicates are eliminated. This seemed to work on all architectures except for Windows. Also, more investigation is needed to identify how to properly identify generic SSA functions and set the linkage.

Example to reproduce (`testdata/generics/value/value.go`).

```golang
package value

type (
	Value[T any] interface {
		Get(Callback[T])
	}

	Callback[T any] func(T)

	Transform[S any, D any] func(S) D
)

func New[T any](v T) Value[T] {
	return &value[T]{
		v: v,
	}
}

type value[T any] struct {
	v T
}

func (v *value[T]) Get(fn Callback[T]) {
	// For example purposes.
	// Normally would be asynchronous callback.
	fn(v.v)
}

func Map[S, D any](v Value[S], tx Transform[S, D]) Value[D] {
	return &mapper[S, D]{
		v:  v,
		tx: tx,
	}
}

type mapper[S, D any] struct {
	v  Value[S]
	tx Transform[S, D]
}

func (m *mapper[S, D]) Get(fn Callback[D]) {
	m.v.Get(func(v S) {
		fn(m.tx(v))
	})
}
```

Reference from package A (`testdata/generics/testa/testa.go`)

```golang
package testa

import (
	"fmt"

	"github.com/tinygo-org/tinygo/testdata/generics/value"
)

func Test() {
	v := value.New(1)
	vm := value.Map(v, Plus100)
	vm.Get(func(v int) {
		fmt.Println("value:", v)
	})
}

// Plus100 is a `Transform` that adds 100 to `value`.
func Plus100(value int) int {
	return value + 100
}
```

Reference from package B (`testdata/generics/testb/testb.go`)

```golang
package testb

import (
	"fmt"

	"github.com/tinygo-org/tinygo/testdata/generics/value"
)

func Test() {
	v := value.New(1)
	vm := value.Map(v, Plus500)
	vm.Get(func(v int) {
		fmt.Println("value:", v)
	})
}

// Plus500 is a `Transform` that adds 500 to `value`.
[...]

PR #3032, opened 2022-07-28
Setting internal linkage and keeping default visibility for anonymous functions

pkedy:
Resolves issue #3031

/cc: @aykevl

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -1039,9 +1039,17 @@ func (b *builder) createFunctionStart() {
 		b.addError(b.fn.Pos(), errValue)
 		return
 	}
+
 	b.addStandardDefinedAttributes(b.llvmFn)
 	if !b.info.exported {
-		b.llvmFn.SetVisibility(llvm.HiddenVisibility)
+		// Do not set visibility for local linkage (internal or private).
+		// Otherwise a "local linkage requires default visibility"
+		// assertion error in llvm-project/llvm/include/llvm/IR/GlobalValue.h:236
+		// is thrown.
+		if b.llvmFn.Linkage() != llvm.InternalLinkage &&
+			b.llvmFn.Linkage() != llvm.PrivateLinkage {
+			b.llvmFn.SetVisibility(llvm.HiddenVisibility)
+		}
 		b.llvmFn.SetUnnamedAddr(true)
 	}
 	if b.info.section != "" {
@@ -1265,6 +1273,7 @@ func (b *builder) createFunction() {
 	// Create anonymous functions (closures etc.).
 	for _, sub := range b.fn.AnonFuncs {
 		b := newBuilder(b.compilerContext, b.Builder, sub)
+		b.llvmFn.SetLinkage(llvm.InternalLinkage)
 		b.createFunction()
 	}
 }
--- a/compiler/testdata/basic.ll
+++ b/compiler/testdata/basic.ll
@@ -196,7 +196,7 @@ entry:
 }
 
 ; Function Attrs: nounwind
-define hidden void @"main.foo$1"(%main.kv.0* dereferenceable_or_null(1) %b, i8* %context) unnamed_addr #1 {
+define internal void @"main.foo$1"(%main.kv.0* dereferenceable_or_null(1) %b, i8* %context) unnamed_addr #1 {
 entry:
   ret void
 }
--- a/compiler/testdata/defer-cortex-m-qemu.ll
+++ b/compiler/testdata/defer-cortex-m-qemu.ll
@@ -115,7 +115,7 @@ declare i8* @llvm.stacksave() #2
 declare void @runtime.setupDeferFrame(%runtime.deferFrame* dereferenceable_or_null(24), i8*, i8*) #0
 
 ; Function Attrs: nounwind
-define hidden void @"main.deferSimple$1"(i8* %context) unnamed_addr #1 {
+define internal void @"main.deferSimple$1"(i8* %context) unnamed_addr #1 {
 entry:
   call void @runtime.printint32(i32 3, i8* undef) #3
   ret void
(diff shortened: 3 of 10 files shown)


pkedy:
@aykevl Looks like I found the issue with the generics test failure under Windows. The test originally had `println` and I added 2 occurrences of `fmt.Println`. The two `println` should output before the two `fmt.Println`, but on Windows that is not the case. They output out the other way around. This tells me that TinyGo is handling stdout with Wasm differently on Windows compared to other OSes. It’s a separate issue but wanted to shed light on my findings.

To fix the test, I made replaced the `println` with `fmt.Println`. Now we just need to find a better way to detect generic functions without looking at the symbol name. This is where I could use your help. Worst case scenario, the logic I added seems to work.

Also, could you please retrigger the tests? Looks like the macOS encountered a hiccup.

aykevl:
There is one other way to detect generic functions: by looking at the `Synthetic` field (it starts with something like "instantiation of ..."). I don't really like that method either but I think it's a more reliable method than looking at the symbol name.

aykevl:
> To fix the test, I made replaced the `println` with `fmt.Println`. Now we just need to find a better way to detect generic functions without looking at the symbol name. This is where I could use your help. Worst case scenario, the logic I added seems to work.

While that fixes the problem, I'd much prefer if you'd use `println` everywhere instead of `fmt.Println` because that reduces the compile time for tests. The `fmt` package is rather large (and imports large packages).

pkedy:
> I'd much prefer if you'd use println everywhere instead

✅  I switched to `println`.  

> by looking at the Synthetic field (it starts with something like "instantiation of ...").

I tried this and see that in some cases `Synthetic` has that prefix, however it is empty for in other functions that are detected by the symbol name and thus produces the linker error.

```
error: Linking globals named '(*github.com/tinygo-org/tinygo/testdata/generics/value.mapper[int, int]).Get[int int]$1': symbol multiply defined!
```

pkedy:
@aykevl I tried this and it works and applies to fewer functions:

```golang
if strings.HasSuffix(name, "]$1") {
	b.llvmFn.SetLinkage(llvm.LinkOnceODRLinkage)
}
```

What is the significance of the `$1` suffix? Is this kind of function something we can detect instead?

pkedy:
The `$1` seems to represent the anonymous function inside the generic function. I was able to produce an error for `$2` by adding a second anonymous func.

```golang
func (m *mapper[S, D]) Get(fn1 Callback[D], fn2 Callback[D]) {
	m.v.Get(func(v S) {
		fn1(m.tx(v))
	}, func(v S) {
		fn2(m.tx(v))
	})
}
```

```
error: Linking globals named '(*github.com/tinygo-org/tinygo/testdata/generics/value.mapper[int, int]).Get[int int]$2': symbol multiply defined!
```

pkedy:
@aykevl It looks like this logic might be headed in the right direction.

```golang
// Set linkage to `linkonce_odr` for anonymous functions inside generic functions.
parent := b.fn.Parent()
insideGenericFunc := false
ancestor := parent
// Search ancestors for generic functions.
for ancestor != nil {
	if strings.HasPrefix(ancestor.Synthetic, "instantiation of ") {
		insideGenericFunc = true
		break
	}
	ancestor = ancestor.Parent()
}
// Test if this function is an anonymous function and set `linkonce_odr`.
if insideGenericFunc {
	for _, anonFunc := range parent.AnonFuncs {
		if anonFunc == b.fn {
			b.llvmFn.SetLinkage(llvm.LinkOnceODRLinkage)
			break
		}
	}
}
```

aykevl:
Yes, those are indeed anonymous functions.
I'm now thinking we might want to use internal linkage for _all_ anonymous functions, because they're only called from one function anyway. Internal linkage means that this function does not appear in the symbol table and will be renamed if it is defined more than once (and I believe the duplicates should normally be discarded during optimization).

Also, it might be worth checking what happens if you use an anonymous function inside an anonymous function, it might have a `$1$1` suffix.

pkedy:
Ok. I’ll try that out. I’ve already updated the test the have anonymous functions inside anonymous functions and that works in the logic in my above comment.

aykevl:
Could be as simple as this:

```go
index 1b588d9c..b8ae01f8 100644
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -1265,6 +1265,7 @@ func (b *builder) createFunction() {
        // Create anonymous functions (closures etc.).
        for _, sub := range b.fn.AnonFuncs {
                b := newBuilder(b.compilerContext, b.Builder, sub)
+               b.llvmFn.SetLinkage(llvm.LinkOnceODRLinkage)
                b.createFunction()
        }
 }
```

(Here I've used `linkonce_odr`, it probably doesn't matter but is more consistent with https://github.com/tinygo-org/tinygo/blob/release/compiler/symbol.go#L199).

aykevl:
Note that when you do this, you need to run `go test ./compiler -update`: a number of compiler tests will change with this change.

pkedy:
I'm happy to report your one line fix works for the tests I added and in my project with the actual use case!

pkedy:
I committed the fix. Many of the `.ll` files in testdata were updated. Please take a look and see if that is expected. Also the `TestAllocs2` test now fails. Looks like this effected memory escaping. Thoughts?

aykevl (approved):
The code looks good to me.
I do need to look into why the Allocs2 test now fails.

aykevl:
Apparently we should use internal linkage, not `linkonce_odr`. Not sure why though, but with `linkonce_odr` many function attributes are not inferred (including the very important `nocapture`).

pkedy:
`TestAllocs2` does not seem to fail using internal linked. I ran `go test ./compiler -update` to update the expected output and tests are passing for me locally. I'm going to go ahead and make this PR ready for review.

pkedy:
I had to add a check to not set visibility to hidden when the linkage is "local" (internal or private)
otherwise the assertion in `llvm-project/llvm/include/llvm/IR/GlobalValue.h:236` fails:

```c++
  void setVisibility(VisibilityTypes V) {
    assert((!hasLocalLinkage() || V == DefaultVisibility) &&
           "local linkage requires default visibility");
    Visibility = V;
    if (isImplicitDSOLocal())
      setDSOLocal(true);
  }
```

```c++
  bool hasLocalLinkage() const { return isLocalLinkage(getLinkage()); }

  static bool isLocalLinkage(LinkageTypes Linkage) {
    return isInternalLinkage(Linkage) || isPrivateLinkage(Linkage);
  }
```

pkedy:
@aykevl Everything is looking good now. What do you think?

Its great that we went from crazy hack to simple tweak. Thanks for all the helpful hints. Never would have gotten to the solution without them!

Also, please please please can this be released in 0.25.0? :innocent:

deadprogram:
Thank you for working on this @pkedy now merging.

PR #3036, opened 2022-07-30
wasm,wasi: stub runtime.buffered, runtime.getchar

sago35:
This PR adds runtime.buffered and runtime.getchar stubs to the wasm/wasi target.
Supported os.Stdin in #2625, but forgot to support the wasm/wasi target because it did not produce a link error.

With the code I am adding, it is not possible to use os.Stdin in wasm/wasi, but the error will no longer occur.
Note that os.Stdin is not supported by wasm/wasi from the original.

### DIFF
--- a/src/runtime/runtime_tinygowasm.go
+++ b/src/runtime/runtime_tinygowasm.go
@@ -49,6 +49,16 @@ func putchar(c byte) {
 	}
 }
 
+func getchar() byte {
+	// dummy, TODO
+	return 0
+}
+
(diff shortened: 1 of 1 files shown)


sago35:
When he tried to run it on Cloudflare, he seemed to get the following error.

image


https://twitter.com/__syumai/status/1553266252847919106?s=20&t=R42Dr81a0gelLtOJSwk8uw

aykevl (approved):
Seems good to me!

PR #3037, opened 2022-07-31
rp2040: support dormant mode

kenbell:
With this, I can get sub 1mA draw during idle (866uA) on a Challenger RP2040 LoRa board (I have to manually put the embedded LoRa chip into deep sleep).

The machine API is not final - it assumes Pin wake, when RP2040 also supports RTC-wake from dormant.

### DIFF
--- a/src/machine/machine_rp2040.go
+++ b/src/machine/machine_rp2040.go
@@ -97,3 +97,22 @@ func CurrentCore() int {
 
 // NumCores returns number of cores available on the device.
 func NumCores() int { return 2 }
+
+func Sleep(pin Pin, change PinChange) error {
+	if pin > 31 {
+		return ErrInvalidInputPin
+	}
+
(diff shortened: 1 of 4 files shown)


src/machine/machine_rp2040.go
```diff
@@ -97,3 +97,22 @@ func CurrentCore() int {
 
 // NumCores returns number of cores available on the device.
 func NumCores() int { return 2 }
+
```
headblockhead, line 100: Function does not have clear usage description.

headblockhead:
Hello! I am also working on a project using tinygo involving the Challenger RP2040 lora device (it is very work-in-progress and incomplete but [you can find it here](https://github.com/headblockhead/picoDoomsdayMessenger)) and I would greatly appreciate it if you could mark this pull request as ready for review so I can use it in my project! :smile:

PR #3038, opened 2022-08-01
examples/wasm: improve Makefile

sago35:
A small improvement.
It works well even if we copy examples/wasm to a different location.
Also, the version of wasm_exec.js will not be mistaken.


deadprogram:
Seems alright to me, anyone who can think of a reason to not do this?

sago35:
I wanted to build examples/wasm below, but the original makefile did not work.
So I would like to see this change applied.

https://github.com/sago35/tinygo-devcontainer

dgryski (approved):
LGTM

PR #3039, opened 2022-08-02
all: remove support for LLVM 13

aykevl:
LLVM 13 and LLVM 15 can't be supported at the same time. In addition, transitioning to LLVM 15 will be a lot of work (because of opaque pointers), so we better start working on it soon. Therefore, to start with LLVM 15 support I'd like to remove LLVM 13 support now.

To be merged _after_ the coming release, so that LLVM 13 will remain supported in version 0.25.0.


dgryski (approved):
LGTM (now that 0.25 is out)

deadprogram:
:rocket: :ringed_planet: merging

PR #3043, opened 2022-08-03
all: format code according to Go 1.19 rules

aykevl:
Go 1.19 started reformatting code in a way that makes it more obvious how it will be rendered on pkg.go.dev. It gets it almost right, but not entirely. Therefore, I had to modify some of the comments so that they are formatted correctly.


dgryski (approved):
LGTM

sago35 (approved):
LGTM.
It is difficult to develop without incorporating this change first.
Every time I do a go fmt, I get a diff.

PR #3045, opened 2022-08-03
ci: add check that TinyGo can be built using Homebrew LLVM

aykevl:
This is supposed to work, but there was no CI check. Therefore, add this check.

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -100,3 +100,28 @@ jobs:
       - name: Smoke tests
         shell: bash
         run: make smoketest TINYGO=$(PWD)/build/tinygo AVR=0
+  test-macos-homebrew:
+    name: homebrew-install
+    runs-on: macos-latest
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.18'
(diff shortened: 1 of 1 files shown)


.github/workflows/build-macos.yml
```diff
@@ -100,3 +100,28 @@ jobs:
       - name: Smoke tests
         shell: bash
         run: make smoketest TINYGO=$(PWD)/build/tinygo AVR=0
+  test-macos-homebrew:
+    name: Test on MacOS with Homebrew
```
deadprogram: Perhaps change this to `homebrew-install` to match the format for the names of all the other workflows?

deadprogram:
I wonder if this should only run when merged into `dev` and `release` branches, to avoid adding yet more time to the CI process?

aykevl:
> Perhaps change this to `homebrew-install` to match the format for the names of all the other workflows?

I couldn't really come up with a good name. `homebrew-install` works but also doesn't really say what it does. Anyway, no strong opinions from me here.

> to avoid adding yet more time to the CI process

Do you mean CI minute usage or time until PRs are finished testing? Because this check should run in parallel with other checks and finish much quicker than others.

deadprogram (approved):

PR #3049, opened 2022-08-04
make interp pass timeout configurable

dgryski:


dgryski:
CI fails because new interp timeout is set to 0 because it's default zero value in the Options struct when running tests from `main_test.go`.

dkegel-fastly:
For https://github.com/tinygo-org/tinygo/issues/2090

aykevl (approved):

deadprogram:
Thank you @dgryski for the enhancement! Now merging.

PR #3050, opened 2022-08-05
syscall: ensure correct C prototype WASI function signature

rvolosatovs:
It's `<type> <name>`, rather than `<name> <type>` in C

Extracted from #2748


deadprogram:
Thanks for the correction @rvolosatovs now merging.

PR #3051, opened 2022-08-05
net: sync net.go with Go 1.18 stdlib

rvolosatovs:
Extracted from https://github.com/tinygo-org/tinygo/pull/2748


deadprogram (approved):
Looks like just comments updated, but still a good thing.

PR #3055, opened 2022-08-05
static analysis fixes

dgryski:


deadprogram (approved):
Thank you for organizing these commits for review!

PR #3058, opened 2022-08-06
nrf52: s140v6 uf2 targets cleanup

ysoldak:
Many targets use "circuitplay-bluefruit.ld" that actually is linker file for uf2 bootloader with s140v6 softdevice on nrf52840.

- Renamed "targets/circuitplay-bluefruit.ld" to "targets/nrf52840-s140v6-uf2.ld"
- Cleaned up unnecessary duplicated lines in targets that use this linker script
- Added "nano-33-ble-s140v6-uf2" target ([Adafruit bootloader](https://github.com/adafruit/Adafruit_nRF52_Bootloader) shall be flashed manually on this board though)


deadprogram:
Thank you for the cleanup! Merging.

PR #3059, opened 2022-08-06
nrf52: s140v7 uf2 targets cleanup

ysoldak:
XIAO BLE's linker file is actually recipe for uf2 bootloader with s140v7 soft device on nrf52840.

- Renamed "targets/xiao-ble.ld" to "targets/nrf52840-s140v7-uf2.ld"
- Added "nano-33-ble-s140v7-uf2" target ([Adafruit bootloader](https://github.com/adafruit/Adafruit_nRF52_Bootloader) shall be flashed manually on this board though)

See also: #3058


deadprogram:
Thank you @ysoldak for more cleanup.

Issue #3057, opened 2022-08-06
tinygo-test fails with undefined symbols


QuLogic:
With Go 1.18, I am getting the following error when running `make tinygo-test`:
```
ld.lld: error: undefined symbol: getpagesize
>>> referenced by os_linux.go:122 (/builddir/build/BUILDROOT/tinygo-0.25.0-1.fc36.x86_64/usr/lib64/tinygo/src/runtime/os_linux.go:122)
>>>               /tmp/tinygo863360999/main.o:(os_test.TestGetpagesize)
error: failed to link /tmp/tinygo863360999/main: exit status 1
```
With Go 1.19, I additionally get the following error on any test that uses crypto:
```
ld.lld: error: undefined symbol: crypto/internal/boring/sig.StandardCrypto
>>> referenced by notboring.go:25 (/usr/lib/golang/src/crypto/internal/boring/notboring.go:25)
>>>               /builddir/build/BUILD/tinygo-0.25.0/tinygo.OFH7PV/tinygo/thinlto/llvmcache-51407FBE491F17C7777BEE75E99D1816E029D4F3:(crypto/internal/boring.Unreachable)
error: failed to link /tmp/tinygo3462869915/main: exit status 1
```

This started with tinygo 0.24.0. It seemed to be related to CGo, so I tried reverting #2774, but that did not help. I also tried force disabling ThinLTO, but that did not help either. So I'm running low on ideas. I am running with external LLVM (14.0.5 against Go 1.19, and 14.0.0 against Go 1.18)

QuLogic:
Though that being said, 0.23.0 never compiled with Go 1.19, so I can't say if it wouldn't have brought up the same errors. Similarly, the test for `getpagesize` was only added in 0.24.0, so I can't say it also wouldn't have errored in earlier versions.

So I don't know if this actually regressed from 0.23.0 to 0.24.0 or there are just more tests that show it was already broken.

QuLogic:
So I backported #2899 to 0.23.0 and it still failed, so this seems to have always been broken and there just wasn't anything that triggered it?

dkegel-fastly:
Could it be that getpagesize() is no longer available everywhere?  
Evidently it is deprecated, see e.g. https://refspecs.linuxbase.org/LSB_4.1.0/LSB-Core-generic/LSB-Core-generic/baselib-getpagesize.html

aykevl:
[...]

PR #3060, opened 2022-08-06
Improve Go 1.19 support

QuLogic:
Fixes some issues running `make tinygo-test` against Go 1.19. It doesn't pass for me though because of #3057.

### DIFF
--- a/src/os/exec.go
+++ b/src/os/exec.go
@@ -1,6 +1,9 @@
 package os
 
-import "syscall"
+import (
+	"errors"
+	"syscall"
+)
 
 type Signal interface {
 	String() string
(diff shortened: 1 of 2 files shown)


src/os/exec.go
```diff
@@ -24,6 +27,9 @@ type ProcAttr struct {
 	Sys   *syscall.SysProcAttr
 }
 
+// ErrProcessDone indicates a Process has finished.
+var ErrProcessDone = errors.New("os: process already finished")
```
dgryski, line 31: Needed to patch this in by hand today.  :+1:

dgryski (approved):
LGTM

deadprogram:
Thank you for the improvement @QuLogic and to @dgryski for the review.

Now merging.

PR #3061, opened 2022-08-07
syscall: add WASI `sock_accept`

rvolosatovs:
Extracted from #2748 
This PR adds functionality to be able to call the syscall directly. #2748 uses this syscall to implement `net.Listener` for WASI

Note that this calls into WASI directly without using libc. Note, also that this syscall is similar to, but different from standard Linux `accept{,4}`

Refs https://github.com/WebAssembly/WASI/pull/458

### DIFF
--- a/src/syscall/syscall_libc_wasi.go
+++ b/src/syscall/syscall_libc_wasi.go
@@ -309,6 +309,14 @@ func Getpagesize() int {
 	return 65536
 }
 
+func SockAccept(fd int, flags int) (nfd int, err error) {
+	var retptr0 __wasi_fd_t
+	if n := sock_accept(__wasi_fd_t(fd), __wasi_fdflags_t(flags), &retptr0); n != 0 {
+		return -1, Errno(n)
+	}
+	return int(retptr0), nil
+}
+
 // int stat(const char *path, struct stat * buf);
 //
 //export stat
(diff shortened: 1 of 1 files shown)


rvolosatovs:
@aykevl @dgryski any update on this PR? Are there any concerns regarding the feature?

Is there any way I can assist in getting this reviewed?

dgryski:
I'll try to take a look this week.

deadprogram:
@dgryski any chance to look at this? :smile_cat:

codefromthecrypt:
maybe make a code comment about it being different than accept/4?

you could also cite the reference impl which seems to only pass only two values as flags: 0 or non block https://github.com/WebAssembly/wasi-libc/commit/63e4489d01ad0262d995c6d9a5f1a1bab719c917#diff-07a5fd5db4b7b518204f053b8fa5f30506522395b61764e5d804931346335725

aykevl:
To be honest, I think it's better to do this directly in the net package. The syscall package is deprecated and frozen upstream, and this adds some unnecessary API surface that we will have to maintain.

rvolosatovs:
> To be honest, I think it's better to do this directly in the net package. The syscall package is deprecated and frozen upstream, and this adds some unnecessary API surface that we will have to maintain.

Closing, will rework using suggested approach (in #2748)

PR #3063, opened 2022-08-08
Fix skip message for missing emulators

QuLogic:
`spec.Emulator` is a string of the command, not an array, so picking the first element just prints the first letter.


deadprogram (approved):

deadprogram:
Thanks for the fix @QuLogic

PR #3069, opened 2022-08-12
goenv: support GOOS=android

aykevl:
TinyGo doesn't currently support Android directly. However, GOOS=linux
works fine on Android. Therefore, force GOOS=linux on Android.

Tested in Termux.


dgryski (approved):
LGTM

PR #3072, opened 2022-08-15
feat:add new target to support wasm freestanding

### LINKED ISSUE #3068: add new target  like wasm32-unknown-unknown in rust
## Simple Summary

Add new wasm32-unkonw-unkonw/unknow_wasm target in TinyGo. Use this target to generate clean wasm which have no dependecies on JS or WASI.

## Abstract

In the current implementation, WASM target depends heavily on the JS execution environment, and the WASI target requires the corresponding execution platform to suppoprt the POSIX protocol or something like POSIX.

When demands come to generate an algorithm to run on a specific WASM VM, there is usually no JS execution environment and no posix support in these vm environment. For example, if one would like to run the WASM code generated by TinyGo, we have to introduce some JS/WASI compatibility layer code. However, when the execution environment of the VM is developed by a third party and not within control, the developer then cannot run the compatibility layer code. At this time, it is forced to use the WASI target which translates to mannual changes of many modules in its project causing much inconvenience and infeasibilities in some cases. 

A functional target similar to Rust's wasm32-unknown-unknown is needed to generate clean WASM without changing the existing library code.

## Change Motivation

When the WASM file is to be run in a foreign VM environment and this target relies heavily on the JS running environment, it becomes impossible to use the WASM target. When the target is WASI, the generated code will have dependency on the WASI protocol resulting in having some import that depends on the WASI protocol. Compared with the WASM target, there is a possibility of realizing the implementation which developer can make mannual changes to the code (user module, standard package) to remove the imports on WASI. However, this approach implies large amount of engineering efforts.

For the general user library code（include dependecy), there are various parts that use POSIX functions such as (file/os/net/time) etc, which can be patched by cherry picking the corresponding module, modifying the module, removing the code that have WASI dependency and then replacing them with edited modules throughout your project. Through this method, we can achieve the goal of running targets, but it will be very cumbersome. In practice, due to the depth of dependencies, it will be difficult to find the module with problems. In addition, replacement will also bring side effects, which will require the replacement of other libraries that references the edited library.

For the standard library, it is even more cumbersome. In practice, `fmt` standard library has been one of the most serious problems. `fmt` use the `stdout` and `stdin` objects, which means that the fmt standard library depends on the `os` standard library that comes with an init function...(shown below)

```go
// Args hold the command-line arguments, starting with the program name.
var Args []string

func init() {
	Args = runtime_args()
}

```

Not only does `args_size` invokes dependency on WASI, but ```runtime_args``` function further includes...
```
  (import "wasi_snapshot_preview1" "args_sizes_get" (func $runtime.args_sizes_get (;0;) (type 3)))
  (import "wasi_snapshot_preview1" "args_get" (func $runtime.args_get (;1;) (type 3)))
  (import "wasi_snapshot_preview1" "clock_time_get" (func $runtime.clock_time_get (;2;) (type 6)))
```

Unfortunately the fmt library is widely used (Eg: ```fmt.Errorf```), which brings great troubles,
 
* Standard library cannot be replaced
* There are too many user modules which use fmt package, making not realistic to replace all of them.

Therefore, we hope to refer to Rust's wasm32-unkonow-unkonwn, and use plateform specific code to handle these problems (io/file/net) when specifying the directive.

## Implementation

1. Similar to WASI, add a new directive ```wasm_unknow``` to isolate the specific implementation to make sure previous WASM/WASI behaviour stays the same.
2. Add new target type ```wasm_unknow``` to generate cleaner WASM files through this kind of utility target.

The main idea is to refer to the implementation of rust wasm32-unknown-unknown https://github.com/rust-lang/rust/tree/master/library/std/src/sys/unsupported.

packages to be changed 
1. src/os 
2. src/syscall
3. src/runtime  add new target support

behavior change
1. os.args/os.env        always return empty slice
2. file/net                     not support ,
3. stdout.                     discard
4. stdin                         return io.EOF
5. time                          flow rust panic behavire (many people use time as rand seed, autoincrease nano time cause predicatable randomness
6. syscall                      use xx_emulated.go, and no Signal support
7. Signal                       remove signal      (Signal always used in main file, less used in library, so remote it)
## Maintainer

I am also very happy to maintain this target, and I also have a team member who can invest in the development of this target.

@dankegel  @deadprogram @aykevl

hunjixin:
resolve https://github.com/tinygo-org/tinygo/issues/3068


hunjixin:
@deadprogram plan to use this feature?

dgryski:
I want to bikeshed the tag.  `unknow_wasm` should probaby be `unknown_wasm` at least (to fix the spelling error), but I wonder if we want something like `barewasm` (similar to `baremetal`) to indicate exactly what we're building for?

hunjixin:
@dgryski  barewasm seems a reasonable tag name.  I think developers are very much in need of https://github.com/tinygo-org/tinygo/issues/3068. Although WASM was originally designed for browser front-end development, new use cases are being raised for VMs in the blockchain dev communities, and various cross-platform algorithm implementations. One of the most important things is that WASM can take good advantage of the existing technical capabilities of various languages. 
This PR aims to better support the Go standard library and the existing Go code on github to be ported to wasm with minimal changes. In fact, Rust achieves this goal through a similar approach (like this PR). As a result, A lot of developers use this feature in RUST to build the WASM projects. If TinyGo can merge this approach, I believe it will help the TinyGo project to have an even larger adoption rate in regard to WASM projects.

codefromthecrypt:
for the bikeshed I think 'freestanding' is in use for wasm for something similar (maybe zig?) would be cool to use common conventions esp if there aren't strong feelings about it. @mathetake do you have any opinion?

mathetake:
rust: unknown, zig: freestanding, so either works for me

hunjixin:
> for the bikeshed I think 'freestanding' is in use for wasm for something similar (maybe zig?) would be cool to use common conventions esp if there aren't strong feelings about it. @mathetake do you have any opinion?

its better to use freestanding than unknown. @dgryski  any better idea
```
zig cc -target wasm32-freestanding test.c -o test -lc
```

hunjixin:
@codefromthecrypt please help to retry the ci?

deadprogram:
From the looks of those CI errors, someone needs to rebase this branch against the latest `dev`

RubixDev:
How far is the progress on this? I would really like to use this to make an [shfmt](https://github.com/mvdan/sh) plugin for [dprint](https://dprint.dev/).

src/os/stat_linux.go
```diff
@@ -1,5 +1,5 @@
-//go:build linux && !baremetal
-// +build linux,!baremetal
+//go:build linux && !baremetal && !unknow_wasm
+// +build linux,!baremetal,!unknow_wasm
```
RubixDev: Is the `unknow_wasm` here correct, or should it also be `wasm_freestanding`?

hunjixin: yes, a bug here

hunjixin:
> How far is the progress on this? I would really like to use this to make an [shfmt](https://github.com/mvdan/sh) plugin for [dprint](https://dprint.dev/).

dprint？this pr remove all print support

RubixDev:
dprint is a code formatter written in Rust that supports plugins in bare wasm. The only thing These wasm files have to do is export some functions that dprint will then call, but it does not provide WASI or emscripten or any other interface. The dprint wasm plugins written in Rust all target `wasm32-unknown-unknown`. It has nothing to do with printing.

hunjixin:
> dprint is a code formatter written in Rust that supports plugins in bare wasm. The only thing These wasm files have to do is export some functions that dprint will then call, but it does not provide WASI or emscripten or any other interface. The dprint wasm plugins written in Rust all target `wasm32-unknown-unknown`. It has nothing to do with printing.

so you are planning to use  golang formatter to generate wasm plugin?

RubixDev:
Yes. I want to create a wasm binary including the shfmt shell formatter that is written in Go. Thus the plugin itself must also be written in Go, as it depends on shfmt.

But back to the original question, how far is the progress on this? Is it already somehow usable? I am sure this target is what I need for my use case.

hunjixin:
> Yes. I want to create a wasm binary including the shfmt shell formatter that is written in Go. Thus the plugin itself must also be written in Go, as it depends on shfmt.
> 
> But back to the original question, how far is the progress on this? Is it already somehow usable? I am sure this target is what I need for my use case.

there are still some behaviors that cannot be determined which method is better, especially the time part,should be panic or follow behavior suggested by @codefromthecrypt. If the current demand is strong, you can also consider this idea https://github.com/ipfs-force-community/go_tinygo_patch

for using increasing value for time return, this could make ramdom worked, but increase risk because randomness can be predicted.

src/os/stat_linux.go
hunjixin: fixed

jsimnz:
Any update on this? Happy to contribute if needed, or are you still trying to establish behavior?

hunjixin:
> Any update on this? Happy to contribute if needed, or are you still trying to establish behavior?

This pr should work normally, but I don't know how to make tinygo accept this target. The behavioral aspect tends to keep the existing implemented behavior. There is a security risk in using self increasing time. development usually use time.Now as a random seed. especially in the financial smart contract.

codefromthecrypt:
We're nearing a year since @hunjixin raised this. The last question was "how to make tinygo accept this target"

Could someone lay out what is required to get this in, sounds like getting the build to pass is a given. What else can be checked off?

deadprogram:
Valid point @codefromthecrypt 

@dgryski @aykevl what are your thoughts on this? Also while keeping in mind upcoming changes for Go1.21+

dajuguan:
when I build it ` tinygo build -o main.wasm -target=unknow_wasm main.go`, and run it with wasmer. The following error shows:
```
error: Unable to instantiate the WebAssembly module
╰─▶ 1: Error while importing "env"."tinygo_getCurrentStackPointer": unknown import. Expected Function(FunctionType { params: [], results: [I32] })
```

deadprogram:
I would like to have this sort of functionality myself to be able to run TinyGo compiled code on WAMR.

I think since the naming of `wasm32-unknown-unknown` is the most standard, so probably  `-target=wasm32-unknown` and build tag `wasm32_unknown` would be the best option since there are more people doing this sort of thing with WASM in C or Rust at the moment.

Regarding getting this change accepted, there needs to be some tests showing that it works as expected. Some tests similar to the wasi target is the most likely place to look for inspiration.

Thanks for working on this, and let me know how I can help out.

deadprogram:
Now that #4125 has been merged, I am closing this one. Thanks again everyone!

PR #3074, opened 2022-08-15
Add aliases for ed25519 field.feMul and field.feSquare.

abrender:
ed25519vectors_test.go still fails because:
* It relies on "go mod download" which doesn't work, as well as fork/exec.
* It relies on JSON parsing which has problems with reflection.

But, with the vectors hard coded in the test file the tests *do* succeed, so the encryption is working.

### DIFF
--- a/compiler/alias.go
+++ b/compiler/alias.go
@@ -16,6 +16,8 @@ import "tinygo.org/x/go-llvm"
 
 var stdlibAliases = map[string]string{
 	// crypto packages
+	"crypto/ed25519/internal/edwards25519/field.feMul":    "crypto/ed25519/internal/edwards25519/field.feMulGeneric",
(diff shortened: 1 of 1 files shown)


dgryski:
This is useful.  I'm pretty sure there are a few other routines in the crypto packages that could use some aliasing.

abrender:
> This is useful. I'm pretty sure there are a few other routines in the crypto packages that could use some aliasing.

I'm unsure if the failing windows build is due to this or just a fluke.  I'm unable to rerun the workflow - any suggestions? Thanks!

dgryski:
@abrender Restarted the windows build for you.

abrender:
> @abrender Restarted the windows build for you.

@dgryski - thanks! All green now :)

dgryski (approved):
LGTM

deadprogram:
Thanks for the contribution @abrender and thanks for reviewing @dgryski now merging.

PR #3080, opened 2022-08-18
loader,crypto: fix link error for crypto/internal/boring/sig.Standard…

dgryski:
…Crypto

### DIFF
--- a/src/crypto/internal/boring/sig/sig_other.go
+++ b/src/crypto/internal/boring/sig/sig_other.go
@@ -1,0 +1,17 @@
+// Package sig stubs crypto/internal/boring/sig
+package sig
+
+// BoringCrypto indicates that the BoringCrypto module is present.
+func BoringCrypto() {
+
+}
+
+// FIPSOnly indicates that package crypto/tls/fipsonly is present.
+func FIPSOnly() {
+
+}
+
+// StandardCrypto indicates that standard Go crypto is present.
+func StandardCrypto() {
+
+}
--- a/loader/goroot.go
+++ b/loader/goroot.go
@@ -46,7 +46,7 @@ func GetCachedGoroot(config *compileopts.Config) (string, error) {
 	}
 
 	// Find the overrides needed for the goroot.
-	overrides := pathsToOverride(needsSyscallPackage(config.BuildTags()))
+	overrides := pathsToOverride(config.GoMinorVersion, needsSyscallPackage(config.BuildTags()))
 
 	// Resolve the merge links within the goroot.
 	merge, err := listGorootMergeLinks(goroot, tinygoroot, overrides)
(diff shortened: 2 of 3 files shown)


src/crypto/internal/boring/sig/sig_other.go
```diff
@@ -0,0 +1,17 @@
+// Package sig stubs internal/boringcrypto/sig
```
joeshaw: Nit: should be `crypto/internal/boring/sig`

src/crypto/internal/boring/sig/sig_other.go
```diff
@@ -0,0 +1,17 @@
...
+package sig
+
+// BoringCrypto indicates that the BoringCrypto module is present.
+func BoringCrypto() {
+
```
joeshaw, line 6: Should we panic here and in `FIPSOnly`?

joeshaw (commented):
👍

src/crypto/internal/boring/sig/sig_other.go
dgryski: These functions aren't called; they're here so tooling can look for the body of the symbols and then figure out stuff about the binary: https://github.com/golang/go/blob/master/src/crypto/internal/boring/sig/sig_amd64.s#L7

(edit: ok, they're called, but have effectively empty bodies)

src/crypto/internal/boring/sig/sig_other.go
dgryski: Fixed and repushed.

deadprogram:
@dgryski does this branch perhaps need to be rebased against latest `dev`?

dgryski:
Redid the Mac job; passed this time.

dgryski:
Rebased anyway.

aykevl:
While this is a fix, and it seems fine by me, this issue will also be fixed by #3103.

dgryski:
I'm happy to close this in favour of the more general fix.

joeshaw:
@dgryski The caveats in #3103 (currently Linux only, depending on a few other PRs to land first) make me think that this PR would be beneficial to land as a stopgap to prevent the regression in crypto package usage from 1.18 to 1.19 caused by this issue unless that other PR is likely to land for the next release.

dgryski:
I suppose we could merge this in the mean time and revert it when the full solution lands.

aykevl (approved):
> I suppose we could merge this in the mean time and revert it when the full solution lands.

Seems fine by me.

src/crypto/internal/boring/sig/sig_other.go
aykevl: Yeah, this is what made me hesitate a bit with this patch. It breaks these signatures. But apparently it's only used for amd64 which makes me a bit less nervous about it (and amd64 isn't even the primary target of TinyGo).

deadprogram:
Thank you @dgryski for the fix and @aykevl for review.

Now merging.

PR #3084, opened 2022-08-20
build: pin public_suffix Ruby gem to version 4.0.7

deadprogram:
This PR pins the `public_suffix` Ruby gem to version 4.0.7 to avoid needing to update Ruby version for build.

Needed to fix the CI build on `dev` branch due to upstream changes in dependencies.


deadprogram:
Merging to get back to green.

PR #3089, opened 2022-08-23
Draft: improve CI test for timers

kenbell:


deadprogram:
@kenbell looks like this is passing CI. What do you think?

kenbell:
@deadprogram I've re-triggered MacOS.  If that works, that will be 3 in a row that worked, so probably good to merge.

deadprogram:
Thank you very much for the quick fix @kenbell now merging.

PR #3092, opened 2022-08-24
main: ignore ports with VID/PID if not candidates

sago35:
Fall-back using `secondaryPorts` can cause uf2 to be written to the wrong device.
For example, the problem occurs in the following cases

* XIAO and WioTerminal are connected to the PC
* While developing WioTerminal, wrote a program that caused panic
* Ran `tinygo flash -target wioterminal` in this state
* touchSerialPortAt1200bps() to XIAO and flash wrong UF2 for Wio Terminal

| json | msd-volume-name |
| -- | -- |
| targets/feather-m0.json |    "msd-volume-name": "FEATHERBOOT" |
| targets/feather-m4.json |    "msd-volume-name": "FEATHERBOOT" |
| targets/microbit-v2.json |    "msd-volume-name": "MICROBIT" |
| targets/microbit.json | "msd-volume-name": "MICROBIT" |
| targets/wioterminal.json |    "msd-volume-name": "Arduino" |
| targets/xiao.json |    "msd-volume-name": "Arduino" |

The following process is similar to that of secondaryPorts, but without the candidates.
This is the case when VID/PID is not defined in `targets/*.json`.
For now, I believe that this process is less problematic without any changes.
However, if changes are needed, please comment.

https://github.com/tinygo-org/tinygo/blob/d1baa392edeefe1ee9b7c4bd7287a1d554af8be8/main.go#L1039-L1045

### DIFF
--- a/main.go
+++ b/main.go
@@ -978,8 +978,7 @@ func getDefaultPort(portFlag string, usbInterfaces []string) (port string, err e
 			preferredPortIDs = append(preferredPortIDs, [2]uint16{uint16(vid), uint16(pid)})
 		}
 
-		var primaryPorts []string   // ports picked from preferred USB VID/PID
-		var secondaryPorts []string // other ports (as a fallback)
+		var primaryPorts []string // ports picked from preferred USB VID/PID
 		for _, p := range portsList {
 			if !p.IsUSB {
 				continue
@@ -1001,8 +1000,6 @@ func getDefaultPort(portFlag string, usbInterfaces []string) (port string, err e
 					continue
 				}
 			}
-
-			secondaryPorts = append(secondaryPorts, p.Name)
 		}
 		if len(primaryPorts) == 1 {
 			// There is exactly one match in the set of preferred ports. Use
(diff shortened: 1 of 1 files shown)


deadprogram:
Seems like a much safer assumption to me. Now merging, thanks @sago35

aykevl:
This change breaks support for the ESP32-C3. Try the following to reproduce:

    tinygo flash -target=esp32c3 examples/serial

It will exit with `error: unable to locate a serial port` even though it has a serial port available: /dev/ttyUSB0.

aykevl:
I think it fails in the following way:

1. The vid/pid pair of this chip is not specified in the target JSON file (because there are too many), therefore `primaryPorts` is empty.
2. `len(ports)` remains 0, so it uses the fallback to find serial ports.
3. The fallback only works for `/dev/ttyACM*` ports, not for `/dev/ttyUSB*` ports (as used on many boards, including my esp32-c3 board).

Therefore, my conclusion is that we should not use the fallback and instead use the other ports as returned from `GetDetailedPortsList`.

aykevl:
I think a complete fix for the issue described by @sago35 is by having a way to say "this board _only_ uses these vid/pid pairs". For example, you could say that Wio Terminal only ever uses 2886:802d, and if that vid/pid pair is not available it should exit with an error instead of attempting to flash using other serial ports. That would have prevented flashing to the wrong board in this case.

(Alternatively, it could be the other way round: let some boards like the Arduino Uno and most esp32 based boards say that the `serial-port` array is not complete).

PR #3094, opened 2022-08-24
runtime: implement resetTimer

joeshaw:
runtime: implement resetTimer

A follow up to #2954 / #1402, which implemented startTimer and stopTimer.

### DIFF
--- a/src/runtime/time.go
+++ b/src/runtime/time.go
@@ -13,6 +13,7 @@ func (t *timerNode) whenTicks() timeUnit {
 }
 
 // Defined in the time package, implemented here in the runtime.
+//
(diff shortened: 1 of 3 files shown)


deadprogram:
Looks like you might need to rebase against the latest `dev` branch?

joeshaw:
It was latest `dev` when I pushed it, but it looks like the return value from `resetTimer` is not consistent. That needs investigating.

joeshaw:
Actually, I think it was just flaky test timing.  Force-pushed an update to hopefully fix that 🤞

deadprogram (approved):
Thanks for working on this @joeshaw

PR #3097, opened 2022-08-26
boards: Add XIAO ESP32C3 board

ysoldak:
Seeed Studio [XIAO ESP32C3](https://www.seeedstudio.com/Seeed-XIAO-ESP32C3-p-5431.html) is an IoT mini development board based on the Espressif ESP32-C3 WiFi/Bluetooth dual-mode chip.


deadprogram:
Thank you for the new board @ysoldak now merging.

PR #3098, opened 2022-08-26
targets: fail fast on duplicate values in target field slices

ysoldak:
This is how it fails now:
```
/Users/yursol/bin/tinygo build -size short -o test.hex -target=p1am-100            examples/blinky1
panic: duplicate value 'atsamd21g18a' in field : BuildTags

goroutine 1 [running]:
github.com/tinygo-org/tinygo/compileopts.(*TargetSpec).overrideProperties(0xc00028e2c0, 0xc00028e000)
	/Users/yursol/Private/code/tinygo/compileopts/target.go:98 +0x848
github.com/tinygo-org/tinygo/compileopts.(*TargetSpec).resolveInherits(0xc00028e000)
	/Users/yursol/Private/code/tinygo/compileopts/target.go:157 +0x4d
github.com/tinygo-org/tinygo/compileopts.LoadTarget(0xc00028c000)
	/Users/yursol/Private/code/tinygo/compileopts/target.go:223 +0x2b3
github.com/tinygo-org/tinygo/builder.NewConfig(0xc00028c000)
	/Users/yursol/Private/code/tinygo/builder/config.go:16 +0x33
main.Build({0x7ff7bfeff9a8, 0x10}, {0x7ff7bfeff98e, 0x8}, 0xc00028c000)
	/Users/yursol/Private/code/tinygo/main.go:143 +0x56
main.main()
	/Users/yursol/Private/code/tinygo/main.go:1443 +0x3c75
make: *** [smoketest] Error 2

```

### DIFF
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -88,8 +88,17 @@ func (spec *TargetSpec) overrideProperties(child *TargetSpec) {
 			if !src.IsNil() {
 				dst.Set(src)
 			}
-		case reflect.Slice: // for slices, append the field
+		case reflect.Slice: // for slices, append the field and check for duplicates
 			dst.Set(reflect.AppendSlice(dst, src))
+			for i := 0; i < dst.Len(); i++ {
+				v := dst.Index(i).String()
+				for j := i + 1; j < dst.Len(); j++ {
+					w := dst.Index(j).String()
+					if v == w {
+						panic("duplicate value '" + v + "' in field : " + field.Name)
+					}
+				}
+			}
 		default:
 			panic("unknown field type : " + kind.String())
 		}
--- a/targets/p1am-100.json
+++ b/targets/p1am-100.json
@@ -1,6 +1,6 @@
 {
     "inherits": ["atsamd21g18a"],
-    "build-tags": ["sam", "atsamd21g18a", "p1am_100"],
+    "build-tags": ["p1am_100"],
     "flash-command": "bossac -d -i -e -w -v -R --port={port} --offset=0x2000 {bin}",
     "flash-1200-bps-reset": "true"
 }


ysoldak:
The target file is fixed, but it fails timer test on macOS.
Seem to be a random glitch.

deadprogram:
@ysoldak please rebase against latest `dev` I think that spurious fail was fixed.

ysoldak:
@deadprogram done

compileopts/target.go
```diff
@@ -88,8 +88,17 @@ func (spec *TargetSpec) overrideProperties(child *TargetSpec) {
...
 			}
-		case reflect.Slice: // for slices, append the field
+		case reflect.Slice: // for slices, append the field and check for duplicates
 			dst.Set(reflect.AppendSlice(dst, src))
+			for i := 0; i < dst.Len(); i++ {
+				v := dst.Index(i).String()
+				for j := i + 1; j < dst.Len(); j++ {
+					w := dst.Index(j).String()
+					if v == w {
+						panic("duplicate value '" + v + "' in field : " + field.Name)
```
deadprogram, line 98: Perhaps add info about which target failed here e.g. target name.

ysoldak: No such information here?
All we have is `TargetSpec` struct, i.e. different fields, but not the actual file name.

I don't think we need this either, since we always know what target we are building or flashing. And the problem is either in the target file itself or in a target up the chain.

In this repo, smoke test  that builds every board target shall catch all such cases if we ever have them and the last target built, and failed, is the one that has the problem.

deadprogram: OK fair enough. I'll go with that.

deadprogram:
Merging, thanks for the improvements @ysoldak

PR #3099, opened 2022-08-27
compiler: use the LLVM builtins everywhere

aykevl:
This gives some more optimization opportunities to LLVM, because it understands these intrinsics. For example, it might convert llvm.sqrt.f64 to llvm.sqrt.f32 if possible (for example, if code does `float32(math.Sqrt(float64(v))`).

---

This change should make some future changes easier to do.

It probably won't pass CI tests, I need to look into an issue on Windows at least.


deadprogram:
@aykevl this PR is passing all tests. Is there still something that needs to be looked into before merge?

aykevl:
Yes it is ready.

deadprogram:
Now merging, thanks @aykevl

PR #3101, opened 2022-08-28
machine: Add support for Adafruit QT2040 board.

schultz-is:
This adds support for the Adafruit QT2040 board, based on the RP2040. The board shares a few properties with the already-supported Adafruit Feather RP2040: both use an 8MB QSPI flash chip, a 12MHz oscillator, and have a STEMMA QT port for I2C. However the QT2040 appears intended mainly for I2C development, and doesn't provide any pin mapping outside of its integrated NeoPixel LED and QT port. Due to the lack of a simple onboard LED, the smoke test I've included is the RP2040 ADC example, but I'm happy to change that to something else if y'all would prefer.

Thanks y'all!

EDIT: Forgot to mention that I pulled the USB VID/PID from the [Arduino library](https://github.com/earlephilhower/arduino-pico/blob/a97dc38e8778f531ec4d7eddfe66f464d8be434a/tools/json/adafruit_trinkeyrp2040qt.json#L6-L7) recommended in Adafruit's documentation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -597,6 +597,8 @@ endif
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=challenger-rp2040    examples/blinky1
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=trinkey-qt2040      examples/adc_rp2040
+	@$(MD5SUM) test.hex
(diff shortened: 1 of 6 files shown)


deadprogram:
Looks great @schultz-is thank you very much! Now merging.

PR #3102, opened 2022-08-28
Math intrinsics

aykevl:
Continuation of #3099. This reduces the number of math operations in alias.go to just four. Aliases are really just a workaround so I prefer to support them in a different way.

This PR needs to be rebased on #3099 once it is merged. (Alternatively, this PR could be merged instead of #3099 as the first commit is identical to #3099).


aykevl:
I mis-remembered that we were using Go 1.17 as the minimum version. We're still at Go 1.16, which is why the CI fails.

aykevl:
This PR should be rebased after merging #2904.

aykevl:
It is now rebased, so ready to merge.

deadprogram:
Thanks @aykevl now merging :1234: :heavy_division_sign: :1234:

Issue #2779, opened 2022-04-15
tinygo:ld.lld: error: undefined symbol: crypto/aes.encryptBlockAsm


d33tah:
I tried to give tinygo a try by building a simple "hello world" web service. Instead, I found that crypto/aes doesn't compile:

```
> sudo docker run --rm tinygo/tinygo:0.22.0 tinygo test crypto/aes 
tinygo:ld.lld: error: undefined symbol: crypto/aes.encryptBlockAsm
>>> referenced by cipher_asm.go:69 (/usr/local/go/src/crypto/aes/cipher_asm.go:69)
>>>               /tmp/tinygo3958504993/main.o:((*crypto/aes.aesCipherAsm).Encrypt)

tinygo:ld.lld: error: undefined symbol: crypto/aes.decryptBlockAsm
>>> referenced by cipher_asm.go:82 (/usr/local/go/src/crypto/aes/cipher_asm.go:82)
>>>               /tmp/tinygo3958504993/main.o:((*crypto/aes.aesCipherAsm).Decrypt)
failed to run tool: ld.lld
FAIL
error: failed to link /tmp/tinygo3958504993/main: exit status 1
```

It looks like those packages are on the list of the ones [supported by TinyGo](https://tinygo.org/docs/reference/lang-support/stdlib/), so what's the problem?

dkegel-fastly:
You need to read the large type at the top of that page :-)

A PR is pending to update that page to show which packages actually pass tests, which ought to help clear up the confusion.

The tests that pass in the dev branch (which will soon become the next version, 0.23) on darwin, linux, wasi, and windows can be found by checking out that branch and typing "make report-stdlib-tests-pass".  Here's what it currently says:

[report.txt](https://github.com/tinygo-org/tinygo/files/8496646/report.txt)

Note that net/http is not in the list whose tests pass.  So, you can import it, but you can't use all the functions in it yet.

fmt is also not in the list, probably because %v doesn't pass tests (because reflection is not fully supported), but it's useful otherwise.

d33tah:
Got it. Thanks for pointing that out! I guess that the issue can be closed then.

dkegel-fastly:
It's a real enough bug.  Easier way to demonstrate it is:

```
C02DG0QJMD6R:tinygo dkegel$ build/tinygo test crypto/aes
tinygo:ld.lld: error: undefined symbol: _crypto/aes.encryptBlockAsm
[...]

Issue #2578, opened 2022-01-24
Error printing time.Now() on 386 and arm; "undefined symbol: syscall.seek"


dkegel-fastly:
With current dev, the test program
```
package main
import "fmt"
import "time"
func main() {
	fmt.Printf("The time is %v\n", time.Now())
}
```
fails to link on 386 and arm, e.g.
```
GOARCH=386 GOOS=linux tinygo run main.go
tinygo:ld.lld: error: undefined symbol: syscall.seek
```

With https://github.com/tinygo-org/tinygo/pull/2571 applied, the program links and seems to produce correct output on Ubuntu amd64 for arm, 386, and native:
```
$ GOARCH=arm tinygo run main.go
The time is 2022-01-24 08:19:08.892313672 -0800 PST m=+1158.118011446
$ GOARCH=386 tinygo run main.go
The time is 2022-01-24 08:19:42.121299131 -0800 PST m=+1191.347873439
$ tinygo run main.go
The time is 2022-01-24 08:19:58.911269014 -0800 PST m=+1208.138197336
```

PR #3103, opened 2022-08-28
Add support for reading Go assembly files

aykevl:
This is a pretty big change that adds support for Go assembly files.
Go assembly files are different from standard assembly and can't be read
by tools like Clang. They also use a different calling convention. On
top of that, the tools that can read them (the gc toolchain) only output
custom (unstable) object files. In a previous attempt I tried to parse
these files but with this attempt I simply read the output of
`go tool asm -S`, which is a more-or-less stable text format (that,
importantly, includes the raw machine code and all relocations).

As a result:

  * This implements syscall.seek on 386 and arm. The following works with this commit: `GOARCH=arm tinygo test os`. See #2578.
  * This changes the math package, probably speeding up some algorithms (although I didn't benchmark them).
  * The crypto/aes package now works (issue #2779).
  * Undefined references to crypto/internal/boring/sig.StandardCrypto in Go 1.19 are fixed (see #3057, #3080).

So far, this is only for Linux (on all supported architectures: 386,
amd64, arm, arm64). Adding support for macOS and Windows should be
feasible by implementing the appropriate file formats (MachO and COFF).
It should be relatively easy to also add support for js/wasm. I'm not so
sure about other systems where the GOARCH doesn't match the actual
architecture, like WASI and most/all baremetal systems. It might be
possible to get them to work by lifting ARM machine code to LLVM IR, but
that will be quite some work in itself and certainly something for the
longer term.

---

This replaces #2688. It is still a draft because it depends on #3102 (math intrinsics) and #2904 (drop Go 1.16/1.17 support). It needs to be rebased on top of the dev branch once those PRs are in.


compiler/goasm.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+	stackStruct := types.NewStruct(stackStructFields, nil)
+	stackStructOffsets := sizes.Offsetsof(stackStructFields)
+
+	// Create the alloca.
+	// WARNING: we're assuming here that this alloca will be the same as the
+	// stack pointer before the call in the assembly. This is not necessarily
+	// the case. But it seems to be working.
+	// I have tried many other approaches and this appears to be the most
+	// reliable way to do it.
```
aykevl: This is not great, and I'd like to experiment with other ways of doing this once this is in. My main priority has been compatibility: making sure Go assembly works at all.

deadprogram:
Looks like macOS and Linux not passing CI on this @aykevl

Makefile
```diff
@@ -346,7 +346,10 @@ TEST_PACKAGES_LINUX := \
 	testing/fstest \
 	text/template/parse
 
-TEST_PACKAGES_DARWIN := $(TEST_PACKAGES_LINUX)
+TEST_PACKAGES_LINUX := \
+	$(TEST_PACKAGES_LINUX) \
```
dgryski: Should this be `TEST_PACKAGES_DARWIN` ?

aykevl: You are correct, thanks for spotting this! I've pushed a fix.

deadprogram:
Looks like needs merge conflict resolved @aykevl

aykevl:
Rebased to fix the merge conflict.

aykevl:
Can someone take a look at this? I have a number of other changes waiting for this PR to be merged.

dgryski:
I tried reviewing this, but there's too much compiler stuff I don't yet understand.  Can we put this behind a flag temporarily so that it can be merged and allow us to do more tests before making it the default?

Also, is there anything inherently preventing this from working on Darwin, or just "haven't gotten around to it yet"?

aykevl:
> I tried reviewing this, but there's too much compiler stuff I don't yet understand.

If you need help with anything, feel free to ask.

>  Can we put this behind a flag temporarily so that it can be merged and allow us to do more tests before making it the default?

I guess, but we'd have to review it some time anyway. Not sure what the advantage is. Avoiding breakage?

> Also, is there anything inherently preventing this from working on Darwin, or just "haven't gotten around to it yet"?

I just need to write a port to MachO, shouldn't be too difficult. I already made one for Windows.
(Those are things I will make PRs for once this PR is in).

dgryski:
Yes.  We can review it and test it more thoroughly without holding up other PRs and without creating large merge conflicts.

Glad to hear Darwin support is possible.

kenbell:
I think this change also means we can get rid of this 'hack' in the test cases: https://github.com/tinygo-org/tinygo/blob/dev/main_test.go#L173-L180

aykevl:
@dgryski I still don't understand why this feature should be disabled in the beginning? If it's disabled, that won't help with testing.
What is it exactly that you're proposing and what's the advantage of it?

> I think this change also means we can get rid of this 'hack' in the test cases: https://github.com/tinygo-org/tinygo/blob/dev/main_test.go#L173-L180

Thanks, yes it does! Happy to see it go away. I've updated this PR.

dgryski:
@aykevl My goal with the `merge but with flag to enable` was to allow this to unblock other patches (as you said above) and give people the opportunity to test it more easily until we made it the default.

I'll try to understand this patch tonight and you can merge it (since it seems to be working.)

dgryski:
My suggestion is to run the test corpus, removing all the `purego` and `noasm` tags.  If that passes, we can merge this and debug any issues as they come up.

deadprogram:
Just catching up on this PR. Did we ever perform the test suggested by @dgryski and if so do we consider it ready to merge?

deadprogram:
@dgryski and @aykevl have we performed the recommended tests and/or are we ready to merge this?

dgryski:
@dkegel-fastly You have a linux box.  You want to take a look at running these tests?

dkegel-fastly:
will do

dankegel:
Hrm.   "make test" fails on my linux box with 
```
tinygo:ld.lld: error: undefined symbol: log
>>> referenced by log.go:80 (/usr/lib/go-1.18/src/math/log.go:80)
>>>               /home/dank/.cache/tinygo/thinlto/llvmcache-CC40CEA9C88595CE4A2271BD52FFCD6A1FF971D3:(runtime.run$1$gowrapper)
>>> referenced by log.go:80 (/usr/lib/go-1.18/src/math/log.go:80)
>>>               /home/dank/.cache/tinygo/thinlto/llvmcache-CC40CEA9C88595CE4A2271BD52FFCD6A1FF971D3:(runtime.run$1$gowrapper)
>>> referenced by log.go:80 (/usr/lib/go-1.18/src/math/log.go:80)
>>>               /home/dank/.cache/tinygo/thinlto/llvmcache-CC40CEA9C88595CE4A2271BD52FFCD6A1FF971D3:(runtime.run$1$gowrapper)
>>> referenced 4 more times

tinygo:ld.lld: error: undefined symbol: ceil
>>> referenced by floor.go:41 (/usr/lib/go-1.18/src/math/floor.go:41)
>>>               /home/dank/.cache/tinygo/thinlto/llvmcache-CC40CEA9C88595CE4A2271BD52FFCD6A1FF971D3:(runtime.run$1$gowrapper)

tinygo:ld.lld: error: undefined symbol: exp
>>> referenced by sinh.go:0 (/usr/lib/go-1.18/src/math/sinh.go:0)
>>>               /home/dank/.cache/tinygo/thinlto/llvmcache-CC40CEA9C88595CE4A2271BD52FFCD6A1FF971D3:(runtime.run$1$gowrapper)
>>> referenced by exp.go:14 (/usr/lib/go-1.18/src/math/exp.go:14)
>>>               /home/dank/.cache/tinygo/thinlto/llvmcache-CC40CEA9C88595CE4A2271BD52FFCD6A1FF971D3:(runtime.run$1$gowrapper)
>>> referenced by exp.go:14 (/usr/lib/go-1.18/src/math/exp.go:14)
>>>               /home/dank/.cache/tinygo/thinlto/llvmcache-CC40CEA9C88595CE4A2271BD52FFCD6A1FF971D3:(runtime.run$1$gowrapper)
>>> referenced 7 more times

tinygo:ld.lld: error: undefined symbol: exp2
>>> referenced by exp.go:140 (/usr/lib/go-1.18/src/math/exp.go:140)
>>>               /home/dank/.cache/tinygo/thinlto/llvmcache-CC40CEA9C88595CE4A2271BD52FFCD6A1FF971D3:(runtime.run$1$gowrapper)

tinygo:ld.lld: error: undefined symbol: floor
>>> referenced by floor.go:13 (/usr/lib/go-1.18/src/math/floor.go:13)
>>>               /home/dank/.cache/tinygo/thinlto/llvmcache-CC40CEA9C88595CE4A2271BD52FFCD6A1FF971D3:(runtime.run$1$gowrapper)

tinygo:ld.lld: error: undefined symbol: trunc
>>> referenced by floor.go:58 (/usr/lib/go-1.18/src/math/floor.go:58)
>>>               /home/dank/.cache/tinygo/thinlto/llvmcache-CC40CEA9C88595CE4A2271BD52FFCD6A1FF971D3:(runtime.run$1$gowrapper)
failed to run tool: ld.lld
--- FAIL: TestBuild (0.00s)
    --- FAIL: TestBuild/Host (0.00s)
        --- FAIL: TestBuild/Host/math.go (1.25s)
            main.go:1192: error: failed to link /tmp/tinygo1886996487/main: exit status 1
FAIL
FAIL	github.com/tinygo-org/tinygo	14.778s
```

It's like it needs -lm or something.

aykevl:
@dankegel most likely you need to run `tinygo clean`: you seem to be using and old cached build of musl. See #3099 for details.

dankegel:
Thanks.  I'm afraid I'm hitting panics on the real test, though.

e.g.
```
=== CONT  TestCorpus/github.com/dgryski/go-postings
    main_test.go:523: go: creating new go.mod: module github.com/tinygo/tinygo-corpus-test
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc0413fe4a0 stack=[0xc0413fe000, 0xc0613fe000]
fatal error: stack overflow

runtime stack:
runtime.throw({0x57c250d?, 0x77bcda0?})
	/usr/lib/go-1.18/src/runtime/panic.go:992 +0x71
runtime.newstack()
	/usr/lib/go-1.18/src/runtime/stack.go:1101 +0x5cc
runtime.morestack()
	/usr/lib/go-1.18/src/runtime/asm_amd64.s:547 +0x8b
...
```

What's the easiest way to tell the corpus test to use a bigger stack?

dgryski:
Pretty sure that test shouldn't be hitting a stack overflow. The assembly version of the code is a few instructions and a table lookup.

Edit: oh, saw the stack overflow was in the compiler not the test.

aykevl:
> What's the easiest way to tell the corpus test to use a bigger stack?

It's using a 1GB stack, more stack is not going to help you here.

(Now running the corpus test to reproduce the issue)

aykevl:
I rebased this PR and I can't reproduce the stack overflow. I think it has been fixed in the dev branch.
(That said, I'm running on a slightly unusual setup with a linux/arm64 host and testing on linux/amd64 via `qemu-x86_64`).

dkegel-fastly:
I'll retest tonight.

aykevl:
With some testing (now with the build tags removed), I don't think this is quite ready yet:

- github.com/dgryski/go-bloomindex crashed QEMU with a SIGILL. I suspect this is actually a QEMU bug (apparently the `ptest` instruction isn't commonly used and isn't implemented in QEMU)
- github.com/dgryski/go-chaskey results in a SIGSEGV, apparently because it jumps somewhere it really shouldn't.
- github.com/dgryski/go-cuckoof also resulted in a SIGSEGV, this time it appears to happen somewhere in the Go assembly wrapper code

So while the first seems to be a QEMU bug, the other two are most likely real bugs that I need to fix. I did not yet test other packages than these three.

dkegel-fastly:
test corpus FTW!

aykevl:
> github.com/dgryski/go-bloomindex crashed QEMU with a SIGILL. I suspect this is actually a QEMU bug (apparently the `ptest` instruction isn't commonly used and isn't implemented in QEMU)


Not anymore with a recent QEMU (apparently they implemented `ptest`) but it now triggers a segmentation fault... so there is still a bug in TinyGo.

dgryski:
I'm always so pleased when testing reveals bugs :D

aykevl:
Took me a while to figure out, but I think I found the bug: LLVM doesn't save/restore the `rbp` register through the call even though the assembly marks it as clobbered.

aykevl:
I know a solution to this, but it's going to need a bit of a rewrite. It's a cleanup I've wanted to do _after_ this PR landed but it appears I now need to clean up my code for it to work :smiling_face_with_tear:

aykevl:
The rewrite works. Those 3 tests that failed before now pass with my new implementation (I didn't test any others yet). The resulting assembly is a lot cleaner so it's probably also a lot faster (if that matters).

Now I need to clean it up, implement all the loose ends, and make sure it also works on all other architectures. And of course test the rest of the corpus.

dankegel:
Currently fails with
```
    main_test.go:523: FAIL	github.com/dgryski/go-bloomindex	0.174s
    main_test.go:523: FAIL	github.com/dgryski/go-metro	0.179s
    main_test.go:523: FAIL	github.com/dgryski/go-marvin32	0.176s
    main_test.go:523: FAIL	github.com/dgryski/go-farm	0.177s
    main_test.go:523: FAIL	github.com/dgryski/go-cuckoof	0.176s
```
and then a stack overflow (which probably hides a bunch of failures).

I narrowed the go-farm failure down slightly; the hash tests pass, but 'tinygo test -run TestFingerprint32' fails. 

Also, go-postings on its own still fails here :-(

aykevl:
I didn't push any of the new code yet. In my branch `github.com/dgryski/go-cuckoof` passes for example. I'll push it when it's in a mergeable state.

aykevl:
> Currently fails with
> 
> ```
>     main_test.go:523: FAIL	github.com/dgryski/go-bloomindex	0.174s
>     main_test.go:523: FAIL	github.com/dgryski/go-metro	0.179s
>     main_test.go:523: FAIL	github.com/dgryski/go-marvin32	0.176s
>     main_test.go:523: FAIL	github.com/dgryski/go-farm	0.177s
>     main_test.go:523: FAIL	github.com/dgryski/go-cuckoof	0.176s
> ```

All these work now.

aykevl:
Almost ready, I just need to do some cleanup.
Can someone run the test corpus on linux/amd64? @dgryski?

dgryski:
I'm on Darwin. Also, any corpus runs will need to adjust the tags to remove `purego` and `noasm`.

aykevl:
Investigating whether I can adjust the corpus test so that I can run the tests themselves on linux/amd64 through QEMU (just like `GOARCH=arm tinygo test foo`). Seems to be working well so far.

> Also, any corpus runs will need to adjust the tags to remove `purego` and `noasm`.

I can test this once I am able to run the linux/amd64 corpus test on my system.

aykevl:
I ran the test corpus. I could indeed remove a bunch of these build tags, but not all of them. I think it's easier to fix the remaining issues in separate PRs.

dgryski:
I wonder if we want to specialize the build tags per platform? Removing the asm ones breaks the corpus until Darwin asm support lands.

aykevl:
> I wonder if we want to specialize the build tags per platform? Removing the asm ones breaks the corpus until Darwin asm support lands.

Sounds like a good idea!
Do you have a suggestion how to best add this to the yaml file? Separate build tags for linux/macos/windows for example?

dgryski:
That's probably the easiest. `tags_linux`, `tags_darwin`, `tags_wasi`. Hopefully we can avoid needing to add `arm` and `amd64` variants.

aykevl:
> That's probably the easiest. `tags_linux`, `tags_darwin`, `tags_wasi`.

Ok, will do.

>  Hopefully we can avoid needing to add `arm` and `amd64` variants.

Actually, the test corpus doesn't pass on linux/arm64. I normally modify the test a bit like this to get it to work on linux/arm64:

```diff
--- a/corpus_test.go
+++ b/corpus_test.go
@@ -110,7 +110,7 @@ func TestCorpus(t *testing.T) {
                                out := ioLogger(t, &wg)
                                defer out.Close()
 
-                               opts := optionsFromTarget(target, sema)
+                               opts := optionsFromOSARCH("linux/amd64", sema)
                                opts.Directory = dir
                                var tags buildutil.TagsFlag
                                tags.Set(repo.Tags)
```

aykevl:
Please take another look. I've rebased the changes, added the new build tags in the test corpus, and squashed the goasm changes into a single commit.

aykevl:
^ reminder, can someone take a look at this?

dgryski:
I will.

deadprogram:
@aykevl this PR needs to be rebased against the latest `dev`.

@dgryski can you please take a look to see if we can get this merged?

Thanks!

deadprogram:
@aykevl can you please rebase this PR one last time?

@dgryski if you can, please take a look. I will do the same, so we can get this merged early in this release cycle.

Thanks!

aykevl:
Rebased. It does show the following warning, which I haven't investigated yet:

```
$ go install; and tinygo test math
warning: GOPATH set to GOROOT () has no effect
warning: GOPATH set to GOROOT () has no effect
warning: GOPATH set to GOROOT () has no effect
warning: GOPATH set to GOROOT () has no effect
warning: GOPATH set to GOROOT () has no effect
warning: GOPATH set to GOROOT () has no effect
warning: GOPATH set to GOROOT () has no effect
ok      math    0.001s
```

deadprogram:
> It does show the following warning, which I haven't investigated yet.

OK I will hold off on merging until you say so @aykevl

jclab-joseph:
I found this PR while looking for an issue in tinygo that prevented me from using asm in the default package.
Thanks for your contribution! I look forward to the day when this becomes a reality :)

dgryski:
@jclab-joseph What package are you trying to use?  Frequently a pure-Go implementation can be selected with the correct build tags.

jclab-joseph:
@dgryski Yes, I have been using it that way.
However, I would like to use AES-NI on x86. This requires asm.

MDr164:
Anything I can do to help move this PR forward?

aykevl:
I think the general idea is that #4116 is the way forward, instead of trying to implement support for Go assembly in an extremely hacky way (that wouldn't support microcontrollers anyway). So closing this PR.

Issue #3106, opened 2022-08-31
macos M1: Flash pybadge or ArduinoNano33 freezes my laptop and restarts it


doniacld:
## Description of the problem

Setup:
MacOS Monterey
MacBook: Pro
Chip: Apple M1 Pro
Memory: 16 GB

When I plug my microntroller through USB to my computer and I flash it, sometimes I have the following behaviour:
1. It freezes my computer
2. It restarts it 
3. It proposes me to report the error below

```
panic(cpu 0 caller 0xfffffe001cd379e4): Kernel data abort. at pc 0xfffffe001ddc599c, lr 0x23907e001ddc59d8 (saved state: 0xfffffe60ba1bb3e0)
	  x0:  0x0000000000000000 x1:  0xfffffe001feaa768  x2:  0xfffffe150e3a4034  x3:  0xfffffe150e3a6e20
	  x4:  0xfffffe150e3a6e00 x5:  0x0000000000000000  x6:  0x96cd7e001cc51c2c  x7:  0xfffffe001f30a160
	  x8:  0x0000000000000000 x9:  0xfffffe150f173f90  x10: 0x00000000000032c0  x11: 0x0000000000004001
	  x12: 0x0001af1129900000 x13: 0x0000000000000154  x14: 0xfffffe150e879808  x15: 0x00000000004b0155
	  x16: 0xfffffe001f309740 x17: 0x1ba6fe001f309740  x18: 0x0000000000000000  x19: 0x00000000e00002ed
	  x20: 0xfffffe60ba1bb7f8 x21: 0xfffffe152fecfba0  x22: 0x0000000000000000  x23: 0x0000000000000001
	  x24: 0x37fdfe001dd5a8ac x25: 0x0000000000000000  x26: 0xfffffe60ba1bb8ae  x27: 0xcda1fe150f049d10
	  x28: 0xfffffe15349d2238 fp:  0xfffffe60ba1bb7b0  lr:  0x23907e001ddc59d8  sp:  0xfffffe60ba1bb730
	  pc:  0xfffffe001ddc599c cpsr: 0x60401208         esr: 0x96000006          far: 0x0000000000000000

Debugger message: panic
Memory ID: 0x1
OS release type: User
OS version: 21A559
Kernel version: Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:01 PDT 2021; root:xnu-8019.41.5~1/RELEASE_ARM64_T6000
[...]

PR #3107, opened 2022-08-31
flash: update serial package to v1.3.5 for latest bugfixes

deadprogram:
This PR updates the serial package to v1.3.5 in order to pick up the latest bugfixes.

Hopefully addressing #3106 and other related flashing problems.


sago35 (approved):
I don't know if this change will fix the problem, but it is a good update.

PR #3110, opened 2022-08-31
wioterminal: add UART3 for RTL8720DN

sago35:
Currently, the RTL8270DN driver is complicated by the fact that the sercomUSARTx definition is not published.
In this PR, UART3 is defined as the UART that must be used when using the RTL8720DN.

https://github.com/tinygo-org/drivers/pull/453

### DIFF
--- a/src/machine/board_wioterminal.go
+++ b/src/machine/board_wioterminal.go
@@ -349,8 +349,11 @@ var (
 
 	UART1 = &sercomUSART2
 
-	// RTL8720D
+	// RTL8720D (tx: PC22, rx: PC23)
 	UART2 = &sercomUSART1
+
+	// RTL8720D (tx: PB24, rx: PC24)
+	UART3 = &sercomUSART0
 )
 
 // I2C pins


src/machine/board_wioterminal.go
```diff
@@ -349,8 +349,11 @@ var (
 
 	UART1 = &sercomUSART2
 
-	// RTL8720D
+	// RTL8720D (tx: PC22, rx: PC23)
```
deadprogram, line 352: @sago35 should this be removed? Or it is possible to use UART2 with the RTL8720D?

sago35: @deadprogram 

It is possible to use UART2 with the RTL8720D.

These ports are not required for networking, but are needed for rewriting and debugging.
Probably most people will never handle it, but driver developers may need it from time to time.

image

deadprogram: Thank you very much for the incredibly clear explanation!

deadprogram (approved):

deadprogram:
Thanks @sago35 now merging.

Issue #2983, opened 2022-07-13
not support #v directive in fmt.Sprintf


hunjixin:
code
```
func main() {
	fmt.Sprintf("%#v", struct {
		str string
	}{str: "asdasdasda"})
}
```

err message
```
panic: (reflect.Value).Interface: unexported
error: failed to run compiled binary /var/folders/_v/lltqd75s3tsdpkr65klgzwch0000gn/T/tinygo2195710682/main: signal: abort trap
```

reason for this issue is reflect. unablel to get value of private field, did plan to support this feature?

deadprogram:
Released as part of `v0.26.0`

Issue #3033, opened 2022-07-29
allow read private fileld  for Interface Method in reflect.Value


hunjixin:
reproducer code
```
package main

import (
	"fmt"
	"reflect"
)

type A struct {
	a string
}


func main() {
	v := reflect.ValueOf(A{"xxxx"})
	fmt.Println(v.Interface())
}
```

go run main.go
{xxxx}
 tinygo run main.go
panic: (reflect.Value).Interface: unexported
error: failed to run compiled binary /var/folders/_v/lltqd75s3tsdpkr65klgzwch0000gn/T/tinygo891702810/main: signal: abort

hunjixin:
https://github.com/tinygo-org/tinygo/issues/2983

aykevl:
Thank you for the bug report! I have found and fixed the issue here: https://github.com/tinygo-org/tinygo/pull/3111

deadprogram:
Released as part of `v0.26.0`

PR #3111, opened 2022-08-31
reflect: implement CanInterface and fix string Index()

aykevl:
This commit fixes two related issues:

 1. CanInterface was unimplemented. It now uses the same check as is used in Interface() itself.
    This issue led to https://github.com/tinygo-org/tinygo/issues/3033
 2. Allow making an interface out of a string char element.

Doing this in one commit (instead of two) because they are shown to be correct with the same tests.

### DIFF
--- a/src/reflect/value.go
+++ b/src/reflect/value.go
@@ -168,8 +168,7 @@ func (v Value) IsValid() bool {
 }
 
 func (v Value) CanInterface() bool {
-	// No Value types of private data can be constructed at the moment.
-	return true
+	return v.isExported()
 }
 
 func (v Value) CanAddr() bool {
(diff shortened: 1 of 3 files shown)


dgryski:
Does this allow any other stdlib tests to pass?

aykevl:
I didn't test it but I don't think so: none exit with a `panic: (reflect.Value).Interface: unexported`.

hunjixin:
i have tested it , work fine, but i am not understand how this pr works for this case

hunjixin:
I found the reason 
```go
func (p *pp) printValue(value reflect.Value, verb rune, depth int) {
	// Handle values with special methods if not already handled by printArg (depth == 0).
	if depth > 0 && value.IsValid() && value.CanInterface() {
		p.arg = value.Interface()
		if p.handleMethods(verb) {
			return
		}
	}
```

hunjixin:
this issue also fix in this pr 
close https://github.com/tinygo-org/tinygo/issues/2983

deadprogram (approved):

deadprogram:
Thanks for the implementation @aykevl and thank you @dgryski and @hunjixin for review/testing. Now merging.

PR #3123, opened 2022-09-02
windows: save and restore xmm registers when switching goroutines

aykevl:
This is required according to the Windows x64 calling convention:
https://docs.microsoft.com/en-us/cpp/build/x64-calling-convention?view=msvc-170#callercallee-saved-registers
We didn't do this, and this didn't appear to cause any problems. But we should do it anyway for correctness.


niaow (approved):
LGTM

deadprogram:
Thanks @aykevl for the improvement and thank you @niaow for reviewing! :smile_cat:

PR #3124, opened 2022-09-02
targets: remove hifive1-qemu target

aykevl:
This target was added purely for running tests, and it is currently unused. When I try to use it, it causes runtime exceptions.
The replacement riscv-qemu is much better behaved.

This doesn't drop support for any actual hardware, the HiFive 1 B will remain supported.

Oh and I also fixed a small issue in the README.


deadprogram (approved):
:scissors:

PR #3125, opened 2022-09-02
Makefile: run more tests where supported

aykevl:
I found that some packages do in fact run on Windows, so I've added them where possible. I've also updated the description of which packages fail tests and why.


aykevl:
Apparently all the tests excluded for WASI don't run there. But for Windows I managed to add four more packages to test.

dgryski (approved):
LGTM

deadprogram:
Thanks everyone! Now merging :microscope:

PR #3130, opened 2022-09-03
ci: build LLVM with thread support on Windows

aykevl:
This should fix a number of concurrency/threading issues.

---

Ok, last attempt (in a new PR because I can't reopen #2433).

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -66,7 +66,7 @@ jobs:
         uses: actions/cache/restore@v3
         id: cache-llvm-build
         with:
-          key: llvm-build-15-windows-v5
+          key: llvm-build-15-windows-v6
           path: llvm-build
       - name: Build LLVM
         if: steps.cache-llvm-build.outputs.cache-hit != 'true'
--- a/Makefile
+++ b/Makefile
@@ -120,10 +120,8 @@ ifeq ($(OS),Windows_NT)
     START_GROUP = -Wl,--start-group
     END_GROUP = -Wl,--end-group
 
-    # LLVM compiled using MinGW on Windows appears to have problems with threads.
-    # Without this flag, linking results in errors like these:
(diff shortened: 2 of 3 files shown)


aykevl:
Ok, this still doesn't work. I don't know why. I will have to debug this locally at some point.

(Note for the future: the "Test TinyGo" step seems to be hanging).

aykevl:
Tested locally with LLVM 15, which appears to work well?? Let's try it again.

aykevl:
Why oh why doesn't this work??

aykevl:
Trying once more. This time my plan is as follows:
- figure out which test hangs.
- run only that test, to check it still hangs
- build a test binary, download it, and debug the issue locally

Updates:
1. it's tinygo.test
2. It's a deadlock. When I run it locally, I can see that 4 extra processes are created, but they simply don't exit. They just continue running. Presumably, these processes are the separate `clang` / `ld.lld` processes.
3. The processes that hang are all `ld.lld` processes (e.g. `tinygo.test.exe ld.lld -m i386pep [...]`).
4. It's also hanging in the ELF linker, it's not just the COFF linker.
5. :tada: `--threads=1` seems to be a workaround, at least for ELF :unicorn:

aykevl:
This passes all tests that are affected by this change, so to me it seems like it's ready for merge.

deadprogram:
Do you also need to remove https://github.com/tinygo-org/tinygo/pull/3525 for this?

aykevl:
> Do you also need to remove #3525 for this?

Eventually, yes. But let's first see whether this PR improves the situation (because #3525 didn't seem to do a whole lot).
Regardless of #3525, I think we need this change to not misuse the LLVM APIs (using goroutines while thread support is disabled).

deadprogram:
Makes sense to me, merging!

aykevl:
It took me just over a year (and many hours debugging this), but we are finally building LLVM with thread support.

PR #3133, opened 2022-09-04
wasm: do not allow undefined symbols

aykevl:
`--allow-undefined` can be a problem: it allows compiling code that will fail when loaded. This change makes sure that if some symbols are undefined, they are reported as an error by the linker.

Previously, people could get away with importing a function that was not defined, like this:

```go
func add(int a, int b) int

func test() {
    println(add(3, 5))
}
```

This was technically unsupported but mostly worked. With this change, it isn't possible anymore. Now every function needs to be marked with `//export` explicitly:

```go
//export add
func add(int a, int b) int

func test() {
    println(add(3, 5))
}
```

As before, functions will be placed in the `env` module with the name set from the `//export` tag. This can be overridden with `//go:import-module`:

```go
//go:import-module math
//export add
func add(int a, int b) int

func test() {
    println(add(3, 5))
}
```

For the syscall/js package, I needed to give a list of symbols that are undefined in any case because the package doesn't mark these functions as imported in any way.

### DIFF
--- a/compiler/symbol.go
+++ b/compiler/symbol.go
@@ -163,18 +163,20 @@ func (c *compilerContext) getFunction(fn *ssa.Function) llvm.Value {
 	// External/exported functions may not retain pointer values.
 	// https://golang.org/cmd/cgo/#hdr-Passing_pointers
 	if info.exported {
-		// Set the wasm-import-module attribute if the function's module is set.
-		if info.module != "" {
-
+		if c.archFamily() == "wasm32" {
 			// We need to add the wasm-import-module and the wasm-import-name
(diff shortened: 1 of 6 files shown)


deadprogram:
I am personally in favor of this change. If any existing user code breaks it was going to probably break at some point already at an unexpected time vs. making an intentional change.

deadprogram:
Any opinions on this @dankegel @dgryski @codefromthecrypt anyone else?

codefromthecrypt:
I agree @anuraaga wdyt about enforcing `//export` at compile time to prevent accidents?

anuraaga:
Sounds like a good idea, personally would expect compilation to fail when forgetting that directive

deadprogram:
Thanks for helping review! Now merging.

PR #3134, opened 2022-09-05
machine: add PWM peripheral comments to pins

aykevl:
These comments will be used in a change to the tinygo.org documentation.


deadprogram:
Thanks @aykevl now merging.

PR #3135, opened 2022-09-06
qtpy-rp2040: set default-stack-size to 4096, remove NEOPIXELS pluralization, remove unconnected LED pin

eyz:


deadprogram (approved):
Thank you @eyz

aykevl:
I don't think we should set the stack size for boards like this. A while ago we decided that we're going to use 2048 byte stacks: https://github.com/tinygo-org/tinygo/pull/1960
I think we should stick to that, and certainly not bump the stack size only for some boards. Instead, I've made a PR so that users can set the stack size the way they want: https://github.com/tinygo-org/tinygo/pull/3159

PR #3137, opened 2022-09-07
examples: adds summary of wasm examples and fixes callback bug

### LINKED ISSUE #2568: Minor inaccuracy in WASM examples README
The README says:

"This examples folder contains two examples that can be built using `make`:
`make export`
`make main`"

 Looking in [Makefile](https://github.com/tinygo-org/tinygo/blob/release/src/examples/wasm/Makefile), it looks like `make` instructions are included for ALL examples (`callback`, `slices`, etc). The folder does include the listed examples, but from the README, developers might assume that `make` instructions for other examples do not exist.

codefromthecrypt:
This adds a summary of each wasm example, as before it was a bit unclear how to do so. This also fixes the callback example which was broken.

Fixes #2568

### DIFF
--- a/src/examples/wasm/README.md
+++ b/src/examples/wasm/README.md
@@ -21,12 +21,14 @@ $ tinygo build -o ./wasm.wasm -target wasm ./main/main.go
 This creates a `wasm.wasm` file, which we can load in JavaScript and execute in
 a browser.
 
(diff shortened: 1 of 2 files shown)


deadprogram:
Thank you @codefromthecrypt now merging.

PR #3139, opened 2022-09-07
avr: adding pin change interrupt handling to atmega328p

mrbell321:
This is what I've come up with for AVR pin change interrupts.

### DIFF
--- a/src/machine/machine_atmega328p.go
+++ b/src/machine/machine_atmega328p.go
@@ -477,14 +477,10 @@ var (
 )
 
 func handlePCINTInterrupts(intr uint8, port *volatile.Register8) {
-	println("Interrupted: ", intr, port)
 	current := port.Get()
-	println("Current: ", current)
 	new := current ^ previous[intr]
-	println("Toggled: ", new)
 	for i := uint8(0); i < 8; i++ {
 		if (current>>i)&0x01 == 0x01 && pinCallbacks[intr][i] != nil {
-			println("Calling: ", intr, i)
 			pinCallbacks[intr][i](Pin(i * (intr + 1)))
 		}
 	}
@@ -504,80 +500,65 @@ func handlePCINT2Interrupts(intr interrupt.Interrupt) {
 }
 
 func SetInterrupt(pin Pin, callback func(Pin)) (err error) {
-	// if p == nil {
-	// 	// Disable this pin interrupt
-	// 	return errors.New("Not implemented")
-	// } else {
-	var PCMSK_REG *volatile.Register8
-	var PCIE uint8
-	var PCINT uint8
-	var IRQ uint8
-	var PIN0 Pin
-	//var PIN *volatile.Register8
-	switch {
-	case pin >= PB0 && pin <= PB7:
-		println("PCMSK0: ", pin)
-		// PCMSK0 - PCINT0-7
-		PCMSK_REG = avr.PCMSK0
-		PCIE = avr.PCICR_PCIE0
-		PIN0 = PD0
-		PCINT = 1 << (pin - PIN0)
-		IRQ = avr.IRQ_PCINT0
-		//PIN = avr.PINB
-		if interrupts[IRQ-avr.IRQ_PCINT0] == nil {
-			println("Registering interrupt: ", IRQ)
-			intr := interrupt.New(avr.IRQ_PCINT0, handlePCINT0Interrupts)
-			interrupts[IRQ-avr.IRQ_PCINT0] = &intr
-		} else {
-			println("interrupt already defined")
-		}
-	case pin >= PC0 && pin <= PC7:
-		println("PCMSK1: ", pin)
-		// PCMSK1 - PCINT8-14
-		PCMSK_REG = avr.PCMSK1
-		PCIE = avr.PCICR_PCIE1
-		PIN0 = PC0
-		PCINT = 1 << (pin - PIN0)
-		IRQ = avr.IRQ_PCINT1
-		//PIN = avr.PINC
-		if interrupts[IRQ-avr.IRQ_PCINT0] == nil {
-			println("Registering interrupt: ", IRQ)
-			intr := interrupt.New(avr.IRQ_PCINT1, handlePCINT1Interrupts)
-			interrupts[IRQ-avr.IRQ_PCINT0] = &intr
-		} else {
-			println("interrupt already defined")
-		}
-	case pin >= PD0 && pin <= PD7:
-		println("PCMSK2: ", pin)
-		// PCMSK2 - PCINT16-23
-		PCMSK_REG = avr.PCMSK2
-		PCIE = avr.PCICR_PCIE2
-		PIN0 = PD0
-		PCINT = 1 << (pin - PIN0)
-		IRQ = avr.IRQ_PCINT2
-		//PIN = avr.PIND
-		if interrupts[IRQ-avr.IRQ_PCINT0] == nil {
-			println("Registering interrupt: ", IRQ)
-			intr := interrupt.New(avr.IRQ_PCINT2, handlePCINT2Interrupts)
-			interrupts[IRQ-avr.IRQ_PCINT0] = &intr
-		} else {
-			println("interrupt already defined")
-		}
-	default:
-		return errors.New("Unsupported pin")
-	}
-
-	if pinCallbacks[IRQ-avr.IRQ_PCINT0][pin-PIN0] == nil {
-		println("Registering callback: ", IRQ-avr.IRQ_PCINT0, pin-PIN0)
-		pinCallbacks[IRQ-avr.IRQ_PCINT0][pin-PIN0] = callback
+	if p == nil {
+		// Disable this pin interrupt
+		return errors.New("Not implemented")
 	} else {
-		return errors.New("callback already defined")
+		var PCMSK_REG *volatile.Register8
+		var PCIE uint8
+		var PCINT uint8
+		var IRQ uint8
+		var PIN0 Pin
+		//var PIN *volatile.Register8
+		switch {
+		case pin >= PB0 && pin <= PB7:
+			// PCMSK0 - PCINT0-7
+			PCMSK_REG = avr.PCMSK0
+			PCIE = avr.PCICR_PCIE0
+			PIN0 = PD0
+			PCINT = 1 << (pin - PIN0)
+			IRQ = avr.IRQ_PCINT0
+			if interrupts[IRQ-avr.IRQ_PCINT0] == nil {
+				intr := interrupt.New(avr.IRQ_PCINT0, handlePCINT0Interrupts)
+				interrupts[IRQ-avr.IRQ_PCINT0] = &intr
+			} else {
+			}
+		case pin >= PC0 && pin <= PC7:
+			// PCMSK1 - PCINT8-14
+			PCMSK_REG = avr.PCMSK1
+			PCIE = avr.PCICR_PCIE1
+			PIN0 = PC0
+			PCINT = 1 << (pin - PIN0)
+			IRQ = avr.IRQ_PCINT1
+			if interrupts[IRQ-avr.IRQ_PCINT0] == nil {
+				intr := interrupt.New(avr.IRQ_PCINT1, handlePCINT1Interrupts)
+				interrupts[IRQ-avr.IRQ_PCINT0] = &intr
+			} else {
+			}
+		case pin >= PD0 && pin <= PD7:
+			// PCMSK2 - PCINT16-23
+			PCMSK_REG = avr.PCMSK2
+			PCIE = avr.PCICR_PCIE2
+			PIN0 = PD0
+			PCINT = 1 << (pin - PIN0)
+			IRQ = avr.IRQ_PCINT2
+			if interrupts[IRQ-avr.IRQ_PCINT0] == nil {
+				intr := interrupt.New(avr.IRQ_PCINT2, handlePCINT2Interrupts)
+				interrupts[IRQ-avr.IRQ_PCINT0] = &intr
+			}
+		default:
+			return errors.New("Unsupported pin")
+		}
+
+		if pinCallbacks[IRQ-avr.IRQ_PCINT0][pin-PIN0] == nil {
+			pinCallbacks[IRQ-avr.IRQ_PCINT0][pin-PIN0] = callback
+		} else {
+			return errors.New("callback already defined")
+		}
+
+		PCMSK_REG.SetBits(PCINT)
+		avr.PCICR.SetBits(PCIE)
+
+		return nil
 	}
-
-	println("Setting PCMSK_REG to PCINT: ", PCMSK_REG, PCINT)
-	PCMSK_REG.SetBits(PCINT)
-	println("Setting PCICR to PCIE:", avr.PCICR, PCIE)
-	avr.PCICR.SetBits(PCIE)
-
-	return nil
 }
--- a/src/machine/machine_atmega328p.go
+++ b/src/machine/machine_atmega328p.go
@@ -5,6 +5,7 @@ package machine
 
 import (
 	"device/avr"
+	"errors"
 	"runtime/interrupt"
 	"runtime/volatile"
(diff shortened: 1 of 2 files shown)


src/machine/machine_atmega328p.go
```diff
@@ -467,3 +468,116 @@ var SPI0 = SPI{
...
+
+// Pin Change Interrupts
+var (
+	pinCallbacks [3][8]func(Pin)
+	interrupts   [3]*interrupt.Interrupt
+	previous     [3]uint8
+)
+
+func handlePCINTInterrupts(intr uint8, port *volatile.Register8) {
+	println("Interrupted: ", intr, port)
```
aykevl: This is debug code. Please remove it.

src/machine/machine_atmega328p.go
```diff
@@ -467,3 +468,116 @@ var SPI0 = SPI{
...
+
+func handlePCINT2Interrupts(intr interrupt.Interrupt) {
+	handlePCINTInterrupts(avr.IRQ_PCINT2-avr.IRQ_PCINT0, avr.PIND)
+}
+
+func SetInterrupt(pin Pin, callback func(Pin)) (err error) {
+	// if p == nil {
+	// 	// Disable this pin interrupt
+	// 	return errors.New("Not implemented")
+	// } else {
```
aykevl: Looks like old code left over after experimenting?

aykevl (commented):
Thanks, we need this! However, this PR is not ready for review: it contains tons of debug code. Can you please clean up the PR?

src/machine/machine_atmega328p.go
mrbell321: Oh, yeah, will do.

mrbell321: I've removed all the debug prints

src/machine/machine_atmega328p.go
mrbell321: I uncommented this because I think it's important to consider, but it's still not "implemented". I'm not sure if anyone ever actually removes pin change interrupt handlers after enabling them. It's pretty heavy-weight for debouncing.

mrbell321:
I created a ticket for this work too: https://github.com/tinygo-org/tinygo/issues/3145
Not sure what the official ticket/git tagging mechanism is...

src/machine/machine_atmega328p.go
```diff
@@ -467,3 +468,96 @@ var SPI0 = SPI{
...
+	if callback == nil {
+		// Disable this pin interrupt
+		return errors.New("Not implemented")
+	} else {
+		var PCMSK_REG *volatile.Register8
+		var PCIE uint8
+		var PCINT uint8
+		var IRQ uint8
+		var PIN0 Pin
+		//var PIN *volatile.Register8
```
aykevl: Here is some more unused code.

src/machine/machine_atmega328p.go
```diff
@@ -467,3 +468,96 @@ var SPI0 = SPI{
...
+
+func handlePCINT2Interrupts(intr interrupt.Interrupt) {
+	handlePCINTInterrupts(avr.IRQ_PCINT2-avr.IRQ_PCINT0, avr.PIND)
+}
+
+func SetInterrupt(pin Pin, callback func(Pin)) (err error) {
+	if callback == nil {
+		// Disable this pin interrupt
+		return errors.New("Not implemented")
+	} else {
```
aykevl: `errors.New` causes a heap allocation, which is not a good idea in this code. Instead, please define this error in a global.
Or better yet: add it to src/machine/machine.go.

There are two other `errors.New` calls below that should be fixed in a similar way (but perhaps not all by defining them in src/machine/machine.go).

Also: please indent this code according to https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow.

src/machine/machine_atmega328p.go
```diff
@@ -467,3 +468,96 @@ var SPI0 = SPI{
...
+			PCMSK_REG = avr.PCMSK0
+			PCIE = avr.PCICR_PCIE0
+			PIN0 = PD0
+			PCINT = 1 << (pin - PIN0)
+			IRQ = avr.IRQ_PCINT0
+			if interrupts[IRQ-avr.IRQ_PCINT0] == nil {
+				intr := interrupt.New(avr.IRQ_PCINT0, handlePCINT0Interrupts)
+				interrupts[IRQ-avr.IRQ_PCINT0] = &intr
+			} else {
+			}
```
aykevl: empty `else`

src/machine/machine_atmega328p.go
mrbell321: Fixed it. What's the convention for "Resolve conversation" here? 
I've worked on projects where the person who opens it should resolve and ones where the person who makes the fix resolves it...

src/machine/machine_atmega328p.go
mrbell321: Ah, good point. Addressed.

src/machine/machine_atmega328p.go
mrbell321: Addressed.

src/machine/machine_atmega328p.go
aykevl: We don't really have a convention here. But if a comment is addressed I would consider that resolved (either by fixing it as requested or explaining why it's the way it is - there may be a good reason the reviewer didn't see!)

src/machine/machine_atmega328p.go
```diff
@@ -467,3 +472,93 @@ var SPI0 = SPI{
 	sdo:  PB3,
 	sdi:  PB4,
 	cs:   PB2}
+
+// Pin Change Interrupts
+var (
+	pinCallbacks [3][8]func(Pin)
+	interrupts   [3]*interrupt.Interrupt
+	previous     [3]uint8
```
aykevl: It looks like `previous` is  never used?

src/machine/machine_atmega328p.go
```diff
@@ -467,3 +472,93 @@ var SPI0 = SPI{
...
+		// PCMSK0 - PCINT0-7
+		PCMSK_REG = avr.PCMSK0
+		PCIE = avr.PCICR_PCIE0
+		PIN0 = PD0
+		PCINT = 1 << (pin - PIN0)
+		IRQ = avr.IRQ_PCINT0
+		if interrupts[IRQ-avr.IRQ_PCINT0] == nil {
+			intr := interrupt.New(avr.IRQ_PCINT0, handlePCINT0Interrupts)
+			interrupts[IRQ-avr.IRQ_PCINT0] = &intr
+		}
```
aykevl: `interrupt.New` is a bit of a magical function and doesn't do anything at runtime (only at compile time). Therefore, I suggest using the following:

```suggestion
		interrupt.New(avr.IRQ_PCINT0, handlePCINT0Interrupts)
```

(same for the other 2 interrupts).

src/machine/machine_atmega328p.go
```diff
@@ -467,3 +472,93 @@ var SPI0 = SPI{
...
+		if interrupts[IRQ-avr.IRQ_PCINT0] == nil {
+			intr := interrupt.New(avr.IRQ_PCINT2, handlePCINT2Interrupts)
+			interrupts[IRQ-avr.IRQ_PCINT0] = &intr
+		}
+	default:
+		return ErrInvalidInputPin
+	}
+
+	if pinCallbacks[IRQ-avr.IRQ_PCINT0][pin-PIN0] != nil {
+		return ErrCallbackAlreadyDefined
```
aykevl: It depends on the chip, but at least on the nrf a new call to `SetInterrupt` simply overwrites the old interrupt. (There is a risk of a race condition in the write though, so perhaps better to leave it as-is).

src/machine/machine_atmega328p.go
```diff
@@ -467,3 +472,93 @@ var SPI0 = SPI{
...
+	var PIN0 Pin
+
+	switch {
+	case pin >= PB0 && pin <= PB7:
+		// PCMSK0 - PCINT0-7
+		PCMSK_REG = avr.PCMSK0
+		PCIE = avr.PCICR_PCIE0
+		PIN0 = PD0
+		PCINT = 1 << (pin - PIN0)
+		IRQ = avr.IRQ_PCINT0
```
aykevl: You could consider the following:

```suggestion
		pinIndex := pin - PD0
		pinCallbacks[0][pinIndex] = callback
		avr.PCMSK0.SetBits(1 << pinIndex)
		avr.PCICR.SetBits(avr.PCICR_PCIE0)
```

It's less abstract, but it's actually a bit shorter and perhaps more readable as well. Also, I think the compiler will be better able to optimize this code.

aykevl (commented):
This looks good, with some nits here and there that could improve the code a bit and make it more efficient. Thank you for your continued work on this PR!

src/machine/machine_atmega328p.go
```diff
@@ -467,3 +472,93 @@ var SPI0 = SPI{
...
+
+func handlePCINT1Interrupts(intr interrupt.Interrupt) {
+	handlePCINTInterrupts(avr.IRQ_PCINT1-avr.IRQ_PCINT0, avr.PINC)
+}
+
+func handlePCINT2Interrupts(intr interrupt.Interrupt) {
+	handlePCINTInterrupts(avr.IRQ_PCINT2-avr.IRQ_PCINT0, avr.PIND)
+}
+
+func SetInterrupt(pin Pin, callback func(Pin)) (err error) {
```
aykevl: Actually, I somehow missed this but this doesn't follow the existing API. Can you update the API to follow the standard API? See https://tinygo.org/docs/reference/machine/

```suggestion
func (pin Pin) SetInterrupt(change PinChange, callback func(Pin)) (err error) {
```

src/machine/machine_atmega328p.go
mrbell321: Ah interesting...

src/machine/machine_atmega328p.go
mrbell321: I don't remember where I saw this idea, but I was translating from something... either one of the other tinygo architectures or something in the arduino framework.
I can see the possibility of wanting to change the interrupt handler at runtime, but I don't personally have a use case, so I thought I'd leave it like this until there's a good reason to implement.

src/machine/machine_atmega328p.go
mrbell321: Hm, sure, I should follow the documented standard.
But I pretty sure I copied this from one of the other architectures tho, so I think there's some non-conformity elsewhere too.
...or I just messed up.

src/machine/machine_atmega328p.go
mrbell321: Ah, yeah, holdover from my _previous_  iteration... :-D

src/machine/machine_atmega328p.go
mrbell321: Hm, why did I think I needed to track all of the interrupts in an array?

src/machine/machine_atmega328p.go
mrbell321: Yeah, I was intentionally being abstract because... well... reasons.
However, it ended up not working out very well that way due to... other reasons. 
This new way simplifies quite a bit so I'll have the MR updated soon.

...I also need to figure out why "make release" is now crashing my machine....

src/machine/machine_atmega328p.go
mrbell321: Aah, well part of it, at least is that the atmega328p doesn't distinguish rising v falling edges on these Port Pin change interrupts, so 'PinChange' is basically meaningless.
But that's OK

aykevl: > But I pretty sure I copied this from one of the other architectures tho

I grepped through the source code and can't find a `SetInterrupt` that is not a member of a `Pin` (or `CAN`) object.
In any case, eventually we should have #3170 and improvements on it that should check for these kinds of inconsistencies.

aykevl: > at least is that the atmega328p doesn't distinguish rising v falling edges on these Port Pin change interrupts, so 'PinChange' is basically meaningless.

Right. In that case, I think only `PinToggle` should be defined (leaving `PinRising` and `PinFalling` undefined will make sure programs fail to compile when used). You could perhaps also check it in `SetInterrupt` but if only `PinToggle` exists it shouldn't really be necessary.

src/machine/machine_atmega328p.go
```diff
@@ -467,3 +472,71 @@ var SPI0 = SPI{
...
+
+func handlePCINT2Interrupts(intr interrupt.Interrupt) {
+	handlePCINTInterrupts(avr.IRQ_PCINT2-avr.IRQ_PCINT0, avr.PIND)
+}
+
+type PinChange uint8
+
+// Pin change interrupt constants for SetInterrupt.
+const (
+	PinBoth PinChange = nil
```
aykevl: The standard name for this is `PinToggle` (because it triggers on every toggle between low and high). See https://tinygo.org/docs/reference/machine/.

(Also, did you test this code? I don't think you can assign `nil` to a uint8).

aykevl (commented):
Just one last thing

src/machine/machine_atmega328p.go
mrbell321: I haven't tested it yet because `make release` is crashing my system at the moment. Just started yesterday. It's probably unrelated to _specifically_ to `make release`, because the first crash I had yesterday seemed to be related to Outlook. 
But after that, `make release` reliably crashes my computer. I probably need to reimage it, which I'm not looking forward too, but pushing up to GitHub ensures I don't lose this code.

I promise not to merge until I get that figured out and actually test it.

src/machine/machine_atmega328p.go
mrbell321: So, I've `make release`, at least to the point that it runs to completion by ripping out all of tinygo and golang and reinstalling them.
However, I now get
```
$ tinygo build -target=arduino   
warning: no avr-gcc installation can be found on the system, cannot link standard libraries
warning: standard library not linked and so no interrupt vector table or compiler runtime routines will be linked
warning: no avr-gcc installation can be found on the system, cannot link standard libraries
warning: standard library not linked and so no interrupt vector table or compiler runtime routines will be linked
```
However, I have avr-gcc
```
$ avr-gcc
avr-gcc: fatal error: no input files
compilation terminated.
```

deadprogram: Please see https://tinygo.org/getting-started/install/linux/#avr-eg-arduino-uno for the install requirements for AVR on Linux.

deadprogram:
So @mrbell321 has this been tested on real hardware?

mrbell321:
I’m on macOS not Linux and I’ve been through the process a couple times.
Not the only one having the issue either. There’s a slack thread.

mrbell321:
The early version, yes, but there have been changes I cannot test yet due
to the above toolchain issues.

src/machine/machine_atmega328p.go
```diff
@@ -467,3 +472,71 @@ var SPI0 = SPI{
...
+func (pin Pin) SetInterrupt(pinChange PinChange, callback func(Pin)) (err error) {
+	if callback == nil {
+		// Disable this pin interrupt
+		return ErrFeatureNotImplemented
+	}
+
+	switch {
+	case pin >= PB0 && pin <= PB7:
+		// PCMSK0 - PCINT0-7
+		pinIndex := pin - PD0
```
ysoldak, line 518:
```suggestion
		pinIndex := pin - PB0
```

ysoldak:
I've tried this change on a custom board with atmega328p chip.

Interrupt fires. Can't use `irremote` driver directly though, since this change misses `PinRising` and `PinFalling` states.
I'm not exactly sure how, but handling of the state shall be possible to add, see for example https://www.avrfreaks.net/s/topic/a5C3l000000URmcEAG/t120592

ysoldak:
This can be closed now, @deadprogram ?

deadprogram:
@ysoldak yes.

@mrbell321 thank you for pointing the way on the implementation. Now closing this PR.

PR #3140, opened 2022-09-08
build: uses setup-go@v3 to simplify cache config

codefromthecrypt:
This updates to setup-go@v3 which is the only updated version (v2 last
updated in feb), and employs its cache to simplify workflow
configuration.

Notably, we can't do this for Alpine until https://github.com/tinygo-org/tinygo/issues/3146

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -34,6 +34,7 @@ jobs:
         with:
           submodules: true
       - name: Install Go
+        # TODO: v2 until https://github.com/tinygo-org/tinygo/issues/3146
         uses: actions/setup-go@v2
         with:
           go-version: '1.18.1'
@@ -121,9 +122,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
       - name: Install Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v3
         with:
           go-version: '1.18.1'
+          cache: true
       - name: Install wasmtime
         run: |
           curl https://wasmtime.dev/install.sh -sSf | bash
@@ -173,9 +175,10 @@ jobs:
               simavr \
               ninja-build
       - name: Install Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v3
         with:
           go-version: '1.18.1'
+          cache: true
       - name: Install Node.js
         uses: actions/setup-node@v2
         with:
@@ -184,13 +187,6 @@ jobs:
         run: |
           curl https://wasmtime.dev/install.sh -sSf | bash
           echo "$HOME/.wasmtime/bin" >> $GITHUB_PATH
-      - name: Cache Go
-        uses: actions/cache@v3
-        with:
-          key: go-cache-linux-asserts-v1-${{ hashFiles('go.mod') }}
-          path: |
-            ~/.cache/go-build
-            ~/go/pkg/mod
       - name: Cache LLVM source
         uses: actions/cache@v3
         id: cache-llvm-source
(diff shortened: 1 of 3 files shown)


.github/workflows/linux.yml
```diff
@@ -30,22 +30,24 @@ jobs:
...
         run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           submodules: true
       - name: Install Go
         uses: actions/setup-go@v2
+        # TODO: with go-version-file and cache
+        # See https://github.com/actions/setup-go/issues/263
```
codefromthecrypt: I'll watch this one and revisit towards whatever end

.github/workflows/linux.yml
```diff
@@ -30,22 +30,24 @@ jobs:
...
         with:
-          go-version: '1.18.1'
+          go-version: '1.18'
```
aykevl: This is better. Apparently, GitHub Actions treats "1.18" as "the latest version of 1.18.x", which is what we want.

aykevl (changes requested):
Thank you for fixing this inconsistency! Unfortunately, it doesn't work in practice because the version number in go.mod specifies the _oldest_ supported version while we should compile with the _newest_ supported version. This is necessary for example when new Go language features are introduced: we want to compile with those new language features.
(Go 1.19 is almost entirely supported - we should fix the last bug and finally switch to it!)

Also, why did you change all these version numbers from v2 to v3? Do they change anything we need? (I don't like updating things just for the sake of updating - it causes noise and may break things unnecessarily).

codefromthecrypt:
@aykevl just so I know how to proceed, maybe I misunderstood you. You think that updating to latest CI versions is not a good idea? I may have misread, but it seems you prefer to use outdated versions?

aykevl:
> You think that updating to latest CI versions is not a good idea? I may have misread, but it seems you prefer to use outdated versions?

If there are any improvements, sure. But if there is no practical improvement, I'm not so sure.
I'm also not much against it. If you want to change it, I'd prefer this in a separate PR or at least separate commit so this PR becomes easier to review. (In this case it's not too much of a problem but I prefer independent changes to go in separate commits).

codefromthecrypt:
ok I don't mind getting my commit count higher :D generally, I prefer to update as then you find the bugs before you *must* update. Finding things out late makes for unnecessary emergencies.

back to the version, if we want to control separately, we can make a file that "1.18" in it https://github.com/actions/setup-go/blob/268d8c0ca0432bb2cf416faae41297df9d262d7f/src/installer.ts#L316-L325

should I make a file named .go-version? That way it is centralized and not required to be the same as go.mod. Just we have to remember to update it. wdyt?

codefromthecrypt:
spun out actions/cache bump here. https://github.com/tinygo-org/tinygo/pull/3141
I'll spin another out to be consistent on latest go version 1.18
then, we can get back to this (how to centralize it)

aykevl:
To be honest, I'm not sure centralizing the version number helps. I've never found it burdensome to update the version number by hand (it's something is usually done by me, and it only happens twice a year), and in fact it can be very useful to do it manually as there are sometimes issues on particular platforms. For example, IIRC we had a case where Windows couldn't use the latest Go version so we delayed it for a bit. This is harder with a centralized version number.

That said, using the latest patch version of Go is definitely a good idea so we get all the latest bug fixes.

codefromthecrypt:
I think at least we can centralize it per file. I've done that before and can raise a PR for review

codefromthecrypt:
I'll switch this one to a partial upgrade to actions/setup-go@v3 noting the gap https://github.com/tinygo-org/tinygo/issues/3146 which is a bit of a time bomb, but we have time.

codefromthecrypt:
@aykevl PTAL I reduced this to only updating the setup-go action to the currently maintained one, except where we can't. I left the go versions alone as that's being done in #3143

deadprogram (approved):
Thanks for the updates/simplifications @codefromthecrypt

PR #3141, opened 2022-09-08
build: updates to actions/cache@v3

codefromthecrypt:
v2 has bugs and hasn't had a release since last November.

See https://github.com/actions/cache/releases


aykevl (approved):
LGTM. Seems like the only reason they bumped the major version number to v3 is because of needing Node.js 16.

aykevl:
> v2 has bugs and hasn't had a release since last November.

I agree this is a good reason to update.

deadprogram:
Thanks @codefromthecrypt now merging.

PR #3144, opened 2022-09-08
machine/rp2040: implement semi-random RNG based on ROSC

deadprogram:
This PR implements `GetRNG()` for the rp2040 by using a semi-random RNG based on the `ROSC` ring oscillator in the same way that the pico-sdk does.

See https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/pico_lwip/random.c

### DIFF
--- a/src/machine/machine_rp2040_rng.go
+++ b/src/machine/machine_rp2040_rng.go
@@ -1,0 +1,38 @@
+//go:build rp2040
+// +build rp2040
+
+// Implementation based on code located here:
+// https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/pico_lwip/random.c
+
+package machine
+
+import (
+	"device/rp"
+)
+
+const numberOfCycles = 32
+
+// GetRNG returns 32 bits of semi-random data based on ring oscillator.
+func GetRNG() (uint32, error) {
+	var val uint32
+	for i := 0; i < 4; i++ {
+		val = (val << 8) | uint32(roscRandByte())
+	}
+	return val, nil
+}
+
+var randomByte uint8
+
+func roscRandByte() uint8 {
+	var poly uint8
+	for i := 0; i < numberOfCycles; i++ {
+		if randomByte&0x80 != 0 {
+			poly = 0x35
+		} else {
+			poly = 0
+		}
+		randomByte = ((randomByte << 1) | uint8(rp.ROSC.GetRANDOMBIT()) ^ poly)
+		// TODO: delay a little because the random bit is a little slow
+	}
+	return randomByte
+}
--- a/src/crypto/rand/rand_baremetal.go
+++ b/src/crypto/rand/rand_baremetal.go
@@ -1,5 +1,5 @@
-//go:build stm32 || (sam && atsamd51) || (sam && atsame5x)
-// +build stm32 sam,atsamd51 sam,atsame5x
+//go:build stm32 || (sam && atsamd51) || (sam && atsame5x) || rp2040
+// +build stm32 sam,atsamd51 sam,atsame5x rp2040
 
 package rand
 


deadprogram:
Currently working as well as a pseudo-random number can be:

```
Terminal ready
d9121ab984aba30dcd7c7d8a42bcbdeb1ee50e452487a349262686cbd2d1bdf4
14f56b22f358b1a3b9c388b21686ec97f4bf643c6e902a2c1d81cde13e9b3913
cb3b8d2d5f9981e9905c9c4bcfd55c8ff5b6041253ed0ffdad5b2af6f4d9a129
2bf748ce51f48857e96ec6ed679328f7e13f37db7d2ff23e9938f41ba04029de
247ffe4bdaf25f5ae194195cf038bb70d6b531f0a4ce08e344cc0b3022a4ca27
bade2b4e3883a6ae6925ba3740517c2215c75316f74dbda12c2deb59d2b4ef6b
3618c3d8b319578e66c30e97e6f126c63612f94c03d0c638517f11589f9a9112
```

src/machine/machine_rp2040_rng.go
```diff
@@ -0,0 +1,36 @@
...
+func GetRNG() (uint32, error) {
+	var val uint32
+	for i := 0; i < 4; i++ {
+		val = (val << 8) | uint32(roscRandByte())
+	}
+	return val, nil
+}
+
+func roscRandByte() uint8 {
+	var val, poly uint8
```
dgryski: The reference code has `val` being a static byte that's persistent across calls.  Did you want to make this a global to achieve the same effect?

deadprogram: Changed, because that is in fact what I want.

Thanks for pointing that out @dgryski

src/machine/machine_rp2040_rng.go
```diff
@@ -0,0 +1,38 @@
...
+// GetRNG returns 32 bits of semi-random data based on ring oscillator.
+func GetRNG() (uint32, error) {
+	var val uint32
+	for i := 0; i < 4; i++ {
+		val = (val << 8) | uint32(roscRandByte())
+	}
+	return val, nil
+}
+
+var val uint8
```
dgryski: Given this is now a global in the `machine` package, does it deserve a better name?

deadprogram: Renamed.

dgryski (approved):
LGTM

deadprogram:
Thanks for the reviews @dgryski

aykevl:
I don't think we should do this, because it is not cryptographically secure:

> If the system clocks are running from the XOSC and/or PLLs the ROSC can be used to generate random numbers. Simply enable the ROSC and read the RANDOMBIT register to get a 1-bit random number and read it n times to get an n-bit value. This does not meet the requirements of randomness for security systems because it can be compromised, but it may be useful in less critical applications. If the cores are running from the ROSC then the value will not be random because the timing of the register read will be correlated to the phase of the ROSC.

In my opinion, the crypto/rand package must _always_ have cryptographically secure random numbers. Here the documentation clearly states that they are not cryptographically secure. Therefore, I don't agree with its use in crypto/rand. One of the most frequent security failures is not using cryptographically secure random numbers, and we'd make it very easy to make this mistake here.

Alternatively, we can make the GetRNG() function available but simply not use it in crypto/rand.

deadprogram:
We are currently supporting urandom which is also only a pseudo-random number generator AFAIK.

Perhaps we can leave it how it is now in but document that it is not a TRNG? Display a warning during compilation?

aykevl:
> We are currently supporting urandom which is also only a pseudo-random number generator AFAIK.

It is pseud-random, yes, but it is cryptographically secure.

> Perhaps we can leave it how it is now in but document that it is not a TRNG?

Nobody will read that. It would mean that crypto/rand in regular Go is good to use for cryptography, so people will use it, and when porting to TinyGo will unknowingly use insecure random numbers.

(Sorry for my somewhat harsh tone here, I care deeply about security and want to avoid potential issues).

deadprogram:
See https://github.com/micropython/micropython/commit/6f06dcaee57cc6f36a4121b4797942200b2d7281

aykevl:
> See [micropython/micropython@6f06dca](https://github.com/micropython/micropython/commit/6f06dcaee57cc6f36a4121b4797942200b2d7281)

What do you mean by that?

> The output of uos.urandom() performs well with the NIST800-22 test suite. In my trial it passed all tests of the sts 2.1.2 test suite.  I also ran a test of the random data with the Common Criteria test suite AIS 31, and it passed all tests too.

This seems to imply it is safe... but I'm not a cryptographer so I don't feel confident saying it is secure.

deadprogram:
There is a lot more conversation about the rp2040 implementation here: https://github.com/micropython/micropython/issues/7031

In any case, we can provide an example function for people who want to hook it up anyhow, with the appropriate caveats in that example. That way I suppose it is not hooked up by default for the unwary,

Out of curiosity, any opinions on all this @dgryski ?

PR #3153, opened 2022-09-11
machine/nrf52840: implement RNG peripheral operation

schultz-is:
Howdy y'all. This adds RNG support for Nordic's nRF52840. Alongside the `GetRNG` function used by `crypto/rand/rand_baremetal.go` I've added some functions to control and configure the RNG peripheral. There's also an ARM CryptoCell 310 subsystem on the machine that contains another TRNG, but interaction with it isn't via an easy register interface like this.

Snippet of sample output from `src/examples/rand` flashed to a `xiao-ble`:
```
e73ba017e959eef43a0307379d595245b114418ba0243e82ba6cf3860830254d
e6825f781dd1a401d9a75afadccce9133e30a27d82193f45cc6353d505e85ed7
51e9e198c65dbbae7a1b37861a409d7d78301467c348197e595d765e71b1fe4c
a066ed186aa7cc4798f68b0cc3fee04d08d4046fe6faa33b9a6ea0db51f87e0e
40447c2d0bcde76a47424b2710baa7e5b1339f7c6c3a322ff275cf5f6f653047
```

### DIFF
--- a/src/crypto/rand/rand_baremetal.go
+++ b/src/crypto/rand/rand_baremetal.go
@@ -1,5 +1,5 @@
-//go:build stm32 || (sam && atsamd51) || (sam && atsame5x) || rp2040
-// +build stm32 sam,atsamd51 sam,atsame5x rp2040
+//go:build nrf52840 || stm32 || (sam && atsamd51) || (sam && atsame5x) || rp2040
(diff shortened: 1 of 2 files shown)


deadprogram (approved):
Looking good @schultz-is thank you!

Issue #3146, opened 2022-09-09
ci: alpine build actions/setup-go is revlocked to v2


codefromthecrypt:
Per https://github.com/actions/setup-go/issues/263 something needs to change in [build-linux](https://github.com/tinygo-org/tinygo/blob/release/.github/workflows/linux.yml#L15) or we won't be able to get fixes or features in actions/setup-go@v3, notably cache integration which simplifies things. actions/setup-go@v2 currently still works, but was last released in feb 2022.

I think the various options are:
* do nothing until it breaks (current approach)
* workaround the GOROOT issue (unsure it will work)
* install go on alpine a different way
* switch off alpine
* your choice here

aykevl:
Here is my attempt to fix this: https://github.com/tinygo-org/tinygo/pull/3158

PR #3158, opened 2022-09-14
ci: use the Go Alpine image for building the release binary

aykevl:
This fixes https://github.com/tinygo-org/tinygo/issues/3146 by using a prebuilt Docker image. I don't remember why I used `setup-go` but probably to make it faster (setup-go usually uses cached binaries).


codefromthecrypt (approved):
looks good! the docker go version will be more or less latest patch, which is what the actions would intend to do.

deadprogram:
Good one @aykevl and thanks for review @codefromthecrypt now merging.

PR #3160, opened 2022-09-15
rp2040: do not use GetRNG in crypto/rand

aykevl:
The crypto/rand package is used for sensitive cryptographic operations. Do not use the rp2040 RNG for this purpose, because it's not strong enough for cryptography.

I think it is _possible_ to make use of the RP2040 RNG to create cryptographically secure pseudo-random numbers, but it needs some entropy calculation and secure hashing (blake2s or so) to make them truly unpredictable.

See the discussion in #3144


deadprogram (approved):
Removing as default seems like the safest option.

PR #3162, opened 2022-09-15
wasm: improve malloc/free heap tracking

aykevl:
Using a slice requires a lot less in code size than a map. This is visible when compiling a very small "hello world" style program.

Before tracking memory in malloc/free:  2873 bytes
With tracking using a map:              6551 bytes
With a slice instead of a map:         3532 bytes

Of course, most of this code size increase won't be visible with https://github.com/tinygo-org/tinygo/pull/3142, but it's still a saving of around 3kB in this minimal example.

### DIFF
--- a/src/runtime/arch_tinygowasm.go
+++ b/src/runtime/arch_tinygowasm.go
@@ -62,53 +62,63 @@ func growHeap() bool {
 }
 
 // The below functions override the default allocator of wasi-libc. This ensures
-// code linked from other languages can allocate memory without colliding with
-// our GC allocations.
+// code linked from other languages can allocate memory on the same heap as the
+// TinyGo heap.
 
-var allocs = make(map[uintptr][]byte)
+// Keep track of all the heap allocations while they're in use.
+// This is not the most efficient solution but it costs a lot less in code size
+// compared to a map.
+var allocs []unsafe.Pointer
+
+func trackAlloc(ptr unsafe.Pointer) {
+	// Try to find some empty space in the allocs slice.
+	for i, slot := range allocs {
+		if slot == nil {
+			allocs[i] = slot
+			return
+		}
+	}
+	// Couldn't find this space. Fall back to appending to the end.
+	allocs = append(allocs, ptr)
+}
+
+func removeAlloc(ptr unsafe.Pointer) {
+	// Remove the pointer so it can be garbage collected.
+	for i, slot := range allocs {
+		if ptr == slot {
+			allocs[i] = nil
+			return
+		}
+	}
+}
 
 //export malloc
 func libc_malloc(size uintptr) unsafe.Pointer {
-	buf := make([]byte, size)
-	ptr := unsafe.Pointer(&buf[0])
-	allocs[uintptr(ptr)] = buf
+	ptr := alloc(size, nil)
+	trackAlloc(ptr)
 	return ptr
 }
 
 //export free
 func libc_free(ptr unsafe.Pointer) {
-	if ptr == nil {
-		return
-	}
-	if _, ok := allocs[uintptr(ptr)]; ok {
-		delete(allocs, uintptr(ptr))
-	} else {
-		panic("free: invalid pointer")
-	}
+	removeAlloc(ptr)
+	free(ptr)
 }
 
 //export calloc
 func libc_calloc(nmemb, size uintptr) unsafe.Pointer {
-	// No difference between calloc and malloc.
+	// Note: we could be even more correct here and check that nmemb * size
+	// doesn't overflow. However the current implementation should normally work
+	// fine.
 	return libc_malloc(nmemb * size)
 }
 
 //export realloc
 func libc_realloc(oldPtr unsafe.Pointer, size uintptr) unsafe.Pointer {
-	// It's hard to optimize this to expand the current buffer with our GC, but
-	// it is theoretically possible. For now, just always allocate fresh.
-	buf := make([]byte, size)
-
-	if oldPtr != nil {
-		if oldBuf, ok := allocs[uintptr(oldPtr)]; ok {
-			copy(buf, oldBuf)
-			delete(allocs, uintptr(oldPtr))
-		} else {
-			panic("realloc: invalid pointer")
-		}
+	newPtr := realloc(oldPtr, size)
+	if newPtr != oldPtr {
+		removeAlloc(oldPtr)
+		trackAlloc(newPtr)
 	}
-
-	ptr := unsafe.Pointer(&buf[0])
-	allocs[uintptr(ptr)] = buf
-	return ptr
+	return newPtr
 }
--- a/tests/runtime_wasi/malloc_test.go
+++ b/tests/runtime_wasi/malloc_test.go
@@ -23,13 +23,13 @@ func libc_calloc(nmemb, size uintptr) unsafe.Pointer
 //export realloc
 func libc_realloc(ptr unsafe.Pointer, size uintptr) unsafe.Pointer
 
-func getFilledBuffer_malloc() uintptr {
+func getFilledBuffer_malloc() unsafe.Pointer {
 	ptr := libc_malloc(5)
 	fillPanda(ptr)
-	return uintptr(ptr)
+	return ptr
 }
 
-func getFilledBuffer_calloc() uintptr {
+func getFilledBuffer_calloc() unsafe.Pointer {
 	ptr := libc_calloc(2, 5)
 	fillPanda(ptr)
 	*(*byte)(unsafe.Add(ptr, 5)) = 'b'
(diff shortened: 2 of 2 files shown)


dgryski:
Should we be concerned about the change from `O(1)` to `O(n)` time for tracking allocations?

aykevl:
Perhaps. But the GC itself is also rather inefficient. We could of course switch this algorithm to an algorithm with the same complexity as the one in the GC (which is probably only marginally better).

dgryski:
Fair enough.  Let's leave the simple implementation for now and if it becomes an issue we can maybe make a build-tag to drop in a map-based implementation for people who care about it.

src/runtime/arch_tinygowasm.go
```diff
@@ -62,53 +62,63 @@ func growHeap() bool {
...
 
-var allocs = make(map[uintptr][]byte)
+func removeAlloc(ptr unsafe.Pointer) {
+	// Remove the pointer so it can be garbage collected.
+	for i, slot := range allocs {
+		if ptr == slot {
+			allocs[i] = nil
+			return
+		}
+	}
```
dgryski, line 92: Do we want to complain if we can't find the pointer?

src/runtime/arch_tinygowasm.go
```diff
@@ -62,53 +62,63 @@ func growHeap() bool {
...
-	if ptr == nil {
-		return
-	}
-	if _, ok := allocs[uintptr(ptr)]; ok {
-		delete(allocs, uintptr(ptr))
-	} else {
-		panic("free: invalid pointer")
-	}
+	removeAlloc(ptr)
+	free(ptr)
```
dgryski, line 105: Just want to confirm that with the `//export free` line, calling `free()` here still does the right thing.

dgryski (approved):
Two questions, but otherwise LGTM.

deadprogram:
Looks like this might be an actual error? https://github.com/tinygo-org/tinygo/actions/runs/3059652993/jobs/4943124870#step:9:58

aykevl:
Yes, that looks like a real bug. I'm glad #3148 added some tests!

src/runtime/arch_tinygowasm.go
aykevl: Yes, it calls the function `free` in gc_conservative.go. The `//export` line does not affect the namespace of the Go package.

src/runtime/arch_tinygowasm.go
aykevl: We could, but it's undefined behavior to call `free` on an invalid pointer so we're not required to.
But of course we could do it when `gcAsserts` is true for example.

src/runtime/arch_tinygowasm.go
```diff
@@ -62,53 +62,63 @@ func growHeap() bool {
...
+		}
+	}
+	// Couldn't find this space. Fall back to appending to the end.
+	allocs = append(allocs, ptr)
+}
 
-var allocs = make(map[uintptr][]byte)
+func removeAlloc(ptr unsafe.Pointer) {
+	// Remove the pointer so it can be garbage collected.
+	for i, slot := range allocs {
```
anuraaga, line 87: Consider iterating in reverse order, complete heuristic but I think newer pointers tend to be freed sooner than older pointers in real code.

anuraaga (commented):
Is it possible to reference the `-opt` flag and pick between slice and map? We'll be executing a lot of logic compiled from C++ or Rust and its unclear what pattern the malloc / free would take. I understand the code size savings is important, but having the chance of pathological behavior makes it scary to run any code compiled with TinyGo doing polyglot. If the slice could be used with `-opt=s` or `z`, and map used with `2`, that could allow both niches to be satisfied with opts that already reflect those semantics.

codefromthecrypt:
I think the tension here, correct me if I'm wrong, is basically you can't override this (a part of how GC works) with build tags until tinygo adds them. chicken egg. Then, the next tension is proving the performance is worse vs having the next impl prove it isn't worse.

Strings are routinely used in plugins. ex regular expression matches etc. It would help in the future balance size/perf in a less conjecturey way. Ex place a representative main.go here that exports regex functions, and benchmark using them. I think that would help balance some discussions in general. A bias towards size can exist with guards against severe performance regressions.

my 2p

aykevl:
> Is it possible to reference the -opt flag and pick between slice and map?

It is possible, but wouldn't help much when it comes to algorithmic complexity. The heap allocator itself is also O(n) in the worst case, where n is the number of blocks (heap size divided by 16, usually). This is why I expect this PR won't have a lot of effect in practice.

Generally when it comes to performance, I want to act on data. In this case, I do have data on code size (because it is easy to measure) but I do not have data on performance. We can always change the implementation if benchmarks suggest this is a bottleneck.

anuraaga:
> Generally when it comes to performance, I want to act on data. In this case, I do have data on code size (because it is easy to measure) but I do not have data on performance.

I do agree with the sentiment but feel that if there are clear pathological cases then those are also similarly clear.

But I didn't know allocation is similarly pathological - in that case indeed this does not need a change.

Is it fair to say TinyGo's wasi target shouldn't be used in cases where performance is a concern given the allocator's limitations?

dgryski:
If performance is a concern, maybe you shouldn't be using WASM...

To be fair, we have the gc disabled for some of our services: `-gc=leaking` is fine for short-lived processes if their initial heap is set correctly.

anuraaga:
> If performance is a concern, maybe you shouldn't be using WASM...

WASM is the only way to extend Envoy for now so we're focusing on it for better or worse 😅  Unfortunately a leaking mode also isn't supported there, but maybe it should be.

For a bit of context, we're trying to bring Coraza WAF to Envoy via WASM - for performance reasons, we've swapped out several libraries from Go to C(++) or Rust

https://github.com/anuraaga/coraza-wasm-filter/tree/main/lib

The biggest was `regexp` itself with ~5x overall performance improvement from swapping in `libre2`. But other's like aho-corasick also provide 20% - the numbers are e2e not microbenchmark.

I think this means that while native will of course perform the best, wasm does have potential for reasonable performance. While I am curious what may be causing such drastic performance difference, if TinyGo itself is designed in a way that we expect lower performance than other compilers, it wouldn't be productive to do a deep dive. So I'm just trying to understand the vision there.

If small code size is the end all (as tiny would represent which makes a lot of sense!) then I would suggest that browser rather than wasi would be the target that fits better with the vision. But that's just a straw-man's proposal. Sorry if too much noise

aykevl:
Performance of TinyGo compiled code varies a lot. I've seen cases where TinyGo outperformed standard Go by a wide margin (for integer heavy code that doesn't allocate memory). For other cases, TinyGo will be a lot slower. In general, you can expect C like code to perform well but code that uses the heap, goroutines, etc to perform much worse.

In particular, for WebAssembly, TinyGo has to jump through hoops to get it supported. WebAssembly is just a really weird instruction set that is incredibly difficult for a Go-like language to target whereas other architectures are much simpler to support. This is getting improved slowly (with [exception handling](https://github.com/WebAssembly/exception-handling), [stack switching](https://github.com/WebAssembly/stack-switching), and a [GC](https://github.com/WebAssembly/gc/)) but this will likely take years. Once these features are supported in WebAssembly and used by TinyGo, I expect TinyGo binaries will perform not very different from C and Rust (perhaps a bit slower but not by much). Until then, we have to keep our workarounds in place which inevitably means code will run slower.

That said, the GC could certainly be improved. It works, but it wasn't optimized for speed (for example, the heap is fully conservative). I believe @dgryski is investigating how to improve this.

dgryski:
The two performance improvements I see for the current garbage collector (without a large scale rewrite) are:
1) increase the size of the mark work queue; an overflow requires an additional scan of the entire heap
2) make note of which allocations do not contain any pointers and this don't require scanning

The first one is easy (although a bit tricky to get right so as not to affect performance on low-end embedded machines) and the second one requires a bit more work to to track the newly required information.

anuraaga:
Thanks for the context everyone - definitely would be nice if the GC spec goes somewhere especially for polyglot! Those GC optimizations, especially the second one seems quite compelling too.

dgryski:
TinyGo is likely to benefit from the stack switching proposal (for goroutines) but not the garbage collection one (because Go supports interior pointers which the proposal doesn't allow.)

aykevl:
> but not the garbage collection one (because Go supports interior pointers which the proposal doesn't allow.)

It can probably still use the WebAssembly GC if all struct methods (etc) are made heap objects. For example, a struct like this:

```go
type Point struct {
    X, Y int
}
```

could be compiled like this:

```go
type Point struct {
    X, Y *int
}
```
This way, it is possible to have interior pointers at the cost of a large heap increase.
This will be even worse for things like byte slices (where it is possible to construct a pointer to each individual byte) but there may be workarounds for that too, like fat pointers. Still, it'll be rather difficult to get this working.

aykevl:
@anuraaga one way you could help here is by making a good (realistic) GC benchmark. Having a good benchmark would help a lot to improve the GC.

aykevl:
Updated the PR. The PR was failing because the tests actually contained a bug: they passed pointers around as `uintptr`. The `uintptr` type is an integer, not a pointer, so the GC wasn't tracking it and the memory was freed before all references were gone.
The first commit fixes the bug, the second commit is the original commit of this PR.

dgryski:
One of the gc-heavy benchmarks I've been playing with is the binarytrees program from the Benchmarks Game: https://benchmarksgame-team.pages.debian.net/benchmarksgame/program/binarytrees-go-2.html

anuraaga:
> The uintptr type is an integer, not a pointer, so the GC wasn't tracking it and the memory was freed before all references were gone.

This was intentional - the tests are calling malloc and the pointer should be valid until free is called even treated as an integer. If this code was c++ it would not be tracked either, hence why the allocator needs to track. 

I'm not sure why but I believe I had issues using `unsafe.Pointer` in `trackAlloc`, it had to be `[]byte` to work.

aykevl:
> This was intentional - the tests are calling malloc and the pointer should be valid until free is called even treated as an integer.

...you are entirely correct. Yes, the pointer should remain valid even if it is treated as `uintptr`.

> I'm not sure why but I believe I had issues using `unsafe.Pointer` in `trackAlloc`, it had to be `[]byte` to work.

That's interesting, sounds like a bug actually. I'll need to investigate this.

dgryski:
What's the status of this PR?  Do we still need it?

aykevl:
Closing, I don't think we really need this. The reduction in binary size would have been nice though.
(So if anybody wants to optimize binary size, here's an easy way to do that!)

PR #3163, opened 2022-09-15
wasm: add legacy target

aykevl:
This should keep at least some compatibility with older browsers. It isn't complete though: wasi-libc is still built with the bulk-memory feature enabled. But it may be good enough for some uses.

If needed, I guess we could ship a separate wasi-libc build with the bulk-memory feature disabled.

### DIFF
--- a/targets/wasm-legacy.json
+++ b/targets/wasm-legacy.json
@@ -1,0 +1,18 @@
+{
+	"llvm-target":   "wasm32-unknown-wasi",
+	"cpu":           "generic",
+	"build-tags":    ["tinygo.wasm"],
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
maybe in this description mention the browsers we are talking about (ex which threshold version has bulk-memory, and before that requires this build?)

deadprogram:
@aykevl is this PR still valid?

aykevl:
We might reconsider this in the future, but the current set of features we use is very widespread among browser by now.

As for newer features like stack switching or threading (once they finally land), I think they're better specified with flags like `-scheduler=`.

PR #3165, opened 2022-09-15
wasm: implement the //go:wasmimport directive

aykevl:
It is ~implemented upstream~ almost implemented upstream and looks pretty stable.

@neelance I'd like to get your opinion on this.
This would mean that TinyGo implements the `//go:wasmimport` directive for all packages (not just runtime and syscall/js). This seems mostly safe to me, and certainly better than the current `//go:wasm-module` directive that we're using. What do you think, am I missing something?
The ABI will be mostly the same as the one used in the Go toolchain, with minor differences here and there. It roughly matches the C ABI.

### DIFF
--- a/compiler/symbol.go
+++ b/compiler/symbol.go
@@ -288,6 +288,15 @@ func (info *functionInfo) parsePragmas(f *ssa.Function) {
 					continue
 				}
 				info.module = parts[1]
+			case "//go:wasmimport":
+				// Import a WebAssembly function, for example a WASI function.
+				// For details, see: https://github.com/golang/go/issues/38248
+				if len(parts) != 3 {
+					continue
+				}
+				info.exported = true
+				info.module = parts[1]
+				info.importName = parts[2]
 			case "//go:inline":
 				info.inline = inlineHint
 			case "//go:noinline":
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
fyi, this isn't quite in upstream, yet, though I personally support using this anyway https://go-review.googlesource.com/c/go/+/350737

aykevl:
> fyi, this isn't quite in upstream, yet

My bad, thanks for correcting.

gedw99:
basic question...

//go:wasmimport directive is for Browser side or server side ?  Or both ?

anuraaga:
https://github.com/golang/go/commit/02411bcd7c8eda9c694a5755aff0a516d4983952 has landed so hopefully this can move forward :)

deadprogram:
Reminder to @aykevl about this one :small_airplane:

aykevl:
Updated and removed the commits from #3142 from this PR.

deadprogram:
We also need to update `examples/wasm/export` and the associated README with this change. I can submit a separate PR.

Thanks everyone for working on this, exciting to see this moving forward! Now merging.

PR #3166, opened 2022-09-15
nrf: make GetRNG available to all chips

aykevl:
All nrf chips have a cryptographically secure RNG on board. Therefore, I've made the code more portable so that it works on all nrf chips.

I did remove a number of exported functions. I am of the opinion that these should only be made available once we have an agreed upon API for multiple chips. People who want to have greater control over the RNG should use the device/nrf package directly instead.

I have also changed the behavior to always enable digital error correction. Enabling it seems like a more conservative (and secure) default to me. Again, people who would like to have it disabled can use the device/nrf package directly.


deadprogram (approved):

deadprogram:
@aykevl please resolve merge conflict.

aykevl:
Rebased.

deadprogram (approved):

deadprogram:
Thanks @aykevl merging now.

PR #3156, opened 2022-09-12
main: add serial port monitoring functionality

sago35:
This PR adds the following features

* `tinygo flash -monitor`
    * After flashing, connect to the serial port
    * Serial port is automatically selected
* `tinygo monitor`
    * Subcommand to connect to serial port
    * `-target` or `-port` can be specified

It can be used as follows.

```
$ tinygo flash --target xiao-ble --size short -monitor examples/echo
   code    data     bss |   flash     ram
   8268     112    7840 |    8380    7952
Connected to COM134. Press Ctrl-C to exit.
hello there
You typed: hello there
こんにちは
You typed: こんにちは
```

```
$ tinygo flash --target xiao-ble --size short examples/serial
   code    data     bss |   flash     ram
   8000     112    7840 |    8112    7952

$ tinygo monitor -target xiao-ble
Connected to COM134. Press Ctrl-C to exit.
hello world!
hello world!
...

$ tinygo monitor -port COM134
Connected to COM134. Press Ctrl-C to exit.
hello world!
hello world!
...
```

I just made this as a trial, but it is very comfortable and I want you to try it.

----

Unfortunately, the Wio Terminal gives an error.
This is because the Bootloader is connected via USB-CDC and then reset.
Since #2992, touchSerialPortAt1200bps to bootloader is no longer needed, wioterminal.json should be changed.
I will create another PR that will make the json changes.

```
$ tinygo flash --target wioterminal --size short -monitor examples/echo
   code    data     bss |   flash     ram
   7688     108    6280 |    7796    6388
Connected to COM4. Press Ctrl-C to exit.
error: read error
```

PR #3168, opened 2022-09-16
wioterminal: remove serial-port setting of bootloader

sago35:
#3156 adds `tinygo flash -monitor`.

With wioterminal, it is possible to connect to a serialport after flashing but before it is reset.
This occurs when the following conditions are present

* bootloader has VID/PID for USBCDC
* VID/PID of the bootloader is listed in `targets/*.json`.
* Reset after flash takes time

before: COM4==bootloader, COM131==application
```
$ tinygo flash --target wioterminal --monitor examples/serial
Connected to COM4. Press Ctrl-C to exit.
error: read error: Port has been closed: The I/O operation has been aborted because of either a thread exit or an application request.
```

after:  COM4==bootloader, COM131==application
```
$ tinygo flash --target wioterminal --monitor examples/serial
Connected to COM131. Press Ctrl-C to exit.
hello world!
hello world!
```

----

Although it probably occurs at other boards than the Wio Terminal, it is necessary to determine which one is the bootloader.
If `tinygo flash -monitor` fails, most likely you are experiencing the same problem as this PR.
If you are using a board that causes problems, you can report and create a PR.

### DIFF
--- a/targets/wioterminal.json
+++ b/targets/wioterminal.json
@@ -2,7 +2,7 @@
     "inherits": ["atsamd51p19a"],
     "build-tags": ["wioterminal"],
     "serial": "usb",
-    "serial-port": ["acm:2886:002d", "acm:2886:802d"],
+    "serial-port": ["acm:2886:802d"],
     "flash-1200-bps-reset": "true",
     "flash-method": "msd",
     "msd-volume-name": "Arduino",


sago35:
On the Adafruit board, 0x8XXX appears to be the application and 0x0XXX appears to be the bootloader.
I would like to create another PR to correct this problem.

https://github.com/adafruit/ArduinoCore-samd/blob/master/boards.txt
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/boards.txt

deadprogram (approved):

deadprogram:
> On the Adafruit board, 0x8XXX appears to be the application and 0x0XXX appears to be the bootloader.
I would like to create another PR to correct this problem.

I can confirm that we need this now.

PR #3169, opened 2022-09-16
machine: add KHz, MHz, GHz constants, deprecate TWI_FREQ_* constants

aykevl:
There are two main issues with these constants:

  * They don't follow the Go naming convention.
  * They call themselves "TWI", while it makes a lot more sense to refer to the actual name which is I2C (or I²C).

I have not removed them but just deprecated them. Perhaps we can remove them when we move towards version 1.0.

---

This may be a somewhat radical change. A less radical change would be to also define new I2C specific constants, like:

```go
const (
    I2CFrequency100KHz = 100 * KHz
    I2CFrequency400KHz = 400 * KHz
)
```

A bit longer but IMHO a lot better than `TWI_FREQ_100KHZ` etc.

Part of #3170.

### DIFF
--- a/src/machine/machine_nrf.go
+++ b/src/machine/machine_nrf.go
@@ -230,7 +230,7 @@ func (i2c *I2C) Configure(config I2CConfig) error {
 
 	// Default I2C bus speed is 100 kHz.
 	if config.Frequency == 0 {
-		config.Frequency = TWI_FREQ_100KHZ
+		config.Frequency = 100 * KHz
 	}
 	// Default I2C pins if not set.
 	if config.SDA == 0 && config.SCL == 0 {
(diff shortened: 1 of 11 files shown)


src/machine/machine_nrf.go
```diff
@@ -253,7 +253,7 @@ func (i2c *I2C) Configure(config I2CConfig) error {
 		(nrf.GPIO_PIN_CNF_DRIVE_S0D1 << nrf.GPIO_PIN_CNF_DRIVE_Pos) |
 		(nrf.GPIO_PIN_CNF_SENSE_Disabled << nrf.GPIO_PIN_CNF_SENSE_Pos))
 
-	if config.Frequency == TWI_FREQ_400KHZ {
+	if config.Frequency >= 400*KHz {
```
aykevl, line 256: This is a small behavioral change but it seems more correct this way.

kenbell:
LGTM

deadprogram:
I really like the way this reads. Good change!

PR #3170, opened 2022-09-16
machine: add tool to check consistency with documentation

aykevl:
This tool checks whether the machine package matches the documentation in https://tinygo.org/docs/reference/machine/.

---

This is very much a draft. Right now it only checks for constants (not even functions, types, methods, etc) and there are many things that need to be changed for this to pass all tests.

* Many constants need to be documented on tinygo.org.
* Many other constants need to be made private, because they are only to be used within the machine package.
* Some constants that are considered valid here (like `ParityNone`) aren't actually documented yet, this must be done before this PR is ready.

The first step in fixing this is #3169.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -236,6 +236,8 @@ jobs:
         run: |
           make ASSERT=1
           echo "$(pwd)/build" >> $GITHUB_PATH
+      - name: Test machine package
+        run: make check-machine
       - name: Test stdlib packages
         run: make tinygo-test
       - name: Install Xtensa toolchain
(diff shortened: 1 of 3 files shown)


dgryski:
Hooray for static analysis!  Any other things we might want to have tooling to verify?

aykevl:
This PR only checks for constants. I eventually want it to check the entire public API of the machine package (global variables, function signatures like `GetRNG`, and methods on peripheral objects lik I2C, PWM, etc).

Other than that, there are some more things that could be worth checking:

- simple checks for heap allocations inside machine package functions (like `return errors.New("...")`)
- imports: only let the machine package import a limited number of packages
- ...others?

dgryski:
```
~/go/src/github.com/tinygo-org/tinygo/src/machine $ awk '/^)/{imports=0}; {if (imports==1) {pkgs[$0]++}}; /^import \(/{imports=1}; END {for (p in pkgs) {print p} }  ' *.go

	"sync"
	"device/esp"
	"math/bits"
	_ "unsafe" // for go:linkname
	"device/kendryte"
	"internal/itoa"
	"math"
	"unsafe"
	"image/color"
	"device/sifive"
	"errors"
	"runtime/volatile"
	"runtime/interrupt"
	"device/avr"
	"machine/usb"
	"device/arm"
	"device/nrf"
	"device/stm32"
	"device/rp"
	"device/riscv"
	"device/nxp"
	"device/sam"
```

It's a start anyway :D

PR #3172, opened 2022-09-17
xiao-ble: fix usbpid

sago35:
0x0045 is the PID for bootloader.

* The problem comes when running tinygo flash -monitor
* Flash from Arduino IDE to 0x8045


deadprogram:
Thanks for the fix @sago35 now merging.

PR #3173, opened 2022-09-17
targets: remove usbpid of bootloader

sago35:
Removing the USBPID in the bootloader allows `tinygo flash -monitor` to work well.
Since #2992, touchSerialPortAt1200bps to bootloader is no longer needed, 

The relevant URLs are
* #3168
* #3172
* #2992 
* #3156


deadprogram:
Thanks for the further updates @sago35 merging.

PR #3175, opened 2022-09-18
machine: improve UARTParity slightly

aykevl:
There are only 3 possible values, so a uint8 seems more appropriate. Also, there is no reason to use specific numbers over a simple enumeration.

---

This is rather subjective but I think it looks better this way. Feel free to disagree :)


deadprogram:
Seems like a good idea to me! Merging, thank you @aykevl

PR #3177, opened 2022-09-18
machine: move SPI Tx function into separate file

aykevl:
This makes the code a bit cleaner because ErrTxInvalidSliceSize isn't redefined in every file that uses SPI and Mode0/Mode1/Mode2/Mode3 is defined for every target that uses SPI.

Found while working on #3170.

### DIFF
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -10,7 +10,6 @@ package machine
 import (
 	"device/arm"
 	"device/sam"
-	"errors"
 	"runtime/interrupt"
 	"unsafe"
 )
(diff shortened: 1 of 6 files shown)


deadprogram:
Seems like a good idea. 

Question: is there a particular reason you chose to name the file `spi-tx.go` instead of `spi_tx.go`?

aykevl:
> Question: is there a particular reason you chose to name the file `spi-tx.go` instead of `spi_tx.go`?

Not really. I've come to dislike the current convention because it can conflict with future GOOS/GOARCH values. But perhaps it's better to stick to the convention so I've updated the PR.

deadprogram (approved):

deadprogram:
Thanks @aykevl now merging.

PR #3178, opened 2022-09-18
machine: rename PinInputPullUp/PinInputPullDown

aykevl:
Some targets used capital PullUp/PullDown, while the documented standard is Pullup/Pulldown. This commit fixes this mismatch.

Perhaps PullUp/PullDown were better names, I don't know. But right now we use Pullup/Pulldown so I'd say we should use that for all chips.

Found while working on #3170. Also, another reason why we need automatic checking (like in #3170) for the machine package.

### DIFF
--- a/src/machine/machine_k210.go
+++ b/src/machine/machine_k210.go
@@ -23,9 +23,15 @@ type PinChange uint8
 // Pin modes.
 const (
 	PinInput PinMode = iota
-	PinInputPullUp
-	PinInputPullDown
+	PinInputPullup
+	PinInputPulldown
 	PinOutput
+)
+
+// Deprecated: use PinInputPullup and PinInputPulldown instead.
+const (
+	PinInputPullUp   = PinInputPullup
+	PinInputPullDown = PinInputPulldown
 )
 
 // FPIOA internal pull resistors.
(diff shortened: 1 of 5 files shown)


deadprogram:
This PR certainly does help expose inconsistencies having spread to several places as a result.

For example, the reason that TInyHCI is failing on this PR:

```
sep 18 21:20:11 tinyhci-UM350 tinygohci[10636]: main.go:118:50: PinInputPullUp not declared by package machine
sep 18 21:20:11 tinyhci-UM350 tinygohci[10636]: main.go:132:52: PinInputPullDown not declared by package machine
sep 18 21:20:11 tinyhci-UM350 tinygohci[10636]: main.go:166:56: PinInputPullDown not declared by package machine
sep 18 21:20:11 tinyhci-UM350 tinygohci[10636]: main.go:182:56: PinInputPullUp not declared by package machine
sep 18 21:20:11 tinyhci-UM350 tinygohci[10636]: main.go:196:58: PinInputPullDown not declared by package machine
sep 18 21:20:11 tinyhci-UM350 tinygohci[10636]: 2022/09/18 21:20:11 Github check run failed on board maixbit for 7fa6c9475b3d21b8d8eb4234bf9b4db166bd61c6
```

aykevl:
@deadprogram if needed I can add (deprecated) fallback constants on these chips. What do you think?

deadprogram:
> if needed I can add (deprecated) fallback constants on these chips. What do you think?

Probably a good idea. There is a lot of code out there already using the current set of values.

aykevl:
Ok, done so. The updated PR should be fully backwards compatible.

Still, I am of the opinion we should eventually remove these constants. Maybe in a few months or a year or so. That may break some code, but it has to be done eventually (certainly before v1.0).

deadprogram:
Thanks for adding the aliases. Perhaps we can look at removing them entirely after the next release?

deadprogram (approved):

PR #3180, opened 2022-09-18
Add ReadTemperature

aykevl:
This PR removes `ADC_TEMP_SENSOR` and adds a plain old function to read the current chip temperature: `ReadTemperature`.

The main motivation for this PR was to remove `ADC_TEMP_SENSOR` which, apart from not following the Go naming convention, is an API specific to the rp2040. With this PR, I've changed the API to a simple function and also added one for the nrf series of chips to show how this API is more general.

We might want to add an error return value, although I'm not entirely sure whether that's useful. On the two implemented chip families (nrf and rp2040), no error is produced by the temperature sensor.

Part of the #3170 effort (`ADC_TEMP_SENSOR` is a constant and therefore shows up in the machine check).


deadprogram (approved):
Much better with the additional abstraction.

PR #3183, opened 2022-09-19
main: restore support for flashing Espressif chips

aykevl:
This reverts commit 303410d3fcfcf3df45bfe6c927a5b963caf3a394 and also removes the unnecessary fallback. See https://github.com/tinygo-org/tinygo/pull/3092 for discussion.

In summary, #3092 broke support for flashing my esp32-c3 board and I think most other Espressif based boards as well.


deadprogram:
Looks like we really need to add an ESP32 to the TinyHCI!

deadprogram (approved):

sago35:
LGTM

PR #3185, opened 2022-09-19
machine: do not expose RESET_MAGIC_VALUE

aykevl:
This is a constant for internal use only, but was (unintentionally?) exported. In addition, it doesn't follow the Go naming convention. This change simply renames the constant so that it is unexported.

Part of #3170.


deadprogram (approved):
:cloud: :fog: :magic_wand:

PR #3188, opened 2022-09-21
compileopts: silently succeed when there's no debug info to strip

codefromthecrypt:
Before, on MacOS, we erred if the user provided configuration to strip debug info.

Ex.
```bash
$ tinygo build -o main.go  -scheduler=none --no-debug  main.go
error: cannot remove debug information: MacOS doesn't store debug info in the executable by default
```

This is a poor experience which results in having OS-specific CLI behavior. Silently succeeding is good keeping with the Linux philosophy and less distracting than logging the same without failing.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -700,21 +700,25 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 	// Add embedded files.
 	linkerDependencies = append(linkerDependencies, embedFileObjects...)
 
+	// Determine whether the compilation configuration would result in debug
+	// (DWARF) information in the object files.
+	var hasDebug = true
+	for _, tag := range config.BuildTags() {
+		if tag == "baremetal" {
+			// Don't use -no-debug on baremetal targets. It makes no sense:
+			// the debug information isn't flashed to the device anyway.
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
Added another commit to keep failing on baremetal based on a conversation with @aykevl on slack. I think there's a sense that failing on the same args if baremetal is a feature not a bug. Fair enough.

Feel free to do whatever you like for example squash, drop the last commit if you change mind, or close out with another PR.

deadprogram:
Perhaps it would be easier to reason about this flag if it were `include-debug` or perhaps `strip-debug` implying some action to be taken.

The current `no-debug` flag implies an end-state e.g. produce a binary without debug information. So I would agree that if the developer includes this flag, and that target does not include debug information anyhow, it would be best to "just work". So IMO this PR is best without the second commit.

codefromthecrypt:
cool. either way works.

Regardless, I'll close out any of my PRs still open at the end of the week as they are both insignificant and not important enough to track for merge/rebase long term (as a volunteer). Until then, I'll keep an eye open!

codefromthecrypt:
obviated by #3200

PR #3190, opened 2022-09-21
main: allow setting the baud rate for serial monitors

sago35:
The m5stamp-c3 is connected via UART-USB conversion.
And the baud rate was 115200 bps.

The following changes were made in this PR

* Default setting changed to 115200 bps (from 9600 bps) to match the standard baud rate of TinyGo's UART
* Can now be changed from the command line with `--baudrate 115200`

With this change, it can now be used as a simple serial monitor tool.

```
# example
$ tinygo monitor --port COM1 --baudrate 9600 
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -1337,6 +1337,7 @@ func main() {
 	llvmFeatures := flag.String("llvm-features", "", "comma separated LLVM features to enable")
 	cpuprofile := flag.String("cpuprofile", "", "cpuprofile output")
 	monitor := flag.Bool("monitor", false, "enable serial monitor")
+	baudrate := flag.Int("baudrate", 0, "baudrate of serial monitor")
 
 	var flagJSON, flagDeps, flagTest bool
 	if command == "help" || command == "list" || command == "info" || command == "build" {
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -1337,6 +1337,7 @@ func main() {
 	llvmFeatures := flag.String("llvm-features", "", "comma separated LLVM features to enable")
 	cpuprofile := flag.String("cpuprofile", "", "cpuprofile output")
 	monitor := flag.Bool("monitor", false, "enable serial monitor")
+	baudrate := flag.Int("baudrate", 0, "baudrate of serial monitor")
```
deadprogram: Wouldn't it be better to just use default value of `115200` here instead of this code?

https://github.com/tinygo-org/tinygo/pull/3190/files#diff-23669db2c16e34742ce163e9f6bc55e0e78d04ee8c1eb292f57b6967805fe12dR36-R38

sago35: Indeed, 115200 bps seems to be the default here.

> https://github.com/tinygo-org/tinygo/pull/3190/files#diff-23669db2c16e34742ce163e9f6bc55e0e78d04ee8c1eb292f57b6967805fe12dR36-R38

I think it is safer to leave error handling in place.

deadprogram:
Thanks @sago35 😺

PR #3191, opened 2022-09-21
runtime: make gc and scheduler asserts/logging settable with build tags

dgryski:


src/runtime/build_asserts.go
```diff
@@ -0,0 +1,9 @@
+//go:build runtime_asserts
+
+package runtime
+
+// enable assertions for the garbage collector
+const gcAsserts = true
+
+// enable asserts for the schedule:w
```
aykevl: Looks like you're using vim? ;)

aykevl (commented):
LGTM with a minor nit

src/runtime/build_asserts.go
dgryski: :D. Fixed.

aykevl (approved):

deadprogram:
@dgryski seems like something might be wrong with submodule? The fail on CircleCI is a bit weird.

dgryski:
Yeah, something's up with Circle CI.  I think the setup script doesn't like the fact that the commit is on a fork and not a branch or something.  I think there's a larger problem going on here, but not one that was introduced with this PR..

deadprogram:
OK, then let's merge.

PR #3195, opened 2022-09-24
ci: build TinyGo using Go 1.19

aykevl:
We've supported Go 1.19 for a while now, let's actually use it in CI.


deadprogram:
:+1:

PR #3196, opened 2022-09-24
build: set circleci resource class to large for CI build

deadprogram:
This PR sets the CircleCI resource class to large for CI build` to speed them up. (approx 43%)


sago35 (approved):
LGTM
It is much faster.

before: 13 - 14 min
after: 10 min

deadprogram:
Thanks for the review @sago35

PR #3200, opened 2022-09-26
compileopts: silently succeed when there's no debug info to strip

deadprogram:
This PR is to use only the first commit from #3188 courtesy of @codefromthecrypt 

Before, on the baremetal target or MacOS, we errored if the user provided configuration to strip debug info.

Ex.
```bash
$ $ tinygo build -o main.go  -scheduler=none --no-debug  main.go
error: cannot remove debug information: MacOS doesn't store debug info in the executable by default
```

This is a poor experience which results in having OS-specific CLI behavior. Silently succeeding is good keeping with the Linux philosophy and less distracting than logging the same without failing.

Signed-off-by: Adrian Cole <adrian@tetrate.io>

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -700,21 +700,25 @@ func Build(pkgName, outpath string, config *compileopts.Config, action func(Buil
 	// Add embedded files.
 	linkerDependencies = append(linkerDependencies, embedFileObjects...)
 
+	// Determine whether the compilation configuration would result in debug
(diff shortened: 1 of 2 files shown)


codefromthecrypt (approved):
thanks for taking this over

deadprogram:
Now merging, thanks @codefromthecrypt

PR #3201, opened 2022-09-26
machine: hide internal pin modes

aykevl:
Many pin modes are only for peripherals like SPI and PWM. Do not make these part of the public API: the public API is the `Configure` method on the peripheral, not the `PinMode` constant.

This may be a somewhat controversial change because it hides things that were previously exported, but I think the API will be in a better shape with these constants hidden.

Part of #3170.

### DIFF
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -17,20 +17,20 @@ import (
 const deviceName = sam.Device
 
 const (
-	PinAnalog    PinMode = 1
-	PinSERCOM    PinMode = 2
-	PinSERCOMAlt PinMode = 3
-	PinTimer     PinMode = 4
-	PinTimerAlt  PinMode = 5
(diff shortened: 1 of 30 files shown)


deadprogram:
It would be wise to run the `drivers` repo tests using this branch before we merge, just in case we are doing something there that will fail with this change.

deadprogram:
Here is an example of something that does not compile with this change: https://github.com/tinygo-org/drivers/blob/release/flash/transport_qspi_samd.go

aykevl:
> Here is an example of something that does not compile with this change: https://github.com/tinygo-org/drivers/blob/release/flash/transport_qspi_samd.go

Yeah, that shouldn't have been in the drivers repo in the first place. It should be fully abstracted in the machine package instead.

...guess I have some refactoring to do before this PR can land.

deadprogram:
Closing since it is out of date. Thank you!

PR #3159, opened 2022-09-14
Add flag for setting the goroutine stack size 

aykevl:
This is helpful in some cases where the default stack size isn't big enough. See https://github.com/tinygo-org/tinygo/pull/3135 for example.

PR #3206, opened 2022-09-28
targets: do not set stack size per board

aykevl:
The needed stack size for a goroutine is hard to determine by the compiler. It will try, but will fail in many common cases. Therefore, the compiler will pick a fixed stack size at compile time.

There is a tradeoff between avoiding stack overflows and wasting RAM. This tradeoff depends on the application: some don't need large stack sizes but do need a lot of memory, while others need deep stacks but aren't so memory constrained. That's why I've added a flag to do this on the command line: https://github.com/tinygo-org/tinygo/pull/3159

It may be reasonable to use a different stack size per chip, for example chips with lots of RAM could default to a larger stack size. But I don't think it's a good idea to do this per board.

@eyz @sago35 @ardnew @bgould what do you think?

### DIFF
--- a/targets/atsame54-xpro.json
+++ b/targets/atsame54-xpro.json
@@ -3,6 +3,5 @@
     "build-tags": ["atsame54_xpro"],
     "serial": "usb",
     "flash-method": "openocd",
-    "openocd-interface": "cmsis-dap",
-    "default-stack-size": 4096
+    "openocd-interface": "cmsis-dap"
 }
--- a/targets/feather-m4-can.json
+++ b/targets/feather-m4-can.json
@@ -6,6 +6,5 @@
     "flash-1200-bps-reset": "true",
     "flash-method": "msd",
     "msd-volume-name": "FTHRCANBOOT",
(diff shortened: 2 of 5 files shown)


eyz:
The bare minimum default stack size to pass some typical examples / tests is fine by me.

ardnew:
As long as we keep the _ability_ to set it per target (but just leave them undefined in the default board JSONs), I don't see an issue. This is just a convenience rather than having to convolute the command-line.

Another option would be to read it from an env var. But if we did this, we should do it for all command-line flags, not just a handful.

Note that this "goroutine stack size" is very likely to be misunderstood by users, as it was for me originally. Two ideas to clarify:

 1. Add a very detailed description in the `tinygo` help/usage output, or a reference to some detailed documentation
 2. Add a compiler warning with docs/reference if the value doesn't meet some sanity criteria (e.g. size >  some percentage of physical RAM)

deadprogram (approved):
I :heart: :heavy_minus_sign:

PR #3207, opened 2022-09-28
targets: remove "acm:"` prefix for USB vid/pid pair

aykevl:
This prefix isn't actually used and only adds noise, so remove it.

It may have been useful on Linux that makes a distinction between /dev/ttyACM* and /dev/ttyUSB* but it isn't now. Also, it's unlikely that the same vid/pid pair will be shared between an acm and usb driver anyway.

---

I made this distinction originally because it seemed like a good idea at the time. But it really isn't.


sago35 (approved):
LGTM!

deadprogram:
Thanks @aykevl and @sago35 now merging.

PR #3208, opened 2022-09-28
testdata: increase timings used for timers test to try to avoid race condition errors on macOS CI

deadprogram:
This PR increases the timings used for the `timers.go` test to try to avoid spurious race condition errors on macOS CI

### DIFF
--- a/testdata/timers.go
+++ b/testdata/timers.go
@@ -4,51 +4,51 @@ import "time"
 
 func main() {
 	// Test ticker.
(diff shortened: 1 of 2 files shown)


deadprogram:
Merging because it seems less painful than rerunning the current `dev` yet again for macOS. :smiley_cat:

aykevl:
I am actually suspecting a more fundamental issue here that messes up the timings.

deadprogram:
That is almost for sure certain, this is just to prevent having to restart the CI so frequently.

PR #3210, opened 2022-09-29
cgo: add support for C.float and C.double

aykevl:
They are not necessary in TinyGo because they always map to float32 and float64, but it's a good idea to add them regardless for compatibility with existing software.

It's a bit of a hack but it works and it doesn't seem too dirty to me.

(Now I think about it, perhaps it would have been better to require explicit casts here just in case we want to support some really weird C system, but then again we even force 64-bit double on AVR even though avr-gcc defaults to 32-bit double).


dgryski (approved):
LGTM, thanks.

cgo/cgo.go
```diff
@@ -309,6 +311,12 @@ func Process(files []*ast.File, dir, importPath string, fset *token.FileSet, cfl
 	// Process CGo imports for each file.
 	for i, f := range files {
 		cf := p.newCGoFile(f, i)
+		// Float and double are aliased, meaning that C.float is the same thing
+		// as float32 in Go.
+		cf.names["float"] = clangCursor{}
```
dgryski, line 316: So why is this hack needed instead of having them added to wherever the integer types and their translations are defined?

aykevl: Because of this check here:

https://github.com/tinygo-org/tinygo/blob/b5ad81c884981f42c1ac491cc1d195aea886ddbe/cgo/cgo.go#L1269

I considered something like this:

```go
        if found, ok := names[fun.Sel.Name]; x.Name == "C" && (ok || fun.Sel.Name == "float" || fun.Sel.Name == "double") {
```

...but I'm not sure whether that's an improvement.

aykevl: - they are different from things like `C.int16_t` because `int16_t` and the like are actual typedefs defined in C
- they are different from things like `C.int` because `C.int` has to be a separate type from Go `int` (they can have a different bit width). Instead, `float32` is an alias of `C.float` and not a new type definition.

deadprogram:
Thanks for the enhancement @aykevl and to @dgryski for reviewing.

Merging!

PR #3211, opened 2022-09-30
builder: refactor Build function to not use a callback

aykevl:
The only reason a callback was used, was so that the temporary directory gets removed once `Build` returns. But that is honestly a really bad reason: the parent function can simply create a temporary function and remove it when it returns. It wasn't worth the code complexity that this callback created.

This change should not cause any observable differences in behavior (it should be a non-functional change). Lots of lines changed, but that's mostly just a change in indentation level.

I have no reason to do this now, but this unclean code has been bugging me and I just wanted to get it fixed.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -42,8 +42,8 @@ type BuildResult struct {
 	// information. Used for GDB for example.
 	Executable string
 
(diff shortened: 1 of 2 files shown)


deadprogram (approved):
The simplification is good, thanks @aykevl

PR #3213, opened 2022-09-30
builder: add test to check for changes in binary size

aykevl:
Something I've wanted to do for a while: a check to make sure binary size doesn't change unintentionally.
This covers only a very limited subset of boards, but I hope that by making it diverse it will still catch most changes.

Depends on #3211 so marked as draft.

### DIFF
--- a/builder/sizes_test.go
+++ b/builder/sizes_test.go
@@ -1,0 +1,92 @@
+package builder
+
+import (
+	"runtime"
+	"testing"
+	"time"
+
+	"github.com/tinygo-org/tinygo/compileopts"
+)
+
+var sema = make(chan struct{}, runtime.NumCPU())
+
+type sizeTest struct {
+	target     string
+	path       string
+	codeSize   uint64
+	rodataSize uint64
+	dataSize   uint64
+	bssSize    uint64
+}
+
+// Test whether code and data size is as expected for the given targets.
+// This tests both the logic of loadProgramSize and checks that code size
+// doesn't change unintentionally.
+//
+// If you find that code or data size is reduced, then great! You can reduce the
+// number in this test.
+// If you find that the code or data size is increased, take a look as to why
+// this is. It could be due to an update (LLVM version, Go version, etc) which
+// is fine, but it could also mean that a recent change introduced this size
+// increase. If so, please consider whether this new feature is indeed worth the
+// size increase for all users.
+func TestBinarySize(t *testing.T) {
+	if runtime.GOOS == "linux" && !hasBuiltinTools {
+		// Debian LLVM packages are modified a bit and tend to produce
+		// different machine code. Ideally we'd fix this (with some attributes
+		// or something?), but for now skip it.
+		t.Skip("Skip: using external LLVM version so binary size might differ")
+	}
+
+	// This is a small number of very diverse targets that we want to test.
+	tests := []sizeTest{
+		// microcontrollers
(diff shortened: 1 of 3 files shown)


builder/sizes_test.go
```diff
@@ -0,0 +1,85 @@
...
+			if err != nil {
+				t.Fatal("could not build:", err)
+			}
+
+			// Check whether the size of the binary matches the expected size.
+			sizes, err := loadProgramSize(result.Executable, nil)
+			if err != nil {
+				t.Fatal("could not read program size:", err)
+			}
+			if sizes.Code != tc.codeSize || sizes.ROData != tc.rodataSize || sizes.Data != tc.dataSize || sizes.BSS != tc.bssSize {
```
dgryski, line 84: Do we want some fudge factors here (+/- 0.01% or something?) to adjust for instruction selection variations?  This test just seems really fragile with all the possible variables (even the full path name for the debug info..)

aykevl: > Do we want some fudge factors here (+/- 0.01% or something?) to adjust for instruction selection variations?

For the same Go and LLVM versions, the selected instructions should be identical.
Also, it's hard to define "small". A small change in the input source can result in a very large change in the binary, just because some compiler heuristic went wrong. So it's hard to say what kind of fudge factor would be good.

>  This test just seems really fragile with all the possible variables (even the full path name for the debug info..)

I believe that binary size should be identical when using the same LLVM and Go versions (which are only updated every half year in CI), and the same C library versions (musl, wasi-libc, etc, which are updated even less frequently). Debug information is not part of the calculated binary size so changing file paths won't affect the test results.

That said, I do see a failure in CI and it might be an actual bug:

```
    sizes_test.go:78: Unexpected code size when compiling: -target=microbit examples/serial
    sizes_test.go:79:             code rodata   data    bss
    sizes_test.go:80: expected:   3012    164      8   2248
    sizes_test.go:81: actual:     3020    164      8   2248
```

I suspect it's because of some patches in Debian LLVM that affect the binary, I'll need to investigate what's going on here.

dgryski: > I believe that binary size should be identical when using the same LLVM and Go versions (which are only updated every half year in CI), and the same C library versions (musl, wasi-libc, etc, which are updated even less frequently). 

Ah right, so this test is really only useful in the context of CI.  That makes sense.  I wonder if we want to conditionalize it on some environment variable so that running all the tests on dev machines doesn't trigger spurious failures.

aykevl: I believe that as long as the test runs using a self-built LLVM version, the output should be stable at a given TinyGo version.
I've updated the PR to only run the tests with a self-built LLVM, not with a precompiled LLVM (apt.llvm.org, Homebrew, etc). I _hope_ that's good enough.

aykevl:
I've had to disable the wasm tests because their output depends on the `wasm-opt` version which we often don't control. Perhaps they can be added back in some way in a later PR. For now, I just want to get something like this merged so we are finally able to test for binary size changes.

aykevl:
This now passes all tests :tada:
@dgryski can you take another look?

dgryski (approved):
LGTM

aykevl:
Updated and rebased on the current dev branch.

dgryski (approved):
Still LGTM

dgryski (approved):
Still LGTM

deadprogram:

PR #3218, opened 2022-10-05
rp2040: remove mem allocation in GPIO ISR

### LINKED ISSUE #3217: rp2040: GPIO interrupt causes GC load
TinyGo version: `tinygo version 0.26.0 linux/amd64 (using go version go1.19 and LLVM version 14.0.0)`

Actual behaviour:
GPIO interrupts disabled for 100ms every 250ms

Expected behaviour:
GPIO interrupts not disabled

Test app:
```go
package main

import (
	"machine"
	"time"
)

func main() {
	println("starting")

	machine.GPIO12.Configure(machine.PinConfig{Mode: machine.PinInput})
	machine.GPIO12.SetInterrupt(machine.PinRising|machine.PinFalling, handleGPIO)

	for {
		time.Sleep(1 * time.Second)
	}
}

var count = 0

func handleGPIO(pin machine.Pin) {
	count++
	machine.LED.Set(count&0x8 != 0)
}
```

Scenario: Run test app above on feather RP2040.  Apply 2KHz signal to GPIO12 using a signal generator.  Observe the frequent periods where LED is not updated (LED has irregular pulsing).  The LED should be pulsed at a consistent rate of approx 2K/8 = 250Hz

Scope output:
SDS00003

Yellow is the LED line
Purple is when the GC runs (I made the GC set GPIO11 high during GC runs):
```go
func runGC() (freeBytes uintptr) {
	machine.GPIO11.Configure(machine.PinConfig{Mode:machine.PinOutput})
	machine.GPIO11.High()
	if gcDebug {
		println("running collection cycle...")
	}
//...
	machine.GPIO11.Low()
	return
}
```

Observations:
1. There appears to be a 50ms pause before the GC runs (possibly low-memory condition, looking for memory blocks?)
2. It appears something is causing frequent memory allocations, possibly in the interrupt handling path, causing GC to run - probably inside the ISR context since multiple GC cycles within time.Sleep.
3. It appears interrupts are disabled during allocation and GC, and which can take 100ms+

kenbell:
Fixes #3217

It's unclear why this fixes the issue (since code is logically equivalent), but avoiding the assignment of register to a variable appears to enable some additional optimization logic to apply that prevents an 'escape'.

Have confirmed the fix by output of `tinygo flash -target feather-rp2040 -print-allocs . -serial uart .`, which no longer shows an allocation inside `machine_rp2040_gpio.go` - also confirmed on scope that test app from #3217 no longer experiences GC pauses.

I think this is a good example of why it would be good to have the allocation path explicitly fail inside ISRs so that subtle issues like this become obvious.

### DIFF
--- a/src/machine/machine_rp2040_gpio.go
+++ b/src/machine/machine_rp2040_gpio.go
@@ -288,8 +288,8 @@ func gpioHandleInterrupt(intr interrupt.Interrupt) {
 			base = &ioBank0.proc1IRQctrl
 		}
 
-		statreg := base.intS[gpio>>3]
-		change := getIntChange(gpio, statreg.Get())
+		statreg := base.intS[gpio>>3].Get()
+		change := getIntChange(gpio, statreg)
 		if change != 0 {
 			gpio.acknowledgeInterrupt(change)
 			callback := pinCallbacks[core][gpio]


src/machine/machine_rp2040_gpio.go
```diff
@@ -288,8 +288,8 @@ func gpioHandleInterrupt(intr interrupt.Interrupt) {
 			base = &ioBank0.proc1IRQctrl
 		}
 
-		statreg := base.intS[gpio>>3]
-		change := getIntChange(gpio, statreg.Get())
+		statreg := base.intS[gpio>>3].Get()
+		change := getIntChange(gpio, statreg)
```
aykevl, line 292: Yes, this is a necessary fix.
Previously it may have worked, but really only by chance. Because `base.intS[gpio>>3]` returns a `volatile.Register32`, which is essentially just an integer. The `.Get()` method uses a pointer receiver, so Go will essentially make sure to pass `&statreg` to it. This circumvents the volatile read (making it a normal read) and of course forces a heap allocation.

The new code should be a lot better.

aykevl (commented):
> I think this is a good example of why it would be good to have the allocation path explicitly fail inside ISRs so that subtle issues like this become obvious.

Absolutely. Perhaps we should just make sure to merge #1460 once it passes CI tests now early in the development cycle, and deal with the likely fallout during this cycle? I really want to get this check in TinyGo because it's too easy to make these kinds of mistakes.

In any case, LGTM assuming pin interrupts still work on the rp2040.

kenbell:
I'm all for getting #1460 merged as soon as possible.  There's random 'it doesn't work' issues because of latent allocation-in-ISRs bugs...

I only found this because i had a relatively complex app for an RF chip that would work for a short period, then die in a fire for absolutely no reason - it just seemed like frustrating random crashes until i really dug in.

src/machine/machine_rp2040_gpio.go
kenbell: Makes sense

dgryski: Is there any way tooling could have detected this?

Edit: ah yes, I see https://github.com/tinygo-org/tinygo/pull/1460

kenbell:
I'll go ahead and merge this fix so we can sequence in #3219 and then #1460 (at least keep RP2040 working after #1460 is merged)

PR #3219, opened 2022-10-05
Rp2040 fix usb allocations inside isr

kenbell:
See #3217 and #1460

This is a quick-and-dirty fix for USB (particularly on RP2040) which runs inside ISRs and was doing some allocation.  Replace buffer allocation with pre-allocated buffers.

### DIFF
--- a/src/machine/usb.go
+++ b/src/machine/usb.go
@@ -72,6 +72,13 @@ var udd_ep_in_cache_buffer [7][64]uint8
 //go:align 4
 var udd_ep_out_cache_buffer [7][64]uint8
 
+// usb_trans_buffer max size is 255 since that is max size
+// for a descriptor (bLength is 1 byte), and the biggest use
+// for this buffer is to transmit string descriptors.  If
+// this buffer is used for new purposes in future the length
+// must be revisited.
+var usb_trans_buffer [255]uint8
+
 var (
 	usbTxHandler    [usb.NumberOfEndpoints]func()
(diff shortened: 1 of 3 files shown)


sago35:
LGTM
I have confirmed that it works with

* feather-nrf52840 (nrf52840)
* feather-m0 (samd21)
* wioterminal (samd51)

sago35:
This PR must be merged after the following are merged

* #3218

kenbell:
@deadprogram - would be good to get this one merged if it looks OK

src/machine/usb.go
```diff
@@ -72,6 +72,22 @@ var udd_ep_in_cache_buffer [7][64]uint8
 //go:align 4
 var udd_ep_out_cache_buffer [7][64]uint8
 
+var usb_trans_buffer []uint8 = makeUSBTransBuffer()
```
deadprogram: You could probably also pre-allocate this as `var usb_trans_buffer [whatever the max size]uint8` then `return usb_trans_buffer[:size]` in the following function.

deadprogram: Looking further into what you did, I think you already probably have enough for now.

kenbell: I've pushed a version that pre-allocates at max size.  The trade-off is probably something like costing 200 bytes, but is a bit simpler.

deadprogram (approved):
I really like removing heap allocations! :scissors:

deadprogram:
Thank you @kenbell now merging.

PR #3222, opened 2022-10-07
main: print ldflags including ThinLTO flags with -x

aykevl:
The -x flag prints commands as they are executed. Unfortunately, for the link command, they were printed too early: before the ThinLTO flags were added. This is somewhat confusing while debugging.


dgryski (approved):
LGTM

deadprogram:
Thanks for the improvement @aykevl and review @dgryski now merging.

PR #3224, opened 2022-10-09
darwin: fix syscall.Open on darwin/arm64

### LINKED ISSUE #2997: "tinygo test os" fails due to varargs problems on M1 mac
Apple documents varargs as being different on darwin arm64 (see https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms) but it appears tinygo dev branch doesn't quite implement that yet.

This shows up as test failures in ```tinygo test os```.  When TestFstat tries to create a file, it ends up with the wrong permissions.

The open system call is apparently varargs, and the 3rd argument is only used if the 2nd argument has O_CREAT set.

Here's a minimal test that uses that call:
```
package syscall_test
import (
	"syscall"
	"testing"
)
func TestOpen(t *testing.T) {
	f, err := syscall.Open("/tmp/foo.dat", syscall.O_RDWR|syscall.O_CREAT, 0644)
	if err != nil {
		t.Fatalf("Open: %v", err)
	}
	if err := syscall.Close(f); err != nil {
		t.Fatalf("Close: %v", err)
	}
}
```
After running this on amd64,  ls -l /tmp/foo.dat shows the right permissions.
Removing the file and running on arm64 creates the file with incorrect permissions.

Here's a log showing lldb's bt thinks we passed the 3rd argument properly; evidently you can't trust bt to print varargs properly? 
```
% grep O_ syscall_libc_darwin.go
        O_RDWR   = 0x2
        O_CREAT  = 0x200
        ...
% perl -e "print 0x202"
514
% grep O_ open_test.go
        f, err := syscall.Open("/tmp/foo.dat", syscall.O_RDWR|syscall.O_CREAT, 0644)
% perl -e "print 0644"
420

--------- works on amd64 ---------
% GOARCH=amd64 ../../build/tinygo  test -c syscall
% file syscall.test
syscall.test: Mach-O 64-bit executable x86_64
% rm -f /tmp/foo.dat
% arch -x86_64 ./syscall.test
% ls -l /tmp/foo.dat
-rw-r--r--  1 dkegel  wheel  0 Jul 15 12:16 /tmp/foo.dat

--------- fails on arm64 ---------
% ../../build/tinygo test -c syscall
% file syscall.test
syscall.test: Mach-O 64-bit executable arm64
% rm -f /tmp/foo.dat
% ./syscall.test
% ls -l /tmp/foo.dat
-r--r-x---  1 dkegel  wheel  0 Jul 15 12:20 /tmp/foo.dat

--------- are wrong args passed to open()?  Well, lldb bt shows args look correct, but ... ---------

% lldb syscall.test
% b open
% r
...
% c
...
% c
Process 86029 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
    frame #0: 0x00000001a00fdfb8 libsystem_kernel.dylib`open
% bt
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
  * frame #0: 0x00000001a00fdfb8 libsystem_kernel.dylib`open
    frame #1: 0x0000000100000ba8 syscall.test`syscall.Open(path=<unavailable>, flag=514, mode=420) at syscall_libc.go:68:20 [opt]
```

aykevl:
Unfortunately the calling convention for variadic functions is different from the calling convention of regular functions on darwin/arm64, and `open` happens to be such a variadic function. The syscall package treated it like a regular function, which resulted in buggy behavior.

This fix introduces a wrapper function. This is the cleanest change I could come up with.

This fixes `tinygo test archive/zip` on darwin/arm64, and hopefully also fixes #2997.

### DIFF
--- a/src/syscall/syscall_libc_darwin.go
+++ b/src/syscall/syscall_libc_darwin.go
@@ -281,3 +281,8 @@ func libc_pipe(fds *int32) int32
 //
 //export getpagesize
 func libc_getpagesize() int32
+
+// int open(const char *pathname, int flags, mode_t mode);
+//
+//export syscall_libc_open
+func libc_open(pathname *byte, flags int32, mode uint32) int32
--- a/src/runtime/os_darwin.c
+++ b/src/runtime/os_darwin.c
@@ -1,0 +1,7 @@
+// Wrapper function because 'open' is a variadic function and variadic functions
+// use a different (incompatible) calling convention on darwin/arm64.
+
+#include <fcntl.h>
+int syscall_libc_open(const char *pathname, int flags, mode_t mode) {
+    return open(pathname, flags, mode);
+}
--- a/src/runtime/os_darwin.go
+++ b/src/runtime/os_darwin.go
@@ -5,6 +5,8 @@ package runtime
 
 import "unsafe"
 
+import "C" // dummy import so that os_darwin.c works
+
 const GOOS = "darwin"
 
 const (
--- a/src/syscall/syscall_libc.go
+++ b/src/syscall/syscall_libc.go
@@ -351,11 +351,6 @@ func libc_pread(fd int32, buf *byte, count uint, offset int64) int
 //export lseek
 func libc_lseek(fd int32, offset int64, whence int) int64
 
-// int open(const char *pathname, int flags, mode_t mode);
-//
-//export open
-func libc_open(pathname *byte, flags int32, mode uint32) int32
-
 // int close(int fd)
 //
 //export close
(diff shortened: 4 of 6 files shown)


src/syscall/syscall_libc_darwin.go
```diff
@@ -281,3 +281,8 @@ func libc_pipe(fds *int32) int32
 //
 //export getpagesize
 func libc_getpagesize() int32
+
+// int open(const char *pathname, int flags, mode_t mode);
```
dkegel-fastly, line 285: is this comment correct?  The function name may be wrong.   Or it could be fine!  I don't know the convention here.

dkegel-fastly (approved):
LGTM

deadprogram:
Thanks @aykevl for the fix and @dkegel-fastly for review. Now merging.

PR #3228, opened 2022-10-14
build/windows: use GH actions parallel execution to speed up build time

deadprogram:
This PR cuts build time in half by using parallel execution in Github Actions when performing the Windows build.

It splits the build into 4 different jobs, the first job does the building, and the other 3 each run their respective tests in parallel.

Before this change, total build time for Windows: `1h 0m 29s`

https://github.com/tinygo-org/tinygo/actions/runs/3249645395

With this PR , total build time for Windows: `35m 5s`

https://github.com/tinygo-org/tinygo/actions/runs/3251822513

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -95,10 +95,87 @@ jobs:
         with:
           name: release-double-zipped
           path: build/release/release.zip
+
+  smoke-test-windows:
+    runs-on: windows-2022
+    needs: build-windows
+    steps:
+      - uses: brechtm/setup-scoop@v2
+        with:
+          scoop_update: 'false'
+      - name: Install Dependencies
+        shell: bash
+        run: |
+          scoop install binaryen
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: '1.19'
+          cache: true
+      - name: Download TinyGo build
+        uses: actions/download-artifact@v2
+        with:
+          name: release-double-zipped
+          path: build/
+      - name: Unzip TinyGo build
+        shell: bash
+        working-directory: build
+        run: 7z x release.zip -r
       - name: Smoke tests
         shell: bash
-        run: make smoketest TINYGO=$(PWD)/build/tinygo AVR=0 XTENSA=0
+        run: make smoketest TINYGO=$(PWD)/build/tinygo/bin/tinygo AVR=0 XTENSA=0
+
+  stdlib-test-windows:
+    runs-on: windows-2022
+    needs: build-windows
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install Go
+        uses: actions/setup-go@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/windows.yml
```diff
@@ -95,10 +95,93 @@ jobs:
...
+        with:
+          scoop_update: 'false'
+      - name: Install Dependencies
+        shell: bash
+        run: |
+          scoop install binaryen
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: true
```
aykevl: Are you sure this is necessary? I believe TinyGo ships with binaryen (wasm-opt) already included.
The only thing that's not included is the `Makefile`, which you can get with a checkout without submodules (submodules slow down the checkout a lot).

.github/workflows/windows.yml
```diff
@@ -95,10 +95,93 @@ jobs:
...
+        run: make smoketest TINYGO=$(PWD)/build/tinygo/bin/tinygo AVR=0 XTENSA=0
+
+  stdlib-test-windows:
+    runs-on: windows-2022
+    needs: build-windows
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: true
```
aykevl: Same here, pretty sure you can skip the submodules.

.github/workflows/windows.yml
```diff
@@ -95,10 +95,93 @@ jobs:
...
+        with:
+          scoop_update: 'false'
+      - name: Install Dependencies
+        shell: bash
+        run: |
+          scoop install binaryen wasmtime
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: true
```
aykevl: Same here, I believe you can skip the submodules and binaryen.

.github/workflows/windows.yml
deadprogram: Doesn't this line in the `Makefile` indicate that the default for Windows is to use the system `wasm-opt` meaning I still need the scoop installed version?

https://github.com/tinygo-org/tinygo/blob/dev/Makefile#L132

Makes sense that I can skip the submodules in the checkout.

.github/workflows/windows.yml
deadprogram: Done.

.github/workflows/windows.yml
deadprogram: Can skip submodules but still need binaryen.

.github/workflows/windows.yml
aykevl: Hmm, I think you're right. I don't remember why this is necessary, would be nice to be able to build wasm-opt on Windows.

aykevl (approved):

aykevl:
This is a good improvement! Maybe we can do the same for assert-test-linux.

sago35 (approved):

deadprogram:
Thanks for the reviews @aykevl and @sago35 now merging.

Issue #2415, opened 2021-12-20
standalone test binaries should print `PASS` or `FAIL` when finished


dgryski:
```
~/go/src/go.googlesource.com/go/src/compress/flate $ tinygo test -c compress/flate && ./flate.test
~/go/src/go.googlesource.com/go/src/compress/flate $ go test -c &&  ./flate.test
PASS
```

aykevl:
Huh, that must have been a bug I've introduced here: https://github.com/tinygo-org/tinygo/pull/2043
So essentially what we need is that test binaries print "PASS" or "FAIL" but if they pass, they shouldn't print this (to avoid excessive output, like before #2043).

BerndCzech:
Can it be closed as #3229 was merged? Or can I help here?

dgryski:
It should be tagged as "next release". Bugs are usually closed when they are merged into the release branch, not just the development version.

deadprogram:
0.27.0 has now been released, so closing. Thank you!

PR #3229, opened 2022-10-14
Print PASS on pass when running standalone test binaries

fredgoya:
This commit allows tests run as standalone binaries to print 'PASS' on a test pass, to address https://github.com/tinygo-org/tinygo/issues/2415

It does this by adding a new test flag, test.batch, which is automatically passed to the test binaries when they are run as part of a batch (via `tinygo test`).

If the "test.batch" flag is *present*, current behavior is maintained; if it is *absent*, then a test suite pass will print out "PASS".

Maybe slightly inelegant, but I couldn't find any other way to suppress the PASS message when running as part of a batch of tests. Also, I'm not sure calling it a "batch" is strictly accurate when you can run just one suite with `tinygo test`, but it was the best name I could come up with.

### DIFF
--- a/main.go
+++ b/main.go
@@ -290,6 +290,7 @@ func Test(pkgName string, stdout, stderr io.Writer, options *compileopts.Options
 
 		// Run the test.
 		start := time.Now()
+		cmd.Args = append(cmd.Args, "--test.batch")
 		err = cmd.Run()
 		duration := time.Since(start)
 
--- a/main.go
+++ b/main.go
@@ -224,8 +224,9 @@ func Test(pkgName string, stdout, stderr io.Writer, options *compileopts.Options
 		flags = append(flags, "-test.benchmem")
 	}
 
+	buf := bytes.Buffer{}
 	passed := false
-	err = buildAndRun(pkgName, config, os.Stdout, flags, nil, 0, func(cmd *exec.Cmd, result builder.BuildResult) error {
+	err = buildAndRun(pkgName, config, &buf, flags, nil, 0, func(cmd *exec.Cmd, result builder.BuildResult) error {
 		if testCompileOnly || outpath != "" {
 			// Write test binary to the specified file name.
 			if outpath == "" {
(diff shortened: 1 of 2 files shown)


fredgoya:
I don't know how to run the tests that are failing the assert-test-linux build, so I'm not sure what to fix. Happy to tweak what I've got here if there's some way I can run them to verify, or to take any suggestions.

aykevl:
I looked into how this works in upstream Go and it is quite different. The test binary doesn't know whether it is running as a batch or is run directly, there is no `-test.batch` flag. Instead, what I assume is that the `go` command simply suppresses all output for passing tests (unless `-v` is given of course, in which case it also prints the `PASS` message along with all other output).
I think we should do the same in TinyGo.

aykevl:
I think what needs to change is this line here:

https://github.com/tinygo-org/tinygo/blob/b5ad81c884981f42c1ac491cc1d195aea886ddbe/main.go#L228

For `-v` runs the stdout can be `os.Stdout`, but when `-v` is not given I guess it can be a `bytes.Buffer` and be printed only if the test fails.
And at the same time, the test binary should start printing `PASS` at all times, not just when `-test.v` is given.

fredgoya:
Thanks for the tip! Reworked what I had and did my best to make it match the output of `go test` for the same tests.

dgryski:
I know upstream-Go had a change to the testing package so that test results were printed as they arrived, rather than being buffered until the end.  This helped diagnose which test was slow or failing during a run, or if you needed to kill a test partway through.  That behaviour is pretty handy.  We can maybe merge this improvement and open another bug to address that.

dgryski (approved):
LGTM

deadprogram:
Hello @fredgoya thanks for the code. Would you like to squash these commits now so we can merge?

dgryski:
@deadprogram I have a "squash-and-merge" button on this PR.  Doesn't that do the same thing?

deadprogram:
@dgryski yes. I asked in case @fredgoya wanted to do it, but we can do so as well, if they so wish.

fredgoya:
Thanks! I'm happy to let you do it. Feel free to press the button. :)

deadprogram:
Thank you @fredgoya now merged.

dgryski:
This broke benchmarking, as results are no longer printed without `-v`.

dgryski:
https://github.com/tinygo-org/tinygo/pull/3253

PR #3230, opened 2022-10-15
wasm: fix GC scanning of allocas

aykevl:
Scanning of allocas was entirely broken on WebAssembly. The code intended to do this was never run. There were also no tests.

Looking into this further, I found that it is actually not really necessary to do that: the C stack can be scanned conservatively and in fact this was already done for goroutine stacks (because they live on the heap and are always referenced). It wasn't done for the system stack however.

With these fixes, I believe code should be both faster *and* more correct.

I found this in my work to get opaque pointers supported in LLVM 15, because the code that was never reached now finally got run and was actually quite buggy.

### DIFF
--- a/src/runtime/gc_stack_portable.go
+++ b/src/runtime/gc_stack_portable.go
@@ -4,6 +4,7 @@
 package runtime
 
 import (
+	"internal/task"
 	"unsafe"
 )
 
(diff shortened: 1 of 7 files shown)


dgryski:
I'll review this Monday.

src/runtime/gc_stack_portable.go
```diff
@@ -17,19 +18,28 @@ type stackChainObject struct {
...
+//     though it may be referenced it will not be scanned by default.
+//
+// Therefore, we only need to scan the system stack.
+// It is relatively easy to scan the system stack while we're on it: we can
+// simply read __stack_pointer and __global_base and scan the area inbetween.
+// Unfortunately, it's hard to get the system stack pointer while we're on a
+// goroutine stack. But when we're on a goroutine stack, the system stack is in
+// the scheduler which means there shouldn't be anything on the system stack
+// anyway.
+// ...I hope this assumption holds, otherwise we will need to store the system
```
dgryski, line 34: Is there any way for us to check this assumption (even behind a build tag) and panic if it doesn't hold?

transform/llvm.go
```diff
@@ -75,26 +75,14 @@ func replaceGlobalIntWithArray(mod llvm.Module, name string, buf interface{}) ll
 	return global
 }
 
-// typeHasPointers returns whether this type is a pointer or contains pointers.
```
dgryski, line 78: This seems like a useful function, even if it's unused at the moment.  Do we have an equivalent function to answer this question?

dgryski (approved):
LGTM

src/runtime/gc_stack_portable.go
deadprogram: @aykevl what do you think?

transform/llvm.go
aykevl: I don't think so. But if needed, we can always bring it back from the git history. I don't like keeping dead code around.

src/runtime/gc_stack_portable.go
aykevl: As I understand it, the system stack is stored as a value in the local here:

https://github.com/tinygo-org/tinygo/blob/release/src/internal/task/task_asyncify_wasm.S#L73

This function is transformed by the asyncify pass so the `__stack_pointer` variable is stored somewhere on the system stack itself. It's probably possible to scan it, but not easily.

In any case, this is not a regression: the old code didn't scan the system stack either when it was on a goroutine stack.

deadprogram:
Thank you @aykevl and @dgryski now merging this fix!

PR #3231, opened 2022-10-16
machine/usb: rename 'New()' to 'Port()' to have the API better match what it does

deadprogram:
This PR modifies the `machine/usb` HID and MIDI package to rename 'New()' to 'Port()'. 

This is intended to have the API better match the singleton that is actually being returned, in part based on feedback during recent workshops.

It does unfortunately break the existing API slightly, but at a very minor cost vs. the readability.

### DIFF
--- a/src/machine/usb/hid/keyboard/keyboard.go
+++ b/src/machine/usb/hid/keyboard/keyboard.go
@@ -63,8 +63,8 @@ func init() {
 	}
 }
 
-// New returns hid-keybord.
-func New() *keyboard {
+// Port returns the USB hid-keyboard port.
+func Port() *keyboard {
 	return Keyboard
 }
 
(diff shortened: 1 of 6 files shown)


dgryski:
We could probably write a quick semgrep rule to handle the API migration.

sago35 (commented):
I think it is good.

However, I have written a book in Japanese and would like to see the New() function to stay around for a while.
I would like Port() to be used primarily, but leave New() for a while with a comment that it is deprecated.

deadprogram:
OK, added back `New()` with comment as you suggested @sago35

sago35 (approved):
LGTM.

src/machine/usb/hid/keyboard/keyboard.go
```diff
@@ -63,8 +63,14 @@ func init() {
 	}
 }
 
-// New returns hid-keybord.
+// New returns the USB hid-keyboard port.
+// Deprecated, better to just use Port()
```
dgryski, line 67: If the deprecation comment starts a new paragraph and uses a colon, it will be flagged up by standard Go tooling (like staticcheck) at call sites
```
//
// Deprecated: Use Port()
```

PR #3232, opened 2022-10-16
Fix bug introduced in #3207

aykevl:
I introduced a bug in https://github.com/tinygo-org/tinygo/pull/3207: looking up a VID/PID pair would result in a slice out of bounds panic. This is a trivial fix for that bug.

I also fixed a cosmetic issue in an error message to be more readable.


dgryski (approved):
LGTM. Is there a test that could have caught this?

deadprogram:
Thanks @aykevl merging now,

aykevl:
> LGTM. Is there a test that could have caught this?

Yes... I would probably have caught this if I actually tried to flash anything. It seems like I didn't try.
But not something that CI could easily do.

PR #3234, opened 2022-10-17
sync: implement simple pooling for sync.Pool

anuraaga:
We use sync.Pool to reuse some buffers and I was surprised to see TinyGo doesn't do any pooling. This adds some simple pooling, with the same caveats as exist in the `sync` package docs, not safe for concurrent use but otherwise it will pool as most users expect (I presume).

### DIFF
--- a/src/sync/pool.go
+++ b/src/sync/pool.go
@@ -1,13 +1,18 @@
 package sync
 
-// Pool is a very simple implementation of sync.Pool. It does not actually
-// implement a pool.
+// Pool is a very simple implementation of sync.Pool.
 type Pool struct {
-	New func() interface{}
+	New   func() interface{}
+	items []interface{}
 }
 
 // Get returns the value of calling Pool.New().
 func (p *Pool) Get() interface{} {
+	if len(p.items) > 0 {
+		x := p.items[len(p.items)-1]
+		p.items = p.items[:len(p.items)-1]
+		return x
+	}
 	if p.New == nil {
 		return nil
 	}
@@ -16,4 +21,5 @@ func (p *Pool) Get() interface{} {
 
 // Put drops the value put into the pool.
 func (p *Pool) Put(x interface{}) {
+	p.items = append(p.items, x)
 }
--- a/src/sync/pool.go
+++ b/src/sync/pool.go
@@ -6,7 +6,7 @@ type Pool struct {
 	items []interface{}
 }
 
-// Get returns the value of calling Pool.New().
+// Get returns an item in the pool, or the value of calling Pool.New() if there are no items.
 func (p *Pool) Get() interface{} {
 	if len(p.items) > 0 {
 		x := p.items[len(p.items)-1]
@@ -19,7 +19,7 @@ func (p *Pool) Get() interface{} {
 	return p.New()
 }
 
-// Put drops the value put into the pool.
+// Put adds a value back into the pool.
 func (p *Pool) Put(x interface{}) {
 	p.items = append(p.items, x)
 }
--- a/src/sync/pool_test.go
+++ b/src/sync/pool_test.go
@@ -1,0 +1,50 @@
+package sync_test
+
+import (
+	"sync"
+	"testing"
+)
+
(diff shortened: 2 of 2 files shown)


dgryski:
I'm all for an implementation of `sync.Pool` that does minimal pooling, but having it not safe for concurrent use by default sounds like a pretty big footgun.  Could you add a simple `sync.Mutex` ?

src/sync/pool.go
```diff
@@ -1,19 +1,25 @@
...
 	if p.New == nil {
 		return nil
 	}
 	return p.New()
 }
 
-// Put drops the value put into the pool.
+// Put adds a value back into the pool.
 func (p *Pool) Put(x interface{}) {
+	p.items = append(p.items, x)
```
dgryski, line 24: The upstream `sync.Pool` is cleared every time the garbage collector runs.  This prevents memory leaks from happening.  This implementation doesn't have the safe-guard.  Maybe we could limit the number of elements in the pool to some small constant or somehow expire items (ad their points) to prevent unbounded memory growth?

src/sync/pool.go
```diff
@@ -1,19 +1,25 @@
 package sync
 
-// Pool is a very simple implementation of sync.Pool. It does not actually
-// implement a pool.
+// Pool is a very simple implementation of sync.Pool.
 type Pool struct {
-	New func() interface{}
+	New   func() interface{}
+	items []interface{}
```
dgryski, line 6: Please add a `sync.Mutex` so this type is safe for concurrent use.

dgryski (changes requested):

src/sync/pool.go
anuraaga: The sync package documents not being safe to use from multiple threads

https://github.com/tinygo-org/tinygo/blob/release/src/sync/doc.go#L2

And we can see the mutex implementation is not really locking for concurrent use

https://github.com/tinygo-org/tinygo/blob/release/src/sync/mutex.go#L24

So I guess it's better to leave out `sync.Mutex` to better reflect the reality that it's not safe?

src/sync/pool.go
anuraaga: Yeah did think about this. FWIU, the main motivation for the upstream clearing is not to prevent unbounded memory growth, it is the job of the code to Get and Put in pairs to prevent this, but to allow periods of high traffic to have a larger pool but to not keep the larger pool during low traffic. This is a nice to have optimization but I don't know if it's critical.

I can't think of an appropriate constant that wouldn't be too small or too large, and trying an approach of frequency tracking I think could become quite complicated. Any suggestion on an algorithm that wouldn't be too much?

src/sync/pool.go
dgryski: @aykevl Do you want to weigh in with an opinion here?

aykevl: Hmm, not sure about this one to be honest.
_Right now_ we don't support parallelism and there are no blocking operations in this code so the code is safe. Once we add parallelism this code will become unsafe, along with many other parts of TinyGo internal code (like channels).

It does seem reasonable to me to use a mutex here, just to make sure this part of the code is forward compatible. But I would also be fine leaving it as-is, as adding parallelism will require reviewing lots of concurrency related code anyway.

Sidenote: concurrency and parallism aren't the same thing. TinyGo fully support concurrency (multiple goroutines). It doesn't support parallelism (goroutines running _at the same time_ on different CPU cores).

src/sync/pool.go
aykevl: I think we can leave this optimization out for now. When in doubt, I prefer the simpler solution (where simpler usually but not always means less code). We can always add this optimization when needed.

aykevl (commented):
See my comments below. Other than that, this looks good to me.

dgryski (approved):
Follwing @aykevl's comments, LGTM

deadprogram:
Thank you very much @anuraaga for the improvements, and to @dgryski and @aykevl for review.

Now merging.

PR #3238, opened 2022-10-18
corpus: remove golang.org/x/crypto/internal/subtle

aykevl:
This subdirectory appears to be gone now.

Perhaps a more future-proof way to fix this would be to test particular versions or hashes, instead of simply the latest version? That way we're always testing the same code.

### DIFF
--- a/testdata/corpus.yaml
+++ b/testdata/corpus.yaml
@@ -113,7 +113,6 @@
   - pkg: curve25519
   - pkg: ed25519
   - pkg: hkdf
-  - pkg: internal/subtle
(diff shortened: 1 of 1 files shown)


dgryski (approved):
LGTM

dgryski:
I think you can actually specify versions in the yaml file, but we just don't (yet).

dgryski:
Weird Mac build failure.   I think we should merge this as the CI issue is unrelated to this patch.

deadprogram:
Thanks everyone! Now merging.

PR #3239, opened 2022-10-18
corpus: remove 'noasm' from some tests

aykevl:
Unfortunately I couldn't fully test these changes (because I'm on linux/arm64), but they don't seem to be needed on linux/amd64.


dgryski (approved):
LGTM

deadprogram:
:scissors: thanks @aykevl and @dgryski now merging.

Issue #2589, opened 2022-01-27
PIO Support for Raspberry Pi Pico


bmentink:
Hi,
Is there any support for the programmable IO module on the Pico? If not, will it be added at some point.
Also, can not find how to do interrupts on GPIO's or ADC etc etc
No idea howto print messages to the USB/Serial port either with printf ..

Maybe a list of what works and what is TODO would be great ..... I can see a lot of projects for this board and TinyGO

Many Thanks in advance ..

deadprogram:
Please see https://github.com/tinygo-org/tinygo/issues/1953 regarding PIO.

bmentink:
Thanks, so WIP ... good to see ..

Any idea how to see "println" messages on the pico?  

EDIT: Never mind, sniffed around with the scope and found it on GPIO_0/1, I have attached an HC-05 bluetooth module, and can see the println's just fine ... I found the baud rate was 115200 by trial and error ..

sgall17a:
I am a complete newbie in Tinygo and apologise for being so presumptive since I don't really know what I am talking about but here goes anyway.

The rp2 library in Micropython uses a simple method for assembling the pio instructions.
pythonfunction(parameters)[waitstates]
In PIO all instructions, parameters and wait states are packed into one 32bit word.
1.  The python function corresponds to the instruction bits.
2.  The parameter for the function builds the next few  "parameter" bits.
3. The square brackets indicate an "attribute" in python and build the (optional) wait bits.  The function returns a python object and this can optionally have attributes to indicate wait values.  There are a few rules limiting the maximum number of wait states which can vary by instruction (and ?globally for the statemachine I think).

5.  "label" is also a python function which can manage the PC.

The assembled PIO program would end up as a slice of 32bit words.

Converting the Python to GO could be done fairly easily ( I think).     
Gluing it into Go could be harder (especially since I dont know much about Go internals).
What I think you would need are (mostly housekeeping):
1.  Some way of poking the go array into the PIO memory.
2. Some way of setting the control registrars so we can have more than one PIO program and communicate with the PIO.
3. Some way of talking to the PIO FIFO's.  Go channels would seem to be great way to do that for both receive and transmit channels.
4. Some way of inserting an external execute instruction.  This is just a register write.
6.  Some way to start and stop the state machine.  This is just a register write.  

[...]

Issue #1953, opened 2021-06-16
Design: How should we support RP2040 PIO?


kenbell:
This issue is a placeholder to discuss how we should support RP2040 PIO

### Background
The Pico SDK comes with pioasm, which is a separate tool to assemble PIO instructions (+ interleaved Python / C).  It's run as a code generator, run before compiling the main application logic and generating python/C code.  `pioasm` can also output the binary as hex.

CircuitPython comes with it's own assembler (adafruit_pioasm) which has a few language differences.  It can be used inline in CircuitPython code:
```python
hello = """
.program hello
loop:
    pull
    out pins, 1
; This program uses a 'jmp' at the end to follow the example.  However,
; in a many cases (including this one!) there is no jmp needed at the end
; and the default "wrap" behavior will automatically return to the "pull"
; instruction at the beginning.
    jmp loop
"""

assembled = adafruit_pioasm.assemble(hello)
```

### TinyGo Options
What model should we use for TinyGo?

Some options:

1. Use pioasm with `//go:generate`
2. Have an inline primitive, similar to `arm.Asm`
3. Allow coding by doing function call per instruction, e.g  `pioapp.AddInstruction(pio.MOV,xxx,xx,xx)`

Quick evaluation...
|Option|Pros|Cons|
|-|-|-|
|1|Quick to implement|Extra dependencies to use TinyGo - install/compile Pico SDK|
|2|Familiar model (like `arm.Asm`) and similar(-ish) to CircuitPython|Relatively complex to implement pioasm in Go, some complexity in TinyGo to treat `pio.Asm` as a special case|
|3|Relatively quick to implement|Relatively non-intuitive (personal opinion?)|

### References:
https://datasheets.raspberrypi.org/rp2040/rp2040-datasheet.pdf section 3.3
https://datasheets.raspberrypi.org/pico/raspberry-pi-pico-c-sdk.pdf section 3.3
https://learn.adafruit.com/intro-to-rp2040-pio-with-circuitpython

kenbell:
My personal preference is to do `pio.Asm` approach.  Apps using PIO would be something like this, based on CircuitPython:
```go
import "machine/pico/pio"

hello := pio.Asm(`
.program hello
loop:
    pull
    out pins, 1
; This program uses a 'jmp' at the end to follow the example.  However,
; in a many cases (including this one!) there is no jmp needed at the end
; and the default "wrap" behavior will automatically return to the "pull"
; instruction at the beginning.
    jmp loop
[...]

PR #3242, opened 2022-10-20
Add basic GPIO support for rp2040 PIO

MJDrgn:
Code stolen from #1983
Also see discussion #1953 and #2589  

Adding the constants for Pin types to support PIO means that any implementation can be done in user code, until a final API is implemented in core for PIO compilation and instruction loading (which would be a more complicated and debatable PR)


deadprogram (approved):
That is a good idea @MJDrgn I'm sure @kenbell won't mind a bit. Thanks for helping keep this moving forward.

PR #3245, opened 2022-10-21
Allow custom wasm malloc implementation

anuraaga:
This moves the existing code for overriding malloc for wasm into a separate file with a `custommalloc` build tag defined to allow excluding it. This allows apps to bring their own malloc for wasm.

I could confirm that with this change, I could swap malloc with [mimalloc](https://github.com/microsoft/mimalloc) and then also swap the GC to one that uses malloc for the underlying heap instead of taking full ownership.

https://github.com/corazawaf/coraza-proxy-wasm/compare/main...anuraaga:mimalloc?expand=1

This dramatically improves performance of the polyglot Envoy wasm filter, bringing pauses from ~1s to ~100ms, and also gives me a playground to make more changes to the GC that aren't "tiny" :) Note that the current GC may be improvable - I tried playing with `isPointerlike` to have it recognize malloc'd pointers as not Go objects, and didn't have much performance difference but may have not finished it properly. Either way, we would expect it to be significantly better to have malloc'd objects outside of the Go GC completely anyways.

This would also resolve the issue in #3162 - no problem making the default malloc tiny and possibly `O(n)` if it can be replaced easily.


anuraaga:
@dgryski @aykevl would it be possible to get a review on this? Thanks!

dgryski (approved):
LGTM; this is only shuffling around some code and adding build tags.

deadprogram:
Thanks for the improvement @anuraaga and to @dgryski for review. Now going to squash/merge.

PR #3246, opened 2022-10-21
Look for LLVM 15 in makefile

anuraaga:
#3189 added support for LLVM 15 but I think this also needs to be updated when it's only available as `clang-15`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ LLD_SRC ?= $(LLVM_PROJECTDIR)/lld
 
 # Try to autodetect LLVM build tools.
 # Versions are listed here in descending priority order.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -10,7 +10,7 @@ LLD_SRC ?= $(LLVM_PROJECTDIR)/lld
 
 # Try to autodetect LLVM build tools.
 # Versions are listed here in descending priority order.
-LLVM_VERSIONS = 14 13 12 11
+LLVM_VERSIONS = 15 14 13 12 11
```
aykevl, line 13: I always forget these 🙈

aykevl (approved):

deadprogram:
Thanks for the fix @anuraaga

PR #3249, opened 2022-10-21
main: fix outputting .ll files

aykevl:
This command didn't work anymore since the refactor in #3211. The reason it doesn't work anymore is that before the refactor, the code in the callback wasn't executed for .ll, .bc and .o files but after the refactor it is, which causes a spurious error.

This commit fixes this oversight.

Example that didn't work anymore and is fixed with this change:

    tinygo build -o test.ll examples/serial

### DIFF
--- a/main.go
+++ b/main.go
@@ -168,37 +168,44 @@ func Build(pkgName, outpath string, options *compileopts.Options) error {
 		return err
 	}
 
-	if outpath == "" {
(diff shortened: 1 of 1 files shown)


dgryski (approved):
I believe you.  LGTM

deadprogram:
Don't stop believing! :Of course, the tests still pass too. smiley_cat: 

Now merging, thanks everyone!

PR #3250, opened 2022-10-21
avoid allocating clock on heap [rp2040]

soypat:
One less heap allocation since the type already has a pointer to struct. This function and its result is used exclusively in `func (clks *clocksType) init()` as far as my IDE can tell. `time.Sleep` is still working after this change.


deadprogram (approved):
Every little bit helps! Thanks @soypat

Issue #3085, opened 2022-08-21
Build error: type reflect.Type has no field or method FieldByIndex


nuxart:
go version go1.18.4 darwin/amd64
tinygo version 0.25.0 darwin/amd64 (using go version go1.18.4 and LLVM version 14.0.0)
```
tinygo build -o main.wasm

# encoding/xml
/usr/local/go/src/encoding/xml/typeinfo.go:318:14: typ.FieldByIndex undefined (type reflect.Type has no field or method FieldByIndex)
/usr/local/go/src/encoding/xml/typeinfo.go:319:14: typ.FieldByIndex undefined (type reflect.Type has no field or method FieldByIndex)
```

gedw99:
Me too on 2 other packages today . Xml encoding and decoding

deadprogram:
0.27.0 has now been released, so closing. Thank you!

PR #3252, opened 2022-10-24
Stub out reflect.Type FieldByIndex

anuraaga:
The same named function on `reflect.Value` is stubbed similarly so there seems to be precedent. This is high impact because currently the `encoding/xml` package can't be compiled at all - with this change, manual use of `xml.Decoder` without reflection-based marshaling works fine with TinyGo so some sort of XML usage becomes possible.

Related #3085 which would be solved by actually implementing this for reflection-based XML usage, but unblocking non-reflection would be good

### DIFF
--- a/src/reflect/type.go
+++ b/src/reflect/type.go
@@ -295,7 +295,7 @@ type Type interface {
 	// to the index sequence. It is equivalent to calling Field
 	// successively for each index i.
 	// It panics if the type's Kind is not Struct.
-	//FieldByIndex(index []int) StructField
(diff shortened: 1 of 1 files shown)


jcchavezs (approved):

dgryski (approved):
LGTM

gedw99:
+1. at least can move forward

deadprogram:
> with this change, manual use of xml.Decoder without reflection-based marshaling works fine with TinyGo so some sort of XML usage becomes possible

That is very reasonable to me! Thanks @anuraaga

PR #3253, opened 2022-10-24
fix printing of benchmark output

dgryski:


deadprogram (approved):
Thanks for the fix @dgryski looks good!

PR #3256, opened 2022-10-28
all: remove libffi warning on macos

aykevl:
Running `go install` on MacOS produces the following warning:

    # github.com/tinygo-org/tinygo
    ld: warning: directory not found for option '-L/opt/homebrew/opt/libffi/lib'

It doesn't look like libffi is used anywhere, so I simply removed it. Not sure why it was included in the first place.

(I updated the Makefile for consistency, but we really should be removing that Makefile especially because the Go bindings are removed in upstream LLVM).

---

This is just a small papercut fix to remove the warning during build. I don't believe it changes the resulting `tinygo` binary in any meaningful way.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	golang.org/x/sys v0.0.0-20220829200755-d48e67d00261
 	golang.org/x/tools v0.1.11
 	gopkg.in/yaml.v2 v2.4.0
-	tinygo.org/x/go-llvm v0.0.0-20220922115213-dcb078a26266
+	tinygo.org/x/go-llvm v0.0.0-20221028183034-8341240c0b32
 )
 
 require (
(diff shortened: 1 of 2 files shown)


deadprogram:
Interestingly, the CI seems to fail only on macOS. I retried several times...

aykevl:
Each time it failed in the timers test. It's a flaky test, but failing 4 times in a row seems unlikely. Still, I have restarted the build once again to see if it is just the CI acting up.

aykevl:
Seems to be fixed now! I'm guessing we were just really unlucky, or the CI system was acting up yesterday. I would be rather surprised if removing libffi would result in testdata/timers.go to fail.

deadprogram:
Thanks for the cleanup @aykevl now merging.

PR #3257, opened 2022-10-28
all: use DWARF version 4

aykevl:
This should hopefully fix the following issue when running `tinygo gdb`:

    DW_FORM_rnglistx index pointing outside of .debug_rnglists offset array [in module /tmp/tinygo4013272868/main]

I didn't actually test this branch. I hope to do so eventually. But I think I tested this particular change before and it worked at the time. I did test that the resulting binaries are indeed DWARF v4, and not the newer DWARF v5 which appears to have bugs.

### DIFF
--- a/builder/library.go
+++ b/builder/library.go
@@ -142,7 +142,7 @@ func (l *Library) load(config *compileopts.Config, tmpdir string) (job *compileJ
 	// Note: -fdebug-prefix-map is necessary to make the output archive
 	// reproducible. Otherwise the temporary directory is stored in the archive
 	// itself, which varies each run.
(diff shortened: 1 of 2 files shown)


dgryski (approved):
LGTM

aykevl:
I have tested this on Linux and it indeed fixes the issue with `DW_FORM_rnglistx`. Specifically:

- GDB 10 + DWARF 4 works fine
- GDB 10 + DWARF 5 has this issue
- GDB 12 + DWARF 5 works fine

So I believe the bug is that older GDB versions can't correctly deal with DWARF 5. It seems reasonable to stay with DWARF 4 for now for that reason.
(One gnarly issue in the future will be that GDB versions that are part of cross compilation toolchains tend to be heavily out of date, so we might need to stay at DWARF 4 for a while for these chips).

deadprogram:
Thank you @aykevl and @dgryski now merging.

PR #3260, opened 2022-10-30
cgo: support anonymous enums included in multiple Go files

aykevl:
Anonymous enums (often used in typedefs) triggered a problem that was already solved for structs but wasn't yet solved for enums. So this patch generalizes the code to work for both structs and enums, and adds testing for both.

### DIFF
--- a/cgo/libclang.go
+++ b/cgo/libclang.go
@@ -533,6 +533,26 @@ func tinygo_clang_globals_visitor(c, parent C.GoCXCursor, client_data C.CXClient
 	return C.CXChildVisit_Continue
 }
 
+// Get the precise location in the source code. Used for uniquely identifying
+// source locations.
+func (f *cgoFile) getUniqueLocationID(pos token.Pos, cursor C.GoCXCursor) interface{} {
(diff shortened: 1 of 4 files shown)


aykevl:
...that homebrew-install failure is really strange.

deadprogram:
```
panic: incorrect build: using LLVM version 15.0.2 in the tinygo.org/x/llvm package, and version 15.0.3 in the ./cgo package
```

That is indeed a strange error from Homebrew...

deadprogram:
@aykevl I think this is probably problem due to Go modules cache. Consider changing https://github.com/tinygo-org/tinygo/blob/cgo-anonymous-enum/.github/workflows/build-macos.yml#L118

aykevl:
@deadprogram yeah, that seems like it!
I've rebased against the dev branch, which appears to fix the issue for now (the cache is invalidated because go.sum was updated). Hopefully it won't happen again. If it does, we may need to undo the simplification made in https://github.com/tinygo-org/tinygo/pull/3140.

deadprogram:
Seems good now @aykevl thanks! Now merging.

PR #3264, opened 2022-11-03
esp: use LLVM Xtensa linker instead of Espressif toolchain

aykevl:
The Espressif fork of LLVM now has Xtensa support in the linker LLD. (This support was written mosly by me). This means we don't have to use the Espressif GNU toolchain anymore and makes installing TinyGo simpler.

In the future, this also paves the way for ThinLTO support for Espressif chips. Right now it is mostly just a way to simplify TinyGo installation and speed up CI slightly.

---

(Note: this PR probably won't pass CI, I will update it later).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,18 +15,8 @@ FROM tinygo-llvm AS tinygo-llvm-build
 RUN cd /tinygo/ && \
     make llvm-build
 
-# tinygo-xtensa stage installs tools needed for ESP32
-FROM tinygo-llvm-build AS tinygo-xtensa
-
-ARG xtensa_version="1.22.0-80-g6c4433a-5.2.0"
-RUN cd /tmp/ && \
-    wget -q https://dl.espressif.com/dl/xtensa-esp32-elf-linux64-${xtensa_version}.tar.gz && \
-    tar xzf xtensa-esp32-elf-linux64-${xtensa_version}.tar.gz && \
-    cp ./xtensa-esp32-elf/bin/xtensa-esp32-elf-ld /usr/local/bin/ && \
-    rm -rf /tmp/xtensa*
-
 # tinygo-compiler stage builds the compiler itself
-FROM tinygo-xtensa AS tinygo-compiler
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -15,18 +15,8 @@ FROM tinygo-llvm AS tinygo-llvm-build
...
-ARG xtensa_version="1.22.0-80-g6c4433a-5.2.0"
-RUN cd /tmp/ && \
-    wget -q https://dl.espressif.com/dl/xtensa-esp32-elf-linux64-${xtensa_version}.tar.gz && \
-    tar xzf xtensa-esp32-elf-linux64-${xtensa_version}.tar.gz && \
-    cp ./xtensa-esp32-elf/bin/xtensa-esp32-elf-ld /usr/local/bin/ && \
-    rm -rf /tmp/xtensa*
-
 # tinygo-compiler stage builds the compiler itself
-FROM tinygo-xtensa AS tinygo-compiler
+FROM tinygo-llvm-build AS tinygo-compiler
```
aykevl, line 19: I'd like some extra eyes on this change, I'm not very experienced with Dockerfiles. The main change is that installing xtensa-esp32-elf-ld should be skipped.

deadprogram: That certainly looks correct to me.

deadprogram:
Do you also need to do something about https://github.com/tinygo-org/tinygo/blob/dev/Makefile#L697-L708 ?

aykevl:
> Do you also need to do something about https://github.com/tinygo-org/tinygo/blob/dev/Makefile#L697-L708 ?

No. The `XTENSA=0` option is still needed for the CircleCI tests.

aykevl:
Pushed again, now with all CI changes included. Hoping the CI will pass...

(This is going to take a few hours)

aykevl:
All tests pass, except for a few TinyHCI tests that look like a problem with the CI.

deadprogram:
This is going to make it SO much easier for people who want to program ESP32 and friends!

Now merging, thank you for your upstream changes @aykevl

PR #3265, opened 2022-11-03
wasm: remove -wasm-abi= flag

aykevl:
This flag controls whether to convert external i64 parameters for use in a browser-like environment.

This flag was needed in the past because back then we only supported wasm on browsers but no WASI. Now, I can't think of a reason why anybody would want to change the default. For `-target=wasm` (used for browser-like environments), the wasm_exec.js file expects this i64-via-stack ABI, which is the default. For WASI, there is no limitation on i64 values and `-wasm-abi=generic` is the default.

So far, all the times I've seen this flag in the wild it wasn't necessary and could just as well be left out. So the main purpose of this PR is to avoid confusion and unnecessary complexity.


dgryski (approved):
LGTM

deadprogram:
Less is better! Thanks @aykevl now merging.

PR #3269, opened 2022-11-05
darwin: fix error with `tinygo lldb`

aykevl:
Before this patch, `tinygo lldb path/to/package` would result in an error:

    (lldb) target create --arch=arm64-unknown-macosx10.12.0 "/var/folders/17/btmpymwj0wv6n50cmslwyr900000gn/T/tinygo2731663853/main"
    error: the specified architecture 'arm64-unknown-macosx10.12.0' is not compatible with 'arm64-apple-macosx10.12.0' in '/var/folders/17/btmpymwj0wv6n50cmslwyr900000gn/T/tinygo2731663853/main'

This patch fixes this error.

Unfortunately, it doesn't get debug information to work yet. I still haven't figured out what's going wrong here. But it's progress, I guess.

### DIFF
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -195,6 +195,7 @@ func LoadTarget(options *Options) (*TargetSpec, error) {
 		default:
 			llvmarch = options.GOARCH
 		}
+		llvmvendor := "unknown"
 		llvmos := options.GOOS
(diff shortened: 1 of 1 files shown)


deadprogram (approved):
Thanks for the fix @aykevl

PR #3273, opened 2022-11-05
builder: print compiler commands while building a library

aykevl:
We should have been doing this all along and makes it easier to debug things that go wrong while compiling a library.


deadprogram (approved):

deadprogram:

PR #3275, opened 2022-11-06
main: more attempts to find MSD volume when flashing

deadprogram:
This PR adds a `retries` flag to set the number of times to retry locating the MSD volume to be used for flashing UF2 devices. The reason we might need to increase this value is to try to let slower systems handle after a reset when the `flash` command needs to locate the correct MSD volume.

### DIFF
--- a/main.go
+++ b/main.go
@@ -928,8 +928,6 @@ func touchSerialPortAt1200bps(port string) (err error) {
 	return fmt.Errorf("opening port: %s", err)
 }
 
-const maxMSDRetries = 10
-
 func flashUF2UsingMSD(volume, tmppath string, options *compileopts.Options) error {
 	// find standard UF2 info path
 	var infoPath string
@@ -951,7 +949,7 @@ func flashUF2UsingMSD(volume, tmppath string, options *compileopts.Options) erro
 		infoPath = path + "/INFO_UF2.TXT"
 	}
 
-	d, err := locateDevice(volume, infoPath)
+	d, err := locateDevice(volume, infoPath, options.Retries)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


deadprogram:
After playing around with settings on the TinyHCI build machine, I cannot get it to mount any faster after flashing new code on the RP2040. Allowing a longer time than the current hard-wired value of 10 retries * 500 milliseconds lets the CI test pass.

main.go
```diff
@@ -1398,6 +1396,7 @@ func main() {
 	ocdCommandsString := flag.String("ocd-commands", "", "OpenOCD commands, overriding target spec (can specify multiple separated by commas)")
 	ocdOutput := flag.Bool("ocd-output", false, "print OCD daemon output during debug")
 	port := flag.String("port", "", "flash port (can specify multiple candidates separated by commas)")
+	maxRetries := flag.Int("retries", 10, "the maximum number of times to retry locating the MSD volume to be used for flashing")
```
ysoldak, line 1399: IMHO, "retries" is very generic, shall have "msd" in the name? "msd-retries" or something like it?

deadprogram: Perhaps `flash-retries` since it specifically has to do with the `tinygo flash` command?

```
tinygo flash -target=pico -flash-retries=15 -port=/dev/ttyACM0 ./path.to/code
```

deadprogram: Hmm, for some reason `retries` just reads better to me:

```
tinygo flash -target=pico -retries=15 -port=/dev/ttyACM0 ./path.to/code
```

Probably because the word "flash" already appears once. Seems like stuttering.

ysoldak: If it does work only for flash command, then yes, “retries” shall be ok actually.
When “retries” specified for, say, build or debug command, it is just ignored or an error happens?

deadprogram: The flag is just ignored for those commands.

sago35 (approved):
LGTM

deadprogram:
Thanks for the feedback/review @ysoldak and @sago35 now merging.

aykevl:
What about using a timeout instead of a number of retries? I think that better captures what the user intends.

deadprogram:
I think we should change to that, yes.

PR #3280, opened 2022-11-09
Enable wasm pointer tracking for gc=none.

### LINKED ISSUE #3278: gc=custom build option to allow custom GC implementations
In [coraza-proxy-wasm](https://github.com/corazawaf/coraza-proxy-wasm), we are creating a wasm binary for use with Envoy in cases where performance matters, while code size doesn't so much. I've previously mentioned gaining a lot of performance by using `mimalloc` for malloc/free instead of delegating to the TinyGo heap for them. This involved `-gc=none` and swapping `runtime.alloc` with the same code as currently in `gc_conservative.go`, except using a fixed-size malloc'd buffer as the heap instead of taking control of the entire wasm heap.

Naturally this has issues of being a fixed size heap, and while performance was much better with the non-Go allocations off the Go heap (no 1s pauses), it still needed to be improved (~100ms pauses). So the next step was replacing the GC with one that isn't tiny, ideally without having to actually write one.

I succeeded in swaping out the GC with `bdwgc` compiled to wasm. Performance is much better, with pauses <10ms from what I can tell. It's a number that is now unnoticable when just running through integration test requests while previously they were clearly noticable.

https://github.com/corazawaf/coraza-proxy-wasm/compare/main...anuraaga:bdwgc?expand=1#diff-eb1422fbb127355685df6612c55a2796be3543928fc25daec007bf9348753104R77

The biggest problem was marking tracked pointers in GC stack slots, and while I somehow found #3277 during my debugging, actually my root cause was much simpler, that stack slot tracking is disabled with `-gc=none` 😂 The input I got while finding #3277 really helped me to find this though

https://github.com/tinygo-org/tinygo/blob/release/compileopts/config.go#L102

I was able to get my custom GC working with this change to TinyGo

https://github.com/tinygo-org/tinygo/compare/dev...anuraaga:gc-hack?expand=1

Naturally, it seems to make sense to disable stack tracking when `gc=none` given it's less for custom GC's and more for disabling GC. Would it be reasonable to have a separate `gc=custom` flag that would enable stack tracking, and also stub out [initHeap](https://github.com/tinygo-org/tinygo/blob/release/src/runtime/gc_none.go#L38) to allow initialization logic to run in the correct order (I had to work around this once in https://github.com/corazawaf/coraza-proxy-wasm/pull/63 and it's still fragile)?

Or otherwise, perhaps we can remove the guards on `gc.conservative` enabling stack tracking, i.e. apply my two-line change above? I guess there won't be any tracking if `gc=none` and no allocations are actually made so it wouldn't hurt to allow the transformation to run in that case too, maybe slightly slower during compile but I'd expect no runtime effect

anuraaga:
Fixes #3278 

This takes the simpler approach of allowing wasm pointer tracking on the stack to be enabled even with `gc=none`. none is [documented](https://github.com/tinygo-org/tinygo/blob/release/src/runtime/gc_none.go#L8) as being useful for combining with a separate garbage collector. However in practice, it is basically impossible to do this with wasm because tracking pointers on the stack is required for any GC to be able to see them to mark them.

It seems reasonable to enable the stack tracking code even when not using the TinyGo GC. Either a) an external GC is being used and the tracking is required or b) an external GC is not being used meaning there are no allocations, and therefore no tracking will actually happen.

Having the ability to swap in a custom GC allows TinyGo compiled wasm to perform quite well with low pause times, so this is very enabling for projects like [coraza-proxy-wasm](https://github.com/corazawaf/coraza-proxy-wasm)

### DIFF
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -105,6 +105,8 @@ func (c *Config) GC() string {
 // that can be traced by the garbage collector.
 func (c *Config) NeedsStackObjects() bool {
 	switch c.GC() {
+	case "none":
+		fallthrough
 	case "conservative":
 		for _, tag := range c.BuildTags() {
 			if tag == "tinygo.wasm" {
--- a/src/runtime/gc_stack_portable.go
+++ b/src/runtime/gc_stack_portable.go
@@ -1,5 +1,6 @@
-//go:build gc.conservative && tinygo.wasm
-// +build gc.conservative,tinygo.wasm
+//go:build (gc.conservative || gc.none) && tinygo.wasm
+// +build gc.conservative gc.none
+// +build tinygo.wasm
 
 package runtime
 


dgryski (approved):
LGTM.  Seems safe enough.

deadprogram:
Thank you for the addition @anuraaga and to @dgryski for reviewing. Now merging.

aykevl:
I don't think we should do this, to be honest. There are many things needed to integrate a GC, and `-gc=none` is not really able to capture it all. The `-gc=none` flag was mostly intended for debugging and to make sure binaries have the smallest possible code size. Enabling pointer tracking defeats this purpose.

deadprogram:
Hmm, perhaps I merged too quickly then. :crying_cat_face:

aykevl:
On the other hand, while I'm not entirely sold on `-gc=custom` yet, if we are going to support custom GCs then using `-gc=custom` is in my opinion the correct way to do so.

Issue #3277, opened 2022-11-08
wasm heap corruption when pointers are in wasm locals


anuraaga:
While understanding how wasm locals get marked during GC, I found this case where they don't seem to get marked causing heap corruption of valid objects.

Using latest `dev` tinygo

https://github.com/anuraaga/tinygo-gctest
```
git clone https://github.com/anuraaga/tinygo-gctest
cd tinygo-gctest
tinygo build -scheduler=none -target=wasi -o gctest.wasm .
go run github.com/tetratelabs/wazero/cmd/wazero@latest run gctest.wasm
```

Should print yogi and garfield, but prints garfield twice because yogi got overwritten.

Definitely not an expert of the code, but looking at [gc.go](https://github.com/tinygo-org/tinygo/blob/release/transform/gc.go) my reading is it creates stack slots for functions with a call to `runtime.alloc` inside them, but doesn't seem to ever go up the call tree from these. In this example, the pointer is returned to the calling function, which itself has no direct call to `runtime.alloc`, but keeps the pointer in one of its wasm slots and would need a GC stack slot allocated for it. Not sure how to update the code to do this if it's missing, but hopefully it's not too hard 😂

anuraaga:
BTW updated my repro command to compile without `-opt=2`, verified it shows the same behavior. Also, while I've kept the `//go:inline` to make `wasm2wat` easier to read, I confirmed the same behavior without them

anuraaga:
I found that reverting #3230 (actually just gc_stack_portable - https://github.com/tinygo-org/tinygo/compare/dev...anuraaga:revert-3230?expand=1) seems to solve the issue for this code. Not a root cause but hopefully can be a good hint.

dgryski:
@anuraaga So do you think the issue was related to the alloca scanning patch or the llvm15 upgrade?

anuraaga:
@dgryski I think the allocas scanning patch but the llvm upgrade is also a significant one that could have had an effect, not sure.

dgryski:
I rebuilt tinygo with my homebrew llvm (15.0.4) and this test case stopped failing for me.  I agree though I'd be happier knowing the actual root cause.

(Rebuilt with llvm from tinygo repo; again no failure detected.)

Edit: n/m, bug present only when `-scheduler=none` provided.

dgryski:
Here's the diff of the body of the for loop, from the `scheduler=none` to the regular wasm scheduler:
```diff
[...]

PR #3291, opened 2022-11-17
wasm: fix scanning of the stack

aykevl:
LLVM wasn't aware that runtime.stackChainStart must be kept live and can't be optimized away. With this hack, it is forced to consider stackChainStart live at the time of the stack scan.

This fixes the corruption in https://github.com/tinygo-org/tinygo/issues/3277

EDIT: I should point out that @dgryski did investigate the bug so much of the credit goes to him.

### DIFF
--- a/src/runtime/gc_stack_portable.go
+++ b/src/runtime/gc_stack_portable.go
@@ -6,6 +6,7 @@ package runtime
 
 import (
 	"internal/task"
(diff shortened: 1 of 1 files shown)


dgryski (approved):
LGTM.

deadprogram:
Well then thanks @dgryski !!

PR #3292, opened 2022-11-17
Revert "Enable wasm pointer tracking for gc=none."

aykevl:
This reverts commit 0b3a7280fa7fd9cf479a6ebcf608c40280d86b9f and updates the documentation a little bit to explain the purpose of -gc=none. (I'm thinking about the attiny10 by the way where defaulting to -gc=none makes sense).

If we're going to do something like this, IMHO it should be in a separate `-gc=custom`.


deadprogram:
Sorry @anuraaga I was a little too quick on the click. I think good point about `-gc=custom`

PR #3294, opened 2022-11-17
avr: support ThinLTO

aykevl:
ThinLTO results in a small code size reduction, which is nice (especially on these very small chips). It also brings us one step closer to using ThinLTO everywhere.

See: #2870

### DIFF
--- a/targets/avr.ld
+++ b/targets/avr.ld
@@ -5,7 +5,7 @@ MEMORY
     RAM (xrw)       : ORIGIN = 0x800000 + __ram_start, LENGTH = __ram_size
 }
 
-ENTRY(__vector_RESET)
+ENTRY(main)
 
 SECTIONS
 {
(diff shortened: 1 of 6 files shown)


targets/avr.ld
```diff
@@ -5,7 +5,7 @@ MEMORY
     RAM (xrw)       : ORIGIN = 0x800000 + __ram_start, LENGTH = __ram_size
 }
 
-ENTRY(__vector_RESET)
+ENTRY(main)
```
aykevl, line 8: This change is necessary because of a quirk in LTO. This way, `main` isn't optimized away (which we certainly don't want!).
In any case, the entry symbol isn't really relevant here. It's just a way to make absolutely sure the given symbol isn't removed by the linker.

deadprogram: Thanks for elaborating on this.

deadprogram (approved):

PR #3295, opened 2022-11-17
transform: remove duplicate if in gc transform

dgryski:


aykevl (approved):
Heh, yeah this is a silly mistake. Probably mine ;)

deadprogram:
Thanks for the correction @dgryski now merging.

PR #3297, opened 2022-11-19
machine/stm32: fix when reading multiple ADC pins

deadprogram:
This PR changes the STM32 ADC pin configuration to always set the pullup resistor on that pin to floating, otherwise I do not think it can actually read a correct value on that pin, if already set to pullup or pulldown.

UPDATE: it also includes a fix when reading multiple analog pins. The existing code was not correctly clearing the register used to determine the rank of which pins to read, which would result in multiple readings overwriting the value of the desired single pin.


kenbell:
LGTM

PR #3298, opened 2022-11-20
Fix XIAO RP2040 pin variable assignment

goropikari:
According to XIAO RP2040 wiki, D9 and D10 correspond to GPIO4 and GPIO3, respectively.


sago35 (approved):
LGTM
Thank you!

https://github.com/earlephilhower/arduino-pico/blob/master/variants/seeed_xiao_rp2040/pins_arduino.h

deadprogram:
Thank you @goropikari for the correction and to @sago35 for review. Now merging.

Issue #1904, opened 2021-05-23
I2C Peripheral Mode not yet supported


pojntfx:
Hi! I'm trying to use an Arduino Uno as an I2C peripheral instead of an I2C controller, but it doesn't seem to be possible to do so using the relevant machine package. To work around this, I though I might use either `device/avr` or CGo, however neither of them seem to work. For the following snippets, I've created minimal, reproducible examples here: [growlapse subdir](https://github.com/pojntfx/growlapse/tree/main/i2c-testing).

## C Sanity Check (`avr-gcc`/`clang`)

So, the first example, which works - written in plain C:

```c
#include <avr/interrupt.h>
#include <util/twi.h>

#define I2C_ADDRESS 0x10

int main() {
  asm volatile("cli" ::: "memory");

  TWAR = I2C_ADDRESS << 1;
  TWCR |= 1 << TWIE;
  TWCR |= 1 << TWEA;
  TWCR |= 1 << TWINT;
  TWCR |= 1 << TWEN;

  asm volatile("sei" ::: "memory");

  while (1) {
  }
}
```

When I run `i2cdetect -y 1`, it shows up at the relevant I2C address:

```shell
pi@raspberrypi:~ $ i2cdetect -y 1
     0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
00:          -- -- -- -- -- -- -- -- -- -- -- -- -- 
10: 10 -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
20: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
30: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
40: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
50: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
60: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
70: -- -- -- -- -- -- -- --
```

So far so good; I've tested this example program using both `avr-gcc` and `clang`, both work; to replicate, clone the growlapse repo and use the relevant [Makefile](https://github.com/pojntfx/growlapse/blob/main/i2c-testing/c/Makefile) with `make install` (or `make install-clang` for clang).

## Idea 1: Using CGo

[...]

PR #3301, opened 2022-11-20
nrf & rp2040: implementation of i2c target mode

kenbell:
Implements i2c target mode on RP2040

Includes example where a single RP2040 can implement both controller and target on it's two I2C controllers (requires them to be physically wired together).

The API is intentionally low-level since not all devices implement a 'memory map' over I2C:
* Some devices have 16 bit registers (not 8 bit)
* Some devices are triggered by accesses to certain addresses

I'd envisage libraries outside Tinygo could provide simpler abstractions (e.g. an interrupt-safe 8-bit / 16-bit registers, etc)

I've updated terminology to controller/target.  The RP2040 spec continues to use the terms 'master/slave', but I thought it preferable to align on the 'controller/target' terminology in TinyGo as this is the first i2c target implementation.  controller/target is used in the latest 'official' spec for I2C: UM10204 Rev. 7.0 — 1 October 2021, which includes this note: _Updated the terms "master/slave" to "controller/target" throughout to align with MIPI I3C specification and NXP's Inclusive Language Project_


deadprogram:
Very cool!

Also see conversation in https://github.com/tinygo-org/tinygo/issues/1904 for some of what we were previously discussing.

src/machine/i2c.go
```diff
@@ -26,6 +26,32 @@ var (
...
+//
+// Handlers are called in an Interrupt context, so should do minimum
+// processing and not perform any allocations.
+//
+// r contains a byte sent by the controller (when evt is I2CReceive),
+// else 0.
+//
+// The return value should be one or more bytes to return to the
+// controller (when evt is I2CRequest), else it is ignored.
+type I2CHandler func(evt I2CTargetEvent, r byte) []byte
```
soypat: Appreciate all the work that went into this, awesome stuff!

Just have one question regarding the I2CHandler- it receives one byte, which I presume is the last read. So is this called one time for each byte received? Maybe worth documenting that each byte in the returned byte slice is added to the buffer (which is max 16 length)?

kenbell: Yep - it's one byte at a time for read.  We could aggregate into an array, but then it needs a pre-allocated buffer, so some decision on the max number of permitted bytes from the controller.

To make it device neutral, I could define a constant (`MaxI2CHandlerResponseBytes`?) that is set to the size of the underlying hardware buffer.

soypat: Yeah I think I'd avoid not writing the bytes over wire immediately (based on intuition alone though). The constant sounds like a good idea to me :+1:

src/machine/machine_rp2040_i2c.go
```diff
@@ -46,18 +50,27 @@ type I2CConfig struct {
...
+
+	// TargetMode configures I2C as a target (receiving commands).
+	// By default, I2C operates as a controller.
+	TargetMode bool
 }
 
 type I2C struct {
 	Bus           *rp.I2C0_Type
 	restartOnNext bool
+	targetMode    bool
```
deadprogram: Perhaps this would be better as:

```go
type I2CMode int
const (
   I2CModeController I2CMode = iota
   I2CModeTarget
)
```

and then this can be defined as:  

```go
type I2C struct {
...
   mode I2CMode
...
}
```

kenbell:
Thanks for the comments.  I'm working on an nrf52 implementation and thinking about re-doing the API on the basis of that.

Something like I2C.WaitForEvent and I2C.Reply, so app logic would look more like this...

```go

var (
	// Max read from Controller is 3 bytes (addr + 16-bit register content)
	readBuf := make([]byte, 3)
	// Max write to Controller is 2 bytes (16-bit register content)
	writeBuf := make([]byte, 2)

	// 256 16-bit registers simulated
	regs := make([]uint16, 256)
	ptr := 0
)



i2c.Listen(addr)

for {
	evt, count, err := i2c.WaitForEvent(readBuf, timeout)
	if err != nil {
		// handle error
	}
	if evt == machine.I2CTargetWrite {
		ptr := buf[0]
		if count > 2 {
			reg[ptr] = uint16(readBuf[1] << 8) | uint16(readBuf[2])
			ptr++
		}
	} else if evt == machine.I2CTargetRead {
		writeBuf[0] = byte(reg[ptr] >> 8)
		writeBuf[1] = byte(reg[ptr] & 0xff)
		ptr++
		i2c.Reply(writeBuf)
	} else {
		panic("invalid i2c event")
	}
}
```

One problem I see is that although `WaitForEvent` should be blocking, it should not block other Go routines.  Would need some kind of backdoor (linker hack) to allow `machine` package to call `runtime.Gosched`.

Alternative would be to have `I2C.IsEventWaiting` (non-blocking) and let the app logic handle the tight loop + yield.

I think this pattern is better since it avoids app logic from having to honor interrupt handler restrictions and allows implementations in the `machine` package that don't use interrupts.

src/machine/machine_rp2040_i2c.go
kenbell: Yep - I prefer this for selecting the mode

aykevl:
I appreciate the thought you put in API design to get it right :)

> I'm working on an nrf52 implementation and thinking about re-doing the API on the basis of that.

Thanks a lot! Having more than one implementation helps to avoid tying the API to a particular implementation.
Can you please also make a PR to update the documentation in https://tinygo.org/docs/reference/machine/#i2c? I've often found that writing this documentation surfaces inconsistencies in a proposed API.

> One problem I see is that although `WaitForEvent` should be blocking, it should not block other Go routines. Would need some kind of backdoor (linker hack) to allow `machine` package to call `runtime.Gosched`.

There are many peripherals that run into this. Essentially, you want to a semaphore: block inside the goroutine, and unblock from within the interrupt. We have some support for that through channels but it requires some chip specific support (which isn't implemented for the RP2040 yet) and channels are a somewhat heavy data structure for this when all we want is a simple semaphore.

sago35:
I started working on samd51.

kenbell:
I've updated the implementation to:

1. Provide the 'WaitForEvent' and 'Reply' API model, which is easier to consume
2. Implement for NRF chipset

Collatoral impact of this:

1. Expose `runtime.Gosched()` as `gosched()` in machine package
2. No longer use interrupts, so waiting for i2c events is now a busy-loop (with yielding via `gosched()`)
3. Completely redo i2c for newer NRF chipsets to use the new peripheral type

Design notes:

This API design precludes the target from providing exactly the right number of bytes to the controller - although RP2040 supports ability to send an unlimited number of bytes (one-at-a-time) in response to a controller query, the NRF peripheral does not support this.  For NRF the target must supply a single buffer with the complete reply present.

This means that where there is no defined maximum read size from the controller, the target implementation must attempt to send the largest possible response in a single call to 'Reply'.  The example demonstrates this.

src/examples/i2c-target/main.go
```diff
@@ -0,0 +1,130 @@
...
+	mem [256]byte
+)
+
+// targetMain implements the 'main loop' for an I2C target
+func targetMain() {
+	buf := make([]byte, maxTxSize)
+	var ptr uint8
+
+	for {
+		evt, c, err := target.WaitForEvent(buf)
```
soypat: Is `c` the number of bytes read? Is this value not traditionally found as `n` in Go programs?

src/machine/machine_rp2040_i2c.go
```diff
@@ -73,11 +82,27 @@ var (
...
 	// timeout in microseconds.
 	const timeout = 40 * 1000 // 40ms is a reasonable time for a real-time system.
 	return i2c.tx(uint8(addr), w, r, timeout)
 }
 
+// Listen starts listening for I2C requests sent to specified address
+//
+// addr is the address to listen to
+// maxRecv controls the receive buffer size
+func (i2c *I2C) Listen(addr uint16, maxRecv uint32) error {
```
soypat: Is maxRecv redundant with the buffer passed into `WaitForEvent`?

kenbell: It’s needed in the API for nrf chipsets where a fixed size receive buffer is passed to the peripheral for receiving messages.

src/examples/i2c-target/main.go
kenbell: thanks - I've renamed to n

src/machine/machine_rp2040_i2c.go
kenbell: I've taken another look - agree that parameter can go.

kenbell:
I think this one is ready for review (and hopefully merge)

deadprogram:
I'm looking it over this weekend. I think the resulting example that the API looks cool.

The implementation is however a bit hard to review just because there is so much in one commit.

Perhaps one commit that deals with the `gosched()` refactoring, then separate ones for the nrf and rp2040?

kenbell:
@deadprogram yeah - it's pretty unwieldy.  I think I'll break it into:

- Move `gosched()`
- Refactor nrf code to allow for separate i2c implementation on newer chips
- New i2c host implementation
- i2c target implementation for rp2040 & nrf

kenbell:
@deadprogram have broken this up into separate commits - hopefully a bunch easier to review

src/machine/machine_nrf528xx.go
```diff
@@ -0,0 +1,102 @@
...
+var (
+	I2C0 = &I2C{Bus: nrf.TWIM0}
+	I2C1 = &I2C{Bus: nrf.TWIM1}
+)
+
+func (i2c *I2C) enableAsController() {
+	i2c.Bus.ENABLE.Set(nrf.TWIM_ENABLE_ENABLE_Enabled)
+}
+
+func (i2c *I2C) disable() {
```
deadprogram, line 31: Perhaps we should export this method? I think it might make it easier later for low power modes on other processors.

deadprogram:
> have broken this up into separate commits - hopefully a bunch easier to review

@kenbell thank you so much for doing that! it makes a big difference at least for me.

src/machine/machine_nrf528xx.go
deadprogram: Actually, we can just leave it as it is for now in your implementation. If we export something, people will automatically expect it, and we probably need to do some additional work to do what I was thinking.

deadprogram:
@kenbell I think this is ready for merge now, unless anyone else has any further feedback?

deadprogram:
@sago35 I see you were working on a samd51 implementation. Any comments before this PR gets merged?

sago35:
@deadprogram 
I think this PR can be merged at this point, since the samd51 implementation is not yet ready.

aykevl:
Two things:

* Can you create a PR to tinygo-site with this API? (This also helps to review it).
* What happens if the target doesn't respond right away? Because you decided to not use interrupts, there is no guarantee that `WaitForEvent()` won't delay for a long time. It could easily delay for many milliseconds if some other goroutine is busy.
  (Are you using clock stretching? But what about chips that don't support clock stretching)?

The second point is my main concern at the moment.

kenbell:
I've started work on a PR to add some docs: https://github.com/tinygo-org/tinygo-site/pull/340

Yep, on the API design, the implication is that clock stretching is supported.  So far, I've done a quick doc scan and it looks like AVR, STM32, SAMD also support clock stretching in their peripherals, so in practice I think it's got broad support.  Are you aware of any platforms that don't support clock stretching?

Thinking about basic use-cases like a controller issuing a command to synchronously update some flash over I2C probably means it's important for the MCU designers to implement clock stretching?  It would be very limiting to require the app to always complete any processing and response buffering/prep within a single I2C clock cycle.

If in practice the model of allowing the scheduler to run proves problematic, I think we have a few possible 'outs':
1 - Add a configuration flag to I2CConfig `NoYield` (default false), if true always tight-loop and don't call `gosched()` for that peripheral
2 - Add an extra call `WaitForEventNoYield` which is the same as `WaitForEvent` but no call to `gosched()`
3 - Add an extra call `IsEventWaiting` which is non-blocking, so the app has more apparent control

In practice, I don't really like (3) personally because:
* It makes the app more complex (how would be best to schedule periodic sensor readings, servicing I2C requests and performing screen updates for a device doing all three?)
* Isn't idiomatic Go (idiomatic Go is blocking IO)
* It doesn't prevent excessive delays in the code (albeit, it might be easier to track them down)

deadprogram:
> If in practice the model of allowing the scheduler to run proves problematic, I think we have a few possible 'outs':

I think option 1 sounds like the least obtrusive, should it be needed. We can probably leave that for a future PR.

deadprogram:
I think we've done enough on this PR, can continue the work of refinement from `dev` branch.

Now merging, thank you very much to @kenbell for the development and to @soypat @sago35 @aykevl for review!

PR #3303, opened 2022-11-21
Fix panic when size 0 passed to malloc

anuraaga:
Currently it will panic (I think Go may translate `&buf[0]` without a length check)

https://linux.die.net/man/3/realloc

- `malloc`/`calloc` can return `NULL` for empty size
- `realloc` with empty is equivalent to `free`


dgryski (approved):
LGTM.

deadprogram:
Thank you for the fixes @anuraaga and thanks for reviewing @dgryski 

Also, thanks for the helpful link @anuraaga 

Now merging.

PR #3306, opened 2022-11-21
avr: added EEPROM support for ATmega boards

SoleimanyBen:
This adds general EEPROM support for AVR devices through the `EEPROM` struct.

First MR here so I was hoping to get feedback on what I could improve and what I did wrong here.

It looks like it goes against convention to not have a `Configure()` function for hardware structures, but I did not see how It could make sense in this instance.

Also, should there be a specific `EEPROM0` constant for each machine file that is then populated within the specific board file?

Thanks again, glad to be able to contribute to such a great project! :)

p.s. im an idiot, sorry about the other PR with the wrong branch selected :sweat_smile:

### DIFF
--- a/src/machine/machine_atmega.go
+++ b/src/machine/machine_atmega.go
@@ -5,10 +5,94 @@ package machine
 
 import (
 	"device/avr"
+	"errors"
 	"runtime/interrupt"
 	"runtime/volatile"
 	"unsafe"
 )
+
+// EEPROM on ATMega
+type EEPROM struct {
+}
+
+var EEPROM0 = EEPROM{}
+
+// setAddress sets the address for a given read or write into the MCUs EEARH/L register.
+func (e EEPROM) setAddress(addr int16) error {
+	if addr < 0 || addr > int16(e.Size()) {
+		return errors.New("address provided is out of bounds")
+	}
+
+	avr.EEARH.Set(uint8(addr >> 8))
+	avr.EEARL.Set(uint8(addr & 0xFF))
+
+	return nil
+}
+
+// WriteAt writes len(data) bytes in the EEPROMs at the provided offset.
+func (e EEPROM) WriteAt(data []byte, off int64) (int, error) {
+	written := 0
+	for i, value := range data {
+		if err := e.WriteByteAt(value, off+int64(i)); err != nil {
+			return written, err
+		}
+		written++
+	}
+	return written, nil
+}
+
+// WriteByteAt performs the logic to writes a byte into the EEPROM at the given address.
+func (e EEPROM) WriteByteAt(value byte, addr int64) error {
+	for avr.EECR.HasBits(avr.EECR_EEPE) {
+	}
+
+	if err := e.setAddress(int16(addr)); err != nil {
+		return err
+	}
+
+	avr.EEDR.Set(value)
+
+	avr.EECR.SetBits(avr.EECR_EEMPE)
+	avr.EECR.SetBits(avr.EECR_EEPE)
+
+	return nil
+}
+
+// ReadAt reads exactly len(buf) into buf at the offset. It will return the amount of bytes copied or an error if one exists.
+// The buffer cannot be empty, and an an error is thrown if fewer bytes are read than the size of the buffer.
+func (e EEPROM) ReadAt(buf []byte, off int64) (int, error) {
+	if len(buf) == 0 {
+		return 0, nil
+	}
+
+	read := 0
+	for i := 0; i < len(buf); i++ {
+		val, err := e.ReadByteAt(off + int64(i))
+		if err != nil {
+			return read, err
+		}
+
+		buf[i] = val
+
+		read++
+	}
+
+	return len(buf), nil
+}
+
+// ReadByteAt reads and returns the byte at the specified address. An error is returned if there is a failure to read.
+func (e EEPROM) ReadByteAt(addr int64) (byte, error) {
+	for avr.EECR.HasBits(avr.EECR_EEPE) {
+	}
+
+	if err := e.setAddress(int16(addr)); err != nil {
+		return byte(0), err
+	}
+
+	avr.EECR.SetBits(avr.EECR_EERE)
+
+	return avr.EEDR.Get(), nil
+}
 
 // I2C on AVR.
 type I2C struct {
--- a/src/machine/machine_atmega1280.go
+++ b/src/machine/machine_atmega1280.go
@@ -11,6 +11,11 @@ import (
 
 const irq_USART0_RX = avr.IRQ_USART0_RX
 
+// Size returns the size of the EEPROM for this machine.
+func (e EEPROM) Size() int64 {
+	return 4096
+}
+
 const (
 	portA Pin = iota * 8
 	portB
--- a/src/machine/machine_atmega1284p.go
+++ b/src/machine/machine_atmega1284p.go
@@ -10,6 +10,11 @@ import (
 
 const irq_USART0_RX = avr.IRQ_USART0_RX
 
+// Size returns the size of the EEPROM for this machine.
+func (e EEPROM) Size() int64 {
+	return 4096
+}
+
 // Return the current CPU frequency in hertz.
 func CPUFrequency() uint32 {
 	return 20000000
--- a/src/machine/machine_atmega2560.go
+++ b/src/machine/machine_atmega2560.go
@@ -13,6 +13,11 @@ const irq_USART1_RX = avr.IRQ_USART1_RX
 const irq_USART2_RX = avr.IRQ_USART2_RX
 const irq_USART3_RX = avr.IRQ_USART3_RX
 
+// Size returns the size of the EEPROM for this machine.
+func (e EEPROM) Size() int64 {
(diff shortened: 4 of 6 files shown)


SoleimanyBen:
~~also now that im thinking about it, maybe the `eepromSize` constant should also be within board specific files?~~

edit: I have made changes to this commit so this comment does not apply anymore

src/machine/machine_atmega.go
```diff
@@ -5,11 +5,120 @@ package machine
...
+// EECR bitfield values.
+// The constants provided by the AVR package are not correct, so these are the correct fields instead.
+const (
+	eecrEERE  = avr.EECR_EERE - 1
+	eecrEEPE  = avr.EECR_EEPE - 1
+	eecrEEMPE = avr.EECR_EEMPE - 2
+	eecrEERIE = avr.EECR_EERIE - 5
+	eecrEEPM0 = avr.EECR_EEPM0 - 12
+	eecrEEPM1 = avr.EECR_EEPM1 - 27
+)
```
aykevl: The bitfields are correct. You have to use them as bitfields, not as bit indices.

src/machine/machine_atmega.go
```diff
@@ -5,11 +5,120 @@ package machine
...
+
+// EEPROM on ATMega
+type EEPROM struct {
+}
+
+var EEPROM0 = EEPROM{}
+
+// setAddress sets the address for a given read or write into the MCUs EEARH/L register.
+func (e EEPROM) setAddress(addr int64) error {
```
aykevl: This is an unexported function, so it's best to use the most efficient data type available. In this case, `int16` (or even `uint16`).

```suggestion
func (e EEPROM) setAddress(addr int16) error {
```

src/machine/machine_atmega.go
```diff
@@ -5,11 +5,120 @@ package machine
...
+	}
+	return written, nil
+}
+
+// WriteByteAt performs the logic to writes a byte into the EEPROM at the given address.
+func (e EEPROM) WriteByteAt(value byte, addr int64) error {
+	notReady := true
+	for notReady {
+		// Wait for the completion of the previous write (if there is one)
+		if !avr.EECR.HasBits(1 << eecrEEPE) {
```
aykevl:
```suggestion
		if !avr.EECR.HasBits(avr.EECR_EEPE) {
```

But actually this whole loop can be simplified to the following:

```go
for avr.EECR.HasBits(avr.EECR_EEPE) {}
```

src/machine/machine_atmega.go
```diff
@@ -5,11 +5,120 @@ package machine
...
+
+	return nil
+}
+
+// ReadAt reads exactly len(buf) into buf at the offset. It will return the amount of bytes copied or an error if one exists.
+// The buffer cannot be empty, and an an error is thrown if fewer bytes are read than the size of the buffer.
+func (e EEPROM) ReadAt(buf []byte, off int64) (int, error) {
+	if len(buf) == 0 {
+		return 0, errors.New("buffer cannot be empty")
+	}
```
aykevl, line 66: Why is this an error?
```suggestion
	if len(buf) == 0 {
		return 0, nil
	}
```

src/machine/machine_atmega.go
```diff
@@ -5,11 +5,120 @@ package machine
...
+
+	size := len(buf)
+	read := 0
+	for i := 0; i < size; i++ {
```
aykevl:
```suggestion
	for i := 0; i < len(buf); i++ {
```

src/machine/machine_atmega.go
```diff
@@ -5,11 +5,120 @@ package machine
...
+		}
+
+		buf[i] = val
+
+		read++
+	}
+
+	if read < size {
+		return read, errors.New("fewer bytes read than size of buffer")
+	}
```
aykevl: This never happens, so you can remove this.

```suggestion
```

src/machine/machine_atmega.go
```diff
@@ -5,11 +5,120 @@ package machine
...
+
+	return read, nil
```
aykevl: This always equal to `len(buf)`.
```suggestion
	return len(buf), nil
```

src/machine/machine_atmega1280.go
```diff
@@ -11,6 +11,11 @@ import (
 
 const irq_USART0_RX = avr.IRQ_USART0_RX
 
+// Size returns the size of the EEPROM for this machine.
+func (e EEPROM) Size() int64 {
+	return 4096
```
aykevl, line 16: I believe this constant is part of the .atdf files, but that can be fixed in a future patch. (It needs a change to tools/gen-device-avr).

aykevl (commented):
I have a number of nits, but overall the code looks quite reasonable.

Can you also make a PR to https://github.com/tinygo-org/tinygo-site with API documentation on the EEPROM? At least the machine package, but if you could also write something in here (https://tinygo.org/docs/concepts/peripherals/) that would be great.

SoleimanyBen:
> I have a number of nits, but overall the code looks quite reasonable.
> 
> 
> 
> Can you also make a PR to https://github.com/tinygo-org/tinygo-site with API documentation on the EEPROM? At least the machine package, but if you could also write something in here (https://tinygo.org/docs/concepts/peripherals/) that would be great.

Absolutely. I'll make a write up once I fix the other issues you mentioned!

src/machine/machine_atmega.go
SoleimanyBen: If I make this change, within this function there is a comparison between an `int64` and `int16`. Should I be casting the `int16` into an `int64` to perform this check, or should I change the result of the `Size()` function to return `int16` instead?

src/machine/machine_atmega.go
```diff
@@ -5,11 +5,95 @@ package machine
...
+// EEPROM on ATMega
+type EEPROM struct {
+}
+
+var EEPROM0 = EEPROM{}
+
+// setAddress sets the address for a given read or write into the MCUs EEARH/L register.
+func (e EEPROM) setAddress(addr int16) error {
+	if addr < 0 || addr > int16(e.Size()) {
+		return errors.New("address provided is out of bounds")
```
aykevl, line 23: Please define a global constant with this error message to return. For example:

```go
var errEEPROMAddress = errors.New("eeprom: address out of bounds")
```

Otherwise this forces a heap allocation when the error is triggered, which is not good.

src/machine/machine_atmega.go
```diff
@@ -5,11 +5,95 @@ package machine
...
 
+// EEPROM on ATMega
+type EEPROM struct {
+}
+
+var EEPROM0 = EEPROM{}
+
+// setAddress sets the address for a given read or write into the MCUs EEARH/L register.
+func (e EEPROM) setAddress(addr int16) error {
+	if addr < 0 || addr > int16(e.Size()) {
```
aykevl, line 22: I think this is an off-by-one.

```suggestion
	if addr < 0 || addr >= int16(e.Size()) {
```

src/machine/machine_atmega.go
aykevl: In this case you know it will never be bigger than an `int16` so casting to the most efficient type is best (for AVR that is `int16`).

SoleimanyBen:
Sorry for the extreme delay on these changes, will go ahead and add them this week.

deadprogram:
@SoleimanyBen I was looking at this PR today, and had a couple ideas. Seems to me that it would be great to implement the `ReadWriteCloser` interface for EEPROM as the main API developers would use.

Imaginary code:

```go
func main() {
    store := machine.EEPROM.Open(address)
    defer store.Close()

    var data [12]byte
    _, err := store.Read(data)
    if err != nil {
        panic("nope")
    }

    println(string(data))


    store.Write([]byte("Hola, mundo!))
}
```

SoleimanyBen:
@deadprogram That's a great idea! I'll write up those changes when I make the other adjustments later this week. :)

aykevl:
> @SoleimanyBen I was looking at this PR today, and had a couple ideas. Seems to me that it would be great to implement the `ReadWriteCloser` interface for EEPROM as the main API developers would use.

Why though? It seems to me that `ReaderAt` and `WriterAt` are a much closer fit to the underlying hardware. With `ReadWriteClose` you have to:
- Open/close the EEPROM, which is not something the hardware needs
- Keep track of the read/write offset, which again isn't natively supported by the hardware

If an application really wanted to layer a `ReadWriteCloser` on top, you could easily write a new package that wraps the EEPROM to provide these methods. But for the machine API, I think it only adds unnecessary overhead.

deadprogram:
@SoleimanyBen I think the interface to create here is actually the `BlockDevice` one, same as https://github.com/tinygo-org/tinyfs/blob/release/tinyfs.go#L42-L69

This is what I have been doing in my Flash PR #3472

aykevl:
@deadprogram EEPROM is not quite the same as flash though. Unlike flash, you can erase and rewrite individual bytes easily so "page size" (a big part of BlockDevice) doesn't make much sense.

In my opinion, this PR is _almost_ ready, it just needs a few small changes to mergeable.

PR #3308, opened 2022-11-23
xiao-rp2040: add pin definitions

sago35:
This PR updates the definition of xiao-rp2040.

One is the addition of pin names related to NEOPIXEL.
To use WS2812 on xiao-rp2040, NEOPIXEL_POWER must be set to High.
The pin names were taken from board_qtpy.go and arduino definitions.

The other is to correct an error in the SPI pin definition.

arduino definitions
https://github.com/earlephilhower/arduino-pico/blob/a7cf5cd1caf1afc8aebc8723776847227fed8966/variants/seeed_xiao_rp2040/pins_arduino.h

pinout
image

schema
https://wiki.seeedstudio.com/XIAO-RP2040/

### DIFF
--- a/src/machine/board_xiao-rp2040.go
+++ b/src/machine/board_xiao-rp2040.go
@@ -38,7 +38,10 @@ const (
 
 // Onboard LEDs
 const (
-	NEOPIXEL = GPIO12
+	NEOPIXEL       = GPIO12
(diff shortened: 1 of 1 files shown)


deadprogram:
Thank you for the corrections @sago35 now merging.

PR #3309, opened 2022-11-24
avr: add channel test

aykevl:
It is working now, so add it as a test. Not sure why, maybe the ThinLTO change fixed something here?

I've also gotten testdata/math.go to work, but it requires a fix in either picolibc or Clang (and I've sent a patch to both).

Also re-enable the AVR CI tests. I hope they'll work now, but if not we can simply disable them again.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -194,10 +194,6 @@ func runPlatTests(options compileopts.Options, tests []string, t *testing.T) {
 				// Breaks interp.
 				continue
 
-			case "channel.go":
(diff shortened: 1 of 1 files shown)


dgryski (approved):
LGTM

deadprogram:
:eight: bit forever! Now merging.

aykevl:
> Also re-enable the AVR CI tests. I hope they'll work now, but if not we can simply disable them again.

Looks like they work reliably now :tada:

PR #3311, opened 2022-11-25
wasm: support ThinLTO

aykevl:
Using ThinLTO manages to optimize binaries significantly. The exact amount varies a lot by program but it's easily a 25-30% reduction.

Don't remove non-ThinLTO support yet. It would not surprise me if this triggered some unintended side effect (especially considering the very large reduction in code size). Eventually, it should be removed though.

### DIFF
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -190,17 +190,11 @@ func (c *Config) StackSize() uint64 {
 	return c.Target.DefaultStackSize
 }
 
-// UseThinLTO returns whether ThinLTO should be used for the given target. Some
-// targets (such as wasm) are not yet supported.
-// We should try and remove as many exceptions as possible in the future, so
-// that this optimization can be applied in more places.
+// UseThinLTO returns whether ThinLTO should be used for the given target.
 func (c *Config) UseThinLTO() bool {
-	parts := strings.Split(c.Triple(), "-")
-	if parts[0] == "wasm32" {
-		// wasm-ld doesn't seem to support ThinLTO yet.
-		return false
-	}
-	// Other architectures support ThinLTO.
+	// All architectures support ThinLTO now. However, this code is kept for the
+	// time being in case there are regressions. The non-ThinLTO code support
+	// should be removed when it is proven to work reliably.
 	return true
 }
 
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -1099,6 +1099,11 @@ func (b *builder) createFunctionStart(intrinsic bool) {
 		// otherwise the function is not exported.
 		functionAttr := b.ctx.CreateStringAttribute("wasm-export-name", b.info.linkName)
 		b.llvmFn.AddFunctionAttr(functionAttr)
+		// Unlike most targets, exported functions are actually visible in
+		// WebAssembly (even if it's not called from within the WebAssembly
+		// module). But LTO generally optimizes such functions away. Therefore,
+		// exported functions must be explicitly marked as used.
+		llvmutil.AppendToGlobal(b.mod, "llvm.used", b.llvmFn)
 	}
 
 	// Some functions have a pragma controlling the inlining level.
--- a/compiler/testdata/pragma.ll
+++ b/compiler/testdata/pragma.ll
@@ -6,6 +6,7 @@ target triple = "wasm32-unknown-wasi"
 @extern_global = external global [0 x i8], align 1
 @main.alignedGlobal = hidden global [4 x i32] zeroinitializer, align 32
 @main.alignedGlobal16 = hidden global [4 x i32] zeroinitializer, align 16
+@llvm.used = appending global [2 x ptr] [ptr @extern_func, ptr @exportedFunctionInSection]
 @main.globalInSection = hidden global i32 0, section ".special_global_section", align 4
 @undefinedGlobalNotInSection = external global i32, align 4
 @main.multipleGlobalPragmas = hidden global i32 0, section ".global_section", align 1024


dgryski:
Did the test corpus run fine with this change?  (I'll probably run it anyway on my laptop..)

aykevl:
@dgryski was planning to but haven't done that yet.

aykevl:
Updated to fix `make wasmtest`. The code size reductions are now a lot less significant (10-15%, still quite nice though).

dgryski:
Still a CI test failure: `    --- FAIL: TestCompiler/pragma.go (0.17s)`

dgryski:
Corpus test failure:
```
Error: failed to run main module `/var/folders/5b/_hr1d1fd3qn4t9vtfx5p61wm0000gp/T/tinygo2693645423/main`

Caused by:
    1: wasm trap: out of bounds memory access
       wasm backtrace:
           0: 0x41f5a - runtime.memmove
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/runtime.go:37:6
                     - runtime.sliceCopy
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/slice.go:51:9
                     - (math/big.nat).set
                           at /Users/dgryski/go/src/go.googlesource.com/go/src/math/big/nat.go:94:6
           1: 0x4875e - (*math/big.Int).Set
                           at /Users/dgryski/go/src/go.googlesource.com/go/src/math/big/int.go:74:20
           2: 0x70952 - (*golang.org/x/crypto/bn256.gfP2).Set
                           at /Users/dgryski/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/golang/crypto/bn256/gfp2.go:37:9
           3: 0x73cfe - (*golang.org/x/crypto/bn256.gfP6).Mul
                           at /Users/dgryski/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/golang/crypto/bn256/gfp6.go:160:9
           4: 0x733b7 - (*golang.org/x/crypto/bn256.gfP12).Mul
                           at /Users/dgryski/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/golang/crypto/bn256/gfp12.go:109:8
           5: 0x7ef1f - golang.org/x/crypto/bn256.TestGFp12Invert
                           at /Users/dgryski/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/golang/crypto/bn256/bn256_test.go:99:15
           6: 0x18492 - testing.tRunner
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/testing/testing.go:367:4
           7: 0x1817f - (*testing.T).Run
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/testing/testing.go:401:9
                     - testing.runTests$1
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/testing/testing.go:468:9
           8: 0x18492 - testing.tRunner
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/testing/testing.go:367:4
           9: 0x1366f - testing.runTests
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/testing/testing.go:466:9
                     - (*testing.M).Run
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/testing/testing.go:444:29
                     - main.main
                           at /Users/dgryski/Library/Caches/go-build/2d/2dfeac23534ef1a4483b6e9c81dad5cbd9377a03b812dcc9745762ba19ea8bf4-d:73:15
                     - runtime.run$1
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/scheduler_any.go:26:11
          10: 0xfd09 - <goroutine wrapper>
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/scheduler_any.go:24:2
          11:  0x834 - tinygo_launch
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/internal/task/task_asyncify_wasm.S:59
          12: 0xfbdf - (*internal/task.Task).Resume
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/internal/task/task_asyncify.go:110:17
                     - runtime.scheduler
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/scheduler.go:236:11
                     - runtime.run
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/scheduler_any.go:29:11
                     - _start
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/runtime_wasm_wasi.go:22:5

FAIL	golang.org/x/crypto/bn256	0.366s
```

Edit: adding printlns trying to hunt this down changes the failure.  Looks like memory corruption or otherwise optimization failure..

dgryski:
Adding another corpus failure.  Maybe there's a connection.
```

Caused by:
    1: wasm trap: wasm `unreachable` instruction executed
       wasm backtrace:
           0: 0xac39 - runtime.abort
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/runtime_tinygowasm.go:73:6
                     - runtime.runtimePanic
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/panic.go:59:7
           1: 0xa72c - runtime.slicePanic
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/panic.go:137:14
           2: 0x58952 - gonum.org/v1/gonum/mat.getDenseWorkspace
                           at /Users/dgryski/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/gonum/gonum/mat/pool.go:103:25
           3: 0x57201 - (*gonum.org/v1/gonum/mat.Dense).isolatedWorkspace
                           at /Users/dgryski/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/gonum/gonum/mat/dense.go:167:8
           4: 0x56896 - (*gonum.org/v1/gonum/mat.Dense).Scale
                           at /Users/dgryski/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/gonum/gonum/mat/dense_arithmetic.go:710:36
           5: 0x8db03 - gonum.org/v1/gonum/optimize/convex/lp.testSimplex
                           at /Users/dgryski/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/gonum/gonum/optimize/convex/lp/simplex_test.go:222:13
           6: 0x8ea89 - gonum.org/v1/gonum/optimize/convex/lp.testRandomSimplex
                           at /Users/dgryski/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/gonum/gonum/optimize/convex/lp/simplex_test.go:173:14
           7: 0x8d1a0 - gonum.org/v1/gonum/optimize/convex/lp.TestSimplex
                           at /Users/dgryski/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/gonum/gonum/optimize/convex/lp/simplex_test.go:137:19
           8: 0x1773b - testing.tRunner
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/testing/testing.go:367:4
           9: 0x17428 - (*testing.T).Run
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/testing/testing.go:401:9
                     - testing.runTests$1
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/testing/testing.go:468:9
          10: 0x1773b - testing.tRunner
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/testing/testing.go:367:4
          11: 0x122e6 - testing.runTests
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/testing/testing.go:466:9
                     - (*testing.M).Run
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/testing/testing.go:444:29
                     - main.main
                           at /Users/dgryski/Library/Caches/go-build/cf/cf672c9554632b6bfe9c62166af1c0fc01ffd44e0121072636c0d4e75e276f3f-d:51:15
                     - runtime.run$1
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/scheduler_any.go:26:11
          12: 0xee92 - <goroutine wrapper>
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/scheduler_any.go:24:2
          13:  0x981 - tinygo_launch
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/internal/task/task_asyncify_wasm.S:59
          14: 0xed6e - (*internal/task.Task).Resume
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/internal/task/task_asyncify.go:110:17
                     - runtime.scheduler
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/scheduler.go:236:11
                     - runtime.run
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/scheduler_any.go:29:11
                     - _start
                           at /Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/runtime_wasm_wasi.go:22:5

```

dgryski:
The problem is an interaction between the new `ThinLTO` optimization and the pooling added in https://github.com/tinygo-org/tinygo/pull/3234

Edit: Looks like removing the pooling logic from `Get()` helps, but reverting the patch entirely still causes a failure.  Still investigating...

Edit: definitely memory corruption-related.  Tests pass if `-test.v` is given to the test binary; panics otherwise.

Edit: Nope, weirder than that.  There is some non-determinism going on here.  Can't figure out where the random is coming from though.  Same binary.  Panics twice in two slightly different ways, and succeeds the third time.  Similar behaviour with `wazero`, so it's not just `wasmtime`.

```
~/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/golang/crypto/bn256 $ wasmtime run ./bn256.test -- -test.v
Error: failed to run main module `./bn256.test`

Caused by:
    1: wasm trap: out of bounds memory access
       wasm backtrace:
           0: 0x70937 - <unknown>!(*golang.org/x/crypto/bn256.gfP2).Set
           1: 0x7450a - <unknown>!(*golang.org/x/crypto/bn256.gfP6).Set
           2: 0x7355e - <unknown>!(*golang.org/x/crypto/bn256.gfP12).Mul
           3: 0x7aa2f - <unknown>!golang.org/x/crypto/bn256.Pair
           4: 0x810c3 - <unknown>!golang.org/x/crypto/bn256.TestBilinearity
           5: 0x18492 - <unknown>!testing.tRunner
           6: 0x1817f - <unknown>!testing.runTests$1
           7: 0x18492 - <unknown>!testing.tRunner
           8: 0x1366f - <unknown>!runtime.run$1
           9: 0xfd09 - <unknown>!runtime.run$1$gowrapper
          10:  0x834 - <unknown>!tinygo_launch
          11: 0xfbdf - <unknown>!_start

~/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/golang/crypto/bn256 $ wasmtime run ./bn256.test -- -test.v
Error: failed to run main module `./bn256.test`

Caused by:
    1: wasm trap: out of bounds memory access
       wasm backtrace:
           0: 0x70937 - <unknown>!(*golang.org/x/crypto/bn256.gfP2).Set
           1: 0x7450a - <unknown>!(*golang.org/x/crypto/bn256.gfP6).Set
           2: 0x7355e - <unknown>!(*golang.org/x/crypto/bn256.gfP12).Mul
           3: 0x7751b - <unknown>!(*golang.org/x/crypto/bn256.gfP12).Exp
           4: 0x77265 - <unknown>!(*golang.org/x/crypto/bn256.GT).ScalarMult
           5: 0x8118f - <unknown>!golang.org/x/crypto/bn256.TestBilinearity
           6: 0x18492 - <unknown>!testing.tRunner
           7: 0x1817f - <unknown>!testing.runTests$1
           8: 0x18492 - <unknown>!testing.tRunner
           9: 0x1366f - <unknown>!runtime.run$1
          10: 0xfd09 - <unknown>!runtime.run$1$gowrapper
          11:  0x834 - <unknown>!tinygo_launch
          12: 0xfbdf - <unknown>!_start

~/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/golang/crypto/bn256 $ wasmtime run ./bn256.test -- -test.v
=== RUN   TestGFp2Invert
--- PASS: TestGFp2Invert (0.00s)
=== RUN   TestGFp6Invert
--- PASS: TestGFp6Invert (0.00s)
=== RUN   TestGFp12Invert
--- PASS: TestGFp12Invert (0.00s)
=== RUN   TestCurveImpl
--- PASS: TestCurveImpl (0.00s)
=== RUN   TestOrderG1
--- PASS: TestOrderG1 (0.01s)
=== RUN   TestOrderG2
--- PASS: TestOrderG2 (0.02s)
=== RUN   TestOrderGT
--- PASS: TestOrderGT (0.11s)
=== RUN   TestBilinearity
--- PASS: TestBilinearity (0.53s)
=== RUN   TestG1Marshal
--- PASS: TestG1Marshal (0.01s)
=== RUN   TestG2Marshal
--- PASS: TestG2Marshal (0.02s)
=== RUN   TestG1Identity
--- PASS: TestG1Identity (0.00s)
=== RUN   TestG2Identity
--- PASS: TestG2Identity (0.00s)
=== RUN   TestTripartiteDiffieHellman
--- PASS: TestTripartiteDiffieHellman (0.45s)
PASS
```

Edit: bah, of course the randomness is coming from the random test values being constructed .
```
bn256_test.go
9:	"crypto/rand"
202:		a, p1, _ := RandomG1(rand.Reader)
203:		b, p2, _ := RandomG2(rand.Reader)
272:	a, _ := rand.Int(rand.Reader, Order)
273:	b, _ := rand.Int(rand.Reader, Order)
274:	c, _ := rand.Int(rand.Reader, Order)
```

Edit: Keeping the pooling logic and building with `-tags="runtime_asserts"` causes the test to panic reliably.

dgryski:
So, I think what's happening here is similar memory corruption to the previous "garbage collector loses track of a pointer and then marks it as free, and returns it for a subsequent allocation" issue we've seen before.

For this particular one, it looks like with the new `thinlto` pass, stronger inlining means that the optimizer ends up deciding a stack chain `push`/`pop` pair for a pointer have "no effect" and are removed.  Obviously this is not the case.

Here's a patch, but I'm not a fan of it
```
diff --git a/transform/gc.go b/transform/gc.go
index a9b5ba74..16ab271f 100644
--- a/transform/gc.go
+++ b/transform/gc.go
@@ -248,6 +248,7 @@ func MakeGCStackSlots(mod llvm.Module) bool {

                        // Store the pointer into the stack slot.
                        store := builder.CreateStore(ptr, gep)
+                       store.SetVolatile(true)
                        pointerStores[store] = struct{}{}
                }

@@ -265,7 +266,8 @@ func MakeGCStackSlots(mod llvm.Module) bool {
                                prev.SetTailCall(false)
                        }
                        builder.SetInsertPointBefore(ret)
-                       builder.CreateStore(parent, stackChainStart)
+                       v := builder.CreateStore(parent, stackChainStart)
+                       v.SetVolatile(true)
                }
        }
```

It marks all the stack chain loads and stores as volatile, so the optimizer is forced to keep them.  Obviously this limits optimization opportunities.  But at least we're not losing pointers.

aykevl:
Wow, that's some heroic bug finding! Thanks a lot!
I will take a look at your proposed fix, I'm sure there is a cleaner fix than the big volatile hammer.

aykevl:
Can you share your small reproducer? I'm not entirely convinced of your push/pop theory as _I think_ LLVM shouldn't do such an optimization (because of the volatile load in `markStack`). So I'd like to take a closer look.
Making these stores volatile might have a significant performance impact and there is a good chance the fix only fixes the bug as a side effect of limiting some legit optimizations.

dgryski:
Here's my current failing test:

```
package main

import (
        "math/big"
        "testing"
)

type G2 struct {
        p *twistPoint
}

func (e *G2) ScalarBaseMult(k *big.Int, pool *bnPool) *G2 {
        if e.p == nil {
                e.p = newTwistPoint(nil)
        }
        TwistMul(twistGen, k, pool)
        return e
}

type bnPool struct {
}

func (pool *bnPool) Get() *big.Int {
        return new(big.Int)
}

func TestOrderG2(t *testing.T) {
        g := new(G2).ScalarBaseMult(p, nil)
        println(g.p)
}

type gfP2 struct {
        x, y *big.Int // value is xi+y.
}

func newGFp2(pool *bnPool) *gfP2 {
        return &gfP2{pool.Get(), pool.Get()}
}

func (e *gfP2) Set(a *gfP2) *gfP2 {
        e.x.Set(a.x)
        e.y.Set(a.y)
        return e
}

func (e *gfP2) SetZero() *gfP2 {
        e.x.SetInt64(0)
        e.y.SetInt64(0)
        return e
}

func (e *gfP2) Add(a, b *gfP2) *gfP2 {
        e.x.Add(a.x, b.x)
        e.y.Add(a.y, b.y)
        return e
}

//go:noinline
func (e *gfP2) Square(a *gfP2, pool *bnPool) *gfP2 {
        //println("square:", a.y, a.x)
        t1 := pool.Get().Sub(a.y, a.x)
        e.x.Mod(t1, p)
        e.y.Set(t1)
        return e
}

func bigFromBase10(s string) *big.Int {
        n, _ := new(big.Int).SetString(s, 10)
        return n
}

var p = bigFromBase10("65000549695646603732796438742359905742825358107623003571877145026864184071783")

var Order = bigFromBase10("50549695646603732796438742359905742570406053903786389881062969044166799969")

type twistPoint struct {
        x, y, z, t *gfP2
}

var twistGen = &twistPoint{
        &gfP2{
                bigFromBase10("21167961636542580255011770066570541300993051739349375019639421053990175267184"),
                bigFromBase10("64746500191241794695844075326670126197795977525365406531717464316923369116492"),
        },
        &gfP2{
                bigFromBase10("20666913350058776956210519119118544732556678129809273996262322366050359951122"),
                bigFromBase10("17778617556404439934652658462602675281523610326338642107814333856843981424549"),
        },
        &gfP2{
                bigFromBase10("0"),
                bigFromBase10("1"),
        },
        &gfP2{
                bigFromBase10("0"),
                bigFromBase10("1"),
        },
}

func newTwistPoint(pool *bnPool) *twistPoint {
        return &twistPoint{
                newGFp2(pool),
                newGFp2(pool),
                newGFp2(pool),
                newGFp2(pool),
        }
}

func (c *twistPoint) Set(a *twistPoint) {
        c.x.Set(a.x)
        c.y.Set(a.y)
        c.z.Set(a.z)
        c.t.Set(a.t)
}

//go:noinline
func TwistDouble(a *twistPoint, pool *bnPool) {
        println("a.x", a.x)
        A := newGFp2(pool).Square(a.x, pool)
        println("imm A", A)
        println("a.y", a.y)
        B := newGFp2(pool).Square(a.y, pool)
        println("B", B.x, B.y)
        newGFp2(pool).Square(B, pool)
        e := newGFp2(pool).Add(B, B)
        println("A", A, A.x, A.y)
        newGFp2(pool).Square(A, pool)
        _ = e
}

const iterations = 256

func TwistMul(a *twistPoint, scalar *big.Int, pool *bnPool) *twistPoint {
        sum := newTwistPoint(pool)
        sum.x.SetZero()

        t := newTwistPoint(pool)

        for i := iterations; i >= 0; i-- {
                TwistDouble(sum, pool)
                if scalar.Bit(i) != 0 {
                        sum.Set(a)
                }
        }

        _ = t
        return sum
}

func main() {
        var t testing.T
        TestOrderG2(&t)
}
```

This needs to be built with:

```
tinygo run -scheduler=none -target=wasi -tags="runtime_asserts" main.go
```

And then it fails with:
```
Error: failed to run main module `/var/folders/5b/_hr1d1fd3qn4t9vtfx5p61wm0000gp/T/tinygo242161356/main`

Caused by:
    1: wasm trap: out of bounds memory access
       wasm backtrace:
           0: 0x1c40f - <unknown>!(*math/big.Int).Sub
           1: 0x1d3ef - <unknown>!(*main.gfP2).Square
           2: 0x7911 - <unknown>!main.TwistDouble
           3: 0x67ca - <unknown>!_start

error: failed to run compiled binary /var/folders/5b/_hr1d1fd3qn4t9vtfx5p61wm0000gp/T/tinygo242161356/main: exit status 134
```

dgryski:
Note though that because it's memory corruption, it's *very* sensitive to small changes in the allocations that happen.  The test corpus passes with the patch applies, whereas it was failing before.

dgryski:
So, looked at this again.  I agree the `SetVolatile()` calls are adding way too many instructions into the function body.  However, I'm still also pretty convinced that llvm is removing the stack chain updates.  (I was looking particularly at the function output of `TwistDouble` from wasm-decompiler https://github.com/WebAssembly/wabt ).

dgryski:
I'll see if there's is better llvm magic for us to use to stop those writes from vanishing.

dgryski:
Here's what I mean.  This is a snippet of the wasm code from `TwistDouble` compiled with the `ThinLTO` patch.

```
    call $main.newGFp2
    local.get 0
    i32.load
    call $_*main.gfP2_.Square
```

Note that `newGFp2` returns a pointer and it isn't stored in the stackchain before the call to `Square` (which will eventually allocate).  Unless I'm missing something, that pointer is effectively lost and not visible to the collector.  Hence the memory corruption.

dgryski:
The problem is here: https://github.com/tinygo-org/tinygo/blob/release/compiler/symbol.go#L159

The call to `runtime.trackPointer` should not be marked as `nocapture`, because the garbage collector absolutely needs to capture the pointer in the shadow stack.  I will file a separate PR with that fix.

dgryski:
https://github.com/tinygo-org/tinygo/pull/3338

aykevl:
I've narrowed down the crash to `(*main.bnPool).Get`. With this patch, the bug is hidden:

```diff
@@ -248,6 +250,10 @@ func MakeGCStackSlots(mod llvm.Module) bool {
 
                        // Store the pointer into the stack slot.
                        store := builder.CreateStore(ptr, gep)
+                       if strings.HasPrefix(fn.Name(), "(*main.bnPool") {
+                               store.SetVolatile(true)
+                               println(fn.Name())
+                       }
                        pointerStores[store] = struct{}{}
                }
 
```

Next step: look at the IR to try and figure out why this matters.

aykevl:
Still trying to get at the IR (`--save-temps` doesn't seem to save the optimized bitcode files). Meanwhile, I've found that it's these three instructions that hide the bug:

```
 01de00: 20 00                      | local.get 0
 01de02: 20 02                      | local.get 2
 01de04: 36 02 08                   | i32.store 2 8
```

dgryski:
Yes because the pointer isn't being stored to the stack record and so ends up being invisible to the garbage collector when it runs.

aykevl:
~The actual diff is in `math/big.Int` for some reason, after optimization (perhaps because of inlining?)~ nevermind I misread

I can't find anything wrong with the function. The optimizer might remove the store, but that should be safe as no allocations happen between the one allocation and the return instruction.
(If you know how to get at the ThinLTO temporaries, please let me know!)

```diff
 ; Function Attrs: minsize nounwind optsize
 define internal fastcc noalias nonnull %"math/big.Int"* @"(*main.bnPool).Get"() unnamed_addr #1 !dbg !21909 {
 entry:
   %gc.stackobject = alloca { %runtime.stackChainObject*, i32, i8* }, align 8, !dbg !21912
   store { %runtime.stackChainObject*, i32, i8* } { %runtime.stackChainObject* null, i32 1, i8* null }, { %runtime.stackChainObject*, i32, i8* }* %gc.stackobject, align 4, !dbg !21912
   %0 = load %runtime.stackChainObject*, %runtime.stackChainObject** @runtime.stackChainStart, align 4, !dbg !21912
   %1 = getelementptr { %runtime.stackChainObject*, i32, i8* }, { %runtime.stackChainObject*, i32, i8* }* %gc.stackobject, i32 0, i32 0, !dbg !21912
   store %runtime.stackChainObject* %0, %runtime.stackChainObject** %1, align 4, !dbg !21912
   %2 = bitcast { %runtime.stackChainObject*, i32, i8* }* %gc.stackobject to %runtime.stackChainObject*, !dbg !21912
   store %runtime.stackChainObject* %2, %runtime.stackChainObject** @runtime.stackChainStart, align 4, !dbg !21912
   call void @llvm.dbg.value(metadata %reflect.badSyntax* undef, metadata !21911, metadata !DIExpression()), !dbg !21912
   %new = tail call fastcc i8* @runtime.alloc(i32 16), !dbg !21913
   %3 = getelementptr { %runtime.stackChainObject*, i32, i8* }, { %runtime.stackChainObject*, i32, i8* }* %gc.stackobject, i32 0, i32 2, !dbg !21913
-  store i8* %new, i8** %3, align 4, !dbg !21913
+  store volatile i8* %new, i8** %3, align 4, !dbg !21913
   %4 = bitcast i8* %new to %"math/big.Int"*, !dbg !21913
   store %runtime.stackChainObject* %0, %runtime.stackChainObject** @runtime.stackChainStart, align 4, !dbg !21914
   ret %"math/big.Int"* %4, !dbg !21914
 }
```

aykevl:
> Yes because the pointer isn't being stored to the stack record and so ends up being invisible to the garbage collector when it runs.

So the thing is, from what I can see, optimizing away the store in my diff above is entirely safe. The allocated object is then returned to the caller which in turn stores the value in its own stack slot to track it. So the bug should _not_ be inside `(*main.bnPool).Get`.

aykevl:
I'm working on reducing the reproducer even further. (I do this by calling `runGC()` at the beginning of `runtime.alloc`, to make the bug hopefully more obvious).

dgryski:
Right. My initial claim that the pool was to blame was incorrect. But the underlying cause of a missed pointer *is* correct. I found minimizing the bug fairly tedious since you're working against the choices the optimization passes make as well as the exact sequence of allocations.

aykevl:
Here is a much shorter reproducer:

```go
package main

type Num struct {
        val *int
}

func newNum() *Num {
        return &Num{new(int)}
}

type Pair struct {
        x *Num
        y *Num
}

const iterations = 16384

func main() {
        sum := &Pair{
                newNum(),
                newNum(),
        }
        for i := 0; i < iterations; i++ {
                if sum.x.val == nil {
                        println("nil:", sum.x.val, i)
                        panic("sum.x.val is nil!")
                }
                _ = &Num{new(int)}
        }
}
```

This works with the following patch:
```diff
@@ -267,6 +270,7 @@ func alloc(size uintptr, layout unsafe.Pointer) unsafe.Pointer {
                return unsafe.Pointer(&zeroSizedAlloc)
        }
 
+       runGC()
        gcTotalAlloc += uint64(size)
        gcMallocs++
 
```

To verify: the same code executes just fine using `go run`.

aykevl:
> But the underlying cause of a missed pointer _is_ correct.

Yeah. The question remains: which pointer, and why? (I will look at the IR next for clues).

aykevl:
Slightly further simplified:

```go
package main

type Num struct {
        val *int
}

//go:noinline
func newNum() *Num {
        return &Num{new(int)}
}

const iterations = 16384

var x *int

var num *Num

func main() {
        num = newNum()
        for i := 0; i < iterations; i++ {
                if num.val == nil {
                        println("nil:", num.val, i)
                        panic("num.val is nil!")
                }
                x = new(int)
        }
}
```

aykevl:
Finally, I've found the store that is incorrectly eliminated (namely, the first store in `main.newNum`). Looking at the IR, I can't think of a reason why LLVM thinks this is safe. Needs further debugging, this might even be a LLVM bug.

aykevl:
It's the DSEPass. I kind of expected that already, but now I have proof.

dgryski:
Nice. I'm glad someone with more llvm knowledge than I have was able to find a better root cause.

aykevl:
I think I found the actual bug, and it's not a bug in LLVM but in our code. Namely: the `tail` attribute on the call to `runtime.alloc`.

dgryski:
Nice. I'm glad someone with more llvm knowledge than I have was able to find a better root cause.

dgryski:
Should I close the trackpointer "nocapture" PR then?

aykevl:
Right now testing whether that fix actually works (it appears as such). You can close now, or wait until I have made a PR with the fix.

aykevl:
I've pushed a commit with the fix.
Unfortunately, it doesn't seem to be a complete fix for golang.org/x/crypto/bn256.

aykevl:
The bn256 tests pass when I use `SetVolatile`, so the bug isn't fixed entirely. Guess I need to start digging more...
That said, the `tail call` issue is certainly a bug so that would be good to have regardless.

dgryski:
I think the `trackPointer` issue is also a real issue we need to address somehow.

transform/gc.go
```diff
@@ -186,11 +186,25 @@ func MakeGCStackSlots(mod llvm.Module) bool {
...
 			}
 
-			if ptr := stripPointerCasts(ptr); !ptr.IsAAllocaInst().IsNil() {
+			ptrStripped := stripPointerCasts(ptr)
+			if !ptrStripped.IsAAllocaInst().IsNil() {
 				// Allocas don't need to be tracked because they are allocated
 				// on the C stack which is scanned separately.
 				continue
 			}
+			if !ptrStripped.IsACallInst().IsNil() {
```
dgryski: Is this change still needed with https://github.com/tinygo-org/tinygo/pull/3395 ?

aykevl: No, I'll rebase it and remove it once #3395 is in.

deadprogram: #3395 has been merged, so this can be rebased now.

aykevl:
I've rebased this on the dev branch so it should be good now.

Did another size comparison (before and after the ThinLTO change). The size difference is typically around -15%. My [`sizediff`](https://gist.github.com/aykevl/bc593ce06b22d64d8dcbb335a40a1e40) tool says it's an 18% reduction but that's probably in part because testdata/timers.go is a bit of an outlier with a reduction of 77% (from 103kB to 23kB). Anyway, a very nice improvement and one that will hopefully allow more improvements in the future.

aykevl:
Isn't this the perfect time for wasmtime to disappear?

```
  Installing latest version of Wasmtime ()
    Checking for existing Wasmtime installation
    Fetching archive for Linux, version 
https://github.com/bytecodealliance/wasmtime/releases/download//wasmtime--x86_64-linux.tar.xz 
curl: (22) The requested URL returned error: 404
Error: Could not download Wasmtime version ''. See https://github.com/bytecodealliance/wasmtime/releases for a list of available releases
```

Luckily this doesn't seem to be our fault, we're just running their install script. I suspect it broke because the latest release is a pre-release (instead of a normal release).

dgryski (approved):
LGTM

dgryski:
This can be merged now.

PR #3313, opened 2022-11-25
atsame5x: reduce heap allocation

sago35:
This PR reduces atsame5x-can's heap allocation.
Data() is easy to use but causes heap allocation, so rxMsg.Length() was added.


sago35:
src/examples/can (old vs new)

image


src/examples/caninterrupt (old vs new)

image

deadprogram:
Good change @sago35 thanks! Now merging.

PR #3315, opened 2022-11-25
machine/nrf51: add ADC implementation

deadprogram:
This PR adds the ADC implementation for the `nrf51` for boards like the BBC:Microbit and MicrobitV2

### DIFF
--- a/src/machine/machine_nrf51.go
+++ b/src/machine/machine_nrf51.go
@@ -187,3 +187,85 @@ func (spi SPI) Tx(w, r []byte) error {
 
 	return nil
 }
+
+// InitADC initializes the registers needed for ADC.
+func InitADC() {
+	return // no specific setup on nrf51 machine.
+}
+
+// Configure configures an ADC pin to be able to read analog data.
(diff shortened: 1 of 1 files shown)


deadprogram:
Tested via https://github.com/tinygo-org/tinyhci/blob/4db55685e4607e47f036db95791e249c8ad717bc/microbit/main.go which is part of https://github.com/tinygo-org/tinyhci/pull/10 and worked as expected.

deadprogram:
Here it is passing now! :smile_cat: 

https://github.com/tinygo-org/tinygo/pull/3315/checks?check_run_id=9720569604

breml:
@deadprogram thanks for this PR.

I just successfully tested this code with a micro:bit V1.5 and an Octopus Analog Rotation Brick. Works like a charm.

The range that I observed is 64 to 65344. ~~So I wonder, if we could do better in extrapolating the 10bit value to 16bit.~~

deadprogram:
Thanks for verifying @breml now merging.

PR #3316, opened 2022-11-27
board: Adafruit KB2040

bgould:
product page: https://www.adafruit.com/product/5302
circuitpython board definition: https://github.com/adafruit/circuitpython/blob/main/ports/raspberrypi/boards/adafruit_kb2040/mpconfigboard.mk
arduino pins: https://github.com/Dygmalab/arduino-RP2040/blob/main/variants/adafruit_kb2040/pins_arduino.h

### DIFF
--- a/src/machine/board_kb2040.go
+++ b/src/machine/board_kb2040.go
@@ -1,0 +1,109 @@
+//go:build kb2040
+
+package machine
+
+import (
+	"device/rp"
+	"runtime/interrupt"
+)
+
+// Onboard crystal oscillator frequency, in MHz.
+const xoscFreq = 12 // MHz
+
+// GPIO Pins
+const (
+	D0  = GPIO0
+	D1  = GPIO1
+	D2  = GPIO2
+	D3  = GPIO3
+	D4  = GPIO4
+	D5  = GPIO5
+	D6  = GPIO6
+	D7  = GPIO7
+	D8  = GPIO8
+	D9  = GPIO9
+	D10 = GPIO10
+)
+
+// Analog pins
+const (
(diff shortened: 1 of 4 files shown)


src/machine/board_kb2040.go
```diff
@@ -0,0 +1,111 @@
...
+const (
+	A0 = GPIO26
+	A1 = GPIO27
+	A2 = GPIO28
+	A3 = GPIO29
+)
+
+const notConnected = 31
+
+const LED = notConnected
```
aykevl: Please don't include constants for pins that don't exist. If the board has no LED, it should not define `machine.LED`.

aykevl (commented):
Looks reasonable, with one minor nit.

src/machine/board_kb2040.go
bgould: Ok, that makes sense. For the LED it is no problem I think ... I also had the SPI1 pins set up that way because the RP2040 `machine` implementation expected it to be there and I was mostly transcribing this information from the Arduino core for this board

bgould:
Removed references to unused pins and fixed pin definitions for SPI1 ... this should be ready now so removing WIP

src/machine/board_kb2040.go
```diff
@@ -0,0 +1,110 @@
+//go:build kb2040
+// +build kb2040
```
deadprogram: Can you please remove the older tag format here?

deadprogram:
Thank you for another awesome contribution @bgould now merging.

deadprogram (approved):

PR #3321, opened 2022-11-30
Adding support for waveshare rp2040-zero

jespino:
This PR is heavily based on the xiao-rp2040 implementation. It is my first PR to the project and I don't know what is expected from my side to be tested. I'm happy to do all the work needed to get this in, but, some guidance about the expectations would be great.


ysoldak:
Thank you for your contribution, @jespino !
Here is documentation on how to add a new hardware target https://github.com/tinygo-org/tinygo/wiki/Adding-a-new-board

kenbell:
@jespino you probably want to add it to the smoketests target in Makefile.  Will help prevent accidental breakage in future.

jespino:
That sounds good. I'll do it. Thanks @kenbell

src/machine/board_waveshare-rp2040-zero.go
```diff
@@ -0,0 +1,129 @@
...
+	A0 Pin = D26
+	A1 Pin = D27
+	A2 Pin = D28
+	A3 Pin = D29
+)
+
+// Onboard LEDs
+const (
+	NEOPIXEL = GPIO16
+	LED      = NoPin
```
aykevl: If there is no LED, please leave the `LED` constant out. Having it there but defined as `NoPin` is only going to cause confusion.

aykevl (commented):
Looks reasonable to me. Assuming you've tested it and it works well for you?
Just one nit below.

src/machine/board_waveshare-rp2040-zero.go
jespino: I'll do it tomorrow

jespino: About testing it, i have it working with an USB device and connecting regular pins, but I haven't tested i2c, uart or spi yet. I have that pending

jespino: @aykevl I removed the LED constant there. And I tested it with an I2C device that I have (an MP6050 accelerometer) and it looks like it's working fine. I don't have any UART or SPI based device at hand right now.

deadprogram:
Looking good, @jespino I am going to squash/merge now. Thank you for the contribution!

PR #3323, opened 2022-12-04
fix wrong pinout

conejoninja:


sago35:
Fixed in #3298

https://github.com/tinygo-org/tinygo/blob/5293d3e5f0fca3afb73295e114fe9f151766b17f/src/machine/board_xiao-rp2040.go#L27-L28

conejoninja:
ouch! thanks I missed the fix. Then this should be closed.

PR #3324, opened 2022-12-04
WIP: usb-hid: support sending "raw" keyboard reports

bgould:
The API for USB-HID Keyboard is too restrictive ... the io.Writer pattern does not map cleanly to implementing HID devices, though is a useful convenience for some cases.  Eventually should consider if this API should be reworked a little more extensively in the future... but for now add this method is enough to support more general use cases such as:

https://github.com/bgould/keyboard-firmware/blob/main/hosts/usbhid/usbhid.go#L28-L40

### DIFF
--- a/src/machine/usb/hid/keyboard/keyboard.go
+++ b/src/machine/usb/hid/keyboard/keyboard.go
@@ -186,6 +186,20 @@ func (kb *keyboard) write(p uint16) error {
 	return kb.writeKeycode(c)
 }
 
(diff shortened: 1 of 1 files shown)


sago35:
Perhaps it is better to make `kb.tx(b)` public.

```go
func (kb *keyboard) sendKey(consumer bool, b []byte) bool {
	kb.tx(b)
	return true
}
```

bgould:
it is possible to do this relatively easily without updating the keyboard package, so closing.

PR #3327, opened 2022-12-06
usbhid: change usage-maximum to 0xFF

sago35:
With the previous configurations, keycodes larger than 0x73 cannot be sent.
For example, we cannot send `Left Windows (0xF0E3)`.

note:

In fact, the Windows key can also be sent at `0xE080`. 
`0xE080` can be sent even before the change.
However, it is not possible to press, for example, the `Muhenkan 無変換 (0xF08B)` key that often appears on Japanese keyboards.
http://www2d.biglobe.ne.jp/~msyk/keyboard/layout/usbkeycode.html

https://www.usb.org/sites/default/files/documents/hid1_11.pdf

```
// P.59 - P.60
Usage Page (Generic Desktop),
Usage (Keyboard),
Collection (Application),
    Report Size (1),
    Report Count (8),
    Usage Page (Key Codes),
    Usage Minimum (224),
    Usage Maximum (231),
    Logical Minimum (0),
    Logical Maximum (1),
    Input (Data, Variable, Absolute), ;Modifier byte
    Report Count (1),
    Report Size (8),
    Input (Constant), ;Reserved byte
    Report Count (5),
    Report Size (1),
    Usage Page (LEDs),
    Usage Minimum (1),
    Usage Maximum (5),
    Output (Data, Variable, Absolute), ;LED report
    Report Count (1),
    Report Size (3),
    Output (Constant), ;LED report padding
    Report Count (6),
    Report Size (8),
    Logical Minimum (0),
    Logical Maximum(255), // ******** 
    Usage Page (Key Codes),
    Usage Minimum (0),
    Usage Maximum (255), // ********
    Input (Data, Array),
End Collection
```

before:
Usage Maximum (0x73) : 0x29, 0x73
Logical Maximum (0x73) : 0x25, 0x73

after:
Usage Maximum (0xFF) : 0x29, 0xFF
Logical Maximum (0xFF) : 0x25, 0xFF


sago35:
This is a keyboard I am developing, made entirely in TinyGo (2 x `xiao-ble`).
This PR change is necessary because I want to use a Japanese key arrangement.
However, this change would be useful not only for Japanese layouts.

image

deadprogram:

PR #3332, opened 2022-12-08
builder: always run wasm-opt for wasm32 binaries

dgryski:


deadprogram:
Seems like an excellent idea to me. Thanks @dgryski now merging.

PR #3335, opened 2022-12-12
Better handling of sub-clusters in SVD

ya-mouse:
The nrf5340_application and nrf9160 SVDs have more complex sub-clusters with nested structures.


deadprogram (approved):
Confirmed that the TinyHCI all still pass with this set of changes. So seems good to me!

deadprogram:
Thank you for the improvement @ya-mouse now merging.

PR #3336, opened 2022-12-12
runtime: move KeepAlive/SetFinalizer to common code

aykevl:
We don't support these yet so let's just put them in a central location. Once these functions are supported we can think about how to structure the code again.

Motivated by #3302.

### DIFF
--- a/src/runtime/gc_conservative.go
+++ b/src/runtime/gc_conservative.go
@@ -640,11 +640,3 @@ func dumpHeap() {
 		}
 	}
 }
-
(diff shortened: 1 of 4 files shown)


dgryski (approved):
LGTM

deadprogram:

aykevl:
Oops too late. I wanted to make this a draft again after https://github.com/tinygo-org/tinygo/pull/3302#discussion_r1046531649. But I guess we can leave it as-is until the discussion about `-gc=custom` is resolved (and then revert if needed).

PR #3337, opened 2022-12-13
Add basic atmega32u support

jespino:


src/runtime/runtime_atmega.go
```diff
@@ -1,5 +1,5 @@
-//go:build avr && atmega
-// +build avr,atmega
+//go:build (avr && atmega) || (avr && atmega32u4)
```
ysoldak:
```suggestion
//go:build avr && (atmega || atmega32u4)
```
makes sense?

jespino: I like the other approach more because it is more consistent with the `+build` line where we have to use `+build avr,atmega avr,atmega32u4`. I'm ok changing it, but I think is more readable and less error prone if both use the same boolean algebra structure.

ysoldak: Valid points. No strong opinion from my side. Sad it’s not possible to express this more concise the older way.

ysoldak: Questioned the need to support old build tags in our slack channel.

ysoldak: Filed cleanup PR for old build tags: #3348

ysoldak: The cleanup PR was merged, now you can drop deprecated build tags in your files too.

jespino: Now merged the dev branch without the old build tags and change it to `avr && (atmega || atmega32u4)`

deadprogram:
Thank you for the contribution @jespino now squashed/merged.

PR #3338, opened 2022-12-14
compiler: mark the trackPointer intrinsic as capturing pointers

dgryski:
The compiler uses this to note which pointers need to be written to the current stack chain object.  When trackPointer is marked as 'nocapture', the llvm optimizer can remove stores of pointers to the shadow stack, thus making them invisible to the garbage collector and causing memory corruption when the GC next runs.

### DIFF
--- a/compiler/symbol.go
+++ b/compiler/symbol.go
@@ -147,8 +147,8 @@ func (c *compilerContext) getFunction(fn *ssa.Function) (llvm.Type, llvm.Value) 
 	case "runtime.trackPointer":
 		// This function is necessary for tracking pointers on the stack in a
 		// portable way (see gc_stack_portable.go). Indicate to the optimizer
-		// that the only thing we'll do is read the pointer.
-		llvmFn.AddAttributeAtIndex(1, c.ctx.CreateEnumAttribute(llvm.AttributeKindID("nocapture"), 0))
+		// that the only thing we'll do is read the pointer.  This function is not
+		// marked as 'nocapture' because the gc pass captures the pointer in the shadow stack.
 		llvmFn.AddAttributeAtIndex(1, c.ctx.CreateEnumAttribute(llvm.AttributeKindID("readonly"), 0))
 	case "__mulsi3", "__divmodsi4", "__udivmodsi4":
 		if strings.Split(c.Triple, "-")[0] == "avr" {


dgryski:
Oops, need to update all the `.ll` files.

dgryski:
Corpus failure:
```
FAIL	gonum.org/v1/gonum/optimize/convex/lp	224.066s
```

I wonder if this is because we're now (correctly!) tracking more pointers?

dgryski:
Looks like this prevents coalescing of pointers into the same storage locations, so "live" pointers are kept alive by being stored in the stack chain entry.  This limits the effectiveness of the garbage collector.  Not ready to merge (yet)..

dgryski:
This also prevents heap-to-stack optimizations.

dgryski:
I wonder if we could special-case trackPointer at https://github.com/tinygo-org/tinygo/blob/release/transform/allocs.go#L168

aykevl:
I do not think this is a correct fix: the way I read it `nocapture` is correct in this case:

> This indicates that the callee does not [capture](https://llvm.org/docs/LangRef.html#pointercapture) the pointer. This is not a valid attribute for return values. This attribute applies only to the particular copy of the pointer passed in this argument. A caller could pass two copies of the same pointer with one being annotated nocapture and the other not, and the callee could validly capture through the non annotated parameter.

> Given a function call and a pointer that is passed as an argument or stored in the memory before the call, a pointer is captured by the call if it makes a copy of any part of the pointer that outlives the call. To be precise, a pointer is captured if one or more of the following conditions hold: [...]

Rationale: `runtime.trackPointer` will be lowered to a store to an alloca. While this alloca is temporarily visible outside the function (via `runtime.stackChainStart`), semantically it does not leave the function as `runtime.stackChainStart` is restored again on return.

> I wonder if we could special-case trackPointer at https://github.com/tinygo-org/tinygo/blob/release/transform/allocs.go#L168

Yes, but it would only be a incomplete fix. The nice thing about `nocapture` is that it is recursively inferred by LLVM passes. For example, in the following example, the heap to stack transform can determine that `n` does not escape purely by looking at `foo` because the `n` parameter to `bar` and `baz` will be marked `nocapture`.

```go
func foo() {
    var n int
    bar(&n)
}
func bar(n *int) {
    baz(n)
}
func baz(n *int) {
    *n = 5
}
```

If `n` is passed to `runtime.trackPointer` in `baz`, that would break the optimization.

In the end, `nocapture` (which is perhaps better named "noescape") is exactly what we need for escape analysis.

dgryski:
Closing as not the right fix.

PR #3339, opened 2022-12-14
Bump CMSIS to 5.9.0

ya-mouse:
Legacy CMSIS 4.x missing some headers like core_m33.h, required by nrf5340 and nrf9160.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,7 +3,7 @@
 	url = https://github.com/NordicSemiconductor/nrfx.git
 [submodule "lib/CMSIS"]
 	path = lib/CMSIS
-	url = https://github.com/ARM-software/CMSIS.git
+	url = https://github.com/ARM-software/CMSIS_5.git
 [submodule "lib/avr"]
 	path = lib/avr
 	url = https://github.com/avr-rust/avr-mcu.git
--- a/Makefile
+++ b/Makefile
@@ -748,7 +748,7 @@ wasmtest:
 build/release: tinygo gen-device wasi-libc $(if $(filter 1,$(USE_SYSTEM_BINARYEN)),,binaryen)
 	@mkdir -p build/release/tinygo/bin
 	@mkdir -p build/release/tinygo/lib/clang/include
-	@mkdir -p build/release/tinygo/lib/CMSIS/CMSIS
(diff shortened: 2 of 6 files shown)


deadprogram:
Very good idea, thanks for the update @ya-mouse now merging.

deadprogram:
@ya-mouse so it turns out there is a problem with this version update, and the TinyGo Bluetooth package.

See https://github.com/tinygo-org/bluetooth/actions/runs/3714845327/jobs/6299516750#step:6:15 for the specific error.

Looks to be this issue:
https://github.com/ARM-software/CMSIS_5/issues/617

I tried a few things, but was not able to get around it. I think it would be best to revert this PR for now, and then we can bring it back in once there is a solution for the Bluetooth package.

ya-mouse:
This snippet should fix the problem: https://github.com/EMCLab-Sinica/ARM-CMSIS_5/commit/f37a11415a37df993e990b8e8a1d5f8b05edb91b

deadprogram:
@ya-mouse but what to do about getting that into upstream repo? Or are you proposing some other solution in the interim so we can be using the newer CMSIS?

ya-mouse:
I did two PRs (different approaches) to separate this change into cmsis_clang.h. Let's see:

https://github.com/ARM-software/CMSIS_5/pull/1596
https://github.com/ARM-software/CMSIS_5/pull/1597

ya-mouse:
Got the fixed merged in upstream: https://github.com/ARM-software/CMSIS_5/pull/1597 into develop.
So, we can bump CMSIS either to a develop commit or wait for the next release.

ya-mouse:
@deadprogram 
Got the fix merged in upstream: https://github.com/ARM-software/CMSIS_5/pull/1597 into develop.
So, we can bump CMSIS either to a develop commit or wait for the next release.

aykevl:
The safest option would probably be to bump to CMSIS 5 and add your patch on top. We can also pick a development version, that's a bit more risky but I don't think the risk is very high (we use only small parts of the CMSIS library).

ya-mouse:
So, we stick with one of commits from CMSIS or I do a patch over latest release? Is there a best practice to incorporate patched version to the source code?

ya-mouse:
@aykevl how do we proceed? :)

PR #3345, opened 2022-12-18
compileopts: replace 'retries' flag with more correct 'timeout' flag

deadprogram:
This PR replaces this poorly named `tinygo flash -retries` flag with the better named `tinygo flash -timeout` flag, which also uses the more understandable values related to time, e.g. `tinygo flash -timeout=30s`.


sago35 (approved):
LGTM

deadprogram:
Thank you for the review @sago35 now merging.

PR #3346, opened 2022-12-18
runtime: GC: implement precise heap

aykevl:
This implements the block-based GC as a partially precise GC. This means that for most heap allocations it is known which words contain a pointer and which don't. This should in theory make the GC faster (because it can skip non-pointer object) and have fewer false positives in a GC cycle. It does however use a bit more RAM to store the layout of each object.

Right now this GC seems to be slower than the conservative GC, but should be less likely to run out of memory as a result of false positives.

This PR defaults to the precise GC on regular OSes (Windows/MacOS/Linux) and doesn't change the default elsewhere (baremetal, wasm, etc). This can be changed in the future if needed.


aykevl:
Looks like I need to update a few tests to get this to pass CI.

dgryski:
Exciting! I'll give this a review as soon as I can.

aykevl:
Rebased to fix the merge conflicts and to fix the interp test failure.

src/runtime/gc_blocks.go
```diff
@@ -168,6 +168,9 @@ func (b gcBlock) markFree() {
 	if gcAsserts && b.state() != blockStateFree {
 		runtimePanic("gc: markFree() was not successful")
 	}
+	if gcAsserts {
+		*(*[4]uintptr)(unsafe.Pointer(b.address())) = [4]uintptr{}
```
aykevl: Oops, should have used `wordsPerBlock` instead of hardcoding the number 4.
UPDATE: fixed :)

aykevl:
Did a quick check to see whether other architectures are affected, and they (surprisingy) are: baremetal systems tend to have a decrease of around 16 bytes binary size and wasm tends to have an increase in binary size of 11 bytes. Probably as a result of some changes around inlining heuristics. Nothing to worry about, honestly.

Of course, where the precise GC is enabled there is a much bigger increase of 400+ bytes. This is expected, and should not be a problem at all for these systems. It might be possible to reduce this somewhat in the future so that precise GC can also be used on AVR (where false positives are especially common).

aykevl:
Currently investigating what appears to be a memory corruption on arm.

aykevl:
Should be fixed now. The issue was that the `runtime.align` function did not correctly align objects: they should be aligned to an 8 byte boundary and not to a 4 byte boundary.

The actual bug is quite interesting. For some reason, the ARM backend used an 'or' instruction on the pointer to add 4. This is unintuitive, but is in fact correct _as long as the pointer is sufficiently aligned_. In this case it wasn't, leading to a wrong pointer being calculated.
Not sure why they use 'or' instead of 'add', I suppose it's faster on some chips or uses less energy or something.

src/internal/task/task_stack.go
```diff
@@ -63,20 +66,20 @@ func (t *Task) Resume() {
...
-	s.canaryPtr = &stack[0]
+	s.canaryPtr = (*uintptr)(stack)
 	*s.canaryPtr = stackCanary
 
 	// Get a pointer to the top of the stack, where the initial register values
 	// are stored. They will be popped off the stack on the first stack switch
 	// to the goroutine, and will start running tinygo_startTask (this setup
 	// happens in archInit).
-	r := (*calleeSavedRegs)(unsafe.Pointer(&stack[uintptr(len(stack))-(unsafe.Sizeof(calleeSavedRegs{})/unsafe.Sizeof(uintptr(0)))]))
+	r := (*calleeSavedRegs)(unsafe.Add(stack, stackSize-unsafe.Sizeof(calleeSavedRegs{})))
```
dgryski, line 82: Any other places in the runtime tree we should switch to using `unsafe.Add` ?

src/runtime/arch_arm.go
```diff
@@ -9,9 +9,9 @@ const TargetBits = 32
 
 const deferExtraRegs = 0
 
-// Align on word boundary.
+// Align on the maximum alignment for this platform (double).
 func align(ptr uintptr) uintptr {
-	return (ptr + 3) &^ 3
+	return (ptr + 7) &^ 7
```
dgryski, line 14: Did you check all the other `align` implementations to make sure they're correct?

src/runtime/gc_precise.go
```diff
@@ -0,0 +1,97 @@
...
+			sizeFieldBits = 4
+		case 32:
+			sizeFieldBits = 5
+		case 64:
+			sizeFieldBits = 6
+		default:
+			runtimePanic("unknown pointer size")
+		}
+
+		// Extract fields.
```
dgryski: An example of what's happening here (vis a vis splitting the layout bits) might help here.

src/runtime/gc_precise.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	if !isOnHeap(word) {
+		// Definitely isn't a pointer.
+		return false
+	}
+
+	// Might be a pointer. Now look at the object layout to know for sure.
+	if scanner.bitmapAddr != nil {
+		if (*(*uint8)(unsafe.Add(scanner.bitmapAddr, index/8))>>(index%8))&1 == 0 {
```
dgryski, line 135: Seems like a helper function might make this more readable, especially if we could change the type from `unsafe.Pointer` to `memoryBitmap` or something.

src/runtime/gc_precise.go
```diff
@@ -0,0 +1,97 @@
...
+
+func newGCObjectScanner(block gcBlock) gcObjectScanner {
+	if gcAsserts && block != block.findHead() {
+		runtimePanic("gc: object scanner must start at head")
+	}
+	scanner := gcObjectScanner{}
+	layout := *(*uintptr)(unsafe.Pointer(block.address()))
+	if layout == 0 {
+		// Unknown layout. Assume all words in the object could be pointers.
+		scanner.size = 1
```
dgryski, line 79: These seem like magic values we might want constants for?  I guess the rest of the pointer scanning logic just "works" with these both set to `1`.

src/runtime/gc_precise.go
```diff
@@ -0,0 +1,97 @@
...
+		return false
+	}
+	// If the bitmap is zero, there are definitely no pointers in the object.
+	return scanner.bitmap == 0
+}
+
+func (scanner *gcObjectScanner) nextIsPointer(word, parent, addrOfWord uintptr) bool {
+	index := scanner.index
+	scanner.index++
+	if scanner.index == scanner.size {
```
dgryski, line 124: I know this is so that the `size = 1` stuff above works, but it seems to me we might like a different way handle this so that we can assert if 'nextIsPointer` is in fact called too many times and overflows the bitmap.

compiler/llvm.go
```diff
@@ -166,6 +166,7 @@ func (c *compilerContext) createObjectLayout(t llvm.Type, pos token.Pos) llvm.Va
 	// Create the global initializer.
 	bitmapBytes := make([]byte, int(objectSizeWords+7)/8)
 	bitmap.FillBytes(bitmapBytes)
+	reverseBytes(bitmapBytes) // big-endian to little-endian
```
dgryski, line 171: Now that we're actually using the layout in the allocator, do we have any tests that it's correct?

src/runtime/gc_precise.go
```diff
@@ -0,0 +1,97 @@
...
+		// While the format allows for large objects without pointers, this is
+		// optimized by the compiler so if bitmapAddr is set, we know that there
+		// are at least some pointers in the object.
+		return false
+	}
+	// If the bitmap is zero, there are definitely no pointers in the object.
+	return scanner.bitmap == 0
+}
+
+func (scanner *gcObjectScanner) nextIsPointer(word, parent, addrOfWord uintptr) bool {
```
dgryski, line 121: I wonder if this ends up being cleaner if we split this into `next()` call that advanced the index and `IsPointer` that does the check?

src/runtime/gc_blocks.go
```diff
@@ -0,0 +1,689 @@
...
+		if preciseHeap {
+			// The first word of the object is just the pointer layout value.
+			// Skip it.
+			start += align(unsafe.Sizeof(uintptr(0)))
+		}
+		for addr := start; addr != end; addr += unsafe.Alignof(addr) {
+			// Load the word.
+			word := *(*uintptr)(unsafe.Pointer(addr))
+
+			if !scanner.nextIsPointer(word, root.address(), addr) {
```
dgryski, line 529: I feel like the fact we have the scanner and the word as two separate things here means they might get out of sync.  For example, `nextIsPointer`, but you're passing in `word`, so it should be `*this*IsPointer`, except that you need to increment the internal scanner index .... We should figure out the best API here.

dgryski (changes requested):
Pretty confident this is correct (it passes the corpus, which is always encouraging).  Just a bunch of questions.

src/internal/task/task_stack.go
aykevl: Yes, lots of places. I have a local branch that I should finish and submit...

aykevl: I'll submit it after this branch (and perhaps #2640) have been merged, to avoid conflicts.

src/runtime/arch_arm.go
aykevl: I should have. I've updated the PR and I think all `align` functions are now correct.

compiler/llvm.go
aykevl: There are some tests in compiler/testdata/gc.go, or do you mean something else?

src/runtime/gc_precise.go
aykevl: This is by design to support arrays. For example, if you do `make([]*byte, 10)` you'll create a datalayout of one word with size == 1 (on most architectures). The `size = 1` above merely uses this fact, it basically treats the object like it would `make([]*T, ...)` (that is, everything could be a pointer).

src/runtime/gc_precise.go
aykevl: I have put a big comment explaining all of this at the top that should hopefully explain what happens here.

src/runtime/gc_blocks.go
aykevl: Yeah I was struggling with the API. Not too happy with what I have here but I couldn't figure out something better that is also generic. Suggestions for a better API are welcome :)

aykevl (commented):
I've put a big comment at the top of gc_precise.go that should answer some questions you have. Please take another look.

src/runtime/gc_precise.go
dgryski: So reading the comment about the `size`, I'm wondering if we know that for slices of structs the padding will always work out such that we'll never get "out-of-sync" with the layout bitmap.

dgryski (approved):
One more question, but otherwise LGTM.

src/runtime/gc_precise.go
aykevl: Well, it should. If it doesn't, that means there is a bug somewhere (which wouldn't be very surprising to me, this code hasn't been tested a lot). If you're interested, the relevant functions are `createObjectLayout` and `getPointerBitmap`.

dgryski: SGTM

PR #3347, opened 2022-12-19
rp2040: fix interrupt issue (2)

sago35:
The method I fixed in #3343 was a bit wrong.
If BUFF_STATUS must updated after IN/OUT processing is performed, the data in usbDPSRAM.EPxBufferControl may be corrupted.

Please review again. @deadprogram


deadprogram:
Good catch on that @sago35 thank you! Now merging.

PR #3348, opened 2022-12-19
build: drop deprecated build tags

ysoldak:
This PR evicts deprecated `+build` tags and ensures no new such created in autogenerated files.
Current `go:build` tags introduced in Go 1.17, that is two versions back.

Fixed mostly by running `go fix ./...` but some files need to be adjusted manually after.


ysoldak:
@aykevl @deadprogram I know, it's huge, but just by number of files, not logic.
There are two commits: one fixes generators, another adds sheer bulk of files auto-fixed with `go fix` and with some manual help.

deadprogram:
Thanks @ysoldak for the PR now merging.

Issue #3341, opened 2022-12-16
recursive array initialization build fatal error: stack overflow


d-enk:
```go
package main

type T struct {
	childs  []T
	childs2 [2]*T
}

type (
	T1 []T1
	T2 [2]*T2
)

func main() {
	_ = T{}

        // build error
	_ = T1{} 
	_ = T2{}
}

```

runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc020c78388 stack=[0xc020c78000, 0xc040c78000]
fatal error: stack overflow

tinygo version 0.26.0 linux/amd64 (using go version go1.19.4 and LLVM version 14.0.0)

aykevl:
Confirmed. This is a bug.

aykevl:
Fixed here, for LLVM 15 only: https://github.com/tinygo-org/tinygo/pull/3349
I'm unlikely to work on a fix for LLVM 14 as it will be rather difficult and LLVM 14 will eventually be deprecated anyway.

Thank you for the very short bug report!

PR #3349, opened 2022-12-22
compiler: fix stack overflow when creating recursive pointer types

aykevl:
There were two types that could result in a compiler stack overflow. This is difficult to fix in LLVM 14, so I won't even bother. However, this is trivial to fix with opaque pointers in LLVM 15. Therefore, this fix is for LLVM 15 only.

Fixes: https://github.com/tinygo-org/tinygo/issues/3341

### DIFF
--- a/compiler/channel.go
+++ b/compiler/channel.go
@@ -43,7 +43,7 @@ func (b *builder) createChanSend(instr *ssa.Send) {
 	}
 
 	// Allocate blockedlist buffer.
-	channelBlockedList := b.mod.GetTypeByName("runtime.channelBlockedList")
+	channelBlockedList := b.getLLVMRuntimeType("channelBlockedList")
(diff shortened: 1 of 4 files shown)


dgryski (approved):
LGTM

compiler/channel.go
```diff
@@ -43,7 +43,7 @@ func (b *builder) createChanSend(instr *ssa.Send) {
 	}
 
 	// Allocate blockedlist buffer.
-	channelBlockedList := b.mod.GetTypeByName("runtime.channelBlockedList")
+	channelBlockedList := b.getLLVMRuntimeType("channelBlockedList")
```
aykevl, line 46: Not sure why this is needed, but `getLLVMRuntimeType` is the one that should have been used here.

PR #3350, opened 2022-12-22
wasi: makes wasmtime "run" explicit

codefromthecrypt:
wasmtime by default will assume the subcommand is "run" vs one of its others, but being explicit helps clarify the actual command invoked.

For example, we pass similar looking args to wasmtime in different places due to how filesystem lookup works at the moment.

Specifically, `build/tinygo test -target wasi os` ends up compiling `main` to wasm and invoking it like so. I found the "run" here was helpful as it is surely the marker of where things start.

```bash
dir=/Users/adrian/Library/Caches/tinygo/goroot-663b9f64fa32ef34422ae1c56b5d12237701f7e39c3efcc1087329a71121caff/src/os
args=[wasmtime run --dir=.. --dir=../.. --dir=/var/folders/vd/1cf8zdb1721f4z5rjggy8bp40000gn/T/tinygotmp2965559219 --env=TMPDIR=/var/folders/vd/1cf8zdb1721f4z5rjggy8bp40000gn/T/tinygotmp2965559219 /var/folders/vd/1cf8zdb1721f4z5rjggy8bp40000gn/T/tinygo2707001624/main --dir=.]
ok  	os	0.094s
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -285,9 +285,9 @@ func Test(pkgName string, stdout, stderr io.Writer, options *compileopts.Options
 			// relative directory up to the module root, even if the test never
 			// reads any files.
 			//
-			// Ex. --dir=.. --dir=../.. --dir=../../..
+			// Ex. run --dir=.. --dir=../.. --dir=../../..
 			dirs := dirsToModuleRoot(result.MainDir, result.ModuleRoot)
-			var args []string
+			args := []string{"run"}
 			for _, d := range dirs[1:] {
 				args = append(args, "--dir="+d)
 			}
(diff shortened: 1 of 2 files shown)


dgryski (approved):
LGTM

dgryski:
Side note: I assume this makes integration with `wazero`'s CLI version easier too?

codefromthecrypt:
> Side note: I assume this makes integration with wazero's CLI version easier too?

yep, indeed it does, though args are still different.

deadprogram:
Thank you for the improvement @codefromthecrypt now merging.

PR #3359, opened 2022-12-28
Add PDM support for circuitplay-bluefruit

mlsorensen:
This PR attempts to add some basic support for PDM and implements it for the Nordic 52840 based SoCs and the CircuitPlayground-Bluefruit board.  

I could imagine maybe the configuration and API can be filled out a bit more to do things like configure and fetch sample rate.  I'm completely new to tinygo and have only played with microcontrollers here and there as a hobby, so hopefully I haven't done anything too boneheaded.

With the change in this PR, I was able to complete a tinygo version of the [Adafruit VU meter](https://learn.adafruit.com/adafruit-circuit-playground-bluefruit/playground-sound-meter) with my son on the Bluefruit.  I couldn't easily figure out how to write to the "CPLAYBTBOOT" onboard flash filesystem to try storing a WAV file, but I did go so far as to echo out a 1 second integer buffer (per the PR's example code) and save that as a WAV on my laptop and play it back. So I think in general it works.

As a side note, I noticed that there is a "microphone" PDM driver that can run I2S in PDM mode. I think maybe this is a quirk of a specific board, to be able to run I2S in PDM mode?  I wonder if it would make sense later to refactor this as a PDM device and just make the controller-specific implementation leverage its I2S.  It was a bit confusing to me to see this microphone driver documented as a PDM but then need an I2S in order to use it.

Thanks!

### DIFF
--- a/src/examples/pdm/pdm.go
+++ b/src/examples/pdm/pdm.go
@@ -20,7 +20,7 @@ func main() {
 	for {
 		if machine.BUTTONA.Get() {
 			println("Recording new audio clip into memory")
-			pdm.Read(&audio[0], uint32(len(audio)))
+			pdm.Read(audio)
 			println(fmt.Sprintf("Recorded new audio clip into memory: %v", audio))
 		}
 	}
--- a/src/examples/pdm/pdm.go
+++ b/src/examples/pdm/pdm.go
@@ -1,0 +1,27 @@
+package main
+
+import (
+	"fmt"
+	"machine"
+)
+
+var (
+	audio = make([]int16, 16)
+	pdm   = machine.PDM{}
+)
+
+func main() {
+	machine.BUTTONA.Configure(machine.PinConfig{Mode: machine.PinInputPulldown})
+	err := pdm.Configure(machine.PDMConfig{})
+	if err != nil {
+		panic(fmt.Sprintf("Failed to configure PDM:%v", err))
+	}
+
+	for {
+		if machine.BUTTONA.Get() {
+			println("Recording new audio clip into memory")
+			pdm.Read(&audio[0], uint32(len(audio)))
+			println(fmt.Sprintf("Recorded new audio clip into memory: %v", audio))
+		}
+	}
+}
--- a/src/examples/pdm/pdm.go
+++ b/src/examples/pdm/pdm.go
@@ -12,7 +12,7 @@ var (
 
 func main() {
 	machine.BUTTONA.Configure(machine.PinConfig{Mode: machine.PinInputPulldown})
-	err := pdm.Configure(machine.PDMConfig{})
+	err := pdm.Configure(machine.PDMConfig{CLK: machine.PDM_CLK_PIN, DIN: machine.PDM_DIN_PIN})
 	if err != nil {
 		panic(fmt.Sprintf("Failed to configure PDM:%v", err))
 	}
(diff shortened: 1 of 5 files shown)


mlsorensen:
Based on other recent PRs I'm assuming the smoke-test-windows is flaky and not because of this PR.

src/machine/machine_nrf52840.go
```diff
@@ -33,3 +35,72 @@ var (
...
+
+	pdm.device.SAMPLE.SetPTR(uint32(uintptr(unsafe.Pointer(&pdm.defaultBuffer))))
+	pdm.device.SAMPLE.SetMAXCNT_BUFFSIZE(1)
+	pdm.device.SetTASKS_START(1)
+	return nil
+}
+
+// Read stores a set of samples in the given target buffer. Pointer should
+// represent first element of buffer slice, not the pointer to the slice itself.
+func (pdm *PDM) Read(bufptr *int16, bufsz uint32) (uint32, error) {
```
deadprogram: Perhaps you can change the signature to:

```go
func (pdm *PDM) Read(buf []int16) (uint32, error)
```

And then perform the pointer casting inside the function itself. What do you think?

src/examples/pdm/pdm.go
```diff
@@ -0,0 +1,27 @@
...
+	machine.BUTTONA.Configure(machine.PinConfig{Mode: machine.PinInputPulldown})
+	err := pdm.Configure(machine.PDMConfig{CLK: machine.PDM_CLK_PIN, DIN: machine.PDM_DIN_PIN})
+	if err != nil {
+		panic(fmt.Sprintf("Failed to configure PDM:%v", err))
+	}
+
+	for {
+		if machine.BUTTONA.Get() {
+			println("Recording new audio clip into memory")
+			pdm.Read(&audio[0], uint32(len(audio)))
```
deadprogram: Making the change I suggested below, would allow changing this line to:

```go
pdm.Read(audio)
```

deadprogram:
First of all, thanks for working on this!

Sorry about the delay in giving feedback. Please see my comments above, and let me know what you think.

src/machine/machine_nrf52840.go
mlsorensen: I'm not a golang expert, I just hack :) My impression was that passing the value would result in more memory consumed, and the audio buffers are usually large. If that's not the case I'll try it, I agree it is cleaner.

Now that I think about it you're just passing the slice reference itself and not the backing array, so I guess it does make sense. Stand by...

mlsorensen:
Integrated your comments, thanks!

src/machine/machine_nrf52840.go
```diff
@@ -33,3 +35,72 @@ var (
...
+		pdm.device.MODE.Set(nrf.PDM_MODE_OPERATION_Mono | nrf.PDM_MODE_EDGE_LeftRising)
+	}
+
+	pdm.device.SAMPLE.SetPTR(uint32(uintptr(unsafe.Pointer(&pdm.defaultBuffer))))
+	pdm.device.SAMPLE.SetMAXCNT_BUFFSIZE(1)
+	pdm.device.SetTASKS_START(1)
+	return nil
+}
+
+// Read stores a set of samples in the given target buffer. Pointer should
```
deadprogram: Can you please update the comment here? Thanks!

deadprogram:
Thanks for making the requested changes. Now going to squash/merge.

aykevl:
It seems to me that this PDM peripheral is shared across all nrf52xxx devices. If so, it would be nice to move the implementation to machine_nrf528xx.go so it can be used across all chips.

PR #3360, opened 2023-01-02
Support for Adafruit Feather M0 Express board

trichner:
- added new board `feather-m0-express` for the [Adafruit Feather M0 Express](https://www.adafruit.com/product/3403) board (inspired by `feather-m0`)

I'm still really confused regarding the UART though, not sure whether this is correct. So far I struggle to test it with a CMSIS-DAP debugger & USB-to-TTY-Serial on the other side.

Pinout used:

### DIFF
--- a/src/machine/board_feather-m0-express.go
+++ b/src/machine/board_feather-m0-express.go
@@ -1,0 +1,101 @@
+//go:build sam && atsamd21 && feather_m0_express
+
+package machine
+
+// used to reset into bootloader
+const resetMagicValue = 0xf01669ef
+
+// GPIO Pins
+const (
+	D0  = PA11  // UART0 RX
+	D1  = PA10  // UART0 TX
+	D2  = NoPin // does not seem to exist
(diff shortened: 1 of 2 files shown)


trichner:
Something that I noticed:
The original Feather M0 board has D10/D11 defined as TX/RX, which corresponds to UART1 (SERCOM1). At least on my Feather M0 Express the TX/RX labelled pins are A10/A11 (PA10/PA11) which correspond to UART0. Though UART0 doesn't seem to be exposed by the machine package.

Is that intentional? Did I miss something?

This begs the question how one is supposed to use the labelled TX/RX ports for UART, i.e. SERCOM0.

deadprogram:
@trichner could you please add an entry to the README for this board?

https://github.com/tinygo-org/tinygo/blob/dev/README.md#supported-boardstargets

Thank you!

deadprogram:
Actually don't worry about the README, I need to update it anyhow with some other missing boards.

Now merging, thank you @trichner

PR #3361, opened 2023-01-02
Support for machine/board bundles

ya-mouse:


deadprogram:
Thanks for the PR @ya-mouse but I am not exactly sue what it is supposed to accomplish exactly and it is quite out of date. Now closing, please reopen a new one with some explanation if this is something you think is important and that you are still working on.

PR #3366, opened 2023-01-06
Add USB HID joystick support

nobonobo:
Add 6-Axis and 16-buttons Joystick support.

## changes

- add endpoint for HID Interrupt Out (need for FFB Controller)
- add machine/usb/joystick package
- HID MIDI Bulk Out Endpoint Number changes 5 to 7
- minimal fixed for hid and usb packages
- add joystick minimum sample

PR #3367, opened 2023-01-07
machine/usb: improve buffer size definition

sago35:
This PR fixes the following problems

> #3366
> Also, I couldn't get the usb-midi branch to work with samd51 on this branch.

I have confirmed that it works on the nobonobo/add-support-joystick branch to which I have added this change.

* examples/usb-midi
    * samd51 : wioterminal
    * rp2040 : pico
    * nrf52840 : feather-nrf52840-sense

However, I could not confirm that it works with samd21.
It appears that usb-midi is not working on samd21 as of the current dev branch.
This is a separate issue.

### DIFF
--- a/src/machine/usb.go
+++ b/src/machine/usb.go
@@ -66,10 +66,10 @@ var (
 var udd_ep_control_cache_buffer [256]uint8
 
 //go:align 4
-var udd_ep_in_cache_buffer [7][64]uint8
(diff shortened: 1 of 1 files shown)


sago35:
As a result of my research on samd21, it seems that if midi.NoteOn(), etc. is executed before USB initialization is completed, it will enter a mode where it cannot transmit.
I will consider how to fix it.

deadprogram:
Thanks for the fix @sago35 now merging.

PR #3368, opened 2023-01-07
src/runtime: add xorshift-based fastrand64

dgryski:
`hash/maphash` uses `runtime_fastrand64` now.

https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/hash/maphash/maphash.go;l=257


deadprogram:
Excellent :1234: as usual @dgryski thanks! Now merging.

PR #3370, opened 2023-01-07
fix resource leak in os.(*File).Close

achille-roussel:
Hello!

I am opening this PR to modify the `os.File` implementation to do two things:
- handle platform-specific logic on `Close`, allowing the release of resources that may be held by the file
- set the internal file handle to `nil` on `Close` and return `ErrClosed` afterwards if a file keeps being used

I am making this change to address the following issues:
- on Darwin, the `dirInfo` type references memory allocated by the C library, as well as a copy of the file descriptor created by `dup`. Because the object wasn't close, both the memory and file descriptor were being leaked.

- on Linux, the `dirInfo` type holds a buffer claimed from an internal pool. While I don't have evidence that this pooling is effective, the buffers were never released, resulting in no buffer pooling at all.
- Because the file handle was not released on close, multiple calls to `os.(*File).Close` could result in closing file descriptors that are reused, likely leading to bugs that would be really difficult to track down.

I am tempted to suggest that we should remove the buffer pooling entirely on Linux, since it was never effective, it seems like unnecessary complexity that has never provided any benefits, and the fix here could introduce new behaviors that may not be desirable (e.g. sync.Pool seems to never release memory in tinygo). Let me know what you think and I can make the change.

I also made a change to have `os.(*File).Fd` return `^uintptr(0)` after close, which matches the behavior of Go and seems better than the historical behavior of returning 0, which could be mistaken for stdin.

I have yet to test the code because I am running into what looks like linking issues against libc (see https://github.com/tinygo-org/tinygo/issues/3057#issuecomment-1374497605).

I do not have a great sense of how to validate that the underlying `dirInfo` value is closed on unix platforms, the compiler does ensure that the `close` method exists because it is called, but that doesn't protect against regressions. Happy to get guidance on this front.

Let me know if you'd like to see anything changed!

### DIFF
--- a/src/os/dir_darwin.go
+++ b/src/os/dir_darwin.go
@@ -147,7 +147,7 @@ func darwinOpenDir(fd syscallFd) (uintptr, string, error) {
 	return dir, "", nil
 }
 
-// Implemented in syscall/syscall_darwin.go.
+// Implemented in syscall/syscall_libc_darwin_*.go.
 
 //go:linkname closedir syscall.closedir
 func closedir(dir uintptr) (err error)
--- a/src/os/dir_other.go
+++ b/src/os/dir_other.go
@@ -13,6 +13,9 @@ import (
 type dirInfo struct {
 }
 
+func (*dirInfo) close() {
+}
+
 func (f *File) readdir(n int, mode readdirMode) (names []string, dirents []DirEntry, infos []FileInfo, err error) {
 	return nil, nil, nil, &PathError{Op: "readdir unimplemented", Err: syscall.ENOTDIR}
 }
--- a/src/os/dir_unix.go
+++ b/src/os/dir_unix.go
@@ -8,43 +8,29 @@ package os
 
 import (
 	"io"
-	"sync"
 	"syscall"
 	"unsafe"
 )
 
 // Auxiliary information if the File describes a directory
 type dirInfo struct {
-	buf  *[]byte // buffer for directory I/O
-	nbuf int     // length of buf; return value from Getdirentries
-	bufp int     // location of next record in buf.
+	nbuf int             // length of buf; return value from Getdirentries
+	bufp int             // location of next record in buf.
+	buf  [blockSize]byte // buffer for directory I/O
 }
 
 const (
 	// More than 5760 to work around https://golang.org/issue/24015.
-	blockSize = 8192
+	blockSize = 8192 - 2*unsafe.Sizeof(int(0))
 )
 
-var dirBufPool = sync.Pool{
-	New: func() interface{} {
-		// The buffer must be at least a block long.
-		buf := make([]byte, blockSize)
-		return &buf
-	},
-}
-
 func (d *dirInfo) close() {
-	if d.buf != nil {
-		dirBufPool.Put(d.buf)
-		d.buf = nil
-	}
 }
 
 func (f *File) readdir(n int, mode readdirMode) (names []string, dirents []DirEntry, infos []FileInfo, err error) {
 	// If this file has no dirinfo, create one.
 	if f.dirinfo == nil {
 		f.dirinfo = new(dirInfo)
-		f.dirinfo.buf = dirBufPool.Get().(*[]byte)
 	}
 	d := f.dirinfo
 
@@ -66,7 +52,7 @@ func (f *File) readdir(n int, mode readdirMode) (names []string, dirents []DirEn
 		if d.bufp >= d.nbuf {
 			d.bufp = 0
 			var errno error
-			d.nbuf, errno = syscall.ReadDirent(syscallFd(f.handle.(unixFileHandle)), *d.buf)
+			d.nbuf, errno = syscall.ReadDirent(syscallFd(f.handle.(unixFileHandle)), d.buf[:])
 			if d.nbuf < 0 {
 				errno = handleSyscallError(errno)
 			}
@@ -79,7 +65,7 @@ func (f *File) readdir(n int, mode readdirMode) (names []string, dirents []DirEn
 		}
 
 		// Drain the buffer
-		buf := (*d.buf)[d.bufp:d.nbuf]
+		buf := d.buf[d.bufp:d.nbuf]
 		reclen, ok := direntReclen(buf)
 		if !ok || reclen > uint64(len(buf)) {
 			break
--- a/src/os/file.go
+++ b/src/os/file.go
@@ -42,11 +42,11 @@ var lstat = Lstat
 func Mkdir(path string, perm FileMode) error {
 	fs, suffix := findMount(path)
 	if fs == nil {
-		return &PathError{"mkdir", path, ErrNotExist}
+		return &PathError{Op: "mkdir", Path: path, Err: ErrNotExist}
 	}
 	err := fs.Mkdir(suffix, perm)
 	if err != nil {
-		return &PathError{"mkdir", path, err}
+		return &PathError{Op: "mkdir", Path: path, Err: err}
 	}
 	return nil
 }
@@ -65,7 +65,7 @@ func fixCount(n int, err error) (int, error) {
 func Remove(path string) error {
 	fs, suffix := findMount(path)
 	if fs == nil {
-		return &PathError{"remove", path, ErrNotExist}
+		return &PathError{Op: "remove", Path: path, Err: ErrNotExist}
 	}
 	err := fs.Remove(suffix)
 	if err != nil {
@@ -84,11 +84,11 @@ func (f *File) Name() string {
 func OpenFile(name string, flag int, perm FileMode) (*File, error) {
 	fs, suffix := findMount(name)
 	if fs == nil {
-		return nil, &PathError{"open", name, ErrNotExist}
+		return nil, &PathError{Op: "open", Path: name, Err: ErrNotExist}
 	}
 	handle, err := fs.OpenFile(suffix, flag, perm)
 	if err != nil {
-		return nil, &PathError{"open", name, err}
+		return nil, &PathError{Op: "open", Path: name, Err: err}
 	}
 	return NewFile(handle, name), nil
 }
@@ -106,10 +106,14 @@ func Create(name string) (*File, error) {
 // Read reads up to len(b) bytes from the File. It returns the number of bytes
 // read and any error encountered. At end of file, Read returns 0, io.EOF.
 func (f *File) Read(b []byte) (n int, err error) {
-	n, err = f.handle.Read(b)
+	if f.handle == nil {
+		err = ErrClosed
+	} else {
+		n, err = f.handle.Read(b)
+	}
 	// TODO: want to always wrap, like upstream, but ReadFile() compares against exactly io.EOF?
 	if err != nil && err != io.EOF {
-		err = &PathError{"read", f.name, err}
+		err = &PathError{Op: "read", Path: f.name, Err: err}
 	}
 	return
 }
@@ -123,13 +127,16 @@ func (f *File) ReadAt(b []byte, offset int64) (n int, err error) {
 	if offset < 0 {
 		return 0, &PathError{Op: "readat", Path: f.name, Err: errNegativeOffset}
 	}
+	if f.handle == nil {
+		return 0, &PathError{Op: "readat", Path: f.name, Err: ErrClosed}
+	}
 
 	for len(b) > 0 {
 		m, e := f.handle.ReadAt(b, offset)
 		if e != nil {
 			// TODO: want to always wrap, like upstream, but TestReadAtEOF compares against exactly io.EOF?
 			if e != io.EOF {
-				err = &PathError{"readat", f.name, e}
+				err = &PathError{Op: "readat", Path: f.name, Err: e}
 			} else {
 				err = e
 			}
@@ -146,9 +153,13 @@ func (f *File) ReadAt(b []byte, offset int64) (n int, err error) {
 // Write writes len(b) bytes to the File. It returns the number of bytes written
 // and an error, if any. Write returns a non-nil error when n != len(b).
 func (f *File) Write(b []byte) (n int, err error) {
-	n, err = f.handle.Write(b)
+	if f.handle == nil {
+		err = ErrClosed
+	} else {
+		n, err = f.handle.Write(b)
+	}
 	if err != nil {
-		err = &PathError{"write", f.name, err}
+		err = &PathError{Op: "write", Path: f.name, Err: err}
 	}
 	return
 }
@@ -160,14 +171,33 @@ func (f *File) WriteString(s string) (n int, err error) {
 }
 
 func (f *File) WriteAt(b []byte, off int64) (n int, err error) {
-	return 0, ErrNotImplemented
+	if f.handle == nil {
+		err = ErrClosed
+	} else {
+		err = ErrNotImplemented
+	}
+	err = &PathError{Op: "writeat", Path: f.name, Err: err}
+	return
 }
 
 // Close closes the File, rendering it unusable for I/O.
 func (f *File) Close() (err error) {
-	err = f.handle.Close()
+	if f.handle == nil {
+		err = ErrClosed
+	} else {
+		// Some platforms manage extra state other than the system handle which
+		// needs to be released when the file is closed. For example, darwin
+		// files have a DIR object holding a dup of the file descriptor.
+		//
+		// These platform-specific logic is provided by the (*file).close method
+		// which is why we do not call the handle's Close method directly.
+		err = f.file.close()
+		if err == nil {
+			f.handle = nil
+		}
+	}
 	if err != nil {
-		err = &PathError{"close", f.name, err}
+		err = &PathError{Op: "close", Path: f.name, Err: err}
 	}
 	return
 }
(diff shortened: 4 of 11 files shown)


dgryski (changes requested):
CI failures due to missing functions on wasi (at least).  But otherwise LGTM once those are fixed.

achille-roussel:
Thanks for the quick reply @dgryski 

I addressed the WASI build issue, but I believe there is another problem with the CI run for OSX https://github.com/tinygo-org/tinygo/actions/runs/3862629890/jobs/6584319540 which seems similar to the issue I reported in https://github.com/tinygo-org/tinygo/issues/3057#issuecomment-1374497605

```
tinygo:ld.lld: error: undefined symbol: _syscall.closedir
>>> referenced by dir_darwin.go:23 (/Users/runner/work/tinygo/tinygo/src/os/dir_darwin.go:23)
>>>               /Users/runner/Library/Caches/tinygo/thinlto/llvmcache-4584E20AB8F095322715CB56E98C1BCE020DF609:(symbol _(*os.File).Close+0x37)
failed to run tool: ld.lld
```

It's possible that the `closedir` call was erased from the binaries built since it was never used, and now that it's included there are CI caches that have gotten invalidated and trigger the issue?

achille-roussel:
Another issue triggered in CI, which I also observe in my development environment, but seems like a panic during the compilation of tests? https://github.com/tinygo-org/tinygo/actions/runs/3862629893/jobs/6584368992

```
$ make tinygo-test-wasi-fast
/home/runner/go/bin/tinygo test -target wasi compress/zlib container/heap container/list container/ring crypto/des crypto/internal/subtle crypto/md5 crypto/rc4 crypto/sha1 crypto/sha256 crypto/sha512 debug/macho embed/internal/embedtest encoding encoding/ascii85 encoding/base32 encoding/csv encoding/hex go/scanner hash hash/adler32 hash/crc64 hash/fnv html internal/itoa internal/profile math math/cmplx net net/http/internal/ascii net/mail os path reflect sync testing testing/iotest text/scanner unicode unicode/utf16 unicode/utf8  crypto/internal/nistec/fiat ./tests/runtime_wasi
panic: local has not been parsed: parameter c : *closeOnce

goroutine 171 [running]:
github.com/tinygo-org/tinygo/compiler.(*builder).getValue(0xc00013ab40?, {0x[5](https://github.com/tinygo-org/tinygo/actions/runs/3862629893/jobs/6584368992#step:7:6)5c45b8?, 0xc009c754a0?})
	/__w/tinygo/tinygo/compiler/compiler.go:1811 +0x585
github.com/tinygo-org/tinygo/compiler.(*builder).createFunction(0xc00013ab40)
	/__w/tinygo/tinygo/compiler/compiler.go:12[6](https://github.com/tinygo-org/tinygo/actions/runs/3862629893/jobs/6584368992#step:7:7)2 +0x[7](https://github.com/tinygo-org/tinygo/actions/runs/3862629893/jobs/6584368992#step:7:8)5c
github.com/tinygo-org/tinygo/compiler.(*compilerContext).createPackage(0xc00d0fbd00, {0xc00a5[8](https://github.com/tinygo-org/tinygo/actions/runs/3862629893/jobs/6584368992#step:7:9)8ea0?}, 0xc00a708[9](https://github.com/tinygo-org/tinygo/actions/runs/3862629893/jobs/6584368992#step:7:10)80)
	/__w/tinygo/tinygo/compiler/compiler.go:877 +0xa0f
github.com/tinygo-org/tinygo/compiler.CompilePackage({0xc0014ef7b7?, 0xc00a421da0?}, 0xc0020d90e0, 0xc00a708980, {0xc0096cacc0?}, 0x55396d0?, 0x9?)
	/__w/tinygo/tinygo/compiler/compiler.go:305 +0x457
github.com/tinygo-org/tinygo/builder.Build.func3(0xc00a774c00)
	/__w/tinygo/tinygo/builder/build.go:348 +0x216
github.com/tinygo-org/tinygo/builder.runJob(0xc00a774c00, 0xc008b68020?)
	/__w/tinygo/tinygo/builder/jobs.go:222 +0x4f
	/__w/tinygo/tinygo/builder/jobs.go:[12](https://github.com/tinygo-org/tinygo/actions/runs/3862629893/jobs/6584368992#step:7:13)3 +0x5db
make: *** [Makefile:4[13](https://github.com/tinygo-org/tinygo/actions/runs/3862629893/jobs/6584368992#step:7:14): tinygo-test-wasi-fast] Error 2
```

achille-roussel:
I think I got to the bottom of the linker issues, which were due to missing functions on Darwin in `src/syscall`.

The one issue that remains is the compiler panic, which I believe is coming from compiling `os/exec/exec.go` in Go, particularly this type:
```go
type closeOnce struct {
    *os.File

    once sync.Once
    err  error
}

func (c *closeOnce) Close() error {
    c.once.Do(c.close)
    return c.err
}

func (c *closeOnce) close() {
    c.err = c.File.Close()
}
```

I added logging to tinygo in `createPackage` and `createFunction` to understand what is happening, and it appears that the `close` method does not see its receiver parameter `c`:
```
(*os/exec.closeOnce).Close
parameter c : *closeOnce
(*os/exec.closeOnce).close$bound
(*os/exec.closeOnce).close
```
I'm not sure what this `$bound` function is (maybe it's the generated closure in `c.once.Do(c.close)`?), will investigate more and try to understand why the parameter is missing.

achille-roussel:
By modifying the code to remove the `close` method, the compiler does not crash anymore:
```go
func (c *closeOnce) Close() error {
    //c.once.Do(c.close)
    c.once.Do(func() { c.err = c.File.Close() })
    return c.err
}
```

I think I have a hint as to what might be happening, the `os/exec.(*closeOnce).close` method appears to be instantiated twice, the second time result in mistakenly assuming that it's a redeclaration and errors. Here are the two instantiation points:

(sorry the line numbers may mismatch a bit because I added debug statements locally)

```
github.com/tinygo-org/tinygo/compiler.(*compilerContext).getFunction(0x14002dfa600, 0x14004d6a680)
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/compiler/symbol.go:63 +0x104
github.com/tinygo-org/tinygo/compiler.(*builder).createFunctionCall(0x14000e2e000, 0x14003a1b1c0)
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/compiler/compiler.go:1742 +0x7d8
github.com/tinygo-org/tinygo/compiler.(*builder).createExpr(0x14000e2e000, {0x109a68840?, 0x14003a1b180?})
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/compiler/compiler.go:1898 +0x1fa0
github.com/tinygo-org/tinygo/compiler.(*builder).createInstruction(0x14000e2e000, {0x109a687f8?, 0x14003a1b180?})
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/compiler/compiler.go:1397 +0x778
github.com/tinygo-org/tinygo/compiler.(*builder).createFunction(0x14000e2e000)
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/compiler/compiler.go:1295 +0x8dc
github.com/tinygo-org/tinygo/compiler.(*compilerContext).getFunction(0x14002dfa600, 0x140033eba00)
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/compiler/symbol.go:210 +0xa94
github.com/tinygo-org/tinygo/compiler.(*builder).parseMakeClosure(0x14000d51e00, 0x14000103a40)
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/compiler/func.go:146 +0xd0
github.com/tinygo-org/tinygo/compiler.(*builder).createExpr(0x14000d51e00, {0x109a690f8?, 0x14000103a40?})
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/compiler/compiler.go:2084 +0x2a80
github.com/tinygo-org/tinygo/compiler.(*builder).createInstruction(0x14000d51e00, {0x109a690b0?, 0x14000103a40?})
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/compiler/compiler.go:1397 +0x778
github.com/tinygo-org/tinygo/compiler.(*builder).createFunction(0x14000d51e00)
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/compiler/compiler.go:1295 +0x8dc
github.com/tinygo-org/tinygo/compiler.(*compilerContext).createPackage(0x14002dfa600, {0x14004c2a340?}, 0x14003167500)
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/compiler/compiler.go:884 +0x930
github.com/tinygo-org/tinygo/compiler.CompilePackage({0x1400285abe7?, 0x14004ad5920?}, 0x1400285fe00, 0x14003167500, {0x14000045418?}, 0x10829b407?, 0x38?)
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/compiler/compiler.go:305 +0x3ec
github.com/tinygo-org/tinygo/builder.Build.func3(0x14004d7e2a0)
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/builder/build.go:348 +0x1a4
github.com/tinygo-org/tinygo/builder.runJob(0x14004d7e2a0, 0x0?)
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/builder/jobs.go:222 +0x48
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/builder/jobs.go:123 +0x43c
```

```
github.com/tinygo-org/tinygo/compiler.(*compilerContext).getFunction(0x14002dfa600, 0x14004d6a680)
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/compiler/symbol.go:63 +0x104
github.com/tinygo-org/tinygo/compiler.newBuilder(0x14002dfa600, {0x14000010020?}, 0x14004d6a680)
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/compiler/compiler.go:170 +0x34
github.com/tinygo-org/tinygo/compiler.(*compilerContext).createPackage(0x14002dfa600, {0x14004c2a340?}, 0x14003167500)
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/compiler/compiler.go:883 +0x92c
github.com/tinygo-org/tinygo/compiler.CompilePackage({0x1400285abe7?, 0x14004ad5920?}, 0x1400285fe00, 0x14003167500, {0x14000045418?}, 0x10829b407?, 0x38?)
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/compiler/compiler.go:305 +0x3ec
github.com/tinygo-org/tinygo/builder.Build.func3(0x14004d7e2a0)
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/builder/build.go:348 +0x1a4
github.com/tinygo-org/tinygo/builder.runJob(0x14004d7e2a0, 0x0?)
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/builder/jobs.go:222 +0x48
	/Users/achilleroussel/dev/src/github.com/achille-roussel/tinygo/builder/jobs.go:123 +0x43c
```

Based on my understanding, when the `os/exec.(*closeOnce).Close` method is created, it iterates over its expressions and find the bound call to `c.close` (maybe via inlining etc...), makes a call to `getFunction`, which registers `os/exec.(*closeOnce).close`.

Later on, as the compiler iterates on methods of the `closeOnce` type, it finds `os/exec.(*closeOnce).close` and tries to create the function. However, it is already registered and the compiler therefore mistakenly assumes that it was redeclared. This causes an early exist at https://github.com/tinygo-org/tinygo/blob/release/compiler/compiler.go#L1037-L1045, which leaves the `b.locals` map unpopulated (much like the rest of the function state) and triggers a panic when the code reaches https://github.com/tinygo-org/tinygo/blob/release/compiler/compiler.go#L1768-L1774.

It seems like the compilation should probably abort if it finds a redeclared function, which would prevent the panic (tho not fix the issue).

I am not familiar enough with the codebase yet to picture what a fix should be, it seems that we should create all methods first, then go through their expressions and instantiate their content?

achille-roussel:
Commenting the check for redeclared functions seems to work around the issue, no compiler tests are failing, is it a fair assumption to say that this code path was never exercised? (maybe redeclaration is validated elsewhere already?)

The WASI test suite completes successfully now but a new error is reported which might be causing CI to fail still:
```
Function return type does not match operand type of return inst!
  ret void, !dbg !1368
 %runtime._interface = type { i32, ptr }error: verification error after compiling package os/exec
```
I will look into this.

achille-roussel:
Taking a step back here: to my knowledge, WASI currently does not have any mechanisms to allow spawning processes (at least according to https://github.com/bytecodealliance/wasmtime/blob/main/docs/WASI-background.md#stick-a-fork-in-it)

Maybe it's not worth spending more time trying to fix compilation of `os/exec` and we should just disable it for now?

achille-roussel:
I investigated this further and here is what I understand of the issue so far:

- `os/exec.closeOnce` has an embedded `*os.File` field for which the compiler has to generate _inherited_ methods
- `os.File` has an embedded `*os.file` field which now has a `close` method
- `os.(*File).Close` is inlined, so is `os/exec.(*closeOnce).close`, resulting in the compiler generating a wrapper function named `os/exec.(*closeOnce).close` calling `os.(*file).close`
- a `closeOnce` value is instantiated and converted to an interface in the `os/exec` code, which causes the first instantiation of the wrapper for `os/exec.(*closeOnce).close`
- the second instantiation occurs when the method is generated from the source code

Based on my understanding, the compilation issue comes from a conflict between the SSA-generated wrapper inheriting the method name of `os/exec.(*closeOnce).close`, and the instantiation of the actual method from the source code, which is also named `os/exec.(*closeOnce).close`.

The wrapper ends up being registered under the source function name by [`llvm.AddFunction`](https://github.com/tinygo-org/tinygo/blob/release/compiler/symbol.go#L95).

Why wasn't it happening before? This is beyond my understanding but it has to do with changing the code of `os.(*File).Close` I believe. It may also have to do with the order in which the generated wrappers and source code are instantiated?

To work around the issue, I have implemented a change to _mangle_ the wrapper names by appending the synthetic value to the function name, see https://github.com/tinygo-org/tinygo/pull/3370/commits/8c3b7a7a2cf18961fa9df6211fd5bc3591763230#diff-131ba75d1f2276484b60c9882004078948adae81cba2abf044098c4ec73218b0R222

While this appears to address the issue, it really needs vetting from someone with more context than I have. For example, I don't know if that may cause issues with the generation of debug symbols. The synthetic value is also a human-readable value like `wrapper for os.(*file).close` which can be somewhat unexpected to find as a symbol name.

compiler/symbol.go
```diff
@@ -218,6 +218,9 @@ func (c *compilerContext) getFunctionInfo(f *ssa.Function) functionInfo {
 		// Pick the default linkName.
 		linkName: f.RelString(nil),
 	}
+	if isWrapperFunction(f) {
+		info.linkName += "$" + f.Synthetic
+	}
```
aykevl: Can you extract this and move it into a separate PR with tests?
You can test this by putting the Go code that led to the bug in one of the compiler/testdata/*.go files and running `go test ./compiler -update`.

aykevl (changes requested):

aykevl:
So yeah it looks like you found a compiler bug, congratulations! :wink:
To better discuss it and to better see what's going on, having a separate PR makes things a lot easier. But I'll reply to some comments already:

> Commenting the check for redeclared functions seems to work around the issue, no compiler tests are failing, is it a fair assumption to say that this code path was never exercised? (maybe redeclaration is validated elsewhere already?)

This check is really just a sanity check (like an assert). A bug-free compiler shouldn't declare the same function twice (although it can be done using `//go:linkname` I think). So any time you hit it, it means there is a bug somewhere else that should be fixed.
Ignoring it will probably produce a working binary, but it may result in duplicates or weird symbol names so it's somewhat incorrect anyway.

> `os.(*File).Close` is inlined, so is `os/exec.(*closeOnce).close`, resulting in the compiler generating a wrapper function named `os/exec.(*closeOnce).close` calling `os.(*file).close`

No, inlining happens at a later stage in the compiler. But the SSA level sometimes wraps functions for various reasons, it's certainly possible there's a bug with this wrapping (either in golang.org/x/tools/go/ssa or in TinyGo).

compiler/symbol.go
achille-roussel: Thanks for the feedback! I will isolate the compiler fix in a separate PR and attempt to create a test case that reproduces the issue 👍

achille-roussel:
I opened https://github.com/tinygo-org/tinygo/pull/3381 and rebased this branch on top of it.

Let me know if you would like to see any other changes made to this branch!

achille-roussel:
I think this is ready for (re)review. 

The code has changed in Go 1.20 and the parts that triggered a panic during compilation do not seem to exist anymore, so there are no more blockers to merging this diff.

I can cleanup the commit history if needed.

Let me know if there are any other changes you would like to see!

achille-roussel:
Squashed into a single commit as discussed on Slack, tho one of the windows tests has failed on the last run. Is this a known flaky test or should I investigate?
```
FAIL	net	0.055s
```

deadprogram:
> Is this a known flaky test

Sadly, yes. So nothing to do with your changes, certainly.

deadprogram:
Thank you for the fixes and all the requested changes @achille-roussel also thanks @dgryski and @aykevl for review.

I verified all tests are currently passing in #3612 

Now merging this PR, thanks again @achille-roussel

achille-roussel:
Thanks for following up and merging the fix!

I'll rebase and send the rest of the PRs I has to complete WASI support later this week.

PR #3371, opened 2023-01-07
document the origin of src/os/file.go and reason why the os.lstat variable exists

achille-roussel:
Hello!

While working on https://github.com/tinygo-org/tinygo/pull/3370, I noticed that there was an unused global variable in the `os` package. This PR removes it.

The comment claimed that the variable was used in test, but I could not find any reference to it in tests. I believe we can safely remove it and replace references to this variable with direct calls to `Lstat` instead.

### DIFF
--- a/src/os/file.go
+++ b/src/os/file.go
@@ -2,6 +2,14 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+// This file was originally copied from Go, see:
+// https://github.com/golang/go/blob/master/src/os/file.go
+//
+// Some of the code inherited from Go is not used anymore in Tinygo, but we keep
+// changes to a minimum to help simplify bringing changes (e.g. the lstat global
+// is not used here anymore, but we might need it if we add tests from Go in
+// this package).
+
 // Package os implements a subset of the Go "os" package. See
 // https://godoc.org/os for details.
 //


achille-roussel:
There is a failure reported when testing the `net` package on windows, tho without much details about what may have gone wrong?
```
FAIL	net	0.062s
```

dgryski:
The `file.go` file is mostly imported directly from upstream Go.  Keeping the changes to this file to a minimum make importing future changes easier.  The `lstat` variable and comment come from the upstream version of the file.

achille-roussel:
I opened because the comment was misleading, I thought it might save someone time if they don't run into this code again.

It seems to me that the files have already diverged quite a lot from Go, so I'm not sure it's practical to aim maintaining similarities, but I'm not a maintainer of this project so I have very little context and what the right trade off is here.

I can update the comment maybe to indicate that it's intended to stay for future opportunities to import future changes? Happy to close as well if this is not a useful change :)

dgryski:
What might be useful would be a comment at the top of the file indicating where this file comes from and why the code might seem unnatural for tinygo.

dgryski:
Thanks.  Can you rebase to remove the initial commit and the revert?

dgryski (approved):
LGTM

aykevl:
Looks good to me. That comment should really have been there from the beginning (for proper attribution of the code).

PR #3373, opened 2023-01-08
machine/usb: change to not send before endpoint initialization

sago35:
Until now, the problem was caused by trying to send before the endpoint was initialized.
This PR will add code to check if endpoint initialization is complete and resolve the issue.

* issues
  * https://github.com/tinygo-org/tinygo/pull/3367#issuecomment-1374365943
  * https://twitter.com/sago35tk/status/1611911029059117057?s=20&t=EbY5u85_AdCGRyooF5CHiQ

USBCDC monitored linestate and had no problems.

https://github.com/tinygo-org/tinygo/blob/b5ad81c884981f42c1ac491cc1d195aea886ddbe/src/machine/usb/cdc/usbcdc.go#L100-L108

### DIFF
--- a/src/machine/usb.go
+++ b/src/machine/usb.go
@@ -10,7 +10,8 @@ import (
 )
 
 type USBDevice struct {
-	initcomplete bool
+	initcomplete         bool
+	InitEndpointComplete bool
 }
 
 var (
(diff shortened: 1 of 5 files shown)


deadprogram:
Good idea @sago35 thanks for the fix/improvement. Now merging.

PR #3378, opened 2023-01-11
machine/rp2040: XIP flash read/write

deadprogram:
This is an experimental branch trying (unsuccessfully so far) to implement something like https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_flash/flash.c

Of course this might not be possible due to https://github.com/raspberrypi/pico-sdk/issues/1089


deadprogram:
Also see https://github.com/adafruit/circuitpython/blob/main/ports/raspberrypi/common-hal/nvm/ByteArray.c

kenbell:
@deadprogram - I tried to do something very similar and hit a brick wall.  My last thought (that I never validated) is needing to run some of this code from RAM, not flash.  If you look at CircuitPython, it's using the Pico SDK, which does this:
```c
void __no_inline_not_in_flash_func(flash_range_program)(uint32_t flash_offs, const uint8_t *data, size_t count) {
```
(https://github.com/raspberrypi/pico-sdk/blob/2062372d203b372849d573f252cf7c6dc2800c0a/src/rp2_common/hardware_flash/flash.c#L86)

That basically amounts to putting the function in a RAM section in the linker script:
```c
#define __not_in_flash(group) __attribute__((section(".time_critical." group)))
```

The linker script maps `*(.time_critical*)` to RAM:
https://github.com/raspberrypi/pico-sdk/blob/afc10f3599c27147a6f34781b7102d86f58aa5f6/src/rp2_common/pico_standard_link/memmap_no_flash.ld#L51

I think this is because underneath, XIP is being disabled, so any attempt to access an address mapped to flash is hanging up.

We might need to figure out how to have go functions execute from RAM using `//go:section`?

deadprogram:
@kenbell I bet you are correct, not exactly sure how to do that. @aykevl any comments here?

kenbell:
Looks like their start-up code has to explicitly copy these bits into RAM super early... https://github.com/raspberrypi/pico-sdk/blob/2e6142b15b8a75c1227dd3edbe839193b2bf9041/src/rp2_common/pico_standard_link/crt0.S#L230

kenbell:
@deadprogram PR #3382 should help

bgould:
Yes! Merged this with the PR from @kenbell and marked the flash methods with //go:section .ramtext and it keeps the program from crashing.  Output is not correct ... but could be worse!

```
writing flash
reading back flash
ha``i  gb`d
reading back flash
ha``i  gb`d
reading back flash
ha``i
writing flash
reading back flash
ha``i  gb`d
reading back flash
ha``i  gb`d
reading back flash
ha``i
```

deadprogram:
Replaced by #3536

Issue #3369, opened 2023-01-07
unexpected token ILLEGAL when compiling <sys/socket.h> with wasi target


charconstpointer:
I was playing around with https://github.com/alphahorizonio/unisockets and I've noticed some issues when compiling its code for `wasi`, narrowed it down to the snippet below and it seems like it's the issue with `<sys/socket.h>` and underlying wasi implementation
```go
package main

/*
#include <sys/socket.h>
*/
import "C"

const (
	SHUT_RDWR = int32(C.SHUT_RDWR)
)

func main() {
}
```

Compiled with `tinygo build -target wasi main.go`
Output `tinygo/lib/wasi-libc/sysroot/include/__header_sys_socket.h:12:35: unexpected token ILLEGAL, expected end of expression`

---
Env
- Tinygo version: 0.26.0
- go version go1.19.4 linux/amd64

aykevl:
I've implemented support for `|` in https://github.com/tinygo-org/tinygo/pull/3379. It doesn't fully fix this bug though, the new error is the following:

```
$ tinygo run -target=wasi ./tmp/issue3369.go
# command-line-arguments
lib/wasi-libc/sysroot/include/__header_sys_socket.h:12:20: undeclared name: C.SHUT_RD
lib/wasi-libc/sysroot/include/__header_sys_socket.h:12:30: undeclared name: C.SHUT_WR
```

Not sure what's going on there.

PR #3379, opened 2023-01-11
cgo: add support for bitwise operators

aykevl:
This is a partial fix for https://github.com/tinygo-org/tinygo/issues/3369.

### DIFF
--- a/cgo/const.go
+++ b/cgo/const.go
@@ -14,6 +14,9 @@ import (
 var (
 	prefixParseFns map[token.Token]func(*tokenizer) (ast.Expr, *scanner.Error)
 	precedences    = map[token.Token]int{
+		token.OR:  precedenceOr,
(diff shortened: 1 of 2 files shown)


dgryski (approved):
LGTM.  Do we have to write our own C parser or is there something else we build upon?

deadprogram:

aykevl:
> Do we have to write our own C parser or is there something else we build upon?

I've looked into alternatives and doing our own parser seems to be the best solution. Note that this isn't a full parser, it really just converts a C AST to a Go AST to let the Go parser deal with it. This will never be able to parse all C expressions, but hopefully enough to be usable.
(Upstream CGo is actually quite terrible: for each requested type, it constructs C code that will trigger an error and then it will read the error messages to figure out the type of the symbol. And I believe for macros it will pull the value out of generated DWARF after compiling).

PR #3381, opened 2023-01-13
Fix function and wrapper name conflict

achille-roussel:
This PR extracts the compiler fixes from https://github.com/tinygo-org/tinygo/pull/3370.

---
> Carrying context from the original PR https://github.com/tinygo-org/tinygo/pull/3370#issuecomment-1374890405

I investigated this further and here is what I understand of the issue so far:

- `os/exec.closeOnce` has an embedded `*os.File` field for which the compiler has to generate _inherited_ methods
- `os.File` has an embedded `*os.file` field which now has a `close` method
- `os.(*File).Close` is inlined, so is `os/exec.(*closeOnce).close`, resulting in the compiler generating a wrapper function named `os/exec.(*closeOnce).close` calling `os.(*file).close`
- a `closeOnce` value is instantiated and converted to an interface in the `os/exec` code, which causes the first instantiation of the wrapper for `os/exec.(*closeOnce).close`
- the second instantiation occurs when the method is generated from the source code

Based on my understanding, the compilation issue comes from a conflict between the SSA-generated wrapper inheriting the method name of `os/exec.(*closeOnce).close`, and the instantiation of the actual method from the source code, which is also named `os/exec.(*closeOnce).close`.

The wrapper ends up being registered under the source function name by [`llvm.AddFunction`](https://github.com/tinygo-org/tinygo/blob/release/compiler/symbol.go#L95).

Why wasn't it happening before? This is beyond my understanding but it has to do with changing the code of `os.(*File).Close` I believe. It may also have to do with the order in which the generated wrappers and source code are instantiated?

To work around the issue, I have implemented a change to _mangle_ the wrapper names by appending the synthetic value to the function name, see https://github.com/tinygo-org/tinygo/pull/3370/commits/8c3b7a7a2cf18961fa9df6211fd5bc3591763230#diff-131ba75d1f2276484b60c9882004078948adae81cba2abf044098c4ec73218b0R222

While this appears to address the issue, it really needs vetting from someone with more context than I have. For example, I don't know if that may cause issues with the generation of debug symbols. The synthetic value is also a human-readable value like `wrapper for os.(*file).close` which can be somewhat unexpected to find as a symbol name.

---

I added the `compiler/testdata/exec.go` file, which is a copy of Go's `src/os/exec/exec.go` file (with a few adjustments to allow standalone compilation).

That being said, it hasn't triggered the bug in this isolated PR, I am only able to reproduce it when adding the changes to `src/os` I made in the original PR.


aykevl:
> That being said, it hasn't triggered the bug in this isolated PR, I am only able to reproduce it when adding the changes to `src/os` I made in the original PR.

Then it's not testing what it is supposed to be testing.

In any case, this test is way, way too large. If it reproduces the bug at all, it's near impossible to find in a 10000 line IR file. I'm sorry, but it really has to be a lot smaller than this to be useful.
One thing you could try is to see which parts of the IR file change when you update the test with or without the patch. There should be a change somewhere.

achille-roussel:
@aykevl I'm sorry but I don't think I will be able to figure out how to reproduce the code using the approach you suggested to add source files to `compiler/testdata`. It is probably due to my lack of experience with the tingyo compiler internals, it seems that there is code removal happening that results in not exercising the conditions that triggered the panic in #3370 

I would like to suggest an alternative testing strategy; I can add a unit test for `compiler.(*compilerContext).getFunction` which will exercise the conditions I observed. It won't be as future-proof as a test based on a full compilation of Go source files, and it will be tied to the internal shape of the compiler code, but it will protect against regressions and will be a warning to anyone who modifies those internals (if it ever happens, `getFunction` seems pretty core to how the compiler is structured).

Let me know what you think!

dgryski:
Can we drop the test for now to get this fix merged?  File a issue for a better test later?

deadprogram:
Still seems like some CI errors happening on this https://github.com/tinygo-org/tinygo/actions/runs/4117575684/jobs/7109884545#step:16:330

achille-roussel:
Actually, after updating to the latest dev branch + using Go 1.20, I can't reproduce the original panic that triggered in  https://github.com/tinygo-org/tinygo/pull/3370

I'm tempted to say let's close this PR then? If the compiler doesn't crash and `tinygo-test-wasi-fast` passes, it's good enough?

achille-roussel:
Ah! the `closeOnce` type does not seem to exist anymore in `os/exec` in Go 1.20... so maybe tinygo still would have trouble with the code but it isn't being exercised anymore.

Nevertheless, I think this can be revisited if the panic reoccurs, and left as-is for now.

dgryski:
While CI only runs against the latest version (currently 1.20) I think we still want to support 1.19, so we should still fix this issue.

aykevl:
This is a really fundamental change to the compiler, so I would really want to see a test for it. _Especially_ if it doesn't seem reproducible anymore on Go 1.20.

> While CI only runs against the latest version (currently 1.20) I think we still want to support 1.19, so we should still fix this issue.

Please check whether the original issue reproduces on the latest release with Go 1.19. While adding support for Go 1.20, there was also an update to the x/tools/go/ssa package which might have fixed the bug.

deadprogram:
@achille-roussel @aykevl @dgryski can we close this now?

achille-roussel:
I believe we can close if the underlying issue I was trying to fix isn't manifesting anymore 👍

deadprogram:
OK thanks everyone!

PR #3382, opened 2023-01-13
arm: enable functions in RAM for go & cgo

kenbell:
The name of the section `.ramtext` is a bit questionable - could be a better name.  Suggestions?

I dropped the functions into the `.data` region of the binaries since the existing logic in `preinit` to initialize globals from flash also magically copies the functions into RAM.  There's an argument for having a dedicated region if there is support for 'NX' bits on any micro-controllers.

### DIFF
--- a/compiler/symbol.go
+++ b/compiler/symbol.go
@@ -279,8 +279,12 @@ func (info *functionInfo) parsePragmas(f *ssa.Function) {
 					info.linkName = parts[2]
 				}
 			case "//go:section":
+				// Only enable go:section when the package imports "unsafe".
+				// go:section also implies go:noinline since inlining could
+				// move the code to a different section than that requested.
 				if len(parts) == 2 && hasUnsafeImport(f.Pkg.Pkg) {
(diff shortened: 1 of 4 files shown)


bgould:
> The name of the section `.ramtext` is a bit questionable - could be a better name. Suggestions?

@aykevl do you have an opinion about this, and/or any concerns with adding this in general?

deadprogram:
> The name of the section .ramtext is a bit questionable - could be a better name. Suggestions?

How about `.ramfunctions`?

Also, this branch needs some modifications to take into account the recent flash related changes in `dev` branch.

kenbell:
@deadprogram I think this is good to go - I've called the section '.ramfuncs'

deadprogram (approved):
Looks good to me, @aykevl any comments from you before I merge?

deadprogram:
This is working well in #3536 which I just merged, so closing this PR since this change was picked up over there. Thank you very much for your work on this @kenbell

PR #3384, opened 2023-01-13
compiler: update golang.org/x/tools/ssa

aykevl:
This package needs to be updated to support Go 1.20. There were a few backwards incompatible changes that required updates to the compiler package.

I've separated out this change to make sure all the CI tests are run and there are no regressions.


aykevl:
I ran the test corpus and it still passes with this change.

dgryski (approved):
Corpus passes.  LGTM.

deadprogram:

PR #3385, opened 2023-01-14
test: print package name when compilation failed

aykevl:
Before this patch, a compile error would prevent the 'ok' or 'FAIL' line to be printed. That's unexpected. This patch changes the code in such a way that it's obvious a test result line is printed in all cases.

Found while working on Go 1.20.


dgryski (approved):
LGTM

deadprogram:
Seems like something did not work as expected here @aykevl please check CI fails.

aykevl:
This should fix the CI error.

deadprogram:

PR #3387, opened 2023-01-14
testing: implement t.Setenv

aykevl:
This method has been added in Go 1.17 and is used in archive/zip starting with Go 1.20. Therefore, this method is now needed in Go 1.20.

I've left out the parts that disable parallel execution of tests, because we don't do that in TinyGo.

See:
* https://github.com/golang/go/issues/41260
* https://go-review.googlesource.com/c/go/+/260577


dgryski (approved):
LGTM

deadprogram:
Thank you @aykevl and @dgryski for review. Now merging.

PR #3391, opened 2023-01-16
Initial Go 1.20 support

aykevl:
Not all features are supported yet and some tests still fail, but most code should compile with these patches.
We can work on full support with this merged, to get it in shape for Go 1.20 when it is released.


deadprogram (approved):

deadprogram:
Thank you for working on this @aykevl now merging.

QuLogic:
If I've applied this correctly this time, I think there's only one failure now, #3408.

PR #3395, opened 2023-01-17
wasm: avoid miscompile with ThinLTO

aykevl:
I found that when I enable ThinLTO, a miscompilation triggers that had
been hidden all the time previously. The bug appears to happen as
follows:

 1. TinyGo generates a function with a runtime.trackPointer call, but
    without an alloca (or the alloca gets optimized away).
 2. LLVM sees that no alloca needs to be kept alive across the
    runtime.trackPointer call, and therefore it adds the 'tail' flag.
    One of the effects of this flag is that it makes it undefined
    behavior to keep allocas alive across the call (which is still safe
    at that point).
 3. The GC lowering pass adds a stack slot alloca and converts
    runtime.trackPointer calls into alloca stores.

The last step triggers the bug: the compiler inserts an alloca where
there was none before but that's not valid as long as the 'tail' flag is
present.

This patch fixes the bug in a somewhat dirty way, by always creating a
dummy alloca so that LLVM won't do the optimization in step 2 (and
possibly other optimizations that rely on there being no alloca
instruction).

While this patch is only visible with #3311, I think it's a latent bug and not one caused by #3311.


aykevl:
Weirdly, this patch itself increases binary size by about 0.8% but when rebased on the ThinLTO patch it actually reduces binary size by around 2%. I suppose the increase is because of the dummy alloca which might not be optimized away. This alloca certainly should be optimized away with ThinLTO (which explains why it doesn't increase binary size anymore) but why it results in a reduction I don't know.

Anyway, normally I'd be worried about a 0.8% binary size increase but because that will all be negated by #3311 I'm not at all worried.

Oh and the corpus test passes with this patch, except for blackfriday/v2 (which is probably unrelated).

dgryski:
I had 10 successful blackfriday test runs with both `dev` and this branch.

dgryski (approved):
LGTM

deadprogram:
Some very fine work here @aykevl and @dgryski thank you! Now merging.

PR #3398, opened 2023-01-19
rp2040: rtc support

ysoldak:
Real Time Clock on RP2040.

Probably, not very useful by itself, but the ability to fire a scheduled interrupt enables us to support RTC-governed Deep Sleep and Dormant states later.

### DIFF
--- a/src/examples/rtc/rtc.go
+++ b/src/examples/rtc/rtc.go
@@ -1,0 +1,61 @@
+//go:build rp2040
+
+package main
+
+// This example demonstrates real time clock (RTC) support.
+//
+// An alarm can be set to execute user callback function once or on a schedule.
+//
+// Unfortunately, it is not possible to use time.Time to work with RTC,
+// that would introduce a circular dependency between "machine" and "time" packages.
+
+import (
+	"fmt"
+	"machine"
+	"time"
+)
+
+// For RTC to work, it must be set to some reference time first
+var rtcTimeReference = machine.RtcTime{
+	Year:  2023,
+	Month: 01,
+	Day:   19,
+	Dotw:  4,
+	Hour:  1,
+	Min:   12,
+	Sec:   45,
+}
+
+// Alarm shall fire every minute at 5 sec
+var rtcTimeAlarm = machine.RtcTime{
+	Year:  -1,
+	Month: -1,
+	Day:   -1,
+	Dotw:  -1,
+	Hour:  -1,
+	Min:   -1,
+	Sec:   5,
+}
+
+func main() {
+
+	// configure RTC and set alarm
+	machine.RTC.SetTime(rtcTimeReference)
+	machine.RTC.SetAlarm(rtcTimeAlarm, func() { println("Pekabo!") }) // the callback function executes on interrupt and shall be as quick as possible
+
+	// wait a bit to let user connect to serial console and for RTC to initialize
+	time.Sleep(3 * time.Second)
+
+	for {
+		rtcTime, err := machine.RTC.GetTime() // reading time from RTC, it shall increase 1 second each read
+		if err != nil {
+			println(err.Error())
+		}
+		printTime(rtcTime)
+		time.Sleep(time.Second)
+	}
+}
+
+func printTime(t machine.RtcTime) {
+	fmt.Printf("%4d-%02d-%02d %s %02d:%02d:%02d\r\n", t.Year, t.Month, t.Day, time.Weekday(t.Dotw).String()[:3], t.Hour, t.Min, t.Sec)
+}
--- a/src/examples/rtc/rtc.go
+++ b/src/examples/rtc/rtc.go
@@ -6,27 +6,17 @@ package main
 //
 // An alarm can be set to execute user callback function once or on a schedule.
 //
-// Unfortunately, it is not possible to use time.Time to work with RTC,
+// Unfortunately, it is not possible to use time.Time to work with RTC directly,
 // that would introduce a circular dependency between "machine" and "time" packages.
 
 import (
 	"fmt"
 	"machine"
+	"runtime"
 	"time"
 )
 
-// For RTC to work, it must be set to some reference time first
-var rtcTimeReference = machine.RtcTime{
-	Year:  2023,
-	Month: 01,
-	Day:   19,
-	Dotw:  4,
-	Hour:  1,
-	Min:   12,
-	Sec:   45,
-}
-
-// Alarm shall fire every minute at 5 sec
+// Alarm shall fire every minute at 20 sec
 var rtcTimeAlarm = machine.RtcTime{
 	Year:  -1,
 	Month: -1,
@@ -34,28 +24,52 @@ var rtcTimeAlarm = machine.RtcTime{
 	Dotw:  -1,
 	Hour:  -1,
 	Min:   -1,
-	Sec:   5,
+	Sec:   20,
 }
 
 func main() {
 
-	// configure RTC and set alarm
-	machine.RTC.SetTime(rtcTimeReference)
+	// Set system clock to some date and time.
+	t, _ := time.Parse(time.RFC3339, "2006-01-02T15:04:05Z")
+	runtime.AdjustTimeOffset(-1 * int64(time.Since(t)))
+
+	// Set RTC from system time and enable recurring alarm
+	machine.RTC.SetTime(toRtcTime(time.Now()))
 	machine.RTC.SetAlarm(rtcTimeAlarm, func() { println("Pekabo!") }) // the callback function executes on interrupt and shall be as quick as possible
 
-	// wait a bit to let user connect to serial console and for RTC to initialize
+	// Wait a bit to let user connect to serial console and for RTC to initialize
 	time.Sleep(3 * time.Second)
 
 	for {
-		rtcTime, err := machine.RTC.GetTime() // reading time from RTC, it shall increase 1 second each read
-		if err != nil {
-			println(err.Error())
-		}
-		printTime(rtcTime)
+		rtcTime, _ := machine.RTC.GetTime() // shall increase 1 second each time
+		nowRtc := toTime(rtcTime)
+		nowSys := time.Now()
+		diff := nowRtc.Sub(nowSys)
+		// Some descrepancy is expected, diff shall be constant
+		fmt.Printf("SYS: %v, RTC: %v, DIFF: %v\r\n", nowSys.Format(time.RFC3339), nowRtc.Format(time.RFC3339), diff)
 		time.Sleep(time.Second)
 	}
 }
 
-func printTime(t machine.RtcTime) {
-	fmt.Printf("%4d-%02d-%02d %s %02d:%02d:%02d\r\n", t.Year, t.Month, t.Day, time.Weekday(t.Dotw).String()[:3], t.Hour, t.Min, t.Sec)
+func toRtcTime(t time.Time) machine.RtcTime {
+	return machine.RtcTime{
+		Year:  int16(t.Year()),
+		Month: int8(t.Month()),
+		Day:   int8(t.Day()),
+		Dotw:  int8(t.Weekday()),
+		Hour:  int8(t.Hour()),
+		Min:   int8(t.Minute()),
+		Sec:   int8(t.Second()),
+	}
+}
+
+func toTime(t machine.RtcTime) time.Time {
+	return time.Date(
+		int(t.Year),
+		time.Month(t.Month),
+		int(t.Day),
+		int(t.Hour),
+		int(t.Min),
+		int(t.Sec),
+		0, time.Local)
 }
--- a/src/examples/rtc/rtc.go
+++ b/src/examples/rtc/rtc.go
@@ -33,21 +33,27 @@ func main() {
 	t, _ := time.Parse(time.RFC3339, "2006-01-02T15:04:05Z")
 	runtime.AdjustTimeOffset(-1 * int64(time.Since(t)))
 
-	// Set RTC from system time and enable recurring alarm
+	// Set RTC from system time.
+	// Some descrepancy is expected. RTC may show 1 second ahead system time.
 	machine.RTC.SetTime(toRtcTime(time.Now()))
-	machine.RTC.SetAlarm(rtcTimeAlarm, func() { println("Pekabo!") }) // the callback function executes on interrupt and shall be as quick as possible
 
-	// Wait a bit to let user connect to serial console and for RTC to initialize
-	time.Sleep(3 * time.Second)
+	// Schedule and enable recurring alarm.
+	// The callback function is executed in the context of an interrupt handler,
+	// so regular restructions for this sort of code apply: no blocking, no memory allocation, etc.
+	machine.RTC.SetAlarm(rtcTimeAlarm, func() { println("Pekabo!") })
 
 	for {
-		rtcTime, _ := machine.RTC.GetTime() // shall increase 1 second each time
+		rtcTime, _ := machine.RTC.GetTime()
 		nowRtc := toTime(rtcTime)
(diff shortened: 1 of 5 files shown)


src/machine/machine_rp2040-rtc.go
```diff
@@ -0,0 +1,256 @@
...
+	ctrl      volatile.Register32
+	irqSetup0 volatile.Register32
+	irqSetup1 volatile.Register32
+	rtc1      volatile.Register32
+	rtc0      volatile.Register32
+	intR      volatile.Register32
+	intE      volatile.Register32
+	intF      volatile.Register32
+	intS      volatile.Register32
+}
```
aykevl: Why this new type definition? What's wrong with `rp.RTC_Type`?

src/machine/machine_rp2040-rtc.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	rtc.setup0.SetBits(uint32(t.Year) << rp.RTC_SETUP_0_YEAR_Pos)
+	rtc.setup0.SetBits(uint32(t.Month) << rp.RTC_SETUP_0_MONTH_Pos)
+	rtc.setup0.SetBits(uint32(t.Day) << rp.RTC_SETUP_0_DAY_Pos)
+
+	rtc.setup1.SetBits(uint32(t.Dotw) << rp.RTC_SETUP_1_DOTW_Pos)
+	rtc.setup1.SetBits(uint32(t.Hour) << rp.RTC_SETUP_1_HOUR_Pos)
+	rtc.setup1.SetBits(uint32(t.Min) << rp.RTC_SETUP_1_MIN_Pos)
+	rtc.setup1.SetBits(uint32(t.Sec) << rp.RTC_SETUP_1_SEC_Pos)
```
aykevl: I think this is wrong: `SetBits` just sets those bits but doesn't clear it.
A better (correct and more efficient) way to do this would be by creating a new `uint32` value with the right values and then using a single `rtc.setup0.Set()` to update it.

src/machine/machine_rp2040-rtc.go
```diff
@@ -0,0 +1,256 @@
...
+
+	// Does it repeat? I.e. do we not match on any of the bits
+	rtcAlarmRepeats = t.AlarmRepeats()
+
+	// Store function pointer we can call later
+	rtcCallback = callback
+
+	// Enable the IRQ at the proc
+	interrupt.New(rp.IRQ_RTC_IRQ, rtcHandleInterrupt).Enable()
+	irqSet(rp.IRQ_RTC_IRQ, true)
```
ysoldak, line 170: Is this really needed?
I see this pattern of `interrupt.New().Enable()` followed by `irqSet()` across rp2040 related code.
Yet, it works w/o `irqSet()` call too.

src/machine/machine_rp2040-rtc.go
```diff
@@ -0,0 +1,256 @@
+//go:build rp2040
```
kenbell, line 1: this file should be machine_rp2040_rtc.go?  (we use underscore in other files)

src/machine/machine_rp2040-rtc.go
```diff
@@ -0,0 +1,256 @@
...
+	// Check the freq is not too big to divide
+	if rtcFreq > rp.RTC_CLKDIV_M1_CLKDIV_M1_Msk {
+		panic("rtc freq is too big to divide")
+	}
+
+	// Write divide value
+	rtc.clkDivM1.Set(rtcFreq)
+}
+
+func (rtc *rtcType) SetTime(t RtcTime) error {
```
kenbell, line 57: Should this accept time.Time?

src/machine/machine_rp2040-rtc.go
```diff
@@ -0,0 +1,256 @@
...
+
+	// Enable RTC and wait for it to be running
+	rtc.ctrl.SetBits(rp.RTC_CTRL_RTC_ENABLE)
+	for !rtc.running() {
+	}
+
+	return nil
+}
+
+func (rtc *rtcType) GetTime() (t RtcTime, err error) {
```
kenbell, line 93: Should this return time.Time?

src/machine/machine_rp2040-rtc.go
```diff
@@ -0,0 +1,256 @@
...
+		Year:  int16((rtc_1 & rp.RTC_RTC_1_YEAR_Msk) >> rp.RTC_RTC_1_YEAR_Pos),
+		Month: int8((rtc_1 & rp.RTC_RTC_1_MONTH_Msk) >> rp.RTC_RTC_1_MONTH_Pos),
+		Day:   int8((rtc_1 & rp.RTC_RTC_1_DAY_Msk) >> rp.RTC_RTC_1_DAY_Pos),
+	}
+
+	return t, nil
+}
+
+// void rtc_set_alarm(datetime_t *t, rtc_callback_t user_callback) {
+func (rtc *rtcType) SetAlarm(t RtcTime, callback func()) {
```
kenbell, line 119: Should be documented that the callback will be in the context of an interrupt handler - ie no blocking, no memory allocation, etc?

src/machine/machine_rp2040-rtc.go
```diff
@@ -0,0 +1,256 @@
...
+
+	// Call user callback function
+	if rtcCallback != nil {
+		rtcCallback()
+	}
+}
+
+// ---
+
+type RtcTime struct {
```
kenbell, line 215: Should this only be used for setting alarms?   For get/set time, use time.Time instead?

Also, how generic is this - should it be in a top-level `rtc.go`, or is it rp2040 specific?

src/machine/machine_rp2040.go
```diff
@@ -76,6 +76,8 @@ func machineInit() {
 
 	clocks.init()
 
+	RTC.init()
```
kenbell: Best not to initialize unconditionally like this - in general, enable peripherals only if the app will use them

src/examples/rtc/rtc.go
```diff
@@ -0,0 +1,61 @@
...
+	Year:  -1,
+	Month: -1,
+	Day:   -1,
+	Dotw:  -1,
+	Hour:  -1,
+	Min:   -1,
+	Sec:   5,
+}
+
+func main() {
```
kenbell, line 30: Might be worth showing how to use `runtime.AdjustTimeOffset` to set the go standard `time` package to have the correct time from an RTC?

src/machine/machine_rp2040-rtc.go
ysoldak: yes, true, will fix, thanks!

src/machine/machine_rp2040-rtc.go
ysoldak: I'd _love_ to. Unfortunately, that makes a cycle:
```
package github.com/tinygo-org/tinygo/src/examples/rtc
        imports machine
        imports time
        imports runtime
        imports machine: import cycle not allowed
```

src/machine/machine_rp2040-rtc.go
ysoldak: see above

src/machine/machine_rp2040-rtc.go
ysoldak: yes, thank you, will add.

aykevl (commented):
I... really don't like that chip vendors think they can implement the Gregorian calendar in hardware. Because they can't:

1. It does not deal with timezones or daylight saving time at all.
2. As admitted in the datasheet, it doesn't correctly deal with the leap century. Admittedly that's rather rare (we're not due for another 77 years) but it shows that this kind of thing is difficult.
3. It does not verify the day of the week.
4. This is really something that only the UI should do. Most time protocols (like NTP) only deal with Unix time, which is a plain and simple number. It only adds complexity and bugs between systems that deal with Unix time and systems that attempt to use the Gregorian calendar.

Anyway, that's my view on this and kind of besides the point of this PR.

But regarding this specific PR: it implements a new API for the machine package which I don't want to add without getting it right. So I suggest making a PR to tinygo-site, see https://github.com/tinygo-org/tinygo-site/pull/317 for example (content/docs/reference/machine.md is the important part, the rest is optional). That way, it's easier to see what it is doing. And we'd need to document the API anyway so better do it right from the start.

src/machine/machine_rp2040-rtc.go
ysoldak: Can't use `time.Time`, unfortunately. See above.

This implementation is RP2040 specific.
We don't have any API for RTC as of yet (nRF family uses it, but does not expose, probably others too).
So the idea was we can try and make it for RP2040 first and generalise once we have at least two families exposing RTC?

src/machine/machine_rp2040.go
ysoldak: OK, will add a check and init in `SetTime()` then.

src/machine/machine_rp2040-rtc.go
ysoldak: Great, will do!

src/machine/machine_rp2040-rtc.go
ysoldak: Nothing, it's just me not knowing it exists :D
Will fix.

src/examples/rtc/rtc.go
ysoldak: This RTC does not keep state on power down or reboot, unfortunately.
So one has to source the correct time somewhere first, like from WiFi Access Point, as I do in some of my projects.

When one just wants to schedule a delayed interrupt (say within minutes, hours), the reference time does not really matter. It can be 1970-01-01, really.

ysoldak:
Example output
```
SYS: 2006-01-02T15:04:08Z, RTC: 2006-01-02T15:04:09Z, DIFF: 999.148ms
SYS: 2006-01-02T15:04:09Z, RTC: 2006-01-02T15:04:10Z, DIFF: 997.986ms
SYS: 2006-01-02T15:04:10Z, RTC: 2006-01-02T15:04:11Z, DIFF: 997.155ms
SYS: 2006-01-02T15:04:11Z, RTC: 2006-01-02T15:04:12Z, DIFF: 996.361ms
SYS: 2006-01-02T15:04:12Z, RTC: 2006-01-02T15:04:13Z, DIFF: 995.58ms
SYS: 2006-01-02T15:04:13Z, RTC: 2006-01-02T15:04:14Z, DIFF: 994.825ms
SYS: 2006-01-02T15:04:14Z, RTC: 2006-01-02T15:04:15Z, DIFF: 994.061ms
SYS: 2006-01-02T15:04:15Z, RTC: 2006-01-02T15:04:16Z, DIFF: 993.29ms
SYS: 2006-01-02T15:04:16Z, RTC: 2006-01-02T15:04:17Z, DIFF: 992.508ms
SYS: 2006-01-02T15:04:17Z, RTC: 2006-01-02T15:04:18Z, DIFF: 991.733ms
SYS: 2006-01-02T15:04:18Z, RTC: 2006-01-02T15:04:19Z, DIFF: 990.949ms
Pekabo!
SYS: 2006-01-02T15:04:19Z, RTC: 2006-01-02T15:04:20Z, DIFF: 990.173ms
SYS: 2006-01-02T15:04:20Z, RTC: 2006-01-02T15:04:21Z, DIFF: 989.389ms
SYS: 2006-01-02T15:04:21Z, RTC: 2006-01-02T15:04:22Z, DIFF: 988.624ms
SYS: 2006-01-02T15:04:22Z, RTC: 2006-01-02T15:04:23Z, DIFF: 987.844ms
```

~Interesting to see the diff decrease. (approx 800us per second.)~
~This means RTC ticks faster than the system clock.~
~Which of them is correct is hard to tell though.~

This is all due to the fact RTC returns seconds precision and the loop iteration takes slightly more than a second (1 second and ~800µs to be precise).

_Both clocks have exactly same value internally, they run from the same source (XOSC)._

The accumulated error is immediately visible in the value of the system clock (µs precision) and is not visible until it adds up to more than a second in RTC case.

The moment error accumulates to a whole second, the output _jumps_ two seconds for both SYS and RTC.

Confirmed by letting the (slightly modified) example run for 20 minutes:
```
...
SYS: 2006-01-02T15:23:33Z, RTC: 2006-01-02T15:23:33Z, DIFF: -998.923ms, DIFF':846µs
SYS: 2006-01-02T15:23:34Z, RTC: 2006-01-02T15:23:34Z, DIFF: -999.788ms, DIFF':865µs
SYS: 2006-01-02T15:23:36Z, RTC: 2006-01-02T15:23:35Z, DIFF: -1.000628s, DIFF':840µs
SYS: 2006-01-02T15:23:37Z, RTC: 2006-01-02T15:23:37Z, DIFF: -1.507ms, DIFF':-999.121ms
SYS: 2006-01-02T15:23:38Z, RTC: 2006-01-02T15:23:38Z, DIFF: -2.366ms, DIFF':859µs
SYS: 2006-01-02T15:23:39Z, RTC: 2006-01-02T15:23:39Z, DIFF: -3.187ms, DIFF':821µs
...
```

I shall probably modify the example and remove the confusing DIFF output.

src/examples/rtc/rtc.go
ysoldak: Implemented in example. Thanks for the idea!

src/machine/machine_rp2040-rtc.go
ysoldak: @aykevl what would be your understanding?
If this `irqSet` is indeed not needed in such cases, I could make a PR that cleans this in RP2040 machine.

src/machine/machine_rp2040-rtc.go
soypat: I'd go for a `SetDate` function that follows the [stdlib format of creating a new date](https://pkg.go.dev/time#Date) (excluding the loc).
This would probably make it more natural to use than with a machine specific type.

soypat: Check this out, it might be a way of solving this cyclical dependency problem given that we all agree it's time the machine package had it's own Time type.

`machine/time.go`
```go
type Time struct {
 // Same fields as time.Time
}

func (rtc *RTC) SetTime[T ~Time](t T) error {
   // do the setting.
}
```

`time/time.go`
```go
// replace contents of time.Time type.
type Time machine.Time
```
I have no idea of the far reaching consequences of this change.

soypat:
If possible I would not expose a `RtcTime` type. That ***should*** be a `time.Time`. I've suggested having a `machine.Time` type that basically implements the whole functionality of `time.Time`, to then be imported by `time` package.

If we do not create a `machine.Time` type I believe we should probably give the `RTC` type similar methods to `time.Time`, such as `Date` (copy [Date](https://pkg.go.dev/time#Time.Date) signature) and `Clock` (copies [Clock](https://pkg.go.dev/time#Time.Clock) signature) and `Nanosecond` (copies [Nanosecond](https://pkg.go.dev/time#Time.Nanosecond) signature).

Same for `SetAlarm`, which could be `SetAlarmDate` and take same parameters as `time.Date` if we choose to drop the `RtcTime` type.

Last Point: If you we do choose to keep `RtcTime` I'd encourage it be renamed to be `RTCTime`, in line with [Go convention of acronyms according to Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

src/machine/machine_rp2040-rtc.go
ysoldak: I've actually made an alternative PR #3403 that sets RTC interrupt with a delay instead of actual time. Probably, that's better than such tricks.

src/machine/machine_rp2040-rtc.go
aykevl: Yeah it looks to me like `irqSet` is not needed and was just copied over from C code. I think it can be removed.

PR #3399, opened 2023-01-19
Allows the emulator to expand {tmpDir}

codefromthecrypt:
This allows you to expand {tmpDir} in the json "emulator" field, and uses it in wasmtime instead of custom TMPDIR mapping logic.

Before, we had custom logic for wasmtime to create a separate tmpDir when running go tests. This overwrite the TMPDIR variable when running, after making a mount point. A simpler way to accomplish the end goal of writing temp files is to use wasmtime's map-dir instead. When code is compiled to wasm with the wasi target, tempDir is always /tmp, so we don't need to add variables (since we know what it is). Further, the test code is the same between normal go and run through wasmtime. So, we don't need to make a separate temp dir first, and avoiding that reduces logic, as well makes it easier to swap out the emulator (for wazero which has no depedencies). To map the correct directory, this introduces a {tmpDir} token whose value is the host-specific value taken from `os.TempDir()`.

The motivation I have for this isn't so much to clean up the wasmtime code, but allow wazero to execute the same tests. After this change, the only thing needed to pass tests is to change the emulator, due to differences in how wazero deals with relative lookups (they aren't restricted by default, so there's not a huge amount of custom logic needed).

In other words, installing wazero from main, `make tinygo-test-wasi` works with no other changes except this PR and patching `targets/wasi.json`.
```json
	"emulator":      "wazero run -mount=.:/ -mount={tmpDir}:/tmp {}",
```

On that note, if there's a way to override the emulator via arg or env, this would be even better, but in any case patching json is fine.

### DIFF
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -544,6 +544,8 @@ func (c *Config) Emulator(format, binary string) ([]string, error) {
 	var emulator []string
 	for _, s := range parts {
 		s = strings.ReplaceAll(s, "{root}", goenv.Get("TINYGOROOT"))
+		// Allow replacement of what's usually /tmp except notably Windows.
+		s = strings.ReplaceAll(s, "{tmpDir}", os.TempDir())
 		s = strings.ReplaceAll(s, "{"+format+"}", binary)
 		emulator = append(emulator, s)
 	}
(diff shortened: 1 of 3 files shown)


dgryski:
Being able to reduce the amount of `wasmtime`-specific code in tinygo is a win.

dgryski (approved):
LGTM

deadprogram:
Thank you @codefromthecrypt for the code and to @dgryski for reviewing. Now merging.

codefromthecrypt:
thanks for the merge!

PR #3401, opened 2023-01-20
add a new target: zkvm

SchmErik:
This is a first attempt to add the zkvm as a new target. At this point, all files in testdata compile except for those involving goroutines and channels. Some tests can run on r0vm but most fail. The failures require further investigation and it's possible that this code being added contains mistakes...


SchmErik:
Sorry about this, targeted the wrong repo. Closing

PR #3402, opened 2023-01-21
example: adjust time offset

ysoldak:
As discussed in #3398 we don't really have an example on how to use `runtime.AdjustTimeOffset()` functionality.
So here it is.

### DIFF
--- a/src/examples/time-offset/time-offset.go
+++ b/src/examples/time-offset/time-offset.go
@@ -1,0 +1,32 @@
+package main
+
+// This example demonstrates how to set the system time.
+//
+// Usually, boards don't keep time on power cycles and restarts, it resets to Unix epoch.
+// The system time can be set by calling runtime.AdjustTimeOffset().
+//
+// Possible time sources: an external RTC clock or network (WiFi access point or NTP server)
+
+import (
+	"fmt"
+	"runtime"
+	"time"
+)
+
+const myTime = "2006-01-02T15:04:05Z" // this is an example time you source somewhere
+
+func main() {
+
+	// measure how many nanoseconds the internal clock is behind
+	timeOfMeasurement := time.Now()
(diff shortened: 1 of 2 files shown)


src/examples/time-offset/time-offset.go
```diff
@@ -0,0 +1,27 @@
...
+	"runtime"
+	"time"
+)
+
+const myTime = "2006-01-02T15:04:05Z" // this is an example time you source somewhere
+
+func main() {
+
+	t, _ := time.Parse(time.RFC3339, myTime)
+	runtime.AdjustTimeOffset(-1 * int64(time.Since(t)))
```
aykevl: I think this may better explain what is happening (I didn't test the code though):

```suggestion
	// measure how many nanoseconds the internal clock is behind
	timeOfMeasurement := time.Now()
	actualTime, _ := time.Parse(time.RFC3339, myTime)
	offset := actualTime.Sub(timeOfMeasurement)

	// adjust internal clock by adding the offset to the internal clock
	runtime.AdjustTimeOffset(int64(offset)
```

aykevl (commented):
Looks reasonable to me.

src/examples/time-offset/time-offset.go
ysoldak: The original idea with this example was to give people a simple recipe to set custom time, not necessarily educational.

The API is a bit confusing, to be fair. It would be much simpler if one could just `time.Parse().UnixNano()` and set that value and the internals figure out how exactly to do that.

But if you insist on educational value of the example, sure we can expand it and probably also write some comments how it looks internally?
Shall we?

ysoldak:
@aykevl PTAL?

ysoldak:
@deadprogram @aykevl do we want this?

deadprogram (approved):

deadprogram:
Now merging, thanks for the improvement @ysoldak

PR #3403, opened 2023-01-21
rp2040: rtc delayed interrupt

ysoldak:
Drastically reduced API surface compared to #3398 
The functionality is limited to setting a delayed interrupt.

`machine.RTC.SetInterrupt()` API was modelled on Pin and CAN interrupt functionality.

Delay is in nanoseconds to simplify conversion from `time.Time` in user code since we can't consume `time.Time` directly due to dependency cycle.

To avoid dealing with leap years and different count of days in a month, delay's upper boundary is set to 1 day and minimal value is 1 second. Boundaries can be revised later. Boundaries are verified in runtime. 

This PR is a ground work for using RTC interrupts to wake up from SLEEP and DORMANT states later.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -465,6 +465,8 @@ smoketest:
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=pca10040            examples/pininterrupt
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=nano-rp2040         examples/rtcinterrupt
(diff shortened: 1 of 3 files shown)


soypat (approved):
LGTM. I believe this is a good starting point to let users start using the `rtc` type with minimal exposed API. This way we can learn more on the needs of the community.

deadprogram:
Thank you for working on this @ysoldak and to @soypat for review. Now merging.

PR #3406, opened 2023-01-22
machine/i2c: add interface check and implementation where missing for SetBaudRate()

deadprogram:
This PR updates machine/i2c by adding an interface check to verify the completeness of the `i2cController` for all required functions, and add implementation where missing for `SetBaudRate()`.

Currently set to draft to work on implementations for the various placeholders.

### DIFF
--- a/src/machine/i2c.go
+++ b/src/machine/i2c.go
@@ -1,4 +1,4 @@
-//go:build atmega || nrf || sam || stm32 || fe310 || k210 || rp2040
+//go:build atmega || nrf || sam || stm32 || fe310 || k210 || rp2040 || mimxrt1062
 
 package machine
 
--- a/src/machine/i2c.go
+++ b/src/machine/i2c.go
@@ -6,6 +6,17 @@ import (
 	"errors"
 )
 
+// If you are getting a compile error on this line please check to see you've
+// correctly implemented the methods on the I2C type. They must match
+// the i2cController interface method signatures type to type perfectly.
+// If not implementing the I2C type please remove your target from the build tags
+// at the top of this file.
+var _ interface { // 2
+	Configure(config I2CConfig) error
+	Tx(addr uint16, w, r []byte) error
+	SetBaudRate(br uint32) error
+} = (*I2C)(nil)
+
 // TWI_FREQ is the I2C bus speed. Normally either 100 kHz, or 400 kHz for high-speed bus.
 //
 // Deprecated: use 100 * machine.KHz or 400 * machine.KHz instead.
--- a/src/machine/i2c.go
+++ b/src/machine/i2c.go
@@ -36,6 +36,7 @@ var (
 	errI2CBusError           = errors.New("I2C bus error")
 	errI2COverflow           = errors.New("I2C receive buffer overflow")
 	errI2COverread           = errors.New("I2C transmit buffer overflow")
+	errI2CNotImplemented     = errors.New("I2C operation not yet implemented")
 )
 
 // I2CTargetEvent reflects events on the I2C bus
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -732,12 +732,13 @@ func (i2c *I2C) Configure(config I2CConfig) error {
 	return nil
 }
 
-// SetBaudRate sets the communication speed for the I2C.
-func (i2c *I2C) SetBaudRate(br uint32) {
+// SetBaudRate sets the communication speed for I2C.
+func (i2c *I2C) SetBaudRate(br uint32) error {
 	// Synchronous arithmetic baudrate, via Arduino SAMD implementation:
 	// SystemCoreClock / ( 2 * baudrate) - 5 - (((SystemCoreClock / 1000000) * WIRE_RISE_TIME_NANOSECONDS) / (2 * 1000));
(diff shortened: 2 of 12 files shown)


aykevl:
What's the status of this PR?

deadprogram:
The `SetBaudRate()` changes are complete for all machines except:

- `machine_mimxrt1062_i2c.go`
- `machine_stm32_i2c_reva.go`
- `machine_stm32_i2c_revb.go`

UPDATE: stm32_i2c_revb.go "Currently, only 100KHz mode is supported". So nothing to be done on that.

On those machine types, I have put placeholder functions. Perhaps someone else can lend a hand on those on a future PR? I think this PR itself is ready for review.

src/machine/i2c.go
```diff
@@ -6,6 +6,15 @@ import (
 	"errors"
 )
 
+// i2cController interface defines what is required to provide a i2c controller.
+// It is primarily to provide compile time checking when adding a new i2c
+// implementation.
+type i2cController interface {
```
soypat: I'd rather we not implement compilation check types in the global namespace if we can avoid it. To do this we can replace this definition with the following in this file to ensure all future implementations fulfill the interface so that we don't have to add the check on each individual interface facilitating maintenance in the long run. We also keep the global namespace free of clutter going forward:
```go
// If you are getting a compile error on this line please check to see you've
// correctly implemented the methods on the I2C type. They must match
// the i2cController interface method signatures type to type perfectly.
// If not implementing the I2C type please remove your target from the build tags at the top of this file.
var _ interface{  // 2
	Configure(config I2CConfig) error
	Tx(addr uint16, w, r []byte) error
	SetBaudRate(br uint32) error
} = (*I2C)(nil)
```

Or 

```go
func _() {
    type i2cController interface {
	    Configure(config I2CConfig) error
	    Tx(addr uint16, w, r []byte) error
	    SetBaudRate(br uint32) error
    }
    // If you are getting a compile error on this line please check to see you've
    // correctly implemented the methods on the I2C type. They must match
    // the i2cController interface method signatures type to type perfectly.
    // If not implementing the I2C type please remove your target from the build tags at the top of this file.
    var _ i2cController = (*I2C)(nil)
}
```

src/machine/machine_atsamd21.go
```diff
@@ -644,6 +644,9 @@ func (uart *UART) handleInterrupt(interrupt.Interrupt) {
 	uart.Bus.INTFLAG.SetBits(sam.SERCOM_USART_INTFLAG_RXC)
 }
 
+// check for ensuring we fulfill interface
+var _ i2cController = (*I2C)(nil)
```
soypat: This seems like a lot of work to add to each file. Can we just have it in the same file as `i2cController`?

src/machine/i2c.go
soypat: See https://github.com/tinygo-org/tinygo/issues/2334#issuecomment-1238872628

src/machine/machine_atsamd21.go
deadprogram: Seems yes. Done.

src/machine/i2c.go
deadprogram: Good idea, done.

deadprogram:
@soypat changes made thanks for the suggestions. Please review again.

src/machine/machine_mimxrt1062_i2c.go
```diff
@@ -171,6 +171,12 @@ func (i2c *I2C) Configure(config I2CConfig) {
 	i2c.reset(freq)
 }
 
+// SetBaudRate sets the communication speed for I2C.
+func (i2c I2C) SetBaudRate(br uint32) error {
+	// TODO: implement
+	return nil
```
soypat: Hmmmm... shouldn this return an error or panic until it is implemented?

soypat (approved):
LGTM.  Noted that unimplemented methods don't raise an error, but otherwise looks great. Thank you for the change Ron!

src/machine/machine_mimxrt1062_i2c.go
deadprogram: Good idea. Done.

deadprogram:
OK had to make a few more changes for the `mimxrt1062` to make sure it was actually conforming to the `i2c` interface.

deadprogram:
Thanks for review and feedback for improvements @soypat now merged.

src/machine/machine_mimxrt1062_i2c.go
```diff
@@ -212,13 +207,13 @@ func (i2c I2C) Tx(addr uint16, w, r []byte) error {
...
 
-// WriteRegister transmits first the register and then the data to the
+// WriteRegisterEx transmits first the register and then the data to the
 // peripheral device.
 //
 // Many I2C-compatible devices are organized in terms of registers. This method
 // is a shortcut to easily write to such registers. Also, it only works for
 // devices with 7-bit addresses, which is the vast majority.
-func (i2c I2C) WriteRegister(address uint8, register uint8, data []byte) error {
+func (i2c I2C) WriteRegisterEx(address uint8, register uint8, data []byte) error {
```
aykevl, line 216: Why `WriteRegisterEx` and `ReadRegisterEx`?
If it's to avoid a duplicate method, I think this is a better solution: https://github.com/tinygo-org/tinygo/pull/4130

PR #3409, opened 2023-01-23
stm32wl: Fix for incomplete RNG initialisation

ofauchon:
STM32WL's RNG was not properly initialized.
RNG Timeout errors were appearing after 1-2 successful requests to rand.Read() function.

This PR completes initialization of RNG registers and add checks for CECS (Clock) and SECS(Seed) status registers

Steps to reproduce the problem from dev branch (on LoraE5 dev board)

```
package main

import (
	"time"
	"crypto/rand"
)

func main() {
	println("*** stm32wl RNG test")

	var randomBytes [2]byte
	for {
		_, err := rand.Read(randomBytes[:])
		if err !=nil{
			println("error:",err)
		}
		println("randomBytes:", randomBytes[0], randomBytes[1])
		time.Sleep(time.Millisecond * 25)
	}
}


$ tinygo flash  -target=lorae5
$  picocom -b 115200 /dev/ttyUSB0

*** stm32wl RNG test
randomBytes: 147 36
randomBytes: 100 21
error: machine: RNG Timeout
randomBytes: 100 21
error: machine: RNG Timeout
randomBytes: 100 21
error: machine: RNG Timeout
randomBytes: 100 21
```

With the fix, I could call rand.Read() hundreds of time without errors.

### DIFF
--- a/src/machine/machine_stm32_rng.go
+++ b/src/machine/machine_stm32_rng.go
@@ -15,6 +15,13 @@ func GetRNG() (uint32, error) {
 		rngInitDone = true
 	}
 
+	if stm32.RNG.SR.HasBits(stm32.RNG_SR_CECS) {
+		return 0, ErrClockRNG
+	}
+	if stm32.RNG.SR.HasBits(stm32.RNG_SR_SECS) {
+		return 0, ErrSeedRNG
+	}
+
 	cnt := RNG_MAX_READ_RETRIES
 	for !stm32.RNG.SR.HasBits(stm32.RNG_SR_DRDY) {
 		cnt--
--- a/src/machine/machine.go
+++ b/src/machine/machine.go
@@ -4,6 +4,8 @@ import "errors"
 
 var (
 	ErrTimeoutRNG         = errors.New("machine: RNG Timeout")
+	ErrClockRNG           = errors.New("machine: RNG Clock Error")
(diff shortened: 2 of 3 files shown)


src/machine/machine_stm32_rng.go
```diff
@@ -18,6 +18,12 @@ func GetRNG() (uint32, error) {
 	cnt := RNG_MAX_READ_RETRIES
 	for !stm32.RNG.SR.HasBits(stm32.RNG_SR_DRDY) {
 		cnt--
+		if stm32.RNG.SR.HasBits(stm32.RNG_SR_CECS) {
```
deadprogram: Should these two checks be before the loop?

ofauchon: Reference manual :

CECS => "When the clock error detection is enabled (CED = 0) and if the RNG clock frequency is too
low, the RNG sets to “1” both the CEIS and CECS bits to indicate that a clock error
occurred."  

SECS => "When a noise source (or seed) error occurs, the RNG stops generating random numbers
and sets to “1” both SEIS and SECS bits to indicate that a seed error occurred. If a value is
available in the RNG_DR register, it must not be used as it may not have enough entropy."

... So it seems to be even better to do the checks before the loop. 

Thanks for the suggestion.

deadprogram:
Thanks for the fixes @ofauchon now merging.

PR #3410, opened 2023-01-24
Return unique pointer when size 0 passed to malloc

lann:
PR #3303 addressed panics when malloc'ing with size 0 by returning nil. While this is permitted by the C standard, some Wasm environments expect the behavior to match that of glibc/dlmalloc, which is to always return a unique pointer.

### DIFF
--- a/src/runtime/arch_tinygowasm_malloc.go
+++ b/src/runtime/arch_tinygowasm_malloc.go
@@ -13,7 +13,7 @@ var allocs = make(map[uintptr][]byte)
 //export malloc
 func libc_malloc(size uintptr) unsafe.Pointer {
 	if size == 0 {
-		return nil
+		size = 1 // Match dlmalloc's behavior to allocate a "minimum-sized chunk"
 	}
 	buf := make([]byte, size)
 	ptr := unsafe.Pointer(&buf[0])
@@ -42,8 +42,7 @@ func libc_calloc(nmemb, size uintptr) unsafe.Pointer {
 //export realloc
 func libc_realloc(oldPtr unsafe.Pointer, size uintptr) unsafe.Pointer {
 	if size == 0 {
-		libc_free(oldPtr)
-		return nil
+		size = 1 // Match dlmalloc's behavior to allocate a "minimum-sized chunk"
 	}
 
 	// It's hard to optimize this to expand the current buffer with our GC, but
--- a/tests/runtime_wasi/malloc_test.go
+++ b/tests/runtime_wasi/malloc_test.go
@@ -127,19 +127,19 @@ func TestMallocFree(t *testing.T) {
 
 func TestMallocEmpty(t *testing.T) {
 	ptr := libc_malloc(0)
-	if ptr != nil {
-		t.Errorf("expected nil pointer, got %p", ptr)
+	if ptr == nil {
+		t.Error("expected pointer but was nil")
 	}
 }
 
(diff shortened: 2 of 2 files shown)


dgryski:
I don't want to commit fixes that work around buggy wasm implementations. Should they have bugs filed against them too?

lann:
I'm not sure that it's buggy implementations per se. I haven't been able to find documentation on what wasi expects out of malloc, but `wasi-libc` embed dlmalloc so it is at least implicitly part of the API.

bgould:
Seems like user code should not be relying on any specific behavior in this case.  

"The malloc() function allocates size bytes and returns a pointer to the allocated memory. The memory is not initialized. If size is 0, then malloc() returns either NULL, or a unique pointer value that can later be successfully passed to free()." (https://linux.die.net/man/3/malloc) ... unless WASI specifies this differently, I agree with @dgryski that relying on this behavior is a bug in the calling code ... at the very least they are relying on undefined behavior.

If we wanted to define a specific behavior for this ... maybe it could be to return _either nil or a pointer randomly_, similar to how standard Go randomizes map iteration to keep users from relying on key ordering (j/k)

If we need to support users who have to rely on one behavior or the other ... then IMO that should be controlled explicitly via build tags, because tomorrow someone will come along and insist that reverting this behavior is necessary because of some reason.

anuraaga:
Agree that there shouldn't really be a need for TinyGo's malloc to match the behavior of any other particular implementation. Note that there is a `custommalloc` build tag that can also be used to swap in something else entirely (I use `mimalloc`).

As far as I can tell, the issue is mostly wit-bindgen preventing NULL in generated code like this one

https://github.com/fermyon/spin/blob/722107325f01a58e5e04b09df3e70f73409c58bf/sdk/go/http/wasi-outbound-http.c#L11-L13

As wit-bindgen's GitHub repo states, it's a relatively young project and this just seems like a bug that could be fixed there. AFAIK, wit is not particularly specific to WASI so shouldn't be making assumptions about wasi-libc, and this is probably just accidental.

lann:
I went looking for evidence that more tools expect the dlmalloc behavior and I found that `wasi-libc` actually supports a second embedded allocator emmalloc which _does_ return NULL on `malloc(0)`, so I think the assertion that consumers _shouldn't_ rely on this alloc behavior is correct. I will look into fixing this in `wit-bindgen`.

The only other thing I'll point out is that this is _sort of_ a behavior regression since 0.25, so there may be other projects relying on the dlmalloc behavior:
- 0.25 delegated to the normal `gc_*` allocators, which afaict returns a unique pointer on `malloc(0)` (and indeed worked for us, at least with `gc_leaking`)
- 0.26 panics on `malloc(0)`; I don't know how many people ran into that issue but I doubt many would have tested with #3303 yet

If that doesn't seem like a big concern I'm happy to close this PR.

aykevl:
Agree with others, this is a bug in the users of `malloc` and therefore should be fixed there. Only if it really isn't fixable for some reason then it's something to consider working around in TinyGo.

> 0.26 panics on `malloc(0)`; I don't know how many people ran into that issue but I doubt many would have tested with

Ah that may be it. But this is fixed in the dev branch now, so the next release won't have this bug. (It was fixed quite some time ago, so doing a release now is perhaps not such a bad idea).

lann:
Fixed in wit-bindgen: https://github.com/bytecodealliance/wit-bindgen/pull/465

PR #3411, opened 2023-01-25
target/lgt92: fix openocd target and interface

ofauchon:
Some fixes for Dragino LGT92 target board: 

- Target MCU was not properly defined for LGT92 target (should be stm32l0)
- Change openocd interface from stlink-v2 to stlink (interface/stlink-v2.cfg is deprecated)

Specifications : https://www.dragino.com/products/lora-lorawan-end-node/item/142-lgt-92.html
Changes were tested on LGT92 device.


deadprogram:
Thank you for the correction @ofauchon now merging.

PR #3412, opened 2023-01-26
runtime/stm32wlx: change init order so clock speeds are correct

deadprogram:
This PR modifies the runtime init order for the `stm32wlx` so the clock speeds are set before peripherals started.


ofauchon:
UART still working fine at baudrates 9600 / 115200 on LoraE5 development board. 
LGTM

ofauchon (approved):
Tested on LoarE5 dev board. 
uart still works as expected

deadprogram:
Thanks for testing @ofauchon now merging.

PR #3414, opened 2023-01-27
Support for the Espressif ESP32-C3-DevKit-RUST-1 development board

inindev:
Signed-off-by: John Clark <inindev@gmail.com>

### DIFF
--- a/src/machine/board_esp32-c3-devkit-rust-1.go
+++ b/src/machine/board_esp32-c3-devkit-rust-1.go
@@ -1,0 +1,81 @@
+//go:build esp32_c3_devkit_rust_1
+
+// This file contains the pin mappings for the Espressif ESP32-C3 Development Board for Rust.
+//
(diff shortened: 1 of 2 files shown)


deadprogram:
Hello @inindev thank you for another board contribution.

Just wondering if the name might be a little too long.

Perhaps just `esp32-c3-devkit-rust-1` would be enough? :thinking:

inindev:
I wondered the same thing, or if I should collapse a couple of dashes such as esp32c3-devkit-rust1 but thought it would break convention to remove the manufacturer name from the front.  I will make your suggested change.

inindev:
I also thought I should update the "Supported boards/targets" of the repo README.md.  Would it be best to put a PR up for this file here or update a different location?

inindev:
renaming complete

deadprogram:
Thank you @inindev now merging.

PR #3416, opened 2023-01-28
avr: use precise GC by default

aykevl:
AVR is especially sensitive to false positives in the GC for a few reasons:

  * Pointer values are only two bytes, instead of four on most other microcontrollers.
  * Pointers are not aligned, so any two-byte value in an object coud be a pointer.
  * Memory typically starts at a low address, which makes it much more likely to clash with a regular integer value (which tend to be small values).

Therefore, use the precise GC (instead of the conservative GC) by default. This increases binary size by around 350 bytes, but I think it's a good tradeoff in most cases: it avoids some runtime errors and generally RAM tends to be a lot more scarce than flash on AVR chips. If this is too much, `-gc=conservative` or `-gc=leaking` can be used instead.

@niaow

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -210,10 +210,6 @@ func runPlatTests(options compileopts.Options, tests []string, t *testing.T) {
 				// limited amount of memory.
 				continue
 
-			case "gc.go":
-				// Does not pass due to high mark false positive rate.
-				continue
-
 			case "json.go", "stdlib.go", "testing.go":
 				// Too big for AVR. Doesn't fit in flash/RAM.
 				continue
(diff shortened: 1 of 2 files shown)


dgryski (approved):
LGTM assuming CI flake fixes itself.

dgryski:
Not a flake; actual CI failure trying to run the gc.go tests on avr: https://github.com/tinygo-org/tinygo/actions/runs/4032796575/jobs/6933624327#step:16:1428

aykevl:
That's odd, it worked for me locally. I'll need to debug this.

deadprogram:
@aykevl perhaps you can rebase this against the latest `dev` to see if it still has same issue?

aykevl:
Rebased, but when I tested it, it hangs.

dgryski:
Does the precise gc still hang on avr?

PR #3418, opened 2023-01-29
windows: add ARM64 support

aykevl:
This was actually surprising once I got TinyGo to build on Windows 11 ARM64. All the changes are exactly what you'd expect for a new architecture, there was no special weirdness just for arm64.

Actually getting TinyGo to build was kind of involved though. The very short summary is: install arm64 versions of some pieces of software (like golang, cmake) instead of installing them though choco. In particular, use the [llvm-mingw](https://github.com/mstorsjo/llvm-mingw/releases) toolchain instead of using standard mingw.

---

The main reason I'm doing this is because my fastest Windows-capable system has an arm64 chip, and therefore it makes testing on Windows a lot easier for me (even though it's somewhat of a nice feature).

### DIFF
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -304,10 +304,19 @@ func defaultTarget(goos, goarch, triple string) (*TargetSpec, error) {
 		// normally present in Go (without explicitly opting in).
 		// For more discussion:
 		// https://groups.google.com/g/Golang-nuts/c/Jd9tlNc6jUE/m/Zo-7zIP_m3MJ?pli=1
+		switch goarch {
+		case "amd64":
+			spec.LDFlags = append(spec.LDFlags,
+				"-m", "i386pep",
+				"--image-base", "0x400000",
+			)
+		case "arm64":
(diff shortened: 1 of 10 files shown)


compileopts/target.go
```diff
@@ -304,10 +304,18 @@ func defaultTarget(goos, goarch, triple string) (*TargetSpec, error) {
 		// normally present in Go (without explicitly opting in).
 		// For more discussion:
 		// https://groups.google.com/g/Golang-nuts/c/Jd9tlNc6jUE/m/Zo-7zIP_m3MJ?pli=1
+		if goarch == "amd64" {
```
deadprogram: Consider 

```go
switch goarch {
   "amd64": ...
}
```

here. No?

aykevl: Good idea, fixed.

deadprogram (approved):
Seems good to me. I do not have any Windows ARM64 machines to test on myself, but since you do @aykevl I will merge. Seems like a very cool feature!

PR #3419, opened 2023-01-30
Allow custom-gc SetFinalizer and clarify KeepAlive

anuraaga:
In #3336, `SetFinalizer` was moved to common code out of GC implementations. This prevents custom-gc from trying to implement finalizers (I'd like to). There is a bit of copy-paste among GC implementations because of it, but especially with `gc_blocks` it's less than it could have been.

At the same time, this reimplements `KeepAlive` in the same way as Go - while the Go document does make a big deal about finalizers, close reading shows `ensures that the object is not freed`, which can be important with unsafe code. I don't think `KeepAlive` has a strong relation to `SetFinalizer` as is currently documented. In tinygo wasi's case, I believe this specifically means that the shadow stack slot for a variable is not reused by a different one until the call to `KeepAlive` (for non-wasi, same thing but registers?). However, I need help in knowing whether the `The compiler cannot see that alwaysFalse is always false,` note is valid for LLVM as well - intuitively for me, just `noinline` would be enough to ensure the variable is kept alive for the function call.

### DIFF
--- a/src/runtime/runtime.go
+++ b/src/runtime/runtime.go
@@ -88,13 +88,45 @@ func LockOSThread() {
 func UnlockOSThread() {
 }
 
+// Mark KeepAlive as noinline so that it is easily detectable as an intrinsic.
+//
+//go:noinline
+
+// KeepAlive marks its argument as currently reachable.
+// This ensures that the object is not freed, and its finalizer is not run,
+// before the point in the program where KeepAlive is called.
+//
+// A very simplified example showing where KeepAlive is required:
+//
+//	type File struct { d int }
+//	d, err := syscall.Open("/file/path", syscall.O_RDONLY, 0)
+//	// ... do something if err != nil ...
+//	p := &File{d}
+//	runtime.SetFinalizer(p, func(p *File) { syscall.Close(p.d) })
+//	var buf [10]byte
+//	n, err := syscall.Read(p.d, buf[:])
+//	// Ensure p is not finalized until Read returns.
+//	runtime.KeepAlive(p)
+//	// No more uses of p after this point.
+//
+// Without the KeepAlive call, the finalizer could run at the start of
+// syscall.Read, closing the file descriptor before syscall.Read makes
+// the actual system call.
+//
+// Note: KeepAlive should only be used to prevent finalizers from
+// running prematurely. In particular, when used with unsafe.Pointer,
+// the rules for valid uses of unsafe.Pointer still apply.
 func KeepAlive(x interface{}) {
-	// Unimplemented. Only required with SetFinalizer().
+	if alwaysFalse {
+		println(x)
+	}
 }
 
-func SetFinalizer(obj interface{}, finalizer interface{}) {
-	// Unimplemented.
-}
+// alwaysFalse is a boolean value that is always false.
+// The compiler cannot see that alwaysFalse is always false,
+// so it emits the test and keeps the call, giving the desired
+// escape analysis result. The test is cheaper than the call.
+var alwaysFalse bool
 
 //go:linkname godebug_setUpdate internal/godebug.setUpdate
 func godebug_setUpdate(update func(string, string)) {
--- a/src/runtime/runtime.go
+++ b/src/runtime/runtime.go
@@ -88,45 +88,10 @@ func LockOSThread() {
 func UnlockOSThread() {
 }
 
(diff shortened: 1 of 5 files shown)


src/runtime/runtime.go
```diff
@@ -88,13 +88,45 @@ func LockOSThread() {
...
 }
 
-func SetFinalizer(obj interface{}, finalizer interface{}) {
-	// Unimplemented.
-}
+// alwaysFalse is a boolean value that is always false.
+// The compiler cannot see that alwaysFalse is always false,
+// so it emits the test and keeps the call, giving the desired
+// escape analysis result. The test is cheaper than the call.
+var alwaysFalse bool
```
anuraaga: In Go, the variable is named `cgoAlwaysFalse` but I think this was just to reuse the same variable after the fact for something unrelated to cgo.

https://github.com/golang/go/blob/master/src/runtime/mfinal.go#L515

dgryski: I'm not convinced the new LTO logic will think `alwaysFalse` is potentially not always `false`.  I think we need something else here that we can ensure at the compiler level.

anuraaga: What do you think about just noinline being enough? If it seems so I could remove this stuff to prevent confusion.

anuraaga:
BTW I could implement `SetFinalizer` with bdwgc here, in case it helps to see what it may look like.

https://github.com/corazawaf/coraza-proxy-wasm/compare/main...anuraaga:setfinalizer?expand=1

Only limitation is finalizers must be `func(interface{})` since I think I would need `reflect` implemented to be able to support arbitrary funcs.

src/runtime/runtime.go
anuraaga: Went ahead and removed down to just adding go:noinline, verifying it actually does keepalive is probably quite tricky but in the meantime it can be attempted along with finalizers in custom GC.

anuraaga:
@dgryski Any other thoughts on this? Thanks!

dgryski:
My only concern is if "no inline" really is enough to convince tinygo/llvm to keep allocations alive. We might need direct compiler support.

/Cc @aykevl

anuraaga:
Not sure how hard that is, but assuming this PR doesn't make anything particularly worse, hopefully it's still possible to continue to fix forward. Mainly hoping to be able to implement a full custom GC for 0.27.0 including finalizer :)

dgryski (approved):
Fair enough.
LGTM

anuraaga:
Saw 0.27.0 update merge to release branch - just curious if this could get in for 0.27.0. Otherwise finalizer would have to wait until 0.28, not the end of the world but figured doesn't hurt to check.

dgryski:
No complaints from my end to get this into 0.27.

deadprogram:
@aykevl any other feedback before I squash/merge?

src/runtime/runtime.go
```diff
@@ -88,11 +88,8 @@ func LockOSThread() {
...
 }
 
+//go:noinline
 func KeepAlive(x interface{}) {
-	// Unimplemented. Only required with SetFinalizer().
-}
-
-func SetFinalizer(obj interface{}, finalizer interface{}) {
 	// Unimplemented.
 }
```
aykevl: `//go:noinline` doesn't do anything here, except avoid calling a no-op function. Most importantly, this does nothing to keep an object alive: LLVM will happily remove the `x` parameter because it is not used.

I think the only reliable way to implement this is either by using a volatile operation, or by calling an assembly stub (that simply returns).

aykevl (commented):
The `SetFinalizer` parts seem fine by me. But `KeepAlive` is still unimplemented with this PR, see my comment below.

src/runtime/runtime.go
anuraaga: Thanks for the hint, I went ahead and removed the inline and added a TODO based on it. Would like to dig into it in a followup

aykevl:
Thank you! I'll take a look at implementing ~SetFinalizer~ KeepAlive.
(The Windows build failed but I'm 99% sure that's not because of a bug in this PR so I merged it anyway - also, I can't see why this would fail on Windows only).

aykevl:
Implemented `KeepAlive`: https://github.com/tinygo-org/tinygo/pull/3455

PR #3421, opened 2023-01-31
Add custom-gc stub for ReadMemStats

anuraaga:
Missed this one since normal memory allocator operations don't touch it, but `tinygo test` does always use it


dgryski (approved):
LGTM

deadprogram:
Thank you for the addition @anuraaga and to @dgryski for review. Now merging.

PR #3423, opened 2023-02-01
Switch WASI emulator from wasmtime to wazero

anuraaga:
wasmtime must be installed separately, but all TinyGo users have Go installed, so wazero can be used to emulate wasi without any installation step at all with the help of `go run`. This should really improve the UX of `tinygo test`.

All the directory hacks for wasmtime can also be removed. Why is this? Because @codefromthecrypt developed CLI features specifically while [targeting](https://github.com/tetratelabs/wazero/blob/main/.github/workflows/integration.yaml#L123) TinyGo's test suite - TinyGo is a first-class target for wazero and its UX will continue to matter to wazero (I guess, sorry for speaking for you @codefromthecrypt :P).

### DIFF
--- a/main.go
+++ b/main.go
@@ -761,6 +761,12 @@ func buildAndRun(pkgName string, config *compileopts.Config, stdout io.Writer, c
 				"runtime": runtimeGlobals,
 			}
 		}
+	} else if strings.Contains(config.Target.Emulator, "wazero") {
+		// wazero needs some special flags to pass environment variables
+		for _, v := range environmentVars {
+			args = append(args, "-env", v)
+		}
+		args = append(args, cmdArgs...)
 	} else {
 		// Pass environment variables and command line parameters as usual.
 		// This also works on qemu-aarch64 etc.
--- a/main.go
+++ b/main.go
@@ -266,39 +266,6 @@ func Test(pkgName string, stdout, stderr io.Writer, options *compileopts.Options
 		// Tests are always run in the package directory.
 		cmd.Dir = result.MainDir
 
-		// wasmtime is the default emulator used for `-target=wasi`. wasmtime
-		// is a WebAssembly runtime CLI with WASI enabled by default. However,
-		// only stdio are allowed by default. For example, while STDOUT routes
-		// to the host, other files don't. It also does not inherit environment
-		// variables from the host. Some tests read testdata files, often from
-		// outside the package directory. Other tests require temporary
-		// writeable directories. We allow this by adding wasmtime flags below.
-		if config.EmulatorName() == "wasmtime" {
-			// At this point, The current working directory is at the package
-			// directory. Ex. $GOROOT/src/compress/flate for compress/flate.
-			// buildAndRun has already added arguments for wasmtime, that allow
-			// read-access to files such as "testdata/huffman-zero.in".
-			//
-			// Ex. main(.wasm) --dir=. -- -test.v
-
-			// Below adds additional wasmtime flags in case a test reads files
-			// outside its directory, like "../testdata/e.txt". This allows any
-			// relative directory up to the module root, even if the test never
-			// reads any files.
-			//
-			// Ex. run --dir=.. --dir=../.. --dir=../../..
-			dirs := dirsToModuleRoot(result.MainDir, result.ModuleRoot)
-			args := []string{"run"}
-			for _, d := range dirs[1:] {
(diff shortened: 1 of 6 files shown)


codefromthecrypt:
My name is @codefromthecrypt and I approve of this message

codefromthecrypt:
by the way, it wasm @anuraaga's idea to relax the mount by default, and literally because of this code in tinygo!

main.go
```diff
@@ -794,18 +761,12 @@ func buildAndRun(pkgName string, config *compileopts.Config, stdout io.Writer, c
 				"runtime": runtimeGlobals,
 			}
 		}
-	} else if config.EmulatorName() == "wasmtime" {
-		// Wasmtime needs some special flags to pass environment variables
-		// and allow reading from the current directory.
-		args = append(args, "--dir=.")
+	} else if strings.Contains(config.Target.Emulator, "wazero") {
+		// wazero needs some special flags to pass environment variables
```
codefromthecrypt, line 765: maybe add the why?

```suggestion
		// wasm is sandboxed by default, so we explicitly propagate ENV variables used in tests.
```

We could also in the CLI allow env propagation, either opt-in or by default.. I wouldn't change the go library default, but for example, the CLI uses real time also which isn't a default.

codefromthecrypt: counterpoint is docker doesn't automatically propagate env from the host, so in general I prefer the explicitness.

anuraaga: Yeah would probably be good to have a flag to opt-in the environment, that would allow tinygo test and go test to behave the same regardless of target, and similar use cases.

anuraaga:
Linux failure isn't a flake will fix it tomorrow

main.go
codefromthecrypt: for this though I would say that since environmentVars are already defined in tinygo we don't need to block on an auto-propagation thing. I also noticed docker doesn't do it, and it is likely not to have a lot of maintenance here in go, as I don't expect them to add host env sensitive variables into tests that often.

dgryski:
The test corpus (at least the version I run) already takes about 30 minutes to complete with wasmtime.  How long does it take with wazero?

(I'm all up for patching tinygo to support wazero, but I'm not sure I want to make it the default.)

codefromthecrypt:
good point on time. I think until we add the fancy optimized wasm in wasm thing at least the crypto stuff might take longer than wasmtime. @anuraaga have you noticed a before/after?

codefromthecrypt:
by the way, we can and already do run wazero for tinygo tests, just we don't use the makeflle and dodge a slow package https://github.com/tetratelabs/wazero/blob/main/.github/workflows/integration.yaml#L166

dgryski:
Could we add `--emulator=wazero` or something to override the default value of wasmtime?

anuraaga:
Thanks @dgryski - on my Mac laptop, corpus tests are

wasmtime: user=1348.19s system=122.36s cpu=764% total=3:12.40
wazero: user=2143.33s system=202.82s cpu=709% total=5:30.84

CI is wasmtime 8m vs wazero 10m for test-wasi-fast.

Corpus in particularly is significantly different - I can definitely see not wanting to regress on the default. Adding the flag feels to me like it then adds complication to the UX where we wanted to simplify it, as the decision of tradeoffs gets handed off to the user, and I suspect in most cases they'll probably just install wasmtime to match the default. It's still possible to use wazero instead by installing a wasmtime CLI-compatible wazero wrapper anyways. Let me go ahead and close this PR for now - there's some fun perf work happening on wazero and I think we'll eventually be able to make a switch without regression.

dgryski:
SGTM.  Please continue to push for fixes that make integration with wazero easier.

anuraaga:
Just for reference, this is what using a wrapper looks like.

https://github.com/wasilibs/go-re2/pull/23
https://github.com/wasilibs/tools/blob/main/cmd/wasmtime/wasmtime.go

It took a while to find that installation syntax for windows (didn't know about scoop) so glad to see that all collapse to a single line.

codefromthecrypt:
for wazero 1.0 which should be march, I'll make sure we have packaging in place, too. It is cool we can naturally work in windows because go does and we test it, but also reasonable to have a package installer, too. It might be a nice time to revisit adding a default emulator here once that's in place, despite the ease of `go install`

thanks for all the attention meanwhile!

Issue #742, opened 2019-11-21
Trying to compile gopher-lua with TinyGo issues


thegrumpylion:
I have forked and striped of any "os" dependencies this package https://github.com/yuin/gopher-lua (my fork https://github.com/thegrumpylion/gopher-lua) and i am trying to build this example https://github.com/thegrumpylion/tinygo/blob/lua/src/examples/lua/lua.go

`tinygo build -target=pca10056 -o lua.hex ./src/examples/lua`

But i hit this wall:

```
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x3810ac0]

runtime stack:
runtime.throw(0x3b8f74d, 0x2a)
        /home/nikolas/go/src/runtime/panic.go:774 +0x72
runtime.sigpanic()
        /home/nikolas/go/src/runtime/signal_unix.go:378 +0x47c

goroutine 1 [syscall]:
runtime.cgocall(0xcb0050, 0xc00c1dc850, 0xc00c1dc848)
        /home/nikolas/go/src/runtime/cgocall.go:128 +0x5b fp=0xc00c1dc820 sp=0xc00c1dc7e8 pc=0x92dafb
tinygo.org/x/go-llvm._Cfunc_LLVMGetConstOpcode(0x7f73582d5e88, 0x7f7300000000)
        _cgo_gotypes.go:5253 +0x49 fp=0xc00c1dc850 sp=0xc00c1dc820 pc=0xb48b99
tinygo.org/x/go-llvm.Value.Opcode.func1(0x7f73582d5e88, 0x8)
        /home/nikolas/golang/pkg/mod/tinygo.org/x/go-llvm@v0.0.0-20191113125529-bad6d01809e8/ir.go:872 +0x56 fp=0xc00c1dc888 sp=0xc00c1dc850 pc=0xb6a7c6
tinygo.org/x/go-llvm.Value.Opcode(0x7f73582d5e88, 0x60f0d50)
        /home/nikolas/golang/pkg/mod/tinygo.org/x/go-llvm@v0.0.0-20191113125529-bad6d01809e8/ir.go:872 +0x2b fp=0xc00c1dc8a8 sp=0xc00c1dc888 pc=0xb573bb
github.com/tinygo-org/tinygo/interp.(*MapValue).PutBinary(0xc0079f04d0, 0xc003334780, 0xc003c77d40)
[...]

PR #3424, opened 2023-02-01
Draft: `runtime`: MemStats.Alloc/HeapAlloc support

bgould:
https://github.com/tinygo-org/tinygo/issues/742#issuecomment-1411026796

### DIFF
--- a/src/runtime/mstats.go
+++ b/src/runtime/mstats.go
@@ -9,6 +9,11 @@ package runtime
 type MemStats struct {
 	// General statistics.
 
+	// Alloc is bytes of allocated heap objects.
+	//
+	// This is the same as HeapAlloc (see below).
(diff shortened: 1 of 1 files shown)


dgryski:
Probably makes sense to fix some of the garbage collectors to track information for memstats or add comments indicating that we don't.

bgould:
Agreed ... digging into this to understand the scope of the work and see if there is any low hanging fruit to start.

Also thinking it might make sense to break out the reflect stubs into a separate PR since it is not really related to MemStats.

bgould:
Split `reflect` changes into separate PR: https://github.com/tinygo-org/tinygo/pull/3425

dgryski:
Merged as part of #4506

PR #3425, opened 2023-02-02
`reflect`: Select constants and stubs

bgould:
Adds constants and stubs to `reflect` package to support compilation mentioned in https://github.com/tinygo-org/tinygo/issues/742#issuecomment-1411026796

### DIFF
--- a/src/reflect/chan.go
+++ b/src/reflect/chan.go
@@ -1,0 +1,30 @@
+package reflect
+
+type SelectDir int
+
+const (
+	_             SelectDir = iota
+	SelectSend              // case Chan <- Send
+	SelectRecv              // case <-Chan:
+	SelectDefault           // default
+)
+
+type SelectCase struct {
+	Dir  SelectDir // direction of case
+	Chan Value     // channel to use (for send or receive)
+	Send Value     // value to send (for send)
+}
+
+func Select(cases []SelectCase) (chosen int, recv Value, recvOK bool) {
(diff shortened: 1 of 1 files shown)


dgryski (changes requested):
These should probably be in `value.go` with everything else. I just put them in chan.go for convenience when I was trying to get the lua stuff to build.

src/reflect/chan.go
```diff
@@ -0,0 +1,30 @@
...
+	Send Value     // value to send (for send)
+}
+
+func Select(cases []SelectCase) (chosen int, recv Value, recvOK bool) {
+	panic("reflect.Select: unimplemented")
+}
+
+func (v Value) Send(x Value) {
+	panic("reflect.Value.Send(): unimplemented")
+
```
aykevl, line 24: Nit: unnecessary newline

src/reflect/chan.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+func (v Value) Close() {
+	panic("reflect.Value.Close(): unimplemented")
+
```
aykevl, line 29: Same here.

aykevl (commented):
> These should probably be in `value.go` with everything else.

Agreed, plus two really minor nits you can fix at the same time. Other than that, this looks good to me.

dgryski:
Incorporated this change into https://github.com/tinygo-org/tinygo/pull/3470

PR #3433, opened 2023-02-07
docker: add .git entry to .dockerignore file

deadprogram:
THis PR adds `.git` to the .dockerignore file so that the `dev` image does not return the error `error obtaining VCS status: exit status 128` when being run on environments such as Github actions where the directory where the files are checked out to are on the "host" and not inside the container itself.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,3 @@
 build/
 llvm-*/
-
+.git


deadprogram:
For example https://github.com/tinygo-org/bluetooth/actions/runs/4118285839/jobs/7110621115#step:6:7 :crying_cat_face:

dgryski (approved):
LGTM

dgryski:
Anything else we should add while we're here?

deadprogram:
I'm just looking for the build fix right now, but agree we should look at that, @dgryski

deadprogram:
OK that did not work as expected, it caused the docker build of the dev image to fail. Instead of reverting I dropped the commit from `dev`. Need to study the solution a bit closer. :crying_cat_face:

PR #3437, opened 2023-02-10
compiler: zero struct padding during map operations

### LINKED ISSUE #3358: map delete not finding keys when value is largeish, key hash seems to be incorrect
I am currently debugging this code that tries to delete the keys in a map (to empty it while retaining bucket storage)

https://github.com/corazawaf/coraza/blob/v3/dev/internal/corazawaf/rulegroup.go#L105

(for the below, note that I tried copying the keys into a slice before deleting to see if it was iterator+delete behavior problem but it was all the same so simplified back to normal map iteration).

I noticed a memory leak and after adding `if len(transformationCache) > 0 { panic }` after the deletion could verify that the map wasn't getting cleared with tinygo wasm. It works fine with Go.

I have added this instrumentation to hashmap.go

https://github.com/anuraaga/tinygo/compare/custom-gc...anuraaga:tinygo:hashmapdelete-debug?expand=1

And shrunk down the test to similar one as last time

https://github.com/corazawaf/coraza-proxy-wasm/compare/main...anuraaga:coraza-1225?expand=1#diff-173fbfd8d8844658344b121461b4290d0a85230caae9825240705df8130e8b75R33
`~/go/bin/tinygo build -scheduler=none -target=wasi -opt=2 -o mapbench.wasm`

The debug output looks something like this

```
hashmapBinaryGet 0x0001019c 0x0000ffd0 16 1666840535
could find key
hashmapBinaryDelete 0x0001019c 0x0000ffb8 16 1666840535
delete could find key 0x0001019c 0x0000ffb8 1666840535
hashmapBinaryGet 0x0001019c 0x0000ffd0 16 1666840535
couldn't find key
```

The key address for get is `0x0000ffd0` and delete is `0x0000ffb8`. That being said the hash is the same in this example so it's being able to clear the map but with the same instrumentation when looking at the output for the original coraza code, the hash values were also different. I'm not sure why I wasn't able to reproduce this hash value difference, but either way, the key is in a local variable `k`, which there is only one of, so intuitively I suppose those addresses must be the same and the difference is unexpected.

One weird thing I found while making the repro is it seems the value struct needs to be more than 2 strings worth of size to exhibit the behavior - with three fields, get and delete have different addresses, while with two fields they are the same.

Looked through the code in `hashmap.go` and `map.go` (compiler) and couldn't see anything suspicious, the code paths for get/lookup vs delete look basically the same for both, but the difference does cause our real-world code to fail with the map not being cleared. With the code looking the same, the issue may be in IR reduction?

Note the above simpleish test case approach is also applied to the real world code here (which is where I was observing the address+key value difference)

https://github.com/corazawaf/coraza/compare/v3/dev...anuraaga:scratch?expand=1

The output looks like this (we can see the different hash values)

```
[2022-12-28 04:04:56.210312][25][info][wasm] [source/extensions/common/wasm/context.cc:1170] wasm log coraza-filter my_vm_id: hashmapBinaryGet 0x09887b70 0x0000fb98 16 2748313615
[2022-12-28 04:04:56.210312][25][info][wasm] [source/extensions/common/wasm/context.cc:1170] wasm log coraza-filter my_vm_id: could find key
[2022-12-28 04:04:56.210312][25][info][wasm] [source/extensions/common/wasm/context.cc:1170] wasm log coraza-filter my_vm_id: hashmapBinaryDelete 0x09887b70 0x0000fbf8 16 3003228291
[2022-12-28 04:04:56.210313][25][info][wasm] [source/extensions/common/wasm/context.cc:1170] wasm log coraza-filter my_vm_id: delete couldn't find key 0x09887b70
[2022-12-28 04:04:56.210313][25][info][wasm] [source/extensions/common/wasm/context.cc:1170] wasm log coraza-filter my_vm_id: hashmapBinaryGet 0x09887b70 0x0000fb98 16 2748313615
[2022-12-28 04:04:56.210314][25][info][wasm] [source/extensions/common/wasm/context.cc:1170] wasm log coraza-filter my_vm_id: could find key
```

dgryski:
Fixes #3358

### DIFF
--- a/compiler/map.go
+++ b/compiler/map.go
@@ -325,7 +325,14 @@ func (b *builder) zeroUndefBytes(llvmType llvm.Type, ptr llvm.Value) error {
 			if fieldEndOffset != nextOffset {
 				n := llvm.ConstInt(b.uintptrType, nextOffset-fieldEndOffset, false)
 				llvmStoreSize := llvm.ConstInt(b.uintptrType, storeSize, false)
-				paddingStart := b.CreateInBoundsGEP(b.i8ptrType, elemPtr, []llvm.Value{llvmStoreSize}, "")
+				gepPtr := elemPtr
+				if gepPtr.Type() != b.i8ptrType {
+					gepPtr = b.CreateBitCast(gepPtr, b.i8ptrType, "") // LLVM 14
+				}
+				paddingStart := b.CreateInBoundsGEP(b.ctx.Int8Type(), gepPtr, []llvm.Value{llvmStoreSize}, "")
+				if paddingStart.Type() != b.i8ptrType {
+					paddingStart = b.CreateBitCast(paddingStart, b.i8ptrType, "") // LLVM 14
+				}
 				b.createRuntimeCall("memzero", []llvm.Value{paddingStart, n}, "")
 			}
 		}
--- a/compiler/map.go
+++ b/compiler/map.go
@@ -89,6 +89,7 @@ func (b *builder) createMapLookup(keyType, valueType types.Type, m, key llvm.Val
 		// growth.
 		mapKeyAlloca, mapKeyPtr, mapKeySize := b.createTemporaryAlloca(key.Type(), "hashmap.key")
 		b.CreateStore(key, mapKeyAlloca)
+		b.zeroUndefBytes(b.getLLVMType(keyType), mapKeyAlloca)
 		// Fetch the value from the hashmap.
 		params := []llvm.Value{m, mapKeyPtr, mapValuePtr, mapValueSize}
 		commaOkValue = b.createRuntimeCall("hashmapBinaryGet", params, "")
@@ -133,6 +134,7 @@ func (b *builder) createMapUpdate(keyType types.Type, m, key, value llvm.Value, 
 		// key can be compared with runtime.memequal
 		keyAlloca, keyPtr, keySize := b.createTemporaryAlloca(key.Type(), "hashmap.key")
 		b.CreateStore(key, keyAlloca)
+		b.zeroUndefBytes(b.getLLVMType(keyType), keyAlloca)
 		params := []llvm.Value{m, keyPtr, valuePtr}
 		b.createRuntimeCall("hashmapBinarySet", params, "")
 		b.emitLifetimeEnd(keyPtr, keySize)
@@ -161,6 +163,7 @@ func (b *builder) createMapDelete(keyType types.Type, m, key llvm.Value, pos tok
 	} else if hashmapIsBinaryKey(keyType) {
 		keyAlloca, keyPtr, keySize := b.createTemporaryAlloca(key.Type(), "hashmap.key")
 		b.CreateStore(key, keyAlloca)
+		b.zeroUndefBytes(b.getLLVMType(keyType), keyAlloca)
 		params := []llvm.Value{m, keyPtr}
 		b.createRuntimeCall("hashmapBinaryDelete", params, "")
 		b.emitLifetimeEnd(keyPtr, keySize)
@@ -240,7 +243,8 @@ func (b *builder) createMapIteratorNext(rangeVal ssa.Value, llvmRangeVal, it llv
 }
 
 // Returns true if this key type does not contain strings, interfaces etc., so
-// can be compared with runtime.memequal.
+// can be compared with runtime.memequal.  Note that padding bytes are undef
+// and can alter two "equal" structs being equal when compared with memequal.
 func hashmapIsBinaryKey(keyType types.Type) bool {
 	switch keyType := keyType.(type) {
 	case *types.Basic:
@@ -262,4 +266,70 @@ func hashmapIsBinaryKey(keyType types.Type) bool {
 	default:
 		return false
 	}
+}
+
+func (b *builder) zeroUndefBytes(llvmType llvm.Type, ptr llvm.Value) error {
+	// We know that hashmapIsBinaryKey is true, so we only have to handle those types that can show up there.
+	// To zero all undefined bytes, we iterate over all the fields in the type.  For each element, compute the
+	// offset of that element.  If it's Basic type, there are no internal padding bytes.  For compound types, we recurse to ensure
+	// we handle nested types.  Next, we determine if there are any padding bytes before the next
+	// element and zero those as well.
+
+	zero := llvm.ConstInt(b.ctx.Int32Type(), 0, false)
+
+	switch llvmType.TypeKind() {
+	case llvm.IntegerTypeKind:
+		// no padding bytes
+		return nil
+	case llvm.PointerTypeKind:
+		// mo padding bytes
+		return nil
+	case llvm.ArrayTypeKind:
+		llvmArrayType := llvmType
+		llvmElemType := llvmType.ElementType()
+
+		for i := 0; i < llvmArrayType.ArrayLength(); i++ {
+			idx := llvm.ConstInt(b.uintptrType, uint64(i), false)
+			elemPtr := b.CreateInBoundsGEP(llvmArrayType, ptr, []llvm.Value{zero, idx}, "")
+
+			// zero any padding bytes in this element
+			b.zeroUndefBytes(llvmElemType, elemPtr)
+		}
+
+	case llvm.StructTypeKind:
+		llvmStructType := llvmType
+		numFields := llvmStructType.StructElementTypesCount()
+		llvmElementTypes := llvmStructType.StructElementTypes()
+
+		for i := 0; i < numFields; i++ {
+			idx := llvm.ConstInt(b.ctx.Int32Type(), uint64(i), false)
+			elemPtr := b.CreateInBoundsGEP(llvmStructType, ptr, []llvm.Value{zero, idx}, "")
+
+			// zero any padding bytes in this field
+			llvmElemType := llvmElementTypes[i]
+			b.zeroUndefBytes(llvmElemType, elemPtr)
+
+			// zero any padding bytes before the next field, if any
+			offset := b.targetData.ElementOffset(llvmStructType, i)
+			storeSize := b.targetData.TypeStoreSize(llvmElemType)
+			fieldEndOffset := offset + storeSize
+
+			var nextOffset uint64
+			if i < numFields-1 {
+				nextOffset = b.targetData.ElementOffset(llvmStructType, i+1)
+			} else {
+				// Last field?  Next offset is the total size of the allcoate struct.
+				nextOffset = b.targetData.TypeAllocSize(llvmStructType)
+			}
+
+			if fieldEndOffset != nextOffset {
+				n := llvm.ConstInt(b.uintptrType, nextOffset-fieldEndOffset, false)
+				llvmStoreSize := llvm.ConstInt(b.uintptrType, storeSize, false)
+				paddingStart := b.CreateInBoundsGEP(b.i8ptrType, elemPtr, []llvm.Value{llvmStoreSize}, "")
+				b.createRuntimeCall("memzero", []llvm.Value{paddingStart, n}, "")
+			}
+		}
+	}
+
+	return nil
 }
--- a/compiler/testdata/zeromap.go
+++ b/compiler/testdata/zeromap.go
@@ -1,0 +1,37 @@
+package main
+
+type hasPadding struct {
+	b1 bool
+	i  int
+	b2 bool
+}
+
+type nestedPadding struct {
+	b bool
+	hasPadding
+	i int
+}
+
+//go:noinline
+func testZeroGet(m map[hasPadding]int, s hasPadding) int {
+	return m[s]
+}
+
+//go:noinline
+func testZeroSet(m map[hasPadding]int, s hasPadding) {
+	m[s] = 5
+}
+
+//go:noinline
+func testZeroArrayGet(m map[[2]hasPadding]int, s [2]hasPadding) int {
+	return m[s]
+}
+
+//go:noinline
+func testZeroArraySet(m map[[2]hasPadding]int, s [2]hasPadding) {
+	m[s] = 5
+}
+
+func main() {
+
+}
--- a/compiler/compiler_test.go
+++ b/compiler/compiler_test.go
@@ -49,6 +49,7 @@ func TestCompiler(t *testing.T) {
 		{"goroutine.go", "cortex-m-qemu", "tasks"},
 		{"channel.go", "", ""},
 		{"gc.go", "", ""},
+		{"zeromap.go", "", ""},
 	}
 	if goMinor >= 20 {
 		tests = append(tests, testCase{"go1.20.go", "", ""})
--- a/compiler/testdata/zeromap.ll
+++ b/compiler/testdata/zeromap.ll
@@ -1,0 +1,170 @@
+; ModuleID = 'zeromap.go'
+source_filename = "zeromap.go"
+target datalayout = "e-m:e-p:32:32-p10:8:8-p20:8:8-i64:64-n32:64-S128-ni:1:10:20"
+target triple = "wasm32-unknown-wasi"
+
+%main.hasPadding = type { i1, i32, i1 }
+
+declare noalias nonnull ptr @runtime.alloc(i32, ptr, ptr) #0
+
+declare void @runtime.trackPointer(ptr nocapture readonly, ptr, ptr) #0
+
+; Function Attrs: nounwind
+define hidden void @main.init(ptr %context) unnamed_addr #1 {
+entry:
+  ret void
+}
+
+; Function Attrs: noinline nounwind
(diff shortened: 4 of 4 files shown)


dgryski:
More testing and cleanup required.  But otherwise I'm happy with this.

dgryski:
Fixes the test case from #3358.

dgryski:
Still need to come up with a proper test case to add, but I think this is ready for review otherwise.

dgryski:
@anuraaga Can you test this fix in your application?

anuraaga:
Seems to be good, thanks @dgryski!

deadprogram:
@aykevl any comments here before merge?

compiler/map.go
```diff
@@ -263,3 +267,73 @@ func hashmapIsBinaryKey(keyType types.Type) bool {
...
+		for i := int64(0); i < typ.Len(); i++ {
+			// for each element, first clear any undef bytes in the element itself
+			idx := llvm.ConstInt(b.uintptrType, uint64(i), false)
+			ptr := b.CreateGEP(llvmArrayType, base, []llvm.Value{idx}, "")
+
+			// recursively zero any padding bytes in this element
+			b.zeroUndefBytes(typ.Elem(), ptr)
+
+			// check for padding between elements
+			// TODO(dgryski): typeSizeEqualStoreSize ?
```
aykevl:
```go
			// TODO(dgryski): typeSizeEqualStoreSize ?
```

What does this mean?

compiler/map.go
```diff
@@ -263,3 +267,73 @@ func hashmapIsBinaryKey(keyType types.Type) bool {
...
+			if allocSize, storeSize := b.targetData.TypeAllocSize(llvmElemType), b.targetData.TypeStoreSize(llvmElemType); allocSize != storeSize {
+				n := llvm.ConstInt(b.uintptrType, allocSize-storeSize, false)
+				llvmSize := llvm.ConstInt(b.uintptrType, storeSize, false)
+				paddingStart := b.CreateGEP(b.ctx.Int8Type(), ptr, []llvm.Value{llvmSize}, "")
+				b.createRuntimeCall("memzero", []llvm.Value{paddingStart, n}, "")
+			}
```
aykevl: Can you give an example when this is necessary? I can't think of a case where array elements leave extra space after the element.

compiler/map.go
```diff
@@ -263,3 +267,73 @@ func hashmapIsBinaryKey(keyType types.Type) bool {
...
+		}
+
+	case *types.Struct:
+		llvmStructType := b.getLLVMType(typ)
+		base := ptr
+
+		for i := 0; i < typ.NumFields(); i++ {
+			offset := b.targetData.ElementOffset(llvmStructType, int(i))
+			llvmOffset := llvm.ConstInt(b.uintptrType, offset, false)
+			ptr := b.CreateGEP(b.ctx.Int8Type(), base, []llvm.Value{llvmOffset}, "")
```
aykevl: Suggestion (I think this is less confusing):

```suggestion
			elemPtr := b.CreateGEP(b.ctx.Int8Type(), ptr, []llvm.Value{llvmOffset}, "")
```

aykevl (commented):
This PR should really get some tests. Perhaps in a new compiler/testdata/map.go, with tests like:

```go
type paddedStruct struct {
    byte a
    int  b
}
func storeToPaddedStruct(m map[paddedStruct]int, key paddedStruct {
    m[key] = 3
}
```

(Didn't test this code and whether it reproduces the issue, but I think it does).

Also a suggestion: you could use LLVM types throughout instead of Go types. That is, use `llvm.Type` instead of `types.Type` for the `typ` parameter. That might be a little bit more reliable as it certainly catches all types that need explicit zeroing for the padding (in case a Go type is ever implemented as a struct with padding).
The zeroing algorithm itself looks reasonable to me.

compiler/map.go
dgryski: This was a reminder to myself to see if I could use `bool llvm::DataLayout::typeSizeEqualsStoreSize()` but it's not exported by `go-llvm`, so I'd need to patch that first.

compiler/map.go
dgryski: You're probably right.  I think I was doing this late at night and now I'm pretty sure the reason I thought I needed to do this was wrong: struct types will be padded to their largest alignment, so consecutive array elements will always be safe and not need any extra padding between them.  I'll remove this.

dgryski:
I'll update this to to use llvm types instead of Go types.

Also I tried to create a unit test for this, and while I *could* create a case where I was accessing a map with a padded struct, I couldn't cause it to have non-zero padding bytes.

dgryski:
I think I addressed all the issues.  Need to run this through the test corpus.  Still need to write an actual test.

aykevl:
Thinking some more, I think you also need to zero the end of structs. For example, with this one:

```go
type someStruct struct {
    a float32
    b char
}
```
This struct is 8 bytes, but only the first 5 are written to.

> Also I tried to create a unit test for this, and while I could create a case where I was accessing a map with a padded struct, I couldn't cause it to have non-zero padding bytes.

Unit tests would be great, but are pretty difficult in the case of undefined behavior.
What I was asking was just IR output to show the memzero calls in IR (add a Go file to compiler/testdata, add the filename to compiler/compiler_test.go, and run `go test ./compiler -update`). That also makes review a bit easier as it clearly shows the calls are inserted when needed.

dgryski:
> Thinking some more, I think you also need to zero the end of structs. For example, with this one:

Ah yes.  When I was looking at the code I assumed that `keySize` ended up being the store size, *not* the alloc size. So indeed I'm missing those bytes.  I'll fix this.

dgryski:
Now with a unit test.  PTAL.

dgryski:
Passes the test corpus, which is always nice.

compiler/map.go
```diff
@@ -263,3 +267,71 @@ func hashmapIsBinaryKey(keyType types.Type) bool {
...
+	case llvm.PointerTypeKind:
+		// mo padding bytes
+		return nil
+	case llvm.ArrayTypeKind:
+		llvmArrayType := llvmType
+		llvmElemType := llvmType.ElementType()
+
+		for i := 0; i < llvmArrayType.ArrayLength(); i++ {
+			idx := llvm.ConstInt(b.uintptrType, uint64(i), false)
+			elemPtr := b.CreateGEP(llvmArrayType, ptr, []llvm.Value{idx}, "")
```
aykevl: You can use `CreateInBoundsGEP` here (and other places in this PR), because you know the GEP is going to be in bounds. ("in bounds", as I understand it, basically means you are 100% sure the pointer is still pointing inside the allocated object - in this case, the alloca).

Also, I believe you need to add a zero index as the first index parameter to the GEP, like here:

https://github.com/tinygo-org/tinygo/blob/0d56dee00f49bd50eb373c02c30062a75ec28f10/compiler/compiler.go#L2020-L2021

More information: https://llvm.org/docs/GetElementPtr.html

compiler/map.go
```diff
@@ -263,3 +267,71 @@ func hashmapIsBinaryKey(keyType types.Type) bool {
...
+			// zero any padding bytes in this element
+			b.zeroUndefBytes(llvmElemType, elemPtr)
+		}
+
+	case llvm.StructTypeKind:
+		llvmStructType := llvmType
+		numFields := llvmStructType.StructElementTypesCount()
+		llvmElementTypes := llvmStructType.StructElementTypes()
+
+		for i := 0; i < llvmStructType.StructElementTypesCount(); i++ {
```
aykevl: Suggestion, I think this looks better:

```suggestion
		for i := 0; i < numFields; i++ {
```

Or even:

```suggestion
		for i, llvmElemType := range llvmElementTypes {
```

compiler/map.go
```diff
@@ -263,3 +267,71 @@ func hashmapIsBinaryKey(keyType types.Type) bool {
...
+			offset := b.targetData.ElementOffset(llvmStructType, i)
+			llvmOffset := llvm.ConstInt(b.uintptrType, offset, false)
+			elemPtr := b.CreateGEP(b.ctx.Int8Type(), ptr, []llvm.Value{llvmOffset}, "")
```
aykevl: This line will likely lead to a verifier failure in LLVM 14. To fix it, you need to cast `ptr` to an `i8*`. You can use `b.CreateBitCast(ptr, b.i8ptrType, "")` to cast the pointer type. And cast back to the pointer element type after the GEP.

compiler/map.go
```diff
@@ -263,3 +267,71 @@ func hashmapIsBinaryKey(keyType types.Type) bool {
...
+
+			// zero any padding bytes in this field
+			llvmElemType := llvmElementTypes[i]
+			b.zeroUndefBytes(llvmElemType, elemPtr)
+
+			// zero any padding bytes before the next field, if any
+			if i < numFields-1 {
+				nextOffset := b.targetData.ElementOffset(llvmStructType, i+1)
+				if storeSize := b.targetData.TypeStoreSize(llvmElemType); offset+storeSize != nextOffset {
+					n := llvm.ConstInt(b.uintptrType, nextOffset-(offset+storeSize), false)
```
aykevl: Suggestion. Doing the calculation only once may be easier to read.

```suggestion
				gap := nextOffset-(offset+b.targetData.TypeStoreSize(llvmElemType))
				if gap > 0 {
					n := llvm.ConstInt(b.uintptrType, gap, false)
```

compiler/map.go
```diff
@@ -263,3 +267,71 @@ func hashmapIsBinaryKey(keyType types.Type) bool {
...
+					llvmSize := llvm.ConstInt(b.uintptrType, storeSize, false)
+					paddingStart := b.CreateGEP(b.ctx.Int8Type(), elemPtr, []llvm.Value{llvmSize}, "")
```
aykevl: Another GEP that will likely fail in LLVM 14.

compiler/map.go
```diff
@@ -263,3 +267,71 @@ func hashmapIsBinaryKey(keyType types.Type) bool {
...
+		allocSize := b.targetData.TypeAllocSize(llvmStructType)
+		lastElemOffset := b.targetData.ElementOffset(llvmStructType, numFields-1)
+		lastElemStore := b.targetData.TypeStoreSize(llvmElementTypes[numFields-1])
+
+		if allocSize != lastElemOffset+lastElemStore {
+			n := llvm.ConstInt(b.uintptrType, allocSize-(lastElemOffset+lastElemStore), false)
+			llvmSize := llvm.ConstInt(b.uintptrType, lastElemOffset+lastElemStore, false)
+			paddingStart := b.CreateGEP(b.ctx.Int8Type(), ptr, []llvm.Value{llvmSize}, "")
+			b.createRuntimeCall("memzero", []llvm.Value{paddingStart, n}, "")
+		}
```
aykevl: You could consider merging these in the for loop above, with logic like this (in pseudocode):

```
if last field:
    nextOffset = one-past-the-end of the struct
else:
    nextOffset = start offset of next field
if offset + TypeAllocSize(field) != nextOffset:
    zero padding
```

Not sure it's more readable though.

compiler/testdata/zeromap.go
```diff
@@ -0,0 +1,20 @@
...
+	var s hasPadding
+
+	for i := 0; i < 10; i++ {
+		s.b1 = i&1 == 0
+		s.i = i
+		s.b2 = i&1 == 1
+		m[s]++
+	}
+	println(len(m))
+}
```
aykevl: Suggestion: use the following code to test:

```go
func testZeroGet(m map[hasPadding]int, s hasPadding) int {
        return m[s]
}

func testZeroSet(m map[hasPadding]int, s hasPadding) {
        m[s] = 5
}
```

This makes zeromap.ll smaller and more readable while still testing the important parts (the calls to `runtime.memzero`). For example, `testZeroGet` compiles to:

```llvm
define hidden i32 @main.testZeroGet(ptr dereferenceable_or_null(40) %m, i1 %s.b1, i32 %s.i, i1 %s.b2, ptr %context) unnamed_addr #1 {
entry:
  %hashmap.key = alloca %main.hasPadding, align 8
  %hashmap.value = alloca i32, align 4
  %s = alloca %main.hasPadding, align 8
  %0 = insertvalue %main.hasPadding zeroinitializer, i1 %s.b1, 0
  %1 = insertvalue %main.hasPadding %0, i32 %s.i, 1
  %2 = insertvalue %main.hasPadding %1, i1 %s.b2, 2
  %stackalloc = alloca i8, align 1
  store %main.hasPadding zeroinitializer, ptr %s, align 8
  call void @runtime.trackPointer(ptr nonnull %s, ptr nonnull %stackalloc, ptr undef) #3
  store %main.hasPadding %2, ptr %s, align 8
  call void @llvm.lifetime.start.p0(i64 4, ptr nonnull %hashmap.value)
  call void @llvm.lifetime.start.p0(i64 12, ptr nonnull %hashmap.key)
  store %main.hasPadding %2, ptr %hashmap.key, align 8
  %3 = getelementptr inbounds i8, ptr %hashmap.key, i32 1
  call void @runtime.memzero(ptr nonnull %3, i32 3, ptr undef) #3
  %4 = getelementptr inbounds i8, ptr %hashmap.key, i32 9
  call void @runtime.memzero(ptr nonnull %4, i32 3, ptr undef) #3
  %5 = call i1 @runtime.hashmapBinaryGet(ptr %m, ptr nonnull %hashmap.key, ptr nonnull %hashmap.value, i32 4, ptr undef) #3
  call void @llvm.lifetime.end.p0(i64 12, ptr nonnull %hashmap.key)
  %6 = load i32, ptr %hashmap.value, align 4
  call void @llvm.lifetime.end.p0(i64 4, ptr nonnull %hashmap.value)
  ret i32 %6
}
```

Also, it might be worth testing an array as well (it could just be a `[2]hasPadding`).

aykevl (commented):
Have you tried running a few tests under LLVM 14? LLVM 14 still requires pointer types to match, so I think you will hit a few validation errors. (We don't really have such a test in CI, although I think it would be caught if a padded struct gets added to testdata/map.go - maybe that's a good idea anyway).

(The "Suggestion" comments are really just that, style suggestions you can apply if you like them).

dgryski:
Trying to get the `CreateBitCast` stuff to work and I keep getting `Invalid bitcast`, so that's what I'm fighting with today.

dgryski:
OK.  Switched to `CreateInBoundsGEP`, switched the upper limit of the for loop to `numFields`, simplified the unit test, combined last-field zero logic, added `zero`s to the GEP calls. I couldn't figure out how to use the `CreateBitCast` without causing verification failures. <phew>.

aykevl:
Did a quick check and this PR as-is is indeed not working in LLVM 14:

```
  %5 = getelementptr inbounds i8*, i1* %4, i64 1, !dbg !43
 i8*  call void @runtime.memzero(i8** %5, i64 7, i8* undef), !dbg !43
```

Will take a more in-depth look later.
I'm pretty sure you can reproduce this on MacOS (`brew install llvm@14`, `go install -tags=llvm14`).

dgryski:
I have LLVM 14 from checking out the reflect fixes.  I'll look at this now.

aykevl:
Didn't fully test this but it seems to fix all the verification failures for me in LLVM 14:

```diff
--- a/compiler/map.go
+++ b/compiler/map.go
@@ -325,7 +325,14 @@ func (b *builder) zeroUndefBytes(llvmType llvm.Type, ptr llvm.Value) error {
                        if fieldEndOffset != nextOffset {
                                n := llvm.ConstInt(b.uintptrType, nextOffset-fieldEndOffset, false)
                                llvmStoreSize := llvm.ConstInt(b.uintptrType, storeSize, false)
-                               paddingStart := b.CreateInBoundsGEP(b.i8ptrType, elemPtr, []llvm.Value{llvmStoreSize}, "")
+                               gepPtr := elemPtr
+                               if gepPtr.Type() != b.i8ptrType {
+                                       gepPtr = b.CreateBitCast(gepPtr, b.i8ptrType, "") // LLVM 14
+                               }
+                               paddingStart := b.CreateInBoundsGEP(b.ctx.Int8Type(), gepPtr, []llvm.Value{llvmStoreSize}, "")
+                               if paddingStart.Type() != b.i8ptrType {
+                                       paddingStart = b.CreateBitCast(paddingStart, b.i8ptrType, "") // LLVM 14
+                               }
                                b.createRuntimeCall("memzero", []llvm.Value{paddingStart, n}, "")
                        }
                }
```

aykevl:
Explanation:

> `if gepPtr.Type() != b.i8ptrType {`

Check for LLVM 14. In LLVM 14, every pointer is different (with different element types), while in LLVM 15 all pointers in the same address space are the same.

> `gepPtr = b.CreateBitCast(gepPtr, b.i8ptrType, "")`

Convert to `i8*`.

dgryski:
PTAL.

aykevl (approved):

PR #3439, opened 2023-02-11
ci: don't link with libzstd in release builds

aykevl:
libzstd was added in LLVM 15, but we don't currently use it. So let's disable it in LLVM just like libzlib.

See: https://reviews.llvm.org/D128465


dgryski (approved):
LGTM

dgryski:
CI still having `wasmtime` issues

Issue #3394, opened 2023-01-17
Serial print stops working after SPI configure on pico target using a picoprobe


incredible-io:
Hello.

I have a ST7735 display connected to SPI1 at a pico and this pico connected with SWD to another pico to be able to debug my code. The following minimal sample code reproduce this issue:
```golang
package main

import (
	"machine"
	"time"
)

func main() {
	time.Sleep(time.Second * 5)
	println("Hello, TinyGo!") // This print works, 
                                  // but the last character (\n) seems not to be printed, if
                                  // directly followed by the next line without time.Sleep()

	if err := machine.SPI1.Configure(machine.SPIConfig{
		SCK: machine.SPI1_SCK_PIN,
		SDO: machine.SPI1_SDO_PIN,
	}); err != nil {
		panic(err)
	}

	for {
		print(".") // These prints are not shown in my serial monitor, 
                           // but the loop is running
		time.Sleep(time.Second)
	}
}
```
I'm using this command to flash the picoprobe:
```
$> tinygo flash -target pico -programmer picoprobe -serial uart .
```
If I comment out the block with `machine.SPI.Configure()`, all prints from the for-loop are shown in my serial monitor. So in my real code I cannot use print for logging status messages, and so on, if I use my display at the same time. If I use only one pico without the debug-probe-setup from the handbook, the sample code runs fine, also with configured SPI.

So I have no idea what I'm missing, or maybe what is wrong with the picoprobe setup?

The wiring is like in [Appendix A](https://datasheets.raspberrypi.com/pico/getting-started-with-pico.pdf) plus the wiring from the [display](https://www.waveshare.com/wiki/Pico-LCD-1.8).

I'm using TinyGo version:
```
$> tinygo version
tinygo version 0.26.0 linux/amd64 (using go version go1.19.5 and LLVM version 14.0.0)

$> tinygo info pico
LLVM triple:       thumbv6m-unknown-unknown-eabi
GOOS:              linux
GOARCH:            arm
build tags:        cortexm baremetal linux arm rp2040 rp pico tinygo math_big_pure_go gc.conservative scheduler.tasks serial.usb // <- this is corrected within the flash command, right?
garbage collector: conservative
scheduler:         tasks
cached GOROOT:     /home/.../.cache/tinygo/goroot-...
```

Regards,
Cid

soypat:
[...]

PR #3443, opened 2023-02-12
rp2040: provide better errors for invalid pins on i2c and spi

soypat:
As title says, it should be easier to get better errors on invalid pin configuration. Would prevent issues like https://github.com/tinygo-org/tinygo/issues/3394

### DIFF
--- a/src/machine/machine_rp2040_i2c.go
+++ b/src/machine/machine_rp2040_i2c.go
@@ -90,7 +90,7 @@ func (i2c *I2C) Tx(addr uint16, w, r []byte) error {
 //	SCL: 3, 7, 11, 15, 19, 27
 func (i2c *I2C) Configure(config I2CConfig) error {
 	const defaultBaud uint32 = 100_000 // 100kHz standard mode
-	if config.SCL == 0 {
+	if config.SCL == 0 && config.SDA == 0 {
 		// If config pins are zero valued or clock pin is invalid then we set default values.
 		switch i2c.Bus {
 		case rp.I2C0:
@@ -100,6 +100,20 @@ func (i2c *I2C) Configure(config I2CConfig) error {
 			config.SCL = I2C1_SCL_PIN
 			config.SDA = I2C1_SDA_PIN
 		}
+	}
+	var okSDA, okSCL bool
+	switch i2c.Bus {
+	case rp.I2C0:
+		okSDA = config.SDA%4 == 0
+		okSCL = (config.SCL+1)%4 == 0
+	case rp.I2C1:
+		okSDA = (config.SDA+2)%4 == 0
+		okSCL = (config.SCL+3)%4 == 0
+	}
+	if !okSDA {
+		return errors.New("invalid I2C SDA pin")
+	} else if !okSCL {
+		return errors.New("invalid I2C SCL pin")
 	}
 	if config.Frequency == 0 {
 		config.Frequency = defaultBaud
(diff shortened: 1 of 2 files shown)


src/machine/machine_rp2040_i2c.go
```diff
@@ -101,6 +101,20 @@ func (i2c *I2C) Configure(config I2CConfig) error {
...
+		okSCL = (config.SCL+1)%4 == 0
+	case rp.I2C1:
+		okSDA = (config.SDA+2)%4 == 0
+		okSCL = (config.SCL+3)%4 == 0
+	}
+	if !okSDA {
+		return errors.New("invalid I2C SDA pin")
+	} else if !okSCL {
+		return errors.New("invalid I2C SCL pin")
+	}
```
aykevl: Please define these as globals. That avoids a heap allocation.

src/machine/machine_rp2040_spi.go
```diff
@@ -175,6 +175,24 @@ func (spi SPI) Configure(config SPIConfig) error {
...
+		okSDO = config.SDO == 11 || config.SDO == 15
+		okSDO = config.SCK == 10 || config.SCK == 14
+	}
+	if !okSDI {
+		return errors.New("invalid SPI SDI pin")
+	} else if !okSDO {
+		return errors.New("invalid SPI SDO pin")
+	} else if !okSCK {
+		return errors.New("invalid SPI SCK pin")
+	}
```
aykevl: Same here.

aykevl (commented):
Good idea. Just a suggestion to make the code a little bit more efficient.

soypat:
@aykevl I've added your suggested fixes

src/machine/machine_rp2040_i2c.go
```diff
@@ -57,6 +57,8 @@ var (
 	ErrInvalidTgtAddr     = errors.New("invalid target i2c address not in 0..0x80 or is reserved")
 	ErrI2CGeneric         = errors.New("i2c error")
 	ErrRP2040I2CDisable   = errors.New("i2c rp2040 peripheral timeout in disable")
+	errInvalidI2CSDA      = errors.New("invalid I2C SDA pin")
```
deadprogram, line 60: These should be exported if the developer wants to compare against them for specific error handling, right?

soypat: I'm not sure in what situation a user would want to handle an error regarding chosen pins. Seems like that would be always chosen as a constant literal in the program setup. Might it be unecessary noise to the machine package?

deadprogram: Since all the other errors are exported, shouldn't we be consistent? But you are probably correct about the nature of the error.

soypat: I'm unsure of the benefits of being consistent in this case. Every error we add here decreases the portability of the code since this is a board specific file.

There's also plenty of examples in the Go standard library where there's a distinction made between what should and should not be made an exported error. Keep in mind these are errors that could be exported in the future if the need ever arised, as has happened recently with the "deadline exceeded" error in the net package, I believe.
image

aykevl: I think not exporting this error is fine. The only thing you can do with this error is show it to the developer so they can fix the code. It's not something the application should handle.

In addition, every thing that is not exported can be exported without breaking the API. Once it is exported, changing it to be non-exported is an API break. So starting out by making it unexported is always safe.

aykevl (approved):

deadprogram:
Thanks @soypat now squash/merging.

PR #3452, opened 2023-02-16
Netdev2

scottfeldman:
This PR adds a network device driver model called netdev.  There will be a companion PR for TinyGo drivers to update the netdev drivers and network examples.  This PR covers the core "net" package.

An RFC for the work is here: #https://github.com/tinygo-org/drivers/issues/487.  Some things have changed from the RFC, but nothing major.

The "net" package is a partial port of Go's "net" package, version 1.19.3.  The src/net/README file has details on what is modified from Go's "net" package.

Most "net" features are working as they would in normal Go.  TCP/UDP/TLS protocol support is there.  As well as HTTP client and server support.  Standard Go network packages such as golang.org/x/net/websockets and Paho MQTT client work as-is.  Other packages are likely to work as-is.

Testing results are [here](https://docs.google.com/spreadsheets/d/e/2PACX-1vT0cCjBvwXf9HJf6aJV2Sw198F2ief02gmbMV0sQocKT4y4RpfKv3dh6Jyew8lQW64FouZ8GwA2yjxI/pubhtml?gid=1013173032&single=true).


src/net/net.go
```diff
@@ -7,10 +7,19 @@
 package net
 
 import (
-	"io"
 	"time"
+
+	"tinygo.org/x/drivers/netdev"
```
bgould: I think maybe this shouldn't be referencing the drivers module; it might be preferable to have a copy of the netdev.Netdever interface in TinyGo instead of in `drivers` (or alternatively, have the primary definition be in TinyGo and referenced from `drivers`).

src/net/udpsock.go
```diff
@@ -0,0 +1,249 @@
...
+		return nil, fmt.Errorf("Lookup of host name '%s' failed: %s", host, err)
+	}
+
+	return &UDPAddr{IP: ip[:4], Port: port}, nil
+}
+
+// UDPConn is the implementation of the Conn and PacketConn interfaces
+// for UDP network connections.
+type UDPConn struct {
+	fd            netdev.Sockfd
```
bgould: Similar comment as for netdev.Netdever ... Sockfd should maybe be defined in the TinyGo repo instead of `drivers`.  Also wondering if Sockfd should be `uintptr` instead of `int` (in `os` for instance file descriptors are `uintptr`)

bgould:
Great work, this is very exciting :)

src/net/net.go
scottfeldman: I had it that way at one point, but changed it as to not add any new exports to "net" package.

src/net/udpsock.go
scottfeldman: int was what socket(2) calls return, so just copied that.  And -1 is a common error return value for socket(2).

deadprogram:
Seems like some dependency problem here @scottfeldman https://github.com/tinygo-org/tinygo/actions/runs/4190521808/jobs/7264005556#step:13:32

scottfeldman:
@deadprogram I'm stuck.  Both netdev PRs for tinygo and drivers builds are failing because they depend on each other.  tinygo needs to import drivers/netdev and drivers needs to import tinygo "net" for net.useNetdev.  These build fine in my go workspace where they're both present.  I'm at a loss on what to do.

scottfeldman:
@deadprogram I started over in a fresh go workspace with a fresh clone of tinygo and drivers,  checked out netdev branch for each and built everything again.  Ran make test on tinygo and drivers and both PASS.

It seems both PRs need to be merged at the same time and folks would have to refresh both to avoid hitting the build issue we're hitting here.

src/net/net.go
deadprogram: I understand the motivation, but we for sure do not wish for the circular reference. Either use a subpackage, just add another file with the exported interface, or copy the interface definition into both the tinygo repo and also a duplicate the drivers package.

scottfeldman: Ok, new commit moving drivers/netdev to tinygo/src/net/netdev.  This should decouple "net" and "netdev" and remove the circular reference.

src/net/netdev/socket.go
```diff
@@ -0,0 +1,143 @@
...
+
+// sockaddr_in
+type SockAddr struct {
+	host string
+	port Port
+	ip   IP
+}
+
+func NewSockAddr(host string, port Port, ip IP) SockAddr {
+	return SockAddr{host: host, port: port, ip: ip}
```
soypat: A lot of these types are shared by the Go standard library. Maybe it's best to reuse them? `net.IP` and friends I mean. `net.IP` is a slice of bytes, If we keep this whole package outside of `net` we can simply use the existing `net` package for all these types. Port is also represented as an `int` in `net` (see `TCPAddr`).

src/net/netdev/netdev.go
```diff
@@ -0,0 +1,133 @@
...
+package netdev
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+type HardwareAddr []byte
```
soypat: `net` already has a `HardwareAddr`! Let's use it! I understand you are not using it here to avoid a cyclic dependency- well that's another reason to keep `netdev` in its own external package for now, either in `drivers` or `tinyio` when https://github.com/tinygo-org/drivers/issues/491 is approved, if it is approved.

src/net/net.go
```diff
@@ -7,10 +7,18 @@
...
-	"io"
+	"net/netdev"
 	"time"
 )
 
+// dev is the current TinyGo network device
+var dev netdev.Netdever
+
+// UseNetdev sets the netdev
+func UseNetdev(n netdev.Netdever) {
```
soypat: Again, this should not be exported. Link at compile time! Our objective should be to maintain compatibility during the rollout of this shiny brand new feature. OVer time we will learn how it works and what we can do to make it better, maybe even adding in this function as exported. But best to play it safe for now. Unexporting it later may break programs.

src/net/tlssock.go
```diff
@@ -0,0 +1,156 @@
...
+
+	return &TLSConn{
+		fd: fd,
+	}, nil
+}
+
+// A TLSConn represents a secured connection.
+// It implements the net.Conn interface.
+type TLSConn struct {
+	fd            netdev.Sockfd
```
soypat: If a Sockfd is just a plain old integer lets use a `int32` or similar for now and call all instances of it `fd*` in code.

src/net/udpsock.go
```diff
@@ -0,0 +1,248 @@
...
+	if laddr.Port == 0 {
+		laddr.Port = ephemeralPort()
+	}
+
+	fd, err := dev.Socket(netdev.AF_INET, netdev.SOCK_DGRAM, netdev.IPPROTO_UDP)
+	if err != nil {
+		return nil, err
+	}
+
+	var ip netdev.IP
```
soypat: Can this be just a plain `net.IP`?

src/net/netdev/socket.go
```diff
@@ -0,0 +1,143 @@
...
+	Socket(family AddressFamily, sockType SockType, protocol Protocol) (Sockfd, error)
+	Bind(sockfd Sockfd, myaddr SockAddr) error
+	Connect(sockfd Sockfd, servaddr SockAddr) error
+	Listen(sockfd Sockfd, backlog int) error
+	Accept(sockfd Sockfd, peer SockAddr) (Sockfd, error)
+	Send(sockfd Sockfd, buf []byte, flags SockFlags, timeout time.Duration) (int, error)
+	Recv(sockfd Sockfd, buf []byte, flags SockFlags, timeout time.Duration) (int, error)
+	Close(sockfd Sockfd) error
+	SetSockOpt(sockfd Sockfd, level SockOptLevel, opt SockOpt, value any) error
+}
```
soypat: ## On the `Socketer`'s method types
1. Keep as is with own types.

2. Use `net` types where possible or primitive types everywhere else. I like this one because its simple. Berkeley sockets are well known. You can define the Constants in the `net` package for the SockOptLevel, SockFlag, Protocol, etc. 
  ```go
  type Sockfd = int32
  type SockFlags = int
  type SockOptLevel = int
  type SockOpt = int
  // Decompose SockAddr
  type IP = net.IP 
  type Port = int    // like net package
  ```
  You can interchange some `int`s for `uint`s if you know they are only positive to give even better type hints to users.

3. Find a middle ground. I've seen that some types have `String` methods. That is fantastic for debuggability and developer happiness. We should probably keep that. That said- `SockAddr` should definetely be a `ip net.IP` and a `port int`, just like the standard library. `Sockfd` looks like it should be a `int32`, there is not much to be gained from giving it a type. It's also pretty apparent what is and isn't a FD from looking at bare source code (usually called `fd` or `fd*`). If we can't add meaningful methods to SockOptLevel, SockFlags, SockOpt, then they or some of them should be integers of fixed size according the the berkeley API.

src/net/netdev/netdev.go
```diff
@@ -0,0 +1,133 @@
+// Netdev is TinyGo's network device driver model.  TinyGo's "net" package
+// interfaces to the netdev driver directly to provide TCPConn, UDPConn,
+// and TLSConn socket connections.
+
+package netdev
```
soypat: I am not in favor regarding the addition of a package that would break native-tinygo compatibility.

soypat (changes requested):
I really appreciate the work you've put into this whole PR. This is exactly where I hoped TinyGo would be in the future a while ago and now it's here- it's unreal. Now being so close to the finish line I really want to take the time to really sit down and make sure we take TinyGo in the right direction- and that may take some thinking.

Right now I'm thinking this new package is not quite what I envisioned for TinyGo. I don't think the netdev package should live inside the standard library unless native Go incorporated it. IT would fundamentally split the ecosystem. We should strive to keep Gophers and Tinygophers united if possible.

Maybe adding this package is the only way of implementing Netdev... but I think it isn't. In fact, I'm quite sure we can do a better job at keeping compatibility with upstream/native. More comments to follow.

Lastly I want to link an issue that is related to this work https://github.com/tinygo-org/drivers/issues/491. This RFC is for the creation of a `tinyio` package that would solve many of the dependency cycle issues we are having, including in the `machine` package. I believe `netdev` would be a perfect fit as a subpackage of `tinyio` since it is neither a driver nor a stdlib package, but rather a bridge between the two. i.e. an interface, which is what `tinyio` seeks to standardize.

src/net/netdev/socket.go
scottfeldman: Doing so would create a circular dependency...

"net" imports "netdev"
"netdev" imports "net"

One way to work around this is to put everything in the "net" package.   So netdev.go and socket.go would just be part of the "net" package; there would be no netdev package.  I'm not a fan of this solution as it adds all those netdev exports to the net package.  The current PR has just the one export added to net: net.UseNetdev() and netdev package does not import "net".

Keeping netdev not dependent on net makes it easier to move its location around (current at net/netdev), and let's one use netdev without net package, if desired.  See rawsocket example in this PR.  It doesn't use "net" at all.

src/net/netdev/netdev.go
scottfeldman: Same circular dependency issue.

src/net/net.go
scottfeldman: Hmmm...let me check if the //go:linkname trick you introduced would work when the netdev package lives in tinygo.  That would hide the UseNetdev() export from net package.

Let me play around with it and see...

src/net/netdev/socket.go
soypat: Snap, you're right. Then probably better just to use a `[]byte` slice for IP and an int for port. We should still go back to how we were doing it when we got the wifinina working. Can we go that way without causing circular dependency? That and also avoid exporting `UseNetdev` from `net`'s side.

src/net/README.md
```diff
@@ -0,0 +1,473 @@
...
+
+## Using "net/http" Package
+
+TinyGo's net/http package is a partial port of Go's net/http package, providing
+a subset of the full net/http package.  There are a few features excluded
+during the porting process, in particular:
+
+- No HTTP/2 support
+- No TLS support for HTTP servers (no https servers)
+- HTTP client request can't be reused
```
aykevl: Can you explain why replacing net/http is necessary?

aykevl (commented):
I agree we really need to have networking support, but I'm worried the code in this PR is going to be very difficult to maintain. Currently we need a few changes here and there to support a new Go version (especially when that Go version adds new language features), but so far we managed to avoid having to pull code from upstream on each release. This PR would change that in a large way.

In particular, I wonder whether we could keep using the Go stdlib version of packages like net/http and crypto/tls, and if that's not appropriate for microcontrollers maybe have a separate package with lower resources (outside the standard library)?
This is especially important for net/http:
1. I expect there are many people who would like to have a fully featured net/http package (with TLS and all that).
2. If we copy+modify upstream net/http that will also put a burden on us to fix any security issues that might arise. I'm not sure we're able to do that.

scottfeldman:
> Right now I'm thinking this new package is not quite what I envisioned for TinyGo. I don't think the netdev package should live inside the standard library unless native Go incorporated it. IT would fundamentally split the ecosystem. We should strive to keep Gophers and Tinygophers united if possible.

@soypat Hmmm, I think I'm missing something here.  I assumed TinyGo is downstream of Go with a direction arrow Go -> TinyGo.  I hadn't considered the opposite of TinyGo -> Go.  Are there examples in the past where TinyGo changes moved to upstream Go?  If the arrow can go both ways, netdev is not the right answer.

netdev would not make sense to integrate into upstream Go "net" package.  netdev is a hack.  It doesn't use the porting layer of Go's "net" package.  The porting layer is at the syscall interface to the OS.  We don't have an OS, so netdev slices above the porting layer, above the syscall interface.  The good news is applications written to Go's "net" package just work in TinyGo (sans features that haven't been ported).

BTW, I tried slipping in netdev at the porting layer to try to retain as much or all of "net" as possible.  I ran into two issues: first, it makes huge binaries.  I was able to compile some of the examples and was getting images that were ~10x bigger in flash and in RAM compared to what we have in this PR.  The full "net" package is huge, mind-numbingly huge (to quote DA).  The second issue was TLS.  netdev takes advantage of these co-processors to offload the TLS protocol to hardware/firmware.  That means the crypto/tls code in Go is pretty much by-passed in favor of direct netdev calls to hardware/firmware.  Go's porting layer doesn't comprehend TLS offload; it assumes TLS protocol is handled in software, and transported over TCP.

So here we are.  This PR provides a TinyGo "net" package that is a subset of Go's.  Just enough of the "net" package *.go files are copied over to give us the three basic net.Conns: TCPConn, UDPConn, TLSConn, as well as most of the "net/http" package.  It is a maintenance burden (keeping sync'ed with upstream Go "net").  But it's awesome also.  The examples for MQTT and websocket use the standard packages built on top of "net", and they just work.  But this PR is no good if the goal is to unify TinyGo and Go's "net" packages.

src/net/README.md
scottfeldman: > I agree we really need to have networking support, but I'm worried the code in this PR is going to be very difficult to maintain. Currently we need a few changes here and there to support a new Go version (especially when that Go version adds new language features), but so far we managed to avoid having to pull code from upstream on each release. This PR would change that in a large way.

There is a maintenance burden, but I didn't see it as difficult.  There are 10 files under src/net, 5 are straight copies, the other 5 are modified.  Someone needs periodically to back-port upstream changes for bug/security fixes.  As part of this PR, I would offer to maintain these files.  Maybe it's just because I've been thru these 10 files so much, it seems easy to maintain.

> In particular, I wonder whether we could keep using the Go stdlib version of packages like net/http and crypto/tls, and if that's not appropriate for microcontrollers maybe have a separate package with lower resources (outside the standard library)? This is especially important for net/http:

tls/crypto: no.  Go's tls/crypto is software support for TLS protocol.  We have mbed TLS offload where the co-controller (e.g. ESP32) handles the TLS protocol.  netdev goes straight to the offload model.

net/http: good luck, I guess is my first thought.  I've tried multiple times to get the full net/http code working without modifications.  Each time I abandon the effort because of too many things are missing in TinyGo.  I should have keep better notes so I could give you details.  I wouldn't mind revisiting if we can make some progress on this PR.  Other packages such as net/url, net/textproto, net/rpc, net/mail, net/smtp just work, as well as the golang.org/x/net (websocket) packages and third party packages (Paho MQTT, etc).

> 1. I expect there are many people who would like to have a fully featured net/http package (with TLS and all that).

There is TLS for net/http, but only on the client side.  So client http.Get/Head/Post/PostForm() calls work for http or https (TLS).  Http server TLS support is not there.  So no http.ListenAndSeverTLS(), but http.ListenAndServe() is OK.

It should be possible to add https server support, if someone needed it.  The mbed TLS offload on these co-controllers we're using can support it.  I'm inclined to wait and see if someone actually needs that support.  I guess it would be cool to run an https on a micro-controller with let's encrypt and plug it directly into the internet?  Are people running internet services on micro-controllers?   My thinking was these embedded devices live on the edge of the internet, not on the interior.  But there is a path forward to move from the edge.

> 2. If we copy+modify upstream net/http that will also put a burden on us to fix any security issues that might arise. I'm not sure we're able to do that.

There are 17 files in net/http, 9 of which are straight copies, and the other 8 are modified.  I can make same offer to maintain these files wrt security and other upstream changes.

src/net/netdev/socket.go
scottfeldman: option 2 is out because of circular dep.

Sockfd is opaque to client code; client code shouldn't understand what's inside or how big it is.  It's passed as an opaque token.

I find user-defined types self-documenting and I like the compiler checks I gain with user-defined types, so I'm having a hard time backing away from that.  Is there a performance (size, time) issue using user-defined types?  I don't believe there is any penalty in the hot path (send/recv).

scottfeldman:
One option is to move everything to drivers and have an empty "net" package in tinygo.  Apps that import "net" would now import "tinygo.org/x/drivers/net", same for "net/http".  (Kind of what's done now).  An app that imports just "net" will get an empty dir compile failure or something, and stick a src/net/README directing users to drivers/net.

That leaves tinygo's "net" namespace available in case something better comes along.

Would this help?

src/net/README.md
bgould: Sort of talking out of my assumptions here ... so feel free to shoot it down if this is way off base  ... but I wonder it _might_ be feasible use the standard `net` package by linking implementations of syscalls that dispatch to `netdev` implementations?

As an example of what I mean ... NestedVM was a compiler for producing Java binaries from C/C++ code ... in that case, POSIX functionality was integrated using [newlib](https://sourceware.org/newlib/) ... newlib allows users to implement a callback for dispatching syscalls including network ones such as: https://github.com/bgould/nestedvm/blob/master/src/org/ibex/nestedvm/UnixRuntime.java#L174-L185

Not sure if this is feasible ... but if so, I would imagine that the relevant syscall definitions in standard Go should be relatively stable ... almost certainly more stable that the `net` package.

src/net/net.go
soypat: If we do want to have netdev package inside of `Tinygo` it should be unreachable by external users. You can do this by putting `netdev` directory under another directory called `internal`. See [Dave Cheney's](https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface) blog post.

src/net/netdev/netdev.go
soypat: As I mentioned above, we can put `netdev` under an internal directory as [per this link](https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface).

src/net/README.md
soypat: While I do see value in merging with upstream Go that may be a herculean task we're asking for. I think we should strive to get this as polished as possible and completely Go-compatible as possible so that in the future we can merge TinyGo net package with upstream net package with not so much as a hiccup. This would mean eliminating any exported function signatures that are not in upstream Go and keeping the Application usage the same, which I believe is identical, if not very similar.

soypat:
~It's fine @scottfeldman you can put `netdev` under a internal directory for now (see [link](https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface), same as above).~ Actually cant be done.

> I assumed TinyGo is downstream of Go with a direction arrow Go -> TinyGo. I hadn't considered the opposite of TinyGo -> Go. Are there examples in the past where TinyGo changes moved to upstream Go? If the arrow can go both ways, netdev is not the right answer.

It's not out of the question, as I believe TinyGo is gaining a lot of attention. That said it is highly unlikely changes would be merged upstream, especially this one which is a "hack".

> netdev would not make sense to integrate into upstream Go "net" package. netdev is a hack. It doesn't use the porting layer of Go's "net" package.

Agreed. But you are still including it in the form of a subpackage accesible to users, this is the same as if it were in upstream Go which can then break compatibility. We should avoid that by using the `internal` package functionality. See link above.

> BTW, I tried slipping in netdev at the porting layer to try to retain as much or all of "net" as possible.

Getting this working would be gold but I think it's a non-goal for this PR. We can strive for keeping the API identical with `net` package and Application usage identical that way we assume nothing about what the final solution to this `net` package dilemma will be (maybe using syscalls like BCG says or other solution).

> One option is to move everything to drivers and have an empty "net" package in tinygo. [...] That leaves tinygo's "net" namespace available in case something better comes along.

We should only do this if we can not maintain an **identical exported API** signature with upstream Go. As far as I can tell it is possible but requires the compiler `//go:linkname` directive and ~using the `internal` package.~ We should really make sure we are not slipping in exported API that is not in upstream Go IMHO so that we can move to a different solution in the future if something would require it.

soypat:
:facepalm: @scottfeldman   I just noticed if we put the `netdev` behind an `internal` package developers won't be able to develop. I have to think about this some more...

soypat:
I'll throw this possible solution I've come up with out there:


### Define the `socketer` interface in `net` but don't export it
This *sounds* like an idiomatic solution. 

1. Define `net.socketer` interface similarly to how you've done it but using standard library primitive types or existing `net` **exported** types. Do the same with `net.netdev` interface, which could be simply `net.dev` to avoid stutter.

2. Define `net.useDev` package function that takes a `net.dev` as an argument.

3. When one writes drivers one implements the `net.dev` interface which is doable since all of its types are primitive or `net` types. 

4. There could be a helper package called `netdev` somewhere outside of tinygo that helps users link their driver with the `net.useDev` package function using the `//go:linkname` compiler directive. i.e: `netdev.Use`. `netdev`  would have a interface mirroring the one in `net` (`net.dev`). 

### What this solves:
1. Tinygo <-> Go compatibility in full
2. Cyclic dependencies. Now there are none.
3. Developers can develop using external `netdev` package that mirrors `net.dev` interface

### Drawbacks
1. `net.socketer`/`netdev.Socketer`  interface is now composed of primitive types which impacts developer friendliness in a negative way.

src/net/README.md
scottfeldman: > Sort of talking out of my assumptions here ... so feel free to shoot it down if this is way off base ... but I wonder it _might_ be feasible use the standard `net` package by linking implementations of syscalls that dispatch to `netdev` implementations?

This would be the ultimate solution.  Replacing syscall/sockets with netdev/sockets at the portability layer in the net package.  I tried that route and didn't get very far.  I saved the branch in case I wanted to revisit, but if I recall I was bumping into things that were not available in tinygo.  Even the crypto/tls package should port, negating the need to offload TLS, assuming we can do all the crypto required on these controllers.  Let's say all of this is possible, and someone can pull this off in the future.  What would be the migration path from this PR to the ultimate solution?  Let's look at the pieces:

--------

#### Apps
In this PR, apps import "net", "net/http" and "crypto/tls", same as they would if running in normal Go env.  Nothing should change from the apps' perspective when migrating to the ultimate solution.  Users are unaware the rug has been pulled out, upgraded, and slipped back in.

#### Drivers
In this PR, drivers implement the netdev interface, which is basically the BSD socket(2) interface.  The ultimate solution ports the "net" package to netdev rather than syscall/sockets, and there is a good chance nothing changes in the netdev interface during this port, so nothing changes in the drivers.  (BSD sockets will probably be with us forever).  We retain our investment in drivers.

#### Interfaces
There is a loose end with net.UseNetdev() where the app must pick which netdev driver to use.  We don't really have support for network interfaces (eth0, wlan0, etc); we're hard wiring the driver to the app with UseNetdev().  If the ultimate solution can mimic the full OS network interfaces as defined in the "net" package, then maybe UseNetdev moves outside of the apps visibility.  So there would be a change required to the app.  But to support a full OS model of interfaces, there would need to be something equivalent to the "ip" command in Linux to configure and manage the interfaces.  Seems way overkill for embedded.

--------

So I'm feeling pretty good about a migration path forward from this PR.  I'm skeptical the full "net" package could be ported at the syscall level, but just in case it can, I think we have a good migration story.  

If someone wants to take a go at a full "net" package port, this PR would be a good starting point as the netdev driver parts are done.  In my attempt, I started with an empty tinygo/src/net directory and tried to compile one of the drivers/examples/net examples.  Adding/modifying "net" package files as needed to get a clean compile.  Adding xxx_tinygo.go files where the xxx_[posix|linux|unix|windows].go files reside, and so on.

scottfeldman:
> I'll throw this possible solution I've come up with out there:
> 
> ### Define the `socketer` interface in `net` but don't export it
> This _sounds_ like an idiomatic solution.
> 
> 1. Define `net.socketer` interface similarly to how you've done it but using standard library primitive types or existing `net` **exported** types. Do the same with `net.netdev` interface, which could be simply `net.dev` to avoid stutter.
> 2. Define `net.useDev` package function that takes a `net.dev` as an argument.
> 3. When one writes drivers one implements the `net.dev` interface which is doable since all of its types are primitive or `net` types.

You lost me here.  How does a driver in the drivers repo get access to a non-exported interface in the tinygo repo?  The driver has to be written to some interface.

Maybe you can pull this PR (and the related driver one) and experiment with different options?  The success criteria would be that the examples in drivers/examples/net still compile and run, including the rawsocket example.

scottfeldman:
Build is busted because of F_SETFL symbol.  F_SETFL was added to src/syscall/net.go, as well as some other missing syscall constants, to support netdev interface using native Go types/constants.  Unfortunately, F_SETFL is defined in syscall_libc_darwin.go and syscall_libc_wasi.go, so not sure where the real F_SETFL should live.

Would someone please advise where to put a single F_SETFL to fix this build issue?

soypat:
Alright, will review in the coming weeks- 11k lines of code is quite a bunch

scottfeldman:
Hi @soypat, pester you on the review...how's it going?

soypat:
My apologies @scottfeldman, I totally missed the updated version. Thanks for letting me know. I'll get to it ASAP

soypat:
> Unfortunately, F_SETFL is defined in syscall_libc_darwin.go and syscall_libc_wasi.go, so not sure where the real F_SETFL should live.

Scott, the way the go tool works is that it interprets files that end with `_windows.go`, `_darwin.go`, `_linux.go`,  etc. as having an applied build tag corresponding to the GOOS environment variable. Similarly `_386.go` and `_amd64.go` suffixed files will have the GOARCH build tag applied. See https://pkg.go.dev/go/build#hdr-Build_Constraints

scottfeldman:
@soypat 

I have some updates in my local tree to this PR.  Should I push those or wait until your review?  Changes are:

- rebase with dev
- fix this F_SETFL issue
- get the espat driver working

soypat:
Push em!

deadprogram:
Hello @scottfeldman looks like some actual CI errors like this https://github.com/tinygo-org/tinygo/actions/runs/4535422239/jobs/7990849307?pr=3452#step:20:34

scottfeldman:
I messed up my git repo for this PR, so I'm closing this PR and replacing it with a fresh one: https://github.com/tinygo-org/tinygo/pull/3614.

PR #3453, opened 2023-02-16
build: always cache LLVM source/build even if the tests fail to avoid extra rebuilds

deadprogram:
This PR modifies the GH actions builds to always cache LLVM source/build even if the tests fail to avoid extra rebuilds.

See https://github.com/actions/cache/tree/main/save#always-save-cache for details

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -29,11 +29,11 @@ jobs:
         with:
           go-version: '1.20'
           cache: true
-      - name: Cache LLVM source
-        uses: actions/cache@v3
+      - name: Restore LLVM source cache
+        uses: actions/cache/restore@v3
         id: cache-llvm-source
         with:
-          key: llvm-source-15-macos-v2
+          key: llvm-source-15-macos-v3
           path: |
             llvm-project/clang/lib/Headers
             llvm-project/clang/include
@@ -43,11 +43,22 @@ jobs:
       - name: Download LLVM source
         if: steps.cache-llvm-source.outputs.cache-hit != 'true'
         run: make llvm-source
-      - name: Cache LLVM build
-        uses: actions/cache@v3
+      - name: Save LLVM source cache
+        uses: actions/cache/save@v3
+        if: steps.cache-llvm-source.outputs.cache-hit != 'true'
+        with:
+          key: ${{ steps.cache-llvm-source.outputs.cache-primary-key }}
+          path: |
+            llvm-project/clang/lib/Headers
+            llvm-project/clang/include
+            llvm-project/compiler-rt
+            llvm-project/lld/include
+            llvm-project/llvm/include
+      - name: Restore LLVM build cache
+        uses: actions/cache/restore@v3
         id: cache-llvm-build
         with:
-          key: llvm-build-15-macos-v3
+          key: llvm-build-15-macos-v4
           path: llvm-build
       - name: Build LLVM
         if: steps.cache-llvm-build.outputs.cache-hit != 'true'
@@ -61,6 +72,12 @@ jobs:
           # build!
           make llvm-build
           find llvm-build -name CMakeFiles -prune -exec rm -r '{}' \;
+      - name: Save LLVM build cache
+        uses: actions/cache/save@v3
+        if: steps.cache-llvm-build.outputs.cache-hit != 'true'
+        with:
+          key: ${{ steps.cache-llvm-build.outputs.cache-primary-key }}
+          path: llvm-build
       - name: Cache wasi-libc sysroot
         uses: actions/cache@v3
         id: cache-wasi-libc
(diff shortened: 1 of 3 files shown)


ysoldak (approved):

aykevl (commented):
Yes yes yes! We had this in CircleCI but at the time it wasn't possible in GitHub Actions.

I don't think the path parameter is necessary for restore. Otherwise it looks good to me.

deadprogram:
The example I looked at in the docs here did that with the path, so I copied it: https://github.com/actions/cache#using-a-combination-of-restore-and-save-actions

deadprogram:
Once I have this working correctly for Windows, I will do the same thing for the other targets in a subsequent PR, just so we can get the Windows builds unstuck.

aykevl:
It doesn't include the cache here: https://github.com/actions/cache/tree/main/save#re-evaluate-cache-key-while-saving
Also, it doesn't make sense to include paths in the restore steps. It just restores whatever was saved. I think the example you used is wrong.

deadprogram:
OK I will remove that in the later PR just to let the current build try to finish.

deadprogram:
@aykevl turns out you do need that:

```
Input required and not supplied: path
```

aykevl:
> @aykevl turns out you do need that:
> 
> ```
> Input required and not supplied: path
> ```

Yeah you are correct, I later found out the path is required. Which doesn't make a lot of sense to me to be honest (it wasn't required in CircleCI).

deadprogram:
@aykevl just wonder if we should either remove the LLVM source cache tasks, or remove this?

https://github.com/tinygo-org/tinygo/blob/ba4553fce4d540c33c222d70b53d0c25f5687e35/.github/workflows/windows.yml#L75-L77

Seems like something not quite correct here.

aykevl:
> @aykevl just wonder if we should either remove the LLVM source cache tasks, or remove this?
> 
> https://github.com/tinygo-org/tinygo/blob/ba4553fce4d540c33c222d70b53d0c25f5687e35/.github/workflows/windows.yml#L75-L77
> 
> Seems like something not quite correct here.

I believe this is working as intended. The way it works now is as follows:
- Cache the bits of llvm-project that are used for all CI runs (which is a small subset of the entire llvm-project tree).
- Download the entire llvm-project directory (not just the cached subset) for LLVM builds.

This provides quick cache restores in the common case when LLVM isn't built (the llvm-project directory is large so caching only a subset provides a significant speedup. But when LLVM needs to be rebuilt, this cached version isn't good enough. It's possible to also cache the full LLVM source tree but as it's only needed for LLVM builds it won't help CI times much.

aykevl:
Some notes:
* I tried building with thread support many times (for example here: https://github.com/tinygo-org/tinygo/pull/2433) but didn't succeed. It works locally, but doesn't work in CI. Very frustrating.
* I doubt stack size is the real issue here. I'm suspecting some sort of memory corruption that is then presented as a stack overflow. Especially as it only happens sometimes.

deadprogram:
> I'm suspecting some sort of memory corruption that is then presented as a stack overflow. Especially as it only happens sometimes.

Finally passed, but seems like it fails more often then passes.

.github/workflows/build-macos.yml
```diff
@@ -43,11 +43,22 @@ jobs:
...
+        uses: actions/cache/save@v3
+        if: steps.cache-llvm-source.outputs.cache-hit != 'true'
+        with:
+          key: ${{ steps.cache-llvm-source.outputs.cache-primary-key }}
+          path: |
+            llvm-project/clang/lib/Headers
+            llvm-project/clang/include
+            llvm-project/compiler-rt
+            llvm-project/lld/include
+            llvm-project/llvm/include
```
aykevl, line 56: Why did you add this step?

.github/workflows/windows.yml
```diff
@@ -100,6 +123,12 @@ jobs:
     runs-on: windows-2022
     needs: build-windows
     steps:
+      - name: Configure pagefile
+        uses: al-cheb/configure-pagefile-action@v1.3
+        with:
+          minimum-size: 8GB
+          maximum-size: 24GB
+          disk-root: "C:"
```
aykevl, line 131: Was the page file too small? Or why did you add this?

.github/workflows/build-macos.yml
deadprogram: All of the LLVM source is cached and also all of the LLVM builds are cached for each of the build platforms which need it.

.github/workflows/windows.yml
deadprogram: The idea is to give more virtual memory via pagefile to the Windows build instances, since they have unused space on C: drive. See https://github.com/easimon/maximize-build-space#how-it-works for the inspiration.

Did it actually help? I cannot say really.

PR #3457, opened 2023-02-17
machine/rp2040: change calling order for device enumeration fix

deadprogram:
This PR changes the calling order for the device enumeration fix on the RP2040. I think it needs to be done first before anything else to work correctly.


sago35:
I will check later.

sago35 (approved):

deadprogram:

PR #3458, opened 2023-02-18
fix bad qt py pin assignment

andyshinn:
We were confused when testing out this board and our A0 ended up being on A1!


sago35 (approved):
Thank you for the correction.

https://learn.adafruit.com/adafruit-qt-py/pinouts

deadprogram:
Thank you for the fix @andyshinn and to @sago35 for the quick review.

PR #3461, opened 2023-02-18
main: support qemu-user debugging

aykevl:
Running binaries in QEMU (when debugging on Linux for example) did not work correctly as qemu-user expects the `-g` flag to be first on the command line before the program name. Putting it after will make it a command line parameter for the emulated program, which is not what we want.

I don't think this ever worked correctly. But now it works for me to debug issues with #3103:

Screenshot_20230218_201218

(Using lldb here as Arch Linux doesn't package gdb-multiarch. The bug turned out to be using the esp register on amd64 which of course won't work).


deadprogram:
Thanks @aykevl now merging.

PR #3462, opened 2023-02-18
build/macos, linux, windows: set the number of make jobs equal to the number of available cores

deadprogram:
This PR sets the number of make jobs equal to the number of cores available to instances being used for Github-hosted runners on macos, Linux, and Windows builds.

See https://github.com/github/codeql/discussions/11260 for a conversation about this topic.

See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources for the runner specs.

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -42,7 +42,7 @@ jobs:
             llvm-project/llvm/include
       - name: Download LLVM source
         if: steps.cache-llvm-source.outputs.cache-hit != 'true'
-        run: make llvm-source
+        run: make llvm-source -j3
       - name: Save LLVM source cache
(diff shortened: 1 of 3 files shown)


aykevl (commented):
In the vast majority of these cases (perhaps even all), setting the number of jobs won't help at all because they're serial. For example, `make llvm-source` is just a single git command and `make llvm-build` is similarly fully serial (and in fact, the `ninja` call within is already parallelized).

aykevl:
Looking a bit more closely, I think these are the only ones that could be parallelized:

- `make gen-device`
- `make release deb`

deadprogram:
Closing because I think this actually does not do anything like what I intended.

PR #3463, opened 2023-02-18
build/linux: correct issue with duplicate name for LLVM source cache for arm/arm64

deadprogram:
This PR contains a needed fix that came out of #3462 

The Github actions build for linux has an issue with a duplicate name for the LLVM source cache for arm and arm64. This PR correct that problem.


deadprogram:
I think it actually is correct right now. Closing.

PR #3465, opened 2023-02-19
arm64: fix register save/restore to include vector registers

aykevl:
Some vector registers must be preserved across calls, but this wasn't happening on Linux and MacOS. When I added support for windows/arm64, I saw that it required these vector registers to be preserved and assumed this was Windows deviating from the standard calling convention. But actually, Windows was just implementing the standard calling convention and the bug was on Linux and MacOS.

This commit fixes the bug on Linux and MacOS and at the same time merges the Go and assembly files as they no longer need to be separate.


dgryski (approved):
LGTM.

dgryski:
I assume we correctly save all needed registers in amd64?

deadprogram:

aykevl:
> I assume we correctly save all needed registers in amd64?

Well I hope so. This is just something I discovered while looking into a possible optimization of the stack switching code so decided to fix it even if there haven't been any bug reports regarding this.

PR #3467, opened 2023-02-19
ci: don't pass -v to `go test`

aykevl:
It can be difficult to find what went wrong in a test. Omitting -v should make it easier to see the failing tests and the output for them (note that output is still printed for tests that fail).

Just something I've been running into and I'd like to change. Comments welcome.


dgryski (approved):
Sure. Most of the time looking at the logs is just trying to find the one "FAIL".

deadprogram:
Having spent a lot of time with the CI process lately, I am totally sold. Thanks! Now merging.

PR #3470, opened 2023-02-20
random reflection improvements

dgryski:


dgryski:
This is in no way its final form; I'll pick and choose and construct better PRs, but this is my current WIP branch.

src/reflect/value.go
```diff
@@ -217,7 +217,15 @@ func (v Value) CanAddr() bool {
 }
 
 func (v Value) Addr() Value {
-	panic("unimplemented: (reflect.Value).Addr()")
+	if !v.CanAddr() {
+		panic("taking addr of un-addr-able value")
+	}
+
+	return Value{
+		typecode: PtrTo(v.Type()).(*rawType),
```
aykevl: Interfaces can be expensive, so what I usually do is create a separate function that works directly with `*rawType`. Something like this:

```suggestion
		typecode: pointerTo(v.typecode),
```

```go
func pointerTo(t *rawType) *rawType { ... }
func PointerTo(t Type) Type {
    return pointerTo(t.(*rawType))
}
```

src/reflect/value.go
```diff
@@ -389,10 +397,10 @@ func (v Value) Slice3(i, j, k int) Value {
 	panic("unimplemented: (reflect.Value).Slice3()")
 }
 
-//go:linkname maplen runtime.hashmapLenUnsafePointer
+//go:linkname maplen runtime.hashmapLen
```
aykevl: This change is probably not compatible with LLVM 14.

src/reflect/value.go
```diff
@@ -641,30 +649,102 @@ func (v Value) OverflowFloat(x float64) bool {
...
+	if v.Kind() != Map {
+		panic(&ValueError{Method: "MapKeys", Kind: v.Kind()})
+	}
+
+	// empty map
+	if v.Len() == 0 {
+		return nil
+	}
+
+	var keys []Value
```
aykevl: This avoids a lot of allocations by doing a single allocation of the correct size:

```suggestion
	keys := make([]Value, 0, v.Len())
```

src/reflect/value.go
```diff
@@ -794,7 +890,30 @@ func (v Value) Convert(t Type) Value {
...
+	if cap < 0 {
+		panic("reflect.MakeSlice: negative cap")
+	}
+	if len > cap {
+		panic("reflect.MakeSlice: len > cap")
+	}
+
+	var slice sliceHeader
+	slice.cap = uintptr(cap)
+	slice.len = uintptr(len)
```
aykevl, line 944: To be extra safe it might be a good idea to check whether these values actually fit:

```go
if int(uintptr(cap)) != cap { panic() }
```

This is only really relevant on AVR, and should be a no-op on other architectures (it should be trivial for the compiler to optimize this check away on 32-bit and 64-bit systems).

src/reflect/value.go
```diff
@@ -794,7 +890,30 @@ func (v Value) Convert(t Type) Value {
...
+	}
+	if len < 0 {
+		panic("reflect.MakeSlice: negative len")
+	}
+	if cap < 0 {
+		panic("reflect.MakeSlice: negative cap")
+	}
+	if len > cap {
+		panic("reflect.MakeSlice: len > cap")
+	}
```
aykevl: You could consider calling `runtime.slicePanic` here, just like the compiler does. It should reduce binary size a bit at the cost of the readability of panic messages.

src/reflect/value.go
```diff
@@ -863,10 +982,41 @@ func alloc(size uintptr, layout unsafe.Pointer) unsafe.Pointer
 //go:linkname sliceAppend runtime.sliceAppend
 func sliceAppend(srcBuf, elemsBuf unsafe.Pointer, srcLen, srcCap, elemsLen uintptr, elemSize uintptr) (unsafe.Pointer, uintptr, uintptr)
 
+//go:linkname hashmapStringGet runtime.hashmapStringGet
+func hashmapStringGet(m unsafe.Pointer, key string, value unsafe.Pointer, valueSize uintptr) bool
```
aykevl, line 1022: I'm not sure this will work in LLVM 14 because the type of the `m` parameter is a different pointer type.

aykevl (commented):
Looks very reasonable. Some early comments and suggestions below.

dgryski:
This is what I've been using to test:
```
package main

import (
        "fmt"
        "reflect"
)

type namedString string

type barStruct struct {
        QuxString string
        BazInt    int
}

type foobar struct {
        Foo string
        Bar barStruct
}

func main() {
        m := make(map[string]int)
        m["two"] = 2
        v := reflect.ValueOf(m)
        k := reflect.ValueOf("two")

        fmt.Println(v.MapIndex(k))

        v.SetMapIndex(k, reflect.ValueOf(3))
        fmt.Println(v.MapIndex(k))

        mapT := reflect.TypeOf(m)

        fmt.Println(mapT.Key().String())

        m["four"] = 4
        m["five"] = 5

        v.SetMapIndex(reflect.ValueOf("two"), reflect.Value{})

        keys := v.MapKeys()

        fmt.Println(keys)

        intm := make(map[int]string)

        intm[10] = "ten"
        intm[20] = "twenty"

        intv := reflect.ValueOf(intm)
        intv.SetMapIndex(reflect.ValueOf(30), reflect.ValueOf("thirty"))

        keys = intv.MapKeys()

        for _, k := range keys {
                fmt.Println(k.Interface())
        }

        it := v.MapRange()
        for it.Next() {
                fmt.Println(it.Key().Interface(), it.Value().Interface())
        }

        maptype := reflect.TypeOf(m)
        refmap := reflect.MakeMap(maptype)
        refmap.SetMapIndex(reflect.ValueOf("six"), reflect.ValueOf(6))
        six := refmap.MapIndex(reflect.ValueOf("six"))
        fmt.Println(six.Interface())

        named := namedString("foo")
        fmt.Println("string:", reflect.TypeOf(named).String())
        fmt.Println("name:", reflect.TypeOf(named).Name())

        var fb foobar
        fb.Bar.QuxString = "qux"

        reffb := reflect.TypeOf(fb)
        q := reffb.FieldByIndex([]int{1, 0})
        fmt.Println(q.Name)

        var ok bool
        q, ok = reffb.FieldByName("Foo")
        fmt.Println(q.Name, ok)

        q, ok = reffb.FieldByName("Snorble")
        fmt.Println(q.Name, ok)
}
```

compiler/interface.go
```diff
@@ -167,9 +171,18 @@ func (c *compilerContext) getTypeCode(typ types.Type) llvm.Value {
 		case *types.Basic:
 			typeFields = []llvm.Value{c.getTypeCode(types.NewPointer(typ))}
 		case *types.Named:
+			name := typ.Obj().Name()
+			var buf []llvm.Value
+
+			for _, b := range []byte(name) {
+				buf = append(buf, llvm.ConstInt(c.ctx.Int8Type(), uint64(b), false))
```
soypat, line 178: Another place where the buffer could be pre-allocated?

dgryski: Good point.

aykevl: I care a lot less about efficiency in the compiler :) but yes, this is another case where preallocating a buffer might be slightly faster. I suspect all the CGo calls are going to be much more expensive than the heap allocation, though.

dgryski:
Bumping the `cortex-m-qemu` stack size to `4096` causes it to pass.  Not sure if that's something we can do in general .

compiler/interface.go
```diff
@@ -167,9 +171,18 @@ func (c *compilerContext) getTypeCode(typ types.Type) llvm.Value {
...
+			}
+
 			typeFields = []llvm.Value{
-				c.getTypeCode(types.NewPointer(typ)), // ptrTo
-				c.getTypeCode(typ.Underlying()),      // underlying
+				c.getTypeCode(types.NewPointer(typ)),                   // ptrTo
+				c.getTypeCode(typ.Underlying()),                        // underlying
+				llvm.ConstInt(c.uintptrType, uint64(len(name)), false), // length
+				llvm.ConstArray(c.ctx.Int8Type(), buf),
```
aykevl, line 185: I _believe_ this has the same effect and may be a bit faster or more readable:

```suggestion
				llvm.ConstString(name, false),
```

compiler/interface.go
```diff
@@ -112,6 +112,8 @@ func (c *compilerContext) getTypeCode(typ types.Type) llvm.Value {
 			typeFieldTypes = append(typeFieldTypes,
 				types.NewVar(token.NoPos, nil, "ptrTo", types.Typ[types.UnsafePointer]),
 				types.NewVar(token.NoPos, nil, "underlying", types.Typ[types.UnsafePointer]),
+				types.NewVar(token.NoPos, nil, "len", types.Typ[types.Uintptr]),
+				types.NewVar(token.NoPos, nil, "name", types.NewArray(types.Typ[types.Int8], int64(len(typ.Obj().Name())))),
 			)
```
aykevl, line 117: You are missing the corresponding update to the comment at the top of src/reflect/type.go.

targets/cortex-m.json
```diff
@@ -8,7 +8,7 @@
 	"rtlib": "compiler-rt",
 	"libc": "picolibc",
 	"automatic-stack-size": true,
-	"default-stack-size": 2048,
+	"default-stack-size": 4096,
```
aykevl, line 11: :+1: for tests this is probably more appropriate.

src/runtime/chan.go
```diff
@@ -148,10 +148,8 @@ func chanLen(c *channel) int {
 	return int(c.bufUsed)
 }
 
-// wrapper for use in reflect
-func chanLenUnsafePointer(p unsafe.Pointer) int {
-	c := (*channel)(p)
-	return chanLen(c)
+func chanLenUnsafePtr(c unsafe.Pointer) int {
+	return chanLen((*channel)(c))
```
aykevl, line 152: Why did you change the name and remove the comment?

src/reflect/type.go
```diff
@@ -430,31 +447,45 @@ type structField struct {
 // Equivalent to (go/types.Type).Underlying(): if this is a named type return
 // the underlying type, else just return the type itself.
 func (t *rawType) underlying() *rawType {
-	if t.meta&flagNamed != 0 {
+	if t.isNamed() {
 		return (*elemType)(unsafe.Pointer(t)).elem
```
aykevl, line 451: Because the struct for named types has changed, you should use a new struct type (`namedType` for example) here. This won't affect amd64, but will affect AVR which doesn't align any struct fields and therefore will have `.elem` at a different offset.

src/reflect/type.go
```diff
@@ -430,31 +447,45 @@ type structField struct {
...
+func pointerTo(t *rawType) *rawType {
 	switch t.Kind() {
 	case Pointer:
 		panic("reflect: cannot make **T type")
 	case Struct:
-		return (*structType)(unsafe.Pointer(t.(*rawType))).ptrTo
+		return (*structType)(unsafe.Pointer(t)).ptrTo
 	default:
-		return (*elemType)(unsafe.Pointer(t.(*rawType))).ptrTo
+		return (*elemType)(unsafe.Pointer(t)).ptrTo
```
aykevl, line 477: Same here.

src/reflect/type.go
```diff
@@ -472,13 +503,21 @@ func (t *rawType) elem() *rawType {
 	switch underlying.Kind() {
 	case Pointer:
 		return (*ptrType)(unsafe.Pointer(underlying)).elem
-	case Chan, Slice, Array:
+	case Chan, Slice, Array, Map:
 		return (*elemType)(unsafe.Pointer(underlying)).elem
-	default: // not implemented: Map
+	default:
 		panic("unimplemented: (reflect.Type).Elem()")
```
aykevl, line 509: It's not unimplemented after this PR, so please change this panic message accordingly.

src/reflect/type.go
```diff
@@ -793,11 +855,24 @@ func (t *rawType) NumMethod() int {
...
 
 func (t *rawType) Name() string {
-	panic("unimplemented: (reflect.Type).Name()")
+	if t.isNamed() {
+		ntype := (*namedType)(unsafe.Pointer(t))
+		name := stringHeader{
+			data: unsafe.Pointer(&ntype.name),
+			len:  ntype.nlen,
+		}
+		return *(*string)(unsafe.Pointer(&name))
```
aykevl, line 864: `unsafe.String` would be best here, but as long as we support Go 1.19 and below we can't do that.

src/reflect/type.go
```diff
@@ -816,12 +891,52 @@ func (t rawType) PkgPath() string {
...
-func (t rawType) FieldByName(name string) (StructField, bool) {
-	panic("unimplemented: (reflect.Type).FieldByName()")
+func (t *rawType) FieldByName(name string) (StructField, bool) {
+	if t.Kind() != Struct {
+		panic(TypeError{"FieldByName"})
+	}
+
+	numField := t.NumField()
+
+	// This is incredibly inefficient
```
aykevl, line 901: Oof, yeah this is O(n²). This should probably be fixed eventually because that kind of performance is going to cause problems at some point.

src/reflect/value.go
```diff
@@ -217,7 +217,15 @@ func (v Value) CanAddr() bool {
...
 func (v Value) Addr() Value {
-	panic("unimplemented: (reflect.Value).Addr()")
+	if !v.CanAddr() {
+		panic("taking addr of un-addr-able value")
+	}
+
+	return Value{
+		typecode: pointerTo(v.typecode),
+		value:    unsafe.Pointer(&v.value),
+		flags:    v.flags,
```
aykevl, line 227: I think this is the bug you mentioned on Slack? Because I think you need to clear the indirect flag here:

```suggestion
		flags:    v.flags &^ valueFlagIndirect,
```

src/reflect/value.go
```diff
@@ -636,35 +644,136 @@ func (v Value) NumMethod() int {
...
+	keys := make([]Value, 0, v.Len())
+
+	it := hashmapNewIterator()
+	k := New(v.typecode.Key())
+	e := New(v.typecode.Elem())
+
+	for hashmapNext(v.pointer(), it, k.value, e.value) {
+		keys = append(keys, k.Elem())
+		k = New(v.typecode.Key())
+	}
```
aykevl, line 686: Took me a bit to understand this code but it looks correct.

src/reflect/value.go
```diff
@@ -636,35 +644,136 @@ func (v Value) NumMethod() int {
...
+	if v.Kind() != Map {
+		panic(&ValueError{Method: "MapIndex", Kind: v.Kind()})
+	}
+
+	// compare key type with actual key type of map
+	if key.typecode != v.typecode.key() {
+		panic(&ValueError{Method: "MapIndex"})
+	}
+
+	elemType := v.Type().Elem()
```
aykevl, line 701: Probably a bit more efficient by avoiding a conversion to an interface:

```suggestion
	elemType := v.typecode.Elem()
```

aykevl (commented):
I reviewed most of the PR, but didn't review value.go completely. But figured it would be useful to send review feedback already.

Assuming this gets more of encoding/json to work, can you also add some tests to testdata/json.go? Just a few lines to prove the new features work (the testdata/* files are run on more platforms than the reflect package tests so might catch more bugs).

src/runtime/chan.go
dgryski: Because I removed them all and when I put them back I ended up choosing a different name.  I'll make sure all the wrappers needed for reflect have `UnsafePointer` at at least a comment indicating that they're for reflect.

src/reflect/type.go
dgryski: Ok.  I was assuming that because `namedType` and `elemType` shared the same initial struct layout this would be safe.

src/reflect/value.go
dgryski: Oops, thanks.

src/reflect/type.go
aykevl: You are correct. I was assuming a different struct layout. No need to change anything here.

aykevl: One possible optimization to named types is to use the following struct layout:

```go
type namedType struct {
	rawType
	nlen       uint16
	ptrTo      *rawType
	underlying *rawType
	name       [1]byte
}
```

This saves a single pointer-sized integer on most architectures (32-bit and 64-bit) for each named type by storing `nlen` in the padding that's between `rawType` and `ptrTo`. But on the other hand, this might not be worth the complexity and can also be done at a later time.

(I was assuming this optimization was used here, in which case my original comment would have been correct).

dgryski:
Closing in favour of https://github.com/tinygo-org/tinygo/pull/3486 (and others)

PR #3472, opened 2023-02-21
machine/stm32, nrf: flash API

deadprogram:
This PR is to provide an API to read/write onboard flash memory, starting with the STM32. 

EDIT: Now working on STM32F4, STM32L4, and STM32WLx boards.

FURTHER EDIT: also works on nrf51, nrf52, and nrf528xx


deadprogram:
I was just able to use this branch to erase/write/read from flash on an STM32WLx specifically a Wio-E5 board.

```
Connected to /dev/ttyUSB0. Press Ctrl-C to exit.                                                                                                                             
flash data start: 0x08002800                                                                                                                                                 
flash data end:   0x08040000                                                                                                                                                 
writing flash data...                                                                                                                                                        
result: 8765432112345678876543211234567887654321123456788765432112345678
```

deadprogram:
Now also working on STM32L4 and should be on STM32F4. Hence moving to ready for review.

ysoldak:
That is status of NRF chip support?
I see pieces of code related to NRF, but example does not compile for an nrf target.
Something to cleanup in this PR or some files left behind?

deadprogram:
The nrf code was started by @aykevl and was just to return the `FlashDataStart()`. I should probably remove the `nrf` specific parts from this PR, so that implementation can be in a separate PR.

aykevl:
I think it would be a good idea to use the same interface as is used in TinyFS (as mentioned by @bgould on Slack).

deadprogram:
This PR now uses the same `BlockDevice` interface as used in TinyFS. Very good suggestion, thanks.

Also, I have implemented the flash interface for all of the Nordic Semi nrf51/nrf52/nrf528xx processors. Hence the name change for the PR.

src/machine/machine_nrf.go
```diff
@@ -382,3 +383,90 @@ func ReadTemperature() int32 {
...
+
+type flashBlockDevice struct {
+}
+
+// ReadAt reads the given number of bytes from the block device.
+func (f flashBlockDevice) ReadAt(p []byte, off int64) (n int, err error) {
+	data := unsafe.Slice((*byte)(unsafe.Pointer(FlashDataStart()+uintptr(off))), len(p))
+	copy(p, data)
+
+	return 0, nil
```
bgould: Should this return `len(data)` for the first value?

_ReadAt reads len(p) bytes into p starting at offset off in the underlying input source. It returns the number of bytes read (0 <= n <= len(p)) and any error encountered._ (https://pkg.go.dev/io#ReaderAt)

deadprogram:
OK that was my last change: small renaming and add a compile time interface check.

src/machine/machine_stm32_flash.go
```diff
@@ -0,0 +1,100 @@
...
+
+type flashBlockDevice struct {
+}
+
+// ReadAt reads the given number of bytes from the block device.
+func (f flashBlockDevice) ReadAt(p []byte, off int64) (n int, err error) {
+	data := unsafe.Slice((*byte)(unsafe.Pointer(FlashDataStart()+uintptr(off))), len(p))
+	copy(p, data)
+
+	return 0, nil
```
bgould: Same comment as above ... should return the number of bytes read into the buffer (perhaps additionallly EOF or another error if attempting to read past the end of flash data address)

deadprogram: Good point. Both done.

src/machine/machine_nrf.go
deadprogram: Done. Also check for EOF on read/write like other suggestion.

bgould:
I don't have any STM32 boards so can't test ... but from reading the code LGTM

> The nrf code was started by @aykevl and was just to return the `FlashDataStart()`. I should probably remove the `nrf` specific parts from this PR, so that implementation can be in a separate PR.

I do have nrf boards so happy to test this part ... but the above comment suggests that maybe it isn't intended to be included?

One other super-minor suggestion ... might be worth documenting the WriteAt method to mention that it expected that pages being written should be pre-erased.  I like that the low-level block device implementation doesn't do this, as it makes most sense to do that in a higher-level implemenation such as FlashBuffer or a filesystem ... but that requirement is somewhat different that most io.WriterAt implementations so could be worth documenting (towards that end such a comment is in the `tinygo.org/x/drivers/flash` package for instance - https://github.com/tinygo-org/drivers/blob/release/flash/flash.go#L244-L246)

deadprogram:
>  but the above comment suggests that maybe it isn't intended to be included?

I ended up adding the nrf support in this PR so I could make sure everything was going to work as I intended on multiple processor types.

ysoldak:
Verified, works on an nrf board, target `xiao-ble`!

I've modified the example, so it reads flash on start to verify data actually persists after power cycle.
Also prints info on flash size in bytes and write block size.
Making a separate branch and PR probably is overkill, so I just post the example here:

```
package main

import (
	"machine"
	"time"
)

var (
	err     error
	message = "Hello, TinyGo!!!"
)

func main() {

	flash := machine.OpenFlashBuffer(machine.Flash, machine.FlashDataStart())
	data := make([]byte, len(message))

	time.Sleep(3 * time.Second)

	// Print out general information
	println("Flash data start:      ", machine.FlashDataStart())
	println("Flash data end:        ", machine.FlashDataEnd())
	println("Flash data size, bytes:", int(machine.FlashDataEnd()-machine.FlashDataStart()))
	println("Flash write block size:", machine.Flash.WriteBlockSize())
	println()

	// Read flash contents on start (data shall survive power off)
	print("Reading data from flash: ")
	_, err = flash.Read(data)
	checkError(err)
	println(string(data))

	// Write the message to flash
	print("Writing data to   flash: ")
	flash.Seek(0, 0) // rewind back to beginning
	_, err = flash.Write([]byte(message))
	println(message)
	checkError(err)

	// Read back flash contents after write (verify data is the same as written)
	print("Reading data from flash: ")
	flash.Seek(0, 0) // rewind back to beginning
	_, err = flash.Read(data)
	checkError(err)
	println(string(data))

	println()
	for {
		println(string(data))
		time.Sleep(time.Second)
	}
}

func checkError(err error) {
	if err != nil {
		for {
			println(err.Error())
			time.Sleep(time.Second)
		}
	}
}
```

Please note, if the message length is changed, it errors on write with "write flash data must align to correct number of bits".
This contradicts the documentation for `machine.Flash.WriteBlockSize()` that states: "non-aligned writes
// should always work correctly".

Another inconsistency in documentation is for `machine.Seek()` that states: "you cannot use seek before write operations, only read", but we have to seek back to the start before write, if there were a read before that; does not work otherwise.

bgould:
Wow, works great! Slightly modified the littlefs example in TinyFS to try this out and it works great: https://github.com/tinygo-org/tinyfs/blob/machine-flash/examples/console/littlefs/machine/main.go

```
~/src/tinyfs/examples/console/littlefs/machine (machine-flash)        
$ tinygo flash -target circuitplay-bluefruit

~/src/tinyfs/examples/console/littlefs/machine (machine-flash)
$ winpty python -m serial.tools.miniterm COM8
--- Miniterm on COM8  9600,8,N,1 ---
--- Quit: Ctrl+] | Menu: Ctrl+T | Help: Ctrl+T followed by Ctrl+H ---

==> lsblk

-------------------------------------
 Device Information:
-------------------------------------
 flash data start: 0003A000
 flash data end:  000ED000
-------------------------------------

==> format
Successfully formatted LittleFS filesystem.

==> ls
Could not open directory /: littlefs: Corrupted
==> mount
Successfully mounted LittleFS filesystem.

==> ls
==> samples
wrote 90 bytes to file0.txt
wrote 90 bytes to file1.txt
wrote 90 bytes to file2.txt
wrote 90 bytes to file3.txt
wrote 90 bytes to file4.txt
==> cat file2.txt
00000000: 20 21 22 23 24 25 26 27 28 29 2a 2b 2c 2d 2e 2f     !"#$%&'()*+,-./
00000010: 30 31 32 33 34 35 36 37 38 39 3a 3b 3c 3d 3e 3f    0123456789:;<=>?
00000020: 40 41 42 43 44 45 46 47 48 49 4a 4b 4c 4d 4e 4f    @ABCDEFGHIJKLMNO
00000030: 50 51 52 53 54 55 56 57 58 59 5a 5b 5c 5d 5e 5f    PQRSTUVWXYZ[\]^_
00000040: 60 61 62 63 64 65 66 67 68 69 6a 6b 6c 6d 6e 6f    `abcdefghijklmno
00000050: 70 71 72 73 74 75 76 77 78 79                      pqrstuvwxy
==>
```

bgould (approved):

bgould:
@deadprogram This all looks good ... are you ok with it if I squash these commits when merging?

deadprogram:
All feedback should now be addressed.

@ysoldak I added the needed code to ensure that `WriteAt` handles alignment if needed. You can verify by changing the data in the example.

@bgould yes, once the tests pass again, I intend to squash everything before merging.

Issue #2987, opened 2022-07-14
Arduino Nano 33 IoT floating pin when using interrupts


bpoetzschke:
Hi there,

I was playing around with configuring pins to input mode and came across the GPIO documentation here https://tinygo.org/docs/concepts/peripherals/gpio/.

So I configured my pin as PinInputPullup and afterward, I configured the interrupt. And in the callback of the interrupt, I can see the pin floating. I need to connect an external PullUp resistor to make it work. Interestingly I can also just configure the interrupt without setting any input mode before. So it seems that configuring an interrupt overrides the previous pin config.

However, the documentation of `SetInterrupts` states I need to configure the pin first.

I attached 3 examples of the behaviors I observed.

Example 1: Floating pin.
```go
func main() {
	waitSerial()

	led.Configure(machine.PinConfig{Mode: machine.PinOutput})
	input.Configure(machine.PinConfig{Mode: machine.PinInputPullup})

	input.SetInterrupt(machine.PinToggle, func(pin machine.Pin) {
		println(fmt.Sprintf("Pin toggled to %t", pin.Get()))
		led.Set(pin.Get())
	})

	for {
		time.Sleep(time.Second)
	}
```

Output:
```
21:37:17.094 -> Pin toggled to true
21:37:17.094 -> Pin toggled to false
21:37:17.094 -> Pin toggled to true
21:37:17.094 -> Pin toggled to false
21:37:17.094 -> Pin toggled to true
21:37:17.121 -> Pin toggled to false
21:37:17.121 -> Pin toggled to true
21:37:17.121 -> Pin toggled to false
21:37:17.121 -> Pin toggled to true
21:37:17.158 -> Pin toggled to false
21:37:17.158 -> Pin toggled to true
21:37:17.158 -> Pin toggled to false
21:37:17.158 -> Pin toggled to true
21:37:17.192 -> Pin toggled to false
21:37:17.192 -> Pin toggled to true
21:37:17.192 -> Pin toggled to false
21:37:17.192 -> Pin toggled to true
...
```

Example 2: Floating pin without configuring it as input at all. Produces the same output.
```go
func main() {
	waitSerial()

	led.Configure(machine.PinConfig{Mode: machine.PinOutput})

	input.SetInterrupt(machine.PinToggle, func(pin machine.Pin) {
		println(fmt.Sprintf("Pin toggled to %t", pin.Get()))
		led.Set(pin.Get())
	})

	for {
		time.Sleep(time.Second)
	}
[...]

PR #3475, opened 2023-02-23
Rearrange switch case for get pin cfg

bpoetzschke:
This MR attempts to fix https://github.com/tinygo-org/tinygo/issues/2987.

I tested it with the example provided in my issue I created and the floating of the pin stops and seems to be connected properly to a pull up resister now.

The tl;dr is that pin 36 in `getPinCfg` of `src/machine/machine_atsamd21g18.go` was missing and caused a wrong arrangement of the switch cases leading to floating inputs for pins higher than 36 on at least Arduino Nano 33 boards.


deadprogram (approved):
Oops! Thanks for the fix @bpoetzschke good catch!

bpoetzschke:
@deadprogram this is my first pull request for TinyGo, so how will this be merged? Will you or someone from the TinyGo team merge it?

deadprogram:
Now merging, thanks @bpoetzschke

PR #3476, opened 2023-02-23
machine/rp2040: correct issues with i2c/spi pin validation

deadprogram:
This PR corrects an issue with i2c pin validation introduced in https://github.com/tinygo-org/tinygo/pull/3443

### DIFF
--- a/src/machine/machine_rp2040_i2c.go
+++ b/src/machine/machine_rp2040_i2c.go
@@ -103,21 +103,23 @@ func (i2c *I2C) Configure(config I2CConfig) error {
 			config.SDA = I2C1_SDA_PIN
 		}
 	}
-	var okSDA, okSCL bool
+	var okSCL, okSDA bool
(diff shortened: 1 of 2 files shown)


deadprogram:
Wait, I just looked at the datasheet more closely. I need to change this a little...

deadprogram:
OK @soypat now I think I have corrected it in a way that preserves your original intent.

deadprogram:
Made a second commit to correct error in the SPI pin validation.

Both of these issues were discovered by TinyHCI :robot:

deadprogram:
These are the RP2040 Pico tests that fail until this PR is applied:

image

soypat (approved):
Thank goodness gracious (and @deadprogram) for TinyHCI!
Looks good to me! Sorry about the faulty PR.

deadprogram:
Thanks for getting things started with the original PR @soypat for for quick review. Now merging.

PR #3480, opened 2023-02-24
FPU support for ARM Cortex-M targets

waj334:
Adds FPU support to the scheduler by allowing the FPU registers to be saved during a context switch. The target must have the `fpu` build tag and `FPU_ENABLED` preprocessor define set. Otherwise, the smaller stack frame is used and the non-FPU context save/restore routine will be performed.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -796,7 +796,7 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 			if config.AutomaticStackSize() {
 				// Modify the .tinygo_stacksizes section that contains a stack size
 				// for each goroutine.
-				err = modifyStackSizes(result.Executable, stackSizeLoads, stackSizes)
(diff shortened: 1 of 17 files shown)


biohazduck:
This adds hardware support only for the M4? How would one go about adding hardware support for a board with a peripheral-based FPU like the raspberry pi pico?

waj334:
> This adds hardware support only for the M4? How would one go about adding hardware support for a board with a peripheral-based FPU like the raspberry pi pico?

It should work on Cortex-M33 and Cortex-M7, but I don't have any such boards to test on. Cortex-M0 should be the only CPU not supported by this since it has no FPU.

To enable support you need the processor define and build tag set as mentioned and "-mfloat-abi=hard" in the cflags so that the compiler will actually emit FPU instructions.

I'm brainstorming a better method of switching this support on so that there won't be any need to duplicate the target json files.

deadprogram:
I think this has already been handled some time ago. In any case, please open a new PR against the latest `dev` if still needed. Thank you!

PR #3483, opened 2023-02-25
runtime: properly turn pointer into empty interface when hashing

dgryski:


dgryski:
Without this fix, when keys were being rehashed during map growth, they weren't properly being seen as empty interfaces with the contents as the type/value we expect, but as structs with two `unsafe.Pointer`s.   This PR also includes a test demonstrating the failure.

aykevl (approved):
Thank you for the quick fix! That's indeed a bug there: the key is loaded as `_interface` but then silently wrapped in an interface in the `hashmapInterfaceHash` call.

PR #3484, opened 2023-02-25
ci: add AVR timers test

aykevl:
Add the timers test because they now work correctly on AVR, probably as a result of the reflect refactor: https://github.com/tinygo-org/tinygo/pull/2640

I've also updated a few of the other tests to indicate the new status and why they don't work. It's no longer because of compiler errors, but because of linker or runtime errors (which is at least some progress). For example, I found that testdata/reflect.go works if you disable `testAppendSlice` and increase the stack size.


deadprogram:
Thanks @aykevl now merging.

PR #3486, opened 2023-02-25
reflect cleanups: Maps

dgryski:
Cleaning up my reflect PRs.  This one is the first in a series that gathers together *only* the improvements for Maps. 

Next up, Slices.

### DIFF
--- a/src/reflect/value.go
+++ b/src/reflect/value.go
@@ -641,7 +641,27 @@ func (v Value) OverflowFloat(x float64) bool {
 }
 
 func (v Value) MapKeys() []Value {
-	panic("unimplemented: (reflect.Value).MapKeys()")
+	if v.Kind() != Map {
+		panic(&ValueError{Method: "MapKeys", Kind: v.Kind()})
+	}
+
+	// empty map
+	if v.Len() == 0 {
+		return nil
+	}
+
+	keys := make([]Value, 0, v.Len())
+
+	it := hashmapNewIterator()
+	k := New(v.typecode.Key())
+	e := New(v.typecode.Elem())
+
+	for hashmapNext(v.pointer(), it, k.value, e.value) {
+		keys = append(keys, k.Elem())
+		k = New(v.typecode.Key())
+	}
+
+	return keys
 }
 
 //go:linkname hashmapStringGet runtime.hashmapStringGetUnsafePointer
--- a/src/reflect/value.go
+++ b/src/reflect/value.go
@@ -687,23 +687,53 @@ func (v Value) MapIndex(key Value) Value {
 	panic("unimplemented: (reflect.Value).MapIndex()")
 }
 
+//go:linkname hashmapNewIterator runtime.hashmapNewIterator
+func hashmapNewIterator() unsafe.Pointer
+
+//go:linkname hashmapNext runtime.hashmapNextUnsafePointer
+func hashmapNext(m unsafe.Pointer, it unsafe.Pointer, key, value unsafe.Pointer) bool
+
 func (v Value) MapRange() *MapIter {
-	panic("unimplemented: (reflect.Value).MapRange()")
+	if v.Kind() != Map {
+		panic(&ValueError{Method: "MapRange", Kind: v.Kind()})
+	}
+
+	return &MapIter{
+		m:   v,
+		it:  hashmapNewIterator(),
+		key: New(v.typecode.Key()),
+		val: New(v.typecode.Elem()),
+	}
 }
 
 type MapIter struct {
+	m   Value
+	it  unsafe.Pointer
+	key Value
+	val Value
+
+	valid bool
 }
 
 func (it *MapIter) Key() Value {
-	panic("unimplemented: (*reflect.MapIter).Key()")
+	if !it.valid {
+		panic("reflect.MapIter.Key called on invalid iterator")
+	}
+
+	return it.key.Elem()
 }
 
 func (it *MapIter) Value() Value {
-	panic("unimplemented: (*reflect.MapIter).Value()")
+	if !it.valid {
+		panic("reflect.MapIter.Value called on invalid iterator")
+	}
+
+	return it.val.Elem()
 }
 
 func (it *MapIter) Next() bool {
-	panic("unimplemented: (*reflect.MapIter).Next()")
+	it.valid = hashmapNext(it.m.pointer(), it.it, it.key.value, it.val.value)
+	return it.valid
 }
 
 func (v Value) Set(x Value) {
--- a/src/reflect/value.go
+++ b/src/reflect/value.go
@@ -972,8 +972,70 @@ func AppendSlice(s, t Value) Value {
 	}
 }
 
+//go:linkname hashmapStringSet runtime.hashmapStringSetUnsafePointer
+func hashmapStringSet(m unsafe.Pointer, key string, value unsafe.Pointer)
+
+//go:linkname hashmapBinarySet runtime.hashmapBinarySetUnsafePointer
+func hashmapBinarySet(m unsafe.Pointer, key, value unsafe.Pointer)
+
+//go:linkname hashmapStringDelete runtime.hashmapStringDeleteUnsafePointer
+func hashmapStringDelete(m unsafe.Pointer, key string)
+
+//go:linkname hashmapBinaryDelete runtime.hashmapBinaryDeleteUnsafePointer
+func hashmapBinaryDelete(m unsafe.Pointer, key unsafe.Pointer)
+
 func (v Value) SetMapIndex(key, elem Value) {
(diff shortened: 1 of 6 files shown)


src/reflect/value.go
```diff
@@ -644,7 +644,45 @@ func (v Value) MapKeys() []Value {
...
+			keyptr = unsafe.Pointer(&key.value)
+		}
+		//TODO(dgryski): zero out padding bytes in key, if any
+		if ok := hashmapBinaryGet(v.pointer(), keyptr, elem.value, elemType.Size()); !ok {
+			return Value{}
+		}
+		return elem.Elem()
+	}
+
 	panic("unimplemented: (reflect.Value).MapIndex()")
```
aykevl, line 707: I suppose other map types will come at a later time?

src/reflect/value.go
```diff
@@ -644,7 +644,45 @@ func (v Value) MapKeys() []Value {
...
+		}
+		return elem.Elem()
+	} else if key.typecode.isBinary() {
+		var keyptr unsafe.Pointer
+		if key.isIndirect() {
+			keyptr = key.value
+		} else {
+			keyptr = unsafe.Pointer(&key.value)
+		}
+		//TODO(dgryski): zero out padding bytes in key, if any
```
aykevl, line 699: I'm not sure this happens in practice. And if it does, we'd have to make sure to zero the original value (not try and zero it here). For example, this is one place that takes special care to zero the undef parts of a value:

https://github.com/tinygo-org/tinygo/blob/0d56dee00f49bd50eb373c02c30062a75ec28f10/compiler/llvm.go#L82-L87

src/reflect/value.go
```diff
@@ -644,7 +644,45 @@ func (v Value) MapKeys() []Value {
...
+			return Value{}
+		}
+		return elem.Elem()
+	} else if key.typecode.isBinary() {
+		var keyptr unsafe.Pointer
+		if key.isIndirect() {
+			keyptr = key.value
+		} else {
+			keyptr = unsafe.Pointer(&key.value)
+		}
```
aykevl: There are three cases to consider here, not two:

1. When `isIndirect()` returns true: in that case `key.value` is always a pointer to the value.
2. When the value fits in a pointer, in which case you can indeed use `unsafe.Pointer(&key.value)`.
3. When the value is larger than a pointer, it is also indirect (but without the flag).

See for example:

https://github.com/tinygo-org/tinygo/blob/0d56dee00f49bd50eb373c02c30062a75ec28f10/src/reflect/value.go#L243-L247

src/reflect/value.go
```diff
@@ -644,7 +644,45 @@ func (v Value) MapKeys() []Value {
...
+
 func (v Value) MapIndex(key Value) Value {
+	if v.Kind() != Map {
+		panic(&ValueError{Method: "MapIndex", Kind: v.Kind()})
+	}
+
+	// compare key type with actual key type of map
+	if key.typecode != v.typecode.key() {
+		// type error?
+		panic(&ValueError{Method: "MapIndex"})
```
aykevl: Well, I guess it's a "value error" in the sense that the value is incorrect (has the wrong type).

src/reflect/value.go
```diff
@@ -686,23 +686,67 @@ func (v Value) MapIndex(key Value) Value {
...
 
 func (it *MapIter) Value() Value {
-	panic("unimplemented: (*reflect.MapIter).Value()")
+	if !it.valid {
+		panic("MapIter: Value() called before Next()")
+	}
+
+	if it.done {
+		panic("MapIter.Value called on exhausted iterator")
+	}
```
aykevl: You could save a bit of const memory (and sacrifice panic message readability) by collapsing these four panic messages into one:

1. Merge `.valid` and `.done` into one flag that simply indicates whether Key()/Value() are the same.
2. Use the same panic message in both cases (the compiler should merge the two strings).

Not sure it's worth it though. Just an idea.

src/reflect/value.go
```diff
@@ -985,8 +985,70 @@ func AppendSlice(s, t Value) Value {
...
+
 func (v Value) SetMapIndex(key, elem Value) {
-	panic("unimplemented: (reflect.Value).SetMapIndex()")
+	if v.Kind() != Map {
+		panic(&ValueError{Method: "SetMapIndex", Kind: v.Kind()})
+	}
+
+	// compare key type with actual key type of map
+	if key.typecode != v.typecode.key() {
+		panic("reflect.Value.SetMapIndex: incompatible types for key")
```
aykevl, line 1014: Slightly inconsistent with `MapIndex` which uses `ValueError`. I'm not too worried about it but might be cleaner to use the same sort of panic message in both cases.

src/reflect/value.go
```diff
@@ -985,8 +985,70 @@ func AppendSlice(s, t Value) Value {
...
+	if key.Kind() == String {
+		if del {
+			hashmapStringDelete(v.pointer(), *(*string)(key.value))
+		} else {
+			var elemptr unsafe.Pointer
+			if elem.isIndirect() {
+				elemptr = elem.value
+			} else {
+				elemptr = unsafe.Pointer(&elem.value)
+			}
```
aykevl: Same issue as before. You should probably use the following:

```go
if elem.isIndirect() || elem.Size() > unsafe.Sizeof(uintptr(0)) {
```

(The same issue is also present a few times below.)

src/reflect/value.go
```diff
@@ -1065,9 +1065,52 @@ func (v Value) FieldByName(name string) Value {
...
+
+// MakeMapWithSize creates a new map with the specified type and initial space
+// for approximately n elements.
+func MakeMapWithSize(typ Type, n int) Value {
+
+	const (
+		hashmapAlgorithmBinary uint8 = iota
+		hashmapAlgorithmString
+		hashmapAlgorithmInterface
+	)
```
aykevl, line 1087: It's kind of unfortunate to have this duplicated.
I guess this could be fixed using an internal package that defines these constants? Not sure it's worth the trouble (and probably better to do in a separate PR if you think it's worth the deduplication).

src/reflect/value.go
```diff
@@ -1065,9 +1065,52 @@ func (v Value) FieldByName(name string) Value {
...
+	val := typ.Elem().(*rawType)
+
+	var alg uint8
+
+	if key.Kind() == String {
+		alg = hashmapAlgorithmString
+	} else if key.isBinary() {
+		alg = hashmapAlgorithmBinary
+	} else {
+		panic("reflect.MakeMap: invalid key type")
```
aykevl: It's not invalid I think? Just not implemented yet.

src/reflect/value.go
```diff
@@ -1065,9 +1065,52 @@ func (v Value) FieldByName(name string) Value {
...
+	}
+
+	m := hashmapMake(key.Size(), val.Size(), uintptr(n), alg)
+
+	return Value{
+		typecode: typ.(*rawType),
+		value:    unsafe.Pointer(m),
```
aykevl: Nit: no need to cast here. (Feel free to ignore this comment, it's very minor).

src/reflect/value.go
dgryski: Yes.  Strings were the easiest, then the binary ones.   This panic was more reminder, but I can make it an explicit `TODO`.

src/reflect/value.go
aykevl: Actually, re-reading the code this does in fact result in undef values, for example if the following value is stored in a struct on a 32-bit architecture:

```go
type small struct {
    x byte
    y int16
}
```

There is some padding, but the `if size < b.targetData.TypeAllocSize(b.i8ptrType) {` check doesn't catch that.

dgryski: Yes, this is exactly the case my recent zero undef padding PR in the compiler fixed.

src/reflect/value.go
dgryski: I was trying to match the error messages from upstream Go.

src/reflect/value.go
dgryski: Good catch. I'll try to make them match.

src/reflect/value.go
dgryski: Yeah.  The runtime needs them, the reflect code needs them, and the compiler needs them.

src/reflect/value.go
dgryski: Yeah, "invalid" in the sense that I haven't gotten around to them yet.  I'll clean up this message.

src/reflect/value.go
dgryski: Oops.  Thanks.  Probably from thinking the returned pointer was a `*hashmap`.

dgryski:
Fixed all review issues and rebased.  PTAL.

aykevl (approved):
Thanks a lot!

PR #3487, opened 2023-02-26
reflect cleanup: Slices

dgryski:
To be rebased once #3486 is merged into dev.

### DIFF
--- a/src/reflect/value.go
+++ b/src/reflect/value.go
@@ -382,7 +382,52 @@ func (v Value) Bytes() []byte {
 }
 
 func (v Value) Slice(i, j int) Value {
-	panic("unimplemented: (reflect.Value).Slice()")
+	switch v.Kind() {
+	case Slice:
+		hdr := *(*sliceHeader)(v.value)
+		i, j := uintptr(i), uintptr(j)
+
+		if j < i || hdr.cap < j {
+			slicePanic()
+		}
+
+		elemSize := v.typecode.underlying().elem().Size()
+
+		hdr.len = j - i
+		hdr.cap = hdr.cap - i
+		hdr.data = unsafe.Add(hdr.data, i*elemSize)
+
+		return Value{
+			typecode: v.typecode,
+			value:    unsafe.Pointer(&hdr),
+			flags:    v.flags,
+		}
+
+	case Array:
+		// TODO(dgryski): can't do this yet because the resulting value needs type slice of v.elem(), not array of v.elem().
+		// need to be able to look up this "new" type so pointer equality of types still works
+
+	case String:
+		i, j := uintptr(i), uintptr(j)
+		str := *(*stringHeader)(v.value)
+
+		if j < i || str.len < j {
+			slicePanic()
+		}
+
+		hdr := stringHeader{
+			data: unsafe.Add(str.data, i),
+			len:  j - i,
+		}
+
+		return Value{
+			typecode: v.typecode,
+			value:    unsafe.Pointer(&hdr),
+			flags:    v.flags,
+		}
+	}
+
+	panic(&ValueError{"Slice", v.Kind()})
 }
 
 func (v Value) Slice3(i, j, k int) Value {
--- a/src/reflect/value.go
+++ b/src/reflect/value.go
@@ -882,8 +882,40 @@ func (v Value) Convert(t Type) Value {
 	panic("unimplemented: (reflect.Value).Convert()")
 }
 
+//go:linkname slicePanic runtime.slicePanic
+func slicePanic()
+
 func MakeSlice(typ Type, len, cap int) Value {
-	panic("unimplemented: reflect.MakeSlice()")
+	if typ.Kind() != Slice {
+		panic("reflect.MakeSlice of non-slice type")
+	}
+
+	rtype := typ.(*rawType)
+
+	ulen := uint(len)
+	ucap := uint(cap)
+	maxSize := (^uintptr(0)) / 2
+	elementSize := rtype.elem().Size()
+	if elementSize > 1 {
+		maxSize /= uintptr(elementSize)
+	}
+	if ulen > ucap || ucap > uint(maxSize) {
+		slicePanic()
+	}
+
+	// This can't overflow because of the above checks.
+	size := uintptr(ucap) * elementSize
+
+	var slice sliceHeader
+	slice.cap = uintptr(ucap)
+	slice.len = uintptr(ulen)
+	slice.data = alloc(size, nil)
+
+	return Value{
+		typecode: rtype,
+		value:    unsafe.Pointer(&slice),
+		flags:    valueFlagExported,
+	}
 }
 
 func Zero(typ Type) Value {
--- a/src/reflect/value.go
+++ b/src/reflect/value.go
@@ -431,6 +431,33 @@ func (v Value) Slice(i, j int) Value {
 }
 
 func (v Value) Slice3(i, j, k int) Value {
+	switch v.Kind() {
+	case Slice:
+		hdr := *(*sliceHeader)(v.value)
+		i, j, k := uintptr(i), uintptr(j), uintptr(k)
+
+		if j < i || k < j || hdr.len < k {
+			slicePanic()
+		}
+
+		elemSize := v.typecode.underlying().elem().Size()
+
+		hdr.len = j - i
+		hdr.cap = k - i
+		hdr.data = unsafe.Add(hdr.data, i*elemSize)
+
+		return Value{
+			typecode: v.typecode,
+			value:    unsafe.Pointer(&hdr),
+			flags:    v.flags,
+		}
+
+	case Array:
+		// TODO(dgryski): can't do this yet because the resulting value needs type v.elem(), not array of v.elem().
+		// need to be able to look up this "new" type so pointer equality of types still works
+
+	}
+
 	panic("unimplemented: (reflect.Value).Slice3()")
 }
 
--- a/src/reflect/value.go
+++ b/src/reflect/value.go
@@ -378,7 +378,27 @@ func (v Value) String() string {
 }
 
 func (v Value) Bytes() []byte {
-	panic("unimplemented: (reflect.Value).Bytes()")
+	switch v.Kind() {
+	case Slice:
+		if v.typecode.elem().Kind() != Uint8 {
+			panic(&ValueError{Method: "Bytes", Kind: v.Kind()})
+		}
+		return *(*[]byte)(v.value)
+
+	case Array:
+		v.checkAddressable()
+
+		if v.typecode.elem().Kind() != Uint8 {
+			panic(&ValueError{Method: "Bytes", Kind: v.Kind()})
+		}
+
+		// Small inline arrays are not addressable, so we only have to
+		// handle addressable arrays which will be stored as pointers
+		// in v.value
+		return unsafe.Slice((*byte)(v.value), v.Len())
+	}
+
+	panic(&ValueError{Method: "Bytes", Kind: v.Kind()})
 }
 
 func (v Value) Slice(i, j int) Value {
--- a/src/reflect/value.go
+++ b/src/reflect/value.go
@@ -958,10 +958,54 @@ func Copy(dst, src Value) int {
 	panic("unimplemented: reflect.Copy()")
 }
 
+//go:linkname sliceGrow runtime.sliceGrow
+func sliceGrow(buf unsafe.Pointer, oldLen, oldCap, newCap, elemSize uintptr) (unsafe.Pointer, uintptr, uintptr)
+
+// extend slice to hold n new elements
+func (v *Value) extendSlice(n int) {
+	if v.Kind() != Slice {
+		panic(&ValueError{Method: "extendSlice", Kind: v.Kind()})
+	}
+
+	var old sliceHeader
+	if v.value != nil {
+		old = *(*sliceHeader)(v.value)
+	}
+
+	var nbuf unsafe.Pointer
+	var nlen, ncap uintptr
+
+	if old.len+uintptr(n) > old.cap {
+		// we need to grow the slice
+		nbuf, nlen, ncap = sliceGrow(old.data, old.len, old.cap, old.cap+uintptr(n), v.typecode.elem().Size())
+	} else {
+		// we can reuse the slice we have
+		nbuf = old.data
+		nlen = old.len
+		ncap = old.cap
+	}
+
+	newslice := sliceHeader{
+		data: nbuf,
+		len:  nlen + uintptr(n),
+		cap:  ncap,
+	}
+
+	v.value = (unsafe.Pointer)(&newslice)
+}
+
 // Append appends the values x to a slice s and returns the resulting slice.
 // As in Go, each x's value must be assignable to the slice's element type.
-func Append(s Value, x ...Value) Value {
-	panic("unimplemented: reflect.Append()")
+func Append(v Value, x ...Value) Value {
+	if v.Kind() != Slice {
+		panic(&ValueError{Method: "Append", Kind: v.Kind()})
+	}
+	oldLen := v.Len()
+	v.extendSlice(len(x))
+	for i, xx := range x {
+		v.Index(oldLen + i).Set(xx)
+	}
+	return v
 }
 
 // AppendSlice appends a slice t to a slice s and returns the resulting slice.
--- a/src/reflect/value.go
+++ b/src/reflect/value.go
@@ -1056,10 +1056,57 @@ func alloc(size uintptr, layout unsafe.Pointer) unsafe.Pointer
 //go:linkname sliceAppend runtime.sliceAppend
 func sliceAppend(srcBuf, elemsBuf unsafe.Pointer, srcLen, srcCap, elemsLen uintptr, elemSize uintptr) (unsafe.Pointer, uintptr, uintptr)
 
+//go:linkname sliceCopy runtime.sliceCopy
+func sliceCopy(dst, src unsafe.Pointer, dstLen, srcLen uintptr, elemSize uintptr) int
+
 // Copy copies the contents of src into dst until either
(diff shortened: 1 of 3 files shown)


src/reflect/value.go
```diff
@@ -971,10 +971,43 @@ func Copy(dst, src Value) int {
...
+
+	nbuf, nlen, ncap := sliceGrow(old.data, old.len, old.cap, old.cap+uintptr(n), v.typecode.elem().Size())
+
+	newslice := sliceHeader{
+		data: nbuf,
+		len:  nlen + uintptr(n),
+		cap:  ncap,
+	}
+
+	v.value = (unsafe.Pointer)(&newslice)
```
aykevl, line 1165: I tried to find a way this heap allocation can be avoided but it seems it can't (otherwise some invariants would be violated). So LGTM.

src/reflect/value.go
```diff
@@ -895,8 +895,28 @@ func (v Value) Convert(t Type) Value {
...
+//go:linkname slicePanic runtime.slicePanic
+func slicePanic()
+
 func MakeSlice(typ Type, len, cap int) Value {
-	panic("unimplemented: reflect.MakeSlice()")
+	if typ.Kind() != Slice {
+		panic("reflect.MakeSlice of non-slice type")
+	}
+
+	if len < 0 || cap < 0 || len > cap || int(uintptr(len)) != len || int(uintptr(cap)) != cap {
```
aykevl: There is actually one more thing that is normally checked: whether the capacity won't overflow the address space. I tried to reproduce this on the Go playground, but it seems to just die: https://go.dev/play/p/qOhvwZUJrEB

Here is some code I used to test this locally (changing `1 << 20` to `1 << 40` makes this reproducible on 64-bit systems):

```go
package main

import "reflect"

func main() {
        t := reflect.TypeOf([][1 << 20]byte(nil))
        slice := reflect.MakeSlice(t, 1<<20, 1<<20)
        println("slice:", slice.Len(), slice.Cap())
}
```

```
$ GOARCH=arm go build -o test tmp/bigslice.go
$ qemu-arm test
panic: runtime: allocation size out of range

goroutine 1 [running]:
reflect.unsafe_NewArray(0x84538, 0x100000)
        /usr/local/go/src/runtime/malloc.go:1281 +0x24
reflect.MakeSlice({0xa2ab8, 0x83c28}, 0x100000, 0x100000)
        /usr/local/go/src/reflect/value.go:3091 +0xa4
main.main()
        /home/ayke/src/tinygo/tinygo/tmp/bigslice.go:7 +0x6c
```

See `maxSliceSize` in the compiler for how this is implemented for regular slice operations. It seems that at runtime, the regular Go compiler uses `math.MulUintptr` to determine whether the buffer size would be out of range.

---

EDIT: these checks can be simplified further, like this:

```go
ulen := uint(len)
ucap := uint(cap)
maxSize := /* some code calculating the maximum possible size, something like ^uintptr(0) / 2 / typ.Elem().Size() */
if ulen > ucap || ucap > maxSize {
```

src/reflect/value.go
```diff
@@ -895,8 +895,28 @@ func (v Value) Convert(t Type) Value {
...
+		slicePanic()
+	}
+
+	var slice sliceHeader
+	slice.cap = uintptr(cap)
+	slice.len = uintptr(len)
+	slice.data = unsafe.Pointer(alloc(typ.Size()*uintptr(cap), nil))
```
aykevl: No need for the cast here:

```suggestion
	slice.data = alloc(typ.Size()*uintptr(cap), nil)
```

src/reflect/value.go
```diff
@@ -382,7 +382,53 @@ func (v Value) Bytes() []byte {
...
+
+	case Array:
+		// TODO(dgryski): can't do this yet because the resulting value needs type slice of v.elem(), not array of v.elem().
+		// need to be able to look up this "new" type so pointer equality of types still works
+
+	case String:
+		i, j := uintptr(i), uintptr(j)
+		str := *(*stringHeader)(v.value)
+
+		if i < 0 || i > str.len || j < 0 || j > str.len {
```
aykevl: This doesn't check for `i > j`, which would lead to a negative string length.

```suggestion
		if i < 0 || j < i || str.len < j {
```

Or even this (because `i < 0` is always false):

```suggestion
		if j < i || str.len < j {
```

src/reflect/value.go
```diff
@@ -382,7 +382,53 @@ func (v Value) Bytes() []byte {
 }
 
 func (v Value) Slice(i, j int) Value {
-	panic("unimplemented: (reflect.Value).Slice()")
+	switch v.Kind() {
+	case Slice:
+		hdr := *(*sliceHeader)(v.value)
+		i, j := uintptr(i), uintptr(j)
+
+		if i < 0 || i > hdr.len || j < 0 || j > hdr.len {
```
aykevl:
```suggestion
		if i < 0 || j < i || hdr.len < j {
```

src/reflect/value.go
```diff
@@ -382,7 +382,53 @@ func (v Value) Bytes() []byte {
...
+
+	case String:
+		i, j := uintptr(i), uintptr(j)
+		str := *(*stringHeader)(v.value)
+
+		if i < 0 || i > str.len || j < 0 || j > str.len {
+			slicePanic()
+		}
+
+		hdr := sliceHeader{
```
aykevl: `sliceHeader` for a string? (by chance, this will in fact work but is probably not what you intended).

src/reflect/value.go
```diff
@@ -432,6 +432,33 @@ func (v Value) Slice(i, j int) Value {
 }
 
 func (v Value) Slice3(i, j, k int) Value {
+	switch v.Kind() {
+	case Slice:
+		hdr := *(*sliceHeader)(v.value)
+		i, j, k := uintptr(i), uintptr(j), uintptr(k)
+
+		if i < 0 || i > hdr.len || j < 0 || j > hdr.len || k < 0 || k > hdr.len {
```
aykevl:
```suggestion
		if i < 0 || j < i || k < j || hdr.len < k {
```

But because `i` is an `uintptr`, `i < 0` doesn't do anything so can be left out:

```suggestion
		if j < i || k < j || hdr.len < k {
```

This is in fact what the compiler does. It's correct because: if `k` is negative, it wraps around and will be bigger than `hdr.len`. Similarly, if `j` is negative (but `k` is in bounds), `j` wraps around and is thus larger than `k`. And so on for `i`.

src/reflect/value.go
```diff
@@ -378,7 +378,31 @@ func (v Value) String() string {
...
 
 func (v Value) Bytes() []byte {
-	panic("unimplemented: (reflect.Value).Bytes()")
+	switch v.Kind() {
+	case Slice:
+		if v.typecode.elem().Kind() != Uint8 {
+			panic(&ValueError{Method: "Bytes", Kind: v.Kind()})
+		}
+		hdr := *(*sliceHeader)(v.value)
+		return *(*[]byte)(unsafe.Pointer(&hdr))
```
aykevl: You can cast to the appropriate type immediately:

```suggestion
		return *(*[]byte)(v.value)
```

src/reflect/value.go
```diff
@@ -378,7 +378,31 @@ func (v Value) String() string {
...
+
+		var hdr sliceHeader
+
+		l := uintptr(v.Len())
+
+		hdr.len = l
+		hdr.cap = l
+		hdr.data = v.value
+
+		return *(*[]byte)(unsafe.Pointer(&hdr))
```
aykevl: You can use `unsafe.Slice` for such an operation:

```suggestion
		return unsafe.Slice((*byte)(v.value), v.Len())
```

src/reflect/value.go
```diff
@@ -1082,10 +1082,57 @@ func alloc(size uintptr, layout unsafe.Pointer) unsafe.Pointer
...
+		hdr := (*sliceHeader)(v.value)
+		buf = hdr.data
+		len = hdr.len
+	case Array:
+		if v.isIndirect() {
+			buf = v.value
+			len = uintptr(v.Len())
+		} else {
+			// TODO(dgryski): Need to fix this
+			panic("Copy: does not support small inline arrays")
```
aykevl: The following should work:

```go
buf = unsafe.Pointer(&v.value)
len = uintptr(v.Len())
```

That means, you're copying bytes straight into the `v.value` member.

EDIT: actually, is this allowed at all if the indirect flag isn't set?

src/reflect/value.go
```diff
@@ -1082,10 +1082,57 @@ func alloc(size uintptr, layout unsafe.Pointer) unsafe.Pointer
...
+		}
+	case String:
+		hdr := (*stringHeader)(v.value)
+		buf = hdr.data
+		len = hdr.len
+	default:
+		panic("reflect.Copy: not slice or array or string")
```
aykevl: Might be useful to add a comment that this should be unreachable.

aykevl (commented):
This is a very well put together PR! Thanks for the cleanup!

I reviewed just the slice commits (not the map commits, those are in #3486).

src/reflect/value.go
dgryski: The issue is not wanting to modify the existing slice header, so we have to allocate a new one :(

src/reflect/value.go
dgryski: Oops.  Forgot that slicing a string returns a string, not a slice.

src/reflect/value.go
dgryski: No, I think you're right.  The value isn't addressable.
```
~/go/src/github.com/dgryski/bug/ref $ cat main.go
package main

import "reflect"

func main() {
	var a [3]byte
	refa := reflect.ValueOf(a)
	reflect.Copy(refa, reflect.ValueOf("abc"))
	println(string(a[:]))
}
~/go/src/github.com/dgryski/bug/ref $ go run  main.go
panic: reflect: unknown method using unaddressable value
```

src/reflect/value.go
dgryski: Checking for multiplication overflow seems to be roughly:
```
   m := x * y
   return x != m/y
```
which is easy enough to add as a small helper in reflect.  Pulling `mat.MulUintptr` from `runtime/internal/math` seems like overkill at the moment.  We might turn out to need it in more places, but for now I'm going to do add this as a tiny helper in `reflect`.

dgryski:
All review comments addressed; rebased.  PTAL.

dgryski:
Rebased against dev now that Maps cleanups are in.

src/reflect/value.go
```diff
@@ -882,8 +882,43 @@ func (v Value) Convert(t Type) Value {
...
-	panic("unimplemented: reflect.MakeSlice()")
+	if typ.Kind() != Slice {
+		panic("reflect.MakeSlice of non-slice type")
+	}
+
+	ulen := uintptr(len)
+	ucap := uintptr(cap)
+	if ulen > ucap {
+		slicePanic()
+	}
```
aykevl: There is one edge case that's not handled here: when `uintptr` is smaller than `int` (on AVR). For example, `reflect.MakeSlice([]T, 0x10004, 5)` would return a slice of length 4 on AVR.

In addition, there is an issue with negative integers. This code:

```go
        t := reflect.TypeOf([][0]byte(nil))
        const large =  - 1 * 1024 * 1024 * 1024
        v := reflect.MakeSlice(t, large, large)
        println("reflect slice size:", v.Len(), v.Cap())
```

Results in the following very incorrect output (when I apply the patches in the next two comments, otherwise it will panic for the wrong reason):

```
$ GOARCH=arm tinygo run ./tmp/big-slice.go
reflect slice size: -1073741824 -1073741824
```

This is roughly (but not exactly) what the compiler does. I didn't test it, but I _think_ it covers all edge cases.

```go
ulen := uint(len)
ucap := uint(cap)
maxSize := (^uintptr(0)) / 2
elementSize := typ.Elem().Size()
if elementSize > 1 {
    maxSize /= uintptr(elementSize)
}
if ulen > ucap || ucap > uint(maxSize) {
    slicePanic()
}
```

src/reflect/value.go
```diff
@@ -882,8 +882,43 @@ func (v Value) Convert(t Type) Value {
...
+
+	size, ok := mulNoOverflow(typ.Size(), ucap)
```
aykevl: You need to use the element size here, not the slice size (which is always `3 * unsafe.Size(uintptr(0))`).

```suggestion
	size, ok := mulNoOverflow(typ.Elem().Size(), ucap)
```

src/reflect/value.go
```diff
@@ -882,8 +882,43 @@ func (v Value) Convert(t Type) Value {
...
+	return Value{
+		typecode: typ.(*rawType),
+		value:    unsafe.Pointer(&slice),
+		flags:    valueFlagExported,
+	}
+}
+
+func mulNoOverflow(x, y uintptr) (uintptr, bool) {
+	if x <= 1 || y <= 1 {
+		return x * y, false
```
aykevl: This code disallows elements that are 0 or 1 bytes in size, like this:

```go
        t := reflect.TypeOf([][0]byte(nil))
        v := reflect.MakeSlice(t, 5, 5)
```
Fix:
```suggestion
		return x * y, true
```

src/reflect/value.go
```diff
@@ -382,7 +382,52 @@ func (v Value) Bytes() []byte {
 }
 
 func (v Value) Slice(i, j int) Value {
-	panic("unimplemented: (reflect.Value).Slice()")
+	switch v.Kind() {
+	case Slice:
+		hdr := *(*sliceHeader)(v.value)
+		i, j := uintptr(i), uintptr(j)
```
aykevl, line 408: This also has an overflow problem on AVR, but I'm not sure that's worth worrying about.
(It might also appear to have an issue with negative numbers, but that's caught by `hdr.cap < j` so it's fine).

src/reflect/value.go
```diff
@@ -378,7 +378,22 @@ func (v Value) String() string {
...
+			panic(&ValueError{Method: "Bytes", Kind: v.Kind()})
+		}
+		return *(*[]byte)(v.value)
+
+	case Array:
+		if v.typecode.elem().Kind() != Uint8 {
+			panic(&ValueError{Method: "Bytes", Kind: v.Kind()})
+		}
+
+		return unsafe.Slice((*byte)(v.value), v.Len())
```
aykevl, line 398: This does not work correctly for arrays that fit in a pointer. The following code prints incorrect values:

```go
package main

import "reflect"

func main() {
        v := reflect.ValueOf([3]byte{1, 2, 3})
        slice := v.Bytes()
        println("slice:", len(slice), cap(slice))
        for _, v := range slice {
                println("v:", v)
        }
}
```

src/reflect/value.go
dgryski: Oops. this was from when I was trying to decide if I should return `true` meaning `overflow` or `true` meaning `ok -- no overflow`.

src/reflect/value.go
dgryski: It seems like lot of `reflect` will break on AVR.  That's too bad :(

dgryski:
Issues addressed; rebased against dev. PTAL.

src/reflect/value.go
aykevl: Actually I realize this code still isn't correct. When run under the regular Go compiler, it panics:

```
panic: reflect.Value.Bytes of unaddressable byte array

goroutine 1 [running]:
reflect.Value.bytesSlow({0x1022ee160?, 0x14000018080?, 0x14000044768?})
	/Users/ayke/lib/go/src/reflect/value.go:321 +0xec
reflect.Value.Bytes(...)
	/Users/ayke/lib/go/src/reflect/value.go:305
main.main()
	/Users/ayke/src/tinygo/tinygo/tmp/smallslice.go:7 +0x108
exit status 2
```

Which makes sense: this would allow the caller to modify the array in-place. Also see the documentation (emphasis mine):

> It panics if v's underlying value is not a slice of bytes or an **addressable array of bytes**.

So in short, it should panic for any array that is not addressable (I assume this means `CanAddr`). This conveniently also fixes the issue with very short arrays that fit in a pointer.

src/reflect/value.go
```diff
@@ -882,8 +974,50 @@ func (v Value) Convert(t Type) Value {
...
+		maxSize /= uintptr(elementSize)
+	}
+	if ulen > ucap || ucap > uint(maxSize) {
+		slicePanic()
+	}
+
+	size, ok := mulNoOverflow(elementSize, uintptr(ucap))
+	if !ok {
+		panic("reflect.MakeSlice: slice size out of range")
+	}
```
aykevl: As far as I can see, this check is now unnecessary because of the check above: `ucap > uint(maxSize)` also checks for overflows.

src/runtime/slice.go
```diff
@@ -51,3 +51,29 @@ func sliceCopy(dst, src unsafe.Pointer, dstLen, srcLen uintptr, elemSize uintptr
 	memmove(dst, src, n*elemSize)
 	return int(n)
 }
+
+// sliceGrow returns a new slice with space for at least newCap elements
+func sliceGrow(oldBuf unsafe.Pointer, oldLen, oldCap, newCap, elemSize uintptr) (unsafe.Pointer, uintptr, uintptr) {
```
aykevl, line 56: The code in this function is almost identical to the code in `sliceAppend`. Was this intentional? (Merging the two might be nice to avoid duplication, but might also increase complexity of the code so not sure).

dgryski: I think I originally had a comment here about the duplication. Upstream has this logic merged. So yes we can do that, but I'd rather get this patch unblocked in first.

src/reflect/value.go
dgryski: I had that thought too. I'll remove it.

dgryski:
Fixed and rebased.  PTAL.

src/reflect/value.go
dgryski: Special casing removed and I added a comment.

PR #3488, opened 2023-02-26
builder: remove non-ThinLTO build mode

aykevl:
All targets now support ThinLTO so let's remove the old unused code.

I haven't seen any ThinLTO related issues since the last release, so it seems to me it is working as intended. And I'd like to work on some improvements that will be easier with this old code removed.


deadprogram:
:scissors: thanks @aykevl now merging.

PR #3489, opened 2023-02-26
Add real ThinLTO mode

aykevl:
We use ThinLTO for linking, but we use it in a way that doesn't give most of its benefits: we merge all the bitcode files into a single LLVM module and run some optimizations on it before linking. Therefore, this works more like a traditional "full" LTO link rather than a true thin link.

This PR adds a new experimental `-lto=thin` option to do a true ThinLTO link. The main benefit is that linking will be a lot faster especially for large programs consisting of many packages.

At the moment, it only works for programs that don't do interface type asserts and don't call interface methods. It also probably won't work on WebAssembly and baremetal systems. But it's part of a larger goal towards a truly incremental build system: https://github.com/tinygo-org/tinygo/issues/2870
Once interface type asserts and method calls are converted to a [vtable-like implementation](https://research.swtch.com/interfaces), most programs should just work on linux/darwin/windows.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -511,24 +511,11 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 
 			// Create runtime.initAll function that calls the runtime
 			// initializer of each package.
-			llvmInitFn := mod.NamedFunction("runtime.initAll")
-			llvmInitFn.SetLinkage(llvm.InternalLinkage)
-			llvmInitFn.SetUnnamedAddr(true)
-			transform.AddStandardAttributes(llvmInitFn, config)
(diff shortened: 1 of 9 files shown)


aykevl:
@dgryski this also includes a small patch that finally enables source-level debugging on MacOS:

Screenshot 2023-02-26 at 23 58 52

(This only works with `-lto=thin`, so it's not very useful yet).

deadprogram:
Seems like some actual build problems on this @aykevl please take a look.

aykevl:
@deadprogram yeah I saw it. Frustratingly, I can't reproduce the MacOS issue locally. To fix both issues (Windows and MacOS), I think I need to do the interface refactor first.

PR #3497, opened 2023-02-28
reflect cleanups: small bits

dgryski:


dgryski:
To be rebased after #3487 is merged.

src/reflect/value.go
```diff
@@ -948,7 +948,15 @@ func (v Value) SetCap(n int) {
 }
 
 func (v Value) SetLen(n int) {
-	panic("unimplemented: (reflect.Value).SetLen()")
+	if v.typecode.Kind() != Slice {
+		panic(&ValueError{"reflect.Value.SetLen", v.Kind()})
+	}
+
+	hdr := (*sliceHeader)(v.value)
+	if uintptr(n) > hdr.cap {
```
aykevl: There is another small issue here on AVR (with `SetLen(0x10005)`), would be nice if it works correctly on those chips too.

```suggestion
	if int(uintptr(n)) != n || uintptr(n) > hdr.cap {
```

src/reflect/value.go
```diff
@@ -36,6 +37,10 @@ func (v Value) isExported() bool {
 	return v.flags&valueFlagExported != 0
 }
 
+func (v Value) isRO() bool {
```
aykevl: Can you put in some documentation what this flag means? Because it isn't entirely clear to me (but perhaps will become clear when I look at the next commits). See `isIndirect` and `isExported` above.

Is this to differentiate between `CanAddr` and `CanSet`?

src/reflect/value.go
```diff
@@ -1035,8 +1035,37 @@ func mulNoOverflow(x, y uintptr) (uintptr, bool) {
...
+			typecode: typ.(*rawType),
+			value:    unsafe.Pointer(&zerobuffer[0]),
+			flags:    flags,
+		}
+	}
+
+	return Value{
+		typecode: typ.(*rawType),
+		value:    alloc(typ.Size(), nil),
+		flags:    valueFlagRO | valueFlagIndirect,
```
aykevl: Why is `valueFlagRO` needed? How is it different from not setting either of these flags? (`flags: valueFlag`)? Because that is what would happen when you would use `reflect.ValueOf` on a zero value, like this (see the implementation of `reflect.ValueOf`):

```go
func myZero() reflect.Value {
    var v someType
    return reflect.ValueOf(v)
}
```

I think this would be correct:

```suggestion
		flags:    valueFlagExported,
```

And similarly for the other instances in this function.

src/reflect/value.go
```diff
@@ -992,12 +1008,29 @@ func (v Value) checkRO() {
...
 }
 
+// OverflowInt reports whether the int64 x cannot be represented by v's type.
+// It panics if v's Kind is not Int, Int8, Int16, Int32, or Int64.
 func (v Value) OverflowInt(x int64) bool {
-	panic("unimplemented: reflect.OverflowInt()")
+	switch v.Kind() {
+	case Int, Int8, Int16, Int32, Int64:
+		bitSize := v.typecode.Size() * 8
+		trunc := (x << (64 - bitSize)) >> (64 - bitSize)
```
aykevl, line 1002: This is going to result in _terrible_ machine code on AVR but it looks correct 🤷

src/reflect/value.go
```diff
@@ -870,31 +1002,118 @@ func (v Value) checkAddressable() {
...
+
+func mulNoOverflow(x, y uintptr) (uintptr, bool) {
+	if x <= 1 || y <= 1 {
+		return x * y, false
+	}
+	m := x * y
+	return m, x == m/y
 }
 
+var zerobuffer [256]byte
```
aykevl: Hmm, this buffer is kinda large. It's fine of course on Linux for example but it would take up a lot of space in the very limited RAM of microcontrollers (if it's needed, globals are optimized away when they're not needed).
A few suggestion:
* make it smaller (perhaps 32 bytes or so)
* make it read-only. This is a bit more complicated, one way to do that is by using strings (`"\x00\x00\x00" etc) and using the backing array that is generally stored in read-only memory instead of the very precious RAM.

aykevl (commented):
Reviewed the "small bits" part of this PR.

As a general comment: it would be nice to also add some tests to testdata/reflect.go: it is tested on almost all platforms/architectures (the exception is AVR, but I hope to include even that), and it is a quick way to check whether the TinyGo reflect implementation has identical behavior to the upstream Go implementation (the output should be identical). I've found various bugs that way in the past, for example for `.CanAddr()`.

src/reflect/value.go
dgryski: This is the code from upstream :(

src/reflect/value.go
dgryski: I can make a 32-byte `\x00` string and use an init() function to extract the data field and put it in this global.

src/reflect/value.go
dgryski: The idea for `ro` was mostly taken from upstream.  I couldn't off the top of my head figure out a case where the other flags would fail to do what I needed, and it just seemed "safer" to have an explicit flag for "You Can't Touch This."  If you're sure the semantics can be the same using the flags we have, I'll simplify this PR to remove it.

dgryski:
PTAL

src/reflect/value.go
```diff
@@ -1018,8 +1018,39 @@ func MakeSlice(typ Type, len, cap int) Value {
...
+var zerobuffer unsafe.Pointer
+
+const zerobufferLen = 32
+
+func init() {
+	// 32 characters of zero bytes
+	zerobufferStr := "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
+	s := (*stringHeader)(unsafe.Pointer(&zerobufferStr))
+	zerobuffer = s.data
+}
```
aykevl, line 1070: I verified this gets resolved entirely at compile time, which is good.

src/reflect/value.go
```diff
@@ -217,7 +217,15 @@ func (v Value) CanAddr() bool {
...
 func (v Value) Addr() Value {
-	panic("unimplemented: (reflect.Value).Addr()")
+	if !v.CanAddr() {
+		panic("reflect.Value.Addr of unaddressable value")
+	}
+
+	return Value{
+		typecode: pointerTo(v.typecode),
+		value:    unsafe.Pointer(&v.value),
+		flags:    v.flags,
```
aykevl: I believe you need to remove the indirect flag here:

```suggestion
		flags:    v.flags ^ valueFlagIndirect,
```

because the following code:

```go
package main

import "reflect"

func main() {
        n := 5
        v := reflect.ValueOf(&n)
        println("CanAddr:", v.Elem().Addr().CanAddr())
}
```

Has a different output in TinyGo:

```
$ go run ./tmp/reflect.go
CanAddr: false

$ tinygo run ./tmp/reflect.go
CanAddr: true
```

aykevl (commented):
Unfortunately I found one issue, but the fix is very small. Feel free to rebase and merge if you agree with my proposed fix: the rest looks good to me.

dgryski:
Two review nits fixed.  I'll merge this after CI clears.

deadprogram:
CI in this branch all green!

PR #3498, opened 2023-03-01
reflect: Named type improvements

dgryski:
To be rebased after https://github.com/tinygo-org/tinygo/pull/3497 is merged.


compiler/interface.go
```diff
@@ -169,9 +171,18 @@ func (c *compilerContext) getTypeCode(typ types.Type) llvm.Value {
 		case *types.Basic:
 			typeFields = []llvm.Value{c.getTypeCode(types.NewPointer(typ))}
 		case *types.Named:
+			name := typ.Obj().Name()
+			var buf []llvm.Value
+
+			for _, b := range []byte(name) {
+				buf = append(buf, llvm.ConstInt(c.ctx.Int8Type(), uint64(b), false))
+			}
```
aykevl: I believe the following

```go
			var buf []llvm.Value
			for _, b := range []byte(name) {
				buf = append(buf, llvm.ConstInt(c.ctx.Int8Type(), uint64(b), false))
			}
// ....
				llvm.ConstArray(c.ctx.Int8Type(), buf),
```

Is the same as the following:

```go
				c.ctx.ConstString(name, false),
```

I think the 2nd is easier to read.

src/reflect/type.go
```diff
@@ -487,7 +488,40 @@ func pointerTo(t *rawType) *rawType {
 }
 
 func (t *rawType) String() string {
-	return "T"
+	if t.isNamed() {
+		// TODO(dgryski): `main` until pkgPath support lands
+		return "main" + "." + t.Name()
```
aykevl: I would have called this something like `"<unknown>"` instead of `"main"` but as this is going to change very soon anyway it's no big deal.

aykevl (commented):
This adds 1.36% of code size when I run the driver smoke tests (in my admittedly not very statistically accurate comparison script). About 2/3 has a relatively small impact of up to 100 bytes, the rest has a much larger code size impact of around 1kB or larger.

The PR itself looks good though, but I'm thinking of ways to reduce the overhead a bit especially for programs that don't rely heavily on reflect. A small patch to shrink and move `nlen` (that doesn't actually work: it crashes at runtime) seems to reduce overhead by 0.2% - which is admittedly not very large but would be nice to have.

src/reflect/type.go
dgryski: I can move the `PkgPath` commit into this branch/commit and avoid this altogether.

dgryski:
We can remove the length field here and for PkgPath by switching to null-terminated strings and using the helper I added for that in the struct field methods.  I wonder if trying to deduplicate the names makes any sense.

deadprogram:
@dgryski the `/root/project/src/reflect/type.go:899:17: String not declared by package unsafe` error is happening in the CircleCI build for this branch.

dgryski:
I have integrated the pkgpath PR and replaced both name fields with null-terminated strings instead.

dgryski:
```
~/go/src/github.com/tinygo-org/tinygo $ sizediff size0.txt size1.txt
before  after   diff
100890 100794    -96  -0.1%  wasm testdata/reflect.go
 49866  49781    -85  -0.2%  wasm testdata/filesystem.go
 42290  42219    -71  -0.2%  wasm testdata/atomic.go
 40532  40461    -71  -0.2%  wasm testdata/binop.go
 59451  59380    -71  -0.1%  wasm testdata/interface.go
 63435  63364    -71  -0.1%  wasm testdata/map.go
 43468  43397    -71  -0.2%  wasm testdata/sort.go
 13677  13677      0   0.0%  wasm testdata/alias.go
 21601  21601      0   0.0%  wasm testdata/calls.go
 51954  51954      0   0.0%  wasm testdata/channel.go
 27196  27196      0   0.0%  wasm testdata/env.go
 20768  20768      0   0.0%  wasm testdata/float.go
 14198  14198      0   0.0%  wasm testdata/gc.go
 17189  17189      0   0.0%  wasm testdata/generics.go
 39811  39811      0   0.0%  wasm testdata/goroutines.go
 16009  16009      0   0.0%  wasm testdata/init.go
 13477  13477      0   0.0%  wasm testdata/init_multi.go
 13421  13421      0   0.0%  wasm testdata/ldflags.go
 32788  32788      0   0.0%  wasm testdata/math.go
 21638  21638      0   0.0%  wasm testdata/print.go
 31970  31970      0   0.0%  wasm testdata/rand.go
 13128  13128      0   0.0%  wasm testdata/recover.go
 18501  18501      0   0.0%  wasm testdata/slice.go
 16226  16226      0   0.0%  wasm testdata/string.go
 16986  16986      0   0.0%  wasm testdata/structs.go
 66396  66396      0   0.0%  wasm testdata/timers.go
 13218  13218      0   0.0%  wasm testdata/zeroalloc.go
218606 218620     14   0.0%  wasm testdata/testing.go
215574 215617     43   0.0%  wasm testdata/json.go
105469 105566     97   0.1%  wasm testdata/stdlib.go
sum: -382 (-0.0%)
```

dgryski:
I'm up for having a compiler flag to strip the pkgpath data for smaller systems.

dgryski:
~Importing the `visiblefields` code and tests from upstream revealed some bugs.   Back to draft state for now until I get those sorted.~

Edit: Wrong PR.

aykevl:
Did a quick check to see whether putting the names in a separate global would reduce code size, but it actually increases the size:

```diff
--- a/compiler/interface.go
+++ b/compiler/interface.go
@@ -113,7 +113,8 @@ func (c *compilerContext) getTypeCode(typ types.Type) llvm.Value {
                                types.NewVar(token.NoPos, nil, "ptrTo", types.Typ[types.UnsafePointer]),
                                types.NewVar(token.NoPos, nil, "underlying", types.Typ[types.UnsafePointer]),
                                types.NewVar(token.NoPos, nil, "pkgpath", types.Typ[types.UnsafePointer]),
-                               types.NewVar(token.NoPos, nil, "name", types.NewArray(types.Typ[types.Int8], 1+int64(len(typ.Obj().Name())))),
+                               //types.NewVar(token.NoPos, nil, "name", types.NewArray(types.Typ[types.Int8], 1+int64(len(typ.Obj().Name())))),
+                               types.NewVar(token.NoPos, nil, "name", types.Typ[types.UnsafePointer]),
                        )
                case *types.Chan, *types.Slice:
                        typeFieldTypes = append(typeFieldTypes,
@@ -193,11 +194,23 @@ func (c *compilerContext) getTypeCode(typ types.Type) llvm.Value {
                                llvm.ConstInt(c.ctx.Int32Type(), 0, false),
                        })
 
+                       nameInitializer := c.ctx.ConstString(name+"\x00", false)
+                       nameGlobal := llvm.AddGlobal(c.mod, nameInitializer.Type(), ".namedtype")
+                       nameGlobal.SetInitializer(nameInitializer)
+                       nameGlobal.SetAlignment(1)
+                       nameGlobal.SetUnnamedAddr(true)
+                       nameGlobal.SetLinkage(llvm.InternalLinkage)
+                       nameGlobal.SetGlobalConstant(true)
+                       namePtr := llvm.ConstGEP(nameGlobal.GlobalValueType(), nameGlobal, []llvm.Value{
+                               llvm.ConstInt(c.ctx.Int32Type(), 0, false),
+                               llvm.ConstInt(c.ctx.Int32Type(), 0, false),
+                       })
+
                        typeFields = []llvm.Value{
-                               c.getTypeCode(types.NewPointer(typ)),  // ptrTo
-                               c.getTypeCode(typ.Underlying()),       // underlying
-                               pkgpathPtr,                            // pkgpath pointer
-                               c.ctx.ConstString(name+"\x00", false), // name
+                               c.getTypeCode(types.NewPointer(typ)), // ptrTo
+                               c.getTypeCode(typ.Underlying()),      // underlying
+                               pkgpathPtr,                           // pkgpath pointer
+                               namePtr,                              // name
                        }
                        metabyte |= 1 << 5 // "named" flag
                case *types.Chan:
diff --git a/src/reflect/type.go b/src/reflect/type.go
index 44879dae..849d3f05 100644
--- a/src/reflect/type.go
+++ b/src/reflect/type.go
@@ -429,7 +429,7 @@ type namedType struct {
        ptrTo *rawType
        elem  *rawType
        pkg   *byte
-       name  [1]byte
+       name  *byte
 }
 
 // Type for struct types. The numField value is intentionally put before ptrTo
@@ -910,7 +910,7 @@ func readStringZ(data unsafe.Pointer) string {
 func (t *rawType) Name() string {
        if t.isNamed() {
                ntype := (*namedType)(unsafe.Pointer(t))
-               return readStringZ(unsafe.Pointer(&ntype.name[0]))
+               return readStringZ(unsafe.Pointer(ntype.name))
        }
 
        return t.Kind().String()
```

Result: all my tests (the drivers smoke tests) stay the same or increase in size, summed up together about 0.09%.

dgryski:
I think the only other way to shrink the size impact here is to actually compress the pkgpath data, but we'll need to make sure the decompression code is small enough that we still get a size savings.  Something like how Crack compressed its dictionaries.  (See point Miscellany.7 at https://alecmuffett.com/alecm/software/crack/c50-faq.html )

aykevl:
Some size statistics on the drivers smoke tests (with my flawed sizediff program):

- The first commit (that adds the type names) added 0.64% in size. For most programs, it's less than 100 bytes and therefore a relatively modest addition.
- The second, third, and fourth commit resulted in no change for over half the programs, for the rest 400-800 bytes. Presumably this is because those programs actually use reflect features (so it's for a good reason).
- The fifth commit (that adds the package paths) adds 0.59% on all tests combined, but again most programs only see a small addition while a smaller fraction shows a much larger increase (and those on the higher end that I checked all use fmt, which in turn uses reflect).

I don't think there is much more we can gain* so I think it's good as-is. But let me check the source code again.

\* actually there is, I think it is possible to optimize package paths when they're not used by leveraging LLVM whole program devirtualization. But this is a rather advanced optimization that I haven't tested yet.

targets/cortex-m-qemu.json
```diff
@@ -2,6 +2,7 @@
 	"inherits": ["cortex-m3"],
 	"build-tags": ["qemu", "lm3s6965"],
 	"linkerscript": "targets/lm3s6965.ld",
+        "default-stack-size": 4096,
```
aykevl, line 5: Nit: please use tabs here for consistency.

aykevl (approved):
Looks good to me. It would be nice to fix the one inconsistency, but feel free to merge whether or not you do it.

targets/cortex-m-qemu.json
dgryski: I'll fix this in a later PR so we can merge this one.

aykevl:
> I don't think there is much more we can gain* so I think it's good as-is. But let me check the source code again.
> 
> \* actually there is, I think it is possible to optimize package paths when they're not used by leveraging LLVM whole program devirtualization. But this is a rather advanced optimization that I haven't tested yet.

I tried this, and found out that virtual function elimination only works with _functions_, not with global variables. Perhaps we can extend LLVM to support this, but I'll let it rest for now.

PR #3502, opened 2023-03-01
reflect: add FieldByName() and FieldByIndex()

dgryski:
For when https://github.com/tinygo-org/tinygo/pull/3498 is merged.

### DIFF
--- a/compiler/interface.go
+++ b/compiler/interface.go
@@ -57,6 +57,7 @@ const (
 	structFieldFlagAnonymous = 1 << iota
 	structFieldFlagHasTag
 	structFieldFlagIsExported
+	structFieldFlagIsEmbedded
 )
 
 // createMakeInterface emits the LLVM IR for the *ssa.MakeInterface instruction.
@@ -242,6 +243,9 @@ func (c *compilerContext) getTypeCode(typ types.Type) llvm.Value {
 				}
 				if token.IsExported(field.Name()) {
 					flags |= structFieldFlagIsExported
+				}
+				if field.Embedded() {
+					flags |= structFieldFlagIsEmbedded
 				}
 				data := string(flags) + field.Name() + "\x00"
 				if typ.Tag(i) != "" {
--- a/compiler/interface.go
+++ b/compiler/interface.go
@@ -84,6 +84,27 @@ func (b *builder) extractValueFromInterface(itf llvm.Value, llvmType llvm.Type) 
 	return b.emitPointerUnpack(valuePtr, []llvm.Type{llvmType})[0]
 }
 
+func (c *compilerContext) pkgPathPtr(pkgpath string) llvm.Value {
+	pkgpathName := "reflect/types.type.pkgpath.empty"
+	if pkgpath != "" {
+		pkgpathName = "reflect/types.type.pkgpath:" + pkgpath
+	}
+
+	pkgpathInitializer := c.ctx.ConstString(pkgpath+"\x00", false)
+	pkgpathGlobal := llvm.AddGlobal(c.mod, pkgpathInitializer.Type(), pkgpathName)
+	pkgpathGlobal.SetInitializer(pkgpathInitializer)
+	pkgpathGlobal.SetAlignment(1)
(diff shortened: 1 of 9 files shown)


dgryski:
Rebased to remove duplicate definition of `readStringZ` now that it's in the `names` branch.

deadprogram:
CircleCI fail:

`/root/project/src/reflect/type.go:580:16: String not declared by package unsafe`

https://app.circleci.com/pipelines/github/tinygo-org/tinygo/9018/workflows/1d36a9da-6c8a-4d0d-a790-c1c6d9edbfca/jobs/39879?invite=true#step-118-9

dgryski:
`unsafe.String` removed.  This should fix the CI failure on older Go versions.

dgryski:
Importing the visiblefields code and tests from upstream revealed some bugs. Back to draft state for now until I get those sorted.

dgryski:
With the extra tests from `VisibleFields` I'm pretty confident in these routines now.  PTAL.

dgryski:
Failing CI due to legit reflect failure.  Looking into it..

dgryski:
PTAL.

dgryski:
Split out unrelated fixes to their own PRs.

src/reflect/value.go
```diff
@@ -1395,7 +1395,24 @@ func (v Value) SetMapIndex(key, elem Value) {
...
+		panic(&ValueError{"FieldByIndex", v.Kind()})
+	}
+	for i, x := range index {
+		if i > 0 {
+			if v.Kind() == Pointer && v.typecode.elem().Kind() == Struct {
+				if v.IsNil() {
+					panic("reflect: indirection through nil pointer to embedded struct")
+				}
+				v = v.Elem()
+			}
```
aykevl, line 1411: Interesting that this is needed for Value but not for Type. I guess the logic is that you'd want to index into fields like in plain Go (`.fieldName`) but that it doesn't make much sense for types?

src/reflect/type.go
```diff
@@ -611,49 +645,56 @@ func (t *rawType) rawField(n int) rawStructField {
...
-	if flagsByte&structFieldFlagIsExported == 0 {
-		// This field is unexported.
-		// TODO: list the real package path here. Storing it should not
-		// significantly impact binary size as there is only a limited
-		// number of packages in any program.
-		pkgPath = "<unimplemented>"
+// rawFieldByName returns nearly the same value as FieldByName but without converting the
+// Type member to an interface.
+//
+// For internal use only.
```
aykevl, line 657: "For internal use only" is already implied by the fact that the method is not exported.

compiler/interface.go
```diff
@@ -84,6 +84,27 @@ func (b *builder) extractValueFromInterface(itf llvm.Value, llvmType llvm.Type)
 	return b.emitPointerUnpack(valuePtr, []llvm.Type{llvmType})[0]
 }
 
+func (c *compilerContext) pkgPathPtr(pkgpath string) llvm.Value {
+	pkgpathName := "reflect/types.type.pkgpath.empty"
+	if pkgpath != "" {
+		pkgpathName = "reflect/types.type.pkgpath:" + pkgpath
+	}
+
+	pkgpathInitializer := c.ctx.ConstString(pkgpath+"\x00", false)
```
aykevl, line 93: Note that `c.ctx.ConstString(pkgpath, true)` would do the same.
(I don't really care about which it is, the latter might be more readable).

aykevl (approved):
Interestingly commit 1-5 didn't introduce any size increase (in fact, there was a small reduction in some cases), but commit 6 ([compiler,reflect: fix pkgpath for struct fields](https://github.com/tinygo-org/tinygo/pull/3502/commits/66b4b4ec9817ae51ba5dc69b87ffd67c318d90c5)) introduces a size regresion of about 0.12%. That's not very big, but i wonder whether I can reduce it somehow...

aykevl:
> That's not very big, but i wonder whether I can reduce it somehow...

probably not (my thinking was to skip the pkgpath if it doesn't exist, but in practice almost all structs have a pkgpath)

Issue #3504, opened 2023-03-02
Reflect refactor caused non-deterministic builds


aykevl:
I found that TinyGo does not produce deterministic results anymore:

```
$ tinygo build -o test.hex -target=cortex-m-qemu ./testdata/calls.go && md5sum test.hex
e1bd69209b6fe86f1360befbd7c47d94  test.hex

$ tinygo build -o test.hex -target=cortex-m-qemu ./testdata/calls.go && md5sum test.hex
05b7cbd854f460a1bf3564b280ffea16  test.hex
```

I bisected this to 4e8453167f42976aad87099ffdb3746fc540d6a6, which is the reflect refactor.

aykevl:
Smaller reproducer:

```go
package main

type Thing struct {
	name string
}

func (t Thing) Foo() {
}

type Printer interface {
	Foo()
}

func main() {
	var t Printer = &Thing{"foo"}
	defer t.Foo()
}
```

aykevl:
This was introduced and fixed during the current release cycle, so closing.

PR #3505, opened 2023-03-02
transform: fix non-determinism in the interface lowering pass

aykevl:
This non-determinism was introduced in
https://github.com/tinygo-org/tinygo/pull/2640. Non-determinism in the compiler is a bug because it makes it harder to find whether a compiler change actually affected the binary.

Fixes https://github.com/tinygo-org/tinygo/issues/3504


dgryski (approved):
LGTM

deadprogram:
Merging, thanks @aykevl and @dgryski

PR #3509, opened 2023-03-03
syscall Stat fixes

dgryski:


dgryski:
`archive/tar` passes on Darwin and would otherwise pass on wasi but `internal/testenv` doesn't know that wasi doesn't have symlinks (and we don't support `t.Skipf`).

src/syscall/syscall_libc_darwin.go
```diff
@@ -168,10 +173,10 @@ type Stat_t struct {
 	Gid       uint32
 	Rdev      int32
 	Pad_cgo_0 [4]byte
-	Atim      Timespec
-	Mtim      Timespec
-	Ctim      Timespec
-	Btim      Timespec
+	Atimespec Timespec
```
dkegel-fastly, line 175: Probably want to remove the misleading comment above, mea culpa.

dkegel-fastly:
If that last commit allows archive/tar to build and pass, should it also add it to TEST_PACKAGES_DARWIN or TEST_PACKAGES_LINUX in the toplevel Makefile?

src/syscall/syscall_libc_darwin.go
dgryski: Good point. I'll remove the comment.

dgryski:
Yeah was thinking about adding it for Darwin. Need the see what's preventing it from running on Linux.

dgryski:
For the record,
```
~/go/src/go.googlesource.com/go/src/archive/tar $ egrep '(go:build)|([AC]tim)' stat_actime*.go
stat_actime1.go://go:build aix || linux || dragonfly || openbsd || solaris
stat_actime1.go:func statAtime(st *syscall.Stat_t) time.Time {
stat_actime1.go:	return time.Unix(st.Atim.Unix())
stat_actime1.go:func statCtime(st *syscall.Stat_t) time.Time {
stat_actime1.go:	return time.Unix(st.Ctim.Unix())
stat_actime2.go://go:build darwin || freebsd || netbsd
stat_actime2.go:func statAtime(st *syscall.Stat_t) time.Time {
stat_actime2.go:	return time.Unix(st.Atimespec.Unix())
stat_actime2.go:func statCtime(st *syscall.Stat_t) time.Time {
stat_actime2.go:	return time.Unix(st.Ctimespec.Unix())
```

dgryski:
Looks like this probably needs `-tags=osusergo`, but adding that to all the testing make rules seems like a larger change than I'd like for a PR about fixing the `Stat_t` field names.

dkegel-fastly (approved):
LGTM

Would have liked this commit to include some change to the tests that gives some evidence the change improves things, but if that's hard, fine.

PR #3511, opened 2023-03-03
runtime: make runtime panics recover()-able

dgryski:


aykevl:
Did you check the binary size impact?
In any case, this results in an unnecessary heap allocation in a very sensitive area (for example, when an "out of memory" runtime panic is triggered, allocating would lead to infinite recursion). So I suggest factoring out the recover part of `runtime._panic` to avoid the conversion to an interface.

(I'm a bit careful with this change, as I think there might be reasons why it's the way it is now. Just want to make sure we don't overlook anything).

dgryski:
No, I didn't. I also noticed that CI failed. Marking as draft for now.

aykevl:
Actually, the fact that CI fails is because of this heap allocation:

```
/Users/runner/work/tinygo/tinygo/build/tinygo build -size short -o test.hex -target=pca10040 -gc=none -scheduler=none examples/blinky1
tinygo:ld.lld: error: undefined symbol: runtime.alloc
>>> referenced by string.go:62 (/Users/runner/work/tinygo/tinygo/src/runtime/string.go:62)
>>>               /Users/runner/Library/Caches/tinygo/thinlto/llvmcache-059CC3AC26AD6D48C9F8BD6A8548A01527722C0E:(runtime.stringConcat)
```

dgryski:
This is needed to make the `bytes` package from 1.23 pass tests.

aykevl:
I have made a new PR with a somewhat different approach: https://github.com/tinygo-org/tinygo/pull/4758

aykevl (changes requested):
I don't think we should make all runtime panics recoverable. Instead, we should take a look at which panics make sense to recover from. See #4758 for example.

dgryski:
Proper fix in https://github.com/tinygo-org/tinygo/pull/5438

PR #3512, opened 2023-03-03
Support GPIO and SPI for ARM Linux SBC

oflebbe:
Hi,

support for generic ARM Linux Single Board computer GPIO and SPI via linux devices /dev/gpiochip0 and /dev/spidev0
without any additional C support, so tinygo can be used for crosscompiling.

Introduces two new targets: raspberrypi32 and raspberrypi64 .

Blinky1 (GPIO) and the st7789 (SPI) had been working for me on a raspberry pi . 

Devices need to be enabled via raspi-config first, in order to load kernel modules.

What do you think ?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ test.hex
 test.nro
 test.wasm
 wasm.wasm
+.DS_Store
--- a/src/machine/machine_generic.go
+++ b/src/machine/machine_generic.go
@@ -1,4 +1,4 @@
-//go:build !baremetal
+//go:build !baremetal && !raspberrypi
 
 package machine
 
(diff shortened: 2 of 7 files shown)


aykevl:
I kind of like the idea, but because these devices are not nearly as much resource constrained, what about doing this dynamically at runtime? It can detect the SBC variant it's running on and use the appropriate API. Cross compiling is still possible, the way it's always been done: using `GOOS` and `GOARCH`.

oflebbe:
Thanks for the feedback!  I can totally see the point that it might not be the best fit to tinygo, since it doesn't depend on anything within tinygo, it only happens to implement tinygo API's for gpio, spi for use with tinygo drivers. Maybe it would make sense to have it as an independent library even for use with vanilla go, tinygo drivers can be used on top of it either ways.

So I am considering withdrawing the proposal.

PR #3513, opened 2023-03-04
reflect: add support for remaining map types

dgryski:


src/reflect/value.go
```diff
@@ -816,6 +819,11 @@ func (v Value) MapIndex(key Value) Value {
...
 		}
 		return elem.Elem()
+	} else {
+		if ok := hashmapInterfaceGet(v.pointer(), key.Interface(), elem.value, elemType.Size()); !ok {
+			return Value{}
+		}
+		return elem.Elem()
 	}
 
 	// TODO(dgryski): Add other map types.  For now, just string and binary types are supported.
```
aykevl: This code is now unreachable, so could be deleted.

aykevl (commented):
One comment, otherwise looks good to me.
(Feel free to merge with the fix).

src/reflect/value.go
dgryski: Oops. thanks.

PR #3515, opened 2023-03-04
Refactor EnableInterrupts and DisableInterrupts

waj334:
Removes usage of AsmFull which required an optimization pass to remove the map parameter passed into it. This caused issues when compiling with -opt=0 where memory for the map was being allocated as an unintended side-effect of using AsmFull since no optimizations are enabled.

### DIFF
--- a/src/device/arm/arm.go
+++ b/src/device/arm/arm.go
@@ -29,6 +29,7 @@
 //	POSSIBILITY OF SUCH DAMAGE.
 package arm
 
+import "C"
 import (
 	"errors"
 	"runtime/volatile"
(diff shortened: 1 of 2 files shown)


src/device/arm/arm.go
```diff
@@ -174,20 +175,13 @@ func SetPriority(irq uint32, priority uint32) {
...
 // DisableInterrupts disables all interrupts, and returns the old interrupt
 // state.
-func DisableInterrupts() uintptr {
-	return AsmFull(`
-		mrs {}, PRIMASK
-		cpsid i
-	`, nil)
-}
+//go:linkname DisableInterrupts DisableInterrupts
+func DisableInterrupts() uintptr
```
aykevl, line 180: Please use `//export` instead of `//go:linkname`. The reason is that `//go:linkname` uses the TinyGo calling convention, while `//export` uses the C calling convention. There might be a small difference between them.

src/device/arm/interrupts.c
```diff
@@ -0,0 +1,22 @@
+#include <stdint.h>
+
+void EnableInterrupts(uintptr_t mask) {
+asm volatile(
+	"msr PRIMASK, %0"
+	:
+	: "r"(mask)
+	: "memory"
+);
```
aykevl: Can you indent this code? That looks better to me at least.
(...I've gotten so used to auto-formatted code that I feel weird about commenting about style issues).

aykevl (commented):
Thank you! Happy to see `AsmFull` go away.

waj334:
> Thank you! Happy to see `AsmFull` go away.

There's a couple more spots (4) where AsmFull is still being used. 1 out of the 4 is arm so I can refactor that and test, but the remaining are xtensa (2) and avr (1) which I don't really have anything to test on.

deadprogram:
Thank you for the refactor @waj334 now merging.

deadprogram:
Hmm, I perhaps should have tested on the TinyHCI before merge. Seems like the ItsyBitsy M4 crashes on start with this commit. Probably any M4 might be affected?

deadprogram:
A subsequent commit passed the CI, so that seems like problem must be unrelated to this PR.

PR #3518, opened 2023-03-05
Refactor SystemStack function for arm targets.

waj334:
Removing usage of arm.AsmFull in favor of writing inline assembly in C.

### DIFF
--- a/src/internal/task/task_stack_cortexm.c
+++ b/src/internal/task/task_stack_cortexm.c
@@ -1,3 +1,4 @@
+//go:build scheduler.tasks && cortexm
 #include <stdint.h>
 
 uintptr_t SystemStack() {
--- a/src/internal/task/task_stack_cortexm.c
+++ b/src/internal/task/task_stack_cortexm.c
@@ -1,0 +1,12 @@
+#include <stdint.h>
+
+uintptr_t SystemStack() {
+    uintptr_t sp;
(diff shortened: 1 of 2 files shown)


waj334:
@aykevl I'm not sure why this one is broken. The existence of task_stack_cortexm.c seems to upset the compiler during the tests.

src/internal/task/task_stack_cortexm.c
```diff
@@ -0,0 +1,12 @@
+#include <stdint.h>
```
aykevl, line 2: I think you need to add build tags here, something like this:

```c
//go:build scheduler.tasks && cortexm
```

waj334:
@aykevl Fixed

deadprogram:
Looks correct to me. @aykevl any last feedback here?

aykevl:
Looks good to me. Might be worth rebasing against the dev branch to make sure it still works with https://github.com/tinygo-org/tinygo/pull/3213, or we could just merge it and hope it doesn't break the test.

deadprogram:
> Might be worth rebasing against the dev branch to make sure it still works with #3213

That is a good idea. Could you please do that @waj334 ?

waj334:
@deadprogram done

deadprogram:
Thank you very much for this @waj334 now merging.

PR #3520, opened 2023-03-05
device, machine, runtime/interrupt: implement GBA register definitions

deadprogram:
This PR implements a number of GBA register definitions mostly obtained from https://problemkaputt.de/gbatek.htm

It also switches the example to use the same register defs.


aykevl (approved):
That's a great improvement, thanks!

Issue #3506, opened 2023-03-02
Race condition


aykevl:
We have a race condition:

```
==================
WARNING: DATA RACE
Write at 0x00c000f23a38 by goroutine 39:
  golang.org/x/tools/go/ssa.(*Function).newBasicBlock()
      /home/ayke/pkg/mod/golang.org/x/tools@v0.5.1-0.20230114154351-e035d0c426c8/go/ssa/func.go:668 +0x468
  golang.org/x/tools/go/ssa.(*Function).startBody()
      /home/ayke/pkg/mod/golang.org/x/tools@v0.5.1-0.20230114154351-e035d0c426c8/go/ssa/func.go:163 +0x494
  golang.org/x/tools/go/ssa.(*builder).buildFunctionBody()
      /home/ayke/pkg/mod/golang.org/x/tools@v0.5.1-0.20230114154351-e035d0c426c8/go/ssa/builder.go:2389 +0x294
  golang.org/x/tools/go/ssa.(*builder).buildFunction()
      /home/ayke/pkg/mod/golang.org/x/tools@v0.5.1-0.20230114154351-e035d0c426c8/go/ssa/builder.go:2326 +0x5c
  golang.org/x/tools/go/ssa.(*builder).buildCreated()
      /home/ayke/pkg/mod/golang.org/x/tools@v0.5.1-0.20230114154351-e035d0c426c8/go/ssa/builder.go:2413 +0x48
  golang.org/x/tools/go/ssa.(*Package).build()
      /home/ayke/pkg/mod/golang.org/x/tools@v0.5.1-0.20230114154351-e035d0c426c8/go/ssa/builder.go:2606 +0xfec
  golang.org/x/tools/go/ssa.(*Package).build-fm()
      <autogenerated>:1 +0x34
  sync.(*Once).doSlow()
      /usr/local/go/src/sync/once.go:74 +0xb0
  sync.(*Once).Do()
      /usr/local/go/src/sync/once.go:65 +0x40
  golang.org/x/tools/go/ssa.(*Package).Build()
      /home/ayke/pkg/mod/golang.org/x/tools@v0.5.1-0.20230114154351-e035d0c426c8/go/ssa/builder.go:2477 +0x2bc
  github.com/tinygo-org/tinygo/compiler.CompilePackage()
      /home/ayke/src/tinygo/tinygo/compiler/compiler.go:280 +0x294
  github.com/tinygo-org/tinygo/builder.Build.func3()
[...]

PR #3522, opened 2023-03-06
compiler: fix a race condition

aykevl:
There was a mostly benign race condition in the compiler. The issue was that there is a check for type aliases (which can alias types in another function), but this check was _after_ accessing a property of the function that might not have been completed.

I don't think this can have any serious effects, as the function is skipped anyway, but such bugs should certainly be fixed.

---

Pretty sure this fixes https://github.com/tinygo-org/tinygo/issues/3506. I can't get any race conditions with this patch applied.


deadprogram (approved):
LGTM

deadprogram:
Now merging, thanks @aykevl

PR #3523, opened 2023-03-06
build/windows: trying to increase stack size for bash/make

deadprogram:
This PR attempts to increase stack size for bash/make. 

See https://github.com/microsoft/WSL/issues/633 and https://github.com/actions/runner-images/issues/3738

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -99,7 +99,9 @@ jobs:
           scoop install wasmtime
       - name: Test TinyGo
         shell: bash
-        run: make test GOTESTFLAGS="-short"
+        run: |
+          ulimit -s unlimited
+          make test GOTESTFLAGS="-short"
       - name: Build TinyGo release tarball
         shell: bash
         run: make build/release -j4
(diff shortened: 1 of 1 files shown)


deadprogram:
Not going to work this way. Closing until I come up with another idea.

aykevl:
I'm 95% sure the stack is not the issue, but rather some sort of race condition (probably that we don't have threads enabled, see #3130).
If the issue was a small stack, then it would segfault, not log a warning sometimes.

I wonder how big the stacks actually are though, because it can't be 8kB - that's way too small, and we probably would have seen many segfaults if it were that small. (For reference, we already ran into problems with a 128kB stack on Alpine Linux).

aykevl:
[Here is why](https://github.com/llvm/llvm-project/blob/main/clang/lib/Basic/Stack.cpp#L21):

```cpp
static LLVM_THREAD_LOCAL void *BottomOfStack = nullptr;
```

If we're compiling with threads disabled on Windows, I doubt this is actually a thread local variable and it will cause chaos if two threads are setting their own bottom of stack. Probably the reason this doesn't trigger all the time is that there is a heuristic in the stack check for unexpected values, and usually the stacks aren't close enough to trigger it.

PR #3524, opened 2023-03-06
build/windows: switch to MSYS64 bash/make

deadprogram:
This PR tries switching to the MSYS64 version of bash.


deadprogram:
Closing, seems like not needed at present.

PR #3525, opened 2023-03-06
windows: disable parallelism on Windows

aykevl:
This is an attempt to figure out why the Windows CI keeps crashing. It should be removed before the next release, by which point we should know whether this has helped or not.

---

You might say I'm getting desperate looking for solutions.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1508,6 +1508,15 @@ func main() {
 		defer pprof.StopCPUProfile()
 	}
 
+	// Limit the number of threads to one.
+	// This is an attempted workaround for the crashes we're seeing in CI on
+	// Windows. If this change helps, it indicates there is a concurrency issue.
+	// If it doesn't, then there is something else going on. Either way, this
+	// should be removed once the test is done.
+	if runtime.GOOS == "windows" {
+		runtime.GOMAXPROCS(1)
+	}
+
 	switch command {
 	case "build":
 		pkgName := "."


deadprogram (approved):
I think this worked. There is a separate issue with Windows CI and the std tests for the `net` package, but that is I think unrelated.

Maybe we should put this behind a flag, if we are unable to resolve otherwise before the next release.

deadprogram:
Merging to get the build back.

deadprogram:
```
--- FAIL: TestCompiler (10.04s)
    --- FAIL: TestCompiler/goroutine.go-cortex-m-qemu-tasks (0.7[9](https://github.com/tinygo-org/tinygo/actions/runs/4346376523/jobs/7592354126#step:16:10)s)
Error:         compiler_test.go:112: could not parse test case goroutine.go: could not compile: <cgo>:1:1: error: stack nearly exhausted; compilation time may suffer, and crashes due to stack overflow are likely
FAIL
FAIL	github.com/tinygo-org/tinygo/compiler	[10](https://github.com/tinygo-org/tinygo/actions/runs/4346376523/jobs/7592354126#step:16:11).249s
```

https://github.com/tinygo-org/tinygo/actions/runs/4346376523/jobs/7592354126#step:16:11

aykevl:
Hmm, that seems bad.

Maybe leave it in at least a few days so we can get a better feel whether anything changed at all?

deadprogram:
Let's let it run for a bit and see.

PR #3526, opened 2023-03-06
build/linux: switch to ubuntu-20.04 for arm builds

deadprogram:
This PR switches to ubuntu-20.04 for arm builds since ubuntu-18.04 is deprecated for GH actions runner. See https://github.com/actions/runner-images/issues/6002


deadprogram:
Rebased in `dev` with the Windows fix, so this PR should now pass :black_cat:

aykevl (approved):
LGTM

PR #3527, opened 2023-03-06
Add CryptoAuth Trust Platform to the target list

hugolgst:
https://www.microchip.com/en-us/development-tool/DM320118

### DIFF
--- a/src/machine/board_cryptoauth.go
+++ b/src/machine/board_cryptoauth.go
@@ -20,7 +20,10 @@ var (
 )
 
 var (
-	I2C0 = sercomI2CM1
+	I2C0 = sercomI2CM0
+	I2C1 = sercomI2CM1
+	I2C2 = sercomI2CM2
+	I2C3 = sercomI2CM3
 )
 
 const (
--- a/src/machine/board_cryptoauth.go
+++ b/src/machine/board_cryptoauth.go
@@ -1,0 +1,45 @@
+package machine
+
+const (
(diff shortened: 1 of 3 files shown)


src/machine/board_cryptoauth.go
```diff
@@ -0,0 +1,48 @@
+package machine
```
deadprogram, line 2: This file needs to have the following build tag in order to work correctly:

```go
//go:build sam && atsamd21 && cryptoauth
```

src/machine/board_cryptoauth.go
```diff
@@ -0,0 +1,48 @@
...
+	UART_RX_PIN = PA23
+)
+
+var (
+	DefaultUART = &sercomUSART0
+)
+
+var (
+	I2C0 = sercomI2CM0
+	I2C1 = sercomI2CM1
```
deadprogram, line 25: You can probably remove I2C1 thru I2C3 and just leave I2C0.

deadprogram:
Hello @hugolgst just wondering if you are still working with this hardware, and are interested/able to try to update this PR? Thanks.

hugolgst:
Unfortunately not but it would be cool to update it if possible

src/machine/board_cryptoauth.go
hugolgst:
```suggestion
//go:build sam && atsamd21 && cryptoauth
package machine
```

PR #3528, opened 2023-03-07
machine/atsamd51: remove extra BK0RDY clear

sago35:
Before the modification, handleUSBIRQ() clears BK0RDY first, and then handleUSBSetAddress() clears it as well.
If there is a USB reception after the first BK0RDY clear, the data is lost.

https://github.com/tinygo-org/tinygo/blob/v0.27.0/src/machine/machine_atsamd51_usb.go#L145-L146

https://github.com/tinygo-org/tinygo/blob/v0.27.0/src/machine/machine_atsamd51_usb.go#L291-L292


deadprogram:
Thanks for the correction @sago35 now merging.

PR #3529, opened 2023-03-07
reflect: NumMethods

dgryski:


src/reflect/type.go
```diff
@@ -428,10 +432,11 @@ type mapType struct {
...
 	rawType
-	ptrTo *rawType
-	elem  *rawType
-	pkg   *byte
-	name  [1]byte
+	ptrTo     *rawType
+	elem      *rawType
+	numMethod uint16
+	pkg       *byte
+	name      [1]byte
```
aykevl: I believe you can save a little bit of storage by packing the struct more tightly. Right now, there will always be two padding bytes after `numMethod` on a 32-bit architecture. To avoid it, you could move the field to just after `rawType` (most optimal, takes up no extra space as it fits entirely in the padding area) or just before `name` (avoids the padding bytes as the next field is the variable-sized `name`).

src/reflect/type.go
```diff
@@ -409,7 +412,8 @@ type elemType struct {
 
 type ptrType struct {
 	rawType
-	elem *rawType
+	elem      *rawType
+	numMethod uint16
```
aykevl: Same here. By moving `numMethod` to after `rawType`, you fill the gap and it won't result in any additional storage on a 32-bit architecture.

src/reflect/type.go
```diff
@@ -443,10 +448,11 @@ type namedType struct {
...
 	rawType
-	numField uint16
-	ptrTo    *rawType
-	pkgpath  *byte
-	fields   [1]structField // the remaining fields are all of type structField
+	numField  uint16
+	numMethod uint16
+	ptrTo     *rawType
+	pkgpath   *byte
+	fields    [1]structField // the remaining fields are all of type structField
```
aykevl: The same actually happens here. Before, `rawField` and `numField` were taking up 4 bytes together including padding (on a 32-bit architecture). But I don't see a way to avoid the extra padding in this case.

src/reflect/type.go
```diff
@@ -983,7 +989,20 @@ func (t *rawType) NumOut() int {
...
+
+	switch t.Kind() {
+	case Pointer:
+		return int((*ptrType)(unsafe.Pointer(t)).numMethod)
+	case Struct:
+		return int((*structType)(unsafe.Pointer(t)).numMethod)
+	}
+
+	// Other types have no methods attached.  Note we don't panic here.
+	return 0
```
aykevl, line 1024: I really hope we don't overlook something here...

src/reflect/type.go
```diff
@@ -498,7 +498,13 @@ func pointerTo(t *rawType) *rawType {
...
 func (t *rawType) String() string {
 	if t.isNamed() {
-		return t.PkgPath() + "." + t.Name()
+		path := t.PkgPath()
+		slash := len(path) - 1
+		for slash >= 0 && path[slash] != '/' {
+			slash--
+		}
+		shortened := path[slash+1:]
+		return shortened + "." + t.Name()
```
aykevl: Are you reading the package name here?
Note that the package name can be different from the package path. For example, the package tinygo.org/x/go-llvm has package name llvm (not go-llvm). But using the ending of the package path might be good enough in practice.

aykevl (commented):
Reviewed just the last 4 commits.

Regarding struct packing, this might be useful: http://www.catb.org/esr/structure-packing/
In this case, it would help to keep the structs mostly the same. I recommend changing the layout to something like this:

```go
type structArrayOrPtrType struct {
    rawType
    numMethod      uint16
    ptrTo/elemType *rawType // name varies by type
}
```

src/reflect/type.go
dgryski: Yeah, I  wasn't paying any attention to the struct packing, preferring to get it working first.   I can go back and tweak these layouts to minimize padding.

src/reflect/type.go
dgryski: You're right.  I'll fix this.

dgryski: Since PkgPath and packge name might not be the same, we'll need to store that field separately for the `namedType` struct.  We might be able to do something sneaky here to save space.

dgryski: Sample patch (compiler changes not included):
```
@@ -1026,10 +1020,21 @@ func readStringZ(data unsafe.Pointer) string {
        }))
 }

+func (t *rawType) name() string {
+       ntype := (*namedType)(unsafe.Pointer(t))
+       return readStringZ(unsafe.Pointer(&ntype.name[0]))
+}
+
        if t.isNamed() {
-               ntype := (*namedType)(unsafe.Pointer(t))
-               return readStringZ(unsafe.Pointer(&ntype.name[0]))
+               pkgDotName := t.name()
+               // strip the package and the `.` that appear before the type name
+               for i := 0; i < len(pkgDotName); i++ {
+                       if pkgDotName[i] == '.' {
+                               return pkgDotName[i+1:]
+                       }
+               }
+               panic("corrupt name field for named type")
```

dgryski:
PTAL.

dgryski:
Fixed the optimization pass to use the new layout of the `named` type.

dgryski:
Looks like my optimization fix is still breaking stuff..

dgryski:
Transform test fixed.

deadprogram (approved):
Looking good!

deadprogram:
Now merging, thanks @dgryski and @aykevl

PR #3533, opened 2023-03-08
targets: format json files

dgryski:
for i in *.json; do
             jq --tab . $i >$i.jq && mv $i.jq $i;
	done


dgryski:
This makes all the json files machine formatted for consistency, but makes some of the entries harder to grep for.

aykevl:
I'm not opposed to this change (consistency is good!), but having `"inherits"` and `"build-tags"` on a single line would make things easier. I have been using grep from time to time on the `"inherits"` line which will become harder with this change.

I suppose special-casing these two keys going to be difficult?
Also, it would be really nice to check for this formatting in CI so that it doesn't regress in the future.

dgryski:
We could do that kind of custom formatting, but I'm not sure how easy it would be to do with `jq`.  Easier might be to write a Go program to keep it correct.  We can close this PR, but it is interesting to note how inconsistent the targets files are.

dgryski:
Now that tinygo will support `encoding/json`, it should be easier to write a custom target formatter that does what we want.

dgryski:
Actually not sure why we can't just write the formatter in standard Go.

aykevl:
Thinking about this again, I agree this is a good idea.

However... if we're going to do this, what about switching to YAML instead? This has been discussed before (for example in #2214, #2222) and if we're going to do a big rewrite we might as well switch to the usual YAML syntax.

(YAML is backwards compatible so we could also do the YAML switch in a separate PR and not touch targets/*.json at all - leave them in the usual JSON syntax).

deadprogram:
Closing due to age. Please open a new PR against latest `dev` if needed. Thank you!

PR #3534, opened 2023-03-08
compiler: only define the package path once

aykevl:
Previously the compiler would create multiple package path globals. This is not intended: the intention is to only create one per package path.

This change (somewhat to my surprise) does not change binary size in any significant way. I suspect this is because the pacakge paths are merged together at a later compilation stage because they're marked as `unnamed_addr`.

### DIFF
--- a/compiler/interface.go
+++ b/compiler/interface.go
@@ -181,13 +181,16 @@ func (c *compilerContext) getTypeCode(typ types.Type) llvm.Value {
 				pkgpathName = "reflect/types.type.pkgpath:" + pkgpath
 			}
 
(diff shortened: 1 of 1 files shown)


dgryski:
This code gets moved in https://github.com/tinygo-org/tinygo/pull/3502 ; can I incorporate this change into my PR?

aykevl:
Sure, go ahead.

dgryski:
This is now part of https://github.com/tinygo-org/tinygo/pull/3529

dgryski:
Closing as a duplicate.

PR #3540, opened 2023-03-11
compiler: add alloc attributes to runtime.alloc

aykevl:
This gives a small improvement now, and is needed to be able to use the Heap2Stack transform that's available in the Attributor pass. This Heap2Stack transform could replace our custom OptimizeAllocs pass.

Most of the changes are just IR that changed, the actual change is relatively small.

To give an example of why this is useful, here is the code size before this change:

    $ tinygo build -o test -size=short ./testdata/stdlib.go
      95620    1812     968 |   97432    2780

    $ tinygo build -o test -size=short ./testdata/stdlib.go
      95380    1812     968 |   97192    2780

That's a 0.25% reduction. Not a whole lot, but nice for such a small patch.

### DIFF
--- a/compiler/symbol.go
+++ b/compiler/symbol.go
@@ -52,6 +52,17 @@ const (
 	inlineNone
 )
 
+// Values for the allockind attribute. Source:
+// https://github.com/llvm/llvm-project/blob/release/16.x/llvm/include/llvm/IR/Attributes.h#L49
+const (
+	allocKindAlloc = 1 << iota
+	allocKindRealloc
+	allocKindFree
+	allocKindUninitialized
+	allocKindZeroed
+	allocKindAligned
+)
+
 // getFunction returns the LLVM function for the given *ssa.Function, creating
 // it if needed. It can later be filled with compilerContext.createFunction().
 func (c *compilerContext) getFunction(fn *ssa.Function) (llvm.Type, llvm.Value) {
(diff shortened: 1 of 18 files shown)


aykevl:
Looks like this is failing under LLVM 14.

aykevl:
LLVM 14 regression should be fixed now.

deadprogram:
Interestingly, this PR causes the `itsybitsy-nrf52840` to fail the TinyHCI. I verified the HW, and then noticed that other PRs were able to still pass.

See https://github.com/tinygo-org/tinygo/pull/3540/checks?check_run_id=11934025697

aykevl:
I ran the entire test corpus, which passed. So if it's a real bug, it's a very specific one.

I also built the itsybitsy-nrf52840 TinyHCI binary locally and there is a difference in the binary output, so a real bug triggered by this change is plausible.

aykevl:
I can reproduce the issue locally.

aykevl:
Huh, there is indeed a rather important bug: somehow the optimizer decided that it's fine to just return 0 instead of the actual value.

aykevl:
Yup, reducing further the compiler converts this line:

```llvm
  %11 = load i16, ptr %value, align 2
  %12 = shl i16 %11, 4
  ret i16 %12
```

Into this:

```llvm
  ret i16 0
```

(`%value` here is `var value int16` from the beginning of the `.Get()` function, and these lines are the IR of the last lines of `.Get()`: `return uint16(value << 4)`).

The reasoning appears to be that `runtime.alloc` returns a zeroed allocation (correct), and there is no store to this before the load at the end (not entirely correct). The problem is that while there is no store, the pointer is stored into a register which captures the pointer (and can therefore lead to arbitrary changes to the value).
In other words, this appears like a LLVM bug.

aykevl:
Reported the bug upstream: https://github.com/llvm/llvm-project/issues/61362

I'll convert this to a draft until the bug is resolved in some way.

aykevl:
Fixed the issue for the nrf. I looked for other issues like this, and there may be some, but not as clear as this one.
(To be clear, the code has always been broken, the optimizer just didn't catch it until we added the alloc attributes).

deadprogram:
Have not been able to get this PR to pass the CircleCI with the older version compatibility tests. Perhaps it in fact is not compatible?

deadprogram:
Perhaps rebase against `dev` might convince it to pass?

aykevl:
No, I can reproduce it locally. This looks more like a bug in the PR.

aykevl:
Whoops, that's what you get from working with two different git repos. I force-pushed an older branch which re-introduced the LLVM 14 bug.
With the latest push, it should finally be fixed.

deadprogram:
All green at last, thank you very much @aykevl now merging.

PR #3541, opened 2023-03-11
wioterminal: fix pin definition of BCM13

sago35:
This PR fixes the pin definition of BCM13

PinOut: https://files.seeedstudio.com/wiki/Wio-Terminal/img/WioT-Pinout.jpg
Schematics: https://files.seeedstudio.com/wiki/Wio-Terminal/res/Wio-Terminal-SCH-v1.2.pdf

image


deadprogram:
Thank you for the correction @sago35 now merging.

PR #3542, opened 2023-03-11
machine/rp2040: change uart to allow for a single pin

sago35:
This PR allows the RP2040's UART to be used only for either transmitting or receiving.
This is especially useful when using a device such as the XIAO-RP2040, which has a small number of pins.

### DIFF
--- a/src/machine/machine_rp2040_uart.go
+++ b/src/machine/machine_rp2040_uart.go
@@ -41,8 +41,12 @@ func (uart *UART) Configure(config UARTConfig) error {
 		rp.UART0_UARTCR_TXE)
 
 	// set GPIO mux to UART for the pins
-	config.TX.Configure(PinConfig{Mode: PinUART})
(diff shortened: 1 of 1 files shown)


deadprogram:
That is interesting, so this actual works correctly? Very useful, certainly. Should we allow this behavior for other platforms?

sago35:
At least it works well on my XIAO-RP2040 keyboard.
Either way, there is no reason to initialize NoPin, so I see no problem with the change.

It would be wonderful if other microcontrollers could work in the same way.

sago35:
My English may have been bad.

Connects TX of XIAO-RP2040-1 and RX of XIAO-RP2040-2.
Connect nothing to RX of XIAO-RP2040-1.
Connect nothing to TX of XIAO-RP2040-2.

deadprogram:
Thanks @sago35 now merging.

PR #3543, opened 2023-03-12
reflect: add Convert() 

dgryski:


aykevl:
I guess this is only about the last commit?

Any reason all the other commits are included? Because this seems like a standalone change to me (and the last commit looks good to me).

dgryski:
The last commit was the issue we identified but I did end up stacking it onto my latest PR because that was needed to run all the various test sites and JSON programs I was working on.

I can try to extract the standalone fixes from this branch I guess but they won't all have associated tests since those are mostly cumulative.

dgryski:
This was branched from the end of NumMethods PR.

dgryski:
All unrelated fixes removed. Rebased against `dev`.

dgryski:
Still a draft because I should probably add support for converting floats.

dgryski:
Probably worth reviewing and merging
 We can add more cases as we hit them.

deadprogram:
Look good to me, @aykevl any feedback before merge?

src/reflect/value.go
```diff
@@ -1024,7 +1024,97 @@ func (v Value) OverflowUint(x uint64) bool {
...
+		*(*uint16)(ptr) = uint16(bits)
+	case 4:
+		*(*uint32)(ptr) = uint32(bits)
+	case 8:
+		*(*uint64)(ptr) = bits
+	}
+	return Value{
+		typecode: t,
+		value:    ptr,
+		flags:    flags | valueFlagIndirect,
```
aykevl: This is not quite correct. For the following code:

```go
package main

import "reflect"

func main() {
        v1 := reflect.ValueOf(int(0))
        v2 := v1.Convert(reflect.TypeOf(int16(0)))
        println("can set:", v2.CanSet())
}
```

There is a difference in output:

```
$ go run ./tmp/convert.go 
can set: false

$ tinygo run ./tmp/convert.go
can set: true
```

This can be fixed by removing the indirect flag, and storing the value directly in the `value` field if it fits.

src/reflect/value.go
```diff
@@ -1024,7 +1024,97 @@ func (v Value) OverflowUint(x uint64) bool {
...
+	switch size {
+	case 4:
+		*(*float32)(ptr) = float32(v)
+	case 8:
+		*(*float64)(ptr) = v
+	}
+	return Value{
+		typecode: t,
+		value:    ptr,
+		flags:    flags | valueFlagIndirect,
```
aykevl: Same here (I assume, didn't test it).

src/reflect/value_test.go
```diff
@@ -266,6 +266,15 @@ func TestNamedTypes(t *testing.T) {
 	}
 }
 
+func TestConvert(t *testing.T) {
+	v := ValueOf(int64(3))
+	c := v.Convert(TypeOf(byte(0)))
+
+	if c.Type().Kind() != Uint8 || c.Uint() != 3 {
+		t.Errorf("Conver(uint64 -> byte failed: kind=%v, value=%d", c.Type().Kind().String(), c.Uint())
+	}
```
aykevl: I think this could use a few more tests. Perhaps using a table, using type/value inputs and expected output?

src/reflect/value.go
```diff
@@ -1052,6 +1052,32 @@ func convertOp(src Value, typ Type) (Value, bool) {
...
+
+			switch src.Type().Elem().Kind() {
+			case Uint8:
+				return cvtBytesString(src, rtype), true
+			case Int32:
+				return cvtRunesString(src, rtype), true
+			}
+		}
+
+		// TODO(dgryski): Implement other cases
```
aykevl: Which other cases? Could be helpful to include those in the comment.

src/runtime/string.go
```diff
@@ -77,6 +77,16 @@ func stringFromBytes(x struct {
...
+		ptr *byte
+		len uintptr
+		cap uintptr
+	})(unsafe.Pointer(&x))
+	s := stringFromBytes(slice)
+	return *(*string)(unsafe.Pointer(&s))
+}
+
 // Convert a string to a []byte slice.
 func stringToBytes(x _string) (slice struct {
```
aykevl, line 91: This function should really be changed to have the signature of `stringToBytesTyped`. That's probably not possible right now, but would be once we drop support for Go 1.19 and below.
Perhaps good to indicate this in a comment (I tend to grep for Go versions like `git grep -i Go 1.19` each release to find cases like these).

src/reflect/value.go
```diff
@@ -1052,6 +1052,32 @@ func convertOp(src Value, typ Type) (Value, bool) {
 		case String:
 			return cvtUintString(src, rtype), true
 		}
+
+	case Slice:
+		if typ.Kind() == String && !src.typecode.elem().isNamed() {
```
aykevl, line 1146: Why the limitation on named types? Is it just not implemented yet?

src/reflect/value_test.go
```diff
@@ -273,6 +273,14 @@ func TestConvert(t *testing.T) {
 	if c.Type().Kind() != Uint8 || c.Uint() != 3 {
 		t.Errorf("Conver(uint64 -> byte failed: kind=%v, value=%d", c.Type().Kind().String(), c.Uint())
 	}
+
+	v = ValueOf("hello")
+	c = v.Convert(TypeOf([]byte("")))
+
+	if c.Type().Kind() != Slice || c.Type().Elem().Kind() != Uint8 && c.Len() != 5 && string(c.Bytes()) != "hello" {
+		t.Errorf("Conver(string -> []byte")
```
aykevl: Typo:

```suggestion
		t.Errorf("Convert(string -> []byte")
```

src/runtime/string.go
```diff
@@ -77,6 +77,16 @@ func stringFromBytes(x struct {
...
 
+func stringFromBytesTyped(x []byte) string {
+	slice := *(*struct {
+		ptr *byte
+		len uintptr
+		cap uintptr
+	})(unsafe.Pointer(&x))
+	s := stringFromBytes(slice)
+	return *(*string)(unsafe.Pointer(&s))
+}
```
aykevl, line 88: Have you tried `//go:linkname`ing to `runtime.stringFromBytes` directly? The signature should be the same at the LLVM IR level.
(Same for `runtime.stringToBytes`).

src/reflect/value.go
```diff
@@ -1159,6 +1216,17 @@ func makeFloat(flags valueFlags, v float64, t *rawType) Value {
...
 }
 
+func makeFloat32(flags valueFlags, v float32, t *rawType) Value {
+	size := t.Size()
+	ptr := alloc(size, nil)
+	*(*float32)(ptr) = float32(v)
+	return Value{
+		typecode: t,
+		value:    ptr,
+		flags:    flags | valueFlagIndirect,
```
aykevl: I suspect this is another instance of the wrong indirect flag (which results in a different output for `CanSet`).

src/reflect/value.go
```diff
@@ -1099,6 +1138,24 @@ func cvtUintFloat(v Value, t *rawType) Value {
...
+
+func cvtFloatUint(v Value, t *rawType) Value {
+	return makeInt(v.flags, uint64(v.Float()), t)
+}
+
+func cvtFloat(v Value, t *rawType) Value {
+	if v.Type().Kind() == Float32 && t.Kind() == Float32 {
+		// Don't do any conversion if both types have underlying type float32.
+		// This avoids converting to float64 and back, which will
+		// convert a signaling NaN to a quiet NaN. See issue 36400.
```
aykevl, line 1206: Ugh, NaN.
I'm not sure we even support signalling NaN at all. I believe most embedded C libraries just assume all NaNs are quiet (IMHO signalling NaN is just bad design).
Anyway, LGTM.

deadprogram:
@dgryski looks like just a couple of changes needed here.

src/reflect/value.go
dgryski: Looks like I was trying to implememt https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/reflect/value.go;l=3438

src/reflect/value.go
dgryski: Bah, no, I think it's more complicated than that.  We need to include the usual `size < unsafe.Sizeof(uintptr(0))` logic.

src/reflect/value_test.go
dgryski: My plan was that the majority of these issues would get sorted out when I managed to port over all the Go unit tests.

src/runtime/string.go
dgryski: I don't remember if I did.  I was probably just trying to play it safe.

src/reflect/value.go
dgryski: ... which seems to suggest that `SetUint` and friends are also wrong..

dgryski: Fixed the issues here.  Still need to look at `SetUint`

dgryski:
PTAL

dgryski:
As for tests, I plan on using the ones from https://github.com/tinygo-org/tinygo/pull/3590 when that is merged.

deadprogram:
#3590 is now merged.

dgryski:
Urk.  So turns out trying to get the Convert tests from upstream running needs a whole bunch of infra we don't have yet.  I'm going to see if I can get some sort of tests up and running here.

dgryski:
Added some tests and a fix for a bug the found.

deadprogram:
This branch now needs rebase @dgryski if you please.

dgryski:
Rebased.

dgryski:
@aykevl This is ready for review.

aykevl (approved):
LGTM

PR #3545, opened 2023-03-13
WIP:riscv64-qemu target

rminnich:
tinygo build -target riscv64-qemu -monitor .
    tinygo:ld.lld: error: /home/rminnich/.cache/tinygo/thinlto/llvmcache-933570AB9628376F0C2EAD9666A92A793D143BE9:\
    (function (reflect.Value).Complex: .text.(reflect.Value).Complex+0x76): relocation R_RISCV_HI20 out of range: \
    524291 is not in [-524288, 524287]; references reflect$string.170
    >>> referenced by value.go:0 (/home/rminnich/tinygo/src/reflect/value.go:0)
    >>> defined in /home/rminnich/.cache/tinygo/thinlto/llvmcache-933570AB9628376F0C2EAD9666A92A793D143BE9

    what's odd is that it the number is
    524291 is 0x80003
    
    I had to dump the llvm output to see the 0xffff_ffff bits.

### DIFF
--- a/targets/riscv64-qemu.json
+++ b/targets/riscv64-qemu.json
@@ -1,0 +1,8 @@
+{
+	"inherits": ["riscv64"],
+	"features": "+a,+c,+m,-relax,-save-restore",
+	"build-tags": ["virt", "qemu"],
+	"default-stack-size": 4096,
+	"linkerscript": "targets/riscv64-qemu.ld",
+	"emulator": "qemu-system-riscv32 -machine virt -nographic -bios none -kernel {}"
+}
--- a/builder/builder_test.go
+++ b/builder/builder_test.go
@@ -34,6 +34,7 @@ func TestClangAttributes(t *testing.T) {
 		"k210",
 		"nintendoswitch",
 		"riscv-qemu",
+		"riscv64-qemu",
 		"wasi",
 		"wasm",
 	}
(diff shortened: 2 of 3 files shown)


aykevl (commented):
Early (non-final) review:

* Please add `riscv64-qemu` to builder/builder_test.go and run the test.
* Please add at least a smoke test for this.
* Are you sure the compiler matches with the code? Because you're compiling with a LLVM 14 based TinyGo version and the error it gives usually indicates a version mismatch.

rminnich:
thanks, that was fast!, will add those things.

targets/riscv64-qemu.json
```diff
@@ -0,0 +1,8 @@
+{
+	"inherits": ["riscv64"],
+	"features": "+a,+c,+m,-relax,-save-restore",
+	"build-tags": ["virt", "qemu"],
+	"default-stack-size": 4096,
+	"linkerscript": "targets/riscv64-qemu.ld",
+	"emulator": "qemu-system-riscv32 -machine virt -nographic -bios none -kernel {}"
```
aykevl, line 7: `qemu-system-riscv32`? That doesn't look right.

targets/riscv64-qemu.json
```diff
@@ -0,0 +1,8 @@
+{
+	"inherits": ["riscv64"],
+	"features": "+a,+c,+m,-relax,-save-restore",
```
aykevl, line 3: These don't seem quite right, the test is complaining about it:

```
    builder_test.go:145: target has LLVM features
        	"+a,+c,+m,-relax,-save-restore"
        but Clang makes it
        	"+64bit,+a,+c,+d,+f,+m,-relax,-save-restore"
```

aykevl (commented):
What's the status of this PR?

PR #3546, opened 2023-03-13
reflect: add type check to Value.Field()

dgryski:


aykevl (approved):
Oops, yes that's a bug.

PR #3547, opened 2023-03-13
reflect: move StructField.Anonymous field to match upstream location

dgryski:


aykevl (commented):
I'm not against this change, but I'm kind of surprised it is needed? Is it just for consistency or did you hit some issue somewhere?

dgryski:
Just noticed different output running tests with TinyGo vs Upstream. Tests passed but the different ordering made it harder to see the output was the same.

aykevl:
Ok, seems harmless anyway.

PR #3548, opened 2023-03-13
reflect: fix off-by-one in Zero sizing

dgryski:
Without this, pointers wouldn't be set to nil.  Add some tests.


aykevl (approved):

aykevl:
Assuming the conflict is trivial to resolve, feel free to resolve it and merge.

dgryski:
Yeah. A bunch of these added tests in more or less the same place after extraction. I'll resolve it and merge.

PR #3549, opened 2023-03-13
reflect: set Index field in Field()

dgryski:


aykevl (commented):
Yeah, that's an oversight.

PR #3550, opened 2023-03-13
reflect: tighten unaddressable array rules for Copy()

dgryski:


aykevl (approved):
Doesn't this loosen the rules instead of tightening it?
In any case, LGTM.

dgryski:
Fixed commit message.  Merging.

PR #3551, opened 2023-03-14
reflect: MapIter.Next() needs to allocate new keys/values every time

dgryski:


aykevl (commented):
Hmm, just tested and this indeed matches the Go implementation. It seems unfortunate to me that it requires new allocations every iteration.

dgryski:
There are other map iteration calls that don't. We can add those later.

PR #3553, opened 2023-03-14
reflect: add UnsafePointer() for Func types

dgryski:


aykevl (commented):
I don't think this does what you think it does.
Func values are implemented differently in TinyGo, and in particular are different for closures and bound methods (bound methods and closures are almost the same thing to the compiler).
Is there any code that actually needs this? Can you point me to some code? Would be interesting what they even do with this pointer.

dgryski:
I'll hunt down the failing test.

dgryski:
I don't remember which package this was from, but I did find my internal failing test I had extracted:

```
package main

import "fmt"

func p() string {
	return "hello"
}

func main() {
	fmt.Println(p)
}
```

This could also easily come from trying to print a struct that has a function pointer (say, a list of callbacks) as a member.

aykevl:
The main difference that I can think of is that TinyGo will return the same pointer value for each instance of a closure, while upstream Go will return a different one.
Maybe there is code that relies on this difference? But most likely it's only for printing, in which case it's just a very small change in behavior.

I guess either way LGTM, and if it breaks something we'll know it and can complain to that package (which relies on implementation specific behavior). It would in any case be unsafe to cast the pointer back to a function, I hope nobody is seriously doing that.

If you want behavior more close to upstream, this would work:

```
if .Context !=nil:
    return .Context
return .Code
```

(I didn't test any of this, I could easily be wrong).

Anyway, just letting you know in case this is relevant or you want to check it, feel free to merge if you think the slight behavioral difference is fine.

dgryski:
I'll update my code to check Context first.

dgryski:
PTAL

PR #3557, opened 2023-03-14
reflect: SetMapIndex: use AssignableTo() instead of type equality

dgryski:


aykevl (commented):
I think I am missing some context here. Is this related to your Convert work? Otherwise I don't really see a reason why this change is necessary.
(And also: this could lead to a problem later on when AssignableTo is extended to support interfaces - at that point, you might for example set an integer value to a map[string]interface{} and that would misbehave in SetMapIndex. So I think the old check should probably remain in place as a TODO for safety).

dgryski:
Yes. This came up in the context of encoding/json trying to assign a value of   `map[string]interface{}` in a higher level `map[string]interface{}`. The assignment is valid, but type pointer equality doesn't work.

dgryski:
Here is the code I was using to test this. (Code demonstrating unpacking into a `struct` removed):
```
package main

import (
        "encoding/json"
        "fmt"
)

const httpbinJSON = `
{
  "slideshow": {
    "author": "Yours Truly",
    "date": "date of publication",
    "slides": [
      {
        "title": "Wake up to WonderWidgets!",
        "type": "all"
      },
      {
        "items": [
          "Why <em>WonderWidgets</em> are great",
          "Who <em>buys</em> WonderWidgets"
        ],
        "title": "Overview",
        "type": "all"
      }
    ],
    "title": "Sample Slide Show"
  }
}
`
func testJSONInterface() {
        var m interface{}
        err := json.Unmarshal([]byte(httpbinJSON), &m)
        if err != nil {
                fmt.Println("error decoding JSON:", err)
        }
        j, err := json.MarshalIndent(m, "", " ")
        if err != nil {
                fmt.Println("error encoding to JSON:", err)
        }
        fmt.Printf("interface:\n%v\n", string(j))
}

func main() {
        testJSONInterface()
}
```

dgryski:
Ah, the issue you're concerned about is addressed by wrapping things in an interface if required, as I did in https://github.com/tinygo-org/tinygo/pull/3558 .

Turns out these two PRs are not as independent as I thought.  I'll close https://github.com/tinygo-org/tinygo/pull/3558 and incorporate the changes into this one.  That should fix the failing test in https://github.com/tinygo-org/tinygo/pull/3558 too.

dgryski:
Forgot this depends on checking `NumMethods`; need to be merged after https://github.com/tinygo-org/tinygo/pull/3529

dgryski:
Rebased now that #3529 is in.

dgryski:

PR #3559, opened 2023-03-14
reflect: call decomposeInterface() directly in TypeOf()

dgryski:


aykevl (commented):
Looks good to me, but what was the motivation for this change? Efficiency?

dgryski:
Was mostly confused why we called ValueOf and then immediately threw away everything but the type.

aykevl (approved):
Probably just laziness. The old code is also a bit more readable.
Anyway, LGTM.

PR #3562, opened 2023-03-15
reflect: add SetBytes()

dgryski:


aykevl:
@dgryski looks like there is a typo in the test

dgryski:
Saw that. Will fix once my kids are in bed.

dgryski:
PTAL

src/reflect/value.go
```diff
@@ -980,7 +980,13 @@ func (v Value) SetString(x string) {
...
 
 func (v Value) SetBytes(x []byte) {
-	panic("unimplemented: (reflect.Value).SetBytes()")
+	v.checkAddressable()
+	if v.typecode.Kind() != Slice || v.typecode.elem().Kind() != Uint8 {
+		panic("reflect.Value.SetBytes called on not []byte")
+	}
+
+	// copy the header contents over
+	*(*sliceHeader)(v.value) = *(*sliceHeader)(unsafe.Pointer(&x))
```
aykevl: I think this works just as well, and seems easier to read:

```suggestion
	*(*[]byte)(v.value) = x
```

aykevl (approved):
Looks good to me!
I was initially not expecting `SetBytes` actually copies the slice bytes (and not the slice itself), but it also happens upstream so :shrug:

src/reflect/value.go
dgryski: Oh yeah, of course.

aykevl (approved):

PR #3564, opened 2023-03-15
reflect: fix Type.Name to return empty string for non-named types

dgryski:
// Name returns the type's name within its package for a defined type. 
    // For other (non-defined) types it returns the empty string.


dgryski:
This is not correct.

dgryski:
Fixed.

aykevl (approved):

PR #3566, opened 2023-03-16
Update transform tests with opaque pointers

aykevl:
Just a maintenance update. Support for non-opaque pointers will be dropped by LLVM at some point, and opaque pointers make it easier to modify these tests.


aykevl:
This now includes all IR tests in transform/testdata.

The test inputs are updated by hand (with the help of search+replace), the test outputs are all generated but I checked that they didn't change in behavior (they just got smaller due to the lack of typed pointers).

dgryski (approved):
LGTM.  Passes tests.

PR #3568, opened 2023-03-16
compiler: support all kinds of recursive types

aykevl:
Previously we only supported recursive types in structs. But there can be other kinds of recursive types, like slices:

    type RecursiveSlice []RecursiveSlice

This doesn't involve structs, so it led to infinite recursion in the compiler. This fix avoids recursion at the proper level: at the place where the named type is defined.


dgryski (approved):
LGTM.

deadprogram:

PR #3569, opened 2023-03-17
reflect: unpack interfaces in MapKeys() if needed

dgryski:


src/reflect/value.go
```diff
@@ -779,8 +779,17 @@ func (v Value) MapKeys() []Value {
...
 
+	keyType := v.typecode.Key().(*rawType)
+	isKeyStoredAsInterface := keyType.Kind() != String && !keyType.isBinary()
+
 	for hashmapNext(v.pointer(), it, k.value, e.value) {
-		keys = append(keys, k.Elem())
+		if isKeyStoredAsInterface {
+			intf := *(*interface{})(k.value)
+			v := ValueOf(intf)
+			keys = append(keys, v)
```
aykevl, line 789: What about maps that actually do have interface keys, like `map[interface{}]int`? Do they also work with this change?
Might be useful to add a test case for that as well.

dgryski: Added another test for maps with interface keys.

dgryski:
PTAL

dgryski:
Going to take out https://github.com/tinygo-org/tinygo/pull/3569/commits/fb8e6f49589734703f9631768cb7b2f711be64ba because it needs https://github.com/tinygo-org/tinygo/pull/3557 merged to be able to check the map key assignment.

dgryski:

PR #3573, opened 2023-03-18
reflect map key interface fixes

dgryski:


src/reflect/value.go
```diff
@@ -846,9 +846,13 @@ func (v Value) MapRange() *MapIter {
 		panic(&ValueError{Method: "MapRange", Kind: v.Kind()})
 	}
 
+	keyType := v.typecode.Key().(*rawType)
```
aykevl:
```suggestion
	keyType := v.typecode.key()
```

src/reflect/value.go
```diff
@@ -846,9 +846,13 @@ func (v Value) MapRange() *MapIter {
...
+	isKeyStoredAsInterface := keyType.Kind() != String && !keyType.isBinary()
```
aykevl, line 850: Perhaps an idea to refactor this into a function that returns the key type as an enum (string, binary, interface)? Or change `isBinary` to be that function (and change the name with it, of course). Can happen in a separate PR.

dgryski:
This started as a follow-up to https://github.com/tinygo-org/tinygo/pull/3569 to make sure `MapIter` also unpacks the keys like `MapKeys()` needed to do.

src/reflect/value.go
```diff
@@ -858,14 +862,21 @@ type MapIter struct {
...
 
 func (it *MapIter) Key() Value {
 	if !it.valid {
 		panic("reflect.MapIter.Key called on invalid iterator")
 	}
 
+	if it.keyInterface {
+		intf := *(*interface{})(it.key.value)
+		v := ValueOf(intf)
+		return v
```
aykevl, line 877: Does this also work with maps that really do have interfaces as map keys? Like `map[interface{}]int`.
~I think there should be tests to show the fixed behavior.~ I see now there are tests in the last commit, but not for maps that really do have interfaces as keys.

compiler/map.go
```diff
@@ -78,6 +78,7 @@ func (b *builder) createMapLookup(keyType, valueType types.Type, m, key llvm.Val
 
 	// Do the lookup. How it is done depends on the key type.
 	var commaOkValue llvm.Value
+	origKeyType := keyType
 	keyType = keyType.Underlying()
```
aykevl, line 82: Oh, yes, I see now that this code was buggy.
:+1:

src/reflect/value_test.go
```diff
@@ -132,6 +132,49 @@ func TestMap(t *testing.T) {
...
+	refut.SetMapIndex(ValueOf(unmarshalerText{"y", "z"}), ValueOf(true))
+
+	// make sure we can get it out with the compiler
+	if !ut[unmarshalerText{"y", "z"}] {
+		t.Errorf("Failed to look up up reflect-set map key with compiler")
+	}
+
+	utKeys := refut.MapKeys()
+
+	// make sure keys extracted via reflection have the correct ype
```
aykevl: nit: typo

```suggestion
	// make sure keys extracted via reflection have the correct type
```

dgryski:
I have a patch to fix `map[interface{}]int` (and tests) but they depend on this and https://github.com/tinygo-org/tinygo/pull/3557 being merged.

Basically, the `map` logic needs to not use the type that was passed in to determine the hashing type, but the key type of the ap instead.

dgryski:

PR #3574, opened 2023-03-19
reflect: fix isBinary() for float types

dgryski:


dgryski:
`isBinary()` needs to match the implementation in the compiler: https://github.com/tinygo-org/tinygo/blob/release/compiler/map.go#L244

aykevl (approved):
Ah yes, I must have missed this in review.

Floating point NaN requires special handing and while we could introduce a special map type for this, I didn't bother with the compiler version of this and just let them be hashed using the interface map type.

PR #3576, opened 2023-03-19
docker: correct path for GHCR dev container build

deadprogram:
This PR corrects the path for the Docker dev container build pushed to GHCR. 

It is currently `ghcr.io/tinygo-org/tinygo/tinygo-dev` and it should be just `ghcr.io/tinygo-org/tinygo-dev`


dgryski (approved):
LGTM

PR #3578, opened 2023-03-19
reflect: tweak Type.String() for interfaces to make encoding/xml happy

dgryski:


aykevl (approved):
This is fine, but I would have preferred if this came with some tests.
Will this be part of the reflect tests you want to port?

dgryski:
In particular `encoding/xml` expects an error message with the name of an interface, specifically `interface{}`.  Because we didn't have any special handling for interfaces, it returned just `interface` causing the test to fail.  That's all this output tweak is fixing.

One we have actual method sets for interfaces we can fix this to properly print them.

PR #3582, opened 2023-03-20
src/examples/pjw: display Peter J. Weinberger

rminnich:
If you search for "bell labs pjw image", you will find an image of pjw from the earliest days of digital photography at Bell Labs. https://spinroot.com/pico/pjw.html is particularly fun.

The image is very popular in Plan 9, including "...  in the Plan 9 software, for instance when you cat a file, all characters missing from the font (as well as NUL and U+FFFD characters) render as tiny pjw faces."

Given that many ideas in Go come from the Bell Labs researchers, it seems only fair to be able to show pjw on devices that work with tinygo.

This command shows several things:
o use of go generate to create img.go
o use of embed
o in the generator, pjw.go, piping the generated code through
  goimports, which has the additional effect of
  formatting it
o a very simple program to display the image

The result on, e.g., the pyportal, is monochrome pjw!


aykevl:
So this sounds like an inside joke at Bell Labs?
Is this also common in the Go community?

I certainly don't want to spoil a joke, but I'm not sure this really belongs in the TinyGo repository: it adds a ton of dependencies and generally I'd like to avoid any dependencies on the drivers repo from inside the tinygo repo.
This image could be used as an example how to draw an image on a screen. I think that would actually be quite helpful.

deadprogram:
I am planning to restart https://github.com/tinygo-org/tinygo-zoo perhaps that would be a good place for this?

deadprogram:
Closing in favor of adding to zoo, or someplace else. Thank everyone!

Issue #3544, opened 2023-03-13
Want to easily modify USB_PID, etc.


sago35:
I would like to modify the following four values at tinygo build time.

* usb_STRING_PRODUCT
* usb_STRING_MANUFACTURER
* usb_VID
* usb_PID

https://github.com/tinygo-org/tinygo/blob/v0.27.0/src/machine/board_xiao-rp2040.go#L99-L108

I tried changing usbn_STRING_PRODUCT from const to var and rewriting it using ldflags, but it did not work.

```
$ tinygo flash --target xiao-rp2040 --ldflags "-X machine.usbn_STRING_PRODUCT='XXX'" examples/blinky1
```

ldflags can rewrite the strings contained in the main package.
However, for some reason, I think it cannot rewrite the machine package.

One way to resolve this Issue is to

1. Make ldflags work with any package that includes the machine package
2. Change machine.usb_STRING_PRODUCT etc. from const to var
3. Prepare string variables to override usb_VID and usb_PID

sago35:
I am making my keyboard in TinyGo.
I use xiao-rp2040 for my keyboard.

In this situation, if you want to use the xiao-rp2040 separately from the keyboard, the tinygo flash command becomes a bit cumbersome.
For example, if -port is specified incorrectly, the keyboard will not work, which is very annoying.


https://github.com/sago35/tinygo-keyboard

deadprogram:
This is part of the `v0.28` release so now closing. Thank you everyone!

PR #3583, opened 2023-03-20
machine/usb: provide ability to override USB VID, PID, Manufacturer name and Product name

deadprogram:
This PR adds a way to override the default USB VID, PID, Manufacturer name and Product name, for those boards that we have full USB support.

It should solve #3544 for your needs @sago35 @bgould and of course anyone else please take a look.

### DIFF
--- a/src/examples/hid-keyboard/main.go
+++ b/src/examples/hid-keyboard/main.go
@@ -1,10 +1,21 @@
+// to override the USB Manufacturer or Product names:
+//
+// tinygo flash -target circuitplay-express -ldflags="-X main.usbManufacturer='TinyGopher Labs' -X main.usbProduct='GopherKeyboard'" examples/hid-keyboard
+//
+// you can also override the VID/PID. however, only set this if you know what you are doing,
+// since changing it can make it difficult to reflash some devices.
 package main
 
 import (
 	"machine"
+	"machine/usb"
 	"machine/usb/hid/keyboard"
+	"strconv"
 	"time"
 )
+
+var usbVID, usbPID string
+var usbManufacturer, usbProduct string
 
 func main() {
 	button := machine.BUTTON
@@ -17,5 +28,25 @@ func main() {
 			kb.Write([]byte("tinygo"))
 			time.Sleep(200 * time.Millisecond)
 		}
+	}
+}
+
+func init() {
+	if usbVID != "" {
+		vid, _ := strconv.ParseUint(usbVID, 0, 16)
+		usb.VendorID = uint16(vid)
+	}
+
+	if usbPID != "" {
+		pid, _ := strconv.ParseUint(usbPID, 0, 16)
+		usb.ProductID = uint16(pid)
+	}
+
+	if usbManufacturer != "" {
+		usb.Manufacturer = usbManufacturer
+	}
+
+	if usbProduct != "" {
+		usb.Product = usbProduct
 	}
 }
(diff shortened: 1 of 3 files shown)


deadprogram:
```
$ tinygo flash -target circuitplay-express -ldflags="-X main.usbManufacturer='TinyGopher Labs' -X main.usbProduct='GopherKeyboard'" examples/hid-keyboard

$ lsusb
Bus 004 Device 001: ID 1d6b:0003 Linux Foundation 3.0 root hub
Bus 003 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub
Bus 002 Device 001: ID 1d6b:0003 Linux Foundation 3.0 root hub
Bus 001 Device 003: ID 8087:0029 Intel Corp. 
Bus 001 Device 015: ID 239a:8018 TinyGopher Labs GopherKeyboard
Bus 001 Device 002: ID 0c45:6723 Microdia Integrated_Webcam_HD
Bus 001 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub
```

sago35:
@deadprogram 
Thank you very much.
It will be a good example.

Overall, I think it is good.
However, I have one minor concern that I will try later.

Perhaps rewriting the VID/PID using this feature will make the next `tinygo flash` more difficult.
In many cases, like the keyboards I make, the PCB is not bare.
In other words, it is difficult to double-tap the reset button or press bootsel.

It may be possible to implement this in another PR, but I believe changes are needed in the tinygo core to better handle VID/PID.

deadprogram:
> Perhaps rewriting the VID/PID using this feature will make the next tinygo flash more difficult.

I already ran into that. :crying_cat_face: 

Not sure how to handle that exactly, since it is a feature you need to have for a professional-level HID device aka real product.

deadprogram:
I added some GoDoc comments along with hopefully suitably scary warning, and force pushed. Please take another look.

deadprogram:
One last change, I removed VID and PID from the comment in the example, so people do ont get in trouble just by copy/paste that command.

src/examples/hid-keyboard/main.go
```diff
@@ -1,11 +1,19 @@
+// to override the USB Manufacturer or Product names:
+//
+// tinygo flash -target circuitplay-express -ldflags="-X main.usbManufacturer='TinyGopher Labs' -X main.usbProduct='GopherKeyboard'" examples/hid-keyboard
```
sago35, line 3: It would be helpful if it said that the VID/PID can be changed in the same way, but with caution.

src/examples/hid-keyboard/main.go
```diff
@@ -19,3 +27,23 @@ func main() {
...
+func init() {
+	if usbVID != "" {
+		vid, _ := strconv.Atoi(usbVID)
+		usb.VendorID = uint16(vid)
+	}
+
+	if usbPID != "" {
+		pid, _ := strconv.Atoi(usbPID)
+		usb.ProductID = uint16(pid)
+	}
```
sago35: Since hexadecimal numbers are used more often than decimal numbers, strconv.ParseUINT(x, 16, 0) may be a better choice.

sago35:
I am very happy that it is working well.
image

src/examples/hid-keyboard/main.go
deadprogram: Updated.

src/examples/hid-keyboard/main.go
deadprogram: good point. done.

sago35 (approved):
LGTM

deadprogram:

PR #3589, opened 2023-03-22
examples/wasm: update export example to show the use of the go:wasmimport directive

deadprogram:
This PR updates the wasm "export" example to show the use of the `go:wasmimport` directive. See #3165 for details.

### DIFF
--- a/src/examples/wasm/export/wasm.go
+++ b/src/examples/wasm/export/wasm.go
@@ -8,12 +8,12 @@ import (
 func main() {
 }
 
-//export add
+//go:wasmimport add
 func add(a, b int) int {
 	return a + b
 }
 
-//export update
+//go:wasmimport update
 func update() {
 	document := js.Global().Get("document")
 	aStr := document.Call("getElementById", "a").Get("value").String()
--- a/src/examples/wasm/README.md
+++ b/src/examples/wasm/README.md
@@ -2,10 +2,10 @@
 
 The examples here show two different ways of using WebAssembly with TinyGo:
 
-1. Defining and exporting functions via the `//export <name>` directive. See
-[the export folder](./export) for an example of this.  Additionally, the Wasm
-module (which has a default value of `env`) can be specified using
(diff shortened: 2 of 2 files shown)


aykevl (commented):
`//go:wasmimport` is _meant_ for only importing functions, but it also unintentionally exports functions.

Do we want to officially support that, or add a new `//go:wasmexport` instead?

deadprogram:
The proposal at https://github.com/golang/go/issues/38248 was accepted and already implemented here https://github.com/golang/go/commit/02411bcd7c8eda9c694a5755aff0a516d4983952 so in the interest of keeping compatibility we should probably do the same thing for now.

src/examples/wasm/export/wasm.go
```diff
@@ -8,12 +8,12 @@ import (
 func main() {
 }
 
-//export add
+//go:wasmimport add
```
achille-roussel, line 11: I think this update is premature, go:wasmimport is only supported on functions with no body that get linked at runtime to exports of other modules. 

Here it seems that the //export directive is intended to ask the compiler to add the function to the list of exports of the module, or am I missing something?

aykevl:
I have made a PR to fix this: https://github.com/tinygo-org/tinygo/pull/3610
It should make this PR (intentionally) fail.

That said, I do agree it would be a good idea to add an example with `//go:wasmimport`, but make it an actual import instead of an export.

src/examples/wasm/export/wasm.go
deadprogram: Ah OK. I totally misunderstood what was happening here. Thanks for the clarification.

deadprogram:
> It should make this PR (intentionally) fail.

I think the best thing to come out of this PR is to prevent me from doing what is in this PR. :smile_cat: 

Thanks @achille-roussel and @aykevl 

I will close this one since it does not actually do anything we need.

Issue #3449, opened 2023-02-15
`defer`s shouldn't re-run when repanic'ing.


dgryski:
Looks like the `panic` in the `defer` block causes the `defer` to rerun, when it shouldn't.

```
~/go/src/github.com/dgryski/bug/repanic $ cat main.go
package main

func main() {
	var rerun int
	defer func() {
		println("defer")
		if r := recover(); r != nil {
			rerun++
			if rerun < 5 {
				panic(r)
			}
		}
	}()

	panic("boom")
}
~/go/src/github.com/dgryski/bug/repanic $ go run main.go
defer
panic: boom [recovered]
	panic: boom

goroutine 1 [running]:
main.main.func1()
	/Users/dgryski/go/src/github.com/dgryski/bug/repanic/main.go:10 +0x6d
panic({0x105ba80, 0x107a8b8})
	/Users/dgryski/go/src/go.googlesource.com/go/src/runtime/panic.go:884 +0x213
main.main()
	/Users/dgryski/go/src/github.com/dgryski/bug/repanic/main.go:15 +0x6c
exit status 2
~/go/src/github.com/dgryski/bug/repanic $ tinygo run main.go
defer
defer
defer
defer
defer
```

PR #3590, opened 2023-03-22
reflect: import / enable all upstream reflect tests

dgryski:


dgryski:
More of these tests will pass when a few of the outstanding PRs are merged.  Some of the table tests *mostly* pass except for one or two small things (which we could either fix or comment out).  All the uses of `shouldPanic()` are returned without  testing becaue of https://github.com/tinygo-org/tinygo/issues/3449 for native and because wasi doesn't support panic.

src/reflect/tostring_test.go
```diff
@@ -0,0 +1,95 @@
...
+// For debugging only.
+func valueToString(val Value) string {
+	var str string
+	if !val.IsValid() {
+		return "<zero Value>"
+	}
+	typ := val.Type()
+	switch val.Kind() {
+	case Int, Int8, Int16, Int32, Int64:
+		return strconv.FormatInt(val.Int(), 10)
```
soypat, line 26: I'm partial to using the switch differently in these `toString` functions where each `case` ends with an assignment rather than a return. Make for more "structured" code, or at least that's how Kevlin Hennesy calls it in his talk "The Lost Art of Structured Programming".

soypat (approved):
LGTM. Left a nitpicky suggestion which is OK either way.

src/reflect/tostring_test.go
dgryski: This is imported from upstream.

deadprogram:
Thanks @dgryski and @soypat now merging.

PR #3598, opened 2023-03-25
reflect: prevent AccidentallyQuadratic behaviour when decoding json s…

dgryski:
…tructs

### DIFF
--- a/src/reflect/type.go
+++ b/src/reflect/type.go
@@ -658,7 +658,13 @@ const fieldOffsetCacheSize = 4 * unsafe.Sizeof(uintptr(0))
 type fieldOffsetCacheType [fieldOffsetCacheSize]struct {
 	t       *structType
 	offsets map[int]uintptr
-	locked  bool
+
+	// While we don't have threads contending for cache entries, we do have
+	// recursive calls. If a struct has other structs as fields, then
+	// calculating the offsets here will generate calls to Align() and
+	// Size() both of which call rawField which will end up back here in
+	// lookup.  If need to protect cache entries from colliding structs.
+	locked bool
 }
 
 var fieldOffsetCache fieldOffsetCacheType
--- a/src/reflect/type.go
+++ b/src/reflect/type.go
@@ -803,7 +803,6 @@ func (t *rawType) rawFieldByName(n string) (rawStructField, []int, bool) {
 			// Also calculate field offset.
 
 			descriptor := (*structType)(unsafe.Pointer(ll.t.underlying()))
-			var offset uintptr
 			field := &descriptor.fields[0]
 
 			for i := uint16(0); i < descriptor.numField; i++ {
@@ -817,6 +816,7 @@ func (t *rawType) rawFieldByName(n string) (rawStructField, []int, bool) {
 				name := readStringZ(data)
 				data = unsafe.Add(data, len(name))
 				if name == n {
+					offset := fieldOffsetCache.lookup(descriptor, int(i))
 					found = append(found, result{
 						rawStructFieldFromPointer(descriptor, field.fieldType, data, flagsByte, name, offset),
 						append(ll.index, int(i)),
@@ -836,17 +836,8 @@ func (t *rawType) rawFieldByName(n string) (rawStructField, []int, bool) {
 					})
 				}
 
-				offset += field.fieldType.Size()
-
-				// update offset/field pointer if there *is* a next field
-				if i < descriptor.numField-1 {
-
-					// Increment pointer to the next field.
-					field = (*structField)(unsafe.Add(unsafe.Pointer(field), unsafe.Sizeof(structField{})))
-
-					// Align the offset for the next field.
-					offset = align(offset, uintptr(field.fieldType.Align()))
-				}
+				// Increment pointer to the next field.
+				field = (*structField)(unsafe.Add(unsafe.Pointer(field), unsafe.Sizeof(structField{})))
 			}
 		}
 
--- a/src/reflect/type.go
+++ b/src/reflect/type.go
@@ -653,7 +653,7 @@ func rawStructFieldFromPointer(descriptor *structType, fieldType *rawType, data 
 	}
 }
 
-const fieldOffsetCacheSize = 16
+const fieldOffsetCacheSize = 4 * unsafe.Sizeof(uintptr(0))
 
 type fieldOffsetCacheType [fieldOffsetCacheSize]struct {
 	t       *structType
--- a/src/reflect/type.go
+++ b/src/reflect/type.go
@@ -890,6 +890,8 @@ func (t *rawType) NumField() int {
 	return int((*structType)(unsafe.Pointer(t.underlying())).numField)
 }
 
+var structSizeCache = make(map[*rawType]uintptr)
+
 // Size returns the size in bytes of a given type. It is similar to
 // unsafe.Sizeof.
 func (t *rawType) Size() uintptr {
@@ -928,12 +930,18 @@ func (t *rawType) Size() uintptr {
 	case Array:
 		return t.elem().Size() * uintptr(t.Len())
 	case Struct:
+		if size, ok := structSizeCache[t]; ok {
+			return size
+		}
+
 		numField := t.NumField()
 		if numField == 0 {
 			return 0
 		}
 		lastField := t.rawField(numField - 1)
-		return align(lastField.Offset+lastField.Type.Size(), uintptr(t.Align()))
+		size := align(lastField.Offset+lastField.Type.Size(), uintptr(t.Align()))
+		structSizeCache[t] = size
+		return size
 	default:
 		panic("unimplemented: size of type")
 	}
@@ -1068,6 +1076,8 @@ func (t *rawType) Comparable() bool {
 	}
 }
 
+var structIsBinaryCache = make(map[*rawType]bool)
+
 // isbinary() returns if the hashmapAlgorithmBinary functions can be used on this type
 func (t *rawType) isBinary() bool {
 	switch t.Kind() {
@@ -1078,12 +1088,17 @@ func (t *rawType) isBinary() bool {
 	case Array:
 		return t.elem().isBinary()
 	case Struct:
+		if b, ok := structIsBinaryCache[t]; ok {
+			return b
+		}
 		numField := t.NumField()
 		for i := 0; i < numField; i++ {
 			if !t.rawField(i).Type.isBinary() {
+				structIsBinaryCache[t] = false
 				return false
 			}
 		}
+		structIsBinaryCache[t] = true
 		return true
 	}
 	return false
--- a/src/reflect/type.go
+++ b/src/reflect/type.go
@@ -658,6 +658,7 @@ const fieldOffsetCacheSize = 16
 type fieldOffsetCacheType [fieldOffsetCacheSize]struct {
 	t       *structType
 	offsets map[int]uintptr
+	locked  bool
 }
 
 var fieldOffsetCache fieldOffsetCacheType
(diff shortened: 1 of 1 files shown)


dgryski:
I have a better hash function that drastically reduces the number of collisions on my test data.

src/reflect/type.go
```diff
@@ -631,6 +632,66 @@ func rawStructFieldFromPointer(descriptor *structType, fieldType *rawType, data
...
+	idx ^= (uintptr(unsafe.Pointer(t.ptrTo)) << (unsafe.Sizeof(uintptr(0)) * (8 / 2)))
+	idx = hashptr(idx)
+	idx ^= idx >> (unsafe.Sizeof(uintptr(0)) * (8 / 2))
+	idx ^= uintptr(bits.RotateLeft32(uint32(idx), int(t.numField)))
+	idx %= fieldOffsetCacheSize
+	if f[idx].t != t {
+		// Cache miss; populate cache for this type.
+		// Clear out the old map so we have space to allocate a new one
+		f[idx].offsets = nil
+		offsets := make(map[int]uintptr)
```
dgryski: Could probably preallocate this map so we don't hit the growth case.

src/reflect/type.go
```diff
@@ -631,6 +632,66 @@ func rawStructFieldFromPointer(descriptor *structType, fieldType *rawType, data
...
+
+type fieldOffsetCacheType [fieldOffsetCacheSize]struct {
+	t       *structType
+	offsets map[int]uintptr
+}
+
+var fieldOffsetCache fieldOffsetCacheType
+
+func (f *fieldOffsetCacheType) lookup(t *structType, n int) uintptr {
+	idx := uintptr(unsafe.Pointer(t))
```
dgryski: Should probably extract this hash logic to its own function.

src/reflect/type.go
```diff
@@ -631,6 +632,66 @@ func rawStructFieldFromPointer(descriptor *structType, fieldType *rawType, data
...
+}
+
+//go:linkname xorshift32 runtime.xorshift32
+func xorshift32(x uint32) uint32
+
+//go:linkname xorshiftMult64 runtime.xorshiftMult64
+func xorshiftMult64(x uint64) uint64
+
+func hashptr(p uintptr) uintptr {
+	if unsafe.Sizeof(p) == 32 {
```
dgryski: Oops, this should be `4`.

dgryski:
PTAL.  Fixed all the issues I identified.

dgryski:
Huh. CI failed.

dgryski:
Two hard problems in computer science. 

When filling a cache, we need to call field.Size(), but for some structs that field was itself another struct which collided with the struct we were in the middle of trying to process, which would overwrite the cache entry and make things Bad.  Ouch.

deadprogram:
Seems like real CI error https://github.com/tinygo-org/tinygo/actions/runs/4530584804/jobs/7980277801?pr=3598#step:7:39

dgryski:
Yup and I know why it happened. This is a collision when the cache is being re populated but it's for the type that was in that slot. Since clearing the offsets map is the first thing we do, the other collision reads from a null map which is zero. Need to pull the locked check out of the big if completely.

dgryski:
Failing windows stdlib test is just the `net` package this time.

dgryski:
All green. PTAL.

deadprogram:
I re-ran a couples times CI is green.

src/reflect/type.go
```diff
@@ -631,6 +632,93 @@ func rawStructFieldFromPointer(descriptor *structType, fieldType *rawType, data
 	}
 }
 
+const fieldOffsetCacheSize = 16
```
dgryski: We could probably scale this like
```
const fieldOffsetCacheSze = 4 * unsafe.Sizeof(uintptr(0))
```
so when building natively or on larger machines we get a larger default cache.  I was just trying to be really conservative here.

src/reflect/type.go
```diff
@@ -850,6 +918,8 @@ func (t *rawType) Size() uintptr {
 	}
 }
 
+var structAlignCache = make(map[*rawType]int)
```
dgryski, line 956: We could probably added additional caches for `Size()`, `Comparable()`, and `isBinary()`. (assuming we didn't want to add Comparable and isBinary to a flag byte somewhere already)

src/reflect/type.go
```diff
@@ -631,6 +632,93 @@ func rawStructFieldFromPointer(descriptor *structType, fieldType *rawType, data
...
+	locked  bool
+}
+
+var fieldOffsetCache fieldOffsetCacheType
+
+func (f *fieldOffsetCacheType) lookup(t *structType, n int) uintptr {
+	idx := hashStructTypePtr(t)
+	idx %= fieldOffsetCacheSize
+
+	if f[idx].locked {
```
kenbell, line 676: I'm not sure about this locking primitive.  I don't think there's anything that guarantees a write to a bool is atomic like this.  There's no memory barrier or anything...

aykevl:
@dgryski have you run some size comparison tests? For example, `make smoke-test` in the drivers repo.

(I haven't really looked at the code yet).

dgryski:
The extra code complexity here is really to avoid having the cache grow without bound.  If we don't care about that we can probably just replace it with a straight map and move on.  That shouldn't impact the code size as much but will probably affect runtime memory when reflection is used because there are no longer any cache evictions.

dgryski:
Non-zero entries from `make smoke-test` and `sizediff`
```
153528 153968    440   0.3%  tinygo build -size short -o ./build/test.hex -target=feather-m4 ./examples/sdcard/tinyfs/
 95572  96020    448   0.5%  tinygo build -size short -o ./build/test.hex -target=wioterminal ./examples/rtl8720dn/mqttsub/
132540 133020    480   0.4%  tinygo build -size short -o ./build/test.hex -target=arduino-nano33 ./examples/wifinina/ntpclient/main.go
132536 133016    480   0.4%  tinygo build -size short -o ./build/test.hex -target=arduino-nano33 ./examples/wifinina/udpstation/main.go
133056 133536    480   0.4%  tinygo build -size short -o ./build/test.hex -target=arduino-nano33 ./examples/wifinina/webclient/main.go
132748 133236    488   0.4%  tinygo build -size short -o ./build/test.hex -target=arduino-nano33 ./examples/wifinina/tcpclient/main.go
 69328  69824    496   0.7%  tinygo build -size short -o ./build/test.hex -target=wioterminal ./examples/rtl8720dn/webserver/
 79964  80468    504   0.6%  tinygo build -size short -o ./build/test.hex -target=wioterminal ./examples/rtl8720dn/webclient/
 20004  20548    544   2.7%  tinygo build -size short -o ./build/test.hex -target=microbit-v2 ./examples/microbitmatrix/main.go
 71140  71708    568   0.8%  tinygo build -size short -o ./build/test.hex -target=p1am-100 ./examples/p1am/main.go
 23984  24552    568   2.4%  tinygo build -size short -o ./build/test.uf2 -target=pico ./examples/ssd1289/main.go
 19956  20532    576   2.9%  tinygo build -size short -o ./build/test.hex -target=microbit ./examples/microbitmatrix/main.go
 66012  67060   1048   1.6%  tinygo build -size short -o ./build/test.hex -target=itsybitsy-m0 ./examples/flash/console/spi
 60836  61908   1072   1.8%  tinygo build -size short -o ./build/test.hex -target=itsybitsy-m0 ./examples/bmp280/main.go
 66452  67524   1072   1.6%  tinygo build -size short -o ./build/test.hex -target=itsybitsy-m0 ./examples/ds3231/main.go
 62316  63388   1072   1.7%  tinygo build -size short -o ./build/test.hex -target=pyportal ./examples/flash/console/qspi
 74812  75884   1072   1.4%  tinygo build -size short -o ./build/test.hex -target=feather-m0 ./examples/dht/main.go
 69424  70496   1072   1.5%  tinygo build -size short -o ./build/test.hex -target=feather-m4 ./examples/sdcard/console/
 58428  59508   1080   1.8%  tinygo build -size short -o ./build/test.hex -target=itsybitsy-m0 ./examples/adt7410/main.go
 60796  61876   1080   1.8%  tinygo build -size short -o ./build/test.hex -target=itsybitsy-m0 ./examples/bmi160/main.go
 64452  65532   1080   1.7%  tinygo build -size short -o ./build/test.hex -target=itsybitsy-m0 ./examples/mcp2515/main.go
 53952  55032   1080   2.0%  tinygo build -size short -o ./build/test.hex -target=microbit ./examples/sht3x/main.go
 53932  55012   1080   2.0%  tinygo build -size short -o ./build/test.hex -target=microbit ./examples/shtc3/main.go
 68316  69396   1080   1.6%  tinygo build -size short -o ./build/test.hex -target=xiao ./examples/pcf8563/alarm/
 67976  69056   1080   1.6%  tinygo build -size short -o ./build/test.hex -target=xiao ./examples/pcf8563/time/
 68252  69332   1080   1.6%  tinygo build -size short -o ./build/test.hex -target=xiao ./examples/pcf8563/timer/
 57660  58828   1168   2.0%  tinygo build -size short -o ./build/test.hex -target=feather-m4 ./examples/i2csoft/adt7410/
 78504  79672   1168   1.5%  tinygo build -size short -o ./build/test.hex -target=nucleo-wl55jc ./examples/lora/lorawan/atcmd/
263736 264912   1176   0.4%  tinygo build -size short -o ./build/test.hex -target=pyportal ./examples/ili9341/slideshow
 57892  59076   1184   2.0%  tinygo build -size short -o ./build/test.hex -target=feather-nrf52840 ./examples/is31fl3731/main.go
 64692  66132   1440   2.2%  tinygo build -size short -o ./build/test.hex -target=feather-m0 ./examples/gps/i2c/main.go
 22232  23676   1444   6.5%  tinygo build -size short -o ./build/test.hex -target=arduino-nano33 ./examples/espat/espconsole/main.go
 22432  23876   1444   6.4%  tinygo build -size short -o ./build/test.hex -target=arduino-nano33 ./examples/espat/esphub/main.go
 22240  23684   1444   6.5%  tinygo build -size short -o ./build/test.hex -target=arduino-nano33 ./examples/espat/espstation/main.go
 65096  66544   1448   2.2%  tinygo build -size short -o ./build/test.hex -target=feather-m0 ./examples/gps/uart/main.go
 19040  20508   1468   7.7%  tinygo build -size short -o ./build/test.hex -target=bluepill ./examples/ds1307/time/main.go
sum: 34504 (1.0%) geomean: 1.1%
```

Edit: technically this geomean calculation was incorrect because of all the 0s.  It should be closer effectively 0 also.

src/reflect/type.go
dgryski: There isn't.  However, TinyGo also doesn't have any *actual* concurrency so we'd only be interrupted by, well, an interrupt.  This is mostly to solve the case of recursive calls from the same thread -- maybe "lock" was a bad name and "in-progress" or "incomplete" would be a better one.  I can replace this with something from the `sync` package if you'd like.

deadprogram:
What is the status with this PR now?

src/reflect/type.go
kenbell: The recursion path isn't obvious (at least to me), perhaps just add some comments explaining why the lock exists and where the recursion can happen?

dgryski: Updated the comment around `locked`.  PTAL

dgryski:
@aykevl Are you still concerned about the code size impact of this change?

aykevl (commented):
Most of these changes seem fine by me (even though I don't really like using hashmaps in the reflect package - it does work).

What I don't quite follow is `fieldOffsetCacheType`. Can you explain what it does, and why it is necessary?

One possible consequence that may be unexpected is that comparing two interfaces can now allocate, while previously I don't think they would allocate. This can have consequences for interrupt code for example.

dgryski:
I created `fieldOffsetCacheType` so I could put a method on it.  I suppose I could just declare the `fieldOffsetCache` as a variable of the appropriate type and call `lookup()` passing along the cache.  I don't know if there's a global "in interrupt" flag, but we could detect that and use the slow non-allocating path in that case.

Note that all this extra work is because we don't store struct sizes or field offsets in the binary, so we have to calculate them.  But all these loops end up being quite expensive in the case where user code is iterating through all the fields in a struct.  In exchange for a larger binary, we could replace all these caches and maps with direct lookups.

dgryski:
Should we prototype adding size and and offsets to structs and seeing the effect that has?

dgryski:
Planning on closing this in favour of the PR that just adds the sizes to the type structures. However the caches for IsBinary and Comparable seem useful.

aykevl:
I believe there are two bits left over in the meta byte at the start of the type struct. Perhaps you can put `isBinary` (and maybe `Comparable`) in there?

dgryski:
I didn't want to take bits in the meta byte without discussion, but yes that would be a good place for them. I will add that to the other struct-size PR.

dgryski:
Filed https://github.com/tinygo-org/tinygo/issues/3683 instead.

PR #3604, opened 2023-03-26
tests: revert PR#2412

kenbell:
This is a speculative fix to see if test concurrency is causing some of our CI breakages.

This is effectively a full-revert of the original PR.  We may want to opt for something more subtle (like restricting number of parallel go routines and/or making it a command-line option)


kenbell:
Closing as this didn't work

PR #3609, opened 2023-03-27
reflect: use direct calls to runtime string functions

aykevl:
The runtime.stringFromBytesTyped and runtime.stringToBytesTyped functions aren't really necessary, because they have the same LLVM IR signature. Therefore, remove them and link directly to the functions that the compiler uses internally.

Context: https://github.com/tinygo-org/tinygo/pull/3543#discussion_r1144038917


dgryski (approved):
LGTM

deadprogram:

PR #3610, opened 2023-03-27
compiler: only support //go:wasmimport on declared functions

aykevl:
Don't support this pragma on defined functions. It is only meant for importing, not for exporting.

See https://github.com/tinygo-org/tinygo/pull/3589.


deadprogram:
Thank you for the guard condition @aykevl now merging.

PR #3613, opened 2023-03-28
reflect enable more upstream tests

dgryski:


dgryski:
Annoyingly the fix for `TestCanSetField` causes `testdata/reflect.go` to fail.  I think we need an additional `read-only` flag.

deadprogram:
Merging these changes so they don't go stale, as requested. Thanks for this massive amount of work @dgryski

PR #3615, opened 2023-03-28
rp2040: remove SPI DataBits property

aykevl:
As discussed on Slack, I believe this property does more harm than good:

  * I don't think it's used anywhere. None of the drivers use it.
  * It is not fully implemented. While values <= 8 might work fine, values larger than 8 result in extra zero bits (instead of anything sensible).
  * Worse, it doesn't return an error when it's out of range. This is not an optional property: if the SPI peripheral doesn't support a particular number of bits, it should return an error instead of silently limiting the number of bits. This will be confusing to users.

Therefore, I propose we drop it. Maybe there are good uses for it (perhaps for displays that use big endian 16-bit values?), but without a good use case like a driver in tinygo.org/x/drivers, I think it's more trouble than it's worth.

### DIFF
--- a/src/machine/machine_rp2040_spi.go
+++ b/src/machine/machine_rp2040_spi.go
@@ -26,8 +26,6 @@ type SPIConfig struct {
 	LSBFirst bool
 	// Mode's two most LSB are CPOL and CPHA. i.e. Mode==2 (0b10) is CPOL=1, CPHA=0
 	Mode uint8
-	// Number of data bits per transfer. Valid values 4..16. Default and recommended is 8.
-	DataBits uint8
 	// Serial clock pin
 	SCK Pin
 	// TX or Serial Data Out (MOSI if rp2040 is master)
@@ -199,9 +197,6 @@ func (spi SPI) Configure(config SPIConfig) error {
 		return errSPIInvalidSCK
 	}
 
-	if config.DataBits < 4 || config.DataBits > 16 {
-		config.DataBits = 8
-	}
 	if config.Frequency == 0 {
 		config.Frequency = defaultBaud
 	}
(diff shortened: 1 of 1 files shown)


src/machine/machine_rp2040_spi.go
```diff
@@ -231,13 +226,13 @@ func (spi SPI) initSPI(config SPIConfig) (err error) {
 }
 
 //go:inline
-func (spi SPI) setFormat(databits, mode uint8, frameFormat uint32) {
```
soypat, line 234: can we leave the databits argument here? I feel like we are removing valuable information about the API available on the RP2040. We can still remove the DataBits field which I agree is more harmful than good.

soypat (changes requested):

src/machine/machine_rp2040_spi.go
adrianre12: If the underlying transfer functions don't support anything but 8bits would the ability to set a different value of databits via setFormat be meaningful?

aykevl: It's a purely internal API, I don't see why it should be kept? I think it's only confusing to keep it if it doesn't do anything.

kenbell: I think I agree with @aykevl on this one - to me, a simpler implementation wins.

kenbell (approved):

src/machine/machine_rp2040_spi.go
soypat: I guess I can agree that from a purely functional standpoint there's no reason to keep databits. Reason I mentioned it was in case someone was to maintain this repo in the future adding 16bit SPI transfer functionality, in which case the `databits` argument would provide intrinsic documentation. I'm fine either way since this is a hypothetical case and the RP2040 documentation is just so gosh darn good by itself.

soypat (approved):

deadprogram:
Thank you @aykevl @kenbell and @soypat for figuring this out for all of us. Now merging.

PR #3616, opened 2023-03-28
builder: provide fallback code for nix

aykevl:
Change ReadBuildID to always use the fallback code if it can't find the build ID in a conventional way.
This is the case when using Nix on MacOS, which for some reason decided to pass `-no-uuid` to the linker breaking our support for reading the build ID.

...I wish this was a feature of the standard library, but it appears we have to invent it ourself.

### DIFF
--- a/builder/buildid.go
+++ b/builder/buildid.go
@@ -77,27 +77,29 @@ func ReadBuildID() ([]byte, error) {
 			}
 			return raw[4:], nil
 		}
-	default:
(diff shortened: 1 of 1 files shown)


deadprogram:
Is there anyone (besides @aykevl) who runs nix and can verify this?

aykevl:
I don't actually run Nix. It was asked on Slack, but IIRC this fix didn't even work.
Closing, as I'm not sure it even fixes a bug.

Issue #3603, opened 2023-03-26
CI failing in 'net' package often on Windows


kenbell:
The stdlib-test-windows is often failing in CI.  It appears the binary that is created to run tests in the 'net' package is not being created, causing the failure.

kenbell:
From some poking, i think this failure is related in some way to the parallelism of the tests.  The test running code tries to setup a go routine for every test package in parallel.  Given the limited resources of the CI runners, I suspect we're hitting resource constraints.  I've tried a hacky solution of running the tests sequentially with (apparently) much better success.

kenbell:
I did an experiment here to disable parallel test execution and it didn't solve the problem (first build failed to run), so I don't think it's directly related to running tests in parallel: https://github.com/tinygo-org/tinygo/pull/3604

deadprogram:
Closing since this was addressed. Thanks!

PR #3617, opened 2023-03-28
build: test net on linux & mac only (random CI fails on windows)

kenbell:
This is a far from ideal answer to https://github.com/tinygo-org/tinygo/issues/3603, but it's causing a lot of need to re-build PRs for no benefit (given no-one has located the root-cause of the PR breaks caused by the net package on Windows).

This change still tests the package on Linux & Mac, providing some protection against regressions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -307,7 +307,6 @@ TEST_PACKAGES_FAST = \
 	internal/profile \
 	math \
 	math/cmplx \
-	net \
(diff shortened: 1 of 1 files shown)


dgryski (approved):
LGTM

deadprogram:
It least it stops the spurious fails while we figure out if there is actual root cause.

PR #3618, opened 2023-03-29
machine/gba: rename display

deadprogram:
This PR changes the machine/gba implementation for the display to use pointer to the one and only display so it can be called safely from interrupt handlers.

### DIFF
--- a/src/machine/machine_gameboyadvance.go
+++ b/src/machine/machine_gameboyadvance.go
@@ -38,27 +38,27 @@ func (p Pin) Set(value bool) {
 	// do nothing
 }
 
-var Display = FramebufDisplay{(*[160][240]volatile.Register16)(unsafe.Pointer(uintptr(gba.MEM_VRAM)))}
+var Display = DisplayMode3{(*[160][240]volatile.Register16)(unsafe.Pointer(uintptr(gba.MEM_VRAM)))}
 
-type FramebufDisplay struct {
+type DisplayMode3 struct {
 	port *[160][240]volatile.Register16
 }
 
-func (d FramebufDisplay) Configure() {
+func (d *DisplayMode3) Configure() {
 	// Use video mode 3 (in BG2, a 16bpp bitmap in VRAM) and Enable BG2
 	gba.DISP.DISPCNT.Set(gba.DISPCNT_BGMODE_3<<gba.DISPCNT_BGMODE_Pos |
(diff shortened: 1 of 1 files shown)


aykevl:
I don't entirely follow, what's the problem this PR fixes?

deadprogram:
I tried to update and run one of these GBA examples: https://gitlab.com/TitiMoby/tinygo-examples/-/blob/main/gba/03-Gopher/main.go using the newest TinyGo and friends.

As part of that testing, I discovered several interrelated problems. Now that we (correctly) do not allow heap allocation to occur in interrupt handlers, I needed to make some changes to TinyFont located in this PR: https://github.com/tinygo-org/tinyfont/pull/41

I found that I also needed to change the display reference to use a pointer, as it otherwise would not work as expected when called as part of the updated example. I then also did some renaming since the current implementation is actually GBA Mode 3. I am starting to play with implementing Mode 4 over in this incomplete repo/branch https://github.com/tinygo-org/tinygba/blob/display/display.go

aykevl:
I agree with the name change, but I don't understand why it needs to be a pointer?
Can you give me a way to reproduce the issue?

deadprogram:
> I agree with the name change, but I don't understand why it needs to be a pointer?

@aykevl you were of course correct about not needing to be pointer. The issue I was having was in TinyFont.

The only other thing I have added to this PR now is enabling the VBLANK interrupt when the display `Configure()` is called, since that is the most likely reason to be using it.

aykevl:
So, you want to update the display in a VBLANK interrupt? I'm not entirely sure that's a good idea to be honest. The stack is very small so you'll easily run out of stack and corrupt some memory.

This is how I implemented it in my [experimental board package](https://github.com/aykevl/board/blob/main/board-gameboy-advance.go#L50-L56):

```go
func (d mainDisplay) WaitForVBlank(time.Duration) {
	// Wait until the VBlank flag is set.
	// TODO: sleep until the next VBlank instead of busy waiting.
	// (See VBlankIntrWait)
	for gba.DISP.DISPSTAT.Get()&(1<<gba.DISPSTAT_VBLANK_Pos) == 0 {
	}
}
```

So basically every loop I wait for a VBlank and then start updating the screen. This has actually worked quite well for me, fixing the timing issues I previously had in [this little game](https://github.com/aykevl/things/blob/master/games/brick-breaker/main.go) on the GameBoy Advance.

deadprogram:
OK, removed the interrupt enable. Thanks for the above technique!

PR #3623, opened 2023-03-30
rp2040: unify all linker scripts using LDFLAGS

aykevl:
The only thing that's different between all these chips is the flash size, which can easily be passed as a linker flag instead. This removes a bunch of duplicate code in an uncommon language (linker script).

I've also fixed a few boards with incorrect flash sizes:

  * nano-rp2040 has 16MB instead of 2MB
  * macropad-rp2040 has 8MB instead of 2MB
  * gopher-badge has 8MB instead of 1MB


deadprogram (approved):
What an excellent idea, thanks @aykevl

deadprogram:
Thanks @aykevl now merging.

PR #3625, opened 2023-03-31
testing cleanups

dgryski:
Make the testing code a bit nicer:
* `-count` for running benchmarks multiple times
*  `-test.v` is now unbuffered so if your test crashes you see the output


deadprogram (approved):
:test_tube: thanks @dgryski

PR #3627, opened 2023-03-31
main: set WASMTIME_BACKTRACE_DETAILS when running in wasmtime.

dgryski:
I find myself consistently running tests, seeing them panic, and then immediately running them again with this environment variable set.  It's easier to just have tinygo do this for me.

### DIFF
--- a/main.go
+++ b/main.go
@@ -792,6 +792,7 @@ func buildAndRun(pkgName string, config *compileopts.Config, stdout io.Writer, c
 		}
 	}
 	var args, env []string
+	var extraCmdEnv []string
 	if needsEnvInVars {
 		runtimeGlobals := make(map[string]string)
 		if len(cmdArgs) != 0 {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -873,7 +879,12 @@ func buildAndRun(pkgName string, config *compileopts.Config, stdout io.Writer, c
 	} else {
 		cmd = exec.Command(name, args...)
 	}
-	cmd.Env = env
+	if len(extraCmdEnv) > 0 {
+		cmd.Env = append(cmd.Env, env...)
+		cmd.Env = append(cmd.Env, extraCmdEnv...)
+	} else {
+		cmd.Env = env
+	}
```
aykevl: I'm not sure this is more readable, but at least it does basically the same thing without needing an if/else.

```suggestion
	cmd.Env = append([]string, env...)
	cmd.Env = append(cmd.Env, extraCmdEnv)
```

To me it's more readable, not sure about other people :)

aykevl (approved):
This seems fine by me, but why not configure your environment to `WASMTIME_BACKTRACE_DETAILS=1`? That's what I've done, so that I don't need to set the environment variable manually each time.

(It's in my `~/.profile` file)

main.go
dgryski: I think I was trying to avoid the extra allocation the default case where we didn't have it, or potentially modifying the `env` array.   But since this is the compiler it probably doesn't matter as much.

dgryski:
I don't know why I never just modified my default environment.  Maybe because I run a bunch of wasmtime things and most of the time I don't want the full backtraces?  I have no idea.. Anyways, I figured this also makes for nicer backtraces out-of-the-box for people who *don't* know about the wasmtime-specific environment variables.

main.go
aykevl: It happens only once per run (in the compiler!), so this is really not a place to worry about performance. So then, the thing that I find important the most is readability :)

Anyway, LGTM either way, you can fix it if you think it's worth it.

PR #3629, opened 2023-03-31
add -test.skip

dgryski:
Should have just pulled in `match.go` from the beginning..

### DIFF
--- a/main.go
+++ b/main.go
@@ -236,6 +236,9 @@ func Test(pkgName string, stdout, stderr io.Writer, options *compileopts.Options
 	if testConfig.RunRegexp != "" {
 		flags = append(flags, "-test.run="+testConfig.RunRegexp)
 	}
+	if testConfig.SkipRegexp != "" {
+		flags = append(flags, "-test.skip="+testConfig.SkipRegexp)
+	}
 	if testConfig.BenchRegexp != "" {
 		flags = append(flags, "-test.bench="+testConfig.BenchRegexp)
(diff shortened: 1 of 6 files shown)


dgryski:
Bah. Need import updated tests too.

main.go
```diff
@@ -1427,6 +1430,7 @@ func main() {
 		flag.BoolVar(&testConfig.Verbose, "v", false, "verbose: print additional output")
 		flag.BoolVar(&testConfig.Short, "short", false, "short: run smaller test suite to save time")
 		flag.StringVar(&testConfig.RunRegexp, "run", "", "run: regexp of tests to run")
+		flag.StringVar(&testConfig.SkipRegexp, "skip", "", "run: regexp of tests to run")
```
aykevl: Copy-paste bug:
```suggestion
		flag.StringVar(&testConfig.SkipRegexp, "skip", "", "run: regexp of tests to skip")
```

aykevl (commented):
Looks good to me.

I assume match.go is mostly from upstream? I didn't check it thoroughly, but if it is basically a copy from upstream Go I'll trust it is correct.

dgryski:
Yes. Match had previously been imported from upstream. Now it's a newer version with skip support.

The only change is the bare metal fake matcher, which I moved out of that file to make updating easier in the future.

dgryski:
PTAL

aykevl (approved):

PR #3630, opened 2023-03-31
rp2040: remove SPI deadline

aykevl:
Removing it improves SPI performance by about 20% for me (updating the display of a Gopher Badge). I also don't think it's actually needed: if the SPI (in controller mode) actually blocks like that, there must be something wrong with the configuration. It can't be caused by an external device that doesn't respond, for example.


deadprogram:
Appears to be working for me exactly as expected with this change. Probably was defensive code that is not actually needed? Now merging, thanks for code removal/speed increase @aykevl

PR #3633, opened 2023-04-01
reflect: handle Convert'ing between identical underlying types

dgryski:
Needed for go-jose/v3


aykevl (approved):

aykevl:
> reflect: handle Convert'ing between identical underlying types

~What did you do here? That's some weird Unicode stuff. (Or is it my system that's weird?)~ nevermind, a logout+login fixed it for me. Probably related to the updates I just installed.

PR #3634, opened 2023-04-01
rp2040: use DMA for send-only SPI transfers

aykevl:
This improves performance slightly. It also is some groundwork for better DMA support in TinyGo in the future.

I'm not entirely sure why it improves performance (in theory the old code should already saturate the SPI bus) but it does, so 🤷

### DIFF
--- a/src/machine/machine_rp2040_spi.go
+++ b/src/machine/machine_rp2040_spi.go
@@ -5,6 +5,7 @@ package machine
 import (
 	"device/rp"
 	"errors"
+	"unsafe"
 )
 
 // SPI on the RP2040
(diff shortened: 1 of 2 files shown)


src/machine/machine_rp2040_spi.go
```diff
@@ -288,14 +289,51 @@ func (spi SPI) isBusy() bool {
...
+
+	// Wait until the transfer is complete.
+	// TODO: do this more efficiently:
+	//   - Add a new API to start the transfer, without waiting for it to
+	//     complete. This way, the CPU can do something useful while the
+	//     transfer is in progress.
+	//   - If we have to wait, do so by waiting for an interrupt and blocking
+	//     this goroutine until finished (so that other goroutines can run or
+	//     the CPU can go to sleep).
+	for ch.CTRL_TRIG.Get()&rp.DMA_CH0_CTRL_TRIG_BUSY != 0 {
```
kenbell, line 331: Instead of a new API, etc - can we just call `gosched()`.  I have a change pending in PR #3301 to make it available to machine package.

That way, goroutines can be used to do 'useful things' while SPI is waiting.

aykevl: Yes that's possible, but it forces the use of multiple goroutines on the users of the API if they want to improve performance. Those goroutines have extra overhead, which you might not want in performance critical code.

In any case, rather than calling `gosched()` the more efficient approach is using some sort of semaphore to block this goroutine which can then be unblocked from an interrupt. Calling `gosched()` in a loop just wastes power if there is no other goroutine to run.

kenbell: Until then, `gosched()` is better than tight loops?    Both consume power - but one allows other code to run.  I added gosched to a bunch of i2c code so that it was possible to run an i2c controller and target in the same program concurrently, talking to each other.

aykevl: > Until then, `gosched()` is better than tight loops?

Yes, that's true. (With one nitpick: `gosched()` is probably a lot slower to return even if there are no other goroutines, so that might add some latency. I'm not sure it's significant though).

deadprogram: Perhaps we should replace the implementation of the i2c code to use the semaphore/IRQ? In the meantime it does not bother me to have both styles of implementation to explore a little bit what works, as long the the API people use does not change it is not really a problem.

deadprogram:
Merging, thanks for this @aykevl and to @kenbell for review.

PR #3635, opened 2023-04-01
rp2040: use 4MHz as default frequency for SPI

aykevl:
This matches other SPI implementations. I think the original value of 115200 was from a confusion with UART.


deadprogram:
Makes sense to me. Now merging, thanks @aykevl

Issue #3474, opened 2023-02-22
USB: Composite HID Devices


neildavis:
Firstly,  should [`joystick`](https://github.com/tinygo-org/tinygo/tree/release/src/machine/usb/joystick) be located under [`machine/usb/hid`](https://github.com/tinygo-org/tinygo/tree/release/src/machine/usb/hid)  (along with [`keyboard`](https://github.com/tinygo-org/tinygo/tree/release/src/machine/usb/hid/keyboard) & [`mouse`](https://github.com/tinygo-org/tinygo/tree/release/src/machine/usb/hid/mouse)) instead of in `machine/usb` (with siblings [`cdc`](https://github.com/tinygo-org/tinygo/tree/release/src/machine/usb/cdc) & [`midi`](https://github.com/tinygo-org/tinygo/tree/release/src/machine/usb/midi))? As I understand, CDC, MIDI & HID are all different USB 'device classes' whereas `joystick` is [clearly](https://github.com/tinygo-org/tinygo/blob/v0.27.0/src/machine/usb/joystick/joystick.go#L46) [using HID](https://github.com/tinygo-org/tinygo/blob/0d56dee00f49bd50eb373c02c30062a75ec28f10/src/machine/usb.go#L277)

Secondly, (and more importantly) it appears that more than one of `mouse`, `keyboard` & `joystick` cannot be used in a 'composite HID device'. Whist `mouse` & `keyboard` can coexist, only one device actually works and adding `joystick` breaks everything. e.g. the program below will not work as written with the following observations:

1. Whilst the mouse & keyboard show up in Linux as `/dev/input/eventX` & `/dev/input/mouse1`, no events are received by `evtest` on these devices.
2. No event device or joystick device (e.g. `/dev/input/js0`) is created in Linux for the joystick.
3. If the joystick code is removed, then the mouse & keyboard devices are both created but **only mouse events** are received by `evtest`

Using: `tinygo version 0.28.0-dev-1065f06 linux/amd64 (using go version go1.20.1 and LLVM version 15.0.0)` from [this build run](https://github.com/tinygo-org/tinygo/actions/runs/4223759755) - latest available commit on `dev` at time of writing:

```go
package main

import (
	"machine/usb/hid/keyboard"
	"machine/usb/hid/mouse"
	"machine/usb/joystick"
	"time"
)

// Entry point
func main() {
	kb := keyboard.Port()
	ms := mouse.Port()
	js := joystick.Port()

[...]

PR #3626, opened 2023-03-31
machine/usb/hid/joystick: move joystick under HID and remove duplicate code

deadprogram:
machine/usb/hid/joystick: move joystick under HID as it belongs and also remove duplicate code.

As discussed here: https://github.com/tinygo-org/tinygo/issues/3474#issuecomment-1478487194

PR #3636, opened 2023-04-02
machine/usb: descriptor refactor

deadprogram:
This PR is to refactor the USB descriptors to make them modular. This is in part to address what is being discussed in https://github.com/tinygo-org/tinygo/issues/3474

There is still further work needed to complete this feature, this PR in intended to continue the conversation in the form of code.


deadprogram:
cc @neildavis

deadprogram:
Note also this PR includes #3626

deadprogram:
This PR now has the mostly complete refactoring of the USB descriptors. The only parts that are still incomplete are to finish creating types for HID keyboard and mouse descriptors.

EDIT: I added comments to the HID reports, that is probably enough for now.

deadprogram:
PTAL @bgould @ardnew @sago35 and anyone else who might be interested.

deadprogram:
Had to do one last thing on this PR, which was to cleanup `usb.EnableJoystick()` so it would use the proper method to change the `ClassLength` instead of slice modification magic.

sago35:
I have started reviewing the code. I think it is good work. On the other hand, I may have to think a little more when adding new devices such as msd.

However, I still think it is a very good improvement.

src/machine/usb/descriptor/midi.go
```diff
@@ -0,0 +1,249 @@
...
+	0x40, // MaxPacketSizeL
+	0x00, // MaxPacketSizeH
+	0x00, // Interval
+	0x00, // refresh
+	0x00, // sync address
+}
+
+var EndpointEP7OUT = EndpointType{
+	data: endpointEP7OUT[:],
+}
```
sago35, line 204: Shouldn't it be implemented in endpoint.go?

deadprogram: I was thinking about that, however since these endpoints have values specific to the MIDI protocol (note the length difference).

Perhaps they should have `MIDI` in the endpoint name to make that more explicit?

sago35:
midi and joystick may not be working.

deadprogram:
> midi and joystick may not be working.

I will check that out later today, thanks for letting me know.

deadprogram:
@sago35 I tested this branch, and was not able to get keyboard or MIDI to work using circuitplayground express board.

Then I tested the same thing on the `dev` branch, and again I was not able to get either keyboard or MIDI to work.

Finally I tested the last release, and it also did not work :frowning_face: 

Are you having this same problem with your devices? Which processor are you using?

deadprogram:
OK, checked out code using Clue (nrf52840)

In this branch  using Clue the Keyboard works, however MIDI does not. Have not tried Joystick yet.

deadprogram:
@sago35 did some more testing last night, appears the `dev` branch is not working on rp2040 or samd21 when I tested using the `usb-midi` example.

The release branch still working as expected when I plugged both devices into my Macbook. However `dev` not working for either board. Looks like we will need to bisect changes to locate the cause.

deadprogram:
Did a `git bisect` and found that https://github.com/tinygo-org/tinygo/commit/f41b6a3b96574c29002b2b6c5585f9faa3766a29 is the commit where this stopped working. Looking into where we are doing the allocation that is causing this problem now...

deadprogram:
DId a session with `gdb` and an ItsyBitsy-M4. The results were not what I expected:

```
$ tinygo gdb -target itsybitsy-m4 -programmer=jlink examples/usb-midi                                                                                                        
GNU gdb (Ubuntu 9.2-0ubuntu1~20.04.1) 9.2                                                                                                                                    
Copyright (C) 2020 Free Software Foundation, Inc.                                                                                                                            
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>                                                                                                
This is free software: you are free to change and redistribute it.                                                                                                           
There is NO WARRANTY, to the extent permitted by law.                                                                                                                        
Type "show copying" and "show warranty" for details.                                                                                                                         
This GDB was configured as "x86_64-linux-gnu".                                                                                                                               
Type "show configuration" for configuration details.                                                                                                                         
For bug reporting instructions, please see:                                                                                                                                  
<http://www.gnu.org/software/gdb/bugs/>.                                                                                                                                     
Find the GDB manual and other documentation resources online at:                                                                                                             
    <http://www.gnu.org/software/gdb/documentation/>.                                                                                                                        
                                                                                                                                                                             
For help, type "help".                                                                                                                                                       
Type "apropos word" to search for commands related to "word"...                                                                                                              
Reading symbols from /tmp/tinygo2055889492/main...                                                                                                                           
Remote debugging using :3333                                                                                                                                                 
0x00000004 in ?? ()                                                                                                                                                          
Loading section .text, size 0xd9b0 lma 0x4000                                                                                                                                
Loading section .tinygo_stacksizes, size 0x4 lma 0x119b0                                                                                                                     
Loading section .data, size 0x68c lma 0x119b4                                                                                                                                
Start address 0x00007638, load size 57408                                                                                                                                    
Transfer rate: 7007 KB/sec, 9568 bytes/write.                                                                                                                                
Expected an decimal digit (0-9)                                                                                                                                              
(gdb) bt                                                                                                                                                                     
#0  0x00007638 in Reset_Handler () at /home/ron/Development/tinygo/tinygo/src/runtime/panic.go:150
```

I get the same error in both the `dev` branch and this feature branch. Calling @aykevl to help on this one, please...

sago35:
@deadprogram @aykevl 
I feel that this change ( https://github.com/tinygo-org/tinygo/commit/f41b6a3b96574c29002b2b6c5585f9faa3766a29 ) has a little too much impact.
For example, the tinygo-org/bluetooth code at hand did not work either.
Also, the joystick code no longer works.

Of course it is better to fix the heap allocation, but it takes time for the user to realize why it is not working.
In the next release, it may be better to keep warnings instead of errors, or to allow optin.

https://github.com/tinygo-org/tinygo/commit/f41b6a3b96574c29002b2b6c5585f9faa3766a29

kenbell:
@deadprogram I think Ayke suggested setting this bp to try to catch it in the original PR: https://github.com/tinygo-org/tinygo/pull/1460
```
b runtime.runtimePanic
```

@sago35 I think we can make interrupt allocation checks either opt-in or opt-out fairly easily using build tags.  The longer we leave it `opt-in` though, the more code will be written that does allocations inside interrupt handlers that needs fixing.  Perhaps a plan like:

1. keep dev opt-out until we're close to next release
2. make it opt-in for a short period for the next release
3. make dev opt-out again
4. revisit at the subsequent release to see if we're ready to make the switch permanent / opt-out

deadprogram:
> I think Ayke suggested setting this bp to try to catch it in the original PR

@kenbell that was what was surprising, that the runtime panic was not that one, but was `unsafeSlicePanic`. So something else perhaps is going on? Not sure, which is why I think I need help from @aykevl to figure this out.

sago35:
My main concern is that the code that was working with tinygo0.27 will no longer work.
It would be best if we could warn them at the time of tinygo build.

deadprogram:
I had forgotten to add the `-opt=1` flag to my command, hence was not getting the correct debugging info before.

Here is debug info from commit https://github.com/tinygo-org/tinygo/commit/f41b6a3b96574c29002b2b6c5585f9faa3766a29

```
$ tinygo gdb -target itsybitsy-m4 -programmer=jlink -opt=1 examples/usb-midi                                                                                                 
GNU gdb (Ubuntu 9.2-0ubuntu1~20.04.1) 9.2                                                                                                                                    
Copyright (C) 2020 Free Software Foundation, Inc.    
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>         
This is free software: you are free to change and redistribute it.                    
There is NO WARRANTY, to the extent permitted by law.                                                    
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".                                                                         
Type "show configuration" for configuration details.                                                                   
For bug reporting instructions, please see:                                                              
<http://www.gnu.org/software/gdb/bugs/>.                                                                 
Find the GDB manual and other documentation resources online at:                                                                                                             
    <http://www.gnu.org/software/gdb/documentation/>.                                                                                                                        
                                                                                                         
For help, type "help".                                                                                                                                                       
Type "apropos word" to search for commands related to "word"...                                                        
Reading symbols from /tmp/tinygo1796678531/main...                                                                     
Remote debugging using :3333         
0xdeadbeee in ?? ()                                                                                                                                                                                                
Loading section .text, size 0x3a90 lma 0x4000                                                                                                                                                                      
Loading section .tinygo_stacksizes, size 0x4 lma 0x7a90                               
Loading section .data, size 0x13c lma 0x7a94                                                                                                                                                                       
Start address 0x00005e8a, load size 15312
Transfer rate: 2136 KB/sec, 5104 bytes/write.                                                                          
Expected an decimal digit (0-9)      
(gdb) b runtime.runtimePanic                        
Breakpoint 1 at 0x51d4                              
(gdb) bt                                                                                                 
#0  0x00005e8a in runtime.preinit () at /home/ron/Development/tinygo/tinygo/src/machine/usb/cdc/usbcdc.go:114                                                                                                                                 
#1  Reset_Handler () at /home/ron/Development/tinygo/tinygo/src/runtime/runtime_atsamd51.go:19                                                                                                                                                
(gdb) f                                             
#0  0x00005e8a in runtime.preinit () at /home/ron/Development/tinygo/tinygo/src/machine/usb/cdc/usbcdc.go:114                                                                                                                                 
114             usbcdc.Write([]byte{c})                    
(gdb) l                                                    
109             return len(data), nil                      
110     }                                                  
111                                                        
112     // WriteByte writes a byte of data to the USB CDC interface.                                                   
113     func (usbcdc *USBCDC) WriteByte(c byte) error {                                                                
114             usbcdc.Write([]byte{c})                    
115             return nil                                 
116     }                                                  
117                                                        
118     func (usbcdc *USBCDC) DTR() bool {                 
(gdb) c                                                    
Continuing.                                                

Program received signal SIGTRAP, Trace/breakpoint trap.                                                                
0xdeadbeee in ?? ()                                        
(gdb) bt                                                   
#0  0xdeadbeee in ?? ()                                    
(gdb) f                                                    
#0  0xdeadbeee in ?? ()
```

deadprogram:
Played around a bit more. By commenting out this line https://github.com/tinygo-org/tinygo/blob/dev/src/examples/usb-midi/main.go#L23 the example works on `dev` branch.

So I think this branch with the refactor has an error, but there is separately something in `dev` branch that does not allow using `fmt` package from the handler function.

kenbell:
@deadprogram if I've understood the midi code correctly, the handler is invoked in the interrupt handler context - it's highly unlikely `fmt.Printf` is free from allocations, so it makes sense that it can't be used in that handler.

It might be an idea to document SetHandler that any handler must be interrupt-safe?  (either that or we need to use a channel or similar to exit the interrupt context)

deadprogram:
Corrected the descriptor errors that were not allowing MIDI and keyboard to work.

The call to `fmt` in the `examples/usb-midi` needs to also be removed for that to work, but is not specific to this PR, so I will make another PR that addresses that.

Now verifying joystick...

deadprogram:
Had to make a few more changes in descriptors, but now `Keyboard`, `Midi`, and `Joystick` all appear to be working.

I will submit a separate PR to remove `fmt` from `examples/usb-midi`.

Now ready for "final" review.

deadprogram:
Added https://github.com/tinygo-org/tinygo/pull/3660 to correct example as mentioned.

sago35:
I believe the code is fine.
But as far as I have tested, it works fine except for the pattern using joystick.UseSettings().
I will investigate a little more.

deadprogram:
@sago35 can you please share the repo with the code in which you are using `joystick.UseSettings()` so I can take a look?

sago35:
@deadprogram 
I am thinking that perhaps UseSettings is not being changed properly.
https://github.com/sago35/pico-ffb-steering-wheel/tree/7a1098872fdc3fb6f3747a69daca0c02974097f9

deadprogram:
Just wondering why you need this code exactly? https://github.com/sago35/pico-ffb-steering-wheel/blob/7a1098872fdc3fb6f3747a69daca0c02974097f9/pid/pid_handler.go#L173-L195

I don't know a lot about HID joysticks, perhaps you can briefly explain what that code is doing?

sago35:
```
(gdb) b runtime.slicePanic
Breakpoint 1 at 0x10004db4: file C:\tinygo\tinygo\src\runtime/panic.go, line 137.
Note: automatically using hardware breakpoints for read-only addresses.
(gdb) c
Continuing.

Breakpoint 1, 0x10004db4 in runtime.slicePanic () at C:\tinygo\tinygo\src\runtime/panic.go:137
137             runtimePanic("slice out of range")
(gdb) bt
#0  0x10004db4 in runtime.slicePanic () at C:\tinygo\tinygo\src\runtime/panic.go:137
#1  0x10005f00 in runtime.run$1 () at C:\tinygo\tinygo\src\machine\usb\descriptor/hid.go:82
#2  0x10005bf2 in <goroutine wrapper> () at C:\tinygo\tinygo\src\runtime/scheduler_any.go:23
#3  0x100002fc in tinygo_startTask () at C:\\tinygo\\tinygo\\src\\internal\\task/task_stack_cortexm.S:26
```

sago35:
It appears that -1 is returned here and runtime.slicePanic is called. I will look into it more tomorrow.

https://github.com/tinygo-org/tinygo/blob/8355ebcf550839543a98ee61aa1c10c55c5edd68/src/machine/usb/descriptor/hid.go#L80

deadprogram:
> It appears that -1 is returned here and runtime.slicePanic is called. I will look into it more tomorrow.

I think that happened because the Joystick had already been enabled by the default. I just added some code to handle that case, please give it a try.

src/machine/usb/descriptor/hid.go
```diff
@@ -0,0 +1,202 @@
...
+
+// FindClassHIDType tries to find the ClassHID data in the descriptor.
+func FindClassHIDType(data, section []byte) (ClassHIDType, error) {
+	if len(data) < ClassHIDTypeLen || len(section) == 0 {
+		return ClassHIDType{}, errNoClassHIDFound
+	}
+
+	// search only for ClassHIDType without the ClassLength,
+	// in case it has already been set.
+	idx := bytes.Index(data[:ClassHIDTypeLen-2], section)
```
sago35: ClassHIDTypeLen is 9, so this will not work.

```suggestion
	idx := bytes.Index(data, section)
```

src/machine/usb.go
```diff
@@ -302,11 +302,15 @@ func EnableMIDI(txHandler func(), rxHandler func([]byte), setupHandler func(usb.
 
 // EnableJoystick enables HID. This function must be executed from the init().
 func EnableJoystick(txHandler func(), rxHandler func([]byte), setupHandler func(usb.Setup) bool, hidDesc []byte) {
-	idx := bytes.Index(usb.DescriptorCDCJoystick.Configuration, []byte{
-		0x09, 0x21, 0x11, 0x01, 0x00, 0x01, 0x22,
-	})
-	binary.LittleEndian.PutUint16(usb.DescriptorCDCJoystick.Configuration[idx+7:idx+9], uint16(len(hidDesc)))
-	usb.DescriptorCDCJoystick.HID[2] = hidDesc
+	class, err := descriptor.FindClassHIDType(descriptor.CDCJoystick.Configuration, descriptor.ClassHIDJoystick.Bytes())
```
sago35, line 305: EnableJoystick has been called twice, and the second call seems to fail to find it because ClassLength has already been set.
Why it is being called twice requires a little more checking.

sago35:
@deadprogram 
Please rebase to the latest dev branch.

deadprogram:
> Please rebase to the latest dev branch.

Done.

src/machine/usb/descriptor/hid.go
deadprogram: Oops, I had it reversed. Renamed the parameter names and corrected, seems to be working now.

deadprogram:
@sago35 I think this PR is now ready, please take another look.

deadprogram:
@sago35 rebased just now again against the latest `dev`.

sago35 (approved):
LGTM

The following example and the FFB JoyStick also worked.

```
$ tinygo flash --target feather-rp2040 --size short --monitor ./src/examples/echo
$ tinygo flash --target feather-rp2040 --size short --monitor ./src/examples/hid-keyboard
$ tinygo flash --target feather-rp2040 --size short --monitor ./src/examples/hid-mouse
$ tinygo flash --target feather-rp2040 --size short --monitor ./src/examples/usb-midi
$ tinygo flash --target feather-rp2040 --size short --monitor ./src/examples/hid-joystick
```

FFB JoyStick
https://github.com/sago35/pico-ffb-steering-wheel/tree/work

deadprogram:
Thank you very much @sago35 and @kenbell for all of your assistance! Now merging.

tuffrabit:
> LGTM
> 
> The following example and the FFB JoyStick also worked.
> 
> ```
> $ tinygo flash --target feather-rp2040 --size short --monitor ./src/examples/hid-joystick
> ```

Hey! The hid-joystick example worked for you on a rp2040 board? I can get it to compile but it does not work. Windows 10 won't even enumerate it after flashing. It consistently gets either a error code 43 or 10 in the device manager. You actually got it to build, flash, enumerate, and then generate input in joy.cpl or something similar?

This is on 0.28.1 btw

PR #3638, opened 2023-04-03
samd51: use correct SPI frequency

aykevl:
The SPI frequency was rounded up, not rounded down. This meant that if you wanted to configure 15MHz for example, it would pick the next available frequency (24MHz). That's unsafe, the safe option is to round down and the SPI support for most other chips also rounds down for this reason.

In addition, I've improved SPI clock selection so that it will pick the best clock of the two, widening the available frequencies. See the comments in the patch for details.

---

Found this bug while investigating DMA for the samd51. The samd21 has a similar bug and should also be fixed.


deadprogram:

PR #3640, opened 2023-04-07
Improve handling of multiple identical MCUs connected via USB

### LINKED ISSUE #3637: -monitor flag is confused with -port flag
I have two identical raspberry pi picos, /dev/ttypACM0 and /dev/ttyACM1.

When I flash using both -monitor and -port together, the right port is flashed but monitor connects to the wrong port:

```
sfeldma@nuc:~/work/tinygo$ ls /dev/ttyACM*
/dev/ttyACM0  /dev/ttyACM1

sfeldma@nuc:~/work/tinygo$ tinygo flash -monitor -target pico -size short -port /dev/ttyACM0 src/examples/echo/echo.go 
   8548     108    3288 |    8656    3396
Connected to /dev/ttyACM1. Press Ctrl-C to exit.       <-- monitor connected to /dev/ttyACM1

sfeldma@nuc:~/work/tinygo$ tinygo flash -monitor -target pico -size short -port /dev/ttyACM1 src/examples/echo/echo.go 
   8548     108    3288 |    8656    3396
Connected to /dev/ttyACM0. Press Ctrl-C to exit.       <-- monitor connected to /dev/ttyACM0
```

If I get real crazy and add a third pico, nothing works:

```
sfeldma@nuc:~/work/tinygo$ ls /dev/ttyACM*
/dev/ttyACM0  /dev/ttyACM1  /dev/ttyACM2
sfeldma@nuc:~/work/tinygo$ tinygo flash -monitor -target pico -size short -port /dev/ttyACM1 src/examples/echo/echo.go 
   8548     108    3288 |    8656    3396
error: unable to search for a default USB device - use -port flag, available ports are /dev/ttyACM0, /dev/ttyACM1, /dev/ttyACM2

sfeldma@nuc:~/work/tinygo$ tinygo flash -monitor -target pico -size short src/examples/echo/echo.go 
   8548     108    3288 |    8656    3396
error: failed to flash /tmp/tinygo1733382844/main.uf2: unable to locate device: RPI-RP2
```

I'm on dev branch of tinygo:

```
sfeldma@nuc:~/work/tinygo$ tinygo version
tinygo version 0.28.0-dev-2c0f61ca linux/amd64 (using go version go1.19.3 and LLVM version 15.0.0)

```

kenbell:
At least partially fixes #3637 

There are a few things:

1. Don't silently fail to flash if there is an error doing reset using the baud-rate trick because no port could be found

2. Improve error message when there are multiple candidate ports with same VID/PID by fixing the conditions when different errors are shown

3. Pass the port to the Monitor function after flashing (not this already works when using the `monitor` command)

### DIFF
--- a/main.go
+++ b/main.go
@@ -1143,12 +1143,12 @@ func getDefaultPort(portFlag string, usbInterfaces []string) (port string, err e
 	}
 
 	if len(portCandidates) == 0 {
-		if len(usbInterfaces) > 0 {
-			return "", errors.New("unable to search for a default USB device - use -port flag, available ports are " + strings.Join(ports, ", "))
-		} else if len(ports) == 1 {
+		if len(ports) == 1 {
 			return ports[0], nil
-		} else {
+		} else if len(ports) > 1 {
 			return "", errors.New("multiple serial ports available - use -port flag, available ports are " + strings.Join(ports, ", "))
+		} else {
+			return "", errors.New("unable to search for a default USB device - use -port flag, available ports are " + strings.Join(ports, ", "))
 		}
 	}
 
--- a/main.go
+++ b/main.go
@@ -532,7 +532,12 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 		return fmt.Errorf("unknown flash method: %s", flashMethod)
 	}
 	if options.Monitor {
-		return Monitor("", options)
+		port, err := getDefaultPort(port, config.Target.SerialPort)
+		if err != nil {
+			return &commandError{"failed to locate port to monitor", port, err}
+		}
+
+		return Monitor(port, options)
 	}
 	return nil
 }


main.go
```diff
@@ -426,14 +426,16 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
...
 	if config.Target.PortReset == "true" && flashMethod != "openocd" {
 		port, err := getDefaultPort(port, config.Target.SerialPort)
-		if err == nil {
-			err = touchSerialPortAt1200bps(port)
-			if err != nil {
-				return &commandError{"failed to reset port", port, err}
-			}
-			// give the target MCU a chance to restart into bootloader
-			time.Sleep(3 * time.Second)
+		if err != nil {
```
kenbell: This particular change may be questionable... before, if the device was already in a UF2 bootloader, flashing would proceed silently - with this change it would fail.

Not failing is bad (confusing) and failing is bad for a common case, perhaps should do the volume check first, and skip the reset if the volume already exists to separate these cases?

deadprogram: > perhaps should do the volume check first, and skip the reset if the volume already exists to separate these cases

That sounds like a very good idea.

deadprogram: @kenbell will you make this update to the PR?

kenbell: I'm thinking for now, I'll just back out this part which causes a regression - i think the rest of the change is good.  We can revisit this bit at a later date.

deadprogram (approved):
Looking good to me @kenbell thanks for the fixes.

aykevl:
Unfortunately, this isn't quite right. See https://github.com/tinygo-org/tinygo/issues/3674.
It is broken for me with a Circuit Playground Bluefruit, while it worked before.

kenbell:
Ughh - i'll revisit and take another pass.  I think the logic to select a serial port needs some significant re-work.  I think it's one that will need to wait for the next release now.  It will need some time in dev to bake.

PR #3644, opened 2023-04-09
fix benchmark logging

dgryski:
The verbosity fix in https://github.com/tinygo-org/tinygo/pull/3625 broke `b.Log()` output.  This fixes the segfault and prints the benchmark output afterwards properly.

### DIFF
--- a/src/testing/benchmark.go
+++ b/src/testing/benchmark.go
@@ -160,6 +160,7 @@ func (b *B) ReportAllocs() {
 // runN runs a single benchmark for the specified number of iterations.
 func (b *B) runN(n int) {
 	b.N = n
+	runtime.GC()
 	b.ResetTimer()
 	b.StartTimer()
 	b.benchFunc(b)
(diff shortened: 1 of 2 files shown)


dgryski:
Benchmarks with logging now looks like:
```
~/go/src/github.com/dgryski/trifles/sshdregex $ tinygo test -bench=. -v
BenchmarkRagel	25813588	        45.78 ns/op
--- BENCH: BenchmarkRagel
    1
    100
    10000
    1000000
    25813588
BenchmarkRegex	 1949697	       601.8 ns/op
BenchmarkFSM  	16752514	        70.50 ns/op
```

src/testing/benchmark.go
```diff
@@ -436,8 +444,9 @@ func (b *B) Run(name string, f func(b *B)) bool {
 	b.hasSub = true
 	sub := &B{
 		common: common{
-			name:  benchName,
-			level: b.level + 1,
+			output: &logger{},
```
rminnich, line 447: this is not something I care about but some people worry when fields are not in alphaetical order ...

dgryski: In this case, `output` is actually the first field in `common`.

src/testing/benchmark.go
```diff
@@ -416,6 +418,12 @@ func (b *B) processBench(ctx *benchContext) {
 				results += "\t" + r.MemString()
 			}
 			fmt.Println(results)
+
+			// Print any benchmark outut
```
deadprogram: `outut`?

dgryski: Fixed.

deadprogram:
Thanks for the fixes @dgryski now merging.

PR #3646, opened 2023-04-10
Handle defers in else blocks (or other out-of-order SSA processing)

### LINKED ISSUE #3643: Issue with defers in elses
From Slack:

```
~/go/src/github.com/u-root/u-root/cmds/core/hexdump $ tinygo test -run='Hex/from_file$' -v
=== RUN   TestHexdump
=== RUN   TestHexdump/hexdump_from_file
opening /var/folders/5b/_hr1d1fd3qn4t9vtfx5p61wm0000gp/T/TestHexdump228065111/000/testfile
returning nil
starting defer chain
defer w
starting defer chain
    --- PASS: TestHexdump/hexdump_from_file (0.00s)
--- PASS: TestHexdump (0.00s)
PASS
ok  	github.com/u-root/u-root/cmds/core/hexdump	0.553s
~/go/src/github.com/u-root/u-root/cmds/core/hexdump $ git diff .
diff --git a/cmds/core/hexdump/hexdump.go b/cmds/core/hexdump/hexdump.go
index d7018ea7..cf2c7156 100644
--- a/cmds/core/hexdump/hexdump.go
+++ b/cmds/core/hexdump/hexdump.go
@@ -31,22 +31,27 @@ func hexdump(filenames []string, reader io.Reader, writer io.Writer) error {
                readers = make([]io.Reader, 0, len(filenames))

                for _, filename := range filenames {
+                       println("opening", filename)
                        f, err := os.Open(filename)
                        if err != nil {
                                return err
                        }
-                       defer f.Close()
+                       defer func() { println("defer file"); f.Close() }()
                        readers = append(readers, f)
                }
        }

        r := io.MultiReader(readers...)
        w := hex.Dumper(writer)
-       defer w.Close()
+       defer func() { println("defer w"); w.Close() }()

        if _, err := io.Copy(w, r); err != nil {
+               println("returning err")
                return err
        }
+       _ = w
+       println("returning nil")
+       defer func() { println("starting defer chain") }()
        return nil
 }
```

We see `starting defer chain` twice but no `defer file`. 
The second `starting defer chain` should be the `defer file` entry; it's like it gets overwritten.

dgryski:
Fixes #3643

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -167,6 +167,8 @@ type builder struct {
 	deferExprFuncs    map[ssa.Value]int
 	selectRecvBuf     map[*ssa.Select]llvm.Value
 	deferBuiltinFuncs map[ssa.Value]deferBuiltin
+	runDefersBlock    []llvm.BasicBlock
+	afterDefersBlock  []llvm.BasicBlock
 }
 
 func newBuilder(c *compilerContext, irbuilder llvm.Builder, f *ssa.Function) *builder {
@@ -1349,6 +1351,15 @@ func (b *builder) createFunction() {
 		}
 	}
 
+	// The rundefers instruction needs to be created after all defer
+	// instructions have been created. Otherwise it won't handle all defer
+	// cases.
+	for i, bb := range b.runDefersBlock {
+		b.SetInsertPointAtEnd(bb)
+		b.createRunDefers()
+		b.CreateBr(b.afterDefersBlock[i])
+	}
+
 	if b.hasDeferFrame() {
 		// Create the landing pad block, where execution continues after a
 		// panic.
@@ -1503,7 +1514,14 @@ func (b *builder) createInstruction(instr ssa.Instruction) {
 			b.CreateRet(retVal)
 		}
 	case *ssa.RunDefers:
-		b.createRunDefers()
+		// Note where we're going to put the rundefers block
+		run := b.insertBasicBlock("rundefers.block")
+		b.CreateBr(run)
+		b.runDefersBlock = append(b.runDefersBlock, run)
+
+		after := b.insertBasicBlock("rundefers.after")
+		b.SetInsertPointAtEnd(after)
+		b.afterDefersBlock = append(b.afterDefersBlock, after)
 	case *ssa.Send:
 		b.createChanSend(instr)
 	case *ssa.Store:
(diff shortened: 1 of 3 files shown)


dgryski:
I feel like we should be able to only render the defer handler once and pass in an integer indicating which basic block to jump back to (basically create a switch indicating the return block).

But I managed to come up with a solution that works as a first pass.

dgryski:
I will add this as a test:
```
package main

//go:noinline
func deferbug(b bool) {
        if !b {
                defer println("defer else func")
        }
}

func main() {
	deferbug(false)
}
```

But try to make it actually fail so we can detect it.

compiler/compiler.go
```diff
@@ -1349,6 +1351,14 @@ func (b *builder) createFunction() {
 		}
 	}
 
+	block := b.GetInsertBlock()
+	for i, bb := range b.runDefersBlock {
+		b.SetInsertPointAtEnd(bb)
+		b.createRunDefers()
+		b.CreateBr(b.afterDefersBlock[i])
+	}
+	b.SetInsertPointAtEnd(block)
```
aykevl: You don't need to stash and restore the insert block. The insert point isn't used at a later point.

compiler/compiler.go
```diff
@@ -1349,6 +1351,14 @@ func (b *builder) createFunction() {
 		}
 	}
 
+	block := b.GetInsertBlock()
+	for i, bb := range b.runDefersBlock {
+		b.SetInsertPointAtEnd(bb)
+		b.createRunDefers()
+		b.CreateBr(b.afterDefersBlock[i])
+	}
```
aykevl, line 1361: Can you add a comment what this does? For example:

> The rundefers instruction needs to be created after all defer instructions have been created. Otherwise it won't handle all defer cases.

aykevl (approved):
This looks correct to me, with some style nits. I'm surprised we didn't hit this bug more often.

> I feel like we should be able to only render the defer handler once and pass in an integer indicating which basic block to jump back to (basically create a switch indicating the return block).

Yeah that would be better. But that is an optimization. The PR here is just a bugfix to fix the current broken behavior.
(With a bit of luck, LLVM merges many of those blocks though).

compiler/compiler.go
dgryski: Fixed.

compiler/compiler.go
dgryski: Fixed.

dgryski:
Rebased to clean up the commits.  I'll merge this when CI passes.

PR #3649, opened 2023-04-12
Support LLVM16

QuLogic:
There was surprisingly little to do, but I was testing with the latest release, and not `dev` yet. I also wasn't sure where to add this to CI since you've dropped a lot of them, so it's only put in CircleCI for now.

Naturally, this depends on https://github.com/tinygo-org/go-llvm/pull/45.

I'm not entirely sure the elaborated typedef change is correct, especially because the only failure I have is a few repeats of:
```
# _/builddir/build/BUILD/tinygo-0.27.0/_build/src/github.com/tinygo-org/tinygo/testdata/cgo
testdata/cgo/main.go:146:48: cannot use (*[7][2]C.int)(nil) (value of type *[7][2]C.int) as [4][7][2]C.int value in argument to C.arraydecay
```
but we'll see how it goes in CI for `dev`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,6 +65,9 @@ commands:
                 lld-<<parameters.llvm>> \
                 cmake \
                 ninja-build
+            # Fix missing symlink (useless for LLVM 14, but harmless.)
+            cd /usr/lib/x86_64-linux-gnu/
+            ln -sf libclang-16.so.16.* libclang-16.so.16
       - hack-ninja-jobs
       - build-binaryen-linux
       - restore_cache:
@@ -105,6 +108,13 @@ jobs:
       - test-linux:
           llvm: "14"
     resource_class: large
+  test-llvm16-go118:
+    docker:
+      - image: golang:1.18-buster
+    steps:
+      - test-linux:
+          llvm: "16"
+    resource_class: large
 
 workflows:
   test-all:
@@ -112,3 +122,4 @@ workflows:
       # This tests our lowest supported versions of Go and LLVM, to make sure at
       # least the smoke tests still pass.
       - test-llvm14-go118
+      - test-llvm16-go118
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ LLD_SRC ?= $(LLVM_PROJECTDIR)/lld
 
 # Try to autodetect LLVM build tools.
 # Versions are listed here in descending priority order.
-LLVM_VERSIONS = 15 14 13 12 11
+LLVM_VERSIONS = 16 15 14 13 12 11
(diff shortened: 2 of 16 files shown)


QuLogic:
I realized there was another case for handling typedefs in `makeDecayingASTType`, so I also handled elaboraed typedefs there, but I was extermely conservative and only unwrapped if the elaborated type was a typedef.

Additionally, it appears that RISCV targets now have a CPU; I've added that in, but without any version checks, so that will probably break something.

QuLogic:
This now passes CI for LLVM16, but because the Circle CI part does not run `go test`, it's actually missing some issues:

- In `cgo` tests, the converted structs now get a translated version of the C name, instead of a generically counted name, i.e., `C._Ctype_struct___0` -> `C.struct_point_t`. Perhaps related to the elaborated types change?
- In `TestCompiler/channel.go`, the `argmemonly` attribute is now `memory(argmem: readwrite)`, but at the end of the list instead. This is the [first note in the LLVM IR release notes](https://releases.llvm.org/16.0.0/docs/ReleaseNotes.html#changes-to-the-llvm-ir)
- In `interp` tests, the pointers seem to have been simplified to `ptr`; it appears that opaque pointers are on by default now, and is probably similar to #3566.

QuLogic:
This now works, AFAICT, at least against 0.27.0 and LLVM 16. I'm not sure which job to modify here to get full LLVM16 testing.

aykevl:
We don't really have full LLVM 16 testing until Espressif releases their rebased LLVM fork.
Testing in CircleCI seems good enough to me for now.

I will look at the changes later.

QuLogic:
Now that I've [succeeded at building 0.27.0](https://koji.fedoraproject.org/koji/buildinfo?buildID=2187817), I've run against `dev` as well, and found one new file that needed changes, so here's a small update to fix that.

aykevl:
I'm looking into this now, in particular I'm working on full support for LLVM 16 (including static linking with LLVM and fixing the CGo issue).

QuLogic:
Rebased with the merged go-llvm commit.

aykevl:
Hi @QuLogic I have made a LLVM 16 PR here: https://github.com/tinygo-org/tinygo/pull/3741
I did it in a separate PR because I prefer switching to a new LLVM version entirely (not off-by-default) to catch more bugs, and because I needed to fix that CGo bug anyway. I hope you don't mind!
One difference that may be important to you is that I don't run all the compiler tests for older LLVM versions: there are frequent IR changes and supporting multiple LLVM versions is just too much hassle for little gain.

deadprogram:
@QuLogic this PR is also included in https://github.com/tinygo-org/tinygo/pull/3741 thank you for your efforts! Now closing.

PR #3654, opened 2023-04-14
Multi msd volume names

kenbell:
PoC: match on multiple possible volume names


aykevl:
Looks good to me.

kenbell:
I tidied up some 'testing' entries in rp2040.json and tweaked mac case to be more consistent with other platforms in the loops.  Assuming this passes CI, should be ready to go.

deadprogram:
Thank you for the addition @kenbell and to @aykevl for review. Now merging.

PR #3657, opened 2023-04-15
main: don't print `ok` for a successful compile-only

dgryski:


aykevl:
This is something I noticed before but hadn't looked into the bug. Thank you for fixing it!

PR #3658, opened 2023-04-15
Switch interp tests to opaque pointers

QuLogic:
Similar to #3566; it is necessary to do this for LLVM 16, as it switched to opaque pointers in output by default, and LLVM 17 will support _only_ opaque pointers.

I'm not 100% certain on the changes in the source files, but it does appear to be passing tests for me at least. I assumed that a) all `bitcast` could be dropped since they'd just be `ptr`->`ptr` anyway, and b) `getelementptr` could be dropped _iff_ the indices were all 0, but since a lot of that input gets optimized away in the output, I didn't have a good reference to confirm these assumptions.


aykevl (approved):

aykevl:
Yes this is a good idea. I have done something similar in #3566.

PR #3661, opened 2023-04-16
Update wasi-libc to SDK 20

### LINKED ISSUE #3910: Update wasi-libc
I noticed wasi-libc is currently from the wasi-sdk 16 release and the latest wasi-sdk is now 20. Would it make sense to update it? I was notably missing the `getpagesize` function that was added in wasi-sdk 17.

QuLogic:
The version 17 SDK adds `getpagesize`, so use it instead of hardcoding a number (even if their implementation is _also_ a hardcoded number.)

### DIFF
--- a/src/syscall/syscall_libc_wasi.go
+++ b/src/syscall/syscall_libc_wasi.go
@@ -398,8 +398,7 @@ func Chmod(path string, mode uint32) (err error) {
 }
 
 func Getpagesize() int {
-	// per upstream
-	return 65536
+	return libc_getpagesize()
 }
 
 type Utsname struct {
@@ -427,6 +426,11 @@ func Syscall(trap, a1, a2, a3 uintptr) (r1, r2 uintptr, err Errno)
 
 // This is a stub, it is not functional.
 func Syscall6(trap, a1, a2, a3, a4, a5, a6 uintptr) (r1, r2 uintptr, err Errno)
+
+// int getpagesize(void);
+//
+//export getpagesize
+func libc_getpagesize() int
 
 // int stat(const char *path, struct stat * buf);
 //
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,10 +75,10 @@ commands:
       - run: go install -tags=llvm<<parameters.llvm>> .
       - restore_cache:
           keys:
-            - wasi-libc-sysroot-systemclang-v6
+            - wasi-libc-sysroot-systemclang-v7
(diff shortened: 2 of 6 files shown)


QuLogic:
Hrm, bit confused since this test works for me locally just fine.

aykevl:
What's the benefit here? The page size is fixed for WebAssembly at 64kB, so I don't see why we need a library call.

(This is different for `getpagesize` in Linux for example, as the page size varies by architecture and sometimes by kernel build - Apple M1 has 16K pages for example. There it really does need to be a library call).

dgryski:
Aside from debating the`getpagesize` change,  seems like upgrading to the latest `wasi-libc` is a Good Thing?

aykevl:
Yes, seems fine by me. I'm a bit careful with updating these libraries because they may introduce new bugs, but we have to do it from time to time.

deadprogram:
@QuLogic could you please rebase this branch against the latest `dev` so we can verify it is passing the latest tests? I would like to merge.

src/syscall/syscall_libc_wasi.go
```diff
@@ -324,8 +324,7 @@ func Pipe2(p []int, flags int) (err error) {
 }
 
 func Getpagesize() int {
-	// per upstream
-	return 65536
+	return libc_getpagesize()
 }
```
aykevl, line 402: EDIT: oops I see I already commented on this.

---

Probably unnecessary, the WebAssembly page size is fixed at 64kB. I don't really see a benefit of calling a libc function. (This is different on an architecture like arm64 which does in fact have multiple page sizes that can only be known at runtime).
On the other hand, it probably won't increase binary size by much either so :shrug:

QuLogic:
Gave it a rebase. I forget which build failed, and I think logs might have gone, so we'll see if it's okay now.

QuLogic:
I just realized there are some `wasi-libc` caches in CI, skipping their build if available, which is probably why CI is failing but not local.

QuLogic:
This should be good to go now.

src/syscall/syscall_libc_wasi.go
QuLogic: Mostly for consistency, and to limit hard-coded numbers. I have no idea whether wasi will evolve in some direction as to not be 64KiB, but if it does, then that information will be encoded in the SDK instead of here.

deadprogram:
Thanks @QuLogic now merging.

PR #3662, opened 2023-04-17
builder: print the status of the job that just completed

dgryski:


deadprogram (approved):
Good fix, thanks @dgryski

PR #3665, opened 2023-04-18
samd51: fix ADC multisampling

aykevl:
Multisampling/averaging (using the Samples configuration property) was returning incorrect values. When I investigated this, I found that the samd51 gives erratic values when using multisampling together with fewer than 16 bits resolution.

I fixed this by forcing 16 bit resolution when multisampling, and adjusting the output to account for multisampling.

Found while reading the battery value on a pybadge, which gave non-sensible values with Samples set to a value larger than 1.


deadprogram:

PR #3670, opened 2023-04-22
device/gba: additional IO mapping for sound, DMA, and SIO

deadprogram:
This PR adds more IO mapping to the device/gba package for sound, DMA, and SIO.


deadprogram:
I also added some of the sprite OBJ support over the weekend. Done for now.

aykevl (approved):
Did not verify this in depth but it looks reasonable. LGTM.

PR #3673, opened 2023-04-24
Print source location for panics when using `-monitor`

aykevl:
This PR adds the ability to print source locations for runtime panics. Here is an example:

    $ tinygo flash -target=circuitplay-bluefruit -monitor ./examples/heartrate
    tick 00:00.810
    tick 00:01.587
    tick 00:02.387
    tick 00:03.244
    panic: runtime error at 0x00027c4d: alloc in interrupt
    [tinygo: panic at /home/ayke/src/tinygo/bluetooth/adapter_sd.go:74:4]

To be clear, this path isn't stored on the microcontroller. It's stored as part of the build, and `-monitor` just looks up the path from the panic message.

Possible enhancements:

  - Print such an address for regular panics as well. I'm not sure that's so useful, as it's usually a lot easier to look up panics just by their message.
  - Use runtimePanicAt (instead of runtimePanic) in other locations, if that proves to be beneficial.
  - Print the TinyGo-generated output in some other color, to distinguish it from the regular console output.
  - Print more details when panicking (registers, stack values), and print an actual backtrace.

### DIFF
--- a/builder/sizes_test.go
+++ b/builder/sizes_test.go
@@ -41,9 +41,9 @@ func TestBinarySize(t *testing.T) {
 	// This is a small number of very diverse targets that we want to test.
 	tests := []sizeTest{
 		// microcontrollers
-		{"hifive1b", "examples/echo", 4556, 272, 0, 2252},
-		{"microbit", "examples/serial", 2680, 380, 8, 2256},
-		{"wioterminal", "examples/pininterrupt", 6109, 1471, 116, 6816},
+		{"hifive1b", "examples/echo", 4612, 276, 0, 2252},
+		{"microbit", "examples/serial", 2724, 384, 8, 2256},
+		{"wioterminal", "examples/pininterrupt", 6159, 1477, 116, 6816},
 
 		// TODO: also check wasm. Right now this is difficult, because
 		// wasm binaries are run through wasm-opt and therefore the
--- a/main.go
+++ b/main.go
@@ -532,7 +532,7 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 		return fmt.Errorf("unknown flash method: %s", flashMethod)
 	}
 	if options.Monitor {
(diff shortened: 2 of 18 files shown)


deadprogram:
@aykevl looks like something failing the various CI tests, maybe a file that was not added?

aykevl:
Fixed, but I can't really test the change because of #3674. At least not on my Circuit Playground Bluefruit.

deadprogram:
Looks like CI still failing.

aykevl:
Fixed, MachO is problematic again and I'm not really interested in fixing that. It's ELF that's important (used on Linux and baremetal).

aykevl:
Breaks on Windows too, so I'll limit testing on Linux only. Which is all we care about, really (we care about the ELF format which isn't used on MacOS or Windows).

deadprogram:
@aykevl can you please resolve merge conflict?

aykevl:
Rebased.

deadprogram:
```
$ tinygo flash -target=circuitplay-bluefruit -monitor ./examples/heartrate                                                                        
tick 00:01.588                                                                                                                                    
tick 00:02.388                                                                                                                                    
tick 00:03.245                                                                                                                                    
tick 00:03.976
...
tick 00:53.848
tick 00:54.771
tick 00:55.512
tick 00:56.323
tick 00:57.123
tick 00:57.902
tick 00:58.811
```

It never seems to fail?

aykevl:
IIRC it only fails when you actually connect to it. You can try connecting using examples/heartrate-monitor for example.

deadprogram:
> it only fails when you actually connect to it. 

Thanks. It works (fails?) as expected.

deadprogram (approved):
Well worth the small increase in size to be able to figure out what is going on!

deadprogram:
Now merging, thanks @aykevl

PR #3675, opened 2023-04-25
Revert "Improve handling of multiple identical MCUs connected via USB"

deadprogram:
Reverts tinygo-org/tinygo#3640


deadprogram:
Reverting until @kenbell can help take a closer look. Sorry everyone about the mess!

PR #3676, opened 2023-04-25
runtime: improve panic message when heap allocating in an interrupt

aykevl:
The old message may have been confusing. With the new message, it should be more clear that we mean a _heap_ allocation.


deadprogram:
Looks like small size change with this PR causing CI to fail.

deadprogram:
@aykevl please resolve merge conflict.

deadprogram:

PR #3678, opened 2023-04-26
all: improve selection of serial ports for flash and monitor

kenbell:
2nd attempt at fixing #3637

This time, I've reworked the code to:
1.  discover ports so that the separation of 'possible list of ports' and 'previously selected port' is clearer.
2.  be explicit about how the code works to identify the 'best' port based on user parameters, VID:PID pairs, etc
3.  made the concept of 'retrying' serial port discovery more explicit

### DIFF
--- a/main.go
+++ b/main.go
@@ -370,7 +370,7 @@ func dirsToModuleRoot(maindir, modroot string) []string {
 }
 
 // Flash builds and flashes the built binary to the given serial port.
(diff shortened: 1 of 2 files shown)


kenbell:
This change needs some time to soak in dev before a release since it affects multiple platforms and is very subject to difference in performance of MCUs and host PCs.

deadprogram:
@kenbell now that we are into the next development cycle, what is the status of this PR?

PR #3680, opened 2023-04-26
cortexm: fix stack size offset

aykevl:
The old code was broken and led to a HardFault in a rather convoluted way:

 1. The CFA offset was incorrect, in fact it was not aligned (the stack is supposed to always be aligned to 4 bytes at least).
 2. This unaligned size was then used for stack size calculations.
 3. A stack that wasn't a multiple of 4 was allocated.
 4. The calleeSavedRegs struct (in `(internal/task.state).archInit`) was not correctly aligned.
 5. Writing to this struct resulted in a HardFault.

### DIFF
--- a/src/internal/task/task_stack_cortexm.S
+++ b/src/internal/task/task_stack_cortexm.S
@@ -114,9 +114,9 @@ tinygo_swapTask:
     pop {r4-r11, pc}
     #else
     pop {r4-r7}
-    .cfi_def_cfa_offset 5*9
+    .cfi_def_cfa_offset 5*4
     pop {r0-r3}
-    .cfi_def_cfa_offset 1*9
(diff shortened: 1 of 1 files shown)


kenbell:
lgtm - value is now changing by 4x the number of registers, which makes some sense.

must have been a fun one to track down

aykevl:
Yeah this one was interesting. It took me a while until I realized it tried to store to an unaligned memory address - and then I went digging from there.

As a coincidence, I was just about to start working on optimizing the hard fault handler to have a smaller impact on code size when I found this bug (and had to fix it first to be able to continue).

deadprogram:
Thanks for the fix @aykevl and to @kenbell for review.

PR #3681, opened 2023-04-26
cortexm: optimize code size for the HardFault_Handler

aykevl:
This function is called when a hard fault occurs. Hard faults happen when something really bad happens - like writing to unwritable memory or an unaligned memory access on Cortex-M0. It is not generally possible to recover from these.

This commit optimizes the code size overhead of hard fault handling:

  * It removes the stack overflow checking code. This may seem like a bad thing, but the only thing this could check were stack overflows outside goroutines. In practice, this could only really happen on a stack overflow in the scheduler (unlikely), or in interrupt code (possible, but interrupts are small so still unlikely). Most stack overflows happen in regular goroutines, and weren't caught in the HardFault.
  * It makes the panic message similar to a regular panic. This has two advantages:
      * It reduces code size, because the string can be reused between the HardFault handler and the runtime panic function.
      * Using the same pattern automatically makes `-monitor` print the source address for the hard fault. Not a big benefit as we could trivially add any other pattern but a nice benefit nonetheless.

Result:

    $ tinygo flash -target=microbit -size=short -programmer=openocd -monitor examples/serial
       3036       8    2256 |    3044    2264
    [...snip]
    panic: runtime error at 0x00000344: HardFault with sp=0x200007d0
    [tinygo: panic at /home/ayke/src/tinygo/tinygo/src/internal/task/task_stack_cortexm.go:48:4]

Previously it would just print the hardfault addresses without source location.

(This is with https://github.com/tinygo-org/tinygo/pull/3680 not yet fixed and some local changes to configure the UART so I can actually see the panic).

For atsamd21/nrf51 chips this results in a binary size reduction of around 100 bytes. For other Cortex-M chips it's around 24 bytes but I hope to change this in the future because a lot of the fault decoding in runtime_cortexm_hardfault_debug.go should IMHO be done by the TinyGo monitor instead (I estimate that this would save around 800 bytes on these chips).

### DIFF
--- a/builder/sizes_test.go
+++ b/builder/sizes_test.go
@@ -43,8 +43,8 @@ func TestBinarySize(t *testing.T) {
 	tests := []sizeTest{
 		// microcontrollers
 		{"hifive1b", "examples/echo", 3884, 280, 0, 2268},
-		{"microbit", "examples/serial", 2924, 388, 8, 2272},
-		{"wioterminal", "examples/pininterrupt", 7365, 1491, 116, 6912},
+		{"microbit", "examples/serial", 2852, 360, 8, 2272},
+		{"wioterminal", "examples/pininterrupt", 7337, 1491, 116, 6912},
 
 		// TODO: also check wasm. Right now this is difficult, because
 		// wasm binaries are run through wasm-opt and therefore the
--- a/src/device/arm/cortexm.S
+++ b/src/device/arm/cortexm.S
@@ -1,29 +1,6 @@
 .syntax unified
 .cfi_sections .debug_frame
 
-.section .text.HardFault_Handler
-.global  HardFault_Handler
-.type    HardFault_Handler, %function
-HardFault_Handler:
-    .cfi_startproc
-    // Put the old stack pointer in the first argument, for easy debugging. This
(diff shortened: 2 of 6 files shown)


deadprogram:
Looks like "real" CI failure here https://app.circleci.com/pipelines/github/tinygo-org/tinygo/9517/workflows/3c85e9d4-e599-4624-828c-400a0ae7ea16/jobs/40570?invite=true#step-118-15

aykevl:
Oh I forgot, this needs a special case for LLVM 14. That, or we need to drop support for LLVM 14.

I think I'd wait with this until we drop support for LLVM 14.

deadprogram:
Hi @aykevl we dropped support for LLVM 14 a while ago. Is this PR still valid?

aykevl:
This still seems like a good idea to me. It reduces code size a little, and removes an unnecessary assembly function (now replaced with `llvm.sponentry.p0()`).

Rebased etc to run CI and make it ready for merge.

aykevl:
Oooh code size changes (see above comment) are looking good! No regressions, and most programs get at least a little bit smaller.

deadprogram:
I've been waiting a while for this @aykevl thanks for getting it finished! Now merging.

PR #3687, opened 2023-04-27
Add impact on binary size as a comment to PRs

aykevl:
Testing this...

### DIFF
--- a/.github/workflows/sizediff.yml
+++ b/.github/workflows/sizediff.yml
@@ -1,0 +1,92 @@
+name: Binary size difference
+
+on:
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  sizediff:
+    runs-on: ubuntu-22.04
+    steps:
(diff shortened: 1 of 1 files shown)


aykevl:
Well, finally, I think this is ready for review.

The comment could be improved quite a bit, which I hope to do in follow-up PRs. This is just basic functionality.

* It can say "there is no difference" when all the hashes are the same (and skip the sizediff output entirely).
* It can say "the size didn't change, but some binaries did" (and maybe show a plain old diff?).
* If there is a difference in binary size, it can provide a short summary, something like "23 out of 112 smoke tests changed in size, of which 20 were reduced in size and 3 increased in size".

sago35:
I have not seen the code at all.
How does it work when the smoke test is added or removed?

aykevl:
> How does it work when the smoke test is added or removed?

The drivers repo is tested, not the smoke tests in the tinygo repo. This repository stays the same when checking out the dev branch.

aykevl:
Can someone take a look at this PR? I think it's very important to catch regressions early.

deadprogram:
Looking good to me! I say we merge it.

PR #3688, opened 2023-04-27
tools: add sizediff tool

aykevl:
This tool can be very useful to compare binary sizes as output by `-size=short`.

This is a tool I wrote a while ago. It's not perfect (we should probably use a geomean) but it works well enough to get a good idea on the binary size impact of a change.

---

I'm adding this in preparation for https://github.com/tinygo-org/tinygo/pull/3687, which will need to have this tool already in the dev branch.

### DIFF
--- a/tools/sizediff
+++ b/tools/sizediff
@@ -1,0 +1,83 @@
+#!/usr/bin/env python3
+
+# Small tool to compare code size between TinyGo runs (with the -size=short flag).
+
+import sys
+
+class Comparison:
+    def __init__(self, command, code0, data0, bss0, code1, data1, bss1):
+        self.command = command
+        self.code0 = code0
+        self.data0 = data0
+        self.bss0 = bss0
+        self.code1 = code1
+        self.data1 = data1
+        self.bss1 = bss1
+
+    @property
+    def flash0(self):
+        return self.code0 + self.data0
+
+    @property
+    def flash1(self):
+        return self.code1 + self.data1
+
+    @property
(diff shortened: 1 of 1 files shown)


dgryski:
My copy has geomean integrated:
It needs
```
  totalProduct = 1
```

```
        if comparison.codediff != 0:
            totalProduct *= (1 + diffPct)
```

and then at the end

```
    geomean = totalProduct ** (1.0 / float(len(comparisons)))
```

The true geomean to print is then `geomean - 1`.

deadprogram:
@dgryski are you suggesting to change the implementation of this PR?

dgryski:
Yes :). I realize it was a very poorly done review.  @aykevl said sizediff should have geomean, which I had previously added to my copy.

aykevl:
I have little experience with geomean, so even if I add it I'm not sure it is correct. I only said it should be added because you said it should be ;)
@dgryski do you have your version published somewhere?
Alternatively, what do you think of merging this, and then you can add geomean in a separate PR?

The way I'm using sizediff is to get a rough idea of size changes. While I do quote the sum at the bottom in PRs, I always look at the rest of the changes to see the patterns (is it a similar percentage everywhere, is it a roughly fixed number of bytes, does it apply only to some architectures, etc). I don't think adding geomean will change that for me.

dgryski:
Sure. Merge your script and I'll submit geomean as a follow-up.

aykevl:
Ok, in that case merging :sparkles:

PR #3656, opened 2023-04-15
add struct size and field offsets to reflect data

dgryski:
An alternate solution to https://github.com/tinygo-org/tinygo/pull/3598

PR #3689, opened 2023-04-27
compiler,reflect: make field offsets varints

### LINKED ISSUE #3686: reflect: make struct field offset a varint
#3656 added struct sizes and field offsets. This adds around 0.36% on average to the binary size according to my test, although the increase is highly variable (and sometimes negative).

I suspect most of the size increase comes from the struct field offsets. One way to reduce the overhead is by not storing the offset as a uint32, but rather as a variable width integer in the `data` byte array. I believe that will save 3 bytes per struct field in the common case.

dgryski:
Fixes #3686

### DIFF
--- a/src/reflect/type.go
+++ b/src/reflect/type.go
@@ -1294,13 +1294,7 @@ func uvarint32(buf []byte) (uint32, int) {
 	var x uint32
 	var s uint
 	for i, b := range buf {
-		if i == maxVarintLen32 {
-			return 0, -(i + 1) // overflow
-		}
 		if b < 0x80 {
-			if i == maxVarintLen32-1 && b > 1 {
-				return 0, -(i + 1) // overflow
-			}
 			return x | uint32(b)<<s, i + 1
 		}
 		x |= uint32(b&0x7f) << s
--- a/src/reflect/type.go
+++ b/src/reflect/type.go
@@ -463,7 +463,6 @@ type structType struct {
 
 type structField struct {
 	fieldType *rawType
-	offset    uint32
 	data      unsafe.Pointer // various bits of information, packed in a byte array
 }
 
@@ -624,7 +623,7 @@ func (t *rawType) Field(i int) StructField {
 	}
 }
 
-func rawStructFieldFromPointer(descriptor *structType, fieldType *rawType, data unsafe.Pointer, flagsByte uint8, name string, offset uintptr) rawStructField {
+func rawStructFieldFromPointer(descriptor *structType, fieldType *rawType, data unsafe.Pointer, flagsByte uint8, name string, offset uint32) rawStructField {
 	// Read the field tag, if there is one.
 	var tag string
 	if flagsByte&structFieldFlagHasTag != 0 {
@@ -651,7 +650,7 @@ func rawStructFieldFromPointer(descriptor *structType, fieldType *rawType, data 
 		Type:      fieldType,
 		Tag:       StructTag(tag),
 		Anonymous: flagsByte&structFieldFlagAnonymous != 0,
-		Offset:    offset,
+		Offset:    uintptr(offset),
 	}
 }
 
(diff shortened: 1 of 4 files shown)


dgryski:
Impact on a production service built with `-target=wasi -no-debug`:

```
-rwxr-xr-x  1 dgryski  staff  2751357 27 Apr 09:25 service.dev*
-rwxr-xr-x  1 dgryski  staff  2750656 27 Apr 09:18 service.field-varint*
```

deadprogram:
@dgryski size changed so CI failing.

aykevl:
That's a much smaller effect than I expected, but I'll test the drivers repo too (hence why I'm working on #3687 to make this easier).

deadprogram:
@aykevl looking good to merge?

aykevl:
I'll try to take a look soon.

aykevl:
Did a quick check using the drivers smoke tests, and this does indeed result in a small binary size reduction (overall reduction of 0.21%) with a few exceptions where it increases binary size. So from that perspective, this PR looks good. I'll look at the implementation next.

(When #3687 is in, checking this will become a ton easier as I don't have to run these tests manually).

EDIT: 0.21% is actually really good, as #3656 added added 0.36% in binary size. That means that the sum increase is only 0.15% - less than half of the original increase.

src/reflect/type.go
```diff
@@ -1284,3 +1286,25 @@ func StructOf([]StructField) Type {
...
+	for i, b := range buf {
+		if i == maxVarintLen32 {
+			return 0, -(i + 1) // overflow
+		}
+		if b < 0x80 {
+			if i == maxVarintLen32-1 && b > 1 {
+				return 0, -(i + 1) // overflow
+			}
+			return x | uint32(b)<<s, i + 1
+		}
```
aykevl: We know that the input data is correct, so we don't need to check for overflow. I suggest removing those checks to reduce binary size.

src/runtime/interface.go
```diff
@@ -96,8 +96,7 @@ func interfaceTypeAssert(ok bool) {
 
 type structField struct {
 	typecode unsafe.Pointer // type of this struct field
-	offset   uint32
-	data     *uint8 // pointer to byte array containing name, tag, and 'embedded' flag
+	data     *uint8         // pointer to byte array containing name, tag, and 'embedded' flag
```
aykevl: This comment needs a slight update to also include the offset.

src/reflect/type.go
```diff
@@ -673,13 +672,14 @@ func (t *rawType) rawField(n int) rawStructField {
...
 	field := (*structField)(unsafe.Add(unsafe.Pointer(&descriptor.fields[0]), uintptr(n)*unsafe.Sizeof(structField{})))
-	offset := uintptr(field.offset)
 	data := field.data
 
 	// Read some flags of this field, like whether the field is an embedded
 	// field. See structFieldFlagAnonymous and similar flags.
 	flagsByte := *(*byte)(data)
 	data = unsafe.Add(data, 1)
+	offset, lenOffs := uvarint32(unsafe.Slice((*byte)(data), maxVarintLen32))
+	data = unsafe.Add(data, lenOffs)
```
aykevl, line 682: One thing to consider (although that might go a bit too far) is to not convert to a slice but work on raw pointers instead. That's a bit more dangerous, but may reduce binary size a little bit.

aykevl (commented):
Looks good to me, with some suggestions how the implementation could be slightly more efficient.

deadprogram:
@dgryski not sure if you saw @aykevl comments on this PR, but looked like pretty small changes then we can merge this one.

dgryski:
PTAL

deadprogram:
The size reduction is working as expected, and all the immediate feedback has been addressed, so now merging. Thank you very much for working on this @dgryski and for the helpful review @aykevl

PR #3692, opened 2023-04-28
machine/usb: move MIDI under ADC

deadprogram:
This PR moves the USB MIDI implementation under a subpackage `adc` (Audio Device Class) for easier future expansion.

It also adds some minor docs.

Lastly, it corrects `examples/usb-midi` so that it does not use any serial communication other than MIDI of course.


src/examples/usb-midi/main.go
```diff
@@ -16,12 +15,11 @@ func main() {
 	led.Configure(machine.PinConfig{Mode: machine.PinOutput})
 
 	button := machine.BUTTON
-	button.Configure(machine.PinConfig{Mode: machine.PinInputPullup})
+	button.Configure(machine.PinConfig{Mode: machine.PinInputPulldown})
```
sago35, line 18: I think there is an optimal pullup/pulldown choice for each target, but it would be confusing to use pulldown only for this example.
I hope we can find some way to make it better.

sago35 (commented):
Looks good, but I would like to think about making it Pulldown as it seems a bit confusing.

src/examples/usb-midi/main.go
sago35:
```
$ git rev-parse HEAD
d28b58e4dd31eee1c7d188a3ac0139f3477be099

$ git grep -i pininput ./src/examples/
src/examples/button/button.go:  button.Configure(machine.PinConfig{Mode: machine.PinInputPullup})
src/examples/button2/button2.go:        button1.Configure(machine.PinConfig{Mode: machine.PinInputPullup})
src/examples/button2/button2.go:        button2.Configure(machine.PinConfig{Mode: machine.PinInputPullup})
src/examples/button2/button2.go:        button3.Configure(machine.PinConfig{Mode: machine.PinInputPullup})
src/examples/button2/button2.go:        button4.Configure(machine.PinConfig{Mode: machine.PinInputPullup})
src/examples/hid-keyboard/main.go:      button.Configure(machine.PinConfig{Mode: machine.PinInputPullup})
src/examples/hid-mouse/main.go: button.Configure(machine.PinConfig{Mode: machine.PinInputPullup})
src/examples/pdm/pdm.go:        machine.BUTTONA.Configure(machine.PinConfig{Mode: machine.PinInputPulldown})
src/examples/pininterrupt/circuitplay-express.go:       buttonMode      = machine.PinInputPulldown
src/examples/pininterrupt/pca10040.go:  buttonMode      = machine.PinInputPullup
src/examples/pininterrupt/stm32.go:     buttonMode      = machine.PinInputPulldown
src/examples/pininterrupt/wioterminal.go:       buttonMode      = machine.PinInput
src/examples/usb-midi/main.go:  button.Configure(machine.PinConfig{Mode: machine.PinInputPullup})
```

sago35: It may be possible to create a default setting for `machine.BUTTON`, I do not know.

deadprogram: The button question is a bit beyond the scope of this PR, I think. No?

sago35: Yes, it is.
So I think it is better not to change it in this PR.
However, since there is no direct problem, either is fine.

sago35 (approved):

deadprogram:
Thanks for the review and feedback, @sago35 now merging.

PR #3695, opened 2023-04-30
wasm: remove i64 workaround

aykevl:
This is my third try. I think now (well, after the coming TinyGo release) is the right time to finally merge this change.

For the history, see: https://github.com/tinygo-org/tinygo/pull/2375

This PR removes around 300 lines of code, which I think is a big win.


aykevl:
Tested the PR in the same way as before, and it passes:

> I have tested this PR with the following steps:
> 
>    1. (before this PR) Modify the syscall/js package to comment out all tests that don't pass yet, until `tinygo test -target=wasm syscall/js` works.
>    2. Remove the i64 workaround.
>    3. Fix wasm_exec.js until `tinygo test -target=wasm syscall/js` passes again.

dgryski:
Can we merge this now?

aykevl:
Yes, this seems like a good time to do it.

deadprogram:
Merging thank you @aykevl for the improvement and to @dgryski for review.

PR #3698, opened 2023-05-01
os: implement os.(*File).Sync

achille-roussel:
This PR implements `os.(*File).Sync`. The use case I am targeting is WASI but the implementation also touches Darwin and other targets using libc.


dgryski (approved):
LGTM

deadprogram:
Thank you for working on this @achille-roussel and to @dgryski for review. Now merging.

PR #3712, opened 2023-05-06
builder: implement Nordic DFU file writer in Go

aykevl:
This avoids a dependency on nrfutil. I have verified that it creates equivalent zip files to a wasp-os DFU zip file I downloaded here: https://github.com/wasp-os/wasp-os/releases/
I have also tested that it produces valid DFU files that can be uploaded using the dfu.py program here to my PineTime:
https://github.com/wasp-os/ota-dfu-python/tree/3d6fd30d33c2b20bc86ff6b9269fddf4a1d4c7c6

There are some minor differences in the generated file that should not matter in practice (JSON whitespace, firmware file name, zip compression).

---

The reason I did this is because nrfutil (downloaded from the Nordic website) does not work well on my linux/arm64 laptop.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -909,13 +909,8 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 		}
 	case "nrf-dfu":
 		// special format for nrfutil for Nordic chips
-		tmphexpath := filepath.Join(tmpdir, "main.hex")
-		err := objcopy(result.Executable, tmphexpath, "hex")
-		if err != nil {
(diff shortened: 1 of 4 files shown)


aykevl:
Hmm, looks like this needs a small change for Go 1.18 compatibility.

aykevl:
This should fix the Go 1.18 compatibility.

deadprogram (approved):

deadprogram:
Looks good, and removing an external dependency is :+1: 

Thanks @aykevl now merging.

PR #3714, opened 2023-05-06
machine/flash: improvements and fixes to FlashBuffer Write() and Seek()

deadprogram:
This PR has improvements and fixes to the `FlashBuffer` `Write()` and `Seek()` functions.

It also updates the example to show that you can perform multiple `Write()` operations, as long as the data is a multiple of the write block size.

### DIFF
--- a/src/machine/flash.go
+++ b/src/machine/flash.go
@@ -35,6 +35,7 @@ var (
 	errFlashCannotReadPastEOF   = errors.New("cannot read beyond end of flash data")
 	errFlashCannotWritePastEOF  = errors.New("cannot write beyond end of flash data")
 	errFlashCannotErasePastEOF  = errors.New("cannot erase beyond end of flash data")
+	errFlashInvalidSeekOffset   = errors.New("cannot seek outside address space of flash data")
 )
 
 // BlockDevice is the raw device that is meant to store flash data.
@@ -74,6 +75,9 @@ type FlashBuffer struct {
 
 	// offset is relative to start
 	offset uintptr
+
+	// erased is the last consecutive address to have been erased
+	erased uintptr
 }
 
 // OpenFlashBuffer opens a FlashBuffer.
@@ -89,38 +93,50 @@ func (fl *FlashBuffer) Read(p []byte) (n int, err error) {
 	return
 }
 
-// Write data to a FlashBuffer.
+// Write data to a FlashBuffer. You must pass in a slice containing a multiple
+// of the WriteBlockSize() for the target device, or the data written will be
+// padded to fill in any missing space.
 func (fl *FlashBuffer) Write(p []byte) (n int, err error) {
-	// any new pages needed?
+	// any new blocks needed?
 	// NOTE probably will not work as expected if you try to write over page boundary
 	// of pages with different sizes.
-	pagesize := uintptr(fl.b.EraseBlockSize())
+	blocksize := uintptr(fl.b.EraseBlockSize())
 
 	// calculate currentPageBlock relative to fl.start, meaning that
 	// block 0 -> fl.start
-	// block 1 -> fl.start + pagesize
-	// block 2 -> fl.start + pagesize*2
+	// block 1 -> fl.start + blocksize
+	// block 2 -> fl.start + blocksize*2
 	// ...
-	currentPageBlock := (fl.start + fl.offset - FlashDataStart()) + (pagesize-1)/pagesize
-	lastPageBlockNeeded := (fl.start + fl.offset + uintptr(len(p)) - FlashDataStart()) + (pagesize-1)/pagesize
+	currentPageBlock := (fl.offset + (blocksize - 1)) / blocksize
+	lastPageBlockNeeded := (fl.offset + uintptr(len(p)) + (blocksize - 1)) / blocksize
+
+	if fl.erased == 0 {
+		// since we are writing for the first time, we need to
+		// erase the first page e.g. page 0
+		currentPageBlock = 0
+		lastPageBlockNeeded = 1
+	}
 
 	// erase enough blocks to hold the data
-	if err := fl.b.EraseBlocks(int64(currentPageBlock), int64(lastPageBlockNeeded-currentPageBlock)); err != nil {
-		return 0, err
+	if lastPageBlockNeeded*blocksize > fl.erased {
+		if err := fl.b.EraseBlocks(int64(currentPageBlock), int64(lastPageBlockNeeded-currentPageBlock)); err != nil {
+			return 0, err
+		}
+		fl.erased = lastPageBlockNeeded * blocksize
 	}
 
 	// write the data
-	for i := 0; i < len(p); i += int(pagesize) {
-		var last int = i + int(pagesize)
-		if i+int(pagesize) > len(p) {
+	for i := 0; i < len(p); i += int(fl.b.WriteBlockSize()) {
+		var last int = i + int(fl.b.WriteBlockSize())
+		if last > len(p) {
 			last = len(p)
 		}
 
-		_, err := fl.b.WriteAt(p[i:last], int64(fl.offset))
+		n, err := fl.b.WriteAt(p[i:last], int64(fl.offset))
 		if err != nil {
 			return 0, err
 		}
-		fl.offset += uintptr(len(p[i:last]))
+		fl.offset += uintptr(n)
 	}
 
 	return len(p), nil
@@ -133,10 +149,14 @@ func (fl *FlashBuffer) Close() error {
 
 // Seek implements io.Seeker interface, but with limitations.
 // You can only seek relative to the start.
-// Also, you cannot use seek before write operations, only read.
+// Also, with write operations you can only seek back to the beginning.
 func (fl *FlashBuffer) Seek(offset int64, whence int) (int64, error) {
 	if whence != io.SeekStart {
 		panic("you can only Seek relative to Start")
+	}
+
+	if offset < 0 || uintptr(offset) >= FlashDataEnd() {
+		return -1, errFlashInvalidSeekOffset
 	}
 
 	fl.offset = uintptr(offset)
--- a/src/examples/flash/main.go
+++ b/src/examples/flash/main.go
@@ -7,7 +7,10 @@ import (
 
 var (
 	err     error
-	message = "1234567887654321123456788765432112345678876543211234567887654321"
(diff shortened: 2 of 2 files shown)


deadprogram:
@sago35 I think this PR should address your issues, with the buffer size limitations mentioned above.

deadprogram:
```
$ tinygo monitor
Flash data start:       0x10003000
Flash data end:         0x10200000
Flash data size, bytes: 2084864
Flash write block size: 256
Flash erase block size: 4096

Reading data from flash: 
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
Writing data to flash: 
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
Reading data back from flash: 
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
1234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321123456788765432112345678876543211234567887654321
Done.
```

src/machine/flash.go
```diff
@@ -133,12 +149,16 @@ func (fl *FlashBuffer) Close() error {
...
 // Seek implements io.Seeker interface, but with limitations.
 // You can only seek relative to the start.
-// Also, you cannot use seek before write operations, only read.
+// Also, with write operations you can only seek back to the beginning.
 func (fl *FlashBuffer) Seek(offset int64, whence int) (int64, error) {
 	if whence != io.SeekStart {
 		panic("you can only Seek relative to Start")
 	}
 
+	if offset < 0 || uintptr(offset) >= FlashDataEnd() {
```
sago35, line 158: Perhaps ` offset < End - Start` is the correct value.

src/machine/flash.go
```diff
@@ -89,38 +93,50 @@ func (fl *FlashBuffer) Read(p []byte) (n int, err error) {
...
 	// block 0 -> fl.start
-	// block 1 -> fl.start + pagesize
-	// block 2 -> fl.start + pagesize*2
+	// block 1 -> fl.start + blocksize
+	// block 2 -> fl.start + blocksize*2
 	// ...
-	currentPageBlock := (fl.start + fl.offset - FlashDataStart()) + (pagesize-1)/pagesize
-	lastPageBlockNeeded := (fl.start + fl.offset + uintptr(len(p)) - FlashDataStart()) + (pagesize-1)/pagesize
+	currentPageBlock := (fl.offset + (blocksize - 1)) / blocksize
+	lastPageBlockNeeded := (fl.offset + uintptr(len(p)) + (blocksize - 1)) / blocksize
```
sago35, line 111: Why add (blocksize-1) here?
If fl.offset is 0, the correct value 0 is returned; if fl.offset is 1, 1 is returned. But isn't it correct to return 0 even when fl.offset is 1?

src/machine/flash.go
```diff
@@ -89,38 +93,50 @@ func (fl *FlashBuffer) Read(p []byte) (n int, err error) {
...
+	}
 
 	// erase enough blocks to hold the data
-	if err := fl.b.EraseBlocks(int64(currentPageBlock), int64(lastPageBlockNeeded-currentPageBlock)); err != nil {
-		return 0, err
+	if lastPageBlockNeeded*blocksize > fl.erased {
+		if err := fl.b.EraseBlocks(int64(currentPageBlock), int64(lastPageBlockNeeded-currentPageBlock)); err != nil {
+			return 0, err
+		}
+		fl.erased = lastPageBlockNeeded * blocksize
```
sago35, line 125: I believe this approach will not work.
For example, when using the filesystem, the areas are not used in order from the beginning to the end.
It is also very confusing in combination with Seek.

I guess the only way is to read, erase, rewrite, and then write.

deadprogram: > I guess the only way is to read, erase, rewrite, and then write.

I think that is probably true when using the `FlashBuffer`. It is just a shortcut? Perhaps it is not really complete enough as @aykevl suggests we should remove it.

For a real file system, try out https://github.com/tinygo-org/tinyfs/tree/dev which uses the `BlockDevice` interface.

sago35: If so, how about the following?
`EraseBlocks()` are a little more difficult to handle.
And even if not necessarily using tinyfs, there may be cases where the user wants to save a little data.

* Write does not delete, but simply writes
* Read simply reads
* Create a function that returns the block number contained in the specified address to assist EraseBlock
    * ex) func XXXXX(addr uintptr) (blockNo, startAddr)

deadprogram:
See https://github.com/tinygo-org/tinygo/pull/3718 for a pure `BlockDevice` implementation.

deadprogram:
Closing since replaced by #3718

Issue #3713, opened 2023-05-06
With change #3693, examples/hid-keyboard no longer works


sago35:
#3693 

I have checked below and perhaps enumeration is not working properly.

```
$ tinygo flash --target feather-rp2040 --size short --programmer cmsis-dap ./src/examples/hid-keyboard
```

sago35:
Fixed by #3715.

PR #3715, opened 2023-05-06
machine/usb/hid: fix hidreport

sago35:
#3713

Below is a table showing the USB HID Report Descriptor items where the value changes depending on whether it is above or below 255, as well as the corresponding values.

| Item |	Range 0-255	| Range 256 and above|
| --- | -- | -- |
|Logical Minimum|	0x15|	0x16| 
|Logical Maximum	|0x25	|0x26|
|Physical Minimum	|0x35	|0x36|
|Physical Maximum	|0x45	|0x46|
|Usage Minimum	|0x19	|0x1A|
|Usage Maximum	|0x29	|0x2A|


deadprogram:

PR #3716, opened 2023-05-07
pinetime: update the target file

aykevl:
* Rename pinetime-devkit0 to pinetime because the production device is almost the same hardware (the only noticeable difference is a different accelerometer, which isn't part of the board file).
  * Remove the UART and set serial to none. The UART uses a lot of current by default, so it seems better to disable it.

This is a breaking change, but honestly I think I'm the only one who has ever actually used TinyGo for the PineTime and I'm fine with this change :slightly_smiling_face:


deadprogram:
With a pitch like that, how can we not merge this! :laughing:

PR #3717, opened 2023-05-09
nrf52: improve ADC support

aykevl:
There are various settings that weren't used, and that can be quite useful.
I'm using these new settings to correctly read the battery voltage on the PineTime.


deadprogram:
@aykevl these are some pretty interesting changes. We should go back and take a look at the other nrf chips and see if we can use anything from this, if you haven't already.

aykevl:
This is for all nrf52xxx chips, not just the nrf52832. Which means the only chip it _doesn't_ apply to is the nrf51822 (used in the original micro:bit).

The nrf51822 looks like it has slightly different register names and such.

PR #3718, opened 2023-05-09
machine/flash: remove FlashBuffer, modify flash example to use BlockDevice interface

deadprogram:
This PR modifies the machine/flash implementation by removing `FlashBuffer` since it may be more confusing than helpful.

Also modifies the flash example to use `BlockDevice` interface, which is a bit clearer to read now.


sago35 (approved):
LGTM

deadprogram:
Thanks for the review @sago35 and for helping work thru the design process on this. Likewise to @aykevl and others.

Now merging.

Issue #3719, opened 2023-05-10
Failure to initialize variables


sago35:
Both of the following codes should return the same value, but CDCHID is not successfully initialized.

```go
package main

import (
	"fmt"
	"machine/usb/descriptor"
)

func main() {
	fmt.Printf("%#v\n", descriptor.CDCHID.HID[2][116:119])
	fmt.Printf("%#v\n", descriptor.HIDUsageMaximum(0x1FFF))
}
```

```
$ tinygo run ./_x/binary/
[]byte{0x2a, 0x0, 0x0}
[]byte{0x2a, 0xff, 0x1f}

$ tinygo version
tinygo version 0.28.0-dev-14fed598 windows/amd64 (using go version go1.20.2 and LLVM version 15.0.0)
```

deadprogram:
@sago35 was this issue resolved?

sago35:
I have checked, and it seems there are no issues now. @deadprogram 

```go
package main

import (
	"fmt"
	"machine/usb/descriptor"
)

func main() {
	for i, b := range descriptor.CDCHID.HID[2][:] {
		fmt.Printf("%03d %02X\n", i, b)
	}
	fmt.Printf("%#v\n", descriptor.CDCHID.HID[2][135:138])
	fmt.Printf("%#v\n", descriptor.HIDUsageMaximum(0x1FFF))
}
```

output:

```
...
[]byte{0x2a, 0xff, 0x1f}
[]byte{0x2a, 0xff, 0x1f}
```

PR #3720, opened 2023-05-10
machine/usb/hid: fix hidreport (2)

sago35:
MediaKey initialization is failing due to issue #3719.
I changed the code a little to make it work.

### DIFF
--- a/src/machine/usb/descriptor/hidreport.go
+++ b/src/machine/usb/descriptor/hidreport.go
@@ -1,9 +1,5 @@
 package descriptor
 
-import (
-	"encoding/binary"
-)
-
 const (
 	hidUsagePage       = 0x05
 	hidUsage           = 0x09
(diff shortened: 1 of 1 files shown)


aykevl (commented):
Huh, that's odd, I don't see why the code would have changed.
Any idea what causes this bug? (I'm wary of changes that are just "to make things work" without a good explanation as to why they work).

---

Also, it looks like CI is failing because the musl submodule fails.

sago35:
@aykevl 
I have no idea, but I suspect that the code in the following state is not working? I am thinking that it is not working.

* file scope variable
* PutUint16() in the initialization binary.LittleEndian.
    * Maybe there is a little more to it than that, I don't know what it is.

sago35:
> I'm wary of changes that are just "to make things work" without a good explanation as to why they work

I have the same idea.
However, if this is not working well, my project will not work well.

deadprogram:
I prefer the simpler code in this case anyhow. But we should investigate further.

deadprogram:
While we investigate, I am going to merge this to avoid breaking @sago35 other code. Thanks for the fix!

PR #3357, opened 2022-12-27
Filter target build-tags if user specified an overriding option

anuraaga:
I noticed that when building for wasi, the `runtime_memhash_` option cannot be overridden because the leveldb option always gets added, but only one file is allowed to be compiled

```
❯ ~/go/bin/tinygo build -scheduler=none -target=wasi -opt=2 -o mapbench.wasm -tags="runtime_memhash_tsip" .
# runtime
/Users/anuraag/git/tinygo/src/runtime/memhash_tsip.go:18:6: ptrToSlice redeclared in this block
/Users/anuraag/git/tinygo/src/runtime/memhash_leveldb.go:16:6: 	other declaration of ptrToSlice
/Users/anuraag/git/tinygo/src/runtime/memhash_tsip.go:47:6: hash64 redeclared in this block
/Users/anuraag/git/tinygo/src/runtime/memhash_leveldb.go:82:6: 	other declaration of hash64
/Users/anuraag/git/tinygo/src/runtime/memhash_tsip.go:119:6: hash32 redeclared in this block
/Users/anuraag/git/tinygo/src/runtime/memhash_leveldb.go:34:6: 	other declaration of hash32
```

PR #3722, opened 2023-05-13
compileopts: don't filter build tags, use specific build tags instead

aykevl:
This basically reverts https://github.com/tinygo-org/tinygo/pull/3357 and replaces it with a different mechanism to get to the same goal.

I do not think filtering tags like this is a good idea: it's the wrong part of the compiler to be concerned with such tags (that part sets tags, but doesn't modify existing tags). Instead, I've written the `//go:build` lines in such a way that it has the same effect: WASI defaults to leveldb, everything else defaults to fnv, and it's possible to override the default using build tags.

I tested the correct tag gets applied by adding code like this in each memhash file and look at the compiler error:

```go
var x fnv
```

Example result:

```
$ tinygo build -o test.wasm -target=wasi -tags=runtime_memhash_tsip ./testdata/alias.go
# runtime
/home/ayke/src/tinygo/tinygo/src/runtime/memhash_tsip.go:18:7: undefined: tsip
```


dgryski (approved):
LGTM

deadprogram:
Thanks for working on this @aykevl and to @dgryski for the review. Now merging.

PR #3723, opened 2023-05-13
reflect: optimize Zero() a little bit

aykevl:
It could be expensive to call Size() three times, and it is unnecessary. Instead, do it only once.

This results in a very small reduction of binary size if Zero() is used.

---

This is a trivial optimization I found while working on a changelog for the next release.


deadprogram (approved):
Every little bit helps!

PR #3724, opened 2023-05-13
windows: re-enable parallelism

aykevl:
This reverts https://github.com/tinygo-org/tinygo/pull/3525, because that change didn't seem to stop the CI failures we have been seeing. Instead, I've added thread support in https://github.com/tinygo-org/tinygo/pull/3130 which IIRC fixed most of the CI crashes.

Re-enabling parallelism should improve the performance of TinyGo a bit on Windows.

(Not sure whether it's better to include this in the coming release to avoid a performance regression, or to push it to the next release to avoid potentially introducing a bug in the release).

### DIFF
--- a/main.go
+++ b/main.go
@@ -1547,15 +1547,6 @@ func main() {
 		defer pprof.StopCPUProfile()
 	}
 
-	// Limit the number of threads to one.
-	// This is an attempted workaround for the crashes we're seeing in CI on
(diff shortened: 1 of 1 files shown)


deadprogram:
I think we should re-enable it. Any other opinions here?

deadprogram (approved):

deadprogram:
OK merging. Let's go! :twisted_rightwards_arrows:

PR #3725, opened 2023-05-14
device: add support for Renesas family of MCU

deadprogram:
This PR is work in progress to add device support for the Renesas R7FA family of MCUs in preparation to add machine package support for some boards based on these processors.


deadprogram:
I think we should merge this PR now, and then start the process of adding various boards based on this family of processors such as the new Arduino boards in a series of separate PRs.

sago35 (approved):
LGTM
At least there is no change in src/device so far.

deadprogram:
Thanks for review for this first step @sago35 now merging.

PR #3728, opened 2023-05-16
machine: add DefaultUART to xiao

sago35:
Activate the -serial option for tinygo.

This is the xiao version of #2626


deadprogram (approved):

deadprogram:
Thanks for the improvement @sago35 now merging.

PR #3729, opened 2023-05-16
tools: use geomean for sizediff

dgryski:


aykevl:
I can't say I fully understand what this does (I honestly don't know much about statistics), but it seems fine by me. If you've tested this locally I'd be fine merging this even without CI.

dgryski:
Yeah, I'm pretty confident this calculation is correct.  It's the same calculation used for the geometric mean of investment returns (which handles combining percentages). 
```
Geometric Mean Formula for Investments
Geometric Mean = [Product of (1 + Rn)] ^ (1/n) -1
Where:
Rn = growth rate for year N

5th Square Root of ((1 + 0.05)(1 + 0.1)(1 + 0.2)(1 – 0.5)(1 + 0.2)) – 1 = -0.03621
Multiply the result by 100 to calculate the percentage. This results in a -3.62% annual return.
```

soypat:
After a cursory review and a brush up on geomean, looks good to me- at least the implementation of the formula!

deadprogram:
Thanks for working on this @dgryski and to @aykevl and @soypat for review. Now merging.

PR #3734, opened 2023-05-19
Restrict `//go:wasmimport` to a limited number of types

aykevl:
This is done upstream, and I think we should do the same so that people won't rely on unsupported `//go:wasmimport` pragmas.

I'd like to get this in before the next release, so that we don't ship a TinyGo release with a `//go:wasimport` that allows too many parameter types.

### DIFF
--- a/compiler/symbol.go
+++ b/compiler/symbol.go
@@ -239,12 +239,16 @@ func (c *compilerContext) getFunction(fn *ssa.Function) (llvm.Type, llvm.Value) 
 // present in *ssa.Function, such as the link name and whether it should be
 // exported.
 func (c *compilerContext) getFunctionInfo(f *ssa.Function) functionInfo {
+	if info, ok := c.functionInfos[f]; ok {
+		return info
+	}
 	info := functionInfo{
 		// Pick the default linkName.
 		linkName: f.RelString(nil),
 	}
 	// Check for //go: pragmas, which may change the link name (among others).
 	info.parsePragmas(f)
+	c.functionInfos[f] = info
 	return info
 }
 
--- a/compiler/symbol.go
+++ b/compiler/symbol.go
@@ -4,6 +4,7 @@ package compiler
 // pragmas, determines the link name, etc.
 
 import (
+	"fmt"
 	"go/ast"
 	"go/token"
 	"go/types"
@@ -247,14 +248,14 @@ func (c *compilerContext) getFunctionInfo(f *ssa.Function) functionInfo {
 		linkName: f.RelString(nil),
 	}
 	// Check for //go: pragmas, which may change the link name (among others).
-	info.parsePragmas(f)
+	c.parsePragmas(&info, f)
 	c.functionInfos[f] = info
 	return info
 }
 
 // parsePragmas is used by getFunctionInfo to parse function pragmas such as
 // //export or //go:noinline.
-func (info *functionInfo) parsePragmas(f *ssa.Function) {
+func (c *compilerContext) parsePragmas(info *functionInfo, f *ssa.Function) {
 	if f.Syntax() == nil {
 		return
 	}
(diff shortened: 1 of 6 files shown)


deadprogram:
@aykevl looks like failing test on Windows related to this change https://github.com/tinygo-org/tinygo/actions/runs/5024866881/jobs/9011086657?pr=3734#step:17:10

aykevl:
Should be fixed now.

deadprogram:
SInce this implements what was accepted in https://github.com/golang/go/issues/59149 seems correct to me. 

Any comments from @codefromthecrypt @achille-roussel @dgryski or anyone else WASM-oriented before we merge this?

compiler/symbol.go
```diff
@@ -354,6 +361,44 @@ func (info *functionInfo) parsePragmas(f *ssa.Function) {
...
+				allowedType = true
+			case types.UnsafePointer:
+				allowedType = true
+			}
+		}
+		if !allowedType {
+			c.addError(param.Pos(), fmt.Sprintf("%s: unsupported parameter type %s", pragma, param.Type().String()))
+		}
+	}
+}
```
achille-roussel: In Go, we also validate the return values (see https://github.com/golang/go/blob/86c6b4763ed486f20cf018d0810cfd7a1fd91998/src/cmd/compile/internal/ssagen/abi.go#L392-L415)

Note the key difference is we don't allow returning pointers, only integer/floats.

aykevl: Thanks, I missed return values! I've updated the PR accordingly.

Also thanks for mentioning pointers, because I almost missed those.

codefromthecrypt: @achille-roussel is who I would ask, so I approve indirectly ;)

codefromthecrypt: but heh, I should mention that some signatures are using unsafe pointers. There was that chat about nil vs zero recently. So, are we saying that folks should coerce to uintptr(0) in the rare case the signature accepted an unsafe pointer? https://github.com/tinygo-org/tinygo/blob/0d56dee00f49bd50eb373c02c30062a75ec28f10/src/runtime/arch_tinygowasm_malloc.go#L43

achille-roussel: uintptr isn't supported actually because unlike the rest of the industry, GOARCH=wasm uses 64 bits pointers so we didn't want to risk weird bugs by implicitly truncating values when crossing import boundaries. 

I guess the answer to "how do I return a pointer to a region that the host allocated in the linear memory?" is to return a uint32. It's not great but we decided it would be better to push that complexity to the code rather than pretending it can be handled safely by the compiler. 

Baby steps :)

codefromthecrypt: fwiw this works for me. At first, I found pointer types semantically helpful, but from a mapping perspective disturbing. I didn't realize GOARCH=wasm uses 64 bits pointers either! I kindof assumed it went with wasm 1 and 2 which don't support large memory.

deadprogram:
Thank you very much for the improvement @aykevl and to @codefromthecrypt and @achille-roussel for the review.

Now merging, thanks again!

compiler/symbol.go
aykevl: > but heh, I should mention that some signatures are using unsafe pointers.

The example you give is from the runtime, which is special-cased. The restrictions on `//go:wasmimport` don't apply to the runtime.

> uintptr isn't supported actually because unlike the rest of the industry, GOARCH=wasm uses 64 bits pointers so we didn't want to risk weird bugs by implicitly truncating values when crossing import boundaries.

...and TinyGo uses 32-bit `uintptr` (because I didn't check when first adding support for wasm, I just assumed it would be 32-bits). Not allowing `uintptr` in `//go:wasmimport` does in fact avoid potential issues there.

codefromthecrypt: Thanks for the clarification!

PR #3735, opened 2023-05-20
Add Settings to debug.BuildInfo

rajatjindal:
While using `tinygo` to compile one of my projects, I ran into a problem because, in `tinygo`, `debug.BuildInfo` does not have the `Settings []BuildSetting` field like in `golang` [here](https://github.com/golang/go/blob/aeb0644bd33e67f24e2411a651ac9ff72ddc96b4/src/runtime/debug/mod.go#L41).

I made this change locally, compiled `tinygo` and then I was able to compile my project successfully. 

Kindly let me know if this makes sense or if I need to make other changes to fix it.

(also when running the `make test` locally (on M2 laptop), I ran into some llvm problem, so submitting the PR to see if I can validate this using CI here).

Many thanks

### DIFF
--- a/src/runtime/debug/debug.go
+++ b/src/runtime/debug/debug.go
@@ -33,9 +33,17 @@ func ReadBuildInfo() (info *BuildInfo, ok bool) {
 // BuildInfo represents the build information read from
 // the running binary.
 type BuildInfo struct {
-	Path string    // The main package path
-	Main Module    // The module containing the main package
-	Deps []*Module // Module dependencies
+	Path     string    // The main package path
+	Main     Module    // The module containing the main package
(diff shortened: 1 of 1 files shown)


rajatjindal:
oops, sorry I opened the PR against `release` branch by mistake. moved it to be against `dev` branch now.

aykevl:
This PR doesn't seem to do anything? It just adds a new field to a struct - there is no effect on the behavior of the compiler.

aykevl (approved):
Oh sorry I completely misread the PR. Yes, this looks fine to me.

rajatjindal:
Thank you very much for accepting this PR

PR #3736, opened 2023-05-20
build: add issues write permission to sizediff job

deadprogram:
The `sizediff` job appears to run successfully when the PR is submitted by TinyGo org team member, but is still failing when PRs are submitted by other contributors.

This PR tries (yet again) to address the `Error: Resource not accessible by integration` error.

Perhaps it is the issues that require this permission?

https://github.com/thollander/actions-comment-pull-request/blob/main/src/main.ts#L114


deadprogram:
Merging since at least it doesn't not work... :smiley_cat:

PR #3737, opened 2023-05-20
interp: support pointer tagging

aykevl:
This is necessary to get https://github.com/tinygo-org/tinygo/pull/3691 working.

I have also refactored/fixed some of the related code to make the last commit a nice clean change.

### DIFF
--- a/interp/interpreter.go
+++ b/interp/interpreter.go
@@ -736,30 +736,25 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 			if err == nil {
 				// The lhs is a pointer. This sometimes happens for particular
 				// pointer tricks.
-				switch inst.opcode {
-				case llvm.Add:
+				if inst.opcode == llvm.Add {
 					// This likely means this is part of a
 					// unsafe.Pointer(uintptr(ptr) + offset) pattern.
 					lhsPtr = lhsPtr.addOffset(int64(rhs.Uint()))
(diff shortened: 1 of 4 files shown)


interp/interpreter.go
```diff
@@ -736,30 +736,25 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 			if err == nil {
 				// The lhs is a pointer. This sometimes happens for particular
 				// pointer tricks.
-				switch inst.opcode {
-				case llvm.Add:
+				if inst.opcode == llvm.Add {
```
deadprogram, line 739: This looks like a good place for:

```go
switch {
    case inst.opcode == llvm.Add:
        // do stuff
    case inst.opcode == llvm.Xor && rhs.Uint() == 0:
        // other stuff
    default:
        // yet other stuff
}
```

deadprogram: I just noticed that it is not quite so easy to replace that code in this PR without changing how the commits are structured.

If I feel strongly I will submit a PR with that change.

aykevl: Oh, yes that's another option. Didn't think of that. It has the same effect though.

I don't really care about the specific form 🤷 I can update the PR if you like.

deadprogram:
Thanks @aykevl for the improvements/fixes. This should unblock @dgryski 

Now merging.

PR #3738, opened 2023-05-21
attiny1616: implement Pin.Get()

aykevl:
I didn't add this method in the initial PR.

Also, I found that a few of my assumptions were incorrect. I've changed the code that configures the pin to make input (floating and pullup) actually work. These chips really are quite different from all the older AVRs.


deadprogram (approved):
:sewing_needle:

PR #3745, opened 2023-05-23
main: use `go env` instead of doing all detection manually

aykevl:
This replaces our own manual detection of various variables (GOROOT, GOPATH, Go version) with a simple call to `go env`.

If the `go` command is not found:

    error: could not find 'go' command: executable file not found in $PATH

If the Go version is too old:

    error: requires go version 1.18 through 1.20, got go1.17

If the Go tool itself outputs an error (using GOROOT=foobar here):

    go: cannot find GOROOT directory: foobar

This does break the case where `go` wasn't available in $PATH but we would detect it anyway (via some hardcoded OS-dependent paths). I'm not sure we want to fix that: I think it's better to tell users "make sure `go version` prints the right value" than to do some automagic detection of Go binary locations.

### DIFF
--- a/builder/config.go
+++ b/builder/config.go
@@ -1,7 +1,6 @@
 package builder
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/tinygo-org/tinygo/compileopts"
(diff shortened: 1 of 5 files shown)


deadprogram:
@aykevl perhaps this PR needs rebase against `dev`?

aykevl:
Rebased.

This PR will probably also conflict with #3824 so would be useful to merge.

deadprogram:
Looking good @aykevl thanks you for the improvement. Now merging.

Issue #3746, opened 2023-05-24
Add Support for BlueMicro Family of boards (nRF52)


jpconstantineau:
Add support for the following boards:
- [BlueMicro](https://www.tindie.com/products/jpconstantineau/bluemicro-nrf52832/) (nRF52832)
- [BlueMicro840](https://www.tindie.com/products/jpconstantineau/bluemicro840/) (nRF52840) (similar to nice!nano)
- [BlueMicro833](https://www.tindie.com/products/jpconstantineau/ebyte-e73-2g4m08s1e-breakout-bluemicro833-pcba/) (nrf52833) 

I am the creator of these boards and they have been available on Tindie for a while.  I would like to add them to TinyGO.  I would be willing to do the work to get them included but before going ahead, I would like to know if a PR would be accepted to be merged in.

Both the BlueMicro840 and BlueMicro833 are currently in [CircuitPython](https://circuitpython.org/downloads?manufacturers=JPConstantineau).  I also have a few keyboards on there too but as these are for a specific purpose, I thought it would be best to leave these out of TinyGo as they can be flashed with a generic rp2040 Pico config.

Please comment on how I should proceed and I'll take care of the work to prepare a PR for inclusion.

jpconstantineau:
I started looking into this.  Here are a few comments and some questions:
## nRF52840:
I far as I could find, all that needs to be done is
- create target file
- create a board.go file

For both cases, I can use the nicenano as  the two boards are very similar.
I noted a key diff in the json target files between the dev and current release branches where the "msd-volume-name" is in [] vs isn't...  I am assuming this is a change across all boards.  I'll test the files with the release I have on my PC and start looking into building and testing from the dev branch.

I'll prepare a PR for it once it's ready.

## nRF52833
This one will require a fair bit of work as none of the other 833 boards used UF2 bootloader and flashing method.  I therefore will need some guidance on getting it supported.  I suspect some USB code from the 840 will need to be added, as well as UF2/softdevice support.

## nRF52832
I havent' looked too much into this one yet but as the original feather nrf52832 isn't there I'll have to rely on the nrf52832 dev kit (PCA10040) with the jlink as the programmer instead of using serial.

jpconstantineau:
Support for the BlueMicro833 will depend on support of the PCA10010 (nRF52833 Dev Kit).

aykevl:
> Support for the BlueMicro833 will depend on support of the PCA10010 (nRF52833 Dev Kit).

[...]

PR #3749, opened 2023-05-25
Add bluemicro840 board

jpconstantineau:
Takes care of the BlueMicro840 of issue #3746 

Schematic is located [here](https://nrf52.jpconstantineau.com/docs/bluemicro840_v1)
USB Vendor ID/Product ID were assigned for firmware on this device by [openmoko](https://github.com/openmoko/openmoko-usb-oui/blob/master/usb_product_ids.psv)

Tested on actual hardware.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -528,6 +528,10 @@ endif
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=pca10059            examples/blinky2
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=bluemicro840        examples/blinky1
+	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=bluemicro840        examples/blinky2
+	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=itsybitsy-m0        examples/blinky1
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=feather-m0          examples/blinky1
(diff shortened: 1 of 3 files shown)


jpconstantineau:
sizediff job fails with this error: Resource not accessible by integration

deadprogram:
> sizediff job fails with this error: Resource not accessible by integration

Please see https://github.com/tinygo-org/tinygo/pull/3757

targets/bluemicro840.json
```diff
@@ -0,0 +1,5 @@
+{
+    "inherits": ["nrf52840", "nrf52840-s140v6-uf2"],
+    "build-tags": ["bluemicro840"],
```
aykevl, line 3: Perhaps you can also add the serial port?

```json
    "serial-port": ["1d50:6161"]
```

aykevl (approved):
Looks good to me, with one small nit.

aykevl (changes requested):
Actually, please also add a smoke test to the Makefile to make sure it continues to build without errors.
(Sorry I forgot about that).

Makefile
```diff
@@ -528,6 +528,10 @@ endif
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=pca10059            examples/blinky2
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=bluemicro840        examples/blinky1
+	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=bluemicro840        examples/blinky2
+	@$(MD5SUM) test.hex
```
aykevl: One example should be good enough - feel free to pick any of the two.
(Each new board adds a second or so to CI time, and we have a lot of boards and a lot of PRs so I like to keep them to a minimum).

aykevl (approved):

PR #3751, opened 2023-05-26
samd21: fix issue with WS2812 driver

aykevl:
The regular port access is around 4 cycles, instead of the usual 2 cycles for a store instruction on Cortex-M0+. The IOBUS however is faster, I didn't measure exactly but I guess it's 2 cycles as expected.

This fixes a bug in the WS2812 driver that only happens on samd21 chips: https://github.com/tinygo-org/drivers/issues/540


ysoldak (approved):
It works, verified on XIAO and Arduino Nano 33 IoT boards.

deadprogram:
Thanks for the fix @aykevl and to @ysoldak for reviewing. Now merging.

PR #3760, opened 2023-06-01
Internal flags

aykevl:
Two small changes:

1. Rename a few flags that are only intended for TinyGo developers.
2. Add a `-internal-nodwarf` flag to avoid debug information generation. This can be useful during development.


deadprogram:
Looks good to me. Anyone else have any thoughts before we merge this?

deadprogram:
Merging since it only impacts developers working on TinyGo itself. Thanks for the improvements @aykevl

PR #3765, opened 2023-06-02
adding new uf2 target for PCA10056

jpconstantineau:
Tested on actual hardware with blink1 and blink2 examples.
Unlike flashing with the "vanilla" pca10056 target, this new target uses the secondary USB port (not the JLINK one) to flash and keeps the bootloader and softdevice on the device.

### DIFF
--- a/targets/pca10056-s140v6-uf2.json
+++ b/targets/pca10056-s140v6-uf2.json
@@ -1,5 +1,5 @@
 {
-	"inherits": ["nrf52840", "nrf52840-s140v6-uf2"],
+    "inherits": ["nrf52840", "nrf52840-s140v6-uf2"],
     "build-tags": ["pca10056"],
     "serial-port": ["239a:0x0029"],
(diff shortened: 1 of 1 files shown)


jpconstantineau:
Adds new target to existing board.
Diff is 0 throughout.

targets/pca10056-s140v6-uf2.json
```diff
@@ -0,0 +1,6 @@
+{
+	"inherits": ["nrf52840", "nrf52840-s140v6-uf2"],
```
deadprogram: Please correct formatting here. Thank you!

jpconstantineau: Agreed.  I think for some reason vscode mixed tabs and spaces and I couldn't see the difference...

jpconstantineau:
Spacing corrected.  Let me know if this is what you had in mind.

deadprogram:
Thanks for the addition @jpconstantineau now squash/merging.

PR #3766, opened 2023-06-02
rp2040: SetPeriod bounds check fix

### LINKED ISSUE #3763: machine/rp2040: Bug in PWM `SetPeriod` bounds check
Drive by issue, might fix this weekend.

The `milliseconds` declared const is actually a whole second! First const block of `SetPeriod`, line 264ish.

soypat:
Resolves https://github.com/tinygo-org/tinygo/issues/3763


deadprogram:
Thanks for the fix @soypat now merging.

PR #3767, opened 2023-06-03
build: only make comment on sizediff job when run from the main repo

deadprogram:
This PR should implement correctly the repo name check to skip trying to make a GH actions comment when that pull request comes from a fork.

https://github.com/orgs/community/discussions/26829#discussioncomment-3253575


deadprogram:
Merging since there are no objections. If this does not work as expected, I think we should remove the commenting part, but leave in the check.

PR #3770, opened 2023-06-04
RP2040 i2c: fix crash when encountering abort while waiting for stop condition

### LINKED ISSUE #3671: rp2040: Transaction with a NDIR CO2 sensor completely disables I2C functionality
A specially crafted transaction will consistently crash the I2C peripheral. This happens after the second `Tx` call when configuring the [Sandbox Electronics 100,000ppm MH-Z16 NDIR CO2 Sensor](https://sandboxelectronics.com/?product=100000ppm-mh-z16-ndir-co2-sensor-with-i2cuart-5v3-3v-interface-for-arduinoraspeberry-pi). 

This driver works fine on the Raspberry Pi 3 and 4 (using Go version) and the ESP32 (using the C++ version), so I suspect it is a RP2040 issue.I have replicated this issue on a Raspberry Pi Pico and a Waveshare Zero. 

I have not been able to reproduce this crash trying to pull SDA permanently high or low. A Pull-up resistor to smooth out the SDA line did not seem to work either.

I rolled back the changes made in [#2743](https://github.com/tinygo-org/tinygo/pull/2743) and this still happens.

## Steps to reproduce
Run the example program here https://github.com/tinygo-org/drivers/pull/551/files

## Events
1. Abort (code number 8) during wait for stop condition: [`!i2c.interrupted(rp.I2C0_IC_RAW_INTR_STAT_STOP_DET)`](https://github.com/tinygo-org/tinygo/blob/0d56dee00f49bd50eb373c02c30062a75ec28f10/src/machine/machine_rp2040_i2c.go#L306)
2. Then it hits the timeout condition in that for loop
3. Returns a `errI2CWriteTimeout`
4. SCL line stays low after this "crash"
5. Future I2C transactions fail returning `ErrRP2040I2CDisable` during [I2C.disable()](https://github.com/tinygo-org/tinygo/blob/0d56dee00f49bd50eb373c02c30062a75ec28f10/src/machine/machine_rp2040_i2c.go#L251)

After the first event no other I2C transaction succeeds and the SCL line stays low. 

<details><summary>Screenshot of I2C transactions</summary>

ndir-rp2040crash-I2C timeout during write
ndir-crasherclosup

</details>

soypat:
Resolves #3671, see issue, it's got all of the context. This PR now allows users to use the NDIR CO2 sensor and other peripherals that could fail during last bits of a write.

**Main Feature:** Fix unrecoverable I2C crash by clearing abort after waiting for stop condition.

**Additional features**:
- Removed the unused `i2c.restartOnNext` and set RESTART bit condition on the first byte being transmitted in a transaction, whether it be a write or read op.
- Added `Reasons` method users can access via an interface cast to stringify an I2C abort condition.

It'd be great get this tested with other peripherals.

### DIFF
--- a/src/machine/machine_rp2040_i2c.go
+++ b/src/machine/machine_rp2040_i2c.go
@@ -52,10 +52,9 @@ type I2CConfig struct {
 }
 
 type I2C struct {
-	Bus           *rp.I2C0_Type
-	restartOnNext bool
-	mode          I2CMode
-	txInProgress  bool
+	Bus          *rp.I2C0_Type
+	mode         I2CMode
+	txInProgress bool
 }
 
 var (
@@ -236,7 +235,6 @@ func (i2c *I2C) init(config I2CConfig) error {
 	if err := i2c.disable(); err != nil {
 		return err
 	}
-	i2c.restartOnNext = false
 
 	i2c.mode = config.Mode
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Thanks for the fix @soypat now merging.

PR #3591, opened 2023-03-23
reflect map fixes

dgryski:
This is a combination of
   https://github.com/tinygo-org/tinygo/pull/3573
   https://github.com/tinygo-org/tinygo/pull/3569
   https://github.com/tinygo-org/tinygo/pull/3557

plus an additional fix and test for interface keys.  Basic YAML now works.

PR #3774, opened 2023-06-10
reflect: use .key() instead of a type assert

aykevl:
This should be ever so slightly more efficient.

This is basically the same change as https://github.com/tinygo-org/tinygo/pull/3591/commits/507d76a64abf5e974fb01357aed7bff3dc8ee872.


dgryski:
Thanks, I missed that one.

dgryski (approved):
LGTM

PR #3778, opened 2023-06-11
goenv: fix version number to 0.28.0 (no -dev)

aykevl:


goenv/version.go
```diff
@@ -12,7 +12,7 @@ import (
 
 // Version of TinyGo.
 // Update this value before release of new version of software.
-const Version = "0.28.0-dev"
+const Version = "0.28.0"
```
deadprogram: We probably need to change this to `0.28.1` now?

aykevl: Done.

deadprogram:
Thanks for the fix @aykevl :smiling_face_with_tear:

PR #3781, opened 2023-06-11
ci: rename release-double-zipped to something more useful

aykevl:
The Linux artifacts have clear names (linux-amd64-double-zipped etc), but the MacOS and Windows ones didn't. This patch renames these artifact names to be more readable, especially when downloading the artifacts.


deadprogram:
@aykevl there are some other references that need to be changed in this PR such as https://github.com/tinygo-org/tinygo/blob/dev/.github/workflows/windows.yml#L217

aykevl:
Oops, that's fixed now.

deadprogram:

PR #3782, opened 2023-06-12
goenv: update to new v0.29.0 development version

sago35:


deadprogram:
Thanks @sago35 now merging.

PR #3795, opened 2023-06-16
reflect: fix iterating over maps with interface{} keys

### LINKED ISSUE #3794: `MapIter.Key()` returns key with wrong type for maps with `any` or `interface{}` keys
When compiled with TinyGo 0.28.1, `iter.Key()` returns `int` key but it should be `interface` key:

```Go
m := map[any]any{1: 2}
rv := reflect.ValueOf(m)
iter := rv.MapRange()

for iter.Next() {
  k := iter.Key()
  fmt.Printf("k.Kind() = %s\n", k.Kind())
}

// Go and TinyGo displays different results:
// - Go 1.19.10:     k.Kind() = interface
// - TinyGo 0.28.1:  k.Kind() = int
```
https://go.dev/play/p/0IkOixuogDH

This bug can cause panic in functions such as `IsNil()` by turning a value of supported type into a value of unsupported type.

dgryski:
Fixes #3794

### DIFF
--- a/src/reflect/value.go
+++ b/src/reflect/value.go
@@ -914,10 +914,11 @@ func (v Value) MapKeys() []Value {
 	e := New(v.typecode.Elem())
 
 	keyType := v.typecode.key()
-	isKeyStoredAsInterface := keyType.Kind() != String && !keyType.isBinary()
+	keyTypeIsEmptyInterface := keyType.Kind() == Interface && keyType.NumMethod() == 0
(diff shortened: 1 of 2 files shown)


fxamacker:
Thanks!  This PR fixed the reported issue.  :+1:

deadprogram (approved):

deadprogram:
Looking good @dgryski thank you for the fix and to @fxamacker for reporting the bug and verifying it. Now merging.

PR #3803, opened 2023-06-21
nrf: wait for stop condition after reading from the I2C bus

aykevl:
Found while working on the PineTime. For some reason it still kind of works in most cases, but I was hitting this issue when interacting with two different I2C devices (the touch sensor and the BMA421).


ysoldak (approved):

deadprogram:
Now merging, thanks @aykevl for the fix and to @ysoldak for review.

Issue #2784, opened 2022-04-19
aws-sdk-go fails tests with "reflect.Value has no field or method FieldByNameFunc"


dkegel-fastly:
This is no surprise, but aws's go sdk needs more reflect.

github.com/aws/aws-sdk-go fails nearly all of its tests in tinygo with the error
```
../../aws/awsutil/path_value.go:70:18: value.FieldByNameFunc undefined (type reflect.Value has no field or method FieldByNameFunc)
```

It also needs encoding/xml, x/sys, and net/http.  It'll be a while before tinygo supports it, but it's interesting to see what the roadblocks will be.

Commenting out the guts of rValuesAtPath() gets past that, kind of, after which the next pervasive problem is
```
../../request/retryer.go:203:20: impossible type assertion: err.(*net.OpError)
	*net.OpError does not implement temporary (missing method Temporary)
```

rockwotj:
I'm getting a similar error trying to include `hamba/avro`

Reference here: https://pkg.go.dev/reflect#Value.FieldByNameFunc

aykevl:
Looks like this was fixed in https://github.com/tinygo-org/tinygo/pull/3805, please comment if it wasn't.

PR #3805, opened 2023-06-22
reflect: Add FieldByNameFunc

rockwotj:
This adds FieldByNameFunc, which some libraries like reflect2 need.

For my usecase I could also just stub FieldByNameFunc to panic, but figured that it would work OK to just make it work. I'm not sure if the overhead to FieldByName using a closure is acceptable.

Fixes: https://github.com/tinygo-org/tinygo/issues/2784


dgryski (approved):
LGTM.  This is what upstream does as well.

deadprogram:
Thank you for adding this @rockwotj and to @dgryski for review. Now merging.

PR #3807, opened 2023-06-23
docker: update clang to version 15

radkomih:
When I try to build the docker image, it fails with this error `Unable to locate package clang-11`.
This commit updates the clang version to match the current one, which is 15.


deadprogram:
Thanks for the improvement @radkomih now merging.

PR #3811, opened 2023-06-27
targets: support comments in target json files

ysoldak:
**New dependency**
- [muzzammil.xyz/jsonc](https://github.com/muhammadmuzzammil1998/jsonc)
This JSONC library is most up-to-date and widely used, also does not bring any extra dependencies.

**Cleanup**
`load()` function was completely dropped, since it does not serve its purpose anymore, as its outdated comment hints upon. Target inheritance is handled by `resolveInherits()` function instead.

**Related**
This PR replaces #2222

### DIFF
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -3,10 +3,8 @@ package compileopts
 // This file loads a target specification from a JSON file.
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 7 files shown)


ysoldak:
@deadprogram can you see what's wrong with CircleCI's "test-all" check?

dgryski:
Obviously this breaks any other tooling that expects "normal" JSON, such as https://github.com/tinygo-org/tinygo/pull/3533 .  We might want to write our own formatting tool to keep the targets files consistent.

ysoldak:
Good point, so I'd rather switch to YAML for target files, but community wants "JSON with comments" 🤷 

Adding comments to JSON obviously breaks compatibility with tools that do not expect comments and this is something was probably overlooked during JSONC discussion.

@soypat @aykevl @deadprogram opinions?

soypat:
Maybe we can avoid comments for internal use? Sorry, I forget if this change was meant for UX or so we could better document internal target.json files.

PR #3816, opened 2023-06-30
main: improve detection of filesystems

aykevl:
This is a rewrite of how filesystems are detected. Specifically, it fixes an issue on Linux where the location of the FAT filesystem can vary between distributions (for example, we supported most distros by checking two different paths, but NixOS uses a different path): it now uses the data in /proc/mounts instead which should be universal.

~WIP: tested on Linux, still need to implement MacOS and Windows support.~

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,6 +19,7 @@ import (
 	"regexp"
 	"runtime"
 	"runtime/pprof"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


aykevl:
Tested on Linux, macOS and Windows. Didn't specifically test on NixOS but it should improve detection of the MSD volumes on there (but even if it doesn't, I think the new method and the refactor is worth it).

sago35:
At least for Windows, I have checked the source code and the behavior.
No problem.

Later, I will check about Linux.

sago35:
Confirmed to work on linux (ubuntu). No problem.
The source code has also been looked at and there are no problems.

LGTM

sago35 (approved):
LGTM

deadprogram:
Thanks for the improvements @aykevl and to @sago35 for review. Now merging.

deadprogram:
Small problem: this change does not work with `usbmount` as used on the TinyHCI server:

https://github.com/tinygo-org/tinygo/runs/14736392846

The problem is that the `/proc/mounts` name for the drive is being set to the next available like `/media/usb2` by `usbmount` and then a symbolic link to the volume name e.g. `/media/tinyhci/CPLAYBOOT`.

https://github.com/rbrito/usbmount is the package being used along with the changes in https://github.com/tinygo-org/tinyhci/tree/main/tools/usbmount

deadprogram:
@aykevl see my comment above. Seems it is not quite so easy to get `usbmount` to actually mount the volumes under a specific name as the code I have been using creates symlinks for that purpose.

What do you think?

deadprogram:
UPDATE: I have made some changes to TinyHCI to use mount/umount for devices that seems to be working. Still testing...

marcofeltmann:
Guess I see your point. 

The previous approach didn't care about the name of the device but scanned for existence of the `INFO_UF2.TXT` file.

That didn't work on NixOS because neither of the scanned directories `/run/media` nor `/media` exist here.

Maybe a hybrid would work?

- search the `/proc/mounts/` for any type of `vfat`
- scan each of the results for the `INFO_UF2.TXT`

Sorry for [fish](https://fishshell.com/) usage, it's more intuitive for me than the (ba)sh stuff.

```fish 
for dir in (cat /proc/mounts | grep vfat | cut -d " " -f 2)
  if test -e $dir/INFO_UF2.TXT
    echo $dir
  end
end

/home/mafe/usb
```

In this particular case the whole NixOS `ramfs`, `tmpfs`, `fuse` and other types are ignored by the `grep vfat`.

The test also excludes `/boot/efi/` from the results list, so only my RP2040 mount remains.

aykevl:
> The previous approach didn't care about the name of the device but scanned for existence of the `INFO_UF2.TXT` file.

Actually it did, see:

https://github.com/tinygo-org/tinygo/blob/d01d85930d19a73010853b7ba8648d8811891bee/main.go#L956

Here it checks for all volumes with a particular name and an `INFO_UF2.TXT` file.

aykevl:
Instead of a name, we could maybe also check the USB vid/pid? I'm not sure I want to go that route though, it is far less intuitive than using the mount name.

PR #3817, opened 2023-07-02
machine/macropad_rp2040: add machine.BUTTON

sago35:
There are many cases in TinyGo examples where BUTTON is used.


deadprogram (approved):

deadprogram:
Thanks for the improvement @sago35 now merging.

PR #3818, opened 2023-07-03
rp2040: add missing suffix to CMD_READ_STATUS

oxplot:
Needed for when using `#define BOARD_QUAD_ENABLE_STATUS_BYTE 1`.


deadprogram:
Thanks for the correction @oxplot now merging.

PR #3821, opened 2023-07-04
machine/rp2040: wait for 1000 us after flash reset to avoid issues with busy USB bus

deadprogram:
This PR modifies the machine/rp2040 USB reset to wait for 1000 microseconds after flash reset to avoid issues with busy USB bus.

See https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/pico_fix/rp2040_usb_device_enumeration/rp2040_usb_device_enumeration.c#L130

### DIFF
--- a/src/machine/machine_rp2040_usb_fix_usb_device_enumeration.go
+++ b/src/machine/machine_rp2040_usb_fix_usb_device_enumeration.go
@@ -3,6 +3,7 @@
 package machine
 
 import (
+	"device/arm"
 	"device/rp"
 )
 
@@ -87,6 +88,7 @@ func hw_enumeration_fix_force_ls_j() {
 	rp.USBCTRL_REGS.USB_MUXING.Set(rp.USBCTRL_REGS_USB_MUXING_TO_DIGITAL_PAD | rp.USBCTRL_REGS_USB_MUXING_SOFTCON)
 
 	// LS_J is now forced but while loop here just to check
+	waitCycles(125000)
 
 	// if timer pool disabled, or no timer available, have to busy wait.
 	hw_enumeration_fix_finish()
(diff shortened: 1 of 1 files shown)


src/machine/machine_rp2040_usb_fix_usb_device_enumeration.go
```diff
@@ -87,6 +88,7 @@ func hw_enumeration_fix_force_ls_j() {
 	rp.USBCTRL_REGS.USB_MUXING.Set(rp.USBCTRL_REGS_USB_MUXING_TO_DIGITAL_PAD | rp.USBCTRL_REGS_USB_MUXING_SOFTCON)
 
 	// LS_J is now forced but while loop here just to check
+	waitCycles(125000)
```
sago35, line 91: waiteCycles() should be added to the line above hw_enumeration_fix_finish().
As much as possible, it should be in the same form as the pico-sdk.

https://github.com/raspberrypi/pico-sdk/blob/6a7db34ff63345a7badec79ebea3aaef1712f374/src/rp2_common/pico_fix/rp2040_usb_device_enumeration/rp2040_usb_device_enumeration.c#L130

sago35: Then, waitCycle(125000) waits for about 6ms.
It might be better to use a shorter value of about `125000 / 5`.

image

soypat:
LGTM- tested and works on my end.

soypat (approved):
LGTM- tested and works on my end with a `-serial=usb` program.

deadprogram:
Thanks for review @soypat now merging.

src/machine/machine_rp2040_usb_fix_usb_device_enumeration.go
```diff
@@ -87,6 +88,7 @@ func hw_enumeration_fix_force_ls_j() {
...
 
 	// if timer pool disabled, or no timer available, have to busy wait.
```
ysoldak, line 93: Comment become broken in two parts

src/machine/machine_rp2040_usb_fix_usb_device_enumeration.go
ysoldak, line 91: Use CPU clock constant?

PR #3824, opened 2023-07-06
all: Go 1.21 support

aykevl:
Work in progress. The main thing that fails is encoding/json.


deadprogram:
@aykevl looks like actual fail in CircleCI https://app.circleci.com/pipelines/github/tinygo-org/tinygo/9801/workflows/61b4c44c-a10f-44c4-b5ed-d1f8603d4c4c/jobs/40999?invite=true#step-118-9

aykevl:
@deadprogram fixed it, now this PR only needs a rebase after #3829 is merged to be ready.

deadprogram:
@aykevl ready for rebase.

aykevl:
I have updated CI for Linux, Windows, and macOS to use Go 1.21-rc2 instead of Go 1.20. This should be changed before the next release (see #3830).

It is now ready for review.

(Note that the last commit isn't really related to Go 1.21, but I added it because it's very small and generally useful).

deadprogram:
@aykevl looks like a bunch of build errors

aykevl:
Looks like this needs some more work:
* There is a build failure on macOS (hopefully just a matter of updating the library).
* It appears that our `//go:wasmimport` is a bit more strict than upstream.

aykevl:
Fixed the issue with `//go:wasmimport`, so at least wasm works with Go 1.21.

* I've reported the issue with macOS upstream: https://github.com/bugst/go-serial/issues/162
* For Windows, it appears we need `sync.OnceValue` (seems to be new in Go 1.21).
* For Linux, there is an issue with AVR that I need to investigate.

I'm glad I didn't wait until after the Go 1.21 release to investigate all these issues.

aykevl:
Did a quick investigation into why AVR fails and it seems we are hitting [this assert](https://github.com/golang/go/blob/master/src/go/types/const.go#L77):

```go
		s := conf.sizeof(T)
		assert(s == 4 || s == 8)
```

Looks like they made the assumption that all `uintptr` values are either 4 or 8 bytes in size.

aykevl:
Reported the AVR bug upstream: https://github.com/golang/go/issues/61249

deadprogram:
> I've reported the issue with macOS upstream: https://github.com/bugst/go-serial/issues/162

Just submitted PR to correct that issue. See https://github.com/bugst/go-serial/pull/163

deadprogram:
> Reported the AVR bug upstream: https://github.com/golang/go/issues/61249

Fix for this has been merged:

https://github.com/golang/go/commit/d983be9cb54de1cde85ba7946bf30c8013cdf5f3

UPDATE: it is not included in Go 1.21 rc3 however it looks like it was merged before the development version was switched to 1.22 so hopefully it will be included in the 1.21 final release.

deadprogram:
> For Windows, it appears we need sync.OnceValue (seems to be new in Go 1.21).

I have created PR #3837 to add these functions, and the Windows tests are now passing in that branch.

deadprogram:
@aykevl PR https://github.com/tinygo-org/tinygo/pull/3838 has been merged into `dev` so please rebase this PR so we can verify it. Thanks!

aykevl:
Rebased.

deadprogram:
@aykevl looks like got further and now has this error in Windows smoketests:

`ld.lld: error: undefined symbol: internal/godebug.registerMetric`

See https://github.com/tinygo-org/tinygo/actions/runs/5613804410/job/15213684058#step:9:60

aykevl:
Pushed a bunch of changes, hopefully it'll pass most tests now...

aykevl:
Yes, looks like all expected tests pass! The only issue remaining is the AVR assert (which will be fixed in RC4 or the final release).

deadprogram:
@aykevl looks like `go1.21rc4` is now available! Can you please try switching this branch to use it to see if everything passes now?

aykevl:
Great, I've updated the PR to use the -rc4 version.

deadprogram (approved):
All tests now passing! I think we can merge this now.... :partying_face:

deadprogram:
Great work on this @aykevl now merging!

builder/config.go
```diff
@@ -27,7 +27,7 @@ func NewConfig(options *compileopts.Options) (*compileopts.Config, error) {
 	if err != nil {
 		return nil, err
 	}
-	if major != 1 || minor < 18 || minor > 20 {
+	if major != 1 || minor < 18 || minor > 21 {
 		// Note: when this gets updated, also update the Go compatibility matrix:
 		// https://github.com/tinygo-org/tinygo-site/blob/dev/content/docs/reference/go-compat-matrix.md
 		return nil, fmt.Errorf("requires go version 1.18 through 1.20, got go%d.%d", major, minor)
```
soypat, line 33: Missed a number `s/1.20/1.21/`

soypat, line 33: Missed a minor here `s/1.20/1.21/l`

compiler/testdata/go1.21.ll
```diff
@@ -0,0 +1,179 @@
+; ModuleID = 'go1.21.go'
```
soypat, line 1: Just curious here- how do you go about writing these `.ll` files? Are you comfortable enough writing them on your own or do you recourse to reference/help?

src/internal/bytealg/bytealg.go
```diff
@@ -251,3 +251,13 @@ func IndexRabinKarp(s, substr string) int {
...
 	return -1
 }
+
+// MakeNoZero makes a slice of length and capacity n without zeroing the bytes.
+// It is the caller's responsibility to ensure uninitialized bytes
+// do not leak to the end user.
+func MakeNoZero(n int) []byte {
+	// Note: this does zero the buffer even though that's not necessary.
+	// For performance reasons we might want to change this (similar to the
+	// malloc function implemented in the runtime).
```
soypat, line 261: Should this start with a `// TODO` ?  Thinking of ease of ctrl+f searching backburner items.

soypat (commented):
Minor nitpicks here and there, looks OK to me.

compiler/testdata/go1.21.ll
aykevl: These files are generated using `go test ./compiler -update`.

src/internal/bytealg/bytealg.go
aykevl: Does anyone every do that? I don't.
I typically use these kinds of comments as a hint to future me (or future someone else). With a TODO, I know something is really missing that should be implemented. In this case I didn't use a TODO because it's working as intended, just a bit slower than possible (and therefore if this shows up in a profiler there's a clear hint how to improve things).

PR #3825, opened 2023-07-06
reflect: implement Value.Grow

aykevl:
This was added in Go 1.20 and becomes necessary for encoding/json in Go 1.21.


deadprogram:
Thank you for the needed addition @aykevl now merging.

PR #3696, opened 2023-05-01
os: implement os.(*File).ReadDir for -target=wasi

achille-roussel:
This PR adds support for reading directories when compiling to programs to WASI.

The implementation uses wasi-libc's DIR type and opendir/readdir/closedir functions.

Please take a look and let me know if you would like to see anything changed.

PR #3827, opened 2023-07-06
wasi: allow zero inodes when reading directories

achille-roussel:
When I submitted https://github.com/tinygo-org/tinygo/pull/3696, I ported the behavior from [Go](https://github.com/golang/go/blob/master/src/os/dir_unix.go#L92-L94) without clear reflection on the reasons why this check existed in the first place.

After further discussion with @codefromthecrypt, it appeared that this check causes more problems than it solves, especially when targeting WASI, due to the guest not knowing the type of underlying file system the host runtime is exposing.

I submitted https://go-review.googlesource.com/c/go/+/507915 to Go to address the issue, and Wazero also recorded more details about it in its [RATIONALE.md](https://github.com/tetratelabs/wazero/blob/008bcdc75ea69d853b3979555758deb7b73dd329/RATIONALE.md#why-dont-we-require-inodes-to-be-non-zero).

This PR addresses the issue for TinyGo by allowing zero inodes when building for the `wasi` target.

### DIFF
--- a/src/os/dir_wasi.go
+++ b/src/os/dir_wasi.go
@@ -53,9 +53,6 @@ func (f *File) readdir(n int, mode readdirMode) (names []string, dirents []DirEn
 		if dirent == nil { // EOF
 			break
 		}
-		if dirent.Ino == 0 {
-			continue
-		}
 		name := dirent.Name()
 		// Check for useless names before allocating a string.
 		if string(name) == "." || string(name) == ".." {


dgryski (approved):
LGTM

codefromthecrypt:
thanks so much @achille-roussel! I'm glad this [landed upstream](https://github.com/golang/go/commit/39c507071251dbf5ea098077ad0e791679dae548) and will get into tinygo, too!

deadprogram:
Thank you for the improvement @achille-roussel and to @dgryski for the quick review. Now merging.

PR #3829, opened 2023-07-07
reflect: add SetZero

aykevl:
This was added in Go 1.20 and is required by encoding/json starting with Go 1.21. See: #3824.


deadprogram (approved):

deadprogram:

PR #3832, opened 2023-07-13
machine.UART refactor

### LINKED ISSUE #3537: Need UART "Flush" method on RP2040

soypat:
Resolves #3537, closes #3538.

Adds:
- `UART.flush`: method implemented by UARTs that blocks until all data has finished being transmitted
- `uart.writeByte`: method implemented by UARTs that writes over the UART pin without blocking.
- Comments to WriteBytes shared by all implementations.

Semantic changes:
- `UART.WriteByte` and `UART.Write`: Methods now expected to block until all data has been finished being transmitted over wire.

### DIFF
--- a/src/machine/uart.go
+++ b/src/machine/uart.go
@@ -73,8 +73,11 @@ func (uart *UART) WriteByte(c byte) error {
 // Write data over the UART's Tx.
 // This function blocks until the data is finished being sent.
 func (uart *UART) Write(data []byte) (n int, err error) {
-	for _, v := range data {
-		uart.writeByte(v)
+	for i, v := range data {
+		err = uart.writeByte(v)
+		if err != nil {
+			return i, err
+		}
 	}
 	uart.flush() // flush() blocks until all data has been transmitted.
 	return len(data), nil
--- a/src/machine/uart.go
+++ b/src/machine/uart.go
@@ -59,10 +59,20 @@ func (uart *UART) Read(data []byte) (n int, err error) {
 	return size, nil
 }
 
+// WriteByte writes a byte of data to the UART.
+func (uart *UART) WriteByte(c byte) error {
+	err := uart.writeByte(c)
+	if err != nil {
+		return err
+	}
+	uart.flush() // flush() blocks until all data has been transmitted.
+	return nil
+}
+
 // Write data to the UART.
 func (uart *UART) Write(data []byte) (n int, err error) {
 	for _, v := range data {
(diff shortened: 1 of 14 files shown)


aykevl (approved):
Looks good to me.

This may break some application that assumed the previous incorrect behavior, but it is certainly an improvement over what we had previously.

We might want to do something similar for `ReadByte`: at the moment it does not block while the vast majority of applications need a blocking interface.

sago35 (approved):
LGTM

src/machine/uart.go
```diff
@@ -59,11 +59,24 @@ func (uart *UART) Read(data []byte) (n int, err error) {
...
+	uart.flush() // flush() blocks until all data has been transmitted.
+	return nil
+}
+
+// Write data over the UART's Tx.
+// This function blocks until the data is finished being sent.
 func (uart *UART) Write(data []byte) (n int, err error) {
 	for _, v := range data {
-		uart.WriteByte(v)
+		uart.writeByte(v)
```
kenbell: Should handle the error return?  Otherwise will 'silently' eat the buffer?

aykevl: While some hardware might cause an error there, the TX side of a UART only sends so I don't see how that would normally result in an error. The error return is mainly there to match the signature of bufio.
(This is different from ReadByte which might actually get an invalid signal as input, although I'm not sure this is even detectable in a UART).

That said, if the `writeByte` method always returns nil that should be const-propagated meaning that the compiler will likely optimize the error check away. So it wouldn't harm to check the return value just to be sure.

soypat:
One of the jobs failed- maybe a flaky build? @deadprogram @aykevl 
https://app.circleci.com/pipelines/github/tinygo-org/tinygo/9827/workflows/dfe682fa-9055-4cf6-9fa2-49b01c463f91/jobs/41028

src/machine/uart.go
soypat: Done!

soypat:
OK, the flaky test passed. Merging now. Thank you everyone for the great feedback!

PR #3833, opened 2023-07-13
nrf: add I2C timeout

aykevl:
This commit adds I2C timeouts for nrf51 and nrf52832 (but not yet for others like nrf52840).

Tested on the PineTime (nrf51832), where I now got a timeout instead of hanging and resetting due to a watchdog reset.

I don't know what a good timeout value is, I picked ~30ms because that seems long enough for me for most devices to respond yet still short enough for a typical watchdog.


deadprogram (approved):

deadprogram:
Now merging, thanks for the addition @aykevl

PR #3836, opened 2023-07-15
rp2040: add NoPin support

### LINKED ISSUE #3819: Cannot use machine.NoPin with rp2040
The machine.Pin implementation of the RP2040 is as follows

https://github.com/tinygo-org/tinygo/blob/v0.28.1/src/machine/machine_rp2040_gpio.go#L123-L129

Perhaps just adding a guard in this area will eliminate the problem.
I will check later.

soypat:
These changes resolve #3819.

I've not added the check to `Get` since it would not actively interfere with a program, though that would mean `NoPin.Get` may return `GPIO0` logic level. Not sure if NoPin is documented as doing so...


deadprogram:
Thanks for the addition @soypat now merging.

PR #3838, opened 2023-07-19
sync: add implementation from upstream Go for OnceFunc, OnceValue, and OnceValues

deadprogram:
This PR adds implementation to the `sync` package from upstream Go for `OnceFunc()`, `OnceValue()`, and `OnceValues()`.

Needed to satisfy requirements for Go 1.21 support under Windows. See https://github.com/tinygo-org/tinygo/pull/3824

### DIFF
--- a/src/sync/oncefunc_test.go
+++ b/src/sync/oncefunc_test.go
@@ -1,0 +1,159 @@
+// Copyright 2022 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package sync_test
+
+import (
+	"sync"
+	"testing"
+)
+
+// We assume that the Once.Do tests have already covered parallelism.
+
+func TestOnceFunc(t *testing.T) {
+	calls := 0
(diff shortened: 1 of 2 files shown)


src/sync/oncefunc_test.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	if v1 != 1 || v2 != 2 {
+		t.Errorf("want v1==1 and v2==2, got %d and %d", v1, v2)
+	}
+	if allocs != 0 {
+		t.Errorf("want 0 allocations per call, got %v", allocs)
+	}
+}
+
+// TODO: need to implement more complete panic handling for these tests.
```
aykevl, line 64: What if you exclude this test on wasm? You can check `runtime.GOARCH == "wasm"` and skip the test.

aykevl (approved):
This looks good to me.

src/sync/oncefunc_test.go
deadprogram: There are too many other unsupported things still, such as `undefined: runtime.PanicNilError`.

We can and should work on that as well, but probably best as a separate PR.

Issue #3501, opened 2023-03-01
WASI linker error: undefined symbol `__multi3`


flavio:
I ran into this linker issue when trying to build some Go code:

```console
$ tinygo build -x \
	-o policy.wasm \
	-target=wasi \
	-no-debug .
clang -mbulk-memory -mnontrapping-fptoint -msign-ext --sysroot=/usr/local/tinygo/lib/wasi-libc/sysroot -gdwarf-4 -Oz --target=wasm32-unknown-wasi -mcpu=generic -MD -MV -MTdeps -MF /tmp/tinygo2136342536/dep-1473137204.d -flto=thin -c -o /home/tinygo/.cache/tinygo/tmp-2233786165.bc /usr/local/tinygo/src/internal/task/task_asyncify_wasm.S -Qunused-arguments
clang -mbulk-memory -mnontrapping-fptoint -msign-ext --sysroot=/usr/local/tinygo/lib/wasi-libc/sysroot -gdwarf-4 -Oz --target=wasm32-unknown-wasi -mcpu=generic -MD -MV -MTdeps -MF /tmp/tinygo2136342536/dep-2756220689.d -flto=thin -c -o /home/tinygo/.cache/tinygo/tmp-1731379903.bc /usr/local/tinygo/src/runtime/asm_tinygowasm.S -Qunused-arguments
wasm-ld --stack-first --no-demangle -L /usr/local/tinygo -o /tmp/tinygo2136342536/main --strip-debug --compress-relocations /tmp/tinygo2136342536/main.o /home/tinygo/.cache/tinygo/obj-0bfb1c88ecd5881542b47db010b5787164fc6a554d252cff6bd66805.bc /home/tinygo/.cache/tinygo/obj-21efc40633d7331a1c71e816bcdcc1a96f69a1f1969201b2c8be7b5f.bc /usr/local/tinygo/lib/wasi-libc/sysroot/lib/wasm32-wasi/libc.a -mllvm -mcpu=generic --lto-O2 --thinlto-cache-dir=/home/tinygo/.cache/tinygo/thinlto -mllvm --rotation-max-header-size=0
tinygo:wasm-ld: error: lto.tmp: undefined symbol: __multi3
failed to run tool: wasm-ld
error: failed to link /tmp/tinygo2136342536/main: exit status 1
```

This happens with tinygo version 0.27.0, using the official container image. I can reproduce it also with tinygo built from the `dev` branch.

## Root cause

[...]

PR #3844, opened 2023-07-26
compiler: add compiler-rt and wasm symbols to table

charlie-haley:
This PR adds `compiler-rt` and updates the `makeArchive` function to read symbols from wasm binaries and add them to the symbol table.

Fixes https://github.com/tinygo-org/tinygo/issues/3501

### DIFF
--- a/builder/ar.go
+++ b/builder/ar.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"time"
 
+	wasm "github.com/aykevl/go-wasm"
 	"github.com/blakesmith/ar"
 )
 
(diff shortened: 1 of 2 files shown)


builder/ar.go
```diff
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"time"
 
+	wasm "github.com/aykevl/go-wasm"
```
aykevl, line 15: Note that the package is already named "wasm", so this would be fine:

```suggestion
	"github.com/aykevl/go-wasm"
```

(Many people think the directory name is used, but that's not true: the name from `package wasm` is used).

aykevl (approved):
Looks good to me. Normally I'd like to see a test for changes like these, but I'm not quite sure how to test this reliably.

targets/wasi.json
```diff
@@ -7,6 +7,7 @@
 	"goarch":        "arm",
 	"linker":        "wasm-ld",
 	"libc":          "wasi-libc",
+	"rtlib":         "compiler-rt",
```
aykevl, line 10: Actually now I realize this probably needs to be added to targets/wasm.json too. Feel free to make a new PR if you'd like.

charlie-haley: New PR for `wasm.json` https://github.com/tinygo-org/tinygo/pull/3848

aykevl:
@charlie-haley FYI: https://github.com/tinygo-org/tinygo/pull/4188

PR #3845, opened 2023-07-28
wasm: fix functions exported through //export

aykevl:
When a function is exported using `//export`, but also had a `//go:wasm-module` pragma, the `//export` name was ignored. The `//go:wasm-module` doesn't actually do anything besides breaking the export (exported functions don't have a module name) so I've made sure to remove this check.

I've also refactored and clarified the code a bit to avoid unnecessary complexity.


deadprogram:

PR #3848, opened 2023-07-31
compiler: add compiler-rt to wasm.json

charlie-haley:
Following on from #3844, adding `compiler-rt` to wasm.json


deadprogram:
Thanks for the addition @charlie-haley now merging.

PR #3849, opened 2023-08-01
machine/usb: remove usbDescriptorConfig

sago35:
This PR deletes usbDescriptorConfig.
Then, the usbDescriptor updates that were performed when sendDescriptor() was executed are moved to EnableXXX().

This allows the user program to initialize any Endpoint above 4.


deadprogram (approved):
Thank you for the improvement @sago35

PR #3850, opened 2023-08-01
machine/usb: allow USB Endpoint settings to be changed externally

sago35:
This PR is configured from machine/usb/xxx instead of machine using the changes in #3849.
This PR needs to be merged after #3849.

### DIFF
--- a/src/machine/usb/config.go
+++ b/src/machine/usb/config.go
@@ -1,7 +1,7 @@
 package usb
 
 type EndpointConfig struct {
-	No        uint8
+	Index     uint8
 	IsIn      bool
 	TxHandler func()
 	RxHandler func([]byte)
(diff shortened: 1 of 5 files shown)


deadprogram:
You can rebase this PR now @sago35 thank you!

sago35:
Rebased. @deadprogram

src/machine/usb/config.go
```diff
@@ -0,0 +1,14 @@
+package usb
+
+type EndpointConfig struct {
+	No        uint8
```
deadprogram: This might be better named `Num` or perhaps `ID`?

sago35: Perhaps `Index` would be a better name.

sago35: Renamed.

deadprogram:
Thank you for the improvements @sago35 now merging.

PR #3851, opened 2023-08-01
machine/usb/hid: add RxHandler interface

sago35:
This PR needs to be merged after #3850

This PR adds `RxHandler()` interface to usbhid.
RxHandler is used, for example, with HID serial. RxHandler is also used for NUMLOCK indicators.

### DIFF
--- a/src/machine/usb/hid/hid.go
+++ b/src/machine/usb/hid/hid.go
@@ -23,7 +23,7 @@ const (
 )
 
 type hidDevicer interface {
-	Handler() bool
+	TxHandler() bool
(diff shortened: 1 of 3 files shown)


deadprogram:
This PR can be rebased now @sago35 if you please.

deadprogram:
@sago35 I am just wondering if we should rename `Handler()` to `TxHandler()` along with this change?

sago35:
OK.
Renamed. @deadprogram

deadprogram:
Thank you for the improvements @sago35 now merging.

PR #3855, opened 2023-08-06
Implement watchdog support on RP2040, SAMD51, NRF528xx, STM32

kenbell:
The watchdog hardware is quite different, but the basics seem consistent enough for a common abstraction.

In my case, I need watchdog because I've got some code that is crashing after an 'indeterminate amount of time' on a tinygo based sensor that i don't have good access to.  A quick restart is beneficial while I track down the bug...

In general watchdog is useful for production use where things 'need to just work'

### DIFF
--- a/src/machine/machine_atsamd51.go
+++ b/src/machine/machine_atsamd51.go
@@ -2299,3 +2299,52 @@ func checkFlashError() error {
 
 	return nil
 }
+
+// Watchdog provides access to the hardware watchdog available
+// in the SAMD51.
+var Watchdog = &watchdogImpl{}
+
+const (
+	// WatchdogMaxTimeout in milliseconds (16s)
+	WatchdogMaxTimeout = (16384 * 1000) / 1024 // CYC16384/1024kHz
+)
+
+type watchdogImpl struct{}
+
+// Configure the watchdog.
+//
+// This method should not be called after the watchdog is started and on
+// some platforms attempting to reconfigure after starting the watchdog
+// is explicitly forbidden / will not work.
+func (wd *watchdogImpl) Configure(config WatchdogConfig) error {
+	// 1.024kHz clock
+	cycles := int((int64(config.TimeoutMillis) * 1024) / 1000)
+
+	// period is expressed as a power-of-two, starting at 8 / 1024ths of a second
+	period := uint8(0)
+	cfgCycles := 8
+	for cfgCycles < cycles {
+		period++
+		cfgCycles <<= 1
+
+		if period >= 0xB {
+			break
+		}
+	}
+
+	sam.WDT.CONFIG.Set(period << sam.WDT_CONFIG_PER_Pos)
+
+	return nil
+}
+
+// Starts the watchdog.
+func (wd *watchdogImpl) Start() error {
+	sam.WDT.CTRLA.SetBits(sam.WDT_CTRLA_ENABLE)
+	return nil
+}
+
+// Update the watchdog, indicating that `source` is healthy.
+func (wd *watchdogImpl) Update() {
+	// 0xA5 = magic value (see datasheet)
+	sam.WDT.CLEAR.Set(0xA5)
+}
--- a/src/machine/machine_rp2040_watchdog.go
+++ b/src/machine/machine_rp2040_watchdog.go
@@ -4,24 +4,67 @@ package machine
 
 import (
 	"device/rp"
-	"runtime/volatile"
-	"unsafe"
 )
 
-type watchdogType struct {
-	ctrl    volatile.Register32
-	load    volatile.Register32
-	reason  volatile.Register32
-	scratch [8]volatile.Register32
-	tick    volatile.Register32
+// Watchdog provides access to the hardware watchdog available
+// in the RP2040.
+var Watchdog = &watchdogImpl{}
+
+const (
+	// WatchdogMaxTimeout in milliseconds (approx 8.3s).
+	//
+	// Nominal 1us per watchdog tick, 24-bit counter,
+	// but due to errata two ticks consumed per 1us.
+	// See: Errata RP2040-E1
+	WatchdogMaxTimeout = (rp.WATCHDOG_LOAD_LOAD_Msk / 1000) / 2
+)
+
+type watchdogImpl struct {
+	// The value to reset the counter to on each Update
+	loadValue uint32
 }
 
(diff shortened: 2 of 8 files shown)


kenbell:
@deadprogram - any idea about the CircleCI failures?  Seems like it's unrelated to these changes - and seeing a couple different failure modes SIGSEGV vs SIGABRT

```
/go/bin/tinygo build -size short -o test.hex -target=pca10040 -opt=0     ./testdata/stdlib.go
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0xfffffffffffffffa addr=0xa359 pc=0x7f77c618c60c]

runtime stack:
runtime.throw({0x835c59?, 0x7f77c618c625?})
	/usr/local/go/src/runtime/panic.go:992 +0x71
runtime.sigpanic()
	/usr/local/go/src/runtime/signal_unix.go:802 +0x389
```

```
cd tests/os/smoke && /go/bin/tinygo test -c -target=pybadge && rm smoke.test
corrupted double-linked list
SIGABRT: abort
PC=0x7fa5a60618eb m=41 sigcode=18446744073709551610
signal arrived during cgo execution
```

```
/go/bin/tinygo build -size short -o test.hex -target=pca10040 -opt=0     ./testdata/stdlib.go
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0xfffffffffffffffa addr=0xa384 pc=0x7fd2f5a06f20]

runtime stack:
runtime.throw({0x835c59?, 0x7fd288021780?})
	/usr/local/go/src/runtime/panic.go:992 +0x71
runtime.sigpanic()
	/usr/local/go/src/runtime/signal_unix.go:802 +0x389
```

deadprogram:
There are spurious build fails on CircleCI nearly every build. I do not know why, might be related to the size of the free instance we get from them? I restart it and it eventually passes... :crying_cat_face:

kenbell:
:-(   When do we drop support for Go 1.18?

deadprogram:
> When do we drop support for Go 1.18?

We can possibly drop it after the Go 1.21 release?

src/machine/machine_atsamd51.go
```diff
@@ -2299,3 +2299,81 @@ func checkFlashError() error {
...
+
+const (
+	// WatchdogMaxSources is 8
+	//
+	// SAMD51 only supports a single source in hardware, so
+	// the support for multiple sources is in software.
+	WatchdogMaxSources = 8
+
+	// WatchdogMaxTimeout is 16s
+	WatchdogMaxTimeout = 16 // CYC16384/1024kHz
```
soypat: Should these units be consistent with the configuration struct units? (Prefer smaller units, milliseconds)

soypat (commented):
WatchdogMaxTimeout units should ideally be identical among implementations and in the config struct

src/machine/machine_atsamd51.go
kenbell: Thanks - good spot.  I've definitely messed up the units, I'll fix it up.

kenbell:
I've put some draft docs here: https://github.com/tinygo-org/tinygo-site/pull/371

deadprogram:
@kenbell finally looking at this a bit more closely. Thanks for working on it!

Since the only one of the processors with multiple "sources" is the Nordic Semi WDT, I wonder if perhaps it might be better to simplify the code by only supporting a single source for now.

I find the syntax `ExtraSources:  0` a little clumsy. If we stick with multiple sources, perhaps better to use something explicit like `NumSources: 2`? The `Configure()` could set `NumSources = 1` if it is set to 0.

Final comment, might be cool to define something like this in `watchdog.go`:

```go
const WatchdogSourceDefault = iota
```

That would allow cleaner syntax with a single source. 


If defining multiple sources in the example, could be like this:

```go
const (
      watchMain = iota
      watchSecond
      watchThird
)
```

What do you think?

src/machine/machine_atsamd51.go
```diff
@@ -2299,3 +2299,81 @@ func checkFlashError() error {
...
+// Watchdog provides access to the hardware watchdog available
+// in the SAMD51.
+var Watchdog = &watchdogImpl{}
+
+const (
+	// WatchdogMaxSources is 8
+	//
+	// SAMD51 only supports a single source in hardware, so
+	// the support for multiple sources is in software.
+	WatchdogMaxSources = 8
```
soypat: I just noticed this value is a result of the software implementation, not the actual hardware. Should we really provide a software abstraction over the hardware at this level? 

Maybe this is worth implementing separately as a WatchdogMultiplexer type, within or out of `machine` package that receives a `Watcher` interface.

deadprogram: > Should we really provide a software abstraction over the hardware at this level?

That is exactly the question, and I think the answer is probably no.

kenbell:
@deadprogram @soypat - thanks.  Yeah, I agree - I'll remove the 'multiple sources' implementation.  It's easy enough for an app to implement if needed.

src/machine/machine_rp2040_watchdog.go
```diff
@@ -4,24 +4,67 @@ package machine
...
+	// The value to reset the counter to on each Update
+	loadValue uint32
 }
 
-var watchdog = (*watchdogType)(unsafe.Pointer(rp.WATCHDOG))
+// Configure the watchdog.
+//
+// This method should not be called after the watchdog is started and on
+// some platforms attempting to reconfigure after starting the watchdog
+// is explicitly forbidden / will not work.
```
soypat, line 31: This seems to have been copy-pasted for each individual watchdog implementation, should we have a single watchdog.go implementation which calls `Watchdog.configure(cfg)`?, same goes for `Start`

kenbell: I don't think we've done that anywhere else, just for docs.  Would be kind of a big change to remove public methods from platform-specific files.  Happy to do it if we think it's beneficial - the public function should easily get inlined.

kenbell:
I've added an implementation for stm32

src/machine/machine_rp2040_watchdog.go
deadprogram: I think we should leave it as it is now, for the reasons @kenbell just mentioned.

deadprogram (approved):
Looking good for me now @kenbell thank you for working on this!

Issue #3856, opened 2023-08-07
dev: on clean systems, git submodule update --init can have authn issues


kenbell:
The new renesas SVD submodule has an SSH 'URL' unlike the other submodules which are HTTPS.  This can lead to public key errors (repro case unclear).

We should use an HTTPS url for this submodule, the same as other submodules.

deadprogram:
Since the Renesas submodule was merged only in the current `dev` branch, now closing this. Thanks @kenbell for the fix!

PR #3857, opened 2023-08-07
all: use https for renesas submodule #3856

kenbell:
Addressess #3856


deadprogram:
Thanks for the fix @kenbell now merging.

PR #3858, opened 2023-08-09
build: switch GH actions builds to use Go 1.21 final release

### LINKED ISSUE #3830: Use Go 1.21 in CI
We need to start using Go 1.21 in CI, once #3824  is merged and it is available in CI (probably once Go 1.21 is released).

deadprogram:
This PR switches the GH actions builds to use Go 1.21 final release.

Fixes #3830

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v3
         with:
-          go-version: '1.21.0-rc.4'
+          go-version: '1.21'
           cache: true
       - name: Restore LLVM source cache
         uses: actions/cache/restore@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/build-macos.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v3
         with:
-          go-version: '1.21.0-rc.4'
+          go-version: '1.21.0'
```
aykevl: Can you change this to `1.21`? My understanding is that `1.21` will pick the latest point release instead of pinning it to 1.21.0.

deadprogram: Done,

aykevl (approved):

PR #3860, opened 2023-08-10
modules: update to go-serial package v1.6.0

deadprogram:
This PR updates TinyGo to use the newest release of the go-serial package v1.6.0 with needed changes for Go 1.21


aykevl:
Glad they updated go-serial, I never really liked the replace situation.

Issue #3685, opened 2023-04-27
wasm: allow GOOS=wasip1 to substitute for target=wasi


codefromthecrypt:
To allow drop-in replacement with go 1.21, it would seem nice to support `GOOS=wasip1` as an alternative to `-target=wasi`.

Per activity on https://github.com/golang/go/issues/58141, `GOOS=wasip1` will ship in Go 1.21 even if a subset of functionality vs the TinyGo equivalent target.

dgryski:
I agree we should do this. We also need to figure out how this will affect build tags. Do we migrate all of tinygo to the new tag to match upstream and add it to the list of tags exported to users, etc.

aykevl:
Same, we should add support for `GOOS=wasip1`. And perhaps also `-target=wasip1` for consistency.

deadprogram:
Is anyone working on this one yet?

aykevl:
I've made a draft PR for this feature: https://github.com/tinygo-org/tinygo/pull/3861

deadprogram:
Now that #3861 has merged, I am marking this PR to close on next release. Thanks everyone!

deadprogram:
Closing since this is part of release 0.29.0 thank you very much everyone!

PR #3700, opened 2023-05-02
os: implement os.Chtimes

achille-roussel:
Opening this PR as a draft because I'm running into an issue that someone more familiar with the TinyGo build may be more equipped to solve.

`utimes` in wasi-libc has a dependency on `__builtin_mul_overflow` via https://github.com/WebAssembly/wasi-libc/blob/3189cd1ceec8771e8f27faab58ad05d4d6c369ef/libc-bottom-half/cloudlibc/src/common/time.h, which gets translated by LLVM to a call to `__muloti4` from compiler-rt, which isn't being linked to `tinygo`.

```
$ ./build/tinygo test -target=wasi os
tinygo:wasm-ld: error: /home/achille_roussel/go/src/github.com/achille-roussel/tinygo/lib/wasi-libc/sysroot/lib/wasm32-wasi/libc.a(utimensat.o): undefined symbol: __muloti4
tinygo:wasm-ld: error: /home/achille_roussel/go/src/github.com/achille-roussel/tinygo/lib/wasi-libc/sysroot/lib/wasm32-wasi/libc.a(utimensat.o): undefined symbol: __muloti4
failed to run tool: wasm-ld
FAIL	os	0.000s
error: failed to link /tmp/tinygo1517931148/main: exit status 1
```

PR #3699, opened 2023-05-02
os: implement os.(*File).Truncate and os.Truncate

achille-roussel:
This PR adds an implementation of `os.(*File).Truncate` and `os.Truncate`. The implementation is based on libc's `truncate` and `ftruncate` functions. I ported the tests from the Go standard library as well.

PR #3861, opened 2023-08-10
wasm: add support for GOOS=wasip1

aykevl:
This is WIP, a preview of what we might want to add (see https://github.com/tinygo-org/tinygo/issues/3685).

This is bigger than you might expect, for two reasons:

1. This uses GOOS=wasip1 support in the Go standard library (added in Go 1.21) instead of using our old support for WASM that basically just pretended to be linux/arm as a workaround for lack of native support.
2. Support is entirely built into the compiler (instead of in a target JSON file), similar to other operating systems (linux, darwin, windows) but unlike our existing js/wasm and wasi/wasm support. I'm not entirely sure this is a good idea, although it does make sense if we expose this feature only using environment variables and not using the `-target` flag.

It's not fully working, I'll need to continue working on it later. But a basic "hello world" already compiles.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -153,6 +153,7 @@ jobs:
           tar -C ~/lib -xf tinygo.linux-amd64.tar.gz
           ln -s ~/lib/tinygo/bin/tinygo ~/go/bin/tinygo
       - run: make tinygo-test-wasi-fast
+      - run: make tinygo-test-wasip1-fast
       - run: make smoketest
   assert-test-linux:
     # Run all tests that can run on Linux, with LLVM assertions enabled to catch
--- a/Makefile
+++ b/Makefile
@@ -417,8 +417,12 @@ tinygo-bench-fast:
 # Same thing, except for wasi rather than the current platform.
 tinygo-test-wasi:
 	$(TINYGO) test -target wasi $(TEST_PACKAGES_FAST) $(TEST_PACKAGES_SLOW) ./tests/runtime_wasi
+tinygo-test-wasip1:
+	GOOS=wasip1 GOARCH=wasm $(TINYGO) test $(TEST_PACKAGES_FAST) $(TEST_PACKAGES_SLOW) ./tests/runtime_wasi
 tinygo-test-wasi-fast:
 	$(TINYGO) test -target wasi $(TEST_PACKAGES_FAST) ./tests/runtime_wasi
+tinygo-test-wasip1-fast:
+	GOOS=wasip1 GOARCH=wasm $(TINYGO) test $(TEST_PACKAGES_FAST) ./tests/runtime_wasi
 tinygo-bench-wasi:
 	$(TINYGO) test -target wasi -bench . $(TEST_PACKAGES_FAST) $(TEST_PACKAGES_SLOW)
 tinygo-bench-wasi-fast:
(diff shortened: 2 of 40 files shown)


deadprogram:
> This is bigger than you might expect

Perhaps we should move this to milestone for `0.30` release so we do not hold up a Go 1.21 compatible release in the meantime.

@aykevl @codefromthecrypt anyone else what do you think?

aykevl:
Actually I think I can finish this PR quickly. It seems to pass all tests locally at least.

aykevl:
Eventually I'd like to remove our old WASI support entirely (currently usable via `-target=wasi`), perhaps aliasing `-target=wasi` to the equivalent of `GOOS=wasip1 GOARCH=wasm` to not break backwards compatibility too badly. But for now it seems reasonable to keep support for both. It's not too difficult to support both at the same time: it's mainly just slightly more complicated `//go:build` lines and somewhat slower CI.

deadprogram:
```
/root/project/src/runtime/os_wasip1.go:8:7: GOOS redeclared in this block
/root/project/src/runtime/os_linux.go:10:7: 	other declaration of GOOS
# runtime
/root/project/src/runtime/os_wasip1.go:8:7: GOOS redeclared in this block
/root/project/src/runtime/os_linux.go:10:7: 	other declaration of GOOS
make: *** [Makefile:449: smoketest] Error 1

Exited with code exit status 2
```

Actual fail in CircleCI build https://app.circleci.com/pipelines/github/tinygo-org/tinygo/9892/workflows/c400192b-ca42-4e4a-a87c-f7743e786c75/jobs/41137?invite=true#step-118-9

aykevl:
Thanks, fixed now.

aykevl:
Fixed an issue on `tinygo test os` (due to the renaming of stat_linux.go).

deadprogram (approved):
Looking good from my point of view.

@dgryski @codefromthecrypt @dkegel-fastly what do you think?

deadprogram:
Last call for comments before merging this PR!

cc @dgryski @codefromthecrypt @dkegel-fastly @achille-roussel and anyone else who is interested.

achille-roussel (approved):
This looks great 🙌 it's actually a much smaller change than I would have expected, which is great 👍 

It might be useful to document when should users select `-target=wasi` vs `GOOS=wasip1` in a README or similar, I believe that the historical implementation still had a few features that were not supported, for example:
- https://github.com/tinygo-org/tinygo/pull/3699
- https://github.com/tinygo-org/tinygo/pull/3700

I imagine users will probably stubble upon those differences and be confused about why certain functionalities work in one WASI but not the other. I expect we would get questions like "when should we use -target=wasi vs GOOS=wasip1?" and recording those answers will likely reduce the maintenance load.

Thanks for doing this work!

dkegel-fastly:
running my app's tests now...

dkegel-fastly:
In preparation for testing this branch, I did a hopefully clean build of the latest dev branch, but it fails "make test", so maybe hold off until that's settled...? see slack.

deadprogram:
@dkegel-fastly see https://github.com/tinygo-org/tinygo/issues/3871 to track that issue. Since it is not related to this PR, I am going to merge this one to keep things moving along.

dgryski:
BTW, this seems to pass the test corpus just fine.

Issue #3865, opened 2023-08-12
modules: error getting modules running `tinygo build` when using Go 1.21


deadprogram:
There is an error under Go 1.21 when TinyGo tries to download needed modules when running `tinygo build`

`GOPROXY list is not the empty string, but contains no entries`

Here is an example of this error in the CI build for TinyFont:
https://github.com/tinygo-org/tinyfont/actions/runs/5840094195/job/15838964200#step:7:7

See https://github.com/hashicorp/vault/issues/22292 for another project with this same issue.

deadprogram:
The root cause is https://github.com/golang/go/issues/61928

deadprogram:
Closing since this is part of release 0.29.0 thank you everyone!

PR #3866, opened 2023-08-12
loader: merge go.env file which is now required starting in Go 1.21

deadprogram:
This PR modifies the loader to merge the `go.env` file which is now required starting in Go 1.21 to correctly get required packages when building.

Solves https://github.com/tinygo-org/tinygo/issues/3865


aykevl (approved):
Looks reasonable, thank you!

PR #3867, opened 2023-08-12
rp2040: move flash related functions into separate file from C import

### LINKED ISSUE #3852: gopls does not work for machine.Flash in rp2040
For example, as shown below, LSP works for machine.LED, but not for machine.Flash.
`import "C"` may be affected.

image

deadprogram:
This PR moves the flash related functions for rp2040 into separate file from the C imports for correct LSP.

Fixes #3852

### DIFF
--- a/src/machine/machine_rp2040_flash.go
+++ b/src/machine/machine_rp2040_flash.go
@@ -1,0 +1,97 @@
+//go:build rp2040
+
+package machine
+
+import (
+	"bytes"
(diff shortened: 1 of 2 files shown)


sago35 (approved):
Confirmed by XIAO-RP2040.
gopls is working very well.
Thanks!

deadprogram:
Glad it is working as expected @sago35 thanks for testing. Now merging.

PR #3882, opened 2023-08-27
update binaryen submodule to version 114

### LINKED ISSUE #3881: wasm-opt error parsing wasm [parse exception: invalid function index (at 0:864478)]
[parse exception: invalid function index (at 0:864478)]
Fatal: error parsing wasm
error: wasm-opt failed: exit status 1

Setting -opt=0 would allow my build to complete, but any other opt setting including default results in a similar error.  I am not really clear on what fundamentally causes this to happen since my project is rather large and the parsing error only pointed me towards wasm-opt.  I decided to try tinkering with tinygo and found that upgrading the binaryen submodule from version 102 to 114 solved it.  Could that be upgraded for a future tinygo release?

DarkByteBen:
This change upgrades the binaryen submodule from version 102 to version 114.

Resolves #3881

One consideration before merging is that the newer version of binaryen requires a submodule initialized under third_party/googletest.  If this change is fine as is, great, but I'm not sure if there might be other documentation or files that need to account for that.


deadprogram:
Hello @DarkByteBen 

I think there are some other changes needed to fully pick up the update.

https://github.com/tinygo-org/tinygo/blob/dev/.github/workflows/linux.yml#L355
I think this needs to be `git submodule update --init --recursive lib/binaryen` in order to update the submodule's submodules.

https://github.com/tinygo-org/tinygo/blob/dev/.github/workflows/linux.yml#L345-L350
We need to change the cache key to `binaryen-linux-arm-v2`

There are some other similar changes that will be needed for all the GH jobs.

Searching the repo source for `binaryen` should turn up more places like the above.

ayang64:
@deadprogram - This would be useful for getting tinygo to compile under FreeBSD as the current version of the `binaryen` fails to build.

It is a small error -- clang stops at a bitwise OR (|) that was supposed to be a boolean OR (||)  -- but this appears to be fixed in later versions.

deadprogram:
Superseded by #3958 so now closing. Thanks again!

PR #3883, opened 2023-08-28
Update golang.org/x/tools to v0.12.0

QuLogic:
This required handling array and struct zero constants, due to https://github.com/golang/tools/commit/71ea8f168c160da91116b4ddbd577a8fc95aa334. I implemented that the same way as channels, which I hope is correct. At least, it appears to work for tests that I am able to run, but we'll see how it is with the more extensive checks on CI.

This also triggers an update of `golang.org/x/sys`, but that is only used for a constant and should be fine.


aykevl (approved):
This looks correct.

Also, updating x/tools/go/ssa is probably a good idea, we haven't updated that dependency in quite a while.

@QuLogic you might also consider updating to v0.13.0, which was released a few days ago.

deadprogram:
Thanks for the update @QuLogic and to @aykevl for review.

Now merging.

PR #3885, opened 2023-09-01
runtime: refactor markGlobals to findGlobals

aykevl:
Instead of `markGlobals` calling `markRoots` unconditionally (which doesn't make sense for `-gc=none` and `-gc=leaking`), provide `markRoots` as a callback function.

This is in preparation for `-gc=boehm`, where the previous design is even more awkward and a callback makes far more sense.

I've tested the size impact using `make smoketest XTENSA=0`. There is none, except for two cases:

  * One with `-opt=0` so const-propagation for the callback didn't take place.
  * One other on AVR, I don't know why but as it's only 16 bytes in a very specific case I'm going to assume it's just a random change in compiler output that caused a size difference.

I made a separate PR out of this because although this is a relatively small change, there's certainly a chance it breaks things.

### DIFF
--- a/src/runtime/gc_blocks.go
+++ b/src/runtime/gc_blocks.go
@@ -420,7 +420,7 @@ func runGC() (freeBytes uintptr) {
 
 	// Mark phase: mark all reachable objects, recursively.
 	markStack()
-	markGlobals()
+	findGlobals(markRoots)
 
 	if baremetal && hasScheduler {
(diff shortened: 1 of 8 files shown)


kenbell:
lgtm

deadprogram:
Thanks for the refactor @aykevl and to @kenbell for review. Now merging.

aykevl:
See #4385 for the new GC

PR #3889, opened 2023-09-03
machine/usb/adc/midi: improve implementation

deadprogram:
This PR changes `machine/usb/adc/midi` to improve the implementation to include several new messages such as program changes and pitch bend.

It also changes the functions to set handlers to be named SetTxHandler() and SetRxHandler(). The previous function `SetHandler()` has been annotated to indicate that it is now deprecated. 

The PR also adds error handling for invalid parameter values such as MIDI channel.

There is however a somewhat breaking change to the current implementation, in that we now use the typical MIDI user system of counting MIDI channels from 1-16 instead of from 0-15 as the lower level USB-MIDI API itself expects.

### DIFF
--- a/src/machine/usb/adc/midi/messages.go
+++ b/src/machine/usb/adc/midi/messages.go
@@ -1,19 +1,233 @@
 package midi
 
-// NoteOn sends a note on message.
-func (m *midi) NoteOn(cable, channel uint8, note Note, velocity uint8) {
-	m.msg[0], m.msg[1], m.msg[2], m.msg[3] = (cable&0xf<<4)|0x9, 0x90|(channel&0xf), byte(note)&0x7f, velocity&0x7f
-	m.Write(m.msg[:])
+import (
+	"errors"
+)
+
+// From USB-MIDI section 4.1 "Code Index Number (CIN) Classifications"
+const (
+	CINSystemCommon2   = 0x2
+	CINSystemCommon3   = 0x3
+	CINSysExStart      = 0x4
+	CINSysExEnd1       = 0x5
+	CINSysExEnd2       = 0x6
+	CINSysExEnd3       = 0x7
+	CINNoteOff         = 0x8
+	CINNoteOn          = 0x9
+	CINPoly            = 0xA
+	CINControlChange   = 0xB
+	CINProgramChange   = 0xC
+	CINChannelPressure = 0xD
+	CINPitchBendChange = 0xE
+	CINSingleByte      = 0xF
+)
+
+// Standard MIDI channel messages
+const (
+	MsgNoteOff           = 0x80
+	MsgNoteOn            = 0x90
+	MsgPolyAftertouch    = 0xA0
+	MsgControlChange     = 0xB0
+	MsgProgramChange     = 0xC0
+	MsgChannelAftertouch = 0xD0
+	MsgPitchBend         = 0xE0
+	MsgSysExStart        = 0xF0
+	MsgSysExEnd          = 0xF7
+)
+
+// Standard MIDI control change messages
+const (
+	CCModulationWheel       = 0x01
+	CCBreathController      = 0x02
+	CCFootPedal             = 0x04
+	CCPortamentoTime        = 0x05
+	CCDataEntry             = 0x06
+	CCVolume                = 0x07
+	CCBalance               = 0x08
+	CCPan                   = 0x0A
+	CCExpression            = 0x0B
+	CCEffectControl1        = 0x0C
+	CCEffectControl2        = 0x0D
+	CCGeneralPurpose1       = 0x10
+	CCGeneralPurpose2       = 0x11
+	CCGeneralPurpose3       = 0x12
+	CCGeneralPurpose4       = 0x13
+	CCBankSelect            = 0x20
+	CCModulationDepthRange  = 0x21
+	CCBreathControllerDepth = 0x22
+	CCFootPedalDepth        = 0x24
+	CCEffectsLevel          = 0x5B
+	CCTremeloLevel          = 0x5C
+	CCChorusLevel           = 0x5D
+	CCCelesteLevel          = 0x5E
+	CCPhaserLevel           = 0x5F
+	CCDataIncrement         = 0x60
+	CCDataDecrement         = 0x61
+	CCNRPNLSB               = 0x62
+	CCNRPNMSB               = 0x63
+	CCRPNLSB                = 0x64
+	CCRPNMSB                = 0x65
+	CCAllSoundOff           = 0x78
+	CCResetAllControllers   = 0x79
+	CCAllNotesOff           = 0x7B
+	CCChannelVolume         = 0x7F
+)
+
+var (
+	errInvalidMIDICable        = errors.New("invalid MIDI cable")
+	errInvalidMIDIChannel      = errors.New("invalid MIDI channel")
+	errInvalidMIDIVelocity     = errors.New("invalid MIDI velocity")
+	errInvalidMIDIControl      = errors.New("invalid MIDI control number")
+	errInvalidMIDIControlValue = errors.New("invalid MIDI control value")
+	errInvalidMIDIPatch        = errors.New("invalid MIDI patch number")
+	errInvalidMIDIPitchBend    = errors.New("invalid MIDI pitch bend value")
+	errInvalidMIDISysExData    = errors.New("invalid MIDI SysEx data")
+)
+
+// NoteOn sends a channel note on message.
+// The cable parameter is the cable number 0-15.
+// The channel parameter is the MIDI channel number 1-16.
+func (m *midi) NoteOn(cable, channel uint8, note Note, velocity uint8) error {
+	switch {
+	case cable > 15:
+		return errInvalidMIDICable
+	case channel == 0 || channel > 16:
+		return errInvalidMIDIChannel
+	case velocity > 127:
+		return errInvalidMIDIVelocity
+	}
+
+	m.msg[0], m.msg[1], m.msg[2], m.msg[3] = (cable&0xf<<4)|CINNoteOn, MsgNoteOn|(channel-1&0xf), byte(note)&0x7f, velocity&0x7f
+	_, err := m.Write(m.msg[:])
+	return err
 }
 
-// NoteOff sends a note off message.
-func (m *midi) NoteOff(cable, channel uint8, note Note, velocity uint8) {
-	m.msg[0], m.msg[1], m.msg[2], m.msg[3] = (cable&0xf<<4)|0x8, 0x80|(channel&0xf), byte(note)&0x7f, velocity&0x7f
-	m.Write(m.msg[:])
+// NoteOff sends a channel note off message.
+// The cable parameter is the cable number 0-15.
+// The channel parameter is the MIDI channel number 1-16.
+func (m *midi) NoteOff(cable, channel uint8, note Note, velocity uint8) error {
+	switch {
+	case cable > 15:
+		return errInvalidMIDICable
+	case channel == 0 || channel > 16:
+		return errInvalidMIDIChannel
+	case velocity > 127:
+		return errInvalidMIDIVelocity
+	}
+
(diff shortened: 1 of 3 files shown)


deadprogram:
UPDATE: I renamed the `SendCC()` function to something more meaningful, added constant values for continuous controller message types, and also added a `SysEx()` function.

src/machine/usb/adc/midi/midi.go
```diff
@@ -53,24 +54,40 @@ func newMidi() *midi {
...
-	for i = 0; i < len(b); i += 4 {
-		m.tx(b[i : i+4])
+	s, e := 0, 0
+	for s = 0; s < len(b); s += 4 {
+		e = s + 4
+		if e > len(b) {
+			e = len(b)
+		}
+
+		m.tx(b[s:e])
```
HattoriHanzo031, line 88: I have no experience with MIDI, but just saw micro optimisation for this loop :)  
```suggestion
	e := 4
	for ; e < len(b); e += 4 {
		m.tx(b[e-4 : e])
	}
	if e < len(b)+4 {
		m.tx(b[e-4:])
	}
	return len(b), nil
```

src/machine/usb/adc/midi/midi.go
```diff
@@ -79,7 +96,11 @@ func (m *midi) sendUSBPacket(b []byte) {
 }
 
 // from BulkIn
-func (m *midi) Handler() {
+func (m *midi) TxHandler() {
+	if m.txHandler != nil {
+		m.txHandler()
+	}
```
HattoriHanzo031, line 102: Is it better for this user handler to be before instead of after sending packets over USB so it does not introduce latency. Should it be used to modify something prior to sending or just as signalling to the user that packet was sent?
Again, I don't know how all this works so I'm not suggesting, just asking :)

src/machine/usb/adc/midi/messages.go
```diff
@@ -1,19 +1,265 @@
...
+			m.msg[2], m.msg[3] = data[i+1], data[i+2]
+
+			if _, err := m.Write(m.msg[:]); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
 }
```
HattoriHanzo031: Just some refactoring. Not sure If the code is 100% correct, but in concept I think it would be clearer if writing is separated in 3 segments
```suggestion
func (m *midi) SysEx(cable uint8, data []byte) error {
	if cable > 15 {
		return errInvalidMIDICable
	}

	if len(data) < 3 {
		return errInvalidMIDISysExData
	}

	// write start
	m.msg[0], m.msg[1] = (cable&0xf<<4)|CINSysExStart, MsgSysExStart
	m.msg[2], m.msg[3] = data[0], data[1]
	if _, err := m.Write(m.msg[:]); err != nil {
		return err
	}

	// write middle
	i := 2
	for ; i < len(data)-2; i += 3 {
		m.msg[0], m.msg[1] = (cable&0xf<<4)|CINSysExStart, data[i]
		m.msg[2], m.msg[3] = data[i+1], data[i+2]
		if _, err := m.Write(m.msg[:]); err != nil {
			return err
		}
	}

	// write end
	switch len(data) - i {
	case 2:
		m.msg[0], m.msg[1] = (cable&0xf<<4)|CINSysExEnd3, data[i]
		m.msg[2], m.msg[3] = data[i+1], MsgSysExEnd
	case 1:
		m.msg[0], m.msg[1] = (cable&0xf<<4)|CINSysExEnd2, data[i]
		m.msg[2], m.msg[3] = MsgSysExEnd, 0
	case 0:
		m.msg[0], m.msg[1] = (cable&0xf<<4)|CINSysExEnd1, MsgSysExEnd
		m.msg[2], m.msg[3] = 0, 0
	}
	if _, err := m.Write(m.msg[:]); err != nil {
		return err
	}

	return nil
}
```

HattoriHanzo031 (commented):
Since I have no knowledge of MIDI I just made some refactoring suggestions

src/machine/usb/adc/midi/midi.go
sago35: Basically, it should be called when the transmission is completed.
It would be better to treat refactoring of the part that originally existed as a separate PR.

src/machine/usb/adc/midi/midi.go
sago35: Thanks for the suggestion. @HattoriHanzo031 
Perhaps both are good codes.
I think the original code does not need to be changed as the loops look much clearer.

src/machine/usb/adc/midi/messages.go
sago35: I think the new code is better.
However, just looking at it, I am not sure if there are any off-by-one errors.

sago35: I think this proposal should also be in a separate PR?

sago35 (approved):
LGTM

I think it would be OK to accept HattoriHanzo031's proposal, but it would be better to have another PR created.

src/machine/usb/adc/midi/midi.go
```diff
@@ -79,7 +96,11 @@ func (m *midi) sendUSBPacket(b []byte) {
 }
 
 // from BulkIn
-func (m *midi) Handler() {
+func (m *midi) TxHandler() {
```
soypat, line 99: Do we want to call this method `TxHandler`? And even more importantly- do we want to export it? What is there to gain?

src/machine/usb/adc/midi/midi.go
```diff
@@ -53,24 +54,40 @@ func newMidi() *midi {
...
+	m.SetRxHandler(rxHandler)
+}
+
+// SetRxHandler sets the handler function for incoming MIDI messages.
+func (m *midi) SetRxHandler(rxHandler func([]byte)) {
 	m.rxHandler = rxHandler
 }
 
+// SetTxHandler sets the handler function for outgoing MIDI messages.
+func (m *midi) SetTxHandler(txHandler func()) {
```
soypat, line 76: Maybe the TxHandler could receive the sent bytes? It would allow for users to inspect outgoing messages and react based on them, i.e: RGB LEDs color based on pitch?

src/machine/usb/adc/midi/messages.go
```diff
@@ -1,19 +1,265 @@
...
+		return errInvalidMIDICable
+	}
+
+	if channel == 0 || channel > 16 {
+		return errInvalidMIDIChannel
+	}
+
+	if patch > 127 {
+		return errInvalidMIDIPatch
+	}
```
soypat: nit: use switch with empty statement.

src/machine/usb/adc/midi/messages.go
```diff
@@ -1,19 +1,265 @@
...
+		return errInvalidMIDICable
+	}
+
+	if channel == 0 || channel > 16 {
+		return errInvalidMIDIChannel
+	}
+
+	if bend > 0x3FFF {
+		return errInvalidMIDIPitchBend
+	}
```
soypat: nit: same as above here

src/machine/usb/adc/midi/messages.go
```diff
@@ -1,19 +1,265 @@
...
+// The data slice should not include the SysEx start (0xF0) or
+// end (0xF7) bytes, only the data in between.
+func (m *midi) SysEx(cable uint8, data []byte) error {
+	if cable > 15 {
+		return errInvalidMIDICable
+	}
+
+	if len(data) < 3 {
+		return errInvalidMIDISysExData
+	}
```
soypat: if adding switch above do same here

src/machine/usb/adc/midi/messages.go
```diff
@@ -1,19 +1,265 @@
...
+		return errInvalidMIDIChannel
+	}
+
+	if control > 127 {
+		return errInvalidMIDIControl
+	}
+
+	if value > 127 {
+		return errInvalidMIDIControlValue
+	}
```
soypat: switch nit

src/machine/usb/adc/midi/messages.go
```diff
@@ -1,19 +1,265 @@
...
+		return errInvalidMIDICable
+	}
+
+	if channel == 0 || channel > 16 {
+		return errInvalidMIDIChannel
+	}
+
+	if velocity > 127 {
+		return errInvalidMIDIVelocity
+	}
```
soypat: switch nit

src/machine/usb/adc/midi/messages.go
```diff
@@ -1,19 +1,265 @@
...
```
soypat: switch nit

soypat (approved):
Looks good as-is and other comments in here are great too. I've left a few minor observations which are not directly tied to functionality or critical performance so apply them at author's discretion

src/machine/usb/adc/midi/midi.go
deadprogram: The reason that I put the handler first was in case the developer wanted to change the data being sent. But that is perhaps unrealistic and could introduce some latency. I think I will move the handler after the transmit, as has been suggested here.

src/machine/usb/adc/midi/messages.go
deadprogram: @HattoriHanzo031 I think this is good idea. I ran thru a few paper iterations and it seems correct. Separating into start, middle, and end is a better way to structure the code.

src/machine/usb/adc/midi/midi.go
deadprogram: This is already the naming for transmit handler function in USB HID implementations elsewhere in TinyGo.

src/machine/usb/adc/midi/midi.go
deadprogram: That is good idea, but I think I will leave it for a future implementation for now.

src/machine/usb/adc/midi/messages.go
deadprogram: Good idea!

src/machine/usb/adc/midi/messages.go
deadprogram: Ditto.

src/machine/usb/adc/midi/messages.go
deadprogram: Natch.

src/machine/usb/adc/midi/messages.go
deadprogram: Same.

src/machine/usb/adc/midi/messages.go
deadprogram: Indeed.

src/machine/usb/adc/midi/messages.go
deadprogram: :smile_cat:

src/machine/usb/adc/midi/midi.go
HattoriHanzo031: @sago35 performance gain is probably negligible so if you think original code is more clear then it is best to leave it as is

src/machine/usb/adc/midi/midi.go
```diff
@@ -79,12 +96,16 @@ func (m *midi) sendUSBPacket(b []byte) {
...
+func (m *midi) TxHandler() {
 	m.waitTxc = false
 	if b, ok := m.buf.Get(); ok {
 		m.waitTxc = true
 		m.sendUSBPacket(b)
 	}
+
+	if m.txHandler != nil {
+		m.txHandler()
+	}
```
sago35: Oops, I made a mistake.

`func (m *midi) TxHandler()` is called when a previous USB transmission has completed.
Line 103 is the next transmission.

Therefore, m.txHandler() should be called immediately after entering this function.

sago35: Calling immediately after entering a function can add a little more data for the next transmission.
If the call is made at the end of a function, USB communication may improve performance slightly, but only by a small percentage.

So it is probably better to call from the beginning of the function this time.

deadprogram: OK change made :smiley_cat: 

Thanks for helping look this over @sago35 it is appreciated!

deadprogram:
Thank you very much everyone for your feedback and reviews. Now merging.

Issue #3890, opened 2023-09-03
Faulty WASM During Initialization


cryi:
## Environment:
- tinygo version: 0.29.0
- operating System: windows and wsl2 (ubuntu 22.04)
- architecture: WASM

## Description:
Compiling the given code for WASM using tinygo  produces incorrect behavior. The function checksum() returns a value that doesn't match the expected one only when it is called during the initialization of global variables. The issue appears to be related to the function panicOnError() that is declared in the code but is not directly used for the failing part.

## Minimal Reproducible Example:
```go
//go:build js && wasm

package main

import (
	"crypto/sha256"
	"errors"
	"fmt"
)

var (
	ProtoAlpha = MustParseProtocolHash("ProtoALphaALphaALphaALphaALphaALphaALphaALphaDdp3zK")
	input      = []byte{2, 170, 11, 205, 123, 44, 173, 205, 135, 236, 176, 213, 197, 3, 48, 251, 89, 254, 237, 116, 50, 191, 254, 206, 222, 138, 9, 162, 184, 109, 21, 39, 198, 165}
)

func checksum(input []byte) (cksum [4]byte) {
	h := sha256.Sum256(input)
	h2 := sha256.Sum256(h[:])
	copy(cksum[:], h2[:4])
	return
}

func test() error {
	expected := [4]byte{220, 247, 41, 156}
	if checksum(input) /* [220 247 41 156]*/ != expected {
		return errors.New("checksum mismatch")
	}
	return nil
}

func MustParseProtocolHash(src string) (h [32]byte) {
	err := test()
	panicOnError(err)
	// uncomment this and code works
	// if err != nil {
	// 	panic(err)
	// }
	return h
}

func panicOnError(err error) {
	if err != nil {
		panic(err)
	}
}

func main() {
	// trigger init of tezos module
	fmt.Println(ProtoAlpha)

	// this works too
	// fmt.Println(MustParseProtocolHash("ProtoALphaALphaALphaALphaALphaALphaALphaALphaDdp3zK"))
}
```

## Steps to reproduce:
1. compile the provided code using tinygo targeting WASM. (`tinygo clean && tinygo build -o main.wasm -target wasm -no-debug`)
2. run the resulting WASM binary. 
3. observe that the checksum generated during global variable initialization doesn't match the expected value.

## Expected behavior:

The checksum generated should match the expected value `[220 247 41 156]`.

## Actual behavior:

[...]

PR #3901, opened 2023-09-09
interp: don't copy unknown values in runtime.sliceCopy

aykevl:
This was bug https://github.com/tinygo-org/tinygo/issues/3890. See the diff for details.
Essentially, it means that `copy` in the interpreter was copying data that wasn't known yet, or copying data into a slice that could be read externally after the interp pass.


deadprogram (approved):
Thank you very much for the fix @aykevl now merging.

PR #3902, opened 2023-09-10
targets: add GoBadge target as alias for PyBadge

deadprogram:
This PR adds a new target `gobadge` as an alias for PyBadge to make life easier for hack session users.


aykevl (approved):

deadprogram:

Issue #3478, opened 2023-02-24
build: circleCI backward compatibility build failing


deadprogram:
The build on CircleCI maintained for backwards compatibility with older versions of Go and LLVM is failing pretty consistently. It can be restarted from the failure, and will fail it all the way thru after a couple of tries.

See https://app.circleci.com/pipelines/github/tinygo-org/tinygo/8927/workflows/5580161d-fc1b-4e32-8761-800a9850e3eb/jobs/39754?invite=true#step-118-8

deadprogram:
This was addressed more or less during the last release cycle, so now closing.

PR #3903, opened 2023-09-10
Board support for Adafruit Gemma M0

trulede:
Board support for Adafruit Gemma M0.

Derived from similar boards (e.g. Itsybitsy M0). Pin and usb device adjustments only.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -558,6 +558,8 @@ endif
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=trinket-m0          examples/blinky1
 	@$(MD5SUM) test.hex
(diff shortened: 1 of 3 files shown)


deadprogram:
Hello @trulede thank you for adding this board.

Can you please also add a smoke test for it similar to https://github.com/tinygo-org/tinygo/blob/dev/Makefile#L559-L560

Thanks!

trulede:
@deadprogram  I'm not sure, but it seems that "test-all" failed for some other reason. The gemma-m0 build I added seemed to run OK.

deadprogram:
@trulede yes that is a spurious error, please see https://github.com/tinygo-org/tinygo/issues/3478

deadprogram (approved):
Thank you for the addition @trulede now squash/merging.

PR #3904, opened 2023-09-11
interp: add 'max depth' parameter to limit cost

kenbell:
Prototype of limiting max depth for interp evaluation (number of functions) as a way of keeping it's cost-to-execute down.

This should still have reproducible builds since the code structure (function call chains) is fixed for any given code being compiled.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -446,7 +446,7 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 				if pkgInit.IsNil() {
 					panic("init not found for " + pkg.Pkg.Path())
 				}
-				err := interp.RunFunc(pkgInit, config.Options.InterpTimeout, config.DumpSSA())
+				err := interp.RunFunc(pkgInit, config.Options.InterpTimeout, config.Options.InterpMaxDepth, config.DumpSSA())
 				if err != nil {
 					return err
 				}
@@ -1043,7 +1043,7 @@ func createEmbedObjectFile(data, hexSum, sourceFile, sourceDir, tmpdir string, c
 // needed to convert a program to its final form. Some transformations are not
 // optional and must be run as the compiler expects them to run.
 func optimizeProgram(mod llvm.Module, config *compileopts.Config) error {
-	err := interp.Run(mod, config.Options.InterpTimeout, config.DumpSSA())
+	err := interp.Run(mod, config.Options.InterpTimeout, config.Options.InterpMaxDepth, config.DumpSSA())
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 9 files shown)


dgryski:
See a previous approach on limiting interp time here: https://github.com/tinygo-org/tinygo/pull/2384

aykevl:
I think this is actually a really interesting solution! It's reproducible which is good, and at least according to the size difference bot there doesn't seem to be much of an impact on binary size for small programs.

(I do need to review the code still, though).

@dgryski what do you think? Also, would this help where you ran into interp limitations?

dgryski:
Interp has problems with the `strings` and `bytes` unit tests. We encounter interp issues mainly trying to construct regexps at the top level. I'll see if this patch solves these for us.

kenbell:
I've been thinking about some additional 'constraints' we could place...

### Limit Instruction Count
Only evaluate X times the number of instructions in a given function.  E.g. if a function is 200 instructions, only evaluate 20,000 instructions before declaring the function needs to be executed at run time.  This could catch gross loop unrolling, like the tests which are iterating through every possible Unicode code-point to test character classification.

### Limit Allocations
If more than X bytes allocated during the evaluation of a function, declare it to evaluate at run time.

### Others?
So long as the conditions are repeatable (based on artifacts of the code, not on asynchronous stuff like wall time) we could look for other 'heuristics'....   maybe the number of function calls made?

dgryski:
Huh, interesting update.  So `strings` and `bytes` no longer fail due to interp issues.  

Previously we would time out trying to execute https://github.com/golang/go/blob/master/src/strings/strings_test.go#L1814 at compile time.  I wonder what changed since they both pass now on `dev` without this patch.

Off to test regexps next...

dgryski:
Still can't run `cloudflare/ahocorasick'
```
~/go/src/github.com/cloudflare/ahocorasick $ time tinygo test
^C

real	5m22.539s
user	22m45.512s
sys	0m37.183s
```

but this regexp-at-compile-time example runs now:
```
~/go/src/github.com/dgryski/bug/re $ tinygo run main.go
true
true
false
false
~/go/src/github.com/dgryski/bug/re $ cat main.go
package main

import (
	"fmt"
	"regexp"
)

var validID = regexp.MustCompile(`^[a-z]+\[[0-9]+\]$`)

func main() {
	fmt.Println(validID.MatchString("adam[23]"))
	fmt.Println(validID.MatchString("eve[7]"))
	fmt.Println(validID.MatchString("Job[48]"))
	fmt.Println(validID.MatchString("snakey"))
}
```

(However, it *also* seems to run on the current `dev` branch -- what else did we change?)

kenbell:
@dgryski - I've pushed a change that 'fixes' cloudflare/ahocorasick - I think it amounts to interp is many, many, many times slower than native execution.  The latest change adds a `-interp-maxinstr` (default 100_000) that if interp runs more than N instructions evaluating a function, it kicks it back to execute at runtime.

I really doubt this will impact baremetal targets very much since it's relatively unlikely this scenario of 'N' nested loops in a variable initializer will crop-up.

PR #3905, opened 2023-09-11
machine/rp2040: always use the USB device enum fix

deadprogram:
This PR modifies machine/rp2040 to always use the USB device enum fix, even in chips that supposedly have the HW fix.

Additionally, correct the amount of time spent waiting after USB reset, based on advice from @sago35 that I misunderstood in https://github.com/tinygo-org/tinygo/pull/3821


sago35 (approved):
LGTM

deadprogram:
Thank you for review @sago35 now merging.

Issue #3893, opened 2023-09-06
crash on interp timeout


dkegel-fastly:
I can reproduce this on both mac and linux.   I boiled it down on mac to this:

main.go:
```
package main
func main() {
	println("hello, world")
}
```

go.mod:
```
module main

go 1.21.0
```

To reproduce:
```
$ go mod tidy
$ tinygo version
tinygo version 0.29.0 darwin/amd64 (using go version go1.21.0 and LLVM version 15.0.0)

$ tinygo build -interp-timeout=0s
  
# main
/Users/dkegel/tobug: interp: running for more than 0s, timing out (executed calls: 1)
  SIGSEGV: segmentation violation
PC=0x10423f89c m=11 sigcode=1
signal arrived during cgo execution

goroutine 1 [syscall]:
runtime.cgocall(0x1003e2960, 0xc0006e9460)
	/Users/runner/hostedtoolcache/go/1.21.0/x64/src/runtime/cgocall.go:157 +0x4b fp=0xc0006e9438 sp=0xc0006e9400 pc=0x10000934b
tinygo.org/x/go-llvm._Cfunc_LLVMDumpValue(0x600001718180)
	_cgo_gotypes.go:4906 +0x3f fp=0xc0006e9460 sp=0xc0006e9438 pc=0x10024c83f
main.printCompilerError.Value.Dump.func1({0xc000120400?})
	/Users/runner/go/pkg/mod/tinygo.org/x/go-llvm@v0.0.0-20221028183034-8341240c0b32/ir.go:722 +0x39 fp=0xc0006e9498 sp=0xc0006e9460 pc=0x1003b6579
tinygo.org/x/go-llvm.Value.Dump(...)
	/Users/runner/go/pkg/mod/tinygo.org/x/go-llvm@v0.0.0-20221028183034-8341240c0b32/ir.go:722
main.printCompilerError(0x104680bc0, {0x1046f2840?, 0xc00018c000?})
	/Users/runner/work/tinygo/tinygo/main.go:1293 +0x1d6 fp=0xc0006e9748 sp=0xc0006e9498 pc=0x1003b5a96
main.handleCompilerError({0x1046f2840?, 0xc00018c000?})
	/Users/runner/work/tinygo/tinygo/main.go:1326 +0x2f fp=0xc0006e9770 sp=0xc0006e9748 pc=0x1003b65cf
main.main()
...

deadprogram:
[...]

PR #3906, opened 2023-09-11
build: #3893 do not return LLVM structs from Build

kenbell:
This looses a bit of info about the instruction causing the failure - but since Dump can't be used after builder.Build is finished, this is probably the best available for now?

Maybe there's a way to enhance go-llvm to provide more detail in the `String` function?

### DIFF
--- a/interp/compiler.go
+++ b/interp/compiler.go
@@ -49,7 +49,10 @@ func (inst *instruction) String() string {
 		operands[i] = op.String()
 	}
 
-	name := instructionNameMap[inst.opcode]
+	name := ""
+	if int(inst.opcode) < len(instructionNameMap) {
+		name = instructionNameMap[inst.opcode]
+	}
 	if name == "" {
 		name = "<unknown op>"
 	}
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -57,16 +57,14 @@ func runTest(t *testing.T, pathPrefix string) {
 	if err != nil {
 		if err, match := err.(*Error); match {
 			println(err.Error())
-			if !err.Inst.IsNil() {
-				err.Inst.Dump()
-				println()
+			if len(err.Inst) != 0 {
+				println(err.Inst)
 			}
 			if len(err.Traceback) > 0 {
 				println("\ntraceback:")
 				for _, line := range err.Traceback {
 					println(line.Pos.String() + ":")
-					line.Inst.Dump()
-					println()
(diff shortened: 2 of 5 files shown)


dgryski (commented):
I don't know this code but the reasoning LGTM.

kenbell:
pushed a change to fix `make test`

deadprogram:
@kenbell looks like CI still failing https://github.com/tinygo-org/tinygo/actions/runs/6156646704/job/16705747355?pr=3906#step:16:32

kenbell:
@deadprogram - thanks, yeah - looks like it exposes a latent issue in converting unknown opcodes to strings.  I've pushed a change to fix that up.

deadprogram:
@aykevl what do you think?

interp/compiler.go
```diff
@@ -49,7 +49,10 @@ func (inst *instruction) String() string {
 		operands[i] = op.String()
 	}
 
-	name := instructionNameMap[inst.opcode]
+	name := ""
+	if int(inst.opcode) < len(instructionNameMap) {
+		name = instructionNameMap[inst.opcode]
+	}
```
aykevl, line 55: If this fails, that's most certainly a bug and it's rather unlikely this actually happens (it's rare that new instructions are added to LLVM, and it's unlikely they're used by TinyGo right away).

...but of course, adding an extra check won't harm here.

aykevl (approved):
I assume this fails because the modules are deallocated at the time the values are dumped.

There is a way to get the same value that was printed before using `.Dump()`, but that'll need a new go-llvm API addition. For now, this fix works and prints the most important information.

interp/interp_test.go
```diff
@@ -57,16 +57,14 @@ func runTest(t *testing.T, pathPrefix string) {
 	if err != nil {
 		if err, match := err.(*Error); match {
 			println(err.Error())
-			if !err.Inst.IsNil() {
-				err.Inst.Dump()
-				println()
+			if len(err.Inst) != 0 {
+				println(err.Inst)
```
aykevl, line 61: Oh and these need to be changed to `t.Log` or something eventually (I think the reason I used println before was so the error messages would interleave in a sensible way with `inst.Dump()`, which I think also writes to stderr).

aykevl:
Followup PR: https://github.com/tinygo-org/tinygo/pull/3921
I've added a `String` method to go-llvm so that the previous stack trace is back.

PR #3908, opened 2023-09-14
atsamd21, atsamd51: add support for USB INTERRUPT OUT

sago35:
This PR adds `INTERRUPT OUT` support to the following microcontrollers

* atsamd21
* atsamd51

Setting of `INTERRUPT OUT` is almost the same as `BULK OUT`.
The only difference is the parameter at `setEPCFG()`.

Note that rp2040 and nrf52840 were already supported.


sago35:
Once this PR is merged, Vial-support will work on atsamd21 and atsamd51.

https://github.com/sago35/tinygo-keyboard

sago35:
If this PR is merge, pybadge/gobadge can also use tinygo-keyboard with vial.

deadprogram:
Great work @sago35 thank you! Now merging.

Issue #3900, opened 2023-09-08
WASM: stack overflow when unmarshalling JSON 


cryi:
## Environment
TinyGo version: 0.29.0 (with #3890 patch)
Environment: WASM

## Description

We've encountered a stack overflow bug specifically when targeting WASM using TinyGo. Interestingly, we were unable to replicate the issue on desktop targets.

## Minimal Reproducible Example
```go
package main

import (
	"encoding/json"
	"fmt"
	"syscall/js"
)

var (
	terminateChannel = make(chan struct{})
)

type Person interface {
}

type JsonResponse struct {
	Generated []Person `json:"generated"`
}

func main() {
	mre := js.Global().Get("Object").New()
	js.Global().Set("mre", mre)
	mre.Set("trigger", js.FuncOf(trigger))
	<-terminateChannel
}

func trigger(this js.Value, args []js.Value) interface{} {
	// removing any level from the JSON below solves the stack overflow issue
	data := `{
		"generated": [
			{
				"metadata": {
					"person_records": {
						"errors": [
							{
								"with": {
									"int": "1004"
								}
							}
						]
					}
				}
			}
		]
	}`
	var response JsonResponse
	err := json.Unmarshal([]byte(data), &response)
	if err != nil {
		panic(err)
	}

	fmt.Println(response) // the program prints this and then experiences a stack overflow
	terminateChannel <- struct{}{}
	return nil
}
```
From JavaScript, it's invoked as:
```js
const tezpay = globalThis.mre
tezpay.trigger()
```

## Steps to reproduce:
1. compile the provided code using tinygo targeting WASM. (`tinygo clean && tinygo build -o mre.wasm -target wasm -no-debug mre.go`)
2. Include `https://github.com/tinygo-org/tinygo/blob/release/targets/wasm_exec.js` in your index.html
3. Run the resulting WASM binary.
```js
async function wasmBrowserInstantiate(wasmModuleUrl, importObject) => {
	let response = undefined;

	if (!importObject) {
		importObject = {
			env: {
				abort: () => console.log("Abort!")
			}
		};
	}


	if (WebAssembly.instantiateStreaming) {
		response = await WebAssembly.instantiateStreaming(
			fetch(wasmModuleUrl),
			importObject
		);
	} else {
		const fetchAndInstantiateTask = async () => {
			const wasmArrayBuffer = await fetch(wasmModuleUrl).then(response =>
				response.arrayBuffer()
			);
[...]

PR #3911, opened 2023-09-17
targets: increase default stack size to 32k for wasi/wasm targets

deadprogram:
This PR increases the default stack size to 32k for wasi/wasm targets, including the Go compatible `wasip1` target.

Intended to address issues like https://github.com/tinygo-org/tinygo/issues/3900 where a larger default avoids them.


aykevl (approved):
Seems reasonable to me. The low default stack size is important for embedded systems with little memory, but doesn't really apply to WebAssembly.

aykevl (approved):

deadprogram:

Issue #3532, opened 2023-03-08
Race condition


aykevl:
We have another race condition, this time it appears to be a bit more serious:

```
==================
WARNING: DATA RACE
Read at 0x00c00056e3b8 by goroutine 44:
  go/types.computeInterfaceTypeSet()
      /usr/local/go/src/go/types/typeset.go:155 +0x44
  go/types.(*Interface).typeSet()
      /usr/local/go/src/go/types/interface.go:29 +0x6c4
  go/types.lookupFieldOrMethod()
      /usr/local/go/src/go/types/lookup.go:184 +0x704
  go/types.LookupFieldOrMethod()
      /usr/local/go/src/go/types/lookup.go:66 +0xc0
  golang.org/x/tools/go/ssa.(*Function).selection()
      /home/ayke/pkg/mod/golang.org/x/tools@v0.5.1-0.20230114154351-e035d0c426c8/go/ssa/func.go:68 +0x1a0
  golang.org/x/tools/go/ssa.(*builder).setCallFunc()
      /home/ayke/pkg/mod/golang.org/x/tools@v0.5.1-0.20230114154351-e035d0c426c8/go/ssa/builder.go:985 +0x100
  golang.org/x/tools/go/ssa.(*builder).setCall()
      /home/ayke/pkg/mod/golang.org/x/tools@v0.5.1-0.20230114154351-e035d0c426c8/go/ssa/builder.go:1120 +0x44
  golang.org/x/tools/go/ssa.(*builder).expr0()
      /home/ayke/pkg/mod/golang.org/x/tools@v0.5.1-0.20230114154351-e035d0c426c8/go/ssa/builder.go:704 +0x320c
  golang.org/x/tools/go/ssa.(*builder).expr()
      /home/ayke/pkg/mod/golang.org/x/tools@v0.5.1-0.20230114154351-e035d0c426c8/go/ssa/builder.go:625 +0x1ac
  golang.org/x/tools/go/ssa.(*builder).stmt()
      /home/ayke/pkg/mod/golang.org/x/tools@v0.5.1-0.20230114154351-e035d0c426c8/go/ssa/builder.go:2160 +0x1e0
  golang.org/x/tools/go/ssa.(*builder).stmtList()
      /home/ayke/pkg/mod/golang.org/x/tools@v0.5.1-0.20230114154351-e035d0c426c8/go/ssa/builder.go:946 +0x78
  golang.org/x/tools/go/ssa.(*builder).stmt()
[...]

PR #3915, opened 2023-09-18
Tentative: Force compilation order

### LINKED ISSUE #3895: Panic during compile - index out of range in createFunctionStart()?
This is an intermittent error compiling a small "hello, http" app with tinygo 0.29 [EDIT: also with 0.28.1, see below]:
```
panic: runtime error: index out of range [0] with length 0

goroutine 65 [running]:
github.com/tinygo-org/tinygo/compiler.(*builder).createFunctionStart(0xc01061c300, 0x0)
        /__w/tinygo/tinygo/compiler/compiler.go:1219 +0x11f8
github.com/tinygo-org/tinygo/compiler.(*builder).createFunction(0xc01061c300)
        /__w/tinygo/tinygo/compiler/compiler.go:1308 +0x2c
github.com/tinygo-org/tinygo/compiler.(*compilerContext).getFunction(0xc0147c8000, 0xc01b0fcf00)
        /__w/tinygo/tinygo/compiler/symbol.go:224 +0xdaf
github.com/tinygo-org/tinygo/compiler.(*builder).createFunctionCall(0xc01061c180, 0xc00ffa2bc0)
        /__w/tinygo/tinygo/compiler/compiler.go:1883 +0x906
github.com/tinygo-org/tinygo/compiler.(*builder).createExpr(0xc01061c180, {0x563ba30, 0xc00ffa2b80})
        /__w/tinygo/tinygo/compiler/compiler.go:2047 +0x4e5
github.com/tinygo-org/tinygo/compiler.(*builder).createInstruction(0xc01061c180, {0x563b910?, 0xc00ffa2b80?})
        /__w/tinygo/tinygo/compiler/compiler.go:1460 +0xa37
github.com/tinygo-org/tinygo/compiler.(*builder).createFunction(0xc01061c180)
        /__w/tinygo/tinygo/compiler/compiler.go:1349 +0xacc
github.com/tinygo-org/tinygo/compiler.(*compilerContext).createPackage(0xc0147c8000, {0xc00dc27500?}, 0xc00dcb1180)
        /__w/tinygo/tinygo/compiler/compiler.go:901 +0x8e5
github.com/tinygo-org/tinygo/compiler.CompilePackage({0xc000444aa8?, 0x53?}, 0xc00044e960, 0xc00dcb1180, {0xe?}, 0x55a15af?, 0x9e?)
        /__w/tinygo/tinygo/compiler/compiler.go:312 +0x42a
github.com/tinygo-org/tinygo/builder.Build.func3(0xc00e0ceae0)
        /__w/tinygo/tinygo/builder/build.go:365 +0x1f4
github.com/tinygo-org/tinygo/builder.runJob(0xc00e0ceae0, 0x0?)
        /__w/tinygo/tinygo/builder/jobs.go:222 +0x4d
created by github.com/tinygo-org/tinygo/builder.runJobs in goroutine 1
        /__w/tinygo/tinygo/builder/jobs.go:123 +0x5bd\n": exit status
```

kenbell:
It seems like the dependency logic in Build is broken.  There's a bunch of code tracking actionid jobs - but nothing that seems to ensure a package is actually fully built before packages dependent on it.

In the world of templates, it seems like actually having dependencies fully built is important - we can't just rely on the linker to match up symbols later.

This branch is a quick hack to see if it fixes #3895 and a bunch of other build crashes by forcing compilation ordering.  The build logic almost certainly needs revisiting more thoroughly - do we even need actionidjobs any more?

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -217,6 +217,10 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 	var packageJobs []*compileJob
 	packageActionIDJobs := make(map[string]*compileJob)
 
+	// Keep track of the actual jobs to compile dependencies.  Actual builds
+	// must occur so that template instantiation can complete in dependencies.
+	actualCompileJobs := make(map[string]*compileJob)
+
 	if config.Options.GlobalValues == nil {
 		config.Options.GlobalValues = make(map[string]map[string]string)
 	}
@@ -293,6 +297,7 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 		// Action ID jobs need to know the action ID of all the jobs the package
 		// imports.
 		var importedPackages []*compileJob
+		var actualCompileDeps []*compileJob
 		for _, imported := range pkg.Pkg.Imports() {
 			job, ok := packageActionIDJobs[imported.Path()]
 			if !ok {
@@ -300,6 +305,15 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 			}
 			importedPackages = append(importedPackages, job)
 			actionIDDependencies = append(actionIDDependencies, job)
+			actualCompileDeps = append(actualCompileDeps, actualCompileJobs[imported.Path()])
+		}
+
+		// This is a placeholder to accumulate all of compilation
+		// dependencies into a single dependency.
+		actualCompileDepsJob := &compileJob{
+			description:  "compilation deps for package " + pkg.ImportPath,
+			dependencies: actualCompileDeps,
+			run:          func(j *compileJob) (err error) { return nil },
 		}
 
 		// Create a job that will calculate the action ID for a package compile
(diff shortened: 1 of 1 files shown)


aykevl:
> In the world of templates

I think you mean generics?

> it seems like actually having dependencies fully built is important - we can't just rely on the linker to match up symbols later.

It's slightly more complicated than that. Even if packages are fully built before their dependencies are, there can be race conditions. I still suspect this is a bug in the x/tools/go/ssa package, but writing a reproducer is hard so that's why I've been postponing it. For more information, see: https://github.com/tinygo-org/tinygo/issues/3532 (in this reproducer, two packages that import the same dependency result in a race condition).

But to clarify: there is nothing on the TinyGo side that requires the dependencies to be fully built. And in fact, this worked just fine before we added support for generics. Hence why this seems like a x/tools/go/ssa bug to me.

I've looked into this recently again. One fix that appears to be working reliably is building the Go SSA serially, like this:

```diff
@@ -344,13 +347,47 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
                }
                packageActionIDJobs[pkg.ImportPath] = packageActionIDJob
 
+               // Build the SSA for the given package.
+               ssaPkg := program.Package(pkg.Pkg)
+               ssaPkg.Build()
```

...but that slows down the build and also builds packages that don't need to be rebuilt. There must be a better way to do this though, but it's kind of complicated to get right and not do unnecessary work.

kenbell:
Closing as issue fixed by #3920

Issue #3894, opened 2023-09-07
Crash during compile


dkegel-fastly:
This is on linux with tinygo 0.29 compiling a fairly small "hello, http" app.

Seems to be intermittent, but often enough to cause CI traffic jams.

```
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x290 pc=0x5148bb2]

runtime stack:
runtime.throw({0x556e85f?, 0x40?})
        /usr/local/go/src/runtime/panic.go:1047 +0x5d fp=0x7f148c85f880 sp=0x7f148c85f850 pc=0x4f3f1d
runtime.sigpanic()
        /usr/local/go/src/runtime/signal_unix.go:825 +0x3e9 fp=0x7f148c85f8e0 sp=0x7f148c85f880 pc=0x50b1c9

goroutine 182 [syscall]:
runtime.cgocall(0x8517d0, 0xc002979fd0)
        /usr/local/go/src/runtime/cgocall.go:157 +0x5c fp=0xc002979fa8 sp=0xc002979f70 pc=0x4c36bc
tinygo.org/x/go-llvm._Cfunc_LLVMConstInt(0x7f148b916268, 0x0, 0x0)
        _cgo_gotypes.go:3065 +0x4d fp=0xc002979fd0 sp=0xc002979fa8 pc=0x6ef08d
tinygo.org/x/go-llvm.ConstInt.func1({0x853b80?}, 0xc00297a038?, 0x0)
        /go/pkg/mod/tinygo.org/x/go-llvm@v0.0.0-20221028183034-8341240c0b32/ir.go:843 +0x5e fp=0xc00297a010 sp=0xc002979fd0 pc=0x70719e
tinygo.org/x/go-llvm.ConstInt({0x7f148b916268?}, 0xc004766240?, 0x60?)
        /go/pkg/mod/tinygo.org/x/go-llvm@v0.0.0-20221028183034-8341240c0b32/ir.go:843 +0x19 fp=0xc00297a038 sp=0xc00297a010 pc=0x7070f9
github.com/tinygo-org/tinygo/compiler.(*compilerContext).getTypeCode(0xc002840280, {0x55da858, 0xc00507e560?})
        /__w/tinygo/tinygo/compiler/interface.go:444 +0x2f4e fp=0xc00297a568 sp=0xc00297a038 pc=0x7ab4ae
[...]

Issue #3874, opened 2023-08-18
Compiler crash


rockwotj:
We had a compiler crash during a CI run, however it's the only time this has happened, so maybe the stacktrace is helpful, but note this is likely some race condition somewhere, as this generally doesn't happen.

```
/vectorized/bin/tinygo build -o /var/lib/buildkite-agent/builds/buildkite-amd64-builders-i-045ec208102639ae1-1/redpanda/redpanda/vbuild/release/clang/src/go/transform-sdk/internal/testdata/wasi.wasm -target wasi wasi/transform.go
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x80 addr=0x0 pc=0x5e62a2d]
runtime stack:
runtime.throw({0x8ebde5?, 0x7fb81d489fc0?})
/usr/local/go/src/runtime/panic.go:1047 +0x5d fp=0x7fb822e822f0 sp=0x7fb822e822c0 pc=0x255b0fd
runtime.sigpanic()
/usr/local/go/src/runtime/signal_unix.go:825 +0x3e9 fp=0x7fb822e82350 sp=0x7fb822e822f0 pc=0x25723a9
goroutine 152 [syscall]:
runtime.cgocall(0x28e2670, 0xc005183b30)
/usr/local/go/src/runtime/cgocall.go:157 +0x5c fp=0xc005183b08 sp=0xc005183ad0 pc=0x252a89c
tinygo.org/x/go-llvm._Cfunc_LLVMVerifyModule(0x7fb8258b0330, 0x1, 0xc006104130)
_cgo_gotypes.go:9315 +0x4c fp=0xc005183b30 sp=0xc005183b08 pc=0x276292c
tinygo.org/x/go-llvm.VerifyModule.func1({0x276a93f?}, 0x258b0330?, 0x7fb82503b610?)
/home/rockwood/go/pkg/mod/tinygo.org/x/go-llvm@v0.0.0-20221028183034-8341240c0b32/analysis.go:38 +0x7d fp=0xc005183b78 sp=0xc005183b30 pc=0x2762e5d
tinygo.org/x/go-llvm.VerifyModule({0xc00043c446?}, 0x53?)
/home/rockwood/go/pkg/mod/tinygo.org/x/go-llvm@v0.0.0-20221028183034-8341240c0b32/analysis.go:38 +0x45 fp=0xc005183bd0 sp=0xc005183b78 pc=0x2762d25
[...]

PR #3917, opened 2023-09-18
compiler: don't use types in the global context

aykevl:
This usually works by chance, but leads to crashes. So we should never ever do this.

I'm pretty sure this is the crash behind this issue: https://github.com/tinygo-org/tinygo/issues/3894

It may also have caused this crash: https://github.com/tinygo-org/tinygo/issues/3874

I have a suspicion this is also behind the rather crash-prone CircleCI jobs, that we haven't been able to find the source of. But we'll find out soon enough once this fix is merged.

To avoid hitting this issue again in the future, I've created a PR to remove these dangerous functions altogether from the go-llvm API: https://github.com/tinygo-org/go-llvm/pull/54

### DIFF
--- a/compiler/interface.go
+++ b/compiler/interface.go
@@ -147,7 +147,7 @@ func (c *compilerContext) getTypeCode(typ types.Type) llvm.Value {
 				c.addError(token.NoPos, fmt.Sprintf("too many levels of pointers for typecode: %s", typstr))
 			}
 			return llvm.ConstGEP(c.ctx.Int8Type(), ptr, []llvm.Value{
-				llvm.ConstInt(llvm.Int32Type(), 1, false),
+				llvm.ConstInt(c.ctx.Int32Type(), 1, false),
 			})
 		}
 	}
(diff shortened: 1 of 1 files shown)


compiler/interface.go
```diff
@@ -147,7 +147,7 @@ func (c *compilerContext) getTypeCode(typ types.Type) llvm.Value {
 				c.addError(token.NoPos, fmt.Sprintf("too many levels of pointers for typecode: %s", typstr))
 			}
 			return llvm.ConstGEP(c.ctx.Int8Type(), ptr, []llvm.Value{
-				llvm.ConstInt(llvm.Int32Type(), 1, false),
```
aykevl, line 150: @dgryski I believe this was you. Just so you know, the `llvm.*Type()` methods should never be used (see https://github.com/tinygo-org/tinygo/pull/1744 for another instance and https://github.com/tinygo-org/go-llvm/pull/54 where I remove this function).

aykevl:
Checked CircleCI crashes and this appears to be the bug behind at least some of those crashes. Perhaps most (but it's hard to say). At least, I see it crashing on one of the lines where I've made a fix.

dgryski:
Thanks for fixing those.  Whenever I delve into the llvm stuff I always feel out of my depth..

deadprogram:
Now merging, thanks @aykevl and @dgryski for review.

PR #3920, opened 2023-09-20
build: build Go SSA serially [issue 3895]

dkegel-fastly:
From
https://github.com/tinygo-org/tinygo/pull/3915#issuecomment-1724405109

According to Ayke, it slows down the build but seems to reliably fix https://github.com/tinygo-org/tinygo/issues/3895

Seems to help on my real app without slowing build down too much

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -344,6 +344,10 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 		}
 		packageActionIDJobs[pkg.ImportPath] = packageActionIDJob
 
+		// Build the SSA for the given package.
+		ssaPkg := program.Package(pkg.Pkg)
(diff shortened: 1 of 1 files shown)


deadprogram (approved):
It is better to slow the build down slightly and consistently build correctly. We can work on additional optimizations in subsequent releases.

aykevl:
Yeah, I'm afraid I have to agree (even though I don't want to). These crashes are not great.

So a not-so-enthousiastic LGTM from me.

aykevl:
To be clear, I don't know whether this will fix #3895 (but I think it will). It will most likely fix #3532.

The performance overhead is around 15% in my quick testing.

deadprogram:
Now merging, thanks @dkegel-fastly and to @aykevl for the original fix.

PR #3922, opened 2023-09-21
all: release 0.30.0

aykevl:
This is a small release just in time for GopherCon US.

Some of the bigger changes of this release are:

  - switch to LLVM 16
  - fixes for two separate hard-to-reproduce compiler crashes
  - added support for the Adafruit Gemma M0


deadprogram:
:rocket:

PR #3923, opened 2023-09-21
transform: fix bug in StringToBytes optimization pass

aykevl:
Previously, this pass would convert any read-only use of a `runtime.stringToBytes` call to use the original string buffer instead. This is incorrect: if there are any writes to the resulting buffer, none of the slice buffer pointers can be converted to use the original read-only string buffer.

This commit fixes that bug and adds a test to prove the new (correct) behavior.

### DIFF
--- a/transform/rtcalls.go
+++ b/transform/rtcalls.go
@@ -28,32 +28,44 @@ func OptimizeStringToBytes(mod llvm.Module) {
 
 		// strptr is always constant because strings are always constant.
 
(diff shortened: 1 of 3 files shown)


deadprogram:
@aykevl please rebase against latest `dev` now, thanks!

aykevl:
Rebased.

deadprogram:
Thanks for the fix @aykevl now merging

PR #3924, opened 2023-09-21
ci: redo LLVM build on Windows

aykevl:
MinGW got updated, so it looks like we need to do a new LLVM build.


aykevl:
Looks like this does indeed fix the Windows build issue.

deadprogram:

Issue #3926, opened 2023-09-22
compile failed with `undefined: C._Bool`


Patrick0308:
# Environment:
tinygo version: 0.30.0
operating System: macOS
architecture: aarch64

# Description:
 Compilation brakes with an undefined C._Bool

# Steps to reproduce:
Given the following Go-code:
```
package main

/*
  #include "stdbool.h"
  bool test() {
     return true;
  }
*/
import "C"

import "fmt"

func main() {
	fmt.Printf("%s", C._Bool(C.test()))
}
```
Run

`tinygo run main.go`

# Expected behavior:
```
❯ tinygo run main.go
%!s(main._Ctype__Bool=true)%    
```

# Actual behavior:

```
❯ tinygo run main.go
# command-line-arguments
main.go:14:19: undefined: C._Bool
```

aykevl:
Well this led me down a rather deep rabbit hole to improve CGo compatibility: https://github.com/tinygo-org/tinygo/pull/3927

I'll look at the actual bug next.

aykevl:
Here is a fix: https://github.com/tinygo-org/tinygo/pull/3928

Patrick0308:
#3928 is merged

deadprogram:
Reopening and tagging to close on next release.

Patrick0308:
I build dev branch on my own computer. I tested 
```
package main

/*
  #include "stdbool.h"
  bool test() {
     return true;
  }
*/
import "C"

import "fmt"

func main() {
	fmt.Printf("%s", C._Bool(C.test()))
}
```
It's ok.

However I tested https://github.com/fermyon/spin/pull/1794 example but building painc.
The error log is 
```
Building component tinygo-mysql with `/Users/patrick/projects/tinygo/build/tinygo build -target=wasi -gc=leaking -o main.wasm main.go`
# github.com/fermyon/spin/sdk/go/mysql
../../../tinygo/llvm-project/clang/lib/Headers/stdbool.h:20:14: undefined: C._Bool
Error: Build command for component tinygo-mysql failed with status Exited(1)
```

Is it another issue?

Patrick0308:
@aykevl Any idea？

aykevl:
@Patrick0308 my guess is that they're shipping their own TinyGo variant, that is still the latest release (and not the dev branch).

PR #3928, opened 2023-09-22
cgo: add C._Bool type

aykevl:
This fixes https://github.com/tinygo-org/tinygo/issues/3926.

While working on this I've found another bug: if C.bool is referenced from within Go, it isn't available anymore on the C side. This is an existing bug that also applies to float and double, but may be less likely to be triggered there.
This bug is something to be fixed at a later time (it has something to do with preprocessor defines).

### DIFF
--- a/cgo/cgo.go
+++ b/cgo/cgo.go
@@ -88,6 +88,7 @@ var cgoAliases = map[string]string{
 	"C.uintptr_t": "uintptr",
 	"C.float":     "float32",
 	"C.double":    "float64",
+	"C._Bool":     "bool",
 }
 
 // builtinAliases are handled specially because they only exist on the Go side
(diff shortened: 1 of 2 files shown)


dgryski (approved):
LGTM.  File a bug for the other issue mentioned in the PR description?

deadprogram (approved):
Thanks for the addition @aykevl looks good to me.

Patrick0308:
Any plan to release this fix?

aykevl:
@Patrick0308 it will be part of the next release. We don't yet have a plan for when that will be.

aykevl:
@Patrick0308 it will be part of the next release. We don't yet have a plan for when that will be.

PR #3931, opened 2023-09-25
builder: do Go SSA build in a separate job

aykevl:
This means it can run in parallel with other jobs - although it can't run in parallel with itself.

This is an imperfect solution that is a slight improvement in compile time over building the SSA sequentially while buliding the jobs graph. But it paves the way for more efficient solutions.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -281,9 +281,6 @@ func CompilePackage(moduleName string, pkg *loader.Package, ssaPkg *ssa.Package,
 	c.runtimePkg = ssaPkg.Prog.ImportedPackage("runtime").Pkg
 	c.program = ssaPkg.Prog
 
-	// Convert AST to SSA.
-	ssaPkg.Build()
-
 	// Initialize debug information.
 	if c.Debug {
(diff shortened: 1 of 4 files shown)


compiler/compiler.go
```diff
@@ -281,9 +281,6 @@ func CompilePackage(moduleName string, pkg *loader.Package, ssaPkg *ssa.Package,
 	c.runtimePkg = ssaPkg.Prog.ImportedPackage("runtime").Pkg
 	c.program = ssaPkg.Prog
 
-	// Convert AST to SSA.
-	ssaPkg.Build()
-
```
aykevl, line 286: This should ideally have been removed in #3920 and although it doesn't really harm (packages are only built once), it is better to remove it to avoid confusion.

aykevl:
Actually I'm working on an entirely different approach now, let's see whether that works at all...

deadprogram:
> Actually I'm working on an entirely different approach now, let's see whether that works at all...

@aykevl does that mean we should close this PR?

aykevl:
Let's close it for now. I can pick it up later if needed.

PR #3935, opened 2023-09-28
Update makefiles to improve portability

ayang64:
I've been working on building tinygo under FreeBSD. I'm submitting some patches that reflect changes I had to make to accomplish this.

These commits first ensure that we invoke the proper make program when calling make recursively.

The second commit renames the makefiles to reflect that they require GNU Make.


dgryski:
The changes to call `$(MAKE)` are fine, but I don't know if we want to rename the makefiles.  Probably updating the BUILDING doc (as was done in another PR) is sufficient.

ayang64:
> The changes to call `$(MAKE)` are fine, but I don't know if we want to rename the makefiles. Probably updating the BUILDING doc (as was done in another PR) is sufficient.

Just curious:  what are your objections to it?  i'm not looking to debate and I won't drag the discussion out; my rationalle is below.

The reasons i'd like to see it happen are:

* we, in fact, use GNU make to build tinygo.  the extensions to make we use are incompatible with traditional make.
* gnu make is what comes with linux and macos.  These programs actually look for `GNUmakefile` first. then `makefile` and `Makefile` last. i believe situations like this is the reason for this order.
* when using FreeBSD or other OSes with traditional make, we initially see a ton of errors and it takes a while to discern if the issue is GNU extensions or some other problem.  when renamed to `GNUmakefile`, traditional make will simplly complain that there's no `Makefile`.  The filename will signal that we need `gmake`.  OSes with gnu make by default will not be affected.
* Putting incompatible extensions in a Makefile introduces unnecessary incompatibly; making this simple fix can alleviate problems without needing to document special cases.
* there is a precedent for doing the right thing in nomad and other hashicorp projects: https://github.com/hashicorp/nomad/blob/main/GNUmakefile
* when the manual suggests naming the files `Makefile`, that is almost certainly in the context of a GNU system; i think they've been kind of dogmatic about that but that doesn't help portability to systems with non-gnu userlands.
* i think it is the right-thing-to-do(tm)  if we use GNU extensions.
* finally, this is a persistent quality-of-life issue that non-GNU people have to deal with that can be easily fixed.  we often lose this battle because people are enamored by the `Makefile` name.  by merging this, tinygo can stand as another example of favoring portability and correctness.

that's all.

dgryski:
Ok, those seems reasonable.

dkegel-fastly:
How about a toplevel forwarding Makefile so people don't have to invoke gmake directly?

ayang64:
> How about a toplevel forwarding Makefile so people don't have to invoke gmake directly?

That's interesting!  How would that work?

Just to be clear:  FreeBSD users have to invoke `gmake` right now and will have to after renaming.  The purpose of renaming the files is to make it clear that is necessary from the start.

I don't mind having to call `gmake` when I see a `GNUmakefile`.  The experience I had with tinygo is that I initially typed `make ...` and got a screen full of errors because there was only a `Makefile`

Anyway -- I'm interested in understanding how a forwarding `Makefile` would work.

aykevl:
This change seems fine by me. I'll have to get used to opening GNUmakefile instead of Makefile but I can live with that :)
As our Makefile is full of GNUisms it seems sensible to make that more explicit.

An alternative would be to change the Makefile to be FreeBSD compatible, but I'm not sure whether that's worth the trouble? And without good CI testing, I'm sure it'll quickly break as we are mainly testing using GNU make.

I'm curious however what the goal is? Do you plan to include TinyGo in FreeBSD? (I know very little about FreeBSD).
Right now, TinyGo doesn't support FreeBSD (or any BSD) as a build target, but I guess people that want to use TinyGo on FreeBSD are mainly interested in the wasm and microcontroller build targets?


> The experience I had with tinygo is that I initially typed `make ...` and got a screen full of errors because there was only a `Makefile`

Thought: would it be possible to check the make variant and exit early somehow with an error message? (I guess the answer is no, as `make` seems to read the whole file before executing any commands inside - it's not a shell script).

> How about a toplevel forwarding Makefile so people don't have to invoke gmake directly?

I suspect this is only going to be more confusing. But that's up to the people who actually use FreeBSD if they prefer it that way.

ayang64:
there is currently a freebsd port but it seems out of date.  I got the most
recent version to build with a few modifications. i don't think it would
take much to allow it to compile out of the box.

you may not explicitly support freebsd but you could do a few things to
allow it to work without modification.

at the end my goal is to be able to use it on my hardware.

aykevl:
> you may not explicitly support freebsd but you could do a few things to allow it to work without modification.

Right, yeah I'd be happy with any small fixes. I just don't think I can promise full FreeBSD support considering the (small) number of people using it.



> at the end my goal is to be able to use it on my hardware.

What do you mean? Produce FreeBSD binaries, or just run the compiler (for all the other non-FreeBSD build targets)?

ayang64:
> What do you mean? Produce FreeBSD binaries, or just run the compiler (for all the other non-FreeBSD build targets)?

run the binary to build non-FreeBSD targets.  i was at gophercon experimenting with a pybadge and i want to continue working with it.

aykevl:
I don't see any objections, so merging. If anybody objects to the name change of the Makefiles, we can of course revert the second commit.



> run the binary to build non-FreeBSD targets. i was at gophercon experimenting with a pybadge and i want to continue working with it.

Ah yes, cool :+1: :robot:

PR #3939, opened 2023-10-03
ci: use matrix instead of duplicating the Linux cross-compiling job

aykevl:
This reduces a lot of copy-pasted code (that was in fact different in a small way!)


aykevl:
Looks like this is working well.

deadprogram:
Thank you for the simplification @aykevl now merging.

PR #3940, opened 2023-10-04
tools: process SVD files in parallel

aykevl:
This speeds up the generation of SVD files, for example as part of a package build (where parallelism may not be desired, or tools should auto-detect the number of threads to use).

I have used the same pattern as in gen-device-avr, which was already parallelized.


aykevl:
Hmm, this doesn't seem to work so well. I suspect there are multiple input SVD files mapping to a single output Go file, hence the concurrency errors.

deadprogram:
Closing since out of date. Thanks!

PR #3948, opened 2023-10-13
Add a Nix flake file

aykevl:
This is a combination of two commits.

The first one refactors the clang include path. I hope I don't break any systems with it, CI should give at least an idea of how well it works.

The second makes some minor modifications, and also adds a flake.nix file. Here is the copy-pasted explanation:

---

This adds a flake.nix file that makes it possible to quickly create a development environment.

You can download Nix here, for use on your Linux or macOS system: https://nixos.org/download.html

After you have installed Nix, you can enter the development environment as follows:

    nix develop

This drops you into a bash shell, where you can install TinyGo simply using the following command:

    go install

That's all! Assuming you've set up your $PATH correctly, you can now use the tinygo command as usual:

    tinygo version

You can also do many other things from this environment. Building and flashing should work as you're used to: it's not a VM or container so there are no access restrictions.


deadprogram:

aykevl:
Oops I found a few missing bits, I'll be adding those in a new PR.

PR #3949, opened 2023-10-14
nix: improve docs and add support for wasi-libc

aykevl:
I forgot a few things in the flake file, but now everything should be included.


deadprogram:
Thanks @aykevl merging.

PR #3950, opened 2023-10-14
refactor: rm io/ioutil funcs

ginglis13:
io/ioutil has been deprecated since Go 1.16
https://pkg.go.dev/io/ioutil


deadprogram (approved):
Thank you for the improvement @ginglis13

PR #3952, opened 2023-10-15
Add Nix support on macOS

aykevl:
Improve support for NixOS on Darwin.
~WIP: it isn't fully supported yet (for example, `make wasi-libc` doesn't work correctly yet).~


aykevl:
Tested on MacOS and Linux, this works.

I have tested some boards (microbit-v2, arduino) and they work. With bossa installed in the environment, the arduino-nano33 also works.

deadprogram:
Thanks for adding this @aykevl now merging!

PR #3954, opened 2023-10-15
machine/spi: use interface to ensure uniformity for all machine implementations

deadprogram:
This PR does the same for `machine/spi` as #3406 did for `i2c`, which is to use an interface check to ensure uniformity for all machine implementations.


sago35 (approved):
LGTM
I think it is very good because the interface will be the same.

deadprogram:
Thanks for the review @sago35 now merging.

PR #3958, opened 2023-10-16
Binaryen116

flavio:
This PR is based on https://github.com/tinygo-org/tinygo/pull/3882 which seems to be no longer active.
I've run into issue https://github.com/tinygo-org/tinygo/issues/3881 too, this is why I would like to see this fix merged.

I've rebased the original PR against latest `dev` branch, and applied the requests from [this comment](https://github.com/tinygo-org/tinygo/pull/3882#issuecomment-1714392816).

Moreover, I've updated to latest release of binaryen v116, instead of v114 of the original PR

I've searched for other places inside of the GitHub actions where the binaryen code is checkout out, but this seems to be specific to Linux. On Mac/Windows the binaryen binaries are installed via package managers (brew and scoop).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ commands:
     steps:
       - run:
           name: "Pull submodules"
-          command: git submodule update --init
+          command: git submodule update --init --recursive
   llvm-source-linux:
     steps:
       - restore_cache:
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -356,13 +356,13 @@ jobs:
         uses: actions/cache@v3
         id: cache-binaryen
         with:
-          key: binaryen-linux-${{ matrix.goarch }}-v1
+          key: binaryen-linux-${{ matrix.goarch }}-v3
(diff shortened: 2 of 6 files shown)


flavio:
@deadprogram this is the new PR, which updates to binaryen 116

deadprogram:
Thank you for the update @flavio I am going to squash/merge now.

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ commands:
     steps:
       - run:
           name: "Pull submodules"
-          command: git submodule update --init
+          command: git submodule update --init --recursive
```
aykevl, line 8: Just wondering, is this `--recursive` a new requirement? (Because it also pulls another submodule that we don't actually need).

flavio: Without that, the code would not compile because of some missing dependency

aykevl: I think I found it. The extra dependency will be disabled here: https://github.com/tinygo-org/tinygo/pull/4154

PR #3964, opened 2023-10-19
Makefile: improve NodeJS version check

aykevl:
If node wasn't installed, the version was obtained anyway leading to a message like this:

    /bin/sh: node: command not found

But we only need to check the NodeJS version in specific cases.

This commit makes sure the version is only read when actually needed, thereby removing the spurious "command not found" error message.

Additionally, this removes one instance of `which`, which is not standard. Apparently the POSIX standard way to check for a command is the `command -v`, which is also built into shells (unlike `which`).

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -268,15 +268,20 @@ lib/wasi-libc/sysroot/lib/wasm32-wasi/libc.a:
 	cd lib/wasi-libc && $(MAKE) -j4 EXTRA_CFLAGS="-O2 -g -DNDEBUG -mnontrapping-fptoint -msign-ext" MALLOC_IMPL=none CC="$(CLANG)" AR=$(LLVM_AR) NM=$(LLVM_NM)
 
 # Check for Node.js used during WASM tests.
-NODEJS_VERSION := $(word 1,$(subst ., ,$(shell node -v | cut -c 2-)))
 MIN_NODEJS_VERSION=16
 
 .PHONY: check-nodejs-version
 check-nodejs-version:
-ifeq (, $(shell which node))
-	@echo "Install NodeJS version 16+ to run tests."; exit 1;
-endif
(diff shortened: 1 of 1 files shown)


deadprogram:
@aykevl please see build errors.

deadprogram:
@aykevl any update here?

ysoldak:
Rebased on top of dev to see if it works better now. Annoyed by this message too.

GNUmakefile
```diff
@@ -268,15 +268,20 @@ lib/wasi-libc/sysroot/lib/wasm32-wasi/libc.a:
...
-NODEJS_VERSION := $(word 1,$(subst ., ,$(shell node -v | cut -c 2-)))
 MIN_NODEJS_VERSION=16
 
 .PHONY: check-nodejs-version
 check-nodejs-version:
-ifeq (, $(shell which node))
-	@echo "Install NodeJS version 16+ to run tests."; exit 1;
-endif
-	@if [ $(NODEJS_VERSION) -lt $(MIN_NODEJS_VERSION) ]; then echo "Install NodeJS version 16+ to run tests."; exit 1; fi
+	@if ! command -v node &> /dev/null; then \
```
ysoldak, line 275:
```suggestion
	@if ! command -v node ; then \
```

ysoldak:
Added a suggestion that _may_ make it work on Ubuntu.
The root cause is Ubuntu uses Dash for `/bin/sh`, see https://wiki.ubuntu.com/DashAsBinSh

Alternative solution may be set this on top of the makefile:
```
SHELL = /bin/bash
```

deadprogram:
Any action on this PR? :smile_cat:

ysoldak:
I guess a micro but still architectural decision to be made here: try fix the failing command or enforce shell with SHELL variable.

@aykevl

deadprogram:
This was replaced by #4829 so now closing. Thanks everyone!

PR #3967, opened 2023-10-20
Allow larger systems to have a larger max stack alloc

### LINKED ISSUE #3331: Be smarter of stack<->heap allocations for "larger" systems.
Since https://github.com/tinygo-org/tinygo/pull/2628 , all allocations larger than 256 bytes are moved to the heap.  This might not make sense on systems where the default stack size is larger.

This would also allow a larger work queue size for the garbage collector's mark phase which will reduce the number of full heap rescans on overflow.

dgryski:
Fixes #3331

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -197,6 +197,7 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 		Scheduler:          config.Scheduler(),
 		AutomaticStackSize: config.AutomaticStackSize(),
 		DefaultStackSize:   config.StackSize(),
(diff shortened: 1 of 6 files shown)


dgryski:
I had a thought previously we needed to inject this value into the runtime somehow, but I can't find that code now so maybe I was wrong?  Anyways, marking as draft for now until we're certain this patch is actually all we need.

deadprogram:
Based on the previous feedback and description this appears ready to go to me.

@aykevl unless you have some other comments I am going to merge today.

deadprogram:
Now merging, thanks for working on this @dgryski

aykevl (commented):
A bit late but LGTM

PR #3968, opened 2023-10-20
nrf,sam,rp2040: add machine.DeviceID function

kenbell:
This seems like a common feature and is super useful in network/lora/mesh scenarios to separate devices without flash config.

I've implemented on 3 chipsets nrf,sam,rp2040 so I'm confident we can have a common API.

### DIFF
--- a/src/machine/machine_atsam.go
+++ b/src/machine/machine_atsam.go
@@ -1,0 +1,31 @@
+//go:build sam
+
+package machine
+
+import (
+	"runtime/volatile"
+	"unsafe"
+)
+
+var deviceID [16]byte
+
+// DeviceID returns an identifier that is unique within
+// a particular chipset.
+//
+// The identity is one burnt into the MCU itself, or the
+// flash chip at time of manufacture.
+//
+// It's possible that two different vendors may allocate
+// the same DeviceID, so callers should take this into
+// account if needing to generate a globally unique id.
+//
+// The length of the hardware ID is vendor-specific, but
+// 8 bytes (64 bits) and 16 bytes (128 bits) are common.
+func DeviceID() []byte {
+	for i := 0; i < len(deviceID); i++ {
+		word := (*volatile.Register32)(unsafe.Pointer(deviceIDAddr[i/4])).Get()
+		deviceID[i] = byte(word >> ((i % 4) * 8))
+	}
+
+	return deviceID[:]
+}
--- a/src/machine/machine_rp2040_flash.go
+++ b/src/machine/machine_rp2040_flash.go
@@ -13,6 +13,32 @@ func EnterBootloader() {
 	enterBootloader()
 }
 
+// 13 = 1 + FLASH_RUID_DUMMY_BYTES(4) + FLASH_RUID_DATA_BYTES(8)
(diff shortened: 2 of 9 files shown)


src/machine/machine_atsam.go
```diff
@@ -0,0 +1,31 @@
...
+// flash chip at time of manufacture.
+//
+// It's possible that two different vendors may allocate
+// the same HardwareID, so callers should take this into
+// account if needing to generate a globally unique id.
+//
+// The length of the hardware ID is vendor-specific, but
+// 8 bytes (64 bits) and 16 bytes (128 bits) are common.
+func HardwareID() []byte {
+	id := make([]byte, 16)
```
deadprogram: Perhaps avoid the heap allocation here?

```
var hardwareID [16]byte

func HardwareID() []byte {
    ...
    return hardwareID[:]
}
```

kenbell: thanks for reviewing.  yeah, i like that change - will update the PR

src/machine/machine_rp2040_flash.go
```diff
@@ -13,6 +13,33 @@ func EnterBootloader() {
...
+//
+// It's possible that two different vendors may allocate
+// the same HardwareID, so callers should take this into
+// account if needing to generate a globally unique id.
+//
+// The length of the hardware ID is vendor-specific, but
+// 8 bytes (64 bits) is common.
+func HardwareID() []byte {
+	// 13 = 1 + FLASH_RUID_DUMMY_BYTES(4) + FLASH_RUID_DATA_BYTES(8)
+	txBuf := make([]byte, 13)
```
deadprogram: Should also do something about these heap allocations.

src/machine/machine_rp2040_rom.go
```diff
@@ -132,13 +159,62 @@ void ram_func flash_erase_blocks(uint32_t offset, size_t count)
...
+
+	flash_init_boot2_copyout();
+
+	__compiler_memory_barrier();
+
+	flash_connect_internal_func();
+	flash_exit_xip_func();
+
+	flash_cs_force(0);
+    size_t tx_remaining = count;
```
deadprogram: Formatting?

kenbell: ack

src/machine/machine_rp2040_flash.go
kenbell: i was less sure about these - i'll think about it.  It's a bit messier given the flash 'protocol stuff' and returning an unaligned slice into it.  I'm wondering if escape analysis would take care of allocating rxBuf on the stack?

kenbell: Looks like we can use just one buffer for tx and rx and return a slice of that, so I've done that.

deadprogram:
Hopefully last feedback: should we perhaps call this something less generic than `HardwareID`? Possible options that come to mind:

- `SerialNumber`
- `UniqueID`
- `DeviceID`

kenbell:
I like DeviceID - so I’ve gone with that

deadprogram:
Thanks very much for this addition @kenbell now merging.

PR #3969, opened 2023-10-20
Fix macos CI builds - github runner python issues

kenbell:
See: https://github.com/actions/runner-images/issues/11545

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -122,6 +122,12 @@ jobs:
     steps:
       - name: Update Homebrew
         run: brew update
+      - name: Fix Python symlinks
+        run: |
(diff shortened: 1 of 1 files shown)


aykevl:
I wonder whether it makes sense to wait a bit, perhaps the CI image will soon be fixed? I can't imagine we're the only ones hitting this.

kenbell:
It's been on-going since December 2022, so I'm not so sure it will be fixed soon.

kenbell:
I suspect this will be one of those things that recurs when there's a new python release.  I think the 'fix' I have is (a) working and (b) fairly innocuous in terms of future maintenance.

deadprogram:
@kenbell thanks for the fix. It is rather annoying to be sure, but seems like upstream is stalled.

Issue #3974, opened 2023-10-23
Xtensa: unable to compile code that needs clang headers


deadprogram:
This commit introduced a problem that only seems to appear when compiling for Xtensa:

https://github.com/tinygo-org/tinygo/commit/d801d0cd539447322228e9f1daa860774584b4d3

The TinyGo drivers has a test case that can reproduce it:

```
tinygo build -size short -o ./build/test.elf -target=m5stack-core2 ./examples/ft6336/basic/
# tinygo.org/x/drivers/delay
delay/sleep.go:10:10: fatal: 'stdbool.h' file not found
```

I have confirmed that this commit is the one that introduced the error.

The only reason this was not discovered sooner, is because of an error in the CI discovered by @trichner as part of https://github.com/tinygo-org/drivers/pull/614

deadprogram:
cc @aykevl

deadprogram:
Corrected by #3975

deadprogram:
Closing since the issue was introduced in the current dev cycle. Thanks @aykevl for the help!

PR #3975, opened 2023-10-23
compileopts: add back cflag '-isystem' for resource directory search

deadprogram:
This PR modifies compileopts to add back the cflag '-isystem' for resource directory search since it is needed for Xtensa.

Fixes https://github.com/tinygo-org/tinygo/issues/3974

### DIFF
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -265,11 +265,17 @@ func (c *Config) CFlags(libclang bool) []string {
 	}
 	resourceDir := goenv.ClangResourceDir(libclang)
 	if resourceDir != "" {
-		// The resoure directory contains the built-in clang headers like
+		// The resource directory contains the built-in clang headers like
 		// stdbool.h, stdint.h, float.h, etc.
 		// It is left empty if we're using an external compiler (that already
 		// knows these headers).
-		cflags = append(cflags, "-resource-dir="+resourceDir)
+		cflags = append(cflags,
+			"-resource-dir="+resourceDir,
+		)
+		if strings.HasPrefix(c.Triple(), "xtensa") {
+			// workaround needed in LLVM 16, see: https://github.com/espressif/llvm-project/issues/83
+			cflags = append(cflags, "-isystem", filepath.Join(resourceDir, "include"))
+		}
 	}
 	switch c.Target.Libc {
 	case "darwin-libSystem":


aykevl:
This looks like a Clang bug, I'll investigate what's going on here.
(We probably still need a workaround, but we can report the issue to Espressif at the same time).

compileopts/config.go
```diff
@@ -265,11 +265,14 @@ func (c *Config) CFlags(libclang bool) []string {
 	}
 	resourceDir := goenv.ClangResourceDir(libclang)
 	if resourceDir != "" {
-		// The resoure directory contains the built-in clang headers like
+		// The resource directory contains the built-in clang headers like
```
aykevl, line 268: :sweat_smile:

aykevl:
Looks like there is indeed something weird in Clang, still digging...

aykevl:
See: https://github.com/espressif/llvm-project/issues/83

compileopts/config.go
```diff
@@ -265,11 +265,14 @@ func (c *Config) CFlags(libclang bool) []string {
...
 		// stdbool.h, stdint.h, float.h, etc.
 		// It is left empty if we're using an external compiler (that already
 		// knows these headers).
-		cflags = append(cflags, "-resource-dir="+resourceDir)
+		cflags = append(cflags,
+			"-resource-dir="+resourceDir,
+			"-isystem", filepath.Join(resourceDir, "include"), // needed by Xtensa
+		)
```
aykevl: Can you do this in a way that only applies to Xtensa?

Something like:

```go
if strings.HasPrefix(c.Triple(), "xtensa") {
    // workaround needed in LLVM 16, see: https://github.com/espressif/llvm-project/issues/83
    cflags = append(cflags, "-isystem", filepath.Join(resourceDir, "include"))
}
```

This makes it clear that it is a workaround that can hopefully be removed in the future. And makes sure it only applies to Xtensa, to avoid affecting other platforms.

Also, I suspect a similar change will be needed in builder/library.go (test after `tinygo clean` to be sure).

aykevl:
I've also made an upstream patch: https://github.com/espressif/llvm-project/pull/84

compileopts/config.go
deadprogram: Updated `compileopts/config.go` based on your suggestion.

I did not need any change to `builder/library.go` even after a `tinygo clean`

deadprogram:
@aykevl anything else needed here before merge?

aykevl (approved):

PR #3981, opened 2023-10-29
doc: fix a typo in the rtcinterrupt example

graugans:
I came across this while glancing through the examples.

Maybe it also makes sense to link to the docs like this: https://tinygo.org/docs/concepts/compiler-internals/heap-allocation/ 
Is it worth adding the hint about blocking and memory to the interrupt documentation page https://tinygo.org/docs/concepts/compiler-internals/interrupts/ as well? This came to my mind when I was thinking if `fmt.Println()` is blocking or not I am pretty sure it does :-)

### DIFF
--- a/src/examples/rtcinterrupt/rtcinterrupt.go
+++ b/src/examples/rtcinterrupt/rtcinterrupt.go
@@ -25,6 +25,8 @@ func main() {
 	// Schedule and enable recurring interrupt.
 	// The callback function is executed in the context of an interrupt handler,
 	// so regular restrictions for this sort of code apply: no blocking, no memory allocation, etc.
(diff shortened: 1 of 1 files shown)


deadprogram:
@graugans those seem like very good suggestions. Can you please add that?

graugans:
> @graugans those seem like very good suggestions. Can you please add that?

I'll do my best.

graugans:
I have added a link to the online documentation. ~~I will create a Pull-Request for the documentation project to provide more hints about interrupts.~~ See https://github.com/tinygo-org/tinygo-site/pull/378

deadprogram:
Thanks @graugans for the doc improvement. Now squash/merging.

PR #3982, opened 2023-10-30
Add -info option to tinygo monitor

sago35:
The output of running `tinygo monitor --info` is as follows.
It displays information about the targets managed by tinygo, making it easier to understand what is connected where.

Even if it is not defined in tinygo's `targets/*.json`, some information is displayed.
ex) COM101

```
$ tinygo monitor --info
COM1
COM155 2886 8045 xiao-ble
COM32
COM33
COM76 239A 8022 feather-m4
COM101 2E8A F00A
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -1436,6 +1436,7 @@ func main() {
 	llvmFeatures := flag.String("llvm-features", "", "comma separated LLVM features to enable")
 	cpuprofile := flag.String("cpuprofile", "", "cpuprofile output")
 	monitor := flag.Bool("monitor", false, "enable serial monitor")
+	info := flag.Bool("info", false, "print information")
 	baudrate := flag.Int("baudrate", 115200, "baudrate of serial monitor")
 
 	// Internal flags, that are only intended for TinyGo development.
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -1731,41 +1732,29 @@ func main() {
...
+				fmt.Printf("%s %4s %4s %s\n", s.Name, s.VID, s.PID, s.Target)
+			}
+		} else {
+			err := Monitor("", *port, options)
+			handleCompilerError(err)
+		}
 	case "targets":
-		dir := filepath.Join(goenv.Get("TINYGOROOT"), "targets")
-		entries, err := os.ReadDir(dir)
+		specs, err := getTargetSpecs()
```
deadprogram: Why export `ListSerialPorts()` but not export `getTargetSpecs()`? They should both have the same level of external visibility, whatever that is, don't you think @sago35 ?

sago35: Yes, I believe your point is valid.
I've made the corrections.
I also wanted to adjust the commit message, so I did a rebase and force-pushed.

monitor.go
```diff
@@ -128,6 +133,94 @@ func Monitor(executable, port string, options *compileopts.Options) error {
...
+				}
+			}
+		}
+		serialPortInfo = append(serialPortInfo, info)
+	}
+
+	return serialPortInfo, nil
+}
+
+func GetTargetSpecs() (map[string]*compileopts.TargetSpec, error) {
```
deadprogram: This file `monitor.go` seems like an odd place to move this function.

monitor.go
```diff
@@ -128,6 +133,94 @@ func Monitor(executable, port string, options *compileopts.Options) error {
...
+			return nil, fmt.Errorf("cnuld not list target: %w", err)
+		}
+		if spec.FlashMethod == "" && spec.FlashCommand == "" && spec.Emulator == "" {
+			// This doesn't look like a regular target file, but rather like
+			// a parent target (such as targets/cortex-m.json).
+			continue
+		}
+		name := entry.Name()
+		name = name[:len(name)-5]
+		//fmt.Println(name)
```
deadprogram: Remove commented line, please.

sago35: done.

monitor.go
sago35: `GetTargetSpecs() has been moved to the compileopts package.

deadprogram:
Thanks for working on this and making the requested changes @sago35 now merging.

PR #3983, opened 2023-11-02
Implement Trylock

soypat:
~I also fixed something which _seems_ like a bug... The unlock method was not resetting the value of the `locked` field. I suspect there may be more than meets the eye and that this is not a bug...~

I now understand what the code is doing, removed the bug I introduced via force push :facepalm:

### DIFF
--- a/src/sync/mutex.go
+++ b/src/sync/mutex.go
@@ -3,10 +3,12 @@ package sync
 import (
 	"internal/task"
 	_ "unsafe"
+
+	"runtime/volatile"
 )
 
 type Mutex struct {
-	locked  bool
+	state   uint8 // Set to non-zero if locked.
 	blocked task.Stack
 }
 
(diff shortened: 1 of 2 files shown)


src/sync/mutex.go
```diff
@@ -37,6 +37,19 @@ func (m *Mutex) Unlock() {
 	}
 }
 
+// TryLock tries to lock m and reports whether it succeeded.
+//
+// Note that while correct uses of TryLock do exist, they are rare,
+// and use of TryLock is often a sign of a deeper problem
+// in a particular use of mutexes.
+func (m *Mutex) TryLock() bool {
+	if m.locked {
```
dgryski: It wouldn't surprise me if we needed calls to volatile loads here: https://github.com/tinygo-org/tinygo/blob/release/src/runtime/volatile/volatile.go

soypat: :thinking: Wouldn't a memory barrier be better suited for this use case? Even then, tinygo is single threaded so we should have consistent ordering?

dgryski: I'm more concerned about LLVM removing loads/unloads it doesn't think are important.  Yes, TInyGo is single threaded so memory barriers are less interesting.

soypat: I've added volatile loads/stores as requested!

deadprogram:
Looking at this code, seems like it should be OK to squash/merge.

@soypat do you want to squash the commits yourself?
@aykevl or anyone else any other feedback before merge?

aykevl (approved):
Looks good to me, though if we're going to care about atomicity we should really use the sync/atomic package instead. Volatile has slightly different semantics (though for most microcontrollers it doesn't really matter).

Rebased on the dev branch, ready to merge once CI passes.

Issue #2597, opened 2022-01-30
fmt.Scanf()  does not work.


bmentink:
I get a  runtime "panic: (319:0x20001970)" whenever I try to use Scanf in the form:

fmt.Scanf("%s", &command)

Is it implemented yet?

deadprogram:
What target are you compiling for @bmentink and can you please show a bit more of your code?

bmentink:
@deadprogram  Hi Ron,

I am compiling for the RaspberryPi pico target. Code is part of a much larger project(sensorless BLDC motor drive), the relevant part is:

import (
 "fmt"
)

var command, data string

println("Enter command:")
fmt.Scanf("%s %s", &command, &data)		
fmt.Printf("Command = %s, Data = %s\r\n", command, data)

EDIT: A single  fmt.Scanf("%s", &command)  .. does the same panic ...
Also, the form fmt.Scan(&command) causes the same panic.

Cheers,

dgryski:
Does the pico have an `os.Stdin`?  `fmt.Scanf` works for me on native but ends up panicing on wasi due to EOF.  I wonder if its a similar issue.

bmentink:
@dgryski  Yes, there is an `os.Stdin`

bmentink:
@deadprogram 
Do you need any more info to progress this?

deadprogram:
> Yes, there is an os.Stdin

On the rp2040? Hmm, I am not too sure that is true. Have to look into that.

bmentink:
@deadprogram 
You might be right. I just assumed there was because the TinyGo plugin/Pico target (VS Code ) gave me the popup prompt for os.Stdin when I started typing it ....

bmentink:
I am implementing a brushless BLDC motor drive as an excuse to test how feasible TinyGo is with coding for the RPi Pico. It seems I have run into the 1st road block, by not having a serial in. Serial  is essential for me, as I am controlling this drive over a Bluetooth->Serial connection. (Also PIO support would have been nice for this project as well, but have got around that with PWM drivers)

@deadprogram  .. any luck with looking into this ?

deadprogram:
You do have serial in/out without any problem for example using the UART. What you do not have is any operating system to provide stdin. Hence why you cannot just use `fmt.Scanf()`.

See https://github.com/tinygo-org/tinygo/blob/release/src/examples/echo/echo.go for some ways to read/write to serial.

bmentink:
[...]

PR #3984, opened 2023-11-04
machine: modify UART and USB CDC ACM Read() to be blocking 

xudongzheng:
Using a callback for receiving serial data eliminates the need for polling.

So far I've only modified the files for RP2040 as that's what I'm testing on. If this PR is in the right direction, I will modify `type UART struct` for the other boards as well.

### DIFF
--- a/src/machine/uart.go
+++ b/src/machine/uart.go
@@ -9,6 +9,11 @@ var errUARTBufferEmpty = errors.New("UART buffer empty")
 // UARTParity is the parity setting to be used for UART communication.
 type UARTParity uint8
 
+type UARTCommon struct {
+	Buffer *RingBuffer
+	RXChan chan bool
+}
+
 const (
 	// ParityNone means to not use any parity checking. This is
 	// the most common setting.
@@ -26,23 +31,24 @@ const (
 // To implement the UART interface for a board, you must declare a concrete type as follows:
 //
 // 		type UART struct {
-// 			Buffer *RingBuffer
+// 			UARTCommon
 // 		}
 //
-// You can also add additional members to this struct depending on your implementation,
-// but the *RingBuffer is required.
-// When you are declaring your UARTs for your board, make sure that you also declare the
-// RingBuffer using the NewRingBuffer() function when you declare your UART:
+// You can also add additional members to this struct depending on your
+// implementation. UARTCommon is required. When you are declaring your UARTs for
+// your board, make sure that you also initialize UARTCommon using the
+// NewUARTCommon() function:
 //
-//		UART{Buffer: NewRingBuffer()}
+//		UART{UARTCommon: NewUARTCommon()}
 //
 
 // Read from the RX buffer.
 func (uart *UART) Read(data []byte) (n int, err error) {
-	// check if RX buffer is empty
+	// If RX buffer is empty, block until data is available.
 	size := uart.Buffered()
 	if size == 0 {
-		return 0, nil
+		<-uart.RXChan
+		size = uart.Buffered()
 	}
 
 	// Make sure we do not read more from buffer than the data slice can hold.
@@ -103,4 +109,16 @@ func (uart *UART) Buffered() int {
 // Usually called by the IRQ handler for a machine.
 func (uart *UART) Receive(data byte) {
 	uart.Buffer.Put(data)
+
+	select {
+	case uart.RXChan <- true:
+	default:
+	}
+}
+
+func NewUARTCommon() UARTCommon {
+	return UARTCommon{
+		Buffer: NewRingBuffer(),
+		RXChan: make(chan bool),
+	}
 }
--- a/src/machine/usb/cdc/usbcdc.go
+++ b/src/machine/usb/cdc/usbcdc.go
@@ -23,10 +23,11 @@ type cdcLineInfo struct {
 
 // Read from the RX buffer.
 func (usbcdc *USBCDC) Read(data []byte) (n int, err error) {
-	// check if RX buffer is empty
+	// If RX buffer is empty, block until data is available.
 	size := usbcdc.Buffered()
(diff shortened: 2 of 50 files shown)


src/machine/usb/cdc/usbcdc.go
```diff
@@ -59,17 +59,27 @@ func (usbcdc *USBCDC) Buffered() int {
...
+
 // Receive handles adding data to the UART's data buffer.
 // Usually called by the IRQ handler for a machine.
 func (usbcdc *USBCDC) Receive(data byte) {
+	if usbcdc.rxCallback != nil {
+		if usbcdc.rxCallback(data) {
+			return
+		}
+	}
 	usbcdc.rxBuffer.Put(data)
```
sago35: The callback receives data as an argument, and data is also added to the ringbuffer. This could potentially lead to some confusion in the handling of data. Do you have any thoughts on this?

xudongzheng: The idea is that the callback can return true if it successfully processed the byte and it doesn't need to be buffered. If it returns false, the byte would be buffered as if the callback did not exist.

sago35: Understood.

sago35 (approved):
LGTM

deadprogram:
The only change I would like to propose is to name this `rxHandler` instead. Great work @xudongzheng

src/machine/uart.go
```diff
@@ -99,8 +100,17 @@ func (uart *UART) Buffered() int {
 	return int(uart.Buffer.Used())
 }
 
+func (uart *UART) SetRXCallback(cb func(byte) bool) {
+	uart.rxCallback = cb
+}
```
aykevl: This needs documentation that includes the following:
- it is an interrupt (not just a regular callback)
- it should describe what the return value is

See for example:

https://github.com/tinygo-org/tinygo/blob/731532cd2b6353b60b443343b51296ec0fafae09/src/machine/machine_nrf.go#L73-L80

aykevl (commented):
> Using a callback for receiving serial data eliminates the need for polling.

Is that the only reason? Because handling things inside an interrupt is a pretty big hammer. If this is the only reason, a much better approach would be to make methods like `ReadByte` blocking (to avoid polling).

Also, this is a new API so should be added here: https://tinygo.org/docs/reference/machine/#uart

xudongzheng:
> Is that the only reason? Because handling things inside an interrupt is a pretty big hammer. If this is the only reason, a much better approach would be to make methods like `ReadByte` blocking (to avoid polling).

Thanks for the suggestion. Please see the latest commit fe186e971e58101b92b003de0c4e64d24e4bc787, which implements blocking `Read()` on both RP2040 UART and USB CDC ACM. If this implementation is preferred, I will go ahead and make the change for other boards as well, and also get rid of the RX callback code.

I opted to change the `Read()` behavior without changing the `ReadByte()` behavior since the error return value becomes meaningless if `ReadByte()` becomes blocking. However I can change `ReadByte()` along with the function return value if that's preferred.

sago35:
@xudongzheng 
I believe the related PR and Issue are as follows.

#2739 
#2597
#2625
#2625

The essence of this PR is to make the `Read()` functions for UART and USBCDC more aligned with the expected Go `io.Reader` interface. In other words, we want to ensure that it does not return a value where `n == 0` and `err == nil`. This will enable standard Go functions like `fmt.Scanf` to work smoothly.

In that case, I believe it should be implemented without using a channel. Additionally, this change will break compatibility with TinyGo. Documentation is also necessary for this, and it needs to be reviewed carefully.

Of course, it might also be an option to define a struct like `UARTxxxx` instead of `UART` to maintain compatibility.

> https://pkg.go.dev/io#Reader
> Implementations of Read are discouraged from returning a zero byte count with a nil error, except when len(p) == 0.
> Callers should treat a return of 0 and nil as indicating that nothing happened; in particular it does not indicate EOF.

xudongzheng:
> In that case, I believe it should be implemented without using a channel.

I don't quite see how this would be implemented without channels. For that reason, I initially opted for adding a callback rather than changing the `Read()` behavior - anyone who wants blocking reads can easily add a channel and write a wrapper to make reads block.

I think the overhead of channel should be pretty low, and it should be one or two at most since there won't be that many serial ports. Performance-wise, it works well enough as I was able to receive about 400KB/s on RP2040 over USB CDC ACM.

> Additionally, this change will break compatibility with TinyGo. Documentation is also necessary for this, and it needs to be reviewed carefully.

I agree, though I would like a bit more clarity on the implementation direction before I start working on the documentation.

lePereT:
Hi! Is there any progress on merging this PR?

aykevl:
I would much prefer an implementation that makes `Read` and `ReadByte` blocking than one that requires a totally new way of interacting with the hardware (via interrupts or channels). See https://github.com/tinygo-org/tinygo/pull/2625 for how this was done for `os.Stdin`. For example by moving the blocking behavior out of `runtime.getchar` and into `ReadByte` itself.

...or is that what this PR already does?

xudongzheng:
> For example by moving the blocking behavior out of `runtime.getchar` and into `ReadByte` itself.

I also think it makes sense to refactor `runtime.getchar` so it's just:

```
func getchar() byte {
	v, _ := machine.Serial.ReadByte()
	return v
}
```

However, are you suggesting I don't use something like channels? The alternative is to have `machine.UART.Read()` do:

```
func (uart *UART) Read(data []byte) (n int, err error) {
	for uart.Buffered() == 0 {
		Gosched()
	}
	...
}
```

My concern is that this wouldn't allow the microcontroller to sleep while waiting for additional data, which is useful for battery-powered boards like nRF. Communicating with a channel should allow sleep in theory.

PR #3985, opened 2023-11-04
machine: add SPI DMA support for the rp2040 and samd51

aykevl:
This adds SPI DMA support for the rp2040, and a fallback for all the other devices.

For more details, see: https://github.com/tinygo-org/tinygo-site/pull/342

Draft, until:
 - [x] I've added samd51 support
 - [ ] https://github.com/tinygo-org/tinygo-site/pull/342 has been resolved

### DIFF
--- a/src/machine/machine_rp2040_spi.go
+++ b/src/machine/machine_rp2040_spi.go
@@ -289,7 +289,26 @@ func (spi SPI) isBusy() bool {
 
 // tx writes buffer to SPI ignoring Rx.
 func (spi SPI) tx(tx []byte) error {
-	if len(tx) == 0 {
+	err := spi.StartTx(tx, nil)
+	if err != nil {
+		return err
+	}
+	return spi.Wait()
+}
+
+// IsAsync returns whether the SPI supports async operation (usually DMA).
+//
+// It returns true on the rp2040.
+func (spi SPI) IsAsync() bool {
+	return true
+}
+
+// Start a transfer in the background.
+//
+// After this, another StartTx() or Wait() must be called. The provided byte
+// slices (tx and rx) may only be accessed again after Wait() was called.
+func (spi SPI) StartTx(tx, rx []byte) error {
+	if len(tx) == 0 && len(rx) == 0 {
 		// We don't have to do anything.
 		// This avoids a panic in &tx[0] when len(tx) == 0.
 		return nil
@@ -306,6 +325,15 @@ func (spi SPI) tx(tx []byte) error {
 		dreq = 18 // DREQ_SPI1_TX
 	}
 
+	// Wait for the previous transmission to complete.
+	for ch.CTRL_TRIG.Get()&rp.DMA_CH0_CTRL_TRIG_BUSY != 0 {
+	}
+
+	if len(rx) != 0 {
+		// Fallback. We don't support receiving data using DMA yet.
+		return spi.Tx(tx, rx)
+	}
+
 	// Configure the DMA peripheral as follows:
 	//   - set read address, write address, and number of transfer units (bytes)
 	//   - increment read address (in memory), don't increment write address (SSPDR)
@@ -319,6 +347,19 @@ func (spi SPI) tx(tx []byte) error {
 		rp.DMA_CH0_CTRL_TRIG_DATA_SIZE_SIZE_BYTE<<rp.DMA_CH0_CTRL_TRIG_DATA_SIZE_Pos |
 		dreq<<rp.DMA_CH0_CTRL_TRIG_TREQ_SEL_Pos |
 		rp.DMA_CH0_CTRL_TRIG_EN)
+	return nil
+}
+
+// Wait until all active transactions (started by StartTx) have finished. The
+// buffers provided in StartTx will be available after this method returns.
+func (spi SPI) Wait() error {
+	// Pick the DMA channel reserved for this SPI peripheral.
+	var ch *dmaChannel
+	if spi.Bus == rp.SPI0 {
+		ch = &dmaChannels[spi0DMAChannel]
+	} else { // SPI1
+		ch = &dmaChannels[spi1DMAChannel]
+	}
 
 	// Wait until the transfer is complete.
 	// TODO: do this more efficiently:
--- a/src/machine/machine_atsamd51.go
+++ b/src/machine/machine_atsamd51.go
@@ -683,6 +683,56 @@ func (p Pin) getPinGrouping() (uint8, uint8) {
 	return group, pin_in_group
 }
 
+// Static DMA channel allocation.
+// If there are a lot of DMA using peripherals, we might need to switch to
+// dynamic allocation instead.
+const (
+	dmaChannelSERCOM0 = iota
+	dmaChannelSERCOM1
+	dmaChannelSERCOM2
+	dmaChannelSERCOM3
+	dmaChannelSERCOM4
+	dmaChannelSERCOM5
+	dmaChannelSERCOM6
+	dmaChannelSERCOM7
+	dmaNumChannels
+)
+
+// DMA descriptor structure. This structure is defined by the hardware, and is
+// described by "22.9 Register Summary - SRAM" in the datasheet.
+type dmaDescriptor struct {
+	btctrl   uint16
+	btcnt    uint16
+	srcaddr  unsafe.Pointer
+	dstaddr  unsafe.Pointer
+	descaddr unsafe.Pointer
+}
+
+//go:align 16
+var dmaDescriptorSection [dmaNumChannels]dmaDescriptor
+
+//go:align 16
+var dmaDescriptorWritebackSection [dmaNumChannels]dmaDescriptor
+
+// Enable and configure the DMAC peripheral if it hasn't been enabled already.
+func enableDMAC() {
+	if !sam.DMAC.CTRL.HasBits(sam.DMAC_CTRL_DMAENABLE) {
+		// Init DMAC.
+		// First configure the clocks, then configure the DMA descriptors. Those
+		// descriptors must live in SRAM and must be aligned on a 16-byte
+		// boundary.
+		// Some examples:
+		// http://www.lucadavidian.com/2018/03/08/wifi-controlled-neo-pixels-strips/
+		// https://svn.larosterna.com/oss/trunk/arduino/zerotimer/zerodma.cpp
+		sam.MCLK.AHBMASK.SetBits(sam.MCLK_AHBMASK_DMAC_)
+		sam.DMAC.BASEADDR.Set(uint32(uintptr(unsafe.Pointer(&dmaDescriptorSection))))
+		sam.DMAC.WRBADDR.Set(uint32(uintptr(unsafe.Pointer(&dmaDescriptorWritebackSection))))
+
+		// Enable peripheral with all priorities.
+		sam.DMAC.CTRL.SetBits(sam.DMAC_CTRL_DMAENABLE | sam.DMAC_CTRL_LVLEN0 | sam.DMAC_CTRL_LVLEN1 | sam.DMAC_CTRL_LVLEN2 | sam.DMAC_CTRL_LVLEN3)
+	}
+}
+
 // InitADC initializes the ADC.
 func InitADC() {
 	// ADC Bias Calibration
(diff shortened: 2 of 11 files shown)


src/machine/machine_rp2040_spi.go
```diff
@@ -306,6 +325,15 @@ func (spi SPI) tx(tx []byte) error {
 		dreq = 18 // DREQ_SPI1_TX
 	}
 
+	// Wait for the previous transmission to complete.
+	for ch.CTRL_TRIG.Get()&rp.DMA_CH0_CTRL_TRIG_BUSY != 0 {
```
kenbell, line 329: Since this is a busy loop, why not yield to the scheduler here - it's a busy loop either way, and yielding would allow stuff like timers to happen?

src/machine/machine_rp2040_spi.go
```diff
@@ -319,6 +347,19 @@ func (spi SPI) tx(tx []byte) error {
...
+// Wait until all active transactions (started by StartTx) have finished. The
+// buffers provided in StartTx will be available after this method returns.
+func (spi SPI) Wait() error {
+	// Pick the DMA channel reserved for this SPI peripheral.
+	var ch *dmaChannel
+	if spi.Bus == rp.SPI0 {
+		ch = &dmaChannels[spi0DMAChannel]
+	} else { // SPI1
+		ch = &dmaChannels[spi1DMAChannel]
+	}
```
kenbell, line 362: Same comment as above - this goes through to a busy loop.  I'm not sure how this enables async behavior if it's busy-loop.  How does:
```go
spi.StartTx()
spi.Wait()
```
 work better than doing:
 ```go
 spi.Tx()
 ```
 
It seems like it only helps in one very specific case where the code between StartTx and Wait can do something meaningful for an app - and I'm not sure how that works in a driver in tinygo drivers.  If a display-chip driver wants to use DMA how does the driver yield to the app logic, so the app can do some useful processing ready for the next frame?
 
I can see how this works if the app is itself calling `spi.StartTx` and `spi.Wait`, but not how it works when there's a driver.

kenbell (commented):
It would be helpful to add one (maybe two?) examples showing best practice on how to use StartTx / Wait in a driver (like tinygo drivers) and directly in an app (if that use-case is expected).

src/machine/machine_rp2040_spi.go
aykevl: In short, I plan on adding methods like `StartDrawBitmap` and `Wait` to the display drivers. I already have these changes locally, but would like to improve them a bit before submitting.
Basically, that way you can do display updates and calculating the next frame at the same time. Pseudocode:

```
for part of display:
    fill half of the buffer
    wait until previous buffer has been sent
    send buffer to screen
    continue with other half of the buffer
wait until previous buffer has been sent
```

I've managed to hit nearly the maximum possible FPS on a st7789 this way, because the app could fill the next buffer to be sent while the previous buffer was still transmitting - thereby doing the two in parallel.

If you'd like to see these changes first in the driver I understand.

aykevl:
Here is a driver that shows how this async API can be used: https://github.com/tinygo-org/drivers/pull/625
And here is how I plan to use this driver in a graphics library: https://github.com/aykevl/tinygl/compare/async-update

While the graphics library needs to be careful to support async operations, in most cases it's pretty simple to use (get a new buffer, write to it, send to the display) and in fact users of the graphics library don't need to know about it at all unless they implement a custom widget.

In practice, I've found that fast display updates (where rendering takes less time than a display update) are limited by the time it takes to update the display (plus some small extra overhead). This means that it's possible to do most display updates on the Gopher Badge in about 20ms (the hardware limit) and it will only take longer when rendering takes longer than 20ms.
So in short, it works and it speeds up display performance substantially - about as much as DMA theoretically could.

src/machine/machine_rp2040_spi.go
soypat: I'm still not convinced this could not be written as application code with the old blocking SPI abstraction, i.e:

```python
workerDisplayThread(workToDoChannel):
   for:
      ourWork <- workToDoChannel;
      if !(ourWork && displayOn):
         return
      fill our part of buffer given ourWork
      lock spi bus
      send buffer to screen while blocking
      unlock spi bus
```

Some advantages of the above abstraction:
- `machine.SPI` kept simple, which in my opinion is a huge gain. I really don't want to end up with something like `smbus` or the micropython `i2c` abstraction...
- Multiplexing on to more than just 2 threads is pretty straightforward in case the rendering is the bottleneck, given you have multicore processing.

aykevl: So if I understand you correctly, you propose doing the work in two separate goroutines that both have half the buffer?
One thing I ran into when I tried this is that this part:

> ```
> send buffer to screen while blocking
> ```

...might actually not happen in parallel. So what could happen is that the first goroutine renders a buffer, then locks the SPI bus (or does something else that can cause a context switch), and then the second goroutine starts rendering into its own buffer _before the first goroutine has started the SPI transfer_. Which means there is no speed benefit at all. This is especially likely to occur when the SPI code sprinkles some `runtime.Gosched()` calls throughout, almost guaranteeing that it will switch over to the second goroutine before starting the SPI transfer.

This is really hard to reason about and I don't see a good way around it. I've found the async transmit and wait calls to be much easier to reason about.

Perhaps you know of a way to avoid this trap, though?

soypat: > This is really hard to reason about and I don't see a good way around it. I've found the async transmit and wait calls to be much easier to reason about.

I agree, it is harder to reason about, but at least it can be abstracted and though about in a way such that we could develop a generic algorithm to solve this problem using current SPI so that we don't expose two abstractions on the same type (AsyncSPI+SPI) at the user level. 

> Perhaps you know of a way to avoid this trap, though?

Hmm, what about two channels? One for starting rendering with the display space to be rendered and another to signal it finished rendering? You'd then need a high level orchestrator to synchronize the processes, but I'm guessing maybe it could consume the display interface plus an interface that renders the scene so that the complexity is implemented once and used for all displays drivers.

Maybe it's not a long lasting solution if moving on to multiple screens but I think it's worth the effort to think about how we can keep the SPI interface minimalist. 

As far as I can tell we already have the async functionality exposed via the current SPI interface in a out-of-band sort of way.
- Users know that calling SPI.Tx will block and most likely schedule the current goroutine while it performs I/O.
- When Tx returns it is a signal that the transaction has ended. Time to deassert CS.

Something I don't like about this new interface is it leaks coupling between StartTx and Wait regarding the CS line. People who write drivers need to implement both the StartAction and Wait methods on their driver device API, and that bubbles up to the user level where users also need to write and use the StartAction and Wait methods to ensure the CS line is deasserted after the action.

As a final note here's a quote by Rob Pike from his dotGo talk "Simplicity is complicated" which I feels applies to the SPI 2 method interface design we enjoy today: "Simplicity is complicated but clarity is worth the fight".

aykevl: > > Perhaps you know of a way to avoid this trap, though?
> 
> Hmm, what about two channels? One for starting rendering with the display space to be rendered and another to signal it finished rendering? You'd then need a high level orchestrator to synchronize the processes, but I'm guessing maybe it could consume the display interface plus an interface that renders the scene so that the complexity is implemented once and used for all displays drivers.

But how does that avoid the trap I described?

Fundamentally, the problem is that the scheduler doesn't know that the goroutine that calls `Tx` should be prioritized over the goroutine that renders the next block of pixels. Otherwise it can run any possible goroutine including the rendering goroutine before the SPI transfer has started, depending on which goroutine last did a blocking call or called `runtime.Gosched()`.
To be extra clear, the problem is that the rendering goroutine can start _before I/O has started_. It would be fine if rendering only started after the SPI transaction was started, but unfortunately there is no guarantee it is (and in my experiment, it was in fact often not started before SPI was started). The `StartTx` API guarantees that the transfer has started before returning.

The only way I can think of to solve this is by adding priorities to goroutines and by setting the goroutine that calls `Tx` to a higher priority. Additionally, this would need the ability to preempt a running goroutine (such as the rendering goroutine) so that a higher priority goroutine can run.
This would be a really big change to the runtime and other parts of TinyGo which currently assume non-preemptive goroutines.

I would most certainly prefer the current simple `Tx` API over adding a new API. But unfortunately I don't see how it's possible to add the background DMA support with that API.

soypat: Aha, I think I understand now. There's no way to reliably deliver control from Tx goroutine to the waiting second goroutine renderer without unadvertently waking up an unrelated process, which would hurt FPS of the display. So the issue is with the user running more than one goroutine which would interfere with the 2 display goroutines.

I still have some questions though

* Could we instead expect users to render responsibly, as in expect them to put their other goroutines to sleep during rendering? Our goal after all is not to be efficient when running zillions of goroutines.
* Might this be an issue we have today that would eventually be solved by making TinyGo better?
  - Adding multicore support
  - Making the scheduler more sophisticated (priorities like you said). I don't know much about how the scheduler works today, but simply having the scheduler select the highest woken goroutine from available goroutines when scheduling should be enough?
  
The last point actually interests me as a talking point because that would mean that maybe by making tinygo better we can avoid this style of async design most Go developers are not familiar with in other APIs (I2S and other real time buffered applications).

aykevl: > Aha, I think I understand now. There's no way to reliably deliver control from Tx goroutine to the waiting second goroutine renderer without unadvertently waking up an unrelated process, which would hurt FPS of the display. So the issue is with the user running more than one goroutine which would interfere with the 2 display goroutines.

It's actually worse than that. It also happens with just two goroutines.
Imagine you have two goroutines, A and B. A calls `StartDrawBitmap` while B renders the next bitmap. And let's say A is currently running.

* A calls `StartDrawBitmap`
* A calls `setWindow`, which does a SPI transaction.
* A starts the SPI transaction, which blocks the goroutine, which causes the scheduler to switch over to B.
* B renders the next bitmap. Because there are no blocking calls, it renders the full bitmap before the scheduler has a chance to intervene.
* B is done, wants to send the rendered bitmap, which causes it to block and the scheduler switches over to goroutine A.
* By now A has set the window, and the SPI bus has been idle while B was busy.
* A finally starts the actual bitmap transaction.

This is not an unusual case. The window needs to be set in most (if not all) cases, which will result in the scheduler switching over to goroutine B.

I would actually be fine with it if this only happened with a third goroutine involved, but it also happens when the program is only updating the screen and nothing else.

> Adding multicore support

For the rp2040 it would solve the case where there is one sending goroutine and one rendering goroutine, but the same issue would happen when you actually try to make use of the extra core by starting another rendering goroutine.
Also, most chips aren't multicore: the rp2040 is really an exception.

> Making the scheduler more sophisticated (priorities like you said).

While this is a possible solution, it would give goroutines properties which they don't have today. I would have to look up the relevant discussions, but as far as I know all proposals to make goroutines unique in some way (by giving them an ID for example) have been rejected. I'd be very careful to add this without fully understanding why the Go designers don't want this.

Also, implementing priorities is going to be very tough because it needs preemption.

soypat: So I think there may be something I still don't understand-

> - A starts the SPI transaction, which blocks the goroutine, which causes the scheduler to switch over to B.

`A` should block on an asynchronous DMA call ideally that runs in the background while `B` starts drawing window. This means the SPI transaction for first half can be in progress while drawing the next window- by the time you finish rendering you will either have a free SPI bus or a busy one, but you never spent CPU time doing something useless since like you said drawing hogs the CPU.

So no issues up to the point where B starts it's SPI transaction? 

When B starts it's SPI transaction is when the CPU is relieved of work- what does it do? Is this the issue?

> By now A has set the window, and the SPI bus has been idle while B was busy.

I read this and it does not match up with my model of what happens, is there something I missed?

aykevl: > A should block on an asynchronous DMA call ideally that runs in the background while B starts drawing window.

Ideally yes. The problem is that there are _two_ asynchronous DMA calls. One in the `setWindow` method (to set the x/y/width of the bitmap), and the other for the actual bitmap.
So A does the first call, blocks, so the CPU switches over to B _before the bitmap has started transmitting_.
The `setWindow` call is very short so barely useful to do in the background. But because it's a blocking call, it pushes A to the background. And because we have a cooperative scheduler, A is only resumed once B is finished rendering.

soypat: > two asynchronous DMA calls. One in the setWindow method (to set the x/y/width of the bitmap), and the other for the actual bitmap.

Hmmm, what is the conceptual difference between the two DMAs? I understand there's one DMA that transmits the pixel buffer over SPI... from what you are saying there's another DMA to create the pixel data? Or to copy it (where?)?

soypat: [Oh.](https://www.youtube.com/watch?v=QyrDgEz3DR0)

Yeah, that makes a lot of sense. 

It does *seem* like the point right after setWindow in goroutine A could send over a buffered channel of length 1 to guarantee goroutine B does not steal control. B would be waiting right before starting rendering on the channel receive. This could be done vice-versa, so B would also send over a channel (maybe the same one!) when done rendering to signal A should also start rendering right after B starts SPI. 

This is considering that a send on a buffered not-full channel does not block, which if I'm not mistaken it does not (?)

kenbell: Hmm - setWindow looks like this?
```
	x += d.columnOffset
	y += d.rowOffset
	copy(d.buf[:4], []uint8{uint8(x >> 8), uint8(x), uint8((x + w - 1) >> 8), uint8(x + w - 1)})
	d.sendCommand(CASET, d.buf[:4])
	copy(d.buf[:4], []uint8{uint8(y >> 8), uint8(y), uint8((y + h - 1) >> 8), uint8(y + h - 1)})
	d.sendCommand(RASET, d.buf[:4])
	d.sendCommand(RAMWR, nil)
```

If so, that's like 3 DMA transfers just to set window size?  @aykevl design is growing on me - but i wonder if buses could/should expose a `startTx` that takes multiple buffers and does them all in one contiguous async operation.  In most cases I guess this would mean using interrupts to switch between buffers when each transaction is complete.  I guess this could theoretically be faster since `setWindow` currently does three blocking SPI transactions that could be async?

aykevl: > It does _seem_ like the point right after setWindow in goroutine A could send over a buffered channel of length 1 to guarantee goroutine B does not steal control. B would be waiting right before starting rendering on the channel receive. 

Yes, this could work. It's a bit finnicky to get right, but I don't see why this wouldn't work.
It does seem to have a lot more non-obvious complexity though. While the API may _look_ simpler, users of the API would have to introduce a lot more complexity to be fast - and it's certainly not very intuitive as evidenced by this discussion!

> This is considering that a send on a buffered not-full channel does not block, which if I'm not mistaken it does not (?)

Correct. (I'd say this is implementation defined, but it makes the most sense to not block when doing a send on a channel with space in the buffer).

> Hmm - setWindow looks like this?
> [...]
> If so, that's like 3 DMA transfers just to set window size?

Actually five, because each `sendCommand` call is actually a single byte transaction to send the command, and then (optionally, depending on the command) some parameter bytes sent in a separate transfer. And it has to be five separate transactions because the command/data pin has to be toggled between each transfer so that the display controller knows which bytes are part of the command and which are part of the command parameters.

(There is an optimization opportunity here if the window didn't change, but I'm talking about the general case).

This is also the reason why the `SetPixel` API is so slow: it would have to do six SPI transactions (13 bytes) per pixel in the common (worst) case!

> [...] I wonder if buses could/should expose a `startTx` that takes multiple buffers and does them all in one contiguous async operation.

This is actually part of the design: `StartTx` can be called multiple times before waiting for the transactions to complete. (The implementation I wrote doesn't make use of this, but it's possible).
...if it were possible, which unfortunately it isn't because of the command/data pin.

soypat: > StartTx can be called multiple times before waiting for the transactions to complete. 

Isn't there an issue with Lock contention when calling StartTx? When there's several StartTx calls the ordering is not guaranteed since they'd be waiting on a lock free which in upstream Go has no guaranteed ordering?

> While the API may look simpler, users of the API would have to introduce a lot more complexity to be fast 

While I do agree this is more work for the user of `machine.SPI` there are some other points worth mentioning:

1. Go and async: Go's design is built on channels and virtual threads. Async API design is not what's usually found in the Go ecosystem (with  exception of `exec` and few other packages) and users are more likely to be unfamiliar with this API.

2. We have one use case for this API and it may not even be strictly necessary to solve the problem. There may be a way to solve this using channels and maintaining performance characteristics of this design. 


Given these points above I'd err on the side of including this API in a external package, not `machine`, and seeing how well it fares on the basis that Yes is forever, No is temporary.

soypat:
Discussion on DMA API here FYI: https://github.com/orgs/tinygo-org/discussions/4835

src/machine/machine_rp2040_spi.go
aykevl: > Isn't there an issue with Lock contention when calling StartTx?

True, if there are multiple goroutines calling StartTx. But peripherals like SPI are not async safe anyway. Today, if more than one goroutine calls Tx on a SPI object, that would also fail in a bad way.



> 2\. We have one use case for this API

Here is an implementation that uses this API on the ST7789, that I've actually tested and verified to work:

```diff
--- a/st7789/st7789.go
+++ b/st7789/st7789.go
@@ -310,6 +310,18 @@ func (d *Device) DrawRGBBitmap8(x, y int16, data []uint8, w, h int16) error {
        return nil
 }
 
+func (d *Device) StartDrawBitmap8(x, y int16, data []uint8, w, h int16) error {
+       d.Wait()
+       k, i := d.Size()
+       if x < 0 || y < 0 || w <= 0 || h <= 0 ||
+               x >= k || (x+w) > k || y >= i || (y+h) > i {
+               return errOutOfBounds
+       }
+       d.setWindow(x, y, w, h)
+       d.startTx(data, false)
+       return nil
+}
+
 // FillRectangleWithBuffer fills buffer with a rectangle at a given coordinates.
 func (d *Device) FillRectangleWithBuffer(x, y, width, height int16, buffer []color.RGBA) error {
        i, j := d.Size()
@@ -424,6 +437,23 @@ func (d *Device) Tx(data []byte, isCommand bool) {
        d.csPin.High()
 }
 
+func (d *Device) startTx(data []byte, isCommand bool) error {
+       if isCommand {
+               d.dcPin.Low()
+       } else {
+               d.dcPin.High()
+       }
+       d.csPin.Low()
+       d.bus.StartTx(data, nil)
+       return nil
+}
+
+func (d *Device) Wait() error {
+       d.bus.Wait()
+       d.csPin.High()
+       return nil
+}
+
 // Rx reads data from the display
 func (d *Device) Rx(command uint8, data []byte) {
        d.dcPin.Low()
```

(It's old code in a git stash, but I remember having tested this and showing much better performance).

PR #3989, opened 2023-11-06
machine/usb: add support for ISERIAL descriptor

sago35:
`ISERIAL` is sent only if usb.Serial is not an empty string.

The implementation of ISERIAL was necessary for compatibility with vial-gui in [sago35/tinygo-keyboard](https://github.com/sago35/tinygo-keyboard).


deadprogram:
Thank you very much for adding this @sago35 now merging.

PR #3993, opened 2023-11-10
Bump default stack size for target pico to 8kb from 2kb

scottfeldman:
The pico-w wifi driver needs a minimum of 8kb stack size.  The default stack size for pico (and pico-w) is 2kb, so -stack-size=8kb is needed when working with wifi driver.  Unfortunately, developers working on the wifi driver forget to specify -stack-size=8kb when testing, resulting in the wifi driver lockup during initialization.  Hours (or days) pass before we realize our mistake, and then HAND smacks HEAD and we specify --stack-size=8kb.

This RP sets the default stack size to 8kb for pico (and pico-w).  This will help developers who forget -stack-size=8kb, and will help users once the pico-w wifi driver is available.

### DIFF
--- a/targets/pico.json
+++ b/targets/pico.json
@@ -4,6 +4,7 @@
     ],
     "build-tags": ["pico"],
     "serial-port": ["2e8a:000A"],
+    "default-stack-size": 8192,
     "ldflags": [
         "--defsym=__flash_size=2048K"
     ],


soypat:
This catches me off way too often when working in a project I left in the backburner that requires it. I don't think it would be strictly necessary if there was a panic message for stack overflow. Sadly it just crashes and prints out nothing.

sago35:
@scottfeldman 
Shouldn't we create a pico-w.json that includes different definitions for machine.BUTTON, even if it's just in json? Of course, I also think that changing the pico to 8kb is not a bad thing.

scottfeldman:
> @scottfeldman Shouldn't we create a pico-w.json that includes different definitions for machine.BUTTON, even if it's just in json? Of course, I also think that changing the pico to 8kb is not a bad thing.

That makes sense to me.  Would the correct approach be to make pico-w.json inherit from pico.json?

pico-w.json:

{
    "inherits": [
        "pico"
    ],
    "default-stack-size": 8192,
}

scottfeldman:
Hmmm...there's probably a lot more to change when adding a new target.  What machine things are different compared with pico?  machine.LED must be diff since the LED goes thru the wifi chip on pico-w.  What else?  Maybe it's too early for this PR...but darn it, getting burned forgetting -stack-size is a time-waster.

sago35:
I think the `pico-w.json` you mentioned earlier should work fine. Just make sure not to forget to set `-target pico-w`, and that should be good enough.

sago35:
I think it's better to add `machine/board_pico_w.go` later.

deadprogram:
Thanks for adding this @scottfeldman and to @sago35 and @soypat for review. Now merging.

Issue #3990, opened 2023-11-07
UART for nrf targets does not implement `machine.Serialer`


bgould:
This function procedure is incompatible with the `machine.Serialer` interface:

https://github.com/tinygo-org/tinygo/blob/2b215955cac75b1308d02269b8ea5b73448708a9/src/machine/machine_nrf.go#L174

When compiling with `-serial=uart`, compilation fails when attempting to use `machine.Serial` as a `machine.Serialer` like:

```
cannot use machine.Serial (variable of type *machine.UART) as machine.Serialer value in struct literal: *machine.UART does not implement machine.Serialer (wrong type for method Configure)
                have Configure(machine.UARTConfig)
                want Configure(machine.UARTConfig) error
```

bgould:
Upon further investigation ... it appears `machine.Serialer` is protected by build tags in `src/machine/usb.go` ... so maybe isn't intended for UART to implement it?  Unless there is some detail I'm missing, I think that UART should implement `machine.Serialer`.

sago35:
The machine.Serialer interface, if I recall correctly, was introduced for the purpose of tidying up implementations such as -serial=uart. Therefore, it probably does not take into account environments without USB support.

As you are writing, I believe UART should implement machine.Serialer.

PR #3995, opened 2023-11-11
WIP: Ensure UART implements `machine.Serialer`

bgould:
Updates to ensure that `machine.UART` implements `machine.Serialer`

Will resolve https://github.com/tinygo-org/tinygo/issues/3990

### DIFF
--- a/src/machine/machine_nrf.go
+++ b/src/machine/machine_nrf.go
@@ -171,7 +171,7 @@ var (
 )
 
 // Configure the UART.
-func (uart *UART) Configure(config UARTConfig) {
+func (uart *UART) Configure(config UARTConfig) error {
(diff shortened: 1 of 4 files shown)


sago35:
> Error:         sizes_test.go:76: could not build: could not compile: /Users/runner/work/tinygo/tinygo/src/machine/uart.go:26:18: cannot use (*UART)(nil) (value of type *UART) as Serialer value in variable declaration: *UART does not implement Serialer (wrong type for method Configure)

It looks like implementation is also needed in `src/machine/machine_fe310.go`.

deadprogram:
@bgould are you still working on this PR? I think the idea of it is still good, but it will require more effort to get it finished.

PR #3996, opened 2023-11-13
Support UEFI OS

jclab-joseph:
Adds support for UEFI OS.

- [X] Working minimal
- [x] findGlobals
- [X] Refactoring
- [X] exit
- [X] amd64
- [x] smoke test
- [x] target.json

not goal
- i386
- arm64 (TODO)

Usage :
```
tinygo build -target uefi-amd64 -o test.efi main.go
```

image


compileopts/target.go
```diff
@@ -375,6 +375,43 @@ func defaultTarget(goos, goarch, triple string) (*TargetSpec, error) {
 			"--no-insert-timestamp",
 			"--no-dynamicbase",
 		)
+	} else if goos == "uefi" {
```
aykevl: Please use a target JSON file. The targets defined in this file are only for the OSes natively supported by Go.

src/runtime/gc_globals.go
```diff
@@ -1,4 +1,4 @@
-//go:build (gc.conservative || gc.precise) && (baremetal || tinygo.wasm)
+//go:build (gc.conservative || gc.precise) && (baremetal || tinygo.wasm) && !uefi
```
aykevl: I would suggest using the build tag `baremetal` for UEFI, because that's basically what it is (at least how I understand it).

src/runtime/runtime_uefi.go
```diff
@@ -0,0 +1,273 @@
...
+func nanosecondsToTicks(ns int64) timeUnit {
+	return timeUnit(x86.ConvertTicksFromNanoSeconds(uefi.TicksFrequency(), uint64(ns)))
+}
+
+// timeOffset
+// TODO: use rtc
+var timeOffset int64
+
+//go:linkname now time.now
+func now() (sec int64, nsec int32, mono int64) {
```
aykevl: You shouldn't need to define this function. I suggest defining the `baremetal` build tag, with which you just need to define the `ticks` function.

src/runtime/runtime_uefi.go
```diff
@@ -0,0 +1,273 @@
...
+	April
+	May
+	June
+	July
+	August
+	September
+	October
+	November
+	December
+)
```
aykevl: Why are these constants needed?

src/runtime/runtime_uefi.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+// timeOffset
+// TODO: use rtc
+var timeOffset int64
+
+//go:linkname now time.now
+func now() (sec int64, nsec int32, mono int64) {
+	mono = nanotime()
+	efiTime, status := uefi.GetTime()
```
aykevl: I don't like how UEFI uses calendar time, but if that's what we have to work with, then so be it...

aykevl (commented):
Did a quick pass over the code (not a full review).

My main comment is this: can you make this more like any other baremetal target instead of making it its own thing besides baremetal and regular OSes? A baremetal target is much easier to maintain than what you have here (for example, the existing Nintendo Switch support is more work than any weird chip someone added support for years ago).

Also, please add a smoke test so we can see how it works.

src/runtime/gc_globals.go
jclab-joseph: FindGlobals needs to be reimplemented for UEFI.
I just pushed the code that implements findGlobals.

src/runtime/runtime_uefi.go
jclab-joseph: removed

src/runtime/runtime_uefi.go
jclab-joseph: x86 has Real-time-clock. This is to call the GetTime API that uses this.

src/runtime/runtime_uefi.go
jclab-joseph: x86 has Real-time-clock. This is to call the GetTime API that uses this.

compileopts/target.go
jclab-joseph: Actually, this is the real problem.
https://github.com/tinygo-org/tinygo/blob/dcea75d974ab5a2f5d3cdc9d039876fa780b87e9/compileopts/target.go#L436-L442
GOOS=linux works better overall in the code. However, you must use windows abi for stack.
Any good ideas?

src/runtime/os_uefi.go
```diff
@@ -0,0 +1,99 @@
...
+var module *exeHeader
+
+// Mark global variables.
+// Unfortunately, the linker doesn't provide symbols for the start and end of
+// the data/bss sections. Therefore these addresses need to be determined at
+// runtime. This might seem complex and it kind of is, but it only compiles to
+// around 160 bytes of amd64 instructions.
+// Most of this function is based on the documentation in
+// https://docs.microsoft.com/en-us/windows/win32/debug/pe-format.
+func findGlobals(found func(start, end uintptr)) {
```
aykevl, line 17: Please don't copy code. This PR is already at almost 3000 lines, which is massive (and will be a huge maintenance burden).

If you want to eventually get this merged, I suggest working on ways to simplify and slim down the code.

src/runtime/gc_globals.go
aykevl: You can still implement this as a baremetal target. The only file you have to modify is src/runtime/gc_globals.go to make sure it doesn't compile for uefi, and add a new `findGlobals` just for UEFI (that then calls shared Windows code).

compileopts/target.go
aykevl: Yes, use GOOS=linux like every other baremetal target.

> However, you must use windows abi for stack. Any good ideas?

What do you mean by this?

src/runtime/runtime_uefi.go
aykevl: Again, you shouldn't need to define this function if you define the `baremetal` build tag. In that case, the common `time.now` function in src/runtime/baremetal.go will be used instead and you only need to implement the relevant `ticks` and `ticksToNanoseconds` functions.

aykevl (commented):
This looks like it will be a huge PR, so I suggest working on ways to simplify it. It can only be merged if it is maintainable, and right now it's so invasive I honestly don't want to merge it like this.

* Don't copy code like you did for `findGlobals`. Instead, find ways to reuse it. For example, you can move it to a new gc_windows.go file that is compiled in both Windows and UEFI.
* Do _not_ make this a new GOOS. Instead, create targets/uefi.json that sets all the necessary properties (and adds at least the `uefi` and `baremetal` build tags). This probably means using `GOOS=linux`, like all the other baremetal targets. Then you can compile for UEFI using `tinygo build -target=uefi ...`.
* Add a smoke test to GNUmakefile (there are lots of smoke tests already in there).

(To be clear: there is a difference between big and invasive. Right now it modifies many parts that shouldn't need to be modified by a PR that adds a new build target. If you need to use many lines of code to implement UEFI support, that's fine as long as it doesn't affect the common code too much).

jclab-joseph:
Yes, it's not a good looking code right now.
It's a work in progress and needs a lot of refactoring.
Your opinion will help me a lot!

~~GOOS=linux is required, but with the exception uefi should use task_stack_windows_amd64 instead of task_stack_linux_amd64 . Cannot change name of task_stack.s without changing target.go.~~
I'm sorry. It was my mistake. It's being edited.

targets/uefi-amd64.json
```diff
@@ -0,0 +1,39 @@
...
+  "cflags": [
+    "-Werror",
+    "-fshort-enums",
+    "-fomit-frame-pointer",
+    "-fno-exceptions", "-fno-unwind-tables", "-fno-asynchronous-unwind-tables",
+    "-ffunction-sections", "-fdata-sections",
+    "-ffreestanding",
+    "-fshort-wchar",
+    "-mno-red-zone"
+  ],
```
aykevl, line 23: I would suggest adding `"uefi-amd64"` to builder/builder_test.go and running the tests. I suspect there will be a difference in the `"features"` JSON key (which is relevant for the ABI among other things).

xplshn:
Any progress?

jclab-joseph:
@xplshn 
Oh, I haven't had time to work on it any more lately.. :crying_cat_face:

xplshn:
> @xplshn Oh, I haven't had time to work on it any more lately.. 😿

That's okay, it's great work anyways, someone could use it as starting point to continue the port.

jclab-joseph:
FYI, here's a similar project: https://github.com/usbarmory/go-boot/

sparques:
I've been working on a fork of this: https://github.com/sparques/tinygo/tree/feat/uefi

Most notably, I fixed the assembly shim code to correctly implement the PE calling ABI and verified it works in qemu and on real hardware. Prior to the fix, "Hello World" would not run on real hardware.

I was getting crashes when trying to use -gc=precise and -scheduler=tasks and fixed that as well. I added (possibly even correctly) support for -gc=leaking and -scheduler=none.

I've also added support for Graphics Output Protocol and have seperate driver code so the screen can be drawn to via the draw.Image interface. I have my [terminal emulator](https://github.com/sparques/fansiterm) working (output, at least) as an EFI application.

I'm not sure Simple Text Input Protocol (essentially , os.Stdin) is working correctly so I'm fixing that next. I also plan on adding support for Serial IO Protocol, File Protocol, and Network Protocols in that order. 

@aykevl: What other clean up or reoganization is needed? 

Also, where should seperate driver code go? Is tinygo.org/x/drivers/uefi/ appropriate? For example, the driver that lets you wrap a Graphics Output Protocol object into one that implements draw.Image can't live in machine/ or runtime/ because of cyclic dependencies, so I was thinking tinygo.org/x/drivers/uefi/image made sense.

jclab-joseph:
@sparques Awesome! Thank you so much for your contribution. I look forward to seeing tinygo used in uefi in the future!

aykevl:
What I'm looking for is something that is somewhat minimal (ask in, not too many lines of code changed) and fits in with the rest of TinyGo. Something that has a minimum (or ideally zero) compiler changes. In other words, something that is easy to maintain. 

You gave a link to a branch, but that doesn't give me a diff. So I can't easily review that.

xplshn:
> I've been working on a fork of this: https://github.com/sparques/tinygo/tree/feat/uefi
> 
> Most notably, I fixed the assembly shim code to correctly implement the PE calling ABI and verified it works in qemu and on real hardware. Prior to the fix, "Hello World" would not run on real hardware.
> 
> I was getting crashes when trying to use -gc=precise and -scheduler=tasks and fixed that as well. I added (possibly even correctly) support for -gc=leaking and -scheduler=none.
> 
> I've also added support for Graphics Output Protocol and have seperate driver code so the screen can be drawn to via the draw.Image interface. I have my [terminal emulator](https://github.com/sparques/fansiterm) working (output, at least) as an EFI application.
> 
> I'm not sure Simple Text Input Protocol (essentially , os.Stdin) is working correctly so I'm fixing that next. I also plan on adding support for Serial IO Protocol, File Protocol, and Network Protocols in that order.
> 
> @aykevl: What other clean up or reoganization is needed?
> 
> Also, where should seperate driver code go? Is tinygo.org/x/drivers/uefi/ appropriate? For example, the driver that lets you wrap a Graphics Output Protocol object into one that implements draw.Image can't live in machine/ or runtime/ because of cyclic dependencies, so I was thinking tinygo.org/x/drivers/uefi/image made sense.

You should create a pull request (https://github.com/tinygo-org/tinygo/compare/release...sparques:tinygo:feat/uefi)

And link it here so that `@aykevl` can check the diff

NOTE: @aykevl you can already review at: https://github.com/tinygo-org/tinygo/compare/release...sparques:tinygo:feat/uefi

sparques:
I created a draft PR: https://github.com/tinygo-org/tinygo/pull/5060

I'm having trouble running the tests / figuring out what I need to fix for the tests. 

I did add support for several other UEFI protocols like Serial IO Protocol and the like, however I could not get it working in a VM or on real hardware and I'm not entirely sure that's my shortcomming--turns out while the UEFI spec is very thorough and featureful, basically no vendor fully implements it, making a UEFI target slightly less useful. But all the basics work and with enough unsafe.Pointer usage you can poke around other bits of hardware on a machine-by-machine basis.

deadprogram:
Closing since basic UEFI support has been added in several other PRs. Thanks for pioneering this effort @jclab-joseph

PR #4000, opened 2023-11-19
add `slog` standard library package support for Tinygo

soypat:
Main change added is `src/slog/value_tinygo_119.go` to bypass the use of `reflect.SliceHeader` type.

### DIFF
--- a/src/slog/attr.go
+++ b/src/slog/attr.go
@@ -1,0 +1,102 @@
+// Copyright 2022 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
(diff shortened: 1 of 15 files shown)


aykevl:
Do you mean the log/slog package? Because right now the PR adds slog, not log/slog.

It also appears that this is only an issue in older Go versions, right? Then can you explain why you need this PR? If this PR only fixes an issue with older Go versions, then I'm inclined to _not_ merge this ~3000 line PR as the problem will go away on its own eventually.

soypat:
Seems like this was fixed before meing merged from golang.org/x/exp/slog  into Go upstream. Closed

PR #4001, opened 2023-11-20
refactor: reuse OptLevel() to get the opt level

radkomih:


deadprogram:
Thanks for helping remove duplication @radkomih now merging.

PR #4007, opened 2023-11-25
lib/cmsis-svd: change to new shared location and update version

deadprogram:
This PR modifies the submodule `lib/cmsis-svd` to change to new shared location and updates to the latest version.

As a result, a small change to the makefile was needed to avoid overwriting the `esp32` implementation in the community version that we are currently using  with an official one that is now part of `cmsis-svd`.

At some point, switching to the use the official implementation would possibly be a good idea. I have not looked too deeply into that part.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,7 +9,8 @@
 	url = https://github.com/avr-rust/avr-mcu.git
 [submodule "lib/cmsis-svd"]
 	path = lib/cmsis-svd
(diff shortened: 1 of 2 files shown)


aykevl (approved):
There's an official version now? Awesome!

Yeah we should update the ESP32 to use the official SVD, but until then this is a good workaround.

deadprogram:

PR #4009, opened 2023-11-25
docker: remove lists after update to reduce image size

deadprogram:
This PR modifies the Dockerfile to remove lists after update to reduce image size.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ RUN apt-get update && \
         /var/tmp/* \
         /tmp/*
 
-COPY ./Makefile /tinygo/Makefile
+COPY ./GNUmakefile /tinygo/GNUmakefile
 
 RUN cd /tinygo/ && \
     make llvm-source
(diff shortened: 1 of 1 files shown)


deadprogram:
This was successfully built in https://github.com/tinygo-org/tinygo/tree/build-llvm-image branch.

It is now needed due to having exceeded the max size for builders in the free GH plan we are using.

Dockerfile
```diff
@@ -2,9 +2,10 @@
 FROM golang:1.21 AS tinygo-llvm
 
 RUN apt-get update && \
-    apt-get install -y apt-utils make cmake clang-15 ninja-build
+    apt-get install -y apt-utils make cmake clang-15 ninja-build && \
+    rm -rf /var/lib/apt/lists/*
```
b0ch3nski: Suggestion: more savings:
```
    rm -rf \
        /var/lib/apt/lists/* \
        /var/log/* \
        /var/tmp/* \
        /tmp/*
```

(copy-pasted from Dockerfiles I have been using since a looong time)

deadprogram: Thanks for the improvement @b0ch3nski I have modified this PR accordingly.

deadprogram:
Thanks for review and improvement @b0ch3nski now merging.

PR #4010, opened 2023-11-25
runtime: allow panic on syscall.Exit instead of halting

rminnich:
In some environments, such as u-root, code can be imported which provides a shell. The shell calls functions, instead of trying to use exec.Command (which is not available). An example can be found in github.com:u-root/u-root/tools/tinygobb.

When the command exits, the shell needs to resume operation, which is not possible when runtime.Exit calls halt.

On tamago, we achieve this by setting a normally nil function, runtime.Exit, to a funtion which panics. The shell has a recover() to catch the panics.

There are many ways in which setting runtime.Exit can go wrong, including different functions setting it to different functions.

tinygo allows a simpler, and hence more robust, model: instead of halting, we can panic. But in most cases, halting is ok.

Add a function, runtime.PanicOnExit, which enables a panic on exit. This function is idempotent, and once panic on exit is enabled, it can not be disabled. This is restrictive by design. It is hoped that by making it simple and non-revocable, we can avoid the problems that can occur with more complex mechanisms, but still provide an option to halting.

A function is provided, instead of just a variable, to allow simpler access from assembly, i.e., both variables and functions have been used in bare metal environments and, for several reasons, functions have proved more flexible.

### DIFF
--- a/src/runtime/baremetal.go
+++ b/src/runtime/baremetal.go
@@ -3,6 +3,7 @@
 package runtime
 
 import (
+	"internal/itoa"
 	"unsafe"
 )
 
@@ -59,8 +60,30 @@ func runtime_putchar(c byte) {
 	putchar(c)
 }
 
+// ErrExitPanic implements error
+type ErrExitPanic int
+
+func (e ErrExitPanic) Error() string {
+	return "errno " + itoa.Itoa(int(e))
+}
+
+var panicOnExit bool
+
+// PanicOnExit enables panic when syscall.Exit is called,
+// as opposed to halting. It can be set by, e.g., a u-root
+// shell to regain control when commands exit.
+// This in turn makes it easier for building unmodified
+// commands into u-root, for bare metal environments.
+func PanicOnExit() {
+	panicOnExit = true
+}
+
 //go:linkname syscall_Exit syscall.Exit
 func syscall_Exit(code int) {
+	if panicOnExit {
+		panic(ErrExitPanic(code))
+	}
+
 	exit(code)
 }
 


rminnich:
This program can be run on the microbit2 with: tinygo flash -target microbit-v2 -monitor .
```
// blink program for the BBC micro:bit
package main

import (
	"fmt"
	"machine"
	"runtime"
	"syscall"
	"time"
)

// The LED matrix in the micro:bit is a multiplexed display: https://en.wikipedia.org/wiki/Multiplexed_display
// Driver for easier control: https://github.com/tinygo-org/drivers/tree/master/microbitmatrix
func main() {
	runtime.PanicOnExit()
	for {
		err := g()
		fmt.Printf("did it %v\n", err)
	}
}

func g() error {

	defer func() {
		if r := recover(); r != nil {
			fmt.Println("Recovered in f", r)
		}
	}()
	ledrow := machine.LED_ROW_1
	ledrow.Configure(machine.PinConfig{Mode: machine.PinOutput})
	ledcol := machine.LED_COL_1
	ledcol.Configure(machine.PinConfig{Mode: machine.PinOutput})
	ledcol.Low()
	for {
		ledrow.Low()
		fmt.Printf("low\n")
		time.Sleep(time.Millisecond * 500)

		ledrow.High()
		fmt.Printf("high\n")
		time.Sleep(time.Millisecond * 500)
		syscall.Exit(1)
	}
}
```

produces the following:
```
low
high
Recovered in f errno 1
did it <nil>
low
high
Recovered in f errno 1
did it <nil>
low
high
Recovered in f errno 1
did it <nil>
low
high
Recovered in f errno 1
did it <nil>
```

Thus, u-root programs which now call syscall.Exit can be run as functions, on bare metal, and the shell will work.

[aside: OMG this tinygo environment is wonderful. Best firmware language I've ever used, and I've used many. 
Thanks!]

dgryski:
I'm not a firmware person but this seems reasonable.

deadprogram:
Just wondering if this is something that would be better off as a compile-time flag instead of a runtime callable function?

deadprogram:
Reminder to @aykevl to PTAL

aykevl (changes requested):
I don't think we should be doing this. It introduces a new feature that's pretty core to Go and I strongly doubt upstream Go would ever accept such a change.

(Also, what about goroutines that were started? They won't be exited with `panic`).

What if you create a new package for use by u-root that implements exactly this behavior? It can export the same `PanicOnExit` function and an `Exit` function that is called everywhere that is needed. It also has the added benefit of not requiring TinyGo to work.

deadprogram:
> What if you create a new package for use by u-root that implements exactly this behavior? It can export the same PanicOnExit function and an Exit function that is called everywhere that is needed. It also has the added benefit of not requiring TinyGo to work.

That is a very good suggestion, IMO. @rminnich what do you think?

Issue #3631, opened 2023-04-01
No serial output on xiao-esp32c3


scottfeldman:
I'm not seeing any serial output for xiao-esp32c3.

```
sfeldma@nuc:~/work/tinygo$ tinygo flash -monitor -target xiao-esp32c3 -size short  src/examples/serial/serial.go 
   code    data     bss |   flash     ram
   4502       0    8818 |    4502    8818
esptool.py v4.6-dev
Serial port /dev/ttyACM0
Connecting...
Chip is ESP32-C3 (revision v0.4)
Features: WiFi, BLE
Crystal is 40MHz
MAC: 34:85:18:00:2b:b8
Uploading stub...
Running stub...
Stub running...
Configuring flash size...
Flash will be erased from 0x00000000 to 0x00001fff...
Compressed 4592 bytes to 3264...
Wrote 4592 bytes (3264 compressed) at 0x00000000 in 0.1 seconds (effective 380.6 kbit/s)...
Hash of data verified.

Leaving...
Hard resetting via RTS pin...
Connected to /dev/ttyACM0. Press Ctrl-C to exit.
```

I expected to see "hello world!" printed to monitor every second.

No output.

Tried same on xiao board (not xiao-esp32c3) and I see output.

sago35:
USB Serial on ESP32-C3 is not available because `USB Serial/JTAG Controller` is not implemented now.

https://wiki.seeedstudio.com/XIAO_ESP32C3_Getting_Started/#resources

scottfeldman:
> USB Serial on ESP32-C3 is not available because `USB Serial/JTAG Controller` is not implemented now.
> 
> https://wiki.seeedstudio.com/XIAO_ESP32C3_Getting_Started/#resources

What would it take to add this support?

sago35:
I don't know much about it because there is not much documentation.

https://files.seeedstudio.com/wiki/XIAO_WiFi/Resources/esp32-c3_datasheet.pdf

![image](https://user-images.githubusercontent.com/9251039/229278384-c0a465b6-f60a-4116-b5f7-48e8e15e61ac.png)



Perhaps there is a hint in the source code around

https://github.com/espressif/arduino-esp32


It will probably finally be implemented in machine/machine_atsamd51.go, for example.
But first, we need to make sure that code like `xxx.WriteByte('a')` works.

deadprogram:
@sago35 you are quite correct.

Further dev info can be found here:

[...]

PR #4011, opened 2023-11-27
machine/esp32c3: add USB_SERIAL allowing USBCDC communication.

deadprogram:
This PR adds to the `machine/esp32c3` the implementation for USB_SERIAL allowing USBCDC communication.

Addresses https://github.com/tinygo-org/tinygo/issues/3631

### DIFF
--- a/src/machine/machine_esp32c3.go
+++ b/src/machine/machine_esp32c3.go
@@ -504,3 +504,95 @@ func (uart *UART) writeByte(b byte) error {
 }
 
 func (uart *UART) flush() {}
+
+type Serialer interface {
+	WriteByte(c byte) error
+	Write(data []byte) (n int, err error)
+	Configure(config UARTConfig) error
+	Buffered() int
+	ReadByte() (byte, error)
+	DTR() bool
+	RTS() bool
+}
+
+// USB Serial/JTAG Controller
+// See esp32-c3_technical_reference_manual_en.pdf
+// pg. 736
+type USB_DEVICE struct {
+	Bus *esp.USB_DEVICE_Type
+}
+
+var (
+	_USBCDC = &USB_DEVICE{
+		Bus: esp.USB_DEVICE,
+	}
+
+	USBCDC Serialer = _USBCDC
+)
+
+var (
+	errUSBWrongSize            = errors.New("USB: invalid write size")
+	errUSBCouldNotWriteAllData = errors.New("USB: could not write all data")
+	errUSBBufferEmpty          = errors.New("USB: read buffer empty")
+)
+
+func (usbdev *USB_DEVICE) Configure(config UARTConfig) error {
+	return nil
+}
+
+func (usbdev *USB_DEVICE) WriteByte(c byte) error {
+	if usbdev.Bus.GetEP1_CONF_SERIAL_IN_EP_DATA_FREE() == 0 {
+		return errUSBCouldNotWriteAllData
+	}
+
+	usbdev.Bus.SetEP1_RDWR_BYTE(uint32(c))
+	usbdev.flush()
+
+	return nil
+}
+
+func (usbdev *USB_DEVICE) Write(data []byte) (n int, err error) {
+	if len(data) == 0 || len(data) > 64 {
+		return 0, errUSBWrongSize
+	}
+
+	for i, c := range data {
+		if usbdev.Bus.GetEP1_CONF_SERIAL_IN_EP_DATA_FREE() == 0 {
+			if i > 0 {
+				usbdev.flush()
+			}
+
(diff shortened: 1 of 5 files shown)


src/machine/machine_esp32c3.go
```diff
@@ -504,3 +504,95 @@ func (uart *UART) writeByte(b byte) error {
...
+
+type Serialer interface {
+	WriteByte(c byte) error
+	Write(data []byte) (n int, err error)
+	Configure(config UARTConfig) error
+	Buffered() int
+	ReadByte() (byte, error)
+	DTR() bool
+	RTS() bool
+}
```
soypat, line 516: Don't we have a serialer.go file or similar? Seems like we could benefit from having a single place to define these kinds of interfaces for all microcontrollers to consume (by adding them in the build tags as they are implemented)

src/runtime/runtime_esp32c3.go
```diff
@@ -94,3 +99,19 @@ func interruptInit() {
...
 //go:extern _vector_table
 var _vector_table [0]uintptr
+
+func putchar(c byte) {
+	machine.Serial.WriteByte(c)
+}
+
+func getchar() byte {
+	for machine.Serial.Buffered() == 0 {
+		Gosched()
```
soypat: gosched should be unexported I believe- `runtime.Gosched` is the available version.

src/machine/machine_esp32c3.go
deadprogram: The other definition for this interface is in `machine/usb.go` but that file included some code that does not apply to the USB-CDC debug interface on the `esp32c3` chips. That was why I did not just add the build tag to that file.

src/runtime/runtime_esp32c3.go
deadprogram: This is existing code pattern from other implementations. If we want to change that it should probably be in a different PR.

src/machine/machine_esp32c3.go
sago35: In the following PR, bgould is attempting to make machine.Serialer accessible from other files.

#3995

src/runtime/runtime_esp32xx.go
```diff
@@ -4,28 +4,11 @@ package runtime
...
-		Gosched()
-	}
-	v, _ := machine.Serial.ReadByte()
-	return v
-}
-
-func buffered() int {
-	return machine.Serial.Buffered()
-}
-
```
sago35, line 28: Is there a reason for this change?
It seems that esp32 and esp32c3 share the same code.

targets/esp32c3.json
```diff
@@ -2,7 +2,7 @@
 	"inherits": ["riscv32"],
 	"features": "+32bit,+c,+m,-64bit,-a,-d,-e,-experimental-zawrs,-experimental-zca,-experimental-zcd,-experimental-zcf,-experimental-zihintntl,-experimental-ztso,-experimental-zvfh,-f,-h,-relax,-save-restore,-svinval,-svnapot,-svpbmt,-v,-xtheadvdot,-xventanacondops,-zba,-zbb,-zbc,-zbkb,-zbkc,-zbkx,-zbs,-zdinx,-zfh,-zfhmin,-zfinx,-zhinx,-zhinxmin,-zicbom,-zicbop,-zicboz,-zihintpause,-zk,-zkn,-zknd,-zkne,-zknh,-zkr,-zks,-zksed,-zksh,-zkt,-zmmul,-zve32f,-zve32x,-zve64d,-zve64f,-zve64x,-zvl1024b,-zvl128b,-zvl16384b,-zvl2048b,-zvl256b,-zvl32768b,-zvl32b,-zvl4096b,-zvl512b,-zvl64b,-zvl65536b,-zvl8192b",
 	"build-tags": ["esp32c3", "esp"],
-	"serial": "uart",
+	"serial": "usb",
```
sago35, line 5: Implementing this change may require modifying the m5stamp-c3.json side. However, I think it's not a problem for another PR.

sago35 (approved):
LGTM

I think this is a very good feature addition.

src/runtime/runtime_esp32xx.go
deadprogram: Good point. I changed back to the shared implementation.

targets/esp32c3.json
deadprogram: Can you please submit that PR and test it @sago35 as I do not have that hardware.

soypat (approved):

deadprogram:
Thanks for the reviews and improvements/feedback @sago35 and @soypat now merging.

PR #4012, opened 2023-11-27
boards: add Adafruit Qtpy-esp32c3

deadprogram:
This PR adds support for the the Adafruit QT Py ESP32-C3 WiFi Dev Board https://www.adafruit.com/product/5405

PR #4011 should be merged first and this PR rebased before merging.

PR #4014, opened 2023-11-28
machine/esp32c3: implement i2c

deadprogram:
This PR is based on the work done by @zdima in PR #2629 with a couple of fixes.

I was able to run the i2c scanner as well as being able to connect to some i2c devices on my QtPY-esp32c3

Note that PR #4012 needs to be merged and this PR rebased before merging.


deadprogram:
#4012 was merged, so this PR is ready for review, please.

src/machine/machine_esp32c3_i2c.go
```diff
@@ -0,0 +1,353 @@
...
+)
+
+var (
+	I2C0 = &I2C{}
+)
+
+type I2C struct{}
+
+// I2CConfig is used to store config info for I2C.
+type I2CConfig struct {
```
bgould, line 18: Should this struct have `Mode I2CMode` such as https://github.com/tinygo-org/tinygo/blob/731532cd2b6353b60b443343b51296ec0fafae09/src/machine/machine_nrf.go#L209-L214?

deadprogram: It needs the `Mode` added and also implemented, which I think will have to be in a separate future PR.

ysoldak (approved):
Verified works on XIAO ESP32C3 with SSD1306 128x32 display (over I2C).
The code that I verified with: https://github.com/tinygo-org/drivers/blob/release/examples/ssd1306/i2c_128x32/main.go

deadprogram:
Thanks for reviews and testing @ysoldak and @bgould now merging.

src/machine/machine_esp32c3_i2c.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+	return nil
+}
+
+// Tx does a single I2C transaction at the specified address.
+// It clocks out the given address, writes the bytes in w, reads back len(r)
+// bytes and stores them in r, and generates a stop condition on the bus.
+func (i2c *I2C) Tx(addr uint16, w, r []byte) (err error) {
+	// timeout in microseconds.
+	const timeout = 40 // 40ms is a reasonable time for a real-time system.
```
aykevl, line 332: Just discovered here that the first comment is wrong. The timeout is given in milliseconds, not microseconds.

PR #4015, opened 2023-11-28
m5stamp-c3: change settings to explicitly use UART

sago35:
Due to #4011, the default has been changed to USB, so I made the necessary corrections.
I have verified the operation on m5stamp.


deadprogram (approved):
Thanks @sago35

PR #4019, opened 2023-12-02
machine/atmega: bufferSize = 32

ysoldak:
`RingBuffer` is used for many things inside `machine` package and has size `128` bytes, hardcoded for optimisation. However, `128` bytes is quite big chunk of memory on `2k` ram targets (~6%) that many atmel chips are.
Among other things, `UART` uses `RingBuffer`, so this is an instant tax that end users can't really avoid.
Would be nice to cut this tax for targets with low memory budget.

This PR conditionally sets `RingBuffer`'s size to `32` bytes on `atmega` tagged targets.
Of course, there can be other targets with small ram and some `atmega` chips have up to `16k` ram and `128` bytes may be fine for them.
So this is a first step, and we may want to adjust the sizes conditionally further, as a follow up.

~This also updates `sizediff` tool to show ram differences and cleans up unused code.~
Change in `sizediff` extracted to a separate PR #4023

```
./tools/sizediff original.txt modified.txt
 flash                          ram
 before   after   diff          before   after   diff
   4184    4184      0   0.00%    2264    2264      0   0.00% /Users/yursol/go/bin/tinygo build -size short -o test.hex -target=pca10040            examples/blinky1
   4312    4312      0   0.00%    2264    2264      0   0.00% /Users/yursol/go/bin/tinygo build -size short -o test.hex -target=pca10040            examples/adc
   4972    4972      0   0.00%    2264    2264      0   0.00% /Users/yursol/go/bin/tinygo build -size short -o test.hex -target=pca10040            examples/blinkm
   4848    4848      0   0.00%    2268    2268      0   0.00% /Users/yursol/go/bin/tinygo build -size short -o test.hex -target=pca10040            examples/blinky2
   4196    4196      0   0.00%    2264    2264      0   0.00% /Users/yursol/go/bin/tinygo build -size short -o test.hex -target=pca10040            examples/button
   4264    4264      0   0.00%    2264    2264      0   0.00% /Users/yursol/go/bin/tinygo build -size short -o test.hex -target=pca10040            examples/button2
   4548    4548      0   0.00%    2264    2264      0   0.00% /Users/yursol/go/bin/tinygo build -size short -o test.hex -target=pca10040            examples/echo
  74624   74624      0   0.00%    3880    3880      0   0.00% /Users/yursol/go/bin/tinygo build -size short -o test.hex -target=pca10040            examples/echo2
[...]

PR #4023, opened 2023-12-03
tools/sizediff: cleanup and calculate ram

ysoldak:
This change extracted from #4019 

See result of it running in #4019


deadprogram:
Thanks for the addition @ysoldak now merging.

PR #4024, opened 2023-12-04
machine/atmega: uart double speed mode

ysoldak:
This PR enables double speed mode for atmega UARTs.
Double speed mode results in less errors and higher throughput.

In very rare cases one wants to use the slow mode.

I've encountered very high error rate when reading from an UART at `115200` baud on `16Mhz` (atmega328p[b]), resulting in inconsistent reads and data corruption.
`atmega328p[b]` chip specification indicates error rate be `3.5%` for slow mode in such setup. Switching to double speed shall bring error rate to `2.1%`. And that indeed solved my problem, no more data corruption.

Looking at the specification, I can not see any downside of using double speed mode by default.
In fact AVR's MiniCore does use double speed by default too.

---

Context: In my little hobby side project, I'm trying to interface with Digital FPV Air Units talking MSP protocol at 115200 speed.

### DIFF
--- a/src/machine/machine_atmega.go
+++ b/src/machine/machine_atmega.go
@@ -179,10 +179,18 @@ func (uart *UART) Configure(config UARTConfig) {
 		config.BaudRate = 9600
 	}
 
-	// Set baud rate based on prescale formula from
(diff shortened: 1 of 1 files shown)


deadprogram:
Thanks for adding this @ysoldak now merging.

PR #4027, opened 2023-12-04
wasi preview 2 support

dgryski:
This PR adds `-target=wasip2` support to TinyGo.  It no longer depends on `wasi-libc`, but. has its own miniature libc that turns the calls into appropriate component calls.


src/runtime/runtime_wasm_wasi.go
```diff
@@ -1,4 +1,4 @@
-//go:build tinygo.wasm && (wasi || wasip1)
```
ydnar: Should this be renamed to `runtime_wasm_wasip1.go`?

dgryski: @ydnar Yes, eventually.  That will be part of the "rip out the duplicated code" bit I'm still working on.

aykevl:
Interesting!
Suggestion: use `-target=wasip2` to more clearly specify the WASI version.

src/runtime/runtime_tinygowasmp2.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"unsafe"
+)
+
+type __wasi_io_stream_list struct {
+	buf unsafe.Pointer
+	len uint32
+}
+
+//go:wasmimport wasi:cli/stdout@0.2.0-rc-2023-11-10 get-stdout
```
johanbrandhorst: Are these modules stable? NGL I figured there'd be something like `wasi_snapshot_preview2:cli/stdout` maybe, not this 😅.

ydnar: According to @ricochet, if a component expresses a version number, then the host must support that exact version.

dgryski: The version number stuff will be solved by component registries.

ricochet: This will be `wasi:cli/stdout@0.2.0` once a second reference implementation is complete.

dgryski:
Now we just need to unstub the 29 calls in `libc_wasip2.go` with Go implementations of the associated C calls using preview2 calls under the hood.

src/syscall/env_wasip2.go
```diff
@@ -0,0 +1,51 @@
...
+	data unsafe.Pointer
+	len  uint32
+}
+
+type _string struct {
+	ptr    *byte
+	length uintptr
+}
+
+//go:wasmimport wasi:cli/environment@0.2.0-rc-2023-11-10 get-environment
```
jordan-rash: FWIW, just noticed that there was a new tag cut
https://github.com/WebAssembly/wasi-cli/releases/tag/v0.2.0-rc-2023-12-05

ydnar: We discussed in last BA SIG Go meeting to pin it to @0.2. @dgryski does that work?

ydnar:
@dgryski the HOWTO in this PR might need to be updated to reflect the new build steps (and no need for a Preview 1 adapter!):

```sh
tinygo build -target=wasip2 -x -o main.wasm ./cmd/wasip2-test
wasm-tools component embed -w wasi:cli/command $(tinygo env TINYGOROOT)/lib/wasi-cli/wit/ main.wasm -o embedded.wasm
wasm-tools component new embedded.wasm -o component.wasm
wasmtime run --wasm component-model component.wasm
```

dgryski:
This is now in a state where it's roughly equivalent to wasip1 support.  I still plan to add networking support, but that can come in a later PR.

src/syscall/libc_wasip2.go
```diff
@@ -0,0 +1,1294 @@
...
+		return newfd
+	}
+
+	// is fd a file?
+	if file, ok := wasiFiles[fd]; ok {
+		// scan for first free file descriptor
+		newfd := findFreeFD()
+		wasiFiles[newfd] = file
+		return newfd
+	}
```
lpereira: If you `dup()` a file FD, then `close()` one of the file descriptors, I would expect the dup-ed file descriptor to still work and point to the same `wasiFile` or `wasiStream`. Maybe you need some sort of refcount thing here both in the `wasiFile` and `wasiStream` structs so that if `refCount > 1`, they're only removed from the maps?

src/syscall/libc_wasip2.go
```diff
@@ -0,0 +1,1294 @@
...
+		result := stream.d.Stat()
+		if err := result.Err(); err != nil {
+			libcErrno = errorCodeToErrno(*err)
+			return -1
+		}
+		stream.offset = int64(result.OK().Size)
+	}
+
+	libcfd := nextLibcFd
+	nextLibcFd++
```
lpereira: Shouldn't this use the `findFreeFd()` function? Or am I missing something?

dgryski: Ah yes. This was code from before I had that function.

src/syscall/libc_wasip2.go
dgryski: I can add ref counting here.

src/syscall/libc_wasip2.go
```diff
@@ -0,0 +1,1294 @@
...
+
+	return int64(stream.offset)
+}
+
+// int close(int fd)
+//
+//go:export close
+func close(fd int32) int32 {
+	if _, ok := wasiStreams[fd]; ok {
+		// TODO(dgryski): Do we need to do any stdin/stdout/stderr cleanup here?
```
lpereira: Maybe call `Flush()` on the output stream, just to be sure?

loader/goroot.go
```diff
@@ -240,6 +240,7 @@ func pathsToOverride(goMinor int, needsSyscallPackage bool) map[string]bool {
 		"internal/fuzz/":        false,
 		"internal/reflectlite/": false,
 		"internal/task/":        false,
+		"internal/wasm/":        false,
```
ydnar: Do we still need this?

loader/goroot.go
```diff
@@ -259,6 +262,7 @@ func pathsToOverride(goMinor int, needsSyscallPackage bool) map[string]bool {
 
 	if needsSyscallPackage {
 		paths["syscall/"] = true // include syscall/js
+		paths["syscall/wasi/"] = false
```
ydnar: Is syscall/wasi the right place for generated WASI bindings?

main.go
```diff
@@ -1543,6 +1570,8 @@ func main() {
 		Monitor:         *monitor,
 		BaudRate:        *baudrate,
 		Timeout:         *timeout,
+		WitPackage:      witPackage,
+		WitWorld:        witWorld,
```
ydnar: Should probably rename these to WITPackage and WITWorld consistent with Go initialisms convention.

(Elsewhere too)

ydnar (commented):
Phone review, WIP

src/syscall/libc_wasip2.go
dgryski: Refcounting added and tested.

loader/goroot.go
dgryski: I wonder if we should move them to `syscall/internal` ?

main.go
dgryski: Done.

src/syscall/libc_wasip2.go
dgryski: Done.

src/syscall/libc_wasip2.go
dgryski: Ref counting added.

loader/goroot.go
ydnar: That works. The versioned package generation would ensure generating updates for 0.2.1 etc would not collide with existing code.

dgryski: Moved to `internal/wasi`

deadprogram:
Seems like this PR has some actual fail on the CI https://github.com/tinygo-org/tinygo/actions/runs/8623266604/job/23785693630

deadprogram:
@dgryski this PR has a merge conflict that needs to be resolved, if you please.

dgryski:
Just rebased against `dev` with no conflicts; let see what happes...

deadprogram:
@dgryski looks like actual test failure of some kind here https://github.com/tinygo-org/tinygo/actions/runs/8788784842/job/24119419435#step:18:36

deadprogram:
@dgryski please see this error:

```bash
main.go:1344: error: wasm-tools failed: exec: "wasm-tools": executable file not found in $PATH
```

dgryski:
@deadprogram Yup, I'm looking at the linux.yml workflow now..

.github/workflows/linux.yml
```diff
@@ -134,17 +134,19 @@ jobs:
...
         with:
           go-version: '1.22'
           cache: true
       - name: Install wasmtime
-        run: |
-          mkdir -p $HOME/.wasmtime $HOME/.wasmtime/bin
-          curl https://github.com/bytecodealliance/wasmtime/releases/download/v14.0.4/wasmtime-v14.0.4-x86_64-linux.tar.xz -o wasmtime-v14.0.4-x86_64-linux.tar.xz -SfL
-          tar -C $HOME/.wasmtime/bin --wildcards -xf wasmtime-v14.0.4-x86_64-linux.tar.xz --strip-components=1 wasmtime-v14.0.4-x86_64-linux/*
-          echo "$HOME/.wasmtime/bin" >> $GITHUB_PATH
+        uses: bytecodealliance/actions/wasmtime/setup@v1
```
deadprogram, line 145: This is better already!

.github/workflows/linux.yml
```diff
@@ -154,8 +156,8 @@ jobs:
           mkdir -p ~/lib
           tar -C ~/lib -xf tinygo.linux-amd64.tar.gz
           ln -s ~/lib/tinygo/bin/tinygo ~/go/bin/tinygo
-      - run: make tinygo-test-wasi-fast
-      - run: make tinygo-test-wasip1-fast
+      # - run: make tinygo-test-wasip1-fast
```
deadprogram: Is this actually meant to be commented out?

dgryski: Oops, no.  Was probably trying to speed up CI so I could iterate on the wasip2 failure.

tools/tgtestjson.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+# Run tests and convert output to json with go tool test2json.
+# This is a workaround for the lack of -json output in tinygo test.
+# Some variables must be set in the environment beforehand.
+# TODO: let's just add -json support to tinygo test.
```
deadprogram, line 6: :100:

deadprogram (approved):
It is very large, but it is also the result of a lot of iterating and is covered by tests.

I say we merge and then continue from there.

@aykevl any comments?

GNUmakefile
```diff
@@ -430,15 +435,40 @@ tinygo-test-wasi:
...
+
+tinygo-test-wasip2-slow:
+	$(TINYGO) test -target=wasip2 $(TEST_PACKAGES_SLOW)
+tinygo-test-wasip2-fast:
+	$(TINYGO) test -target=wasip2 $(TEST_PACKAGES_FAST) ./tests/runtime_wasi
+
+tinygo-test-wasip2-wip:
+	$(TINYGO) test -target wasip2 -x -v $(TEST_PACKAGES_FAST) ./tests/runtime_wasi
+tinygo-test-wasip2-dev:
+	$(TINYGO) test -target wasip2 -wit-package $$(tinygo env TINYGOROOT)/lib/wasi-cli/wit/ -wit-world wasi:cli/command -x -work encoding/csv
```
aykevl: 'wip' as in "work in progress"? Was this meant to be included in the PR?
Also, `-x -work` suggests this was mostly useful during testing?

builder/build.go
```diff
@@ -842,6 +846,80 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
...
+				if err != nil {
+					// Preserve a copy of the failure if -work option is used
+					if config.Options.Work {
+						dest := filepath.Join(
+							filepath.Dir(result.Executable),
+							fmt.Sprintf("%s.embed-fail.wasm", strings.Replace(pkgName, "/", "_", -1)),
+						)
+						copyCmd := exec.Command("cp", "-f", result.Executable, dest)
+						copyCmd.Run()
+					}
```
aykevl: Why do you need to copy anything here?
If anything, it would make sense to me to work from the work directory and copy a binary _out_ if it succeeds (so that if there's an error, no further action is needed).

Also, `cp` is specific to Linux/Unix systems (not Windows I think) and Go is perfectly capable of copying a file itself. If you need to copy anything you can use built-in functions for that (for example, there's `copyFile` in main.go).

compiler/symbol.go
```diff
@@ -375,19 +375,26 @@ func (c *compilerContext) checkWasmImport(f *ssa.Function, pragma string) {
 
 // Check whether the type maps directly to a WebAssembly type, according to:
 // https://github.com/golang/go/issues/59149
+// TODO(ydnar): document why we relaxed this for WASI Preview 2.
```
aykevl: This looks like a TODO that would need fixing?
I would certainly like to know why we're diverging from the Go spec for this (and why removing these checks for specific packages like is done for syscall/js only isn't enough). If reasonably possible, I'd like to stick to the Go spec for this.

compiler/symbol.go
```diff
@@ -375,19 +375,26 @@ func (c *compilerContext) checkWasmImport(f *ssa.Function, pragma string) {
...
+			return true
+		case types.Int8, types.Uint8, types.Int16, types.Uint16, types.Int32, types.Uint32, types.Int64, types.Uint64:
 			return true
 		case types.Float32, types.Float64:
 			return true
-		case types.UnsafePointer:
-			if !isReturn {
-				return true
-			}
+		case types.Uintptr, types.UnsafePointer:
```
aykevl, line 393: Pointers and uintptr are 64-bit in standard Go while they are 32-bits in TinyGo. Therefore, whitelisting them here will introduce an incompatibility which I think is a bad thing.

compiler/testdata/errors.go
```diff
@@ -16,14 +16,6 @@ type Uint uint32
 //go:wasmimport modulename validparam
 func validparam(a int32, b uint64, c float64, d unsafe.Pointer, e Uint)
 
-// ERROR: //go:wasmimport modulename invalidparam: unsupported parameter type int
-// ERROR: //go:wasmimport modulename invalidparam: unsupported parameter type string
-// ERROR: //go:wasmimport modulename invalidparam: unsupported parameter type []byte
-// ERROR: //go:wasmimport modulename invalidparam: unsupported parameter type *int32
-//
-//go:wasmimport modulename invalidparam
-func invalidparam(a int, b string, c []byte, d *int32)
```
aykevl: Why remove these tests? I think they're important.

main.go
```diff
@@ -876,7 +899,26 @@ func buildAndRun(pkgName string, config *compileopts.Config, stdout io.Writer, c
...
+			// and allow reading from the current directory.
+			switch config.Options.Target {
+			case "wasip1":
+				emuArgs = append(emuArgs, "--dir=.")
+			case "wasip2":
+				dir := result.MainDir
+				if isSingleFile {
+					cwd, _ := os.Getwd()
+					dir = cwd
+				}
```
aykevl, line 916: That's an odd distinction here, does this match upstream go and why is this distinction needed?

main.go
```diff
@@ -1464,6 +1506,12 @@ func main() {
 		flag.StringVar(&outpath, "o", "", "output filename")
 	}
 
+	var witPackage, witWorld string
+	if command == "help" || command == "build" || command == "test" {
+		flag.StringVar(&witPackage, "wit-package", "", "wit package for wasm component embedding")
+		flag.StringVar(&witWorld, "wit-world", "", "wit world for wasm component embedding")
+	}
```
aykevl: What about subcommands like `run` or `info`?

src/runtime/runtime_tinygowasmp2.go
```diff
@@ -0,0 +1,80 @@
...
+func putchar(c byte) {
+	putcharBuffer[putcharPosition] = c
+	putcharPosition++
+	if c == '\n' || putcharPosition >= putcharBufferSize {
+		list := cm.NewList(&putcharBuffer[0], putcharPosition)
+		result := putcharStdout.BlockingWriteAndFlush(list)
+		if err := result.Err(); err != nil {
+			// TODO(ydnar): handle error case
+			panic(*err)
+		}
```
aykevl: Does it even make sense to do anything with the error here? The `panic` will just call `putchar` again leading to a stack overflow (probably).
Is there any sensible case under which writing to stdout will fail and there is a reasonable fallback?
Honestly I'd suggest just ignoring the error - you'll notice soon enough if there's no output.

src/runtime/runtime_wasm_wasip2.go
```diff
@@ -0,0 +1,71 @@
...
+	"internal/wasi/cli/v0.2.0/environment"
+	monotonicclock "internal/wasi/clocks/v0.2.0/monotonic-clock"
+)
+
+type timeUnit int64
+
+// libc constructors
+//
+//export __wasm_call_ctors
+func __wasm_call_ctors()
```
aykevl: What does this do, considering that we don't use a libc?

src/runtime/runtime_wasm_wasip2.go
```diff
@@ -0,0 +1,71 @@
...
+	return 0
+}
+
+//export _start
+func _start() {
+	// These need to be initialized early so that the heap can be initialized.
+	heapStart = uintptr(unsafe.Pointer(&heapStartSymbol))
+	heapEnd = uintptr(wasm_memory_size(0) * wasmPageSize)
+	run()
+}
```
aykevl, line 26: Any reason not to inline _start into `__wasi_cli_run_run` and remove `_start`?
If it's to avoid a linker error, you can use `--no-entry`.

src/syscall/env_libc.go
```diff
@@ -0,0 +1,59 @@
+//go:build (darwin || nintendoswitch || wasip1) && !wasip2
```
aykevl: Doesn't `wasip1` imply that `wasip2` is not set?
In other words, I think `!wasip2` is not needed. (I see this pattern in a few other files too).

src/syscall/errno_wasip2.go
```diff
@@ -0,0 +1,7 @@
+//go:build wasip2
+
+package syscall
+
+// The errno for libc_wasip2.go
+
+var libcErrno Errno
```
aykevl, line 7: ~Where does this error come from, if not from a libc?~ nevermind answered later.

src/syscall/syscall_libc_wasi.go
```diff
@@ -118,11 +119,8 @@ const (
 	PATH_MAX = 4096
 )
 
-//go:extern errno
-var libcErrno uintptr
-
 func getErrno() error {
-	return Errno(libcErrno)
+	return libcErrno
 }
```
aykevl: ~Instead of moving `libcErrno`, what about moving `getErrno` to a separate wasip1-only file? That way we don't have a misleading `libcErrno` variable (misleading because it doesn't actually refer to an errno variable in an external libc).~

Oh I see now, you've used `libcErrno` in many places. I guess that works if we want to maintain some sort of libc-compatibility?
Maybe a comment would help to clarify this?

src/syscall/libc_wasip2.go
```diff
@@ -0,0 +1,1355 @@
...
+		return nil
+	}
+
+	if size < uint(len(cwd)+1) {
+		libcErrno = ERANGE
+		return nil
+	}
+
+	// TODO(dgryski): null termination?
+	copy(unsafe.Slice(buf, size), cwd)
```
aykevl: As long as the only caller is `Getwd` this seems fine (because the buffer is implicitly zeroed). Still feels a bit risky.

aykevl (commented):
Read through most of the PR, and skimmed libc_wasip2.go. Looks reasonable to me though I have a few concerns here and there.

And some more comments that I can't really do inline:

* Can you add wasip2 to builder/builder_test.go? That way we can be sure the LLVM flags etc are the same between Clang and TinyGo.
* I see a lot of places where you use `//go:export`. This is equivalent to `//export`, and deviates to from upstream Go. I'd prefer using `//export` everywhere.  
    (I should never have added `//go:export` in the first place, I was just being lazy...)

compiler/symbol.go
ydnar: The changes here correspond to and informed this proposal in Go: https://github.com/golang/go/issues/66984

ydnar: We can update the comment to point at the proposal.

compiler/symbol.go
ydnar: These will be 32 bits wide in standard Go with `GOARCH=wasm32`.

compiler/testdata/errors.go
ydnar: They're mostly no longer valid with the relaxed types in wasm args. We can re-add the slice test.

src/runtime/runtime_wasm_wasip2.go
ydnar: Nothing. We could probably remove it.

src/runtime/runtime_wasm_wasip2.go
ydnar: Eventually the runtime initialization will move to _initialize and be called by component initialization before a Run export is called.

src/syscall/env_libc.go
ydnar: This may predate the change to eliminate the wasi build tag. We'll fix.

compiler/symbol.go
ydnar: Done.

builder/build.go
ydnar: @omnide ^^

src/syscall/syscall_libc_wasi.go
dgryski: Done

main.go
dgryski: This is the logic that allows "tinygo run a/b/c/foo.go" to run with the current working directory, as opposed to having `a/b/c/`.  Upstream also has this restrictions for `go run`, and also why package names can't have `.go` at the end.

GNUmakefile
dgryski: Yup, these are no longer needed.  Removed.

src/runtime/runtime_tinygowasmp2.go
dgryski: Removed.

dgryski:
@aykevl PTAL.  I think we've addressed everything.

builder/build.go
deadprogram: @ydnar @omnide the `copyFile()` function that @aykevl is referring to is located here https://github.com/tinygo-org/tinygo/blob/dev/main.go#L75

deadprogram:
@dgryski see https://github.com/tinygo-org/tinygo/pull/4027/files#r1589100941

builder/build.go
dgryski: @deadprogram I think this will actually just be removed; we don't need it anymore.

deadprogram: Even better! :scissors:

omnide: I don't believe these copies are valuable, so let's just remove them.
https://github.com/dgryski/tinygo/pull/15

src/runtime/runtime_wasm_wasi.go
deadprogram: @dgryski is this still something needed to be done?

builder/build.go
deadprogram: Thanks!

src/runtime/runtime_wasm_wasi.go
dgryski: Pretty sure this file doesn't exist anymore.

deadprogram:
@aykevl any other comments before we merge?

compiler/symbol.go
```diff
@@ -375,19 +375,28 @@ func (c *compilerContext) checkWasmImport(f *ssa.Function, pragma string) {
...
+			return true
+		case types.String:
+			return true
 		}
+	case *types.Struct:
+		return true
```
aykevl: This is actually not part of https://github.com/golang/go/issues/66984 anymore (though it is when it's a pointer-to-struct and pointer-to-array, this probably needs some more logic):

> ### Supporting `struct` and `[...]T` by value
> 
> A previous version of this proposal included support for passing `struct` and `[...]T` types by value by expanding each field recursively into call parameters. This was removed in favor of a simpler initial implementation but could be readded if users require it.

compiler/symbol.go
```diff
@@ -375,19 +375,28 @@ func (c *compilerContext) checkWasmImport(f *ssa.Function, pragma string) {
...
 		case types.Float32, types.Float64:
 			return true
-		case types.UnsafePointer:
-			if !isReturn {
-				return true
-			}
+		case types.Uintptr, types.UnsafePointer:
+			return true
+		case types.String:
+			return true
```
aykevl: Note that this works more or less by accident (because the way both Go and TinyGo represent strings is the most obvious way to do it so they match).

compiler/symbol.go
aykevl: I see. It'll still be slightly incompatible but only because TinyGo uses 32-bit pointers in GOOS=wasm (so GOOS=wasm in TinyGo behaves like GOOS=wasm32). Seems fine by me.

compiler/testdata/errors.go
aykevl: I'd prefer if these tests were updated instead of removed altogether. ABIs are tricky to get right, and I'd like to have some tests for them.
For example: slices, maps, structs, and func values are still not allowed.

To make it really perfect, it'd be nice to add some tests to compiler/testdata/pragma.go to show that the compiler indeed produces the expected LLVM IR for the given types. Not a blocker though.

compiler/testdata/errors.go
```diff
@@ -16,14 +16,6 @@ type Uint uint32
 //go:wasmimport modulename validparam
 func validparam(a int32, b uint64, c float64, d unsafe.Pointer, e Uint)
```
aykevl: Might be a good idea to update this to add some new types (like `string` and `*struct{}`).

src/runtime/runtime_wasm_wasip2.go
aykevl: Is there a reason we can't do this now already?

src/runtime/runtime_wasm_wasip2.go
aykevl: I'd prefer if it was removed then (though that's not a blocker, it just removes noise).

src/runtime/runtime_wasm_wasip2.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Read the command line arguments from WASI.
+// For example, they can be passed to a program with wasmtime like this:
+//
+//	wasmtime run ./program.wasm arg1 arg2
+func init() {
+	__wasm_call_ctors()
+}
```
aykevl: If `__wasm_call_ctors()` does nothing, the comment is misleading.

aykevl (commented):
Having a proposal eases my worries about relaxing the `//go:wasmimport` rules, though I still have a few small things about the code that I'd like to see improved. (I'm careful with this because experience tells me that once the compiler allows something, it's very difficult to restrict it again in the future - it's much easier the other way round if needed!).

Other than that, it's mostly just minor nits.

compiler/symbol.go
ydnar: This also matches how WASI 0.2 represents strings (ptr, len pair)

compiler/symbol.go
aykevl: Also, now I think about it, we should _definitely_ be stricter here and also check every field in the struct. Otherwise `func(x *struct{field [2]int})` would be legal which it shouldn't be.

See: https://github.com/golang/go/issues/66984#issuecomment-2107661575

src/runtime/runtime_wasm_wasip2.go
dgryski: `__wasm_call_ctors()` stuff removed.

dgryski:
Is there anything else left unaddressed before we merge?

aykevl:
@dgryski yes there is, namely with `//go:wasmimport`. We are currently incompatible with the [upstream proposal for loosening `//go:wasmimport`](https://github.com/golang/go/issues/66984):

 * The upstream proposal doesn't allow passing structs as-is.
 * Also, the proposal now requires `structs.HostLayout` markers in every struct to ensure they match the underlying platform. The only thing TinyGo needs to do here is to check that the field is present: I haven't checked specifically but I'm pretty sure we already match the Canonical ABI for the in-memory layout.
 * We don't validate whether struct fields are allowed. So for example the following code would be allowed, even though it clearly is invalid:
    ```go
    //go:wasmimport foo bar
    func bar(x *struct{a func()})
    ```

We could decide to deviate from the upstream proposal (especially when it comes to passing structs by value) but if we do that I'd like to see at least a good explanation for that (and the comment should be more than just "this reflects the relaxed type restrictions proposed here" because it doesn't).

deadprogram:
@dgryski can you please rebase against the latest `dev` so this branch can build using LLVM 18?

dgryski:
Rebased against dev and squashed.

dgryski:
@aykevl So what's the easiest set of short-term hacks to get this merged?  Merge with a note indicating the current incompatibilities but that we will update it to match upstream Go before the next Go release that properly includes support for wasip2?

deadprogram:
@dgryski please rebase against `dev` again and correct the merge conflict.

> Merge with a note indicating the current incompatibilities but that we will update it to match upstream Go before the next Go release that properly includes support for wasip2?

IMO if we do that, we can merge this right now. wasi *preview* in the naming even is a signal to developers that they may have to deal with at least some breaking changes along the way. And as long as we are compatible with the lastest shipping Go releases then we have done the correct thing.

deadprogram:
@dgryski what did you think about my last request/comment?

dgryski:
@deadprogram Yeah, I think that's fine.  Big Go doesn't have wasip2 support yet, so we're leading the charge here.  There are proposals (some of whic have been accepted) but we're still in unknown territory.  "Nobody" should be using this yet for any real workload.

aykevl:
@dgryski but you said yourself that passing structs wasn't going to work with WASI and that more work was needed to lower to primitive types?

What about a different solution: the relaxed `//go:wasmimport` rules are reverted and instead the specific packages in-tree (`internal/wasi/*`) are whitelisted to allow any type?

To be clear, it's these two lines that are a problem:

```go
	case *types.Struct:
		return true
```

This needs to be more strict. We can't just pass any struct, because there is no defined ABI for passing structs and there is no checking whether these fields are allowed. The other changes (loosening integers and pointers) seem safe to me.

> So what's the easiest set of short-term hacks to get this merged? Merge with a note indicating the current incompatibilities but that we will update it to match upstream Go before the next Go release that properly includes support for wasip2?

The relaxed constraints of `//go:wasmimport` aren't limited to wasip2, but apply to any wasm target. So then we'd be allowing things now and possibly breaking programs again in the future once we (inevitably) restrict things again. A short term hack would be to revert these changes to `//go:wasmimport` and instead disable checking for `internal/wasi/*` packages (they're internal so we can do with them whatever we want).

deadprogram:
@dgryski @aykevl from what I can tell, all of the outstanding comments have been addressed, the branch is rebased, and is passing all tests.

Are we ready to merge?

dgryski:
@deadprogram I still need to merge @ydnar 's https://github.com/dgryski/tinygo/pull/16 PR and rebase, then I think we're good.

aykevl:
I'd like to take another look once that PR is merged, but initially it seems to resolve most of my concerns.

deadprogram:
@aykevl please take a look now.

compiler/symbol.go
```diff
@@ -373,21 +373,42 @@ func (c *compilerContext) checkWasmImport(f *ssa.Function, pragma string) {
...
 			return true
 		case types.Float32, types.Float64:
 			return true
-		case types.UnsafePointer:
-			if !isReturn {
-				return true
+		case types.Uintptr, types.UnsafePointer:
+			return true
+		case types.String:
+			return isPointerOrField
```
aykevl: Passing strings as parameters is allowed by the spec (and will do the right thing if I'm not mistaken).

compiler/symbol.go
```diff
@@ -373,21 +373,42 @@ func (c *compilerContext) checkWasmImport(f *ssa.Function, pragma string) {
...
+		}
+	case *types.Array:
+		return isPointerOrField && isValidWasmType(typ.Elem(), true)
```
aykevl: This would allow arrays as struct fields, which doesn't seem to be allowed by the spec. I think this is an oversight in the spec because I don't see why arrays in struct fields would be disallowed while pointers to arrays are allowed.

I'd suggest ignoring this, or even better proposing a small change to the spec.

compiler/testdata/errors.go
aykevl: It would still be nice to update these tests instead of removing them.

aykevl (commented):
I mostly looked at compiler/symbol.go which looks good to go (but with some optional suggestions).

Were there any other changes?

compiler/symbol.go
johanbrandhorst: I agree that this is an oversight in the spec, as long as the argument is a pointer to a struct, that struct should be fine to contain array elements. I will update the proposal.

aykevl: @johanbrandhorst thank you!

deadprogram:
@dgryski please resolve merge conflicts. Hopefully for the last time!

compiler/testdata/errors.go
ydnar: Agreed. I'll TAL

rajatjindal:
hello folks, thank you for your amazing work on this project as well as this feature. 👍 

I was trying out the latest changes in this branch, and ran into a problem that I thought I will mention here. 

when building a project for `target wasip2`, I am getting following error:

```
# os
/Users/rajatjindal/go/src/github.com/tinygo-org/tinygo/build/release/tinygo/src/os/stat_unix.go:28:6: statNolog redeclared in this block
/Users/rajatjindal/go/src/github.com/tinygo-org/tinygo/build/release/tinygo/src/os/stat_other.go:15:6:  other declaration of statNolog
/Users/rajatjindal/go/src/github.com/tinygo-org/tinygo/build/release/tinygo/src/os/stat_unix.go:41:6: lstatNolog redeclared in this block
/Users/rajatjindal/go/src/github.com/tinygo-org/tinygo/build/release/tinygo/src/os/stat_other.go:20:6:  other declaration of lstatNolog
/Users/rajatjindal/go/src/github.com/tinygo-org/tinygo/build/release/tinygo/src/os/stat_unix.go:15:16: method File.Stat already declared at /Users/rajatjindal/go/src/github.com/tinygo-org/tinygo/build/release/tinygo/src/os/stat_other.go:10:16
```

which I was able to fix by making following change to `stat_other.go`:

```
diff --git a/src/os/stat_other.go b/src/os/stat_other.go
index d3e0af6e..234c59db 100644
--- a/src/os/stat_other.go
+++ b/src/os/stat_other.go
@@ -1,4 +1,4 @@
-//go:build baremetal || (tinygo.wasm && !wasip1)
+//go:build baremetal || (tinygo.wasm && !wasip1 && !wasip2)
 
 // Copyright 2016 The Go Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
```

could you please confirm if that is the right fix (and if so, if you could pls include that fix in changes) or kindly point me in the right direction.

thank you

compiler/testdata/errors.go
ydnar: I updated the tests, will submit a PR.

compiler/symbol.go
ydnar: Fixed in https://github.com/dgryski/tinygo/pull/17

compiler/testdata/errors.go
ydnar: Fixed in https://github.com/dgryski/tinygo/pull/17

compiler/testdata/errors.go
ydnar: https://github.com/dgryski/tinygo/pull/17

dgryski:
@rajatjindal Thanks.  Those build tags must have broken on one my rebases.  Fixed now.

.github/workflows/linux.yml
```diff
@@ -134,17 +134,19 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          submodules: true
```
aykevl, line 138: This should not be necessary.
The whole point of test-linux-build is that it uses the release tarball created in build-linux and tests whether it works. If you need these submodules, it seems like some files aren't included in the release tarball that should be in there.

src/crypto/rand/rand_arc4random.go
```diff
@@ -1,4 +1,4 @@
-//go:build darwin || tinygo.wasm
+//go:build darwin || wasip1 || wasip2 || wasm
```
aykevl, line 1: This removes support for crypto/rand in wasm_unknown, which is probably a good thing (since it doesn't have an underlying system to read random values from).

src/runtime/runtime_tinygowasmp2.go
```diff
@@ -0,0 +1,76 @@
+//go:build wasip2
+
```
aykevl, line 2: What about merging this file and runtime_wasm_wasip2.go?
Or was there a reason you kept them separate?

aykevl (approved):
I did an in-depth review and it looks good to me.

This is probably far too late (I don't want to delay this PR any further), but what about _not_ checking in all those generated files?
Alternatives:
- generating the files on build, like with `gen-device`
- putting those files in a separate repository

It's not a big issue, but I feel like adding so many generated files in the repo leads to churn. But maybe that's just me.

src/runtime/runtime_tinygowasmp2.go
dgryski: Probably just an artifact of how the port was initially written.  We can do a bit more cleanup later.

.github/workflows/linux.yml
dgryski: This is probably needed for the generated files which were originally a submodule.

ydnar: Package `cm` is a submodule in `vendor`. Should we hard vendor it?

ydnar:
> I did an in-depth review and it looks good to me.
> 
> This is probably far too late (I don't want to delay this PR any further), but what about _not_ checking in all those generated files? Alternatives:
> 
> * generating the files on build, like with `gen-device`
> * putting those files in a separate repository
> 
> It's not a big issue, but I feel like adding so many generated files in the repo leads to churn. But maybe that's just me.

The generated files are part of (and compiled with every program that uses) the standard library. I guess we could generate them in CI and build. We’d need to add Rust dependencies (`wasm-tools`), and the standard library wouldn’t build for `GOOS=wasip2` unless those files are present.

On the other hand, the diff history of the generated files might be useful over time.

Do you have an opinion on whether package `cm` should be copied into `src/vendor`, rather than a submodule?

adamreese:
Hi folks! I've been doing some experiments using this branch and it's some awesome work.

I noticed `init()` aren't called when using export functions other than `_start`. When using Stdin/Stdout/Stderr I'm getting a panic because they're being initialized inside an `init()`  https://github.com/tinygo-org/tinygo/pull/4027/files#diff-519198709ce5f198021f73cc0136481092b1104ed96e2f21c3bbbd0ffe819d9dR132

Here's an example and the stacktrace https://github.com/adamreese/component-example/blob/stdout-issue/adder/main.go

```
$ wasmtime run final.wasm 1 2 add
Error: failed to run main module `final.wasm`

Caused by:
    0: failed to invoke `run` function
    1: error while executing at wasm backtrace:
           0: 0xdf617 - wit-component:shim!indirect-wasi:io/streams@0.2.0-[method]output-stream.blocking-write-and-flush
           1: 0x9aab2 - (internal/wasi/io/v0.2.0/streams.OutputStream).BlockingWriteAndFlush
                           at /Users/areese/p/src/github.com/tinygo-org/tinygo/src/internal/wasi/io/v0.2.0/streams/streams.wit.go:314:46
           2: 0x9fb18 - syscall.writeStream
                           at /Users/areese/p/src/github.com/tinygo-org/tinygo/src/syscall/libc_wasip2.go:205:45
           3: 0x9f7cb - write
                           at /Users/areese/p/src/github.com/tinygo-org/tinygo/src/syscall/libc_wasip2.go:46:21
           4: 0x9a66d - syscall.Write
                           at /Users/areese/p/src/github.com/tinygo-org/tinygo/src/syscall/syscall_libc.go:32:16              - (os.unixFileHandle).Write
                           at /Users/areese/p/src/github.com/tinygo-org/tinygo/src/os/file_anyos.go:110:24
           5: 0x9a0ac - docs:adder/add@0.1.0#add
                           at /Users/areese/p/src/github.com/tinygo-org/tinygo/src/os/file.go:162:14
           6:   0xc4 - <unknown>!<wasm function 2>
           7: 0x94f84 - calculator.wasm!docs:calculator/calculate@0.1.0#eval-expression
           8:   0xad - <unknown>!<wasm function 1>
           9: 0x60c0 - command-6059807ff4c3095a.wasm!command::main::h86d1f8485843e4d5
          10: 0x3182 - command-6059807ff4c3095a.wasm!std::sys_common::backtrace::__rust_begin_short_backtrace::h61307f8ecbd66741
          11: 0x3191 - command-6059807ff4c3095a.wasm!std::rt::lang_start::{{closure}}::h19e864a4fb53b568
          12: 0x54ba2 - command-6059807ff4c3095a.wasm!std::rt::lang_start_internal::hdc5a291adcbf2591
          13: 0x787e - command-6059807ff4c3095a.wasm!__main_void
          14: 0x253a - command-6059807ff4c3095a.wasm!_start
          15: 0x8ddd3 - wit-component:adapter:wasi_snapshot_preview1!wasi:cli/run@0.2.0#run
    2: unknown handle index 0
make: *** [run] Error 1
```

ydnar:
We could have the code export both _start and _initialize, but it would break on certain runtimes (e.g. Node).

This is a separate area of development (see https://github.com/golang/go/issues/42372 and https://github.com/tinygo-org/tinygo/pull/4082).

Ideally it could detect missing func main() and export _initialize instead of _start.

deadprogram:
I would like to propose that we squash/merge this PR now, and make any further changes in smaller PRs.

@dgryski do you want to do this? :smile_cat:

aykevl:
> I noticed `init()` aren't called when using export functions other than `_start`. When using Stdin/Stdout/Stderr I'm getting a panic because they're being initialized inside an `init()` https://github.com/tinygo-org/tinygo/pull/4027/files#diff-519198709ce5f198021f73cc0136481092b1104ed96e2f21c3bbbd0ffe819d9dR132

~This is a separate issue, can you create a new issue if needed?
But to explain this issue: this is working as intended. You always need to call `_start` before calling anything else: `_start` initializes the runtime (among other things).
For some background, see: https://github.com/WebAssembly/WASI/issues/19~

@dgryski @ydnar I think this is working as intended? I think `_start` is still the function that must be called on init to initialize the runtime and such?

ydnar:
_start should call init and then main.

_initialize should call init (and initialize the runtime) but not call main. Host expects to call other exported functions.

.github/workflows/linux.yml
aykevl: @ydnar no that seems fine, that directory is in src/ and is already copied to the release directory.

aykevl:
Yeah I think we should just squash+merge this.

dgryski:
🎉

Mossaka:
Hey folks, this is huge! Thank y'all for your dedication and effort 🎉🥂

johanbrandhorst:
Congrats everyone on hitting this milestone!!!

PR #4031, opened 2023-12-07
modules: update net submodule to latest commit with http Client Transport interface

deadprogram:
This PR updates the `net` submodule to the latest commit, that includes `http.Client` `Transport` and `RoundTripper` interface.

It is needed for Fermyon Spin and any other WASI code that might need these interfaces.


deadprogram:
Merging since this is both needed, and also since it is mostly just code from upstream.

brendandburns:
@deadprogram because the client code in tinygo doesn't use the `Transport` interface for round tripping, it's not feasible to use the standard go HTTP client with the `wasi-http` specification.

Would you be open to a PR that adds support for a user supplied round tripper in `Client`? I'd be happy to send the PR.

PR #4032, opened 2023-12-08
runtime: stub out Breakpoint() function

deadprogram:
This PR stubs out `runtime.Breakpoint()` just to help more code compile.

### DIFF
--- a/src/runtime/debug.go
+++ b/src/runtime/debug.go
@@ -18,3 +18,8 @@ func NumCgoCall() int {
 func NumGoroutine() int {
 	return 1
 }
+
+// Stub for Breakpoint, does not do anything.
+func Breakpoint() {
+	panic("Breakpoint not supported")
+}


src/runtime/debug.go
```diff
@@ -18,3 +18,7 @@ func NumCgoCall() int {
 func NumGoroutine() int {
 	return 1
 }
+
+// Stub for Breakpoint, does not do anything.
+func Breakpoint() {
+}
```
aykevl: I'm not sure I understand the meaning of `runtime.Breakpoint()`. While in a debugger it's supposed to trap, but outside of it I'm not sure? Perhaps it's best to just panic here until a proper breakpoint instruction can be inserted (`bkpt` in ARM for example).

deadprogram: OK added panic.

deadprogram:
Thanks for review @aykevl changes made, so now merging.

PR #4033, opened 2023-12-08
builder/picolib: add needed file for compiling math functions with error support

deadprogram:
This PR updates builder/picolib to add a needed file for compiling math functions with error support.

Thanks to @aykevl for actually finding and providing this fix, I really just reported the problem and tested that the fix worked as expected.


builder/picolibc.go
```diff
@@ -337,6 +337,7 @@ var picolibcSourcesLarge = []string{
 	"libm/common/math_err_may_uflow.c",
 	"libm/common/math_err_check_uflow.c",
 	"libm/common/math_err_check_oflow.c",
+	"libm/common/math_errf_invalidf.c",
```
aykevl, line 341: We might want to add all the other `math_err_*.c` files here too.

aykevl (commented):
It would be nice to have a test for this fix. Can you add a variant of the testcase to testdata/math.go? (basically `float32(math.Sqrt(float64(someValue)))`).

builder/picolibc.go
deadprogram: OK doing it now.

deadprogram: Done.

deadprogram:
Thanks for the help resolving this @aykevl now merging.

PR #4037, opened 2023-12-10
stub out more types/funcs to compile against golang.org/x/net/internal/socket

scottfeldman:
These are changes need to compile github.com/domainr/dnsr/ with TinyGo. See issue https://github.com/tinygo-org/net/issues/14.

These change are mostly to fix missing symbols in src/crypto/tls and src/net.  Missing types and functions are cut-and-pasted from go1.21.4. Functions are stubbed out returning errors.New("not implemented").

DNRS is compiled by running tinygo test:

`   sfeldma@nuc:~/work/dnsr$ tinygo test -target=wasi`

With this patch, and a corresponding patch for tinygo-org/net to fixup src/net, you should get a clean compile.

### DIFF
--- a/src/crypto/tls/common.go
+++ b/src/crypto/tls/common.go
@@ -6,7 +6,442 @@
 
 package tls
 
+import (
+	"context"
+	"crypto"
+	"crypto/x509"
+	"io"
+	"net"
+	"sync"
+	"time"
+)
+
+// CurveID is the type of a TLS identifier for an elliptic curve. See
+// https://www.iana.org/assignments/tls-parameters/tls-parameters.xml#tls-parameters-8.
+//
+// In TLS 1.3, this type is called NamedGroup, but at this time this library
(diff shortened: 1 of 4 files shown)


deadprogram:
@scottfeldman looks like you need to run `go fmt`.

```
Unformatted:
  src/syscall/syscall.go
make: *** [GNUmakefile:184: fmt-check] Error 1
```

deadprogram:
Tested alongside https://github.com/tinygo-org/net/pull/17 and worked as expected, so now squash/merging this PR. We also need a PR to update the with`net` submodule.

scottfeldman:
>  We also need a PR to update the with`net` submodule.

Sorry, I think I missed something here...what needs updated?

Should the "dev" branch be removed from .gitmodules?

```
[submodule "src/net"]
        path = src/net
        url = https://github.com/tinygo-org/net.git
        branch = dev
```

Issue #2446, opened 2021-12-30
Multicore support


aykevl:
We need to support multi core systems some day:

 - Some microcontrollers have more than one core that we'd like to use (RP2040, ESP32)
 - Most operating systems have multiple cores.

At the same time, support for multi core shouldn't (significantly) regress single-core performance on microcontrollers, because almost all microcontrollers are single core.

To support multi core, there need to be changes to:

 - [ ] the scheduler, to schedule over more than one core
 - [ ] channels (implemented in the runtime)
 - [ ] the garbage collector (it needs to stop all goroutines to scan their stacks)
 - [ ] wrappers for atomic instructions (see #2307)
 - [ ] packages like sync
 - [ ] ... probably more

I propose we add a new `unicore` (or `singlecore`, `singlethread`, ...) build tag for single core chips and schedulers, so that we can know at build time whether a configuration can assume it's only running in a single (hardware) thread and optimize accordingly.

I suspect it's easiest to not add multi core support first to a chip (RP2040, ESP32) but to an operating system. Operating system threads are already well defined and require the vast majority of changes that are also required for multi core support on a chip (see the list above). I propose we add a `-scheduler=threads` for that, where each goroutine is running in a separate OS thread. This is _much_ simpler to implement than Go runtime style green threads because it doesn't require special handling of blocking system calls.

Nerzal:
Guess we can update the Non-Goals of the readme.md now 
![image](https://user-images.githubusercontent.com/9110370/147753866-359b6393-b978-4004-b457-bf7e3bc91a34.png)

eyz:
Perhaps target a FreeRTOS or similar to start?

*EDIT*: I now see you mentioned you would rather start with an OS, and did not actually imply that it would be on a RTOS.

aykevl:
> Guess we can update the Non-Goals of the readme.md now

Yes, feel free to send a PR :)

> Perhaps target a FreeRTOS or similar to start?

No, that would be far more difficult. We already have a RTOS-like scheduler, adding FreeRTOS or similar would make things quite complicated. And add an external dependency that I'd really like to avoid.

migueleliasweb:
Hi all, any updates on this? 

Although I haven't worked much with embedded systems I would like to lend a hand to add support for this.

couillonnade:
I am curious if any progress has been made for supporting multiple cores? Did not find any hack or PR about it. Should we consider this is not a primary goal for TinyGo? Thank you!

[...]

PR #4047, opened 2023-12-18
Draft: threads scheduler

jclab-joseph:
- [X] amd64/linux createThread
- [ ] amd64/windows createThread
- [ ] 386/linux createThread
- [ ] arm/linux createThread
- [ ] arm64/linux createThread
- [ ] arm64/linux createThread
- [ ] testdata
- [ ] Leave possible pause to the OS.

https://github.com/tinygo-org/tinygo/pull/2290 must be merged first.

This is a draft of a threads scheduler implementation as a cornerstone of a multi-core implementation.

I can't do it alone. I would love to work together.

See https://github.com/tinygo-org/tinygo/issues/2446

### DIFF
--- a/builder/musl.go
+++ b/builder/musl.go
@@ -130,6 +130,7 @@ var Musl = Library{
 			"thread/*.c",
 			"time/*.c",
 			"unistd/*.c",
+			"linux/gettid.c",
 		}
 		if arch == "arm" {
 			// These files need to be added to the start for some reason.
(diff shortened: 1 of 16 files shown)


dgryski:
In addition to the scheduler, the `sync/*` packages probably needs to be rewritten to account for concurrency.

jclab-joseph:
Yes,
Need to test it further, but once #2290 is merged, Mutex, Cond, etc. will likely be thread safe.

aykevl:
Another important piece is the garbage collector. You need to find a way to pause all other threads while the GC is running.

jclab-joseph:
Close by #4559

PR #4052, opened 2023-12-20
machine/serial, rp2040: add support for hardware flow control

deadprogram:
This PR extends the `machine` package serial implementation for the rp2040 to add support for hardware flow control.


bgould:
thank you @deadprogram , merging

PR #4053, opened 2023-12-20
machine/nano-rp2040: add UART1 and correct mappings for NINA via UART.

deadprogram:
This PR adds ad corrects the `machine` package nano-rp2040 implementation to add UART1 and also correct mappings for controlling the NINA chip via UART.


bgould:
thank you @deadprogram , merging

PR #4054, opened 2023-12-22
Add `-serial=rtt` for Segger RTT input/output

aykevl:
This adds Segger RTT support to TinyGo.

Segger RTT is a protocol to set up a fast input/output channel over a debug connection. I've found this useful when working with the PineTime, which has a debug connection but does not have a UART. The same may be true for other boards.

The protocol is not limited to Segger debuggers: it also works with OpenOCD and CMSIS-DAP (in my case, the Particle debugger).

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -777,6 +777,8 @@ endif
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=pca10040 -serial=none examples/echo
 	@$(MD5SUM) test.hex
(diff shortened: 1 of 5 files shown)


bgould:
Looks good to me, merging.

aykevl:
Thank you!

Oh and for completeness: the way you'd use this is by adding `-serial=rtt -monitor` to the `tinygo flash` command, and you'd see the log directly in the console. For example, my full command line for the PineTime looks like this:

    tinygo flash -target=./pinetime-wasp-bootloader.json -programmer=cmsis-dap -opt=2 -size=short -serial=rtt -monitor

(...yes, it's getting kinda long)

PR #4056, opened 2023-12-27
os/user: add bare-bones implementation of the os/user package

eliasnaur:


deadprogram:
Thanks for the addition @eliasnaur now merging.

PR #4068, opened 2024-01-03
all: statically link to LLVM 17 instead of LLVM 16

aykevl:
We can now finally do it, now that Espressif has updated their fork.


deadprogram:
We also need to update the Docker `dev` build. Should I add a commit to this PR for that, or should it be in a different PR?

aykevl:
@deadprogram I wasn't sure about that, but looking more closely yeah it looks like it's doing a build of LLVM (instead of linking to system libraries)? I've updated the PR.

builder/cc1as.cpp
```diff
@@ -1,5 +1,12 @@
 //go:build byollvm
 
+// Source: https://github.com/llvm/llvm-project/blob/main/clang/tools/driver/cc1as_main.cpp
+// This file needs to be updated each LLVM release.
+// There are a few small modifications to make, like:
+//   * ExecuteAssembler is made non-static.
+//   * The struct AssemlberImplementation is moved to cc1as.h so it can be
```
QuLogic:
```suggestion
//   * The struct AssemblerImplementation is moved to cc1as.h so it can be
```

builder/cc1as.h
```diff
@@ -89,6 +92,10 @@ struct AssemblerInvocation {
   /// Whether to emit DWARF unwind info.
   EmitDwarfUnwindType EmitDwarfUnwind;
 
+  // Whether to emit compact-unwind for non-canonical entries.
+  // Note: maybe overriden by other constraints.
```
QuLogic, line 96:
```suggestion
  // Note: maybe overridden by other constraints.
```

deadprogram:
@aykevl this commit needs to be added to this branch for it to be able to build correctly: https://github.com/tinygo-org/tinygo/commit/579b86a55464978918f6efa76437ed2c9f4cd73b

builder/cc1as.cpp
aykevl: Thanks, fixed!

builder/cc1as.h
aykevl: This is code copied from the LLVM project, I'd prefer to keep it as-is (or fix the typo upstream).

aykevl:
@deadprogram I've updated the PR and added your commit.

(To be honest, I'm not exactly sure what it does - I'm not so familiar with Docker).

deadprogram:
@aykevl I still need to correct something in that commit, it did not push the base docker image to ghcr, only to dockerhub.

deadprogram:
OK corrected and now merging. Thanks for working on this @aykevl and to @QuLogic for review/feedback.

PR #4069, opened 2024-01-03
targets: add ninafw tag to nano-rp2040 for ninafw BLE support

deadprogram:
This PR adds the `ninafw` tag to the `nano-rp2040` target for ninafw BLE support in https://github.com/tinygo-org/bluetooth/pull/207


bgould:
LGTM, merging

PR #4075, opened 2024-01-06
Use LLVM 17 by default

aykevl:
This switches the `go install` command on Linux and macOS to use LLVM 17 by default, instead of LLVM 16. These versions have been out a long time, so it's about time we switched the default.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,5 +128,5 @@ workflows:
       - test-llvm15-go118
       # This tests the upcoming Go 1.22 support.
       - test-llvm16-go122
-      # This tests the upcoming LLVM 17 support.
+      # This tests LLVM 17 support when linking against system libraries.
       - test-llvm17-go121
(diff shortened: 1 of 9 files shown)


deadprogram:
@aykevl I think this shell script needs to be moved into the GH action working directory e.g. the root of the repo.

https://github.com/tinygo-org/tinygo/actions/runs/7431727178/job/20222762002?pr=4075#step:4:5

aykevl:
I've fixed the sizediff job by specifying the path relative to the root of the repository (and fixing the permissions).

The sizediff job still fails, but if you look at it that's just because it can't find the script _after checking out the dev branch_. Therefore, it's an expected error we can ignore. After merging this PR, this shouldn't be a problem anymore (perhaps I should have put that change in a separate PR, oh well).

deadprogram:
OK thanks for the update @aykevl now merging!

PR #4076, opened 2024-01-06
machine/pyportal: add needed values to board file for ninafw BLE support

deadprogram:
This PR adds needed values to the `machine` package board file for the Adafruit `PyPortal` for ninafw BLE support.


aykevl (approved):
Cool, I didn't realize the PyPortal had BLE support (though with the ESP32 on board it's kinda obvious).

deadprogram:
Thanks for the review @aykevl now merging.

PR #4081, opened 2024-01-10
cgo: fix calling CGo callback inside generic function

aykevl:
The package of the generic function might not be set, so we have to call Origin() for it.

Found while porting mcufont to TinyGo.


deadprogram:

PR #4082, opened 2024-01-12
runtime, builder: WebAssembly reactor mode

### LINKED ISSUE #2735: WASI reactor mode
Hi all,

I've been experimenting with tinygo having some fun prototyping a go wasm application hosted by wasmtime-java. Unfortunately, I hit a bug/missing feature when trying to access the memory: https://github.com/kawamuray/wasmtime-java/issues/21. As described in that issue, tinygo cannot compile in [reactor mode](https://github.com/WebAssembly/WASI/blob/2ccb606926bee332a7b0a486e150d1053a493977/design/application-abi.md#current-unstable-abi), hence preventing wasmtime-java to access its memory. Although this might be recognized as a limitation of the engine (apparently others, at least from the documentation, can switch the usage mode independently from the exports https://wasmedge.org/book/en/dev/go.html#a-simple-function-run), I think it would be nice if tinygo can support the reactor mode.

I tried to prototype the support myself following what was done for clang https://reviews.llvm.org/D62922 and this is roughly what I needed to come up with a correct export:

* Patch manually [`wasi.json`](https://github.com/tinygo-org/tinygo/blob/release/targets/wasi.json#L11) adding the ldflag `"--entry=_initialize"`
* Rename the export `_start` to `_initialize`: https://github.com/tinygo-org/tinygo/blob/release/src/runtime/runtime_wasm_wasi.go#L15

If you want I can also provide a proper PR, but i need some guidance :)

ydnar:
This is an initial attempt at enabling reactor mode for WebAssembly programs, where the host can repeatedly call into the guest Wasm program via functions exported with `//go:wasmexport`.

The entrypoint for a WebAssembly reactor module is `_initialize` instead of `_start`.

Assuming that the WebAssembly runtime is not reentrant (e.g. not threaded), then this works if the //go:wasmexport calls are detected and wrapped in a function that starts the Go scheduler.

When the exported function ends, all other goroutines are paused.

Goroutines started in a global init() function will run while the host has called into the guest. They are paused when the guest call returns and restarted on the next call.

TODO: figure out how to enable reactor mode:

1. Should it be a flag to tinygo build?
2. Should it be a build tag (e.g. -tags reactor)?
3. Should the compiler detect the omission of main.main and automatically enable reactor mode?

TODO: figure out how and where best to wrap `//go:wasmexport` calls.

TODO: figure out how this works with `-scheduler none`

Fixes #2735.


aykevl:
> TODO: figure out how to enable reactor mode:

My suggestion would be to add a new target, for example `-target=wasi-reactor`.

> TODO: figure out how and where best to wrap `//go:wasmexport` calls.

I would suggest to do that in a separate PR, and make sure to fully support reactor mode first.

PR #4083, opened 2024-01-12
build: use llvm-17 base image correctly for faster docker dev builds

deadprogram:
This PR corrects the GH actions docker builds to use the caches from the separate `llvm` and `docker` builds correctly.

A change to the major version of LLVM needs to be pushed to the `build-llvm-image` branch, which I have now done for `llvm-17`. This PR picks up that change, and also corrects the slowness I introduced myself with some other recent build changes.


dgryski (approved):
I believe you.

deadprogram:
Thanks @dgryski now merging.

PR #4085, opened 2024-01-15
targets: add ninafw pins and settings to additional boards

deadprogram:
This PR adds the `ninafw` pins and settings to the Adafruit Metro M4 Airlift board definition and target.

UPDATE: I have also added the mapping for the PyBadge when using AirLift FeatherWing in a separate commit. Lastly I have added a third commit that adds the `ninafw` tag to a couple of other boards. This will help simplify the maintenance of the `drivers` repo.


bgould:
LGTM

PR #4090, opened 2024-01-17
Go 1.22 support

aykevl:
Work in progress Go 1.22 support. ~Some features haven't been implemented yet, and cgo crashes.~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,6 +105,13 @@ jobs:
       - test-linux:
           llvm: "15"
     resource_class: large
+  test-llvm16-go122:
+    docker:
+      - image: golang:1.22-rc-bullseye
+    steps:
+      - test-linux:
+          llvm: "16"
+    resource_class: large
   test-llvm17-go121:
     docker:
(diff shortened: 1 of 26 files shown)


aykevl:
I think the only thing missing is `reflect.TypeFor`, which is now required by encoding/json (and should be trivial to implement).

aykevl:
This is ready for review.

Two things are still missing:

  * The standard library doesn't have a Go module, which means tooling won't detect a Go version. This needs to be fixed, for the next point.
  * For some reason, golang.org/x/tools/go/ssa doesn't seem to use new-style loop variables (see [this page](https://go.dev/wiki/LoopvarExperiment) for details). I need to investigate why this is.

I might fix those in a separate PR.

deadprogram (approved):
From my point of view, looks good to merge now and let people report any issues.

aykevl:
Thanks! Merged it :shipit:

aykevl:
New Go 1.22 loop variables support: https://github.com/tinygo-org/tinygo/pull/4092

PR #4092, opened 2024-01-19
loader: make sure Go version is plumbed through

aykevl:
This fixes the new loop variable behavior in Go 1.22.

Specifically:
  * The compiler (actually, the x/tools/go/ssa package) now correctly picks up the Go version.
  * If a module doesn't specify the Go version, the current Go version (from the `go` tool and standard library) is used. This fixes `go run`. As a side effect, this also fixes the Go version of the standard library (though for that we should really copy the go.mod file from the standard library).
  * The tests in testdata/ that use a separate directory are now actually run in that directory. This makes it possible to use a go.mod file there.
  * There is a test to make sure old Go modules still work with the old Go behavior, even on a newer Go version.

### DIFF
--- a/compileopts/options.go
+++ b/compileopts/options.go
@@ -23,6 +23,7 @@ type Options struct {
 	GOOS            string // environment variable
 	GOARCH          string // environment variable
 	GOARM           string // environment variable (only used with GOARCH=arm)
+	Directory       string // working dir, leave it unset to use the current working dir
 	Target          string
 	Opt             string
 	GC              string
@@ -48,7 +49,6 @@ type Options struct {
 	Programmer      string
 	OpenOCDCommands []string
 	LLVMFeatures    string
-	Directory       string
 	PrintJSON       bool
 	Monitor         bool
 	BaudRate        int
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -27,6 +27,8 @@ import (
 	"github.com/tinygo-org/tinygo/goenv"
 )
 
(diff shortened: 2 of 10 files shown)


aykevl:
Pushed a fix that should fix the build failure on some older Go versions (like Go 1.18).

aykevl:
@deadprogram looks like something's up with TinyHCI?

deadprogram:
>  looks like something's up with TinyHCI?

Restarted.

deadprogram:

deadprogram:
@aykevl I think this PR is not working as expected. Looks like it is causing an error in the CI build for the `bluetooth` package here: https://github.com/tinygo-org/bluetooth/actions/runs/7586401583/job/20664471703#step:7:8

aykevl:
> @aykevl I think this PR is not working as expected. Looks like it is causing an error in the CI build for the `bluetooth` package here: https://github.com/tinygo-org/bluetooth/actions/runs/7586401583/job/20664471703#step:7:8

Looks like it is working exactly as expected. The bluetooth go.mod file says Go 1.18, while `unsafe.SliceData` was added in Go 1.20. We just didn't see the error before.

The solution would be to fix the Bluetooth package (either to bump the go.mod version, or use something other than `unsafe.SliceData`).

deadprogram:
> We just didn't see the error before.

Aha! That is actually good news, then. I will submit a PR with update to Bluetooth package.

PR #4093, opened 2024-01-19
reflect: update to Go 1.22 semantics

aykevl:
The IsZero function was changed slightly in Go 1.22, so we'll need to update it.

While this could in theory break existing programs that rely on the old behavior, they'd also break with the imminent Go 1.22 release so I don't think this is a real problem.


deadprogram:
> While this could in theory break existing programs that rely on the old behavior, they'd also break with the imminent Go 1.22 release so I don't think this is a real problem.

Very accurate observation, and thanks for the fix @aykevl now merging.

PR #4097, opened 2024-01-23
Initial support for Waveshare RP2040-Zero 

conejoninja:
[Waveshare RP2040-Zero](https://www.waveshare.com/wiki/RP2040-Zero)

I think I'm missing to add some smoke test or something, it's been a while since I last contributed and not sure if I'm missing something.

### DIFF
--- a/src/machine/board_rp2040-zero.go
+++ b/src/machine/board_rp2040-zero.go
@@ -1,0 +1,121 @@
+//go:build rp2040_zero
+
+// This file contains the pin mappings for the Waveshare RP2040 Zero boards.
+//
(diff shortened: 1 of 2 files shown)


deadprogram:
@conejoninja reminder that the smoke tests are located here https://github.com/tinygo-org/tinygo/blob/dev/GNUmakefile#L455-L797 :rabbit:

ysoldak:
We have support for this board already https://github.com/tinygo-org/tinygo/blob/release/src/machine/board_waveshare-rp2040-zero.go

conejoninja:
Thanks @ysoldak , somehow I missed it, closing the PR as it's no longer needed

PR #4101, opened 2024-01-27
cgo: add support for printf

aykevl:
The C printf function is sometimes needed for C files included using CGo. This PR makes sure they're available on all systems where CGo is fully supported (that is, everywhere except on AVR).

For baremetal systems using picolibc, I've picked the integer-only version of printf to save on flash size. In the future we might want to consider providing a way to pick the floating point version instead, if needed.

PR #4102, opened 2024-01-27
esp32c3: prepare for wifi/BLE support

aykevl:
I've been investigating wifi support on the ESP32-C3. It might not be terribly difficult by reusing most of the work done by the Espressif Rust people (which also need a stack separate from ESP-IDF). But it does need some tweaks to the linker script and a bunch of ROM functions.

Source for the new functions: https://github.com/espressif/esp-idf/blob/master/components/esp_rom/esp32c3/ld/esp32c3.rom.ld

Related: #4101 (logging from the binary blobs becomes a lot easier with printf support).


deadprogram:
This is quite exciting! Now merging, thanks @aykevl

PR #4103, opened 2024-01-27
all: make TinyGo code usable with "big Go" CGo

aykevl:
I managed to get CGo sort-of working in VSCode (meaning that it will typecheck code in the IDE itself) using a few crude hacks, but it requires a few minor changes to the TinyGo standard library.

I intend to eventually add this support in the TinyGo extension for VSCode directly, but for now I've manually updated .vscode/settings.json to get IDE support. In any case, it would be nice to have this for when I hopefully add this to the TinyGo extension eventually.

For reference, I'm using TinyGo target "gopher-badge" in VSCode with the following custom settings:

```json
{
    "cmake.configureOnOpen": false,
    "go.toolsEnvVars": {
        "CGO_ENABLED": "1",
        "CC": "clang",
        "GOOS": "linux",
        "GOARCH": "arm",{
        "CGO_CFLAGS": "-Werror -fshort-enums -fomit-frame-pointer -mfloat-abi=soft -fno-exceptions -fno-unwind-tables -fno-asynchronous-unwind-tables -ffunction-sections -fdata-sections -nostdlibinc -isystem /home/ayke/.cache/tinygo/picolibc-thumbv6m-unknown-unknown-eabi-cortex-m0plus/include -isystem /home/ayke/src/tinygo/tinygo/lib/picolibc/newlib/libc/include -isystem /home/ayke/src/tinygo/tinygo/lib/picolibc/newlib/libc/tinystdio -Oz --target=thumbv6m-unknown-unknown-eabi -mcpu=cortex-m0plus -mthumb -fno-PIC",
        "GOROOT": "/home/ayke/.cache/tinygo/goroot-8d8f8a7da5d85cd177e3196b59dcd519d76b00dc6a0aae2635059910f11c4e8c",
        "GOFLAGS": "-tags=cortexm,baremetal,linux,arm,rp2040,rp,gopher_badge,math_big_pure_go,gc.conservative,scheduler.tasks,serial.usb",
        "CCC_OVERRIDE_OPTIONS": "+-mthumb",
    },
}
```

Some notes:

  * I got the list of CFlags by patching the tinygo binary, I should make a separate PR for that too.
  * I had to remove the `tinygo` build tag to get things to work (the `tinygo` files often can't be parsed by upstream CGo).
  * The `CCC_OVERRIDE_OPTIONS` is the especially crude hack: Go adds `-marm` and there doesn't appear to be a way to override it, except using this special (undocumented) environment variable. It's really meant to be used for tests, not for production code.

### DIFF
--- a/src/machine/machine_rp2040_rom.go
+++ b/src/machine/machine_rp2040_rom.go
@@ -1,4 +1,4 @@
-//go:build rp2040
+//go:build tinygo && rp2040
 
 package machine
 
--- a/src/internal/task/task_stack_arm.S
+++ b/src/internal/task/task_stack_arm.S
@@ -1,3 +1,5 @@
+//go:build tinygo
+
 // Only generate .debug_frame, don't generate .eh_frame.
 .cfi_sections .debug_frame
 
--- a/src/internal/task/task_stack_avr.S
+++ b/src/internal/task/task_stack_avr.S
@@ -1,3 +1,5 @@
+//go:build tinygo
+
 .section .bss.tinygo_systemStack
 .global tinygo_systemStack
 .type   tinygo_systemStack, %object
--- a/src/internal/task/task_stack_cortexm.S
+++ b/src/internal/task/task_stack_cortexm.S
@@ -1,3 +1,5 @@
+//go:build tinygo
+
 // Only generate .debug_frame, don't generate .eh_frame.
 .cfi_sections .debug_frame
 
--- a/src/internal/task/task_stack_cortexm.go
+++ b/src/internal/task/task_stack_cortexm.go
@@ -52,17 +52,17 @@ func (s *state) archInit(r *calleeSavedRegs, fn uintptr, args unsafe.Pointer) {
 }
 
 func (s *state) resume() {
-	switchToTask(s.sp)
+	tinygo_switchToTask(s.sp)
 }
 
 //export tinygo_switchToTask
-func switchToTask(uintptr)
+func tinygo_switchToTask(uintptr)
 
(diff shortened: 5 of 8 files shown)


src/machine/machine_rp2040_rom.go
```diff
@@ -1,4 +1,4 @@
-//go:build rp2040
+//go:build tinygo && rp2040
```
ysoldak, line 1: What's special about this single machine .go file in PR?

ysoldak:
Thank you for looking into this. I'm pretty annoyed with VSCode red marks all over when working on something that uses our bluetooth lib.

src/machine/machine_rp2040_rom.go
deadprogram: Good question. @aykevl could you please elaborate?

deadprogram: @aykevl any comment on this before I merge it?

aykevl: It uses CGo, which leads to problems. None of the other files do.

deadprogram:
Thanks for adding this @aykevl and to @ysoldak for review. Now merging.

deadprogram:
Looks like this might have had an unintended side effect @aykevl 

Please see https://github.com/tinygo-org/drivers/actions/runs/7961592170/job/21733165140#step:8:509

aykevl:
@deadprogram pretty sure that's unrelated. I've seen many weird flaky tests lately, I think we have another memory corruption error.

PR #4104, opened 2024-01-28
rp2040: add definition for machine.PinToggle

bgould:
@soypat To your knowledge, is this correct? It seems to be working as expected in this example - https://github.com/tinygo-org/drivers/pull/648/files#diff-a95e9433f1eeac9e6a4babe7393dc048b03c22d172365ff9016d42862c7cc71a - However I'm still reading through the datasheet and stepping through interrupt setup trying to be sure, and your opinion is appreciated.

### DIFF
--- a/src/machine/machine_rp2040_gpio.go
+++ b/src/machine/machine_rp2040_gpio.go
@@ -241,6 +241,8 @@ const (
 	PinFalling PinChange = 4 << iota
 	// Edge rising
 	PinRising
+
+	PinToggle = PinFalling | PinRising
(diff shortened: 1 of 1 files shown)


soypat:
Hey @bgould- From what I've read in the source code it does seem OK based on how the pinchange is stored into register. I might have some more time during the week to read the datasheet in depth (couldn't quite find the page in a quick search).

deadprogram:
Thanks for the addition @bgould and to @soypat for review. Now merging.

PR #4109, opened 2024-02-01
machine/rp2040: set XOSC startup delay multiplier

xudongzheng:
XOSC requires additional time to stablize on certain RP2040 boards.

This is done in Pico SDK: https://github.com/search?q=repo%3Araspberrypi%2Fpico-sdk%20PICO_XOSC_STARTUP_DELAY_MULTIPLIER&type=code

This sets the multiplier globally for simplicity. I don't believe there are any practical drawbacks in doing so. Without it, the compiled firmware will hard fault on boot (on affected boards), which is quite difficult to debug.

### DIFF
--- a/src/machine/machine_rp2040_xosc.go
+++ b/src/machine/machine_rp2040_xosc.go
@@ -8,6 +8,11 @@ import (
 	"unsafe"
 )
 
+// On some boards, the XOSC can take longer than usual to stabilize. On such
+// boards, this is needed to avoid a hard fault on boot/reset. Refer to
+// PICO_XOSC_STARTUP_DELAY_MULTIPLIER in the Pico SDK for additional details.
+const XOSC_STARTUP_DELAY_MULTIPLIER = 64
+
 type xoscType struct {
 	ctrl     volatile.Register32
 	status   volatile.Register32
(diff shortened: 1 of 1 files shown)


deadprogram:
Looking thru docs and posts, seems like a reasonable change to me. Thanks for looking into this @xudongzheng

src/machine/machine_rp2040_xosc.go
```diff
@@ -30,7 +30,8 @@ func (osc *xoscType) init() {
 	osc.ctrl.Set(rp.XOSC_CTRL_FREQ_RANGE_1_15MHZ)
 
 	// Set xosc startup delay
-	delay := (((xoscFreq * MHz) / 1000) + 128) / 256
+	multiplier := 64
+	delay := (((xoscFreq * MHz) / 1000) + 128) / 256 * multiplier
```
ysoldak:
```suggestion
	delay := (xoscFreq * KHz + 128) / 256 * XOSC_STARTUP_DELAY_MULTIPLIER
```
And define `XOSC_STARTUP_DELAY_MULTIPLIER` as const upper in this file with comment along the lines PICO SDK does it "On some samples, the xosc can take longer to stabilize than is usual" or similar.

xudongzheng: Changing `((xoscFreq * MHz) / 1000)` to `xoscFreq * KHz + 128` can be potentially confusing since `xoscFreq` is defined as 12 (MHz) in the various board files.

deadprogram:
Thanks for adding this @xudongzheng and to @ysoldak for review. Now merging.

PR #4113, opened 2024-02-05
ci: add macOS ARM64 builder

aykevl:
This means we can finally release native arm64 builds of TinyGo on macOS!

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -14,7 +14,17 @@ concurrency:
 jobs:
   build-macos:
     name: build-macos
-    runs-on: macos-11
+    strategy:
+      matrix:
+        # macos-12: amd64 (oldest supported version as of 05-02-2024)
+        # macos-14: arm64 (oldest arm64 version)
+        os: [macos-12, macos-14]
+        include:
+          - os: macos-12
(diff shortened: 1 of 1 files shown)


aykevl:
This PR is ready!

deadprogram:
Looking good, this is really great for Mac people. Now merging, thanks @aykevl

edwinjhlee:
Hello friends in tinygo.

I notice arm64 tinygo build is ready. When will the official arm64 binary release?

I am developing x-cmd, and I really want to add tinygo into x-cmd pkg. You can consider x-cmd pkg is like nvm/pyenv/asdf, it help user install software without root priviledge, and manage different version of package for developers.

We already add some language and a lot of binaries into the x-cmd pkg ( a lot of pkg are written in go )

This is an example:
https://www.x-cmd.com/pkg/bat

Usually I would prefer using official build as release source instead of build our own. So I am wondering when tinygo macos arm64 build will be released.

aykevl:
The next release will happen any moment now! Probably today or tomorrow.

edwinjhlee:
I see ~ Thank you. Tiny go is awesome.

https://github.com/tinygo-org/tinygo/releases/tag/v0.31.0

PR #4115, opened 2024-02-05
ci: update from Node.js 16 to Node.js 18

aykevl:
Node.js 16 is no longer supported, so we can drop support for it as well.

This also means updating a whole lot of GitHub Actions versions, because they were updated to work on Node.js 20 instead. For most actions this should be a relatively small change, but the upload-aftifact action has had some major changes (which should generally improve things a lot).


deadprogram:
Necessary upgrades. Now merging, thanks @aykevl

PR #4116, opened 2024-02-05
compileopts: set 'purego' build tag by default

aykevl:
This is needed for various crypto libraries.
It adds the `purego` build tag, and removes it from testdata/corpus.yaml.

Tested by running `make test-corpus` and `make test-corpus-wasi`, both passed on this PR.

CC @FiloSottile who told me yesterday the `purego` build tag is most definitely supported for cryptographic libraries to avoid Go assembly.

### DIFF
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -74,7 +74,12 @@ func (c *Config) GOARM() string {
 
 // BuildTags returns the complete list of build tags used during this build.
 func (c *Config) BuildTags() []string {
-	tags := append(c.Target.BuildTags, []string{"tinygo", "math_big_pure_go", "gc." + c.GC(), "scheduler." + c.Scheduler(), "serial." + c.Serial()}...)
+	tags := append(c.Target.BuildTags, []string{
+		"tinygo",                                     // that's the compiler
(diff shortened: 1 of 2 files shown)


dgryski (approved):
LGTM

FiloSottile:
After this is merged and golang/go#58636 is fixed in Go 1.23, you should be able to remove

https://github.com/tinygo-org/tinygo/blob/731532cd2b6353b60b443343b51296ec0fafae09/compiler/alias.go#L18-L25

and I believe all failures at https://tinygo.org/docs/reference/lang-support/stdlib/#crypto will resolve.

If they don't, please open an issue!

FYI, you might be able to remove the whole stdlibAliases mechanism if golang/go#63270 is fixed, too.

aykevl:
Oh yes if we can remove those aliases that would be great! They were basically just a hack to get those packages compiling, I'd be very happy to remove that code.

(But also see https://github.com/golang/go/issues/58636#issuecomment-1931785836 where I describe some problems with the `purego` build tag).

deadprogram:
Thank you @aykevl for adding this and to @dgryski and @FiloSottile for reviews/feedback.

Now merging.

PR #4117, opened 2024-02-06
build: remove more files from host on Docker build

deadprogram:
This PR modifies the build to remove more files from the GH action host runner for the Docker build to avoid running out of disk space.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -22,11 +22,14 @@ jobs:
       - name: Free Disk space
         shell: bash
         run: |
+          df -h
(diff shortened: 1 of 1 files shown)


deadprogram:
I have verified that this PR does in fact fix the build issue with running out of space:

https://github.com/tinygo-org/tinygo/actions/runs/7805843348/job/21290928603

dgryski (approved):
LGTM.  Anything else we can remove while we're here?

deadprogram:
This is probably more that can be removed, but I will need to revisit that in a future PR.

Thanks for review @dgryski now merging.

PR #4118, opened 2024-02-07
ci: switch to Go 1.22

aykevl:
Let's see how well it runs.

I have removed the Go 1.22 test from .circleci/config.yml because it's not needed anymore: we're testing everything in the normal CI process now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,16 +105,9 @@ jobs:
       - test-linux:
           llvm: "15"
     resource_class: large
-  test-llvm16-go122:
+  test-llvm17-go122:
     docker:
-      - image: golang:1.22-rc-bullseye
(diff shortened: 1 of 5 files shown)


aykevl:
I need to investigate why `node` doesn't work anymore in the new Alpine Linux images.

aykevl:
> I need to investigate why `node` doesn't work anymore in the new Alpine Linux images.

Apparently I just needed to run it again (I guess this was fixed in the meantime). Let's run CI again...

aykevl:
Looks like TinyHCI is stuck? I can do a rebase/force-push to try and get it unstuck if needed.

deadprogram:
I think we are OK on the TinyHCI checks. Now merging, thanks for the update @aykevl

PR #4119, opened 2024-02-07
esp32: switch over to the official SVD file

aykevl:
This change only affects the esp32. The esp8266 is still community-only and other chips (like the esp32c3) were never supported by community SVD files.

I think it's a good idea to use the official SVD files instead of the community maintained ones (which probably won't really be maintained now anymore).

I've checked that examples/blinky1 and examples/serial produce the same binary, so they won't have any change in behavior.


deadprogram:
Much better, thanks @aykevl now merging.

PR #4121, opened 2024-02-09
nrf52840: generic board support

bgould:
Adding minimum definitions need to compile firmware for generic nRF52840 boards not otherwise defined in the `machine` package.  The following sort of target file can be used to take advantage of this feature (the key element being the `nrf52840_generic` build tag:

```json
{
  "inherits": ["nrf52840", "nrf52840-s140v6-uf2"],
  "build-tags": ["adv360pro", "nrf52840_generic"],
  "serial-port": ["1209:9090", "29ea:0362", "239a:00b3"],
  "msd-volume-name": ["ADV360PRO"]
}
```

The pin definitions can be overridden by end users in an `init()` function such as:  https://github.com/bgould/kinadv360pro-firmware/blob/split-refactor/main_generic.go#L1-L13

The USB vendor and product ID are valid as documented at: https://pid.codes/org/TinyGo/

USB details can also be easily overridden by end users, such as: https://github.com/bgould/kinadv360pro-firmware/blob/split-refactor/main.go#L27-L33

### DIFF
--- a/src/machine/board_nrf52840_generic.go
+++ b/src/machine/board_nrf52840_generic.go
@@ -1,0 +1,27 @@
+//go:build nrf52840 && nrf52840_generic
+
+package machine
+
+const HasLowFrequencyCrystal = true
+
+var (
+	LED          = NoPin
+	SDA_PIN      = NoPin
+	SCL_PIN      = NoPin
+	UART_TX_PIN  = NoPin
+	UART_RX_PIN  = NoPin
+	SPI0_SCK_PIN = NoPin
+	SPI0_SDO_PIN = NoPin
+	SPI0_SDI_PIN = NoPin
+
+	// https://pid.codes/org/TinyGo/
+	usb_VID uint16 = 0x1209
+	usb_PID uint16 = 0x9090
+
+	usb_STRING_MANUFACTURER = "TinyGo"
+	usb_STRING_PRODUCT      = "nRF52840 Generic board"
+)
+
+var (
+	DefaultUART = UART0
+)
--- a/src/machine/board_nrf52840_generic.go
+++ b/src/machine/board_nrf52840_generic.go
@@ -2,8 +2,6 @@
 
 package machine
 
-const HasLowFrequencyCrystal = true
-
 var (
 	LED          = NoPin
 	SDA_PIN      = NoPin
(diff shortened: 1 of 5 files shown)


src/machine/board_nrf52840_generic.go
```diff
@@ -0,0 +1,27 @@
+//go:build nrf52840 && nrf52840_generic
+
+package machine
+
+const HasLowFrequencyCrystal = true
```
aykevl: This is not a given, some boards in fact don't have a low frequency crystal (though this is rare).
But maybe it's rare enough that we can leave it as-is.

src/machine/board_nrf52840_generic.go
```diff
@@ -0,0 +1,27 @@
...
+	SCL_PIN      = NoPin
+	UART_TX_PIN  = NoPin
+	UART_RX_PIN  = NoPin
+	SPI0_SCK_PIN = NoPin
+	SPI0_SDO_PIN = NoPin
+	SPI0_SDI_PIN = NoPin
+
+	// https://pid.codes/org/TinyGo/
+	usb_VID uint16 = 0x1209
+	usb_PID uint16 = 0x9090
```
aykevl, line 17: Where did this one come from? Did you allocate this yourself?

aykevl (commented):
Can you add a smoke test to the GNUmakefile file?
Other than that, this looks good to me.

src/machine/board_nrf52840_generic.go
aykevl: Oh I found it: https://github.com/pidcodes/pidcodes.github.com/pull/717
Apparently it was @ardnew who requested it.

bgould: I refactored this a bit so that it is configurable via build flag.

src/machine/board_nrf52840_generic.go
bgould: I refactored this a bit so that it configurable via build flag.

src/machine/board_nrf52840_generic.go
```diff
@@ -0,0 +1,25 @@
...
+	LED          = NoPin
+	SDA_PIN      = NoPin
+	SCL_PIN      = NoPin
+	UART_TX_PIN  = NoPin
+	UART_RX_PIN  = NoPin
+	SPI0_SCK_PIN = NoPin
+	SPI0_SDO_PIN = NoPin
+	SPI0_SDI_PIN = NoPin
+
+	// https://pid.codes/org/TinyGo/
```
deadprogram, line 15: I was today years old when I discovered that we have our own PID. :smile_cat:

deadprogram:
Let's squash/merge this since it appears that all feedback was resolved.

Thanks @bgould and to @aykevl for review.

PR #4125, opened 2024-02-12
runtime, targets: add new target for wasm_unknown

deadprogram:
This PR adds a new target `wasm_unknown` based in part from PR #3072

### DIFF
--- a/src/runtime/runtime_tinygowasm_unknown.go
+++ b/src/runtime/runtime_tinygowasm_unknown.go
@@ -31,10 +31,9 @@ func abort() {
 
 //go:linkname syscall_Exit syscall.Exit
 func syscall_Exit(code int) {
-	//proc_exit(uint32(code))
 }
 
-// TinyGo does not yet support any form of parallelism on WebAssembly, so these
+// There is not yet any support for any form of parallelism on WebAssembly, so these
 // can be left empty.
 
 //go:linkname procPin sync/atomic.runtime_procPin
--- a/src/runtime/runtime_tinygowasm_unknown.go
+++ b/src/runtime/runtime_tinygowasm_unknown.go
@@ -1,0 +1,50 @@
+//go:build wasm_unknown
+
+package runtime
+
+const (
+	stdout = 1
+)
+
+func putchar(c byte) {
+}
+
+func getchar() byte {
+	// dummy, TODO
+	return 0
+}
+
+func buffered() int {
+	// dummy, TODO
+	return 0
+}
+
+//go:linkname now time.now
+func now() (sec int64, nsec int32, mono int64) {
+	return 0, 0, 0
+}
+
+// Abort executes the wasm 'unreachable' instruction.
+func abort() {
+	trap()
+}
+
+//go:linkname syscall_Exit syscall.Exit
+func syscall_Exit(code int) {
+	//proc_exit(uint32(code))
+}
+
+// TinyGo does not yet support any form of parallelism on WebAssembly, so these
+// can be left empty.
+
+//go:linkname procPin sync/atomic.runtime_procPin
+func procPin() {
+}
+
+//go:linkname procUnpin sync/atomic.runtime_procUnpin
+func procUnpin() {
+}
+
+func hardwareRand() (n uint64, ok bool) {
+	return 0, false
+}
--- a/src/runtime/runtime_wasm_unknown.go
+++ b/src/runtime/runtime_wasm_unknown.go
@@ -1,0 +1,49 @@
+//go:build wasm_unknown
+
+package runtime
+
+import (
+	"unsafe"
+)
+
+type timeUnit int64
+
+// libc constructors
+//
+//export __wasm_call_ctors
+func __wasm_call_ctors()
+
+//export _start
+func _start() {
+	// These need to be initialized early so that the heap can be initialized.
+	heapStart = uintptr(unsafe.Pointer(&heapStartSymbol))
+	heapEnd = uintptr(wasm_memory_size(0) * wasmPageSize)
+	run()
+}
+
+// Read the command line arguments from WASI.
+// For example, they can be passed to a program with wasmtime like this:
+//
+//	wasmtime run ./program.wasm arg1 arg2
+func init() {
+	__wasm_call_ctors()
+}
+
+var args []string
+
+func ticksToNanoseconds(ticks timeUnit) int64 {
+	return int64(ticks)
+}
+
+func nanosecondsToTicks(ns int64) timeUnit {
+	return timeUnit(ns)
+}
+
+const timePrecisionNanoseconds = 1000 // TODO: how can we determine the appropriate `precision`?
+
+func sleepTicks(d timeUnit) {
+}
+
+func ticks() timeUnit {
+	return timeUnit(0)
+}
--- a/src/runtime/runtime_wasm_unknown.go
+++ b/src/runtime/runtime_wasm_unknown.go
@@ -2,9 +2,7 @@
 
 package runtime
 
-import (
-	"unsafe"
-)
+import "unsafe"
 
 type timeUnit int64
 
@@ -21,10 +19,6 @@ func _start() {
 	run()
 }
 
-// Read the command line arguments from WASI.
-// For example, they can be passed to a program with wasmtime like this:
-//
-//	wasmtime run ./program.wasm arg1 arg2
 func init() {
 	__wasm_call_ctors()
 }
@@ -39,7 +33,8 @@ func nanosecondsToTicks(ns int64) timeUnit {
 	return timeUnit(ns)
 }
 
-const timePrecisionNanoseconds = 1000 // TODO: how can we determine the appropriate `precision`?
+// with the wasm32-unknown-unknown target there is no way to determine any `precision`
+const timePrecisionNanoseconds = 1000
 
 func sleepTicks(d timeUnit) {
 }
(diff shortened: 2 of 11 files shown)


deadprogram:
This branch in its current state does work when run like this:

```
$ tinygo build -size full -o hello-unknown.wasm -target wasm-unknown ./src/examples/hello/
    466       0      37  130985 |     503  131022 | (unknown)
     22       0       0       4 |      22       4 | main
   3947       0       4      42 |    3951      46 | runtime
      4       0       0       0 |       4       0 | runtime/volatile
   4439       0      41  131031 |    4480  131072 | total
```
given this input program:

```go
package main

var x int32

//go:wasmimport pybadge echo_i32
func echo(x int32)

//go:export update
func update() {
	x++
	echo(x)
}

func main() {
}
```

One question I have, is how can I control the amount of RAM being added here? I would like to allocate a single page by default. The code here produces the following WASM:

```
Memory[1]:                                                                                                                                        
 - memory[0] pages: initial=2
```

deadprogram:
Can we get rid of any of these functions?

```
Function[25]:                                                                                                                                     
 - func[1] sig=2 <__wasm_call_ctors>                                                                                                              
 - func[2] sig=6 <tinygo_getCurrentStackPointer>                                                                                                  
 - func[3] sig=0 <runtime.memequal>                                                                                                               
 - func[4] sig=0 <runtime.hash32>                                                                                                                 
 - func[5] sig=3 <malloc>                                                                                                                         
 - func[6] sig=3 <runtime.alloc>                                                                                                                  
 - func[7] sig=5 <runtime.hashmapBinarySet>                                                                                                       
 - func[8] sig=1 <free>                                                                                                                           
 - func[9] sig=4 <runtime.hashmapBinaryGet>                                                                                                       
 - func[10] sig=1 <runtime.hashmapBinaryDelete>                                                                                                   
 - func[11] sig=4 <calloc>                                                                                                                        
 - func[12] sig=4 <realloc>                                                                                                                       
 - func[13] sig=2 <_start>                                                                                                                        
 - func[14] sig=2 <runtime.calculateHeapAddresses>                                                                                                
 - func[15] sig=5 <runtime.markRoots>                                                                                                             
 - func[16] sig=3 <(runtime.gcBlock).state>                                                                                                       
 - func[17] sig=1 <(runtime.gcBlock).markFree>                                                                                                    
 - func[18] sig=6 <runtime.growHeap>                                                                                                              
 - func[19] sig=1 <runtime.startMark>                                                                                                             
 - func[20] sig=5 <(runtime.gcBlock).setState>                                                                                                    
 - func[21] sig=0 <runtime.hashmapGet>                                                                                                            
 - func[22] sig=4 <runtime.hashmapBucketAddrForHash>                                                                                              
 - func[23] sig=7 <runtime.hashmapSet>                                                                                                            
 - func[24] sig=3 <(runtime.gcBlock).findHead>                                                                                                    
 - func[25] sig=2 <update>
 ```

**Comment by deadprogram** [comment]
Thanks to @dgryski suggestion, this branch can now build the following test program to a miniscule 307 bytes in size:

```go
package main

var x int32

//go:wasmimport pybadge echo_i32
func echo(x int32)

//go:export update
func update() {
	x++
	echo(x)
}

func main() {
}
```

```
$ tinygo build -size short -o hello-unknown.wasm -target wasm-unknown -gc=leaking -no-debug ./src/examples/hello/
     30       4    1024 |      34    1028
$ ls -l hello-unknown.wasm 
-rwxrwxr-x 1 ron ron 307 feb 15 14:37 hello-unknown.wasm
```

The resulting WAT

```wat
(module
  (type (;0;) (func))
  (type (;1;) (func (param i32)))
  (import "pybadge" "echo_i32" (func $main.echo (type 1)))
  (import "env" "memory" (memory (;0;) 2))
  (func $_start (type 0))
  (func $update (type 0)
    (local i32)
    i32.const 1024
    i32.const 1024
    i32.load
    i32.const 1
    i32.add
    local.tee 0
    i32.store
    local.get 0
    call $main.echo)
  (export "_start" (func $_start))
  (export "update" (func $update))
  (data (;0;) (i32.const 1024) "\00\00\00\00"))
```

So small!

Marking ready for review.

src/runtime/runtime_tinygowasm_unknown.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// Abort executes the wasm 'unreachable' instruction.
+func abort() {
+	trap()
+}
+
+//go:linkname syscall_Exit syscall.Exit
+func syscall_Exit(code int) {
+	//proc_exit(uint32(code))
```
orsinium: You could trap for a non-zero exit code. Then it will stop code execution, as a syscall.Exit is expected to do. On the downside, it won't execute defers.

src/runtime/runtime_tinygowasm_unknown.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// TinyGo does not yet support any form of parallelism on WebAssembly, so these
```
orsinium: Wasm doesn't provide any in the first place. There is a threading proposal but it only covers shared memory and atomic operations. Actually spawning the threads is left to the host environment, you can't do that from wasm.

src/runtime/runtime_wasm_unknown.go
```diff
@@ -0,0 +1,47 @@
...
+
+func ticksToNanoseconds(ticks timeUnit) int64 {
+	return int64(ticks)
+}
+
+func nanosecondsToTicks(ns int64) timeUnit {
+	return timeUnit(ns)
+}
+
+const timePrecisionNanoseconds = 1000 // TODO: how can we determine the appropriate `precision`?
```
orsinium: You probably can't. You have clock precision information on wasi but the whole point of `wasm_unknown` target is to not use wasi. And the whole point of the wasm is to not give you any information about the host environment in any other way. 

On a positive side, clock precision doesn't matter if you don't have a clock access anyway.

orsinium (approved):

src/runtime/runtime_wasm_unknown.go
deadprogram: Comment changed to note this.

src/runtime/runtime_tinygowasm_unknown.go
deadprogram: Comment changed. Thanks for the feedback!

src/runtime/runtime_tinygowasm_unknown.go
deadprogram: Removed commented out code.

deadprogram:
Thanks to @hunjixin @dgryski @codefromthecrypt @mathetake @orsinium and everyone else who helped get this to happen!

Now merging.

aykevl:
Can you add this target to builder/builder_test.go? It's a new target with a different set of default LLVM flags and this test makes sure that the LLVM features match with whatever Clang produces (a mismatch there can lead to problems sometimes).

targets/wasm-unknown.json
```diff
@@ -0,0 +1,26 @@
...
+	"features":      "+mutable-globals,+nontrapping-fptoint,+sign-ext",
+	"build-tags":    ["tinygo.wasm", "wasm_unknown"],
+	"goos":          "linux",
+	"goarch":        "arm",
+	"linker":        "wasm-ld",
+	"rtlib":         "compiler-rt",
+	"scheduler":     "none",
+	"default-stack-size": 4096,
+	"cflags": [
+		"-mno-bulk-memory",
```
aykevl, line 13: Note that wasm-libc by default _does_ include bulk memory instructions, which will likely be invoked by the compiler sometimes (for example, when copying a large slice).

dgryski: But this target shouldn't be using wasi-libc anyways since there *is* no `wasi` here.

aykevl: So which libc do you propose we use instead?
We _require_ the use of a libc, because LLVM expects us to have one:
* Copying large amounts of memory is typically lowered to calls to `memmove` or `memcpy`.
* Zeroing memory (for example, when allocating memory) is usually done using a call to `memset`.
* Many math functions will be emitted as libc calls if wasm doesn't have appropriate instructions for them.

The math ones we could perhaps work around if it's really needed (though I'd rather not for consistency), but the memory builtins need to be supplied somehow.

dgryski: Oh, I thought the memory ones were in the compiler runtime lib.  Totally forgot about the math ones though.  I was thinking about the file access bits and spaced on everything else.

aykevl: There has been discussion inside LLVM of adding the math functions to compiler-rt which would in fact fix some subtle bugs in LLVM (the libc versions may set errno while the LLVM intrinsics don't, so if the LLVM intrinsics get lowered to the libc versions they might unexpectedly set errno). But I haven't seen anybody actually put in the work to make it happen - probably because this is only relevant when compiling with `-fno-math-errno` and targeting instruction sets that don't have these math functions as intrinsics in the instruction set. In other words, nobody really cares enough.

Regarding the memory builtins: I haven't seen any discussion on those, I think they're in libc for historic reasons. I suspect that building with bulk memory instructions enabled will avoid those intrinsics - but then the host needs support for those functions.

What we _could_ do is make a new "minimal wasm libc" based on wasi-libc but keeping only the parts we need (like memory and math functions) and excluding any WASI parts. I don't even think this will be very difficult: just compile a subset of the files in wasi-libc in a similar way that libraries like compiler-rt and picolibc are built today (on demand and cached).

PR #3986, opened 2023-11-04
fix crash on type assert on empty interface

dgryski:

PR #4126, opened 2024-02-13
Checking for methodset existance

### LINKED ISSUE #3788: compiler crash on interface assert
I originally noticed this issue while investigating another issue: https://github.com/tinygo-org/tinygo/issues/3771#issuecomment-1585869068

this is a small reproducer:

```
/Volumes/git/tinygo/x (dev)*$ uname -v
Darwin Kernel Version 20.6.0: Thu Mar  9 20:39:26 PST 2023; root:xnu-7195.141.49.700.6~1/RELEASE_X86_64
/Volumes/git/tinygo/x (dev)*$ 
/Volumes/git/tinygo/x (dev)*$ go version
go version go1.20.5 darwin/amd64
/Volumes/git/tinygo/x (dev)*$ 
/Volumes/git/tinygo/x (dev)*$ tinygo version
tinygo version 0.28.1 darwin/amd64 (using go version go1.20.5 and LLVM version 15.0.7)
/Volumes/git/tinygo/x (dev)*$ 
/Volumes/git/tinygo/x (dev)*$ cat main.go 
package main

func main() {
        // _ = any(0).(interface{ x() }) // this is ok
}

func init() {
        _ = any(0).(interface{ x() }) // this crashes the compiler
}
/Volumes/git/tinygo/x (dev)*$ tinygo build .
panic: interp: offset out of range

goroutine 66 [running]:
github.com/tinygo-org/tinygo/interp.pointerValue.addOffset(...)
        /Volumes/git/tinygo/interp/memory.go:523
github.com/tinygo-org/tinygo/interp.(*runner).run(0xc0010c0b40, 0xc000f75db0, {0xc00072f110, 0x1, 0x0?}, 0xc00111ba10, {0xc0007160d0, 0x8})
        /Volumes/git/tinygo/interp/interpreter.go:413 +0x78c8
github.com/tinygo-org/tinygo/interp.(*runner).run(0xc0010c0b40, 0xc000f75d60, {0x0, 0x0, 0x10001ba50?}, 0x0, {0x10041ef36, 0x4})
        /Volumes/git/tinygo/interp/interpreter.go:512 +0x7d45
github.com/tinygo-org/tinygo/interp.RunFunc({0xc0001ea120?}, 0xc001503ce0?, 0x40?)
        /Volumes/git/tinygo/interp/interp.go:238 +0x3a5
github.com/tinygo-org/tinygo/builder.Build.func3(0xc001115560)
        /Volumes/git/tinygo/builder/build.go:437 +0xcf9
github.com/tinygo-org/tinygo/builder.runJob(0xc001115560, 0x0?)
        /Volumes/git/tinygo/builder/jobs.go:222 +0x4f
        /Volumes/git/tinygo/builder/jobs.go:123 +0x5be
/Volumes/git/tinygo/x (dev)*$ 
```

### LINKED ISSUE #4079: type assertion panics
The behavior of type assertions in tinygo differs from the standard go compiler. I get panic when I try to make a valid type assertion.

Example:
https://github.com/gggrafff/tinygo_issue

marco6:
This PR adds a check for the existence of the methodset in the interpreter implementation of typeasserts. This is necessary since this [field is optional](https://github.com/tinygo-org/tinygo/blob/release/compiler/interface.go#L250).

This should be the last bit that fixes #3788 since pull #3986 only covers `XXX.(any)` cases and still crashes if the interface has methods. As such, it should also fix #4079.

As a last note, I also tried a different approach [here](https://github.com/marco6/tinygo/tree/mandatory-methodset) where I forced the methodset to be mandatory as I felt it was more robust, but the simplicity of these 4 lines seems to be unmatched.

### DIFF
--- a/interp/interpreter.go
+++ b/interp/interpreter.go
@@ -427,13 +427,20 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 				if err != nil {
 					return nil, mem, r.errorAt(inst, err)
 				}
+				// typecodePtr always point to the numMethod field in the type
+				// description struct. The methodSet, when present, comes right
+				// before the numMethod field (the compiler doesn't generate
+				// method sets for concrete types without methods).
+				// Considering that the compiler doesn't emit interface type
+				// asserts for interfaces with no methods (as the always succeed)
+				// then if the offset is zero, this assert must always fail.
 				if typecodePtr.offset() == 0 {
 					locals[inst.localIndex] = literalValue{uint8(0)}
 					break
 				}
 				typecodePtrOffset, err := typecodePtr.addOffset(-int64(r.pointerSize))
 				if err != nil {
-					return nil, mem, r.errorAt(inst, err) // unlikely
+					return nil, mem, r.errorAt(inst, err)
 				}
 				methodSetPtr, err := mem.load(typecodePtrOffset, r.pointerSize).asPointer(r)
 				if err != nil {
--- a/interp/interpreter.go
+++ b/interp/interpreter.go
@@ -427,6 +427,10 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 				if err != nil {
 					return nil, mem, r.errorAt(inst, err)
 				}
+				if typecodePtr.offset() == 0 {
+					locals[inst.localIndex] = literalValue{uint8(0)}
+					break
+				}
 				typecodePtrOffset, err := typecodePtr.addOffset(-int64(r.pointerSize))
 				if err != nil {
 					return nil, mem, r.errorAt(inst, err) // unlikely
--- a/testdata/init.go
+++ b/testdata/init.go
@@ -109,3 +109,14 @@ func sliceString(s string, start, end int) string {
 func sliceSlice(s []int, start, end int) []int {
 	return s[start:end]
 }
+
+type outside struct{}
+
+func init() {
+	_, _ = any(0).(interface{ DoesNotExist() })
+	_, _ = any("").(interface{ DoesNotExist() })
+	_, _ = any(outside{}).(interface{ DoesNotExist() })
+
+	type inside struct{}
+	_, _ = any(inside{}).(interface{ DoesNotExist() })
+}


interp/interpreter.go
```diff
@@ -427,6 +427,10 @@ func (r *runner) run(fn *function, params []value, parentMem *memoryView, indent
 				if err != nil {
 					return nil, mem, r.errorAt(inst, err)
 				}
+				if typecodePtr.offset() == 0 {
+					locals[inst.localIndex] = literalValue{uint8(0)}
+					break
+				}
```
aykevl, line 440: Can you please explain in a comment why this is correct? Because it is very non-obvious just reading the source code. It would need to include something about what `.offset() == 0` means (no method set), and why this implies the type assert is always false. Also, it's worth pointing out that `literalValue{uint8(0)}` is in fact setting the type assert value to false.

marco6: I'm not sure commenting on that single piece of code is too useful, unless we comment also the logic after that. Before reading the writer (which is in another file) I couldn't understand any part of that logic. For example:

```go
typecodePtrOffset, err := typecodePtr.addOffset(-int64(r.pointerSize))
```

Why should we go back a pointer to get the typecode?
The reason is: because in [interface.go](https://github.com/tinygo-org/tinygo/blob/release/compiler/interface.go#L449), the typecodePtr always returns a pointer to the `numMethod` field. So, when the first member of typecode is the `methodsSet`, `typecodePtr` will point to the *second* member (offset != 0) and you must go back of a pointer size to get the method set. When it's not present, instead, it will point at the beginning (offset == 0) and the subtraction will fail (crashing the compiler).

So, maybe just a pointer to the method would be enough for a comment? Something along the lines of 

```go
// see compilerContext.getTypeCode in interface.go
```
could be enough?

aykevl: > Before reading the writer (which is in another file) I couldn't understand any part of that logic.

I admit I didn't explain the code well enough in comments when I wrote it. That doesn't mean we should keep the code hard-to-read. It's difficult to write too many comments (unless they're wrong).

Also, I've found that writing comments helps me avoid bugs. I've often found that once I tried to explain my code in a comment, I discovered my thinking was flawed and I had just written buggy code.

> [...] unless we comment also the logic after that.

You are free to add more comments explaining the logic if you want. Personally I often do this for code I've written a while ago and don't fully understand, so the next person doesn't have to reverse engineer the whole thing again.

If you don't know what to write, you could write something like this:

> The offset is 0, which means the concrete type doesn't include a method set (the compiler doesn't generate method sets for concrete types without methods). The compiler doesn't emit interface type asserts for interfaces with no methods (they would always succeed) so this interface type assert must fail.

marco6: Ok, I added a comment. I hope it's not too long/pedantic and it's clear. If not, just tell me and I'll try to reword it.

TheSignPainter98:
Any progress on this? I'm also having this issue

aykevl (approved):
LGTM

PR #4128, opened 2024-02-14
add crypto/subtle

crypto-smoke:
Adds `crypto/subtle` package which is needed for AES stream cipher use.


aykevl:
While I appreciate the help, in this case the correct fix would be to add `-tags=purego` (or wait until the next release, which includes #4116). Whenever reasonably possible, we use the standard library instead of copying or rewriting standard library packages.

PR #4130, opened 2024-02-14
machine: use native WriteRegister/ReadRegister functions on mimxrt1062

aykevl:
This seems better than adding extra WriteRegisterEx/ReadRegisterEx methods that do the same thing but use hardware features instead of implementing it in software.

Note: **I have not actually tested this** except for checking that it compiles on the teensy41.

See: https://github.com/tinygo-org/tinygo/pull/3406/files#r1489950066

### DIFF
--- a/src/machine/i2c-register.go
+++ b/src/machine/i2c-register.go
@@ -1,0 +1,29 @@
+//go:build atmega || nrf || sam || stm32 || fe310 || k210 || rp2040 || (esp32c3 && !m5stamp_c3)
+
+package machine
+
+// This file implements WriteRegister and ReadRegister for chips that do not
+// have direct hardware support for this feature.
+
+// WriteRegister transmits first the register and then the data to the
+// peripheral device.
+//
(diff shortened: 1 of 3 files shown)


kenbell:
This change in drivers removed all use of i2c bus-level ReadRegister / WriteRegister...
https://github.com/tinygo-org/drivers/commit/e6f82fad2e9d3f51051dafb07a225102d6e87b08

The issue was compat with periph.io...  https://github.com/tinygo-org/drivers/issues/559

It seems like we're oscillating a bit on what the right interface for i2c should be?

aykevl:
@kenbell while I agree ReadRegister/WriteRegister aren't great APIs, we do have them now and lots of code depends on it. What this PR does is it reduces the number of ReadRegister/WriteRegister methods from 2 to 1 (or 4 to 2).

I'd be open to removing ReadRegister/WriteRegister entirely - but again, there's a lot of existing code that expects it to be there so it's kinda difficult to do.

PR #4132, opened 2024-02-15
targets/wasm_unknown: use proper defaults in target

deadprogram:
This PR improves `targets/wasm_unknown` by using proper defaults for GC to avoid always needing to add the extra flag which is now required for this target.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -773,7 +773,7 @@ endif
 ifneq ($(WASM), 0)
 	$(TINYGO) build -size short -o wasm.wasm -target=wasm               examples/wasm/export
 	$(TINYGO) build -size short -o wasm.wasm -target=wasm               examples/wasm/main
-	$(TINYGO) build -size short -o wasm.wasm -target=wasm-unknown -gc=leaking -no-debug examples/hello-wasm-unknown
+	$(TINYGO) build -size short -o wasm.wasm -target=wasm-unknown       examples/hello-wasm-unknown
 endif
 	# test various compiler flags
 	$(TINYGO) build -size short -o test.hex -target=pca10040 -gc=none -scheduler=none examples/blinky1
--- a/builder/builder_test.go
+++ b/builder/builder_test.go
@@ -35,6 +35,7 @@ func TestClangAttributes(t *testing.T) {
 		"riscv-qemu",
 		"wasi",
 		"wasm",
+		"wasm-unknown",
 	}
 	if hasBuiltinTools {
 		// hasBuiltinTools is set when TinyGo is statically linked with LLVM,
--- a/src/runtime/runtime_wasm_unknown.go
+++ b/src/runtime/runtime_wasm_unknown.go
@@ -16,7 +16,7 @@ func _start() {
 	// These need to be initialized early so that the heap can be initialized.
 	heapStart = uintptr(unsafe.Pointer(&heapStartSymbol))
 	heapEnd = uintptr(wasm_memory_size(0) * wasmPageSize)
-	run()
+	initAll()
 }
 
 func init() {
--- a/src/runtime/runtime_wasm_unknown.go
+++ b/src/runtime/runtime_wasm_unknown.go
@@ -11,8 +11,8 @@ type timeUnit int64
 //export __wasm_call_ctors
 func __wasm_call_ctors()
 
-//export _start
-func _start() {
+//export _initialize
+func _initialize() {
 	// These need to be initialized early so that the heap can be initialized.
 	heapStart = uintptr(unsafe.Pointer(&heapStartSymbol))
 	heapEnd = uintptr(wasm_memory_size(0) * wasmPageSize)
--- a/targets/wasm-unknown.json
+++ b/targets/wasm-unknown.json
@@ -1,13 +1,14 @@
 {
 	"llvm-target":   "wasm32-unknown-unknown",
 	"cpu":           "generic",
-	"features":      "+mutable-globals,+nontrapping-fptoint,+sign-ext",
+	"features":      "+mutable-globals,+nontrapping-fptoint,+sign-ext,-bulk-memory",
 	"build-tags":    ["tinygo.wasm", "wasm_unknown"],
 	"goos":          "linux",
 	"goarch":        "arm",
 	"linker":        "wasm-ld",
 	"rtlib":         "compiler-rt",
 	"scheduler":     "none",
+	"gc":            "leaking",
 	"default-stack-size": 4096,
 	"cflags": [
 		"-mno-bulk-memory",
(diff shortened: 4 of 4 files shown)


orsinium (approved):

aykevl:
I believe what you're looking for is reactor mode: https://github.com/WebAssembly/WASI/blob/main/legacy/application-abi.md. Basically, it means exporting an `_initialize` function (which might set up the heap if used and calls `initAll`) instead of `_start`.

Also see #2735 .

orsinium:
`heapStart` and `heapEnd` initialization sounds useful 👀 I'd test it on some big code examples before merging.

deadprogram:
> Basically, it means exporting an _initialize function (which might set up the heap if used and calls initAll) instead of _start.

OK updated.

orsinium (dismissed):
Both `_start` and `_initialize` exported functions are part of wasi p1. They might not be called in non-wasi or wasi p2 environments. Most likely, any initialization in `wasi-unknown` should be done using the core spec [start section](https://webassembly.github.io/spec/core/syntax/modules.html#start-function). The spec has a note that the start function is called before any exports (including memory) and so shouldn't call any imported functions (and this is the reason why wasi and [the linking convention](https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md#start-section) introduced `_start` in the first place) but that's not an issue for what you're doing: by default, `wasm-unknown` target does not call any exported functions. And even better, in [the component model](https://github.com/WebAssembly/component-model), which supersedes both wasi p1 and the linking convention, [there is no shared memory](https://github.com/WebAssembly/component-model/blob/main/design/high-level/Choices.md), so you'll be able to pass things into external functions before exports are initialized.

deadprogram:
@orsinium I am not understand which changes you are asking for in this PR?

deadprogram:
I just added another commit with the test mentioned by @aykevl here https://github.com/tinygo-org/tinygo/pull/4125#issuecomment-1946813715

aykevl:
So the question is now: how should packages be initialized? Because all those global initializers and `init` calls need to be called somewhere.

> The spec has a note that the start function is called before any exports (including memory) and so shouldn't call any imported functions (and this is the reason why wasi and [the linking convention](https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md#start-section) introduced `_start` in the first place) but that's not an issue for what you're doing: by default, `wasm-unknown` target does not call any exported functions.

I don't agree here. Go `init` is free to call anything, including functions imported using `//go:wasmimport`. Therefore, I'd argue we have the same restriction as WASI and need a separate exported function like `_initialize`.
And while we're not explicitly targetting WASI here, using `_initialize` makes a lot of sense to me for consistency.

deadprogram:
Related question: how can we add something to the .S file to generate a WASM `start` section?

https://webassembly.github.io/spec/core/binary/modules.html#start-section

Seems like we should be able to do this in similar fashion to https://github.com/tinygo-org/tinygo/blob/dev/src/runtime/asm_tinygowasm.S

deadprogram:
> And while we're not explicitly targetting WASI here, using _initialize makes a lot of sense to me for consistency.

Renamed the function to `_initialize()`.

deadprogram:
<s>Renamed function back to `_start` because I could not get it to work as expected. That will be for a future PR.</s>

Never mind. Changed back to `_initialize()`

deadprogram:
Now merging, thank you @orsinium @aykevl for reviews/feedback.

I will still attempt to get the start section enabled, but that will have to wait for a future PR.

aykevl:
> I will still attempt to get the start section enabled, but that will have to wait for a future PR.

How do you intend to use the start function? (I doubt it can used for package initializers).
Also see: https://lists.llvm.org/pipermail/llvm-dev/2019-October/135981.html

deadprogram:
It is for future use. See https://github.com/WebAssembly/WASI/issues/487#issuecomment-1186012592

orsinium:
@aykevl it is a very good point. There are two things that we currently do in the `_start` function in the wasm-wasi target, if I understand it correctly:

1. Heap initialization and other TinyGo internal preparations. I believe the best place for them is the start section. This way, it is guaranteed that they are always executed before any other Go code in the module. It also guarantees that it is executed only once. This use case was exactly the reason why the start section was added into wasm MVP.
2. Running the Go `init` function. You're right, there are certain limitations if we put it into the start section too, and I'm not 100% sure if we should do that or not. If the module calls from init an external function which expects the module's memory to be initialized, it won't work. There won't be a need for that in the component-model-powered environment, but the point of wasm-unknown is to not assume any environment.

So, I'm sure that the best place for TinyGo internals like heap initialization is the start section. However, I'm not that sure about the Go `init` function. Probably, it should also go there, considering that the component model gets a lot of traction (and you;ll have to support it anyway for wasi p2), but then the users should be aware of the limitations (and perhaps use `main` instead when they need to share memory).

aykevl:
Right, with the component model coming up it makes sense to try and use the start section - at least for runtime initialization and maybe for running init functions.

> but then the users should be aware of the limitations (and perhaps use `main` instead when they need to share memory).

You mean users should do initialization from the main function instead of inside the init functions?
Note that this also applies to _dependencies_ which may be difficult or impossible to change. Perhaps dependencies won't call out to external functions, but it's a bit of a risk.

orsinium:
> You mean users should do initialization from the main function instead of inside the init functions?

Precisely! Instead of this:

```
var someGlobal = externalFunc()
```

Or this:

```
var someGlobal int

func init() {
  someGlobal = externalFunc()
}
```

Do this:

```
var someGlobal int

func main() {
  someGlobal = externalFunc()
}
```

> Perhaps dependencies won't call out to external functions

Correct, there shouldn't be a scenario when a dependency calls an external function in wasm-unknown. If a dependency tries to use a wasi API (like FS access), it won't work with wasm-unknown regardless of if it is called from init or main, because there is no wasi in wasm-unknown by design. And if the dependency is designed specifically to work with TinyGo's wasm-unknown (like my [wasm4go](https://github.com/orsinium-labs/wasm4go) project), it will be already designed with the limitations in mind.

aykevl:
> Correct, there shouldn't be a scenario when a dependency calls an external function in wasm-unknown.

I wouldn't be so sure about that. If compiler development has taught me anything, it's that if something is possible, people _will_ eventually do it. For example, people might want to put helper functions for their special WASM environment in a separate package and maybe do some package-level initialization.

orsinium:
If it is designed for wasm-unknown, it will be tested and designed with whatever limitations are already present in the target. If it is designed for wasm-wasi and calls external functions from init, it's quite likely to not work anyway because these functions are most likely to be wasi functions.

Also, none of that would matter in wasip2 and component model because the component model allows start functions to work with linear memory and pass through linear memory external call results:

> start functions reuse the same Canonical ABI machinery as normal imports and exports for getting component-level values into and out of core linear memory.
>
> https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#-start-definitions

Issue #4123, opened 2024-02-10
time.NewTimer() fails from init()


xudongzheng:
The following code results in a fault:

```
package main

import "time"

var timer = time.NewTimer(time.Second * 5)

func main() {
	println("Hello world!")
}
```

Issue also happens if `time.NewTimer()` is called from `init()`.

The following code works correctly

```
package main

import "time"

var timer *time.Timer

func main() {
	timer = time.NewTimer(time.Second * 5)
	println("Hello world!")
}
```

This was tested on both desktop `tinygo build -o ~/hello ~/hello.go` and RP2040 Pico.

It looks like `tn.timer.f` here is `nil`: https://github.com/tinygo-org/tinygo/blob/f4395147039fc356277399853821cb772ebda883/src/runtime/time.go#L39

aykevl:
Thanks, here is a fix: https://github.com/tinygo-org/tinygo/pull/4134

xudongzheng:
Thanks! Closing as I've confirmed the fix.

PR #4134, opened 2024-02-19
interp: do not register runtime timers during interp

aykevl:
The runtime hasn't been initialized yet so this leads to problems.

This fixes https://github.com/tinygo-org/tinygo/issues/4123.


deadprogram:

PR #4135, opened 2024-02-19
compileopts: remove workaround for LLVM 16

aykevl:
This workaround is no longer needed now that we've switched to LLVM 17 where this bug has been fixed upstream: https://github.com/espressif/llvm-project/pull/84
See: https://github.com/tinygo-org/tinygo/issues/3974 and https://github.com/tinygo-org/tinygo/pull/3975.


deadprogram:
Thanks for the cleanup @aykevl now merging.

Issue #3450, opened 2023-02-15
tinygo 0.27 cannot compile to GOOS=darwin


gtsteffaniak:
Anytime I try to add `GOOS=darwin` during compile-time, I get the following error:

```
GOOS=darwin tinygo build .
# runtime
/usr/local/lib/tinygo/src/runtime/scheduler.go:22:24: undefined: GOOS
/usr/local/lib/tinygo/src/runtime/runtime_unix.go:181:17: undefined: clock_MONOTONIC_RAW
/usr/local/lib/tinygo/src/runtime/runtime_unix.go:191:16: undefined: clock_REALTIME
/usr/local/lib/tinygo/src/runtime/runtime_unix.go:225:34: undefined: flag_PROT_READ
/usr/local/lib/tinygo/src/runtime/runtime_unix.go:225:49: undefined: flag_PROT_WRITE
/usr/local/lib/tinygo/src/runtime/runtime_unix.go:225:66: undefined: flag_MAP_PRIVATE
/usr/local/lib/tinygo/src/runtime/runtime_unix.go:225:83: undefined: flag_MAP_ANONYMOUS
```

- Using tinygo_0.27.0_arm64.deb on linux docker image (host machine is m1 macbook).
- Also doesn't work when using amd64 linux image.
- WORKS with GOOS=linux and GOOS=windows

aykevl:
I can't reproduce this in a Debian LXC container (on arm64).
I suspect something went wrong with your installation. Can you share the output of `which tinygo` and `tinygo env`?

dgryski:
I develop entirely on Darwin and I also don't have this problem. Definitely something with the install.

gtsteffaniak:
I am using golang:1.20 image from dockerhub:

```
root@3ad92dcfe559:/app/tinyVisualizer# cat /etc/*release
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

root@3ad92dcfe559:/app/tinyVisualizer# which tinygo
/usr/local/bin/tinygo

root@3ad92dcfe559:/app/tinyVisualizer# tinygo env
GOOS="linux"
GOARCH="arm64"
[...]

PR #4136, opened 2024-02-19
compileopts: always enable CGo

aykevl:
CGo is needed for the rp2040 and for macOS (GOOS=darwin), without it these targets just won't work. And there really isn't a benefit from disabling CGo: we don't need any external linkers for example.

This avoids a somewhat common issue of people having CGO_ENABLED=0 somewhere in their environment and not understanding why things don't work. See for example: https://github.com/tinygo-org/tinygo/issues/3450

### DIFF
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -87,12 +87,6 @@ func (c *Config) BuildTags() []string {
 	return tags
 }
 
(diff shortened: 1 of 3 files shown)


deadprogram:

PR #4140, opened 2024-02-20
Dockerfile: reduce size of resulting image

aykevl:
This reduces the size of the Docker image from 13GB to 1.71GB, and should therefore make CI of the drivers package much faster. It should hopefully also avoid the out-of-space problem we currently have when building the smoke tests for the drivers repo.

This size reduction is done by using multistage builds and only copying the necessary files in the final stage.

---

I don't know whether this is the correct fix though: while it should help with the drivers CI, it would remove the option of using this Docker container as a base container for developing TinyGo itself. (But is anybody actually doing that, and do we need to build a new container for that in CI?)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,26 +20,24 @@ FROM tinygo-llvm AS tinygo-llvm-build
 RUN cd /tinygo/ && \
     make llvm-build
 
-# tinygo-compiler stage builds the compiler itself
-FROM tinygo-llvm-build AS tinygo-compiler
+# tinygo-compiler-build stage builds the compiler itself
+FROM tinygo-llvm-build AS tinygo-compiler-build
 
 COPY . /tinygo
 
-# update submodules
+# build the compiler and tools
 RUN cd /tinygo/ && \
-    rm -rf ./lib/*/ && \
-    git submodule sync && \
-    git submodule update --init --recursive --force
-
-RUN cd /tinygo/ && \
-    make
-
-# tinygo-tools stage installs the needed dependencies to compile TinyGo programs for all platforms.
-FROM tinygo-compiler AS tinygo-tools
-
-RUN cd /tinygo/ && \
-    make wasi-libc binaryen && \
+    git submodule update --init --recursive && \
     make gen-device -j4 && \
-    cp build/* $GOPATH/bin/
+    make build/release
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -20,26 +20,24 @@ FROM tinygo-llvm AS tinygo-llvm-build
...
-    make
+    git submodule update --init --recursive && \
+    make gen-device -j4 && \
+    make build/release
 
-# tinygo-tools stage installs the needed dependencies to compile TinyGo programs for all platforms.
-FROM tinygo-compiler AS tinygo-tools
+# tinygo-compiler copies the compiler build over to a base Go container (without
+# all the build tools etc).
+FROM golang:1.21 AS tinygo-compiler
```
deadprogram, line 36: Shouldn't this be using 1.22 now?

deadprogram:
> I don't know whether this is the correct fix though: while it should help with the drivers CI, it would remove the option of using this Docker container as a base container for developing TinyGo itself. (But is anybody actually doing that, and do we need to build a new container for that in CI?)

This is the only reason to go to all this trouble. If in fact no one is using this for working on tinygo itself, then we can save a lot of hassle/storage/computation.

Dockerfile
deadprogram: Just noticed PR #4118 does that. Never mind.

aykevl: Yeah these two PRs conflict with each other.

aykevl:
So there are essentially _two_ kinds of images we need, which are currently both served by the tinygo-dev image:

1. People wanting to work on TinyGo without modifying their environment.
2. Running tests in CI (like the drivers smoke tests).

1) needs all the necessary tools and LLVM and such, but in my opinion doesn't need to be run on every commit. Instead, we could provide a simple way to set this up on various systems.
2) only needs the release binary, essentially.

I would suggest splitting these in two separate containers, so that the drivers repo can run CI tests quickly while still making all tools available to developers. What do you think?

deadprogram:
> I would suggest splitting these in two separate containers, so that the drivers repo can run CI tests quickly while still making all tools available to developers. What do you think?

Very good idea. We could start with the CI container, and then if anyone asks for it work on the development container later.

And since this PR does exactly that, I will merge it now.

aykevl:
Sounds good to me!

PR #4143, opened 2024-02-20
targets: add support for Thumby

deadprogram:
This PR adds support for the TinyCircuits Thumby 

https://thumby.us/

### DIFF
--- a/src/machine/board_thumby.go
+++ b/src/machine/board_thumby.go
@@ -1,0 +1,94 @@
+//go:build thumby
+
+// This contains the pin mappings for the Thumby.
+//
+// https://thumby.us/
+package machine
+
+import (
+	"device/rp"
+	"runtime/interrupt"
+)
+
+const (
+	THUMBY_SCK_PIN = I2C1_SDA_PIN
+	THUMBY_SDA_PIN = I2C1_SCL_PIN
+
+	THUMBY_CS_PIN    = GPIO16
+	THUMBY_DC_PIN    = GPIO17
+	THUMBY_RESET_PIN = GPIO20
+
+	THUMBY_LINK_TX_PIN = UART0_TX_PIN
+	THUMBY_LINK_RX_PIN = UART0_RX_PIN
+	THUMBY_LINK_PU_PIN = GPIO2
+
+	THUMBY_BTN_LDPAD_PIN = GPIO3
+	THUMBY_BTN_RDPAD_PIN = GPIO5
+	THUMBY_BTN_UDPAD_PIN = GPIO4
+	THUMBY_BTN_DDPAD_PIN = GPIO6
+	THUMBY_BTN_B_PIN     = GPIO24
+	THUMBY_BTN_A_PIN     = GPIO27
+
+	THUMBY_AUDIO_PIN = GPIO28
+
+	THUMBY_SCREEN_RESET_PIN = GPIO20
+)
+
+// I2C pins
+const (
+	I2C0_SDA_PIN Pin = NoPin
+	I2C0_SCL_PIN Pin = NoPin
+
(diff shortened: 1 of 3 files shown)


deadprogram:
PXL_20240220_231119706

src/machine/board_thumby.go
```diff
@@ -0,0 +1,94 @@
...
+// UART on the Thumby
+var (
+	UART0  = &_UART0
+	_UART0 = UART{
+		Buffer: NewRingBuffer(),
+		Bus:    rp.UART0,
+	}
+)
+
+var DefaultUART = UART0
```
rminnich, line 90: This one is totally idiomatic :-)

src/machine/board_thumby.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+var (
+	usb_VID uint16 = 0x2E8A
+	usb_PID uint16 = 0x0005
+)
+
+// UART pins
+const (
+	UART0_TX_PIN = GPIO0
```
rminnich, line 75: An idiomatic name might be UART0TXPin and UART0RXPin, etc.

src/machine/board_thumby.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+// Onboard crystal oscillator frequency, in MHz.
+const (
+	xoscFreq = 12 // MHz
+)
+
+// USB CDC identifiers
+const (
+	usb_STRING_PRODUCT      = "Thumby"
```
rminnich, line 64: so, when I was at Google, _ in names was strongly discourated. As, in the sense, rejected. I notice these all over tinygo but maybe it's time to become more idiomatic. Otherwise it starts to look like "C in Go" -- that was the argument made to me.

also, convention has it that usb should USB, as it is an acronym. 

USBProductName and USBManufacturerName might be ok, but why not just ProductName and or Product and Manufacturer? 

I understand there are conventions you have, so maybe you don't want to change at this point, but these identifier styles would be deemed non-idiomatic ...

If you don't want to export, then you could say _USBProduct or something? 

Same for VID and PID.

src/machine/board_thumby.go
```diff
@@ -0,0 +1,94 @@
...
+// https://thumby.us/
+package machine
+
+import (
+	"device/rp"
+	"runtime/interrupt"
+)
+
+const (
+	THUMBY_SCK_PIN = I2C1_SDA_PIN
```
rminnich, line 14: I wonder, to make the name idiomatic, you might say
ThumbySCKPin

src/machine/board_thumby.go
deadprogram: The names that we have chosen in the past have been generally trying to match datasheets just to have better searchability for non-Go people.

If we want to make a chance to be more in line with Go norms, which we probably want to do, we should most likely have a bigger conversation on how to make that happen.

deadprogram:
Other than the bigger question of pin naming, any other feedback on this PR before I merge?

src/machine/board_thumby.go
soypat: @deadprogram Agreed- I'd strongly suggest against making identifier name changes/decisions until we have (ideally) an automated way of "correcting" these names or a reference document on how TinyGo names identifiers- as in a tool that can be run from the command line to avoid code churn and much back and forth on opinions.

deadprogram: > a tool that can be run from the command line to avoid code churn and much back and forth on opinions.

Oh yes please!

kenbell:
lgtm

deadprogram:
Thanks for review @rminnich @soypat @kenbell now merging.

PR #4147, opened 2024-02-22
Dockerfile: don't remove cached files

aykevl:
This reverts 649f49e00080e2317617e19c76450eeb2af491f7. It is not necessary anymore: the container is an intermediary container so doesn't need to be extra small.

Not sure whether this is really needed but it seems a little cleaner to me.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,12 +2,7 @@
 FROM golang:1.22 AS tinygo-llvm
 
 RUN apt-get update && \
-    apt-get install -y apt-utils make cmake clang-15 ninja-build && \
-    rm -rf \
-        /var/lib/apt/lists/* \
-        /var/log/* \
-        /var/tmp/* \
-        /tmp/*
+    apt-get install -y apt-utils make cmake clang-15 ninja-build
 
 COPY ./GNUmakefile /tinygo/GNUmakefile
 


b0ch3nski:
@aykevl The final stage is still on Go 1.21: `FROM golang:1.21 AS tinygo-compiler` - was this on purpose or by mistake? :)

If it was by mistake, here is my suggestion:
```
FROM golang:1.22 AS golang-base
FROM golang-base AS tinygo-llvm
<...>
FROM golang-base AS tinygo-compiler
```
Something like this could prevent such version skew in future.

aykevl:
@b0ch3nski that was a mistake (see https://github.com/tinygo-org/tinygo/pull/4158#issuecomment-1966432185).
It was essentially a merge failure. I don't think it's very useful to add more complexity to the Dockerfile just for this purpose.

deadprogram:
I expect we still want to remove these files in the interest of reducing the overall size.

aykevl:
> I expect we still want to remove these files in the interest of reducing the overall size.

What do you mean? They don't affect the resulting docker container, only the intermediary container.

deadprogram:
> They don't affect the resulting docker container, only the intermediary container.

That is correct. However it was the LLVM build step that needed the extra space in order to build correctly, which is why we need to clear space for that.

aykevl:
> However it was the LLVM build step that needed the extra space in order to build correctly

Ah, I see. Thanks for clarifying!

PR #4150, opened 2024-02-25
Stub CallSlice for Value

akspi:
Trying to stub [this](https://github.com/golang/go/blob/08d9397170e5870b72a95233e8e4ec43d2d70e30/src/reflect/value.go#L381) method to get tinygo to compile a dependency to wasm. Should be a fairly harmless change?


aykevl (approved):
Yes LGTM

PR #4154, opened 2024-02-26
ci: fix binaryen build

aykevl:
Disable the googletest dependency so that we can avoid that submodule dependency.


deadprogram:

PR #4155, opened 2024-02-26
Don't add `--recursive` to `git submodule update --init`

aykevl:
It's not generally needed. It was added in https://github.com/tinygo-org/tinygo/pull/3958 to fix an issue with binaryen that has since been fixed in #4154, so we don't need the googletest dependency anymore.

Draft until the release is done. #4154 is the hotfix, this cleans up some things after the hotfix and is more risky.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -85,7 +85,7 @@ Now that we have a working static build, it's time to make a release tarball:
 
 If you did not clone the repository with the `--recursive` option, you will get errors until you initialize the project submodules:
 
-    git submodule update --init --recursive
+    git submodule update --init
 
 The release tarball is stored in build/release.tar.gz, and can be extracted with
 the following command (for example in ~/lib):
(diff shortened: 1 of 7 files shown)


aykevl:
Ready for review.

BUILDING.md
```diff
@@ -85,7 +85,7 @@ Now that we have a working static build, it's time to make a release tarball:
 
 If you did not clone the repository with the `--recursive` option, you will get errors until you initialize the project submodules:
```
deadprogram, line 86: Should this comment also be modified?

deadprogram (approved):
The tests pass, so it must be working.

BUILDING.md
aykevl: No seems fine by me.
You can clone the repository with the `--recursive` option, but it will be a bit slower than needed.
If you didn't, you can fix it using `git submodule update --init`.

aykevl (commented):
> The tests pass, so it must be working.

That's a bold statement!

deadprogram:
Now merging, thanks @aykevl

PR #4157, opened 2024-02-26
runtime: add Frame.Entry field

ydnar:
This enables compatibility with golang.org/x/tools/internal/pkgbits.

https://github.com/golang/tools/blob/master/internal/pkgbits/frames_go17.go


aykevl:
Thank you!

PR #4159, opened 2024-02-27
Fix 'out of range' panic in interp

aykevl:
See: https://github.com/tinygo-org/drivers/issues/658


interp/memory.go
```diff
@@ -517,12 +517,12 @@ func (v pointerValue) offset() uint32 {
 // addOffset essentially does a GEP operation (pointer arithmetic): it adds the
 // offset to the pointer. It also checks that the offset doesn't overflow the
 // maximum offset size (which is 4GB).
-func (v pointerValue) addOffset(offset int64) pointerValue {
+func (v pointerValue) addOffset(offset int64) (pointerValue, error) {
```
deadprogram, line 520: This is a lot better than the panic.

deadprogram:
Now merging, thanks for the quick fix @aykevl

PR #4169, opened 2024-03-02
compiler: handle missing LLVM math intrinsics on wasm targets

ydnar:
This enables WebAssembly targets (unknown, wasip2) to use package math without linking libc.

The missing intrinsics are:

```
llvm.exp.f64
llvm.exp2.f64
llvm.log.f64
```

This is required for and cherry-picked from the WASI Preview 2 fork we’re working on with @dgryski.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -830,7 +830,7 @@ func (c *compilerContext) createPackage(irbuilder llvm.Builder, pkg *ssa.Package
 			}
 			// Create the function definition.
 			b := newBuilder(c, irbuilder, member)
-			if _, ok := mathToLLVMMapping[member.RelString(nil)]; ok {
+			if intrinsic := b.mathToLLVMMapping(member.RelString(nil)); intrinsic != "" {
 				// The body of this function (if there is one) is ignored and
 				// replaced with a LLVM intrinsic call.
 				b.defineMathOp()
--- a/compiler/intrinsics.go
+++ b/compiler/intrinsics.go
@@ -111,14 +111,32 @@ func (b *builder) createKeepAliveImpl() {
 	b.CreateRetVoid()
 }
 
-var mathToLLVMMapping = map[string]string{
+// mathToLLVMMapping maps a relative Go symbol to an LLVM intrinsic
+func (b *builder) mathToLLVMMapping(rel string) string {
+	i := mathToLLVM[rel]
+	if strings.HasPrefix(b.Triple, "wasm") && wasmMissingIntrinsics[i] {
+		return ""
+	}
+	return i
+}
+
+var mathToLLVM = map[string]string{
+	"math.Abs":   "llvm.fabs.f64",
 	"math.Ceil":  "llvm.ceil.f64",
-	"math.Exp":   "llvm.exp.f64",
-	"math.Exp2":  "llvm.exp2.f64",
+	"math.Exp":   "llvm.exp.f64",  // requires libc
+	"math.Exp2":  "llvm.exp2.f64", // requires libc
 	"math.Floor": "llvm.floor.f64",
-	"math.Log":   "llvm.log.f64",
+	"math.Log":   "llvm.log.f64", // requires libc
 	"math.Sqrt":  "llvm.sqrt.f64",
 	"math.Trunc": "llvm.trunc.f64",
+}
+
+// wasmMissingIntrinsics maps intrinsics not present in LLVM wasm
+// which require linking to libc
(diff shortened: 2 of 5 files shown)


ydnar:
It could be interesting to add //export annotations to certain tests in testdata so they can be run under wasm32-unknown.

aykevl:
While one possible solution, I'd actually much prefer to use a very stripped down libc anyway (for consistency). This libc might only implement these math and perhaps memory builtins (memset, memcpy, memmove).

LLVM is built in a way that require a libc. If you go without, it might work for a bit but eventually you'll run into issues.

If you really don't want a libc, can you explain why?

ydnar:
@dgryski and I are working on WASI Preview 2 support for TinyGo (GOOS=wasip2). To our knowledge, this is the first native language target of Preview 2 that does not require an adapter shim.

wasi-libc does not yet support WASI Preview 2, and there is not yet a clear timeline when it will.

Second, there are reasons we may want to avoid using it for Go due to how the Component Model interacts with the runtime and GC. 

Given that big Go has non-asm implementations of all math functions, and LLVM has intrinsics for most, it seems reasonable to avoid linking libc (or even libm) for two symbols.

Happy to discuss this in detail with you and @dgryski.

aykevl:
> wasi-libc does not yet support WASI Preview 2, and there is not yet a clear timeline when it will.

That wouldn't matter for math builtins. They don't need any WASI support (v2 or otherwise) - they're just pure functions.

> Second, there are reasons we may want to avoid using it for Go due to how the Component Model interacts with the runtime and GC.
> 
> Given that big Go has non-asm implementations of all math functions, and LLVM has intrinsics for most, it seems reasonable to avoid linking libc (or even libm) for two symbols.

You'll also lose out on many optimization opportunities. Many math builtins are supported directly in wasm and the only way to use those is by using LLVM builtins. Moreover, LLVM knows these builtins and for example converts the equivalent of `float32(sqrt64(n))` to `sqrt32(float32(n))` (pseudocode).

To be clear: I'm only talking about the math and memory builtins. I understand that most other libc features will need to be reimplemented in some other way. There's no reason to put all these functions together in a single library called "libc".

I can make a PR for this if you're interested, to show what I mean.

ydnar:
Huh. I'm confused then. This PR doesn't inhibit LLVM intrinsics, only skipping those that aren't present in Wasm.

Maybe it's worth benchmarking the libc exp/exp2 against Go's under wasm?

Edit: as for memory intrinsics, does LLVM lean on libc for those on wasm? It works in our wasip2 branch with no memcpy impl or libc.

dgryski:
We could also add implementations that call into native wasm instructions to our mini wasip2 libc.

aykevl:
> We could also add implementations that call into native wasm instructions to our mini wasip2 libc.

That's not necessary. LLVM will do that for you. We need separate implementions for the ones that aren't supported by wasm, such as `log` and `exp`.

(I'm working on a math-only version of wasi-libc but of course I run into weird toolchain issues I need to sort out - EDIT: actually that toolchain issue is an existing bug that might also affect compiler-rt for wasm).

aykevl:
> Edit: as for memory intrinsics, does LLVM lean on libc for those on wasm? It works in our wasip2 branch with no memcpy impl or libc.

Yes. But if you have enabled bulk memory operations, LLVM will use those instead (because they'll be faster than what you can implement in a libc). So it's only needed for environments where bulk memory operations aren't available.

ydnar:
Benchmarks:

**Before vs after (this PR):**

```
benchmark             old ns/op     new ns/op     delta
BenchmarkMathAbs      3.52          3.91          +11.26%
BenchmarkMathCeil     3.52          3.97          +12.83%
BenchmarkMathExp      3.75          13.0          +247.56%
BenchmarkMathExp2     3.53          14.5          +310.03%
BenchmarkMathLog      3.75          4.50          +19.98%
```

**Before vs after (with pure Go `exp`, `exp2` and `log`:**

```
benchmark             old ns/op     new ns/op     delta
BenchmarkMathAbs      3.52          3.53          +0.43%
BenchmarkMathCeil     3.52          3.62          +2.64%
BenchmarkMathExp      3.75          13.5          +259.03%
BenchmarkMathExp2     3.53          15.1          +326.75%
BenchmarkMathLog      3.75          4.52          +20.46%
```

ydnar:
Will close this in favor of #4176 when it’s ready.

dgryski:
https://github.com/tinygo-org/tinygo/pull/4176 has been merged.  This can be closed.

PR #4172, opened 2024-03-03
cgo: do a basic test that math functions work

aykevl:
They should, but we weren't testing this.
I discovered this while working on https://github.com/tinygo-org/macos-minimal-sdk/pull/4 which will likely make math.h not work anymore (it will need a single-line fix in TinyGo). So I wanted to make sure we have a test in place before we update that dependency.

### DIFF
--- a/testdata/cgo/main.c
+++ b/testdata/cgo/main.c
@@ -1,3 +1,4 @@
+#include <math.h>
 #include "main.h"
 
 int global = 3;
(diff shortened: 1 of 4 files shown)


deadprogram:
@aykevl seems some problem with macOS for this PR.

aykevl:
I think I've worked around the failure for now.
The issue was that the `puts` output was moved to the end. Not sure how or why, probably something to do with stdout vs stderr or maybe it needed a fflush or something. But I've moved the line, so the test should pass.

deadprogram:
Thanks for adding this @aykevl should help ensure this stays working.

PR #4180, opened 2024-03-09
compileopts: apply OpenOCD commands after target configuration

eliasnaur:
The openocd-commands option cannot refer to commands defined by the target configuration. Lift this restriction by moving the commands to after target loading.

The other two users of 'openocd-commands' are:

```text
$ rg openocd-commands
targets/bluepill-clone.json
3:	"openocd-commands": ["set CPUTAPID 0x2ba01477"]

targets/esp32c3.json
20: 	"openocd-commands": ["gdb_memory_map disable"],
```

I don't know whether this change affects them, nor users of the `-ocd-commands` command line option. Let me know if you'd rather have some other way of achieving the goal.

### DIFF
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -475,9 +475,6 @@ func (c *Config) OpenOCDConfiguration() (args []string, err error) {
 		return nil, fmt.Errorf("unknown OpenOCD transport: %#v", c.Target.OpenOCDTransport)
 	}
 	args = []string{"-f", "interface/" + openocdInterface + ".cfg"}
-	for _, cmd := range c.Target.OpenOCDCommands {
-		args = append(args, "-c", cmd)
-	}
 	if c.Target.OpenOCDTransport != "" {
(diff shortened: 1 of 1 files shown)


aykevl:
Is it possible to execute arbitrary code from OpenOCD commands? So far I've made sure all these OpenOCD commands can only use a safe subset.
So far I've tried to use a safe subset, but maybe that's not really necessary after all.

eliasnaur:
I'm not sure what you're asking. Does this PR introduce (additional) risk of arbitrary command execution? FWIW, my use case is https://github.com/tinygo-org/tinygo/pull/4181.

deadprogram (approved):
This PR would not introduce any code execution that is not already permitted by OpenOCD already. So it seems reasonable to me.

sago35 (approved):
I also think there's no problem with this change.

deadprogram:
Now merging, thanks for the improvement @eliasnaur and to @sago35 for review.

Issue #4114, opened 2024-02-05
WASM code that uses `__multi3` produces module that cannot be parsed by `wasm-opt`


flavio:
This is a follow up on issue https://github.com/tinygo-org/tinygo/issues/3501. The issue has been fixed with tinygo 0.29.0.
However, the [code linked as reproducer](https://github.com/flavio/tinygo-wasi-symbol-bug) is still broken. This time the build fails during the optimization phase, when running `wasm-opt`.

The compilation fails with the following error:
```console
[parse exception: invalid function index (at 0:33795)]
```

This happens when building with the following Tinygo docker images:

* `tinygo/tinygo:0.29.0`
* `tinygo/tinygo:0.30.0`

Both images ship with `wasm-opt` version 114. I've tried to use the latest version of `wasm-opt` (v116), but I got the same error.

**UPDATE**: this happens also with the latest version of tinygo (`tinygo/tinygo:0.31.0`), which contains `wasm-opt` version 116

## Reproducer

I've updated the code inside of the original reproducer, it can be found inside of [this GitHub repository](https://github.com/flavio/tinygo-wasi-symbol-bug).

flavio:
This is happening also with version 0.31.0. I've updated the reproduced to reflect that

aykevl:
I found that #3844 was actually incorrect. I have fixed this locally, I'll make a PR soon.

aykevl:
Here is the PR that fixes this issue: https://github.com/tinygo-org/tinygo/pull/4188.
I can't confirm the output binary is correct, but at least it compiles.

flavio:
@aykevl good news, I've checked out the PR you linked above, built tinygo from scratch and then I successfully built and tested the real code (not the small reproducer). The fix is working :partying_face:

aykevl:
@flavio thanks for testing!

PR #4176, opened 2024-03-04
wasm-unknown: add math and memory builtins that LLVM needs

aykevl:
WIP, not working yet due to existing bugs. But this is the general idea.

PR #4188, opened 2024-03-12
wasm: fix symbol table index for archives

aykevl:
The symbol table was generated incorrectly. The correct way is to use the custom linking WebAssembly section, which I implemented in go-wasm for this purpose (see https://github.com/aykevl/go-wasm/commit/50275154210c1816c2a4ef2fdd33b78e412f08a2).

This fixes https://github.com/tinygo-org/tinygo/issues/4114 and is a prerequisite for https://github.com/tinygo-org/tinygo/pull/4176.


aykevl:
Made a small fix so that it works correctly with #4176.

deadprogram (approved):

PR #4189, opened 2024-03-13
compileopts: adapt GOARM to follow new 1.22 scheme

### LINKED ISSUE #4177: Add support for new ARM softfloat/hardfloat distinction added in Go 1.22
When trying to compile a binary with the new ARM subtargets, e.g. `GOARM=6,softfloat`, tinygo bails out. The support for ARM added in #2260 would need to be extended for that.

MDr164:
This ads the optional softfloat and hardfloat suffix to GOARM the same way it was introduced upstream with Go 1.22. By default armv5 is softfloat while v6 and v7 are hardfloat. Resolves #4177 

TODO: Actually figure out the correct combination of LLVM arch flags.


compileopts/target.go
```diff
@@ -181,15 +181,28 @@ func LoadTarget(options *Options) (*TargetSpec, error) {
...
-			case "5":
-				llvmarch = "armv5"
-			case "6":
-				llvmarch = "armv6"
-			case "7":
-				llvmarch = "armv7"
-			default:
-				return nil, fmt.Errorf("invalid GOARM=%s, must be 5, 6, or 7", options.GOARM)
+			armOpts := strings.Split(options.GOARM, ",")
+			if len(armOpts) == 1 || len(armOpts) == 2 {
```
aykevl: Can you change this condition to return an early error instead?
See: https://go.dev/wiki/CodeReviewComments#indent-error-flow

Something like this:

```go
if len(armOpts) > 2 {
    // return error here
}
// rest of the code
```

compileopts/target.go
```diff
@@ -181,15 +181,28 @@ func LoadTarget(options *Options) (*TargetSpec, error) {
...
+				}
+				if len(armOpts) == 2 {
+					switch armOpts[1] {
+					case "softfloat":
+						llvmarch += "sf"
+					case "hardfloat":
+						llvmarch += "hf"
+					default:
+						return nil, fmt.Errorf("invalid extension GOARM=%s, must be softfloat or hardfloat", options.GOARM)
+					}
```
aykevl: I don't think this is correct, I think you need to end the triple with `-gnueabi` or `-gnueabihf` instead (see the definition of `target` a few lines below - GitHub doesn't let me comment there).

aykevl (changes requested):
You need to add a new test case to builder/builder_test.go, there's a list of GOOS/GOARCH/GOARM combinations that get tested. Once you do that, you can test these flags using `make test`. This will tell you the exact LLVM features you need to use.
This currently tests all of TinyGo and takes a while, you can make a modification similar to https://github.com/tinygo-org/tinygo/pull/4190 to only test the builder package.

compileopts/target.go
MDr164: Sure, I can change that around

aykevl:
It might be useful to try out combinations of flags using Godbolt: https://godbolt.org/z/9cGfzhG6x

compileopts/target.go
MDr164: Isn't that part being substituted further down the line anyways? Afaik the armvX part gets turned back into just arm as LLVM doesn't explicitly check the arm32 version based on the first element of the target triple.

aykevl: I've found that whatever you put after `armv7` or `armv6` appears to be ignored entirely.
What I mean is that this piece of code needs updating to have the correct suffix:

```go
		// Target triples (which actually have four components, but are called
		// triples for historical reasons) have the form:
		//   arch-vendor-os-environment
		target := llvmarch + "-" + llvmvendor + "-" + llvmos
		if options.GOOS == "windows" {
			target += "-gnu"
		} else if options.GOARCH == "arm" {
			target += "-gnueabihf"
		}
```

Here you'll need to modify the environment part (`-gnueabihf`) to something else to disable floating point instructions (which will automatically result in a softfloat fallback).

You may find this page useful to figure out the correct flags: https://embeddedartistry.com/blog/2017/10/11/demystifying-arm-floating-point-compiler-options/

MDr164: I updated the PR to change that part of the target triple now but also change the flags further down. I'm actually not too sure still what the correct subset would be that still supports the majority of chips and still gives decent performance. It looks like most of the FPU stuff is optional anyways so what would be a sane subset when someone explicitly requests hardfloat support for a given arm version?

MDr164:
Looks like CI is failing for code I didn't even touch in this PR. Otherwise this looks fine now. Will have to run a few more tests to be certain all the flags and targets check out on my hardware here but I think it's ready to go.

compileopts/target.go
```diff
@@ -308,13 +322,32 @@ func defaultTarget(goos, goarch, triple string) (*TargetSpec, error) {
...
 		spec.CFlags = append(spec.CFlags, "-fno-unwind-tables", "-fno-asynchronous-unwind-tables")
+		hardfloat := strings.HasSuffix(triple, "hf")
 		switch strings.Split(triple, "-")[0] {
 		case "armv5":
-			spec.Features = "+armv5t,+strict-align,-aes,-bf16,-d32,-dotprod,-fp-armv8,-fp-armv8d16,-fp-armv8d16sp,-fp-armv8sp,-fp16,-fp16fml,-fp64,-fpregs,-fullfp16,-mve.fp,-neon,-sha2,-thumb-mode,-vfp2,-vfp2sp,-vfp3,-vfp3d16,-vfp3d16sp,-vfp3sp,-vfp4,-vfp4d16,-vfp4d16sp,-vfp4sp"
+			if hardfloat {
+				spec.CFlags = append(spec.CFlags, "-mfloat-abi=hard")
+				spec.Features = "+armv5t,+strict-align,-aes,-bf16,-d32,-dotprod,-fp-armv8,-fp-armv8d16,-fp-armv8d16sp,-fp-armv8sp,-fp16,-fp16fml,-fp64,-fpregs,-fullfp16,-mve.fp,-neon,-sha2,-thumb-mode,-vfp2,-vfp2sp,-vfp3,-vfp3d16,-vfp3d16sp,-vfp3sp,-vfp4,-vfp4d16,-vfp4d16sp,-vfp4sp"
+			} else {
+				spec.CFlags = append(spec.CFlags, "-mfloat-abi=soft") // maybe use softfp instead?
```
aykevl: `softfp` means to still use the FPU (but use an ABI compatible with soft-float code), which is not what we want here. So this comment can be removed.

```suggestion
				spec.CFlags = append(spec.CFlags, "-mfloat-abi=soft")
```

Same goes for the other two similar comments below.

aykevl (commented):
If I read this PR correctly, it will now default to softfloat everywhere except when hardfloat is specified explicitly? Is that true? That seems like a bad idea. A better default would be to use softfloat on armv5 and hardfloat on armv6 and armv7 (or softfloat on armv5+armv6 and hardfloat on armv7).

Also, I don't think this affects musl, so you will likely still have some floating point instructions (or ABI incompatibilities). Try running this on an ARM system:

```
GOARM=5,softfloat tinygo run ./testdata/math.go
GOARM=5,hardfloat tinygo run ./testdata/math.go
GOARM=7,softfloat tinygo run ./testdata/math.go
GOARM=7,hardfloat tinygo run ./testdata/math.go
```

They should all produce identical output.

MDr164:
> If I read this PR correctly, it will now default to softfloat everywhere except when hardfloat is specified explicitly? Is that true? That seems like a bad idea. A better default would be to use softfloat on armv5 and hardfloat on armv6 and armv7 (or softfloat on armv5+armv6 and hardfloat on armv7).
> 
> Also, I don't think this affects musl, so you will likely still have some floating point instructions (or ABI incompatibilities). Try running this on an ARM system:
> 
> ```
> GOARM=5,softfloat tinygo run ./testdata/math.go
> GOARM=5,hardfloat tinygo run ./testdata/math.go
> GOARM=7,softfloat tinygo run ./testdata/math.go
> GOARM=7,hardfloat tinygo run ./testdata/math.go
> ```
> 
> They should all produce identical output.

I'll have to run some more tests to make sure this works as expected on musl as well. And yes the idea was that v5 defaults to soft while v6 and v7 default to hard to match upstream behavior. I'll rework this if the musl tests turn out to be what you expect.

MDr164:
Closed in favor of #4374

PR #4190, opened 2024-03-13
Makefile: allow overriding the packages to test in `make test`

aykevl:
This way you can for example run `make test GOTESTPKGS=./builder` to only test the builder package.
I've often done this by manually modifying the Makefile, so having a make parameter available would make this much easier.


deadprogram:
Good idea @aykevl now merging.

PR #4020, opened 2023-12-02
board: Adafruit LED Glasses Driver nRF52840

bgould:
Adafruit LED Glasses Driver - nRF52840 Sensor Board
https://www.adafruit.com/product/5217

References:
https://cdn-learn.adafruit.com/assets/assets/000/105/134/original/adafruit_products_EyeLightsD_sch.png?1633457726
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/variants/ledglasses_nrf52840/variant.h
https://github.com/adafruit/Adafruit-EyeLights-LED-Glasses-and-Driver-PCB/blob/main/Adafruit%20EyeLights%20LED%20Glasses%20Driver.pdf

- [x] create board definition
- [x] verify USB serial
- [x] verify ws2812
- [x] verify button
- [x] verify i2c
- [x] verify LIS3DH
- [x] verify bluetooth
- [x] verify QSPI/flash
- [x] verify battery
- [ ] verify microphone

PR #4197, opened 2024-03-18
feather-nrf52840-sense doesn't use LFXO

goderiich:
Submitting a PR based on discussion with @aykevl here: https://github.com/tinygo-org/bluetooth/issues/251

In the adafruit github comparing feather_nrf52840_express to feather_nrf52840_sense there does seem to be a difference between the two boards, this issue was also experienced with the ledglasses-nrf52840 board https://github.com/tinygo-org/tinygo/pull/4020#issuecomment-1837556982

Express
```
#define USE_LFXO      // Board uses 32khz crystal for LF
// define USE_LFRC    // Board uses RC for LF
```

Sense
```
#define USE_LFRC    // Board uses RC for LF
```

### DIFF
--- a/src/machine/board_feather-nrf52840-sense.go
+++ b/src/machine/board_feather-nrf52840-sense.go
@@ -2,7 +2,7 @@
 
 package machine
 
-const HasLowFrequencyCrystal = true
+const HasLowFrequencyCrystal = false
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Thank you for the correction @goderiich now merging.

PR #4203, opened 2024-03-21
Add pca10059-s140v7 as a target

Schwusch:
This aims to fix https://github.com/tinygo-org/bluetooth/issues/253


Schwusch:
Not sure if anything needs to be added or why the one test is failing. All guidance is welcome

aykevl (approved):

aykevl:
Actually we may need a smoke test for this, just to be sure.
Can you add one to GNUmakefile?

deadprogram:
Thank you for adding this @Schwusch and to @aykevl for review. Now merging.

PR #4205, opened 2024-03-21
crypto/ed25519 

leongross:
This PR should fix the compile error for crypto/ed25519. Changing the path for the aliasing resolved the issue on the website [1].
Using an example program to verify the functionality yields success for me [2]. 

Note:
Contrary to the standard implementation for `ed25519.GenerateKey` the call in `tinygo` always generates a different key [3].
Maybe we should look into this issue.

[1]: https://tinygo.org/docs/reference/lang-support/stdlib/#cryptoed25519
[2]: https://asecuritysite.com/golang/ed25519
[3]: https://pkg.go.dev/crypto/ed25519#GenerateKey

### DIFF
--- a/compiler/alias.go
+++ b/compiler/alias.go
@@ -16,8 +16,8 @@ import "tinygo.org/x/go-llvm"
 
 var stdlibAliases = map[string]string{
 	// crypto packages
-	"crypto/ed25519/internal/edwards25519/field.feMul":    "crypto/ed25519/internal/edwards25519/field.feMulGeneric",
-	"crypto/ed25519/internal/edwards25519/field.feSquare": "crypto/ed25519/internal/edwards25519/field.feSquareGeneric",
+	"crypto/ed25519/internal/edwards25519/field.feMul": "crypto/ed25519/internal/edwards25519/field.feMulGeneric",
+	"crypto/internal/edwards25519/field.feSquare":      "crypto/ed25519/internal/edwards25519/field.feSquareGeneric",
 	"crypto/md5.block":         "crypto/md5.blockGeneric",
 	"crypto/sha1.block":        "crypto/sha1.blockGeneric",
 	"crypto/sha1.blockAMD64":   "crypto/sha1.blockGeneric",


deadprogram:
Hello @leongross thank you for the PR. Looks like you need to run `go fmt` on your changes, however:

```
Unformatted:
  compiler/alias.go
make: *** [GNUmakefile:184: fmt-check] Error 1
```

Could you please do that?

leongross:
@deadprogram regarding the second issue: from your experience, do you have an idea why the key generation is non-deterministic although it should be deterministic?

leongross:
CI run `test-llvm17-go122` failed due to missing `node` dependency:
> /bin/sh: 1: node: not found

leongross:
As I see it resolving this import issue could resolve the following includes:


|req|pkg|
|---|---|
|crypto/ed25519 | crypto/x509| 
|crypto/ed25519| crypto/x509 -> crypto/tls|
|crypto/tls | net/http/httptrace|
|crypto/tls | net/smtp|
|crypto/tls, net/http/httptrace | net/http| 
|crypto/tls, net/http | net/http/cgi|
|net/http | net/http/cookiejar|
|net/http | net/rpc|
|net/http | net/http/pprof| 
|net/http, net/http/cgi | net/http/fcgi|
|crypto/tls, crypto/x509, net/http | net/http/httptest |

aykevl:
> Contrary to the standard implementation for `ed25519.GenerateKey` the call in `tinygo` always generates a different key [3](https://pkg.go.dev/crypto/ed25519#GenerateKey).

What do you mean? Note that "deterministic" means it will generate the same key if the input is the same, meaning `rand` does not provide a random source. If you set the `rand` parameter to nil, then it must generate a different key pair each time (if it didn't, that would be a massive security bug).
If this behavior is different between standard Go and TinyGo, that would be a bug.

leongross:
> > Contrary to the standard implementation for `ed25519.GenerateKey` the call in `tinygo` always generates a different key [3](https://pkg.go.dev/crypto/ed25519#GenerateKey).
> 
> What do you mean? Note that "deterministic" means it will generate the same key if the input is the same, meaning `rand` does not provide a random source. If you set the `rand` parameter to nil, then it must generate a different key pair each time (if it didn't, that would be a massive security bug). If this behavior is different between standard Go and TinyGo, that would be a bug.

I must have misread the documentation, sorry for the confusion.
I assumed that the behavior would be similar to the one in C, using `srand(NULL)` and `rand`, which it clearly is not.

leongross:
I'm not quite sure what the problem with the smoke test is, has someone an idea?

leongross:
can someone re-run the ci?

deadprogram:
CI is now passing, thanks for the fix @leongross now merging.

leongross:
@deadprogram Regarding the list on the [Packages supported by TinyGo](https://tinygo.org/docs/reference/lang-support/stdlib/) on the website: How is the data that the table represents raised? Is this done manually, are there any tests importing the packages, or how is this dealt with?
Anyway, I would be glad to assist you with that as well.

EDIT: I found the [table generation](https://github.com/tinygo-org/tinygo-site/blob/596e3ba7a682df0fdafc5927d1d3603e07e9a4f5/imports/main.go), I will have a look at it.

PR #4208, opened 2024-03-26
os/link

leongross:
Add `os.Link` functionality to create linux `hardlinks`

### DIFF
--- a/src/os/os_hardlink_test.go
+++ b/src/os/os_hardlink_test.go
@@ -1,0 +1,53 @@
+//go:build !windows && !baremetal && !js && !wasi && !wasip1 && !wasm_unknown
+
+// Copyright 2024 The Go Authors. All rights reserved.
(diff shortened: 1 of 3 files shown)


deadprogram:
@rminnich any comment here before merge? :smile_cat:

src/os/os_hardlink_test.go
```diff
@@ -0,0 +1,55 @@
+//go:build !windows && !baremetal && !js && !wasi && !wasip1
```
deadprogram: Do we also need to have `!wasm_unknown` in this list?

leongross: this sounds like a reasonable thing to do

deadprogram:
Thanks for adding this @leongross now merging.

leongross:
Thank you for taking care of that!

PR #4211, opened 2024-03-28
nix: fix wasi-libc include headers

aykevl:
Apparently Nix really doesn't like --sysroot, so we have to use `-nostdlibinc` and `-isystem` instead.

~I've also added some tests to try to detect common issues like these.~ these tests would actually slow down the Nix test a lot with little benefit (the main tests are already done for other OSes) so I've removed them again.

This issue was reported to me by someone on Telegram.


deadprogram:
Thanks for the quick fix for nix.

Now merging.

PR #4213, opened 2024-03-28
os/Chown

leongross:
Add `os.Chown()`

### DIFF
--- a/src/os/file_anyos.go
+++ b/src/os/file_anyos.go
@@ -151,9 +151,6 @@ func Chmod(name string, mode FileMode) error {
 // If the file is a symbolic link, it changes the uid and gid of the link's target.
 // A uid or gid of -1 means to not change that value.
 // If there is an error, it will be of type *PathError.
-//
-// On Windows or Plan 9, Chown always returns the syscall.EWINDOWS or
-// EPLAN9 error, wrapped in *PathError.
 func Chown(name string, uid, gid int) error {
 	e := ignoringEINTR(func() error {
 		return syscall.Chown(name, uid, gid)
--- a/src/os/file_anyos.go
+++ b/src/os/file_anyos.go
@@ -147,6 +147,23 @@ func Chmod(name string, mode FileMode) error {
 	return nil
 }
 
+// Chown changes the numeric uid and gid of the named file.
+// If the file is a symbolic link, it changes the uid and gid of the link's target.
+// A uid or gid of -1 means to not change that value.
+// If there is an error, it will be of type *PathError.
+//
+// On Windows or Plan 9, Chown always returns the syscall.EWINDOWS or
+// EPLAN9 error, wrapped in *PathError.
+func Chown(name string, uid, gid int) error {
+	e := ignoringEINTR(func() error {
+		return syscall.Chown(name, uid, gid)
+	})
+	if e != nil {
+		return &PathError{Op: "chown", Path: name, Err: e}
+	}
+	return nil
+}
+
 // ignoringEINTR makes a function call and repeats it if it returns an
 // EINTR error. This appears to be required even though we install all
 // signal handlers with SA_RESTART: see #22838, #38033, #38836, #40846.
--- a/src/os/os_chmod_test.go
+++ b/src/os/os_chmod_test.go
@@ -33,75 +33,41 @@ func TestChmod(t *testing.T) {
 
 }
 
-func TestChown(t *testing.T) {
-	var (
-		TEST_UID_1           = 1001
-		TEST_GID_1           = 127
-		TEST_UID_2           = 0
-		TEST_GID_2           = 0
-		ERR_PERM_DENIED      = "permission denied"
-		ERR_OP_NOT_PERMITTED = "operation not permitted"
-	)
-
+// Since testing syscalls requires a static, predictable environment that has to be controlled
+// by the CI, we don't test for success but for failures and verify that the error messages are as expected.
+// EACCES is returned when the user does not have the required permissions to change the ownership of the file
+// ENOENT is returned when the file does not exist
+// ENOTDIR is returned when the file is not a directory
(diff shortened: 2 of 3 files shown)


src/os/os_chmod_test.go
```diff
@@ -28,4 +30,26 @@ func TestChmod(t *testing.T) {
...
 	checkMode(t, f.Name(), fm)
+
+}
+
+func TestChown(t *testing.T) {
+	f := newFile("TestChown", t)
+	defer Remove(f.Name())
+	defer f.Close()
+
+	if runtime.GOOS != "windows" {
```
rminnich: what about plan 9?

src/os/os_chmod_test.go
```diff
@@ -28,4 +30,26 @@ func TestChmod(t *testing.T) {
...
+		if err := Chown(f.Name(), 0, 0); err != nil {
+			t.Fatalf("chown %s 0 0: %s", f.Name(), err)
```
rminnich: So in terms of errors like this, and maybe this project is different, but the pattern is either
Chown(%s, 0, 0): %v != nil
or 
Chown(%s, 0, 0): got %v, want nil
It's just so you have a very regular pattern for when you need to deal with logs.
This was pretty much de rigeur at google.

src/os/os_chmod_test.go
```diff
@@ -28,4 +30,26 @@ func TestChmod(t *testing.T) {
...
+		}
+
+		fi, err := Stat(f.Name())
+		if err != nil {
+			t.Fatalf("stat %s: %s", f.Name(), err)
+		}
+
+		if fi.Sys().(*syscall.Stat_t).Uid != 0 || fi.Sys().(*syscall.Stat_t).Gid != 0 {
```
rminnich: I've had tests explode with complex casts like this. A more common pattern woudl be
if fi.Sys() == nil {
error
}
s, ok := fi.Sys().(*syscall.Stat_t)
if  ! ok {
bad
}
uid, gid = s.Uid, s.Gid // just useful so you're not repeating fi.Sys().(*syscall.Stat_t). everywhere

src/os/os_chmod_test.go
```diff
@@ -28,4 +30,26 @@ func TestChmod(t *testing.T) {
 		t.Fatalf("chmod %s %#o: %s", f.Name(), fm, err)
 	}
 	checkMode(t, f.Name(), fm)
+
+}
+
+func TestChown(t *testing.T) {
+	f := newFile("TestChown", t)
```
rminnich, line 53: What you might do first is
if runtime.GOOS == "windows" || runtime.GOOS == "plan9" {
t.Skip()
}

Then you can drop the if on line 41.

leongross:
To design meaningful test cases for `chown`, I think you would need multiple users with different `uids` and `gids` and a user who has permission to change ownership of a file to this, a super user. In my opinion, there are 2 ways to integrate this:

1. update the container image for the CI so that there are accounts with said properties
2. mocking the file system and permissions 

The problem I see with solution 2 is that the implementation of `chown` relies on the syscall `chown`, so I wouldn't know how to mock that.

I would appreciate your input on that @rminnich 

EDIT: Maybe I am mistaken but as I see it, `golang` does not do any of the described tests but checks for an [error of an invalid gid](https://github.com/golang/go/blob/master/src/syscall/syscall_linux_test.go), maybe this is the way to go here as well?

leongross:
@deadprogram do you have an opinion on that?

leongross:
@deadprogram

leongross:
@rminnich

src/os/file_anyos.go
```diff
@@ -147,6 +147,23 @@ func Chmod(name string, mode FileMode) error {
 	return nil
 }
 
+// Chown changes the numeric uid and gid of the named file.
+// If the file is a symbolic link, it changes the uid and gid of the link's target.
+// A uid or gid of -1 means to not change that value.
+// If there is an error, it will be of type *PathError.
+//
+// On Windows or Plan 9, Chown always returns the syscall.EWINDOWS or
+// EPLAN9 error, wrapped in *PathError.
```
aykevl: TinyGo doesn't support Plan 9, so I'm not sure how useful this comment is?

aykevl (commented):
I think it's fine to keep this lightly tested, for example by checking that the syscall returns an expected permission error (or even just "file not found" - it also means the call works at least somewhat).

Yes it would be great to be able to test everything, but that's very hard to do for things like this and you'll end up putting a lot of effort into what's essentially just a wrapper for a libc function.

src/os/file_anyos.go
leongross: You're right, this should be removed.

deadprogram:
@leongross can you please rebase this PR against the latest `dev`? Thanks!

leongross:
@deadprogram rebase done, but noq quite sure where the runtime panic originates. Do you know anything about this?

leongross:
@deadprogram @aykevl

deadprogram:
@leongross looks like some CI fails on this PR. You probably also need to rebase against latest `dev`.

leongross:
@deadprogram after rebasing against the current dev branch the test still pancis:

```
panic: runtime error at 0x0000000104659d7c: type assert failed
FAIL	os	0.019s
```

leongross:
The call stack I could back trace so far looks like this:

* runtime panic
* src/runtime/interface.go:87
* compiler/interface.go:768
* compiler/compiler.go:2468

leongross:
The old errors should be fixed except for the ci (failing probably due to https://github.com/tinygo-org/tinygo/issues/4347). Could you review it? @deadprogram

leongross:
@deadprogram I think this was the last fix now, mind to review?

deadprogram:
Thanks for the additions @leongross and too @rminnich and @aykevl for review.

Now squash/merging.

PR #4215, opened 2024-03-30
Make most RP2040 boards available for simulation

aykevl:
This makes these boards available to be used on play.tinygo.org, once the appropriate board specs (SVG files etc) are created. In particular, this includes the Gopher Badge which I'm currently working on (but adding most other boards was trivial so I included those as well).


deadprogram:
@aykevl looks like actual error here? https://github.com/tinygo-org/tinygo/actions/runs/8490815316/job/23262246941?pr=4215#step:21:130

aykevl:
Yes indeed, I need to think of a proper fix.

aykevl:
Ok, unless I made a mistake this should now pass the smoke test.

deadprogram:
Moving those things out of the board files and removing the duplication is great! Thanks @aykevl

Issue #4206, opened 2024-03-25
core: compile failures due to SSA race condition


scottfeldman:
Several folks on the Slack TinyGo channel reported in-frequent compile failures with 0.31.2.

Here's some I've captured in my build logs.  I would estimate it happens 1/40 compiles.  A re-compile is the work-around.

```
#13 144.9 /usr/local/bin/tinygo build -target wioterminal -o ../../relays-wioterminal.uf2 -stack-size 8kb -size short /tmp/build-3411892484.go
#13 148.5 2024/03/14 07:35:32 Error generating UF2s: exit status 1: Basic Block in function '(*github.com/***/dean.mutex).islocked' does not have terminator!
#13 148.5 label %entry
#13 148.5 error: verification error after compiling package command-line-arguments
```

```
#14 141.8 /usr/local/bin/tinygo build -target wioterminal -o ../../relays-wioterminal.uf2 -stack-size 8kb -size short /tmp/build-1005049736.go
#14 145.2 2024/03/17 04:38:03 Error generating UF2s: exit status 1: Basic Block in function '(*struct{github.com/***/dean.Thing; github.com/***/device/target.Targets "json:\"-\""; ModelStruct string "json:\"-\""; github.com/***/device.WifiAuth "json:\"-\""; DialURLs string "json:\"-\""; DeployParams string; deployValues net/url.Values; github.com/***/device.ViewMode "json:\"-\""; WsScheme string "json:\"-\""; fs embed.FS; github.com/***/device.deviceOS}).TestFlag' does not have terminator!
#14 145.2 label %entry
#14 145.2 error: verification error after compiling package command-line-arguments
```

```
#13 197.4 /usr/local/bin/tinygo build -target wioterminal -o ../../relays-wioterminal.uf2 -stack-size 8kb -size short /tmp/build-3732778842.go
#13 200.7 2024/03/14 07:05:36 Error generating UF2s: exit status 2: SIGSEGV: segmentation violation
#13 200.7 PC=0x813d1c m=3 sigcode=18446744073709551610 addr=0x4a5
#13 200.7
#13 200.7 goroutine 110 gp=0xc017982e00 m=3 mp=0xc000051008 [running]:
#13 200.7 github.com/tinygo-org/tinygo/compiler.getPos({0x0, 0x0})
[...]

PR #4218, opened 2024-04-05
os.FindProcess, process.Release

leongross:
This implementation of `os.FindProcess` provides minimal functionality while being aligned to the process design pattern of `tinygo`.

### DIFF
--- a/src/os/exec_test.go
+++ b/src/os/exec_test.go
@@ -6,7 +6,6 @@ package os_test
 
 import (
 	. "os"
-	"runtime"
 	"testing"
 )
 
--- a/src/os/exec_test.go
+++ b/src/os/exec_test.go
@@ -1,0 +1,32 @@
+// Copyright 2024 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package os_test
+
+import (
+	. "os"
+	"runtime"
+	"testing"
+)
+
+func TestFindProcess(t *testing.T) {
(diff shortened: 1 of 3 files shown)


src/os/exec_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	. "os"
+	"runtime"
+	"testing"
+)
+
+func TestFindProcess(t *testing.T) {
+	// NOTE: For now, we only test the Linux case since only exec_posix.go is currently the only implementation.
+	if runtime.GOOS == "linux" {
```
aykevl: exec_posix.go does in fact include Windows, and because it's all no-ops anyway I think it would be fine to test this on any platform (including Windows).

aykevl (commented):
Looks fine to me.

src/os/exec_test.go
leongross: that makes sense, I will remove it

leongross:
To me it looks like these errors are unrelated to the code this patch covers. Any proposal on how to proceed here @aykevl ?

leongross:
@aykevl any idea what's the problem here?

leongross:
seems to be blocked by https://github.com/tinygo-org/tinygo/issues/4206

aykevl:
It's a race that only happens sometimes. I've restarted the test.

leongross:
@aykevl I guess we won the ci race, let's merge this :)

deadprogram:
Now merging, thanks for the addition @leongross and to @aykevl for review. Now merging.

leongross:
thanks a lot @deadprogram!

PR #4224, opened 2024-04-12
machine: add __tinygo_spi_tx function to simulator

aykevl:
This is much, _much_ faster than `__tinygo_spi_transfer` which can only transfer a single byte at a time. This is especially important for simulated displays.

I've already implemented the browser side of this on the playground and have used this patch for local testing where it massively speeds up display operations.


deadprogram (approved):
This is great, thanks @aykevl

Issue #4225, opened 2024-04-12
tinygo source displeases linters


dkegel-fastly:
Eagle-eyed new contributors have noticed that the tinygo source tree does not pass some lint checks.

The project probably shouldn't rely on volunteers to run lint.   

To address this, we might run lint in ci and/or add a "make lint" target in GNUMakefile.

To avoid adding too many new dependencies, we might want to keep the number of linters small, e.g. just go vet and revive, for starters, and avoid metalinters.

dkegel-fastly:
Plain old lint target exists, runs one linter now, and several problems have been fixed.

I think we can close this now; further lint enhancements can continue without this bug report being open.

deadprogram:
This was released with `v0.32.0` so now closing. Thank you everyone!

PR #4226, opened 2024-04-12
Add lint make target, use it in ci

dkegel-fastly:
Seems to work, and seems fast enough.

This is just a proof of concept of the first step in fixing https://github.com/tinygo-org/tinygo/issues/4225
It only checks a subset of files with a subset of revive rules, and ignores all but a small subset of those errors.
The first commit finds one error, and its build on circleci is red.
The 2nd commit fixes one such error, and its build on circleci gets past the lint check.

Also runs lint check in github ci, since pull requests don't run circleci (?).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,10 +85,13 @@ commands:
           condition: <<parameters.fmt-check>>
           steps:
           - run:
+              name: Install linter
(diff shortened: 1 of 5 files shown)


dkegel-fastly:
dammit, wanted to do that against dev.

PR #4227, opened 2024-04-12
Add lint make target, use it in ci (this time against dev branch)

dkegel-fastly:
See https://github.com/tinygo-org/tinygo/issues/4225
    
Runs in both circleci and github, circleci is run on branch push, github is run on PR
    
Revive builds so fast, don't bother installing it; saves us wondering which one we get 
    
Uses tools.go idiom to give control over linter versions to go.mod.
    
Also pacifies linter re AppendToGlobal as a token first fix.
    
TODO: gradually expand the number of directories that are linted,
uncomment more entries in revive.toml, and fix or suppress the 
warnings lint finds.
    
TODO: add linters "go vet" and staticcheck
    
NOT TODO: don't add metalinters like golangci-lint that pull in
lots of new of dependencies; we'd rather not clutter go.mod that
much, let alone open ourselves up to the additional attack surface.


.github/workflows/linux.yml
```diff
@@ -113,6 +113,10 @@ jobs:
           gem install --version 4.0.7 public_suffix
           gem install --version 2.7.6 dotenv
           gem install --no-document fpm
+      - name: Install linter
+        run: go install github.com/mgechev/revive@v1.3.7
```
ydnar: This might be an opportunity to add a `internal/tools/tools.go` package with `go generate` lines, so the specific version(s) of Go tools can be kept in go.mod and updated with normal dependency updates.

ydnar: Also: maybe add `staticcheck`?

dkegel-fastly:
Let's add staticcheck in a future PR.  I've implemented the internal/tools/tools.go idea.

deadprogram:
This seems like a very good idea. @dkegel-fastly do you want to squash these commits into a single one?

dkegel-fastly:
Squashed, updated commit message a bit.

deadprogram:
Thank you very much @dkegel-fastly now merging.

PR #4228, opened 2024-04-14
Lint: lint and fix src/{os,reflect}

dkegel-fastly:
1. lint: expand to src/reflect, fix or suppress what it found.
2. fmt-check: pacify after lint fix, sigh
3. add go:build tools guard on internal/tools/tools.go to avoid complains from "go test ./..."

(A previous draft also fixed spelling errors, but that was landed separately for easier review.)
(A previous draft commented out some unreachable code, but this draft just suppresses the lint warning on those two files.)

See https://github.com/tinygo-org/tinygo/issues/4225


deadprogram:
Hi @dkegel-fastly looks like some actual CI errors here: https://github.com/tinygo-org/tinygo/actions/runs/8678079555/job/23794537729?pr=4228#step:16:40

dkegel-fastly:
Fixed, and expanded to src/os as well.

src/reflect/all_test.go
```diff
@@ -4146,30 +4146,33 @@ func TestValuePanic(t *testing.T) {
 
 func shouldPanic(expect string, f func()) {
 	return
-	defer func() {
```
deadprogram: @dkegel-fastly did you mean to make this change?

dkegel-fastly: Yes.  lint pointed out that the defer never ran.

dgryski: Oh, that's my fault.  `all_test.go` was a copy from upstream with patches to make it run on tinygo + wasi.  Since tinygo doesn't support panic/recover on wasi, I stubbed that out.  We should probably have a wasm-specific check to disable that, and leave a comment for the next person.

dkegel-fastly:
ah, right, thanks!

src/reflect/all_test.go
dkegel-fastly: OK!  For now I undid my comments, and just suppress the lint warning on the two files where this happened.

dkegel-fastly:
Ready for re-review!

dgryski (approved):
LGTM

PR #4241, opened 2024-04-23
crypto/generic

leongross:
Building applications with cryptographic operations still lacks a lot of support due to excessive usage of assembly in the golang standard library. This PR intends to replace them in the tinygo build process with the generic implementations.

### DIFF
--- a/compiler/alias.go
+++ b/compiler/alias.go
@@ -24,6 +24,10 @@ var stdlibAliases = map[string]string{
 	"crypto/sha256.block":      "crypto/sha256.blockGeneric",
 	"crypto/sha512.blockAMD64": "crypto/sha512.blockGeneric",
 
+	// AES
+	"crypto/aes.decryptBlockAsm": "crypto/aes.decryptBlock",
+	"crypto/aes.encryptBlockAsm": "crypto/aes.encryptBlock",
+
 	// math package
 	"math.archHypot": "math.hypot",
 	"math.archMax":   "math.max",


dgryski:
This has been fixed on Go `tip` with the `purego` build tag.

aykevl:
> This has been fixed on Go `tip` with the `purego` build tag.

That would be a better long-term fix, yes. But the change seems fine by me if it unblocks someone right now, any reason not to merge this?

dgryski:
If it works it's probably fine to merge, assuming it doesn't then break stuff when we build with the `purego` build tag. (But yes, let's merge for now and fix later if needed.)

deadprogram:
Thanks for the code @leongross and for the review @dgryski and @aykevl now merging.

leongross:
Great, thanks for taking care of this!

leongross:
I have quite a few new cryptography functions that I need to use; what is the status of adjusting tinygo to make use `purego` usable? Adding them all manually to the aliasing does not seem to be a good and reliable idea.

Could you tell me what steps need to be taken? I would we willing to help with this if needed.
@dgryski @deadprogram @aykevl

dgryski:
@leongross Does simply adding `-tags=purego` to your build command line not work?  What pieces are missing?  IIRC that *should* start working (starting with 1.23).

leongross:
Since go 1.23 is not yet released, I would have to use the go1.23rc1 for that right?

What I did was
1. Download said `go1.23rc1`
2. Get an example implementation using the golang standard library
3. Build using `GOROOT=$HOME/sdk/go1.23rc1 tinygo build -tags=purego .`

I get the following errors:
```
$HOME/sdk/go1.23rc1/src/bytes/bytes.go:116:18: undefined: bytealg.LastIndexByte
$HOME/sdk/go1.23rc1/src/bytes/bytes.go:125:17: undefined: bytealg.LastIndexRabinKarp
$HOME/sdk/go1.23rc1/src/bytes/bytes.go:130:17: undefined: bytealg.LastIndexByte
$HOME/sdk/go1.23rc1/src/bytes/bytes.go:1329:32: cannot use s[i:] (value of type []byte) as string value in argument to bytealg.IndexRabinKarp
$HOME/sdk/go1.23rc1/src/bytes/bytes.go:1329:39: cannot use sep (variable of type []byte) as string value in argument to bytealg.IndexRabinKarp
```

Using the default golang compiler on that version works fine although (`go1.23rc1 build -tags=purego .`)

Tinygo is the upstream dev branch `tinygo version 0.33.0-dev-e6caa3fe linux/amd64 (using go version go1.21.11 and LLVM version 18.1.2)`. I see that tinygo version states that it uses go1.21.11, but I'm not sure how to change that.

Also, the purego tag seems to be set as [default build tag](https://github.com/tinygo-org/tinygo/blob/release/compileopts/config.go#L80) for a while now

PR #4242, opened 2024-04-26
rp2040: fix timeUnit type

aykevl:
It should be int64 like for all other systems, not uint64.

See: https://github.com/tinygo-org/tinygo/pull/4239


deadprogram:

PR #4243, opened 2024-04-27
compileopts: fix race condition

aykevl:
Appending to a slice can lead to a race condition if the capacity of the slice is larger than the length (and therefore the returned slice will overwrite some fields in the underlying array).

This fixes that race condition. I don't know how severe this particular one is. It shouldn't be that severe, but it is a bug and fixing it makes it easier to hunt for possibly more serious race conditions.

See: https://github.com/tinygo-org/tinygo/issues/4206

### DIFF
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -74,7 +74,8 @@ func (c *Config) GOARM() string {
 
 // BuildTags returns the complete list of build tags used during this build.
 func (c *Config) BuildTags() []string {
-	tags := append(c.Target.BuildTags, []string{
+	tags := append([]string(nil), c.Target.BuildTags...) // copy slice (avoid a race)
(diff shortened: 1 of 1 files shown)


deadprogram:
@aykevl looks like a possible CI fail here https://app.circleci.com/pipelines/github/tinygo-org/tinygo/10816/workflows/c6200aa8-e8c0-4174-8e0c-2610f4f57e28/jobs/42996#step-118-3605_94

aykevl:
@deadprogram honestly that looks like just another race condition to me. (This PR doesn't fix the race we're hitting all the time, but it does fix _a_ race condition).

deadprogram:
> honestly that looks like just another race condition to me. 

I think you are quite correct. My bad.

deadprogram:

PR #4245, opened 2024-04-28
unix, wasm: call exit(2) instead of abort or unreachable

aykevl:
This matches upstream Go, and I think it's a better idea in general.

Specifically, it avoids triggering an 'undefined' instruction in wasm when there's a panic. Many people confuse this 'undefined' instruction (with associated backtrace) with the actual bug.

A possible downside is that this unreachable instruction does show a backtrace, which may be helpful in finding the underlying bug. Calling exit(2) doesn't have this behavior.

### DIFF
--- a/src/runtime/runtime_tinygowasm.go
+++ b/src/runtime/runtime_tinygowasm.go
@@ -65,8 +65,14 @@ func now() (sec int64, nsec int32, mono int64) {
 	return
 }
 
-// Abort executes the wasm 'unreachable' instruction.
+// Abort exits the program with an error code.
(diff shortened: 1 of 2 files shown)


deadprogram:
Seems like some failures with this PR @aykevl

dgryski:
Given that we don't have stack traces on panic wasm, I'd be much happier leaving this as the `trap` instruction so that we at least get a stack trace from the wasm runtime.

aykevl:
@dgryski ok, makes sense. I agree stack traces are important.

PR #4246, opened 2024-04-28
simulator: fix I2C support

aykevl:
- Add ReadRegister and WriteRegister (because they're still used by some packages).
  - Fix out-of-bounds panic in I2C.Tx when either w or r has a length of zero (or is nil).


deadprogram (approved):

PR #4248, opened 2024-04-28
Add 'make spell' target, fix what it finds.

dkegel-fastly:
In source files, only checks comments.

Seems to miss a few words, but it is fast and easy.

We can add more checks to "make spell" later.


dgryski (approved):
LGTM

deadprogram (approved):
Thanks for adding this feature @dkegel-fastly and to @dgryski for review. Now merging.

PR #4252, opened 2024-04-29
make: use release esp-17.0.1_20240419 tag for LLVM source build

deadprogram:
This PR changes the makefile to use the release `esp-17.0.1_20240419` tag for LLVM source from the Espressif LLVM fork for the TinyGo builds. This is intended to make our builds more reproducible since the current branch based approach the target branch can change at any time.


dkegel-fastly (approved):
LGTM

deadprogram:
Thanks for review @dkegel-fastly now merging.

PR #4254, opened 2024-05-04
Makefile: add lld to list of build targets for wasm-ld

paralin:
The current list of targets does not build wasm-ld.

wasm-ld is a symlink created in ./llvm-build/bin pointing to ./lld.

Add the "lld" build target to get wasm-ld into ./llvm-build/bin.

Fixes a build failure where wasm-ld is not found.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -166,7 +166,7 @@ LIB_NAMES = clang $(CLANG_LIB_NAMES) $(LLD_LIB_NAMES) $(EXTRA_LIB_NAMES)
 # library path (for ninja).
 # This list also includes a few tools that are necessary as part of the full
 # TinyGo build.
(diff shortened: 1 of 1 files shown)


paralin:
CI Failure looks unrelated to changes in this PR

deadprogram:
> CI Failure looks unrelated to changes in this PR

You are correct. Restarted build and it passes.

deadprogram (approved):
Seems legit to me, thanks @paralin

aykevl:
I wonder why this is needed? I don't see where wasm-ld is needed (unlike `llvm-ar`, `llvm-nm` etc which are needed in GNUmakefile).

paralin:
I'm not sure, I just got a build failure saying that tool is missing, and then after adding it to the build targets, it's fixed. So, not sure where the actual issue was.

Issue #2311, opened 2021-11-24
nucleo-l432kc, nucleo-l031k6, nucleo-f722ze seemingly lack the ability to change I2C frequency


rayozzie:
For both of these boards there are consts for TWI_FREQ, however the I2CConfig type has no Frequency field and thus it is not apparent how to configure bus speed for I2C.

https://tinygo.org/docs/reference/microcontrollers/machine/nucleo-l432kc/
https://tinygo.org/docs/reference/microcontrollers/machine/nucleo-l031k6/
https://tinygo.org/docs/reference/microcontrollers/machine/nucleo-f722ze

Thank you for your consideration of this fix/enhancement.

deadprogram:
@kenbell I was looking at this briefly, and saw that we are using some constant values based on clock speed generated by STM32CubeMX. We could keep doing this as an interim solution if we can also generate those 4 values for the typical standard and fast mode speeds for each of the boards this is currently being used for.

I was looking at STM32CubeMX but it was not immediately obvious how to regenerate those values. Could you show me how you did that and/or help out to generate values for the various `getFreqRange()` functions?

kenbell:
@deadprogram you first have to set the peripheral clock to the correct MHz in the 'Clock Configuration' tab.  For L432, it's this (80MHz on PCLK1):
![image](https://github.com/tinygo-org/tinygo/assets/509378/8daa9f00-192e-4836-8837-fa65414b68e8)

Resolve the clock configuration.

Then enable I2C1 peripheral, you'll now see in gray the magic value in the 'timing' setting (0x10909CEC in this case for 100kHz standard mode):
![image](https://github.com/tinygo-org/tinygo/assets/509378/41a05e61-c5ae-4146-8650-043fa43b6779)

deadprogram:
Thanks @kenbell that is exactly what I was looking for. PR coming later.

deadprogram:
Merged into `dev`, tagged to close on next release. Thanks @rayozzie for report and to @kenbell for helping me out.

deadprogram:
This was released with `v0.32.0` so now closing. Thank you everyone!

PR #4255, opened 2024-05-06
machine/stm32: add i2c Frequency and SetBaudRate() function for missing boards

deadprogram:
This PR adds i2c config `Frequency` and `SetBaudRate()` function for STM32 boards that were lacking this implementation.

Addresses #2311

### DIFF
--- a/src/machine/machine_stm32_i2c_revb.go
+++ b/src/machine/machine_stm32_i2c_revb.go
@@ -45,11 +45,21 @@ type I2C struct {
 
 // I2CConfig is used to store config info for I2C.
 type I2CConfig struct {
-	SCL Pin
-	SDA Pin
+	Frequency uint32
+	SCL       Pin
+	SDA       Pin
 }
 
 func (i2c *I2C) Configure(config I2CConfig) error {
+	// Frequency range
+	switch config.Frequency {
+	case 0:
+		config.Frequency = 100 * KHz
+	case 10 * KHz, 100 * KHz, 400 * KHz, 500 * KHz:
+	default:
+		return errI2CNotImplemented
+	}
+
 	// disable I2C interface before any configuration changes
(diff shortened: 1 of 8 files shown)


deadprogram:
Pushed update, since I realized my implementation would not work correctly to change the baudrate after the initial implementation due to needing to disable the i2c before changing frequency. Now corrected.

src/machine/machine_stm32_i2c_revb.go
```diff
@@ -86,8 +95,23 @@ func (i2c *I2C) Configure(config I2CConfig) error {
 
 // SetBaudRate sets the communication speed for I2C.
 func (i2c *I2C) SetBaudRate(br uint32) error {
```
kenbell, line 97: I'm not sure what the correct semantics we want are, but the logic here differs from NRF in that an unsupported value in NRF results in silently using 100KHz and will leave I2C enabled in any case.

NRF: https://github.com/tinygo-org/tinygo/blob/6384ecace093df2d0b93915886954abfc4ecfe01/src/machine/machine_nrf.go#L292

It seems like there's three possible behaviors for unsupported rates:
1. Change nothing (do validation before any changes) and return error
2. Force unsupported to the default 100KHz (hard to debug for by app devs?)
3. Disable the peripheral and return an error

(I'm not sure NRF implementation chose the best option)

deadprogram: > Change nothing (do validation before any changes) and return error

That seems like the best option to me.

deadprogram: Updated to ensure that is what happens e.g. make no change and return error if invalid.

kenbell (approved):
lgtm

deadprogram:
Thank you for review and feedback @kenbell now merging.

PR #4258, opened 2024-05-10
machine: add TxFifoFreeLevel() for CAN

sago35:
This PR adds a function called TxFifoFreeLevel() for the CAN functionality on ATSAME5X.
It allows checking if there is space available in the FIFO before setting it, enabling safe transmission even with a high volume of data to be sent.


deadprogram:
Thanks for the addition @sago35 now merging.

PR #4259, opened 2024-05-12
Add I2C support for esp32

jfreymuth:
Mostly copied from `machine_esp32c3_i2c.go`, but there are some differences, especially with timeout handling (see line 165).

Tested on an M5Paper, where I could successfully communicate with multiple devices and recover from errors in a long-running program.

### DIFF
--- a/src/machine/machine_esp32_i2c.go
+++ b/src/machine/machine_esp32_i2c.go
@@ -47,7 +47,6 @@ func (i2c *I2C) Configure(config I2CConfig) error {
 }
 
 func (i2c *I2C) initAll() {
-
 	i2c.initClock()
 	i2c.initNoiseFilter()
 	i2c.initPins()
@@ -120,7 +119,6 @@ func (i2c *I2C) initPins() {
 
 //go:inline
 func (i2c *I2C) initFrequency() {
-
 	clkmDiv := i2cClkSourceFrequency/(i2c.config.Frequency*1024) + 1
 	sclkFreq := i2cClkSourceFrequency / clkmDiv
 	halfCycle := sclkFreq / i2c.config.Frequency / 2
@@ -228,7 +226,6 @@ type i2cCommand struct {
 func nanotime() int64
 
 func (i2c *I2C) transmit(addr uint16, cmd []i2cCommand, timeoutMS int) error {
-
 	if i2c.Bus.GetSR_BUS_BUSY() == 1 {
 		i2c.resetBus()
 	}
@@ -388,7 +385,7 @@ func (i2c *I2C) Tx(addr uint16, w, r []byte) (err error) {
 }
 
 func (i2c *I2C) SetBaudRate(br uint32) error {
-	return nil
+	return errI2CNotImplemented
 }
 
 func (p Pin) pinReg() *volatile.Register32 {
--- a/src/machine/machine_esp32_i2c.go
+++ b/src/machine/machine_esp32_i2c.go
@@ -1,0 +1,415 @@
+//go:build esp32
+
+package machine
+
+import (
+	"device/esp"
+	"runtime/volatile"
+	"unsafe"
+)
+
+var (
+	I2C0 = &I2C{Bus: esp.I2C0, funcSCL: 29, funcSDA: 30}
+	I2C1 = &I2C{Bus: esp.I2C1, funcSCL: 95, funcSDA: 96}
+)
+
+type I2C struct {
+	Bus              *esp.I2C_Type
+	funcSCL, funcSDA uint32
+	config           I2CConfig
+}
+
+// I2CConfig is used to store config info for I2C.
+type I2CConfig struct {
+	Frequency uint32 // in Hz
+	SCL       Pin
+	SDA       Pin
+}
+
+const (
+	i2cClkSourceFrequency = uint32(80 * MHz)
+)
+
+func (i2c *I2C) Configure(config I2CConfig) error {
+	if config.Frequency == 0 {
+		config.Frequency = 400 * KHz
+	}
+	if config.SCL == 0 {
+		config.SCL = SCL_PIN
+	}
+	if config.SDA == 0 {
+		config.SDA = SDA_PIN
+	}
+	i2c.config = config
+
+	i2c.initAll()
+	return nil
+}
+
+func (i2c *I2C) initAll() {
+
+	i2c.initClock()
+	i2c.initNoiseFilter()
+	i2c.initPins()
+	i2c.initFrequency()
+	i2c.startMaster()
+}
+
+//go:inline
+func (i2c *I2C) initClock() {
+	// reset I2C clock
+	if i2c.Bus == esp.I2C0 {
+		esp.DPORT.SetPERIP_RST_EN_I2C0_EXT0_RST(1)
+		esp.DPORT.SetPERIP_CLK_EN_I2C0_EXT0_CLK_EN(1)
+		esp.DPORT.SetPERIP_RST_EN_I2C0_EXT0_RST(0)
+	} else {
+		esp.DPORT.SetPERIP_RST_EN_I2C_EXT1_RST(1)
+		esp.DPORT.SetPERIP_CLK_EN_I2C_EXT1_CLK_EN(1)
+		esp.DPORT.SetPERIP_RST_EN_I2C_EXT1_RST(0)
+	}
+	// disable interrupts
+	i2c.Bus.INT_ENA.Set(0)
+	i2c.Bus.INT_CLR.Set(0x3fff)
+
+	i2c.Bus.SetCTR_CLK_EN(1)
+}
+
+//go:inline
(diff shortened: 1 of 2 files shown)


src/machine/machine_esp32_i2c.go
```diff
@@ -0,0 +1,415 @@
...
+		config.SDA = SDA_PIN
+	}
+	i2c.config = config
+
+	i2c.initAll()
+	return nil
+}
+
+func (i2c *I2C) initAll() {
+
```
deadprogram: Nit: remove extra line.

src/machine/machine_esp32_i2c.go
```diff
@@ -0,0 +1,415 @@
...
+	cmd  i2cCommandType
+	data []byte
+	head int
+}
+
+//go:linkname nanotime runtime.nanotime
+func nanotime() int64
+
+func (i2c *I2C) transmit(addr uint16, cmd []i2cCommand, timeoutMS int) error {
+
```
deadprogram: Nit: remove blank line.

src/machine/machine_esp32_i2c.go
```diff
@@ -0,0 +1,415 @@
...
+	if len(r) > 0 {
+		cmd = append(cmd, i2cCommand{cmd: i2cCMD_READ, data: r})
+	}
+	cmd = append(cmd, i2cCommand{cmd: i2cCMD_STOP})
+
+	return i2c.transmit(addr, cmd, timeout)
+}
+
+func (i2c *I2C) SetBaudRate(br uint32) error {
+	return nil
```
deadprogram: Perhaps return `errI2CNotImplemented` instead? https://github.com/tinygo-org/tinygo/blob/dev/src/machine/i2c.go#L39

src/machine/machine_esp32_i2c.go
```diff
@@ -0,0 +1,415 @@
...
+}
+
+func (p Pin) pinReg() *volatile.Register32 {
+	return (*volatile.Register32)(unsafe.Pointer((uintptr(unsafe.Pointer(&esp.GPIO.PIN0)) + uintptr(p)*4)))
```
deadprogram, line 392: Why not also use `unsafe.Add()` here in similar fashion to the implementation of `nextAddress()` function just below?

src/machine/machine_esp32_i2c.go
```diff
@@ -0,0 +1,415 @@
...
+}
+
+func nextAddress(reg *volatile.Register32) *volatile.Register32 {
+	return (*volatile.Register32)(unsafe.Add(unsafe.Pointer(reg), 4))
+}
+
+// CheckDevice does an empty I2C transaction at the specified address.
```
deadprogram, line 399: Seems like a cool function, but I am just wondering if it really belongs in this PR? Perhaps it can either be added as something available to all i2c implementations, which as least right now it is not. What do you think?

deadprogram:
This is really great @jfreymuth thank you very much for working on it. Please see my comments inline.

src/machine/machine_esp32_i2c.go
jfreymuth: The function is copied from [machine_esp32c3.go#L114](https://github.com/tinygo-org/tinygo/blob/1d9f26cee1f1501b09647186a25ba29aa6a0c58c/src/machine/machine_esp32c3.go#L114), which uses uintptr addition (but it seems to be an outlier in that file too). For consistency this function should probably be moved to `machine_esp32.go`?

src/machine/machine_esp32_i2c.go
jfreymuth: Another option would be to modify `Tx` so that `Tx(addr, nil, nil)` returns an error if there is no device with that address. This seems to be how other microcontroller libraries do it, e.g. Arduino:
```c++
WIRE.beginTransmission(address);
error = WIRE.endTransmission();
```
That behavior also matches the method comment better, which implies that the address is always written:
```go
// It clocks out the given address, writes the bytes in w, reads back len(r)
// bytes and stores them in r, and generates a stop condition on the bus.
```
(currently the method just does nothing if both slices are empty)

Ideally that would be reflected in the generic i2c documentation and other implementations as well, which – I agree – is not part of this PR

src/machine/machine_esp32_i2c.go
deadprogram: > For consistency this function should probably be moved to machine_esp32.go

Agreed, but that can happen in a separate PR.

src/machine/machine_esp32_i2c.go
deadprogram: > Ideally that would be reflected in the generic i2c documentation and other implementations as well, which – I agree – is not part of this PR

That makes sense to me! :smile:

deadprogram:
Oh yeah, one thing I almost forgot is we need to also add a smoke test for i2c on esp32.

aykevl:
> Mostly copied from `machine_esp32c3_i2c.go`, but there are some differences, especially with timeout handling (see line 165).

That's going to increase the maintenance burden. Can you refactor the code so that most/all common code is shared and only the chip-specific things are in the appropriate chip specific files?
For example:
- `machine_esp32xx_i2c.go` for the esp32 and esp32c3 with most of the I2C code
- `machine_esp32c3_i2c.go` contains the esp32c3 specific bits (for example, constants or functions called from machine_esp32xx_i2c.go)
- same for `machine_esp32_i2c.go`

This is also important because I expect other chips in the ESP32 family to be very similar, they can share code in a similar way.

aykevl:
We have something similar for the nrf chips:
- https://github.com/tinygo-org/tinygo/blob/release/src/machine/machine_nrf528xx.go contains most of the I2C/UART/PWM etc implementation
- https://github.com/tinygo-org/tinygo/blob/release/src/machine/machine_nrf52840.go contains a `setPins` method because it varies by chip

jfreymuth:
I can try, but I don't have any esp32c3 devices, so I'm not sure how to verify I'm not breaking anything with the refactor

deadprogram:
Perhaps we can merge this PR (with a smoketest added, please) and then refactor the i2c implementation as described by @aykevl in a separate PR. I have both boards and am willing to help out with this.

aykevl:
Sounds good to me!
(I'm just hoping we won't forget this...)

I'm not sure about the smoketest though: what would be gained from that? We already have a smoketest for the esp32, so any syntax or type errors would already be caught.

deadprogram:
> I'm not sure about the smoketest though: what would be gained from that? We already have a smoketest for the esp32, so any syntax or type errors would already be caught.

Just to ensure that the i2c parts compile for the `esp32` build tag.

aykevl:
In that case I'd suggest adding the ESP32 here:
https://github.com/tinygo-org/tinygo/blob/6384ecace093df2d0b93915886954abfc4ecfe01/src/machine/i2c.go#L1

We already have plenty of smoke tests that show the machine package (including I2C) compiles for the ESP32:
https://github.com/tinygo-org/tinygo/blob/6384ecace093df2d0b93915886954abfc4ecfe01/GNUmakefile#L751-L764

deadprogram:
> In that case I'd suggest adding the ESP32 here:

https://github.com/tinygo-org/tinygo/pull/4259/files#diff-0fff984543b29e2e510fb57be78ac8464fc10ba2d494ba94fa7b686e275b3ee0R1

That is in fact in this very PR already! :) In which case I think we can merge now.

deadprogram:
Now going to squash/merge. Thank you @jfreymuth for working on this and to @aykevl for review.

src/machine/machine_esp32_i2c.go
```diff
@@ -0,0 +1,412 @@
...
+}
+
+// Tx does a single I2C transaction at the specified address.
+// It clocks out the given address, writes the bytes in w, reads back len(r)
+// bytes and stores them in r, and generates a stop condition on the bus.
+func (i2c *I2C) Tx(addr uint16, w, r []byte) (err error) {
+	// timeout in microseconds.
+	const timeout = 40 // 40ms is a reasonable time for a real-time system.
+
+	cmd := make([]i2cCommand, 0, 8)
```
soypat, line 374: This could be added as a field to I2C to not risk heap allocation
```go
 cmdbuf [8]i2cCommand
 ```

File: src/machine/machine_esp32_i2c.go [hunk 7]
```diff
@@ -0,0 +1,412 @@
...
+}
+
+// CheckDevice does an empty I2C transaction at the specified address.
+// This can be used to find out if any device with that address is
+// connected, e.g. for enumerating all devices on the bus.
+func (i2c *I2C) CheckDevice(addr uint16) bool {
+	// timeout in microseconds.
+	const timeout = 40 // 40ms is a reasonable time for a real-time system.
+
+	cmd := []i2cCommand{
```
soypat, line 406: Same as above, we could store this memory in I2C.

PR #4263, opened 2024-05-16
fix fpm ci installion

leongross:
The CI seems to fail on the fpm installation and recommends to install the `rub-dev` package.
Doing this fixes the issue (#4218)


deadprogram:
Thank you for the fix @leongross I just noticed those failures had started.

Now merging.

PR #4265, opened 2024-05-18
machine/renesas: WIP on adding support for Arduino Uno R4

deadprogram:
This is a very very early draft PR to hopefully help jumpstart adding support for the Arduino Uno R4.


gandarez:
What's the status of this?

dndeus:
Nothing yet with this one? :)

Any broad estimation?? I think we should disclose in the docs that it won't work with tinygo yet :)

deadprogram:
Made some progress here, now need to implement the runtime timers and then should be able to test blinky1.

dndeus:
That's awesome news, can't wait for it!

gandarez:
There are two versions for Uno R4, which is the Minima and Wifi. I think they might have distinct build tags and board definition.

Frelsaren:
@deadprogram I'm guessing this has been abandoned?

PR #4268, opened 2024-05-20
compiler,reflect: fix NumMethods for Interface type

### LINKED ISSUE #4260: `reflect.Type.NumMethod()` returns 0 for interface types, impacting CBOR unmarshalling
# Description
It appears that `reflect.Type.NumMethod()` returns 0 when called on interface types. This behavior is particularly problematic for functionalities like CBOR unmarshalling that rely on NumMethod() for type resolution.

# Potential Source of Issue
This issue may stem from changes made in the following commit: [acba074](https://github.com/tinygo-org/tinygo/commit/acba0748f1daad128166e09e82badb9f1f0b2a29)
The relevant code is:
```go
var numMethods int
if hasMethodSet {
  for i := 0; i < ms.Len(); i++ {
    if ms.At(i).Obj().Exported() {
      numMethods++
    }
  }
}
```

Here, `numMethods` is incremented only under the `hasMethodSet` condition, which is set to false if the type is an interface. This might not be the intended functionality.

# Example of Affected Code

```go
package main

import (
	"fmt"

	"github.com/fxamacker/cbor/v2"
)

type Marshallable interface {
	MarshalBinary() ([]byte, error)
	UnmarshalBinary(data []byte) error
}

type A struct {
	Data []byte
}

func (a *A) MarshalBinary() ([]byte, error) {
	return a.Data, nil
}

func (a *A) UnmarshalBinary(data []byte) error {
	a.Data = data
	return nil
}

type Message struct {
	SomeMarshallable Marshallable
}

func runMarshalUnmarshal() error {
	messageToMarshal := Message{
		SomeMarshallable: &A{Data: []byte("hello")},
	}

	// Marshal
	msgBytes, err := cbor.Marshal(messageToMarshal)
	if err != nil {
		return err
	}

	// Initiate with concrete type
	messageFromUnmarshal := Message{
		SomeMarshallable: &A{},
	}

	// Unmarshal
	err = cbor.Unmarshal(msgBytes, &messageFromUnmarshal)
	if err != nil {
		return err
	}

	// Print type of Marshallable
	fmt.Printf("%T\n", messageFromUnmarshal.SomeMarshallable)
	return nil
}

func main() {
	err := runMarshalUnmarshal()
	if err != nil {
		panic(err)
	}
}
```

# Expected Behavior
The code above is expected to print the type `*main.A` but instead it prints `[]uint8`. Removing the `if hasMethodSet` condition from the code reference above results with the desired behavior.

frenkel26:
Fixes #4260


aykevl (changes requested):
This change looks good to me.

Can you add a test? For example, you can add this one-line change:

```diff
--- a/testdata/reflect.go
+++ b/testdata/reflect.go
@@ -456,6 +456,7 @@ func showValue(rv reflect.Value, indent string) {
        case reflect.Interface:
                println(indent + "  interface")
                println(indent+"  nil:", rv.IsNil())
+               println(indent+"  NumMethod:", rv.NumMethod())
                if !rv.IsNil() {
                        showValue(rv.Elem(), indent+"  ")
                }
```

And then you can run the following:

```
go run ./testdata/reflect.go 2> testdata/reflect.txt
```

And then `tinygo test` _should_ pass (it doesn't for me, but it probably will pass with your fix).

dgryski (approved):
LGTM.  Thanks for the test.

aykevl (approved):
Looks good to me!
I confirmed it matches what `go` produces.

dgryski:
I'm going to see if there are any more tests we can uncomment from `reflect/all_test.go`.

dgryski:
https://github.com/tinygo-org/tinygo/pull/4276

PR #4266, opened 2024-05-18
LLVM 18 support

aykevl:

PR #4271, opened 2024-05-21
LLVM 18 support

aykevl:
2nd try, replaces https://github.com/tinygo-org/tinygo/pull/4266


aykevl:
Ready for review. The failed tests all seem to be problems with TinyHCI (devices not found), not necessarily a problem with this PR.

Unfortunately a change in LLVM 18 broke ld.lld for AVR (see https://github.com/llvm/llvm-project/pull/92695) so I had to patch LLVM. This means a few things:

 * We have to use a patched version of LLVM for release binaries. Honestly not a big deal, we've done this before. It just means we pick a different branch in the `make llvm-source` step.
 * AVR won't work correctly when linking against a system installed LLVM (so using `go install` without any flags).
   This not only affects people who build TinyGo manually on Linux and macOS, but also Linux distributions. Linux distributions might want to consider applying the first patch (or both patches) in https://github.com/llvm/llvm-project/pull/92695 to their LLVM build. CC @QuLogic

While this problem will likely be fixed in LLVM 19, it will likely continue to be a problem for LLVM 18 because the LLVM team recently said there probably won't be any point releases for LLVM 18.x anymore.

deadprogram:
> AVR won't work correctly when linking against a system installed LLVM

So this only impacts people who want to compile for AVR targets, who are also building their own TinyGo, correct? And if LLVM releases `18.1.7` then problem solved?

If this is correct, then we can probably go ahead and merge this. For people using ARM, RISCV, or WASM targets this will not impact with dev versions while it gets sorted out.

aykevl:
> So this only impacts people who want to compile for AVR targets

Correct

> who are also building their own TinyGo, correct?

Yes, but also for distributions that typically link against their own LLVM shared library. (But those distributions also don't support Xtensa, so AVR wouldn't be the only option). An example would be the Fedora tinygo package.

> And if LLVM releases `18.1.7` then problem solved?

Yes... but it's unlikely they will. From [the release notes](https://discourse.llvm.org/t/18-1-6-released/79068):

> There are no more planned 18.1.x releases, but there could be one if a critical issue is found. The release/19.x branch will be created on July 23, 2024.

So to get this fixed we'll probably have to wait until LLVM 19.

(...I should have tested LLVM 18 a while ago, so I could have gotten this fixed in a point release and not end up in this mess...).

In other news, this has been fixed in the LLVM main branch so it'll definitely be fixed in LLVM 19.

aykevl:
A workaround would be to build TinyGo against LLVM 17, which remains possible. You can use `go build -tags=llvm17` for this. So it not as big of a breakage as it might seem.

deadprogram:
> You can use `go build -tags=llvm17` for this.

That works to resolve those specific use cases, and LLVM 19 will be here soon enough anyhow.

Now merging, thanks @aykevl

deadprogram:
@aykevl just FYI https://github.com/llvm/llvm-project/releases/tag/llvmorg-18.1.7 was released after all. :smile:

aykevl:
@deadprogram yeah I saw that, but that's only to fix a single bug (that was bad enough that Clang couldn't compile itself, so yeah pretty critical). It doesn't include this AVR linker fix.

PR #4274, opened 2024-05-24
main: add -serial=rtt flag as possible option

aykevl:
I added this a while ago but forgot to change the help documentation.


deadprogram:

PR #4279, opened 2024-05-31
tweaks for wasm-opt running logic

dgryski:
1) if `-work` was passed don't overwrite original. .wasm binary
2) if `-x` was passed print out wasm-opt command line.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -825,10 +825,18 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 					args = append(args, "--asyncify")
 				}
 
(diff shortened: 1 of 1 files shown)


jameysharp:
I asked Damian for this because it helps with troubleshooting bugs introduced by wasm-opt if we can check what the wasm looked like before that. For example, I was just now able to use this PR to verify that `llvm-dwarfdump --verify` finds no errors in the DWARF data before wasm-opt, but finds a bunch of errors after wasm-opt, for this program:

```go
package main
import "os"
func main() {
        os.Lstat("some-filename")
}
```

ydnar (approved):

PR #4282, opened 2024-06-03
Add nodejs_20 to flake.nix, as it's required to run `make llvm-source`

elliottt:
When running through the commands listed in `flake.nix` for building locally, I found that I would get a build failure when running `make llvm-source` without having node available. As I don't have a global node install this failed, but it was easy enough to pin a version in `flake.nix`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -62,6 +62,7 @@
             # these tools are available in the host environment.
             #gdb
             #openocd
+            nodejs_20
           ];
           shellHook= ''
             # Configure CLANG, LLVM_AR, and LLVM_NM for `make wasi-libc`.


aykevl:
Can you share the build failure?
Asking, because it shouldn't be needed and I know there's a warning that can be ignored.
(And because I'd like to keep the list of dependencies small so it remains fast).

elliottt:
```
trevor@trogdor ~/src/tinygo-org/tinygo                                                                                                                                                                                   [0/1917]
% nix develop
warning: Git tree '/home/trevor/src/tinygo-org/tinygo' is dirty
trevor@trogdor:~/src/tinygo-org/tinygo$ make llvm-source
sh: line 1: node: command not found
make: Nothing to be done for 'llvm-source'.
trevor@trogdor:~/src/tinygo-org/tinygo$
```

aykevl:
Yeah, that's not an error. Everything seems to work as intended.

> ```sh: line 1: node: command not found```

You can ignore this. (I started work to remove this warning in https://github.com/tinygo-org/tinygo/pull/3964 but didn't finish it).

> ```make: Nothing to be done for 'llvm-source'.```

This just means the directory llvm-project already exists.

Closing, because I don't think there's anything to fix here.

PR #4283, opened 2024-06-04
compiler/empty basic block

leongross:
To generate a function, tinygo builds a function using llvm and a list of pre-defined basic blocks. Therefore it needs the first basic block for the entry point of a function. If the list of `BasicBlock`s is empty, this indicates that the function is external [1].
Getting the first basic block is then impossible and results in a panic (observed in [2]).

This PR aims to fix this issue and end the compilation gracefully. Since my experience with compiler development is non though, it would be great if the original authors (@aykevl) could take care of the logic behind this and probably make it behave sanely.

[1]: https://cs.opensource.google/go/x/tools/+/refs/tags/v0.21.0:go/ssa/ssa.go;l=338
[2]: https://github.com/tinygo-org/tinygo/actions/runs/9365468429/job/25780736332?pr=4218#step:14:17686

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -1204,7 +1204,14 @@ func (b *builder) createFunctionStart(intrinsic bool) {
 			b.blockExits[block] = llvmBlock
 		}
 		// Normal functions have an entry block.
-		entryBlock = b.blockEntries[b.fn.Blocks[0]]
+		// If Blocks is nil, this indicates an external function for which no Go source code is available.
+		if b.fn.Blocks == nil {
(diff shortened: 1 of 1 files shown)


aykevl (commented):
This is not the issue that you're seeing. What actually happens is that a synthetic function has not yet been created successfully:

https://github.com/tinygo-org/tinygo/blob/6384ecace093df2d0b93915886954abfc4ecfe01/compiler/symbol.go#L213-L222

So basically, TinyGo is trying to define a function while the Go SSA hasn't been fully created yet (in a separate goroutine). This is #4206. You will never hit this condition for regular functions that don't have a function body.

Issue #4285, opened 2024-06-05
Cgo doesn't support bitshifted constants


hectorchu:
If I change the `#define foo 3` to `#define foo (3 << 0)` in `cgo/testdata/const.go`, the test fails with `testdata/const.go:4:17: unexpected token ILLEGAL, expected end of expression`.

aykevl:
Thank you for the report! I have created a PR to fix this: https://github.com/tinygo-org/tinygo/pull/4286

deadprogram:
This was released with `v0.32.0` so now closing. Thank you everyone!

PR #4286, opened 2024-06-06
cgo: implement shift operations in preprocessor macros

aykevl:
This fixes https://github.com/tinygo-org/tinygo/issues/4285.


deadprogram:

PR #4287, opened 2024-06-07
runtime: Simplify slice growing/appending code

lpereira:
Refactor the slice appending function to rely on the slice growing function, and remove branches/loops to use a branchfree variant.

CC @dgryski

### DIFF
--- a/src/reflect/type.go
+++ b/src/reflect/type.go
@@ -774,7 +774,7 @@ func (t *rawType) rawFieldByNameFunc(match func(string) bool) (rawStructField, [
 				if match(name) {
 					found = append(found, result{
 						rawStructFieldFromPointer(descriptor, field.fieldType, data, flagsByte, name, offset),
-						append(ll.index, int(i)),
+						append(ll.index[:len(ll.index):len(ll.index)], int(i)),
 					})
 				}
 
@@ -787,7 +787,7 @@ func (t *rawType) rawFieldByNameFunc(match func(string) bool) (rawStructField, [
 
 					nextlevel = append(nextlevel, fieldWalker{
 						t:     embedded,
-						index: append(ll.index, int(i)),
+						index: append(ll.index[:len(ll.index):len(ll.index)], int(i)),
 					})
 				}
 
--- a/src/runtime/slice.go
+++ b/src/runtime/slice.go
@@ -3,42 +3,26 @@ package runtime
 // This file implements compiler builtins for slices: append() and copy().
 
 import (
+	"math/bits"
 	"unsafe"
 )
 
 // Builtin append(src, elements...) function: append elements to src and return
 // the modified (possibly expanded) slice.
-func sliceAppend(srcBuf, elemsBuf unsafe.Pointer, srcLen, srcCap, elemsLen uintptr, elemSize uintptr) (unsafe.Pointer, uintptr, uintptr) {
-	if elemsLen == 0 {
-		// Nothing to append, return the input slice.
-		return srcBuf, srcLen, srcCap
+func sliceAppend(srcBuf, elemsBuf unsafe.Pointer, srcLen, srcCap, elemsLen, elemSize uintptr) (unsafe.Pointer, uintptr, uintptr) {
+	newLen := srcLen + elemsLen
+	if newLen > srcLen {
+		// Allocate a new slice with capacity for elemsLen more elements, if necessary;
+		// otherwise, reuse the passed slice.
+		srcBuf, _, srcCap = sliceGrow(srcBuf, srcLen, srcCap, newLen, elemSize)
+
+		if elemsLen > 0 {
+			// Append the new elements in-place.
+			memmove(unsafe.Add(srcBuf, srcLen*elemSize), elemsBuf, elemsLen*elemSize)
+		}
 	}
 
-	if srcLen+elemsLen > srcCap {
-		// Slice does not fit, allocate a new buffer that's large enough.
-		srcCap = srcCap * 2
-		if srcCap == 0 { // e.g. zero slice
-			srcCap = 1
-		}
-		for srcLen+elemsLen > srcCap {
-			// This algorithm may be made more memory-efficient: don't multiply
-			// by two but by 1.5 or something. As far as I can see, that's
-			// allowed by the Go language specification (but may be observed by
-			// programs).
-			srcCap *= 2
-		}
-		buf := alloc(srcCap*elemSize, nil)
-
-		// Copy the old slice to the new slice.
-		if srcLen != 0 {
-			memmove(buf, srcBuf, srcLen*elemSize)
-		}
-		srcBuf = buf
-	}
-
-	// The slice fits (after possibly allocating a new one), append it in-place.
-	memmove(unsafe.Add(srcBuf, srcLen*elemSize), elemsBuf, elemsLen*elemSize)
-	return srcBuf, srcLen + elemsLen, srcCap
+	return srcBuf, newLen, srcCap
 }
 
 // Builtin copy(dst, src) function: copy bytes from dst to src.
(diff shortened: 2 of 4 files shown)


dgryski:
Some quick testing shows this makes the multiplication factor 1.125.  However the expression doesn't seem to work for `newCap <= 8` so I think we need to make sure the edge cases are handled properly.

dgryski:
I'm not sure this actually saves us anything.  I threw together a quick simulation and it seems that the new algorithm requires more calls to "grow" (since we're not growing as much, we need to do it more frequently, requiring more CPU) *and* ends up generating more garbage, because the additional calls to grow means we have to throw more stuff away.

You can play with my sample code here: https://go.dev/play/p/QcHiMqnGlym

lpereira:
Yup, that seems to be the case from some of the testing I did last week to understand why some tests were failing. I'm going to send a new version that does pretty much the same thing as before but keeps the cleanup.

lpereira:
Alright, @dgryski, pushed a new version, rebased on the current `dev` branch, that goes back to the "align to the next power of two" code, but using a branchfree version instead.

lpereira:
The failure inside the `reflect` test is very weird.

cee-dub:
The Index field of "want" for the reflect field `F` looks incorrect, it is field 1, but want is `0,0`.

src/runtime/slice.go
```diff
@@ -54,29 +36,21 @@ func sliceCopy(dst, src unsafe.Pointer, dstLen, srcLen uintptr, elemSize uintptr
...
-	// allow nil slice
-	if oldCap == 0 {
-		oldCap++
-	}
-
-	// grow capacity
-	for oldCap < newCap {
-		oldCap *= 2
-	}
+	// This can be made more memory-efficient by multiplying by some other constant, such as 1.5,
```
ldemailly, line 44: It's 2x only for small slices (and then toward 1.25x for big ones) in big go - should it also be same for tinygo or expectation is always "small"-ish in tinygo ?

https://github.com/golang/go/blob/e76353d5a923dbc5e22713267104b56a2c856302/src/runtime/slice.go#L304

src/runtime/slice.go
```diff
@@ -54,29 +36,21 @@ func sliceCopy(dst, src unsafe.Pointer, dstLen, srcLen uintptr, elemSize uintptr
...
+	// which seems to be allowed by the Go language specification (but this can be observed by
+	// programs).
+	newCap = 1 << bits.Len32(uint32(newCap))
```
ldemailly: so tinygo is 32bits only? 

also tinygo doesn't need to account for malloc headers like big go?

src/runtime/slice.go
```diff
@@ -54,29 +36,21 @@ func sliceCopy(dst, src unsafe.Pointer, dstLen, srcLen uintptr, elemSize uintptr
...
 
-	buf := alloc(oldCap*elemSize, nil)
+	buf := alloc(newCap*elemSize, nil)
```
ldemailly, line 50: should this have some overflow check?

ldemailly (commented):
lgtm. nice to share the code in grow. just some (non merge blocking) questions

src/runtime/slice.go
lpereira: Good idea!  Doing 1.5x for slices over 64 elements now.

src/runtime/slice.go
lpereira: Changed this to use `bits.Len()`, passing an `uint` instead of `uint32`.  Hopefully this should work on 64-bit too.

src/runtime/slice.go
lpereira: Yes, it should. Even calculating the new `newCap` can overflow.  However, handling overflow conditions doesn't seem to be something that's done pervasively in TinyGo, so I didn't do it here so this is closer to the rest of the code.

lpereira:
> The Index field of "want" for the reflect field `F` looks incorrect, it is field 1, but want is `0,0`.

Right, and I have no idea why!  I've spent some time in the reflect code, and @dgryski too, but we don't yet know what's happening here.  Our best guesses is that there's either memory corruption somewhere, or that, although the code seems correct, its behavior is different from before, and the reflect code is probably depending on that somehow.

cee-dub:
Hmm, I was suggesting/agreeing with @dgryski that the test may be faulty and the old implementation matched. I think the expectation looks incorrect, though I have very little experience or deep knowledge of this code.

lpereira:
> Hmm, I was suggesting/agreeing with @dgryski that the test may be faulty and the old implementation matched. I think the expectation looks incorrect, though I have very little experience or deep knowledge of this code.

If I understood the test correctly, it seems correct. The index is a slice of indices, starting with the current type, and following any nested type. So an index of `[]int{0}` means the field is the first in that struct; and an index of `[]int{0, 1}` means it's the second of the first of that struct.

The test defines two structs: `Rec1` and `Rec2`; `Rec1` has a pointer to `Rec2`, and `Rec2` has a `F string` and a pointer to `Rec1`. To find `F` from `Rec1`, you need to go through the first element, and then find the first element (`[]int{0, 0}`), which seems to agree with the test.

I might be mistaken about how this works, though.

src/runtime/slice.go
ldemailly: Thanks,

and for the other part (which may not be applicable to tiny go, depends on what kind of heap allocator there is but I think in big go they subtract some meta info size so for instance (making actual numbers up) when requesting 3500 bytes you get a full 4096-headers page of memory instead of 4096+fragment)

ps: I have no idea of tinygo's internal

cee-dub:
Reading more of #tinygo-dev, feel free to ignore me. I didn't look closely enough at the `Index` type in the test.

src/runtime/slice.go
aykevl: I think it is very unlikely to overflow in practice, but yes it would be a good idea to have some sort of check for this.
For 32-bit and 64-bit systems it would be possible to use math/bits for this: https://pkg.go.dev/math/bits#Add32
But maybe there are universal expressions for detecting such an overflow condition efficiently.

src/runtime/slice.go
aykevl: Whether there is metadata or not depends on `-gc=precise` vs `-gc=conservative`. But that should be an implementation detail of the GC/heap: all the slice code needs to know is that there's the `alloc` function and it's not allowed to write outside the returned memory area.

lpereira:
So I copied the failing test to its own file, and `reflect.VisibleFields()` gives the right output, agreeing with Big Go.  So the problem seems to be in the actual test driver somewhere.

lpereira:
Fixed the `reflect` test; let's see if CI likes it.

lpereira:
Alright, `archive/zip` failing is probably the `reflect.DeepEqual()` calls inside the tests, most likely some unbounded recursion or something of the sort.

dgryski:
@lpereira Ok, so it does seem like the bug is somewhere in `reflect` then.  I'm going to try to hunt down the bug looking at that function.

dgryski:
This fixes the `reflect` test:
```
diff --git a/src/reflect/type.go b/src/reflect/type.go
index 1356f67c..32bf68d7 100644
--- a/src/reflect/type.go
+++ b/src/reflect/type.go
@@ -774,7 +774,7 @@ func (t *rawType) rawFieldByNameFunc(match func(string) bool) (rawStructField, [
                                if match(name) {
                                        found = append(found, result{
                                                rawStructFieldFromPointer(descriptor, field.fieldType, data, flagsByte, name, offset),
-                                               append(ll.index, int(i)),
+                                               append(ll.index[:len(ll.index):len(ll.index)], int(i)),
                                        })
                                }

@@ -787,7 +787,7 @@ func (t *rawType) rawFieldByNameFunc(match func(string) bool) (rawStructField, [

                                        nextlevel = append(nextlevel, fieldWalker{
                                                t:     embedded,
-                                               index: append(ll.index, int(i)),
+                                               index: append(ll.index[:len(ll.index):len(ll.index)], int(i)),
                                        })
                                }

```

I'll get this in as another PR and you can rebase against that with only the slice allocation changes.

dgryski:
https://github.com/tinygo-org/tinygo/pull/4367

lpereira:
I rebased this on top of #4387 (please merge it first!), and removed the commit changing `reflect`.  CI is still failing with some error I was unable to track down.

dgryski:
I'll look at the `archive/zip` failure this evening.

dgryski:
The `archive/zip` failure seems to be related to the stepped doubling logic function.

dgryski:
Nope; failure seems intermittent.

dgryski:
This is due to memory fragmentation which (annoyingly) has different behaviour with the new growth algorithm.  A smarter allocator that has free lists for small allocation sizes to reduce fragmentation would help here, but that's outside the scope of this PR.

lpereira:
> A smarter allocator that has free lists for small allocation sizes to reduce fragmentation would help here, but that's outside the scope of this PR.

Agreed.  I've rebased on the current `dev` branch and changed the code to have only one overallocation strategy until we have a better allocator.  This won't reduce the amount of memory used by slices, but it simplifies the code quite a bit, which is already a gain IMHO.

ldemailly:
fragmentation because of not taking headers/meta into account was my point in https://github.com/tinygo-org/tinygo/pull/4287#discussion_r1692116535

lpereira:
sizediff failure is unrelated.

dgryski (approved):
LGTM

dgryski:
Merging.  Will close my PR that has the reflect fix.

src/reflect/type.go
```diff
@@ -774,7 +774,7 @@ func (t *rawType) rawFieldByNameFunc(match func(string) bool) (rawStructField, [
 				if match(name) {
 					found = append(found, result{
 						rawStructFieldFromPointer(descriptor, field.fieldType, data, flagsByte, name, offset),
-						append(ll.index, int(i)),
+						append(ll.index[:len(ll.index):len(ll.index)], int(i)),
```
aykevl, line 777: This will force a new allocation on every `append`, is that intentional?

dgryski: Yes.  This fixes a test failure in `reflect` that was triggered by the new slice allocation code which made this sometimes be a copy and sometimes not.  (The old growth code made this always a copy in the test, but the new growth code expanded it more and so sometimes reused the old slice, which then had entries in the slice overwritten by later loop iterations, causing the faiure.)

PR #4288, opened 2024-06-07
main_test: Diff expected and actual results when tests fail

lpereira:
Uses a vendored "internal/diff" from Big Go to show the differences between the expected and actual outputs when tests fail.


deadprogram:
@aykevl or anyone else have opinion on this PR?

deadprogram:
Seems like a pretty good idea to me. @aykevl or anyone else have any comments here?

dgryski (approved):
LGTM

Issue #4045, opened 2023-12-16
`bytes.Equal` shouldn't allocate


dgryski:
TinyGo uses the `bytes` package from upstream Go.

```
// Equal reports whether a and b
// are the same length and contain the same bytes.
// A nil argument is equivalent to an empty slice.
func Equal(a, b []byte) bool {
        // Neither cmd/compile nor gccgo allocates for these string conversions.
        return string(a) == string(b)
}
```

However, the `[]byte->string` conversion *does* allocate on TinyGo.   We should replace calls to `bytes.Equal` with a call to our own function that doesn't allocate.

aykevl:
I'd like to avoid an optimization like that because it just seems a bit hacky to me. (We have a number of those optimizations in TinyGo, but generally they're to replace a call to a LLVM intrinsic). Instead, I'd propose we add a new transform just like our current `OptimizeStringToBytes` but the other way round.

Here is the transform:

https://github.com/tinygo-org/tinygo/blob/731532cd2b6353b60b443343b51296ec0fafae09/transform/rtcalls.go#L12-L18

The benefit of doing it like that is that it also fixes every other cast that looks like that, instead of just the `bytes.Equal` case.

kevyonan:
shouldn't that do something equivalent like `memcmp`?

dgryski:
@assyrianic Yes.  You can't just immediately defer to `memcmp` though because you need to check for nil slices,  check the lengths, and dereference the data pointers.

aykevl:
> We should replace calls to `bytes.Equal` with a call to our own function that doesn't allocate.

Maybe you're right. I looked at #4289 but realized that optimization is a lot less trivial than the existing string-to-[]bytes optimization. I think it can be replaced with a call to `runtime.stringEqual` (using the `len` part of the byte slice for the string length).

@lpereira

lpereira:
There are quite a bit of `string(b)` conversions in the standard library, too, that would benefit from #4289.  I can add a transform that makes `bytes.Equal()` better by itself, but I'm not sure how beneficial it would be, to be honest.

soypat:
This is still an issue in dev btw

soypat:
@lpereira Is there a PR to merge https://github.com/lpereira/tinygo/commit/71f0663f4debd0db40d76d45f975de7d1e037129 ?

PR #4289, opened 2024-06-08
Optimize `[]byte` to `string` conversions

lpereira:
This enables the optimization of functions such as "bytes.Equal()", which is currently implemented as:

```go
func Equal(a, b []byte) bool { return string(a) == string(b) }
```

Which causes allocations and copies in TinyGo, but not in Big Go.

Since we can prove that neither conversions will escape in this case, we can convert the calls to `runtime.stringFromBytes()` to an inlined version of it that doesn't allocate or copies the original string. 

This is a draft PR, because, although I can verify that the relevant calls to the runtime alloc calls are removed from the final binary in a simple test program, `make test` is failing.  I'd like some help here to understand what's happening.

Another thing that I didn't have time today to check was why `isReadOnly()` was returning `false` for my test program; I'll pick this up on Monday to take a closer look, but for now, I'm omitting the call here, which is most likely why `make test` isn't happy.

Closes: #4045
CC: @aykevl @dgryski

### DIFF
--- a/transform/allocs.go
+++ b/transform/allocs.go
@@ -136,7 +136,7 @@ func valueEscapesAt(value llvm.Value) llvm.Value {
 			panic("expected instruction use")
 		}
 		switch use.InstructionOpcode() {
-		case llvm.GetElementPtr:
+		case llvm.GetElementPtr, llvm.ExtractValue:
 			if at := valueEscapesAt(use); !at.IsNil() {
 				return at
 			}
--- a/transform/optimizer.go
+++ b/transform/optimizer.go
@@ -65,6 +65,7 @@ func Optimize(mod llvm.Module, config *compileopts.Config) []error {
 
 		// Run TinyGo-specific optimization passes.
 		OptimizeStringToBytes(mod)
+		OptimizeBytesToString(mod)
 		OptimizeReflectImplements(mod)
 		maxStackSize := config.MaxStackAlloc()
 		OptimizeAllocs(mod, nil, maxStackSize, nil)
--- a/transform/rtcalls.go
+++ b/transform/rtcalls.go
@@ -73,6 +73,52 @@ func OptimizeStringToBytes(mod llvm.Module) {
 	}
 }
 
+// OptimizeBytesToString transforms runtime.stringFromBytes(...) calls that do not
+// escape into non-alloc versions.  This is useful, for instance, when optimizing
+// "bytes.Equal()", which is implemented as:
+//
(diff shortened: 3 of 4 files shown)


transform/allocs.go
```diff
@@ -163,6 +163,8 @@ func valueEscapesAt(value llvm.Value) llvm.Value {
 		case llvm.ICmp:
 			// Comparing pointers don't let the pointer escape.
 			// This is often a compiler-inserted nil check.
+		case llvm.ExtractValue:
+			// Extracting values don't let the pointer escape.
```
aykevl: There's a bug here: the extracted value itself might escape. This needs a check on the result.

aykevl (commented):
It looks like this PR won't handle a case like this correctly:

```go
b := []byte{...}
s := string(b) // convert to string (without alloc)
b[0] = 5 // this modifies the string
```

Basically it doesn't just need a check whether the byte slice escapes, but also that it is read-only after it is converted to a string.

transform/allocs.go
lpereira: Done.

lpereira:
Rebased and pushed a new version, that also improves the read-only checks to catch more usages.  I still need to fine-comb the table I added for this; however, local tests (`make tinygo-test-wasi`) seem to be working, but I don't have all the dependencies for things other than WASM building, so I'll wait for the CI to see if I need to fix something unexpected.

lpereira:
Rebased to current `dev` and marked it as ready for review.

dgryski:
Ping @aykevl

aykevl:
I did a quick check and this doesn't seem to optimize the `bytes.Equal` case?

aykevl:
This optimization pass should have at least one test that shows that it actually does what it says that it does, and maybe a few tests that show that it doesn't apply in incorrect cases.

I think this is actually pretty tricky to get right. For example, in the following case the optimization would be legal:

```go
func foo(b []byte) {
    s := string(b)
    return s[0]
}
```

but in the following case, it would _not_ be legal:


```go
func foo(b []byte) {
    s := string(b)
    bar() // might modify b
    return s[0]
}
```

lpereira:
(I'm sorry, but I won't be pursuing this optimization anymore as this was done as part of my previous job. If someone wants to pick this up, please feel free to.)

aykevl:
@lpereira thanks for the update!

That's unfortunate, but understandable. I'll close the PR to avoid clutter, but of course it can be reopened if you ever decided to continue work on it.

PR #4290, opened 2024-06-08
transform/rtcalls: Bail fast if can't convert pointer

lpereira:
There's no need to keep looping if one of the uses makes it impossible to convert a call to `runtime.stringToBytes()` with a raw pointer.

### DIFF
--- a/transform/rtcalls.go
+++ b/transform/rtcalls.go
@@ -34,7 +34,7 @@ func OptimizeStringToBytes(mod llvm.Module) {
 			if use.IsAExtractValueInst().IsNil() {
 				// Expected an extractvalue, but this is something else.
 				canConvertPointer = false
-				continue
+				break
 			}
 			switch use.Type().TypeKind() {
 			case llvm.IntegerTypeKind:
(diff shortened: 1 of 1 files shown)


deadprogram:
Thanks for the PR @lpereira 

Question: is it correct to say that if *any* of the uses are invalid, that the string cannot be converted? If so then this looks good to me.

transform/rtcalls.go
```diff
@@ -49,7 +49,7 @@ func OptimizeStringToBytes(mod llvm.Module) {
 					// There is a store to the byte slice. This means that none
 					// of the pointer uses can't be propagated.
 					canConvertPointer = false
-					continue
+					break
```
aykevl, line 52: Note that this actually breaks from the switch statement, not from the for loop. So it only avoids a single `append` statement.

aykevl (approved):
Seems fine to me.

aykevl:
> Question: is it correct to say that if _any_ of the uses are invalid, that the string cannot be converted?

Yes.

PR #4291, opened 2024-06-10
targets: add cyw43439 tag to badger2040-w and also add new pico-w target

deadprogram:
This PR adds to the targets, first by adding the `cyw43439` tag to the `badger2040-w` target, and then also adding a new `pico-w` target for seamless wireless support, initially as part of the tinygo bluetooth package.


sago35 (approved):
LGTM

deadprogram:

PR #4295, opened 2024-06-14
Add smoke tests for machine package

aykevl:
The machine package wasn't tested for every board. Therefore, add a new serial-like test that also tries to import the machine package. This should highlight potential issues in the future.


deadprogram:
Good idea @aykevl thanks for adding this. Now merging.

PR #4302, opened 2024-06-21
gc_leaking: Don't zero out new allocations on Wasm targets

lpereira:
Wasm linear memory is always initialized to zero by definition, so there's no need to waste time zeroing out this allocation.

### DIFF
--- a/src/runtime/gc_leaking.go
+++ b/src/runtime/gc_leaking.go
@@ -44,7 +44,7 @@ func alloc(size uintptr, layout unsafe.Pointer) unsafe.Pointer {
 		runtimePanic("out of memory")
 	}
 	pointer := unsafe.Pointer(addr)
-	memzero(pointer, size)
+	zero_new_alloc(pointer, size)
 	return pointer
 }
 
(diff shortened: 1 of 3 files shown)


aykevl:
For `-gc=leaking` this seems correct.
But in fact, you can do this optimization for every system except for baremetal (and nintendoswitch I think). Linux and Windows zero the memory in `mmap` and `VirtualAlloc`. For MacOS I couldn't find it anywhere online, but I would be very surprised if the returned pages weren't zeroed: every other Unix variant does this.

dgryski:
@aykevl So I guess we can just tweak the build tags here.  LGTM otherwise.

dgryski:
@aykevl Are you find to merge this now and add the other build tags later or do you want to wait until the build tags are done for all the other platforms before merging?

lpereira:
Updated the tags to zero-out memory on anything other than `baremetal || nintendoswitch` tags, and rebased on the current dev branch.

dgryski (approved):
LGTM

PR #4303, opened 2024-06-21
all: use latest version of x/tools

aykevl:
We previously picked a work-in-progress patch, but this is the proper fix for this race condition. I think we should use that instead of relying on the previous work-in-progress patch.

(I didn't verify this still fixes the race condition, but it should, and if not we'll find out soon enough).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	go.bug.st/serial v1.6.0
 	golang.org/x/net v0.26.0
 	golang.org/x/sys v0.21.0
-	golang.org/x/tools v0.21.1-0.20240508182429-e35e4ccd0d2d
+	golang.org/x/tools v0.22.1-0.20240621165957-db513b091504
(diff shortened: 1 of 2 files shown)


deadprogram:
@aykevl looks like something went wrong on this PR https://app.circleci.com/pipelines/github/tinygo-org/tinygo/10982/workflows/26ac7071-bdda-414d-b768-62b6b79b5bf5/jobs/43358?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary&invite=true#step-118-3695_73

aykevl:
Huh, weird. Looks like the race isn't quite fixed yet.

aykevl:
I picked the wrong git hash, it was the commit right before the actual fix.
I've updated the PR and confirmed it still fixes the race.

deadprogram:
Thanks for updating @aykevl now merging.

PR #4305, opened 2024-06-23
Go 1.23 support

aykevl:
Still a work in progress.

PR #4306, opened 2024-06-24
machine: use new internal/binary package

aykevl:
The encoding/binary package in Go 1.23 imports the slices package, which results in circular import when imported from the machine package. Therefore, encoding/binary cannot be used in the machine package.

This commit fixes that by introducing a new internal/binary package that is just plain Go code without dependencies. It can be safely used anywhere (including the runtime if needed).

---

This PR is in preparation of Go 1.23 (see #4305), but it made sense to me to extract it and review separately.


deadprogram:
Seems like an OK solution to me. Thanks @aykevl now merging.

PR #4307, opened 2024-06-24
compiler: remove old atomics workaround for AVR

aykevl:
The bug should have been fixed in any LLVM version that we currently support.

Extracted from #4305 because it's unrelated to Go 1.23 support, just something I found while working on it.

### DIFF
--- a/compiler/atomic.go
+++ b/compiler/atomic.go
@@ -1,9 +1,6 @@
 package compiler
 
 import (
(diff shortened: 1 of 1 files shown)


deadprogram (approved):
:scissors:

deadprogram:
Thanks for keeping things clean @aykevl now merging.

PR #4308, opened 2024-06-24
all: add linux/mipsle support

aykevl:
Very bare bones, doesn't work yet. But it's a start.

Small programs can work, for example `GOARCH=mipsle tinygo run -scheduler=none -gc=leaking ./testdata/alias.go`


aykevl:
This should be working for the most part, including syscalls, goroutines, defer/recover, the FPU, etc.


A few things are still missing:
- ~testdata/math.go crashes. I'm suspecting a LLVM bug but I need to investigate more deeply.~ fixed. Of course this wasn't an LLVM bug but rather a bug in TinyGo (see #4312).
- 64-bit atomics aren't supported yet. These don't seem to be supported by the hardware, I don't know how they're normally implemented?

Marked as draft until #4312 is merged and I have updated the CI files to rebuild LLVM (with Mips support).

aykevl:
Added support for big-endian mips (`GOARCH=mips`) too, though it currently has bugs.

aykevl:
Ready for review!

* `GOARCH=mipsle` works, and is tested in CI. The only thing that doesn't work is 64-bit atomics. (It looks like upstream uses a simple lock mechanism, so that's something we could copy).
* `GOARCH=mips` compiles but has bugs. Small programs tend to work but a few features (like CGo, reflect and maps) have bugs. Probably because this is the first big-endian system that TinyGo tries to support.

deadprogram (approved):
I think this looks pretty great! 

Anyone who is using MIPS want to chime in? Or maybe once we merge this it will be easier to try it out.

deadprogram:
Let's merge and let MIPS people play with it.

PR #4312, opened 2024-06-25
transform: fix incorrect alignment of heap-to-stack transform

aykevl:
It assumed the maximum alignment was equal to `sizeof(void*)`, which is definitely not the case. So this only worked more or less by accident previously.

I found this bug while working on MIPS support, but it is also a bug on many other architectures like amd64 (which has a maximum alignment of 16 bytes, not 8 as would be assumed here).

This can be fixed in a better way, by adding a call-site attribute to the `runtime.alloc` call that stores the expected alignment (or something similar). But at least the current fix should be safe.

(It might actually still be buggy, since amd64 has a maximum alignment of 16 instead of 8. This should be fixed later).

### DIFF
--- a/transform/allocs.go
+++ b/transform/allocs.go
@@ -29,9 +29,13 @@ func OptimizeAllocs(mod llvm.Module, printAllocs *regexp.Regexp, maxStackAlloc u
 
 	targetData := llvm.NewTargetData(mod.DataLayout())
 	defer targetData.Dispose()
-	ptrType := llvm.PointerType(mod.Context().Int8Type(), 0)
-	builder := mod.Context().NewBuilder()
+	ctx := mod.Context()
+	builder := ctx.NewBuilder()
 	defer builder.Dispose()
+
+	// Determine the maximum alignment on this platform.
+	complex128Type := ctx.StructType([]llvm.Type{ctx.DoubleType(), ctx.DoubleType()}, false)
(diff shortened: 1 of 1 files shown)


transform/allocs.go
```diff
@@ -29,10 +29,14 @@ func OptimizeAllocs(mod llvm.Module, printAllocs *regexp.Regexp, maxStackAlloc u
...
 	defer targetData.Dispose()
-	ptrType := llvm.PointerType(mod.Context().Int8Type(), 0)
-	builder := mod.Context().NewBuilder()
+	ctx := mod.Context()
+	builder := ctx.NewBuilder()
 	defer builder.Dispose()
 
+	// Determine the maximum alignment on this platform.
+	complex128Type := ctx.StructType([]llvm.Type{ctx.DoubleType(), ctx.DoubleType()}, false)
+	maxAlign := int64(targetData.ABITypeAlignment(complex128Type))
```
dgryski, line 38: This seems like the sort of thing we should compute once and then store in some target config somewhere.

dgryski (approved):
LGTM

aykevl:
I have written a better version of this PR: https://github.com/tinygo-org/tinygo/pull/4317
It's similar, but actually fixes the same issue on all architectures (including amd64 where we just haven't seen it happen yet).

Issue #4314, opened 2024-06-27
build with target "wasm-unknown" fail in windows10 22H2


longkui-clown:
when I use cmd `tinygo build -o demo_go.wasm -target wasm-unknown -no-debug main.go` to build wasm file, it return errors like below:
```
# os
C:\tinygo0.32.0.windows-amd64\tinygo\src\os\types_unix.go:20:18: undefined: syscall.Stat_t
C:\tinygo0.32.0.windows-amd64\tinygo\src\os\stat_linuxlike.go:50:32: undefined: syscall.Timespec
C:\tinygo0.32.0.windows-amd64\tinygo\src\os\dir_unix.go:55:28: undefined: syscall.ReadDirent
C:\tinygo0.32.0.windows-amd64\tinygo\src\os\dir_unix.go:82:49: undefined: syscall.Dirent
C:\tinygo0.32.0.windows-amd64\tinygo\src\os\dirent_linux.go:15:46: undefined: syscall.Dirent
C:\tinygo0.32.0.windows-amd64\tinygo\src\os\dirent_linux.go:15:83: undefined: syscall.Dirent
C:\tinygo0.32.0.windows-amd64\tinygo\src\os\dirent_linux.go:19:46: undefined: syscall.Dirent
C:\tinygo0.32.0.windows-amd64\tinygo\src\os\dirent_linux.go:19:86: undefined: syscall.Dirent
C:\tinygo0.32.0.windows-amd64\tinygo\src\os\dirent_linux.go:27:49: undefined: syscall.Dirent
C:\tinygo0.32.0.windows-amd64\tinygo\src\os\dirent_linux.go:31:33: undefined: syscall.Dirent
C:\tinygo0.32.0.windows-amd64\tinygo\src\os\dirent_linux.go:37:15: undefined: syscall.DT_BLK
C:\tinygo0.32.0.windows-amd64\tinygo\src\os\dirent_linux.go:39:15: undefined: syscall.DT_CHR
C:\tinygo0.32.0.windows-amd64\tinygo\src\os\dirent_linux.go:41:15: undefined: syscall.DT_DIR
C:\tinygo0.32.0.windows-amd64\tinygo\src\os\dirent_linux.go:43:15: undefined: syscall.DT_FIFO
C:\tinygo0.32.0.windows-amd64\tinygo\src\os\dirent_linux.go:45:15: undefined: syscall.DT_LNK
[...]

PR #4315, opened 2024-06-27
wasm-unknown: make sure the os package can be imported

aykevl:
See: https://github.com/tinygo-org/tinygo/issues/4314

The os package isn't particularly useful on wasm-unknown, but just like on baremetal systems it's imported by a lot of packages so it should at least be possible to import this package.

### DIFF
--- a/src/examples/hello-wasm-unknown/main.go
+++ b/src/examples/hello-wasm-unknown/main.go
@@ -3,6 +3,10 @@
 // tinygo build -size short -o hello-unknown.wasm -target wasm-unknown -gc=leaking -no-debug ./src/examples/hello-wasm-unknown/
 package main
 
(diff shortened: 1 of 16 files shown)


aykevl:
I think we should introduce a new build tag, something like `nonhosted` (or `nonos`, or `noos`, or whatever, feel free to bikeshed) that indicates that the program is compiled for a non-OS-like environment. OSes are the obvious ones (Windows, MacOS, Linux), but also WASI because it provides typical OS services like files and command line arguments. Things that are _not_ OSes are baremetal, wasm_unknown, browsers (GOOS=js), and anything else that isn't a proper OS.

This however is a bigger change that should be done in a separate PR.

dgryski (approved):
LGTM

PR #4316, opened 2024-06-27
test: support GOOS/GOARCH pairs in the -target flag

aykevl:
This means it's possible to test just a particular OS/arch with a command like this:

    go test -run=Build -target=linux/arm

I found it useful while working on MIPS support.


deadprogram:
Very simple and good change. Thanks @aykevl now merging.

PR #4317, opened 2024-06-27
transform: fix incorrect alignment of heap-to-stack transform (2nd try)

aykevl:
This is an improved version of #4312 which doesn't use heuristics anymore but rather relies on frontend (./compiler directory) to add the right alignment attribute.

I needed a small addition to go-llvm for this to work: https://github.com/tinygo-org/go-llvm/commit/3b50c76783a8b502623ef40530728b2c746b34fc


deadprogram:
@dgryski any thoughts on this PR?

dgryski:
Not sure I understand the llvm-specific fix but I'll believe that it's a bug and that this is a fix for it.

deadprogram:
Thanks @dgryski that was my own assessment more or less. :smile_cat: 

Now merging.

PR #4321, opened 2024-06-30
Support for the CH32V003 chip

aykevl:
Work in progress.

To try, you need to make the lib/cmsis-svd/data/WCH-Community/ directory and download https://github.com/ch32-rs/ch32-rs/blob/main/svd/vendor/CH32V003xx.svd into that directory.

Also, you need to install probe-rs.

Not nearly finished. I'd like to get at least the timer working before marking this PR as ready.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,8 @@ src/device/kendryte/*.go
 src/device/kendryte/*.s
 src/device/rp/*.go
 src/device/rp/*.s
+src/device/wch/*.go
+src/device/wch/*.s
 vendor
 llvm-build
 llvm-project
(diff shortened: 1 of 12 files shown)


aykevl:
We might need to add a new way of generating register files: https://github.com/ch32-rs/ch32-rs/issues/21
Basically it seems like development on register files now happens here: https://github.com/ch32-rs/ch32-data, but unfortunately they store the files in a custom YAML based format (not in standard SVD format).

thomas-maurice:
Is there any progress on this ? It would be super cool to have these chips supported 👀

jwetzell:
Looking at maybe reviving this effort. I've got a few nanoCH32V003 boards coming in.

PR #4322, opened 2024-07-02
add signal stubs to runtime

leongross:
Linux applications might ignore a certain set of received signals.
For that the `tinygo` runtime has to expose said signals.

Since the `tinygo` runtime only has limited linux signal support, this PR adds stubs to the `runtime` and the `os` package.

Package to build with this PR: https://github.com/u-root/u-root/tree/main/cmds/exp/vmboot

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -338,6 +338,7 @@ TEST_PACKAGES_FAST = \
 	net/http/internal/ascii \
 	net/mail \
 	os \
+	os/signal \
 	path \
 	reflect \
 	sync \
--- a/loader/goroot.go
+++ b/loader/goroot.go
@@ -247,12 +247,12 @@ func pathsToOverride(goMinor int, needsSyscallPackage bool) map[string]bool {
 		"net/":                  true,
 		"net/http/":             false,
 		"os/":                   true,
-		// "os/signal":             false,
-		"os/user/": false,
-		"reflect/": false,
-		"runtime/": false,
-		"sync/":    true,
-		"testing/": true,
+		"os/signal":             false,
+		"os/user/":              false,
+		"reflect/":              false,
+		"runtime/":              false,
+		"sync/":                 true,
+		"testing/":              true,
 	}
 
 	if goMinor >= 19 {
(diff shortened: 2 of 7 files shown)


leongross:
The waspi2 test seem to fail due to

> /opt/hostedtoolcache/go/1.22.5/x64/src/internal/testpty/pty_cgo.go:11:10: fatal: 'fcntl.h' file not found

Building for non-wasm seems to work as expected (testes on x86_64 linux)
@deadprogram @aykevl does someone have an idea why that occurs / how to fix that?

deadprogram:
@leongross can you please rebase this PR against the latest `dev` to see if it still fails CI?

leongross:
Running the a local build results in symbols not being found. Also it looks like the wrong single library is used?
```
ld.lld: error: undefined symbol: os/signal.signal_recv
>>> referenced by signal_unix.go:23 (/usr/lib/golang/src/os/signal/signal_unix.go:23)
>>>               /tmp/tinygo2823474359/main.lto.main.o:(os/signal.loop)

ld.lld: error: undefined symbol: os/signal.signal_enable
>>> referenced by signal_unix.go:49 (/usr/lib/golang/src/os/signal/signal_unix.go:49)
>>>               /tmp/tinygo2823474359/main.lto.main.o:(os/signal.Notify)
```

Searching for the strings in the binary yields the linking symbols
```
$ nm build/tinygo | grep "os/signal.signal_recv"
0000000000585d20 t os/signal.signal_recv
```

The lubmusl error can be traced down to [this](https://git.musl-libc.org/cgit/musl/tree/include/fcntl.h#n22) include in muslbic that seems to miss `fcnt.h` in [x86_64](https://git.musl-libc.org/cgit/musl/tree/arch/x86_64/bits).

leongross:
@dgryski any ideas why this might fail? I saw you worked on the runtime before.

dgryski:
The `os/signal` entry in `goroot.go` is still commented out.  I think that needs to be set to `false` since we have our own complete implementation of that package now.  Without that it's trying to run the Big Go `os/signal` tests which fail because our syscall packages are different.

aykevl:
@leongross I looked at how difficult it would be to support signals (instead of just stubbing them out), and got it working on Linux: https://github.com/tinygo-org/tinygo/pull/4378
What do you think?

(Note: if you try this, make sure to run `tinygo clean` so that musl is rebuilt. Otherwise signals will just result in a hang or crash.)

leongross:
This PR is superseded by https://github.com/tinygo-org/tinygo/pull/4378.

PR #4323, opened 2024-07-02
add os.StartProcess

leongross:
This PR adds a working implementation for `os.StartProcess`, just a concatenation of the `fork` and `exec` syscalls.
The upstream Golang implementation has a lot more safeguards with mutices etc but to keep this feasible I think this will do as well.

I would really appreciate some feedback on this to keep this code as bug-free as possible :)


leongross:
@deadprogram @ayang64

src/os/exec_posix.go
```diff
@@ -33,3 +34,55 @@ func (p *Process) release() error {
 	runtime.SetFinalizer(p, nil)
 	return nil
 }
+
+// Combination of fork and exec, careful to be thread safe.
+
+// https://cs.opensource.google/go/go/+/master:src/syscall/exec_unix.go;l=143?q=forkExec&ss=go%2Fgo
```
ayang64: can you comment this in a way that describes the compromises you made instead of simply calling it "hacky"?

why is it hacky?

src/os/exec_posix.go
```diff
@@ -33,3 +34,55 @@ func (p *Process) release() error {
...
+		if ret != 0 {
+			// exec failed
+			syscall.Exit(1)
+		}
+		// 3. TODO: use pipes to communicate back child status
+		return int(ret), nil
+	}
+}
+
+// in regular go this is where the forkExec thingy comes in play
```
ayang64: is there a better way to comment this function in more detail than where the "thingy comes in play"?

ayang64 (changes requested):
In general I think this is good but I'd like more context and better comments before we give this a detailed review.

Can you reword some of your comments and maybe add additional commentary to describe some of the compromises you had to make without having to read the code in detail?

Not that I don't want to read the code but comments would help me understand some of the considerations you're talking about.

src/syscall/syscall.go
```diff
@@ -3,6 +3,9 @@ package syscall
 import (
 	"errors"
 	"sync/atomic"
+	"internal/bytealg"
```
ayang64: this is a weird double import of `bytealg`.

src/os/exec_posix.go
leongross: This seems to be a working comment of mine that slipped into the commit. I will improve this comment, thank you.

leongross:
> In general I think this is good but I'd like more context and better comments before we give this a detailed review.
> 
> Can you reword some of your comments and maybe add additional commentary to describe some of the compromises you had to make without having to read the code in detail?
> 
> Not that I don't want to read the code but comments would help me understand some of the considerations you're talking about.

This is very reasonable, will do that!

leongross:
@ayang64 @aykevl @deadprogram @dgryski

deadprogram:
@leongross please rebase this branch and resolve merge conflicts. Thank you!

leongross:
The CI errors are difficult for me to explain. When I run the test locally I get the following output:
```
$ make
$ ./build/tinygo test -v -run TestForkExec os
=== RUN   TestForkExec
    forkExec succeeded: new process has pid &{255262}
--- PASS: TestForkExec (0.00s)
PASS
ok      os      0.001s
```

deadprogram:
Looks to me like the problem is something to do with needing some additional build tags. The CI builds that fail are Windows and macOS. For example:

https://github.com/tinygo-org/tinygo/actions/runs/10061289668/job/27810982068?pr=4323#step:17:32

leongross:
Internal test dependencies seem to be broken now, although this PR does not change something related to `os.Interrupt`.

```
main_test.go:572: test error: could not compile: /Users/runner/hostedtoolcache/go/1.22.5/arm64/src/testing/internal/testdeps/deps.go:149:63: undefined: os.Interrupt
```

src/os/exec_posix_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+	proc, err := StartProcess("echo", []string{"echo", "hello", "world"}, &ProcAttr{})
+	if err != nil {
+		t.Fatalf("forkExec failed: %v", err)
+		return
+	}
+
+	if proc.Pid == 0 {
+		t.Fatalf("forkExec failed: new process has pid 0")
+	} else {
```
rminnich: Don't need the else here, since Fatal does not return. This as well to the return after the fatal, above.

src/syscall/syscall_unix.go
```diff
@@ -16,3 +21,54 @@ type SockaddrInet6 struct {
 	Addr   [16]byte
 	raw    RawSockaddrInet6
 }
+
+func Fork() (err error) {
+	fail := int(libc_fork())
+	if fail < 0 {
+		// TODO: parse the syscall return codes
+		return errors.New("fork failed")
```
rminnich, line 29: if you can wrap some standard error in here, tests are easier.
syscall.EAGAIN is at least as good as this errors.New I think?

src/syscall/syscall_unix.go
```diff
@@ -16,3 +21,54 @@ type SockaddrInet6 struct {
...
+	env1 := make([]*byte, len(envv)+1)
+	for i, env := range envv {
+		env1[i] = &cstring(env)[0]
+	}
+	env1[len(envv)] = nil
+
+	fail := int(libc_execve(&argv0[0], &argv1[0], &env1[0]))
+	if fail < 0 {
+		// TODO: parse the syscall return codes
+		return errors.New("fork failed")
```
rminnich, line 54: same here. fail can even be
fail := syscall.Errno(libc_execve(&argv0[0], &argv1[0], &env1[0]))

leongross:
I suspect that the import errors originate in this syscall package requirement code:
https://github.com/tinygo-org/tinygo/blob/release/loader/goroot.go#L219
The golang standard library does not provide implementations for Fork and Execve and it looks as if the custom `syscall` package for `GOOS=linux` is not loaded but the standard one which then makes sense.
A merge that just adds Frok and Execve to the syscall package could be the solution here.|
Alternatively, we could put Execve and Fork into another package such as `exec`  or `os`.

aykevl:
Closing since #4377 was merged.

PR #4217, opened 2024-04-04
os/user.Lookup

leongross:
Add support for User lookup via `os/user.Lookup()`

PR #4324, opened 2024-07-02
os/user: use stdlib version of this package

aykevl:
I tried implementing enough CGo support to get the native os/user package to work. But I hit a few bugs, probably in CGo itself. Then I realized I could just as well set the osusergo build tag to disable CGo for this specific case.

This actually gets the os/user package to work correctly on Linux (I confirmed it returns my name/uid/homedir etc). On other systems, it probably just returns an error if it can't determine these kinds of things. But that's no worse than the current behavior which just doesn't do anything at all.

CC @eliasnaur


aykevl:
I missed #4217, see https://github.com/tinygo-org/tinygo/pull/4217#issuecomment-2203531206 for details.

eliasnaur:
Successfully tested with my programs. Thank you!

deadprogram:
@aykevl looks like actual problem on this PR see https://github.com/tinygo-org/tinygo/actions/runs/9763774599/job/27608561750#step:9:40

leongross:
```
2: wasm trap: wasm `unreachable` instruction executed
```

There might be a check missing for an argument not to be nil

aykevl:
The problem was that os/user was also tested on WebAssembly, but because `t.Skipf` isn't implemented yet it resulted in a failure. I've modified the PR to only run the tests on Linux/MacOS/Windows instead.

Probably we should just stop the test when `t.Skip()` is called, instead of continuing which will inevitably result in weird behavior. But that's a separate issue.

aykevl:
Reminder, this PR is ready for review! See https://github.com/tinygo-org/tinygo/pull/4410#issuecomment-2289975625

archie2x (approved):
Nice. Sorry I didn't see this (or @leongross's) before submitting my own!

dgryski (approved):
LGTM

deadprogram:
Thanks for making this change @aykevl and to everyone who helped review. Now merging.

PR #4325, opened 2024-07-03
runtime: implement dummy getAuxv to satisfy golang.org/x/sys/cpu

eliasnaur:
Fixes the program

```go
  package main

  import _ "golang.org/x/sys/cpu"

  func main() {
  }
```


dgryski (approved):
LGTM

PR #4327, opened 2024-07-08
Add chromeos bind mount support for flash command

taigrr:
ChromeOS uses the Plan 9 9p protocol for sharing files, drives, etc. The code in `findFATMounts` splits on the mount type using `strings.Fields`. Therefore, the vfat mounts actual mount type is obscured, and the /proc/mount file shows 9p in place of vfat.
image

I am marking this PR as a draft, as the function is named `findFATMounts` and likely should be renamed, or the logic may need to be moved elsewhere, as this new (one-line) fix will allow ext3/ext4 etc. filesystem types to leak through if they are mounted through the 9p protocol. I don't think this is a huge deal, but is less than ideal.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1072,7 +1072,8 @@ func findFATMounts(options *compileopts.Options) ([]mountPoint, error) {
 				continue
 			}
 			fstype := fields[2]
-			if fstype != "vfat" {
+			// chromeos bind mounts use 9p
+			if !(fstype == "vfat" || fstype == "9p") {
 				continue
 			}
 			fspath := strings.ReplaceAll(fields[1], "\\040", " ")


taigrr:
hey @deadprogram I will chat with you in person on this tomorrow, flagging it so you have a point of reference

taigrr:
After discussing with @deadprogram in person I'm marking this as ready for merging.

skirmish:
I also independently verified this afternoon it works on the showroom floor. Thanks @taigrr for walking me through it.

taigrr:
Once this gets approved I'll send my PR for updating docs to show people how to use it since the process is a bit more complicated on crostini

main.go
```diff
@@ -1072,7 +1072,8 @@ func findFATMounts(options *compileopts.Options) ([]mountPoint, error) {
 				continue
 			}
 			fstype := fields[2]
-			if fstype != "vfat" {
+			// chromeos bind mounts use 9p
+			if fstype != "vfat" && fstype != "9p" {
```
deadprogram:
```suggestion
			if !(fstype == "vfat" || fstype == "9p") {
```

Nitpick: perhaps more maintainable to stick to demorgan's law?

taigrr: Sure. Do you prefer a rebase + force push or a second commit?

deadprogram: rebase + force push, please

taigrr: done

taigrr:
sorry for the second force push. editor reordered some import statements to be alphabetical that weren't in the first commit and it slipped by me at first.

deadprogram:
Thanks for the improvement @taigrr and to @skirmish for testing. Now merging.

PR #4329, opened 2024-07-09
fix: remove message after test binary built

dshemin:
Hello!

After test binary build i've see message about test fails and it's a bit confusing. Here is a fix for this.


aykevl (approved):
Thank you! This is something I hit myself sometimes but never got around to fixing.

PR #4330, opened 2024-07-09
Refactor compiler error messages and add tests

aykevl:
This refactors the compiler error messages in various ways. It should not (significantly) change the behavior, but it will make it easier to change things later on with confidence that nothing breaks.

There are a few things that I'd like to implement some day, maybe:

* Provide JSON output for error messages, so that tools (in particular the playground) can show better error messages.
* Provide nicely formatted error messages, like you can see in GCC, Clang and Rust.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1288,10 +1288,9 @@ func usage(command string) {
 
 // try to make the path relative to the current working directory. If any error
 // occurs, this error is ignored and the absolute path is returned instead.
-func tryToMakePathRelative(dir string) string {
-	wd, err := os.Getwd()
-	if err != nil {
-		return dir
+func tryToMakePathRelative(dir, wd string) string {
+	if wd == "" {
+		return dir // working directory not found
 	}
 	relpath, err := filepath.Rel(wd, dir)
 	if err != nil {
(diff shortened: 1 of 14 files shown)


dgryski (approved):
LGTM

aykevl:
@dgryski I've fixed the Windows error (I hope), renamed some tests, and made sure that when the package loader (`go list`) returns multiple errors, these are all shown and not just the first.

aykevl:
Fixed a Go 1.19 regression in my previous change.

aykevl:
@dgryski can you take another look at this PR? If you're okay with it, I'll merge it.
I have a few things that depend on this PR.

main.go
```diff
@@ -1339,19 +1335,35 @@ func printCompilerError(logln func(...interface{}), err error) {
...
 		for _, err := range err.Errs {
-			printCompilerError(logln, err)
+			printCompilerError(err, logln, wd)
 		}
 	case loader.Error:
-		logln(err.Err.Error())
-		logln("package", err.ImportStack[0])
-		for _, pkgPath := range err.ImportStack[1:] {
-			logln("\timports", pkgPath)
+		if err.Err.Pos.Filename != "" {
```
dgryski, line 1347: Is there a way we can know for certain which of these cases we're in ("probably a syntax error" / "probably an import cycle error") without guessing?  What happens if we guess wrong?  Where do other error messages end up?

dgryski (approved):
One question, but otherwise I'm happy for improved messages and tests.

main.go
aykevl: These errors come from the `go list` command.
An import cycle looks like this:

```json
        "Error": {
                "ImportStack": [
                        "command-line-arguments",
                        "github.com/tinygo-org/tinygo/testdata/errors/importcycle",
                        "github.com/tinygo-org/tinygo/testdata/errors/importcycle"
                ],
                "Pos": "",
                "Err": "import cycle not allowed"
        },
```

A syntax error looks like this:

```json
        "Error": {
                "ImportStack": [
                        "command-line-arguments",
                        "github.com/tinygo-org/tinygo/testdata/errors/invaliddep"
                ],
                "Pos": "testdata/errors/invaliddep/invaliddep.go:1:1",
                "Err": "expected 'package', found ppackage"
        }
```
The only difference I see is the `Pos` field (and the error message itself).



>  What happens if we guess wrong?

We may get a differently formatted error message.
This is why I added all these tests, to ensure that the error messages are actually what they should be. And that's how I discovered that the previous code had an error message that wasn't formatted correctly.

>  Where do other error messages end up?

Most other error messages are stored as `scanner.Error` or `types.Error`, which are always for a given source location.

Eventually I'd like to flip this logic around and let packages return the right error messages. That's one reason I introduced the separate `diagnostics` package in https://github.com/tinygo-org/tinygo/pull/4335. Eventually (with another refactor) it enables the packages that produce an error to return the right diagnostics which avoids all this guessing.

Issue #4331, opened 2024-07-09
RP2040 I2C Target mode doubles first byte sent in Reply


sparques:
I'm using a Raspberry Pi Pico as an I2C Target. The host is a Raspberry Pi Zero 2W.

After one transfer, the first byte is always doubled up.

On the Host (Pi Zero):

```
/home/sparques # i2cget -y 1 0x2c 0x20 i 4 # read from i2c bus 1; device id 0x2c; register 0x20, block of data 4 bytes long
0x1e 0x1e 0x00 0x00
```

I have the Pico dumping output on USB serial. It shows the below, which is correct.

```
host request against 0x20
Sending to host: [ 1e 0 0 1 18 0 1 0 2 0 b 0 3 0 0 0 4 0 5 0 ad de ef be 1 0 0 0 0 0] 
```

And here is a trace with a logic analyzer. The pico sends the first byte; pauses, then sends the first byte again along with the rest.

![image](https://github.com/tinygo-org/tinygo/assets/15929821/c1d83a9b-d570-4128-811a-0ce2db0578a5)

Any idea's what's going on here? I based my code on https://github.com/tinygo-org/tinygo/tree/release/src/examples/i2c-target

It seems like something's wrong with the Pico's I2C peripheral (or how it's configued, or how it's called, etc), or the code implementing Reply(). 

I have an http-like router for handling I2C r/w requests to registers. I've tried to pare down the code to just what's relevant. This seems to work fine but 🤷🏽 if it's related to the issue I'm seeing.

```
type Router struct {
	bus      I2C
	addr     uint16
	register uint16
	maxTx    int

	regMap map[uint16]Handler

	Finish func()
}

// called in main(); HIDDescRegister = 0x20; 
router.Handle(HIDDescRegister, func([]byte) (rx []byte) {
	return HIDDesc // this is just a []byte = 1e 0 0 1 18 0 1 0 2 0 b 0 3 0 0 0 4 0 5 0 ad de ef be 1 0 0 0 0 0
})

// the meat of Router
func (r *Router) Handle(register uint16, f Handler) {
	r.regMap[register] = f
	if f == nil {
		delete(r.regMap, register)
	}
}

func (r *Router) route(register uint16, buf []byte) {
	handler, ok := r.regMap[register]
	if !ok {
		// HID spec says if a request is made to a non-supported register, we should reply with zero-byte response
		// not sure if I understood this part of the spec right or if this is good for other I2C protocols
[...]

PR #4334, opened 2024-07-11
Rewrite Reply() to fix sending long replies in I2C Target Mode

sparques:
This addresses https://github.com/tinygo-org/tinygo/issues/4331.

I think the changes that actually did something were  using the RAW rather than MASK registers and adding a wait for the TX_FIFO to clear. 

In my debugging, I had the original loop print the value of txPtr and it printed zero twice--given the txPtr++ increment comes right after the call to put a byte into the FIFO, I'm really at a loss as to how that's possible. Something wrong with the task-pause-and-resume? A scheduler bug? Removing gosched() from the loop had no effect on the double-send-first-byte issue.

At any rate, this code now works per a logic analyzer as well as finally getting my HID-over-I2C keypad working.

### DIFF
--- a/src/machine/machine_rp2040_i2c.go
+++ b/src/machine/machine_rp2040_i2c.go
@@ -501,15 +501,23 @@ func (i2c *I2C) Reply(buf []byte) error {
 	}
 
 	for txPtr < len(buf) {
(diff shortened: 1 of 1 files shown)


deadprogram (approved):
Thanks for the fix @sparques

PR #4335, opened 2024-07-12
Refactor error message printing and sort them

aykevl:
This is a big refactor of error messages (enabled by #4330) that should enable a number of future improvements to error messages. Because it's a refactor, there's a decent chance it will break some existing error messages. That's why I added a load of tests for these error messages: the hope is that with all those tests, no major issues will arrive.

And as a bonus to show why this refactor is needed, I've made a small change that will sort error messages. This is something that's been bugging me for a while but was difficult to fix in a nice way with the previous design.

~This is a draft because it depends on #4330.~

### DIFF
--- a/errors_test.go
+++ b/errors_test.go
@@ -23,6 +23,7 @@ func TestErrors(t *testing.T) {
 		"loader-invaliddep",
 		"loader-invalidpackage",
 		"loader-nopackage",
+		"optimizer",
 		"syntax",
 		"types",
 	} {
--- a/errors_test.go
+++ b/errors_test.go
@@ -17,6 +17,7 @@ import (
 func TestErrors(t *testing.T) {
 	for _, name := range []string{
 		"cgo",
+		"compiler",
 		"interp",
 		"loader-importcycle",
 		"loader-invaliddep",
--- a/errors_test.go
+++ b/errors_test.go
@@ -17,6 +17,7 @@ import (
 func TestErrors(t *testing.T) {
 	for _, name := range []string{
 		"cgo",
+		"interp",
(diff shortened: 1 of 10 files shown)


aykevl:

deadprogram:
Pinging @dgryski @ydnar and any other interested parties...

errors_test.go
```diff
@@ -56,26 +59,44 @@ func testErrorMessages(t *testing.T, filename string) {
...
-		text = strings.ReplaceAll(text, "\\", "/")
+func matchErrors(t *testing.T, pattern, actual string) bool {
+	patternLines := strings.Split(pattern, "\n")
+	actualLines := strings.Split(actual, "\n")
+	if len(patternLines) != len(actualLines) {
+		return false
 	}
-	return text
+	for i, patternLine := range patternLines {
+		indices := regexp.MustCompile(`\{\{.*?\}\}`).FindAllStringIndex(patternLine, -1)
```
dgryski, line 78: Lift the regexp compilation out of the loop?

dgryski (approved):
LGTM

errors_test.go
deadprogram: @aykevl want to make this change, or just merge as it is?

deadprogram: If actually needed this can be done in a future PR.

deadprogram:
Now merging, thanks @aykevl for adding this and to @dgryski for review.

errors_test.go
aykevl: I can lift it out, but this isn't very performance critical. I don't think there's a real need (apart from aesthetics).

PR #4336, opened 2024-07-12
cgo: support preprocessor macros passed on the command line

aykevl:
Go code might sometimes want to use preprocessor macros that were passed on the command line (such as `-DMACRO_NAME=3`). This wasn't working before and resulted in the following error:

    internal error: could not find file where macro is defined

This is now supported, though location information isn't available (which makes sense: the command line is not a file).

I had to use the `clang_tokenize` API for this and reconstruct the original source code. Apparently this is the only way to do it: https://stackoverflow.com/a/19074846/559350
In the future we could consider replacing our own tokenization with the tokenizer that's built into Clang directly. This should reduce the possibility of bugs a bit.

### DIFF
--- a/cgo/cgo_test.go
+++ b/cgo/cgo_test.go
@@ -7,6 +7,7 @@ import (
 	"go/ast"
 	"go/format"
 	"go/parser"
+	"go/scanner"
 	"go/token"
 	"go/types"
 	"os"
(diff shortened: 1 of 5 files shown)


dgryski (approved):
LGTM

aykevl:
Thank you for the review!

PR #4337, opened 2024-07-12
builder: remove workaround for generics race condition

aykevl:
This commit reverts commit 13a8eae0d.

It appars that the race has been fixed by https://github.com/golang/tools/commit/db513b091504, which we now use because it fixes another race condition as well.
See: https://github.com/tinygo-org/tinygo/issues/4206 for the other race condition.


dkegel-fastly (approved):
LGTM

PR #4340, opened 2024-07-15
esp32-dev board support

Fluffy-Bean:
Hello!

I want to add support for the `esp32-dev` board, these are the following references I've used:

**Reference 1**
Pinout for the espressif esp32 devboard
https://docs.espressif.com/projects/esp-idf/en/stable/esp32/_images/esp32-devkitC-v4-pinout.png

**Reference 2**
This is the specific board I have, but it's silkscreen labels are slightly different which I hope isn't going to be an issue
https://ae01.alicdn.com/kf/Sa74c2ababf3640c5b77022481509c13bi.jpg

I was able to successfully compile tinygo using workflows, and run the following program using the ESP32 I2C support added in the last release

```go
package main

import (
	"image/color"

	"machine"

	"tinygo.org/x/drivers/ssd1306"
	"tinygo.org/x/tinyfont"
	"tinygo.org/x/tinyfont/proggy"
)

func main() {
	White := color.RGBA{R: 255, G: 255, B: 255, A: 255}

	machine.I2C0.Configure(machine.I2CConfig{
		SCL: machine.SCL_PIN,
		SDA: machine.SDA_PIN,
	})

	display := ssd1306.NewI2C(machine.I2C0)
	display.Configure(ssd1306.Config{
		Address:  ssd1306.Address_128_32,
		VccState: ssd1306.SWITCHCAPVCC,
		Width:    128,
		Height:   64,
	})

	display.ClearBuffer()
	display.ClearDisplay()

	tinyfont.WriteLine(&display, &proggy.TinySZ8pt7b, 5, 5, "Hello World!", White)

	display.Display()
}

```

Some pins are commented out, as I'm unsure if they're available, or if I got them correct.

### DIFF
--- a/src/machine/board_esp32_dev.go
+++ b/src/machine/board_esp32_dev.go
@@ -1,0 +1,82 @@
+//go:build esp32_dev
+
+package machine
+
+// Reference 1:
+// https://docs.espressif.com/projects/esp-idf/en/stable/esp32/_images/esp32-devkitC-v4-pinout.png
+
+// Reference 2:
+// This is the board I have, but it's slightly different, so for now I'm going to avoid pins that
+// are missing/unlabeled
+// https://ae01.alicdn.com/kf/Sa74c2ababf3640c5b77022481509c13bi.jpg
+
+// Silkscreen Pins
+const (
+	VP  = GPIO36
+	VN  = GPIO39
+	CMD = GPIO11
+	TX  = GPIO1
+	RX  = GPIO3
+	CLK = GPIO6
+)
+
+// Analog-to-Digital
+const (
+	// ADC(Bus)_(Channel)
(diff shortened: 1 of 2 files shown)


deadprogram:
Hello @Fluffy-Bean 

I am just wondering if the board actually is the same as the https://github.com/tinygo-org/tinygo/blob/release/src/machine/board_esp32-coreboard-v2.go

A superficial review of a few pins led me to thing this might be the case.

What do you think?

Fluffy-Bean:
That does indeed seem to be the same board based on the pins! An unchanged program of mine and the example I provided seem to compile correctly when switching the compile target to the `esp32-coreboard-v2`.

As there don't seem to be issues, I'll close this pull request. Thank you for your time!

PR #4343, opened 2024-07-16
Allows compilation of code using debug.BuildInfo and show correct tinygo version

ldemailly:
That let's me build a 245k tinygo target https://github.com/grol-io/grol/pull/37
(also needed some change in fortio.org/log)

Many thanks to @ydnar and @dankegel for their help and guidance.

### DIFF
--- a/src/runtime/debug/debug.go
+++ b/src/runtime/debug/debug.go
@@ -1,6 +1,8 @@
 // Package debug is a dummy package that is not yet implemented.
 package debug
 
+import "runtime"
+
 // SetMaxStack sets the maximum amount of memory that can be used by a single
 // goroutine stack.
 //
@@ -27,16 +29,17 @@ func Stack() []byte {
 //
 // Not implemented.
 func ReadBuildInfo() (info *BuildInfo, ok bool) {
-	return nil, false
+	return &BuildInfo{GoVersion: "tinygo" + runtime.Version()}, true
 }
 
 // BuildInfo represents the build information read from
 // the running binary.
 type BuildInfo struct {
-	Path     string    // The main package path
-	Main     Module    // The module containing the main package
-	Deps     []*Module // Module dependencies
-	Settings []BuildSetting
+	GoVersion string    // version of the Go toolchain that built the binary, e.g. "go1.19.2"
+	Path      string    // The main package path
+	Main      Module    // The module containing the main package
+	Deps      []*Module // Module dependencies
+	Settings  []BuildSetting
 }
 
 type BuildSetting struct {
@@ -57,4 +60,11 @@ type Module struct {
 // Not implemented.
 func SetGCPercent(n int) int {
 	return n
+}
+
+// String implements Stringer for BuildInfo.
+//
+// Not implemented.
+func (bi *BuildInfo) String() string {
+	return "<build info placeholder>\n"
 }
--- a/src/runtime/debug/debug.go
+++ b/src/runtime/debug/debug.go
@@ -29,7 +29,7 @@ func Stack() []byte {
 //
 // Not implemented.
 func ReadBuildInfo() (info *BuildInfo, ok bool) {
-	return &BuildInfo{GoVersion: "tinygo" + runtime.Version()}, true
+	return &BuildInfo{GoVersion: runtime.Compiler + runtime.Version()}, true
 }
 
 // BuildInfo represents the build information read from
--- a/src/runtime/debug/debug.go
+++ b/src/runtime/debug/debug.go
@@ -1,4 +1,8 @@
-// Package debug is a dummy package that is not yet implemented.
+// Portions copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package debug is a very partially implemented package to allow compilation.
 package debug
 
 import (
--- a/src/runtime/debug/debug.go
+++ b/src/runtime/debug/debug.go
@@ -1,7 +1,12 @@
 // Package debug is a dummy package that is not yet implemented.
 package debug
 
-import "runtime"
+import (
+	"fmt"
+	"runtime"
+	"strconv"
+	"strings"
+)
 
 // SetMaxStack sets the maximum amount of memory that can be used by a single
 // goroutine stack.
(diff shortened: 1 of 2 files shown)


dkegel-fastly (approved):
LGTM

src/runtime/debug/debug.go
```diff
@@ -27,16 +29,17 @@ func Stack() []byte {
 //
 // Not implemented.
 func ReadBuildInfo() (info *BuildInfo, ok bool) {
-	return nil, false
+	return &BuildInfo{GoVersion: "tinygo" + runtime.Version()}, true
```
ydnar: Can this reference runtime.Compiler?

src/runtime/debug/debug.go
```diff
@@ -27,16 +29,17 @@ func Stack() []byte {
...
 }
 
 // BuildInfo represents the build information read from
 // the running binary.
 type BuildInfo struct {
```
ydnar, line 46: Does this match big Go now?

src/runtime/debug/debug.go
```diff
@@ -58,3 +61,10 @@ type Module struct {
 func SetGCPercent(n int) int {
 	return n
 }
+
+// String implements Stringer for BuildInfo.
+//
+// Not implemented.
+func (bi *BuildInfo) String() string {
+	return "<build info placeholder>\n"
```
ydnar: Can you implement this?

src/runtime/debug/debug.go
ldemailly: yup, changing it

src/runtime/debug/debug.go
ldemailly: yes (at least go 1.22.5's)

src/runtime/debug/debug.go
ldemailly: given everything else is empty in the struct, what would that do?

could we just import the real code from big go?

ldemailly:
output comparison (through fortio.org/version) - edited for latest commit:
```
$ ./grol.tiny buildinfo
dev  tinygo0.33.0-dev arm64 darwin
go	tinygo0.33.0-dev
$ ./grol buildinfo
dev  go1.22.5 arm64 darwin
go	go1.22.5
path	grol.io/grol
mod	grol.io/grol	(devel)	
dep	fortio.org/cli	v1.7.0	h1:w+uXZLGi4t3Vn/BvbeMuSw84Z1pvNPG9HqeGfpP68cc=
dep	fortio.org/log	v1.15.0-pre4	h1:jAr6cn7AHhKBBdD6M2WlCqNWnePyCMp9A6++o0Zl0jY=
dep	fortio.org/struct2env	v0.4.1	h1:rJludAMO5eBvpWplWEQNqoVDFZr4RWMQX7RUapgZyc0=
dep	fortio.org/version	v1.0.4	h1:FWUMpJ+hVTNc4RhvvOJzb0xesrlRmG/a+D6bjbQ4+5U=
dep	github.com/kortschak/goroutine	v1.1.2	h1:lhllcCuERxMIK5cYr8yohZZScL1na+JM5JYPRclWjck=
build	-buildmode=exe
build	-compiler=gc
build	-tags=no_net,no_json
build	-trimpath=true
build	CGO_ENABLED=0
build	GOARCH=arm64
build	GOOS=darwin
build	vcs=git
build	vcs.revision=9396c9b051e054a7783f75195aac0084f20911d2
build	vcs.time=2024-07-16T07:02:48Z
build	vcs.modified=true
```

src/runtime/debug/debug.go
ldemailly: The original is
```go
// BuildInfo represents the build information read from a Go binary.
type BuildInfo struct {
	// GoVersion is the version of the Go toolchain that built the binary
	// (for example, "go1.19.2").
	GoVersion string

	// Path is the package path of the main package for the binary
	// (for example, "golang.org/x/tools/cmd/stringer").
	Path string

	// Main describes the module that contains the main package for the binary.
	Main Module

	// Deps describes all the dependency modules, both direct and indirect,
	// that contributed packages to the build of this binary.
	Deps []*Module

	// Settings describes the build settings used to build the binary.
	Settings []BuildSetting
}
```

src/runtime/debug/debug.go
dkegel-fastly: "else"?  It's not printing anything at all.  It could at least return "go. " + GoVersion, couldn't it?

ldemailly: ic the first line sure, I didn't notice because when I use buildinfo I print the GoVersion separately let me add something

btw should it be
```
go	tinygo0.33.1
```
or
```
tinygo tinygo0.33.1
```

ldemailly: doing the former, copying the code

ldemailly: @dkegel-fastly so I copied the real code but... it's ugly... what is the general philosophy/must not be the first time tinygo is just a small variant of an existing go file, can you import+patch or some other way to avoid copy pasta (this is more for my education than anything specific for this PR - hopefully it's ok now with https://github.com/tinygo-org/tinygo/pull/4343/commits/41dd04a20b5675741156621549c3c9df2900e3f9 )

ydnar (approved):

src/runtime/debug/debug.go
```diff
@@ -1,6 +1,13 @@
 // Package debug is a dummy package that is not yet implemented.
```
ydnar: Should this comment be removed?

ldemailly: it's still kinda dummy - I was wondering about the comment
```go
// Not implemented.
func ReadBuildInfo() 
```
and figured to leave it as it's... only giving an empty one (no modules, no settings etc) except for GoVersion

ldemailly:
I see the approvals, thanks, would one of you merge, I can't (wouldn't anyway)

src/runtime/debug/debug.go
```diff
@@ -58,3 +66,60 @@ type Module struct {
 func SetGCPercent(n int) int {
 	return n
 }
+
+// Start of stolen from big go. TODO: import/reuse without copy pasta.
```
dkegel-fastly, line 74: I do believe a copyright notice should be copied here...?

ldemailly: isn’t this covered by https://github.com/tinygo-org/tinygo/blob/release/LICENSE#L3 ?

dkegel-fastly: I think the copyright folks like to have a greppable marker in each file.

I like the way runtime/timer.go does it: copy the three top llines from the file you're copying, and change "Copyright" to "Portions copyright".

ldemailly: Thanks, I was looking for an example, thanks for mentioning src/runtime/timer.go

Done

ldemailly:
~~hmm I messed up my fork trying to start a new PR earlier (about TLS) let me fix this (hopefully)~~ fixed PTAL

ldemailly:
The previous build https://github.com/tinygo-org/tinygo/actions/runs/9977696304 of size difference, before llvm18 got broken, passed, so can we merge this ? (only difference is the comment at the top of the file, since that build)

ldemailly:
ping

deadprogram:
@ldemailly please rebase against the latest `dev` and this PR should pass CI now.

deadprogram:
Actually, I can just squash/merge this now, I think.

Thank you very much @ldemailly for working on this and to @ydnar and @dkegel-fastly for guidance and reviews.

PR #4344, opened 2024-07-17
syscall: remove some dead code

aykevl:
Not sure why it's here but it's not referenced anywhere.

Found by @ydnar.

### DIFF
--- a/src/syscall/syscall.go
+++ b/src/syscall/syscall.go
@@ -2,7 +2,6 @@ package syscall
 
 import (
 	"errors"
-	"sync/atomic"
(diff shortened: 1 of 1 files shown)


ydnar (changes requested):
Probably needs a gofmt pass to remove the `sync/atomic` import.

aykevl:
Should be fixed now!

(Nitpick: that's one of the things the editor does on save, but it's not actually `go fmt` that fixes up the imports.)

dgryski (approved):
LGTM

PR #4345, opened 2024-07-18
Add enough tls.ConnectionState and tls.CipherSuiteName() to compile fortio/log

ldemailly:
Note that net/http is broken on macOS but this allows it to work on Linux (and waspi)

edit: linux also doesn't work #4346

### DIFF
--- a/src/crypto/tls/common.go
+++ b/src/crypto/tls/common.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"crypto"
 	"crypto/x509"
+	"fmt"
 	"io"
 	"net"
 	"sync"
@@ -23,9 +24,21 @@ import (
 // only supports Elliptic Curve based groups. See RFC 8446, Section 4.2.7.
 type CurveID uint16
 
+// CipherSuiteName returns the standard name for the passed cipher suite ID
+//
+// Not Implemented.
+func CipherSuiteName(id uint16) string {
+	return fmt.Sprintf("0x%04X", id)
+}
+
 // ConnectionState records basic TLS details about the connection.
 type ConnectionState struct {
-	// TINYGO: empty; TLS connection offloaded to device
+	// TINYGO: empty; TLS connection offloaded to device [?]
+	//
+	// Minimum (empty) fields for fortio.org/log http logging to
+	// compile:
+	PeerCertificates []*x509.Certificate
+	CipherSuite      uint16
 }
 
 // ClientAuthType declares the policy the server will follow for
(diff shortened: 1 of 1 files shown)


src/crypto/tls/common.go
```diff
@@ -23,9 +24,21 @@ import (
...
+func CipherSuiteName(id uint16) string {
+	return fmt.Sprintf("0x%04X", id)
+}
+
 // ConnectionState records basic TLS details about the connection.
 type ConnectionState struct {
-	// TINYGO: empty; TLS connection offloaded to device
+	// TINYGO: empty; TLS connection offloaded to device [?]
+	//
+	// Minimum (empty) fields for fortio.org/log http logging to
```
ydnar: What is the broader implication? `net/http` compatibility?

ldemailly: well... if I ever get a super simple server working with tinygo, it'd have the logging I use elsewhere for servers... I haven't tried others but mine does need these fields even empty to compile


edit: the wip test is https://github.com/ldemailly/go-scratch/blob/main/tinyhttp/tinyhttp.go

ldemailly: in short it's incremental progress - can we merge?

src/crypto/tls/common.go
```diff
@@ -23,9 +24,21 @@ import (
 // only supports Elliptic Curve based groups. See RFC 8446, Section 4.2.7.
 type CurveID uint16
 
+// CipherSuiteName returns the standard name for the passed cipher suite ID
+//
+// Not Implemented.
+func CipherSuiteName(id uint16) string {
+	return fmt.Sprintf("0x%04X", id)
```
dgryski, line 31: Do we want to use `strconv` or something cheaper than `fmt` here?

ldemailly: I copy pasted the fallback case from the go library - also I don't think FormatUint can do leading 0s ?
so while it could be cheaper to print the value differently, at least that part could stay however big go does it?

dgryski: SGTM

dgryski (approved):
LGTM

ldemailly:
@deadprogram thanks for having merged the other, what about this one :) cheers

ldemailly:
(just rebased)

deadprogram:
Thanks for the addition @ldemailly and to @dgryski for review. Now merging.

ldemailly:
thx! ❤️

PR #4350, opened 2024-07-20
fix: set ubuntu version to 24.04 to address #4347

deadprogram:
This PR intends to provide a fix for #4347 by setting the Ubuntu version to 24.04


deadprogram:
Merging to fix #4347

PR #4351, opened 2024-07-20
net: update to latest net package

deadprogram:
This PR is to: update to the latest `net` package to obtain recent updates for both wasi and embedded uses.


dgryski (approved):
LGTM

deadprogram:
Thanks for review @dgryski now merging.

PR #4352, opened 2024-07-20
Avoid error spam on wasm finalizer call and ship a potentially functional version

### LINKED ISSUE #1140: syscall/js.finalizeRef not implemented
The error happened when I try an example (https://github.com/tinygo-org/tinygo/blob/master/src/examples/wasm/slices/)

Steps to reproduce:

* Clone the tinygo repository `git clone git@github.com:tinygo-org/tinygo.git`
* `cd tinygo/src/examples/wasm/slices`
* Build the wasm with `tinygo build -o wasm.wasm -target wasm --no-debug ./wasm.go`
* Copy the file `tinygo/targets/wasm_exec.js` to current `slices` folder
* Run the server, open the address in browser and start typing a number. You should see the error above.

ldemailly:
This is a slighly modified version of https://github.com/tinygo-org/tinygo/issues/1140#issuecomment-1697415864 credit to @evilnoxx and @FrankReh

Fixes #1140 

Note that... it's not set so noop mostly, you can see on https://grol.io where it's running, open the console, click run:

Screenshot 2024-07-21 at 8 46 21 AM

but hopefully for some other tinygo code it does something useful, at least it doesn't ~~crash~~ show errors all the time anymore

### DIFF
--- a/targets/wasm_exec.js
+++ b/targets/wasm_exec.js
@@ -302,7 +302,6 @@
 						// for one specific case, by js.go:jsString. and can/might leak memory.
 						const id = mem().getUint32(unboxValue(v_ref), true);
 						// Note that this if is so far seemingly never true. Someone should investigate why.
-						// (ie we are silently, no log even, ignoring an unexpected case)
 						if (this._goRefCounts?.[id] !== undefined) {
 							this._goRefCounts[id]--;
 							if (this._goRefCounts[id] === 0) {
@@ -311,6 +310,9 @@
 								this._ids.delete(v);
 								this._idPool.push(id);
 							}
+						} else {
+							// Log as a hint and reminder that something is probably off.
+							console.log("syscall/js.finalizeRef: unknown id", id);
 						}
 					},
 
--- a/targets/wasm_exec.js
+++ b/targets/wasm_exec.js
@@ -298,11 +298,22 @@
 
 					// func finalizeRef(v ref)
 					"syscall/js.finalizeRef": (v_ref) => {
-						// Note: TinyGo does not support finalizers so this should never be
-						// called.
-						console.error('syscall/js.finalizeRef not implemented');
+						// Note: TinyGo does not support finalizers so this is only called
+						// for one specific case, by js.go:jsString. and can/might leak memory.
+						const id = mem().getUint32(unboxValue(v_ref), true);
+						if (this._goRefCounts?.[id] !== undefined) {
+							console.log("syscall/js.finalizeRef", id, this._goRefCounts[id]);
+							this._goRefCounts[id]--;
+							if (this._goRefCounts[id] === 0) {
+								const v = this._values[id];
+								this._values[id] = null;
+								this._ids.delete(v);
+								this._idPool.push(id);
+							}
+						} else {
+							console.log("syscall/js.finalizeRef: unknown id", id);
+						}
 					},
-
 					// func stringVal(value string) ref
 					"syscall/js.stringVal": (value_ptr, value_len) => {
 						const s = loadString(value_ptr, value_len);
--- a/targets/wasm_exec.js
+++ b/targets/wasm_exec.js
@@ -301,8 +301,9 @@
 						// Note: TinyGo does not support finalizers so this is only called
 						// for one specific case, by js.go:jsString. and can/might leak memory.
 						const id = mem().getUint32(unboxValue(v_ref), true);
+						// Note that this if is so far seemingly never true. Someone should investigate why.
+						// (ie we are silently, no log even, ignoring an unexpected case)
 						if (this._goRefCounts?.[id] !== undefined) {
-							console.log("syscall/js.finalizeRef", id, this._goRefCounts[id]);
 							this._goRefCounts[id]--;
 							if (this._goRefCounts[id] === 0) {
 								const v = this._values[id];
@@ -310,10 +311,9 @@
 								this._ids.delete(v);
 								this._idPool.push(id);
 							}
-						} else {
-							console.log("syscall/js.finalizeRef: unknown id", id);
 						}
 					},
+
 					// func stringVal(value string) ref
 					"syscall/js.stringVal": (value_ptr, value_len) => {
 						const s = loadString(value_ptr, value_len);


dyaskur (approved):

targets/wasm_exec.js
```diff
@@ -298,11 +298,22 @@
...
 					// func finalizeRef(v ref)
 					"syscall/js.finalizeRef": (v_ref) => {
-						// Note: TinyGo does not support finalizers so this should never be
-						// called.
-						console.error('syscall/js.finalizeRef not implemented');
+						// Note: TinyGo does not support finalizers so this is only called
+						// for one specific case, by js.go:jsString. and can/might leak memory.
+						const id = mem().getUint32(unboxValue(v_ref), true);
+						if (this._goRefCounts?.[id] !== undefined) {
+							console.log("syscall/js.finalizeRef", id, this._goRefCounts[id]);
```
aykevl: Why this extra check?
1. `this._goRefCounts` won't be undefined, so the `?.` operator is not needed here.
2. I don't see an extra check in wasm_exec.js from upstream Go, so why would it be needed here? (I think it can be removed, or if it can't that indicates there's probably a bug somewhere else).

Also, can you remove these `console.log` logging statements? All they will do is spam the console without providing useful information to the developer (and if the developer actually wants to debug the finalizer, they can easily insert some temporary debug loggin).

aykevl (commented):
See the inline comment.

targets/wasm_exec.js
ldemailly: it actually is needed as in it doesn't crash but it's also not set - index seems to always be 0 and the ref not exist in my case so it's defensive

ldemailly: another way to say this is I am not sure the code inside the if ever runs, thus the leak log for further investigation- ie it's better than crashing like it does today but... not much better. 

edit: _goRefCount is seemingly set in some cases in this same file but I didn't understand it/how it matches with it not actually being set in my case (outside of that comment)

ldemailly: @aykevl I updated the description, PTAL. If you have a (manual or otherwise) test to suggest where it wouldn't be unset/0 lmk, I would feel better seeing the code for `--` exercised at least once

ldemailly:
and if using the same instance of wasm from the console:
Screenshot 2024-07-21 at 8 55 51 AM

ldemailly:
gentle ping :)

targets/wasm_exec.js
deadprogram: > I don't see an extra check in wasm_exec.js from upstream Go, so why would it be needed here?

I suppose that means this is either also a bug in big Go or else as @aykevl said a bug someplace else?

targets/wasm_exec.js
```diff
@@ -298,11 +298,22 @@
...
+							this._goRefCounts[id]--;
+							if (this._goRefCounts[id] === 0) {
+								const v = this._values[id];
+								this._values[id] = null;
+								this._ids.delete(v);
+								this._idPool.push(id);
+							}
+						} else {
+							console.log("syscall/js.finalizeRef: unknown id", id);
```
deadprogram, line 315: Should probably remove this call to `log()` entirely, or else switch to `error()`. 

Assuming this entire PR is a good idea, which I don't have an opinion on at the moment. @ldemailly and @aykevl you both have a better grasp on the issues involved.

ldemailly: It's definitely not an error() because that would break/defeat the fix and I think it helps showing what happens for future troubleshooting but if it's all that stands between merging I can also remove it

targets/wasm_exec.js
ldemailly: > I suppose that means this is either also a bug in big Go

No, big go "works": it doesn't get invalid ids/not found

Screenshot 2024-07-23 at 7 55 13 AM

(added the same if/else logging there)

```javascript
					// func finalizeRef(v ref)
					"syscall/js.finalizeRef": (sp) => {
						sp >>>= 0;
						const id = this.mem.getUint32(sp + 8, true);
						if (this._goRefCounts?.[id] !== undefined) {
							console.log("syscall/js.finalizeRef", id, this._goRefCounts[id]);
							this._goRefCounts[id]--;
							if (this._goRefCounts[id] === 0) {
								const v = this._values[id];
								this._values[id] = null;
								this._ids.delete(v);
								this._idPool.push(id);
							}
						} else {
							console.log("syscall/js.finalizeRef: unknown id", id);
						}
					},
```

So yes there is a leak or issue, thus the comments and the log, but at least it's functional instead of making every single wasm js non functional like it is today afaik

ldemailly: to see (modified) big go's version https://github.com/grol-io/grol/blob/big_go_wasm/wasm/wasm_exec.js#L302-L303

targets/wasm_exec.js
deadprogram: It appears from looking at the big Go implementation and also the current TinyGo implementation, and also @aykevl comment, that they think the code change should probably be something like this:

```js
// func finalizeRef(v ref)
"syscall/js.finalizeRef": (v_ref) => {
        const id = mem().getUint32(unboxValue(v_ref), true);
	this._goRefCounts[id]--;
	if (this._goRefCounts[id] === 0) {
		const v = this._values[id];
		this._values[id] = null;
		this._ids.delete(v);
		this._idPool.push(id);
	}
}
```

Is that a correct summary?

ldemailly: no, not sure how you get to that conclusion, sorry

https://github.com/tinygo-org/tinygo/pull/4352/files#r1688222655

doing that would just silently ignore the case where you get an unexpected id, which appears to be _all_ the cases and I don't think that's more helpful than a log

yet I suppose anything that doesn't involve `console.error` would be progress, not sure I get the resistance to this current version though.

ldemailly: this being said... I thought console.error() was fatal (ie current version of the code) but it's not or I used a different setup when it hit me - so technically this is isn't needed at all, outside of generating console errors every invocation. I think this is progress but....

deadprogram: This is the big Go implementation here: https://github.com/golang/go/blob/master/misc/wasm/wasm_exec.js#L298-L309

There are no calls to `console.log` or any other logging. Hence TinyGo should not do any logging by default to keep the implementations as close as possible.

What I think @aykevl was saying is that anyone who needs to specifically log this, could modify the `wasm_exec.js` themselves if they need to.

ldemailly: https://github.com/golang/go/blob/master/misc/wasm/wasm_exec.js#L22

ldemailly: as in when something is dubious a log is appropriate imo - but... will remove and nobody will know that it's non functional... but maybe we can move forward

ldemailly: done

deadprogram: > https://github.com/golang/go/blob/master/misc/wasm/wasm_exec.js#L22

The `wasm_exec.js` implementation for `finalizeRef` does not call `console.log` to be more specific:
https://github.com/golang/go/blob/master/misc/wasm/wasm_exec.js#L298-L309

ldemailly: I am hiding the bug now,  even if I think it's wrong, can we merge?

(my point was there is precedent in big go for logging unexpected cases https://github.com/tinygo-org/tinygo/pull/4352#discussion_r1688371365)

targets/wasm_exec.js
```diff
@@ -298,9 +298,20 @@
...
+						// (ie we are silently, no log even, ignoring an unexpected case)
+						if (this._goRefCounts?.[id] !== undefined) {
+							this._goRefCounts[id]--;
+							if (this._goRefCounts[id] === 0) {
+								const v = this._values[id];
+								this._values[id] = null;
+								this._ids.delete(v);
+								this._idPool.push(id);
+							}
+						}
```
deadprogram, line 316:
```suggestion
						} else {
							console.error("syscall/js.finalizeRef: unknown id", id);
						}
```

deadprogram: So something like this is basically what you were proposing to catch the error you referred to? (but with the correct formatting)

deadprogram: Sorry about all the edits, GH did not like my spaces.

targets/wasm_exec.js
deadprogram: @ldemailly please see my other code comment, I think I understand better now what you intended.

targets/wasm_exec.js
ldemailly: ? that's how it was in the first commit ... https://github.com/tinygo-org/tinygo/pull/4352/commits/e4e6e0d32c02e21ebdcba026ed96be7e550b8a54#diff-3c36e56a725ebf4b2f5fd4a5a117e50894d15570f979b76d0affc410430ccec5R314

deadprogram: That first commit added two lines with calls to `console.log()`, one in each clause of the `if`. So not quite the same.

ldemailly: so you want my original commit, but without the first log?

ldemailly: done, PTAL

targets/wasm_exec.js
```diff
@@ -298,9 +298,22 @@
...
+							this._goRefCounts[id]--;
+							if (this._goRefCounts[id] === 0) {
+								const v = this._values[id];
+								this._values[id] = null;
+								this._ids.delete(v);
+								this._idPool.push(id);
+							}
+						} else {
+							// Log as a hint and reminder that something is probably off.
+							console.log("syscall/js.finalizeRef: unknown id", id);
```
deadprogram, line 315:
```suggestion
							console.error("syscall/js.finalizeRef: unknown id", id);
```

deadprogram: This is what I meant.

ldemailly: That would get back to previous state where every invocation would error, which is what I'm trying to avoid, I think it's fine as low level log "for developers", as a reminder it needs more work,

For users of a wasm module I think they shouldn't worry about it imo
maybe a compromise is a Warning but even that would probably make me want to not produce wasm using tinygo

targets/wasm_exec.js
ldemailly: I missed btw you had put `error` because the whole point is not not have `error` (title and description of this PR) thus I didn't get your point (not to say I agree with it, I don't, but my bad on autocorrecting your .error to .log)

targets/wasm_exec.js
deadprogram: > That would get back to previous state where every invocation would error

I suppose I do not understand the purpose of this PR in that case. I thought you wanted to see the error in logs from some of your previous comments.

ldemailly: The purpose is two(*) fold:

- avoid showing errors on otherwise functional or seemingly functional wasm js
- use the code mentioned in and make some progress on the 4 year old open issue that might apply in some cases
- yet not just assume the id or ref object exist when they don't, allowing further investigation 


*: off by one error

ldemailly: in other words see the not quite expected case in the log, but not as error

ldemailly:
ping

ldemailly:
so?

targets/wasm_exec.js
```diff
@@ -298,9 +298,22 @@
...
 					// func finalizeRef(v ref)
 					"syscall/js.finalizeRef": (v_ref) => {
-						// Note: TinyGo does not support finalizers so this should never be
-						// called.
-						console.error('syscall/js.finalizeRef not implemented');
+						// Note: TinyGo does not support finalizers so this is only called
+						// for one specific case, by js.go:jsString. and can/might leak memory.
+						const id = mem().getUint32(unboxValue(v_ref), true);
+						// Note that this if is so far seemingly never true. Someone should investigate why.
+						if (this._goRefCounts?.[id] !== undefined) {
```
prochac, line 305: The `unboxValue` returns the string as a byte array. It leads to `id` being `0` in my case.
So I looked around, and the `unboxValue` contained an interesting part of code:
```
const id = v_ref & 0xffffffffn;
return this._values[id];
```

So I tried this:
```suggestion
						// magicValue taken from `unboxValue` function.
						const magicValue = 0xffffffffn;
						const id = v_ref & magicValue;
						if (this._goRefCounts?.[id] !== undefined) {
```
And when I checked `this._ids.size` before and after the `delete`, it really deleted something :D Can it be it? Is it working? Like, I have no idea what I'm doing and what the hex number is for.

@deadprogram any idea what the constant does?

ribasushi:
Womdering what is the status of this...

targets/wasm_exec.js
deadprogram: @prochac I think in this case is is the highest ID allowed.

Also, from a superficial reading of your proposed change, that looks like it could in fact be the correct way to obtain the ID.

Is there anyone who can test this out?

deadprogram:
@ribasushi or anyone else regarding https://github.com/tinygo-org/tinygo/pull/4352#discussion_r1961879277 do you have a test example to see this working? I am not sure how to verify it.

targets/wasm_exec.js
ribasushi: @deadprogram I'd be happy to test, but I am a bit lost wich version... The file as seen here https://github.com/tinygo-org/tinygo/blob/99e6e54e5f0/targets/wasm_exec.js works for me without errors, although the console does say `syscall/js.finalizeRef: unknown id 0`, unsure if expected or not...

deadprogram: It would be that same file but with this line https://github.com/tinygo-org/tinygo/blob/99e6e54e5f0/targets/wasm_exec.js#L303C7-L303C59

changed to:

```js
						const id = v_ref & 0xffffffffn;
```

as proposed by @prochac

ribasushi: @deadprogram seems to work flawlessly, console is silent

targets/wasm_exec.js
```diff
@@ -298,9 +298,22 @@
 
 					// func finalizeRef(v ref)
 					"syscall/js.finalizeRef": (v_ref) => {
-						// Note: TinyGo does not support finalizers so this should never be
-						// called.
-						console.error('syscall/js.finalizeRef not implemented');
+						// Note: TinyGo does not support finalizers so this is only called
+						// for one specific case, by js.go:jsString. and can/might leak memory.
+						const id = mem().getUint32(unboxValue(v_ref), true);
```
deadprogram, line 303:
```suggestion
						const id = v_ref & 0xffffffffn;
```

Based on @prochac suggestion.

ribasushi: @ldemailly I can confirm this works (tested with: https://github.com/rsc/qr/tree/master/qart)

targets/wasm_exec.js
deadprogram: I added a commit with the @prochac change, but some how it ended up in @ldemailly fork, and also now this PR has been "Checking for ability to merge automatically…" for about 30 minutes. Something went wrong someplace?

I will try to `git rebase` on my local branch to see if I can get this finally merged into `dev`,,,,

deadprogram:
Closing since replaced by #4744 thanks everyone!

PR #4354, opened 2024-07-21
build: add package ID to compiler and optimization error messages

aykevl:
This improves error reporting slightly. You can see the improvements in the changed tests.


deadprogram (approved):
Looks good, thanks @aykevl

PR #4356, opened 2024-07-21
More testing.T and go-cmp compatibility

ldemailly:
- [x] Adding missing T.Deadline() (used by testscript) -> moved to #4362 
- [x] Avoid panic in deep equal with map of interfaces -> moved to #4360 
- [x] Avoid panic in [go-cmp](https://github.com/google/go-cmp/blob/v0.6.0/cmp/compare.go#L312-L314) when it tries to find an Equal method  -> moved to #4361 
  - This one is a bit dicey as I don't know how to make it conditionally crash/not crash (crash for people who think it works, not crash for people like go-cmp for which it's only one option)

Diff is stack on top of https://github.com/tinygo-org/tinygo/pull/4352 but once that one merges and I rebase it'll have only the top commit (not sure how you guys handle this)

### DIFF
--- a/src/reflect/type.go
+++ b/src/reflect/type.go
@@ -1083,8 +1083,10 @@ func (t rawType) Method(i int) Method {
 	panic("unimplemented: (reflect.Type).Method()")
 }
 
-func (t rawType) MethodByName(name string) (Method, bool) {
-	panic("unimplemented: (reflect.Type).MethodByName()")
+func (t rawType) MethodByName(name string) (m Method, ok bool) {
+	// Avoid panic in in go-cmp t.MethodByName("Equal")
+	// https://github.com/google/go-cmp/blob/v0.6.0/cmp/compare.go#L312-L314
+	return // false
 }
 
 func (t *rawType) PkgPath() string {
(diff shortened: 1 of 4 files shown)


dkegel-fastly:
Witholding comment for tonight other than to say fixing the little nits you have been running into is great... ideally each one of the little fixes would be its own PR, and some might even benefit from a simple test.

deadprogram:
@ldemailly this PR would be a lot easier to review if it were actually 3 different PRs:

- PR for "Add missing T.Deadline":
https://github.com/tinygo-org/tinygo/pull/4356/commits/c8bd98d5dad4c6a44c4e870bea37a11b33ecb2d8

- PR for https://github.com/tinygo-org/tinygo/pull/4356/commits/6465bee19f217f61d1a153eb01a9b8d7c43608e8 and https://github.com/tinygo-org/tinygo/pull/4356/commits/c8bd98d5dad4c6a44c4e870bea37a11b33ecb2d8

- PR for https://github.com/tinygo-org/tinygo/pull/4356/commits/84ad1a00a40c4ed51f3317331985d789ac3c5101

None of these needs the PR in https://github.com/tinygo-org/tinygo/pull/4352 so it can just be based on the `dev` branch.

Please let me know if this makes sense to you, and/or if I can help out in any way. Thanks!

ldemailly:
done

PR #4360, opened 2024-07-23
Getting DeepEqual to work with maps whose key is an interface

ldemailly:
-[x]  Avoid panic in deep equal with map of interfaces

Splitting #4356 into 3 PRs...

### DIFF
--- a/src/reflect/all_test.go
+++ b/src/reflect/all_test.go
@@ -1102,6 +1102,20 @@ func init() {
 	cycleMap3["different"] = cycleMap3
 }
 
+type deepEqualInterface interface {
+	Foo()
+}
+type deepEqualConcrete struct {
+	int
+}
+
+func (deepEqualConcrete) Foo() {}
+
+var (
+	deepEqualConcrete1 = deepEqualConcrete{1}
+	deepEqualConcrete2 = deepEqualConcrete{2}
+)
+
 var deepEqualTests = []DeepEqualTest{
 	// Equalities
 	{nil, nil, true},
@@ -1119,6 +1133,7 @@ var deepEqualTests = []DeepEqualTest{
 	{[]byte{1, 2, 3}, []byte{1, 2, 3}, true},
 	{[]MyByte{1, 2, 3}, []MyByte{1, 2, 3}, true},
 	{MyBytes{1, 2, 3}, MyBytes{1, 2, 3}, true},
+	{map[deepEqualInterface]string{deepEqualConcrete1: "a"}, map[deepEqualInterface]string{deepEqualConcrete1: "a"}, true},
 
 	// Inequalities
 	{1, 2, false},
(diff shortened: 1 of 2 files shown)


dgryski:
So does this make it *work* or just make it not *fail* ?  Can we uncomment or add any tests here?

ldemailly:
Edit: `tinygo test reflect` is how, thanks for the tip

~~@dgryski I can add a test (have one in grol) - just not sure... can you share how do I run the tests ?~~

~~I tried~~
```
$ tinygo test ./src/reflect/
package std/reflect
	src/reflect/type.go:67:2: use of internal package internal/itoa not allowed
cannot resolve packages: exit status 1
```

ldemailly:
@dgryski  done, added new test (the file is huge and messy so not sure which test to uncomment, sorry) without the fix:
```
$ tinygo test reflect 
panic: reflect: unimplemented: AssignableTo with interface
FAIL	reflect	0.234s
```

dgryski:
`all_test.go` is from upstream.  If you want, it will probably be easier to either create a new `deepequal_test.go` file or (probably better) add a test to `value_test.go` instead.

ldemailly:
mind contributing that? I demonstrated the fix works, has a test added with minimal duplication (table driven), there are plenty of changes in all_test.go - I don't have more time for this

dgryski (approved):
LGTM

dgryski:
@ldemailly This is fine to merge as-is.  I see we've already made a bunch of patches to `all_test.go`, so it's not just a pristine copy of upstream (plus comments).

ldemailly:
Thanks, cool, can you click squash&merge?

also the other PRs too if possible

deadprogram:
Squash/merged. Thank you very much @ldemailly for all the changes and to @dgryski for review.

aykevl:
This change is not correct: it now allows code while it should panic.
Here is an example:
```go
package main

import (
        "fmt"
        "reflect"
)

type Foo interface {
        Foo()
}

type Bar interface {
        Bar()
}

func main() {
        m := map[Foo]int{}
        v := reflect.ValueOf(m)
        key := reflect.ValueOf(struct{
                Key Bar
        }{}).Field(0)
        val := v.MapIndex(key)
        fmt.Println("val:", val)
}
```

Go:

```
panic: reflect.Value.MapIndex: value of type main.Bar is not assignable to type main.Foo

goroutine 1 [running]:
reflect.Value.assignTo({0xab3a0?, 0x4000010070?, 0x4000010070?}, {0xbcbd1, 0x16}, 0xab320, 0x0)
        /usr/local/go1.22.2/src/reflect/value.go:3356 +0x20c
reflect.Value.MapIndex({0xaa900?, 0x4000104ef8?, 0x4000104dc8?}, {0xab3a0, 0x4000010070, 0x94})
        /usr/local/go1.22.2/src/reflect/value.go:1821 +0xd0
main.main()
        /home/ayke/src/tinygo/tinygo/tmp/pr4360.go:22 +0x138
exit status 2
```

TinyGo:

```
val: <invalid reflect.Value>
```

ldemailly:
when is that panic is useful, vs getting DeepEqual working? clearly one can construct code that changes compared to the fix, yet what does that buy?

ldemailly:
There are thousands of valid go programs that fail with tinygo because tinygo is incomplete, my assumption is that the goal of taking contributions is to incrementally shrink that set and make it more likely for valid programs to succeed, as long as it doesn't cause bad side effects (like say crashes)

This MR accomplishes that, the counter example and the revert, I'm not sure what they accomplish? 

You may want to update https://tinygo.org/docs/guides/contributing/#something-in-go-that-you-wantneed-does-not-appear-to-be-in-tinygo if I'm misunderstanding project goals

PR #4361, opened 2024-07-23
make go-cmp Equal succeed

ldemailly:
- [x] Avoid panic in [go-cmp](https://github.com/google/go-cmp/blob/v0.6.0/cmp/compare.go#L312-L314) when it tries to find an Equal method 
  - This one is a bit dicey as I don't know how to make it conditionally crash/not crash (crash for people who think it works, not crash for people like go-cmp for which it's only one option)
  
Splitting commits from #4356

### DIFF
--- a/src/reflect/type.go
+++ b/src/reflect/type.go
@@ -1083,8 +1083,10 @@ func (t rawType) Method(i int) Method {
 	panic("unimplemented: (reflect.Type).Method()")
 }
 
-func (t rawType) MethodByName(name string) (Method, bool) {
-	panic("unimplemented: (reflect.Type).MethodByName()")
+func (t rawType) MethodByName(name string) (m Method, ok bool) {
+	// Avoid panic in in go-cmp t.MethodByName("Equal")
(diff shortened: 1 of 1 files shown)


ldemailly:
'Test' is here

https://github.com/grol-io/grol/blob/v0.24.0/ast/modify_test.go#L130

panics without this fix, also 

 https://github.com/google/go-cmp/blob/v0.6.0/cmp/compare.go#L312-L314

in the comment hopefully makes it clear how this helps
(of note, if you _do_ have an equal method it won't be used, yet... still better than panic imo)

aykevl (changes requested):
This will make it very difficult to debug existing code that calls `MethodByName`: now it's very difficult to find why existing code (that calls `MethodByName`) works differently in TinyGo. Therefore, I think this is a bad idea.

aykevl:
Would it be possible to change go-cmp instead to not call MethodByName?

ldemailly:
I don't think go-cmp are (or should be, really) concerned with adding special case for incomplete go, so it would need to be on tinygo's side (or users of both like me)

Not sure btw that it's hard to find callers without panic

maybe something with some build tag then?

PR #4362, opened 2024-07-23
Add missing testing.T.Deadline

ldemailly:
Allows compilation of code calling Deadline (like testscript) and is also somewhat functional given the api (it never timesout but ... works)

Split from #4356


deadprogram:
Thanks for the addition here @ldemailly now merging.

PR #4363, opened 2024-07-24
WIP: implement realloc() for slice growth

dgryski:


aykevl (commented):
I think this deviates from the Go language specification.

Consider the following code:

```go
slice := make([]int, 4)
slice2 := append(slice, 1234)
slice[0] = 20
println(slice[1])
```

Normally the `append` would create a new slice instead of reusing the old one (unless `cap` allows extending it, which is not the case here). But here, the same slice is reused, which changes the behavior: println might print 20 instead of 0.

I have actually written code myself that relies on this behavior. I'm pretty sure it's specified in the Go spec (though I haven't checked).

dgryski:
Hmm. Good point. I was looking at this for an optimization where the caller is doing a lot of appends. I'll check the spec and see if we have any wiggle room.

dgryski:
```
If the capacity of s is not large enough to fit the additional values, append [allocates](https://go.dev/ref/spec#Allocation) a new, sufficiently large underlying array that fits both the existing slice elements and the additional values. Otherwise, append re-uses the underlying array.
```

I think the word "new" there certainly implies we can't `realloc()`, and the the three-form slice notation was added to allow people to append to a slice and *know* that it would be a copy that was returned.  Oh well, back to the drawing board...

PR #4364, opened 2024-07-25
Fix typos

deining:
This PR fixes various typos I spotted in the project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -217,7 +217,7 @@
   - `reflect`: add SetZero
   - `reflect`: fix iterating over maps with interface{} keys
   - `reflect`: implement Value.Grow
-  - `reflect`: remove unecessary heap allocations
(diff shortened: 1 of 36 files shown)


dkegel-fastly:
Changes should be based on dev, not release.

That said, I took the liberty of improving golangci/misspell to handle (some of?) those typos, created a PR for that, and then created https://github.com/tinygo-org/tinygo/pull/4387 to use it to fix (most of?) the typos this PR fixes.

So probably want to just close this PR and maybe have another look when 4387 lands...?

dkegel-fastly:
That PR has landed, and https://github.com/tinygo-org/tinygo/pull/4406 will mop up a few it missed.

Closing in favor of those efforts.
Feel free to open a new PR that adds entries to spellfix.csv and includes the results of 'make spellfix'.

Thanks for pointing out the problems!

PR #4365, opened 2024-07-26
GNUmakefile: add install target for linux systems

leongross:
Small PR adding the option to quickly install custom build binaries to the system path, similar to the way the [docs](https://tinygo.org/docs/guides/build/manual-llvm/#adding-tinygo-to-your-path) describe it. There are 2 options: local and system-wide installation. The default is system-wide installation.

```sh
# defaults to system installation
make install

# system installation, INSTALL_SYSTEM_DIR defaults to /usr/bin
INSTALL_SYSTEM_DIR=<customDir> make install

# user installation, INSTALL_LOCAL_DIR defaults to $HOME/go/bin
INSTALL_LOCAL=1 INSTALL_LOCAL_DIR=<customDir> make install
```

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -2,6 +2,8 @@
 # aliases
 all: tinygo
 
+#SHELL := /usr/bin/env bash
+
 # Default build and source directories, as created by `make llvm-build`.
 LLVM_BUILDDIR ?= llvm-build
 LLVM_PROJECTDIR ?= llvm-project
@@ -35,6 +37,10 @@ GOTESTPKGS ?= ./builder ./cgo ./compileopts ./compiler ./interp ./transform .
 
 # tinygo binary for tests
 TINYGO ?= $(call detect,tinygo,tinygo $(CURDIR)/build/tinygo)
+
+# isntalltion location defaults
+INSTALL_LOCAL_DIR ?= $(HOME)/go/bin/
+INSTALL_SYSTEM_DIR ?= /usr/bin/
 
 # Check for ccache if the user hasn't set it to on or off.
 ifeq (, $(CCACHE))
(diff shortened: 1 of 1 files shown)


aykevl:
Is there a reason you want to put it in a system path, and not effectively run `go install` (that puts it in `$GOBIN`)? Putting it in `$GOBIN` would make much more sense to me (and in fact, I'd use that a lot if it were available).

leongross:
On a multi-user system, it might be useful to install the binary on a system path.
Especially when uninstalling the tinygo package from the system path via the systems package manager for development purposes.
But the local installation should be replaced by a `go install`. I still think it is a good idea to wrap this into a `make install`, since all the rest of the build system is also wrapped into that so it would just make sense.

leongross:
@dkegel-fastly I saw you have been working on the GNUmakefile, thoughts on this PR?

GNUmakefile
```diff
@@ -36,6 +38,10 @@ GOTESTPKGS ?= ./builder ./cgo ./compileopts ./compiler ./interp ./transform .
 # tinygo binary for tests
 TINYGO ?= $(call detect,tinygo,tinygo $(CURDIR)/build/tinygo)
 
+# isntalltion location defaults
+INSTALL_LOCAL_DIR ?= $(HOME)/go/bin/
```
dkegel-fastly, line 42: Have you considered providing the old-school standard variables and targets that packagers rely on and experienced "make" users expect?  

In particular, I think autoconf-generated configure scripts let users control installation location in two ways
```
make install prefix=/alternate/directory
make install DESTDIR=/alternate/directory
```
If I understand correctly, both are needed because the first method works on windows (as it replaces the whole directory including drive letter), and the second is expected by e.g. dh_auto_install, but they might have identical effect for us.

Debian packagers would thank you :-)

See https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/html_node/Installation-Names.html
and e.g. https://manpages.debian.org/testing/debhelper/dh_auto_install.1.en.html

dankegel: Gosh darn it, now I remember: GNUmakefile is only for the static build.  "Make install" should copy build/TinyGo, not do "go install".   Glad you got that right.

Also, using profile and DESTDIR must stack; DESTDIR applies after profile. (It is for package wranglers only.)

leongross: Thanks @dkegel-fastly for the feedback. I should really adapt this patch to the convention utilizing `prefix` and `DESTDIR`.

dkegel-fastly: also fix the typos in the comments else the dreaded Spelling Police will come after you :-)

deadprogram:
Hi @leongross are you still working on this PR?

leongross:
I stalled it due to other work, maybe someone else can pick this up.

PR #4366, opened 2024-07-26
Revert "Getting DeepEqual to work with maps whose key is an interface (#4360)"

aykevl:
This reverts commit 1fd75ffbda1ff021289f2feb6a062434df68d8b7.

The change is not correct and allows code to continue while it should panic. For details, see:
https://github.com/tinygo-org/tinygo/pull/4360#issuecomment-2252943012

A correct implementation shouldn't just remove the panic, but instead actually check whether the key is assignable.
This requires an implementation of `AssignableTo` for interfaces, which is going to be a lot more work to get right.

### DIFF
--- a/src/reflect/all_test.go
+++ b/src/reflect/all_test.go
@@ -1102,20 +1102,6 @@ func init() {
 	cycleMap3["different"] = cycleMap3
 }
 
-type deepEqualInterface interface {
-	Foo()
-}
-type deepEqualConcrete struct {
-	int
-}
-
-func (deepEqualConcrete) Foo() {}
-
-var (
-	deepEqualConcrete1 = deepEqualConcrete{1}
-	deepEqualConcrete2 = deepEqualConcrete{2}
-)
-
 var deepEqualTests = []DeepEqualTest{
 	// Equalities
 	{nil, nil, true},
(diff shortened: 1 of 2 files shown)


dgryski:
Yup, you're right.

dgryski (approved):

ldemailly:
when is the panic useful, versus having DeepEqual functioning? ~~can we make code that does both?~~

ldemailly:
Repeating here:

There are thousands of valid go programs that fail with tinygo because tinygo is incomplete, my assumption is that the goal of taking contributions is to incrementally shrink that set and make it more likely for valid programs to succeed, as long as it doesn't cause bad side effects (like say crashes)

The original PR accomplished that, the counter example and the revert, I'm not sure what they accomplish?

You may want to update https://tinygo.org/docs/guides/contributing/#something-in-go-that-you-wantneed-does-not-appear-to-be-in-tinygo if I'm misunderstanding project goals

What is "correct" vs "useful" ? also valuing the work people put into contributions.

aykevl:
> What is "correct" vs "useful" ?

It is more important to run programs _correctly_ than to be able to run them but with bugs. And if some part of TinyGo is incomplete, it is better to say so (using a panic, compiler error, etc) than it is to ignore the problem and produce incorrect programs. The reason is that if it fails loudly, the user knows where to look for a fix (for example, in this case by avoiding `MapIndex`) than just have a program that compiles, runs, but does something unexpected.

aykevl:
Meanwhile I'm working on a correct fix for this issue (that would fully support AssignableTo and Implements).

dgryski:
@aykevl that's great! There's a bunch of code that will enable to run.

ldemailly:
Deepequal worked , with your revert it stopped working (nice removing the tests demonstrating so)

This is terribly disappointing and tbh poor project leadership

aykevl:
@ldemailly I'm aware of your objections. But I still stand by my opinion that this check is necessary.

ldemailly:
Got it,

Working functional useful code (DeepEqual) = bad
Random stuff that happens to panic but doesn't do anything useful = good justification to break the good code.


Looking forward to seeing a (better) fix for #4373 which again ... would have been a lot better to keep the previous fix until a better one rather than revert for the sake of... keeping something useless panicking at the cost of something useful  not working

PR #4367, opened 2024-07-26
reflect: rawFieldByNameFunc: copy index slice to avoid later overwrites

dgryski:


lpereira:
In #4287, there's a [commit implementing `FieldByNameFunc()` in terms of `VisibleFields()` (and implementing `FieldByName()` interms of `FieldByNameFunc()`)](https://github.com/tinygo-org/tinygo/pull/4287/commits/a3c5eb5c0bb4854fd99852285e36d795896f0648).  It removes over 120 lines of code, simplifying the code quite a bit.  Is there any downside is using that version of the code instead?

dgryski:
Because `FieldByName` is able to short-circuit walking the struct while `VisibileFields` necessarily needs to process every single field.

lpereira:
Ah, makes sense. Yeah, we can clean that up later. I'm rebasing #4287 on this patch then.

dgryski:
Commit merged as part of https://github.com/tinygo-org/tinygo/pull/4287

PR #4368, opened 2024-07-28
transform: use thinlto-pre-link passes

aykevl:
This improves compilation time by about 5% in my quick test, while increasing binary size on average by  0.13% when comparing the smoke tests in the drivers repo (and about two thirds of that 0.13% is actually caused by a single smoke test).

I think this is still a good idea because it aligns the TinyGo optimization sequence with what the ThinLTO passes (in the linker) expect.


dgryski (approved):
LGTM

PR #4369, opened 2024-07-28
reflect: move binary flag into map type

aykevl:
The map type had a byte of padding ready for such a flag (on all systems except AVR). So it's basically free to use.

I want to use the now-free flag bit in the meta byte in a future change, this just lays the groundwork.

### DIFF
--- a/src/reflect/type.go
+++ b/src/reflect/type.go
@@ -30,8 +30,9 @@
 //     elem         *typeStruct // element type of the array
 //     arrayLen     uintptr     // length of the array (this is part of the type)
 //     slicePtr     *typeStruct // pointer to []T type
-// - map types (this is still missing the key and element types)
+// - map types
 //     meta         uint8
+//     extraMeta    uint8
 //     nmethods     uint16 (0)
 //     ptrTo        *typeStruct
 //     elem         *typeStruct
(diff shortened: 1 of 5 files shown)


aykevl:
There's a very small binary size increase, probably because the reflect code changed a bit. I don't see an increase in the type metadata itself.

aykevl:
Actually I _don't_ need this. But it might be useful in the future I guess?

aykevl:
Actually this would still be useful. I think I can save some binary size this way.

src/reflect/type.go
```diff
@@ -433,10 +438,11 @@ type arrayType struct {
 
 type mapType struct {
 	rawType
-	numMethod uint16
-	ptrTo     *rawType
-	elem      *rawType
-	key       *rawType
+	extraFlags uint8
```
dgryski, line 441: All the other types that have an `elem` pointer match up with the `elemType` base struct (I suppose I could have embedded it, but I didn't ...).  Can you add the `extraFlags` field there as well as to the other elemType structs?

dgryski (approved):
Small tweak for consistency, but otherwise LGTM.

src/reflect/type.go
aykevl: Oh, I think you are right. This would be a problem if the struct doesn't have any padding (like on AVR - not a big platform but still kind of a lurking bug). I'll look into a fix.

dgryski:
Still picking this up?  On hold until it's actually needed?

dgryski:
Ping

aykevl:
I don't think it's needed right now. Perhaps later. The change as-is doesn't actually improve anything. Closing for now.

PR #4370, opened 2024-07-28
reflect: return correct name for unsafe.Pointer type

aykevl:
For some reason, the type kind name is "unsafe.Pointer" while the .Name() method just returns "Pointer".

Not sure why this difference exists, but to be able to test .Name() in testdata/reflect.go it needs to match.


dgryski (approved):
LGTM

dgryski (approved):
LGTM

PR #4371, opened 2024-07-29
mips: add GOMIPS=softfloat support

aykevl:
Previously, the compiler would default to hardfloat. This is not supported by some MIPS CPUs.

This took me much longer than it should have because of a quirk in the LLVM Mips backend: if the target-features string is not set (like during LTO), the Mips backend picks the first function in the module and uses that. Unfortunately, in the case of TinyGo this first function is `llvm.dbg.value`, which is an LLVM intrinsic and doesn't have the target-features string.
I fixed it by adding a `-mllvm -mattr=` flag to the linker.

### DIFF
--- a/builder/builder_test.go
+++ b/builder/builder_test.go
@@ -57,8 +57,10 @@ func TestClangAttributes(t *testing.T) {
 		{GOOS: "linux", GOARCH: "arm", GOARM: "6"},
 		{GOOS: "linux", GOARCH: "arm", GOARM: "7"},
 		{GOOS: "linux", GOARCH: "arm64"},
-		{GOOS: "linux", GOARCH: "mips"},
-		{GOOS: "linux", GOARCH: "mipsle"},
+		{GOOS: "linux", GOARCH: "mips", GOMIPS: "hardfloat"},
+		{GOOS: "linux", GOARCH: "mipsle", GOMIPS: "hardfloat"},
+		{GOOS: "linux", GOARCH: "mips", GOMIPS: "softfloat"},
+		{GOOS: "linux", GOARCH: "mipsle", GOMIPS: "softfloat"},
 		{GOOS: "darwin", GOARCH: "amd64"},
 		{GOOS: "darwin", GOARCH: "arm64"},
 		{GOOS: "windows", GOARCH: "amd64"},
@@ -68,6 +70,9 @@ func TestClangAttributes(t *testing.T) {
 		name := "GOOS=" + options.GOOS + ",GOARCH=" + options.GOARCH
 		if options.GOARCH == "arm" {
 			name += ",GOARM=" + options.GOARM
+		}
+		if options.GOARCH == "mips" || options.GOARCH == "mipsle" {
+			name += ",GOMIPS=" + options.GOMIPS
 		}
 		t.Run(name, func(t *testing.T) {
 			testClangAttributes(t, options)
--- a/builder/builder_test.go
+++ b/builder/builder_test.go
@@ -64,7 +64,6 @@ func TestClangAttributes(t *testing.T) {
 		{GOOS: "windows", GOARCH: "amd64"},
 		{GOOS: "windows", GOARCH: "arm64"},
 		{GOOS: "wasip1", GOARCH: "wasm"},
-		{GOOS: "wasip2", GOARCH: "wasm"},
 	} {
 		name := "GOOS=" + options.GOOS + ",GOARCH=" + options.GOARCH
 		if options.GOARCH == "arm" {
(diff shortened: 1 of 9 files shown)


aykevl:
Actually I want to change the code a little bit, so converted to draft.

aykevl:
Ok, now it's ready (and hopefully it passes!). I've refactored the code a bit to avoid some unnecessary complications so that it is easier to add subarch support (GOARM, GOMIPS, GO386, etc).

clabby:
Trying this out at the moment (thank you for working on BE MIPS32 support!)

Noticing that the resulting MIPS32 BE binary includes some MIPS64 instructions, some from r6 (it looks like we're only enabling r2 in the features you pass to LLVM, so unsure why they're landing in there).

Host OS: MacOS Sonoma 14.5
TinyGo Rev: `d8cbd0ccd5074b519376e481bcc00a7a3e9d38a4`
LLVM version: 18.1.8
Build command: `GOOS=linux GOARCH=mips GOMIPS=softfloat tinygo build -o hello.elf -scheduler=none -gc=leaking`
Example binary:
```go
package main

import (
	"os"
	"strconv"
)

func main() {
	a := [][]byte{}
	for i := 0; i < 1000; i++ {
		b := doWork(byte(i % 256))
		a = append(a, b)
	}

	b := len(a)
	_, _ = os.Stdout.Write([]byte("done with work, len(b) = " + strconv.Itoa(b) + "\n"))
}

func doWork(x byte) []byte {
	b := make([]byte, 4096)
	for i := 0; i < len(b); i++ {
		b[i] = x
	}
	return b
}
```

aykevl:
> Noticing that the resulting MIPS32 BE binary includes some MIPS64 instructions, some from r6 (it looks like we're only enabling r2 in the features you pass to LLVM, so unsure why they're landing in there).

Are you sure about that? Which r6 instructions do you see?

EDIT: at least qemu-mips-static doesn't complain about unknown instructions. But who knows, maybe it implements some 64-bit instructions for some reason.

builder/builder_test.go
aykevl, line 1: @dgryski I see that this file tests GOOS=wasip2, but we don't actually support GOOS=wasip2 passed via environment variables (only `-target=wasip2`).
I think the test can be removed? What do you think?

I'm seeing this issue because I added some error checking for supported GOOS/GOARCH values (not in combination, so far only whether any given GOOS or GOARCH is supported at all).

aykevl: Actually I went ahead and removed the test.
Once we actually support GOOS=wasip2 (passed as an environment variable) we can add it again.

clabby:
> > Noticing that the resulting MIPS32 BE binary includes some MIPS64 instructions, some from r6 (it looks like we're only enabling r2 in the features you pass to LLVM, so unsure why they're landing in there).
> 
> Are you sure about that? Which r6 instructions do you see?

Apologies - was mistaken, `SWC2` is not from r6, rather an extension - misread a note where it was changed in r6. From the MIPS spec:

> Coprocessor 2 is optional and may be omitted. Software may determine if Coprocessor 2 is implemented by checking the state of the C2 bit in the Config1 CP0 register. If Coprocessor 2 is implemented, the Coprocessor 2 interface instructions (BC2, CFC2, COP2, CTC2, DMFC2, DMTC2, LDC2, LWC2, MFC2, MTC2, SDC2, and SWC2) may be omitted on an instruction-by-instruction basis.

With the standard "big Go," it seems to use r1, which might be why the coprocessor instructions aren't used w/ `GOARCH=mips`? Also saw an instance of `DADDIU`, which is from MIPS64.

(Let me know if there's a better place to discuss - sorry for cluttering your PR!)

aykevl:
> With the standard "big Go," it seems to use r1, which might be why the coprocessor instructions aren't used w/ `GOARCH=mips`?

Yeah we should probably switch to mips32r1, I can do that in a separate PR (after the refactor in this PR).

> Also saw an instance of `DADDIU`, which is from MIPS64.

I don't, when I disassemble the binary. I also don't see SWC2. Where do you see it? Can you maybe give me the binary and the instruction address?

> (Let me know if there's a better place to discuss - sorry for cluttering your PR!)

I guess the original issue is a place we could discuss this.

builder/builder_test.go
dgryski: SGTM

deadprogram:
So is this PR ready to merge?

aykevl:
From my point of view, yes. It makes TinyGo a bit more compatible with older/simpler MIPS cores. A future PR will move to mips32r1 for even wider compatibility.

deadprogram:
OK, great. Thanks @aykevl now merging.

PR #4374, opened 2024-07-31
ARM softfloat support

aykevl:
This adds softfloat support to GOARM, as proposed here: https://github.com/golang/go/issues/61588

This is similar to https://github.com/tinygo-org/tinygo/pull/4189 but with a few differences:

  * It is based on the work to support MIPS softfloat.
  * It fixes the issue that the default changed to softfloat everywhere. This PR defaults to softfloat on ARMv5 and hardfloat on ARMv6 and ARMv7.
  * It also compiles the C libraries (compiler-rt, musl) using the same soft/hard floating point support. This is important because otherwise a GOARM=7,softfloat binary could still contain hardware floating point instructions.

I checked that the generated instructions (both for Go code and for the musl library) are as expected (softfloat vs hardfloat).

---

~Draft because it needs to be rebased after #4371 is merged.~

CC @MDr164 I hope you don't mind me taking over the work, but this seems important to have (and #4189 would have needed some big refactoring anyway after #4371).


MDr164:
Yeah sorry I didn't get around testing more on ARM as I was totally swamped the past weeks. Feel free to copy any code from my PR if that helps.

aykevl:
This PR is ready now, assuming CI passes.

MDr164:
> This PR is ready now, assuming CI passes.

Looks good! I assume the two boards failing in CI are not failing due to this PR. On one flashing failed and the other error also doesn't look related.

deadprogram:
Thanks for the addition @aykevl and to @MDr164 for review. Now merging.

PR #4375, opened 2024-07-31
compiler: reimplement interface type asserts

aykevl:
This is a big reimplementation that simplifies the compiler a lot. Instead of storing the method set in metadata and lowering the type asserts as a whole program pass, this change just puts the list of methods in the type code (and a separate global for the interface type).

I'm still working on this. I'm not very happy with the binary size increase (~4%) but I have a few ideas how to reduce it. But I made a PR for initial review for those interested (@dgryski?)

This is also a step in the direction of #2870.


dgryski:
Are you still thinking of working on this?

reteps:
cc @aykevl -- my assumption is no?

jakebailey:
FWIW I took a crack at this and #4376 in #5304, since this unblocks my own usage.

PR #4376, opened 2024-07-31
reflect: fully implement Type.AssignableTo

aykevl:
This fully implements AssignableTo, fixing a number of bugs in the previous implementation. In particular, it now supports AssignableTo for interface types.

Depends on #4375


transform/optimizer.go
```diff
@@ -65,7 +65,6 @@ func Optimize(mod llvm.Module, config *compileopts.Config) []error {
 
 		// Run TinyGo-specific optimization passes.
 		OptimizeStringToBytes(mod)
-		OptimizeReflectImplements(mod)
```
dgryski, line 68: So what happens with type asserts that are statically known to be true or false?  Optimization pushed for future work?

aykevl: This is specifically for the reflect `Implements` method, which is used in some important code but not really used that much in total (so I don't think optimizing it will provide much benefit). The main reason this "optimization" existed was to get the encoding/json package to work.

But yeah, good point about other (normal Go style) interface type asserts. I can do a check how common that is and whether optimizing that case is beneficial. That is something that should work well as a package level optimization though.

reteps:
@dgryski any progress on this? This is causing issues for me. I would like to use tinygo to generate WASM bindings, but the lack of a JSON encoder isn't great. See https://github.com/reteps/dockerfmt/issues/25.

progrium:
Is the dependency on #4375 a hard one? Could this be implemented without it?

jakebailey:
FWIW I took a crack at this and #4375 in #5304, since this unblocks my own usage.

aykevl:
Closing since #5304 is just a better version of this PR.

PR #4377, opened 2024-08-02
os/StartProcess 

leongross:
This is a rework of https://github.com/tinygo-org/tinygo/pull/4323 (based on that branch) to move the Execve and Fork wrappers to the `os` package to circumvent the goroot overrides.


src/os/osexec.go
```diff
@@ -0,0 +1,56 @@
...
+
+package os
+
+import (
+	"errors"
+	"syscall"
+	"unsafe"
+)
+
+func Fork() (pid int, err error) {
```
aykevl: Please unexport `Fork` and `Execve` because they aren't provided by the "big Go" os package (and are easy to misuse).

src/os/osexec.go
```diff
@@ -0,0 +1,56 @@
+//go:build linux
+// +build linux
```
aykevl: You only need `//go:build linux`, the old `// +build` lines are deprecated.
Also, you'll probably want `//go:build linux && !baremetal`.

src/os/exec_other.go
```diff
@@ -0,0 +1,21 @@
+//go:build !aix && !android && !freebsd && !linux && !netbsd && !openbsd && !plan9 && !solaris
+// +build !aix,!android,!freebsd,!linux,!netbsd,!openbsd,!plan9,!solaris
+
+package os
```
aykevl, line 3: I think you need to define `Interrupt` and `Kill` here to get the Windows and MacOS CI to pass.

src/os/exec_linux.go
```diff
@@ -0,0 +1,84 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build posix || aix || linux || dragonfly || freebsd || (js && wasm) || netbsd || openbsd || solaris || wasip1
+// +build posix aix linux dragonfly freebsd js,wasm netbsd openbsd solaris wasip1
```
aykevl: I think you can simplify this to `//go:build linux`.
Also, what about MacOS? I don't see why it wouldn't work.

aykevl (commented):
Quick initial (incomplete) review

leongross:
> main_test.go:394: /home/runner/work/tinygo/tinygo/src/os/osexec.go:14:23: undefined: syscall.Syscall

[ci](https://github.com/tinygo-org/tinygo/actions/runs/10250765856/job/28357229355?pr=4377#step:19:92) fails with undefined `syscall.Syscall`. I thought since it is just a package the compiler would pick that up when building it. @aykevl 

EDIT: it looks like this is only happening on architectures that the `createRawSyscall` function does not target, which makes sense, namely `EmulatedCortexM3 : cortex-m-qemu` and `EmulatedRISCV : riscv-qemu`. Anyways, this should not happen.
https://github.com/tinygo-org/tinygo/blob/dev/main_test.go#L522


Anyhow the smoke tests fail without any further information of what went wrong on linux, windows and mac. the smoke test cannot be removed, it is not clear though why it wasn't build in the first place

```
rm: cannot remove 'smoke.test': No such file or directory
```

leongross:
Not sure if something like https://github.com/tinygo-org/tinygo/pull/4398 would fix this? @aykevl

aykevl:
> Not sure if something like #4398 would fix this? @aykevl

No, that's the wrong approach. If `syscall.Syscall` fails on baremetal systems, then the solution is not to implement these syscalls (because they're baremetal, so there is no OS to call into), but rather to figure out _why_ these calls are made.

src/os/osexec.go
```diff
@@ -0,0 +1,55 @@
+//go:build linux
```
aykevl:
```suggestion
//go:build linux && !baremetal
```

Baremetal systems act like Linux, so you need to explicitly exclude these.

leongross: But then most of the `os` package should have that set, right? 
The CI fails for other missing implementations such as `os.Getpid` and others.
I am note sure though why this is happening here and has not happened before?

src/os/tempfile.go
```diff
@@ -17,7 +17,8 @@ var minrandMutex sync.Mutex
 func init() {
 	// Avoid getting same results on every run
 	now := time.Now()
-	seed := uint32(Getpid()) ^ uint32(now.Nanosecond()) ^ uint32(now.Unix())
+	// seed := uint32(Getpid()) ^ uint32(now.Nanosecond()) ^ uint32(now.Unix())
```
leongross: To get this PR working I needed to adjust the seeding function. @aykevl is this still fine or should we increase entropy somehow?

leongross:
@deadprogram @dgryski @aykevl build and tests finally succeed, would appreciate another review.

leongross:
@aykevl @deadprogram @dgryski would someone mind reviewing?

src/os/tempfile.go
deadprogram: @aykevl what do you think?

Also @leongross you should probably remove the commented out line.

src/os/osexec.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	argv1[len(argv)] = nil
+
+	env1 := make([]*byte, len(envv)+1)
+	for i, env := range envv {
+		env1[i] = &cstring(env)[0]
+	}
+	env1[len(envv)] = nil
+
+	// fail := libc_execve(&argv0[0], &argv1[0], &env1[0])
```
deadprogram: Any reason these lines are commented here? :smile_cat:

leongross: No, this surely can be removed.

src/os/exec.go
```diff
@@ -57,8 +61,9 @@ type Process struct {
 	Pid int
 }
 
+// Wrapper for the forkExec function, which is a wrapper around the fork and execve system calls.
```
aykevl: That's not a super descriptive comment (users don't need to know that an internal function is called `forkExec`).
Maybe [`os.StartProcess`](https://pkg.go.dev/os#StartProcess) can give an idea what to put here?

src/syscall/syscall_unix.go
```diff
@@ -1,3 +1,6 @@
+//go:build linux || unix
+// +build linux unix
```
aykevl:
```suggestion
//go:build linux || unix
```

You can remove the old `// +build` lines.

src/os/tempfile.go
aykevl: Why didn't this work?

src/syscall/syscall_unix.go
```diff
@@ -16,3 +19,8 @@ type SockaddrInet6 struct {
 	Addr   [16]byte
 	raw    RawSockaddrInet6
 }
+
+const (
+	SYS_FORK   = 57
+	SYS_EXECVE = 59
+)
```
aykevl: These can be removed, because every actual operating system we have (Linux, MacOS, Windows) doesn't use the TinyGo syscall package.
(If removing these leads to compilation errors, you're probably using them from something that's not an operating system and probably need to adjust build tags at the call site).

src/os/osexec.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"errors"
+	"syscall"
+	"unsafe"
+)
+
+func fork() (pid int, err error) {
+	// ret := libc_fork()
+	// ret, _, _ := syscall.Syscall(syscall.SYS_FORK, 0, 0, 0)
+	ret, _, _ := syscall.Syscall(57, 0, 0, 0)
```
aykevl: Don't use magic numbers. Use `syscall.SYS_FORK` instead, for example.

src/os/exec_other.go
```diff
@@ -0,0 +1,28 @@
+//go:build (!aix && !android && !freebsd && !linux && !netbsd && !openbsd && !plan9 && !solaris) || baremetal || tinygo.wasm
+// +build !aix,!android,!freebsd,!linux,!netbsd,!openbsd,!plan9,!solaris baremetal tinygo.wasm
```
aykevl:
```suggestion
```

`// +build` lines can be removed.

src/os/osexec.go
```diff
@@ -0,0 +1,55 @@
...
+
+	env1 := make([]*byte, len(envv)+1)
+	for i, env := range envv {
+		env1[i] = &cstring(env)[0]
+	}
+	env1[len(envv)] = nil
+
+	// fail := libc_execve(&argv0[0], &argv1[0], &env1[0])
+	// fail, _, _ := syscall.Syscall(syscall.SYS_EXECVE, uintptr(unsafe.Pointer(&argv0[0])), uintptr(unsafe.Pointer(&argv1[0])), uintptr(unsafe.Pointer(&env1[0])))
+	fail, _, _ := syscall.Syscall(59, uintptr(unsafe.Pointer(&argv0[0])), uintptr(unsafe.Pointer(&argv1[0])), uintptr(unsafe.Pointer(&env1[0])))
```
aykevl: Please don't use magic numbers. It looks like you can use `syscall.Exec` instead.

src/os/osexec.go
```diff
@@ -0,0 +1,55 @@
...
+	if fail < 0 {
+		// TODO: parse the syscall return codes
+		return errors.New("execve failed")
```
aykevl: Can you implement this TODO?
(If you use `syscall.Exec` you already have an error value to return!)

src/os/exec_linux.go
```diff
@@ -0,0 +1,83 @@
...
+
+// This function is a wrapper around the forkExec function, which is a wrapper around the fork and execve system calls.
+// The StartProcess function creates a new process by forking the current process and then calling execve to replace the current process with the new process.
+// It thereby replaces the newly created process with the specified command and arguments.
+// Differences to upstream golang implementation (https://cs.opensource.google/go/go/+/master:src/syscall/exec_unix.go;l=143):
+// * No setting of Process Attributes
+// * Ignoring Ctty
+// * No ForkLocking (might be introduced by #4273)
+// * No parent-child communication via pipes (TODO)
+// * No waiting for crashes child processes to prohibit zombie process accumulation / Wait status checking (TODO)
```
aykevl, line 46: Is it possible to check for these cases, and if users use any of these to return an error instead of continuing and not doing what the user expected?
For example, if `Stdin` is set, it should either implement that feature or return an error.

src/os/exec.go
```diff
@@ -47,6 +47,10 @@ func (p *ProcessState) Sys() interface{} {
 	return nil // TODO
 }
 
+func (p *ProcessState) Exited() bool {
+	return false // TODO
+}
```
aykevl, line 58: That's... a little ugly.
Is there any way we can track whether the process is still running?

aykevl (commented):
* There is some commented out code, that looks like testing code. IMHO it should either be used, or removed (it doesn't really add any value in documentation).
* You make a few changes to the syscall package, but why? All actual operating systems (Linux, macOS, Windows) use the default syscall package. If you need any changes to the syscall package, that probably indicates that it is called from somewhere (for example, baremetal) and you need to adjust build tags.

src/os/tempfile.go
leongross: I am not sure anymore what I thought the issue was there. Using the prior implementation seems to work now so I will revert this change.

src/os/osexec.go
leongross: I removed the explicit syscall number and found out, that in the CI the definition for `SYS_FORK` cannot be found. Grepping through the source tree I found out, that there is no definition for this, only the local go installation provides this, e.g. via `src/syscall/zsysnum_linux_amd64.go`. What confused me is that the CI should have picked this up but did not. On my local machine, this works perfectly fine.

src/os/exec_linux.go
leongross: That sounds like a reasonable thing to do. I will evaluate this and see how/if this is possible.

src/os/exec.go
leongross: For now, I stubbed it out the same way all the other methods over `ProcessState` (which is an empty struct) are implemented. Making this functional would mean a quite large rework of the tinygo process representation, which atm is just a pid. I think it is a good idea to rework this, but [not as part of this PR](https://github.com/tinygo-org/tinygo/issues/4471).

deadprogram:
@leongross still are some CI failures from this PR is would appear.

leongross:
I figured out the error here: there is no fork syscall on arm64 [1]. The original golang implementation abstracts this better and does not do raw syscalls as my implementation does. So for now, arm64 has to be excluded as a possible target architecture for this.

[1]: https://chromium.googlesource.com/chromiumos/docs/+/master/constants/syscalls.md#cross_arch-numbers

leongross:
Windows CI fails on, any will investigate.
```
--- FAIL: TestBuild/Host/go1.21.go (35.61s)
```

aykevl:
I think you need to drop the first commit, now that you've rebased on the dev branch?

The Windows test failures are probably just because the Windows CI tends to be flaky (sometimes, tests don't produce any output and we don't know why) so you can ignore these for now.

leongross:
Thanks for the feedback @aykevl. It looks like now just the said windows CI is failing.

leongross:
@aykevl any idea how we can fix the flaky windows CI?

src/syscall/syscall_unix.go
```diff
@@ -1,3 +1,5 @@
+//go:build linux || unix
```
aykevl, line 1:
```suggestion
//go:build unix
```

src/os/exec.go
aykevl: Oh I see, `Wait` doesn't return anything really.

src/os/osexec.go
```diff
@@ -0,0 +1,56 @@
+//go:build linux && !baremetal && !darwin && !tinygo.wasm && !arm64
```
aykevl:
```suggestion
//go:build linux && !baremetal && !tinygo.wasm && !arm64
```

It can't be both linux and darwin.

src/os/osexec.go
```diff
@@ -0,0 +1,56 @@
...
+func fork() (pid int, err error) {
+	ret, _, err := syscall.Syscall(syscall.SYS_FORK, 0, 0, 0)
+	if int(ret) != 0 {
+		errno := err.(syscall.Errno)
+		if int(errno) != 0 {
+			return -1, errno
+		}
+	}
+	return int(ret), nil
+}
```
aykevl: Can you please either implement `fork()` for arm64, or just use the libc function?

For example (this code is entirely untested but something like that should work):

```suggestion
func fork() (pid int32, err error) {
	pid = libc_fork()
	if pid != 0 {
		err = syscall.Errno(*libc_errno())
	}
	return
}

//export fork
func libc_fork() int32

// Internal musl function to get the C errno pointer.
//export __errno_location
func libc_errno() *int32
```

src/os/exec_linux_test.go
```diff
@@ -0,0 +1,32 @@
...
+// Test the functionality of the forkExec function, which is used to fork and exec a new process.
+// This test is not run on Windows, as forkExec is not supported on Windows.
+// This test is not run on Plan 9, as forkExec is not supported on Plan 9.
+func TestForkExec(t *testing.T) {
+	if runtime.GOOS != "linux" {
+		t.Logf("skipping test on %s", runtime.GOOS)
+		return
+	}
+
+	proc, err := StartProcess("/bin/echo", []string{"hello", "world"}, &ProcAttr{})
```
aykevl, line 22: Can you also add a negative test, to make sure that running `/bin/non-existing-program` for example returns the expected error message?

Can you also add some tests that verify that unsupported features like `ProcAttr.Dir`, `ProcAttr.Sys`, etc will return an error as they should? (These checks probably also need to be implemented). If a program sets `Dir` for example but then the program executes in another directory without error, that's going to be weird and difficult to debug. I'd rather not have users deal with such bugs.

src/syscall/exec_unix.go
```diff
@@ -0,0 +1,7 @@
+//go:build unix
+
+package syscall
+
+// Implemented in runtime package.
+func runtime_BeforeExec()
+func runtime_AfterExec()
```
aykevl: Looks like these are never called? Why did you declare them?

src/os/osexec.go
```diff
@@ -0,0 +1,56 @@
...
+	env1 := make([]*byte, len(envv)+1)
+	for i, env := range envv {
+		env1[i] = &cstring(env)[0]
+	}
+	env1[len(envv)] = nil
+
+	ret, _, err := syscall.Syscall(syscall.SYS_EXECVE, uintptr(unsafe.Pointer(&argv0[0])), uintptr(unsafe.Pointer(&argv1[0])), uintptr(unsafe.Pointer(&env1[0])))
+	if int(ret) != 0 {
+		// errno := err.(syscall.Errno)
+		// return errno
```
aykevl: Please clean up these kinds of leftover comments.

src/os/signal/signal.go
aykevl, line 1: Please remove this file.

aykevl (commented):
Reviewed.

A number of comments I made are basically TODO items left in the code that really should have been implemented in some way, or just stuff that isn't needed and should have been removed from this PR. When submitting a PR, can you take a look at it to make sure you implement/cleanup/remove as much of this as possible? That makes reviewing a lot faster.

As for the Windows CI, that's a long standing issue and unrelated to this PR. I don't know how to fix it (I'm not running Windows myself and it's difficult and time consuming to fix issues like those).

src/syscall/exec_unix.go
leongross: These were leftover stubs back when https://github.com/tinygo-org/tinygo/pull/4426 was not merged yet.

src/os/osexec.go
leongross: Thanks for the input, this seems to be a better idea. I had a look at the execve implementation, but I think there it does not make sense to call into musl since it is the exact same thing we do here:

```c
int execve(const char *path, char *const argv[], char *const envp[])
{
	return syscall(SYS_execve, path, argv, envp);
}

```

leongross:
The CI seems to have problems finding musl files that are there although. Locally running these commands does not produce errors. @aykevl do you have any idea what is going on there? Only the smoke tests fail, the unit tests seem to work out.


https://github.com/tinygo-org/tinygo/actions/runs/11590162170/job/32267411673?pr=4377#step:10:11
https://github.com/tinygo-org/tinygo/actions/runs/11590162170/job/32267411673?pr=4377#step:9:700

leongross:
@aykevl

src/os/exec_linux.go
```diff
@@ -0,0 +1,103 @@
...
+	if attr == nil {
+		attr = new(ProcAttr)
+	}
+
+	p, err := fork()
+	pid = int(p)
+
+	if err != nil {
+		return 0, err
+	} else {
```
b0ch3nski: nitpick: redundant `else`

src/os/exec_linux.go
```diff
@@ -0,0 +1,103 @@
...
+// * Ignoring Ctty
+// * No ForkLocking (might be introduced by #4273)
+// * No parent-child communication via pipes (TODO)
+// * No waiting for crashes child processes to prohibit zombie process accumulation / Wait status checking (TODO)
+func forkExec(argv0 string, argv []string, attr *ProcAttr) (pid int, err error) {
+	if argv == nil {
+		return 0, errors.New("exec: no argv")
+	}
+
+	if len(argv) == 0 {
```
aykevl, line 52: Nitpick: with this check, the previous check is unnecessary. `len(argv)` of a nil slice is always 0.

src/os/os.test
aykevl, line 1: This file needs to be removed.

src/os/osexec_aarch64.go
```diff
@@ -0,0 +1,18 @@
+//go:build linux && !baremetal && !darwin && !tinygo.wasm && aarch64
```
aykevl: `aarch64` is not a valid Go build tag.

I think this file can be removed entirely now, since these things are implemented in osexec.go?

aykevl (commented):
Just a few small things, otherwise this PR seems ready for merging.

src/os/osexec_aarch64.go
```diff
@@ -0,0 +1,18 @@
...
+
+// arm64 does not have a fork syscall, so ignore it for now
+// TODO: add support for arm64 with clone or use musl implementation
```
aykevl, line 4: This comment needs to be updated or removed since it is outdated.

src/os/osexec.go
```diff
@@ -0,0 +1,61 @@
+//go:build linux && !baremetal && !tinygo.wasm
+
+// arm64 does not have a fork syscall, so ignore it for now
+// TODO: add support for arm64 with clone or use musl implementation
```
aykevl: This comment needs to be updated or removed since it is outdated.

aykevl:
Thank you! I have merged the PR.

src/os/exec_linux.go
b0ch3nski: I think this was left behind and got merged :slightly_smiling_face:

aykevl: Yeah, it wasn't a blocker for me. It's not a bug, just code that could have been cleaner.

leongross:
Thanks for merging and the thorough review @aykevl!

PR #4378, opened 2024-08-02
runtime: add support for os/signal

aykevl:
This adds support for enabling and listening to signals on Linux and MacOS.

~TODO: also support disabling signals.~

---

Not tested on MacOS yet. CI will show whether this works or not.
Also, this will need a few changes to not run the signal test systems that don't support it: Windows, WebAssembly and baremetal.


aykevl:
Fixed support for MacOS. It needed an implementation of `syscall.Kill` which was simple to add.

src/runtime/runtime_unix.go
```diff
@@ -256,3 +256,41 @@ func growHeap() bool {
 	setHeapEnd(heapStart + heapSize)
 	return true
 }
+
+func init() {
+	// Set up a channel to receive signals into.
+	// A channel size of 1 should be sufficient in most cases, but using 4 just
+	// to be sure.
+	signalChan = make(chan uint32, 4)
```
leongross: Could you elaborate on that? Could you tell me what the requirements/context are for this?

src/runtime/runtime_unix.go
```diff
@@ -256,3 +256,41 @@ func growHeap() bool {
...
+}
+
+var signalChan chan uint32
+
+//go:linkname signal_enable os/signal.signal_enable
+func signal_enable(s uint32) {
```
leongross, line 372: I think it is a good idea to add the remaining signal functions (`signal_disable`, `signal_ignore`, `signal_ignored`) golang provides as well, or at least stub them out. I remember building applications relying on the signal packages and often requiring most of them. 

Also functions like `signal.Ignore(.)` can easily be added as public API wrappers for the same functionality implemented in the runtime.

aykevl:
I'm working on extending this PR to include all remaining signal features, but it's a bit harder than I expected to do correctly without race conditions.

leongross:
> but it's a bit harder than I expected to do correctly without race conditions

how do you test for that systematically to detect them? Is there already a testing setup for that or did you observe this in manual testing?

aykevl:
I don't have tests, I reason through it by reading how POSIX signals work and by knowing how `chan` works in TinyGo. And I realized that channels are probably not safe. (They are made safe for interrupts on baremetal, but I don't think there's something similar for POSIX signals - I need to check to be sure).

deadprogram:
@aykevl please resolve merge conflict... thanks!

aykevl:
This is a bit more tricky than I expected, I need to work on it a bit more to make it safe. So it's not just a matter of rebasing.

leongross:
So if this turns out to be more tricky than initially anticipated it might be a good idea to first merge some stubbed implementations and then take on the more complicated work after. We would like to get this working with u-root by the middle of September and would appreciate some quick prototyping efforts.

leongross:
@aykevl any thoughts?

aykevl:
Implemented disabling of signals.
Also fixed _some_ race conditions, but while doing that I found a few others. I'm not entirely sure how to fix those since MacOS does not implement the necessary POSIX function to do that (`sigtimedwait`), all the existing ones like `usleep` can't be used in a race-free way as far as I can see. But I've documented all that I know in comments.

aykevl:
Actually I think I thought of a way to safely use signals on MacOS: by using `sigsetjmp` and `siglongjmp`. That's kinda ugly, but as long as `sigtimedwait` isn't implemented that might be the most reasonable way to do it.

src/os/signal/signal.go
```diff
@@ -1,14 +0,0 @@
...
-
-package signal
-
-import (
-	"os"
-)
-
-// Just stubbing the functions for now since signal handling is not yet implemented in tinygo
-func Reset(sig ...os.Signal)                      {}
-func Ignore(sig ...os.Signal)                     {}
```
leongross, line 13: What is the reason you want to remove these functions? They are widely used and we need an API to interact with the runtime. I proposed a fairly basic implementation for that in #4384, maybe we can do something similar here?

```suggestion
func Ignore(sig ...os.Signal) {
	for _, s := range sig {
		signal_ignore(uint32(signum(s)))
	}
}
```

src/runtime/runtime_unix.go
```diff
@@ -307,3 +326,164 @@ func growHeap() bool {
...
+
+// Mask of signals that have been received. The signal handler atomically ORs
+// signals into this value.
+var receivedSignals uint32
+
+//go:linkname signal_enable os/signal.signal_enable
+func signal_enable(s uint32) {
+	if s >= 32 {
+		// TODO: to support higher signal numbers, we need to turn
+		// receivedSignals into a uint32 array.
```
leongross, line 375: If we want to do it this way uint64 might even be sufficient.
checking the values for `SIGRTMAX` on my system leaves me with `64`. From a quick online look up is seems to be the case for many linux systems.

testdata/signal.go
```diff
@@ -0,0 +1,42 @@
...
+
+	// Send the signal.
+	syscall.Kill(syscall.Getpid(), syscall.SIGUSR1)
+
+	time.Sleep(time.Millisecond * 100)
+
+	// Stop notifying.
+	// (This is just a smoke test, it's difficult to test the default behavior
+	// in a unit test).
+	signal.Ignore(syscall.SIGUSR1)
```
leongross, line 37: I am confused here, I thought the go API was removed by removing signal.go

src/os/signal/signal.go
aykevl: Instead of using our own implementation, we use the one in the Go standard library. That one is already well tested so as long as the runtime side is implemented correctly, the signal package should also be correct.

By removing the package entirely, loader/goroot.go will pick the one from the standard library.

src/runtime/runtime_unix.go
aykevl: Unfortunately, uint64 is not atomic on some systems such as MIPS32.
But signal numbers higher than 31 are rarely used (I think), so we can leave that for now and implement it when someone needs it.

testdata/signal.go
aykevl: It's replaced by the standard library package, see my other comment.

src/runtime/runtime_unix.go
leongross: that makes sense. I also had a deeper look into the standard signals on `signal (7)` and indeed it looks like 1-31 are the common signals which end at `SIGRTMIN`. So for now there are only 2 entirely user-definable signals available (`SIGUSR1`, `SIGUSR2`) but for basic signal support, this should suffice.

src/runtime/runtime_unix.go
```diff
@@ -307,3 +326,164 @@ func growHeap() bool {
...
+//go:linkname signal_disable os/signal.signal_disable
+func signal_disable(s uint32) {
+	if s >= 32 {
+		// TODO: to support higher signal numbers, we need to turn
+		// receivedSignals into a uint32 array.
+		runtimePanicAt(returnAddress(0), "unsupported signal number")
+	}
+	tinygo_signal_disable(s)
+}
+
```
leongross, line 405: We might also want to add [signal_ignored](https://cs.opensource.google/go/go/+/master:src/os/signal/signal_unix.go;drc=a9bd84e03786e8bd4bfb3780ce3283b0f3e1f3bf;bpv=0;bpt=1;l=18). This could implemented via a bitmap as well.

```suggestion
func signal_ignored(s uint32) 
```

aykevl: Do you have a code sample that requires this?
The test includes `signal.Ignore(...)` call, which doesn't result in `signal_ignored` being called.

leongross: I had something in the back of my mind but I had a look at my build logs and could not find it, so let's leave it for now.

leongross:
@aykevl I think the CI failure is just racy, maybe re-run it? Also, is there anything I can further help on?

aykevl:
Yeah that's an unrelated failure. TinyHCI does fail sometimes for no apparent reason.

src/runtime/runtime_unix.go
```diff
@@ -217,8 +222,22 @@ func nanosecondsToTicks(ns int64) timeUnit {
...
+	// Check for incoming signals.
+	if checkSignals() {
+		// Received a signal, so there's probably at least one goroutine that's
+		// runnable again.
+		return
+	}
+
+	// TODO: there is a race condition here. If a signal arrives between
+	// checkSignals() and usleep(), the usleep() call will not exit early so the
+	// signal is delayed until usleep finishes or another signal arrives.
```
leongross: Iirc we talked about this issue and the only way to get this working asynchronously is by adding a second listening thread/go routine which is not possible at the moment, right? So for the initial support, we should not see this as a blocking factor.

src/runtime/runtime_unix.go
```diff
@@ -307,3 +326,164 @@ func growHeap() bool {
...
+			// Could not send the signal number on the channel. This means
+			// there's still a signal pending. In that case, let it be received
+			// at which point checkSignals is called again to put the next one
+			// in the channel buffer.
+			return gotSignals
+		}
+
+		// Atomically clear the signal number from receivedSignals.
+		// TODO: use atomic.Uint32.Or once we drop support for Go 1.22 instead
+		// of this loop.
```
leongross, line 488: Since Go 1.22 support will not be dropped in the next view releases, we should also not consider this a blocking factor here.

src/runtime/runtime_unix.go
```diff
@@ -307,3 +326,164 @@ func growHeap() bool {
...
+// void tinygo_signal_handler(int sig);
+//
+//export tinygo_signal_handler
+func tinygo_signal_handler(s int32) {
+	// This loop is essentially the atomic equivalent of the following:
+	//
+	//   receivedSignals |= 1 << s
+	//
+	// TODO: use atomic.Uint32.And once we drop support for Go 1.22 instead of
+	// this loop.
```
leongross, line 438: > Since Go 1.22 support will not be dropped in the next view releases, we should also not consider this a blocking factor here.

Same here

src/runtime/runtime_unix.go
```diff
@@ -307,3 +326,164 @@ func growHeap() bool {
...
+	// Function called from os/signal to get the next received signal.
+	val := <-signalChan
+	checkSignals()
+	return val
+}
+
+// Atomically find a signal that previously occured and send it into the
+// signalChan channel. Return true if at least one signal was delivered this
+// way, false otherwise.
+func checkSignals() bool {
```
leongross, line 460: It looks like the return code for checkSignals has not been really checked. Is this on purpose or by accident?

leongross:
I have trouble running the tests locally, the test always fails with a segfault. This does not look to me as the mentioned race condition:

```
$ tinygo test os/signal 
FAIL    os/signal       0.000s
# internal/testpty
../lib/musl/include/fcntl.h:22:10: fatal: 'bits/fcntl.h' file not found
/usr/local/go/src/internal/testpty/pty_cgo.go:11:10: note: in file included from pty_cgo.go!cgo.c:4:
```

```
$ tinygo run testdata/signal.go
panic: runtime error at 0x0000000000000007: caught signal SIGSEGV
```

src/runtime/runtime_unix.go
aykevl: Correct, all it will do is make the code look better (and _maybe_ a bit faster though that's probably not noticeable). I intentionally put in "Go 1.22" because when dropping support for a Go version I typically will grep for such strings to fix cases like these.

src/runtime/runtime_unix.go
aykevl: It is used in `sleepTicks`, where it is used to exit early before sleeping.

leongross:
What do you think about https://github.com/tinygo-org/tinygo/pull/4508 ?

aykevl:
I have updated the PR and made signals non-racy on Linux (I hope!).

* Using `sigtimedwait` on Linux for sleeping. Fall back to a racy `usleep` on MacOS (there's not much we can do about that).
* Using `sigwait` in `waitForEvents`. While `sigsuspend` would work just fine, `sigwait` is more similar to `sigtimedwait` and so keeps the code hopefully a bit easier to follow.

Apart from some minor things your PR looked good, but I wanted to make sure the other race condition was also fixed at the same time.

leongross:
@aykevl nix-ci seems to be broken due to llvm on [other branches as well](https://github.com/tinygo-org/tinygo/actions/runs/11291350989/job/31405090798?pr=4498#step:8:107)

leongross:
@aykevl please rebase to fix the CI errors.

EDIT: Updating the chromium driver go package seems to fix the wasm CI errors:
* https://github.com/tinygo-org/tinygo/pull/4530
* https://github.com/tinygo-org/tinygo/pull/4531

aykevl:
Rebased to fix the Nix CI error.
I think the Chrome issue is unrelated (this PR doesn't modify anything related to wasm) and it's flaky, so with a bit of luck it will work this time.

leongross (approved):
nano and pico CI failed due to not being available. Either re-run the CI or let's merge it since the supported platforms test ran successfully.

src/runtime/signal.c
```diff
@@ -0,0 +1,89 @@
+//go:build none
+
+// Ignore the //go:build above. This file is manually included on Linux and
+// MacOS to provide os/signal support.
```
ydnar, line 4: Can this just have `//go:build linux || darwin`?

aykevl: Unfortunately not. Using it as a regular C file would imply CGo, which forces an import of the syscall package (not needed in TinyGo but that's how the Go toolchain works), which then imports the runtime package. So doing it this way avoids a circular dependency.

The same is true for runtime_unix.c (which is included in the same way).

deadprogram:
I think all feedback here has now been addressed or commented on.

Thank you @aykevl for the implementation and to @leongross for providing so much good feedback. Also @ydnar :100: 

Now merging.

leongross:
Awesome that we got this finally merged, thanks, everyone!

PR #4380, opened 2024-08-04
wasm: add `recover()` support

### LINKED ISSUE #2914: Implement recover for wasm architecture
`recover` was implemented for most architectures in #2331, however `wasm` was left out.

As suggested here https://github.com/tinygo-org/tinygo/blob/4c647847247316bae553f9b9a8c7da4034edd199/compiler/defer.go#L29-L33 it should probably utilize the proposed exception handling: https://github.com/WebAssembly/exception-handling

Opened as per expected closure of #891

aykevl:
This fixes https://github.com/tinygo-org/tinygo/issues/2914

Add unwinding and recover support for wasm using WebAssembly exception handling. This still has a few gotchas:

  * Many WASI systems don't support exception handling yet, so this is only enabled by default in `-target=wasm` (for browsers and Node.js).
    For example, see:  https://github.com/bytecodealliance/wasmtime/issues/2049
  * Asyncify doesn't support wasm exception handling:
    https://github.com/WebAssembly/binaryen/issues/4470
    This means it's not possible to use goroutines together with panic/recover.
  * The current way that exceptions are implemented pretend to be C++ exceptions, but work slightly differently. If C++ code is called (for example through CGo) that raises an exception, that exception will be eaten by TinyGo and not be propagated. This is fixable, it just hasn't been implemented (because we don't actually support C++ right now).

I hope that these issues will be resolved over time. At least for now, people who need `recover()` have a way to use it.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -747,6 +747,7 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 			}
 			ldflags = append(ldflags, "-mllvm", "-mcpu="+config.CPU())
 			ldflags = append(ldflags, "-mllvm", "-mattr="+config.Features()) // needed for MIPS softfloat
+			ldflags = append(ldflags, "-mllvm", "-wasm-enable-eh")
 			if config.GOOS() == "windows" {
 				// Options for the MinGW wrapper for the lld COFF linker.
 				ldflags = append(ldflags,
--- a/compiler/calls.go
+++ b/compiler/calls.go
@@ -70,21 +70,25 @@ func (b *builder) createRuntimeInvoke(fnName string, args []llvm.Value, name str
 // createCall creates a call to the given function with the arguments possibly
 // expanded.
 func (b *builder) createCall(fnType llvm.Type, fn llvm.Value, args []llvm.Value, name string) llvm.Value {
-	expanded := make([]llvm.Value, 0, len(args))
-	for _, arg := range args {
-		fragments := b.expandFormalParam(arg)
-		expanded = append(expanded, fragments...)
-	}
-	return b.CreateCall(fnType, fn, expanded, name)
+	return b.CreateCall(fnType, fn, b.expandFormalParams(args), name)
 }
 
 // createInvoke is like createCall but continues execution at the landing pad if
 // the call resulted in a panic.
 func (b *builder) createInvoke(fnType llvm.Type, fn llvm.Value, args []llvm.Value, name string) llvm.Value {
-	if b.hasDeferFrame() {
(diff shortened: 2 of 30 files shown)


deadprogram:
@aykevl please note merge conflict here.

aykevl:
Fixed the merge conflicts!

dgryski (commented):
I can't review all the LLVM stuff, but I just want to confirm that there should be no differences on non-wasm platforms here.

aykevl:
It would be useful to have this merged, because this change will conflict with the "recover from runtime panics" change I'd like to make.

If there's any way I can help to review this (explain what it does or improve comments for example) please let me know.

dgryski:
I think this is fine to merge.  I've read through it and everything looks fine from a high level.  But obviously any bugs are goingn to be in the details.  Any way we could make a fuzzer to tease out any issues?  Or any additional tests we could run/import from upstream Go to test these?  (I guess the fact that it already passes the existing panic/recover tests we have is probably sufficient..)

aykevl:
Once we have a way to combine `recover()` with goroutines we could run a bunch of tests that depend on having `recover()` available (for example, the tests currently disabled in `make tinygo-test-wasip1`). But until that, I'm not sure. But as you said, we do have panic/recover tests and they do pass so that should count for something :)

deadprogram:
It has a lot of limitations, to be sure, but it is at least some progress. Anyone has opinion on this before merge?

deadprogram:
@aykevl can you please resolve merge conflicts in this PR?

deadprogram:
@aykevl

deadprogram:
@aykevl what do you think about completing this PR? :smile_cat:

dgryski:
What's the status of this PR (aside from all the merge conflicts)?

Edit: Seems like "waiting on support in wasmtime at least".  And for those who want to follow along at home: https://github.com/bytecodealliance/wasmtime/issues/1677#issuecomment-2827482734

aykevl:
It needs two things:

* Some place we can test it. We could use NodeJS for that I guess until wasmtime is ready. Since wasmtime support doesn't seem to be far off, we could wait for that.
* One of these:
  * exception handling support in Asyncify (see https://github.com/WebAssembly/binaryen/pull/5475 for the latest attempt).
  * Threading support as an alternative to asyncify (e.g. through Web Workers in browsers and NodeJS).
  * Disabling goroutine support. Good enough for a small test, but not feasible as a default option and won't support all the standard library tests.
  * Maaaybe something in wasip3 that we can use for goroutines?

jakebailey:
I'm going to close this since #5550 just got merged.

PR #4381, opened 2024-08-04
ci: don't include prebuilt libraries in the release

aykevl:
These libraries will be automatically built when needed and cached.

The main reason these were needed is for play.tinygo.org, but I've now prebuilt them there directly (so they don't need to be built for every tarball).


dgryski (approved):
LGTM

PR #4383, opened 2024-08-05
main: show runtime panic addresses for `tinygo run` and `tinygo test`

aykevl:
This adds the same panic locations that are already present for `tinygo flash -monitor`, but for `tinygo run` and `tinygo test`.

For example, this is the output that I get while working on some GC code. It now shows the source location instead of just an address:

    $ tinygo test -v archive/zip
    === RUN   TestReader
    === RUN   TestReader/test.zip
    panic: runtime error at 0x000000000024d9b4: goroutine stack overflow
    [tinygo: panic at /home/ayke/src/tinygo/tinygo/src/internal/task/task_stack.go:58:15]
    FAIL    archive/zip     0.139s

(This particular location isn't all that useful, but it shows that the feature works).

### DIFF
--- a/main.go
+++ b/main.go
@@ -936,7 +936,7 @@ func buildAndRun(pkgName string, config *compileopts.Config, stdout io.Writer, c
 
 	// Configure stdout/stderr. The stdout may go to a buffer, not a real
 	// stdout.
-	cmd.Stdout = stdout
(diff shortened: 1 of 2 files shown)


dgryski (approved):
LGTM

PR #4387, opened 2024-08-06
GNUmakefile: add spellfix target, use it.

dkegel-fastly:
Also:
- point to an updated version of the misspell package.
- give internal/tools its own go.mod temporarily because new misspell needs a newer go than toplevel go.mod specifies
- make lint target depend on tools target and use the revive it installs (assuming go install causes it to be found via PATH)
- fix what newer misspell and lint find
- add spell to CI


compiler/compiler.go
```diff
@@ -242,7 +242,7 @@ func NewTargetMachine(config *Config) (llvm.TargetMachine, error) {
 }
 
 // Sizes returns a types.Sizes appropriate for the given target machine. It
-// includes the correct int size and aligment as is necessary for the Go
+// includes the correct int size and allignment as is necessary for the Go
```
dgryski: `alignment`

src/internal/wasi/clocks/v0.2.0/monotonic-clock/monotonic-clock.wit.go
```diff
@@ -78,7 +78,7 @@ func wasmimport_Resolution() (result0 uint64)
 // SubscribeInstant represents the imported function "subscribe-instant".
 //
 // Create a `pollable` which will resolve once the specified instant
-// occured.
+// occurred.
```
dgryski: Sadly this is a misspelling upstream, *and* this is a generated file.  We should probably ignore `src/internal/wasi` for now.

src/internal/wasi/io/v0.2.0/streams/streams.wit.go
```diff
@@ -413,7 +413,7 @@ func wasmimport_OutputStreamFlush(self0 uint32, result *cm.Result[StreamError, s
 //
 // Read from one stream and write to another.
 //
-// The behavior of splice is equivelant to:
+// The behavior of splice is equivalent to:
```
dgryski: I might submit spelling fixes for the original `.wit` files upstream.

dgryski (changes requested):
I'm a fan of automation.  Even better if the tooling is reliable enough to run as part of CI.

compiler/compiler.go
dkegel-fastly: unbroke

src/internal/wasi/clocks/v0.2.0/monotonic-clock/monotonic-clock.wit.go
dkegel-fastly: ignored

dkegel-fastly:
Thanks for the review!  Ready for another look.

dgryski (approved):
LGTM

go.mod
```diff
@@ -1,14 +1,16 @@
 module github.com/tinygo-org/tinygo
 
-go 1.19
+go 1.21
```
b0ch3nski: I think I saw somewhere on tinygo.org docs mentioning a requirement of Go 1.20 - would be good to confirm that and also bump the version in docs

archie2x (approved):
LGTM

go.mod
leongross, line 1: Ci [fails](https://github.com/tinygo-org/tinygo/actions/runs/10288011952/job/28472435670?pr=4387#step:19:39) with
```
go: updates to go.mod needed; to update it:
	go mod tidy
```
So update go modules and/or rebase against dev?

leongross (changes requested):

go.mod
dkegel-fastly: I suspect separate go.mod to specify a newer go for now, to be removed when top level go.mod catches up

go.mod
dkegel-fastly: Not yet!  I think we still work with go 1.19.  So the proper workaround is probably a separate go.mod just for the spell checker for the moment.

dgryski (approved):
LGTM

PR #4388, opened 2024-08-06
compileopts: add CanonicalArchName to centralize architecture detection

aykevl:
It's possible to detect the architecture from the target triple, but there are a number of exceptions that make it unpleasant to use for this purpose. There are just too many weird exceptions (like mips vs mipsel, and armv6m vs thumv6m vs arm64 vs aarch64) so it's better to centralize these to canonical architecture names.

I picked the architecture names that happen to match the musl architecture names, because those seem the most natural to me.


dgryski (approved):
LGTM

PR #4389, opened 2024-08-07
duplicate big-go version tags functionality [ fix #4391 ]

### LINKED ISSUE #4391: tinygo version build-tags (duplicate go1.X build-tag functionality)
## Background
You can conditionally compile go source files based on go versions with `//go:build` lines

require go >= v1.22 `//go:build go1.22`
require go < v1.22 `//go:build !go1.22`

Ideally *tinygo* would offer all the same functionality of *big-go* and this mechanism would suffice.

## Motivation
*tinygo* offers a subset of the functionality of big-go (rapidly closing gap!)
- it can be difficult to manage downstream projects with release/dev versions of *tinygo*
- work-arounds are temporary and have to be edited to track *tinygo* release

## Example

While porting [u-root](https://github.com/u-root/u-root), all files for all commands that fail to link with release *tinygo* (currently *0.32.0*) have been marked with `//go:build !tinygo && (...)`

But, many of these commands now link with *tinygo-dev* branch, whose version predicts the next release in the (current) form *0.33.0-dev-SHA*

Right now, support for using *tinygo*(-dev) with these commands is being managed by developers with branches.

One solution is to make a more complicated build line and supply the build tags directly, like: `//go:build (!tinygo || tinygodev) && (...)` and build with `tinygo build -tags tinygodev`

### Problem
Once the next version of *tinygo* is released, above solution requires modifying source and/or flow to accommodate.

## Proposal
I'm proposing *tinygo* mimic/offer the big-go version build-tags behavior with its own build tags so that we can track its growing support of the big-go functionality. Developers can write `//go:build` lines that won't have to necessarily be changed as functionality is added to *tinygo*.

`//go:build (!tinygo || tinygo0.33) && (...)`

This solution allows specifying a minimum *tinygo* version, does not require defining custom build-tags, and does not need to be modified once `tinygo v0.33.0` is actually released assuming functionality that supported the build isn't removed.

### Additional Functionality
Have a way to export the generated version build-tags so that dependent tools can use them. 

- *big-go* does not do this so *tinygo* generates the go0.X tags itself.
- gobusybox could be modified to use *tinygo* as its compiler, but use big-go tools, to do so, it needs to forward all the build tags of tinygo (see u-root/gobusybox#121)

### Additional Thought
*tinygo-dev* branch should have a *tinygodev* build tag?

Anyway: supplying build tags, particularly version, only seems like a "win" and doesn't really have any drawback that I can think of.

archie2x:
[ see #4391 for motivation ]
This attempts to duplicate big-Go go1.X build tags in tinygo with a series of build tags in the form `tinygo0.1,tinygo0.2,...`

Also adds a '-full' command-line flag to see all of the tags.

```
% ./build/tinygo info -json -full
{
  "goarch": "arm64",
  "goarm": "6",
  [...]
  "build_tags": [
    "darwin",
    "arm64",
    "tinygo",
    "purego",
    "math_big_pure_go",
    "gc.precise",
    "scheduler.tasks",
    "serial.none",
    "tinygo0.1",       <---
    "tinygo0.2",       <---
    [...]
    "tinygo0.32",      <---
    "tinygo0.33"       <---
  ],
  "garbage_collector": "precise",
  "scheduler": "tasks",
  "llvm_triple": "arm64-apple-macosx11.0.0"
}

```

TODO:
- tests
- Version specification is constantly confusing. I'd be inclined to create a separate package for clarity s.t. the constants were of the form`tinygoVersion.Major` etc. instead of goenv.VersionMajor etc.
- define `tinygo-dev` build tag if "-dev" ?
```
package tinygoVersion

const Major = 0
const Minor = 33
const Patch = 0
const Dev = "-dev" // "" for release
```

### DIFF
--- a/builder/config.go
+++ b/builder/config.go
@@ -33,14 +33,15 @@ func NewConfig(options *compileopts.Options) (*compileopts.Config, error) {
 		return nil, fmt.Errorf("requires go version 1.19 through 1.22, got go%d.%d", major, minor)
 	}
 
+	// tinygo major / minor version
+	tgMajor, tgMinor := goenv.GetVersion()
+
 	return &compileopts.Config{
 		Options:        options,
 		Target:         spec,
 		GoMinorVersion: minor,
 		TestConfig:     options.TestConfig,
-
-		// tinygo version
-		VersionMajor:   goenv.VersionMajor,
-		VersionMinor:   goenv.VersionMinor,
+		VersionMajor:   tgMajor,
+		VersionMinor:   tgMinor,
 	}, nil
 }
--- a/builder/config.go
+++ b/builder/config.go
@@ -38,5 +38,9 @@ func NewConfig(options *compileopts.Options) (*compileopts.Config, error) {
 		Target:         spec,
 		GoMinorVersion: minor,
 		TestConfig:     options.TestConfig,
+
+		// tinygo version
+		VersionMajor:   goenv.VersionMajor,
+		VersionMinor:   goenv.VersionMinor,
 	}, nil
 }
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -20,6 +20,10 @@ type Config struct {
 	Target         *TargetSpec
 	GoMinorVersion int
 	TestConfig     TestConfig
+
+	// tinygo versions
+	VersionMajor int
+	VersionMinor int
 }
 
 // Triple returns the LLVM target triple, like armv6m-unknown-unknown-eabi.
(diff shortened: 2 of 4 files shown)


aykevl:
I'm not entirely convinced we should be doing this. Can you explain why this is needed (and why you can't just use the latest tinygo for example)?

archie2x:
> I'm not entirely convinced we should be doing this. Can you explain why this is needed (and why you can't just use the latest tinygo for example)?

Thanks! Sorry: it was an early-morning PR. See Issue #4391 where I tried to illustrate the motivation.

archie2x:
As mentioned in #4391, I agree this doesn't make sense for now.

PR #4398, opened 2024-08-08
support for riscv syscalls

leongross:
Intention to add riscv syscall support to tinygo. I have no prior experience with porting tinygo to another architecture, also I am not sure if all the efforts I am making here are required for that. It is an experiment. 

I am orienting myself at https://github.com/tinygo-org/tinygo/pull/4308.


aykevl:
@leongross would you like to see RISC-V supported in TinyGo?
I can take a look at adding it, it shouldn't be too difficult since we already have baremetal RISC-V supported.

leongross:
I think this could be very promising but for the projects I currently work on this has no priority. I was just experimenting around with it.

PR #4209, opened 2024-03-27
Add os/file.Truncate

leongross:
Add `file.Truncate()` to `file.go`

PR #4401, opened 2024-08-09
add os.Truncate(name, size) (see #4209)

archie2x:
See https://pkg.go.dev/os#Truncate

@leongross authored #4209 which defined the *os.File receiver Truncate but was missing os.Truncate(filename, size).


dgryski (approved):
LGTM

deadprogram:
Now merging, thank you @archie2x for the addition and to @dgryski for review.

PR #4402, opened 2024-08-12
reflect, runtime: remove *UnsafePointer wrappers for functions

aykevl:
This was needed in the past because LLVM used typed pointers and there was a mismatch between pointer types. But we've dropped support for typed pointers a while ago so now we can remove these wrappers.

This is just a cleanup, it shouldn't have any practical effect.


deadprogram (approved):

deadprogram:
Thanks for the cleanup on this @aykevl now merging.

PR #4403, opened 2024-08-12
unix: print a message when a fatal signal happens

aykevl:
Print a message for SIGBUS, SIGSEGV, and SIGILL when they happen. These signals are always fatal, but it's very useful to know which of them happened.

Also, it prints the location in the binary which can then be parsed by `tinygo run` (see https://github.com/tinygo-org/tinygo/pull/4383).

While this does add some extra binary size, it's for Linux and MacOS (systems that typically have plenty of RAM/storage) and could be very useful when debugging some low-level crash such as a runtime bug.

---

I still need to make this work on MacOS.

### DIFF
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -346,6 +346,8 @@ func defaultTarget(options *Options) (*TargetSpec, error) {
 			"-arch", llvmarch,
 			"-platform_version", "macos", platformVersion, platformVersion,
 		)
+		spec.ExtraFiles = append(spec.ExtraFiles,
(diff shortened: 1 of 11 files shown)


aykevl:
This would also be useful for MIPS for example, so that when a SIGILL happens ("illegal instruction"), we know the location in the binary and can more easily find it to fix the bug.
CC @DJAGABATWORK

deadprogram:
Also probably of interest to @leongross and others.

dgryski (approved):
LGTM

deadprogram:

PR #4407, opened 2024-08-14
static analysis cleanups

dgryski:
A few cleanups found by some tools

### DIFF
--- a/main.go
+++ b/main.go
@@ -958,7 +958,7 @@ func buildAndRun(pkgName string, config *compileopts.Config, stdout io.Writer, c
 	err = run(cmd, result)
 	if err != nil {
 		if ctx != nil && ctx.Err() == context.DeadlineExceeded {
-			stdout.Write([]byte(fmt.Sprintf("--- timeout of %s exceeded, terminating...\n", timeout)))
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -958,7 +958,7 @@ func buildAndRun(pkgName string, config *compileopts.Config, stdout io.Writer, c
 	err = run(cmd, result)
 	if err != nil {
 		if ctx != nil && ctx.Err() == context.DeadlineExceeded {
-			stdout.Write([]byte(fmt.Sprintf("--- timeout of %s exceeded, terminating...\n", timeout)))
+			fmt.Fprintf(stdout, "--- timeout of %s exceeded, terminating...\n", timeout)
```
archie2x, line 961: I mean... why isn't that:
```go
fmt.Printf("--- timeout of %s exceeded, terminating...\n", timeout)
```
?

archie2x (approved):

main.go
dgryski: Because `stdout` in this case is only *sometimes* `os.Stdout`, and other times we wrap it in our own buffering code.

archie2x: Figured. Thanks!

deadprogram:
Thank you for the cleanup @dgryski and to @archie2x for review. Now merging.

PR #4411, opened 2024-08-14
misspell.csv: add new misspellings

dkegel-fastly:
Also:

- check in result of 'make spellfix'

- Fix 'numer of', since 'NUMER' by itself is valid abbrev for numerator.
  Oddly, misspell was not able to detect this one, even though in
  theory it allows detecting multiword typos, so the fix was done by
  hand.

- misspell.csv: add another misspelling: interrput
  Fix by hand, since spellfix is too timid to fix in non-whitespace context?

- make spellfix: fix top level files, too
  Do manual fix in GNUmakefile, since spellchecking that is just too meta


dgryski (approved):
LGTM

archie2x (approved):

deadprogram:
Thanks for the fixes @dkegel-fastly and to @dgryski and @archie2x for review. Now merging.

PR #4413, opened 2024-08-16
renesas: update Renesas svd generation

deadprogram:
This PR updates the still incomplete support for the Renesas family of processors by switching to the main `cmsis-svd` repo for the SVD data, and removing the separate repo that we had previously added ourselves.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -185,7 +185,7 @@ fmt-check: ## Warn if any source needs reformatting
 	@unformatted=$$(gofmt -l $(FMT_PATHS)); [ -z "$$unformatted" ] && exit 0; echo "Unformatted:"; for fn in $$unformatted; do echo "  $$fn"; done; exit 1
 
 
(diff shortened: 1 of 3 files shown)


GNUmakefile
```diff
@@ -234,7 +234,7 @@ gen-device-rp: build/gen-device-svd
 	GO111MODULE=off $(GO) fmt ./src/device/rp
 
 gen-device-renesas: build/gen-device-svd
-	./build/gen-device-svd -source=https://github.com/tinygo-org/renesas-svd lib/renesas-svd/ src/device/renesas/
+	./build/gen-device-svd -source=https://github.com/cmsis-svd/cmsis-svd-data/tree/master/data/Renesas lib/cmsis-svd/data/Renesas/ src/device/renesas/
```
sago35, line 237: Isn't the addition to gen-devices at line 188 necessary? Other than that, I think it's good.

deadprogram: Thanks for catching that omission @sago35 just added it.

deadprogram:
Thanks for the review and feedback @sago35 now merging.

PR #4414, opened 2024-08-16
targets: remove import-memory flag from wasm-unknown target to fix #4319

### LINKED ISSUE #4319: wasm-unknown target shouldn't set `--import-memory` by default
The `wasm-unknown` target invokes `wasm-ld` with the [`--import-memory`](https://lld.llvm.org/WebAssembly.html#cmdoption-import-memory) option, which causes the following import to be emitted in the output:

```wat
(import "env" "memory" (memory $mimport$0 2))
```

This is undesirable to have as default behavior for `wasm-unknown`, because it introduces a mandatory import into what is otherwise a freestanding module.

If the `--import-memory` were to be removed from `targets/wasm-unknown.json` ldflags, then `wasm-ld` would emit this instead:

```wat
;; Declare and export the Memory, which requires no special import binding in the host
(memory $0 2)
(export "memory" (memory $0))
```

This option appears to have been added as part of a larger change in https://github.com/tinygo-org/tinygo/pull/4125/commits/2cd966d6d577df17cfde3869ee7e0dbb4da9307f, but it's unclear from the commit message why `--import-memory` was added. Based on discussion in the PR, I think it was simply to reduce the number of bytes in an absolute bare-bones "hello world" module, which IMO is not a compelling reason to have it set by default.

deadprogram:
This PR modifies the wasm-unknown target by removing the `-import-memory` flag as discussed in #4319

### DIFF
--- a/targets/wasm-unknown.json
+++ b/targets/wasm-unknown.json
@@ -19,8 +19,7 @@
 	"ldflags": [
 		"--stack-first",
 		"--no-demangle",
-		"--no-entry",
-		"--import-memory"
+		"--no-entry"
 	],
 	"extra-files": [
 		"src/runtime/asm_tinygowasm.S"


deadprogram:
cc @jmillikin @orsinium anyone else with interest in `wasm-unknown`

orsinium (approved):
LGTM! I agree, it should not be imported. None of the other wasm compilers I know import memory by default.

> it's unclear from the commit message why --import-memory was added.

@jmillikin I think I know the reason, if you are curious :) We were testing wasm-unknown by making games for [wasm-4](https://wasm4.org/) and [gamgee](https://github.com/orsinium-labs/gamgee) (my wasm-4 re-implementation). And wasm-4 requires games to import memory because then it can control the memory layout. 

However, wasm-4 is the only runtime I know that requires the guest code to import memory. Usually, the guest is in full control of the whole memory and all host functionality is exposed through host-defined functions. Wasm-4 went with direct memory manipulation because it's a retro console, and all retro games (like GB, GBA, Sega, etc.) always worked with memory directly.

deadprogram:
Thanks for review @orsinium now merging.

PR #4417, opened 2024-08-18
darwin: replace custom syscall package with Go native syscall package

aykevl:
This required a few compiler and runtime tricks to work, but I ran a bunch of tests and it seems fine. (CI will of course do more exhaustive testing).

The main benefit here is that we don't need to maintain the darwin version of the syscall package, and reduce extra risks for bugs (because we reuse the well-tested syscall package). For example, Go 1.23 needed a bunch of new constants in the syscall package. That would have been avoided if we had used the native syscall package on MacOS.

---

I might do the same thing for wasm/js and wasm/wasi. Those might be a bit more difficult to do though, and certainly need to be in a separate PR.


aykevl:
Weird, this test fails in CI but passes for me when I try to debug this locally. I'll need to investigate this more, I think.

aykevl:
Tests are passing now! Except for a few TinyHCI tests.

deadprogram:
Restarted TinyHCI and all tests passing now. Once we resolve https://github.com/tinygo-org/tinygo/pull/4431 we can merge this one.

aykevl:
How is #4431 related to this PR? They seem entirely unrelated to me.

aykevl:
Oh, do you mean that we want #4431 merged and then make a point release?

deadprogram:
> Oh, do you mean that we want https://github.com/tinygo-org/tinygo/pull/4431 merged and then make a point release?

Yes, that is correct.

deadprogram:
Now merging, good work @aykevl thanks!

PR #4418, opened 2024-08-19
ci: update apt repo for sizediff toolchain

aykevl:
This was still at jammy (22.04), while the CI container was noble (24.04). Somehow this didn't break, but it certainly isn't ideal to install packages across Ubuntu versions!

See: #4350 

Discovered while working on the CHANGELOG.


deadprogram:
Thanks for the update @aykevl now merging.

PR #4423, opened 2024-08-20
tinygo: detect GOOS=wasip1 for relative WASI paths via config

ydnar:
Fixes running tests without `-target=wasip1`, e.g.:

```sh
GOOS=wasip1 GOARCH=wasm tinygo test ./...
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -305,13 +305,11 @@ func Test(pkgName string, stdout, stderr io.Writer, options *compileopts.Options
 			//
 			// Ex. run --dir=.. --dir=../.. --dir=../../..
 			var dirs []string
-			switch config.Options.Target {
+			switch config.Target.GOOS {
 			case "wasip1":
 				dirs = dirsToModuleRootRel(result.MainDir, result.ModuleRoot)
-			case "wasip2":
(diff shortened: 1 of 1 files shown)


dgryski (approved):
Approving for now to fix CI issues, but a better fix exists.

deadprogram:
Is this fix urgent enough to justify a patch release?

aykevl (commented):
- `Options` is what is passed on the command line or via environment variables.
- `Target` is what is loaded from targets/*.json (or constructed at runtime for Linux/MacOS/Windows and some wasm targets).

A cleaner fix would be to use `config.GOOS()`. The `Config` struct is the one source of truth given the two inputs which might conflict. (And in this case, it just returns `config.Target.GOOS` so the fix is correct).

ydnar:
I'll make a new PR that uses that API.

I think I want to understand the differences between wasip1 and wasip2 w/r/t directories, Cwd, etc. Will talk with @dgryski to understand.

I think post second patch it's worth shipping a new release.

deadprogram:
> I think post second patch it's worth shipping a new release.

OK, sounds like a plan.

PR #4425, opened 2024-08-21
mips: fix crash with GOMIPS=softfloat

aykevl:
`defer` and `GOMIPS=softfloat` together would result in a crash. This patch fixes this issue.


deadprogram:
Thanks for the quick fix on this @aykevl now merging.

PR #4428, opened 2024-08-22
mips: fix big-endian (GOARCH=mips) support

aykevl:
I made an awkward mistake, mixing up GOOS and GOARCH. So here is a fix, with an associated test.


deadprogram:
:smiley_cat: now merging thanks for the fix @aykevl

aykevl:
Next fix for big-endian support: https://github.com/tinygo-org/tinygo/pull/4430

PR #4429, opened 2024-08-22
mips: use MIPS32 (instead of MIPS32R2) as the instruction set

aykevl:
This should widen compatibility a bit, so that older CPUs can also execute programs built by TinyGo. The performance may be lower, if that's an issue we can look into implementing the proposal here: https://github.com/golang/go/issues/60072

This still wouldn't make programs usable on MIPS II CPUs, I suppose we can lower compatiblity down to that CPU if needed.

I tried setting the -cpu flag in the QEMU command line to be able to test this, but it looks like there are no QEMU CPU models that are mips32r1 and have a FPU. So it's difficult to test this.


deadprogram:
Now merging, thanks @aykevl

PR #4430, opened 2024-08-22
mips: better support big-endian MIPS

aykevl:
Support big-endian MIPS, for the most part. The only thing that's still failing is CGo support: that will need a separate fix.

### DIFF
--- a/compiler/llvmutil/llvm.go
+++ b/compiler/llvmutil/llvm.go
@@ -8,6 +8,7 @@
 package llvmutil
 
 import (
+	"encoding/binary"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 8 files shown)


aykevl:
Updated the PR to also fix the reflect package. Now, most MIPS big-endian tests succeed so I've switched the tests over to big-endian (little-endian is still tested with a single test to be sure we won't break it, see #4428).

TjeuKayim:
In tinygo v0.33.0, `GOMIPS=softfloat` segfaulted for me. But with this MR applied, it successfully compiles. I then ran the Linux binaries on both mips and mipsle hardware, from Qualcomm and MediaTek respectively, and it worked as expected! Thank you for making this possible 😃 🚀

deadprogram:
Thank you @aykevl for working on this and to @TjeuKayim for testing it out. Now merging.

PR #4431, opened 2024-08-22
tinygo: add relative and absolute --dir options to wasmtime args

ydnar:
This normalizes the `--dir` arguments passed to `wasmtime` for both `wasip1 `and `wasip2` when running `tinygo test` or `tinygo run`. The additional `--dir` and `PWD` arguments will now be reflected when using `-x`.


dgryski:
I think there is some code in the waisp1 or wasip2 branch that detects "something" (will need to double check) based on the first or last --dir option provided.  Are we breaking anything else by effectively having two names for each directory in the module path here?

ydnar:
> I think there is some code in the waisp1 or wasip2 branch that detects "something" (will need to double check) based on the first or last --dir option provided. Are we breaking anything else by effectively having two names for each directory in the module path here?

It worked for the tests I ran. Did you discover anything?

dgryski:
No, I was just wondering because I know I had issues with argument ordering.  I didn't see any failures, but I can run my larger test corpus and see what falls out.

aykevl:
@ydnar did you get any further with this PR?
You could try running the test locally (`make test GOTESTPKGS=. GOTESTFLAGS="-v -run=Build"`), maybe it'll give some more information as to what goes wrong?

ydnar:
I haven't. I'll try that!

deadprogram:
@ydnar looks like CI still failing here:
https://github.com/tinygo-org/tinygo/actions/runs/10874218316/job/30180223920?pr=4431#step:19:33

Possible timeout?

ydnar:
Looks like a logic error on my part. Fixed!

I wanted `-x` to accurately reflect the arguments passed to `wasmtime`, but I initially failed to account for the difference between running `go test` and `go run` in the section where I moved the `--dir` arguments to. The last commit fixes this.

deadprogram:
@ydnar LGTM do you want to squash these commits yourself, or would you prefer that I do it for you?

ydnar:
Is squash merge okay?

deadprogram:
> Is squash merge okay?

Done. Thank you very much!

PR #4432, opened 2024-08-23
fix: add missing Truncate() function stub to os/file for bare-metal systems

deadprogram:
This PR adds the missing `Truncate()` function stub to `os/file` for bare-metal systems. Discovered when attempting to compile the latest Wazero with the latest TinyGo.

### DIFF
--- a/src/os/file_other.go
+++ b/src/os/file_other.go
@@ -135,3 +135,17 @@ func Readlink(name string) (string, error) {
 func tempDir() string {
 	return "/tmp"
 }
+
+// Truncate is unsupported on this system.
+func Truncate(filename string, size int64) (err error) {
+	return ErrUnsupported
+}
+
+// Truncate is unsupported on this system.
+func (f *File) Truncate(size int64) (err error) {
+	if f.handle == nil {
+		return ErrClosed
+	}
+
+	return Truncate(f.name, size)
+}


src/os/file_other.go
```diff
@@ -135,3 +135,8 @@ func Readlink(name string) (string, error) {
 func tempDir() string {
 	return "/tmp"
 }
+
+// Truncate is unsupported on this system.
+func (f *File) Truncate(size int64) (err error) {
+	return ErrUnsupported
+}
```
archie2x: Do you want to also add the non-receiver version?

```

// Truncate is unsupported on this system.
func Truncate(string filename, size int64) (err error) {
	return ErrUnsupported
}

func (f *File) Truncate(size int64) (err error) {
	if f.handle == nil {
		return ErrClosed
	}

	return Truncate(f.name, size)
}
```

archie2x (changes requested):
Suggest adding the non-receiver version of Truncate as well.

src/os/file_other.go
deadprogram: OK change made. Thanks for the suggestion @archie2x

archie2x (approved):

deadprogram:
Thanks for review and feedback @archie2x now merging.

PR #4435, opened 2024-08-25
Add support for LLVM 19

aykevl:
This is incomplete. I've only made this so that we have a LLVM 19 branch ready to go once LLVM 19 is released (and the Xtensa fork has been rebased).

I will make a new PR once it is ready.

Some things that need to be resolved:

  - [x] LLVM 19 needs to be released (probably early September)
  - [x] Homebrew needs to be updated (see https://github.com/tinygo-org/go-llvm/pull/65 and https://github.com/Homebrew/homebrew-core/pull/181351)
  - [x] https://github.com/llvm/llvm-project/issues/105987 needs to be fixed, or we need to use a hacky workaround.
  - [x] The Xtensa fork needs to be rebased (though we can probably have partial LLVM 19 support before then for distro maintainers).


deadprogram:
LLVM 19 is out and the default for homebrew installs: https://formulae.brew.sh/formula/llvm

aykevl:
Yes. But it should continue to work, if people install LLVM as `brew install llvm@18` (as IIRC it says in the documentation).

I can make a PR to support LLVM 19 for `go install` only if needed, but I'd prefer to do that only when needed (it is extra work to make and test). Instead, I'd prefer to wait until Espressif releases their LLVM 19 fork.

QuLogic:
I noticed with `lld` 19 that several `experimental` flags warn that they don't exist and are being ignored. As this changes those to the non-prefixed versions, doesn't that mean that `lld` 18 will now warn that _those_ don't exist and start ignoring them? In other words, don't we need some translation for `-experimental-`-prefixed features between >=19 and <19?

aykevl:
> I noticed with `lld` 19 that several `experimental` flags warn that they don't exist and are being ignored. As this changes those to the non-prefixed versions, doesn't that mean that `lld` 18 will now warn that _those_ don't exist and start ignoring them? In other words, don't we need some translation for `-experimental-`-prefixed features between >=19 and <19?

These flags often change with LLVM releases, especially for RISC-V. So far it hasn't been breaking anything to use the flags of a different LLVM release, but the warnings are kinda ugly.
One solution would be to use a different set of flags for different LLVM versions, though I don't really like to maintain those. It's a solution though if needed.

(To be clear: these warnings only happen when using a LLVM version that's not the "primary"/latest supported LLVM version. Most users shouldn't see those).

aykevl:
In other news, I've updated this PR since the Xtensa fork got updated with LLVM 19 support!

deadprogram (approved):
At last! Thanks for all the work on this @aykevl now merging.

PR #4437, opened 2024-08-26
wasip1: deduplicate target selection

aykevl:
Before this change, `GOOS=wasip1 GOARCH=wasm` and `-target=wasip1` were entirely separate configurations that just happened to be the same in practice. This change makes sure we keep only one of them to avoid inconsistencies.

This fixes the TODO here: https://github.com/tinygo-org/tinygo/pull/4175#issuecomment-1977326088

It is also intended to simplify things for https://github.com/golang/go/issues/65199 (`//go:wasmexport`).


dgryski:
If we're deleting the wasip1 target json file, what is the proper fix for a custom target json that wants to inherit from that wasip1?  For example, https://github.com/fastly/compute-sdk-go/blob/main/targets/fastly-compute-wasip1.json

ydnar:
Would inheritance work for implied or config-less targets?

aykevl:
> If we're deleting the wasip1 target json file, what is the proper fix for a custom target json that wants to inherit from that wasip1?

Oh, that's a good point. Thanks for letting me know this is being done!

I could do it the other way round, and remove the built-in configuration? In that case, `GOOS=wasip1 GOARCH=wasm` would be special-cased to load targets/wasip1.json instead.

deadprogram:
>  In that case, GOOS=wasip1 GOARCH=wasm would be special-cased to load targets/wasip1.json instead.

That sounds more flexible to me, but I will let @dgryski weigh in first before any conclusive decision.

ydnar:
The JSON config files seem like the right path. Would this also apply to other GOOS/GOARCH pairs?

(I'd considered renaming wasip1.json to wasm-waspi1.json, so the target name was the implied double)

dgryski:
I agree telling the compiler to load a different `json` file does seem like a nicer fix.

ydnar (changes requested):
My preference is to retain the `wasi.json` and `wasip1.json` configuration files, so they can be inherited from first-party and user-defined targets.

Perhaps worth having `GOOS=wasip1 GOARCH=wasm` load `wasip1.json`?

ydnar:
Do we want to close this in favor of keeping the target files as canonical?

deadprogram:
> Do we want to close this in favor of keeping the target files as canonical?

I think we should close it, yes. target files FTW! :smile:

deadprogram:
Now closing since #4746 answers this: the target file is authoritative source of build settings.

PR #4438, opened 2024-08-27
wasip2: do not export the _start function

aykevl:
It seems to have been replaced with the Component Model `run` function.

Maybe I'm misunderstanding something, but at least to me I don't see why we would need two separate ways to start a module when a Component Model `run` also works. It simplifies things.

### DIFF
--- a/src/runtime/runtime_wasm_wasip2.go
+++ b/src/runtime/runtime_wasm_wasip2.go
@@ -13,19 +13,11 @@ type timeUnit int64
 
 //export wasi:cli/run@0.2.0#run
 func __wasi_cli_run_run() uint32 {
-	_start()
-	return 0
-}
-
-//export _start
-func _start() {
 	// These need to be initialized early so that the heap can be initialized.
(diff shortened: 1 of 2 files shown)


ydnar:
Once wasmexport is implemented, this PR could also initialize the runtime in _initialize, then the wasi:cli/command#run export can run main.main.

dgryski (approved):
LGTM.

aykevl:
> Once wasmexport is implemented, this PR could also initialize the runtime in _initialize, then the wasi:cli/command#run export can run main.main.

I tried defining `_initialize`, but wasmtime (or wasm-tools?) seems to ignore it entirely. It's not being called. So no idea how the `//go:wasmexport` proposal (with `-buildmode=c-shared`) is supposed to work on wasip2.

deadprogram:
Thanks for the improvement @aykevl and to @ydnar and @dgryski for review. Now merging.

ydnar:
When we merge the wasmexport PR, we can make the #run export use //go:wasmexport

PR #4445, opened 2024-08-29
Added Script to install tiny go in Windows with one click

Thepathakarpit:
Open notepad paste script save with .bat extension. Double click on file in file manager. Done Tinygo installed in your windows system.

### DIFF
--- a/tinygo_install.bat
+++ b/tinygo_install.bat
@@ -1,0 +1,46 @@
+@echo off
+SETLOCAL ENABLEEXTENSIONS
+
+:: Install Scoop if not present
+where scoop >nul 2>&1
+if %ERRORLEVEL% neq 0 (
+    echo Scoop not found. Installing Scoop...
+    powershell -Command "iwr -useb get.scoop.sh | iex"
+)
+
+:: Install Go if not present
+where go >nul 2>&1
+if %ERRORLEVEL% neq 0 (
+    echo Go not found. Installing Go...
+    scoop install go
+)
+
+:: Install TinyGo if not present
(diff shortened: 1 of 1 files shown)


ysoldak:
Feel like this belongs to the website, can be a separate file to download or inlined in .md file.
See https://github.com/tinygo-org/tinygo-site/blob/release/content/getting-started/install/windows.md

leongross:
Hi @Thepathakarpit, as a quick heads up, I am not a Windows user and do not know a lot about the best practices.
I like the idea of simplifying Windows installs, but I am not sure if adding another script is the solution here since we already have a GNUmakefile for sth similar.
I know that one can install GNU make via [chocolatey](https://community.chocolatey.org/packages/make) and I assume it works similar to the linux version? I don't know how the integration for batch files is there but to hide the complexity and yet another script in the tree, maybe it's possible to go GNUmakefile way?

ydnar (changes requested):
The [installation instructions](https://tinygo.org/getting-started/install/windows/) for Windows are not that complicated.

I struggle to see how TinyGo providing a `bat` file that downloads and executes arbitrary code from two different sources other than TinyGo (excluding Scoop itself) on the internet to save a few keystrokes is a benefit.

I do not approve this PR.

dgryski:
I agree.  We should continue to support `scoop` as the expected way to install tinygo on Windows.  Any issues (extra dependencies, etc) can be handled by the scoop json configuration.

@Thepathakarpit Did you have any specific issues with the `scoop` installer?

Issue #4439, opened 2024-08-27
Wasi target command flag vs environment variables


mattjohnsonpint:
My understanding is that setting `GOOS=wasip1 GOARCH=wasm` should produce identical output as passing the `-target wasi` command line flag, but it doesn't.  Rather, it seem to match `-target wasm`.

Repro:

```go
package main

func main() {
	println("Hello, World!")
}
```

You'll need `wasm2wat` from [The WebAssembly Binary Toolkit](https://github.com/WebAssembly/wabt).

Then:

- Compile the different ways.  Optimizations off to see everything.
```bash
tinygo build -target wasi -opt 0 -o test1.wasm
tinygo build -target wasm -opt 0 -o test2.wasm
GOOS=wasip1 GOARCH=wasm tinygo build -opt 0 -o test3.wasm
```

- Convert output to WAT format for readability
```bash
wasm2wat test1.wasm -o test1.wat
wasm2wat test2.wasm -o test2.wat
wasm2wat test3.wasm -o test3.wat
```

Open each wat file in a text editor.  Imports are near the top, exports are near the bottom.

test1.wat

```wasm
...
(import "wasi_snapshot_preview1" "fd_write" (func $runtime.fd_write (type 4)))
(import "wasi_snapshot_preview1" "poll_oneoff" (func $runtime.poll_oneoff (type 4)))
(import "wasi_snapshot_preview1" "clock_time_get" (func $runtime.clock_time_get (type 16)))
...
(export "memory" (memory 0))
(export "malloc" (func $malloc))
(export "free" (func $free))
(export "calloc" (func $calloc))
(export "realloc" (func $realloc))
(export "_start" (func $_start))
(export "asyncify_start_unwind" (func 168))
(export "asyncify_stop_unwind" (func 169))
(export "asyncify_start_rewind" (func 170))
(export "asyncify_stop_rewind" (func 171))
(export "asyncify_get_state" (func 172))
...
```

test2.wat / test3.wat

```wasm
...
(import "gojs" "runtime.ticks" (func $runtime.ticks (type 15)))
(import "gojs" "runtime.sleepTicks" (func $runtime.sleepTicks (type 16)))
(import "wasi_snapshot_preview1" "fd_write" (func $runtime.fd_write (type 4)))
...
(export "memory" (memory 0))
(export "malloc" (func $malloc.command_export))
(export "free" (func $free.command_export))
(export "calloc" (func $calloc.command_export))
[...]

PR #4447, opened 2024-08-29
main: don't detect WebAssembly from the .wasm extension

aykevl:
Currently this overrides GOOS/GOARCH, which are also used for wasm. This will break people who rely on a command like this:

    tinygo build -o foo.wasm path/to/package

They will need to update to explicitly set the target, for example:

    tinygo build -o foo.wasm -target=wasm path/to/package

An alternative would be to only set the target when the environment variables are unset, but that gets hacky very quickly. I feel like a better solution is to not do anything and expect users to set the right target (wasm or wasip1).

Fixes: https://github.com/tinygo-org/tinygo/issues/4439

### DIFF
--- a/main.go
+++ b/main.go
@@ -1685,9 +1685,6 @@ func main() {
 			usage(command)
 			os.Exit(1)
 		}
-		if options.Target == "" && filepath.Ext(outpath) == ".wasm" {
-			options.Target = "wasm"
-		}
 
 		err := Build(pkgName, outpath, options)
 		handleCompilerError(err)


ydnar (approved):

deadprogram:
Seems like something not quite working as expected with this PR and code that targets our simulated hardware. Hence the CI failures.

deadprogram:
@aykevl please note CI failures here. :smile_cat:

dgryski:
I think this makes sense given that `wasm` now has multiple meanings: best to make *which* `wasm` explicit.

deadprogram:
@aykevl I think the modules compiled for JS but including the `machine` package are the ones failing here.

dgryski:
I still think we should merge this.  Make people get use to needing an explicit `-target=wasip1`.

ydnar:
Can we also update the documentation here? https://tinygo.org/docs/reference/usage/subcommands/

```
Compile the given program. The output binary is specified using the -o parameter. The generated file type depends on the extension:
```

deadprogram:
@aykevl what would be the correct command now for this?

```
/go/bin/tinygo build -size short -o test.wasm -tags=arduino              examples/blinky1
ld.lld-18: error: undefined symbol: __tinygo_gpio_configure
>>> referenced by machine_generic.go:31 (/root/project/src/machine/machine_generic.go:31)
>>>               /tmp/tinygo1892339853/main.lto.main.o:(main.main)
ld.lld-18: error: undefined symbol: __tinygo_gpio_set
>>> referenced by machine_generic.go:35 (/root/project/src/machine/machine_generic.go:35)
>>>               /tmp/tinygo1892339853/main.lto.main.o:(main.main)
>>> referenced by machine_generic.go:35 (/root/project/src/machine/machine_generic.go:35)
>>>               /tmp/tinygo1892339853/main.lto.main.o:(main.main)
make: *** [GNUmakefile:573: smoketest] Error 1
```

deadprogram:
Closing this since it has been superseded by #4746

PR #4449, opened 2024-08-31
compiler: fix passing weirdly-padded structs to new goroutines

aykevl:
The values were stored in the passed object as the values itself (not expanded like is common in the calling convention), and read back after assuming they were expanded. This often works for simple parameters (int, pointer, etc), but not for more complex parameters. Especially when there's padding.

Found this while working on `//go:wasmexport`.

The first commit just moves some code around, the second commit is the actual fix.


ydnar (approved):

deadprogram:
Thank you for the fix @aykevl and to @ydnar for review.

PR #4451, opened 2024-09-01
wasip1: add `//go:wasmexport` support

aykevl:
This adds support for the `//go:wasmexport` pragma as proposed here: https://github.com/golang/go/issues/65199

It is currently implemented only for wasip1, but it is certainly possible to extend it to other targets like `GOOS=js`, `-target=wasm-unknown`, or `-target=wasip2`. ~It is also currently limited to `-buildmode=c-shared`, this is a limitation that could easily be lifted in the future.~

WIP: this needs tests, but I'm not entirely sure how to test this (I've tested it by using the `--invoke` flag to `wasmtime`).

This will likely replace #4082.

If you want fast calls into the binary, I'd recommend to disable the scheduler (`-scheduler=none`). That should give a significant speedup since lots of complexity is avoided. Though there are also some possible optimizations that I haven't implemented yet (like caching stacks).

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -197,6 +197,7 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 		ABI:             config.ABI(),
 		GOOS:            config.GOOS(),
 		GOARCH:          config.GOARCH(),
+		BuildMode:       config.BuildMode(),
 		CodeModel:       config.CodeModel(),
 		RelocationModel: config.RelocationModel(),
 		SizeLevel:       sizeLevel,
@@ -649,6 +650,13 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 	result.Binary = result.Executable // final file
 	ldflags := append(config.LDFlags(), "-o", result.Executable)
 
+	if config.Options.BuildMode == "c-shared" {
+		if !strings.HasPrefix(config.Triple(), "wasm32-") {
+			return result, fmt.Errorf("buildmode c-shared is only supported on wasm at the moment")
+		}
+		ldflags = append(ldflags, "--no-entry")
+	}
+
 	// Add compiler-rt dependency if needed. Usually this is a simple load from
 	// a cache.
 	if config.Target.RTLib == "compiler-rt" {
@@ -884,7 +892,7 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 
 				err := cmd.Run()
 				if err != nil {
-					return fmt.Errorf("wasm-tools failed: %w", err)
+					return fmt.Errorf("`wasm-tools component embed` failed: %w", err)
 				}
 
 				// wasm-tools component new embedded.wasm -o component.wasm
@@ -904,7 +912,7 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 
 				err = cmd.Run()
 				if err != nil {
-					return fmt.Errorf("wasm-tools failed: %w", err)
+					return fmt.Errorf("`wasm-tools component new` failed: %w", err)
 				}
 			}
 
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -33,6 +33,17 @@ func (c *Config) CPU() string {
 	return c.Target.CPU
 }
 
+// The current build mode (like the `-buildmode` command line flag).
+func (c *Config) BuildMode() string {
+	if c.Options.BuildMode != "" {
+		return c.Options.BuildMode
+	}
+	if c.Target.BuildMode != "" {
+		return c.Target.BuildMode
+	}
+	return "default"
+}
+
 // Features returns a list of features this CPU supports. For example, for a
 // RISC-V processor, that could be "+a,+c,+m". For many targets, an empty list
 // will be returned.
--- a/compileopts/options.go
+++ b/compileopts/options.go
@@ -8,6 +8,7 @@ import (
 )
 
 var (
+	validBuildModeOptions     = []string{"default", "c-shared"}
 	validGCOptions            = []string{"none", "leaking", "conservative", "custom", "precise"}
 	validSchedulerOptions     = []string{"none", "tasks", "asyncify"}
 	validSerialOptions        = []string{"none", "uart", "usb", "rtt"}
@@ -26,6 +27,7 @@ type Options struct {
 	GOMIPS          string // environment variable (only used with GOARCH=mips and GOARCH=mipsle)
 	Directory       string // working dir, leave it unset to use the current working dir
 	Target          string
+	BuildMode       string // -buildmode flag
 	Opt             string
 	GC              string
 	PanicStrategy   string
@@ -61,6 +63,14 @@ type Options struct {
 
 // Verify performs a validation on the given options, raising an error if options are not valid.
 func (o *Options) Verify() error {
+	if o.BuildMode != "" {
+		valid := isInArray(validBuildModeOptions, o.BuildMode)
+		if !valid {
+			return fmt.Errorf(`invalid buildmode option '%s': valid values are %s`,
+				o.BuildMode,
+				strings.Join(validBuildModeOptions, ", "))
+		}
+	}
 	if o.GC != "" {
 		valid := isInArray(validGCOptions, o.GC)
 		if !valid {
--- a/compiler/goroutine.go
+++ b/compiler/goroutine.go
@@ -7,6 +7,7 @@ import (
 	"go/token"
 	"go/types"
 
+	"github.com/tinygo-org/tinygo/compiler/llvmutil"
 	"golang.org/x/tools/go/ssa"
 	"tinygo.org/x/go-llvm"
 )
@@ -101,7 +102,7 @@ func (b *builder) createGo(instr *ssa.Go) {
 
 	paramBundle := b.emitPointerPack(params)
 	var stackSize llvm.Value
-	callee := b.createGoroutineStartWrapper(funcType, funcPtr, prefix, hasContext, instr.Pos())
+	callee := b.createGoroutineStartWrapper(funcType, funcPtr, prefix, hasContext, false, instr.Pos())
 	if b.AutomaticStackSize {
 		// The stack size is not known until after linking. Call a dummy
 		// function that will be replaced with a load from a special ELF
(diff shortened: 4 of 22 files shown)


compiler/symbol.go
```diff
@@ -342,22 +373,17 @@ func (c *compilerContext) parsePragmas(info *functionInfo, f *ssa.Function) {
...
 // The list of allowed types is based on this proposal:
 // https://github.com/golang/go/issues/59149
-func (c *compilerContext) checkWasmImport(f *ssa.Function, pragma string) {
+func (c *compilerContext) checkWasmImportExport(f *ssa.Function, pragma string) {
 	if c.pkg.Path() == "runtime" || c.pkg.Path() == "syscall/js" || c.pkg.Path() == "syscall" {
 		// The runtime is a special case. Allow all kinds of parameters
 		// (importantly, including pointers).
 		return
 	}
-	if f.Blocks != nil {
```
ydnar, line 360: This should probably be retained for `go:wasmimport`

builder/build.go
```diff
@@ -649,6 +650,13 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 	result.Binary = result.Executable // final file
 	ldflags := append(config.LDFlags(), "-o", result.Executable)
 
+	if config.Options.BuildMode == "c-shared" {
+		if config.GOOS() != "wasip1" {
```
ydnar: I think we’ll quickly relax this, perhaps better to check the triple or GOARCH for a `wasm` prefix?

compileopts/config.go
```diff
@@ -91,6 +99,12 @@ func (c *Config) BuildTags() []string {
 	for i := 1; i <= c.GoMinorVersion; i++ {
 		tags = append(tags, fmt.Sprintf("go1.%d", i))
 	}
+	if c.BuildMode() == "c-shared" && strings.HasPrefix(c.Triple(), "wasm") {
+		// Use reactor mode, which means the application is initialized in
+		// _initialize and main.main is not called by default. See for example:
+		// https://github.com/WebAssembly/WASI/blob/main/legacy/application-abi.md
+		tags = append(tags, "tinygo_wasm_reactor")
```
ydnar: Proposal: rename `tinygo_wasm_reactor` to `tinygo.buildmode.c_shared`

compileopts/options.go
```diff
@@ -26,6 +27,7 @@ type Options struct {
 	GOMIPS          string // environment variable (only used with GOARCH=mips and GOARCH=mipsle)
 	Directory       string // working dir, leave it unset to use the current working dir
 	Target          string
```
ydnar, line 29: Worth adding `BuildMode` to `TargetSpec` as well, so `wasm-unknown` and `wasip2` can set build mode to `c-shared` by default?

compiler/symbol.go
```diff
@@ -292,10 +294,39 @@ func (c *compilerContext) parsePragmas(info *functionInfo, f *ssa.Function) {
...
+				if f.Blocks == nil {
+					c.addError(f.Pos(), "can only use //go:wasmexport on definitions")
+					continue
+				}
+				if len(parts) != 2 {
+					c.addError(f.Pos(), fmt.Sprintf("expected one parameter to //go:wasmimport, not %d", len(parts)-1))
+					continue
+				}
+				name := parts[1]
+				if name == "_start" || name == "_initialize" {
```
ydnar, line 334: So `//export` remains for exports like `_start` and `_initialize` that don’t get the goroutine / scheduler dance?

compiler/symbol.go
```diff
@@ -292,10 +294,39 @@ func (c *compilerContext) parsePragmas(info *functionInfo, f *ssa.Function) {
...
+					c.addError(f.Pos(), fmt.Sprintf("//go:wasmexport does not allow %#v", name))
+					continue
+				}
+				if c.BuildMode != "c-shared" && f.RelString(nil) == "main.main" {
+					c.addError(f.Pos(), fmt.Sprintf("//go:wasmexport does not allow main.main to be exported with -buildmode=%s", c.BuildMode))
```
ydnar, line 339: Does big Go blow up if `main.main` is declared with `-buildmode=c-shared`?

ydnar:
Looks like big Go supports `go:wasmexport` under `js` and with `main.main`: https://github.com/golang/go/issues/65199#issuecomment-2299416894

builder/build.go
ydnar: I’m fairly certain that @deadprogram will want to use this for `unknown`.

ydnar:
For backwards compatibility, if both `//export` and `//go:wasmexport` are found, only implement `go:wasmexport`?

ydnar:
For tests, we could implement `wasip1`, `wasip2`, and `unknown` tests.

For `wasip2`, we can implement both worlds: `wasi:cli/command` and `wasi:http/proxy`, verifying:

1. Goroutines started during `init` run during the reactor call (`wasi:cli/command#run` and [`wasi:http/incoming-handler#handle`](https://github.com/WebAssembly/wasi-http/blob/main/wit/handler.wit#L19)).
2. The expected reactor function is called, by verifying output to either stdout/stderr or the HTTP response.

For `wasip1` and `unknown`, using `wasmtime` with `--invoke` should be sufficient.

compiler/symbol.go
aykevl: Yes, but it's an implementation detail. Eventually I'd like to make `//export` work as it does on every other architecture (that is, it makes the function available to the linker but doesn't export it on a wasm level) - but that will probably require a deprecation period so people have time to switch to `//go:wasmexport`.

builder/build.go
aykevl: Yes. But to avoid too much complexity in this PR I've left it out. I tried to do it all at once but it became a bit more complex than I liked. Better to get it working for wasip1 first and then expand support to other targets (which should be relatively easy).

compileopts/config.go
aykevl: I guess we can do that, but `tinygo_wasm_reactor` seemed a bit more explicit. We can always change it later (I hope, if no external packages suddenly start depending on the build tag).

compileopts/options.go
aykevl: We can do that in the PR that adds `//go:wasmexport` to `wasm-unknown` :)
This kind of complexity is why I kept `//go:wasmexport` to wasip1 for the initial version.

compiler/symbol.go
aykevl: Yes. This is what the proposal says:

> When the `-buildmode` flag is unset, the `_start` function and any exported functions will be exported to the host. Using `//go:wasmexport` on the `main` function in this mode will result in a compilation error. In this mode, only `_start` will initialize the runtime, and so must be the first export called from the host. Any other exported functions may only be called through calling into host functions that call other exports during the execution of the `_start` function. Once the `_start` function has returned, no other exports may be called on the same instance.

aykevl (commented):
This PR still needs tests, that's what's needed to make this mergeable.

> Looks like big Go supports `go:wasmexport` under `js` and with `main.main`: [golang/go#65199 (comment)](https://github.com/golang/go/issues/65199#issuecomment-2299416894)

Yes, I've intentionally left that out of the first PR. Focusing on wasip1 first made the PR a lot simpler.
(And to be clear, the comment does not say that `main.main` can be exported, just that if you call an imported function that imported function can then call back into an exported function. Which is exactly what I've implemented for wasip1 here).

> For backwards compatibility, if both `//export` and `//go:wasmexport` are found, only implement `go:wasmexport`?

I guess we can do this. I don't see any obvious problems with that, as far as I can think the external ABI is the same (though of course `//go:wasmexport` has clearly defined semantics and `//export` does not). I would like to remove the ability for `//export` to produce wasm exported functions at some point though, because it's a non-standard TinyGo extension.

compiler/symbol.go
aykevl: It is, the check has been moved into `case "//go:wasmimport":` above.

compileopts/config.go
ydnar: Context: I think mapping the tag to the flag eases the developer experience, both for TinyGo developers and end users.

compileopts/options.go
ydnar: Ok. I'll make a follow up PR for wasip2 and unknown.

compiler/symbol.go
ydnar: I think the proposal allows main.main, and wasmexport on it, when -buildmode=c-shared:

> When using -buildmode=c-shared, the resulting Wasm binary will not contain a _start function, and will only contain the _initialize function and any exported functions. The Go main function will not be exported to the host. The user can choose to export it like any other function using the //go:wasmexport directive. The _initialize function will not automatically call main. The main function will not initialize the runtime.

ydnar: Ah, I missed the ! in the line. My mistake!

compileopts/config.go
aykevl: I have fixed this by removing the tag altogether and selecting between entry points inside the compiler. (This is similar to how big Go does it).

aykevl:
* Added some light testing
* Removed the `tinygo_wasm_reactor` build tag.
* Added runtime asserts to check that `//go:wasmexport` functions aren't being called at a time that it is not allowed.
* Added support for `//go:wasmexport` with `-buildmode=default` (though I couldn't really test this: it's impossible to test through `wasmtime --invoke`).

This is now ready for review.

main_test.go
```diff
@@ -375,10 +383,10 @@ func optionsFromOSARCH(osarch string, sema chan struct{}) compileopts.Options {
 }
 
 func runTest(name string, options compileopts.Options, t *testing.T, cmdArgs, environmentVars []string) {
-	runTestWithConfig(name, t, options, cmdArgs, environmentVars)
+	runTestWithConfig(name, t, options, cmdArgs, environmentVars, "")
 }
 
-func runTestWithConfig(name string, t *testing.T, options compileopts.Options, cmdArgs, environmentVars []string) {
+func runTestWithConfig(name string, t *testing.T, options compileopts.Options, cmdArgs, environmentVars []string, emulator string) {
```
ydnar: Rather than override `emulator`, what if struct `TargetSpec` had an `EmulatorArgs` field that is injected into the argument string for the default emulator for a target?

It could be injected in `(*Config).Emulator()`, appending `" " + c.Target.EmulatorArgs`.

src/runtime/runtime_wasip1.go
```diff
@@ -100,6 +134,51 @@ func ticks() timeUnit {
 	return timeUnit(nano)
 }
 
+// Track which state we're in: before (or during) init, running inside
```
ydnar: Can these lines be moved out to a `runtime_wasm.go` file, so they can be shared with `wasip1`, `wasip2`, and `unknown`?

main_test.go
aykevl: Fixed this by removing the new test entirely and replacing it with a much more extensive test using wazero (one that I'm actually confident in will work).

(Also found and fixed a bug where exported functions with no return values would crash the compiler).

src/runtime/runtime_wasip1.go
aykevl: Done, moved to runtime_wasmentry.go.

aykevl:
* Added `-target=wasm-unknown`. Defaulting to `-buildmode=c-shared` but `-buildmode=default` can also be used to make `_start` the entry point.
* Added some more testing for different configurations.

I looked into wasip2 support, but that's a bit harder to test because wazero doesn't support wasm components (and wasip2) yet.

ydnar (approved):

compiler/goroutine.go
```diff
@@ -121,6 +122,147 @@ func (b *builder) createGo(instr *ssa.Go) {
 	b.createCall(fnType, start, []llvm.Value{callee, paramBundle, stackSize, llvm.Undef(b.dataPtrType)}, "")
 }
 
+// Create an exported wrapper function for functions with the //go:wasmexport
+// pragma. This wrapper function is quite complex when the scheduler is enabled:
+// it needs to start a new goroutine each time the exported function is called.
+func (b *builder) createWasmExport() {
+	pos := b.info.wasmExportPos
+	if b.info.exported {
```
ydnar, line 130: Is it possible to flip this so `//export` is ignored if `//go:wasmexport` is present?

(and maybe warn?)

ydnar:
Merging per @aykevl’s message in Slack: https://gophers.slack.com/archives/CN4R2DQV7/p1728076418126569?thread_ts=1727844162.259119&cid=CN4R2DQV7

```
Yes, the wasmexport PR is ready from my POV.
```

Issue #4452, opened 2024-09-01
Add support for Raspberry Pi's RP2350 Dual Arm Cortex-M33  


soypat:
This issue would seek a replication of [this PR done for the RP2040](https://github.com/tinygo-org/tinygo/pull/1897/files) for the new RP2350 chip.

- Page: https://www.raspberrypi.com/products/rp2350/
- Datasheet: https://datasheets.raspberrypi.com/rp2350/rp2350-datasheet.pdf
- PR with FPU support for M33: https://github.com/tinygo-org/tinygo/pull/3480

Support for RISCV Hazard3 cores is out of the scope of this issue.

deadprogram:
See https://github.com/cmsis-svd/cmsis-svd-data/pull/8 for one important part of this work to be done.

maxprofit:
Does any more work have to be done for this? It appears as if there have been a couple big merges, but I don't know the expected ending state of this issue?

aykevl:
I think this is https://github.com/tinygo-org/tinygo/pull/4459?

(To be clear: I think this issue is about supporting TinyGo at all. Using both cores can be added later).

deadprogram:
Now closing since it does appear to have been completed.

PR #4459, opened 2024-09-08
Add RP2350 support

soypat:
Seeks to solve #4452. 

Notes:
- `gen-device-svd.go` Changed to support multiline Metadata descriptions such as the one found for the RP2350.
- `rp2350.ld` takes rust version as reference. Ideally I'd like a minimal version, I've left it all in so people with more experience can note of what is required and what is "extra"
- There's two packages, QFN-60 with 30 GPIOs and QFN80 with 48 GPIOs. I imagine we could expose all 48 GPIOs for the rp2350 and avoid making the distinction at the package level. Board would expose GP* pins.
- [Pico-sdk's second stage bootloader](https://github.com/raspberrypi/pico-sdk/tree/master/src/rp2350/boot_stage2) is spagghetified from what I can tell. Not sure which is the assembly we should be running... that said:
- [Likely candidate is boot2_generic_03h](https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2350/boot_stage2/boot2_generic_03h.S) which is apparently what [Rust uses](https://github.com/rp-rs/rp-hal/blob/512b102061e6397e46b941e2e627de313eb210b3/rp2040-hal-examples/src/bin/gpio_dyn_pin_array.rs#L40C45-L40C68). I've tried adding it as best as I could


deadprogram:
This is great @soypat thank you for getting going on it!

Please see https://github.com/tinygo-org/cmsis-svd-data/pull/1 for some related supporting work.

soypat:
Yes! I've looked at it and generated the device definitions. I just added a new commit with stubs for machine and runtime and am getting the following error during linking:

```
$ tinygo build -target=rp2350 -serial=none  examples/blinky1 
ld.lld: error: unable to insert .start_block after .vector_table
ld.lld: error: no memory region specified for section '.ARM.exidx'
ld.lld: error: no memory region specified for section '.text'
```

deadprogram:
I'm not too sure that the `ld` file you added will actually work with TinyGo. You probably need to have mostly the same symbols in same order as other platforms.

soypat:
OK, blinky1 now compiles. I am afraid of flashing the resulting .uf2 😅

soypat:
I've noticed the bootloader is not included in the build pipeline. I can write invalid assembly in it and everything still compiles...

src/machine/machine_rp2350.go
```diff
@@ -0,0 +1,87 @@
...
+// PinMode sets the direction and pull mode of the pin. For example, PinOutput
+// sets the pin as an output and PinInputPullup sets the pin as an input with a
+// pull-up.
+// type PinMode uint8
+
+// type PinConfig struct {
+// 	Mode PinMode
+// }
+
+func (p Pin) Set(b bool)          {}
```
rminnich: For unimplemented things I prefer a panic. The reason is that you almost never want to end up calling them, in their unimplemented form, so you don't want to accidentally call them. That would be the one thing I'd change.

src/machine/machine_rp2350.go
```diff
@@ -0,0 +1,87 @@
...
+func (p Pin) Get() bool           { return false }
+func (p Pin) Configure(PinConfig) {}
+func putchar(b byte)              {}
+
+type timeUnit int64
+
+// ticks returns the number of ticks (microseconds) elapsed since power up.
+//
+//go:linkname ticks runtime.machineTicks
+func ticks() uint64 {
```
rminnich: I'd say at the very least make this increment ... but ticks should not be hard to fill in.

src/runtime/runtime_rp2350.go
```diff
@@ -0,0 +1,83 @@
...
+	"machine"
+)
+
+// machineTicks is provided by package machine.
+func machineTicks() uint64
+
+// machineLightSleep is provided by package machine.
+func machineLightSleep(uint64)
+
+type timeUnit int64
```
rminnich, line 17: This seems a bit odd, what happens if it's accidently set to a negative value.

src/runtime/runtime_rp2350.go
```diff
@@ -0,0 +1,83 @@
...
+		// With scheduler, sleepTicks may return early if an interrupt or
+		// event fires - so scheduler can schedule any go routines now
+		// eligible to run
+		machineLightSleep(uint64(d))
+		return
+	}
+
+	// Busy loop
+	sleepUntil := ticks() + d
+	for ticks() < sleepUntil {
```
rminnich, line 48: This will be  a surprise: it will never exit. For this reason, I think ticks() should also be a panic until it's right.

targets/rp2350.ld
```diff
@@ -0,0 +1,45 @@
+/* See Rust for a more complete reference: https://github.com/rp-rs/rp-hal/blob/main/rp235x-hal-examples/memory.x */
+MEMORY
+{
+   /* 2MiB safe default.  */
```
rminnich, line 4: Are these known to match the machine hardware? This is one of the places you really want to get it right, or you'll be in Debug Hell forever.

targets/rp2350_boot2_generic03h.s
```diff
@@ -0,0 +1,392 @@
...
+#ifdef __riscv
+    mv t0, ra
+    li a3, XIP_QMI_BASE
+    li a0, INIT_M0_TIMING
+    sw a0, QMI_M0_TIMING_OFFSET(a3)
+    li a0, INIT_M0_RCMD
+    sw a0, QMI_M0_RCMD_OFFSET(a3)
+    li a0, INIT_M0_RFMT
+    sw a0, QMI_M0_RFMT_OFFSET(a3)
+#else
```
rminnich: I would make the #else have an actual  value, #elseif arm and have the fallthrough #else do a #error. Maybe that's too picky, but I've been burned by default behavior.

rminnich (commented):
my main concerns:
absolutely, make sure serial out is functional
make sure panic works
empty functions should panic
make sure all that stuff in ldscript is right
#if should have no defaults, there should always be a value, and the default should be "something was not set that is expected to be set"

soypat:
@rminnich I'll leave all functions so far in panic("unimplemented") state. We should focus on getting the bootloader working and setting the LED pin high/low.

soypat:
@deadprogram @aykevl I'm melding the RP2040 and RP2350 APIs... it's got some design decisions that some may consider questionable. I've managed to reuse a lot of logic between them at the cost of some magic constant weirdness. There are a few cases where code reuse was just not worth it though such as with the peripheral reset. Love to hear your input on how it is coming along.

soypat:
OK I feel like I've gotten the hang of writing Cortex assembler and understand what is going on with the linker scripts. I cannot however get to the point of flashing the RP2350 correctly- after I flash it the RP2350 enters bootoader mode and appears as a mass storage device ready to get flashed again.

soypat:
Aha, looks like things have changed since the RP2040. We now have to build a valid `IMAGE_DEF`, which is a binary blob with specific format.
image

soypat:
Working on a a pico binary encoder/decoder https://github.com/soypat/picobin. Raspberry Pi's picotool is missing lots of features to inspect binaries, which makes sense- it's not designed to reverse engineer their stuff. 

Also RP2350 documentation on pico binary layout is VERY lackluster compared to their hardware documentation so having code to describe the layout gives me confidence on what exactly is going on.

soypat:
OK. I've run into a limitation of the `debug/elf` package. I need to patch the bootloader by extending its length. I'm thinking of adding a elf implementation that can get the job done to https://github.com/soypat/tinyboot.

Note: @aykevl This PR would add aforementioned tinyboot as a dependency of tinygo compiler. I'm in favor of third partying build tools so that others can use the package and find bugs in parallel to the TinyGo compiler.

Edit: progress here: https://github.com/soypat/tinyboot/tree/main/build/xelf

cibomahto:
Thanks for all you great work on this! I was able to bodge together enough fixes to build a working RP2350 GPIO out example here: https://github.com/Blinkinlabs/tinygo/commit/fbb5234497a9973f5ff7b884f838db16f6f4fab1

I cheated in several ways to get it to work so it isn't pullable as-is. The main changes were:
* remove the second stage bootloader- the documentation suggests that it's no longer needed unless you have a weird spi flash chip, and elf files generated by the pico 2 sdk don't seem to include it. The functionality has been moved into the boot rom.
* add a stub embedded block right after the isr vector table, so that the boot rom can identify the image. I borrowed this from the rp2350 datasheet. A more complete system like you are describing would include a second block at the end of the image. The elf files generated by the pico 2 sdk include this second block. I think the only function it serves is to verify that the entire image was written into flash, as the default block doesn't include an image checksum.

After disabling some other initialization bits which were causing the image to hang, I succeeded in running a very short that just turns the LED on and off repeatedly.

```
package main

import (
        "machine"
)

func main() {

    led := machine.LED

    for {
        led.Low()
        led.High()
    }
}
```

cibomahto:
Here's also a very basic python tool for extracting IMAGE_BLOCKs from rp2350 ELF files: https://gist.github.com/cibomahto/790c1a34d6001a088de9bfba6e27ed23

aykevl:
> I need to patch the bootloader by extending its length.

@soypat What do you mean by this? It's generally not possible to move stuff in the firmware file after linking.

aykevl:
Would it be possible to write binaries in a similar way that ESP8266/ESP32 binaries are created? So the ELF file is not modified, instead the ELF file is converted to a binary in a custom way that adds the special header.

cibomahto:
@aykevl Looking at the [esp builder](https://github.com/tinygo-org/tinygo/blob/release/builder/esp.go):

I think not, because for the RP2350, the pico SDK generates a normal ELF image, just with some magic headers embedded in it.

For RP2350 ELF output, the pico SDK generates a normal ELF file directly, with one info block tucked between the isr table and the start of program image, and another in a second ELF section, at the end of the program image. The first one needs to include an offset to the second one (which needs to be calculated after the program image size is determined), but the second one points back to the memory address of the first one, which should be fixed. If this is done as a post-processing step, it does require rewriting the ELF file in order to make those two changes. Perhaps the builder could insert a dummy section for the second block, and then patch it with [elfpatch](https://github.com/tinygo-org/tinygo/blob/release/builder/elfpatch.go)? At a glance, this seems like it should also allow UF2 files to be generated from the patched ELF.

However, the datasheet also gives an example of a minimum implementation that doesn't need any post processing, just a [simple header](https://github.com/Blinkinlabs/tinygo/blob/fbb5234497a9973f5ff7b884f838db16f6f4fab1/targets/embedded_block.S) that needs to be placed in the first 4k of the image. I've chosen this approach to get a basic output working. The disadvantage of this is that it won't support any of the fancier validation options that the rp2350 boot rom offers, such as image signing or partition support.

aykevl:
> However, the datasheet also gives an example of a minimum implementation that doesn't need any post processing, just a [simple header](https://github.com/Blinkinlabs/tinygo/blob/fbb5234497a9973f5ff7b884f838db16f6f4fab1/targets/embedded_block.S) that needs to be placed in the first 4k of the image.

That looks like a much simpler way to do it!
This doesn't require any patching or rewriting of the ELF file, right? Just a way to tell the linker to put that magic block at the right position (presumably after the interrupt table).


> The first one needs to include an offset to the second one (which needs to be calculated after the program image size is determined),

This can probably also be done by the linker, with some special assembler magic.

cibomahto:
> > However, the datasheet also gives an example of a minimum implementation that doesn't need any post processing, just a [simple header](https://github.com/Blinkinlabs/tinygo/blob/fbb5234497a9973f5ff7b884f838db16f6f4fab1/targets/embedded_block.S) that needs to be placed in the first 4k of the image.
> 
> That looks like a much simpler way to do it! This doesn't require any patching or rewriting of the ELF file, right? Just a way to tell the linker to put that magic block at the right position (presumably after the interrupt table).

Correct. These 'blocks' each contain a pointer to the next block, in the form of an offset. The rp2350 bootrom uses the offsets to walk through the block chain, and to verify that they form a closed loop back to the first block. The trick is that if you only have one block, you can point it back to itself by specifying a zero offset, so no calculation needs to be done.

For the proof of concept, I gave it an explicit position by modifying the arm linker file: https://github.com/Blinkinlabs/tinygo/commit/fbb5234497a9973f5ff7b884f838db16f6f4fab1#diff-2048f2f7e5dcd7f7efd321f90018d841b002168ec4aa103d33363f4f4f316171R12 .

This is very straightforward, but seems a little invasive to add that for all ARM targets. Another straightforward option is to include a modified version of arm.ld just for the rp2350, but I don't see any other examples that do that and it's probably more difficult to maintain. Maybe there's an assembler directive to put something at the beginning of .text?

aykevl:
> For the proof of concept, I gave it an explicit position by modifying the arm linker file: [Blinkinlabs@fbb5234#diff-2048f2f7e5dcd7f7efd321f90018d841b002168ec4aa103d33363f4f4f316171R12](https://github.com/Blinkinlabs/tinygo/commit/fbb5234497a9973f5ff7b884f838db16f6f4fab1#diff-2048f2f7e5dcd7f7efd321f90018d841b002168ec4aa103d33363f4f4f316171R12) .

Actually in my opinion it's fine, if the name is slightly different. Something more generic/descriptive. What about `KEEP(*(.after_isr_vector))`? Maybe with a short note like `/* for the RP2350 */`.

soypat:
> What do you mean by this? It's generally not possible to move stuff in the firmware file after linking.

Not without great effort, which I was motivated to do at one point but no longer- Seeing @cibomahto PR that works looks like a good starting point. Let me know what I can do to get this merged ASAP.

cibomahto:
I sent a pull request for the changes that I knew how to add properly. I know of at least these issues that are remaining:

- GPIO: Set the new 'isolation' bit for the RP2350 only: https://github.com/Blinkinlabs/tinygo/commit/fbb5234497a9973f5ff7b884f838db16f6f4fab1#diff-92842a5fcd6e32b41bc401586ea33fd04c1cfc107babed817b12118f08372667L129
- Decide what to do for resets- the behavior of the frceoff/frceon bits has changed in rp2350. Commenting them out at least let the image boot: https://github.com/Blinkinlabs/tinygo/commit/fbb5234497a9973f5ff7b884f838db16f6f4fab1#diff-a7f56abb731d6475be3a5d0641c86fe6fd50b065ff75b2d5d5fbf21a71fcb30cL28
- Figure out why the clock init hangs. Commenting it out allowed the image to boot: https://github.com/Blinkinlabs/tinygo/commit/fbb5234497a9973f5ff7b884f838db16f6f4fab1#diff-9e41e0242756386de153301b40712f687ae1925d01da8c6ec477d770c18bc4d8L61

cibomahto:
In my 'experimental' branch, I've succeeded in getting the resets, clocks, timer, UART, and USB running, for at least enough functionality to use time.Sleep(), set a correct UART baud rate, and enable the USB CDC function for a virtual serial terminal.

For testing, I commented out the code that changed between the RP2040 and RP2350, but it needs to be integrated more cleanly. 

For instance, in machine_rp2_clocks, the init code is mostly the same, but the registers are laid out differently:
[src/machine/machine_rp2_clocks.go](https://github.com/Blinkinlabs/tinygo/commit/96df708f3bdf8d9a17fb5b349ed4dd2f067bf93b#diff-4babe1e3aead6c34db2f8de3a7d1680b3f073b142370939eebf3d9397f5666a2)

Does anyone have a suggestion or reference for how these should be set up correctly? I think these peripherals are still close enough that it makes sense to have a common '_clocks' file for them.


Reset is here:
https://github.com/Blinkinlabs/tinygo/commit/96bb22511f82a8a50667b24afeb6067e276d0429
https://github.com/Blinkinlabs/tinygo/commit/0ddbe7da80e9ab44a76b3f18bc61d963483e6ab3

System clocks (and tick source) are here:
https://github.com/Blinkinlabs/tinygo/commit/96df708f3bdf8d9a17fb5b349ed4dd2f067bf93b
https://github.com/Blinkinlabs/tinygo/commit/f2327dc7b7b5b18a3bce204d5b38751ed1817d85

and USB is here (the only difference from rp2040 are the CMSIS register names, and one pad disconnect):
https://github.com/Blinkinlabs/tinygo/commit/361bd1e3c430e8efa0b0ae1ebea8ee82123c7584

deadprogram:
The CI builds show an error `failed to calculate ELF ROM addresses: no ELF ROM sections found`...

cibomahto:
> The CI builds show an error `failed to calculate ELF ROM addresses: no ELF ROM sections found`...

It looks like this started failing at f56a0c74 . @soypat Do you know if those changes are necessary if we don't need to patch the ELF file? 

Note: I bisected the history using this pattern:

```
git checkout f56a0c74
git submodule update
go install
make gen-device
make llvm-source
tinygo build -size short -o test.hex -target=pca10040            examples/blinky1
```

soypat:
@cibomahto The use of tinyboot is no longer needed. Reverting use of tinyboot- didn't think we'd solve this one so quickly :)

deadprogram:
All unit tests now pass, including the TinyHCI hardware test on a RP2040 Pico.

Is this PR ready to be squashed and merged? @soypat @cibomahto @aykevl ?

builder/build.go
```diff
@@ -1428,6 +1434,23 @@ func printStacks(calculatedStacks []string, stackSizes map[string]functionStackS
 	}
 }
 
+func applyPatches(executable string, bootPatches []string) (err error) {
+	for _, patch := range bootPatches {
+		switch patch {
+		case "rp2040":
+			err = patchRP2040BootCRC(executable)
+		// case "rp2350":
```
deadprogram, line 1448: Is this supposed to be commented out here?

src/machine/machine_rp2.go
```diff
@@ -1,15 +1,44 @@
...
 
+const (
+	// Number of spin locks available
+	_NUMSPINLOCKS = 32
+	// Number of interrupt handlers available
+	_NUMIRQ               = 32
+	_PICO_SPINLOCK_ID_IRQ = 9
+)
+
+// UART on the RP2040
```
deadprogram, line 21: We should update comment:
```suggestion
// UART on the RP2040/RP2350
```

targets/rp2350.ld
deadprogram: Yes. 

`the RP2354 devices have internal 2MB flash memories` via https://datasheets.raspberrypi.com/rp2350/hardware-design-with-rp2350.pdf page 10

cibomahto:
@deadprogram These changes still need to be merged in a way that's compatible with rp2040: https://github.com/tinygo-org/tinygo/pull/4459#issuecomment-2513108065 . I made a first pass at it last night, will try to get it submitted tomorrow.

deadprogram:
Sounds good @cibomahto please let me know anything I can do to help out!

cibomahto:
@deadprogram Here is the pull request: #4655

I ran into an issue using a USB console and fmt that I'm not sure how to debug. Given a basic program that is compiled with ```-serial=usb```:

```
package main

import (
        "machine"
        "time"
        "fmt"
)

var (
    usbcdc = machine.USBCDC
)

func main() {
    led := machine.LED

    count := 0
    for {
        led.Low()
        fmt.Println("Hello world!", count)
//        usbcdc.WriteByte('a')

        time.Sleep(500*time.Millisecond)

        led.High()
        fmt.Println("xxx")
//        usbcdc.WriteByte('b')
        time.Sleep(500*time.Millisecond)

        count++
    }
}
```

If I use fmt(), it panics at the start of [go/src/fmt/print.go:303](https://github.com/golang/go/blob/master/src/fmt/print.go#L303) . The underlying USB seems to work because if I call usbcdc directly, then I can transfer data to and from a PC.

I think this should probably be debugged before merging.

deadprogram:
>If I use fmt(), it panics at the start of [go/src/fmt/print.go:303](https://github.com/golang/go/blob/master/src/fmt/print.go#L303) . The underlying USB seems to work because if I call usbcdc directly, then I can transfer data to and from a PC.

Seems odd, I will take a peek.

deadprogram:
> that I'm not sure how to debug

@cibomahto do we have GDB support for rp2350 yet? seems like backtrace might show something.

I can confirm the crash behavior you describe, and that `usbcdc.WriteByte()` and `usbcdc.Write()` both work.

Interestingly, when the modified version is running, I did notice something else: when trying to reflash, the pico2 seems to stop running (LED stops blinking) but never actually seems to start running the bootloader to receive the new code as no MSD device appears until a hard restart of the board.

cibomahto:
> > that I'm not sure how to debug
> 
> @cibomahto do we have GDB support for rp2350 yet? seems like backtrace might show something.

@deadprogram I'm using openocd/gdb directly (with a pico debug probe): https://github.com/Blinkinlabs/rp2350_tinygo_dev/blob/main/docker/Makefile

I wasn't able to figure out any more from the backtrace, except to confirm that it happened from the very first call to fmt(). My initial suspicion was that the buffer pool that the print library uses was not being initialized.

> I can confirm the crash behavior you describe, and that `usbcdc.WriteByte()` and `usbcdc.Write()` both work.
> 
> Interestingly, when the modified version is running, I did notice something else: when trying to reflash, the pico2 seems to stop running (LED stops blinking) but never actually seems to start running the bootloader to receive the new code as no MSD device appears until a hard restart of the board.

That's probably because of this:
https://github.com/Blinkinlabs/tinygo/blob/adecf5383066d8af59fc35d3b6ea9b53a874d122/src/machine/machine_rp2_2350.go#L208

I've been using a pico debug probe to do this development, loading the code using openocd instead of the bootloader. It shouldn't be too tough to implement this, I can take a stab at it.

deadprogram:
> That's probably because of this:
https://github.com/Blinkinlabs/tinygo/blob/adecf5383066d8af59fc35d3b6ea9b53a874d122/src/machine/machine_rp2_2350.go#L208

That would certainly explain it! :smile_cat:

deadprogram:
OK I found the USB on rp2350 problem, see https://github.com/tinygo-org/tinygo/pull/4655/files#r1884304340

deadprogram:
Rebased against latest `dev` branch.

builder/build.go
soypat: No patch required for rp2350 from my understanding

cibomahto:
@deadprogram I 'rebased' pull request #4655 against the updated rp2350-add branch, and added the usb serial fix.

Unfortunately the CI test seems to be failing in a new place now.

deadprogram:
@cibomahto I see that..
https://github.com/tinygo-org/tinygo/actions/runs/12322471622/job/34419991680?pr=4655#step:17:33

Hmm not sure why that is happening. Maybe @aykevl might know something?

deadprogram:
Tests in #4655 are now all passing!

cibomahto:
@deadprogram Great! I just pushed an implementation of enterBootloader(). I verified that it can put the device into USB bootloader mode, but haven't been able to test it with tinygo -flash because that doesn't seem compatible with my container-based build environment.

I think that covers the minimum features for the rp2350. It would still be nice to run it at 150MHz instead of 125, but that could be added later.

deadprogram:
> I just pushed an implementation of enterBootloader(). I verified that it can put the device into USB bootloader mode, 

Excellent news!

> but haven't been able to test it with tinygo -flash because that doesn't seem compatible with my container-based build environment.

Not a problem, I will test that myself shortly.

deadprogram:
@soypat I think we should merge PR #4655 into this PR now, and then mark this PR as ready for review.

deadprogram:
OK now #4655 has been rebased into this branch. GPIO and USB should be fully functional, as well as being able to reflash. That means we can *really* start development/integration :smile_cat:

builder/build.go
deadprogram: Then we should remove this, no?

src/machine/machine_rp2_pins.go
```diff
@@ -1,4 +1,4 @@
-//go:build rp2040 || ae_rp2040 || badger2040 || challenger_rp2040 || feather_rp2040 || gopher_badge || kb2040 || macropad_rp2040 || nano_rp2040 || pico || qtpy_rp2040 || thingplus_rp2040 || thumby || tufty2040 || waveshare_rp2040_zero || xiao_rp2040
+//go:build rp2040 || rp2350 || ae_rp2040 || badger2040 || challenger_rp2040 || feather_rp2040 || gopher_badge || kb2040 || macropad_rp2040 || nano_rp2040 || pico || qtpy_rp2040 || thingplus_rp2040 || thumby || tufty2040 || waveshare_rp2040_zero || xiao_rp2040
```
deadprogram, line 1: Seems to me that one of two things must be true about this line:

- either we need to add `pico2`
- OR we should replace the entire line with this because it should already be inherited via the json:
- 
```suggestion
//go:build rp2040 || rp2350
```

cibomahto: There's a wrinkle here that will have to be solved eventually- the RP2350A chip has the same 30 GPIO+ADC channels as the RP2040, however the larger RP2350B package has 48 GPIO pins, with a different ADC mapping. Since we don't have a way to specify the RP2350 variant yet, I think your suggestion is reasonable.

If a board developer wanted to provide alternate names for them, it looks like the convention is to alias the machine definitions:
https://github.com/tinygo-org/tinygo/blob/ba1919f252a2e027581d4f061ef5b7014a372071/src/machine/board_itsybitsy-m4.go#L9
https://github.com/tinygo-org/tinygo/blob/ba1919f252a2e027581d4f061ef5b7014a372071/src/machine/board_arduino.go#L7

deadprogram:
@soypat we need whatever feedback or changes very soon if we are going to be able to get this into the xmas release. :snowman_with_snow:

src/machine/machine_rp2_pins.go
deadprogram: It would be best if we address any superfluous tags in a more focused PR later.

deadprogram:
Just added Pico2 to smoketests. If everything passes, I think we should squash this PR into `dev` and then continue working on it from there.

deadprogram:
Now going to squash/merge. Thank you very much especially @soypat and @cibomahto for doing all the heavy lifting on this so far, and to everyone else who helped out!

tdunning:
Just a quick comment on this. I wound up continuing on a project using Pico2 instead of the Pico I had been working with.

I had NO idea that that you guys had *just* added this support. It is amazingly functional already. I have found a few small issues that I will file as separate items but this does not feel like software that is 3 weeks old.

Congratulations on amazing work!

PR #4461, opened 2024-09-11
wasm/js: use standard library syscall package

aykevl:
This switches `-target=wasm` (browser wasm) over from our own syscall package to the one used in the Go standard library.

While this doesn't remove any code (so we can't simplify anything), the idea is that this improves compatibility with existing code a bit more. So It's a similar reasoning as for https://github.com/tinygo-org/tinygo/pull/4417.


aykevl:
Pushed a fix, that will hopefully get CI to pass.

dgryski:
Still failing CI.  Our `os` package depends on some bits of our `syscall` package that I guess we'll need to port over:

https://github.com/tinygo-org/tinygo/actions/runs/11150962028/job/30993307747?pr=4461#step:19:582

aykevl:
I don't think this is the right way to go about it. For example: `syscall.EINVAL` in the stdlib syscall version is 22, while wasi-libc uses 28. So closing.

PR #4467, opened 2024-09-12
main: rework usage

leongross:
Remove unused switch case and improve/add command-specific usage from website.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1256,36 +1256,151 @@ func getBMPPorts() (gdbPort, uartPort string, err error) {
 	}
 }
 
-func usage(command string) {
-	switch command {
-	default:
-		fmt.Fprintln(os.Stderr, "TinyGo is a Go compiler for small places.")
-		fmt.Fprintln(os.Stderr, "version:", goenv.Version())
-		fmt.Fprintf(os.Stderr, "usage: %s <command> [arguments]\n", os.Args[0])
-		fmt.Fprintln(os.Stderr, "\ncommands:")
-		fmt.Fprintln(os.Stderr, "  build:   compile packages and dependencies")
-		fmt.Fprintln(os.Stderr, "  run:     compile and run immediately")
-		fmt.Fprintln(os.Stderr, "  test:    test packages")
-		fmt.Fprintln(os.Stderr, "  flash:   compile and flash to the device")
-		fmt.Fprintln(os.Stderr, "  gdb:     run/flash and immediately enter GDB")
-		fmt.Fprintln(os.Stderr, "  lldb:    run/flash and immediately enter LLDB")
-		fmt.Fprintln(os.Stderr, "  monitor: open communication port")
-		fmt.Fprintln(os.Stderr, "  ports:   list available serial ports")
-		fmt.Fprintln(os.Stderr, "  env:     list environment variables used during build")
-		fmt.Fprintln(os.Stderr, "  list:    run go list using the TinyGo root")
-		fmt.Fprintln(os.Stderr, "  clean:   empty cache directory ("+goenv.Get("GOCACHE")+")")
-		fmt.Fprintln(os.Stderr, "  targets: list targets")
-		fmt.Fprintln(os.Stderr, "  info:    show info for specified target")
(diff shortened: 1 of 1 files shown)


leongross:
From a cluttering perspective, I am unsure if we want to keep all the static usage messages in the middle of the code or refactor it somewhere else, something like `const.go` holding all that static data?

dkegel-fastly:
The message is significantly longer now.... I wonder if we want the old version as the default, and the new one as "more help" somehow?

leongross:
The old version is still there. I aligned the usage behavior to big go. 
`tinygo help` displays the old message.`tinygo help <subcommand>` shows the additive help if there is any.

leongross:
So with that in mind, what do you think? @dankegel @dkegel-fastly @deadprogram

leongross:
@deadprogram what do you think of this?

leongross:
If there are no change requests, can we merge this? @dankegel @archie2x @dkegel-fastly @dgryski

dgryski (approved):
LGTM

main.go
```diff
@@ -1256,36 +1256,151 @@ func getBMPPorts() (gdbPort, uartPort string, err error) {
...
+
+	.ll: 			
+					Create textual LLVM IR, after optimization. This is mainly useful for debugging.
+
+	.bc: 			
+					Create LLVM bitcode, after optimization. This may be useful for debugging or for
+					linking into other programs using LTO.
+
+	.hex 
+					Create an Intel HEX file to flash it to a microcontroller.
```
archie2x: - limit help output to 80 chars wide (even `BigGo` does this)
- inconsistent list values (.o, .ll, and .bc end in ':' while the rest do not)
- the formatting of lists is odd, the 'definition' is indented too far

I think 'Compile the given program'  is misleading since 'program' indicates something directly runnable and the list following then contradicts that idea. In any case, BigGo says: "Build compiles the packages named by the import paths, along with their dependencies, but it does not install the results." I would think if `tinygo` is largely drop-in tooling for `BigGo` it would make most sense to follow its lead and note (only?) any different behavior?

archie2x (commented):
I have have nits about the wording of things but more accessible documentation is great.

main.go
dgryski: I think the wording is mostly pulled from the website; any fixes should be made there too.

dgryski (changes requested):
Just noticed in my editor (not visible on github diff view) that there are a few lines with trailing space that should probably be cleaned up.

main.go
leongross: Indeed as @dgryski pointed out I took the existing help from the website and put it into the CLI. I agree, that the wording could use a rework.

I adjusted the tab distance of the definitions to 2 tabs like golang does.

leongross: https://github.com/tinygo-org/tinygo-site/pull/432

archie2x: I'm not any kind of authority in this project. I just have opinions like anybody else. I'm not familiar enough with this project to approve any PR especially about documentation / usage().

I was wrong:  "go help" does exceed 80 chars, uses tabs, is tied to how wide the tab is in your terminal, and: yuck.

I'll opine that the documentation closer to the code, as here, is more accessible, likely to be up to date(?), and generally good. Having it in two places is not so good.

aykevl:
I think this is an excellent idea! So if nobody objects, I'd like to see this merged :)
The only thing I'm worried about is keeping it in sync with the documentation at tinygo.org. I'm afraid we're going to forget updating the documentation in the code when we update the website.

deadprogram:
@dgryski I think your change request was already made, no?

leongross:
@aykevl keeping it in sync is also a problem I see. First, I think adding this for now is a big improvement to the current state. In the long run it mayb be a good idea to automatically pull this from the website source on a regular basis. I thought how to do this but since the webiste uses html amd css styling this is not trivial to convert.

dgryski (approved):
LGTM

deadprogram:
Thank you very much for working on this @leongross and to @dgryski @aykevl @archie2x @dkegel-fastly for reviews. Now squash/merging.

PR #4468, opened 2024-09-15
targets/wasip2: add wasmtime -S options for network access

ydnar:
This adds `wasmtime` emulator args in the `wasip2` target: `-Sinherit-network -Sallow-ip-name-lookup`

This enables prototyping an implementation of package `net` using `wasi:sockets`. It grants access to the host machine’s network, with effectively equal privileges to a non-wasm process.

### DIFF
--- a/targets/wasip2.json
+++ b/targets/wasip2.json
@@ -23,7 +23,7 @@
 	"extra-files": [
 		"src/runtime/asm_tinygowasm.S"
 	],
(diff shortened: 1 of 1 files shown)


dgryski (approved):
LGTM

deadprogram:
Now merging, thanks @ydnar and to @dgryski for review.

PR #4472, opened 2024-09-17
support -extldflags

### LINKED ISSUE #4320: Ability to pass flags to external linker (such as `wasm-ld`)
While looking into https://github.com/tinygo-org/tinygo/issues/4319, I discovered that there doesn't seem to be a way to pass flags to the `wasm-ld` external linker from the `tinygo build` command. The only way I could find is to create a custom target JSON, which is a bit heavyweight.

There is `-ldflags`, but it's documented as being for flags compatible with the Go linker.

The standard Go build system defines `-extldflags` which seems to be what I want, but `tinygo build` doesn't have that flag.

dgryski:
Fixes #4320


deadprogram:

PR #4476, opened 2024-09-18
runtime maps: random iteration and maps.clone

### LINKED ISSUE #3726: tinygo does not quite use a random offset for map iteration?
Go for some time has used a random offset for map iteration.
I think Tinygo tries to, but since fastrand() is always initialized with a seed of 1, every time you run the program, map iteration is in the same order.
This confused me quite a bit, and it might bite other programmers, too.

Presumably it would be easy to fix this by tiptoeing closer to upstream, e.g. making each runtime provide a cputicks() that can be called safely very early, and using that in runtime/algorithm.go to initialize xorshift32State and xorshift64State? 

machine.GetRNG() and getentropy() sound tempting, too, if available, available early enough, and cheap.

dgryski:
This PR fixes a few map-related issues:

* make map iteration less predictable.  It's still not a fully random key, but then again neither is Big Go's. 
* fix building the tsip runtime hash.  At some point this broken when an import cycle was introduced, so just inline the bits we need from `math/bits` and `encoding/binary`. 
* add `maps.clone` so the `maps` package can use it.
* initialize fastrand() on process startup so we actually get random numbers for map seeds
* fix build for tsip memhash

### DIFF
--- a/builder/sizes_test.go
+++ b/builder/sizes_test.go
@@ -42,8 +42,8 @@ func TestBinarySize(t *testing.T) {
 	tests := []sizeTest{
 		// microcontrollers
 		{"hifive1b", "examples/echo", 4484, 280, 0, 2252},
-		{"microbit", "examples/serial", 2732, 388, 8, 2256},
-		{"wioterminal", "examples/pininterrupt", 6016, 1484, 116, 6816},
+		{"microbit", "examples/serial", 2808, 388, 8, 2256},
+		{"wioterminal", "examples/pininterrupt", 6064, 1484, 116, 6816},
 
 		// TODO: also check wasm. Right now this is difficult, because
 		// wasm binaries are run through wasm-opt and therefore the
--- a/src/runtime/algorithm.go
+++ b/src/runtime/algorithm.go
@@ -23,7 +23,13 @@ func fastrand() uint32 {
 	return xorshift32State
 }
 
-var xorshift32State uint32 = 1
+func init() {
+	r, _ := hardwareRand()
+	xorshift64State = uint64(r | 1) // protect against 0
+	xorshift32State = uint32(xorshift64State)
+}
+
+var xorshift32State uint32
 
 func xorshift32(x uint32) uint32 {
 	// Algorithm "xor" from p. 4 of Marsaglia, "Xorshift RNGs".
@@ -43,7 +49,7 @@ func fastrand64() uint64 {
 	return xorshift64State
 }
 
-var xorshift64State uint64 = 1
+var xorshift64State uint64
 
 // 64-bit xorshift multiply rng from http://vigna.di.unimi.it/ftp/papers/xorshift.pdf
 func xorshiftMult64(x uint64) uint64 {
--- a/src/runtime/hashmap.go
+++ b/src/runtime/hashmap.go
@@ -281,13 +281,26 @@ func hashmapInsertIntoNewBucket(m *hashmap, key, value unsafe.Pointer, tophash u
 }
 
 func hashmapGrow(m *hashmap) {
+	// allocate our new buckets twice as big
+	n := hashmapCopy(m, m.bucketBits+1)
+	*m = n
+}
+
+//go:linkname hashmapClone maps.clone
+func hashmapClone(intf _interface) _interface {
+	typ, val := decomposeInterface(intf)
+	m := (*hashmap)(val)
+	n := hashmapCopy(m, m.bucketBits)
+	return composeInterface(typ, unsafe.Pointer(&n))
+}
+
+func hashmapCopy(m *hashmap, sizeBits uint8) hashmap {
 	// clone map as empty
 	n := *m
 	n.count = 0
 	n.seed = uintptr(fastrand())
 
-	// allocate our new buckets twice as big
-	n.bucketBits = m.bucketBits + 1
+	n.bucketBits = sizeBits
 	numBuckets := uintptr(1) << n.bucketBits
 	bucketBufSize := hashmapBucketSize(m)
 	n.buckets = alloc(bucketBufSize*numBuckets, nil)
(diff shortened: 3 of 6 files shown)


src/runtime/algorithm.go
```diff
@@ -23,7 +23,12 @@ func fastrand() uint32 {
 	return xorshift32State
 }
 
-var xorshift32State uint32 = 1
+func init() {
+	xorshift64State = uint64(nanotime() | 1) // protect against nanotime() returning 0
```
ydnar: So this truncates 1 bit off `nanotime` for the seed. Would it make sense to get nanotime, then if 0, set to 1 or some NOMS 32-bit number?

src/runtime/memhash_tsip.go
```diff
@@ -10,11 +10,32 @@
 package runtime
 
 import (
-	"encoding/binary"
-	"math/bits"
```
aykevl, line 14: Are you sure you can't use math/bits directly? Using it directly can give some important speed boosts because the compiler treats most of the functions in that package as runtime intrinsics and they often compile down to a single fast instruction.
I checked the package and didn't see an import of any package (except for the unsafe package which itself is a built-in package with no dependencies AFAIK).

src/runtime/hashmap.go
```diff
@@ -281,13 +281,26 @@ func hashmapInsertIntoNewBucket(m *hashmap, key, value unsafe.Pointer, tophash u
...
+	*m = n
+}
+
+//go:linkname hashmapClone maps.clone
+func hashmapClone(intf _interface) _interface {
+	typ, val := decomposeInterface(intf)
+	m := (*hashmap)(val)
+	n := hashmapCopy(m, m.bucketBits)
+	return composeInterface(typ, unsafe.Pointer(&n))
+}
```
aykevl, line 295: Good idea, I didn't think of reusing the growing code for map cloning.

src/runtime/algorithm.go
aykevl: I guess an alternative would be to use `nanotime() | 1<<63`, or `nanotime() + 1` (I don't think `nanotime` goes negative).

builder/sizes_test.go
```diff
@@ -41,9 +41,9 @@ func TestBinarySize(t *testing.T) {
 	// This is a small number of very diverse targets that we want to test.
 	tests := []sizeTest{
 		// microcontrollers
-		{"hifive1b", "examples/echo", 4484, 280, 0, 2252},
-		{"microbit", "examples/serial", 2732, 388, 8, 2256},
-		{"wioterminal", "examples/pininterrupt", 6016, 1484, 116, 6816},
+		{"hifive1b", "examples/echo", 4504, 280, 0, 2252},
+		{"microbit", "examples/serial", 2736, 388, 8, 2256},
+		{"wioterminal", "examples/pininterrupt", 6020, 1484, 116, 6816},
```
aykevl: That's unfortunate :(
Not sure why this is happening, I'm guessing something with reflect and maybe interface comparisons?

Did you check which commit caused these changes? It would be interesting to know (and perhaps should be squashed together with that commit).

aykevl (commented):
I find it ironic that big Go said map iteration order is undefined, so made map iteration random, and now there's a lot of code depending on the random iteration order.
(There's a difference between random and undefined! Always producing the map keys in sorted order would be entirely valid according to the Go spec).

Anyway, see my comment below. Didn't investigate the code that changes the iteration order in detail, but it seems reasonable on the surface.

src/runtime/memhash_tsip.go
dgryski: Looks like the circular import was actually from `encoding/binary` so I've put back the use of `math/bits`.

Edit: `encoding/binary` -> `slices` -> `iter` -> `runtime`.

builder/sizes_test.go
dgryski: This is almost certainly seeding-runtime-fastrand change, but I'll double check.

Edit: Yup, it was.

src/runtime/memhash_tsip.go
aykevl: Thanks for the update!
I only now realize I've had this problem before, and fixed it with https://github.com/tinygo-org/tinygo/pull/4306. Can you check whether you can use that internal package instead? (You'll probably need to add `LittleEndian.Uint64`). That would avoid some duplication.

builder/sizes_test.go
aykevl: Right, I see. There's not much to do about that apart from making map iteration order non-random again.

src/runtime/algorithm.go
aykevl: Unfortunately, `nanotime()` doesn't really add any randomness on baremetal because it starts at 0 on reset (so will still be 0 or very close to it in `init()`, in a very predictable way).

There are two ways this can be handled:

  * It could be left at a default value like 1 on baremetal (keeping the non-random behavior).
  * It could also use the `hardwareRand` function to obtain real random numbers from the hardware, if that's supported (it is on most 32-bit baremetal chips and on all common OSes). That would in fact be more random than `nanotime()` even on Linux/MacOS/Windows/WASI.

src/runtime/memhash_tsip.go
dgryski: `Uint64` added to `internal/binary`, and I switched the `tsip` code over to that: https://github.com/tinygo-org/tinygo/pull/4476/commits/a74be5d833f7af56c33ef0b7dc55545d6518a42f

src/runtime/algorithm.go
dgryski: Switched to hardware randomess in https://github.com/tinygo-org/tinygo/pull/4476/commits/a47c0c1386eefbf642c8c7c27e79d9e29ef03a71

ydnar (approved):

dgryski:
Need to fix the binary sizes again ..

PR #4481, opened 2024-09-24
runtime: add runtime.fcntl link directives

leongross:
Building u-root commands with tinygo (see https://github.com/u-root/u-root/issues/2979) returns linking errors for the system runtime function `runtime.fcntl`. This PR intends to fix this issue. Affected u-root cmdlets:

* localboot
* netbootxyz
* newsshd
* systemboot

### DIFF
--- a/src/runtime/os_linux.go
+++ b/src/runtime/os_linux.go
@@ -142,6 +142,6 @@ func hardwareRand() (n uint64, ok bool) {
 //export getrandom
 func libc_getrandom(buf unsafe.Pointer, buflen uintptr, flags uint32) uint32
 
-//go:linknam runtime_fcntl syscall/syscall_fcntl
-//go:export
-func runtime_fcntl(fd, cmd, arg int32) (ret int32, errno int32)
+func fcntl(fd, cmd, arg int32) (ret int32, errno int32){
+    return 0, 0
+}
--- a/src/runtime/os_linux.go
+++ b/src/runtime/os_linux.go
@@ -5,7 +5,9 @@ package runtime
 // This file is for systems that are _actually_ Linux (not systems that pretend
 // to be Linux, like baremetal systems).
 
-import "unsafe"
+import (
+	"unsafe"
+)
 
 const GOOS = "linux"
 
@@ -139,3 +141,7 @@ func hardwareRand() (n uint64, ok bool) {
 //
 //export getrandom
 func libc_getrandom(buf unsafe.Pointer, buflen uintptr, flags uint32) uint32
+
+//go:linknam runtime_fcntl syscall/syscall_fcntl
+//go:export
+func runtime_fcntl(fd, cmd, arg int32) (ret int32, errno int32)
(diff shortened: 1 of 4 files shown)


src/syscall/syscall_libc.go
```diff
@@ -212,6 +212,13 @@ func Truncate(path string, length int64) (err error) {
 	return
 }
 
+//go:linkname syscall_fcntl runtime/runtime.fcntl
+func syscall_fcntl(fd, cmd, arg int32) (ret int32, errno int32) {
+	// https://cs.opensource.google/go/go/+/master:src/runtime/os_linux.go;l=452?q=runtime.fcntl&ss=go%2Fgo
+	r, _, err := Syscall6(SYS_FCNTL, uintptr(fd), uintptr(cmd), uintptr(arg), 0, 0, 0)
```
aykevl, line 219: Why are you trying to do a direct system call in syscall_libc.go? `Syscall6` and the like are only supported on Windows and Linux, not on many other systems.

What about calling the libc function `fcntl` instead?

src/runtime/os_linux.go
```diff
@@ -139,3 +141,7 @@ func hardwareRand() (n uint64, ok bool) {
 //
 //export getrandom
 func libc_getrandom(buf unsafe.Pointer, buflen uintptr, flags uint32) uint32
+
+func fcntl(fd, cmd, arg int32) (ret int32, errno int32){
+    return 0, 0
+}
```
aykevl: This is a bad idea. It should not just ignore the call and return without an error.
Is this intended to call `syscall_fcntl` perhaps?

And then, any reason not to call the libc function instead (like with `syscall_Getpagesize` for example).

src/syscall/syscall_libc.go
```diff
@@ -212,6 +212,13 @@ func Truncate(path string, length int64) (err error) {
 	return
 }
 
+//go:linkname syscall_fcntl runtime/runtime.fcntl
```
aykevl, line 216: I think you meant `runtime.fcntl`?

```suggestion
//go:linkname syscall_fcntl runtime.fcntl
```

leongross:
Sorry for wasting your time, this PR was kind of outdated and does not represent the current state. I will push the latest changes by tomorrow.

src/syscall/syscall_libc.go
leongross: Creating a wrapper to the libc sounds like a good idea.

leongross:
I am perplexed by the linker's behavior. After reconsidering your design choices, what I want to do is the following:

* create function `Fcntl` in `syscall_unix` and `syscall_libc`
* in `os_linux` add the function declaration for `fcntl`
* link the syscall function over the `runtime.fcntl` function (since we cannot import the syscall package due to circular dependencies)

I tried:
* link the `runtime/fcnlt` function to `syscall.Fcntl` -> did not work
* link the `syscall.Fcntl` function(s) to `runtime.fcntl` -> did not work
* declare `runtime.fcntl` as linkable (`//go:linkname fcntl`) and link from the syscall package up -> did not work

I realized it only worked when I defined `runtime.fcntl` to return nothing. I would really appreciate your input on this.

Issue #4488, opened 2024-10-02
`init()` for `wasip2` reactor components


rvolosatovs:
Currently, `init()` functions do not run in `wasip2` reactor components run in Wasmtime 25.

A recent change to `wasm-tools` introduced support for `_initialize` in `component new` step https://github.com/bytecodealliance/wasm-tools/pull/1747, however even with:
- `wasm-tools 0.218`
- `tinygo 0.33`

`init()` functions do not run in reactors, which manifests in e.g. stdio not being available and runtime panics e.g. printing to `os.Stdout`

It's possible that the outdated WASI-libc version is the cause here, I'll try updating it and report

Refs #4482 

Update 1:

neither updating vendored `wasi-libc` to https://github.com/WebAssembly/wasi-libc/commit/5ed3ec5701b58347d28a87c84ccc95f9c54b0148, nor https://github.com/WebAssembly/wasi-libc/commits/wasi-sdk-22 seemed to have changed anything, but I'm also not familiar with the build process enough to know if I had to do something more that just `go install` to make sure TinyGo uses the latest `wasi-libc`

PR #4490, opened 2024-10-02
fix(wasip2): export `_initialize`

rvolosatovs:
Closes https://github.com/tinygo-org/tinygo/issues/4488
Refs https://github.com/tinygo-org/tinygo/pull/4482

Turns out `_initialize` was not exported by `wasip2` components and so `init()` was never called in reactors. This PR fixes that, by initializing the heap and calling all initializers, just like `run()` would normally do

Note: `wasm-tools 1.217` or later are *required* for `init()` to work in reactors (since `_initialize` is ignored in earlier versions, see https://github.com/bytecodealliance/wasm-tools/pull/1747)

Thanks @alexcrichton for assisting in debugging this in https://bytecodealliance.zulipchat.com/#narrow/stream/217126-wasmtime/topic/calling.20.60_initialize.60

### DIFF
--- a/src/runtime/runtime_wasm_wasip2.go
+++ b/src/runtime/runtime_wasm_wasip2.go
@@ -3,6 +3,7 @@
 package runtime
 
 import (
+	"sync"
 	"unsafe"
 
 	"internal/wasi/cli/v0.2.0/environment"
@@ -11,12 +12,34 @@ import (
 
 type timeUnit int64
 
-//export wasi:cli/run@0.2.0#run
-func __wasi_cli_run_run() uint32 {
+var callInitAll = sync.OnceFunc(initAll)
+
+var initialize = sync.OnceFunc(func() {
 	// These need to be initialized early so that the heap can be initialized.
 	heapStart = uintptr(unsafe.Pointer(&heapStartSymbol))
 	heapEnd = uintptr(wasm_memory_size(0) * wasmPageSize)
-	run()
+	initHeap()
+})
+
+//export _initialize
+func _initialize() {
+	initialize()
+	callInitAll()
+}
+
+//export wasi:cli/run@0.2.0#run
+func __wasi_cli_run_run() uint32 {
+	initialize()
+	if hasScheduler {
+		go func() {
+			callInitAll()
+			callMain()
+			schedulerDone = true
+		}()
+		scheduler()
+	} else {
+		callMain()
+	}
 	return 0
 }
 


lxfontes:
ran same scenario as in #4482

Before
```
varHolder: {str:top ptr:0x15b30 mapStr:map[] mapPtr:map[]}
ptrHolder: &{str:top ptr:0x15b40 mapStr:map[] mapPtr:map[]}
callHolder: {str:top ptr:0x15df0 mapStr:map[a:b] mapPtr:map[c:0x16140]}
```

With this PR
```
varHolder: {str:top ptr:0x15b38 mapStr:map[a:b] mapPtr:map[c:0x15b40]}
ptrHolder: &{str:top ptr:0x15b48 mapStr:map[a:b] mapPtr:map[c:0x15b50]}
callHolder: {str:top ptr:0x164c0 mapStr:map[a:b] mapPtr:map[c:0x16720]}
```

🎉

aykevl (changes requested):
With this change, if `wasi:cli/run@0.2.0#run` is called, the package initializers will be called twice. I don't think that's the intention?

rvolosatovs:
> With this change, if `wasi:cli/run@0.2.0#run` is called, the package initializers will be called twice. I don't think that's the intention?

how about https://github.com/tinygo-org/tinygo/pull/4490/commits/03d535de3b1a726d8f532c536187a0574f5b71bf ?

rvolosatovs:
Looking at tests, it does not look like `_initialize` is actually called for `run` implementations?
Perhaps we need to call `_initialize` from within `run` and guard by `sync.Once` or such?

rvolosatovs:
Reworked the PR a bit - initialization is now done always, but guarded by `sync.OnceValue` - since the spec is not completely clear on the semantics here, this ensures that initialization is done for both reactors and command components.

For reactors, runtime must always call `_initialize` and for commands it may not

aykevl:
> Looking at tests, it does not look like `_initialize` is actually called for `run` implementations?

That's not how I understand the wasip2 spec, my understanding is that `_initialize` needs to be called first in all cases but that it wasn't always done previously.

@ydnar?

ydnar:
What version of `wasm-tools` is used to create the component? It needs at least v1.217.0: https://github.com/bytecodealliance/wasm-tools/compare/v1.216.0...v1.217.0

The relevant commit is: https://github.com/bytecodealliance/wasm-tools/commit/108c37d94fea2e1b764ecf329edbbeea1e9326b0

ydnar:
Once #3165 lands, the `wasi:cli/command#run` export should be wired up to call `main.main` and do nothing else.

ydnar:
Fixed in #4496.

ydnar:
#4496 merged. Want to try `dev` branch?

ydnar:
@rvolosatovs is still necessary? Can you verify against dev branch?

rvolosatovs:
> @rvolosatovs is still necessary? Can you verify against dev branch?

latest `dev` branch seems to work

PR #4433, opened 2024-08-23
ci: test newly supported packages

aykevl:
In https://github.com/tinygo-org/tinygo-site/pull/423 I see that there are a lot of newly supported packages. This PR adds testing for them, so that we won't regress.

PR #4492, opened 2024-10-02
GNUmakefile: add some more passing stdlib tests

dgryski:


dgryski:
Just realized this is mostly duplicating https://github.com/tinygo-org/tinygo/pull/4433 , which for some reason I thought was only adding a much smaller number of packages..

aykevl:
This fails on MacOS:

```
panic: runtime error at 0x0000000100f36ba0: goroutine stack overflow
FAIL	go/printer	0.199s
```

Kinda surprised by that one, perhaps it has some test for the maximum stack depth or something?

These fail on Windows:

```
FAIL	debug/dwarf	0.626s
```

```
--- FAIL: Test (0.00s)
    no testdata
    FailNow is incomplete, requires runtime.Goexit()
FAIL
FAIL	internal/diff	0.058s
```

dgryski:
Removed some packages and now everything is green.  If we merge this we can update `dev` with whatever is left-over from your PR.

PR #4493, opened 2024-10-03
builder: keep wasm temporary files

aykevl:
Don't rename them when `-work` is set, instead update `result.Binary` each time and leave `result.Executable` be the linker output (as intended: `result.Executable` should be the unmodified linker output).


dgryski (approved):
LGTM.  My initial patch was hacky anyway because I needed the intermediate files to debug an wasm-opt issue.  Thanks for cleaning this up.

deadprogram:
Thanks for the improvements @aykevl and to @dgryski for review. Now merging.

PR #4494, opened 2024-10-03
wasm: add `//go:wasmexport` support to js/wasm 

aykevl:
This adds support for //go:wasmexport with `-target=wasm` (in the browser). This follows the //go:wasmexport proposal, meaning that blocking functions are not allowed.

Both `-buildmode=default` and `-buildmode=c-shared` are supported. The latter allows calling exported functions after `go.run()` has returned.

This PR depends on #4451 so will remain a draft until that one is merged.

### DIFF
--- a/src/runtime/runtime_wasm_js.go
+++ b/src/runtime/runtime_wasm_js.go
@@ -2,25 +2,14 @@
 
 package runtime
 
-import "unsafe"
-
 type timeUnit float64 // time in milliseconds, just like Date.now() in JavaScript
 
 // wasmNested is used to detect scheduler nesting (WASM calls into JS calls back into WASM).
 // When this happens, we need to use a reduced version of the scheduler.
+//
+// TODO: this variable can probably be removed once //go:wasmexport is the only
+// allowed way to export a wasm function (currently, //export also works).
 var wasmNested bool
-
-//export _start
-func _start() {
-	// These need to be initialized early so that the heap can be initialized.
-	heapStart = uintptr(unsafe.Pointer(&heapStartSymbol))
-	heapEnd = uintptr(wasm_memory_size(0) * wasmPageSize)
-
-	wasmNested = true
-	run()
-	__stdio_exit()
-	wasmNested = false
-}
 
 var handleEvent func()
 
@@ -50,3 +39,7 @@ func sleepTicks(d timeUnit)
 
 //go:wasmimport gojs runtime.ticks
 func ticks() timeUnit
+
+func beforeExit() {
+	__stdio_exit()
+}
--- a/targets/wasm_exec.js
+++ b/targets/wasm_exec.js
@@ -466,20 +466,13 @@
 			this._idPool = [];      // unused ids that have been garbage collected
 			this.exited = false;    // whether the Go program has exited
 
-			while (true) {
-				const callbackPromise = new Promise((resolve) => {
-					this._resolveCallbackPromise = () => {
-						if (this.exited) {
-							throw new Error("bad callback: Go program has already exited");
-						}
-						setTimeout(resolve, 0); // make sure it is asynchronous
-					};
-				});
+			if (this._inst.exports._start) {
 				this._inst.exports._start();
-				if (this.exited) {
-					break;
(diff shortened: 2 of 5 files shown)


ydnar:
@aykevl want to rebase?

aykevl:
Rebased, ready for review!

src/runtime/runtime_wasm_js.go
```diff
@@ -2,26 +2,15 @@
...
-//export _start
-func _start() {
-	// These need to be initialized early so that the heap can be initialized.
-	heapStart = uintptr(unsafe.Pointer(&heapStartSymbol))
-	heapEnd = uintptr(wasm_memory_size(0) * wasmPageSize)
-
-	wasmNested = true
-	run()
-	__stdio_exit()
-	wasmNested = false
```
aykevl, line 22: I'm not entirely sure about removing `wasmNested` here. I'm not entirely sure how this is supposed to work, actually. I believe it is for reentrant calls exposed by `//export` - which is better supported anyway with `//go:wasmexport` now.

src/runtime/runtime_wasm_js.go
```diff
@@ -2,26 +2,15 @@
...
 package runtime
 
-import "unsafe"
-
 type timeUnit float64 // time in milliseconds, just like Date.now() in JavaScript
 
 // wasmNested is used to detect scheduler nesting (WASM calls into JS calls back into WASM).
 // When this happens, we need to use a reduced version of the scheduler.
+//
+// TODO: this variable can probably be removed once //go:wasmexport is the only
```
ydnar, line 10: Should the other mechanism for calling an export in js use the same semantics?

targets/wasm_exec.js
```diff
@@ -466,20 +466,13 @@
...
-					};
-				});
+			if (this._inst.exports._start) {
 				this._inst.exports._start();
-				if (this.exited) {
-					break;
-				}
-				await callbackPromise;
+
+				// TODO: wait until the program exists.
```
ydnar, line 472: Seems important?

deadprogram:
@aykevl and @ydnar what is the status of this PR?

@aykevl can you please rebase against latest `dev`?

deadprogram:
All tests are passing. @ydnar any further feedback? 
@aykevl any other changes you wish to make?

ydnar:
I don't have much experience with wasm/js. It seems fine?

targets/wasm_exec.js
aykevl: The behavior should be the same as it was before. If you read the code carefully, you see that it wait for `callbackPromise`, which is resolved when `this._resolveCallbackPromise` is called, but `this._resolveCallbackPromise` is never called.
So `await new Promise(() => {});` should be exactly the same behavior: never resolving. It basically just simplifies the code.

src/runtime/runtime_wasm_js.go
aykevl: I find it difficult to say. `//export` currently doesn't have clearly documented semantics around the scheduler.
The only thing I was able to find was this comment in https://github.com/tinygo-org/tinygo/pull/2011:

> If WASM calls into JS which calls back into WASM, it is possible for the scheduler to nest.
> The event from the callback must be handled immediately, so the task cannot simply be deferred to the outer scheduler.
> This creates a minimal scheduler loop which is used to handle such nesting.

However, all tests still pass with this change. So it's not documented, and there are no tests. I don't even know what the correct behavior is supposed to be exactly. So, uh, :woman_shrugging:
My suggestion would be to leave this alone, and if it causes any issues we can fix them when they happen.

I think I would even like to remove the `minSched`/`wasmNested` stuff in the future and hope things don't break. Because it's not clearly documented what they do exactly and I don't like code with no clear purpose.

aykevl:
No further changes I want to make, though there's some weird stuff with some older code that I don't fully understand and for which there isn't any documentation or tests. I propose we ignore those.

deadprogram:
Now merging, thanks @aykevl for working on this and to @ydnar for review.

Issue #4479, opened 2024-09-23
runtime should pass layout info to alloc() when possible


dgryski:
Right now we only have layout info for allocations created via the compiler.  We should make it possible for the runtime to pass layout info for common types at least ([]byte, []string, etc).

dgryski:
Tried a quick POC below
```
diff --git a/src/runtime/string.go b/src/runtime/string.go
index aeefe1d4..ceb5ff13 100644
--- a/src/runtime/string.go
+++ b/src/runtime/string.go
@@ -32,6 +32,12 @@ func stringEqual(x, y string) bool {
        return true
 }

+var (
+       // Constants for use with alloc()
+       gcLayoutZero   uintptr
+       gcLayoutNoPtrs = unsafe.Pointer(&gcLayoutZero)
+)
+
 // Return true iff x < y.
 //
 //go:nobounds
@@ -59,7 +65,7 @@ func stringConcat(x, y _string) _string {
                return x
        } else {
                length := x.length + y.length
-               buf := alloc(length, nil)
+               buf := alloc(length, gcLayoutNoPtrs)
                memcpy(buf, unsafe.Pointer(x.ptr), x.length)
                memcpy(unsafe.Add(buf, x.length), unsafe.Pointer(y.ptr), y.length)
                return _string{ptr: (*byte)(buf), length: length}
@@ -72,7 +78,7 @@ func stringFromBytes(x struct {
        len uintptr
        cap uintptr
 }) _string {
-       buf := alloc(x.len, nil)
+       buf := alloc(x.len, gcLayoutNoPtrs)
        memcpy(buf, unsafe.Pointer(x.ptr), x.len)
        return _string{ptr: (*byte)(buf), length: x.len}
 }
@@ -83,7 +89,7 @@ func stringToBytes(x _string) (slice struct {
        len uintptr
        cap uintptr
 }) {
-       buf := alloc(x.length, nil)
+       buf := alloc(x.length, gcLayoutNoPtrs)
        memcpy(buf, unsafe.Pointer(x.ptr), x.length)
        slice.ptr = (*byte)(buf)
        slice.len = x.length
@@ -100,7 +106,7 @@ func stringFromRunes(runeSlice []rune) (s _string) {
        }

        // Allocate memory for the string.
-       s.ptr = (*byte)(alloc(s.length, nil))
+       s.ptr = (*byte)(alloc(s.length, gcLayoutNoPtrs))

[...]

PR #4495, opened 2024-10-03
add gc layout for some basic types

dgryski:


dgryski:
For https://github.com/tinygo-org/tinygo/issues/4479

src/runtime/slice.go
```diff
@@ -47,7 +47,13 @@ func sliceGrow(oldBuf unsafe.Pointer, oldLen, oldCap, newCap, elemSize uintptr)
 	// memory allocators, this causes some difficult to debug issues.
 	newCap = 1 << bits.Len(uint(newCap))
 
-	buf := alloc(newCap*elemSize, nil)
+	var layout unsafe.Pointer
+	if elemSize == 1 {
```
ydnar: Could this check < size of pointer?

dgryski: Yeah, I was just being ultra-conservative for this initial POC, and also my test case was focussed on strings.  This could absolutely be smarter.

dgryski:
Here's the result of https://github.com/dgryski/trifles/tree/master/tinyjsonbench with the layout patches:
```
   3.1  +
        |                   |
        |         +-------------------+
        |                   |
   3.0  +
        |
        |
        |
   2.9  +                                        *
        |
        |
   2.8  +                              +-------------------+
        |                              +---------*---------+
   2.7  +--------------------------------------------------------------

File       N   Mean  Stddev
old.times  20  3.06  0.02    (control)
new.times  20  2.79  0.04    (2.79 < 3.06 ± 0.02, p = .000)
```
and on wasi
```
        |
  10.0  +
        |
        |
   9.8  +                   |
        |                   |
        |                   |
   9.6  +                   |
        |         +-------------------+
        |         |         *         |
        |         +-------------------+          |
   9.4  +                   |                    |
        |                   |          +---------*---------+
   9.2  +                                        |
        |
        |
   9.0  +--------------------------------------------------------------

File       N   Mean  Stddev
old.times  20  9.49  0.11    (control)
new.times  20  9.31  0.07    (9.31 < 9.49 ± 0.06, p = .000)
```

dgryski:
Ideally I think the layout should be computed by the compiler and stored with the rest of the type info data, but I'm leaving that for a later PR.  I just did the easy ones for now.

src/reflect/gclayout_precise.go
```diff
@@ -0,0 +1,37 @@
...
+
+	if r.Kind() == String {
+		return gcLayoutString
+	}
+
+	if r.Kind() == Slice {
+		return gcLayoutSlice
+	}
+
+	// Unknown (for now); less the conservative pointer scanning handle it
```
aykevl: Typo?

src/reflect/gclayout_precise.go
```diff
@@ -0,0 +1,37 @@
...
+
+func (r *rawType) gcLayout() unsafe.Pointer {
+
+	if r.Kind() < String {
+		return gcLayoutNoPtrs
+	}
+
+	if r.Kind() == Pointer || r.Kind() == UnsafePointer {
+		return gcLayoutPointer
+	}
```
aykevl: Channels and maps are also pointers, and can be added here.
Also, you could consider doing something like this:

```go
kind := r.Kind()
if kind < String {
    return gcLayoutNoPtrs
}
switch kind {
case Pointer, UnsafePointer, Map, Chan:
    return gcLayoutPointer
case String:
    // etc
}
```

src/reflect/gclayout_precise.go
```diff
@@ -0,0 +1,37 @@
...
+
+var (
+	// Constants for use with alloc()
+	// See runtime/gc_precise.go
+	// pppppppp_pppppppp_pppppppp_ppsssss1
+	gcLayoutNoPtrs  = unsafe.Pointer(uintptr(0b0000_00001_1))
+	gcLayoutPointer = unsafe.Pointer(uintptr(0b0001_00001_1))
+	gcLayoutString  = unsafe.Pointer(uintptr(0b0001_00010_1))
+	gcLayoutSlice   = unsafe.Pointer(uintptr(0b0001_00011_1))
+)
```
aykevl: Unfortunately all these constants (specifically the amount of bits used for the size of the object) vary by pointer size (16, 32, or 64 bits).
I think it may be possible to get the right constants with some const bit twiddling?
Much easier would probably be to write a function that calculates these at runtime (switching over `unsafe.Sizeof(uintptr(0))`) and hope the function gets inlined and optimized to a single constant.

src/reflect/gclayout_none.go
```diff
@@ -0,0 +1,9 @@
+//go:build !gc.precise
+
+package reflect
+
+import "unsafe"
+
+func (r *rawType) gcLayout() unsafe.Pointer {
+    return nil
+}
```
aykevl: I think it's fine to pass the real `gcLayout` in all cases. When not using the precise GC, the parameter is ignored. I think the optimizer will be smart enough to optimize it all away. (And honestly, I think the precise GC is the way forward anyway on most platforms).

src/runtime/slice.go
```diff
@@ -47,7 +47,13 @@ func sliceGrow(oldBuf unsafe.Pointer, oldLen, oldCap, newCap, elemSize uintptr)
 	// memory allocators, this causes some difficult to debug issues.
 	newCap = 1 << bits.Len(uint(newCap))
 
-	buf := alloc(newCap*elemSize, nil)
+	var layout unsafe.Pointer
+	// less type info here; can only go off element size
+	if elemSize < unsafe.Sizeof(uintptr(0)) {
+		layout = gcLayoutNoPtrs
+	}
```
aykevl: This is probably best solved by a small compiler change that will pass the layout to `runtime.sliceAppend` (like it does today with `runtime.alloc`).

src/runtime/string.go
```diff
@@ -32,6 +32,11 @@ func stringEqual(x, y string) bool {
 	return true
 }
 
+var (
+	// Constants for use with alloc()
+	gcLayoutNoPtrs = unsafe.Pointer(uintptr(0x3))
+)
```
aykevl: Perhaps we should move constants like these to a separate internal package? This is quite finicky stuff and easy to get wrong.

aykevl (commented):
Good idea! It's one of those "I should do that some day" but never actually did.

dgryski:
PTAL.  I think I've addressed everything except the compiler change to pass more type info to the runtime slice functions.

ydnar:
Is a GC layout in this PR the same thing that big Go refers to as a GC shape?

dgryski:
@ydnar Yes, basically.  (For those playing along at home: https://github.com/golang/proposal/blob/master/design/generics-implementation-dictionaries-go1.18.md )

For our implementation, see https://github.com/tinygo-org/tinygo/blob/dev/src/runtime/gc_precise.go .

aykevl:
> Is a GC layout in this PR the same thing that big Go refers to as a GC shape?

More or less ([source](https://go.googlesource.com/proposal/+/refs/heads/master/design/generics-implementation-gcshape.md)):

> The _GC shape_ of a type means how that type appears to the allocator / garbage collector. It is determined by its size, its required alignment, and which parts of the type contain a pointer.

The TinyGo GC layout value doesn't contain alignment and it barely contains size (it doesn't repeat pointer locations, so for a slice or array it would only use the element layout and expect the GC to repeat that). But most importantly, it contains information on which parts (may) contain a pointer - or rather, which parts most definitely don't contain a pointer.

Also, am I glad now that I wrote some rather extensive documentation for this feature!

aykevl (approved):
LGTM. Assuming it passes the test corpus, I think this is fine to merge?

dgryski:
Yes, it passes the test corpus.   If you're happy with the code in `gclayout.go` that constructs the layout bits, then yes we can merge this.

aykevl:
Interestingly there are some binary size and RAM usage changes:

```
flash                          ram
before   after   diff          before   after   diff
[...]
244264  245468   1204   0.49%    9516   10488    972  10.21% tinygo build -size short -o ./build/test.hex -target=nano-rp2040 -stack-size 8kb ./examples/net/websocket/dial/
285816  287104   1288   0.45%   17240   18212    972   5.64% tinygo build -size short -o ./build/test.hex -target=wioterminal -stack-size 8kb ./examples/net/mqttclient/paho/
337000  338372   1372   0.41%   15748   15396   -352  -2.24% tinygo build -size short -o ./build/test.hex -target=matrixportal-m4 -stack-size 8kb ./examples/net/webstatic/
284680  286076   1396   0.49%   18172   17812   -360  -1.98% tinygo build -size short -o ./build/test.hex -target=wioterminal -stack-size 8kb ./examples/net/webserver/
295148  296612   1464   0.50%   12760   13788   1028   8.06% tinygo build -size short -o ./build/test.hex -target=pyportal -stack-size 8kb ./examples/net/http-get/
```

These don't use the precise GC yet, and the changes are too large to reasonably be caused by the added code in this PR, so I think these are actually due to interp changes. Which I think is a good thing?

PR #4496, opened 2024-10-03
compiler, runtime: enable go:wasmexport for wasip2

ydnar:
This builds on #4451 to enable `//go:wasmexport` for `wasip2` (WASI Preview 2).

- `//go:wasmexport` is now permitted and supersedes `//go:export` and `//export` if both are present.
- The `wasip2` target now defaults to `-buildmode=c-shared`, since all WASI 0.2+ programs are reactors.
- `main.main` is called by the `wasi:cli/run#run` export in a `wasmexport`.

### DIFF
--- a/src/runtime/runtime_wasip2.go
+++ b/src/runtime/runtime_wasip2.go
@@ -6,18 +6,19 @@ import (
 	"unsafe"
 
 	"internal/wasi/cli/v0.2.0/environment"
+	wasi_run "internal/wasi/cli/v0.2.0/run"
 	monotonicclock "internal/wasi/clocks/v0.2.0/monotonic-clock"
+
+	"internal/cm"
 )
 
 type timeUnit int64
 
-//export wasi:cli/run@0.2.0#run
-func __wasi_cli_run_run() uint32 {
-	// These need to be initialized early so that the heap can be initialized.
-	heapStart = uintptr(unsafe.Pointer(&heapStartSymbol))
-	heapEnd = uintptr(wasm_memory_size(0) * wasmPageSize)
-	run()
-	return 0
+func init() {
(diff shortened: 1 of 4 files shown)


src/runtime/runtime_wasip2.go
```diff
@@ -6,18 +6,19 @@ import (
...
-	// These need to be initialized early so that the heap can be initialized.
-	heapStart = uintptr(unsafe.Pointer(&heapStartSymbol))
-	heapEnd = uintptr(wasm_memory_size(0) * wasmPageSize)
-	run()
-	return 0
+func init() {
+	wasi_run.Exports.Run = func() cm.BoolResult {
+		callMain()
+		return false
+	}
```
aykevl, line 21: Interesting, any reason not to use `//export` here?

ydnar: Great question! All exports defined in WIT should use `//go:wasmexport`, including the wiring up from `wasi:cli/command#run` to `main.main`. This is the standard pattern for wiring up exports generated from `wit-bindgen-go`.

1. Developer generates Go (e.g. `wasi:http/proxy`)
2. Developer writes Go library that wraps the low-level APIs in the generated code (e.g. a `net/http` compatible shim)
3. In the shim code, developer assigns the relevant `Exports` struct field to their implementation. For example: https://github.com/ydnar/wasi-http-go/blob/76673aea4c888a0e9b36d20604b3f11a4831bdd8/wasihttp/server.go#L21-L24

ydnar: FWIW, I believe `//export` should still be used for `cabi_realloc` since the Component Model defines different reentrancy rules for low-level exports like this.

PR #4498, opened 2024-10-04
net: enable native golang linux networking

leongross:
This is a rework of the plain linux network POC https://github.com/tinygo-org/tinygo/pull/4460. This PR adds adaptive switching of network packages depending on the `GOOS`.

The CI fails due to unadjusted build tags for MacOs and WASm.
Successfully tested on `linux/amd64`.


loader/goroot.go
```diff
@@ -225,9 +225,22 @@ func needsSyscallPackage(buildTags []string) bool {
...
 
-// The boolean indicates whether to merge the subdirs. True means merge, false
-// means use the TinyGo version.
-func pathsToOverride(goMinor int, needsSyscallPackage bool) map[string]bool {
+// linuxNetworking returns whether the unmodified go linux net stack should be used
+// until the full rework of the net package is done.
+func linuxNetworking(buildTags []string) bool {
+	for _, tag := range buildTags {
+		if tag == "linux" {
+			return true
```
aykevl: Unfortunately, `linux` is also set on baremetal systems (because we have to pretend to be _some_ OS). So this isn't quite right.

I think checking for the equivalent of `linux && !baremetal && !nintendoswitch && !tinygo.wasm` would work here.

.gitmodules
```diff
@@ -1,41 +1,40 @@
 [submodule "lib/nrfx"]
-	path = lib/nrfx
-	url = https://github.com/NordicSemiconductor/nrfx.git
+    path = lib/nrfx
+    url = https://github.com/NordicSemiconductor/nrfx.git
```
aykevl: Can you please undo all the whitespace changes in this file?
Also, some seem to be reordered. Was this intentional?

src/runtime/sync.go
```diff
@@ -4,10 +4,14 @@ package runtime
 
 //go:linkname semacquire internal/poll.runtime_Semacquire
 func semacquire(sema *uint32) {
-	panic("todo: semacquire")
+	// TODO the "net" pkg calls this, so panic() isn't an option.  Right
+	// now, just ignore the call.
+	// panic("todo: semacquire")
```
aykevl: Don't do this. There is a very good reason this panics: it is a reminder this should be implemented when it is used.
So either restore the panic, or actually implement `semacquire` and `semrelease`.

(It's probably easier to implement the internal/poll package instead in a way that works with TinyGo - but even that might be rather complex).

src/os/exec.go
```diff
@@ -47,6 +47,10 @@ func (p *ProcessState) Sys() interface{} {
 	return nil // TODO
 }
 
+func (p *ProcessState) Exited() bool {
+	return false // TODO
+}
```
aykevl, line 58: Please don't implement functions like this. This could break existing programs that rely on `Exited` returning a correct value.

(This is different from `Sys()` for example, where programs would type assert anyway and the type assert would probably fail - therefore clearly indicating the issue).

aykevl (commented):
How did you test this?

To accept this into TinyGo, I'd like to see at least some tests. For example, you can create a new test like testdata/net.go (and add it to main_test.go for Linux only). It doesn't have to be exhaustive, but for example you can start a socket server (`net.Listen`) in a goroutine and try to connect to it from the same program.

.gitmodules
leongross: Neither the spaces nor the order were intentional, my IDE did the formatting. I will revert these chnages.

src/os/exec.go
leongross: So what do you propose to do here? The ProcessState afaik is just a stub, so the functions implemented over this struct should also be stubs. Re-implementing the process model of tinygo would definitely overshoot this PRs scope.

leongross:
@aykevl I will add some basic tests. Since I already did some testing (I hope) it should be fairly easy to integrate this here.

src/runtime/sync.go
leongross: Maybe I am missing something here, but since tinygo at the moment only supports a single os thread, it would be sufficient to implement this using a simple mutex, right? Making this code panic will eventually reduce usability for this drastically.

aykevl: No. All this API provides is a `uint32` to store state. That's not enough to store a queue of waiting goroutines and the counter as would be needed for proper semaphore support. I don't think this can even be implemented in the current scheduler - at leat, not without some expensive data structures on the side. It is relatively easy to implement with a futex though, as I have in my threading PR: #4559

leongross: I see that PR #4559 is still a wip. Maybe we should consider pulling out the futex functionality first to make some progress here?

src/syscall/forklock.go
```diff
@@ -0,0 +1,38 @@
...
+//
+// package tinygo.org/x/drivers/examples/net/tcpclient
+//        imports bytes
+//        imports io
+//        imports sync
+//        imports internal/task
+//        imports runtime/interrupt
+//        imports device/arm
+//        imports syscall
+//        imports sync: import cycle not allowed
```
aykevl: This shouldn't be happening: you have excluded the `baremetal` build tag so `device/arm` should not be imported.

src/syscall/forklock.go
```diff
@@ -0,0 +1,38 @@
...
+//
+// So for now, make our own stubbed-out ForkLock that doesn't use sync..
+
+type forklock struct{}
+
+func (f forklock) RLock()   {}
+func (f forklock) RUnlock() {}
+
+var ForkLock forklock
```
aykevl: Don't. You're basically removing all safety here and breaking the API contract.

src/runtime/sync.go
```diff
@@ -1,13 +1,72 @@
...
+	"unsafe"
+)
+
 // This file contains stub implementations for internal/poll.
+// The official golang implementation states:
+//
+// "That is, don't think of these as semaphores.
+// Think of them as a way to implement sleep and wakeup
+// such that every sleep is paired with a single wakeup,
+// even if, due to races, the wakeup happens before the sleep."
```
aykevl, line 17: Ah, that's useful to know. So they're not really semaphores but something like it.

src/runtime/sync.go
```diff
@@ -1,13 +1,72 @@
...
+const semTabSize = 251
+
+type semTable [semTabSize]struct {
+	root semaRoot
+	pad  [64 - unsafe.Sizeof(semaRoot{})]byte // only 64 x86_64, make this variable
+}
+
+func (t *semTable) rootFor(addr *uint32) *semaRoot {
+	return &t[(uintptr(unsafe.Pointer(addr))>>3)%semTabSize].root
+}
```
aykevl: Can you explain what this is doing, and what it's for?

src/runtime/sync.go
```diff
@@ -1,13 +1,72 @@
...
 
 //go:linkname semacquire internal/poll.runtime_Semacquire
 func semacquire(sema *uint32) {
-	panic("todo: semacquire")
+	if cansemacquire(sema) {
+		return
+	}
```
aykevl: This doesn't look like a complete implementation?
You need to wait for the other goroutine to do a `semrelease`. (This would be easy with multithreading and real OS futexes, but they don't exist yet).
Either you can busy loop with `Gosched()` (terrible idea) or put the waiting goroutine in a hashmap to be awoken by `semrelease` (slightly less terrible idea).

src/runtime/sync.go
```diff
@@ -1,13 +1,72 @@
...
 }
 
 //go:linkname semrelease internal/poll.runtime_Semrelease
 func semrelease(sema *uint32) {
-	panic("todo: semrelease")
+	root := semtable.rootFor(sema)
+	atomic.AddUint32(sema, 1)
+	if root.nwait.Load() == 0 {
+		return
+	}
```
aykevl: You are not awaking the waiting goroutine here.

testdata/net.go
```diff
@@ -0,0 +1,82 @@
...
+			fmt.Printf("error: received data does not match sent data: '%s' != '%s'\n", recvBuf.String(), string(testDialListenData))
+			return
+		}
+		conn.Close()
+
+		return
+	}()
+
+	// hacky way to wait for the listener to start
+	time.Sleep(1 * time.Second)
```
aykevl: Can you fix this? One easy way would be using a channel.

testdata/net.go
```diff
@@ -0,0 +1,82 @@
+package main
+
+import (
+	"bytes"
+	"fmt"
```
aykevl: Can you try to remove the `fmt` package? It's a bit heavy and slows down tests.
I think you can use `println` in all cases.

testdata/net.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+var (
+	testDialListenData = []byte("Hello tinygo :)")
+)
+
+func TestDialListen() {
+	// listen thread
+	go func() {
+		ln, err := net.Listen("tcp", ":"+strconv.FormatInt(TEST_PORT, 10))
```
aykevl: Can you use port 0 and pass the used port number back to the main goroutine? That will avoid issues with multiple tests running at the same time and all using port 9000.

testdata/net.go
```diff
@@ -0,0 +1,82 @@
...
+			return
+		}
+
+		conn, err := ln.Accept()
+		if err != nil {
+			fmt.Printf("error accepting: %v\n", err)
+			return
+		}
+
+		recvBuf.Reset()
```
aykevl: Why do you need to reset this buffer?

testdata/net.go
```diff
@@ -0,0 +1,82 @@
...
+		_, err = conn.Read(recvBuf.Bytes())
```
aykevl: That looks way more complicated than it needs to be. Why not just use a plain byte slice?
(Also, `conn.Read` is not guaranteed to read all available data. I think you may need `io.ReadFull` or `io.ReadAll`).

testdata/net.go
```diff
@@ -0,0 +1,82 @@
...
+
+	sendBuf.Reset()
+	fmt.Fprint(sendBuf, testDialListenData)
+	conn, err := net.Dial("tcp4", "127.0.0.1:"+strconv.FormatInt(TEST_PORT, 10))
+	if err != nil {
+		fmt.Printf("error dialing: %v\n", err)
+		return
+	}
+
+	if _, err = conn.Write(sendBuf.Bytes()); err != nil {
```
aykevl: Why not just
```suggestion
	if _, err = conn.Write(testDialListenData); err != nil {
```

testdata/net.go
```diff
@@ -0,0 +1,82 @@
...
+		if err != nil {
+			fmt.Printf("error reading: %v\n", err)
+			return
+		}
+
+		// TODO: this is racy
+		if recvBuf.String() != string(testDialListenData) {
+			fmt.Printf("error: received data does not match sent data: '%s' != '%s'\n", recvBuf.String(), string(testDialListenData))
+			return
+		}
```
aykevl: What's racy about this?

.gitmodules
aykevl, line 1: Can you restore this file? It doesn't seem to have anything to do with this PR.

GNUmakefile
```diff
@@ -343,6 +343,7 @@ TEST_PACKAGES_FAST = \
 	net/http/internal/ascii \
 	net/mail \
 	os \
+	os/signal \
```
aykevl: Please remove the first commit of this PR (you can use `git rebase -i dev`).

src/runtime/netpoll_generic.go
```diff
@@ -0,0 +1,31 @@
...
+//
+// No heap pointers.
+// For linux to call create Fds with a pollDesc, it needs a ctxRuntime pointer, so use the original pollDesc struct.
+// On linux we have a heap.
+type pollDesc struct{}
+
+//go:linkname poll_runtime_pollReset internal/poll.runtime_pollReset
+func poll_runtime_pollReset(pd *pollDesc, mode int) int {
+	println("poll_runtime_pollReset not implemented", pd, mode)
+	return 1
```
aykevl: This is also a bad idea. Either implement it, or exit with an error.

src/runtime/poll.go
```diff
@@ -4,20 +4,21 @@ package runtime
 
 //go:linkname poll_runtime_pollServerInit internal/poll.runtime_pollServerInit
 func poll_runtime_pollServerInit() {
-	panic("todo: runtime_pollServerInit")
+	// fmt.Printf("poll_runtime_pollServerInit not implemented, skipping panic\n")
```
aykevl: If you put in debugging code, please remove it before sending in a PR. It makes the diff bigger and more work to review.
Generally, take a look at the PR diff and see if there's any part of it that is not necessary and can be removed (such as the changes to `.gitmodules` and the signal changes that are still part of this PR).
(Also, how do you know that you can just remove this panic?)

src/os/exec.go
aykevl: Ah you are correct.

aykevl (changes requested):

testdata/net.go
leongross: I rand into situations where `recvBuf.String()` would return an empty string, because (I assume) the recv wasn't done yet?
EDIT: Using `io.ReadFull` solved this issues.

src/syscall/forklock.go
leongross: Sounds reasonable. This is leftover code from the old net rework PR I based this one on, I dragged it with me and forgot to remove it.

src/runtime/sync.go
leongross: @aykevl does this sound like a good approach to you?

src/syscall/forklock.go
leongross: This again was stub code to get the poc working. Since your PR introduces threat safe mutexes we could replace it with

 ```suggestion
var ForkLock sync.RWMutex
```

File: src/runtime/netpoll_generic.go (hunk 19 above)
Comment (leongross) [reply src/runtime/netpoll_generic.go]: Returning 1 here indicates an error for this context. I added the proper definitions that should make it more clear.

File: src/runtime/sync.go (hunk 7 above)
Comment (leongross) [reply src/runtime/sync.go:17]: https://cs.opensource.google/go/go/+/master:src/runtime/sema.go;l=12?q=semac&ss=go%2Fgo

File: main_test.go [hunk 21]
```diff
@@ -341,8 +341,7 @@ func emuCheck(t *testing.T, options compileopts.Options) {
 	}
 	if spec.Emulator != "" {
 		emulatorCommand := strings.SplitN(spec.Emulator, " ", 2)[0]
-		_, err := exec.LookPath(emulatorCommand)
-		if err != nil {
+		if _, err := exec.LookPath(emulatorCommand); err != nil {
```
aykevl: Please don't make unrelated changes in the PR. While this is probably a good idea, it would be better to make a separate cleanup PR instead of putting everything in one huge PR.

dgryski: A nice phrase I picked up somewhere was "clean or fix, not both".

src/runtime/sync.go
```diff
@@ -1,13 +1,52 @@
...
+// implement the modularity that the semacquire/semacquire1 implementation model
+// offers, which in fact is emitted here entirely.
+// This means we assume the following constant settings from the golang standard
+// library: lifo=false,profile=semaBlock,skipframe=0,reason=waitReasonSemaquire
 
 //go:linkname semacquire internal/poll.runtime_Semacquire
 func semacquire(sema *uint32) {
-	panic("todo: semacquire")
+	var semaBlock futex.Futex
+	semaBlock.Store(*sema)
```
aykevl, line 29: You are working with a copy of the data. This semaphore implementation doesn't actually work.

aykevl (changes requested):
What's the status of this PR?

PR #4499, opened 2024-10-04
compiler, runtime: enable go:wasmexport for wasip2

ydnar:
This builds on #4451 to enable `//go:wasmexport` for `wasip2` (WASI Preview 2).

- `//go:wasmexport` is now permitted and supersedes `//go:export` and `//export` if both are present.
- The `wasip2` target now defaults to `-buildmode=c-shared`, since all WASI 0.2+ programs are reactors.
- `main.main` is called by the `wasi:cli/run#run` export in a `wasmexport`.

### DIFF
--- a/src/runtime/runtime_wasip2.go
+++ b/src/runtime/runtime_wasip2.go
@@ -6,18 +6,19 @@ import (
 	"unsafe"
 
 	"internal/wasi/cli/v0.2.0/environment"
+	wasi_run "internal/wasi/cli/v0.2.0/run"
 	monotonicclock "internal/wasi/clocks/v0.2.0/monotonic-clock"
+
+	"internal/cm"
 )
 
 type timeUnit int64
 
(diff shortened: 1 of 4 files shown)


src/runtime/runtime_wasip2.go
```diff
@@ -6,18 +6,19 @@ import (
 	"unsafe"
 
 	"internal/wasi/cli/v0.2.0/environment"
+	wasi_run "internal/wasi/cli/v0.2.0/run"
```
dgryski: The underscore here looks weird.  Maybe `wasiclirun` ?

dgryski (approved):
LGTM, except for the renamed import (which I'm fine to handle later)

ydnar:
Want me to rename back?

dgryski:
Yes, but if you just want this merged we can do it in a separate PR later.

dgryski (approved):
LGTM

PR #4501, opened 2024-10-05
compiler: conform to latest iteration of wasm types proposal

ydnar:
Updates to current version of the wasm types proposal at https://github.com/golang/go/issues/66984. 

Changes include:

- Removal of `int` and `uint` as allowed types.
- Restriction of `int8`, `uint8`, `int16`, and `uint16` to struct fields, arrays, or pointer values.
- Requires `struct` types in wasm params with 1 or more fields to include `structs.HostLayout`. It enforces this for Go 1.23 or later.
    - TODO: revise this to minimum Go 1.24?

### DIFF
--- a/compiler/symbol.go
+++ b/compiler/symbol.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	"github.com/tinygo-org/tinygo/compiler/llvmutil"
+	"github.com/tinygo-org/tinygo/goenv"
 	"github.com/tinygo-org/tinygo/loader"
 	"golang.org/x/tools/go/ssa"
 	"tinygo.org/x/go-llvm"
@@ -422,14 +423,14 @@ func (c *compilerContext) checkWasmImportExport(f *ssa.Function, pragma string) 
 		c.addError(f.Signature.Results().At(1).Pos(), fmt.Sprintf("%s: too many return values", pragma))
 	} else if f.Signature.Results().Len() == 1 {
 		result := f.Signature.Results().At(0)
-		if !isValidWasmType(result.Type(), siteResult) {
+		if !c.isValidWasmType(result.Type(), siteResult) {
 			c.addError(result.Pos(), fmt.Sprintf("%s: unsupported result type %s", pragma, result.Type().String()))
 		}
 	}
 	for _, param := range f.Params {
 		// Check whether the type is allowed.
 		// Only a very limited number of types can be mapped to WebAssembly.
-		if !isValidWasmType(param.Type(), siteParam) {
+		if !c.isValidWasmType(param.Type(), siteParam) {
 			c.addError(param.Pos(), fmt.Sprintf("%s: unsupported parameter type %s", pragma, param.Type().String()))
 		}
 	}
@@ -442,7 +443,7 @@ func (c *compilerContext) checkWasmImportExport(f *ssa.Function, pragma string) 
 //
 // This previously reflected the additional restrictions documented here:
 // https://github.com/golang/go/issues/59149
-func isValidWasmType(typ types.Type, site wasmSite) bool {
+func (c *compilerContext) isValidWasmType(typ types.Type, site wasmSite) bool {
 	switch typ := typ.Underlying().(type) {
 	case *types.Basic:
 		switch typ.Kind() {
@@ -461,19 +462,35 @@ func isValidWasmType(typ types.Type, site wasmSite) bool {
 			return site == siteParam || site == siteIndirect
 		}
 	case *types.Array:
-		return site == siteIndirect && isValidWasmType(typ.Elem(), siteIndirect)
+		return site == siteIndirect && c.isValidWasmType(typ.Elem(), siteIndirect)
 	case *types.Struct:
 		if site != siteIndirect {
 			return false
 		}
+		// Structs with no fields do not need structs.HostLayout
+		if typ.NumFields() == 0 {
+			return true
+		}
+		hasHostLayout := true // default to true before detecting Go version
+		// (*types.Package).GoVersion added in go1.21
+		if gv, ok := any(c.pkg).(interface{ GoVersion() string }); ok {
+			if goenv.Compare(gv.GoVersion(), "go1.23") >= 0 {
+				hasHostLayout = false // package structs added in go1.23
+			}
+		}
 		for i := 0; i < typ.NumFields(); i++ {
-			if !isValidWasmType(typ.Field(i).Type(), siteIndirect) {
+			ftyp := typ.Field(i).Type()
+			if ftyp.String() == "structs.HostLayout" {
+				hasHostLayout = true
+				continue
+			}
+			if !c.isValidWasmType(ftyp, siteIndirect) {
 				return false
 			}
 		}
-		return true
+		return hasHostLayout
 	case *types.Pointer:
-		return isValidWasmType(typ.Elem(), siteIndirect)
+		return c.isValidWasmType(typ.Elem(), siteIndirect)
 	}
 	return false
 }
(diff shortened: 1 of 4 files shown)


ydnar:
@dgryski @aykevl would you prefer I split this into two parts, one that upgrades the `wasm-tools-go` dependency and regenerates the bindings, and a separate one with the wasm types changes to package `compiler`?

deadprogram:
> would you prefer I split this into two parts, one that upgrades the wasm-tools-go dependency and regenerates the bindings, and a separate one with the wasm types changes to package compiler?

I personally would prefer that. :smile_cat:

ydnar:
> > would you prefer I split this into two parts, one that upgrades the wasm-tools-go dependency and regenerates the bindings, and a separate one with the wasm types changes to package compiler?
> 
> I personally would prefer that. 😸

Here you go: https://github.com/tinygo-org/tinygo/pull/4529

deadprogram:
```
# github.com/tinygo-org/tinygo/compiler
compiler/symbol.go:469:33: c.pkg.GoVersion undefined (type *types.Package has no field or method GoVersion)
```
https://app.circleci.com/pipelines/github/tinygo-org/tinygo/11640/workflows/e46166fe-40b9-410a-95a9-d49be472b0d1/jobs/44676?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary&invite=true#step-112-0_39

deadprogram:
@ydnar note the CI failure...

ydnar:
What's the minimum version of Go supported by TinyGo?

Looks like this field was added in 1.21: https://pkg.go.dev/go/types#Package.GoVersion

deadprogram:
"Officially" should be same as Go itself. But we have been able to keep backwards compat quite a few versions of both Go and LLVM, as shown by the CircleCI tests.

ydnar:
> "Officially" should be same as Go itself. But we have been able to keep backwards compat quite a few versions of both Go and LLVM, as shown by the CircleCI tests.

How does this project define compatibility?

1. TinyGo can be built with version N of Go?
2. or TinyGo can be *used* with version N of Go?

Asking because what % of users build TinyGo from source, and what % of those users are using older versions of Go? Versus what % of end-users of TinyGo who use a binary distribution use older versions of Go?

aykevl:
Usually we keep compatibility as long as it isn't too much of an issue to support. We tend to drop older versions when it's getting annoying to support these older versions. It's always surprising how many people rely on older Go versions but still expect to use a newer TinyGo version.

Also, to be clear: because it makes CI easier "can be built" and "can be used with" is usually the same version. The lower bound is currently LLVM 15 and Go 1.19:

https://github.com/tinygo-org/tinygo/blob/ef4f46f1d1550beb62324d750c496b2b4a7f76d0/.circleci/config.yml#L121-L123

goenv/version.go
```diff
@@ -34,7 +34,12 @@ func GetGorootVersion() (major, minor int, err error) {
 	if err != nil {
 		return 0, 0, err
 	}
+	return ParseGoVersion(s)
+}
 
+// ParseGoVersion returns the major and minor version for a given Go version.
+// The input must be in the form of "go$MAJOR.$MINOR$TRAILING", e.g. "go1.23" or "go1.24.0-rc2".
+func ParseGoVersion(s string) (major, minor int, err error) {
```
aykevl: Please see https://github.com/tinygo-org/tinygo/pull/4520 where I made a very similar change.

compiler/symbol.go
```diff
@@ -453,19 +456,32 @@ func isValidWasmType(typ types.Type, site wasmSite) bool {
...
 		if site != siteIndirect {
 			return false
 		}
+		// Structs with no fields do not need structs.HostLayout
+		if typ.NumFields() == 0 {
+			return true
+		}
+		var hasHostLayout bool
+		if !goenv.WantGoVersion(c.pkg.GoVersion(), 1, 23) {
+			hasHostLayout = true // package structs does not exist before go1.23
```
aykevl: I guess one way to work around this is to put this in a separate file with build tags. Together with https://github.com/tinygo-org/tinygo/pull/4501, the compiler won't try to compile projects with a Go version that's newer than the one TinyGo was compiled with (which means that when TinyGo was compiled with Go before 1.23, it will refuse to even compile code that requires Go 1.23+).

It's a bit ugly, but will keep compatibility for a bit longer.

goenv/version.go
```diff
@@ -54,7 +59,18 @@ func GetGorootVersion() (major, minor int, err error) {
...
+}
+
+// WantGoVersion returns true if Go version s is >= major and minor.
+// Returns false if s is not a valid Go version string. See [ParseGoVersion] for more information.
+func WantGoVersion(s string, major, minor int) bool {
+	ma, mi, err := ParseGoVersion(s)
+	if err != nil {
+		return false
+	}
+	return ma >= major && mi >= minor
```
aykevl: What if we're on Go 2.2 and the code is written for Go 1.20? It would fail this check, though 2.20 would work.

```suggestion
	return ma > major || (ma == major && mi >= minor)
```

All very hypothetical of course, since I don't think there will ever be a Go 2.0 (and the compatibility situation may not be what we expect when that happens). So perhaps we can just ignore the possibility of Go 2.0 for now.

aykevl (commented):
LGTM, though please see https://github.com/tinygo-org/tinygo/pull/4520 because it will conflict.

aykevl:
I guess we could drop support for Go 1.19 and Go 1.20. That would leave Go 1.21-1.23 supported. I find it difficult to guess how many people would have an issue with this (note that Go 1.19 is the Go version in Debian 12).

compiler/symbol.go
ydnar: I can do this.

ydnar: @aykevl thanks! I rebased and my branch now uses `Parse`. I included the tests for `Parse` in my PR as well.

ydnar:
@aykevl I updated this quite a bit. Can you take a look?

ydnar:
I can split this into two PRs, one for the package `goenv` changes (`Parse` and `Compare`), and a second for the wasm types.

deadprogram:
> I can split this into two PRs, one for the package goenv changes (Parse and Compare), and a second for the wasm types.

Sorry to be a pain, but I am fairly certain that will make it a lot easier to review. Thank you!

ydnar:
> > I can split this into two PRs, one for the package goenv changes (Parse and Compare), and a second for the wasm types.
> 
> Sorry to be a pain, but I am fairly certain that will make it a lot easier to review. Thank you!

Done! See #4536.

deadprogram:
@ydnar I think this PR can be rebased now, if you please.

ydnar:
@deadprogram done!

deadprogram (approved):
Did some testing and my current code still compiles and runs without issue.

LGTM!

compiler/symbol.go
```diff
@@ -459,19 +462,35 @@ func isValidWasmType(typ types.Type, site wasmSite) bool {
...
 		}
+		// Structs with no fields do not need structs.HostLayout
+		if typ.NumFields() == 0 {
+			return true
+		}
+		hasHostLayout := true // default to true before detecting Go version
+		// (*types.Package).GoVersion added in go1.21
+		if gv, ok := any(c.pkg).(interface{ GoVersion() string }); ok {
+			if goenv.Compare(gv.GoVersion(), "go1.23") >= 0 {
+				hasHostLayout = false // package structs added in go1.23
```
aykevl, line 478: Interesting solution, I hadn't thought of that! Much cleaner than using build tags.

aykevl (approved):

aykevl:
Was this PR also meant to be squashed?

compiler/symbol.go
ydnar: Yeah! It's great when the only difference is a new method.

deadprogram:
Now merging, thank you @ydnar for working on this and to @aykevl for helping review.

PR #4503, opened 2024-10-06
runtime: implement newcoro, coroswitch to support package iter

eliasnaur:


dgryski:
Do we want to add any tests here?

eliasnaur:
The standard library `iter` tests fail:

```sh
$ tinygo test iter
panic: runtime error at 0x00000001001d3680: nil pointer dereference
FAIL	iter	0.279s
```

(fixed by https://go-review.googlesource.com/c/go/+/617917, but not released).

Is there somewhere to add test exceptions to standard library tests, without copying them?

dgryski:
We don't have a framework for "run this stdlib test but skip=REGEX", which I guess is what we'd need.

(When we upgrade to the next Go version we can hopefully just add `iter` to the list of support stdlib packages.)

eliasnaur:
I looked closer; the stdlib `iter` tests also rely on `runtime.NumGoroutine` not lying.

I've added a smoketest to testdata/go1.23 in the meantime.

dgryski:
(Filed https://github.com/tinygo-org/tinygo/issues/4507 )

src/runtime/coro.go
```diff
@@ -0,0 +1,31 @@
...
+// package iter.
+
+type coro struct {
+	f  func(*coro)
+	ch chan struct{}
+}
+
+//go:linkname newcoro
+
+func newcoro(f func(*coro)) *coro {
```
aykevl, line 13: It doesn't look like this should work?
The syntax is `//go:linkname <functionname> <linkname>` and should be immediately before the function. Otherwise they get ignored.

EDIT: oh I see, you want to make these functions available for linking? This hasn't been implemented in TinyGo yet, though putting these pragmas in won't harm either.

testdata/go1.23/main.go
```diff
@@ -8,6 +10,15 @@ func testFuncRange(f func(yield func(int) bool)) {
...
 	}
+	next, stop := iter.Pull(f)
+	defer stop()
+	for {
+		i, ok := next()
+		if !ok {
+			break
+		}
+		println(i)
+	}
```
aykevl: Can you put this in a separate function? `testFuncRange` is for testing range over func, you could add a new `testCoro` for example.

src/runtime/coro.go
eliasnaur: Yeah, I just copied those lines from Big Go's coro.go. I'll leave them.

testdata/go1.23/main.go
eliasnaur: Done.

deadprogram:
@aykevl anything else needed here or is it good to merge?

aykevl:
Thank you!

PR #4506, opened 2024-10-07
Add MemStats.HeapAlloc

dgryski:
This is @bgould's https://github.com/tinygo-org/tinygo/pull/3424 plus some infra to track those metrics for gc_blocks.  Still need to add in logic for gc_leaking.


aykevl (approved):

dgryski:
The binary size jump seems large to me.

PR #4510, opened 2024-10-08
targets: add bulk memory flags to wasm-unknown

deadprogram:
This PR modifies the `wasm-unknown` target to add bulk memory flags since basically every runtime has it now.


dgryski (approved):
LGTM

deadprogram:
Apparently we are a bit stricter about what we allow. Removing the `-mno-bulk-memory` appears to be what actually did what I wanted.

deadprogram:
Thank you for review @dgryski now merging.

Issue #4515, opened 2024-10-11
ci: error while loading shared libraries: libLLVM-17.so.1: cannot open shared object file: No such file or directory


leongross:
In the ci the `nix-test` step fails with the following error [1]

```
/home/runner/go/bin/tinygo: error while loading shared libraries: libLLVM-17.so.1: cannot open shared object file: No such file or directory
```

The `Restore LLVM source cache` uses LLVM v.18, so we should bump the version here.

[1]: https://github.com/tinygo-org/tinygo/actions/runs/11272306601/job/31346952510?pr=4378#step:8:108

aykevl:
> The `Restore LLVM source cache` uses LLVM v.18, so we should bump the version here.

We use LLVM 18, so this is correct.

PR #4516, opened 2024-10-11
nix: use LLVM 18 instead of LLVM 17

aykevl:
This should hopefully fix some CI issues we're currently having: https://github.com/tinygo-org/tinygo/issues/4515

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -15,6 +15,11 @@ jobs:
   nix-test:
     runs-on: ubuntu-latest
     steps:
+      - name: Uninstall system LLVM
+        # Hack to work around issue where we still include system headers for
+        # some reason.
+        # See: https://github.com/tinygo-org/tinygo/pull/4516#issuecomment-2416363668
+        run: sudo apt-get remove llvm-18
       - name: Checkout
         uses: actions/checkout@v4
       - name: Pull musl
--- a/flake.lock
+++ b/flake.lock
@@ -20,16 +20,16 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1703068421,
-        "narHash": "sha256-WSw5Faqlw75McIflnl5v7qVD/B3S2sLh+968bpOGrWA=",
+        "lastModified": 1728500571,
+        "narHash": "sha256-dOymOQ3AfNI4Z337yEwHGohrVQb4yPODCW9MDUyAc4w=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "d65bceaee0fb1e64363f7871bc43dc1c6ecad99f",
+        "rev": "d51c28603def282a24fa034bcb007e2bcb5b5dd0",
         "type": "github"
       },
       "original": {
         "id": "nixpkgs",
-        "ref": "nixos-23.11",
+        "ref": "nixos-24.05",
         "type": "indirect"
       }
     },
--- a/flake.nix
+++ b/flake.nix
@@ -35,7 +35,7 @@
   inputs = {
     # Use a recent stable release, but fix the version to make it reproducible.
     # This version should be updated from time to time.
-    nixpkgs.url = "nixpkgs/nixos-23.11";
+    nixpkgs.url = "nixpkgs/nixos-24.05";
     flake-utils.url = "github:numtide/flake-utils";
   };
   outputs = { self, nixpkgs, flake-utils }:
@@ -49,11 +49,11 @@
           buildInputs = [
             # These dependencies are required for building tinygo (go install).
             go
-            llvmPackages_17.llvm
-            llvmPackages_17.libclang
+            llvmPackages_18.llvm
+            llvmPackages_18.libclang
             # Additional dependencies needed at runtime, for building and/or
             # flashing.
-            llvmPackages_17.lld
+            llvmPackages_18.lld
             avrdude
             binaryen
             # Additional dependencies needed for on-chip debugging.
@@ -68,7 +68,7 @@
             # Without setting these explicitly, Homebrew versions might be used
             # or the default `ar` and `nm` tools might be used (which don't
             # support wasi).
-            export CLANG="clang-17 -resource-dir ${llvmPackages_17.clang.cc.lib}/lib/clang/17"
+            export CLANG="clang-18 -resource-dir ${llvmPackages_18.clang.cc.lib}/lib/clang/18"
             export LLVM_AR=llvm-ar
             export LLVM_NM=llvm-nm
 
@@ -77,7 +77,7 @@
             export MD5SUM=md5sum
 
             # Ugly hack to make the Clang resources directory available.
-            export GOFLAGS="\"-ldflags=-X github.com/tinygo-org/tinygo/goenv.clangResourceDir=${llvmPackages_17.clang.cc.lib}/lib/clang/17\" -tags=llvm17"
+            export GOFLAGS="\"-ldflags=-X github.com/tinygo-org/tinygo/goenv.clangResourceDir=${llvmPackages_18.clang.cc.lib}/lib/clang/18\" -tags=llvm18"
           '';
         };
       }


leongross (commented):
I do not know a lot about Nix, but if we do changes in the flake, don't we also have to update the [`flake.lock`](https://github.com/tinygo-org/tinygo/blob/release/flake.lock#L32)?

aykevl:
Yes, apparently! I ran `nix develop` locally and it automatically updated flake.lock.

aykevl:
It's still failing, and I don't know why because it works fine on my system:

```
/home/runner/go/bin/tinygo: symbol lookup error: /home/runner/go/bin/tinygo: undefined symbol: LLVMInitializeXtensaTarget, version LLVM_18.1
```

leongross:
Here also the linux wasm tests seem to fail, this has been observed in many recent PRs.

deadprogram:
```
/home/runner/go/bin/tinygo: symbol lookup error: /home/runner/go/bin/tinygo: undefined symbol: LLVMInitializeXtensaTarget, version LLVM_18.1
```

Any nix experts able to help out, please?

sylv-io:
I have tried your PR and have not been able to reproduce the symbol lookup error in the workflow either locally or with [act -w nix-test](https://github.com/nektos/act). Can you re-run the workflow and see if the error persists?

deadprogram:
I just retried the workflow and same error @sylv-io

leongross:
@katexochen do you know what is going on here?

sylv-io:
A rather strange bug indeed. I was able to reproduce this bug and access the binary by [forking tinygo and modifying the workflow](https://github.com/9elements/tinygo/actions/runs/11353913986).

Comparing the resulting binary with ghidra and ghidriff shows that the binary built by nix within the github runner adds support for two additional LLVM targets: M68k and Xtensa.

Here is a decompiler diff:
```diff
 void _cgo_ec38bf0848b9_Cfunc_LLVMInitializeAllTargets(void)
 
 {
   LLVMInitializeAArch64Target();
   LLVMInitializeAMDGPUTarget();
   LLVMInitializeARMTarget();
   LLVMInitializeAVRTarget();
   LLVMInitializeBPFTarget();
   LLVMInitializeHexagonTarget();
   LLVMInitializeLanaiTarget();
   LLVMInitializeLoongArchTarget();
   LLVMInitializeMipsTarget();
   LLVMInitializeMSP430Target();
   LLVMInitializeNVPTXTarget();
   LLVMInitializePowerPCTarget();
   LLVMInitializeRISCVTarget();
   LLVMInitializeSparcTarget();
   LLVMInitializeSystemZTarget();
   LLVMInitializeVETarget();
   LLVMInitializeWebAssemblyTarget();
   LLVMInitializeX86Target();
   LLVMInitializeXCoreTarget();
+  LLVMInitializeM68kTarget();
+  LLVMInitializeXtensaTarget();
   return;
 }
```

IIUC the Xtensa target is not supported by default in libLLVM, causing the runtime symbol lookup error. I cannot explain why this only happens in the GitHub runner, but perhaps this information can help you further.

Here is the full ghidriff report: 
[tinygo.local-tinygo.github.ghidriff.md](https://github.com/user-attachments/files/17385126/tinygo.local-tinygo.github.ghidriff.md)

sylv-io:
@aykevl:

Okay, I understand the problem. tinygo is built using the system llvm headers, but links to the dynamic library provided by nix.

diff /nix/store/8bbp6b9wb0lc4kiv7a6qrljsp8mbp4lf-llvm-18.1.8-dev/include/llvm/Config/Targets.def /usr/include/llvm-18/llvm/Config/Targets.def
```
44a45,46
> LLVM_TARGET(M68k)
> LLVM_TARGET(Xtensa)
```

Removing the llvm system packages fixes the workflow, but this behavior is unintended and we should probably find out why the build is not using the nix header instead. 

Here the workaround:
Commit: https://github.com/9elements/tinygo/commit/531178741120e542eaba2da835b133e660246453
Workflow: https://github.com/9elements/tinygo/actions/runs/11363440497

sylv-io:
In other words, we should fix the flake to use/prioritize the nix header instead. :+1:

sylv-io:
By patching Target.def with an error macro, we can see that [go-llvm](https://github.com/tinygo-org/go-llvm) is accessing the system headers: https://github.com/9elements/tinygo/actions/runs/11364600184/job/31610962741

I assume this is due to the hardcoded cgo flags in `llvm_config_linux_llvm18.go`: https://github.com/tinygo-org/go-llvm/blob/master/llvm_config_linux_llvm18.go#L5

aykevl:
@sylv-io thank you! That is indeed a workaround (CI passes now!)

I'm not sure how this should ideally be fixed: ideally we'd want to keep `go install` working without any special flags. One option would be to use `-idirafter` to add the include directory after all the other directories, so that the Nix LLVM headers get searched first (hopefully!).

sylv-io:
> @sylv-io thank you! That is indeed a workaround (CI passes now!)

np :grin: 

> I'm not sure how this should ideally be fixed: ideally we'd want to keep `go install` working without any special flags. One option would be to use `-idirafter` to add the include directory after all the other directories, so that the Nix LLVM headers get searched first (hopefully!).

Yes, it looks like a difficult problem that probably needs to be addressed in `go-llvm`. Even if we define the correct CGO_CPPFLAGS in the flake shellHooks, the cgo flag CPPFLAGS in `llvm_config_linux_llvm18.go` seems to take precedence...

e.g.
```
export CGO_CPPFLAGS="-I${llvmPackages_18.llvm.dev}/include"
```

[go build -x -a -v output:](https://github.com/9elements/tinygo/actions/runs/11365754126/job/31614621446)
```
[...]
TERM='dumb' CGO_LDFLAGS='"-O2" "-g" "-L/usr/lib/llvm-18/lib" "-lLLVM-18"' /nix/store/j9hk9x78z301jqldfyvw72z24vb992qi-go-1.22.6/share/go/pkg/tool/linux_amd64/cgo -objdir $WORK/b099/ -importpath tinygo.org/x/go-llvm -- -I/nix/store/8bbp6b9wb0lc4kiv7a6qrljsp8mbp4lf-llvm-18.1.8-dev/include -I/usr/include/llvm-18 -I/usr/include/llvm-c-18 -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -I $WORK/b099/ -O2 -g ./analysis.go ./bitreader.go ./bitwriter.go ./dibuilder.go ./executionengine.go ./ir.go ./irreader.go ./linker.go ./llvm_config_linux_llvm18.go ./passes.go ./support.go ./target.go ./version.go
# tinygo.org/x/go-llvm
In file included from /usr/include/llvm-c-18/llvm-c/Target.h:43,
                 from /usr/include/llvm-c-18/llvm-c/ExecutionEngine.h:23,
                 from ../../../go/pkg/mod/tinygo.org/x/go-llvm@v0.0.0-20240627184919-3b50c76783a8/executionengine.go:17:
/usr/include/llvm-18/llvm/Config/Targets.def:1:2: error: #error ("nope")
    1 | #error("nope")
      |  ^~~~~
In file included from /usr/include/llvm-c-18/llvm-c/Target.h:47:
/usr/include/llvm-18/llvm/Config/Targets.def:1:2: error: #error ("nope")
    1 | #error("nope")
      |  ^~~~~
In file included from /usr/include/llvm-c-18/llvm-c/Target.h:52:
/usr/include/llvm-18/llvm/Config/Targets.def:1:2: error: #error ("nope")
    1 | #error("nope")
      |  ^~~~~
In file included from /usr/include/llvm-c-18/llvm-c/Target.h:78:
/usr/include/llvm-18/llvm/Config/Targets.def:1:2: error: #error ("nope")
    1 | #error("nope")
      |  ^~~~~
In file included from /usr/include/llvm-c-18/llvm-c/Target.h:87:
/usr/include/llvm-18/llvm/Config/Targets.def:1:2: error: #error ("nope")
    1 | #error("nope")
      |  ^~~~~
In file included from /usr/include/llvm-c-18/llvm-c/Target.h:96:
/usr/include/llvm-18/llvm/Config/Targets.def:1:2: error: #error ("nope")
    1 | #error("nope")
      |  ^~~~~
[...]
```

IIUC: To fix this properly, the cgo flags in `go-llvm` must be generated dynamically, which currently only works with [pkg-config](https://pkg.go.dev/cmd/cgo). Unfortunately, [llvm does not support pkg-config](https://github.com/llvm/llvm-project/issues/9777) and has its own tool for this, `llvm-config --cflags` and `llvm-config --ldflags`, and I don't see a way to integrate it automatically. Apart from using `go generate`, but ["(that) is intended for use by the author of the go package, not its clients."](https://go.dev/blog/generate)

So I'm out of ideas. :/

deadprogram:
This PR at least gets the CI back to passing while the more complete solution is considered.

Now merging, thank you @aykevl and @sylv-io

PR #4520, opened 2024-10-12
Various improved error checks

aykevl:
This is a bunch of small error checking and reporting changes.

Most importantly, it will now check whether TinyGo has been compiled with a too-old Go toolchain version. This should make issues like the following much clearer: https://github.com/NixOS/nixpkgs/pull/341170#issuecomment-2359237471
It may be overly aggressive though: compiling a package that specifies a newer Go version than the one TinyGo was compiled with may still work if there weren't any language changes. But in that case, the real issue is probably packaging: TinyGo should have been compiled with the newest Go version that it supports.

### DIFF
--- a/builder/config.go
+++ b/builder/config.go
@@ -2,6 +2,7 @@ package builder
 
 import (
 	"fmt"
+	"runtime"
 
 	"github.com/tinygo-org/tinygo/compileopts"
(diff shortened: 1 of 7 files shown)


deadprogram:

aykevl:
Rebased!

dgryski (approved):
LGTM

PR #4517, opened 2024-10-11
os: net.newUnixFile

leongross:
Building the u-root cmdlet `sluinit` with my experimental `net` PR #4498 following error occurs:

```
/usr/local/go/src/net/fd_unix.go:205: linker could not find symbol net.newUnixFile
``` 

This PR resolves the issue of the missing function `net_newUnixFile` and links it into the net package.

Since this package implements functionality for the linux net package, this code can currently not really be tested (see #4498).

PR #4521, opened 2024-10-14
Add sponsor button to key repositories

conejoninja:
Add the sponsor button linking to open collective in the repository.

Needs to be enable ( https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/displaying-a-sponsor-button-in-your-repository )

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,3 @@
+# These are supported funding model platforms
+
+open_collective: tinygo


leongross (commented):
Interestingly the CI for linux fails, although you did not touch this. I have the same issue in https://github.com/tinygo-org/tinygo/pull/4517, so `dev` seems to be broken? Same for `nix-test`.

deadprogram:

conejoninja:
done, but now Linux/assert-test-linux fails :/

leongross (commented):
```
/usr/bin/ld: ASTContext.cpp:(.text._ZL17DecodeTypeFromStrRPKcRKN5clang10ASTContextERNS3_19GetBuiltinTypeErrorERbb+0x670): undefined reference to `__isoc23_strtoul'
```
It looks to me as if there is still a version mismatch. The linker is unable to find the symbol `__isoc23_strtoul` which I think is defined in [glibc](https://github.com/bminor/glibc/blame/2c1903cbbac0022153a67776f474c221250ad6ed/stdlib/stdlib.h#L252), so it looks like the glibc version is too old.

aykevl:
This PR doesn't change any code, so I'd suggest to just ignore the CI errors.
The CI errors are weird though, if they keep happening I think we should do a new llvm-build (by incrementing the cache key) to see whether that helps. It looks like a version mismatch indeed.

leongross (approved):
I agree @aykevl. Since we had some trouble in the past few days with the CI, I initially thought we should investigate this while we are debugging the CI. But on other PRs like https://github.com/tinygo-org/tinygo/pull/4517 this could not be observed, so let's get this merged.

deadprogram:
Thank you @conejoninja now merging.

PR #4522, opened 2024-10-14
runtime: disallow defer in interrupts

aykevl:
This often doesn't work because there might not be a current task to push the defer frame to. It will instead show an unhelpful nil pointer dereference panic.

We could make this work with a separate defer stack for interrupts (as if they were newly started goroutines) but that is difficult with multiple interrupts happening at the same time (we shouldn't jump to a previous interrupt in `panic()`!). So instead, disable defer altogether in interrupts and adjust panic/recover accordingly.

### DIFF
--- a/src/runtime/panic.go
+++ b/src/runtime/panic.go
@@ -2,6 +2,7 @@ package runtime
 
 import (
 	"internal/task"
+	"runtime/interrupt"
 	"unsafe"
 )
 
@@ -50,7 +51,9 @@ func _panic(message interface{}) {
 	if panicStrategy() == panicStrategyTrap {
 		trap()
 	}
-	if supportsRecover() {
+	// Note: recover is not supported inside interrupts.
+	// (This could be supported, like defer, but we currently don't).
+	if supportsRecover() && !interrupt.In() {
 		frame := (*deferFrame)(task.Current().DeferFrame)
 		if frame != nil {
 			frame.PanicValue = message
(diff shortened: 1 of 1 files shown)


dgryski (approved):
LGTM

dgryski:
Do we also want to tweak the compiler so that defers get printed as allocations when `-print-allocs` is on?

aykevl:
> Do we also want to tweak the compiler so that defers get printed as allocations when `-print-allocs` is on?

Those allocations are regular allocations, they should already get printed with `-print-allocs=.`. (Note that many simple defers don't allocate, it's mainly the ones in loops that require a heap allocation).

deadprogram:

aykevl:
Rebased!

deadprogram:
Thanks for the improvement @aykevl and to @dgryski for reviewing. Now merging.

glococo:
Hi @aykevl , I refactorized tinygo somehow in this way:
machine.Serial.Enable().Receive( rx_callback )
however when an interrupt is fired, Arm jump to "handler mode", execute the interrupt, and when finish I would like to leave the handler mode and run the "callback" function in thread mode.

Why ? because runing the callback funcion in thread mode frees "interrupt active" register and in case of race conditions, interrupt can be called again while the callback is still running/paused and could keep filling a buffer.

I tried using defer (with previous tinygo version) but jump from iniside the handler mode, not freeing the interrupt active.
Any help is appreciated.

aykevl:
@glococo the best way to do this right now is using a non-blocking send on a channel. Something like this:

```go
var ch = make(chan struct{}, 1)

// inside the interrupt handler
select {
case ch <- struct{}{}:
default:
}

// in a separate goroutine
for range ch {
    // handle event
}
```

As long as every operation is non-blocking, you can use channels inside interrupts.

PR #4523, opened 2024-10-14
transform: optimize range over []byte(string)

eliasnaur:


eliasnaur:
The test failures don't *seem* relevant, but I could easily be mistaken.

deadprogram:

dgryski:
LGTM but I really want @aykevl's take on this.

transform/util.go
```diff
@@ -38,15 +38,17 @@ func hasFlag(call, param llvm.Value, kind string) bool {
...
 			if !isReadOnly(use) {
 				return false
 			}
-		} else if !use.IsACallInst().IsNil() {
+		case !use.IsACallInst().IsNil():
 			if !hasFlag(use, value, "readonly") {
 				return false
 			}
-		} else {
+		case !use.IsALoadInst().IsNil():
```
dgryski, line 50: Maybe add a comment to this block so it doesn't get lost next to the `default` label?

aykevl (approved):
Yes this looks reasonable (and I agree a comment before `default:` would look slightly better).

transform/util.go
eliasnaur: Done.

dgryski (approved):
LGTM

PR #4529, opened 2024-10-17
internal: update wasm-tools-go to v0.3.0

ydnar:
This updates the version of github.com/bytecodealliance/wasm-tools-go to v0.3.0 and regenerates the WASI 0.2 bindings in packages `internal/wasi/*` and `internal/cm`. This is a cherry-pick from #4501. cc @deadprogram


deadprogram (approved):
I am using this same version of `wasm-tools-go` elsewhere quite successfully so far.

deadprogram:
Now merging, thanks for update @ydnar

src/internal/cm/abi.go
```diff
@@ -84,6 +89,25 @@ func F64ToU64(v float64) uint64 { return *(*uint64)(unsafe.Pointer(&v)) }
...
+// [Canonical ABI]: https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md
+func F32ToU64(v float32) uint64 { return uint64(*(*uint32)(unsafe.Pointer(&v))) }
+
+// U64ToF32 maps the bits of a [uint64] into a [float32].
+// Used to lift a Core WebAssembly i64 into a [float32] when required by the [Canonical ABI].
+//
+// [uint64]: https://pkg.go.dev/builtin#uint64
+// [float32]: https://pkg.go.dev/builtin#float32
+// [Canonical ABI]: https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md
+func U64ToF32(v uint64) float32 {
```
dgryski, line 106: This comment should probably apply to the upstream generator, but note that `math` already has
```
[func Float32bits(f float32) uint32](https://pkg.go.dev/math#Float32bits)
[func Float32frombits(b uint32) float32](https://pkg.go.dev/math#Float32frombits)
[func Float64bits(f float64) uint64](https://pkg.go.dev/math#Float64bits)
[func Float64frombits(b uint64) float64](https://pkg.go.dev/math#Float64frombits)
```

src/internal/wasi/io/v0.2.0/streams/streams.wit.go
```diff
@@ -54,6 +64,16 @@ func (self *StreamError) Closed() bool {
...
 }
 
+var stringsStreamError = [2]string{
+	"last-operation-failed",
+	"closed",
+}
+
+// String implements [fmt.Stringer], returning the variant case name of v.
+func (v StreamError) String() string {
+	return stringsStreamError[v.Tag()]
```
dgryski, line 74: Again a comment for the generator here: should we be handling "unknown" types (`Tag() >= 2`) or just assume that will "never happen" and a crash regardless?

dgryski (approved):
LGTM to merge with the exception of some comments about the generated code (but those shouldn't block merging this..)

src/internal/cm/abi.go
ydnar: Memorializing IRL conversation this past week in San Francisco:

"We duplicate functionality from package `math` here to avoid a potential circular dependency in package `runtime`. Package `cm` intentionally depends only on `unsafe`."

src/internal/wasi/io/v0.2.0/streams/streams.wit.go
ydnar: Memorializing IRL conversation this past week in San Francisco:

"Invalid variant cases are a bug, and a panic/crash is the intended behavior."

PR #4532, opened 2024-10-18
Remove `minSched` hack for wasm

aykevl:
I am not entirely sure what it's doing (it seems related to `js.FuncOf`), but tests still seem to pass when this code is removed. So let's remove it.

Added test to make sure `js.FuncOf` is at least working to some degree.

See the discussion in #4494 for context.


deadprogram:
Good to remove this, I think. Thanks @aykevl now merging.

PR #4533, opened 2024-10-18
ci: use macos-13 instead of macos-12 for amd64 builds

aykevl:
The macos-12 runner is being deprecated, so we have to switch to a new runner: https://github.com/actions/runner-images/issues/10721

The next one is macos-13, which is still amd64.


deadprogram:
Thanks for the update @aykevl now merging.

PR #4534, opened 2024-10-18
compiler: mark stringFromBytes as nocapture/readonly to help escape a…

### LINKED ISSUE #4525: compiler: strconv.AppendUint allocates
```
$ cat byteslicegarbage_test.go
package byteslicegarbage

import (
	"strconv"
	"testing"
)

func TestAllocs(t *testing.T) {
	dst := make([]byte, 50)
	res := testing.Benchmark(func(b *testing.B) {
		for range b.N {
			strconv.AppendUint(dst[:0], 10000, 10)
		}
	})
	if string(dst)[:2] != "10" {
		t.Errorf("AppendUint: %q", string(dst))
	}
	if a := res.AllocsPerOp(); a != 0 {
		t.Errorf("AppendUint allocates %d, expected none", a)
	}
}
$ go test
$ tinygo test
--- FAIL: TestAllocs (2.42s)
    AppendUint allocates 1, expected none
FAIL
FAIL	blah	2.625s
```

This allocation is caused by the folowing pattern (from `strconv.formatBits`):

```
func formatBits(dst []byte, ..., append bool) ([]byte, string) {
    var buf [65]byte
    ...
    // Fill buf.
    ....
    if append {
        return append(dst), ""
    }
    return nil, string(buf)   
}
```

As you can see, the conversion to string is conditional, but TinyGo nevertheless decides to allocate buf on the heap. This saves a copy in the non-append case, but incurs an allocation in the performance-critical append case.

From a cursory glance at `compiler/compiler.go`, `ssa.Convert` operations are translated to `runtime.stringFromBytes`, so some other pass must transform that to the escaping buffer. Perhaps just inlining and general optimization?

dgryski:
…nalysis

Fixes #4525

### DIFF
--- a/compiler/symbol.go
+++ b/compiler/symbol.go
@@ -175,6 +175,9 @@ func (c *compilerContext) getFunction(fn *ssa.Function) (llvm.Type, llvm.Value) 
 	case "runtime.stringFromBytes":
 		llvmFn.AddAttributeAtIndex(1, c.ctx.CreateEnumAttribute(llvm.AttributeKindID("nocapture"), 0))
 		llvmFn.AddAttributeAtIndex(1, c.ctx.CreateEnumAttribute(llvm.AttributeKindID("readonly"), 0))
+	case "runtime.stringFromRunes":
+		llvmFn.AddAttributeAtIndex(1, c.ctx.CreateEnumAttribute(llvm.AttributeKindID("nocapture"), 0))
+		llvmFn.AddAttributeAtIndex(1, c.ctx.CreateEnumAttribute(llvm.AttributeKindID("readonly"), 0))
 	case "runtime.trackPointer":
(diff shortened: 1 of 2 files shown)


dgryski:
I have some time at airports today; I might look at some of the other `strconv` escape analysis issues and see if there's anything else there with an easy fix.

aykevl:
Would it be possible to add a test to transform/testdata/allocs2.go?

dgryski:
Added. 

New test fails on `dev` with
```
        allocs2.go:53: object allocated on the heap: escapes at line 54
```
and passes on this branch.

deadprogram:
OK this is great, I quite need this actually. Thank you very much @dgryski and to @aykevl for review.

Now merging.

PR #4535, opened 2024-10-18
use unsafe.Slice in memhash routines

dgryski:


ydnar (approved):

aykevl:
Great! This looks so much better.

PR #4536, opened 2024-10-19
goenv: parse patch version, add func Compare to compare two Go version strings

ydnar:
This PR implements 3 changes:

- `goenv.Parse` now parses patch version out of a Go version string.
- Adds `goenv.Compare` to compare two Go version strings. This is used in #4501 to enable `structs.HostLayout` enforcement.
- Adds tests for `Parse` and `Compare`.

Extracted from #4501.

### DIFF
--- a/builder/config.go
+++ b/builder/config.go
@@ -43,7 +43,7 @@ func NewConfig(options *compileopts.Options) (*compileopts.Config, error) {
 	// compiled with the latest Go version.
 	// This may be a bit too aggressive: if the newer version doesn't change the
 	// Go language we will most likely be able to compile it.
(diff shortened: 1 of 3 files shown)


aykevl (approved):
LGTM.

Was this intended to be kept as separate commits or can this be squashed when merging?

deadprogram:
I think we can squash it, and keep this moving along.

Thank you very much @ydnar for the fixes/improvements and to @aykevl for review. Now merging

ydnar:
> LGTM.
> 
> 
> 
> Was this intended to be kept as separate commits or can this be squashed when merging?

Squashed! 

Thanks for the review.

PR #4537, opened 2024-10-21
esp32c3: add smoke tests for a few boards

aykevl:
These boards probably haven't been working since the addition of I2C, because they were missing some constants in the machine package.


deadprogram:
Thank you for the additional test coverage @aykevl now merging.

PR #4538, opened 2024-10-21
targets: add WaveShare ESP-C3-32S-Kit

aykevl:
I've had this board for a while now, but never added proper TinyGo support. So here is a PR to do just that.


deadprogram:
@aykevl perhaps you forgot a file or something?
https://github.com/tinygo-org/tinygo/actions/runs/11438159651/job/31831984697?pr=4538#step:19:632

aykevl:
Oops I used the wrong name for the target JSON file. Should be fixed now.

deadprogram:

PR #4539, opened 2024-10-21
fe310: support GPIO PinInput

aykevl:
This is needed to support switching between input and output.


deadprogram:

PR #4542, opened 2024-10-22
wasm: use precise GC for WebAssembly (including WASI)

aykevl:
With a few small modifications, all the problems with `-gc=precise` in WebAssembly seem to have been fixed.

I didn't do any performance measurements, but this is supposed to improve GC performance.

One issue was that it was allocating a `[]uintptr` for the stack, even though the stack can contain pointers. Another issue was a bad check for stack overflows that triggered when it should not.

### DIFF
--- a/src/internal/task/task.go
+++ b/src/internal/task/task.go
@@ -30,3 +30,6 @@ type Task struct {
 // the given function and falls back to the default stack size. It is replaced
 // with a load from a special section just before codegen.
 func getGoroutineStackSize(fn uintptr) uintptr
+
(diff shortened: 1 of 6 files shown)


ydnar:
This generally seems OK. How does the precise GC behave with things that look like pointers but are not?

dgryski:
@ydnar If there are no type bits, then it falls back on the conservative logic.

dgryski (approved):
LGTM

deadprogram:
Thanks for the improvement @aykevl and to @ydnar and @dgryski fior review. Now merging.

PR #4543, opened 2024-10-24
reflect: add `Value.Clear`; support `anytype->interface{}`, `Slice->(*)Array` in `Value.Convert`, fix `Copy`

### LINKED ISSUE #4554: `reflect.Copy` Incorrect When Source is Array of Len > 8
Tested from `dev`. Minimal reproduction:

```go
//go:build ignore

package main

import (
	"fmt"
	"reflect"
)

func main() {
	guid := [9]byte{0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09}
	b := make([]byte, 9)

	// Copy builtin works
	copy(b, guid[:])
	fmt.Printf("%x\n", b)

	// Reflect with slices works
	reflect.Copy(reflect.ValueOf(b), reflect.ValueOf(guid[:]))
	fmt.Printf("%x\n", b)

	// Reflect with array doesn't work
	reflect.Copy(reflect.ValueOf(b), reflect.ValueOf(guid))
	fmt.Printf("%x\n", b)
}
```

Result:

```console
$ tinygo run ./test.go
010203040506070809
010203040506070809
303ac73ca07f000002

$ go run ./test.go
010203040506070809
010203040506070809
010203040506070809
```

Not shown in the reproduction: a `[8]byte` array works fine, so the issue seems to only be for lengths > 8.

ben-krieger:
Fixes #4554

This PR adds a few things to the reflect package.

1. Add `Value.Clear`
2. Convert anything to `interface{}`, i.e. `reflect.ValueOf(42).Convert(reflect.TypeFor[interface{}]())`
3. Convert slices to arrays, i.e. `reflect.ValueOf([]int{1, 2, 3}).Convert(reflect.TypeFor[[3]int]())`
4. Convert slices to array pointers... because after doing 3, I felt like I had to
5. Fixes `reflect.Copy` when the source is an array

The motivation was to allow compiling the CBOR library at https://github.com/fido-device-onboard/go-fdo/tree/tinygo-server-support/cbor. However, this PR can be split into smaller ones if desired.

### DIFF
--- a/src/reflect/type.go
+++ b/src/reflect/type.go
@@ -424,8 +424,8 @@ type rawType struct {
 	meta uint8 // metadata byte, contains kind and flags (see constants above)
 }
 
-// All types that have an element type: named, chan, slice, array, map (but not
-// pointer because it doesn't have ptrTo).
+// All types that have an element type: named, chan, slice, array, map, interface
+// (but not pointer because it doesn't have ptrTo).
 type elemType struct {
 	rawType
 	numMethod uint16
@@ -437,6 +437,12 @@ type ptrType struct {
 	rawType
 	numMethod uint16
 	elem      *rawType
+}
+
+type interfaceType struct {
+	rawType
+	ptrTo *rawType
+	// TODO: methods
 }
 
 type arrayType struct {
@@ -995,6 +1001,10 @@ func (t *rawType) AssignableTo(u Type) bool {
 		return true
 	}
 
+	if t.underlying() == u.(*rawType).underlying() && (!t.isNamed() || !u.(*rawType).isNamed()) {
+		return true
+	}
+
 	if u.Kind() == Interface && u.NumMethod() == 0 {
 		return true
 	}
@@ -1060,6 +1070,9 @@ func (t *rawType) NumMethod() int {
 		return int((*ptrType)(unsafe.Pointer(t)).numMethod)
 	case Struct:
 		return int((*structType)(unsafe.Pointer(t)).numMethod)
+	case Interface:
+		//FIXME: Use len(methods)
+		return (*interfaceType)(unsafe.Pointer(t)).ptrTo.NumMethod()
 	}
 
 	// Other types have no methods attached.  Note we don't panic here.
--- a/src/reflect/value.go
+++ b/src/reflect/value.go
@@ -689,6 +689,25 @@ func (v Value) Cap() int {
 	}
 }
 
+//go:linkname mapclear runtime.hashmapClear
+func mapclear(p unsafe.Pointer)
+
+// Clear clears the contents of a map or zeros the contents of a slice
+//
+// It panics if v's Kind is not Map or Slice.
+func (v Value) Clear() {
+	switch v.typecode.Kind() {
+	case Map:
+		mapclear(v.pointer())
+	case Slice:
+		hdr := (*sliceHeader)(v.value)
+		elemSize := v.typecode.underlying().elem().Size()
+		memzero(hdr.data, elemSize*hdr.len)
+	default:
+		panic(&ValueError{Method: "Clear", Kind: v.Kind()})
+	}
+}
+
 // NumField returns the number of fields of this struct. It panics for other
 // value types.
 func (v Value) NumField() int {
@@ -888,6 +907,9 @@ func (v Value) Index(i int) Value {
 }
 
 func (v Value) NumMethod() int {
+	if v.typecode == nil {
+		panic(&ValueError{Method: "reflect.Value.NumMethod", Kind: Invalid})
+	}
 	return v.typecode.NumMethod()
 }
 
@@ -1062,7 +1084,7 @@ func (v Value) Set(x Value) {
 	v.checkAddressable()
 	v.checkRO()
 	if !x.typecode.AssignableTo(v.typecode) {
-		panic("reflect: cannot set")
+		panic("reflect.Value.Set: value of type " + x.typecode.String() + " cannot be assigned to type " + v.typecode.String())
 	}
 
 	if v.typecode.Kind() == Interface && x.typecode.Kind() != Interface {
@@ -1240,7 +1262,9 @@ func (v Value) OverflowUint(x uint64) bool {
 }
 
 func (v Value) CanConvert(t Type) bool {
-	panic("unimplemented: (reflect.Value).CanConvert()")
+	// TODO: Optimize this to not actually perform a conversion
+	_, ok := convertOp(v, t)
+	return ok
 }
 
 func (v Value) Convert(t Type) Value {
@@ -1259,6 +1283,15 @@ func convertOp(src Value, typ Type) (Value, bool) {
 			typecode: typ.(*rawType),
 			value:    src.value,
 			flags:    src.flags,
+		}, true
+	}
+
+	if rtype := typ.(*rawType); rtype.Kind() == Interface && rtype.NumMethod() == 0 {
+		iface := composeInterface(unsafe.Pointer(src.typecode), src.value)
+		return Value{
+			typecode: rtype,
+			value:    unsafe.Pointer(&iface),
+			flags:    valueFlagExported,
 		}, true
 	}
 
@@ -1302,14 +1335,33 @@ func convertOp(src Value, typ Type) (Value, bool) {
 		*/
 
 	case Slice:
-		if typ.Kind() == String && !src.typecode.elem().isNamed() {
-			rtype := typ.(*rawType)
-
-			switch src.Type().Elem().Kind() {
-			case Uint8:
-				return cvtBytesString(src, rtype), true
-			case Int32:
-				return cvtRunesString(src, rtype), true
+		switch rtype := typ.(*rawType); rtype.Kind() {
+		case Array:
+			if src.typecode.elem() == rtype.elem() && rtype.Len() <= src.Len() {
+				return Value{
+					typecode: rtype,
+					value:    (*sliceHeader)(src.value).data,
+					flags:    src.flags | valueFlagIndirect,
+				}, true
+			}
+		case Pointer:
+			if rtype.Elem().Kind() == Array {
+				if src.typecode.elem() == rtype.elem().elem() && rtype.elem().Len() <= src.Len() {
+					return Value{
+						typecode: rtype,
+						value:    (*sliceHeader)(src.value).data,
+						flags:    src.flags & (valueFlagExported | valueFlagRO),
+					}, true
+				}
+			}
+		case String:
+			if !src.typecode.elem().isNamed() {
+				switch src.Type().Elem().Kind() {
+				case Uint8:
+					return cvtBytesString(src, rtype), true
+				case Int32:
+					return cvtRunesString(src, rtype), true
+				}
 			}
 		}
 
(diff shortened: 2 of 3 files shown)


src/reflect/value.go
```diff
@@ -689,6 +689,27 @@ func (v Value) Cap() int {
...
+//
+// It panics if v's Kind is not Map or Slice.
+func (v Value) Clear() {
+	switch v.typecode.Kind() {
+	case Map:
+		mapclear(v.pointer())
+	case Slice:
+		hdr := (*sliceHeader)(v.value)
+		elemSize := v.typecode.underlying().elem().Size()
+		for i := uintptr(0); i < hdr.len; i++ {
```
dgryski: Shouldn't this just be a single `memzero` call instead of one call for each element in the slice?

dgryski (commented):
Just noticed an issue, but I'll do a more complete review Thursday.

src/reflect/value.go
ben-krieger: Updated.

src/reflect/type.go
```diff
@@ -970,10 +976,18 @@ func (t *rawType) FieldAlign() int {
 // AssignableTo returns whether a value of type t can be assigned to a variable
 // of type u.
 func (t *rawType) AssignableTo(u Type) bool {
-	if t == u.(*rawType) {
+	if u := u.(*rawType); t == u || t.underlying() == u || t == u.underlying() {
```
dgryski: `underlying()` returns the type itself if it's not a named type; so you should just be able to use `t.underlying() == u.underlying()` (or use `t == u` first if you want to potentially avoid the function calls).

src/reflect/type.go
```diff
@@ -970,10 +976,18 @@ func (t *rawType) FieldAlign() int {
...
 		return true
 	}
 
+	if u.Kind() == Array && t.Kind() == Slice {
```
dgryski: These additions make our `reflect` package differ from upstream:
```
~/go/src/github.com/dgryski/refcvt $ tinygo run main.go
true
true
~/go/src/github.com/dgryski/refcvt $ go run main.go
false
false
~/go/src/github.com/dgryski/refcvt $ cat main.go
package main

import "reflect"

func main() {

	s := make([]byte, 10)
	var hw [5]byte

	st := reflect.TypeOf(s)
	hwt := reflect.TypeOf(hw)
	hwtp := reflect.TypeOf(&hw)

	println(st.AssignableTo(hwt))
	println(st.AssignableTo(hwtp))
}
```

src/reflect/value.go
```diff
@@ -1262,6 +1286,15 @@ func convertOp(src Value, typ Type) (Value, bool) {
 		}, true
 	}
 
+	if rtype := typ.(*rawType); rtype.Kind() == Interface && rtype.NumMethod() == 0 {
```
dgryski, line 1289: Do you have a test that triggers this?

src/reflect/value.go
```diff
@@ -1302,25 +1335,47 @@ func convertOp(src Value, typ Type) (Value, bool) {
...
-			case Uint8:
-				return cvtBytesString(src, rtype), true
-			case Int32:
-				return cvtRunesString(src, rtype), true
+		switch rtype := typ.(*rawType); rtype.Kind() {
+		case Array:
+			if src.typecode.elem() == rtype.elem() && rtype.Len() <= src.Len() {
+				return Value{
+					typecode: rtype,
+					value:    (*sliceHeader)(src.value).data,
```
dgryski, line 1343: Pretty sure you need to add the `valueFlagIndirect` here to the set of flags, otherwise we'll miss the extra pointer load when we need to get the value.

```
~/go/src/github.com/dgryski/refcvt $ go run main.go
706050403020100 c000102020
~/go/src/github.com/dgryski/refcvt $ tinygo run main.go
10b2514b0 10b2514b0
~/go/src/github.com/dgryski/refcvt $ cat main.go
package main

import (
	"encoding/binary"
	"fmt"
	"reflect"
	"unsafe"
)

func main() {

	s := make([]byte, 8)
	var arr [8]byte

	for i := range s {
		s[i] = byte(i)
	}

	arrT := reflect.TypeOf(arr)
	sV := reflect.ValueOf(s)

	a := sV.Convert(arrT).Interface().([8]byte)

	ptr := binary.LittleEndian.Uint64(a[:])
	fmt.Printf("%x %x\n", ptr, unsafe.SliceData(s))
}
```

src/reflect/value_test.go
```diff
@@ -624,6 +648,123 @@ func TestConvert(t *testing.T) {
...
+	_ = v.Interface().(interface{}).([]struct{})
+
+	v = ValueOf(map[string]string{"A": "B"}).Convert(anyType)
+	if v.Kind() != Interface || v.NumMethod() > 0 {
+		t.Error("Convert(map -> interface{})")
+	}
+	_ = v.Interface().(interface{}).(map[string]string)
+}
+
+func TestClearSlice(t *testing.T) {
```
dgryski, line 743: Also need `TestClearMap`

src/reflect/value_test.go
```diff
@@ -1,6 +1,7 @@
 package reflect_test
```
dgryski, line 1: Are there any tests in `all_test.go` you can uncomment?

dgryski (changes requested):

src/reflect/type.go
ben-krieger: That's actually subtly different and I'm not sure what the correct behavior is.

Currently, I'd say that these two types below are not assignable to one another, though either is assignable to `[]byte` or vice versa.

```go
type Named1 []byte
type Named2 []byte
```

I'm going to look at what's correct in the language spec.

src/reflect/type.go
ben-krieger: Hmmm... So the correct change would have been to update `Value.Set` to check `AssignableTo || CanConvert`?

src/reflect/value.go
ben-krieger: `TestConvertToEmptyInterface`

src/reflect/value.go
ben-krieger: Thank you. I'm fumbling in the dark with the flags, because it's hard to come up with tests to see if they're set correctly. The behavior of when a value is addressable, etc is quite confusing.

src/reflect/value_test.go
ben-krieger: I did not know to look there. I'll check!

ben-krieger:
Thanks for the thorough review @dgryski! I'll spend some time this weekend fixing it up and adding more tests.

src/reflect/type.go
ben-krieger: Updated and now your example returns false false for both.

src/reflect/value.go
ben-krieger: Updated. Here's my test program:

```go
//go:build ignore

package main

import (
	"fmt"
	"reflect"
)

func main() {
	s := make([]byte, 8)
	for i := range s {
		s[i] = byte(i)
	}

	a := reflect.ValueOf(s).Convert(reflect.TypeFor[[8]byte]()).Interface().([8]byte)
	b := reflect.ValueOf(s).Convert(reflect.TypeFor[*[8]byte]()).Interface().(*[8]byte)

	fmt.Printf("&a: %p\n&b: %p\n&s: %p\n", a[:], b[:], s)
	fmt.Printf("a: %v\nb: %v\ns: %v\n", a[:], (*b)[:], s)
}
```

Results:

```console
$ go run ./test.go
&a: 0xc00009a048
&b: 0xc00009a040
&s: 0xc00009a040
a: [0 1 2 3 4 5 6 7]
b: [0 1 2 3 4 5 6 7]
s: [0 1 2 3 4 5 6 7]

$ tinygo run ./test.go
&a: 0x7f07aca106d0
&b: 0x7f07aca106b0
&s: 0x7f07aca106b0
a: [0 1 2 3 4 5 6 7]
b: [0 1 2 3 4 5 6 7]
s: [0 1 2 3 4 5 6 7]
```

src/reflect/type.go
ben-krieger: Yes, your suggestion is the correct behavior. Updated.

src/reflect/value_test.go
ben-krieger: I don't think any can be uncommented without significant additions. The most obvious candidate is the conversion tests, but this would require implementing `ConvertibleTo`.

dgryski:
`testdata/reflect.go` is still failing.  That's Bad because the desired output is generated by upstream Go, so that indicates incompatibilities.

ben-krieger:
> `testdata/reflect.go` is still failing. That's Bad because the desired output is generated by upstream Go, so that indicates incompatibilities.

Got it. So it seems that arrays in some cases have become addrable and settable, when they shouldn't. Let me see if I can figure out why...

dgryski:
It's almost certainly the flag changes for `indirect` support.  You probably need some `RO` bits too.

ben-krieger:
> It's almost certainly the flag changes for `indirect` support. You probably need some `RO` bits too.

Tried RO, but couldn't get it working, so I reverted the change that broke it. (I was also uncomfortable with that code, because I didn't understand the underlying mechanism that distinguished between arrays > 64 bits in size.) However, the test I added for `reflect.Copy` is still failing.

Any thoughts on the proper way to fix #4554?

ben-krieger:
> It's almost certainly the flag changes for `indirect` support. You probably need some `RO` bits too.

@dgryski I think this is ready for (hopefully the last) review. All tests are passing now and the fix for #4554 matches a lot of other code within the reflect package that check either indirect OR size > uintptr.

I don't see any more low-hanging fruit for test coverage, but let me know what you think. Thanks!

src/reflect/value.go
```diff
@@ -1061,8 +1083,8 @@ func (it *MapIter) Next() bool {
 func (v Value) Set(x Value) {
 	v.checkAddressable()
 	v.checkRO()
-	if !x.typecode.AssignableTo(v.typecode) {
-		panic("reflect: cannot set")
+	if !x.typecode.AssignableTo(v.typecode) && !x.typecode.ConvertibleTo(v.typecode) {
```
dgryski: Should this be `||` instead of `&&` ?

dgryski:
One quick comment.  I'll do a full review tomorrow (Monday) and then hopefully we can get this merged.  Thanks for all your work on this!

src/reflect/value.go
ben-krieger: According to the go docs:

> As in Go, x's value must be assignable to v's type and must not be derived from an unexported field.

So this change shouldn't be needed... However, I was not getting correct behavior matching Go without it. 

I'll see if I can add some tests and update AssignableTo with whatever implicit conversions are allowed, instead. 

But yes, the && was correct because it was not A and not B, meaning not (A or B), i.e fail if not assignable or convertible.

dgryski: Thanks.  I think I must've missed one of the `!`'s when I skimmed this.

src/reflect/type.go
```diff
@@ -995,6 +1001,10 @@ func (t *rawType) AssignableTo(u Type) bool {
 		return true
 	}
 
+	if t.underlying() == u.(*rawType).underlying() && (!t.isNamed() || !u.(*rawType).isNamed()) {
```
ben-krieger, line 1004: See https://go.dev/ref/spec#Assignability

If the underlying types are the same, only one can be named.

dgryski (approved):
LGTM

dgryski:
@aykevl Any comments?

aykevl:
Took a look, looks good to me.

PR #4544, opened 2024-10-25
machine/usb/adc/midi: Fix PitchBend()

sago35:
In the first commit, I clarify the operator precedence.
In practice, the behavior does not change.
However, `(channel-1 & 0xf)` is not the intended code.

https://go.dev/play/p/wZxeYl0yr0-

The second commit fixes the issue with PitchBend.
I believe a correction was necessary due to an error in the shift amount.

| m.msg[2] | m.msg[3] | pitchbend value |
|-----------|-----------|-----------------|
| 0x7F | 0x00 | 0b0000_0000_0111_1111 |
| 0x00 | 0x7F | 0b0011_1111_1000_0000 |

### DIFF
--- a/src/machine/usb/adc/midi/messages.go
+++ b/src/machine/usb/adc/midi/messages.go
@@ -177,7 +177,7 @@ func (m *midi) PitchBend(cable, channel uint8, bend uint16) error {
 		return errInvalidMIDIPitchBend
 	}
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Thank you for the corrections @sago35 now merging. :musical_keyboard:

PR #4548, opened 2024-10-25
main: parse extldflags early so we can report the error message

aykevl:
This avoids some weird behavior when the -extldflags flag cannot be parsed by TinyGo.


dgryski (approved):
LGTM

PR #4549, opened 2024-10-25
runtime: move scheduler code around

aykevl:
This moves all scheduler code into a separate file that is only compiled when there's a scheduler in use (the tasks or asyncify scheduler, which are both cooperative). The main goal of this change is to make it easier to add a new "scheduler" based on OS threads.

It also fixes a few subtle issues with `-gc=none`:

  - `runtime.Gosched()` panicked. This is now fixed to just return immediately (the only logical thing to do when there's only one goroutine).
  - Timers aren't supported without a scheduler, but the relevant code was still present and would happily add a timer to the queue. It just never ran. So now it exits with a runtime error, similar to any blocking operation.

Other than that, this really just moves code around.


aykevl:
Note: this PR conflicts with some changes I'd like to make to the sync package to make it thread-safe, so waiting until this one is merged before I submit those other PRs (see #4559).

aykevl:
Rebased to resolve merge conflicts.

src/runtime/arch_tinygoriscv.go
```diff
@@ -36,7 +36,8 @@ func procUnpin() {
 
 func waitForEvents() {
 	mask := riscv.DisableInterrupts()
-	if !runqueue.Empty() {
+	runqueue := schedulerRunQueue()
+	if !(runqueue != nil && runqueue.Empty()) {
```
eliasnaur: Nit: `runqueue == nil || !runqueue.Empty()` is clearer to me, but I understand why you write it like this.

Nittier: `if rq := schedulerRunQueue(); rq == nil || !rq.Empty() {`

src/runtime/scheduler_none.go
```diff
@@ -11,18 +23,41 @@ func sleep(duration int64) {
...
+
+func removeTimer(tim *timer) bool {
+	runtimePanic("timers not supported without a scheduler")
+	return false
+}
+
+func schedulerRunQueue() *task.Queue {
+	// This function is not actually used, it is only called when hasScheduler
+	// is true. So we can just return nil here.
+	return nil
```
eliasnaur: Nit: `panic("unreachable")` like `scheduler` below?

src/runtime/scheduler_none.go
```diff
@@ -11,18 +23,41 @@ func sleep(duration int64) {
...
+}
+
+func scheduler(returnAtDeadlock bool) {
+	// This *should* be unreachable.
```
eliasnaur: Nit: unhelpful comment.

eliasnaur (commented):
Drive-by review.

deadprogram:
@aykevl this PR needs to be rebased against `dev` to resolve merge conflicts. Also, what do you think about @eliasnaur comments?

aykevl:
Rebased, and fixed the review issues pointed out by @eliasnaur (thanks for the review!).

aykevl:
Fixed some build failures.

deadprogram:
Thank you for working on the improvements @aykevl and thank you very much @eliasnaur for review/feedback!

Now merging.

PR #4552, opened 2024-10-26
ci: use TinyGo version in artifact files

aykevl:
This avoids needing to rename them ourselves (which is kinda annoying) and also avoids mistakes in the process.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -19,6 +19,8 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: golang:1.23-alpine
+    outputs:
+      version: ${{ steps.version.outputs.version }}
     steps:
       - name: Install apk dependencies
         # tar: needed for actions/cache@v4
@@ -32,6 +34,9 @@ jobs:
         uses: actions/checkout@v4
         with:
           submodules: true
+      - name: Extract TinyGo version
+        id: version
+        run: ./.github/workflows/tinygo-extract-version.sh | tee -a "$GITHUB_OUTPUT"
       - name: Cache Go
         uses: actions/cache@v4
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/tinygo-extract-version.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+# Extract the version string from the source code, to be stored in a variable.
+grep 'const version' goenv/version.go | sed 's/^const version = "\(.*\)"$/version=\1/g'
```
dgryski, line 4: grep not needed
```
sed -n '/^const version =/{ s/^const version = "\(.*\)"$/version=\1/; p;}' goenv/version.go
```

.github/workflows/linux.yml
```diff
@@ -152,11 +157,11 @@ jobs:
       - name: Download release artifact
         uses: actions/download-artifact@v4
         with:
-          name: linux-amd64-double-zipped
+          name: linux-amd64-double-zipped-${{ needs.build-linux.outputs.version }}
       - name: Extract release tarball
         run: |
           mkdir -p ~/lib
-          tar -C ~/lib -xf tinygo.linux-amd64.tar.gz
+          tar -C ~/lib -xf tinygo*.linux-amd64.tar.gz
```
dgryski: Any reason we're globbing here instead of using the `...outputs.version` variable?

.github/workflows/linux.yml
```diff
@@ -297,6 +302,9 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+      - name: Get TinyGo version
+        id: version
+        run: ./.github/workflows/tinygo-extract-version.sh >> "$GITHUB_OUTPUT"
```
dgryski: Above you used `tee -a` instead of `>>`.  Why the change?

dgryski (approved):
LGTM.

.github/workflows/tinygo-extract-version.sh
aykevl: True, though I think the `grep` and `sed` together are more readable? This way the `sed` is just a simple regex search-and-replace instead of the more complicated code you wrote that I don't quite understand.
I can update if you think avoiding `grep` is better.

.github/workflows/linux.yml
aykevl: Fixed. This was the older code. I updated the code in most places (from `>>` to `tee -a` to be able to see the output of the step), but apparently missed this one.

.github/workflows/linux.yml
aykevl: No, it was just easier at the time. Fixed.

.github/workflows/tinygo-extract-version.sh
dgryski: Nah, you can leave the grep if you prefer it.

Here's a breakdown of the `sed` code.

Sed, like awk, is can do `/pattern/ { action }`, so that's what we have here.  The `-n` flag tells `sed` not to print every line (the default), so  in our action after the replacement we need to explicitly `p` to print the current line.  If this was happening on a huge file I'd probably add a `q` to quit so `sed` didn't continue to process the entire file after it found the one match it was looking for

dgryski (approved):
LGTM

dgryski:
Conflict that needs to be fixed before we can merge this.

aykevl:
Fixed the conflict!

PR #4553, opened 2024-10-26
ci: remove 'shell: bash' lines from MacOS build

aykevl:
Unlike Windows, we can just use the default shell here.


dgryski (approved):
LGTM.  CI passes so this must be right :D

PR #4555, opened 2024-10-27
tinygo: revise and simplify wasmtime argument handling

ydnar:
This PR implements 3 related changes to how TinyGo processes arguments for the `wasmtime` emulator. I ran across the issue related to trying to use `wasmtime serve` in a custom target file for my WasmCon talk next month.

It adds support for alternative `wasmtime` subcommands like `serve`, simplifies the Wasmtime-specific code paths, and fixes PWD handling for `tinygo run` when not running tests. I recommend reviewing by commit to see the individual changes.

Detailed changes:

- Simplify `buildAndRun`, condensing `wasmtime` handling into a single section handling `--env` and `--dir` args.
- Use `wasmtime run` subcommand instead of bare `wasmtime`. This enables use of `wasmtime serve` or other subcommands. The key change is that emulator arguments are inserted *after* the subcommand, e.g. `run`.
- Set `PWD` to the package dir *only when running tests*, which allows `tinygo run --target wasip1` to run with `PWD=.`.
- Fixed a related bug with `-x` (`PrintCommands`) where it repeated the full path and name of the command.

### DIFF
--- a/main.go
+++ b/main.go
@@ -925,7 +925,7 @@ func buildAndRun(pkgName string, config *compileopts.Config, stdout io.Writer, c
 
 	// Run binary.
 	if config.Options.PrintCommands != nil {
-		config.Options.PrintCommands(cmd.Path, cmd.Args...)
+		config.Options.PrintCommands(cmd.Path, cmd.Args[1:]...)
 	}
 	err = run(cmd, result)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -925,7 +927,7 @@ func buildAndRun(pkgName string, config *compileopts.Config, stdout io.Writer, c
 
 	// Run binary.
 	if config.Options.PrintCommands != nil {
-		config.Options.PrintCommands(cmd.Path, cmd.Args...)
+		config.Options.PrintCommands(cmd.Path, cmd.Args[1:]...)
```
ydnar, line 930: Running `tinygo run -x ...` repeated the full path to `wasmtime` and `wasmtime`.

deadprogram (approved):

dgryski:
The `isSingleFile` logic was for cases where we did `tinygo run some/dir/file.go` to make sure the directories were set correctly.  If that still works then these cleanups LGTM.

ydnar:
Yes that still works. 

The equivalent `tinygo run ./cmd/foo` also works now (without specifying `main.go`).

dgryski (approved):
LGTM

PR #4556, opened 2024-10-27
runtime: rewrite channel implementation

aykevl:
This rewrite simplifies the channel implementation considerably, with 34% fewer LOC. Perhaps the most important change is the removal of the channel state, which made sense when we had only send and receive operations but only makes things more compliated when multiple select operations can be pending on a single channel.

I did this rewrite originally to make it possible to make channels parallelism-safe (see https://github.com/tinygo-org/tinygo/issues/2446). The current implementation is not parallelism-safe, but it will be easy to make it so (the main additions will be a channel lock, a global select lock, and an atomic compare-and-swap in chanQueue.pop).

This change also reduces binary size: for the driver examples that do change (presumably the ones that use a channel), the code size is reduced by around 400 bytes.

### DIFF
--- a/src/runtime/chan.go
+++ b/src/runtime/chan.go
@@ -1,27 +1,45 @@
 package runtime
 
 // This file implements the 'chan' type and send/receive/select operations.
-
-// A channel can be in one of the following states:
-//     empty:
-//       No goroutine is waiting on a send or receive operation. The 'blocked'
-//       member is nil.
-//     recv:
-//       A goroutine tries to receive from the channel. This goroutine is stored
-//       in the 'blocked' member.
-//     send:
-//       The reverse of send. A goroutine tries to send to the channel. This
-//       goroutine is stored in the 'blocked' member.
-//     closed:
-//       The channel is closed. Sends will panic, receives will get a zero value
-//       plus optionally the indication that the channel is zero (with the
-//       comma-ok value in the task).
 //
-// A send/recv transmission is completed by copying from the data element of the
-// sending task to the data element of the receiving task, and setting
-// the 'comma-ok' value to true.
-// A receive operation on a closed channel is completed by zeroing the data
-// element of the receiving task and setting the 'comma-ok' value to false.
+// Every channel has a list of senders and a list of receivers, and possibly a
+// queue. There is no 'channel state', the state is inferred from the available
+// senders/receivers and values in the buffer.
+//
+// - A sender will first try to send the value to a waiting receiver if there is
+//   one, but only if there is nothing in the queue (to keep the values flowing
+//   in the correct order). If it can't, it will add the value in the queue and
+//   possibly wait as a sender if there's no space available.
+// - A receiver will first try to read a value from the queue, but if there is
+//   none it will try to read from a sender in the list. It will block if it
+//   can't proceed.
+//
+// State is kept in various ways:
+//
+// - The sender value is stored in the sender 'channelOp', which is really a
+//   queue entry. This works for both senders and select operations: a select
+//   operation has a separate value to send for each case.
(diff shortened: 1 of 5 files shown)


src/runtime/chan.go
```diff
@@ -1,520 +1,329 @@
...
-	// If this is a send operation, then the send buffer can be found in this select state.
-	s *chanSelectState
-
-	// allSelectOps is a slice containing all of the channel operations involved with this select statement.
-	// Before resuming the task, all other channel operations on this select statement should be canceled by removing them from their corresponding lists.
-	allSelectOps []channelBlockedList
+const (
+	chanOperationWaiting = iota // waiting for a send/receive operation to continue
+	chanOperationOk             // successfully sent or received (not closed)
+	chanOperationClosed         // channel was closed, the value has been zeroed
```
eliasnaur: Nit: because of `chanOperationMask`, these would be clearer (to me) spelled out in binary literal form.

src/runtime/chan.go
```diff
@@ -1,520 +1,329 @@
...
-		case &b.allSelectOps[i] == b:
-			// This entry is the one that was already detached.
-			continue
-		case v.t == nil:
-			// This entry is not used (nil channel).
-			continue
+// Pop the next waiting channel from the queue. Channels that are no longer
+// waiting (for example, when they're part of a select operation) will be
+// skipped.
+// This function must be called with interrupts disabled.
```
eliasnaur, line 85: Is 'interrupts disabled' still accurate when our lord and saviour, multithreading support, arrives? Here and below.

src/runtime/chan.go
```diff
@@ -1,520 +1,329 @@
...
-	ch.state = chanStateRecv
-	receiver.Ptr, receiver.Data = value, 1
-	*blockedlist = channelBlockedList{
-		next: ch.blocked,
-		t:    receiver,
-	}
-	ch.blocked = blockedlist
-	chanDebug(ch)
-	interrupt.Restore(i)
+	// We can't proceed, so we add ourselfs to the list of receivers and wait
```
eliasnaur: Nit: "ourselves".

eliasnaur (commented):
Drive-by review.

deadprogram:
@aykevl any responses to @eliasnaur comments? :smile_cat:

deadprogram:
Also note that I changed the branch for the PR just now to `dev`.

src/runtime/chan.go
aykevl: Seems reasonable, thanks!

src/runtime/chan.go
aykevl: Yes, though the comment will need to be updated to say something like "this function must be called with the channel lock held and interrupts disabled".  Interrupts still need to be disabled on microcontrollers. What will happen exactly on a chip like the rp2040 (that has both interrupts and two cores) is something I'll figure out when that happens.

src/runtime/chan.go
aykevl: Thanks, fixed!

aykevl:
Rebased and fixed the issues pointed out by @eliasnaur.

Note that this PR and #4549 do conflict with each other (even though that won't be visible in git/GitHub). Once one is merged, the other will need to be rebased and updated.

deadprogram:
This PR can now be rebased @aykevl if you please.

aykevl:
Rebased after merging #4549!

deadprogram (approved):
Good to reduce complexity, thanks for working on this!

deadprogram:
Now merging, thank you to @aykevl for working on this and to @eliasnaur for review.

PR #4558, opened 2024-10-28
sync: don't use `volatile` in Mutex

aykevl:
Volatile loads/stors are only useful for communication with interrupts or for memory-mapped I/O. They do not provide any sort of safety for sync.Mutex, while making it *appear* as if it is more safe.

  * `sync.Mutex` cannot be used safely inside interrupts, because any blocking calls (including `Lock`) will cause a runtime panic.
  * For multithreading, `volatile` is also the wrong choice. Atomic operations should be used instead, and the current code would not work for multithreaded programs anyway.

---

Looking at the original PR in https://github.com/tinygo-org/tinygo/pull/3983 I don't see why volatile is needed here. It makes it look like it's doing something safe, while it is not (the default without `volatile` is already safe for the current scheduler because there's no threading).
I'm making this PR in advance because it will make it easier to implement a `sync.Mutex` that does in fact support multithreading (using atomics and futex).

The only thing that this would provide, is to make the following unconventional pattern somewhat safe:

```go
var mutex sync.Mutex

func startTransfer() {
    mutex.Lock()
    startSomeInterruptHandler()

    // wait until the interrupt unlocks the mutex
    mutex.Lock()
}

func handleInterrupt() {
    // do something with the interrupt

    // signal to the main goroutine that we're done
    mutex.Unlock()
}
```

I'm not aware of anybody doing this, and I don't think it's documented anywhere. And in any case, we'd be better off providing a proper primitive for this feature of waiting for an interrupt to finish (a TODO for me for years already).

### DIFF
--- a/src/sync/mutex.go
+++ b/src/sync/mutex.go
@@ -3,12 +3,10 @@ package sync
 import (
 	"internal/task"
 	_ "unsafe"
-
-	"runtime/volatile"
 )
 
 type Mutex struct {
-	state   uint8 // Set to non-zero if locked.
+	locked  bool
 	blocked task.Stack
 }
 
@@ -16,18 +14,18 @@ type Mutex struct {
 func scheduleTask(*task.Task)
 
 func (m *Mutex) Lock() {
-	if m.islocked() {
+	if m.locked {
 		// Push self onto stack of blocked tasks, and wait to be resumed.
 		m.blocked.Push(task.Current())
 		task.Pause()
 		return
 	}
 
-	m.setlock(true)
+	m.locked = true
 }
 
 func (m *Mutex) Unlock() {
-	if !m.islocked() {
+	if !m.locked {
 		panic("sync: unlock of unlocked Mutex")
 	}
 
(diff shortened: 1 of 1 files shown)


dgryski:
My guess is I had probably just been looking at a tinygo runtime but where llvm was doing something unexpected and so decided to take a more defensive route.

dgryski (approved):
LGTM

aykevl:
> My guess is I had probably just been looking at a tinygo runtime but where llvm was doing something unexpected and so decided to take a more defensive route.

There are most certainly valid use cases for volatile! I just don't see it in this case - the code is basically entirely synchronous (stack switching is possible, but that also happens at defined points in the program).

soypat (approved):
I honestly don't recall why I stuck volatile in there 🫠. Looks and sounds good to me!

aykevl:
I will in fact be adding atomic instructions in the future, see #4559 (commit name "sync: implement futex-based Mutex"). That's how I found the volatile instructions and was wondering how they got there - hence this PR.

PR #4385, opened 2024-08-05
Boehm GC support

aykevl:
This adds support for the well-known Boehm GC. It's significantly faster than our own naive GC and could be used as an alternative on bigger systems.

In the future, this GC might also be supported on WebAssembly with some extra work. Right now it's Linux only (though Windows/MacOS shouldn't be too difficult to add).

---

~Still a draft because it depends on #4101 and I might improve the PR a little bit more.~

PR #4559, opened 2024-10-28
Threading support

aykevl:
Work in progress. Many necessary parts are not yet implemented, but this gives context for PRs I want to make in the future.

See https://github.com/tinygo-org/tinygo/issues/2446 for motivation (in particular, why this focuses on Linux first) and for a list of things that need to be implemented to make multithreading work in TinyGo.

### DIFF
--- a/src/internal/task/task_threads.c
+++ b/src/internal/task/task_threads.c
@@ -1,0 +1,104 @@
+//go:build none
+
+#define _GNU_SOURCE
+#include <pthread.h>
+#include <semaphore.h>
+#include <signal.h>
+#include <stdint.h>
+#include <stdio.h>
+
+// BDWGC also uses SIGRTMIN+6 on Linux, which seems like a reasonable choice.
+#ifdef __linux__
+#define taskPauseSignal (SIGRTMIN + 6)
+#endif
+
+// Pointer to the current task.Task structure.
+// Ideally the entire task.Task structure would be a thread-local variable but
+// this also works.
+static __thread void *current_task;
+
+struct state_pass {
+    void      *(*start)(void*);
+    void      *args;
+    void      *task;
+    uintptr_t *stackTop;
+    sem_t     startlock;
+};
+
+// Handle the GC pause in Go.
+void tinygo_task_gc_pause(int sig);
+
+// Initialize the main thread.
+void tinygo_task_init(void *mainTask, pthread_t *thread, void *context) {
+    // Make sure the current task pointer is set correctly for the main
+    // goroutine as well.
+    current_task = mainTask;
+
+    // Store the thread ID of the main thread.
+    *thread = pthread_self();
+
+    // Register the "GC pause" signal for the entire process.
+    // Using pthread_kill, we can still send the signal to a specific thread.
+    struct sigaction act = { 0 };
+    act.sa_flags = SA_SIGINFO;
+    act.sa_handler = &tinygo_task_gc_pause;
+    sigaction(taskPauseSignal, &act, NULL);
+}
+
+void tinygo_task_exited(void*);
+
+// Helper to start a goroutine while also storing the 'task' structure.
+static void* start_wrapper(void *arg) {
+    struct state_pass *state = arg;
+    void *(*start)(void*) = state->start;
+    void *args = state->args;
+    current_task = state->task;
+
+    // Save the current stack pointer in the goroutine state, for the GC.
+    int stackAddr;
+    *(state->stackTop) = (uintptr_t)(&stackAddr);
+
+    // Notify the caller that the thread has successfully started and
+    // initialized.
+    sem_post(&state->startlock);
+
+    // Run the goroutine function.
+    start(args);
+
+    // Notify the Go side this thread will exit.
+    tinygo_task_exited(current_task);
+
+    return NULL;
+};
+
+// Start a new goroutine in an OS thread.
+int tinygo_task_start(uintptr_t fn, void *args, void *task, pthread_t *thread, uintptr_t *stackTop, void *context) {
+    // Sanity check. Should get optimized away.
+    if (sizeof(pthread_t) != sizeof(void*)) {
+        __builtin_trap();
+    }
+
+    struct state_pass state = {
+        .start     = (void*)fn,
+        .args      = args,
+        .task      = task,
+        .stackTop  = stackTop,
+    };
+    sem_init(&state.startlock, 0, 0);
+    int result = pthread_create(thread, NULL, &start_wrapper, &state);
+
+    // Wait until the thread has been created and read all state_pass variables.
+    sem_wait(&state.startlock);
+
+    return result;
+}
+
+// Return the current task (for task.Current()).
+void* tinygo_task_current(void) {
+    return current_task;
+}
+
+// Send a signal to cause the task to pause for the GC mark phase.
+void tinygo_task_send_gc_signal(pthread_t thread) {
+    pthread_kill(thread, taskPauseSignal);
+}
--- a/src/internal/task/task_threads.go
+++ b/src/internal/task/task_threads.go
@@ -1,0 +1,275 @@
+//go:build scheduler.threads
+
+package task
+
+import (
+	"sync/atomic"
+	"unsafe"
+)
+
+// If true, print verbose debug logs.
+const verbose = false
+
+// Scheduler-specific state.
+type state struct {
+	// Goroutine ID. The number here is not really significant and after a while
+	// it could wrap around. But it is useful for debugging.
+	id uintptr
+
+	// Thread ID, pthread_t or similar (typically implemented as a pointer).
+	thread threadID
+
+	// Highest address of the stack. It is stored when the goroutine starts, and
+	// is needed to be able to scan the stack.
+	stackTop uintptr
+
+	// Next task in the activeTasks queue.
+	QueueNext *Task
+
+	// Semaphore to pause/resume the thread atomically.
+	pauseSem Semaphore
+
+	// Semaphore used for stack scanning.
+	// We can't reuse pauseSem here since the thread might have been paused for
+	// other reasons (for example, because it was waiting on a channel).
+	gcSem Semaphore
+}
+
+// Goroutine counter, starting at 0 for the main goroutine.
+var goroutineID uintptr
+
+var mainTask Task
+
+// Queue of tasks (see QueueNext) that currently exist in the program.
+var activeTasks = &mainTask
+var activeTaskLock PMutex
+
+func OnSystemStack() bool {
+	runtimePanic("todo: task.OnSystemStack")
+	return false
+}
+
+// Initialize the main goroutine state. Must be called by the runtime on
+// startup, before starting any other goroutines.
+func Init(sp uintptr) {
+	mainTask.state.stackTop = sp
+	tinygo_task_init(&mainTask, &mainTask.state.thread)
+}
+
+// Return the task struct for the current thread.
+func Current() *Task {
+	t := (*Task)(tinygo_task_current())
+	if t == nil {
+		runtimePanic("unknown current task")
+	}
+	return t
+}
+
+// Pause pauses the current task, until it is resumed by another task.
+// It is possible that another task has called Resume() on the task before it
+// hits Pause(), in which case the task won't be paused but continues
+// immediately.
+func Pause() {
+	// Wait until resumed
+	t := Current()
+	if verbose {
+		println("*** pause:  ", t.state.id)
+	}
+	t.state.pauseSem.Wait()
+}
+
+// Resume the given task.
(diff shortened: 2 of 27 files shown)


aykevl:
Updated after some refactoring (moved internal/llsync into internal/task) and added GC support (stop-the-world scanning by pausing all threads and scanning their stacks).
Main issues remaining are:

 - [x] Some problem with `bits/fcntl.h` that I need to investigate (see CI failures).
 - [x] Timer support.
 - [x] Fixing tests that assume a deterministic order in which goroutines are run.
 - [x] Fixing signal support.

src/internal/task/futex_linux.c
```diff
@@ -0,0 +1,28 @@
...
+// import.
+
+#include <stdint.h>
+#include <sys/syscall.h>
+#include <time.h>
+#include <unistd.h>
+
+#define FUTEX_WAIT    0
+#define FUTEX_WAKE    1
+#define FUTEX_PRIVATE 128
```
leongross: Why redefine these here if they can be included from `<linux/futex.h>`?

src/internal/task/futex_linux.c
```diff
@@ -0,0 +1,28 @@
+//go:build none
+
+// This file is manually included, to avoid CGo which would cause a circular
+// import.
+
+#include <stdint.h>
+#include <sys/syscall.h>
+#include <time.h>
+#include <unistd.h>
+
```
leongross:
```suggestion
#include <unistd.h>
#include <linux/futex.h>
```

src/internal/task/futex_linux.c
leongross: nit: also afaik the flag itself is called [`FUTEX_PRIVATE_FLAG`](https://elixir.bootlin.com/linux/v6.11/source/include/uapi/linux/futex.h#L26), just to be consistent with [`FUTEX_WAIT`](https://elixir.bootlin.com/linux/v6.11/source/include/uapi/linux/futex.h#L11) and `FUTEX_WAKE`

aykevl: Musl (which we use) doesn't provide `linux/futex.h`.

aykevl: Thanks! Fixed locally, will be included the next time I push the branch.

aykevl:
Whoa, looks like all tests are passing!

jclab-joseph:
@aykevl Awesome! Thanks for your hard work

aykevl:
I don't think there's a whole lot more I can extract and make separate PRs for, so let's mark this as ready for review!

src/internal/task/task_threads.c
```diff
@@ -0,0 +1,104 @@
...
+    struct state_pass state = {
+        .start     = (void*)fn,
+        .args      = args,
+        .task      = task,
+        .stackTop  = stackTop,
+    };
+    sem_init(&state.startlock, 0, 0);
+    int result = pthread_create(thread, NULL, &start_wrapper, &state);
+
+    // Wait until the thread has been crated and read all state_pass variables.
```
eliasnaur: Nit: "created"

src/internal/task/task_threads.go
```diff
@@ -0,0 +1,265 @@
...
+	// and the stop-the-world GC won't see threads that haven't started yet or
+	// are not fully started yet.
+	activeTaskLock.Lock()
+	errCode := tinygo_task_start(fn, args, t, &t.state.thread, &t.state.stackTop)
+	if errCode != 0 {
+		runtimePanic("could not start thread")
+	}
+	t.state.QueueNext = activeTasks
+	activeTasks = t
+	activeTaskLock.Unlock()
```
eliasnaur, line 116: Nit: defer the unlock?

src/internal/task/task_threads.go
```diff
@@ -0,0 +1,265 @@
...
+	// Wake each paused thread for the second time, so they will resume normal
+	// operation.
+	for t := activeTasks; t != nil; t = t.state.QueueNext {
+		if t != current {
+			t.state.gcSem.Post()
+		}
+	}
+
+	// Allow goroutines to start and exit again.
+	activeTaskLock.Unlock()
```
eliasnaur, line 212: Nit: defer?

src/internal/task/task_threads.go
```diff
@@ -0,0 +1,265 @@
...
+		runtimePanic("taskExited failed")
+	}
+}
+
+// Futex to wait on until all tasks have finished scanning the stack.
+// This is basically a sync.WaitGroup.
+var scanDoneFutex Futex
+
+// GC scan phase. Because we need to stop the world while scanning, this kinda
+// needs to be done in the tasks package.
```
eliasnaur, line 149: FWIW, I find this structure backwards. I would expect the package `runtime` to have `GCScan` that then calls into this package `task`. Especially in light of `runtime.gcScanGlobals` and `runtime.scanStack`.

src/runtime/gc_blocks.go
```diff
@@ -444,7 +452,9 @@ func free(ptr unsafe.Pointer) {
 
 // GC performs a garbage collection cycle.
 func GC() {
+	gcLock.Lock()
 	runGC()
+	gcLock.Unlock()
```
eliasnaur, line 458: Nit: defer?

src/runtime/gc_blocks.go
```diff
@@ -733,6 +743,7 @@ func ReadMemStats(m *MemStats) {
 	m.Sys = uint64(heapEnd - heapStart)
 	m.HeapAlloc = (gcTotalBlocks - gcFreedBlocks) * uint64(bytesPerBlock)
 	m.Alloc = m.HeapAlloc
+	gcLock.Unlock()
```
eliasnaur, line 757: Nit: defer?

eliasnaur (commented):
Very much look forward to this work!

src/runtime/scheduler.go
```diff
@@ -27,6 +29,34 @@ func scheduleLogChan(msg string, ch *channel, t *task.Task) {
 	}
 }
 
+func timerQueueAdd(tim *timerNode) {
```
leongross: nit: I think naming the timerNode `tim` is confusing. We should name it either `timer`, `tn`, or just `t`.
I see that name throughout the changes, is it an abbreviation or common term that I am not familiar with?

src/internal/task/task_threads.go
```diff
@@ -0,0 +1,265 @@
...
+
+	// Highest address of the stack. It is stored when the goroutine starts, and
+	// is needed to be able to scan the stack.
+	stackTop uintptr
+
+	// Next task in the activeTasks queue.
+	QueueNext *Task
+
+	// Semaphore to pause/resume the thread atomically.
+	pauseSem Semaphore
```
b0ch3nski, line 30: Shouldn't `Semaphore` be used with pointer receiver?

src/sync/mutex.go
```diff
@@ -6,131 +6,96 @@ import (
...
+	// Writer lock. Held between Lock() and Unlock().
+	writerLock Mutex
 }
 
-const (
-	rwMutexStateUnlocked = uint32(0)
-	rwMutexStateWLocked  = ^uint32(0)
-	rwMutexMaxReaders    = rwMutexStateWLocked - 1
-)
+const rwMutexMaxReaders = 1 << 30
```
leongross, line 32: Is there a reason you chose this specific value?

src/internal/task/task_threads.c
aykevl: Thanks, fixed!

src/internal/task/task_threads.go
aykevl: I avoid `defer` in places like these since:

1. `defer` can cause heap allocations in some cases (mainly in loops, so not really relevant here, but it's not very visible).
2. Using `defer` currently makes every call in the function slower and larger, since it needs to be able to recover.
3. We don't need the run-defer-on-panic behavior here: if we panic, something is very wrong and it doesn't help to unlock.

Admittedly it _could_ be used here, but in many other places (e.g. `GCScan`) it can lead to unexpected side effects if it allocates.

So in short, `defer` as it is currently implemented is inefficient and can have unexpected side effects so I avoid its use in the runtime.

(Same goes for all the other comments regarding `defer`).

src/internal/task/task_threads.go
aykevl: I agree with you, and I first tried to put the equivalent of `GCScan` in the runtime. But `GCScan` also needs access to various private symbols like `activeTaskLock`, `activeTasks`, `scanDoneFutex`, etc so to avoid exporting many private details of the tasks package I put it there.

I can however make `scanCurrentStack` and `gcScanGlobals` callbacks, so they don't directly call back to the runtime. But then I don't know what to do with `scanCurrentStack` in `tinygo_task_gc_pause`. In the end, the GC and the scheduler do need some tight integration, I can't see a way around that to be honest.

src/internal/task/task_threads.go
aykevl: No. The `state` struct (or actually the `Task` struct where it is embedded) will always be accessed via a pointer, like `*Task`. So essentially when calling a function like `pauseSem.Post()`, what happens is something like `SemaphorePost(&t.state.pauseSem)`.

Also, I think you mean to say that `pauseSem` should have the type `*Semaphore`? Since the methods on the `Semaphore` type do in fact have a pointer receiver:

```go
func (s *Semaphore) Post() { ... }
func (s *Semaphore) Wait() { ... }
```

src/sync/mutex.go
aykevl: No reason, except that it's the same value as used in the [upstream `RWMutex` implementation](https://github.com/golang/go/blob/master/src/sync/rwmutex.go#L47):

```go
const rwmutexMaxReaders = 1 << 30
```

src/runtime/scheduler.go
aykevl: Good point, I've updated the naming of some variables here to be more consistent.
(I think `tim` is just short for `timer`).

aykevl:
Updated the PR after some review comments. And rebased on dev to incorporate #4435.

src/internal/task/task_threads.go
b0ch3nski: My point was about a pointer (pun intended :smile:) on `*Semaphore`. While `Semaphore` methods do have pointer receiver, the instances of `pauseSem` and `gcSem` members would get copied when `state` struct instance is ever passed around.

I've created playground snippet which uses stdlib `sync.Mutex` as an example to even trigger `go vet` warnings: https://go.dev/play/p/cV7h16_JS1R

aykevl: > While `Semaphore` methods do have pointer receiver, the instances of `pauseSem` and `gcSem` members would get copied when `state` struct instance is ever passed around.

That's why it should never be copied around!
Because it's an internal API, we can make sure this is the case. (And it is).

src/sync/mutex.go
leongross: Sounds reasonable enough to me.

aykevl:
Rebased (which was a bit of work after #4385 got merged), let's see whether CI is happy with this.

aykevl:
Added some comments to clarify stuff! Ready for review!

(Waiting for this, since lots of people want multicore support on the RP2040 and this PR is a prerequisite).

src/runtime/gc_blocks.go
```diff
@@ -492,6 +501,10 @@ func runGC() (freeBytes uintptr) {
 		finishMark()
 	}
 
+	// If we're using threads, resume all other threads before starting the
+	// sweep.
+	gcResumeWold()
```
eliasnaur: nit: typo: `gcResumeWorld`

eliasnaur (commented):
LGTM, but I'm certainly no expert in this area.

src/runtime/gc_blocks.go
aykevl: Oops, fixed!
That's the danger of autocomplete (it makes it very easy to repeat a single typo).

aykevl:
I'd love to get this merged, since it blocks a few other things I'm working on (including the much-demanded RP2040 multicore support).

b0ch3nski:
LGTM. I've read the code and there is nothing in there I could be worried about, although I understand mostly the general idea of this.

deadprogram:
Thanks for all of the useful comments in this PR @aykevl it has made it a lot easier to review. Amazing work!

Thank you very much to @leongross @jclab-joseph @eliasnaur @b0ch3nski for all of the thoughtful comments and feedback!

Now merging :rocket:

jfcg:
Amazing !
Are goroutines mapped to OS-threads one-to-one ?
Or is there a multiplexer like in gc toolchain?
Great work, thanks..

aykevl:
Goroutines are mapped one-to-one to OS threads for ease of implementation. It really simplifies the implementation a lot.

PR #4560, opened 2024-10-28
Prepare leaking GC for multithreading

aykevl:
This makes the leaking GC (`-gc=leaking`) ready for multithreading. The leaking GC is the easiest real heap allocator to add support for, since it can increment the heap pointer with a simple atomic add operation.
This PR does not actually make the leaking GC support multithreading, but it makes it very easy to do so in the future simply by changing the new llsync package (see https://github.com/tinygo-org/tinygo/pull/4559).

This is the first PR of (hopefully) a bunch to follow that will gradually add support for multithreading to various parts of TinyGo. I split it in many pieces to make it hopefully easier to review.

This PR is intended to be merged by rebasing.

### DIFF
--- a/src/runtime/gc_leaking.go
+++ b/src/runtime/gc_leaking.go
@@ -7,6 +7,7 @@ package runtime
 // may be the only memory allocator possible.
 
 import (
+	"internal/task"
 	"unsafe"
 )
 
@@ -19,6 +20,9 @@ var gcTotalAlloc uint64
 // Total number of calls to alloc()
 var gcMallocs uint64
 
+// Heap lock for parallel goroutines. No-op when single threaded.
+var gcLock task.PMutex
+
 // Total number of objected freed; for leaking collector this stays 0
 const gcFrees = 0
 
@@ -30,6 +34,7 @@ func alloc(size uintptr, layout unsafe.Pointer) unsafe.Pointer {
 	// TODO: this can be optimized by not casting between pointers and ints so
 	// much. And by using platform-native data types (e.g. *uint8 for 8-bit
 	// systems).
+	gcLock.Lock()
 	size = align(size)
 	addr := heapptr
 	gcTotalAlloc += uint64(size)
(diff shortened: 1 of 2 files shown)


aykevl:
Note: not super convinced about the "llsync" name (meaning "low level sync"). Others I have considered: "sync" (can be confused by the stdlib sync package), "atomic" (same), "futex" (there's more in there than just futexes), "mutex" (it's more than mutexes), "internalsync" (too long), etc.
Please let me know if you can think of a better name!

dgryski:
`tinysync` ?

aykevl:
Actually I realized I'd like to be able to use these synchronization primitives inside the internal/task package and some synchronization primitives (such as the pseudo-futex for the non-threaded scheduler) need internal/task. So that's going to be a circular dependency. Therefore I've moved the new stuff into the internal/task package as well, to simplify things.
Also, putting them there does make some sense: these primitives depend on which kind of scheduler (and therefore which kind of task) is in use.

This nicely sidesteps the naming question :)

aykevl:
Can someone take a look at this PR?

eliasnaur:
Not a maintainer, but LGTM. The changes are fairly harmless: adding the passthrough `pseudoAtomic` type and refactoring the leaking GC to use it.

dgryski:
I think there's more than just sprinkling a few atomics onto some integers to truly make this thread-safe.  For example, the growing logic and checking `heapEnd` needs to be similarly protected.   Generally to make things thread-safe without wrapping the whole thing in a mutex means a lot more atomic test-and-set loops.  I can try to take a look at this next week.

aykevl:
> For example, the growing logic and checking `heapEnd` needs to be similarly protected.

Oops, you are correct! I believe it's correct for a non-growing heap, but that's only the case on microcontrollers.

aykevl:
Updated the PR. With `growHeap`, there doesn't seem to be a trivial way to use atomic operations so I've used a lock instead.

src/runtime/gc_leaking.go
```diff
@@ -43,6 +48,8 @@ func alloc(size uintptr, layout unsafe.Pointer) unsafe.Pointer {
 		// Failed to make the heap bigger, so we must really be out of memory.
 		runtimePanic("out of memory")
```
dgryski, line 49: Unlock mutex before throwing `panic`?

aykevl: It's a runtime panic, which cannot be recovered from. So unlocking the mutex won't help anybody.

deadprogram:
Any other feedback on this PR @dgryski ?

dgryski (approved):
LGTM.

PR #4562, opened 2024-10-29
ci: run at least some tests on older Go/LLVM versions

aykevl:
These should make sure basic functionality is still working. Using the `-short` flag to avoid taking too long to run all tests (and to install all the necessary emulators), and because some targets might not work in older Go/LLVM versions (such as WASI).


aykevl:
Seems to be working!

deadprogram:
Thanks for looking after the older versions @aykevl now merging.

PR #4564, opened 2024-10-30
test: show output even when a test binary didn't exit cleanly

aykevl:
This was a problem on wasm, where `node` would exit with a non-zero exit code when there was a panic.


dgryski (approved):
LGTM

PR #4565, opened 2024-10-30
wasm: use wasip1 API for parameter/env passing

aykevl:
Instead of hardcoding the command line parameters and the environment variables in the binary, pass them at runtime to Node.js and use the WASIp1 API to retrieve them in wasm_exec.js.

The only real benefit right now is that it becomes possible to change `go.argv` and `go.env` before running a wasm binary.

This also changes the syscall package for GOOS=js: it now becomes more like a libc (using wasi-libc), which means error values like `syscall.EEXIST` will actually match the one returned by the relevant libc function.

Wasm binary size for packages that import the os package will be increased somewhat.

---

Now that I've written it, I realized I don't need it after all. So I'm not sure whether this change is a good idea: I haven't seen anybody request support for `go.argv` or `go.env` yet and it does increase binary size for wasm binaries in many cases. Leaving it as a draft for now, we can merge it if there's demand for it I guess. Or close it if not.

### DIFF
--- a/main.go
+++ b/main.go
@@ -770,12 +770,11 @@ func Run(pkgName string, options *compileopts.Options, cmdArgs []string) error {
 // for the given emulator.
 func buildAndRun(pkgName string, config *compileopts.Config, stdout io.Writer, cmdArgs, environmentVars []string, timeout time.Duration, run func(cmd *exec.Cmd, result builder.BuildResult) error) (builder.BuildResult, error) {
 	// Determine whether we're on a system that supports environment variables
-	// and command line parameters (operating systems, WASI) or not (baremetal,
-	// WebAssembly in the browser). If we're on a system without an environment,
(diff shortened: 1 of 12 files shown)


deadprogram:
@aykevl there are requests now for the features in this PR. Any chance you can bring it back to life?

aykevl:
> @aykevl there are requests now for the features in this PR. Any chance you can bring it back to life?

This request is a misunderstanding of what this PR does. We already have parameter passing support for wasip1. This PR just uses the same API for browser wasm, it doesn't (or shouldn't) affect WASI.

deadprogram:
I also misunderstood the PR lol. Thanks for the clarification.

dgryski:
Should we close this?

aykevl:
Closing for now.

Issue #4563, opened 2024-10-29
wasm code never returns to javascript after await go.run(instance)


glycerine:
filing an issue per Ayke's request from gopher slack:

Under tinygo 0.34.0 (and v0.33.0) my javascript can call the wasm but 
the wasm never returns back to the javascript after the await go.run(instance);
 like the big Go code does. 

I tried calling an explicit os.Exit(0) too, but on execution I get: `process.exit is not a function`.

Is there some trick to getting the tinygo wasm to finish? 

I tested the simplest package main  func main() { println("hiya") } program. 

(to repro: https://gist.github.com/glycerine/50faaa852a8e41792ea7133732ec1655)

glycerine:
https://github.com/tinygo-org/tinygo/blob/release/targets/wasm_exec.js#L473

seems to hang forever so maybe that is the culprit?

glycerine:
Yes, if I just comment out that line (line 473 targets/wasm_exec.js), things seem to work fine; the wasm code returns to the javascript.

(Not to say there isn't the work Aykevl mentioned on slack to do to make sure everything else (other goroutines?) is finished properly first, just suggesting that removing that wait on a promise that will never resolve will certainly have to be a part of any fix).

glycerine:
fixed by https://github.com/tinygo-org/tinygo/pull/4567

PR #4567, opened 2024-10-30
Fix two wasm bugs

aykevl:
Fix two related bugs:

 1. `time.Sleep` in a normal (non-reactor) browser wasm binary would cause every subsequent exported function (`//go:wasmexport`) to fail.
 2. https://github.com/tinygo-org/tinygo/issues/4563

### DIFF
--- a/testdata/wasmexport-noscheduler.go
+++ b/testdata/wasmexport-noscheduler.go
@@ -1,5 +1,7 @@
 package main
 
+import "time"
+
 func init() {
 	println("called init")
 }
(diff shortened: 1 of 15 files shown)


testdata/wasmexport-noscheduler.go
```diff
@@ -9,6 +11,7 @@ func callTestMain()
 
 func main() {
 	// main.main is not used when using -buildmode=c-shared.
+	time.Sleep(time.Millisecond)
```
eliasnaur, line 15: Leftover debug? Otherwise a comment would be nice.

aykevl: Fixed! It's actually part of the test, I've added a comment.

ydnar (approved):
LGTM

dgryski:
I think `TestTest/WASM` has started failing since this PR was merged.

aykevl:
Yes, that does seem likely. My bad!

PR #4570, opened 2024-10-31
runtime/trace: stub more functions

joonas:
I came across the following error while I was looking at using the opentelemetry libraries with TinyGo:
```console
# go.opentelemetry.io/otel/sdk/trace
../../../../pkg/mod/go.opentelemetry.io/otel/sdk@v1.31.0/trace/span.go:820:9: undefined: rt.IsEnabled
../../../../pkg/mod/go.opentelemetry.io/otel/sdk@v1.31.0/trace/span.go:824:19: undefined: rt.NewTask
```

I thought it would be nice to have these functions stubbed out so that it would be possible to get further along with libraries that might try to make use of these `runtime/trace` functions, if for no other reason than to be able to use the `go:linkname` directive to do something else with them.

### DIFF
--- a/src/runtime/trace/trace.go
+++ b/src/runtime/trace/trace.go
@@ -2,6 +2,7 @@
 package trace
 
 import (
+	"context"
 	"errors"
 	"io"
 )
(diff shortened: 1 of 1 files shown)


src/runtime/trace/trace.go
```diff
@@ -11,3 +12,29 @@ func Start(w io.Writer) error {
...
+
+type Task struct{}
+
+func (t *Task) End() {}
+
+func Log(ctx context.Context, category, message string) {}
+
+func Logf(ctx context.Context, category, format string, args ...any) {}
+
+func WithRegion(ctx context.Context, regionType string, fn func()) {}
```
aykevl: I don't think this is correct. The documentation says:

> WithRegion starts a region associated with its calling goroutine, runs fn, and then ends the region.

So even if we don't implement tracing, I think `WithRegion` should at least call `fn`.

aykevl (commented):
Can you run `go fmt` on src/runtime/trace/trace.go`? The CI complains that it is not correctly formatted (probably because of the missing newline at the end of the file).

Other than that and the comment below, LGTM.

src/runtime/trace/trace.go
joonas: Good call, I've adjusted the function to call the passed in `fn`, thanks!

joonas:
~Decided to force push to see if the Windows test fails consistently, [since it seems to have passed in prior push](https://github.com/tinygo-org/tinygo/actions/runs/11606320373/job/32318147605).~

Force push seems to have done the trick 👍

aykevl:
Thank you!

Issue #3731, opened 2023-05-17
`defer panic()` doesn't work


dgryski:
```
~/go/src/github.com/dgryski/bug/defpan $ tinygo run main.go
todo: builtin: panic
todo: builtin: panic
~/go/src/github.com/dgryski/bug/defpan $ cat main.go
package main

func main() {
	println("before")
	defer panic("panic")
	println("after")
}
```

`panic` is `*ssa.Panic` which isn't handled by `defer` (or `createBuiltin`?).  I guess we need to turn it into a `b.createRuntimeInvoke("_panic", ...)`.

joonas:
just wanted to chime in here that I'm also interested in this, but not quite sure how I could best help move things along 🙂

aykevl:
Here is a fix: https://github.com/tinygo-org/tinygo/pull/4571

PR #4571, opened 2024-10-31
compiler: allow deferred panic

aykevl:
This is rare, but apparently some programs do this:

    defer panic("...")

This is emitted in the IR as a builtin function.

See: https://github.com/tinygo-org/tinygo/issues/3731


dgryski (approved):
LGTM

joonas:
I just wanted to share that after applying these changes along with my changes from #4570 on to a local dev build of tinygo, the library I was trying to use compiles as expected, so I'm very excited to see these land!

Issue #4527, opened 2024-10-17
nRF52: a single interrupt activation doesn't work


perttierkkila:
TLDR: nRF52 blinky example doesn't work when serial is disabled.

Env
> tinygo version 0.33.0 darwin/arm64 (using go version go1.22.5 and LLVM version 18.1.2)

This is a pretty wild bug which goes down to interrupt enable and this one line https://github.com/tinygo-org/tinygo/blob/release/src/runtime/runtime_nrf.go#L61
> intr.Enable()

When this line is replaced code from Enable() with constant value, IRQ works properly. Only difference is internal irq.num is not used. Enabling any other IRQ doesn't do the trick.
> arm.EnableIRQ(uint32(nrf.IRQ_RTC1))

And the weird part, if application enables ANY IRQ (for example IRQ num 240 which doesn't exist), IRQ_RTC1 starts working also. Due this, default settings with serial enabled, keeps IRQ_RTC1 working properly and hides a bug.

At least 0.31 worked properly as I have used serial=none setting a long time. Please keep older versions at brew to allow easily allow switching version. Now it is pretty PIA.

perttierkkila:
Same problem persists with 0.34.0.
`tinygo version 0.34.0 darwin/arm64 (using go version go1.22.8 and LLVM version 18.1.2)`

Using pca10040-board with blink-code
```
package main

import (
	"machine"
	"time"
)

func main() {
	machine.LED.Configure(machine.PinConfig{Mode: machine.PinOutput})
	for {
		machine.LED.Set(!machine.LED.Get())
		time.Sleep(300 * time.Millisecond)
	}
}
```

When using normal pca10040-target, blinking works. But when modifying pca10040.json and disabling serial, nothing happens anymore. I tracked earlier the reason to IRQ enabling.
```
	"serial": "uart",
```
=>
```
	"serial": "none",
```

perttierkkila:
Tested building images at CI-server using official docker-image. Same problem still exists.
`tinygo version 0.33.0 linux/amd64 (using go version go1.23.0 and LLVM version 18.1.2)`

And tried 0.34 at CI, the problem still exists
`tinygo version 0.34.0 linux/amd64 (using go version go1.23.2 and LLVM version 18.1.2)`

aykevl:
[...]

PR #4572, opened 2024-10-31
interrupt: fix bug in interrupt lowering

aykevl:
The alignment wasn't set, so defaulted to 4 (for a 32-bit int). LLVM saw this, and therefore assumed that a ptrtoint of the pointer would have had the lowest bits unset. That's an entirely valid optimization, except that we are using these globals for arbitrary values (and aren't actually using these globals).

Fixed by setting alignment to 1. It works, though long-term we should maybe find a different solution for this.

This is the root cause of https://github.com/tinygo-org/tinygo/issues/4527, which I finally found after some extensive debugging why a single constant would change whether I called `arm.EnableIRQ` with an arbitrary value or not.

### DIFF
--- a/compiler/interrupt.go
+++ b/compiler/interrupt.go
@@ -41,6 +41,8 @@ func (b *builder) createInterruptGlobal(instr *ssa.CallCommon) (llvm.Value, erro
 
 	// Create a new global of type runtime/interrupt.handle. Globals of this
 	// type are lowered in the interrupt lowering pass.
+	// It must have an alignment of 1, otherwise LLVM thinks a ptrtoint of the
+	// global has the lower bits unset.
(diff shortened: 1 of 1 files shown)


dgryski (approved):
LGTM.  Any other downsides of having an alignment of 1?

aykevl:
No, not that I can think of. If it were a global that was actually used it would mean that on systems that don't support unaligned memory access (mostly Cortex-M0 and Cortex-M0+) it might cause problems. But this global is purely used for tracking compiler state, it's never actually accessed.

I also see there are some driver example binaries that get slightly larger. Not sure why, the only thing I can think of is an optimization that happened slightly differently. Or, indeed, it actually fixes a bug there.

PR #4573, opened 2024-10-31
sync: implement WaitGroup using a futex

aykevl:
This prepares sync.WaitGroup for multithreading.
Code size for the cooperative scheduler is nearly unchanged.

~This PR depends on #4560 and #4549.~

### DIFF
--- a/src/sync/waitgroup.go
+++ b/src/sync/waitgroup.go
@@ -3,35 +3,65 @@ package sync
 import "internal/task"
 
 type WaitGroup struct {
-	counter uint
-	waiters task.Stack
+	futex task.Futex
 }
 
 func (wg *WaitGroup) Add(delta int) {
-	if delta > 0 {
-		// Check for overflow.
-		if uint(delta) > (^uint(0))-wg.counter {
-			panic("sync: WaitGroup counter overflowed")
-		}
-
-		// Add to the counter.
-		wg.counter += uint(delta)
-	} else {
-		// Check for underflow.
-		if uint(-delta) > wg.counter {
-			panic("sync: negative WaitGroup counter")
-		}
-
-		// Subtract from the counter.
-		wg.counter -= uint(-delta)
-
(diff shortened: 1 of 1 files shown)


aykevl:
Rebased, this PR should be ready now.

aykevl:
Any feedback on this PR?

deadprogram:
@dgryski can you take a peek at this PR please?

src/sync/waitgroup.go
```diff
@@ -3,35 +3,62 @@ package sync
...
-		// Check for underflow.
-		if uint(-delta) > wg.counter {
-			panic("sync: negative WaitGroup counter")
+			// Add to the counter.
+			if wg.futex.CompareAndSwap(counter, counter+uint32(delta)) {
+				// Successfully added.
+				return
+			}
 		}
+	} else {
```
dgryski: Remove `else`.  The `if` case lands in a `for {}` that either `panic`s or `return`s.

dgryski (approved):
LGTM.  Code cleanup to unindent the `else` is optional.

src/sync/waitgroup.go
aykevl: I prefer to keep the structure as-is: while for things like fast paths and error return it makes sense to have an early return, in this case there are two equal cases: positive or negative/zero.
I can use a switch instead, though?

```go
switch {
case delta > 0:
    // positive
default:
    // negative (or zero)
}
```

aykevl:
Merged the first two commits to get other PRs ready. Now it's just the actual WaitGroup commit.
(Also changed comments slightly).

src/sync/waitgroup.go
deadprogram: > I can use a switch instead, though?

:+1: IMO

aykevl: Ok, updated.

deadprogram:
Thank you for working on these improvements @aykevl and thank you @dgryski for review. Now merging.

PR #4574, opened 2024-10-31
runtime: optimize findHead

aykevl:
This is similar to https://github.com/tinygo-org/tinygo/pull/3899, but smaller and hopefully just as efficient.
Needs a bit of cleanup before being ready.


dgryski (approved):
LGTM

PR #4581, opened 2024-11-03
goenv: read git hash embedded in the binary

aykevl:
This PR makes two practical changes:

  * The git hash is read from the binary itself instead of embedding it with a `-ldflags` flag. This means it is also present when building TinyGo using `go build` or `go install`.
  * A flag is added to indicate whether the tree was modified. This might aid in helping people: it is now easier to see whether they have made local modifications.

### DIFF
--- a/goenv/version.go
+++ b/goenv/version.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"runtime/debug"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


goenv/version.go
```diff
@@ -4,29 +4,47 @@ import (
...
+	hash := ""
+	modified := false
+	if info, ok := debug.ReadBuildInfo(); ok {
+		for _, setting := range info.Settings {
+			println("setting:", setting.Key, setting.Value)
+			switch setting.Key {
+			case "vcs.revision":
+				hash = setting.Value[:8]
+			case "vcs.modified":
+				modified = setting.Value == "true"
```
sago35: If possible, I'd like the state to be unmodified when building from CI. Note that `vcs.modified` will still return true if there are any untracked files present.

The following has been verified using the binary generated by CI.

```
>.\bin\tinygo.exe version
.....
setting: vcs git
setting: vcs.revision e5f95be5d00ceb83ddce36374bfb06ba81c9264e
setting: vcs.time 2024-11-03T09:08:14Z
setting: vcs.modified true
tinygo version 0.35.0-dev-e5f95be5-modified windows/amd64 (using go .\bin\tinygo.exe versionM version 18.1.2)

```

aykevl:
@sago35 Thanks, that's certainly not the intention!
I've removed the `-modified` suffix for now, it can be added later if needed (with the right `.gitignore` so that it won't show up as modified in the Windows CI).

I've also removed some debugging code I accidentally left in (a `println`) and simplified the code a bit.

sago35 (approved):
LGTM

eliasnaur (commented):
FWIW, LGTM.

PR #4585, opened 2024-11-06
cgo: support second return value (errno)

aykevl:
This adds support for code like this:

```go
val, err := C.some_func()
if val == -1 {
    println("failed:", err.Error())
}
```

That is, the second return value is the errno value (a `syscall.Errno` value) after the call. Because of how C errno works, you can't just check errno: you have to check the normal return value to know whether the call succeeded or not, and only use errno when you know it failed and was therefore set by the call.

To make this work, I also had to implement some CGo features that weren't implemented yet (mainly for WASI which likes to use rather indirect errno values). I also changed the syscall package for wasm/js because the errno values in there didn't actually match the values in wasi-libc.

The reason I did this was to be able to run `tinygo test os/syscall`. That test needs some more CGo stuff but this is one of the prerequisites.

### DIFF
--- a/builder/picolibc.go
+++ b/builder/picolibc.go
@@ -34,6 +34,7 @@ var libPicolibc = Library{
 			"-D__OBSOLETE_MATH_FLOAT=1", // use old math code that doesn't expect a FPU
 			"-D__OBSOLETE_MATH_DOUBLE=0",
 			"-D_WANT_IO_C99_FORMATS",
+			"-D__PICOLIBC_ERRNO_FUNCTION=__errno_location",
 			"-nostdlibinc",
 			"-isystem", newlibDir + "/libc/include",
 			"-I" + newlibDir + "/libc/tinystdio",
--- a/cgo/cgo.go
+++ b/cgo/cgo.go
@@ -138,7 +138,8 @@ typedef unsigned long long  _Cgo_ulonglong;
 // first.
 // These functions will be modified to get a "C." prefix, so the source below
 // doesn't reflect the final AST.
-const generatedGoFilePrefix = `
+const generatedGoFilePrefixBase = `
+import "syscall"
 import "unsafe"
 
 var _ unsafe.Pointer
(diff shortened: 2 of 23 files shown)


aykevl:
Looks like I need to fix some issues in the cgo tests.

dgryski:
Does this increase or decrease our upstream Go compatibility?

aykevl:
Increase, because Go also supports this feature.
From the [docs](https://pkg.go.dev/cmd/cgo):

> Any C function (even void functions) may be called in a multiple assignment context to retrieve both the return value (if any) and the C errno variable as an error (use _ to skip the result value if the function returns void). For example:
>
>     n, err = C.sqrt(-1)
>     _, err := C.voidFunc()
>     var n, err = C.sqrt(1)

(We don't support the void return value yet, but if the C function returns a value it should now match upstream Go).

aykevl:
The testing failure with the cgo package should be fixed now.

aykevl:
There are some more errors. Should be easy to fix, I'll do that another day.

aykevl:
From CI:

```diff
--- expected
+++ actual
@@ -75,8 +75,8 @@
 len(C.GoBytes(nil, 0)): 0
 len(C.GoBytes(C.CBytes(nil),0)): 0
 rountrip CBytes: hello
-EINVAL: true
-EAGAIN: true
+EINVAL: false
+EAGAIN: false
 copied string: foobar
 CGo sqrt(3): +1.732051e+000
 C   sqrt(3): +1.732051e+000
```

Well that's just great. The reason is that while mingw-w64 defines actual error values, the Go stdlib syscall package fakes these error values on Windows. I wonder what's up with that.

aykevl:
Apparently it's a known issue in the Go toolchain as well: https://github.com/golang/go/issues/23468

I guess I can fix it by modifying `__get_errno` to map the C errno value to the fake syscall.Errno value on Windows.

cgo/libclang.go
```diff
@@ -488,6 +452,62 @@ func (f *cgoFile) createASTNode(name string, c clangCursor) (ast.Node, any) {
...
+			// Later tokens are the macro contents.
+			for int(tokenOffset) > (startOffset + sourceBuf.Len()) {
+				// Pad the source text with whitespace (that must have been
+				// present in the original source as well).
+				sourceBuf.WriteByte(' ')
+			}
+			sourceBuf.WriteString(spelling)
+		}
+	}
+	C.clang_disposeTokens(tu, rawTokens, numTokens)
```
eliasnaur: Nit: convert to earlier `defer`?

eliasnaur (commented):
Comment from a cursory review.

cgo/libclang.go
aykevl: Fixed!

aykevl:
Updated, hopefully this makes most errno values on Windows actually work.

aykevl:
This PR would be useful to simplify #4603.

dgryski:
This is still failing CI with what looks like legit errors.

aykevl:
Yes, I need to fix that. Specifically, I think it's the third commit (`syscall: use wasi-libc tables for wasm/js target`).
I extracted https://github.com/tinygo-org/tinygo/pull/4607 to make this PR smaller and easier to debug.

aykevl:
Also extracted that third commit into https://github.com/tinygo-org/tinygo/pull/4611

aykevl:
Rebased so it's just the addition of the errno return value.

dgryski (approved):
LGTM

PR #4586, opened 2024-11-07
Implement race-free signals using futexes

aykevl:
Use a futex instead of the previous rather complicated code with signal masking. It was race free (I hope!) on Linux but not on MacOS. This PR makes it race-free on MacOS too.

Sleeping is very simple: just use a futex with a timeout and if the futex value changed we know there was a signal. There shouldn't be any race conditions with this approach.

This also adds an `internal/futex` package which I intend to reuse in #4559.


src/internal/futex/futex_linux.c
```diff
@@ -0,0 +1,33 @@
...
+
+#include <limits.h>
+#include <stdint.h>
+#include <sys/syscall.h>
+#include <time.h>
+#include <unistd.h>
+
+#define FUTEX_WAIT         0
+#define FUTEX_WAKE         1
+#define FUTEX_PRIVATE_FLAG 128
```
eliasnaur, line 14: Why not `#include <linux/futex.h>`?

src/internal/futex/futex_darwin.c
```diff
@@ -0,0 +1,37 @@
+//go:build none
+
+// This file is manually included, to avoid CGo which would cause a circular
+// import.
+
+#include <stdint.h>
+
+int __ulock_wait(uint32_t operation, void *addr, uint64_t value, uint32_t timeout_us);
+int __ulock_wait2(uint32_t operation, void *addr, uint64_t value, uint64_t timeout_ns, uint64_t value2);
+int __ulock_wake(uint32_t operation, void *addr, uint64_t wake_value);
```
eliasnaur, line 14: Link to documentation, if it exists somewhere (semi-)stable.

src/internal/futex/futex.go
```diff
@@ -0,0 +1,74 @@
...
+// to sleep (atomically) to prevent lost wake-ups.
+type Futex struct {
+	atomic.Uint32
+}
+
+// Atomically check for cmp to still be equal to the futex value and if so, go
+// to sleep. Return true if we were definitely awoken by a call to Wake or
+// WakeAll, and false if we can't be sure of that.
+func (f *Futex) Wait(cmp uint32) bool {
+	tinygo_futex_wait((*uint32)(unsafe.Pointer(&f.Uint32)), cmp)
```
eliasnaur, line 26: It's probably ok to assume the layout of `atomic.Uint32` here and below, but it seems to me you can avoid it by

```
type Futex struct {
    v uint32
}
```

and adding methods that forwards to atomic.(Load|And|Store...)Uint32 functions.

src/internal/futex/futex.go
```diff
@@ -0,0 +1,74 @@
...
+	//
+	// I'm not sure whether we do anything like pthread does, so to be on the
+	// safe side we say we don't know whether the wakeup was spurious or not and
+	// return false.
+	return false
+}
+
+// Like Wait, but times out after the number of nanoseconds in timeout.
+// If timeout is 0, it may or may not be treated as Wait with infinite timeout.
+// Therefore, make sure the timeout value is non-zero.
```
eliasnaur: Given https://github.com/tinygo-org/tinygo/issues/4568 et al can a zero timeout be tightened to always do the least surprising thing, even if it means `timeout = max(timeout, 1)` here or in the platform implementations?

src/internal/futex/futex_darwin.c
```diff
@@ -0,0 +1,37 @@
...
+
+void tinygo_futex_wait(uint32_t *addr, uint32_t cmp) {
+    __ulock_wait(UL_COMPARE_AND_WAIT|ULF_NO_ERRNO, addr, (uint64_t)cmp, 0);
+}
+
+void tinygo_futex_wait_timeout(uint32_t *addr, uint32_t cmp, uint64_t timeout) {
+    // Note: __ulock_wait2 is available since MacOS 11.
+    // I think that's fine, since the version before that (MacOS 10.15) is EOL
+    // since 2022. Though if needed, we could certainly use __ulock_wait instead
+    // and deal with the smaller timeout value.
```
eliasnaur, line 39: Nit: I would put (most of) this note in the commit message. Go 1.23 also requires macOS 11 and it doesn't seem reasonable to do better in TinyGo.

src/runtime/runtime_unix.go
```diff
@@ -222,46 +224,30 @@ func nanosecondsToTicks(ns int64) timeUnit {
...
 
 func sleepTicks(d timeUnit) {
-	// When there are no signal handlers present, we can simply go to sleep.
-	if !hasSignals {
-		// timeUnit is in nanoseconds, so need to convert to microseconds here.
-		usleep(uint(d) / 1000)
-		return
+	var until timeUnit
+	if !hasScheduler {
+		until = ticks() + d
```
eliasnaur: I don't understand why this code depends on `hasScheduler`.

src/runtime/runtime_unix.go
```diff
@@ -222,46 +224,30 @@ func nanosecondsToTicks(ns int64) timeUnit {
...
-		}
-	} else {
-		// Linux (and various other POSIX systems) implement sigtimedwait so we
-		// can do this in a non-racy way.
-		tinygo_wfi_mask(activeSignals)
-		if checkSignals() {
-			tinygo_wfi_unmask()
+		// Set duration (in next loop iteration) to the remaining time.
+		d = until - ticks()
+		if d <= 0 {
```
eliasnaur, line 249: If `hasScheduler`, d will always be less than zero, no?

eliasnaur (commented):
Nice improvement. A few comments from the peanut gallery.

src/internal/futex/futex_linux.c
aykevl: Because this header file is not available in musl, which we use.

src/internal/futex/futex.go
aykevl: I'd prefer to keep using `atomic.Uint32`. I don't expect `atomic.Uint32` to ever change to something else, and if it does we can always update this code.

src/runtime/runtime_unix.go
aykevl: Yeah, that's a bug. Thanks for spotting it! (I probably modified the code a few times and it probably made sense to not calculate `until` without a scheduler before).
I'll push a fix.

src/runtime/runtime_unix.go
aykevl: Not necessarily! There are cases where `Wait` and `WaitUntil` might return due to spurious wakeups and not due to a signal. I'm not entirely sure whether this is possible, but the futex documentation on Linux is a bit vague on it (see src/internal/futex/futex.go) and since there's no real documentation for the MacOS syscalls I think it's better to defensively assume they can return spuriously. (I know that on Windows, the equivalent syscalls do return spuriously sometimes).

src/internal/futex/futex.go
aykevl: Done. I don't see a way this can happen currently (the two places that call `sleepTicks` make sure it's called with a positive timeout) but since this is MacOS and not some constrainted microcontroller it doesn't really hurt to put in an extra check.

src/internal/futex/futex_darwin.c
aykevl: Done. There's no real documentation, but someone has figured out these calls: https://outerproduct.net/futex-dictionary.html

dgryski:
Can you rebase so we get the fix for the wasm failure hitting the linux-assert build?

aykevl:
Rebased!

deadprogram:
Looks like everything has been addressed, so now merging.

Thank you @aykevl for implementing and to @eliasnaur and @dgryski for reviewing.

PR #4589, opened 2024-11-08
[DO NOT MERGE] testing Windows CI failures

aykevl:
Trying to figure out why Windows CI keeps failing without error.

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -26,6 +26,8 @@ jobs:
             goarch: arm64
     runs-on: ${{ matrix.os }}
     steps:
+      - name: exit early
+        run: command-does-not-exist
       - name: Install Dependencies
         run: |
           HOMEBREW_NO_AUTO_UPDATE=1 brew install qemu binaryen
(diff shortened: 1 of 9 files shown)


aykevl:
Update: so it's not that the binaries are different. Sometimes they run, sometimes they don't with the same binary (same hash). So there seems to be some sort of non-determinism _inside the compiled binary_. Perhaps something related to ASLR?

aykevl:
...but when I run the same binary a 1000 times, it passes every time? So maybe something to do with parallel execution?

aykevl:
The build fails when running tests in parallel, but succeeds when done serially? Even though the actual binary is the same (same hash). Something weird is going on here.

aykevl:
It appears that adding a sleep just before exiting works around the issue? That definitely points to some sort of race.

(I suspect it's needed to call `exit(0)` instead of returning, since that's what the C equivalent is doing in mingw-w64-crt/crt/crtexe.c).

aykevl:
I think I found it! It does indeed appear that exiting via `exit(0)` fixes this problem.
See: https://github.com/tinygo-org/tinygo/pull/4612

...it only took me a few hours to figure this out :weary:

aykevl:
Closing, since this issue has now been fixed.

PR #4593, opened 2024-11-11
compiler: link .syso files found in Go package directories

ydnar:
This is an experiment, driven by a need to link in custom sections into the resulting .wasm binary. Using a .syso file in a package directory is a well-known way to accomplish this with other architectures in big Go.

We have a POC that proves this works: https://github.com/ydnar/wasi-http-go/pull/13

Another example: https://tip.golang.org/src/crypto/internal/boring/syso/

@dgryski co-authored this live at WasmCon.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -711,6 +711,12 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 			}
 			linkerDependencies = append(linkerDependencies, job)
 		}
+
+		// Add .syso files
+		// TODO: is this the right way to do this?
+		for _, filename := range pkg.SysoFiles {
+			ldflags = append(ldflags, filepath.Join(pkg.Dir, filename))
+		}
 	}
 
 	// Linker flags from CGo lines:
--- a/compiler/symbol.go
+++ b/compiler/symbol.go
@@ -350,6 +350,10 @@ func (c *compilerContext) parsePragmas(info *functionInfo, f *ssa.Function) {
 			info.wasmModule = parts[1]
 			info.wasmName = parts[2]
 		case "//go:wasmexport":
+			if c.archFamily() != "wasm32" {
+				// go:wasmimport is ignored on non-wasm architectures
+				continue
+			}
 			if f.Blocks == nil {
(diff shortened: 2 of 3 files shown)


aykevl:
While this is one option, I don't think we should be going for this solution. The reason is that `.syso` files would fundamentally be difficult to support in TinyGo, because existing `.syso` files will be expecting to use the Go calling convention (and not the TinyGo calling convention). So it would remain a wasm-only solution, which doesn't feel right to me (it feels like the wrong tool for the job).

Do you have any objections against a special-purpose pragma for this purpose? It would basically just add the file to the linker, nothing else.

Also, since this is going to be generated code and the uglyness is hidden, I don't see any fundamental reason why the `#cgo` hack couldn't be used? Certainly not pretty, no, but nobody except wasm-bindgen-go developers need to care about this. And it doesn't need any TinyGo changes as far as I can see.

ydnar:
The Cgo hack can’t be used because it creates a circular dependency on package `syscall`.

What if I limited this PR to just `wasm32`?

aykevl:
> The Cgo hack can’t be used because it creates a circular dependency on package `syscall`.

Yeah, you're right.

> What if I limited this PR to just `wasm32`?

Limiting to wasm would indeed help against my worries. Or even only accepting a single filename (I don't know, something like `wasi-component.syso`) so that the intended use is very clear. It's easy to lift such restrictions in the future if needed, but adding them in the future may be difficult if people started using these .syso files for other purposes.

ydnar:
One complication: -Target=wasip2 sets GOARCH=arm GOOS=linux.

This means that foo_wasm.syso would not be found.

Do you have any plans to change the GOARCH/GOOS handling for architectures that TinyGo supports that Go doesn't?

aykevl:
> Do you have any plans to change the GOARCH/GOOS handling for architectures that TinyGo supports that Go doesn't?

We can't, really. We use `go list` to get a list of files to compile. So we have to pretend to be something that the standard library supports (such as linux/arm). Because wasip1 support exists in the stdlib, we can use the actual GOOS/GOARCH values, but sadly that's not the case for wasip2.

Issue #4557, opened 2024-10-27
Cannot build anything with tinygo on windows


RodargDev:
I have followed all steps given in this installation guide (https://tinygo.org/getting-started/install/windows/)

I currently have go and tinygo installed through scoop with versions:
```go version go1.23.2 windows/amd64```
```tinygo version 0.34.0 windows/amd64 (using go version go1.23.2 and LLVM version 18.1.2)```

I have a small program with the following code:
```
package main

import "fmt"

func main() {
	fmt.Println("Hello, World!")
}
```

But when i try to use tinygo to build this for the pico (Or any other targer) i get a huge stacktrace with a lot of gibberish characters. Only the start and end of the stacktrace give some context into whats going on.

The command i use
```tinygo flash -target=pico .\main.go```

The output i get in console
```
PS C:\Projects\WaterLevelSensor> tinygo flash -target=pico .\main.go
error: unable to make temporary file: No such file or directory
# 1 "C:\\Users\\info\\scoop\\apps\\tinygo\\current\\lib\\compiler-rt-builtins\\arm\\aeabi_cdcmp.S"
error: unable to make temporary file: No such file or directory
# 1 "<builerror: unable to make temporary file: No such file or directory
t-in>" 1
# 1 "<built-in>" 3
# 1 "C:\\Users\\info\\scoop\\apps\\tinygo\\current\\lib\\compiler-rt-builtins\\arm\\aeabi_cfcmp# 1# 377 "<built-in>" 3
# 1 "<command line>" 1
 "C:\.S"
\Users\\info\\scoop\\apps\\tinygo\\curren# error: unable to make temporary file: No such file or directory
t\\# 1 "<built-in>" 1
lib\\compile# 1 "<built-in>" 31 "<buir
lt-in>" 2
-# 1 rt-built"C:\\User# 377 "<built-in>" 3
# s\\# 1 "C:\\Users\\info\\scoop\\apps\\tininfo\\scygo\\cins\\arm\\aourrent\\lib\\compieabi_dcmp.S"
o# 1 1 ler-rp"<bt-builtiuierror: unable to make temporary file: No such file or directoryns\\arl"<com
\\ammt-and line>" 1
i\\aeabi_#n>" 1
# ppfc1 "<s\\tinygo\\current\\lib\\com 1 "<bmbpiler-uilt-in>"p.S"
uilt-in>" 3
 2
[...]

PR #4596, opened 2024-11-12
builder: whitelist temporary directory env var for Clang invocation

aykevl:
It looks like this breaks on Windows: https://github.com/tinygo-org/tinygo/issues/4557

I haven't confirmed this is indeed the problem, but it would make sense. And passing through the temporary directory seems like a good idea regardless, there's not much that could break due to that.

### DIFF
--- a/builder/tools.go
+++ b/builder/tools.go
@@ -32,9 +32,25 @@ func runCCompiler(flags ...string) error {
 	cmd.Stderr = os.Stderr
 
 	// Make sure the command doesn't use any environmental variables.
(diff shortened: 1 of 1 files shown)


dgryski (approved):
I'll say LGTM so we can unbreak Windows, but I think I'd prefer looping over the array calling `os.LookupEnv` for each environment variable, adding it to `cmd.Env` if we get a hit.

deadprogram:
Thanks for the fix @aykevl and to @dgryski for review. Now merging.

PR #4599, opened 2024-11-13
targets/wasm_exec: call process.exit() when go.run() returns

dgryski:
`TestTest/WASM/Fail` started failing with the last patch to this file.  This (I hope) restores the correct behaviour.

### DIFF
--- a/targets/wasm_exec.js
+++ b/targets/wasm_exec.js
@@ -531,7 +531,10 @@
 
 		const go = new Go();
 		WebAssembly.instantiate(fs.readFileSync(process.argv[2]), go.importObject).then((result) => {
-			return go.run(result.instance);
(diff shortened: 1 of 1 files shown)


targets/wasm_exec.js
```diff
@@ -531,7 +531,10 @@
 
 		const go = new Go();
 		WebAssembly.instantiate(fs.readFileSync(process.argv[2]), go.importObject).then((result) => {
-			return go.run(result.instance);
+			go.run(result.instance).then((result) => {
+                            process.exit(result);
+                        }).
+                        catch((e) => { throw e });
```
aykevl, line 537: There seem to be some whitespace issues here?

aykevl:
Merging because it fixes CI.

PR #4601, opened 2024-11-15
support to parse devl version

### LINKED ISSUE #4474: How to build source code using Go built from source code
## Go version

```cli
go version
go version devel go1.24-b049837d97 Tue Sep 17 20:44:51 2024 +0000 darwin/arm64
```

## TinyGo version

```cli
tinygo version
tinygo version 0.33.0 darwin/arm64 (using go version devel go1.24-b049837d97 Tue Sep 17 20:44:51 2024 +0000 and LLVM version 18.1.2)
```

## Source codes

I attempted to run these.

ref. https://tinygo.org/docs/guides/webassembly/wasm/

```go
package main

// This calls a JS function from Go.
func main() {
	println("adding two numbers:", add(2, 3)) // expecting 5
}

// This function is imported from JavaScript, as it doesn't define a body.
// You should define a function named 'add' in the WebAssembly 'env'
// module from JavaScript.
//
//export add
func add(x, y int) int

// This function is exported to JavaScript, so can be called using
// exports.multiply() in JavaScript.
//
//export multiply
func multiply(x, y int) int {
	return x * y
}
```

```js
// Providing the environment object, used in WebAssembly.instantiateStreaming.
// This part goes after "const go = new Go();" declaration.
go.importObject.env = {
    'add': function(x, y) {
        return x + y
    }
    // ... other functions
}

// Calling the multiply function:
console.log('multiplied two numbers:', wasm.exports.multiply(5, 3));
```

### Build

```cli
GOOS=js GOARCH=wasm tinygo build -o wasm.wasm ./main.go

could not parse Go version: version does not start with 'go' prefix
```

This error occurs at [this line](https://github.com/tinygo-org/tinygo/blob/release/goenv/version.go#L39).

I want to know the way how to avoid this error. Who knows tips ?
Or, ideally, if the `devel` string is skipped, this error will be solved.
At least, Go version can't be modified, so the aggressive parsing is valuable for me or someone (maybe). But I'm not familiar with TinyGo, so I want to know the mainter's opinion.
How about ?

```cli
go env -w GOVERSION='1.22'
go: GOVERSION cannot be modified
```

sivchari:
close #4474 

Support to parse devl version (Go upstream version).

### DIFF
--- a/goenv/version.go
+++ b/goenv/version.go
@@ -50,6 +50,9 @@ func GetGorootVersion() (major, minor int, err error) {
 // major, minor, and patch version: 1, 3, and 2 in this example.
 // If there is an error, (0, 0, 0) and an error will be returned.
 func Parse(version string) (major, minor, patch int, err error) {
+	if strings.HasPrefix(version, "devel ") {
+		version = strings.Split(strings.TrimPrefix(version, "devel "), version)[0]
+	}
 	if version == "" || version[:2] != "go" {
 		return 0, 0, 0, errors.New("could not parse Go version: version does not start with 'go' prefix")
 	}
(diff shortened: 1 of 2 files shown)


aykevl:
Looks reasonable, thank you!

PR #4602, opened 2024-11-15
compiler, runtime: move constants into shared package

aykevl:
Use a single package for certain constants that must be the same between the compiler and the runtime.

While just using the same values in both places works, this is much more obvious and harder to mess up. It also avoids the need for comments pointing to the other location the constant is defined. And having it in code makes it possible for IDEs to analyze the source.

In the future, more such constants and maybe algorithms can be added. For example: constants for the reflect package.

---

Fee free to bikeshed about the package name. I can't put it under `internal/` (the compiler won't let me) so that's why it is top level.

The direct reason why I want to have this package is because I'd like to use this for a GC optimization I'm working on that's kinda tricky (reusing the remaining stack for the GC mark stack). For that, some constants really do need to be the same that may look like tunable constants.

### DIFF
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/tinygo-org/tinygo/compiler/llvmutil"
 	"github.com/tinygo-org/tinygo/loader"
+	"github.com/tinygo-org/tinygo/src/tinygo"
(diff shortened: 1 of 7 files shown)


dgryski (approved):
LGTM.  Useful, thanks.

PR #4603, opened 2024-11-15
runtime/fcntl

leongross:
Rework of #4481

### DIFF
--- a/src/runtime/os_linux.go
+++ b/src/runtime/os_linux.go
@@ -5,7 +5,9 @@ package runtime
 // This file is for systems that are _actually_ Linux (not systems that pretend
 // to be Linux, like baremetal systems).
 
-import "unsafe"
+import (
+	"unsafe"
+)
 
 const GOOS = "linux"
 
@@ -83,6 +85,11 @@ type elfProgramHeader32 struct {
 //go:extern __ehdr_start
 var ehdr_start elfHeader
 
+// int *__errno_location(void);
+//
+//export __errno_location
+func libc_errno_location() *int32
+
 // findGlobals finds globals in the .data/.bss sections.
 // It parses the ELF program header to find writable segments.
 func findGlobals(found func(start, end uintptr)) {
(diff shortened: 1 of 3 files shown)


src/runtime/os_linux.go
```diff
@@ -139,3 +139,12 @@ func hardwareRand() (n uint64, ok bool) {
 //
 //export getrandom
 func libc_getrandom(buf unsafe.Pointer, buflen uintptr, flags uint32) uint32
+
+// int fcntl(int fd, int cmd, int arg);
+//
+// export fcntl
+func libc_fcntl(fd int, cmd int, arg int) (ret int)
+
+func fcntl(fd int, cmd int, arg int) int {
```
aykevl: This must match the signature in the syscall/unix package:

```go
func fcntl(fd int32, cmd int32, arg int32) (int32, int32)
```

src/runtime/os_linux.go
```diff
@@ -139,3 +146,16 @@ func hardwareRand() (n uint64, ok bool) {
...
+
+// int fcntl(int fd, int cmd, int arg);
+//
+// export fcntl
+func libc_fcntl(fd int, cmd int, arg int) (ret int)
+
+func fcntl(fd int32, cmd int32, arg int32) (ret int32, errno int32) {
+	ret = int32(libc_fcntl(int(fd), int(cmd), int(arg)))
+	if ret < 0 {
+		return 0, int32(uintptr(*libc_errno_location()))
```
aykevl: Why this double cast?

```suggestion
		return 0, *libc_errno_location()
```

src/runtime/os_linux.go
```diff
@@ -139,3 +146,16 @@ func hardwareRand() (n uint64, ok bool) {
...
+//
+//export fcntl
+func libc_fcntl(fd int, cmd int, arg int) (ret int)
+
+func fcntl(fd int32, cmd int32, arg int32) (ret int32, errno int32) {
+	ret = int32(libc_fcntl(int(fd), int(cmd), int(arg)))
+	if ret < 0 {
+		return 0, int32(*libc_errno_location())
+	}
+	return ret, 0
```
aykevl: This is not quite correct. If you look at where `runtime.fcntl` is used, the code looks like this:

```go
func Fcntl(fd int, cmd int, arg int) (int, error) {
        val, errno := fcntl(int32(fd), int32(cmd), int32(arg))
        if val == -1 {
                return int(val), syscall.Errno(errno)
        }
        return int(val), nil
}
```

So `val` needs to be -1 when there's an error.
This probably means you can simplify the code like this:

```suggestion
	ret = int32(libc_fcntl(int(fd), int(cmd), int(arg)))
	errno = *libc_errno_location()
	return
```

leongross:
Thanks a lot for reviewing and merging @aykevl

PR #4607, opened 2024-11-16
cgo: support more macros

aykevl:
This PR adds better support for C macros (`#define`) in CGo.

1. It makes sure that if a macro contains another macro (such as `#define foo (bar+1)`) the referenced macro is also defined on the Go side, to avoid errors (`undefined: C.something`).
2. It supports function-like macros inside macros. For example, `#define add(a, b) (a+b)` that can be used like `#define added add(3, 5)`.

Both of these will be necessary to fully support CGo errno values with wasi-libc.

I extracted these two from #4585 because these seem like relatively simple changes, to make the other PR smaller.

### DIFF
--- a/cgo/const.go
+++ b/cgo/const.go
@@ -54,8 +54,8 @@ func init() {
 }
 
 // parseConst parses the given string as a C constant.
-func parseConst(pos token.Pos, fset *token.FileSet, value string) (ast.Expr, *scanner.Error) {
-	t := newTokenizer(pos, fset, value)
+func parseConst(pos token.Pos, fset *token.FileSet, value string, f *cgoFile) (ast.Expr, *scanner.Error) {
+	t := newTokenizer(pos, fset, value, f)
 	expr, err := parseConstExpr(t, precedenceLowest)
 	t.Next()
 	if t.curToken != token.EOF {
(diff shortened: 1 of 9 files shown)


cgo/const.go
```diff
@@ -96,6 +160,65 @@ func parseConstExpr(t *tokenizer, precedence int) (ast.Expr, *scanner.Error) {
...
+		return val, nil
+	}
+
+	if t.f != nil {
+		// Check whether this identifier is actually a macro "call" with
+		// parameters. In that case, we should parse the parameters and pass it
+		// on to a new invocation of parseConst.
+		if t.peekToken == token.LPAREN {
+			if cursor, ok := t.f.names[t.curValue]; ok && t.f.isFunctionLikeMacro(cursor) {
+				t.Next()
```
b0ch3nski, line 178: `Next()` called twice - is it intentional? If so, would be good to have a comment here

aykevl: Yes this is intentional. There is the "cur" (current) token and the "peek" (next) token. At this point we know the current token and the peek token, and we want to skip ahead so that the "current" token is the first unknown token. Hence two `Next()` calls.

aykevl: Added a comment that should hopefully clarify this.

b0ch3nski: Thanks, makes sense :ok_hand:

deadprogram:
Thanks for the addition @aykevl and to @b0ch3nski for review. Now merging.

PR #4610, opened 2024-11-18
runtime: heapptr only needs to be initialized once

aykevl:
There is no need to initialize it twice.
See: https://github.com/tinygo-org/tinygo/pull/2082


dgryski (approved):
LGTM.  Fixing my very first PR :)

aykevl:
Thanks for the review!

PR #4611, opened 2024-11-18
syscall: refactor wasm environment variables and constants

aykevl:
Two changes for WebAssembly:
1. Move environment variables into their own files, and refactor WASIp2 environment variables to be a lot simpler (not going through the C API).
2. Use the correct constants for `-target=wasm`. Instead of using fake constants, use the ones that match wasi-libc.

This PR is extracted from #4585 because it's a bit more tricky than it initially seemed.

### DIFF
--- a/src/syscall/env_libc.go
+++ b/src/syscall/env_libc.go
@@ -55,5 +55,57 @@ func Environ() []string {
 	return envs
 }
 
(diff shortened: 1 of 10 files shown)


dgryski (approved):
LGTM

PR #4612, opened 2024-11-19
windows: don't return in mainCRTStartup, exit via `exit(0)` instead

aykevl:
This fixes a bug where output would not actually be written to stdout before exiting the process, leading to a flaky Windows CI. Exiting using `exit(0)` appears to fix this.

For some background, see: https://github.com/tinygo-org/tinygo/pull/4589

### DIFF
--- a/src/runtime/runtime_windows.go
+++ b/src/runtime/runtime_windows.go
@@ -52,7 +52,16 @@ func mainCRTStartup() int {
 	stackTop = getCurrentStackPointer()
 	runMain()
 
-	// For libc compatibility.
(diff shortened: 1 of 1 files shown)


aykevl:
At least this time it passed the tests :tada: 
But I'm fairly sure this is it, since the tests in https://github.com/tinygo-org/tinygo/actions/runs/11910781343/job/33191101738?pr=4589 also passed. And even if not, `exit(0)` should be harmless.

deadprogram (approved):
:tada: thank you!!!

PR #4613, opened 2024-11-19
ci: cache the Go cache across Windows builds

aykevl:
This should hopefully make the build slightly faster.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -100,6 +100,13 @@ jobs:
       - name: Build wasi-libc
         if: steps.cache-wasi-libc.outputs.cache-hit != 'true'
         run: make wasi-libc
+      - name: Cache Go cache
+        uses: actions/cache@v4
+        with:
+          key: go-cache-windows-v1-${{ hashFiles('go.mod') }}
+          path: |
+            C:/Users/runneradmin/AppData/Local/go-build
+            C:/Users/runneradmin/go/pkg/mod
       - name: Install wasmtime
         run: |
           scoop install wasmtime@14.0.4


.github/workflows/windows.yml
```diff
@@ -100,6 +100,13 @@ jobs:
       - name: Build wasi-libc
         if: steps.cache-wasi-libc.outputs.cache-hit != 'true'
         run: make wasi-libc
+      - name: Cache Go cache
+        uses: actions/cache@v4
+        with:
+          key: go-cache-windows-v1-${{ hashFiles('go.mod') }}
+          path: |
+            C:/Users/runneradmin/AppData/Local/go-build
```
deadprogram, line 108: How did you determine the path here?

deadprogram:
Looking at the results, I do not see any change in the overall build time.

.github/workflows/windows.yml
aykevl: By running `go env`, though I probably shouldn't hardcode these.

deadprogram:
@aykevl seems like this PR did not actually reduce the build time?

aykevl:
It's going to be a small effect, yes. And in this CI run did not apply since it was the first time and the cache wasn't stored yet.

You can see the cache got stored though:

```
Post job cleanup.
"C:\Program Files\Git\usr\bin\tar.exe" --posix -cf cache.tzst --exclude cache.tzst -P -C D:/a/tinygo/tinygo --files-from manifest.txt --force-local --use-compress-program "zstd -T0"
Cache Size: ~56 MB (58389498 B)
Cache saved successfully
Cache saved with key: go-cache-windows-v1-ca733cc5ef186bec57b6bb434b4a52f1d3749001aff847ca883100f35bbcc805
```

deadprogram:
> You can see the cache got stored though

OK, in that case, merging. Thanks!

PR #4616, opened 2024-11-20
compiler: Fix wasmimport -> wasmexport in error message

### LINKED ISSUE #4615: Wrong error for invalid wasmexport
Hi! Using tinygo 0.34.0 I tried compiling this example from the [doc](https://tinygo.org/docs/guides/webassembly/wasi/).
I got the same error as #4464.
Trying to fix it I wrote:
```go
//go:wasmexport yourmodulename add
func add(x, y uint32) uint32 {
	return x + y
}
```
and got the error:
```
main.go:4:6: expected one parameter to //go:wasmimport, not 2
```
which wrongly refers to `//go:wasmimport` instead of `//go:wasmexport`.
This is easily fix here:
https://github.com/tinygo-org/tinygo/blob/2a76ceb7dd5ea5a834ec470b724882564d9681b3/compiler/symbol.go#L348

dgryski:
Fixes #4615

### DIFF
--- a/compiler/symbol.go
+++ b/compiler/symbol.go
@@ -345,7 +345,7 @@ func (c *compilerContext) parsePragmas(info *functionInfo, f *ssa.Function) {
 				continue
 			}
 			if len(parts) != 2 {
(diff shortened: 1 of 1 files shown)


aykevl:
Oops, thanks for the fix!

PR #4618, opened 2024-11-20
interp: align created globals

aykevl:
Use the alignment from the align attribute of the runtime.alloc call. This is going to be a more accurate alignment, and is typically smaller than the default.

This also reduces binary size and static RAM usage slightly in some cases.

### DIFF
--- a/builder/sizes_test.go
+++ b/builder/sizes_test.go
@@ -43,7 +43,7 @@ func TestBinarySize(t *testing.T) {
 		// microcontrollers
 		{"hifive1b", "examples/echo", 4600, 280, 0, 2268},
 		{"microbit", "examples/serial", 2908, 388, 8, 2272},
-		{"wioterminal", "examples/pininterrupt", 6140, 1484, 116, 6832},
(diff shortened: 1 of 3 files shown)


builder/sizes_test.go
```diff
@@ -43,7 +43,7 @@ func TestBinarySize(t *testing.T) {
 		// microcontrollers
 		{"hifive1b", "examples/echo", 4600, 280, 0, 2268},
 		{"microbit", "examples/serial", 2908, 388, 8, 2272},
-		{"wioterminal", "examples/pininterrupt", 6140, 1484, 116, 6832},
+		{"wioterminal", "examples/pininterrupt", 6140, 1484, 116, 6824},
```
aykevl, line 46: :tada:

aykevl:
This does indeed reduce SRAM and flash usage slightly in many cases (see the "size difference" comment above).

builder/sizes_test.go
deadprogram: Every byte helps!

deadprogram:
Thanks @aykevl now merging.

PR #4619, opened 2024-11-20
sync: only use a real lock in `sync.Map` when needed

aykevl:
There's no need for a real mutex with a cooperative scheduler (like `-scheduler=tasks`). So we can use a fake one instead.

This is a small PR mainly to show/implement `task.PMutex` which I'm going to use in other PRs.

### DIFF
--- a/src/sync/map.go
+++ b/src/sync/map.go
@@ -1,10 +1,12 @@
 package sync
 
+import "internal/task"
+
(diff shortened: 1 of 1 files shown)


dgryski (approved):
LGTM once CI is fixed

aykevl:
Also rebased.

deadprogram:
Thanks for the improvement @aykevl and to @dgryski for review.

PR #3887, opened 2023-09-02
Add `#cgo noescape` directive

aykevl:
This would have been useful to fix https://github.com/tinygo-org/bluetooth/issues/176.

Unfortunately, once I finished the implementation I found that for this to work correctly, Go 1.22 needs to be used. So marking this as draft until Go 1.22 is released.

PR #4621, opened 2024-11-20
fix: add updated test output to match recent changes

deadprogram:
This PR is to correct test fails from recent changes to runtime channel implementation in #5874


deadprogram:
Committing fix to get CI build back to green.

aykevl:
Also, LGTM. I think it was #3887 and #4556 together that caused this problem.

PR #4622, opened 2024-11-21
runtime: fix regression introduced by merging conflicting PRs

aykevl:
This was probably #4586 and #4549 that conflicted.


deadprogram:

PR #4623, opened 2024-11-21
runtime: implement `Goexit`

aykevl:
This is needed for full support of the testing package.

### DIFF
--- a/src/runtime/panic.go
+++ b/src/runtime/panic.go
@@ -38,12 +38,24 @@ type deferFrame struct {
 	JumpPC     unsafe.Pointer                 // pc to return to
 	ExtraRegs  [deferExtraRegs]unsafe.Pointer // extra registers (depending on the architecture)
 	Previous   *deferFrame                    // previous recover buffer pointer
-	Panicking  bool                           // true iff this defer frame is panicking
+	Panicking  panicState                     // not panicking, panicking, or in Goexit
 	PanicValue interface{}                    // panic value, might be nil for panic(nil) for example
 }
 
+type panicState uint8
+
+const (
+	panicFalse panicState = iota
+	panicTrue
+	panicGoexit
+)
+
 // Builtin function panic(msg), used as a compiler intrinsic.
 func _panic(message interface{}) {
+	panicOrGoexit(message, panicTrue)
+}
+
+func panicOrGoexit(message interface{}, panicking panicState) {
 	if panicStrategy() == tinygo.PanicStrategyTrap {
 		trap()
 	}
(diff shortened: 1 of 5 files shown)


testdata/recover.go
```diff
@@ -22,6 +27,9 @@ func main() {
 
 	println("\n# defer panic")
 	deferPanic()
+
+	println("\n# runtime.Goexit")
```
dgryski, line 31: Shouldn't `recover.txt` golden output also be updated here?

src/runtime/panic.go
```diff
@@ -53,11 +65,16 @@ func _panic(message interface{}) {
...
-			frame.Panicking = true
+			frame.Panicking = panicking
 			tinygo_longjmp(frame)
 			// unreachable
 		}
 	}
+	if panicking == panicGoexit {
+		// Call to Goexit() instead of a panic.
+		// Exit the goroutine instead of printing a panic message.
+		deadlock()
```
dgryski, line 76: I'm not sure `deadlock` is actually the right thing here.   This means (for example) if we spawn 1000 goroutines and  `Goexit()` all of them, they'll still be hanging around in the scheduler.  How do we get them to *actually* clean up?  What does a "regular" goroutine do at the end of its lifecycle?

dgryski (changes requested):
Test output needs updating.

src/runtime/panic.go
aykevl: `deadlock` needs to be changed, yes. But at the moment, it will do the exact same thing a goroutine does when it returns: it calls `Pause` without a way to wake it up. At that point the GC should be able to clean them up (assuming no false positives and such).

I plan to change this to something better because with threading support, the OS thread really does need to be exited.

testdata/recover.go
aykevl: Oops, yes! Fixed now :grimacing:

src/runtime/panic.go
dgryski: OK, so maybe for me it's just the name of the routine "deadlock" that didn't communicate what was actually happening.

dgryski:
CI failure.  Otherwise LGTM.
```
    --- FAIL: TestCompiler/defer.go-cortex-m-qemu (0.47s)
        compiler_test.go:125: output does not match expected output:
```

aykevl:
Updated the test.

deadprogram:
Since all feedback was addressed and @dgryski said LGTM I think this is now ready to merge. Thanks @aykevl and @dgryski

Issue #4608, opened 2024-11-17
gen-device-svd.go generates invalid assembler files for svd files with multi-line descriptions


cibomahto:
The [rp2350 SVD file](https://github.com/cmsis-svd/cmsis-svd-data/blob/main/data/RaspberryPi/rp2350.svd) has a long, multi-line description tag inside the root device element:

```
<?xml version="1.0" encoding="utf-8"?>
<!--
Copyright (c) 2024 Raspberry Pi Ltd.

SPDX-License-Identifier: BSD-3-Clause
-->
<device schemaVersion="1.1" xmlns:xs="http://www.w3.org/2001/XMLSchema-instance" xs:noNamespaceSchemaLocation="CMSIS-SVD.xsd">
    <vendor>Raspberry Pi</vendor>
    <name>RP2350</name>
    <series>RP</series>
    <version>0.1</version>
    <description>
        Dual Cortex-M33 or Hazard3 processors at 150MHz
        520kB on-chip SRAM, in 10 independent banks
        Extended low-power sleep states with optional SRAM retention: as low as 10uA DVDD
        8kB of one-time-programmable storage (OTP)
        Up to 16MB of external QSPI flash/PSRAM via dedicated QSPI bus
            Additional 16MB flash/PSRAM accessible via optional second chip-select
        On-chip switched-mode power supply to generate core voltage
            Low-quiescent-current LDO mode can be enabled for sleep states
        2x on-chip PLLs for internal or external clock generation
        GPIOs are 5V-tolerant (powered), and 3.3V-failsafe (unpowered)
        Security features:
            Optional boot signing, enforced by on-chip mask ROM, with key fingerprint in OTP
            Protected OTP storage for optional boot decryption key
            Global bus filtering based on Arm or RISC-V security/privilege levels
            Peripherals, GPIOs and DMA channels individually assignable to security domains
            Hardware mitigations for fault injection attacks
            Hardware SHA-256 accelerator
        Peripherals:
            2x UARTs
            2x SPI controllers
            2x I2C controllers
            24x PWM channels
            USB 1.1 controller and PHY, with host and device support
            12x PIO state machines
            1x HSTX peripheral
    </description>
    <width>32</width>
    <size>32</size>
```

Running `make gen-device` on this svd file produces a .s file with only the first line commented correctly, which cause a compilation error when including the file:

```
// Automatically generated file. DO NOT EDIT.
[...]

PR #4625, opened 2024-11-21
Update cmsis-svd library, and fix incompatiblitiy with gen-device-svd

cibomahto:
This pull request addresses #4608 , and also updates the cmsis-svd library to bring in the rp2350 svd file for #4452. I've grouped them together because the offending condition only seems to be present in the updated cmsis-svd library.

I wrote a python script to check that all of the generated .go files are syntactically correct, however i have not managed to check the .s files.

```
import pathlib
import subprocess
import sys

def check_go_files(target_dir):
    issues = []

    src = pathlib.Path(target_dir)
    for go_file in list(src.rglob('*.go')):
        command = f'gofmt -e {go_file}'

        output = subprocess.run(command.split(), capture_output=True)
        if output.returncode != 0:
            issues.append(go_file)

    return issues

target_dir = 'tinygo/src/device'

issues = []
issues.extend(check_go_files(target_dir))

if len(issues) > 0:
    print("the folling files have issues:", file=sys.stderr)
    for issue in issues:
        print(issue.as_posix())
    exit(1)
```


aykevl (approved):
LGTM.
@deadprogram any comments from you?

deadprogram:
Thanks for the update and fix for the generator @cibomahto now merging.

PR #4628, opened 2024-11-26
cgo: fix build warnings on Windows ARM

aykevl:
Fix the warning, and also remove `tinygo_clang_enum_visitor` which was unused.

See: https://github.com/golang/go/issues/49721


deadprogram (approved):
Thanks for the fix @aykevl

Issue #3353, opened 2022-12-24
Windows 386 build support


alchem1ster:
**Are there plans to add x86 support?**

- Windows 10 Pro 22H2 x64
- tinygo version 0.26.0 windows/amd64 (using go version go1.19.4 and LLVM version 14.0.0)
- GCC 12.2.0 + LLVM/Clang/LLD/LLDB 14.0.6 + MinGW-w64 10.0.0 (MSVCRT)

Input:
`$Env:GOARCH=386`
`tinygo build -no-debug -o result.exe`
Result:
`error: failed to hash file: open O:\TinyGo\src\runtime\asm_386_windows.S: The system cannot find the file specified.`

aykevl:
Correct, windows/386 is not currently supported. (But we should really be printing a more useful error message here).

I think windows/386 support can be added, but maintaining it will be some work. Can you describe why you need it?

alchem1ster:
>  Can you describe why you need it?

Creating tiny standalone utilities for x86-based Windows systems, with very weak CPUs (similar to Core 2 Duo) and limited RAM (up to 1GB). I can use the full Go, but tinyGo allows me to significantly reduce memory consumption.

BieHDC:
I started working on this and have a simple hello world printing before it crashes for yet unknown reasons.
I would be willing to keep working on this, however i am afraid i will require the support of someone who has deeper knowledge in the subject.

BieHDC:
I did the other things i wanted to achieve, now it is just about that crash i cant debug.
And i took down a semi fixme in the os_windows aswell while i was at it.

I parked the current work [here](https://github.com/tinygo-org/tinygo/compare/release...BieHDC:tinygo:release) for anyone who wants to take a look.

![goals](https://github.com/tinygo-org/tinygo/assets/43303168/c7dc6bbf-f618-4e75-adef-692bdbdef230)

aykevl:
My first guess would be that you use the wrong calling convention in task_stack_386_windows.S. You can try to build with `-scheduler=none` to skip that code and see if it works any better. I don't know exactly which calling convention is used on windows/386, there are a lot of them (see https://en.wikipedia.org/wiki/X86_calling_conventions).
Another thing you can try is use `-gc=leaking` to avoid including a GC, there's a small chance that might also skip some brokenness on windows/386.

BieHDC:
[...]

PR #4629, opened 2024-11-26
Windows: add i386 (windows/386) support

aykevl:
This is not fully working, but very simple programs do work.

The issue I'm currently hitting is that `syscall.loadsystemlibrary` (defined in src/runtime/runtime_windows.go) appears to be corrupting memory. Or maybe it just overwrites registers, I'm not sure.

See: #3353

### DIFF
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -398,15 +398,25 @@ func (c *Config) LibcCFlags() []string {
 	case "mingw-w64":
 		root := goenv.Get("TINYGOROOT")
 		path := c.LibraryPath("mingw-w64")
-		return []string{
+		cflags := []string{
 			"-nostdlibinc",
 			"-isystem", filepath.Join(path, "include"),
 			"-isystem", filepath.Join(root, "lib", "mingw-w64", "mingw-w64-headers", "crt"),
 			"-isystem", filepath.Join(root, "lib", "mingw-w64", "mingw-w64-headers", "include"),
 			"-isystem", filepath.Join(root, "lib", "mingw-w64", "mingw-w64-headers", "defaults", "include"),
-			"-D_UCRT",
-			"-D_WIN32_WINNT=0x0a00", // target Windows 10
 		}
+		if c.GOARCH() == "386" {
+			cflags = append(cflags,
+				"-D__MSVCRT_VERSION__=0x700", // Microsoft Visual C++ .NET 2002
+				"-D_WIN32_WINNT=0x0501",      // target Windows XP
+			)
+		} else {
+			cflags = append(cflags,
+				"-D_UCRT",
+				"-D_WIN32_WINNT=0x0a00", // target Windows 10
+			)
+		}
+		return cflags
 	case "":
 		// No libc specified, nothing to add.
 		return nil
--- a/compiler/symbol.go
+++ b/compiler/symbol.go
@@ -208,6 +208,12 @@ func (c *compilerContext) getFunction(fn *ssa.Function) (llvm.Type, llvm.Value) 
 			// > circumstances, and should not be exposed to source languages.
 			llvmutil.AppendToGlobal(c.mod, "llvm.compiler.used", llvmFn)
 		}
+	case "GetModuleHandleExA", "GetProcAddress", "GetSystemInfo", "GetSystemTimeAsFileTime", "LoadLibraryExW", "QueryUnbiasedInterruptTime", "SetEnvironmentVariableA", "Sleep", "VirtualAlloc":
+		// On Windows we need to use a special calling convention for some
+		// external calls.
+		if c.GOOS == "windows" && c.GOARCH == "386" {
+			llvmFn.SetFunctionCallConv(llvm.X86StdcallCallConv)
+		}
 	}
 
 	// External/exported functions may not retain pointer values.
--- a/compiler/symbol.go
+++ b/compiler/symbol.go
@@ -208,7 +208,7 @@ func (c *compilerContext) getFunction(fn *ssa.Function) (llvm.Type, llvm.Value) 
 			// > circumstances, and should not be exposed to source languages.
 			llvmutil.AppendToGlobal(c.mod, "llvm.compiler.used", llvmFn)
 		}
-	case "GetModuleHandleExA", "GetProcAddress", "GetSystemInfo", "GetSystemTimeAsFileTime", "LoadLibraryExW", "QueryUnbiasedInterruptTime", "SetEnvironmentVariableA", "Sleep", "VirtualAlloc":
+	case "GetModuleHandleExA", "GetProcAddress", "GetSystemInfo", "GetSystemTimeAsFileTime", "LoadLibraryExW", "QueryPerformanceCounter", "QueryPerformanceFrequency", "QueryUnbiasedInterruptTime", "SetEnvironmentVariableA", "Sleep", "SystemFunction036", "VirtualAlloc":
 		// On Windows we need to use a special calling convention for some
 		// external calls.
 		if c.GOOS == "windows" && c.GOARCH == "386" {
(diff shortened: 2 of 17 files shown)


aykevl:
Improved the PR slightly: using stdcall on windows/386 so testdata/gc.go works. Importing the time package still crashes the program though.

aykevl:
I *think* all important things are working now!
The only thing that still fails is this:

```
$ GOARCH=386 build/tinygo test -v go/format
=== RUN   TestNode
FAIL    go/format       0.726s
```

...but that's not part of CI (and I'd guess that there are many Linux architectures where some packages don't pass tests).
CI does run some windows/386 tests, so base functionality is being tested.

b0ch3nski:
I think I have some ancient Windows 7 x86 installation around which I could use to test this later this week :slightly_smiling_face:

aykevl:
@b0ch3nski
I tested it on an ARM system (actually a Windows VM inside Asahi Linux running on a Macbook). If the X86 emulation in Windows is good, this should show that it's working as intended. But testing on a real x86 (or even amd64) system would confirm that, so if you want to test that would be great!

BieHDC:
@aykevl Thanks for your work!
For simple testing i just compiled `src/examples/test` and ran it. `GOOS=windows GOARCH=386 ./build/tinygo build -o src/examples/test/test.exe src/examples/test/test.go`
It works on Windows 11 64Bit and Windows Vista 32Bit as they ship new enough dlls and accept the default PE version.
It does not work on Windows XP and lower due to PE version and the dependent dlls. See my [previous work](https://github.com/tinygo-org/tinygo/compare/dev...BieHDC:tinygo:dev) on how i fixed that.

My goal with the 32Bit version was explicitly for the older Windows Versions. Is that also in your interest? The 32/64Bit sliding line for Windows is between XP with 32Bit and 7 with 64Bit, which all are unsupported by now and do not require or benefit from the more secure APIs used right now.

A side question: is there a testsuite we can compile into a binary for extensive testing beyond a print?

compiler/symbol.go
```diff
@@ -208,6 +208,12 @@ func (c *compilerContext) getFunction(fn *ssa.Function) (llvm.Type, llvm.Value)
 			// > circumstances, and should not be exposed to source languages.
 			llvmutil.AppendToGlobal(c.mod, "llvm.compiler.used", llvmFn)
 		}
+	case "GetModuleHandleExA", "GetProcAddress", "GetSystemInfo", "GetSystemTimeAsFileTime", "LoadLibraryExW", "QueryUnbiasedInterruptTime", "SetEnvironmentVariableA", "Sleep", "VirtualAlloc":
```
BieHDC: Can this be done as a catch-all when a kernel32.dll function is being imported? 
Also this fixup might be required for all external dll calls in general. Would need more testing.

b0ch3nski:
I can also confirm that this indeed works on Windows 7 x86 :tada:

And I agree with @BieHDC - if it could be made working also for Windows XP only by tweaking ldflags, let's go for it. I can think of some nostalgia projects running modern software on XP :smile:

BieHDC:
> And I agree with @BieHDC - if it could be made working also for Windows XP only by tweaking ldflags, let's go for it. I can think of some nostalgia projects running modern software on XP 😄

That is also my idea. Would be cool to code stuff for old Windows in Go just for the heck of it.
I personally would go as low as Windows 98, but my ultimate goal would be 95. It depends on OP how low this PR goes. But if OP stays above and this gets merged before christmas, then i know what i will do in the next holidays.

aykevl:
> It does not work on Windows XP and lower due to PE version and the dependent dlls. See my [previous work](https://github.com/tinygo-org/tinygo/compare/dev...BieHDC:tinygo:dev) on how i fixed that.

It looks like there's a version of UCRT for Windows XP. Perhaps that would work?
https://learn.microsoft.com/en-us/cpp/windows/universal-crt-deployment?view=msvc-170#deployment-on-microsoft-windows-xp

I tried using MSVCRT.DLL but didn't get it to run (though binaries do compile). Here is how far I got: https://github.com/tinygo-org/tinygo/compare/dev...windows-386-msvcrt
You could try running such a binary in WinDbg to see why it crashes.

My proposal would be to merge this PR and then develop based on that for those who would like to support even older Windows versions.

> My goal with the 32Bit version was explicitly for the older Windows Versions. Is that also in your interest?

I was working on it because there was demand for it (see #3353). I don't use Windows nor X86 so it was purely out of interest to see whether it could work.



> A side question: is there a testsuite we can compile into a binary for extensive testing beyond a print?

You can run `GOARCH=386 make tinygo-test-fast`, which runs most supported standard library tests. Or you can compile individula packages using `tinygo build -c packagename` and then run the resulting `packagename.test.exe` on the target system. There is no single binary that runs all tests.

compiler/symbol.go
aykevl: I suppose it can, by parsing `//sys` comments in the source code. But just adding the list of names was a lot easier.

Programs that would like to call kernel32.dll functions should use the syscall package, which uses the appropriate calling convention.

aykevl:
Actually I managed to create very simple programs that run with msvcrt.dll, and updated my branch: https://github.com/tinygo-org/tinygo/compare/dev...windows-386-msvcrt

Feel free to use this code for Windows XP support. (Older Windows versions are likely going to be difficult because they aren't really supported by anything anymore, even Windows XP is barely working because it's so old).

aykevl:
Updated to use MSVCRT.DLL on GOARCH=386. This _might_ allow binaries to run on Windows XP, but I haven't tested that.

aykevl:
Sadly, it looks like it doesn't quite work yet on Windows XP (this is with `--major-os-version 4` added to the linker). Unfortunately the error message isn't very helpful either.

Screenshot_winxp-i686_2025-04-08_15:31:44

(Why did I put the test file in an ISO? Because I couldn't get networking to work at all, and apparently file sharing is difficult on Windows XP so I didn't even try).

I don't think I'll be experimenting with this further, but anybody wanting to dig deeper is free to try :)

Anyway, this PR is ready from my POV.

aykevl:
I actually got it to work!

Screenshot_winxp-i686_2025-04-09_10:56:39

compileopts/config.go
```diff
@@ -382,15 +382,25 @@ func (c *Config) LibcCFlags() []string {
...
 		}
+		if c.GOARCH() == "386" {
+			cflags = append(cflags,
+				"-D__MSVCRT_VERSION__=0x700", // Microsoft Visual C++ .NET 2002
+				"-D_WIN32_WINNT=0x0501",      // target Windows XP
+			)
+		} else {
+			cflags = append(cflags,
+				"-D_UCRT",
+				"-D_WIN32_WINNT=0x0a00", // target Windows 10
```
b0ch3nski, line 416: What consequences we would have with `0x0601` targeting Windows 7 here on x64 architecture?

aykevl: Honestly I don't know. It probably works just fine (and if it doesn't we can change that). Note that it didn't change: `"-D_WIN32_WINNT=0x0a00"` was already there, so this doesn't modify behavior for modern (amd64/arm64) systems.

BieHDC:
> (Why did I put the test file in an ISO? Because I couldn't get networking to work at all, and apparently file sharing is difficult on Windows XP so I didn't even try).

I am doing the exact same thing and it is the easiest way for one way transferring.

And nice to see you went down to XP, thanks!

b0ch3nski:
@aykevl Could you please resolve merge conflicts? Apart from it, I think that it's good to go :slightly_smiling_face:

aykevl:
Rebased!

deadprogram:
@aykevl thanks for working on this retro feature. Also thanks @b0ch3nski and @BieHDC for all the reviewing.

Now merging.

PR #4631, opened 2024-11-26
targets: add implementation for Tillitis TKey device

deadprogram:
This PR is to start adding an implementation for the Tillitis TKey device:

https://tillitis.se/products/tkey/

The only thing working is the `blinky1` example.

```shell
$ tinygo flash -size short -target=tkey examples/blinky1
    320       0    2076 |     320    2076
Auto-detected serial port /dev/ttyACM0
Connecting to device on serial port /dev/ttyACM0 ...
Firmware name0:'tk1 ' name1:'mkdf' version:5
UDI: 00010:8:3:00bc614e
Loading app from /tmp/tinygo2087949463/main.bin onto device
```
tkey-blinky1


deadprogram:
UPDATE: I have added the UART support, so now the `examples/serial` also works:

```shell
$ tinygo flash -size short -target=tkey examples/serial
    428       0    2080 |     428    2080
Auto-detected serial port /dev/ttyACM0
Connecting to device on serial port /dev/ttyACM0 ...
Firmware name0:'tk1 ' name1:'mkdf' version:5
UDI: 00010:8:3:00bc614e
Loading app from /tmp/tinygo1965928096/main.bin onto device

$ tinygo monitor -baudrate=62500
hell
hello world!
hello world!
hello world!
hello world!
```

Note that the baud rate is fixed at 62500 and cannot be easily changed. See https://github.com/tillitis/tillitis-key1/pull/288

deadprogram:
FURTHER UPDATE: now added `Pin` interface to touch sensor, so the `examples/button` code works.

```shell
$ tinygo flash -size short -target=tkey examples/button
    312       0    2084 |     312    2084
Auto-detected serial port /dev/ttyACM0
Connecting to device on serial port /dev/ttyACM0 ...
Firmware name0:'tk1 ' name1:'mkdf' version:5
UDI: 00010:8:3:00bc614e
Loading app from /tmp/tinygo874847465/main.bin onto device
```

tkey-touch

deadprogram:
YET FURTHER UPDATE: have now implemented the RNG interface, so the following works:

```shell
$ tinygo flash -size short -target=tkey examples/rand                                                                                                                                                               
  11756      56    2076 |   11812    2132                                                                                                                                                                           
Auto-detected serial port /dev/ttyACM0                                                                                                                                                                              
Connecting to device on serial port /dev/ttyACM0 ...                                                                                                                                                                
Firmware name0:'tk1 ' name1:'mkdf' version:5                                                                                                                                                                        
UDI: 00010:8:3:00bc614e        
Loading app from /tmp/tinygo2643520006/main.bin onto device                                                                                                                                                         

$ tinygo monitor -baudrate=62500                                                                                                                                                                                    
4d0a0ab1fdaea355bc5a4a755770fa085dc12c83eea81043062edd43a4d754fb                                          
f3d709ce9ed4f246ae79730a56bf40f827c84ce15e02e8fec2bcaa037e79799e                                          
c878e6bae7c1283ccae4d9198b7ac7320e4e4482beed24250601a362e5a23c12
7879cb8d916dd0222c1b3d4054fb37dce2be0b33e60719f24c2be7edee9617a6                                          
5c88b3e2de8e10ef9087e418d89e755ac0c6ca68f9ea49cc850f25638e4bdcc3                                          
e8bcf6ab894a9366d4c224eeabe484da0c1c9f77f94bb22243c8204f0ac5c350                                          
3e6707a530606a472c12bf7c6b180e9c786c78f3dff3291e7af962d28f017cb3                                          
4bd4bdc231a4ead00ed726c68e173bdbb07966045acc7b509638d6949f4db8de                                          
0dd34b3f408ffef9be126e649dd0a91a22231f998c7054e054053c7178aab64b
```

deadprogram:
Added commit from the weekend to conform to `machine.Serialer` interface.

The `examples/echo` code is now also working:

```shell
$ tinygo flash -size short -target=tkey examples/echo
   1060       0    2088 |    1060    2088      
Auto-detected serial port /dev/ttyACM0
Connecting to device on serial port /dev/ttyACM0 ...
Firmware name0:'tk1 ' name1:'mkdf' version:5                                                              
UDI: 00010:8:3:00bc614e                                                                                   
Loading app from /tmp/tinygo2398999357/main.bin onto device    

$ tinygo monitor -baudrate=62500                                                                           
Echo console enabled. Type something then press enter:                           
hello from TKey!                                                                                          
You typed: hello from TKey!
```

targets/tkey.ld
```diff
@@ -0,0 +1,78 @@
...
+        KEEP(*(.init))
+        . = ALIGN(4);
+        *(.text)
+        *(.text.*)
+        *(.rodata)
+        *(.rodata.*)
+        . = ALIGN(4);
+    } >RAM
+
+    /* Put the stack at the bottom of RAM, so that the application will
```
mchack-work: The comment says the stack is at bottom of RAM. But it's above the beginning of .text. If it's growing downwards as usual doesn't it corrupt the loaded device app code? Is there some protection for every use that stops this from happening?

If you truly want the stack to occupy the first part of the bottom of RAM you will need either cooperation from firmware to begin load and execution at a higher address (incidentally, this was just what the TKey firmware did previously), or include some kind of jump instruction among the first instructions that is safe to be overwritten by later stack operations, not include the entire .text before the stack. Something like this layout of the device binary:

- _start: jump to crt0
- reserved space for stack
- crt0
- rest of .text

Am I missing something?

deadprogram: The current implementation is valid only for the pre-allocated stack size. So you think it is better to allocate it before the `.text` section? @aykevl any thoughts on this?

As the link script comment mentions, the application will crash on stack overflow instead of silently corrupting memory, so that is no protection, but it does tell you it has happened. Allocation of a larger initial stack is controlled via the `stack-size` param.

mchack-work: I agree it will likely crash when the stack starts to write over the application. Eventually, it's likely to crash. I didn't realize that the intention was to noisily crash by just happening to try to execute some random data from the stack. I thought you just wanted to have the stack stop growing by stopping at the bottom of RAM.

A noisy crash like that is extra interesting in the TKey because that will most likely trigger the unimplemented instruction trap in the CPU and start blinking the status LED in red by the hardware design, only breakable by a power cycle.

A write outside of memory, at least from some early serial number, should also trigger that blink-red-forever hardware behaviour according to documentation but I need to verify that.

If true, that means placing the stack at the true bottom and writing outside of RAM should immediately halt the CPU on the TKey in a visible way.

deadprogram:
More recent commit adds the call for the TKey's built-in `BLAKE2s` function call.

The following modified `echo` program:

```go
package main

import (
	"encoding/hex"
	"machine"
	"time"
)

var (
	uart = machine.Serial
)

func main() {
	// use default settings for UART
	uart.Configure(machine.UARTConfig{})

	input := make([]byte, 64)
	hash := make([]byte, 32)

	i := 0
	for {
		if uart.Buffered() > 0 {
			data, _ := uart.ReadByte()

			switch data {
			case 13:
				// return key
				uart.Write([]byte("\r\n"))
				uart.Write([]byte("You typed: "))
				uart.Write(input[:i])
				uart.Write([]byte("\r\n"))
				if err := calcHash(input[:i], hash); err != nil {
					uart.Write([]byte("Failed to calculate BLAKE2s\r\n"))
					i = 0
					continue
				}

				uart.Write([]byte(hex.EncodeToString(hash)))
				uart.Write([]byte("\r\n"))
				i = 0
			default:
				// just echo the character
				uart.WriteByte(data)
				input[i] = data
				i++
			}
		}
		time.Sleep(10 * time.Millisecond)
	}
}

func calcHash(input []byte, hash []byte) error {
	key := []byte("key")
	err := machine.BLAKE2s(hash, key, input)
	if err != nil {
		return err
	}
	return nil
}
```

Produces the following output:


```shell
$ tinygo flash -size short -target=tkey examples/echo
   1680       0    2076 |    1680    2076
Auto-detected serial port /dev/ttyACM0
Connecting to device on serial port /dev/ttyACM0 ...
Firmware name0:'tk1 ' name1:'mkdf' version:5
UDI: 00010:8:3:00bc614e
Loading app from /tmp/tinygo2868810019/main.bin onto device

$ tinygo monitor -baudrate=62500                                                                              
Echo console enabled. Type something then press enter:           
hello                                                
You typed: hello                         
f3400892fd799ac07396701cabc0c4aab892dc320645ed106c66b6ae5e681181
what a cool hash                      
You typed: what a cool hash                                                                               
e207695eb539abddd8cba0066e6665e022c799e2769dc2d5bf049cfd1426aa29
what a cool hash                                                                                          
You typed: what a cool hash                                                                               
e207695eb539abddd8cba0066e6665e022c799e2769dc2d5bf049cfd1426aa29
```

deadprogram:
I added another commit to implement the runtime `abort()` function by triggering an illegal instruction, which I think is in keeping with your previous comments @mchack-work what do you think?

deadprogram:
Confirmed that the GPIO `Pin` implementation does work on the actual hardware:

tkey-gpio-works

targets/tkey.ld
deadprogram: > If true, that means placing the stack at the true bottom and writing outside of RAM should immediately halt the CPU on the TKey in a visible way.

With heap allocations, this is exactly the current behavior provided by the code in this branch.

For example, this code:

```go
package main

import (
	"runtime"
	"time"
)

func main() {
	ms := runtime.MemStats{}
	runtime.ReadMemStats(&ms)
	println("Starting Heap Used: ", ms.HeapInuse)

	big := make([]byte, 8192)
	for i := 1; i < len(big); i++ {
		big[i] = big[i-1] + 1
	}

	for {
		runtime.ReadMemStats(&ms)
		println("Heap Used: ", ms.HeapInuse)

		b := more()
		big = append(big, b...)
		time.Sleep(10 * time.Millisecond)
	}
}

func more() []byte {
	big := make([]byte, 8192)
	for i := 1; i < len(big); i++ {
		big[i] = big[i-1] + 1
	}
	return big
}
```

When flashed, produces the expected blinking red LED and also outputs panic data via connected serial port:

```shell
$ tinygo flash -size short -target tkey -monitor -baudrate 62500 crashme.go
   2096       0    2076 |    2096    2076
Auto-detected serial port /dev/ttyACM0
Connecting to device on serial port /dev/ttyACM0 ...
Firmware name0:'tk1 ' name1:'mkdf' version:5
UDI: 00010:8:3:00bc614e
Loading app from /tmp/tinygo338584284/main.bin onto device
Starting Heap Used:  0
Heap Used:  8192
Heap Used:  49152
Heap Used:  57344
Heap Used:  65536
panic: runtime error at 0x40000790: out of memory
[tinygo: panic at /home/ron/Development/tinygo/tinygo-122/src/runtime/slice.go:57:14]
```

src/device/tkey/crt0.S
```diff
@@ -0,0 +1,53 @@
...
+	li x22,0
+	li x23,0
+	li x24,0
+	li x25,0
+	li x26,0
+	li x27,0
+	li x28,0
+	li x29,0
+	li x30,0
+	li x31,0
```
aykevl: Why is this needed?
(I see no reason why these registers would need to be cleared).

src/device/tkey/crt0.S
aykevl, line 1: Can you try using src/device/riscv/start.S instead?

src/machine/machine_tkey.go
```diff
@@ -0,0 +1,227 @@
...
+	Bus *tkey.UART_Type
+}
+
+var (
+	DefaultUART = UART0
+	UART0       = &_UART0
+	_UART0      = UART{Bus: tkey.UART}
+)
+
+// Thw TKey UART is fixed at 62500 baud, 8N1.
```
aykevl: Typo

```suggestion
// The TKey UART is fixed at 62500 baud, 8N1.
```

Also, you could consider returning an error if any other configuration than this one is specified by the user.

targets/tkey.json
aykevl, line 1: Since you add a new target (with new LLVM target options and such), can you also add it in builder/builder_test.go?

Also, can you let this file inherit from riscv32? (`"inherits": "riscv32"`). That will avoid some duplication. You can verify the  target is still the same using `tinygo info -json tkey`.

src/device/tkey/crt0.S
```diff
@@ -0,0 +1,53 @@
...
+
+	/* zero-init bss section */
+	la a0, _sbss
+	la a1, _ebss
+	bge a0, a1, end_init_bss
+
+loop_init_bss:
+	sw zero, 0(a0)
+	addi a0, a0, 4
+	blt a0, a1, loop_init_bss
```
aykevl: This shouldn't be needed because bss is zeroed in `runtime.preinit`.

targets/tkey.ld
aykevl: > The comment says the stack is at bottom of RAM. But it's above the beginning of .text. If it's growing downwards as usual doesn't it corrupt the loaded device app code?

Indeed.

I think the simplest fix would be to move `.stack` to just after `.text.init` but before `.text`. The `.text.init` section is not needed anymore after init and can be freely overwritten.

> I didn't realize that the intention was to noisily crash by just happening to try to execute some random data from the stack. I thought you just wanted to have the stack stop growing by stopping at the bottom of RAM.

The comment comes from Cortex-M targets, where we generally place the stack at the bottom of RAM which will generally lead to a crash when reading/writing there.

targets/tkey.ld
aykevl, line 1: Would it be possible to inherit from targets/riscv.ld instead? Just like targets/hifive1b.ld for example.
Or put differently, why did you need to write a completely new linker script?

If it's because there's no flash, you can use `REGION_ALIAS`:

https://sourceware.org/binutils/docs/ld/REGION_005fALIAS.html

deadprogram: > If it's because there's no flash, you can use REGION_ALIAS

That sounds interesting, could you please show me example of how I would use that here?

src/device/tkey/crt0.S
deadprogram: That seems to be working for me as expected, so I can remove the extra file. Thanks!

targets/tkey.ld
aykevl: I think something like this?
```
MEMORY
{
    RAM (rwx)       : ORIGIN = 0x40000000, LENGTH = 0x20000 /* 128 KB */
}

REGION_ALIAS("FLASH_TEXT", RAM);
```

deadprogram: What I am actually trying to achieve is not including that section, since the tkey does not have interrupts.

This is part of why I did not inherit from riscv.ld

deadprogram: @aykevl thanks for all these good suggestions, I am removing files with great pleasure for an upcoming commit.

src/machine/machine_tkey.go
deadprogram: Good idea, done.

src/device/tkey/crt0.S
deadprogram: Removed this file anyhow.

targets/tkey.json
deadprogram: Done, thanks for that suggestion @aykevl

targets/tkey.ld
deadprogram: Done.

deadprogram: Oops forgot to add to `builder_test.go`. Just did that now.

Obviously all of these commits will be squashed before merge :smile_cat:

src/device/tkey/crt0.S
deadprogram: I removed this file based on your other comment.

targets/tkey.ld
aykevl: > What I am actually trying to achieve is not including that section, since the tkey does not have interrupts.

You mean `.text.handleInterruptASM`?
I think the linker will remove it, since it's not referenced anywhere. So in theory it shouldn't be a problem.

targets/tkey.json
```diff
@@ -1,7 +1,7 @@
...
     "cflags": [
         "-mcmodel=medany",
```
aykevl: I think `-mcmodel` is not needed (but am not sure). It should already be the default.

targets/tkey.ld
deadprogram: That appears to be exactly the case.

targets/tkey.json
aykevl, line 1: These options are already the default, so they don't need to be set here:

- default-stack-size
- linker
- rtlib
- libc

targets/tkey.ld
deadprogram: I tested overflow of the stack using the following code:

```go
package main

func fibonacci() func() int {
	first, second := 0, 1
	return func() int {
		ret := first
		first, second = second, first+second
		return ret
	}
}

func main() {
	f := fibonacci()
	for i := 0; i < 10; i++ {
		println(f())
	}
}
```

It appears to work as desired:

```shell
$ tinygo flash -size full -target tkey -monitor -baudrate 62500 fib.go
    158       0       0       0 |     158       0 | (unknown)
    398       0       0       0 |     398       0 | C compiler-rt
     16       0       0       0 |      16       0 | C picolibc
     24       0       0       0 |      24       0 | device/riscv
     22       0       0       0 |      22       0 | machine
     20       0       0       0 |      20       0 | main
    338       0       0       4 |     338       4 | runtime
     16       0       0       0 |      16       0 | runtime/volatile
    992       0       0    2052 |     992    2052 | total
Auto-detected serial port /dev/ttyACM0
Connecting to device on serial port /dev/ttyACM0 ...
Firmware name0:'tk1 ' name1:'mkdf' version:5
UDI: 00010:8:3:00bc614e
Loading app from /tmp/tinygo2162457694/main.bin onto device
0
1
1
2
3
5
8
13
21
34
```

At that point the console output freezes and the red LED starts blinking. So it would appear that both stack and heap overflow will trigger hardware fault, which is the desired behavior.

targets/tkey.ld
deadprogram: OK I think I have made all changes. Any further feedback @aykevl @mchack-work anyone else?

src/device/tkey/tkey.go
```diff
@@ -0,0 +1,139 @@
+//go:build tkey
+
+// Hand written file based on https://github.com/tillitis/tkey-libs/blob/main/include/tkey/tk1_mem.h
```
aykevl, line 3: That's unfortunate :( but not much to do about that.

src/machine/machine_tkey.go
```diff
@@ -0,0 +1,232 @@
...
+
+package machine
+
+import (
+	"device/tkey"
+	"errors"
+	"strconv"
+)
+
+const deviceName = "TKey"
```
aykevl, line 11: This refers to the CPU type. My understanding is that this is a soft-CPU on an FPGA, right? So then I guess this value works as well.

src/machine/machine_tkey.go
```diff
@@ -0,0 +1,232 @@
...
+			pushed = true
+		}
+	}
+
+	switch mode {
+	case PinInputPullup:
+		return !pushed
+	case PinInput, PinInputPulldown:
+		return pushed
+	}
```
aykevl: Oh, I guess that's one way of doing it.
Somewhat unexpected maybe, it doesn't exactly match other systems. But I guess it works.

One thing it doesn't do is return the current value for output pins. This is implemented for many other chips. Not sure how important this is.
See: https://tinygo.org/docs/reference/machine/

src/machine/machine_tkey.go
```diff
@@ -0,0 +1,232 @@
...
+	DefaultUART = UART0
+	UART0       = &_UART0
+	_UART0      = UART{Bus: tkey.UART}
+)
+
+// The TKey UART is fixed at 62500 baud, 8N1.
+func (uart *UART) Configure(config UARTConfig) error {
+	if config.BaudRate != 62500 {
+		return errors.New("uart: only 62500 baud rate is supported")
+	}
```
aykevl: I guess the value 0 would also be acceptable? On most chips, the value 0 means use the default (115200 or 9600).

src/machine/machine_tkey.go
```diff
@@ -0,0 +1,232 @@
...
+
+	return nil
+}
+
+// Write a slice of data bytes to the UART.
+func (uart *UART) Write(data []byte) (n int, err error) {
```
aykevl, line 129: I see you're not using src/machine/uart.go, I guess this is because the implementation here is not interrupt based and therefore `ReadByte` needs to be different?

src/runtime/runtime_tkey_baremetal.go
```diff
@@ -0,0 +1,24 @@
+//go:build tkey && !qemu
+
+package runtime
+
+import "device/riscv"
+
+// ticksToNanoseconds converts ticks (at 18MHz) to nanoseconds.
+func ticksToNanoseconds(ticks timeUnit) int64 {
+	return int64(ticks) * 10000
+}
```
aykevl: The comment here doesn't match the source code!
The code suggests one tick is actually 10_000 nanoseconds, or 10µs.

Same for `nanosecondsToTicks`!

targets/tkey.ld
aykevl: Yes, looks good to me!

aykevl (commented):
Looks good to me, below mostly just minor nits.
The most important one is `ticksToNanoseconds` and `nanosecondsToTicks`, which seem to have an incorrect comment.

LGTM with any possible fixes.

targets/tkey.json
deadprogram: I think you are correct. Removed.

targets/tkey.json
deadprogram: Removed.

src/device/tkey/tkey.go
deadprogram: Indeed.

src/machine/machine_tkey.go
deadprogram: > My understanding is that this is a soft-CPU on an FPGA, right?

That is correct.

src/machine/machine_tkey.go
deadprogram: > One thing it doesn't do is return the current value for output pins. 

Modified to now do that, just to avoid surprises.

src/machine/machine_tkey.go
deadprogram: Good point. Modified to allow for `0` aka default.

src/machine/machine_tkey.go
deadprogram: > I guess this is because the implementation here is not interrupt based

That is correct.

src/runtime/runtime_tkey_baremetal.go
deadprogram: Oops, I forgot to change the comment from the original values I tried. Modified comment to match the current reality.

aykevl (approved):

deadprogram:
Thanks for review and feedback @aykevl @mchack-work now squash/merging.

PR #4633, opened 2024-11-27
runtime: use uint32 for the channel state and select index

aykevl:
This uses uint32 instead of uint64. The reason for this is that uint64 atomic operations aren't universally available (especially on 32-bit architectures). We could also use uintptr, but that seems needlessly complicated: it's unlikely real-world programs will use more than a billion select states (2^30).

(This could also reduce code size slightly, but I didn't test that).


deadprogram (approved):
Seems reasonable to me! Thanks @aykevl

PR #4634, opened 2024-11-27
examples: use default UART settings in echo example

deadprogram:
This PR modifies the `echo` example to use the default UART settings now that they work on all supported platforms.

### DIFF
--- a/src/examples/echo/echo.go
+++ b/src/examples/echo/echo.go
@@ -7,15 +7,13 @@ import (
 	"time"
 )
 
-// change these to test a different UART or pins if available
 var (
 	uart = machine.Serial
-	tx   = machine.UART_TX_PIN
-	rx   = machine.UART_RX_PIN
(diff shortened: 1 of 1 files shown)


src/examples/echo/echo.go
```diff
@@ -10,12 +10,11 @@ import (
 // change these to test a different UART or pins if available
```
b0ch3nski: this comment doesn't make much sense after the change

deadprogram: Should I just remove it entirely?

sago35: I also think it would be better to delete the comments.

deadprogram: OK, done. Thanks for the feedback!

deadprogram:
Now merging, thanks for reviews @b0ch3nski and @sago35

PR #4637, opened 2024-11-29
testdata/corpus: add wazero and its examples to testdata/corpus

### LINKED ISSUE #4636: Consider adding wazero as a testing candidate
There are some real use cases of TinyGo WASI target + [Wazero](https://github.com/tetratelabs/wazero), like https://github.com/wasmvision. I am happy to contribute to a corpus test to ensure each build of TinyGo can run a set of third party Go packages, including wazero. I would appreciate if anyone can point to where I can get started with.

Mossaka:
close #4636

### DIFF
--- a/testdata/corpus.yaml
+++ b/testdata/corpus.yaml
@@ -281,3 +281,7 @@
   version: v2
 - repo: github.com/soypat/mu8
 - repo: github.com/brandondube/pctl
+- repo: github.com/tetratelabs/wazero
+  subdirs:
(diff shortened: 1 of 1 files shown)


aykevl:
This seems like a good idea. @dgryski what do you think?

dgryski:
If they still pass, sure.  I'm all for adding things to our test corpus.

aykevl:
@Mossaka can you verify that the test corpus still passes with these added tests?

PR #4639, opened 2024-11-30
sync: make Cond MT-safe

aykevl:
This actually simplifies the code and avoids a heap allocation in the call to Wait. Instead, it uses the Data field of the task to store information on whether a task was signalled early.

### DIFF
--- a/src/sync/cond.go
+++ b/src/sync/cond.go
@@ -1,19 +1,26 @@
 package sync
 
-import "internal/task"
+import (
+	"internal/task"
+	"unsafe"
+)
+
+// Condition variable.
+// A goroutine that called Wait() can be in one of a few states depending on the
+// Task.Data field:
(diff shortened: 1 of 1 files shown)


aykevl:
Rebased since the first commit was merged in another PR, no code change.

deadprogram:
@dgryski PTAL

dgryski (approved):
I stared at this a long time, specifically the `dataPtr.Swap()` bits.  I'm pretty sure the comments are correct, and I *think* the code is then also correct.  I'm not sure what could convince me more.  Maybe some constants for the `0`/`1`?  So we could have `oldState == awake` or `oldState == aboutToSleep` or something, instead of `!= 0`.  However, I also believe you that this code is correct.  So ...

deadprogram:
Thanks for the addition @aykevl and thank you @dgryski for review. Now merging.

PR #4640, opened 2024-11-30
test: make tests deterministic with -scheduler=threads

aykevl:
This prepares the tests for threading.

### DIFF
--- a/testdata/channel.go
+++ b/testdata/channel.go
@@ -3,6 +3,7 @@ package main
 import (
 	"runtime"
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
@@ -70,11 +71,13 @@ func main() {
 	// Test multi-receiver.
 	ch = make(chan int)
 	wg.Add(3)
-	go fastreceiver(ch)
-	go fastreceiver(ch)
-	go fastreceiver(ch)
+	var result atomic.Uint32
+	go fastreceiveradd(ch, &result)
+	go fastreceiveradd(ch, &result)
+	go fastreceiveradd(ch, &result)
 	slowsender(ch)
 	wg.Wait()
+	println("sum of sums:", result.Load())
 
 	// Test iterator style channel.
 	ch = make(chan int)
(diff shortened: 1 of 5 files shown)


testdata/channel.go
```diff
@@ -244,7 +250,7 @@ func receive(ch <-chan int) {
 func sender(ch chan int) {
 	for i := 1; i <= 8; i++ {
 		if i == 4 {
-			time.Sleep(time.Microsecond)
+			time.Sleep(time.Millisecond)
```
aykevl, line 253: This seems to be long enough to make the test deterministic with the Go toolchain.
But of course it's still racy, if it becomes a problem we can try to find another way using a WaitGroup or something.

testdata/channel.txt
```diff
@@ -12,9 +12,7 @@ received num: 8
 recv from closed channel: 0 false
 complex128: (+7.000000e+000+1.050000e+001i)
 sum of n: 149
-sum: 25
-sum: 29
-sum: 33
+sum of sums: 87
```
aykevl, line 15: These numbers were spread out differently when using threading. Sometimes the right numbers came out, but other times the numbers were balanced differently. Of course that's all correct: the important part is that every number is received by exactly one receiver.

testdata/channel.go
```diff
@@ -88,7 +91,10 @@ func main() {
 	println("sum(100):", sum)
 
 	// Test simple selects.
-	go selectDeadlock() // cannot use waitGroup here - never terminates
+	wg.Add(1)
+	go selectDeadlock()
+	wg.Wait()
```
aykevl, line 96: Interestingly this was easily fixable using a WaitGroup.

deadprogram:
@dgryski PTAL :smile_cat:

testdata/channel.txt
dgryski: The mathematician in me really wants this to switch from adding integers to multiplying unique primes so that we can be sure (due to https://en.wikipedia.org/wiki/Fundamental_theorem_of_arithmetic ) that everything was received exactly once.

dgryski (approved):
LGTM.  We can hunt down other issues as they become apparent.

deadprogram:

testdata/channel.txt
aykevl: Actually that's not a bad idea!
Not sure I'll make that PR, but you're free to do so of course :)

PR #4642, opened 2024-11-30
Prepare channels for parallelism

aykevl:
This is part of #4559 but extracted for easier review.
In theory this should not change the binary size of any existing code, since all the new synchronisation should be no-ops when there are no threads involved.

### DIFF
--- a/src/internal/task/task.go
+++ b/src/internal/task/task.go
@@ -27,7 +27,7 @@ type Task struct {
 }
 
 // DataUint32 returns the Data field as a uint32. The value is only valid after
-// setting it through SetDataUint32.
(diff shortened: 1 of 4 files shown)


aykevl:
Huh, interesting. In some tests the binary size _decreased_ by 40 bytes. Not sure why that's happening but I'll take it. It doesn't increase, so that's good.

aykevl:
Rebased after I merged the "non-atomic atomics" commit into dev.

deadprogram:
@dgryski what do you think?

deadprogram:
@dgryski any feedback on this PR?

deadprogram (approved):
Finally had a chance to take a look at this PR over the weekend. LGTM! :+1:

PR #4644, opened 2024-12-01
mips: fix a bug when scanning the stack

aykevl:
Previously the assembler was changing this code:

    jal tinygo_scanstack
    move $a0, $sp

Into this:

    jal tinygo_scanstack
    nop
    move $a0, $sp

So it was "helpfully" inserting a branch delay slot, even though this was already being taken care of.
Somehow this didn't lead to a crash before, but it does break in the WIP threading branch (https://github.com/tinygo-org/tinygo/pull/4559).

The `.set noreorder` avoids the extra inserted nop, fixing this issue.


deadprogram (approved):
Good catch on that one @aykevl thanks!

PR #4645, opened 2024-12-01
runtime: lock output in print/println

aykevl:
This ensures that calls to print/println happening in different threads are not interleaved. It's a `task.PMutex`, so this should only change things when threading is used.

This matches the Go compiler, which does the same thing: https://godbolt.org/z/na5KzE7en

The locks are not recursive, which means that we need to be careful to not call `print` or `println` inside a runtime.print* implementation, inside putchar (recursively), and inside signal handlers. Making them recursive might be useful to do in the future, but it's not really necessary.

Extracted from #4559 to make reviewing easier.


deadprogram (approved):
Thanks for the improvement to this @aykevl now merging.

PR #4646, opened 2024-12-01
runtime: remove Cond

aykevl:
I don't think this is used anywhere right now, and it would need to be updated to work with multithreading. So instead of fixing it, I think we can remove it.

My original intention was to have something like this that could be used in the machine package, but since this is in the runtime package (and the runtime package imports the machine package on baremetal) it can't actually be used that way.

I checked the TinyGo repo and the drivers repo, and `runtime.Cond` isn't used anywhere except in that one test.

If we want to re-add this feature (for use in the machine package for example!), we can add it again but in a different package (not the runtime package).

Also see: https://github.com/tinygo-org/tinygo/pull/1172

CC @niaow


dgryski (approved):
LGTM

deadprogram:
Thanks for the removal @aykevl and thanks @dgryski for review. Now merging.

PR #4647, opened 2024-12-01
runtime: use exitGoroutine instead of deadlock

aykevl:
This makes the meaning more clear. And while the behavior is currently the same, when we start to use threads we should actually exit the goroutine when calling runtime.Goexit() instead of just leaving it deadlocked.

See: https://github.com/tinygo-org/tinygo/pull/4623#discussion_r1852789402

### DIFF
--- a/src/runtime/panic.go
+++ b/src/runtime/panic.go
@@ -71,9 +71,9 @@ func panicOrGoexit(message interface{}, panicking panicState) {
 		}
 	}
 	if panicking == panicGoexit {
-		// Call to Goexit() instead of a panic.
+		// This is a call to Goexit() instead of a panic.
 		// Exit the goroutine instead of printing a panic message.
-		deadlock()
+		exitGoroutine()
 	}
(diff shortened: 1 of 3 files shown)


aykevl:
Oh cool, this reduces binary size a little! Probably because `trap()` is smaller than `panic("unreachable")`.

aykevl:
But I see there are some AVR failures I need to fix.

dgryski (approved):
LGTM once CI is fixed.

deadprogram:
@aykevl what is the status of this PR?

aykevl:
I still think it's a good idea, to make the code cleaner.
I'll update the PR later to actually make use of `pthread_exit` etc.

dgryski:
I feel like this has been superceeded by @jakebailey's recent scheduler work?

jakebailey:
Some of the platforms still deadlock because they can't truly exit, but fewer do; probably worth a recheck

PR #4648, opened 2024-12-02
fix: allow nintendoswitch target to compile

deadprogram:
This PR is to fix the build errors for the `nintendoswitch` target as mentioned in #2530

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -867,7 +867,7 @@ endif
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=pca10040 -serial=rtt examples/echo
 	@$(MD5SUM) test.hex
-	$(TINYGO) build             -o test.nro -target=nintendoswitch      examples/serial
+	$(TINYGO) build             -o test.nro -target=nintendoswitch      examples/echo2
 	@$(MD5SUM) test.nro
 	$(TINYGO) build -size short -o test.hex -target=pca10040 -opt=0     ./testdata/stdlib.go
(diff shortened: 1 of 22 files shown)


dkegel-fastly (approved):
LGTM

One of these days maybe we should go through all the sources and sort the buildtags into some canonical order...

dkegel-fastly:
Might want to add a note about how timeOffset might be set in the future...?

aykevl (commented):
Looks good to me.

Can you add a smoke test (or modify the existing one) to make sure this keeps working? I think all that is needed is a package that imports the os package (even if it's `import _ "os"`).

deadprogram:
> Can you add a smoke test (or modify the existing one) to make sure this keeps working? 

Could you please point me to exactly where I could add this?

aykevl:
Right now we use the following smoketest:

```makefile
	$(TINYGO) build             -o test.nro -target=nintendoswitch      examples/serial
	@$(MD5SUM) test.nro
```

You could add a new Go file (instead of examples/serial) for the Nintendo Switch, or you can use something like ./testdata/stdlib.go (though that might test more than intended and compiling fmt is a bit slow).

deadprogram:
> You could add a new Go file (instead of examples/serial) for the Nintendo Switch, or you can use something like ./testdata/stdlib.go (though that might test more than intended and compiling fmt is a bit slow).

@aykevl instead of either of those, I changed which file is built for the smoketest to be one that includes the `os` package.

deadprogram:
Thanks for the feedback/comments. Now merging.

PR #4649, opened 2024-12-04
properly handle unix read on directory

noboruma:
On linux systems the `syscall.Read` returns -1 in case the read is being performed on a directory.
This does not play well with the documentation of this function which states that `n` matches the number of bytes count being read. As a result, other code parts may panic, like here:
https://github.com/noboruma/tinygo/blob/fix-os-unix-read/src/os/file.go#L471

The proposition here is to simply set `n` to 0 in case of directory `EISDIR` error.

### DIFF
--- a/src/os/file_anyos.go
+++ b/src/os/file_anyos.go
@@ -1,6 +1,6 @@
 //go:build !baremetal && !js && !wasm_unknown
 
-// Portions copyright 2009 The Go Authors. All rights reserved.
+// Portions copyright 2009-2024 The Go Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
@@ -97,6 +97,11 @@ type unixFileHandle uintptr
 // read and any error encountered. At end of file, Read returns 0, io.EOF.
 func (f unixFileHandle) Read(b []byte) (n int, err error) {
 	n, err = syscall.Read(syscallFd(f), b)
+	// In case of EISDIR, n == -1.
+	// This breaks the assumption that n always represent the number of read bytes.
+	if err == syscall.EISDIR {
+		n = 0
+	}
 	err = handleSyscallError(err)
 	if n == 0 && len(b) > 0 && err == nil {
 		err = io.EOF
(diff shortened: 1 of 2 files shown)


src/os/file_anyos.go
```diff
@@ -97,6 +97,11 @@ type unixFileHandle uintptr
 // read and any error encountered. At end of file, Read returns 0, io.EOF.
 func (f unixFileHandle) Read(b []byte) (n int, err error) {
 	n, err = syscall.Read(syscallFd(f), b)
+	// In case of EISDIR, n == -21.
+	// This breaks the assumption that n always represent the number of read bytes.
+	if err == syscall.EISDIR {
+		n = 0
+	}
```
leongross: I am confused by this change. I tried [1] to recreate your described behavior on my linux machine but could not observe, that the read syscall would return `-EISDIR` for the bytes read when this error is encountered.

[1]: https://gist.github.com/leongross/790669a0e83369970f9e7b9bdff30bda

aykevl: ~I suspect `n` should be set to 0 for any error.~ actually probably not, there's also EOF.
This might actually be a bug in the Syscall function (implemented in the compiler).

noboruma: I am using llvm-18 on debian 12 (kernel 6.1)
@leongross on my machine running your C code shows `bytes_read = -1`, with both `gcc 12.2` and `clang-18`.

noboruma: I have updated the commit to mention -1 instead of -21.
Fix stays the same, in case of `EISDIR` we can force `n = 0`

leongross: I think the proper thing to do here is to check the `err` returned in https://github.com/noboruma/tinygo/blob/fix-os-unix-read/src/os/file.go#L471 and then abort the operation. If there really is an error, we should abort anyway, so the value of `n` should not be read anyway.

noboruma: @leongross we can, this was actually the first fix I did. 
However this code is a copy-paste from official Golang source (GC), so any modification here means we are forking away from std. In std library they handle the `n` issue at a lower layer, as proposed here.

aykevl (approved):

deadprogram:
Thank you @noboruma for the code changes and to @leongross and @aykevl for review. Now merging.

PR #4650, opened 2024-12-06
sync: implement futex-based Mutex

aykevl:
This is largely based on this paper by Ulrich Drepper: https://dept-info.labri.fr/~denis/Enseignement/2008-IR/Articles/01-futex.pdf

See the comment in the source code for details.

Part of #4559.

### DIFF
--- a/src/internal/task/mutex.go
+++ b/src/internal/task/mutex.go
@@ -1,0 +1,69 @@
+package task
+
+// Futex-based mutex.
+// This is largely based on the paper "Futexes are Tricky" by Ulrich Drepper.
+// It describes a few ways to implement mutexes using a futex, and how some
+// seemingly-obvious implementations don't exactly work as intended.
+// Unfortunately, Go atomic operations work slightly differently so we can't
+// copy the algorithm verbatim.
+//
+// The implementation works like this. The futex can have 3 different values,
+// depending on the state:
+//
+//   - 0: the futex is currently unlocked.
+//   - 1: the futex is locked, but is uncontended. There is one special case: if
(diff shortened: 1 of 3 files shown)


aykevl:
Hmm, there's a bit of a binary size increase. I can maybe look into that, it would be nice to avoid.

deadprogram:
> Hmm, there's a bit of a binary size increase. I can maybe look into that, it would be nice to avoid.

@aykevl did you have a chance to do that yet?

aykevl:
Yes, I still want to do this. Probably by keeping the old implementation for the non-parallel case.

src/internal/task/mutex.go
```diff
@@ -0,0 +1,69 @@
...
+func (m *Mutex) Lock() {
+	// Fast path: try to take an uncontended lock.
+	if m.futex.CompareAndSwap(0, 1) {
+		// We obtained the mutex.
+		return
+	}
+
+	// The futex is contended, so we enter the contended loop.
+	// If we manage to change the futex from 0 to 2, we managed to take the
+	// look. Else, we have to wait until a call to Unlock unlocks this mutex.
```
leongross, line 38: nit
```suggestion
	// lock. Else, we have to wait until a call to Unlock unlocks this mutex.
```

leongross (approved):
Lgtm

src/internal/task/mutex.go
aykevl: Thanks, fixed in #4559!

aykevl:
Replaced by #4695.

PR #4651, opened 2024-12-08
feature: modify i2s interface/implementation to better match spec

deadprogram:
This PR modifies the i2s interface/implementation on SAMD21 to better match the i2s specification.

Based on some learning from implementation for rp2040 pio in https://github.com/tinygo-org/pio/pull/14

### DIFF
--- a/src/machine/i2s.go
+++ b/src/machine/i2s.go
@@ -1,4 +1,4 @@
-//go:build sam
+//go:build sam && atsamd21
 
 // This is the definition for I2S bus functions.
 // Actual implementations if available for any given hardware
@@ -9,6 +9,22 @@
 
 package machine
 
+import "errors"
+
+// If you are getting a compile error on this line please check to see you've
+// correctly implemented the methods on the I2S type. They must match
+// the interface method signatures type to type perfectly.
+// If not implementing the I2S type please remove your target from the build tags
+// at the top of this file.
+var _ interface {
+	SetSampleFrequency(freq uint32) error
+	ReadMono(b []uint16) (int, error)
+	ReadStereo(b []uint32) (int, error)
+	WriteMono(b []uint16) (int, error)
+	WriteStereo(b []uint32) (int, error)
+	Enable(enabled bool)
+} = (*I2S)(nil)
+
 type I2SMode uint8
 type I2SStandard uint8
 type I2SClockSource uint8
@@ -18,6 +34,7 @@ const (
 	I2SModeSource I2SMode = iota
 	I2SModeReceiver
 	I2SModePDM
+	I2SModeSourceReceiver
 )
 
 const (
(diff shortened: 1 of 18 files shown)


deadprogram:
Note that this PR includes #4652 to pass CI. So that one should be merged first, and this PR rebased.

deadprogram:
Now rebased and ready for review.

deadprogram:
@soypat @sago35 or anyone else interested, PTAL.

src/machine/i2s.go
```diff
@@ -9,6 +9,23 @@
...
+// the interface method signatures type to type perfectly.
+// If not implementing the I2S type please remove your target from the build tags
+// at the top of this file.
+var _ interface {
+	SetSampleFrequency(freq uint32) error
+	ReadMono(b []uint16) (int, error)
+	ReadStereo(b []uint32) (int, error)
+	WriteMono(b []uint16) (int, error)
+	WriteStereo(b []uint32) (int, error)
+	Paused(disabled bool)
```
soypat: What does Paused do?

deadprogram: Maybe it should be called `Enabled()`? I took that name from here https://github.com/tinygo-org/pio/blob/main/rp2-pio/piolib/i2s.go#L97 but now that you mention it I think we should change them both.

soypat: Maybe we can omit it for now. I'm not sure what fraction of i2s hardware implements the pause functionality. Same goes for Stop... I'm not sure what Stop does exactly, especially seeing Paused _seems_ to do the same thing. I really didn't think much of the implementation through, just copy pasted the C SDK code  :sweat_smile:

soypat: I'm also thinking Stop and Paused could be implemented by a higher-level software layer which splits up incoming buffers to allow interrupting read/writes whereas hardware read/writes would not allow pausing and stopping.

deadprogram: I think we should have an `Enable()` function. But also remove `Paused()` and `Stop()` for the reasons you just expressed. What do you think?

soypat: We can always add it later on if it is really needed, but I've got a feeling from what I can tell it may be an early abstraction to add it at the hardware level. Can we leave them all out for now?

deadprogram: YAGNI is usually the correct choice. OK I will remove.

deadprogram:
Changed based on feedback from @soypat thank you!

deadprogram:
Any final feedback before merge?

sago35 (approved):
`src/machine/board_grandcentral-m4.go` and `src/machine/board_wioterminal.go` are for atsamd51, so they might not need to be changed in this PR. Everything else looks good.

deadprogram:
@sago35 I am hoping to add the i2s support for atsamd51 in the near future.

deadprogram:

PR #4655, opened 2024-12-12
Implement RP2350 initialization clocks, uart, and (partial) USB support

cibomahto:
This patch enables the clocks, uart, and partial USB support for the 2350.

I left the default console as 'uart' because if it is set to USB, using 'fmt' causes a panic. USB transfers work fine if only a single byte is transferred at a time.

Another issue is that the clock speed for RP2350 is set to 125MHz, because there were a number of places where this was hardcoded when setting up peripherals. This should at least be increased to 150MHz, perhaps as a new constant in machine_rp2_2xxxx.


src/runtime/runtime_rp2350.go
```diff
@@ -73,6 +74,10 @@ func machineInit()
 
 func init() {
 	machineInit()
+
+	machine.InitSerial()
```
deadprogram: The order of these commands are incorrect.  They should be the same as the rp2040:

```go
	cdc.EnableUSBCDC()
	machine.USBDev.Configure(machine.UARTConfig{})
	machine.InitSerial()
```

deadprogram: With that change the USB seems to work as expected:

```shell
$ tinygo flash -size full -target pico2 -monitor examples/serial                                                                                                                                                    
      0       6       3       1 |       9       4 | (padding)                                                                                                                                                       
    529       0       8      34 |     537      42 | (unknown)                                                                                                                                                       
    500       0       0       0 |     500       0 | C compiler-rt                                                                                                                                                   
      0     248       0       0 |     248       0 | C interrupt vector                                                                                                                                              
     76       0       0       0 |      76       0 | C picolibc                                                                                                                                                      
      0      16       0       0 |      16       0 | Go types
     92       0       0       0 |      92       0 | device/arm
     28       0       0       0 |      28       0 | device/rp
     14       0       0       0 |      14       0 | internal/binary
    284      48       4       4 |     336       8 | internal/task
   1214     143       0     314 |    1357     314 | machine
    400       0       0       7 |     400       7 | machine/usb/cdc
      0       0      93       0 |      93      93 | machine/usb/descriptor
     28      12       0       0 |      40       0 | main
   2474     598       0     752 |    3072     752 | runtime
    458       0       0       0 |     458       0 | runtime/volatile
   6097    1071     108    3160 |    7276    3268 | total
hello world!
hello world!
...
```

I also tested the `echo` example, and it also worked,

deadprogram:
Once you make the fix I suggest @cibomahto you should rebase this branch against the `rp2350-add` branch, since it is now rebased against the latest `dev`. If you do that, this PR should pass all CI tests.

deadprogram:
@cibomahto I had to rebuild the cached LLVM because of reasons unrelated to this PR.

All tests are now passing! :partying_face:

deadprogram:
@cibomahto looks like you need to run `go fmt` on `src/machine/machine_rp2350_rom.go`

cibomahto:
@deadprogram Everything passes again

deadprogram:
I just confirmed that the bootloader changes work as expected @cibomahto great work!

@soypat IMO this PR can now be merged into #4459 and that PR can be set to ready for review

soypat (commented):
We can merge this but I can see a lot of things I'm going to want to unexport before next release

deadprogram:
> but I can see a lot of things I'm going to want to unexport before next release

@soypat do you mean functions that you think should not be exported from the package?

soypat:
@deadprogram yes. A cursory glance showed a lot of USB and Clock exported functions which did not need to be exported or could better be hid from package scope behind methods on a type

deadprogram:
> A cursory glance showed a lot of USB and Clock exported functions which did not need to be exported or could better be hid from package scope behind methods on a type

That is probably the case with the current rp2040 implementation as well. In any case we can address those things separately.

deadprogram:
Now rebasing this branch into the `rp2350-add` branch for further development. Thank you very much @cibomahto and @soypat

cibomahto:
> @deadprogram yes. A cursory glance showed a lot of USB and Clock exported functions which did not need to be exported or could better be hid from package scope behind methods on a type

Yes, apologies that I'm not fluent enough in Go to know the right way to do this yet :-)

cibomahto:
Another thing for whatever list these go on, the USB driver both chips are almost exactly the same*, but the CMSIS register names are slightly different between the two. It would probably be best to merge them back together if possible.

* Except for an extra init in the rp2350, and the rp2350 not needing the errata from the rp2040

PR #4660, opened 2024-12-15
feat(nrf52xxx/adc|spi): adjustable resolution, improve SPI

milkpirate:
# What I did
## SPI
- Create `SPIMode` type
- Make SPI implement `io.Reader`, `io.Writer`
- Update formatting to align with other switch-cases
- Update SPI documentation reference

## ADC
- Change ADC methods to pointer receivers
- Make ADC resolution adjustable
- Remove nonsensical calculations (at least to me they seemed like, please correct me if I'm wrong)
- Update ADC documentation reference

# Open questions
- Will make sense to change the signature of `ADC.Get()` to return an error? Since it will change the interface and might break things.
- How to run the added tests?

### DIFF
--- a/src/machine/machine_nrf52xxx.go
+++ b/src/machine/machine_nrf52xxx.go
@@ -52,8 +52,23 @@ func (a *ADC) Configure(config ADCConfig) {
 		// TODO: return an error, will that interfere with any interfaced if one will be?
 	}
 
-	// Source resistance, according to table 89 on page 364 of the nrf52832 datasheet.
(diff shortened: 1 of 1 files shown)


ysoldak (changes requested):
I'd prefer this PR to be split into two: separately ADC and SPI changes.

You also need to justify the testify (:D) addition -- we don't have tests in machine package, probably better to make a separate PR for that too.

milkpirate:
in favor of other PRs:
- https://github.com/tinygo-org/tinygo/pull/4699
- https://github.com/tinygo-org/tinygo/pull/4700
- https://github.com/tinygo-org/tinygo/pull/4701

PR #4661, opened 2024-12-17
Binary size report in HTML

aykevl:
This adds a `-size-report` flag that writes a HTML file with more detailed information on the binary size. This is similar to `-size=full` but also adds file level size usage.
Here is an example, with "C compiler-rt" expanded:
Screenshot_20241217_114048

Here you can see that 28 bytes go to `memset`, 92 bytes go to the count-leading-zeroes builtin (clzsi3), and the remaining bytes (92+14 = 106) go to an unsigned 32-bit divide builtin. This makes sense: the chip I used for testing is the nrf51 (in the microbit) which doesn't support clz or division so these need to be done in software.

If you expand the runtime package, you can also see that around 1kB of binary size is taken up by the GC, the rest is for various other runtime features like initialization and the scheduler.

I intend to add more to this size report in the future. Most importantly, a way to answer the question "why does my binary take up so much space, I didn't import that many packages!". Basically, a way to trace from the main package where all the binary size goes and how much a given package is responsible for the binary size of the program. While this could in theory all be done on the command line, doing this in HTML/CSS/JS makes the UX a whole lot better and more intuitive.

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -925,15 +925,16 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 			}
 
 			// Print code size if requested.
-			if config.Options.PrintSizes == "short" || config.Options.PrintSizes == "full" {
+			if config.Options.PrintSizes != "" {
 				sizes, err := loadProgramSize(result.Executable, result.PackagePathMap)
 				if err != nil {
 					return err
 				}
-				if config.Options.PrintSizes == "short" {
+				switch config.Options.PrintSizes {
+				case "short":
 					fmt.Printf("   code    data     bss |   flash     ram\n")
 					fmt.Printf("%7d %7d %7d | %7d %7d\n", sizes.Code+sizes.ROData, sizes.Data, sizes.BSS, sizes.Flash(), sizes.RAM())
-				} else {
+				case "full":
 					if !config.Debug() {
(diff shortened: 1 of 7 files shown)


deadprogram:
Any particular reason you did not just add another option to `-size=report`?

aykevl:
What would you propose as a command line flag?
Note that we'd need to specify the file name somehow, or pick a default one to store in `/tmp`.

deadprogram:
I misunderstood that the usage was `-size-report=<filename>` . Let me consider it...

deadprogram:
How about adding a new flag `size-report-file` with a default value of `size-report.html`.

In addition, add the `report` option to the `-size` flag.

That way people can use `-size=report` and it basically does the same thing as other `-size` flag options, but CI can override the output location for the report if they need to for some reason.

dgryski:
Just to bikeshed a bit here, I think most Go tooling would use `-size=html` here and either write to a "known" output file or stdout.

deadprogram:
I like @dgryski suggestion.

aykevl:
Updated the PR with `-size=html`.
It will now also print the filename when writing the file.

deadprogram:
@aykevl please note CI failure.

aykevl:
Oops, yes, fixed.

deadprogram:
Thanks for the changes to this feature @aykevl it works really well! Now merging.

PR #4662, opened 2024-12-17
fix: ensure that tkey target has stubs for runtime/interrupt package

deadprogram:
This PR is add build tags to ensure that tkey target has stubs for `runtime/interrupt` package. That way if that package is included by some other package, it does not try to bring in any standard RISC-V IRQ implementation.

Discovered when working on https://github.com/hybridgroup/tinygo-tkey/pull/2 with solution suggested by @aykevl


deadprogram:
@aykevl looks ok to you?

deadprogram:
Merging since @aykevl had already told me what to do about this.

PR #4664, opened 2024-12-17
feature: make i2c implementation shared for rp2040/rp2350

deadprogram:
This PR makes the i2c implementation shared between the `rp2040` and the `rp2350`. 

It works on the real hardware, just tested on my Pico2 with an ssd1306 display. :smiley_cat:
PXL_20241217_194237882


sago35 (approved):

deadprogram:
Thanks for reviewing @sago35 now merging.

PR #4665, opened 2024-12-18
feature: make SPI implementation shared for rp2040/rp2350

deadprogram:
This PR makes the SPI implementation shared between the rp2040 and the rp2350.

Tested on real hardware on my Pico2 with an mcp3008 ADC.

I simplified the `setFormat()` function becuase the additional param was not actually doing anything (`rp.XIP_SSI_CTRLR0_SPI_FRF_STD = 0`)


sago35 (approved):

deadprogram:

PR #4666, opened 2024-12-18
feature: make RNG implementation shared for rp2040/rp2350

deadprogram:
This PR makes the RNG implementation shared between the rp2040 and the rp2350.

Worth noting that it is not a true RNG so is not enabled by default on rp2040 or rp2350.

I did test on real hardware with my usual insecure hackaround. Works as expected.

### DIFF
--- a/src/machine/machine_rp2_rng.go
+++ b/src/machine/machine_rp2_rng.go
@@ -1,4 +1,4 @@
-//go:build rp2040
+//go:build rp2040 || rp2350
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Any feedback on this PR? It is pretty minimal change based on existing functionality.

sago35 (approved):

sago35:
LGTM

deadprogram:
Thanks @sago35 for review now merging.

PR #4668, opened 2024-12-19
reflect: fix incorrect comment on elemType

aykevl:
PR that introduced this: https://github.com/tinygo-org/tinygo/pull/4543

`elemType` has a different set of fields than `interfaceType`, but the comment says something else.


deadprogram (approved):
:scroll:

Issue #4673, opened 2024-12-30
rp2040 / rp2350 xoscFreq setting is nonfuctional


cibomahto:
The RP2040 and 2350 board files all specify a parameter for the external crystal frequency, [xoscFreq](https://github.com/tinygo-org/tinygo/blob/release/src/machine/board_pico.go#L38). This appears to be so that a different external crystal could be used on a particular board.

Unfortunately, the [PLL configurations for the system and USB clock](https://github.com/tinygo-org/tinygo/blob/52983794d702af7a00833ae12b0d2e7175e46017/src/machine/machine_rp2_clocks.go#L149) are hard-coded to assume a 12MHz input clock, so this setting doesn't actually work. Finding an optimal setting for the PLL [is complicated](https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_clocks/scripts/vcocalc.py), and it's probably best not to do this on the fly during the startup code.

I can think of two straightforward solutions:
1. Remove the xoscFreq setting. The majority of RP2040 or RP2350 designs should be expected to use a 12MHz crystal.
2. Add the two PLL configurations to the board (or target) configuration, so that a custom crystal frequency can be used correctly.

I think option 1 better, unless  someone has a board with a nonstandard clock that they need to support. I couldn't find examples of other machines/boards that support a similar feature. If either of these sound reasonable, I'd be happy to make a pull request.

cc @soypat @deadprogram

cibomahto:
I created a [clock configuration structure](https://github.com/tinygo-org/tinygo/pull/4674/commits/586c88e4cdbaf4fbad178f3ecca171b53ba98a6b#diff-4babe1e3aead6c34db2f8de3a7d1680b3f073b142370939eebf3d9397f5666a2), made a default in [machine_rp2_2xxx.go](https://github.com/tinygo-org/tinygo/pull/4674/commits/586c88e4cdbaf4fbad178f3ecca171b53ba98a6b#diff-f4c70a0146ba047fe88ad88a2bd140a6d3001ef7aa845b7b69fb27931ada2532), then modified the [board files](https://github.com/tinygo-org/tinygo/pull/4674/commits/586c88e4cdbaf4fbad178f3ecca171b53ba98a6b#diff-c40f5ef7da9ed2e0faab5e546d8ca6b875a50f55af44223617dd9d9e0b87dd0f) to create a new object based on the default.

[...]

PR #4674, opened 2024-12-31
[WIP] Separate system clock config for rp2xxx devices

cibomahto:
Here is a first pass at separating the PLL configuration options for the rp2xxx devices, to allow the rp2350 to be configured for 150MHz. I tested (using a logic analyzer) that the UART, USB, I2C, and SPI generate roughly correct speeds.

I moved the configuration into the machine files, but it could be nice to allow a board to override them. If you have a suggestion for how to do that, please let me know.

I know of three issues that should be fixed before submitting the pull request:
* I2C is slow (90khz instead of 100khz), but this appears to be an existing issue and is similar on both parts.
* The 150MHz clock creates a non-integer 6.66ns period, which makes the PWM output inaccurate on the rp2350 
* Currently, the board files have a separate, nonfuctional configuration for the on-board crystal. I created a separate issue for this: https://github.com/tinygo-org/tinygo/issues/4673 , but this touches on the same code.

cc @soypat @deadprogram

### DIFF
--- a/src/machine/machine_rp2_2040.go
+++ b/src/machine/machine_rp2_2040.go
@@ -113,6 +113,19 @@ const (
 	fnXIP pinFunc = 0
 )
 
+// System clock configuration
+const (
+	pllSysFreq uint32 = 125*MHz
+	pllSysVcoFreq = 1500*MHz
+	pllSysPostDiv1 = 6
+	pllSysPostDiv2 = 2
+
+	pllUSBFreq uint32 = 48*MHz
+	pllUSBVcoFreq = 480*MHz
+	pllUSBPostDiv1 = 5
+	pllUSBPostDiv2 = 2
+)
+
 // Configure configures the gpio pin as per mode.
 func (p Pin) Configure(config PinConfig) {
 	if p == NoPin {
@@ -184,7 +197,7 @@ func (clks *clocksType) initRTC() {
 	crtc := clks.clock(clkRTC)
 	crtc.configure(0, // No GLMUX
 		rp.CLOCKS_CLK_RTC_CTRL_AUXSRC_CLKSRC_PLL_USB,
-		48*MHz,
+		pllUSBFreq,
 		46875)
 }
 
--- a/src/machine/machine_rp2_2040.go
+++ b/src/machine/machine_rp2_2040.go
@@ -113,18 +113,20 @@ const (
 	fnXIP pinFunc = 0
 )
 
-// System clock configuration
-const (
-	pllSysFreq     uint32 = 125 * MHz
-	pllSysVcoFreq         = 1500 * MHz
-	pllSysPostDiv1        = 6
-	pllSysPostDiv2        = 2
+// Clock frequency and PLL configuration
+var defaultSystemClockCfg = systemClockCfg{
+	xoscFreq: 12,
 
-	pllUSBFreq     uint32 = 48 * MHz
-	pllUSBVcoFreq         = 480 * MHz
-	pllUSBPostDiv1        = 5
-	pllUSBPostDiv2        = 2
-)
+	pllSysFreq:     125 * MHz,
+	pllSysVcoFreq:  1500 * MHz,
+	pllSysPostDiv1: 6,
+	pllSysPostDiv2: 2,
+
+	pllUSBFreq:     48 * MHz,
+	pllUSBVcoFreq:  480 * MHz,
+	pllUSBPostDiv1: 5,
+	pllUSBPostDiv2: 2,
+}
 
 // Configure configures the gpio pin as per mode.
 func (p Pin) Configure(config PinConfig) {
@@ -197,7 +199,7 @@ func (clks *clocksType) initRTC() {
 	crtc := clks.clock(clkRTC)
 	crtc.configure(0, // No GLMUX
 		rp.CLOCKS_CLK_RTC_CTRL_AUXSRC_CLKSRC_PLL_USB,
-		pllUSBFreq,
+		clockCfg.pllUSBFreq,
 		46875)
 }
 
--- a/src/machine/machine_rp2_2040.go
+++ b/src/machine/machine_rp2_2040.go
@@ -201,8 +201,6 @@ func (clks *clocksType) initRTC() {
 		46875)
 }
 
-func (clks *clocksType) initTicks() {} // No ticks on RP2040
-
 // startTick starts the watchdog tick.
 // cycles needs to be a divider that when applied to the xosc input,
 // produces a 1MHz clock. So if the xosc frequency is 12MHz,
--- a/src/machine/machine_rp2_2040.go
+++ b/src/machine/machine_rp2_2040.go
@@ -115,15 +115,15 @@ const (
 
 // System clock configuration
 const (
-	pllSysFreq uint32 = 125*MHz
-	pllSysVcoFreq = 1500*MHz
-	pllSysPostDiv1 = 6
-	pllSysPostDiv2 = 2
+	pllSysFreq     uint32 = 125 * MHz
+	pllSysVcoFreq         = 1500 * MHz
+	pllSysPostDiv1        = 6
+	pllSysPostDiv2        = 2
 
-	pllUSBFreq uint32 = 48*MHz
-	pllUSBVcoFreq = 480*MHz
-	pllUSBPostDiv1 = 5
-	pllUSBPostDiv2 = 2
+	pllUSBFreq     uint32 = 48 * MHz
+	pllUSBVcoFreq         = 480 * MHz
+	pllUSBPostDiv1        = 5
+	pllUSBPostDiv2        = 2
 )
 
 // Configure configures the gpio pin as per mode.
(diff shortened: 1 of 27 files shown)


deadprogram:
Thank you for the PR @cibomahto 

Can you please run `go fmt` on the changed files?

```shell
/bin/sh: 1: node: not found
Unformatted:
  src/machine/machine_rp2_2040.go
  src/machine/machine_rp2_2350.go
make: *** [GNUmakefile:185: fmt-check] Error 1
```

src/machine/machine_rp2_2040.go
```diff
@@ -113,6 +113,19 @@ const (
 	fnXIP pinFunc = 0
 )
 
+// System clock configuration
+const (
+	pllSysFreq     uint32 = 125 * MHz
```
eliasnaur, line 121: The rp2040 datasheet says that its cores are capable of 133 MHz. A cursory glance at the datasheet didn't mention any downside to that speed, so why is TinyGo running at 125 MHz? Perhaps it's because of non-integer periods that you mention in the PR description for rp2350? If so, this PR might consider rp2040 at 133 MHz for the same reasons you chose 150 MHz for rp2350.

Failing all that, a comment would be nice.

cibomahto: That's a good point, I don't know why it was running at 125 either, but this seems plausible.

soypat:
Changing out the xoscFreq constant for a package level variable feels off. I also don't understand what we gained from switching out the constant for a xoscFreq field on clockCfg. In any case the constants could remain in each board-level file and the variable be set in a single package level file.

cibomahto:
> Changing out the xoscFreq constant for a package level variable feels off. I also don't understand what we gained from switching out the constant for a xoscFreq field on clockCfg. In any case the constants could remain in each board-level file and the variable be set in a single package level file.

As is, the xoscFreq constant doesn't work- you need to configure the PLLs to match the xoscFreq setting, and they were hard-coded in the clock configuration routine. Otherwise, you won't get the correct CPU clock speed, or the PLL might not actually start.

I guess some other options are:
* Port the PLL configuration solver to go, and run it during the configuration routine to determine the needed settings. Usually this isn't done because it is a little heavy for a boot routine, and there is some nuance about which configuration is best for a given design.
* Make a table of known PLL configurations for the given xoscFreq, and panic if the correct one isn't on the list.
* Drop xoscFreq, it's unlikely that any boards will use a different value than the default. This also implies that you can only run the RP2040 or RP2350 at a single clock speed, but maybe that's good enough.
* Copy all the PLL and clock settings into each board as constants, as you suggested, but then there is a lot of duplication.

In C-style apis, it's a common pattern to have default tables that you can pass to init routines, but I haven't been able to find something similar in TinyGo. The closest so far is for instance the SPI or I2C configuration objects ([for example](https://github.com/tinygo-org/tinygo/blob/52983794d702af7a00833ae12b0d2e7175e46017/src/machine/machine_atsamd21.go#L680)), but in this case the defaults are 'secret', and [applied when fields are left at 0](https://github.com/tinygo-org/tinygo/blob/52983794d702af7a00833ae12b0d2e7175e46017/src/machine/machine_atsamd21.go#L706), which seems a little weird to me.

eliasnaur:
Gentle ping from a nobody who looks forward to run my pico2 at 150MHz :)

It seems to me dropping `xoscFreq` (or its configurability) is the shortest path forward, given that [its value is fixed](https://github.com/tinygo-org/tinygo/issues/4673) anyway. Then another PR can deal with any future board that uses a different crystal.

soypat (changes requested):
My main gripe is preserving the clarity of the existing code. From what I can tell the actual changes to the logic of what is happening are:
- pllSysFreq, dependent on rp2350 vs rp2040
- pllSysPostDiv1, dependent on rp2350 vs rp2040

The way we express hardware/board dependent constants in tinygo are using constants guarded by [**build tags**](https://www.digitalocean.com/community/tutorials/customizing-go-binaries-with-build-tags). This gives us clarity of hardware values which are constant throughout the life of the program. Even in the case where their conceptual value may vary we can still choose to express their default/startup value as a constant.

To implement these constants I'd suggest creating two constants inside src/machine/machine_rp2_2040.go and src/machine/machine_rp2_2350.go to express this hardware dependent variation.

```go
// src/machine/machine_rp2_2040.go
const (
    pllSysFreq     = 125 * MHz
    pllSysPostDiv1 = 6
)

// src/machine/machine_rp2_2350.go
const (
    pllSysFreq     = 150 * MHz
    pllSysPostDiv1 = 5
)
```

And then use these values where they correspond.

Let me know if it is still unclear and I can try my hand at adding a commit here

cibomahto:
@soypat That sounds good. Is there a mechanism that allows constants defined in machine-level files to be overridden by a board-level tag?

Two notes:
* Those values for 'pllSysFreq' and 'pllSysPostDiv1' assume xoscFreq is 12, so xoscFreq should be removed from the board files, and moved to machine_rp2_clocks.
* The 133 and 150MHz clock speeds cause don't divide into an integer ns clock frequency, so moving to these frequencies will make at least the PWM peripheral timings inaccurate.

deadprogram:
>  Is there a mechanism that allows constants defined in machine-level files to be overridden by a board-level tag?

No, however those could need to be `var` instead of `const`, which would allow that to be changed in a board file.

> Those values for 'pllSysFreq' and 'pllSysPostDiv1' assume xoscFreq is 12, so xoscFreq should be removed from the board files, and moved to machine_rp2_clocks.

Seems reasonable.

> The 133 and 150MHz clock speeds cause don't divide into an integer ns clock frequency, so moving to these frequencies will make at least the PWM peripheral timings inaccurate.

That is however possibly a problem, since things like servos would malfunction.

eliasnaur:
> > Is there a mechanism that allows constants defined in machine-level files to be overridden by a board-level tag?
> 
> No, however those could need to be `var` instead of `const`, which would allow that to be changed in a board file.
> 

Note that some code, such as [`delay.Sleep`](https://github.com/tinygo-org/drivers/blob/019bbbe5fb66ff6ecd50949520253a50ecc72733/delay/sleep.go#L42), depends on `machine.CPUFrequency` being constant.

> 
> > The 133 and 150MHz clock speeds cause don't divide into an integer ns clock frequency, so moving to these frequencies will make at least the PWM peripheral timings inaccurate.
> 
> That is however possibly a problem, since things like servos would malfunction.

Is there no way to work around these malfunctions? Does the frequency really have to be runtime configurable to avoid them?

cibomahto:
> > > Is there a mechanism that allows constants defined in machine-level files to be overridden by a board-level tag?
> > 
> > 
> > No, however those could need to be `var` instead of `const`, which would allow that to be changed in a board file.
> 
> Note that some code, such as [`delay.Sleep`](https://github.com/tinygo-org/drivers/blob/019bbbe5fb66ff6ecd50949520253a50ecc72733/delay/sleep.go#L42), depends on `machine.CPUFrequency` being constant.

That's good insight! I'm confused by that code and comment. It seems like the function will work even if CPUFrequency() isn't constant, but just be somewhat inaccurate for short delays. It also seems confusing that it expects a function call to evaluate to a constant.

> > > The 133 and 150MHz clock speeds cause don't divide into an integer ns clock frequency, so moving to these frequencies will make at least the PWM peripheral timings inaccurate.
> > 
> > 
> > That is however possibly a problem, since things like servos would malfunction.
> 
> Is there no way to work around these malfunctions? Does the frequency really have to be runtime configurable to avoid them?

The scope of this commit is to allow rp2040 and rp2350 to run at different (fixed) clock speeds, and it wasn't intended to be run-time configurable. The PWM timing issue can be corrected by updating the rp2 PWM driver to work with these clock frequencies.

eliasnaur:
> > > > Is there a mechanism that allows constants defined in machine-level files to be overridden by a board-level tag?
> > > 
> > > 
> > > No, however those could need to be `var` instead of `const`, which would allow that to be changed in a board file.
> > 
> > 
> > Note that some code, such as [`delay.Sleep`](https://github.com/tinygo-org/drivers/blob/019bbbe5fb66ff6ecd50949520253a50ecc72733/delay/sleep.go#L42), depends on `machine.CPUFrequency` being constant.
> 
> That's good insight! I'm confused by that code and comment. It seems like the function will work even if CPUFrequency() isn't constant, but just be somewhat inaccurate for short delays. It also seems confusing that it expects a function call to evaluate to a constant.
> 

Note that I'm not familiar enough with the compiler to know, but: it's possible your `clockCfg` variable will be deemed read-only, placed in ROM, and then `CPUFrequency` may still be constant enough for `delay.Sleep`.

cibomahto:
I've checked in an updated version of this pull request, changing the clock configurations back to constants.

The PWM generation is still a blocker, and since the performance gain from 125 to 150MHz isn't really significant, I'm no longer sure it's worth rewriting, so I'd be fine with dropping this change set.

cibomahto:
Closing this for now, if someone wants to take over please do.

amken3d:
Addressed in #4723.

Issue #4678, opened 2025-01-02
top level 'make lint' target fails


cibomahto:
Using the latest tinygo release 2a76ceb7dd5ea5a834ec470b724882564d9681b3 (but also the dev branch), if I attempt to run 'make lint', it fails:

```
make lint
/bin/sh: 1: node: not found
cd internal/tools && go generate -tags tools ./
go: downloading github.com/mgechev/revive v1.3.9
go: downloading github.com/golangci/misspell v0.6.0
go: downloading github.com/mitchellh/go-homedir v1.1.0
go: downloading github.com/pkg/errors v0.9.1
go: downloading github.com/fatih/color v1.17.0
go: downloading github.com/mgechev/dots v0.0.0-20210922191527-e955255bf517
go: downloading github.com/spf13/afero v1.11.0
go: downloading github.com/hashicorp/go-version v1.7.0
go: downloading github.com/BurntSushi/toml v1.4.0
go: downloading github.com/chavacava/garif v0.1.0
go: downloading github.com/olekukonko/tablewriter v0.0.5
go: downloading github.com/fatih/structtag v1.2.0
go: downloading golang.org/x/tools v0.23.0
go: downloading golang.org/x/text v0.14.0
go: downloading github.com/mattn/go-colorable v0.1.13
go: downloading github.com/mattn/go-isatty v0.0.20
go: downloading github.com/mattn/go-runewidth v0.0.9
go: downloading golang.org/x/sys v0.22.0
revive -version
version 1.3.9
# TODO: lint more directories!
# revive.toml isn't flexible enough to filter out just one kind of error from a checker, so do it with grep here.
# Can't use grep with friendly formatter.  Plain output isn't too bad, though.
# Use 'grep .' to get rid of stray blank line
revive -config revive.toml compiler/... src/{os,reflect}/*.go | grep -v "should have comment or be unexported" | grep '.' | awk '{print}; END {exit NR>0}'
linting - getting packages: getting packages - resolving packages in dots: malformed import path "src/{os,reflect}/*.go": invalid char '{'
```

cibomahto:
If I replace the failing glob with find, so that it checks all of the files in src/:

```
LINTFILESCMD=find src/ -type f -name '*.go'
.PHONY: lint
lint: tools ## Lint source tree
        revive -version
        # TODO: lint more directories!
        # revive.toml isn't flexible enough to filter out just one kind of error from a checker, so do it with grep here.
[...]

PR #4679, opened 2025-01-02
Fix for #4678: top-level 'make lint' wasn't working

cibomahto:
As noted in #4678, the revive command seems to have had a syntax error in the file input glob. It appears to have been broken in a way that did not result in a return code being set. This change uses 'find' to build the input to the linter.

Note that it is expected to fail the CI script, because it is uncovering some existing lint issues that were not being caught.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1000,7 +1000,7 @@ endif
 tools:
 	cd internal/tools && go generate -tags tools ./
 
-LINTFILESCMD=find src/os/ src/reflect/ -type f -name '*.go'
+LINTDIRS=src/os/ src/reflect/
 .PHONY: lint
 lint: tools ## Lint source tree
 	revive -version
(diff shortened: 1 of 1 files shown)


GNUmakefile
```diff
@@ -1000,14 +1000,18 @@ endif
 tools:
 	cd internal/tools && go generate -tags tools ./
 
+LINTFILESCMD=find src/os/ src/reflect/ -type f -name '*.go'
```
b0ch3nski: How about changing this variable to contain a list of directories to scan? That would be easier to maintain/adjust/expand in future.

cibomahto: @b0ch3nski Similar to this? https://github.com/tinygo-org/tinygo/pull/4679/commits/7190e019ae8a6715628125025906f037405f6839

b0ch3nski: @cibomahto LGTM :+1:

b0ch3nski (approved):

deadprogram:
This PR seems like a good thing, but then suddenly will require a followup to fix all of the issues it raises, if I am not mistaken?

cibomahto:
You are correct! If the issues that this surfaces aren't important, then an alternative is to just delete this test.

deadprogram:
Closing this PR since the commits were merged in #4997 

Thank you @cibomahto for the fixes and to @b0ch3nski for review.

PR #4683, opened 2025-01-04
machine: adding support for waveshare-rp2040-tiny

pobochiigo:
Adding support for Waveshare [rp2040-tiny](https://www.waveshare.com/wiki/RP2040-Tiny) board

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -754,12 +754,6 @@ endif
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=feather-m4          examples/pwm
 	@$(MD5SUM) test.hex
-	# test uart
-	$(TINYGO) build -size short -o test.hex -target=waveshare-rp2040-tiny examples/uart
-	@$(MD5SUM) test.hex
-	# test i2c
-	$(TINYGO) build -size short -o test.hex -target=waveshare-rp2040-tiny examples/i2c-target
-	@$(MD5SUM) test.hex
 	# test usb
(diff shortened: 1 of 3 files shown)


ysoldak (changes requested):
@pobochiigo  You want to add your board to the smoketest suite. 
See https://github.com/tinygo-org/tinygo/blob/release/GNUmakefile#L516

pobochiigo:
> @pobochiigo You want to add your board to the smoketest suite. See https://github.com/tinygo-org/tinygo/blob/release/GNUmakefile#L516

done

ysoldak:
@aykevl @deadprogram are we OK with multiple smokes for a board or we want just one per board?

ysoldak (approved):
Fine with me, I'll let @aykevl or @deadprogram decide if 3 smokes is to much :D

deadprogram:
I think one is enough. The smoke tests already take a while to run.

ysoldak:
@pobochiigo we just want to be sure we [smoke-]test all boards, so one smoke per board is enough.

To test permutations of all boards with multiple examples would be good, but is a non-goal for now -- tests already take considerable time.

Would you drop changes in lines from 757 to 762, please? Then we can merge the PR. Thank you!

pobochiigo:
> @pobochiigo we just want to be sure we [smoke-]test all boards, so one smoke per board is enough.
> 
> To test permutations of all boards with multiple examples would be good, but is a non-goal for now -- tests already take considerable time.
> 
> Would you drop changes in lines from 757 to 762, please? Then we can merge the PR. Thank you!

yep, sure

ysoldak (approved):

PR #4686, opened 2025-01-05
builder: Fix parsing of external ld.lld error messages

QuLogic:
If `ld.lld` is a version-specific binary (e.g., `ld.lld-18`), then its error messages include the version. The parsing previously incorrectly assumed it would be unversioned.


aykevl:
Thank you!

PR #4692, opened 2025-01-12
fix: correctly handle calls for GetRNG() when being made from nrf devices with SoftDevice enabled

deadprogram:
This PR is to correctly handle calls for GetRNG() when being made from nrf devices with SoftDevice enabled.

Thanks @aykevl for pointing me towards the needed changes.

Fixes https://github.com/tinygo-org/bluetooth/issues/310

### DIFF
--- a/src/machine/machine_nrf_sd.go
+++ b/src/machine/machine_nrf_sd.go
@@ -1,0 +1,59 @@
+//go:build nrf && softdevice
+
+package machine
+
+import (
+	"device/arm"
+	"device/nrf"
+
+	"errors"
+)
+
+// avoid a heap allocation in GetRNG.
+var (
+	softdeviceEnabled uint8
+	bytesAvailable    uint8
+	buf               [4]uint8
+
+	errNoSoftDeviceSupport = errors.New("rng: softdevice not supported on this device")
+	errNotEnoughRandomData = errors.New("rng: not enough random data available")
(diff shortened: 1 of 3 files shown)


src/machine/machine_nrf_sd.go
```diff
@@ -0,0 +1,56 @@
...
+	if softdeviceEnabled == 0 {
+		return getRNG()
+	}
+
+	// call into the SoftDevice to get random data bytes available
+	switch nrf.Device {
+	case "nrf51":
+		// sd_rand_application_bytes_available_get: SOC_SVC_BASE_NOT_AVAILABLE + 4
+		arm.SVCall1(0x2B+4, &bytesAvailable)
+	case "nrf52", "nrf52840", "nrf52833":
```
ysoldak, line 38: Is there a case when we have `"nrf52840", "nrf52833"` and not `"nrf52"`?
I mean this can probably be dropped: `", nrf52840", "nrf52833"`?

deadprogram: I stole this code from https://github.com/tinygo-org/tinygo/blob/release/src/runtime/runtime_nrf_softdevice.go#L32

aykevl (approved):

deadprogram:
Thanks for review @aykevl now merging.

PR #4695, opened 2025-01-13
sync: move Mutex to internal/task

aykevl:
The mutex implementation needs a different implementation once support for threading lands. This implementation just moves code to the internal/task package to centralize these algorithms.
We need a mutex implementation in internal/task so it's better to share code, by moving this one here as well to keep things organized.

---

This is a replacement for #4650.


deadprogram:
Thanks for the movement @aykevl now merging.

PR #4699, opened 2025-01-14
Feat/nrf52xxx/spi/improve

milkpirate:
As requested per https://github.com/tinygo-org/tinygo/pull/4660#pullrequestreview-2538408575

### SPI
- Create SPIMode type
- Make SPI implement io.Reader, io.Writer
- Update formatting to align with other switch-cases
- Update SPI documentation reference


src/machine/machine_nrf52xxx.go
```diff
@@ -234,23 +265,7 @@ func (spi SPI) Configure(config SPIConfig) error {
...
-		conf |= (nrf.SPIM_CONFIG_CPOL_ActiveLow << nrf.SPIM_CONFIG_CPOL_Pos)
-		conf &^= (nrf.SPIM_CONFIG_CPHA_Leading << nrf.SPIM_CONFIG_CPHA_Pos)
-	case 3:
-		conf |= (nrf.SPIM_CONFIG_CPOL_ActiveLow << nrf.SPIM_CONFIG_CPOL_Pos)
-		conf |= (nrf.SPIM_CONFIG_CPHA_Trailing << nrf.SPIM_CONFIG_CPHA_Pos)
-	default: // to mode
-		conf &^= (nrf.SPIM_CONFIG_CPOL_ActiveHigh << nrf.SPIM_CONFIG_CPOL_Pos)
-		conf &^= (nrf.SPIM_CONFIG_CPHA_Leading << nrf.SPIM_CONFIG_CPHA_Pos)
-	}
+	conf = config.Mode.ApplyTo(conf)
```
ysoldak: I don't feel this is necessary?
`SPIMode` type 👍 
`ApplyTo()` method 👎

src/machine/machine_nrf52xxx.go
```diff
@@ -176,39 +218,28 @@ func (a ADC) Get() uint16 {
...
-	// Pick a default frequency.
-	if config.Frequency == 0 {
-		config.Frequency = 4000000 // 4MHz
-	}
-
 	// set frequency
 	var freq uint32
 	switch {
+	case config.Frequency == 0: // default MCU SPI speed
+		freq = nrf.SPIM_FREQUENCY_FREQUENCY_M4
```
ysoldak, line 234: I don't see this as improvement :( Previous code looks cleaner to me: separates concerns of a) ensuring default value and b) handling frequency.

milkpirate: Found the separate if-statement superfluous, since we already have a flow control on the frequency.

src/machine/machine_nrf52xxx.go
milkpirate: I like it because:

- It says what it does. It applies the mode to the config
- It make the switch case testable
- It hides the implementation details which are not necessary at this level of logic
- Though they are still inspect-able if one is really interested

I would like to keep it in if you have not striking counter arguments. I can also make it private if you dont like it being public.

soypat: This change is a machine level exported API change, which is not in itself a bad change but it must be discussed beforehand. A lot of things must be checked before making such a change.

> It hides the implementation details which are not necessary at this level of logic

Configure takes a SPIConfig and configures hardware. There is no intermediate step to abstract. There is no shared logic here between other MCUs. "Not necessary" are too strong words. Maybe one could argue that modularizing functions within Configure allows sharing logic between other higher level functions; but that is not the case here. The only argument I see for this change is that Configure is too long of a function to comfortably read- and even in that case I'd argue: on what basis? There seems to be ~6 actions performed within Configure which add to the cognitive load-  and your change does not actually reduce the amount of actions performed, it just shortens the code, which might make it easier to read but does not reduce the intrinsic amount of congitive load of the Configure method.

I'd strongly err on the side of avoiding changes that make code less localized without reducing cognitive load.

src/machine/machine_nrf52xxx.go
```diff
@@ -324,6 +338,16 @@ func (spi SPI) Tx(w, r []byte) error {
...
+// Read implements [io.Reader]. And reads as many bytes as the given buffer is long
+func (spi *SPI) Read(r []byte) (int, error) {
+	return len(r), spi.Tx(nil, r)
+}
+
+// Write implements [io.Writer]. And writes as long as there are bytes in w.
+func (spi *SPI) Write(w []byte) (int, error) {
+	return len(w), spi.Tx(w, nil)
+}
+
```
soypat, line 358: Read and Write are not part of the SPI interface. It is easy enough to implement an abstraction if needed that works for all SPI hardwares of all MCUs

```go
func writeSPI(w []byte, spi interface {Tx(w,r []byte) error}) (int,error) {
    return len(w), spi.Tx(w,nil)
}
```

src/machine/machine_nrf52xxx.go
```diff
@@ -8,6 +8,48 @@ import (
...
+)
+
+// There are 3 SPI interfaces on the NRF528xx.
+var (
+	SPI0 = SPI{Bus: nrf.SPIM0}
+	SPI1 = SPI{Bus: nrf.SPIM1}
+	SPI2 = SPI{Bus: nrf.SPIM2}
+)
+
+type SPIMode uint8
```
soypat, line 30: This is a breaking change. It must be discussed in an issue beforehand.

src/machine/machine_nrf52xxx.go
```diff
@@ -176,39 +218,28 @@ func (a ADC) Get() uint16 {
...
 	SDI       Pin
 	LSBFirst  bool
-	Mode      uint8
+	Mode      SPIMode
 }
 
-// Configure is intended to setup the SPI interface.
-func (spi SPI) Configure(config SPIConfig) error {
+// Configure is intended to set up the SPI interface.
+func (spi *SPI) Configure(config SPIConfig) error {
```
soypat, line 226: The pointer receiver is not a change I'm against, in fact I think it may make sense to change it. I'd be curious as to why you are changing it :)

soypat (changes requested):
machine package breaking changes must be discussed beforehand. And some nits

src/machine/machine_nrf52xxx.go
milkpirate: Test ability is no argument for you, ok. Will inline.

src/machine/machine_nrf52xxx.go
milkpirate: Yeah, lets do it! Actually I was surprised that SPI is not supplying it, since it would have seemed the more natural way to interface with SPI than the current one. Can you point me to the relevant file/code section?

src/machine/machine_nrf52xxx.go
milkpirate: Ok, will remove it from the config struct definition, so one can pass to the config but the type is not required.

src/machine/machine_nrf52xxx.go
milkpirate: - Well the SPI interfaces are created as singletons, so why should we copy it on its receivers.
- One would usually expect to configure the very instance where `Configure` is attached to and not a copy (that gets thrown away. Technically we configure the hardware not the instance but it still feels odd and as I said counter intuitive.

src/machine/machine_nrf52xxx.go
milkpirate: can you two @ysoldak, @b0ch3nski agree on that?

b0ch3nski: I find switch statements much easier to read but I think it's just a matter of personal preference. Nevertheless, we already have a switch here and that lonely if statement looks like it has rejoined it's people :smiley:

milkpirate: > I find switch statements much easier to read but I think it's just a matter of personal preference. Nevertheless, we already have a switch here and that lonely if statement looks like it has rejoined it's people 😃

Yes, felt the same to me.

ysoldak: Regarding switch, I must confess, I was struggling to understand what's going on at first.
I was expecting every 'case' to produce a unique value (number of unique values = number of cases in switch).
I would propose merge two cases, but then case condition becomes long and look ugly.

So I still prefer deal with default value first, then handle the value, regardless default it is or not.
It may look nice/smart to have only one switch without an extra "if" before it, because we can, but I'd say readability suffers, at least in my case.

aykevl (changes requested):
There are a few different things this PR does, which I think should be done separately. Some I agree with, some I don't. Especially since it affects the public API. Example: I think making the SPI type a pointer receive is a good idea, but that should probably be done for all SPI implementations at the same time to ensure consistency. And be done separate from other changes, like the mode constants (which, again, should be done for all chips at the same time in my opinion).

PR #4701, opened 2025-01-14
Feat/nrf52xxx/adc/make resolution changeable

milkpirate:
As requested per https://github.com/tinygo-org/tinygo/pull/4660#pullrequestreview-2538408575

### ADC
- Change ADC methods to pointer receivers
- Make ADC resolution adjustable
- Remove nonsensical calculations (at least to me they seemed like, please correct me if I'm wrong)
- Update ADC documentation reference

## Open questions
- Will make sense to change the signature of `ADC.Get()` to return an error? Since it will change the interface and might break things.

### DIFF
--- a/src/machine/machine_nrf52xxx.go
+++ b/src/machine/machine_nrf52xxx.go
@@ -25,7 +25,6 @@ func InitADC() {
 // SampleTime will be ceiled to 3(default), 5, 10, 15, 20 or 40(max) µS respectively
 // Samples can be 1(default), 2, 4, 8, 16, 32, 64, 128, 256 samples
 func (a *ADC) Configure(config ADCConfig) {
-
 	var configVal uint32 = nrf.SAADC_CH_CONFIG_RESP_Bypass<<nrf.SAADC_CH_CONFIG_RESP_Pos |
 		nrf.SAADC_CH_CONFIG_RESP_Bypass<<nrf.SAADC_CH_CONFIG_RESN_Pos |
 		nrf.SAADC_CH_CONFIG_REFSEL_Internal<<nrf.SAADC_CH_CONFIG_REFSEL_Pos |
@@ -169,7 +168,22 @@ func (a *ADC) Get() uint16 {
 	}
 	nrf.SAADC.EVENTS_STOPPED.Set(0)
 
-	return rawValue.Get()
+	// convert to 16 bit resolution/value
+	var resolutionAdjustment uint8
+	switch nrf.SAADC.RESOLUTION.Get() {
+	case nrf.SAADC_RESOLUTION_VAL_8bit:
+		resolutionAdjustment = 8
+	case nrf.SAADC_RESOLUTION_VAL_10bit:
+		resolutionAdjustment = 6
+	case nrf.SAADC_RESOLUTION_VAL_12bit:
+		resolutionAdjustment = 4
+	case nrf.SAADC_RESOLUTION_VAL_14bit:
+		resolutionAdjustment = 2
+	default:
+		resolutionAdjustment = 4 // 12bit
+	}
+
+	return rawValue.Get() << resolutionAdjustment
 }
 
 // SPI on the NRF.
--- a/src/machine/machine_nrf52xxx.go
+++ b/src/machine/machine_nrf52xxx.go
@@ -52,8 +52,23 @@ func (a *ADC) Configure(config ADCConfig) {
 		// TODO: return an error, will that interfere with any interfaced if one will be?
 	}
 
-	// Source resistance, according to table 89 on page 364 of the nrf52832 datasheet.
-	// https://infocenter.nordicsemi.com/pdf/nRF52832_PS_v1.4.pdf
+	var resolution uint32
+	switch config.Resolution {
+	case 8:
+		resolution = nrf.SAADC_RESOLUTION_VAL_8bit
+	case 10:
+		resolution = nrf.SAADC_RESOLUTION_VAL_10bit
+	case 12:
+		resolution = nrf.SAADC_RESOLUTION_VAL_12bit
+	case 14:
+		resolution = nrf.SAADC_RESOLUTION_VAL_14bit
+	default:
+		resolution = nrf.SAADC_RESOLUTION_VAL_12bit
+	}
+	nrf.SAADC.RESOLUTION.Set(resolution)
+
+	// Source resistance, according to table 41 on page 676 of the nrf52832 datasheet.
+	// https://docs-be.nordicsemi.com/bundle/ps_nrf52840/attach/nRF52840_PS_v1.11.pdf?_LANG=enus
 	if config.SampleTime <= 3 { // <= 10kΩ
 		configVal |= nrf.SAADC_CH_CONFIG_TACQ_3us << nrf.SAADC_CH_CONFIG_TACQ_Pos
 	} else if config.SampleTime <= 5 { // <= 40kΩ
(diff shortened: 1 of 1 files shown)


src/machine/machine_nrf52xxx.go
```diff
@@ -164,13 +169,7 @@ func (a ADC) Get() uint16 {
...
 	nrf.SAADC.EVENTS_STOPPED.Set(0)
 
-	value := int16(rawValue.Get())
-	if value < 0 {
-		value = 0
-	}
-
-	// Return 16-bit result from 12-bit value.
-	return uint16(value << 4)
+	return rawValue.Get()
```
ysoldak: Can this break something? Like a client code out there in the wild expects 16bit value.

src/machine/machine_nrf52xxx.go
```diff
@@ -14,19 +14,17 @@ func CPUFrequency() uint32 {
...
-}
-
-// Configure configures an ADC pin to be able to read analog data.
-func (a ADC) Configure(config ADCConfig) {
 	// Enable ADC.
-	// The ADC does not consume a noticeable amount of current simply by being
-	// enabled.
+	// The ADC does not consume a noticeable amount of current by being enabled.
 	nrf.SAADC.ENABLE.Set(nrf.SAADC_ENABLE_ENABLE_Enabled << nrf.SAADC_ENABLE_ENABLE_Pos)
+}
```
ysoldak, line 20: So now we going to have ADC automatically enabled on NRF chips. If the comment is true, then probably it's fine.
How can we confirm that? NRF chip is particularly good for projects there extreme low power consumption is required (hibernation). Will this change jeopardise this?

src/machine/machine_nrf52xxx.go
milkpirate: Well, honestly this whole block makes no sense to me:

- Why convert a `Uint` to an `int` and then check if its `<0`? 
- Then `value << 4` does not convert something from 12 to 16 bit, it multiplies the value by $2^4 = 16$. Why?

I may not be aware of some special cases why this was done like so, maybe @aykevl (the original author) can shed some light on this.

src/machine/machine_nrf52xxx.go
milkpirate: Well I would expect initialization code to be in a function called `InitADC()`, and not in `Configure()`. If(!) `InitADC()` is always called (maybe by tinygo somewhere), then yes, then the ADC will be always on. Further I do not know if the comment is correct 🤷🏻 probably @aykevl known.

If not it will only be turned on if one does `InitADC()`. I guess it will break every code that does not `InitADC()`.

This poses the question: Whats valued more coherent code or not breaking existing one?

I am defiantly more leaning to the first, but this is not my party 😄

src/machine/machine_nrf52xxx.go
ysoldak: I believe the idea is (though I can't find this in documentation, probably picked up this from slack conversations) the value is always 16bit (i.e. ADC max is always 65535). But when hardware can't deliver that, we simply have less resolution. This way the client code does not need to worry about ADC MAX.

src/machine/machine_nrf52xxx.go
ysoldak: Right, I did assume wrong, `InitADC()` is not called in runtime init method on NRF chips, like `InitSerial()` does, for example. See https://github.com/tinygo-org/tinygo/blob/release/src/runtime/runtime_nrf.go#L29-L33
And this example confirms that, client code need to call it explicitly: https://github.com/tinygo-org/tinygo/blob/release/src/examples/adc/adc.go#L12

Not sure at the moment if this is by design or simply inconsistency that we want to fix.
@aykevl

src/machine/machine_nrf52xxx.go
milkpirate: Ahhh! Ohhh!... Ok, well then the shift must depend on the resolution. Can add that if you like. Then it remain compatible with current usages.

src/machine/machine_nrf52xxx.go
deadprogram: This is by design. It you are not using ADC then you don't have to bring it with you.

src/machine/machine_nrf52xxx.go
ysoldak: Yes, correct, shift depends on ADC resolution.

src/machine/machine_nrf52xxx.go
ysoldak: Hm, why do we init serial unconditionally then?

deadprogram: Several reasons, one of which is that without serial, you cannot use the auto-detect of 1200 baud for kicking into bootloader mode.

src/machine/machine_nrf52xxx.go
```diff
@@ -164,13 +168,22 @@ func (a ADC) Get() uint16 {
...
+	case nrf.SAADC_RESOLUTION_VAL_8bit:
+		rawValue <<= 8
+	case nrf.SAADC_RESOLUTION_VAL_10bit:
+		rawValue <<= 6
+	case nrf.SAADC_RESOLUTION_VAL_12bit:
+		rawValue <<= 4
+	case nrf.SAADC_RESOLUTION_VAL_14bit:
+		rawValue <<= 2
+	default:
+		rawValue <<= 4 // 12bit
```
milkpirate: The switch case can be replaced by:
```golang
	// max precision 14bit (<<= 2, for 16 bit value)
	// min precision 8bit (<<= 8, for 16 bit value)
	// resolution constants 8bit ~ 0, 14bit ~ 3
	precisionAdjustment := 8 - 2*nrf.SAADC.RESOLUTION.Get()
        rawValue <<= precisionAdjustment
```
its more compact, but also "magical". Just tell me what you prefer.

src/machine/machine_nrf52xxx.go
milkpirate: So leave it in? @deadprogram

src/machine/machine_nrf52xxx.go
ysoldak: I like this simple linear function conversion instead of a multiline switch 👍

ysoldak (commented):
Looks fine to me.
Not 100% onboard with (unnecessary) pointer receiver change, but also have nothing against.
Feel free do 'magic' conversion for resolution adjustment,  but it is fine with switch too.
Ready to approve once PR is not a draft anymore.

ysoldak (approved):

deadprogram:
Hello @milkpirate and @ysoldak 

Looks like a new error now since this was merged in the TinyHCI hardware tests for nrf52840:

```shell
analogReadGround = fail
expected: 'val <= 4096'
actual: 52374
```

See https://github.com/tinygo-org/tinygo/runs/35896914622

The code that runs this test is located here: 
https://github.com/tinygo-org/tinyhci/blob/main/itsybitsy-nrf52840/main.go#L197-L221

ysoldak:
Hmm, sad. How did the test pass in PR then? 🤔

ysoldak:
@deadprogram could you probably re-check the test setup?
I can't see how this code change would make that test fail.

Besides, "itsybitsy-nrf52840" test was failing already couple commits back (albeit due to different reason) and after totally unrelated change.
https://github.com/tinygo-org/tinygo/runs/35880998371

deadprogram:
I did retry several times. The reason for the previous failure was something else unrelated.

ysoldak:
See https://github.com/tinygo-org/tinygo/pull/4710

PR #4703, opened 2025-01-16
stm32: Fix ADC channel selecting and ADC value reading

grudzinski:


src/machine/machine_stm32_adc_f1.go
```diff
@@ -24,7 +24,7 @@ func InitADC() {
 	enableAltFuncClock(unsafe.Pointer(stm32.ADC1))
 
 	// enable
-	stm32.ADC1.CR2.SetBits(stm32.ADC_CR2_ADON)
+	stm32.ADC1.CR2.SetBits(stm32.ADC_CR2_ADON | stm32.ADC_CR2_ALIGN)
```
grudzinski, line 27: ADC can `ALIGN` left as we did in code `<< 4`  by setting flag `stm32.ADC_CR2_ALIGN`.

src/machine/machine_stm32_adc_f1.go
```diff
@@ -49,7 +49,7 @@ func (a ADC) Configure(ADCConfig) {
 func (a ADC) Get() uint16 {
 	// set rank
 	ch := uint32(a.getChannel())
-	stm32.ADC1.SQR3.SetBits(ch)
+	stm32.ADC1.SetSQR3_SQ1(ch)
```
grudzinski, line 52: Unset prev value and set new

src/machine/machine_stm32_adc_f1.go
deadprogram: Good trick!

src/machine/machine_stm32_adc_f1.go
deadprogram: :+1: for using setter function here.

deadprogram:
Thanks for the additional fix here @grudzinski now merging.

PR #4705, opened 2025-01-18
add support for Elecrow Pico rp2040 W5 boards

scottfeldman:
https://www.elecrow.com/pico-w5-microcontroller-development-boards-rp2040-microcontroller-board-support-wifi-2-4ghz-5ghz-bluetooth5.html

Just the basics to get several test to work (blinky1, flash).

This board has an rtl8720d Wifi/bluetooth chip wired to UART1, but the rtl8720d firmware installed is the AT cmd set, not the RPC interface used by the rtl8720dn driver, so no wifi support at the moment.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -861,7 +861,7 @@ endif
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=tkey                examples/blinky1
 	@$(MD5SUM) test.hex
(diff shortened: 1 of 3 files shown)


ysoldak (changes requested):
Please add board to the smoke test suite (one test per board is enough)
https://github.com/tinygo-org/tinygo/blob/release/GNUmakefile#L516

ysoldak (approved):

ysoldak:
Thanks, @scottfeldman, merging now!

PR #4710, opened 2025-01-21
nrf: fix adc read near zero

ysoldak:
Hopefully fixes bug introduced in #4701

I still don't understand:
1) how tests passed on PR but failed on dev after merge
2) why would `rawValue.Get()` return negative value

### DIFF
--- a/src/machine/machine_nrf52xxx.go
+++ b/src/machine/machine_nrf52xxx.go
@@ -183,7 +183,12 @@ func (a *ADC) Get() uint16 {
 		resolutionAdjustment = 4 // 12bit
 	}
 
-	return rawValue.Get() << resolutionAdjustment
+	value := int16(rawValue.Get())
+	if value < 0 {
+		value = 0
+	}
+
+	return uint16(value << resolutionAdjustment)
 }
 
 // SPI on the NRF.


ysoldak:
@deadprogram tinyhci tests fail due to mount errors, ptal

src/machine/machine_nrf52xxx.go
```diff
@@ -183,7 +183,12 @@ func (a *ADC) Get() uint16 {
 		resolutionAdjustment = 4 // 12bit
 	}
 
-	return rawValue.Get() << resolutionAdjustment
+	value := int16(rawValue.Get())
```
deadprogram, line 186: This should probably be a cast to `uint16`

deadprogram:
> how tests passed on PR but failed on dev after merge

The TinyHCI tests are only run on PRs from team members.

> why would rawValue.Get() return negative value

I think that needs to be cast to `uint16`.

src/machine/machine_nrf52xxx.go
ysoldak: This is how it was before, see https://github.com/tinygo-org/tinygo/pull/4701/files#diff-e8032f214fa1ae692ba79fc6179cd57e7325ec7d8da3152ff0e4223587f1a041L167

deadprogram: Isn't that previous cast to int16?

```go
value := int16(rawValue.Get())
```

src/machine/machine_nrf52xxx.go
```diff
@@ -183,7 +183,12 @@ func (a *ADC) Get() uint16 {
...
+	if value < 0 {
+		value = 0
+	}
+
+	return uint16(value << resolutionAdjustment)
```
ysoldak, line 191: @deadprogram we do return `uint16` in the end, if that's you after.

ysoldak: I've just repeated exact logic as we had before: casting register value to `int16`, checking for `0` and then casting to `uint16` after applying resolution adjustment

ysoldak:
@deadprogram seem like HCI tests pass now with this change

deadprogram:
Thanks for the fix @ysoldak now merging.

PR #4713, opened 2025-01-22
targets: match Pico2 stack size to Pico

eliasnaur:
Took me a while to debug weird crashes after switching from Pico to Pico2.


deadprogram:
Thanks for adding that @eliasnaur now merging.

PR #4720, opened 2025-01-28
Add support for threading on MacOS

aykevl:
This is based on #4559. This PR itself is very small, most of the work was in making all headers available in lib/macos-minimal-sdk.

### DIFF
--- a/src/internal/task/task_threads.c
+++ b/src/internal/task/task_threads.c
@@ -2,16 +2,28 @@
 
 #define _GNU_SOURCE
 #include <pthread.h>
-#include <semaphore.h>
 #include <signal.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <unistd.h>
 
-// BDWGC also uses SIGRTMIN+6 on Linux, which seems like a reasonable choice.
 #ifdef __linux__
+#include <semaphore.h>
+
+// BDWGC also uses SIGRTMIN+6 on Linux, which seems like a reasonable choice.
 #define taskPauseSignal (SIGRTMIN + 6)
-#endif
+
+#elif __APPLE__
+#include <dispatch/dispatch.h>
+// SIGIO is for interrupt-driven I/O.
+// I don't think anybody should be using this nowadays, so I think we can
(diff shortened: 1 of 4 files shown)


aykevl:
This PR is now ready!

src/internal/task/task_threads.c
```diff
@@ -22,7 +30,11 @@ struct state_pass {
     void      *args;
     void      *task;
     uintptr_t *stackTop;
+    #if __APPLE__
```
eliasnaur, line 38: Preprocessor directives are somewhat ugly, and get uglier with the addition of platforms (e.g. Windows). I suggest adding a GOOS-specific typedef for the semaphore type along with a handful of small inline functions that abstracts the semaphore API.

eliasnaur (approved):
Change looks good. Small suggestion below.

src/internal/task/task_threads.c
aykevl: That is what I originally tried, but in my opinion didn't result in code that was easier to read. Note that the way these functions are called is very different between standard POSIX (`sem_*` API) and MacOS.

I will likely make a new file for Windows. Maybe I can use the pthread library of MinGW and reuse this code, but I expect Windows will be so different a separate file makes things easier. We'll see.

deadprogram:
@aykevl something not quite right with this PR, I think. The CI tests on macOS run "forever" until they timeout.

aykevl:
Yeah that doesn't look quite right, I need to debug this on MacOS.

I've cancelled the CI for now, it's not going to get any further this way.

aykevl:
Fixed the CI hang: there was a problem with the stop-the-world phase (other threads never received the signal) that is now fixed.

debug/dwarf still fails when I run it locally, but it also fails on the dev branch so that seems to be an unrelated problem. Other than that, stdlib packages now pass the tests.

eliasnaur:
Still LGTM.

deadprogram:
Thanks for working on this @aykevl and to @eliasnaur for reviews. Now merging.

PR #4723, opened 2025-02-02
Rp2350 b support

amken3d:
Adding support for the RP2350B
a. Support for Pins 31 thru 47
b. Updated ADC
c. Updated PWM
d. Updated Pico2, elecrow and tiny2350 json to add rp2350a buildtag
e. Added Amken Max14 board (Intelligent motor controller board in Alpha testing stages)

All test pass using make test
ok      github.com/tinygo-org/tinygo/builder    (cached)
ok      github.com/tinygo-org/tinygo/cgo        (cached)
ok      github.com/tinygo-org/tinygo/compileopts        (cached)
ok      github.com/tinygo-org/tinygo/compiler   (cached)
ok      github.com/tinygo-org/tinygo/interp     (cached)
ok      github.com/tinygo-org/tinygo/transform  (cached)
ok      github.com/tinygo-org/tinygo    60.810s


targets/pico2.json
```diff
@@ -2,7 +2,7 @@
     "inherits": [
         "rp2350"
     ],
-    "build-tags": ["pico2"],
+    "build-tags": ["pico2", "rp2350a"],
```
deadprogram, line 5: I wonder if we should deprecate the existing rp2350 target in favor of a new `rp2350a.json` so specific boards can just inherit from either `rp2350a` or `rp2350b` depending on which chip they are using? That just seems a bit more straightforward to me.

If we really do not want to remove the `rp2350` target completely, it could inherit from `rp2350a`.

eliasnaur: FWIW, I argue for the other way around: define *only* the `rp2350` that covers both rp2350a and rp2350b by exposing all the extra pins etc (`rp2350b` in this PR). It's then up to boards and/or particular applications to know which model they're running on.

deadprogram: It is not completely clear that we can do that with just pin mapping, since it appears that there are also things like additional ADC channels that only exist on the "B" version. But perhaps that does not matter for compilation as long as you do not try to use a thing that does not exist on the board you are targeting?

I personally still think being explicit about which variation you are using with our current build tags based system is probably a good thing.

eliasnaur: (Still FWIW) I see your point about being explicit, however consider the advantage of portable firmware builds. That is, assuming (1) that model A is a subset of B, and (2) the added pins/units can easily be detected from registers, I consider it a win that you can build 1 firmware image for both variants.

As a personal example, I'm building on rp2350a but may run out of pins for a future hardware revision and upgrade to rp2350b. I'd love to have 1 firmware image for both boards.

src/machine/machine_rp2_pwm.go
```diff
@@ -0,0 +1,388 @@
...
+
+// Counter returns the current counter value of the timer in this PWM
+// peripheral. It may be useful for debugging.
+func (p *pwmGroup) Counter() uint32 {
+	return (p.CTR.Get() & rp.PWM_CH0_CTR_CH0_CTR_Msk) >> rp.PWM_CH0_CTR_CH0_CTR_Pos
+}
+
+// Period returns the used PWM period in nanoseconds.
+func (p *pwmGroup) Period() uint64 {
+	periodPerCycle := cpuPeriod()
```
eliasnaur, line 119: Given https://github.com/tinygo-org/tinygo/pull/4674#issuecomment-2639124943, I don't think `cpuPeriod` should be used anymore; its calculation `1e9/CPUFrequency()` won't be integer at 133MHz or 150MHz.

soypat: Why should we not use it anymore?

targets/pico2.json
soypat: I support being able to build 1 firmware for both- if it means an extra register read so be it :)

src/machine/machine_rp2_pins.go
```diff
@@ -1,4 +1,4 @@
-//go:build rp2040 || rp2350 || ae_rp2040 || badger2040 || challenger_rp2040 || feather_rp2040 || gopher_badge || kb2040 || macropad_rp2040 || nano_rp2040 || pico || qtpy_rp2040 || thingplus_rp2040 || thumby || tufty2040 || waveshare_rp2040_zero || xiao_rp2040
+//go:build rp2040 || rp2350a || ae_rp2040 || badger2040 || challenger_rp2040 || feather_rp2040 || gopher_badge || kb2040 || macropad_rp2040 || nano_rp2040 || pico || qtpy_rp2040 || thingplus_rp2040 || thumby || tufty2040 || waveshare_rp2040_zero || xiao_rp2040
```
soypat: i feel like most of these tags are superfluous, no? Shouldn't it work with just `rp2040 || rp2350a` ?

src/machine/machine_rp2_pwm.go
eliasnaur: [cpuPeriod's value](https://github.com/tinygo-org/tinygo/blob/52983794d702af7a00833ae12b0d2e7175e46017/src/machine/machine_rp2_clocks.go#L18-L20) is `1e9/CPUFrequency()` which for 125MHz is a nice round 8, but for 133MHz or 150MHz is not and thus PWM accuracy will suffer.

targets/pico2.json
amken3d: You still need two build tags as the pin alignment is different. ADC0 starts at GPIO26 for A and at GPIO40 for B variant.

src/machine/machine_rp2_pwm.go
amken3d: You will like the change that this new commit introduces. You can select from a wide variety of good frequencies all the way upto 300Mhz, many of which produce clean CPU periods.

amken3d:
Tested on Pico and Pico2. Good stable overclocking is achieved on both
Console output : 
Hello,
sysClockFrequency 300
 Updated sysClockFrequency 200
CPU period: 5000000
 Running at full speed 300
CPU period: 3333333
Hello,
sysClockFrequency 300

Frequency can now be set at runtime. :) 
Enjoy running your Pico's at double the speed.

src/machine/machine_rp2_pins.go
amken3d: I removed the board tags.

deadprogram:
@amken3d please note build failures in latest commit.

amken3d:
@deadprogram Is there some caching in circleci? I am able to build cleanly with gopher_board selected as target.  I updated the board file in [28f938f], but the smoke test keep failing at that spot.

# machine
/root/project/src/machine/board_gopher-badge.go:19:15: undefined: GPIO13
/root/project/src/machine/board_gopher-badge.go:21:18: undefined: GPIO2
/root/project/src/machine/board_gopher-badge.go:22:18: undefined: GPIO15
/root/project/src/machine/board_gopher-badge.go:23:18: undefined: GPIO15
/root/project/src/machine/board_gopher-badge.go:25:21: undefined: GPIO10
/root/project/src/machine/board_gopher-badge.go:26:21: undefined: GPIO11
/root/project/src/machine/board_gopher-badge.go:27:21: undefined: GPIO25
/root/project/src/machine/board_gopher-badge.go:28:21: undefined: GPIO24
/root/project/src/machine/board_gopher-badge.go:29:21: undefined: GPIO22
/root/project/src/machine/board_gopher-badge.go:30:21: undefined: GPIO23
/root/project/src/machine/board_gopher-badge.go:32:22: undefined: GPIO21
/root/project/src/machine/board_gopher-badge.go:33:22: undefined: GPIO19
/root/project/src/machine/board_gopher-badge.go:34:22: undefined: GPIO16
/root/project/src/machine/board_gopher-badge.go:35:22: undefined: GPIO17
/root/project/src/machine/board_gopher-badge.go:36:22: undefined: GPIO18
/root/project/src/machine/board_gopher-badge.go:37:22: undefined: GPIO20
/root/project/src/machine/board_gopher-badge.go:38:22: undefined: GPIO12
/root/project/src/machine/board_gopher-badge.go:40:23: undefined: GPIO14
/root/project/src/machine/board_gopher-badge.go:41:23: undefined: GPIO3
/root/project/src/machine/board_gopher-badge.go:46:21: undefined: GPIO0
/root/project/src/machine/board_gopher-badge.go:47:21: undefined: GPIO1
/root/project/src/machine/board_gopher-badge.go:55:21: undefined: GPIO18
/root/project/src/machine/board_gopher-badge.go:56:21: undefined: GPIO19
/root/project/src/machine/board_gopher-badge.go:57:21: undefined: GPIO16
/root/project/src/machine/board_gopher-badge.go:66:2: undefined: SysClockFrequency
/root/project/src/machine/board_gopher-badge.go:66:22: undefined: Freq133MHz
/root/project/src/machine/board_gopher-badge.go:67:2: undefined: xoscFreq
/root/project/src/machine/board_gopher-badge.go:67:13: undefined: Freq12MHz
/root/project/src/machine/board_gopher-badge.go:83:17: undefined: GPIO0
/root/project/src/machine/board_gopher-badge.go:84:17: undefined: GPIO1
/root/project/src/machine/board_gopher-badge.go:85:17: undefined: GPIO4
/root/project/src/machine/board_gopher-badge.go:86:17: undefined: GPIO5
make: *** [GNUmakefile:587: smoketest] Error 1

amken3d:
@deadprogram , need some help. I am not able to figure out why it fails.  make test runs clean on my end. 
ok      github.com/tinygo-org/tinygo/builder    2.770s
ok      github.com/tinygo-org/tinygo/cgo        (cached)
ok      github.com/tinygo-org/tinygo/compileopts        (cached)
ok      github.com/tinygo-org/tinygo/compiler   1.225s
ok      github.com/tinygo-org/tinygo/interp     (cached)
ok      github.com/tinygo-org/tinygo/transform  0.087s
ok      github.com/tinygo-org/tinygo    76.003s

deadprogram:
https://github.com/tinygo-org/tinygo/actions/runs/13227290372/job/36919834194?pr=4723#step:10:151

I don't think is cache related

src/machine/machine_rp2_pwm.go
eliasnaur: Thank you! However, do you know whether `CPUFrequency` is still constant enough to satisfy the constraints of `delay.Sleep`? See https://github.com/tinygo-org/tinygo/pull/4674#issuecomment-2634937017

eliasnaur: Hm, you're still using `cpuPeriod`, so the PWM frequency issue still exists at "uneven" frequencies, no?

I would expect something like:

```go
        cpuFreqMHz := uint64(CPUFrequency()/MHz)
	top := p.getWrap()
	phc := p.getPhaseCorrect()
	Int, frac := p.getClockDiv()
	// Line below can overflow if operations done without care.
	return (16*uint64(Int) + uint64(frac)) * uint64((top+1)*(phc+1))*1e3 / (cpuFreqMHz*16) // cycles = (TOP+1) * (CSRPHCorrect + 1) * (DIV_INT + DIV_FRAC/16)
```

amken3d: I will take a look at it again. 
My simplistic solution was that if you cared about "uneven" frequency, just change it to one where it is even. I provided the following carefully selected options
[50,100,125,133,150,175,200,225,240,250,275,300]
But to your point, there are some on this list which are not factors of 2

eliasnaur: I see. To be clear, the current frequency, 125MHz, that both rp2040 and rp2350 run on work fine with `cpuPeriod`. I bring this is up now because you're in the area, and I'd like to bump both variants to their datasheet maximum (133MHz and 150MHz) in a future PR. I don't want to use frequencies outside datasheet guaranteed frequencies.

amken3d: 133 and 150 are for the reference design of the Pico. Anyway the datasheet says  "Most chips in most normal
environments can run significantly faster than the quoted maximum, and therefore support overclocking" pg 512.  The speeds that they choose is a function of the crystal that they used. In my boards, I use a much higher quality cmos oscillator at 40mhz, which i share with the CAN controller as well, which has error in the <10ppm range. In practical applications, even on the pico, you can run 300 mhz without the chip complaining or overheating.  I have run rp2040 at 500Mhz with a higher core voltage. The datasheet even tells you how to do it.

eliasnaur: Sure, I'm not saying overclocking is impossible or undesirable. I'm only saying that the PWM divider calculations should handle frequencies that don't nicely divide 1e9, as `cpuPeriod` assumes.

amken3d: Yes. Agreed. I will do that. Thanks for reminding.

deadprogram:
The build failures are related to missing pin mappings when building `gopher_badge` for WASM target e.g. for the TinyGo playground.

src/machine/board_ae_rp2040.go
```diff
@@ -2,6 +2,11 @@
 
 package machine
 
+func init() {
+	SysClockFrequency = Freq133MHz
```
soypat, line 6: This exposes new API on machine package. Why would we want to expose a variable that can be modified by users? Is CPUFrequency() not good enough for your applications?

src/machine/board_ae_rp2040.go
```diff
@@ -2,6 +2,11 @@
 
 package machine
 
+func init() {
```
soypat, line 5: Avoid init functions in machine package always unless absolutely required. I'd like to understand the problem you are trying to solve so I can propose a better way of doing this.

soypat:
@amken3d  Please take a look at my propsed strategy for adding the RP2350B here: https://github.com/tinygo-org/tinygo/pull/4728

My ultimate motivation for this PR I'm showing you is avoiding adding extraneous complexity to the machine package. I want to keep this maintainable. Thank you for understanding :)

amken3d:
@soypat , Thanks for the review and PR. Your changes for the ADC looks fine to me.  My PR does include other things as well. Do you plan to address those as well?

src/machine/board_ae_rp2040.go
amken3d: CPUFrequency() only gave me what the current  sysClk frequency was and not the ability to overclock or under clock. There are several use cases for being able to change system frequency at runtime like peri synchronization, power management etc. This change makes changing clock speeds on the fly possible.

eliasnaur: I don't think changing clock frequency at runtime is in scope of this PR. If for nothing else, runtime clock support doesn't answer the question of what to do about all the code that assumes `CPUFrequency` never changes.

deadprogram: I would agree that any changes about clock frequency should be in a separate PR.

deadprogram:
@amken3d @soypat seems from a superficial overview that since the pins and ADC are addressed already in PR #4728 the only thing really "missing" is the PWM support, which would appear to be similar to the ADC changes? Probably another commit to #4728 would do it?

After that, adding the new board support could be done in a separate PR @amken3d 

Then any changes to clock speed can be discussed in yet another separate PR/issue. @cibomahto I think had some ideas on how to do this, before it got pulled into a bigger set of objectives and dropped.

amken3d:
@deadprogram , thanks for the guidance. I will close this PR for now.

soypat:
Got intellisense back working. Will try to get around to porting your changes @amken3d. Thank you so much for the work you put in :D

PR #4728, opened 2025-02-10
Rp2350b implementation

soypat:
This simplifies changes proposed by Amken to fewer files and even simplifies the ADC internals of the machine package at the cost of some safety, which honestly does not seem like such a big loss compared to the complexity that could be added otherwise. 

I really did not have much time to prepare this commit nor did I have the help of intellisense, my environment is broken since tinygo extension mysteriously dissapeared from VSCode extension list. Still, I think this is the direction worth exploring via build tags. I still do not understand some of the changes proposed, namely those regarding init functions.

### DIFF
--- a/src/machine/machine_rp2_pins.go
+++ b/src/machine/machine_rp2_pins.go
@@ -1,4 +1,4 @@
-//go:build rp2040 || rp2350 || ae_rp2040 || badger2040 || challenger_rp2040 || feather_rp2040 || gopher_badge || kb2040 || macropad_rp2040 || nano_rp2040 || pico || qtpy_rp2040 || thingplus_rp2040 || thumby || tufty2040 || waveshare_rp2040_zero || xiao_rp2040
+//go:build rp2040 || rp2350 || gopher_badge
 
 package machine
 
@@ -34,10 +34,4 @@ const (
 	GPIO27 Pin = 27 // peripherals: PWM5 channel B
 	GPIO28 Pin = 28 // peripherals: PWM6 channel A
 	GPIO29 Pin = 29 // peripherals: PWM6 channel B
-
-	// Analog pins
-	ADC0 Pin = GPIO26
-	ADC1 Pin = GPIO27
-	ADC2 Pin = GPIO28
-	ADC3 Pin = GPIO29
 )
--- a/src/machine/machine_rp2_pll.go
+++ b/src/machine/machine_rp2_pll.go
@@ -4,6 +4,9 @@ package machine
 
 import (
 	"device/rp"
+	"errors"
+	"math"
+	"math/bits"
 	"runtime/volatile"
 	"unsafe"
 )
@@ -29,12 +32,11 @@ var (
 // Post Divider 1, postDiv1 with range 1-7 and be >= postDiv2.
 //
 // Post Divider 2, postDiv2 with range 1-7.
-func (pll *pll) init(refdiv, vcoFreq, postDiv1, postDiv2 uint32) {
+func (pll *pll) init(refdiv, fbdiv, postDiv1, postDiv2 uint32) {
 	refFreq := xoscFreq / refdiv
 
 	// What are we multiplying the reference clock by to get the vco freq
 	// (The regs are called div, because you divide the vco output and compare it to the refclk)
-	fbdiv := vcoFreq / (refFreq * MHz)
 
 	// Check fbdiv range
 	if !(fbdiv >= 16 && fbdiv <= 320) {
@@ -54,13 +56,14 @@ func (pll *pll) init(refdiv, vcoFreq, postDiv1, postDiv2 uint32) {
 		panic("postdiv1 should be greater than or equal to postdiv2")
 	}
 
-	// Check that reference frequency is no greater than vco / 16
+	// Check that reference frequency is no greater than vcoFreq / 16
+	vcoFreq := calcVCO(xoscFreq, fbdiv, refdiv)
 	if refFreq > vcoFreq/16 {
 		panic("reference frequency should not be greater than vco frequency divided by 16")
 	}
 
 	// div1 feeds into div2 so if div1 is 5 and div2 is 2 then you get a divide by 10
-	pdiv := postDiv1<<rp.PLL_SYS_PRIM_POSTDIV1_Pos | postDiv2<<rp.PLL_SYS_PRIM_POSTDIV2_Pos
+	pdiv := uint32(postDiv1)<<rp.PLL_SYS_PRIM_POSTDIV1_Pos | uint32(postDiv2)<<rp.PLL_SYS_PRIM_POSTDIV2_Pos
 
 	if pll.cs.HasBits(rp.PLL_SYS_CS_LOCK) &&
 		refdiv == pll.cs.Get()&rp.PLL_SYS_CS_REFDIV_Msk &&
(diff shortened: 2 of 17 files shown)


soypat:
@amken3d @deadprogram @eliasnaur  Still working on it. Added PWM and GPIO support originally implemented by amken3d. I did it in a way to maximize the amount of code reused from the RP2040. Please take note that this is a design decision that takes into account the cost of maintaining these APIs. 

I want the code to be as little fragmented as possible between RP2 systems. It seems like 100% of PWM code was reusable. Had to resort to some tricks with the is48Pin package constant which should avoid compiling the extra code on RP2040 platforms, but we'll see soon enough.

If someone can please test out the new PWM period calculation... I didn't really validate the calculations...

There's a lot of crystal oscillator code I did NOT port because I don't quite understand what it was going for.

src/machine/machine_rp2_pwm.go
```diff
@@ -146,12 +146,13 @@ func (p *pwmGroup) Counter() uint32 {
...
 func (p *pwmGroup) Period() uint64 {
-	periodPerCycle := cpuPeriod()
+	freq := CPUFrequency()
 	top := p.getWrap()
 	phc := p.getPhaseCorrect()
 	Int, frac := p.getClockDiv()
-	// Line below can overflow if operations done without care.
-	return (16*uint64(Int) + uint64(frac)) * uint64((top+1)*(phc+1)*periodPerCycle) / 16 // cycles = (TOP+1) * (CSRPHCorrect + 1) * (DIV_INT + DIV_FRAC/16)
+	// Lines below can overflow if operations done without care.
+	term2 := 16 * uint64((top+1)*(phc+1)) * 1e9 / uint64(freq) // 1e9/freq == CPU period in nanoseconds.
```
eliasnaur: Can 16e9*(top+1)*(phc+1) overflow uint64? If so, you could divide freq by 1e3 or 1e6 to compensate.

Here and below.

src/machine/machine_rp2_pins.go
```diff
@@ -1,4 +1,4 @@
-//go:build rp2040 || rp2350 || ae_rp2040 || badger2040 || challenger_rp2040 || feather_rp2040 || gopher_badge || kb2040 || macropad_rp2040 || nano_rp2040 || pico || qtpy_rp2040 || thingplus_rp2040 || thumby || tufty2040 || waveshare_rp2040_zero || xiao_rp2040
+//go:build rp2040 || rp2350
```
deadprogram:
```suggestion
//go:build rp2040 || rp2350 || gopher_badge
```

This needs to have the `gopher_badge` tag included in order to build the WASM code used for the TinyGo Playground. See build error at https://github.com/tinygo-org/tinygo/actions/runs/13277344229/job/37069264325?pr=4728#step:22:151

src/machine/machine_rp2_pwm.go
soypat: I just realized there's a bug in the calculation- I extracted the 16 from Int,Frac multiplication without dividing Frac. 

I also realize on closer inspection we might be able to get away by just multiplying the 1e9 up front:

OK, took a look at the datasheet and this is what I could grep: 
`maxtop=maxuint16, maxphc=1, maxInt=255, maxFrac=15` => `maxNumerator = (16*255+15)*65536*2*1e9=5.4e17` whereas `maxuint64=2.1e19`
 
Would seem we are fine?

soypat: If implemented this way it would leave us with maximum resolution of period calculation :D

deadprogram:
What else do we need for this PR? An actual board/smoke test, correct?

soypat:
Alright, ready for testing I believe :)

src/machine/machine_rp2_pll.go
```diff
@@ -98,3 +101,179 @@ func (pll *pll) init(refdiv, vcoFreq, postDiv1, postDiv2 uint32) {
...
+//
+// Example for 12MHz crystal and RP2350:
+//
+//	fbdiv, refdiv, pd1, pd2, _ := pllSearch{LockRefDiv:1}.CalcDivs(12*MHz, 133*MHz, MHz)
+type pllSearch struct {
+	LowerVCO   bool
+	LockRefDiv uint8
+}
+
+func (ps pllSearch) CalcDivs(xoscRef, targetFreq, MHz uint64) (fbdiv uint64, refdiv, pd1, pd2 uint8, err error) {
```
soypat, line 120: I tried to add a function that dynamically calculates the PLL dividers and optimized enough so that maybe  TinyGo interp would kick in and solve for the dividers at compile time with the XoscFreq and target CPU frequency as solving parameters but it seems I was unsucessful as far as I could tell.

deadprogram: This function is not yet in use in this PR it would seem? How did you test it, @soypat

deadprogram:
> Alright, ready for testing I believe :)

Tested on RP2350A Pico2 and RP2350B Pico Plus2 boards with both `blinky1` and also `pwm` examples, and they are both worked as expected.

Also tested with both of those examples on RP2040 Pico board just to check for regressions, and is working as expected.

src/machine/machine_rp2_pwm.go
```diff
@@ -266,6 +268,9 @@ func (pwm *pwmGroup) setPeriod(period uint64) error {
 		// Maximum Period is 268369920ns on rp2040, given by (16*255+15)*8*(1+0xffff)*(1+1)/16
 		// With no phase shift max period is half of this value.
 		maxPeriod = 268 * milliseconds
+		// This will be a compile time error if this method is at risk of overflowing. cpufreq=155MHz for typical RP2350.
```
eliasnaur, line 271: Nit: did you mean "150MHz for typical lrp2350"?

amken3d:
> @amken3d @deadprogram @eliasnaur Still working on it. Added PWM and GPIO support originally implemented by amken3d. I did it in a way to maximize the amount of code reused from the RP2040. Please take note that this is a design decision that takes into account the cost of maintaining these APIs.
> 
> I want the code to be as little fragmented as possible between RP2 systems. It seems like 100% of PWM code was reusable. Had to resort to some tricks with the is48Pin package constant which should avoid compiling the extra code on RP2040 platforms, but we'll see soon enough.
> 
> If someone can please test out the new PWM period calculation... I didn't really validate the calculations...
> 
> There's a lot of crystal oscillator code I did NOT port because I don't quite understand what it was going for.

@soypat  Looks like you figured out most of it. 
Couple of things for your consideration
1. The RP2350 bumped its max crystal freq to 50 MHZ  (from 15 for rp2040)
     In my hardware designs, I am going for a 40Mhz CMOS oscillator instead of 12Mhz crystal, what you typically see. Primarily for BOM consolidation as well as better stability than a 12 Mhz crystal. If there was flexibility, then other board makers will choose different setups based on different usecases.
2. Did not see a way to set the target frequency at build-time or runtime (I would prefer runtime, but I will leave it up to the maintainers). 
 I see being able to set target frequency at runtime as an advantage for low power usecases or for getting more performance by overclocking. The 133Mhz and 150Mhz set by the ref design is just a ultra conservative limit. I would argue that as long as it is within the vcocalc boundaries, it is not even considered as overclocking. Overclocking would be when you raise the core voltage, which tinygo does not yet do.
 @eliasnaur , has brought up that  there is some dependency in other parts of the code (tinygo drivers) on a fixed stable frequency. My request to the maintainers is to revisit that and consider the flexibility that runtime settable freq would bring to the table.

eliasnaur:
>     2. Did not see a way to set the target frequency at build-time or runtime (I would prefer runtime, but I will leave it up to the maintainers).
>        I see being able to set target frequency at runtime as an advantage for low power usecases or for getting more performance by overclocking. The 133Mhz and 150Mhz set by the ref design is just a ultra conservative limit. I would argue that as long as it is within the vcocalc boundaries, it is not even considered as overclocking. Overclocking would be when you raise the core voltage, which tinygo does not yet do.
>        @eliasnaur , has brought up that  there is some dependency in other parts of the code (tinygo drivers) on a fixed stable frequency. My request to the maintainers is to revisit that and consider the flexibility that runtime settable freq would bring to the table.

A configurable clock speed is probably out of scope of this PR. Nevertheless, allow me to bring up a few points:

- Even if `delay.Sleep` didn't require a (Go) constant `CPUFrequency` (say, integer division is fast enough on larger cores), how do you notify every driver about a frequency change? Drivers typically query `CPUFrequency` at initialization and then assume it (value) constant from then on.
- Can a runtime configurable frequency even save that much power? It's my (naive?) understanding that modern cores save power by shutting off hardware units and (deep) sleeping.

(Sorry for derailing the discussion of this PR)

deadprogram:
> A configurable clock speed is probably out of scope of this PR. 

Agreed. Also, IMO not needed at runtime. Being able to set different clock speed at compile-time would be sufficient. In any case, that would be for a future PR.

> (Sorry for derailing the discussion of this PR)

Not at all, this is a good conversation to refer back to for a follow-on PR.

deadprogram:
Rebased against latest `dev`

src/machine/machine_rp2_pll.go
soypat: I can hardly say I tested it. I compared the results of this function with the ones of vcocalc.py for a few inputs and saw they matched. I left it in in case someone wanted to try to use it in the future seeing how many PRs included xosc changes for the rp2350 recently.

soypat: Do we have a "scripts" place in tinygo main repo we could put this in and write tests for?

deadprogram: > Do we have a "scripts" place in tinygo main repo we could put this in and write tests for?

We have the `tests` directory which is where all of the various testing has gone.

I would suggest we remove that code from this PR, and add it instead to a new draft PR that can be worked on further as part of implementing speed changes.

deadprogram: @soypat what do you think?

eliasnaur: I agree that `CalcDivs` is useful code, in particular for future clock changes. How about moving the code to a `pll_test.go` file and add a test that verifies that the (otherwise magic) pll constants match the output of `CalcDivs`? A future PR would then change, say, `targetFreq`, get a failing test and adjust the other constants according to `CalcDivs`.

deadprogram: @eliasnaur that sounds like a very cool idea. Perhaps you could create a PR off of this one to make that change, if you think you see a straightforward way to do it?

eliasnaur: I don't think it's that complicated, but sure, if @soypat doesn't do it as part of this PR, I can can do it after merge.

deadprogram:
OK in the interest of moving this PR forward, I have squashed some of the commits and rebased against `dev`.

Once it passes CI I intend to merge it, then further changes can take place against the `dev` branch.

deadprogram:
Now merging, thank you @soypat @amken3d @eliasnaur for working on this to get it to the current state.

PR #4730, opened 2025-02-11
targets: turn on GC for TKey1 device

deadprogram:
This PR turns on GC for the TKey1 device target, since it does in fact work.

Also very much needed, since many crypto packages require heap allocations.


deadprogram:
Merging since no opposition to this change. :smile_cat:

PR #4731, opened 2025-02-12
build: update Linux builds to run on ubuntu-latest since 20.04 is being retired

deadprogram:
This PR updates Linux CI builds to run on `ubuntu-latest` since "The Ubuntu 20.04 runner image will be fully unsupported by April 1, 2025"

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -169,7 +169,7 @@ jobs:
   assert-test-linux:
     # Run all tests that can run on Linux, with LLVM assertions enabled to catch
     # potential bugs.
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -297,7 +297,7 @@ jobs:
           - goarch: arm
             toolchain: arm-linux-gnueabihf
             libc: armhf
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     needs: build-linux
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


deadprogram:
This PR also adds commit removing sandbox from headless chrome used for WASM testing, which is now a requirement for Ubuntu.

ydnar (approved):

.github/workflows/linux.yml
```diff
@@ -169,7 +169,7 @@ jobs:
   assert-test-linux:
     # Run all tests that can run on Linux, with LLVM assertions enabled to catch
     # potential bugs.
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
```
aykevl, line 172: :+1:

deadprogram:
Thanks for review @ydnar now merging.

.github/workflows/linux.yml
```diff
@@ -297,7 +297,7 @@ jobs:
           - goarch: arm
             toolchain: arm-linux-gnueabihf
             libc: armhf
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
aykevl, line 300: I specifically picked an older version for wider compatibility. I'm not sure TinyGo binaries built on ubuntu-latest will run on ubuntu-22.04 for example (or any random old Raspberry Pi install).

aykevl:
> This PR also adds commit removing sandbox from headless chrome used for WASM testing, which is now a requirement for Ubuntu.

Sad, but there's no real security impact for us of course. It's all just tests.

.github/workflows/linux.yml
deadprogram: Hmm, that is possible. Should we leave it how it is now, since I have merged it, or should we add a new PR that downgrades?

aykevl: See: https://github.com/tinygo-org/tinygo/pull/4753

PR #4733, opened 2025-02-12
ci: upgrade to Ubuntu 22.04

aykevl:
Ubuntu 20.04 got deprecated, but we want to keep using an older version for the widest compatibility. So we'll switch to Ubuntu 22.04 instead.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -297,7 +297,7 @@ jobs:
           - goarch: arm
             toolchain: arm-linux-gnueabihf
             libc: armhf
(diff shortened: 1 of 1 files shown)


deadprogram:
Ahem #4731

deadprogram:
We will need https://github.com/tinygo-org/tinygo/pull/4731/commits/e0c2fda1eb9be5c6f8a36ce8fc2befbb9f17ba39 at some point sooner or later, I think.

aykevl:
Oops, didn't see that one

Issue #4726, opened 2025-02-09
TinyGo `v0.35.0` breaks convention on Command modules for exported functions


inliquid:
Until very recently TinyGo had no support for compiling Reactor modules which have `_initialize` function as opposed to `_start` in Command modules. This, as well as some other reasons lead to community convention which worked for some years. In short according to WASI, Commands should exit after `_start` finished and as a result module with its exported functions becomes invalid.

However, TinyGo as well as some other compilers did not exit, which made it possible to call exported functions even after `_start` exited.

For instance Wazero, the most popular Wasm runtime in Go world has special [note](https://github.com/tetratelabs/wazero/blob/main/RATIONALE.md#why-do-we-only-return-a-sysexiterror-on-a-non-zero-exit-code) in documentation on this:

> ### Why do we only return a `sys.ExitError` on a non-zero exit code?
> 
> It is reasonable to think an exit error should be returned, even if the code is success (zero). Even on success, the module is no longer functional. For example, function exports would error later. However, wazero does not. The only time `sys.ExitError` is on error (non-zero).
> 
> This decision was to improve performance and ergonomics for guests that both use WASI (have a `_start` function), and also allow custom exports. Specifically, Rust, TinyGo and normal wasi-libc, don't exit the module during `_start`. If they did, it would invalidate their function exports. This means it is unlikely most compilers will change this behavior.
>
> `GOOS=waspi1` from Go 1.21 does exit during `_start`. However, it doesn't support other exports besides `_start`, and `_start` is not defined to be called multiple times anyway.
>
> Since `sys.ExitError` is not always returned, we added Module.IsClosed for defensive checks. This helps integrators avoid calling functions which will always fail.

There are well-established and popular tools and libraries in the wild which are built on top of this assumption, for instance https://github.com/knqyf263/go-plugin which generates helper stubs based on `*.proto` definitions. Generated code is based heavily on existing tools and their proven behavior, such as Wazero and TinyGo (pre-v0.35.0). 

And a lot of tooling and code (including my own and the company I work in) is already in production.

This is good post by Matt Johnson-Pint in [Slack](https://gophers.slack.com/archives/C040AKTNTE0/p1734994475553679) which summarizes what's happened:

> Hi folks.  I've been trying out the recently released TinyGo 0.35.0 with Wazero and ran into some gotchas.  Thought I would share my findings:
> - There's apparently two types of WASI programs (in WASI p1) - "commands" and "reactors".
[...]

PR #4734, opened 2025-02-12
feature: add buildmode=wasi-legacy for wasi modules

deadprogram:
This PR adds a `-buildmode=wasi-legacy` flag to support existing base of users who expected the older behavior for wasi modules to not return an exit code as if they were reactors.

See #4726 for some details on what this is intended to address.

To try it out:

```shell
tinygo build -o plugin.wasm -scheduler=none -target=wasi -buildmode=wasi-legacy plugin.go
```

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -667,6 +667,16 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 		ldflags = append(ldflags, "--no-entry")
 	}
 
+	if config.Options.BuildMode == "wasi-legacy" {
+		if !strings.HasPrefix(config.Triple(), "wasm32-") {
+			return result, fmt.Errorf("buildmode wasi-legacy is only supported on wasm")
+		}
+
+		if config.Options.Scheduler != "none" {
+			return result, fmt.Errorf("buildmode wasi-legacy only supports scheduler=none")
+		}
+	}
+
 	// Add compiler-rt dependency if needed. Usually this is a simple load from
 	// a cache.
 	if config.Target.RTLib == "compiler-rt" {
--- a/compileopts/options.go
+++ b/compileopts/options.go
@@ -8,7 +8,7 @@ import (
 )
 
 var (
-	validBuildModeOptions     = []string{"default", "c-shared"}
+	validBuildModeOptions     = []string{"default", "c-shared", "wasi-legacy"}
 	validGCOptions            = []string{"none", "leaking", "conservative", "custom", "precise"}
 	validSchedulerOptions     = []string{"none", "tasks", "asyncify"}
 	validSerialOptions        = []string{"none", "uart", "usb", "rtt"}
--- a/compiler/symbol.go
+++ b/compiler/symbol.go
@@ -283,6 +283,11 @@ func (c *compilerContext) getFunctionInfo(f *ssa.Function) functionInfo {
 		info.wasmName = "_start"
 		info.exported = true
 	}
+	if info.linkName == "runtime.wasmEntryLegacy" && c.BuildMode == "wasi-legacy" {
+		info.linkName = "_start"
+		info.wasmName = "_start"
+		info.exported = true
+	}
 
 	// Check for //go: pragmas, which may change the link name (among others).
 	c.parsePragmas(&info, f)
--- a/main.go
+++ b/main.go
@@ -1501,7 +1501,7 @@ func main() {
 	var tags buildutil.TagsFlag
 	flag.Var(&tags, "tags", "a space-separated list of extra build tags")
 	target := flag.String("target", "", "chip/board name or JSON target specification file")
-	buildMode := flag.String("buildmode", "", "build mode to use (default, c-shared)")
+	buildMode := flag.String("buildmode", "", "build mode to use (default, c-shared, wasi-legacy)")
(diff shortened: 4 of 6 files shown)


deadprogram:
Tested with this branch on the https://github.com/knqyf263/go-plugin repo.

After making this change:

```
diff --git a/Makefile b/Makefile
index 6cb7303..e508516 100644
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ tinygo_tests := $(shell find tests -path "*/plugin*/*.go")
 build.tests: $(tinygo_tests:.go=.wasm)
 
 %.wasm: %.go $(GOBIN)/protoc-gen-go-plugin
-       tinygo build -o $@ -scheduler=none --no-debug --target=wasi $<
+       tinygo build -o $@ -buildmode=legacy -scheduler=none --no-debug --target=wasi $<
 
 proto_files := $(shell find . -name "*.proto")
 .PHONY: protoc
```

All of the tests in that repo pass with this change, and fail without it. So it does appear to solve the issue.

deadprogram:
Setting to ready for review, since I have now added a test, and also verified using the https://github.com/knqyf263/go-plugin tests that it corrects the issue.

aykevl (commented):
Hmm. Not very happy with this, but I can see why it is needed. I'm of the opinion that users should switch to `-buildmode=c-shared` eventually, since that is specified (and our old behavior was not).

Perhaps we can rename the build mode to something else, like `-buildmode=wasi-command-legacy` to make it very clear that this is only for WASI? Other than that, LGTM.

deadprogram:
>  I'm of the opinion that users should switch to -buildmode=c-shared eventually, since that is specified (and our old behavior was not).

Agreed, hopefully that will happen pretty quickly, but in the meantime this should help those who cannot easily change their code.

> Perhaps we can rename the build mode to something else, like -buildmode=wasi-command-legacy to make it very clear that this is only for WASI? 

Does making it longer make it clearer? I tried to pick the shortest option.

deadprogram:
Changed flag to `-buildmode=wasi-legacy` to address feedback but still keep it short.

deadprogram:
Any feedback before I merge this from @ydnar @dgryski or anyone else?

ydnar:
I'm not sure this is the right implementation. It doesn't operate the scheduler like buildmode=reactor, so goroutines will not execute.

Is the issue just mainExited?

This also presumes that other projects would change -buildmode. Could they just use reactor?

deadprogram:
> I'm not sure this is the right implementation. It doesn't operate the scheduler like buildmode=reactor, so goroutines will not execute.

The way this is intended to be used is without a scheduler, e.g. 

```
tinygo build -o plugin.wasm -scheduler=none -target=wasi -buildmode=wasi-legacy plugin.go
```

For projects such as https://github.com/knqyf263/go-plugin

> Is the issue just mainExited?

So it seems from my testing with go-plugin see https://github.com/tinygo-org/tinygo/pull/4734#issuecomment-2654635530

> This also presumes that other projects would change -buildmode. Could they just use reactor?

The proposed change here allows the existing generated code from projects such as go-plugin to recompile with the latest tinygo code without regenerating, just adding the flag.

The use case here is for existing users of wasi that expect the older behavior.

deadprogram:
Another project that needs this is Trivy.

See https://github.com/aquasecurity/trivy/blob/main/pkg/module/module.go#L316

deadprogram:
I added a check to ensure that `scheduler=none` is set when trying to use `buildmode=wasi-legacy`.

deadprogram:
@ydnar please note the extra check I added to ensure that `scheduler=none` is set when trying to use `buildmode=wasi-legacy`.

https://github.com/tinygo-org/tinygo/pull/4734/files#diff-d871f892c3b7d415467578ac1742cccab676ec4cc21022b0b9b34f5741237eb7R675

deadprogram:
Now merging this PR since it does correct the issue it is intended to.

Thank you for the feedback everyone, and thanks to @inliquid and others who provided a very complete description of the situation.

builder/build.go
```diff
@@ -667,6 +667,16 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
...
 	}
 
+	if config.Options.BuildMode == "wasi-legacy" {
+		if !strings.HasPrefix(config.Triple(), "wasm32-") {
+			return result, fmt.Errorf("buildmode wasi-legacy is only supported on wasm")
+		}
+
+		if config.Options.Scheduler != "none" {
+			return result, fmt.Errorf("buildmode wasi-legacy only supports scheduler=none")
+		}
```
aykevl, line 677: Slightly better would be:

```go
if config.Scheduler() != "none"
```

The reason is that for example `-target=wasm-unknown` already defaults to scheduler "none", so the explicit command line flag (found via `config.Options.Scheduler`) isn't needed.

aykevl (commented):
@deadprogram
Didn't want to hold up this PR on a naming issue, wasi-legacy seems fine. It makes it clear that this is a special build mode just for WASI.

@ydnar 

> I'm not sure this is the right implementation. It doesn't operate the scheduler like buildmode=reactor, so goroutines will not execute.

It does, the `run` function runs the scheduler. So I don't think we need to limit this to just the "none" scheduler.

PR #4735, opened 2025-02-13
target: add Pimoroni Pico Plus2

deadprogram:
This PR is to complete #4728 by adding an actual board that uses the rp2350b.

Note the commit that was needed to separate the ADC pin defs for rp2350a vs. rp2350b. Small difference, but needed to compile.


deadprogram:
Fixed formatting. @soypat this is ready.

PR #4737, opened 2025-02-14
os/file: add File.Chmod

leongross:
Add `file.Chmod` to `os/file`.


soypat (approved):
Not a fan of naked return but alas looks ok to me

deadprogram:
Thanks for adding this @leongross and to @soypat for review. Now merging.

leongross:
Thanks @soypat for the review and @deadprogram for the merge.

PR #4732, opened 2025-02-12
all: add Go 1.24 support

aykevl:
Not entirely finished, but let's see how far we get with just a few adjustments.

PR #4742, opened 2025-02-18
interp: correctly mark functions as modifying memory

aykevl:
This is a fix for an interp bug I found in #4732.
Making a draft PR since I want to see the impact of this, it could be big. I might need to modify the PR.

### DIFF
--- a/builder/sizes_test.go
+++ b/builder/sizes_test.go
@@ -43,8 +43,8 @@ func TestBinarySize(t *testing.T) {
 	tests := []sizeTest{
 		// microcontrollers
 		{"hifive1b", "examples/echo", 4560, 280, 0, 2268},
-		{"microbit", "examples/serial", 2908, 388, 8, 2272},
-		{"wioterminal", "examples/pininterrupt", 7293, 1487, 116, 6912},
+		{"microbit", "examples/serial", 2916, 388, 8, 2272},
+		{"wioterminal", "examples/pininterrupt", 7315, 1489, 116, 6912},
 
 		// TODO: also check wasm. Right now this is difficult, because
 		// wasm binaries are run through wasm-opt and therefore the
--- a/interp/interpreter.go
+++ b/interp/interpreter.go
@@ -970,9 +970,9 @@ func (r *runner) runAtRuntime(fn *function, inst instruction, locals []value, me
 	case llvm.Call:
 		llvmFn := operands[len(operands)-1]
 		args := operands[:len(operands)-1]
-		for _, arg := range args {
-			if arg.Type().TypeKind() == llvm.PointerTypeKind {
-				err := mem.markExternalStore(arg)
+		for _, op := range operands {
+			if op.Type().TypeKind() == llvm.PointerTypeKind {
+				err := mem.markExternalStore(op)
 				if err != nil {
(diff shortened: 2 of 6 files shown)


deadprogram:
```
--- FAIL: TestBinarySize (0.00s)
    --- FAIL: TestBinarySize/microbit/examples/serial (40.40s)
        sizes_test.go:67: Unexpected code size when compiling: -target=microbit examples/serial
        sizes_test.go:68:             code rodata   data    bss
        sizes_test.go:69: expected:   2908    388      8   [22](https://github.com/tinygo-org/tinygo/actions/runs/13391197573/job/37399093533?pr=4742#step:15:23)72
        sizes_test.go:70: actual:     4352    404    176   2552
    --- FAIL: TestBinarySize/wioterminal/examples/pininterrupt (41.07s)
        sizes_test.go:67: Unexpected code size when compiling: -target=wioterminal examples/pininterrupt
        sizes_test.go:68:             code rodata   data    bss
        sizes_test.go:69: expected:   7293   1487    116   6912
        sizes_test.go:70: actual:     8691   1501    [28](https://github.com/tinygo-org/tinygo/actions/runs/13391197573/job/37399093533?pr=4742#step:15:29)8   7200
FAIL
```

Impact was not small on your first attempt, it would seem. :elephant:

aykevl:
Oops, yeah, I was afraid of this.
I'll need to investigate this a bit deeper.

aykevl:
Trying a workaround that should hopefully fix the worst of this. Let's see what happens.

aykevl:
So this is fixing a real bug, but also uncovering a bunch of cases where the old behavior was incorrect but resulted in smaller binary sizes. Most importantly (at least for small programs), the `runtime.fastrand` function was previously just returning 0 during interp since `xorshift32State` wasn't initialized but interp (incorrectly) continued interpreting it anyway. This bugfix fixes this issue - but also means that callers of fastrand (most notably the hashmap code) now can't be run during interp.

I've added a fix that should fix the worst of this, by initializing the xorshift state to a valid (non-random) value so that fastrand actually returns pseudo-random numbers that can be used by interp while still making fastrand use hardware random numbers at runtime.

@dgryski you wrote some of this, what do you think of this solution?

deadprogram:
@aykevl I think we can afford the extra 2 bytes here https://github.com/tinygo-org/tinygo/actions/runs/13415479905/job/37475187142?pr=4742#step:19:16 :smiley_cat:

aykevl:
Oops, yeah I missed that one.

deadprogram:
@aykevl looking at the sizediff, it would appear that this PR is a success.

dgryski:
I have the feeling this will fix a number of other interp-related issues on our bug tracker.

deadprogram (approved):
Thanks for working on this @aykevl

deadprogram:
I will merge after the current build finishes.

deadprogram:
Now merging, thanks @aykevl

dgryski:
I tracked down a regression that was preventing an internal service from building with recent releases of tinygo to this.  This PR causes the `asyncify` pass to take 10x longer for a 10% binary increase size, and the `precompute-propagate` to basically never finish.

PR #4744, opened 2025-02-20
fix: better handling for wasm JS finalizer

deadprogram:
This PR is to complete the work done by @ldemailly in https://github.com/tinygo-org/tinygo/pull/4352 along with suggestion from @prochac

### DIFF
--- a/targets/wasm_exec.js
+++ b/targets/wasm_exec.js
@@ -305,8 +305,7 @@
 					"syscall/js.finalizeRef": (v_ref) => {
 						// Note: TinyGo does not support finalizers so this is only called
 						// for one specific case, by js.go:jsString. and can/might leak memory.
-						const id = mem().getUint32(unboxValue(v_ref), true);
-						// Note that this if is so far seemingly never true. Someone should investigate why.
+						const id = v_ref & 0xffffffffn;
 						if (this._goRefCounts?.[id] !== undefined) {
 							this._goRefCounts[id]--;
 							if (this._goRefCounts[id] === 0) {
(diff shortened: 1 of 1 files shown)


targets/wasm_exec.js
```diff
@@ -303,9 +303,21 @@
...
+							this._goRefCounts[id]--;
+							if (this._goRefCounts[id] === 0) {
+								const v = this._values[id];
+								this._values[id] = null;
+								this._ids.delete(v);
+								this._idPool.push(id);
+							}
+						} else {
+							// Log as a hint and reminder that something is probably off.
+							console.log("syscall/js.finalizeRef: unknown id", id);
```
prochac: So you go with console.log, not console.error?

deadprogram: This was in response to @ldemailly trying to reduce the number of errors in the browser console. However now that it is working more correctly, perhaps it can now actually use `console.warning` as had been suggested.

What do you think @prochac about that? Also @ribasushi could you please try that out with your test setup?

ribasushi: @deadprogram with https://github.com/tinygo-org/tinygo/blob/94e7fd80e0f9102e67b5e9a9cd15abafb8eabe63/targets/wasm_exec.js I get no console output, and things seem to function well

I concur: `.warning` or perhaps even leaving as `.error` is warranted.

prochac: @deadprogram It shouldn't happen, so error from my point of view. If it occurs, the error will ideally motivate someone to open new issue.

deadprogram: OK, the people have spoken: `console.error` it is.

Changed and updated this PR.

ribasushi: @deadprogram note this also fixes (not currently linked) https://github.com/tinygo-org/tinygo/issues/1140

deadprogram:
OK at long last! Thank you to @ldemailly for starting this off, to @prochac for the final twist, and to @ribasushi for helping test/review and get it landed.

Now merging.

PR #4746, opened 2025-02-21
fix: infer target for wasm when GOOS and GOARCH are set

deadprogram:
This PR fixes the `tinygo build` command to only infer `target` for wasm targets when `GOOS` and `GOARCH` are set correctly, not just based on file extension.

This is intended to be a better version of #4447 by handling all of the current wasm ENV var variations for highest compatibility with big Go.

### DIFF
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -356,17 +356,7 @@ func defaultTarget(options *Options) (*TargetSpec, error) {
 			return nil, fmt.Errorf("invalid GOMIPS=%s: must be hardfloat or softfloat", options.GOMIPS)
 		}
 	case "wasm":
-		llvmarch = "wasm32"
-		spec.CPU = "generic"
-		spec.Features = "+bulk-memory,+mutable-globals,+nontrapping-fptoint,+sign-ext,-multivalue,-reference-types"
-		spec.BuildTags = append(spec.BuildTags, "tinygo.wasm")
-		spec.CFlags = append(spec.CFlags,
-			"-mbulk-memory",
-			"-mnontrapping-fptoint",
-			"-mno-multivalue",
-			"-mno-reference-types",
-			"-msign-ext",
-		)
+		return nil, fmt.Errorf("GOARCH=wasm but GOOS is unset. Please set GOOS to wasm, wasip1, or wasip2.")
 	default:
 		return nil, fmt.Errorf("unknown GOARCH=%s", options.GOARCH)
 	}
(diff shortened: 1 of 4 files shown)


deadprogram:
I have added a third commit to this PR that removes the separate code for WASM compile options now that the target files are the authoritative source.

ydnar:
Nice, thank you!

deadprogram:
Any other feedback @aykevl @dgryski anyone else?

main.go
```diff
@@ -1684,8 +1684,15 @@ func main() {
...
-			options.Target = "wasm"
+		if options.Target == "" && options.GOARCH == "wasm" {
+			switch options.GOOS {
+			case "js":
+				options.Target = "wasm"
+			case "wasip1":
+				options.Target = "wasip1"
+			case "wasip2":
+				options.Target = "wasip2"
+			}
```
aykevl, line 1704: Perhaps we should add a default case here with an error message?
Something like (pseudocode-ish):

```go
case "":
    return errors.New("GOOS is unset but GOARCH is set to wasm. Please set GOOS to wasm, wasip1, or wasip2.")
default:
    return errors.New(fmt.Sprintf("Unknown GOOS=%s for GOARCH=wasm. Please set GOOS to wasm, wasip1, or wasip2.", options.GOOS))
```

aykevl (commented):
The behavior with this PR is:

```
$ tinygo build -o test.wasm ./testdata/alias.go
```

Creates a native binary (not a wasm binary!). This is a change from the previous behavior, and I'm afraid a lot of people are going to be confused by this. Perhaps we should still default to the equivalent of `-target=wasm` when using the .wasm extension?

```
$ GOARCH=wasm tinygo build -o test.wasm ./testdata/alias.go
should have used target for GOARCH=wasm
```

That's a bit of a confusing error, at least to me. I think it is saying the user should have used the `-target=` flag?
A better error might be to tell the user that GOOS needs to be set in this case.

```
$ GOARCH=wasm GOOS=foobar tinygo build -o test.wasm ./testdata/alias.go
should have used target for GOARCH=wasm
```

Same output as before with GOOS set to an invalid value. I think we need better error handling here.

```
$ GOOS=js tinygo build -o test.wasm ./testdata/alias.go
unknown GOOS=js
```

That's contradictory, since we do in fact support GOOS=js. A better error message might be something like "GOOS=js is only supported with GOARCH=wasm".

deadprogram:
> Creates a native binary (not a wasm binary!). This is a change from the previous behavior, and I'm afraid a lot of people are going to be confused by this.

@aykevl I am now returning an error that the user needs to set either a target or else GOOS?GOARCH in this case.

I have also made some error message improvements based on your other comments.

main.go
deadprogram: Thanks for the feedback. See my changes.

deadprogram:
@aykevl any further feedback here?

compileopts/target.go
```diff
@@ -446,23 +436,8 @@ func defaultTarget(options *Options) (*TargetSpec, error) {
...
-			"--no-demangle",
-		)
-		spec.Emulator = "wasmtime run --dir={tmpDir}::/tmp {}"
-		spec.ExtraFiles = append(spec.ExtraFiles,
-			"src/runtime/asm_tinygowasm.S",
-			"src/internal/task/task_asyncify_wasm.S",
-		)
-		llvmos = "wasi"
+	case "wasm", "wasip1", "wasip2":
+		return nil, fmt.Errorf("GOOS=%s but GOARCH is unset. Please set GOARCH to wasm or js", options.GOOS)
```
aykevl:
```suggestion
		return nil, fmt.Errorf("GOOS=%s but GOARCH is unset. Please set GOARCH to wasm", options.GOOS)
```

GOARCH=js isn't a thing (that would mean we produce JavaScript code, which we don't).

main.go
```diff
@@ -1684,8 +1684,24 @@ func main() {
...
+				switch options.GOOS {
+				case "js":
+					options.Target = "wasm"
+				case "wasip1":
+					options.Target = "wasip1"
+				case "wasip2":
+					options.Target = "wasip2"
+				default:
+					fmt.Fprintln(os.Stderr, "GOARCH=wasm but GOOS is unset. Please set GOOS to wasm, wasip1, or wasip2.")
+					os.Exit(1)
```
aykevl: There's also the case of `GOARCH=wasm GOOS=foobar`, which would produce a "GOOS is unset" error. That's somewhat wrong (though hopefully not very common). To be really sure it might be better to have a `""` case.

aykevl (commented):
Looks good to me, with a few nits.

main.go
deadprogram: I ended up with "GOOS is not set correctly" which would be a valid statement for either case.

compileopts/target.go
deadprogram: Fixed.

deadprogram:
OK finally ready to merge. Thank you @ydnar and @aykevl for your reviews/feedback!

PR #4747, opened 2025-02-21
machine: compute rp2 clock dividers from crystal and target frequency

eliasnaur:
Follow-up to #4728 which implemented the algorithm for finding the dividers.

The calculation is computed at compile time by interp, as verified by the unchanging size for example/blinky1 for -target pico.

@soypat you [mentioned](https://github.com/tinygo-org/tinygo/pull/4728#discussion_r1955270154) that you couldn't get TinyGo to compute the dividers at compile time. However, this PR doesn't change the size of the blinky1 example for `-target pico`. What did I miss?

### DIFF
--- a/src/machine/machine_rp2_clocks.go
+++ b/src/machine/machine_rp2_clocks.go
@@ -137,7 +137,19 @@ func (clk *clock) configure(src, auxsrc, srcFreq, freq uint32) {
 
 }
 
-const pllsysFB, pllsysPD1, pllsysPD2 uint32 = 125, 6, 2 // RP2040 running 125MHz with 1500MHz VCO.
+var pllsysFB, pllsysPD1, pllsysPD2 uint32
+
+// Compute clock dividers.
+//
+// Note that the entire init function is computed at compile time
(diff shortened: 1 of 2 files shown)


soypat:
Interesting you got the same size. I tried something similar, but instead of using `init` I tried directly initializing the variable at package level `var pllSys, pll... = pllSearch{...}.CalcDiv(...)` and I actually got a smaller binary size than the blinky example which told me something was VERY wrong. sure enough when flashing a RP2040 the light would not light up, presumably having crashed/panicked in calculation.

Will try to test it

eliasnaur:
Gentle ping. Eager to boost my Picos :)

soypat:
Amazing, tested and works. Very nice win by interp. Curious to see how we let users set the frequency. Hopefully we can do some linker stuff so that a `machine.cpuFrequency` variable can be set via a compiler flag?

eliasnaur:
Thank you. Speed bump at https://github.com/tinygo-org/tinygo/pull/4766.

PR #4750, opened 2025-02-23
reflect: Go 1.24 support

ydnar:
This is WIP, currently broken due to a cyclic import between packages `reflect` and `runtime` in TinyGo.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,9 +109,9 @@ jobs:
           # "make lint" fails before go 1.21 because internal/tools/go.mod specifies packages that require go 1.21
           fmt-check: false
     resource_class: large
-  test-llvm19-go123:
+  test-llvm19-go124:
     docker:
-      - image: golang:1.23-bullseye
+      - image: golang:1.24-bullseye
     steps:
       - test-linux:
           llvm: "19"
@@ -124,4 +124,4 @@ workflows:
       # least the smoke tests still pass.
       - test-llvm15-go119
       # This tests LLVM 19 support when linking against system libraries.
-      - test-llvm19-go123
+      - test-llvm19-go124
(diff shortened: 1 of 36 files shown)


ydnar:
@aykevl @deadprogram @dgryski I need some help bringing this over the finish line, particularly with the circular dependency.

deadprogram:
Looks like 2 packages in the TinyGo `runtime` import the `reflect` package: `hashmap.go` and `interface.go`. I think `interface.go` is only used by `hashmap.go` so the hashmap implementation is probably the key here, so to speak.

@dgryski we probably need your help here, I looked over the files in question, and it was not immediately clear to me how to avoid using reflection in the current implementation.

dgryski:
The solution is to split `reflect` into a tiny `reflectlite` package for use by the runtime. We don't need all of `reflect`, just some of the type introspection bits.

deadprogram:
> The solution is to split reflect into a tiny reflectlite package for use by the runtime. We don't need all of reflect, just some of the type introspection bits.

@dgryski any chance you can take a pass at that? It would be incredibly helpful.

deadprogram:
<s>Looking from progress in https://github.com/tinygo-org/tinygo/pull/4732 perhaps this PR will not be needed?</s>

Was explained to me last night what this PR is supposed to be doing. Never mind.

aykevl:
I think we still need this, but needs some refactoring first. Most likely by splitting reflect and reflectlite. (Also it needs to be rebased on the dev branch after #4732 got merged).

I looked into splitting reflectlite today, but wasn't happy with the two ways I approached it. I wanted to avoid too much code duplication, but that's difficult.

ydnar:
@aykevl I was able to get around the `reflectlite` duplication with a bit of `go:linkname`.

PR #4753, opened 2025-02-24
ci: use older image for cross-compiling builds

aykevl:
This ensures that the resulting binaries are compatible with a wide range of Linux systems, not just the most recent ones.
See: https://github.com/tinygo-org/tinygo/pull/4731

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -297,7 +297,7 @@ jobs:
           - goarch: arm
             toolchain: arm-linux-gnueabihf
             libc: armhf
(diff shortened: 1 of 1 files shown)


deadprogram:
This failure is why I updated to the latest version I did. There is some problem with the cross-compiling toolchain still unresolved for older version.

aykevl:
Forced a rebuild of LLVM, that will hopefully fix this problem.
Using a LLVM build from a different Ubuntu version is likely not going to work anyway.

deadprogram:
It passed! Now merging, thanks for the improvement @aykevl

PR #4754, opened 2025-02-24
ci: use native arm64 builder for arm64 binaries

aykevl:


aykevl:
Hmm, that doesn't seem to work. Now the build-linux job doesn't run at all.

mikesmitty:
Looks like the new variables are missing the leading `$`

aykevl:
Looks like this is not working at the moment.
We're running into this issue: https://github.com/actions/runner/issues/801#issuecomment-2394425757

> `Error: JavaScript Actions in Alpine containers are only supported on x64 Linux runners. Detected Linux Arm64`


(There's a workaround but I'm not entirely comfortable with that one, I'd rather wait a bit longer to see whether it gets fixed).

aykevl:
@mikesmitty thanks, that was indeed one of the issues. The other was that I missed a comma in the `goarch:` line.

deadprogram:
Rebased against latest `dev` and merge conflicts corrected. Let's see if it can run successfully now.

deadprogram:
The `JavaScript Actions in Alpine containers are only supported on x64 Linux runners. Detected Linux Arm64` error is still an issue. We must continue to wait for it to be addressed upstream.

deadprogram:
Closing since out of date. Thanks!

PR #4756, opened 2025-02-25
refactor: use *SPI everywhere to make consistent for implementations.

### LINKED ISSUE #4663: rp2040 SPI0 variable is a pointer and not struct
I am developing for the RPI Pico and noticed that the definition of the SPI0 & SPI1 variables are pointers to the SPI struct. This differ to the SPI0 definition for generic boards, which is simply a struct.  As a result the following code compiles with target=pico but does not compile if target is generic/wasm.

```
func main() {
	machine.SPI1.Configure(machine.SPIConfig{
		SDO: machine.GP15, // default SDO pin - TX
		SCK: machine.GP14, // default sck pin
		SDI: machine.GP28, // default sdi pin
		Frequency: 10000000,
	})
	driver = max72xx.NewDevice(*machine.SPI1, machine.GP13) // Selector
       // compiler error if target is generic
```

This is because the declaration of SPI0 is different depending which tag you use:
machine/machine_rp2040_spi.go
```
// SPI on the RP2040
var (
	SPI0  = &_SPI0
	_SPI0 = SPI{
		Bus: rp.SPI0,
	}
	SPI1  = &_SPI1
	_SPI1 = SPI{
		Bus: rp.SPI1,
	}
)
```

machine/machine_generic_peripherals.go
```
var (
	UART0 = hardwareUART0
	UART1 = hardwareUART1
	SPI0  = SPI{0}
	SPI1  = SPI{1}
	I2C0  = &I2C{0}
)
```

It would be more consistent for rp2040 SPI to also be a struct, not a pointer. unless you know a particular reason why this is like that.
Happy to submit a change request if the change makes sense.

deadprogram:
This PR is to fix #4663 "in reverse" by making SPI a pointer everywhere, as discussed in the comments.

See https://github.com/tinygo-org/tinygo/issues/4663#issuecomment-2553136448

### DIFF
--- a/src/machine/board_feather-stm32f405.go
+++ b/src/machine/board_feather-stm32f405.go
@@ -186,15 +186,15 @@ const (
 )
 
 var (
-	SPI1 = SPI{
+	SPI1 = &SPI{
 		Bus:             stm32.SPI2,
 		AltFuncSelector: AF5_SPI1_SPI2,
 	}
-	SPI2 = SPI{
+	SPI2 = &SPI{
 		Bus:             stm32.SPI3,
 		AltFuncSelector: AF6_SPI3,
 	}
-	SPI3 = SPI{
+	SPI3 = &SPI{
 		Bus:             stm32.SPI1,
 		AltFuncSelector: AF5_SPI1_SPI2,
 	}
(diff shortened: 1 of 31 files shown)


bgould:
Looks good to me ... seems like there is a conflict to resolve though

deadprogram:
Resolved and rebased. Now waiting for CI to pass...

deadprogram:
Thanks for reviewing @bgould now merging.

PR #4757, opened 2025-02-25
Revert "fix: implement testing {Skip,Fail}Now"

aykevl:
This reverts commit c5879c682c834e242aa86f86df2ba4984dfd9ba9.

It doesn't look like this is working (see https://github.com/tinygo-org/tinygo/pull/4736#issuecomment-2679670226), and it doesn't have any tests anyway to prove that it does work. So I think it's best to revert it for now and add a working implementation with tests in the future.

### DIFF
--- a/src/testing/testing.go
+++ b/src/testing/testing.go
@@ -17,7 +17,6 @@ import (
 	"io/fs"
 	"math/rand"
 	"os"
-	"runtime"
 	"strconv"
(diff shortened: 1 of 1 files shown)


deadprogram (approved):
Sorry that I did not look a little closer at the PR, I should have realized that those functions were not actually being called anyplace.

PR #4758, opened 2025-02-25
compiler, runtime: implement recoverable divide-by-zero panic

aykevl:
This gets the math/bits package tests to pass.

Unfortunately, this also has a binary size impact of around 150-200 bytes in many cases. I'm a bit on the edge on whether this is worth it, since it's mostly used for getting package tests to work. But at the same time, having working package tests is very valuable.


aykevl:
Not happy with the implementation just yet. Currently it removes runtime panic locations, which are very useful for debugging.

jakebailey:
#5438 did this, as part of making runtime panics recoverable.

PR #4761, opened 2025-02-26
internal/syscall/unix: use our own version of this package

aykevl:
The upstream one assumes it's running on a Unix system (which makes sense), but this package is also used on baremetal. So replace it on systems that need a replaced syscall package.


aykevl:
@dgryski there is a chance this will break wasip2 in some way. I ran the test corpus (see https://github.com/tinygo-org/tinygo/pull/4762) and it didn't break anything, but still a heads up.

deadprogram:
Now merging, thanks for the quick fix @aykevl

PR #4762, opened 2025-02-27
make: add test-corpus-wasip2

aykevl:
I wanted to run the test corpus with WASIp2 so I added these lines.


deadprogram (approved):
It is running the tests, and they are all passing, so looking good to me!

aykevl:
The test corpus is not part of CI, so this won't have any effect on the tests passing in CI. It's just convenience for manually running these tests.

deadprogram:
> The test corpus is not part of CI

Should we be adding it?

aykevl:
Perhaps, but the test corpus is also very big and very slow. Maybe it would work better as a nightly or weekly test?
@dgryski probably has some opinions here.

dgryski:
I always ran it by hand. I agree it's too big to run for every commit. Once a week is nice if we could automate it. Otherwise just a reminder somewhere so people know it exists and that they should run it occasionally.

I will say I don't think it finds bugs that the stdlib tests don't. It's more that it provides confidence in a change and also a nice list of known working packages.

aykevl:
I find it useful to run after a possibly tricky change (such as #4761). But yeah, it doesn't find all that many bugs. The standard library tests already exercise the compiler very well.

aykevl:
In any case, I don't think this is needed in the 0.36.0 milestone. We could add it, or not, but it doesn't matter much.

deadprogram:
OK thanks for the clarification. Seems like we can merge just to have the additional coverage when useful to run locally.

PR #4766, opened 2025-03-01
machine: bump rp2350 CPUFrequency to 150 MHz

eliasnaur:
Leave rp2040 speed bump to 200 MHz for a future change, because it requires bumping the core voltage as well[0].

[0] https://github.com/raspberrypi/pico-sdk/releases/tag/2.1.1

### DIFF
--- a/src/machine/machine_rp2_2040.go
+++ b/src/machine/machine_rp2_2040.go
@@ -9,6 +9,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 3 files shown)


soypat (approved):
Excellent- we now have supa fast rp2350

soypat:
(went ahead and tested it on a pico 2 with blinky1 after approving and it does in fact work :)

PR #4767, opened 2025-03-01
machine: replace hard-coded cpu frequencies on rp2xxx

eliasnaur:
Missed from the earlier change that bumped the rp2350 frequency.

@soypat sorry for missing this.

### DIFF
--- a/src/machine/machine_rp2_clocks.go
+++ b/src/machine/machine_rp2_clocks.go
@@ -144,7 +144,7 @@ var pllsysFB, pllsysPD1, pllsysPD2 uint32
 // Note that the entire init function is computed at compile time
 // by interp.
 func init() {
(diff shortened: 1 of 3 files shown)


soypat (changes requested):
Hmmm these should be a call to CPUFrequency(). And conversely the call in CalcDivs should take the constant cpuFreq instead of calling the function

deadprogram:
@soypat these changes looking good to merge?

soypat (approved):

soypat:
Itll work, but for future reference drivers should call CPUFrequency() inatead of referring to the constsnt directly for future proofing

eliasnaur:
PTAL. Sorry for misunderstanding your review comment.

eliasnaur:
I changed the hardcoded values in the clocks initialization procedure. PTAL.

soypat (approved):

soypat:
Nice find with the xoscFreqs

PR #4770, opened 2025-03-02
reflect, runtime: additional Go 1.24 support

ydnar:
Two major changes for Go 1.24 support:

1. Removes package `runtime` dependency on package `reflect`, required for `iter` support in `reflect`.
2. Copies the implementation of `reflect.Value.Equal` from big Go. This is necessary for JSON parsing in Go 1.24. See: https://go-review.googlesource.com/c/go/+/626116

### DIFF
--- a/src/reflect/hashmap.go
+++ b/src/reflect/hashmap.go
@@ -1,0 +1,87 @@
+package reflect
+
+import (
+	"unsafe"
+)
+
+//go:linkname hash32 runtime.hash32
+func hash32(ptr unsafe.Pointer, n, seed uintptr) uint32
+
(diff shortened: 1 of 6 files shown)


aykevl:
Unfortunately this requires the reflect package to be included in every build, which is not the case.

deadprogram:
Is it possible to use https://github.com/tinygo-org/tinygo/pull/4770/commits/dd75c5cfb61b1a25d7268b2383d13852c14e9805 on its own? Can that help with decoding json leaving the `iter` issue to be resolved separately?

ydnar:
Closing in favor of #4787.

PR #4771, opened 2025-03-03
src/reflect: implement Value.Equal

deadprogram:
Implementation copied from Go.

This PR pulls one of the commits from https://github.com/tinygo-org/tinygo/pull/4770

See https://go-review.googlesource.com/c/go/+/626116


deadprogram:
Thanks for the original commit on this @ydnar now merging.

PR #4774, opened 2025-03-04
internal/reflectlite: do not use the reflect package

aykevl:
This avoids a dependency on the reflect package.

I've tried to deduplicate as much as reasonably possible, while keeping the code clean. Especially some tricky methods for the reflect.Type type have a shared implementation (using //go:linkname). In other cases, I've simply used the same type using a type alias, in the other direction (the reflect package now imports the internal/reflectlite package which should be safe).

~TODO: actually let the runtime package use this package. For that, a bit more work is needed.~ done


aykevl:
Hitting a weird issue with io/fs, this might indicate a deeper problem. Looking into it.

EDIT: I think I fixed it (needed a fix to OptimizeReflectImplements).

src/reflect/value.go
```diff
@@ -11,6 +12,8 @@ type valueFlags uint8
 // contained in an interface{} directly, like whether this value was exported at
 // all (it is possible to read unexported fields using reflection, but it is not
 // possible to modify them).
+//
+// These flags are shared with the internal/reflectlite package.
 const (
 	valueFlagIndirect valueFlags = 1 << iota
```
ydnar, line 18: If these are shared, should these constants `=` their `reflectlite` counterparts?

ydnar (commented):
Thanks! LGTM. I look forward to rebasing my `iter` PR on top of this.

ydnar:
@aykevl can you look at #4787? It’s an alternative implementation to this, with the Go 1.24 (`iter`) additions and updates to package `runtime`.

aykevl:
Oops, should have tested this one:

```
panic: todo: internal/reflectlite.Type.String
FAIL	reflect	2.158s
```

aykevl:
Fixed the above panic.

deadprogram:
@ydnar and @dgryski are we cool to merge this PR and continue work on `iter` and friends using this implementation?

ydnar:
Can we evaluate this vs the approach I took?

deadprogram:
At least from the binaries that had size increase, here is what I see:

This PR:

```shell
 flash                          ram
 before   after   diff          before   after   diff
64340   64488    148   0.23%    6204    6196     -8  -0.13% tinygo build -size short -o ./build/test.hex -target=itsybitsy-m0 ./examples/bmi160/main.go
64048   64196    148   0.23%    6228    6220     -8  -0.13% tinygo build -size short -o ./build/test.hex -target=itsybitsy-m0 ./examples/bmp280/main.go
61472   61628    156   0.25%    6188    6180     -8  -0.13% tinygo build -size short -o ./build/test.hex -target=itsybitsy-m0 ./examples/adt7410/main.go
69720   69876    156   0.22%    6376    6368     -8  -0.13% tinygo build -size short -o ./build/test.hex -target=itsybitsy-m0 ./examples/ds3231/main.go
70692   70888    196   0.28%    3656    3648     -8  -0.22% tinygo build -size short -o ./build/test.hex -target=pinetime     ./examples/bma42x/main.go
```

PR #4787 

```shell
 flash                          ram
 before   after   diff          before   after   diff
63988   64740    752   1.18%    6196    6196      0   0.00% tinygo build -size short -o ./build/test.hex -target=itsybitsy-m0 ./examples/bmi160/main.go
64048   64784    736   1.15%    6228    6228      0   0.00% tinygo build -size short -o ./build/test.hex -target=itsybitsy-m0 ./examples/bmp280/main.go
61472   62224    752   1.22%    6188    6188      0   0.00% tinygo build -size short -o ./build/test.hex -target=itsybitsy-m0 ./examples/adt7410/main.go
69720   70480    760   1.09%    6376    6376      0   0.00% tinygo build -size short -o ./build/test.hex -target=itsybitsy-m0 ./examples/ds3231/main.go
70692   71412    720   1.02%    3656    3656      0   0.00% tinygo build -size short -o ./build/test.hex -target=pinetime     ./examples/bma42x/main.go
```

src/reflect/type.go
```diff
@@ -412,6 +343,7 @@ type Type interface {
 }
 
 // Constants for the 'meta' byte.
+// These constants are also defined in the internal/reflectlite package.
 const (
 	kindMask       = 31  // mask to apply to the meta byte to get the Kind value
```
ydnar, line 348: Given `reflectlite` is an internal package, can these constants be shared, rather than duplicated? e.g. `kindMask = reflectlite.KindMask`?

ydnar (changes requested):
My primary feedback is the duplication of `const` values and structural types between the two packages. Given `reflectlite` is an internal package, can it export the constants (like `kindMask`) and structural types (like `rawType`, `elemType`, `ptrType`, etc), to avoid 1:1 duplication between these two packages?

With that and merging of #4806, this should be good to go.

src/reflect/type.go
deadprogram: @aykevl what do you think? Seems like a good idea to me.

aykevl: Yes, that's definitely possible! I can update this PR if this PR is the one we're going for.

ydnar:
Implemented in #4787.

PR #4775, opened 2025-03-04
Add goroutine benchmark to examples

soypat:
Added `src/examples/bench-goro` example.

Output on Raspberry Pi Pico (2040)

```
bench: 3.730954s goroutine ctx switch: 1.865µs
```

And now with new changes to the CPU frequency (125->200MHz:

```
bench: 2.372861s goroutine ctx switch: 1.186µs
```

### DIFF
--- a/src/examples/bench-goro/bench.go
+++ b/src/examples/bench-goro/bench.go
@@ -1,0 +1,40 @@
+package main
+
+import (
+	"machine"
+	"runtime"
+	"sync"
+	"time"
+)
+
+const N = 500000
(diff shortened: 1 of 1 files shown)


soypat:
@aykevl @deadprogram What do y'all think of adding this example?

deadprogram (approved):
I like it, seems pretty useful to me as a minimum viable benchmark.

deadprogram:
Changed target branch to `dev`.

deadprogram:
Now merging, thanks @soypat

PR #4781, opened 2025-03-06
machine: don't block the rp2xxx UART interrupt handler

eliasnaur:
Don't block forever if there's nothing to receive. On the other hand, process the entire FIFO, not just a single byte.

This fixes an issue where the rp2350 would hang after programming through openocd, because the UART0 interrupt would be spuriously pending.

CC @deadprogram who wrote https://github.com/tinygo-org/tinygo/commit/87e48c105791187ae1bbd096ae55029de2b072a3 and may know if I missed a subtlety here.

### DIFF
--- a/src/machine/machine_rp2_uart.go
+++ b/src/machine/machine_rp2_uart.go
@@ -67,7 +67,7 @@ func (uart *UART) Configure(config UARTConfig) error {
 	uart.Interrupt.SetPriority(0x80)
 	uart.Interrupt.Enable()
 
-	// setup interrupt on receive
+	// Setup interrupt on receive.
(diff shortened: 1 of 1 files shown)


soypat:
Would this cause UART.Read to not block now?

eliasnaur:
Did UART.Read ever block? Even if it did, blocking the interrupt handler wouldn't work because it's only called when a byte is received (or sometimes spuriously after programming).

deadprogram:
Excellent work @eliasnaur this is much better now! Also thanks for review @soypat

Thanks for digging into it, now merging.

soypat:
@eliasnaur thanks for the reply- I now see the issue :)

PR #4782, opened 2025-03-06
machine: correct register address for Pin.SetInterrupt for rp2350

### LINKED ISSUE #4689: GPIO interrupts not recognized on RP2350 (Pico 2)
I have written a simple program that should recognize edges on an input and trigger an interrupt. The interrupt doesn't happen.

Here is the simple program:
```go
package main

import (
	"fmt"
	"machine"
	"time"
)

func bcast(x bool) int {
	if x {
		return 1
	} else {
		return 0
	}
}

func main() {
	// square wave is fed to pin 10
	p := machine.Pin(10)
	p.Configure(machine.PinConfig{Mode: machine.PinInputPullup})
	// and should be echoed to pin 11
	q := machine.Pin(11)
	q.Configure(machine.PinConfig{Mode: machine.PinOutput})
	err := p.SetInterrupt(machine.PinRising, func(pin machine.Pin) {
		fmt.Printf("verify pin: %d\n", pin)
		fmt.Printf("interrupt: %d, %d\n", bcast(p.Get()), bcast(pin.Get()))
		q.Set(p.Get())
	})
	if err != nil {
		fmt.Printf("err set interrupt: %s\n", err.Error())
		panic(err.Error())
	} else {
		fmt.Printf("set interrupt ok\n")
	}
	for i := 0; i < 300; i++ {
		fmt.Printf("tick %d, pin=%d\n", i, bcast(p.Get()))
		time.Sleep(1 * time.Second)
	}
}
```
And here is the output when run on a Pico2. The main loop runs fine and the pin is clearly changing state but the interrupt is not recognized.

I suspect that there is something about the RP2350 that requires a slight additional bit of work in order to enable the control of interrupts. It may have to do with the newly introduced security model. 

I don't have a 2040 based Pico to test on right now so I can't confirm this works on the RP2040
```
 tinygo flash -target=pico2 -scheduler=tasks -gc=conservative -size=full -stack-size=20kb -monitor IrqBug/main.go
      0      45       3       5 |      48       8 | (padding)
   1560       0      10      45 |    1570      55 | (unknown)
   1220       0       0       0 |    1220       0 | C compiler-rt-builtins
      0     248       0       0 |     248       0 | C interrupt vector
     76       0       0       0 |      76       0 | C picolibc
    160       0       0       0 |     160       0 | Go interface assert
    292       0       0       0 |     292       0 | Go interface method
      0    1148       0       0 |    1148       0 | Go types
     96       0       0       0 |      96       0 | device/arm
     68       0       0       0 |      68       0 | device/rp
      6       0       0       0 |       6       0 | errors
   8286     216       0       0 |    8502       0 | fmt
     14       0       0       0 |      14       0 | internal/binary
   1136      21       0       0 |    1157       0 | internal/fmtsort
    132       2       0       0 |     134       0 | internal/itoa
      0      19       8       0 |      27       8 | internal/oserror
    338      48       4       4 |     390       8 | internal/task
     46       2       0       0 |      48       0 | io/fs
   1662     159       8    1082 |    1829    1090 | machine
    400       0       0       7 |     400       7 | machine/usb/cdc
      0       0      93       0 |      93      93 | machine/usb/descriptor
    440      88       0       0 |     528       0 | main
    134       0       0       0 |     134       0 | math/bits
    160      39      32       0 |     231      32 | os
   4826     773      48       0 |    5647      48 | reflect
   4350     722       4     753 |    5076     757 | runtime
    792       0       0       0 |     792       0 | runtime/volatile
    510       0       0       0 |     510       0 | sort
   7632   16012    1338       0 |   24982    1338 | strconv
     76       0       0       0 |      76       0 | sync
     32       6       0       0 |      38       0 | syscall
   1036     288       0       0 |    1324       0 | unicode/utf8
  35480   19836    1548    3944 |   56864    5492 | total
Connected to /dev/cu.usbmodem14101. Press Ctrl-C to exit.
tick 1, pin=0
tick 2, pin=0
tick 3, pin=0
tick 4, pin=0
tick 5, pin=0
tick 6, pin=0
tick 7, pin=0
tick 8, pin=0
tick 9, pin=0
tick 10, pin=1
tick 11, pin=1
tick 12, pin=1
tick 13, pin=1
tick 14, pin=0
tick 15, pin=0
tick 16, pin=0
tick 17, pin=1
tick 18, pin=0
tick 19, pin=0
```

eliasnaur:
Fixes #4689

### DIFF
--- a/src/machine/machine_rp2_gpio.go
+++ b/src/machine/machine_rp2_gpio.go
@@ -33,6 +33,7 @@ type irqSummary struct {
 
 type ioBank0Type struct {
 	io                 [_NUMBANK0_GPIOS]ioType
+	_                  [rp2350ExtraReg][128]byte
 	irqsum             [rp2350ExtraReg]irqSummary
 	intR               [_NUMBANK0_IRQS]volatile.Register32
 	proc0IRQctrl       irqCtrl


soypat:
LGTM. `0x200-(0x17c+4) = 128`
image

PR #4783, opened 2025-03-07
machine: fix RP2040 Pico board on the playground

aykevl:
Right now it doesn't compile, with errors like the following:

    # machine
    /app/tinygo/src/machine/board_pico.go:7:13: undefined: GPIO0
    /app/tinygo/src/machine/board_pico.go:8:13: undefined: GPIO1
    /app/tinygo/src/machine/board_pico.go:9:13: undefined: GPIO2
    /app/tinygo/src/machine/board_pico.go:10:13: undefined: GPIO3
    [...etc...]

This patch should fix that.

Also added a smoke test, since somehow this wasn't being tested (hence why it broke, probably).


deadprogram:

PR #4784, opened 2025-03-07
compiler: crypto/internal/sysrand is allowed to use unsafe signatures

aykevl:
Apparently this package imports `runtime.getRandomData` from the gojs module. This is not yet implemented, but simply allowing this package to do such imports gets crypto/sha256 to compile.


aykevl:
Oh, fun.

```
Run bytecodealliance/actions/wasm-tools/setup@v1
finding latest release for platform linux and architecture x86_64
Error: no releases found for platform linux and architecture x86_64
```

EDIT: that was just a fluke apparently.

dkegel-fastly:
last I heard, they were planning a release later this month for other reasons, which is pretty soon

deadprogram:
Thank you for the quick fix here @aykevl now merging.

PR #4785, opened 2025-03-07
ci: add more tests for wasm and baremetal

aykevl:
Run a range of tests in CI, to make sure browser wasm and baremetal don't regress too badly.

I have intentionally filtered out tests (instead of making a list of working tests), so that newly added tests to TEST_PACKAGES_FAST will be added here as well (and can be excluded if needed).

Depends on #4784, must be rebased after that PR is merged.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -278,7 +278,7 @@ jobs:
         run: make tinygo-test
       - run: make smoketest
       - run: make wasmtest
-      - run: make tinygo-baremetal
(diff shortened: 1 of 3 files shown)


aykevl:
Rebased, ready for review!
This will make CI slower, but I think these are tests that we need to run. Right now we have just the compiler tests to test basic functionality, adding this list of packages adds a lot more code we run through the tests.

ydnar (approved):
LGTM.

PR #4791, opened 2025-03-10
riscv-qemu: add VirtIO RNG device

aykevl:
This implements machine.GetRNG() using VirtIO. This gets the tests to pass for crypto/md5 and crypto/sha1 that use crypto/rand in their tests.

The implementation is currently very specific to the RNG. But the code can be refactored in the future to support other types of devices if needed (such as a network driver, to test the net package). It is also not at all optimized for performance: it reads 4 bytes at a time instead of a larger buffer. But that should be good enough to run some tests.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -429,7 +429,6 @@ TEST_PACKAGES_NONWASM = \
 #
 #   * No filesystem is available, so packages like compress/zlib can't be tested
 #     (just like wasm).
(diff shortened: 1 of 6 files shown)


deadprogram:
This is a very useful @aykevl and I'm glad I did not have to do the research on implementation. :smile_cat: 

Now merging, thanks!

PR #4792, opened 2025-03-10
ci: bump stack size for baremetal package tests

aykevl:
Bumping the stack size to 8kb (previous was 4kb) gets most remaining tests to pass on baremetal.
The only remaining one is the math package, which seems to be due to an inaccuracy in the picolibc package.

Depends on #4791 (it would conflict otherwise, and crypto/elliptic needs both of these changes to pass).


aykevl:
This is now ready for review.

deadprogram:
> gets most remaining tests to pass on baremetal.

That is pretty awesome! Thanks for working on this effort @aykevl now merging.

PR #4798, opened 2025-03-11
fix: ensure use of pointers for SPI interface on missing machines

deadprogram:
This PR is to ensure use of pointers for SPI interface on atsam21/atsam51 and other machines that were missing the correct implementation.

Recent build errors in `drivers` repo and other indicated that PR #4756 was not actually complete. This PR is to address that.


ysoldak (approved):
LGTM

gandarez (approved):

PR #4803, opened 2025-03-13
machine/rp2350: add flash support for rp2350

### LINKED ISSUE #4799: Missing internal flash support for pico2 boards
https://github.com/tinygo-org/tinygo/blob/8c54e3dd885824638015f319907f75c01570cdad/src/machine/flash.go#L1
Shall we include rp2350 tag above?

mikesmitty:
Hi all, this still needs a bit of cleanup and I haven't had the chance to properly test it on hardware, but it builds cleanly. Figured I'd push it up to see if any big changes need to happen before I give it a final polish.

Fixes #4799


src/machine/machine_rp2350_flash.go
```diff
@@ -0,0 +1,111 @@
+//go:build rp2350
```
eliasnaur: Is this file a straight copy of `machine_rp2040_flash.go`? If so, rename `machine_rp2040_flash.go` to `machine_rp2_flash.go` and change its header to say `//go:build rp2040 || rp2350`.

mikesmitty: Oh yeah, will do

mikesmitty:
Ok, I had a chance to test the console functions in https://github.com/tinygo-org/tinyfs/blob/release/examples/console/littlefs/machine/main.go and it seems to be working as far as I can tell, but I've not used tinyfs before

eliasnaur:
I'm not the right reviewer for the details of the flash interface. Perhaps @deadprogram who authored the other implementation.

mikesmitty:
Sounds good. It's mostly just copy/paste from pico-sdk with the RISC-V and RP2040-specific macro branches trimmed out and I grouped everything by source file with a header comment so it should hopefully be relatively straightforward to cross-reference. The only real changes I made were to mimic the conversion of `flash_range_program()`/`flash_range_erase()` to `flash_range_write()`/`flash_erase_blocks()` that was done in `machine_rp2040_rom.go` and to ditch the enum `flash_devinfo_get_cs_size()` used as a return value in favor of a `uint8_t`. It's only used once and only to check against the zero value from 4 register bits. I left that bit in even though it's only used for checking if a PSRAM chip exists to make it easier to add support for it down the road, but I can trim it out if that makes more sense

deadprogram:
Thank you for working on this @mikesmitty and to @eliasnaur for review.

I tested on real hardware, working exactly as expected. Now squash/merging. Thanks again!

mikesmitty:
Thanks!

PR #4787, opened 2025-03-08
reflect: embed internal/reflectlite types

ydnar:
This PR is an alternative to #4774, using type aliases and struct embedding in lieu of `go:linkname`.

- Package `reflect` now depends on package `internal/reflectlite`, reversing the existing dependency.
- Adds Go 1.24 `iter.Seq` methods to `reflect`.
    - Adds fix for https://github.com/golang/go/issues/71905.
- Fixes the circular dependency between package `runtime` and `reflect` introduced via package `iter` in Go 1.24.

### Note

An alternative to this PR could implement the common functionality in `reflect` and `reflectlite` in a new, third package (e.g. `internal/tinyreflect`).

PR #4806, opened 2025-03-14
reflect, runtime, internal/reflectlite: add iter support

### LINKED ISSUE #4487: `reflect.Value` methods `Seq` and `Seq2` missing
These two methods are new in Go 1.23 as part of the new iterators feature:
- https://pkg.go.dev/reflect#Value.Seq
- https://pkg.go.dev/reflect#Value.Seq2

```
% cat main.go 
package main

import "reflect"

func main() {
	var v reflect.Value
	_ = v.Seq
	_ = v.Seq2
}
% go build -o main
% tinygo build -o main
# example.com/example
main.go:7:8: v.Seq undefined (type reflect.Value has no field or method Seq)
main.go:8:8: v.Seq2 undefined (type reflect.Value has no field or method Seq2)
```

ydnar:
This PR adds Go 1.24 `iter` support to package `reflect`.

- `Type.CanSeq` and `Type.CanSeq2`
- `Value.Seq` and `Value.Seq2`

Fixes #4487.


src/reflect/iter_test.go
```diff
@@ -0,0 +1,416 @@
...
+			}
+			if len(copy) != 0 {
+				t.Fatalf("should loop four times")
+			}
+		}},
+		// {"func", ValueOf(func(f func(int, int) bool) {
+		// 	for i := range 4 {
+		// 		f(i, i+1)
+		// 	}
+		// }), func(t *testing.T, s iter.Seq2[Value, Value]) {
```
deadprogram, line 315: Are these supposed to be commented?

ydnar: TinyGo reflect doesn't support functions so I disabled those tests.

deadprogram:
This PR will also address https://github.com/tinygo-org/tinygo/issues/4487

ydnar:
Implemented in #4787.

Issue #4810, opened 2025-03-17
builder: runtime.buildVersion not a string


ydnar:
When running tests under `-target=wasip2`, the build for #4809 was [failing](https://github.com/tinygo-org/tinygo/actions/runs/13887548775/job/38854292078?pr=4809) when testing `crypto/elliiptic`:

```console
$ tinygo test -v -target=wasip2 crypto/elliptic
FAIL	crypto/elliptic	0.000s
runtime.buildVersion: not a string
```

I traced the issue to an error type in package `internal/cm`:

```go
var (
	errEmpty          = &stringError{err: "empty text"}
	errNoMatchingCase = &stringError{err: "no matching case"}
)

type stringError struct {
	_   struct{}
	err string
}

func (err *stringError) Error() string {
	return err.err
}
```

Removing this type fixes the build-time error.

aykevl:
That's definitely not supposed to happen!
How can I reproduce this?

aykevl:
Oh I have a reproducer.
At commit afd4dc222e2eef0d751de4d08e8ef103511fa033, I got:

```
$ tinygo run -target=wasip2 ./testdata/alias.go
runtime.buildVersion: not a string
```

aykevl:
Huh. It looks like LLVM has decided that `{ ptr, i32 }` and `%runtime._string` are the same type. And they do have the same layout in memory. But this is a bit unexpected.

aykevl:
This fixes the bug: https://github.com/tinygo-org/tinygo/pull/4811

PR #4811, opened 2025-03-17
builder: use a separate module for command-line set strings

aykevl:
Strings that are set via the command line (as in `-ldflags="-X ..."`) are now created in a separate module to avoid type renaming issues. LLVM sometimes renames types or merges types that are structurally the same, and putting these strings in a separate module avoids this issue (and lets llvm.LinkModules deal with the difference).

This fixes https://github.com/tinygo-org/tinygo/issues/4810. Even with https://github.com/bytecodealliance/go-modules/pull/318 I still think it's important, since this is a real bug that might happen again in the future.


ydnar:
LGTM

ydnar:
LGTM

PR #4812, opened 2025-03-17
Boehm GC for WebAssembly

aykevl:
See: https://github.com/tinygo-org/tinygo/pull/4385

This is an incomplete port. Things that need to be resolved are:

 - [x] #4385 needs to be merged first, and this PR needs to be rebased.
 - [x] We need to have two separate builds of wasi-libc since the `MALLOC_IMPL` macro is going to be different. I think it's easiest to start doing this build not when creating a TinyGo tarball but when running TinyGo itself (similar to picolibc, musl, etc). This makes it easier to build different variants easily and when needed.

The performance improvements can be large. For example, for the html package some benchmarks are more than 50% faster:

```
$ benchstat html-precise.txt html-boehm.txt
               │ html-precise.txt │           html-boehm.txt           │
               │      sec/op      │   sec/op     vs base               │
Escape               110.29µ ± 0%   73.62µ ± 3%  -33.25% (p=0.002 n=6)
EscapeNone            48.37µ ± 0%   49.15µ ± 1%   +1.62% (p=0.002 n=6)
Unescape             192.31µ ± 5%   84.53µ ± 3%  -56.04% (p=0.002 n=6)
UnescapeNone          2.638µ ± 1%   2.559µ ± 4%        ~ (p=0.132 n=6)
UnescapeSparse        45.44µ ± 2%   18.55µ ± 1%  -59.17% (p=0.002 n=6)
UnescapeDense        182.55µ ± 1%   79.63µ ± 1%  -56.38% (p=0.002 n=6)
geomean               53.11µ        32.40µ       -39.00%
```

Pause times should also be reduced quite a bit, likely much more than the performance improvements. I did not test those, though.

If you want to try this PR, you will have to rebuild wasi-libc:

```
$ rm -rf lib/wasi-libc/sysroot/ lib/wasi-libc/build
$ make wasi-libc
```

(Also note that only wasip1 is supported right now, other wasm variants should be easy to add in the future).

### DIFF
--- a/builder/bdwgc.go
+++ b/builder/bdwgc.go
@@ -14,7 +14,7 @@ var BoehmGC = Library{
 	name: "bdwgc",
 	cflags: func(target, headerPath string) []string {
 		libdir := filepath.Join(goenv.Get("TINYGOROOT"), "lib/bdwgc")
-		return []string{
+		flags := []string{
 			// use a modern environment
 			"-DUSE_MMAP",              // mmap is available
 			"-DUSE_MUNMAP",            // return memory to the OS using munmap
@@ -30,6 +30,7 @@ var BoehmGC = Library{
 			// Use a minimal environment.
 			"-DNO_MSGBOX_ON_ERROR", // don't call MessageBoxA on Windows
 			"-DDONT_USE_ATEXIT",
+			"-DNO_GETENV",
 
 			// Special flag to work around the lack of __data_start in ld.lld.
 			// TODO: try to fix this in LLVM/lld directly so we don't have to
@@ -49,12 +50,13 @@ var BoehmGC = Library{
 
 			"-I" + libdir + "/include",
 		}
+		return flags
 	},
 	needsLibc: true,
 	sourceDir: func() string {
 		return filepath.Join(goenv.Get("TINYGOROOT"), "lib/bdwgc")
 	},
-	librarySources: func(target string) ([]string, error) {
+	librarySources: func(target string, _ bool) ([]string, error) {
 		sources := []string{
 			"allchblk.c",
 			"alloc.c",
--- a/builder/builtins.go
+++ b/builder/builtins.go
@@ -226,7 +226,7 @@ var libCompilerRT = Library{
 		// Development build.
 		return filepath.Join(goenv.Get("TINYGOROOT"), "lib/compiler-rt-builtins")
 	},
-	librarySources: func(target string) ([]string, error) {
+	librarySources: func(target string, _ bool) ([]string, error) {
 		builtins := append([]string{}, genericBuiltins...) // copy genericBuiltins
 		switch compileopts.CanonicalArchName(target) {
 		case "arm":
--- a/builder/library.go
+++ b/builder/library.go
@@ -38,7 +38,7 @@ type Library struct {
 	sourceDir func() string
 
 	// The source files, relative to sourceDir.
-	librarySources func(target string) ([]string, error)
+	librarySources func(target string, libcNeedsMalloc bool) ([]string, error)
 
 	// The source code for the crt1.o file, relative to sourceDir.
 	crt1Source string
@@ -226,7 +226,7 @@ func (l *Library) load(config *compileopts.Config, tmpdir string) (job *compileJ
 
 	// Create jobs to compile all sources. These jobs are depended upon by the
 	// archive job above, so must be run first.
-	paths, err := l.librarySources(target)
+	paths, err := l.librarySources(target, config.LibcNeedsMalloc())
 	if err != nil {
 		return nil, nil, err
 	}
--- a/builder/mingw-w64.go
+++ b/builder/mingw-w64.go
@@ -48,7 +48,7 @@ var libMinGW = Library{
 		}
 		return flags
 	},
-	librarySources: func(target string) ([]string, error) {
+	librarySources: func(target string, _ bool) ([]string, error) {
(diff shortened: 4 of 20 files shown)


deadprogram:
@aykevl you can now rebase this PR against `dev` now that #4385 was merged. Thanks!

aykevl:
A few more changes are needed to make this PR ready, it needs more than just rebasing. I'm working on it!

aykevl:
Should be ready now.
The Boehm GC is disabled by default, it can be used with `-gc=boehm`.

aykevl:
The main drawback right now is that it will often output messages like the following:

```
GC Warning: Repeated allocation of very large block (appr. size 68 KiB):
        May lead to memory leak and poor performance
```

I think the only reasonable fix is to just disable output entirely (see https://github.com/ivmai/bdwgc/pull/721) or disable it using `GC_set_warn_proc` since using `GC_malloc_ignore_off_page` is not safe with Go (inner pointers are allowed, even far into an object, so `GC_malloc_ignore_off_page` is unsafe even if it would probably work for most programs in practice).

deadprogram:
@aykevl now that https://github.com/ivmai/bdwgc/commit/6d10870ba8ca4410276af90c7fc305f1d2949c00 was merged, I think that this PR can proceed?

deadprogram:
@aykevl reminder about this PR :smiley_cat:

aykevl:
Updated the PR.
* rebased on dev
* added `GC_DISABLE_SNPRINTF=1` on WebAssembly to avoid linking in printf related functions (saves ~18kB of binary size)
* disabled warnings on WebAssembly since they happen frequently and there's not much we can do about them.

deadprogram:
@aykevl looks like actual error here: https://github.com/tinygo-org/tinygo/actions/runs/15212794972/job/42791925530?pr=4812#step:17:24

aykevl:
@deadprogram I think it's more likely to be a flaky test. I don't see how this PR should affect windows/386.

aykevl:
@deadprogram actually, you might be right. The Boehm GC is also used on Windows, and this PR updates the bdwgc library.

aykevl:
I can reproduce this issue in my Windows VM (running on ARM!).

aykevl:
I think it's related to https://github.com/ivmai/bdwgc/issues/736. Probably our bug and not one in bdwgc, but asking to be sure.

EDIT: might not be the actual GOOS=386 issue here.

aykevl:
Oh it is actually https://github.com/ivmai/bdwgc/commit/e8aedf635e1ad42102799478ea8a2ab55abd2ba5 that is causing this issue on Windows.

aykevl:
Pushed a workaround to disable bdwgc on Windows/386 for now. I don't think that's a big loss, I think having bdwgc on WASIp1 is much more important.

deadprogram:
@aykevl not quite passing yet https://github.com/tinygo-org/tinygo/actions/runs/15388465010/job/43292271205#step:17:24

aykevl:
Should be fixed now.

deadprogram:
OK, new error https://github.com/tinygo-org/tinygo/actions/runs/15389453659/job/43296579801?pr=4812#step:7:28

aykevl:
I can't reproduce this issue locally. Trying to re-run the tests to see whether it's a flaky one.

deadprogram:
> I can't reproduce this issue locally. Trying to re-run the tests to see whether it's a flaky one.

That time it failed on yet a different test. I restarted one more time...

deadprogram:
It fails on a different test each time, looking like some kind of race? In any case, still fails eventually on some test or another. :crying_cat_face:

deadprogram:
Passing all tests in the last run. Whatever race condition is happening, I do not think it is related to this PR.

deadprogram (approved):
From my point of view, this PR appears ready to merge now.

Anyone have any further feedback before I do so?

aykevl:
I'm not so sure about that. I think you just managed to avoid a somewhat rare crash this time, but the bug is still there.

deadprogram:
> I'm not so sure about that. I think you just managed to avoid a somewhat rare crash this time, but the bug is still there.

@aykevl I don't think that is specific to this PR, do you?

aykevl:
> I don't think that is specific to this PR, do you?

I think it is, since we were also hitting issues on windows/386. I think the windows/386 issues and the flaky builds are related.

Currently trying to reproduce it on my system.

aykevl:
Changed to use the bdwgc commit just before https://github.com/ivmai/bdwgc/commit/e8aedf635e1ad42102799478ea8a2ab55abd2ba5 which I think introduced the Windows issues (GOARCH=386 at least, but I also suspect it is causing the flaky CI issue). See https://github.com/ivmai/bdwgc/issues/736#issuecomment-2929578337 for details. This does mean we lose some of the recent size improvements, but fixing these issues will need a bit more time. To at least have (opt-in) bdwgc support in wasm for this release (even if it grows the binary more than needed) is better than having nothing.

Also, this brings back bdwgc support in GOARCH=386, which isn't super useful but a nice side effect I guess.

deadprogram:
> To at least have (opt-in) bdwgc support in wasm for this release (even if it grows the binary more than needed) is better than having nothing.

Agreed!

aykevl:
And all tests pass in the first CI run, this gives me some confidence I was right about that one bdwgc commit being a problem.

deadprogram:
Thanks for working on this @aykevl and the various needed fixes along the way. Now merging.

PR #4813, opened 2025-03-17
runtime: remove unused file func.go

aykevl:
This was used in the past, but we don't use it anymore.


ydnar:
Would this have been used for `reflect` func support?

deadprogram:
:scissors: :broom: thanks @aykevl now merging.

aykevl:
A long time ago we resolved all indirect calls on WebAssembly to direct calls (by giving every function an ID and then inserting a switch at the point of an indirect function call, basically). This was necessary for our homebrewed Asyncify-like pass which only supported direct calls. But eventually we switched to Asyncify which does support indirect calls, so this support wasn't needed anymore and was removed. Apparently, not all of it was removed.

PR #4820, opened 2025-03-20
Wasilibc builtin

aykevl:
Instead of relying on a build when TinyGo is being built, do it like all other libraries when it is needed.

This brings a few benefits:

  * No more running `make wasi-libc` on the command line as a special case for WebAssembly. The generic `git submodule update --init` step is now enough for wasi-libc.
  * It becomes much easier to customize the build per system. For example: include/exclude malloc as needed, disable/enable bulk memory operations per target, etc.

---

In particular, this PR makes it easier to build wasi-libc with and without malloc support: #4812 needs malloc but I'd rather not enable it in all cases (wasi-libc malloc would conflict with the `malloc` symbol we provide in the TinyGo runtime using our own GC).

To the WebAssembly people (@dgryski, @ydnar, etc): while I think this is a step forward, it might break things. This PR might need some extra testing (though `make test-corpus-wasi` passes so that's something).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,14 +73,6 @@ commands:
             - go-cache-v4-{{ checksum "go.mod" }}
       - llvm-source-linux
       - run: go install -tags=llvm<<parameters.llvm>> .
-      - restore_cache:
-          keys:
-            - wasi-libc-sysroot-systemclang-v7
-      - run: make wasi-libc
-      - save_cache:
-          key: wasi-libc-sysroot-systemclang-v7
-          paths:
-            - lib/wasi-libc/sysroot
       - when:
           condition: <<parameters.fmt-check>>
           steps:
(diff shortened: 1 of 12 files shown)


aykevl:
Rebased to fix a merge conflict.

aykevl:
Rebased again after #4826 got merged.

aykevl:
Would love to get a review on this so that I can finish #4812 (Boehm GC support for WebAssembly).

dgryski:
My only initial concern is with the `unlock` return parameter and ensuring that 1) the lock is undone in case of error and 2) the lock is cleaned up in all other cases, and lookig at the code I think we're fine.  There's the `ok` bool that covers calling unlock if `load` returns and the places that call `load` have `defer unlock()` afterwards.

dgryski (approved):
LGTM

deadprogram:
Thanks for all the work on this @aykevl and to @dgryski for review. Now merging.

PR #4821, opened 2025-03-20
compileopts: re-enable support for GOARCH=wasm

aykevl:
Support for GOARCH=wasm was broken entirely. With this PR, it should work again for supported values of GOOS.

This fixes the following error for example:

    $ make tinygo-test-wasip1
    GOOS=wasip1 GOARCH=wasm /home/ayke/bin/tinygo test cmp compress/lzw compress/zlib [...etc]
    cannot resolve packages: GOARCH=wasm but GOOS is unset. Please set GOOS to wasm, wasip1, or wasip2.
    make: *** [GNUmakefile:489: tinygo-test-wasip1] Error 1

### DIFF
--- a/compileopts/target.go
+++ b/compileopts/target.go
@@ -178,6 +178,21 @@ func (spec *TargetSpec) resolveInherits() error {
 
 // Load a target specification.
 func LoadTarget(options *Options) (*TargetSpec, error) {
+	if options.Target == "" && options.GOARCH == "wasm" {
+		// Set a specific target if we're building from a known GOOS/GOARCH
+		// combination that is defined in a target JSON file.
+		switch options.GOOS {
+		case "js":
+			options.Target = "wasm"
+		case "wasip1":
+			options.Target = "wasip1"
+		case "wasip2":
+			options.Target = "wasip2"
+		default:
+			return nil, errors.New("GOARCH=wasm but GOOS is not set correctly. Please set GOOS to wasm, wasip1, or wasip2.")
+		}
+	}
+
 	if options.Target == "" {
 		return defaultTarget(options)
 	}
--- a/main.go
+++ b/main.go
@@ -1684,24 +1684,9 @@ func main() {
 			usage(command)
 			os.Exit(1)
 		}
-		if options.Target == "" {
-			switch {
-			case options.GOARCH == "wasm":
-				switch options.GOOS {
-				case "js":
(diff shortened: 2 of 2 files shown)


compileopts/target.go
```diff
@@ -179,7 +179,15 @@ func (spec *TargetSpec) resolveInherits() error {
 // Load a target specification.
 func LoadTarget(options *Options) (*TargetSpec, error) {
 	if options.Target == "" {
-		return defaultTarget(options)
+		if options.GOARCH == "wasm" && options.GOOS == "js" {
```
deadprogram: I probably would have written it using a switch:

```go
		switch {
		case options.GOARCH == "wasm" && options.GOOS == "js":
			options.Target = "wasm"
		case options.GOARCH == "wasm" && options.GOOS == "wasip1":
			options.Target = "wasip1"
		case options.GOARCH == "wasm" && options.GOOS == "wasip2":
			options.Target = "wasip2"
		default:
			return defaultTarget(options)
		}
```

compileopts/target.go
```diff
@@ -179,7 +179,15 @@ func (spec *TargetSpec) resolveInherits() error {
...
+			options.Target = "wasm"
+		} else if options.GOARCH == "wasm" && options.GOOS == "wasip1" {
+			options.Target = "wasip1"
+		} else if options.GOARCH == "wasm" && options.GOOS == "wasip2" {
```
ydnar: Will setting `GOOS=wasip2` even work? Will Go choke on the unknown `GOOS`?

aykevl:
Reworked the PR. I realize now why `tinygo test` wasn't working: the special handling for GOOS=wasm was only done for `tinygo build` (not for any of the other subcommands). I've removed the old handling and moved it to LoadTarget which is a more appropriate place.

compileopts/target.go
aykevl: Yes, it will work. I've tested it. GOOS is rewritten to what is set in targets/wasip2.json before `go` commands are invoked.

compileopts/target.go
aykevl: Rewritten this as a switch now, in the updated version of this PR.

deadprogram:

edidim (approved):

edidim (approved):

PR #4823, opened 2025-03-22
windows: add support for the Boehm GC

aykevl:
This adds support for the Boehm GC on Windows, and sets it as the default. It also fixes a rather important bug for the GC on Windows, that I'm surprised we haven't found earlier.


deadprogram:
Thanks @aykevl looking good! Now merging.

PR #4824, opened 2025-03-22
arm64: remove unnecessary .section directive

aykevl:
This directive caused the code to be put in a non-executable area on Windows which caused a segmentation fault. This patch fixes the issue by removing `.section` directives, fixing windows/arm64 support.


deadprogram:
Great find on this one @aykevl thank you very much! Now merging.

PR #4825, opened 2025-03-22
cgo: *actually* fix build warnings on Windows ARM

aykevl:
See: https://github.com/tinygo-org/tinygo/pull/4628

This removes the comment, which actually fixes the issue. #4628 was incorrect.


deadprogram:
Oops I should have noticed that too. Thanks for fixing @aykevl now merging.

PR #4832, opened 2025-03-26
all: shrink bdwgc library

aykevl:
By disabling some configuration options, the library becomes a lot smaller. On Linux, binaries are ~12kb smaller.

This isn't really important on Linux/macOS/Windows, but will become important when we start using bdwgc on WebAssembly for example.

### DIFF
--- a/builder/bdwgc.go
+++ b/builder/bdwgc.go
@@ -30,7 +30,10 @@ var BoehmGC = Library{
 			// Use a minimal environment.
 			"-DNO_MSGBOX_ON_ERROR", // don't call MessageBoxA on Windows
 			"-DDONT_USE_ATEXIT",
-			"-DNO_GETENV",
+			"-DNO_GETENV",          // smaller binary, more predictable configuration
+			"-DNO_CLOCK",           // don't use system clock
+			"-DNO_DEBUGGING",       // reduce code size
+			"-DGC_NO_FINALIZATION", // finalization is not used at the moment
 
 			// Special flag to work around the lack of __data_start in ld.lld.
 			// TODO: try to fix this in LLVM/lld directly so we don't have to
@@ -39,6 +42,8 @@ var BoehmGC = Library{
 
 			// Do not scan the stack. We have our own mechanism to do this.
 			"-DSTACK_NOT_SCANNED",
+			"-DNO_PROC_STAT",  // we scan the stack manually (don't read /proc/self/stat on Linux)
+			"-DSTACKBOTTOM=0", // dummy value, we scan the stack manually
 
 			// Assertions can be enabled while debugging GC issues.
 			//"-DGC_ASSERTIONS",
@@ -63,7 +68,6 @@ var BoehmGC = Library{
 			"blacklst.c",
 			"dbg_mlc.c",
 			"dyn_load.c",
-			"finalize.c",
 			"headers.c",
 			"mach_dep.c",
 			"malloc.c",
@@ -71,7 +75,6 @@ var BoehmGC = Library{
 			"mark_rts.c",
 			"misc.c",
 			"new_hblk.c",
-			"obj_map.c",
 			"os_dep.c",
 			"reclaim.c",
 		}


aykevl:
Another ~16kB (on Linux) can be shrunk with a new option I made a PR for: https://github.com/ivmai/bdwgc/pull/721
With that change, the added binary size of bdwgc (in comparison to `-gc=precise`) is ~27kB on linux/arm64 after stripping. But we'll have to see whether it will get accepted.

builder/bdwgc.go
```diff
@@ -59,7 +65,6 @@ var BoehmGC = Library{
 			"blacklst.c",
 			"dbg_mlc.c",
 			"dyn_load.c",
-			"finalize.c",
```
ivmai, line 66: I think there's no need to exclude finalize.c as it should compile to zero size if GC_NO_FINALIZATION is defined.

builder/bdwgc.go
```diff
@@ -30,6 +30,10 @@ var BoehmGC = Library{
 			// Use a minimal environment.
 			"-DNO_MSGBOX_ON_ERROR", // don't call MessageBoxA on Windows
 			"-DDONT_USE_ATEXIT",
+			"-DNO_CLOCK",           // don't use system clock
+			"-DNO_DEBUGGING",       // reduce code size
+			"-DGC_NO_FINALIZATION", // finalization is not used at the moment
+			"-DNO_GETENV",          // smaller binary, more predictable configuration
```
ivmai: "-DGC_DISABLE_INCREMENTAL" should also help in decreasing code size

ivmai: Also, consider `-D SMALL_CONFIG` (which implies `GC_DISABLE_INCREMENTAL`) and `-D IGNORE_DYNAMIC_LOADING`.

builder/bdwgc.go
aykevl: Correct, but there's no reason to include it either :shrug:

builder/bdwgc.go
aykevl: I already use `-DGC_DISABLE_INCREMENTAL` and `-DIGNORE_DYNAMIC_LOADING`, see a few lines above.
(I might try to enable incremental GC at some point, but I'll need to be careful around handling SIGSEGV etc so it's disabled at the moment. Even with incremental GC disabled, bdwgc is still a lot faster than our own naive GC).

`-D SMALL_CONFIG` docs say:

> Tries to tune the collector for small heap sizes, usually causing it to use less space in such situations. Incremental collection no longer works in this case.

Some people might want to use a large heap size, I'd like to keep that ability. This is a generic configuration, not for a specific program so would like to avoid tuning for a specific configuration.

ivmai: You could define -D LARGE_CONFIG in conjunction with -D SMALL_CONFIG. It might seem confllusing a bit:)

deadprogram:
This work here is quite exciting thanks for helping us out @ivmai

builder/bdwgc.go
ivmai: Another available option (with the latest bdwgc master): `-D NO_FIND_LEAK`

builder/bdwgc.go
```diff
@@ -59,7 +65,6 @@ var BoehmGC = Library{
 			"blacklst.c",
 			"dbg_mlc.c",
```
ivmai, line 69: With the latest bdwgc master: `dbg_mlc.c` is not needed unless you use `GC_debug_*` API.

deadprogram:
@aykevl did you see https://github.com/ivmai/bdwgc/commit/6d10870ba8ca4410276af90c7fc305f1d2949c00 ?

deadprogram:
@aykevl what about this PR, and does it work with PR #4812 ?

aykevl:
My proposal would be to first merge #4812 and after that is done, I can rebase and update this PR with the suggestions from @ivmai.

deadprogram:
@aykevl now that #4812 was merged, perhaps we can get back to this one?

builder/bdwgc.go
ivmai: If you update to latest bdwgc/master, remove obj_map.c from the list below.

dgryski:
Ping @aykevl

deadprogram:
@aykevl any action on this PR? :smile_cat:

aykevl:
Updated the PR!

deadprogram:
Thank you for finishing this one up @aykevl now merging.

ivmai:
BTW. Also consider adding `-DGC_NO_DEINIT` option to exclude a rarely needed feature `GC_deinit` in bdwgc.
The macro was recently [added](https://github.com/bdwgc/bdwgc/commit/7b27a9d40b8370a0754f64a115eae4d65290afe2) to upstream.

aykevl:
@ivmai this is only needed when we update the bdwgc library, right? I'd prefer to keep it as-is at the current commit hash, since it is working now.

ivmai:
> this is only needed when we update the bdwgc library, right? I'd prefer to keep it as-is at the current commit hash, since it is working now.

Right, after updating of the referred commit hash.
Just a reminder.

Issue #4828, opened 2025-03-24
linker could not find symbol ...


jakebailey:
I'm trying to build a version of the Go port of the TypeScript compiler with tinygo. The code is here: https://github.com/microsoft/typescript-go/tree/main/cmd/tsgo

When I just do a regular build, I get:

```console
$ ~/tinygo/bin/tinygo build -o tsgo-tinygo ./cmd/tsgo
/home/jabaile/tinygo/src/internal/futex/futex_linux.c:24: linker could not find symbol syscall
../../../../usr/local/lib/tinygo/lib/bdwgc/misc.c:896: linker could not find symbol strtoul
../../../../usr/local/lib/tinygo/lib/bdwgc/misc.c:1037: linker could not find symbol open
../../../../usr/local/lib/tinygo/lib/bdwgc/misc.c:1128: linker could not find symbol atol
../../../../usr/local/lib/tinygo/lib/bdwgc/misc.c:1139: linker could not find symbol atoi
../../../../usr/local/lib/tinygo/lib/bdwgc/misc.c:1148: linker could not find symbol atol
../../../../usr/local/lib/tinygo/lib/bdwgc/misc.c:1160: linker could not find symbol atoi
../../../../usr/local/lib/tinygo/lib/bdwgc/misc.c:1173: linker could not find symbol atoi
../../../../usr/local/lib/tinygo/lib/bdwgc/os_dep.c:235: linker could not find symbol open
../../../../usr/local/lib/tinygo/lib/bdwgc/os_dep.c:1170: linker could not find symbol open
../../../../usr/local/lib/tinygo/lib/bdwgc/os_dep.c:1200: linker could not find symbol strtoul
../../../../usr/local/lib/tinygo/lib/bdwgc/os_dep.c:2916: linker could not find symbol strtoul
../../../../usr/local/lib/tinygo/lib/bdwgc/os_dep.c:3849: linker could not find symbol fcntl
../../../../usr/local/lib/tinygo/lib/bdwgc/os_dep.c:3946: linker could not find symbol uname
```

But, I definitely have glibc installed such that those should be found.

Am I just "holding it wrong", and there's a different way to build a plain binary for my system?

aykevl:
A `tinygo clean` should fix this.

(We should eventually really fix this issue, but for now `tinygo clean` is a workaround - essentially, you have a stale cache).

jakebailey:
[...]

PR #4837, opened 2025-03-31
Add version to cached C libraries

aykevl:
This may be a bit of a weird place to put the library path, but otherwise it's difficult to get the pathname for the `Config.CFlags` function. So I've put it here.

This should help to avoid stale library caches. The idea is to increment it every time something changes to a library that means it needs to be recompiled. It's a manual process.

See for example: https://github.com/tinygo-org/tinygo/issues/4828

### DIFF
--- a/builder/library.go
+++ b/builder/library.go
@@ -15,6 +15,9 @@ import (
 
 // Library is a container for information about a single C library, such as a
 // compiler runtime or libc.
+//
+// Note: whenever a library gets changed, the version in compileopts/config.go
+// probably also needs to be incremented.
 type Library struct {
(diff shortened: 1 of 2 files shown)


deadprogram:
@aykevl can you please rebase and resolve the merge conflicts on this PR? Thanks!

aykevl:
Rebased!

b0ch3nski:
@aykevl oops!
```
compileopts/config.go:390:13: c.LibcPath undefined (type *Config has no field or method LibcPath)
```

aykevl:
Fixed!

deadprogram:

PR #4838, opened 2025-03-31
Makefile: avoid some shell commands

aykevl:
This avoids 8 shell commands on Windows that were run at all times (and 7 on other OSes). 5 by using `$(OS)` instead of an uname call, and three other commands that were run unconditionally.

This won't make much of a difference on Unix-like operating systems with fast subprocess calls, but it makes a very noticeable difference on Windows. Timing varies a lot, but before this change `make` would often take over 2 seconds just to evaluate the make file:

```
$ time make foo
make: *** No rule to make target 'foo'.  Stop.

real    0m2.454s
user    0m0.015s
sys     0m0.280s
```

With this change, it's a lot faster:

```
$ time make foo
make: *** No rule to make target 'foo'.  Stop.

real    0m0.427s
user    0m0.015s
sys     0m0.265s
```

I think this avoids all shell commands on Windows when they are not needed.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -44,14 +44,10 @@ TINYGO ?= $(call detect,tinygo,tinygo $(CURDIR)/build/tinygo)
 
 # Check for ccache if the user hasn't set it to on or off.
 ifeq (, $(CCACHE))
-    # Use CCACHE for LLVM if possible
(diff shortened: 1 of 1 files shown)


deadprogram:
Thanks for the improvement here @aykevl now merging.

PR #4840, opened 2025-04-01
fix: display all of the current GC options for the -gc flag

deadprogram:
This PR is to display all of the current GC options for the -gc flag when running `tinygo help`


aykevl (approved):
LGTM. The failure in assert-test-linux looks unrelated (if we see this more often, we'd need to investigate what's going on there since it looks like a rather bad issue).

deadprogram:
Thanks @aykevl now merging.

PR #4841, opened 2025-04-01
riscv-qemu: actually sleep in time.Sleep()

aykevl:
Instead of just incrementing the timestamp, this causes the system to actually sleep when calling time.Sleep. The direct effect is that this works as expected:

    $ tinygo run -target=riscv-qemu examples/serial
    hello world!
    hello world!
    hello world!
    [..etc]

This commit also adds a bare bones handler for exceptions (such as invalid memory writes), since we're adding an interrupt handler anyway.

While this patch doesn't add that much functionality, having interrupt support is going to be needed for multicore support on riscv-qemu. My plan is to first add this support to riscv-qemu (based on the earlier work I did for the RP2040 and demoed at FOSDEM 2025) and once the basics are in place and fully tested we can extend this support to the RP2040. Writing for QEMU first makes it much easier to debug any issues that will come up.

### DIFF
--- a/src/runtime/runtime_tinygoriscv_qemu.go
+++ b/src/runtime/runtime_tinygoriscv_qemu.go
@@ -11,32 +11,98 @@ import (
 // This file implements the VirtIO RISC-V interface implemented in QEMU, which
 // is an interface designed for emulation.
 
+// One tick is 100ns by default in QEMU.
+// (This is not a standard, just the default used by QEMU).
 type timeUnit int64
-
-var timestamp timeUnit
 
 //export main
 func main() {
 	preinit()
+
+	// Set the interrupt address.
+	// Note that this address must be aligned specially, otherwise the MODE bits
+	// of MTVEC won't be zero.
+	riscv.MTVEC.Set(uintptr(unsafe.Pointer(&handleInterruptASM)))
+
+	// Enable global interrupts now that they've been set up.
+	// This is currently only for timer interrupts.
+	riscv.MSTATUS.SetBits(riscv.MSTATUS_MIE)
+
 	run()
 	exit(0)
 }
 
+//go:extern handleInterruptASM
+var handleInterruptASM [0]uintptr
+
+//export handleInterrupt
+func handleInterrupt() {
+	cause := riscv.MCAUSE.Get()
+	code := uint(cause &^ (1 << 31))
+	if cause&(1<<31) != 0 {
+		// Topmost bit is set, which means that it is an interrupt.
+		switch code {
+		case riscv.MachineTimerInterrupt:
+			// Signal timeout.
+			timerWakeup.Set(1)
+			// Disable the timer, to avoid triggering the interrupt right after
+			// this interrupt returns.
+			riscv.MIE.ClearBits(riscv.MIE_MTIE)
+		}
+	} else {
+		// Topmost bit is clear, so it is an exception of some sort.
+		// We could implement support for unsupported instructions here (such as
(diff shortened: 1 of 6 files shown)


src/runtime/runtime_tinygoriscv_qemu.go
```diff
@@ -13,30 +13,94 @@ import (
...
+	}
+
+	// Zero MCAUSE so that it can later be used to see whether we're in an
+	// interrupt or not.
+	riscv.MCAUSE.Set(0)
+}
+
 func ticksToNanoseconds(ticks timeUnit) int64 {
-	return int64(ticks)
+	return int64(ticks) * 100
```
eliasnaur: Nit: explain the magic number or name the constant?

src/runtime/runtime_tinygoriscv_qemu.go
```diff
@@ -13,30 +13,94 @@ import (
...
 }
 
 func nanosecondsToTicks(ns int64) timeUnit {
-	return timeUnit(ns)
+	return timeUnit(ns / 100)
```
eliasnaur: Ditto.

src/runtime/runtime_tinygoriscv_qemu.go
```diff
@@ -13,30 +13,94 @@ import (
...
+	code := uint(cause &^ (1 << 31))
+	if cause&(1<<31) != 0 {
+		// Topmost bit is set, which means that it is an interrupt.
+		switch code {
+		case 7: // Machine timer interrupt
+			// Signal timeout.
+			timerWakeup.Set(1)
+			// Disable the timer, to avoid triggering the interrupt right after
+			// this interrupt returns.
+			riscv.MIE.ClearBits(1 << 7) // MTIE bit
```
eliasnaur: Nit: name 1<<7?

src/runtime/runtime_tinygoriscv_qemu.go
```diff
@@ -13,30 +13,94 @@ import (
...
+var handleInterruptASM [0]uintptr
+
+//export handleInterrupt
+func handleInterrupt() {
+	cause := riscv.MCAUSE.Get()
+	code := uint(cause &^ (1 << 31))
+	if cause&(1<<31) != 0 {
+		// Topmost bit is set, which means that it is an interrupt.
+		switch code {
+		case 7: // Machine timer interrupt
```
eliasnaur: Nit: name the constant?

src/runtime/runtime_tinygoriscv_qemu.go
```diff
@@ -13,30 +13,94 @@ import (
...
 
+var timerWakeup volatile.Register8
+
 func sleepTicks(d timeUnit) {
-	// TODO: actually sleep here for the given time.
-	timestamp += d
+	// Enable the timer.
+	target := uint64(ticks() + d)
+	aclintMTIMECMP.Set(target)
+	riscv.MIE.SetBits(1 << 7) // MTIE
```
eliasnaur: Ditto.

aykevl:
@eliasnaur I've added the constants (or explained what they mean in the case of ticks). I've also added the same constants in the other RISC-V chips: fe310, k210, and esp32c3.

eliasnaur:
LGTM, thanks!

deadprogram:
Thank you for working on this @aykevl it is also much more readable now. Thanks @eliasnaur for the reviews! Now merging.

PR #4849, opened 2025-04-11
device/arm: clear pending interrupts before enabling them

eliasnaur:
Without this change, a pending interrupt would spuriously trigger immediately after enabling. This happens if an interrupt is triggered during flashing (e.g. by DMA), which survives the subsequent reset.

This behaviour matches e.g. `machine.irqSet` in machine_rp2_rp2350.go.

See 7f970a45, whose symptoms were likely caused by spurious interrupts.

### DIFF
--- a/builder/sizes_test.go
+++ b/builder/sizes_test.go
@@ -43,8 +43,8 @@ func TestBinarySize(t *testing.T) {
 	tests := []sizeTest{
 		// microcontrollers
 		{"hifive1b", "examples/echo", 4560, 280, 0, 2268},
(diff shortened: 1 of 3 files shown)


deadprogram:
Pinging @aykevl and @soypat for review, please.

soypat (approved):
LGTM. Thanks for the contrib (thats what the kids call it nowadays)

deadprogram:
Now merging, thanks for working on this @eliasnaur and to @soypat for review.

Issue #4751, opened 2025-02-23
stm32 needs USB implementation


deadprogram:
It would be great to have a USB implementation for the STM32 boards.

Who wants this too? Who wants to help? 😸

deadprogram:
What I mean is STM32 implementation(s) similar to https://github.com/tinygo-org/tinygo/blob/dev/src/machine/machine_rp2040_usb.go

aykevl:
I might actually do this for a project of mine :) but need to pick my battles so who knows.
(I've come to like stm32 chips, mostly because they're cheap and flexible, and can use very little power if configured right).

digitalentity:
I might need this soon-ish (sooner than STM32H7xx support). Unless someone is working on this already, I might have a crack at it over the weekend.

digitalentity:
Done, CDC, HID and MSC all verified to work on the STM32F722. I expect it to work on F4 series as well because the USB IP block is the same.

PR #4850, opened 2025-04-11
machine/rp2: expose usb endpoint stall handling

mikesmitty:
Since the USB mass storage PR (#4844) is getting a bit monolithic I figured I'd break out smaller parts of it as I can. I figure I'll work my way through the MCUs that support USB natively, but here's the RP2 changes to start. If there's any suggestions on which MCUs would make the most sense to add support next I'm happy to hear it

### DIFF
--- a/src/machine/machine_rp2040_usb.go
+++ b/src/machine/machine_rp2040_usb.go
@@ -67,7 +67,7 @@ func handleUSBIRQ(intr interrupt.Interrupt) {
 
 		if !ok {
 			// Stall endpoint?
-			sendStallViaEPIn(0)
+			USBDev.SetStallEPIn(0)
 		}
 
 	}
--- a/src/machine/machine_rp2_usb.go
+++ b/src/machine/machine_rp2_usb.go
@@ -21,26 +21,33 @@ func initEndpoint(ep, config uint32) {
 	offset := ep*2*usbBufferLen + 0x100
 	val |= offset
 
+	// Bulk and interrupt endpoints must have their Packet ID reset to DATA0 when un-stalled.
+	epXPIDReset[ep] = false // Default to false in case an endpoint is re-initialized.
+
 	switch config {
 	case usb.ENDPOINT_TYPE_INTERRUPT | usb.EndpointIn:
 		val |= usbEpControlEndpointTypeInterrupt
 		_usbDPSRAM.EPxControl[ep].In.Set(val)
+		epXPIDReset[ep] = true
 
 	case usb.ENDPOINT_TYPE_BULK | usb.EndpointOut:
 		val |= usbEpControlEndpointTypeBulk
 		_usbDPSRAM.EPxControl[ep].Out.Set(val)
 		_usbDPSRAM.EPxBufferControl[ep].Out.Set(usbBufferLen & usbBuf0CtrlLenMask)
 		_usbDPSRAM.EPxBufferControl[ep].Out.SetBits(usbBuf0CtrlAvail)
+		epXPIDReset[ep] = true
 
 	case usb.ENDPOINT_TYPE_INTERRUPT | usb.EndpointOut:
 		val |= usbEpControlEndpointTypeInterrupt
 		_usbDPSRAM.EPxControl[ep].Out.Set(val)
 		_usbDPSRAM.EPxBufferControl[ep].Out.Set(usbBufferLen & usbBuf0CtrlLenMask)
 		_usbDPSRAM.EPxBufferControl[ep].Out.SetBits(usbBuf0CtrlAvail)
+		epXPIDReset[ep] = true
 
 	case usb.ENDPOINT_TYPE_BULK | usb.EndpointIn:
 		val |= usbEpControlEndpointTypeBulk
 		_usbDPSRAM.EPxControl[ep].In.Set(val)
+		epXPIDReset[ep] = true
 
 	case usb.ENDPOINT_TYPE_CONTROL:
 		val |= usbEpControlEndpointTypeControl
@@ -109,7 +116,12 @@ func handleEndpointRx(ep uint32) []byte {
 }
 
 func handleEndpointRxComplete(ep uint32) {
-	epXdata0[ep] = !epXdata0[ep]
+	setEPDataPID(ep, !epXdata0[ep])
+}
+
+// Set the USB endpoint Packet ID to DATA0 or DATA1.
+func setEPDataPID(ep uint32, dataOne bool) {
+	epXdata0[ep] = dataOne
 	if epXdata0[ep] || ep == 0 {
 		_usbDPSRAM.EPxBufferControl[ep].Out.SetBits(usbBuf0CtrlData1Pid)
 	}
@@ -138,7 +150,8 @@ func sendViaEPIn(ep uint32, data []byte, count int) {
 	_usbDPSRAM.EPxBufferControl[ep&0x7F].In.Set(val)
 }
 
-func sendStallViaEPIn(ep uint32) {
+// Set ENDPOINT_HALT/stall status on a USB IN endpoint.
+func (dev *USBDevice) SetStallEPIn(ep uint32) {
 	// Prepare buffer control register value
 	if ep == 0 {
 		armEPZeroStall()
@@ -147,6 +160,37 @@ func sendStallViaEPIn(ep uint32) {
 	_usbDPSRAM.EPxBufferControl[ep&0x7F].In.Set(val)
 	val |= uint32(usbBuf0CtrlStall)
 	_usbDPSRAM.EPxBufferControl[ep&0x7F].In.Set(val)
+}
+
+// Set ENDPOINT_HALT/stall status on a USB OUT endpoint.
+func (dev *USBDevice) SetStallEPOut(ep uint32) {
+	if ep == 0 {
+		panic("SetStallEPOut: EP0 OUT not valid")
+	}
+	val := uint32(usbBuf0CtrlStall)
+	_usbDPSRAM.EPxBufferControl[ep&0x7F].Out.Set(val)
+}
+
+// Clear the ENDPOINT_HALT/stall on a USB IN endpoint.
+func (dev *USBDevice) ClearStallEPIn(ep uint32) {
+	ep = ep & 0x7F
+	val := uint32(usbBuf0CtrlStall)
+	_usbDPSRAM.EPxBufferControl[ep].In.ClearBits(val)
+	if epXPIDReset[ep] {
+		// Reset the PID to DATA0
+		setEPDataPID(ep&0x7F, false)
+	}
+}
+
+// Clear the ENDPOINT_HALT/stall on a USB OUT endpoint.
+func (dev *USBDevice) ClearStallEPOut(ep uint32) {
+	ep = ep & 0x7F
+	val := uint32(usbBuf0CtrlStall)
+	_usbDPSRAM.EPxBufferControl[ep].Out.ClearBits(val)
+	if epXPIDReset[ep] {
+		// Reset the PID to DATA0
+		setEPDataPID(ep, false)
+	}
 }
 
 type usbDPSRAM struct {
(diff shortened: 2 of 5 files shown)


src/machine/machine_rp2040_usb.go
```diff
@@ -77,7 +77,7 @@ func handleUSBIRQ(intr interrupt.Interrupt) {
 
 		if !ok {
 			// Stall endpoint?
-			sendStallViaEPIn(0)
+			SetStallEPIn(0)
```
eliasnaur: I know this is not directly related to the goal of this change, but it seems to me there's quite a lot of duplication between the `machine_rpxxxx_usb.go` files. Perhaps a `machine_rp2_usb.go` would make it easier to do changes now and in future.

mikesmitty: That is a really good point. I'm going to try putting together a PR to see how much can be consolidated

deadprogram:
Now that https://github.com/tinygo-org/tinygo/pull/4856 has been merged, this PR is the next up @mikesmitty 

Can you please resolve the merge conflicts with the `dev` branch?

mikesmitty:
Sure thing, done

deadprogram:
> If there's any suggestions on which MCUs would make the most sense to add support next I'm happy to hear it

The SAMD21/SAMD51 are the next logical addition, in part because we also need more USB implementations! See https://github.com/tinygo-org/tinygo/issues/4751

mikesmitty:
Sounds good, coincidentally I started looking at the USB support in the atsamd21 the other day. I'm still working through the FAT12 emulation layer right now, but I'll come back to it when I'm done

deadprogram:
@eliasnaur any other feedback now on this PR? Looks good to me.

src/machine/usb.go
```diff
@@ -212,6 +213,9 @@ func handleStandardSetup(setup usb.Setup) bool {
 		if setup.WValueL == 1 { // DEVICEREMOTEWAKEUP
 			isRemoteWakeUpEnabled = false
 		} else if setup.WValueL == 0 { // ENDPOINTHALT
+			if usbStallHandler[setup.WIndex&0x7F] != nil {
+				return usbStallHandler[setup.WIndex&0x7F](setup)
```
eliasnaur: `WIndex&0x7f` could use a comment. Will it ever overflow on rogue input?
Also below.

src/machine/machine_rp2_usb.go
```diff
@@ -138,7 +150,8 @@ func sendViaEPIn(ep uint32, data []byte, count int) {
 	_usbDPSRAM.EPxBufferControl[ep&0x7F].In.Set(val)
 }
 
-func sendStallViaEPIn(ep uint32) {
+// Enable ENDPOINT_HALT/stall on a USB IN endpoint
+func SetStallEPIn(ep uint32) {
```
eliasnaur: Can you explain why the `SetStall`* functions must be exported? Can't you infer their calls from the nil-ness of `ep.StallHandler` below?

src/machine/usb.go
```diff
@@ -123,6 +123,7 @@ var (
 	usbTxHandler    [usb.NumberOfEndpoints]func()
 	usbRxHandler    [usb.NumberOfEndpoints]func([]byte)
 	usbSetupHandler [usb.NumberOfInterfaces]func(usb.Setup) bool
+	usbStallHandler [usb.NumberOfEndpoints]func(usb.Setup) bool
```
eliasnaur, line 126: Does the stall handler really need the `usb.Setup` as well?

eliasnaur (commented):
Some comments. I don't know the USB protocol nor the rpxxxx USB hardware, so my comments may well be misguided.

src/machine/machine_rp2_usb.go
mikesmitty: Sure, the short version is that a USB endpoint halt/stall is a form of flow control/out-of-band communication that specific protocols on the device can use to signal to the host when something goes wrong. Individual USB drivers on the device are assigned endpoints (akin to network ports) and the mass storage driver needs to be able to stall those endpoints to notify the host of fatal errors so just the mass storage connection can be reset when needed. In response the host sends a secret handshake to the mass storage driver via a series of USB Setup Packets, the mass storage driver sees this and recognizes the host is back in sync, un-stalls its endpoints and mass storage communication can resume.

If you're interested in more detail about the flow, this is the long version:

In the mass storage Bulk-Only Transfer (BOT) protocol for example the typical communication flow goes like this:
1. USB host sends a CBW (Command Block Wrapper) to the device containing a SCSI command (e.g. a READ command)
2. USB device validates the CBW and if everything checks out it immediately starts sending data from the requested sector/block, packet by packet as the host ACKs them. Each sector/block takes 8 full packets to send so the host will keep accepting raw packets as the requested data until it gets the number of bytes it requested in the original CBW
3. When the device finishes transferring the data it finishes the transaction by sending a CSW (Command Status Wrapper) with the final status of the command and a few other details, then starts waiting for the next CBW to arrive

If the CBW gets corrupted in transit however, the device won't starting sending the requested data to ensure data integrity. There's no frame encapsulating the data, it's just full buffers of data so sending a CSW immediately would be interpreted by the host as part of the response to its READ command. Instead, the device stalls its TX and RX endpoints and refuses to perform any commands. The stall is an actively negotiated status between the host and the device by the underlying host/device USB firmwares so the host will stop listening for the response data and perform a BOT Reset Recovery process (or sometimes take the easy way out, power-cycling the device haha). The Reset Recovery process is three requests sent to the control endpoint (endpoint 0) in this specific order:
1. A "Bulk-Only Mass Storage Reset" request (class-specific)
2. A standard CLEAR_FEATURE request with the ENDPOINT_HALT function, addressed to the stalled mass storage IN (device TX) endpoint. CLEAR_FEATURE is also used for other features, but in this case it's for 
3. Another standard CLEAR_FEATURE request, addressed to the stalled mass storage OUT (device RX) endpoint

This is where the Setup Packets come in, all three of these commands get sent using USB Setup Packets. Typically just for initial negotiation/configuration, but also for recovery procedures. Since the first request is marked as a class-specific request and addressed to the mass storage interface id it gets routed through the usb.SetupConfig.Handler function, but since CLEAR_FEATURE is a standard request it gets routed to machine.handleStandardSetup() where most standard requests are handled. Endpoint 0 stalls/un-stalls a lot in the course of its normal processes so that makes sense, but these specific requests are addressed to the stalled mass storage endpoints with the wIndex field and the mass storage driver needs to receive them so it knows when to un-stall its endpoints. That's the reason behind adding the usb.EndpointConfig.StallHandler

src/machine/usb.go
mikesmitty: Yep, I mentioned it in the novel I wrote just now, but the usbSetupHandler handles class-specific requests addressed to a protocol's interface, but stall handling is done via standard requests and addressed to the protocol endpoints individually

src/machine/usb.go
mikesmitty: Oh, that's a good point. I'll push up a fix in a sec

mikesmitty:
It's interesting that the sizes_test is failing again. I just updated the size for that test from 7359 to 7395 in the previous PR and now it's flipped back

src/machine/machine_rp2_usb.go
mikesmitty: Or to put it less verbosely (apologies for the novel), the mass storage driver calls `SetStall*()` to signal the host of an issue then waits for the host to acknowledge it via Setup Packets to the usb.SetupConfig.Handler and usb.EndpointConfig.StallHandler functions, then the mass storage driver can resume the communications by clearing the stall conditions

src/machine/usb.go
```diff
@@ -212,6 +213,12 @@ func handleStandardSetup(setup usb.Setup) bool {
 		if setup.WValueL == 1 { // DEVICEREMOTEWAKEUP
 			isRemoteWakeUpEnabled = false
 		} else if setup.WValueL == 0 { // ENDPOINTHALT
+			if setup.WIndex < usb.NumberOfEndpoints && usbStallHandler[setup.WIndex&0x7F] != nil {
```
eliasnaur: Can you compare with the number of endpoints without masking the direction out? E.g.

```go
if idx := setup.WIndex & 0x7f; idx < usb.NumberOfEndpoints && usbStallHandler[idx] != nil
```

here and below?

src/machine/machine_rp2_usb.go
eliasnaur: Nice explanation, thank you! (now that you went through the trouble, consider adding (some of) your explanation somewhere in the documentation).

I think I understand the need for the exports now. However, I don't understand why they're not methods on `USBDevice`. I see there's already `machine.ReceiveUSBControlPacket`, so I suppose it's not a new pattern. I do worry locking us into a singleton `USBDevice` will bite our behinds in future.

src/machine/machine_rp2_usb.go
```diff
@@ -149,6 +162,37 @@ func sendStallViaEPIn(ep uint32) {
 	_usbDPSRAM.EPxBufferControl[ep&0x7F].In.Set(val)
 }
 
+// Enable ENDPOINT_HALT/stall on a USB OUT endpoint
```
eliasnaur: Do you mean `Enable` and not `Set` (as in the name) or `Flag`? As far a I understood your description, this is called to tell the hardware and/or the peer that there is a stall. If so, `Enable` contributed to my initial confusion about the purpose of the function.

Nit: end documentation comments with periods.

deadprogram:
@mikesmitty looks like this PR needs to be rebased against the latest `dev`, please.

src/machine/usb.go
mikesmitty: Yep, good callout. Tried to push that one out the door too quickly

src/machine/machine_rp2_usb.go
mikesmitty: I think of them as synonymous, but sure if it's generally more clear I can update it

src/machine/machine_rp2_usb.go
mikesmitty: Sure, that makes sense. I'll update that here in a sec

src/machine/machine_rp2_usb.go
```diff
@@ -149,6 +162,37 @@ func sendStallViaEPIn(ep uint32) {
...
+}
+
+// Clear the ENDPOINT_HALT/stall on a USB IN endpoint.
+func (dev *USBDevice) ClearStallEPIn(ep uint32) {
+	ep = ep & 0x7F
+	val := uint32(usbBuf0CtrlStall)
+	_usbDPSRAM.EPxBufferControl[ep].In.ClearBits(val)
+	if epXPIDReset[ep] {
+		// Reset the PID to DATA0
+		setEPDataPID(ep&0x7F, false)
```
eliasnaur, line 181: Nit: the masking is redundant here.

src/machine/machine_rp2_usb.go
```diff
@@ -149,6 +162,37 @@ func sendStallViaEPIn(ep uint32) {
 	_usbDPSRAM.EPxBufferControl[ep&0x7F].In.Set(val)
 }
 
+// Set ENDPOINT_HALT/stall status on a USB OUT endpoint.
+func (dev *USBDevice) SetStallEPOut(ep uint32) {
+	if ep == 0 {
```
eliasnaur, line 167: Should `ep` be masked before comparing with 0?

eliasnaur (approved):
A question and another nit. LGTM, with the caveat that I don't speak USB :)

src/machine/machine_rp2_usb.go
deadprogram: >  consider adding (some of) your explanation somewhere in the documentation).

Yes please!!

deadprogram:
Thanks for the work on this PR @mikesmitty and to @eliasnaur for the excellent feedback.

Now merging.

eliasnaur:
@mikesmitty I'd still like to resolve the two comments above. Sorry for approving too soon (my intention was to say that I found no more issues with the PR).

src/machine/machine_rp2_usb.go
mikesmitty: Will do, for sure!

src/machine/machine_rp2_usb.go
mikesmitty: The 0 endpoint is a special case. It's both an in and out endpoint in one so there's no direction bit applied for it, but couldn't hurt

mikesmitty:
Sure, I'll make those changes in the main PR

PR #4851, opened 2025-04-12
Baremetal multicore support

aykevl:
Use multiple cores of the "baremetal" riscv-qemu target. The goal is to iron out all the bugs on an easy-to-debug target before moving on to more interesting targets like RP2040. After this PR is done, the RP2040 should be relatively easy to add support for (since I already wrote most of that code).

Work in progress. This already mostly works, but there are a few things that still need to be fixed:

  - [x] ~There's a timing bug where if one goroutine pauses, and another already running goroutine pauses for a shorter time, the second may not be woken up at the earlier time.~ fixed!
  - [x] Timers (`time.NewTimer` etc) need to be implemented.
  - [x] Stack scanning (for the GC) needs to be implemented.

PR #4852, opened 2025-04-12
Small scheduler refactors

aykevl:
Two commits that I extracted from #4851 that shouldn't have much of a practical impact but I think clean up the code a little bit.


eliasnaur:
LGTM. The change is just a rename and pushing the `mainExited` variable into speecific schedulers.

deadprogram:
Thanks for the improvements @aykevl and to @eliasnaur for review. Now merging.

PR #4853, opened 2025-04-12
runtime: implement NumCPU for `-scheduler=threads`

aykevl:
For the threads scheduler, it makes sense to have NumCPU available. For all other schedulers, the number of available CPUs is practically limited to one by the scheduler (even though the system might have more CPUs).

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -972,6 +972,7 @@ endif
 	@cp -rp lib/musl/crt/crt1.c          build/release/tinygo/lib/musl/crt
 	@cp -rp lib/musl/COPYRIGHT           build/release/tinygo/lib/musl
 	@cp -rp lib/musl/include             build/release/tinygo/lib/musl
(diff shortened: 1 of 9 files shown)


deadprogram:
I think there is race condition in this test, since restarting it on a previous PR cause the test to pass:
https://app.circleci.com/pipelines/github/tinygo-org/tinygo/12569/workflows/82555dfc-6731-4e0d-94d2-54fa617489cb/jobs/46524

aykevl:
@deadprogram yeah that looks like a race condition, indeed

aykevl:
@deadprogram this should hopefully fix that race condition: https://github.com/tinygo-org/tinygo/pull/4854
(It is not a bug in the scheduler, it's a bug in the test since it assumed the goroutine would start within 4ms or so which might not be the case especially in CI with systems that are probably heavily loaded).

deadprogram:
Thanks @aykevl now merging.

PR #4854, opened 2025-04-12
ci: try to fix race condition in testdata/goroutines.go

aykevl:
A race condition was possible because the 'acquire' goroutine might not have started in 4 milliseconds which changed the ordering of the test.

This patch fixes it by making sure the goroutine has started (and locked the mutex) before continuing with the test.

### DIFF
--- a/testdata/goroutines.go
+++ b/testdata/goroutines.go
@@ -62,12 +62,15 @@ func main() {
 	time.Sleep(2 * time.Millisecond)
 
 	var m sync.Mutex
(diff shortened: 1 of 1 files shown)


deadprogram:
Good idea, thanks for the fix @aykevl now merging.

deadprogram:
Looks like it did not completely fix it: https://github.com/tinygo-org/tinygo/actions/runs/14421450173/job/40444543566#step:17:56

PR #4855, opened 2025-04-12
runtime: use int64 for `timeUnit` everywhere

aykevl:
This changes `timeUnit` to be int64 everywhere. This required a small change to the wasm (browser) target, and allowed to move the type to a central location to reduce duplication.

### DIFF
--- a/targets/wasm_exec.js
+++ b/targets/wasm_exec.js
@@ -283,12 +283,12 @@
 					},
 				},
 				gojs: {
-					// func ticks() float64
+					// func ticks() int64
 					"runtime.ticks": () => {
-						return timeOrigin + performance.now();
+						return BigInt((timeOrigin + performance.now()) * 1e6);
 					},
 
-					// func sleepTicks(timeout float64)
+					// func sleepTicks(timeout int64)
 					"runtime.sleepTicks": (timeout) => {
 						// Do not sleep, only reactivate scheduler after the given timeout.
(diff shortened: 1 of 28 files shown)


targets/wasm_exec.js
```diff
@@ -283,12 +283,12 @@
 					},
 				},
 				gojs: {
-					// func ticks() float64
+					// func ticks() int64
 					"runtime.ticks": () => {
-						return timeOrigin + performance.now();
+						return BigInt((timeOrigin + performance.now()) * 1e6);
```
eliasnaur, line 288: Is `BigInt` performant enough for frequent calls to `runtime.ticks`? For example, is this line reasonably sure not to allocate?

eliasnaur (approved):
LGTM as this is mostly a refactor. One question about performance below.

targets/wasm_exec.js
deadprogram: > Native support offered by BigInt often leads to better performance due to optimized operations built directly into the JavaScript engine.

I think it will be sufficient.

deadprogram:
Thank you for the refactor @aykevl and to @eliasnaur for review. Now merging.

targets/wasm_exec.js
aykevl: If it turns out to be a performance issue, we can also do the conversion on the Go side. The code will just look a little bit uglier.

PR #4856, opened 2025-04-12
machine/rp2: merge common usb code

mikesmitty:
Originating from a comment by @eliasnaur on #4850, this merges the shared components of the rp2 USB stacks


src/machine/machine_rp2_usb.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+}
+
+// SendUSBInPacket sends a packet for USB (interrupt in / bulk in).
+func SendUSBInPacket(ep uint32, data []byte) bool {
+	sendUSBPacket(ep, data, 0)
+	return true
+}
+
+//go:noinline
```
eliasnaur, line 61: This `noinline` directive probably deserves a comment, or outright removal.
I `git blame`d this directive to the introduction of USB support, in 0bc7c2a61 by @sago35, so I couldn't infer it's meaning.

eliasnaur (approved):
LGTM and thank you. I verified that `machine_rp2_usb.go` is a straightforward copy of code common to the `machine_rp2xxx_usb.go` files.

src/machine/machine_rp2_usb.go
mikesmitty: I did some digging and it looks like this is where it originated: c01f81144e7d7a990cf038e0553352be51749a70
Apparently it was getting inlined repeatedly and causing significant file size increases. Added a note with more info

deadprogram:
Thank you very much for splitting out this PR and for the excellent work overall @mikesmitty this is very exciting!

Also thanks for the detailed reviews @eliasnaur :star: 

Now squash/merging.

mikesmitty:
Thanks!

PR #4858, opened 2025-04-15
add -nobounds (similar to -gcflags=-B)

dgryski:


eliasnaur (approved):

deadprogram:
Thanks for the addition @dgryski and to @eliasnaur for quick review. Now merging.

aykevl:
Hmm, I wonder whether it would perhaps be cleaner to implement `-gcflags=-B` directly? It can be parsed in a similar way to how we parse `-ldflags` already.

PR #4862, opened 2025-04-18
main: change -json flag to match upstream Go

aykevl:
This changes the -json flag for build/run/flash/gdb etc commands to not print `compileopts.Config` but instead match upstream Go and print build errors in a structured way.

For more details, see the proposal: https://github.com/golang/go/issues/62067

---

I plan to build upon this proposal to emit slightly more verbose output that also shows start and end positions in the output. This will be useful for better diagnostics in the TinyGo Playground. See: https://github.com/tinygo-org/playground/pull/26

### DIFF
--- a/compileopts/options.go
+++ b/compileopts/options.go
@@ -53,7 +53,6 @@ type Options struct {
 	Programmer      string
 	OpenOCDCommands []string
 	LLVMFeatures    string
-	PrintJSON       bool
 	Monitor         bool
 	BaudRate        int
 	Timeout         time.Duration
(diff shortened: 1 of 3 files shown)


deadprogram:
This seems fine as long as there is not something that is expecting the current output format. Not sure what that might be?

deadprogram:
Merging since seems good to me. Thanks @aykevl

aykevl:
> This seems fine as long as there is not something that is expecting the current output format. Not sure what that might be?

Yeah not sure, I think a better place to put this information is as extra information in the `tinygo info` subcommand.

PR #4864, opened 2025-04-22
all: add support for LLVM 20

aykevl:
This adds support for the 'go install' case, where we link against the system LLVM (Homebrew, Linux distro version, etc). This does not have an effect on `make`, which still uses LLVM 19 for now until the Espressif Xtensa fork is updated.

The main reason for doing this is because newer distros like Fedora 42 and Homebrew have switched to LLVM 20, so switching to this newer version helps those people. (I'm one of those people, I'm on Fedora 42).

---

Somewhat surprisingly, it seems like everything just works with this newer LLVM version. ~No~ only a few small changes needed to TinyGo itself, for WebAssembly:

  * nontrapping-fptoint was enabled in the wasmbuiltin library used for wasm-unknown. This matches wasm-unknown which already had this enabled, so it doesn't add any new instructions.
  * bulk-memory was enabled in wasi-libc. This was previously enabled in all the other WebAssembly targets (wasm, wasip1, wasip2) so this does't add any new instructions. I think it was also enabled in the wasi-libc build before https://github.com/tinygo-org/tinygo/pull/4820 but I don't know for sure.


deadprogram:
Thanks for updating this @aykevl now merging.

PR #4868, opened 2025-04-23
main: add StartPos and EndPos to `-json` output

aykevl:
This is useful for the TinyGo Playground: with this change it can show the error span like in an IDE, instead of just the (single) error position. Errors become a bit more readable as a result.

It is non-standard however, the Go toolchain doesn't show these positions. But I think they're fairly straightforward and I hope the Go toolchain will follow at some point.

See https://github.com/tinygo-org/playground/pull/26 for what this might look like once integrated in the playground. (Note that we technically only need `EndPos`, so I could remove `StartPos` if preferred).

### DIFF
--- a/diagnostics/diagnostics.go
+++ b/diagnostics/diagnostics.go
@@ -10,6 +10,7 @@ import (
 	"go/types"
 	"io"
 	"path/filepath"
+	"reflect"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


deadprogram:
> It is non-standard however, the Go toolchain doesn't show these positions. But I think they're fairly straightforward and I hope the Go toolchain will follow at some point.

Since they are additional data items and not changing any expected behavior that is not a problem. And this is very useful for the playground.

Now merging thanks @aykevl

PR #4870, opened 2025-04-24
fix: use OpenOCD flash method on microbit v2 when using Nordic Semi SoftDevice

deadprogram:
This PR fixes the Microbit v2 flash method when using Nordic Semi SoftDevice to use OpenOCD. Otherwise, the device actually just locks up and does not run until a hard reset.


deadprogram:
@conejoninja

conejoninja:
It worked, merged, thanks!!

PR #4871, opened 2025-04-24
targets: add target for Microbit v2 with SoftDevice S140 support

deadprogram:
This PR adds a target for Microbit v2 with SoftDevice S140 support, which includes both peripheral and central, albeit at a higher memory cost, leaving less memory for user programs.


deadprogram:
@conejoninja this one too :)

conejoninja:
It worked, merged, thanks!!

Issue #4874, opened 2025-04-26
[WASM] wasm export func block until other goroutinue exit, this behavior not same go,go will exit right now and not block other goroutinue


labulakalia:
env
```
go version go1.24.0 linux/amd64
```
```
tinygo version 0.37.0 linux/amd64 (using go version go1.24.0 and LLVM version 19.1.2)
```
#### greet_wasm.go
```
package main

import (
	"fmt"
	"runtime"
	"time"
)

// main is required for TinyGo to compile to Wasm.
func main() {}

//go:wasmexport greet
func _greet() {
	go func() {
		fmt.Println(time.Now(), "wasm call host func log")
		for i := 0; i < 10; i++ {
			runtime.Gosched()
			fmt.Println(time.Now(), "sleep", i)
			time.Sleep(time.Second)
			runtime.Gosched()
		}
		fmt.Println(time.Now(), "wasm end call host func log")
	}()
	fmt.Println(time.Now(), "start go sched")
	runtime.Gosched()
	time.Sleep(time.Second * 3)
	fmt.Println(time.Now(), "wasm greet exit")

}

```
#### greet.go
```greet.go
package main

import (
	"context"
	"crypto/rand"
	_ "embed"
	"fmt"
	"log"
	"os"

	"github.com/tetratelabs/wazero"
	"github.com/tetratelabs/wazero/imports/wasi_snapshot_preview1"
)

func main() {
	greetWasm, _ := os.ReadFile(os.Args[1])
	// Choose the context to use for function calls.
	ctx := context.Background()

	// Create a new WebAssembly Runtime.
	r := wazero.NewRuntime(ctx)
	defer r.Close(ctx) // This closes everything this Runtime created.

	// Instantiate a Go-defined module named "env" that exports a function to
	// log to the console.
	_, err := r.NewHostModuleBuilder("env").
		Instantiate(ctx)
	if err != nil {
		log.Panicln(err)
	}
	wasi_snapshot_preview1.MustInstantiate(ctx, r)
	conf := wazero.NewModuleConfig().WithStartFunctions("_initialize").
		WithStdout(os.Stdout).
		WithStderr(os.Stderr).
		WithStdin(os.Stdin).
		WithRandSource(rand.Reader).
		WithSysNanosleep().
		WithSysNanotime().
		WithSysWalltime()
	mod, err := r.InstantiateWithConfig(ctx, greetWasm, conf)
	if err != nil {
		log.Panicln(err)
	}
[...]

PR #4875, opened 2025-04-28
wasm: don't block //go:wasmexport because of running goroutines

aykevl:
This fixes bug https://github.com/tinygo-org/tinygo/issues/4874.


deadprogram (approved):

deadprogram:

PR #4876, opened 2025-04-29
machine: [rp2] discount scheduling delays in I2C timeouts

eliasnaur:
The `gosched` call introduces arbitrary long delays in general, and in TinyGo particular because the goroutine scheduler is cooperative and doesn't preempt busy (e.g. compute-heavy) goroutines. This change discounts such delays from the rp2xxx I2C timeout logic.

I tested this change by simulating a busy goroutine:

	go func() {
		for {
			// Busy.
			before := time.Now()
			for time.Since(before) < 100*time.Millisecond {
			}
			// Sleep.
		}
	}()

and testing that I2C transfers would no longer time out.

### DIFF
--- a/src/machine/machine_rp2_i2c.go
+++ b/src/machine/machine_rp2_i2c.go
@@ -86,10 +86,7 @@ func (i2c *I2C) Tx(addr uint16, w, r []byte) error {
 	if i2c.mode != I2CModeController {
 		return ErrI2CWrongMode
 	}
-
-	// timeout in microseconds.
-	const timeout = 40 * 1000 // 40ms is a reasonable time for a real-time system.
-	return i2c.tx(uint8(addr), w, r, timeout)
+	return i2c.tx(uint8(addr), w, r)
 }
 
 // Listen starts listening for I2C requests sent to specified address
@@ -218,8 +215,8 @@ func (i2c *I2C) enable() {
 //
 //go:inline
 func (i2c *I2C) disable() error {
-	const MAX_T_POLL_COUNT = 64 // 64 us timeout corresponds to around 1000kb/s i2c transfer rate.
-	deadline := ticks() + MAX_T_POLL_COUNT
+	const timeout_us = 4_000
+	deadline := ticks() + timeout_us
 	i2c.Bus.IC_ENABLE.Set(0)
 	for i2c.Bus.IC_ENABLE_STATUS.Get()&1 != 0 {
 		if ticks() > deadline {
@@ -285,7 +282,8 @@ func (i2c *I2C) deinit() (resetVal uint32) {
 }
 
 // tx performs blocking write followed by read to I2C bus.
-func (i2c *I2C) tx(addr uint8, tx, rx []byte, timeout_us uint64) (err error) {
+func (i2c *I2C) tx(addr uint8, tx, rx []byte) (err error) {
+	const timeout_us = 4_000
 	deadline := ticks() + timeout_us
 	if addr >= 0x80 || isReservedI2CAddr(addr) {
 		return ErrInvalidTgtAddr
@@ -338,7 +336,9 @@ func (i2c *I2C) tx(addr uint8, tx, rx []byte, timeout_us uint64) (err error) {
 				return errI2CWriteTimeout // If there was a timeout, don't attempt to do anything else.
 			}
 
+			before := ticks()
 			gosched()
+			deadline += ticks() - before
 		}
 
 		abortReason = i2c.getAbortReason()
@@ -361,7 +361,9 @@ func (i2c *I2C) tx(addr uint8, tx, rx []byte, timeout_us uint64) (err error) {
 					return errI2CWriteTimeout
 				}
 
+				before := ticks()
 				gosched()
+				deadline += ticks() - before
 			}
 			i2c.Bus.IC_CLR_STOP_DET.Get()
 		}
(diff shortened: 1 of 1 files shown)


soypat:
These changes may have a negative impact in cases where the timeout is real. Have you tried "refreshing" the deadline between the write and read section of I2C.tx and seeing if that solves the issue as well? (right above rxStart declaration)

```
deadline += timeout_us
```

eliasnaur:
Good idea! Please take another look. I also expanded the commit message.

soypat:
Hol up. I believe the deadline can never trigger now. You are declaring the deadline in-loop, so it is refreshed every time a byte is written over I2C. I suggest going back to the deadline declaration at the start of the function and refreshing it after the write section (right above rxStart declaration). This is because I suspect your program is timing out during the read section due to how the conditions are ordered. If this is not the case let me know

eliasnaur:
What is the deadline for, other than buggy i2c hardware or faulty data lines?

The calling code says

```go
	// timeout in microseconds.
	const timeout = 40 * 1000 // 40ms is a reasonable time for a real-time system.
	return i2c.tx(uint8(addr), w, r, timeout)
```

but TinyGo is no way near "a real-time system" unless there's only one goroutine running.

With your proposed change (deadline extension just before starting read), there's still a risk of unfortunate goroutine scheduling triggering a timeout.

Simplified case, for the transmission loop:

```go
deadline := now() + 40ms
for _, b := range txBuf { // for each byte to transmit
   i2cTransmit(b)
   for i2cTxBufferFull() {
      if now() > deadline { return ... }
      gosched() // Takes, say, 100ms of time.
   }
}
```

As far as I can tell, the second byte will immediately timeout if `i2cTxBufferFull` returns true just once. My proposed change does

```go
for _, b := range txBuf {
   i2cTransmit(b)
   deadline := now() + 40ms
   for i2cTxBufferFull() {
      if now() > deadline { return ... }
      gosched()
   }
}
```

which ensure that no matter how long a `gosched` takes, the timeout will only happen if at least 40ms passed with a full buffer.

soypat:
OK, I've just realized I had a miconception about the looping logic when I asked you to perform both changes. Can we try your original change and doing some basic math to calculate ideal timeout? 

Lets look at a scenario: We have our RP2040 running a slow 10kHz I2C baud. The time it takes for transaction to complete is at least (len(w)+len(r))*bitSendTime where bitSendTime is 1./10000 seconds = 0.1ms = 100us. We can thus set the timeout to be (len(w)+len(r))*100, and we can calculate it correctly. With your original change then we can then discard the time spent between byte transmits (which I had not realized the I2C algorithm could block between every single byte.

soypat:
I'd also add in a security factor, like *2 or something of the sort

eliasnaur:
Done. I didn't add a security factor, because that's already builtin to the pessimistic bus frequency (10kHz).

eliasnaur:
I've pushed an update that replaces the bitrate computation with a simple "long" timeout. I also replaced another tight deadline in `I2C.disable`. I was still seeing timeouts, presumably because of an interrupt handler that can run for a while because of real-time constraints.

Longer deadlines are not great, but my rationale is that the deadlines are only there for faulty hardware or programming errors. And they're not *that* long (40ms). EDIT: finally, the deadlines can be tightened in a future DMA implementation.

src/machine/machine_rp2_i2c.go
```diff
@@ -218,8 +215,8 @@ func (i2c *I2C) enable() {
 //
 //go:inline
 func (i2c *I2C) disable() error {
-	const MAX_T_POLL_COUNT = 64 // 64 us timeout corresponds to around 1000kb/s i2c transfer rate.
-	deadline := ticks() + MAX_T_POLL_COUNT
+	const timeout_us = 40000
```
soypat: Why did you change this line? This loop does not call gosched

soypat: Did it have a noticeable effect? What baud are you running?

eliasnaur: I run i2c at 400kHz.

I had to bump the `disable` deadline otherwise I saw timeouts from it (presumably because of a long interrupt handler in my program). Unlike `gosched`, interrupt handlers can reasonably be assumed to complete in short time (<40ms), but not in (in this case) 64μs.

soypat: I have some reservations jumping the disable timeout by factor ~1000 than what the reference implementation suggests. If a interrup handler runs in over a few ms i believe that to be a real issue for tinygo which usually encourages really short interrupts to not negatively interfere with the runtime

soypat: Can you try 400us and 4ms and see if any of those work?

eliasnaur: I changed the timeouts to 4ms; my interrupt handler completes in ~1.5ms. Long interrupts are not great, but what else can be done about real-time requirements (e.g. interrupt "bottom halves")? TinyGo doesn't offer goroutine priorities.

Just for my curiosity, where do you see the disable deadline in the pico-sdk reference? I couldn't see it in https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_i2c/i2c.c.

soypat: image
Currently on vacations so cant spend too much time right now. Looks like I diverted from the Pico SDK and took the datasheet at its word here.  Prolly get back to this by mid week

eliasnaur: Thank you for the reference.

> Increment POLL_COUNT by one. If POLL_COUNT >= MAX_T_POLL_COUNT, exit with [...] error [...]

reads to me that MAX_T_POLL_COUNT is a loop counter, not a deadline. And 

> Define a maximum time-out paramater, MAX_T_POLL_COUNT

reads to me as the value being implementation defined. Only the loop iteration delay, `t_i2c_poll`, is defined in terms of bus speed.

Am I reading the datasheet wrong here?

soypat: You are correct! Looks like I slipped on the naming

soypat (approved):
LGTM!

PR #4877, opened 2025-04-30
machine: add simulated PWM/timer peripherals

aykevl:
I will soon use these as part of the TinyGo tour, to explain the PWM peripherals on common chips.

### DIFF
--- a/src/machine/machine_generic.go
+++ b/src/machine/machine_generic.go
@@ -48,6 +48,103 @@ func gpioSet(pin Pin, value bool)
 //export __tinygo_gpio_get
 func gpioGet(pin Pin) bool
 
+// Generic PWM/timer peripheral. Properties can be configured depending on the
+// hardware.
+type timerType struct {
+	// Static properties.
+	instance    int32
+	frequency   uint64
+	bits        int
+	prescalers  []int
+	channelPins [][]Pin
+
+	// Configured 'top' value.
+	top uint32
+}
+
+// Configure the PWM/timer peripheral.
+func (t *timerType) Configure(config PWMConfig) error {
+	// Note: for very large period values, this multiplication will overflow.
+	top := config.Period * t.frequency / 1e9
+	if config.Period == 0 {
+		top = 0xffff // default for LEDs
+	}
+
+	// The maximum value that can be stored with the given number of bits in
+	// this timer.
+	maxTop := uint64(1)<<uint64(t.bits) - 1
+
+	// Look for an appropriate prescaler value.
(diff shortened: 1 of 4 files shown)


src/machine/machine_nrf52840_simulator.go
```diff
@@ -0,0 +1,60 @@
+//go:build !baremetal && (bluemicro840 || circuitplay_bluefruit || clue_alpha || feather_nrf52840_sense || feather_nrf52840 || itsybitsy_nrf52840 || mdbt50qrx || nano_33_ble || nicenano || nrf52840_mdk || particle_3rd_gen || pca10056 || pca10059 || rak4631 || reelboard || xiao_ble)
+
+// Simulator support for nrf52840 based boards.
+
+package machine
+
+// Channel values below are nil, so that they get filled in on the first use.
+// This is the same as what happens on barmetal.
```
eliasnaur: Nit: "baremetal"

src/machine/machine_generic.go
```diff
@@ -48,6 +48,103 @@ func gpioSet(pin Pin, value bool)
...
+	}
+
+	// The maximum value that can be stored with the given number of bits in
+	// this timer.
+	maxTop := uint64(1)<<uint64(t.bits) - 1
+
+	// Look for an appropriate prescaler value.
+	var prescaler int
+	for _, div := range t.prescalers {
+		if top/uint64(div) <= maxTop {
```
eliasnaur, line 80: Nit: `if scaled := top/uint64(div); scaled <= maxTop {` and then `top = scaled` below. Not sure the performance and/or the readability gain is worth it, though.

eliasnaur (approved):
LGTM (FWIW).

src/machine/machine_nrf52840_simulator.go
aykevl: Thanks, fixed.

src/machine/machine_generic.go
aykevl: Actually that won't help performance at all. Compilers are very good at recognizing this sort of pattern and only doing the operation once. In LLVM, this is probably done by InstCombine.

(And in any case this code is typically run in WebAssembly in a browser, so performance is less of an issue).

aykevl:
Fixed the typo, and also fixed a bug I found for the atsamd21 (wrong pin in TCC0).

deadprogram:
Thanks for the addition @aykevl and to @eliasnaur for review. Now merging.

PR #4878, opened 2025-04-30
runtime,os: add os.Executable() for Darwin

dgryski:


src/runtime/os_darwin.go
```diff
@@ -229,3 +229,38 @@ func call_syscall6(fn, a1, a2, a3, a4, a5, a6 uintptr) int32
...
+	// next string is exe path
+	argv = (*unsafe.Pointer)(unsafe.Add(unsafe.Pointer(argv), unsafe.Sizeof(argv)))
+
+	cstr := unsafe.Pointer(*argv)
+	length := strlen(cstr)
+	argString := _string{
+		length: length,
+		ptr:    (*byte)(cstr),
+	}
+	executablePath := *(*string)(unsafe.Pointer(&argString))
```
eliasnaur, line 254: Why not `executablePath := unsafe.String((*byte)(cstr), length)`?

eliasnaur (commented):
LGTM, other than a single question below.

aykevl (approved):
LGTM assuming you've tested this

src/runtime/os_darwin.go
aykevl: We still support Go 1.19, and unsafe.String was introduced in Go 1.20.

dgryski: We don't (yet) use `unsafe.String()` in the runtime package, although there is a comment to that effect: https://github.com/tinygo-org/tinygo/blob/3e60eeb368f25f237a512e7553fd6d70f36dc74c/src/runtime/runtime.go#L114

dgryski:
Hmm..
```
~/go/src/github.com/dgryski/bug/exe $ cat main.go
package main

import "os"

func main() {
	p, err := os.Executable()
	if err != nil {
		println("error: ", err)
	} else {
		println("exe: ", p)
	}
}
~/go/src/github.com/dgryski/bug/exe $ tinygo run main.go
exe:  /var/folders/5b/_hr1d1fd3qn4t9vtfx5p61wm0000gp/T/tinygo2514738286/main
~/go/src/github.com/dgryski/bug/exe $ tinygo build
~/go/src/github.com/dgryski/bug/exe $ ./exe
exe:  ./exe
~/go/src/github.com/dgryski/bug/exe $ go run main.go
exe:  /var/folders/5b/_hr1d1fd3qn4t9vtfx5p61wm0000gp/T/go-build1161502719/b001/exe/main
~/go/src/github.com/dgryski/bug/exe $ go build
~/go/src/github.com/dgryski/bug/exe $ ./exe
exe:  /Users/dgryski/go/src/github.com/dgryski/bug/exe/exe
```

Big Go gives the full path but we only get the relative path for a full binary.  I should probably fix that.

src/runtime/os_darwin.go
eliasnaur: I see, thanks. I *did* in fact look up when `unsafe.String` was introduced but then went ahead and assumed 1.20 was more than safe. Out of curiosity, is there any reason for keeping support for such a old version of Go?

deadprogram:
Thanks for the addition @dgryski and to @eliasnaur and @aykevl for reviewing. Now merging.

src/runtime/os_darwin.go
aykevl: We could indeed drop support for this older version by now. One consideration might be that Debian bookworm (current stable) uses Go 1.19, dropping support for that means people will need to upgrade the Go version on Debian.

aykevl:
@dgryski I suspect the result is wrong if you call `os.Chdir` and then call `os.Executable()`.

dgryski:
Oops, oh right.  I guess we need to do the entire resolution at process startup instead. I'll rework this (although it means adding in a hook to `main()` to call any platform-specific argc/argv processing).

dgryski:
Fix for `os.Chdir` issue: https://github.com/tinygo-org/tinygo/pull/4881

PR #4879, opened 2025-05-01
machine: use pointer receiver in simulated PWM peripherals

aykevl:
I only discovered this issue after a while. All the baremetal PWM implementations use pointers to a PWM instance, instead of the PWM instance itself. For consistency (and because it's a better idea in general), the simulated PWMs need to work the same.


deadprogram (approved):
Thanks for the consistency @aykevl I probably should have noticed that.

Now merging.

aykevl:
Thank you!

PR #4881, opened 2025-05-01
os,runtime: make sure executable path is fully resolved at process init

dgryski:


dgryski:
Not sure I like this solution; it pulls in a bunch of code from syscall to get the cwd at process init.

src/runtime/os_darwin.go
```diff
@@ -232,10 +232,18 @@ func call_syscall6X(fn, a1, a2, a3, a4, a5, a6 uintptr) uintptr
 
 //go:linkname os_runtime_executable_path os.runtime_executable_path
 func os_runtime_executable_path() string {
-	argv := (*unsafe.Pointer)(unsafe.Pointer(main_argv))
+	return executablePath
+}
+
+var executablePath string
 
+func platform_argv(argc int32, argv *unsafe.Pointer) {
```
aykevl, line 240: Hmm, I'm not sure I like this name. What about `storeExecutablePath` or something like that?

src/runtime/os_darwin.go
```diff
@@ -264,3 +282,40 @@ func stringsTrimPrefix(s, prefix string) string {
...
+	if len(dir) > 0 && dir[len(dir)-1] == '/' {
+		return dir + name
+	}
+	return dir + "/" + name
+}
+
+// from syscall
+func getcwd() string {
+	const pathMax = 1024
+	var buf [4 * pathMax]byte
```
aykevl, line 296: Why use 1024 and then use 4 times that size in the actual buffer? (Also, it would help to give a pointer to why that is the right value, like a link to a man page or something).

src/runtime/os_darwin.go
```diff
@@ -264,3 +282,40 @@ func stringsTrimPrefix(s, prefix string) string {
...
+	s := libc_getcwd(&buf[0], uint(len(buf)))
+	if s == nil {
+		return ""
+	}
+	n := clen(buf[:])
+	if n < 1 {
+		return ""
+	}
+	return string(buf[:n])
```
aykevl, line 305: You could consider using `unsafe.String` here to avoid an allocation. It should be safe, since `buf` is never modified afterwards. (The potential downside is that `executablePath` will then always be 4kB in size - not sure whether that's a good tradeoff).

aykevl (commented):
This seems like a reasonable solution, even if not perfect (I can't think of a better way right now). Some nits below.

PR #4883, opened 2025-05-03
machine/rp2: unexport machine-specific errors

eliasnaur:
Errors are part of API, and the exported rp2 errors seemed arbitrary. For example, the very particular `ErrRP2040I2CDisable` was exported, but `errI2CWriteTimeout` (which is defined on all platforms) is not.

While here, remove "RP2040" from an error name and make the messages consistent and idiomatic.

### DIFF
--- a/src/machine/machine_rp2_i2c.go
+++ b/src/machine/machine_rp2_i2c.go
@@ -58,15 +58,15 @@ type I2C struct {
 }
 
 var (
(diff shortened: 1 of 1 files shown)


soypat:
I'm in support of this change but I'd like to hear from @deadprogram as well

deadprogram:
Thanks for cleaning this up here @eliasnaur and to @soypat for review. Now merging.

PR #4657, opened 2024-12-14
compiler: report error instead of crashing on missing function body

aykevl:
This can happen with generic functions, see: https://github.com/tinygo-org/tinygo/issues/4486

PR #4885, opened 2025-05-03
go.*: upgrade golang.org/x/tools to v0.30.0, Go to 1.22

### LINKED ISSUE #4884: miscompilation of a break after a range-loop
This program runs forever on TinyGo:

```go
package main

import (
	"fmt"
	"iter"
)

func splitSeq() iter.Seq[string] {
	return func(yield func(string) bool) {
	}
}

func main() {
	fmt.Println("start")
	defer fmt.Println("finish")
	for {
		for range splitSeq() {
		}
		break
	}
}
```

To reproduce:

```sh
$ go run stub2.go
start
finish
$ tinygo version
tinygo version 0.38.0-dev-612a38e3 darwin/arm64 (using go version go1.24.2 and LLVM version 18.1.8)
$ tinygo run stub2.go
start
<hang>
```

Adding an explicit break target fixes the hang:

```go
package main

import (
	"fmt"
	"iter"
)

func splitSeq() iter.Seq[string] {
	return func(yield func(string) bool) {
	}
}

func main() {
	fmt.Println("start")
	defer fmt.Println("finish")
loop:
	for {
		for range splitSeq() {
		}
		break loop
	}
}
```

```
$ tinygo run stub3.go
start
finish
```

eliasnaur:
Fixes #4884 by upgrading the ssa package.

The fix is in v0.26.0, which also bumps the minimum Go to 1.22. The latest x/tools module still depending on 1.22 is v0.30.0.


eliasnaur:
The builders fail on the same error as described in https://github.com/tinygo-org/tinygo/pull/4657#issuecomment-2566343500.

eliasnaur:
Commented out `SanityCheckFunctions` and left a note about https://github.com/golang/go/issues/73594.

eliasnaur:
The [`test-llvm15-go119`](https://app.circleci.com/pipelines/github/tinygo-org/tinygo/12643/workflows/f12b913a-8403-4465-8fbd-0b7c8ce9f705/jobs/46676) job fails, but I see nowhere to bump the builder to Go 1.22.

deadprogram:
I think you are looking for https://github.com/tinygo-org/tinygo/blob/dev/.circleci/config.yml#L97

eliasnaur:
Indeed, thanks. I forgot how to `rg` for hidden directories 🤦.

deadprogram:
I am generally supportive of this PR. Not sure if anyone else has concerns about dropping support for such old Go versions?

dkegel-fastly:
Dropping 1.21 now seems fine.

As a tinygo user, I would feel comfortable if tinygo supported not just the two currently supported versions of go, but also the previous one. 

e.g. when tinygo releases support for go 1.25, I would be nervous if it dropped go 1.23, but comfortable with dropping go 1.22.

This is based on my own experience keeping downstream projects in all sorts of languages running over the years, and is admittedly somewhat paranoid.

deadprogram:
Totally agree with your comments @dkegel-fastly 

> “Just because you're paranoid doesn't mean they aren't after you.”
― Joseph Heller, [Catch-22](https://www.goodreads.com/work/quotes/814330)

deadprogram:
Merging in the interest of keeping things moving along. Thank you @eliasnaur for the PR and to @dkegel-fastly for feedback.

Issue #4072, opened 2024-01-05
Missing runtime.Pinner


inkeliz:
I'm using the version 30.0:
> tinygo version 0.30.0 linux/amd64 (using go version go1.21.1 and LLVM version 16.0.1)

But seems that it's missing the new `runtime.Pinner`, which was introduced in Go 1.21:

```
undefined: runtime.Pinner
```

PR #4889, opened 2025-05-08
transform: PtrToInt and IntToPtr do not escape

ysoldak:
This hints compiler that constructs like `uintptr(unsafe.Pointer(&x))` do not escape. Handled by `llvm.PtrToInt` case.
(I've also added `llvm.IntToPtr`, for completeness, but I'm not sure what code matches it. Feel free to reject.)

### Disclaimer

This change may very well be completely wrong; can such constructs escape?
Hence a draft, merely a conversation starter.

### Context

I2C implementations on RP2040 and NRF52 behave differently regarding allocation behavior.
NRF52 implementation escapes to heap while RP2040 does not.
The issue was tracked down to lines 52 and 61 of this file: https://github.com/tinygo-org/tinygo/blob/release/src/machine/machine_nrf528xx.go

### Reproducer
```
$ tinygo build -target=feather-nrf52840 -print-allocs=main -o test.uf2 ./src/examples/i2c-target
.../tinygo/src/examples/i2c-target/main.go:101:13: object allocated on the heap: escapes at line 105
.../tinygo/src/examples/i2c-target/main.go:77:43: object allocated on the heap: escapes at line 77
.../tinygo/src/examples/i2c-target/main.go:76:12: object allocated on the heap: escapes at line 77
.../tinygo/src/examples/i2c-target/main.go:66:43: object allocated on the heap: escapes at line 66
.../tinygo/src/examples/i2c-target/main.go:57:43: object allocated on the heap: escapes at line 57
.../tinygo/src/examples/i2c-target/main.go:56:13: object allocated on the heap: escapes at line 57
```

And this does not escape at all
```
$ tinygo build -target=nano-rp2040 -print-allocs=main -o test.uf2 ./src/examples/i2c-target
```

~After this change both do not escape anymore.~
I was over-optimistic and probably tired, it still escapes in NRF52 case, but I don't understand why.

Thanks @dgryski for pointing at `allocs.go`

### DIFF
--- a/transform/allocs.go
+++ b/transform/allocs.go
@@ -160,6 +160,10 @@ func valueEscapesAt(value llvm.Value) llvm.Value {
 		case llvm.ICmp:
 			// Comparing pointers don't let the pointer escape.
 			// This is often a compiler-inserted nil check.
+		case llvm.PtrToInt:
+			// Pointer to int conversion doesn't let the pointer escape.
+		case llvm.IntToPtr:
+			// Int to pointer conversion doesn't let the pointer escape.
 		default:
 			// Unknown instruction, might escape.
 			return use
--- a/transform/testdata/allocs2.go
+++ b/transform/testdata/allocs2.go
@@ -1,11 +1,13 @@
 package main
 
+import "unsafe"
+
 func main() {
 	n1 := 5
 	derefInt(&n1)
 
 	// This should eventually be modified to not escape.
-	n2 := 6 // OUT: object allocated on the heap: escapes at line 9
+	n2 := 6 // OUT: object allocated on the heap: escapes at line 11
 	returnIntPtr(&n2)
 
 	s1 := make([]int, 3)
@@ -15,7 +17,7 @@ func main() {
 	readIntSlice(s2[:])
 
 	// This should also be modified to not escape.
-	s3 := make([]int, 3) // OUT: object allocated on the heap: escapes at line 19
+	s3 := make([]int, 3) // OUT: object allocated on the heap: escapes at line 21
 	returnIntSlice(s3)
 
 	useSlice(make([]int, getUnknownNumber())) // OUT: object allocated on the heap: size is not constant
@@ -23,14 +25,14 @@ func main() {
 	s4 := make([]byte, 300) // OUT: object allocated on the heap: object size 300 exceeds maximum stack allocation size 256
 	readByteSlice(s4)
 
-	s5 := make([]int, 4) // OUT: object allocated on the heap: escapes at line 27
+	s5 := make([]int, 4) // OUT: object allocated on the heap: escapes at line 29
 	_ = append(s5, 5)
 
 	s6 := make([]int, 3)
 	s7 := []int{1, 2, 3}
 	copySlice(s6, s7)
 
-	c1 := getComplex128() // OUT: object allocated on the heap: escapes at line 34
+	c1 := getComplex128() // OUT: object allocated on the heap: escapes at line 36
 	useInterface(c1)
 
 	n3 := 5
@@ -38,13 +40,13 @@ func main() {
 		return n3
 	}()
 
-	callVariadic(3, 5, 8) // OUT: object allocated on the heap: escapes at line 41
+	callVariadic(3, 5, 8) // OUT: object allocated on the heap: escapes at line 43
 
-	s8 := []int{3, 5, 8} // OUT: object allocated on the heap: escapes at line 44
+	s8 := []int{3, 5, 8} // OUT: object allocated on the heap: escapes at line 46
 	callVariadic(s8...)
 
-	n4 := 3 // OUT: object allocated on the heap: escapes at line 48
-	n5 := 7 // OUT: object allocated on the heap: escapes at line 48
+	n4 := 3 // OUT: object allocated on the heap: escapes at line 50
+	n5 := 7 // OUT: object allocated on the heap: escapes at line 50
 	func() {
 		n4 = n5
 	}()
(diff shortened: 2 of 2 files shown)


ysoldak:
In `machine` package we have many cases of `uintptr(unsafe` use.
See https://github.com/search?q=repo%3Atinygo-org%2Ftinygo%20uintptr(unsafe%20path%3Asrc%2Fmachine&type=code

Each of such cases potentially unnecessarily allocates.

dgryski:
From a conservation on Slack:

I think we need a bit more; for example if you have a pointer conversion between two pointer types, we want to make sure the first one "escapes" if the second pointer "escapes" too.
That's probably why the official unsafe rules require all pointer conversions to happen in a single expression, so there's no requirement to track "does this int eventually become a pointer".
This code pattern with `IntToPtr` will happen with code predating https://pkg.go.dev/unsafe#Add.

eliasnaur:
As far as I can read the NRF code `i2c.Bus.TXD.PTR.Set(uint32(uintptr(unsafe.Pointer(&w[0]))))` fundamentally *does* escape the pointer, no? I realize that `Tx` does wait for the DMA transfer to complete, but the `TinyGo` compiler doesn't know that.

In other words, I can't see how this optimization can be done even with perfect tracking of the `uintptr` value.

As a side note, I believe code such as the NRF `i2c.Tx` technically needs a `runtime.Pinner` to keep the `w` and `r` buffers pinned during the DMA transfer. Because of https://github.com/tinygo-org/tinygo/issues/4072 and because TinyGo doesn't move pointers, a `runtime.KeepAlive` is probably enough.

aykevl:
> As far as I can read the NRF code `i2c.Bus.TXD.PTR.Set(uint32(uintptr(unsafe.Pointer(&w[0]))))` fundamentally _does_ escape the pointer, no? I realize that `Tx` does wait for the DMA transfer to complete, but the `TinyGo` compiler doesn't know that.
> 
> In other words, I can't see how this optimization can be done even with perfect tracking of the `uintptr` value.

Agreed. The value does escape here.

We could possibly do something tricky like telling the compiler "please keep the value alive until this point (after the DMA transfer completed), but don't consider this pointer escaped". That would be correct, and also avoid the heap allocation. But I'm not entirely sure how to do that - perhaps some tricks with inline assembly can be used.

EDIT: actually, `runtime.KeepAlive` does exactly what I describe above. And it is already implemented in TinyGo. The only thing missing is a way to tell the compiler "this PtrToInt doesn't escape".

eliasnaur:
> > As far as I can read the NRF code `i2c.Bus.TXD.PTR.Set(uint32(uintptr(unsafe.Pointer(&w[0]))))` fundamentally _does_ escape the pointer, no? I realize that `Tx` does wait for the DMA transfer to complete, but the `TinyGo` compiler doesn't know that.
> > In other words, I can't see how this optimization can be done even with perfect tracking of the `uintptr` value.
> 
> Agreed. The value does escape here.
> 
> We could possibly do something tricky like telling the compiler "please keep the value alive until this point (after the DMA transfer completed), but don't consider this pointer escaped". That would be correct, and also avoid the heap allocation. But I'm not entirely sure how to do that - perhaps some tricks with inline assembly can be used.

Isn't that what `runtime.Pinner` is for? That is, fixing #4072 will allow low-level code to extend a pointer's lifetime, but shouldn't interfere with escape analysis. Longer term, `runtime.Pinner` will mark a pointer pinned for a future moving garbage collector.

eliasnaur:
> EDIT: actually, `runtime.KeepAlive` does exactly what I describe above. And it is already implemented in TinyGo. The only thing missing is a way to tell the compiler "this PtrToInt doesn't escape".

Pedantically, `runtime.KeepAlive` doesn't say that the pointer address should be held constant, only that finalizers won't run. DMA requires pinned addresses, as provided by `runtime.Pinner`.

aykevl:
> Isn't that what `runtime.Pinner` is for?

No, that's to avoid moving the object (which doesn't currently happen and realistically is unlikely to happen in the future). A quick look at `runtime.Pinner` seems to show that the object will likely escape.
What we actually want is `runtime.KeepAlive`. See the edit to my previous comment.

aykevl:
> Pedantically, `runtime.KeepAlive` doesn't say that the pointer address should be held constant, only that finalizers won't run. DMA requires pinned addresses, as provided by `runtime.Pinner`.

Well you are _technically_ correct. Do note that `runtime.KeepAlive` also promises to keep the object alive, it just says nothing about the address:

> KeepAlive marks its argument as currently reachable. This ensures that the object is not freed, and its finalizer is not run, before the point in the program where KeepAlive is called.

In TinyGo, I think it is unlikely we'll ever have a moving garbage collector that would require pinning. Not just because of DMA, but also because it means interrupts can't run while the GC moves the heap.

eliasnaur:
> which doesn't currently happen and realistically is unlikely to happen in the future

I've seen this comment before, but I'd like to say I find the current garbage collector too unreliable for dynamic allocation because of memory fragmentation. I frequently run into out-of-memory issues before even half of available memory (512kB) has been filled because of fragmentation. Coupled with the escape analyzer missing key allocations I find myself spending significant effort (and API contortions) to avoid allocations altogether.

I don't see any other way of fixing fragmentation and making the GC reliable than making it moving.

eliasnaur:
> In TinyGo, I think it is unlikely we'll ever have a moving garbage collector that would require pinning. Not just because of DMA, but also because it means interrupts can't run while the GC moves the heap.

DMA code as well as interrupts can mark objects pinned, no? I realize this is not done today, but I don't see a fundamental reason that's impossible.

ysoldak:
This was just an experiment and did not actually fix the problem I've been facing in my project.

PR #4890, opened 2025-05-09
runtime: avoid an allocation in (*time.Timer).Reset

eliasnaur:


dgryski (approved):
LGTM.  My only question is if we can just reuse the timer struct as-is and we don't need to reset or update any other fields.

eliasnaur:
Good point. PTAL.

deadprogram:
Looks good now, thanks for the improvement @eliasnaur and to @dgryski for review/feedback.

Now merging.

src/runtime/scheduler_cooperative.go
```diff
@@ -117,11 +117,11 @@ func addTimer(tim *timerNode) {
 
 // removeTimer is the implementation of time.stopTimer. It removes a timer from
 // the timer queue, returning true if the timer is present in the timer queue.
```
aykevl, line 119: This comment is now incorrect: it doesn't return a bool anymore.

aykevl (commented):
A bit late, LGTM but there's a small nit in that the documentation now doesn't match the function. Might be a good idea to fix in a separate PR if you want.

src/runtime/scheduler_cooperative.go
eliasnaur: https://github.com/tinygo-org/tinygo/pull/4902

PR #4898, opened 2025-05-21
wasm: fix C realloc and optimize it a bit

aykevl:
- Do not use make([]byte, ...) to allocate, instead allocate a slice of pointers. This makes sure the precise GC will scan the contents of the allocation, since C could very well put pointers in there.
  - Simplify the map to use the pointer as the key and the size as the value, instead of storing the slices directly in the map.

---

I didn't really test this apart from running ./testdata/cgo. I'm not entirely sure how to properly test this. But if we start using libc_realloc for cabi_realloc I think @ydnar has some test cases on hand?

### DIFF
--- a/compiler/map.go
+++ b/compiler/map.go
@@ -250,6 +250,9 @@ func (b *builder) createMapIteratorNext(rangeVal ssa.Value, llvmRangeVal, it llv
 func hashmapIsBinaryKey(keyType types.Type) bool {
 	switch keyType := keyType.Underlying().(type) {
 	case *types.Basic:
+		// TODO: unsafe.Pointer is also a binary key, but to support that we
+		// need to fix an issue with interp first (see
+		// https://github.com/tinygo-org/tinygo/pull/4898).
 		return keyType.Info()&(types.IsBoolean|types.IsInteger) != 0
 	case *types.Pointer:
 		return true
--- a/src/runtime/arch_tinygowasm_malloc.go
+++ b/src/runtime/arch_tinygowasm_malloc.go
@@ -8,16 +8,21 @@ import "unsafe"
 // code linked from other languages can allocate memory without colliding with
 // our GC allocations.
 
-var allocs = make(map[uintptr][]byte)
+// Map of allocations, where the key is the allocated pointer and the value is
+// the size of the allocation.
+// TODO: make this a map[unsafe.Pointer]uintptr, since that results in slightly
+// smaller binaries. But for that to work, unsafe.Pointer needs to be seen as a
+// binary key (which it is not at the moment).
+// See https://github.com/tinygo-org/tinygo/pull/4898 for details.
+var allocs = make(map[*byte]uintptr)
 
 //export malloc
 func libc_malloc(size uintptr) unsafe.Pointer {
 	if size == 0 {
 		return nil
 	}
-	buf := make([]byte, size)
-	ptr := unsafe.Pointer(&buf[0])
-	allocs[uintptr(ptr)] = buf
+	ptr := alloc(size, nil)
+	allocs[(*byte)(ptr)] = size
 	return ptr
 }
 
@@ -26,8 +31,8 @@ func libc_free(ptr unsafe.Pointer) {
 	if ptr == nil {
 		return
 	}
-	if _, ok := allocs[uintptr(ptr)]; ok {
-		delete(allocs, uintptr(ptr))
+	if _, ok := allocs[(*byte)(ptr)]; ok {
+		delete(allocs, (*byte)(ptr))
 	} else {
 		panic("free: invalid pointer")
 	}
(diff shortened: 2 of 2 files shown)


ydnar:
The Component Model never calls free on these allocations. We'd need some way to assert ownership of them so the GC can free them later.

deadprogram:
Looks like real test failure here https://github.com/tinygo-org/tinygo/actions/runs/15159995695/job/42637182590#step:17:28

ydnar:
> The Component Model never calls free on these allocations. We'd need some way to assert ownership of them so the GC can free them later.

It looks like we can have the generated code call `runtime.libc_free` on any allocated pointers it knows about, since `libc_free` doesn’t actually free the memory. I think I can wire this up into package `cm` with a small TinyGo-specific code path.

@aykevl would it help to have `libc_free` call some inner function `release` or something, that explicitly doesn’t *free* the memory, doing what it does now, allowing the GC to clean it up later?

aykevl:
Huh, this is weird. This PR increases binary size by a large amount for wasm. For example, testdata/alias.go went from 17kb to 49kb. I don't know why this PR causes such a large increase, but I am pretty sure that we wouldn't have this problem if `malloc` and `free` weren't exported by default on wasm.

aykevl:
> The Component Model never calls free on these allocations. We'd need some way to assert ownership of them so the GC can free them later.

@ydnar so the host calls `cabi_realloc` (implicitly claiming ownership of the memory) and then at some point in the future transfers ownership to the guest/Go size? Is that correct? (Or at least the spec says something like "we won't use the memory after this point"?)

ydnar:
Transfer is implicit. There is not a specific call for each pointer that transfers ownership.

The component model does specify a post-call mechanism for exported functions to allow the host to free allocs.

This is where we can call libc_free.

ydnar:
> Huh, this is weird. This PR increases binary size by a large amount for wasm. For example, testdata/alias.go went from 17kb to 49kb. I don't know why this PR causes such a large increase, but I am pretty sure that we wouldn't have this problem if `malloc` and `free` weren't exported by default on wasm.

Can this split the implementation from the `//export`?

ydnar:
@aykevl for this to support the Component Model, I think we need a function like `libc_free` that:

1. Removes the pointer from the `allocs` map.
2. Does not panic if it’s an unknown pointer.

Can it be something like `runtime.Release(ptr unsafe.Pointer) unsafe.Pointer` that returns `nil` if the pointer was unknown?

dgryski:
I'll try to help out here.

deadprogram:
It would be very good to get this fix, or some version of it, merged in time for the next release... :smile_cat:

deadprogram:
I think this PR will pass CI with this change:

```
diff --git a/builder/sizes_test.go b/builder/sizes_test.go
index 931e9970..e4d5396a 100644
--- a/builder/sizes_test.go
+++ b/builder/sizes_test.go
@@ -81,8 +81,9 @@ func TestSizeFull(t *testing.T) {
                "wasip1",
        }
 
-       libMatch := regexp.MustCompile(`^C [a-z -]+$`) // example: "C interrupt vector"
-       pkgMatch := regexp.MustCompile(`^[a-z/]+$`)    // example: "internal/task"
+       libMatch := regexp.MustCompile(`^C [a-z -]+$`)      // example: "C interrupt vector"
+       goMatch := regexp.MustCompile(`^Go [a-z -]+$`)      // example: "Go interface method"
+       pkgMatch := regexp.MustCompile(`^[a-z][a-z/0-9]+$`) // example: "internal/task"
 
        for _, target := range tests {
                target := target
@@ -106,6 +107,9 @@ func TestSizeFull(t *testing.T) {
                                if libMatch.MatchString(pkg) {
                                        continue
                                }
+                               if goMatch.MatchString(pkg) {
+                                       continue
                                if pkgMatch.MatchString(pkg) {
                                        continue
                                }
```

aykevl:
@deadprogram it would, but that's not the problem. The problem is that it increases binary size for wasm substantially. I'll need to investigate why that is.

deadprogram:
> The problem is that it increases binary size for wasm substantially. I'll need to investigate why that is.

Thanks for the clarification.

aykevl:
I think I fixed the issue. The problem was that `unsafe.Pointer` was not treated like all the other pointers and forced map lookups using reflection. This fix might also optimize other code, we'll see.

aykevl:
Also, I optimized the code slightly by calling `alloc` directly instead of doing the weird round-up-and-divide to allocate a slice of pointers. The code is simpler now.

deadprogram:
@aykevl still error in this PR failing CI:

```shell
panic: cannot convert pointer value to byte
```

deadprogram:
@aykevl any chance of getting back to this PR soon? I think it is kinda needed...

aykevl:
Rebased since it's been a while and I want to see why CI fails (logs were expired).

aykevl:
Removed the first commit that was causing problems.

aykevl:
Actually no I think it is needed.

aykevl:
Ok, this should do it for now. Not pretty, but fixing the interp issue is more work. We can do that later.

deadprogram:
@dgryski @ydnar anyone else WASM oriented please take a look.

deadprogram:
Last call for comments! :smile_cat:

deadprogram:
OK, now merging. Thanks for working on this @aykevl

PR #4900, opened 2025-05-21
machine: don't inline RTT WriteByte everywhere

aykevl:
With `-opt=2`, WriteByte gets inlined everywhere a println statement exists. This blows up binary size for very little gain. In my case, the binary size roughly doubled. Instead, don't inline it so that the binary size remains somewhat reasonable. This might slow down WriteByte a tiny bit, but likely not by any significant amount.


eliasnaur (approved):
LGTM. Thanks.

deadprogram:
Thanks for the improvement @aykevl and to @eliasnaur for review. Now merging.

PR #4901, opened 2025-05-21
wasm: refactor/modify stub signal handling

aykevl:
This commit changes signal handling in a few ways:

  * It stubs signals for all wasm targets (not just wasi) and baremetal, since none of those have traditional POSIX signals. And moves the code for that into a single file, instead of duplicating it.
  * It removes the stub for signal_ignored since the value `false` might be wrong in some cases and it doesn't usually seem to be called (it is not called in tsgo). Should be trivial to re-add if it is shown to be needed.
  * It adds a stub for `os/signal.signalWaitUntilIdle` which _is_ called by tsgo.

### DIFF
--- a/src/runtime/runtime_wasip1.go
+++ b/src/runtime/runtime_wasip1.go
@@ -149,20 +149,3 @@ type (
 		}
 	}
 )
-
-// Need to stub these for `os/signal`, but wasi doesn't support signals.
-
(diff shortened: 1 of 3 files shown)


dgryski (approved):
LGTM

aykevl:
Thank you for the review!

PR #4902, opened 2025-05-21
runtime: update comment

eliasnaur:
I forgot to change the comment in PR #4890.

CC @aykevl


deadprogram (approved):
Thanks for the correction here @eliasnaur

Issue #4690, opened 2025-01-06
RP2350: SPI with DMA seems not working


advancedlogic:
First of all, thanks for the amazing effort to develop TinyGo.
I am creating a library to experiment with the Geeek Pico Breadboard [https://github.com/geeekpi/pico_breadboard_kit](url) using the RP PICO 2 (RP2350).

- NeoPixel: OK
- Joystick:  OK
- Leds: OK
- Buzzer (Speaker): KO, PWM seems not yet available for RP PICO 2
- Display (ST7796U): OK but without DMA
- Touch: Not implemented yet

I was able to modify the driver ST7789 to work with the ST7796U (480x320) but if I try to use the method Tx(w,r []byte) with more than 1 byte, the PICO 2 freezes during the initialization process and I did not find a way to understand why. Everything works if I send each byte using Transfer (but it's slow for a decent refresh rate). - [https://github.com/tinygo-org/tinygo/blob/release/src/machine/machine_rp2_spi.go](url)

I also tried with an ST7789 using the driver available [https://github.com/tinygo-org/drivers/tree/release/st7789](url), with the RP PICO 2 and it freezes during the initialization process.

It seems related to the DMA access. In the RP2350 Datasheet it seems that for payloads less than 250MB it should be retro-compatible with RP2040 but the DMA SPI addresses seem different. I maybe wrong (I am not an expert in microcontrollers)

advancedlogic:
I don't think the behavior is supposed to be this one, but I managed to make it work, transferring 4 Bytes at the time:

```
func (d *DeviceOf[T]) sendCommand(command uint8) error {
	d.dcPin.Low()
	d.csPin.Low()
	err := d.bus.Tx([]byte{command}, nil)
	d.csPin.High()
	return err
}

func (d *DeviceOf[T]) sendData(data []byte) error {
	d.dcPin.High()
	d.csPin.Low()
	for i := 0; i < len(data); i += 4 {
		end := i + 4
		if end > len(data) {
			end = len(data)
		}
		if err := d.bus.Tx(data[i:end], nil); err != nil {
			return err
		}
	}
	d.csPin.High()
	return nil
}
```

hb9cwp:
@advancedlogic Which SPI port are you using, e.g. SPI0 or SPI1? 

[...]

PR #4903, opened 2025-05-22
Fix DMA to SPI transmits on RP2350

hb9cwp:
DMA DREQ line numbers for "flow control" between SPI bus and DMA channels on RP2350 differ from RP2040. Tested with st7789 driver for Pico-1.14-LCD from Waveshare on Pico 2W. Without this fix func st7789.tx() blocks indefinitely while attempting to use DMA to SPI transfers, see #4690.


src/machine/machine_rp2_2040.go
```diff
@@ -70,6 +70,15 @@ const (
 	numClocks
 )
 
+// Additional definitions that are missing in generated src/device/rp/rp2040.go
+const (
+	// from 2040 Datasheet: 2.5.3.1. System DREQ Table
+	DMA_DREQ_SPI0_TX = 16
+	DMA_DREQ_SPI0_RX = 17
+	DMA_DREQ_SPI1_TX = 18
+	DMA_DREQ_SPI1_RX = 19
```
eliasnaur: Why export these? They're useful, sure, but do they belong in the `machine` package from the perspective of importing packages?

hb9cwp: > Why export these?

Because the generated `src/device/rp/rp2040|rp2350.go` don't provide them, and they are global and specific for these processors as well.

> but do they belong in the machine package 

What alternatives do you propose which would be a better fit?

deadprogram: See https://github.com/tinygo-org/tinygo/blob/release/src/device/sam/atsamd51x-bitfields.go for example of how to handle values that are not part of the files generated from the SVDs.

eliasnaur: > > Why export these?
> 
> Because the generated `src/device/rp/rp2040|rp2350.go` don't provide them, and they are global and specific for these processors as well.
> 
> > but do they belong in the machine package
> 
> What alternatives do you propose which would be a better fit?

For every exported API the question should be: what are the use cases for outside users? If there's no identifiable outside need, it's better to keep the API unexported and deal with isssues such as naming, other platforms etc. when [the need arises](https://github.com/orgs/tinygo-org/discussions/4835).

If you or some other maintainer (@soypat?) decide to export anyway, I suggest package `device/rp`. The `machine` package is for abstractions across a wide range of hardware, and I know of no restriction that say `device/rp` is for machine generated code only.

hb9cwp: Agreed, `device/rp` would be a good place, similar to `device/tkey` or `device/riscv` for example. 

Unfortunately, it is currently excluded by .gitignore. It would be helpful to have also generated *.go and *.s files in the repo that are versioned. This would facilitate searching and automatic cross-referencing, etc.

src/device/rp/rp2040-plus.go
```diff
@@ -0,0 +1,51 @@
+// Hand created file. DO NOT DELETE.
+// Definitions that are missing in src/device/rp/rp2040.go generated from SVDs
+
```
eliasnaur: This comment seems redundant: everything not marked generated is not, and I'm not even sure the definitions are missing per se. Are DREQ constants in scope for SVD?

Also, I believe a consistent file name would use underscores. `plus` is not that informative. Perhaps `rp2350_consts.go` for now?

src/device/rp/rp2350-plus.go
```diff
@@ -0,0 +1,66 @@
+// Hand created file. DO NOT DELETE.
+// Definitions that are missing in src/device/rp/rp2040.go generated from SVDs
+
```
eliasnaur: Ditto.

src/machine/machine_rp2_2040.go
hb9cwp: Moved the additional global constants for DMA to `device/rp/rp2nnn0-plus.go`, see refactoring below.
But I did not touch some other lines in `src/machine/machine_rp2_2nnn.go` that could be refactored as well, such as:
```
const (
	cpuFreq          = 150 * MHz
	_NUMBANK0_GPIOS  = 48
	_NUMBANK0_IRQS   = 6
	rp2350ExtraReg   = 1
	_NUMIRQ          = 51
	notimpl          = "rp2350: not implemented"
	...
```

eliasnaur: > Moved the additional global constants for DMA to `device/rp/rp2nnn0-plus.go`, see refactoring below. But I did not touch some other lines in `src/machine/machine_rp2_2nnn.go` that could be refactored as well, such as:
> 
> ```
> const (
> 	cpuFreq          = 150 * MHz
> 	_NUMBANK0_GPIOS  = 48
> 	_NUMBANK0_IRQS   = 6
> 	rp2350ExtraReg   = 1
> 	_NUMIRQ          = 51
> 	notimpl          = "rp2350: not implemented"
> 	...
> ```

These constants are internal implementation details. Why should they be exported?

src/device/rp/rp2040-plus.go
hb9cwp: For naming and comments, I tried to follow the example provided by @deadprogram above.
About the definitions, I was unable to spot them in the generated files, and also did a brief search in SVDs.

deadprogram:
Looks like the new files need to have `go fmt` run on them, if you please:

```
Unformatted:
  src/device/rp/rp2040-plus.go
  src/device/rp/rp2350-plus.go
make: *** [GNUmakefile:185: fmt-check] Error 1
```

deadprogram:
Thanks for the format fixes @hb9cwp 

Only suggestion I have left is perhaps name the new files `src/device/rp/rp2040-bitfields.go` to match the `sam` files.

hb9cwp:
@deadprogram Thanks for re-launching the tests. Yesterday, I could not figure out why one failed. Eventually #4898 is affected as well.

About the naming, I was considering `-const` first, then contemplated more generic `-bis`, `-extra`, or `-plus` that stipulate extensions of the generated files. Because later, we might also include additional types/classes and functions/methods, much like the generated files. Unless we add those to additional files with specific suffixes `rp2nnn-xyz.go`?

deadprogram:
@hb9cwp I think `-extra` would be my personal choice of the options, now that you have explained your thinking. Whichever we end up with here, we should consider changing the names of the existing similar files to match.

deadprogram:
Thank you for adding this @hb9cwp and for making all the requested changes. Also, thanks @eliasnaur for your in-depth reviews and feedback!

Now squash/merging.

hb9cwp:
Thanks for the merge. After building LLVM and TinyGo from sources yesterday, I can confirm that it fixed the issue between DMA and SPI on RP2350: `tinygo version 0.38.0-dev-3f4f4e0e linux/amd64 (using go version go1.24.2 and LLVM version 19.1.2)`.
Although DMA transfers to the LCD now work without further changes, I might propose another PR for the ST7789 display driver. While debugging, I noticed a few timing parameters which appear a bit "aggressive", notably in the inititialization of the ST7789 and LCD where it does not really affect performance, when compared to the ST7789WV datasheet and its implementation in CircuitPython.

PR #4907, opened 2025-05-23
machine/samd21: implement watchdog

mateusznowakdev:
PR #3855 added watchdog implementation for some chip families, but not for SAMD21. This PR adds the missing support.

Note that SAMD21 does not have a default watchdog clock source, and none of the existing ones was a good choice:

- GCLK0 is set to 48 MHz (DFLL48M)
- GCLK1 and GCLK2 are set to 32 kHz (OSC32K)
- GCLK3 is set to 8 MHz (OSC8M)

The `machine.Watchdog.Configure()` function will enable the last clock generator, GCLK8, set its source to OSCULP32K, and set prescaler to 32 to get nice 1.024kHz clock. This is similar to how the clock is configured for I2S peripheral -- it is configured on demand.

The remaining parts of the Configure method are the same as in SAMD51.

### DIFF
--- a/src/machine/machine_atsamd51.go
+++ b/src/machine/machine_atsamd51.go
@@ -2357,6 +2357,5 @@ func (wd *watchdogImpl) Start() error {
 
 // Update the watchdog, indicating that `source` is healthy.
 func (wd *watchdogImpl) Update() {
-	// 0xA5 = magic value (see datasheet)
-	sam.WDT.CLEAR.Set(0xA5)
+	sam.WDT.CLEAR.Set(sam.WDT_CLEAR_CLEAR_KEY)
 }
--- a/src/machine/machine_atsamd21.go
+++ b/src/machine/machine_atsamd21.go
@@ -2027,3 +2027,62 @@ func checkFlashError() error {
 
 	return nil
 }
+
+// Watchdog provides access to the hardware watchdog available
+// in the SAMD21.
+var Watchdog = &watchdogImpl{}
+
+const (
+	// WatchdogMaxTimeout in milliseconds (16s)
+	WatchdogMaxTimeout = (16384 * 1000) / 1024
+)
+
+type watchdogImpl struct{}
+
+// Configure the watchdog.
+//
+// This method should not be called after the watchdog is started and on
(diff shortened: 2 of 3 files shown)


src/machine/machine_atsamd51.go
```diff
@@ -2357,6 +2357,5 @@ func (wd *watchdogImpl) Start() error {
 
 // Update the watchdog, indicating that `source` is healthy.
 func (wd *watchdogImpl) Update() {
-	// 0xA5 = magic value (see datasheet)
-	sam.WDT.CLEAR.Set(0xA5)
+	sam.WDT.CLEAR.Set(sam.WDT_CLEAR_CLEAR_KEY)
```
deadprogram, line 2360: Thanks for replacing the magic number with a defined value here! :broom:

deadprogram:
> The remaining parts of the Configure method are the same as in SAMD51.

Perhaps it would be good to put the shared parts into a single place, a new file such as `machine_atsamdx_watchdog.go`

Then just init the clock (as needed) in the individual files for samd21/samd51 with a new function such as `initWatchdogClock()`.

What do you think?

mateusznowakdev:
I've moved the common watchdog code to `machine_atsam_watchdog.go`, to match naming convention for other platforms (nrf, rp2). SAMD21 works. SAMD51 compiles.

Unfortunately, the `Start` method can't be reused because it accesses different hardware registers.

Not sure if it makes code easier to read :grinning: and there is a lot of repeated code between `machine_atsamd21.go` and `machine_atsamd51.go` anyway. We can merge this PR in its current state, and maybe someone will be interested in refactoring out the remaining common code in the future.

deadprogram:
> Not sure if it makes code easier to read

I think you are correct. It was probably better before. Maybe you should remove the second commit?

mateusznowakdev:
Sure thing, done

deadprogram:
Thanks for working on this @mateusznowakdev now merging.

mateusznowakdev:
Thank you for merging these patches so far, it's nice to have these upstreamed instead of patching SDK every single time :grinning:

deadprogram:
Thank YOU for working on it! :smile_cat:

Issue #4819, opened 2025-03-19
TinyGo Panics on Generic Type Aliases


Yeaseen:
TinyGo crashes when compiling a **generic type alias**, even when using `GODEBUG=gotypesalias=1`. Though I am not sure whether TinyGo supports generic type aliases correctly. 

```bash
$ cat test.go
package main
import "fmt"
// Define a generic type
type Wrapper[T any] struct {
	Value T
}
// Alias declaration with type parameter
type Alias[T any] = Wrapper[T]

func main() {
	// Using the alias
	var a Alias[int] = Alias[int]{Value: 42}
	fmt.Println("a:", a.Value)
}
```

```
GODEBUG=gotypesalias=1 tinygo run test.go
panic: unknown type: main.Alias[int]

goroutine 200 [running]:
github.com/tinygo-org/tinygo/compiler.(*compilerContext).makeLLVMType(0xc003bb2000?, {0x7020428?, 0xc00287d340?})
        /__w/tinygo/tinygo/compiler/compiler.go:468 +0x539
```

gandarez:
It produces different errors when setting target.

Without target set
```
GODEBUG=gotypesalias=1 tinygo run main.go

panic: unknown type: sync.Mutex

goroutine 54 [running]:
github.com/tinygo-org/tinygo/compiler.(*compilerContext).makeLLVMType(0x14001ca8000?, {0x10aa97468?, 0x1400587a800?})
        /Users/runner/work/tinygo/tinygo/compiler/compiler.go:468 +0x578
```

With pico2-w target set
```
GODEBUG=gotypesalias=1 tinygo run -target=pico2-w main.go

panic: unknown type: any

goroutine 116 [running]:
github.com/tinygo-org/tinygo/compiler.(*compilerContext).makeLLVMType(0x1400af8a280?, {0x10b4ab468?, 0x140000221c0?})
        /Users/runner/work/tinygo/tinygo/compiler/compiler.go:468 +0x578
```

Even though changing it from `any` to `int` return the same result `panic: unknown type: any`.

eliasnaur:
Probable dupe: https://github.com/tinygo-org/tinygo/issues/4416

ofauchon:
Hello. 

Same behaviour with ArchLinux (and llvm19 packages)

```
$ pacman -Q llvm 
llvm 19.1.7-1

$ cd tinygo
$ git switch dev
$ git pull --recurse-submodules
$ rm -fr llvm-project
[...]

PR #4910, opened 2025-05-24
flake.*: upgrade to nixpkgs 25.05

eliasnaur:


aykevl:
Good idea, it would probably also be a good idea to update the LLVM version at the same time (from 18 to 20). With a bit of luck it'll just work when you replace all occurences of `18` with `20` in flake.nix.

eliasnaur:
Done.

eliasnaur:
I also added a workaround for #4819.

deadprogram:
Thank you for the update @eliasnaur and to @aykevl for review. Now merging.

flake.nix
```diff
@@ -67,8 +67,11 @@
             # has `md5sum`).
             export MD5SUM=md5sum
 
+            # Work around #4819, missing support for generic type aliases.
+            export GODEBUG=gotypesalias=0
```
aykevl, line 71: Why is this needed? Isn't this still disabled by default?

eliasnaur: The upgrade bumped Go to 1.24 which has `gotypesalias=1` as default.

aykevl: Huh, weird that I never ran into it. I'm using Go 1.24.

PR #4911, opened 2025-05-24
compiler: add support for GODEBUG=gotypesalias=1

aykevl:
Since we now require Go 1.22, this became a lot simpler (since we can now refer to `types.Alias` and `types.Unalias`).


eliasnaur (approved):
Can't say I'm an expert on the compiler, but the change looks good to me and works for my programs.

Thanks for fixing this!

deadprogram:
Thanks for the fix @aykevl and to @eliasnaur for testing. Now merging.

ofauchon:
Thanks for the fix !

PR #4919, opened 2025-06-10
build: go back to using MinoruSekine/setup-scoop for Windows CI builds

deadprogram:
This PR goes back to using the `MinoruSekine/setup-scoop` Github Action for Windows CI builds. We were using a very old fork due to reasons which I think no longer apply.


deadprogram:
Merging to keep us up to date...

PR #4922, opened 2025-06-11
riscv32: use `gdb` binary as a fallback

aykevl:
At least on my system (Fedora 42) the standard gdb binary is capable of debugging riscv-qemu binaries (running inside qemu-system-riscv32).


deadprogram:
Thanks for the fallback @aykevl now merging.

PR #4925, opened 2025-06-12
Multi-core scheduler for the RP2040

aykevl:
This PR depends on #4851. It needs to be rebased after that PR gets merged.

I've been testing it a bit with my [badge code](https://github.com/aykevl/things/tree/master/badge) and it seems to be working well so far! It's unfortunately a bit more difficult to fully test than #4851, but I tried running a few tests in ./testdata and they seem to work correctly.

This PR also includes a few fixes for #4851, issues that I later realized and fixed in this PR as separate commits.

### DIFF
--- a/src/runtime/runtime_rp2040.go
+++ b/src/runtime/runtime_rp2040.go
@@ -4,9 +4,16 @@ package runtime
 
 import (
 	"device/arm"
+	"device/rp"
+	"internal/task"
 	"machine"
 	"machine/usb/cdc"
+	"runtime/interrupt"
+	"runtime/volatile"
+	"unsafe"
 )
+
+const numCPU = 2
 
 // machineTicks is provided by package machine.
 func machineTicks() uint64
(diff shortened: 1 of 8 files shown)


aykevl:
This PR does increase binary size for RP2040 chips (1½-2kB), but since they generally have plenty of flash memory I don't think that's a problem. And having multicore support by default seems more important to me.

deadprogram:
> having multicore support by default seems more important to me.

:100: :exclamation:

deadprogram:
Just tried this branch on an rp2040 app with 3 goroutines, one controlling display, another one running web server.

Worked perfectlyl!! Noticeably "snappier" than the single-core version. :heart_eyes:

deadprogram:
I will rebase this as soon as the `dev` build is complete, to avoid any extra cache invalidation/LLVM rebuilds.

deadprogram:
OK rebased.

ysoldak:
Verified on XIAO-RP2040, works!
I can see my test code executes faster on this branch, compared to dev branch, so I assume it runs on two cores.

Test code
```
package main

import (
	"sync"
	"time"
	"math/rand"
)

func main() {

	time.Sleep(2 * time.Second)
	println(time.Now().Format("15:04:05.999999999"), "Starting goroutines")

	numbers := make(chan int, 100)

	wg := &sync.WaitGroup{}

	for i := 0; i < 2; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()
			v := 0
			for j := 0; j < 1000000; j++ {
				v += rand.Intn(100)
			}
			numbers <- v
		}()
	}

	wg.Wait()

	println(<-numbers)
	println(<-numbers)

	println(time.Now().Format("15:04:05.999999999"), "All goroutines finished")

}
```

Output on "dev" branch:
```
$ tinygo flash -target xiao-rp2040 -monitor ./src/examples/threads
Connected to /dev/cu.usbmodem101. Press Ctrl-C to exit.
00:00:02.007123 Starting goroutines
49519772
49530931
00:00:09.348277 All goroutines finished
```

Output on "threading-rp2040-3" branch:
```
$ tinygo flash -target xiao-rp2040 -monitor ./src/examples/threads
Connected to /dev/cu.usbmodem101. Press Ctrl-C to exit.
00:00:02.007233 Starting goroutines
49519204
49519305
00:00:06.583655 All goroutines finished
```

ysoldak:
Sizes of the test code (see above).

On "dev" branch:
```
  38608    1588    3272 |   40196    4860
```

On "threading-rp2040-3" branch:
```
  40636    1580    5344 |   42216    6924
```

deadprogram:
https://github.com/tinygo-org/tinygo/blob/dev/main.go#L1545

Shouldn't `cores` be added to this list?

Side note: this PR also appears to still work as expected if you use `-scheduler tasks` with the previous behavior.

From reading of the code/datasheet, it appears that this will save power usage on the second core.

targets/rp2040.json
```diff
@@ -12,6 +13,9 @@
     "extra-files": [
         "src/device/rp/rp2040.s"
     ],
+    "ldflags": [
+        "--defsym=__num_stacks=2"
```
deadprogram, line 17: Only possible drawback here is using extra memory if someone wants `-scheduler=tasks` instead of the multicore. But we can probably just document the extra flags that are needed to deal with that for the people who are concerned about power use.

aykevl: Yeah, that's true. There are ways to fix this, if needed. I can do that in a followup PR.

aykevl:
@deadprogram see https://github.com/tinygo-org/tinygo/pull/4927 for the outdated help text

src/runtime/runtime_rp2040.go
```diff
@@ -43,6 +50,284 @@ func sleepTicks(d timeUnit) {
 	}
 }
 
+// Currently sleeping core, or 0xff.
```
eliasnaur, line 53:
```sh
$ diff src/runtime/runtime_rp2*
1c1
< //go:build rp2040
---
> //go:build rp2350
31a32,35
>       if d <= 0 {
>               return
>       }
>
```

Can the two files be merged, unlocking multicore on rp2350 for (approximately) free?

src/runtime/runtime_rp2040.go
```diff
@@ -43,6 +50,284 @@ func sleepTicks(d timeUnit) {
...
+}
+
+func (l *spinLock) Lock() {
+	// Wait for the lock to be available.
+	spinlock := l.spinlock()
+	for spinlock.Get() == 0 {
+		// TODO: use wfe and send an event when unlocking so the CPU can go to
+		// sleep while waiting for the lock.
+		// Unfortunately when doing that, time.Sleep() seems to hang somewhere.
+		// This needs some debugging to figure out.
```
eliasnaur, line 317: This could hide some other subtle error that affects not just `time.Sleep`. It seems risky to enable the multicore scheduler by default in v0.38.0 with so little bake time.

deadprogram: "the voice of reason"? :smiley_cat:

This should build with multicore I think:
 
```
tinygo flash -target=pico -scheduler=cores -ldflags="--defsym=__num_stacks=2" path/to/code
```

Perhaps we should consider it experimental for this release?

deadprogram: @eliasnaur from reading the comment carefully, I interpret it to mean that the implementation in this PR could be improved by using wfe, but it does not yet due to the `time.Sleep` issue mentioned.

That does not change the point you are making overall. :+1:

deadprogram: Another implication of changing the default to multicore, is that default power use will be higher with the second core active. :zap:

src/runtime/runtime_rp2040.go
aykevl: Probably, but I don't have a RP2350 to test. I think there are small differences that need to be accounted for.
I plan on getting a RP2350 soon!

src/runtime/runtime_rp2040.go
aykevl: I don't think power consumption will be affected much: when one of the cores has nothing to do it will just go to sleep. But I didn't measure this. (Should be easy to do by compiling with -scheduler=tasks and -scheduler=cores and measuring the difference).

aykevl: So, yeah, I did not investigate that problem with wfe/sev in the spinlock. I would need to attach a debugger to see where it gets stuck.

We could disable multicore support for now, and leave it for people who want to test it. It's possible there's a bug I didn't discover yet. That comment is just there since I had to back out wfe/sev at the last minute and wanted to make sure the next person knows there's an opportunity for optimization there.

deadprogram: > That comment is just there since I had to back out wfe/sev at the last minute and wanted to make sure the next person knows there's an opportunity for optimization there.

That was what I was trying to say. There is not a known bug here, this is a chance to optimize further.

deadprogram: > I don't think power consumption will be affected much: when one of the cores has nothing to do it will just go to sleep.

So it might use some extra power initially, but then would go down to minimal use on that core?

src/runtime/runtime_rp2040.go
deadprogram: I will give you one next week!

src/runtime/runtime_rp2040.go
aykevl: So the way it works (at least how I understand it), is that both cores are always running. The second core is just asleep on reset. With multicore support, it is pulled out of this sleep state and starts the scheduler - which in many cases will put it right back to sleep waiting for a goroutine to run. So yeah it spends a little time configuring stuff and then goes back to sleep.
...but again, I didn't measure anything so I'm only speculating how it *should* work.

deadprogram: > So yeah it spends a little time configuring stuff and then goes back to sleep.

That is what all of the docs and references I could find are claiming about power use. So it would appear to be fine as it is from that point of view.

deadprogram: > We could disable multicore support for now, and leave it for people who want to test it. It's possible there's a bug I didn't discover yet.

As much as this feature excites me, I think @eliasnaur is probably correct that we should not make it the default yet.

eliasnaur: @aykevl has a better idea of the robustness of the multi-core support than I; I merely pointed out the timing.

I'm also very excited about this code, and I would be comfortable making multi-core default, if the release notes spelled out instructions for switching back to the single-core scheduler. It's just a `--scheduler=...` argument, right? Assuming `__num_stacks=2` is benign.

deadprogram: I had already confirmed that `-scheduler=tasks` does what you would expect and runs on single core.

I assume that the extra stack space is wasted in that case, so I also tested this command (note I missed the `extldflags` in my original comment):

```
tinygo flash -target nano-rp2040 -ldflags="--extldflags '--defsym=__num_stacks=1'" -stack-size 8kb -scheduler tasks /path/to/my/code
```

This also worked correctly, and according to `-size full` did save some bytes.

Based on this, I think we can merge this PR as it is, and just add the docs to fallback to single core.

deadprogram:
Thank you very much for all your hard work on this amazing feature @aykevl and to everyone who helped review @ysoldak @eliasnaur others too.

Now merging!

hb9cwp:
Having no RP2040 on hand, I monkey-patched `src/runtime/runtime_rp2350.go` as suggested by @eliasnaur and @aykevl above and built both single and dual core binaries for Pico 2W of the test code from @ysoldak above using tinygo 0.38.0 release.
I am happy to report that a) the dual core version runs on RP2350 as well, and b) it takes only about 0.6 s compared to 1.0 s of the single core binary :-)

```
$ tinygo build -target pico2-w -o testSingle.uf2 -scheduler=tasks -stack-size=8kb -size short test.go
  79308    2932    3292 |   82240    6224
...
Hello from SingleCore, CPU Freq is  150000000
CPU Freq is  150000000
00:00:15.032134 Starting goroutines
49519772
49530931
00:00:16.07982 All goroutines finished

$ tinygo build -target pico2-w -o testDual.uf2 -scheduler=cores -ldflags="-extldflags '--defsym=__num_stacks=2'" -stack-size=8kb -size short test.go
  81168    2948    5344 |   84116    8292
Hello from MultiCore, CPU Freq is  150000000
00:00:15.000231 Starting goroutines
49519772
49519085
00:00:15.598207 All goroutines finished
```
Thanks to all for the great work that makes this possible!

Issue #4519, opened 2024-10-12
tinygo flash may not work on Sequoia (macOS 15)


Senoue:
Hi. 
I wonder if tinygo flash can't be used on Sequoia (macOS 15).
In my case,

```
failed to flash /var/folders/m9/xxxx/T/xxx/main.uf2: unable to locate any volume: [RPI-RP2]
```

Is everyone okay?

Senoue:
sorry. It was macOS 15 instead of 18

sago35:
The error is being output from here, and we can predict that the cause lies in `findFATMounts()`. Since I don't have a macOS system, I need someone's help.

https://github.com/tinygo-org/tinygo/blob/v0.33.0/main.go#L994-L1018

aykevl:
I have MacOS (though I don't use it much) but am not yet on Sequoia. Kinda hesitant to upgrade, am afraid it will break stuff (last upgrade didn't exactly go very smoothly).

Senoue:
Thank you, 
when I checked the mount point on Sequoia, it was mounted as NO NAME in /Volumes, 
so I added NO NAME to targets/rp2040.json

```
"msd-volume-name": ["RPI-RP2", "NO NAME"],
```

I found out that tinygo flash can be done!

deadprogram:
Is this true for all Sequoia users?

Senoue:
I had the same problem with two Macs and a friend's environment.

zwindler:
I have the issue as well

I installed tinygo as a brew package, and I'm not sure where I can find the "targets/rp2040.json" file with this mean of installation

Can you give a bit more details on how you solved it @Senoue?

I'll try to fix it myself with the information you (kindly) gave in the meantime and report if I find on my own :)

zwindler:
Ok, found it. For future readers

When you install with homebrew, the config file is here

```
/opt/homebrew/Cellar/tinygo/0.35.0/targets
```

(Replace 0.35.0 with your version)

And also don't forget to read "Additional Requirements for Microcontrollers" from https://tinygo.org/getting-started/install/macos, or else the rpi2040 will unmount (with an error message)

I still have the issue though, but I'm progressing xD

```
failed to flash /var/folders/dr/xxx/T/tinygo12345/main.uf2: unable to locate any volume: [RPI-RP2,NO NAME]
```

sago35:
> /opt/homebrew/Cellar/tinygo/0.35.0/targets

[...]

PR #4928, opened 2025-06-13
Use diskutil on macOS to extract volume name and path for FAT mounts

mi-bear:
### Summary

On macOS, this change improves the detection of FAT volumes by parsing `diskutil info` output to extract the volume name and path. This ensures accurate identification of mounted FAT-based filesystems under `/Volumes`.

Fixes https://github.com/tinygo-org/tinygo/issues/4519

### Details

- Replaces hardcoded assumptions about the filesystem layout with dynamic parsing of `diskutil info`
- Extracts `VolumeName` and `FileSystemPersonality` for each mount point
- Adds a check for `scanner.Err()` to handle unexpected parsing issues gracefully
- Sorts results by volume path for consistent ordering

### Platform

- This change only affects the `darwin` (macOS) platform
- No impact on other platforms or existing logic for `linux`/`windows`

### DIFF
--- a/main.go
+++ b/main.go
@@ -1013,15 +1013,43 @@ func findFATMounts(options *compileopts.Options) ([]mountPoint, error) {
 			return nil, fmt.Errorf("could not list mount points: %w", err)
 		}
 		for _, elem := range list {
-			// TODO: find a way to check for the filesystem type.
-			// (Only return FAT filesystems).
-			points = append(points, mountPoint{
-				name: elem.Name(),
-				path: filepath.Join("/Volumes", elem.Name()),
-			})
+			volumePath := filepath.Join("/Volumes", elem.Name())
+			if _, err := os.Stat(volumePath); err != nil {
+				continue
+			}
+
+			cmd := exec.Command("diskutil", "info", volumePath)
+			var out bytes.Buffer
+			cmd.Stdout = &out
+			if err := cmd.Run(); err != nil {
+				continue // skip if diskutil failed
+			}
+
+			var volName, fsType string
+			scanner := bufio.NewScanner(&out)
+			for scanner.Scan() {
+				line := scanner.Text()
+				if strings.HasPrefix(line, "  Volume Name:") {
+					volName = strings.TrimSpace(strings.TrimPrefix(line, "  Volume Name:"))
+				}
+				if strings.HasPrefix(line, "  File System Personality:") {
+					fsType = strings.TrimSpace(strings.TrimPrefix(line, "  File System Personality:"))
+				}
+			}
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -1013,15 +1013,43 @@ func findFATMounts(options *compileopts.Options) ([]mountPoint, error) {
...
+			cmd.Stdout = &out
+			if err := cmd.Run(); err != nil {
+				continue // skip if diskutil failed
+			}
+
+			var volName, fsType string
+			scanner := bufio.NewScanner(&out)
+			for scanner.Scan() {
+				line := scanner.Text()
+				if strings.HasPrefix(line, "  Volume Name:") {
```
b0ch3nski: To consider: you could simply build the `diskutil` output into a map, like:
```go
if colonIndex := strings.Index(line, ":"); colonIndex != -1 {
	key := strings.TrimSpace(line[:colonIndex])
	value := strings.TrimSpace(line[colonIndex+1:])
	diskInfo[key] = value
}
```
And then after the scanner loop, check and read values quickly, like:
```go
volName, ok := diskInfo["Volume Name"]
if !ok {
	continue
}
```
_Warning: untested_

mi-bear: Thanks a lot for the thoughtful suggestion — turning the `diskutil` output into a map definitely made the logic clearer and more maintainable. That approach really helped me rethink how to handle this kind of key-value parsing.

I ended up using `strings.SplitN` instead of `Index` to make the split a bit safer in case the value ever contains a colon, but the idea came directly from your comment. Learned something new from this — appreciate it!

main.go
```diff
@@ -1013,15 +1013,50 @@ func findFATMounts(options *compileopts.Options) ([]mountPoint, error) {
...
+			}
+			if err := scanner.Err(); err != nil {
+				continue
+			}
+
+			volName, okv := diskInfo["Volume Name"]
+			fsType, okf := diskInfo["File System Personality"]
+			if !(okv && okf) {
+				continue
+			}
```
sago35: I think this is better.

```suggestion
			if !okv || !okf {
				continue
			}
```

mi-bear: Agreed, the simpler conditional is more readable. I've made the change, thanks for the suggestion!

sago35 (approved):
Looks good to me.

main.go
deadprogram: Demorgan's Law says those two expressions are equivalent. :heavy_equals_sign:

deadprogram:
Thank you for the improvement @mi-bear and to @sago35 & @b0ch3nski for review. Now squash/merging.

PR #4930, opened 2025-06-20
machine: make sure DMA buffers do not escape unnecessarily

aykevl:
Writing the pointer of a buffer to memory-mapped I/O will normally cause it to escape, which forces the compiler to heap-allocate the buffer. But we do know how long the value stays alive, so we can tell the compiler to keep it alive exactly until it is not needed anymore - and tell it to not treat the pointer-to-uintptr cast as escaping.

This PR tries to solve the same problem as https://github.com/tinygo-org/tinygo/pull/4889 but in a different way. In fact, see https://github.com/tinygo-org/tinygo/pull/4889#issuecomment-2904700154 where I first came up with this idea. Basically, with this PR, the allocation inside `legacy.ReadRegister` (but probably not `legacy.WriteRegister`!) should be fixed. See: https://github.com/tinygo-org/drivers/pull/766.

It's perhaps not the cleanest solution, but it seems to work in my tests.

@ysoldak can you take a look?

@dgryski @eliasnaur you may also be interested in this PR.

### DIFF
--- a/src/machine/machine.go
+++ b/src/machine/machine.go
@@ -1,6 +1,9 @@
 package machine
 
-import "errors"
+import (
+	"errors"
+	"unsafe"
+)
 
 var (
 	ErrTimeoutRNG         = errors.New("machine: RNG Timeout")
@@ -62,3 +65,30 @@ func (p Pin) Low() {
 type ADC struct {
 	Pin Pin
 }
+
+// Convert the pointer to a uintptr, to be used for memory I/O (DMA for
+// example). It also means the pointer is "gone" as far as the compiler is
+// concerned, and a GC cycle might deallocate the object. To prevent this from
+// happening, also call keepAliveNoEscape at a point after the address isn't
+// accessed anymore by the hardware.
+// The only exception is if the pointer is accessed later in a volatile way
+// (volatile read/write), which also forces the value to stay alive until that
+// point.
+//
+// This function is treated specially by the compiler to mark the 'ptr'
+// parameter as not escaping.
+//
+// TODO: this function should eventually be replaced with the proposed ptrtoaddr
+// instruction in LLVM. See:
+// https://discourse.llvm.org/t/clarifiying-the-semantics-of-ptrtoint/83987/10
+// https://github.com/llvm/llvm-project/pull/139357
+func unsafeNoEscape(ptr unsafe.Pointer) uintptr {
+	return uintptr(ptr)
+}
+
+// Make sure the given pointer stays alive until this point. This is similar to
+// runtime.KeepAlive, with the difference that it won't let the pointer escape.
+// This is typically used together with unsafeNoEscape.
+//
+// This is a compiler intrinsic.
+func keepAliveNoEscape(ptr unsafe.Pointer)
--- a/src/machine/machine_nrf528xx.go
+++ b/src/machine/machine_nrf528xx.go
@@ -49,7 +49,7 @@ func (i2c *I2C) Tx(addr uint16, w, r []byte) (err error) {
 
 	// Configure for a single shot to perform both write and read (as applicable)
 	if len(w) != 0 {
-		i2c.Bus.TXD.PTR.Set(uint32(uintptr(unsafe.Pointer(&w[0]))))
+		i2c.Bus.TXD.PTR.Set(uint32(unsafeNoEscape(unsafe.Pointer(unsafe.SliceData(w)))))
 		i2c.Bus.TXD.MAXCNT.Set(uint32(len(w)))
 
 		// If no read, immediately signal stop after TX
@@ -58,7 +58,7 @@ func (i2c *I2C) Tx(addr uint16, w, r []byte) (err error) {
 		}
 	}
 	if len(r) != 0 {
-		i2c.Bus.RXD.PTR.Set(uint32(uintptr(unsafe.Pointer(&r[0]))))
+		i2c.Bus.RXD.PTR.Set(uint32(unsafeNoEscape(unsafe.Pointer(unsafe.SliceData(r)))))
 		i2c.Bus.RXD.MAXCNT.Set(uint32(len(r)))
 
 		// Auto-start Rx after Tx and Stop after Rx
@@ -89,6 +89,11 @@ func (i2c *I2C) Tx(addr uint16, w, r []byte) (err error) {
 		}
 	}
 
+	// Make sure the w and r buffers stay alive until this point, so they won't
+	// be garbage collected while the buffers are used by the hardware.
+	keepAliveNoEscape(unsafe.Pointer(unsafe.SliceData(w)))
+	keepAliveNoEscape(unsafe.Pointer(unsafe.SliceData(r)))
+
 	return
 }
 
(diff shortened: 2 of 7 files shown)


src/machine/machine.go
```diff
@@ -62,3 +65,30 @@ func (p Pin) Low() {
...
+func unsafeNoEscape(ptr unsafe.Pointer) uintptr {
+	return uintptr(ptr)
+}
+
+// Make sure the given pointer stays alive until this point. This is similar to
+// runtime.KeepAlive, with the difference that it won't let the pointer escape.
+// This is typically used together with unsafeNoEscape.
+//
+// This is a compiler intrinsic.
+func keepAliveNoEscape(ptr unsafe.Pointer)
```
eliasnaur, line 94: This is, as you say, hacky and won't help external drivers that don't have access to the unexported `unsafeNoEscape` and `keepAliveNoEscape`.

Can't `runtime.KeepAlive` be made to not move the allocation to the heap in TinyGo? If so, that leaves `unsafeNoEscape`. I suggest the more drastical solution of accepting #4889 that relaxes `uintptr(unsafe.Pointer(&x))` to not escape. My reasoning is that for something like DMA, you need something like `runtime.KeepAlive` to keep the pointer alive anyway (it's not enough for the pointer to be kept alive at the point of conversion to `uintptr`).

That way, all code can use the same tricks to avoid allocations.

aykevl: > This is, as you say, hacky and won't help external drivers that don't have access to the unexported `unsafeNoEscape` and `keepAliveNoEscape`.

I intentionally kept them unexported. Drivers (like the tinygo.org/x/drivers/* packages) won't need it as long as they don't access these hardware registers directly - which is the vast majority. If there are drivers which really need these functions, we can export these functions.

> Can't `runtime.KeepAlive` be made to not move the allocation to the heap in TinyGo?

Yes, but that won't help the machine package due to a circular dependency (machine -> runtime -> machine).

> I suggest the more drastical solution of accepting #4889 that relaxes `uintptr(unsafe.Pointer(&x))` to not escape.

The only way to implement that at the moment is to let the compiler do something exactly like `unsafeNoEscape`: insert a call at the point of the cast. I much prefer to keep this explicit and keep the default (escaping) behavior for these casts. In the future there might be ptrtoaddr but not today.

> My reasoning is that for something like DMA, you need something like `runtime.KeepAlive` to keep the pointer alive anyway (it's not enough for the pointer to be kept alive at the point of conversion to `uintptr`).

Yeah that's the goal of this PR.

ysoldak:
Any idea why it fails on Windows?

ysoldak:
Do we have a test that proves this helps?
Asking because with tricks like https://github.com/tinygo-org/drivers/pull/766 I was able to eliminate all I2C escapes, at least in my project.

eliasnaur:
> Do we have a test that proves this helps? Asking because with tricks like [tinygo-org/drivers#766](https://github.com/tinygo-org/drivers/pull/766) I was able to eliminate all I2C escapes, at least in my project.

I agree with this sentiment. Why avoid escaping buffers that could surely be scratch space in some larger, already allocated, structure?

eliasnaur:
What about drivers written on top of an interface, say [drivers.I2C](https://github.com/tinygo-org/drivers/blob/c4ff8242a7726cb0d281d7c583d124e8d52bca00/i2c.go#L5)? AFAIK, Unless TinyGo can see through the interface indirection, any buffers passed to `I2C.Tx` will escape.

aykevl:
Rebased, and added tests.

> What about drivers written on top of an interface, say [drivers.I2C](https://github.com/tinygo-org/drivers/blob/c4ff8242a7726cb0d281d7c583d124e8d52bca00/i2c.go#L5)? AFAIK, Unless TinyGo can see through the interface indirection, any buffers passed to `I2C.Tx` will escape.

Yeah, this is only a partial solution. Callers that use the `drivers.I2C` interface should probably still allocate a static buffer to be sure (in the driver object, for example). But I do think it fixes a few real world cases.

src/machine/machine_nrf528xx.go
```diff
@@ -89,6 +89,15 @@ func (i2c *I2C) Tx(addr uint16, w, r []byte) (err error) {
 		}
 	}
 
+	// Make sure the w and r buffers stay alive until this point, so they won't
+	// be garbage collected while the buffers are used by the hardware.
+	if len(w) > 0 {
+		keepAliveNoEscape(unsafe.Pointer(&w[0]))
```
eliasnaur: Nit: why not use `unsafe.SliceData` and drop the `len(w) > 0` check? Here and elsewhere.

aykevl: Because it's an old PR that I rebased.
I've updated the PR to use `unsafe.SliceData` instead where possible.

ysoldak:
I have a reproducer that confirms this solves unnecessary heap escapes, at least for nrf52 chip.

In the drivers repo, on 0.33.0 version, on release branch of tinygo, i.e. 0.39.0 (same results on dev branch).

`$ tinygo build -target nano-33-ble -o build/test.hex -size=short -opt=z -print-allocs=drivers ./examples/lsm6dsox`
```
~/tinygo-drivers/lsm6dsox/lsm6dsox.go:43:16: object allocated on the heap: escapes at line 43
~/tinygo-drivers/internal/legacy/i2clegacy.go:10:13: object allocated on the heap: escapes at line 13
~/tinygo-drivers/internal/legacy/i2clegacy.go:6:36: object allocated on the heap: escapes at line 6
  62376    1580    6472 |   63956    8052
```

Interestingly, rc2040 targets do not have this problem:
`$ tinygo build -target nano-rp2040 -o build/test.hex -size=short -opt=z -print-allocs=drivers  ./examples/lsm6dsox`
```
~/tinygo-drivers/lsm6dsox/lsm6dsox.go:43:16: object allocated on the heap: escapes at line 43
  65840    1636    5240 |   67476    6876
```

After switching to to `machine-noescape-dma` branch and doing `go install`, I now have this:
`$ tinygo build -target nano-33-ble -o build/test.hex -size=short -opt=z -print-allocs=drivers ./examples/lsm6dsox`
```
~/tinygo-drivers/lsm6dsox/lsm6dsox.go:43:16: object allocated on the heap: escapes at line 43
  62392    1580    6472 |   63972    8052
```

This change seems to be fixing escapes for short-living slices, I'm happy.
@deadprogram @soypat makes our live easier in "drivers".
We may, as @soypat says, want to implement some CI to check for such escapes even if this change to be merged, to monitor regressions.

src/machine/machine.go
ysoldak: > In the future there might be ptrtoaddr but not today.

We may expect this function added in LLVM 22, seems like.
https://llvm.org/docs/ReleaseNotes.html#changes-to-the-llvm-ir

deadprogram:
I personally am in favor of this PR, mainly due to being able to at least catch some of these unintentional escapes now.

Also agree that avoiding them in the first place is the ideal, but this is at least another defensive step.

@soypat any opinions on this?

soypat:
Seems like a neat compiler intrinsic to help out in these cases. Don't think it helps out too much in drivers package, but nice to have

deadprogram:
Rebased against latest `dev` to make sure tests all pass.

deadprogram:
All passing so now merging. Thank you everyone!

PR #4934, opened 2025-06-23
Allow implementing different WIT worlds in wasip2 target

### LINKED ISSUE #4843: Allow compiling Wasm components with worlds besides `wasi:cli/command`
AFAICT, TinyGo already supports the Wasm component model, but only if you're compiling a component for the `wasi:cli/command` world. That's [hardwired](https://github.com/tinygo-org/tinygo/blob/v0.37.0/targets/wasip2.json#L31-L32) into the `wasip2` target.

What about compiling components with a different WIT interface? Is that something other people already know how to do, but it's just not well documented? Or are there changes that have to be made? I'm happy to help develop this, but would appreciate some pointers from anybody with context.

tomhoule:
As mentioned in [this comment](https://github.com/tinygo-org/tinygo/issues/4843#issuecomment-2988622249), we are also interested in Wasm components with custom worlds (for [Grafbase extensions](https://grafbase.com/docs/gateway/extensions)).

I have succesfully compiled and used examples based on the changes in this branch. They are mostly from @daedric's patch in the issue.

So this PR is meant as a starting point for discussions, a working branch for anyone else looking at trying this, and to discuss how best to test it (I've made attempts, but they added up to a lot of code, and I wasn't too happy with them).

I understand it's work to review and merge something like this, and even more to provide guidance on a proper test setup, so while feedback would be much appreciated, I understand if it's not a priority.

fixes #4843

### DIFF
--- a/builder/build.go
+++ b/builder/build.go
@@ -914,6 +914,11 @@ func Build(pkgName, outpath, tmpdir string, config *compileopts.Config) (BuildRe
 			}
 
 			// Run wasm-tools for component-model binaries
+			wasiPackage := strings.ReplaceAll(config.Target.WASIPackage, "{root}", goenv.Get("TINYGOROOT"))
+			if config.Options.WASIPackage != "" {
+				wasiPackage = config.Options.WASIPackage
+			}
+
 			witPackage := strings.ReplaceAll(config.Target.WITPackage, "{root}", goenv.Get("TINYGOROOT"))
 			if config.Options.WITPackage != "" {
 				witPackage = config.Options.WITPackage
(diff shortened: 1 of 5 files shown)


daedric:
LGTM (but it does not carry any weight :)), I'm happy that my preliminary work has been useful :)

tomhoule:
I see the failing tests, will try and get these running locally and figure out the cause.

cataggar:
This isn't needed. I explained in https://github.com/tinygo-org/tinygo/issues/4843#issuecomment-3367784501

PR #4938, opened 2025-06-27
rp2: use the correct channel mask for rp2350 ADC; hold lock during read

eliasnaur:
Found while investigating https://github.com/orgs/tinygo-org/discussions/4936.

Untested.

### DIFF
--- a/src/machine/machine_rp2_adc.go
+++ b/src/machine/machine_rp2_adc.go
@@ -73,14 +73,15 @@ func (c ADCChannel) Configure(config ADCConfig) error {
 func (c ADCChannel) getOnce() uint16 {
 	// Make it safe to sample multiple ADC channels in separate go routines.
 	adcLock.Lock()
-	rp.ADC.CS.ReplaceBits(uint32(c), 0b111, rp.ADC_CS_AINSEL_Pos)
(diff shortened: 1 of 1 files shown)


soypat (commented):
Why are we locking? Is this due to new multicore semantics?

eliasnaur:
I guess so. And because the ADC is shared between the analog pins, it's reasonable to support separate goroutines using it simultaneously.

soypat:
Either we test it or we could do a side-by-side comparison of existing pico-sdk code... or a datasheet walkthrough would work

eliasnaur:
It's tested here: https://github.com/orgs/tinygo-org/discussions/4936#discussioncomment-13603066.

soypat:
I've noted it is tested but exhibits no change to behaviour so I'm still curious as to what this CL solves.

eliasnaur:
The patch does 2 things:

- Move the reading of the ADC inside the lock, on the basis that if the lock is needed, the actual reading the value should be locked as well.
- Use `rp.ADC_CS_AINSEL_Msk` instead of the hard-coded `0b111` mask, which is wrong for rp2350.

soypat:
Alright, LGTM in that case! Thank you for the contribution @eliasnaur !

PR #4942, opened 2025-07-02
flash: add -o flag support to save built binary (Fixes #4937)

### LINKED ISSUE #4937: Add -o flag support to flash command
The `tinygo flash` command currently doesn't support the `-o` flag to save the built binary to a specified output file, while the `tinygo build` command does. This inconsistency makes it inconvenient when users want to both flash a device and save the binary file for later use or distribution.

## Current Behavior
```bash
tinygo build -o firmware.hex -target arduino main.go  # Works - saves to firmware.hex
tinygo flash -o firmware.hex -target arduino main.go  # Doesn't work - error message to stop.
```

## Expected Behavior
```bash
tinygo flash -o firmware.hex -target arduino main.go  # Should flash device AND save to firmware.hex
```

## Proposed Implementation

I believe there are two possible approaches:

### Approach 1: Minimal Change (Immediate Solution)
- Modify `Flash` function to copy the built binary to the specified output path after successful build
- Pass `outpath` as an additional parameter to the Flash function
- Pros: Minimal risk, easy to review, maintains backward compatibility
- Cons: Duplicates output logic between `build` and `flash` commands, inconsistent parameter handling

### Approach 2: Refactor Build Function (Better Long-term Solution)
- Modify `Build` function in `main.go` to use `compileopts.Options` instead of separate `outpath` parameter
- This would unify the output handling logic between `build` and `flash` commands
- Pros: Cleaner architecture, consistent design, eliminates code duplication
- Cons: Larger change, requires more careful testing

**Personal Preference**: I lean toward **Approach 2** (refactoring the Build function) as it creates a more consistent and maintainable codebase. However, I understand that Approach 1 might be preferred for minimizing change scope.

Would appreciate feedback on which approach the maintainers prefer before submitting a pull request.

rdon-key:
This PR adds `-o` flag support to the `tinygo flash` command, allowing users to save the built binary to a specified output file. This brings `flash` in line with the behavior of `tinygo build`, promoting consistency and improving usability.

Fixes #4937

## Changes
- **Added `-o` flag support** to the `flash` command
- **Introduced `validateOutputFormat` function** to check compatibility with the target format
- **Saved output file after build and before flashing**
- **Updated help text** for the `flash` command with usage examples

## Behavior
- When `-o` is specified, the binary is both **saved** and **flashed**
- Output file format must match the target's expected format (e.g., `.uf2`, `.hex`)
- If the format is invalid, the command fails **before building**
- The saved file is identical to what is flashed

## Testing Results
```bash
# Build works as usual
$ tinygo build -o build.uf2 --target waveshare-rp2040-zero .
 258388    1664    5352 |  260052    7016

# Flash with -o saves file and flashes the device
$ tinygo flash -o flash.uf2 --target waveshare-rp2040-zero .
 258388    1664    5352 |  260052    7016

# Format validation works
$ tinygo flash -o flash.hex --target waveshare-rp2040-zero .
output format .hex does not match target format .uf2

# File content is identical between build and flash
$ sha256sum flash.uf2 build.uf2
95cc493412d9937433cd0ba9e8d667be92de19bab18eaba6cc49ada4b8d3a8c5 *flash.uf2
95cc493412d9937433cd0ba9e8d667be92de19bab18eaba6cc49ada4b8d3a8c5 *build.uf2

# Updated Help Output
$ tinygo help flash
Flash the program to a microcontroller. Some common flags are described below.

        -target={name}:
                        Specifies the type of microcontroller that is used. The name of the
                        microcontroller is given on the individual pages for each board type
                        listed under Microcontrollers
                        (https://tinygo.org/docs/reference/microcontrollers/).
                        Examples: "arduino-nano", "d1mini", "xiao".

        -o={filename}:
                        Save the built binary to the specified output file. The file
                        format must match the target's expected format (e.g., .hex,
                        .uf2). Both flashing and saving will be performed.

        -monitor:
                        Start the serial monitor (see below) immediately after
                        flashing. However, some microcontrollers need a split second
                        or two to configure the serial port after flashing, and
                        using the "-monitor" flag can fail because the serial
                        monitor starts too quickly. In that case, use the "tinygo
                        monitor" command explicitly.
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -341,10 +341,6 @@ func dirsToModuleRootAbs(maindir, modroot string) []string {
 
 // validateOutputFormat checks if the output file extension matches the expected format
 func validateOutputFormat(outpath, expectedExt string) error {
-	if outpath == "" {
-		return nil // no output file specified
-	}
-
 	actualExt := filepath.Ext(outpath)
 	if actualExt != expectedExt {
 		return fmt.Errorf("output format %s does not match target format %s", actualExt, expectedExt)
--- a/main.go
+++ b/main.go
@@ -339,8 +339,21 @@ func dirsToModuleRootAbs(maindir, modroot string) []string {
 	return dirs
 }
 
+// validateOutputFormat checks if the output file extension matches the expected format
+func validateOutputFormat(outpath, expectedExt string) error {
+	if outpath == "" {
+		return nil // no output file specified
+	}
+
+	actualExt := filepath.Ext(outpath)
+	if actualExt != expectedExt {
+		return fmt.Errorf("output format %s does not match target format %s", actualExt, expectedExt)
+	}
+	return nil
+}
+
 // Flash builds and flashes the built binary to the given serial port.
-func Flash(pkgName, port string, options *compileopts.Options) error {
+func Flash(pkgName, port, outpath string, options *compileopts.Options) error {
 	config, err := builder.NewConfig(options)
 	if err != nil {
 		return err
@@ -389,13 +402,24 @@ func Flash(pkgName, port string, options *compileopts.Options) error {
 	if !options.Work {
 		defer os.RemoveAll(tmpdir)
 	}
-
+	// Validate output format before building
+	if outpath != "" {
+		if err := validateOutputFormat(outpath, fileExt); err != nil {
+			return err
+		}
+	}
 	// Build the binary.
 	result, err := builder.Build(pkgName, fileExt, tmpdir, config)
 	if err != nil {
 		return err
 	}
 
+	// Save output file if specified (after build, before flashing)
+	if outpath != "" {
+		if err := copyFile(result.Binary, outpath); err != nil {
+			return fmt.Errorf("failed to save output file: %v", err)
+		}
+	}
 	// do we need port reset to put MCU into bootloader mode?
 	if config.Target.PortReset == "true" && flashMethod != "openocd" {
 		port, err := getDefaultPort(port, config.Target.SerialPort)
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -339,8 +339,21 @@ func dirsToModuleRootAbs(maindir, modroot string) []string {
 	return dirs
 }
 
+// validateOutputFormat checks if the output file extension matches the expected format
+func validateOutputFormat(outpath, expectedExt string) error {
+	if outpath == "" {
```
b0ch3nski: Your code checks for this condition already before calling this function, so this seems redundant

rdon-key: Done. That check is indeed redundant since we validate outpath before calling this function. I cleaned this up.

b0ch3nski (approved):

rdon-key:
The macOS CI job (build-macos-14) is being canceled after running for almost 6 hours due to a timeout.
Both Windows and Linux jobs complete successfully, and the tests seem to pass otherwise.
I have tried three times with the same result.

Could a maintainer advise on how to proceed in this situation?

deadprogram:
@rdon-key can you please rebase this branch against the latest `dev` branch? Thanks!

rdon-key:
@deadprogram 
Rebased onto latest dev as requested. Ready for merge.

rdon-key:
@deadprogram @sago35

 Hi! It's been about 2 weeks since I rebased this PR as requested. 

Could you please let me know if there's anything else needed for this PR to move forward? I'm happy to address any additional feedback.

Thanks!

deadprogram:
OK going to now squash/merge this one unless any objections?

PR #4943, opened 2025-07-03
fix: add SPI and I2C to teensy 4.1

emricks:
Adds support for SPI and I2C to the teensy41 board.

SPI3 was removed because the pin definitions are incorrect and a part of the SD card.

Tested SPI1 and SPI2 and all three I2C buses with basic examples.

### DIFF
--- a/src/machine/board_teensy41.go
+++ b/src/machine/board_teensy41.go
@@ -299,45 +299,45 @@ const (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


emricks:
Added SPI3 variables back with correct values to fix tests.

deadprogram:
Thank you very much for working on this @emricks 

I changed the target branch for the PR to `dev` as mentioned here: https://tinygo.org/docs/guides/contributing/#how-to-use-our-github-repository

Now squash/merging.

PR #4945, opened 2025-07-07
machine: make I2C usable in the simulator

aykevl:
This fixes/improves a few issues with I2C support:

  * Validate I2C pins, so only pins that are supported by the hardware can be used (similar to how it's done with PWM).
  * Add address to Tx API (without it, the simulator can't really simulate I2C).
  * Add frequency when configuring. Not currently used, but might be useful in the future and adding it now avoids possibly breaking changes.

This is a breaking change, but since the simulator doesn't support I2C yet that seems fine to me. (It does in my local changes, but those need to be cleaned up before I can push them).

### DIFF
--- a/src/machine/i2c.go
+++ b/src/machine/i2c.go
@@ -37,6 +37,9 @@ var (
 	errI2COverflow           = errors.New("I2C receive buffer overflow")
 	errI2COverread           = errors.New("I2C transmit buffer overflow")
 	errI2CNotImplemented     = errors.New("I2C operation not yet implemented")
+	errI2CNoDevices          = errors.New("i2c: bus has no devices")                         // simulator only
+	errI2CMultipleDevices    = errors.New("i2c: bus has address conflict")                   // simulator only
+	errI2CWrongAddress       = errors.New("i2c: bus has devices but none with this address") // simulator only
 )
 
 // I2CTargetEvent reflects events on the I2C bus
--- a/src/machine/i2c.go
+++ b/src/machine/i2c.go
@@ -26,17 +26,17 @@ const (
 )
 
 var (
-	errI2CWriteTimeout       = errors.New("I2C timeout during write")
-	errI2CReadTimeout        = errors.New("I2C timeout during read")
-	errI2CBusReadyTimeout    = errors.New("I2C timeout on bus ready")
-	errI2CSignalStartTimeout = errors.New("I2C timeout on signal start")
-	errI2CSignalReadTimeout  = errors.New("I2C timeout on signal read")
-	errI2CSignalStopTimeout  = errors.New("I2C timeout on signal stop")
-	errI2CAckExpected        = errors.New("I2C error: expected ACK not NACK")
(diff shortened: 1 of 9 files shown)


deadprogram:
@aykevl please note build errors: https://github.com/tinygo-org/tinygo/actions/runs/16116006879/job/45469916956?pr=4945#step:17:30

aykevl:
Should be fixed now.

src/machine/i2c.go
```diff
@@ -37,6 +37,9 @@ var (
 	errI2COverflow           = errors.New("I2C receive buffer overflow")
 	errI2COverread           = errors.New("I2C transmit buffer overflow")
 	errI2CNotImplemented     = errors.New("I2C operation not yet implemented")
+	errI2CNoDevices          = errors.New("I2C bus has no devices")                         // simulator only
+	errI2CMultipleDevices    = errors.New("I2C bus has address conflict")                   // simulator only
+	errI2CWrongAddress       = errors.New("I2C bus has devices but none with this address") // simulator only
```
eliasnaur: The prefixes should be changed to the more idiomatic "i2c: " like in machine_rp2_i2c.go (or vice versa).

src/machine/machine_generic.go
```diff
@@ -239,7 +242,33 @@ type I2CConfig struct {
...
+				matchSDA = true
+				break
+			}
+		}
+		if !matchSCL && !matchSDA {
+			return errors.New("I2C: SCL and SDA pins are incorrect for this I2C instance")
+		} else if !matchSCL {
+			return errors.New("I2C: SCL pin is incorrect for this I2C instance")
+		} else if !matchSDA {
+			return errors.New("I2C: SDA pin is incorrect for this I2C instance")
```
eliasnaur: Nit: "i2c:".

src/machine/machine_generic.go
```diff
@@ -239,7 +242,33 @@ type I2CConfig struct {
...
+				break
+			}
+		}
+		matchSDA := false
+		for _, pin := range i2c.PinsSDA {
+			if pin == config.SDA {
+				matchSDA = true
+				break
+			}
+		}
```
eliasnaur: Nit: shorten with `slices.IndexFunc`?

eliasnaur (approved):
LGTM

deadprogram:
CI test failure appears to be similar to https://github.com/chromedp/chromedp/issues/1484

src/machine/machine_generic.go
aykevl: Thank you, good idea.

aykevl:
@eliasnaur I've implemented your suggestions.

deadprogram:
@aykevl please rebase against lastest `dev` to pass CI.

aykevl:
Rebased!

deadprogram:
Thank you for the improvements @aykevl and to @eliasnaur for review. Now merging.

PR #4951, opened 2025-07-09
chore: update chromedp to try to avoid test failures on updated Linux versions

deadprogram:
THis PR is to update the `chromedp` package to try to avoid test failures on updated Linux versions.


aykevl:
Was the Go version bump intentional?

deadprogram:
> Was the Go version bump intentional?

It is required by the newer version, apparently.

In any case, this PR not only did not fix the build, but also breaks some of the wasip2 code.

deadprogram:
Closing since it appears to cause even more problems.

PR #4957, opened 2025-07-15
runtime: ensure time.Sleep(d) sleeps at least d

eliasnaur:
Account for the sleep queue base time in the computation of the wakeup time.

Tested with the following program on pico2.

  func main() {
  	go func() {
  		for i := range 60 {
  			const delay = 20 * time.Millisecond
  			before := time.Now()
  			time.Sleep(delay)
  			if d := time.Since(before); true || d < delay {
  				log.Println(i, "actual", d, "delay", delay)
  			}
  		}
  	}()
  	time.Sleep(500 * time.Millisecond)
  	log.Println("******** done sleeping ********")
  	select {}
  }

Without this change, the program would print lines such as:

  17 actual 15.494ms delay 20ms
  18 actual 15.49ms delay 20ms
  19 actual 15.585ms delay 20ms
  20 actual 15.493ms delay 20ms
  21 actual 15.494ms delay 20ms
  22 actual 15.487ms delay 20ms
  23 actual 15.498ms delay 20ms
  ******** done sleeping ********
  24 actual 15.548ms delay 20ms
  25 actual 20.011ms delay 20ms
  26 actual 20.01ms delay 20ms
  27 actual 20.011ms delay 20ms
  28 actual 20.015ms delay 20ms

Note that while more than one sleeping goroutine is in the timer queue, the sleep duration is 5ms short.

PR #4958, opened 2025-07-17
tests: de-flake goroutines test

eliasnaur:
Hopefully helps with #4957.

Added a few `t.Helper` calls to better distinguish call sites.


deadprogram:
Seems like a good idea to me! Thanks for the improvement here @eliasnaur now merging.

PR #4962, opened 2025-07-20
Add flag to skip Renesas SVD builds

QuLogic:
Much like the STMicro SVDs, Renesas SVDs are fully proprietary and not under an FOSS license.


deadprogram:
> Renesas SVDs are fully

That is a bummer, I did not realize that. Alas, now merging.

PR #4964, opened 2025-07-20
Makefile: Install missing dlmalloc files

QuLogic:
Running tests against an installed TinyGoRoot fails with:
```
=== CONT  TestBuild/WebAssembly/gc.go-boehm
    main_test.go:441: wasi-libc: did not find any files for pattern builder.filePattern{glob:"dlmalloc/src/dlmalloc.c", exclude:[]string(nil)}
```
because these files are not there, and are required with Boehm GC.


deadprogram:
cc @aykevl

aykevl:
Yes looks correct, thank you!

PR #4973, opened 2025-08-02
chore: correct GOOS=js name in error messages for WASM

dsseng:
These changes only affect user-facing error strings, and bring no functional change


deadprogram:
Thanks for the correction @dsseng now merging.

PR #4977, opened 2025-08-04
internal/task: add SA_RESTART flag to GC interrupts

aykevl:
This makes sure system calls like read don't return EINTR but instead restart the call on an interrupt. This is by far the more sensible option, the default POSIX behavior of returning EINTR is extremely error-prone.

Found this bug while trying to use the upstream testing package instead of our own.


eliasnaur (approved):

deadprogram:
Thanks for the fix @aykevl and to @eliasnaur for quick review. Now merging.

PR #4978, opened 2025-08-04
internal/task: use -stack-size flag when starting a new thread

aykevl:
Found this bug while trying to use the upstream testing package instead of our own. The io/fs package wasn't passing, because the test was run in a separate goroutine (and therefore a separate thread, with its own stack) instead of all in the same thread with our own stack creation/switching implementation.


eliasnaur (approved):
LGTM.

deadprogram:
Thanks for the correction @aykevl and to @eliasnaur for review.

Now merging.

PR #4979, opened 2025-08-04
Full LLVM 20 support

aykevl:
We've been using LLVM 19 so far, since that was available in the Espressif fork. However I noticed they have an unreleased LLVM 20 fork. So here's a start at porting TinyGo to use this newer LLVM version.

It's not yet finished, I've seen issues with AVR at least.


aykevl:
Reported the AVR crash upstream: https://github.com/llvm/llvm-project/issues/152097

aykevl:
Fixed the AVR crash (https://github.com/llvm/llvm-project/pull/152269) and cherry-picked it in a branch of https://github.com/tinygo-org/llvm-project.

Also hopefully fixed an issue with LLVM 15, which we still support. I did this by switching the wasm-unknown `cpu` to "mvp", which is a WebAssembly CPU that only supports the absolute bare minimum of WebAssembly features. The previous value of "generic" changes across LLVM releases, which is probably not what we want.

deadprogram:
Looks like something is still a problem? https://github.com/tinygo-org/tinygo/actions/runs/16775128593/job/47510760185#step:22:15

aykevl:
Yeah I need to investigate that one. Will probably do that after WHY2025.

deadprogram:
Now that Homebrew is already on to LLVM21 we probably should get on this.

aykevl:
Rebased, hopefully fixed one of the baremetal tests. No clue about `crypto/elliptic: ScalarMult was called on an invalid point`

aykevl:
Spent a few hours debugging. The crypto/elliptic failure appears to be a miscompilation. When I add `//go:noinline` to `crypto/internal/fips140/nistec/fiat.p256Mul` the test passes. Which is really frustrating since that will make it much harder to find the actual bug.

aykevl:
Disabling the machine outliner for RISC-V appears to fix the issue with crypto/elliptic. So it's likely a LLVM bug. I have not investigated this bug further, I just hope it gets fixed in a future version :sweat_smile:

aykevl:
CI is now passing :tada:

GNUmakefile
```diff
@@ -263,7 +263,7 @@ gen-device-renesas: build/gen-device-svd
 	GO111MODULE=off $(GO) fmt ./src/device/renesas
 
 $(LLVM_PROJECTDIR)/llvm:
-	git clone -b xtensa_release_19.1.2 --depth=1 https://github.com/espressif/llvm-project $(LLVM_PROJECTDIR)
+	git clone -b tinygo_20.x --depth=1 https://github.com/tinygo-org/llvm-project $(LLVM_PROJECTDIR)
```
deadprogram, line 266: Yes! Thank you for pointing this to our own fork!! :package:

targets/wasip2.json
```diff
@@ -1,7 +1,7 @@
 {
 	"llvm-target":   "wasm32-unknown-wasi",
 	"cpu":           "generic",
-	"features":      "+bulk-memory,+mutable-globals,+nontrapping-fptoint,+sign-ext,-multivalue,-reference-types",
+	"features":      "+bulk-memory,+bulk-memory-opt,+call-indirect-overlong,+mutable-globals,+nontrapping-fptoint,+sign-ext,-multivalue,-reference-types",
```
deadprogram, line 4: These changes might be of particular note to the WASM oriented people who might be using their own wasm-unknown JSON targets. cc/ @dgryski @ydnar @orsinium

deadprogram:
Thank you for getting this working @aykevl now merging.

targets/wasip2.json
orsinium: all is good from my side, [we use a custom target.json](https://github.com/firefly-zero/firefly-cli/blob/main/src/target.json) 👍🏿

deadprogram: I mean, you might need to modify your custom JSON if you use LLVM20+. Not sure on that, just mentioning the possibility.

aykevl: I don't think these changes actually change anything. For example `+bulk-memory-opt` was split out of `+bulk-memory` so nothing changed there - it was already included just not named separately.

PR #4982, opened 2025-08-06
ci: make the goroutines test less racy

aykevl:
Since we switched to OS threads for goroutines, the testdata/goroutines.go test has been flaky. Not surprising, if threads need to be scheduled within 1ms on a busy CI system. This PR makes the test a bit less flaky (hopefully) by increasing the time to 100ms.


eliasnaur (approved):
LGTM

deadprogram:
OK merging. Thanks everyone!

PR #4986, opened 2025-08-07
feat: enable multi-core scheduler for rp2350

mikesmitty:
Been testing this locally on a project with multiple goroutines and it seems to be working great


eliasnaur (approved):
LGTM, if for nothing else that the rp2xxx implementation are so nearly identical.

deadprogram:
Thank you for the addition @mikesmitty and to @eliasnaur for review. Now merging.

PR #4987, opened 2025-08-08
fix: add TryLock methods to sync.RWMutex

mikesmitty:
I saw these were recently added to sync.Mutex, figured they should probably be on RWMutex too. Implementations sourced from big go with the race detection removed

### DIFF
--- a/src/sync/mutex.go
+++ b/src/sync/mutex.go
@@ -31,6 +31,9 @@ type RWMutex struct {
 
 const rwMutexMaxReaders = 1 << 30
 
+// Lock locks rw for writing.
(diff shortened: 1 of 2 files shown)


deadprogram (approved):
Looks good to me. @aykevl or anyone else have any comment before merge?

deadprogram:
@mikesmitty sorry about the late comment here, but perhaps you can add this to the tests for this? What do you think?

mikesmitty:
Oh yeah, sure thing

deadprogram:
Thanks for the addition here @mikesmitty now merging.

PR #4994, opened 2025-08-14
os: remove VFS layer

aykevl:
This was largely non-working and I don't think anybody actually used it. So let's remove it.

I hope to implement some sort of virtual filesystem support later that actually works, which is simpler with the non-working abstraction removed.


aykevl:
Hmm, perhaps I should first make a better abstraction before randomly deleting code.

deadprogram:
Also what about https://github.com/tinygo-org/tinyfs/blob/dev/tinygo_os.go ?

PR #5001, opened 2025-08-18
targets: set default stack size to 8k for rp2040/rp2350 based boards

deadprogram:
This PR sets the default stack size to 8k for rp2040/rp2350 based boards where this was not already the case.

Should avoid a lot of errors from just forgetting to add that flag, since either everyone practically just sets it, or in a number of cases is already the default.


eliasnaur (approved):

deadprogram:
Thank you for the quick review @eliasnaur now merging.

Issue #4998, opened 2025-08-16
xiao-ble (sense), i2c, nrf52840 stucks with unconnected bme280


gen2thomas:
In my program I read the sensor each 10 sec. After some time the MCU stop working always on the same line. I print output before calling the sensor and also inside the sensor driver to catch this issue. I have already optimized the driver code [to prevent all heap allocations](https://github.com/tinygo-org/drivers/issues/760) and for the first time it seems to improve the behavior, but after some repeated tests it shows that nothing has really changed.

```go
// this is called each 10 sec and calls the driver implementation
// the caller implements an println(err.Error()) in case of an error is returned
func (as *AirSensor) RefreshMeasurement() error {
	// TODO: this call seems to hang after some time if the sensor is not present	
	if !as.driver.Connected() {
		println("ss0E")
		return errors.New("BME280 not connected")
	}
...

```
```go
// Connected returns whether a BME280 has been found.
// It does a "who am I" request and checks the response.
// dataBuf [8]byte, regBuf  [2]byte
func (d *Device) Connected() bool {
	println("b0")
	//data := []byte{0}
	d.dataBuf[0] = 0
	data := d.dataBuf[:1]
	println("b1")
	//legacy.ReadRegister(d.bus, uint8(d.Address), WHO_AM_I, data)
	//d.bus.Tx(uint16(d.Address), []byte{WHO_AM_I}, data)
	d.regBuf[0] = WHO_AM_I
	d.bus.Tx(uint16(d.Address), d.regBuf[:1], data)
	println("b2")
	return data[0] == CHIP_ID
}
```

If working like expected, the output looks like this:

```
b0
b1
b2
ss0E
BME280 not connected
```

In case of stuck it looks always like this;

```
b0
b1
```

So I had a look to the TX implementation of tinygo. May it be possible, that in case of error there [is no return path of the loop](https://github.com/tinygo-org/tinygo/blob/733a17d470870afc9c821524ce291ec5062d682b/src/machine/machine_nrf528xx.go#L88) and this causes the issue?:
```go
	for i2c.Bus.EVENTS_STOPPED.Get() == 0 {
		// Allow scheduler to run
		gosched()

		// Handle errors by ensuring STOP sent on bus
		if i2c.Bus.EVENTS_ERROR.Get() != 0 {
			if i2c.Bus.EVENTS_STOPPED.Get() == 0 {
				// STOP cannot be sent during SUSPEND
[...]

PR #5007, opened 2025-08-21
machine,nrf528,i2c: fix multiple resume-stop on bus error

gen2thomas:
As already shown in #4998,  in case of an error the i2c TX method in nrf528xx loops until a stop condition is fulfilled. In case of an error, a resume-stop command on bus will be called, so the bus can finish its work and exit the loop afterwards - no "hard" return in case of error is done and needed.

With the current code, after the first error occurs, each loop checks again for an error and practically (assuming the error is not gone) sends the resume-stop sequence again and again until the stop state is reached (triggered by the first resume-stop-sequence) and the loop is finished. On long run with existing errors, e.g the device is not connected on the bus, and cyclic calls (e.g. each 10 seconds) this leads reproducible to hang of the loop with the [bme280](https://github.com/tinygo-org/drivers/tree/release/bme280) driver.

The provided code change improves the code for this topics:
* checks for the error code before the resume-stop is called, so the real error is finally contained in err variable
* skips the complete error check if the first error was detected - this fixes the originated problem

### DIFF
--- a/src/machine/machine_nrf528xx.go
+++ b/src/machine/machine_nrf528xx.go
@@ -78,14 +78,14 @@ func (i2c *I2C) Tx(addr uint16, w, r []byte) (err error) {
 		// Allow scheduler to run
 		gosched()
 
-		// Handle errors by ensuring STOP sent on bus
-		if i2c.Bus.EVENTS_ERROR.Get() != 0 {
+		// Handle first occurrence of error by ensuring STOP sent on bus
+		if err == nil && i2c.Bus.EVENTS_ERROR.Get() != 0 {
+			err = twiCError(i2c.Bus.ERRORSRC.Get())
 			if i2c.Bus.EVENTS_STOPPED.Get() == 0 {
 				// STOP cannot be sent during SUSPEND
 				i2c.Bus.TASKS_RESUME.Set(1)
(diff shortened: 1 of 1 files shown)


eliasnaur:
LGTM, but please copy the relevant parts of the PR explanation to the commit message.

gen2thomas:
@eliasnaur second topic added to the commit message - is the prefix ok? Is there a special guideline for commit messages on tinygo?

eliasnaur:
I'm fairly new here, and not aware of particular guidelines for commit messages. My personal preference is to retain details in the commit message, because GitHub PRs are separate from the repository. That is, PR descriptions are for helping reviewers, and commit messages to describe the changes.

For your commit, I suggest something like

nrf528: stop the bus only once on I2C bus error

In some cases, e.g nothing connected on the bus, repeated resume-stop sequences can lead to the bus never reaching the stop state, hanging Tx.
This change ensures the resume-stop sequence is submitted once on error. It also moves the error code read to before the sequence to ensure it's valid.

Fixes: #4998

gen2thomas:
thanks @eliasnaur - are there any other suggestion @deadprogram ?

deadprogram:
Looks good to me, thank you very much @gen2thomas for the fix and to @eliasnaur for review.

PR #5009, opened 2025-08-22
fix: createInlineAsm:avoid mapupdate collect after call

### LINKED ISSUE #5008: Bug Report: AsmFull incorrectly processes MapUpdate operations after the call
## Summary

I discovered an issue with `device.AsmFull` where it incorrectly processes `MapUpdate` operations that occur **after** the AsmFull call, causing the function to return unexpected values from future map modifications instead of the values present at call time.

## Problem Description

The issue is in `compiler/inlineasm.go` at line 62, where the code uses `break` to exit a `switch` statement but continues processing subsequent `MapUpdate` operations in the referrers loop:

```go
case *ssa.Call:
    if r.Common() == instr {
        break  // ❌ Only breaks from switch, continues for loop
    }
```

This means that map updates happening **after** the AsmFull call are still collected and can affect the result.

## Reproduction

```go
package main

import (
    "device"
    "machine"
    "time"
)

func main() {
    machine.Serial.Configure(machine.UARTConfig{BaudRate: 115200})
    time.Sleep(1 * time.Second)
    
    place := map[string]interface{}{
        "input": uint32(42),
    }
    
    result := device.AsmFull("mov {}, {input}", place)
    place["input"] = uint32(44) // This MapUpdate affects the result!
    
    println("Expected: 42, Got:", uint32(result))
}
```

### Expected Result
```
Expected: 42, Got: 42
```

### Actual Result (Before Fix)
```
Expected: 42, Got: 44
```

## Root Cause Analysis

The original code appears to intend to stop processing referrers when encountering the AsmFull call itself, but uses `break` which only exits the `switch` statement, not the entire `for` loop. This allows subsequent MapUpdate operations to be processed incorrectly.

## Impact

This bug can cause:  **Unexpected behavior** where AsmFull returns "future" values instead of call-time values

Thank you for maintaining TinyGo! This is an amazing project and I'm happy to contribute to its improvement. 😊

luoliwoshang:
fixed https://github.com/tinygo-org/tinygo/issues/5008

Problem:
The original code used a break statement inside a switch within a for loop, which only exited the switch but continued processing subsequent MapUpdate operations. This led to AsmFull returning register values from map updates  that happened after the function call, violating the expected behavior.

Solution:
- Stops processing when encountering the AsmFull call itself

### DIFF
--- a/compiler/inlineasm.go
+++ b/compiler/inlineasm.go
@@ -47,6 +47,7 @@ func (b *builder) createInlineAsmFull(instr *ssa.CallCommon) (llvm.Value, error)
 	asmString := constant.StringVal(instr.Args[0].(*ssa.Const).Value)
 	registers := map[string]llvm.Value{}
 	if registerMap, ok := instr.Args[1].(*ssa.MakeMap); ok {
+	referrers:
 		for _, r := range *registerMap.Referrers() {
 			switch r := r.(type) {
 			case *ssa.DebugRef:
@@ -59,7 +60,9 @@ func (b *builder) createInlineAsmFull(instr *ssa.CallCommon) (llvm.Value, error)
 				registers[key] = b.getValue(r.Value.(*ssa.MakeInterface).X, getPos(instr))
 			case *ssa.Call:
 				if r.Common() == instr {
-					break
+					// Stop processing when we encounter the AsmFull call itself
+					// to avoid including MapUpdate operations that happen after the call
+					break referrers
 				}
 			default:
 				return llvm.Value{}, b.makeError(instr.Pos(), "don't know how to handle argument to inline assembly: "+r.String())
--- a/compiler/compiler_test.go
+++ b/compiler/compiler_test.go
@@ -50,6 +50,7 @@ func TestCompiler(t *testing.T) {
 		{"channel.go", "", ""},
 		{"gc.go", "", ""},
 		{"zeromap.go", "", ""},
+		{"asmfull.go", "", ""},
 	}
 	if goMinor >= 20 {
 		tests = append(tests, testCase{"go1.20.go", "", ""})
--- a/compiler/testdata/asmfull.go
+++ b/compiler/testdata/asmfull.go
@@ -1,0 +1,21 @@
+package main
+
+import "device"
+
+// Test the specific bug case: AsmFull should use call-time input register value (42), not post-call value (44)
+func testAsmFullBug() uintptr {
+	place := make(map[string]interface{})
+	place["input"] = uint32(42)
+
+	// This should use input=42 at call time, return the moved value
+	result := device.AsmFull("mov {}, {input}", place)
+
+	// This update should NOT affect the AsmFull result (was the bug)
+	place["input"] = uint32(44)
+
+	return result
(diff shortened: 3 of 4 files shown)


compiler/inlineasm.go
```diff
@@ -46,24 +75,12 @@ func (b *builder) createInlineAsm(args []ssa.Value) (llvm.Value, error) {
...
-				// ignore
-			case *ssa.MapUpdate:
-				if r.Block() != registerMap.Block() {
-					return llvm.Value{}, b.makeError(instr.Pos(), "register value map must be created in the same basic block")
-				}
-				key := constant.StringVal(r.Key.(*ssa.Const).Value)
-				registers[key] = b.getValue(r.Value.(*ssa.MakeInterface).X, getPos(instr))
-			case *ssa.Call:
-				if r.Common() == instr {
-					break
```
eliasnaur: Instead of extracting `collectInlineAsmRegisters`, why not name the outer loop and use `break <name>`? Something like:

```go
referrers:
    for _, r := range *registerMap.Referrers() {
...
        if r.Common() == instr {
	        break referrers
        }
```

eliasnaur (commented):
Probably for @aykevl

compiler/inlineasm.go
luoliwoshang: Thanks for the great suggestion! You're absolutely right - using labeled break is cleaner and more idiomatic. I've updated the code accordingly. I initially went with function extraction due to personal preference, but your approach is definitely better with fewer code changes. Thanks for the review!

eliasnaur:
The test disappeared, was that on purpose? Also, please squash the two "goto" and "named break" commits into one.

Probably @aykevl for the meat of this change.

luoliwoshang:
> The test disappeared, was that on purpose? Also, please squash the two "goto" and "named break" commits into one.

I have merged the two commits into one. I temporarily removed the added tests to confirm whether the test failure issue is only coming from the CI environment. I will add back the tests shortly.

luoliwoshang:
Hey @aykevl! 👋 I found a little bug in TinyGo and managed to fix it - would love to get your eyes on the PR when you have a moment!
Thanks for building such an awesome project! TinyGo makes embedded Go so much fun, and I'd love to be part of the contributor community

aykevl:
While I appreciate you fixing this, I would actually prefer if we could remove AsmFull entirely. It's not very well defined, and C (GCC/Clang) has a far more flexible way to write inline assembly. It's annoying to use, yes, but doesn't require TinyGo to know all the details of an architecture.

Here is an example of how it is done in the WS2812 driver:
https://github.com/tinygo-org/drivers/blob/release/ws2812/ws2812-asm_cortexm.go

luoliwoshang:
Thank you for your feedback and for sharing the WS2812 example.
I completely understand your preference to remove AsmFull in favor of standard C inline assembly, which is indeed more flexible and reduces maintenance burden on TinyGo.
However, if `AsmFull` won't be removed in the immediate next release, would this fix still provide some value in the interim?
I totally respect your decision either way and am happy to follow whatever direction you think is best for the project.

PR #5010, opened 2025-08-23
machine: clarify WriteAt semantics of BlockDevice

aykevl:
Flash can generally only be written after it has been erased, which wasn't spelled out explicitly in the interface.

This should clarify issues like the write-without-erase bug I found in https://github.com/tinygo-org/tinygo/pull/4844.


eliasnaur (approved):

deadprogram:
:scroll: thanks for the documentation clarification. Now merging.

Issue #4995, opened 2025-08-14
ci: macOS CI builds have intermittent lockup during the stdlib tests


deadprogram:
There is an intermittent lockup during the stdlib tests when running the macOS CI builds on GH actions.

It occurs both with macOS 13 and macOS 14 during the execution of `make tinygo-test`

The lockup can occur on different tests. The result is the same: test execution stops until the runner times out.

eliasnaur:
Maybe (hopefully?) related: https://github.com/tinygo-org/tinygo/issues/4974

eliasnaur:
It seems like there are no time limit to the tests, other than the GH Actions limit? It would be nice to have a forced timelimit like `go test` has 10 minutes, and then get a goroutine dump.

eliasnaur:
Another hang: https://github.com/tinygo-org/tinygo/actions/runs/17197258565/job/48781195261?pr=5009

deadprogram:
@aykevl this is actually happening quite frequently. Any chance you might look into the root cause, please?

deadprogram:
Labeling to close this issue on next release.

PR #5013, opened 2025-08-25
testing: implement -timeout flag

eliasnaur:
Note that this PR renames the existing `-timeout` flag to `-vol-timeout`.

Should reduce the multi-hour hangs on macOS builders (#4995) to ~10 minutes.

### DIFF
--- a/compileopts/config.go
+++ b/compileopts/config.go
@@ -10,6 +10,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/google/shlex"
 	"github.com/tinygo-org/tinygo/goenv"
@@ -659,4 +660,5 @@ type TestConfig struct {
 	BenchTime         string
 	BenchMem          bool
 	Shuffle           string
+	Timeout           time.Duration
 }
--- a/main.go
+++ b/main.go
@@ -1628,7 +1628,7 @@ func main() {
 	ocdCommandsString := flag.String("ocd-commands", "", "OpenOCD commands, overriding target spec (can specify multiple separated by commas)")
 	ocdOutput := flag.Bool("ocd-output", false, "print OCD daemon output during debug")
 	port := flag.String("port", "", "flash port (can specify multiple candidates separated by commas)")
-	timeout := flag.Duration("timeout", 20*time.Second, "the length of time to retry locating the MSD volume to be used for flashing")
+	timeout := flag.Duration("volume-timeout", 20*time.Second, "the length of time to retry locating the MSD volume to be used for flashing")
 	programmer := flag.String("programmer", "", "which hardware programmer to use")
 	ldflags := flag.String("ldflags", "", "Go link tool compatible ldflags")
 	llvmFeatures := flag.String("llvm-features", "", "comma separated LLVM features to enable")
--- a/main.go
+++ b/main.go
@@ -238,6 +238,9 @@ func Test(pkgName string, stdout, stderr io.Writer, options *compileopts.Options
 	if testConfig.Shuffle != "" {
 		flags = append(flags, "-test.shuffle="+testConfig.Shuffle)
 	}
+	if t := testConfig.Timeout; t != 0 {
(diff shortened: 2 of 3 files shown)


deadprogram:
> Note that this PR renames the existing `-timeout` flag to `-vol-timeout`.

Is that a problem for anyone?

> Should reduce the multi-hour hangs on macOS builders (#4995) to ~10 minutes.

Yes please! :smile_cat:

deadprogram:
Just wondering do we actually need to rename the flag? Do your other changes actually require that, @eliasnaur?

eliasnaur:
> Just wondering do we actually need to rename the flag? Do your other changes actually require that, @eliasnaur?

I'm not sure I understand your question. Are you suggesting to keep `-timeout` for volume timeouts, or to overload its meaning to depend on whether you're running `tinygo flash` or `tinygo test`?

deadprogram:
> overload its meaning to depend on whether you're running `tinygo flash` or `tinygo test`

I think that is what I meant.

eliasnaur:
> > overload its meaning to depend on whether you're running `tinygo flash` or `tinygo test`
> 
> I think that is what I meant.

It's possible, but I don't think it's worth the confusion. Besides, "timeout" is not a great name for something as specific as "the length of time to retry locating the MSD volume to be used for flashing".

deadprogram:
Fair point. In that case, how about renaming it to `flash-timeout`?

eliasnaur:
> Fair point. In that case, how about renaming it to `flash-timeout`?

The timeout is for waiting for the flash volume, not flashing in itself. For example, if you're flashing through a debugger `-flash-timeout` would be ignored.

How about `-volume-timeout`?

deadprogram:
Sounds good to me!

deadprogram:
@eliasnaur can you please rebase this branch against the latest `dev`? Not sure why the CI is failing.

eliasnaur:
I already did, and I'm not sure either.

For some reason, my Linux VM fails to run `make tinygo-test` with the following error:

```sh
$ nix develop
$ go install
$ make tinygo-test
/home/e/go/bin/tinygo test -skip='TestExtraMethods|TestParseAndBytesRoundTrip/P256/Generic' cmp compress/lzw compress/zlib container/heap container/list container/ring crypto/ecdsa crypto/elliptic crypto/md5 crypto/sha1 crypto/sha256 crypto/sha512 database/sql/driver debug/macho embed/internal/embedtest encoding encoding/ascii85 encoding/asn1 encoding/base32 encoding/base64 encoding/csv encoding/hex go/ast go/format go/scanner go/version hash hash/adler32 hash/crc64 hash/fnv html internal/itoa internal/profile math math/cmplx net/http/internal/ascii net/mail os path reflect sync testing testing/iotest text/scanner unicode unicode/utf16 unicode/utf8 unique  archive/zip compress/flate crypto/aes crypto/des crypto/hmac debug/dwarf debug/plan9obj image io/ioutil mime mime/multipart mime/quotedprintable net net/mail net/textproto os/user regexp/syntax strconv text/tabwriter text/template/parse compress/bzip2 crypto/dsa index/suffixarray 
package net/http/internal/ascii is not in std (/home/e/.cache/tinygo/goroot-d6ba3aaea39ea12a38e3f8b292ba2e412e6bfc30c3b3b8a879b97c4a782dc8a2/src/net/http/internal/ascii)
../../.cache/tinygo/goroot-d6ba3aaea39ea12a38e3f8b292ba2e412e6bfc30c3b3b8a879b97c4a782dc8a2/src/net/mail/message.go:27:2: package net is not in std (/home/e/.cache/tinygo/goroot-d6ba3aaea39ea12a38e3f8b292ba2e412e6bfc30c3b3b8a879b97c4a782dc8a2/src/net)
cannot resolve packages: exit status 1
make: *** [GNUmakefile:491: tinygo-test] Error 1
```

deadprogram:
@eliasnaur see #5080 for a different fix for the test race, although not for the root cause itself.

PR #5014, opened 2025-08-25
feat: allow bidirectional usb endpoint numbers

mikesmitty:
This one is still a bit fresh, will update once it's had a bit more time to cook


mikesmitty:
Ok, all ready for review now. I don't have any atsamd21 boards, but I've tested USB CDC on an atsamd51, nrf52840, and a pico-w as well as USB mass storage on the pico-w and everything looks good to me

src/machine/machine_atsamd21_usb.go
```diff
@@ -20,19 +20,24 @@ const (
 	usb_DEVICE_PCKSIZE_MULTI_PACKET_SIZE_Pos  = 14
 	usb_DEVICE_PCKSIZE_MULTI_PACKET_SIZE_Mask = 0x3FFF
 
-	NumberOfUSBEndpoints = 8
+	NumberOfUSBEndpoints = 6
 )
 
 var (
-	endPoints = []uint32{
+	inEndpoints = []uint32{
```
aykevl: You could consider changing the code to something like:

```suggestion
	inEndpoints = [...]uint32{
```

That makes it a fixed-size array instead of a slice.

If you do that, you can probably also make `NumberOfUSBEndpoints` more descriptive (possibly moving it to common code):

```go
const NumberOfUSBEndpoints = max(len(inEndpoints), len(outEndpoints))
```

aykevl (commented):
One suggestion, didn't review the code in depth but I see nothing weird at a quick glance.

src/machine/machine_atsamd21_usb.go
mikesmitty: Ah, good idea. That reminds me, this PR fixes the reason that endpoints were moved from machine/usb.go out into each platform, would be a good time to undo that.

aykevl:
Did a quick test on a Circuit Playground Express (atsamd21) and it seems to be working fine!

deadprogram:
@mikesmitty I just switched the branch for this to `dev`. No merge conflicts. Do you want to squash any of these commits?

Anyone else have any comments on this PR?

mikesmitty:
Sure, I'll squash them all

deadprogram:
Before merge, just want to confirm with @sago35 that this does not affect any TinyGo Keeb code. :keyboard:

sago35:
I’ll make sure to review it early next week.

src/machine/usb/descriptor/midi.go
```diff
@@ -199,10 +199,15 @@ var endpointEP7OUT = [endpointMIDITypeLen]byte{
...
-	data: endpointEP7OUT[:],
+var EndpointEP3OUT = EndpointType{
+	data: endpointEP3OUT[:],
 }
 
+var (
+	EndpointMIDIIN  = &EndpointEP4IN
+	EndpointMIDIOUT = &EndpointEP3OUT
+)
+
```
sago35, line 210: I think it would be better to move it to `src/machine/usb/descriptor/endpoint.go`.

src/machine/machine_atsamd21_usb.go
```diff
@@ -197,7 +182,8 @@ func handleUSBIRQ(intr interrupt.Interrupt) {
 			if usbRxHandler[i] == nil || usbRxHandler[i](buf) {
 				AckUsbOutTransfer(i)
 			}
-		} else if (epFlags & sam.USB_DEVICE_EPINTFLAG_TRCPT1) > 0 {
+		}
+		if (epFlags & sam.USB_DEVICE_EPINTFLAG_TRCPT1) > 0 {
```
sago35, line 186: What effect does this change have?
Depending on the nature of the change, it might be better to split it into a separate PR.

src/machine/machine_atsamd51_usb.go
```diff
@@ -200,7 +185,8 @@ func handleUSBIRQ(intr interrupt.Interrupt) {
 			if usbRxHandler[i] == nil || usbRxHandler[i](buf) {
 				AckUsbOutTransfer(i)
 			}
-		} else if (epFlags & sam.USB_DEVICE_ENDPOINT_EPINTFLAG_TRCPT1) > 0 {
+		}
+		if (epFlags & sam.USB_DEVICE_ENDPOINT_EPINTFLAG_TRCPT1) > 0 {
```
sago35, line 189: What effect does this change have?
Depending on the nature of the change, it might be better to split it into a separate PR.

src/machine/machine_nrf52840_usb.go
```diff
@@ -183,15 +170,16 @@ func handleUSBIRQ(interrupt.Interrupt) {
...
 			// Check if endpoint has a pending interrupt
 			inDataDone := epDataStatus&(nrf.USBD_EPDATASTATUS_EPIN1<<(i-1)) > 0
 			outDataDone := epDataStatus&(nrf.USBD_EPDATASTATUS_EPOUT1<<(i-1)) > 0
 			if inDataDone {
 				if usbTxHandler[i] != nil {
 					usbTxHandler[i]()
 				}
-			} else if outDataDone {
+			}
+			if outDataDone {
```
sago35, line 182: What effect does this change have?
Depending on the nature of the change, it might be better to split it into a separate PR.

sago35 (commented):
Some parts of my TinyGo Keeb code didn’t work, but that turned out to be an issue on my end.
So overall, everything looks really good.

There are just a few parts that may need slight changes, and a couple of things I'd like to understand the reasoning behind.

deadprogram:
I think this PR has been deprecated by https://github.com/tinygo-org/tinygo/pull/5405 is that correct?

PR #5015, opened 2025-08-27
rp2040: allow writing to the UART inside interrupts

aykevl:
Writing to the UART takes time and that may not be a good idea inside an interrupt, but it is essential for debugging sometimes (especially since USB-CDC typically doesn't work inside an interrupt). However, `runtime.Gosched()` tries to switch to a different goroutine and will panic when it is called inside an interrupt.

This fixes UART support in interrupts for the RP2040 at least. You can test it with `-serial=uart` and connecting a USB-UART adapter to the right pins.


eliasnaur (approved):

deadprogram:
Thank you for the improvement @aykevl and to @eliasnaur for review. Now merging...

PR #5016, opened 2025-08-27
Pause for interrupt

aykevl:
This should make usb/msc a whole lot more efficient by pausing the worker goroutine and waiting for an interrupt to unpause it instead of waiting in a loop and sleeping for 0.1ms each cycle.

In other words, this should make it both faster (no unnecessary delay due to the time.Sleep) and more efficient (no polling).

To make this possible, I've added `internal/task.Waiter` which allows waiting for a signal from an interrupt. Basically, a goroutine can pause itself and wait for the interrupt to unpause itself when some event happens. I imagine this feature will be useful in many other places in the machine package for example (e.g. inside an SPI transfer using DMA, the goroutine can pause itself and wait for a "transfer finished" interrupt instead of polling using `runtime.Gosched` or similar).

@mikesmitty can you take a look at the USB logic to verify it is correct? It seems to work under MacOS.

---

Unfortunately this suffers from a potential hang with `-scheduler=cores`: `scheduleTask` takes the scheduler lock from interrupt context which would cause a deadlock when this interrupt happens with the scheduler lock already held. I should try and fix this in a separate PR. So maybe it's best to hold off of merging this PR until that bug is fixed.

### DIFF
--- a/src/internal/task/waiter.go
+++ b/src/internal/task/waiter.go
@@ -1,0 +1,99 @@
+package task
+
+import (
+	"runtime/interrupt"
+	"unsafe"
+)
+
+// A waiter waits for an interrupt to fire. Call Wait() from a goroutine to
+// pause it, and call Resume() from an interrupt handler to resume the
+// goroutine.
+type Waiter struct {
+	waiting *Task
+	lock    PMutex
+}
+
+var resumeTaskSentinel = (*Task)(unsafe.Pointer(uintptr(2)))
+var workingTaskSentinel = (*Task)(unsafe.Pointer(uintptr(1)))
+
+// Wait from a main goroutine until an interrupt calls Resume(). It will also
+// return if the interrupt called Resume() before the Wait() call (to avoid a
+// race condition).
+func (w *Waiter) Wait() {
+	if interrupt.In() {
+		runtimePanic("Waiter: called Wait from interrupt")
+	}
+
+	mask := interrupt.Disable()
+	w.lock.Lock()
+	switch w.waiting {
+	case nil, workingTaskSentinel:
+		w.waiting = Current()
+		w.lock.Unlock()
+		interrupt.Restore(mask)
+		Pause()
+	case resumeTaskSentinel:
+		// Marked as 'resume now', can return immediately.
+		w.waiting = workingTaskSentinel
+		w.lock.Unlock()
+		interrupt.Restore(mask)
+	default:
+		w.lock.Unlock()
+		interrupt.Restore(mask)
+		runtimePanic("Waiter: task is waiting already")
+	}
+}
+
+// Resume a waiting goroutine from an interrupt handler. If there is a goroutine
+// waiting, it will resume. If not, the next one that calls Wait() will
+// immediately resume.
+func (w *Waiter) Resume() {
+	if !interrupt.In() {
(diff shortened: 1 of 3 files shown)


mikesmitty:
Sure, I'll give it a look here in a few. That bit was an ugly hack, glad to see it go

src/internal/task/waiter.go
```diff
@@ -0,0 +1,98 @@
...
+type Waiter struct {
+	waiting *Task
+	lock    PMutex
+}
+
+var resumeTaskSentinel = (*Task)(unsafe.Pointer(uintptr(2)))
+var workingTaskSentinel = (*Task)(unsafe.Pointer(uintptr(1)))
+
+// Wait from a main goroutine until an interrupt calls Release(). It will also
+// return if the interrupt called Release() before the Wait() call.
```
mikesmitty:
```suggestion
// Wait from a main goroutine until an interrupt calls Resume(). It will also
// return if the interrupt called Resume() before the Wait() call.
```
I assume Release() was renamed to Resume()?

mikesmitty (commented):
Aside from the little documentation fix LGTM

eliasnaur:
What's the difference between `Waiter` and a channel?

```go
var ch = make(chan struct{}, 1)

// In the waiting code
<-ch

// In the interrupt handler
select{
case ch<-struct{}{}:
default:
}
```

src/internal/task/waiter.go
aykevl: Oops, fixed.

aykevl:
@eliasnaur semantically there is very little difference. However, this waiter:

  * Is much more lightweight than a channel. It's going to be much more efficient (untested but that should be the case) and typically only uses 4 bytes of memory instead of 40 or so for a channel. Also it can be zero-initialized together with the rest of an object and doesn't need a separate heap allocation.
  * There is a separate "done" state, which doesn't exist for channels. Not entirely sure whether it is needed for USB MSC but it was used in the previous code.

The reason why it can be so much more lightweight is because it doesn't support `select`, multiple senders, or multiple receivers. Only one waiting goroutine and one interrupt handler is supported.

There have been many cases in the machine package etc where we would have wanted something like a Waiter but a channel would have been overkill.

mikesmitty:
> What's the difference between `Waiter` and a channel?

I actually tried to do a channel initially, but the channel got effectively optimized into a function call and sustained the interrupt context it was specifically trying to avoid

deadprogram:
> Unfortunately this suffers from a potential hang with `-scheduler=cores`: `scheduleTask` takes the scheduler lock from interrupt context which would cause a deadlock when this interrupt happens with the scheduler lock already held. I should try and fix this in a separate PR. So maybe it's best to hold off of merging this PR until that bug is fixed.

@aykevl any update on this?

aykevl:
No, haven't made a fix for that yet.

aykevl:
Actually, I think we can merge this as-is: https://github.com/tinygo-org/tinygo/commit/394116e99844f98507c52314713e632ed6761c98 disabled `-scheduler=cores` for all actually used targets by default. We can fix the race condition later, since that is a more general problem. (I suspect that the race condition I described above might be the same one that caused all the other issues and therefore fixing that might also help us re-enable the cores scheduler).

PR #5017, opened 2025-08-27
usb/msc: erase blocks before writing to them

aykevl:
We could perhaps check whether they've been erased already (via unmap or something) but the default of most OSes seems to be to _not_ unmap blocks so we will have to erase in most cases anyway.

Tested on MacOS, and tested similar changes on Linux. They allow a FAT filesystem to be mounted and modified without causing problems due to a missing erase step.

@mikesmitty I haven't looked deeply into USB MSC or SCSI, from the behavior of MacOS at least it appears that it will always write full blocks. Do you know whether this is true generally, or only happens to be true in this case?

### DIFF
--- a/src/machine/usb/msc/msc.go
+++ b/src/machine/usb/msc/msc.go
@@ -1,11 +1,13 @@
 package msc
 
 import (
+	"errors"
 	"machine"
 	"machine/usb"
 	"machine/usb/descriptor"
 	"machine/usb/msc/csw"
 	"machine/usb/msc/scsi"
+	"math/bits"
 	"time"
 )
 
@@ -22,6 +24,8 @@ const (
 const (
 	mscInterface = 2
 )
+
+var errInvalidBlockSize = errors.New("usb/msc: invalid block size")
 
 var MSC *msc
 
@@ -60,20 +64,41 @@ type msc struct {
 }
 
 // Port returns the USB Mass Storage port
-func Port(dev machine.BlockDevice) *msc {
+func Port(dev machine.BlockDevice) (*msc, error) {
 	if MSC == nil {
-		MSC = newMSC(dev)
+		msc, err := newMSC(dev)
+		if err != nil {
+			return nil, err
+		}
+		MSC = msc
 	}
-	return MSC
+	return MSC, nil
 }
 
-func newMSC(dev machine.BlockDevice) *msc {
+func newMSC(dev machine.BlockDevice) (*msc, error) {
 	// Size our buffer to match the maximum packet size of the IN endpoint
 	maxPacketSize := descriptor.EndpointMSCIN.GetMaxPacketSize()
+
+	// Windows only supports block sizes of 512 or 4096 bytes, other systems are
+	// probably similar.
+	blockSize := max(dev.EraseBlockSize(), dev.WriteBlockSize())
+	if bits.OnesCount32(uint32(blockSize)) != 1 {
+		return nil, errInvalidBlockSize // not a power of two
+	}
+	var blockSizeUSB uint32
+	switch {
+	case blockSize <= 512:
+		blockSizeUSB = 512
+	case blockSize <= 4096:
+		blockSizeUSB = 4096
+	default:
+		return nil, errInvalidBlockSize
+	}
+
 	m := &msc{
 		// Some platforms require reads/writes to be aligned to the full underlying hardware block
 		blockCache:    make([]byte, dev.WriteBlockSize()),
-		blockSizeUSB:  512,
+		blockSizeUSB:  blockSizeUSB,
 		buf:           make([]byte, dev.WriteBlockSize()),
 		cswBuf:        make([]byte, csw.MsgLen),
 		cbw:           &CBW{Data: make([]byte, 31)},
@@ -114,7 +139,7 @@ func newMSC(dev machine.BlockDevice) *msc {
 
 	go m.processTasks()
 
-	return m
+	return m, nil
 }
 
 func (m *msc) processTasks() {
--- a/src/machine/usb/msc/scsi_readwrite.go
+++ b/src/machine/usb/msc/scsi_readwrite.go
@@ -91,6 +91,16 @@ func (m *msc) writeBlock(b []byte, lba, offset uint32) (n int, err error) {
 		return 0, invalidWriteError
 	}
 
+	// Erase the block first if needed.
+	// Data packets arrive in order, so if we want to write to the start of a
+	// block, that means it's the first write to this erase block and it needs
+	// to be erased first.
+	if uint32(blockStart)%m.blockSizeUSB == 0 {
+		firstBlock := uint32(blockStart) / uint32(m.dev.EraseBlockSize())
+		numBlocks := (m.blockSizeUSB + uint32(m.dev.EraseBlockSize()) - 1) / uint32(m.dev.EraseBlockSize())
+		m.dev.EraseBlocks(int64(firstBlock), int64(numBlocks))
+	}
+
 	// Write the full block to the underlying device
 	n, err = m.dev.WriteAt(b, blockStart)
 	n -= int(blockOffset)


src/machine/usb/msc/msc.go
```diff
@@ -73,7 +73,7 @@ func newMSC(dev machine.BlockDevice) *msc {
 	m := &msc{
 		// Some platforms require reads/writes to be aligned to the full underlying hardware block
 		blockCache:    make([]byte, dev.WriteBlockSize()),
-		blockSizeUSB:  512,
+		blockSizeUSB:  max(uint32(dev.EraseBlockSize()), 512),
```
aykevl: Any idea why this was 512? If the block size is lower (128 for example), could this use a lower block size instead to reduce wear?

src/machine/usb/msc/scsi_readwrite.go
```diff
@@ -91,6 +91,13 @@ func (m *msc) writeBlock(b []byte, lba, offset uint32) (n int, err error) {
 		return 0, invalidWriteError
 	}
 
+	// Erase the block first if needed.
+	if uint32(blockStart)%m.blockSizeUSB == 0 {
+		firstBlock := uint32(blockStart) / uint32(m.dev.EraseBlockSize())
+		numBlocks := (m.blockSizeUSB + uint32(m.dev.EraseBlockSize()) - 1) / uint32(m.dev.EraseBlockSize())
+		m.dev.EraseBlocks(int64(firstBlock), int64(numBlocks))
+	}
```
aykevl, line 102: This makes the assumption that data is always written in full blocks, and that packets arrive sequentially. @mikesmitty can you confirm that this is true?

src/machine/usb/msc/msc.go
mikesmitty: The short version is that 512 bytes has been a near-universal standard sector size for disk storage for the last 40 years (and 4k for newer hardware). Linux won't even try to read a scsi disk with smaller sectors for example, and I doubt most other OSes will either. I would strongly recommend not changing this, higher or lower. It can only reduce hardware compatibility in exchange for the potential benefit of a small flash wear reduction. I would think it more important that a USB drive work universally than have marginally lower wear on a flash chip that should last hundreds of thousands of writes

src/machine/usb/msc/scsi_readwrite.go
mikesmitty: That's how the protocol operates, yeah. A read or write command is sent from the host in a CDB packet along with the transfer length and then nothing but un-encapsulated raw data is sent/received until the transfer length is met

mikesmitty:
> @mikesmitty I haven't looked deeply into USB MSC or SCSI, from the behavior of MacOS at least it appears that it will always write full blocks. Do you know whether this is true generally, or only happens to be true in this case?

Yeah, in the SCSI protocol read/write commands are specified by their starting block number and the length is given in number of full blocks. Transferring less than a full block would be expected to be discarded as incomplete.

src/machine/usb/msc/msc.go
mikesmitty: Yeah, windows doesn't support non-standard sector sizes either: https://learn.microsoft.com/en-us/troubleshoot/windows-server/backup-and-storage/support-policy-4k-sector-hard-drives

aykevl: Ok, will change to 512 or 4096 depending on the erase/write block size. Since most systems should support 4096 that seems like a relatively safe value, especially since the flash in microcontrollers seems to cap out at 4096 usually.

> I would think it more important that a USB drive work universally than have marginally lower wear on a flash chip that should last hundreds of thousands of writes

10.000 erase cycles per page (erase block) in the case of the nrf52 and stm32l031 chips for example. It's quite low actually. Normally a flash translation layer is used to spread wear out over the entire chip to not destroy it by writing to the FAT or main directory over and over again (which might make it wear out after doing just 10.000 file operations).
...but it's not something we can do a whole lot about in the USB MSC driver except for avoiding write amplification.

aykevl:
Updated the PR.

 * Use only 512 or 4096 byte blocks depending on the underlying block size. Modern systems should be capable of mounting devices with 4K blocks (Windows 8, Linux since 2010 or so, and probably any still-supported MacOS version but I couldn't find much information about that).
 * Improve comment before `EraseBlocks` call.

One thing I could change is to essentially add a new task specifically for erasing blocks, so that all blocks of a write operation are erased at once (this can be a bit faster than doing it block-by-block). But I guess the code as-is works fine.

> Yeah, in the SCSI protocol read/write commands are specified by their starting block number and the length is given in number of full blocks. Transferring less than a full block would be expected to be discarded as incomplete.

Right, so erasing at the start of the write would be safe? Or is the device expected to buffer the entire block and only write it after it received the entire block of data? (Probably a moot point since any proper OS will write full blocks anyway).

src/machine/usb/msc/msc.go
```diff
@@ -60,20 +64,41 @@ type msc struct {
...
+		blockSizeUSB = 4096
+	default:
+		return nil, errInvalidBlockSize
+	}
+
 	m := &msc{
 		// Some platforms require reads/writes to be aligned to the full underlying hardware block
 		blockCache:    make([]byte, dev.WriteBlockSize()),
-		blockSizeUSB:  512,
+		blockSizeUSB:  blockSizeUSB,
```
mikesmitty, line 101: It would probably be better to leave the default blockSizeUSB here to 512 and move the new block size checking logic to `RegisterBlockDevice()` in disk.go (called a few lines below here). I put the logic around setting up the block device in there with the idea that it could be swapped out during runtime if someone wanted to.

src/machine/usb/msc/scsi_readwrite.go
mikesmitty: Looking at this again, it might make sense to refactor the block erase logic like in #5027, then you could simplify this like so:
```
if offset == 0 && (lba % m.blockSizeUSB) == m.eraseBlockOffset {
    err := m.unmapUSBBlocks(lba, 1)
    // Error handling
}
```
writeBlock() gets called when a full flash device write block is buffered and `offset` here is that flash block's byte offset within the emulated USB sector, so it would depend on the usb block size being at least as large as the advertised usb block, but you've got that case covered now with the checks above

src/machine/usb/msc/msc.go
mikesmitty: Ah, fair enough. I have a lot of experience with traditional disks, but these flash chips less so

mikesmitty:
> One thing I could change is to essentially add a new task specifically for erasing blocks, so that all blocks of a write operation are erased at once (this can be a bit faster than doing it block-by-block). But I guess the code as-is works fine.

Yeah, I think that would be more ideal too

> Right, so erasing at the start of the write would be safe? Or is the device expected to buffer the entire block and only write it after it received the entire block of data? (Probably a moot point since any proper OS will write full blocks anyway).

As soon as the write starts happening it should be safe to erase the block, though it could be made a moot point by setting the `m.blockCache` buffer capacity to the emulated USB block size or at least the erase block size and then flash writes would automatically be buffered until the full USB or erase block is received. That would probably be for the best actually, to ensure data loss is avoided

src/machine/usb/msc/msc.go
mikesmitty: Also, sorry if I had a weird vibe earlier, got some life stuff going on right now and I'll be more chill going forward. I'm really impressed with what you've done with this project

deadprogram:
@aykevl @mikesmitty what is the status on this PR?

src/machine/usb/msc/msc.go
aykevl: You mean, you'd like to still emulate a block size of 512 even if the underlying flash has an erase block size of 4096?
I think that's a bad idea:
1. It takes up more space.
2. It causes write amplification (up to 8x as many erase operations - remember: flash has a finite amount of erase cycles!)
3. It seems that all modern operating systems support a block size of 4096. So I don't see the point of emulating a smaller block size.

mikesmitty: I was more thinking in the context that it'd be unusual for TinyGo to see a device with 4k erase blocks so 512 would be a safe default, but after the unexpected edge case that just bit me with the nrf52840 buffer thing, might as well plan to handle it better

aykevl: 4k is pretty common, it's what the nrf52* devices uses and IIRC also what rp2040 uses. And the code in this PR uses 512 bytes when possible anyway, so devices with a smaller erase block size can use that.

deadprogram:
Checking in on this PR @aykevl and @mikesmitty ? Thanks!

deadprogram:
Hello @aykevl or @mikesmitty or anyone else any update on this PR? :smile_cat:

src/machine/usb/msc/msc.go
```diff
@@ -60,20 +64,41 @@ type msc struct {
 }
 
 // Port returns the USB Mass Storage port
-func Port(dev machine.BlockDevice) *msc {
+func Port(dev machine.BlockDevice) (*msc, error) {
```
deadprogram, line 67: Since this is a breaking API change, just want to make sure everyone would be fine with that?

src/machine/usb/msc/scsi_readwrite.go
```diff
@@ -91,6 +91,16 @@ func (m *msc) writeBlock(b []byte, lba, offset uint32) (n int, err error) {
...
 	}
 
+	// Erase the block first if needed.
+	// Data packets arrive in order, so if we want to write to the start of a
+	// block, that means it's the first write to this erase block and it needs
+	// to be erased first.
+	if uint32(blockStart)%m.blockSizeUSB == 0 {
+		firstBlock := uint32(blockStart) / uint32(m.dev.EraseBlockSize())
+		numBlocks := (m.blockSizeUSB + uint32(m.dev.EraseBlockSize()) - 1) / uint32(m.dev.EraseBlockSize())
+		m.dev.EraseBlocks(int64(firstBlock), int64(numBlocks))
```
deadprogram, line 101: Shouldn't we return error from `EraseBlocks()` if there is one?

src/machine/usb/msc/scsi_readwrite.go
```diff
@@ -91,6 +91,16 @@ func (m *msc) writeBlock(b []byte, lba, offset uint32) (n int, err error) {
 		return 0, invalidWriteError
 	}
 
+	// Erase the block first if needed.
+	// Data packets arrive in order, so if we want to write to the start of a
+	// block, that means it's the first write to this erase block and it needs
+	// to be erased first.
+	if uint32(blockStart)%m.blockSizeUSB == 0 {
```
deadprogram, line 98: The blockStart used in the erase condition is a byte offset (from `usbToRawOffset`), but it's being divided by `EraseBlockSize()` which expects a block index, We should verify the math is consistent with how `EraseBlocks()` expects its parameters (block number vs byte offset)

PR #5023, opened 2025-09-02
chore: update version for 0.40 development cycle

sago35:


deadprogram:
Thanks @sago35 now merging.

PR #5025, opened 2025-09-05
fix: expand RTT debugger compatibility

mikesmitty:
I recently saw/fell in love with the idea of the https://github.com/piersfinlayson/airfrog debugger tool (wireless SWD+RTT) and got a bit jealous of the [probe-rs debugger](https://probe.rs/docs/tools/debugger/#a-visual-guide-of-implemented-features) features/tight integration with VSCode. Turns out it is largely powered by the https://github.com/Marus/cortex-debug extension, which was not originally written with rust in mind so I've been working on some things to make it easier to use (like an example launch.json for VSCode, etc.).

This PR in particular expands RTT debugger compatibility by adding a `_SEGGER_RTT` symbol alias to `machine.rttSerialInstance` for RTT debuggers (such as Marus/cortex-debug) that are hard-coded to look for the `_SEGGER_RTT` symbol specifically and struggle with memory scanning. probe-rs does [something similar-ish](https://github.com/probe-rs/rtt-target/blob/117d9519a5d3b1f4bc024bc05f9e3c5dec0a57f5/rtt-target/src/lib.rs#L262), but since [monitor.go](https://github.com/tinygo-org/tinygo/blob/3869f76887feef6c444308e7e1531b7cac1bbd10/monitor.go#L38) specifically looks for `machine.rttSerialInstance` we need both to exist to avoid breaking compatibility.

Example output from an nm dump:
```
$ arm-none-eabi-nm -S -l -p my-program.elf |egrep 'rttSerialInst|SEGGER'
20001068 00000048 b machine.rttSerialInstance   /Users/michaelsmith/git/tinygo/src/machine/serial-rtt.go:19
20001068 00000048 b _SEGGER_RTT
```

### DIFF
--- a/compiler/symbol.go
+++ b/compiler/symbol.go
@@ -630,7 +630,7 @@ func (c *compilerContext) addStandardAttributes(llvmFn llvm.Value) {
 // linkName is equal to .RelString(nil) on a global and extern is false, but for
 // some symbols this is different (due to //go:extern for example).
 type globalInfo struct {
-	linkName string // go:extern
+	linkName string // go:extern, go:linkname
 	extern   bool   // go:extern
 	align    int    // go:align
 	section  string // go:section
@@ -715,14 +715,14 @@ func (c *compilerContext) getGlobalInfo(g *ssa.Global) globalInfo {
 	// Check for //go: pragmas, which may change the link name (among others).
 	doc := c.astComments[info.linkName]
 	if doc != nil {
-		info.parsePragmas(doc)
+		info.parsePragmas(doc, c, g)
 	}
 	return info
 }
 
 // Parse //go: pragma comments from the source. In particular, it parses the
-// //go:extern pragma on globals.
-func (info *globalInfo) parsePragmas(doc *ast.CommentGroup) {
+// //go:extern and //go:linkname pragmas on globals.
+func (info *globalInfo) parsePragmas(doc *ast.CommentGroup, c *compilerContext, g *ssa.Global) {
 	for _, comment := range doc.List {
 		if !strings.HasPrefix(comment.Text, "//go:") {
 			continue
@@ -742,6 +742,17 @@ func (info *globalInfo) parsePragmas(doc *ast.CommentGroup) {
 		case "//go:section":
 			if len(parts) == 2 {
 				info.section = parts[1]
+			}
+		case "//go:linkname":
+			if len(parts) != 3 || parts[1] != g.Name() {
+				continue
+			}
+			// Only enable go:linkname when the package imports "unsafe".
+			// This is a slightly looser requirement than what gc uses: gc
+			// requires the file to import "unsafe", not the package as a
+			// whole.
+			if hasUnsafeImport(g.Pkg.Pkg) {
+				info.linkName = parts[2]
 			}
 		}
 	}
--- a/monitor.go
+++ b/monitor.go
@@ -35,8 +35,8 @@ func Monitor(executable, port string, config *compileopts.Config) error {
 		// Use the RTT interface, which is documented (in part) here:
 		// https://wiki.segger.com/RTT
 
-		// Try to find the "machine.rttSerialInstance" symbol, which is the RTT
-		// control block.
+		// Try to find the "_SEGGER_RTT" symbol (machine.rttSerialInstance)
+		// symbol, which is the RTT control block.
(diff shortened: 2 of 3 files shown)


mikesmitty:
Man, having a debugger to give you stack traces and live readouts of memory and registers is so nice, not having stack traces felt like living in the stone age. This is the minimal launch.json I'm working with right now (goes in `${workspaceFolder}/.vscode/`). I'm using a nightly build of openocd since v0.12.0 doesn't support rp2350 so some of the flags might be different with older versions. Also, it would be neat if we could add some of this into the TinyGo VSCode plugin like probe-rs does
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "TinyGo Debug",
            "type": "cortex-debug",
            "request": "launch",
            "servertype": "openocd",
            "executable": "./tinygo-build.elf",
            "configFiles": [
                "interface/cmsis-dap.cfg",
                "target/rp2350.cfg"
            ],
            "svdFile": "RaspberryPi::RP2xxx_DFP@0.9.5",
            "preLaunchTask": "${defaultBuildTask}",
            "openOCDLaunchCommands": [
                "adapter speed 10000"
            ],
            "rttConfig": {
                "enabled": true,
                "clearSearch": false,
                "polling_interval": 10,
                "decoders": [
                    {
                        "label": "RTT channel 0",
                        "port": 0,
                        "timestamp": true,
                        "type": "console"
                    }
                ]
            }
        }
    ]
}
```
And tasks.json I'm using to automatically build it before running the debugger:
```json
{
    "version": "2.0.0",
    "tasks": [
       {
            "label": "TinyGo Build",
            "type": "shell",
            "command": "${userHome}/git/tinygo/build/tinygo build -target=metro-rp2350 -serial=rtt -opt=1 -o ${workspaceFolder}/tinygo-build.elf",
            "group": {
                "kind": "build",
                "isDefault": true
            },
            "presentation": {
                "reveal": "always",
                "panel": "new",
                "close": true
            }
        }
    ]
}
```

aykevl:
Personally I don't care about backwards compatibility in this case. You'd only really use this while debugging, and would rebuild the binary anyway.

What about just renaming the global? That seems like a simpler solution to me than adding an alias. (If going this route, please also add a comment to machine.rttSerialInstance explaining that it is renamed in the compiler).

```diff
--- a/compiler/symbol.go
+++ b/compiler/symbol.go
@@ -712,6 +712,9 @@ func (c *compilerContext) getGlobalInfo(g *ssa.Global) globalInfo {
                // Pick the default linkName.
                linkName: g.RelString(nil),
        }
+       if info.linkName == "machine.rttSerialInstance" {
+               info.linkName = "_SEGGER_RTT"
+       }
        // Check for //go: pragmas, which may change the link name (among others).
        doc := c.astComments[info.linkName]
        if doc != nil {
```

(Ideally this would be done using `//go:linkname` but that hasn't been implemented for globals yet, though it could be).

mikesmitty:
Oh, sure, that works for me too. I'll look into the globals issue while I'm at it

mikesmitty:
I ended up chasing my tail for a bit thinking the debug info wasn't getting properly attached since it wasn't showing in the `nm` output, but it's definitely in the DWARF info. I guess it just looks for the name field and not the linkage name
```
0x0000f262:   DW_TAG_variable
                DW_AT_name	("machine.rttSerialInstance")
                DW_AT_type	(0x0000f18f "machine.rttSerial")
                DW_AT_external	(true)
                DW_AT_decl_file	("/Users/michaelsmith/git/tinygo/src/machine/serial-rtt.go")
                DW_AT_decl_line	(21)
                DW_AT_alignment	(4)
                DW_AT_location	(DW_OP_addr 0x20001068)
                DW_AT_linkage_name	("_SEGGER_RTT")
```

deadprogram:
@mikesmitty I switched this PR to `dev`. Can you please rebase it against the latest `dev` branch? Thanks!

mikesmitty:
Oh whoops, yeah one sec

deadprogram:
@mikesmitty you might also want to squash this PR into 2 commits, one for the RTT debugger, additions, the other for the `go:linkname` change. What do you think?

mikesmitty:
Yep, done

deadprogram:
Thank you @mikesmitty and to @aykevl for review. Now merging.

PR #5030, opened 2025-09-09
ci:uninstall cmake before install

luoliwoshang:
fixed  https://github.com/tinygo-org/tinygo/actions/runs/17543650117/job/49860578495?pr=5028

```bash
Updating files: 100% (146556/146556), done.
Error: cmake was installed from the local/pinned tap
but you are trying to install it from the homebrew/core tap.
Formulae with the same name from different taps cannot be installed at the same time.

To install this version, you must first uninstall the existing formula:
  brew uninstall cmake
Then you can install the desired version:
  brew install cmake

Error: Process completed with exit code 1.
```

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -79,6 +79,7 @@ jobs:
           rm -rf llvm-project
           make llvm-source
           # install dependencies
+          brew uninstall cmake || true
           HOMEBREW_NO_AUTO_UPDATE=1 brew install cmake ninja
           # build!
           make llvm-build


aykevl:
If cmake is already installed, I think we can skip that one? There's no reason we need a specific variant of it. (And it speeds up CI if we don't reinstall something).

deadprogram:
https://github.com/tinygo-org/tinygo/pull/5031 tries to do what you suggest @aykevl 

Let's see if it passes...

luoliwoshang:
The reason for uninstalling and then reinstalling cmake is that I want to maintain the original logic of installing the latest cmake via brew.

deadprogram:
> The reason for uninstalling and then reinstalling cmake is that I want to maintain the original logic of installing the latest cmake via brew.

That does make sense to me, especially since Homebrew does update formulas. Now merging, thanks for the fix @luoliwoshang!

aykevl:
> The reason for uninstalling and then reinstalling cmake is that I want to maintain the original logic of installing the latest cmake via brew.

But _why_ would you do that? We don't need the latest cmake.

luoliwoshang:
Sorry, I may have misread that. What I meant to express was to keep the original logic of installing cmake through brew, not necessarily the latest version. In fact, I also didn't install the latest cmake.

PR #5031, opened 2025-09-09
fix: do not install cmake on macOS CI build

deadprogram:
This PR  fixes the macOS CI by not installing cmake on macOS CI. It appears to be installed by default and since Homebrew has recently been updated, there is now a version mismatch.

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -79,7 +79,7 @@ jobs:
           rm -rf llvm-project
           make llvm-source
           # install dependencies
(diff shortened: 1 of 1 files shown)


aykevl:
The "Build LLVM" step is skipped, so the CI didn't actually test this change.

I would suggest updating the llvm-build-19 cache key for MacOS.

deadprogram:
> The "Build LLVM" step is skipped, so the CI didn't actually test this change.
> 
> I would suggest updating the llvm-build-19 cache key for MacOS.

Oops I merged #5030 before I read your comment @aykevl 

Too quick on the click. I will create a new PR to update the LLVM build.

PR #5032, opened 2025-09-09
GNUmakefile: shrink TinyGo binaries on Linux

aykevl:
With these flags, the TinyGo binary gets 17MB (10.6%) smaller. That seems like a quite useful win for such a small change!

This is only for Linux for now. MacOS and Windows can be tested later, the flags for those probably need to be modified.

Originally inspired by:
https://discourse.llvm.org/t/state-of-the-art-for-reducing-executable-size-with-heavily-optimized-program/87952/18 There are some other flags like -Wl,--pack-dyn-relocs=relr that did not shrink binary size in my testing, so I've left them out.

---

Of course the size of the compiler is not nearly as important as the size of the binaries it produces, but I still think this is a small win.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -26,7 +26,7 @@ jobs:
         # tar: needed for actions/cache@v4
         # git+openssh: needed for checkout (I think?)
         # ruby: needed to install fpm
-        run: apk add tar git openssh make g++ ruby-dev
+        run: apk add tar git openssh make g++ ruby-dev mold
       - name: Work around CVE-2022-24765
         # We're not on a multi-user machine, so this is safe.
         run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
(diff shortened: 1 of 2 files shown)


b0ch3nski:
@aykevl

```
/usr/bin/ld: unrecognized option '--icf=all'
```

shouldn't `lld` be called here instead of `ld`?

aykevl:
@b0ch3nski yes, though I've actually switched the PR over to mold instead since it produces even smaller binaries (no idea why). They're now almost 19MB smaller than what I started out with.

aykevl:
Removed the `mold` install for the cross compilation builds to try and fix an error. That will mean `-icf=all` won't work for those, but other size reductions still apply.

deadprogram:
Looking good, thanks for working on this @aykevl and to @b0ch3nski for review. Now merging.

PR #5033, opened 2025-09-09
fix: update macOS CI to build a new v2 cached LLVM to verify cmake install fix

deadprogram:
This PR updates the macOS CI to build a new v2 cached LLVM 19 to verify cmake install fix that was part of #5030


aykevl:
I'd still prefer if we didn't unnecessarily reinstall cmake. Slowing down CI because "this is how we've done it before" is not a very strong reason?

PR #5034, opened 2025-09-10
fix(rp2350): add software spinlocks

mikesmitty:
As it turns out, the RP2350 has hardware spinlocks that can be unlocked by writes to nearby addresses, the lower spinlocks currently in use in TinyGo happen to be unlocked by writes to the doorbell interrupt registers used to signal between cores, very possibly leading to some unexpected unlocks. This was not corrected in the A3 or A4 steppings and instead software spinlocks are used by default on RP2350 in pico-sdk: 
> RP2350 Warning. Due to erratum RP2350-E2, writes to new SIO registers above an offset of +0x180 alias the spinlocks, causing spurious lock releases. This SDK by default use atomic memory accesses to implement the hardware_sync_spin_lock API, as a workaround on RP2350 A2.

https://www.raspberrypi.com/documentation/pico-sdk/hardware.html#group_hardware_sync

Another thing I noticed in that section, they also always disable interrupts when the spinlocks are being held, we may want to do the same:
> [...] the default spinlock related methods here (e.g. [spin_lock_blocking](https://www.raspberrypi.com/documentation/pico-sdk/hardware.html#group_hardware_sync_1ga091e88a7608ed8dbf3e1e984bae54602)) always disable interrupts while the lock is held as use by IRQ handlers and user code is common/desirable, and spin locks are only expected to be held for brief periods.

These are the software spinlock macros ported over:
https://github.com/raspberrypi/pico-sdk/blob/2.2.0/src/rp2_common/hardware_sync_spin_lock/include/hardware/sync/spin_lock.h#L112
https://github.com/raspberrypi/pico-sdk/blob/2.2.0/src/rp2_common/hardware_sync_spin_lock/include/hardware/sync/spin_lock.h#L197

### DIFF
--- a/src/runtime/runtime_rp2.go
+++ b/src/runtime/runtime_rp2.go
@@ -13,7 +13,6 @@ import (
 )
 
 const numCPU = 2
-const numSpinlocks = 32
 
 // machineTicks is provided by package machine.
 func machineTicks() uint64
@@ -273,44 +272,6 @@ func coreStackTop(core uint32) uintptr {
 		runtimePanic("unexpected core")
 		return 0
 	}
-}
-
-// These spinlocks are needed by the runtime.
-var (
-	printLock     = spinLock{id: 20}
-	schedulerLock = spinLock{id: 21}
-	atomicsLock   = spinLock{id: 22}
-	futexLock     = spinLock{id: 23}
-)
-
-func resetSpinLocks() {
-	for i := uint8(0); i < numSpinlocks; i++ {
-		l := &spinLock{id: i}
-		l.spinlock().Set(0)
-	}
-}
-
-// A hardware spinlock, one of the 32 spinlocks defined in the SIO peripheral.
-type spinLock struct {
-	id uint8
-}
-
-// Return the spinlock register: rp.SIO.SPINLOCKx
-func (l *spinLock) spinlock() *volatile.Register32 {
-	return (*volatile.Register32)(unsafe.Add(unsafe.Pointer(&rp.SIO.SPINLOCK0), l.id*4))
-}
-
-func (l *spinLock) Lock() {
-	// Wait for the lock to be available.
-	spinlock := l.spinlock()
-	for spinlock.Get() == 0 {
-		arm.Asm("wfe")
-	}
-}
-
-func (l *spinLock) Unlock() {
-	l.spinlock().Set(0)
-	arm.Asm("sev")
 }
 
 // Wait until a signal is received, indicating that it can resume from the
--- a/src/runtime/runtime_rp2350.go
+++ b/src/runtime/runtime_rp2350.go
@@ -3,8 +3,10 @@
 package runtime
 
 import (
+	"device/arm"
 	"device/rp"
 	"runtime/interrupt"
+	"sync/atomic"
 )
 
 const (
@@ -36,4 +38,51 @@ func handleSIOFifoInterrupt(intr interrupt.Interrupt) {
 	case 1:
 		gcInterruptHandler(currentCPU())
 	}
+}
+
+// These spinlocks are needed by the runtime.
+//
+// Unlike on the RP2040, these are software spinlocks: erratum RP2350-E2 makes
+// the hardware SIO spinlocks unreliable (a core can observe a spinlock as
+// acquired when it isn't). The pico-sdk likewise replaced them with LL/SC
+// based software spinlocks on this chip. The Cortex-M33 supports ldrex/strex,
+// so 32-bit atomics compile to native instructions and are multicore-safe.
+var (
+	printLock     spinLock
+	schedulerLock spinLock
+	atomicsLock   spinLock
+	futexLock     spinLock
+)
+
+// Software spinlocks don't survive in locked state across a soft reset: they
+// live in .bss, which preinit() has already cleared before this is called.
+func resetSpinLocks() {
+}
+
+// A software spinlock, implemented using atomic operations.
+type spinLock struct {
+	atomic.Uint32
+}
+
+func (l *spinLock) Lock() {
+	// Try to replace 0 with 1. Once we succeed, the lock has been acquired.
+	for !l.Uint32.CompareAndSwap(0, 1) {
+		// Wait until the current holder calls Unlock, which executes a "sev".
+		arm.Asm("wfe")
+	}
+}
+
+func (l *spinLock) Unlock() {
+	// Safety check: the spinlock should have been locked.
+	if schedulerAsserts && l.Uint32.Load() != 1 {
+		runtimePanic("unlock of unlocked spinlock")
+	}
+
+	// Unlock the lock. Simply write 0, because we already know it is locked.
+	l.Uint32.Store(0)
+
+	// Wake up cores that are waiting for this lock in Lock() or for some other
+	// event (like a runnable task) in schedulerUnlockAndWait. The hardware
+	// spinlock version on the RP2040 has the same behavior.
+	arm.Asm("sev")
 }
--- a/src/runtime/runtime_rp2040.go
+++ b/src/runtime/runtime_rp2040.go
@@ -3,14 +3,57 @@
 package runtime
 
 import (
+	"device/arm"
 	"device/rp"
 	"runtime/interrupt"
+	"runtime/volatile"
+	"unsafe"
 )
 
 const (
 	sioIrqFifoProc0 = rp.IRQ_SIO_IRQ_PROC0
 	sioIrqFifoProc1 = rp.IRQ_SIO_IRQ_PROC1
 )
+
+const numSpinlocks = 32
+
+// These spinlocks are needed by the runtime.
+var (
+	printLock     = spinLock{id: 20}
+	schedulerLock = spinLock{id: 21}
+	atomicsLock   = spinLock{id: 22}
+	futexLock     = spinLock{id: 23}
+)
+
+func resetSpinLocks() {
+	for i := uint8(0); i < numSpinlocks; i++ {
+		l := &spinLock{id: i}
+		l.spinlock().Set(0)
+	}
+}
+
+// A hardware spinlock, one of the 32 spinlocks defined in the SIO peripheral.
+type spinLock struct {
+	id uint8
+}
+
+// Return the spinlock register: rp.SIO.SPINLOCKx
+func (l *spinLock) spinlock() *volatile.Register32 {
+	return (*volatile.Register32)(unsafe.Add(unsafe.Pointer(&rp.SIO.SPINLOCK0), l.id*4))
+}
+
+func (l *spinLock) Lock() {
+	// Wait for the lock to be available.
+	spinlock := l.spinlock()
+	for spinlock.Get() == 0 {
+		arm.Asm("wfe")
+	}
+}
+
+func (l *spinLock) Unlock() {
+	l.spinlock().Set(0)
+	arm.Asm("sev")
+}
 
 // On RP2040, each core has its own SIO FIFO IRQ. Core0 enables
 // IRQ_SIO_IRQ_PROC0 and Core1 enables IRQ_SIO_IRQ_PROC1, so each handler can


mikesmitty:
Oh, whoops. My go fmt extension has been flaking out on me. Will have the missing rp2040 imports updated in a moment. Here's the lock/unlock disassembled output with inlining disabled:
```
   0x10001178 <(*runtime.spinLock).Lock+0>:     cbz     r0, 0x10001192 <(*runtime.spinLock).Lock+26>
   0x1000117a <(*runtime.spinLock).Lock+2>:     ldaexb  r2, [r0]
   0x1000117e <(*runtime.spinLock).Lock+6>:     movs    r1, #1
   0x10001180 <(*runtime.spinLock).Lock+8>:     cmp     r2, #0
   0x10001182 <(*runtime.spinLock).Lock+10>:    bne.n   0x1000117a <(*runtime.spinLock).Lock+2>
   0x10001184 <(*runtime.spinLock).Lock+12>:    strexb  r2, r1, [r0]
   0x10001188 <(*runtime.spinLock).Lock+16>:    cmp     r2, #0
   0x1000118a <(*runtime.spinLock).Lock+18>:    bne.n   0x1000117a <(*runtime.spinLock).Lock+2>
   0x1000118c <(*runtime.spinLock).Lock+20>:    dmb     sy
   0x10001190 <(*runtime.spinLock).Lock+24>:    bx      lr
   0x10001192 <(*runtime.spinLock).Lock+26>:    bl      0x100013f4 <runtime.nilPanic>
```
```
   0x100013e4 <(*runtime.spinLock).Unlock+0>:   cbz     r0, 0x100013ee <(*runtime.spinLock).Unlock+10>
   0x100013e6 <(*runtime.spinLock).Unlock+2>:   movs    r1, #0
   0x100013e8 <(*runtime.spinLock).Unlock+4>:   stlb    r1, [r0]
   0x100013ec <(*runtime.spinLock).Unlock+8>:   bx      lr
   0x100013ee <(*runtime.spinLock).Unlock+10>:  bl      0x100013f4 <runtime.nilPanic>
```

src/runtime/runtime_rp2350.go
```diff
@@ -14,3 +15,60 @@ const (
 	sioIrqFifoProc0 = rp.IRQ_SIO_IRQ_FIFO
 	sioIrqFifoProc1 = rp.IRQ_SIO_IRQ_FIFO
 )
+
+// Due to hardware errata RP2350-E2 spinlocks are emulated in software as in pico-sdk.
+type spinLock struct {
+	// state field must be first field so its address is the same as the struct address.
+	state  uint8
+	id     uint8
+	uint16 // Padding to prevent false sharing
```
eliasnaur: Nit: `uint16` has its own alignment, so I think `_ [2]uint8` is more clear.

src/runtime/runtime_rp2350.go
```diff
@@ -14,3 +15,60 @@ const (
...
+
+func (l *spinLock) Lock() {
+	// Original reference:
+	// https://github.com/raspberrypi/pico-sdk/blob/2.2.0/src/rp2_common/hardware_sync_spin_lock/include/hardware/sync/spin_lock.h#L112
+
+	// r0 is automatically filled with the pointer value "l" here.
+	// We create a variable to permit access to the state byte (l.state) and
+	// avoid a memory fault when accessing it in assembly.
+	state := &l.state
+	_ = state
```
eliasnaur: Hoping `state` ends up in `r0` seems brittle to me, and I'm surprised the compiler doesn't optimize it away. Are you sure you can't bind `state` to an asm register a better way? https://tinygo.org/docs/concepts/compiler-internals/inline-assembly/ mentions that Cgo assembly is more full-featured and also inlined.

src/runtime/runtime_rp2350.go
```diff
@@ -14,3 +15,60 @@ const (
...
+	// Jump back to the loop start ("1:") if the lock is already held.
+	arm.Asm("bne 1b")
+
+	// Attempt to store '1' into the lock state byte.
+	// The return code (0 for success, 1 for failure) is placed in r2.
+	arm.Asm("strexb r2, r1, [r0]")
+	// Check if the result was successful (r2 == 0).
+	arm.Asm("cmp r2, #0")
+	// Jump back to the loop start ("1:") if the lock was not acquired.
+	arm.Asm("bne 1b")
```
eliasnaur: With register binding through Cgo assembly, it seems to me that the assembly can be cut down to just the special instructions (`ldaexb` and `strexb`) and the rest kept in Go.

eliasnaur (commented):
I support the switch to atomic instructions, but if you need something that works right away, RP2350-E2 mentions that some spinlocks are not affected:

> The following SIO spinlocks can be used normally because they don’t alias with writable registers: 5, 6, 7,
10, 11, and 18 through 31. Some of the other lock addresses may be used safely depending on which of
the high-addressed SIO registers are in use.
Locks 18 through 24 alias with some read-only TMDS encoder registers, which is safe as only writes are
mis-decoded.

src/runtime/runtime_rp2350.go
aykevl: Yeah, don't do this. This _will_ break eventually. The compiler is free to put it in any register it likes, store it on the stack, whatever.
Also, all the assembly instructions below are independent from a compiler POV so the compiler is free to modify registers between them if it wants to (it probably won't, but it would be allowed to).

A much better way would be to use atomic operations directly, and with that I mean sync/atomic. See the section I posted before:

https://github.com/tinygo-org/tinygo/blob/3869f76887feef6c444308e7e1531b7cac1bbd10/src/runtime/runtime_tinygoriscv_qemu.go#L360-L375

This _should_ result in similar assembly, and if it doesn't we'd have to investigate why.

aykevl (changes requested):
See my comment below.

Also,

> Another thing I noticed in that section, they also always disable interrupts when the spinlocks are being held, we may want to do the same: [...]

The runtime does this in various places if needed. The spinlock implementation doesn't need to disable interrupts too.

src/runtime/runtime_rp2350.go
```diff
@@ -14,3 +15,60 @@ const (
 	sioIrqFifoProc0 = rp.IRQ_SIO_IRQ_FIFO
 	sioIrqFifoProc1 = rp.IRQ_SIO_IRQ_FIFO
 )
+
+// Due to hardware errata RP2350-E2 spinlocks are emulated in software as in pico-sdk.
+type spinLock struct {
+	// state field must be first field so its address is the same as the struct address.
+	state  uint8
+	id     uint8
```
aykevl: Why is there an `id` field here? It doesn't seem to be used?

src/runtime/runtime_rp2350.go
```diff
@@ -14,3 +15,60 @@ const (
 	sioIrqFifoProc0 = rp.IRQ_SIO_IRQ_FIFO
 	sioIrqFifoProc1 = rp.IRQ_SIO_IRQ_FIFO
 )
+
+// Due to hardware errata RP2350-E2 spinlocks are emulated in software as in pico-sdk.
```
eliasnaur: Nit: I would remove this comment. It seems to me that even without RP2350-E2, atomic instructions are superior to hardware spinlocks. In other words, the purpose of hardware spinlocks is to enable multiple cores on the atomic-less M0+ architecture of the rp2040.

src/runtime/runtime_rp2350.go
mikesmitty: I agree it's questionable, but because the Go receiver variable is always passed as the first parameter and r0 is where the first parameter will always be per the AAPCS this should always be consistent. I won't die on that hill if atomics can produce a similar enough result though. I don't recall what method I was testing, but the atomic lock method I was initially looking at disassembled to about 4x as long as this, hence the hacky setup. Looking at that one though, it's only ~2x the size which seems reasonable to me.

Hacky version:
```
   0x10001178 <(*runtime.spinLock).Lock+0>:     cbz     r0, 0x10001192 <(*runtime.spinLock).Lock+26>
   0x1000117a <(*runtime.spinLock).Lock+2>:     ldaexb  r2, [r0]
   0x1000117e <(*runtime.spinLock).Lock+6>:     movs    r1, #1
   0x10001180 <(*runtime.spinLock).Lock+8>:     cmp     r2, #0
   0x10001182 <(*runtime.spinLock).Lock+10>:    bne.n   0x1000117a <(*runtime.spinLock).Lock+2>
   0x10001184 <(*runtime.spinLock).Lock+12>:    strexb  r2, r1, [r0]
   0x10001188 <(*runtime.spinLock).Lock+16>:    cmp     r2, #0
   0x1000118a <(*runtime.spinLock).Lock+18>:    bne.n   0x1000117a <(*runtime.spinLock).Lock+2>
   0x1000118c <(*runtime.spinLock).Lock+20>:    dmb     sy
   0x10001190 <(*runtime.spinLock).Lock+24>:    bx      lr
   0x10001192 <(*runtime.spinLock).Lock+26>:    bl      0x100013f4 <runtime.nilPanic>
```
Extensive version:
```
   0x10001178 <(*runtime.spinLock).Lock+0>:     cbz     r0, 0x100011d8 <(*runtime.spinLock).Lock+96>
   0x1000117a <(*runtime.spinLock).Lock+2>:     adds    r0, #4
   0x1000117c <(*runtime.spinLock).Lock+4>:     movs    r1, #1
   0x1000117e <(*runtime.spinLock).Lock+6>:     nop
   0x10001180 <(*runtime.spinLock).Lock+8>:     ldaex   r2, [r0]
   0x10001184 <(*runtime.spinLock).Lock+12>:    cbnz    r2, 0x10001190 <(*runtime.spinLock).Lock+24>
   0x10001186 <(*runtime.spinLock).Lock+14>:    stlex   r2, r1, [r0]
   0x1000118a <(*runtime.spinLock).Lock+18>:    cmp     r2, #0
   0x1000118c <(*runtime.spinLock).Lock+20>:    bne.n   0x10001180 <(*runtime.spinLock).Lock+8>
   0x1000118e <(*runtime.spinLock).Lock+22>:    b.n     0x100011d2 <(*runtime.spinLock).Lock+90>
   0x10001190 <(*runtime.spinLock).Lock+24>:    clrex
   0x10001194 <(*runtime.spinLock).Lock+28>:    ldaex   r2, [r0]
   0x10001198 <(*runtime.spinLock).Lock+32>:    cbnz    r2, 0x100011a4 <(*runtime.spinLock).Lock+44>
   0x1000119a <(*runtime.spinLock).Lock+34>:    stlex   r2, r1, [r0]
   0x1000119e <(*runtime.spinLock).Lock+38>:    cmp     r2, #0
   0x100011a0 <(*runtime.spinLock).Lock+40>:    bne.n   0x10001194 <(*runtime.spinLock).Lock+28>
   0x100011a2 <(*runtime.spinLock).Lock+42>:    b.n     0x100011d2 <(*runtime.spinLock).Lock+90>
   0x100011a4 <(*runtime.spinLock).Lock+44>:    clrex
   0x100011a8 <(*runtime.spinLock).Lock+48>:    ldaex   r2, [r0]
   0x100011ac <(*runtime.spinLock).Lock+52>:    cbnz    r2, 0x100011b8 <(*runtime.spinLock).Lock+64>
   0x100011ae <(*runtime.spinLock).Lock+54>:    stlex   r2, r1, [r0]
   0x100011b2 <(*runtime.spinLock).Lock+58>:    cmp     r2, #0
   0x100011b4 <(*runtime.spinLock).Lock+60>:    bne.n   0x100011a8 <(*runtime.spinLock).Lock+48>
   0x100011b6 <(*runtime.spinLock).Lock+62>:    b.n     0x100011d2 <(*runtime.spinLock).Lock+90>
   0x100011b8 <(*runtime.spinLock).Lock+64>:    clrex
   0x100011bc <(*runtime.spinLock).Lock+68>:    ldaex   r2, [r0]
   0x100011c0 <(*runtime.spinLock).Lock+72>:    cbnz    r2, 0x100011cc <(*runtime.spinLock).Lock+84>
   0x100011c2 <(*runtime.spinLock).Lock+74>:    stlex   r2, r1, [r0]
   0x100011c6 <(*runtime.spinLock).Lock+78>:    cmp     r2, #0
   0x100011c8 <(*runtime.spinLock).Lock+80>:    bne.n   0x100011bc <(*runtime.spinLock).Lock+68>
   0x100011ca <(*runtime.spinLock).Lock+82>:    b.n     0x100011d2 <(*runtime.spinLock).Lock+90>
   0x100011cc <(*runtime.spinLock).Lock+84>:    clrex
   0x100011d0 <(*runtime.spinLock).Lock+88>:    b.n     0x10001180 <(*runtime.spinLock).Lock+8>
   0x100011d2 <(*runtime.spinLock).Lock+90>:    dmb     sy
   0x100011d6 <(*runtime.spinLock).Lock+94>:    bx      lr
   0x100011d8 <(*runtime.spinLock).Lock+96>:    bl      0x10001438 <runtime.nilPanic>
```
That version:
```
   0x10001178 <(*runtime.spinLock).Lock+0>:     cbz     r0, 0x100011ae <(*runtime.spinLock).Lock+54>
   0x1000117a <(*runtime.spinLock).Lock+2>:     adds    r0, #4
   0x1000117c <(*runtime.spinLock).Lock+4>:     movs    r1, #1
   0x1000117e <(*runtime.spinLock).Lock+6>:     nop
   0x10001180 <(*runtime.spinLock).Lock+8>:     ldaex   r2, [r0]
   0x10001184 <(*runtime.spinLock).Lock+12>:    cbnz    r2, 0x10001192 <(*runtime.spinLock).Lock+26>
   0x10001186 <(*runtime.spinLock).Lock+14>:    stlex   r2, r1, [r0]
   0x1000118a <(*runtime.spinLock).Lock+18>:    cmp     r2, #0
   0x1000118c <(*runtime.spinLock).Lock+20>:    it      eq
   0x1000118e <(*runtime.spinLock).Lock+22>:    bxeq    lr
   0x10001190 <(*runtime.spinLock).Lock+24>:    b.n     0x10001180 <(*runtime.spinLock).Lock+8>
   0x10001192 <(*runtime.spinLock).Lock+26>:    movs    r1, #1
   0x10001194 <(*runtime.spinLock).Lock+28>:    clrex
   0x10001198 <(*runtime.spinLock).Lock+32>:    wfe
   0x1000119a <(*runtime.spinLock).Lock+34>:    nop
   0x1000119c <(*runtime.spinLock).Lock+36>:    ldaex   r2, [r0]
   0x100011a0 <(*runtime.spinLock).Lock+40>:    cmp     r2, #0
   0x100011a2 <(*runtime.spinLock).Lock+42>:    bne.n   0x10001194 <(*runtime.spinLock).Lock+28>
   0x100011a4 <(*runtime.spinLock).Lock+44>:    stlex   r2, r1, [r0]
   0x100011a8 <(*runtime.spinLock).Lock+48>:    cmp     r2, #0
   0x100011aa <(*runtime.spinLock).Lock+50>:    bne.n   0x1000119c <(*runtime.spinLock).Lock+36>
   0x100011ac <(*runtime.spinLock).Lock+52>:    bx      lr
   0x100011ae <(*runtime.spinLock).Lock+54>:    bl      0x10001414 <runtime.nilPanic>
   ```

File: src/runtime/runtime_rp2350.go (hunk 5 above)
Comment (mikesmitty) [reply src/runtime/runtime_rp2350.go]: Sure, seems reasonable

File: src/runtime/runtime_rp2350.go (hunk 4 above)
Comment (mikesmitty) [reply src/runtime/runtime_rp2350.go]: I originally just carried it over, but it's actually used for the hardware spinlocks in runtime_rp2.go:
```
/root/project/src/runtime/runtime_rp2.go:294:27: unknown field id in struct literal of type spinLock
/root/project/src/runtime/runtime_rp2.go:295:27: unknown field id in struct literal of type spinLock
/root/project/src/runtime/runtime_rp2.go:296:27: unknown field id in struct literal of type spinLock
/root/project/src/runtime/runtime_rp2.go:297:27: unknown field id in struct literal of type spinLock
```

File: src/runtime/runtime_rp2350.go (hunk 3 above)
Comment (mikesmitty) [reply src/runtime/runtime_rp2350.go]: Are you talking about the `arm.AsmFull()` functions? I originally tried that, but it doesn't allow passing through pointer types for some reason (has a note about having been removed in v0.23.0)
If you mean something else I'm curious though. Initial google results aren't bringing up much

**Comment by mikesmitty** [comment]
> I support the switch to atomic instructions, but if you need something that works right away, RP2350-E2 mentions that some spinlocks are not affected:
> 
> > The following SIO spinlocks can be used normally because they don’t alias with writable registers: 5, 6, 7,
> > 10, 11, and 18 through 31. Some of the other lock addresses may be used safely depending on which of
> > the high-addressed SIO registers are in use.
> > Locks 18 through 24 alias with some read-only TMDS encoder registers, which is safe as only writes are
> > mis-decoded.

Thank you for tracking those down. I figured there would probably be some, but I couldn't find them

File: src/runtime/runtime_rp2350.go (hunk 2 above)
Comment (mikesmitty) [reply src/runtime/runtime_rp2350.go]: I just noticed this one isn't doing a memory barrier at the end. I assume we'll want to add that to the rp2350 atomics, but I'm not sure where this implementation is coming from exactly. Is this generated from the arm assembly in the mainstream sync/atomic package?

File: src/runtime/runtime_rp2350.go (hunk 3 above)
Comment (eliasnaur) [reply src/runtime/runtime_rp2350.go]: I'm talking about "Inline assembly using CGo": https://tinygo.org/docs/concepts/compiler-internals/inline-assembly/

File: src/runtime/runtime_rp2350.go (hunk 2 above)
Comment (aykevl) [reply src/runtime/runtime_rp2350.go]: > I agree it's questionable, but because the Go receiver variable is always passed as the first parameter and r0 is where the first parameter will always be per the AAPCS this should always be consistent.

The AAPCS only applies on non-inlined externally available functions. That doesn't apply here. The compiler is free to inline these anywhere and use any register.

There are a few cases where you can rely on the calling convention, but this is not one of them.

File: src/runtime/runtime_rp2350.go (hunk 3 above)
Comment (aykevl) [reply src/runtime/runtime_rp2350.go]: Inline assembly through CGo is indeed an option, and can make the code slightly more efficient. I recommend reading this page to get an understanding of how it works: http://www.ethernut.de/en/documents/arm-inline-asm.html

Other than that, it's just standard CGo. You can make the function `static` and put it directly in the Go file like so:

```go
// static void spinlock_lock(unsigned *lock) { ... }
import "C"
```

src/runtime/runtime_rp2350.go
mikesmitty: Ah yeah, I didn't consider inlining, that could have been problematic.

aykevl: Even without inlining there is no reason the compiler would be required to keep `state` in any particular register. It might have picked any register, you're lucky it picked r0 in this case. In fact I had expected it would have optimized it out entirely.

src/runtime/runtime_rp2.go
```diff
@@ -297,31 +297,6 @@ var (
 	futexLock     = spinLock{id: 3}
```
eliasnaur: `id` is an implementation detail of rp2040 spinlocks, whereas on rp2350 `id`s have some meaning but are unusued. I suggest moving the spinlock variables to the rpXXXX.go files and avoid the `id` field on rp2350.

src/runtime/runtime_rp2350.go
```diff
@@ -14,3 +15,25 @@ const (
 	sioIrqFifoProc0 = rp.IRQ_SIO_IRQ_FIFO
 	sioIrqFifoProc1 = rp.IRQ_SIO_IRQ_FIFO
 )
+
+type spinLock struct {
+	atomic.Uint32
+	id uint8
```
eliasnaur: Delete field.

eliasnaur (approved):
Other than a nit, this looks good to me. @aykevl WDYT?

mikesmitty:
After doing some testing, I think this might not actually be locking properly. Going to do some more thorough testing

aykevl:
Yes, this looks good. @mikesmitty can you apply the changes proposed by @eliasnaur?

Also, you may want to rebase on the dev branch to hopefully fix the assert-test-linux CI failure.

src/runtime/runtime_rp2350.go
mikesmitty: That one's still needed for compatibility with how the rp2040 hardware spinlocks are initialized here:
https://github.com/tinygo-org/tinygo/blob/109e0767e763f1911828c5e59a563111e3c55f61/src/runtime/runtime_rp2.go#L295-L298

mikesmitty:
Sure, here's the rebase. I haven't had time to test or diagnose it yet, been stuck working on a project that's consumed all my free time, but I'm pretty sure it's not functioning properly as confirmed by that CI failure

mikesmitty:
Hmm, no I guess it is actually locking. I'm not sure what's happening with that CI test though

src/runtime/runtime_rp2350.go
eliasnaur: Indeed. Above, I suggested moving that `var ()` block to the `runtime_rp2xxx.go` files to get rid of that dependency.

mikesmitty: Ah yep, I missed that

aykevl (approved):
LGTM, thank you!

aykevl:
Restarting CI, these are probably flaky tests.

src/runtime/runtime_rp2350.go
```diff
@@ -2,8 +2,40 @@
 
 package runtime
 
+// #include <stdint.h>
+//
+// static void spinlock_lock(void *lock) {
```
eliasnaur, line 7: What was wrong with the `sync/atomic` operations? Custom assembly for just one target (rp2350) seems to me like a maintenance headache for little gain.

mikesmitty:
Ok, finally coming back to take a look at this again. I still have some reservations about the stability of this one. I'm getting mysterious panics in various runtime locations that I haven't tracked down yet. I suspect some race conditions are going on that I haven't narrowed down yet

For example:
```
[2025-10-03T13:28:06.865Z, +005191ms] panipanic: runtime  ee[2025-10-03T13:28:06.877Z, +005203ms] rrrroorr  aatt  0x100036ed0x10003a5d: unsafe.Slice/String: len out of range
```

src/runtime/runtime_rp2350.go
mikesmitty: I forget, but I'll take a second look at them. One reason was the size of them, but I think there was another reason also. In hindsight, the size difference is fairly small all things considered

src/runtime/runtime_rp2350.go
```diff
@@ -2,8 +2,40 @@
...
+//     "movs %0, #1\n" /* fill dependency slot */                 \
+//     "cmp %1, #0\n"                                             \
+//     /* Immediately retry if lock is seen to be taken */        \
+//     "bne 1b\n"                                                 \
+//     /* Attempt to claim */                                     \
+//     "strexb %1, %0, [%2]\n"                                    \
+//     "cmp %1, #0\n"                                             \
+//     /* Claim failed due to intervening write, so retry */      \
+//     "bne 1b\n"                                                 \
+//     : "=&r" (_tmp0), "=&r" (_tmp1) : "r" (lock)                \
```
aykevl, line 21: Just realized this is incomplete. You also need to clobber flags and memory at least.

src/runtime/runtime_rp2350.go
```diff
@@ -2,8 +2,40 @@
...
+//     );                                                         \
+//     __asm volatile ("dmb" : : : "memory");
+// }
+//
+// static void spinlock_unlock(void *lock) {
+//     /* Release-ordered store is available: use instead of separate fence */ \
+//     uint32_t zero = 0;                                                      \
+//     __asm volatile (                                                        \
+//         "stlb %0, [%1]\n"                                                   \
+//         : : "r" (zero), "r" (lock)                                          \
```
aykevl, line 31: Same here. It modifies memory, though I don't think it updates any flags.

aykevl (changes requested):
Just realized there are some issues with the assembly code.

deadprogram:
I tested against and this PR hangs multicore on two RP2350 boards where unpatched dev passes 200/200, reproducibly; the CAS primitive itself is fine, and neither interrupt masking nor ACTLR.EXTEXCLALL fixes it, so the root cause is still unknown. 

Meanwhile b9387fe already moved the locks clear of the doorbell aliases and #5482 fixed the deadlock in #5151, so this PRs original justification is largely gone.

I will open some new PRs and comments based on what we learned from all this @mikesmitty and everyone else who looked into it, so thank you very much!

PR #5035, opened 2025-09-10
fix: update chrome cdproto dependencies for CI

deadprogram:
This PR is to update chrome cdproto dependencies for CI to correct recent build errors.


deadprogram:
We cannot simply update these packages if we want to maintain backwards compatibility with older Go versions. :crying_cat_face:

b0ch3nski:
@deadprogram to keep Go 1.22 you can update `chromedp` to `v0.11.0` and `cdproto` to commit `cb3a7a1d51d76a913c548aa402970bbd616ffb6e` - just blamed these repos, anything newer will break compatibility :smile_cat:

deadprogram:
Just created #5038 to try to resolve the build error mentioned in #5036

PR #5039, opened 2025-09-11
chore: create separate go.mod file for testing dependencies for wasm tests

### LINKED ISSUE #5036: ci: Linux wasm tests failing due to Chrome
Once again, headless Chrome rears its head. CI is failing, probably due to some Chrome update.

For example:
```
target.go:126: could not unmarshal event: unknown IPAddressSpace value
```

https://github.com/tinygo-org/tinygo/actions/runs/17604358791/job/50012091228#step:21:59

I had previously corrected a similar issue with https://github.com/tinygo-org/tinygo/pull/4953

Others appear to be running into similar issues: https://github.com/chromedp/chromedp/issues/1578

deadprogram:
This PR is to create separate go.mod file for testing dependencies for wasm tests, since they use Chromium headless browser. This is to avoid use of older incompatible versions.

Fixes #5036


tests/wasm/go.mod
```diff
@@ -0,0 +1,17 @@
+module github.com/tinygo-org/tinygo/tests/wasm
+
+go 1.24
+
+require (
+	github.com/chromedp/cdproto v0.0.0-20250803210736-d308e07a266d
+	github.com/chromedp/chromedp v0.14.1
```
ysoldak, line 7: Why would you not use these newer versions directly in main go.mod?

deadprogram: https://github.com/tinygo-org/tinygo/pull/5035 tried that. Cannot do that and also maintain backward compatibility. :crying_cat_face:

ysoldak (approved):

deadprogram:
Thanks for review @ysoldak now merging.

PR #5041, opened 2025-09-16
fix(rp2): misc spinlock-related fixes

mikesmitty:
DBGPAUSE pauses the timer used for sleep pauses when a debugger is attached, which gets annoying to debug. Might be related to https://github.com/tinygo-org/tinygo/blob/v0.39.0/src/runtime/runtime_rp2.go#L314

### DIFF
--- a/src/machine/machine_rp2_timer.go
+++ b/src/machine/machine_rp2_timer.go
@@ -101,3 +101,15 @@ func (tmr *timerType) lightSleep(us uint64) {
 	// Disable interrupt
 	intr.Disable()
 }
+
+// setDbgPause sets whether this timer is paused when a debugger is connected.
+func (tmr *timerType) setDbgPause(enable bool) {
+	const bitPos = 1
+	const bitMask = 0b11
+	val := uint32(0b00)
+	if enable {
+		// Disable timer when debugger is connected to either core.
+		val = 0b11
+	}
+	tmr.dbgPause.ReplaceBits(val, bitMask, bitPos)
+}
--- a/src/runtime/runtime_rp2.go
+++ b/src/runtime/runtime_rp2.go
@@ -311,15 +311,13 @@ func (l *spinLock) Lock() {
 	// Wait for the lock to be available.
 	spinlock := l.spinlock()
 	for spinlock.Get() == 0 {
-		// TODO: use wfe and send an event when unlocking so the CPU can go to
-		// sleep while waiting for the lock.
-		// Unfortunately when doing that, time.Sleep() seems to hang somewhere.
-		// This needs some debugging to figure out.
+		arm.Asm("wfe")
 	}
 }
 
 func (l *spinLock) Unlock() {
 	l.spinlock().Set(0)
+	arm.Asm("sev")
 }
 
 // Wait until a signal is received, indicating that it can resume from the
--- a/src/runtime/runtime_rp2.go
+++ b/src/runtime/runtime_rp2.go
@@ -14,6 +14,7 @@ import (
 )
 
 const numCPU = 2
+const numSpinlocks = 32
 
 // machineTicks is provided by package machine.
 func machineTicks() uint64
@@ -297,6 +298,13 @@ var (
 	futexLock     = spinLock{id: 23}
 )
 
+func resetSpinLocks() {
+	for i := uint8(0); i < numSpinlocks; i++ {
+		l := &spinLock{id: i}
+		l.spinlock().Set(0)
+	}
+}
+
 // A hardware spinlock, one of the 32 spinlocks defined in the SIO peripheral.
 type spinLock struct {
 	id uint8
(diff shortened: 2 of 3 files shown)


src/machine/machine_rp2_timer.go
```diff
@@ -101,3 +101,14 @@ func (tmr *timerType) lightSleep(us uint64) {
 	// Disable interrupt
 	intr.Disable()
 }
+
+// setDbgPause sets whether this timer is paused when a debugger is connected.
+func (tmr *timerType) setDbgPause(enable bool) {
+	const bitPosCore0 = 1
+	const bitMask = 0b11
+	var val uint32
```
eliasnaur: Nit: explicitly initialize `val` to the disabled value (supposedly `val := 0b00`).

src/machine/machine_rp2_timer.go
```diff
@@ -101,3 +101,14 @@ func (tmr *timerType) lightSleep(us uint64) {
 	// Disable interrupt
 	intr.Disable()
 }
+
+// setDbgPause sets whether this timer is paused when a debugger is connected.
+func (tmr *timerType) setDbgPause(enable bool) {
+	const bitPosCore0 = 1
```
eliasnaur: What about core 1? Is the (main) timer always guaranteed to run on core 0? A comment would be nice.

mikesmitty: The timer is shared between both cores, this just configures whether or not the timer is paused when a debugger is attached to a given core (core0 is bit 1, core1 is bit 2, and bit 0 is reserved). Added a comment to help indicate

src/machine/machine_rp2_timer.go
```diff
@@ -101,3 +101,15 @@ func (tmr *timerType) lightSleep(us uint64) {
...
+// setDbgPause sets whether this timer is paused when a debugger is connected.
+func (tmr *timerType) setDbgPause(enable bool) {
+	const bitPos = 1
+	const bitMask = 0b11
+	val := uint32(0b00)
+	if enable {
+		// Disable timer when debugger is connected to either core.
+		val = 0b11
+	}
+	tmr.dbgPause.ReplaceBits(val, bitMask, bitPos)
```
eliasnaur, line 114: Nit: use constants from package `rp`:

```go
val := uint32(0)
mask := rp.WATCHDOG_CTRL_PAUSE_DBG0_Msk|
    rp.WATCHDOG_CTRL_PAUSE_DBG1_Msk
if enable {
    val = mask
}
tmr.dbgPause.ReplaceBits(val, mask, 0)
```

(untested)

eliasnaur (commented):
LGTM, but you probably want to change the PR title to reflect that you're fixing multiple issues.

src/machine/machine_rp2_timer.go
mikesmitty: Yeah, I'd prefer to use a mask constant, but that's for a different register with the same name unfortunately. They also switched from one timer on the RP2040 to two on RP2350 (one for secure code, one for non-secure) so the naming switched (TIMER_DBGPAUSE_DBG[01]_Msk vs TIMER[01]_DBGPAUSE_DBG[01]_Msk)

mikesmitty:
> LGTM, but you probably want to change the PR title to reflect that you're fixing multiple issues.

Ah yeah, good point. I ended up adding more things than I planned originally

mikesmitty:
I thought I was done, but here's one more. schedulerLock can stay held through a soft reset, especially if the reset happens while all tasks are sleeping, and since spinlocks don't get reset during a soft reset it will hang the scheduler before `runtime.init()` is called

src/runtime/runtime_rp2.go
```diff
@@ -291,12 +292,20 @@ func coreStackTop(core uint32) uintptr {
...
+	atomicsLock   = spinLock{id: 22}
+	futexLock     = spinLock{id: 23}
 )
 
+func resetSpinLocks() {
+	l := &spinLock{id: 0}
+	for i := uint8(0); i < numSpinlocks; i++ {
+		l.id = i
+		l.spinlock().Set(0)
+	}
```
eliasnaur: Nit:

```go
for i := range uint8(numSpinLocks) {
   l := &spinLock{id:i}
   l.spinlock().Set(0)
}
```
}

eliasnaur (commented):
LGTM

src/runtime/runtime_rp2.go
mikesmitty: Sure, that seems reasonable. One sec

eliasnaur: You didn't like my range-over-int suggestion? :)

deadprogram:
Thank you for the fixes here @mikesmitty and to @eliasnaur for review. Now merging.

PR #5042, opened 2025-09-16
LLVM 21 support

aykevl:
This is nowhere near finished, I've seen a few issues already. So it is marked as WIP until then.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,7 +108,7 @@ jobs:
       - image: golang:1.25-bullseye
     steps:
       - test-linux:
-          llvm: "20"
+          llvm: "21"
     resource_class: large
 
 workflows:
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -117,7 +117,7 @@ jobs:
     runs-on: macos-latest
     strategy:
       matrix:
-        version: [16, 17, 18, 19, 20]
+        version: [16, 17, 18, 19, 20, 21]
     steps:
       - name: Set up Homebrew
         uses: Homebrew/actions/setup-homebrew@master
@@ -141,8 +141,8 @@ jobs:
       - name: Check binary
         run: tinygo version
       - name: Build TinyGo (default LLVM)
-        if: matrix.version == 20
+        if: matrix.version == 21
         run: go install
       - name: Check binary
-        if: matrix.version == 20
+        if: matrix.version == 21
         run: tinygo version
--- a/.github/workflows/sizediff-install-pkgs.sh
+++ b/.github/workflows/sizediff-install-pkgs.sh
@@ -2,11 +2,11 @@
 # still works after checking out the dev branch (that is, when going from LLVM
 # 16 to LLVM 17 for example, both Clang 16 and Clang 17 are installed).
 
-echo 'deb https://apt.llvm.org/noble/ llvm-toolchain-noble-20 main' | sudo tee /etc/apt/sources.list.d/llvm.list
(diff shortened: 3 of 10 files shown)


aykevl:
There is a rather tricky bug in WebAssembly. It looks like it has something to do with concurrency (`-scheduler=none` hides the bug, but inserting the right `println` lines also hides the bug so :woman_shrugging:). This is going to be fun to debug.

@dgryski is this something you'd want to take a look at? Reproducing it is straightforward: take this branch, do `go install` on a system with LLVM 21, and running `tinygo run -target=wasip1 ./testdata/gc.go` (it crashes on exit).

dgryski:
```
# llvm 21
~/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/dgryski/go-nibz $ tinygo build -print-allocs=startMark -tags=runtime_asserts -target=wasip1 -scheduler=none .
/Users/dgryski/go/src/github.com/tinygo-org/tinygo/src/runtime/gc_blocks.go:575:3: object allocated on the heap: escapes at line 591

# llvm 20
~/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/dgryski/go-nibz $ tinygo build -print-allocs=startMark -tags=runtime_asserts -target=wasip1 -scheduler=none .
~/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/dgryski/go-nibz $
```

Something in the escape analysis pass is causing what *should* be a stack allocation to be moved to the heap and thus call `runtime.alloc` in the middle of our garbage collection, with predictable results.

dgryski:
The escape analysis pass needs to be updated to use the updated `captures` attribute: https://github.com/tinygo-org/tinygo/blob/release/transform/allocs.go#L157 instead of `nocapture`.  (Note that we'll need to thread through the value of this attribute from the compiler where it's set.)

dgryski:
```
~/go/src/github.com/dgryski/tinygo-test-corpus/_corpus/golang/crypto/nacl/sign $ tinygo test -target=wasip1 -gc=leaking  -tags="purego noasm" -scheduler=none
--- FAIL: TestSign (0.01s)
    signed message did not match, got
        00000000000000000000000000000000000000000000000000000000000000004961dfea4f9435ddfba291b4ef883b9f9e3a29cc7120b5b8ef1e1ce1a303ca0368656c6c6f20776f726c64
        , expected
        26a0a47f733d02ddb74589b6cbd6f64a7dab1947db79395a1a9e00e4c902c0f185b119897b89b248d16bab4ea781b5a3798d25c2984aec833dddab57e0891e0d68656c6c6f20776f726c64
```

dgryski:
This failure is still `nocapture`-related.

dgryski:
I think the switch away from `nocapture` requires moving to a new attribute API that requires changes in `go-llvm` too.

aykevl:
> The escape analysis pass needs to be updated to use the updated `captures` attribute: https://github.com/tinygo-org/tinygo/blob/release/transform/allocs.go#L157 instead of `nocapture`. (Note that we'll need to thread through the value of this attribute from the compiler where it's set.)

I don't think this needs anything new in go-llvm. What's needed is reading the `captures` flag (using [`GetEnumValue`](https://pkg.go.dev/tinygo.org/x/go-llvm#Attribute.GetEnumValue), probably), and checking whether the right bits are set/clear. (The bits are somewhere in the LLVM source as an enum I think).

theoparis:
LLVM 22 is almost out - is this still being worked on?

dgryski:
Fixed in https://github.com/tinygo-org/tinygo/pull/5505

PR #5044, opened 2025-09-18
feat(os): add UserConfigDir and UserCacheDir

### LINKED ISSUE #5040: os.UserConfigDir and os.UserCacheDir are missing from os package
The tinygo standard library doesn't include [os.UserConfigDir](https://cs.opensource.google/go/go/+/refs/tags/go1.25.1:src/os/file.go;l=560) and [os.UserCacheDir](https://cs.opensource.google/go/go/+/refs/tags/go1.25.1:src/os/file.go;l=507}. This seems like a simple oversight given that `os.UserHomeDir` is implemented.

Minimal example:

```go
package main

import "os"

func main() {
	_, _ = os.UserConfigDir()
}
```

```
$ tinygo build -o main main.go
# command-line-arguments
main.go:6:12: undefined: os.UserConfigDir
```

mikesmitty:
Fixes #5040

### DIFF
--- a/src/internal/filepathlite/path_plan9.go
+++ b/src/internal/filepathlite/path_plan9.go
@@ -1,0 +1,6 @@
+package filepathlite
+
+// IsAbs reports whether the path is absolute.
+func IsAbs(path string) bool {
+	return len(path) > 0 && (path[0] == '/' || path[0] == '#')
+}
--- a/src/internal/filepathlite/path_unix.go
+++ b/src/internal/filepathlite/path_unix.go
@@ -1,0 +1,8 @@
+//go:build unix || (js && wasm) || wasip1
+
+package filepathlite
+
+// IsAbs reports whether the path is absolute.
+func IsAbs(path string) bool {
+	return len(path) > 0 && path[0] == '/'
+}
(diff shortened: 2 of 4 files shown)


deadprogram:
@mikesmitty looks like this needs to be behind a build tag since those functions do not exist in older Go versions.

aykevl:
internal/filepathlite was added in Go 1.23, so we can't rely on it.

We could, however, have our own internal/filepathlite package instead. For that you'd need to modify the following code to add the package as well:
https://github.com/tinygo-org/tinygo/blob/3869f76887feef6c444308e7e1531b7cac1bbd10/loader/goroot.go#L245
(Add `"internal/filepathlite": false,`, and preferably also a note to the files like "This is needed for Go 1.22 compatibility").

dgryski:
We'd have to import the entire `internal/filepathlite` then since the upstream `path/filepath` (which we use unmodified) was rewritten to use it for most functions.

b0ch3nski:
... or after 1.26 release we could drop 1.22 support :smile_cat:

aykevl:
Can you rebase on the current dev branch? We've dropped Go 1.22 support.

dgryski:
Closed as https://github.com/tinygo-org/tinygo/pull/5303 has been merged.

PR #5046, opened 2025-09-18
net: update submodule to latest commits

b0ch3nski:
Update submodule to include https://github.com/tinygo-org/net/pull/41


deadprogram:
Thanks for updating @b0ch3nski now merging.

PR #5048, opened 2025-09-18
fix: point the submodule for musl-libc to TinyGo org mirror

deadprogram:
This PR fixes the intermittent CI errors by pointing the submodule for musl-libc to a mirror in the TinyGo GitHub org. The git server for git.musl-libc.org is having troubles, and it also seems like a safer bet to have our own mirror just in case.


deadprogram:
Note that all tests pass except macOS which require PR #5049

cee-dub (approved):

deadprogram:
Handled in #5049

PR #5049, opened 2025-09-18
fix: remove macOS 15 from CI build matrix. It was already being teste…

deadprogram:
This PR fixes a build failure remove macOS 15 from CI build matrix. It is already being tested by the `test-macos-homebrew` job, and the build artifact conflicts with the actual build artifact that we want, which is built on macOS 14 for backwards-compatibility.


deadprogram:
Build is failing due to git.musl-libc.org server not accepting connections. This is addressed by #5048

deadprogram:
I have added the commit from #5048 so we can see that everything should pass with both PRs.

cee-dub (approved):

deadprogram:
Thanks for reviewing @cee-dub

PR #5051, opened 2025-09-19
Added Gopher ARCADE board

conejoninja:
Gopher ARCADE is a custom small board based on the attiny85, more information in this unfinished repository : https://github.com/conejoninja/arcade


deadprogram:
Very cool @conejoninja :joystick: 

Now merging.

PR #4952, opened 2025-07-10
machine: add I2C pin comments

aykevl:
Similar to PWM, I2C can only be used on some pins. To automatically generate this information per board, we need to add extra comments that can then be interpreted by doc-gen for the tinygo.org website.

I will make a PR to tinygo-site later, that uses this information.

PR #5053, opened 2025-09-23
fe310: add I2C pins for the HiFive1b

aykevl:
This extends https://github.com/tinygo-org/tinygo/pull/4952 with some missing pins.


deadprogram:
Thanks @aykevl now merging.

Issue #4959, opened 2025-07-17
wasm-ld: error: couldn't allocate output register for constraint '{r0}'


anshuman-mor:
Getting this error intermittently when compiling to wasip2

> tinygo build -o test-tinygo-wasi2.wasm --target wasip2 --no-debug

wasm-ld: error: couldn't allocate output register for constraint '{r0}'

I am running this on Windows.

> tinygo version 
tinygo version 0.38.0 windows/amd64 (using go version go1.23.2 and LLVM version 19.1.2)

deadprogram:
Hello @anshuman-mor could you please provide some code that reproduces this error? Thank you.

joelnpdr:
Hello, I have the same issue.  I am compiling this library, which has some crypto operations:

https://github.com/gagliardetto/solana-go

I am using the same version as @anshuman-mor 


```go
package main

import sgo "github.com/gagliardetto/solana-go"

func main() {
        _, err := sgo.NewRandomPrivateKey()
        if err != nil {
                panic(err)
        }
        time.Sleep(50 * time.Second)
}
```

I am trying to generate a ed25519 key pair.

joelnpdr:
I see that for wasip2, tinygo pulls in ARM code:

https://github.com/tinygo-org/tinygo/blob/release/targets/wasip2.json#L8

aykevl:
Oh, huh. Somehow this program is calling syscalls, which of course won't work on wasip2. I wonder how that happened.

aykevl:
I traced the calls a little bit:

* `github.com/streamingfast/logging.maybeNewLogger`
* `golang.org/x/crypto/ssh/terminal.IsTerminal`
* `golang.org/x/term.IsTerminal`
* `golang.org/x/term.isTerminal`
* `golang.org/x/sys/unix.IoctlGetTermios`
* `golang.org/x/sys/unix.ioctl`

So yeah this all makes sense, considering we emulate linux/arm for wasip2 (since upstream doesn't support wasip2 in the stdlib). But the compiler should not have emitted these syscalls at least.

aykevl:
Ok, here you go: https://github.com/tinygo-org/tinygo/pull/5054
Not a fix, but a better error message.

[...]

PR #5054, opened 2025-09-24
compiler: emit an error when the actual arch doesn't match GOARCH

aykevl:
See: https://github.com/tinygo-org/tinygo/issues/4959
This makes sure to emit an error instead of generating inline assembly, which leads to hard-to-debug linker errors. Instead, it will now produce errors like the following:

    # golang.org/x/sys/unix
    ../../../../pkg/mod/golang.org/x/sys@v0.0.0-20220722155257-8c9f86f7a55f/unix/affinity_linux.go:20:23: system calls are not supported: target emulates a linux/arm system on wasm32
    ../../../../pkg/mod/golang.org/x/sys@v0.0.0-20220722155257-8c9f86f7a55f/unix/fcntl.go:17:29: system calls are not supported: target emulates a linux/arm system on wasm32
    ../../../../pkg/mod/golang.org/x/sys@v0.0.0-20220722155257-8c9f86f7a55f/unix/fcntl.go:32:24: system calls are not supported: target emulates a linux/arm system on wasm32

@dgryski you may want to verify this doesn't break anything, since a package might have been previously working. I can adapt the PR if these new error messages break existing code.

### DIFF
--- a/compiler/syscall.go
+++ b/compiler/syscall.go
@@ -74,6 +74,14 @@ func (b *builder) createRawSyscall(call *ssa.CallCommon) (llvm.Value, error) {
 		return b.CreateCall(fnType, target, args, ""), nil
 
 	case b.GOARCH == "arm" && b.GOOS == "linux":
+		if arch := b.archFamily(); arch != "arm" {
+			// Some targets pretend to be linux/arm for compatibility but aren't
+			// actually such a system. Make sure we emit an error instead of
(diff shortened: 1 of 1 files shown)


deadprogram:
@dgryski any feedback on this PR before merge? Thanks!

dgryski:
Test corpus passes on wasip2 and wasip1 so I guess we're good to go?

deadprogram:

PR #5055, opened 2025-10-01
interp: better errors when debugging interp

aykevl:
When debugging is enabled for interp, print better errors in a specific case.

Before:

    !! revert because of error: interp: unsupported instruction (to be emitted at runtime)

After:

    !! revert because of error: /usr/local/go1.24.0/src/regexp/syntax/parse.go:927:27: interp: unsupported instruction (to be emitted at runtime)

So this adds error location information, which can be quite useful.


deadprogram (approved):
Thanks for the improvement @aykevl

Issue #3356, opened 2022-12-27
Endless loop in runtime_nrf.go: sleepTicks()


temnok:
Endless loop in `runtime_nrf.go: sleepTicks()` function can happen for parameter `d` >= 0x800000 (>= 256 seconds for 32768Hz ticks) because `& 0x7fffff` operation will return 0 for second and following loop runs and `d` variable remains unchanged:

Current version:
```
func sleepTicks(d timeUnit) {
	for d != 0 {
		ticks := uint32(d) & 0x7fffff // 23 bits (to be on the safe side)
		rtc_sleep(ticks)
		d -= timeUnit(ticks)
	}
}
```

Fixed version:
```
func sleepTicks(d timeUnit) {
	for d != 0 {
		ticks := uint32(d)
		if ticks > 0x7fffff { // 23 bits (to be on the safe side)
			ticks = 0x7fffff
		}
		rtc_sleep(ticks)
		d -= timeUnit(ticks)
	}
}
```

aykevl:
Ugh, you are correct (I wrote this code). And I think there is a lot more code like this that is broken in the same way.
Feel free to make a PR :)

deadprogram:
@temnok @aykevl is this issue still valid?

temnok:
@deadprogram Yes that's still an issue since the code did not change at https://github.com/tinygo-org/tinygo/blob/release/src/runtime/runtime_nrf.go#L80

aykevl:
@temnok can you check whether https://github.com/tinygo-org/tinygo/pull/5056 works for you?

(Note that the fix you proposed is also subtly wrong, since it would still ignore all the bits beyond the lower 32 - though at a 32768Hz clock that would be something like 1.5 day so less likely to hit in practice).

PR #5056, opened 2025-10-01
runtime: fix sleep duration for long sleeps

aykevl:
Long sleep durations weren't implemented correctly, by simply casting to a smaller number of bits. What we want is a saturating downcast, which is what I've used here instead.

This should fix https://github.com/tinygo-org/tinygo/issues/3356.

---

Tested blinky on nrf52840 and samd21, didn't test any other board/chip.

### DIFF
--- a/builder/sizes_test.go
+++ b/builder/sizes_test.go
@@ -43,8 +43,8 @@ func TestBinarySize(t *testing.T) {
 	tests := []sizeTest{
 		// microcontrollers
 		{"hifive1b", "examples/echo", 3884, 280, 0, 2268},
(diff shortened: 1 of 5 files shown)


eliasnaur (approved):

deadprogram:
@aykevl please note small code size change is causing the test to fail: https://github.com/tinygo-org/tinygo/actions/runs/18158535177/job/51767081892?pr=5056#step:17:11

aykevl:
Binary sizes test should be fixed now.

Looking at the sizediff comment above, it does on average increase binary size slightly, but it also fixes a real bug so :woman_shrugging:

deadprogram:

PR #5058, opened 2025-10-01
Add `testing.T.Context()` and `testing.B.Context()`

ben-krieger:
Matches https://pkg.go.dev/testing#T.Context and https://pkg.go.dev/testing#B.Context

### DIFF
--- a/src/testing/testing.go
+++ b/src/testing/testing.go
@@ -10,6 +10,7 @@ package testing
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"flag"
 	"fmt"
@@ -78,6 +79,9 @@ type common struct {
 	tempDir    string
 	tempDirErr error
 	tempDirSeq int32
+
+	ctx       context.Context
+	cancelCtx context.CancelFunc
 }
 
 type logger struct {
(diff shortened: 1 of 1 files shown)


src/testing/testing.go
```diff
@@ -488,12 +505,15 @@ func (t *T) Run(name string, f func(t *T)) bool {
 	}
 
 	// Create a subtest.
+	ctx, cancelCtx := context.WithCancel(context.Background())
```
ben-krieger, line 508: It seems kinda weird that the parent context is not from the parent `testing.T`, but this seems to be how it is in the Go stdlib.

ben-krieger: Oh wait, there's specifically a comment on this in the Go stdlib:

> There's no reason to inherit this context from parent. The user's code can't observe the difference between the background context and the one from the parent test.

aykevl:
As an alternative, it might be possible to switch to the standard library `testing` package so we don't need to play catch-up with Go. I've tried it in the past, and I don't think it needs many changes to work. I might look into that later.

ben-krieger:
> As an alternative, it might be possible to switch to the standard library `testing` package so we don't need to play catch-up with Go. I've tried it in the past, and I don't think it needs many changes to work. I might look into that later.

@aykevl I absolutely agree with this approach! I'll even try to implement it myself. In the meantime, is it okay to merge this PR, since it's pretty small in scope and doesn't conflict with eventually changing to the upstream testing package?

deadprogram (approved):
Switching to upstream Go testing is the way, but this at least solves the immediate needs while that other work continues.

deadprogram:
Now merging, thanks for working on this @ben-krieger

PR #5059, opened 2025-10-03
fix ticker stop race condition by clearing period before removing timer

cataggar:
The job failed during the TestBuild/ARM64Linux/timers.go test. The log says: "fail: ticker should have stopped!" and "ticker was stopped (didn't send anything after 750ms)", indicating that the test expected the ticker to stop, but it did not behave as intended.

I was hitting this error in CI, so I [asked ](https://github.com/cataggar/tinygo/issues/4)Copilot to fix it and the fix looks good to me. At least the tests now pass.

### DIFF
--- a/src/runtime/time_go122.go
+++ b/src/runtime/time_go122.go
@@ -53,6 +53,11 @@ func startTimer(tim *timer) {
 
 //go:linkname stopTimer time.stopTimer
 func stopTimer(tim *timer) bool {
+	// Set period to 0 to prevent the timer from being re-added to the queue
+	// if the callback is currently running. This must be done before removing
+	// the timer to avoid a race condition where the callback re-adds the timer
+	// after it's been removed.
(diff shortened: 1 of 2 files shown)


aykevl:
This is a race condition. So it's not surprising it passes now, since the test passes most of the time.

Can you please explain why this is a correct fix (without using Copilot)?

cataggar:
@aykevl , unfortunately no. My Go skills aren't good enough to validate Copilot's reasoning:

> 	// Set period to 0 to prevent the timer from being re-added to the queue
	// if the callback is currently running. This must be done before removing
	// the timer to avoid a race condition where the callback re-adds the timer
	// after it's been removed.
	
If it does not make sense, please close. I just used some tokens trying to help out.

deadprogram:
Closing, thanks anyhow.

PR #5060, opened 2025-10-04
Add UEF as a target

sparques:


aykevl:
For this PR, I have the same reservations as for #3996:

  * It's very big, and therefore hard to review. I'd prefer to start with something that is minimal, to which we can add more over time.
  * I'd prefer if this was written like a baremetal target, and not something separate. Baremetal targets (that set the `baremetal` build tag etc) are easier to maintain since they can be treated similar to a microcontroller from a runtime/stdlib perspective.

deadprogram:
@sparques I think this PR is no longer needed due to your other PRs, is that correct?

sparques:
Correct, this is obsolete now; closing.

PR #5061, opened 2025-10-05
testing: switch to upstream version

aykevl:
This is a work in progress.

We currently have our own testing package, and while it works, ideally we'd use the version in the standard library. This PR is a start on that: `make tinygo-test` passes, but `make tinygo-test-baremetal` for example doesn't. More changes will be needed.

An example PR that wouldn't be needed anymore if we used the Go standard library testing package: https://github.com/tinygo-org/tinygo/pull/5058


aykevl:
Oh, huh. This fails because `AllocsPerRun` actually works (at least on wasip1) and our GC escape analysis isn't as good as the escape analysis in upstream Go. So tests start failing that passed before, but what changed is that `AllocsPerRun` actually checks allocations.

Not sure what to do here. We could modify `runtime.ReadMemStats` to run fake numbers during testing, but that doesn't seem like a clean solution. Or, maybe, try to improve the GC so that these tests start passing.

b0ch3nski:
@aykevl also `CheckCorpus()` needs to be added: https://github.com/tinygo-org/tinygo/blob/testing-upstream/testdata/testing.go#L94

```
matchStringOnly does not implement testing.testDeps (missing method CheckCorpus)
```

aykevl:
@b0ch3nski yes, there are a few of such things that needs changes. But those aren't blockers - they're just work. I was more looking at fundamental problems we have with the testing package.

PR #5062, opened 2025-10-05
Create "pico2-ice" target board

tinkerator:
This board has an rp2350b chip on it, as well as a Lattice Semiconductor iCE40UP5K FPGA. More details of this open hardware board here:

  https://pico2-ice.tinyvision.ai/

Tested on a pico2-ice board with:

  ~/go/bin/tinygo flash -target=pico2-ice src/examples/blinky1/blinky1.go

which blinks the GREEN LED (connected to GPIO0) on this board.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -621,7 +621,7 @@ smoketest: testchdir
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=feather-rp2040      examples/device-id
 	@$(MD5SUM) test.hex
-	$(TINYGO) build -size short -o test.hex -target=pico2-ice	    examples/blinky1
+	$(TINYGO) build -size short -o test.hex -target=pico2-ice	        examples/blinky1
 	@$(MD5SUM) test.hex
 	# test simulated boards on play.tinygo.org
 ifneq ($(WASM), 0)
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -621,6 +621,8 @@ smoketest: testchdir
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=feather-rp2040      examples/device-id
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=pico2-ice	    examples/blinky1
+	@$(MD5SUM) test.hex
 	# test simulated boards on play.tinygo.org
 ifneq ($(WASM), 0)
 	GOOS=js GOARCH=wasm $(TINYGO) build -size short -o test.wasm -tags=arduino              examples/blinky1
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -621,7 +621,7 @@ smoketest: testchdir
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=feather-rp2040      examples/device-id
 	@$(MD5SUM) test.hex
-	$(TINYGO) build -size short -o test.hex -target=pico2-ice	        examples/blinky1
+	$(TINYGO) build -size short -o test.hex -target=pico2-ice           examples/blinky1
 	@$(MD5SUM) test.hex
 	# test simulated boards on play.tinygo.org
 ifneq ($(WASM), 0)
--- a/src/machine/board_pico2_ice.go
+++ b/src/machine/board_pico2_ice.go
@@ -59,7 +59,8 @@ const (
 	GP46 Pin = 46
 	GP47 Pin = 47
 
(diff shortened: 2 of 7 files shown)


src/machine/board_pico2_ice.go
```diff
@@ -0,0 +1,173 @@
...
+	USB_BOOT = GP42
+
+	// Button
+	SW1     = GP42
+	BOOTSEL = GP42
+
+	// Tricolor LEDs
+	RED   Pin = GP1
+	GREEN Pin = GP0
+	BLUE  Pin = GP9
```
aykevl: These are usually called `LED_RED` etc for other boards.

src/machine/board_pico2_ice.go
```diff
@@ -0,0 +1,173 @@
...
+	// Onboard LED
+	LED Pin = GREEN
+
+	// Onboard crystal oscillator frequency, in MHz.
+	xoscFreq = 12 // MHz
+)
+
+// This board does not define default i2c pins.
+const (
+	I2C0_SDA_PIN Pin = 0
```
aykevl: Please use `NoPin` in that case.

src/machine/board_pico2_ice.go
```diff
@@ -0,0 +1,173 @@
...
+	GP21 Pin = GPIO21
+	GP22 Pin = GPIO22
+	GP23 Pin = GPIO23
+	GP24 Pin = GPIO24
+	GP25 Pin = GPIO25
+	GP26 Pin = GPIO26
+	GP27 Pin = GPIO27
+	GP28 Pin = GPIO28
+	GP29 Pin = GPIO29
+	GP30 Pin = 30
```
aykevl: Can you add these pin numbers (GPIO30 etc) to a new file for the rp2350b? So they can be shared across boards with an rp2350b.

aykevl (commented):
Can you also add this new board as a smoke test to GNUmakefile?

src/machine/board_pico2_ice.go
tinkerator: On closer inspection, these are already defined in `machine_rp2_2350b.go`. Will rely on that.

src/machine/board_pico2_ice.go
tinkerator: Will do.

src/machine/board_pico2_ice.go
tinkerator: Will fix.

tinkerator:
Thanks for the review. I've addressed all of these. Wasn't sure where to add the `smoketest`, but picking a point just before the `WASM` conditional (that `WASM` set of tests appear to fail for compiler flag issues). Hope that placement is OK.

Ran smoketest to completion with `make TINYGO=~/go/bin/tinygo WASM=0 XTENSA=0 smoketest`.

GNUmakefile
```diff
@@ -621,6 +621,8 @@ smoketest: testchdir
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=feather-rp2040      examples/device-id
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=pico2-ice	    examples/blinky1
```
aykevl: This is really a nitpick (sorry!) but can you make sure the alignment here is like the other smoke tests?

```suggestion
	$(TINYGO) build -size short -o test.hex -target=pico2-ice	        examples/blinky1
```

aykevl (approved):
Looks good to me, thank you!

tinkerator:
FWIW I was a bit confused by your tabbing suggestion at first. I guess, in my editor, I use default 8-character tabs. I think you must be using 4-character tabs. When I do that, the tabbing of the `GNUMakefile` is way more consistent!

This detail might be worth a mention in the https://tinygo.org/docs/guides/contributing/ guide. Hilariously, in my mailer, this is how your suggestion rendered: 

rendered

aykevl:
Oh, I didn't notice the tab character! Ideally no tab characters should be used there at all.

GNUmakefile
```diff
@@ -621,6 +621,8 @@ smoketest: testchdir
 	@$(MD5SUM) test.hex
 	$(TINYGO) build -size short -o test.hex -target=feather-rp2040      examples/device-id
 	@$(MD5SUM) test.hex
+	$(TINYGO) build -size short -o test.hex -target=pico2-ice	        examples/blinky1
```
aykevl:
```suggestion
	$(TINYGO) build -size short -o test.hex -target=pico2-ice           examples/blinky1
```

PR #5063, opened 2025-10-06
fix(rp2): possible integer overflow while computing factors for SPI  baudrate

oflebbe:
Because of integer overflows incorrect factors are calculated for baudrates which are a bit larger than integer multiples of 4194304.
For example for baudrates of 8_400_000 or 58_800_000. Fixed the same way in newer versions of the pico SDK.


aykevl:
Thank you! This looks like a good idea to me. 

(There might be more efficient ways of doing this, such as dividing some of these numbers by a power of two to avoid the overflow, but this shouldn't be performance sensitive anyway. Also, 64-bit multiplies are going to be decently fast on the RP2350 with the Cortex-M33).

